PR #1333, opened 2015-12-14
Add AsciiBinder info to templates & contrib guide

nhr:
@adellape please review.


adellape:
@nhr LGTM. Possibly worth having a note for people who were already using the previous method. Even if it's basically "if you were already using the old stuff, just install the new gem and start using the new asciibinder commands instead" (and if applicable, any other migration steps needed, though IIRC there really weren't any).

PR #1349, opened 2015-12-15
clarify limitrange behavior

bparees:
three fixes:

1) changed a / to a .
2) replaced <= with "lesser or equal to" because <= was being converted to a left-arrow symbol, so it was not clear it was intended as a mathematical expression.  In addition the table itself didn't really provide enough context to understand the content of the table was a mathematical constraint, so i think converting to pure english makes it clearer.
3) also updated the column heading to make it clearer what the column represents.

@derekwaynecarr ptal.

### DIFF
--- a/dev_guide/limits.adoc
+++ b/dev_guide/limits.adoc
@@ -74,16 +74,16 @@ Across all containers in a pod, the following must hold true:
 [cols="3a,8a",options="header"]
 |===
 
-|Constraint |Behavior
+|Constraint |Enforced Behavior
 
 |`*Min*`
-|Min[resource] <= container.resources.requests[resource] (required) <= container/resources.limits[resource] (optional)
+|Min[resource] less than or equal to container.resources.requests[resource] (required) less than or equal to container.resources.limits[resource] (optional)
 
 |`*Max*`
-|container.resources.limits[resource] (required) <= Max[resource]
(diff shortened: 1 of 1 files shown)


bparees:
@ncdc can you pick someone else to review this?

dev_guide/limits.adoc
```diff
@@ -74,16 +74,16 @@ Across all containers in a pod, the following must hold true:
 [cols="3a,8a",options="header"]
 |===
 
-|Constraint |Behavior
+|Constraint |Enforced Behavior
 
 |`*Min*`
-|Min[resource] <= container.resources.requests[resource] (required) <= container/resources.limits[resource] (optional)
+|Min[resource] lesser or equal to container.resources.requests[resource] (required) lesser or equal to container.resources.limits[resource] (optional)
```
ncdc: I would probably say "less than or equal to"

ncdc:
1 optional nit, otherwise LGTM

dev_guide/limits.adoc
bparees: agreed, will change.

bparees:
nit addressed.  @openshift/team-documentation ptal.

adellape:
LGTM, waiting on Travis.

PR #1354, opened 2015-12-16
Fixed yaml formatting in insecure repositories

soltysh:
My bad :( My previous PR had wrongly formatted yaml. Fixing it here.
@bfallonf ptal


bfallonf:
@soltysh cool cool. Thanks for that!

PR #1361, opened 2015-12-16
fixed the KUBERNETES_MASTER value

christian-posta:
in the docs to point to the CDK to match the value of docker... not a big deal either way as the docs aren't CDK specific, but just for consistency (and so i can copy/paste directly ) :)


adellape:
LGTM

PR #1364, opened 2015-12-17
Remove contradictory note about NFS and SELinux

pmorie:
@erinboyd @swagiaal @jeffvance


erinboyd:
I was under the impression with NFSv4 and OSE 3.1 we were able to respect the selinux labels that existed already on the NFS mount using that option in the pod. The subtlety was just that block devices does this automatically through 'relabeling' and shared just offers that extra level of security.
@swagiaal ?

swagiaal:
@erinboyd NFSv3 has limited SELinux support. The thing that is most notably lacking is MCS labels. They are not transmitted from server to client. Which means that you can give your pods access to the export but you cant give access to some pods and exclude others. So I would say this PR LGTM.

erinboyd:
@pmorie @swagiaal LGTM

bfallonf:
:cool: Merging.

PR #1370, opened 2015-12-21
Add section about using external artifacts during the build.

php-coder:
Document how to use external artifacts during the build.

@rhcarvalho PTAL.

Addressed to the following Trello card: https://trello.com/c/HTVCfQss/777-3-document-binary-build-via-dockerfile-and-custom-assemble

### DIFF
--- a/dev_guide/builds.adoc
+++ b/dev_guide/builds.adoc
@@ -1590,3 +1590,69 @@ Docker and Source builds set the following labels on output images:
 |*io.openshift.build.source-location*
 |Source URL for the build
 |===
+
+
+[[using-external-artifacts]]
+== Using External Artifacts During a Build
+
+It is not recommended to store binary files in a source repository. Therefore
+you may find it necessary to define a build which pulls additional files (such
+as Java jar dependencies) during the build process. How this is done depends on
+the build strategy you are using:
+
+For a `*Source*` build strategy you need to put appropriate shell commands into
+the `assemble` script.
+
+.`.sti/bin/assemble` file
+====
+
+[source,bash]
+----
+#!/bin/sh
+APP_VERSION=1.0
+wget http://repository.example.com/app/app-$APP_VERSION.jar -O app.jar
+----
+====
+
+.`.sti/bin/run` file
+====
+
+[source,bash]
+----
+#!/bin/sh
+exec java -jar app.jar
+----
+====
+
+For more information on how to control which assemble and run script is used by a source bulid,
+see: link:#override-builder-image-scripts[Override Builder Image Scripts]
+
+For a `*Docker*` build strategy you need to modify the `Dockerfile` and invoke
+shell commands with the https://docs.docker.com/engine/reference/builder/#run[`RUN` instruction].
+
+.Excerpt of `Dockerfile`
+====
+
+[source]
+----
+FROM jboss/base-jdk:8
+
+ENV APP_VERSION 1.0
+RUN wget http://repository.example.com/app/app-$APP_VERSION.jar -O app.jar
+
+EXPOSE 8080
+CMD [ "java", "-jar", "app.jar" ]
+----
+====
+
+In practice you may want to use an environment variable for the file location
+so that the specific file to be downloaded can be customized via an environment
+variable defined on the `BuildConfig` rather than updating the assemble script or
+`Dockerfile`.
+
+You could choose between a different ways of defining environment variables:
+
+- link:#environment-files[put into the `.sti/environment` file] (only for a `*Source*` build strategy)
+- link:#buildconfig-environment[set in `*BuildConfig*`]
+- link:../cli_reference/basic_cli_operations.html#build-and-deployment-cli-operations[provide explicitly with `*oc start-build --env*`]
+  (only for builds that are triggered manually)


dev_guide/builds.adoc
```diff
@@ -1557,3 +1557,84 @@ Docker and Source builds set the following labels on output images:
...
+the `assemble` script.
+
+.`.sti/bin/assemble` file
+====
+
+[source,bash]
+----
+#!/bin/sh -x
+wget http://code.jquery.com/jquery-2.1.4.min.js
+----
```
rhcarvalho: This assemble script is not really usable the way it is, it would override the default assemble script and not really "build".
I'd try to make it a more realistic example.

dev_guide/builds.adoc
```diff
@@ -1557,3 +1557,84 @@ Docker and Source builds set the following labels on output images:
...
+For `*Docker*` build strategy you need to modify `Dockerfile` and invoke shell
+commands with https://docs.docker.com/engine/reference/builder/#run[`RUN` instruction].
+
+.Excerpt of `Dockerfile`
+====
+
+[source]
+----
+RUN wget http://code.jquery.com/jquery-2.1.4.min.js
+----
```
rhcarvalho: Also here, I'd try to make a more usable example. Can we layer on top of an existing image and add a new binary or other artifact?
Don't need to be "copy-pastable", but at least something closer to a legit use:

```Dockerfile
FROM base-image:tag
...
RUN curl -O https://my.artifact.repository.com/builds/latest/app.war && do-something with app.war
...
CMD ["java", "..."]
```

dev_guide/builds.adoc
```diff
@@ -1557,3 +1557,84 @@ Docker and Source builds set the following labels on output images:
...
+====
+
+Both of the examples will work, but what if you don't want to have a hard coded
+version of the artifact?
```
rhcarvalho: Don't ask questions in the documentation

dev_guide/builds.adoc
```diff
@@ -1557,3 +1557,84 @@ Docker and Source builds set the following labels on output images:
...
+
+.  provide them explicitly with `*oc start-build --env*` (only for builds that are triggered manually)
++
+[source,shell]
+----
+oc start-build --env=JQUERY_VERSION=2.1.4 <BuildConfigName>
+----
+
+Also, consider one more way: you could link:#override-builder-image-scripts[create your custom version of the `assemble` script]
+and use it instead of the default or from the repository.
```
rhcarvalho: Wait, this was the first example, wasn't it?!

dev_guide/builds.adoc
```diff
@@ -1557,3 +1557,84 @@ Docker and Source builds set the following labels on output images:
...
+
+For the `*Docker*` strategy there is only one option and is to link:#docker-strategy-environment[use `*BuildConfig*`]:
```
rhcarvalho: Why `oc start-build --env` would not work?!
And obviously the env can be set in the Dockerfile as well, though less useful (we could probably make it part of the original example Dockerfile above).

dev_guide/builds.adoc
php-coder: @rhcarvalho As far I understand, no, it's a different story. The first one is when you're using `.sti/bin/assemle` script that resides in the repository. And the latter is when you're put this script to the internet and configure `BuildConfig` to download and use it (this way probably could be helpful when you don't have an access to the repository).

rhcarvalho: I didn't quite understand that. Where in the docs above is it mentioned that the assemble script is in a remote location?!

php-coder: There is a link: https://docs.openshift.org/latest/dev_guide/builds.html#override-builder-image-scripts

rhcarvalho: Hmm, but reading 

> Also, consider one more way: you could create your custom version of the `assemble` script and use it instead of the default or from the repository.

That doesn't suggest what you're explaining here in the comments.

php-coder:
@rhcarvalho Thanks for the feedback. I've addressed all of your comments, could you review it again?

dev_guide/builds.adoc
```diff
@@ -1557,3 +1557,114 @@ Docker and Source builds set the following labels on output images:
...
 |Source URL for the build
 |===
+
+
+[[using-external-artifacts]]
+= Using External Artifacts During The Build
+
+In order to use external artifacts during the build you don't need any special
+support from OpenShift platform. You could download and use any artifacts
+as you want.
```
rhcarvalho: Isn't the assemble script, for example, a "special platform support" mechanism?

rhcarvalho: I'd start with a more objective sentence explaining what this section is about and why you would want to "Use external artifacts during a build".
A user reading the docs in a rush wants to tell if this is the right place to read and what he'll get out of continuing to read the next paragraphs.

php-coder: I mean that there is no special commands, settings and so on. How I could improve this?

dev_guide/builds.adoc
```diff
@@ -1557,3 +1557,114 @@ Docker and Source builds set the following labels on output images:
...
+
+For `*Source*` build strategy you need to put appropriate shell commands into
+the `assemble` script.
+
+.`.sti/bin/assemble` file
+====
+
+[source,bash]
+----
+#!/bin/sh -ex
```
rhcarvalho: Keep it simple, the `-ex` can be omitted here.

dev_guide/builds.adoc
```diff
@@ -1557,3 +1557,114 @@ Docker and Source builds set the following labels on output images:
...
+
+For `*Docker*` build strategy you need to modify `Dockerfile` and invoke shell
+commands with https://docs.docker.com/engine/reference/builder/#run[`RUN` instruction].
+
+.Excerpt of `Dockerfile`
+====
+
+[source]
+----
+FROM java:8
```
rhcarvalho: Can we use a Java image we provide? `java:8` uses ROOT users, among other pitfalls we already handle in our images.
This is our Java builder image: `fabric8io/java-main`, https://github.com/fabric8io/java-main/tree/master/

But we might want to use another image that is not a builder (need to find the name/link).

rhcarvalho:
For completeness, I propose we show a custom `run` script that leverages what was installed by the `assemble`. Just like in the Dockerfile we set CMD to use artifacts that were pulled in during build.

dev_guide/builds.adoc
bparees: "It is not recommended to store binary files in a source repository.  Therefore you may find it necessary to define a build which pulls additional files (such as Java jar dependencies) during the build process.  How this is done depends on the build strategy you are using:"

dev_guide/builds.adoc
```diff
@@ -1556,3 +1556,109 @@ Docker and Source builds set the following labels on output images:
 |*io.openshift.build.source-location*
 |Source URL for the build
 |===
+
+
+[[using-external-artifacts]]
+= Using External Artifacts During The Build
```
bparees: s/The/a/

dev_guide/builds.adoc
```diff
@@ -1556,3 +1556,109 @@ Docker and Source builds set the following labels on output images:
...
+
+In order to use external artifacts during the build you don't need any special
+support from OpenShift platform. You could download and use any artifacts
+as you want.
+
+For `*Source*` build strategy you need to put appropriate shell commands into
```
bparees: "For a"

dev_guide/builds.adoc
```diff
@@ -1556,3 +1556,109 @@ Docker and Source builds set the following labels on output images:
...
+====
+
+[source,bash]
+----
+#!/bin/sh
+exec java -jar app.jar
+----
+====
+
+For `*Docker*` build strategy you need to modify `Dockerfile` and invoke shell
```
bparees: "For a" 
"modify the"
"with the"

dev_guide/builds.adoc
```diff
@@ -1556,3 +1556,109 @@ Docker and Source builds set the following labels on output images:
...
+ENV APP_VERSION 1.0
+RUN wget http://repository.example.com/app/app-$APP_VERSION.jar -O app.jar
+
+EXPOSE 8080
+CMD [ "java", "-jar", "app.jar" ]
+----
+====
+
+Both of the examples will work, but you could improve it by replacing local
+variable by the environment variable.
```
bparees: "In practice you may want to use an environment variable for the file location so that the specific file to be downloaded can be customized via an environment variable defined on the BuildConfig rather than updating the assemble script or Dockerfile"

dev_guide/builds.adoc
```diff
@@ -1556,3 +1556,109 @@ Docker and Source builds set the following labels on output images:
...
+      - name: APP_VERSION
+        value: "1.0"
+----
+
+.  provide them explicitly with `*oc start-build --env*` (only for builds that are triggered manually)
++
+[source,shell]
+----
+oc start-build --env=APP_VERSION=1.0 <BuildConfigName>
+----
```
bparees: dump this whole discussion of environment variables and just link to the appropriate documentation.

dev_guide/builds.adoc
```diff
@@ -1556,3 +1556,109 @@ Docker and Source builds set the following labels on output images:
...
+
+If you can't modify source repository and put `assemble` script there, consider
+one more way: publish custom version of the `assemble` script to some server and
+link:#override-builder-image-scripts[configure `BuildConfig` to download and use
+this script instead of the default or from the repository].
```
bparees: dump this (i'll make another comment above about how to handle it better)

dev_guide/builds.adoc
```diff
@@ -1556,3 +1556,109 @@ Docker and Source builds set the following labels on output images:
...
+.`.sti/bin/run` file
+====
+
+[source,bash]
+----
+#!/bin/sh
+exec java -jar app.jar
+----
+====
+
```
bparees: add: "For more information on how to control which assemble and run script is used by a source bulid, see: https://docs.openshift.org/latest/dev_guide/builds.html#override-builder-image-scripts"

dev_guide/builds.adoc
```diff
@@ -1556,3 +1556,109 @@ Docker and Source builds set the following labels on output images:
...
+oc start-build --env=APP_VERSION=1.0 <BuildConfigName>
+----
+
+If you can't modify source repository and put `assemble` script there, consider
+one more way: publish custom version of the `assemble` script to some server and
+link:#override-builder-image-scripts[configure `BuildConfig` to download and use
+this script instead of the default or from the repository].
+
+For the `*Docker*` strategy there are only two options:
+
```
bparees: the content after this line looks like a cut+paste of env discussion, not dockerfile overriding.

in anycase move the explanation of how to override the dockerfile directly below the example dockerfile, and if there is appropriate documentation that already exists, link to it rather than re-summarizing here.

bparees:
@php-coder bump

php-coder:
@bparees I remember about it. I will take a look on it tomorrow, ok?

bparees:
@php-coder np.

php-coder:
@bparees I've update the doc according to your comments (thanks! they were very helpful). PTAL.

dev_guide/builds.adoc
```diff
@@ -1577,3 +1577,69 @@ Docker and Source builds set the following labels on output images:
...
+====
+
+[source,bash]
+----
+#!/bin/sh
+exec java -jar app.jar
+----
+====
+
+For more information on how to control which assemble and run script is used by a source bulid,
```
rhcarvalho: s/bulid/build

dev_guide/builds.adoc
```diff
@@ -1577,3 +1577,69 @@ Docker and Source builds set the following labels on output images:
...
+CMD [ "java", "-jar", "app.jar" ]
+----
+====
+
+In practice you may want to use an environment variable for the file location
+so that the specific file to be downloaded can be customized via an environment
+variable defined on the `BuildConfig` rather than updating the assemble script or
+`Dockerfile`.
+
+You could choose between a different ways of defining environment variables:
```
rhcarvalho: "There are different ways to define build environment variables:"

dev_guide/builds.adoc
```diff
@@ -1577,3 +1577,69 @@ Docker and Source builds set the following labels on output images:
...
+
+- link:#environment-files[put into the `.sti/environment` file] (only for a `*Source*` build strategy)
```
rhcarvalho: Adding a `.sti/environment` file in the source repository (only for `*Source*` build strategy)

dev_guide/builds.adoc
```diff
@@ -1577,3 +1577,69 @@ Docker and Source builds set the following labels on output images:
...
+- link:#buildconfig-environment[set in `*BuildConfig*`]
+- link:../cli_reference/basic_cli_operations.html#build-and-deployment-cli-operations[provide explicitly with `*oc start-build --env*`]
+  (only for builds that are triggered manually)
```
rhcarvalho: I would suggest reordering this list so that the first item is the one that applies in all cases:

* set in BuildConfig
* set in .sti/environment
* set in start-build

Obviously the env can also be set and updated in the Dockerfile itself, but I imagine that would be a dumb idea.

bparees:
lgtm.  @openshift/team-documentation ptal.

dev_guide/builds.adoc
```diff
@@ -1577,3 +1577,69 @@ Docker and Source builds set the following labels on output images:
 |*io.openshift.build.source-location*
 |Source URL for the build
 |===
+
+
+[[using-external-artifacts]]
+= Using External Artifacts During a Build
```
adellape: Needs two `=`s.

adellape:
@php-coder One comment inline, otherwise LGTM.

php-coder:
@adellape Fixed. Could you take a look again?

adellape:
@php-coder :+1: Thanks.

PR #1383, opened 2015-12-30
Troubleshooting: recommend diagnostics

sosiouxme:
Recommend diagnostics in the context of troubleshooting with a bit of actual description.

I looked for a better place to put this, but the troubleshooting section seemed best at the moment. It didn't seem to warrant a separate topic. But feel free to pull this out somewhere else if you think it best.

BTW, I think this troubleshooting doc has already moved well beyond just SDN troubleshooting and might be better named more generically as "Troubleshooting" or "Troubleshooting the cluster".

### DIFF
--- a/admin_guide/sdn_troubleshooting.adoc
+++ b/admin_guide/sdn_troubleshooting.adoc
@@ -25,6 +25,94 @@ will work through some scenarios that, hopefully, will cover the
 majority of cases.  If your problem is not covered, the tools and
 concepts that are introduced should help guide debugging efforts.
 
+== Run Diagnostics First
+
+Diagnostics are intended to assist with troubleshooting by detecting
+and explaining many common problems with the OpenShift cluster and
+ultimately automating troubleshooting effort. The diagnostic tool is
+under development along with the project and is currently considered
+experimental; it is included in the `openshift` binary itself so that
+you can run it wherever you have a server or client installed.
+
+=== Expected Diagnostics Usage
+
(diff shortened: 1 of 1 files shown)


adellape:
@sosiouxme I think there's enough here to warrant a stand-alone "Running Diagnostics" topic near the end of the Cluster Administration book. The "Running Diagnostics First" section in this "Troubleshooting OpenShift SDN" topic could then just be some brief explanation + a link to a "Running Diagnostics" topic.

And eventually when we have a separate Troubleshooting book, we could move both of these topics there.

WDYT?

sosiouxme:
@adellape I guess this is long enough to justify its own page; just seemed like there were already an awful lot of admin topics, maybe it's time for another sublevel...

adellape:
@sosiouxme Definitely; the Cluster Admin guide should be getting grouped into better subcategories in the coming weeks.

ahardin-rh:
@sosiouxme @adellape Is this still relevant/correct? Raising this PR from the dead. :ghost:  Thanks.

sosiouxme:
@ahardin-rh I think it's still relevant, but would need a little jiggering as "openshift ex diagnostics" became "oadm diagnostics" and then "oc adm diagnostics". Also @adellape 's suggestion for having it be a separate page would be good.

ahardin-rh:
Closing this and continuing work in https://github.com/openshift/openshift-docs/pull/3302

PR #1388, opened 2016-01-04
syncing_groups_with_ldap: typo fixes, cleanups

Miciah:
Use consistent capitalisation for references to OpenShift "Group" objects, as opposed to LDAP "groups" or "groups" in the general sense.

Use consistent styling for file names.

Reword some statements to avoid confusing sentence structures (such as dangling non-participial modifiers, lacking parallelism, or ending a relative clause with a stranded preposition and then following it with a prepositional phrase).

Delete a spurious "at".

Use consistent capitalisation for "augmented Active Directory".

Consistently write "RFC 2307" (as opposed to "RFC2307").

Fix some typos: "intOrgPerson", "idenitifer", "syncheonized", and "Adminstrators".
## 

@stevekuznetsov, should "objectclass" be changed to "objectClass" in LDAPSyncConfigs? AFAICT compliant LDAP servers should not care, and we use "objectClass" elsewhere.

### DIFF
--- a/install_config/syncing_groups_with_ldap.adoc
+++ b/install_config/syncing_groups_with_ldap.adoc
@@ -10,13 +10,13 @@
 toc::[]
 
 == Overview
-As an OpenShift administrator, grouping users allows you to better manage them,
-change their permissions, or to enhance collaboration. Your organization may
-have already created user groups and stored them in an LDAP server. OpenShift
-can sync those LDAP records with internal OpenShift records, allowing you
-to simply manage your groups in one place. OpenShift currently supports group
-sync with LDAP servers using three common schemas for defining group
-membership: RFC 2307, Active Directory, and augmented Active Directory.
+As an OpenShift administrator, you can use groups to manage users, change their
+permissions, and enhance collaboration. Your organization may have already
+created user groups and stored them in an LDAP server. OpenShift can sync those
+LDAP records with internal OpenShift records, enabling you to manage your groups
+in one place. OpenShift currently supports group sync with LDAP servers using
+three common schemas for defining group membership: RFC 2307, Active Directory,
+and augmented Active Directory.
 
 [NOTE]
 ====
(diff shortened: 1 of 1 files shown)


adellape:
@Miciah Whoops, I just merged https://github.com/openshift/openshift-docs/pull/1387 which took care of some of these!

Miciah:
Oh, snap! I'll rebase now.  Sorry about that.

Miciah:
Rebased and kept a few changes:

    Use consistent capitalisation for references to OpenShift "Group" objects,
    as opposed to LDAP "groups" or "groups" in the general sense.
    
    Use consistent styling for file names.
    
    Reword some statements to avoid confusing sentence structures (such as
    dangling non-participial modifiers, lacking parallelism, or ending
    a relative clause with a stranded preposition and then following it with
    a prepositional phrase).
    
    Delete a spurious "at".
    
    Use consistent capitalisation for "augmented Active Directory".
    
    Consistently write "RFC 2307" (as opposed to "RFC2307").
    
    Fix a typo: "intOrgPerson" should be "inetOrgPerson".

I'm still not sure about "objectClass".  Also, thinking about it more, I wonder whether we should change "Group records" to "Group objects" or just "Groups"—or maybe it would be clearer just to switch to lower-case "group" everywhere (perhaps specifying "OpenShift group" as needed to disambiguate).

install_config/syncing_groups_with_ldap.adoc
```diff
@@ -637,8 +635,8 @@ augmentedActiveDirectory:
     groupMembershipAttributes: [ testMemberOf ] <4>
```
stevekuznetsov: @Miciah We may want to change `testMemberOf` to a more appropriate attribute name. I believe `testMemberOf` was chosen to not conflict with `memberOf` on the test LDAP server we have set up, which calculates `memberOf` instead of storing it as an attribute.

stevekuznetsov:
@Miciah  thanks for taking the time, differentiation between Groups, groups, records, entries, *etc* was driving me insane even in the proposal phase of LDAP integration. I don't really know the best way to keep those concepts separate.

Re: `objectClass` - I don't think LDAP cares about case, but for consistency's sake let's use camel-case.

stevekuznetsov:
Also in some cases I may have written "LDAP group entry" - I think in those cases I was intentionally referring to an entry on the LDAP server (as opposed to AD & eAD schema).

adellape:
:+1:

PR #1398, opened 2016-01-07
fuse.adoc: Fix a misspelling of "OpenShift"

Miciah:


tnguyen-rh:
Label rationale:
PR #1332 introduced the typo, so we copy its label as well.

mfojtik:
We should probably avoid using "shit" in GIT commit messages ;-) Now it will be there forever.

Miciah:
The diffs are just as much a permanent part of the Git history, so there's no getting around that.  I'll change the PR title though...

tnguyen-rh:
@adellape 
Looks like @tpoitras did the needful w/ #2490 so we can revert the orig labels/milestone.

PR #1399, opened 2016-01-07
Add docs for adding SCC to user/group

pmorie:
Add docs for adding SCC to users/groups.

@adellape @liggitt


adellape:
:+1: Thanks.

PR #1401, opened 2016-01-07
Add note about nodeName.

jsafrane:
nodeName must be set to OpenStack InstanceID of the node, otherwise
OpenShift's provider for OpenStack cloud can't be initialized.

This is not always necessary, but it's much easier to document it as
mandatory (or recommended) instead of describing cases where OpenShift could
work event without 'nodeName' in the node config.


adellape:
:+1: Thanks.

PR #1410, opened 2016-01-08
Secret example metadata should contain name in a "metadata" field

tiwillia:
Kubernetes v1.1 requires that the name of a secret be placed in the metadata field. The current example secret will fail to create for this reason.

See http://kubernetes.io/v1.1/docs/user-guide/secrets.html


adellape:
:+1:

PR #1412, opened 2016-01-08
logging conf: saner way to edit scc

sosiouxme:
It turns out you don't have to `oc edit` to add a user to SCC. This is far easier to explain and harder to do wrong. It applies since the release of 3.1 apparently.


adellape:
Cool, some other spots can get updated to use this too. Thanks.

PR #1415, opened 2016-01-12
v2 vs v3 topic edits

bfallonf:
@adellape PTAL

Or did you mean this should be conditionalized so it doesn't show up in Dedicated?

### DIFF
--- a/_build_cfg.yml
+++ b/_build_cfg.yml
@@ -34,8 +34,6 @@ Distros: openshift-enterprise,openshift-dedicated
 Topics:
   - Name: Overview
     File: index
-  - Name: OpenShift v2 vs. OpenShift v3
-    File: v2_vs_v3
   - Name: OpenShift Enterprise 3.1 Release Notes
     File: ose_3_1_release_notes
     Distros: openshift-enterprise
@@ -45,6 +43,8 @@ Topics:
   - Name: xPaaS Release Notes
     File: xpaas_release_notes
     Distros: openshift-enterprise
+  - Name: Comparing OpenShift Enterprise 2 and OpenShift Enterprise 3
+    File: v2_vs_v3
 
 ---
 Name: What's New?
--- a/release_notes/v2_vs_v3.adoc
+++ b/release_notes/v2_vs_v3.adoc
@@ -1,4 +1,4 @@
-= Comparing OpenShift version 2 and OpenShift version 3
+= Comparing OpenShift Enterprise 2 and OpenShift Enterprise 3
 {product-author}
 {product-version}
 :icons: font
@@ -11,10 +11,16 @@
 toc::[]
 
 == Overview
-While still an application platform, OpenShift v3 is a very different product than OpenShift v2. 
-Many of the same terms are used, and the same functions are performed, but the terminology can be different, and behind the scenes things may be happening very differently.
+OpenShift version 3 (v3) is a very different product than OpenShift version 2 (v2). 
+Many of the same terms are used, and the same functions are performed, but the terminology can be different, and behind the scenes things may be happening very differently. 
+Still, OpenShift remains an application platform.
 
 This topic discusses these differences in detail, in an effort to help OpenShift users in the transition from OpenShift v2 to OpenShift v3.
+
+ifdef::openshift-dedicated[]
+[NOTE]
+OpenShift Dedicated 3 follows the numbering of the product's major version, and uses the same code base as OpenShift Enterprise 3.
+endif::[]
 
 == Architecture Changes
 


adellape:
This may warrant an `openshift-dedicated`-only line or two to give some context as to why there's an OSE topic in the OSD docs. For example, see https://docs.openshift.com/dedicated/3.1/release_notes/osd_3_1_release_notes.html#how-is-openshift-dedicated-different-than-openshift-enterprise.

bfallonf:
@adellape Like this? I wasn't sure what else to put in, so it's pretty much a copy/paste from the link.

_build_cfg.yml
```diff
@@ -34,7 +34,7 @@ Distros: openshift-enterprise,openshift-dedicated
 Topics:
   - Name: Overview
     File: index
-  - Name: OpenShift v2 vs. OpenShift v3
+  - Name: Comparing OpenShift Enterprise 2 and OpenShift Enterprise 3
```
adellape: Sorry, I meant to mention yesterday also a suggestion to move this topic to the end of the Release Notes category vs the beginning. Or really, would this better belong in the Architecture category? Seems it would go well right after Architecture -> Overview, right before the Infrastructure Components subcategory. WDYT? Was there a request somewhere about it being specifically in the Rel Notes section?

release_notes/v2_vs_v3.adoc
```diff
@@ -16,6 +16,10 @@ Many of the same terms are used, and the same functions are performed, but the t
 
 This topic discusses these differences in detail, in an effort to help OpenShift users in the transition from OpenShift v2 to OpenShift v3.
 
+ifdef::openshift-dedicated[]
+OpenShift Dedicated uses the same code base as OpenShift Enterprise 3. OpenShift Dedicated is hosted on the public cloud and managed by Red Hat. An OpenShift Enterprise 3 subscription entitles you to host and manage the software on your infrastructure of choice.
+endif::[]
```
adellape: Maybe this can be a Note box. Something like:

> "OpenShift v2" and "OpenShift v3" refer to the second and third architectural revisions of OpenShift, respectively. OpenShift Dedicated 3, referring to the product's major version, uses the same code base as OpenShift Enterprise 3, both based on OpenShift v3.

Gives some context to why this topic is appearing in the OSD docs, and also explains what "v2" and "v3" are referring to. I don't believe that's explained anywhere in the OSE 3.1 docs anymore, since we axed the "What's New" category post-3.0 (there was a similar Note at the end of https://docs.openshift.org/latest/whats_new/index.html, which persists in the Origin docs).

If you're into something like that, I could possibly see it moved up to become the first paragraph instead of a Note box, too.

_build_cfg.yml
bfallonf: I moved it to the bottom of the rel notes section. It's in the rel notes section because this is the information from the What's New section leftover from previously. I don't think Architecture is a better fit, as it's not so much about the architecture itself, but rather what's changed.

release_notes/v2_vs_v3.adoc
bfallonf: Changed to:

"OpenShift Dedicated 3 follows the numbering of the product's major version, and uses the same code base as OpenShift Enterprise 3."

I don't think the sentence explaining what "v2/3" is too relevant, as it's pretty self explanatory. I did, however, change the first instance of "OpenShift v2 and v3" to OpenShift version 2 and OpenShift version 3" just in case. Like how you spell out abbreviations the first time, then use them then on.

Kept it as a note box as well.

bfallonf:
@adellape Comments addressed. Let me know what you think.

release_notes/v2_vs_v3.adoc
adellape: There's an earlier use of v2/v3 in the first para/sentence above this, if you were wanting to spell it out in their first appearance.

adellape: On its own (without juxtaposing the "v#" language), I'm not sure that "OpenShift Dedicated 3 follows the numbering of the product's major version" is worth saying. At that point, I'd probably just drop it and leave the bit about the code base.

bfallonf: Done, with some rewording.

I think with just the "using the same codebase as v3" sentence in there, it's probably not worth making the point at all. It says in the Overview and the Rel Notes that this is a thing. If anything, the important point to make would be that "This topic is relevant because Dedicated 3 uses the same codebase as Enterprise 3", which I think it's doing atm, but really, I'd be happy to take the notebox out and maybe even take this topic out of the Dedicated books.

bfallonf:
@adellape One comment on your comment. I hope we're on the same page. Let me know.

adellape:
@bfallonf I'm good with this. Thanks for indulging. :smiley_cat:

I'll merge this now, and we can ponder further on whether to keep it in the Dedicated book later.

bfallonf:
@adellape Cool cool. Thanks.

PR #1420, opened 2016-01-12
Include note about required hostnames for the custom Metric certificates

mwringe:
In reference to https://lists.openshift.redhat.com/openshift-archives/users/2016-January/msg00071.html 

See also https://github.com/openshift/origin-metrics/pull/59


adellape:
:+1:

PR #1421, opened 2016-01-13
Add link to compute resources from quotas & limits

sspeiche:
Helps with setting context when users jump directly into section Quotas and Limits


adellape:
:+1:

PR #1426, opened 2016-01-14
Update statement about no builder images supporting incremental flag

### LINKED ISSUE #1414: Updated docs to match new incremental feature of EAP 1.2 images
Currently [the docs say](https://docs.openshift.org/latest/dev_guide/builds.html#incremental-builds):

> The S2I images provided by OpenShift do not implement artifact reuse, so setting incremental to true will have no effect on builds using those builder images.

Which is no longer true. So I recommend removing this statement.

sspeiche:
Fixes #1414

### DIFF
--- a/dev_guide/builds.adoc
+++ b/dev_guide/builds.adoc
@@ -199,9 +199,8 @@ previously-built images. To create an incremental build, create a
 }
 ----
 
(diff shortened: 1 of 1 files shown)


adellape:
:+1:

PR #1307, opened 2015-12-09
arch/additional_concepts 3.1 changes to SCC settings

jeffvance:
quick pr to get scc section current to version 3.1

PR #1427, opened 2016-01-14
Add docs on using anyuid SCC for dockerhub images

pmorie:
Add some guidance for using dockerhub images to prevent openshift/origin#6282

@pweil- @jeffvance @adellape

### DIFF
--- a/admin_guide/manage_scc.adoc
+++ b/admin_guide/manage_scc.adoc
@@ -274,6 +274,24 @@ This allows images to run as the root UID if no *USER* is specified in the
 *_Dockerfile_*.
 ====
 
+[[enable-dockerhub-images-that-require-root]]
+
+=== Enable dockerhub images that require root
+
+Some Dockerhub images (examples: `postgres` and `redis`) require root access and have certain
+expectations about how volumes are owned.  For these images, add the service account to the
+`anyuid` SCC.
+
+----
+$ oc edit scc anyuid
+----
+
+Add the following under `users`:
+
+----
+   - system:serviceaccount:My_Project:My_SVCACCT
+----
+
 [[use-mount-host-on-the-registry]]
 
 === Use --mount-host on the Registry


pweil-:
LGTM.  An aside, I think this "how do I" section is getting a bit muddy now that we've introduced new SCCs.  For instance, the section directly above this addition is also addressing running as any UID using the restricted SCC which is effectively what we're describing in the proposed new section (with slight differences due to the SCC definition that doesn't relate to UIDs)

pmorie:
@pweil- Where are all the SCCs we provide listed?

pweil-:
> @pweil- Where are all the SCCs we provide listed?

https://github.com/openshift/openshift-docs/pull/1307/files#diff-ac08c0344bc91609b46aae5f51152c8dR249

pmorie:
@pweil- I think we should give those treatment in the admin guide too, probably, or at least link to it from there.  Which sounds better to you? I'll open a PR.

pweil-:
@pmorie - it would be good to frame them in a use case fashion in the admin guide.  Ie. "I want to run a secure cluster" and "I'm running a dev cluster".

pmorie:
I'm thinking we need to rework this doc into something like:

1.  These are the SCCs we provide
2.  This is what you should use them for
3.  Suggested policy combinations for different cluster use-cases

Also, do we need to tell people to edit the SCC just to add a user or
should we have people use the oadm shortcuts for adding a user to an SCC?

pweil-:
> Also, do we need to tell people to edit the SCC just to add a user or
> should we have people use the oadm shortcuts for adding a user to an SCC?

we should use the command now that it is available

pmorie:
@adellape PTAL

admin_guide/manage_scc.adoc
```diff
@@ -274,6 +274,24 @@ This allows images to run as the root UID if no *USER* is specified in the
 *_Dockerfile_*.
 ====
 
+[[enable-dockerhub-images-that-require-root]]
+
+=== Enable dockerhub images that require root
+
+Some Dockerhub images (examples: `postgres` and `redis`) require root access and have certain
+expectations about how volumes are owned.  For these images, add the service account to the
+`anyuid` SCC.
```
jeffvance: The "anyuid" SCC is one of the new predefined SCCs in version 3,1; however, it is not listed in the 2 predefined SCCs shown in the ```oc get scc``` commands above. I think the outputs of the two ```oc get scc``` commands need to be updated.

pmorie: Good catch, I'll update.

pmorie:
@adellape Could we merge this one?  I'm going to probably rewrite this page in the next week, but I'd like to get this information into the online docs.

smarterclayton:
I was going to open an issue that contained this exact discussion, but I see that it has already happened.  Excellent work, carry on, chop chop.

I'd like have a clear intro at the beginning of the section explaining the basics of how we restrict security, then a list of the SCCs, and walk people through the common steps (that we have) but with enough detail so you can fix any problems you hit.  Definitely need to be able to give this page to *anyone* and say "if you need X, grant the project or service account this SCC".  Also, be sure to use the `-z SERVICEACCOUNT` option, not directly inputting the full service account name.

pmorie:
why is there no emoji for the concept of two horses in a harness? WHY

adellape:
@pmorie K, I'll merge as-is. I take it the rewrite you had in mind would also switch to using `add-scc-to-user`/`add-scc-to-group`?

smarterclayton:
One word: PETA

PR #1429, opened 2016-01-14
OSC_EDITOR -> OC_EDITOR

Miciah:
Fix references to the outdated `OSC_EDITOR` environment variable to instead refer to the current `OC_EDITOR` variable.

The corresponding code change was https://github.com/openshift/origin/pull/3086.


bfallonf:
Thanks @Miciah . I can see the code change was from July last year, so I've attached the 3.0 label so its goes into the 3.0 books too. Let me know if this is wrong in anyway.

PR #1431, opened 2016-01-15
Fix default node config file path

codificat:
A minor change here: the default path to the node's config file does not contain the node's hostname.


adellape:
:+1: Thanks.

PR #1445, opened 2016-01-18
describe imagesource inputs

bparees:
adds doc for the new input from an image feature.  do not merge until https://github.com/openshift/origin/pull/6546 has merged.

### DIFF
--- a/dev_guide/builds.adoc
+++ b/dev_guide/builds.adoc
@@ -978,6 +978,59 @@ containing an inline Dockerfile that potentially replaces one within the
 contents of the binary archive.
 ====
 
+[[image-source]]
+
+== Image Source
+
+Additional files can be provided to the build process via images.  Input
+images are referenced in the same way the `From` and `To` image targets
+are defined.  This means both docker images and image stream
+tags can be referenced.  In conjunction with the image, you must provide one or 
+more path pairs to indicate the path of the files/directories to copy out of the
+image and the destination to place them in the build context.
+
+The source path can be any absolute path within the image specified.  The destination
+must be a relative directory path.  At build time, the image will be loaded
+and the indicated files/directories will be copied into the context directory
+of the build process.  This is the same directory into which the source repository
+content (if any) is cloned.
+
+Image inputs are specified in the `*source*` definition of the `*BuildConfig*`:
+
+====
+
+----
+    source:
+      git:
+        uri: https://github.com/openshift/ruby-hello-world.git
+      images: <1>
+      - from: <2>
+          kind: ImageStreamTag
+          name: myinputimage:latest
+          namespace: mynamespace
+        paths: <3>
+        - destinationDir: injected/dir <4>
+          sourcePath: /usr/lib/somefile.jar <5>
+      - from:
+          kind: ImageStreamTag
+          name: myotherinputimage:latest
+          namespace: myothernamespace
+        pullSecret: mysecret <6>
+        paths:
+        - destinationDir: injected/dir
+          sourcePath: /usr/lib/somefile.jar
+----
+
+<1> An array of one or more input images and files.
+<2> A reference to the image containing the files to be copied.
+<3> An array of source/destination paths.
+<4> The directory relative to the build root where the build process can access the file.
+<5> The location of the file to be copied out of the referenced image.
+<6> An optional secret provided if credentials are needed to access the input image.
+====
+
+NOTE: This feature is not supported for builds using the `*Custom*` strategy.
+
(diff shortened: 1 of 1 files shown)


bparees:
@csrwng ptal

dev_guide/builds.adoc
```diff
@@ -978,6 +978,59 @@ containing an inline Dockerfile that potentially replaces one within the
...
+== Image Source
+
+Additional files can be provided to the build process via images.  Input
+images are referenced in the same way the `From` and `To` image targets
+are defined.  This means both docker images and image stream
+tags can be referenced.  In conjunction with the image, you must provide one or 
+more path pairs to indicate the path of the files/directories to copy out of the
+image and the destination to place them in the build context.
+
+The source path can be any location within the image specified.  The destination
```
csrwng: You may want to mention that source should be an absolute path

dev_guide/builds.adoc
```diff
@@ -978,6 +978,59 @@ containing an inline Dockerfile that potentially replaces one within the
...
+        pullSecret: mysecret <6>
+        paths:
+        - destinationDir: injected/dir
+          sourcePath: /usr/lib/somefile.jar
+----
+
+<1> An array of one or more input images and files.
+<2> A reference to the image containing the files to be copied.
+<3> An array of source/destination paths.
+<4> The build process (assemble script or Dockerfile) can reference the file at this relative path.
```
csrwng: A description of what it is may be more straightforward:
'The directory relative to the build root where the build process can access the file'

One thing to note that I just realized is that the location is relative to the place where we extract the source and not the directory that we use for the build. For example, you may have a git repo with a  context dir of 'foo/bar' and an image source with destinationDir of 'foo/baz'. The files from the image will end up copied outside of the directory we're using for building.

dev_guide/builds.adoc
bparees: will do

dev_guide/builds.adoc
bparees: hm. will they?  we'll clone the repo, then cd into foo/bar and move that content into the staging folder

then we'll copy the image paths into the staging folder

then we tar up the staging folder and inject it as the conext directory for the build.

right?

so you'd have 
<contextdir>/source
<contextdir>/foo/baz

bparees: (will change the description to what you suggest, i was definitely struggling w/ that one)

csrwng: We extract both git and image source (to the same dir) before moving content to the staging folder:
https://github.com/openshift/origin/blob/master/pkg/build/builder/source.go#L53-L75

bparees: this may call for experimentation but if you're right i think the code is wrong.

ContextDir is another field (along with SourceSecret) that should have been attached to the Git structure instead of being a standalone field.  What it should mean is that effectively we just clone that subdirectory of your git repo and put whatever that is in your build's contextdir.

bparees: it does look like you're right, for both s2i and docker we:

1) pull everything down to a base dir
2) move into the contextdir (or rename the contextdir to be the basedir in the s2i case)
3) build

that is..unfortunate.

csrwng: I guess we should fix it :)

bparees: i'm not sure it can easily be fixed at this point w/o breaking backwards compatibility.  the binary input and dockerfile inputs also make this behavioral assumption.

the dockerfile one we can probably deal with, not so sure about the binary inputs.

csrwng: I don't think we need to change the behavior for binary input, given that it's like a git repo. I think we still want to potentially use a context directory within it. However dockerfile and image source can/should be fixed.

bparees: i don't like the idea that we have one field that applies to a random subset of the input sources.

i think i'd rather keep the current behavior (it applies to all of them in the same way) and then in the future introduce a contextDir field on each input source that does the right thing, while deprecating the existing one.

csrwng:
LGTM

bparees:
@openshift/team-documentation ptal, ready for merge.

adellape:
@bparees Sorry, just merged https://github.com/openshift/openshift-docs/pull/1452 which makes this need a rebase.

bparees:
@adellape np, rebased.

adellape:
[rev_history]
|link:../dev_guide/builds.html[Builds]
|Added an link:../dev_guide/builds.html#image-source[Image Source] section.
%

PR #1448, opened 2016-01-19
Run post commit and pre-push command

PI-Victor:
trello card ref: https://trello.com/c/I1v1YeOf/844-8-implement-post-build-command
Signed-off-by: pi-victor ipalade@redhat.com

[Review on Reviewable](https://reviewable.io/reviews/openshift/openshift-docs/1448)

### DIFF
--- a/dev_guide/builds.adoc
+++ b/dev_guide/builds.adoc
@@ -93,6 +93,8 @@ spec:
     to:
       kind: "ImageStreamTag"
       name: "origin-ruby-sample:latest"
+  postCommit: <6>
+      script: "bundle exec rake test"
 ----
 
 <1> This specification will create a new `*BuildConfig*` named
@@ -100,16 +102,18 @@ spec:
 <2> You can specify a list of link:#build-triggers[triggers], which cause a new
 build to be created.
 <3> The `*source*` section defines the source of the build. The source type
-determines the primary source of input, and can be either `*Git*`, to point to a
-code repository location, `*Dockerfile*`, to build from an inline Dockerfile, or
-`*Binary*`, to accept binary payloads. It is possible to have multiple sources
-at once, refer to the documentation for each source type for details.
+determines the primary source of input, and can be either `*Git*`, to point to
+a code repository location, `*Dockerfile*`, to build from an inline Dockerfile,
+or `*Binary*`, to accept binary payloads. It is possible to have multiple
+sources at once, refer to the documentation for each source type for details.
 <4> The `*strategy*` section describes the build strategy used to execute the
 build. You can specify `*Source*`, `*Docker*` and `*Custom*` strategies here.
 This above example uses the `*ruby-20-centos7*` Docker image that
 Source-To-Image will use for the application build.
 <5> After the Docker image is successfully built, it will be pushed into the
 repository described in the `*output*` section.
+<6> The `*postCommit*` section defines an optional link:#build-hooks[build
+hook].
 ====
 
 [[source-to-image-strategy-options]]
@@ -1425,6 +1429,110 @@ Configuration change triggers currently only work when creating a new
 `*BuildConfig*`. In a future release, configuration change triggers will also be
 able to launch a build whenever a `*BuildConfig*` is updated.
 ====
+
+[[build-hooks]]
+
+== Build Hooks
+
+Build hooks allow behavior to be injected into the build process.
+
+There is one hook type, post commit, that can be used to execute commands
+inside a temporary container running the build output image. The hook is
+executed immediately after the last layer of the image has been committed and
+before the image is pushed to a registry.
+
+The current working directory is set to the image's `*WORKDIR*`, the default
+working directory of the Docker image. For most images, this is where
+the source code is located.
+
+The hook will fail if the script or command returns a non-zero exit code or if
+there's any other error related to starting the temporary container. When the
+hook fails it will mark the build as failed and the image will not be pushed to
+a registry. The exact failure reason can be inspected by looking at the build
+logs.
+
+Build hooks can be used to run unit tests to verify the image before the build
+is marked complete and the image is made available in a registry. If all tests
+pass and the test runner returns exit code 0, the build will be marked
+successful. Otherwise, in case of any test failure, the build will be marked as
+failed. In all cases, the build log will contain the output of the test runner,
+which can be used to identify failed tests.
+
+The post commit hook is not limited to running tests but can be used for other
+commands as well. Changes made by the hook will not be persisted, since it runs
+in a temporary container. In other words, the hook execution cannot affect the
+final image. This behavior allows for, among other uses, the installation and
+usage of test dependencies that will be automatically discarded and not present
+in the final image.
+
+There are different ways to configure the post build hook. All forms in the
+examples below are equivalent and will execute `*bundle exec rake test
+--verbose*`:
+
+1. Shell script:
++
+[source,yaml]
+----
+postCommit:
+  script: "bundle exec rake test --verbose"
+----
+In this form, `*script*` is a shell script to be run with `*/bin/sh -ic*`. Use
+this when a shell script is appropriate to execute the build hook, for example
+for running unit tests as above. If you need control over the image entrypoint,
+or if the image does not have `*/bin/sh*`, use `*command*` and/or `*args*`.
++
+NOTE: The additional `*-i*` flag was introduced to improve the experience
+working with CentOS and RHEL images, and may be removed in a future release.
++
+2. Command as the image entrypoint:
++
+[source,yaml]
+----
+postCommit:
+  command: ["/bin/bash", "-c", "bundle exec rake test --verbose"]
+----
++
+In this form, `*command*` is the command to run. It overrides the image
+entrypoint in the exec form, as documented in the
+link:https://docs.docker.com/engine/reference/builder/#entrypoint[Dockerfile
+reference]. This might be needed if the image does not have `*/bin/sh*` or if
+you do not want to use a shell. In all other cases, using `*script*` might be
+more convenient.
++
+3. Pass arguments to the default entrypoint:
++
+[source,yaml]
+----
+postCommit:
+  args: ["bundle", "exec", "rake", "test", "--verbose"]
+----
+In this form, `*args*` is a list of arguments that are provided to the default
+entrypoint of the image. The image entrypoint must be able to handle arguments.
++
+4. Shell script with arguments:
++
+[source,yaml]
+----
+postCommit:
+  script: "bundle exec rake test $1"
+  args: ["--verbose"]
+----
++
+This form is useful if you need to pass arguments that would otherwise be hard
+to quote properly in the shell script. In the `*script*`, `$0` will be "/bin/sh"
+and `$1`, `$2`, etc, are the positional arguments from `*args*`.
++
+5. Command with arguments:
++
+[source,yaml]
+----
+postCommit:
+  command: ["bundle", "exec", "rake", "test"]
+  args: ["--verbose"]
+----
+This form is equivalent to appending the arguments to `*command*`.
+
+NOTE: It is invalid to provide both `*script*` and `*command*` simultaneously.
 
 [[using-docker-credentials-for-pushing-and-pulling-images]]
 == Using Docker Credentials for Pushing and Pulling Images


PI-Victor:
@rhcarvalho ping.

dev_guide/builds.adoc
```diff
@@ -1109,6 +1109,27 @@ The following example shows the part of a `*BuildConfig*` specifying
 ----
 ====
 
+[[post-build-actions]]
```
rhcarvalho: It is not "post build", we've been calling it "post commit".

PI-Victor:
@rhcarvalho can you review this? i think it's in a more finished state now. also you were right about replacing json with yaml so i also had to convert the examples. also eliminated the changes to api_v1.adoc.
make some comments if you find anything that needs rewording i might have forgotten something from our conversation today.

dev_guide/builds.adoc
```diff
@@ -93,6 +93,10 @@ spec:
     to:
       kind: "ImageStreamTag"
       name: "origin-ruby-sample:latest"
+  postCommit: <6>
+      command: "sh -c",
+      args: "rake test"
```
rhcarvalho: Note that command and args are string arrays.

This example is invalid according to our API validation, as one cannot have command and script together.
I'd keep this simple and just use "script". We can describe the other possibilities further down.

dev_guide/builds.adoc
```diff
@@ -110,6 +114,8 @@ This above example uses the `*ruby-20-centos7*` Docker image that
 Source-To-Image will use for the application build.
 <5> After the Docker image is successfully built, it will be pushed into the
 repository described in the `*output*` section.
+<6> This post-commit hook will run a command in a container after the image is committed and before it is pushed to the registry. `command` and `script` can't be specified together.
```
rhcarvalho:
```asciidoc
<6> You can specify in `*postCommit*` a command to be executed in a temporary
container running the build output image, immediately after the last layer of
the image is committed and before the image is pushed to a registry.
```

rhcarvalho: Well, and include a link to the details.

dev_guide/builds.adoc
```diff
@@ -1184,6 +1190,80 @@ spec:
 ----
 ====
 
+[[build-hooks]]
```
rhcarvalho: How about we put this section after triggers? Deployments talk about triggers first, hooks later.
In general the sections are not ordered mirroring deployments, but I think in this case it makes more sense to have hooks later.

dev_guide/builds.adoc
```diff
@@ -1184,6 +1190,80 @@ spec:
...
+
+== Build Hooks
+Build hooks can be used as a part of the build flow to run tests inside a temporary container and validate your image before it is being pushed to the registry. The command is executed with the current working directory set to the image's `WORKDIR`.
```
rhcarvalho: - [ ] "before it is pushed to a registry."
- [ ] I'm not 100% sure, but I think it might be \`\*WORKDIR\*\`, here and everywhere else (AsciiDoc syntax).
- [ ] You can also add that "WORKDIR" is normally where your code resides (with your own words), as that will make it less puzzling what is "WORKDIR" and what does that actually mean.
- [ ] I suggest breaking these long lines into multiple lines, as that's the pattern we follow in `.adoc` files. See the paragraphs above and below the new content. With Atom, ctrl-alt-q on a line will reformat the current paragraph.

rhcarvalho: Although we're introducing this as a means to run tests, it is worth mentioning that this is not to the exclusion of other possibilities.

- [ ] Briefly mention that this feature can be used for other purposes as well. Keep in mind that:
  1. The command runs in the same image that would be pushed to a registry
  2. If the command makes any changes, those won't be persisted. This can be "good" or "bad", depending on what you're trying to do, and knowing about this helps a user determine if this is the right feature she's looking for. You can install test dependencies prior to running tests, and those won't affect the final image. You cannot use this feature to "sign" a binary, or anything else one would expect to persist. You cannot use this to remove files produced during the build, etc.

dev_guide/builds.adoc
```diff
@@ -1184,6 +1190,80 @@ spec:
...
+The hook will fail if the script or command returns a non-zero exit code or if there's any other error related to starting the temporary container. When the hook fails it will mark the build as failed.
+====
+----
+spec:
+  postCommit: <1>
+    command:
```
rhcarvalho: - [ ] According to the description below, the `<1>` marker should be in this line, and not in the line above.
- [ ] JSON is valid YAML, so I think here it would be more clear to write `command: ["/bin/sh", "-c"]` instead of using multiple lines.

dev_guide/builds.adoc
```diff
@@ -1184,6 +1190,80 @@ spec:
...
+      -"sh"
+      -"-c"
+    args: "rake test" <2>
```
rhcarvalho: - [ ] args is an array of strings

dev_guide/builds.adoc
```diff
@@ -1184,6 +1190,80 @@ spec:
...
+----
+<1> `command` is the command to run. This might be needed if the image doesn't have "/bin/sh" or if you do not want to use a shell. In all other cases, using `script` might be more convenient.
+
+<2> `args` is a list of arguments that are provided to either `command`,
+`script` or the default entrypoint of the image. The arguments are
+placed immediately after the command to be run.
+
+<3> `script` is a shell script to be run with `/bin/sh -c`. Use `script` when a shell script is appropriate to execute the post build hook, for example for running unit tests with "rake test". If you need control over the image entrypoint, or if the image does not have "/bin/sh", use `command` and/or `args`.
+
+*NOTE: It is invalid to provide both `Script` and `Command` simultaneously. If none of the fields are specified, the hook is not executed.
```
rhcarvalho: There is some special syntax for note blocks. For a one-liner note, the case here, it is:

```asciidoc
NOTE: this is a note.
```

Note that there should be no "*" in the beginning of the line.
And if the note is lengthier, there's this format as well:

```asciidoc
[NOTE]
====
Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec a diam lectus.
Sed sit amet ipsum mauris. Maecenas congue ligula ac quam viverra nec
consectetur ante hendrerit. Donec et mollis dolor. Praesent et diam eget libero
egestas mattis sit amet vitae augue.
====
```

dev_guide/builds.adoc
```diff
@@ -1184,6 +1190,80 @@ spec:
 ----
 ====
 
+[[build-hooks]]
+
+== Build Hooks
+Build hooks can be used as a part of the build flow to run tests inside a temporary container and validate your image before it is being pushed to the registry. The command is executed with the current working directory set to the image's `WORKDIR`.
+The hook will fail if the script or command returns a non-zero exit code or if there's any other error related to starting the temporary container. When the hook fails it will mark the build as failed.
+====
```
rhcarvalho: This section between `====` is good to describe what the fields in `postCommit` are, but it does so with an invalid example. I'd suggest we rework this content to be with the numbered examples below (1, 2, ..., 5).

That is:

```
... When the hook fails it will mark the build as failed.

There are five different ways to configure the post build hook....

(include details within the examples below)
1. ...
2. ...
...
```

dev_guide/builds.adoc
```diff
@@ -1184,6 +1190,80 @@ spec:
...
+<3> `script` is a shell script to be run with `/bin/sh -c`. Use `script` when a shell script is appropriate to execute the post build hook, for example for running unit tests with "rake test". If you need control over the image entrypoint, or if the image does not have "/bin/sh", use `command` and/or `args`.
+
+*NOTE: It is invalid to provide both `Script` and `Command` simultaneously. If none of the fields are specified, the hook is not executed.
+====
+
+There are five different ways to configure the post build hook. As an example, all forms below are equivalent and will execute `rake test --verbose`.
+
+1. Shell script:
+
+  postCommit:
```
rhcarvalho: - [ ] I think all the examples here need to be surrounded with `----`.

rhcarvalho:
@PI-Victor we're in good shape, just need to review the syntax and improve on the parts where we're including invalid examples (command and script together). We don't want copy-pastes from the docs to result in validation error :smile:

dev_guide/builds.adoc
PI-Victor: this didn't work. i know what you want but aesthetically it looks bad. the ==== block is more for defining attributes rather than examples, otherwise it stretches till the bottom of the postcommit definition and looks like nothing similar in these docs.

dev_guide/builds.adoc
```diff
@@ -1426,6 +1432,94 @@ Configuration change triggers currently only work when creating a new
 able to launch a build whenever a `*BuildConfig*` is updated.
 ====
 
+[[build-hooks]]
+
+== Build Hooks
+
+Build hooks can be used as a part of the build flow to run tests inside a
+temporary container and validate your image before it is being pushed to a
```
bparees: s/being//

dev_guide/builds.adoc
```diff
@@ -1426,6 +1432,94 @@ Configuration change triggers currently only work when creating a new
...
+registry. The command is executed with the current working directory set to the
+image's `*WORKDIR*` which is your current working directory where your code is
+located. The hook will fail if the script or command returns a non-zero exit
+code or if there's any other error related to starting the temporary container.
+When the hook fails it will mark the build as failed.
```
bparees: it will mark the build as failed and the image will not be pushed to a registry.

dev_guide/builds.adoc
```diff
@@ -1426,6 +1432,94 @@ Configuration change triggers currently only work when creating a new
...
+
+[source,yaml]
+spec:
+  postCommit:
+    command: ["/bin/sh", "-c"] <2>
+    args: ["rake, ,"test"] <3>
+
+<2> `command` is the command to run. This might be needed if the image doesn't
+have "/bin/sh" or if you do not want to use a shell. In all other cases, using
+`script` might be more convenient.
```
bparees: additional final sentence.  "This value is used as the entrypoint for the container."

dev_guide/builds.adoc
```diff
@@ -1426,6 +1432,94 @@ Configuration change triggers currently only work when creating a new
...
+<3> `args` is a list of arguments that are provided to either `command`,
+`script` or the default entrypoint of the image. The arguments are placed
+immediately after the command to be run.
+
+
+  NOTE: It is invalid to provide both `Script` and `Command` simultaneously. If
+  none of the fields are specified, the hook is not executed.
+====
+
+There are five different ways to configure the post build hook. As an example,
```
bparees: s/five/several/ so we don't have to keep this number updated if things change.

dev_guide/builds.adoc
```diff
@@ -1426,6 +1432,94 @@ Configuration change triggers currently only work when creating a new
...
+
+1. Shell script:
+[source,yaml]
+postCommit:
+  script: "rake test --verbose"
+
+2. Command as the image entrypoint:
+[source,yaml]
+postCommit:
+  command: ["rake", "test", "verbose"]
```
bparees: --verbose?

dev_guide/builds.adoc
```diff
@@ -1426,6 +1432,94 @@ Configuration change triggers currently only work when creating a new
...
+ 5. Command with arguments:
+[source,yaml]
+postCommit:
+  command: ["rake", "test"]
+  args: ["--verbose"]
+
+ * This form is equivalent to appending the arguments to `command`.
+
+The post commit hook is not limited to running tests but can be used for other
+commands as well, what you have to bare in mind is that changes made by the
```
bparees: "comands as well.  It is important to be aware that changes made by the post-build command will not be persisted since it runs in a temporary container.  As such, no actions performed by the post-build command will affect the final image.  This behavior allows for, among other uses, the installation and usage of test dependencies without affecting the final image."

PI-Victor:
@bparees corrected @rhcarvalho hey, updated this, what do you think?

dev_guide/builds.adoc
```diff
@@ -110,6 +112,11 @@ This above example uses the `*ruby-20-centos7*` Docker image that
 Source-To-Image will use for the application build.
 <5> After the Docker image is successfully built, it will be pushed into the
 repository described in the `*output*` section.
+<6> In `*postCommit*` you can specify a link:#build-hooks[command] to be
+executed in a temporary container running the build output image, immediately
+after the last layer of the image is committed and before the image is pushed to
```
rhcarvalho: "is committed" -> "has been committed"?

dev_guide/builds.adoc
```diff
@@ -1426,6 +1432,96 @@ Configuration change triggers currently only work when creating a new
...
 ====
 
+[[build-hooks]]
+
+== Build Hooks
+
+Build hooks can be used as a part of the build flow to run tests inside a
+temporary container and validate your image before it is pushed to a
+registry. The command is executed with the current working directory set to the
+image's `*WORKDIR*` which is your current working directory where your code is
```
rhcarvalho: "which is your current working directory where your code is located"

I think we cannot affirm that for 100% of cases. I'd probably reword this to reflect that it is the common case, but not a guarantee.

dev_guide/builds.adoc
```diff
@@ -1426,6 +1432,96 @@ Configuration change triggers currently only work when creating a new
...
+located. The hook will fail if the script or command returns a non-zero exit
+code or if there's any other error related to starting the temporary container.
+When the hook fails it will mark the build as failed and the image will not be pushed to a registry.
+====
+[source,yaml]
+spec:
+  postCommit:
+    script: "rake test" <1>
+
+<1> `script` is a shell script to be run with `/bin/sh -c`. Use this when a
```
rhcarvalho: Not 100% sure, but thinking ``` `script` ``` is markdown syntax, and we normally use ``` `*script*` ``` in our adocs...

dev_guide/builds.adoc
```diff
@@ -1426,6 +1432,96 @@ Configuration change triggers currently only work when creating a new
 able to launch a build whenever a `*BuildConfig*` is updated.
 ====
 
+[[build-hooks]]
+
+== Build Hooks
+
+Build hooks can be used as a part of the build flow to run tests inside a
```
rhcarvalho: We should probably start with a more generic purpose statement, and then nail down to "use this for testing the image".

dev_guide/builds.adoc
```diff
@@ -1426,6 +1432,96 @@ Configuration change triggers currently only work when creating a new
...
+
+ 5. Command with arguments:
+[source,yaml]
+postCommit:
+  command: ["rake", "test"]
+  args: ["--verbose"]
+
+ * This form is equivalent to appending the arguments to `command`.
+
+The post commit hook is not limited to running tests but can be used for other
```
rhcarvalho: I think this block/paragraph could go well in the beginning of the section, so we have a few (1-3) paragraphs explaining the feature, it's purpose, constraints, etc, and then we dive into the "running unit tests" examples with the various syntaxes and their particularities.

dev_guide/builds.adoc
bparees: I *think* either is ok, but we can let the tech writers decide :)

dev_guide/builds.adoc
bparees: +1

dev_guide/builds.adoc
bparees: +1

dev_guide/builds.adoc
PI-Victor: i'll change it to has been committed.

dev_guide/builds.adoc
PI-Victor: actually i think i did start like that and somehow ended up with tests. will redo though.

dev_guide/builds.adoc
adellape: +1

adellape:
@PI-Victor LGTM in general. LMK when you've addressed the last couple comments and I can merge if you're ready.

rhcarvalho:
@PI-Victor let's get this up to date and merged soon.

We need to update the patch and add a note about the `-i` flag.

We will also need a note along these lines:

> NOTE: The build hook introduced in v1.1.3 has had its `script` flavor changed to
be executed with `/bin/sh -ic` instead of `/bin/sh -c`.
> The additional `-i` flag was introduced to improve the experience working with
CentOS and RHEL images at the moment, and may be removed in a future release.

dev_guide/builds.adoc
```diff
@@ -1426,6 +1430,111 @@ Configuration change triggers currently only work when creating a new
...
+----
+postCommit:
+  script: "bundle exec rake test --verbose"
+----
+In this form, `*script*` is a shell script to be run with `*/bin/sh -ic*`. Use
+this when a shell script is appropriate to execute the build hook, for example
+for running unit tests as above. If you need control over the image entrypoint,
+or if the image does not have `*/bin/sh*`, use `*command*` and/or `*args*`.
++
+NOTE: The build hook introduced in v1.1.3 used `*/bin/sh -c*`. The additional
```
rhcarvalho: @adellape @bparees how can we best convey this note? Version numbers will differ depending own the product branch and we can use the conditional macros for that. Is 1.1.3 becoming some OSE release?

rhcarvalho:
I sat with @PI-Victor and we did a major review of this, updating the points that were needed.

@bparees @adellape could you please take another, perhaps last, look? Thanks!

dev_guide/builds.adoc
bparees: I think we ignore the v1.1.3 bit and just say that for now it uses -ic and we may change that in the future.

1.1.3 is not becoming an OSE release, we already have a 1.1.4 now, and 3.2 won't necessarily match to any exact origin release anyway.

dev_guide/builds.adoc
```diff
@@ -1426,6 +1430,111 @@ Configuration change triggers currently only work when creating a new
...
+
+Build hooks allow behavior to be injected into the build process.
+
+There is one hook type, post commit, that can be used to execute commands
+inside a temporary container running the build output image. The hook is
+executed immediately after the last layer of the image has been committed and
+before the image is pushed to a registry.
+
+The current working directory is set to the image's `*WORKDIR*`, the default
+working directory of the Docker image. Normally, for most images, this is where
```
bparees: s/Normally, //

bparees:
@rhcarvalho done, one nit and one suggestion for your question.

PI-Victor:
@bparees thanks i've updated
@adellape i think we're good, can you take a look

bparees:
lgtm

rhcarvalho:
lgtm

@adellape ptal

rhcarvalho:
@PI-Victor please remove the "WIP" from the title.

@adellape ptal

PI-Victor:
:+1: ping @openshift/team-documentation

rhcarvalho:
@openshift/team-documentation @tnguyen-rh @adellape ping

adellape:
@PI-Victor @rhcarvalho :+1: Thanks.

adellape:
[rev_history]
|link../dev_guide/builds.html[Builds]
|Added a link:../dev_guide/builds.html#build-hooks[Build Hooks] section.
%

PR #1458, opened 2016-01-20
As of openshift-ansible-3.0.35 openshift_master_cluster_method must b…

brenton:
…e set


brenton:
@adellape, This is already set in the 3.1 docs.  However, we'll soon be releasing an update that will require setting this value even for 3.0 installs.

PR #1460, opened 2016-01-21
Bug 1291594, added warning admonition

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1291594

@adellape @nak3 How does this look?

### DIFF
--- a/dev_guide/builds.adoc
+++ b/dev_guide/builds.adoc
@@ -1009,6 +1009,22 @@ Manually cancel a build using the web console, or with the following CLI command
 $ oc cancel-build <build_name>
 ----
 
+[[deleting-a-buildconfig]]
+
+== Deleting a BuildConfig
+Delete a `*BuildConfig*` using the following command:
+
+----
+$ oc delete bc <BuildConfigName>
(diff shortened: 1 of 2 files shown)


nak3:
LGTM. Thank you!

adellape:
@ahardin-rh @nak3 It wasn't immediately clear to me from the Warning whether it was referring to the /etc/resolv.conf file of the master or node host, or from within a container. I'm assuming from the context of th topic and the linked material in the BZ that it's the former, but clarifying either way (e.g., "in the /etc/resolv.conf file on each host") may be helpful.

Also, the `nameserver` and `search` lines are both in the /etc/resolv.conf file, so the sentence could probably be restructured a bit.

nak3:
Ah, I see. I agree @adellape 
Sorry, @ahardin-rh please add `in the /etc/resolv.conf file on each node host` (master is not supposed to be affected.)

ahardin-rh:
@nak3 @adellape thanks for your feedback! I made the updates. Please let me know if further edits are needed.

PR #1463, opened 2016-01-21
Bug 1296333, added step 4 to Configuring Docker Storage

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1296333

@bfallonf @adellape @tpoitras @tnguyen-rh how does this look?

### DIFF
--- a/install_config/install/prerequisites.adoc
+++ b/install_config/install/prerequisites.adoc
@@ -708,10 +708,10 @@ monitoring.
 This will destroy any Docker containers or images currently on the host.
 ====
 +
+. Ensure that Docker is running properly:
 ----
-# systemctl stop docker
-# rm -rf /var/lib/docker/*
-# systemctl restart docker
+# systemctl enable docker
+# systemctl start docker
 ----
 
 [[reconfiguring-docker-storage]]


adellape:
@ahardin-rh The "Reinitialize Docker" step is still in there, which is no longer what's going on, with these edits. That line and the Warning box can be removed.

A nit, but, "Ensure it's running properly" suggests to me that it's already running and I should just check its status (like with "systemctl status docker"). May be more accurate (per the commands being shown) to just plainly say "Enable and start Docker".

I do think the step is still worthwhile to include, in general (even if the installer does enable/start it), to make sure there's no problems w/ it starting up, instead of finding out later during the OpenShift installation.

bfallonf:
@ahardin-rh Yeah I think the sentence might need a tweak, too. Another option is to do something like:

Make sure docker is running:

$ systemctl status docker

and if it is not, start it:

$ systemctl start docker

vikram-redhat:
@ahardin-rh Is this fix only applicable to 3.1?

tnguyen-rh:
@ahardin-rh
I think the original BZ is slightly misguided.  To me, the term "prerequisites" has checklist nature but not necessarily sequential execution nature.  That is, someone installing OpenShift might indeed already have Docker installed (and used for other purposes), so the filesystem storage (i.e., `/var/lib/docker`) might very well be non-empty.

There is another wrinkle: Deletion is necessary only if the storage backend changes.  If it is already thin-pool-on-logical-volume, then that `rm -rf` step is superfluous.

To satisfy the BZ OP probably we could add a blurb describing the difference between "ground-up installation" and "prerequisites verification" as well as conditionalize (using "if") some of the steps along the lines of what @bfallonf suggested above.

adellape:
> That is, someone installing OpenShift might indeed already have Docker installed (and used for other purposes), so the filesystem storage (i.e., /var/lib/docker) might very well be non-empty.

@tnguyen-rh Yeah, that was the original intent of including the step.

adellape:
@ahardin-rh @tnguyen-rh Actually, another wrinkle: Starting with OSE 3.1.1, Atomic Host is a supported OS for OSE hosts. So for those AH systems, Docker actually _will_ already be running. So, just another reason really to make the distinction clear. And actually, this topic will be getting updated for that inclusion of AH anyway, so there will probably be further impact here too.

ahardin-rh:
Closing in favor of https://github.com/openshift/openshift-docs/pull/1475, which made the changes for whether or not Docker is already running.

PR #1464, opened 2016-01-22
image related updates for online beta

gabemontero:
@luciddreamz @abhgupta @sspeiche @danmcp  FYI - my very _initial_ pass at a write up for the documentation for the [online beta image/template curation card](https://trello.com/c/zgWyXOg8/474-8-curating-the-set-of-images-templates)

As I've added a whole new section to Using Images, rather than needing ifdef's, I believe simply limiting the Distros in _build_cfg.yml to online should suffice, but of course correct me as needed.

Otherwise of course content / structure / location are all fair game at this early stage.

PTAL - thx.

@bparees - FYI - this PR starts to delve somewhat (more of a toe tipping really) into documenting (or at least claiming the existence of) the example repos in the official documentation.  DevExp weigh in of course welcomed as well.

### DIFF
--- a/_build_cfg.yml
+++ b/_build_cfg.yml
@@ -449,9 +449,16 @@ Topics:
         File: decision_server
       - Name: JBoss Data Grid
         File: data_grid
+  - Name: OpenShift V3 Online
+    Dir: online_images
+    Distros: openshift-online
+    Topics:
+      - Name: Overview
+        File: index
   - Name: Revision History
     File: revhistory_using_images
     Distros: openshift-enterprise,openshift-dedicated
+
 
 ---
 Name: REST API Reference
--- a/using_images/online_images/index.adoc
+++ b/using_images/online_images/index.adoc
@@ -1,0 +1,39 @@
+= Current restrictions and constraints
+{product-author}
+{product-version}
+:data-uri:
+:icons:
+:experimental:
+
+The OpenShift Online Developer Preview offering scopes the inventory of images it provides out of the box
+with a few contraints in mind.  These constrainst also apply to any images you choose to
+inport into your Project.  The various constraints are enforced via the OpenShift Quota
+system.  See the link:../../dev_guide/compute_resources.html[computing guide] for a description of how this system works.
+
+Now, the specific constraints:
+
+* A memory limit of 2GiB is in place.  The 2 GiB is spread out across the project's pods and containers.
+* Up to 10 pods are allowed.
+* Up to 20 Replication Controllers are allowed.
+* Up to 10 Services are allowed.
+* Up to 16 Secrets are allowed (though some amount of these secrets will be needed by the system's build and deployer service accounts)
+* No templates that map a DockerFile VOLUME to an emptyDir Kubernetes Volume are allowed for now.  The current intent is to lift this restriction and the V3 OpenShift Online Developer Preview offering progresses out of the beta phase.
+* In fact, any Dockerfile VOLUME must be mounted with a Persistent Volume Claim at this time with the V3 OpenShift Online Developer Preview offering.
+* The Project associated with a user can allocate up to 2 Persistent Volume Claims.
+* No images that run as ROOT are allowed.
+* Only the Source-To-Image build strategy is allowed for any BuildConfigs imported into your Project.
+
+If you want to know your Project's current resource usage, you can log into the console and view them from the settings tab of your Project view, or use the CLI command  `oc describe quota [your project's quota object's name]`.
+
+As part of providing a set of templates out of the box, the Online Developer Preview offering has taken various, publicly accessible templates and added a memory limit template parameter with
+a default setting for the deployments.  These default settings were based on tuning exercises of some common developer flows, with the 2 GiB memory limit in mind.
+
+You can change the defaults as you see fit, based on the needs of the specific scenario you want to try.  You'll just have to keep in mind the 2 GiB overall memory limit when adjusting the settings
+of your various deployments.
+
+[NOTE]
+Read about memory limits  link:../../dev_guide/compute_resources.html#memory-limits[here] and template parameters link:../../architecture/core_concepts/templates.html#parameters[here] if you require more details on those concepts.
+
+
+
+
--- a/getting_started/developers/developers_cli.adoc
+++ b/getting_started/developers/developers_cli.adoc
@@ -27,13 +27,73 @@ start by reading about
 ifdef::openshift-enterprise[]
 link:../../release_notes/index.html[what's new].
 endif::[]
-ifdef::openshift-origin[]
+ifdef::openshift-origin,openshift-online[]
 link:../../whats_new/index.html[what's new].
 endif::[]
-This version
 of OpenShift is significantly different from version 2.
 
-The following sections guide you through creating a project that contains a
+That said, similar to version 2, OpenShift provides out of the box for developers
+a set of languages and link:../../using_images/db_images/index.html[databases] with corresponding implementations and tutorials that allow you to kickstart your application
+development.  The support around languages center around the link:../../dev_guide/templates.html#using-the-instantapp-templates[InstantApp Templates].
+InstantApp Templates in turn leverage link:../../using_images/s2i_images/index.html[Builder Images].
+
+|===
+|Language|Implementations and Tutorials
+
+.^|Ruby
+|https://github.com/openshift/rails-ex[Rails]
+
+.^|Python
+|https://github.com/openshift/django-ex[Django]
+
+.^|Node.js
+|https://github.com/openshift/nodejs-ex[Node.js]
+
+.^|PHP
+|https://github.com/openshift/cakephp-ex[CakePHP]
+
+.^|Perl
+|https://github.com/openshift/dancer-ex[Dancer]
+
+.^|Java
+|link:../../using_images/s2i_images/java.html[Maven]
+
+|===
+
+Other OpenShift Docker Images
+
+* https://github.com/openshift/mysql[MySQL]
+
+* https://github.com/openshift/mongodb[MongoDB]
+
+* https://github.com/openshift/postgresql[PostgreSQL]
+
+* https://github.com/openshift/jenkins[Jenkins]
+
+
+ifdef::openshift-enterprise,openshift-dedicated,openshift-online[]
+=== More on Java
+JBoss Middleware has put together a broad range of https://github.com/jboss-openshift/application-templates[OpenShift templates] as well as link:../../using_images/xpaas_images/index.html[images] as part of their xPaaS services.
+
+The technologies available with the xPaaS services in particular include:
+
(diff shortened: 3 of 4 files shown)


bparees:
what about the pure DB deployment templates?

bparees:
(https://github.com/openshift/origin/tree/master/examples/db-templates)

gabemontero:
Yep I should have mentioned I was going to add those once I did a sanity
check on there resource consumption out of the box.

On the subject of upcoming additions, I plan to include the wordpress
example you have (following the resource eval) and the see if a useful
Laravel example could be constructed.  They are the 2 most popular V2
quickstarts not already covered by the V3 content listed here.

Thanks

bparees:
i'd be wary of the wordpress one.  when deployed, it builds from the
wordpress master repo, so it could get broken at any time.

(it's also a little clunky in terms of how the DB gets configured)


On Fri, Jan 22, 2016 at 11:06 AM, Gabe Montero <notifications@github.com>
wrote:

> Yep I should have mentioned I was going to add those once I did a sanity
> check on there resource consumption out of the box.
>
> On the subject of upcoming additions, I plan to include the wordpress
> example you have (following the resource eval) and the see if a useful
> Laravel example could be constructed. They are the 2 most popular V2
> quickstarts not already covered by the V3 content listed here.
>
> Thanks
>
> On Friday, January 22, 2016, Ben Parees <notifications@github.com> wrote:
>
> > (https://github.com/openshift/origin/tree/master/examples/db-templates)
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <
> https://github.com/openshift/openshift-docs/pull/1464#issuecomment-173961468
> >
> > .
>
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/openshift/openshift-docs/pull/1464#issuecomment-173962898>
> .
>



-- 
Ben Parees | OpenShift

gabemontero:
Ah thanks for the heads up

adellape:
@gabemontero Do you envision more topics being added to this new subcategory? It looks from this like all of the standard OSE/xPaaS images are being included for Online, so I think this could work just as `ifdef`'d content in the normal Using Images -> Overview topic, so that it appears as the intro to the discussion of the images, instead of a separate topic down at the bottom after all the individual image docs.

gabemontero:
@adellape hey - yeah, the situation is still somewhat fluid wrt which images we'll include, and how much elaboration via additional topics we'll need so that online users can ramp up quickly.  I'm curious if we are going to end up needing tweeks to the web console for example to better guide users through instantiating templates in the online env, which we might document here.  Also, while we've certainly been told to incorporate online specific doc with the existing doc, I'm still unclear on if/how the various online pieces might be exposed in some aggregated way.  There is in fact a [general online documentation card](https://trello.com/c/ekmsWdIf/476-8-documentation) out there.  Is it going to end up being a placeholder while individual updates occur in the other cards, or is something "generic" coming out of it.

So with that preamble :-), this was in part why I went with the separate section approach, at least for now.  I thought it better to isolate it more explicitly.  Certainly though let's keep an eye on this.  If/When it becomes clear that the content should merge elsewhere, we'll certainly do that.

thanks

adellape:
@gabemontero Got it, thanks!

gabemontero:
FYI - I've pushed an update that points to the DB-only templates.

gabemontero:
@bparees - re: Wordpress, the example readme also talks about how the Wordpress Dockerhub image binds Apache to port 80 in the container, which requires root access.

I subsequently see that the OpenShift service is exposed on port 5055, but unless we create our own extension image to not bind on port 80, the dockerhub images is a no-go.  Or am I missing something.

In any event, the net is there is more ammunition to add to your warning regarding wordpress.  I'm pretty much decided into not pulling that in.

I am considering the golang-ex (golang was a pretty popular v2 quickstart as well) that @mfojtik put together last year.  Either of you have strong reasons for not including that in the online beta?

bparees:
@gabemontero there are two ways to run wordpress.  one is the readme described here, which you are referring to and just uses an existing dockerhub image:
https://github.com/openshift/origin/tree/master/examples/wordpress

The other is the way I created which uses a template to define a buildconfig which builds a custom wordpress image using our php s2i build image, which means it does not listen on 80, because we configure php/apache correctly in the s2i builder image:
https://github.com/openshift/origin/tree/master/examples/wordpress/template

The concern I expressed earlier was about using that template because of the fact that it defines a buildconfig that just points to (and attempts to build) the wordpress master repo, so there's no guarantee it doesn't get broken.

We could of course fork the wordpress repo and update the template to build the fork instead, so we'd have control over it changing.

Anyway run through the template readme steps, you'll have a better sense of what does/doesn't do compared to the pre-canned image example.

gabemontero:
@bparees right, there are 2 ways, and neither way is too attractive IMO, minus possibly the fork of the wordpress repo as a form of version control.  Not sure what exactly you meant by the "DB clunkiness" wrt config after looking at the template (elaborate if you could when time permits), but technically speaking, its use of emptyDir at the moment still precludes it from the beta, though I'm sure you are aware of the hot debate regarding storage and the beta.

So for the moment wordpress is out (though several things are fluid).

But talking it out was helpful to me - thx as always.

gabemontero:
Oh, and I mixed two topics in an earlier post ... when time permits @bparees @mfojtik , opinions on exposing golang-ex for the online beta would be welcome.

bparees:
@gabemontero the db clunkiness is that you have to copy the generated DB password that's given to you when the template is instantiated, and then use it to configure wordpress once it starts up.  you also have to manually configure the db hostname.  If you walk through the readme you'll see what i mean.  it's not awful but it's not seamless.

as for golang-ex, i don't see any particular problem w/ offering it, but it's definitely not one of our better tested/supported offerings.

gabemontero:
@bparees yep, gotcha now (re: clunkiness), and thx (re: golang-ex) ... i'm thinking let's give golang-ex a shot (pending any worries from @mfojtik )

using_images/online_images/index.adoc
```diff
@@ -0,0 +1,101 @@
...
+
+The inventory of images are being pulled from a variety of sources, including
+images documented in other sections of link:../index.adoc[Using Images].
+
+The OpenShift Online offering scopes the inventory of images it provides out of the box
+with a few contraints in mind.  Specifically:
+
+* The Project associated with a user can only allocate any number of pods, but
+* A memory limit of 2GiB is place on any project a user creates.  The 2 GiB is spread out across the project's pods, and enforced via the OpenShift Quota system; see the link:../../dev_guide/compute_resources.adoc[computing guide] for a description of how this system works.
+* No images with an emptyDir Kubernets Volume are allowed.
```
dgoodwin: Couple suggestions for this very tricky section:

- We should state emptyDir volumes are not permitted "at this time". (hopefully they will be soon)
- emptyDir volumes would be in a template rather than an image. (specifically a PodSpec somewhere within a template)
- Specific to images though we should also note that if you are using an image which includes a VOLUME in it's Dockerfile, that volume must be mounted with a persistent volume claim, otherwise the image will not run. (this is also an Online specific limitation)
- We should probably document here that the user will have a quota of 2 persistent volume claims available. (last I heard, the number may change)

gabemontero:
I've pushed an update based on the comments from @dgoodwin here, along with some ongoing work and quick chat with @sspeiche

gabemontero:
I've made another push based on clarifications from @dgoodwin on which images are acceptable under the current persistence restrictions.

gabemontero:
I've made (potentially placeholder) update to the xPaas section based on some walkthroughs of the xPaas templates and quickstarts.  Among other things, I've found some issues in the EAP/xPaas online documentation, for which xPaas test has opened a project JIRA issue to address.  Once they have corrected things, some of what I've added could possibly be  removed.  We'll see as things progress.

gabemontero:
Otherwise, after some additional discussion with @sspeiche , I think this PR is basically close to have the WIP prefix removed and have it get the final considerations for merging.

The last thing(s) at least from my perspective holding that up are essentially:
1) The link refs to the actual templates - what do we ultimately point to (currently we point to the versions under openshift-ansible) ... as things stand, if we still want to have default memory limits for the templates, either those openshift-ansible templates need to be updated quickly when the origin ones get merged, or we point to more recent copies, or we don't point at all and just list / ref the ones that show up in the console

2) how this page gets referenced by the Getting started, etc. to the [online branch](https://github.com/openshift/openshift-docs/tree/online) as noted in the [documentation trello card](https://trello.com/c/ekmsWdIf/476-8-documentation)

gabemontero:
OK, I've removed the WIP tag for this PR.  I've made updates, including connecting back to the getting started section, based on meeting between @abhgupta , @sspeiche , and myself.  I will note in the getting started that I've bracketed the cli and eclipse entries with ifdef online for now, but do think they should be generically accessible long term.  I just wanted to afford us some wiggle room for pushing the content for online review and percolating on it as part of the online effort before exposing generically.

@abhgupta asked me to set up a meeting with the two of us and @sspeiche , @adellape , @spurtell and @luciddreamz to go over the flow / content / structure, and hopefully get this to the point were we push this content through.

I'll set up that meeting for early next week.

thanks

_build_cfg.yml
```diff
@@ -449,10 +449,17 @@ Topics:
         File: decision_server
       - Name: JBoss Data Grid
         File: data_grid
+  - Name: OpenShift V3 Online
+    Dir: online_images
+    Distros: openshift-online
+    Topics:
+      - Name: Overview
+        File: index
```
adellape: The developers_cli topic also needs to be added earlier up in the "Getting Started" section in this file.

_build_cfg.yml
```diff
@@ -449,10 +449,17 @@ Topics:
         File: decision_server
       - Name: JBoss Data Grid
         File: data_grid
+  - Name: OpenShift V3 Online
```
gabemontero: Developer Preview Considerations

using_images/online_images/index.adoc
```diff
@@ -0,0 +1,39 @@
+= Current restrictions and constraints
+{product-author}
+{product-version}
+:data-uri:
+:icons:
+:experimental:
+
+The OpenShift Online Developer Preview offering scopes the inventory of images it provides out of the box
+with a few contraints in mind.  These constrainst also apply to any images you choose to
```
spurtell: typos - constraints (x2)

gabemontero:
The meeting I mentioned 4 days ago occurred earlier today.  The net is that while certainly additional massaging of this PR's content is needed, the overall outline was agreed to.

What also was agreed to (much to my delight :-) ) is that @adellape will suck in this PR into his own, apply some of the wordsmith'ing and clean up from the comments today, and merge this in with his own soon to be revealed PR.  Thanks @adellape !

adellape:
@gabemontero np. See https://github.com/openshift/openshift-docs/pull/1650 which will autoclose this when merged.

PR #1465, opened 2016-01-22
Bug 129939, destinationCaCertificate noted as required

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1299393

@jkaurredhat @adellape please let me know if further edits are needed. Thanks!

### DIFF
--- a/architecture/core_concepts/routes.adoc
+++ b/architecture/core_concepts/routes.adoc
@@ -599,8 +599,9 @@ spec:
       -----END CERTIFICATE-----
 ----
 
-<1> The `*termination*` field is set to `reencrypt`. Other fields are as in edge termination.
-<2> The `*destinationCaCertificate*` field optionally specifies a CA
-certificate to validate the endpoint certificate, securing the connection
-from the router to the destination.
+<1> The `*termination*` field is set to `reencrypt`. Other fields are as in edge
+termination.
+<2> The `*destinationCaCertificate*` field specifies a CA certificate to
+validate the endpoint certificate, securing the connection from the router to
+the destination. This field is required, but only for re-encryption.
 ====


ahardin-rh:
@jkaurredhat @adellape, any edits here? Thanks!

rest_api/openshift_v1.adoc
```diff
@@ -9278,7 +9278,7 @@ Patch is provided to give a concrete name and type to the Kubernetes PATCH reque
 |certificate|provides certificate contents|false|string|
 |key|provides key file contents|false|string|
 |caCertificate|provides the cert authority certificate contents|false|string|
-|destinationCACertificate|provides the contents of the ca certificate of the final destination.  When using re-encrypt termination this file should be provided in order to have routers use it for health checks on the secure connection|false|string|
+|destinationCACertificate|provides the contents of the ca certificate of the final destination.  When using re-encrypt termination this file should be provided in order to have routers use it for health checks on the secure connection|true|string|
```
adellape: @jkaurredhat The description here "When using re-encrypt termination this file should be provided [...]" makes it sound like it's more accurate to be listed as false overall, and only required when using re-encryption. But that's just my interpretation; Engineering likely would have to weigh in on what's more valid for "Required" in that context.

@ahardin-rh BTW, the content for the API Guide is generated from the origin product itself (see https://github.com/openshift/origin/tree/master/api), which is run periodically (though not as frequent as we probably should) to update this doc. So, any manual changes made directly here would be overwritten next time we generated the doc. If we do change this value here, we would want to follow up and see about getting it changed in the origin product itself so that it persists next time we re-generate.

adellape:
@ahardin-rh One question I'm not sure about inline, otherwise LGTM.

rest_api/openshift_v1.adoc
adellape: @pweil- @ramr Could one of you weigh in on my above question about whether it's more accurate for this value to show true or false? Or point us in the right direction. :bow:

pweil-: You are correct, it is only required for the reencrypt termination type

adellape: Thanks, @pweil-.

@ahardin-rh @jkaurredhat To me then it seems then that it should stay `false` in the API Guide, and just make sure the "Re-encryption Termination" section in the Arch topic makes it clear that `destinationCACertificate` is required, but only for re-encryption (which this PR is on its way towards doing).

ahardin-rh:
@jkaurredhat @adellape , updates are complete. PTAL. Thanks!

ahardin-rh:
**revhistory note:**
Updated  to clarify that `destinationCaCertificate` is required, but only for re-encryption.

PR #1468, opened 2016-01-22
Registry install Advanced Configuration section

aweiteka:
Add Registry install Advanced Configuration section with secure, production config procedure.

### DIFF
--- a/install_config/install/docker_registry.adoc
+++ b/install_config/install/docker_registry.adoc
@@ -558,8 +558,32 @@ Digest: sha256:3662dd821983bc4326bee12caec61367e7fb6f6a3ee547cbaff98f77403cab55
 To enable managing the registry configuration file directly it is recommended that the configuration file be mounted as a secret volume.
 
 . link:#deploy-registry[Deploy the registry].
-. Edit the configuration file locally as needed. See details on
-https://github.com/docker/distribution/blob/master/docs/configuration.md#override-configuration-options[overriding registry options]. The initial file may be found in the link:https://github.com/openshift/origin/blob/master/images/dockerregistry/config.yml[Docker image registry source repository].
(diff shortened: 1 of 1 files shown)


install_config/install/docker_registry.adoc
```diff
@@ -545,6 +545,71 @@ Digest: sha256:3662dd821983bc4326bee12caec61367e7fb6f6a3ee547cbaff98f77403cab55
 ----
 ====
 
+== Advanced Configuration
+
+To enable managing the registry configuration file directly it is recommended that the configuration file be mounted as a secret volume.
+
+. link:#deploy-registry[Deploy the registry].
+. Edit the configuration file locally as needed. See details on
+https://github.com/docker/distribution/blob/master/docs/configuration.md#override-configuration-options[overriding registry options]. The initial file may be found in the link:https://github.com/openshift/origin/blob/master/images/dockerregistry/config.yml[Docker image registry source repository].
```
adellape: Looks like you could link here instead for the registry override doc:

https://docs.docker.com/registry/configuration/

For the `config.yml` file, it looks short enough that maybe we could just inline it here instead of linking to the upstream code repo? Looking at the file's history, it looks like it hasn't had too much action (at least so far), as far as concerns of an in-line version here getting out of date.

adellape:
@aweiteka Sorry for the delay. One comment inline. Do you think this should go into the OSE doc as soon as it's ready? Wondering if there would be any support issues wrt a modified integrated registry.

aweiteka:
> Do you think this should go into the OSE doc as soon as it's ready? Wondering if there would be any support issues wrt a modified integrated registry.

Thanks for the comments. Updated. There are no support issues I'm aware of. I think this is very safe to merge any time.

adellape:
Will add revhistory note in follow-up PR.

PR #1474, opened 2016-01-25
Document admission control plugin configuration

csrwng:
Documents admission control plugins and the configuration exposed in master-config.yaml

### DIFF
--- a/install_config/configuring_admission_control.adoc
+++ b/install_config/configuring_admission_control.adoc
@@ -1,0 +1,161 @@
+= Configuring Admission Control Plugins
+{product-author}
+{product-version}
+:data-uri:
+:icons:
+:experimental:
+:toc: macro
+:toc-title:
+
+toc::[]
+
+== Overview
+
+An admission control plug-in is a piece of code that intercepts requests to the Master API server prior to
+persistence of a resource, but after the request is authenticated and authorized.
+
+Each admission control plug-in is run in sequence before a request is accepted into the cluster. If any plug-in
+in the sequence rejects the request, the entire request is rejected immediately and an error is returned to the end-user.
+
+Admission control plug-ins may modify the incoming object in some cases to apply system configured defaults.
+In addition, admission control plug-ins may modify related resources as part of request processing to do things like increment quota usage.
+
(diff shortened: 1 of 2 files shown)


csrwng:
@deads2k @derekwaynecarr ptal
@dgoodwin fyi

install_config/configuring_admission_control.adoc
```diff
@@ -0,0 +1,161 @@
...
+
+[source, yaml]
+----
+- NamespaceLifecycle
+- OriginPodNodeEnvironment
+- LimitRanger
+- ServiceAccount
+- SecurityContextConstraint
+- ResourceQuota
+- SCCExecRestrictions
```
derekwaynecarr: ResourceQuota should always be last in the sequence.

Please include wording to that affect.

csrwng: It's not in the code. We should open an issue. 
https://github.com/openshift/origin/blob/master/pkg/cmd/server/kubernetes/master_config.go#L38

install_config/configuring_admission_control.adoc
```diff
@@ -0,0 +1,161 @@
...
+link:../dev_guide/quota.html[resource quota] that has been configured for a particular project.
+
+* *SCCExecRestrictions* Prevents users from remotely executing processes on pods that they cannot create according to current
+link:../architecture/additional_concepts/authorization.html#security-context-constraints[security context constraints].
+
+== Providing configuration for Admission Control Plugins
+
+Certain admission control plugins may take an additional configuration file. The configuration for a plugin may be provided either
+as a file location or an embedded configuration object in
+link:master_node_configuration.html#master-configuration-files[master-config.yaml].
```
sosiouxme: Am I right in thinking that kubernetes plugins can't accept embedded configuration? The interface init method only accepts an io.Reader.

I don't think any actually take config yet, but I'm working on one...

csrwng: On the startup of our master we write the embedded config to a temp file and pass that file's stream to the plugin.

csrwng:
Closing in favor of #1476

PR #1477, opened 2016-01-26
Hawkular Metrics and Heapster Access

mwringe:
We need to add some documentation on how to access Hawkular Metrics and Heapster directly with OpenShift Origin.

This PR adds a brief description at the end of the metrics install guide on how to do so. Not sure if we need to add a separate section somewhere instead of adding it to the install section.

### DIFF
--- a/install_config/cluster_metrics.adoc
+++ b/install_config/cluster_metrics.adoc
@@ -490,3 +490,52 @@ specified in `*metricsPublicURL*` and accept that certificate.
 To avoid this issue, use certificates which are configured to be acceptable by
 your browser.
 ====
+
+ifdef::openshift-origin[]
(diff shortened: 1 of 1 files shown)


mwringe:
Any thing keeping this from being merged?

adellape:
@mwringe Nope, sorry. Merging.

Adding to_followup label for general cleanup (e.g., add unique anchors) before picking into the OSE branch (just to keep the file up to date).

**revhistory note:**

None needed because it only applies to Origin.

adellape:
I'm removing the to_followup label here so that it's included in Next Release for ease-of-cherrypick. It's conditionalized to not show up in the non-Origin builds anyway right now.

**revhistory note:** None needed, only applies to Origin.

PR #1481, opened 2016-01-27
Make clear that OpenStack instance names can be used in nodeName.

jsafrane:
Instance ID can be interpreted as instance UUID. `nodeName` should be  human-friendlier instance name.


adellape:
:+1: Thanks.

PR #1482, opened 2016-01-27
OpenStack provider configuration Note that you can use instance_name

chmouel:
You can use instance_name of the openstack vm instead of just instance_id (and this 
is actually less awkward to us)


jsafrane:
Looking at the code, only instance_name is actually allowed, instance UUID won't work. See https://github.com/openshift/openshift-docs/pull/1481

chmouel:
even better i don't think id should ever be exposed :) thanks

PR #1483, opened 2016-01-27
Correct documentation around setting AWS Access keys

jeremyeder:
Docs say you should export access keys. But that doesn't work.  It needs to be set in the service's environment as specified in the systemd unitfile, which are /etc/sysconfig/atomic-openshift-master and /etc/sysconfig/atomic-openshift-node.

http://stackoverflow.com/questions/34731990/openshift-3-origins-persistent-volume-issue


adellape:
:+1: Thanks.

Issue #678, opened 2015-07-02
Explain project-level node selectors more explicitly


liggitt:
Administrators may want to limit which nodes a particular project can run pods on.

A node selector can be set as an annotation on a project/namespace, and pods within that project are restricted to that selector (in addition to any node selector on the pod itself).

A default node selector can be set in the master config file under the "projectConfig" stanza. For example:
```
projectConfig:
  defaultNodeSelector: "type=user-node,zone=east"
  ...
```

It can also be set on a project-by-project basis by an administrator, when creating a project. For example:
```
oadm new-project test --node-selector='type=user-node,zone=east'
```

bfallonf:
I've already started a BZ on this. I'll add this issue to that to track it:

https://bugzilla.redhat.com/show_bug.cgi?id=1253125

PR #1490, opened 2016-01-28
Bug 1253125 : added nodeselector information

bfallonf:
@thoraxe @sosiouxme @liggitt  @adellape Here's a first draft of the nodeselector info.

I have:
1. Changed the "Self-provisioned Projects" file to "Managing Projects" and made sure all the links were good. I added an overview and moved the "Self-provisioning Projects" bit down, then added a "Node Selector" bit after.
2. I got the node selector info from the resources from the [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1253125), [Jordan's GH issue](https://github.com/openshift/openshift-docs/issues/678), and [Luke's aggregating logging info](https://github.com/openshift/openshift-docs/blob/72cc83bb424c10824d9a1c1f09bfc28c832aa921/install_config/aggregate_logging.adoc).

One question though, is there a difference between node labels and the labels outlined in the "[Architecture](https://docs.openshift.com/enterprise/3.1/architecture/core_concepts/pods_and_services.html#labels)" section?


sosiouxme:
@bfallonf if I understand your question, yes, they are the same labels. All objects in the system can be given metadata specifying labels, so node labels look the same as labels on pods etc.

node *selectors* on pods just specify that the nodes they're scheduled on need to have a certain set of labels.

admin_guide/managing_projects.adoc
```diff
@@ -94,3 +101,40 @@ or:
 To request a new project, fill out the project request form located at
 https://internal.example.com/openshift-project-request.
 ----
+
+== Node Selectors
+
+Node selectors are used in conjunction with labels to control pod placement. You can assign a pod or set of pods to specific nodes, or limit a project to only run on a desired node.
```
sosiouxme: I would say a set of nodes. The set might only have one node, but we really don't want to encourage people to direct a pod to one specific node, because if that node goes down, their pod can't be rescheduled.

admin_guide/managing_projects.adoc
```diff
@@ -94,3 +101,40 @@ or:
...
+
+For example, if you have a OpenShift topology with multiple regions, node selectors help restrict specific OpenShift projects to only deploy pods into a specific region. 
+
+To create a project with a node selector, use the `--node-selector` option. For example, the following creates a new project named `MyProject`, but runs only on the nodes labeled `user-node` and `east`:
+
+====
+----
+$ oadm new-project MyProject --node-selector='type=user-node,zone=east'
```
sosiouxme: It's a minor thing, but I would rather say "region=east" than "zone" here. Typically you want your pods deployed in one region but across multiple zones.

admin_guide/managing_projects.adoc
```diff
@@ -94,3 +101,40 @@ or:
...
+----
+...
+metadata:
+  annotations:
+    openshift.io/node-selector: type=user-node,zone=east
+...
+----
+====
+
+For already-existing projects, edit the *_/etc/origin/master/master-config.yaml_* master configuration file and add a node selector value:
```
sosiouxme: I believe this is applied to any project without its own node-selector annotation, not just existing ones. It's a global default.

sosiouxme:
Some nits, but largely looks good.

admin_guide/managing_projects.adoc
bfallonf: @sosiouxme Edited to:

"Alternatively, edit the /etc/origin/master/master-config.yaml master configuration file and add a node selector value. This adds the node selector to all current projects without a nodeSelector value, as well as all future projects:"

Can I please get an ack that's what you meant?

bfallonf:
@sosiouxme All suggestions edited in. Just one question about the last point you made.

If there's anything else, let me know.

thoraxe:
How much nitpicking do you want?

admin_guide/managing_projects.adoc
sosiouxme: ack

sosiouxme: Well, I guess I'd say it a little differently. The default node selector applies for all projects, unless the admin overrides (per project). I think it's a little clearer as:

"This applies the node selector for pods created in all projects without a specified nodeSelector value:"

admin_guide/managing_projects.adoc
```diff
@@ -94,3 +101,40 @@ or:
...
+
+For example, if you have a OpenShift topology with multiple regions, node selectors help restrict specific OpenShift projects to only deploy pods into a specific region. 
+
+To create a project with a node selector, use the `--node-selector` option. For example, the following creates a new project named `MyProject`, but runs only on the nodes labeled `user-node` and `east`:
+
+====
+----
+$ oadm new-project MyProject --node-selector='type=user-node,region=east'
+----
+====
```
adellape: The project name can't have capital letters:

```
The project "MyProject" is invalid.
metadata.name: invalid value 'MyProject', Details: must be a DNS label (at most 63 characters, matching regex [a-z0-9]([-a-z0-9]*[a-z0-9])?): e.g. "my-name"
```

It may be worth adding a [NOTE] box here to clarify that `--node-selector` is only available with the `oadm` version of `new-project` and is therefore only a cluster admin operation. The `oc new-project` command doesn't have the `--node-selector` option, so non-cluster-admins cannot use the `--node-selector` option when self-provisioning projects:

```
$ oc new-project my-project-2 --node-selector='type=user-node,region=east'
Error: unknown flag: --node-selector
```

admin_guide/managing_projects.adoc
```diff
@@ -94,3 +101,40 @@ or:
 To request a new project, fill out the project request form located at
 https://internal.example.com/openshift-project-request.
 ----
+
+== Node Selectors
+
+Node selectors are used in conjunction with labels to control pod placement. You can assign a pod or set of pods to specific nodes, or limit a project to only run on a desired set of nodes.
```
adellape: To provide some more context on what's going on here wrt node labels (e.g., maybe they didn't know about them during initial installation), maybe at the front of this paragraph start with a line something like:

> Labels can be assigned to node hosts link:../../install_config/install/advanced_install.html#configuring-node-host-labels[during an advanced installation] or link:../../admin_guide/manage_nodes.html#updating-labels-on-nodes[updated post-installation].

Links to those existing sections for your clicking's sake:

https://docs.openshift.com/enterprise/latest/install_config/install/advanced_install.html#configuring-node-host-labels

https://docs.openshift.com/enterprise/latest/admin_guide/manage_nodes.html#updating-labels-on-nodes

admin_guide/managing_projects.adoc
```diff
@@ -94,3 +101,40 @@ or:
...
+Alternatively, edit the *_/etc/origin/master/master-config.yaml_* master configuration file and add a node selector value. This adds the node selector to all current projects without a `nodeSelector` value, as well as all future projects:
+
+====
+----
+...
+projectConfig:
+  nodeSelector: "type=user-node,region=east"
+...
+----
+====
```
adellape: Should mention you would need to restart the **atomic-openshift-master** service after making this change for it to take effect.

adellape:
@bfallonf Couple comments inline.

admin_guide/managing_projects.adoc
bfallonf: @sosiouxme  Ok. Done. Thanks for that.

bfallonf:
@thoraxe go crazy!

@adellape Done. 

Please, let me know if there's anything else for this PR.

admin_guide/managing_projects.adoc
```diff
@@ -10,12 +10,19 @@
...
+
+== Self-provisioning Projects
+
 You can allow developers to create their own projects. There is an endpoint
 that will provision a project according to a
 link:../dev_guide/templates.html[template]. The web console and `oc new-project` 
 command use this endpoint when a developer link:../dev_guide/projects.html[creates a new project].
 
-== Template for New Projects
+=== Template for New Projects
```
adellape: Minor nit, but possibly update this to be a gerund phrase, and add a manual anchor:

```
[[modifying-the-template-for-new-projects]]
=== Modifying the Template for New Projects
```

admin_guide/managing_projects.adoc
```diff
@@ -94,3 +101,55 @@ or:
 To request a new project, fill out the project request form located at
 https://internal.example.com/openshift-project-request.
 ----
+
+== Node Selectors
```
adellape: Same here, wrt gerund/anchor:

```
[[using-node-selectors-with-projects]]
== Using Node Selectors with Projects
```

Or possibly simply "Using Node Selectors"... A reason to also include the "with Projects", even if it seems redundant in a topic called "Managing Projects", is that the concept of "node selector" appears in other, non-strictly-project-related material.

Or maybe, "Setting Node Selectors for Projects".

admin_guide/managing_projects.adoc
```diff
@@ -94,3 +101,55 @@ or:
...
+
+Node selectors are used in conjunction with labels to control pod placement. 
+
+[NOTE]
+Labels can be assigned link:../install_config/install/advanced_install.html#configuring-node-host-labels[during an advanced installation], or link:../admin_guide/manage_nodes.html#updating-labels-on-nodes[added to a node after installation].
```
adellape: It's not broken, but, wrap this in `====` if you want to match the guideline style.

admin_guide/managing_projects.adoc
```diff
@@ -94,3 +101,55 @@ or:
...
+----
+...
+metadata:
+  annotations:
+    openshift.io/node-selector: type=user-node,region=east
+...
+----
+====
+
+Alternatively, edit the *_/etc/origin/master/master-config.yaml_* master configuration file and add a node selector value. This applies the node selector to pods created in all projects without a specified nodeSelector value:
```
adellape:
```
`*nodeSelector*`
```

admin_guide/managing_projects.adoc
```diff
@@ -94,3 +101,55 @@ or:
...
+  nodeSelector: "type=user-node,region=east"
+...
+----
+====
+
+Ensure to restart the master OpenShift service for the changes to take effect:
+
+====
+----
+$ systemctl restart atomic-openshift-master
```
adellape: This requires root privileges, so change the $ prompt to a # to signal it should be run as root (or as non-root w/ sudo).

adellape:
@bfallonf Looking good, some nits inline.

admin_guide/managing_projects.adoc
```diff
@@ -94,3 +101,55 @@ or:
...
+
+====
+----
+$ oadm new-project myproject \
+    --node-selector='type=user-node,region=east'
+----
+====
+
+[NOTE]
+While the `new-project` option is available for both `oadm` and `oc`, the cluster administrator and developer commands respectively, creating a new project with a node selector is only available with the `oadm` command. The `new-project` option is not available to project developers when self-provisioning projects.
```
adellape: Same here, wrt `====`.

admin_guide/managing_projects.adoc
bfallonf: I think "using Node Selectors" is fine here. This is kinda the first point for node selectors (which is why I kept it as "Node Selectors" in the first place), and any other times they might be mentioned in anchors they would have a "Using Node Selectors for X" in the title, or something similar.

admin_guide/managing_projects.adoc
bfallonf: I'm...not sure if I've ever done that. Done now.

bfallonf:
@adellape Updated

We can wait for an ack from @sosiouxme before we merge, as per #1519 .

admin_guide/managing_projects.adoc
```diff
@@ -94,3 +102,60 @@ or:
 To request a new project, fill out the project request form located at
 https://internal.example.com/openshift-project-request.
 ----
+
+[[using-node-selectors]]
+== Using Node Selectors
+
+Node selectors are used in conjunction with labels to control pod placement.
```
sosiouxme: This would be clearer if it more specifically said "with node labels".

admin_guide/managing_projects.adoc
```diff
@@ -94,3 +102,60 @@ or:
...
+== Using Node Selectors
+
+Node selectors are used in conjunction with labels to control pod placement. 
+
+[NOTE]
+====
+Labels can be assigned link:../install_config/install/advanced_install.html#configuring-node-host-labels[during an advanced installation], or link:../admin_guide/manage_nodes.html#updating-labels-on-nodes[added to a node after installation].
+====
+
+Using node selectors, you can assign a pod or set of pods to specific nodes, or limit a project to only run on a desired set of nodes. For example, if you have a OpenShift topology with multiple regions, node selectors help restrict specific OpenShift projects to only deploy pods into a specific region.
```
sosiouxme: Separate the use of node selectors into two use cases: developer and administrator. The developer can set node selectors on their pod specification (`nodeSelector` entry). The administrator can set a cluster-wide default node selector and override that at the project level. We really need to talk about the interactions at the three levels.

- The project node selector (if any) is the cluster-wide default, potentially overridden by one specified in the project annotation. Thus the project can be less restrictive than the default. In particular, if a project annotation has an empty value (`openshift.io/node-selector: ""`) then that project has no node selector, even if there is a cluster-wide default. This is of particular interest to admins who may want to prescribe a default for users and ignore it for infrastructure projects.
- Whatever the project node selector ends up being, if a user specifies a node selector in their pod spec, this is *merged* with the project node selector, making it more restrictive. The pod node selector cannot end up less restrictive than the project, so the project contents really are restricted to the admin's wishes.

admin_guide/managing_projects.adoc
bfallonf: @sosiouxme Yes, that's why i was asking if these labels are the same as the other labels. I think if we refer to them as "node labels" it might lead to confusion within people who have the same or less understanding as me... There's the note box after this that talks more about the labels. I think that's enough. Let me know if you disagree.

bfallonf:
@sosiouxme I added more info. Tried to separate between what a admin can do from what a dev can do. One Question:

How does the developer add a node selector exactly? @adellape said in a comment above that a developer cannot use the --new-project option in the same was an admin can here. Is it just as simple as modifying the pod spec, like it says here?

Let me know if there's anything else. Thanks for the help on this :fireworks:

admin_guide/managing_projects.adoc
sosiouxme: Alright I can see that potential confusion. However "with labels" is ambiguous as to where the labels go (nodes? pods? something else?). How about: "Node selectors are used in conjunction with labels on nodes ... " (or, labeled nodes)?

adellape: > (or, labeled nodes)?

+1

sosiouxme:
@bfallonf the developer cannot fiddle with project-level node selectors which is what alex was talking about. But they can set them on their pods; as best I can tell, we don't mention this *anywhere* in dev docs, and only obliquely in admin docs (https://docs.openshift.org/latest/admin_guide/scheduler.html#configurable-predicates). It's a Kubernetes concept and well-explained at e.g. http://kubernetes.io/v1.1/docs/user-guide/node-selection/README.html so I don't know that we need to duplicate those docs; can probably just link to them, somewhere in developer docs and in this section on how they are merged with the project-level node selector.

admin_guide/managing_projects.adoc
```diff
@@ -0,0 +1,171 @@
...
+$ oadm new-project myproject \
+    --node-selector='type=user-node,region=east'
+----
+====
+
+Once this command is run, this becomes the default node selector for all projects.
+
+[NOTE]
+====
+While the `new-project` option is available for both `oadm` and `oc`, the cluster administrator and developer commands respectively, creating a new project with a node selector is only available with the `oadm` command. The `new-project` option is not available to project developers when self-provisioning projects.
```
sosiouxme: `new-project` is the subcommand.... `--node-selector` is the option that is/not available.

admin_guide/managing_projects.adoc
```diff
@@ -0,0 +1,171 @@
...
+To create a project with a node selector, use the `--node-selector` option. The following creates a new project named `myproject` dictating that pods be deployed onto nodes labeled `user-node` and `east`:
+
+====
+----
+$ oadm new-project myproject \
+    --node-selector='type=user-node,region=east'
+----
+====
+
+Once this command is run, this becomes the default node selector for all projects.
```
sosiouxme: Nor for all projects... just for `myproject`, it becomes the base node selector for all pods therein. It might be helpful to introduce that terminology ("base node selector") to lay the groundwork for understanding that the developers can specify more selectors on top of these.

admin_guide/managing_projects.adoc
```diff
@@ -0,0 +1,171 @@
...
+----
+...
+metadata:
+  annotations:
+    openshift.io/node-selector: type=user-node,region=east
+...
+----
+====
+
+If the `openshift.io/node-selector` value remains empty, the project will not have a node selector, even if the cluster-wide default has been set. This means that, as an OpenShift administrator, you can set a default for any developers, but want to keep projects separate for infrastructure projects.
```
sosiouxme: "remains empty" is ambiguous here. There is a difference between not having any annotation (which is the default, leaving the node selector to the cluster default if any), and having one with an empty value "" (which replaces the cluster default with an empty selector). To make it more confusing, there's no way AFAIK to create the latter condition with `oadm new-project`; you have to edit the namespace (not the project!) after creating it. Yuk.

Also the second sentence doesn't read right. Suggestion given all this:

"
If the `openshift.io/node-selector` annotation is given the empty string (`""`), the project will have an empty base node selector, even if the cluster-wide default has been set. This means that, as an OpenShift administrator, you can set a default to restrict developer projects to a subset of nodes and still enable infrastructure projects to schedule the entire cluster.
"

admin_guide/managing_projects.adoc
```diff
@@ -0,0 +1,171 @@
...
+
+However, you can still dictate node selector values for all projects with a node selector value. If a developer changes the node selector in a pod configuration, the pod will have both the default and new node selectors, because the default node selector cannot be overridden by a developer.
```
sosiouxme: Not sure this makes sense. Perhaps an example will help.

If a project has been created with the above annotation (`openshift.io/node-selector: type=user-node,region=east`) and the user then sets a nodeSelector of "clearance=classified" on a pod in that project, the pod will only ever be scheduled on nodes that have all three labels (type=user-node *and* region=east *and* clearance=classified). If they decided they wanted to schedule on region=west and set a pod nodeSelector accordingly, they would be demanding nodes with labels region=east *and* region=west, which is impossible (labels can only get one value) so it would never be scheduled. Thus the developer can, per-pod, augment (but not override) the project node selector.

admin_guide/managing_projects.adoc
bfallonf: Fair enough. Changed to 'labeled nodes'.

bfallonf:
@sosiouxme I've included your thoughts in this revision. I'm not in love with the "base node selector" phrase, and I tried to do the same thing with default, but default doesn't work too well either. I've used "administrator-set node selector" a couple times, but I think the sentiment is there. I've also included some info in the dev_guide/projects topic.

My question about that is, why is the admin changing the project config, but the dev is changing the pod config? Should that be consistent?

Also, @adellape @tpoitras @ahardin-rh @tnguyen-rh if there's a better suggestion for the placement of the dev info, I'm all ears. For reference: this topic is about controlling the placement of new pods onto specific nodes. As the admin stuff is grouped with the project stuff, I placed it in dev_guide/projects, but tossed up whether it needs to be a new topic, or perhaps grouped with the dev_guide/pod_autoscaling topic and changing that to "Managing Projects" or something and making "Pod Auto-scaling" a part within, and "Node Selectors" another. LMKWYT!

sosiouxme:
> why is the admin changing the project config, but the dev is changing the pod config? 

@bfallonf projects are the main point at which the cluster-admin can set limits to rein in their multitenant users. The project user doesn't have the ability to change the project once it's created (and has only limited options for creating the project in the first place). This allows the cluster-admin to designate a particular set of nodes for a project to run on and know that the project's pods will actually be restricted to those nodes. But it still allows the users actually defining the pods in the project to further specify node placement needs they might have.

admin_guide/managing_projects.adoc
```diff
@@ -0,0 +1,171 @@
...
+Node selectors are used in conjunction with labeled nodes to control pod placement. 
+
+[NOTE]
+====
+Labels can be assigned link:../install_config/install/advanced_install.html#configuring-node-host-labels[during an advanced installation], or link:../admin_guide/manage_nodes.html#updating-labels-on-nodes[added to a node after installation].
+====
+
+As an OpenShift cluster administrator, you can set the cluster-wide default node selector to restrict pod placement to specific nodes. For example, if you have an OpenShift topology with multiple regions, node selectors help restrict specific OpenShift projects to only deploy pods onto nodes in a specific region.
+
+OpenShift developers can set a node selector on their project configuration if they wish to restrict nodes even further. This will be added to the cluster-wide node selector, meaning it will be more restrictive.
```
sosiouxme: No, they can't; they can only set node selectors on pods. This would be fine if it just said "pod configuration" instead of "project configuration".

admin_guide/managing_projects.adoc
```diff
@@ -0,0 +1,171 @@
...
+
+However, you can still dictate node selector values for all projects with a node selector value. If a project has been created with the above annotation (`openshift.io/node-selector: type=user-node,region=east`) and a developer sets another node selector on a pod in that project, for example `clearance=classified`, the pod will only ever be scheduled on nodes that have all three labels (`type=user-node`, `region=east` and `clearance=classified`). If they decided they wanted to schedule on `region=west` and set a pod node selector accordingly, they would be demanding nodes with labels `region=east` and `region=west`, which cannot work because labels can only be set to one value, so it would never be scheduled. Thus the developer can, per-pod, augment (but not override) the project node selector.
+
+Alternatively, you can edit the *_/etc/origin/master/master-config.yaml_* master configuration file and add a node selector value. This applies the node selector to pods created in all projects without a specified `*nodeSelector*` value:
```
sosiouxme: This would make more sense to me at the top of the section. First discuss the global default, then the project-specific override, then the pod-specific narrowing of node selector.

dev_guide/projects.adoc
```diff
@@ -165,3 +165,29 @@ To delete a project using the CLI:
...
+== Assigning Pods to Specific Nodes
+
+You can use node selectors in conjunction with labeled nodes to control pod placement.
+
+[NOTE]
+====
+OpenShift administrators can assign labels link:../install_config/install/advanced_install.html#configuring-node-host-labels[during an advanced installation], or link:../admin_guide/manage_nodes.html#updating-labels-on-nodes[added to a node after installation].
+====
+
+OpenShift cluster administrators can set the cluster-wide default node selector to restrict pod placement to specific nodes. As an OpenShift developer, you can set a node selector on a project configuration to restrict nodes even further. This will be added to the cluster-wide node selector, meaning it will be more restrictive.
```
sosiouxme: > As an OpenShift developer, you can set a node selector on a project configuration

nope, just on pods.

dev_guide/projects.adoc
```diff
@@ -165,3 +165,29 @@ To delete a project using the CLI:
...
+
+To add a node selector, edit a pod configuration, and add the `nodeSelector` value:
```
sosiouxme: This is really something the developer would configure as part of the pod creation (this text kind of suggests editing a running pod which you can't). It goes in the pod definition, whether that be for a standalone pod, or in a pod template in a DC or RC.

dev_guide/projects.adoc
```diff
@@ -165,3 +165,29 @@ To delete a project using the CLI:
...
+
+====
+----
+...
+nodeSelector:
+  type:user-node,region:east
```
sosiouxme: Proper YAML would look like:

```
nodeSelector:
  type: user-node
  region: east
```

I'm not sure I'd use this example though as the labels sound more cluster-admin-y. Maybe:

```
nodeSelector:
  disktype: ssd
```

I would prefer to link to kube docs (e.g. http://kubernetes.io/v1.1/docs/user-guide/node-selection/README.html) for a better in-context example. And would like to mention here that your labels are added to any the cluster-admin has specified by default / for your project.

dev_guide/projects.adoc
```diff
@@ -165,3 +165,29 @@ To delete a project using the CLI:
...
+...
+----
+====
+
+Any pods created will now be assigned to the labeled nodes.
```
sosiouxme: No, each pod specification has to include any labels you want on that pod.

"Pods created with a nodeSelector will be assigned to nodes with the specified label(s) and any labels specified for the project by the cluster administrator."

admin_guide/managing_projects.adoc
bfallonf: Hazards of adding a lot to the doc... Should be ok now.

dev_guide/projects.adoc
bfallonf: We're trying not to add links to upstream too much these days. I think the example works. And there was info about the admin-set ones in there already, but I changed it around a little and it should be better now.

bfallonf:
@sosiouxme Done to suggestions. Should be ok. Let me know if there's anything else.

admin_guide/managing_projects.adoc
```diff
@@ -0,0 +1,169 @@
...
+== Using Node Selectors
+
+Node selectors are used in conjunction with labeled nodes to control pod placement. 
+
+[NOTE]
+====
+Labels can be assigned link:../install_config/install/advanced_install.html#configuring-node-host-labels[during an advanced installation], or link:../admin_guide/manage_nodes.html#updating-labels-on-nodes[added to a node after installation].
+====
+
+As an OpenShift cluster administrator, you can set the cluster-wide default node selector to restrict pod placement to specific nodes. For example, if you have an OpenShift topology with multiple regions, node selectors help restrict specific OpenShift projects to only deploy pods onto nodes in a specific region.
```
sosiouxme: The regions example doesn't really make sense as a global default. I'd just say:

"As an OpenShift cluster administrator, you can set the cluster-wide default node selector to restrict pod placement to specific nodes for most projects."

And then I would pull in the text below that describes how to actually set the global default. The part that starts with "Alternatively,..."

The text "For example, if you have an OpenShift topology with multiple regions, node selectors help restrict specific OpenShift projects to only deploy pods onto nodes in a specific region." would go better as an intro to setting project node selectors below. "For example... To create a project..."

admin_guide/managing_projects.adoc
```diff
@@ -0,0 +1,169 @@
...
+
+OpenShift developers can set a node selector on their pod configuration if they wish to restrict nodes even further. This will be added to the cluster-wide node selector, meaning it will be more restrictive.
```
sosiouxme: This part I would say after describing the project-level node selector. It is added to whatever selector applies to the project, whether that's cluster default or project-specific, so I'd rephrase the last sentence as "This will be added to the project's node selector, meaning it will be more restrictive."

dev_guide/projects.adoc
```diff
@@ -165,3 +165,36 @@ To delete a project using the CLI:
 ----
 $ oc delete project <project_name>
 ----
+
+[[assigning-pods-to-specific-nodes]]
+
+== Assigning Pods to Specific Nodes
```
sosiouxme: this feels a little out of place here; if i were looking for how to affect node placement of pods, i wouldn't look under projects, and the only tie-in really is that the default is/can be by project.

not sure what would be a better section... something about creating apps, or editing deployment configurations...

dev_guide/projects.adoc
```diff
@@ -165,3 +165,36 @@ To delete a project using the CLI:
...
+
+You can use node selectors in conjunction with labeled nodes to control pod placement.
+
+[NOTE]
+====
+OpenShift administrators can assign labels link:../install_config/install/advanced_install.html#configuring-node-host-labels[during an advanced installation], or link:../admin_guide/manage_nodes.html#updating-labels-on-nodes[added to a node after installation].
+====
+
+OpenShift cluster administrators can set the cluster-wide default node selector to restrict pod placement to specific nodes. As an OpenShift developer, you can set a node selector on a pod configuration to restrict nodes even further.
```
sosiouxme: "OpenShift cluster administrators can set the cluster-wide default node selector" well, it could be cluster-wide or specific to your project. Since we don't give them any way to know the difference, might as well just say "OpenShift cluster administrators can set the default node selector for your project ..."

dev_guide/projects.adoc
```diff
@@ -165,3 +165,36 @@ To delete a project using the CLI:
...
+====
+----
+...
+nodeSelector:
+  disktype: ssd
+...
+----
+====
+
+Pods created using a template configuration with a node selector are assigned to nodes with the specified labels.
```
sosiouxme: s/using a template configuration//

dev_guide/projects.adoc
```diff
@@ -165,3 +165,36 @@ To delete a project using the CLI:
...
+
+The labels specified here are added in conjunction to the labels link:../admin_guide/managing_projects.html#using-node-selectors[added by a cluster administrator]. For example, if a project has the `type=user-node` and `region=east` labels added to a project by the OpenShift administrator, and you add the above `disktype: ssd` label to a pod, the pod will only ever be scheduled on nodes that have all three labels. 
+
+[[NOTE]]
+====
+Labels can only be set to one value, so setting a node selector of `region=west` in a pod configuration that has `region=east` as the adnimistrator-set default, results in a pod that will never be scheduled.
```
sosiouxme: administrator

dev_guide/projects.adoc
bfallonf: @sosiouxme Yes, i brought it up with the team a few updates ago, but got no response. I can worry about that when we get to peer review. As long as all the info is there for now.

admin_guide/managing_projects.adoc
bfallonf: @sosiouxme I moved the "Alternatively..." section up, but from what I understand, this is the exact same process, just a different way to go about it. If that's not true, then I've definitely received conflicting information from somewhere. 

Either way, now the example use case I was putting forward doesn't exist anymore, and this is just a topic full of info...

bfallonf:
@sosiouxme I've made edits as to your suggestions. Can I ask if they are in the right direction? This has gone through so many revisions that's it's become a proper Frankenstein's Monster. I'll admit, my understanding of all this has had to change with each revision and I get pretty confused each time you've added suggestions. I think what's there now conflicts with what I've had each revision... 

Either way, if the right info is there, I can work on the placement. I can't spend much more time on this.

dev_guide/projects.adoc
adellape: How bout at the end of `dev_guide/deployments`? At least for now.

dev_guide/projects.adoc
```diff
@@ -165,3 +165,36 @@ To delete a project using the CLI:
...
+
+[[assigning-pods-to-specific-nodes]]
+
+== Assigning Pods to Specific Nodes
+
+You can use node selectors in conjunction with labeled nodes to control pod placement.
+
+[NOTE]
+====
+OpenShift administrators can assign labels link:../install_config/install/advanced_install.html#configuring-node-host-labels[during an advanced installation], or link:../admin_guide/manage_nodes.html#updating-labels-on-nodes[added to a node after installation].
```
adellape: Consider that the Dedicated and Online versions of this can't link to these install_config and admin_guide books.

admin_guide/managing_projects.adoc
```diff
@@ -0,0 +1,169 @@
...
+|*PROJECT_REQUESTING_USER*
+|The username of the requesting user.
+|===
+
+Access to the API is granted to developers with the
+link:../architecture/additional_concepts/authorization.html#roles[`self-provisioner`
+role] and the `self-provisioners` cluster role binding. This role is available
+to all authenticated developers by default.
+
+=== Disabling Self-provisioning
```
adellape: Set a unique manual anchor.

adellape:
@bfallonf @sosiouxme Few comments inline.

admin_guide/managing_projects.adoc
bfallonf: @adellape Isn't this below? Is that not right?

bfallonf:
@adellape comments addressed :pineapple:

admin_guide/managing_projects.adoc
adellape: Yea, looks like it's there. Must have missed it before.

admin_guide/managing_projects.adoc
```diff
@@ -0,0 +1,170 @@
...
+====
+----
+$ oadm new-project myproject \
+    --node-selector='type=user-node,region=east'
+----
+====
+
+Once this command is run, this becomes the adminstrator-set node selector for all pods contained in the indicated project.
+
+OpenShift developers can set a node selector on their pod configuration if they wish to restrict nodes even further. This will be added to the project node selector, meaning it will be more restrictive.
```
sosiouxme: Feels out of place in the middle of describing the project node selector. Suggest moving down

admin_guide/managing_projects.adoc
```diff
@@ -0,0 +1,170 @@
...
+
+However, you can still dictate node selector values for all projects with a node selector value. If a project has been created with the above annotation (`openshift.io/node-selector: type=user-node,region=east`) and a developer sets another node selector on a pod in that project, for example `clearance=classified`, the pod will only ever be scheduled on nodes that have all three labels (`type=user-node`, `region=east` and `clearance=classified`). If they decided they wanted to schedule on `region=west` and set a pod node selector accordingly, they would be demanding nodes with labels `region=east` and `region=west`, which cannot work and will never be scheduled, because labels can only be set to one value. Thus the developer can, per-pod, augment (but not override) the project node selector.
```
sosiouxme: "However" seems like a non-sequitur here. How about (combining the intro from above):

> OpenShift developers can set a node selector on their pod configuration if they wish to restrict nodes even further. This will be added to the project node selector, meaning it will be more restrictive. Therefore, the administrator can still dictate node selector values for all projects with a node selector value.
> 
> As an example, if a project has been created...

admin_guide/managing_projects.adoc
```diff
@@ -0,0 +1,170 @@
...
+[[using-node-selectors]]
+== Using Node Selectors
+
+Node selectors are used in conjunction with labeled nodes to control pod placement. 
+
+[NOTE]
+====
+Labels can be assigned link:../install_config/install/advanced_install.html#configuring-node-host-labels[during an advanced installation], or link:../admin_guide/manage_nodes.html#updating-labels-on-nodes[added to a node after installation].
+====
+
```
sosiouxme: Are these three sections long enough to merit sub-headers? E.g. "Setting the cluster default node selector"

admin_guide/managing_projects.adoc
```diff
@@ -0,0 +1,170 @@
...
+Ensure to restart the master OpenShift service for the changes to take effect:
+
+====
+----
+# systemctl restart atomic-openshift-master
+----
+====
+
```
sosiouxme: E.g. subheader "Setting per-project node selectors"

admin_guide/managing_projects.adoc
```diff
@@ -0,0 +1,170 @@
...
+...
+metadata:
+  annotations:
+    openshift.io/node-selector: type=user-node,region=east
+...
+----
+====
+
+If `openshift.io/node-selector` is set to an empty string (`oadm new-project --node-selector=""`), the project will not have an adminstrator-set node selector, even if the cluster-wide default has been set. This means that, as an OpenShift administrator, you can set a default to restrict developer projects to a subset of nodes and still enable infrastructure or other projects to schedule the entire cluster.
+
```
sosiouxme: E.g. subheader "Developer-specified node selectors"

dev_guide/deployments.adoc
```diff
@@ -670,3 +670,39 @@ $ oc scale dc frontend --replicas=3
...
+
+OpenShift cluster administrators can set the default node selector for your project in order to restrict pod placement to specific nodes. As an OpenShift developer, you can set a node selector on a pod configuration to restrict nodes even further.
+
+To add a node selector when creating a pod, edit the pod configuration, and add the `nodeSelector` value. This can be added to a single pod configuration, or in a pod template:
+
+====
+----
+...
+nodeSelector:
+  disktype: ssd
```
sosiouxme: If we're not going to link to upstream doc, then this should have a little more context as they'd have to look at the API spec to figure out where exactly this goes. I suggest the example look like:

```
apiVersion: v1
kind: Pod
spec:
  nodeSelector:
    disktype: ssd
...
```

dev_guide/deployments.adoc
```diff
@@ -670,3 +670,39 @@ $ oc scale dc frontend --replicas=3
...
+...
+----
+====
+
+Pods created with a node selector are assigned to nodes with the specified labels.
+
+The labels specified here are added in conjunction to the labels link:../admin_guide/managing_projects.html#using-node-selectors[added by a cluster administrator]. For example, if a project has the `type=user-node` and `region=east` labels added to a project by the OpenShift administrator, and you add the above `disktype: ssd` label to a pod, the pod will only ever be scheduled on nodes that have all three labels.
```
sosiouxme: "added in conjunction to" sounds odd to me at least. How about "are used in conjunction with"?

sosiouxme:
@bfallonf @adellape sorry for the long silence, and for picking nits off the dead horse a bit longer, but this is mostly pretty good.

General question: sometimes it's "OpenShift cluster administrator", sometimes "OpenShift administrator", sometimes "cluster administrator". They're all for the same thing, and need to be distinguished at least a little from project administrators (which will be most users). Is it worth trying to standardize this terminology? The three-word version seems unwieldy but it seems like it should be consistent. "Cluster administrator" seems about right to me.

adellape:
> The three-word version seems unwieldy but it seems like it should be consistent. "Cluster administrator" seems about right to me.

+1

bfallonf:
@sosiouxme Thanks for the suggestions. The newest revision should have them all. 

I'll presume that's all and put this onto peer review. @adellape @ahardin-rh @tnguyen-rh @tpoitras if there's any comments, let me know. If not, merge away.

sosiouxme:
:+1:

admin_guide/managing_projects.adoc
```diff
@@ -0,0 +1,178 @@
...
+To create your own custom project template:
+
+. Start with the current default project template:
++
+----
+$ oadm create-bootstrap-project-template -o yaml > template.yaml
+----
+
+. Modify the template by adding objects or modifying existing objects, then load the template:
```
tpoitras: Should this be two steps?

admin_guide/managing_projects.adoc
```diff
@@ -0,0 +1,178 @@
...
+[NOTE]
+====
+Labels can be assigned link:../install_config/install/advanced_install.html#configuring-node-host-labels[during an advanced installation], or link:../admin_guide/manage_nodes.html#updating-labels-on-nodes[added to a node after installation].
+====
+
+=== Setting the Cluster-wide Default Node Selector
+
+As a cluster administrator, you can set the cluster-wide default node selector to restrict pod placement to specific nodes. 
+
+Edit the *_/etc/origin/master/master-config.yaml_* master configuration file and add a node selector value. This applies the node selector to pods created in all projects without a specified `*nodeSelector*` value:
```
tpoitras: This reads awkward/redundant to me, I think you could lose "master configuration" and just say "file":

Edit the *_/etc/origin/master/master-config.yaml_* file...

admin_guide/managing_projects.adoc
```diff
@@ -0,0 +1,178 @@
...
+====
+----
+...
+projectConfig:
+  nodeSelector: "type=user-node,region=east"
+...
+----
+====
+
+Ensure to restart the master OpenShift service for the changes to take effect:
```
tpoitras: "Ensure to restart" feels weird and overly formal. It's not wrong, but would you consider just "Restart the master..."? Or, reworded "For the changes to take effect, you must restart the master OpenShift service:"

admin_guide/managing_projects.adoc
```diff
@@ -0,0 +1,178 @@
...
+The following creates a new project named `myproject` and dictates that pods be deployed onto nodes labeled `user-node` and `east`:
+
+====
+----
+$ oadm new-project myproject \
+    --node-selector='type=user-node,region=east'
+----
+====
+
+Once this command is run, this becomes the adminstrator-set node selector for all pods contained in the indicated project.
```
tpoitras: s/indicated/specified ?

admin_guide/managing_projects.adoc
```diff
@@ -0,0 +1,178 @@
...
+
+[NOTE]
+====
+While the `new-project` subcommand is available for both `oadm` and `oc`, the cluster administrator and developer commands respectively, creating a new project with a node selector is only available with the `oadm` command. The `new-project` subcommand is not available to project developers when self-provisioning projects.
+====
+
+Using the `oadm new-project` command adds an `annotation` section to the project configuration. Changing the `openshift.io/node-selector` value overrides the default:
```
tpoitras: Should you mention the location of the project configuration file?

admin_guide/managing_projects.adoc
```diff
@@ -0,0 +1,178 @@
...
+
+=== Developer-specified Node Selectors
+
+OpenShift developers link:../dev_guide/deployments.adoc#assigning-pods-to-specific-nodes[can set a node selector on their pod configuration] if they wish to restrict nodes even further. This will be added to the project node selector, meaning that you can still dictate node selector values for all projects that have a node selector value.
+
+For example, if a project has been created with the above annotation (`openshift.io/node-selector: type=user-node,region=east`) and a developer sets another node selector on a pod in that project, for example `clearance=classified`, the pod will only ever be scheduled on nodes that have all three labels (`type=user-node`, `region=east` and `clearance=classified`). If they decided they wanted to schedule on `region=west` and set a pod node selector accordingly, their pods would be demanding nodes with labels `region=east` and `region=west`, which cannot work. The pods will never be scheduled, because labels can only be set to one value. Thus the developer can, per-pod, augment (but not override) the project node selector.
```
tpoitras: I'd get rid of "above" and the parentheses, since you're including the values here anyway, and because you don't know if/when someone will modify the above.

dev_guide/deployments.adoc
```diff
@@ -670,3 +670,42 @@ $ oc scale dc frontend --replicas=3
...
+    disktype: ssd
+...
+...
+----
+====
+
+Pods created when the node selector is in place are assigned to nodes with the specified labels.
+
+The labels specified here are used in conjunction with the labels link:../admin_guide/managing_projects.html#using-node-selectors[added by a cluster administrator]. For example, if a project has the `type=user-node` and `region=east` labels added to a project by the cluster administrator, and you add the above `disktype: ssd` label to a pod, the pod will only ever be scheduled on nodes that have all three labels. 
+
```
tpoitras: As an OpenShift developer, how can I view the default node selector configuration that was configured by the cluster administrator?

admin_guide/managing_projects.adoc
bfallonf: Hmm this topic was here before I got in. I think the command is doing both...

admin_guide/managing_projects.adoc
bfallonf: Changed to:

"Restart the OpenShift service for the changes to take effect"

admin_guide/managing_projects.adoc
bfallonf: You have to edit the project configuration. But I getcha. I threw a line about that in there.

admin_guide/managing_projects.adoc
bfallonf: I think if I do that it's harder to understand the example... But I will try rewrite this better explain it.

bfallonf:
@tpoitras Thanks for the review. If there's nothing else (from you or anyone else) I'll get this merged.

admin_guide/managing_projects.adoc
sosiouxme: You modify the template by editing the file. Then you load the file as below. Yes, it's two steps.

admin_guide/managing_projects.adoc
sosiouxme: On Wed, Mar 2, 2016 at 12:20 AM, Timothy Poitras <notifications@github.com>
wrote:

> In admin_guide/managing_projects.adoc
> <https://github.com/openshift/openshift-docs/pull/1490#discussion_r54680881>
> :
>
> > +Using the `oadm new-project` command adds an `annotation` section to the project configuration. Changing the `openshift.io/node-selector` <http://openshift.io/node-selector> value overrides the default:
>
> Should you mention the location of the project configuration file?
>
>

At this point we're not talking about the project configuration file. We're
talking about the object that represents the project in the API, so that
users have some idea where they can edit it after creation.

dev_guide/deployments.adoc
sosiouxme: On Wed, Mar 2, 2016 at 12:34 AM, Timothy Poitras <notifications@github.com>
wrote:
>
> As an OpenShift developer, how can I view the default node selector
> configuration that was configured by the cluster administrator?
>
>
>
I don't believe there is a way at this time. They can see the project
annotation if it exists but not the global default. Although it's easy to
figure out what it is by creating a pod and looking at the node selector it
ends up with. I don't think that's worth mentioning.

admin_guide/managing_projects.adoc
bfallonf: @sosiouxme Cool. Thanks. I've added in an 'edit the template' step.

admin_guide/managing_projects.adoc
bfallonf: Ok. Changed to refer to project and not project config.

dev_guide/deployments.adoc
bfallonf: Ah ok. I was mistaken. I'll take what I added.

bfallonf:
Ok. Suggestions taken in. I'm going to merge this. If there's anything else, I'll make a follow-up PR.

bfallonf:
**revhistory note:**

(Cluster Administration -> Managing Projects)

```
Added the link:managing_projects.adoc#using-node-selectors[Using Node Selectors] section.
```

(Developer Guide -> Deployments)

```
Added the link:deployments.adoc#assigning-pods-to-specific-nodes[Assigning Pods to Specific Nodes] section.
```

adellape:
@bfallonf s/adoc/html in the links

bfallonf:
@adellape Oops

PR #1501, opened 2016-01-29
Updated images doc

soltysh:
@adellape while looking at one bug I found we have outdated the images part of the doc. Here are the modifications I've applied:
- updated all jsons to yamls
- moved Image Pull Policy to the end of the docs, since it's loosely coupled with the docs itself
- updated tag tracking, changed to adding tag
- added private registry access using secrets.

@smarterclayton ptal the secrets part

### DIFF
--- a/architecture/core_concepts/builds_and_image_streams.adoc
+++ b/architecture/core_concepts/builds_and_image_streams.adoc
@@ -134,39 +134,33 @@ a build or a deployment.
 .Image Stream Object Definition
 ====
 
-[source,json]
+[source,yaml]
 ----
-{
-  "kind": "ImageStream",
-  "apiVersion": "v1",
-  "metadata": {
-    "name": "origin-ruby-sample",
-    "namespace": "p1",
-    "selfLink": "/osapi/v1/namesapces/p1/imageStreams/origin-ruby-sample",
-    "uid": "480dfe73-f340-11e4-97b5-001c422dcd49",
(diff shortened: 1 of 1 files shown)


smarterclayton:
Looks great, thanks for writing it up

adellape:
@soltysh The private registry secrets stuff is at least partly discussed in the Dev Guide here:

https://docs.openshift.org/latest/dev_guide/image_pull_secrets.html#allowing-pods-to-reference-images-from-other-secured-registries

But, I think I'd like to merge your PR as-is for now and then follow up by taking some of the task-based stuff (e.g., the oc tag/secrets examples) that's accumulated in this Arch topic and move/work it into to the Dev Guide topic, and possibly rename it "Managing Image Streams" or something.

There's also this ["Accessing the Registry Directly"](https://docs.openshift.org/latest/install_config/install/docker_registry.html#access) section in the "Deploying a Docker Registry" topic that could get shared into the Dev Guide topic, around that "Integrated Registry Authentication and Authorization" subsection.

If you're into this approach, I'll merge here and follow up.

soltysh:
@adellape it's closer to the first link you've mentioned but here it specifically is talking about importing images. I'm ok with merging and followups.

adellape:
@soltysh Yeah, understood. I'll make sure not to lose that aspect while reworking/moving things around. Thanks.

Issue #1500, opened 2016-01-29
Need to change the liveness/readiness probe scheme as part of securing the registry


liggitt:
https://github.com/openshift/openshift-docs/blob/master/install_config/install/docker_registry.adoc#securing-the-registry

When changing a registry to serve over TLS, in the deploymentconfig, change the livenessProbe and readinessProbe scheme from `HTTP` to `HTTPS`:
```
        livenessProbe:
          httpGet:
            path: /healthz
            port: 5000
            scheme: HTTPS
            ...
        readinessProbe:
          httpGet:
            path: /healthz
            port: 5000
            scheme: HTTPS
            ...
```

PR #1510, opened 2016-02-01
Bug 1302956: update scheme in probes when securing a registry

0xmichalis:
Fixes https://github.com/openshift/openshift-docs/issues/1500
and https://bugzilla.redhat.com/show_bug.cgi?id=1302956


adellape:
:+1: Thanks.

PR #1528, opened 2016-02-02
Document new SCC allowEmptyDirVolumePlugin restriction.

dgoodwin:
This new setting is similar to allowHostDirVolumePlugin, but is assumed to be
true by default for compatibility with existing deployments after upgrade.

Functionality is not yet merged but hopefully soon via origin PR https://github.com/openshift/origin/pull/6659

PR #1529, opened 2016-02-02
Outline Online EmptyDir volume restrictions and workaround.

dgoodwin:
EmptyDir volumes are expected to be disabled in Online (potentially for a short
period of time so this patch will likely get reverted someday).

EmptyDir volumes are something a user can explicitly add to their PodSpecs, and
in some cases new-app will automatically do it for them. As such we need to
clarify firstly that this volume type is disabled in Online, how the error will
surface (which can vary depending on if the user is creating a template/DC/RC
vs a pod itself), and how to work around the problem.

This PR depends on the following origin PR: https://github.com/openshift/origin/pull/6659

And is somewhat related to the SCC Docs PR for AllowEmptyDirVolumePlugin: https://github.com/openshift/openshift-docs/pull/1528

### DIFF
--- a/dev_guide/new_app.adoc
+++ b/dev_guide/new_app.adoc
@@ -214,7 +214,7 @@ link:../install_config/install/docker_registry.html#securing-the-registry[secure
 with SSL],
 endif::[]
 ifdef::openshift-dedicated[]
-secured with SSL, 
+secured with SSL,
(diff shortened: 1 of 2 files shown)


adellape:
:+1: Thanks.

revhistory note to be provided in follow-up PR.

PR #1555, opened 2016-02-08
Document the RunOnceDuration plug-in

csrwng:
Adds documentation for the RunOnceDuration plugin. 

This should only be able to releases that include the admin guide (origin and enterprise).


csrwng:
@sosiouxme @dgoodwin ptal

dgoodwin:
LGTM, good reference for what I'm working on as well, thanks.

adellape:
:+1: Thanks.

Will add **revhistory note** in follow-up PR.

PR #1565, opened 2016-02-09
Add GitHub organization config

liggitt:
Feature added in https://github.com/openshift/origin/pull/7161

### DIFF
--- a/install_config/configuring_authentication.adoc
+++ b/install_config/configuring_authentication.adoc
@@ -843,6 +843,13 @@ Using GitHub as an identity provider requires users to get a token using
 `<master>/oauth/token/request` to use with command-line tools.
 ====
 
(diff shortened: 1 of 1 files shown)


adellape:
:+1: Thanks.

adellape:
**revhistory note** will appear in follow-up PR.

liggitt:
the revhistory refers to noting which version the `organizations` attribute appeared in? if so, +1

adellape:
@liggitt That, yes, but also about just noting what has changed in the docs themselves, specifically the downstream docs (e.g., https://docs.openshift.com/enterprise/latest/welcome/revhistory_full.html).

I should probably send a PSA to the list about our use of revhistory notes in PRs so these comments of ours are less mysterious.

PR #1571, opened 2016-02-10
autoscale cmd docs

0xmichalis:
autoscale command is in master: https://github.com/openshift/origin/pull/6861


adellape:
:+1: Thanks.

PR #1572, opened 2016-02-10
Clarify when we switch to running commands inside a master host.

cben:
Please verify this makes sense, I'm following these instructions for the first time in my life.

### DIFF
--- a/install_config/install/advanced_install.adoc
+++ b/install_config/install/advanced_install.adoc
@@ -930,9 +930,9 @@ link:#verifying-the-installation[Verifying the Installation].
 
 == Verifying the Installation
 
-After the installer completes, you can verify that the master is started and
-nodes are registered and reporting in *Ready* status by running the following as
-*root*:
+After the installer completes, verify that the master is started and
+nodes are registered and reporting in *Ready* status.
+*On the master host*, run the following as root:
 
 ====
 ----


install_config/install/advanced_install.adoc
```diff
@@ -931,8 +931,8 @@ link:#verifying-the-installation[Verifying the Installation].
 == Verifying the Installation
 
 After the installer completes, you can verify that the master is started and
-nodes are registered and reporting in *Ready* status by running the following as
-*root*:
+nodes are registered and reporting in *Ready* status by sshing into the master
+host and running the following as *root*:
```
adellape: How about:

"After the installer completes, verify that the master is started and nodes are registered and reporting in Ready status. On the master host, run the following as root:"

cben: Amended (but decided to add emphasis *On the master host*).

It also crossed my mind that switching to full `root@master.example.com#` prompts could help, as they convey exactly the desired information - which user on which host.  But that might be more confusing than useful unless it's applied consistently across docs...

adellape:
:+1: Thanks.

adellape:
**revhistory note:**

(Install & Config -> Installing -> Advanced Installation)

```
Clarify in the link:../install_config/install/advanced_install.html#verifying-the-installation[Verifying the Installation] section to run the `oc get nodes` command on the master host.
```

PR #1576, opened 2016-02-10
[DO NOT MERGE] Add page-heading navigation bar

nhr:
@luciddreamz thought you might be interested in this; it adds a right-hand in-page nav similar to what you can see in the [Bootstrap.js docs](http://getbootstrap.com/components/#glyphicons).


adellape:
Hey that's neat.

adellape:
What's the thought on the body text becoming more narrow as a side-effect?

luciddreamz:
Having both a left and right navigation is a bit concerning. I wonder if there is some way to combine that effect in with the left navigation like with this: https://www.fullcontact.com/developer/docs/name/

nhr:
Yeah, I don't love the specific way I've got it with this PR, but just wanted to throw it out there as an idea. @CowboysFan had asked for something along these lines and I am totally open to looking at other approaches.

PR #1581, opened 2016-02-12
Use patch instead of sed

liggitt:


adellape:
@liggitt In OSE 3.1.1.6, I have a readiness + liveness probe in the router dc but only a liveness in the registry dc (created new in 3.1.1):

http://pastebin.test.redhat.com/350234

But I see PRs/Issues in origin talking about readiness probes in the registry so I'm confused.

liggitt:
readiness probe was added in https://github.com/openshift/origin/pull/6439, exists in 1.1.2+

liggitt:
you could separate the command into two commands, one to patch liveness and one to patch readiness, so they could be run independently if they have an older registry config

adellape:
Closing in favor of https://github.com/openshift/openshift-docs/pull/1727 for the liveness probe and https://github.com/openshift/openshift-docs/pull/1728 for adding an additional step for the readiness probe (separated so we can get it into OSE docs starting with 3.2).

PR #1582, opened 2016-02-12
Bug 1298619, Added SELinux guidance for installation

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1298619

### DIFF
--- a/install_config/install/prerequisites.adoc
+++ b/install_config/install/prerequisites.adoc
@@ -340,6 +340,28 @@ link:../../install_config/persistent_storage/persistent_storage_gce.html[GCE
 Persistent Disks], and
 link:../../install_config/persistent_storage/persistent_storage_iscsi.html[iSCSI].
 
+[[prereq-selinux]]
+
+=== SELinux
+
+Security-Enhanced Linux (SELinux) must be enabled on all of the servers before
+installing OpenShift or the installer will fail. Also, configure
+`*SELINUXTYPE=targeted*` in the *_etc/selinux/config_* file:
+
+----
+# This file controls the state of SELinux on the system.
+# SELINUX= can take one of these three values:
+#     enforcing - SELinux security policy is enforced.
+#     permissive - SELinux prints warnings instead of enforcing.
+#     disabled - No SELinux policy is loaded.
+SELINUX=enforcing
+# SELINUXTYPE= can take one of these three values:
+#     targeted - Targeted processes are protected,
+#     minimum - Modification of targeted policy. Only selected processes are protected.
+#     mls - Multi Level Security protection.
+SELINUXTYPE=targeted
+----
+
 [[host-preparation]]
 
 == Host Preparation


ahardin-rh:
@adellape , @bfallonf, can you please eyeball this?

ahardin-rh:
Approved by reporter, Eric Jones

bfallonf:
@ahardin-rh Heyo. So this flagged a memory I had from a security talk they gave in Brisbane one day, so I brought it up with one of the security guys (Wade Mealing), and yes, we're not supposed to be recommending using "setenforce=0" anywhere. So just taking out the final part ("Alternatively...") and this should be ok.

My own question is, why would Selinux be disabled? Obviously it's not by default, as I've never had this issue anytime I've tried to install. Also, should we really be recommending that we disable Selinux? From what I can tell, that might be a bad idea security-wise, right?

trevorjay:
@bfallonf You'd be happy to know Wade is who told me about this thread. <sidious>The circle is complete.</sidious>

I would strongly recommend striking how to disable SELinux full stop. It isn't just a security issue (which, of course it is) or a support issue for the community/Red Hat (again, which it is). Disabling SELinux is likely to cause the worst kind of problems with OpenShift. By worst I mean intermittent or seemingly random and that depend on *exactly* what containers you are running. You know, the kind where debugging them destroys your sanity.

bfallonf:
@trevorjay Cool. Thanks for that. 

@ahardin-rh So maybe a warning box about "The OpenShift installer needs Selinux enabled. You shouldn't disable Selinux, but if you do, here's how to enable it again:...". But I'll let you figure out the details :)

ahardin-rh:
@trevorjay @bfallonf thanks for your comments. I removed the details on disabling and kept the focus on the fact that SELinux should be enabled.

ahardin-rh:
**revhistory note:**

For Install Config --> Prerequisites

    Added a link:../install_config/install/prerequisites.htmll#prereq-selinux[SELinux] section to include guidance that SELinux must be enabled, or the installer will fail.

trevorjay:
@ahardin-rh Looks good to me.

install_config/install/prerequisites.adoc
```diff
@@ -340,6 +340,28 @@ link:../../install_config/persistent_storage/persistent_storage_gce.html[GCE
 Persistent Disks], and
 link:../../install_config/persistent_storage/persistent_storage_iscsi.html[iSCSI].
 
+[[prereq-selinux]]
+
+=== SELinux
+
+Security-Enhanced Linux (SELinux) must be enabled on all of the servers before
+installing OpenShift or the installer will fail. Also, configure
+`*SELINUXTYPE=targeted*` in the *_etc/selinux/config_* file:
```
adellape: @ahardin-rh Missing a `/` at the beginning of the file path.

ahardin-rh: Follow-up PR: https://github.com/openshift/openshift-docs/pull/1644

PR #1584, opened 2016-02-12
Describe system:authenticated:oauth

liggitt:


adellape:
**revhistory note:**

(Arch -> Addt'l Concepts -> Authentication)

```
Added the *system:authenticated:oauth* virtual group.
```

PR #1456, opened 2016-01-20
Clarify description of default identity providers

rhuss:
Moved the description of the default identity providers into conditional asciidoc sections for OSE and Origin, so that it fits to the selected providers.

PR #1601, opened 2016-02-16
fix typo for setting up the openshift-origin default authentication

PI-Victor:
this fixes the below error which occurs when the user sees the docs as part of origin and not ose.

``````
the Allow All identity provider is used by default, which denies access for all user names and passwords. ```

``````

### DIFF
--- a/install_config/configuring_authentication.adoc
+++ b/install_config/configuring_authentication.adoc
@@ -21,23 +21,17 @@ access tokens] to authenticate themselves to the API.
 
 As an administrator, you can configure OAuth using the
 link:../install_config/master_node_configuration.html[master configuration file] to specify an
-link:#identity-providers[identity provider]. If you installed OpenShift using
-the
+link:#identity-providers[identity provider].
 ifdef::openshift-enterprise[]
-link:../install_config/install/quick_install.html[Quick Installation] or
-endif::[]
-link:../install_config/install/advanced_install.html[Advanced Installation]
-method, the
-ifdef::openshift-enterprise[]
-link:#DenyAllPasswordIdentityProvider[Deny All]
-endif::[]
-ifdef::openshift-origin[]
-link:#AllowAllPasswordIdentityProvider[Allow All]
-endif::[]
-identity provider is used by default, which denies access for all user names and
-passwords. To allow access, you must choose a different identity provider and
-configure the master configuration file appropriately (located at
+If you installed OpenShift using the
+link:../install_config/install/quick_install.html[Quick Installation]  method,
+the link:#DenyAllPasswordIdentityProvider[Deny All] identity provider is used
+by default, which denies access for all user names and passwords. To allow
+access, you must choose a different identity provider and configure the master
+configuration file appropriately (located at
 *_/etc/origin/master/master-config.yaml_* by default).
+endif::[]
+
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by PI-Victor** [comment]
ping @openshift/team-documentation

File: install_config/configuring_authentication.adoc [hunk 1]
```diff
@@ -21,23 +21,17 @@ access tokens] to authenticate themselves to the API.
...
-link:#DenyAllPasswordIdentityProvider[Deny All]
-endif::[]
-ifdef::openshift-origin[]
-link:#AllowAllPasswordIdentityProvider[Allow All]
-endif::[]
-identity provider is used by default, which denies access for all user names and
-passwords. To allow access, you must choose a different identity provider and
-configure the master configuration file appropriately (located at
+If you installed OpenShift using the
+link:../install_config/install/quick_install.html[Quick Installation]  method,
```
tpoitras: Why remove mention of the advanced installation method?

install_config/configuring_authentication.adoc
```diff
@@ -21,23 +21,17 @@ access tokens] to authenticate themselves to the API.
...
+the link:#DenyAllPasswordIdentityProvider[Deny All] identity provider is used
```
tpoitras: This means that Origin no longer has Allow All as the default? And now uses Deny All?

PI-Victor: this line only shows up in the OSE docs, not the origin ones. It is clarified what the default behavior is in the paragraph below.

install_config/configuring_authentication.adoc
PI-Victor: again, for the same reason as below. bu if you have any suggestions as to how to reword this and keep the link, i can update it. it's best seen when you look at the compiled docs rather than the diff, and switch between ose and origin

adellape:
This was fixed in https://github.com/openshift/openshift-docs/pull/1456.

PI-Victor:
do you guys read the docs before you merge them? you're repeating the same stuff twice, read the paragraphs:

```
As an administrator, you can configure OAuth using the master configuration file to specify an identity provider. If you installed OpenShift using the Advanced Installation method, the Allow All identity provider is used by default, which allows access for all user names and passwords.

When running a master without a configuration file, the Allow All identity provider is used by default, which allows any non-empty user name and password to log in. This is useful for testing purposes. To use other identity providers, or to modify any token, grant, or session options, you must run the master from a configuration file.
```

tpoitras:
@PI-Victor this is because the text has been conditionalized between Origin and Enterprise, since Deny All is used or OSE. The Enterprise docs read as follows:

> As an administrator, you can configure OAuth using the master configuration file to specify an identity provider. If you installed OpenShift using the Quick Installation or Advanced Installation method, the Deny All identity provider is used by default, which denies access for all user names and passwords. To allow access, you must choose a different identity provider and configure the master configuration file appropriately (located at /etc/origin/master/master-config.yaml by default).
> 
> When running a master without a configuration file, the Allow All identity provider is used by default, which allows any non-empty user name and password to log in. This is useful for testing purposes. To use other identity providers, or to modify any token, grant, or session options, you must run the master from a configuration file.

I agree that this could be improved for Origin to sound less redundant.

PR #1607, opened 2016-02-17
[WIP] Add global proxy config for advanced install

sdodson:
Add advanced install configuration variables.


sdodson:
@smunilla global proxy config work here.

PR #1609, opened 2016-02-17
Added troubleshooting and e2e sections for persistent storage with stubbed content

erinboyd:


erinboyd:
@adellape @screeley44 @jeffvance Please view

screeley44:
I guess I'm still wondering if this best fits in the /install_config/...  section of the docs?  Would this be better in a new section on the same level as "install_config"?  the install_config section is pretty specific to covering installation and is really geared to the admin/cluster-deployer type audience IMO.

And the e2e and troubleshooting seem more geared to a wider range of users - including devs, etc...

thoughts?  @erinboyd @adellape @jeffvance

jeffvance:
1) is the plan to have a link in each persistent storage doc (eg, nfs, glusterfs, ceph...) pointing to the e2e doc?
2) "e2e" is often associated with testing. Perhaps an altername name of "storage-examples" is clearer?
3) referring to Scott's comment above, have you considered moving up the storage e2e to, perhaps, openshift-docs/examples/storage/ ?

erinboyd:
@jeffvance Actually the other way around. The e2e will pull the information from the other docs as those will be the 'root' documents more likely changed. Thus changes to those will be reflected in these.

There is a directory called examples/storage but it doesn't actually exist in the official doc. Just in the git repo (is my understanding)

erinboyd:
@screeley44 the reason behind the install_config was most of what has to happen in the e2e starts with the admin. Therefore it seemed like a more natural progression for them to look there. I am open to suggestions, it's really up to @adellape

screeley44:
I tend to agree with @jeffvance - maybe we need a main section called "Examples" and then another separate section for "troubleshooting" and from there, we could index and link to (or pull from) origin/examples (where most people put stuff, like our gluster and ceph examples) and also link to it from other sections like "install_config", "dev guide" "using images", etc...  - notice the using images already has a DB category under it, with some linking back to the origin/examples.  But yeah, agree, whatever @adellape thinks is best.

erinboyd:
@adellape can you please weigh in

adellape:
Sorry for the delay, all.

> 2) "e2e" is often associated with testing. Perhaps an altername name of "storage-examples" is clearer?

While looking at a preview build, I think now too it would be clearer to just call it "Persistent Storage Examples". Are there any cons to that title that come to mind?

> And the e2e and troubleshooting seem more geared to a wider range of users - including devs, etc...

While I agree (i.e., they'd likely be a mix of things that devs can do but also some things that only cluster admins can do), it seems safer to still frame it for mostly an admin audience. For example, the book could easily show up in the Enterprise and Origin doc sets, but for Dedicated and Online (where we don't provide docs for full cluster admin tasks), it's tougher. Showing a guide for those doc sets that mixes admin and dev tasks (even if it's interesting to see what the full process looks like) could be a bad experience for a dev-only audience.

> the reason behind the install_config was most of what has to happen in the e2e starts with the admin. Therefore it seemed like a more natural progression for them to look there.

I think it still makes sense either in the install_config book, or possibly the admin_guide (the Cluster Administration book), for the above reason. A boon for it being in install_config is being able to directly follow the "Configuring Persistent Storage" directory.

Side note: WRT pulling the examples up a level in general, this would also yield a new book on the Customer Portal build of these docs, which would require some legwork to set up there.

2nd side note: Eventually we do want to have a top-level "Troubleshooting Guide", where our hodgepodge of troubleshooting topics we currently have could get wrangled up (and added to) as its own book.

erinboyd:
@screeley44 @jeffvance @copejon @adellape please review

_build_cfg.yml
```diff
@@ -218,6 +218,15 @@ Topics:
...
       - Name: Volume Security
         File: pod_security_context
+      - Name: Persistent Storage Troubleshooting
+        File: storage_troubleshooting
+  - Name: Persistent Storage Examples
+    Dir: storage_examples
+    Topics:
+      - Name: Overview
+        File: index
+      - Name: Sharing an NFS mount across two persistent volume claims
```
adellape: Use title case:

Sharing an NFS Mount Across Two Persistent Volume Claims

However, I'd suggest a modified, shorter name here for the left-nav name, maybe:

Sharing an NFS Mount Across PVCs

But let the topic title in the shared_storage.adoc below be the full "Sharing an NFS Mount Across Two Persistent Volume Claims".

_build_cfg.yml
```diff
@@ -218,6 +218,15 @@ Topics:
...
+        File: shared_nfs_e2e
```
adellape: I believe the Travis build failed because this was not updated yet to the new `shared_storage` file name.

install_config/storage_examples/shared_storage.adoc
```diff
@@ -0,0 +1,16 @@
+= Sharing an NFS mount across two persistent volume claims
```
adellape: See above comment about title case.

install_config/persistent_storage/storage_troubleshooting.adoc
```diff
@@ -0,0 +1,33 @@
...
+:toc: macro
+:toc-title:
+:prewrap!:
+
+toc::[]
+
+== Overview
+This guide provides methods for resolving common issues encountered when configuring your cluster to use persistent storage.
+
+The initial setup of each persistent volume type can be found in the following guide for https://docs.openshift.org/latest/install_config/persistent_storage/index.html[Persistent Volumes].
```
adellape:
```
can be found in link:../../install_config/persistent_storage/index.html[Configuring Persistent Storage].
```

install_config/persistent_storage/storage_troubleshooting.adoc
```diff
@@ -0,0 +1,33 @@
...
+
+Detailed examples of various use cases can also be found here:
```
adellape:
```
can also be found in link:../../install_config/storage_examples/index.html[Persistent Storage Examples].
```

adellape:
@erinboyd Some comments inline.

BTW remind me, were you wanting to merge this sooner than later so others would have someplace to more to, or did you want to add any more at this time to the stub articles? If the former, we should probably make the stubs more obvious that they're WIP and have more coming soon. If the latter, just ping me again when ready. :space_invader:

screeley44:
@adellape @erinboyd  - We want to merge sooner than later, so agree we should have some WIP label set

adellape:
OK. Then I'd say, in this PR, any spot where it's currently left sorta hanging / incomplete, just replace with or add something to the effect of "More to be added soon." Only because, once this is merged onto master, it will appear in the nightly Origin build on docs.openshift.org/latest.

erinboyd:
@adellape @jeffvance @screeley44 please review

adellape:
LGTM

jeffvance:
Under install_config/storage_examples/  is _shared_storage.adoc_, which will show a complete nfs example. It might be more flexible and expandable to change this structure to:
   `install_config/storage_examples/nfs.adoc`
   `install_config/storage_examples/glusterfs.adoc`
   `install_config/storage_examples/ceph-rbd.adoc`
etc...
The idea being to easily add new storage examples. To me, the doc name, _shared_storage.adoc_, is generic. But, maybe the idea is to add *all* of our shared storage examples to this doc, and then create a _block_storage.adoc_ for all of the block storage examples?

erinboyd:
@adellape Can we merge this now so we can gets some examples going in here please

adellape:
@erinboyd @jeffvance @screeley44 This is now merged.

PR #1627, opened 2016-02-19
document quickstart examples

bparees:


bparees:
@smarterclayton ptal.  I think you asked for this a while ago.

bparees:
@mfojtik ptal.

adellape:
@bparees Is going back to calling them QuickStarts instead of InstantApps (or InstantApp Templates or something) intentional?

luciddreamz:
@bparees @adellape With OpenShift 2, we had a concrete concept of [QuickStarts](https://hub.openshift.com/api/v1/quickstarts.json) (see also [OSE v2 Docs](https://access.redhat.com/documentation/en-US/OpenShift_Enterprise/2/html/Administration_Guide/Adding_QuickStarts_to_the_Management_Console.html)). In OpenShift 3, these are actually just predefined [Templates](https://docs.openshift.com/enterprise/latest/dev_guide/templates.html). Wherever you use the word **QuickStart** or **InstantApp** in OpenShift 3, make sure to follow it with the word **Templates** so that users know what we're talking about.

bparees:
@adellape @luciddreamz it's an attempt to correct a terminology error we've made with v3.

quickstart:  provides a basic skeleton for developing an application on a particular language/framework (eg gives you a Rails hello world app and defines a DB service).  Assumption is you will then edit the code to do something interesting and rebuild/redeploy.

instantapp: provides a useful application out of the box (eg Jenkins or Wordpress), no editing/coding needed.  It may build itself from a source repo, but there's no assumption the user needs to fork the repo and edit the source code to make it useful/interesting.


A template could define a QuickStart or an InstantApp or neither.

An instantapp could technically be defined w/o a template (eg "oc new-app openshift/jenkins-1-centos7" will get you a running Jenkins, no template needed, though it's not ideal since there's no persistent storage in that case)

Sound reasonable?

adellape:
@bparees Are the same distinctions being made in the template tags / UI? Should most (all?) of what we currently call InstantApp templates be reclassified as QuickStart templates?

bparees:
@adellape they should be, they are not yet.  i've been reluctant to bring it up w/ @jwforres because i think we already went through this once when we picked instantapp as the tag :)

luciddreamz:
@bparees So, @corey112358  and I tried to move away from providing "example" code in the OpenShift v2 QuickStarts and instead focused on configuring languages/frameworks so that they were ready to use on OpenShift. Yes they would deploy to the platform in one click with a database, and default page would show at the index route if you navigated there in browser, but more than anything the QuickStart served as a template that the end-user could pickup and run with. Core configuration options for the framework were integrated to work with the corresponding concepts/features in OpenShift ([persistent storage and logging](https://github.com/luciddreamz/laravel/blob/master/.openshift/action_hooks/deploy#L65-L80) for example, with [docs](https://github.com/luciddreamz/laravel#log-files)).

Even [WordPress](https://github.com/openshift/wordpress-example) has been heavily updated to integrate core WordPress features with core OpenShift features (for example using the PHP cart's development mode variable to also set [WordPress into development mode](https://github.com/openshift/wordpress-example#development-mode)). I guess my preference is to make the assumption that users will always want to modify the code and/or should at least look at the relevant GitHub readme.

What I would like to do with OpenShift 3, if possible, is try to standardize the naming around predefined Templates where we would have something like this:

- Application Templates (tag: application)
- Database Templates (tag: database)
- xPaaS Templates (tag: xpaas)

luciddreamz:
I guess to make this more clear, I want to get rid of the distinction between InstantApp Templates and QuickStart Templates and just call them both "Application Templates" @bparees @corey112358.

bparees:
@luciddreamz ok i removed the db templates entirely since i realized we already talk about them here:
for example here:
https://docs.openshift.org/latest/using_images/db_images/mysql.html#creating-a-database-service-from-a-template

bparees:
@luciddreamz thoughts on the current iteration?  it explicitly calls everything a template now (sometimes prefixed with quickstart)

dev_guide/app_tutorials/quickstarts.adoc
```diff
@@ -0,0 +1,27 @@
...
+* Dancer - a Perl web framework (includes a MySQL database)
+** link:https://github.com/openshift/origin/tree/master/examples/Quickstarts/dancer-mysql.json[Template definition]
+** link:https://github.com/openshift/dancer-ex[Source repository]
+* Django - a Python web framework (includes a PostgreSQL database)
+** link:https://github.com/openshift/origin/tree/master/examples/Quickstarts/django-postgresql.json[Template definition]
+** link:https://github.com/openshift/django-ex[Source repository]
+* NodeJS - a NodeJS web application (includes a MongoDB database)
+** link:https://github.com/openshift/origin/tree/master/examples/Quickstarts/nodejs-mongodb.json[Template definition]
+** link:https://github.com/openshift/nodejs-ex[Source repository]
+* Rails - a Ruby web framework (includes a PostgreSQL database)
```
adellape: Should we provide a link here to the `ruby_on_rails` tutorial as well?

bparees: you mean this:
https://docs.openshift.org/latest/dev_guide/app_tutorials/ruby_on_rails.html
?

i'd say no, given that that's another topic under the same section as this topic already.

luciddreamz:
@bparees @adellape Note: In the console, the word used is "Instant Apps"

bparees:
@luciddreamz this pull doesn't mention instant apps.  anything other concerns?

gabemontero:
fyi - some refs to the quickstarts and instant apps were recently added to the getting started section as part of the online work (developers_cli.adoc).

We'll very well may need to massage those refs slightly to be in sync with, and to have ptrs to when appropriate, the content being worked here.  I can certainly drive that after this PR lands if needed.

bparees:
@adellape ptal

bparees:
@adellape or rather, please merge :)

dev_guide/app_tutorials/quickstarts.adoc
```diff
@@ -0,0 +1,27 @@
...
+
+Many quickstarts include a reference to a source repository that contains the application source code.  To customize the quickstart, you will need to fork the repository and then, when instantiating the template, substitute the default source repository name with your forked repository.  This will result in builds that are performed using your source code instead of the provided example source.  You can then update the code in your source repository and launch a new build to see the changes reflected in the deployed application.
+
+== Web Framework Quickstart Templates
+
+These quickstarts provide a minimum application implemented in the framework and language indicated.
+
+* CakePHP - a PHP web framework (includes a MySQL database)
+** link:https://github.com/openshift/origin/tree/master/examples/Quickstarts/cakephp-mysql.json[Template definition]
```
adellape: Looks like all the links to the json files have `Quickstarts` in the path but should be lowercase `quickstarts`.

adellape:
@bparees After the links are fixed, I'll merge for Origin's sake, and label this for follow-up before getting downstream.

bparees:
@adellape thanks, fixed the capitalization issue.

bparees:
@adellape bump

PR #1635, opened 2016-02-22
Moved the docker port from port 9 to 3.  Removed openshift from slides.

knobunc:
Reflected the changes in the code and relocated the docker bridge to port 3.

Changed the slides so the text is generic across Origin and Atomic OpenShift.


knobunc:
@adellape Please can you review this?

PR #1636, opened 2016-02-22
Redeploy step not necessary since it is triggered on dc config change

aweiteka:


aweiteka:
@adellape this is a minor thing but it does prevent an extra deployment of the reg.

adellape:
Sorry for the delay. LGTM, though maybe in its place we could say as much about the dc change trigger happening at that point.

adellape:
No revhistory needed.

PR #1638, opened 2016-02-22
BZ 1299000 -- Document uploading images to the internal registry

tnguyen-rh:
This is for [BZ 1299000](https://bugzilla.redhat.com/show_bug.cgi?id=1299000).
It adds an xref from the Creating Images "Overview" to Install/Config "Docker Registry", highlights the term `pull specification` in Architecture, emphasizes the `<project>` component of the pull specification in the `docker tag` command, and adds a section on the registry's IP address (pulling from https://github.com/openshift/openshift-docs/issues/1494).

### DIFF
--- a/creating_images/index.adoc
+++ b/creating_images/index.adoc
@@ -6,3 +6,7 @@
 :experimental:
 
 This guide provides best practices on writing and testing Docker images that can be used on OpenShift.
+ifdef::openshift-enterprise,openshift-origin[]
+Once you have created an image, you can push it to
+the link:../install_config/install/docker_registry.html[internal registry].
+endif::[]
--- a/install_config/install/docker_registry.adoc
+++ b/install_config/install/docker_registry.adoc
@@ -101,6 +101,7 @@ $ oc get -o yaml svc docker-registry | \
       oc replace -f -
 ----
 
+[[non-production-use]]
 ==== Non-Production use
 
 For non-production use, you can use the `--mount-host=<path>` option to specify
@@ -169,7 +170,6 @@ $ oadm registry --service-account=registry \
 ----
 endif::[]
 
-[[viewing-logs]]
 
 ==== Setting storage quota
 To protect the storage from being accidentally filled up or to prevent users from abusing the space,
@@ -193,6 +193,80 @@ You may then
 link:../../admin_guide/pruning_resources.html#pruning-images[prune obsoleted images]
 to retrieve some space back.
 
+
+[[maintaining-the-registry-ip-address]]
+=== Maintaining the Registry IP Address
+
+OpenShift refers to the integrated registry by its service IP address,
+so if you decide to delete and recreate the *docker-registry* service,
+you can ensure a completely transparent transition by arranging to
+re-use the old IP address in the new service.
+If a new IP address cannot be avoided, you can minimize cluster
+disruption by rebooting only the masters.
+
+[[re-using-the-address]]
+Re-using the Address::
+
+To re-use the IP address, you must save the IP address of the old *docker-registry*
+service prior to deleting it, and arrange to replace the newly assigned IP address
+with the saved one in the new *docker-registry* service.
+
+// NB: Snarfed from <https://github.com/openshift/openshift-docs/issues/1494>.
+. Make a note of the `ClusterIP` for the service:
++
+----
+$ oc get svc/docker-registry -o yaml | grep clusterIP:
+----
+
+. Delete the service:
++
+----
+$ oc delete svc/docker-registry dc/docker-registry
+----
+
+. Create the registry definition in *_registry.yaml_*,
+replacing `<options>` with, for example, those used in step 3 of the
+instructions in the link:#non-production-use[Non-Production use] section:
++
+----
+$ oadm registry <options> -o yaml > registry.yaml
+----
+
+. Edit *_registry.yaml_*, find the `Service` there,
+and change its `ClusterIP` to the address noted in step 1.
+
+. Create the registry using the modified *_registry.yaml_*:
++
+----
+$ oc create -f registry.yaml
+----
+
+[[rebooting-the-masters]]
+Rebooting the Masters::
+
+If you are unable to re-use the IP address, any operation that uses a link:../../architecture/core_concepts/builds_and_image_streams.html#referencing-images-in-image-streams[pull specification]
+that includes the old IP address will fail.
+To minimize cluster disruption, you must reboot the masters:
++
+----
+ifdef::openshift-origin[]
+# systemctl restart origin-master
+endif::[]
+ifdef::openshift-enterprise[]
+# systemctl restart atomic-openshift-master
+endif::[]
+----
+// Code block snarfed from ../http_proxies.adoc, w/ node-reboot stuff removed.
+// tnguyen opines: It would be nice to #define this somewhere and include it here...
++
+This ensures that the old registry URL, which includes the old IP address,
+is cleared from the cache.
++
+[NOTE]
+We recommend against rebooting the entire cluster because that incurs
+unnecessary downtime for pods and does not actually clear the cache.
+
+[[viewing-logs]]
 == Viewing Logs
 
 To view the logs for the Docker registry, run the `oc logs` indicating the desired pod:
@@ -409,6 +483,9 @@ In the following examples, we use:
 |*<image>*
 |`busybox`
 
+|*<tag>*
+| omitted (defaults to `latest`)
+
 |====
 
 . Pull an arbitrary image:
@@ -419,7 +496,11 @@ $ docker pull docker.io/busybox
 ----
 ====
 
-. Tag the new image with the form `<registry_ip>:<port>/<project>/<image>`:
+. Tag the new image with the form `<registry_ip>:<port>/<project>/<image>`.
+The project name *must* appear in this
+link:../../architecture/core_concepts/builds_and_image_streams.html#referencing-images-in-image-streams[pull specification]
+for OpenShift to
+correctly place and later access the image in the registry.
 +
 ====
 ----
--- a/architecture/core_concepts/builds_and_image_streams.adoc
+++ b/architecture/core_concepts/builds_and_image_streams.adoc
@@ -317,7 +317,7 @@ image stream and image name. It uses the following convention for its name:
 `<image stream name>@<name>`.
 
 A `*DockerImage*` is used to reference or retrieve an image for a given
-external registry. It uses standard docker pull specification for its name,
+external registry. It uses standard docker _pull specification_ for its name,
 eg `openshift/ruby-20-centos7:2.0`. When no tag is specified it is assumed
 the `latest` will be used.
 


tnguyen-rh:
@pweil- @bfallonf 
PTAL

install_config/install/docker_registry.adoc
```diff
@@ -169,9 +169,41 @@ $ oadm registry --service-account=registry \
...
+[[registry-ip-address]]
+=== Registry IP Address
+
+Because the integrated registry is referenced by its service IP address,
+if you decide to delete and recreate the docker-registry service,
+you should either take care to re-use its address,
+or reboot the cluster after creating the new docker-registry service.
+
+[[re-using-the-address]]
+==== Re-using the Address
```
bfallonf: @tnguyen-rh  Did we ever talk about never going more than 3 ==='s down? There's a chance I made that rule up in my head...

I feel you could have the same effect here by calling "Re-using the Address" and "Rebooting the Cluster" into " *Option 1: Re-using the Address* " and the like.

install_config/install/docker_registry.adoc
```diff
@@ -169,9 +169,41 @@ $ oadm registry --service-account=registry \
...
+service prior to deleting it, and arrange to replace the newly assigned IP address
+with the saved one in the new docker-registry service.
+
+// NB: Snarfed from <https://github.com/openshift/openshift-docs/issues/1494>.
+. Make a note of the `ClusterIP` for the service.
+. `oc delete svc/docker-registry dc/docker-registry`
+. `oadm registry ... -o yaml > registry.yaml`
+. Edit *_registry.yaml_*, find the `Service` there,
+and set its `ClusterIP` to the address noted in step 1.
+. `oc create -f registry.yaml`
```
bfallonf: The above doesn't seem to mark up correctly. Too many periods before each command?

install_config/install/docker_registry.adoc
```diff
@@ -169,9 +169,41 @@ $ oadm registry --service-account=registry \
...
+you must reboot the cluster after creating the new docker-registry service.
+In some circumstances it is sufficient to reboot only the master,
+but that may leave stale references to the old IP address in
+link:../../architecture/infrastructure_components/kubernetes_infrastructure.html#master[etcd].
+
+
+[[viewing-logs]]
 == Viewing Logs
 
 To view the logs for the Docker registry, run the `oc logs` indicating the desired pod:
```
bfallonf: I know this was here before this PR, but "run the `oc logs`" might need a "command" thrown in there, or a deleting of "the"

"run `oc logs`"
"run the `oc logs` command"

bfallonf:
@tnguyen-rh A few pokes :point_right:

install_config/install/docker_registry.adoc
tnguyen-rh: Yeah, i had a feeling this was too deep (and the volume in each sub-sub-sub section too little to warrant it).  Will rework to use a labelled list.

install_config/install/docker_registry.adoc
tnguyen-rh: Good idea.  Will do.

install_config/install/docker_registry.adoc
tnguyen-rh: I suppose even better would be to omit the inline markup.  But, as you note, this is out of scope for this PR.  @vikram-redhat WDYT?

kargakis: We know support getting application logs just by running oc logs against the dc

kargakis: So once the docker registry is deployed, `oc logs dc/docker-registry` should be the same as `oc logs docker-registry-1-da73t`.

install_config/install/docker_registry.adoc
```diff
@@ -169,9 +169,41 @@ $ oadm registry --service-account=registry \
...
 
-[[viewing-logs]]
 
+[[registry-ip-address]]
+=== Registry IP Address
+
+Because the integrated registry is referenced by its service IP address,
+if you decide to delete and recreate the docker-registry service,
+you should either take care to re-use its address,
+or reboot the cluster after creating the new docker-registry service.
```
ncdc: As mentioned in the bz, the only thing you need to restart is the master api. Restarting anything else will cause unnecessary downtime.

install_config/install/docker_registry.adoc
```diff
@@ -169,9 +169,41 @@ $ oadm registry --service-account=registry \
...
+. `oc create -f registry.yaml`
 
+[[rebooting-the-cluster]]
+Rebooting the Cluster::
+
+If you are unable to re-use the IP address,
+you must reboot the cluster after creating the new docker-registry service.
+In some circumstances it is sufficient to reboot only the master,
+but that may leave stale references to the old IP address in
+link:../../architecture/infrastructure_components/kubernetes_infrastructure.html#master[etcd].
```
ncdc: Same comment as above. Only restart the master api.

install_config/install/docker_registry.adoc
```diff
@@ -398,7 +430,11 @@ $ docker pull docker.io/busybox
 ----
 ====
 
-. Tag the new image with the form `<registry_ip>:<port>/<project>/<image>`:
+. Tag the new image with the form `<registry_ip>:<port>/<project>/<image>`.
```
ncdc: You can also append a tag after `<image>`. If you omit it, it defaults to latest.

tnguyen-rh: Right.  I'll mention that in the intro blurb.

install_config/install/docker_registry.adoc
tnguyen-rh: Thanks for the tip.  I think this definitely warrants another issue:
https://github.com/openshift/openshift-docs/issues/1645

tnguyen-rh:
@bfallonf 
Thanks for the review.  With the exception of the `oc log`-related change (which has its [own issue](#1645) now), i've installed the suggested changes.
@ncdc 
I've bifurcated the "reboot cluster" section to describe "only master" and "entire cluster" scenarios.  The reason i want to keep "entire cluster" is abstracted to *NORMALLY-HARMLESS* and *MILDLY-UNDESIRABLE-CONSEQUENCE*.  Could you suggest some text for those placeholders?

install_config/install/docker_registry.adoc
```diff
@@ -169,9 +170,68 @@ $ oadm registry --service-account=registry \
...
 
-[[viewing-logs]]
 
+[[registry-ip-address]]
+=== Registry IP Address
+
+Because the integrated registry is referenced by its service IP address,
+if you decide to delete and recreate the docker-registry service,
+you should either take care to re-use its address,
+or reboot some or all of the cluster after creating the new docker-registry service.
```
ncdc: Please revise this to only mention restarting the masters.

install_config/install/docker_registry.adoc
```diff
@@ -169,9 +170,68 @@ $ oadm registry --service-account=registry \
...
++
+----
+oc create -f registry.yaml
+----
+
+[[rebooting-the-cluster]]
+Rebooting the Cluster::
+
+If you are unable to re-use the IP address,
+you must reboot some or all of the cluster after creating the new docker-registry service.
```
ncdc: Please revise this to only mention restarting the masters.

install_config/install/docker_registry.adoc
```diff
@@ -169,9 +170,68 @@ $ oadm registry --service-account=registry \
...
+is cleared from the cache.
++
+[horizontal]
+Only master:::
+Rebooting only the master avoids unnecessary downtime.
+However, it may leave stale references to the old IP address in
+link:../../architecture/infrastructure_components/kubernetes_infrastructure.html#master[etcd].
+This is *NORMALLY-HARMLESS* but *MILDLY-UNDESIRABLE-CONSEQUENCE*.
+
+Entire cluster:::
```
ncdc: Please remove this section

install_config/install/docker_registry.adoc
```diff
@@ -169,9 +170,68 @@ $ oadm registry --service-account=registry \
...
+
+If you are unable to re-use the IP address,
+you must reboot some or all of the cluster after creating the new docker-registry service.
+This ensures that the old registry URL, which includes the old IP address,
+is cleared from the cache.
++
+[horizontal]
+Only master:::
+Rebooting only the master avoids unnecessary downtime.
+However, it may leave stale references to the old IP address in
```
ncdc: There is no way around stale IP addresses if the docker-registry service's IP changes. You either have to preserve the original IP, or write some utility/script to traverse etcd and find-and-replace the old IP with the new IP.

ncdc:
@tnguyen-rh 

> I've bifurcated the "reboot cluster" section to describe "only master" and "entire cluster" scenarios. The reason i want to keep "entire cluster" is abstracted to NORMALLY-HARMLESS and MILDLY-UNDESIRABLE-CONSEQUENCE. Could you suggest some text for those placeholders?

I've added comments to individual lines. In summary, please remove any reference to rebooting the entire cluster.

cc @pweil- @miminar

install_config/install/docker_registry.adoc
tnguyen-rh: (from IRC)
> [R]ebooting the cluster does not remove stale registry IPs.
Thanks for the clarification.  This was the key misunderstanding.

install_config/install/docker_registry.adoc
tnguyen-rh: (from IRC in response to followup question "what is the consequence of stale IP addrs?")

> possibility of failed image pulls (leading to failed container/pod starts) and pushes

and also

> if you restart the master, so it picks up the new ip, new builds & deployments will start using the new ip correctly

Cool, thanks.  Will now fill in the placeholders appropriately.  (The reason for all this fuss is to give the reader a bit of "why" behind the "do this" instructions, part of the effort to preempt some support calls.)

creating_images/index.adoc
```diff
@@ -6,3 +6,7 @@
 :experimental:
 
 This guide provides best practices on writing and testing Docker images that can be used on OpenShift.
+ifdef::openshift-enterprise,openshift-origin[]
```
pweil-: why is this scoped to just enterprise and origin?

pweil-:
+1 to just restarting master to pick up the new IP if it was unable to be reused.  Thanks for putting this together.  Will take another look when the new changes land.

install_config/install/docker_registry.adoc
bfallonf: Looks good!

creating_images/index.adoc
tnguyen-rh: @pweil-
The conditional is around a link to the Installation and Configuration document, which is OSO and OSE only, IIUC its [`_build_cfg.yml` Distros key](https://github.com/openshift/openshift-docs/blob/master/_build_cfg.yml#L152) correctly.
Am i missing something?

install_config/install/docker_registry.adoc
```diff
@@ -169,9 +170,68 @@ $ oadm registry --service-account=registry \
...
+Re-using the Address::
+
+To re-use the IP address, you must save the IP address of the old docker-registry
+service prior to deleting it, and arrange to replace the newly assigned IP address
+with the saved one in the new docker-registry service.
+
+// NB: Snarfed from <https://github.com/openshift/openshift-docs/issues/1494>.
+. Make a note of the `ClusterIP` for the service.
+
+. Delete the service.
```
miminar: Save the address first with something like:

    oc describe svc/docker-registry | sed -n 's/^IP:\s*\(.*\)/\1/p'
    # or
    oc get svc/docker-registry -o yaml | grep 'clusterIP:'

?

tnguyen-rh: @miminar 
Thanks for the tip.  I think the second command is easier to understand on quick reading, and have added it.

creating_images/index.adoc
pweil-: Probably not, I was just trying to think of a reason that this link wouldn't be shown.  If install is keyed to those distros then this is fine.

tnguyen-rh:
@pweil- @ncdc 
PTAL.

pweil-:
Changes LGTM

tnguyen-rh:
@ahardin-rh @adellape @bfallonf 
PTAL (2nd time for @bfallonf  due to big changes).
In particular, i would like to confirm that the labels are correct.

install_config/install/docker_registry.adoc
```diff
@@ -169,9 +170,68 @@ $ oadm registry --service-account=registry \
 ----
 endif::[]
 
-[[viewing-logs]]
 
+[[registry-ip-address]]
+=== Registry IP Address
```
bfallonf: Can this have a more descriptive title? Something like "Defining the Registry IP Address" maybe? Something that describes the process the reader is about to undertake.

install_config/install/docker_registry.adoc
```diff
@@ -169,9 +170,68 @@ $ oadm registry --service-account=registry \
...
+
+OpenShift refers to the integrated registry by its service IP address,
+so if you decide to delete and recreate the docker-registry service,
```
bfallonf: I think the markup for services is iffy in the rest of the docs, but should it have some *'s around docker-registry?

install_config/install/docker_registry.adoc
```diff
@@ -169,9 +170,68 @@ $ oadm registry --service-account=registry \
...
+you can ensure a completely transparent transition by arranging to
+re-use the old IP address in the new service.
+If a new IP address cannot be avoided, you can minimize cluster
+disruption by rebooting only the masters.
+
+[[re-using-the-address]]
+Re-using the Address::
+
+To re-use the IP address, you must save the IP address of the old docker-registry
```
bfallonf: *'s here as well?

install_config/install/docker_registry.adoc
```diff
@@ -169,9 +170,68 @@ $ oadm registry --service-account=registry \
...
+service prior to deleting it, and arrange to replace the newly assigned IP address
+with the saved one in the new docker-registry service.
+
+// NB: Snarfed from <https://github.com/openshift/openshift-docs/issues/1494>.
+. Make a note of the `ClusterIP` for the service.
```
bfallonf: Some :'s after the step descriptions leading into the command:

"Make a note of the `ClusterIP` for the service:"

install_config/install/docker_registry.adoc
```diff
@@ -169,9 +170,68 @@ $ oadm registry --service-account=registry \
...
+----
+$ oc create -f registry.yaml
+----
+
+[[rebooting-the-masters]]
+Rebooting the Masters::
+
+If you are unable to re-use the IP address, any operation that uses a link:../../architecture/core_concepts/builds_and_image_streams.html#referencing-images-in-image-streams[pull specification]
+that includes the old IP address will fail.
+To minimize cluster disruption, you must reboot the masters.
```
bfallonf: Reboot command example?

----
# systemctl enable atomic-openshift-master
----

install_config/install/docker_registry.adoc
bfallonf: @tnguyen-rh I think maybe I was misunderstood. This is a typo-level suggestion :P There's a word missing and I thought it'd be easier to make the change while you're in here already.

install_config/install/docker_registry.adoc
```diff
@@ -390,6 +450,8 @@ In the following examples, we use:
 
 |====
 
+We omit the optional trailing `:<tag>`, so `<tag>` defaults to `latest`.
```
bfallonf: I'm not sure what this means. Who is we? Where does the <tag> go?

bfallonf:
@tnguyen-rh A few comments, but cool :cool:

install_config/install/docker_registry.adoc
adellape: I agree w/ the asterisk suggestion, at least as far as matching what we (mostly) use elsewhere for "unique names of objects".

Side note: I wish we used backticks more liberally in our markup guidelines, like for values of parameters/settings, and possibly for general "names of objects" like these. :disappointed:

install_config/install/docker_registry.adoc
tnguyen-rh: @bfallonf 
> Can this have a more descriptive title? Something like "Defining the Registry IP Address" maybe? Something that describes the process the reader is about to undertake.
IIUC the procedure is only necessary if you want to remove and then re-add the `docker-registry` service.  The purpose of the procedure is to maintain the IP address, so perhaps a better title might be "Maintaining the Registry IP Address".

tnguyen-rh:
@bfallonf @adellape 
Thanks for the review.  I've munged things according to your suggestions.  WDYT?

install_config/install/docker_registry.adoc
bfallonf: @tnguyen-rh I like it!

install_config/install/docker_registry.adoc
bfallonf: +1!

bfallonf:
@tnguyen-rh Works for me :clap:

tnguyen-rh:
**revhistory note:**

(Install & Config -> Install -> Deploying a Docker Registry)

- Mentioned default tag `latest`.
- Clarified importance of *<project-name>* in pull spec.
- Added section **Maintaining the Registry IP Address**.

PR #1564, opened 2016-02-09
Update persistent_storage_nfs.adoc

jeffvance:
I've tried to cover several security topics under nfs persistent storage. This is just a first stab at attempting to improve the documentation around security and storage. It is likely that much of the nfs security will be moved to other sec docs or to a new sec doc, so that these practices do not need to be repeated for each storage type. However, for now, it seem prudent to make this update available and restructure the best fit for storage & security a bit later.

PR #1649, opened 2016-02-23
storage security update

jeffvance:
This is the replacement PR for #1564

### DIFF
--- a/install_config/persistent_storage/persistent_storage_nfs.adoc
+++ b/install_config/persistent_storage/persistent_storage_nfs.adoc
@@ -11,122 +11,364 @@
 toc::[]
 
 == Overview
-You can provision your OpenShift cluster with
-link:../../architecture/additional_concepts/storage.html[persistent storage] using
-https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Storage_Administration_Guide/ch-nfs.html[NFS].
-Some familiarity with Kubernetes and NFS is assumed.
 
-The Kubernetes
-link:../../architecture/additional_concepts/storage.html[persistent volume]
-framework allows administrators to provision a cluster with persistent storage
-and gives users a way to request those resources without having any knowledge of
-the underlying infrastructure. Persistent volumes are not bound to a single
-project or namespace; they can be shared across the OpenShift cluster.
-link:../../architecture/additional_concepts/storage.html#persistent-volume-claims[Persistent
-volume claims], however, are specific to a project or namespace and can be
-requested by users.
+OpenShift clusters can be provisioned with
+link:../../architecture/additional_concepts/storage.html[persistent storage] using NFS
+Some familiarity with Openshift,
+link:../../architecture/additional_concepts/storage.html#persistent-volume-claims[persistent volume claims (PVCs)],
+and https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Storage_Administration_Guide/ch-nfs.html[NFS] is beneficial.
 
-For a detailed example, see the guide for
-https://github.com/openshift/origin/tree/master/examples/wordpress[WordPress and
-MySQL using NFS].
+The Openshift persistent volume (PV) framework allows administrators to provision a cluster
+with persistent storage and gives developers the ability to request those resources without
+having specific knowledge of the underlying infrastructure. Persistent volumes are not bound
+to a single project; they can be shared across the entire OpenShift cluster. Persistent volume
+claims (PVCs), however, are specific to a project and are created and used by developers.
 
-[IMPORTANT]
-====
-High-availability of storage in the infrastructure is left to the underlying
-storage provider.
-====
-
-[[nfs-provisioning]]
+(Note: "project" and "namespace" are used interchangeably throughout this document)
 
 == Provisioning
-Storage must exist in the underlying infrastructure before it can be mounted as
-a volume in OpenShift. To provision NFS volumes in OpenShift, all that is
-required is a distinct list of NFS servers and paths and the
-`*PersistentVolume*` API.
 
-You must define your persistent volume in an object definition before creating
-it in OpenShift:
+Storage must exist in the underlying infrastructure before it can be mounted as a volume in
+OpenShift. To provision NFS volumes a list of NFS servers and export paths are all that is
+required.
 
-.Persistent Volume Object Definition Using NFS
-====
-
-[source,yaml]
-----
-apiVersion: "v1"
-kind: "PersistentVolume"
+A PV spec file is usually defined before creating the PV:
+```
+apiVersion: v1
+kind: PersistentVolume
 metadata:
-  name: "pv0001" <1>
+  name: pv0001 <1>
 spec:
   capacity:
-    storage: "5Gi" <2>
+    storage: 5Gi <2>
   accessModes:
-    - "ReadWriteOnce" <3>
+  - ReadWriteOnce <3>
   nfs: <4>
-    path: "/tmp" <5>
-    server: "172.17.0.2" <6>
-  persistentVolumeReclaimPolicy: "Recycle" <7>
-----
-<1> The name of the volume. This will be how it is identified via
-link:../../architecture/additional_concepts/storage.html[persistent volume
-claims] or from pods.
-<2> The amount of storage allocated to this volume.
-<3> This is the policy for restricting read/writes to a shared volume.
-<4> This defines the volume type being used, in this case the *nfs* plug-in.
-<5> The path that is exported by the NFS server.
-<6> The host name or IP address of the NFS server.
-<7> Defines what happens to a volume when released from its claim. Valid options
+    path: /tmp <5>
+    server: 172.17.0.2 <6>
+  persistentVolumeReclaimPolicy: Recycle <7>
+```
+<1> the name of the volume. This is the PV identify in various `oc ... pod` commands.
+<2> the amount of storage allocated to this volume.
+<3> though this appears to be related to controlling access to the volume, in fact, it
+is actually used similarly to labels and used to match a PVC to a PV. Currently, no
+access rules are enforced based on the `accessModes`.
+<4> this defines the volume type being used, in this case the *nfs* plug-in.
+<5> the path that is exported by the NFS server.
+<6> the host name or IP address of the NFS server.
+<7> defines what happens to a volume when released from its claim. Valid options
 are *Retain* (default) and *Recycle*. See
 link:#nfs-reclaiming-resources[Reclaiming Resources].
-====
 
-Save your definition to a file, for example *_nfs-pv.yaml_*, and create the
-persistent volume:
-
-====
-----
+Save the definition to a file, for example *_nfs-pv.yaml_*, and create the persistent
+volume:
+```
 # oc create -f nfs-pv.yaml
 persistentvolume "pv0001" created
-----
-====
+```
 
 Verify that the persistent volume was created:
-
-====
-----
+```
 # oc get pv
 NAME                     LABELS    CAPACITY     ACCESSMODES   STATUS      CLAIM     REASON    AGE
 pv0001                   <none>    5368709120   RWO           Available                       31s
-----
-====
+```
 
-Users can then link:../../dev_guide/persistent_volumes.html[request storage
-using persistent volume claims], which can now utilize your new persistent
-volume.
+The next step can be to create a persistent volume claim (PVC) which will bind to the new PV.
+This is easy to do, as shown in the "nfs-claim.yaml example below:
+```
+#claim yaml file: "nfs-claim.yaml"
+
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: nfs-claim1
+spec:
+  accessModes:
+    - ReadWriteOnce <1>
+  resources:
+    requests:
+      storage: 1Gi <2>
+
+# oc create -f nfs-claim.yaml
+```
+<1> as mentioned above for PVs, the `accessModes` do not enforce security, but rather act as
+labels to match a PV to a PVC.
+<2> this claim will look for PVs offering 1Gi or greater capacity.
+
+*Note:* PVs, and PVCs are not necessary, just convenient, and make sharing a volume across a
+project simpler. The NFS specific information contained in the PV spec can also be defined
+directly in a pod spec.
 
 [[nfs-enforcing-disk-quotas]]
-
 === Enforcing Disk Quotas
-Use disk partitions to enforce disk quotas and size constraints. Each partition
-can be its own export. Each export is one persistent volume. Kubernetes enforces
-unique names for persistent volumes, but the uniqueness of the NFS volume's
-server and path is up to the administrator.
+Disk partitions can be used to enforce disk quotas and size constraints. Each partition
+can be its own export. Each export is one persistent volume (PV). Openshift enforces
+unique names for PVs, but the uniqueness of the NFS volume's server and path is up to
+the administrator.
 
-Enforcing quotas in this way allows the end user to request persistent storage
+Enforcing quotas in this way allows the developer to request persistent storage
 by a specific amount (e.g, 10Gi) and be matched with a corresponding volume of
 equal or greater capacity.
 
 [[nfs-volume-security]]
+== Volume Security
+This section covers aspects of Security Context Constraints (SCCs), which are described
+elsewhere, see
+link:../../admin_guide/manage_scc.html[managing SCCs] and
+link:../../architecture/additional_concepts/authorization.html#security-context-constraints[SCC concepts].
+The reader is expected to understand the basics of POSIX permissions, process UIDs,
+supplemental groups, and selinux.
 
-=== Volume Security
-Users request storage with a `*PersistentVolumeClaim*`. This claim only lives in
-the user's namespace and can only be referenced by a pod within that same
-namespace. Any attempt to access a persistent volume across a namespace causes
-the pod to fail.
+*NOTE:* The
+link:pod_security_context.html[pod and storage security] document should be read before
+implementing NFS volumes.
 
+Developers request NFS storage by referencing, in the `volumes` section of their
+pod spec, either a _PersistentVolumeClaim_ by name, or the NFS volume plug-in directly.
+
+The *_/etc/exports_* file, on the NFS server, contains the accessible NFS directories.
+The target NFS directory has the familiar POSIX owner and group ids. The Openshift NFS
+plug-in mounts the container's NFS directory with the same POSIX ownership and permissions
+found on the exported NFS directory. However, the container is not run with its effective
+UID equal to the owner of the NFS mount, which is the desired behavior.
+
+As an example, if the target NFS directory appears as:
+[[nfs-export]]
+```
+#on the nfs server:
+# ls -lZ /opt/nfs -d
+drwxrws---. nobody 5555 unconfined_u:object_r:usr_t:s0   /opt/nfs
+
+# id nobody
+uid=99(nobody) gid=99(nobody) groups=99(nobody)
+```
+then the container will need to match selinux labels, and either run with a UID of 99
+(_nobody_ owner), or with 5555 in its supplemental groups, in order to access the directory.
+
+=== User IDs
+User ids can be defined in the container image or in the pod spec. The
+link:pod_security_context.html#user-id[pod and storage security] document covers
+controlling storage access based on user ids.
+
+The container needs it's UID set to 99 (ignoring group ids for the moment), so the
+following fragement can be added to the pod spec:
+```
+spec:
+  containers: <1>
+  - name: ...
+    securityContext:
+      runAsUser: 99  #nobody
+```
+<1> pods contain a `securtityContext` specific to each container (shown above), and a global
+`securityContext` which applies to all containers defined in the pod.
+
+Assuming the _default_ project and the _restricted_ SCC, the pod's requested user id of 99
+will, unfortunetely, *not* be allowed, and therefore the pod will fail. The pod fails because:
+
+- it requests 99 as its user id,
+- all SCCs available to the pod are examined to see which SCC will allow a user id of 99
+(actually, all policies of the SCCs are checked but the  focus here is on user id),
+- since all available SCCs use _MustRunAsRange_ for their `runAsUser` strategy, uid
+range checking is required, 
+- 99 is not included in the SCC or namespace's user id range, so the pod fails.
+
+To best way to fix this situation is to create a custom SCC.
+
+===== Custom SCC for User IDs
+It's generally considered a good practice to *not* modify the predefined SCCs. The preferred
+approach is to create a custom SCC that better fits an organization's security needs, or
+create a new project that supports the desired user ids. See
+link:../../dev_guide/projects.html#create-a-project[projects] if interested in creating
+a new project.
+
+[[nfs-scc]]
+A custom SCC can be created such that a min and max user ids are defined, UID range
+checking is still enforced, and the UID of 99 will be allowed. Here is an exert:
+```
+# oc export scc nfs-scc <1>
+allowHostDirVolumePlugin: false  #the allow* bools are the same as for the restricted scc
+...
+kind: SecurityContextConstraints
+metadata:
+  ...
+  name: nfs-scc <1>
+priority: 9 <2>
+runAsUser:
+  type: MustRunAsRange <3>
+  uidRangeMax: 99 <4>
+  uidRangeMin: 99
+...
+```
+<1> the name of the new SCC is "nfs-scc".
+<2> numerically larger numbers have greater priority, nil or omitted is the lowest priority.
+Higher priority SCCs sort before lower pri SCCs and thus have a better chance of matching a new
+pod.
+<3> the `runAsUser` strategy is set to _MustRunAsRange_, which means uid range checking is 
+enforced.
+<4> the uid range is 99-99 (a range of one value). The min value, 99, would become the user id
+default if a user id was not supplied in the pod spec.
+
+Now, using `runAsUser: 99`, shown in the pod fragment above, the pod matches the "nfs-scc"
+and is able to run with a UID of 99.
+
+=== Group IDs
+Another way to handle NFS access (assuming it's not a choice to change permissions on the NFS export)
+is to use supplemental groups. Supplemental groups in Openshift are used for shared storage, of which
+NFS is an exmaple. In contrast, block storage, such as Ceph RBD or iSCSI, use the `fsGroup` SCC strategy
+and  the `fsGroup` value in the pod's `securityContext`.
+
+*NOTE:* this is covered in the
+link:pod_security_context.adoc#supplemental-groups[pod and storage security] document.
+
+Since the group id on the link:#nfs-export[target NFS directory], shown above, is 5555,
+the pod can define that group id using `suplementalGroups` under pod's global
+`securityContext` definition. For example:
+```
+spec:
+  containers:
+    - name: ...
+      #runAsUser: 99 from above has been commented out here
+  securityContext: <1>
+    supplementalGroups: [5555] #an array of GIDs defined globally for the pod
+```
+<1> securityContext must be defined globally to the pod, not under a specific container
+
+Since group id is the focus here, it's worth seeing the ranges defined for the _default_
+project:
+```
+# oc export ns default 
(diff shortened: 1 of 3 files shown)


screeley44:
@jeffvance  - one thing that jumped out at me when I click "view" is that we are discussing SELinux in two places
        SELinux under the Volume Security

        SELinux and NFS Exports Settings

Is this what we want?

install_config/persistent_storage/persistent_storage_nfs.adoc
```diff
@@ -11,122 +11,364 @@
...
+pod's matching SCC, or, if absent in the SCC, then the namespace's `sa.scc.mcs:` value is used. 
+
+*Note:* if pods are not authorizing, or if the NFS mount is failing due to permissions errors, then
+there is a possibility that selinux enforcement is interfering. One way to check for this is to run:
+```
+ausearch -m avc --start recent
+```
+which examines the log file for AVC (Access Vector Cache) errors.
+
+It is also usually necessary to enable *virt_sandbox_use_nfs*, as seen below:
```
screeley44: is this required, I thought only virt_use_nfs label was required?

jeffvance: @screeley44 we need to nail down exactly which *nfs* labels are needed to access the nfs mount.

jeffvance: @screeley44 I'll look at the selinux portion of the persistent_storage_nfs doc to make sure I'm not repeating too much.

jeffvance: re-pushed to clean up the nfs selinux section.
@pmorie @pweil-  PTAL when you get a chance -- thanks!

install_config/persistent_storage/persistent_storage_nfs.adoc
```diff
@@ -11,122 +11,359 @@
...
+(_nobody_ owner), or with 5555 in its supplemental groups, in order to access the directory.
+
+=== User IDs
+User ids can be defined in the container image or in the pod spec. The
+link:pod_security_context.html#user-id[pod and storage security] document covers
+controlling storage access based on user ids.
+
+The container needs it's UID set to 99 (ignoring group ids for the moment), so the
+following fragement can be added to the pod spec:
+```
```
erinboyd: I don't believe we want to advocate the use of nobody:nobody anymore

install_config/persistent_storage/pod_security_context.adoc
```diff
@@ -11,144 +11,395 @@
...
+#(Note: showmount needs access to the ports used by rpcbind and rpc.mount on the nfs server)
+showmount -e <nfs-server-ip-or-hostname>
+Export list for f21-nfs.vm:
+/opt/nfs  *
+
+#on the nfs server:
+# cat /etc/exports
+/opt/nfs *(rw,sync,no_root_squash)
+...
+
```
erinboyd: I thought our export was supposed to be root_squash?

jeffvance: @erinboyd yes, root_squash is the default.

install_config/persistent_storage/persistent_storage_nfs.adoc
jeffvance: @erinboyd I see many references to "nobody" in various docs. What are we advocating instead?

install_config/persistent_storage/pod_security_context.adoc
```diff
@@ -11,144 +11,396 @@
...
+*Note:* the link:#scc[SCC overview], above, should be read before working with FS groups.
+
+`*fsGroup*` defines a pod's "file system group" id, which gets added to the container's supplemental
+groups. As mentioned link:#supplemental-groups[above], the `supplementalGroups` id applies to shared
+storage; whereas, the `fsGroup` id is used for block storage.
+
+Block storage, such as Ceph-RBD, iSCSI, and various cloud storage, is typically dedicated to a single
+pod which has requested the block storage volume, either directly or via a persistent volume claim (PVC).
+Unlike shared storage, block storage is *_taken over_* by a pod, meaning that user and group ids supplied
+in the pod spec (or image) are applied to the actual, physical block device. Typically, block storage is
```
erinboyd: I don't think you can share block storage. Once another pod takes it over, it becomes inaccessible by the other pod.

install_config/persistent_storage/pod_security_context.adoc
```diff
@@ -11,144 +11,396 @@
...
+pod will fail because its group id of *5555* is not within the SCC's or the namespace's range of allowed
+group ids. 
+
+[[scc-supplemental-groups]]
+==== Supplemental Groups and Custom SCCs
+To remedy this situation a custom SCC can be created such that a min and max group id are defined,
+id range checking is enforced, and the group id of 5555 is allowed. It is considered a better
+practice to create new SCCs versus modifying a predefined SCC, or changing the range of allowed
+ids in the predefined projects. 
+
```
erinboyd: Is it worth mentioning what the user gets with the defaults out of the box? Also that supp groups are meant to accommodate existing permissions for shared storage...?

install_config/persistent_storage/pod_security_context.adoc
```diff
@@ -11,144 +11,396 @@
...
+to the new SCC), it will start because the group *5555*, supplied in the pod spec, is now allowed
+by the custom SCC.
+
+[[fsgroup]]
+== FS Group
+*Note:* the link:#scc[SCC overview], above, should be read before working with FS groups.
+
+`*fsGroup*` defines a pod's "file system group" id, which gets added to the container's supplemental
+groups. As mentioned link:#supplemental-groups[above], the `supplementalGroups` id applies to shared
+storage; whereas, the `fsGroup` id is used for block storage.
```
erinboyd: Same thing here, should we mention that OSE will assign a default fsgroup if the user doesn't specify one? I believe this is the preferred method we are trying to promote as well.

install_config/persistent_storage/pod_security_context.adoc
```diff
@@ -158,116 +410,155 @@ Currently the list of volumes which support ownership management includes:
...
+
+- it requests 99 as its user id,
+- since all available SCCs use _MustRunAsRange_ for their `*runAsUser*` strategy, uid range
+checking is required, 
+- 99 is not included in the SCC or in namespace's user id range, so the pod fails.
+
+To fix this situation:
+
+- the _restricted_ SCC could be modified to include 99 within its min and max user id range
+(*not* recommended),
```
erinboyd: I agree this is a proven/known method to fix this situation, however it was recently strongly stated in an email that we are not to advocate editing the restricted scc, but should encourage users to create their own.

install_config/persistent_storage/pod_security_context.adoc
```diff
@@ -158,116 +410,155 @@ Currently the list of volumes which support ownership management includes:
...
+Now, with `runAsUser: 99`, shown in the pod fragment above, the pod matches the new nfs-scc and is
+able to run with a UID of 99.
+
+[[selinux]]
 == SELinuxOptions
-
-The pod security context allows you to specify SELinux labels with which to run
-containers in your pod. Additionally, volumes which support SELinux management
-will be relabeled so that they are accessible by the specified label and,
+All predefined SCCs, except for the _privileged_ SCC, set the `seLinuxContext`
```
erinboyd: Should we mention that selinux relabeling is only supported for non-shared devices?

install_config/persistent_storage/pod_security_context.adoc
```diff
@@ -158,116 +410,155 @@ Currently the list of volumes which support ownership management includes:
...
-
-Currently the list of volumes which support SELinux management includes:
-
-* AWS Elastic Block Store
-* OpenStack Cinder
-* GCE Persistent Disk
-* iSCSI
-* emptyDir
-* Ceph RBD
-* gitRepo
```
erinboyd: I thought gitRepo wasn't supported??

install_config/persistent_storage/pod_security_context.adoc
jeffvance: thus the bold "not recommended" statement... but it is something that the reader might consider and therefore I mention it and say not to do it... Is a stronger warning needed? Is it better to not recognize that an admin might want to modify an existing scc?

install_config/persistent_storage/pod_security_context.adoc
jeffvance: This is from the original doc. I did not change this list. If gitRepo is not supported I will delete it. Can we get confirmation on this @pmorie  or @pweil ?

install_config/persistent_storage/pod_security_context.adoc
jeffvance: The default is mentioned in the section above, specifically see 6 above. Will will be more explicit re. supp-groups and existing perms for shared storage.

install_config/persistent_storage/pod_security_context.adoc
jeffvance: again, see 6. above.

install_config/persistent_storage/pod_security_context.adoc
jeffvance: I was able to share ceph-rbd storage following the (brief) description in the pod_security_context doc. Here's what I did:
 CEPH-POD1:
============
...
spec:
  containers:
  - name: ceph-busybox
    image: busybox
    command: ["sleep", "60000"]
    securityContext:
      runAsUser: 88
    volumeMounts:
    - name: ceph-vol1
      mountPath: /usr/share/busybox
  securityContext:
    fsGroup: 5555
...
 CEPH-POD2:
============
...
spec:
  containers:
  - name: ceph-busybox
    image: busybox
    command: ["sleep", "60000"]
    securityContext:
      runAsUser: 98  # <-- different UID
    volumeMounts:
    - name: ceph-vol1
      mountPath: /usr/share/busybox
  securityContext:
    fsGroup: 5555  # <--- fsgroup is the same
...
oc create -f both ceph pods
#rsh into both pods:
```
# oc rsh ceph-pod1 sh
/ $ id
uid=98 gid=0(root) groups=5000,5555
/ $ ls -l /usr/share/busybox/ -d
drwxrwsr-x    4 root     5555          4096 Feb 18 23:13 /usr/share/busybox/
/ $ touch /usr/share/busybox/FOO-ceph-pod2
/ $ ls -l /usr/share/busybox/
-rw-rwSr--    1 12345    5555             0 Nov  3 19:02 FOO
-rw-r--r--    1 98       5555             0 Feb 26 23:24 FOO-ceph-pod2
/ $ exit
# oc rsh ceph-pod1 sh
/ $ id
uid=88 gid=0(root) groups=5000,5555
/ $ ls -l /usr/share/busybox/ -d
drwxrwsr-x    4 root     5555          4096 Feb 26 23:24 /usr/share/busybox/
/ $ touch /usr/share/busybox/FOO-ceph-pod1
/ $ ls -l /usr/share/busybox/ 
-rw-rwSr--    1 12345    5555             0 Nov  3 19:02 FOO
-rw-rwSr--    1 88       5555             0 Feb 26 23:24 FOO-ceph-pod1
-rw-rwSr--    1 98       5555             0 Feb 26 23:24 FOO-ceph-pod2
/ $ exit

Comment (jeffvance) [inline install_config/persistent_storage/pod_security_context.adoc]: Here's the explanation: My ose env has a single node so both ceph pods are scheduled on the same host. In this case the ceph PV can be shared (regardless of the readonly setting). However, there's a k8s PR (https://github.com/kubernetes/kubernetes/pull/15956) to make the scheduler sensitive to rbd noDiskConflicts. Base on this forward direction I will remove the sentence that mentions the ability to share block storage.

File: install_config/persistent_storage/pod_security_context.adoc (hunk 9 above)
Comment (pweil-) [inline install_config/persistent_storage/pod_security_context.adoc]: Looks like it.  From `git_repo.go`:

```
func (b *gitRepoVolumeBuilder) GetAttributes() volume.Attributes {
	return volume.Attributes{
		ReadOnly:        false,
		Managed:         true,
		SupportsSELinux: true, // xattr change should be okay, TODO: double check
	}
}
```

File: install_config/persistent_storage/persistent_storage_nfs.adoc [hunk 10]
```diff
@@ -11,124 +11,244 @@
...
+- all SCCs available to the pod are examined to see which SCC will allow a user id of 99
+(actually, all policies of the SCCs are checked but the  focus here is on user id),
+- since all available SCCs use _MustRunAsRange_ for their `runAsUser` strategy, uid
+range checking is required, 
+- 99 is not included in the SCC or namespace's user id range, so the pod fails.
+
+It's generally considered a good practice to *not* modify the predefined SCCs. The preferred 
+way to fix this situation is to create a custom SCC, as described in 
+link:pod_security_context#scc-runasuser[pod security and custom SCCs].
+A custom SCC can be created such that a min and max user ids are defined, UID range checking is still enforced, and the UID of 99 will be allowed.
```
pweil-: Not sure if we want to recommend the `anyuid` SCC here or not.  It would satisfy this use case but opens up the possibility of running as root.

pweil-: Actually, thinking about this a bit more, do we even want to recommend the RunAsUser route here?  I thought we wanted to steer people towards the GID solutions.  I understand that this section is about UIDs but as soon as someone reads it they may assume that this is the solution to their problem when we actually want them to run as a random UID if possible and use groups.

install_config/persistent_storage/pod_security_context.adoc
```diff
@@ -11,144 +11,400 @@
...
+in the pod. User ids can also be global, or specific to each container. Four stanzas
+control access to volumes:
+link:#fsgroup[`fsGroup`],
+link:#supplemental-groups[`supplementalGroups`],
+link:#user-id[`runAsUser`], and
+link:#selinux[`SELinuxOptions`].
+
+[[scc]]
+==== SCCs, Defaults, and Allowed Ranges
+SCCs influence whether or not a pod is given a default user id, fsgroup id, supplemental
```
pweil-: Most of this section enhances the SCC documentation (under auth and managing sccs).  Can we work the improvements into those documents and link to it here?  Leaving storage related examples is good.

adellape:
@jeffvance Needs a rebase.

install_config/persistent_storage/persistent_storage_nfs.adoc
jeffvance: I don't see where the anyuid SCC is mentioned above. And in places where I discuss `runAsUser` set to _RunAsAny_ I do mention the risk of allowing a pod to run as root. If developers cannot create their own SCCs then I think we're ok here...
Re. your 2nd comment: if the pod does not have access to the _privileged_ and _anyuid_ SCCs, and if the developer cannot create her own SCC then I think it's ok to discuss user ids. However, I agree that the emphasis should be group ids, and therefore I will move the User IDs section below Group IDs, and will add a comment to both that group ids are the preferred approach.

install_config/persistent_storage/pod_security_context.adoc
jeffvance: @pweil- 2 questions:
1) what is the 1st sentence for 1) in the ...architecture/additional_concepts/authorization.adoc#security-context-constraints doc supposed to be? It reads "Running of"  <-- that's it
2) most of the SCC content of my pod_security_context doc changes are related to user and group ids and the interaction of these ids and SCC and defaults. This seems (to me) more storage specific than what is in the above authorization doc.
I'm thinking there is not very much totally generic scc info in my pod_security_context doc...

jeffvance:
@adellape I rebased but there may be one extraneous merge commit. Let me know...

install_config/persistent_storage/pod_security_context.adoc
pweil-: > This seems (to me) more storage specific than what is in the above authorization doc.

I can live with this.  The way I've tried to break it up is:

1.  all mechanics of SCCs in the auth doc
2.  all runtime commands and "how do I" in the managing SCCs doc

I think at a minimum we can get by with having a blurb in the auth doc like `for an in depth discussion of how RunAsUser, FSGroup, and SupplementalGroups affect storage please see <link>`

install_config/persistent_storage/persistent_storage_nfs.adoc
pweil-: > I don't see where the anyuid SCC is mentioned above

My sentence was not clear.  I was trying to say that we could mention the anyuid SCC as an alternative to creating a custom SCC.  However, if it is mentioned it should contain a warning about the root UID.

install_config/persistent_storage/pod_security_context.adoc
jeffvance: agreed, and I will add that sentence to both the "managing sccs" and "authorization#scc" docs.
@pweil- Can you help me with this sentence in:
https://github.com/openshift/openshift-docs/blob/master/architecture/additional_concepts/authorization.adoc#security-context-constraints?
"SCCs are objects that define a set of conditions that a pod must run with in order to be accepted into the system. They allow an administrator to control the following:
1) Running of"
                ^^^^^^^^ what is the rest of this sentence supposed to be?  Thanks!

pweil-: https://docs.openshift.org/latest/architecture/additional_concepts/authorization.html#security-context-constraints .  It should say "privileged containers".  It looks like it is wrapped in a conditional in your link:

```
. Running of
ifdef::openshift-enterprise,openshift-origin[]
link:../../install_config/install/prerequisites.html#security-warning[privileged
containers].
endif::[]
```

install_config/persistent_storage/persistent_storage_nfs.adoc
jeffvance: @pweil- My 3rd bullet in the fragment above reads:
"since all available SCCs use MustRunAsRange for their runAsUser strategy, uid range checking is required", with _available_ be the key word. What does the pod spec need to cause it to  use the _anyuid_ SCC? I assumed
```
groups:
- system:cluster-admins
```
but maybe not? Anyway, before I mention the _anyuid_ SCC I want to understand how a pod can match it...

pweil-: The anyuid scc has a priority set so it will always be tested first IF the actor creating the pod has access to it via user or group grants - with the assumption that the pod is not making other privileged requests that anyuid would  deny.  By default it is only granted to admins but it is our recommended way of running dockerhub images that require root (https://docs.openshift.org/latest/admin_guide/manage_scc.html#enable-dockerhub-images-that-require-root)

jeffvance:
@pweil- I think this latest push covers all of your comments. Please let me know. Thanks!

pweil-:
@jeffvance minor nits

jeffvance:
@adellape I pushed my latest changes. PTAL. Thanks.

pweil-:
:+1:

erinboyd:
LGTM

install_config/persistent_storage/persistent_storage_nfs.adoc
```diff
@@ -11,124 +11,248 @@
...
-[IMPORTANT]
-====
-High-availability of storage in the infrastructure is left to the underlying
-storage provider.
-====
-
-[[nfs-provisioning]]
+
+OpenShift clusters can be provisioned with
+link:../../architecture/additional_concepts/storage.html[persistent storage] using NFS
```
adellape: Punctuation.

install_config/persistent_storage/persistent_storage_nfs.adoc
```diff
@@ -11,124 +11,248 @@
...
+link:../../architecture/additional_concepts/storage.html#persistent-volume-claims[persistent volume claims (PVCs)],
+and https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Storage_Administration_Guide/ch-nfs.html[NFS] is beneficial.
+
+The Openshift persistent volume (PV) framework allows administrators to provision a cluster
+with persistent storage and gives developers the ability to request those resources without
+having specific knowledge of the underlying infrastructure. Persistent volumes are not bound
+to a single project; they can be shared across the entire OpenShift cluster. Persistent volume
+claims (PVCs), however, are specific to a project and are created and used by developers.
+
+(*Note:* "project" and "namespace" are used interchangeably throughout this document)
```
adellape: Put this in a [NOTE] block:

```
[NOTE]
====
This document uses "project" and "namespace" interchangeably throughout.
====
```

install_config/persistent_storage/persistent_storage_nfs.adoc
```diff
@@ -11,124 +11,248 @@
...
 
 == Provisioning
```
adellape: The unique ID of `[[nfs-provisioning]]` has been lost from up above. We need to retain that as anchor IDs must be unique across the entire doc set.

install_config/persistent_storage/persistent_storage_nfs.adoc
```diff
@@ -11,124 +11,248 @@
...
-Storage must exist in the underlying infrastructure before it can be mounted as
-a volume in OpenShift. To provision NFS volumes in OpenShift, all that is
-required is a distinct list of NFS servers and paths and the
-`*PersistentVolume*` API.
 
-You must define your persistent volume in an object definition before creating
-it in OpenShift:
+Storage must exist in the underlying infrastructure before it can be mounted as a volume in
+OpenShift. To provision NFS volumes a list of NFS servers and export paths are all that is required.
```
adellape: volumes,

install_config/persistent_storage/persistent_storage_nfs.adoc
```diff
@@ -11,124 +11,248 @@
...
 
-.Persistent Volume Object Definition Using NFS
-====
+*Note:*
+Each NFS volume must be mountable by all schedulable nodes in the cluster.
```
adellape: Use a [NOTE] block.

install_config/persistent_storage/persistent_storage_nfs.adoc
```diff
@@ -11,124 +11,248 @@
...
 
-[source,yaml]
-----
-apiVersion: "v1"
-kind: "PersistentVolume"
+A PV spec file is usually defined before creating the PV:
```
adellape: s/spec file/object definition/

To match how (most of) the rest of the docs refer to these sorts of definitions.

install_config/persistent_storage/persistent_storage_nfs.adoc
```diff
@@ -11,124 +11,248 @@
...
+```
+apiVersion: v1
```
adellape: Example files should use formatting as seen in https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#examples

Basically:

```
====
[source,yaml]
----
Stuff
in
a file.
----
====

File: install_config/persistent_storage/persistent_storage_nfs.adoc [hunk 19]
```diff
@@ -11,124 +11,248 @@
...
-<3> This is the policy for restricting read/writes to a shared volume.
-<4> This defines the volume type being used, in this case the *nfs* plug-in.
-<5> The path that is exported by the NFS server.
-<6> The host name or IP address of the NFS server.
-<7> Defines what happens to a volume when released from its claim. Valid options
+    path: /tmp <5>
+    server: 172.17.0.2 <6>
+  persistentVolumeReclaimPolicy: Recycle <7>
+```
+<1> the name of the volume. This is the PV identify in various `oc ... pod` commands.
```
adellape: s/the/The/

Need init caps on each of the beginnings of these callouts below as well.

adellape: I assume:

s/identify/identity/

For `oc ... pod`, maybe use:

```
in various `oc <command> pod` operations.
```

install_config/persistent_storage/persistent_storage_nfs.adoc
```diff
@@ -11,124 +11,248 @@
...
-====
 
-Save your definition to a file, for example *_nfs-pv.yaml_*, and create the
-persistent volume:
-
-====
-----
+Save the definition to a file, for example *_nfs-pv.yaml_*, and create the persistent
+volume:
+```
```
adellape: Switch back to former block formatting. I'll stop commenting to this effect; applies to the rest of the changes.

install_config/persistent_storage/persistent_storage_nfs.adoc
```diff
@@ -11,124 +11,248 @@
...
 pv0001                   <none>    5368709120   RWO           Available                       31s
-----
-====
+```
 
-Users can then link:../../dev_guide/persistent_volumes.html[request storage
-using persistent volume claims], which can now utilize your new persistent
-volume.
+The next step can be to create a persistent volume claim (PVC) which will bind to the new PV.
+This is easy to do, as shown in the "nfs-claim.yaml example below:
```
adellape:
```
*_nfs-claim.yaml_*
```

install_config/persistent_storage/persistent_storage_nfs.adoc
```diff
@@ -11,124 +11,248 @@
...
+    requests:
+      storage: 1Gi <2>
+
+# oc create -f nfs-claim.yaml
+```
+<1> as mentioned above for PVs, the `accessModes` do not enforce security, but rather act as
+labels to match a PV to a PVC.
+<2> this claim will look for PVs offering 1Gi or greater capacity.
 
+*Note:* PVs, and PVCs are not necessary, just convenient, and make sharing a volume across a
```
adellape: "Use a [NOTE] block" applies to the rest of the `*Note:*`s below as well.

Drop the comma after "PVs".

install_config/persistent_storage/persistent_storage_nfs.adoc
```diff
@@ -11,124 +11,248 @@
...
+project simpler. The NFS specific information contained in the PV spec can also be defined
```
adellape: s/spec/definition/

install_config/persistent_storage/persistent_storage_nfs.adoc
```diff
@@ -11,124 +11,248 @@
...
+directly in a pod spec.
```
adellape: Assume s/spec/definition/ throughout the rest.

install_config/persistent_storage/persistent_storage_nfs.adoc
```diff
@@ -11,124 +11,248 @@
...
 
-Enforcing quotas in this way allows the end user to request persistent storage
+Enforcing quotas in this way allows the developer to request persistent storage
 by a specific amount (e.g, 10Gi) and be matched with a corresponding volume of
 equal or greater capacity.
 
 [[nfs-volume-security]]
+== Volume Security
+This section covers NFS volume security, including matching permissions and selinux considerations. 
+The reader is expected to understand the basics of POSIX permissions, process UIDs, supplemental groups, and selinux.
```
adellape: s/selinux/SELinux/

install_config/persistent_storage/persistent_storage_nfs.adoc
```diff
@@ -11,124 +11,248 @@
...
+
+*Note:* The
+link:pod_security_context.html[pod and storage security] document should be read before
```
adellape:
```
See link:pod_security_context.html[Volume Security] before implementing NFS volumes.
```

install_config/persistent_storage/persistent_storage_nfs.adoc
jeffvance: Alex, what is a "doc set"? I'm assuming the tag id must be unique across more than the containing doc then...

adellape: @jeffvance Each of the root-level directories in the repo build out as their own "book" on the Customer Portal and as the left-nav categories on docs.openshift.com/enterprise (for example). So I just mean they must unique between all the books, i.e., across the whole repo.

adellape:
Spoke with @jeffvance today and I'm going to consume his commits here into my own PR, do some editing there (e.g., the comments I've made so far in this PR and more), then we'll continue in my PR and ultimately close this one.

install_config/persistent_storage/persistent_storage_nfs.adoc
jeffvance: ok, good to know!

jeffvance:
@adellape what's the latest on this PR ?

adellape:
Closing in favor of https://github.com/openshift/openshift-docs/pull/1726.

Issue #1404, opened 2016-01-07
Aggregate Logging steps need to be updated


detiber:
https://docs.openshift.com/enterprise/3.1/install_config/aggregate_logging.html

On Topic 1. 
Should inform to join on project, after created. 

oc project logging

Or use "oc new-project logging" to create instead oadm...

2. Inform the option without certs:
oc secrets new logging-deployer nothing=/dev/null

akram:
Hi @detieber 

it is also required to update the oc edit scc section, otherwise we got the following error.

```
Deleting existing secrets
+ echo 'Deleting existing secrets'
+ oc delete secret logging-fluentd logging-elasticsearch logging-kibana logging-kibana-proxy logging-kibana-ops-proxy
Error from server: User "system:serviceaccount:default:logging-deployer" cannot delete secrets in project "default"
Error from server: User "system:serviceaccount:default:logging-deployer" cannot delete secrets in project "default"
Error from server: User "system:serviceaccount:default:logging-deployer" cannot delete secrets in project "default"
Error from server: User "system:serviceaccount:default:logging-deployer" cannot delete secrets in project "default"
Error from server: User "system:serviceaccount:default:logging-deployer" cannot delete secrets in project "default"
Creating secrets
+ :
+ echo 'Creating secrets'
+ oc secrets new logging-elasticsearch key=/etc/deploy/keystore.jks truststore=/etc/deploy/truststore.jks searchguard.key=/etc/deploy/searchguard_node_key.key
Error from server: User "system:serviceaccount:default:logging-deployer" cannot create secrets in project "default"
```

It is missing:
```
- system:serviceaccount:default:logging-deployer
```

bfallonf:
@detiber That issue will be fixed in #1641 

@akram I'm not too sure which section you're referring to. Is it step 2 of the pre-deployment phase?

https://docs.openshift.com/enterprise/3.1/install_config/aggregate_logging.html#pre-deployment-configuration

ewolinetz:
@bfallonf so that they're not lost there are also two PRs that I opened to improve documentation for EFK as well.
#1613 and #1625

bfallonf:
@ewolinetz Ah I see. Thanks for the notification. I can make some comments on #1613  and get the in, but should we close #1625 in favor of my #1641 ?

ewolinetz:
@bfallonf it makes sense to me to close #1625. I'll address your comments in #1613

bfallonf:
[...]

PR #1663, opened 2016-02-29
Changed service account section error

bfallonf:
@akram as per #1404 

Also, I didn't think to ask until after I created this PR, but am I replacing the line, or adding your suggested line? Both seemed to work for me...

### DIFF
--- a/install_config/aggregate_logging.adoc
+++ b/install_config/aggregate_logging.adoc
@@ -145,7 +145,7 @@ project created above in the following example:
 ====
 ----
 $ oc policy add-role-to-user edit \
-            system:serviceaccount:logging:logging-deployer
+            system:serviceaccount:default:logging-deployer
 ----
 ====
 


akram:
I added it as well, but apparently, since we are doing everything in the default namespace, the previous line is not required.
I suggest to remove it.

bfallonf:
That works, because that's what I've already done :)

I'll get this merged. Thanks @akram

bfallonf:
**revhistory note:**

(Install & Config -> Aggregating Container Logs )

```
Fixed an service account error in the link:../install_config/aggregate_logging.adoc[Aggregating Container Logs] topic.
```

adellape:
@bfallonf Needs label/milestone.

adellape:
Removed enterprise-3.0 label. This topic was `admin_guide/aggregate_logging.html` back then, but also this change isn't applicable to logging back then.

PR #1665, opened 2016-02-29
Clarified text around access modes

markturansky:
Updated text WRT access modes.  Two questions were recently brought up:
1. How does access mode matching work?
2. Are modes enforced at runtime?

The two changes in this PR attempt to answer both questions.


adellape:
:+1: Thanks.

ahardin-rh:
revhistory in followup

PR #1670, opened 2016-03-01
YAML edit as per suggestion

bfallonf:
Edit reported to me via email

### DIFF
--- a/dev_guide/builds.adoc
+++ b/dev_guide/builds.adoc
@@ -881,28 +881,26 @@ Image inputs are specified in the `*source*` definition of the `*BuildConfig*`:
 ====
 [source,yaml]
 ----
-    source:
-      git:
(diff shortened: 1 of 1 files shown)


GrahamDumpleton:
Looks better now presuming is then rendered properly in browser.

bfallonf:
@GrahamDumpleton The problem seemed to stem that there were too many spaces and indents and the like. Should be good now.

Thanks!

bfallonf:
Just fixed a yaml eg so it worked. Probably no rev history needed here.

adellape:
@bfallonf Label/milestone.

adellape:
Removed this from "Next Release" as it modifies https://github.com/openshift/openshift-docs/pull/1445 which is for OSE 3.2.

Issue #1629, opened 2016-02-20
iOS displays footer that obscures page in portrait mode


smarterclayton:
![image](https://cloud.githubusercontent.com/assets/1163175/13197911/710ce4ce-d7c9-11e5-9f38-636e34e5005b.png)

adellape:
cc @nhr

PR #1673, opened 2016-03-01
Include an alternate mobile footer that is position:relative so that it doesn't interfere with the toggle off canvas nav at mobile resolutions.

sg00dwin:
Fixes https://github.com/openshift/openshift-docs/issues/1629


sg00dwin:
@nhr

sg00dwin:
@nhr updated pr with your changes

PR #1682, opened 2016-03-02
[do not merge] update information about OAuth grant strategies

stevekuznetsov:
Updates information about how we handle OAuth client grant strategies, follow-up to https://github.com/openshift/origin/pull/7739

@liggitt PTAL

@adellape FYI, second commit changes a JSON example to YAML as it's infinitely easier to read YAML and YAML also seems to be the de-facto standard for the rest of these docs

### DIFF
--- a/architecture/additional_concepts/authentication.adoc
+++ b/architecture/additional_concepts/authentication.adoc
@@ -145,11 +145,13 @@ $ oc create -f <(echo '
   "redirectURIs": [
     "http://www.example.com/" <3>
   ]
+  "grantStrategy": "...", <4>
 }')
 ----
 <1> The `name` of the OAuth client is used as the `client_id` parameter when making requests to `_<master>_/oauth/authorize` and `_<master>_/oauth/token`.
 <2> The `secret` is used as the `client_secret` parameter when making requests to `_<master>_/oauth/token`.
 <3> The `redirect_uri` parameter specified in requests to `_<master>_/oauth/authorize` and `_<master>_/oauth/token` must be equal to (or prefixed by) one of the URIs in `redirectURIs`.
+<4> The `grantStrategy` is used to determine what action to take when this client requests tokens and has not yet been granted access by the user. Uses the same values seen in link:./../../install_config/configuring_authentication.html#grant-options[Grant Options].
 ====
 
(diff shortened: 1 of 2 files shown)


architecture/additional_concepts/authentication.adoc
```diff
@@ -135,23 +135,29 @@ To register additional clients:
...
+grantStrategy: "...", <4>
+')
 ----
 <1> The `name` of the OAuth client is used as the `client_id` parameter when making requests to `_<master>_/oauth/authorize` and `_<master>_/oauth/token`.
 <2> The `secret` is used as the `client_secret` parameter when making requests to `_<master>_/oauth/token`.
 <3> The `redirect_uri` parameter specified in requests to `_<master>_/oauth/authorize` and `_<master>_/oauth/token` must be equal to (or prefixed by) one of the URIs in `redirectURIs`.
+<4> The `grantStrategy` is used to determine what action to take when this client requests tokens and has not yet been granted access by the user.
 ====
 
+Valid values for `*grantStrategy*` are:
```
stevekuznetsov: @adellape is this the best way to present these values here? Should we link the two tables together instead of duplicating them?

adellape: Linking seems like it'd work. Something like:

```
<4> The `grantStrategy` is used to determine what action to take when this client requests tokens and has not yet been granted access by the user. Uses the same values seen in link:#grant-options[Grant Options].
```

adellape:
> YAML also seems to be the de-facto standard for the rest of these docs

+1, thanks.

adellape:
@stevekuznetsov When this is ready to merge, please squash and lmk. :bow:

stevekuznetsov:
Waiting on @liggitt approval of upstream pull https://github.com/openshift/origin/pull/7739

stevekuznetsov:
@sgallagher you are taking this over, right?

sgallagher:
@stevekuznetsov Yeah, I think this needs to be rolled into https://github.com/openshift/origin/pull/9616

sgallagher:
I think we can close this in favor of https://github.com/openshift/openshift-docs/pull/2488

PR #1692, opened 2016-03-04
add clarity to hpa doc

stevekuznetsov:
@adellape I'm assuming that this is what is actually meant by the sentence - the HPA will _only_ work if the last deployment is complete.

### DIFF
--- a/dev_guide/pod_autoscaling.adoc
+++ b/dev_guide/pod_autoscaling.adoc
@@ -65,10 +65,9 @@ and scales up or down accordingly. The scaling will occur at a regular interval,
 but it may take one to two minutes before metrics make their way into Heapster.
 
 For replication controllers, this scaling corresponds directly to the replicas
-of the replication controller. For deployment configurations, this will scale
-the latest deployment, or, if no deployment is present, the replication
-controller template in the deployment configuration. Note that scaling will
-occur if the latest deployment is marked as `Complete`.
+of the replication controller. For deployment configurations, scaling corresponds
+directly to the replica count of the deployment configuration. Note that autoscaling
+applies only to the latest deployment in the `Complete` phase.
 
 [[creating-a-hpa]]
 == Creating a Horizontal Pod Autoscaler


adellape:
That is my assumption. @DirectXMan12 Can you confirm?

DirectXMan12:
I believe it's actually the current "active" deployment, but yes.

stevekuznetsov:
@adellape is there a concept of 'active' deployment in the current documentation? If we changed this to read 'only if the active deployment...' would that be comprehensible?

adellape:
> is there a concept of 'active' deployment in the current documentation?

@stevekuznetsov Hm, not really. While I believe I understand the meaning, I'm struggling to find a distinction looking in the CLI and web console. Isn't the current, active deployment always identified as "latest" in the UI? Even when you "rollback" to a previous deployment, a new deployment is ultimately created and marked "latest". Or am I misunderstanding? cc @sallyom 

If that is the case, then probably better to stick with "latest" in the sentence, as I didn't see anything to indicate a deployment as being "active" (in those words).

DirectXMan12:
So, in the code, there's the "latest" deployment (which might have failed), and the "active" deployment (the most recent deployment marked "complete").  When you scale, you scale the "active" deployment -- for example, see https://github.com/openshift/origin/blob/master/pkg/deploy/controller/deploymentconfig/controller_test.go#L285-L298, which demonstrates the case


```
somedc-1: complete, 5 replicas
somedc-2: failed, 0 replicas

$ oc scale dc somedc --replicas=1

somedc-1: complete, 1 replica
somedc-2: failed, 0 replicas
```

What does the web console do in this case?

jwforres:
@spadgett is the right person for HPAs in web console

spadgett:
> What does the web console do in this case?

@DirectXMan12 The web console is consistent with the CLI.

dev_guide/pod_autoscaling.adoc
```diff
@@ -68,7 +68,7 @@ For replication controllers, this scaling corresponds directly to the replicas
 of the replication controller. For deployment configurations, this will scale
 the latest deployment, or, if no deployment is present, the replication
 controller template in the deployment configuration. Note that scaling will
-occur if the latest deployment is marked as `Complete`.
+only occur if the latest deployment is marked as `Complete`.
```
spadgett: @DirectXMan12 @ironcladlou Is this more accurate? I'm not sure the last sentence is even necessary.

"For deployment configurations, this will scale the last `Complete` deployment, or, if no `Complete` deployment is present, the replication controller template in the deployment configuration."

DirectXMan12: That seems better to me.

stevekuznetsov: I'll update on @ironcladlou ack.

ironcladlou: When the autoscaler targets a deploymentConfig, it should only be setting `deploymentConfig.spec.replicas`.

>For deployment configurations, this will scale the last Complete deployment, 

This is true, but only indirectly: the scaler sets `spec.replicas` on the deploymentConfig, causing the latest deployment to be scaled by another mechanism (the deployment controllers). Not sure if that matters in the context of this documentation, but felt it worth clarifying on this thread.

>or, if no Complete deployment is present, the replication controller template in the deployment configuration.

The deploymentConfig has a spec (which is a deploymentConfigSpec with a replicas field) and a podTemplateSpec (which isn't a replication controller template).

The way I would have worded this is: "For deployment configurations, scaling corresponds directly to the replica count of the deployment configuration." and go from there if you need to explain how changing that field affects deployment scaling.

stevekuznetsov:
@ironcladlou @DirectXMan12 @adellape  what do you think about:

> For deployment configurations, scaling corresponds directly to the replica count of the deployment configuration. Note that autoscaling applies only to the latest complete deployment.

ironcladlou:
@stevekuznetsov 
>This leaves room for the "latest deployment failed but one before it didn't" scenario without going into too many technical details.

SGTM.

adellape:
@stevekuznetsov :+1:

stevekuznetsov:
@adellape updated the PR

dev_guide/pod_autoscaling.adoc
```diff
@@ -65,10 +65,9 @@ and scales up or down accordingly. The scaling will occur at a regular interval,
 but it may take one to two minutes before metrics make their way into Heapster.
 
 For replication controllers, this scaling corresponds directly to the replicas
-of the replication controller. For deployment configurations, this will scale
-the latest deployment, or, if no deployment is present, the replication
-controller template in the deployment configuration. Note that scaling will
-occur if the latest deployment is marked as `Complete`.
+of the replication controller. For deployment configurations, scaling corresponds
+directly to the replica count of the deployment configuration. Note that auto-
+scaling applies only to the latest deployment in the `Complete` phase.
```
adellape: @stevekuznetsov In the GH preview at least, this is showing up as "auto- scaling" (with a space). Can you check on that?

stevekuznetsov: @adellape good catch, forgot the newline could do that. fixed.

adellape:
@stevekuznetsov LGTM, thanks.

adellape:
[rev_history]
|link:../dev_guide/pod_autoscaling.html[Pod Autoscaling]
|Updated to add clarity to link:../dev_guide/pod_autoscaling.html#hpa-autoscaling[autoscaling] behavior.
%

PR #1697, opened 2016-03-07
Update docs for SDN defaults changes

danwinship:
Update the docs for https://github.com/openshift/origin/pull/6597. (This will be in 3.2.)


adellape:
:+1: Thanks.

No revhistory needed, but this does seem like 3.2 rel notes fodder.

PR #1699, opened 2016-03-07
adjust to the more fluid resource limits; other fine tuning

gabemontero:
@adellape PTAL

Just some minor edits:
1) added a link to Quota section, and treat Quota and Compute Resource as equal / peer concepts
2) some of the limits on resources are still somewhat fluid; for those, rather than note the specific number, just note a limit will be placed on them
3) we backed off of the fine tuned mem limits based on scenario, and now set everyone at 512Mi to aid density and maintenance of the online env.  So adjusted the verbiage to back off the fine tuning slant.


gabemontero:
@adellape bump - can we get this reviewed / edited / merged?  thanks

PR #1274, opened 2015-11-30
Add oc create route subcommands docs

0xmichalis:
Origin PR: https://github.com/openshift/origin/pull/6819

PR #1717, opened 2016-03-09
Update router docs to use oc create route

0xmichalis:


install_config/install/deploy_router.adoc
```diff
@@ -350,12 +350,12 @@ named `www.example.test`, see the example shown below:
...
 ----
 
-Generate a route configuration file using the above certificate and key.
-Make sure to replace servicename `my-service` with the name of your service.
+Generate a route using the above certificate and key.
 
 ----
-# servicename="my-service"
-# echo "
+$ oc create route edge secured-edge-route --service=my-service --hostname=www.example.test --key=example-test.key --cert=example-test.crt
```
tnguyen-rh: Better to split this into multiple lines, all but the last ending in backslash.  Because the rendering of the docs on the portal is (still!) very vertical, i suggest line break after `secured-edge-route` and also after every option.  For more readability, you can omit the `=` between option name and option value, but that particular change is ... optional. :smile:

tnguyen-rh:
@kargakis 
One refinement opportunity, otherwise LGTM.
I presume this is for 3.2 -- is that correct?

install_config/install/deploy_router.adoc
kargakis: Will split the flags into separate lines though I prefer to keep `=` because it makes it explicit that this is not a boolean flag:)

kargakis:
@tnguyen-rh I have made the change you asked for and also dropped the route name (if we don't specify one the service name is re-used) for better readability. PTAL

`oc create route` will be shipped in 3.2 so it's good to have docs synced. Actually this change should be part of https://github.com/openshift/openshift-docs/pull/1274, I just missed it.

tnguyen-rh:
@kargakis LGTM.
@adellape PTAL label and r.note.

[revhistory note]
|link:../install_config/install/deploy_router.html#using-secured-routes[Installation and Configuration -> Installing -> Deploying a Router -> Using Secured Routes]

|Updated the example to use `oc create route`.
%

install_config/install/deploy_router.adoc
```diff
@@ -350,35 +350,33 @@ named `www.example.test`, see the example shown below:
...
+Generate a route using the above certificate and key.
 
 ----
-# servicename="my-service"
-# echo "
+$ oc create route edge --service=my-service \ 
+    --hostname=www.example.test \
+    --key=example-test.key --cert=example-test.crt
+route "my-service" created
+$ oc get route/secured-edge-route -o yaml
```
adellape: @kargakis Should `secured-edge-route` instead be `my-service` in this example? It seems the route name is just inherited from the service name?

Also, should we use `oc export route/my-service` instead? Looks like basically the same output `get` with `-o yaml`, just missing a few lines of metadata (AFAICT: namespace, resourceVersion, selfLink, uid).

adellape:
> |link:../install_config/install/deploy_router.html#using-secured-routes[Installing -> Deploying a Router -> Using Secured Routes]

@tnguyen-rh So far the "Affected Topic" column has stopped short of listing the affected section, and leaves that part to the "Description of Change" column.

e.g.: "Updated an example in the Using Secured Routes section to use oc create route."

tnguyen-rh:
@adellape
Thanks for the explanation.  IIUC then, the general form is to use the link w/o specific `#id` for "Affected Topic" and include the same link (w/ `#id`s, if appropriate) in the "Description of Change" text.  Here's attempt 2 (if ok, i'll delete the previous attempt):

[revhistory note]
|link:../install_config/install/deploy_router.html[Installing -> Deploying a Router]

|Updated an example in the link:../install_config/install/deploy_router.html#using-secured-routes[Using Secured Routes] section to use `oc create route`.
%

adellape:
@tnguyen-rh Yeah, LGTM. That general form you describe I think gives us wiggle room on the DoC text, as A) there may not always be a specific `#id` we want to link to (or any specific link for that matter), and B) sometimes we may have multiple rows of DoC for a single Affected Topic.

tnguyen-rh:
@adellape OK, thanks.  I've not deleted the attempt 1, but instead crippled it.

PR #1721, opened 2016-03-10
[QA] Bug 1303361 - restricting node placement

bfallonf:
@dobbymoodge FYI


admin_guide/preventing_node_placement.adoc
```diff
@@ -0,0 +1,35 @@
...
+As a cluster administrator, you can prevent users from targeting specific nodes when scheduling pods. Constraining node placement can be grouped with the scheduler and node selectors to control pod placement. 
+
+== Constraining Pod Placement
+
+To prevent users from targeting specific nodes, modify the master configuration file to look similar to the following:
+
+====
+----
+...
+admissionConfig
```
dobbymoodge: This should also mention the `kubernetesMasterConfig`, as in here: https://github.com/openshift/origin/blob/master/pkg/scheduler/admission/podnodeconstraints/doc.go

admin_guide/preventing_node_placement.adoc
```diff
@@ -0,0 +1,74 @@
...
+As a cluster administrator, you can prevent users from targeting specific nodes
+when scheduling pods.  
+
+== Constraining Pod Placement
+
+To prevent users from targeting specific nodes, modify the OpenShift master
+configuration, and the Kubernetes resource configuration.
+
+Adding a label to the `nodeSelectorLabelBlacklist` section of a master
+configuration file prevents pods being deployed onto the nodes with that label.
```
dobbymoodge: This isn't correct - adding a label to the blacklist prevents Pods from being created if their NodeSelector carries any of the blacklisted labels. If the NodeSelector is empty or carries no blacklisted labels, then the scheduler may still place the Pod on a Node that has a blacklisted label.

Say you have a Node with these labels:
```
  - kubernetes.io/hostname: node1
  - foo: bar1
  - baz: quux
```

Also, your `NodeSelectorBlacklist` looks like this:
```
  - kubernetes.io
  - foo
  - shmekky
```

Users which lack the `pods/binding` permission will be able to:

* Create Pods with empty NodeSelector fields
* Create Pods with a NodeSelector label which isn't blacklisted
  * This can be a label that exists on a Node, like `baz: quux`
  * One could also specify a label that doesn't exist on a Node, and it should still work, such as `fizz: buzz` in this example

Users which lack the `pods/binding` permission will **not** be able to:

* Create Pods which specify a NodeName
  * The NodeName field is pretty much the same as a NodeSelector of `kubernetes.io/hostname: node1` in this example
* Create Pods with a NodeSelector label that **is** blacklisted
  * This can be a label that exists on a Node, like `foo: bar`
  * Labels which don't exist on any Nodes but are in the blacklist will also be prohibited, such as `shmekky: blekky`

For Users, ServiceAccounts, or Groups which *do* have the `pods/binding` permission, the NodeName constraint and NodeSelector label blacklist will have no effect.

In all of these cases, the scheduler may *still* correctly schedule a Pod on a Node which bears a blacklisted label as long as it's in the set of schedulable Nodes. That set is determined by a number of other criteria, such as project-level NodeSelectors, etc.

One more example that might make the function of this plugin more clear. Say your cluster is spread across five datacenter "regions": "us-east", "us-central", "us-west", and across the ocean "apac-east" and "apac-west". Nodes which are physically located in each region are labelled accordingly, i.e. `region: us-west` for example. In addition, the specific regions are grouped by labels into "super-regions": `superregion: us` and `superregion: apac`

As the cluster administrator, you may want to group Users by region and configure their projects so that their new Pods are scheduled on Nodes which are geographically local to the User. For Users in the US, their projects are created with the project-level NodeSelector `superregion: us`, and for users in the Asia-Pacific region, their projects are created with project-level NodeSelector `superregion: apac`.

For resource management reasons, you want to make sure load is spread evenly across all Nodes in a super-region. You also want to allow administrators to place new Pods in particular regions for monitoring purposes. To prevent Users from weighting a particular region, you add the `region` label to the `NodeSelectorLabelBlacklist`. Now whenever a User in the US - who by policy doesn't have `pods/binding` permission - creates a Pod, it gets scheduled to a Node in any of the three regions that overlap with the `us` super-region with equal likelihood, according to the best-fit criteria for your cluster. If the User tries to target a `region` with their Pod, it will fail. If they try again, but omit the NodeSelector on their Pod, it may be scheduled on the region they tried to target, because the `us` super-region is set as the project-level NodeSelector, and those Nodes are labeled as part of the `us` super-region.

Your administrators are still able to set the `region` label in their Pod NodeSelector fields because the `cluster-admin` role has the `pods/binding` permission.

I hope that makes more sense.

bfallonf:
@dobbymoodge Thanks very much for the detailed comment. I've added the second use case you mention, which I think is a true winner. Let me know if I took out the wrong details or something.

Another question: Does nothing else have to be added to the procedure? I know it presumes there are labels and node selectors set (or one of each or whatever), but that's not necessarily part of the procedure, and really just used to forward the use case. WDYT?

Thanks again!

bfallonf:
@dobbymoodge Sorry, one more question:

What's the location of the kubernetes master config? I couldn't find it in my own test instance...?

bfallonf:
@dobbymoodge bumping this. It's 3.2 work, so it'd be great to get it out the door. Thanks!

dobbymoodge:
@bfallonf The kubernetes master config is specified in the same place as
the origin master config, generally `master-config.yaml`. It's the section
of the config under `kubernetesMasterConfig`, as can be seen here:
https://docs.openshift.com/enterprise/3.1/install_config/master_node_configuration.html#master-configuration-files

admin_guide/preventing_node_placement.adoc
```diff
@@ -0,0 +1,101 @@
...
+
+== Overview
+
+As a cluster administrator, you can prevent users from targeting specific nodes
+when scheduling pods using labels and node selectors, as well as the node
+blacklist setting.  
+
+== Constraining Pod Placement
+
+Adding a label to the `nodeSelectorLabelBlacklist` section of a master
```
dobbymoodge: This is incomplete and ambiguous; maybe it could be better worded like:

The `nodeSelectorLabelBlacklist` configuration field allows an administrator to control which labels certain users are able to specify in new Pod spec `nodeSelector` fields. Users, service accounts and members of groups which have the `pods/binding` permission can specify any `nodeSelector`; those without the `pods/binding` permission are prohibited from setting a `nodeSelector` for any label that appears in `nodeSelectorLabelBlacklist`.

admin_guide/preventing_node_placement.adoc
```diff
@@ -0,0 +1,101 @@
...
+configuration file prevents pods from being created if their labels match any
+values in the `nodeSelectorLabelBlacklist`. This means, if no node selector is
+available, then the scheduler can still place a pod on a node with a blacklisted
```
dobbymoodge: @bfallonf This sentence seems unnecessary, and the wording is ambiguous and confusing. The example below explains this particular behavior of the plugin clearly enough, with the benefit of more complete context.

admin_guide/preventing_node_placement.adoc
```diff
@@ -0,0 +1,101 @@
...
+:experimental:
+:toc: macro
+:toc-title:
+
+toc::[]
+
+== Overview
+
+As a cluster administrator, you can prevent users from targeting specific nodes
+when scheduling pods using labels and node selectors, as well as the node
```
dobbymoodge: more accurate: "As a cluster administrator, you can set policy to prevent certain roles from targeting single nodes for pod scheduling, as well as defining how users may control which groups of nodes their new pods may be scheduled on."

admin_guide/preventing_node_placement.adoc
```diff
@@ -0,0 +1,101 @@
...
+blacklist setting.  
+
+== Constraining Pod Placement
+
```
dobbymoodge: Maybe add a bit more detailed introduction here:

"Enabling this plugin will require the `pods/binding` permission for roles which need to target particular nodes via the `nodeName` Pod spec field. It will also enforce constraints on the use of the `nodeSelector` Pod spec field based on the `pods/binding` permission and the `nodeSelectorLabelBlacklist` configuration option."

admin_guide/preventing_node_placement.adoc
```diff
@@ -0,0 +1,101 @@
...
+Adding a label to the `nodeSelectorLabelBlacklist` section of a master
+configuration file prevents pods from being created if their labels match any
+values in the `nodeSelectorLabelBlacklist`. This means, if no node selector is
+available, then the scheduler can still place a pod on a node with a blacklisted
+label.
+
+For example, if you have a cluster consisting of five datacentres spread across
```
dobbymoodge: https://obriend.fedorapeople.org/WritingStyleGuide/#chap-Red_Hat_Technical_Publications-Writing_Style_Guide-D

"datacenters" or "data centers" depending on context

admin_guide/preventing_node_placement.adoc
```diff
@@ -0,0 +1,101 @@
...
+link:../admin_guide/managing_projects.html#using-node-selectors[create a node
+selector], grouping the US datacentres into `superregion: us` and the APAC
+datacentres into `superregion: apac`.
+
+To prevent developers from overloading a particular region, you can add the
+specified `region` label to the `NodeSelectorLabelBlacklist` section of a master
+configuration. Then, whenever a developer in the located in the US creates a
+Pod, it is deployed onto a Node in one of the regions with the `superregion: us`
+label. If the developer tries to target a specific region for their pod, they
+will receive an error. Howver, if they try again, without the NodeSelector on
```
Miciah: s/Howver/However/

admin_guide/preventing_node_placement.adoc
```diff
@@ -0,0 +1,101 @@
...
+labeled accordingly]. For example, `region: us-east`.
+
+As a cluster administrator, you can create an infrastructure where application
+developers are deploying pods onto the nodes closest to their geographical
+location. You can
+link:../admin_guide/managing_projects.html#using-node-selectors[create a node
+selector], grouping the US datacentres into `superregion: us` and the APAC
+datacentres into `superregion: apac`.
+
+To prevent developers from overloading a particular region, you can add the
```
dobbymoodge: "overloading" might not be the best term here, especially since the scheduler should generally prevent that from happening. I think rephrasing this to something along the lines of "To maintain even loading of resources per-datacenter, you can add..." This encapsulates both the abuse-prevention use case (implied by the term "overloading") and the cost-management use case, which is what I was basing my original example upon.

admin_guide/preventing_node_placement.adoc
```diff
@@ -0,0 +1,101 @@
...
+specified `region` label to the `NodeSelectorLabelBlacklist` section of a master
+configuration. Then, whenever a developer in the located in the US creates a
```
dobbymoodge: syntax: "in the located in the"

admin_guide/preventing_node_placement.adoc
dobbymoodge: typo: "However"

admin_guide/preventing_node_placement.adoc
```diff
@@ -0,0 +1,101 @@
...
+...
+admissionConfig
+  pluginConfig
+    PodNodeConstraints
+      configuration
+        apiversion: v1
+        kind: PodNodeConstraintsConfig
+        nodeSelectorLabelBlacklist:
+          - kubernetes.io/hostname
+          - <labe1_1>
```
dobbymoodge: typo? labe**1**_1 <-- should be lowercase "L"?

bfallonf:
@dobbymoodge Wow, good find on the labe1 typo. 

Thanks for the review. Is there anything else needed for this PR?

In the meantime, @tpoitras @ahardin-rh @tnguyen-rh @adellape available for peer review if anyone wants to make a comment or two.

Thanks, all.

mdshuai:
@bfallonf 
* This feature restrict both nodeName and nodeSelector in pod. need doc for nodeName too. 
* This feature only restrict normal user, normal user can also get the permission by add "create pods/binding" policy.
[some cases](https://url.corp.redhat.com/f9f4729 ) may help you understand.
thanks.

bfallonf:
@mdshuai Yes, I looked at these test cases when writing this. I've added an extra step about ensuring the right label and selector setup exists. However, the test cases you sent me seem to be performing the same process. Is there a difference between using the "NodeSelector" process or the "NodeName" process? They seem to do the same job and there isn't a point in documenting both if this is true.

@dobbymoodge Do you have the answers to this?

Thanks.

admin_guide/preventing_node_placement.adoc
```diff
@@ -0,0 +1,119 @@
...
+
+The `nodeSelectorLabelBlacklist` field of a master configuration file gives you
+control over the labels that certain roles can specify in new
+`nodeSelector` fields of a pod configuration. Users, service accounts, and
+groups that have the *pods/binding* permission can specify any node selector.
+Those without the *pods/binding* permission are prohibited from setting a
+`nodeSelector` for any label that appears in `nodeSelectorLabelBlacklist`.
+
+For example, you might have a cluster consisting of five data centers spread
+across two regions. In the US, "us-east", "us-central", and "us-west"; and
```
tnguyen-rh: Hmm, s/US/U.S./ maybe?

admin_guide/preventing_node_placement.adoc
```diff
@@ -0,0 +1,119 @@
...
+in APAC, "apac-east" and "apac-west". Each node in each geographical region
```
tnguyen-rh: Need to expand APAC on first use.

admin_guide/preventing_node_placement.adoc
```diff
@@ -0,0 +1,119 @@
...
+link:../admin_guide/manage_nodes.html#updating-labels-on-nodes[should be labeled
+accordingly]. For example, `region: us-east`.
+
+As a cluster administrator, you can create an infrastructure where application
+developers are deploying pods onto the nodes closest to their geographical
```
tnguyen-rh: s/are deploying/deploy/

admin_guide/preventing_node_placement.adoc
```diff
@@ -0,0 +1,119 @@
...
+...
+admissionConfig
+  pluginConfig
+    PodNodeConstraints
+      configuration
+        apiversion: v1
+        kind: PodNodeConstraintsConfig
+        nodeSelectorLabelBlacklist:
+          - kubernetes.io/hostname
+          - <label_1>
```
tnguyen-rh: Invalid YAML: Map keys should end with colon.

tnguyen-rh:
@bfallonf 
One real error and a few suggestions.  Otherwise, LGTM.

dobbymoodge:
The NodeName and NodeSelector fields in the Pod spec serve specific
purposes. It's true that there is some conceptual overlap, but they should
be treated as separate things in the documentation.

If you review my comments I have pointed out specifically how the
PodNodeConstraints plugin treats these fields differently.

bfallonf:
@dobbymoodge @mdshuai  This has confused me even more... My understanding of this might be a little skewed. The test cases look like two different ways of performing the same process. 

@dobbymoodge This needs to complete by this week. I'll try come online later tonight (my time) and try to catch you online during the your workday. Maybe talking in real time will help iron out the kinks.

bfallonf:
@dobbymoodge OK. I've got rid of this stuff in its own file, and added it to the bottom of the schedular topic. I've rearranged the example intro stuff, and added the first subsection about  using labels. I also found a link that didn't work in another file, so fixed that. Is there anything else you think this needs? I kept the example in the intro purely about using nodeselectors, as I didn't feel the need to use both options. I'll check again later tonight, and maybe ask questions online if needed.

@tnguyen-rh I did not do your suggested edits. I figured there'd be more after this push...

Thanks!

admin_guide/scheduler.adoc
```diff
@@ -367,3 +367,204 @@ Since the scheduler is a plug-in, it can be replaced in favor of an alternate
...
+[[controlling-pod-placement]]
+== Controlling Pod Placement
+
+As a cluster administrator, you can set a policy to prevent application
+developers with certain roles from targeting specific nodes when scheduling
+pods. 
+
+[IMPORTANT]
+====
+This process involves the *pod/binding* permission
```
dobbymoodge: **pods/binding**

admin_guide/scheduler.adoc
```diff
@@ -367,3 +367,204 @@ Since the scheduler is a plug-in, it can be replaced in favor of an alternate
...
+link:../admin_guide/manage_authorization_policy.html[role], which is needed to
+target particular nodes. The constraint on the use of the `nodeSelector` field
+of a pod configuration is based on the *pods/binding* permission and the
+`nodeSelectorLabelBlacklist` configuration option.
+====
+
+The `nodeSelectorLabelBlacklist` field of a master configuration file gives you
+control over the labels that certain roles can specify in new
```
dobbymoodge: "new" doesn't belong here

dobbymoodge: looking at e.g. https://docs.openshift.com/enterprise/3.0/admin_guide/scheduler.html#configurable-predicates it seems like this could be rewritten like "...certain roles can specify in a pod's `nodeSelector` field."

admin_guide/scheduler.adoc
```diff
@@ -367,3 +367,204 @@ Since the scheduler is a plug-in, it can be replaced in favor of an alternate
...
+Those without the *pods/binding* permission are prohibited from setting a
+`nodeSelector` for any label that appears in `nodeSelectorLabelBlacklist`.
+
+For example, you might have a cluster consisting of five data centers spread
+across two regions. In the US, "us-east", "us-central", and "us-west"; and
+in APAC, "apac-east" and "apac-west". Each node in each geographical region
+link:../admin_guide/manage_nodes.html#updating-labels-on-nodes[should be labeled
+accordingly]. For example, `region: us-east`.
+
+As a cluster administrator, you can create an infrastructure where application
```
dobbymoodge: Maybe change the tone here to make it more clear that this is a hypothetical example, e.g. "As a cluster administrator, you might have an infrastructure where..."

I don't want people to read this and think "well here's how we would set up a geographic constraint for pod deployment!" :smiley:

admin_guide/scheduler.adoc
```diff
@@ -367,3 +367,204 @@ Since the scheduler is a plug-in, it can be replaced in favor of an alternate
...
+location. You can
+link:../admin_guide/managing_projects.html#using-node-selectors[create a node
+selector], grouping the US data centers into `superregion: us` and the APAC data
+centers into `superregion: apac`.
+
+To maintain an even loading of resources per data center, you can add the
+desired `region` label to the `nodeSelectorLabelBlacklist` section of a master
+configuration. Then, whenever a developer located in the US creates a
+pod, it is deployed onto a node in one of the regions with the `superregion: us`
+label. If the developer tries to target a specific region for their pod, they
```
dobbymoodge: "If the developer tries to target a specific region for their pod - for example `region: us-east` - they will receive an error."

admin_guide/scheduler.adoc
```diff
@@ -367,3 +367,204 @@ Since the scheduler is a plug-in, it can be replaced in favor of an alternate
...
+will receive an error. If they try again, without the node selector on
+their pod, it can still be deployed onto the region they tried to target,
+because `superregion: us` is set as the project-level node selector, and those
+nodes are labeled with `superregion: us`.
```
dobbymoodge: "...and nodes labeled `region: us-east` are also labeled `superregion: us`."

admin_guide/preventing_node_placement.adoc
bfallonf: @tnguyen-rh Sorry, not too sure what you mean here. I did just try put this into a YAML validator, but you're right, it failed. Any advice?

bfallonf:
@dobbymoodge Thanks for those suggestions. I did as suggested, and tried to make the the example a little more hypothetical, but I'm not sure what else to do, really. I think the reader should know the deal. If there's anything else, please let me know.

@tnguyen-rh I put in your suggestions as well, and except for the YAML error, it should all be good.

@mdshuai Can I ask for another QA look, please? Let me know if there are any issues. 

Thanks, all!

admin_guide/preventing_node_placement.adoc
tnguyen-rh: @bfallonf
Sorry, i was not specific.  The map keys requiring a terminating colon are:

- `admissionConfig`
- `pluginConfig`
- `PodNodeConstraints`
- `configuration`

For a quick read on fundamental YAML syntax, see also [YAML in a Nutshell](https://access.redhat.com/documentation/en/red-hat-enterprise-linux-atomic-host/7/getting-started-with-containers/chapter-13-yaml-in-a-nutshell) (one of the first writings i did, so as yet unhindered by normal enterprise-y title humbuggery... :smile:).

bfallonf: @tnguyen-rh Ah I getcha. Updated, and I used an online yaml validator. Should be good now. Good eye!

admin_guide/scheduler.adoc
```diff
@@ -367,3 +367,205 @@ Since the scheduler is a plug-in, it can be replaced in favor of an alternate
...
+configuration. Then, whenever a developer located in the U.S. creates a pod, it
+is deployed onto a node in one of the regions with the `superregion: us` label.
+If the developer tries to target a specific region for their pod (for example,
+`region: us-east`), they will receive an error. If they try again, without the
+node selector on their pod, it can still be deployed onto the region they tried
+to target, because `superregion: us` is set as the project-level node selector,
+and nodes labeled `region: us-east` are also labeled `superregion: us`.
+
+[[constraining-pod-placement-labels]]
+=== Constraining Pod Placement Using Node Labels
```
mdshuai: This part should be "Constraining Pod Placement Using Node Name".  it's not related to node label. if you enable this controller, no matter what's the nodeName is it will rejected to create.

admin_guide/scheduler.adoc
```diff
@@ -367,3 +367,205 @@ Since the scheduler is a plug-in, it can be replaced in favor of an alternate
...
+apiVersion: v1
+kind: Pod
+spec:
+ nodeName: <node1>
+----
+
+. Modify the master configuration file
+(*_/etc/origin/master/master-config.yaml_*) in two places:
++
+.. Add `nodeSelectorLabelBlacklist` to the `admissionConfig` section with
```
mdshuai: Restrict nodeName, actually no need add "nodeSelectorLabelBlacklist" part, just enable PodNodeConstraints admission controller is enough. If still add this part, there is no different with the below example case.

admin_guide/scheduler.adoc
```diff
@@ -367,3 +367,205 @@ Since the scheduler is a plug-in, it can be replaced in favor of an alternate
...
++
+----
+apiVersion: v1
+kind: Pod
+spec:
+ nodeName: <node1>
+----
+
+. Modify the master configuration file
+(*_/etc/origin/master/master-config.yaml_*) in two places:
```
mdshuai: The reason to set two places is that k8s part will restrict create pod directly, the origin part will restrict create a deploymentConfig which contain a pod using nodeName/nodeSelector.
https://github.com/openshift/online/issues/68#issuecomment-200491633

mdshuai:
@bfallonf Just a little comment, others LGTM. thanks for your work.

admin_guide/scheduler.adoc
bfallonf: @mdshuai Sorry, I'm not sure what you mean. Should I just delete the everything below `nodeSelectorBlacklist` in the YAML example?

bfallonf:
Finalised this with @mdshuai and put in the suggestions. @tnguyen-rh @adellape @tpoitras @ahardin-rh If there's anything else, let me know. Otherwise, this should be good to merge.

Thanks, all!

ahardin-rh:
@bfallonf LGTM :clap:

mdshuai:
@bfallonf LGTM

bfallonf:
Stuffed up the squash in this PR. Opened #1862 in favour of this one. I'll close this.

PR #1722, opened 2016-03-10
fix route name

0xmichalis:
@tnguyen-rh we missed that:) PTAL


tnguyen-rh:
@kargakis My bad -- was too quick on the merge button.

No revhistory note necessary as this follows on #1717 which has its own.

PR #1723, opened 2016-03-10
Port1936 clarification regarding later errata

sdodson:


sdodson:
@nak3 per your comments at https://bugzilla.redhat.com/show_bug.cgi?id=1293578#c36

nak3:
Awesome! Thank you for your quick update @sdodson

install_config/upgrading/manual_upgrades.adoc
```diff
@@ -934,6 +934,11 @@ security rules allowing access to your node's public IP on port 1936:
 
 If this port is not allowed, the router pod will fail to start after
 installation.
+
+An errata has been released which ensures that the router liveness probe uses
+localhost rather than the public interface. This means you no longer need to open
+the firewall rule above. Please see https://access.redhat.com/errata/product/290/ver=3.1/rhel---7/x86_64/RHBA-2016:0293[RHBA-2016:0293] for details
+of that errata.
```
sdodson: I'm not sure if this should be left in or if we should simply removed the entire section. If they're upgrading right now their likelihood of hitting the issue here is near zero as the current packages don't have this problem.

vikram-redhat:
@tpoitras - can you take a look today and mark for inclusion in next release?

vikram-redhat:
Marking it for a future release as @nak3  was looking at the followup PR: #1734

adellape:
@vikram-redhat For something that needs to go into Next Release (for the current docs) but just is blocked on a follow-up, we should just keep the `to_follow` label, not switch to Future Release (which suggests a later release of OpenShift).

@tpoitras I've added the `to_followup` label back until https://github.com/openshift/openshift-docs/pull/1734 merges.

adellape:
Revhistory in https://github.com/openshift/openshift-docs/pull/1734.

PR #1736, opened 2016-03-14
Document new registry role policy

aweiteka:


adellape:
LGTM.

Looks like this was added in https://github.com/openshift/origin/pull/6963 and https://github.com/openshift/origin/pull/7760, both of which are in OSE 3.2.

[rev_history]
|link:../admin_guide/manage_authorization_policy.html[Managing Authorization Policies]
|Added new registry roles to output in the link:../admin_guide/manage_authorization_policy.html#viewing-cluster-policy[Viewing Cluster Policy] section.
%

PR #1745, opened 2016-03-15
update binary build doc

gabemontero:
@bparees @rhcarvalho @mfojtik @csrwng PTAL

Note, I tried to cross reference the document with what I saw in the code, in particular https://github.com/openshift/origin/blob/master/pkg/build/builder/source.go#L44-L101

As such, it seemed like dockerfile's could be used even if asFile was set.  That is counter to what I at least interpreted the prior version of the documentation implying.  So let's confirm/deny that along with assessing the changes' attempt to make things clearer.

### DIFF
--- a/dev_guide/builds.adoc
+++ b/dev_guide/builds.adoc
@@ -33,17 +33,21 @@ link:#docker-strategy-options[options])
 (link:../architecture/core_concepts/builds_and_image_streams.html#custom-build[description],
 link:#custom-strategy-options[options])
 
-And there are three types of build source:
+And there are four types of build source:
 
 - link:#source-code[Git]
 - link:#dockerfile-source[Dockerfile]
+- link:#image-source[Image]
 - link:#binary-source[Binary]
 
 It is up to each build strategy to consider or ignore a certain type of source,
 as well as to determine how it is to be used.
 
-Binary and Git are mutually exclusive source types, while Dockerfile can be used
-by itself or together with Git and Binary.
+Binary and Git are mutually exclusive source types, while Dockerfile and Image can be used
+by themselves, with each other, or together with either Git or Binary.
+
+Also, the Binary build source type is unique from the other options in how it is
+specified to the system.  Exactly how is detailed link:#binary-source[below].
 
 
 [[defining-a-buildconfig]]
@@ -846,42 +850,47 @@ source:
 
 == Binary Source
 
-When the `*BuildConfig.spec.source.type*` is `*Binary*`, the build will expect a
-binary as input, and an inline Dockerfile is optional.
+Streaming content in binary format from a local file system to the builder is called a `*binary type build*`.
+The corresponding value of `*BuildConfig.spec.source.type*` is `*Binary*` for such builds.
 
-The binary is generally assumed to be a tar, gzipped tar, or zip file depending
-on the strategy. For `*Docker*` builds, this is the build context and an
-optional Dockerfile may be specified to override any Dockerfile in the build
-context. For `*Source*` builds, this is assumed to be an archive as described
-above. For `*Source*` and `*Docker*` builds, if `*binary.asFile*` is set the
-build context will consist of a single file named by the value of `*binary.asFile*`. 
-The `*contextDir*` field may be used when an archive is provided. Custom builds will 
-receive this binary as input on standard input (`stdin`).
+This source type is unique in that it is leveraged solely based on your use of the `oc start-build`.
 
-A binary source potentially extracts content (if `*asFile*` is not set), in which case `*contextDir*`
-allows changing to a subdirectory within the content before the build executes.
-
-The source definition is part of the `*spec*` section in the `*BuildConfig*`:
-
+[NOTE]
 ====
-[source,yaml]
-----
-source:
-  type: "Binary"
-  binary: <1>
-    asFile: "webapp.war" <2>
-  contextDir: "app/dir" <3>
-  dockerfile: "FROM centos:7\nRUN yum install -y httpd" <4>
-----
-<1> The `*binary*` field specifies the details of the binary source.
-<2> The `*asFile*` field specifies the name of the file that will be created with
-the binary contents.
-<3> The `*contextDir*` field specifies a subdirectory within the extracted contents of a
-binary archive.
-<4> If the optional `*dockerfile*` field is provided, it should be a string
-containing an inline Dockerfile that potentially replaces one within the
-contents of the binary archive.
+Since binary type builds require content to be streamed from the local file system, it is not possible
+to automatically trigger a binary type build (e.g. via an image change trigger) because the binary files could
+not be provided. Similarly, you cannot launch binary type builds from the web console.
 ====
+
+To utilize binary builds, invoke `oc start-build` with one of these options:
+
+1) `--from-file`:  The contents of the file you specify will be sent as a binary stream to the builder.  Then
+the builder will store the data in a file with the same name at the top of the build context.
+
+2) `--from-dir` and `--from-repo`:  The contents will be archived and sent as a binary stream to the builder.
+Then the builder will extract the contents of the archive within the build context directory.
+
+In each of the above cases:
+
+1) If your `*BuildConfig*` already has a `*Binary*` source type defined, it will effectively be ignored and replaced by what the
+client sends.
+
+2) If your `*BuildConfig*` has a `*Git*` source type defined, it will be dynamically disabled, since `*Binary*` and `*Git*` are
+mutually exclusive, and the data in the binary stream provided to the builder will take precedence.
+
+When using `oc new-build --binary=true`, the command ensures that the restrictions associated with binary builds are enforced.
+The resulting `*BuildConfig*` will have a source type of `*Binary*`, meaning that the only valid way to run a build for this `*BuildConfig*`
+will be to use `oc start-build` with one of the `--from` options to provide the requisite binary data.
+
+The `*dockerfile*` and `*contextDir*` link:#source-code[source options] have special meaning with binary builds.
+
+The `*dockerfile*` can be used with any form of `*Binary*` build source.  If `*dockerfile*` is used and the binary stream
+is an archive, its contents serve as a replacement Dockerfile to any Dockerfile in the archive.  If `*dockerfile*` is used, and if the
+`--from-file` argument is used, and the file argument is named "Dockerfile", the value from `*dockerfile*` replaces the value from
+the binary stream. 
+
+In the case of the binary stream encapsulating extracted archive content, the value of the `*contextDir*` field is interpreted as
+a subdirectory within the archive, and if valid, the builder changes into that subdirectory before executing the build.
 
 [[image-source]]
 


dev_guide/builds.adoc
```diff
@@ -846,20 +846,28 @@ source:
...
-A binary source potentially extracts content (if `*asFile*` is not set), in which case `*contextDir*`
-allows changing to a subdirectory within the content before the build executes.
+When the `*BuildConfig.spec.source.type*` is `*Binary*`, the source content will be sent
+as input to the build via a binary stream and  `oc start-build` should be executed with either
+the `--from-dir`, `--from-file`, or `--from-repo` options.
+
+The binary stream will be processed in one of two ways:
+
+1) If `*binary.asFile*` is set, then the bytes from the binary stream will be considered a single file.  The binary
+stream will be stored in a file on the builder's file system with the name as specified as `*binary.asFile*` for the builder to use.
```
bparees: it's not clear to me how this works with --from-dir or --from-repo.  Does it mean that if i specify `binary.asFile=foo` and i invoke `oc start-build --from-dir /some/dir`, that i'll end up with an archive file named "foo" in my build context?  If so, what's the archive format?

or does binary.asFile really only make sense when using the --from-file argument?

dev_guide/builds.adoc
```diff
@@ -846,20 +846,28 @@ source:
...
+
+2) Otherwise, the bytes from the binary stream are assumed to be an archive in either zip, tar, gzipped tar format, and the
+contents of the stream will be extracted onto the builder's file system for the builder to use.  The stream is dynamically analyzed
+to determine the precise archive format, and then the specific archive's formatting rules are used to extract to content.
+
+Note:  Custom builds receive the binary stream on standard input (`stdin`).
+
+The resulting single file (via `*binary.asFile*`) or extracted directory tree of the archive serve as the build context for the builder.
+
+Irregardless of the contents of the binary stream or the build strategy, the `*dockerfile*` field can optionallly be used.  If used, its contents serve as a replacement Dockerfile to any Dockerfile found in
```
bparees: s/Irregardless/Regardless/

bparees:
> As such, it seemed like dockerfile's could be used even if asFile was set. That is counter to what I at least interpreted the prior version of the documentation implying. 

yes that's correct, but the example currently in the docs does illustrate that (the sample buildconfig shows a binary build with an asFile value and a dockerfile value).

in any case i think your added text helps clarify the behavior.

dev_guide/builds.adoc
gabemontero: Yeah, I just included any of the `oc start-build` options that mentioned binary input.  Here is a copy paste for those three from the `oc help start-build` output:

`      --from-dir='': A directory to archive and use as the binary input for a build.`
`      --from-file='': A file to use as the binary input for the build; example a pom.xml or Dockerfile. Will be the only file in the build source.`
`      --from-repo='': The path to a local source code repository to use as the binary input for a build.`

I assumed (and then confirmed by looking at https://github.com/openshift/origin/blob/master/pkg/cmd/cli/cmd/startbuild.go#L214 after seeing your question) that the from-dir and from-repo translate into the client sending an archive of those over the binary stream, and yes, asFile only makes sense with --from-file (though I could see --from-file not needing asFile if say the one file was a pom.xml or Dockerfile).

That sound right to everybody?

dev_guide/builds.adoc
gabemontero: will push update in a sec

gabemontero:
pushed Regardless fix, cool deal on the clarification on the dockerfile applicability.

the `oc start-build` options examination still ongoing, along with any new comments that come in

dev_guide/builds.adoc
bparees: well it sounds right but i'm still not clear on what the behavior actually is if you do such a thing.

gabemontero: ah, meaning setting asFile and using --from-dir ... yeah, unless the whoever created the binary build stuff (don't remember who it was) can chime in, I'd have to look at the code some more and/or actually try it out.

gabemontero: OK ... I thought I would look at some of the tests to confirm the interpretation of the behavior.

I found https://github.com/openshift/origin/blob/master/test/extended/builds/start.go#L85-L146

Turns out, you can use --from-dir/from-file/from-repo even when you do *NOT* have a binary source type in the BC.  Circling back and going through the `oc start-build` code path, using --from* results in a BinaryBuildRequest, and by the time it lands on the receiving side, that object is translated into a BinaryBuildSource (a la the source type Binary) ... so the --from* options are a dynamic way to turn on binary builds.

So I wondered about the source.type field in general, and discovered that it is removed in https://github.com/openshift/origin/blob/master/pkg/build/api/types.go, but still exists in the older v1 version, https://github.com/openshift/origin/blob/master/pkg/build/api/v1/types.go.

So I revisited https://github.com/openshift/origin/blob/master/pkg/build/builder/source.go#L44-L101, and sure enough, it in fact checks all the different *types* (binary, git, image, dockerfile) and essentially will overlay them from binary, to git, to images, to dockerfile as part of retrieving the *source*.

So  this implies to me that post v1, you can have multiple source types in a build config (if true, and if that was well known, it has previously escaped my awareness).

So ... my updated take on the net for the doc:
0) not with this PR, but if the removal of build.source.type becomes the next version (and is consumed by etcd I suppose), we have references to type that need to be removed.
1) If type binary.asFile is specified in the build config, and none of the --from* options are used with start-build, then that file must exist in the builder image
2) Otherwise, --from-file must be specified on the start-build, and the data comes over stdin
3) a binary build type without an as-file (that field is marked omitemtpy), and without the start-build having a --from* option, means either the source is already baked in the builder image (is that at all an interesting scenario ??), or it is totally not useful.
4) so there is essentially a matrix of behavior permutations with the bc source.binary obj and the `oc start-build --from-*` options ... 

I'll go ahead and start updating the PR based on my above interpretation, but of course will adjust as further responses occur.

gabemontero: Yeah, in fact, the documentation also has a blurb that "Binary and Git are mutually exclusive source types" ... from a BuildConfig perspective, perhaps, but with the `--from*` options, that does not seem entirely true, at least from a broader perspective.  Note the extended tests at https://github.com/openshift/origin/blob/master/test/extended/builds/start.go#L85-L146

gabemontero:
OK, essentially rewrote the section, based on ongoing review of the client and server side code paths.

Aside from continued review in this PR, perhaps a group discussion in our next scrum is warranted.

bparees:
so after some experimentation and @gabemontero's explanations of the behavior, i've come to a new understanding of how this feature works.

A user would never explicitly define a buildconfig with an "asFile" value (if you do set one, it doesn't matter, it will get removed or overwritten).

if you run start-build --from-file foo, then the resulting build will have an asFile value with the name of the file you provided (foo).  Even if the original build did not have an asFile field, or had a different value in that field.

if you run start-build --from-dir or --from-repo, there will be no asFile field.  Even if the original BC had an asFile field defined.

So perhaps we need to revisit how we're describing this feature to users, because it's not "just another strategy" they might use when writing a new BuildConfig.  It's implicit behavior they get when using specific start-build arguments, regardless of how their BuildConfig is defined.

Will take another pass through the PR and see how it flows given this new understanding.

gabemontero:
On Thu, Mar 17, 2016 at 9:58 AM, Ben Parees <notifications@github.com>
wrote:

> so after some experimentation and @gabemontero
> <https://github.com/gabemontero>'s explanations of the behavior, i've
> come to a new understanding of how this feature works.
>
> A user would never explicitly define a buildconfig with an "asFile" value
> (if you do set one, it doesn't matter, it will get removed or overwritten).
>
Certainly if that file existed in the builder image beforehand, it would
get replaced with the binary stream coming over stdin, if that is what you
are referring to.

> if you run start-build --from-file foo, then the resulting build will have
> an asFile value with the name of the file you provided (foo). Even if the
> original build did not have an asFile field, or had a different value in
> that field.
>
Yep, as I mentioned in my earlier comment, a binary build "type" is
dynamically created - doesn't matter what was in the BC.

> if you run start-build --from-dir or --from-repo, there will be no asFile
> field. Even if the original BC had an asFile field defined.
>
If your experiments showed that behavior, I did not pick up on that from
code examination....I'll take another look.

> So perhaps we need to revisit how we're describing this feature to users,
> because it's not "just another strategy" they might use when writing a new
> BuildConfig. It's implicit behavior they get when using specific
> start-build arguments, regardless of how their BuildConfig is defined.
>
> Will take another pass through the PR and see how it flows given this new
> understanding.
>
I did go as far as removing the BC type and just describing the start-build
options, though I did start down the path of describing how it interplays
with git, images, dockerfile.  Pending inline comments from you on the
existing content, I can get started on that level of refactor.

bparees:
Ok, I hate to say it, but i think we need to rework this again :(

I think for the Binary Source section, we should walk users through how each of the --from-dir, --from-repo, and --from-file commands work, in conjunction with an existing BuildConfig (which may or may not have a Git source, and may or may not have an in-line Dockerfile).

and then we can get rid of (or move into this section) the parts of this:

bparees:
So it'll be a bit different from the other sections because it's not going to illustrate yaml, etc.  it's all about how to invoke it from the cli side and what will happen.

gabemontero:
Yep ... that is where I was headed with my last comment.

bparees:
> Certainly if that file existed in the builder image beforehand, it would get replaced with the binary stream coming over stdin, if that is what you are referring to.

no, i was referring to the asField field value in the BuildConfig definition, not the actual file in the image.

> If your experiments showed that behavior, I did not pick up on that from code examination....I'll take another look.

yeah, they did.

bparees:
> Pending inline comments from you on the existing content, I can get started on that level of refactor.

yeah, let's do that... sorry this is turning into a mess, but we've learned some useful things :)

gabemontero:
OK, I wouldn't call it perfect, but I've pushed what is an initial draft of the new direction.  Hopefully it is close enough that we can iterate on it and move toward the final version.

dev_guide/builds.adoc
```diff
@@ -846,42 +851,43 @@ source:
...
-the binary contents.
-<3> The `*contextDir*` field specifies a subdirectory within the extracted contents of a
-binary archive.
-<4> If the optional `*dockerfile*` field is provided, it should be a string
-containing an inline Dockerfile that potentially replaces one within the
-contents of the binary archive.
-====
+2) `--from-dir` and `--from-repo`:  The contents will be archived (either zip, tar, or gzipped tar based on what
+is available on the client system) and sent as a binary stream to the builder.  Then the builder will extract
+the contents of the archive off of the top of the build context.
```
bparees: s/off of the top of/within/
s/build context/build context directory/

dev_guide/builds.adoc
```diff
@@ -846,42 +851,43 @@ source:
...
+
+In each of the above cases:
+
+1) If your `*BuildConfig*` already has a `*Binary*` source type defined, it will effectively be ignored and replaced by what the
+client sends.
+
+2) If your `*BuildConfig*` has a `*Git*` source type defined, it will be dynamically disabled, since `*Binary*` and `*Git*` are
+mutually exclusive, and the data in the binary stream provided to the builder will take precedence.
+
+3) Triggers will be ignored.
```
bparees: Probably don't need to say this, if you're in a path were you're running "oc start-build" triggers aren't relevant.

but somewhere in here we should probably have a note that says something like "Note: binary type builds require content to be streamed from the local file system.  As such, it is not possible to automatically trigger a binary type build (eg via an image change trigger) because the binary files could not be provided.  Similarly you cannot launch binary type builds from the web console."

dev_guide/builds.adoc
```diff
@@ -846,42 +851,43 @@ source:
...
+
+When using `oc new-build --binary=true`, the command essentially ensures that any of the restrictions related with binary builds
```
bparees: s/essentially//

dev_guide/builds.adoc
```diff
@@ -846,42 +851,43 @@ source:
...
+are enforced.  And technically source type of `*Binary*` will be created.  But it is really an descriptive placeholder.  You still have to
```
bparees: "When using `oc new-build --binary=true`, the command ensures that the restrictions associated with binary builds are enforced.  The resulting `*BuildConfig*` will have a source type of `*Binary*`, meaning that the only valid way to run a build for this `*BuildConfig*` will be to use `*oc start-build*` with one of the `*--from*` options to provide the requisite binary data."

dev_guide/builds.adoc
```diff
@@ -846,42 +851,43 @@ source:
...
+use `oc start-build` with one of the "--from" options to actually provide the binary data.
+
+When any form of `*Binary*` build source is being used, the `*dockerfile*` field can optionallly be used.  If used, its contents serve as a replacement Dockerfile to any Dockerfile found in
+the binary stream if the binary stream is an archive.  If the binary stream is a single file, the `*dockerfile*` field contents are stored in a Dockerfile stored at the top of the build context, and it among
+other things can access the file specified by `*binary.asFile*`.  In the case of the binary stream encapsulating extracted archive content, the value of the `*contextDir*` field is interpreted as a subdirectory
```
bparees: "When any form of `*Binary*` build source is being used, the `*dockerfile*` field can optionallly be used.  If used, its contents serve as a replacement Dockerfile to any Dockerfile found in
the binary stream if the binary stream is an archive, or the single file if the --from-file argument is used and the file argument is named "Dockerfile".  The binary content (either the extracted archive or the single file) will be placed in the context directory used by the Docker build which is performed using the Dockerfile.  

In the case of the binary stream encapsulating extracted archive content, the value of the `*contextDir*` field is interpreted as a subdirectory within the archive, and if valid, the builder changes into that subdirectory before executing the build."

bparees:
couple suggested rewordings, but i think it's looking better.  Maybe we should get @gshipley to take a look since I think he was having trouble understanding binary builds from the existing doc.

dev_guide/builds.adoc
gabemontero: part of next push

dev_guide/builds.adoc
gabemontero: Removed 3), and added your note with a [NOTE] ==== .... ===== just prior to the explanation of `oc start-build`

dev_guide/builds.adoc
gabemontero: part of next push

gabemontero: nevermind ... you reworded the entire block with your next comment.

gabemontero:
OK,  I believe I've captured the rewordings and pushed them.

dev_guide/builds.adoc
```diff
@@ -33,17 +33,22 @@ link:#docker-strategy-options[options])
...
 - link:#dockerfile-source[Dockerfile]
+- link:#image-source[Image]
 - link:#binary-source[Binary]
 
 It is up to each build strategy to consider or ignore a certain type of source,
 as well as to determine how it is to be used.
 
 Binary and Git are mutually exclusive source types, while Dockerfile can be used
-by itself or together with Git and Binary.
+by itself or together with Git and Binary.  Image is intended to together with the
```
bparees: Image is doesn't have to be used w/ the other options, you can have a build that takes its entire content from image sources (also you're missing a word or two up there)

dev_guide/builds.adoc
```diff
@@ -33,17 +33,22 @@ link:#docker-strategy-options[options])
...
+other options as a way to augment the files available to the build.
+
+Also, the Binary build source type is unique from the other options in how it is
+specified to the system.  Exactly how will be detailed link:#binary-source[below].
```
bparees: s/will be/is/

dev_guide/builds.adoc
```diff
@@ -846,43 +851,47 @@ source:
...
-<3> The `*contextDir*` field specifies a subdirectory within the extracted contents of a
-binary archive.
-<4> If the optional `*dockerfile*` field is provided, it should be a string
-containing an inline Dockerfile that potentially replaces one within the
-contents of the binary archive.
+Binary type builds require content to be streamed from the local file system. As such, it is not possible to automatically trigger a binary type build (eg via an image change trigger)
+because the binary files could not be provided. Similarly you cannot launch binary type builds from the web console.
 ====
 
+With `oc start-build` consider these options:
```
bparees: "To utilize binary builds, invoke `oc start-build` with one of these options:"

dev_guide/builds.adoc
```diff
@@ -846,43 +851,47 @@ source:
...
+client sends.
+
+2) If your `*BuildConfig*` has a `*Git*` source type defined, it will be dynamically disabled, since `*Binary*` and `*Git*` are
+mutually exclusive, and the data in the binary stream provided to the builder will take precedence.
+
+When using `oc new-build --binary=true`, the command ensures that the restrictions associated with binary builds are enforced.
+The resulting `*BuildConfig*` will have a source type of `*Binary*`, meaning that the only valid way to run a build for this `*BuildConfig*`
+will be to use `oc start-build` with one of the `--from` options to provide the requisite binary data.
+
+When any form of `*Binary*` build source is being used, the `*dockerfile*` field can optionallly be used. If used, its contents serve as a replacement Dockerfile to any Dockerfile found in
```
bparees: extra "l" in optionally

dev_guide/builds.adoc
gabemontero: ok, that is what i thought after looking at the code, but could have swore during our scrum discussion that someone declared it had to be used on conjunction with one of the others ... thx for the confirm

bparees: to be fair i haven't confirmed in the code myself, but logically it should not require any other inputs, so i'd consider it a bug if it does.

gabemontero:
edits pushed

bparees:
lgtm.  @mfojtik @rhcarvalho @php-coder @PI-Victor @mnagy would like to have a least one of you give this a read as well and see if it makes sense to you since, as before, i'm too familiar with this feature to review the doc objectively.

dev_guide/builds.adoc
```diff
@@ -846,43 +850,47 @@ source:
...
-
-The binary is generally assumed to be a tar, gzipped tar, or zip file depending
-on the strategy. For `*Docker*` builds, this is the build context and an
-optional Dockerfile may be specified to override any Dockerfile in the build
-context. For `*Source*` builds, this is assumed to be an archive as described
-above. For `*Source*` and `*Docker*` builds, if `*binary.asFile*` is set the
-build context will consist of a single file named by the value of `*binary.asFile*`. 
-The `*contextDir*` field may be used when an archive is provided. Custom builds will 
-receive this binary as input on standard input (`stdin`).
+`*Binary*` is a valid value for `*BuildConfig.spec.source.type*`, but is it not something you will
```
PI-Victor: `but is it not something you will`  *`it is not something`
also `valid value` sounds a bit funky it could be `it is a value that is valid for` or rather in this context `it is a value that could be valid for`

dev_guide/builds.adoc
```diff
@@ -846,43 +850,47 @@ source:
...
-----
-<1> The `*binary*` field specifies the details of the binary source.
-<2> The `*asFile*` field specifies the name of the file that will be created with
-the binary contents.
-<3> The `*contextDir*` field specifies a subdirectory within the extracted contents of a
-binary archive.
-<4> If the optional `*dockerfile*` field is provided, it should be a string
-containing an inline Dockerfile that potentially replaces one within the
-contents of the binary archive.
+Binary type builds require content to be streamed from the local file system. As such, it is not possible to automatically trigger a binary type build (eg via an image change trigger)
```
PI-Victor: this line leads me astray a bit and i hope i didn't misunderstand it.
what about something alone the line: `As such, it is not possible to automatically trigger a binary type build (eg via an image change trigger) or launch it from the web console because the binary source needs to be specified at **build time`
**not sure if `build time` is correct but i think it gives a bit more context.

dev_guide/builds.adoc
```diff
@@ -846,43 +850,47 @@ source:
...
+client sends.
+
+2) If your `*BuildConfig*` has a `*Git*` source type defined, it will be dynamically disabled, since `*Binary*` and `*Git*` are
+mutually exclusive, and the data in the binary stream provided to the builder will take precedence.
+
+When using `oc new-build --binary=true`, the command ensures that the restrictions associated with binary builds are enforced.
+The resulting `*BuildConfig*` will have a source type of `*Binary*`, meaning that the only valid way to run a build for this `*BuildConfig*`
+will be to use `oc start-build` with one of the `--from` options to provide the requisite binary data.
+
+When any form of `*Binary*` build source is being used, the `*dockerfile*` field can optionally be used. If used, its contents serve as a replacement Dockerfile to any Dockerfile found in
```
PI-Victor: this paragraph is mostly about what the `*dockerfile*` is/does so my suggestion would be to start with something similar to:
`The *dockerfile* is an optional field that can be used with any form of *Binary* build source, its contents serving as a replacement...`

dev_guide/builds.adoc
```diff
@@ -846,43 +850,47 @@ source:
...
+the binary stream if the binary stream is an archive, or the single file if the `--from-file` argument is used and the file argument is named "Dockerfile". The binary content (either the extracted
```
PI-Victor: maybe this is just me, found it hard at first read but what about `or it will replace the single file if the...` or maybe something else entirely because it sounds like it loses context when it gets to the `or` part.

dev_guide/builds.adoc
```diff
@@ -846,43 +850,47 @@ source:
...
+2) If your `*BuildConfig*` has a `*Git*` source type defined, it will be dynamically disabled, since `*Binary*` and `*Git*` are
+mutually exclusive, and the data in the binary stream provided to the builder will take precedence.
+
+When using `oc new-build --binary=true`, the command ensures that the restrictions associated with binary builds are enforced.
+The resulting `*BuildConfig*` will have a source type of `*Binary*`, meaning that the only valid way to run a build for this `*BuildConfig*`
+will be to use `oc start-build` with one of the `--from` options to provide the requisite binary data.
+
+When any form of `*Binary*` build source is being used, the `*dockerfile*` field can optionally be used. If used, its contents serve as a replacement Dockerfile to any Dockerfile found in
+the binary stream if the binary stream is an archive, or the single file if the `--from-file` argument is used and the file argument is named "Dockerfile". The binary content (either the extracted 
+archive or the single file) will be placed in the context directory used by the Docker build which is performed using the Dockerfile.
```
PI-Victor: what about `The binary content (either the extracted archive or the single file) will be placed in the context directory used by the Docker build according to the Dockerfile` ?

PI-Victor:
made a few suggestions to some phrases that i think are formulated tricky and make for a hard read.
other from that i learned through reviewing :-) LGTM

dev_guide/builds.adoc
gabemontero: Yeah, as we talked about in scrum on Friday, Binary builds are a bit funky, and certainly that funkiness is more apparent here than perhaps in other parts of this description.  This type is unique in that it is truly optional in the need to specify it.  But let's see if I can convey that more clearly.

I'm not sure the suggestion necessarily works either, but I think I have a thought on how to combine the first two sentences.  I'll push that shortly, and let me know if it seems clearer to you.

thanks.

dev_guide/builds.adoc
gabemontero: part of next push

dev_guide/builds.adoc
gabemontero: Yeah, I agree that it was clunky.  I reworked, though differently than the suggestion.  The rework included pruning parts that in hindsight seem redundant.  Let's see if folks like it better.

gabemontero:
The @PI-Victor inspired edits have been pushed.  Thanks!

bparees:
still lgtm

dev_guide/builds.adoc
```diff
@@ -33,17 +33,21 @@ link:#docker-strategy-options[options])
...
 It is up to each build strategy to consider or ignore a certain type of source,
 as well as to determine how it is to be used.
 
-Binary and Git are mutually exclusive source types, while Dockerfile can be used
-by itself or together with Git and Binary.
+Binary and Git are mutually exclusive source types, while Dockerfile and Image can be used
+by themselves, with each other, or together with either Git or Binary.
+
+Also, the Binary build source type is unique from the other options in how it is
+specified to the system.  Exactly how is detailed link:#binary-source[below].
```
rhcarvalho: @gabemontero, curious reader here; is the double spacing in the beginning of the sentence intentional or accidental? (`.--Exactly`)
If intentional, what is the purpose, is there some meaning to it?

If it's a typo, then it needs to be fixed in other places below as well.

dev_guide/builds.adoc
```diff
@@ -33,17 +33,21 @@ link:#docker-strategy-options[options])
 (link:../architecture/core_concepts/builds_and_image_streams.html#custom-build[description],
 link:#custom-strategy-options[options])
 
-And there are three types of build source:
+And there are four types of build source:
 
 - link:#source-code[Git]
 - link:#dockerfile-source[Dockerfile]
+- link:#image-source[Image]
```
rhcarvalho: Good catch! :tada:

dev_guide/builds.adoc
```diff
@@ -846,43 +850,44 @@ source:
...
-The `*contextDir*` field may be used when an archive is provided. Custom builds will 
-receive this binary as input on standard input (`stdin`).
+Streaming content in binary format from a local file system to the builder is called a `*binary type build*`.
+The corresponding value of `*BuildConfig.spec.source.type*` is `*Binary*` for such builds.
 
-A binary source potentially extracts content (if `*asFile*` is not set), in which case `*contextDir*`
-allows changing to a subdirectory within the content before the build executes.
-
-The source definition is part of the `*spec*` section in the `*BuildConfig*`:
+This source type is unique in that it is leveraged soley based on your use of the `oc start-build`.
```
rhcarvalho: s/soley/solely?

dev_guide/builds.adoc
```diff
@@ -846,43 +850,44 @@ source:
...
-----
-<1> The `*binary*` field specifies the details of the binary source.
-<2> The `*asFile*` field specifies the name of the file that will be created with
-the binary contents.
-<3> The `*contextDir*` field specifies a subdirectory within the extracted contents of a
-binary archive.
-<4> If the optional `*dockerfile*` field is provided, it should be a string
-containing an inline Dockerfile that potentially replaces one within the
-contents of the binary archive.
+Since binary type builds require content to be streamed from the local file system, it is not possible to automatically trigger a binary type build (eg via an image change trigger)
```
rhcarvalho: I'm always referring to [this page](http://www.wikihow.com/Use-%22i.e.%22-Versus-%22e.g.%22) when I use "e.g." and "i.e." for I'm not a native speaker. According to that, it should be written with dots and a following comma: "e.g., via an image change trigger". Any native could confirm/refute that?

dev_guide/builds.adoc
```diff
@@ -846,43 +850,44 @@ source:
...
+because the binary files could not be provided. Similarly you cannot launch binary type builds from the web console.
```
rhcarvalho: "Similarly, you cannot...", missing comma?

dev_guide/builds.adoc
```diff
@@ -846,43 +850,44 @@ source:
...
 
+To utilize binary builds, invoke `oc start-build` with one of these options:
+
+1) `--from-file`:  The contents of the file you specify will be sent as a binary stream to the builder.  Then
+the builder will store the data in a file with the same name at the top of the build context (this internally maps
+to the `*binary.asFile*` field of the `*Binary*` source type).
+
+2) `--from-dir` and `--from-repo`:  The contents will be archived (either zip, tar, or gzipped tar based on what
+is available on the client system) and sent as a binary stream to the builder.  Then the builder will extract
+the contents of the archive within the build context directory.
```
rhcarvalho: Does the format depend solely on the client? Does the builder need to have support for one of those for successfully extracting the payload, or the extraction is handled by the platform?

dev_guide/builds.adoc
```diff
@@ -846,43 +850,44 @@ source:
...
-contents of the binary archive.
+Since binary type builds require content to be streamed from the local file system, it is not possible to automatically trigger a binary type build (eg via an image change trigger)
+because the binary files could not be provided. Similarly you cannot launch binary type builds from the web console.
 ====
 
+To utilize binary builds, invoke `oc start-build` with one of these options:
+
+1) `--from-file`:  The contents of the file you specify will be sent as a binary stream to the builder.  Then
+the builder will store the data in a file with the same name at the top of the build context (this internally maps
+to the `*binary.asFile*` field of the `*Binary*` source type).
```
rhcarvalho: Is it possible to use a directory as input and somehow make it a file (.zip, .tar, .tar.gz) in the target image? Does `binary.asFile` require the path to be a regular file?

dev_guide/builds.adoc
```diff
@@ -846,43 +850,44 @@ source:
...
+client sends.
+
+2) If your `*BuildConfig*` has a `*Git*` source type defined, it will be dynamically disabled, since `*Binary*` and `*Git*` are
+mutually exclusive, and the data in the binary stream provided to the builder will take precedence.
+
+When using `oc new-build --binary=true`, the command ensures that the restrictions associated with binary builds are enforced.
+The resulting `*BuildConfig*` will have a source type of `*Binary*`, meaning that the only valid way to run a build for this `*BuildConfig*`
+will be to use `oc start-build` with one of the `--from` options to provide the requisite binary data.
+
+The optional *dockerfile* can be used with any form of *Binary* build source.  If *dockerfile* is used and the binary stream is an archive, its contents serve as a replacement Dockerfile to any Dockerfile in the archive.  If *dockerfile* is used, and if the `--from-file` argument is used, and the file argument is named "Dockerfile", the value from *dockerfile* replaces the value from the binary stream.
```
rhcarvalho: Did you mean

```
`*dockerfile*`
```

and 

```
`*Binary*`
```

?

Now that we're deleting the YAML example, won't the reference to a "dockerfile" field be missing context?

dev_guide/builds.adoc
```diff
@@ -846,43 +850,44 @@ source:
...
+mutually exclusive, and the data in the binary stream provided to the builder will take precedence.
+
+When using `oc new-build --binary=true`, the command ensures that the restrictions associated with binary builds are enforced.
+The resulting `*BuildConfig*` will have a source type of `*Binary*`, meaning that the only valid way to run a build for this `*BuildConfig*`
+will be to use `oc start-build` with one of the `--from` options to provide the requisite binary data.
+
+The optional *dockerfile* can be used with any form of *Binary* build source.  If *dockerfile* is used and the binary stream is an archive, its contents serve as a replacement Dockerfile to any Dockerfile in the archive.  If *dockerfile* is used, and if the `--from-file` argument is used, and the file argument is named "Dockerfile", the value from *dockerfile* replaces the value from the binary stream. 
+
+In the case of the binary stream encapsulating extracted archive content, the value of the `*contextDir*` field is interpreted as a subdirectory within the archive, and if valid, the builder changes
```
rhcarvalho: `contextDir` here may also be missing context without the code example?

Honestly, this whole sentence gave my brain a hard time! It left me with a question, what happens if the path in *not* valid? Build fails? We don't change into that directory and proceed as if `contextDir` was empty?

I believe `contextDir` must be a relative path? Didn't see this mentioned.

rhcarvalho:
If using binary builds requires `oc start-build --from-*`, when would one create a build with type binary, either via template of `oc new-build --binary`?
Is it just to avoid having a fake git source repo?

If we'd drop the requirement for having a build source, then perhaps there would be less need to create a BC with type binary -- one could just create a BC with no source and perform binary builds.

rhcarvalho:
Is there any special consideration about the builder (custom, s2i or Docker) to be able to handle binary builds, or is it transparently handled by the platform?

rhcarvalho:
Since we have no written standards that I'm aware of, I feel bad for mentioning this, but most files follow some "unspoken standard" of wrapping lines at 80 or so columns.
When I'm making docs changes I normally make my editor reflow my paragraphs to fit that column width.
Perhaps I'm mistaken and there are some oficial formatting rules?

rhcarvalho:
> Since we have no written standards that I'm aware of,

FWIW, there's this:

https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc

But no mention to max column width in particular.

dev_guide/builds.adoc
mfojtik: it is handled by platform.

rhcarvalho:
@gabemontero thanks for doing this review on the docs, awesome stuff! Sorry for coming late in the game with comments and questions ;)

dev_guide/builds.adoc
rhcarvalho: Is it really required to have some binary in the client system, like `zip` or `tar`? Why isn't the client side also handled by the platform, `oc` in this case?

dev_guide/builds.adoc
bparees: @rhcarvalho many of us older folks were taught to use two spaces after a period, not one.  Personally I still prefer it.  I haven't audited our docs to see if we're doing both or one exclusively (any docs i've written i expect have two)., but i'd leave it up to the docs folks to audit for stylistic decisions like that and so far they don't seem concerned.

dev_guide/builds.adoc
mfojtik: https://github.com/openshift/origin/blob/master/pkg/cmd/cli/cmd/startbuild.go#L495

dev_guide/builds.adoc
bparees: yeah it should have dots, i think that's my fault (I wrote the sample gabe used).

dev_guide/builds.adoc
bparees: no it is not possible to do that.

and it's not possible to control binary.asFile, binary.asFile gets set if you use --from-file, if you don't use --from-file, it is not set.

dev_guide/builds.adoc
bparees: not sure what happens if the path isn't valid, that's a good question, but really one that should be addressed at the point in the docs where we describe the contextDir field, not repeated here.

Same for whether or not it needs to be a relative path.

dev_guide/builds.adoc
rhcarvalho: That's interesting, first time I hear about it! Is that something coming from typing courses?

In HTML, the output will be rendered as a single space no matter how many spaces we put there :-)

dev_guide/builds.adoc
rhcarvalho: Hmm, what was that link supposed explain? It's just a function that checks if a file is a known archive format.

We're about to update the docs to say:

> The contents will be archived (either zip, tar, or gzipped tar based on what is available on the client system)

And I'm trying to understand why the format depends on the client.

dev_guide/builds.adoc
rhcarvalho: Since the `contextDir` have multiple semantics depending the other fields, I doubt the behavior in code is the same in all cases. Agree it would be nice to have it addressed in a single place, assuming the answer to the question is one and the same.

dev_guide/builds.adoc
gabemontero: I learned two spaced after a period ending a sentence in English/Grammar classes in elementary school back in the 70's, though you ended up exaggerating it with handwritten assignments ;-) (didn't get to take typing until high school).   

Sorry, but I will not conform to whoever wrote the HTML rendering engine ;-)

dev_guide/builds.adoc
gabemontero: yep .... part of next push

dev_guide/builds.adoc
gabemontero: yep I missed it too ... part of next push

dev_guide/builds.adoc
gabemontero: Yeah that *may* be a less stringent rule but you are correct, it is better with comma ... part of next push

dev_guide/builds.adoc
bparees: yeah apparently it started w/ typewriters (though i never used a typewriter).  I guess since it's being converted that might explain why the docs folks didn't complain.

rhcarvalho: Learning this grammar/historical/cultural thing pays off the trouble of asking. Thanks @gabemontero and @bparees!

dev_guide/builds.adoc
bparees: it really doesn't.  the flow is (or should be):

1) collect all the content from the inputs, dump it in the base directory
2) change to the contextDir from within the base directory, if one was specified
3) execute build logic.

dev_guide/builds.adoc
gabemontero: Yeah, the code does not leverage a local binary.  I had seen that during a code walkthrough after writing that and had meant to correct this statement, but it slipped through the cracks.

I think the more illustrative lines in startbuild.go are here:  https://github.com/openshift/origin/blob/master/pkg/cmd/cli/cmd/startbuild.go#L418-L486 ... it either tar, tar.gz, or just streams the file bytes based on exactly which option.

dev_guide/builds.adoc
gabemontero: Should be using backtick and asterik for dockerfile and Binary in that paragraph now .. part of next push

dev_guide/builds.adoc
gabemontero: I'm taking the net of leaving this sentence as is.

gabemontero:
Yeah the `oc new-build --binary` and "why would you create a build config with binary" is klunky, because binary build broke from the existing convention of the BuildConfig solely driving the determination of your source type.  Earlier iterations of the doc pull attempted to go into that, but through review we've backed off of that.

As you implied, the product will continue to evolve in this space.  Let's let that evolution play out a bit before diving back down this rabbit hole.

As to special s2i/docker/custom considerations, this doc originally had some of that, but upon (re)reviewing the code and trying out example, we came to the conclusion that there were in fact no special differences to note.

On the width, I'll compromise with you :-) ... 80 seems to narrow, but I did split some of the lines up
into the 100s range.

gabemontero:
updates from @rhcarvalho / @bparees review pushed ... thanks !

dev_guide/builds.adoc
rhcarvalho: So we need to update this statement to reflect (or omit?!) the truth.
A quick glance at that code and it seems that it does depend on having `git` to try a `git archive` first, or fallback to something else. No code for zipping the file (just gzip (the tar stream)).

Thanks for the link, indeed illustrates the point.

dev_guide/builds.adoc
```diff
@@ -846,43 +850,48 @@ source:
...
-  dockerfile: "FROM centos:7\nRUN yum install -y httpd" <4>
-----
-<1> The `*binary*` field specifies the details of the binary source.
-<2> The `*asFile*` field specifies the name of the file that will be created with
-the binary contents.
-<3> The `*contextDir*` field specifies a subdirectory within the extracted contents of a
-binary archive.
-<4> If the optional `*dockerfile*` field is provided, it should be a string
-containing an inline Dockerfile that potentially replaces one within the
-contents of the binary archive.
```
rhcarvalho: Doesn't it make sense to keep this YAML example to illustrate some of the discussion in prose?
We'd just need to point out that `<2>` is not part of a build config, but how `oc start-build` would setup a  build.

gabemontero: @bparees (I believe) called for this deletion, so I want to hear his take, but I interpretation of the motivation for removing it was that it implied it needed to be created in order to do binary builds.  It falls back to the uniqueness of binary build in comparison to the other source types.

I'm still inclined to delete it.

gabemontero:
I just realized I did not fully address the tar/zip/gzip/client system piece.... will push momentarily.

dev_guide/builds.adoc
bparees: I can see either argument but yeah i preferred removing it because really it's an implementation detail and i don't want to mislead a user down a path of actually writing a buildconfig in that way.

gabemontero:
Pushed the tar/zip/gzip/etc. changed ... decided to just say we archive it ... whether we use tar with one form or call git.Archive (which gzip tars) in another was an implementation detail we didn't need to dive into.

dev_guide/builds.adoc
gabemontero: :-)

dev_guide/builds.adoc
rhcarvalho: I agree with the fact users should not be creating BCs with this. Well, actually they can, right? The Binary type, as documented here, is a valid type and all of the other options apply. One can also mix and match Binary + Image Source or Dockerfile... The only invalid/misleading field is `asFile`, isn't it?

The problem is we reference `*binary.asFile*`, `*dockerfile*` and `*contextDir*` down below and these things relate to nothing concrete in the context; they're like null pointers, aren't they?

dev_guide/builds.adoc
```diff
@@ -846,43 +850,47 @@ source:
...
+
+In each of the above cases:
+
+1) If your `*BuildConfig*` already has a `*Binary*` source type defined, it will effectively be ignored and replaced by what the
+client sends.
+
+2) If your `*BuildConfig*` has a `*Git*` source type defined, it will be dynamically disabled, since `*Binary*` and `*Git*` are
+mutually exclusive, and the data in the binary stream provided to the builder will take precedence.
+
+When using `oc new-build --binary=true`, the command ensures that the restrictions associated with binary builds are enforced.
```
rhcarvalho: I normally write `oc new-build --binary` in the command line :-)

dev_guide/builds.adoc
```diff
@@ -846,43 +850,47 @@ source:
...
-contents of the binary archive.
+Since binary type builds require content to be streamed from the local file system, it is not possible
+to automatically trigger a binary type build (e.g. via an image change trigger) because the binary files could
+not be provided. Similarly, you cannot launch binary type builds from the web console.
 ====
 
+To utilize binary builds, invoke `oc start-build` with one of these options:
+
+1) `--from-file`:  The contents of the file you specify will be sent as a binary stream to the builder.  Then
+the builder will store the data in a file with the same name at the top of the build context (this internally maps
```
rhcarvalho: > at the top of the build context

Not something introduced here, but looking at the [build dev guide](https://docs.openshift.org/latest/dev_guide/builds.html) seems that we never defined what is the "build context". Is it something obvious to the intended public of the docs?

dev_guide/builds.adoc
gabemontero: Being more explicit/precise is never a bad thing though :-)

dev_guide/builds.adoc
gabemontero: Both "context" and "build context" are used somewhat generously in this file.  But you are correct that it is not explicitly defined.  If you have a proposed definition and location to include the definition, I can include it as part of this pull.  Otherwise, I'm going to defer.

dev_guide/builds.adoc
gabemontero: The null pointer analogy from @rhcarvalho is swaying me to put these back in, with some sort of preamble that you don't *have* to specify this, etc. ....

... but stop the pendulum swinging in my head @bparees if @rhcarvalho 's last point did not sway your preference.

dev_guide/builds.adoc
rhcarvalho: Not that I have a perfect definition, but I'd start with what is a context for Docker:

https://docs.docker.com/engine/reference/builder/
https://docs.docker.com/engine/reference/glossary/ (under build)

In our case, "build context" can either mean the same as the Docker build context, or in case of S2I it's normally conveying to me the idea that there's a directory in the container's file system (`/opt/app-root/src`) and some data (e.g., coming from a Git repo).

We can use `contextDir`, a relative path, to change the working directory / build context relative to the original context.

So, `--from-file` takes a file from my local system and transfers it to the image I'm building by placing it at the image's WORKDIR in case of S2I, which happens to be /opt/app-root/src for our builders. For a Docker strategy build, the file is sent as part of the Docker build context.

dev_guide/builds.adoc
bparees: > The problem is we reference *binary.asFile*, *dockerfile* and *contextDir* down below and these things relate to nothing concrete in the context; they're like null pointers, aren't they?

the binary.asFile reference could/should be removed.  the dockerfile and contextDir references are referencing to concepts/fields described elsewhere in the doc, they are not specific not binary build, so no, they aren't null pointers.

The only argument in my head for showing how to define a BuildConfig with a Binary build source is because it's possible a user might want to intentionally define a buildconfig that they only intend to build via the binary process (therefore they do not have a valid Dockerfile, Git, or Image source they can define as part of the BuildConfig).

However even if we instruct them on how to do that, we should not be explaining asFile because that is still a field they have zero control over (it doesn't matter if they provide it or not, or what value they provide, it always gets overwritten).

so in sum i'm ok w/ putting the example back, but with modifications.

dev_guide/builds.adoc
gabemontero: OK, I'm going to let you or someone else handle that with a different doc PR :-)

dev_guide/builds.adoc
gabemontero: OK, I'm going to remove the asFile ref in the doc, even if I add the example back in (where I would remove it from there as well).

dockerfile and contextDir are illustrated with examples in other parts of this page ...

... I just got my afternoon coffee and will ruminate while I drink, seeing how many more times I flip-flop.  Where ever I land when I finish my coffee is what I'll go with :-)

gabemontero: OK, I landed with this:
1) removed all ref to asFile
2) did *NOT* add the sample back in
3) put a link back to the Git source type section, in case a user wanted to revisit the yaml snippet introduced there

gabemontero:
Latest responses pushed.

dev_guide/builds.adoc
```diff
@@ -846,43 +850,48 @@ source:
...
+client sends.
+
+2) If your `*BuildConfig*` has a `*Git*` source type defined, it will be dynamically disabled, since `*Binary*` and `*Git*` are
+mutually exclusive, and the data in the binary stream provided to the builder will take precedence.
+
+When using `oc new-build --binary=true`, the command ensures that the restrictions associated with binary builds are enforced.
+The resulting `*BuildConfig*` will have a source type of `*Binary*`, meaning that the only valid way to run a build for this `*BuildConfig*`
+will be to use `oc start-build` with one of the `--from` options to provide the requisite binary data.
+
+The `*dockerfile*` and `*contextDir*` link:#source-code[Git source options] have special meaning with binary builds.
```
bparees: they are just source options, not git source options (ie they are applicable even when you don't have a git source).  the section you linked to needs some reworking/retitling to clarify that, but we can do that in another PR.

bparees:
106 comments later, @gabemontero wishes he'd never signed up to clean up this doc...

gabemontero:
link changed to "source options", and, well, LOL to that last comment :-)

bparees:

adellape:
> 106 comments later

Holy moly.

Went straight to the diff and it LGTM for merge, with some @openshift/team-documentation edits to follow-up after.

dev_guide/builds.adoc
rhcarvalho: So that at least we have a reminder: https://github.com/openshift/openshift-docs/issues/1786

Issue #1645, opened 2016-02-23
oc log works w/ the dc directly


tnguyen-rh:
In https://github.com/openshift/openshift-docs/pull/1638 @kargakis notes:

> We now support getting application logs just by running oc logs against the dc

and furthermore

> So once the docker registry is deployed, `oc logs dc/docker-registry` should be the same as `oc logs docker-registry-1-da73t`.

PR #1749, opened 2016-03-16
Use oc logs with the registry deployment config

0xmichalis:
Fixes https://github.com/openshift/openshift-docs/issues/1645

@tnguyen-rh PTAL

### DIFF
--- a/install_config/install/docker_registry.adoc
+++ b/install_config/install/docker_registry.adoc
@@ -273,11 +273,11 @@ unnecessary downtime for pods and does not actually clear the cache.
 [[viewing-logs]]
 == Viewing Logs
 
-To view the logs for the Docker registry, run the `oc logs` indicating the desired pod:
+To view the logs for the Docker registry, use the `oc logs` command with the deployment config:
 
 ====
 ----
-$ oc logs docker-registry-1-da73t
+$ oc logs dc/docker-registry
 2015-05-01T19:48:36.300593110Z time="2015-05-01T19:48:36Z" level=info msg="version=v2.0.0+unknown"
 2015-05-01T19:48:36.303294724Z time="2015-05-01T19:48:36Z" level=info msg="redis not configured" instance.id=9ed6c43d-23ee-453f-9a4b-031fea646002
 2015-05-01T19:48:36.303422845Z time="2015-05-01T19:48:36Z" level=info msg="using inmemory layerinfo cache" instance.id=9ed6c43d-23ee-453f-9a4b-031fea646002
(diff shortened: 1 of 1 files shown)


tnguyen-rh:
@kargakis LGTM.  This is for 3.2, right?

kargakis:
Not necessarily but I would guess every fix we can squeeze in is ok

tnguyen-rh:
@kargakis 
I've added the label for 3.1.  How about 3.0?

kargakis:
This was added in 3.1.1.2. So I guess it's 3.2?

tnguyen-rh:
[rev_history]
|link:../install_config/install/deploy_registry.html[Installing -> Deploying a Docker Registry]
|Used `oc logs` in the link:../install_config/install/deploy_registry.html#access-pushing-and-pulling-images[deployment configuration example].
%

vikram-redhat:
@tnguyen-rh @adellape - there is a reference to OSE 3.2 in this. Should this not be tagged with enterprise-3.2?

adellape:
@vikram-redhat @tnguyen-rh According to @kargakis above it was added in 3.1.1.2, so that would warrant branch/enterprise-3.1 + Next Release.

I tried this on a 3.2 machine and it seemed to work as described (pod logs looked like they matched up with latest dc logs). On a 3.1.1.6 machine, it looks like it's showing the latest deployment logs, but since it was successful there's nothing to show (as mentioned in https://docs.openshift.com/enterprise/3.1/dev_guide/deployments.html#viewing-deployment-logs):

http://pastebin.test.redhat.com/358308

When the deployment was running (or when there's a failed deployment), the `oc logs dc/<name>` showed deployment logs. But from what this PR is suggesting, I would have expected to see the same pod logs from the `oc logs dc/docker-registry` for a successful deployment (like I'm seeing in 3.2).

If I'm misunderstanding the described new feature here, please correct me.

vikram-redhat:
@adellape, @tnguyen-rh , @kargakis 

I was looking at this particular line:

. If your registry was initially deployed on {product-title}
ifdef::openshift-enterprise[]
3.2
endif::[]
ifdef::openshift-origin[]
1.1.2
endif::[]
or later, update the scheme used for the registry's readiness probe from HTTP to HTTPS:

If we merge it in for the next release it will be incorrect? Does this need to be fixed to say 3.1.1.6 instead?

adellape:
@vikram-redhat That is from https://github.com/openshift/openshift-docs/pull/1728 (marked for enterprise-3.2/Future Release) and needs to stay out of the 3.1 docs.

kargakis:
> On a 3.1.1.6 machine, it looks like it's showing the latest deployment logs, but since it was successful there's nothing to show (as mentioned in https://docs.openshift.com/enterprise/3.1/dev_guide/deployments.html#viewing-deployment-logs):

Great for digging this out, we need to update it too.:) Complete deployments should return pod logs.

tnguyen-rh:
@adellape 
Thanks for the detective work.  To summarize, the actual funtionality will be user-visible in 3.2.  I've adjusted te labels/milestone accordingly.  Updates/correction welcome.

PR #1763, opened 2016-03-18
Clarify license of OpenShift docs.

richardfontana:
The OpenShift docs repository needs an explicit license. I would say there has been an implicit assumption that it inherits the general Apache License 2.0 policy of OpenShift Origin; at the same time downstream OpenShift product documentation is using CC BY-SA 3.0. Therefore I suggest using an Apache License 2.0|CC BY-SA 3.0 dual license.


pmorie:
LGTM

PR #1764, opened 2016-03-18
Introduce node allocatable resources

ingvagabund:
Kubelet has introduced concept of allocatable resources so scheduler is more aware of node's capabilities and does not overcommit node's resources.

### DIFF
--- a/admin_guide/node-allocatable-resources.adoc
+++ b/admin_guide/node-allocatable-resources.adoc
@@ -1,0 +1,103 @@
+= Node Allocatable resources
+{product-author}
+{product-version}
+:data-uri:
+:icons:
+:experimental:
+:toc: macro
+:toc-title:
+
+toc::[]
+
+== Overview
+To provide more reliable scheduling and minimize node resource overcommitment,
+each node has a chance to reserve a portion of its resources for use by all
+openshift-node underlying components (e.g. kubelet, kube-proxy, docker)
+and remaining system components (e.g. sshd, NetworkManager) on a node.
+Once specified, scheduler has more information about how much resources
+(e.g. memory, cpu) a node has allocated for pods.
+
+== Settings
+
+Resources reserved for components are based on two node settings:
+
+[options="header",cols="1,2"]
+|===
+
+|Setting |Description
+
+|`*kube-reserved*`
+| A set of `ResourceName=ResourceQuantity` (e.g. cpu=200m,memory=30G) pairs that describe resources reserved for openshift-node components. Default is none.
+
+|`*system-reserved*`
+|A set of `ResourceName=ResourceQuantity` (e.g. cpu=200m,memory=30G) pairs that describe resources reserved for remaining system components. Default is none.
+|===
+
+You can specify values for these settings in the `*kubeletArguments*` section of
+the
+link:../install_config/master_node_configuration.html#node-configuration-files[node
+configuration file] (the *_/etc/origin/node/node-config.yaml_* file by default). Add the section if it does not already exist:
+
+====
+.Node Allocatable Resources Settings
+[source,yaml]
+----
+kubeletArguments:
+  kube-reserved:
+    - "cpu=200m,memory=30G"
+  system-reserved:
+    - "cpu=200m,memory=30G"
+----
+====
+
+Currently, `cpu` and `memory` resource types are supported.
+
+Each resource types is specified in `ResourceQuantity` units. For `cpu` it is
+in units of cores (e.g. 200m, 100Ki, 50M),
+for ``memory`` in units of bytes (e.g. 200Ki, 100M, 50Gi).
+
+See
+link:../dev_guide/compute_resources.html[Compure Resources]
+for more details.
+
+If a flag is not set, it defaults to `0`.
+If none of the flags is set, allocated resource is set to node's capacity as it was
+before introduction of allocatable resources.
+
+== Computation of allocated resource
+
+An allocated amount of a resource is computed based on the following formula:
+
+----
+[Allocatable] = [Node Capacity] - [Kube-Reserved] - [System-Reserved]
+----
+
+If `[Allocatable]` is negative, it is set to `0`.
+
+== Node status
+
+To see node's current capacity and allocatable resources, you can run:
+
+----
+$ oc get node/NODENAME -o json
+...
+    "status": {
+        "capacity": {
+            "cpu": "2",
+            "memory": "3882324Ki",
+            "pods": "110"
+        },
+        "allocatable": {
+            "cpu": "2",
+            "memory": "3882324Ki",
+            "pods": "110"
+        },
+...
+----
+
+== Scheduler
+
+Scheduler now uses value of `node.Status.Allocatable` instead of `node.Status.Capacity`
+to decide if a node will become a candidate for pod scheduling.
+
+By default, the node will report its machine capacity as fully schedulable by the cluster.
--- a/admin_guide/overcommit.adoc
+++ b/admin_guide/overcommit.adoc
@@ -175,9 +175,19 @@ for the system daemons that are required to run on your node for your cluster to
 function (*sshd*, *docker*, etc.). In particular, it is recommended that you
 reserve resources for incompressible resources such as memory.
 
-If you want to explicitly reserve resources for non-pod processes, you can
-create a *resource-reserver* pod that does nothing but reserve capacity from
-being scheduled to on the node by the cluster. For example:
+If you want to explicitly reserve resources for non-pod processes, there are two
+ways to do it:
+
+* Node allocatable resources by specifying resources available for scheduling.
+See
+link:../admin_guide/node-allocatable-resources.html[node allocatable resources]
+for more details.
+* Create a *resource-reserver* pod that does nothing but reserve capacity from
+being scheduled to on the node by the cluster.
+
+Node allocatable resources is preferred to resource-reserver.
+
+*Resource-reserver* example:
 
 .Definition for a *resource-reserver* Pod
 ====


jeremyeder:
LGTM

ncdc:
@derekwaynecarr

ncdc:
@ingvagabund do you need to update/remove https://github.com/openshift/openshift-docs/blob/master/admin_guide/overcommit.adoc#reserving-resources-for-system-processes?

ingvagabund:
@ncdc, yeah, it should be prunned. All paragraphs starting from the third. Then put a link to this document for more information about setting --kube-reserved and --system-reserved flags that does basically the same. I will make another commit.

ingvagabund:
@ncdc Having second thoughts, the resource-reserver pod allows dynamical change of the reserved resources. It makes sense to provide both solutions. WDYT?

ncdc:
I'm pretty sure we want to avoid dynamic changes to reservations, right @derekwaynecarr ?

derekwaynecarr:
I am not sure if I want to remove the resource-reserver pod section completely since some users will have been utilizing that feature and perform an upgrade.

Can we present it as two options, and give a strong preference for users to move to the new option?

ncdc:
WFM

admin_guide/node-allocatable-resources.adoc
```diff
@@ -0,0 +1,58 @@
...
+:experimental:
+:toc: macro
+:toc-title:
+
+toc::[]
+
+== Overview
+To provide more reliable scheduling and minimize node resource overcommitment,
+each node has a chance to specify how much resources get consumed
+by kubernetes related components (e.g. kubelet, kube-proxy, docker) running on a node
```
derekwaynecarr: I think we should avoid use of 'kubernetes' and 'non-kubernetes' and just use 'openshift'

derekwaynecarr: i think this needs to be documented in terms of units we ship.

'openshift-node' is the openshift node-related component.

admin_guide/node-allocatable-resources.adoc
```diff
@@ -0,0 +1,58 @@
...
+and by remaining non-kubernetes components (e.g. system daemons, databases).
```
derekwaynecarr: probably just give examples like sshd, docker.service, NetworkManager.service are all non openshift related services.

derekwaynecarr: I think we need something that says something like:  by default, the node will report its machine capacity as fully schedulable by the cluster.

admin_guide/node-allocatable-resources.adoc
```diff
@@ -0,0 +1,58 @@
...
+Once specified, scheduler has more information about how much resources
+(e.g. memory, cpu) a node has allocated for pods.
+
+== Setting reserved resources
+
+Both kubernetes and non-kubernetes components reserved resources are specified through kubelet command-line flags:
```
derekwaynecarr: we should show these flags as they would be provided in the `node-config.yaml` file for openshift-node.

admin_guide/node-allocatable-resources.adoc
detiber: Should we use the naming from the systemd units? If so, it's atomic-openshift-node for OSE/AP and origin-node for Origin

ingvagabund: @derekwaynecarr openshift-node is one binary that starts all underlying components (kube-proxy, kubelet, docker, ...). So what about:
```
by all underlying components (e.g. kubelet, kube-proxy, docker) that openshift-node runs on a node
```

@detiber IIUC, openshift-node is a generic term for a set of components/daemons that are run/spawn on each node openshift-node component occupies. It should cover both OSE/AP and Origin. But I may be mistaken.

admin_guide/node-allocatable-resources.adoc
ingvagabund:
```
and by remaining components (e.g. sshd, NetworkManager).
```
That is a question. My interpretation of ``--kube-reserved`` was covering docker as well [1]. I interpret ``--system-reserved`` as resource reserved for everything that is not kubernetes/openshift related. NetworkManager could be misinterpreted incorrectly this way but as it does not have to be used with k8s/openshift, it falls out of ``--kube-reserved``.

[1] https://github.com/timstclair/kubernetes/blob/54a6da979c6d85dc81fae364124d5c7dd465bc6c/docs/proposals/node-allocatable.png

ingvagabund: > I think we need something that says something like: by default, the node will report its machine capacity as fully
> schedulable by the cluster.

Suitable for **scheduler** part?

admin_guide/node-allocatable-resources.adoc
ingvagabund: Agree. ``node-config`` is preferable or both (with flags as well) are suitable? I was wondering about it. As far as I saw all docs use ``node-config`` to describe flags/options.

admin_guide/node-allocatable-resources.adoc
detiber: @ingvagabund as long as we aren't worried about user confusion between the generic openshift-node and the packages they install and services they run, then I have no objections with using openshift-node here.

ingvagabund:
@ncdc, @derekwaynecarr update overcommit.adoc to mention both ways but prefer node allocatable resources.

admin_guide/node-allocatable-resources.adoc
```diff
@@ -0,0 +1,58 @@
...
+:icons:
+:experimental:
+:toc: macro
+:toc-title:
+
+toc::[]
+
+== Overview
+To provide more reliable scheduling and minimize node resource overcommitment,
+each node has a chance to specify how much resources get consumed
```
ncdc: each node has a change to reserve a portion of its resources for use by kubernetes related components

admin_guide/node-allocatable-resources.adoc
```diff
@@ -0,0 +1,58 @@
...
+Once specified, scheduler has more information about how much resources
+(e.g. memory, cpu) a node has allocated for pods.
+
+== Setting reserved resources
+
+Both kubernetes and non-kubernetes components reserved resources are specified through kubelet command-line flags:
+
+====
+----
+--kube-reserved=: A set of ResourceName=ResourceQuantity (e.g. cpu=200m,memory=150G) pairs that describe resources reserved for kubernetes system components. [default=none]
```
mdshuai: For user in openshift they more care about how to setup this in "kubeletArguments" part in node-config.yaml.
"--kube-reserved=" is not a good choice in openshift-doc.

ingvagabund:
Updated. PTAL

admin_guide/node-allocatable-resources.adoc
```diff
@@ -0,0 +1,82 @@
...
+
+Resources reserved for components are based on two node settings:
+
+[options="header",cols="1,2"]
+|===
+
+|Setting |Description
+
+|`*kube-reserved*`
+| A set of `ResourceName=ResourceQuantity` (e.g. cpu=200m,memory=150G) pairs that describe resources reserved for openshift-node components. Default is none.
```
derekwaynecarr: reserving 200m of cpu is sensible, 150G of memory is not.  maybe reduce to a smaller number to avoid making us look like a memory hog or having confusion.

admin_guide/node-allocatable-resources.adoc
```diff
@@ -0,0 +1,82 @@
...
+
+|`*system-reserved*`
+|A set of `ResourceName=ResourceQuantity` (e.g. cpu=200m,memory=150G) pairs that describe resources reserved for remaining system components. Default is none.
```
derekwaynecarr: same as above

admin_guide/node-allocatable-resources.adoc
```diff
@@ -0,0 +1,82 @@
...
+link:../install_config/master_node_configuration.html#node-configuration-files[node
+configuration file] (the *_/etc/origin/node/node-config.yaml_* file by default). Add the section if it does not already exist:
+
+====
+.Node Allocatable Resources Settings
+[source,yaml]
+----
+kubeletArguments:
+  kube-reserved:
+    - "cpu=200m,memory=150G"
```
derekwaynecarr: same

admin_guide/node-allocatable-resources.adoc
```diff
@@ -0,0 +1,82 @@
...
+  system-reserved:
+    - "cpu=200m,memory=150G"
```
derekwaynecarr: same

derekwaynecarr:
Add a section on how they can introspect a node to see its capacity versus its allocatable value.

Show a YAML snippet like the following:

```
apiVersion: v1
items:
- apiVersion: v1
  kind: Node
  status:
    allocatable:
      cpu: "2"
      memory: 1017192Ki
      pods: "110"
    capacity:
      cpu: "2"
      memory: 1017192Ki
      pods: "110"
...
```

I suspect operators will want to visually introspect the node is set as expected so this will help.

ingvagabund:
Updated. Showing json format instead of yaml. PTAL

derekwaynecarr:
LGTM

ingvagabund:
@csrwng believe it is ready for merging

csrwng:
@adellape

ingvagabund:
@adellape can we have this merged? I will have another PR updating admin_guide/overcommit.adoc.

adellape:
@ingvagabund I've pulled your commits into https://github.com/openshift/openshift-docs/pull/1891 and continued with more edits. Closing this PR in favor of that one, PTAL there.

PR #1780, opened 2016-03-21
fix registry sa scc call

pweil-:
Fix call when adding the service account

@aweiteka

### DIFF
--- a/install_config/install/docker_registry.adoc
+++ b/install_config/install/docker_registry.adoc
@@ -28,8 +28,7 @@ the *_admin.kubeconfig_* file's location, as a user with cluster administrator
 privileges:
 
 ----
-$ oadm registry --config=admin.kubeconfig \//<1>
-    --credentials=openshift-registry.kubeconfig <2>
+$ oadm registry --credentials=openshift-registry.kubeconfig <1>
 ----
 endif::[]
 ifdef::openshift-enterprise[]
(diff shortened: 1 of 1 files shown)


aweiteka:
Tested. Thanks for the update.

aweiteka:
Related to this, should we also remove the credentials arg from the following step? Currently it fails but it may be due to a bug. It would seem credentials should not be required when using with serviceaccount. The command that works:

```
$ oadm registry --service-account=registry \
    --config=admin.kubeconfig \
    --mount-host=<path>
```

pweil-:
Yes, I'll remove it.  Looking at the code it does appear to be correct to use either the service account or the credentials.  If credentials are specified the service account will not be set.

pweil-:
@aweiteka - removed `credentials` and `config`.  Config is no longer a flag and not mentioned in help anyway.

adellape:
@pweil- Sorry, this needs a rebase.

aweiteka:
This may need more work since this https://github.com/openshift/origin/pull/8296 was merged.

@liggitt was the a docs PR to accompany your work in https://github.com/openshift/origin/pull/8296?

liggitt:
not yet

pweil-:
@adellape rebased

pweil-:
> This may need more work since this openshift/origin#8296 was merged.

I removed the need to set credentials in this PR already so that it shows the example that will use a service account by default.  Let me know you're seeing something I missed.

adellape:
From @liggitt in https://github.com/openshift/origin/issues/8285#issuecomment-203442167:

> You don't actually need to pass --credentials any more, the service account can be used for everything.

Currently we only show use of `--service-account=registry` in these `--mount-host` examples. So it sounds like for Origin and OSE 3.2+ we can update all of the `oadm registry` examples to use `--service-account` instead of `--credentials`?

Also, even back in 3.1, it doesn't look like we need `--images` anymore in any of these examples. 

@pweil- And speaking of 3.1, I take it this update to drop `--credentials` only applies to 3.2+. At least, I tried it out on 3.1.1.6 vs 3.2 and those were my findings here: http://pastebin.test.redhat.com/364004. I think the Origin version of this command can also be updated, yeah? At https://github.com/openshift/openshift-docs/pull/1780/files#diff-d7dbe137f42b2f8f406596161a89db90R151.

liggitt:
yes, 3.2+ only for dropping --credentials

pweil-:
> only applies to 3.2+.

correct

> So it sounds like for Origin and OSE 3.2+ we can update all of the oadm registry examples to use --service-account instead of --credentials

yep

> even back in 3.1, it doesn't look like we need --images anymore in any of these examples

images was always defaulted, correct. I left the existing command just because it looked like it was directed specifically at using the rhel images.  I can remove it if you want.

adellape:
> images was always defaulted, correct. I left the existing command just because it looked like it was directed specifically at using the rhel images. I can remove it if you want.

@pweil- I think it's safe to remove. In OSE docs they're instructed to install docker from RHEL7, which has `--add-registry registry.access.redhat.com` set by default:

```
# egrep registry.access /etc/sysconfig/docker
ADD_REGISTRY='--add-registry registry.access.redhat.com'
```

So AFAICT, everything works out:

```
# oc describe pod docker-registry-1-b5h0i | grep Image
Image(s):			openshift3/ose-docker-registry:v3.1.1.6

# docker images | grep ose-docker
registry.access.redhat.com/openshift3/ose-docker-registry   v3.1.1.6            3c272743b20a        2 weeks ago         478.5 MB
```

pweil-:
@adellape that is adding the redhat registry to the docker daemon.  The images flag still tells which images to pull so I think it's still relevant.  I removed some extra references to config that I found.  PTAL

adellape:
> The images flag still tells which images to pull so I think it's still relevant.

OK. I had thought it was built in to `oadm registry` now to know which image to pull (distro/component/version). Either way can't hurt. Thanks.

adellape:
Closed in favor of https://github.com/openshift/openshift-docs/pull/2240, continuing there.

PR #1782, opened 2016-03-22
bz-1297659: updated line to specifically mention which command(s) nee…

mishaone:
…d to be run with the stated privileges

Looked into the rest of the "section" and the only command being used is "oc create" so I've changed the line to be more specific about this as requested in the BZ. @tpoitras could you review this when you have time?

### DIFF
--- a/install_config/imagestreams_templates.adoc
+++ b/install_config/imagestreams_templates.adoc
@@ -14,34 +14,34 @@ toc::[]
 
 ifdef::openshift-enterprise[]
 Your OpenShift installation includes useful sets of Red Hat-provided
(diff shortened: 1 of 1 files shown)


bfallonf:
@mishaone Looks fine to me. I'll let @vikram-redhat merge it though cos I'm not sure which labels and milestones it needs and it might get in the way of this week's release.

@vikram-redhat All yours :pineapple:

adellape:
No revhistory needed for this, I guess.

tnguyen-rh:
/me musing ...oO (Probably we'll need to (at some point) add something to the style guide about eol whitespace...)

PR #1789, opened 2016-03-22
fixed issues with xpaas docs markup

vikram-redhat:


vikram-redhat:
No rev history or labels applied as it is quick fix to update the markup issues with xPaaS docs.

PR #1797, opened 2016-03-22
Bug 1286858, Documented the infra region

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1286858

@adellape PTAL. Thanks!

### DIFF
--- a/install_config/install/advanced_install.adoc
+++ b/install_config/install/advanced_install.adoc
@@ -146,6 +146,14 @@ invalid if used. These values override other settings in node configuration
 which may cause invalid configurations. Example usage:
 *{'image-gc-high-threshold': ['90'],'image-gc-low-threshold': ['80']}*.
 
+|`*openshift_router_selector*` 
+|Default node selector for automatically deploying router pods. See
+link:#configuring-node-host-labels[Configuring Node Host Labels] for details.
+
+|`*openshift_registry_selector*`
+|Default node selector for automatically deploying registry pods. See
+link:#configuring-node-host-labels[Configuring Node Host Labels] for details.
+
 |===
 
 [[configuring-cluster-variables]]
@@ -303,15 +311,40 @@ link:../../architecture/core_concepts/pods_and_services.html#labels[labels] to
 node hosts during the Ansible install by configuring the *_/etc/ansible/hosts_*
 file. Labels are useful for determining the placement of pods onto nodes using
 the link:../../admin_guide/scheduler.html#configurable-predicates[scheduler].
+Other than *region=infra* (discussed below), the actual label names and values
+are arbitrary and can be assigned however you see fit per your cluster's
+requirements.
 
 To assign labels to a node host during an Ansible install, use the
 `*openshift_node_labels*` variable with the desired labels added to the desired
-node host entry in the *[nodes]* section. For example:
+node host entry in the *[nodes]* section. In the following example, labels are
+set for a region called *primary* and a zone called *east*:
 
 ====
 ----
 [nodes]
 node1.example.com openshift_node_labels="{'region': 'primary', 'zone': 'east'}"
+----
+====
+
+The `*openshift_router_selector*` and `*openshift_registry_selector*` Ansible
+settings are set to *region=infra* by default:
+
+====
+----
+# default selectors for router and registry services
+# openshift_router_selector='region=infra'
+# openshift_registry_selector='region=infra'
+----
+====
+
+The default router and registry will be automatically deployed if nodes exist
+that match the selector settings above. For example:
+
+====
+----
+[nodes]
+node1.example.com openshift_node_labels="{'region':'infra','zone':'default'}"
 ----
 ====
 
@@ -337,6 +370,7 @@ master. For example:
 master.example.com openshift_node_labels="{'region':'infra','zone':'default'}" openshift_schedulable=false
 ----
 ====
+
 
 [[advanced-install-session-options]]
 


install_config/install/advanced_install.adoc
```diff
@@ -315,6 +315,23 @@ node1.example.com openshift_node_labels="{'region': 'primary', 'zone': 'east'}"
...
+====
+----
+# default selectors for router and registry services
+# openshift_router_selector='region=infra'
+# openshift_registry_selector='region=infra'
+----
+====
+
+The default router and registry are automatically deployed. The default label is
+`*region=infra*`. In all other aspects, this is a normal node. Not all nodes are
```
brenton: The router / registry will only be automatically deploy if Nodes exist with labels that match the selector settings above.  Here's an example of how ansible labels can apply labels to Nodes: https://github.com/openshift/openshift-ansible/blob/master/inventory/byo/hosts.ose.example#L289

ahardin-rh: @brenton Thanks! Made some edits. PTAL and let me know if it's clearer.

install_config/install/advanced_install.adoc
```diff
@@ -315,6 +315,34 @@ node1.example.com openshift_node_labels="{'region': 'primary', 'zone': 'east'}"
 ----
 ====
 
+Nodes should have a label signifying the router. The registry should be
+installed on them. For example:
```
brenton: This part sounds confusing to me.  Nodes have labels that often indicate their region.  A convention is to set the 'region=infra' label for the router and registry selector.

ahardin-rh:
@bfallonf @tpoitras @tnguyen-rh PTAL as well. Thanks!

install_config/install/advanced_install.adoc
```diff
@@ -315,6 +315,34 @@ node1.example.com openshift_node_labels="{'region': 'primary', 'zone': 'east'}"
...
+----
+# default selectors for router and registry services
+# openshift_router_selector='region=infra'
+# openshift_registry_selector='region=infra'
+----
+====
+
+The default label is `*region=infra*`. In all other aspects, this is a normal
+node. Not all nodes are running mission-critical containers. While a node
+labeled `*dev*` might be rebooted or deleted at any time, the `*infra*` nodes
```
tnguyen-rh: A label comprises key and value, so "labelled *dev*" is incomplete (likewise, naked *infra*).  For the latter, i suppose the key would be *region* (thus, "the `*region=infra*` nodes").  Maybe the same key is used for *dev*, but i'm guessing not.

tnguyen-rh:
@ahardin-rh One small clarification required, otherwise LGTM.

ahardin-rh:
@tnguyen-rh Fixed. Thanks!

install_config/install/advanced_install.adoc
```diff
@@ -315,6 +315,34 @@ node1.example.com openshift_node_labels="{'region': 'primary', 'zone': 'east'}"
 ----
 ====
 
+Nodes have labels that often indicate their region. A convention is to set the
+`*region=infra*` label for the router and registry selector. For example:
```
bfallonf: @ahardin-rh I think the default setting is a little confused here. Above, you're telling the reader to set the region=infra label for the router and registry, but below, you're telling them that nodes have the label by default. Should the below say "All new nodes will have the *region=infra* label by default"?

bfallonf:
@ahardin-rh One question, but otherwise looks good!

install_config/install/advanced_install.adoc
brenton: Nodes only receive labels today if you explicitly add them in the inventory using `openshift_node_labels` as described below.  `region=infra` is the default selector and if you create Nodes with that selector then the router and registry will be installed.

install_config/install/advanced_install.adoc
```diff
@@ -315,6 +315,34 @@ node1.example.com openshift_node_labels="{'region': 'primary', 'zone': 'east'}"
...
+====
+----
+# default selectors for router and registry services
+# openshift_router_selector='region=infra'
+# openshift_registry_selector='region=infra'
+----
+====
+
+The default label is `*region=infra*`. In all other aspects, this is a normal
+node. Not all nodes are running mission-critical containers. While a node
```
brenton: This is probably the section that @bfallonf is getting confused by.  We're talking about label selectors and then the phrase 'this is a normal node' is used.  That seems out of place.  I would simply say something like, "If `openshift_node_labels` on a Node includes a label that matches the router or registry selector those services will be deployed by ansible."

brenton: Looking again I would say this paragraph could simply be removed.  The following one covers everything I just mentioned.

ahardin-rh:
@bfallonf @brenton Thanks! Made some updates per your comments.

install_config/install/advanced_install.adoc
```diff
@@ -315,6 +315,30 @@ node1.example.com openshift_node_labels="{'region': 'primary', 'zone': 'east'}"
...
+selector. If you create nodes with that selector, then the router and registry
+will be installed.
+
+====
+----
+# default selectors for router and registry services
+# openshift_router_selector='region=infra'
+# openshift_registry_selector='region=infra'
+----
+====
```
bfallonf: @ahardin-rh Sorry, i should have picked up on this yesterday, but is the "default selectors for router and registry services" statement supposed to be a uncommented meant for context? The other two lines are commands for the reader to run, right? If this is all true, the I think that line can probably be removed. The para above is stating that anyway.

bfallonf:
@ahardin-rh Sorry, one more comment :pineapple:

install_config/install/advanced_install.adoc
brenton: Those lines are likely taken directly from the configuration file.  It's common for configuration files to show the defaults by comments.  That way should an admin wish to change them they simply uncomment the line and then change the value.  If they don't change the value then there really isn't any point to uncomment them.  Personally, I would leave these lines as is.

ahardin-rh: Thanks, @bfallonf and @brenton  I will leave the lines as is per Brenton's comment.

ahardin-rh:
@brenton @bfallonf made a few edits per conversation with @adellape. This should be clearer/less redundant now. PTAL. Thanks!

ahardin-rh:
@brenton Did openshift_router_selector and openshift_registry_selector start working with 3.1? Thanks!

brenton:
@ahardin-rh, I know we didn't have them in 3.0.  How exact do you need the answer to be?  I'm not sure if it was 3.1 or 3.1.1.

ahardin-rh:
@brenton perfect. Thanks for clarifying!

ahardin-rh:
[rev_history]
|link:../install_config/install/advanced_install.html[Advanced Installation]
|Added information about *region=infra* to the link:../install_config/install/advanced_install.html#configuring-node-host-labels[Configuring Node Host Labels] section and added `*openshift_router_selector*` and `*openshift_registry_selector*` to the link:../install_config/install/advanced_install.html#configuring-host-variables[Host Variables] table.
%

anpingli:
The installing Ansible version  should be 1.8.4 to 1.9.x.  I don't think we support Ansible 2.0 and later now.

PR #1800, opened 2016-03-23
Unit for ES_RAM is not specified and default is incorrect

akram:
Unit for ES_RAM is not specified and default is incorrect


ahardin-rh:
@akram sorry for the delay on this and thank you! I am closing this in favor of https://github.com/openshift/openshift-docs/pull/2243.

PR #1808, opened 2016-03-24
Document Puma's performance tuning

mnagy:
Original sti-ruby pull: https://github.com/openshift/sti-ruby/pull/96
Trello card: https://trello.com/c/kvzlecgq/853-5-tune-ruby-based-on-memory-limits-evg

### DIFF
--- a/using_images/s2i_images/ruby.adoc
+++ b/using_images/s2i_images/ruby.adoc
@@ -96,6 +96,12 @@ and ruby gem installation. The application assets are only compiled if
 Asset compilation only happens when the application runs in a production
 environment. Therefore, you can use this variable when assets have already been
 compiled.
+
+|`*PUMA_MIN_THREADS*`, `*PUMA_MAX_THREADS*`
+|Indicates the minimum and maximum threads that will be available in https://github.com/puma/puma[Puma]'s thread pool.
+
+|`*PUMA_WORKERS*`
+|Indicates the number of worker processes that will be launched. See documentation on Puma's https://github.com/puma/puma#clustered-mode[clustered mode]. If not explicitly set, the default behavior sets `*PUMA_WORKERS*` to an appropriate value for the memory available to the container and the number of cores on the host.
 |===
 
 [[ruby-hot-deploy]]


mfojtik:
@mnagy @bparees since this is openshift documentation, I would like to see an example YAML of how I can tweak these settings... does it make sense?

bparees:
@mfojtik i dunno, i think i'd rather see us reference existing docs on how to set env variables in a deploymentconfig, than duplicating yaml in every image doc.

using_images/s2i_images/ruby.adoc
```diff
@@ -96,6 +96,13 @@ and ruby gem installation. The application assets are only compiled if
 Asset compilation only happens when the application runs in a production
 environment. Therefore, you can use this variable when assets have already been
 compiled.
+
+|`*PUMA_MIN_THREADS*`, `*PUMA_MAX_THREADS*`
+|These variables indicate the minimum and maximum threads that will be available
```
bparees: duplicated line?

using_images/s2i_images/ruby.adoc
```diff
@@ -96,6 +96,13 @@ and ruby gem installation. The application assets are only compiled if
...
+|These variables indicate the minimum and maximum threads that will be available in link:https://github.com/puma/puma[Puma]'s thread pool.
+
+|`*PUMA_WORKERS*`
+|This variable indicate the number of worker processes that will be launched. See documentation on Puma's link:https://github.com/puma/puma#clustered-mode[clustered mode].
```
bparees: s/indicate/indicates/

Also you can remove "This variable/These variables" from all these descriptions, so it would just read "Indicates the number of....."

using_images/s2i_images/ruby.adoc
```diff
@@ -117,6 +124,40 @@ link:../../dev_guide/environment_variables.html#set-environment-variables[the
...
 
+== Performance tuning
+You can tune the number of threads per worker using the
+`*PUMA_MIN_THREADS*` and `*PUMA_MAX_THREADS*` environment variables.
+Additionally, the number of worker processes is determined by the number of CPU
+cores that the container has available, as recommended by
+link:https://github.com/puma/puma[Puma]'s documentation. This is determined using
+the cgroup link:https://www.kernel.org/doc/Documentation/cgroup-v1/cpusets.txt[cpusets]
+subsystem. You can specify the cores that the container is allowed to use by passing
+the `--cpuset-cpus` parameter to the link:http://docker.io[Docker] run command:
```
bparees: we shouldn't be talking about docker run'ing images in the openshift-docs.  we should only doc things that are relevant to using the image in openshift.

using_images/s2i_images/ruby.adoc
```diff
@@ -117,6 +124,40 @@ link:../../dev_guide/environment_variables.html#set-environment-variables[the
...
+
+----
+$ docker run -e PUMA_MAX_THREADS=32 --cpuset-cpus='0-2,3,5' -p 8080:8080 sinatra-app
```
bparees: this would be fine in the sti-ruby repo readme, but not here.

using_images/s2i_images/ruby.adoc
```diff
@@ -117,6 +124,40 @@ link:../../dev_guide/environment_variables.html#set-environment-variables[the
...
+container is computed using the following formula:
+
+----
+50 + 15 * WORKERS + 0.125 * WORKERS * PUMA_MAX_THREADS
+----
+
+You can specify a memory limit using the `--memory` flag:
+
+----
+$ docker run -e PUMA_MAX_THREADS=32 --memory=300m -p 8080:8080 sinatra-app
```
bparees: same thing here.

we should be talking about resource limits/quota, not docker command args

mnagy:
I've struggled about how to make the tuning section relevant for OpenShift, but in the end decided that it would be better to just throw the changes away and make it simple. The end-user doesn't have direct control over the number of cores available on the node running his pod anyway.

using_images/s2i_images/ruby.adoc
```diff
@@ -96,6 +96,12 @@ and ruby gem installation. The application assets are only compiled if
 Asset compilation only happens when the application runs in a production
 environment. Therefore, you can use this variable when assets have already been
 compiled.
+
+|`*PUMA_MIN_THREADS*`, `*PUMA_MAX_THREADS*`
+|Indicates the minimum and maximum threads that will be available in link:https://github.com/puma/puma[Puma]'s thread pool.
+
+|`*PUMA_WORKERS*`
+|Indicates the number of worker processes that will be launched. See documentation on Puma's link:https://github.com/puma/puma#clustered-mode[clustered mode]. This defaults to the number of cores available on the node.
```
bparees: Should mention something about this being also memory based.  doesn't have to be complicated, but:  "If not explicitly set, the default behavior sets PUMA_WORKERS to an appropriate value for the memory available to the container and the number of cores on the host"

mnagy:
That's great, thanks! Updated.

bparees:
lgtm

using_images/s2i_images/ruby.adoc
```diff
@@ -96,6 +96,12 @@ and ruby gem installation. The application assets are only compiled if
 Asset compilation only happens when the application runs in a production
 environment. Therefore, you can use this variable when assets have already been
 compiled.
+
+|`*PUMA_MIN_THREADS*`, `*PUMA_MAX_THREADS*`
+|Indicates the minimum and maximum threads that will be available in link:https://github.com/puma/puma[Puma]'s thread pool.
```
tpoitras: Since it's a link to an external site, you don't need "link:" in the markup for the Puma link, as per https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#links-to-external-websites

tpoitras: Are there any limits on the number that can be specified for each variable? Or is it valid as long as PUMA_MIN_THREADS < PUMA_MAX_THREADS ?

using_images/s2i_images/ruby.adoc
```diff
@@ -96,6 +96,12 @@ and ruby gem installation. The application assets are only compiled if
...
+
+|`*PUMA_WORKERS*`
+|Indicates the number of worker processes that will be launched. See documentation on Puma's link:https://github.com/puma/puma#clustered-mode[clustered mode]. If not explicitly set, the default behavior sets `*PUMA_WORKERS*` to an appropriate value for the memory available to the container and the number of cores on the host.
```
tpoitras: Since it's a link to an external site, you don't need "link:" in the markup for the Puma link, as per https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#links-to-external-websites

using_images/s2i_images/ruby.adoc
mnagy: There are no limits. And I think min thread doesn't necessarily need to be lower than max. It is certainly valid that they are equal, that will cause the number of threads to be exactly that number at all times. I suspect that if min > max, the result will be the same. In any case, we only validate that those values are numbers and if they are, we just pass them to Puma and let it handle the rest.

mnagy:
Removed redundant `link:`'s, PTAL

using_images/s2i_images/ruby.adoc
```diff
@@ -96,6 +96,12 @@ and ruby gem installation. The application assets are only compiled if
 Asset compilation only happens when the application runs in a production
 environment. Therefore, you can use this variable when assets have already been
 compiled.
+
+|`*PUMA_MIN_THREADS*`, `*PUMA_MAX_THREADS*`
+|Indicates the minimum and maximum threads that will be available in https://github.com/puma/puma[Puma]'s thread pool.
+
+|`*PUMA_WORKERS*`
+|Indicates the number of worker processes that will be launched. See documentation on Puma's https://github.com/puma/puma#clustered-mode[clustered mode]. If not explicitly set, the default behavior sets `*PUMA_WORKERS*` to an appropriate value for the memory available to the container and the number of cores on the host.
```
tpoitras: How is clustered mode related? Are worker processes always launched in clustered mode?

tpoitras:
A small question of context for the link to the puma docs, other than that LGTM. :octocat:

using_images/s2i_images/ruby.adoc
mnagy: Clustered mode is when Puma's running more than two processes.

mnagy:
@tpoitras sorry, this fell through the cracks, just now noticed your question :(

mnagy:
@tpoitras ping?

adellape:
@mnagy LGTM. Is this applicable for images available for both OSE 3.2 and 3.1? Trying to figure out which docs to pick this into.

@tpoitras Since you added the `to_followup`, I'll leave the follow up (or removal of the label) to you. :space_invader:

vikram-redhat:
Removing to_followup.

PR #1811, opened 2016-03-25
removed references to global pod definitions, changed to 'pod-level'

jeffvance:
Minor changes to nfs doc per a comment by @pmorie.
@adellape PTAL


adellape:
:+1:

adellape:
revhistory not needed because it will go in a the same time as https://github.com/openshift/openshift-docs/pull/1726, which has a revhistory note.

PR #1813, opened 2016-03-25
Wrong location of registry.crt and registry.key 

Jooho:
Change the location of registry.crt and registry.key to /etc/secrets at step3,4 in 'Securing the Registry' chapter. This doc set secret-volume to '/etc/secrets/' but it does not specify the location when it creates new crt/key and it creates secrets. Therefore, it does not work if I just follow the doc.


ahardin-rh:
@Jooho Thanks! Looks like we took care of this in April:
https://github.com/openshift/openshift-docs/commit/4ea3fde64bf8369246a28224b375a50658adef6a

PR #1820, opened 2016-03-28
rake import_api should work on a different directory

smarterclayton:
Used infrequently, but don't force directory structure.


smarterclayton:
@nhr

nhr:
Looks good, thanks for the PR.

PR #1100, opened 2015-10-21
Docs about deleting tags

0xmichalis:
@soltysh PTAL

Depends on https://github.com/openshift/origin/pull/4120 and https://github.com/openshift/origin/pull/5319

PR #1641, opened 2016-02-23
Fixed errors from GH issue 1294

bfallonf:
A few errors from #1294, as well as some basic edits.

I'll ask for an ack on the GH issue.

PR #1825, opened 2016-03-29
Followup edits to #1100

bfallonf:
Basic followup edits to #1100 , and the little bit above the addition.

@adellape  Good to merge if there's no comments. Encountered that problem with the 4 ='s causing issues with the admonition.


adellape:
LGTM, merging.

> Encountered that problem with the 4 ='s causing issues with the admonition.

@bfallonf How so? I don't recall the issue.

bfallonf:
@adellape We talked about it in #1641 . Having an example command (eg, something using 4 ='s) inside of an admonition makes it render kinda off.

PR #1830, opened 2016-03-29
fix typo

wanghaoran1988:
when a container with request set and no limit set, if it uses more memory than requested, it will not be killed .....


tnguyen-rh:
@wanghaoran1988 
I believe the current text is correct.  The previous paragraph describes the "more than requested" case and the paragraph in question describes the "less than requested case".
@derekwaynecarr 
Could you please confirm?

derekwaynecarr:
The current text is correct, this PR should be closed.

PR #1832, opened 2016-03-29
[WIP]Bug 1318737, added Recovering Deleted Default Projects section

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1318737

### DIFF
--- a/admin_guide/managing_projects.adoc
+++ b/admin_guide/managing_projects.adoc
@@ -309,3 +309,40 @@ in order, this rule should be specified last.
 ====
 
 endif::openshift-origin,openshift-enterprise,atomic-registry[]
+
+[[recovering-deleted-default-projects]]
+
+== Recovering Deleted Default Projects
+
+If you delete any of the default projects like `*default*` or `*openshift*`,
+then you will likely experience issues that impact the whole environment. To
+recover from this, restart the master service:
+
+====
+----
+# systemctl restart atomic-openshift-master
+----
+====
+
+[WARNING]
+====
+The default project `*openshift-infra*` is unrecoverable in OpenShift 3.0 and 3.1.
+====
+
+The master creates the missing namespaces and bindings on startup.
+
+After restarting, verify the namespaces exist and are not in the *Terminating*
+state:
+
+====
+----
+# oc get namespace/default namespace/openshift namespace/openshift-infra
+----
+====
+
+Next, you need to manually recreate the content (including templates, image
+streams, the registry, the router, and so on) by following the steps in
+link:../install_config/install/docker_registry.html[Deploying a Docker
+Registry], link:../install_config/install/deploy_router.html[Deploying a
+Router], and link:../install_config/imagestreams_templates.html[Loading the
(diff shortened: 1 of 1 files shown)


ahardin-rh:
@liggitt @adellape @tnguyen-rh PTAL. Thanks!

admin_guide/managing_projects.adoc
```diff
@@ -309,3 +309,26 @@ in order, this rule should be specified last.
...
+
+====
+----
+# systemctl restart atomic-openshift-master
+----
+====
+
+The master creates the missing namespaces and bindings on startup.
+
+Then, you need to manually recreate the content (including templates, image
```
tnguyen-rh: @ahardin-rh 
Suggest s/Then/Next/

tnguyen-rh:
@ahardin-rh Small suggestion, otherwise LGTM.

ahardin-rh:
@liggitt PTAL. Thanks!

admin_guide/managing_projects.adoc
```diff
@@ -309,3 +309,26 @@ in order, this rule should be specified last.
...
+
+== Recovering Deleted Default Projects
+
+If you delete any of the default projects like `*default*`, `*openshift*`, or
+`*openshift-infra*`, then you will likely experience issues that impact the
+whole environment. To recover from this, restart the master service:
+
+====
+----
+# systemctl restart atomic-openshift-master
```
liggitt: after restarting, verify the projects exist and are not in the terminating state:
```
oc get namespace/default namespace/openshift namespace/openshift-infra
```

If they are still terminating, wait for the termination to complete, then restart the master service again.

ahardin-rh:
@liggitt Thanks! Made some edits.

ahardin-rh:
@liggitt PTAL at QE's comments in the BZ for this work: 
https://bugzilla.redhat.com/show_bug.cgi?id=1318737
I'm not sure how to address this. Any thoughts?

liggitt:
looking at this again, recovering the deleted openshift-infra project will be very difficult... restarting won't help in 3.2. https://github.com/openshift/origin/pull/4318 is opened to protect those two namespaces from deletion.

ahardin-rh:
@liggitt Thanks. I made an edit regarding openshift-infra. However, to be clear, do you think that the recovery steps are not worth documenting here? Do you think that those would be better suited for a KB article?

ahardin-rh:
@liggitt bump

liggitt:
I probably wouldn't put it in the core docs here. With the fix in https://github.com/openshift/origin/pull/4318 it won't be possible to completely delete those namespaces any more

ahardin-rh:
@liggitt  Thanks. Do you think we should add this for 3.0 and 3.1 branches only (omitting this from Origin and 3.2)?

liggitt:
I wouldn't put it in core docs

ahardin-rh:
@liggitt Thanks. Moved it to KB: https://access.redhat.com/solutions/2260211

PR #1839, opened 2016-03-29
updated direct storage example

erinboyd:
Created example with two references:
ClaimRef
Direct storage via plugin

### DIFF
--- a/install_config/storage_examples/direct_storage.adoc
+++ b/install_config/storage_examples/direct_storage.adoc
@@ -1,0 +1,481 @@
+= Direct access to storage using two methods
+{product-author}
+{product-version}
+:data-uri:
+:icons:
+:experimental:
+:toc: macro
+:toc-title:
+:prewrap!:
+
+toc::[]
+
+== Overview
+In OpenShift a Persistent Volume Claim (PVC) will bind to a Persistent Volume (PV) based on matching the
+spec labels of 'accessModes' and 'capacity' values. In some cases, a user may want to be specific about
+the storage they are acessing. This example will show how to a user can prebind a PV to a specifc claim 
+[METHOD 1] as well as go directly to the storage using the volume plugin [METHOD 2] inside the pod spec.
+
+In  this example we are using NFS, but the same method can be use for any of the storage plugins
+Please see the section on link:../persistent_storage/index.html[Persistent Storage Plugins] for the supports types.
+
+
+[NOTE]
+====
+All `oc ...` commands are executed on the OpenShift master node.
+====
+
+== [METHOD 1] Create the Persistent Volume using a predefined Persistent Volume Claim reference
+=== 1. Creating the persistent volume (PV) 
+The user may define the spec to include a claim name (ClaimRef) to restrict this PV to only
+bind to a particular persistent volume claim.
+
+.Persistent Volume Object Definition Using NFS
+====
+
+[source,yaml]
+----
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: nfs-pv-prebind <1>
+spec:
+  capacity:
+    storage: 1Gi <2>
+  accessModes:
+    - ReadWriteMany
+  persistentVolumeReclaimPolicy: Retain <3>
+  ClaimRef:
+    Namespace: default <4>
+    Name: nfs-claim-prebind <5>
+  nfs: <6>
+    path: /opt/nfs
+    server: nfs.f22
+    readOnly: false
+----
+<1> The name of the PV which is referenced in pod definitions or displayed in
+various `oc` volume commands.
+<2> The amount of storage allocated to this volume.
+<3> A volume reclaim policy of "retain" indicates to preserve the volume after the pods 
+<4> This is the namespace the claim will exist within 
+<5> This is the claim name that the PVC will need to match EXACTLY to in it's definition  
+<6> This indicates the volume plugin being used for the persistent volume definition
+====
+
+Save the PV definition to a file, for example *_nfs-pv.yaml_*,
+and create the persistent volume:
+
+====
+----
+# oc create -f nfs-pv.yaml 
+persistentvolume "nfs-pv-prebind" created
+----
+====
+
+Verify that the persistent volume was created:
+
+====
+----
+# oc get pv
+NAME                     LABELS    CAPACITY   ACCESSMODES   STATUS      CLAIM                         REASON    AGE
+nfs-pv-prebind           <none>    1Gi        RWX           Available                                           4s
+----
+====
+
+=== 2. Creating the Persistent Volume Claim (PVC)
+In general a persistent volume claim (PVC) specifies the desired access mode and storage capacity.
+Currently, based on only these two attributes, a PVC is bound to a single PV. However in this example,
+we pre-bound the PVC to our PV we created.
+
+.PVC Object Definition
+====
+[source,yaml]
+----
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: nfs-claim-prebind  <1>
+spec:
+  accessModes:
+  - ReadWriteMany      <2>
+  resources:
+     requests:
+       storage: 1Gi    <3>
+----
+<1> The claim name referenced by the PV spec definition ClaimRef Name.
+<2> As mentioned above for PVs, the `*accessModes*` do not enforce access right,
+but rather act as labels to match a PV to a PVC.
+<3> This claim will look for PVs offering *1Gi* or greater capacity.
+====
+
+Save the PVC definition to a file, for example *_nfs-pvc.yaml_*,
+and create the PVC:
+
+====
+----
+# oc create -f nfs-pvc.yaml 
+persistentvolumeclaim "nfs-claim-prebound" created
+
+#and verify the PVC was created and bound to the expected PV:
+# oc get pvc
+NAME                       LABELS    STATUS    VOLUME               CAPACITY   ACCESSMODES   AGE
+nfs-claim-prebound         <none>    Bound     nfs-pv-prebind       1Gi        RWX           24s
+                                    <1>
+----
+<1> the claim (nfs-claim-prebind) was bound to the "nfs-pv-prebind" PV.
+====
+
+=== 3. Creating the Pod
+A pod definition file or a template file can be used to define a pod. Below is a pod spec that
+creates a single container and mounts the nfs volume for read-write access:
+
+.Pod Object Definition
+====
+[source,yaml]
+----
+apiVersion: v1
+kind: Pod
+metadata:
+  name: nginx-nfs-pod <1>
+  labels:
+    name: nginx-nfs-pod
+spec:
+  containers:
+    - name: nginx-nfs-pod
+      image: fedora/nginx <2>
+      ports:
+        - name: web
+          containerPort: 80
+      volumeMounts:
+        - name: nfsvol <3>
+          mountPath: /usr/share/nginx/html <4>
+  securityContext:
+      supplementalGroups: [100003] <5>
+      privileged: false
+  volumes:
+    - name: nfsvol
+      persistentVolumeClaim:
+        claimName: nfs-claim-prebind
+----
+<1> The name of this pod as displayed by `oc get pod`.
+<2> The image run by this pod.
+<3> The name of the volume. This name must be the same in both the `containers` and `volumes` sections.
+<4> The mount path as seen in the container.
+<5> The group id to be assigned to the container.
+<6> The PVC that was created in the previous step.
+====
+
+Save the pod definition to a file, for example *_nfs.yaml_*,
+ and create the pod:
+
+====
+----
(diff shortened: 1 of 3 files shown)


erinboyd:
@jeffvance @screeley44 @adellape @copejon PTAL

install_config/storage_examples/direct_storage.adoc
```diff
@@ -0,0 +1,481 @@
...
+  4m		4m		1	{kubelet ose70.rh7}	spec.containers{busybox-nfs-pod}	Started		Started with docker id 346d432e5a48
+----
+[NOTE] 
+====
+Instead of a Persistent Volume Claim being listed under the 'Volumes' section we have 'nfsvol' indicating we used the NFS plugin to access the storage asset.
+====
+
+====
+
+== [Security Requirement] NFS Volume Access
```
jeffvance: I think most of this section should be replaced by a link to ../persistent_storage/pod_security_context.html

install_config/storage_examples/direct_storage.adoc
```diff
@@ -0,0 +1,481 @@
...
+Access is necessary to a node in the NFS server for either example. On the node examine
```
jeffvance: Maybe this comment should be sooner in the review... NFS is used as the example here for "direct access to storage" yet we already have (at least) two NFS docs: 1) ../persistent_storage/persistent_storage_nfs.html, and shared_storage.html. Disucssion here about an NFS-specific selinux boolean setting, fuse mount, etc are specific to NFS. So I don't understand why this doc's title is generic, nor why some of the content in this doc isn't just moved to the storage_examples/shared_storage.html doc.

install_config/storage_examples/direct_storage.adoc
```diff
@@ -0,0 +1,481 @@
...
+
+toc::[]
+
+== Overview
+In OpenShift a Persistent Volume Claim (PVC) will bind to a Persistent Volume (PV) based on matching the
+spec labels of 'accessModes' and 'capacity' values. In some cases, a user may want to be specific about
+the storage they are acessing. This example will show how to a user can prebind a PV to a specifc claim 
+[METHOD 1] as well as go directly to the storage using the volume plugin [METHOD 2] inside the pod spec.
+
+In  this example we are using NFS, but the same method can be use for any of the storage plugins
```
jeffvance: except selinux bool mentioned at the end is specific to nfs... and non-shared storage may have some unique considerations.

install_config/storage_examples/direct_storage.adoc
```diff
@@ -0,0 +1,481 @@
...
+:icons:
+:experimental:
+:toc: macro
+:toc-title:
+:prewrap!:
+
+toc::[]
+
+== Overview
+In OpenShift a Persistent Volume Claim (PVC) will bind to a Persistent Volume (PV) based on matching the
```
jeffvance: consider adding links to ../storage_examples/shared_storage.html and ../persistent_storage/persistent_storage_nfs.htlm and /architecture/additional_concepts/storage.html.

jeffvance: Consider adding a note that both of the methods you describe may be not be directly supported in future releases since 1) defining the plug-in directly in the pod spec will,  by default, not be allowed (need SCC setting to allow it), and 2) there is pressure to remove support for ClaimRef in a PV spec.

install_config/storage_examples/direct_storage.adoc
```diff
@@ -0,0 +1,481 @@
+= Direct access to storage using two methods
```
jeffvance: The title is a bit confusing to me. It's not clear what "direct access" means, though I understand it from your examples. Maybe:
"Ensuring a Pod has Access to the Target Persistent Storage"  -- though that's not great...

install_config/storage_examples/direct_storage.adoc
```diff
@@ -0,0 +1,481 @@
...
+
+
+[NOTE]
+====
+All `oc ...` commands are executed on the OpenShift master node.
+====
+
+== [METHOD 1] Create the Persistent Volume using a predefined Persistent Volume Claim reference
+=== 1. Creating the persistent volume (PV) 
+The user may define the spec to include a claim name (ClaimRef) to restrict this PV to only
```
jeffvance: "user" is vague. The cluster or storage admin can do this but typically not the developer user.

install_config/storage_examples/direct_storage.adoc
```diff
@@ -0,0 +1,481 @@
...
+    Namespace: default <4>
+    Name: nfs-claim-prebind <5>
+  nfs: <6>
+    path: /opt/nfs
+    server: nfs.f22
+    readOnly: false
+----
+<1> The name of the PV which is referenced in pod definitions or displayed in
+various `oc` volume commands.
+<2> The amount of storage allocated to this volume.
```
jeffvance: Though this is really just a selector-like label and there is no concept of allocating physical storage.

install_config/storage_examples/direct_storage.adoc
```diff
@@ -0,0 +1,481 @@
...
+<3> A volume reclaim policy of "retain" indicates to preserve the volume after the pods 
+<4> This is the namespace the claim will exist within 
+<5> This is the claim name that the PVC will need to match EXACTLY to in it's definition
```
jeffvance: Not only does the claim name need to match, but also the capacity defined in the PVC must be <= to the PV's capacity and the PVC's accessModes must be compatible with the access label defined in the PV. So all three must match (today).

install_config/storage_examples/direct_storage.adoc
```diff
@@ -0,0 +1,481 @@
...
+  name: nfs-claim-prebind  <1>
+spec:
+  accessModes:
+  - ReadWriteMany      <2>
+  resources:
+     requests:
+       storage: 1Gi    <3>
+----
+<1> The claim name referenced by the PV spec definition ClaimRef Name.
+<2> As mentioned above for PVs, the `*accessModes*` do not enforce access right,
```
jeffvance: I don't see this mentioned above...

install_config/storage_examples/direct_storage.adoc
```diff
@@ -0,0 +1,481 @@
...
+  volumes:
+    - name: nfsvol
+      persistentVolumeClaim:
+        claimName: nfs-claim-prebind
+----
+<1> The name of this pod as displayed by `oc get pod`.
+<2> The image run by this pod.
+<3> The name of the volume. This name must be the same in both the `containers` and `volumes` sections.
+<4> The mount path as seen in the container.
+<5> The group id to be assigned to the container.
```
jeffvance: There has been, up to this point, no mention of group id 100003 so it seems "out of the blue" here.

install_config/storage_examples/direct_storage.adoc
```diff
@@ -0,0 +1,481 @@
...
+    nfs:
+      path: /opt/nfs  <6>   
+      server: nfs.f22 <7>
+
+----
+<1> The name of this pod as displayed by `oc get pod`.
+<2> The image run by this pod.
+<3> The name of the volume. This name must be the same in both the `containers` and `volumes` sections.
+<4> The mount path as seen in the container.
+<5> The group id to be assigned to the container.
```
jeffvance: same group id comment from above...

install_config/storage_examples/direct_storage.adoc
```diff
@@ -0,0 +1,481 @@
...
+<6> The path to the storage on the NFS server
```
jeffvance: some the numbered notes end in a period and others don't. I think the approach in these docs has been to end these sentences with a period.

install_config/storage_examples/direct_storage.adoc
```diff
@@ -0,0 +1,481 @@
...
+In order to access the NFS mount, the container must match the SELinux label, and
+either run with a UID of 0, or with 100003 in its supplemental groups range. It is recommended to gain
+access to the volume by matching the NFS mount's groups, which will be defined in the pod
+definition below.
+
+By default, SELinux does not allow writing from a pod to a remote NFS server. To enable
+writing to NFS volumes with SELinux enforcing on each node, run:
+
+----
+# setsebool -P virt_sandbox_use_nfs on
```
jeffvance: The ../persistent_storage/persistent_storage_nfs.html doc shows setting two se bools:
# setsebool -P virt_use_nfs 1
# setsebool -P virt_sandbox_use_nfs 1

install_config/storage_examples/direct_storage.adoc
adellape: "Targeting a Specific PVC"
"Consuming a Specific PVC"
"Consuming a Target PVC"
?

install_config/storage_examples/direct_storage.adoc
adellape: We've been adding a line to dissuade "1)" to the storage provider docs recently, e.g. in https://docs.openshift.com/enterprise/latest/install_config/persistent_storage/persistent_storage_nfs.html:

"While the NFS-specific information contained in a PV definition could also be defined directly in a pod definition, doing so does not create the volume as a distinct cluster resource, making the volume more susceptible to conflicts."

install_config/storage_examples/direct_storage.adoc
adellape: I think we could probably just drop this section? It's stated in the Overview the example uses NFS, and the NFS doc already has its own NFS-specific "Volume Security" section (https://docs.openshift.com/enterprise/latest/install_config/persistent_storage/persistent_storage_nfs.html#nfs-volume-security), so that seems like enough and that it would just be a repeat here. (Plus there's the full, dedicated "Volume Security" topic at https://docs.openshift.com/enterprise/latest/install_config/persistent_storage/pod_security_context.html, as already mentioned.)

adellape:
@erinboyd Couple more comments inline, adding to @jeffvance's.

install_config/storage_examples/direct_storage.adoc
erinboyd: Because the shared example is showing how to share the storage between two pods. This is showing how you use a specific storage asset. They are very different.

install_config/storage_examples/direct_storage.adoc
erinboyd: @adellape why would it be more susceptible to conflicts. I actually think using the PV/PVC is more prone to this from a user's eyes/

install_config/storage_examples/direct_storage.adoc
erinboyd: Well, one uses the plugin directly and completely bypasses the pv/pvc, that is why I left it out of the title

install_config/storage_examples/direct_storage.adoc
erinboyd: Removed

install_config/storage_examples/direct_storage.adoc
erinboyd: I don't think I want to add anything about releases, etc...I just want to include what is currently supported. If we deprecate it later, we will notate that. But as of now it seems pretty inconclusive.

install_config/storage_examples/direct_storage.adoc
jeffvance: Then I think the title should be less generic. Perhaps "Sharing NFS Storage Across Two Pods"?

install_config/storage_examples/direct_storage.adoc
jeffvance: "for *any* of the storage plugins" is what I object to. This doc covers NFS. Youl will need to do things a bit differently for GlusterFs.

install_config/storage_examples/direct_storage.adoc
```diff
@@ -0,0 +1,481 @@
...
+  name: nfs-pv-prebind <1>
+spec:
+  capacity:
+    storage: 1Gi <2>
+  accessModes:
+    - ReadWriteMany
+  persistentVolumeReclaimPolicy: Retain <3>
+  ClaimRef:
+    Namespace: default <4>
+    Name: nfs-claim-prebind <5>
```
sosiouxme: This does not work as far as I can tell, and @markturansky confirms (with code reference) that it shouldn't. When you create the PV with a `ClaimRef` (or even a `claimRef`) it is silently ignored and binding occurs as usual. The claimRef isn't even displayed in the PV API object you end up with.

markturansky: Correct. This requires the UID of the claim to be properly bound.

jeffvance: The case of the stanza won't work in v 3.2 (AFIK). Needs to be:
  claimRef:
    name:
    namespace:

ahardin-rh:
@erinboyd Should this be closed? Thanks!

adellape:
Spoke w/ @erinboyd today, closing this one.

vikram-redhat:
Since this is closed, I will remove the to_followup tag.

PR #1851, opened 2016-04-01
Replace .sti -> .s2i in docs

rhcarvalho:
Fixes https://github.com/openshift/sti-python/issues/106.

@bparees @soltysh PTAL. Safe to point users to use `.s2i` instead of the deprecated `.sti`, correct?

### DIFF
--- a/creating_images/s2i.adoc
+++ b/creating_images/s2i.adoc
@@ -63,7 +63,7 @@ executable inside the builder image. S2I supports multiple options providing
 each build in the following order:
 
 1. A script found at the `--scripts-url` URL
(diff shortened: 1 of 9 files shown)


bparees:
lgtm.  @openshift/team-documentation ptal

adellape:
Tried this out in 3.2 (w/ nodejs builder) and it worked, didn't work in 3.1. So I take it this is 3.2+ only.

@rhcarvalho @bparees LMK if i'm mistaken.

bparees:
forget exactly when we made the change but yeah that's plausible.

PR #1863, opened 2016-04-06
Changing example to a matching configuration for docker-registry pvc

akram:
The provided example for docker-registry using a PVC is not correct and references a volume-named called v1 where the correct name is docker-registry.
This is error prone.


ahardin-rh:
Thanks! Sorry for the delay!

ahardin-rh:
[rev_history]
|link:../install_config/install/docker_registry.html[Deploying a Docker Registry]
|Updated the example of using an existing persistent volume claim (PVC) to a matching configuration for docker-registry PVC.
%

Issue #1866, opened 2016-04-06
Issue with parallel image pulls documentation


rikatz:
So, on the docs, the suggested usage of parallel image pulls is as the following:

kubeletArguments:
  serialize-image-pulls:
  - true

This is a bit config, the default configuration enables the parallel image pulls? Is this safe, on an environment as RHEL7/CentOS 7 that uses Docker 1.8.2?

Also, changing the config to the suggested will not work, apparently Openshift expects an YAML with the 'true' (boolean) between quotation marks (as this is an array), so the correct would be:

kubeletArguments:
  serialize-image-pulls:
  - 'true'

Is this right? Thanks!!

install_config/master_node_configuration.adoc

adellape:
> Also, changing the config to the suggested will not work

@rikatz Did you get an error? The master service restarted fine for me using the doc's syntax.

> the default configuration enables the parallel image pulls? Is this safe, on an environment as RHEL7/CentOS 7 that uses Docker 1.8.2?

The way I read it, it sounds like the default configuration uses serialized (i.e., "pull images one at a time"), as shown in the example. The <1> callout in the example is saying to change "true" to "false" to enable parallel.

cc @aveshagarwal @ncdc

ncdc:
You risk image corruption if you set serialize-image-pulls to false with Docker prior to version 1.9.

ncdc:
Each entry under `kubeletArguments` does require an array of strings, so doing `"true"` makes sense.

aveshagarwal:
To enable parallel image pulls:

```
kubeletArguments:
  serialize-image-pulls:
  - "false"
```
Or

```
kubeletArguments:
  serialize-image-pulls:
  - 'false'
```

Both works. Also by default, serial image pulls is true.

rikatz:
@adellape Yes, I get an error, as shown here by @ncdc and @aveshagarwal probably the quotation is mandatory.

Just need to fix this on the Docs :) Do you wan't me to open a PR to fix this?

Also, need to clarify that the 'true' is the default configuration, and that to change this config to 'parallel' this need to me changed to 'false'

aveshagarwal:
@rikatz  Yes the doc needs to be corrected. Please open a PR if you can. 
Yes you are correct about the other part.

PR #1870, opened 2016-04-06
Improve the 'Parallel image pulls' documentation

rikatz:
This improves the 'Parallel Image Pulls' configuration, clarifying the default behaviour, changing the parameter to 'false' (that enables the described feature) and also adding 'quotes' on the selected behaviour (as this is a array of strings, not a boolean).

It solves the issoe #1866

### DIFF
--- a/install_config/master_node_configuration.adoc
+++ b/install_config/master_node_configuration.adoc
@@ -354,9 +354,9 @@ If you are using Docker 1.9+, you may want to consider enabling parallel image p
 ----
 kubeletArguments:
   serialize-image-pulls:
-  - true <1>
+  - "false" <1>
 ----
-<1> Change to false to enable parallel pulls.
+<1> Change to true to disable parallel pulls. (This is the default config)
 ====
 
 [[passwords]]


ahardin-rh:
@rikatz Sorry for the delay. Thanks! :thumbsup:

ahardin-rh:
[rev_history]
|link:../install_config/master_node_configuration.html[Master and Node Configuration]
|Updated the setting guidance in link:../install_config/master_node_configuration.html#master-node-configuration-parallel-image-pulls-with-docker[Parallel Image Pulls with Docker 1.9+].
%

install_config/master_node_configuration.adoc
```diff
@@ -354,9 +354,9 @@ If you are using Docker 1.9+, you may want to consider enabling parallel image p
 ----
 kubeletArguments:
   serialize-image-pulls:
-  - true <1>
+  - "false" <1>
 ----
-<1> Change to false to enable parallel pulls.
+<1> Change to true to disable parallel pulls. (This is the default config)
```
aditya-konarde: @ahardin-rh hello, I reached here looking for the same clarification. 

Saying that 'this' is the default configuration is a bit confusing, since we've noted both cases of it being false and true in the example and comment respectively. 

Would it be better worded if we say:

By default, this flag is set to `foo` (true or false, not sure which one)

PR #1876, opened 2016-04-08
Bug 1305403 - Describe router sharding

tnguyen-rh:
- Add section "Router Sharding" to arch "Routes" chapter
- Add section "Configuring Router Sharding"

@ramr WDYT?

### DIFF
--- a/architecture/core_concepts/routes.adoc
+++ b/architecture/core_concepts/routes.adoc
@@ -610,3 +610,30 @@ termination.
 validate the endpoint certificate, securing the connection from the router to
 the destination. This field is required, but only for re-encryption.
 ====
+
+
+[[router-sharding]]
+== Router Sharding
+
+In {product-title}, each route can have any number of
+link:../../architecture/core_concepts/pods_and_services.html#labels[labels]
+in its `metadata` field.
+A router selects which subset of routes from the entire pool of routes
+to serve via a _selection expression_ involving those labels.
+A selection expression can also involve
+labels on the route's namespace.
+The selected routes form a _router shard_.
+ifdef::openshift-enterprise,openshift-origin[]
+You can
+link:../../install_config/install/deploy_router.html#creating-router-shards[create]
+and
+link:../../install_config/install/deploy_router.html#modifying-router-shards[modify]
+router shards independently from the routes, themselves.
+endif::[]
+
+This design allows for, but does not obligate,
+a route to _possibly_ belong to many different shards.
+For example, a single route may belong to a `SLA=high` shard
+(but not `SLA=medium` or `SLA=low` shards),
+as well as a `geo=west` shard
+(but not a `geo=east` shard).
--- a/install_config/install/deploy_router.adoc
+++ b/install_config/install/deploy_router.adoc
@@ -268,6 +268,222 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
 ----
 ====
 
+
+[[creating-router-shards]]
+=== Creating Router Shards
+
+Router sharding is
+link:../../architecture/core_concepts/routes.html#router-sharding[based on labels];
+you set labels on the routes in the pool,
+and express the desired subset of those routes for the router to serve
+with a selection expression via the `oc env` command.
+
+The rest of this section describes an extended example.
+Suppose there are 26 routes, named `a` -- `z`,
+in the pool, with various labels:
+
+.Possible labels on routes in the pool
+----
+sla=high       geo=east     hw=modest     dept=finance
+sla=medium     geo=west     hw=strong     dept=dev
+sla=low                                   dept=ops
+----
+
+These labels express the concepts:
+service level agreement, geographical location,
+hardware requirements, and department.
+The routes in the pool can have at most one label from each column.
+Some routes may have other labels, entirely, or none at all.
+
+[options="header",cols="1,1,1,1,1,3"]
+|===
+|Name(s) |SLA |Geo |HW |Dept |Other Labels
+
+|`a`
+|`high`
+|`east`
+|`modest`
+|`finance`
+|`type=static`
+
+|`b`
+|
+|`west`
+|`strong`
+|
+|`type=dynamic`
+
+|`c`, `d`, `e`
+|`low`
+|
+|`modest`
+|
+|`type=static`
+
+|`g` -- `k`
+|`medium`
+|
+|`strong`
+|`dev`
+|
+
+|`l` -- `s`
+|`high`
+|
+|`modest`
+|`ops`
+|
+
+|`t` -- `z`
+|
+|`west`
+|
+|
+|`type=dynamic`
+
+|===
+
+Here is a convenience script *_mkshard_*  that
+ilustrates how `oadm router`, `oc env`, and `oc scale`
+work together to make a router shard.
+
+====
+[source,bash]
+----
+#!/bin/bash
+# Usage: mkshard ID SELECTION-EXPRESSION
+id=$1
+sel="$2"
+router=router-shard-$id           //<1>
+oadm router $router --replicas=0  //<2>
+dc=dc/router-shard-$id            //<3>
+oc env   $dc ROUTE_LABELS="$sel"  //<4>
+oc scale $dc --replicas=3         //<5>
+----
+<1> The created router has name `router-shard-<id>`.
+<2> Specify no scaling for now.
+<3> The deployment configuration for the router.
+<4> Set the selection expression using `oc env`.
+    The selection expression is the value of
+    the `ROUTE_LABELS` environment variable.
+<5> Scale it up.
+====
+
+Running *_mkshard_* several times creates several routers:
+
+[options="header",cols="2,3,2"]
+|===
+|Router |Selection Expression |Routes
+
+|`router-shard-1`
+|`sla=high`
+|`a`, `l` -- `s`
+
+|`router-shard-2`
+|`geo=west`
+|`b`, `t` -- `z`
+
+|`router-shard-3`
+|`dept=dev`
+|`g` -- `k`
+
+|===
+
+
+[[modifying-router-shards]]
+=== Modifying Router Shards
+
+Because a router shard is a construct
+link:../../architecture/core_concepts/routes.html#router-sharding[based on labels],
+you can modify either the labels (via
+link:../../cli_reference/basic_cli_operations.html#application-modification-cli-operations[`oc label`])
+or the selection expression.
+
+This section extends the example started in the
+link:#creating-router-shards[Creating Router Shards] section,
+demonstrating how to change the selection expression.
+
+Here is a convenience script *_modshard_* that modifies
+an existing router to use a new selection expression:
+
+====
+[source,bash]
+----
+#!/bin/bash
+# Usage: modshard ID SELECTION-EXPRESSION...
+id=$1
+shift
+router=router-shard-$id       //<1>
+dc=dc/$router                 //<2>
+oc scale $dc --replicas=0     //<3>
+oc env   $dc "$@"             //<4>
+oc scale $dc --replicas=3     //<5>
+----
+<1> The modified router has name `router-shard-<id>`.
+<2> The deployment configuration where the modifications occur.
+<3> Scale it down.
+<4> Set the new selection expression using `oc env`.
+    Unlike `mkshard` from the
+    link:#creating-router-shards[Creating Router Shards]
+    section, the selection expression specified as the
+    non-`ID` arguments to `modshard` must include the
+    environment variable name as well as its value.
+<5> Scale it back up.
+====
+
+[NOTE]
+====
+In `modshard`, the `oc scale` commands are not necessary if the
+link:../../dev_guide/deployments.html#strategies[deployment strategy]
+for `router-dhsard-<id>` is `Rolling`.
+====
+
+For example, to expand the department for `router-shard-3`
+to include `ops` as well as `dev`:
+
+----
+$ modshard 3 ROUTE_LABELS='dept in (dev, ops)'
+----
+
+The result is that `router-shard-3` now selects routes `g` -- `s`
+(the combined sets of `g` -- `k` and `l` -- `s`).
+
+This example takes into account that
+there are only three departments in this example scenario,
+and specifies a department to leave out of the shard,
+thus achieving the same result as the preceding example:
+
+----
+$ modshard 3 ROUTE_LABELS='dept != finanace'
+----
+
+This example specifies shows three comma-separated qualities,
+and results in only route `b` being selected:
+
+----
+$ modshard 3 ROUTE_LABELS='hw=strong,type=dynamic,geo=west'
+----
+
+Similarly to `ROUTE_LABELS`, which involve a route's labels,
+you can select routes based on the labels of the route's namespace labels,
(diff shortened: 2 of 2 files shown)


tnguyen-rh:
@bfallonf PTAL

install_config/install/deploy_router.adoc
```diff
@@ -230,6 +230,45 @@ endif::[]
...
 router] on your OpenShift cluster using IP failover.
 
+
+[[configuring-router-sharding]]
+=== Configuring Router Sharding
+
+Router sharding is
+link:../../architecture/core_concepts/routes.html#router-sharding[based on labels];
+you set labels on the router
+and make sure there is a corresponding node selector
```
bfallonf: Might be a good chance to link to the node selector docs:

https://docs.openshift.com/enterprise/3.1/admin_guide/managing_projects.html#using-node-selectors

install_config/install/deploy_router.adoc
```diff
@@ -230,6 +230,45 @@ endif::[]
...
+when enabling that router.
+This can be done with a single command, per shard.
+The following creates three shards, by geo.
+
+----
+$ router-shard latam
+$ router-shard emea
+$ router-shard apac
+----
```
bfallonf: If this creates a router shard (I'm guessing called the `latam` or whatever), then what do you do? What does this do to your OpenShift instance? The answer to this might be better at home in the core_concepts file above. 

Also, if this section really only mentions how to create a shard, should it be called "Creating Router Shards" instead of "Configuring Router Sharding"?

install_config/install/deploy_router.adoc
```diff
@@ -230,6 +230,45 @@ endif::[]
...
+
+It uses the convenience script `router-shard`:
+
+====
+[source,bash]
+----
+#!/bin/bash
+geo=$1
+oadm router router-$geo --replicas=3 \
```
bfallonf: Does this need a # at the start? Perhaps not, cos it's in a script-thing, but I thought I'd ask.

architecture/core_concepts/routes.adoc
```diff
@@ -605,3 +605,12 @@ termination.
...
 ====
+
+
+[[router-sharding]]
+== Router Sharding
+
+In OpenShift, router sharding is implemented via
+link:../../architecture/core_concepts/pods_and_services.html#labels[labels]
+in the router's `metadata` field.
+These labels can be used to specialize the service selector.
```
bfallonf: I feel like this could really benefit from an example. I'm actually not following this at all. Also, does "service selector" = "node selector"?

bfallonf:
@tnguyen-rh Made a few comments. Let me know if anything doesn't make sense.

install_config/install/deploy_router.adoc
ramr: This needs some more clarification - its not labels on the router but rather on the routes themselves.
We are basically saying: you can run multiple routers with a route selection criteria that restricts the routers to serving a subset of the routes - a divide and conquer strategy ... taking a set and breaking it up into smaller sets (subsets if you will) and running multiple routers to service those each of those subsets. 

Example a simple naive method would be to alphabetically split routes names starting with A-J get served by router rs1, L-R by router rs2 and the rest by router rs3. 
Or you could even define a custom more complicated method based on SLAs on routes - let's assume you have some routes that have labels on them re: what SLA they need - and let's use high, medium, low as the tag values. We could now select all routes with SLA=high by routers that run on `beefier` (more performant) instances and similarly select low and medium by routers that run on appropriate instances. 
Or you could do this based on namespaces as well (namespace=finance is more critical than namespace=dev ... damn, that was too close - reality h{ur,i}ts!!). 

In essence, a method to slice and dice routes that an individual router (note there could be many pods backing each  router) would service. And if you create these subsets appropriately, you can cover all the routes with different router instances.

Some examples covering the above:
```  
#  The general pool of routers (3)  - where SLA is not high.  
#  Assuming that the routes that need a high SLA have a "SLA=high" label on them, you can run:    
oadm router rone --replicas=0  ...   
oc edit dc/rone  ROUTE_LABELS="SLA != high"     
oc scale dc/rone --replicas=3  
  
# The higher SLA set + run on beefier instances.
oadm router rtwo --selector="zone=beefier-instances" - --replicas=0  ...   
oc edit dc/rtwo  ROUTE_LABELS="SLA=high"    
oc scale dc/rtwo --replicas=7  

#  This set is how to divide the routes based on namespaces.  
#  You could do it based on projects as well. The ones from finance and prod.  
oadm router cthree -replicas=0  ...   
oc edit dc/cthree  NAMESPACE_LABELS="name in (finance, prod)"    
oc scale dc/cthree --replicas=10  

#  The ones from marketing.
oadm router p0 --replicas=0  ...   
oc edit dc/p0 NAMESPACE_LABELS="name=marketing"    
oc scale dc/p0 --replicas=5  

#  The general ones.   
oadm router bb8 --replicas=0  ...   
oc edit dc/bb8  NAMESPACE_LABELS="name not in (finance, prod, marketing)"    
oc scale dc/bb8--replicas=2  

#  Other possibilities include using PROJECT_LABELS, ROUTE_FIELDS and just running 
#  in the scope of a single namespace.  
```   

Hopefully that explained it better.

tnguyen-rh:
@ramr Thanks for the feedback.

My understanding now is that labels are on the routes and not the routers.  I've largely rewritten the section (now renamed from "Configuring Router Sharding" to "Creating Router Shards") with this model in mind.  There are some **TODO** items that i will look to finish tomorrow.

Questions:
- Previously, there was `--labels` in the `oadm router` command in the `mkshard` script.  I've now removed that under the assumption that specifying it there would add labels to the router (old thinking) rather than the route (current thinking).  Is this correct?
- The extended example defines and makes use of labels attached to routes.  How they are attached is not documented here.  Should it be?
- My understanding is that `ROUTE_LABELS  are the most suitable env var to use for this extended example.  Is that correct?
- The other possibilities (`ROUTE_FIELDS`, `PROJECT_LABELS`, and `NAMESPACE_LABELS`) are all unsuitable for this example since they specify areas to match against other than the labels on routes.  Is that correct?
- I see in the source code (*pkg/cmd/infra/router/router.go*) that the env vars are each associated with a command-line option.  For example, `ROUTE_FIELDS` is associated with `--fields`.  Which command are the options associated with?  (I'm guessing `oadm router`.)
- If those options are indeed specifiable w/ `oadm router` does that mean there are two paths to creating shards (e.g., `oadm router` w/ `--fields` vs `oc edit` w/ `ROUTE_FIELDS`)?  If so, are these identical?  If not, what is the difference?

I hope these questions are clear, and that i will be able to understand your answers! :smile:

tnguyen-rh:
@ramr

BTW, since the source has a large table, it is not easy to understand at a glance.  I've rendered the *deploy_router.adoc* file and placed the output [here](http://file.brq.redhat.com/tnguyen/bz1305403-router-sharding/deploy_router.html#creating-router-shards).

tnguyen-rh:
@ramr

More changes (but the questions above still pertain).
The directory:
http://file.brq.redhat.com/tnguyen/bz1305403-router-sharding
now contains two rendered files.

tnguyen-rh:
@ramr PTAL

Another request: Could you please post an example that uses `ROUTE_FIELDS`?  (I presume `NAMESPACE_LABELS` and `PROJECT_LABELS` are nothing exotic.)

install_config/install/deploy_router.adoc
```diff
@@ -265,6 +265,193 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+----
+#!/bin/bash
+# Usage: mkshard ID SELECTION-EXPRESSION
+id=$1
+sel="$2"
+oadm router router-shard-$id \//<1>
+    --selector="$sel" \//<2>
+    --replicas=3 \
+    --credentials=/etc/origin/master/openshift-router.kubeconfig \
+    --service-account=ipfailover
```
ramr: You don't need to specify `--credentials` and `--service-account` unless you want to override the defaults. I'd just drop those from the example here.

install_config/install/deploy_router.adoc
```diff
@@ -265,6 +265,193 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+
+----
+$ modshard 3 ROUTE_LABELS='dept != finanace'
+----
+
+This example specifies shows three comma-separated qualities,
+and results in only route `b` being selected:
+
+----
+$ modshard 3 hw=strong,type=dynamic,geo=west
```
ramr: Typo missing environment variable name for `hw=strong,type=dynamic,geo=west`.  
Ala: `ROUTE_LABELS="hw=strong,type=dynamic,geo=west"`

ramr:
@tnguyen-rh  let's chat in the meeting on all the points - hopefully that will clarify all your questions. Thx

tnguyen-rh:
@ramr @bfallonf

Thanks for the reviews and (extensive :smile:) discussion.  I have just squashed and will remove the WIP from the title shortly.  This PR is ready for review.

install_config/install/deploy_router.adoc
```diff
@@ -265,6 +265,217 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+----
+#!/bin/bash
+# Usage: modshard ID SELECTION-EXPRESSION...
+id=$1
+shift
+router=router-shard-$id       //<1>
+dc=dc/$router                 //<2>
+oc scale $dc --replicas=0     //<3>
+oc env   $dc "$@"             //<4>
+oc scale $dc --replicas=3     //<5>
```
ramr: I would actually drop the 2 lines `oc scale $dc --replicas=0` and `oc scale $dc --replicas=3` from the `modshard` script. 
Its done for the `mkshard` as we start with 0 replicas. If you change a deployment config (`oc env` will change `$dc`), a new deployment would be automatically triggered. So its not required to scale down/up as the rolling deployment strategy will take care of that. 

So something like:
```  
#!/bin/bash  
# Usage: modshard ID SELECTION-EXPRESSION...   
id=$1  
shift  
router=router-shard-$id       //<1>  
dc=dc/$router                 //<2>  
oc env   $dc "$@"             //<4> or  the new <3>  
exit $?  
``` 
That makes it simpler as well.

ramr:
LGTM - minor comment/cleanup to remove some lines (and still provide the same behavior).

install_config/install/deploy_router.adoc
```diff
@@ -265,6 +265,217 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+|
+|`west`
+|
+|
+|`type=dynamic`
+
+|===
+
+Here is a convenience script *_mkshard_*  that
+ilustrates how `oadm router`, `oc env` and `oc scale`
```
bfallonf: Oxford comma needed above

install_config/install/deploy_router.adoc
```diff
@@ -265,6 +265,217 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+(the combined sets of `g` -- `k` and `l` -- `s`).
+
+This example takes into account that
+there are only three departments in this example scenario,
+and specifies a department to leave out of the shard,
+thus achieving the same result as the preceding example:
+
+----
+$ modshard 3 ROUTE_LABELS='dept != finanace'
+----
```
bfallonf: Should the above just be 'finance'?

bfallonf:
@tnguyen-rh Aside from the two comments above, this looks fine sentence-level-ey. However, I have no idea what's going on. I think this is asking the reader to understand the two scripts you've given examples for, and if they're like me, they have no idea what each step is doing (there is the footnotes, but I'm still not 100% sure). This might be why we don't do this in other parts of the docs and we have procedures and the like. 

If this is supposed to be a technical process, and a reader who wants router sharding in their instance is expected to know their stuff, this is probably all good. If this is meant for readers with not a lot of technical know-how, I think more context might need to be given...

install_config/install/deploy_router.adoc
tnguyen-rh: @bfallonf Thanks, good catch.

install_config/install/deploy_router.adoc
tnguyen-rh: @bfallonf I don't think so.  The "language feature" highlighted here is the `!=` operator, which i believe means "does not equal".  So:

    ROUTE_LABELS='dept=finance'

is valid, but selects those routes key `dept` *equal to* `finance`, which is opposite of the preceding example (and contradicts the example's intro blurb).  Thus, to complete the logical chain, the desired operator should be flipped from `=` to `!=`.

tnguyen-rh:
@bfallonf Does the new "Router Sharding" section in Architecture "Routes" chapter (i.e., `architecture/core_concepts/routes.adoc`) help explain the underlying concepts?  This is, btw, prettily rendered [here](http://file.brq.redhat.com/~tnguyen/bz1305403-router-sharding/routes.html#router-sharding).

install_config/install/deploy_router.adoc
tnguyen-rh: @ramr Thanks for the feedback.

I think dropping the `exit $?` is fine.  I will change that shortly.

However, i'm not convinced abut avoiding `oc scale` altogether due to the `Rolling` strategy handling that automagically.  As `Rolling` is the default deployment strategy, i'd prefer to mention in the docs that the script can be simplified by omitting those lines.  This way readers who use `Rolling` can benefit, as well as those who don't.

tnguyen-rh:
@ramr @bfallonf Feedback items addressed.  PTAL.

bfallonf:
@tnguyen-rh Yeah I think that works. Should they link to each other? Or maybe use includes to have the info? I think that might make it easier to understand. Other than that, LGTM.

tnguyen-rh:
@bfallonf 
There are two links from inst/cfg to arch, and thanks to your suggestion, two (new ones) in the other direction -- see https://github.com/openshift/openshift-docs/pull/1876/commits/3c22f9fd31ff3ade93caaf9a048b547a7c6d6305).  NB, the [pretty render](http://file.brq.redhat.com/~tnguyen/bz1305403-router-sharding/routes.html) was built w/ attributes `openshift-enterprise` and `product-title="OpenShift Enterprise"`.

bfallonf:
@tnguyen-rh Groovy. This looks great now. Big :+1: from me!

ramr:
LGTM

tnguyen-rh:
[rev_history]
|link:../architecture/core_concepts/routes.html[Core Concepts -> Routes]
|Added the link:../architecture/core_concepts/routes.html#router-sharding[Router Sharding] section.
%
|link:../install_config/install/deploy_router.html[Installing -> Deploying a Router]
|Added sections describing how to link:../install_config/install/deploy_router.html#creating-router-shards[create] and link:../install_config/install/deploy_router.html#modifying-router-shards[modify] router shards.
%

tnguyen-rh:
@adellape @vikram-redhat
IIUC, this PR is supposed to be "async 3.2".  Is "Next Release" the proper milestone?

adellape:
@tnguyen-rh Yeah, we won't make a new version-specific milestone until we get closer to another numbered release.

adellape:
To explain the current `to_followup` label, Thien-Thi is going to remove it after QE has taken a look. We just didn't want this to get downstream too early.

PR #1877, opened 2016-04-08
Add in extra warnings for Cassandra and its disk size.

mwringe:
We need to make it more clear that Cassandra can lose metric data if its persistent volume becomes full.


adellape:
[rev_history]
|link:../install_config/cluster_metrics.html[Enabling Cluster Metrics]
|Added extra warnings for Cassandra and its disk size in the link:../install_config/cluster_metrics.html#metrics-persistent-storage[Persistent Storage] link:../install_config/cluster_metrics.html#deployer-template-parameters[Deployer Template Parameters] sections.
%

PR #1879, opened 2016-04-08
clarify how to avoid copying the base directory

### LINKED ISSUE #8383: Image source's sourcePath doesn't allow for wildcard copy
I want to copy all the contents of a directory in a source image into my destination image, but it's not possible

##### Version
oc v1.1.4
kubernetes v1.2.0-origin-41-g91d3e75

##### Steps To Reproduce
1. Create an image with files and dirs in /files

````
$ ls ./files
Dockerfile  
apply*
config/
```` 

````
FROM scratch
COPY apply /files/config
COPY config/ /files/config

CMD ["/bin/sh"]
````

2. Use it as source image in BC to get the files
````
        ....
        "source": {
          "type": "Git",
          "git": {
            "uri": "https://github.com/bparees/openshift-jee-sample.git",
            "ref": "master"
          },
          "images": [
            {
              "from": {
                "kind": "ImageStreamTag",
                "name": "addons:latest"
              },
              "paths": [
                {
                  "destinationDir": "addon/",
                  "sourcePath": "/files/"
                }
              ]
            }
          ]
        },
       ...
````
##### Current Result
I have in the image:
````
/opt/app-root/src/addon/files/apply
/opt/app-root/src/addon/files/config/...
````

##### Expected Result
I would expect to have:
````
/opt/app-root/src/addon/apply
/opt/app-root/src/addon/config/...
````

##### Additional Information
I've tried with "sourcePath": "/files/*"
Only option I see is to include 2 paths, but this is unmanageable as the number of files can be big. I want to be able to copy everything in sourcePath into destinationPath.

### PR DESCRIPTION
**Pull request by bparees**
fixes https://github.com/openshift/origin/issues/8383

### DIFF
--- a/dev_guide/builds.adoc
+++ b/dev_guide/builds.adoc
@@ -907,7 +907,9 @@ The source path can be any absolute path within the image specified.  The destin
 must be a relative directory path.  At build time, the image will be loaded
 and the indicated files/directories will be copied into the context directory
 of the build process.  This is the same directory into which the source repository
-content (if any) is cloned.
+content (if any) is cloned.  If the source path ends in `*/.*` then the content
+of the directory will be copied, but the directory itself will not be created
+at the destination.
 
 Image inputs are specified in the `*source*` definition of the `*BuildConfig*`:
 

### CODE REVIEW

**Comment by bparees** [comment]

File: dev_guide/builds.adoc [hunk 1]
```diff
@@ -907,7 +907,9 @@ The source path can be any absolute path within the image specified.  The destin
 must be a relative directory path.  At build time, the image will be loaded
 and the indicated files/directories will be copied into the context directory
 of the build process.  This is the same directory into which the source repository
-content (if any) is cloned.
+content (if any) is cloned.  If the source path ends in `*/.*` then the content
```
tnguyen-rh: @bparees 
I don't understand if that is a literal string (asterisk, slash, dot, asterisk), or a glob pattern (any or no character, slash, dot, any or no character).

bparees: It's "/." the rest is formatting.

tnguyen-rh:
[rev_history]
|link:../dev_guide/builds.html[Builds]
|Clarified how to avoid copying the base directory when including extra files in the link:../dev_guide/builds.html#image-source[image source].

adellape:
@tnguyen-rh This is actually 3.2 only; adjusted labels/milestone.

tnguyen-rh:
@adellape Thanks.  I saw "fixes" in the initial comment and jumped to the wrong conclusion.

PR #1881, opened 2016-04-08
explain how contextdir works

bparees:


bparees:
@csrwng @jorgemoralespou ptal.

dev_guide/builds.adoc
```diff
@@ -519,9 +519,21 @@ customStrategy:
...
+== Build Inputs
+
+There are several ways to provide content for the build to operate on:  
+
+* Git repositories
+* Inline Dockerfile definitions
+* Binary inputs
+* Content extracted from existing images
+
+These inputs can also be combined in a single build.  When the build is run, a working directory is constructed and all input content is placed in the working directory (e.g. the input git repository is cloned into the working directory, files specified from input images are copied into the working directory using the target path).  Next the build process will `*cd*` into the `*contextDir*` if one is defined.  Finally the content from the current directory will be provided to the build process for reference by the `*Dockerfile*`, `*assemble*` script, or custom builder logic.
```
csrwng: > These inputs can also be combined in a single build.

It might be useful to explain how they can be combined and in what order. First you can have either git or binary, then on top of that image source and finally Dockerfile. 

> the input git repository is cloned into the working directory, files specified from input images are copied into the working directory using the target path

It's probably also worth noting that while binary, git and image source are placed at the root of the working directory, Dockerfile's are actually placed wherever your contextDir points.

We could also add a warning that states that if the destination directory of your image source falls outside of the context directory, then no part of your image source will be used in the build.

bparees: thanks, good points, i've updated it.

jorgemoralespou:
@bparees I think that an example would illustrate this better.

bparees:
@jorgemoralespou updated, ptal again.

jorgemoralespou:
@bparees excellent

bparees:

bparees:
@openshift/team-documentation bump

dev_guide/builds.adoc
```diff
@@ -519,9 +519,52 @@ customStrategy:
...
+== Build Inputs
+
+There are several ways to provide content for the build to operate on:  
+
+* Git repositories
+* Inline Dockerfile definitions
+* Binary inputs
+* Content extracted from existing images
+
+These inputs can also be combined in a single build.  If multiple inputs are specified and include overlapping filenames, the inline Dockerfile takes precedence, thereby overwriting any file named Dockerfile provided by another input.  Next in precedence is content from existing images, and finally content from the git repository or binary input.  Binary input and Git repository are mutually exclusive inputs.
```
bfallonf: @bparees I feel like a lot of into in the above para could be taken away, and listing the bullet points in order or precedence. So:

"There are several ways to provide content for builds to operate on. In existing order of precedence:

* Inline Dockerfile definitions
* Content extracted from existing images
* Git repositories
* Binary inputs

These can be combined into a single build. As the Dockerfile takes precedence, it can overwrite any other file named Dockerfile provided by another input. Binary input and Git repository are mutually exclusive inputs."

Or something like this.

dev_guide/builds.adoc
```diff
@@ -519,9 +519,52 @@ customStrategy:
...
+  contextDir: "app/dir" <3>
+  dockerfile: "FROM centos:7\nRUN yum install -y httpd" <4>
+
+----
+
+<1> The repository to be cloned into the working directory for the build
+<2> `*/usr/lib/somefile.jar*` from `*myinputimage*` will be stored in `*<workingdir>/app/dir/injected/dir*`
+<3> The working dir for the build will become `*<original_workingdir>/app/dir*`
+<4> A `*Dockerfile*` with this content will be created in `*<original_workingdir>/app/dir*`, overwriting any existing file with that name
+====
```
bfallonf: @bparees This might be a little bit personal preference-ey, but I think the above example can be combined with the para above a little more. The para is a little hard to follow until the example gives it a little more context. The specific example filenames etc could be put into the para, which might make it a little easier to follow. This could mean the little para above about "Here is an example" could disappear as well. This would leave the footnotes of the example to be a little less specific. Let me know what you think.

bfallonf:
@bparees This is largely all good. I had a few style suggestions. As an alternative to fixing them up, you could let me know what you think and apply a follow_up label and I could fix it up post-merge.

dev_guide/builds.adoc
bparees: great suggestion, thanks.

dev_guide/builds.adoc
bparees: Sorry, i'm not entirely sure what you're suggesting..I think you're saying i should describe what the example definition does (including specific discussion of the paths) in the paragraph prior to the example.  Personally I like the example as is because it flows in order describing how each element from the definition is applied during the build.  (the footnotes are in a specific order which matches the order they are used during the build process)

so it is a bit of an abuse of the footnotes, but i feel like it makes a direct connection between the definition elements and their usage/behavior, whereas a setup paragraph and a generic example/footnotes wouldn't do that.

bparees:
@bfallonf updated for your one suggestion, we can continue to discuss the other or you can merge it and we can iterate on that second part in your followup.

dev_guide/builds.adoc
```diff
@@ -519,9 +519,52 @@ customStrategy:
...
+There are several ways to provide content for builds to operate on. In order of precedence:
+
+* Inline Dockerfile definitions
+* Content extracted from existing images
+* Git repositories
+* Binary inputs
+
+These can be combined into a single build. As the inline Dockerfile takes precedence, it can overwrite any other file named Dockerfile provided by another input. Binary input and Git repository are mutually exclusive inputs.
+
+When the build is run, a working directory is constructed and all input content is placed in the working directory (e.g. the input git repository is cloned into the working directory, files specified from input images are copied into the working directory using the target path).  Next the build process will `*cd*` into the `*contextDir*` if one is defined.  Then the inline `*Dockerfile*` (if any) is written to the current directory.  Last, the content from the current directory will be provided to the build process for reference by the `*Dockerfile*`, `*assemble*` script, or custom builder logic.  This means any input content that resides outside the `*contextDir*` will be ignored by the build.
```
adellape: I think this could work as an ordered list instead of having to dig through a big paragraph?

adellape:
@bparees @bfallonf I actually liked the callouts providing the explanations directly related to the object. Maybe my suggestion inline about breaking up that big paragraph into an ordered list would help ease from the high-level process -> example.

I think this is fine to merge as-is for now, with follow-up to come. I also hope to get this Builds topic broken up (i.e., a "Builds" subdirectory with multiple topics)  in time for 3.2, so I'll be back in here shortly.

bparees:
>  I also hope to get this Builds topic broken up (i.e., a "Builds" subdirectory with multiple topics) in time for 3.2, so I'll be back in here shortly.

awww, i take pride in the devex team owning the largest file in the docs :)

vikram-redhat:
Since the builds topic is now broken up with https://github.com/openshift/openshift-docs/milestone/100?closed=1, I am going to remove the to_followup label. @adellape - please let me know if otherwise.

PR #1887, opened 2016-04-11
Bug 1271414, documented maximum name lengths for objects

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1271414

### DIFF
--- a/architecture/core_concepts/deployments.adoc
+++ b/architecture/core_concepts/deployments.adoc
@@ -54,7 +54,7 @@ spec:
   template:    <3>
     metadata:
       labels:  <4>
-        name: frontend
+        name: frontend <5>
     spec:
       containers:
       - image: openshift/hello-openshift
@@ -69,6 +69,8 @@ spec:
 2. The label selector of the pod to run.
 3. A template for the pod the controller creates.
 4. Labels on the pod should include those from the label selector.
+5. The maximum name length after expanding any parameters is 63 characters.
+
 
 [[jobs]]
 
--- a/architecture/core_concepts/routes.adoc
+++ b/architecture/core_concepts/routes.adoc
@@ -16,10 +16,10 @@ An OpenShift route is a way to expose a
 link:pods_and_services.html#services[service] by giving it an
 externally-reachable hostname like `www.example.com`.
 
-A defined route and the endpoints identified by its service can be consumed by
-a router to provide named connectivity that allows external clients to reach
-your applications. Each route consists of a route name, service selector,
-and (optionally) security configuration.
+A defined route and the endpoints identified by its service can be consumed by a
+router to provide named connectivity that allows external clients to reach your
+applications. Each route consists of a route name (limited to 63 characters),
+service selector, and (optionally) security configuration.
 
 == Routers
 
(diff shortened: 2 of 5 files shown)


ahardin-rh:
@adellape PTAL. Thanks!

architecture/core_concepts/routes.adoc
```diff
@@ -125,9 +125,10 @@ image::router_model.png[HAProxy Router Data Flow]
 Implementing sticky sessions is up to the underlying router configuration. The
 default HAProxy template implements sticky sessions using the `balance source`
 directive which balances based on the source IP. In addition, the template
-router plug-in provides the service name and namespace to the underlying
-implementation. This can be used for more advanced configuration such as
-implementing stick-tables that synchronize between a set of peers.
+router plug-in provides the service name (limited to 63 characters) and
```
adellape: I think this mention can be dropped since you already mention the service name limit in the services topic.

adellape:
@ahardin-rh I think this approach works.

ahardin-rh:
@mfojtik @Miciah PTAL. Thanks!

architecture/core_concepts/routes.adoc
```diff
@@ -448,28 +448,29 @@ will be used for TLS termination.
...
       -----END CERTIFICATE-----
-    caCertificate: |-            <4>
+    caCertificate: |-            <5>
       -----BEGIN CERTIFICATE-----
       [...]
       -----END CERTIFICATE-----
 ----
-<1> The `*termination*` field is `edge` for edge termination.
+<1> The name of the object, which is limited to 63 characters.
+<2> The `*termination*` field is `edge` for edge termination.
```
Miciah: Key and certificate were already swapped before your change, and you forgot to shift the numbers after 2 here, so now there are two 2's here, and the numbers down here don't match up with the numbers above.

architecture/core_concepts/deployments.adoc
```diff
@@ -69,6 +69,8 @@ spec:
 2. The label selector of the pod to run.
 3. A template for the pod the controller creates.
 4. Labels on the pod should include those from the label selector.
+5. The maximum name length is 63 characters.
```
Miciah: It might be worth adding the phrase "after expanding any parameters" after "maximum name length".  This specific example doesn't use parameters, but it does use a template, and the fact that these limits apply *after* parameter expansion should be documented in the context of templates.  A better location to call this fact out might be [the documentation on templates in the Developer Guide](https://github.com/openshift/openshift-docs/blob/master/dev_guide/templates.adoc#templates-parameters)—except the examples there don't use parameters in the name field.  Hm...

architecture/core_concepts/routes.adoc
ahardin-rh: Oops! Fixed now. Thanks!

ahardin-rh:
[rev_history]
|link:../architecture/core_concepts/pods_and_services.html[Pods and Services]
|Added information about maximum name lengths for objects.
%
|link:../architecture/core_concepts/projects_and_users.html[Projects and Users]
|Added information about maximum name lengths for objects.
%
|link:../architecture/core_concepts/builds_and_image_streams.html[Builds and Image Streams]
|Added information about maximum name lengths for objects.
%
|link:..architecture/core_concepts/deployments.html[Deployments]
|Added information about maximum name lengths for objects.
%
|link:../architecture/core_concepts/routes.html[Routes]
|Added information about maximum name lengths for objects.
%

PR #1596, opened 2016-02-16
Add KeystonePasswordIdentityProvider docs

liggitt:
Fixes #1073
Added in https://github.com/openshift/origin/pull/4951

PR #1902, opened 2016-04-14
Add url configuration directive for keystone auth

chmouel:
Url is needed for configuring keystone authentication, let's mention it.

### DIFF
--- a/install_config/configuring_authentication.adoc
+++ b/install_config/configuring_authentication.adoc
@@ -290,6 +290,7 @@ oauthConfig:
     provider:
       apiVersion: v1
       kind: KeystonePasswordIdentityProvider
+      url: https://keystone_url:5000/v2.0
       domainName: default <5>
       ca: ca.pem <6> 
       certFile: keystone.pem <7> 


install_config/configuring_authentication.adoc
```diff
@@ -290,6 +290,7 @@ oauthConfig:
     provider:
       apiVersion: v1
       kind: KeystonePasswordIdentityProvider
+      url: https://keystone_url:5000/v2.0
```
liggitt: make this follow the format of other example urls (see other places in the doc). Also, is the v2.0 right? I thought it had to be a v3 server

chmouel: I tested with v2.0 since that was the default on devstack, from what I can tell Gophercloud would match the url suffix to detect if it's v3 or v2 : 

https://github.com/openshift/origin/blob/0c674d492cf2cdc026a8549e12b49972299c116b/Godeps/_workspace/src/github.com/rackspace/gophercloud/openstack/client.go#L66-L69

I can surely specify v3 by default and note that v2 is supported as well,

liggitt: v2 is not supported, only v3

ahardin-rh:
@chmouel @liggitt resurrecting discussion on this. :ghost: Are the changes discussed here still in motion?

chmouel:
I was under the impression this was worked on another pr would be happy to
resurrect this if needed

ahardin-rh:
@chmouel @liggitt My apologies. Looks like this was addressed in https://github.com/openshift/openshift-docs/pull/1596. I will close this PR in favor of that one. Thanks!

PR #1906, opened 2016-04-14
Bug 1326595, added Managing Users section

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1326595

### DIFF
--- a/admin_guide/manage_users.adoc
+++ b/admin_guide/manage_users.adoc
@@ -1,0 +1,135 @@
+= Managing Users
+{product-author}
+{product-version}
+:data-uri:
+:icons:
+:experimental:
+:toc: macro
+:toc-title:
+
+toc::[]
+
+== Adding Users
+
+Once new users log in to OpenShift, an account is created for that user and the
+cluster administrator can
+link:../admin_guide/manage_authorization_policy.html[manage the access level of
+each user].
+
+== Deleting Users
+
+=== Improper Deletion
+
+OpenShift user configuration is stored in several locations within OpenShift.
+Regardless of the identity provider, OpenShift internally stores details like
+role-based access control (RBAC) information and group membership. To completely
+remove user information, this data must be removed in addition to the user
+account.
+
+If an account is improperly deleted, the following type of error can be seen,
+even though the identification provider already shows the user *myuser* as
+existing:
+
+====
+----
+{
+    "MESSAGE": "E0412 06:07:48.142862     465 login.go:141] Unable to authenticate password: Error creating or updating mapping for: &api.DefaultUserIdentityInfo{ProviderName:\"htpasswd_auth\", ProviderUserName:\"myuser\", Extra:map[string]string{}} due to users \"myuser\" not found",
+    "PRIORITY": "6",
+    "SYSLOG_FACILITY": "3",
+    "SYSLOG_IDENTIFIER": "atomic-openshift-master",
+    "_BOOT_ID": "18e98783c33546b4a4346f6750ef84a5",
+    "_CAP_EFFECTIVE": "1fffffffff",
+    "_CMDLINE": "/usr/bin/openshift start master --config=/etc/origin/master/master-config.yaml --loglevel=2",
+    "_COMM": "openshift",
+    "_EXE": "/usr/bin/openshift",
+    "_GID": "0",
+    "_HOSTNAME": "master.example.com",
+    "_MACHINE_ID": "f9370ed252a14f73b014c1301a9b6d1b",
+    "_PID": "465",
+    "_SELINUX_CONTEXT": "system_u:system_r:init_t:s0",
+    "_SYSTEMD_CGROUP": "/system.slice/atomic-openshift-master.service",
+    "_SYSTEMD_SLICE": "system.slice",
+    "_SYSTEMD_UNIT": "atomic-openshift-master.service",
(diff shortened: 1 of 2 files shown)


ahardin-rh:
@adellape @tnguyen-rh @bfallonf PTAL thanks!

admin_guide/manage_users.adoc
```diff
@@ -0,0 +1,135 @@
...
+NAME                  IDP NAME        IDP USER NAME   USER NAME   USER UID
+htpasswd_auth:ashley  htpasswd_auth   ashley          ashley      75e4b80c-dbf1-11e5-8dc6-0e81e52cc949
+----
+====
+
+Note the UID between the two matches. If you attempt to change the
+authentication provider after starting to use OpenShift, the user names that
+overlap will not work because of the entries in the identity list, which will
+still point to the old authentication method.
+
```
bfallonf: @ahardin-rh I got a little lost and had to reread the first sentence. Can I suggest:

"Note the matching UID between the two results."

I think the problem was that 'matches' was too far from what it's modifying (UID).

Also, from what I understand, are you talking about two different improper ways to delete a user? It seems it goes from deleting the data in the wrong order, but then talking about changing identity providers. I don't think i can see a link between the two... Maybe a little elaboration is needed.

bfallonf:
@ahardin-rh Just the one comment from me (which I think is actually two comments...). Otherwise, LGTM :+1:

ahardin-rh:
[rev_history]
|link:../admin_guide/manage_users.html[Managing Users]
|Added the link:../admin_guide/manage_users.html[Managing Users] section.
%

admin_guide/manage_users.adoc
```diff
@@ -0,0 +1,135 @@
...
+{product-version}
+:data-uri:
+:icons:
+:experimental:
+:toc: macro
+:toc-title:
+
+toc::[]
+
+== Adding Users
```
adellape: Remember to add manual, unique IDs for headings (for example, check to make sure there's not some other heading in the repo named "adding-users").

admin_guide/manage_users.adoc
```diff
@@ -0,0 +1,135 @@
...
+
+Once new users log in to OpenShift, an account is created for that user and the
+cluster administrator can
+link:../admin_guide/manage_authorization_policy.html[manage the access level of
+each user].
+
+== Deleting Users
```
adellape: It's nicer if we can avoid headings that just lead into a subheading. For example, if there's some intro or explanation we can add here before going into Improper Deletion.

admin_guide/manage_users.adoc
```diff
@@ -0,0 +1,135 @@
...
+
+=== Improper Deletion
+
+OpenShift user configuration is stored in several locations within OpenShift.
+Regardless of the identity provider, OpenShift internally stores details like
+role-based access control (RBAC) information and group membership. To completely
+remove user information, this data must be removed in addition to the user
+account.
+
+If an account is improperly deleted, the following type of error can be seen,
```
adellape: Seen by who and when? I assume it's seen by the user/developer when attempting to authenticate or do anything?

admin_guide/manage_users.adoc
```diff
@@ -0,0 +1,135 @@
...
+even though the identification provider already shows the user *myuser* as
+existing:
+
+====
+----
+{
+    "MESSAGE": "E0412 06:07:48.142862     465 login.go:141] Unable to authenticate password: Error creating or updating mapping for: &api.DefaultUserIdentityInfo{ProviderName:\"htpasswd_auth\", ProviderUserName:\"myuser\", Extra:map[string]string{}} due to users \"myuser\" not found",
+    "PRIORITY": "6",
```
adellape: We can probably trim this error message to just the "MESSAGE" here? I.e., just line 36.

admin_guide/manage_users.adoc
```diff
@@ -0,0 +1,135 @@
...
+    "_UID": "0",
+}
+----
+====
+
+This error is generated because OpenShift sees the user data in the internal
+database that does not correspond to the external identity.
+
+In OpenShift, two object types contain user data outside the identification
+provider: type user and type identity.
```
adellape: We don't really refer to objects as "type <name>" anywhere else, so this feels weird here. What about just:

```
In OpenShift, two object types contain user data outside the identification provider: `*user*` and `*identity*`.
```

admin_guide/manage_users.adoc
```diff
@@ -0,0 +1,135 @@
...
+
+Type user:
```
adellape: Maybe make these say what they're showing, e.g., "To get the current list of users:" (Same for "type identity" below)

admin_guide/manage_users.adoc
```diff
@@ -0,0 +1,135 @@
...
+To delete a user, all three sources of the user must be removed:
+
+. Identification manager
+. User record
+. Identity record
+
+Once you complete these steps, a new account will be created in OpenShift
+when the user logs in again.
+
+==== Deleting the Identification Manager Entry
```
adellape: Can we make these a procedure w/ numbered steps instead of individual subheadings per step?

ahardin-rh:
Follow-up PR: https://github.com/openshift/openshift-docs/pull/1921

PR #1925, opened 2016-04-20
DOCKER REPO should be new REGISTRY ip.

Jooho:
The final Docker repo should be new $REGISTRY not registry.access.redhat.com. I think it makes people confused so I correct it.


adellape:
:+1: Thanks.

No revhistory needed.

PR #1597, opened 2016-02-16
add from override (buildconfig vs dockerfile)

jfcoz:
Add documentation for https://github.com/openshift/origin/issues/6842

PR #1929, opened 2016-04-21
Followup to PR 1597: Edited title caps

bfallonf:
Follow up to #1597 

I don't think this requires too much else. Anybody want to merge?

@tpoitras @tnguyen-rh @ahardin-rh @adellape


adellape:
:+1:

PR #1936, opened 2016-04-22
Fix metric re-encrypted route port for enterprise containers.

mwringe:
The port value for the re-encrypting port needs to be different between the enterprise and origin metric containers.

Without this fix people viewing how to setup the re-encrypting route will be entering the wrong port and they will not be able to properly get the route working.

### DIFF
--- a/install_config/cluster_metrics.adoc
+++ b/install_config/cluster_metrics.adoc
@@ -446,7 +446,12 @@ metadata:
 spec:
   host: hawkular-metrics.example.com <1>
   port:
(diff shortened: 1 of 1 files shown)


adellape:
[rev_history]
|link:../install_config/cluster_metrics.html[Enabling Cluster Metrics]
|Updated the port value for the re-encrypting port to 8444 for OpenShift Enterprise, which is different from the value for OpenShift Origin, which uses 8443.
%

PR #1939, opened 2016-04-25
Update templates.adoc

alberttwong:
fixing the oc export all command.


ahardin-rh:
Continuing work in https://github.com/openshift/openshift-docs/pull/3430

Issue #1949, opened 2016-04-26
Add Thresholds + Capacity Planning To System Requirements.  


timothysc:
We've finished the online testing and need to draft language that outlines our current thresholds.  

**Supported thresholds are : **
110 pods/node 
300 nodes 

I'm not certain how we want to craft the language and discuss over-riding ```max-pods``` b/c I'm not certain what our support will be. 

**Capacity & Deployment Planning: **
Depending on the total number of resource objects that are planned for a cluster, both the masters deployment as well as sizing of the masters needs to be considered prior to deployment.  For small proof of concept clusters this is not an issue.   

Considerations in planning: 
- Will the cluster be HA
-- Recommended for large deployments or if master fault tolerance is a requirement for your deployment. 
- What are the total number of pods running in the system? 
- What are the total number of pods/machine?  

A general estimate is about 1 cpu core, and 2.5 GB of memory atop of the defaults per 1k pods.    So for 20k pods that would be (20+2) cpu cores and and (50+8) GB of ram. 

I figure this is a good start, but we'll need to rally on language. 

/cc @jeremyeder

bfallonf:
Info in #1961 . Closing this issue.

PR #1961, opened 2016-04-28
Bug 1306890 : Added maximum pod and node info

bfallonf:
@timothysc @jeremyeder As per #1949 and BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1306890

The max-pods stuff seems to be covered here: https://docs.openshift.com/enterprise/3.1/admin_guide/manage_nodes.html#configuring-node-resources

I do have a question about the CPU requirements. Why is it 20+2 and 50+8? Why not 22 and 58? Is there a difference between the two numbers used there? Other than that, if there's anything else, please let me know. 

Thanks for the inital info in the GH issue!

### DIFF
--- a/architecture/core_concepts/pods_and_services.adoc
+++ b/architecture/core_concepts/pods_and_services.adoc
@@ -38,6 +38,11 @@ maintain state when recreated. Therefore pods should usually be managed by
 higher-level link:deployments.html#replication-controllers[controllers],
 rather than directly by users.
 
+[IMPORTANT]
+====
+The recommended maximum number of pods within an {product-title} cluster is 110.
+====
+
 Below is an example definition of a pod that provides a long-running
 service, which is actually a part of the OpenShift infrastructure: the
 private Docker registry. It demonstrates many features of pods, most of
--- a/architecture/infrastructure_components/kubernetes_infrastructure.adoc
+++ b/architecture/infrastructure_components/kubernetes_infrastructure.adoc
@@ -241,6 +241,11 @@ link:../../install_config/master_node_configuration.html[dedicated node
 configuration files].
 endif::[]
 
+[IMPORTANT]
+====
+The recommended maximum number of nodes is 300.
+====
+
 [[kubelet]]
 
 === Kubelet
--- a/install_config/install/advanced_install.adoc
+++ b/install_config/install/advanced_install.adoc
@@ -1105,6 +1105,11 @@ your inventory file. You must then specify the file location with `-i` when
 calling `ansible-playbook` later.
 endif::[]
 
+[IMPORTANT]
+====
+The recommended maximum number of nodes is 300.
+====
+
 To add nodes to an existing cluster:
 
 . Ensure you have the latest playbooks by updating the *atomic-openshift-utils*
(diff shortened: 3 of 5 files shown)


architecture/core_concepts/pods_and_services.adoc
```diff
@@ -38,6 +38,11 @@ maintain state when recreated. Therefore pods should usually be managed by
 higher-level link:deployments.html#replication-controllers[controllers],
 rather than directly by users.
 
+[IMPORTANT]
+====
+The maximum supported number of pods within an {product-title} cluster is 110.
```
timothysc: I don't know if we want to have a section on allowing over-rides here?  

Also might want to craft the language to be softer, as these limits can be stretched, but we want to give customers guidance, while allowing power-users to tinker with the engine. 

/cc @jeremyeder ?

architecture/infrastructure_components/kubernetes_infrastructure.adoc
```diff
@@ -241,6 +241,11 @@ link:../../install_config/master_node_configuration.html[dedicated node
 configuration files].
 endif::[]
 
+[IMPORTANT]
+====
+The maximum supported number of nodes is 300.
```
timothysc: Same comment.

timothysc:
The +2 and +8 are additions from the base requirements listed for the master.

architecture/core_concepts/pods_and_services.adoc
bfallonf: @timothysc The max-pods setting is talked about here: https://docs.openshift.com/enterprise/3.1/admin_guide/manage_nodes.html#configuring-node-resources

I think adding a section about over-riding this might be out of scope for this BZ, but we can make another and work on it. 

Also, I considered changing the language to say "recommended minimums" instead of "supported minimums" but that might be against the idea of this BZ. I don't think we can have both... @timothysc @jeremyeder Do you have thoughts on this? Which to prefer?

bfallonf:
@timothysc Thanks for the review! I changed the +2 and +8 stuff to mention the minimum prerequisite requirements and made one comment above. Let me know what you think.

jeremyeder:
Haven't forgotten about this, just haven't gotten to it yet.

architecture/core_concepts/pods_and_services.adoc
timothysc: I prefer "recommended"

timothysc:
Only comment is to change to language to "recommended"

bfallonf:
Ok. I made that change. @timothysc @jeremyeder Let me know if there's any more thoughts. This is for 3.2, so it'd be good to get it done ASAP.

bfallonf:
@timothysc @jeremyeder This is going to have to be given the thumbs up by the Wednesday otherwise it's going to miss the 3.2 release. Is there anything particularly preventing this from merging? Do we need docs on over-riding before this can merge?

timothysc:
Please merge, and if needed we'll follow up with another issue/PR.

bfallonf:
[rev_history]
|Added prerequisite information for CPU and GB size requirements to link:../install_config/install/prerequisites.html#system-requirements[System Requirements], and Important boxes recommending the node and pod limits.
%

bfallonf:
@timothysc Sure thing. Thanks for that. Let me know what followup information you want with this. We do a release every week, so we can do it anytime.

PR #1985, opened 2016-05-02
Add port 8053 to the list of required ports for new installs

sdodson:


sdodson:
Probably needs a good bit more discussion about why we're using dnsmasq and how we're using it. I'm going to look for where best to document that.

sdodson:
@adellape looks like 3.2 release notes already address dnsmasq bits pretty well so I think this is the only additional change necessary.

adellape:
Closing in favor of https://github.com/openshift/openshift-docs/pull/2075

PR #2020, opened 2016-05-06
explain impersonation

deads2k:
We added a way for users to impersonate other users.  This explains how to use it.

I'm not certain exactly where to put it.

### DIFF
--- a/architecture/additional_concepts/authentication.adoc
+++ b/architecture/additional_concepts/authentication.adoc
@@ -88,6 +88,23 @@ the `system:anonymous` virtual user and the `system:unauthenticated` virtual
 group to the request. This allows the authorization layer to determine which
 requests, if any, an anonymous user is allowed to make.
 
+=== Impersonation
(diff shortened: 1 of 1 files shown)


ahardin-rh:
@deads2k Thanks! Jumping in here since @adellape is on PTO this week. I think your current positioning looks fine. We can always shuffle it later if we want to. Ready to merge?

deads2k:
@liggitt any comments on technical details?

deads2k:
> Ready to merge?

@ahardin-rh  Got an out of band ok from @liggitt .

ahardin-rh:
@deads2k @liggitt Thanks!

adellape:
@deads2k Would this https://github.com/openshift/origin/pull/8824 be the beginning of this code work? The associated Trello card there mentions 3.3.

This docs PR got mislabeled I think, so we'll revert from the 3.1/3.2 docs and put it back in queue for 3.3.

PR #2023, opened 2016-05-06
Remove doc about plugin order override

csrwng:
Given that all admission control plugins are now part of the default admission control chain, documentation on how to override this chain is no longer needed and it's preferable that end users do not make changes to this. Changing which plugins are in the chain and their order can lead to all kinds of support issues.

### DIFF
--- a/admin_guide/managing_projects.adoc
+++ b/admin_guide/managing_projects.adoc
@@ -233,18 +233,6 @@ The number of self-provisioned projects requested by a given user can be limited
 the `*ProjectRequestLimit*`
 link:../install_config/configuring_admission_control.html[admission control plug-in].
 
-The plug-in is not enabled by default. Therefore, it must be specified along
-with default {product-title} admission control plug-ins in the configuration:
-
-[source,yaml]
-----
-admissionConfig:
-  pluginOrderOverride:
-  - OriginNamespaceLifecycle
-  - BuildByStrategy
-  - ProjectRequestLimit
-----
-
 In order to specify limits for users, a configuration must be specified for the
 plug-in. The plug-in configuration takes a list of user label selectors and the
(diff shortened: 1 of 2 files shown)


csrwng:
@deads2k @adellape ptal

adellape:
@csrwng Thanks. So it's no longer worth mentioning what each one is for at all, I guess?

adellape:
> all admission control plugins are now part of the default admission control chain

@csrwng ^ And is this true as of OSE 3.2? Or later?

csrwng:
@adellape that doc is more useful to understand the internals of admission control, but if we don't want end users changing anything in the list and order, I don't think it's very useful. It's more useful as developer doc.

> @csrwng ^ And is this true as of OSE 3.2? Or later?

yes that is true as of 3.2

adellape:
@csrwng Now that the "Configuring Admission Control Plug-ins" topic has been reduced to a definition of what admission controllers are, and a section generically about how plug-ins can be configured (depending on a given plug-in's usage), it seems to me now like instead we could remove this topic and:

- Move the "Overview" content here to an Arch topic
- Just lose the "Providing Configuration..." section altogether, and let the dedicated areas of the docs discuss where configuration for a plug-in is necessary, as they already are (e.g., "Managing Projects" for the `ProjectRequestLimit` and "Configuring Global Build Defaults and Overrides" for `BuildDefaults` and `BuildOverrides`, so far).

WDYT?

csrwng:
SGTM, do you want me to make that change with this PR? or would you do it in a follow up ?

adellape:
@csrwng Cool, I'll pull this commit into another PR and continue.

adellape:
Closing in favor of https://github.com/openshift/openshift-docs/pull/2071.

PR #2043, opened 2016-05-11
subsitute s2i for sti in doc

bparees:


bparees:

adellape:
:+1:

PR #2054, opened 2016-05-11
Replace sti command by s2i

php-coder:
PTAL @mfojtik @openshift/team-documentation

### DIFF
--- a/creating_images/s2i_testing.adoc
+++ b/creating_images/s2i_testing.adoc
@@ -50,11 +50,11 @@ but it should exercise the crucial steps of `assemble` and `run` scripts.
 
 == Generating Scripts and Tools
 The S2I tooling comes with powerful generation tools to speed up the process of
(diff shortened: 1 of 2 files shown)


adellape:
@php-coder LGTM. 

BTW, is the following still accurate (from elsewhere in this topic), in that it's the only way to get the actual `s2i` binary? We don't ship anything that packages it, right (I haven't found anything yet)?

> The test/run script performs the S2I build, so you must have the S2I binary available in your $PATH. If required, follow the installation instructions in the S2I README[1].

[1] https://github.com/openshift/source-to-image/blob/master/README.md#installation

php-coder:
>We don't ship anything that packages it, right (I haven't found anything yet)?

I don't know for sure, because I always build from the sources. Maybe @bparees or @mfojtik know more about this.

bparees:
right, we don't ship the s2i binary with openshift, they'd have to go get an s2i release:
https://github.com/openshift/source-to-image/releases

PR #1724, opened 2016-03-10
switch terminology (instant app vs. quickstart)

gabemontero:
@adellape PTAL - I'm pretty sure this is not the only place where InstantApp vs. Quickstart terminology conversions need to occur, but since I know I created this content for the online dev preview (based on input from @sspeiche) just prior to @bparees 's work with the quickstart tutorial, I knew this at least had to be changed :-)

And of course comments from one and all for further massaging of this content always welcome.

PR #2058, opened 2016-05-11
Follow-up to Switch terminology (instant app vs. quickstart) PR#1724

ahardin-rh:
Follow-up to https://github.com/openshift/openshift-docs/pull/1724

### DIFF
--- a/getting_started/developers/developers_cli.adoc
+++ b/getting_started/developers/developers_cli.adoc
@@ -17,8 +17,8 @@ toc::[]
 == Overview
 
 This getting started experience walks you through the simplest way to get a
-sample project up and running on OpenShift. There are a few different ways to
-launch images within a project, but this topic focuses on the quickest and
+sample project up and running on {product-title}. There are a few different ways
+to launch images within a project, but this topic focuses on the quickest and
 easiest method.
 
 If this is the first part of the documentation you have read, and you are
@@ -64,7 +64,7 @@ link:../../using_images/s2i_images/index.html[builder images].
 
 |===
 
-Other OpenShift Docker Images
+Other {product-title} Docker Images
 
 * https://github.com/openshift/mysql[MySQL]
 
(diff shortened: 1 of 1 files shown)


ahardin-rh:
@adellape @tnguyen-rh @bfallonf @tpoitras PTAL. A little uncertain about this follow-up. Looks like "Quickstart" terminology is favored, but both terms are used in Dev Preview. Not sure what the difference is exactly.

tnguyen-rh:
@ahardin-rh
IIRC, both are valid and describe slightly different concepts.  One is intended to mean "no tinkering required" while the other is intended to mean "tinkering required" (which is which, i can't recall, unfortunately).

Since i'm here, i did a quick peer review.  LGTM.

adellape:
@ahardin-rh @tnguyen-rh Here's how we're defining quickstarts:

https://docs.openshift.com/enterprise/latest/dev_guide/app_tutorials/quickstarts.html

Whereas I believe an "instant app" is meant to suggest that you'll end up with something actually usable out the gate, currently our only default template example of this being Jenkins. But something like WordPress could be an instant app, if we had a template for that. cc @bparees for corrections.

bparees:
@adellape +1

quickstart is basically a starting point for app development.  it gives you something that's running, but not something useful.  the assumption is you'll take the code it gives you and start from there to develop your real app.

ahardin-rh:
@tnguyen-rh @adellape @bparees  Thanks! Very helpful. :bow:

Do you think I should make a brief comment about that within this topic? Perhaps in the Overview?

tnguyen-rh:
@ahardin-rh I think that's a good idea.

ahardin-rh:
@adellape @tnguyen-rh Added a note box about Quickstarts vs Instant Apps. PTAL. Thanks!

tnguyen-rh:
@ahardin-rh The body of the NOTE is fine, but the syntax is incorrect.  Here is a command you can use to see other (correct) examples in the tree:

    git grep NOTE

(Invoke from root dir for best results.)

Otherwise, LGTM. :+1:

ahardin-rh:
@tnguyen-rh Oops! Good catch! fixed.

ahardin-rh:
[rev_history]
|link:../getting_started/developers/developers_cli.html[Developers -> Getting Started for Developers: CLI]
|Added Instant App and Quickstart information, including a Note box explaining the difference between them.
%

adellape:
We never actually published the developers_cli topic into anything, so I've disabled the revhistory note above.

PR #2089, opened 2016-05-13
Bug 1253128 : Mass conversion of json to yaml

bfallonf:
Converted all the json I could find to yaml as per https://bugzilla.redhat.com/show_bug.cgi?id=1253128

@tnguyen-rh You've proven yourself a json/yaml expert. Can I ask for a review? I think there's a few times where it could be set out a little better. I'll note that I did check this all on a conversion validator:

http://jsontoyaml.com/
http://yamltojson.com/

Also, there's three times where I couldn't work it out:
- dev_guide/templates
  - The example under "[Parameters](https://docs.openshift.com/enterprise/3.2/dev_guide/templates.html#writing-parameters)"
- dev_guide/deployments
  - Everything from "[Rolling Strategy](https://docs.openshift.com/enterprise/3.2/dev_guide/deployments.html#rolling-strategy)" onwards
- using_images/db_images/mongodb
  - The example under "[Creating Deployment Configurations](https://docs.openshift.com/enterprise/3.2/using_images/db_images/mongodb.html#creating-the-deploymentconfig)"

Are you up to the task? Thanks!

### DIFF
--- a/admin_guide/scheduler.adoc
+++ b/admin_guide/scheduler.adoc
@@ -251,22 +251,21 @@ applicable nodes as evenly as possible.
 The configuration below specifies the default scheduler configuration, if it
 were to be specified via the scheduler policy file.
 ----
-{
-	"kind" : "Policy",
-	"version" : "v1",
-	"predicates" : [
-		{"name" : "PodFitsPorts"},
-		{"name" : "PodFitsResources"},
-		{"name" : "NoDiskConflict"},
-		{"name" : "MatchNodeSelector"},
-		{"name" : "HostName"}
-	],
-	"priorities" : [
-		{"name" : "LeastRequestedPriority", "weight" : 1},
-		{"name" : "BalancedResourceAllocation", "weight" : 1},
-		{"name" : "ServiceSpreadingPriority", "weight" : 1}
-	]
-}
+kind: "Policy"
+version: "v1"
+predicates: 
+  - name: "PodFitsPorts"
+  - name: "PodFitsResources"
+  - name: "NoDiskConflict"
+  - name: "MatchNodeSelector"
+  - name: "HostName"
+priorities: 
+  - name: "LeastRequestedPriority"
+    weight: 1
+  - name: "BalancedResourceAllocation"
+    weight: 1
+  - name: "ServiceSpreadingPriority"
+    weight: 1
 ----
 
 [IMPORTANT]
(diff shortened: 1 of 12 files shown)


tnguyen-rh:
@bfallonf Is there any way to configure the translation service(s) to use two space characters for each indentation level, instead of a single tab character?

tnguyen-rh:
@bfallonf 
While i'm waiting for no-tabs, i figured i'd tackle the difficult cases (the rest) that you mentioned.  I've uploaded them as [bz1253128-pr2089-the-rest.yaml.txt](https://github.com/openshift/openshift-docs/files/266609/bz1253128-pr2089-the-rest.yaml.txt).  The `.txt` extension is for the sake of GitHub, which does accept `,yaml`-named files.  You'll have to add back the callouts, of course...

bfallonf:
@tnguyen-rh I updated the rest of the files. Good job figuring out what was wrong with them. You're such a champ!

As for the tabs. Does that make them not render right? Is that breaking? I went back to fix some up and it turns out that putting two spaces into atom puts in a tab instead. I realised this when I went back to fix some I had already fixed up... i could use another editor.

tnguyen-rh:
@bfallonf Check out the `expand` utility.  In light testing, `expand -t 2 INFILE > OUTFILE` seems to work well.

bfallonf:
@tnguyen-rh Ok. I couldn't get that to work for me, but what I did do was figure out the hard-coded tab spacing happen in atom and change it to 1 space instead of 2. Then I pushed again. This should mean where everywhere there was 1 tab there is now 2 spaces. I checked in the 'files changed' tab, and this seems to be fine. 

Does this solve the issue?

admin_guide/scheduler.adoc
```diff
@@ -279,77 +286,131 @@ most, if not all, of the default predicates and priority functions listed above.
...
-}
+kind: "Policy"
+version: "v1"
+predicates:
+...
+  - 
+    name: "RequireRegion"
+    argument: 
+      labelsPresence: 
+				labels:
```
tnguyen-rh: There are tab characters here.

tnguyen-rh:
@bfallonf Getting close, but...

- There are still some tab characters, unfortunately.  They need to be converted to two spaces, each (i believe).  I added a line-comment on one of the first instances detected, but it's not the only one.  Maybe @adellape  knows how to configure atom to show tab chars prominantly?

- The other general comment is that everywhere you see:

```yaml
...
foo:
  -
    a: b
    c: d
...
```

you can safely (w/o changing semantics) delete the line break after the hyphen and the following line's indentation, resulting in:

```yaml
...
foo:
  - a: b
    c: d
...
```

The resulting style is more readable (IMHO) and in line w/ the rest of the YAML in the docs.

bfallonf:
@tnguyen-rh OK.

I spent a while taking away all the tabs and putting in spaces instead of relying on atom, which I think I should have done initially... And also took away all the line breaks. Let me know if I missed any, or you have any other suggestions.

Thanks a bunch!

dev_guide/builds.adoc
```diff
@@ -1756,16 +1747,11 @@ The *_.docker/config.json_* file is found in your home directory by default and
...
-    "https://index.docker.io/v1/": { <1>
-      "auth": "YWRfbGzhcGU6R2labnRib21ifTE=", <2>
-      "email": "user@example.com" <3>
-    }
-  }
-}
+https://index.docker.io/v1/: <1>
+  auth: "YWRfbGzhcGU6R2labnRib21ifTE=" <2>
+  email: "user@example.com" <3>
 ----
```
tnguyen-rh: I think this is missing the `auths` key. Should be:

```yaml
auths:
  https://index.docker.io/v1/:
    auth: ...
    email: ...
```

tnguyen-rh:
@bfallonf One last glitch, otherwise LGTM.

bfallonf:
/rev history/
|Various files
|Changed JSON example output to YAML. 
%

dev_guide/builds.adoc
bfallonf: Done.

bfallonf:
Thanks @tnguyen-rh . Good eyes.

I don't think any further review is required. I'll merge this.

PR #2096, opened 2016-05-13
logging: use new-app instead of process|create

sosiouxme:
(For origin and ose-3.2)

@adellape Small changes that simplify some of the steps, conceptually and as less typing.

### DIFF
--- a/install_config/aggregate_logging.adoc
+++ b/install_config/aggregate_logging.adoc
@@ -176,8 +176,7 @@ account] and custom roles:
 +
 ====
 ----
-$ oc process logging-deployer-account-template -n openshift \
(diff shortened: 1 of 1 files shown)


sosiouxme:
@ahardin-rh this one isn't really priority but there's no reason I know of for it to wait, if you're in the neighborhood :)

ahardin-rh:
@sosiouxme Thanks! Looks good!

ahardin-rh:
[rev_history]
|link:../install_config/aggregate_logging.html[Aggregating Container Logs]
|Updated with guidance to use  `oc new-app` instead of `oc process | oc create` for logging.
%

PR #2107, opened 2016-05-16
Bug 1333465, added SSH topic to Dev Guide

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1333465

### DIFF
--- a/dev_guide/ssh_environment.adoc
+++ b/dev_guide/ssh_environment.adoc
@@ -1,0 +1,56 @@
+= Opening a Remote Shell to Containers
+{product-author}
+{product-version}
+:data-uri:
+:icons:
+:experimental:
+:toc: macro
+:toc-title:
+:prewrap!:
+
+toc::[]
+
+== Overview
+
+The `oc rsh` command allows you to locally access and manage tools that are on
+the system. The secure shell (SSH) is the underlying technology and industry
+standard that provides a secure connection to the application. Access to
+applications with the shell environment is protected and restricted with
+Security-Enhanced Linux (SELinux) policies.
+
+[[ssh-environment-start-a-secure-shell-session]]
+== Start a Secure Shell Session
+
+Open a remote shell session to a container:
+----
+$ oc rsh <pod>
+----
+
+While in the remote shell, you can issue commands as if you are inside the
+container and perform local operations like monitoring, debugging, and using CLI
+commands specific to what is running in the container.
+
+For example, in a MySQL container, you can count the number of records in the
+database by invoking the link:../using_images/db_images/mysql.html[`mysql`
+command], then using the the prompt to type in the `SELECT` command. You can
+also use use commands like `ps(1)` and `ls(1)` for validation.
+
+`*BuildConfigs*` and `*DeployConfigs*` map out how you want things to look and
+pods (with containers inside) are created and dismantled as needed. Your changes
+are not persistent. If you make changes directly within the container and that
+container is destroyed and rebuilt, your changes will no longer exist.
+
+[NOTE]
+====
+link:../dev_guide/executing_remote_commands.html[`oc exec`] can be used to
+execute a command remotely. However, the `oc rsh` command provides an easier way
+to keep a remote shell open persistently.
+====
+
+[[ssh-environment-secure-shell-session-help]]
+== Secure Shell Session Help
+
+For help with usage, options, and to see examples:
+----
+$ oc rsh -h
+----
--- a/_build_cfg.yml
+++ b/_build_cfg.yml
@@ -493,6 +493,8 @@ Topics:
         File: quickstarts
       - Name: Ruby on Rails
         File: ruby_on_rails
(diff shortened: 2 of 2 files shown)


ahardin-rh:
@adellape @tnguyen-rh @tpoitras @bfallonf PTAL

dev_guide/ssh_environment.adoc
```diff
@@ -0,0 +1,37 @@
+= Managing Applications in a Secure Shell Environment
```
tnguyen-rh: This is singular "an SSH env" while the `_build_cfg.yml` Name is plural.  Suggest changing one to be consistent w/ the other.

dev_guide/ssh_environment.adoc
```diff
@@ -0,0 +1,37 @@
...
+
+Open a remote shell session to a container:
+----
+$ oc rsh <pod_ID>
+----
+
+[[ssh-environment-secure-shell-session-help]]
+== Secure Shell Session Help
+
+For help with usage, options, and to see examples, use the following command.
```
tnguyen-rh: Probably you can s/, use the following comamnd./:/

dev_guide/ssh_environment.adoc
```diff
@@ -0,0 +1,37 @@
...
+:experimental:
+:toc: macro
+:toc-title:
+:prewrap!:
+
+toc::[]
+
+== Overview
+
+Managing applications in a secure shell (SSH) environment provides specialized
```
tnguyen-rh: This boils down to "Managing provides tools", which i don't understand.

tnguyen-rh:
@ahardin-rh One fuzziness and a couple nits.

bfallonf:
@ahardin-rh I think this could do with a bit of an explanation on what you can do once you've ssh'ed into the container. You mention advanced operations and general debugging. Could you give examples of this?

Also, is there a difference between this process and what's outlined in [executing remote commands](https://docs.openshift.com/enterprise/3.2/dev_guide/executing_remote_commands.html)? Is ssh'ing about getting into the container and staying in there, but executing remote commands a one-time thing?

dev_guide/ssh_environment.adoc
ahardin-rh: @tnguyen-rh, you're right and I keep beating my head off of the desk trying to figure out a better way to rewrite this. Thinking on it...

ahardin-rh:
@tnguyen-rh Reworked this topic based on our conversation this morning and your most excellent pointers. PTAL and let me know what you think. Thanks for all of your help!

dev_guide/ssh_environment.adoc
```diff
@@ -0,0 +1,55 @@
...
+----
+$ oc rsh <pod_ID>
+----
+
+While in the remote shell, you can issue commands as if you are inside the
+container and perform local operations like monitoring, debugging, and using CLI
+commands specific to what is running in the container.
+
+For example, in a MySQL container, you can use `SELECT COUNT([column]) FROM
+[table];` to count the number of records in the database. You can also use use
```
tnguyen-rh: That `SELECT` command is pretty cool.  Unfortunately, it's not a command that you can type directly into the shell; the actual shell command is called `mysql`.  You invoke that command to get another prompt, and then type in the `SELECT` command at that prompt.  Poking around, i see that we already have an example documented -- see `using_images/db_images/mysql.adoc` around line 106.  Perhaps you can marry this `SELECT` command w/ that example (copying it here).  Another idea is to link to it.

tnguyen-rh:
@ahardin-rh I found only one weirdness.  Apart from that, LGTM! :+1:

ahardin-rh:
@tnguyen-rh Thanks! Fixed!

bfallonf:
@ahardin-rh As discussed, this looks good! I give it a thumbs up :+1:

luciddreamz:
This looks good. At some point it might be worth mentioning that users should SSH into their pods and use the `env` command to list all available environment variables, but we can probably mention that in another section.

xingxingxia:
As QA of the bug, examined the PR, LGTM except:
1. An extra 's' in "make a changes"
2. Better to write as "oc rsh <pod>" instead of "oc rsh <pod_ID>". I suggest this because several days ago someone asks me about the info in `oc delete -h` which uses "ID" in many places but actually `oc delete pod $uid` outputs:  pods "9fddc973-2b8d-11e6-9576-0e3d364e19a5" not found. (Where $uid comes from oc get pod <pod> -o yaml | grep uid). So "ID" seems a bit of confusing. (When you have time, could update this issue in other docs: such as https://docs.openshift.org/latest/cli_reference/basic_cli_operations.html which tells users to use "ID")
Thank you!

ahardin-rh:
[rev_history]
|xref:../dev_guide/ssh_environment.adoc[Opening a Remote Shell to Containers]
|Added a new topic on opening a remote shell to containers.
%

Issue #2095, opened 2016-05-13
Sizing recommendations have misprint


etsauer:
The sizing recommendations here: https://github.com/openshift/openshift-docs/blob/master/install_config/install/prerequisites.adoc#sizing-recommendations

It specifies 1 CPU and 2.5 GB Ram per _**1000**_ pods... I believe that should say _**100**_.

adellape:
@etsauer That addition came from this PR: https://github.com/openshift/openshift-docs/pull/1961

Which got its info from this Issue: https://github.com/openshift/openshift-docs/issues/1949

The Issue mentions "per 1k pods". @timothysc Can you verify?

etsauer:
@adellape there's no way the recommendation could be 1 core & 2.5 GB RAM for 1000 pods. Also, in the next line there's some math that calculates to 20 cores and 50 GB RAM for 2000 pods. That math doesn't add up. Either way, i'd like to know where the numbers came from. Even it was meant to be _**100**_, that's assuming VERY tiny pods. A single java pod could easily consume 2GB Ram on its own. These measures are highly dependent on the language, and nature of the workload running inside of it. I'm all for recommending reasonable defaults to start with, but these don't seem very reasonable.

timothysc:
The sizing is for the **master daemons in a HA configuration**.  

If you want to run 20k pod cluster you will need to size your 
3 Masters, 3 - Etcd servers appropriately. 

1 cpu core, and 2.5 GB of memory atop of the defaults per 1k pods. So for 20k pods that would be (20+2) cpu cores and and (50+8) GB of ram.

That means each **master** machine is large. 

This **is not** the sizing of compute nodes.. It's the sizing of master daemons.

etsauer:
ok, that makes way more sense. That section needs to be more clear about being sizing for masters. However, if we're going to have sizing for masters, it seems we should also talk about calculating sizing for nodes.

adellape:
@bfallonf FYI on the above, per https://github.com/openshift/openshift-docs/pull/1961.

timothysc:
Sizing of the nodes, is entirely based on workloads being run.  Operators will need to determine their workload profile and size appropriately based on the profiles of their pods - some system reserve for those machines.

etsauer:
@timothysc agreed its very subjective, but its confusing if we provide guidelines for masters and not nodes. At minimum we need to say something about calculating workload size, leaving room for overhead, and determining node size based on that.

[...]

PR #2108, opened 2016-05-17
Issue 2095: Fixed System Requirements phrasing and added node host info

bfallonf:
@etsauer @timothysc As per #2095 

I'll admit, the part about node hosts is just kinda...blah. I don't think it's saying anything the admin would not expect to do for themselves. Is there anything else that can be added here?

I also changed the wording of the master host section.

Please let me know if this needs adjusting.

### DIFF
--- a/install_config/install/prerequisites.adoc
+++ b/install_config/install/prerequisites.adoc
@@ -114,12 +114,25 @@ this during or after installation.
 
 [[sizing-recommendations]]
 === Sizing Recommendations
+The following apply to production environments. Test or sample environments will
+function with the minimum requirements.
 
-In general, an {product-title} cluster should have 1 CPU core and 2.5 GB of
-memory, on top of the defaults in the table above, for each 1000 pods.
-Therefore, the recommended size of a cluster of 2000 pods would be 20 CPU cores
-and 50 GB of RAM, as well as the minimum requirements for a master host of 2 CPU
-cores and 8 GB of RAM.
+*Master Hosts*
+
+In a highly-available {product-title} cluster with external etcd, a master host
+should have 1 CPU core and 2.5 GB of memory, on top of the defaults in the table
+above, for each 1000 pods. Therefore, the recommended size of master host in an
+{product-title} cluster of 2000 pods would be 20 CPU cores and 50 GB of RAM, as
+well as the minimum requirements for a master host of 2 CPU cores and 8 GB of
+RAM.
+
+*Node Hosts*
+
+The size of a node host depends on the expected size of its workload. As an
+{product-title} cluster administrator, you will need to calculate the expected
+workload, then add about 10% for overhead. For production environments, allocate
+enough resources so that node host failure does not affect your maximum
+capacity.
 
 [[configuring-core-usage]]
 


etsauer:
@bfallonf @timothysc this needs some major updating, as it references the 3.0 docs and (i'm just realizing) has some sections out of order, but this is the guidance the NA Consulting teams give on the subject.

http://playbooks-rhtconsulting.rhcloud.com/playbooks/installation/installation.html#node-sizing

Perhaps we can take some things from here.

install_config/install/prerequisites.adoc
```diff
@@ -114,12 +114,20 @@ this during or after installation.
 
 [[sizing-recommendations]]
 === Sizing Recommendations
+*Master Hosts*
 
-In general, an {product-title} cluster should have 1 CPU core and 2.5 GB of
+In general, an {product-title} master host should have 1 CPU core and 2.5 GB of
```
timothysc: In a HA-deployment

bfallonf:
@etsauer @timothysc Cool. Added the info I thought was relevant from that 3.0 doc and changed some of the master info to talk about Tim's suggestion.

I'd be hesitant to add more info from that doc. It even talks about how outdated it will be soon.

Thoughts?

install_config/install/prerequisites.adoc
```diff
@@ -114,12 +114,27 @@ this during or after installation.
...
+RAM.
+
+*Node Hosts*
+
+The size of a node host depends on the size of the workload that will be run on
+the host. As an {product-title} cluster administrator, you will need to
+calculate the expected workload, making sure to leave some room for reserve
+resources, and use this to determine the size of the node host.
+
+In general, add 30% to the maximum expected workload. This will account for 10%
```
etsauer: The suggestion to add 20% for node failure is for planning capacity for an entire cluster, not for individual nodes. I think the better thing to focus on is that nodes are simply container hosts, and their size should be dictated by the profiles of the workloads expected to run on them. Adding 10% for overhead is still valid though.

bfallonf: @etsauer I read the doc you linked differently. I took the 20% to mean it would be a good size to pick up the slack if other nodes fail. Otherwise, the way you suggest it, what are we adding 20% to? Where does that 20% come from, if not node memory or whatever?

etsauer: @bfallonf yes that's true, the 20% is meant for picking up slack in case a node fails. It just sounds a bit strange when we're talking about individual nodes, seemingly suggesting that 20% of a node could fail somehow. That figure also makes the assumption that there are at least 5 schedulable nodes in the cluster. I think in this context, when we're trying to be brief it makes more sense to just say calculate expected maximum workload per node, add 10% for overhead, and for production deployments, make sure to allocate enough resources such that a host failure doesn't restrict your maximum capacity.

bfallonf:
@etsauer Thanks for the additional help. Changed to:

"The size of a node host depends on the expected size of its workload. As an {product-title} cluster administrator, you will need to calculate the expected workload, then add about 10% for overhead. For production environments, allocate enough resources so that node host failure does not affect your maximum capacity."

WDYT?

etsauer:
lgtm

bfallonf:
@etsauer Cool. Thanks. I'll move forward with this.

@tpoitras @tnguyen-rh @ahardin-rh @adellape Peer review, please?

bfallonf:
[rev_history]
|Added prerequisite information to node host section of link:../install_config/install/prerequisites.html#system-requirements[System Requirements].
%

tnguyen-rh:
@bfallonf LGTM.

bfallonf:
@tnguyen-rh Great. Thanks. I'll merge.

PR #2109, opened 2016-05-17
fix template to new asciibinder standard requirements

tpoitras:
Ping @ahardin-rh @adellape @bfallonf @tnguyen-rh @vikram-redhat 

Brice and Vikram and I met with Lee Newson to discuss the build system and how our docs get ported to the portal.

I've modified our standard topic template here. Check it out, it turns out that a lot of the stuff we've been including in all our topic headers is no longer needed, due to advancements in Asciibinder. It's much simpler and leaner now, so please use this one from now on.

It also helps the tooling team!

### DIFF
--- a/contributing_to_docs/templates/topic_template.adoc
+++ b/contributing_to_docs/templates/topic_template.adoc
@@ -1,13 +1,8 @@
 = Replace with topic title
-{product-author}
-{product-version}
-:icons: font
 :experimental:
 :toc: macro
 :toc-title:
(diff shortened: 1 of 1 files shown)


adellape:
@tpoitras I tried using this in a new topic today and noticed that I still needed `:data-uri:` which isn't mentioned here, otherwise images wouldn't embed.

http://asciidoctor.org/docs/render-documents/#managing-images

PR #2111, opened 2016-05-17
add explanation of scoped tokens

deads2k:
Right now, scoped tokens can only be created by cluster admins.  This severely limits their utility.  We have a card to fix this: https://trello.com/c/Tcabgp9F/696-3-auth-scopes-user-mgmt-of-tokens, but we haven't gotten to it yet.

@wjiangjay

### DIFF
--- a/_build_cfg.yml
+++ b/_build_cfg.yml
@@ -349,6 +349,8 @@ Topics:
     Distros: openshift-origin,openshift-enterprise,atomic-*
   - Name: Managing Authorization Policies
     File: manage_authorization_policy
+  - Name: Scoped Tokens
+    File: scoped_tokens
   - Name: Managing Security Context Constraints
     File: manage_scc
     Distros: openshift-origin,openshift-enterprise
--- a/admin_guide/scoped_tokens.adoc
+++ b/admin_guide/scoped_tokens.adoc
@@ -1,0 +1,48 @@
+= Scoped Tokens
+{product-author}
+{product-version}
+:data-uri:
+:icons:
+:experimental:
+:toc: macro
+:toc-title:
+
+toc::[]
+
+== Overview
+A user may wish to give another entity the power to act as him, but only in a limited way.
+For example, a project admin may wish to delegate the power to create pods.  One way to do 
+this is to create a scoped token.
+
+A scoped token, is a token that identifies as a given user, but is limited to certain actions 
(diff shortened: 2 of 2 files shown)


sosiouxme:
@deads2k can I ask how scoping is relevant in the context of an oauthclient?

deads2k:
An oauthclient can now be scope limited by a cluster-admin.  This means that the oauth client can only request certain scopes and these restrictions are a way to prevent oauth clients from gaining unintended power.

After https://github.com/openshift/origin/pull/8878, an SA and SA token will become valid oauth clients.  Those oauth clients are scope restricted to their own namespace and user:info and user:check-access permissions.  This makes it possible for a user to install an pod in their namespace that can authenticate using openshift, verify identity, verify user access, and even get a token that is valid inside of the namespace to perform actions.  We want to make sure that an SA oauth client cannot get a token valid for other namespaces, to prevent naughty access.

This SA as oauth client is useful for cases like:
 1. jenkins managing my project's builds
 2. personal web console of some description
 3. daikon spark cluster management in my namespace.

When we add the ability to link namespaces for SA client token restrictions (https://trello.com/c/zXFsdnCF/720-allow-sas-to-request-cross-namespace-scopes-if-they-are-explicitly-allowed), we'll have even more power.

deads2k:
@liggitt you want this to wait on https://trello.com/c/Tcabgp9F/696-5-auth-scopes-user-mgmt-of-tokens ?

deads2k:
@adellape @ahardin-rh I don't think we'll wait on the next story.  This is getting mentioned in release notes.

ahardin-rh:
@deads2k Thanks! Can you please check my labels and milestones before I merge? Is this for OSE 3.2 next release, or a future release?

deads2k:
> @deads2k Thanks! Can you please check my labels and milestones before I merge? Is this for OSE 3.2 next release, or a future release?

It's for future.  1.3.

deads2k:
I think I retagged it to the correct one.

ahardin-rh:
@deads2k Thanks!

ahardin-rh:
rev history to be added in follow-up PR

adellape:
To whoever follows up on this, I had to add `Distros: openshift-origin` to this scoped_tokens topic in the build_cfg as part of https://github.com/openshift/openshift-docs/pull/2138. So during the follow-up PR, it should remove that distro.

Issue #2113, opened 2016-05-17
scaleup docs incorrect. References a scaleup.yml file that doesn'T exist 


ivanthelad:
https://docs.openshift.com/enterprise/3.2/install_config/install/advanced_install.html#adding-nodes-advanced

Docs state the file should be under 
https://github.com/openshift/openshift-ansible/tree/master/playbooks/byo/openshift-cluster
but file exists under 
https://github.com/openshift/openshift-ansible/blob/master/playbooks/byo/openshift-node/scaleup.yml

adellape:
Fixed in https://github.com/openshift/openshift-docs/pull/2135

PR #2135, opened 2016-05-19
Correct path of `scaleup.yml` script in install

cpitman:
Ran into this one today. There is no `scaleup.yml` script in the in the `openshift-cluster` folder. There is one in both `openshift-master` and `openshift-node`. From my testing, `openshift-node/scaleup.yml` does the required task.


bfallonf:
GH issue #2113 seems to be related

PR #2137, opened 2016-05-19
BZ1336608: Add note regarding secrets and Ceph

tpoitras:
Added a note to specify that the secret must be created in the same project as the PVC and pod in order for Ceph storage to work with a PV.


tpoitras:
hey @bfallonf @ahardin-rh @tnguyen-rh 
is this cool to merge?

ahardin-rh:
LGTM!

PR #2139, opened 2016-05-20
Add doc for expose with sharding

ghost:
What I missed is the documentation how to use the router sharding.

This PR starts the doc for dev users

### DIFF
--- a/dev_guide/routes.adoc
+++ b/dev_guide/routes.adoc
@@ -137,12 +137,12 @@ host name, resulting in validation errors.
 == Creating Routes with Labels
 
 You can create a route based on labels. A explanation how can be found at
-link:../architecture/core_concepts/routes.adoc#router-sharding
+link:../architecture/core_concepts/routes.adoc#router-sharding[Router Sharding]
+The labels can be
+link:../install_config/install/deploy_router.html#creating-router-shards[selected by routers]
+to create router shards.
 
-How router sharding is created is explained at
-link:../install_config/install/deploy_router.adoc#creating-router-shards[Creating Router Shards]
-you need to add the labels for the dedicated routes.
-
+For example, to create a route with label `sla=low`:
 ----
 $ oc expose svc/low-sla-frontend \
     --hostname=low-sla.example.com \
--- a/dev_guide/routes.adoc
+++ b/dev_guide/routes.adoc
@@ -133,16 +133,20 @@ TLS termination and a default certificate, which may not match the requested
 host name, resulting in validation errors.
 ====
 
-[[creating-routes-shards]]
+[[creating-labeld-routes]]
+== Creating Routes with Labels
 
-== Creating Routes with router shards
+You can create a route based on labels. A explanation how can be found at
+link:../architecture/core_concepts/routes.adoc#router-sharding
 
-In case us use the router sharding possibility 
-link: ../install_config/install/deploy_router.adoc#creating-router-shards[Creating Router Shards]
+How router sharding is created is explained at
+link:../install_config/install/deploy_router.adoc#creating-router-shards[Creating Router Shards]
 you need to add the labels for the dedicated routes.
 
 ----
-oc expose svc/low-sla-frontend --hostname=low-sla.example.com --labels='sla=low'
+$ oc expose svc/low-sla-frontend \
+    --hostname=low-sla.example.com \
+    --labels='sla=low'
 ----
 
 Further information on all types of
--- a/dev_guide/routes.adoc
+++ b/dev_guide/routes.adoc
@@ -133,7 +133,7 @@ TLS termination and a default certificate, which may not match the requested
 host name, resulting in validation errors.
 ====
 
-[[creating-labeld-routes]]
+[[creating-labeled-routes]]
 == Creating Routes with Labels
 
 You can create a route based on labels. A explanation how can be found at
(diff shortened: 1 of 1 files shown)


dev_guide/routes.adoc
```diff
@@ -133,6 +133,18 @@ TLS termination and a default certificate, which may not match the requested
 host name, resulting in validation errors.
 ====
 
+[[creating-routes-shards]]
+
+== Creating Routes with router shards
```
tnguyen-rh: - The conceptual construct of "router sharding" involves labels on routes, it's true.  However, the mere existence of labels on a route does not constitute a router shard, only the *possibility* (given a corresponding selection expression in the router object) of forming a router shard.  How about changing the header to "Creating Routes with Labels" and then explaining this in the paragraph (or better yet, linking to the "Router Sharding" section in Architecture -> Core Concepts -> Routes?

- No need for blank line between the anchor and the header.

dev_guide/routes.adoc
```diff
@@ -133,6 +133,18 @@ TLS termination and a default certificate, which may not match the requested
...
+
+In case us use the router sharding possibility 
+link: ../install_config/install/deploy_router.adoc#creating-router-shards[Creating Router Shards]
```
tnguyen-rh: - The `link:` syntax does not permit a space between `link:` and the URL.
- I suggest rewriting the paragraph (as explained above), and adding another link.

dev_guide/routes.adoc
```diff
@@ -133,6 +133,18 @@ TLS termination and a default certificate, which may not match the requested
...
+you need to add the labels for the dedicated routes.
+
+----
+oc expose svc/low-sla-frontend --hostname=low-sla.example.com --labels='sla=low'
```
tnguyen-rh: - This command needs a `$` prompt.
- Because these docs are sometimes displayed very narrowly, we prefer to use backslash continuation style.  For example:
```bash
$ some-very-long-command \
    --interesting-option \
    --more args
```
(There are four SPC characters at the beginning of each continuation line.)

tnguyen-rh:
@cw-aleks Thanks for the patch!  I think it's a good idea and fills a definite gap in the docs.  I've made some line-comments and look forward to merging this PR once those issues are properly addressed.

cw-aleks:
@tnguyen-rh thanks for feedback. I have now Adopted your comments.

dev_guide/routes.adoc
```diff
@@ -133,6 +133,22 @@ TLS termination and a default certificate, which may not match the requested
 host name, resulting in validation errors.
 ====
 
+[[creating-labeled-routes]]
+== Creating Routes with Labels
+
+You can create a route based on labels. A explanation how can be found at
+link:../architecture/core_concepts/routes.adoc#router-sharding
```
tnguyen-rh: - This is not accurate -- the route is not **based on** labels.
- There is no link text here.
- The URL needs `.html` rather than `.adoc`.

How about we address all three issues, and make things more concise, as well, with:

```asciidoc
link:../architecture/core_concepts/routes.adoc#router-sharding[Router sharding]
requires routes with labels.
```

dev_guide/routes.adoc
```diff
@@ -133,6 +133,22 @@ TLS termination and a default certificate, which may not match the requested
...
+
+How router sharding is created is explained at
+link:../install_config/install/deploy_router.adoc#creating-router-shards[Creating Router Shards]
+you need to add the labels for the dedicated routes.
+
```
tnguyen-rh: Similarly to the first sentence, how about if we use:

```asciidoc
The labels can be
link:../install_config/install/deploy_router.html#creating-router-shards[selected by routers]
to create router shards.
```

and remove the intervening blank line, so that the two sentences form one paragraph?

dev_guide/routes.adoc
```diff
@@ -133,6 +133,22 @@ TLS termination and a default certificate, which may not match the requested
...
+----
```
tnguyen-rh: This example needs an introductory blurb, something like, "For example, to create a route with label `sla=low`:" (note the trailing `:`).

tnguyen-rh:
@cw-aleks 
Thanks.  :+1:  I appreciate your persistence.  Just a few more refinements, and this can be merged w/o requiring further touch-ups.

dev_guide/routes.adoc
cw-aleks: Well I think this code snippets say something different.

https://github.com/openshift/origin/blob/master/pkg/cmd/infra/router/router.go#L53
https://github.com/openshift/origin/blob/master/pkg/cmd/infra/router/router.go#L92
https://github.com/kubernetes/kubernetes/blob/master/pkg/api/v1/types.go#L2606

As far as I understand the code right you need labels for router selection.

In case the VAR ROUTE_LABELS or LabelSelector is not set, which is default, the else tree 'o.Labels = labels.Everything()' is in use.

For router sharding you NEED to label ALL Routers or the apps wil also be deployed to the not labeled router.

That's the purpose of this sentence ;-)

tnguyen-rh:
@cw-aleks Could you please squash the changes down to one commit?  I'll be glad to merge it once that's done.

cw-aleks:
@tnguyen-rh sorry i don't understand what you mean with **squash**.

Never the less I think this pr is not useful any more.
I will close it.

tnguyen-rh:
@cw-aleks
If you do a web search for `git squash` you'll see quite a few results.
[Here](https://github.com/mozilla/pdf.js/wiki/Squashing-Commits) is one of them.

Anyway, thanks for revisiting this PR.  I hope we can work together in the future.

PR #2143, opened 2016-05-24
Fix `oc process | oc create` example

chrira:
oc create command fix


tnguyen-rh:
[rev_history]
|link:../dev_guide/templates.html[Templates]
|Fixed `oc process` example in the link:../dev_guide/templates.html#templates-parameters[Parameters] section.

PR #2145, opened 2016-05-24
Add "TLS Handshake Timeout" section

tnguyen-rh:
Fixes 1309881

I decided against adding links to the new section because i couldn't find a good place in the proposed locations, "Master and Node Configuration" and "Configuring the SDN".  Suggestions welcome.

@mwoodson WDYT?

### DIFF
--- a/admin_guide/sdn_troubleshooting.adoc
+++ b/admin_guide/sdn_troubleshooting.adoc
@@ -630,6 +630,23 @@ provide the output from the debugging script)
 * Kube - declare as a LoadBalancer and something _you_ have to write does the rest
 * OS/AE - Both use the router
 
+[[tls-handshake-timeout]]
+=== TLS Handshake Timeout
+
+When a pod fails to deploy, check its docker log for a TLS handshake timeout:
+
+----
+$ docker log <container_id>
+...
+[...] couldn't get deployment [...] TLS handshake timeout
+...
+----
+
+This condition, and generally, errors in connection authentication,
+may be caused by a large difference in the MTU values between tun0
+and the primary interface (e.g., eth0),
+such as when tun0 MTU is 1500 and eth0 MTU is 9000 (jumbo frames).
+
 === Other debugging notes
 * Peer interfaces (of a Linux virtual ethernet pair) can be determined with `ethtool -S _ifname_`
 * Driver type: `ethtool -i _ifname_`
--- a/admin_guide/sdn_troubleshooting.adoc
+++ b/admin_guide/sdn_troubleshooting.adoc
@@ -642,7 +642,7 @@ $ docker log <container_id>
 ...
 ----
 
-This condition, and generally, errors in connection authentication,
+This condition, and generally, errors in establishing a secure connection,
 may be caused by a large difference in the MTU values between tun0
 and the primary interface (e.g., eth0),
 such as when tun0 MTU is 1500 and eth0 MTU is 9000 (jumbo frames).


admin_guide/sdn_troubleshooting.adoc
```diff
@@ -630,6 +630,23 @@ provide the output from the debugging script)
...
+When a pod fails to deploy, check its docker log for a TLS handshake timeout:
+
+----
+$ docker log <container_id>
+...
+[...] couldn't get deployment [...] TLS handshake timeout
+...
+----
+
+This condition, and generally, errors in connection authentication,
```
liggitt: I wouldn't call this "authentication", since it's too easy to confuse with OpenShift authentication.

tnguyen-rh: @liggitt Good point.  I've changed it to read "errors in establishing a secure connection".  WDYT?

tnguyen-rh:
@adellape @ahardin-rh @tpoitras @bfallonf PTAL

admin_guide/sdn_troubleshooting.adoc
```diff
@@ -630,6 +630,23 @@ provide the output from the debugging script)
...
 
+[[tls-handshake-timeout]]
+=== TLS Handshake Timeout
+
+When a pod fails to deploy, check its docker log for a TLS handshake timeout:
+
+----
+$ docker log <container_id>
+...
+[...] couldn't get deployment [...] TLS handshake timeout
```
bfallonf: What are the [...]'s supposed to be? Pods name variables? I realise this is example readout, but would they be better as <pod_name> or a made up name like pod1?

bfallonf:
@tnguyen-rh One question, but the rest of the info looks good. 

However, unless I'm misunderstanding, I'm wondering if the placement could be more appropriate. There's a section about "Debugging Pod Ingress". Could this be contained in an overarching section about pods, with this as another section inside it? I'd also suggest the title be more about the problem, rather than the solution. So:

Debugging Pods
- Pod Ingress
- - (Then the ingress stuff)
- Pods Failing to Deploy
- - (Then this bit about handshakes)

I'm just not sold on the "Miscellaneous Notes" heading, and this is my way around it :P

admin_guide/sdn_troubleshooting.adoc
tnguyen-rh: It's supposed represent elided text, unimportant for the task of recognizing this error condition.  How about i drop the square-braces and just use `...` for that?

tnguyen-rh:
@bfallonf
I see your point.  It's a tricky question.  IIUC, pods failing to deploy is just one symptom of a TLS handshake timeout, the underlying condition; any number of strange behaviors, initiated by any CLI command, can be traced to this.  I see two major branches in the going-forward documentation strategy around this issue:

- If the weirdness (in the networking layer) persists, we can link to this section from the various places where the problem manifests.  For example, from dev guide "Deployments" topic, or (worst case, looks bad for everyone, scenario) from the various CLI topics.

- Otherwise (the networking layer is "fixed" somehow), we can just drop this section once it's no longer relevant.

Anyway, that's the reason for this text's placement.

admin_guide/sdn_troubleshooting.adoc
bfallonf: Ah OK. Hmm probably works best as is then.

bfallonf:
I understand the placement now. I can see what you mean.

Would you consider the title change though? With troubleshooting docs, it's always best to title the topic with the symptom rather than the cure, for findability's sake. But from what you're saying, it might be harder to determine the title. Thoughts?

tnguyen-rh:
@liggitt PTAL

@bfallonf I'd consider it up until the point when @liggitt gives the LGTM. :smile:

tnguyen-rh:
@liggitt Thanks for the virtual LGTM (over email).  :+1:

[rev_history]
|link:../admin_guide/sdn_troubleshooting.html[Troubleshooting OpenShift SDN]
|Added link:../admin_guide/sdn_troubleshooting.html#tls-handshake-timeout[TLS Handshake Timeout] section.
%

PR #2150, opened 2016-05-25
Information about deployment triggers

0xmichalis:
@bparees we can start from here. 

cc: @adellape @tnguyen-rh would love your suggestions for working and how to structure the table

### DIFF
--- a/dev_guide/deployments.adoc
+++ b/dev_guide/deployments.adoc
@@ -217,7 +217,7 @@ failed and has not been manually deleted or pruned).
 == Triggers
 
 A deployment configuration can contain triggers, which drive the creation of
-new deployments in response to events, both inside and outside OpenShift.
+new deployments in response to events, only inside OpenShift at the moment.
 
 WARNING: If no triggers are defined on a deployment configuration, deployments
 must be link:#start-deployment[started manually].
@@ -225,10 +225,10 @@ must be link:#start-deployment[started manually].
 [[config-change-trigger]]
 === Configuration Change Trigger
 
-The ConfigChange trigger results in a new deployment whenever changes are
-detected to the replication controller template of the deployment configuration.
+The ConfigChange trigger results in a new deployment whenever new changes are
+detected in the pod template of the deployment configuration.
 
-NOTE: If a ConfigChange trigger is defined on a deployment configuration,
+NOTE: If only a ConfigChange trigger is defined on a deployment configuration,
 the first deployment will be automatically created soon after the deployment
 configuration itself is created.
 
@@ -247,7 +247,7 @@ triggers:
 === Image Change Trigger
 
 The ImageChange trigger results in a new deployment whenever the value of an
-image stream tag changes.
+image stream tag changes either by a build or because it was imported.
 
 The following is an example of an ImageChange trigger:
 
@@ -269,9 +269,15 @@ triggers:
 ====
 
 With the above example, when the `latest` tag value of the `origin-ruby-sample`
-image stream changes and the new tag value differs from the current image
+image stream changes and the new image value differs from the current image
 specified in the deployment configuration's `helloworld` container, a new
-deployment is created using the new tag value for the `helloworld` container.
+deployment is created using the new image for the `helloworld` container.
+
+NOTE: If an ImageChange trigger is defined on a deployment configuration (with
+a ConfigChange trigger or with automatic=true) and the ImageStreamTag pointed
+by the ImageChange trigger does not exist yet then the first deployment will
+automatically start as soon as an image is imported or pushed by a build to the
+ImageStreamTag.
 
 [[strategies]]
 == Strategies


dev_guide/deployments.adoc
```diff
@@ -247,7 +247,7 @@ triggers:
 === Image Change Trigger
 
 The ImageChange trigger results in a new deployment whenever the value of an
-image stream tag changes.
+image stream tag changes either by a build or because it was imported.
```
bparees: end this sentence with something like "if a configchangetrigger exists."

dev_guide/deployments.adoc
```diff
@@ -273,6 +273,14 @@ image stream changes and the new tag value differs from the current image
 specified in the deployment configuration's `helloworld` container, a new
 deployment is created using the new tag value for the `helloworld` container.
 
+The following table explains the possible combinations between ConfigChange
+and ImageChange triggers.
+
+  Trigger                  | true                                      | false
+  -----------------------  | ----------------------------------------- | ------------------------------------------------------
+  ConfigChange             | Deployments are started automatically     | Deployments need to be started manually
+  ImageChange (automatic)  | Images are always resolved automatically  | Only the initial image will be resolved in the template
```
bparees: I think this can work but i was thinking of a table like:
```                       
            configchange | true | false
-----------------------------------------
ICT |
-----------------------------------------
true |                            X   |      Y
------------------------------------------
false |                          Z   |       W
```

that's harder to extend if we add more behavior later though.

at a minimum i'd change "images are always resolved automatically" to "Image value is resolved from the imagestreamtag referenced in the imagechangetrigger".  That may not fix in a box though :)

the important thing someone needs to be able to take away from this table is understanding the interaction of the settings...  i'm not sure knowing that "images are resolved automatically" and "deployments are started automatically" would lead me to realize that if I have an ICT=true and a CCT, that i'm going to get deployments when the image changes (because i don't realize that "images are resolved automatically" really means "the pod template is updated automatically which will result in the CCT firing"

dev_guide/deployments.adoc
rhcarvalho: @bparees do you mean that a "ConfigChangeTrigger" need to exist for an "ImageChange" trigger to function?

dev_guide/deployments.adoc
rhcarvalho: This is enlightening! So I understand that `ImageChange` is not really a "trigger" in the sense that it can trigger deployments, instead it can trigger a change in the `PodTemplate` that, in turn, may trigger a deployment if a `ConfigChange` trigger is present...

dev_guide/deployments.adoc
rhcarvalho: > @bparees do you mean that a "ConfigChangeTrigger" need to exist for an "ImageChange" trigger to function?

Oh yeah looking at the table and other comments that seems to be indeed the case... This explains some puzzling behavior @bbguimaraes was observing with the logging templates lately!

dev_guide/deployments.adoc
rhcarvalho: @bbguimaraes take a look at this ^^^ !

I believe it answers your questions about how ConfigChange and ImageChange triggers in DeploymentConfigs interact.

dev_guide/deployments.adoc
bparees: yeah, without the CCT (but with an ICT in automatic mode) the DC will be updated w/ the new image information (at a minimum this will happen when a deployment is explicilty requested), but no deployment will be triggered automatically. I believe.

kargakis: Correct. CCT is responsible for triggering new deployments on pod template
changes. If it's on, **any template change** should trigger a deployment,
if it's off, template changes will not trigger deployments automatically.
Essentially, CCT is (almost) a global on/off for a DC. ICTs set to
automatc: true will always update images in templates, automatic: false
will update only for the initial deployment so people wont get stuck with
ImagePullBackOff errors

bparees: which still leaves me thinking "automatic" is the wrong name for that field (should be something like "initialOnly" or something) but i guess that shipped has sailed.  

/me waves to ship in distance.

dev_guide/deployments.adoc
bbguimaraes: @rhcarvalho: (sorry, I was out last week) my issue is the opposite: the images
are already on the node (they aren't built, so openshift doesn't "know" about
them) and the only thing I need is for the deployment to start when the DC is
created.  If I only have the ConfigChange trigger, that's what happens.  But
when I add an ImageChangeTrigger, I get this on the log:

    Couldn't create initial deployment for deploymentConfig "x/y": DeploymentConfig "y" is invalid: triggers[1].imageChange.tag: Invalid value: "latest": no image recorded for x/y:latest

kargakis: Bruno, can you retry what you are doing on HEAD of Origin and report back
what you observe?
The two controllers in play have been refactored since your working version.

> so openshift doesn't "know" about them

Honestly, I don't understand this part. Your image should be in an image
stream so OpenShift
knows about it, right?

kargakis:
All: we restore ICTs to run deployments without CCs in https://github.com/openshift/origin/pull/9132

tnguyen-rh:
@kargakis Here's the same table transformed into standard asciidoc syntax:

```asciidoc
[cols="2,5,5",options="header"]
|===
| Trigger | true | false

|ConfigChange
|Deployments are started automatically
|Deployments need to be started manually

|ImageChange (automatic)
|Images are always resolved automatically
|Only the initial image will be resolved in the template
|===
```

The `cols="2,5,5"` bit specifies the relative width of each column.  More info on asciidoc tables [here](http://asciidoctor.org/docs/asciidoc-writers-guide/#a-new-perspective-on-tables).

tnguyen-rh:
@kargakis @rhcarvalho @bparees 
I've just added labels.  Could you please verify they are correct?

dev_guide/deployments.adoc
bbguimaraes: @kargakis: my scenario was a DeploymentConfig using an image external to
openshift (think docker.io/nginx or whatever) with a ConfigChangeTrigger and
(for a reason that's not relevant to the discussion) an ImageChangeTrigger.
The problem is the deployment is not started automatically when the DC is
created, since no imagestream exists.

I don't think this case has any practical application, as it is fairly
artificial.  Still, I was surprised adding a ICT to a DC would prevent the CCT
from working.

I believe the relevant parts of the code are:

https://github.com/openshift/origin/blob/6e83535e99fad9420ea8979e97701700905ba9ed/pkg/deploy/controller/configchange/controller.go#L50
https://github.com/openshift/origin/blob/6e83535e99fad9420ea8979e97701700905ba9ed/pkg/deploy/controller/configchange/controller.go#L92

kargakis: Why do you need the ICT since the image is external? Where that ICT is pointing too? Are you running more than one containers in your DC?

bbguimaraes: I don't need the ICT in this case.  It points to an imagestream that is never
updated.  I'm not doing any builds since I know for certain that the images
will already be on the node.  I also have the `IfNotPresent` pull policy
because of that.  As I said, not a common setup.

The thing is: everything works fine if I have only the CC trigger.  It seems
that the current situation is: if I have an ICT I am forced to populate the IS
it points to, even if the ICT is not used and all I need is the CCT.  From what
I understand, this "forced dependency" that the ICT introduces on the CCT is a
side-effect of our race-prevention code.  I still think that's surprising.

kargakis: Actually this behavior has always been desirable. The controllers should wait on the image the ICT points to, so that we won't try to deploy things that don't exist. ICTs are a convenient way of notifying for image updates in the cluster. If you have no use for it, you shouldn't have it on your DC.

dev_guide/deployments.adoc
kargakis: > end this sentence with something like "if a configchangetrigger exists."

Since we restored the old behavior, this is no longer true. You can have automatic deployments just with an image change trigger.

kargakis:
Since we restored to the old behavior there is no need for the table. I have added a note instead about ImageChange triggers. PTAL

kargakis:
@tnguyen-rh the labels are fine. We have been supporting this behavior since I can remember.

tnguyen-rh:
Rev-hist in followup PR.  Merging.

PR #2156, opened 2016-05-26
Remove contradictory note about NFS in 3.0 docs

codificat:
This is the same change as 158f5b4 (PR #1364) but for the 3.0 branch. Fixes BZ 1284695 (for the 3.0 docs).

It seems that when that change was made the persistent storage for NFS had moved to a different section, and the change missed that fact and did not update the previous location.

The reason to have this fix in the first place is to avoid confusion. Having the latest version show one thing and an earlier version show a different one is still causing confusion.


tnguyen-rh:
@codificat Your analysis is correct.  Thanks for catching this.

PR #2172, opened 2016-05-31
fixed various links and added anchors

vikram-redhat:
I fixed some links to xrefs, and added some missing anchor tags. Some of this has already made into the Enterprise 3.2 branch, so when cherry picking, it is likely to cause some merge issues.

Remember, xrefs --> links. html --> adoc.


adellape:
@vikram-redhat The xrefs don't work on docs.openshift. Fixing the ones in deployer_router https://github.com/openshift/openshift-docs/pull/2252 because a missing # was breaking the Portal build too.

PR #2178, opened 2016-05-31
Add NetworkManager as a dependency

sdodson:
NetworkManager is required as of Origin 1.2 / OSE 3.2 in order to provide the dnsmasq functionality. The installer will fail when NetworkManager is not installed.

### DIFF
--- a/install_config/install/prerequisites.adoc
+++ b/install_config/install/prerequisites.adoc
@@ -95,6 +95,7 @@ ifdef::openshift-enterprise[]
 - Base OS: RHEL 7.1 or later with "Minimal" installation option, or RHEL Atomic
 Host 7.2.4 or later.
 endif::[]
+- NetworkManager 1.0 or later
 - 1 vCPU.
 - Minimum 8 GB RAM.
 - Minimum 15 GB hard disk space for the file system containing *_/var/_*.


jawnsy:
@sdodson How come NetworkManager is needed for this? dnsmasq only needs /etc/resolv.conf to exist, as far as I know, so systemd-networkd & systemd-resolved should be an appropriate alternative to NetworkManager, particularly for headless servers.

jawnsy:
Oh, I see, you explained:

> A NetworkManager dispatcher script runs dnsmasq and re-configures the
> host's resolv.conf to use dnsmasq. The node also sets pods to use
> dnsmasq too. dnsmasq forwards 'cluster.local' requests to the masters
> and other requests to the node's normal resolvers. The dnsmasq config
> generated by the dispatcher is located in
> /etc/dnsmasq.d/origin-*.conf.

Can we include the rationale in the docs as well, at least as a comment visible in the source and not output in the docs? It's generally good to know why we require things, so that we can remove those requirements from the docs if the situation changes (i.e. avoid cargo-cult installation of dependencies that may no longer be needed)

And if it's a hard dependency, shouldn't the RPMs be updated to reflect that?

sdodson:
We already specify that we require a minimal install of RHEL7 and NetworkManager is included in a minimal install. If someone has uninstalled NetworkManager I want them to be aware of the need to install it and deal with the ramifications of that change on their own, I don't want to be responsible for installing it automatically and potentially altering their network configuration.

vikram-redhat:
Setting this for next release as this doesn't require tech or peer reviews. @adellape

adellape:
[rev_history]
|Added NetworkManager as a requirement for nodes in the link:../install_config/install/prerequisites.html#system-requirements[System Requirements] section.
%

PR #2196, opened 2016-06-02
fix numbering

aweiteka:


tnguyen-rh:
@aweiteka Good catch, thanks.

tnguyen-rh:
[rev_history]
|xref:../admin_guide/service_accounts.adoc#[Configuring Service Accounts]
|Fixed callout numbering in the xref:../admin_guide/service_accounts.adoc#managed-service-accounts[Managed Service Accounts] example.

PR #2199, opened 2016-06-02
bz1332510 - unable to find namespaces for router

pecameron:
Added documentation to explain observed behavior.

Signed-off-by: Phil Cameron pcameron@redhat.com

### DIFF
--- a/architecture/core_concepts/routes.adoc
+++ b/architecture/core_concepts/routes.adoc
@@ -13,19 +13,26 @@ toc::[]
 
 == Overview
 
-An {product-title} route is a way to expose a
-xref:pods_and_services.adoc#services[service] by giving it an
-externally-reachable hostname like `www.example.com`.
+An {product-title} xref:../../architecture/core_concepts/routes.adoc#[route] exposes a
+xref:../../architecture/core_concepts/pods_and_services.adoc#services[service] at a
+host name, like _www.example.com_, so that external clients can reach it by
+name.
 
-A defined route and the endpoints identified by its service can be consumed by a
-router to provide named connectivity that allows external clients to reach your
-applications. Each route consists of a route name (limited to 63 characters),
-service selector, and (optionally) security configuration.
+DNS resolution for a host name is handled separately from routing.
+Your administrator may have configured a
+xref:../../install_config/install/prerequisites.adoc#prereq-dns[DNS wildcard entry]
+that will resolve to the {product-title} node that is running the
+{product-title} router. If you are using a different host name you may
+need to modify its DNS records independently to resolve to the node that
+is running the router.
+
+Each route consists of a route name (limited to 63 characters), service selector,
+and (optionally) security configuration.
 
 [[routers]]
 == Routers
 
-An {product-title} administrator can deploy _routers_ in an {product-title}
+An {product-title} administrator can deploy _routers_ to nodes in an {product-title}
 cluster, which enable routes
 xref:../../dev_guide/routes.adoc#creating-routes[created by developers] to be
 used by external clients. The routing layer in {product-title} is pluggable, and
@@ -40,15 +47,54 @@ Configuration] guide for information on deploying a router.
 ====
 endif::[]
 
-{product-title} routers provide external host name mapping and load balancing to
-xref:pods_and_services.adoc#services[services] over protocols that pass
-distinguishing information directly to the router; the host name must be present
-in the protocol in order for the router to determine where to send it.
+A router uses the service selector to find the
+xref:pods_and_services.adoc#services[service] and the endpoints backing
+the service. Service-provided load balancing is bypassed and replaced with the
+router's own load balancing. Routers watch the cluster API and automatically
+update their own configuration according to any relevant changes to the IP
+addresses of pods that provide the service. Custom routers can be deployed to
+communicate modifications of API objects to an external routing solution.
 
-Router plug-ins assume they can bind to host ports 80 and 443. This is to allow
-external traffic to route to the host and subsequently through the router.
-Routers also assume that networking is configured such that it can access all
-pods in the cluster.
+In order to reach a router in the first place, requests for host names must
+resolve via DNS to a router or set of routers. The suggested method is to define
+a cloud domain with a wildcard DNS entry pointing to one or more virtual IP (VIP)
+addresses backed by multiple router instances on designated nodes.
+
+A single VIP along with multiple routers means that there will be just one
+active router and the other router(s) are hot-standby(s) - aka passive.
+With > 1 VIPs, more than one router could be in the mix and so one could have
+an "active-active" config (or even more complex setups ala "active-active-passive").
+
+Routes can be
+xref:../../architecture/core_concepts/routes.adoc#router-sharding[sharded]
+among the set of routers. Administrators can set up sharding on a cluster
+wide basis and users can set up sharding for the namespace in their project.
+Sharding allows the operator to define multiple router groups. Each router
+in the group serves only a subset of traffic.
+
+Router VIP configuration is described in
+ifdef::openshift-enterprise,openshift-origin[]
+xref:../../admin_guide/high_availability.adoc#configuring-a-highly-available-routing-service[High Availability].
+endif::[]
+ifdef::openshift-dedicated[]
+the https://docs.openshift.com/enterprise/3.1/admin_guide/high_availability.html#configuring-a-highly-available-routing-service[{product-title} Enterprise Cluster Administration documentation].
+endif::[]
+Routes using names and addresses outside the cloud domain require
+configuration of individual DNS entries.
+
+{product-title} routers provide external host name mapping and load balancing
+of xref:pods_and_services.adoc#services[service] end points over protocols that
+pass distinguishing information directly to the router; the host name
+must be present in the protocol in order for the router to determine
+where to send it.
+
+Router plug-ins assume they can bind to host ports 80 and 443, by default.  This
+allows external traffic to route to the host and subsequently through the router
+to a pod in the cluster.  Because a router binds to ports on the host node, only
+one router listening on those ports can be on each node if the router uses host
+networking (`--host-network=true`).. A router can listen on
+other ports through use of environment variables.  Cluster networking is
+configured such that all routers can access all pods in the cluster.
 
 Routers support the following protocols:
 
@@ -62,28 +108,6 @@ Routers support the following protocols:
 WebSocket traffic uses the same route conventions and supports the same TLS
 termination types as other traffic.
 ====
-
-A router uses the service selector to find the service and the endpoints backing
-the service. Service-provided load balancing is bypassed and replaced with the
-router's own load balancing. Routers watch the cluster API and automatically
-update their own configuration according to any relevant changes in the API
-objects. Routers may be containerized or virtual. Custom routers can be deployed
-to communicate modifications of API objects to an external routing solution.
-
-In order to reach a router in the first place, requests for host names must
-resolve via DNS to a router or set of routers. The suggested method is to define
-a cloud domain with a wildcard DNS entry pointing to a virtual IP backed by
-multiple router instances on designated nodes. Router
-VIP configuration is described in
-ifdef::openshift-enterprise,openshift-origin[]
-xref:../../admin_guide/high_availability.adoc#configuring-a-highly-available-routing-service[High Availability].
-endif::[]
-ifdef::openshift-dedicated[]
-the https://docs.openshift.com/enterprise/3.1/admin_guide/high_availability.adoc#configuring-a-highly-available-routing-service[OpenShift Enterprise Cluster Administration documentation].
-endif::[]
-DNS for addresses
-outside the cloud domain would need to be configured individually. Other
-approaches may be feasible.
 
 [[routes-template-routers]]
 
@@ -168,15 +192,15 @@ xref:passthrough-termination[passthrough terminated] routes matching on HTTP
 vhost and request path.
 
 The F5 router has feature parity with the xref:haproxy-template-router[HAProxy
-template router], which means it has feature parity, and then additional
-features, compared to to the *F5 BIG-IP®* support in
+template router], and it has additional
+features, compared with the *F5 BIG-IP®* support in
 ifdef::openshift-enterprise[]
 OpenShift Enterprise 2.
 endif::[]
 ifdef::openshift-origin[]
 OpenShift v2.
 endif::[]
-When comparing with the {product-title} *routing-daemon* used in earlier
+Compared with the {product-title} *routing-daemon* used in earlier
 versions, the F5 router additionally supports:
 
 - path-based routing (using policy rules),
@@ -630,7 +654,7 @@ the destination. This field is required, but only for re-encryption.
 In {product-title}, each route can have any number of
 xref:../../architecture/core_concepts/pods_and_services.adoc#labels[labels]
 in its `metadata` field.
-A router selects which subset of routes from the entire pool of routes
+A router uses selectors to select a subset of routes from the entire pool of routes
 to serve via a _selection expression_ involving those labels.
 A selection expression can also involve
 labels on the route's namespace.
@@ -649,3 +673,34 @@ For example, a single route may belong to a `SLA=high` shard
 (but not `SLA=medium` or `SLA=low` shards),
 as well as a `geo=west` shard
 (but not a `geo=east` shard).
+
+For example, this allows the operator to partition the routes (via namespace/selectors)
+to be served by different router instances. And we could easily partition the
+routes in such a manner that the routers service overlapping sets. For example,
+deploying routers that service
+namespaces [A-J] *, [K-T] * and the last one is overlapping with the previous
+[Q-Z] *. Traditional sharding/distribution in such a case becomes a special
+case of partitioning with no overlapping sets.
+
+When routers are sharded, a given route is bound to zero or more routers in the
+group.  The route binding ensures uniqueness of the route across the shard.
+Uniqueness allows secure and non-secure versions of the same route to exist
+within a single shard.  This implies that routes now have a visible life cycle
+that moves from created to bound to active.
+
+In the sharded environment the first route to hit the shard reserves the right to
+exist there indefinitely, even across restarts.
+
+During a green/blue deployment a route may be be selected in multiple routers.
+An {product-title} application administrator may wish to bleed traffic from one
+version of the app to another and then turn off the old version.
+
+Sharding can be done by the administrator at a cluster level and by the user
+at a project/namespace level. When namespace labels are used, the service
+account for the router must have `cluster-reader` permission to permit the
+router to access the labels in the namespace.
+
+Since the routers listen on ports 80 (http) and 443 (https) by default, the
+routers must be placed on nodes where those ports are not otherwise in use, or
+the ports must be changed using the `ROUTER_SERVICE_HTTP_PORT` and
+`ROUTER_SERVICE_HTTPS_PORT` environment variables.
--- a/install_config/install/deploy_router.adoc
+++ b/install_config/install/deploy_router.adoc
@@ -76,9 +76,20 @@ $ oc create serviceaccount router -n default
 
 To add a privileged SCC to the *router* service account in the *default* namespace:
 
+====
 ----
 $ oadm policy add-scc-to-user privileged system:serviceaccount:default:router
 ----
+====
+
+Use of labels requires `cluster-reader` permission. So if the router definition sets
+`NAMESPACE_LABELS` then you need to give it the following permission.
+
+====
+----
+$ oadm policy add-cluster-role-to-user cluster-reader system:serviceaccount:default:router
+----
+====
 endif::[]
 
 [[haproxy-router]]
@@ -284,8 +295,114 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
 ====
 
 
+[[working-with-multiple-routers]]
+=== Working With Multiple Routers
+
+An administrator can create multiple routers with the same definition
+that will all serve the same set of routes. By having different groups
+of routers with different namespace or route selectors, they can vary
+the routes that the router will serve.
+
+Multiple routers can be grouped to distribute routing load in the cluster
+and separate tenants to different routers or shards.  Each router or shard in the
+group handles routes based on the selectors in the router.  The administrator
+can create shards over the whole cluster using `ROUTE_LABELS`.  The user can
+create shards over a namespace (project) by using `NAMESPACE_LABELS`.
+
+
+[[adding-nodeselector-to-a-deployment]]
+=== Adding a Node Selector to a Deployment Configuration
+
+You can make specific routers deploy on specific nodes by labeling the nodes and
+adding a node selector to the router deployment controller.
+
+Add a label to the desired node.
+====
+----
+$ oc label node 10.254.254.28 "router=first"
+----
+====
+Add a node selector to the deployment configuration.
+====
+----
+$ oc edit dc <deploymentConfigName>
+----
+====
+Find the section “spec” under “template” and add node selector section.
+Also, add the nodeSelector.
+====
+----
+  template:
+    metadata:
+      creationTimestamp: null
+      labels:
+        router: router1
+    spec:
+      nodeSelector:
+        router: "first"
+----
+====
+When the edit is saved the deployment is updated.
+
+[[using-router-shards]]
+=== Using Router Shards
+
+Each xref:../../architecture/core_concepts/projects_and_users.adoc#projects[project has its own namespace].
+By default, a router selects routes from all the namespaces it has
+access to. The access controls are based on the service account that the
+router is run with.
+
+Using `NAMESPACE_LABELS` and/or `ROUTE_LABELS`, a router can filter out the
+namespaces and/or routes that it should service. This enables you to
+partition routes amongst multiple router deployments effectively
+distributing the set of routes.
+
+Example: A router deployment `finops-router` is run with route selector
+        `NAMESPACE_LABELS="name in (finance, ops)"`
+        and a router deployment `dev-router` is run with route selector
+        `NAMESPACE_LABELS="name=dev"`
+
+If all routes are in the 3 namespaces `finance`, `ops` or `dev`,
+then this could effectively distribute our routes across two
+router deployments.
+
+In the above scenario, sharding becomes a special case of partitioning
+with no overlapping sets. Routes are divided amongst multiple router shards.
+
+The criteria for route selection governs how the routes are distributed - it
+is possible to have routes that overlap accross multiple router deployments.
+
+Example: In addition to the `finops-router` and `dev-router` in the example
+         above, we also have an `devops-router` which is run with a route
+         selector `NAMESPACE_LABELS="name in (dev, ops)"`.
+
+The routes in namespaces `dev` or `ops` now are serviced by two different
+router deployments. This becomes a case where we have partitioned the
+routes with an overlapping set.
+
+
+In addition, this enables us to create more complex routing rules ala
+divert high priority traffic to the dedicated `finops-router` but send
+the lower priority ones to the `devops-router`.
+
+`NAMESPACE_LABELS` allows filtering the projects to service and selecting
+all the routes from those projects. But we may want to partition routes
+based on other criteria in the routes themselves. The `ROUTE_LABELS`
+selector allows you to slice-and-dice the routes themselves.
+
+Example: A router deployment `prod-router` is run with route selector
+        `ROUTE_LABELS="mydeployment=prod"`
+        and a router deployment `devtest-router` is run with route selector
+        `ROUTE_LABELS="mydeployment in (dev, test)"`
+        
+Example assumes you have all the routes you wish to serviced tagged with a
+label `"mydeployment=<tag>"`.
+
 [[creating-router-shards]]
 === Creating Router Shards
+
+Router sharding lets you select how routes are distributed among a set of
+routers. As above the routers are created on different nodes.
 
 Router sharding is
 xref:../../architecture/core_concepts/routes.adoc#router-sharding[based on labels];
@@ -498,6 +615,74 @@ $ modshard 3 \
     NAMESPACE_LABELS='frequency=weekly' \
     ROUTE_LABELS='sla=low'
 ----
+
+[[using-namespace-router-shards]]
+=== Using Namespace Router Shards
+
+Each
+xref:../../architecture/core_concepts/projects_and_users.adoc#projects[project has its own namespace].
+The routes for a project can be handled by a selected router by using
+`NAMESPACE_LABELS`.  The router is given a selector for a `NAMESPACE_LABELS`
+label and the project that wants to use the router applies the `NAMESPACE_LABELS`
+label to its namespace.
+
+Because `NAMESPACE_LABELS` labels are used, the router service account associated with the
+router must have `cluster-reader` permission. This permits the router to read the labels
+that are applied to the namespaces. Multiple routers can be deployed using the
+same service account.
+
+Add `cluster-reader` to the router service account, in this case the default route:
+====
+----
+$ oadm policy add-cluster-role-to-user cluster-reader system:serviceaccount:default:router
+----
+====
+Now create and label the router:
+====
+----
+$ oadm router ...  --service-account=router
+$ oc env dc/router NAMESPACE_LABELS="router=r1"
+----
+====
+Because the router has a selector for a namespace, the router will handle
+routes for that namespace.  So, for example:
+====
+----
+$ oc label namespace default "router=r1"
+----
+====
+Now create routes in the default namespace, and the route is
+available in the default router:
+====
+----
+$ oc create -f route1.yaml
+----
+====
+Now create a new project (namespace) and create a route, route2.
+====
+----
+$ oc new-project p1
+$ oc create -f route2.yaml
+----
+====
+And notice the route is not available in your router.
+Now label namespace p1 with "router=r1"
+====
+----
+$ oc label namespace p1 "router=r1"
+----
+====
+Which makes the route available to the router.
+
+Note that removing the label from the namespace won't have immediate effect
+(as we don't see the updates in the router), so if you redeploy/start a new
+router pod, you should see the unlabelled effects.
+====
+----
+$ oc scale dc/router --replicas=0 && oc scale dc/router --replicas=1
+----
+====
+
 
 [[customizing-the-default-routing-subdomain]]
 === Customizing the Default Routing Subdomain
@@ -853,6 +1038,7 @@ link:http://golang.org/pkg/text/template/[*golang* template] that
 generates the HAProxy configuration file from a list of routes. If you
 want a customized template router to meet your needs, you can customize
 the template file, build a new Docker image, and run a customized router.
+Alternatively you can use a xref:../../dev_guide/configmaps.adoc#[configMap].
 
 One common case for this might be implementing new features within the
 application back ends. For example, it might be desirable in a highly-available
@@ -880,7 +1066,7 @@ Save this content to a file for use as the basis of your customized template.
 
 === Using a ConfigMap to Replace the Router Configuration Template
 
-A link:../../dev_guide/configmaps.html[configMap] permits you to customize
+A xref:../../dev_guide/configmaps.adoc#[configMap] permits you to customize
 the router instance without rebuilding the router image. The
 *_haproxy-config.template_*, *_reload-haproxy_* and other scripts can be modified
 as well as creating and modifying router environment variables.
@@ -889,7 +1075,7 @@ The following example shows how to use a custom *_haproxy-config.template_* in
 your router.
 
 First get a copy of the *_haproxy-config.template_* that you want to modify as
-link:#deploying-customized-router[described above]. Modify the
+xref:#deploying-customized-router[described above]. Modify the
 *_haproxy-config.template_* as desired.
 
 Create a configmap:


tnguyen-rh:
handy link: [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1332510)

architecture/core_concepts/routes.adoc
```diff
@@ -12,22 +12,27 @@ toc::[]
...
+link:../architecture/core_concepts/pods_and_services.html#services[service] at a
+host name, like _www.example.com_, so that external clients can reach it by
+name.
 
-A defined route and the endpoints identified by its service can be consumed by a
-router to provide named connectivity that allows external clients to reach your
-applications. Each route consists of a route name (limited to 63 characters),
-service selector, and (optionally) security configuration.
+DNS resolution for a host name is handled separately from routing;
+your administrator may have configured a cloud domain that will always
```
knobunc: Cloud domain?  Do you mean a DNS wildcard entry as described here: https://docs.openshift.com/enterprise/3.2/install_config/install/prerequisites.html#prereq-dns

architecture/core_concepts/routes.adoc
```diff
@@ -37,30 +42,6 @@ Configuration] guide for information on deploying a router.
 ====
 endif::[]
 
-OpenShift routers provide external host name mapping and load balancing
-to link:pods_and_services.html#services[services] over protocols that
-pass distinguishing information directly to the router; the host name
-must be present in the protocol in order for the router to determine
-where to send it.
```
knobunc: Can you make sure this section isn't deleted?  It's useful preamble for the next paragraph that you moved.

architecture/core_concepts/routes.adoc
```diff
@@ -621,6 +622,11 @@ the destination. This field is required, but only for re-encryption.
 [[router-sharding]]
 == Router Sharding
 
+An OpenShift administrator can deploy multiple routers in an OpenShift cluster. Since the
+routers always listen on ports 80 (http) and 443 (https), the routers are placed on
```
knobunc: That's not true.  You can change the ports with the ROUTER_SERVICE_HTTP_PORT and ROUTER_SERVICE_HTTPS_PORT environment variables.  They just default to 80 and 443 respectively.

architecture/core_concepts/routes.adoc
```diff
@@ -621,6 +622,11 @@ the destination. This field is required, but only for re-encryption.
...
+nodes where the ports are available. Also, the service account for the routers must
```
knobunc: I don't like the also...  how about:

In order to have permission to read the namespace labels, the service account a sharded router is running with must have the `cluster-reader` permission.

install_config/install/deploy_router.adoc
```diff
@@ -75,9 +75,19 @@ $ oc create serviceaccount router -n default
 
 To add a privileged SCC to the *router* service account in the *default* namespace:
 
+====
 ----
 $ oadm policy add-scc-to-user privileged system:serviceaccount:default:router
 ----
+====
+
+Use of multiple name spaces requires cluster-reader premission.
```
knobunc: It's the use of labels that requires the permission.  So if the router definition sets `NAMESPACE_LABELS` then you need to give it the permission.

install_config/install/deploy_router.adoc
```diff
@@ -269,9 +279,89 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
 ====
 
 
+[[working-with-multiple-routers]]
+=== Working With Multiple Routers
+
+Multiple routers can be used to distribute routing load in the cluster. Because
```
knobunc: Hm... not sure if that is entirely the reason.  You can run multiple load-balanced routers that don't use sharding.

Security is certainly part of the picture.  We need to expand this motivation more... might be worth getting @ramr to weigh in if you can't get the information from the trello card he did the work for (you will need to search the archived cards for sharding).

install_config/install/deploy_router.adoc
```diff
@@ -269,9 +279,89 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+each router uses ports 80 (http) and 443 (https) the routers are placed on nodes
```
knobunc: "by default"

s/are/must be/

or the ports must be changed by setting....

install_config/install/deploy_router.adoc
```diff
@@ -269,9 +279,89 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+in the cluster that are not already using those ports.
+
+Each router handles a set of routes that are tagged using labels. Router sharding
+uses ROUTE_LABELS. A project's (namespace) routes can be associated with a router
```
knobunc: hm... wording is a little unclear, the (namespace) reads like it applies to routes.  Not sure how to fix it.

install_config/install/deploy_router.adoc
```diff
@@ -269,9 +279,89 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+using NAMESPACE_LABELS.
+
+
+[[using-namespace-labels]]
+=== Using NAMESPACE_LABELS
+
+Each project has its own namespace. The routes for a project can be handled by a
```
knobunc: I think we reword this to talk about namespaces more than projects.  It's worth linking to https://docs.openshift.com/enterprise/3.0/architecture/core_concepts/projects_and_users.html#projects and mentioning that a project will create a new namespace... but focus more on the namespaces than projects.

install_config/install/deploy_router.adoc
```diff
@@ -269,9 +279,89 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+selected router by using NAMESPACE_LABELS. The router is given a label and each
+project that uses the router is also given the same label.
+
+In order for this to work the router service account associated with the router
+needs cluster-reader permission.
```
knobunc: Pleas put `cluster-reader` in back-ticks to make it obvious it's a literal string.

install_config/install/deploy_router.adoc
```diff
@@ -269,9 +279,89 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+The same service account can be used in creating multiple routers.
```
knobunc: Perhaps:

... can be used by multiple routers.

install_config/install/deploy_router.adoc
```diff
@@ -269,9 +279,89 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+----
+====
+Now create and label the router:
+====
+----
+$ oadm router ...
+$ oc env dc/router NAMESPACE_LABELS="router=r1"
+----
+====
+Each project has a namespace and that namespace can be labeled with the
```
knobunc: We already talked about the relationship between namespaces and projects, let's just stick to namespaces here.

install_config/install/deploy_router.adoc
```diff
@@ -269,9 +279,89 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+And notice the route is not available in your router.
+Now label namespace p1 with "router=r1"
+====
+----
+$ oc label namespace p1 "router=r1"
+----
+====
+Which makes the route available to the router.
+
+Note that removing the label from the namespace won't have immediate effect
```
knobunc: Is that also true of adding / changing the label?

And... is this a bug?  @ramr

architecture/core_concepts/routes.adoc
```diff
@@ -621,6 +622,11 @@ the destination. This field is required, but only for re-encryption.
 [[router-sharding]]
 == Router Sharding
```
knobunc: Careful... this isn't the router sharding you talk about at the bottom.  We either need a new name, or we need to talk about both types in the overview and clarify which we mean whenever we say router sharding.

install_config/install/deploy_router.adoc
```diff
@@ -269,9 +279,89 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+
+
 [[creating-router-shards]]
 === Creating Router Shards
 
+Router sharding distributes routes among a set of routers. As above the routers are
+created on different nodes.
+
 Router sharding is
 link:../../architecture/core_concepts/routes.html#router-sharding[based on labels];
```
knobunc: I think it's worth adding a link to the overview of whatever we call the namespace label kind of separation from here.

knobunc:
@ramr PTAL

architecture/core_concepts/routes.adoc
pecameron: That was a cut and paste. I assumed it was OK. Added the link...

architecture/core_concepts/routes.adoc
pecameron: I added it back in.

architecture/core_concepts/routes.adoc
pecameron: Nice wording.

architecture/core_concepts/routes.adoc
pecameron: Changed...

install_config/install/deploy_router.adoc
pecameron: Changed

install_config/install/deploy_router.adoc
pecameron: Done.

install_config/install/deploy_router.adoc
pecameron: I revised the wording.

install_config/install/deploy_router.adoc
pecameron: OK

install_config/install/deploy_router.adoc
pecameron: OK

install_config/install/deploy_router.adoc
pecameron: OK

pecameron:
I edited the doc based on the above comments. @knobunc @ramr PTAL

architecture/core_concepts/routes.adoc
```diff
@@ -12,22 +12,28 @@ toc::[]
...
+link:../architecture/core_concepts/pods_and_services.html#services[service] at a
+host name, like _www.example.com_, so that external clients can reach it by
+name.
+
+DNS resolution for a host name is handled separately from routing;
+your administrator may have configured a
+link:../../install_config/install/prerequisites.html#prereq-dns[DNS wildcard entry]
+that will always correctly resolve to the OpenShift router, or if using an unrelated
+host name you may need to modify its DNS records independently to resolve
+to the router.
```
ramr: Took a wee bit more `parsing` power - can we break this up into 2/3 sentences ala: 
```
The DNS resolution for a route's host name is handled separately from routing.   
Your administrator may have ... that will resolve to the OpenShift router.  
If you are using a different host name, you may need to ...
```  
(parts snipped for brevity). 

There is also the last sentence that got chopped and it does have some info (maybe we can somehow keep that in):  
```  
Each route consists of a route name (limited to 63 characters), service selector,   
and (optionally) security configuration.  
```

architecture/core_concepts/routes.adoc
```diff
@@ -37,6 +43,29 @@ Configuration] guide for information on deploying a router.
 ====
 endif::[]
 
+A router uses the service selector to find the service and the endpoints backing
+the service. Service-provided load balancing is bypassed and replaced with the
+router's own load balancing. Routers watch the cluster API and automatically
+update their own configuration according to any relevant changes in the API
+objects. Routers may be containerized or virtual. Custom routers can be deployed
```
knobunc: "Routers may be containerized or virtual." -- what does that mean?

architecture/core_concepts/routes.adoc
```diff
@@ -37,6 +43,29 @@ Configuration] guide for information on deploying a router.
 ====
 endif::[]
 
+A router uses the service selector to find the service and the endpoints backing
+the service. Service-provided load balancing is bypassed and replaced with the
+router's own load balancing. Routers watch the cluster API and automatically
+update their own configuration according to any relevant changes in the API
```
knobunc: I'd replace "in the API objects" with "to the IP addresses of pods that provide the service"

architecture/core_concepts/routes.adoc
```diff
@@ -37,6 +43,29 @@ Configuration] guide for information on deploying a router.
...
+objects. Routers may be containerized or virtual. Custom routers can be deployed
+to communicate modifications of API objects to an external routing solution.
+
+In order to reach a router in the first place, requests for host names must
+resolve via DNS to a router or set of routers. The suggested method is to define
+a cloud domain with a wildcard DNS entry pointing to a virtual IP (VIP) backed by
+multiple router instances on designated nodes. Routes can be sharded among the set
```
knobunc: I wouldn't talk about sharding yet, since the term hasn't been defined.

architecture/core_concepts/routes.adoc
```diff
@@ -37,6 +43,29 @@ Configuration] guide for information on deploying a router.
...
+of routers. Router
+VIP configuration is described in
```
knobunc: Why the weird line break?

architecture/core_concepts/routes.adoc
```diff
@@ -37,6 +43,29 @@ Configuration] guide for information on deploying a router.
...
+ifdef::openshift-enterprise,openshift-origin[]
+link:../../admin_guide/high_availability.html#configuring-a-highly-available-routing-service[High Availability].
+endif::[]
+ifdef::openshift-dedicated[]
+the https://docs.openshift.com/enterprise/3.1/admin_guide/high_availability.html#configuring-a-highly-available-routing-service[OpenShift Enterprise Cluster Administration documentation].
+endif::[]
+DNS for addresses
+outside the cloud domain would need to be configured individually. Other
+approaches may be feasible.
```
ramr: Would it be cleaner to say `If routes use host names/addresses outside the cloud domain, you will need to configure the DNS entries individually`?

The last sentence `other approaches may be feasible` leaves it hanging - can we either add more details or drop it.

architecture/core_concepts/routes.adoc
```diff
@@ -621,6 +629,30 @@ the destination. This field is required, but only for re-encryption.
...
+When routers are sharded, a given route is bound to a router in the group. The route
+binding ensures uniqueness of the route across the shard.  Uniqueness allows secure
+and non-secure versions of the same route to exist within a single shard.
+This implies that routes now have a visible life cycle that moves from created to
+bound to active.
+
+In the sharded environment the first route to hit the shard reserves the right to
+exist there indefinitely. Even across restarts.
+
+During a green/blue deployment a route may exist multiple times.  An OpenShift
```
knobunc: Yeah, but that requires two routers... one for the old, and one from the new.  And this paragraph doesn't cover it.

Of course, Rajat is working on the A/B bleedover which will do this for a cutover when the services that are provided are identical.  This case is for a situation where you want to have two different versions of an app and serve them at the same time to do some usability or feature testing.

architecture/core_concepts/routes.adoc
```diff
@@ -621,6 +629,30 @@ the destination. This field is required, but only for re-encryption.
...
+application administrator may wish to bleed traffic from one version of the app to
+another and then turn off the old version.
+
+Sharding is done at a cluster level and at a namespace level. When namespace labels
```
knobunc: I think we need to clearly define and name the two types of sharding.  And say that it "can be done" rather than "is done".

install_config/install/deploy_router.adoc
```diff
@@ -269,9 +280,63 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
 ====
 
 
+[[working-with-multiple-routers]]
+=== Working With Multiple Routers
+
+Multiple routers can be grouped to distribute routing load in the cluster and separate
+tenants to different routers.  Each router in the group shardes a set of routes that are
```
knobunc: typo - shardes

install_config/install/deploy_router.adoc
```diff
@@ -269,9 +280,63 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
 ====
 
 
+[[working-with-multiple-routers]]
+=== Working With Multiple Routers
+
+Multiple routers can be grouped to distribute routing load in the cluster and separate
```
knobunc: How about:

An administrator can create multiple routers with the same definition that will all serve the same set of routes.  By having different groups of routers with different namespace or route selectors, they can vary the routes that the router will serve.

install_config/install/deploy_router.adoc
```diff
@@ -269,9 +280,63 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+tenants to different routers.  Each router in the group shardes a set of routes that are
+tagged using labels. Router sharding over the whole cluster uses ROUTE_LABELS.
```
knobunc: I'd call this one "user-driven router sharding" vs. "administrator-driven route sharding".

install_config/install/deploy_router.adoc
```diff
@@ -269,9 +280,63 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+Sharding over a namespace (project) uses NAMESPACE_LABELS.
+
+
+Because each router, by default, uses ports 80 (http) and 443 (https) the routers must be
+placed on nodes in the cluster that are not already using those ports, or the ports must
+be changed using the ROUTER_SERVICE_HTTP_PORT and ROUTER_SERVICE_HTTPS_PORT environment
+variables. When namespace labels are used, the service account for the router must have
```
knobunc: You said that already... I think we need to pick one place and just say it clearly there.

install_config/install/deploy_router.adoc
```diff
@@ -269,9 +280,63 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+`cluster-reader` permission.
+
+[[adding-noedselector-to-a-deployment]]
```
knobunc: typo

install_config/install/deploy_router.adoc
```diff
@@ -269,9 +280,63 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+=== Adding a Node Selector to a Deployment Config
+
+Make specific routers deploy on specific nodes by labeling the nodes and
```
knobunc: Rather than "Make ..." how about "You can make specific ..."

I want to make it clear that this is a way to solve the problem from the previous section of having the routers be on different nodes.  (you might want to say that you are going to show how to do that in the next section when you talk about "placed on nodes ..." above)

install_config/install/deploy_router.adoc
```diff
@@ -269,9 +280,63 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+      nodeSelector:
+        router: "first"
+----
+====
+When the edit is saved the deployment is updated.
+
 [[creating-router-shards]]
 === Creating Router Shards
 
+Router sharding distributes routes among a set of routers. As above the routers are
```
knobunc: This wording bothers me.  It seems to imply that sharding will let the routes naturally distribute.  Whereas, what it does it let you select what routes are served by a router.

install_config/install/deploy_router.adoc
```diff
@@ -269,9 +280,63 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+created on different nodes.
```
knobunc: This "as above" doesn't consider the different port case... and I don't understand why this paragraph is even here... it doesn't seem to add anything new.

install_config/install/deploy_router.adoc
```diff
@@ -484,6 +549,72 @@ $ modshard 3 \
     ROUTE_LABELS='sla=low'
 ----
 
+[[using-namespace-shards]]
+=== Using Namespace Shards
```
knobunc: Using Namespace Router Shards?

install_config/install/deploy_router.adoc
```diff
@@ -484,6 +549,72 @@ $ modshard 3 \
...
+
+Each
+link:../../architecture/core_concepts/projects_and_users.html#projects[project has its own namespace].
+The routes for a project can be handled by a selected router by using NAMESPACE_LABELS.
+The router is given a NAMESPACE_LABELS label and each project that uses the router is
+also given the same NAMESPACE_LABELS label.
+
+Because NAMESPACE_LABELS labels are used, the router service account associated with the
+router must have `cluster-reader` permission.  Multiple routers can be deployed using the
```
knobunc: ... permission, in order for the router to read the labels that are applied to the namespaces.

install_config/install/deploy_router.adoc
```diff
@@ -484,6 +549,72 @@ $ modshard 3 \
...
+same service account.
+
+Add `cluster-reader` to the router service account, in this case the default route:
+====
+----
+$ oadm policy add-cluster-role-to-user cluster-reader system:serviceaccount:default:router
+----
+====
+Now create and label the router:
```
knobunc: We aren't labelling it... we are providing it with a label selector.  The labels are on the namespaces.

install_config/install/deploy_router.adoc
```diff
@@ -484,6 +549,72 @@ $ modshard 3 \
...
+====
+----
+$ oadm router ...  --service-account=router
+$ oc env dc/router NAMESPACE_LABELS="router=r1"
+----
+====
+Each namespace can be labeled with the same label as the router and the router will
```
knobunc: This is confusing.

If a namespace has a label that matches the label selector of the router, then the router will consider the routes inside that namespace.  For example:

architecture/core_concepts/routes.adoc
```diff
@@ -621,6 +629,30 @@ the destination. This field is required, but only for re-encryption.
 [[router-sharding]]
 == Router Sharding
 
+Sharding allows the operator to define multiple router groups. Each
+router in the group serves only a subset of traffic (sharding).
```
ramr: This is not necessarily true. What we do is really partition the routes (via namespace/selectors) to be served by different router instances. And we could easily partition the routes in such a manner that the routers service overlapping sets (ala namespaces `[A-J]*`, `[K-T]*` and the last one is overlapping with the previous `[Q-Z]*`).  Traditional sharding/distribution in such as case becomes a special case of partitioning with no overlapping sets.

architecture/core_concepts/routes.adoc
```diff
@@ -621,6 +629,30 @@ the destination. This field is required, but only for re-encryption.
...
+
+When routers are sharded, a given route is bound to a router in the group. The route
+binding ensures uniqueness of the route across the shard.  Uniqueness allows secure
+and non-secure versions of the same route to exist within a single shard.
```
ramr: See above comment.

architecture/core_concepts/routes.adoc
ramr: I agree. Document that separately from here.

install_config/install/deploy_router.adoc
```diff
@@ -75,9 +75,20 @@ $ oc create serviceaccount router -n default
 
 To add a privileged SCC to the *router* service account in the *default* namespace:
 
+====
 ----
 $ oadm policy add-scc-to-user privileged system:serviceaccount:default:router
 ----
+====
+
+Use of labels requires `cluster-reader` premission. So if the router definition sets
```
ramr: typo `permission`.

install_config/install/deploy_router.adoc
ramr: typo `noedselector` - its a link though.

install_config/install/deploy_router.adoc
```diff
@@ -269,9 +280,63 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+Make specific routers deploy on specific nodes by labeling the nodes and
+adding a node selector to the router deployment controller.
+
+Add a label to the desired node.
+====
+----
+$ oc label node 10.254.254.28 "router=first"
+----
+====
+Don't forget to restart the node when the config is changed.
```
ramr: You don't need to restart a node if you add/label a node. You can drop this.

install_config/install/deploy_router.adoc
ramr: Not on add. But modify/remove won't cause it to show up on the watcher.  That's expected `etcd` behavior. The selection from the objects from `etcd` is based on the selector value passed, so if you remove/change a label, that object will no longer be part of the selection and as a result, no notifications will be delivered to the `watcher` (the router in this case). I would call it a feature not a `bug`!!

architecture/core_concepts/routes.adoc
pecameron: I made the changes.

architecture/core_concepts/routes.adoc
pecameron: It was in the original. I think you can have a router in a container or not. How would you phrase this? For now I dropped "or virtual"

architecture/core_concepts/routes.adoc
pecameron: Done.

architecture/core_concepts/routes.adoc
pecameron: This is the overview of routing so I think its relevant. I can point to the section and introduce it better.

architecture/core_concepts/routes.adoc
pecameron: I am not seeing a weird line break in the rendered file. Could you point it out to me?

architecture/core_concepts/routes.adoc
pecameron: I like your wording. I dropped the "feasible" sentence.

architecture/core_concepts/routes.adoc
pecameron: @knobunc @ramr, This is the architecture, core concepts, routes section. If sharding doesn't go here, where does it go? I am leaving this here until we discuss it a bit more.

OK I am confused, I thought sharding permitted a selection of external names for a service to go one router and other external names to go to a different router. The external DNS points to the IP of the selected router (that handles the desired subset of names). I didn't relate this to anything like a a/b bleedover.

architecture/core_concepts/routes.adoc
pecameron: I'll drop by and we can talk about this. Clearly, I don't understand it yet.

install_config/install/deploy_router.adoc
pecameron: fixed

architecture/core_concepts/routes.adoc
knobunc: No, in the source...

install_config/install/deploy_router.adoc
pecameron: OK

install_config/install/deploy_router.adoc
pecameron: Deleted it from this section. Left it in th ecore_concepts section.

install_config/install/deploy_router.adoc
pecameron: Fixed

install_config/install/deploy_router.adoc
pecameron: OK

install_config/install/deploy_router.adoc
pecameron: Changed as you suggest.

install_config/install/deploy_router.adoc
pecameron: It bridges the above discussion into the example.

install_config/install/deploy_router.adoc
pecameron: OK

install_config/install/deploy_router.adoc
pecameron: New wording. PTAL

install_config/install/deploy_router.adoc
pecameron: new wording...

architecture/core_concepts/routes.adoc
pecameron: I went with your wording.

architecture/core_concepts/routes.adoc
pecameron: OK

install_config/install/deploy_router.adoc
pecameron: fixed

install_config/install/deploy_router.adoc
pecameron: OK

pecameron:
@knobunc @ramr , I made a bunch of changes based on your comments. You both don't want the sharding discussion in the core_concepts so we need to find a suitable home. This still doesn't gel with me. There is something missing that would make this all come together. PTAL

pecameron:
@sosiouxme @ahardin-rh @adellape @vikram-redhat @ramr @bfallonf @tnguyen-rh @tpoitras @liggitt @pweil- @arsogukpinar @kargakis @pravisankar @sdodson @knobunc 
Hi all, I am making changes to the routing discussion and I want to reach out to the people that have been contributing to these files for your thoughts on my changes. This is work in progress so there will be more changes coming so this is a good time to help get this right. Thanks in advance for your thoughts and comments.  Phil

architecture/core_concepts/routes.adoc
```diff
@@ -12,22 +12,31 @@ toc::[]
 
 == Overview
 
-An OpenShift route is a way to expose a
-link:pods_and_services.html#services[service] by giving it an
-externally-reachable hostname like `www.example.com`.
+An OpenShift link:../../architecture/core_concepts/routes.html[route] exposes a
```
tnguyen-rh: s/OpenShift/{product-title}/

tnguyen-rh: We're starting to use `xref` instead of `link`.  So this needs to change to:

```asciidoc
xref:../../architecture/core_concepts/routes.adoc#[route]
```

Note that in addition to `s/link:/xref:/`, there is also `s/html/.adoc#/` (note trailing `#`).  I will omit commenting on the other instances in the file, but generally they should undergo the same transform.

architecture/core_concepts/routes.adoc
```diff
@@ -12,22 +12,31 @@ toc::[]
...
+link:../../architecture/core_concepts/pods_and_services.html#services[service] at a
+host name, like _www.example.com_, so that external clients can reach it by
+name.
 
-A defined route and the endpoints identified by its service can be consumed by a
-router to provide named connectivity that allows external clients to reach your
-applications. Each route consists of a route name (limited to 63 characters),
-service selector, and (optionally) security configuration.
+DNS resolution for a host name is handled separately from routing.
+your administrator may have configured a
```
tnguyen-rh: s/your/Your/

architecture/core_concepts/routes.adoc
```diff
@@ -12,22 +12,31 @@ toc::[]
...
+link:../../install_config/install/prerequisites.html#prereq-dns[DNS wildcard entry]
+that will resolve to the OpenShift router. If you are using a different
```
tnguyen-rh: s/OpenShift/{product-title}/g

architecture/core_concepts/routes.adoc
```diff
@@ -37,16 +46,44 @@ Configuration] guide for information on deploying a router.
...
+Router VIP configuration is described in
+ifdef::openshift-enterprise,openshift-origin[]
+link:../../admin_guide/high_availability.html#configuring-a-highly-available-routing-service[High Availability].
+endif::[]
+ifdef::openshift-dedicated[]
+the https://docs.openshift.com/enterprise/3.1/admin_guide/high_availability.html#configuring-a-highly-available-routing-service[OpenShift Enterprise Cluster Administration documentation].
+endif::[]
+If routes use host names/addresses outside the cloud domain, you will
+need to configure the DNS entries individually.
+
```
tnguyen-rh: Suggest: "Routes using names and addresses outside the cloud domain require individual DNS entries configuration."

architecture/core_concepts/routes.adoc
```diff
@@ -37,16 +46,44 @@ Configuration] guide for information on deploying a router.
...
 pass distinguishing information directly to the router; the host name
 must be present in the protocol in order for the router to determine
 where to send it.
 
 Router plug-ins assume they can bind to host ports 80 and 443. This is to allow
 external traffic to route to the host and subsequently through the router.
-Routers also assume that networking is configured such that it can access all
-pods in the cluster.
+Sets of routers are configured onto different nodes.  Routers also assume that
+networking is configured such that it can access all pods in the cluster.
```
tnguyen-rh: s/it/they/ (to concord with 3rd-person plural subject "Routers")

install_config/install/deploy_router.adoc
```diff
@@ -269,9 +280,62 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+=== Working With Multiple Routers
+
+An administrator can create multiple routers with the same definition
+that will all serve the same set of routes. By having different groups
+of routers with different namespace or route selectors, they can vary
+the routes that the router will serve.
+
+
+Multiple routers can be grouped to distribute routing load in the cluster and separate
+tenants to different routers.  Each router in the group shards a set of routes that are
```
tnguyen-rh: My understanding is that a shard is conceptual construct based on two separate situations: a **route** having a label `foo=bar` and a **router** having a selection expression that specifies `foo=bar`.

I don't think it's accurate to use "shard" as a verb.

install_config/install/deploy_router.adoc
```diff
@@ -484,6 +548,74 @@ $ modshard 3 \
     ROUTE_LABELS='sla=low'
 ----
 
+[[using-namespace-router-shards]]
+=== Using Namespace Router Shards
+
+Each
+link:../../architecture/core_concepts/projects_and_users.html#projects[project has its own namespace].
+The routes for a project can be handled by a selected router by using
+NAMESPACE_LABELS.  The router is given a selector for a NAMESPACE_LABELS
```
tnguyen-rh: Need mark-up for these env vars, e.g.,`NAMESPACE_LABELS`.

tnguyen-rh:
@pecameron
I suggest in section **Router Sharding** moving the namespace-labeling stuff after the (original) route-labeling stuff, perhaps pushing the two variants down into subsections.

Other than that, there are some style-related changes required.  We (docs team) can probably take care of that in the followup PR, but if you're in the mood... :smile:

architecture/core_concepts/routes.adoc
pecameron: Thanks for letting me know. I went through and made the changes and verified the links work.

architecture/core_concepts/routes.adoc
pecameron: fixed

architecture/core_concepts/routes.adoc
pecameron: fixed

architecture/core_concepts/routes.adoc
pecameron: fixed

architecture/core_concepts/routes.adoc
pecameron: fixed

install_config/install/deploy_router.adoc
pecameron: OK

install_config/install/deploy_router.adoc
pecameron: Fixed.

pecameron:
@knobunc @tnguyen-rh @ramr 
Thanks for all of your comments! I have addressed them in this revision. Please take another look.
Thanks, phil

architecture/core_concepts/routes.adoc
```diff
@@ -12,41 +12,86 @@ toc::[]
...
 ====
 endif::[]
 
-OpenShift routers provide external host name mapping and load balancing
-to link:pods_and_services.html#services[services] over protocols that
+A router uses the service selector to find the service and the endpoints backing
+the service. Service-provided load balancing is bypassed and replaced with the
+router's own load balancing. Routers watch the cluster API and automatically
+update their own configuration according to any relevant changes to the IP
+addresses of pods that provide the service.  Routers may be containerized.
```
knobunc: What does "Routers may be containerized." mean?  They are always containerized.  Did you mean "customized"?

architecture/core_concepts/routes.adoc
```diff
@@ -12,41 +12,86 @@ toc::[]
...
+Custom routers can be deployed to communicate modifications of API objects to
```
knobunc: Can we indicate somehow that there are other reasons you may want to customize a router?

architecture/core_concepts/routes.adoc
```diff
@@ -12,41 +12,86 @@ toc::[]
...
+of xref:pods_and_services.adoc#services[service] end points over protocols that
 pass distinguishing information directly to the router; the host name
 must be present in the protocol in order for the router to determine
 where to send it.
 
 Router plug-ins assume they can bind to host ports 80 and 443. This is to allow
 external traffic to route to the host and subsequently through the router.
-Routers also assume that networking is configured such that it can access all
-pods in the cluster.
+Sets of routers are configured onto different nodes.  Routers also assume that
```
knobunc: "Sets of routers are configured onto different nodes." is a bit wishy-washy.  And I'm not sure what you mean by "sets of routers".  Are you trying to say that only one router can be on a node?  If so, say:

Because a router that is using host networking binds to ports on the host node, only one router can be on each node (unless the ports are changed).

architecture/core_concepts/routes.adoc
```diff
@@ -12,41 +12,86 @@ toc::[]
...
+If routes use host names/addresses outside the cloud domain, you will
+need to configure the DNS entries individually.
+
+{product-title} routers provide external host name mapping and load balancing
+of xref:pods_and_services.adoc#services[service] end points over protocols that
 pass distinguishing information directly to the router; the host name
 must be present in the protocol in order for the router to determine
 where to send it.
 
 Router plug-ins assume they can bind to host ports 80 and 443. This is to allow
```
knobunc: Router plug-ins assume they can bind to host ports 80 and 443, by default.  This can be customized with environment variables.

architecture/core_concepts/routes.adoc
```diff
@@ -621,8 +644,38 @@ the destination. This field is required, but only for re-encryption.
...
 == Router Sharding
 
+Sharding allows the operator to partition the routes (via namespace/selectors)
+to be served by different router instances. And we could easily partition the
+routes in such a manner that the routers service overlapping sets (ala
+namespaces [A-J] *, [K-T] * and the last one is overlapping with the previous
+[Q-Z] *). Traditional sharding/distribution in such as case becomes a special
+case of partitioning with no overlapping sets.
+
+When routers are sharded, a given route is bound to a router in the group. The route
```
knobunc: A route can be served by multiple routers... it depends what label selectors the router is configured with.  This paragraph seems to be saying a route is served by at most one router.  Whereas, a route can be served by zero or more routers.

architecture/core_concepts/routes.adoc
```diff
@@ -621,8 +644,38 @@ the destination. This field is required, but only for re-encryption.
...
+binding ensures uniqueness of the route across the shard.  Uniqueness allows secure
+and non-secure versions of the same route to exist within a single shard.
+This implies that routes now have a visible life cycle that moves from created to
+bound to active.
+
+In the sharded environment the first route to hit the shard reserves the right to
+exist there indefinitely. Even across restarts.
+
+During a green/blue deployment a route may exist multiple times.  An {product-title}
```
knobunc: What does "a route may exist multiple times" mean?

install_config/install/deploy_router.adoc
```diff
@@ -269,11 +280,64 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+[[working-with-multiple-routers]]
+=== Working With Multiple Routers
+
+An administrator can create multiple routers with the same definition
+that will all serve the same set of routes. By having different groups
+of routers with different namespace or route selectors, they can vary
+the routes that the router will serve.
+
+Multiple routers can be grouped to distribute routing load in the cluster
+and separate tenants to different routers.  Each router, shard, in the
```
knobunc: "or shard"

install_config/install/deploy_router.adoc
```diff
@@ -269,11 +280,64 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+group handles routes based on the selectors in the router.  The administrator
+can create shards over the whole cluster using `ROUTE_LABELS`.  The user can
+create shards over a namespace (project) by using `NAMESPACE_LABELS`.
+
+The next section explains how to do this.
```
knobunc: Remove this... feels like pointless noise.

knobunc:
Getting close!  Thanks

architecture/core_concepts/routes.adoc
pecameron: That was in the original. I deleted it.

architecture/core_concepts/routes.adoc
pecameron: What are the reasons we customize the router?

architecture/core_concepts/routes.adoc
pecameron: I like your wording

architecture/core_concepts/routes.adoc
pecameron: changed

architecture/core_concepts/routes.adoc
pecameron: I reworded this. It may need more changes.

architecture/core_concepts/routes.adoc
pecameron: I think it means in selected by multiple routers.

install_config/install/deploy_router.adoc
pecameron: Changed

pecameron:
@knobunc @tnguyen-rh @ramr  Latest comments are integrated. PTAL
thanks, phil

knobunc:
LGTM

pecameron:
@adellape @knobunc @nhr 
What more do we need to do before this is merged?

architecture/core_concepts/routes.adoc
```diff
@@ -152,28 +176,28 @@ addresses; because of the NAT configuration, the originating IP address
...
 
-The F5 router has feature parity with the link:#haproxy-template-router[HAProxy
-template router], which means it has feature parity, and then additional
-features, compared to to the *F5 BIG-IP®* support in
+The F5 router has feature parity with the xref:#haproxy-template-router[HAProxy
+template router], and it has additional
+features, compared with the *F5 BIG-IP®* support in
 ifdef::openshift-enterprise[]
-OpenShift Enterprise 2.
+{product-title} Enterprise 2.
```
adellape: Same comment re: product-title.

architecture/core_concepts/routes.adoc
```diff
@@ -12,41 +12,87 @@ toc::[]
...
+wide basis and users can set up sharding for the namespace in their project.
+Sharding allows the operator to define multiple router groups. Each router
+in the group serves only a subset of traffic.
+
+Router VIP configuration is described in
+ifdef::openshift-enterprise,openshift-origin[]
+xref:../../admin_guide/high_availability.adoc#configuring-a-highly-available-routing-service[High Availability].
+endif::[]
+ifdef::openshift-dedicated[]
+the https://docs.openshift.com/enterprise/3.1/admin_guide/high_availability.html#configuring-a-highly-available-routing-service[{product-title} Enterprise Cluster Administration documentation].
```
adellape: This `{product-title}` should actually just be "OpenShift" since it's referring to a specific distro name outside of its own. The way it is currently, it would get transformed to "OpenShift Dedicated Enterprise Cluster Administration".

architecture/core_concepts/routes.adoc
```diff
@@ -12,41 +12,87 @@ toc::[]
...
+endif::[]
+Routes using names and addresses outside the cloud domain require
+iconfiguration of individual DNS entries.
```
adellape: This supposed to be "configuration"?

architecture/core_concepts/routes.adoc
```diff
@@ -12,41 +12,87 @@ toc::[]
...
+
+
+If routes use host names/addresses outside the cloud domain, you will
+need to configure the DNS entries individually.
```
adellape: Isn't this the same thing as the previous sentence?

architecture/core_concepts/routes.adoc
```diff
@@ -152,28 +176,28 @@ addresses; because of the NAT configuration, the originating IP address
 ifdef::openshift-enterprise[]
 [NOTE]
 ====
-The F5 router plug-in is available starting in OpenShift Enterprise 3.0.2.
+The F5 router plug-in is available starting in {product-title} Enterprise 3.0.2.
```
adellape: `{product-title}` will expand to "OpenShift Enterprise" here, so change to:

```
{product-title} 3.0.2.
```

architecture/core_concepts/routes.adoc
adellape: Same comment re: product-title.

architecture/core_concepts/routes.adoc
```diff
@@ -188,15 +212,15 @@ Passthrough routes are a special case: path-based routing is technically
...
 template router; it is a technical limitation of passthrough encryption, not a
-technical limitation of OpenShift.
+technical limitation of {product-title}.
 ====
 
 *Routing Traffic to Pods Through the SDN*
 
 Because *F5 BIG-IP®* is external to the
-link:../../architecture/additional_concepts/sdn.html[OpenShift SDN], a cluster
+xref:../../architecture/additional_concepts/sdn.adoc#[{product-title} SDN], a cluster
```
adellape: This is probably a spot that could stay just "OpenShift".

architecture/core_concepts/routes.adoc
```diff
@@ -643,3 +665,34 @@ For example, a single route may belong to a `SLA=high` shard
 (but not `SLA=medium` or `SLA=low` shards),
 as well as a `geo=west` shard
 (but not a `geo=east` shard).
+
+For example, this allows the operator to partition the routes (via namespace/selectors)
+to be served by different router instances. And we could easily partition the
+routes in such a manner that the routers service overlapping sets (ala
+namespaces [A-J] *, [K-T] * and the last one is overlapping with the previous
+[Q-Z] *). Traditional sharding/distribution in such as case becomes a special
```
adellape: s/as/a/

architecture/core_concepts/routes.adoc
```diff
@@ -643,3 +665,34 @@ For example, a single route may belong to a `SLA=high` shard
...
+case of partitioning with no overlapping sets.
+
+When routers are sharded, a given route is bound to zero or more routers in the
+group.  The route binding ensures uniqueness of the route across the shard.
+Uniqueness allows secure and non-secure versions of the same route to exist
+within a single shard.  This implies that routes now have a visible life cycle
+that moves from created to bound to active.
+
+In the sharded environment the first route to hit the shard reserves the right to
+exist there indefinitely. Even across restarts.
```
adellape: s/./,/

install_config/install/deploy_router.adoc
```diff
@@ -75,9 +75,20 @@ $ oc create serviceaccount router -n default
...
 To add a privileged SCC to the *router* service account in the *default* namespace:
 
+====
 ----
 $ oadm policy add-scc-to-user privileged system:serviceaccount:default:router
 ----
+====
+
+Use of labels requires `cluster-reader` permission. So if the router definition sets
+`NAMESPACE_LABELS` then you need to give it the permission.
```
adellape: s/permission./following permission:/

install_config/install/deploy_router.adoc
```diff
@@ -283,11 +294,63 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+[[working-with-multiple-routers]]
+=== Working With Multiple Routers
+
+An administrator can create multiple routers with the same definition
+that will all serve the same set of routes. By having different groups
+of routers with different namespace or route selectors, they can vary
+the routes that the router will serve.
+
+Multiple routers can be grouped to distribute routing load in the cluster
+and separate tenants to different routers or shards.  Each router, shard, in the
```
adellape: Should "shard" be in parentheses here?

adellape: Or be "each router or shard"?

install_config/install/deploy_router.adoc
```diff
@@ -283,11 +294,63 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+group handles routes based on the selectors in the router.  The administrator
+can create shards over the whole cluster using `ROUTE_LABELS`.  The user can
+create shards over a namespace (project) by using `NAMESPACE_LABELS`.
+
+
+[[adding-nodeselector-to-a-deployment]]
+=== Adding a Node Selector to a Deployment Config
```
adellape: s/Config/Configuration/

install_config/install/deploy_router.adoc
```diff
@@ -283,11 +294,63 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+----
+====
+Add a node selector to the deployment configuration.
+====
+----
+$ oc edit dc <deploymentConfigName>
+----
+====
+Find the section “spec” under “template” and add node selector section
+and add the nodeSelector.
```
adellape: The back half of this sentence gets weird.

adellape:
@pecameron See comments inline.

Regarding the `xref` / `link` switch, we're actually getting ready to do a mass conversion across the repo for this, supposed to be happening on Monday by @vikram-redhat. So, just mentioning because the timing is a little tricky.

It may actually be best to switch this PR back to using `link` & `.adoc`, in the interest of merging, and let the conversion script on Monday handle it (we discovered some issues earlier this week about xrefs, which led to us switching back to `link` and the upcoming mass conversion, after fixing some things in our build process).

But anyway, after the inline comments are addressed, we can merge here, and a docs writer will do some follow-up edits for style, markup, any other edits, etc.

@tnguyen-rh Could you maybe do the follow-up after this is merged, since you were most recently in these topics for the sharding additions?

architecture/core_concepts/routes.adoc
pecameron: Done

architecture/core_concepts/routes.adoc
pecameron: Yes

architecture/core_concepts/routes.adoc
pecameron: Yes. I deleted the 81-82

architecture/core_concepts/routes.adoc
pecameron: Done

architecture/core_concepts/routes.adoc
pecameron: OK

architecture/core_concepts/routes.adoc
pecameron: OK

install_config/install/deploy_router.adoc
pecameron: OK

install_config/install/deploy_router.adoc
pecameron: I chose "each router or shard"

install_config/install/deploy_router.adoc
pecameron: New wording. Need both the selector and the thing being selected.

architecture/core_concepts/routes.adoc
```diff
@@ -12,41 +12,83 @@ toc::[]
...
+A router uses the service selector to find the service and the endpoints backing
+the service. Service-provided load balancing is bypassed and replaced with the
+router's own load balancing. Routers watch the cluster API and automatically
+update their own configuration according to any relevant changes to the IP
+addresses of pods that provide the service. Custom routers can be deployed to
+communicate modifications of API objects to an external routing solution.
+
+In order to reach a router in the first place, requests for host names must
+resolve via DNS to a router or set of routers. The suggested method is to define
+a cloud domain with a wildcard DNS entry pointing to a virtual IP (VIP) backed by
```
ramr: maybe `pointing to one or more virtual IP (VIP) addresses backed ...`? 

I know its implied but just for some clarity ... 
A single VIP along with multiple routers means that there will be just one active router and the 
other router(s) are hot-standby(s)  - aka passive.
With `> 1` VIPs, more than one router could be in the mix and so one could have an "active-active" config (or even more complex setups ala "active-active-passive").

pecameron: I made the change and added your discussion.

install_config/install/deploy_router.adoc
```diff
@@ -283,9 +294,61 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+
+An administrator can create multiple routers with the same definition
+that will all serve the same set of routes. By having different groups
+of routers with different namespace or route selectors, they can vary
+the routes that the router will serve.
+
+Multiple routers can be grouped to distribute routing load in the cluster
+and separate tenants to different routers or shards.  Each router or shard in the
+group handles routes based on the selectors in the router.  The administrator
+can create shards over the whole cluster using `ROUTE_LABELS`.  The user can
```
ramr: using `ROUTE_LABELS` or really even `NAMESPACE_LABELS`.

install_config/install/deploy_router.adoc
```diff
@@ -283,9 +294,61 @@ $ iptables -A INPUT -p tcp --dport 10443 -j ACCEPT
...
+create shards over a namespace (project) by using `NAMESPACE_LABELS`.
```
ramr: only using `ROUTE_LABELS` for routes in a particular namespace.

install_config/install/deploy_router.adoc
```diff
@@ -498,6 +561,74 @@ $ modshard 3 \
...
 
+[[using-namespace-router-shards]]
+=== Using Namespace Router Shards
+
+Each
+link:../../architecture/core_concepts/projects_and_users.html#projects[project has its own namespace].
+The routes for a project can be handled by a selected router by using
+`NAMESPACE_LABELS`.  The router is given a selector for a `NAMESPACE_LABELS`
+label and the project that wants to use the router applies the `NAMESPACE_LABELS`
+label to its namespace.
```
ramr: This is not very clear - does this mean sharding a single project or sharding across projects? You'd use `NAMESPACE_LABELS` to filter out routes (from multiple projects/namespaces) you want a router to service. And so for `"sharding"` across projects that works with `NAMESPACE_LABELS`. 
Within the scope of a single project however, `NAMESPACE_LABELS` makes no sense - you'd use `ROUTE_LABELS` to filter routes within that project/namespace.

install_config/install/deploy_router.adoc
pecameron: OK

install_config/install/deploy_router.adoc
pecameron: I am throughly confused by NAMESPACE_LABELS and ROUTER_LABELS, what routing problems does each address? When do use each and why? How is the customer supposed to think about using these labels? If I am an admin how do I this about this? If I am a user with a project/namespace how do I think about this? How do I work out what will happen to any given path? I thought I understood but reading your comment I don't.

Could you take a shot at wording the above introduction and this example. I can edit the changes.

ramr: Routes are `objects` that are associated with `projects/namespaces`. A router just takes the information from routes and provides a mechanism to `direct` traffic to the appropriate `endpoints` (associated with the `service`).
By default, the router `"services"` (warning: overloaded term here) all the routes in all namespaces.
You can filter this set of routes a router `"services"` with selectors.

The NAMESPACE_LABELS selector allows you to select a set of namespaces that the router should `"service"` routes for.  For example, consider a case where you have dev, staging and test all running on a single installation - and you use labels on the projects to indicate which environment a project belongs to. Assuming you have 10 projects and for the sake of this example, let's say 3 of those projects have a label `deployment:env=dev`, you can select just the dev  projects with `NAMESPACE_LABELS="deployment:env=dev"` and run a router that services only the development environment. 

The ROUTER_LABELS on the other hand allows you to filter routes across one or more namespaces. Meaning counter to the above example of having projects tagged with the environment ... let's assume instead you have 1000 routes in different projects and 42 of those routes were tagged with a label "deployment:type=life". To select just those 42 routes from multiple namespaces/projects, you can then use `ROUTER_LABELS="deployment:type=life"`.

I'll write this up and post it here. 
(Aside: Formatting content (that contains markup inside of it)  in a markup in a comment is going to be a bit painful but let's see how it goes).

ramr: Sorry for the delay - got sidetracked and then it was pain doing the markup in here.
So I just created a gist for the content.
Please see:  https://gist.github.com/ramr/586a16ede5403357a4709fd01927614a

Let me know if that clarifies things or then we can probably have a call since that might make it clearer.

tnguyen-rh:
@adellape 
ack on followup

@pecameron 
I see two paths forward wrt the gist @ramr posted:
- You incorporate it in this PR.
- You indicate where it should be incorporated in this PR (perhaps w/ an asciidoc comment along the lines of: `// add gist here`.
In both cases, i'll do style and other edits in the followup.
Which path shall we take?

install_config/install/deploy_router.adoc
pecameron: Thanks, I will add your prose.

pecameron:
I will add @ramr prose to the PR. I am away at class most of today.

bmeng:
Cannot add inline comment, but there is a typo here. https://github.com/openshift/openshift-docs/pull/2199/files#diff-7c33c875858cf1c9cf05f070744861faL452

pecameron:
@rarmr I added your prose to the install page. PTAL
@adellape, @tnguyen-rh  - when @ramr is OK with this, it is ready for merge. (I will make further changes as needed)

bmeng:
Not sure what you have changed, the typo is still there, just move one line down.

`` for `router-dhsard-<id>` is `Rolling`.``

pecameron:
@bmeng Ah! that is where it is. Thanks. I have fixed it.

pecameron:
@adellape What else needs to be done before you merge this?

architecture/core_concepts/routes.adoc
ramr: @pecameron , still got the typo - can you please fix. Thx

architecture/core_concepts/routes.adoc
ramr: @pecameron  ok - nm, saw this comment - however the line numbers 81/82 still has the same content (6 of one or half-a-dozen of the other!) ... get rid of one of them.

architecture/core_concepts/routes.adoc
```diff
@@ -39,15 +46,56 @@ Configuration] guide for information on deploying a router.
...
+{product-title} routers provide external host name mapping and load balancing
+of link:pods_and_services.html#services[service] end points over protocols that
+pass distinguishing information directly to the router; the host name
+must be present in the protocol in order for the router to determine
+where to send it.
+
+Router plug-ins assume they can bind to host ports 80 and 443, by default.  This
+allows external traffic to route to the host and subsequently through the router
+to a pod in the cluster.  Because a router binds to ports on the host node, only
+one router listening on those ports can be on each node. A router can listen on
```
ramr: So there's a caveat here - routers that run with `--host-network=false` (not host networking aka container networking mode) don't have this restriction of one listener per port as they use the container's network. 
So maybe something like: 
```  
only one router listening on those ports can be on each node if the router uses host networking  
(`--host-network=true`).   
```

pecameron: I made the change.

architecture/core_concepts/routes.adoc
```diff
@@ -644,3 +670,34 @@ For example, a single route may belong to a `SLA=high` shard
 (but not `SLA=medium` or `SLA=low` shards),
 as well as a `geo=west` shard
 (but not a `geo=east` shard).
+
+For example, this allows the operator to partition the routes (via namespace/selectors)
+to be served by different router instances. And we could easily partition the
+routes in such a manner that the routers service overlapping sets (ala
```
ramr: I know I wrote `ala` but that might possibly need some cleanup for docs -- maybe we can say 
`For example, deploying routers that service namespaces [A-J]*, [K-T]*  ... ` 
in a second second sentence.

pecameron: OK

ramr:
@pecameron , my review is still showing old content - so it still has some comments not addressed by the looks of it. Maybe you didn't push all your changes out? I will let you push and then do another review. Thx

pecameron:
On 06/27/2016 03:30 PM, Ram Ranganathan wrote:
>
> @pecameron <https://github.com/pecameron> , my review is still showing 
> old content - so it still has some comments not addressed by the looks 
> of it. Maybe you didn't push all your changes out? I will let you push 
> and then do another review. Thx
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub 
> <https://github.com/openshift/openshift-docs/pull/2199#issuecomment-228850064>, 
> or mute the thread 
> <https://github.com/notifications/unsubscribe/ANUgeoSMw43uScjf-Wwt6MRPh-UVtqyvks5qQCTmgaJpZM4Is_Df>.
>
I just pushed the changes you suggested in the previous set of emails. 
Hopefully, everything is actually there...
phil

tnguyen-rh:
@pecameron Needs a rebase.

pecameron:
@tnguyen-rh @adellape rebase done PTAL

tnguyen-rh:
@vikram-redhat I changed the milestone back as we discussed.

tnguyen-rh:
Revhistory in #2481.  Removing label `to_followup`.

PR #2203, opened 2016-06-03
removed EAP/A-MQ image docs from build_cfg

vikram-redhat:


vikram-redhat:
I haven't removed the actual topic files just the links to it from the build cfg. Keeping these topic files around just in case.

vikram-redhat:
rev history not required.

vikram-redhat:
Followed up by: https://github.com/openshift/openshift-docs/pull/2204

PR #2204, opened 2016-06-03
removed xpass images

vikram-redhat:


adellape:
@vikram-redhat For this and https://github.com/openshift/openshift-docs/pull/2203, are you planning to update the rest of the OpenShift docs that have in-repo links to these topics? Or do you expect that the redirects in place should handle things?

vikram-redhat:
@adellape Was hoping the redirects would take care of things. If not, and if you notice something odd, let me know.

PR #2207, opened 2016-06-03
fix distros for new admin solutions guide

tpoitras:
accidentally had new admin solutions guide set to origin


tpoitras:
no revhistory note required

PR #2212, opened 2016-06-03
advanced/introduction: Atomic Host cleanup

cgwalters:
Let's say "Atomic Host variant" since it exists for Fedora and CentOS
as well.  I'd like to have more of our community docs to point to
CentOS Atomic Host where previously we would just say RHEL Atomic
Host.

Also it's certainly _possible_ to run Ansible from inside Docker on an
Atomic Host system...but we probably shouldn't rathole on that early
in the docs introduction.  If Ansible upstream introduces some docs
(like
http://victorlin.me/posts/2014/11/13/provision-with-ansible-from-inside-docker
) we can link to that.

### DIFF
--- a/install_config/install/advanced_install.adoc
+++ b/install_config/install/advanced_install.adoc
@@ -17,17 +17,12 @@ installation_ method for installing OpenShift hosts. Familiarity with Ansible is
 assumed, however you can use this configuration as a reference to create your
 own implementation using the configuration management tool of your choosing.
 
-While RHEL Atomic Host is supported for running containerized OpenShift
-services, the advanced installation method utilizes Ansible, which is not
-available in RHEL Atomic Host, and must therefore be run from
-ifdef::openshift-enterprise[]
-a RHEL 7 system.
-endif::[]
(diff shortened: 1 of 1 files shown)


jberkus:
@cgwalters I'm working on the openshift-ansible container deploy.  I agree that we shouldn't recommend it until it actually works, though.

ahardin-rh:
@cgwalters @JacobTanenbaum Resurrecting this. Should this be closed? Thanks!

jberkus:
I'm +1 on the doc changes.  We should add them.

cgwalters:
Of course since this we've shipped package layering for AH, which does work to install Ansible.  Though personally I tend to run it from a container.

ahardin-rh:
Looks like this need a rebase.  Thanks!

ahardin-rh:
@cgwalters Trying to resurrect this again. :ghost: Can you please rebase this, or close if the change is no longer needed? Thanks!

ahardin-rh:
Closing this due to inactivity, but please let me know if we should reopen.

PR #2224, opened 2016-06-07
BZ1343292: Changed plugin to plug-in, where applicable

tpoitras:
https://bugzilla.redhat.com/show_bug.cgi?id=1343292

Still had some pesky uses of "plugin" laying around. There were plenty I couldn't change, due to them being in file names, env var names, lines of YAML, etc. But I changed the ones I could.

@bfallonf - mind taking a squizz here? Thanks! :octopus:

### DIFF
--- a/admin_guide/sdn_troubleshooting.adoc
+++ b/admin_guide/sdn_troubleshooting.adoc
@@ -355,8 +355,8 @@ These are the interfaces that the OpenShift SDN creates:
 
 * `br0`: The OVS bridge device that containers will be attached to.
 OpenShift SDN also configures a set of non-subnet-specific flow rules
-on this bridge. (The `multitenant` plugin does this immediately; the
-`ovssubnet` plugin waits until the SDN master announces the creation
+on this bridge. (The `multitenant` plug-in does this immediately; the
+`ovssubnet` plug-in waits until the SDN master announces the creation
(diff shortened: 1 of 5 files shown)


bfallonf:
@tpoitras Looks good to me :fish:

adellape:
@tpoitras This commit was picked to the 3.1 branches, but I'm going to revert them on those branches because the API topics changed a lot between 3.1 and 3.2. If you want to make the plug-in fixes for the 3.1 REST API topics, you'll probably need to go make those specifically on those branches. cc @vikram-redhat

adellape:
Reverting the commit will unfortunately undo the legitimate changes in other topics tho. So basically at this point I think you'll need to go make these changes directly on the 3.1 branches for all the topics. Sorry.

tpoitras:
@adellape - hey thanks for the heads-up on that one, sorry I seem to have missed this notification earlier.

@vikram-redhat - do you want me to make the plugin -> plug-in changes directly into the 3.1 branch?

vikram-redhat:
@tpoitras - yeah. If I understand correctly, all the other branches have this change now and only 3.1 doesn't because of the other commits that got pulled in. So, yes, at this stage, I think the best thing to do is to make these changes directly in the 3.1 branch.

PR #2226, opened 2016-06-07
better uid permissions example

bparees:


bparees:
@danmcp ptal

creating_images/guidelines.adoc
```diff
@@ -273,9 +273,23 @@ that a process which is able to escape the container due to a vulnerability in
 the container framework will not have specific user permissions on the host
 system.
 
-To accomplish this, directories that must be written to by processes in the
-image should be world-writable. In addition, the processes running in the
-container must not listen on privileged ports (ports below 1024).
+To support running as an arbitray user, directories and files that must be written to by
```
danmcp: I don't think this is the only option to do this.  So maybe change the wording to list it as an option.

https://docs.openshift.com/enterprise/3.2/install_config/persistent_storage/pod_security_context.html#supplemental-groups
https://docs.openshift.com/enterprise/3.2/install_config/persistent_storage/pod_security_context.html#fsgroup

bparees: @danmcp those examples are for persistent storage permissions, not permissions of files within the image, pretty sure it's unrelated.

but i'll update this to make it clearer it's referring to how you define files within the image during image construction.

danmcp: Ah sorry.  Was confused.

bparees:
@danmcp i took out the reference to the volume permissions, i never liked that paragraph anyway and it seems less relevant now that we have fsgroup, as you note.

bparees:

creating_images/guidelines.adoc
```diff
@@ -261,21 +261,29 @@ link:s2i.html[S2I Requirements] topic.
...
-that a process which is able to escape the container due to a vulnerability in
-the container framework will not have specific user permissions on the host
-system.
-
-To accomplish this, directories that must be written to by processes in the
-image should be world-writable. In addition, the processes running in the
-container must not listen on privileged ports (ports below 1024).
+To provide additional security against processes escaping the container due
+to a container engine vulnerability and thereby achieving escalated permissions 
+on the host node, OpenShift runs containers using an arbitrarily assigned uid by default.
```
tnguyen-rh: s/OpenShift/{product-title}/

creating_images/guidelines.adoc
```diff
@@ -261,21 +261,29 @@ link:s2i.html[S2I Requirements] topic.
...
+
+For an image to support running as an arbitray user, directories and files that may be 
+written to by processes in the image should be owned by the root group and read/writable 
+by that group.  Files to be executed should also have group execute permissions.  The following 
+commands will recursively set the correct permissions starting in the current directory:
```
tnguyen-rh: Suggest: "To set the correct permissions in the current and all child directories:"

creating_images/guidelines.adoc
```diff
@@ -261,21 +261,29 @@ link:s2i.html[S2I Requirements] topic.
...
+
+----
+chgrp -R 0 .
+chmod -R g+rw .
+find . -type d -exec chmod g+x {} +
+----
```
tnguyen-rh: Need prompts.

creating_images/guidelines.adoc
```diff
@@ -261,21 +261,29 @@ link:s2i.html[S2I Requirements] topic.
...
+
+This should be done as part of the image construction.
+
+The container user is always a member of the root group, so this will give the container user
+access to read/write the files. The root group does not have any special permissions (unlike the root
```
tnguyen-rh: Suggest: "As a member of the root group, the container user can read and write files."

tnguyen-rh:
@bparees A few suggestions and refinements.  Otherwise, LGTM.

creating_images/guidelines.adoc
bparees: this is a bit tricky since you wouldn't actually run these commands at a prompt, you'd do them in a script that is executed in your Dockerfile, like:
RUN fix-permissions.sh

as seen here:
https://github.com/openshift/postgresql/blob/master/9.2/root/usr/libexec/fix-permissions
https://github.com/openshift/postgresql/blob/master/9.2/Dockerfile#L40

do you have another recommendation?

creating_images/guidelines.adoc
bparees: partially accepted.  I went with "Because the container user is always a member of the root group, the container user can read and write these files."

I want to be very clear about the container user being a member of the root group, and i don't think just saying "as a member of the root group, the container user..." conveys the fact that we make the container user a member of that group.

creating_images/guidelines.adoc
tnguyen-rh: s/commands/Dockerfile fragment/

then prefix all lines w/ "RUN "
then add a blurb explaining the (noninteractive) Dockerfile context.

bparees:
@tnguyen-rh next iteration pushed, ptal... thanks!

creating_images/guidelines.adoc
```diff
@@ -261,21 +261,28 @@ link:s2i.html[S2I Requirements] topic.
...
-image should be world-writable. In addition, the processes running in the
-container must not listen on privileged ports (ports below 1024).
+To provide additional security against processes escaping the container due
+to a container engine vulnerability and thereby achieving escalated permissions 
+on the host node, {product-title} runs containers using an arbitrarily assigned uid by default.
+
+For an image to support running as an arbitray user, directories and files that may be 
+written to by processes in the image should be owned by the root group and read/writable 
+by that group.  Files to be executed should also have group execute permissions.  The following
+Dockerfile fragment will set permissions for `/some/directory` and all child directories:
```
tnguyen-rh: We try to avoid "this WILL do that".  Suggest: "The following Dockerfile fragment sets permissions..."

tnguyen-rh: Also, filenames are marked up:

```asciidoc
*_/etc/hosts_*
```

That is: STAR UNDERSCORE FILENAME UNDERSCORE STAR.

creating_images/guidelines.adoc
```diff
@@ -261,21 +261,28 @@ link:s2i.html[S2I Requirements] topic.
...
+
+----
+RUN chgrp -R 0 /some/directory
+RUN chmod -R g+rw /some/directory
+RUN find /some/directory -type d -exec chmod g+x {} +
+----
+
+Adding these commands to your Dockerfile will set the directory and file permissions to allow users
```
tnguyen-rh: s/will set/sets/

creating_images/guidelines.adoc
```diff
@@ -261,21 +261,28 @@ link:s2i.html[S2I Requirements] topic.
...
+in the root group to access them in the built image.  Because the container user is always a member 
+of the root group, the container user can read and write these files.  The root group does not have 
+any special permissions (unlike the root user) so there are no security concerns with this arrangement.
+
+In addition, the processes running in the container must not listen on privileged ports (ports below 
+1024), since it is not running as a privileged user.
```
tnguyen-rh: s/it is/they are/

tnguyen-rh:
@bparees Couple of refinements.  Getting close!

bparees:
@tnguyen-rh thanks, latest comments addressed.

bparees:
@tnguyen-rh bump

install_config/persistent_storage/pod_security_context.adoc
```diff
@@ -126,7 +126,7 @@ It may not be immediately obvious which SCC was matched by a pod, so the command
 above can be very useful in understanding the UID, supplemental groups, and
 SELinux relabeling in a live container.
 
-Any SCC with a strategy set to *RunAsAny* allows arbitrary values for that
+Any SCC with a strategy set to *RunAsAny* allows specific values for that
```
tnguyen-rh: I don't understand this change.  Is it intentional?  To me, "run as any" and "arbitrary" sound like a better match than "run as any" and "specific".

bparees: yes it's intentional.

RunAsAny means the user can specify the uid, so saying it allows "specific" values makes more sense to me than "arbitrary".  Because elsewhere we use arbitray when saying "openshift runs your image w/ an arbitrary uid".  So i'm trying to straighten out the usage:

arbitrary - a randomly assigned value the user has no control over
specific - a value chosen by the user.

I think it makes sense if you read the full sentence:
"allows specific values for that strategy to be defined in the pod definition"  - the user can specify the value in the pod definition.

"allows arbitrary values for that strategy to be defined....." - this implies to me that i have no control over the value.

bparees:
@tnguyen-rh bump

bparees:
@tnguyen-rh bump

adellape:
LGTM.

@bparees Put this back into OSE 3.1 as well?

adellape:
[rev_history]
|link:../creating_images/guidelines.html[Guidelines]
|Added a better better UID permissions example for *Support Arbitrary User IDs* in the link:../creating_images/guidelines.html#openshift-specific-guidelines[OpenShift-Specific Guidelines] section.
%

bparees:
Sure.

Ben Parees | OpenShift

PR #2232, opened 2016-06-08
Bug 1337328 : Added text mongodb variable

bfallonf:
@bparees @mfojtik As per email.

I've pretty much just added the variable to the table with a link to the text section of the Mongo docs. I don't think this warrants its own section explaining it. Let me know if you disagree.

Any other thoughts?

### DIFF
--- a/using_images/db_images/mongodb.adoc
+++ b/using_images/db_images/mongodb.adoc
@@ -162,15 +162,19 @@ MongoDB settings can be configured with the following environment variables:
 
 |`*MONGODB_NOPREALLOC*`
 |Disable data file preallocation.
-|true
+|`*true*`
 
 |`*MONGODB_SMALLFILES*`
 |Set MongoDB to use a smaller default data file size.
-|true
+|`*true*`
 
 |`*MONGODB_QUIET*`
 |Runs MongoDB in a quiet mode that attempts to limit the amount of output.
-|true
+|`*true*`
+
+|`*MONGODB_TEXT_SEARCH_ENABLED*`
+|(MongoDB version 2.4 only) Enables the https://docs.mongodb.org/v2.6/core/index-text/#text-search-text-command[text search] feature.
+|`*false*`
 |===
 
 === Volume Mount Points


mfojtik:
lgtm

bparees:
needs a note that it's only applicable to mongodb 2.4 (mongodb 2.6 doesn't have an equivalent setting)

bfallonf:
@mfojtik @bparees Cool. Thanks! I added a note in parenthesis.

If there's nothing else, @tpoitras @adellape @ahardin-rh @tnguyen-rh could probably have a read/merge/both.

bparees:
lgtm

using_images/db_images/mongodb.adoc
```diff
@@ -162,15 +162,19 @@ MongoDB settings can be configured with the following environment variables:
 
 |`*MONGODB_NOPREALLOC*`
 |Disable data file preallocation.
-|true
+|True
```
adellape: @bparees Do you know whether these are case sensitive? It looks like a spec will accept it init-capped, at least.

bparees: i don't know, we stuff the value in the mongo config here:
https://github.com/openshift/mongodb/blob/master/2.4/root/usr/share/container-scripts/mongodb/mongodb.conf.template

so it's up to mongo.  i'd guess mongo is not case sensitive.

adellape: Thanks

adellape:
@bfallonf LGTM. Up to you on the inline case sensitivity thing.

bfallonf:
@adellape I changed to caps because it wasn't really about what's in the file, but it's the value. I can't imagine people are going to copy/paste one word...

But I changed to being parameter markup and pushed again.

Nothing of note changes. I'll merge away.

bfallonf:
[rev_history]
|xref:../using_images/db_images/mongodb.adoc#[Database Images -> MongoDB]
|Added the `*MONGODB_TEXT_SEARCH_ENABLED*` variable to the xref:../using_images/db_images/mongodb.adoc#environment-variables[Environment Variables] section.
%

PR #2235, opened 2016-06-08
[WIP] Bug 1296199 - Add section "Comprehensive List of Events"

tnguyen-rh:
Fixes bug 1296199 ([link](https://bugzilla.redhat.com/show_bug.cgi?id=1296199)).

@pweil-
PTAL at the TODO comments (line 52).  I'd like to either validate that approach, or request guidance towards a better one (including other programs and files to look at).  Thanks!

### DIFF
--- a/dev_guide/events.adoc
+++ b/dev_guide/events.adoc
@@ -45,3 +45,172 @@ This section will be expanded as more information becomes available
 
 // TODO: out-of-band example
 ////
+
+[[events-reference]]
+== Comprehensive List of Events
+
+// TODO:
+// - Consult various source code; create tables.
+//   - Kubernetes
+//     pkg/kubelet/container/event.go
+//     [add here.]
+//   - OpenShift
+//     [add here.]
+//   - ??? etcd
+// - Reorganize tables.
+
+This section describes the events of {product-title}
+in a series of tables.
+The first table is for _system events_,
+those not associated with any particular API object.
+The rest of the tables group the events according to API object.
+
+The *Name* is a short string that uniquely identifies the event.
+A trailing `*` means that the event includes timestamp information.
+
+.System Events
+[cols="2,8",options="header"]
+|===
+| Name | Description
+
+|SystemOOM *
+|There was an OOM (out of memory) situation on the cluster.
+|===
+
+.Container Events
+[cols="2,8",options="header"]
+|===
+| Name | Description
+
+|Created
+|
+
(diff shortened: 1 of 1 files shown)


dev_guide/events.adoc
```diff
@@ -45,3 +45,172 @@ This section will be expanded as more information becomes available
...
 // TODO: out-of-band example
 ////
+
+[[events-reference]]
+== Comprehensive List of Events
+
+// TODO:
+// - Consult various source code; create tables.
+//   - Kubernetes
+//     pkg/kubelet/container/event.go
```
pweil-: This is a good place to start however it is not a comprehensive list since not all events are using these constants.  In order to find all events I would look for usages of `Event`, `Eventf`, and `PastEventf` methods that are defined in pkg/client/record/event.go

cc @ncdc

ncdc: This is going to be a pain to keep up to date. From what I understand, events are not guaranteed to remain backwards compatible, so they could change at any time.

ahardin-rh:
@ncdc @pweil- I am picking up this PR from where @tnguyen-rh left off, as he is no longer with the company. This work is verified, but your comments above suggest that maintaining this list will be a challenge. Should we publish what is here, then build off as needed in a separate PR, or do you advise against this? Thanks!

pweil-:
While I think the information is useful it will definitely need reviewed every rebase and every release for accuracy.  I guess my question here is what will providing the events give to an end user or administrator?  Is the thinking that it would help someone write a monitoring tool or is it just for informational purposes?

ahardin-rh:
@pweil- Thanks. After discussion with @adellape, we are going to opt to include this information and create a routine release task for our team to update the list as necessary. I am not sure where exactly to source descriptions from. I see a pointer to [event.go](https://github.com/pweil-/kubernetes/blob/5f9ee58de071515a4c8b439c9ccd048dfd728c85/pkg/client/record/event.go#L54-54) in the [BZ ](https://bugzilla.redhat.com/show_bug.cgi?id=1296199)(comment 7), but not sure how to source descriptions from that. Any guidance? Thanks!

pweil-:
The only way to reliably find the reasons for the event would be looking for calls to those fields in the code which is what will probably make this tedious :cry:

ahardin-rh:
Closing in favor of https://github.com/openshift/openshift-docs/pull/2756

PR #2242, opened 2016-06-09
Edit section on daemonsets

bfallonf:
@pweil- As discussed.

As it's still a WIP, and it's the only thing that's supported with Daemonsets, I can't think of a better place to put it. So I just gave your PR some more context.

However, is there any reason why my 3.2 install does not have the --daemonset option for the oadm registry command? Am I missing something?

### DIFF
--- a/install_config/install/docker_registry.adoc
+++ b/install_config/install/docker_registry.adoc
@@ -80,19 +80,20 @@ endif::[]
 
 === Deploying the Registry as a DaemonSet
 
-You may use the `oadm registry` command to deploy the registry as a
-link:http://kubernetes.io/docs/admin/daemons/[DaemonSet] with
-the `--daemonset` option.
+Use the `oadm registry` command to deploy the registry as a DaemonSet with the
+`--daemonset` option.
 
-A link:http://kubernetes.io/docs/admin/daemons/[DaemonSet] will run a pod
-on every node.  For more information on
-link:http://kubernetes.io/docs/admin/daemons/[DaemonSets] please see the
+Daemonsets ensure that when nodes are created, they contain copies of a
+specified pod. When the nodes are removed, the pods are garbage collected.
+
+For more information on DaemonSets, see the
 link:http://kubernetes.io/docs/admin/daemons/[DaemonSet documentation].
 
 [NOTE]
 ====
-Updating of DaemonSets is still a work in progress.
-It is recommended that a DeploymentConfig be used to deploy the registry.
+Using DaemonSets with {product-title} is a work in progress, and the only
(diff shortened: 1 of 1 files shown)


pweil-:
> is there any reason why my 3.2 install does not have the ...

I was mistaken in my email convo with you.  The related PR was only merged a month ago so it wouldn't have made the 3.2 cut.  Apologies for that

install_config/install/docker_registry.adoc
```diff
@@ -80,19 +80,20 @@ endif::[]
...
+For more information on DaemonSets, see the
 link:http://kubernetes.io/docs/admin/daemons/[DaemonSet documentation].
 
 [NOTE]
 ====
-Updating of DaemonSets is still a work in progress.
-It is recommended that a DeploymentConfig be used to deploy the registry.
+Currently, using DaemonSets with {product-title} is a work in progress, and the
+only supported process is creating a registry. Using deployment configurations
+to deploy a registry might be ideal.
```
pweil-: I think I'd make this a little more scary and recommend using the DC.  We don't want to recommend running as a DaemonSet until the upstream work is completed.

bfallonf:
@pweil- If this didn't make the 3.2 cut, when does that mean it will be released? Maybe it'd be a better idea to hold of on documenting this then if such things as the logging use case, as discussed in the BZ, are not too valid ATM.

Thoughts?

pweil-:
> If this didn't make the 3.2 cut, when does that mean it will be released

The command flag is merged for 3.3.  My stance on this is that anywhere we document it needs to be followed with a note that says it isn't recommended just yet.  I don't mind having the type documented so people know it's available for use but it comes with the burden of not having rolling updates.

bfallonf:
@pweil- OK. I think what's best is:

* Merge this so that the info is in upstream
* Label this PR as 3.3
* Keep looking at the Daemonsets work so that this stuff can be elaborated on for 3.3

I've also changed the wording of this to more recommend-ey like you suggest above.

Does that sound good to you?

pweil-:
sounds good to me, thanks

bfallonf:
@pweil- Great. I'll do that. Thanks!

adellape:
@bfallonf Does this need a revhistory?

bfallonf:
@adellape it does!

bfallonf:
[rev_history]
|xref:../install_config/install/docker_registry.adoc[Installing -> Deploying a Docker Registry]
|Added section on deploying the registry as a daemonset.
%

adellape:
@bfallonf Need a `#` anchor after the `.adoc` in the revhistory. :penguin:

PR #2254, opened 2016-06-10
link fix

ahardin-rh:
xref was broken


ahardin-rh:
no rev history needed

PR #2259, opened 2016-06-13
changed xrefs to links

vikram-redhat:


vikram-redhat:
@adellape ptal.

adellape:
@vikram-redhat @ahardin-rh Since Harrison upgraded AsciiDoctor for d.o.c. to 1.5.4, most of the s/xref/links/ in this PR are no longer needed, and we would just in the future be switching them back to xref. 

However, there are 3 instances I found where a raw xref is still showing through on the Customer Portal builds, which are all because of the reason cited in item 9 of http://ose-docs.etherpad.corp.redhat.com/26: "For links within the same .adoc file, you shouldn't have an anchor".

So that's the 2 here in scheduler.adoc:

https://github.com/openshift/openshift-docs/pull/2259/files#diff-ac165e2e343d723d917cc003c5d90250L231

And the 1 here in http_proxies.adoc:

https://github.com/openshift/openshift-docs/pull/2259/files#diff-f41cad1a91c282f2d6cd2f01706c4ff4L173

Spoke with @ahardin-rh and she's going to make separate PRs for those and just drop the `#` instead of switching to `link`, and pick them as part of the weekly release today.

adellape:
So I think we can close this PR in favor of hers after it's merged, but will leave this PR open until then.

admin_solutions/master_node_config.adoc
```diff
@@ -138,7 +138,7 @@ https://github.com/openshift/openshift-ansible/blob/master/inventory/byo/hosts.o
 === Making Manual Configuration Changes
 
 After installing {product-title} using the
-xref:../install_config/install/quick_install.adoc#[quick install],
+link:../install_config/install/quick_install.html#[quick install],
```
adellape: Leaving a `#` hanging on a `.html` is OK, yea?

vikram-redhat: Yeah.

vikram-redhat:
As discussed, I am closing this PR in favor of creating a new one.

adellape:
Dropping milestone.

PR #2262, opened 2016-06-13
Updated instances of OpenShift to {product-title} in Getting Started

ahardin-rh:
Blanket update of name change in the Getting Started book

### DIFF
--- a/getting_started/devpreview_faq.adoc
+++ b/getting_started/devpreview_faq.adoc
@@ -21,15 +21,15 @@ during the preview period.
 == Product
 
 HOW IS THE NEXT GEN DEVELOPER PREVIEW DIFFERENT THAN OPENSHIFT ONLINE?::
-{product-title} currently runs on the OpenShift 2.X platform. The Next Gen
-Developer Preview brings OpenShift 3, built on Docker and Kubernetes, to the public 
+{product-title} currently runs on the {product-title} 2.X platform. The Next Gen
+Developer Preview brings {product-title} 3, built on Docker and Kubernetes, to the public
 cloud in a multi-tenant environment.
 link:../getting_started/online_v2_vs_v3.html[Learn more].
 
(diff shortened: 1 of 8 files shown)


ahardin-rh:
@bfallonf @adellape @tnguyen-rh PTAL. I will continue with separate PRs for the other books.

getting_started/devpreview_faq.adoc
```diff
@@ -20,16 +20,16 @@ during the preview period.
...
 == Product
 
-HOW IS THE NEXT GEN DEVELOPER PREVIEW DIFFERENT THAN OPENSHIFT ONLINE?::
-{product-title} currently runs on the OpenShift 2.X platform. The Next Gen
-Developer Preview brings OpenShift 3, built on Docker and Kubernetes, to the public 
+HOW IS THE NEXT GEN DEVELOPER PREVIEW DIFFERENT THAN {product-title} ONLINE?::
+{product-title} currently runs on the {product-title} 2.X platform. The Next Gen
+Developer Preview brings {product-title} 3, built on Docker and Kubernetes, to the public
 cloud in a multi-tenant environment.
 link:../getting_started/online_v2_vs_v3.html[Learn more].
```
bfallonf: @ahardin-rh Do we know if these in the caps are going to come out in upper case or lower case?

bfallonf:
@ahardin-rh Just one nit, but it LGTM, probably even with the nit still as is.

getting_started/devpreview_faq.adoc
adellape: I think these all-caps spots should stay as-is, for that case reason. Case aside, with this current change it would come up as "OpenShift Online Online", too, since "OpenShift Online" is the full product-title for Online.

ahardin-rh:
No rev history needed

PR #2279, opened 2016-06-14
added ifdef around Infrastructure Topological Levels section

ahardin-rh:


ahardin-rh:
@adellape PTAL before I cherry-pick to dedicated-3.2. Thanks!

PR #2282, opened 2016-06-15
Added keystone & LDAP to Authentication topic in Admin Solutions guide

tpoitras:
https://bugzilla.redhat.com/show_bug.cgi?id=1339837

Working on adding Keystone & LDAP auth to the admin solutions guide.

### DIFF
--- a/admin_solutions/authentication.adoc
+++ b/admin_solutions/authentication.adoc
@@ -1,22 +1,262 @@
-[[admin-solutions-authentication]]
-= Authentication
-{product-author}
-{product-version}
-:data-uri:
-:icons:
-:experimental:
-:toc: macro
-:toc-title:
-:prewrap!:
-
 toc::[]
 
 == Overview
 
 {product-title} supports many different authentication methods, as defined in
-xref:../install_config/configuring_authentication.adoc#install-config-configuring-authentication[Configuring Authentication].
-For now, this topic focuses on GitHub as a source of authentication, but more
-providers will be added in a future release.
+xref:../install_config/configuring_authentication.adoc#install-config-configuring-authentication[Configuring Authentication]:
+
+- xref:keystone-auth[Keystone]
+- xref:ldap-auth[LDAP]
+- xref:github-auth[GitHub]
+
+[[keystone-auth]]
+== Keystone Authentication
+
+http://docs.openstack.org/developer/keystone/[Keystone] is an OpenStack project
+that provides identity, token, catalog, and policy services. You can integrate
+your {product-title} cluster with Keystone to enable shared authentication with
+an OpenStack Keystone v3 server configured to store users in an internal
+database. Once configured, this configuration allows users to log in to
+{product-title} with their Keystone credentials.
+
+[[config-keystone-auth-on-master]]
+=== Configuring Authentication on the Master
+
+. If you have:
+.. Already completed the installation of Openshift, then copy the
+*_/etc/origin/master/master-config.yaml_* file into a new directory; for example:
++
+----
+$ cd /etc/origin/master
+$ mkdir keystoneconfig; cp master-config.yaml keystoneconfig
+----
+.. Not yet installed {product-title}, then start the {product-title} API server,
+specifying the hostname of the (future) {product-title} master and a directory
+to store the configuration file created by the start command:
++
+----
+$ openshift start master --public-master=<apiserver> --write-config=<directory>
+----
++
+For example:
++
+----
+$ openshift start master --public-master=https://myapiserver.com:8443 --write-config=keystoneconfig
+----
++
+[NOTE]
+====
+If you are installing with Ansible, then you must add the
+`*identityProvider*` configuration to the Ansible playbook.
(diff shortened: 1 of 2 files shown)


tpoitras:
Nicely rendered docs build: http://file.bne.redhat.com/~tpoitras/2016/keystoneLDAP/openshift-enterprise/keystoneLDAP-BZ1339837/admin_solutions/authentication.html

tpoitras:
Incorporated tech review feedback from @sgallagher 

Have not yet heard back from scott seago about Keystone.

admin_solutions/authentication.adoc
```diff
@@ -14,9 +14,236 @@ toc::[]
...
+.. Change `*preferredUsername*` to the attribute to use as the preferred user
+name when provisioning a user for this identity. This option can accept multiple
+attributes. If more than one is specified, they will be checked in order and the
+first non-empty attribute will be used.
+.. Change the `*ca*` to the certificate bundle to use in order to validate
+server certificates for the configured URL. If empty, system trusted roots are
+used. This setting only applies if *insecure: false*. If the LDAP server
+requires a different certificate chain, this attribute should contain the
+filesystem path of that certificate or certificate bundle.
+.. If required, modify the `*insecure*` parameter. The default is *false*. When *true*, no TLS connection is made to the server. When *false*, `ldaps://` URLs connect using TLS, and `ldap://` URLs are upgraded to TLS.
```
sgallagher: I'm not sure what exactly happens to `insecure: "true"` when the URI is `ldaps://`. I assume that it just doesn't work (it shouldn't), but it's also possible that the use of `ldaps://` overrides the value of `insecure`.

@liggitt Do you know off the top of your head before I go digging through the code?

admin_solutions/authentication.adoc
```diff
@@ -14,9 +14,236 @@ toc::[]
...
+.. Define an RFC 2255 URL that
+link:../install_config/configuring_authentication.html#ldap-url[specifies the LDAP host and search parameters] to use.
+. Save your changes and close the file.
+. Start the {product-title} API server, specifying the configuration file you just
+modified:
++
+----
+$ openshift start master --config=<path/to/modified/config>/master-config.yaml
+----
```
sgallagher: This command is incorrect. You need to specify `--master-config` not `--config` on start.

admin_solutions/authentication.adoc
liggitt: Validation prevents that from happening.  That is considered an invalid configuration

sgallagher: Thanks.

OK, then it should probably be noted that the insecure parameter must be "false" when using `ldaps://` URIs.

tpoitras:
Updated as per comments above.

@liggitt - Can you please weigh in on @sgallagher comment from the email?

>>     4c/4f: What is the difference between 'id' and 'preferredUsername'? When would
>>     one be used instead of the other? Both would have to be unique, so why are they
>>     different, etc?
>>
> I'm not 100% sure, I pulled that info from elsewhere in our docs. If I had to
> guess I'd think that 'id' is the string of DN attributes required to identify
> the user internally within OSE, whereas that 'preferredUsername' would be the
> string of attributes used to create a username for display purposes only. But
> that's just a guess. Are there any other LDAP devs on your team who might know?
>
CCing Jordan Liggitt so he can help explain the difference. I *think* the idea
is that the preferredUsername has to be unique for all of OpenShift and the id
is unique within the provider, but Jordan can correct me on that.

tpoitras:
Heard back from Scott Seago with some feedback on Keystone. He no longer has a server for testing, so with help from @vikram-redhat we got our own Keystone server setup here in the Brisbane office. I managed to find out that I was missing something from the example stanza provided in our docs, so that was a good thing.

Updated Keystone docs here with latest commit (will squash all commits to one before merging).

tpoitras:
Made changes based on details emailed by Jordan Liggitt.

Moving to QE review now.

tpoitras:
Passed QE (Verified in BZ). Moving to Peer Review.
Can someone peer review this for me, please? :smile: @bfallonf @ahardin-rh @tnguyen-rh @adellape

admin_solutions/authentication.adoc
```diff
@@ -1,22 +1,262 @@
...
+
+- xref:keystone-auth[Keystone]
+- xref:ldap-auth[LDAP]
+- xref:github-auth[GitHub]
+
+[[keystone-auth]]
+== Keystone Authentication
+
+http://docs.openstack.org/developer/keystone/[Keystone] is an OpenStack project
+that provides identity, token, catalog and policy services. You can integrate
```
bfallonf: oxford comma

bfallonf:
One point. LGTM otherwise.

tpoitras:
thank you, @bfallonf ... I added that comma.

Will merge once Travis is done checking the build.

tpoitras:
This is marked as "future release" milestone right now because the Admin Solutions guide is currently not yet published

tpoitras:
[rev_history]
|link:../ admin_solutions/authentication.html[Authentication]
|Added Keystone and LDAP as authentication providers in this guide.
%

PR #2300, opened 2016-06-17
Need re-encrypting route for cluster metrics using generated self-signed certificates

tpoitras:
ping @TheDiemer for tech review, here's a nice-lookin' doc instead of a diff: http://file.bne.redhat.com/~tpoitras/2016/reencryptBZ1342170/openshift-enterprise/selfsigned-BZ1342170/install_config/cluster_metrics.html#metrics-using-secrets-autogenerated

https://bugzilla.redhat.com/show_bug.cgi?id=1342170

Added requirement of using re-encrypting route for cluster metrics that use generated self-signed certs.

### DIFF
--- a/install_config/cluster_metrics.adoc
+++ b/install_config/cluster_metrics.adoc
@@ -323,14 +323,28 @@ properly.
 ==== Using Generated Self-Signed Certificates
 
 The Metrics Deployer can accept multiple certificates using secrets. If a
-certificate is not passed as a secret, the deployer will generate a self-signed
-certificate to be used instead. For the deployer to generate certificates for
-you, a secret is still required before it can be deployed. In this case, create
-a "dummy" secret that does not specify a certificate value:
+certificate is not passed as a secret, then the deployer generates a self-signed
+certificate instead, forcing users to accept the certificate as a security
+exception.
+
+In order to use official certificates for the web console, you must use a
+link:../install_config/cluster_metrics.html#metrics-reencrypting-route[re-encrypting route].
+This allows the self-signed certificates to remain in use internally,
+while allowing your own certificates to be used for external access. When
+using a re-encrypting route, do not set the certificates as a secret. A
+"dummy" secret named *metrics-deployer* must still exist for the Metrics
+Deployer to generate certificates.
+
+To create a "dummy" secret that does not specify a certificate value:
 
 ----
 $ oc secrets new metrics-deployer nothing=/dev/null
 ----
+
+[CAUTION]
+====
+If you do not use a re-encrypting route when using generated self-signed certificates you will encounter errors.
+====
 
 [[modifying-the-deployer-template]]
 === Modifying the Deployer Template


tpoitras:
Pasting in from the BZ:

_Eric Jones 2016-06-18 05:22:16 EST

Hi Timothy,

So it appears that I incorrectly understood the situation. If you want to have official certs for your webUI for the metrics then you must use the re-encrypting route. If it does not matter that the users will have to accept the certificate as an exception, then you can simply use the sel-signed certs. 

I apologize for this confusion._

tpoitras:
Updated based on Eric's previous comment

tpoitras:
Passed tech review: "Hey Timothy, The most recent version at that link looks good to me. Explains that you would likely want to use the re-encrypting route with the self-signed certs, but it is not required. I think with that update, we will be good to close this bug whenever you are ready to."

tpoitras:
ping @ahardin-rh @bfallonf @tnguyen-rh @adellape for peer review, please. :smile_cat:

install_config/cluster_metrics.adoc
```diff
@@ -323,15 +323,28 @@ properly.
 ==== Using Generated Self-Signed Certificates
 
 The Metrics Deployer can accept multiple certificates using secrets. If a
-certificate is not passed as a secret, the deployer will generate a self-signed
-certificate to be used instead. For the deployer to generate certificates for
-you, a secret is still required before it can be deployed. In this case, create
-a "dummy" secret that does not specify a certificate value:
+certificate is not passed as a secret, then the deployer will generate and use a
+self-signed certificate instead, however this forces users to accept the certificate as a security exception.
```
bfallonf: There's something up with using "however" here, but I can't put my finger on why. Maybe because it usually denotes a negative? I think it might be wise to do away with it and say something like:

"...then the deployer generates a self-signed certificate instead, forcing users to accept the certificate as a security exception."

install_config/cluster_metrics.adoc
```diff
@@ -323,15 +323,28 @@ properly.
...
+
+In order to use official certificates for the web console, you must use a
+link:../install_config/cluster_metrics.html#metrics-reencrypting-route[re-encrypting route].
+This allows the self-signed certificates to remain in use internally
```
bfallonf: "...remain in use internally, while..."

and typo with externally access below? Maybe "external access"?

install_config/cluster_metrics.adoc
```diff
@@ -323,15 +323,28 @@ properly.
...
+while allowing your own certificates to be used for externally access. When
+using a re-encrypting route, do not set the certificates as a secret, but a
+"dummy" secret named *metrics-deployer* must still exist for the Metrics
+Deployer to generate certificates.
```
bfallonf: I'm confused about the above. It goes from a warning into an intrsuction? Maybe:

"When using a re-encrypting route, do not set the certificate as a secret. A "dummy" secret named *metrics-deployer* must still exist for the metrics deployer to generate certificates."

bfallonf:
@tpoitras Some nits. Looks good otherwise :+1:

tpoitras:
Thanks @bfallonf  for the peer review! :+1: 
Changes made.

tpoitras:
[rev_history]
|link:../install_config/cluster_metrics.html[Cluster Metrics]
|Added requirement of using re-encrypting route for cluster metrics that use generated self-signed certs.
%

PR #2319, opened 2016-06-21
Migration: Fixed final set of links for what's new book for OSE 3.0 a…

vikram-redhat:
…s a trial


vikram-redhat:
Migration: Fixed final set of links for whats new book for OSE 3.0 as a trial.

PR #2321, opened 2016-06-21
Minor follow-up edits to PR#2136 and PR#2290

ahardin-rh:


ahardin-rh:
No rev history needed

PR #2322, opened 2016-06-21
Improve repo enablement in upgrade docs

sdodson:
The extras repo should already be enabled however lets make sure of it. Also clean yum caches.


adellape:
@sdodson Thanks, worked these additions into https://github.com/openshift/openshift-docs/pull/2352.

PR #2103, opened 2016-05-16
Update PostgreSQL documentation

mnagy:
Two settings were added and were out of sync, see:
- https://github.com/openshift/postgresql/blob/master/9.2/README.md
- https://github.com/openshift/postgresql/tree/master/9.4/README.md

PR #2342, opened 2016-06-23
Follow-up edits for PR #2103

tnguyen-rh:
Follow-up edits for PR #2103

@adellape @ahardin-rh @tpoitras @bfallonf 
Peer review, please?

### DIFF
--- a/using_images/db_images/postgresql.adoc
+++ b/using_images/db_images/postgresql.adoc
@@ -121,6 +121,7 @@ default=>
 
 When you are finished, enter *\q* to leave the PostgreSQL session.
 
+[[pg-env-vars]]
 === Environment Variables
 
 The PostgreSQL user name, password, and database name must be configured with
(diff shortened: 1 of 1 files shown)


tnguyen-rh:
[rev_history]
|link:../using_images/db_images/postgresql.html[Database Images -> PostgreSQL]
|Updated `*POSTGRESQL_MAX_CONNECTIONS*` and added `*POSTGRESQL_MAX_PREPARED_TRANSACTIONS*` and `*POSTGRESQL_EFFECTIVE_CACHE_SIZE*` to the link:../using_images/db_images/postgresql.html#pg-env-vars[Environment Variables] section.
%

bfallonf:
@tnguyen-rh This is fine. 

Originally I was questioning why the "the"s were gone, but I can see that it's not perfect grammar to have them in. I think what would be ideal would be something like "The value specified here does X", but that's not really consistent with the rest of the docs.

Anyway, this all LGTM.

PR #2355, opened 2016-06-27
Deployment Logs: added slash to syntax

tpoitras:
RE: https://bugzilla.redhat.com/show_bug.cgi?id=1350002

Example commands were missing a slash. Simple fix.


tpoitras:
No revhistory needed

PR #2014, opened 2016-05-05
add docker login using service account tokens

aweiteka:
This addresses a common question since `docker login -p $(oc whoami -t) -u unused -e unused REGISTRY:5000` expires every 24 hours by default. I suspect we should incorporate some of this as a separate PR for users who do not want or cannot use openshift to perform docker build.

PR #2460, opened 2016-07-05
Follow-up to PR#2014: Fixed xrefs and added an anchor

ahardin-rh:
Follow-up edits to to https://github.com/openshift/openshift-docs/pull/2014


ahardin-rh:
Minor edits. No rev history needed.

PR #2454, opened 2016-07-05
added information about using xref to the contributing guide

vikram-redhat:

PR #2472, opened 2016-07-06
update guidelines to use xref macro

aweiteka:


tnguyen-rh:
@aweiteka Thanks.  Turns out, this is already addressed in #2454 -- closing.

PR #2477, opened 2016-07-08
Adding Basic auth and Request Header auth to Admin Solutions guide

tpoitras:
https://bugzilla.redhat.com/show_bug.cgi?id=1339838

### DIFF
--- a/admin_solutions/authentication.adoc
+++ b/admin_solutions/authentication.adoc
@@ -5,9 +5,306 @@ toc::[]
 {product-title} supports many different authentication methods, as defined in
 xref:../install_config/configuring_authentication.adoc#install-config-configuring-authentication[Configuring Authentication]:
 
+- xref:basic-auth-remote[Basic Authentication (Remote)]
+- xref:request-header-auth[Request Header]
 - xref:keystone-auth[Keystone]
 - xref:ldap-auth[LDAP]
 - xref:github-auth[GitHub]
+
+[[basic-auth-remote]]
+== Basic Authentication (Remote)
+
+[[BasicAuthPasswordIdentityProvider]]
+
+Basic Authentication is a generic backend integration mechanism that allows
+users to log in to {product-title} with credentials validated against a remote
+identity provider.
+
+[CAUTION]
+====
+Basic Authentication must use an HTTPS connection to the remote server in order
+to prevent potential snooping of the user ID and password, and to prevent
+man-in-the-middle attacks.
+====
+
+With `*BasicAuthPasswordIdentityProvider*` configured, users send their username
+and password to {product-title}, which then validates those credentials against
+a remote server by making a server-to-server request, passing the credentials as
+a Basic Auth header. This requires users to send their credentials to
+{product-title} during login. 
+
+[NOTE]
+====
+This only works for username/password login mechanisms, and {product-title} must
+be able to make network requests to the remote authentication server. 
+====
+
+[[configuring-basic-auth-on-master]]
+=== Configuring Authentication on the Master
+
+. If you have:
++
+- Already completed the installation of Openshift, then copy the
+*_/etc/origin/master/master-config.yaml_* file into a new directory; for example:
++
+----
+$ mkdir basicauthconfig; cp master-config.yaml basicauthconfig
+----
++
+- Not yet installed {product-title}, then start the {product-title} API server,
+specifying the hostname of the (future) {product-title} master and a directory
+to store the configuration file created by the start command:
++
+----
+$ openshift start master --public-master=<apiserver> --write-config=<directory>
+----
++
+For example:
++
+----
+$ openshift start master --public-master=https://myapiserver.com:8443 --write-config=basicauthconfig
+----
++
+[NOTE]
+====
+If you are installing with Ansible, then you must add the
+`*identityProvider*` configuration to the Ansible playbook.
+If you use the following steps to modify your configuration manually after installing with Ansible, then you will lose any modifications whenever you re-run the install tool or upgrade.
+====
++
+. Edit the new *_master-config.yaml_* file's `*identityProviders*` stanza.
+. Copy
+xref:../install_config/configuring_authentication.adoc#basic-auth-example-config[the example `*BasicAuthPasswordIdentityProvider*` configuration] and paste it to replace the existing stanza.
+. Make the following modifications to the `*identityProviders*` stanza:
+.. Set the provider `*name*` to something unique and relevant to your
+deployment. This name is prefixed to the returned user ID to form an identity
+name.
+.. If required,
+xref:../install_config/configuring_authentication.adoc#mapping-identities-to-users[set `*mappingMethod*`] to control how mappings are established between the
+provider's identities and user objects.
+.. Specify the HTTPS `*url*` to use to connect to a server that accepts credentials in Basic authentication headers.
+.. Optionally, set the `*ca*` to the certificate bundle to use in order to validate server certificates for the configured URL, or leave it empty to use the system-trusted roots.
+.. Optionally, remove or set the `*certFile*` to the client certificate to present when making requests to the configured URL.
+.. If `*certFile*` is specified, then you must set the `*keyFile*` to the key for the client certificate.
+. Save your changes and close the file.
+. Start the {product-title} API server, specifying the configuration file you just
+modified:
++
+----
+$ openshift start master --config=<path/to/modified/config>/master-config.yaml
+----
+
+Once configured, any user logging in to the {product-title} web console will be
+prompted to log in using their Basic authentication credentials.
+
+[[basic-troubleshooting]]
+=== Troubleshooting
+
+The most common issue relates to network connectivity to the backend server. For
+simple debugging, run `curl` commands on the master. To test for a successful
+login, replace the `<user>` and `<password>` in the following example command
+with valid credentials. To test an invalid login, replace them with false
+credentials.
+
+----
+curl --cacert /path/to/ca.crt --cert /path/to/client.crt --key /path/to/client.key -u <user>:<password> -v https://www.example.com/remote-idp
+----
+
+*Successful responses*
+
+A `200` status with a `sub` (subject) key indicates success:
+
+----
+{"sub":"userid"}
+----
+The subject must be unique to the authenticated user, and must not be able to
+be modified.
+
+A successful response may optionally provide additional data, such as:
+
+* A display name using the `name` key:
++
+----
+{"sub":"userid", "name": "User Name", ...}
+----
+* An email address using the `email` key:
++
+----
+{"sub":"userid", "email":"user@example.com", ...}
+----
+* A preferred user name using the `preferred_username` key: 
++
+----
+{"sub":"014fbff9a07c", "preferred_username":"bob", ...}
+----
++
+The `preferred_username` key is useful when
+the unique, unchangeable subject is a database key or UID, and a more
+human-readable name exists. This is used as a hint when provisioning the
+{product-title} user for the authenticated identity.
+
+*Failed responses*
+
+- A `401` response indicates failed authentication.
+- A non-`200` status or the presence of a non-empty "error" key indicates an
+error: `{"error":"Error message"}`
+
+[[view-users-basic-auth]]
+=== Viewing Users
+
+Once one or more users have logged in, you can run `oc get users` to view a
+list of users and verify that users were created successfully. 
+
+From here, you might want to learn how to
+xref:../admin_solutions/user_role_mgmt.adoc#control-user-roles[control user roles].
+
+
+[[request-header-auth]]
+== Request Header Authentication
+
+Configuring Request Header authentication allows users to log in to
+{product-title} using request header values, such as `X-Remote-User`. It is
+typically used in combination with an authenticating proxy, which authenticates
+the user and then provides {product-title} with the user's identity via a
+request header value. This is similar to how
+link:https://access.redhat.com/documentation/en-US/OpenShift_Enterprise/2/html/Deployment_Guide/Configuring_OpenShift_Enterprise_Authentication.html[the remote user plug-in in OpenShift Enterprise 2] 
+allowed administrators to provide Kerberos, LDAP, and many other forms of
+enterprise authentication. The benefit of this configuration is that user
+credentials can be handled by the proxy and never seen by OpenShift.
+
+The proxy must be able to make network requests to the {product-title} server.
+Unauthenticated login attempts are redirected to a configured proxy URL. The
+proxy can authenticate browser clients regardless of how it is configured, but
+it must (currently) use either Basic Auth or Kerberos in order to work with the
+`oc` CLI tooling.
+
+For users to authenticate using this identity provider, they must access
+_\https://<master>/oauth/authorize_ via an authenticating proxy. You can configure
+the OAuth server to redirect unauthenticated requests to the proxy.
+
+[[configuring-request-header-auth-on-master]]
+=== Configuring Authentication on the Master
+
+. If you have:
++
+- Already completed the installation of Openshift, then copy the
+*_/etc/origin/master/master-config.yaml_* file into a new directory; for example:
++
+----
+$ mkdir reqheadauthconfig; cp master-config.yaml reqheadauthconfig
+----
++
+- Not yet installed {product-title}, then start the {product-title} API server,
+specifying the hostname of the (future) {product-title} master and a directory
+to store the configuration file created by the start command:
++
+----
+$ openshift start master --public-master=<apiserver> --write-config=<directory>
+----
++
+For example:
++
+----
+$ openshift start master --public-master=https://myapiserver.com:8443 --write-config=reqheadauthconfig
+----
++
+[NOTE]
+====
+If you are installing with Ansible, then you must add the
+`*identityProvider*` configuration to the Ansible playbook.
+If you use the following steps to modify your configuration manually after installing with Ansible, then you will lose any modifications whenever you re-run the install tool or upgrade.
+====
++
+. Edit the new *_master-config.yaml_* file's `*identityProviders*` stanza.
+. View
+xref:../install_config/configuring_authentication.adoc#reqhead-auth-example-config[the example `*RequestHeaderIdentityProvider*` configuration] 
+and use it as a guide to replace the existing stanza.
+. Modify the `*identityProviders*` stanza based on which headers you plan to
+pass in.
+.. Set the provider `*name*` to something unique and relevant to your
+deployment. This name is prefixed to the returned user ID to form an identity
+name.
+.. If required,
+xref:../install_config/configuring_authentication.adoc#mapping-identities-to-users[set `*mappingMethod*`] 
+to control how mappings are established between the provider's identities and
+user objects.
+.. Set the `*challenge*` parameter to *true* to redirect unauthenticated
+requests from clients expecting `WWW-Authenticate` challenges. 
+.. Set the `*provider.challengeURL*` parameter to the proxy URL to which to send
+clients expecting `WWW-Authenticate` challenges, like the `oc` CLI client.
+This parameter can include the 
+xref:../install_config/configuring_authentication.adoc#RequestHeaderIDP-urlquerytokens[`${url}` and `${query}` tokens]
+in the query portion of the URL. 
+.. Set the `*login*` parameter to *true* to redirect unauthenticated requests
+from clients expecting login flows. 
+.. Set the `*provider.loginURL*` parameter to the proxy URL to which to send
+clients expecting login flows, like web browser clients. This parameter can include the 
+xref:../install_config/configuring_authentication.adoc#RequestHeaderIDP-urlquerytokens[`${url}` and `${query}` tokens]
+in the query portion of the URL. 
+.. Set the `*clientCA*` parameter to the certificate bundle to use to check
+incoming requests for a valid client certificate before the request's headers
+are checked for a user name. 
++
+[WARNING]
+====
+If you expect unauthenticated requests to reach the OAuth server, a `*clientCA*`
+parameter (and optionally, `*clientCommonNames*`) should be set for this
+identity provider. Otherwise, any direct request to the OAuth server can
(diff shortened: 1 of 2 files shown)


admin_solutions/authentication.adoc
```diff
@@ -5,10 +5,308 @@ toc::[]
...
 - xref:ldap-auth[LDAP]
 - xref:github-auth[GitHub]
 
+[[basic-auth-remote]]
+== Basic Authentication (Remote)
+
+[[BasicAuthPasswordIdentityProvider]]
+
+Configuring Basic Authentication allows users to log in to {product-title} with
+credentials stored on a remote identity provider.
```
liggitt: This is not specific enough... all the auth integrations do that.

BasicAuthPasswordIdentityProvider and RequestHeaderIdentityProvider are the most generic integration mechanisms.

BasicAuthPasswordIdentityProvider is a backend integration. It makes users send their username/password to OpenShift, and OpenShift then verifies those credentials against an external server by making a server-to-server request, passing the given credentials as a basic auth header.
* This only works for username/password login mechanisms
* OpenShift must be able to make network requests to the remote auth server
* Requires users to send their credentials to OpenShift during login

RequestHeaderIdentityProvider makes users talk to OpenShift through a proxy, which is responsible for 1) authenticating the user any way it wants to, and 2) providing the user's identity to OpenShift via a request header.
* The proxy must be able to make network requests to the OpenShift server
* Unauthenticated login attempts are redirected to a configured proxy URL
* The proxy can authenticate browser clients any way it wants to, but must currently use basic auth or kerberos if it wants to work with the `oc` CLI tooling.
* Means user credentials can be handled by the proxy and never seen by OpenShift

admin_solutions/authentication.adoc
```diff
@@ -5,10 +5,308 @@ toc::[]
...
+
+Once configured, any user logging in to the {product-title} web console will be
+prompted to log in using their Basic authentication credentials.
+
+[[basic-troubleshooting]]
+=== Troubleshooting
+
+[WARNING]
+====
+*[NEEDINFO]*: Is this troubleshooting information? It was in introduction of the install guide authentication topic, but seems out of place there. What purpose does this serve? Does any of it belong in the overview here, in other words does any of this need to be considered before configuring authentication?
```
liggitt: Most troubleshooting will probably be around network connectivity to the back end server.

Could include equivalent `curl` commands that could be run on the master for debugging. Something like this would be the equivalent of what the OpenShift API server would do, given the sample config. Replace "sampleuser:samplepassword" with a valid username/password to test a valid login, and replace with an invalid username/password to test an invalid login.

```
curl --cacert /path/to/ca.crt --cert /path/to/client.crt --key /path/to/client.key -u sampleuser:samplepassword -v https://www.example.com/remote-idp
```

admin_solutions/authentication.adoc
```diff
@@ -5,10 +5,308 @@ toc::[]
...
+
+You do not create users in {product-title} when integrating with an external
+authentication provider, such as, in this case, Basic authentication. The Basic
+authentication server is the system of record, meaning that users are defined
+there, and any user with a valid user name for the configured authentication
+server can log in.
+
+To add a user to {product-title}, the user must exist on the Basic
+authentication server, and if required you must create a new Basic
+authentication account for the user.
```
liggitt: a "basic authentication account" is not a thing... "Basic authentication" refers to the way the credentials are sent from OpenShift to the remote auth server, nothing more

admin_solutions/authentication.adoc
```diff
@@ -5,10 +5,308 @@ toc::[]
...
+typically used in combination with an authenticating proxy, which sets the
+request header value. This is similar to how
+link:https://access.redhat.com/documentation/en-US/OpenShift_Enterprise/2/html/Deployment_Guide/Configuring_OpenShift_Enterprise_Authentication.html[the
+remote user plug-in in OpenShift Enterprise 2] allowed administrators to
+provide Kerberos, LDAP, and many other forms of enterprise authentication.
+
+For users to authenticate using this identity provider, they must access
+_<master>/oauth/authorize_ via an authenticating proxy. You can either proxy the
+entire master API server so that all access goes through the proxy, or you can
+configure the OAuth server to redirect unauthenticated requests to the proxy.
```
liggitt: I would not recommend proxying the entire API

admin_solutions/authentication.adoc
```diff
@@ -5,10 +5,308 @@ toc::[]
...
+
+[WARNING]
+====
+*NEEDINFO*: The value <master>/oauth/authorize, is that a URL and not a file path?
```
liggitt: yes, a URL

admin_solutions/authentication.adoc
```diff
@@ -5,10 +5,308 @@ toc::[]
...
+.. Set the `*provider.challengeURL*` parameter to the proxy URL to which to send clients expecting `WWW-Authenticate` challenges. 
+This parameter can include
+the 
+xref:../install_config/configuring_authentication.adoc#RequestHeaderIDP-urlquerytokens[`${url}` and `${query}` tokens]
+in the query portion of the URL. 
+.. Set the `*login*` parameter to *true* to redirect unauthenticated requests from clients expecting login flows. 
+.. Set the `*provider.loginURL*` parameter to the proxy URL to which to send clients expecting login flows. This parameter can include the 
+xref:../install_config/configuring_authentication.adoc#RequestHeaderIDP-urlquerytokens[`${url}` and `${query}` tokens]
+in the query portion of the URL. 
+.. Optionally, change the `*clientCA*` parameter to check incoming requests for a valid client certificate before the request's headers are checked for a user name.
```
liggitt: and `*clientCommonNames*`

admin_solutions/authentication.adoc
```diff
@@ -5,10 +5,308 @@ toc::[]
...
+in the query portion of the URL. 
+.. Set the `*login*` parameter to *true* to redirect unauthenticated requests from clients expecting login flows. 
+.. Set the `*provider.loginURL*` parameter to the proxy URL to which to send clients expecting login flows. This parameter can include the 
+xref:../install_config/configuring_authentication.adoc#RequestHeaderIDP-urlquerytokens[`${url}` and `${query}` tokens]
+in the query portion of the URL. 
+.. Optionally, change the `*clientCA*` parameter to check incoming requests for a valid client certificate before the request's headers are checked for a user name. 
++
+[WARNING]
+====
+If you expect unauthenticated requests to reach the OAuth server, a `*clientCA*`
```
liggitt: and optionally `clientCommonNames`

admin_solutions/authentication.adoc
```diff
@@ -5,10 +5,308 @@ toc::[]
...
+====
+. Save your changes and close the file.
+. Start the {product-title} API server, specifying the configuration file you just modified:
++
+----
+$ openshift start master --config=<path/to/modified/config>/master-config.yaml
+----
+
+Once configured, any user logging in to the {product-title} web console will be
+prompted to log in using Request Header credentials.
```
liggitt: "Request Header credentials" aren't a thing. They'll be redirected to the authenticating proxy, which will authenticate the user in some unspecified way.

admin_solutions/authentication.adoc
```diff
@@ -5,10 +5,308 @@ toc::[]
...
+[[create-users-request-header-auth]]
+=== Creating Users with Request Header Authentication
+
+You do not create users in {product-title} when integrating with an external
+authentication provider, such as, in this case, Request Header. The Request
+Header authentication server is the system of record, meaning that users are
+defined there, and any user with a valid user name for the configured
+authentication server can log in.
+
+To add a user to {product-title}, the user must exist on the Request Header authentication server, and if required you must create a new Request Header authentication account for the user.
```
liggitt: This has to be more vague... they'll need to add the users to whatever system the proxy is using to authenticate them.

admin_solutions/authentication.adoc
```diff
@@ -5,10 +5,308 @@ toc::[]
...
+$ oc get users
+NAME         UID                                    FULL NAME   IDENTITIES
+bobsmith     a0c1d95c-1cb5-11e6-a04a-002186a28631   Bob Smith   requestheader:bobsmith <1>
+----
+<1> Identities in {product-title} are comprised of the identity provider name prefixed to the Request Header authentication username.
+====
+
+[WARNING]
+====
+*[NEEDINFO]*: Do you have access to a request header authentication setup? If so, can you please provide me with the output for the `oc get users` command when request header is configured as the auth provider for OpenShift? Thank you.
```
liggitt: the users that result from all the different auth methods are identical. they only vary based on the name that was chosen for the provider, and whether the provider makes display name info available.

admin_solutions/authentication.adoc
liggitt: Need to describe setting the headers that will be used for the identity, preferred username, display name, and email

liggitt: I don't know that I would have them copy/paste that config as a starting point... I'd want them to decide what headers they were going to pass in, not start with a list designed to illustrate looking at multiple headers, when they might not want that

admin_solutions/authentication.adoc
```diff
@@ -5,10 +5,308 @@ toc::[]
...
+$ oc get users
+NAME         UID                                    FULL NAME   IDENTITIES
+bobsmith     a0c1d95c-1cb5-11e6-a04a-002186a28631   Bob Smith   basicauth:bobsmith <1>
+----
+<1> Identities in {product-title} are comprised of the identity provider name prefixed to the Basic authentication username.
+====
+
+[WARNING]
+====
+*[NEEDINFO]*: Do you have access to a basic authentication setup? If so, can you please provide me with the output for the `oc get users` command when basic auth is configured as the auth provider for OpenShift? Thank you.
```
liggitt: this step doesn't make sense to repeat for every identity provider... `oc get users` looks the same, no matter what IDP you use.

tpoitras:
Made updates based on technical review from @liggitt

tpoitras:
Made minor update based on review from Takayoshi Kimura and Eido Inoue

tpoitras:
I'll squash before merging.

Ping @bfallonf @ahardin-rh @tnguyen-rh @adellape for Peer Review, please. :alien:

tpoitras:
Moved BZ along to QA, once it passes will do peer review.

admin_solutions/authentication.adoc
```diff
@@ -5,10 +5,303 @@ toc::[]
...
+. Copy
+xref:../install_config/configuring_authentication.adoc#basic-auth-example-config[the example `*BasicAuthPasswordIdentityProvider*` configuration] and paste it to replace the existing stanza.
+. Make the following modifications to the `*identityProviders*` stanza:
+.. Change the provider `*name*` ("my_remote_basic_auth_provider") to something
+unique and relevant to your deployment. This name is prefixed to the returned
+user ID to form an identity name.
+.. If required,
+xref:../install_config/configuring_authentication.adoc#mapping-identities-to-users[change `*mappingMethod*`] to control how mappings are established between the
+provider's identities and user objects.
+.. Specify the `*url*` to use to connect to a server that accepts credentials in Basic authentication headers.
```
liggitt: note here that this should be an HTTPS URL

admin_solutions/authentication.adoc
```diff
@@ -5,10 +5,303 @@ toc::[]
...
+
+Basic Authentication is a generic backend integration mechanism that allows
+users to log in to {product-title} with credentials stored on a remote identity
+provider.
+
+[CAUTION]
+====
+Basic Authentication is not a secure exchange in principle; it needs to be
+combined with encryption or certificate authentication to prevent potential
+snooping of the user ID and password, and to prevent man-in-the-middle attacks.
```
liggitt: I understand what this is saying, but I would not lead with "Basic Authentication is not a secure exchange in principle". I would rephrase as "Basic Authentication should use a HTTPS connection to the remote server in order to prevent potential snooping of the user ID and password, and to prevent man-in-the-middle attacks."

admin_solutions/authentication.adoc
```diff
@@ -5,10 +5,303 @@ toc::[]
...
 - xref:ldap-auth[LDAP]
 - xref:github-auth[GitHub]
 
+[[basic-auth-remote]]
+== Basic Authentication (Remote)
+
+[[BasicAuthPasswordIdentityProvider]]
+
+Basic Authentication is a generic backend integration mechanism that allows
+users to log in to {product-title} with credentials stored on a remote identity
```
liggitt: s/stored on/validated against/

admin_solutions/authentication.adoc
```diff
@@ -5,10 +5,303 @@ toc::[]
...
+xref:../install_config/configuring_authentication.adoc#basic-auth-example-config[the example `*BasicAuthPasswordIdentityProvider*` configuration] and paste it to replace the existing stanza.
+. Make the following modifications to the `*identityProviders*` stanza:
+.. Change the provider `*name*` ("my_remote_basic_auth_provider") to something
+unique and relevant to your deployment. This name is prefixed to the returned
+user ID to form an identity name.
+.. If required,
+xref:../install_config/configuring_authentication.adoc#mapping-identities-to-users[change `*mappingMethod*`] to control how mappings are established between the
+provider's identities and user objects.
+.. Specify the `*url*` to use to connect to a server that accepts credentials in Basic authentication headers.
+.. Optionally, change the `*ca*` to the certificate bundle to use in order to validate server certificates for the configured URL.
```
liggitt: note, "or leave it empty to use the system trusted roots"

liggitt: none of the example sample paths are likely to work for them... I would say "set" instead of "change"

admin_solutions/authentication.adoc
```diff
@@ -5,10 +5,303 @@ toc::[]
...
+.. Optionally, change the `*certFile*` to the client certificate to present when making requests to the configured URL.
```
liggitt: set or remove

admin_solutions/authentication.adoc
```diff
@@ -5,10 +5,303 @@ toc::[]
...
+the user and then provides {product-title} with the user's identity via a
+request header value. This is similar to how
+link:https://access.redhat.com/documentation/en-US/OpenShift_Enterprise/2/html/Deployment_Guide/Configuring_OpenShift_Enterprise_Authentication.html[the remote user plug-in in OpenShift Enterprise 2] 
+allowed administrators to provide Kerberos, LDAP, and many other forms of
+enterprise authentication. The benefit of this configuration is that user
+credentials can be handled by the proxy and never seen by OpenShift.
+
+The proxy must be able to make network requests to the {product-title} server.
+Unauthenticated login attempts are redirected to a configured proxy URL. The
+proxy can authenticate browser clients however it is configured, but it must
```
liggitt: "however it is configured" reads oddly.

admin_solutions/authentication.adoc
```diff
@@ -5,10 +5,303 @@ toc::[]
...
+pass in.
+.. Set the provider `*name*` to something unique and relevant to your
+deployment. This name is prefixed to the returned user ID to form an identity
+name.
+.. If required,
+xref:../install_config/configuring_authentication.adoc#mapping-identities-to-users[change `*mappingMethod*`] 
+to control how mappings are established between the provider's identities and
+user objects.
+.. Set the `*challenge*` parameter to *true* to redirect unauthenticated
+requests from clients expecting `WWW-Authenticate` challenges.
```
liggitt: "(like the CLI)"

admin_solutions/authentication.adoc
```diff
@@ -5,10 +5,303 @@ toc::[]
...
+deployment. This name is prefixed to the returned user ID to form an identity
+name.
+.. If required,
+xref:../install_config/configuring_authentication.adoc#mapping-identities-to-users[change `*mappingMethod*`] 
+to control how mappings are established between the provider's identities and
+user objects.
+.. Set the `*challenge*` parameter to *true* to redirect unauthenticated
+requests from clients expecting `WWW-Authenticate` challenges. 
+.. Set the `*provider.challengeURL*` parameter to the proxy URL to which to send
+clients expecting `WWW-Authenticate` challenges.
```
liggitt: ", like the `oc` CLI client"

admin_solutions/authentication.adoc
```diff
@@ -5,10 +5,303 @@ toc::[]
...
+user objects.
+.. Set the `*challenge*` parameter to *true* to redirect unauthenticated
+requests from clients expecting `WWW-Authenticate` challenges. 
+.. Set the `*provider.challengeURL*` parameter to the proxy URL to which to send
+clients expecting `WWW-Authenticate` challenges. 
+This parameter can include the 
+xref:../install_config/configuring_authentication.adoc#RequestHeaderIDP-urlquerytokens[`${url}` and `${query}` tokens]
+in the query portion of the URL. 
+.. Set the `*login*` parameter to *true* to redirect unauthenticated requests
+from clients expecting login flows.
```
liggitt: ", like web browser clients"

admin_solutions/authentication.adoc
```diff
@@ -5,10 +5,303 @@ toc::[]
...
+requests from clients expecting `WWW-Authenticate` challenges. 
+.. Set the `*provider.challengeURL*` parameter to the proxy URL to which to send
+clients expecting `WWW-Authenticate` challenges. 
+This parameter can include the 
+xref:../install_config/configuring_authentication.adoc#RequestHeaderIDP-urlquerytokens[`${url}` and `${query}` tokens]
+in the query portion of the URL. 
+.. Set the `*login*` parameter to *true* to redirect unauthenticated requests
+from clients expecting login flows. 
+.. Set the `*provider.loginURL*` parameter to the proxy URL to which to send
+clients expecting login flows. This parameter can include the
```
liggitt: ", like web browser clients"

admin_solutions/authentication.adoc
```diff
@@ -5,10 +5,303 @@ toc::[]
...
+This parameter can include the 
+xref:../install_config/configuring_authentication.adoc#RequestHeaderIDP-urlquerytokens[`${url}` and `${query}` tokens]
+in the query portion of the URL. 
+.. Set the `*login*` parameter to *true* to redirect unauthenticated requests
+from clients expecting login flows. 
+.. Set the `*provider.loginURL*` parameter to the proxy URL to which to send
+clients expecting login flows. This parameter can include the 
+xref:../install_config/configuring_authentication.adoc#RequestHeaderIDP-urlquerytokens[`${url}` and `${query}` tokens]
+in the query portion of the URL. 
+.. Optionally, change the `*clientCA*` parameter to check incoming requests for
```
liggitt: This should absolutely be done. Without checking the request containing user info in headers against a client CA, anyone that can hit the API server can assume any identity they want

liggitt: I would remove the "Optionally", and just say "Change the clientCA parameter to the certificate bundle to use..."

tpoitras:
Thanks for the feedback, @liggitt - I have incorporated it.

tpoitras:
Peer Review, please: @bfallonf @ahardin-rh @adellape

tpoitras:
bump @bfallonf @ahardin-rh @adellape

admin_solutions/authentication.adoc
```diff
@@ -5,10 +5,303 @@ toc::[]
...
+====
+This only works for username/password login mechanisms, and {product-title} must
+be able to make network requests to the remote authentication server. 
+====
+
+[[configuring-basic-auth-on-master]]
+=== Configuring Authentication on the Master
+
+. If you have:
+.. Already completed the installation of Openshift, then copy the
```
bfallonf: These substeps (1a and 1b) are options rather than substeps, correct? Can they be marked up with points rather than steps? It's one or the other, so not sure if numbered list is right?

admin_solutions/authentication.adoc
```diff
@@ -5,10 +5,303 @@ toc::[]
...
+====
+If you are installing with Ansible, then you must add the
+`*identityProvider*` configuration to the Ansible playbook.
+If you use the following steps to modify your configuration manually after installing with Ansible, then you will lose any modifications whenever you re-run the install tool or upgrade.
+====
++
+. Edit the new *_master-config.yaml_* file's `*identityProviders*` stanza.
+. Copy
+xref:../install_config/configuring_authentication.adoc#basic-auth-example-config[the example `*BasicAuthPasswordIdentityProvider*` configuration] and paste it to replace the existing stanza.
+. Make the following modifications to the `*identityProviders*` stanza:
```
bfallonf: Same comment as above

admin_solutions/authentication.adoc
```diff
@@ -5,10 +5,303 @@ toc::[]
...
+
+For users to authenticate using this identity provider, they must access
+_\https://<master>/oauth/authorize_ via an authenticating proxy. You can configure
+the OAuth server to redirect unauthenticated requests to the proxy.
+
+[[configuring-request-header-auth-on-master]]
+=== Configuring Authentication on the Master
+
+. If you have:
+.. Already completed the installation of Openshift, then copy the
```
bfallonf: Gah, same comment!

bfallonf:
@tpoitras Largely great. A few comments about list-type stuff. Otherwise, LGTM :peach:

admin_solutions/authentication.adoc
tpoitras: no, the comment you made above was valid, but these are required steps aside from the "optionally" ones, and we still number "optional" steps in our enumerated lists, so, I'm leaving it as is.

tpoitras:
Thanks for the review, @bfallonf , and good catch on that list formatting.

I've incorporated your feedback, and pushed the changes (now squashed flat to one commit)

Rendered docs updated: http://file.bne.redhat.com/tpoitras/2016/basicrequest/openshift-enterprise/basicrequest-BZ1339838/admin_solutions/authentication.html

Merging for Next Release.

tpoitras:
[rev_history]
|link:../admin_solutions/authentication.html[Source-to-Image (S2I) -> Ruby]
|Added xref:../admin_solutions/authentication.adoc#basic-auth-remote[Basic Auth] and xref:../admin_solutions/authentication.adoc#request-header-auth[Request Header] to the Authentication topic.
%

PR #2493, opened 2016-07-13
Bug 1354147 added info on viewing a build

bfallonf:
As per: https://bugzilla.redhat.com/show_bug.cgi?id=1354147

I'll ask for an ack in the BZ

### DIFF
--- a/dev_guide/builds.adoc
+++ b/dev_guide/builds.adoc
@@ -1285,16 +1285,25 @@ $ oc delete --cascade=false bc <BuildConfigName>
 [[viewing-build-details]]
 == Viewing Build Details
 
-You can view build details using the web console or the following CLI command:
+You can view build details with the web console or by using the `oc describe`
+CLI command:
 
+====
 ----
 $ oc describe build <build_name>
 ----
+====
 
-The output of the `describe` command includes details such as the build source,
-strategy, and output destination. If the build uses the Docker or Source
-strategy, it will also include information about the source revision used for
-the build: commit ID, author, committer, and message.
+This displays information such as:
+
+* The build source
+* The strategy
+* The output destination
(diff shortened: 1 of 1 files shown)


bfallonf:
@PI-Victor Actually I'll ask here... Victor, could I ask if you think anything more is needed for this BZ? I don't think it's worth adding an example more that the CLI command, as it does change.

Thanks!

PI-Victor:
@bfallonf it would be good to add the fact that it shows webhook trigger details whenever it was triggered via webhook. it also shows the obfuscated secret. this is worth mentioning because it's in the actual web-console.
i did have docs to the trello card for this, but in the end i decided it wasn't much to account for, exactly because it will change.
i also have it in plan to add user details/token info to this output but i'm stuck with some technical issues and won't be able to do it sooner than two weeks from now.

bfallonf:
@PI-Victor I think if we describe one part of the oc describe output we'd have to describe it all, so I might have to disagree. The section does mention that you can see it in the web console, and that's pretty self-explanatory, so if there's no other issues, I think it's fine.

Overall, this doesn't go into the docs until 3.3, so we have more time to doc the other stuff you mention. I can keep an eye out.

bfallonf:
@tpoitras @ahardin-rh @tnguyen-rh @adellape Any issues?

PI-Victor:
@bfallonf nope, i agree you're right. keep it that way!

ahardin-rh:
@bfallonf LGTM! :star2:

bfallonf:
@PI-Victor Great, thanks for the thumbs up!

@ahardin-rh And thanks for the eyes. I'll merge.

bfallonf:
I don't think a rev history update is needed here.

PR #2496, opened 2016-07-13
Bug 1235445, added a Versioning Policy

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1235445

### DIFF
--- a/release_notes/index.adoc
+++ b/release_notes/index.adoc
@@ -20,3 +20,41 @@ technical notes, see the
 https://docs.openshift.com/enterprise/3.1/release_notes/index.html[OpenShift
 Enterprise 3.1 Release Notes].
 endif::[]
+
+[[release-versioning-policy]]
+== Versioning Policy
+
+{product-title} provides strict backwards compatibility guarantees for all
+supported APIs (excluding alpha APIs, which may be changed without notice, and
+beta APIs, which may occasionally be changed in a non-backwards compatible
+manner). {product-title} will continue to support older clients against newer
+servers. For example, a 3.1 client will work against 3.0, 3.1, and 3.2 servers.
+
+Changes of APIs for non-security related reasons will involve, at minimum, two
+minor releases (3.1 to 3.2 to 3.3, for example) to allow older clients to
+update. Using new capabilities may require newer clients. A 3.2 server may have
+additional capabilities that a 3.1 client cannot use and a 3.2 client may have
+additional capabilities that are not supported by a 3.1 server.
+
+.Compatibility Matrix
+
+[cols="1,1,1"]
+|===
+
+|
+|*X.Y* (Client)
+|*X.Y+N* footnoteref:[versionpolicyn,Where *N* is a number greater than 1.] (Client)
+
+|*X.Y* (Server)
+|1
+|3
+
+|*X.Y+N* footnoteref:[versionpolicyn] (Server)
+|2
+|1
+
+|===
+
+<1> Fully compatible.
+<2> Client may not be able to access server features.
+<3> Client may provide options and features that may not be compatible with the accessed server.


ahardin-rh:
@smarterclayton This is my first stab at trying to define this versioning policy. PTAL. Thanks!

release_notes/index.adoc
```diff
@@ -20,3 +20,11 @@ technical notes, see the
...
 Enterprise 3.1 Release Notes].
 endif::[]
+
+[[release-versioning-policy]]
+== Versioning Policy
+{product-title} does not have non-backwards compatible changes and, therefore,
+does not drop features on versions earlier than v3. Some upgrades may require
+changes by the administrator. Clients will continue to work against major
+version servers. For example, a 3.1 client will work against 3.0, 3.1, and 3.2.
+Not all features are available in 3.1 clients against 3.2 servers.
```
akostadinov: Will all features of `3.1` client be available against a `3.0` server?

ahardin-rh:
@smarterclayton can you please review? Thanks!

ahardin-rh:
@akostadinov can you please review this? Thanks!

release_notes/index.adoc
```diff
@@ -20,3 +20,11 @@ technical notes, see the
 https://docs.openshift.com/enterprise/3.1/release_notes/index.html[OpenShift
 Enterprise 3.1 Release Notes].
 endif::[]
+
+[[release-versioning-policy]]
+== Versioning Policy
+{product-title} does not have non-backwards compatible changes and, therefore,
+does not drop features on versions earlier than v3. Some upgrades may require
+changes by the administrator. Clients will continue to work against major
+version servers. For example, a 3.1 client will work against 3.0, 3.1, and 3.2.
```
adellape: 3.0, 3.1, and 3.2 servers

adellape:
@ncdc @fabianofranz Do you have any feedback on the server/client versioning policy being proposed for docs here?

ncdc:
Looks reasonable. cc @deads2k @smarterclayton

vikram-redhat:
@sferich888 - can you have a look as well?

sferich888:
I think your trying to say the following: 
- Not all features are available in 3.1 clients against 3.2 servers.
+Note: all features available in 3.1 clients will work against 3.2 servers. However new 3.2 features will not be callable ot accessible from 3.1 clients.

But other than this it looks good.

release_notes/index.adoc
deads2k: I think it is trying to express:
 1. a 3.2 server may have additional capabilities that a 3.1 client cannot use.
 2. a 3.2 client may have additional capabilities that are not supported by a 3.1 server.

deads2k:
> @ncdc @fabianofranz Do you have any feedback on the server/client versioning policy being proposed for OSE docs here?

The policy seems reasonable to me.

release_notes/index.adoc
akostadinov: Makes sense, IMO good to elaborate on that in the text as it is not clear for me.

sferich888: What exactly in the text is not clear? What needs further elaboration?

akostadinov: text only say one part "a 3.2 server may have additional capabilities that a 3.1 client cannot use.", the other part missing: "a 3.2 client may have additional capabilities that are not supported by a 3.1 server."

sferich888: So what your looking for is a compatibility matrix:

X.____ | X.Y | X.Y+N | CLIENTS
X.Y__  |  [0]  | [2]
X.Y+N |  [1]  | [0]
Server

[0] Compatible
[1] Client may not be able to access server features
[2] Client may provide options and/features that may not be compatible with
the server being accessed.

release_notes/index.adoc
```diff
@@ -20,3 +20,11 @@ technical notes, see the
 https://docs.openshift.com/enterprise/3.1/release_notes/index.html[OpenShift
 Enterprise 3.1 Release Notes].
 endif::[]
+
+[[release-versioning-policy]]
+== Versioning Policy
+{product-title} does not have non-backwards compatible changes and, therefore,
```
smarterclayton: double negative, suggest  `product provides strict backwards compatibility guarantees for all supported APIs (excluding alpha APIs, which may be changed without notice, and beta APIs, which may occasionally be changed in a non-backwards compatible manner) and will continue to support older clients against newer servers.  Changes of APIs for non-security related reasons will involve at minimum two minor releases (3.1 -> 3.2 -> 3.3, for example) to allow older clients to update.  Utilizing new capabilities may require newer clients.

akostadinov: I'm not against your explanation. But what is the reason to have at least 3 minor releases? Should we mention bugfix releases anywhere (3.2.1 for example)?
Wouldn't it be easier to just adopt http://semver.org ? I don't see it having stricter requirement from what is written above. Bumping version for non-functionality reasons would still be compatible with semver I think. At least would not break user expectations too much.

ahardin-rh:
Updated per suggestions. Thanks to all!

ahardin-rh:
Build for review: http://file.rdu.redhat.com/~ahardin/081216/versioning-policy/release_notes/

release_notes/index.adoc
```diff
@@ -20,3 +20,41 @@ technical notes, see the
...
+additional capabilities that are not supported by a 3.1 server.
+
+.Compatibility Matrix
+
+[cols="1,1,1"]
+|===
+
+|
+|*X.Y* (Client)
+|*X.Y+N* footnoteref:[versionpolicyn,Where *N* is 1 or 2.] (Client)
```
sferich888: This should read, `Where *N* is some number grater than 1` 

However with that said, the matrix only really works if your within 1-2 versions.

ahardin-rh:
@sferich888 Thanks! This is updated.

ahardin-rh:
[rev_history]
|xref:../release_notes/index.adoc#release-notes-index[Overview]
|Added a xref:../release_notes/index.adoc#release-versioning-policy[Versioning Policy].
%

bfallonf:
@ahardin-rh I moved this to future release. If it's for 3.3 I'm guessing that's right?

PR #2508, opened 2016-07-15
BZ 1349294 -- describe EFK behavior in the Kibana interface

tnguyen-rh:
Fixes bug 1348294 [link](https://bugzilla.redhat.com/show_bug.cgi?id=1348294).

This is WIP because it only addresses [comment 0](https://bugzilla.redhat.com/show_bug.cgi?id=1348294#c0).  The suggestions in [comment 2](https://bugzilla.redhat.com/show_bug.cgi?id=1348294#c2) will be addressed.

@thoraxe PTAL, especially https://github.com/openshift/openshift-docs/commit/001f9e8a671f38ea4cad117dfc3e4ab5372c64d7 -- is that heading in the right direction?

### DIFF
--- a/install_config/aggregate_logging.adoc
+++ b/install_config/aggregate_logging.adoc
@@ -328,6 +328,13 @@ Running the deployer creates a deployer pod and prints its name.
 . Wait until the pod is running.
 This can take up to a few minutes to retrieve the deployer image
 from the registry.
++
+In the Kibana interface, you can watch for the aggregation of
+*default*, *openshift* and *openshift-infra* projects into
+into an overarching project called *.operations*.
++
+NOTE: The *logging* project is not aggregated into *.operations*.
++
 In a terminal, you can watch its progress with:
 +
 ----
--- a/install_config/aggregate_logging.adoc
+++ b/install_config/aggregate_logging.adoc
@@ -230,9 +230,9 @@ The EFK stack is deployed xref:../dev_guide/templates.adoc#dev-guide-templates[u
 ====
 ----
 $ oc new-app logging-deployer-template \
-             --param KIBANA_HOSTNAME=kibana.example.com \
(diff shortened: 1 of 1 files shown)


install_config/aggregate_logging.adoc
```diff
@@ -323,9 +323,19 @@ ifdef::openshift-enterprise[]
 endif::openshift-enterprise[]
 |===
 +
-Running the deployer creates a deployer pod and prints its name. Wait until the
-pod is running. This can take up to a few minutes to retrieve the deployer image
-from the registry. You can watch its process with:
+Running the deployer creates a deployer pod and prints its name.
+
+. Wait until the pod is running.
+This can take up to a few minutes to retrieve the deployer image
```
thoraxe: Strange construction. Perhaps "It may take several minutes for OpenShift to retrieve..."

install_config/aggregate_logging.adoc
```diff
@@ -323,9 +323,19 @@ ifdef::openshift-enterprise[]
...
+from the registry.
++
+In the Kibana interface, you can watch for the aggregation of
```
thoraxe: This paragraph is also a little strange. Perhaps the following:

The logs for the default, openshift, and openshift-infra projects are automatically aggregated. These aggregated logs can be found in the ".operations" item in the Kibana interface. These projects are not found individually in the Kibana interface.

The project where you have deployed the EFK stack (eg: logging, as documented here) is **NOT** aggregated into .operations. It will be found under its ID.

bfallonf:
@thoraxe Thanks for the suggestions. I'll merge this and do a followup PR. I'll tag you in it, but your suggestions above are pretty straight-forward, so if there's nothing else, I'll move forward with it. Thanks again!

PR #2523, opened 2016-07-20
Move metrics to 'openshift-metrics' project

sdodson:
While preparing the installer for automated installation of metrics components in 1.3/3.3 I ran into the fact that metrics were being deployed into `openshift-infra`. That project is a special project containing many highly privileged components and we should avoid using it. Switch docs to use `openshift-metrics` instead, I'm open to suggestions for a better project name. Logging uses `logging` but i sort of like the `openshift-` prefix.

### DIFF
--- a/install_config/cluster_metrics.adoc
+++ b/install_config/cluster_metrics.adoc
@@ -65,16 +65,16 @@ address of the node, then Heapster will fail to retrieve any metrics.
 ====
 endif::[]
 
-The components for cluster metrics must be deployed to the *openshift-infra*
+The components for cluster metrics must be deployed to the *openshift-metrics*
 project. This allows xref:../dev_guide/pod_autoscaling.adoc#dev-guide-pod-autoscaling[horizontal pod
 autoscalers] to discover the Heapster service and use it to retrieve metrics
 that can be used for autoscaling.
 
 All of the following commands in this topic must be executed under the
-*openshift-infra* project. To switch to the *openshift-infra* project:
+*openshift-metrics* project. To switch to the *openshift-metrics* project:
 
 ----
-$ oc project openshift-infra
+$ oc project openshift-metrics
 ----
 
 To enable cluster metrics, you must next configure the following:
@@ -112,11 +112,11 @@ API
 ----
 
 . Before it can deploy components, the *metrics-deployer* service account must
-also be granted the `edit` permission for the *openshift-infra* project:
+also be granted the `edit` permission for the *openshift-metrics* project:
 +
 ----
 $ oadm policy add-role-to-user \
-    edit system:serviceaccount:openshift-infra:metrics-deployer
+    edit system:serviceaccount:openshift-metrics:metrics-deployer
 ----
 
 [[heapster-service-account]]
@@ -128,7 +128,7 @@ this, the Heapster service account requires the `cluster-reader` permission:
 
 ----
 $ oadm policy add-cluster-role-to-user \
-    cluster-reader system:serviceaccount:openshift-infra:heapster
+    cluster-reader system:serviceaccount:openshift-metrics:heapster
 ----
 
 [NOTE]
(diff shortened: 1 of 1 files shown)


sdodson:
@deads2k @mwringe -- thoughts?

deads2k:
I'd like to move external components out of `openshift-infra`.  That namespace holds credentials used by the core infrastructure of openshift (controllers baked into the core product).  I don't think that I would assume that the people maintaining metrics and granting access/routes/etc to it would want to be working in the same namespace as their core infrastructure "mess this up and your cluster falls over until you restart" components.

@liggitt Something you'd also be concerned about or would you just assume that cluster-admins would always be very careful and never want to "just nuke this metrics thing and start over", etc.

sdodson:
Actually, the first paragraph outlines why this is so. Perhaps this is something we must live with?
`This allows horizontal pod autoscalers to discover the Heapster service and use it to retrieve metrics
 that can be used for autoscaling.`

mwringe:
@sdodson no can do, it has to be `openshift-infra` due to the HPA (horizontal pod autoscaler). The HPA is hardcoded to look for it under `openshift-infra`. If it wasn't for the HPA, the metric components can be deployed anywhere (and I agree it should be under some other namespace)

@DirectXMan12 Is there anything we can do to allow the metrics components to be deployed somewhere else?

liggitt:
c.f. https://github.com/openshift/origin/issues/5781 for making the namespace configurable

install_config/cluster_metrics.adoc
```diff
@@ -112,11 +112,11 @@ API
 ----
 
 . Before it can deploy components, the *metrics-deployer* service account must
-also be granted the `edit` permission for the *openshift-infra* project:
+also be granted the `edit` permission for the *openshift-metrics* project:
```
deads2k: This is exactly the sort of thing we're worried about.  This just granted root access to the cluster to the deployer pod installing metrics.  build controller sa -> privileged pods -> pod creation -> docker socket access -> node secrets on disk.

That permission can be revoked later on (its not today), but there's no real reason for it need that access.

DirectXMan12: I think that's a quirk of how the metrics deployment process works -- it's not an inherent need of Heapster.

mwringe: How does giving edit permission to a service account for a specific namespace give it root access to the whole cluster? Or is this just because of how the permissions are configured for the `openshift-infra` project?

DirectXMan12:
I'm not against moving the namespace and/or making the it configurable, but I do think we should keep in mind that Heapster is a dependency of the horizontal pod autoscaler controller, and will eventually be a dependency of idling, so it's at least partially a dependency for having a working cluster.

IMO, it should probably be in a configurable location, *but* the default should be something that implies that it's a thing that's actually needed by cluster components.  I think I originally stuck it in `openshift-infra` because in upstream Kubernetes, they look for it in `kube-system` (the names seemed analogous, and nobody questioned it at the time ;-)).

install_config/cluster_metrics.adoc
liggitt: transitively, because openshift-infra contains many service accounts with elevated permissions, and the edit role in that namespace allows reading all of their secrets

deads2k: > I think that's a quirk of how the metrics deployment process works -- it's not an inherent need of Heapster.

It's inherent for anything with edit permission in this particular namespace.  This isn't a problem in any other namespace.

DirectXMan12: Fair.  Heapster itself should not need any edit permissions anywhere, though.

sdodson:
Should we close this PR and create a card for supporting metrics out of its own namespace?

ahardin-rh:
@sdodson Should I officially close this?

sdodson:
yeah

PR #2540, opened 2016-07-22
Bug 1351597 : added section on executing commands in a container

bfallonf:
As per: https://bugzilla.redhat.com/show_bug.cgi?id=1351597

@miheer FYI

### DIFF
--- a/dev_guide/deployments.adoc
+++ b/dev_guide/deployments.adoc
@@ -191,6 +191,55 @@ To see what the rollback would look like without performing the rollback:
 $ oc rollback <deployment_config> --dry-run
 ----
 
+[[executing-commands-inside-a-container-deployments]]
+== Executing Commands Inside a Container
+
+You can add a command to a container, which modifies the container's startup
+behavior by overruling the image's `ENTRYPOINT`. This is different from a
+xref:../dev_guide/deployments.html#pod-based-lifecycle-hook[lifecycle hook],
+which instead can be run once per deployment at a specified time.
+
+Add the `*command*` parameters to the `*spec*` field of the deployment
+configuration. You can also add an `*args*` field, which modifies the
+`*command*` (or the `ENTRYPOINT` if `*command*` does not exist).
+
+====
+----
+...
+spec:
+  containers:
+    -
+    name: <container_name>
+    image: 'image'
+    command:
+      - '<command>'
+    args:
+      - '<argument_1>'
+      - '<argument_2>'
+      - '<argument_3>'
+...
+----
+====
+
+For example, to execute the `java` command with the '-jar' and '/opt/app-root/springboots2idemo.jar' arguments:
+
+====
+----
+...
+spec:
+  containers:
+    -
+    name: example-spring-boot
+    image: 'image'
+    command:
+      - java
+    args:
+      - '-jar'
+      - /opt/app-root/springboots2idemo.jar
+...
+----
+====
+
 [[viewing-deployment-logs]]
 
 == Viewing Deployment Logs


adellape:
@bparees Should this ability to run commands in containers via `spec.containers` be contrasted with https://docs.openshift.com/enterprise/3.2/dev_guide/deployments.html#pod-based-lifecycle-hook somehow, as far as when/why you might choose to use one method versus the other? Or am I misunderstanding / inventing similarities here... ?

bparees:
in this case you're actually changing the primary command the container is going to run when the pod starts.

w/ hooks you're just defining additional commands to be run at other points in the lifecycle (like something to initialize a DB before the app starts)

so i don't think you'd likely choose one or the other.

One key difference is the hook runs once per deployment, whereas the container command runs once per container that starts (ie once in each replica)

bfallonf:
Thanks @bparees for that info, and @adellape for roping him in.

@bparees I've made an update to the PR. Can I get an ack that it's all correct? Thanks again.

dev_guide/deployments.adoc
```diff
@@ -191,6 +191,52 @@ To see what the rollback would look like without performing the rollback:
 $ oc rollback <deployment_config> --dry-run
 ----
 
+[[executing-commands-inside-a-container-deployments]]
+== Executing Commands Inside a Container
+
+You can add a command to a container, which it runs every time the container
+starts. This is different to a
```
bparees: "different from a"

dev_guide/deployments.adoc
```diff
@@ -191,6 +191,52 @@ To see what the rollback would look like without performing the rollback:
 $ oc rollback <deployment_config> --dry-run
 ----
 
+[[executing-commands-inside-a-container-deployments]]
+== Executing Commands Inside a Container
+
+You can add a command to a container, which it runs every time the container
```
bparees: it's really more than adding a command to the container.  The command field replaces the image's entrypoint, so whatever startup behavior the image would normally have, is superceded by this value.

Similarly the "args" field will be passed as arguments to the ENTRYPOINT if the image has one (or the command if that field is being specified).  So again, changes the behavior of the container entirely, replacing the normal behavior.  If the image has no ENTRYPOINT and the "command" field is not being specified, then the "args" will be invoked directly, i believe (rather than being passed as arguments to the entrypoint/command value).

bfallonf:
Thanks @bparees for the read-through. I've made some changes. Let me know if there's anything more you think needs to be there.

In the meantime, @tpoitras @ahardin-rh @tnguyen-rh peer review please?

bparees:
lgtm

dev_guide/deployments.adoc
```diff
@@ -191,6 +191,53 @@ To see what the rollback would look like without performing the rollback:
 $ oc rollback <deployment_config> --dry-run
 ----
 
+[[executing-commands-inside-a-container-deployments]]
+== Executing Commands Inside a Container
```
tnguyen-rh: I suppose the discrepency between the anchor id and the title is due to the topic title being appended, to avoid potential conflict when other sections w/ identical title are created.  But, i wonder, is that possibility likely?

dev_guide/deployments.adoc
```diff
@@ -191,6 +191,53 @@ To see what the rollback would look like without performing the rollback:
...
+
+You can add a command to a container, which modifies the container's startup
```
tnguyen-rh: Comma unnecessary.

dev_guide/deployments.adoc
```diff
@@ -191,6 +191,53 @@ To see what the rollback would look like without performing the rollback:
...
+...
+spec:
+  containers:
+    -
+    name: <container_name>
+    image: 'image'
+    command:
+      - '<command>'
+    args:
+      - '<arguments>'
```
tnguyen-rh: Having `<arguments>` (plural) as the single list element might lead the reader to put all the arguments on one line.  Maybe we could reformulate the `args` value to be a list of three elements: `<argument_1>`, `<argument_2>`, and `...`.

tnguyen-rh:
@bfallonf One pondering and one suggestion, otherwise LGTM. :+1:

dev_guide/deployments.adoc
bfallonf: Yes, this is something I've been doing in order to prevent conflicts. Not sure if I've discussed it with anyone... It's probably not necessary in this case, but I'm nothing if not consistent :P

dev_guide/deployments.adoc
bfallonf: Hmm I'll disagree with this one. Getting rid of the comma would ask for a s/which/that, which isn't right here. It's the whole "independent/dependent clause" thing.

bfallonf:
@tnguyen-rh Thanks for the read! I fixed the example.

bfallonf:
[revhistory]
|xref:../dev_guide/deployments.adoc#executing-commands-inside-a-container-deployments[Deployments]
|Added xref:../dev_guide/deployments.adoc[Executing Commands Inside a Container] section.
%

dev_guide/deployments.adoc
tpoitras: I've been making mine pretty crazy unique, too.

tpoitras:
LGTM, @bfallonf

dev_guide/deployments.adoc
adellape: This is the first I've seen the title be appended at the end. I usually prepend it, e.g.:

```
[[deployments-executing-commands-inside-a-container]]
```

adellape:
> |xref:../dev_guide/deployments.adoc#[Deployments]

@bfallonf Remember the full anchor ID here.

PR #2317, opened 2016-06-20
Updated instances of OpenShift to {product-title} across Install Config

ahardin-rh:

PR #2541, opened 2016-07-22
Fixed formatting in the Configuring OpenStack Variables heading

ahardin-rh:
I mistakenly introduced the error in https://github.com/openshift/openshift-docs/pull/2317


ahardin-rh:
Minor edit. No revision history needed.

PR #2570, opened 2016-07-28
Bug 1353400 arranged and added to info on adding hosts

bfallonf:
As per: https://bugzilla.redhat.com/show_bug.cgi?id=1353400

I've taken out the "Adding nodes" section from both the quick and advanced install topics and put them into one topic here, which is a level up from them, as it's not an install task, it's a post-install task. I've also made the tasks (especially in the advanced install section) less specific about nodes, as it's the same process for nodes and masters. I've also fixed up the links.

@abutcher @smunilla can I ask for a quick look-over please? I'd hate to think I lost something in the translation.

cc @ahardin-rh


_build_cfg.yml
```diff
@@ -252,6 +252,8 @@ Topics:
   - Name: Master and Node Configuration
     File: master_node_configuration
     Distros: openshift-origin,openshift-enterprise
+  - Name: Adding Hosts to an Existing Install
+    File: adding_hosts_to_isntalls
```
adellape: Typo here is probably what's causing the build to fail.

install_config/adding_hosts_to_installs.adoc
```diff
@@ -0,0 +1,195 @@
+[[install-config-adding-hosts-to-installs]]
+= Adding Hosts to an Existing Install
```
adellape: I think we should stick with calling it an "Existing Cluster" instead of an "install". IBM Style Guide says to only use "install" as a verb, too. "Installation" is permitted as a noun, but I still think sticking with "cluster" is more specific, and (considering consistency) we have a lot of other spots that refer to it in that way currently.

install_config/adding_hosts_to_installs.adoc
```diff
@@ -0,0 +1,195 @@
...
+[[adding-nodes-or-reinstalling-quick]]
+== Adding Hosts Using the Quick Installer Tool
+
+If you used the quick install tool to install your {product-title} cluster, you
+can use the quick install tool to add a new node host to your existing cluster,
+or to reinstall the cluster entirely.
+
+[NOTE]
+====
+Currently, you can not add new master hosts if you used the quick install.
```
adellape: This makes it sound like you can never add new masters if you started w/ the quick install (which you could, if you instead switched to using the advanced method). I assume this is more specifically trying to say that you can't use the quick installer tool to add new masters.

adellape:
Some comments inline. Also:

> I've taken out the "Adding nodes" section from both the quick and advanced install topics and put them into one topic here, which is a level up from them, as it's not an install task, it's a post-install task.

I don't know that I agree with it being up a level. Yes, you can come back to this topic post-initial deployment of OSE, but you're still "installing" OSE on a new host (master or node). Turning these procedures into a new single topic seems appropriate at this point now that we're including masters, but how about keeping under the "Installing" subdir, and just put it at the end:

```
Installing
  Overview
  Prerequisites
  RPM vs Containerized
  Quick Installation
  Advanced Installation
  Disconnected Installation
  Deploying a Docker Registry
  Deploying a Router
  Adding Hosts to an Existing Cluster
```

_build_cfg.yml
bfallonf: :|

install_config/adding_hosts_to_installs.adoc
bfallonf: Hmm now I can't tell if the info from Andrew and Sam was one way or the other. I certainly intended that you can't add new masters if you've installed using the quick install. @abutcher @smunilla thoughts?

bfallonf:
> I don't know that I agree with it being up a level. Yes, you can come back to this topic post-initial deployment of OSE, but you're still "installing" OSE on a new host (master or node).

I'm...respectfully disagreeing. While technically, sure, you're "installing" on the new hosts, I think it's still more accurate to say you're adding to your existing cluster. Look at the quick install option:

"Do you want to (1) only add additional nodes or (2) perform a clean install?:"

You're adding nodes.

And the method for the advanced install is using the "scaleup" file, which I feel is technically the same.

@vikram-redhat @tpoitras @ahardin-rh @tnguyen-rh Thoughts?

smunilla:
@bfallonf I don't know how much this helps, but my future plans for the quick installer include pulling the 'scaleup' workflow out of the install subcommand into its own subcommand. Users have reported in the past that this functionality being tucked under 'install' is a little confusing. See BZ#1339621 .

adellape:
@bfallonf @smunilla Fair enough. The target new hosts will still need to have gone through all the steps outlined in [Prerequisties: Host Preparation](https://docs.openshift.com/enterprise/latest/install_config/install/prerequisites.html#host-preparation) before running the tool/playbook, right? (Side thought: just noticed "Ensuring Host Access" should probably be a subsection of "Host Preparation".) I guess that was another reason I was thinking it belonged more in-line w/ the Installing subdir, but if we make that clear up front in this new topic and link appropriately, that should cover it?

IIRC, the installer yells at you if you don't have some of those prereqs taken care of, but still better to let them know here to make sure it's taken care of, versus being reprimanded by the tool.

vikram-redhat:
@ahardin-rh - hey, do you want to take over this PR while @bfallonf is on leave and finish it for this sprint? The reason I ask is that it is that duplicate of adding master and nodes that we have talked about. https://bugzilla.redhat.com/show_bug.cgi?id=1268973

ahardin-rh:
@vikram-redhat Sure! Working on it now.

adellape:
@bfallonf @vikram-redhat @ahardin-rh Continued and merged via https://github.com/openshift/openshift-docs/pull/2602.

PR #2571, opened 2016-07-28
Update "accessing hawkular metrics" section.

bbguimaraes:
- access is now officially supported, remove `openshift-origin` ifdef
- port docs from origin-metrics

PR #2579, opened 2016-07-29
Follow-up edits to PR #2571

ahardin-rh:
Minor follow-up edits to https://github.com/openshift/openshift-docs/pull/2571


ahardin-rh:
Very minor edits. No revision history needed.

PR #2582, opened 2016-07-29
added missing //s around include tags

ahardin-rh:


ahardin-rh:
minor edit; no revision history needed

PR #2598, opened 2016-08-02
Bug 1289441, added clarifying details in Providing Your Own Certificates

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1289441


ahardin-rh:
@rjhowe please review my first stab at this.

In the BZ you mention, "Then we add this step at the bottom showing how to provide your custom certs to a secret and then redeploying the metrics template."  What are the steps for that? Thanks!

vikram-redhat:
bump @rjhowe

install_config/cluster_metrics.adoc
```diff
@@ -268,6 +270,19 @@ xref:../install_config/cluster_metrics.adoc#metrics-reencrypting-route[re-encryp
...
+----
+--sink=hawkular:https://hawkular-metrics:443?tenant=_system&;....
+----
+
+To:
+
+----
+ --sink=hawkular:https://${CUSTOM_HOSTNAME}:443?tenant=_system&;....
+----
+
```
rjhowe: --sink is hard coded in the deployer image as the heapster.yaml template that is called. 

https://github.com/openshift/origin-metrics/blob/master/deployer/templates/heapster.yaml#L83

This would need to be changed in the metrics code and should not be in the docs until the variable is added.

rjhowe:
The flow of the documentation should point out that the preferred metrics deployment, would be to pass the metrics secret no certs.

$ oc secrets new metrics-deployer nothing=/dev/null


And use the a Re-encrypting Route to pass your custom certs to heapster. As this allow for greater control in modifying the certs in the future.

ahardin-rh:
@rjhowe I made some updates. Please let me know if I'm getting closer. Thanks!

install_config/cluster_metrics.adoc
```diff
@@ -319,6 +331,24 @@ properly.
...
+`--sink` is hard-coded in the deployer image as the *_heapster.yaml_* template
+is called. The
+link:https://github.com/kubernetes/heapster/blob/master/docs/sink-configuration.md#L5[sink]
+option is passed to the Heapster service that in running.
+
```
rjhowe: I would leave out everything around --sink option.  And just say that the "hawkular-metrics" URL is hard coded. 

This name will get the search domain attached and resolve to the service IP

ahardin-rh:
@rjhowe Updated PTAL. Thanks!

rjhowe:
Looks good.

ahardin-rh:
Thanks!

ahardin-rh:
[rev_history]
|xref:../install_config/cluster_metrics.adoc#install-config-cluster-metrics[Enabling Cluster Metrics]
|Added clarifying details to the xref:../install_config/cluster_metrics.adoc#metrics-using-secrets-byo-certs[Providing Your Own Certificates] section.
%

PR #2599, opened 2016-08-02
logging: EFK must avoid NFS

sosiouxme:
It came to our attention via
https://bugzilla.redhat.com/show_bug.cgi?id=1347666
and further research (
http://mail-archives.apache.org/mod_mbox/lucene-java-user/201210.mbox/%3C01a401cda09e$17b00160$47100420$@thetaphi.de%3E
and
https://lucene.apache.org/core/4_8_0/core/org/apache/lucene/store/NativeFSLockFactory.html
) that NFS is a not suitable for Lucene storage. This documents how to
use local storage, that NFS is not supported, and what to do if NFS is
all you have.

### DIFF
--- a/install_config/aggregate_logging.adoc
+++ b/install_config/aggregate_logging.adoc
@@ -416,24 +416,95 @@ The deployer creates an ephemeral deployment in which all of a pod's data is
 lost upon restart. For production usage, add a persistent storage volume to each
 Elasticsearch deployment configuration.
 
-The following example specifies a volume for an Elasticsearch replica (using a
-xref:../architecture/additional_concepts/storage.adoc#persistent-volume-claims[PersistentVolumeClaim]):
+The best-performing volumes are local disks, if it is possible to use
+them. Doing so requires some preparation as follows.
 
+. The relevant service account must be given the privilege to mount and edit a local volume, as follows:
++
 ====
 ----
-$ oc volume dc/logging-es-rca2m9u8 \
+$ oadm policy add-scc-to-user privileged  \
+       system:serviceaccount:logging:aggregated-logging-elasticsearch <1>
+----
+<1> Use the new project you created earlier (e.g., *logging*) when specifying
+this service account.
+====
+
+. Each Elasticsearch replica definition must be patched to claim that privilege, for example:
++
+----
+$ for dc in $(oc get deploymentconfig --selector logging-infra=elasticsearch -o name); do
+    oc scale $dc --replicas=0
+    oc patch $dc \
+       -p '{"spec":{"template":{"spec":{"containers":[{"name":"elasticsearch","securityContext":{"privileged": true}}]}}}}'
+  done
+----
+
+. The Elasticsearch pods must be located on the correct nodes to use
+the local storage, and should not move around even if those nodes are
+taken down for a period of time. This requires giving each Elasticsearch
+replica a node selector that is unique to the node where an administrator
+has allocated storage for it.  xref:#logging-node-selector[See below
+for directions on setting a node selector].
+
+. Once these steps are taken, a local host mount can be applied to each replica
+as in this example (where we assume storage is mounted at the same path on each node):
++
+ifdef::openshift-origin[]
+----
+$ for dc in $(oc get deploymentconfig --selector logging-infra=elasticsearch -o name); do
+    oc set volume $dc \
+          --add --overwrite --name=elasticsearch-storage \
+          --type=hostPath --path=/usr/local/es-storage
+    oc deploy --latest $dc
+    oc scale $dc --replicas=1
+  done
+----
+endif::openshift-origin[]
+ifdef::openshift-enterprise[]
+----
+$ for dc in $(oc get deploymentconfig --selector logging-infra=elasticsearch -o name); do
+    oc set volume $dc \
+          --add --overwrite --name=elasticsearch-storage \
+          --type=hostPath --path=/usr/local/es-storage
+    oc scale $dc --replicas=1
+  done
+----
+endif::openshift-enterprise[]
+
+If using host mounts is impractical or
+undesirable, it may be necessary to attach block storage as a
+xref:../architecture/additional_concepts/storage.adoc#persistent-volume-claims[PersistentVolumeClaim])
+as in the following example:
+
+----
+$ oc set volume dc/logging-es-<unique> \
           --add --overwrite --name=elasticsearch-storage \
           --type=persistentVolumeClaim --claim-name=logging-es-1
 ----
+
+[WARNING]
+====
+Using NFS storage directly or as a PersistentVolume (or via other NAS
+such as Gluster) is not supported for Elasticsearch storage, as Lucene
+relies on filesystem behavior that NFS does not supply. Data corruption
+and other problems can occur. If NFS storage is a requirement, you can
+allocate a large file on that storage to serve as a storage device and
+treat it as a host mount on each host. For example:
+
+----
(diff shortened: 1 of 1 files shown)


sosiouxme:
@ewolinetz @richm PTAL

@adellape or @ahardin-rh this is relatively high priority per the bug. It's relevant to origin and OSE 3.1+

sosiouxme:
heck /cc @thoraxe too.

install_config/aggregate_logging.adoc
```diff
@@ -416,24 +416,84 @@ The deployer creates an ephemeral deployment in which all of a pod's data is
...
 ----
-$ oc volume dc/logging-es-rca2m9u8 \
+$ oadm policy add-scc-to-user privileged  \
+       system:serviceaccount:logging:aggregated-logging-elasticsearch <1>
+----
+<1> Use the new project you created earlier (e.g., *logging*) when specifying
+this service account.
+====
+
+. Each Elasticsearch replica definition must be patched to claim that privilege, for example:
```
ewolinetz: Should probably remind users to stop their cluster first

ewolinetz:
oi... do we forsee having a flag to do this for users with the deployer?

sosiouxme:
I hadn't thought much about deployer parameters. It might be difficult to specify individual nodeselectors for each instance. But it probably wouldn't be hard to patch in the local mounts and the privileged security context.

BTW, need to re-examine whether there's a way short of "privileged" that gets us past the SELinux problem with local mounts.

install_config/aggregate_logging.adoc
sosiouxme: I suppose so. I figured they were gonna lose any ephemeral data anyway...

install_config/aggregate_logging.adoc
```diff
@@ -416,24 +416,84 @@ The deployer creates an ephemeral deployment in which all of a pod's data is
...
-xref:../architecture/additional_concepts/storage.adoc#persistent-volume-claims[PersistentVolumeClaim]):
+The best-performing volumes are local disks, if it is possible to use
+them. There are some complications with doing so.
 
+. The relevant service account must be given the privilege to mount and edit a local volume, as follows:
++
 ====
 ----
-$ oc volume dc/logging-es-rca2m9u8 \
+$ oadm policy add-scc-to-user privileged  \
```
pweil-: do you have to access to the privileged SCC here or will `hostmount-anyuid` (which does not allow privileged) be enough?

sferich888:
Should we also warn customers to avoid gluster (which uses NFS on the backend)?

install_config/aggregate_logging.adoc
sosiouxme: @pweil- I tried `hostmount-anyuid` first and it did not have access due to SELinux context. I believe it's much the same problem we had with fluentd - https://github.com/openshift/origin-aggregated-logging/issues/89#issuecomment-204127737

It seems like less-than-privileged may be possible, but I'm not quite sure how and it seems like it would be a PITA for a user to set up. What do you think?

pweil-: Hey, whaddya know... https://github.com/openshift/origin/issues/8504

sosiouxme: I'm a little foggy on whether exactly the same fix will apply. The problem with fluentd was that it was trying to read and write in /var/log. Here we're trying to read and write in an admin-supplied storage volume; I suppose we could have them chcon the volume to whatever would be convenient? If so, what would that be - is there a context that will allow read/write for any context the pod may be running in?

sosiouxme:
Think I've addressed existing concerns... any further?

install_config/aggregate_logging.adoc
pweil-: The kubelet (when a pod is using host namespaces) or docker should be performing a relabeling of the volume when it can.  It uses the docker opts to pass in the selinux context that is being used.  If that isn't working or this is a different use case then we can figure out what is different.  cc @pmorie who is very familiar with the selinux code for volumes

sosiouxme: What the AVC looks like, FYI:

```
type=AVC msg=audit(1470323991.042:27487): avc:  denied  { write } for  pid=9883 comm="java" name="es-storage" dev="dm-
0" ino=68862303 scontext=system_u:system_r:svirt_lxc_net_t:s0:c2,c8 tcontext=unconfined_u:object_r:usr_t:s0 tclass=dir
type=SYSCALL msg=audit(1470323991.042:27487): arch=c000003e syscall=83 success=no exit=-13 a0=7ff7d43d3780 a1=1ff a2=7
ff7d43d3780 a3=7ff7c47bd728 items=0 ppid=15669 pid=9883 auid=4294967295 uid=1000 gid=0 euid=1000 suid=1000 fsuid=1000 
egid=0 sgid=0 fsgid=0 tty=(none) ses=4294967295 comm="java" exe="/usr/lib/jvm/java-1.8.0-openjdk-1.8.0.101-3.b13.el7_2
.x86_64/jre/bin/java" subj=system_u:system_r:svirt_lxc_net_t:s0:c2,c8 key=(null)
```

install_config/aggregate_logging.adoc
```diff
@@ -416,24 +416,82 @@ The deployer creates an ephemeral deployment in which all of a pod's data is
 lost upon restart. For production usage, add a persistent storage volume to each
 Elasticsearch deployment configuration.
 
-The following example specifies a volume for an Elasticsearch replica (using a
-xref:../architecture/additional_concepts/storage.adoc#persistent-volume-claims[PersistentVolumeClaim]):
+The best-performing volumes are local disks, if it is possible to use
+them. There are some complications with doing so.
```
ahardin-rh: what is an example complication?

install_config/aggregate_logging.adoc
```diff
@@ -416,24 +416,82 @@ The deployer creates an ephemeral deployment in which all of a pod's data is
...
+treat it as a host mount on each host. For example:
+
+----
+$ truncate -s 1T /nfs/storage/elasticsearch-1
+$ mkfs.xfs /nfs/storage/elasticsearch-1
+$ mount -o loop /nfs/storage/elasticsearch-1 /usr/local/es-storage
+$ chown 1000:1000 /usr/local/es-storage
+----
+
+You can then proceed to use `/usr/local/es-storage` as a host-mount as
```
ahardin-rh: s/You can then proceed to use/Then, use...

Use format:

    *_/usr/local/es-storage_*

install_config/aggregate_logging.adoc
```diff
@@ -416,24 +416,82 @@ The deployer creates an ephemeral deployment in which all of a pod's data is
...
+described above. Performance under this solution will be significantly
```
ahardin-rh: s/will be/is

ahardin-rh:
@sosiouxme just a few minor comments from me. Thanks!

install_config/aggregate_logging.adoc
sosiouxme: The points that follow are the complications. Perhaps I should call them
something else... considerations?

sosiouxme:
@ahardin-rh think I addressed your comments now.

ahardin-rh:
@sosiouxme Looks good!  Thanks! Just a squash and we're good to go :beers:

sosiouxme:
ready, then.

ahardin-rh:
[rev_history]
|Added that NFS is a not suitable for Lucene storage, NFS is not supported, and how to
use local storage.
%

PR #2604, opened 2016-08-04
Removed Azure from list of storage drivers

ahardin-rh:
Not currently claiming support for Azure per Mike Barrett and Xiaoli Tran.


adellape:
:+1:

PR #2629, opened 2016-08-10
Get the right haproxy-config.template

ghost:
The shell command line example for enterprise used a hardcoded version.
With this PR the haproxy template will be used from the same router image version as the oc version is in use.


cw-aleks:
closed without commit

PR #2631, opened 2016-08-10
Document OAuth scopes

liggitt:


ahardin-rh:
Continuing this work in https://github.com/openshift/openshift-docs/pull/2859  Thanks!

PR #2641, opened 2016-08-11
Enabling tech preview features in the console

jwforres:
tech_preview_doc

### DIFF
--- a/install_config/web_console_customization.adoc
+++ b/install_config/web_console_customization.adoc
@@ -174,6 +174,36 @@ assetConfig:
 ====
 endif::[]
 
+[[enable-tech-preview-feature]]
+=== Enabling Features in Tech Preview
+
+Sometimes features will be available in tech preview, by default these features may be hidden
+from end users.  
+
+The current list of features in tech preview is:
+
+* pipelines - Enabling this will add the Pipelines navigation item underneath the Builds menu
+
+To enable a feature:
+
+====
+----
+window.OPENSHIFT_CONSTANTS.ENABLE_TECH_PREVIEW_FEATURE.feature_name = true;
+----
+====
+
+Save this script to a file (for example, *_tech-preview.js_*) and add it to the
+master configuration file:
+
+====
+----
+assetConfig:
+  ...
+  extensionScripts:
+    - /path/to/tech-preview.js
+----
+====
+
 [[serving-static-files]]
 == Serving Static Files
 


jwforres:
@spadgett @bparees @danmcp @smarterclayton

bparees:
lgtm

adellape:
@jwforres Is this tech preview for both Origin and OCP (3.3)?

jwforres:
@adellape yes

vikram-redhat:
@jwforres - Is this the only tech preview feature for the next release?

@adellape, @bfallonf - I don't think this is the right file to note the tech preview features or perhaps the title could be changed to reflect what is being enabled here?

@bfallonf - can you please peer review?

install_config/web_console_customization.adoc
```diff
@@ -174,6 +174,36 @@ assetConfig:
...
 
+[[enable-tech-preview-feature]]
+=== Enabling Features in Tech Preview
+
+Sometimes features will be available in tech preview, by default these features may be hidden
+from end users.  
+
+The current list of features in tech preview is:
+
+* pipelines - Enabling this will add the Pipelines navigation item underneath the Builds menu
```
bfallonf: I'm not sure if this is the right place to list the features currently in tech preview. We do have a place for this in the release notes:

https://docs.openshift.com/enterprise/3.2/release_notes/ose_3_2_release_notes.html#ose-32-technology-preview

If the current features are not accurate this should probably be changed.

install_config/web_console_customization.adoc
```diff
@@ -174,6 +174,36 @@ assetConfig:
...
+
+To enable a feature:
+
+====
+----
+window.OPENSHIFT_CONSTANTS.ENABLE_TECH_PREVIEW_FEATURE.feature_name = true;
+----
+====
```
bfallonf: I'm not sure what to do with this. Judging by the next step, this is a script? Do they just throw this into the command line? Is this supposed to be in an edit in a file somewhere?

install_config/web_console_customization.adoc
```diff
@@ -174,6 +174,36 @@ assetConfig:
...
+
+====
+----
+assetConfig:
+  ...
+  extensionScripts:
+    - /path/to/tech-preview.js
+----
+====
+
```
bfallonf: I also think the above could be a two step procedure.

bfallonf:
@jwforres A few questions, but looks good overall!

install_config/web_console_customization.adoc
spadgett: @bfallonf It's not just listing the features. It's telling you the property names you need to use in this config to turn them on. (We can adjust the wording here.)

Agree it should be added to the release notes.

bfallonf: @spadgett @jwforres Wait, is the reader supposed to replace (in this example ) "pipelines" where it says "feature_name" in the below example? If that's the case, then that definitely needs to be clarified.

spadgett: > Wait, is the reader supposed to replace (in this example ) "pipelines" where it says "feature_name" in the below example?

Yes, that's right

ahardin-rh:
Continuing work in https://github.com/openshift/openshift-docs/pull/2936

PR #2646, opened 2016-08-12
Customization top, primary, and secondary navigation in console

jwforres:


jwforres:
@spadgett could use a technical review / do we want to explicitly call all these options out in the docs?

install_config/web_console_customization.adoc
```diff
@@ -172,6 +172,131 @@ assetConfig:
...
+==== Project Left Navigation
+
+When you are navigating within a project, a menu appears on the left with primary and secondary navigation.
+This menu structure is defined as a constant and can be overridden or modified.
+
+[NOTE]
+====
+Significant customizations to the project navigation may affect the user experience and should be done with
+careful consideration. You will need to update this customization in future upgrades to guarantee you are picking
+up any new pages that are available.
```
spadgett: This is a good point. I wonder if we can't have an example that adds something to the existing nav object without replacing the entire menu. Is it asking administrators to write too much code?

```js
window.OPENSHIFT_CONSTANTS.PROJECT_NAVIGATION.push({
  // my primary nav menu
});
```

or

```js
var applicationsMenu = _.find(window.OPENSHIFT_CONSTANTS.PROJECT_NAVIGATION, { label: 'Applications' });
applicationsMenu.secondaryNavSections.push({
  // my menu item
});
```

spadgett: If we ask them to replace the entire object, we'd need to give them some link to constants.js with the menu for the version of OpenShift they have. The approach above would avoid that.

jwforres: Yeah I debated doing that as well. Probably better since it promotes smaller changes.  I think the caution note still applies though since they could still overwrite the whole thing or remove items.

spadgett: But where do they get the entire menu? They could look at the origin-web-console repo I guess, but they'd probably want the tag for their release and not whatever is in master.

jwforres:
Updated the examples for the primary nav

install_config/web_console_customization.adoc
```diff
@@ -172,6 +172,142 @@ assetConfig:
...
+window.OPENSHIFT_CONSTANTS.PROJECT_NAVIGATION.push({
+  label: "Overview",            // The text label
+  iconClass: "fa fa-dashboard", // The icon you want to appear
+  href: "/overview"             // Where to go when this nav item is clicked
+});
+
+// Splice a primary nav item to a specific spot in the list.  This primary item has
+// a secondary menu.
+window.OPENSHIFT_CONSTANTS.PROJECT_NAVIGATION.splice(2, 0, { // Insert at the third spot
+  label: "Applications",
```
spadgett: Is it confusing for the example to add a menu item that already exists? Looks like you might be modifying the existing one or adding a secondary item to it.

jwforres:
Updated again

install_config/web_console_customization.adoc
```diff
@@ -172,6 +172,144 @@ assetConfig:
...
+master configuration file:
+
+====
+----
+assetConfig:
+  ...
+  extensionScripts:
+    - /path/to/navigation.js
+----
+====
```
adellape: Then restart master, I assume?

adellape:
:+1:

install_config/web_console_customization.adoc
jwforres: yes

vikram-redhat:
@bfallonf - can you please follow up early next week?

PR #2674, opened 2016-08-18
Modifications from testing and review, including:

danmacpherson:
- Added basic description of the pods used in the example app
- Converted the Github webhook URL locations to a set of list items. Previously, this was instructions to use the URL on the post-creation page with an admonition for the location in the UI. However, due to the previous instructions users might have navigated away from the post-creation page (to Overview). It was easy to overlook the instructions for the location in the web UI.

### DIFF
--- a/getting_started/basic_walkthrough.adoc
+++ b/getting_started/basic_walkthrough.adoc
@@ -16,7 +16,11 @@ This guide demonstrates how to get a simple project up and running on
 
 The following sections guide you through creating a project that contains a
 sample Node.js application that will serve a welcome page and the current hit
-count (stored in a database) using the {product-title}  web console.
+count (stored in a database) using the {product-title}  web console. This 
+involves creating two xref:../architecture/core_concepts/pods_and_services.adoc#pods[pods]:
+
+- one to host the Node.js application
+- one to host the MongoDB database
 
 The tutorial assumes that you have:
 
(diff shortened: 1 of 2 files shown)


vikram-redhat:
@bfallonf - can you please peer review? This is for the online docs: https://docs.openshift.com/online/welcome/index.html and a follow on from the discussion about https://github.com/openshift/openshift-docs/pull/2119

getting_started/basic_walkthrough.adoc
```diff
@@ -103,27 +107,26 @@ MongoDB pod then starts up and displays its newly-assigned IP address.
 
 In this section, you will configure a GitHub webhook to automatically trigger a
 rebuild of your application whenever you push code changes to your forked
-repository.
+repository. This involves adding the Github webhook URL from your application
+into your Github repository. You obtain this webhook from these locations:
```
bfallonf: I'm not sure if the "You obatin this webhook from these locations:" bit is necessary. I know it's moving the bits below up to here, but in doing so I think it makes it unnecessary.

bfallonf:
@danmacpherson One comment from me. Other than that, this all looks fine :train:

getting_started/basic_walkthrough.adoc
danmacpherson: My reasoning was that when I tested it, I didn't know there was a second location to obtain the webhook URL (until I later scrolled down past the screenshot, I thought it was only obtainable from the post-creation page). So I ended up clicking back a couple of times to try and find the post-creation screen on my browser. It's not huge deal but it does save a minute of time knowing that the URL is available in multiple locations, which is why I think it's important to explicitly say that.

bfallonf: @danmacpherson OK. I looked at this more closely. You changed the steps to dot points. This makes sense. I'll allow it :P

bfallonf:
@vikram-redhat this looks good. Can be merged if you'd like.

vikram-redhat:
@adellape - FYI. Dan has also updated the related spreadsheet. I am gonna go ahead and merge it.

PR #2675, opened 2016-08-18
Limit Pod Access with Egress Policy

tpoitras:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1364773

Created a new section in the admin guide for this, since it seemed to be its own thing:

"Limit Pod Access with Egress Firewall"

Rendered docs: http://file.bne.redhat.com/tpoitras/2016/egress/openshift-enterprise/egress-BZ1364773/admin_guide/limit_pod_access_egress.html

### DIFF
--- a/admin_guide/limit_pod_access_egress.adoc
+++ b/admin_guide/limit_pod_access_egress.adoc
@@ -1,0 +1,104 @@
+[[admin-guide-limit-pod-access-egress]]
+= Limit Pod Access with Egress Firewall
+{product-author}
+{product-version}
+:data-uri:
+:icons:
+:experimental:
+:toc: macro
+:toc-title:
+
+toc::[]
+
+== Overview
+
+As an {product-title} cluster administrator, you can use egress policy to limit
+the addresses that some or all pods can access from within the cluster, so that:
+
+- A pod can only talk to internal hosts, and cannot initiate connections to the
+public Internet.
++
+Or, 
+- A pod can only talk to the public Internet, and cannot initiate connections to
+internal hosts (outside the cluster).
++
+Or,
+- A pod cannot reach specified internal subnets/hosts that it should have no
+reason to contact.
+
+For example, you can configure projects with different egress policies, allowing
+`<project A>` access to a specified IP range, but denying the same access to
+`<project B>`. 
+
+[CAUTION]
+====
+You must have the 
+xref:../install_config/configuring_sdn.adoc#install-config-configuring-sdn[*ovs-multitenant* plug-in] enabled in order to limit pod access via egress policy.
+====
+
+Project administrators can neither create `*EgressNetworkPolicy*` objects, nor
+edit the ones you create in their project. 
+
+The `default` project (and any other global namespace) cannot have egress
+policy. 
+
+[NOTE]
+====
+If you merge two projects together (via `oadm pod-network join-projects`), 
+then you cannot use egress policy in _any_ of the joined projects.
+
+If you make a project global (via `oadm pod-network make-projects-global`), 
+then it cannot have an `*EgressNetworkPolicy*`. 
+
+If an allowed network overlaps with a denied network, then the rules are 
+checked in order, and the first one that matches is enforced.
+====
+
+[[admin-guide-config-pod-access]]
+== Configuring Pod Access Limits
+
+
+To configure pod access limits, you must use the `oc` command or the REST API.
+You can use `oc [create|replace|delete]` to manipulate `*EgressNetworkPolicy*`
+objects. The *_api/swagger-spec/oapi-v1.json_* file has API-level details on how
+the objects actually work. 
+
+To configure pod access limits:
+
+. Navigate to the project you want to affect. 
+. Create a JSON file for the pod limit policy:
++
+----
+# oc create -f <policy>.json
+----
+. Configure the JSON file with policy details. For example:
++
+----
+{
+    "kind": "EgressNetworkPolicy",
+    "apiVersion": "v1",
+    "metadata": {
+        "name": "default"
+    },
+    "spec": {
+        "egress": [
+            {
+                "type": "Allow",
+                "to": {
+                    "cidrSelector": "1.2.3.0/24"
+                }
+            },
+            {
+                "type": "Deny",
+                "to": {
+                    "cidrSelector": "0.0.0.0/32"
+                }
+            }
+        ]
+    }
+}
+----
++
+When the example above is added in a project, it allows traffic to `1.2.3.0/24`,
+but denies access to all other external IP addresses. This would not affect
+traffic to other pods.
--- a/_build_cfg.yml
+++ b/_build_cfg.yml
@@ -441,6 +441,8 @@ Topics:
   - Name: Limit Run-once Pod Duration
     File: limit_runonce_pod_duration
     Distros: openshift-origin,openshift-enterprise
+  - Name: Limit Pod Access with Egress Firewall
+    File: limit_pod_access_egress
   - Name: Monitoring Routers
     File: router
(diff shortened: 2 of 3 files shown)


tpoitras:
ping @danwinship for tech review, please.

bmeng:
Should better add a point to explain that which takes higher priority when the allowed network overlaps with the denied network.

tpoitras:
@bmeng So, which one takes priority?

admin_guide/limit_pod_access_egress.adoc
```diff
@@ -0,0 +1,117 @@
...
+policy. 
+
+If you merge two projects together (via `oadm pod-network join-projects`), then you cannot use egress policy in _any_ of the joined projects.
+====
+
+== Examples and Restrictions
+
+- You can create `*EgressNetworkPolicy*` objects with rules that
+allow or deny outgoing pod traffic to specific CIDR ranges. For example: 
+`10.0.0.0/8`.
```
danwinship: This is redundant

admin_guide/limit_pod_access_egress.adoc
```diff
@@ -0,0 +1,117 @@
...
+
+- Different projects can have different `*EgressNetworkPolicies*`, allowing you 
+to allow `<project A>` access a specified IP range, but deny the same access to 
+`<project B>`.
```
danwinship: and this is pretty much implied by the earlier docs (or if you think it needs to be stated explicitly, it should be stated earlier)

admin_guide/limit_pod_access_egress.adoc
```diff
@@ -0,0 +1,117 @@
...
+
+- Different projects can have different `*EgressNetworkPolicies*`, allowing you 
+to allow `<project A>` access a specified IP range, but deny the same access to 
+`<project B>`. 
+
+- If you merge two projects together via `oadm pod-network join-projects`, then
+you cannot have an `*EgressNetworkPolicy*` object in either namespace. If you
+do, it logs an error and then adds a rule dropping all external traffic from
+both projects. This happens regardless of the order in which you merge the
+projects and create the policy object.
```
danwinship: The exact semantics of what happens when you try to join projects that use egress policy is undefined. Just say that you can't do it. (The Trello card specifies what happens in detail because we wanted QE to check that the expected thing happens.)

admin_guide/limit_pod_access_egress.adoc
```diff
@@ -0,0 +1,117 @@
...
+- If you make a project global via `oadm pod-network make-projects-global`, then
+it cannot have an `*EgressNetworkPolicy*`. If you do, it logs a warning and then
+ignores the policy. This happens regardless of the order in which you make the
+project global and create the policy object.
+
+- Do not create an `*EgressNetworkPolicy*` denying access to an IP address and
+create a Service that points to that IP with a manually-created Endpoints
+object. If you do, then {product-title} ignores the Endpoints, logs an error,
+and does not include the bad endpoint in the iptables rules generated for the
+service you created.
```
danwinship: I don't think this needs to be mentioned at all.

danwinship:
> Should better add a point to explain that which takes higher priority when the allowed network overlaps with the denied network.

The rules are checked in order, and the first one that matches wins.

admin_guide/limit_pod_access_egress.adoc
tpoitras: moved to intro

admin_guide/limit_pod_access_egress.adoc
tpoitras: TMI, got it. :smile:

tpoitras:
Thank you, @danwinship for the speedy tech review. I have incorporated your feedback! :smile: :wrench: 

Moving along to peer review.

tpoitras:
Ping @bfallonf @ahardin-rh @adellape for peer review, please and thank you.

admin_guide/limit_pod_access_egress.adoc
```diff
@@ -0,0 +1,104 @@
...
+- A pod can only talk to the public Internet, and cannot initiate connections to
+internal hosts (outside the cluster).
++
+Or,
+- A pod cannot reach specified internal subnets/hosts that it should have no
+reason to contact.
+
+For example, you can configure projects with different egress policies, allowing
+`<project A>` access to a specified IP range, but denying the same access to
+`<project B>`.
```
bfallonf: I don't think the "Or" things work too well here, plus, they don't fit with other bullet points we've donw in the rest of the docs. I think they could go, and the "so that" bit at the top could be more like "for three reasons". Let me know if that doesn't make sense.

admin_guide/limit_pod_access_egress.adoc
```diff
@@ -0,0 +1,104 @@
...
+`<project B>`. 
+
+[CAUTION]
+====
+You must have the 
+xref:../install_config/configuring_sdn.adoc#install-config-configuring-sdn[*ovs-multitenant* plug-in] enabled in order to limit pod access via egress policy.
+====
+
+Project administrators can neither create `*EgressNetworkPolicy*` objects nor
+edit the ones you create in their project.
```
bfallonf: Comma before 'nor'

admin_guide/limit_pod_access_egress.adoc
```diff
@@ -0,0 +1,104 @@
...
+====
+- If you merge two projects together (via `oadm pod-network join-projects`), 
+then you cannot use egress policy in _any_ of the joined projects.
+
+- If you make a project global (via `oadm pod-network make-projects-global`), 
+then it cannot have an `*EgressNetworkPolicy*`. 
+
+- If an allowed network overlaps with a denied network, then the rules are 
+checked in order, and the first one that matches is enforced.
+====
```
bfallonf: I'm not sure if the bullets above are needed. There's no 'introduction' sentence there to link the points. I think it works as three separate sentences, or paras if you want to keep it that way.

bfallonf:
@tpoitras Just a couple comments. LG otherwise.

admin_guide/limit_pod_access_egress.adoc
tpoitras: had paras before, made bullets. Will revert to paras.

tpoitras:
Thanks, @bfallonf :octocat: :zap: :octopus:

tpoitras:
[rev_history]
|xref:../admin_guide/limit_pod_access_egress.adoc[Cluster Administration -> Limit Pod Access with Egress Firewall]
|Added new topic covering how to limit the IP addresses and traffic that a pod can access.
%

cw-aleks:
As I have understand the renderd file right I can oly limit the **whole** project not dedicated pods, right.

When a customer want to define a dedicated node for egress, in the project, is this with this **EgressNetworkPolicy** and labels possible?

vikram-redhat:
@tpoitras - not sure if you can answer @cw-aleks question?

tpoitras:
@cw-aleks this specific documentation update only deals with setting limits per-project.

As far as I can tell, that is (currently) the only way to implement this. Tagging @danwinship for a definitive answer here.

danwinship:
Correct, this only does namespace/project-level filtering

cw-aleks:
Thanks.

This means that it is not possible to define a dedicated egress node which is different then the application/compute node, isn't it?

danwinship:
well, you can use `nodeSelector`s in the pod specs to control what pods go on what nodes if you want

cw-aleks:
A perfect. So the normal openshift handling with the `nodeSelector`s works. thanks for clarification.

tpoitras:
Thanks @danwinship :+1:

PR #2680, opened 2016-08-18
Fixed the docs on how to protect the router from packet drops

knobunc:
The previous docs were missing a piece for how to set up the
deployment config.  This corrects that.  As written, this will only
work with 3.3.  For 3.2 you need to set the environment variable to
'1' not 'true' and then it will work.

For bug 1269488


knobunc:
@DirectXMan12 @adellape PTAL

adellape:
Closing in favor of https://github.com/openshift/openshift-docs/pull/2680.

PR #2700, opened 2016-08-22
update Rakefile to reflect new script name

aweiteka:


ahardin-rh:
@aweiteka Resurrecting this. Is this still valid? If so, looks like we need a rebase. Thanks!

ahardin-rh:
@aweiteka Should this be closed? Thanks!

PR #2689, opened 2016-08-19
Add section "Backup and Restore" to the Cluster Administration Guide.

ahardin-rh:
This is a continuation of the work in https://github.com/openshift/openshift-docs/pull/2140

Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1259544

PR #2701, opened 2016-08-22
Added missing anchor to top of Backup and Restore topic

ahardin-rh:
Missed adding anchor in https://github.com/openshift/openshift-docs/pull/2689


ahardin-rh:
No rev history needed

PR #2719, opened 2016-08-24
Correct the figure path.

warmchang:
Correct the figure path.


ahardin-rh:
Thanks for the suggestion! I confirmed that the image appears correctly in the docs. images/ is not needed in the path.

PR #2697, opened 2016-08-22
Using Images: add chapter "Invoking scripts embedded in an image"

php-coder:
Add chapter about wrapping S2I scripts.

PTAL @openshift/team-documentation 
CC @bparees 
Trello: https://trello.com/c/FJ7ZVLhR/1016-3-document-and-investigate-s2i-exit-hooks

PR #2724, opened 2016-08-24
Follow-up edits to PR#2697

ahardin-rh:
Follow-up to https://github.com/openshift/openshift-docs/pull/2697

### DIFF
--- a/_build_cfg.yml
+++ b/_build_cfg.yml
@@ -386,7 +386,7 @@ Dir: admin_guide
 Distros: openshift-origin,openshift-enterprise,openshift-dedicated,atomic-*
 Topics:
   - Name: Overview
-    File: index   
+    File: index
   - Name: Managing Nodes
     File: manage_nodes
     Distros: openshift-origin,openshift-enterprise
@@ -615,8 +615,8 @@ Topics:
     Topics:
       - Name: Overview
         File: index
-      - Name: Using S2I images
-        File: customization
+      - Name: Customizing S2I images
+        File: customizing_s2i_images
       - Name: Node.js
         File: nodejs
       - Name: Ruby
--- a/using_images/s2i_images/customizing_s2i_images.adoc
+++ b/using_images/s2i_images/customizing_s2i_images.adoc
@@ -1,5 +1,5 @@
-[[using-s2i-images]]
-= Using S2I images
+[[customizing-s2i-images]]
+= Customizing S2I Images
 {product-author}
 {product-version}
 :data-uri:
@@ -18,7 +18,7 @@ but the default behavior of those scripts may not be suitable for all users.
 This topic covers a few approaches for customizing the behavior of an
 S2I builder that includes default scripts.
 
-[[using-S2I-images-invoking-scripts-embedded-in-an-image]]
+[[customizing-S2I-images-invoking-scripts-embedded-in-an-image]]
 == Invoking Scripts Embedded in an Image
 
 Typically, builder images provide their own version of the S2I scripts that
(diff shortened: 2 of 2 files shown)


ahardin-rh:
@adellape @tpoitras @bfallonf PTAL. Also, do you think this topic should live in the Creating Images or Using Images section? Thanks!

using_images/s2i_images/customizing_s2i_images.adoc
```diff
@@ -69,10 +70,9 @@ The example shows a custom *_assemble_* script that prints the message,
...
 depending on the exit code of the *_assemble_* script.
 
-Wrapping the *_run_* script is a bit tricky because we have to use `exec` for
-invoking it (see chapter
-xref:../../creating_images/guidelines.adoc#general-docker-guidelines[Use `exec` in
-Wrapper Scripts]). As a consequence, it is impossible to have a post-run step
+Wrapping the *_run_* script is a bit tricky because you have to
+xref:../../creating_images/guidelines.adoc#general-docker-guidelines[use `exec`
+for invoking it]. As a consequence, it is impossible to have a post-run step
 because the code after `exec /usr/libexec/s2i/run` never gets executed.
```
bfallonf: The above is verging on colloquialism and...kinda confusing. Maybe:

"You may have difficulties wrapping the *_run_* script using the `exec` command. As a consequence, a post-run step will also impossible(?), because..."

I use (?) because one thing being tricky leads to another being impossible? That's the confusing part I talk about...

cc: @bparees @php-coder

bfallonf:
@ahardin-rh One comment from me. cc'ed people might be able to elaborate...

using_images/s2i_images/customizing_s2i_images.adoc
bparees: one thing being tricky (running commands before the existing run script) leads to another thing being impossible (running things after the existing run script).  but i agree it can use a rephrasing.

better would be:
When wrappering the run script you must use exec for invoking it to ensure signals are handled properly.  Unfortunately the use of exec also precludes the ability to run additional commands after invoking the default image run script.

bfallonf: @bparees Works for me, thanks. I'll let @ahardin-rh follow up.

using_images/s2i_images/customizing_s2i_images.adoc
```diff
@@ -36,7 +36,8 @@ To determine the location of the scripts inside of the builder image, look at
 at the value of `*io.openshift.s2i.scripts-url*` label. Use `docker inspect`:
 
 ----
-$ docker inspect --format='{{ index .Config.Labels "io.openshift.s2i.scripts-url" }}' openshift/wildfly-100-centos7
+$ docker inspect --format='{{ index .Config.Labels "io.openshift.s2i.scripts-url" }}' /
```
php-coder: Does the `/` have a special meaning here? Where I can read about it?

ahardin-rh: It should be a backslash (not a forward slash) to indicate a line break, but I can remove it.

using_images/s2i_images/customizing_s2i_images.adoc
```diff
@@ -36,7 +36,8 @@ To determine the location of the scripts inside of the builder image, look at
...
+openshift/wildfly-100-centos7 /
```
php-coder: We should remove the trailing slash here, then.

ahardin-rh:
Thanks to all! These updates are complete.

ahardin-rh:
Rev history in original PR

_build_cfg.yml
```diff
@@ -615,8 +615,8 @@ Topics:
     Topics:
       - Name: Overview
         File: index
-      - Name: Using S2I images
-        File: customization
+      - Name: Customizing S2I images
```
adellape: @ahardin-rh Blast from the past! But...

s/images/Images/

I think it could also be at the end of the list of images, instead of the first?

adellape: @ahardin-rh Just FYI, I went ahead and did this via https://github.com/openshift/openshift-docs/pull/3524.

ahardin-rh: @adellape Thanks so much!

PR #2733, opened 2016-08-24
Fixed {product-title} instances in code block, removed oc from systemctl commands

ahardin-rh:


ahardin-rh:
Fixed per discussion with aburden

cc @vikram-redhat

ahardin-rh:
No rev history needed

PR #2765, opened 2016-08-30
I felt that correcting a typo and wording was important because the example would not complete properly and the wording was not as clear as it could be.

TheDiemer:


ahardin-rh:
@TheDiemer LGTM, although it looks like we'll need a rebase. Thanks!

ahardin-rh:
Work continued in https://github.com/openshift/openshift-docs/pull/2849

PR #2784, opened 2016-09-02
routing template data structure changes in 3.3 release

rajatchopra:
This document needs to exist in the release_notes section conveying the details of how/what/why the routing structures have changed in 3.3. release. This doc needs to be referenced appropriately in the 3.3 release notes. Please check this comment:
https://github.com/openshift/openshift-docs/pull/2774#issuecomment-244440648

@ahardin-rh @aheslin @adellape 

cc @knobunc @eparis

### DIFF
--- a/release_notes/routing_template_changes_3_3.adoc
+++ b/release_notes/routing_template_changes_3_3.adoc
@@ -1,0 +1,115 @@
+[[routing_template_changes]]
+= Routing Data Structure Changes
+{product-author}
+{product-version}
+:icons: font
+:experimental:
+:toc: macro
+:toc-title:
+:prewrap!:
+:description: This topic is an explanation of the changes in routing template structures in 3.3 release
+
+The underlying data structure that a router template can use has changed in origin release 1.3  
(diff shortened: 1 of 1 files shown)


knobunc:
The technical content LGTM.  There's probably some magic way to say 1.2 and 1.3 so that it expands for OSE correctly, but the docs team will have to weigh in on that.

Should we link to this from the admin guide?

vikram-redhat:
@rajatchopra - Hey Rajat - this seems like a lot of content to put in the release notes. Is there a reason why this can't go into a guide?

knobunc:
I think @rajatchopra was proposing that we add the document to the release_notes section and then just link to the doc from the actual release notes (with a summary explaining when you need to care). But if you'd rather that we put the main doc somewhere else, please propose a path.

eparis:
I'm not to picky about where the long form explanation lives. But a pointer to this needs to be big and bold and maybe using the <blink> tag in the release notes.

adellape:
@rajatchopra This was handled via https://github.com/openshift/openshift-docs/pull/2935.

PR #2787, opened 2016-09-06
Bug 1290487 added information on overlayfs to the prereqs

bfallonf:
As per: https://bugzilla.redhat.com/show_bug.cgi?id=1290487

I'll as for an ack in the BZ.

### DIFF
--- a/install_config/install/prerequisites.adoc
+++ b/install_config/install/prerequisites.adoc
@@ -623,8 +623,16 @@ xref:../../install_config/persistent_storage/persistent_storage_iscsi.adoc#insta
 === SELinux
 
 Security-Enhanced Linux (SELinux) must be enabled on all of the servers before
-installing {product-title} or the installer will fail. Also, configure
-`*SELINUXTYPE=targeted*` in the *_/etc/selinux/config_* file:
+installing {product-title}. 
+
+[IMPORTANT]
+====
+Without SELinux enabled, the installer will fail and there is an increased risk
+of a container accessing the contents of another, ensuring you will not have a
+true multi-tenant system.
+====
+
+Also, configure `*SELINUXTYPE=targeted*` in the *_/etc/selinux/config_* file:
 
 ----
 # This file controls the state of SELinux on the system.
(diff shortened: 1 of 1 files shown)


install_config/install/prerequisites.adoc
```diff
@@ -640,6 +640,18 @@ SELINUX=enforcing
 SELINUXTYPE=targeted
 ----
 
+[IMPORTANT]
```
sferich888: This note should include: 
```
without selinux, there is an increased risk of one container reading the contents of another and we can't claim to have a true multi-tenant system.  It's only valid to use OpenShift in this way if all users of the system are trusted.
```

bfallonf: @sferich888 I didn't think this worked with the Important box that's currently there, so I created a new one incorporating it into the intro para and. WDYT?

I also got info from Eduardo (BZ reporter) in the BZ and I'll add some more info when he gets back to me.

sferich888: We may want to separate the note about the installer failing and multi tenancy because they really are two separate comments or topics.

install_config/install/prerequisites.adoc
```diff
@@ -623,8 +623,16 @@ xref:../../install_config/persistent_storage/persistent_storage_iscsi.adoc#insta
...
-installing {product-title} or the installer will fail. Also, configure
-`*SELINUXTYPE=targeted*` in the *_/etc/selinux/config_* file:
+installing {product-title}. 
+
+[IMPORTANT]
+====
+Without SELinux enabled, the installer will fail and there is an increased risk
+of a container accessing the contents of another. Also, you will not have a true
+multi-tenant system.
+====
```
bfallonf: @sferich888 I made a change. But can I ask if the above is what you meant? 

Now that I'm looking at this more, I'm wondering if this is correct myself...

sferich888: Looks good however the "Also," does not flow. Maybe we should make it one sentence? Thoughts on changing: "another. Also, " to "another, ensuring "?

bfallonf: @sferich888 done. I'll wait for Eduardo's final ack in the BZ then move forward with this. Thanks!

bfallonf:
Closing this PR in favour of #4969

PR #2794, opened 2016-09-06
clarify multimaster requirement wording

brennv:
To make it clearer that the total in this example should be 4 CPU cores and 13 GB of RAM


brennv:
closing in favor of #2804

PR #2797, opened 2016-09-06
clean up the mongo replica example doc

bparees:


bparees:
@rhcarvalho ptal

using_images/db_images/mongodb.adoc
```diff
@@ -456,54 +456,76 @@ of the *rs0* replication set, as the value of `*MONGODB_REPLICA_NAME*` is set to
 === Creating the Service Pod
 
 To initialize members created by xref:creating-the-deploymentconfig[the
-deployment configuration], a _service pod_ is defined in the template. This pod
-starts MongoDB with the `initiate` argument, which instructs the container
-entrypoint to behave xref:mongo-creating-headless-service[slightly differently]
+deployment configuration], the pods are started with the `initiate` argument, 
+which instructs the container entrypoint to behave xref:mongo-creating-headless-service[slightly differently]
```
rhcarvalho: The "[container entrypoint](https://github.com/sclorg/mongodb-container/blob/master/2.4/Dockerfile#L57)" for that image just [execs](https://github.com/sclorg/mongodb-container/blob/master/2.4/root/usr/bin/container-entrypoint) the command/args, so one needs to make it [`['run-mongod', 'initiate']`](https://github.com/sclorg/mongodb-container/blob/master/2.4/examples/replica/mongodb-clustered.json#L93-L96).

using_images/db_images/mongodb.adoc
```diff
@@ -456,54 +456,76 @@ of the *rs0* replication set, as the value of `*MONGODB_REPLICA_NAME*` is set to
...
+  spec:
+    strategy:
+      type: Recreate
+      recreateParams:
+        post:
+          failurePolicy: Retry
+          execNewPod:
+            command:
+            - run-mongod
+            - initiate
```
rhcarvalho: I don't know if we're using this syntax everywhere for YAML snippets, but particularly for telling the command I'd use the single-line YAML form for clarity:

```yaml
command: [run-mongod, initiate]
```

Quotes are optional.

This suggestion is also optional :) it's just me trying to make sure people don't accidentally miss part of the command definition.

using_images/db_images/mongodb.adoc
```diff
@@ -456,54 +456,76 @@ of the *rs0* replication set, as the value of `*MONGODB_REPLICA_NAME*` is set to
...
+            containerName: mongodb
+            env:
+            - name: MONGODB_INITIAL_REPLICA_COUNT
+              value: '3'
```
rhcarvalho: Is the quote required here? It forces the value to be treated as a string, I wonder if not having it makes the YAML template invalid / unusable with OpenShift... just curiosity. We have the same [string](https://github.com/sclorg/mongodb-container/blob/master/2.4/examples/replica/mongodb-clustered.json#L101)/[int](https://github.com/sclorg/mongodb-container/blob/master/2.4/examples/replica/mongodb-clustered.json#L113) in the [JSON example template](https://github.com/sclorg/mongodb-container/blob/master/2.4/examples/replica/mongodb-clustered.json#L93-L96), though JSON has more strict syntax.

using_images/db_images/mongodb.adoc
```diff
@@ -456,54 +456,76 @@ of the *rs0* replication set, as the value of `*MONGODB_REPLICA_NAME*` is set to
...
+          ports:
+          - containerPort: 27017
 ----
 ====
 
 [[mongo-creating-headless-service]]
 === Creating a Headless Service
 
-The `initiate` argument in the xref:example-service-pod[container specification
+The `initiate` argument in the xref:example-deployment-config[container specification
```
rhcarvalho: Probably more chunks of this prose don't reflect the current example template...

I find little value in this docs talking about "creating a headless service" or describing in too much specificity how the replication example is implemented, as it easily rots.

We may well stop using headless services, or change the exact steps for initializing the replica set.

using_images/db_images/mongodb.adoc
```diff
@@ -549,22 +571,7 @@ using `rs.initiate(config)`. The script waits until MongoDB elects the *PRIMARY*
 member of the cluster.
 
 Once the *PRIMARY* member has been elected, the entrypoint script starts
```
rhcarvalho: s/entrypoint/replica set initialization ?

rhcarvalho:
In https://github.com/openshift/openshift-docs/pull/2797/files#diff-d6d70ef13c52e6fa08b0ff612b483578R565

There's no such thing as a "Replication set", it should be "Replica Set".

rhcarvalho:
@bparees thanks for doing this!

I wonder if we really need to have so much YAML with copies of the template (after a JSON -> YAML conversion) and so on here in the docs...

Theoretically, we should watch for any changes in the example template to update the docs in that case.

using_images/db_images/mongodb.adoc
bparees: i just ran a json to yaml converter on the existing json replica template object, i don't have a strong opinion either way.  I can change it.

using_images/db_images/mongodb.adoc
bparees: I think it is accurate as is, the initiate argument is being passed as part of the command, as seen in the updated example.

using_images/db_images/mongodb.adoc
bparees: incidentally the format you get back from "oc get foo -o yaml" is as it's represented in this PR as well, so i can change it, but it's non-standard as far as go seems to be concerned.

using_images/db_images/mongodb.adoc
bparees: this is an EnvVar struct value, not a true replica count field, so yes it must be a string.

using_images/db_images/mongodb.adoc
rhcarvalho: but the 'initiate' argument is passed to 'run-mongod', *not* to the container entrypoint.

using_images/db_images/mongodb.adoc
rhcarvalho: YAML is "non-standard" :-)
There are multiple ways to represent things, and I doubt a mechanic serializer would choose to use sometimes one form, sometimes other.

I don't have a problem leaving it as it is output, just suggested inlining for make it more clear, if that helps prevent user error.

using_images/db_images/mongodb.adoc
rhcarvalho: I've seen parsers that accept either int or string in the input, and attempt convert to the appropriate type (Python, Ruby). Not that I find that good behavior, but was curious whether the parser we use also does that sort of magic :)

rhcarvalho: My curiosity gone: https://play.golang.org/p/ONtWaRbj5H

```
json: cannot unmarshal number into Go value of type string
```

:smile:

using_images/db_images/mongodb.adoc
bparees: this whole doc is going to change a lot when we move to petsets, i'm not interested in a significant rework right now.

aslo i do think there is value in discussing the headless service aspect because
1) it's a relatively rare thing to be doing, i don't know if we have any other examples that do it
2) it has a meaningful impact on how you use it (you get a list of endpoints, it's not automatically load balanced)

using_images/db_images/mongodb.adoc
bparees: I think entrypoint script is a sufficiently generic description of what it's happening here.  it's the script (including embedded scripts) that are executed for the entrypoint.  We're not trying to do a line by line code review here, just give a basic overview of how the image bootstraps.  As you said, anything more detailed is going to rot.

using_images/db_images/mongodb.adoc
bparees: i'll change "container entrypoint" to "startup script" which is generic enough to no longer matter.

bparees:
> I wonder if we really need to have so much YAML with copies of the template (after a JSON -> YAML conversion) and so on here in the docs...

worth revisiting when we update this example to use petsets, but not interested in trying to refactor the yaml out of it right now.  This doc is a cross between a sample, a tutorial, and general documentation, so it's kind of muddled.

bparees:

rhcarvalho:
LGTM

ahardin-rh:
LGTM. Thanks!

ahardin-rh:
[rev_history]
|xref:../using_images/db_images/mongodb.adoc#using-images-db-images-mongodb[Database Images -> MongoDB]
|Updated some of the language surrounding the xref:../using_images/db_images/mongodb.adoc#example-deployment-config[Deployment Configuration Object Definition in the Example Template].
%

using_images/db_images/mongodb.adoc
```diff
@@ -456,54 +456,74 @@ of the *rs0* replication set, as the value of `*MONGODB_REPLICA_NAME*` is set to
 === Creating the Service Pod
```
php-coder: Section's title now outdated.

PR #2804, opened 2016-09-07
clarify multimaster requirement math

brennv:
Minor edit to make it clearer that the total requirement for this example would be 4 CPU and 13 GB of RAM.


itewk:
Thank you!

PR #2817, opened 2016-09-08
Follow-up to PR#2723

ahardin-rh:
Follow-up to https://github.com/openshift/openshift-docs/pull/2723


ahardin-rh:
Rev history in original PR

PR #2832, opened 2016-09-13
There is no "(see the section above)." in the OpenShift docs. That se…

jnordell:
…ction only exists in https://github.com/openshift/jenkins/README.md. Removed that statment and added a short description with no example of the format used in plugins.txt

But maybe this is included automatically from https://github.com/openshift/jenkins ? Then maybe the change needs to be applied there.


ahardin-rh:
Sorry for the delay. Thank you!

ahardin-rh:
No revision history needed

PR #2841, opened 2016-09-14
Add 3.3 support and move to new topic map filename

nhr:
This is a PR in two parts. Part of this PR depends upon a new feature of AsciiBinder, introduced in v0.1.5, the other part depends upond a feature introduced in v0.1.4. In short, if you are running an old version of AsciiBinder, it's time to upgrade:

```
$ gem update ascii_binder
```

Notable changes advanced by this PR:
1. `_build_cfg.yml` is now `_topic_map.yml` - the internal format of the file is unchanged; the new name is more consistent with its companion file `_distro_map.yml` and more meaningful.
2. The distro map file now supports the `distro-overrides` key within branch records. Any distro-level keys (but most significantly, `name`) can be overridden on a branch-by-branch basis. In this PR, I apply this new feature to enable a name change that does not affect every branch.

Note that line 44 of the updated file references a nonexistent branch. That's okay. AsciiBinder will happily skip it for now and when the branch exists, it will be correctly processed.


nhr:
@vikram-redhat - nope. The `_distro_map.yml` file is only read from the master branch during publishing runs.

_distro_map.yml
```diff
@@ -32,12 +29,21 @@ openshift-enterprise:
...
       name: '3.1'
       dir: enterprise/3.1
+      distro-overrides:
+        name: OpenShift Enterprise
     enterprise-3.2:
       name: '3.2'
       dir: enterprise/3.2
+      distro-overrides:
+        name: OpenShift Enterprise
+    container-platform-3.3:
```
adellape: @vikram-redhat We had said we were gonna stick with `enterprise-3.3` for the actual branch name. Still the case? If so, we'll need to adjust this line.

nhr:
Either way, it will not actually get published until I add a branch to the jenkins setup. So whichever way you go, just give me a day of lead time to make adjustments to Jenkins.

_distro_map.yml
vikram-redhat: @adellape - yeah that is correct. @nhr - Sticking with enterprise-3.3.

PR #2844, opened 2016-09-14
Bug 1372457, explained the impact of disabling dnsmasq

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1372457

and 

https://bugzilla.redhat.com/show_bug.cgi?id=1372456

### DIFF
--- a/install_config/install/prerequisites.adoc
+++ b/install_config/install/prerequisites.adoc
@@ -361,6 +361,17 @@ $ dig node1.example.com @10.64.33.1 +short
 10.64.33.101
 ----
 
+[[dns-config-prereq-disabling-dnsmasq]]
+==== Disabling dnsmasq
+
+If you want to disable *dnsmasq* (for example, if your *_/etc/resolv.conf_* is
+managed by a configuration tool other than NetworkManager), then set
+`openshift_use_dnsmasq` to *false* in the Ansible playbook.
+
+However, certain containers do not properly move to the next nameserver when the
+first issues *SERVFAIL*. Red Hat Enterprise Linux (RHEL)-based containers do not
+suffer from this, but certain versions of *uclibc* and *musl* do.
+
 [[wildcard-dns-prereq]]
 ==== Configuring a DNS Wildcard
 


ahardin-rh:
@miheer @sferich888 PTAL. Thanks!

sferich888:
@rjhowe PTAL

install_config/install/prerequisites.adoc
```diff
@@ -363,6 +363,23 @@ $ dig node1.example.com @10.64.33.1 +short
 10.64.33.101
 ----
 
+*The Impact of Disabling dnsmasq*
+
+If you want to set up {product-title} with *dnsmasq* disabled (for example, if
+your *_/etc/resolv.conf_*  is managed by a configuration tool other than
+NetworkManager), then `openshift_use_dnsmasq=False` is set in the Ansible
+playbook.
```
bfallonf: @ahardin-rh I think I'm a little confused by the above. Should it say:

"If you want to set up X, then set `openshift_use_dnsmasq` in the Ansible playbook to `false`."

or

"If you set up X, then {product-title} sets `openshift_use_dnsmasq` in the Ansible playbook to `false`"

?

bfallonf:
@ahardin-rh Would it be worth suggesting that the entire addition here be included in a Warning box or something? The previous procedure seems to use dnsmasq, so it's a bit jarring seeing this topic as a new section.

Otherwise, LGTM!

install_config/install/prerequisites.adoc
ahardin-rh: Fixed. Thanks!

ahardin-rh:
@bfallonf Thanks! I think it's a lot of information to place in a warning box, so I made it its own true section.

install_config/install/prerequisites.adoc
```diff
@@ -361,6 +361,22 @@ $ dig node1.example.com @10.64.33.1 +short
...
+
+If you want to disable *dnsmasq* (for example, if your *_/etc/resolv.conf_* is
+managed by a configuration tool other than NetworkManager), then set
+`openshift_use_dnsmasq` to *false* in the Ansible playbook.
+
+However, certain containers do not properly move to the next nameserver when the
+first issues *SERVFAIL*. Red Hat Enterprise Linux (RHEL)-based containers do not
+suffer from this, but certain versions of *uclibc* and *musl* do.
+
+Enabling *dnsmasq* also enables pushing to the registry via the DNS name
```
rjhowe: I do not understand this statement, are we talking about the route exposed for the registry or the SKYDNS service name. Even with out dnsmasq we should be able to use both if the cluster is configured correctly.  We inject the kubernetes service ip and search domains in the container resolv.conf  and take resolv.conf from the host. So we should be able to resolve everything still. 

Unless things have changed in later versions.

ahardin-rh: @rjhowe Thanks. I removed the statement.

ahardin-rh:
[rev_history]
|xref:../install_config/install/prerequisites.adoc#install-config-install-prerequisites[Installing -> Prerequisites]
|Added information about xref:../install_config/install/prerequisites.adoc#dns-config-prereq-disabling-dnsmasq[disabling dnsmasq].
%

PR #2846, opened 2016-09-15
Bug 1373234 add persistentcy to the ramp node section

bfallonf:
As per: https://bugzilla.redhat.com/show_bug.cgi?id=1373234

To make the steps persistent, just had to add that the steps here had to be in the noted file.

@b-mchugh Can I get an ack that this fulfills this BZ?


bfallonf:
As per discussion in the BZ and over email, a KBase article was chosen to be the best to document this BZ.

Closing this PR.

PR #2870, opened 2016-09-20
document git_ssl_no_verify support

bparees:


bparees:
@csrwng ptal.  yes this is kind of a lousy recommendation to have to make, but the mechanisms for overriding the trusted CAs list are not something i feel comfortable documenting right now.  We do now have a card in our backlog to make it easier to provide your own trusted CAs that will be used during the clone.

csrwng:
It would be good to include an example of how to set the var, like:
```
oc set env bc/mybuild GIT_SSL_NO_VERIFY=true
```

bparees:
@csrwng updated.

csrwng:
not exactly an example :) but I think it's enough to lead users in the right direction.
LGTM

bparees:
> not exactly an example :) but I think it's enough to lead users in the right direction.

yeah that was an intentional choice.  I always prefer linking to the primary source doc for something over providing an example that will bit rot if the tool in question changes.  Also it teaches users to fish instead of giving them a fish.

@openshift/team-documentation ptal/merge

csrwng:
@bparees you give users too much credit. In the end, all I want to know is how to do something and copy/paste if possible. What, you think I'd actually want to read the doc? That's so last century ;)

bparees:
> @bparees you give users too much credit. In the end, all I want to know is how to do something and copy/paste if possible. What, you think I'd actually want to read the doc? That's so last century ;)

:(

ahardin-rh:
@bparees Is this for 3.3?

bparees:
@ahardin-rh yes. labels added.

ahardin-rh:
Thanks!

PR #2778, opened 2016-09-01
Added a tutorial for setting up a Nexus service.

oatmealraisin:
The tutorial covers setting up a containerized Nexus Repository for Maven
dependency caching.

Continued from [here](https://github.com/openshift/origin/pull/10645).

I'm not sure if where I've put the tutorial is in the right section.

@bparees @openshift/team-documentation ptal?

PR #2875, opened 2016-09-20
Changed Nexus tutorial to add a Readiness probe

oatmealraisin:
Sorry, another small change.  The tutorial now adds a readiness probe, which will help with some stability.  @bparees @csrwng @openshift/team-documentation ptal?

### DIFF
--- a/dev_guide/app_tutorials/maven_tutorial.adoc
+++ b/dev_guide/app_tutorials/maven_tutorial.adoc
@@ -28,6 +28,11 @@ capability, read
 link:https://books.sonatype.com/nexus-book/reference/config.html[the Nexus documentation]
 to configure your build properly.
 
+Furthermore, make sure that you give each pod enough resources to
+function. You may have to edit the pod template in the Nexus deployment
+configuration to request more resources.
+xref:../deployment#creating-a-deployment-configuration[Learn more about setting deployment resources in pod templates].
+
 [[nexus-setting-up-nexus]]
 == Setting up Nexus
 
@@ -78,6 +83,11 @@ $ oc set probe dc/nexus \
 	--failure-threshold 3 \
 	--initial-delay-seconds 30 \
 	-- echo ok
+$ oc set probe dc/nexus \
+	--readiness \
+	--failure-threshold 3 \
+	--initial-delay-seconds 30 \
+	--get-url=http://:8081/content/groups/public
 ----
 
 [[nexus-adding-persistence-to-nexus]]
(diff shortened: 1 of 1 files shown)


dev_guide/app_tutorials/maven_tutorial.adoc
```diff
@@ -78,6 +78,11 @@ $ oc set probe dc/nexus \
 	--failure-threshold 3 \
 	--initial-delay-seconds 30 \
 	-- echo ok
+oc set probe dc/nexus \
+	--readiness \
+	--failure-threshold 3 \
+	--initial-delay-seconds 30 \
+	--get-url=http://:8081/content/groups/public
```
bparees: why isn't the liveness probe the same?

csrwng: should they be? One is to test whether the container is live (which a simple cmd would test), vs the other one to test whether it's ready to serve content

bparees: it depends, but in this case if it's not serving content, something has gone wrong and you probably want it restarted, so i'd expect them to be the same.

bparees: i'd expect a more "typical" liveness vs readiness pair to be:
1) liveness checks that the app server is still running (listening on port 80 and accepting connections)
2) readiness checks that the app itself is available/up/initialized

i don't think checking that the container is still live serves much purpose, if the process inside the container is hung.

csrwng:
@oatmealraisin you may want to mention about resource requirements and link to the place in the doc where we tell you how to set them.

oatmealraisin:
@csrwng Roger, testing out steps defined [here](https://docs.openshift.org/latest/dev_guide/compute_resources.html#dev-limit-ranges) to see if it's the right document to point at.

oatmealraisin:
@csrwng I added a mention for it at the top, ptal?

oatmealraisin:
@bparees @csrwng ptal? Did I miss something?

bparees:
@oatmealraisin anything you've already linked in the main doc you can remove from the references at the end.

dev_guide/app_tutorials/maven_tutorial.adoc
bparees: @oatmealraisin I didn't really see a satisfying answer on this.

oatmealraisin:
@bparees I don't believe I have anything left to change, ptal?

bparees:
lgtm. @openshift/team-documentation ptal/merge.

vikram-redhat:
@bfallonf - PTAL and merge. This is a follow up to https://github.com/openshift/openshift-docs/pull/2778

bfallonf:
Cool. Merging. I'll make a couple followup edits. Thanks, all!

PR #2886, opened 2016-09-21
Clarify how routes claim hostnames

knobunc:
This clarifies how the oldest route claims a hostname for an entire
namespace.  Other routes in the namespace can divide up the hostname
with different paths.  But other routes in different namespaces that
claim the hostname are ignored.

### DIFF
--- a/architecture/core_concepts/routes.adoc
+++ b/architecture/core_concepts/routes.adoc
@@ -322,9 +322,19 @@ In order for services to be exposed externally, an {product-title} route allows
 you to associate a service with an externally-reachable host name. This edge
 host name is then used to route traffic to the service.
 
-When two routes claim the same host, the oldest route wins. If additional routes
-with different path fields are defined in the same namespace, those paths will be
-added. If multiple routes with the same path are used, the oldest takes priority.
+When multiple routes from different namespaces claim the same host,
+the oldest route wins and claims it for the namespace. If additional
+routes with different path fields are defined in the same namespace,
+those paths will be added. If multiple routes with the same path are
+used, the oldest takes priority.
+
+Please note that a consequence of this behavior is that if you have
+two routes for a hostname: an older one and a newer one.  And if
+someone else has a route for the same hostname that they created
+between when you created the other two routes.  Then, if you delete
+your older route, your claim to the hostname will no longer be in effect.
+The other namespace will now claim the hostname and your claim will be
+lost.
 
 .A Route with a Specified Host:
 ====


knobunc:
@rajatchopra @ramr PTAL

rajatchopra:
Text is good.
Will it be better if the sentences were itemized in bullet points? They kind of suggest cases that are mutually exclusive.

ramr:
Am ok - as there is some overlap (no pun intended) with the prior content. 
LGTM

knobunc:
I added the case that @marun wanted to be clarified.  @ramr @rajatchopra is that still reasonable or should I re-work it.

@ramr as you observed, I kept much of the original paragraph.  I just had to re-flow the paragraph... is that correct, or should I have left the long line to do as little damage to adjacent text as possible?  Or is there a way to tell github to use a different diff algorithm here?

architecture/core_concepts/routes.adoc
```diff
@@ -322,9 +322,19 @@ In order for services to be exposed externally, an {product-title} route allows
 you to associate a service with an externally-reachable host name. This edge
 host name is then used to route traffic to the service.
 
-When two routes claim the same host, the oldest route wins. If additional routes
-with different path fields are defined in the same namespace, those paths will be
-added. If multiple routes with the same path are used, the oldest takes priority.
+When multiple routes from different namespaces claim the same host,
```
ramr: Well, there's also the case to consider that 2 routes in the same namespace claiming the same host - so maybe drop the part about `from different namespaces` and `claims it for the namespace`.

architecture/core_concepts/routes.adoc
```diff
@@ -322,9 +322,19 @@ In order for services to be exposed externally, an {product-title} route allows
...
+the oldest route wins and claims it for the namespace. If additional
+routes with different path fields are defined in the same namespace,
+those paths will be added. If multiple routes with the same path are
+used, the oldest takes priority.
+
+Please note that a consequence of this behavior is that if you have
+two routes for a hostname: an older one and a newer one.  And if
+someone else has a route for the same hostname that they created
```
ramr: Maybe `And if another route uses/claims/asks for the same hostname` since it could be someone else or another route (created by the same user).

architecture/core_concepts/routes.adoc
```diff
@@ -322,9 +322,19 @@ In order for services to be exposed externally, an {product-title} route allows
...
+between when you created the other two routes.  Then, if you delete
+your older route, your claim to the hostname will no longer be in effect.
+The other namespace will now claim the hostname and your claim will be
```
ramr: `The other route` would cover the case of same/different namespaces.

ramr (commented):
Trying out github review ...

ramr:
@knobunc on the other point, I think the formatting changes are ok.

architecture/core_concepts/routes.adoc
knobunc: Perhaps I am misunderstanding how it works... but the way I read the code, if I have:

Namespace 1:

- Route 1 Host "foo.com" Path "/foo" [Created 1/1/2001]
- Route 2 Host "foo.com" Path "/" [Created 9/1/2001]


Namespace 2:

- Route 3 Host "foo.com" Path "/bar" [Created 2/1/2001]
- Route 4 Host "foo.com" Path "/foo" [Created 3/1/2001]

If all of those are in effect, then only the routes from Namespace 1 will be able to apply any routes to "foo.com" because Route 1 was created first.  So requests to /foo and below go to route 1, everything else goes to route 2.

Now... if I delete Route 1, then since Route 3 is next oldest, the routes in namespace 2 will take effect for foo.com.  So /bar and below goes to route 3, /foo and below goes to route 4, and everything else gets an HTTP not found.

Is my understanding correct?

ramr: So `/bar` is a different path, so it would be available along with Route 1 and 2 as we sort on the paths. Route 4 is a dup, so the oldest one for that path `/foo` would win. If you delete route 1, route 4 becomes the oldest.

knobunc: This code seems not to implement that:
  https://github.com/knobunc/origin/blob/master/pkg/router/controller/unique_host.go#L106

And allowing someone else to inject a distinct path seems scary too.  If I claimed the path '/' and was serving the URL '/secret_thing' as one of my items, then someone else could put a path in with the same hostname and a more specific path underneath my general one and steal traffic.

ramr: My bad, just looked at the route paths (!namespaces and shouldn't reply during meetings).

knobunc: Given the confusion, do you want to re-review?  I'm not sure which comments still obtain.

Thanks.

knobunc:
@ramr Can you please take another look now we cleared up the behavior of the code?

ramr:
LGTM

knobunc:
@adellape PTAL

knobunc:
@openshift/team-documentation Please merge

ahardin-rh:
@knobunc Is this for 3.4? Thanks!

knobunc:
@ahardin-rh definitely 3.4, but if you can get it in 3.3 it applies there too.  Sorry, I have been trying to put that into the main descriptions, but apparently forgot here :-(

ahardin-rh:
@knobunc No worries! Thanks!

ahardin-rh:
[rev_history]
|xref:../architecture/core_concepts/routes.adoc#architecture-core-concepts-routes[Core Concepts -> Routes]
|Clarified how routes claim xref:../architecture/core_concepts/routes.adoc#route-hostnames[host names].
%

vikram-redhat:
Hey @knobunc and @ahardin-rh - if this applies to both 3.3 and 3.4, and if it can be merged now, then we don't need the branch/enterprise-3.4 label on it, and the milestone can be Next Release. Are there any changes in here that don't apply to 3.3?

knobunc:
@vikram-redhat there is nothing in this PR that does not apply to 3.3.

ahardin-rh:
@knobunc @vikram-redhat Thanks! Updating the labels/milestone now.

Issue #2461, opened 2016-07-05
browser requirements are out of date


danmcp:
https://docs.openshift.org/latest/architecture/infrastructure_components/web_console.html#browser-requirements

Not a single item listed isn't horribly out of date:

Windows 8, ipad 2, iphone 4 and macOSX (is now just macOS).  Even Fedora should be 24 now.  But I am not sure what good this list really does.  If it's going to be out of date this quickly, perhaps we should just remove it.

danmcp:
@adellape @vikram-redhat FYI

vikram-redhat:
@danmcp - I agree. We should just remove it. I will get it removed soon.

adellape:
@danmcp @vikram-redhat This was originally added not that long ago via https://bugzilla.redhat.com/show_bug.cgi?id=1322352 and https://github.com/openshift/openshift-docs/pull/1874, requested by GSS. It looks like the table was based on the support matrix from QE: https://mojo.redhat.com/docs/DOC-1013063. Maybe that needs to get updated? But agree that this table will be easy to get out of date on.

cc @ahardin-rh

danmcp:
@vikram-redhat @adellape Can we move it to the release notes or some other place.  I know it isn't right for windows/iphone/ipad versions.  And macOS was just renamed.  It was correct for the Fedora version but isn't like we aren't going to support it on the latest Fedora now that it's out.  If we have to put it somewhere then so be it.  But given how off the messaging is from reality, it's only going to confuse people.  No one is going to see that list and say thank you for letting me know my ipad 2 is supported.

vikram-redhat:
@danmcp I believe this was correct at the time of publishing for 3.2. This is what had been tested by QE, as opposed to what will work, and I think matters most to support.

I think the solution might be to add this to the release notes, as you suggest, for that particular version to say that this version of OpenShift has been tested with these OS/Browser settings.

Asking for @sferich888 's input.

sferich888:
Sounds good to me. That is what we say on the tested configuration matrix.

bfallonf:
Thanks @danmcp I've created a PR for this:

https://github.com/openshift/openshift-docs/pull/2889

I'll close this issue and if there's anything to say, we can move the conversation to there.

PR #2889, opened 2016-09-22
Issue 2461: moved supported browsers to the rel notes

bfallonf:
As per: https://github.com/openshift/openshift-docs/issues/2461

Should be straight-forward, but @adellape @ahardin-rh @tpoitras please let me know if there's any issues.

### DIFF
--- a/release_notes/ose_3_2_release_notes.adoc
+++ b/release_notes/ose_3_2_release_notes.adoc
@@ -619,6 +619,35 @@ behavior, run:
 $ oc rsh <name> -- /bin/bash
 ----
 
+[[browser-requirements]]
+== Browser Requirements
+
+ifdef::openshift-enterprise[]
+Review the https://access.redhat.com/articles/2176281[tested integrations] for
+{product-title}.
+endif::openshift-enterprise[]
+The following browser versions and operating systems have been tested and
+support the web console:
+
+[cols="1,2"]
+.Browser Requirements
+|===
+|Browser (Latest Stable) |Operating System
+
+|Firefox
+|Fedora 23, Windows 8
+
+|Internet Explorer
+|Windows 8
+
+|Chrome
+|Fedora 23, Windows 8, and MacOSX
+
+|Safari
+|MacOSX, iPad 2, iPhone 4
+
+|===
+
 [[ose-32-bug-fixes]]
 == Bug Fixes
 
--- a/architecture/infrastructure_components/web_console.adoc
+++ b/architecture/infrastructure_components/web_console.adoc
@@ -66,35 +66,6 @@ endif::openshift-enterprise,openshift-origin[]
 
 image::about_page.png["About Page"]
 
-[[browser-requirements]]
-== Browser Requirements
-
-ifdef::openshift-enterprise[]
-Review the https://access.redhat.com/articles/2176281[tested integrations] for
-{product-title}.
-endif::openshift-enterprise[]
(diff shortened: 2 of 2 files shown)


bfallonf:
Actually, @adellape have you been working on the 3.3 rel notes? Can we make sure the move here sticks around for 3.3?

bfallonf:
@adellape bump!

release_notes/ose_3_2_release_notes.adoc
```diff
@@ -619,6 +619,35 @@ behavior, run:
 $ oc rsh <name> -- /bin/bash
 ----
 
+[[browser-requirements]]
```
adellape: Maybe give this a more specific ID like `ose-3-2-browser-requirements`?

release_notes/ose_3_2_release_notes.adoc
```diff
@@ -619,6 +619,35 @@ behavior, run:
...
+== Browser Requirements
+
+ifdef::openshift-enterprise[]
+Review the https://access.redhat.com/articles/2176281[tested integrations] for
+{product-title}.
+endif::openshift-enterprise[]
```
adellape: Can drop `openshift-enterprise` from the `endif`.

adellape (commented):
Whoa, an oldie! Couple minor things but LGTM in general. Maybe we should look to do this for 3.3 and later after, as well.

release_notes/ose_3_2_release_notes.adoc
adellape: Well actually, you could drop the `ifdef` completely, since this is an enterprise-only topic overall.

ahardin-rh:
@bfallonf Looks like this will also address https://github.com/openshift/openshift-docs/issues/4218 if we can get the list updated for each version of OCP.

ahardin-rh:
@bfallonf bump

bfallonf:
@ahardin-rh @adellape Oop this is an old one. Hmm if these are the minimum requirements, they probably haven't changed? Would it be accurate to say this should be in the master branch, then cherry-pick it to each release branch (after making Alex's suggested changes above)?

adellape:
@bfallonf @ahardin-rh Browser support actually came up again recently: jwforres was telling erich and I that the [OpenShift Container Platform Tested Integrations](https://access.redhat.com/articles/2176281) can be updated to include the mobile browsers supported, as well (have that in email). We already link to that KBase as far back as 3.0 docs, so I think we should just remove the table here and let the KBase handle keeping that support matrix updated per OCP version.

The only other issue may be that the KBase only specifically mentions the OCP distro by name (and version #), and not Origin/Online/Dedicated... I guess we'd need to either A) get the KBase updated to address those distros as well or create separate KBases for each (work with GSS), B) just link to it anyway, or C) update and maintain tables here for those distros.

bfallonf:
@adellape sure. Hmm not sure which option is better. I agree that it would be easier to just link to the KBase. No point in doubling-up the efforts.

I'm not too sure what to do next. Is there someone in GSS that would be best to include the distro info in the KBase?

adellape:
> Is there someone in GSS that would be best to include the distro info in the KBase?

@bfallonf We'd wanna chat w/ erich about it.

ahardin-rh:
Closing in favor of https://github.com/openshift/openshift-docs/pull/5745 after discussion with the web console team

PR #2891, opened 2016-09-22
Add oc get netnamespace command

ahardin-rh:


ahardin-rh:
[rev_history]
|xref:../install_config/configuring_sdn.adoc#install-config-configuring-sdn[Configuring the SDN]
|Added that `oc get netnamespace` can be run to check VNIDs.
%

PR #2800, opened 2016-09-07
remove throttling page

brennv:
As discussed in https://github.com/openshift/openshift-docs/issues/2464

PR #2913, opened 2016-09-23
remove throttling page

ahardin-rh:
Continues work from https://github.com/openshift/openshift-docs/pull/2800


ahardin-rh:
No revision history needed

PR #2916, opened 2016-09-24
remove python server opinions

### LINKED ISSUE #2902: Statement about mod_wsgi no longer being supported is not accurate.
In:
- https://github.com/openshift/openshift-docs/blob/master/dev_guide/migrating_applications/web_framework_applications.adoc

the documentation says:

> S2I-python does not support mod_wsgi anymore. However, gunicorn is supported and it is an alternative for mod_wsgi. So, add gunicorn package to requirements.txt.

This is not accurate and gives an impression that it is not possible to use mod_wsgi at all with the Python S2I builder.

All the documentation should state is that by default the Python S2I builder will use `gunicorn`, it need not mention mod_wsgi at all, or it could perhaps say that to use other WSGI servers additional configuration steps would need to be performed.

There are details on how to use mod_wsgi and uWSGI in as yet unmerged discovery centre documentation. For example:
- https://github.com/openshift/discoverycenter/blob/master/source/developer/languages/python/advanced-deployment.html.adoc#running-with-apache-mod_wsgi

The migration docs for Python also neglect to mention that because gunicorn is now used, static files cannot be hosted unless you use a separate web server, or use WSGI middleware called Whitenoise. For that reason it is still better to use mod_wsgi or uWSGI as they can handle static file handling without needed to change the application and will do it more efficiently as well.

brennv:
resolves #2902


rhcarvalho:
@brennv there is an extraneous merge commit in your PR. Could you please rebase on top of the latest master?
There should be just the contents of https://github.com/openshift/openshift-docs/pull/2916/commits/3070b16d2f2472649a91cc6033abd20f82a893d1.

@openshift/team-documentation please review.

bfallonf:
@brennv Yes, thanks for this. Maybe a squash of the two commits and I can get this merged. Thanks!

ahardin-rh:
@brennv This is an oldie. :ghost: Do you mind squashing this into one commit if this if it is still valid? If not, can we close it?

brennv:
Closing in favor of #4440

PR #2924, opened 2016-09-26
Clean-up / follow-up on cluster metrics content 

tpoitras:
As per vikram, found some content during release cherry-picking that hadn't been followed up on.

nicely rendered: http://file.bne.redhat.com/tpoitras/2016/metrics/openshift-enterprise/clustermetrics-fix/install_config/cluster_metrics.html#capacity-planning-for-openshift-metrics

ping @adellape @bfallonf peer review?

### DIFF
--- a/install_config/cluster_metrics.adoc
+++ b/install_config/cluster_metrics.adoc
@@ -167,118 +167,125 @@ $ oc process -f metrics-deployer.yaml -v \
     | oc create -f -
 ----
 
-The size requirement of the Cassandra storage is dependent on the number of pods.
-It is the administrator's responsibility to ensure that the size requirements
-are sufficient for their setup and to monitor usage to ensure that the disk does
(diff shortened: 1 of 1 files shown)


tpoitras:
[fix up cluster metrics stuff (no followup had been done)](https://trello.com/c/ngR1t0i1/36-fix-up-cluster-metrics-stuff-no-followup-had-been-done)

tpoitras:
@adellape wrote on IRC:

tpoitras: briefly looked thru it and didn't see anything big jump out. did see some generic headings like "Prerequisites" that might as well get manual IDs at this point

tpoitras: with those ifdef'd OCP "metrics.yaml" references, it would be nice if we could have verified if there's an on-disk equivalent location on a host that we could mention instead, but this works i think at least for now. we could follow-up w/ that if needed

PR #2949, opened 2016-09-27
Adding Nuage to Openshift docs

abhat:
Please review cc @vishpat @hasahni


vikram-redhat:
@bfallonf - PTAL.

bfallonf:
Thanks @abhat . It might be best to merge this and then I'll make some followups in another PR.

install_config/configuring_nuagesdn.adoc
```diff
@@ -0,0 +1,112 @@
...
+ # VSP related parameters
+ vsd_api_url=https://192.168.103.200:8443
+ vsp_version=v4_0
+ enterprise=nuage
+ domain=openshift
+ vsc_active_ip=192.168.103.201
+ vsc_standby_ip=192.168.103.202
+ uplink_interface=eth0
+
+ # rpm locations
```
sferich888: If Red Hat is supporting this plugin the 'location_of_rpm_server' should come from the Red Hat CDN, and this should be a channel based option. What channel will be providing these plugins?

install_config/configuring_nuagesdn.adoc
```diff
@@ -0,0 +1,112 @@
...
+ nuage_master_adminuserpasswd=admin
+ nuage_master_cspadminpasswd=admin
+ nuage_openshift_monitor_log_dir=/var/log/nuage-openshift-monitor
+
+ # Required for brownfield install (where a {product-title} cluster exists without Nuage as the networking plugin)
+ nuage_dockker_bridge=lbr0
+
+=== Installation for multiple masters (HA)
+
+A High Availability (HA) environment can be configured with multiple masters and multiple nodes. There are two supported configuration methods for HA--native HA configuration and configuration using Pacemaker.
```
sferich888: We moved away from using pacemaker for HA deployments in 3.1 (https://access.redhat.com/node/2176281), while it is OK to call out that HA can be achieved, with pacemaker, we should only document how to use an Active Active configuration, to align with the recommendations we make in other parts of the documentation.

architecture/additional_concepts/nuagesdn.adoc
```diff
@@ -0,0 +1,117 @@
...
+
+[[nuage-openshift-monitor]]
+
+=== nuage-openshift-monitor
+
+The nuage-openshift-monitor is a service that monitors the {product-title} API server for creation of projects, services, users, user-groups etc.
+
+[NOTE]
+=====
+In case of Highly Available (HA) {product-title} cluster with multiple masters, nuage-openshift-monitor process runs on all the masters independently without any change in functionality.
```
sferich888: Is this done with DaemonSets? Does that need to be explained?

architecture/additional_concepts/nuagesdn.adoc
```diff
@@ -0,0 +1,117 @@
...
+
+For the operations workflow, the users specify Nuage recognized labels on their application/pod specification to resolve the pods into specific user-defined zones/subnets. This however cannot be used to resolve pods in the zones/subnets created via the developer workflow by nuage-openshift-monitor.
+
+[NOTE]
+=====
+In the operations workflow, the administrator is responsible for pre-creating the VSD constructs to map the pods into a specific zone/subnet as well as allow communication between OpenShift entities (ACL rules, policy groups, network macros, and network macro groups). Detailed description of how to use Nuage labels is provided in the link:http://support.alcatel-lucent.com[Nuage VSP Openshift Integration Guide].
```
sferich888: If an admin is responsible for these actions, and these actions are supported, we need documentation on how to make these mappings.

architecture/additional_concepts/networking.adoc
```diff
@@ -92,9 +92,18 @@ routing for its endpoints. Service DNS can still be used and responds with
 multiple A records, one for each pod of the service, allowing the client to
 round-robin between each pod.
 
+[[network-plugins]]
+
+== Network Plugins
+
+{product-title} supports the same plugin model as Kubernetes for networking pods. 
+ifdef::openshift-enterprise,openshift-origin[]
+The following network plugins are currently supported by {product-title}.
```
abhat: We can change the verbiage to say the following:

"In addition to Openshift SDN, the following third party plugins can be configured". and in the Nuage SDN section, I can call out that the "Nuage SDN network plugin is a third party plugin supported by Nuage Networks". Will that work?

install_config/configuring_nuagesdn.adoc
abhat: Since RH is not supporting this plugin officially, we can call out that the rpms will be provided/hosted by Nuage.

install_config/configuring_nuagesdn.adoc
abhat: I will take out the line for Pacemaker altogether to avoid confusion.

architecture/additional_concepts/nuagesdn.adoc
abhat: No it isn't yet. Each master will run this service under systemd. DaemonSets is a fairly new feature, IMO adding that to the mix here will confuse people than clarify.

architecture/additional_concepts/nuagesdn.adoc
abhat: Just to be clear, these actions aren't supported via Openshift api but are to be done OOB. As per our conversation with the RH partner team (Chris Morgan), we were asked to include only the overview of Nuage as pertinent to the plugin. The details of how to configure these constructs are in the Nuage VSP user-guide and details around how make the mappings are in Nuage's Openshift integration guide. Both these documents are made available to customers through our support page (hence the link). The idea is to keep Openshift documentation as clean and separated from third party config. details as possible, while pointing users to the right documentation hosted by third party vendors.

architecture/additional_concepts/networking.adoc
sferich888: Yes, but I I still have questions about support as it relates to this, because we have install steps.

install_config/configuring_nuagesdn.adoc
sferich888: What is the host or way customer get the RPMs the links here seem like placeholders

architecture/additional_concepts/nuagesdn.adoc
sferich888: We likely just need to outline that this is a systemd service

architecture/additional_concepts/networking.adoc
abhat: The idea is the plugin will be supported by Nuage completely. The install steps are listed so users know  what playbook parameters they need in their nodes file in order to use Nuage as a plugin. In no way is this meant to put the support onus on RH as far as installing openshift with Nuage is concerned. If there are any other questions/concerns, please let us know.

install_config/configuring_nuagesdn.adoc
abhat: The links are indeed placeholders. Nuage doesn't have public mirrors - but we do provide tarballs and help customers set their own repos to which they can point their nodes files.

architecture/additional_concepts/nuagesdn.adoc
abhat: Will do.

architecture/additional_concepts/networking.adoc
sferich888: IMO this means that the playbooks need to be packaged and provided by nuage. This will help to make it clear that they support the install.

We should also make our install docs a reference to nuage docs that cover this information.

bfallonf:
@abhat @sferich888 Thanks for the comments, guys.

I've made a followup PR and shared with the docs content strategist, Ali @aheslin . She's started some conversations with people to determine what we can put into the docs. We can work going further with this after that.

vikram-redhat:
@bfallonf - thanks.

@Eric and @abhat - thanks for the conversation. Just to make it clear, this is at the moment, only in the Origin (upstream) docs. It will move to the downstream, RH supported docs, only after approval from the content strategist @aheslin and GSS @sferich888 .

architecture/additional_concepts/nuagesdn.adoc
munchee13: abhat is correct. It is important to ensure that Red Hat is NOT distributing any Nuage software. However, since SDN configuration is something that needs to be accounted for in advance of an install for the best customer experience, configuration documentation should be included for this partner. This would be akin to how we describe configuration of 3rd party storage plugins, for example. The instructions for installing Nuage itself should remain with Nuage.

aheslin:
I confirmed with @sferich888 that we're ok to publish this downstream as well. Just FYI.

munchee13:
Which OCP release will include this?

PR #2961, opened 2016-09-28
REST API Added manual ids

vikram-redhat:


vikram-redhat:
It wasn't the case of missing ids - but an upstream bug that is brought into the docs: https://github.com/kubernetes/kubernetes/pull/33007/files. Closing this, and submitting another PR for duplicating the fix for the upstream bug.

PR #2964, opened 2016-09-29
Prefer short flag for whoami -t

liggitt:
In light of https://github.com/openshift/origin/pull/11150, we should prefer the short flag for `oc whoami -t`

### DIFF
--- a/rest_api/index.adoc
+++ b/rest_api/index.adoc
@@ -44,7 +44,7 @@ command:
 ----
 $ oc login -u test_user
 Using project "test".
-$ oc whoami --token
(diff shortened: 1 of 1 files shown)


liggitt:
If possible, could we pick this back to 3.2 and 3.1?

ahardin-rh:
Thanks!

ahardin-rh:
[rev_history]
|xref:../rest_api/index.adoc#rest-api-index[Overview]
|Updated  the `oc whoami --token` command to show the shorter flag of `oc whoami --t`.
%

PR #2967, opened 2016-09-29
Set the DomainName or DomainID in the OpenStack cloud provider

lhuard1A:
Whereas with the OpenStack keystone v2 authentication API, the username and
the password are enough, the keystone v3 API requires an additional domain
parameter.

This domain parameter was not yet documented, but already [taken into account
by released OpenShift](https://github.com/openshift/origin/blob/master/vendor/k8s.io/kubernetes/pkg/cloudprovider/providers/openstack/openstack.go#L109).

relates to openshift/openshift-ansible#2515

### DIFF
--- a/install_config/configuring_openstack.adoc
+++ b/install_config/configuring_openstack.adoc
@@ -27,6 +27,7 @@ nodes:
 auth-url = <OS_AUTH_URL>
 username = <OS_USERNAME>
 password = <password>
+domain-id = <OS_USER_DOMAIN_ID>
 tenant-id = <OS_TENANT_ID>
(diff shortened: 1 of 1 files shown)


ahardin-rh:
LGTM. Thank you! Sorry for the delay!

ahardin-rh:
[rev_history]
|xref:../install_config/configuring_openstack.adoc#install-config-configuring-openstack[Configuring for OpenStack]
|Added `openshift_cloudprovider_openstack_domain_id` and `openshift_cloudprovider_openstack_domain_name` to the list of configurable parameters.
%

PR #2974, opened 2016-10-03
Registry 2.4 updates

miminar:
Registry 2.4 has been back-ported!


miminar:
@legionus PTAL
@sdodson, @mfojtik  fyi

sdodson:
@miminar So if they've not provided a custom configuration file then there's really not much they need to do when upgrading?

legionus:
LGTM!

miminar:
@sdodson There's really not much to do. All should just work as before. I could probably mention at least enabling `acceptschema2` in the upgrade guide.

Shall I update release notes as well?

Shall I also mention recommended order of components upgrade? Or is this a job for installer?

miminar:
@sdodson Updated release notes and mentioned `acceptschema2` in update guide.

miminar:
The registry 2.4 hasn't been back-ported and, most probably, never will.

PR #2987, opened 2016-10-04
Bug 1381408, Fix sizing guidelines

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1381408

### DIFF
--- a/install_config/install/planning.adoc
+++ b/install_config/install/planning.adoc
@@ -52,7 +52,10 @@ If you install initially using the quick installer, you can always further tweak
 [[sizing]]
 == Sizing Considerations
 Determine how many nodes and pods you require for your {product-title} cluster.
-The following table provides the sizing limits for nodes and pods:
+Cluster scalability correlates to the number of pods in a cluster environment.
+That number influences the other numbers in your setup.
+
+The following table provides the maximum sizing limits for nodes and pods:
 
 [cols="8,2",options="header"]
 |===
@@ -62,17 +65,11 @@ The following table provides the sizing limits for nodes and pods:
 
 |Maximum pods per cluster |120,000
 
-|Maximum pods per nodes |250
+|Maximum pods per node |250
 
 |Maximum pods per core |10
 
 |===
-
-For example, if you require 2200 pods, you need at least 20 nodes in your
-environment.
-
-This also means the maximum number of pods that can run safely in your
-environment is 33,000.
 
 [IMPORTANT]
 ====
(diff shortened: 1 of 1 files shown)


ahardin-rh:
@jeremyeder can you please take a look? Thanks!

ahardin-rh:
@jeremyeder says that maximum pods per nodes should still be 250. Not sure how this needs to be updated, but GSS is saying that this section of the docs is incorrect. @timothysc are you able to help? Thanks!

timothysc:
The table with the maximum pods/node is correct.  Perhaps we should add the hint that you need to do the math then.  e.g. @ 250 pods/node you'll top out at 480 nodes.  

Those guidelines are upper bounds, everything is correct.

ahardin-rh:
@adellape Made some updates. Can you PTAL? 

Preview build: http://file.rdu.redhat.com/~ahardin/10262016/sizing-fixes/install_config/install/planning.html#sizing

install_config/install/planning.adoc
```diff
@@ -52,27 +52,50 @@ If you install initially using the quick installer, you can always further tweak
...
+The following table provides the maximum sizing limits for nodes and pods:
 
 [cols="8,2",options="header"]
 |===
 |Type |Maximum
 
 |Maximum nodes per cluster |1000
 
-|Maximum pods per cluster |120,000
+|Maximum pods per cluster |120000
```
adellape: http://stylepedia.net/#chap-Red_Hat_Technical_Publications-Writing_Style_Guide-N

See "numbers" for comma usage.

install_config/install/planning.adoc
```diff
@@ -52,27 +52,50 @@ If you install initially using the quick installer, you can always further tweak
...
+----
+
+If you increase the number of nodes to 20, then the pod distribution changes to
+5000 pods per cluster:
+
+----
+20 * 250 = 5000
+----
 
 [IMPORTANT]
```
adellape: Maybe this IMPORTANT note can go up after the table now, but before Scenario A?

timothysc:
Better, but focuses too much on the high end number which is more an edge case. 

Maximum Pods / (Expected Pods/Node <= Max Pods/Node) = Total Number of nodes. 

Admins should profile their pods and nodes to determine exactly how many pods they expect to fit per node.

ahardin-rh:
@timothysc Okay, thanks!

Do you think the scenario approach is okay? If so, do you have any suggested (more realistic/less edge case) scenarios we can use?

Also, is it Scenario A that focuses too much on the high end numbers?

ahardin-rh:
@adellape I made some updates. @timothysc Any suggested scenarios I should use to make the numbers more realistic? Should I just nix the first scenario altogether? Thanks.

ahardin-rh:
Made further updates. Sending to QE for review.

ahardin-rh:
QE verified

ahardin-rh:
[rev_history]
|xref:../install_config/install/planning.adoc#install-config-install-planning[Installing -> Planning]
|Updated the xref:../install_config/install/planning.adoc#sizing[Sizing Considerations] section for clarity.
%

PR #2989, opened 2016-10-04
Bug 1380563, updated Enabling Features in Technology Preview section

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1380563

### DIFF
--- a/install_config/web_console_customization.adoc
+++ b/install_config/web_console_customization.adoc
@@ -366,7 +366,7 @@ window.OPENSHIFT_CONSTANTS.ENABLE_TECH_PREVIEW_FEATURE.<feature_name> = true;
 ====
 
 . Add it to the master configuration file:
-
++
 ====
 ----
 assetConfig:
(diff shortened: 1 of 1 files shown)


ahardin-rh:
@adellape PTAL

ahardin-rh:
@bfallonf peer review please :bowing_woman:

bfallonf:
@ahardin-rh LGTM :star2:

ahardin-rh:
@jwforres PTAL. Thanks!

install_config/web_console_customization.adoc
```diff
@@ -376,6 +376,41 @@ assetConfig:
 ----
 ====
 
+. As a cluster administrator, run the following command to give users the
```
jwforres: this is really specific to jenkins pipelines, this wont be true of all tech preview features

jwforres (changes requested):

install_config/web_console_customization.adoc
jwforres: @bparees are there docs about this in the build doc already?

bparees: permission to use pipeline strategies should be on by default already, not sure where this is coming from?

i think all strategies are on by default, though we've been discussing turning off custom by default.

ahardin-rh: @jwforres @bparees This was sourced from https://github.com/nickschuetz/ocp-3.3-pipelines/blob/master/README.md. However, it is noted that the steps arose from a possible edge case during an upgrade from 3.2 to 3.3, so this content may not be valid.

bparees: yeah, i'm pretty sure that guy just didn't run reconcile roles which is a standard part of the upgrade process.

bparees: so yes i'm pretty sure this should be closed.

PR #2990, opened 2016-10-04
Additions and clarifications to doc guidelines

ahardin-rh:
@adellape @aheslin @bfallonf @tpoitras @vikram-redhat Here is my first pass at the revised doc guidelines. I really want to remove the guess work from the markdown we use. We were running into a lot of inconsistencies and I think that simplifying things will help. The biggest proposed change in that regard is changing the formatting for environment variables to back-ticks. Other product groups seem to use that convention as well.  Per a conversation with @aheslin I also added in guidelines on writing tasks vs concepts vs user goals, and a few other bits of guidance that I think were missing. Please share your thoughts and let me know if I missed anything.

### DIFF
--- a/contributing_to_docs/doc_guidelines.adoc
+++ b/contributing_to_docs/doc_guidelines.adoc
@@ -1,5 +1,5 @@
 [[contributing-to-docs-doc-guidelines]]
-= Documentation guidelines
+= Documentation Guidelines
 :icons:
 :toc: macro
 :toc-title:
@@ -8,7 +8,7 @@
 
 toc::[]
 
-== Topic metadata
+== Topic Metadata
 Every topic file should contain the following metadata at the top, with no line spacing in between, except where noted:
 
 ----
@@ -42,7 +42,7 @@ If using section anchors, they must be all lowercase letters, with no line space
 === Section Title
 ----
 
-== Titles, headings, and filenames
+== Titles, Headings, and Filenames
 There are some general guidelines and conventions we follow for titles,
 headings, and topic filenames. These are described in the following table.
 
@@ -61,7 +61,14 @@ would be appropriate for a topic titled Managing Authorization Policies.
 a|- Use title case in all topic titles and section headings. See
 http://titlecase.com/ for more information.
 - Try to be as descriptive as possible with the topic title or section headings
-without making them unnecessarily too long.
+without making them unnecessarily too long. Use a verb form in headings. Some
+suggestions:
+* Creating
+* Managing
+* Using
+* Understanding
+- Use "Overview" as a heading sparingly. Choose a more descriptive heading
+instead.
 - The content in this repo is used to build docs that publish to multiple
 locations (e.g., docs.openshift.com, docs.openshift.org, and the Red Hat
 Customer Portal). While heading IDs will be created automatically from any
@@ -79,7 +86,7 @@ IDs use the following syntax:
 ----
 |===
 
-== Product name & version attributes
+== Product Name & Version Attributes
 
 When possible, generalize references to the product name and/or version using
 the `{product-title}` and/or `{product-version}` attributes. These attributes
@@ -139,7 +146,9 @@ separated, global use of `{product-version}` across all branches is probably
 less useful, but it is available if you come across a need for it. Just consider
 how it will render across any branches that the content appears in.
 
-== Links, hyperlinks, and cross references
+Do not use markup in headings.
+
+== Links, Hyperlinks, and Cross References
 Links can be used to cross-reference internal topics, or send customers to external information resources for further reading.
 
 In OpenShift docs:
@@ -255,7 +264,7 @@ xref:subtopic
 
 Note: There is no `#` used when linking to a subtopic within the same topic.
 
-== Sharing content between files
+== Sharing Content Between Files
 
 If you want to share content from one topic so that it appears in another topic,
 you can use the `include` directive. See the Asciidoctor documentation for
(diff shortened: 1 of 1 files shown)


contributing_to_docs/doc_guidelines.adoc
```diff
@@ -310,6 +391,23 @@ For all of the system blocks including table delimiters, use four characters. Fo
 ---- for code blocks
 ....
 
+Use this format when embedding footnotes into the code block:
```
adellape: On the Customer Portal builds, having `====` wrapped around it doesn't actually make a difference. On docs.openshift builds, it does show a small line outline around the whole thing, e.g.:

https://docs.openshift.com/container-platform/3.3/admin_guide/manage_authorization_policy.html#managing-role-bindings

Since it only makes a difference in one of the builds, and since it only yields a small cosmetic change at all, I'd suggest for simplicity to just drop using `====` around any codeblocks, callouts or no. And I never liked the previous distinction of our usage being "command syntax" vs "example command/output" either. So yeah, I suggest we just stick with `----` for all codeblocks, no matter the context, and just be done with it.

contributing_to_docs/doc_guidelines.adoc
```diff
@@ -344,6 +442,18 @@ $ lorem.sh
...
 
+For long lines of code that you want to break up among multiple lines, use a backslash to show the line break. For example:
+
+----
+$ oc get endpoints --all-namespaces --template \
+    '{{ range .items }}{{ .metadata.namespace }}:{{ .metadata.name }} \
+    {{ range .subsets }}{{ range .addresses }}{{ .ip }} \
+    {{ end }}{{ end }}{{ "\n" }}{{ end }}' | awk '/ 172\.30\./ { print $1 }'
+----
+
```
adellape: We should talk about the situation where there's a callout at the end of a line, and the syntax for that.

adellape (commented):
Dumping my comments so far, though I still need to get to the "Quick markup reference" section at the end.

contributing_to_docs/doc_guidelines.adoc
```diff
@@ -61,7 +61,8 @@ would be appropriate for a topic titled Managing Authorization Policies.
 a|- Use title case in all topic titles and section headings. See
 http://titlecase.com/ for more information.
 - Try to be as descriptive as possible with the topic title or section headings
-without making them unnecessarily too long.
+without making them unnecessarily too long. Use a verb form in headings (for example,
+Creating, Managing, Using, etc.) whenever possible.
```
bfallonf: I like the idea behind "Understanding". I think a lot of topics could use that and it might be a good idea for a house for the reference type stuff we need to include sometimes.

There might be a danger of overlap between "Understanding" sections and the "Overview" section. Should we be defining what should go into an "Overview" section somewhere? (Do we do that somewhere already? I should really check before I make this comment...)

contributing_to_docs/doc_guidelines.adoc
```diff
@@ -268,6 +271,84 @@ another topic, see the following usage:
 
 http://asciidoctor.org/docs/user-manual/#include-multiple
 
+== Writing concepts
+Concepts should support the things that users want to do and should not include
```
bfallonf: @aheslin I can't find the doc you showed me the other day (and Ashley seems to have put most of it in this topic), but was there an example for a concept topic in there? Further to what Alex says above, an example could help define the idea.

contributing_to_docs/doc_guidelines.adoc
```diff
@@ -268,6 +271,84 @@ another topic, see the following usage:
...
+== Writing tasks
+Write tasks in the following format.
+
+Task Title: Use a verb in the task title (for example, Create or Creating).
+
+Include a paragraph explaining why the user must perform this task. This should be 1-2 sentences maximum.
+
+.About This Procedure
+
+Include any gotchas (things that could trip up the user or cause the task to fail).
```
bfallonf: I had some questions about this as well when @aheslin initially showed it to me. Maybe "About this procedure" is not the correct title for the section? Maybe something as straight-forward as "Warning"? "Pre-Troubleshooting"?

contributing_to_docs/doc_guidelines.adoc
bfallonf: 1000% agree

contributing_to_docs/doc_guidelines.adoc
adellape: "About This Procedure" does seem a bit broad; it's not immediately obvious from the title that the section is going to be about gotchas. Maybe "Precautions"... "Procedure Tips"... "Considerations"... ?

contributing_to_docs/doc_guidelines.adoc
adellape: There's been some casual talk recently about not having "Overview" sections (as we've specifically known/required them in our guidelines previously) just to tick a box. Though of course there are definite places for overview-style material. Agree that we should try to define a bit what that sort of overview material should look like.

Deferring to @aheslin though re: the "Overview" topic requirement specifically; I think we should try to get those sentiments down on paper or at least hash it out a little more (seems like a good time w/ these guidelines updates).

contributing_to_docs/doc_guidelines.adoc
aheslin: I agree with both of you and I'll work on getting an example that's more applicable to OpenShift.

contributing_to_docs/doc_guidelines.adoc
aheslin: After talking to Brice, I don't really know that we gain anything by adding this in tasks. I think it will confuse people. I think having 1-3 sentences about why the user would want to perform the task is sufficient and we can include things like gotchas there. If there procedures where there are a lot of things for the user to consider, I'd suggest linking to a separate section. I'm open to suggestions though.

contributing_to_docs/doc_guidelines.adoc
```diff
@@ -268,6 +271,84 @@ another topic, see the following usage:
...
+
+* A bulleted list of pre-requisites that MUST be performed before the user can complete this task. Skip if there isn't any related information.
+
+.Procedure
+
+. Step 1 - One command per step.
+
+. Step 2 - One command per step.
+
+. Step N
```
aheslin: That's just example text to call it out. Mostly we're trying to avoid multiple steps and extra info in a command type step.

contributing_to_docs/doc_guidelines.adoc
```diff
@@ -139,6 +140,8 @@ separated, global use of `{product-version}` across all branches is probably
 less useful, but it is available if you come across a need for it. Just consider
 how it will render across any branches that the content appears in.
 
+Markup, such as in commands, persists in headings.
```
ahardin-rh: @adellape yeah, it does render strangely. I can revise this to state that we should not include markup in headings. As long as we're consistent, I'm happy! :smiley:

contributing_to_docs/doc_guidelines.adoc
```diff
@@ -268,6 +271,84 @@ another topic, see the following usage:
...
+.After You Finish
+
+You can explain any other tasks that MUST be completed after this task. You can skip this if there are none.
+
+.Related Information
+
+* A bulleted list of links to related information about this task. Skip if there isn't any related information.
+
+== Writing user goals (with numbered steps)
+Use the following format.
```
ahardin-rh: yeah, maybe we can omit this for now

contributing_to_docs/doc_guidelines.adoc
tpoitras: I'd wager that most readers just quickly scan the overview. I don't think it's a good place to put crucial information.

contributing_to_docs/doc_guidelines.adoc
tpoitras: What sorts of things would necessitate markup in headings? I think we should avoid putting file names or variable names into headings wherever possible.

vikram-redhat:
@ahardin-rh - Hey Ashley - looks like this should be good to go. Can you make the changes as suggested and merge?

ahardin-rh:
All comments are now addressed.

PR #2991, opened 2016-10-05
document new git noproxy field

bparees:


bparees:
@openshift/team-documentation ptal.

vikram-redhat:
@bfallonf - PTAL. @bparees - is this for dedicated or online as well?

bparees:
Anything running 3.4 and above.

Ben Parees | OpenShift

bfallonf:
Thanks @bparees I don't think this needs a followup. I'll merge.

bfallonf:
I don't think rev history is needed.

Issue #1947, opened 2016-04-26
iptables.adoc should talk about firewalld


danwinship:
iptables.adoc talks about saving and restoring iptables rules if you have to restart iptables.service, but it would be better if it said to use firewalld to manage your persistent firewall rules rather than iptables.service, since OpenShift/kubernetes/docker all integrate seamlessly with firewalld (and will handle recreating their iptables rules as needed if firewalld restarts, etc).

right @knobunc?

knobunc:
That is reasonable.

We should also make the rules we add to iptables refresh in the same way that the core kubernetes ones do.

eparis:
https://docs.openshift.com/container-platform/3.3/admin_guide/sdn_troubleshooting.html#is-code-firewalld-code-disabled should be fixed to not suggest that firewalld is somehow bad.

knobunc:
This is being worked on as https://github.com/openshift/openshift-docs/pull/3000

knobunc:
This is being worked on as https://github.com/openshift/openshift-docs/pull/3000

knobunc:
The PR has merged

PR #3000, opened 2016-10-05
Document iptables interaction with Openshift

pecameron:
Openshift 3.3 and 3.4 feature

iptables, firewalls, Openshift, Docker all interact with the kernel
iptables. None of the packages knows about what the other packages
have done or need to do.

It is very easy to misconfigure iptables and break Openshift and/or
docker networking.  These changes document the risks and some of the
interactions between packages.

https://trello.com/c/k6TgLcXS/307-3-document-un-allowed-iptables-mucking-with-behavior-sdn-supportability
openshift/openshift-docs: Issue #1947

Signed-off-by: Phil Cameron pcameron@redhat.com

### DIFF
--- a/admin_guide/iptables.adoc
+++ b/admin_guide/iptables.adoc
@@ -11,18 +11,67 @@
 toc::[]
 
 == Overview
-This topic describes how administrators should work with iptables. openshift-sdn takes care of adding the
-necessary iptables rules to make it work. Kubernetes and Docker also manage iptables for port forwarding
-and services.
+There are many system components including {product-title}, Docker, and software that manages local
+firewall policies that rely on the kernel iptables configuration for proper network operation. In
+addition the iptables configuration of all nodes in the cluster must be correct for networking to work.
 
+All components independently work with iptables without knowledge of how other components
+are using them. This makes it very easy for one component to break another component's configuration.
+Further, {product-title} and Docker assume that iptables remains set up exactly as they have set it up.
+They may not detect changes introduced by other components and if they do there maybe some lag in
+implementing the fix. In particular, {product-title} does monitor and fix problems, however, Docker does not.
 
-== Restarting
-Docker doesn't monitor the iptables rules that it adds for exposing ports from containers and hence if
-iptables service is restarted, then these rules are lost. So, to safely restart iptables, it is
-recommended that the rules are saved and restored. 
+IMPORTANT: You must make sure that whatever changes you make to the iptables configuration on a node
+do not impact the operation of {product-title} and Docker. Also, changes will often need to be made on
+all nodes in the cluster. Be careful, iptables is not designed to have multiple concurrent users and it
+is very easy to break {product-title} and Docker networking.
+
+The chains, order of the chains and rules in the kernel iptables must be properly set up on each node
+in the cluster for {product-title} and Docker networking to work properly. There are several tools and
+services that are commonly used in the system that interact with the kernel iptables and can
+accidentally impact {product-title} and Docker.
+
+=== iptables
+`iptables` is a low level tool that can be used to set up, maintain, and inspect the tables of IPv4
+packet filter rules in the Linux kernel.
+
+Independent of other use, such as a firewall, {product-title} and Docker manage chains in some of the tables.
+The chains are inserted in specific order and the rules are specific to their needs.
+
+IMPORTANT: `iptables --flush [chain]` can remove key required configuration. Don't execute this command.
+
+=== iptables.service
+The iptables service supports a local network firewall. It assumes total control of the iptables
+configuration and when it starts it flushes and restores the complete iptables configuration.
+The restored rules are from its configuration file (/etc/sysconfig/iptables).
+The config file is not kept up to date during operation so every restart the dynamically added rules are lost.
+
+WARNING: Stopping and starting the iptables.service will destroy configuration that is required by
+{product-title} and Docker. {product-title} and Docker are not notified of the change.
 
 ----
-$ iptables-save > /path/to/iptables.bkp
-$ systemctl restart iptables
-$ iptables-restore < /path/to/iptables.bkp
+# systemctl disable iptables.service
+# systemctl mask iptables.service
+----
+
+If you need to run iptables.service, keep a limited configuration in the config file and
+rely on Docker and openshift to install their needed rules.
+
+The iptables.service configuration is loaded from:
+
+----
+/etc/sysconfig/iptables
+----
+
+To make permanent rules changes edit the changes into this file. Do not include Docker or {product-title} rules.
+
+After iptables.service is started or restarted on a node, Docker and atomic-openshift-node.service
+must be restarted to reconstruct the needed iptables configuration.
+
+IMPORTANT: Restarting Docker will cause all containers running on the node to be stopped and restarted.
+
+----
+# systemctl restart iptables.service
+# systemctl restart docker
+# systemctl restart atomic-openshift-node.service
 ----
--- a/admin_guide/sdn_troubleshooting.adoc
+++ b/admin_guide/sdn_troubleshooting.adoc
@@ -11,7 +11,6 @@
 toc::[]
 
 == Overview
-
 As described in the
 xref:../architecture/additional_concepts/sdn.adoc#architecture-additional-concepts-sdn[SDN documentation]
 there are multiple layers of interfaces that are created to correctly
@@ -394,13 +393,6 @@ Of course the opposite traffic flows are also possible.
 ==== Is IP Forwarding Enabled?
 
 Check that `sysctl net.ipv4.ip_forward` is set to 1 (and check the host if this is a VM)
-
-==== Is `firewalld` Disabled?
-
-Check that `firewalld` is disabled using `systemctl status firewalld`.
-If it is running, you either need to disable it, or check
-that it is not blocking traffic.  That is outside the scope of this
-guide.
 
 ====  Are your routes correct?
 
--- a/getting_started/administrators.adoc
+++ b/getting_started/administrators.adoc
@@ -39,6 +39,12 @@ Docker and {product-title} must run on the Linux operating system. If you wish t
 run the server from a Windows or Mac OS X host, you should download and run the
 Origin Vagrant image as described in xref:building-from-source[Method 3].
 
+[CAUTION]
+====
+{product-title} and Docker use xref:../admin_guide/iptables.adoc#[iptables] to manage networking. Be careful that local firewall
+rules and other software that makes iptable changes do not alter the {product-title} and Docker setup.
+====
+
 [[installation-methods]]
 == Installation Methods
 


pecameron:
@knobunc @eparis @danwinship Here is a pass at documenting iptables. PTAL

admin_guide/iptables.adoc
```diff
@@ -1,5 +1,5 @@
-[[admin-guide-iptables]]
-= IPtables
+[[admin-guide-firewalld]]
```
knobunc: Do we want to change this?  It means any links to the anchor will break.  @adellape, what's the recommendation?

admin_guide/iptables.adoc
```diff
@@ -11,18 +11,89 @@
 toc::[]
 
 == Overview
-This topic describes how administrators should work with iptables. openshift-sdn takes care of adding the
-necessary iptables rules to make it work. Kubernetes and Docker also manage iptables for port forwarding
-and services.
+There are many system components including {product-title}, Docker, and the Firewall that rely on
```
knobunc: "The Firewall"?  I know what youy mean, but it's not immediately clear.

 How about "the local firewall policies"?

admin_guide/iptables.adoc
```diff
@@ -11,18 +11,89 @@
...
+the kernel iptables configuration for proper network operation. In addition the iptables
+configuration of all nodes in the cluster must be correct for networking to work.
 
+All components independently work with iptables without knowledge of how other components
+are using them. This makes it very easy for one component to break another component's configuration.
+Further, {product-title} and Docker assume that iptables remains set up exactly as they have set it up.
+They do not detect problems introduced by other components.
```
knobunc: s/problems/changes/

Also, OpenShift does monitor and fix, but there can be some lag.  Docker doesn't monitor.

admin_guide/iptables.adoc
```diff
@@ -11,18 +11,89 @@
...
 
-== Restarting
-Docker doesn't monitor the iptables rules that it adds for exposing ports from containers and hence if
-iptables service is restarted, then these rules are lost. So, to safely restart iptables, it is
-recommended that the rules are saved and restored. 
+IMPORTANT: You must make sure that whatever changes you make to the iptables configuration on a node
+do not impact the operation of {product-title} and Docker. Also, changes will frequently need to be made on
```
knobunc: "will frequently"?  Why?  Or do you mean the same change needs to be applied across the cluster affecting all nodes?

If so, we need to talk about how to do it safely across the cluster.

admin_guide/iptables.adoc
```diff
@@ -11,18 +11,89 @@
...
+all nodes in the cluster. Be careful, the setup is very complex and it is very easy to break {product-title}
+and Docker networking.
+
+The chains, order of the chains and rules in the kernel iptables must be properly set up on each node
+in the cluster for {product-title} and docker networking to work properly. There are several tools and
```
knobunc: s/docker/Docker/ ?

admin_guide/iptables.adoc
```diff
@@ -11,18 +11,89 @@
...
+
+command is executed both Openshift and Docker are notified so they can restore any needed rules.
+
+Detailed instructions for working with firewalld are beyond the scope of this document.
+
+
+=== iptables.service
+The iptables service supports a network firewall. It assumes total control of the iptables configuration
+and it flushes and restores the complete iptables configuration. It is not integrated with either Docker
+or {product-title} so when it is started it breaks cluster networking.
```
knobunc: when it is started or reloaded it restores a saved snapshot of the firewall rules and usually breaks cluster networking.

admin_guide/iptables.adoc
```diff
@@ -11,18 +11,89 @@
...
+The iptables.service configuration is loaded from:
+
+----
+/etc/sysconfig/iptables
+----
+
+To make permanent rules changes edit the changes into this file. Do not include Docker or {product-title} rules.
+
+After iptables.service is started on a node, docker must be restarted and atomic-openshift-node.service
+must be restarted to reconstruct the needed iptables configuration. Restarting docker will cause all containers
```
knobunc: Please emphasize that part about restarting docker.

admin_guide/iptables.adoc
```diff
@@ -11,18 +11,89 @@
...
+running on the node to be stopped and restarted.
 
 ----
-$ iptables-save > /path/to/iptables.bkp
-$ systemctl restart iptables
-$ iptables-restore < /path/to/iptables.bkp
+# systemctl restart iptables.service
+# systemctl restart docker
+# systemctl restart atomic-openshift-node.service
```
knobunc: I think we also need to lay out the alternate way of saving everything to /etc/sysconfig/iptables when running, editing that, and restarting iptables.service.  That doesn't necessitate a restart of docker, but we should advise against it because it will make the riles file hideous and impossible to understand.

admin_guide/sdn_troubleshooting.adoc
```diff
@@ -398,9 +397,9 @@ Check that `sysctl net.ipv4.ip_forward` is set to 1 (and check the host if this
 ==== Is `firewalld` Disabled?
```
knobunc: I think we can just remove this section.  Or at least tell them how to see if firewalld is being used and point them elsewhere for docs.

architecture/core_concepts/pods_and_services.adoc
```diff
@@ -225,6 +225,10 @@ installing], or after deploying the cluster by editing the
 xref:../../install_config/master_node_configuration.adoc#install-config-master-node-configuration[node
 configuration file].
 
+IMPORTANT: xref:../../admin_guide/iptables.adoc#[iptables] has many uses in addition to their use in {product-title}
+including the user's firewall and docker.  Care must be taken to make sure that iptable rule
```
knobunc: docker or Docker?  Pick one throughout the docs.

getting_started/administrators.adoc
```diff
@@ -39,6 +39,12 @@ Docker and {product-title} must run on the Linux operating system. If you wish t
 run the server from a Windows or Mac OS X host, you should download and run the
 Origin Vagrant image as described in xref:building-from-source[Method 3].
 
+[CAUTION]
+====
+{product-title} and Docker use iptables to manage networking. Be careful
+that firewall and other software, when used, does not alter the setup.
```
knobunc: firewall isn't software...  Perhaps "local firewall rules, or other software that makes changes to iptables"

admin_guide/iptables.adoc
pecameron: I grepped for admin-guide-iptables and got no hits.

admin_guide/iptables.adoc
pecameron: OK

admin_guide/iptables.adoc
pecameron: OK

admin_guide/iptables.adoc
pecameron: Its, "often". How to safely operate is discussed later.

admin_guide/iptables.adoc
pecameron: OK - will always be "Docker"

admin_guide/iptables.adoc
pecameron: OK

admin_guide/iptables.adoc
pecameron: OK

admin_guide/iptables.adoc
pecameron: I am reluctant to do this. I think that in an active cluster with a continuous flow of containers being started and stopped iptables will be in constant churn. It will be almost impossible to do a save/edit/restore without missing some dynamic changes. Packets that arrive after the flush and before the sate is fully restored may be dropped or end up in an unexpected replica. There is a lot of down side and not much up side doing this.

admin_guide/sdn_troubleshooting.adoc
pecameron: OK

architecture/core_concepts/pods_and_services.adoc
pecameron: OK - "Docker"

pecameron: OK

getting_started/administrators.adoc
pecameron: OK

pecameron:
@knobunc Here are the changes PTAL

admin_guide/iptables.adoc
```diff
@@ -11,18 +11,92 @@
...
+IMPORTANT: iptables --flush [chain] can remove key required configuration. Be very careful when using this.
+
+When making changes consider limiting your changes to the OS_FIREWALL_ALLOW chain in the filter table.
+
+
+=== firewalld
+
+When you need to set up a firewall on the cluster's nodes, use firewalld.
+Firewalld can be used to safely manage access to the cluster in a way that
+does not confiict with {product-title} or Docker. For example, when a
```
dcbw: "conflict"

dcbw:
Generally LGTM

admin_guide/iptables.adoc
pecameron: done

admin_guide/iptables.adoc
knobunc: Openshift knows how to repair that damage... the real problem is that docker doesn't, and those rules are more static.

knobunc:
@pecameron as discussed, for now we need to not recommend firewalld since the installer doesn't do it, and I bet it has not been heavily QAed.  So let's make this be about how to better use iptables today and cover firewalld when the installer supports it.

pecameron:
@knobunc I removed the references to firewalld. 
The install_config/install/host_preparation.adoc file talks about explicitly installing iptables.services
It looks like the ansible install approach has not be fully integrated into the doc.

install_config/install/host_preparation.adoc
```diff
@@ -80,7 +80,7 @@ For RHEL 7 systems:
 . Install the following base packages:
 +
 ----
-# yum install wget git net-tools bind-utils iptables-services bridge-utils bash-completion
+# yum install wget git net-tools bind-utils firewalld firewall-config bridge-utils bash-completion
```
knobunc: Is this still supposed to be here?

knobunc (changes requested):

knobunc:
Other than the nit, LGTM.  @adellape PTAL

install_config/install/host_preparation.adoc
pecameron: fixed... forgot to save the file :-(

admin_guide/iptables.adoc
```diff
@@ -11,18 +11,76 @@
...
+implementing the fix. In particular, {product-title} does monitor and fix problems, however, Docker does not.
 
-== Restarting
-Docker doesn't monitor the iptables rules that it adds for exposing ports from containers and hence if
-iptables service is restarted, then these rules are lost. So, to safely restart iptables, it is
-recommended that the rules are saved and restored. 
+IMPORTANT: You must make sure that whatever changes you make to the iptables configuration on a node
+do not impact the operation of {product-title} and Docker. Also, changes will often need to be made on
+all nodes in the cluster. Be careful, the setup is very complex and it is very easy to break {product-title}
+and Docker networking.
```
danwinship: The reason it is easy to break OpenShift networking is not because "the setup is very complex", it's because iptables.service is not designed to be used in an environment where other processes are also modifying the iptables rules.

admin_guide/iptables.adoc
```diff
@@ -11,18 +11,76 @@
...
+
+The chains, order of the chains and rules in the kernel iptables must be properly set up on each node
+in the cluster for {product-title} and Docker networking to work properly. There are several tools and
+services that are commonly used in the system that interact with the kernel iptables and can
+accidentally impact {product-title} and Docker.
+
+=== iptables
+Iptables is a low level tool that can be used to set up, maintain, and inspect the tables of IPv4
```
danwinship: When referring to the iptables command itself, put it in lowercase and with backticks around it: `iptables`

admin_guide/iptables.adoc
```diff
@@ -11,18 +11,76 @@
...
+
+WARNING: This is a very powerful tool that can manage all aspects of the kernel iptables. It assumes
+you know what you are doing.
+
+Independent of other use, such as a firewall, {product-title} and Docker manage chains in some of the tables.
+The chains are inserted in specific order and the rules are specific to their needs.
+
+IMPORTANT: iptables --flush [chain] can remove key required configuration. Be very careful when using this.
+
+When making changes consider limiting your changes to the OS_FIREWALL_ALLOW chain in the filter table.
```
danwinship: `OS_FIREWALL_ALLOW` only exists if you are running firewalld, and in that case firewalld owns that table and you shouldn't modify it directly.

admin_guide/iptables.adoc
```diff
@@ -11,18 +11,76 @@
...
+configuration and when it starts it flushes and restores the complete iptables configuration.
+The restored rules are from its configuration file (/etc/sysconfig/iptables).
+The config file is not kept up to date during operation so every restart the dynamically added rules are lost.
+
+WARNING: Stopping and starting the iptables.service will destroy configuration that is required by
+{product-title} and Docker. {product-title} and Docker are not notified of the change.
+
+WARNING: Use of iptables.service is not recommended in conjunction with {product-title} or Docker.
+
+Iptables.service can be disabled and masked to prevent accidental execution:
```
danwinship: It's a separate package from the iptables binary; you can just not install it.

admin_guide/iptables.adoc
```diff
@@ -11,18 +11,76 @@
...
+
+----
+# systemctl disable iptables.service
+# systemctl mask iptables.service
+----
+
+If you need to run iptables.service keep a limited configuration in the config file and
```
danwinship: comma before "keep"

admin_guide/iptables.adoc
```diff
@@ -11,18 +11,76 @@
...
+
+The iptables.service configuration is loaded from:
+
+----
+/etc/sysconfig/iptables
+----
+
+To make permanent rules changes edit the changes into this file. Do not include Docker or {product-title} rules.
+
+After iptables.service is started on a node, Docker must be restarted and atomic-openshift-node.service
```
danwinship: "started or restarted on a node"

danwinship: Also, I think unconditionally saying "atomic-openshift-node.service" is wrong. There's a variable (like with "{product-title}") or you need to conditionalize or something. I think.

admin_guide/iptables.adoc
danwinship: OpenShift currently knows how to repair that damage, but the trello card that this is part of explicitly calls for removing that functionality. Also, if you do this then you need to remember to delete the saved rules after restarting so they don't get re-reloaded later.

architecture/core_concepts/pods_and_services.adoc
```diff
@@ -225,6 +225,10 @@ installing], or after deploying the cluster by editing the
 xref:../../install_config/master_node_configuration.adoc#install-config-master-node-configuration[node
 configuration file].
 
+IMPORTANT: xref:../../admin_guide/iptables.adoc#[iptables] has many uses in addition to their use in {product-title}
+including the user's firewall and Docker.  Care must be taken to make sure that iptable rule
+changes don't break other components and that they are applied on all relevant nodes in the cluster.
```
danwinship: I don't think this belongs here

danwinship:
Is documenting how to use firewalld going to be a separate PR?

admin_guide/iptables.adoc
pecameron: OK I will change this.

admin_guide/iptables.adoc
pecameron: Changed

admin_guide/iptables.adoc
pecameron: The install part of the doc explicitly installs it. The intention is to move to firewalld when ansible is ready. so there will be more churn here and hopefully we can eventually just delete the discussion on iptables.service.

admin_guide/iptables.adoc
pecameron: Done

admin_guide/iptables.adoc
pecameron: "atomic-openshift-" appears in a lot of places in the doc. I couldn't find variable for it. Do you happen to know what to code here?

architecture/core_concepts/pods_and_services.adoc
pecameron: Deleted section.

pecameron:
@danwinship  @knobunc firewalld is not currently supported in the ansible installer (there is a card to do this). We are in a transition where iptables.service is explicitly installed (in the install section) and it is mentioned throughout. When we change to firewalld, there will need to be a lot of changes in the docs. 

We also need to decide whether to support both firewalld and iptables.service during a transition period for existing installations.

The card wants to remove iptables.service recovery software and I think that this is not the time to do that since iptables.service is the currently supported firewall. What do you think?

pecameron:
@danwinship @knobunc I made the changes, PTAL

knobunc (approved):
LGTM

@pecameron can you clarify what docs versions you want these changes to go into.

knobunc:
@adellape

pecameron:
@knobunc You have a "changes requested" set. Do you need more changes?

knobunc:
@pecameron are you sure?  I did a forced reload of the page and don't see that.

danwinship:
Hm... I hadn't realized that the ansible firewall rules were set up via iptables.service.

In that case I think we should install appropriate systemd mojo so that if iptables.service gets restarted, then docker and openshift get restarted afterward as well, and then just warn people that if they restart iptables.service, it will kill all of their running pods. (And suggest that they should perhaps therefore refrain from doing this.)

pecameron:
Dan,
The ansible installer folks are working on a card to switch to 
firewalld. The current discussion includes iptables.service related 
information because that is what we support. When the new firewalld 
install is done and we switch over a lot of the iptables.service 
discussion can go away. firewalld is already supported in the software, 
we just need to support it in the installer.
Phil

danwinship:
I know the ansible install doesn't currently use firewalld. But I had thought that it didn't use iptables.service either. (I thought it installed a script that loaded its firewall rules manually via /sbin/iptables calls.)

So what I was saying is that given that this is *not* the case, then a lot of my earlier reasoning (both in this PR and in the Trello card) doesn't make much sense; if we depend on iptables.service, then we can't go around telling people that we don't support it. So maybe we should be focused more on interoperating semi-sanely with iptables.service rather than telling people to not use it. Or alternatively, we need to more clearly delimit what things you are and aren't allowed to do with iptables.service.

pecameron:
@danwinship I took the explain how things work and how to work with them approach. I think there is value in what I have here and when the new firewalld install is ready we can revisit this. Since we ship both iptables.service and firewalld packages and permit people to install either independent of ansible installs I think we need to tell people how to work with both.

knobunc (approved):
LGTM.  @adellape PTAL

pecameron:
@knobunc @ramr I rebased upstream/master. Some formatting changes. PTAL

pecameron:
@knobunc @ramr Don't bother with this. I messed up the merge.

pecameron:
@knobunc @ramr merge is OK. This should be ready to go...

admin_guide/iptables.adoc
```diff
@@ -11,18 +11,76 @@
...
+in the cluster for {product-title} and Docker networking to work properly. There are several tools and
+services that are commonly used in the system that interact with the kernel iptables and can
+accidentally impact {product-title} and Docker.
+
+=== iptables
+`iptables` is a low level tool that can be used to set up, maintain, and inspect the tables of IPv4
+packet filter rules in the Linux kernel.
+
+WARNING: This is a very powerful tool that can manage all aspects of the kernel iptables. It assumes
+you know what you are doing.
```
sferich888: I'm not sure we should be wording this warning this way.

admin_guide/iptables.adoc
```diff
@@ -11,18 +11,76 @@
...
+
+IMPORTANT: `iptables --flush [chain]` can remove key required configuration. Be very careful when using this.
+
+When making changes consider limiting your changes to the OS_FIREWALL_ALLOW chain in the filter table.
+
+=== iptables.service
+The iptables service supports a local network firewall. It assumes total control of the iptables
+configuration and when it starts it flushes and restores the complete iptables configuration.
+The restored rules are from its configuration file (/etc/sysconfig/iptables).
+The config file is not kept up to date during operation so every restart the dynamically added rules are lost.
```
sferich888: Should we be recommended that the service be disabled? 

```
systemctl disable iptables.service
```

sferich888: It looks like this is added later.

admin_guide/iptables.adoc
pecameron: @sferich888 I will reword this. My background is Kernel guy running as root: very dangerous tools, no safety features. Have a nice day...
I will soften the message.

ramr:
LGTM - but will let @danwinship / @dcbw  verify that this is good.

admin_guide/iptables.adoc
```diff
@@ -11,18 +11,76 @@
...
+in the cluster for {product-title} and Docker networking to work properly. There are several tools and
+services that are commonly used in the system that interact with the kernel iptables and can
+accidentally impact {product-title} and Docker.
+
+=== iptables
+`iptables` is a low level tool that can be used to set up, maintain, and inspect the tables of IPv4
+packet filter rules in the Linux kernel.
+
+WARNING: This is a very powerful tool that can manage all aspects of the kernel iptables. It is very
+easy to break networking in difficult to debug ways so be careful.
```
danwinship: I still think this section is misleading; as far as I remember, *all* of the iptables-related problems we have seen have been caused by careless use of iptables.service, not by using /sbin/iptables directly.

admin_guide/iptables.adoc
```diff
@@ -11,18 +11,76 @@
...
+
+Independent of other use, such as a firewall, {product-title} and Docker manage chains in some of the tables.
+The chains are inserted in specific order and the rules are specific to their needs.
+
+IMPORTANT: `iptables --flush [chain]` can remove key required configuration. Be very careful when using this.
```
danwinship: What would "being very careful" entail? There's no way to use "--flush" in a way that won't break things. Just say "don't do it". (Or else refer to the section below about restarting docker and OpenShift in this case.)

admin_guide/iptables.adoc
```diff
@@ -11,18 +11,76 @@
...
+
+When making changes consider limiting your changes to the OS_FIREWALL_ALLOW chain in the filter table.
```
danwinship: That chain only exists if you used the advanced (ansible) install method, but at any rate, I don't think we need to say this anyway; our problem is not with people adding their own iptables rules, it's with people deleting ours.

admin_guide/iptables.adoc
```diff
@@ -11,18 +11,76 @@
...
+=== iptables.service
+The iptables service supports a local network firewall. It assumes total control of the iptables
+configuration and when it starts it flushes and restores the complete iptables configuration.
+The restored rules are from its configuration file (/etc/sysconfig/iptables).
+The config file is not kept up to date during operation so every restart the dynamically added rules are lost.
+
+WARNING: Stopping and starting the iptables.service will destroy configuration that is required by
+{product-title} and Docker. {product-title} and Docker are not notified of the change.
+
+WARNING: Use of iptables.service is not recommended in conjunction with {product-title} or Docker.
```
danwinship: Needs to be reworded in light of the fact that our default advanced installation *does* use iptables.service.

admin_guide/iptables.adoc
danwinship: If you are using the ansible-defined firewall, then you need iptables.service to be running, so you can't disable and mask it. If you are not using the ansible-defined firewall, then we recommend that you don't use iptables.service, and so you can uninstall it.

admin_guide/iptables.adoc
```diff
@@ -11,18 +11,76 @@
...
+The iptables.service configuration is loaded from:
+
+----
+/etc/sysconfig/iptables
+----
+
+To make permanent rules changes edit the changes into this file. Do not include Docker or {product-title} rules.
+
+After iptables.service is started on a node, Docker must be started or restarted and atomic-openshift-node.service
+must be restarted to reconstruct the needed iptables configuration.
```
danwinship: Sorry, my earlier comment here was ambiguous... this should be: After iptables.service is started or restarted on a node, Docker and atomic-openshift-node.service must be restarted to reconstruct the needed iptables configuration.

getting_started/administrators.adoc
```diff
@@ -39,6 +39,12 @@ Docker and {product-title} must run on the Linux operating system. If you wish t
 run the server from a Windows or Mac OS X host, you should download and run the
 Origin Vagrant image as described in xref:building-from-source[Method 3].
 
+[CAUTION]
+====
+{product-title} and Docker use iptables to manage networking. Be careful that local firewall
+rules and other software that makes iptable changes do not alter the {product-title} and Docker setup.
```
danwinship: should probably have a link to iptables.adoc

danwinship (changes requested):

admin_guide/iptables.adoc
pecameron: OK

admin_guide/iptables.adoc
pecameron: Deleted. We are getting into a "let the buyer beware" attitude. Ansible is/will be the inatall method of choice.

admin_guide/iptables.adoc
pecameron: OK

pecameron:
@danwinship I made the changes you suggested. PTAL

admin_guide/iptables.adoc
```diff
@@ -11,18 +11,67 @@
...
+The restored rules are from its configuration file (/etc/sysconfig/iptables).
+The config file is not kept up to date during operation so every restart the dynamically added rules are lost.
+
+WARNING: Stopping and starting the iptables.service will destroy configuration that is required by
+{product-title} and Docker. {product-title} and Docker are not notified of the change.
+
+----
+# systemctl disable iptables.service
+# systemctl mask iptables.service
+----
```
danwinship: I think you meant to delete this... the text that goes with them is gone now

admin_guide/sdn_troubleshooting.adoc
```diff
@@ -395,13 +394,6 @@ Of course the opposite traffic flows are also possible.
 
 Check that `sysctl net.ipv4.ip_forward` is set to 1 (and check the host if this is a VM)
 
-==== Is `firewalld` Disabled?
-
-Check that `firewalld` is disabled using `systemctl status firewalld`.
-If it is running, you either need to disable it, or check
-that it is not blocking traffic.  That is outside the scope of this
-guide.
```
danwinship: Actually, after helping someone deal with another firewalld-related issue, I realize we should keep this text for now; although we are compatible with firewalld in the general case, we are not compatible with its default configuration (which will block incoming connections that OpenShift wants). So for now the text is basically correct: if it is running, you need to make sure that it is not blocking traffic, and that is outside the scope of this guide.

danwinship (commented):
Basically LGTM

knobunc:
@adellape @ahardin-rh PTAL

knobunc:
@pecameron is there a reason this should not be in the 3.3 docs too?  I don't think any of this discusses new features.

pecameron:
@knobunc I think this is 3.3 as well as 3.4 I changed the description.

knobunc:
@pecameron thanks :+1:

pecameron:
@knobunc  @adellape is there something else to do here before this gets merged?

knobunc:
Nope, it's fine by me to merge as soon as the @openshift/team-documentation gets to it.

vikram-redhat:
@adellape - you're assigned to this, but if you don't have time this week, perhaps @ahardin-rh can take a look?

ahardin-rh:
Work continued in https://github.com/openshift/openshift-docs/pull/3000

pecameron:
@ahardin-rh Why was this closed and not merged? cc: @knobunc

knobunc:
@pecameron the docs team often takes the PR over and does the work under a new PR so that they can make tweaks and get a clean merge.  They do this for efficiency.  If you look at #3000 you will see that it is based on your PR.

ahardin-rh:
@pecameron Yeah, sorry for the confusion. I pulled your branch in locally and continued the work in https://github.com/openshift/openshift-docs/pull/3188 to speed along progress and make minor style edits.

PR #2980, opened 2016-10-04
Added rev history file to Admin Solutions guide

tpoitras:
as per @vikram-redhat

PR #3005, opened 2016-10-06
Adding revision history for administrator solutions guide for 3.3

tpoitras:
as per vikram


tpoitras:
replaces #2980 which was created in the wrong branch

PR #3006, opened 2016-10-06
Added guide on upgrading / updating the node OS

knobunc:
This adds recommendations on how to upgrade the OS of, or apply system
updates to, a node so that the iptables rules are not broken.

The purpose of this is both to warn administrators of the risks, but
to also provide a place to point people to in release notes and other
articles.

Fixes bug 1378693

For 3.3 and 3.4.

### DIFF
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -279,6 +279,8 @@ Topics:
         File: manual_upgrades
       - Name: Blue-Green Deployments
         File: blue_green_deployments
+      - Name: Operating System Updates and Upgrades
+        File: os_upgrades
   - Name: Downgrading
     File: downgrade
     Distros: openshift-enterprise
--- a/install_config/upgrading/os_upgrades.adoc
+++ b/install_config/upgrading/os_upgrades.adoc
@@ -1,0 +1,28 @@
+[[install-config-upgrading-os-upgrades]]
+= Overview
+{product-author}
+{product-version}
+:data-uri:
+:icons:
+:experimental:
+:prewrap!:
+
+When you need to update or upgrade the Operating System, either
+changing OS versions or just updates to the system software, you need
+to be aware that this can have an impact on the {product-title}
+software running on those machines.
+
+In particular, these updates can affect the `iptables` rules or `ovs`
+flows that {product-title} requires to operate.  To be safe, we
+recommend rebooting the node to ensure that all of the software is
+running the new versions (especially if a new kernel has been
+installed).  A reboot will also make sure that the `docker` and
+{product-title} processes have been restarted, which will force them
+to make sure that all of the rules in other services are correct.
+
+However, if you do not wish to do a reboot of the node, then it is
+possible to just restart the services that are affected, or to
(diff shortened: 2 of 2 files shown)


knobunc:
@adellape

install_config/upgrading/os_upgrades.adoc
```diff
@@ -0,0 +1,27 @@
...
+:icons:
+:experimental:
+:prewrap!:
+
+When you need to upgrade the Operating System, either changing OS
+versions or just updates to the system software, you need to be aware
+that this can have an impact on the {product-title} software running
+on those machines.
+
+In particular, these updates can affect the `iptables` rules that
```
danwinship: iptables isn't the only source of problems... if OVS gets updated, then the ovs db will get restarted, which also causes problems. Although the solution is basically the same as in the iptables case: restart the processes that care.

It looks like the docker daemon gets restarted on docker update too, which will kill any running pods.

Maybe we should just say "if you want to update, stop the node, update, and reboot"?

_topic_map.yml
```diff
@@ -279,6 +279,8 @@ Topics:
         File: manual_upgrades
       - Name: Blue-Green Deployments
         File: blue_green_deployments
+      - Name: Operating System Upgrades
```
danwinship: "upgrade" to me suggests an OS version change... maybe say "update" everywhere? What do other people think?

Also, are there other places we can link to this new topic from? I feel like no one's going to notice it...

install_config/upgrading/os_upgrades.adoc
knobunc: But we own OVS and update the rules.  The fundamental problem is that docker doesn't monitor and refresh its iptables rules... so if you get lucky and you snapshotted the handful of docker rules that matter, then things will work when you restart because openshift will fix the rest of the rules.

Anyway, I'll call ot ovs too.

_topic_map.yml
knobunc: It's a bit of a placeholder for reference from a kbase article and release notes for the OS.  But it would be nice if it were more prominent in the docs too. @adellape can you suggest somewhere better.

install_config/upgrading/os_upgrades.adoc
danwinship: We update the rules *if you restart OpenShift*. But if you just do `dnf update openvswitch` and then go away, then your node will break.

knobunc:
@adellape can we please get this merged so that we can reference it from the OS release notes and a kbase article.

pravisankar:
@knobunc This is good for warning the admins. 
We may not guarantee our software works if they do random updates/upgrades? I'm thinking of the case where we tested on one version(ovs or iptables) and then the user updated/upgraded to some other version which we didn't test.

knobunc:
@adellape @ahardin-rh Can someone please merge this?  I need to reference it from a kbase article.

ahardin-rh:
Work continued in https://github.com/openshift/openshift-docs/pull/3157

PR #3007, opened 2016-10-06
Publish scoped tokens

liggitt:
This should apply to all distributions


ahardin-rh:
Thanks!

ahardin-rh:
No rev history necessary

PR #3014, opened 2016-10-07
Fix deadlink

mhutter:


mhutter:
On hold, there are many more. Will reopen

PR #3018, opened 2016-10-07
Bug 1381336, deploy metrics as the deployer service account.

mwringe:
This fixes https://bugzilla.redhat.com/show_bug.cgi?id=1381336 where the user who deploys metrics could have special scc permissions which could prevent the deployer from functioning properly.

The deployer service account is the more correct user to be handling the deployer pod.

### DIFF
--- a/install_config/cluster_metrics.adoc
+++ b/install_config/cluster_metrics.adoc
@@ -582,7 +582,8 @@ The following command sets the Hawkular Metrics route to use
 You must have a persistent volume of sufficient size available.
 
 ----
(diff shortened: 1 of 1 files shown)


mwringe:
@adellape If you could take a look that would be greatly appreciated. This affects 3.3 and we have already had 2 bugzillas open about this.

mwringe:
@adellape any update on this? This really needs to be updated because its affecting customers in 3.3

ahardin-rh:
@mwringe LGTM. Thanks! Flipping to QE in the BZ. Sorry for the delay.

PR #2975, opened 2016-10-03
more detailed seccomp documentation

pweil-:
https://bugzilla.redhat.com/show_bug.cgi?id=1379761 provide more detailed documentation about seccomp alpha feature.

@sttts @pmorie please tech review

@vikram-redhat

PR #3020, opened 2016-10-10
Bug 1379761 moved and edited seccomp info

bfallonf:
Followup to #2975 

@pweil- I moved the seccomp info to the admin guide, and put the Examples section more into the Configuring section, but kept the Custom Profile stuff on the outside. WDYT?

I do have some questions on areas I think can be elaborated on though. I'll put them inline. Thanks!


admin_guide/seccomp.adoc
```diff
@@ -0,0 +1,127 @@
...
+====
+
+Seccomp is used to restrict the set of system calls applications can make,
+allowing administrators greater control over the security of workloads running
+in {product-title}.
+
+Seccomp support is achieved via two annotations in the pod configuration:
+
+* *seccomp.security.alpha.kubernetes.io/pod*: profile applies to all containers in the pod that do not override
+* *container.seccomp.security.alpha.kubernetes.io/<container name>*: container-specific profile override
```
bfallonf: @pweil- I'm not sure what the above are. Why are we configuring pods here? Is this something you have to configure before you can go through the tasks below?

admin_guide/seccomp.adoc
```diff
@@ -0,0 +1,127 @@
...
+[NOTE]
+====
+Seccomp is a feature of the Linux kernel. To ensure seccomp is enabled on your
+system run:
+
+----
+$ cat /boot/config-`uname -r` | grep CONFIG_SECCOMP=
+CONFIG_SECCOMP=y
+----
+====
```
bfallonf: Should this be inside the topics below first? Or would this be more at home in something like a "Enabling Seccomp" section instead of a admonition?

admin_guide/seccomp.adoc
```diff
@@ -0,0 +1,127 @@
...
+to take when a syscall is invoked. The
+link:https://github.com/docker/docker/blob/master/profiles/seccomp/default.json[default
+profile] is sufficient in many cases, the cluster administrator must define the
+security constraints of an individual system.
++
+To create your own custom profile, create a file on every node in the
+`seccomp-profile-root` directory.
++
+If you are using the default *runtime/default* profile, you do not need to
+create one.
```
bfallonf: The above was taken over from the Examples section, but now I think I've confused it all. Is it that the file exists by default? Or is it that you have to take the file from the example inline above, so that you don't have to create it? Sorry if this is a confusing question...

admin_guide/seccomp.adoc
```diff
@@ -0,0 +1,127 @@
...
+. In order to control which profiles may be used and to set the default profile,
+xref:../architecture/additional_concepts/authorization.adoc#authorization-seccomp[configure
+your SCC] via the *seccompProfiles* field. The first profile will be used as a
+default.
++
+When configuring seccomp profiles to use, follow these formatting rules:
+
+* *runtime/default*: the default profile for the container runtime (no profile on disk required)
+* *unconfined*: unconfined profile, ie, no seccomp sandboxing
+* *localhost/<profile-name>*: the profile installed to the node's local seccomp profile root
```
bfallonf: I'm not sure what the above is saying. Are they example profiles? can they be substituted into the step above this one?

admin_guide/seccomp.adoc
```diff
@@ -0,0 +1,127 @@
...
++
+For example, if you are using the default *runtime/default* profile, configure the *restricted* SCC with:
++
+----
+seccompProfiles:
+- runtime/default
+----
+
+[[seccomp-configuring-openshift-with-custom-seccomp]]
+==  Configuring {product-title} for a Custom Seccomp Profile
```
bfallonf: Please let me know if anything below is incorrect.

admin_guide/seccomp.adoc
pweil-: both of the annotations belong at the pod level.

admin_guide/seccomp.adoc
pweil-: Since this is not related to OpenShift and is more about the operating system I called it out separately.  I could see it going in the section below if you'd prefer it there.

admin_guide/seccomp.adoc
pweil-: these are the allowable formats of the annotation values.  `runtime/default` is the container runtime's default, built-in profile and doesn't require a file.  `unconfined` is turning off seccomp and doesn't require a file.  `localhost/<name>` is pointing to a specific file on the node.  The `localhost` portion tells the system that it is, in fact, a file that needs read and the name corresponds to the actual file name that would be stored in the seccomp profile root that was configured on the kubelet.

admin_guide/seccomp.adoc
```diff
@@ -0,0 +1,127 @@
...
+kubeletArguments:
+  seccomp-profile-root:
+    - "/your/path"
+----
+
+. Configure *seccomp-profile-root*:
++
+----
+seccompProfiles:
+- runtime/default
```
pweil-: this would be the `localhost/<name>` option if you wanted to use a custom profile.  If you're just using the `runtime/default` you would not need to set the seccomp-profile-root.

admin_guide/seccomp.adoc
```diff
@@ -0,0 +1,127 @@
...
+----
+
+.  Restart the node host.
```
pweil-: this should be done after the kubelet args are changed.  They just need to restart the node service(s), not the whole host.

bfallonf:
@pweil- Thanks for the prompt comments. I've made the changes I think need to happen and that you've suggested, but kept some of them because your comments helped them make sense to me.

Could I ask for a final ack that all is well with the topic? Then I can move it to review. Thanks!

admin_guide/seccomp.adoc
```diff
@@ -0,0 +1,149 @@
...
+
+. Create the seccomp profile in *seccomp-profile-root*:
++
+----
+kubeletArguments:
+  seccomp-profile-root:
+    - "/your/path"
+----
+
+. Configure *seccomp-profile-root*:
```
pweil-: This is a duplicate step of configuring the restricted SCC

pweil-:
LGTM

admin_guide/seccomp.adoc
bfallonf: @pweil- So should i delete this step in favour of the above one?

bfallonf:
Thanks, @pweil- 

@ahardin-rh @adellape @tpoitras Should be ok to peer review now if anyone wants. Just got one thing to sort out above.

bfallonf:
@pweil- I made change to what i _think_ you suggested. Please let me know if I'm wrong.

@ahardin-rh @tpoitras @adellape totally ready for peer review.

Thanks, everybody.

pweil-:
> I made change to what i think you suggested. Please let me know if I'm wrong.

@bfallonf - here is what I would suggest.  Everything looks good except for the example of creating a profile in the root which I think can be removed.  It's a file on the host machines:

```
To ensure pods in your cluster run with a custom profile in the *restricted* SCC:

. Create the seccomp profile in *seccomp-profile-root* on all nodes:
+

. Configure *seccomp-profile-root*:
+
----
kubeletArguments:
  seccomp-profile-root:
    - "/your/path"
----

. Restart node service to apply the changes:

ifdef::openshift-enterprise[]
----
# systemctl restart atomic-openshift-node
----
endif::[]
ifdef::openshift-origin[]
----
# systemctl restart origin-node
----
endif::[]

* Configure the *restricted* SCC:
+
----
seccompProfiles:
- localhost/<profile-name>
----
```

ahardin-rh:
@bfallonf just one comment from me :star:

_topic_map.yml
```diff
@@ -496,6 +494,8 @@ Topics:
   - Name: Securing Builds by Strategy
     File: securing_builds
     Distros: openshift-origin,openshift-enterprise
+  - Name: Seccomp
```
tpoitras: Is this a sufficient title? Should you use the long form of "Secure Computing Mode"?

admin_guide/seccomp.adoc
```diff
@@ -0,0 +1,149 @@
...
+== Overview
+
+[IMPORTANT]
+====
+Seccomp is
+link:https://github.com/kubernetes/kubernetes/blob/release-1.4/docs/design/seccomp.md[currently
+in alpha support].
+====
+
+Seccomp is used to restrict the set of system calls applications can make,
```
tpoitras: Maybe the first instance you could say something like "Secure computing mode, or Seccomp, is used to..."

tpoitras: s/calls applications/calls that applications

admin_guide/seccomp.adoc
```diff
@@ -0,0 +1,149 @@
...
+
+For detailed design information, refer to the
+link:https://github.com/kubernetes/kubernetes/blob/release-1.4/docs/design/seccomp.md[seccomp
+design document].
+
+[[seccomp-enabling-seccomp]]
+== Enabling Seccomp
+
+Seccomp is a feature of the Linux kernel. To ensure seccomp is enabled on your
+system run:
```
tpoitras: comma after "system"?
Maybe not, I tend to over-use the comma.

admin_guide/seccomp.adoc
```diff
@@ -0,0 +1,149 @@
...
+$ cat /boot/config-`uname -r` | grep CONFIG_SECCOMP=
+CONFIG_SECCOMP=y
+----
+
+[[seccomp-configuring-openshift-for-seccomp]]
+== Configuring {product-title} for Seccomp
+
+. Create the seccomp profile. 
++
+A seccomp profile is a json file providing syscalls and the appropriate action
```
tpoitras: s/json/JSON
or
s/json/.json

tpoitras: what is a "syscall"? Is this defined elsewhere?

admin_guide/seccomp.adoc
```diff
@@ -0,0 +1,149 @@
...
+to take when a syscall is invoked. The
+link:https://github.com/docker/docker/blob/master/profiles/seccomp/default.json[default
+profile] is sufficient in many cases, the cluster administrator must define the
```
tpoitras: s/in many cases, the cluster/in many cases. The cluster...

^ This should be two separate sentences, no?

admin_guide/seccomp.adoc
```diff
@@ -0,0 +1,149 @@
...
+
+[IMPORTANT]
+====
+Seccomp is
+link:https://github.com/kubernetes/kubernetes/blob/release-1.4/docs/design/seccomp.md[currently
+in alpha support].
+====
+
+Seccomp is used to restrict the set of system calls applications can make,
+allowing administrators greater control over the security of workloads running
```
tpoitras: allowing cluster administrators

admin_guide/seccomp.adoc
```diff
@@ -0,0 +1,149 @@
...
+
+. Configure *seccomp-profile-root*:
++
+----
+kubeletArguments:
+  seccomp-profile-root:
+    - "/your/path"
+----
+
+. Restart node service to apply the changes:
```
tpoitras: Restart the node service...

tpoitras:
@bfallonf - found about a dozen things here during peer review. :octopus:

admin_guide/seccomp.adoc
```diff
@@ -0,0 +1,149 @@
...
+
+toc::[]
+
+== Overview
+
+[IMPORTANT]
+====
+Seccomp is
+link:https://github.com/kubernetes/kubernetes/blob/release-1.4/docs/design/seccomp.md[currently
+in alpha support].
```
pweil-: Alpha should be defined as https://github.com/openshift/origin#alpha-and-unsupported-kubernetes-features

nak3: @pweil-  What's the policy of RH? Does it mean technical preview?

> The feature may have significant bugs and is suitable for testing and prototyping.

It looks like not production ready feature.

pweil-: yes, tech preview

nak3: I see. Thank you!

bfallonf: Paul made a comment below that tech preview works.

_topic_map.yml
bfallonf: Hmm I'd changed to a better title already. Maybe I didn't push.

admin_guide/seccomp.adoc
bfallonf: From what I know, it's a linux thing...

bfallonf:
OK. Thanks for the comments, all. Looks like this is ready to merge.

bfallonf:
[rev_history]
|xref:../admin_guide/seccomp.adoc#admin-guide-seccomp[Restricting Application Capabilities Using Seccomp]
|Added the Restricting Application Capabilities Using Seccomp file.
%

adellape:
@bfallonf So this is Technology Preview for OCP 3.3 (vs what it may be in Origin)? This feature is listed in the OCP 3.3 Release Notes as a new feature (https://docs.openshift.com/container-platform/3.3/release_notes/ocp_3_3_release_notes.html#ocp-33-scc-profiles-seccomp), but we did not list it as a TP feature (https://docs.openshift.com/container-platform/3.3/release_notes/ocp_3_3_release_notes.html#ocp-33-technology-preview). If it is, we should also update the Release Notes.

cc @pweil-

bfallonf:
@adellape Hmm didn't see that. NP. I'll see what @pweil- says and update the docs if I need to.

pweil-:
@adellape @bfallonf this is considered an alpha api in upstream so it would be alpha downstream as well unless someone thinks differently.

@sttts - do you, by chance, know when this is moving out of alpha?

pweil-:
@adellape @bfallonf - double checked this with the BU, it is being considered supported so I think you are fine with the current documentation.  Will ping via IRC

sttts:
@pweil- what is missing for beta? Don't know about any discussion around seccomp beta.

pweil-:
@sttts nothing afaik other than agreement it is ready.  I'm not aware of any push for beta upstream either.  Maybe something we can push for in 1.6

sttts:
1.6 definitely makes sense. Can you create a feature issue for that? Who sponsors this? sig-node?

pweil-:
> 1.6 definitely makes sense. Can you create a feature issue for that? Who sponsors this? sig-node?

Created and tagged you.  I assume it would be sig-node.

sttts:
Should these docs go into the upstream docs repo as well?

pweil-:
> Should these docs go into the upstream docs repo as well?

In some form if they are not already in the psp documentation (covered by the PSP feature request).  However the seccomp field in SCC was introduced first so it is an actual field and not an annotation.

sttts:
Added links to the feature https://github.com/kubernetes/features/issues/135. I think the only beta blocker are the missing docs.

adellape:
@bfallonf @vikram-redhat I think because this PR (and https://github.com/openshift/openshift-docs/pull/2975) make changes to the topic_map and an all-distro shared Arch topic, that this A) needs to get picked to dedicated-3.3/online eventually and B) needs to be ifdef'd a little bit more than it is. WDYT?

It may need a follow-up PR to put a more specific distro on the topic_map entry, unless this is also relevant for Dedicated admins to know about. cc @spurtell 

And probably also ifdefs (depending on the answer to the above) around the small bit in `architecture/additional_concepts/authorization.adoc`, which currently links to the admin_guide (which doesn't exist for Online).

adellape:
This was picked to enterprise-3.3 this week and published, but I'm not sure whether to put this back into Next Release now w/ a `to_followup` for the other branches, or what. @vikram-redhat ?

bfallonf:
@adellape I fully trust you to do what you find best. Seems you have the info.

vikram-redhat:
Sorry @adellape - I haven't had the chance to look at this today (10/31). Will look at this tomorrow in detail.

adellape:
Removed enterprise-3.3 branch because it's been picked/published there already. Added dedicated-3.3 + online + to_followup labels and moved to the Next Release milestone for further conditionalizing for Dedicated/Online.

PR #2690, opened 2016-08-19
Add documentation for memory eviction, update documentation about swap

derekwaynecarr:
Adds the documentation for memory pressure based pod evictions that were done in kube 1.3 and are now in origin master.

cc @ncdc @pmorie @jeremyeder PTAL

PR #3023, opened 2016-10-11
Bug 1375350 Edits to OOR info

bfallonf:
As per: https://bugzilla.redhat.com/show_bug.cgi?id=1375350

But also, I went through all of #2690 because I wanted to get to some of the vague bits.

@derekwaynecarr  I'll put some comments in the PR. Can I get your thoughts? And of course if you have any comments on my changes.

Thanks!

cc: @ahardin-rh

### DIFF
--- a/admin_guide/out_of_resource_handling.adoc
+++ b/admin_guide/out_of_resource_handling.adoc
@@ -1,5 +1,5 @@
-[[admin-guide-out-of-resource-handling]]
-= Out of Resource Handling
+[[admin-guide-handling-out-of-resource-errors]]
+= Handling Out of Resource Errors
 {product-author}
 {product-version}
 :data-uri:
@@ -12,30 +12,46 @@ toc::[]
 
 == Overview
 
-The node must preserve node stability when available compute resources are
-low. This is especially important when dealing with incompressible resources
-such as memory or disk. If either resource is exhausted, the node becomes
-unstable.
+The node must preserve node stability when available compute resources are low.
+This is especially important when dealing with incompressible resources such as
+memory or disk. If either resource is exhausted, the node becomes unstable.
+
+[WARNING]
+====
+Failure to disable swap memory makes the node not recognize it is under
+*MemoryPressure*.
+
+To take advantage of memory based evictions, operators must
+xref:../admin_guide/overcommit.adoc#disabling-swap-memory[disable swap].
+====
 
 [[out-of-resource-eviction-policy]]
 == Eviction Policy
 
-The node can proactively monitor for and prevent against total starvation of a
-compute resource. In cases where this appears to occur, the node can
-proactively fail one or more pods in order to reclaim the starved resource.
-When the node fails a pod, it terminates all containers in the pod, and the
-`*PodPhase*` is transitioned to *Failed*.
+Using eviction policies, a node can proactively monitor for and prevent
+against total starvation of a compute resource.
+
+In cases where a node is running low on available resources, it can proactively
+fail one or more pods in order to reclaim the starved resource using an eviction
+policy. When the node fails a pod, it terminates all containers in the pod, and
+the `PodPhase` is transitioned to *Failed*.
 
 Platform administrators can configure eviction settings within the
 *_node-config.yaml_* file.
-xref:out-of-resource-schedulable-resources-and-eviction-policies[See an example
-scenario].
 
 [[out-of-resource-eviction-signals]]
 === Eviction Signals
-The node can support the ability to trigger eviction decisions on the signals
+
+The node can be configured to trigger eviction decisions on the signals
 described in the table below. The value of each signal is described in the
 description column based on the node summary API.
+
+To view the signals:
+
+----
+curl <certificate details> \
+  https://<master>/api/v1/nodes/<node>/proxy/stats/summary
+----
 
 .Supported Eviction Signals
 [cols="2a,10a",options="header"]
@@ -47,28 +63,41 @@ description column based on the node summary API.
 |`memory.available` = `node.status.capacity[memory]` - `node.stats.memory.workingSet`
 |===
 
+[NOTE]
+====
 In future releases, the node will support the ability to trigger eviction
 decisions based on disk pressure. Until then, use
 xref:../admin_guide/garbage_collection.adoc#admin-guide-garbage-collection[container
 and image garbage collection].
+====
 
 [[out-of-resource-eviction-thresholds]]
 === Eviction Thresholds
 
-The node supports the ability to specify eviction thresholds that trigger the
-node to reclaim resources.
+You can configure a node to specify eviction thresholds, which trigger the node
+to reclaim resources.
 
-Each threshold is of the following form:
+Eviction thresholds can be soft, for when you allow a grace period before
+reclaiming resources, and hard, for when the node takes immediate action when a
+threshold is met.
+
+Thresholds are configured in the following form:
 
 ----
 <eviction_signal><operator><quantity>
 ----
 
-* Valid `eviction-signal` tokens as defined in
-xref:out-of-resource-eviction-signals[Eviction Signals].
+* Valid `eviction-signal` tokens as defined by eviction signals.
 * Valid `operator` tokens are `<`.
-* Valid `quantity` tokens must match the quantity representation used by
+* Valid `quantity` tokens must match the link:https://github.com/kubernetes/kubernetes/blob/master/docs/design/resources.md#resource-quantities[quantity representation] used by
 Kubernetes.
+
+For example, using the `memory.available` signal, in order to construct a
+threshold for when the memory available drops below 500Mi, the form would be:
+
+----
+memory.available<500Mi
+----
 
 
 [[out-of-resource-soft-eviction-thresholds]]
@@ -82,18 +111,18 @@ grace period is provided, the node errors on startup.
 In addition, if a soft eviction threshold is met, an operator can specify a
 maximum allowed pod termination grace period to use when evicting pods from the
 node. If specified, the node uses the lesser value among the
-`*pod.Spec.TerminationGracePeriodSeconds*` and the maximum-allowed grace period.
+`pod.Spec.TerminationGracePeriodSeconds` and the maximum-allowed grace period.
 If not specified, the node kills pods immediately with no graceful termination.
 
 To configure soft eviction thresholds, the following flags are supported:
 
-* `eviction-soft` describes a set of eviction thresholds (for example,
+* `eviction-soft`: a set of eviction thresholds (for example,
 `memory.available<1.5Gi`) that, if met over a corresponding grace period,
 triggers a pod eviction.
-* `eviction-soft-grace-period` describes a set of eviction grace periods (for
+* `eviction-soft-grace-period`: a set of eviction grace periods (for
 example, `memory.available=1m30s`) that correspond to how long a soft eviction
 threshold must hold before triggering a pod eviction.
-* `eviction-max-pod-grace-period` describes the maximum-allowed grace period (in
+* `eviction-max-pod-grace-period`: the maximum-allowed grace period (in
 seconds) to use when terminating pods in response to a soft eviction threshold
 being met.
 
@@ -107,30 +136,41 @@ termination.
 
 To configure hard eviction thresholds, the following flag is supported:
 
-* `eviction-hard` describes a set of eviction thresholds (for example,
+* `eviction-hard`: a set of eviction thresholds (for example,
 `memory.available<1Gi`) that, if met, triggers a pod eviction.
+
+[[out-of-resource-oscillation-of-node-conditions]]
+=== Oscillation of Node Conditions
+
+If a node is oscillating above and below a soft eviction threshold, but not
+exceeding its associated grace period, the corresponding node condition
+oscillates between *true* and *false*, which can confuse the scheduler.
+
+To protect this, set the following flag to control how long the node must wait
+before transitioning out of a pressure condition:
+
+* `eviction-pressure-transition-period`: the duration that the node has
+to wait before transitioning out of an eviction pressure condition.
+
+Before toggling the condition back to *false*, the node ensures that it has not
+observed a met eviction threshold for the specified pressure condition for the
+period specified.
 
 [[out-of-resource-eviction-monitoring-interval]]
 === Eviction Monitoring Interval
 
-The node evaluates eviction thresholds per its housekeeping interval. It
-monitors every 10 seconds and the value can not be modified.
+The node evaluates and monitors eviction thresholds every 10 seconds and the
+value can not be modified. This is the housekeeping interval.
 
-* `housekeeping-interval` is the interval between container housekeepings.
 
-For more information about housekeeping, see the
-link:https://github.com/google/cadvisor/blob/master/docs/runtime_options.md[cAdvisor
-documentation].
-
-[[out-of-resource-node-conditions]]
-=== Node Conditions
+[[out-of-resource-mapping-eviction-signals-to-node-conditions]]
+=== Mapping Eviction Signals to Node Conditions
 
 The node can map one or more eviction signals to a corresponding node
 condition.
 
-If a hard eviction threshold is met, or a soft eviction threshold is met
-independent of its associated grace period, the node reports a condition
-indicating that the node is under pressure.
+If an eviction threshold is met, independent of its associated grace period, the
+node reports a condition indicating that the node is under pressure.
 
 The following node conditions are defined that correspond to the specified
 eviction signal.
@@ -146,26 +186,9 @@ eviction signal.
 |Available memory on the node has satisfied an eviction threshold.
 |===
 
-The node continues to report node status updates at the frequency specified by
-the `node-status-update-frequency` argument, which defaults to *10s*.
-
-[[out-of-resource-oscillation-of-node-conditions]]
-=== Oscillation of Node Conditions
-
-If a node is oscillating above and below a soft eviction threshold, but not
-exceeding its associated grace period, the corresponding node condition
-constantly oscillates between *true* and *false*, which can cause poor
-scheduling decisions.
-
-To protect against this oscillation, the following flag is defined to control
-how long the node must wait before transitioning out of a pressure condition.
-
-* `eviction-pressure-transition-period` is the duration for which the node has
-to wait before transitioning out of an eviction pressure condition.
-
-The node ensures that it has not observed an eviction threshold being met for
-the specified pressure condition for the period specified before toggling the
-condition back to *false*.
+When the above is set the node continues to report node status updates at the
+frequency specified by the `node-status-update-frequency` argument, which
+defaults to *10s*.
 
 [[out-of-resource-eviction-of-pods]]
 === Eviction of Pods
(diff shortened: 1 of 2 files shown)


admin_guide/out_of_resource_handling.adoc
```diff
@@ -13,26 +13,27 @@ toc::[]
...
+node fails a pod, it terminates all containers in the pod, and the `*PodPhase*`
+is transitioned to *Failed*.
 
 Platform administrators can configure eviction settings within the
 *_node-config.yaml_* file.
-xref:out-of-resource-schedulable-resources-and-eviction-policies[See an example
-scenario].
 
 [[out-of-resource-eviction-signals]]
 === Eviction Signals
```
bfallonf: I do think this could be changed to something like "How the node signals that it's almost full", (though that's a terrible suggestion). But really, the paragraph below seems to be describing a setting that says that. Any other suggestions?

admin_guide/out_of_resource_handling.adoc
```diff
@@ -13,26 +13,27 @@ toc::[]
...
+
 The node can support the ability to trigger eviction decisions on the signals
```
bfallonf: "The node can support the ability"

Does this mean it is not enabled by default? Is this what we're enabling here?

admin_guide/out_of_resource_handling.adoc
```diff
@@ -47,10 +48,13 @@ description column based on the node summary API.
 |`memory.available` = `node.status.capacity[memory]` - `node.stats.memory.workingSet`
 |===
```
bfallonf: Where are these signals actually given out? Is this in the logs? Which files exactly?

admin_guide/out_of_resource_handling.adoc
```diff
@@ -47,10 +48,13 @@ description column based on the node summary API.
...
+[NOTE]
+====
 In future releases, the node will support the ability to trigger eviction
 decisions based on disk pressure. Until then, use
 xref:../admin_guide/garbage_collection.adoc#admin-guide-garbage-collection[container
 and image garbage collection].
+====
 
 [[out-of-resource-eviction-thresholds]]
 === Eviction Thresholds
```
bfallonf: I think what's missing is an explanation of what an eviction threshold, and why you would want to use a soft one over a hard one.

bfallonf: With the below, and this is said a few times, but we're saying "The node supports the ability to X" a lot. I take that to mean something needs to be enabled in order to use it. Is that the case? If it's by default, it should just be saying "The node can X".

admin_guide/out_of_resource_handling.adoc
```diff
@@ -64,8 +68,7 @@ Each threshold is of the following form:
 <eviction_signal><operator><quantity>
 ----
```
bfallonf: Where is this found? A file somewhere?

admin_guide/out_of_resource_handling.adoc
```diff
@@ -64,8 +68,7 @@ Each threshold is of the following form:
...
 
-* Valid `eviction-signal` tokens as defined in
-xref:out-of-resource-eviction-signals[Eviction Signals].
+* Valid `eviction-signal` tokens as defined by eviction signals.
 * Valid `operator` tokens are `<`.
```
bfallonf: What does that mean? Is that the only valid value for the <operator> bit?

admin_guide/out_of_resource_handling.adoc
```diff
@@ -64,8 +68,7 @@ Each threshold is of the following form:
...
 * Valid `quantity` tokens must match the quantity representation used by
 Kubernetes.
```
bfallonf: How do you find that? Quantity of what?

admin_guide/out_of_resource_handling.adoc
```diff
@@ -87,13 +90,13 @@ If not specified, the node kills pods immediately with no graceful termination.
 
 To configure soft eviction thresholds, the following flags are supported:
```
bfallonf: With the below, I think an example would be a lot better... @derekwaynecarr Do you have one we could put into the docs?

admin_guide/out_of_resource_handling.adoc
```diff
@@ -116,21 +119,20 @@ To configure hard eviction thresholds, the following flag is supported:
 The node evaluates eviction thresholds per its housekeeping interval. It
 monitors every 10 seconds and the value can not be modified.
 
-* `housekeeping-interval` is the interval between container housekeepings.
+* `housekeeping-interval`: the interval between container housekeepings.
 
 For more information about housekeeping, see the
 link:https://github.com/google/cadvisor/blob/master/docs/runtime_options.md[cAdvisor
 documentation].
```
bfallonf: Is cAdvisor something supported by OpenShift? I've not heard of it before. Is there a better place we could link out to?

admin_guide/out_of_resource_handling.adoc
```diff
@@ -224,12 +204,66 @@ pods on the node.
 |`BestEffort` pods are not scheduled to the node.
 |===
```
bfallonf: I'm not sure what this is saying. Is it saying that the scheduler can read the eviction signal and do something accordingly? Is so, I'd move this above to the signal section.

admin_guide/overcommit.adoc
```diff
@@ -330,35 +330,34 @@ required.
...
-It is important to understand that oversubscribing the physical resources on a
-node affects resource guarantees the Kubernetes scheduler makes during pod
-placement. For example, suppose two guaranteed pods have reached their memory
-limit. Each container may start using swap. Eventually, if there is not enough
-swap space, processes in the pods can be terminated (due to the system being
-oversubscribed).
+You can disable swap by default on your nodes in order to preserve quality of
+service guarantees. Otherwise, physical resources on a node can oversubscribe,
+affecting the resource guarantees the Kubernetes scheduler makes during pod
+placement.
```
bfallonf: @derekwaynecarr This seems odd. It's as though having it enabled is a bad thing. Is there a reason why it's not just disabled by default instead of the user needing to do this??

bfallonf:
@derekwaynecarr ^ bump

admin_guide/out_of_resource_handling.adoc
derekwaynecarr: an example for a soft eviction and hard eviction threshold would be the following:

* operator wants to evict immediately if memory falls below <5%  (hard threshold)
* operator wants to evict if memory falls below <30% for 1 min (soft threshold)

in this scenario, the operator would like their machines to steady at 70% utilization, but are willing to go above that for short periods of time.  that is the general idea.  our ops team will typically use two thresholds for a similar reason.

derekwaynecarr: we have no default eviction thresholds enabled in 3.3, so a user today must opt-in to this behavior.  Meaning they need to setup the node to do this now.  we will get defaults in the future.

admin_guide/out_of_resource_handling.adoc
derekwaynecarr: this is basically describing a syntax for a threshold.

in 3.3, we just support literal thresholds (memory.available<100Mi)

in 3.4, we will support percentage thresholds as well (memory.available<10%)

so in 3.4, we will need to update this doc to reflect that both input styles are valid.

a sample node configuration is given in the "Example scenario" section.

admin_guide/out_of_resource_handling.adoc
```diff
@@ -64,8 +68,7 @@ Each threshold is of the following form:
 <eviction_signal><operator><quantity>
 ----
 
-* Valid `eviction-signal` tokens as defined in
-xref:out-of-resource-eviction-signals[Eviction Signals].
+* Valid `eviction-signal` tokens as defined by eviction signals.
```
derekwaynecarr: its meant to mean the signals in table 1 (which now lists one), but in 3.4 needs to list 4 more (all disk related)

admin_guide/out_of_resource_handling.adoc
derekwaynecarr: right now, < is the only supported operator, but there was discussion on potentially offering more.  i kept the doc this way to allow it to grow without major re-writes.

admin_guide/out_of_resource_handling.adoc
derekwaynecarr: this basically means 'the syntax you used to express a quantity anywhere in openshift/kube' - so whether its how you declare the amount of cpu or memory for a pod, a constraint on quota, a limit in a limit range, etc.  they all use the quantity representation.  maybe its just obvious.  we don't appear to have a good doc to describe: 
https://github.com/kubernetes/kubernetes/blob/master/docs/design/resources.md#resource-quantities

admin_guide/out_of_resource_handling.adoc
derekwaynecarr: can we hold on a soft eviction scenario until we have disk?  i think that will make more sense in that context for 3.4.

admin_guide/out_of_resource_handling.adoc
```diff
@@ -116,21 +119,20 @@ To configure hard eviction thresholds, the following flag is supported:
 The node evaluates eviction thresholds per its housekeeping interval. It
```
derekwaynecarr: As noted earlier, I think we should just say:

"The node evaluates eviction thresholds every 10s."

In the future, hard eviction thresholds for memory will not use polling every 10s, and instead we will have the kernel tell us the threshold has been passed and act immediately.  That is planned for Kubernetes 1.5 / Origin 1.5.

derekwaynecarr: so drop any mention of housekeeping interval.

admin_guide/out_of_resource_handling.adoc
```diff
@@ -116,21 +119,20 @@ To configure hard eviction thresholds, the following flag is supported:
...
 monitors every 10 seconds and the value can not be modified.
 
-* `housekeeping-interval` is the interval between container housekeepings.
+* `housekeeping-interval`: the interval between container housekeepings.
```
derekwaynecarr: just drop this since its not an actual flag that can be modified at this time.  sorry for the error on my original doc.

admin_guide/out_of_resource_handling.adoc
```diff
@@ -146,26 +148,9 @@ eviction signal.
 |Available memory on the node has satisfied an eviction threshold.
 |===
 
-The node continues to report node status updates at the frequency specified by
-the `node-status-update-frequency` argument, which defaults to *10s*.
-
-[[out-of-resource-oscillation-of-node-conditions]]
-=== Oscillation of Node Conditions
-
-If a node is oscillating above and below a soft eviction threshold, but not
```
derekwaynecarr: this is important.   the default eviction-pressure-transition-period is 5m0s, so if you see a node that reports MemoryPressure, it will wait 5m0s to actually report that MemoryPressure is not longer occuring.

admin_guide/out_of_resource_handling.adoc
```diff
@@ -174,31 +159,26 @@ If an eviction threshold is met and the grace period is passed, the node
...
-relative to their request for that resource are killed first. If no pod has
-exceeded its request, the strategy targets the largest consumer of the starved
-resource.
-* `Guaranteed` pods that consume the greatest amount of the starved resource
-relative to their request are killed first. If no pod has exceeded its request,
+* `Burstable`: pods that consume the most of the starved resource relative to their
+request for that resource are failed first. If no pod has exceeded its request,
 the strategy targets the largest consumer of the starved resource.
+* `Guaranteed`: pods that consume the most of the starved resource relative to
+their request are killed first. If no pod has exceeded its request, the strategy
```
derekwaynecarr: i like that you used 'failed' consistently above, we should do the same here and s/killed/failed

admin_guide/out_of_resource_handling.adoc
```diff
@@ -174,31 +159,26 @@ If an eviction threshold is met and the grace period is passed, the node
...
+targets the largest consumer of the starved resource. 
 
 A `Guaranteed` pod is guaranteed to never be evicted because of another pod's
-resource consumption.
-
-If:
+resource consumption if:
```
derekwaynecarr: The if is confusing here.

A `Guaranteed` pod is guaranteed to never be evicted as a result of another pod's resource consumption.

The next section is trying to say:

The only reason a `Guaranteed` pod could be evicted is as follows:

* a system daemon...
* ...

admin_guide/out_of_resource_handling.adoc
```diff
@@ -174,31 +159,26 @@ If an eviction threshold is met and the grace period is passed, the node
...
 
 * A system daemon (for example, node, *docker*, *journald*, or others) is
 consuming more resources than were reserved via *system-reserved* or
-*kube-reserved* allocations, and
+*kube-reserved* allocations
```
derekwaynecarr: i think it should be "and the node only has Guaranteed pods remaining" in the same sentence.

admin_guide/out_of_resource_handling.adoc
```diff
@@ -174,31 +159,26 @@ If an eviction threshold is met and the grace period is passed, the node
...
 * The node only has `Guaranteed` pods remaining
 
 Then:
```
derekwaynecarr: basically, I am trying to explain that the only rationale thing to do when a node is running only guaranteed pods but system services are consuming too much resource is to fail the guaranteed pods since i cant really fail node system services.

admin_guide/out_of_resource_handling.adoc
derekwaynecarr: I would not move this to the signal section.

The list of reported Node conditions will grow in 3.4 to include DiskPressure.

This table is saying the following:

The scheduler looks at the NodeConditions reported by the node, and if it sees the node reporting "MemoryPressure" it will not place BestEffort pods on that node.

In 3.4, if the scheduler sees nodes that report "DiskPressure", it will not schedule any pods to that node.

The list of pressure conditions will grow, and the scheduler will do something slightly different for each.

derekwaynecarr: so the scheduler does NOT read eviction signals, it reads Node Conditions that are driven based on the configured eviction thresholds.  So for example, if i set an eviction threshold like the following:

eviction-hard is "memory.available<500Mi"

if available memory falls below that value, the Node has a value reported in `Node.Status.Conditions[]` whose `Type` will be `MemoryPressure` and whose `Status` will be `True`.  It's that value on the node object that the scheduler integrates with when making scheduling decisions.

admin_guide/out_of_resource_handling.adoc
derekwaynecarr: I would not nest this below Troubleshooting.  I would keep it after Eviction Thresholds in a dedicated section.

admin_guide/out_of_resource_handling.adoc
```diff
@@ -251,79 +285,41 @@ of service for the pod.
...
-[[out-of-resource-best-practices]]
-== Best Practices
+[[out-of-resource-recommended-practices]]
+== Recommended Practices
 
 [[out-of-resource-best-practice-disable-swap-memory]]
 === Disable Swap Memory
 
-Failure to disable swap memory makes the node not recognize it is under
+Failure to disable swap memory makes the node not recognize it under
```
derekwaynecarr: s/it/it's

I think this is so important that it should be called out maybe in the top of the document with something like ensuring your node has been configured correctly.

admin_guide/overcommit.adoc
derekwaynecarr: It is a bad thing, but some customers want to have it enabled because they used that feature to meet certain densities in v2.  unfortunately, swap being enabled means you can use other features.

@sdodson -- would it be bad for us to disable swap by default in our install, and instead write doc to discuss how it could be turned back on if desired?  is that something we can look to do in 3.4/3.5?

derekwaynecarr (changes requested):
Thanks for improving the documentation, please address the comments.

admin_guide/overcommit.adoc
sdodson: Yeah, it's not hard, just consensus building. We'll try to get it in the first update after 3.4. https://github.com/openshift/openshift-docs/pull/3023

bfallonf: Sounds good. The initial BZ is about this contradiction, so I'll check if that's enough for Eric ( @TheDiemer ) and continue with the rest of the comments. Thanks, all.

admin_guide/out_of_resource_handling.adoc
bfallonf: Changed to "the node can be configured to..." cos 3.3 is the current.

admin_guide/out_of_resource_handling.adoc
bfallonf: OK. I get you. I made a link out to the docs above, but you're right, it might be a good idea for us on docs to doc this at some point.

admin_guide/out_of_resource_handling.adoc
bfallonf: I did some minor rewrites so that it's specified that 10 seconds is the housekeeping interval.

bfallonf: Scratch that, after your next couple comments I scrapped it all instead.

bfallonf:
@derekwaynecarr Thanks for taking a look. I've made edits, pretty much to what you suggested. Can I get a final ack there's nothing else before I move forward with this? Thanks, again.

bfallonf:
@derekwaynecarr ^ Bump. (Thanks!)

derekwaynecarr:
@bfallonf -- this is a big improvement.  LGTM

bfallonf:
Big thanks @derekwaynecarr ! Good to see it's an improvement. Can I ask you to please approve the changes? Seems that new GitHub review feature means people need to give the thumbs up by clicking a button.

@adellape @ahardin-rh Any comments before I merge?

derekwaynecarr (approved):

derekwaynecarr:
@bfallonf -- approved changes, if we can merge this today I can send my updates for the disk-eviction support in 1.4 origin release.

bfallonf:
@derekwaynecarr Sure thing. Thanks much. I'll get this merged. If this has been reviewed before tomorrow morning, I can maybe get someone in BNE to take a look.

cc @adellape @ahardin-rh

admin_guide/out_of_resource_handling.adoc
```diff
@@ -12,31 +12,47 @@ toc::[]
...
-proactively fail one or more pods in order to reclaim the starved resource.
-When the node fails a pod, it terminates all containers in the pod, and the
-`*PodPhase*` is transitioned to *Failed*.
+Using eviction policies, a node can proactively monitor for and prevent
+against total starvation of a compute resource.
+
+In cases where a node is running low on available resources, it can proactively
+fail one or more pods in order to reclaim the starved resource using an eviction
+policy. When the node fails a pod, it terminates all containers in the pod, and
+the `*PodPhase*` is transitioned to *Failed*.
```
ahardin-rh: With the updated doc guidelines, we can apply the new formatting here:

    `PodPhase`

admin_guide/out_of_resource_handling.adoc
```diff
@@ -174,45 +197,39 @@ If an eviction threshold is met and the grace period is passed, the node
...
-The node reports a condition when a compute resource is under pressure. The
-scheduler views that condition as a signal to dissuade placing additional
-pods on the node.
+The scheduler views node conditions when placing additional pods on the node. For example, if the node has an eviction threshold like the following:
+
+----
+eviction-hard is "memory.available<500Mi"
+----
+
+and available memory falls below 500Mi, the node reports a value in `*Node.Status.Conditions*` as `*MemoryPressure*` as true.
```
ahardin-rh: Apply updated formatting here:

    `Node.Status.Conditions`
    `MemoryPressure`

admin_guide/out_of_resource_handling.adoc
```diff
@@ -224,64 +241,15 @@ pods on the node.
...
-[[out-of-resource-best-practice-disable-swap-memory]]
-=== Disable Swap Memory
-
-Failure to disable swap memory makes the node not recognize it is under
-*MemoryPressure*.
-
-To take advantage of memory based evictions, operators must
-xref:../admin_guide/overcommit.adoc#disabling-swap-memory[disable swap].
-
+This means that if the scheduler sees the node reporting "MemoryPressure" it
```
ahardin-rh: `MemoryPressure`

admin_guide/out_of_resource_handling.adoc
```diff
@@ -312,18 +280,58 @@ This configuration ensures that the scheduler does not place pods on a node that
...
-It is never desired for a node to evict a pod that was derived from
-a `*DaemonSet*` since the pod will immediately be recreated and rescheduled
-back to the same node.
+[[out-of-resource-node-out-of-resource-and-out-of-memory]]
+== Out of Resource and Out of Memory
+
+If the node experiences a system out of memory (OOM) event before it is able to
+reclaim memory, the node depends on the OOM killer to respond.
+
+The node sets a `*oom_score_adj*` value for each container based on the quality
```
ahardin-rh: `oom_score_adj`

admin_guide/out_of_resource_handling.adoc
```diff
@@ -312,18 +280,58 @@ This configuration ensures that the scheduler does not place pods on a node that
...
+of service for the pod.
+
+.Quality of Service OOM Scores
+[cols="3a,8a",options="header"]
+|===
+
+| Quality of Service |`*oom_score_adj*` Value
```
ahardin-rh: `oom_score_adj`

admin_guide/out_of_resource_handling.adoc
```diff
@@ -224,64 +241,15 @@ pods on the node.
...
-=== Disable Swap Memory
-
-Failure to disable swap memory makes the node not recognize it is under
-*MemoryPressure*.
-
-To take advantage of memory based evictions, operators must
-xref:../admin_guide/overcommit.adoc#disabling-swap-memory[disable swap].
-
+This means that if the scheduler sees the node reporting "MemoryPressure" it
+will not place BestEffort pods on that node.
```
ahardin-rh: `BestEffort`

ahardin-rh:
@bfallonf just a few minor comments from me regarding updated style guidelines. :star:

bfallonf:
Thanks @ahardin-rh . I'll pay more attention to the new guidelines... I'll merge away!

PR #3046, opened 2016-10-13
Fix Typo and add PORT ENVs

ghost:
Fix Typo.
The ROUTER_SERVICE_\* ports are necessary for router sharding.


ahardin-rh:
Thank you! This was previously addressed in https://github.com/openshift/openshift-docs/pull/3050

PR #3082, opened 2016-10-20
ScheduledJobs docs

soltysh:
@ahardin-rh ptal, I'll open one another for 3.3 branch with an extra comment about limitations.


ahardin-rh:
Thanks! Continuing work in https://github.com/openshift/openshift-docs/pull/3088

PR #3087, opened 2016-10-20
Fix broken links in router/index

superdump:
I have not been able to test this due to having too old ruby and not seeing a good way to update it on macOS. It should be theoretically correct. 😄


ahardin-rh:
Thanks! Work continued in https://github.com/openshift/openshift-docs/pull/3115

PR #3097, opened 2016-10-24
document build nodeselector and annotation configuration

bparees:


bparees:
@openshift/team-documentation ptal/merge.

vikram-redhat:
@bfallonf - PTAL.

bfallonf:
@bparees This looks largely pretty great. Big thanks.

There is a bit of overlap between what's here and what's in the Deployment section though:
https://docs.openshift.com/container-platform/3.3/dev_guide/deployments.html#assigning-pods-to-specific-nodes

It's sorta the same process, but with deployments switched with builds. Is that really the only difference? I might be able to play with some of the context info so we're not repeating ourselves... 

Other than that, thanks, again!

bparees:
@bfallonf yes it's basically the same concepts.  If you can find a way to refactor the content to a common location and then reference it from both sections that would be fine.

but it's important to have an actual example of how to define a nodeselector within a buildconfig, so you can't refactor too much of it out.

bfallonf:
@bparees Sure thing. Thanks. I'll leave it as suggested for now, but it might be something I think about for the future. And you're correct that an example for both is a good idea.

Thanks again. I'll merge.

ahardin-rh:
Rev history in follow-up PR

PR #3111, opened 2016-10-25
add flag for SA

mjudeikis:
Add flag for assigning access to SA


ahardin-rh:
This still looks valid: http://docs.projectatomic.io/registry/latest/registry_quickstart/developers.html

Atomic only. No revision history needed.

PR #3112, opened 2016-10-25
Document how to reconfigure clusterNetworkCIDR/serviceNetworkCIDR

danwinship:
For 3.4 we need to fully document how to manually reconfigure your network (https://trello.com/c/mb9sKIUR)

@openshift/networking PTAL

### DIFF
--- a/install_config/configuring_sdn.adoc
+++ b/install_config/configuring_sdn.adoc
@@ -164,6 +164,78 @@ Check VNIDs by running:
 $ oc get netnamespace
 ----
 
+[[renumbering-sdn-networks]]
+== Renumbering the SDN Network
+
+It is difficult to change the networking configuration of an
+{product-title} cluster after it has been deployed, but if absolutely
+necessary, it can be done in some cases.
+
+=== Expanding the Cluster IP Range By Adding Nodes
+
+The easiest reconfiguration involves changing the
+`*clusterNetworkCIDR*` to make room to add additional nodes, while
+leaving existing Nodes and Pods untouched. (That is, changing
+`*clusterNetworkCIDR*` to a new value which is larger than the old
+value, and which includes the old value as a subset.)
+
+For example, with the default `*clusterNetworkCIDR*` value of
+*10.128.0.0/14* and the default `*hostSubnetLength*` of *9*, you can
+have 512 nodes, with Pod subnets from *10.128.0.0/23* to
+*10.131.254.0/23*. (Older versions of {product-title} had smaller
+defaults.) If you changed `*clusterNetworkCIDR*` to *10.128.0.0/13*,
+then you could have an additional 512 nodes, with Pod IPs going up to
+*10.135.255.255*.
+
+To do this, simply change the `*clusterNetworkCIDR*` value in the
+`*networkConfig*` stanza of the *_master-config.yaml_* file, and then
+restart the master; for single master clusters:
+
+----
+# systemctl restart atomic-openshift-master
+----
+
+For multi-master clusters, on each master:
+
+----
+# systemctl restart atomic-openshift-master-controllers
+----
+
+You will also need to restart the node service on the master and on
+each of the nodes, to make them update their IP routing information to
+take the larger cluster network into account:
+
+----
+# systemctl restart atomic-openshift-node
+----
+
+As stated above, this only works when the new `*clusterNetworkCIDR*`
+value fully contains the original value; if you try to change it to
+any other value, then when you restart the master, it will complain
+about the mismatch and refuse to start.
+
+=== Other Changes to the Cluster or Service Network IP Range
+
+If you want to completely change `*clusterNetworkCIDR*` (eg, moving
+from *10.128.0.0/14* to *192.168.9.0/8*) or you want to change
+`*hostSubnetLength*` (to allow more or fewer pods per node) or
+`*serviceNetworkCIDR*`, simply changing the configuration and
+restarting will not work.
+
+ifndef::openshift-enterprise[]
+Although it is theoretically possible to make such changes in a
+deployed cluster, it is not supported and not documented here. While
+{product-name} may support such changes in the future, for now the
+best solution is to reinstall the cluster with the new correct values.
+endif::[]
+ifdef::openshift-enterprise[]
+Although it is possible to make such changes in a deployed cluster, it
+is complicated and not fully supported; if it all possible, you should
+consider reinstalling the cluster instead. FIXME talk to your support
+person or look up the answer in the support db or something, I'm not
+sure what to write here...
+endif::[]
+
 [[external-access-to-the-cluster-network]]
 == External Access to the Cluster Network
 


install_config/configuring_sdn.adoc
```diff
@@ -164,6 +164,181 @@ Check VNIDs by running:
...
+
+It is difficult to change the networking configuration of an
+{product-title} cluster after it has been deployed, but if absolutely
+necessary, it can be done.
+
+=== Expanding the Cluster IP Range By Adding Nodes
+
+The easiest reconfiguration involves changing the
+`*clusterNetworkCIDR*` to make room to add additional nodes, while
+leaving existing Nodes and Pods untouched.
```
knobunc: Can we make this a little more clear, perhaps by saying something like:

"In order to use this method, the new range must completely contain the original clusterNetworkCIDR."

install_config/configuring_sdn.adoc
```diff
@@ -164,6 +164,181 @@ Check VNIDs by running:
...
+defaults.) If you changed `*clusterNetworkCIDR*` to *10.128.0.0/13*,
+then you could have an additional 512 nodes, with Pod IPs going up to
+*10.135.255.255*.
+
+To do this, simply change the `*clusterNetworkCIDR*` value in the
+`*networkConfig*` stanza of the *_master-config.yaml_* file, and then
+restart the master; for single master clusters:
+
+----
+# systemctl restart atomic-openshift-master
```
knobunc: @adellape Does this need to be conditionalized (or does a variable need to be used?)

install_config/configuring_sdn.adoc
```diff
@@ -164,6 +164,181 @@ Check VNIDs by running:
...
+----
+
+For multi-master clusters:
```
knobunc: (on each machine running the master?)

install_config/configuring_sdn.adoc
```diff
@@ -164,6 +164,181 @@ Check VNIDs by running:
...
+
+You will also need to restart the node service on the master and on
+each of the nodes, to make them update their IP routing information to
+take the larger cluster network into account:
+
+----
+# systemctl restart atomic-openshift-node
+----
+
+Note that this only works if you are changing `*clusterNetworkCIDR*`
```
knobunc: Ah, you say it here.  I think it's worth saying it at the top too, in case they leap into the procedure without reading all of it.

install_config/configuring_sdn.adoc
```diff
@@ -164,6 +164,181 @@ Check VNIDs by running:
...
+case.
+
+=== Other Changes to the Cluster Network IP Range
+
+If you want to completely change `*clusterNetworkCIDR*` (eg, moving
+from *10.128.0.0/14* to *192.168.9.0/8*) or you want to change
+`*hostSubnetLength*` (to allow more or fewer pods per node), you will
+need to completely take down the cluster temporarily.
+
+. On every node and master, do:
```
knobunc: @nhr tells me there should be no space after the . to be good form.

http://asciidoctor.org/docs/user-manual/#sidebar

install_config/configuring_sdn.adoc
```diff
@@ -164,6 +164,181 @@ Check VNIDs by running:
...
++
+----
+# systemctl start atomic-openshift-master-controllers
+# systemctl start atomic-openshift-master-api
+----
+
+. Restart the node service on all of the nodes and masters:
++
+----
+# systemctl start atomic-openshift-node
```
knobunc: Do we want to restart ovs before this?  Or at least delete the bridge with our rules to get a clean slate?  I know we will work when we boot, but extra cruft would be left behind.

install_config/configuring_sdn.adoc
```diff
@@ -164,6 +164,181 @@ Check VNIDs by running:
...
+----
+
+Unlike in the `*clusterNetworkCIDR*`-renumbering case, this will not
```
knobunc: What deletes them?

Could we (later) provide a script that assigns new addresses from the new pool?  Or is there no way to run that between when the old service range was active and when you restart and the new one is?

knobunc (commented):
Looks great.  Thanks Dan!

I just have some minor questions, but it's thorough and clear (at least to me).

install_config/configuring_sdn.adoc
```diff
@@ -164,6 +164,181 @@ Check VNIDs by running:
...
+----
+# systemctl restart atomic-openshift-node
+----
+
+Note that this only works if you are changing `*clusterNetworkCIDR*`
+to a new value which is larger than the old value, and which includes
+the old value as a subset; if you try to change it to any other value,
+then when you restart the master, it will complain about the mismatch
+and refuse to start. You will need to use the procedure below in this
+case.
```
pecameron: Would it be useful to move this discussion up higher since it lays out context?

pecameron: Should we mention that additional nodes can be created and point to how to do it?

install_config/configuring_sdn.adoc
pecameron: Question: if we reduce the network space to less than the existing nodes (if we have 300 nodes and go from /23 to /24) does openshift just remove the extra nodes?

install_config/configuring_sdn.adoc
```diff
@@ -164,6 +164,181 @@ Check VNIDs by running:
...
+If you need to change `*serviceNetworkCIDR*`, either because you need
+more than 65,536 service IPs, or because you have belatedly discovered
+a conflict between the service IP range and other IPs on your network,
+you will have to take down the cluster (as above), but you will also
+need to delete and recreate all Service objects. There is no simple
+way to do this, and in particular, to get all of the permissions
+correct, each Service will have to be recreated by the correct user;
+you can't simply have an administrator recreate them. Thus, this will
+result in a more noticeable outage than the `*clusterNetworkCIDR*`
+renumbering case.
```
pecameron: Is it worth mentioning that as each service is recreated it will get a new address from the new serviceNetworkCIDR? There may be other required changes in the apps when the service ip changes.

install_config/configuring_sdn.adoc
```diff
@@ -164,6 +164,181 @@ Check VNIDs by running:
...
+
+. Restart the node service on all of the nodes and masters:
++
+----
+# systemctl start atomic-openshift-node
+----
+
+Unlike in the `*clusterNetworkCIDR*`-renumbering case, this will not
+kill previously-running pods. However, it leaves all Services deleted,
+and users will have to recreate them after this.
```
pecameron: And any part of the cluster configuration that relies on the service address, e.g., pulling from the openshift registry  will need to be updated in the DCs and RCs as well. Applications configurations that rely on service ip addresses will need to be changed as well. It can take quite a lot of work to get everything back up and running.

install_config/configuring_sdn.adoc
pecameron: @knobunc what is our position on debris left behind when we do this? Is the admin who is setting up something unrelated expected debug why the obvious setup is not working and clean up whatever gets in his way? How does the admin differentiate what is debris from what is needed? Just curious...

pecameron (commented):
Lots of good info here. Changes like this can impact the cluster is a lot of ways. I think pointing out the impact along the way would be helpful. This is in the spirit of "doing this may/will  impact "x" so look _here_ for how to handle it."

install_config/configuring_sdn.adoc
danwinship: That's copied from elsewhere in the docs

install_config/configuring_sdn.adoc
danwinship: > What deletes them?

The instructions above (`oc delete services --all`). I'll clarify.

> Could we (later) provide a script that assigns new addresses from the new pool?

Maybe... but we're supposed to be providing actual support for live renumbering in 1.5 or 1.6 or something anyway

install_config/configuring_sdn.adoc
danwinship: No, although they'd fail to get assigned new HostSubnets, so they'd never become ready. But obviously you should not do that anyway. I'll mention that.

install_config/configuring_sdn.adoc
weliang1: Do we need adding one seesion about changing hostSubnetLength mention in our trello card?

weliang1:
Did a quick test in v3.4.0.21 after updating both  clusternetworkCIDR and servicenetworkworkCIDR, router pod failed with below error:
https://10.121.224.1:443/api/v1/namespaces/https/replicationcontrollers/router1-1: x509: certificate is valid for 10.18.41.92, 172.30.0.1, not 10.121.224.1

Same testing passed in v3.3.1.4

weliang1:
Sometimes can an not ping node tunnel’s ip address after updating hostSubnetLength, 
rebooting both master and node will recover ping

danwinship:
> Sometimes can an not ping node tunnel’s ip address after updating hostSubnetLength,
> rebooting both master and node will recover ping

(more details in https://bugzilla.redhat.com/show_bug.cgi?id=1393083)

It looks like this is because if you change hostSubnetLength but not clusterNetworkCIDR, then the SDN doesn't get recreated on node restart. Fixing this by adding `ovs-vsctl del-br br0` to the procedure, to force the SDN to be recreated.

weliang1:
"ovs-vsctl del-br br0" step fix the ping issue mentioned above.

One more question, should "ovs-vsctl del-br br0" delete old iptables rules for old clusterNetworkCIDR?
The testing result I got so far is iptables keep adding new rules for updating clusterNetworkCIDRs, but it does not remove the rules for old clusterNetworkCIDR.

danwinship:
> @nhr tells me there should be no space after the . to be good form.

That doesn't match the existing style in the file, and it also doesn't generate the right output.

nhr:
@danwinship Are you using AsciiBinder to test this? What does the output look like whn you omit the space?

danwinship:
Updated for most of the earlier comments, plus addressed @weliang1's ping problem, plus some additional problems with changing the service network. In particular, kubernetes automatically recreates the "kubernetes" service any time you delete it, which means that if you change delete all services, change servicesNetworkCIDR to something non-overlapping, and restart, then the master will fail to start because there is a Service object with an IP that doesn't match servicesNetworkCIDR. So you end up needing to run the master briefly with no network plugin (since the check for services-outside-serviceNetworkCIDR happens in openshift-sdn code), fix things up, and then restart it with the correct network plugin again. Ugh.

This does not currently address the problem of the master's certificate still having the old kubernetes service IP address in it. There doesn't seem to be a completely generic way to regenerate the cert correctly. I'm thinking that we may have to make the procedure only work for people who used the ansible install?

It also doesn't address the iptables problem. Again, if we only support ansible installs then this is easy (just restart iptables.service), but if we want to support arbitrary configurations, then we probably have to require them to reboot the nodes...

danwinship:
> @danwinship Are you using AsciiBinder to test this? What does the output look like whn you omit the space?

    --- /tmp/good.html      2016-11-10 09:57:14.860855443 -0500
    +++ /tmp/bad.html       2016-11-10 09:55:51.306776031 -0500
    @@ -1168,10 +1168,10 @@
     </tr>
     </table>
     </div>
    -<div class="olist arabic">
    -<ol class="arabic">
    -<li>
    -<p>On every node and master, do:</p>
    +<div class="paragraph">
    +<div class="title">On every node and master, do:</div>
    +<p>+</p>
    +</div>
     <div class="listingblock">
     <div class="content">
     <pre># systemctl stop atomic-openshift-node
    ...

Good:
good

Bad:
bad

nhr:
@danwinship I stand corrected. Thanks for providing comparative output.

knobunc (approved):
LGTM thanks @danwinship and @weliang1

knobunc:

danwinship:
This is not yet complete and there was a question in my comment above about how to proceed.

Also, Marc said he is OK with us doing this as a support article rather than as formal documentation (although the first part, about expanding clusterNetworkCIDR, should probably stay in the docs, since it is actually fully supported).

danwinship:
Pushed an updated branch that only documents the easy "make clusterNetworkCIDR larger" part, which we would like to have in the docs for 1.4. The rest will be turned into a support article

install_config/configuring_sdn.adoc
```diff
@@ -164,6 +164,78 @@ Check VNIDs by running:
...
+{product-name} may support such changes in the future, for now the
+best solution is to reinstall the cluster with the new correct values.
+endif::[]
+ifdef::openshift-enterprise[]
+Although it is possible to make such changes in a deployed cluster, it
+is complicated and not fully supported; if it all possible, you should
+consider reinstalling the cluster instead. FIXME talk to your support
+person or look up the answer in the support db or something, I'm not
+sure what to write here...
+endif::[]
```
danwinship: note particularly this part... need help from docs people to figure out what the right thing to say here is

ahardin-rh: @danwinship Is there a specific resource we can direct them to?

danwinship: No, not yet... we haven't finished writing it yet. But I guess we can get a placeholder article created now to reserve a URL.

ahardin-rh:
@danwinship any updates on this PR? Looks like there was still work to be done. Thanks!

danwinship:
yeah, it got dropped in favor of other stuff that had to be completed before code freeze. I still need to get back to this...

DanyC97:
@danwinship my 0.02$ cents here ...

you comment from Nov 2016 to turn it into a support article .. most of the time users first turn to docs to find info, if you an article is preferred how end users (who don't have a support contract etc) will be able to get this info ?

danwinship:
@DanyC97 The stuff that will be in the support article is stuff that is officially unsupported, requires hacking things behind OpenShift's back, and carries a high risk of breaking your cluster. Over time, we intend to make it possible to do those things in a supported way, and then we'll add docs on that to openshift-docs.

ahardin-rh:
I am closing this due to inactivity. Please let me know if we need to reopen it.

PR #3150, opened 2016-11-02
update seccomp runtime/default to docker/default

pweil-:
Docs were using the wrong constant for the runtime default profile.  Switched to `docker/default`.

@sttts please review?

### DIFF
--- a/admin_guide/seccomp.adoc
+++ b/admin_guide/seccomp.adoc
@@ -66,7 +66,7 @@ the security constraints of an individual system.
 To create your own custom profile, create a file on every node in the
 `seccomp-profile-root` directory.
 +
-If you are using the default *runtime/default* profile, you do not need to
+If you are using the default *docker/default* profile, you do not need to
 create one.
 
 . Configure your nodes to use the *seccomp-profile-root* where your profiles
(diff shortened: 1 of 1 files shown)


pweil-:
@adellape FYI this is a bug in the seccomp docs that were recently merged (in case it was back ported anywhere)

sttts:
Didn't we want to change `docker/default` into `runtime/default` in the code? AppArmor uses the later.

pweil-:
I think we do...however the upstream code is still checking `runtime/default` both in the kubelet and validation so this should be corrected until that update is made.

vikram-redhat:
@ahardin-rh - assigned to you for this bug: https://bugzilla.redhat.com/show_bug.cgi?id=1390976

ahardin-rh:
Looks like https://github.com/kubernetes/community/pull/390 is now merged. Is this good to go?

@pweil- @sttts I have this labeled for 3.4. Does this also impact earlier versions of the docs? Thanks!

pweil-:
This should be included in any 1.3 seccomp docs as well (if they existed)

ahardin-rh:
@pweil- Thank you! Labels are now updated.

ahardin-rh:
[rev_history]
|xref:../admin_guide/seccomp.adoc#admin-guide-seccomp[Restricting Application Capabilities Using Seccomp]
|Updated Seccomp references of `runtime/default` to be `docker/default`.
%

PR #3199, opened 2016-11-11
Cherrypick of "Removed rogue ifdef" to OCP 3.3

ahardin-rh:


ahardin-rh:
cc @adellape @aheslin

PR #3200, opened 2016-11-11
Cherry-pick of "Removed rogue ifdef" to Dedicated

ahardin-rh:


ahardin-rh:
cc @aheslin @adellape

PR #3201, opened 2016-11-11
Cherry-pick "Removed rogue ifdef" to Online

ahardin-rh:


ahardin-rh:
cc @adellape @aheslin

PR #3205, opened 2016-11-11
[WIP] Kubernetes deployments

0xmichalis:
@mfojtik still WIP but mostly done I think

https://trello.com/c/oAlPryPh/793-3-need-to-document-how-to-use-upstream-deployments

### DIFF
--- a/dev_guide/kube_deployments.adoc
+++ b/dev_guide/kube_deployments.adoc
@@ -1,0 +1,99 @@
+[[dev-guide-deployments]]
+= Kubernetes Deployments
+{product-author}
+{product-version}
+:data-uri:
+:icons:
+:experimental:
+:toc: macro
+:toc-title:
+
+toc::[]
+
+== Overview
+
+Deployments have been added in Kubernetes and serve as a descendant of {product-title} 
+DeploymentConfigs. Since DeploymentConfigs have existed since prior to the 1.0 release in
+{product-title} and Deployments were added in Kubernetes 1.2, it is natural that the latter
+will slightly diverge from the former. This document serves to call out the differences
+between the two resources. Our long-term goal is to provide full feature parity in upstream
+Deployments and switch to use them as a single resource that will provide fine-grained
+management over applications. If you want to look at documentation specifically related to
+upstream Deployments, see xref:http://kubernetes.io/docs/user-guide/deployments/[here].
+
+[NOTE]
+====
+As of writing this, upstream Deployments are still a beta feature and {product-title} provides
+them as a tech preview whereas DeploymentConfigs have been stable since 1.0.
+====
+
+[[Features-not-in-upstream]]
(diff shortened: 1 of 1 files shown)


adellape:
@kargakis Can you confirm my observations on these changes? Just to make sure I get what's being proposed so far.

- It looks like the main content for this PR is the new `deployments.adoc`, which mostly serves to describe the differences between kube deployments and openshift deploymentconfigs.
- The renamed `deployments.adoc` -> `deployment_configs.adoc` is largely unchanged.
- For actual info / usage on kube deployments, we're linking to the upstream kube docs (http://kubernetes.io/docs/user-guide/deployments/).

I have a WIP PR (https://github.com/openshift/openshift-docs/pull/3179) that breaks up the old, enormous `deployments.adoc` file into a subdirectory of multiple topics:

http://file.rdu.redhat.com/~adellape/110816/deployments_breakup/dev_guide/deployments/how_deployments_work.html

I can take your content/commits here and work it into the new subdirectory structure of my https://github.com/openshift/openshift-docs/pull/3179, but I want to make sure I understand to what lengths we want to document kube deployments in OpenShift docs, during this tech preview period (in particular, thinking about OCP 3.4). cc @mfojtik

kargakis:
@adellape cool work in #3179! You've got it right about this PR. Let me finish the second commit and then you can take it over. I will remove the file change and include the change in a separate new file.

mfojtik:
@kargakis I think we should also highlight when to choose Deployments over DeploymentConfigs (I guess we will be getting that questions sooner or latest). Initially I would say you can use upstream Deployments when you don't plan to have any hooks or run any custom strategies and you don't want to use ImageStream. We should mention that Deployments are supported sorely to make sure that the upstream examples/projects that use the Deployments over DeploymentConfigs are assured to work smoothly when running on OpenShift.

kargakis:
> @kargakis I think we should also highlight when to choose Deployments over DeploymentConfigs (I guess we will be getting that questions sooner or latest). Initially I would say you can use upstream Deployments when you don't plan to have any hooks or run any custom strategies and you don't want to use ImageStream. We should mention that Deployments are supported sorely to make sure that the upstream examples/projects that use the Deployments over DeploymentConfigs are assured to work smoothly when running on OpenShift.

I would expect the reasons to choose D over DC would stem from the differences highlighted in this doc. We can make it explicit but it will essentially be a reiteration of the differences. Also I don't think we added Deployments just for upstream examples to work in OpenShift but also because Deployments are evolved out of DeploymentConfigs and since they are embraced in Kubernetes and we already have experience with DCs, we can drive their development (Deployments) and merge the two objects into one so the Kubernetes community can take advantage of what we have been using so long.

kargakis:
Superseded by https://github.com/openshift/openshift-docs/pull/3255

PR #3227, opened 2016-11-15
Fix typo in devGuide imagestream docker tag description

tiwillia:
Before:
`Container images can have names added to them that make it more intuitive what they contain, called a tag. `

After:
`Container images can have names added to them that make it more intuitive to determine what they contain, called a tag.`

Please suggest better wording if preferred.


ahardin-rh:
LGTM. Thanks! Good eye. :eyes:

PR #3231, opened 2016-11-16
Bug 1393017 - Fixed red font with red background issue

gaurav-nelson:
BUG 1393017 - Fixed red font with red background issue and updtaed .travis.yaml to latest

@vikram-redhat Please check.


vikram-redhat:
No Rev History required

PR #3233, opened 2016-11-16
Remove NFS Attach Example for Docker Registry for Production Use

pospispa:
The deleted command:
$ oc volume deploymentconfigs/docker-registry \
     --add --overwrite --name=registry-storage --mount-path=/registry \
     --source='{"nfs": { "server": "<fqdn>", "path": "/path/to/export"}}'
does not work because the default docker-registry SCC profile, i.e. restricted, does not allow direct mounting of NFS volumes.

In addition, docker-registry based on NFS does not scale because NFS can not "guarantee" read-after-write consistency, when scaled.

To sum it up, I propose to delete the above command from the documentation from the place where it is described as one of the options for production use.

For additional information refer to bug: https://bugzilla.redhat.com/show_bug.cgi?id=1393568

### DIFF
--- a/install_config/registry/deploy_registry_existing_clusters.adoc
+++ b/install_config/registry/deploy_registry_existing_clusters.adoc
@@ -152,14 +152,6 @@ $ oc volume deploymentconfigs/docker-registry --add --name=registry-storage -t p
      --claim-name=<pvc_name> --overwrite
 ----
 
-Or, to attach an existing NFS volume to the registry:
-
-----
-$ oc volume deploymentconfigs/docker-registry \
-     --add --overwrite --name=registry-storage --mount-path=/registry \
-     --source='{"nfs": { "server": "<fqdn>", "path": "/path/to/export"}}'
-----
-
 [NOTE]
 ====
 See xref:registry_known_issues.adoc#install-config-registry-known-issues[Known Issues] if using a scaled registry with a


pospispa:
PTAL @mfojtik @openshift/team-documentation 

Target versions: 3.3 and 3.4

vikram-redhat:
@ahardin-rh - PTAL/

mfojtik:
@pospispa is the NFS still usable for non-production environments?

pospispa:
@mfojtik sure, NFS is usable for non-production environments.

mfojtik:
@pospispa can't we instead of removing the attach add notice that this is not suitable for production environments? (maybe outline the reasons why).

pospispa:
There is already a note that links to the [Known issue](https://docs.openshift.org/latest/install_config/registry/registry_known_issues.html#install-config-registry-known-issues) that describes push errors while using shared NFS volume.

The reason why I propose to remove the command from documentation is because the command does not work as is and because it is in "Production Use" section.

mfojtik:
@pospispa I see now, thanks!

LGTM

ahardin-rh:
LGTM. Thank you!

ahardin-rh:
[rev_history]
|xref:../install_config/registry/deploy_registry_existing_clusters.adoc#install-config-deploy-registry-existing-clusters[Setting up the Registry -> Deploying a Registry on Existing Clusters]
|Removed inaccurate example showing how to attach an existing NFS volume to the registry.
%

PR #3236, opened 2016-11-16
Add admin guide for configuring cluster-wide IPsec encryption

dcbw:
This document is relevant for OpenShift 3.3 and later.

@openshift/networking

### DIFF
--- a/admin_guide/ipsec.adoc
+++ b/admin_guide/ipsec.adoc
@@ -1,0 +1,173 @@
+[[admin-guide-ipsec]]
+= ipsec
+{product-author}
+{product-version}
+:data-uri:
+:icons:
+:experimental:
+:toc: macro
+:toc-title:
+
+toc::[]
+
+== Overview
+IPsec is a mechanism for encrypting communication between hosts which can
+already communicate with each other via the Internet Protocol (IP). The simplest
+way of protecting traffic within an OpenShift cluster is to ensure that the
+master and all nodes use IPsec to encrypt their communication.  This document
+describes how to achieve this.
+
+In this example, we will secure communication of an entire IP subnet from which
+the OpenShift hosts receive their IP addresses. Since you are securing the
+host-to-host communication, this automatically includes all cluster management
+and pod data traffic. Note that as OpenShift management traffic uses HTTPS,
+enabling IPsec will encrypt management traffic a second time.
+
+[[requirements]]
+== Requirements
+This guide requires that you have the `libreswan` package, version 3.19 or later,
+installed on cluster hosts.  Only 3.19 and later include the necessary
+opportunistic group functionality that allows hosts to be configured without
+knowledge of every other host in the cluster.
+
+[[procedure]]
+== Procedure
+This procedure should be repeated for each host (both nodes and masters) in your
+cluster.  Any host in the cluster that does not have IPsec enabled will not be
+able to communicate with a host that does. It is suggested to perform these steps
+on masters first, and then each node.
+
+[[certificates]]
+=== Certificates
+By default, OpenShift secures cluster management communication with mutually
+authenticated HTTPS communication. This means that both the client (like an
+openshift node) and the server (like the openshift api-server) send each other
+their certificates, which are checked against a known Certificate Authority (CA).
+These certificates are generated at cluster setup time and typically live on
+each host.
+
+These certificates can also be used to secure pod communications with IPsec. You
+need three files on each host:
+
+- cluster Certificate Authority file
+- host client certificate file
+- host private key file
+
+First, determine what the certificate's nickname will be after it has been
+imported into the `libreswan` certificate database. The nickname is taken
+directly from the certificate's subject's Common Name (CN):
+
+----
+openssl x509 -in /path/to/client-certificate -subject -noout | sed -n 's/.*CN=\(.*\)/\1/p'
+----
+
+Save the nickname for later.
+
+The client certificate, CA certificate, and private key files must be combined
+into a PKCS#12 file, which is a common file format for multiple
+certificates and keys.  To do this, you can use the `openssl` program:
+
+----
+  openssl pkcs12 -export \
+    -in /path/to/client-certificate \
+    -inkey /path/to/private-key \
+    -certfile /path/to/certificate-authority \
+    -passout pass: \
+    -out certs.p12
+----
+
+This PKCS#12 file must then be imported into the `libreswan` certificate
+database. The -W option is left empty because we do not need to assign a
+password to the PKCS#12 file as it is only temporary.
+
+----
+  ipsec initnss
+  pk12util -i certs.p12 -d sql:/etc/ipsec.d -W ""
+  rm certs.p12
+----
+
+[[ipsec-policy]]
+=== libreswan IPsec Policy
+
+Now that the necessary certificates have been imported into the `libreswan`
+certificate database, you can create a policy that uses them to secure
+communication between hosts in your cluster. The following configuration
+creates two `libreswan` connections.  The first encrypts traffic using the
+OpenShift certificates, while the second creates exceptions to the encryption
+for cluster-external traffic.
+
+Place the following text into the file `/etc/ipsec.d/openshift-cluster.conf`:
+
+----
+conn private
+	left=%defaultroute
+	leftid=%fromcert
+	# our certificate
+	leftcert="NSS Certificate DB:openshift-node-2" <1>
+	right=%opportunisticgroup
+	rightid=%fromcert
+	# their certificate transmitted via IKE
+	rightca=%same
+	ikev2=insist
+	authby=rsasig
+	failureshunt=drop
+	negotiationshunt=hold
+	auto=ondemand
+
+conn clear
+	left=%defaultroute
+	right=%group
+	authby=never
+	type=passthrough
+	auto=route
+	priority=100
+----
+<1> replace the text after the colon (eg `openshift-node-2`) with the certificate
+nickname you saved above. For example, on a different host, the full line might be
+`leftcert="NSS Certificate DB:openshift-master"`.
+
+Now that the configuration has been defined, you need to tell `libreswan`
+what IP subnets and hosts to apply each policy to. This is done through policy
+files in `/etc/ipsec.d/policies/` where each configured connection has a
+corresponding policy file. In our example above, we have two connections,
+`private` and `clear` and each will have a file in `/etc/ipsec.d/policies/`.
+
+`/etc/ipsec.d/policies/private` should contain the IP subnet of your cluster,
+which your hosts receive IP addresses from. This, by default, will cause all
+communication between hosts in the cluster subnet to be encrypted if
+the remote host's client certificate authenticates against the local host's
+Certificate Authority certificate. If the remote host's certificate does not
+authenticate, all traffic between the two hosts will be blocked.
+
+For example, if all your hosts are configured to use addresses in the
+172.16.0.0/16 address space, your `private` policy file would contain:
+
+----
+172.16.0.0/16 <1>
+----
+<1> any number of additional subnets to encrypt may be added to this file, which
+will result in all traffic to those subnets using IPsec as well.
+
+Next, encryption between all hosts and the subnet gateway must be un-encrypted,
+to ensure that traffic can enter and exit the cluster.  To do this, add the
+gateway to the `/etc/ipsec.d/policies/clear` file like so:
+
+----
+172.16.0.1/32 <1>
+----
+<1> additional hosts and subnets may be added to this file, which will result in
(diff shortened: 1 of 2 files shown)


admin_guide/ipsec.adoc
```diff
@@ -0,0 +1,174 @@
...
+:toc-title:
+
+toc::[]
+
+== Overview
+IPsec is a mechanism for encrypting communication between hosts which can
+already communicate with each other via the Internet Protocol (IP). The simplest
+way of protecting traffic within an Openshift cluster is to ensure that the
+master and all nodes use IPsec to encrypt their communication.  This document
+describes the simplest way to do this.
```
knobunc: Can we reword the last sentence to avoid repetition of 'simplest'.

Perhaps:
This document describes how to achieve that.

admin_guide/ipsec.adoc
```diff
@@ -0,0 +1,174 @@
...
+	right=%group
+	authby=never
+	type=passthrough
+	auto=route
+	priority=100
+----
+
+Now that the configuration has been defined, you need to tell `libreswan`
+what IP subnets and hosts to apply each policy to. This is done through policy
+files in /etc/ipsec.d/policies/ where each configured connection has a
```
knobunc: Do the paths need to be made into fixed width text?

admin_guide/ipsec.adoc
```diff
@@ -0,0 +1,174 @@
...
+corresponding policy file. In our example above, we have two connections,
+`private` and `clear` and each will have a file in /etc/ipsec.d/policies.
```
knobunc: Above policies has a trailing /, here it doesn't.  We should be consistent.

knobunc (approved):
Really nice stuff, thanks.  LGTM

Just a couple of nits.

admin_guide/ipsec.adoc
```diff
@@ -0,0 +1,174 @@
...
+directly from the certificate's subject's Common Name (CN):
+
+----
+openssl x509 -in /path/to/client-certificate -subject -noout | sed -n 's/.*CN=\(.*\)/\1/p'
+----
+
+Save the nickname for later.
+
+The client certificate, CA certificate, and private key files must be combined
+into a PKCS#12 file, which is a common container format for multiple
```
pecameron: Container is a loaded term w.r.t.openshift/docker. Maybe use something else?

pecameron (commented):
LGTM
Nicely written.
Should ansible be updated to install libreswan?
Do we expect a performance impact for using this?

admin_guide/ipsec.adoc
```diff
@@ -0,0 +1,174 @@
...
+:experimental:
+:toc: macro
+:toc-title:
+
+toc::[]
+
+== Overview
+IPsec is a mechanism for encrypting communication between hosts which can
+already communicate with each other via the Internet Protocol (IP). The simplest
+way of protecting traffic within an Openshift cluster is to ensure that the
```
danwinship: "OpenShift" (everywhere)

admin_guide/ipsec.adoc
```diff
@@ -0,0 +1,174 @@
...
+master and all nodes use IPsec to encrypt their communication.  This document
+describes the simplest way to do this.
+
+In this example, we will secure communication of an entire IP subnet from which
+the Openshift hosts receive their IP addresses. Since you are securing the
+host-to-host communication, this automatically includes all cluster management
+and pod data traffic.
```
danwinship: Management traffic is already over https, so this adds a second layer of encryption...

admin_guide/ipsec.adoc
```diff
@@ -0,0 +1,174 @@
...
+[[requirements]]
+== Requirements
+This guide requires that you have the `libreswan` package, version 3.19 or later,
+installed on cluster hosts.  Only 3.19 and later include the necessary
+opportunistic group functionality that allows hosts to be configured without
+knowledge of every other host in the cluster.
+
+[[procedure]]
+== Procedure
+This procedure should be repeated for each host in your cluster.  Any host
```
danwinship: maybe explicitly clarify that "host" means both nodes and masters here

admin_guide/ipsec.adoc
```diff
@@ -0,0 +1,174 @@
...
+in the cluster that does not have IPsec enabled will not be able to communicate
+with a host that does. It is suggested to perform these steps on the master
```
danwinship: "the master(s)"

admin_guide/ipsec.adoc
```diff
@@ -0,0 +1,174 @@
...
+first, and then each node.
+
+[[certificates]]
+=== Certificates
+By default, Openshift secures cluster management communication with mutually
+authenticated HTTPS communication. This means that both the client (like an
+openshift node) and the server (like the openshift api-server) send each other
+their certificates, which are checked against a know Certificate Authority (CA).
```
danwinship: "known"

admin_guide/ipsec.adoc
```diff
@@ -0,0 +1,174 @@
...
+These certificates are generated at cluster setup time and typically live on
+each host.
+
+These certificates can also be used to secure pod communications with IPsec. You
+need at three files on each host:
```
danwinship: either an extra or a missing word there (depending on whether you meant "at least three" or "three")

admin_guide/ipsec.adoc
```diff
@@ -0,0 +1,174 @@
...
+    -in /path/to/client-certificate \
+    -inkey /path/to/private-key \
+    -certfile /path/to/certificate-authority \
+    -passout pass: \
+    -out certs.p12
+----
+
+This PKCS#12 file must then be imported into the `libreswan` certificate
+database. The -W option is left empty because we do not need to assign a
+password to the PKCS#12 file as it is only temporary.
```
danwinship: so it should be deleted afterward?

admin_guide/ipsec.adoc
```diff
@@ -0,0 +1,174 @@
...
+Now that the necessary certificates have been imported into the `libreswan`
+certificate database, you can create a policy that uses them to secure
+communication between hosts in your cluster. The following configuration
+creates two `libreswan` connections.  The first encrypts traffic using the
+Openshift certificates, while the second creates exceptions to the encryption
+for cluster-external traffic.
+
+Place the following text into the file /etc/ipsec.d/openshift-cluster.conf,
+being sure to replace the `<certificate-nickname-here>` text with the 
+certificate nickname you saved above:
```
danwinship: I believe standard openshift-docs style is to use a literal in the example, with a footnote indicating where it comes from/what it means. Eg, see the "Configuring the Pod Network on Masters" section in install_config/configuring_sdn.adoc.

admin_guide/ipsec.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+Now that the configuration has been defined, you need to tell `libreswan`
+what IP subnets and hosts to apply each policy to. This is done through policy
+files in /etc/ipsec.d/policies/ where each configured connection has a
+corresponding policy file. In our example above, we have two connections,
+`private` and `clear` and each will have a file in /etc/ipsec.d/policies.
+
+/etc/ipsec.d/policies/private should contain the IP subnet of your cluster,
+which your hosts receive IP addresses from. This, by default, will cause all
+communication between hosts in the 172.16.0.0/16 subnet to be encrypted if
```
danwinship: Makes it sound like regardless of what you put in the file, it will affect traffic to 172.16.0.0/16... I think you rearranged some text here and didn't fully fix up everything for it.

admin_guide/ipsec.adoc
```diff
@@ -0,0 +1,174 @@
...
+the remote host's client certificate authenticates against the local host's
+Certificate Authority certificate. If the remote host's certificate does not
+authenticate, all traffic between the two hosts it will be blocked.
```
danwinship: drop "it"

admin_guide/ipsec.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+For example, if all your hosts are configured to use addresses in the
+172.16.0.0/16 address space, your `private` policy file would contain:
+
+----
+172.16.0.0/16
+----
+
+Next encryption between all hosts and the subnet gateway must be un-encrypted,
```
danwinship: comma after "Next"

admin_guide/ipsec.adoc
```diff
@@ -0,0 +1,174 @@
...
+the remote host's client certificate authenticates against the local host's
+Certificate Authority certificate. If the remote host's certificate does not
+authenticate, all traffic between the two hosts it will be blocked.
+
+For example, if all your hosts are configured to use addresses in the
+172.16.0.0/16 address space, your `private` policy file would contain:
+
+----
+172.16.0.0/16
+----
```
danwinship: Presumably one can add additional IP ranges or specific addresses here? It would be good to explain that.

admin_guide/ipsec.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+== Troubleshooting
+When authentication cannot be completed between two hosts, you will not even
+be able to ping between them as all IP traffic will be rejected. If the `clear`
+policy is not configured correctly, you will also not be able to SSH to the host
+from another host in the cluster.
+
+The most common failure is certificates that do not verify against a common
+Certificate Authority certificate. To verify this:
```
danwinship: It seems unlikely that this would happen in an OpenShift cluster, since the certificates are being created by automated processes; if people were ending up with node certs that used the wrong CA, then that would cause https failures even if they weren't using ipsec.

dcbw:
@openshift/networking PTAL, thanks!

knobunc (approved):
LGTM.  Please just add the versions you want this to be applied to to the description.

dcbw:
@knobunc done

admin_guide/ipsec.adoc
```diff
@@ -0,0 +1,173 @@
...
+already communicate with each other via the Internet Protocol (IP). The simplest
+way of protecting traffic within an OpenShift cluster is to ensure that the
+master and all nodes use IPsec to encrypt their communication.  This document
+describes how to achieve this.
+
+In this example, we will secure communication of an entire IP subnet from which
+the OpenShift hosts receive their IP addresses. Since you are securing the
+host-to-host communication, this automatically includes all cluster management
+and pod data traffic. Note that as OpenShift management traffic uses HTTPS,
+enabling IPsec will encrypt management traffic a second time.
```
danwinship: I was pointing that out before as more of a question: do we really want to do that? (Do our customers really want to do that?) Would it be better to just specifically encrypt the currently-unencrypted traffic (which is just VXLAN?) rather than encrypting all traffic between the hosts? (Does libreswan even let us configure something like that?)

(If we do actually want to do this, I don't think it needs to be pointed out in the docs; you can go back to whatever this said before.)

letoams:
On Wed, 30 Nov 2016, Dan Winship wrote:

> +In this example, we will secure communication of an entire IP subnet from which
> +the OpenShift hosts receive their IP addresses. Since you are securing the
> +host-to-host communication, this automatically includes all cluster management
> +and pod data traffic. Note that as OpenShift management traffic uses HTTPS,
> +enabling IPsec will encrypt management traffic a second time.
> 
> I was pointing that out before as more of a question: do we really want to do that? (Do our customers really want to do that?) Would it be better to just specifically encrypt the currently-unencrypted traffic
> (which is just VXLAN?) rather than encrypting all traffic between the hosts? (Does libreswan even let us configure something like that?)
> 
> (If we do actually want to do this, I don't think it needs to be pointed out in the docs; you can go back to whatever this said before.)

We are working on supporting port exclusion in an easier way that
currently. So I would advise to ignore the double encryption for now.

If you really want to exclud it, you can add excluding port 443 using something
like:

conn passthrough-outgoing-https
        left=%defaultroute
        leftprotoport=tcp/%any
        right=0.0.0.0
        rightsubnet=0.0.0.0/0
        rightprotoport=tcp/443
        auto=route
        type=passthrough
        priority=1

conn passthrough-incoming-https
        left=%defaultroute
        leftprotoport=tcp/443
        right=0.0.0.0
        rightsubnet=0.0.0.0/0
        rightprotoport=tcp/%any
        priority=1
        auto=route
        type=passthrough


This will result in high priority kernel policy that will match packets
to/from port 443 and will let that traffic go out in the clear. It has
a higher priority that the regular opportunistic connections.

Paul

knobunc:
@openshift/team-documentation this has passed tech review.  Over to you.  Thanks.

vikram-redhat:
@knobunc - Hey Ben. Is this for 3.4 or 3.3 onwards?

@bfallonf - PTAL.

bfallonf:
I created #3398 as a follow-up to this (for 3.3 ->). 

I'll close this and continue the edits there, and put that onto peer review.

Let me know if there's any issues.

PR #3241, opened 2016-11-17
Bug 1390663 - Updated persistence storage sections

gaurav-nelson:
Updated persistence storage sections to include that they do not support 'Recycle' recycling policy.


gaurav-nelson:
@vikram-redhat  Please check

vikram-redhat:
@bfallonf - PTAL.

vikram-redhat:
Should the text be "does not" rather than "do not"?

PR #3108, opened 2016-10-25
take public url for cockpit documentation

chrira:
Replaced access**.qa**.redhat.com with access.redhat.com to make link work.
New link to cockpit documentation:
https://access.redhat.com/documentation/en/red-hat-enterprise-linux-atomic-host/7/single/getting-started-with-cockpit/

PR #3257, opened 2016-11-18
Follow-up to PR#3108

ahardin-rh:
Follow-up to https://github.com/openshift/openshift-docs/pull/3108


ahardin-rh:
No revision history needed

PR #3269, opened 2016-11-22
Update audit configuration description

soltysh:
With k8s 1.4 we switched to using upstream audit mechanics, which gave us a few additional options. I've updated the description with those new options.

@ahardin-rh ptal

### DIFF
--- a/install_config/master_node_configuration.adoc
+++ b/install_config/master_node_configuration.adoc
@@ -689,7 +689,8 @@ Each audit log contains two entries:
 .. The source IP of the request
 .. The HTTP method being invoked
 .. The original user invoking the operation
-.. The impersonated user for the operation
+.. The impersonated user for the operation (`self` meaning himself here)
+.. The impersonated group for the operation (`lookup` meaning user's group)
 .. The namespace of the request or <none>
 .. The URI as requested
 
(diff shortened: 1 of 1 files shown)


install_config/master_node_configuration.adoc
```diff
@@ -689,7 +689,8 @@ Each audit log contains two entries:
 .. The source IP of the request
 .. The HTTP method being invoked
 .. The original user invoking the operation
-.. The impersonated user for the operation
+.. The impersonated user for the operation (`self` meaning himself here)
```
ahardin-rh: @soltysh Is it accurate to say that self means the user's self?

ahardin-rh:
@soltysh Thank you! I can do some minor style edits in a follow-up PR.

ahardin-rh:
[rev_history]
|xref:../install_config/master_node_configuration.adoc#install-config-master-node-configuration[Master and Node Configuration]
|Updated the xref:../install_config/master_node_configuration.adoc#master-node-config-audit-config[Audit Configuration] section description and added Audit Configuration Parameters.
%

install_config/master_node_configuration.adoc
soltysh: It's basically himself. When you're not using `--as` flag you're acting as yourself, this is when `self` will show up here, which is most of the time. When you do specify `--as` flag for any command this will be the username which you specified as the flag value. I hope that makes sense :)

PR #3277, opened 2016-11-23
Bug 1384994 edited information on embedded etcd

bfallonf:
As per: https://bugzilla.redhat.com/show_bug.cgi?id=1384994

Changed some of the recommended information around installs. Specifically, embedded etcd and the examples.

Will ask for ack over email.

### DIFF
--- a/install_config/downgrade.adoc
+++ b/install_config/downgrade.adoc
@@ -58,9 +58,8 @@ which have the node component on them) with a timestamp:
 ----
 ====
 
-If you are using an external etcd cluster (versus the single embedded etcd), the
-backup is likely created on all etcd members, though only one is required for
-the recovery process.
+When using an external etcd cluster, the backup is created on all etcd
+members, though only one is required for the recovery process.
 
 The RPM downgrade process in a later step should create *_.rpmsave_* backups of
 the following files, but it may be a good idea to keep a separate copy
@@ -141,9 +140,8 @@ If you are using external etcd, also remove the *etcd* package:
 ----
 ====
 
-For embedded etcd, you can leave the *etcd* package installed, as the package is
-only required so that the `etcdctl` command is available to issue operations in
-later steps.
+If using the embedded etcd, leave the *etcd* package installed. It is required
+for running the `etcdctl` command to issue operations in later steps.
 
 [[downgrade-docker]]
 == Downgrading Docker
@@ -251,23 +249,32 @@ member:
 [[downgrading-restoring-etcd]]
 == Restoring etcd
 
-Whether using embedded or external etcd, you must first restore the etcd backup
+When using an external etcd host, you must first restore the etcd backup
 by creating a new, single node etcd cluster. If using external etcd with
 multiple members, you must then also add any additional etcd members to the
 cluster one by one.
 
+////
 However, the details of the restoration process differ between
 xref:downgrading-restoring-embedded-etcd[embedded] and
 xref:downgrading-restoring-external-etcd[external] etcd. See the following
-section that matches your etcd configuration and follow the relevant steps
+section and follow the relevant steps
 before continuing to
 xref:downgrade-bringing-openshift-services-back-online[Bringing OpenShift
 Services Back Online].
+////
 
 Follow the xref:../admin_guide/backup_restore.adoc#cluster-restore[Cluster
 Restore] procedure to restore single-member etcd clusters.
 
 [[downgrading-restoring-embedded-etcd]]
+[NOTE]
+====
+Using the {product-title} embedded etcd in production environments is not
+supported.
+====
+
+////
 === Embedded etcd
 
 Restore your etcd backup and configuration:
@@ -320,6 +327,7 @@ command, which should look similar to the following near the end:
 
 . Continue to xref:downgrade-bringing-openshift-services-back-online[Bringing
 {product-title} Services Back Online].
+////
 
 [[downgrading-restoring-external-etcd]]
 === External etcd
--- a/install_config/install/advanced_install.adoc
+++ b/install_config/install/advanced_install.adoc
@@ -683,7 +683,7 @@ volume provisioning for your cloud platform:
 == Single Master Examples
 
 You can configure an environment with a single master and multiple nodes, and
-either a single embedded *etcd* or multiple external *etcd* hosts.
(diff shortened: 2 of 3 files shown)


install_config/install/planning.adoc
```diff
@@ -117,7 +117,7 @@ not supported.
 === Single Master and Multiple Nodes
 
 The following table describes an example environment for a single
-xref:../../architecture/infrastructure_components/kubernetes_infrastructure.adoc#master[master] (with embedded *etcd*)
+xref:../../architecture/infrastructure_components/kubernetes_infrastructure.adoc#master[master] (with *etcd* on the same host)
```
liggitt: Embedded is not the same as merely being on the same host.

liggitt: Note that it doesn't mention installing etcd. That's because it is using the etcd embedded in the openshift binary

install_config/install/advanced_install.adoc
```diff
@@ -1029,7 +1036,7 @@ using the `native` HA method:
 |Host Name |Infrastructure Component to Install
 
 |*master1.example.com*
-.3+.^|Master (clustered using native HA) and node with etcd on each host
+.3+.^|Master (clustered using native HA) and node
```
liggitt: I wouldn't drop this. It's important to clarify that etcd is installed

install_config/install/advanced_install.adoc
```diff
@@ -683,7 +683,7 @@ volume provisioning for your cloud platform:
 == Single Master Examples
 
 You can configure an environment with a single master and multiple nodes, and
-either a single embedded *etcd* or multiple external *etcd* hosts.
+either a single or multiple number of external *etcd* hosts.
```
liggitt: Are we removing all mentions of the embedded etcd?

install_config/install/planning.adoc
bfallonf: > Embedded is not the same as merely being on the same host.
  
Right, but I thought that the change is more toward what we should be recommending.

> Note that it doesn't mention installing etcd. That's because it is using the etcd embedded in the openshift binary

Hmm OK. Maybe this is leading to something missing from the docs. i'll talk about this more in a comment below, but it's my understanding that we should not be recommending embedded etcd.

install_config/install/advanced_install.adoc
bfallonf: Ack. I'll clear up the other points first.

bfallonf:
@liggitt Thanks a bunch for the comments above. This came from this BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1384994

And a conversation with @detiber . From the email:

"We should also get away from any configuration involving embedded etcd, so even our single master examples should have an etcd host defined.

The correct scenarios we should mention are:
- single master/etcd host (listed as both a master and etcd host)
- multi-master w/ co-located etcd (master and etcd hosts are duplicated between the master and etcd groups)
- multi-master w/ external etcd (master and etcd hosts are distinct)"

I thought I was making the recommended changes, but I may have misunderstood. Jason had yet to give me an ack though.

Jason, do you have any thoughts?

detiber:
@bfallonf I do like the idea of being explicit about etcd where possible, but we do definitely want to drop references to running with the embedded etcd, since we don't consider it a "supported" option at this time.

liggitt:
That's fine, just need to clarify what needs to be installed for each scenario. I also wasn't sure what product the doc was targeting… the origin docs still reference all-in-one and embedded in places

bfallonf:
@liggitt @detiber I've updated this per Jordan's comments.

Also, am I to be removing all other cases in the rest of the docs of the embedded etcd? Is there a difference between what should be in the Origin vs OCP docs for this?

detiber:
@bfallonf The only case where Origin should reference embedded etcd is in the developer related docs in the origin repo.

bfallonf:
OK. I fixed up the docs a little more and took out _most_ cases where the embedded etcd is mentioned. The few that are still there are sorta about the difference between embedded and external. But I my reasonging is that people are going to be using it regardless of if it's supported, so the couple references are relevant. I think?

@detiber @liggitt WDYT?

bfallonf:
@detiber @liggitt bump!

In the meantime, if anyone wants to peer review ( @gaurav-nelson @ahardin-rh @adellape ) feel free to go ahead.

install_config/downgrade.adoc
```diff
@@ -58,9 +58,8 @@ which have the node component on them) with a timestamp:
 ----
 ====
 
-If you are using an external etcd cluster (versus the single embedded etcd), the
-backup is likely created on all etcd members, though only one is required for
-the recovery process.
+When using an external etcd cluster, the backup is likely created on all etcd
```
gaurav-nelson: I guess we can remove 'likely'. _the backup is created_

install_config/downgrade.adoc
```diff
@@ -141,9 +140,9 @@ If you are using external etcd, also remove the *etcd* package:
 ----
 ====
 
-For embedded etcd, you can leave the *etcd* package installed, as the package is
-only required so that the `etcdctl` command is available to issue operations in
-later steps.
+If using embedded etcd, leave the *etcd* package installed, because the package
```
gaurav-nelson: Can we change it to: 

> If you are using embedded etcd, please leave the *etcd* package installed. It is required for running `etcdctl` command used in later steps.

gaurav-nelson:
Rest looks fine! :baggage_claim:

ahardin-rh:
LGTM :star2:

bfallonf:
@detiber @liggitt bumping again!

detiber (approved):

bfallonf:
Thanks @detiber I'll merge this.

bfallonf:
[rev_history]
|xref:../install_config/downgrade.adoc#install-config-downgrade[Downgrading OpenShift]
|Changed recommendations on embedded etcd.
%
[rev_history]
|xref:../install_config/install/advanced_install.adoc#install-config-install-advanced-install[INstalling a Cluster -> Advanced Installation]
|Changed recommendations on embedded etcd.
%
[rev_history]
|xref:../install_config/install/planning.adoc#install-config-install-planning[Installing a Cluster -> Planning]
|Changed recommendations on embedded etcd.
%

PR #3298, opened 2016-11-30
Prevent LD_PRELOAD error

gbraad:
This will prevent errors of the library not being found.

### DIFF
--- a/creating_images/guidelines.adoc
+++ b/creating_images/guidelines.adoc
@@ -305,7 +305,7 @@ start script:
 export USER_ID=$(id -u)
 export GROUP_ID=$(id -g)
 envsubst < ${HOME}/passwd.template > /tmp/passwd
-export LD_PRELOAD=libnss_wrapper.so
+export LD_PRELOAD=/usr/lib64/libnss_wrapper.so
 export NSS_WRAPPER_PASSWD=/tmp/passwd
 export NSS_WRAPPER_GROUP=/etc/group
 ----


bfallonf:
Thanks for the PR @gbraad .

Hey @bparees can I get an ack this is correct? Big thanks.

bparees:
this isn't something we've found necessary in any of our images, probably because /usr/lib64 is part of the normal ld_library_path, do you have a scenario where this change is necessary?
https://github.com/openshift/jenkins/blob/master/2/contrib/jenkins/jenkins-common.sh#L27

gbraad:
This was necessary for custom application images based on fedora:24 and fedora:25. They looked as follows:

`Dockerfile`
```
FROM fedora:25
RUN dnf install -y gettext nss_wrapper && dnf clean all
RUN adduser user -u 1000 -g 0 -r -m -d /home/user/ -c "Application User" -l
ADD run.sh /home/user/
USER 1000
CMD ["/home/user/run.sh"]
```

`run.sh`
```
#!/bin/sh
export USER_ID=$(id -u)
export GROUP_ID=$(id -g)
envsubst < /tmp/passwd_template > /tmp/passwd
export LD_PRELOAD=libnss_wrapper.so
export NSS_WRAPPER_PASSWD=/tmp/passwd
export NSS_WRAPPER_GROUP=/etc/group
/home/user/apptorun
```

The app to run throws this error when forking a new process. When adding the `export LD_PRELOAD=/usr/lib64/libnss_wrapper.so` this did not occur. Haven't tried if using `exec` would have fixed the issue.

bparees:
ok, well I guess it shouldn't hurt, so lgtm.

bfallonf:
Cool. Thanks for the eyes @bparees 

Also, Ben, @gbraad  would version of OCP would this apply to?

bparees:
Any/all

Ben Parees | OpenShift

gbraad:
as @bparees says, 'any/all'. This is not version specific.
However, I have tested this on OpenShift Online (preview), and Origin v1.3.1

ahardin-rh:
Should a label for OCP 3.3 also be added?

bparees:
@ahardin-rh  this change is not specific to any particular release of openshift.  it's about image construction.  So however you want to label it to account for that.

ahardin-rh:
Thanks!

ahardin-rh:
[rev_history]
|xref:../creating_images/guidelines.adoc#creating-images-guidelines[Guidelines]
|In the xref:../creating_images/guidelines.adoc#use-uid[Support Arbitrary User IDs] section, updated `export LD_PRELOAD=libnss_wrapper.so` to be `export LD_PRELOAD=/usr/lib64/libnss_wrapper.so` to prevent errors from the library not being found.
%

PR #3311, opened 2016-12-01
Fix EgressNetworkPolicy example

danwinship:
The example actually worked as documented in 3.3, but that's because there was a bug in the 3.3 code. In 3.4 the bug has been fixed so the example no longer works as-is.

(Thanks to @weliang1 for noticing)

### DIFF
--- a/admin_guide/managing_pods.adoc
+++ b/admin_guide/managing_pods.adoc
@@ -325,7 +325,7 @@ To configure pod access limits:
             {
                 "type": "Deny",
                 "to": {
-                    "cidrSelector": "0.0.0.0/32"
+                    "cidrSelector": "0.0.0.0/0"
                 }
             }
         ]


knobunc (approved):
LGTM

knobunc:
@openshift/team-documentation please can we get this merged into 3.3 docs and beyond.

danwinship:
> please can we get this merged into 3.3 docs and beyond.

3.**4** and beyond. We should leave it as-is in the 3.3 docs since the 3.3 code is buggy and won't work if you specify it correctly

ahardin-rh:
LGTM. Thanks!

ahardin-rh:
[rev_history]
|Updated the `EgressNetworkPolicy` example in the xref:../admin_guide/managing_pods.adoc#admin-guide-config-pod-access[Configuring Pod Access Limits] section.
%

PR #3323, opened 2016-12-05
EgressNetworkPolicy clarifications

danwinship:
Some clarifications based on problems @weliang1 ran into while testing:
- be more explicit that it only affects external traffic, not pod-to-pod traffic
- be more explicit about rule ordering
- mention that you can only have a single EgressNetworkPolicy in a namespace
- mention that breaking the rules may result in all egress traffic from the namespace being dropped
- merge together the redundant warnings about global namespaces

### DIFF
--- a/admin_guide/managing_pods.adoc
+++ b/admin_guide/managing_pods.adoc
@@ -254,7 +254,7 @@ address.
 === Limiting Pod Access with Egress Firewall
 
 As an {product-title} cluster administrator, you can use egress policy to limit
-the addresses that some or all pods can access from within the cluster, so that:
+the external addresses that some or all pods can access from within the cluster, so that:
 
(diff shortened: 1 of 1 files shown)


ahardin-rh:
@danwinship Thanks. Is this for 3.4?

danwinship:
yes

ahardin-rh:
Thanks! I will apply some follow-up edits in a separate PR.

ahardin-rh:
[rev_history]
|Added clarifying details around `EgressNetworkPolicy`, including that it only affects external traffic (not pod-to-pod traffic) and that you can only have a single `EgressNetworkPolicy` in a namespace; added additional details about rule ordering and that breaking the rules can result in all egress traffic from the namespace being dropped.
%

PR #3300, opened 2016-11-30
Added Azure Block Provisioner

pospispa:
PTAL @rootfs @mfojtik @openshift/team-documentation
Target versions: 3.3 and 3.4

I know that the file `install_config/persistent_storage/storage_classes.adoc` is currently not used because it's not in the `_topic_map.yml` file as it was removed in [this commit](https://github.com/openshift/openshift-docs/commit/95c8d88057016513175a24ec739e3458c8fddd52).

IMHO, information about `StorageClass` `parameters` present in the file `install_config/persistent_storage/storage_classes.adoc` should be present somewhere in OpenShift documentation, however, I don't know which place is the right one.

Anyway, this PR adds information on Azure block provisioner that was added to Kubernetes in [this PR](https://github.com/kubernetes/kubernetes/pull/30091).

PR #3324, opened 2016-12-05
Removed Azure block plug-in references since it doesn't support storage class yet

ahardin-rh:
Per discussion in https://github.com/openshift/openshift-docs/pull/3300


ahardin-rh:
No rev history needed

PR #3331, opened 2016-12-06
Use latest epel-release package in documentation

lmello:
Using specific epel-release versions will demand maintenance in the documentation every time a new release of epel-release is launched. The current release is 7-8, but the documentation refer to 7-5, which does not exist in the repo and is broken.

This commits fix this problem by using the epel-release latest version provided for RHEL 7. This package name will not change when the epel release is updated.

### DIFF
--- a/install_config/install/host_preparation.adoc
+++ b/install_config/install/host_preparation.adoc
@@ -135,7 +135,7 @@ package source for Ansible:
 +
 ----
 # yum -y install \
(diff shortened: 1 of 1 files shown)


ahardin-rh:
Thank you!

ahardin-rh:
[rev_history]
|xref:../install_config/install/host_preparation.adoc#install-config-install-host-preparation[Installing a Cluster -> Host Preparation]
|Updated the path to the latest *epel-release* package.
%

PR #3334, opened 2016-12-06
refactor template author content

bparees:
@jwforres ptal


bparees:
@openshift/team-documentation technically this applies to older OCP versions too (it's just moving content around) but some of the content being moved is specific to OCP3.4+ (the ${{foo}} syntax), so i don't know how you want to handle that.

bparees:
@jwforres fixed up the duplication.

ahardin-rh:
Thank you! I will make some minor style edits in a separate follow-up PR.

ahardin-rh:
[rev_history]
|xref:../dev_guide/templates.adoc#dev-guide-templates[Templates]
|Added examples of a template object definition, template description metadata, template object labels, generating a parameter value, setting an explicit value as the default value, and a full template with parameter definitions and references,
%

PR #3342, opened 2016-12-07
Bug 1390363, Expanding the Ingress Documentation

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1390363

Preview builds: 

http://file.rdu.redhat.com/~ahardin/12092016/ingress-IP/dev_guide/getting_traffic_into_cluster.html#using-externalIP

http://file.rdu.redhat.com/~ahardin/12092016/ingress-IP/admin_guide/high_availability.html#configuring-a-highly-available-service

### DIFF
--- a/admin_guide/high_availability.adoc
+++ b/admin_guide/high_availability.adoc
@@ -270,7 +270,6 @@ The list of command options described here are a subset that are relevant to thi
 * Are accessible on the configured hosts from outside the cluster.
 * Are not used for any other purpose within the cluster.
 
-
 *Keepalived* on each node determines whether the needed service is running. If it is, VIPs are supported and *Keepalived* participates in the negotiation to determine which node will serve the VIP. For a node to participate, the service must be listening on the watch port on a VIP or the check must be disabled.
 
 [NOTE]
@@ -406,9 +405,10 @@ Before starting the *keepalived* daemon, the startup script verifies the
 `iptables` rule that allows multicast traffic to flow.  If there is no such
 rule, the startup script creates a new rule and adds it to the IP tables
 configuration.  Where this new rule gets added to the IP tables configuration
-depends on the `--iptables-chain=` option.  If there is an `--iptables-chain=`
+depends on the `--iptables-chain=` option. If there is an `--iptables-chain=`
 option specified, the rule gets added to the specified chain in the option.
 Otherwise, the rule is added to the `INPUT` chain.
+
 
 [IMPORTANT]
 ====
@@ -627,8 +627,11 @@ $ oadm ipfailover ipf-ha-router-us-west \
 endif::[]
 
 +
-Below is ipfailover command for the *geo-cache* service that is listening on port 9736.  Since there are two ipfailover deployment configration the `--vrrp-id-offset` must be set, so that each VIP gets its own offset. In this case, setting a value of 10 means that the `ipf-ha-router-us-west` can have a maximum of 10 VIPs (0-9) since `ipf-ha-geo-cache`
-is starting at 10.
+Below is ipfailover command for the *geo-cache* service that is listening on
+port 9736.  Since there are two ipfailover deployment configration the
+`--vrrp-id-offset` must be set, so that each VIP gets its own offset. In this
+case, setting a value of 10 means that the `ipf-ha-router-us-west` can have a
+maximum of 10 VIPs (0-9) since `ipf-ha-geo-cache` is starting at 10.
 +
 ifdef::openshift-enterprise[]
 ====
@@ -660,8 +663,26 @@ In the commands above, there are ipfailover, *router*, and *geo-cache* pods on e
 The five VIP addresses in each example, 10.245.{2,3}.101-105, are served by the twoipfailover deployment configration.
 IP failover dynamically selects which address is served on which node.
 +
-The admin sets up external DNS to point to the VIP addresses knowing that all the *router* VIPs point to the same *router*, and all the *geo-cache* VIPs point to the same *geo-cache* service. As long as one node
+The administrator sets up external DNS to point to the VIP addresses knowing that all the *router* VIPs point to the same *router*, and all the *geo-cache* VIPs point to the same *geo-cache* service. As long as one node
 remains running, all the VIP addresses are served.
+
+. Deploy the ipfailover router to monitor postgresql listening on node
+port 32439 and the external IP address, as defined in the *postgresql-ingress*
+service:
++
+====
+----
+$ oadm ipfailover ipf-ha-postgresql \
+    --replicas=1 <1> --selector="app-type=postgresql" <2> \
+    --virtual-ips=10.9.54.100 <3> --watch-port=32439 <4>  \
+    --credentials=/etc/origin/master/openshift-router.kubeconfig \
+    --service-account=ipfailover --create
+----
+<1> Specifies the number of instances to deploy.
+<2> Restricts where the ipfailover is deployed.
+<3> Virtual IP address to monitor.
+<4> Port on which ipfailover will monitor on each node.
+====
 
 [[dynamically-updating-vips-for-a-highly-available-service]]
 === Dynamically Updating Virtual IPs for a Highly-available Service
--- a/dev_guide/getting_traffic_into_cluster.adoc
+++ b/dev_guide/getting_traffic_into_cluster.adoc
@@ -27,9 +27,9 @@ xref:using-externalIP[ExternalIP], or xref:using-nodeport[NodePort].
 TCP or UDP offers several approaches:
 
 - Use the non-cloud xref:using-the-loadbalancer[Load Balancer]. This limits you to
-a single ingress IP (which can be a virtual IP (VIP), but still is a single machine for
-initial load balancing). It simplifies the administrator's job, but uses one IP
-per service.
+a single edge router IP (which can be a virtual IP (VIP), but still is a single
+machine for initial load balancing). It simplifies the administrator's job, but
+uses one IP per service.
 - Manually assign xref:using-externalIP[ExternalIPs] to the service. You can
 assign a set of IPs, so you can have multiple machines for the incoming load
 balancing. However, this requires elevated permissions to assign, and manual
@@ -74,8 +74,8 @@ covers web applications.
 
 An administrator can create a
 xref:../install_config/install/prerequisites.adoc#prereq-dns[wildcard DNS]
-entry, and then set up a router. Afterward, the users can self-service the
-ingress without having to contact the administrators. The router has controls to
+entry, and then set up a router. Afterward, the users can self-service the edge
+router without having to contact the administrators. The router has controls to
 allow the administrator to specify whether the users can self-provision host
 names, or if they must fit a pattern the administrator defines. The other
 solutions require the administrator to do the provisioning, or they require that
@@ -106,7 +106,7 @@ options are also available.
 The non-cloud
 xref:../admin_guide/tcp_ingress_external_ports.adoc#admin-guide-expose-external-ports[load
 balancer] allocates a unique IP from a configured pool. This limits you to a
-single ingress IP, which can be a VIP, but still will be a single machine for
+single edge router IP, which can be a VIP, but still will be a single machine for
 initial load balancing. The non-cloud load balancer simplifies the
 administrator's job by providing the needed IP address, but uses one IP per
 service.
@@ -114,8 +114,12 @@ service.
 [[using-externalIP]]
 == Using a Service ExternalIP
 
-Administrators can assign a list of xref:../architecture/core_concepts/pods_and_services.adoc#service-externalip[externalIPs], for which nodes in the cluster will also accept traffic for the service. These IPs are not managed by {product-title} and administrators are responsible for ensuring that traffic arrives at a node with this IP.
-A common example of this is configuing a
+Administrators can assign a list of
+xref:../architecture/core_concepts/pods_and_services.adoc#service-externalip[externalIPs],
+for which nodes in the cluster will also accept traffic for the service. These
+IPs are not managed by {product-title} and administrators are responsible for
+ensuring that traffic arrives at a node with this IP. A common example of this
+is configuring a
 ifdef::openshift-enterprise,openshift-origin[]
 xref:../../admin_guide/high_availability.adoc#admin-guide-high-availability[highly available]
 endif::[]
@@ -124,12 +128,19 @@ highly available
 endif::[]
 service.
 
-The supplied list of IP addresses is used for load balancing
-incoming requests. The service port is opened on the externalIPs on all nodes running *kube-proxy*.
+The supplied list of IP addresses is used for load balancing incoming requests.
+The service port is opened on the externalIPs on all nodes running *kube-proxy*.
+
 [NOTE]
 ====
-ExternalIPs require elevated permissions to assign, and manual tracking of the IP:ports that are in use.
+ExternalIPs require elevated permissions to assign, and manual tracking of the
+IP:ports that are in use.
 ====
+
+Traffic from hosts that are external to the cluster that is going to the
+external IP address ends up on a node in the cluster. When it arrives, the
+service that set up the external IP redirects it to a service endpoint that it
+is managing. The service load balances among its endpoints.
 
 An externally visible IP for the service can be configured in several ways:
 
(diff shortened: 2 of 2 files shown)


ahardin-rh:
@knobunc @rjhowe PTAL. Thanks!

dev_guide/getting_traffic_into_cluster.adoc
```diff
@@ -133,10 +146,298 @@ xref:../admin_guide/high_availability.adoc#admin-guide-high-availability[VIP] ad
...
+* There are no protections in place to restrict the usage of a particular external
+address configured within the cluster. This allows the potential for a single
+address to be used by multiple services targeting the same port. When this
+situation occurs, the service which requested the port first is given use of the
+port.
+
+xref:using-ingress[Ingress] provides a solution.
+
+[[using-ingress]]
+=== Ingress
```
knobunc: We can not call it this.  Ingress is too generic and the term is already used in kubernetes for something else.

I have no great ideas for another name, just some bad ones:
- External IP Self-Service
- External IP Auto-allocation
- Ingress IP Self-Service or Auto-Allocation

dev_guide/getting_traffic_into_cluster.adoc
```diff
@@ -133,10 +146,298 @@ xref:../admin_guide/high_availability.adoc#admin-guide-high-availability[VIP] ad
...
+used to run the pod containing the router.
+
+. Pre-configure the service account to be a member of the privileged security
+context constraint by executing the following command:
++
+----
+$ oadm policy add-scc-to-user privileged system:serviceaccount:default:ipfailover
+----
+
+. *Multicast* must be enabled on each node the router will be deployed to in order
```
knobunc: We are not enabling multicast here, we are allowing it to pass through the firewall.

dev_guide/getting_traffic_into_cluster.adoc
```diff
@@ -133,10 +146,298 @@ xref:../admin_guide/high_availability.adoc#admin-guide-high-availability[VIP] ad
...
+deploying additional routers for each external services, multiple addresses or a
+range of addresses can be configured within a single router to limit the amount
+of administrative actions required and reduce the amount of required resources.
+This reduces the need to have an administrator intervene when configuring new
+services using an externalIP.
+
+When deploying the *ipfailover* router with the `oadm ipfailover` command, the
+`--virtual-ips` flag defines the addresses served by the router. To allow for
+more than one address to be served, a comma-separated list of addresses
+(10.9.54.100,10.9.54.110) or a range of multiple addresses (10.9.54.100-110) can
```
knobunc: There is a limit of 254 IP addresses that can be exposed.

knobunc (dismissed):
Good start.  @marun @pecameron PTAL

dev_guide/getting_traffic_into_cluster.adoc
```diff
@@ -133,10 +146,298 @@ xref:../admin_guide/high_availability.adoc#admin-guide-high-availability[VIP] ad
...
+availability] strategies by deploying the IP failover router to provide access
+services configured with external IP addresses. This allows cluster
+administrators the flexibility of defining the ingress points within a cluster,
+and ensuring the service is readily available.
+
+[NOTE]
+====
+Nodes that have IP failover routers deployed to them must be in the same *Layer
+2* switching domain for ARP broadcasts to communicate to switches what
+appropriate port the destination should flow to.
```
pecameron: "switch" port. 

I am not sure where this is going. What does it take to have a working cluster that breaks this?

Stray comment: each port on the switch is wired to a NIC on one of the nodes. The switch keeps track of which mac addresses arrive on each switch port so when a packet arrives on some port it can be directly "switched" the port that saw its destination mac. Hence the device is called a switch. Switches are layer 2 and arp broadcasts are layer 2 so an arp broadcast goes to all ports on the layer 2 segment. I think that all of the nodes are in the same layer 2 segment (Ben ?)

dev_guide/getting_traffic_into_cluster.adoc
```diff
@@ -133,10 +146,298 @@ xref:../admin_guide/high_availability.adoc#admin-guide-high-availability[VIP] ad
...
+====
+
+[CAUTION]
+====
+Only 254 addresses can be allocated in a single VLAN due to the restrictions
+placed by the Virtual Router Redundancy Protocol (VRRP), which is used by the
+IP failover router.
```
pecameron: xref:...[High availability] is limited to a maximum of 255 VIPs. This is a limitation of the VRRP protocol. The VIPs do not have to be sequential.

dev_guide/getting_traffic_into_cluster.adoc
pecameron: @knobunc VRP requires 244.0.0.18/32 which is multicast. The NICs must be configured for multicast (linux layer, not openshift)

dev_guide/getting_traffic_into_cluster.adoc
```diff
@@ -133,10 +146,298 @@ xref:../admin_guide/high_availability.adoc#admin-guide-high-availability[VIP] ad
...
+. Pre-configure the service account to be a member of the privileged security
+context constraint by executing the following command:
++
+----
+$ oadm policy add-scc-to-user privileged system:serviceaccount:default:ipfailover
+----
+
+. *Multicast* must be enabled on each node the router will be deployed to in order
+for it to receive traffic for 224.0.0.18 (the VRRP multi-cast IP). Depending on
+your environment, enable *multicast* on each of the nodes by executing:
```
pecameron: This really a ipfailover discussion. I would just point to ipfailover.

dev_guide/getting_traffic_into_cluster.adoc
pecameron: @knobunc Actually 255 (range 1..255) 0 is reserved in the RFC

dev_guide/getting_traffic_into_cluster.adoc
pecameron: @knobunc I am not sure about this ipfailover discussion. I think that ipfailover section should cover this at this level of detail. Maybe a higher level treatment would make the point?

pecameron (changes requested):
Lots of good stuff here. This material is not easy to present in a clear way that leads the reader along. I took a pass at it a while ago and I am glad this is getting more attention. I try to be precise (and I don't always succeed) with use of terms in an effort to keep confusion down.

dev_guide/getting_traffic_into_cluster.adoc
```diff
@@ -133,10 +146,298 @@ xref:../admin_guide/high_availability.adoc#admin-guide-high-availability[VIP] ad
...
+kind: Service
+metadata:
+  creationTimestamp: null
+  labels:
+    app: postgresql-persistent
+    template: postgresql-persistent-template
+  name: postgresql-ingress
+spec:
+  deprecatedPublicIPs:
+  - 10.9.54.100
```
ahardin-rh: @rjhowe do you know what depricated public IP represents? I could not find any other reference to it. Sourced from the example content you pointed me to.

ahardin-rh:
@knobunc @pecameron I made some updates and agree that some of the IP Failover discussion can just point to that section of the docs. These details were sourced from https://github.com/sabre1041/openshift-playbooks/blob/54ef2b1119225674241dbffb3824701b448202c2/playbooks/operationalizing/ingress.adoc so maybe some of it is in the weeds. Please let me know if I'm getting closer. Thanks!

dev_guide/getting_traffic_into_cluster.adoc
```diff
@@ -133,10 +146,212 @@ xref:../admin_guide/high_availability.adoc#admin-guide-high-availability[VIP] ad
...
+    port: 5432
+  type: LoadBalancer <1>
+  selector:
+    name: postgresql
+INGRESS
+----
+<1> The `LoadBalancer` type of service will make the request for an external service
+on behalf of the application user.
+
+Alternatively, the `oc expose` command can be used to create the service:
```
pecameron: "expose" creates a route.

dev_guide/getting_traffic_into_cluster.adoc
```diff
@@ -133,10 +146,212 @@ xref:../admin_guide/high_availability.adoc#admin-guide-high-availability[VIP] ad
...
+
+.Example oc get Output
+----
+NAME         CLUSTER-IP      EXTERNAL-IP   PORT(S)   AGE
+postgresql-ingress    172.30.74.106   10.9.54.100,10.9.54.100    5432/TCP    30s
+----
+
+Specifying the type `LoadBalancer` also configures the service with a
+`nodePort` value, allowing direct access to the service on each of the nodes the
+application is deployed on. To discover the node port automatically assigned,
```
pecameron: nodePort exposes the service port on all nodes in the cluster.Any packet that arrives on any node in the cluster targeting the nodeport ends up in the service. From there it is load balanced to the service's endpoints.

pecameron (commented):
Coming along. I wish I knew some of the details that you need for this. We need to get @knobunc involved in this.

ahardin-rh:
@knobunc Can you please provide some follow-up direction to @pecameron 's comments? Thanks!

ahardin-rh:
@knobunc can you please take a look at the comments here when you get a chance? I'd like to get this moving forward again, but need your expertise. Thanks!

dev_guide/getting_traffic_into_cluster.adoc
```diff
@@ -133,10 +146,298 @@ xref:../admin_guide/high_availability.adoc#admin-guide-high-availability[VIP] ad
...
+
+[[using-ingress-defining-the-ingress-IP-range]]
+==== Defining the Ingress IP Range
+
+The ability for cluster administrators to automatically allocate External IP
+addresses using Ingress is enabled by default within {product-title} and
+configured to use the `172.46.0.0/16` range. An alternate range can be specified
+by configuring the `ingressIPNetworkCIDR` parameter in the
+*_/etc/origin/master-config.yaml_* file:
+
```
knobunc: If an IP address was allocated from the ingressIPNetworkCIDR range, and then the range is changed I believe the service's externalIP gets reallocated an address that is in the new range.  @marun is that correct?

Maru also mentioned that the IP allocations _could_ change at any time, but likely would not... Maru, am I misrepresenting your statement, and if not, can you clarify what we want to say in the docs?

dev_guide/getting_traffic_into_cluster.adoc
```diff
@@ -133,10 +146,298 @@ xref:../admin_guide/high_availability.adoc#admin-guide-high-availability[VIP] ad
...
+----
+$ oc get svc postgresql-ingress
+----
+
+.Example oc get Output
+----
+NAME         CLUSTER-IP      EXTERNAL-IP   PORT(S)   AGE
+postgresql-ingress    172.30.74.106   10.9.54.100,10.9.54.100    5432/TCP    30s
+----
```
knobunc: That is the way the CLI behaves (for stupid reasons).  I believe there is a bug logged, so it may change.

dev_guide/getting_traffic_into_cluster.adoc
knobunc: > api/swagger-spec/openshift-openapi-spec.json
> 52844:     "deprecatedPublicIPs": {
> 52845:      "description": "deprecatedPublicIPs is deprecated and replaced by the externalIPs field with almost the exact same semantics.  This field is retained in the v1 API for compatibility until at least 8/20/2016.  It will be removed from any new API revisions.  If both deprecatedPublicIPs *and* externalIPs are set, deprecatedPublicIPs is used.",
> 52852:      "description": "externalIPs is a list of IP addresses for which nodes in the cluster will also accept traffic for this service.  These IPs are not managed by Kubernetes.  The user is responsible for ensuring that traffic arrives at a node with this IP.  A common example is external load-balancers that are not part of the Kubernetes system.  A previous form of this functionality exists as the deprecatedPublicIPs field.  When using this field, callers should also clear the deprecatedPublicIPs field.",
> 

Basically it's historical and is going away.

dev_guide/getting_traffic_into_cluster.adoc
```diff
@@ -133,10 +146,298 @@ xref:../admin_guide/high_availability.adoc#admin-guide-high-availability[VIP] ad
...
+    targetPort: 5432
+  selector:
+    name: postgresql
+  sessionAffinity: None
+  type: LoadBalancer
+----
+<1> Automatically assigned port.
+
+A PostgreSQL client can now be configured to connect directly to any node using
+the value of the assigned `nodePort`.
```
knobunc: I don't have the context for this in the review.  Does the question still hold?

A nodeport would work with ANY IP address that allows traffic to terminate at any node in the cluster.

dev_guide/getting_traffic_into_cluster.adoc
```diff
@@ -133,10 +146,212 @@ xref:../admin_guide/high_availability.adoc#admin-guide-high-availability[VIP] ad
...
+addresses using a CIDR notation, which allows an application user to make a
+request against the cluster for an external IP address. When a service is
+configured with the type `LoadBalancer`, an External IP address will be
+automatically assigned from the designated range.
+
+[NOTE]
+====
+Ingress IP Self-Service is only applicable in non-cloud based environments.
+====
+
```
knobunc: We only set up the rules to allow the traffic to get in if it arrives on the external IP and port.  If they have other rules that block access to the port, then they need to relax those.

The point is to allow people to self-provision external IP addresses.  Normally that requires cluster admin privs, but by turning it into LoadBalancer services, users can create them (subject to quota restrictions).

As to the DNS names, yes... the user would have to examine the service and update DNS.  Ideally they would have something that they ran that watched the service and saw the externalIP and then updated DNS whenever there was a change (and that would be a nice sample container and docs to provide).  But initially people will probably do it manually.

dev_guide/getting_traffic_into_cluster.adoc
knobunc: No, it can do both.  Surprise!

If you expose a dc, rc, or pod you get a service.

If you expose a service, you get a route.

admin_guide/high_availability.adoc
```diff
@@ -244,17 +243,17 @@ Each VIP in the set may end up being served by a different node.
...
 
-Before starting the *keepalived* daemon, the startup script verifies the `iptables` rule that allows multicast traffic to flow. 
-If there is no such rule, the startup script creates a new rule and adds it to the iptables configuration. 
-Where this new rule gets added to the iptables configuration, depends on the `--iptables-chain=` option. 
-If there is an `--iptables-chain=` option specified, the rule gets added to the specified chain in the option; 
+Before starting the *keepalived* daemon, the startup script verifies the `iptables` rule that allows multicast traffic to flow.
+If there is no such rule, the startup script creates a new rule and adds it to the iptables configuration.
+Where this new rule gets added to the iptables configuration, depends on the `--iptables-chain=` option.
+If there is an `--iptables-chain=` option specified, the rule gets added to the specified chain in the option;
 otherwise the rule is added to the `INPUT` chain.
```
knobunc: I'm fine with the original wording

admin_guide/high_availability.adoc
```diff
@@ -264,7 +263,7 @@ The `iptables` rule must be present whenever there is one or more *keepalived* d
 
 The `iptables` rule can be removed after the last *keepalived* daemon terminates. The rule is not automatically removed.
 
-You can manually manage the `iptables` rule on each of the nodes. It only gets created when none is present (as long as ipfailover is not created with the -`-iptable-chain=""` option).  
+You can manually manage the `iptables` rule on each of the nodes. It only gets created when none is present (as long as ipfailover is not created with the -`-iptable-chain=""` option).
```
knobunc: I'm fine with the original wording

dev_guide/getting_traffic_into_cluster.adoc
```diff
@@ -133,10 +146,212 @@ xref:../admin_guide/high_availability.adoc#admin-guide-high-availability[VIP] ad
...
+----
+<1> The IP address assigned to the external IP.
+
+The external IP address is not managed by the underlying Kubernetes
+infrastructure and must be maintained and provided by a cluster administrator.
+While external IPs provide a solution for accessing services on the {product-tlte}
+cluster, there are several shortcomings:
+
+* The cluster administrator user must ensure the external IP is not in any range
+that the cluster can use. The user needs to work with the administrator that
```
knobunc: "cluster can use" -> "cluster is configured to use"

dev_guide/getting_traffic_into_cluster.adoc
```diff
@@ -133,10 +146,212 @@ xref:../admin_guide/high_availability.adoc#admin-guide-high-availability[VIP] ad
...
+controls the network external to the cluster to assign the external IP address,
+a DNS name for the IP, and ensure traffic to the IP can get to the nodes in the
```
knobunc: The DNS name is not strictly necessary... and in fact may not be assigned by the operator of the cluster.

For instance, if I use an externalIP on a public cluster (e.g. openshift.com), I need to configure the DNS I control for benbennett.com to point to the IP address that was assigned.

dev_guide/getting_traffic_into_cluster.adoc
```diff
@@ -133,10 +146,212 @@ xref:../admin_guide/high_availability.adoc#admin-guide-high-availability[VIP] ad
...
+cluster. The cluster administrator must ensure firewall rules permit the packets
+to get into the nodes.
```
knobunc: for the ports they want to expose.

knobunc (commented):
Good stuff.  Sorry for ignoring it for ages.  @pecameron thanks for the diligent and useful reviews.

git001:
I have created some pictures to display the example setup and the packet flow.

https://me2digital.online/2017/01/12/openshift-ingress-setup/

Mind you to use some of this picture or create a own one, because it's easier to explain when you have a picture.

What I miss in the current doc is the call of

```
oadm ipfailover ipf-ha-postgresql ...
```
as mentioned in the link below

https://github.com/sabre1041/openshift-playbooks/blob/54ef2b1119225674241dbffb3824701b448202c2/playbooks/operationalizing/ingress.adoc

@sabre1041 thanks for the detailed description.

knobunc:
That's good stuff.  But can we please try to avoid the work "ingress" or "ingress router" as much as possible?  The problem is that upstream Kubernetes has defined an Ingress object (that is very much like our route object), and we really don't want to overload the terminology.  Perhaps "edge router" is less confusing?

ahardin-rh:
@knobunc @pecameron Picking this back up. Thanks for your feedback. I made some edits. Please let me know if I am missing anything.

@git001 Thanks for your feedback as well. I can see if our graphics team can create a graphic for us to use, but that will come in a future PR. As for the `oadm ipfailover ipf-ha-postgresql` command, I added this to the High Availability topic. Please let me know if the placement is correct.

pecameron (commented):
Looks good. General question on line breaks: what is the max number of characters before a new line? I thought it was 120 but one of the docs has entire paragraphs in a single line and I noticed you re-did some of the longer lines. Just curious...

ahardin-rh:
@pecameron Thanks! Edits complete. As for your question about line length, most of us on the docs team use Atom as our editor. Atom will re-flow at 80 by default, so we typically hard wrap lines at 80 to make the source easy to read.

ahardin-rh:
Approved by QE in the BZ

ahardin-rh:
[rev_history]
|xref:../admin_guide/high_availability.adoc#admin-guide-high-availability[High Availability]
|Added a step for deploying the ipfailover router to monitor postgresql listening to the xref:../admin_guide/high_availability.adoc#configuring-a-highly-available-service[Configuring a Highly-available Service] section
%
|xref:../dev_guide/getting_traffic_into_cluster.adoc#getting-traffic-into-cluster[Getting Traffic into the Cluster]
|Added more details about Ingress.
%

PR #3353, opened 2016-12-09
Adding Create From URL documentation

rhamilto:
*  Adding "Configuring the Create From URL Namespace Whitelist" to web_console_customization and fixing heading semantics on that page (several headings were improperly declared
as h3s or h4s when they should have been h2s or h3s).
* Adding "Create from URL" page to the developer guide describing [the "API" for create from URL](https://github.com/openshift/origin-web-console/pull/884#issue-189848824).

### DIFF
--- a/dev_guide/create_from_url.adoc
+++ b/dev_guide/create_from_url.adoc
@@ -1,0 +1,86 @@
+[[dev-guide-create-from-url]]
+= Create from URL
+{product-author}
+{product-version}
+:data-uri:
+:icons:
+:experimental:
+:toc: macro
+:toc-title:
+:prewrap!:
+
+toc::[]
+
+[[overview]]
+== Overview
+
+Create From URL is a function that allows you to construct a URL from an image
+stream, image tag, or template.
+
+Create from URL only works with image streams or templates from
+namespaces that have been explicitly whitelisted.  The whitelist contains
+the 'openshift' namespace by default.  To add namespaces to the whitelist, see
+xref:../install_config/web_console_customization.adoc#configuring-the-create-
+from-url-namespace-whitelist[Configuring the Create From URL Namespace Whitelist].
+
+[[create-for-url-using-an-image-stream-and-image-tag]]
+== Using an Image Stream and Image Tag
+
+[[image-stream-and-image-tag-query-string-parameters]]
+=== Query String Parameters
+
+[options="header"]
+|===
+|Name|Description|Required|Schema|Default
+|imageStream|The value metadata.name as defined in the image stream to be used.|
+true|string|
+|imageTag|The value spec.tags.name as defined in the image stream to be used.|
+true|string|
+|namespace|The name of the namespace containing the image stream and image tag
+to use.|false|string|openshift
+|name|Identifies the resources created for this application.|false|string|
+|sourceURI|The git repository URL containing the application source code.|false|
+string|
+|sourceRef|The branch, tag, or commit for the application source code specified
+in sourceURI.|false|string|
+|contextDir|The subdirectory for the application source code specified in
+sourceURI, used as the context directory for the build.|false|string|
+|===
+
+[NOTE]
+====
+link:https://en.wikipedia.org/wiki/Percent-encoding#Percent-encoding_reserved_characters[Reserved characters] in parameter values should be URL encoded.
+====
+
+[[example-usage-of-an-image-stream-and-image-tag]]
+==== Example
+----
+ create?imageStream=nodejs&imageTag=4&name=nodejs&sourceURI=https%3A%2F%2Fgithub.com%2Fopenshift%2Fnodejs-ex.git&sourceRef=master&contextDir=%2F
+----
+
+[[create-from-url-using-a-template]]
+== Using a Template
+
+[[template-query-string-parameters]]
+=== Query String Parameters
+
+[options="header"]
+|===
+|Name|Description|Required|Schema|Default
+|template|The value of metadata.name as defined in the template to be used.|
+true|string|
+|templateParamsMap|A JSON parameters map containing the template parameter name
+and corresponding value you wish to override.|false|JSON|
+|namespace|The name of the namespace containing the template to use.|false|string|openshift
+|===
+
+[NOTE]
+====
+link:https://en.wikipedia.org/wiki/Percent-encoding#Percent-encoding_reserved_characters[Reserved characters] in parameter values should be URL encoded.
+====
+
+[[example-usage-of-a-template]]
+==== Example
+----
+ create?template=nodejs-mongodb-example&templateParamsMap={"SOURCE_REPOSITORY_URL"%3A"https%3A%2F%2Fgithub.com%2Fopenshift%2Fnodejs-ex.git"}
+----
--- a/install_config/web_console_customization.adoc
+++ b/install_config/web_console_customization.adoc
@@ -115,7 +115,7 @@ window.OPENSHIFT_EXTENSION_PROPERTIES = {
 ====
 
 [[customizing-the-logo]]
-=== Customizing the Logo
+== Customizing the Logo
 
 The following style changes the logo in the web console header:
 
@@ -142,7 +142,7 @@ assetConfig:
 ----
 
 [[changing-links-to-documentation]]
-=== Changing Links to Documentation
+== Changing Links to Documentation
 
 Links to external documentation are shown in various sections of the web
 console. The following example changes the URL for two given links to the
@@ -164,7 +164,7 @@ assetConfig:
 ----
 
 [[adding-or-changing-links-to-download-the-cli]]
-=== Adding or Changing Links to Download the CLI
+== Adding or Changing Links to Download the CLI
 
 The *About* page in the web console provides download links for the
 xref:../cli_reference/index.adoc#cli-reference-index[command line interface (CLI)] tools. These
@@ -250,9 +250,9 @@ assetConfig:
 ----
 
 [[configuring-navigation-menus]]
-=== Configuring Navigation Menus
+== Configuring Navigation Menus
 
-==== Top Navigation Dropdown Menus
+=== Top Navigation Dropdown Menus
 
 The top navigation bar of the web console contains the help icon and the user
 dropdown menus. You can add additional menu items to these using the
@@ -300,7 +300,7 @@ angular
 hawtioPluginLoader.addModule('<myExtensionModule>');
 ----
 
-==== Project Left Navigation
+=== Project Left Navigation
 
 When navigating within a project, a menu appears on the left with primary and
 secondary navigation. This menu structure is defined as a constant and can be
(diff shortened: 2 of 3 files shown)


jwforres:
@sspeiche @luciddreamz any thoughts on where in the docs we should put the information for the API for "Run on OpenShift"  it doesn't really go in the console customization docs because that is for cluster administrators.

luciddreamz:
Put it in the developers section as it's own page.

rhamilto:
ALL:  Thoughts on this [initial draft of the new Create from URL page](https://github.com/rhamilto/openshift-docs/blob/b8d24e89e87330830ec66a0573e78d2e128770c2/dev_guide/create_from_url.adoc)?  Words are hard.

screen shot 2016-12-12 at 3 54 31 pm

rhamilto:
@sspeiche, @luciddreamz:  what do we want to call this feature?  The card describes it as "Run on OpenShift", but I've been referring to it as "Create from URL" as that arguably better describes the functionality.  I completely defer to y'all.

luciddreamz:
**Create from URL** sounds like a better description - that or **Deploy on OpenShift Button**.

sspeiche:
I like what you have, I'd suggest a few updates:

- not sure users would know what you mean by a preordained namespace
- do we need to talk about parameter encoding?
- an example or two would be helpful, using something like the ever helpful numbered items pointing to it which each part does
- any value in showing screenshot or two?

rhamilto:
> not sure users would know what you mean by a preordained namespace

good point.  i've eliminated "from a preordained namespace" since the next paragraph explains about namespace whitelisting.

> do we need to talk about parameter encoding?

i don't think so.  @jwforres or @spadgett?

> an example or two would be helpful, using something like the ever helpful numbered items pointing to it which each part does

i'm guessing you didn't notice the examples since they weren't in the original screenshot?  i've corrected that by adding a second screenshot with the rest of the page.

> any value in showing screenshot or two?

it might be confusing since the page displays the title and description of the image stream or template being used?

rhamilto:
^ bump @jwforres, @spadgett

dev_guide/create_from_url.adoc
```diff
@@ -0,0 +1,73 @@
...
+== Using a Template
+
+=== Query String Parameters
+
+[options="header"]
+|===
+|Name|Description|Required|Schema|Default
+|template|The value of metadata.name as defined in the template to be used.|
+true|string|
+|templateParamsMap|A JSON parameters map containing the name key value and
```
jwforres: this description is really confusing, particularly 'value key value'.  Also an example is probably helpful here.

dev_guide/create_from_url.adoc
```diff
@@ -0,0 +1,73 @@
...
+=== Query String Parameters
+
+[options="header"]
+|===
+|Name|Description|Required|Schema|Default
+|imageStream|The value metadata.name as defined in the image stream to be used.|
+true|string|
+|imageTag|The value spec.tags.name as defined in the image stream to be used.|
+true|string|
+|namespace|The name of the namespace containing the image stream and image tag
```
jwforres: seems redundant to say that it defaults to openshift when you have a whole column dedicated to the Default values

dev_guide/create_from_url.adoc
jwforres: nevermind on the example part, i see it below.  This description could still be improved though

rhamilto:
> @rhamilto I would mention that the parameter values need to be correctly encoded. If you're passing things like URLs, it's likely they'll have special characters (like their own query parameters) that need to be URL escaped.

Based on my own testing and [this comment](https://perishablepress.com/stop-using-unsafe-characters-in-urls/#comment-90254), I don't think we need to do any encoding given these URLs will be included in HTML.

dev_guide/create_from_url.adoc
rhamilto: I struggled with this description and agree it's confusing.  Ideas on how to more clearly state?

dev_guide/create_from_url.adoc
rhamilto: Fair point.  Will remove the two instances.

spadgett:
@rhamilto I'm not sure I follow what the comment is saying, but for some values, it's not possible for the browser to automatically encode them for you. For instance, a source URL that contains query parameters like

http://example.com/code?branch=master&name=myname

Using that as the `sourceURI` in your example without escaping will pass two different name parameters. The trailing `&name=myname` will be treated as a different query parameter, not part of `sourceURI`. (The `?` might also be a problem.)

This could a problem with JSON template parameter values that could have characters like `&` in them. The parameters MUST be encoded if they contain certain characters.

spadgett:
Another example: Source URLs with a `#` like 

http://example.com/code#master

will cause anything from the `#` in the URL to be ignored. The value need to be encoded.

rhamilto:
Ah.  Thanks for clarifying, @spadgett!  I completely overlooked the parenthesis in your initial comment.  Apologies.

rhamilto:
@ahardin-rh, please put this on your radar for 3.5.  This PR is probably ok as is, but could probably benefit from a writer's touch.

install_config/web_console_customization.adoc
```diff
@@ -506,8 +506,59 @@ endif::[]
 
 endif::[]
 
+[[configuring-the-create-from-url-namespace-whitelist]]
+== Configuring the Create From URL Namespace Whitelist
+
+link:../dev_guide/create_from_url.html[Create from URL] only works with image streams or templates from
```
ahardin-rh: You can remove this link since it points to this same page. If you want to point to an anchor on the page, you can use the format xref:link-name

dev_guide/create_from_url.adoc
```diff
@@ -0,0 +1,81 @@
...
+|===
+
+[NOTE]
+====
+link:https://en.wikipedia.org/wiki/Percent-encoding#Percent-encoding_reserved_characters[Reserved characters] in parameter values should be URL encoded.
+====
+
+==== Example
+
+ create?template=nodejs-mongodb-example&templateParamsMap={"SOURCE_REPOSITORY_URL"%3A"https%3A%2F%2Fgithub.com%2Fopenshift%2Fnodejs-ex.git"}
```
ahardin-rh: place the example in a code block by putting ---- before and after it

dev_guide/create_from_url.adoc
```diff
@@ -0,0 +1,81 @@
...
+Create from URL only works with image streams or templates from
+namespaces that have been explicitly whitelisted.  The whitelist contains
+the 'openshift' namespace by default.  To add namespaces to the whitelist, see
+xref:../install_config/web_console_customization.adoc#configuring-the-create-
+from-url-namespace-whitelist[Configuring the Create From URL Namespace Whitelist].
+
+[[using-an-image-stream-and-image-tag]]
+== Using an Image Stream and Image Tag
+
+=== Query String Parameters
```
ahardin-rh: Each heading needs a unique ID before it so that the docs build properly, so for Query String Parameters:
[[image-stream-and-tag-query-string-parameters]]

dev_guide/create_from_url.adoc
```diff
@@ -0,0 +1,81 @@
...
+|contextDir|The subdirectory for the application source code specified in
+sourceURI, used as the context directory for the build.|false|string|
+|===
+
+[NOTE]
+====
+link:https://en.wikipedia.org/wiki/Percent-encoding#Percent-encoding_reserved_characters[Reserved characters] in parameter values should be URL encoded.
+====
+
+==== Example
```
ahardin-rh: This heading needs a unique ID before it

dev_guide/create_from_url.adoc
```diff
@@ -0,0 +1,81 @@
...
+ create?imageStream=nodejs&imageTag=4&name=nodejs&sourceURI=https%3A%2F%2Fgithub.com%2Fopenshift%2Fnodejs-ex.git&sourceRef=master&contextDir=%2F
+
+
+[[using-a-template]]
+== Using a Template
+
+=== Query String Parameters
```
ahardin-rh: Needs a unique ID so maybe:
[[using-a-template-query-string-parameters]]

dev_guide/create_from_url.adoc
```diff
@@ -0,0 +1,81 @@
...
+====
+link:https://en.wikipedia.org/wiki/Percent-encoding#Percent-encoding_reserved_characters[Reserved characters] in parameter values should be URL encoded.
+====
+
+==== Example
+ create?imageStream=nodejs&imageTag=4&name=nodejs&sourceURI=https%3A%2F%2Fgithub.com%2Fopenshift%2Fnodejs-ex.git&sourceRef=master&contextDir=%2F
+
+
+[[using-a-template]]
+== Using a Template
```
ahardin-rh: Update the ID to be unique since this is a pretty generic heading:
[[create-from-url-using-a-template]]

dev_guide/create_from_url.adoc
```diff
@@ -0,0 +1,81 @@
...
+and corresponding value you wish to override.|false|JSON|
+|namespace|The name of the namespace containing the template to use.|false|string|openshift
+|===
+
+[NOTE]
+====
+link:https://en.wikipedia.org/wiki/Percent-encoding#Percent-encoding_reserved_characters[Reserved characters] in parameter values should be URL encoded.
+====
+
+==== Example
```
ahardin-rh: this needs a unique ID also

dev_guide/create_from_url.adoc
```diff
@@ -0,0 +1,81 @@
+[[dev-guide-create-from-url]]
+= Create From URL
```
ahardin-rh: We use title, so this should be Create from URL

dev_guide/create_from_url.adoc
```diff
@@ -0,0 +1,81 @@
...
+:toc: macro
+:toc-title:
+:prewrap!:
+
+toc::[]
+
+[[overview]]
+== Overview
+
+Create from URL allows the construction of a URL that drops the user into the end
```
ahardin-rh: This needs to me more clear. Maybe something like:

Create From URL is a function that allows you to construct a URL from an image stream, image tag, or template. (Not sure if this is an accurate rewrite, but not sure what is meant by "drops the user into"

dev_guide/create_from_url.adoc
```diff
@@ -0,0 +1,81 @@
...
+|templateParamsMap|A JSON parameters map containing the template parameter name
+and corresponding value you wish to override.|false|JSON|
+|namespace|The name of the namespace containing the template to use.|false|string|openshift
+|===
+
+[NOTE]
+====
+link:https://en.wikipedia.org/wiki/Percent-encoding#Percent-encoding_reserved_characters[Reserved characters] in parameter values should be URL encoded.
+====
+
```
ahardin-rh: This heading needs a unique ID

rhamilto:
Thanks, @ahardin-rh.  Comments addressed.

ahardin-rh:
@rhamilto Thank you!

ahardin-rh:
[rev_history]
|xref:../dev_guide/create_from_url.adoc#dev-guide-create-from-url[Create from URL]
|New topic on Create from URL, a function that allows you to construct a URL from an image stream, image tag, or template.
%

PR #3356, opened 2016-12-09
add rslave propagation mode for volume mounts

pweil-:
Reference: https://github.com/openshift/origin/issues/11601 and https://github.com/openshift/origin/pull/10571

Use rslave mode for volume mount when running origin in a container

@pmorie @derekwaynecarr 

@adellape


pweil-:
@adellape also see this conversation https://github.com/openshift/origin/pull/10571#issuecomment-241528802 about support statements for < docker 1.10.  If we need to call that out in this guide let me know.

pweil-:
@ahardin-rh mind reviewing?  This is for a 3.4 blocker bug

ahardin-rh:
@pweil- LGTM. I will create a follow-up PR noting that rslave mode will only work if the Docker version is 1.10 and later and a Red Hat distro.

ahardin-rh:
No revision history needed (shows up in Origin only)

PR #3357, opened 2016-12-09
fix capitalization of clusterIP

sjwalter2:
The docs use "ClusterIP" when the field in yaml is "clusterIP"; using the capital C will cause the steps to not work but without error message or indication why


ahardin-rh:
Thank you!

PR #465, opened 2015-06-08
Added incremental buildConfig

soltysh:
This PR waits for #444 being merged first. I'll rebase on top of it.

@danmcp @bparees ptal in the meantime

PR #313, opened 2015-04-22
Making changes to the scheduler doc based on feedback

abhgupta:
@thoraxe PTAL

PR #281, opened 2015-04-07
Update session secret doc

liggitt:
Wait until https://github.com/openshift/origin/pull/1540 is reviewed/merged

PR #240, opened 2015-03-23
Add PostgreSQL documentation

mnagy:
Documentation for [PostgreSQL Image](https://github.com/openshift/postgresql). As of this writing, it has not been merged yet, though there won't be any documentation needing changes most likely: https://github.com/openshift/postgresql/pull/1

PR #3373, opened 2016-12-13
Fixed faulty links and cross-calling : #37 #42 #99 #192 #211 #240 #281 #313 #465

prakashmishra1598:
Updated index-commercial, fixed cross calling and faulty links on lines #37 #42 #99 #192 #211 #240 #281 #313 #465


ahardin-rh:
Thanks!

PR #3375, opened 2016-12-13
Provide clarification on custom routes during dev preview

wgordon17:
I believe this update is important to lessen questions, and give customers the clarification that this is a dev preview only limitation


wgordon17:
Nevermind, it looks like this information is already covered, just not in the FAQs. (https://docs.openshift.com/online/dev_guide/routes.html#custom-route-and-hosts-and-certificates-restrictions)

PR #2600, opened 2016-08-03
changed link to gluster storage doc

vikram-redhat:

PR #3400, opened 2016-12-15
Applying changes found in PR#2600

ahardin-rh:
Closed https://github.com/openshift/openshift-docs/pull/2600 and applying those changes here since vikram's branch was likely deleted.

The Ifdef in the MongoDB was introduced in https://github.com/openshift/openshift-docs/pull/2780


ahardin-rh:
No revision history needed.

PR #3444, opened 2017-01-04
Remove all deprecated 'oc env' command by 'oc set env'

guangxuli:


guangxuli:
@ahardin-rh ptal,thanks.

ahardin-rh:
Thank you!

PR #3471, opened 2017-01-10
Bug 1374890 added Configuring a Registry Location section

bfallonf:
As per: https://bugzilla.redhat.com/show_bug.cgi?id=1374890

Will ask @nak3 for an ack in the BZ.

### DIFF
--- a/install_config/install/advanced_install.adoc
+++ b/install_config/install/advanced_install.adoc
@@ -340,6 +340,31 @@ default is
 If you alter this variable, ensure the host name is accessible via your router.
 |===
 
+[[advanced-install-configuring-registry-location]]
+=== Configuring a Registry Location
+
+If you are using an image registry other than the default at
+`registry.access.redhat.com`, specify the desired registry within the
+*_/etc/ansible/hosts_* file.
+
+----
+oreg_url=example.com/openshift3/ose-${component}:${version}
+openshift_examples_modify_imagestreams=true
+----
+
+.Registry Variables
+[options="header"]
+|===
+
+|Variable |Purpose
+|`*oreg_url*`
+|Set to the alternate image location. Necessary if you are not using the default registry at `registry.access.redhat.com`.
+
+|`*openshift_examples_modify_imagestreams*`
+|Set to `true` if pointing to a registry other than the default. Modifies the image stream location to the value of `*oreg_url*`.
+|===
+
+
 [[advanced-install-configuring-global-proxy]]
 === Configuring Global Proxy Options
 


install_config/install/advanced_install.adoc
```diff
@@ -335,6 +335,30 @@ default is
...
+.Registry Variables
+[options="header"]
+|===
+
+|Variable |Purpose
+|`*oreg_url*`
+|Set the alternate image location. Necessary if you are not using the default registry at `registry.access.redhat.com`.
+
+|`*openshift_examples_modify_imagestreams*`
+|Set to `true` if pointing to a registry other than the default.
```
nak3: I am wondering that it should mention that following two things:
1. `openshift_examples_modify_imagestreams` modifies the image stereams's configuration.
2. It modifies the registry as pointed by `oreg_url`.

So, as [upstream's example](https://github.com/openshift/openshift-ansible/blob/master/inventory/byo/hosts.ose.example#L94-L95) wrote, how about "If oreg_url points to a registry other than registry.access.redhat.com we can modify image streams to point at that registry"

bfallonf:
Thanks @nak3 . I've updated the text to what you suggested. Let me know if you would like to change more.

@adellape @ahardin-rh @gaurav-nelson @bmcelvee Available for peer review thanks!

nak3:
No problem. LGTM. Thank you, @bfallonf

install_config/install/advanced_install.adoc
```diff
@@ -340,6 +340,31 @@ default is
 If you alter this variable, ensure the host name is accessible via your router.
 |===
 
+[[advanced-install-configuring-registry-location]]
+=== Configuring a Registry Location
+
+If you are using an image registry other than the default at
+`registry.access.redhat.com`, specify the desired registry within the
```
ahardin-rh: Should this be formatted as *_registry.access.redhat.com_* since it is technically a file path in that it is a URL? I would have to see how we treat that elsewhere in the docs. I could be wrong.

install_config/install/advanced_install.adoc
```diff
@@ -340,6 +340,31 @@ default is
...
+openshift_examples_modify_imagestreams=true
+----
+
+.Registry Variables
+[options="header"]
+|===
+
+|Variable |Purpose
+|`*oreg_url*`
+|Set to the alternate image location. Necessary if you are not using the default registry at `registry.access.redhat.com`.
```
ahardin-rh: same here

ahardin-rh:
@bfallonf Just one question/nit from me. Other than that, LGTM!

bfallonf:
@ahardin-rh I think this one might be ok, actually. It's what you set the variable to, so it's a value. It's a file path, sure, but it's a value in the example.

But, thanks for the eyes!

bfallonf:
[rev_history]
|xref:../install_config/install/advanced_install.adoc#install-config-install-advanced-install[Installing a Cluster -> Advanced Install]
|Added the xref:../install_config/install/advanced_install.adoc#advanced-install-configuring-registry-location[Configuring a Registry Location] section.
%

PR #3474, opened 2017-01-11
fix the high availble link in dev_guide part

### LINKED ISSUE #3687: highly available link response 404 Not Found in Using a Service ExternalIP session
Using a Service ExternalIP
Administrators can assign a list of externalIPs, for which nodes in the cluster will also accept traffic for the service. These IPs are not managed by OpenShift Origin and administrators are responsible for ensuring that traffic arrives at a node with this IP. A common example of this is configuing a **_highly available_** service.

https://docs.openshift.org/latest/dev_guide/getting_traffic_into_cluster.html#using-externalIP
https://docs.openshift.org/admin_guide/high_availability.html#admin-guide-high-availability

guangxuli:
fix : https://github.com/openshift/openshift-docs/issues/3687


guangxuli:
@ahardin-rh ptal,thanks.

PR #3475, opened 2017-01-11
Bug 1411421 added information on linking pods to serviceaccounts

bfallonf:
As per: https://bugzilla.redhat.com/show_bug.cgi?id=1411421

@jim-minter I'll ask again in the BZ, but can I get an ack this fulfills the request? I am wondering if I've gotten some of the wording wrong. It's hard to tell with the info that's there...

Thanks!

### DIFF
--- a/dev_guide/builds.adoc
+++ b/dev_guide/builds.adoc
@@ -947,6 +947,14 @@ source secrets used. Access is granted with the following command:
 $ oc secrets link builder mysecret
 ----
 
+[NOTE]
+====
+Limiting secrets to only the service accounts that reference them is disabled by
+default. This means that if `serviceAccountConfig.limitSecretReferences` is set
+to `false` (the default setting) in the master configuration file, linking
+secrets to a service is not required.
+====
+
 [[automatic-addition-of-a-source-secret-to-a-build-configuration]]
 ===== Automatic Addition of a Source Secret to a Build Configuration
 
--- a/dev_guide/service_accounts.adoc
+++ b/dev_guide/service_accounts.adoc
@@ -170,6 +170,17 @@ To allow a secret to be mounted by a service account's pods, run:
 $ oc secrets link --for=mount <serviceaccount-name> <secret-name>
 ----
 
+[NOTE]
+====
+Limiting secrets to only the service accounts that reference them is disabled by
+default. This means that if `serviceAccountConfig.limitSecretReferences` is set
+to `false` (the default setting) in the master configuration file, mounting
+secrets to a service account's pods with the `--for=mount` option is not
+required. However, using the `--for=pull` option to enable using an image pull
+secret is required, regardless of the
+`serviceAccountConfig.limitSecretReferences` value.
+====
+
 This example creates and adds secrets to a service account:
 
 ====
--- a/install_config/registry/securing_and_exposing_registry.adoc
+++ b/install_config/registry/securing_and_exposing_registry.adoc
@@ -76,6 +76,14 @@ $ oc secrets link registry registry-secret
 $ oc secrets link default  registry-secret
 ----
 +
+[NOTE]
+====
+Limiting secrets to only the service accounts that reference them is disabled by
+default. This means that if `serviceAccountConfig.limitSecretReferences` is set
+to `false` (the default setting) in the master configuration file, linking
+secrets to a service is not required.
+====
++
 . Add the secret volume to the registry deployment configuration:
 +
 ----


dev_guide/builds.adoc
```diff
@@ -947,6 +947,15 @@ source secrets used. Access is granted with the following command:
 $ oc secrets link builder mysecret
 ----
 
+[NOTE]
+====
+As of {product-title} version 3.0.2, limiting secrets to only the service
```
jim-minter: ("3.0.2" only applies when {product-title} is OCP)

dev_guide/builds.adoc
```diff
@@ -947,6 +947,15 @@ source secrets used. Access is granted with the following command:
...
+accounts that reference them is disabled by default. This means that unless
+`serviceAccountConfig.limitSecretReferences` is set to `true` in the master
+configuration file, linking secrets to a service account will do
+nothing.
```
jim-minter: "is not required" may be better than "will do nothing"

dev_guide/builds.adoc
jim-minter: slightly clumsy wording, maybe just easier to say something along the lines of: if limitSecretReferences is false (default), this step is not required ?

dev_guide/managing_images.adoc
```diff
@@ -477,8 +477,9 @@ $ oc secrets new-dockercfg <pull_secret_name> \
     --docker-password=<password> --docker-email=<email>
 ----
 
-To use a secret for pulling images for pods, you must add the secret to your
-service account. The name of the service account in this example should match
+If `serviceAccountConfig.limitSecretReferences` is set to `true` in the master
```
jim-minter: I believe this paragraph change is wrong and should be reverted: --for=pull and serviceAccountConfig.limitSecretReferences are separate things.

dev_guide/service_accounts.adoc
```diff
@@ -170,6 +170,15 @@ To allow a secret to be mounted by a service account's pods, run:
 $ oc secrets link --for=mount <serviceaccount-name> <secret-name>
 ----
 
+[NOTE]
+====
+As of {product-title} version 3.0.2, limiting secrets to only the service
```
jim-minter: same notes as above apply

install_config/registry/securing_and_exposing_registry.adoc
```diff
@@ -76,6 +76,15 @@ $ oc secrets link registry registry-secret
 $ oc secrets link default  registry-secret
 ----
 +
+[NOTE]
+====
+As of {product-title} version 3.0.2, limiting secrets to only the service
```
jim-minter: same notes as above apply

dev_guide/service_accounts.adoc
```diff
@@ -170,6 +170,15 @@ To allow a secret to be mounted by a service account's pods, run:
 $ oc secrets link --for=mount <serviceaccount-name> <secret-name>
 ----
 
+[NOTE]
```
jim-minter: here you follow a section discussing --for=pull and --for=mount; it may be good to be clearer that the NOTE refers to --for=mount (which is also implied if no --for= is specified) only.

jim-minter (changes requested):
some suggestions; the change to dev_guide/managing_images.adoc needs to be reverted. Thanks!

dev_guide/managing_images.adoc
bfallonf: @jim-minter Sorry, maybe I'm missing something. Is that not what you meant in the BZ:

"https://docs.openshift.com/container-platform/3.3/dev_guide/managing_images.html / Allowing Pods to Reference Images from Other Secured Registries
- the advice about 'oc secrets link' is superfluous unless serviceAccountConfig.limitSecretReferences is on, which it is probably not in the majority of installs."

Am I missing something obvious?

dev_guide/service_accounts.adoc
bfallonf: @jim-minter I see what you mean, which is why I made sure the last part of the note box says "...mounting secrets to a service account's pods is not required." Or did you mean that it should be more obvious?

dev_guide/managing_images.adoc
jim-minter: The BZ wasn't desperately clear (sorry), but did mention not to review the 'oc secrets link --for=pull' case(s), of which this is one. serviceAccountConfig.limitSecretReferences being enabled affects the requirement to run 'oc secrets link' or 'oc secrets link --for=mount', but not 'oc secrets link --for=pull'.

dev_guide/service_accounts.adoc
jim-minter: Ah, I hadn't spotted the wording change (so IMO it's not clear enough ;).  Now I re-read it, I'm concerned that the following statement is actually not true and doesn't make sense:

> unless `serviceAccountConfig.limitSecretReferences` is set to `true` in the master configuration file, mounting secrets to a service account's pods will do nothing.

Something like the following would make sense:

if the admin has set limitSecretReferences to false (default), it is not required to use oc link --for=mount to enable use of a mountable secret by a service account.  However oc link --for=pull is always required to enable use of an image pull secret, regardless of the value of limitSecretReferences.

dev_guide/managing_images.adoc
bfallonf: @jim-minter OK. I'm still a bit confused why this was mentioned in the initial BZ, but I've reverted it. Let me know if this still is not what you're requesting.

dev_guide/service_accounts.adoc
bfallonf: OK. I combined the two. Should be better now.

bfallonf:
@gaurav-nelson @ahardin-rh @bmcelvee This should be ok for peer review now :star:

dev_guide/builds.adoc
```diff
@@ -947,6 +947,14 @@ source secrets used. Access is granted with the following command:
 $ oc secrets link builder mysecret
 ----
 
+[NOTE]
+====
+Limiting secrets to only the service accounts that reference them is disabled by
+default. This means that if `serviceAccountConfig.limitSecretReferences` is set
+to `false` (the default setting)) in the master configuration file, linking
```
jim-minter: double ))

install_config/registry/securing_and_exposing_registry.adoc
```diff
@@ -76,6 +76,14 @@ $ oc secrets link registry registry-secret
 $ oc secrets link default  registry-secret
 ----
 +
+[NOTE]
+====
+Limiting secrets to only the service accounts that reference them is disabled by
+default. This means that if `serviceAccountConfig.limitSecretReferences` is set
+to `false` (the default setting)) in the master configuration file, linking
```
jim-minter: double ))

jim-minter:
apart from the double ))s, lgtm, thanks

dev_guide/builds.adoc
```diff
@@ -947,6 +947,14 @@ source secrets used. Access is granted with the following command:
 $ oc secrets link builder mysecret
 ----
 
+[NOTE]
+====
+Limiting secrets to only the service accounts that reference them is disabled by
```
gaurav-nelson: to only the → only to the

dev_guide/service_accounts.adoc
```diff
@@ -170,6 +170,17 @@ To allow a secret to be mounted by a service account's pods, run:
...
+====
+Limiting secrets to only the service accounts that reference them is disabled by
```
gaurav-nelson: same as above for all instances

gaurav-nelson:
minor change, everything else is fine. :ballot_box_with_check:

bfallonf:
@gaurav-nelson As discussed face-to-face, Going to leave it as is.

If there's nothing else, I'll merge this.

bfallonf:
[rev_history]
|xref:../dev_guide/builds.adoc#dev-guide-builds[Builds]
|Added note box about mounting secrets to service accounts.
%
[rev_history]
|xref:../dev_guide/service_accounts.adoc#dev-guide-service-accounts[Service Accounts]
|Added note box about mounting secrets to service accounts.
%
[rev_history]
|xref:../install_config/registry/securing_and_exposing_registry.adoc#install-config-registry-securing-exposing[Setting up the Registry -> Securing and Exposing the Registry]
|Added note box about mounting secrets to service accounts.
%

PR #3476, opened 2017-01-11
Fix command to create new service account.

bachp:
There was a "s" too much that caused the command to create
a normal user instead of a service account.


ahardin-rh:
Thank you!

ahardin-rh:
Minor typo. No revision history needed.

PR #3480, opened 2017-01-12
clean up build admission config examples

bparees:


bparees:

jeremyeder:
Thanks to Ben, this is what worked for me:

Taint a node:
```oadm taint nodes <node_name_from_oc_get_nodes> builder=true:NoScheduleNoAdmitNoExecute```

Set up the builddefaults:
```
# cat /etc/origin/master/master-config.yaml|head
admissionConfig:
  pluginConfig:
    BuildDefaults:
      configuration:
        apiVersion: v1
        kind: BuildDefaultsConfig
        nodeSelector:
          kubernetes.io/hostname: <node_name_from_oc_get_nodes>
        annotations:
          scheduler.alpha.kubernetes.io/tolerations: '[{"key":"builder", "value":"true"}]'
```

Restart the master:
```
Look for log message like:
NodeSelector:map[string]string{"kubernetes.io/hostname":"ip-xxx-xxx-xxx-xxx.us-west-2.compute.internal"}, Annotations:map[string]string{"scheduler.alpha.kubernetes.io/tolerations":"[{\"key\":\"builder\", \"value\":\"true\"}]"}}
```
Launch a build, and describe the build pod:
```
# oc describe pod cakephp-mysql-example-1-build -n cakephp-mysql0|grep Toleration
Tolerations:    builder=true
```

aveshagarwal:
@jeremyeder  I haven't looked into it in detail, but NoExecute is not implemented yet.

jeremyeder:
Sorry, I meant NoSchedule.  I tested with NoSchedule.

aveshagarwal:
@jeremyeder lgtm. I am guessing that the use case here is that we want only build pods to schedule on these builder nodes. We should make sure that the nodes were registered/initialized with this taint "builder=true:NoSchedule." Then, NoSchedule is the right taint in this scenario.

bfallonf:
@bparees Can i ask why this was labeled for both 3.4 and 3.5? I don't think the 3.4 docs branch has been cut yet, and if that's the case, then there's no reason for the 3.5 label. Just wondering if there's a reason I'm not aware of...

bfallonf:
@bparees no problems! I've taken off the 3.5 label. It'll go in with 3.4, so it'll be in 3.5.

Is this ok to merge, then?

bparees:
@bfallonf yup.

bfallonf:
Thanks @bparees . I'll merge.

I don't think a rev history is needed.

jeremyeder:
Taints are still alpha in 3.4.

bparees:
> Taints are still alpha in 3.4.

yeah that's why i'm still holding off actually putting them into the doc specifically, and just discussing the generic capability to add annotations.

bparees:
@openshift/team-documentation this change does not seem to have made it into the OCP docs?  Tracking in bug https://bugzilla.redhat.com/show_bug.cgi?id=1466153 now.

vikram-redhat:
@bfallonf - can you take a look?

bfallonf:
@bparees @vikram-redhat This is in the 3.4 docs, but not 3.5. I have no idea how this happened. Only thing I can think of is that there was a conflict when the 3.5 branch was cut and resolved wrong. Vikram, not sure what you want me to do with this one. I'll chat to you today.

bparees:
@bfallonf @vikram-redhat thanks, let's make sure it's right in 3.6 too.

vikram-redhat:
Thanks @bfallonf and @bparees . Looks like this PR (https://github.com/openshift/openshift-docs/pull/4149/files#diff-e4021ff92d7da26d0039f3305a8c0973) for OCP 3.5 release didn't conficted and didn't pick the right fix. 

@bfallonf  - can you do a follow up to fix this as part of the weekly release? Since 3.6 is not out yet, a follow up fix will make sure that it lands in 3.5 as well.

vikram-redhat:
@bparees @bfallonf this is now pushed to both portal and docs.openshift.com (https://docs.openshift.com/container-platform/3.5/install_config/build_defaults_overrides.html#manually-setting-global-build-defaults). @bfallonf - can you link to the docs and close the bug?

bfallonf:
Thanks @vikram-redhat . Here's the links:

https://access.redhat.com/documentation/en-us/openshift_container_platform/3.5/html-single/installation_and_configuration/#manually-setting-global-build-defaults

https://access.redhat.com/documentation/en-us/openshift_container_platform/3.5/html-single/installation_and_configuration/#manually-setting-global-build-overrides

I'll put these into the BZ and close it. @bparees , please let us know if this requires any more thought.

PR #3492, opened 2017-01-13
Adding back in Azure content for 3.4

ahardin-rh:
Previously mistakenly introduced content into 3.3  via https://github.com/openshift/openshift-docs/pull/3300 and was pulled down via https://github.com/openshift/openshift-docs/pull/3491 and https://github.com/openshift/openshift-docs/pull/3489.

This is adding back topics to 3.4. Some other content in https://github.com/openshift/openshift-docs/pull/3300 related to dynamic provisioning of Azure can't be added in until 3.5. That will be reintroduced in a separate PR.

PR #3371, opened 2016-12-13
Update glusterfs dynamic provisioning doc.

humblec:
Signed-off-by: Humble Chirammal <hchiramm@redhat.com>

PR #3316, opened 2016-12-02
Adding Back Storage Classes file and Fix Links to Dynamic Provisioning

pospispa:
@screeley44 @ahardin-rh PTAL

PR #2972, opened 2016-09-30
Add StorageClass doc under persistent storage

screeley44:
Starting to put this together for OCP 3.4 - any input is appreciated, not sure I like the flow of this, especially with the already existing Dynamic Provisioning topic - at the least we should probably somehow tie the Dynamic Provisioning to this StorageClass doc

@erinboyd @eparis @childsb

PR #3493, opened 2017-01-13
Adding in Azure dynamic provisioning content for 3.5

ahardin-rh:
Previously mistakenly introduced content into 3.3 via https://github.com/openshift/openshift-docs/pull/3300 and was pulled down via https://github.com/openshift/openshift-docs/pull/3491 and https://github.com/openshift/openshift-docs/pull/3489.

Content specific to 3.4 was added back in via https://github.com/openshift/openshift-docs/pull/3492. This content is related to dynamic provisioning of Azure, which can't be added in until 3.5.

### DIFF
--- a/install_config/persistent_storage/dynamically_provisioning_pvs.adoc
+++ b/install_config/persistent_storage/dynamically_provisioning_pvs.adoc
@@ -74,6 +74,10 @@ the master node. To do this, set the
 |xref:../../install_config/configuring_openstack.adoc#install-config-configuring-openstack[Configuring for OpenStack]
 |
 
+|Azure Disk
+|`kubernetes.io/azure-dd`
+|xref:../../install_config/configuring_azure.adoc#install-config-configuring-azure[Configuring for Azure]
+
 |===
 
 
@@ -124,7 +128,7 @@ from plug-in to plug-in.
 [[storage-class-annotations]]
 === StorageClass Annotations
 
-To set a _StorageClass_ as the cluster-wide default: 
+To set a _StorageClass_ as the cluster-wide default:
 ----
    storageclass.beta.kubernetes.io/is-default-class: "true"
 ----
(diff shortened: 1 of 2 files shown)


adellape (approved):
:+1:

vikram-redhat:
@ahardin-rh @adellape  - can you please check if this content should be in 3.4? By the looks of it, it has come back in, but am not really sure which parts, given how many times this has moved around: https://docs.openshift.com/container-platform/3.4/install_config/persistent_storage/dynamically_provisioning_pvs.html#azure-disk

adellape:
@vikram-redhat Ashley had done a set of PRs to tidy up the Azure 3.4 vs 3.5 stuff ahead of release, but it looks like it was reintroduced into 3.4 at the last minute here:

https://github.com/openshift/openshift-docs/pull/3517

I now suspect that the cause for the above issues was probably related to the following early storageclasses-related PR never getting its `to_followup` label removed (and still hasn't):

https://github.com/openshift/openshift-docs/pull/2972#issuecomment-266094524

Also this one (my fault):

https://github.com/openshift/openshift-docs/pull/3316

I don't think we did a very good job of double-checking our `to_followup` queue ahead of 3.4 publish.

ahardin-rh:
@vikram-redhat @adellape The Azure dynamic provisioning content is now removed from 3.4 via https://github.com/openshift/openshift-docs/pull/3541 and https://github.com/openshift/openshift-docs/pull/3542

ahardin-rh:
@vikram-redhat @adellape https://github.com/openshift/openshift-docs/pull/3371 regarding the GlusterFS example was marked with a to-followup label and pulled into the 3.4 GA release somehow, even though it was still flagged for follow-up. https://github.com/openshift/openshift-docs/pull/3543 takes care of some strange formatting introduced in there and satisfies the to-followup.

PR #3466, opened 2017-01-09
document the supports annotation on imagestreams

bparees:

PR #3335, opened 2016-12-06
Add an image stream authoring guide for best end user experience

jwforres:
Implements https://trello.com/c/LBm5DB8b

image-stream-docs-part1

continues from above...

image-stream-docs-part2

PR #3495, opened 2017-01-15
OCP 3.2: Weekly Release: 16/Jan/2017

vikram-redhat:


dev_guide/managing_images.adoc
```diff
@@ -111,7 +111,7 @@ xref:../install_config/install/docker_registry.adoc#registry-known-issues[known
 issue]
 endif::[]
 ifdef::openshift-online,openshift-dedicated,atomic-registry[]
-https://docs.openshift.com/enterprise/3.2/release_notes/ose_3_2_release_notes.html#ose-32-known-issues[known issue]
+https://docs.openshift.com/enterprise/3.2/release_notes/ose_3_2_release_notes.adoc#ose-32-known-issues[known issue]
```
adellape: @vikram-redhat Not that it matters here since this is the enterprise-3.2 repo, but in this case this is link to an external URL, so shoulda stayed `.html`.

dev_guide/managing_images.adoc
```diff
@@ -777,3 +777,68 @@ NAME       DOCKER REPO                        TAGS      UPDATED
 my-image   172.30.48.125:5000/test/my-image   latest    3 seconds ago
 ----
 ====
+[[writing-image-streams-for-s2i-builders]]
+== Writing Image Streams for S2I Builders
```
adellape: @vikram-redhat Looks like this section got added accidentally while trying to pick https://github.com/openshift/openshift-docs/pull/3466 to enterprise-3.2. The actual PR this came from (https://github.com/openshift/openshift-docs/pull/3335) doesn't list enterprise-3.2 as a label so I'm gonna yoink this from 3.2.

adellape: Oh, the main reason I was looking at this was because it was breaking the formatting in the Customer Portal version, since the section title was too close to the code block above it.

PR #3515, opened 2017-01-18
OCP 3.4: Link Fixes: 9

vikram-redhat:


adellape:
I don't get this one? It also was now causing a 404/CP build failure after `cron_job.adoc` was removed, so I've reverted this via https://github.com/openshift/openshift-docs/pull/3591.

PR #3529, opened 2017-01-18
Deploy metrics using ansible: https://trello.com/c/piaWOfvB/

jcantrill:
This PR is a first attempt to document:

* deploying metrics using the ansible role 'openshift_metrics'

cc @sdodson @mwringe @bbguimaraes @bmcelvee 

I think there are additional changes that will be necessary in 'install/advanced_install.adoc' regarding nfs configuration for storage, but I do not think we have moved the 'hosted_metrics' vars to have an equivalent in the new role.  Not really certain we need to.

### DIFF
--- a/install_config/cluster_metrics.adoc
+++ b/install_config/cluster_metrics.adoc
@@ -105,7 +105,7 @@ to configure persistent storage for your metrics pods.
 The size requirement of the Cassandra storage is dependent on the number of
 pods. It is the administrator's responsibility to ensure that the size
 requirements are sufficient for their setup and to monitor usage to ensure that
-the disk does not become full.  The size of the persisted volume is specified with the `*openshift_metrics_cassandra_pv_size*`
+the disk does not become full.  The size of the persisted volume claim is specified with the `*openshift_metrics_cassandra_pvc_size*`
 xref:../install_config/cluster_metrics.adoc#metrics-ansible-variables[ansible variable] which is set to 10 GB by default.
 
 If you would like to use xref:../install_config/persistent_storage/dynamically_provisioning_pvs.adoc#install-config-persistent-storage-dynamically-provisioning-pvs[dynamically provisioned persistent volumes]
@@ -324,12 +324,12 @@ be added to your inventory file if it is necessary to override them.
 | The frequency that metrics are gathered defined as a number and time identifier(s,m,h)
 (e.g. 15s for 15 seconds).
 
-|`*openshift_metrics_cassandra_pv_prefix*`
+|`*openshift_metrics_cassandra_pvc_prefix*`
 |The persistent volume claim prefix created for Cassandra. A serial number is appended to the prefix starting
   from 1.
 
-|`*openshift_metrics_cassandra_pv_size*`
-| The persistent volume size for each of the Cassandra  nodes.
+|`*openshift_metrics_cassandra_cpvc_size*`
+| The persistent volume claim size for each of the Cassandra  nodes.
 
 |`*openshift_metrics_cassandra_storage_type*`
 | Use `emptydir` for ephemeral storage (for  testing), `pv` to use persistent volumes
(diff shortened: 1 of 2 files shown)


install_config/cluster_metrics.adoc
```diff
@@ -194,20 +129,23 @@ After the metrics deployer runs, the output of `oc get pods` should resemble the
...
 
-By default, metrics data is stored for 7 days. You can configure this with the `*METRIC_DURATION*` parameter in the
+By default, metrics data is stored for 7 days. You can configure this by setting the `*openshift_metrics_duration*`
+variable in your hosts file which will update the upon deployment
 ifdef::openshift-origin[]
-link:https://github.com/openshift/origin-metrics/blob/master/metrics.yaml[*_metrics.yaml_* configuration file]. 
+link:https://github.com/openshift/origin-metrics/blob/master/metrics.yaml[*_metrics.yaml_* configuration file].
 endif::openshift-origin[]
 ifdef::openshift-enterprise[]
 *_metrics.yaml_* configuration file.
```
bbguimaraes: We should remove part about `metrics.yaml`, that was specific for the deployer pod.

install_config/cluster_metrics.adoc
```diff
@@ -331,217 +268,153 @@ service.
...
 
 [[metrics-using-secrets-byo-certs]]
 ==== Providing Your Own Certificates
 
 To provide your own certificate which will be used by the
 xref:../architecture/core_concepts/routes.adoc#secured-routes[re-encrypting route],
-you can pass these values as
-xref:../dev_guide/secrets.adoc#dev-guide-secrets[secrets] to the Metrics Deployer.
+you can set the 'cert, key, and ca' xref:../install_config/cluster_metrics.adoc#metrics-ansible-variables[variables]
+in your inventory file.
```
bbguimaraes: It's probably better to spell out the full variable names here: `openshift_metrics_hawkular_ca`, `openshift_metrics_hawkular_key` and `openshift_metrics_hawkular_ca`.

install_config/cluster_metrics.adoc
```diff
@@ -331,217 +268,153 @@ service.
...
 
 The `hawkular-metrics.pem` value needs to contain the certificate in its *_.pem_*
 format. You may also need to provide the certificate for the Certificate Authority
 which signed this *_pem_* file via the `hawkular-metrics-ca.cert` secret.
```
bbguimaraes: These are the files that would go into the secret for the deployer, no longer relevant.

bmcelvee:
@jcantrill Looking at Trello, it looks like this is targeted for 3.5. Is that correct?

jcantrill:
@bmcelvee  that is correct.

install_config/cluster_metrics.adoc
```diff
@@ -583,9 +453,10 @@ The following command sets the Hawkular Metrics route to use
...
 
 ----
-$ oc new-app --as=system:serviceaccount:openshift-infra:metrics-deployer \
-    -f metrics-deployer.yaml \
-    -p HAWKULAR_METRICS_HOSTNAME=hawkular-metrics.example.com
+$ ansible-playbook <OPENSHIFT_ANSIBLE_DIR>/common/openshift-cluster/openshift_metrics.yml \
+   -e openshift_metrics_install_metrics=True \
+   -e openshift_metrics_hawkular_hostname=hawkular-metrics.example.com \
+   -e openshift_metrics_cassandra_storage_type=pv
 ----
```
mwringe: For origin we need to provide a link to where we can download the file if OpenShift ansible is not installed (which is a valid use case for origin installations).

For OCP, the location should be known before hand, why its it <OPENSHIFT_ANSIBLE_DIR>? or is this a resolved to something during the build?

mwringe:
The documentation here makes a lot of assumptions about the user already having ansible installed and setup on there system. 

Ansible being used should be assumed when installing OCP (https://docs.openshift.com/container-platform/3.4/install_config/install/advanced_install.html) but it is not the case when installing Origin (https://docs.openshift.org/latest/getting_started/administrators.html#getting-started-administrators).

Trying to follow the docs when ansible was not used to install the OpenShift cluster doesn't exactly work. I think we need an Origin only section which outlines any additional steps you may need to do if ansible was not used to installed OpenShift.

jcantrill:
@bmcelvee We may wish to change this PR such that we fully display the old version for 'enterprise' and the 'new version' for origin since the deployment method is significant enought

bmcelvee:
@jcantrill Thanks for the heads up. I'll look into the best way to make it happen.

bmcelvee:
Continuing edits in https://github.com/openshift/openshift-docs/pull/3765/files.

PR #3484, opened 2017-01-12
Added Storage Classes topic for Online for BZ#1411222

ahardin-rh:
Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1411222

This topic is for Online only to prevent content duplication. Other Storage Classes content lives in the Installation and Configuration section of the docs, but Installation and Configuration does not appear in Online documentation.

PR #3537, opened 2017-01-19
Added topic on Selector-Label Volume Binding for Online only

ahardin-rh:


ahardin-rh:
@damemi This is the new topic for Online only per our discussion. PTAL.

ahardin-rh:
Path will be https://docs.openshift.com/online/dev_guide/selector_label_volume_binding.html

ahardin-rh:
@abhgupta @damemi Is the content for this page okay? Thanks!

abhgupta:
Is this even applicable for Online?

damemi:
@abhgupta it's similar to the storage classes doc page... not really relevant for Online but a broken link otherwise

ahardin-rh:
@vikram-redhat @aheslin Can we discuss the best course of action for this? As it turns out, this content is actually not relevant for Online, but there is a need to direct users to a page from the UI and preserve some links there. @damemi could probably provide more context on the use case and the need, but now I do not think that this page should be built if it is not relevant to our audience. We recently built a similar page due to UI needs on Storage Classes for Online only via https://github.com/openshift/openshift-docs/pull/3484. We built out that page because the Installation and Configuration guide (where a Storage Class topic exists) does not build in Online, but in this case the content in that PR was actually relevant for Online. I discussed this with the team today and it is uncertain whether there will be need to build similar pages in the future.

vikram-redhat:
Hey @ahardin-rh and @damemi - I am slightly confused. Please correct any assumptions I may have made here:

1. We have already added a link in online UI to a page in the docs about Selector-Label Volume Binding.
2. The link should lead to this content in the online only branch.
3. The content isn't really relevant, but if we don't provide this link, anyone using the UI who clicks the link will get a 404 error.
4. So users don't get the 404 error, we should add *some* content, even though, users clicking the link have no need for that information.

If all this sounds correct, I think the best course of action, from both a customers and docs POV, would be to remove the link from the online UI rather than apply a band-aid solution to the docs.

@aheslin

abhgupta:
@vikram-redhat Your understanding is spot on. 

How about we come up with a standard single page inside the online documentation section, that simply says, "This feature is not available for OpenShift Online"?

We could reference that page from all such links. Would be nice if we didn't have to list that page in the index.

vikram-redhat:
@abhgupta - understood. I think your suggestion is great. However, it still puts the onus of having something on the docs-set. Is it not possible to remove those links to the documentation from the Online UI? What am I missing here?

abhgupta:
@vikram-redhat The links are part of the console and something we cannot customize. We do have the ability to customize the target URLs, which we are trying do here. 

OpenShift Online uses OCP builds and hence we cannot modify the console source code to remove these links. Reusing the same builds make a HUGE simplification and is a MAJOR focus for V3 - our ability to reuse OCP builds and manage customizations on top of it or outside it.

vikram-redhat:
@abhgupta - Ah I understand now. :+1: 

I think your suggestion to have a generic "This feature is not available in OpenShift Online" page sounds good to me then. I will wait for @aheslin to reply and confirm.

@ahardin-rh - can you also ping @aheslin on Monday?

ahardin-rh:
@vikram-redhat where would the generic page be built and is there a way that we can keep it hidden from navigation, or even search? I will discuss our options with @aheslin

vikram-redhat:
Sorry for the delay @ahardin-rh . @aheslin  is reaching out to engineering for more info.

aheslin:
As discussed in the program meeting, we can merge this docs PR. This is a documentation workaround for an issue in the console and it should be fixed to improve the user experience in the short to mid term.

ahardin-rh:
No revision history needed

PR #3546, opened 2017-01-20
Make gluster endpoints and service the same name

eparis:
This prevents the endpoints from being automatically deleted

PR #3547, opened 2017-01-20
Make gluster endpoints and service the same name

eparis:
This prevents the endpoints from being automatically deleted

### DIFF
--- a/install_config/storage_examples/gluster_example.adoc
+++ b/install_config/storage_examples/gluster_example.adoc
@@ -58,7 +58,7 @@ The named endpoints define each node in the Gluster-trusted storage pool:
 apiVersion: v1
 kind: Endpoints
 metadata:
-  name: gluster-endpoints <1>
+  name: gluster-cluster <1>
 subsets:
 - addresses:              <2>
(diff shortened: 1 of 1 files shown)


humblec:
LGTM. Thanks!

nak3:
May I ask you to modify the notes in the line:

> <1> The name of the endpoints is used in the PV definition below.

and

> <1> The name of the service.

to like "This name must match the {service,endpoints}". [This section](https://docs.openshift.com/container-platform/3.3/install_config/persistent_storage/persistent_storage_glusterfs.html#creating-gluster-endpoints) has the notes and it was really helpful.

ahardin-rh:
Thanks!

@nak3 I will follow up with that edit, as well as some other edits, in a separate PR while I continue to address https://bugzilla.redhat.com/show_bug.cgi?id=1408293

ahardin-rh:
These changes were applied to 3.3 and 3.4 via:
https://github.com/openshift/openshift-docs/pull/3545
https://github.com/openshift/openshift-docs/pull/3546

PR #3560, opened 2017-01-23
Weekly Release - Online - Mon Jan 23 2017

ahardin-rh:


vikram-redhat:
As discussed @ahardin-rh - commits look good so merging it.

PR #3561, opened 2017-01-24
Bug 1406755 added note box checking the right repos are installed

bfallonf:
As per: https://bugzilla.redhat.com/show_bug.cgi?id=1406755

I'll ask for an ack in the BZ.

### DIFF
--- a/install_config/install/host_preparation.adoc
+++ b/install_config/install/host_preparation.adoc
@@ -63,16 +63,43 @@ packages.
 # subscription-manager attach --pool=<pool_id>
 ----
 
-. Disable all repositories and enable only the required ones:
+. Disable all yum repositories:
+.. Disable all the enabled RHSM repositories:
 +
 ----
 # subscription-manager repos --disable="*"
+----
+
+.. List the remaining yum repositories and note their names under `repo id`, if any:
++
+----
+# yum repolist
+----
+
+.. Use `yum-config-manager` to disable the remaining yum repositories:
++
+----
+# yum-config-manager --disable <repo_id>
+----
++
+Alternatively, disable all repositories:
++
+----
+ yum-config-manager --disable \*
+----
++
+Note that this could take a few minutes if you have a large number of available repositories
+
+. Enable only the repositories required by {product-title} 3.4:
++
+----
 # subscription-manager repos \
     --enable="rhel-7-server-rpms" \
     --enable="rhel-7-server-extras-rpms" \
     --enable="rhel-7-server-ose-3.4-rpms"
 ----
 endif::[]
+
 
 == Installing Base Packages
 


install_config/install/host_preparation.adoc
```diff
@@ -72,6 +72,20 @@ packages.
     --enable="rhel-7-server-extras-rpms" \
     --enable="rhel-7-server-ose-3.4-rpms"
 ----
++
+[NOTE]
+====
+If the `subscription-manager repos --disable="*"` command does not disable all
+repositories, use `yum-config-manager --disablerepo=<repo_name>`. If the
+repositories are not disabled then enabled in the method above, your system
```
rhcarvalho: The part that says "not disabled then enabled in the method above" is unclear / confusing to me.
What does "method above" refer to, the four `subscription-manager ...` commands in the Host Registration section, or something else?
"not disabled then enabled", I didn't understand that.

rhcarvalho (commented):
@bfallonf thanks for the PR. As to the correctness of the commands I'd defer to someone with more `yum` experience, maybe @tdawson? He was the one helping me originally when I ran into the issue.

@tdawson would you have a better suggestion as to how to ensure only the proper repos are enabled when installing OCP?

tdawson:
I do not believe { subscription-manager repos --disable="*"` }should be used.  This command is inconsistent.  The reason can be found in the man page.

"This command (subscription-manager repos) is  only  used  for  organizations which have a locally-hosted content service of some kind, like Subscription Asset Manager.  With Red Hat's hosted content service, there  is  only  one central repository."

My recommendation is to use yum-config-manager to do those commands.
```
yum-config-manager --disable \*
yum-config-manager \
    --enable "rhel-7-server-rpms" \
    --enable "rhel-7-server-extras-rpms" \
    --enable "rhel-7-server-ose-3.3-rpms"
```

bfallonf:
@tdawson Hmm that's kind of a big change. We've used subscription-manager commands in the docs as long as I've worked on OpenShift. I'm wondering if there's a business case in using the Red Hat thing in subscription manager and it's causing problems. @adellape do you have any thoughts on this?

adellape:
@bfallonf @tdawson A blanket `yum-config-manager --disable \*` can take a really long time if you have a lot of available repos per your RHSM subscriptions, even if they're all disabled already. And the RHSM docs do suggest using `subscription-manager repos` in general, so it'd be good to align with that where we can (even though we know that `--disable` doesn't necessarily get _all_ yum repos):

https://access.redhat.com/documentation/en-US/Red_Hat_Subscription_Management/1/html-single/RHSM/index.html#entitlements-and-yum

I'd suggest breaking this up into two numbered steps: a multi-part step one for listing and disabling, and a step for enabling. Something like:

```
. Disable all yum repositories:

.. Disable all enabled RHSM repositories:
+
----
# subscription-manager repos --disable="*"
----

.. List all remaining yum repositories and note their names under `repo id`, if any:
+
----
# yum repolist
----

.. Disable the remaining yum repositories:
+
----
# yum-config-manager --disable <repo_id>
----
+
Alternatively, disable all repositories using `yum-config-manager`, though note that this could take a few minutes if you have a large number of available repositories:
+
----
 yum-config-manager --disable \*
----

. Enable only the repositories required by {product-title} 3.4:
+
----
# subscription-manager repos \
    --enable="rhel-7-server-rpms" \
    --enable="rhel-7-server-ose-3.4-rpms"
----
```

tdawson:
I like that.  :+1:

bfallonf:
@rhcarvalho @tdawson @adellape I've put in basically what Alex suggested. If there's any disagreements, please let me know.

Otherwise, @ahardin-rh @bmcelvee this is probably up for peer review.

Thanks, all.

ahardin-rh:
@bfallonf LGTM

bfallonf:
[rev_history]
|xref:../install_config/install/host_preparation.adoc#install-config-install-host-preparation[Installing a Cluster -> Host Preparation]
|Added steps on using `yum-config-manager` to the host registration steps.
%

PR #3563, opened 2017-01-24
Remove CronJobs which will available since 3.5

soltysh:
CronJob will be available only in 3.5, before the resource is called ScheduledJob. 
This solves: https://bugzilla.redhat.com/show_bug.cgi?id=1415916.


soltysh:
@vikram-redhat fyi, since I see you've added cronjobs.adoc

vikram-redhat:
@ahardin-rh - since 3.4 hasn't been done yet for this week, and this needs to go out now, I have added it directly to staging.

adellape:
@vikram-redhat This topic is showing up in search results still, which 404s, and needs a redirect set up.

PR #3564, opened 2017-01-24
Fix typo: 'specific file' vs 'literal value'

brennv:


bfallonf:
Thanks @brennv I'll merge.

I don't think a rev history is needed.

PR #3566, opened 2017-01-24
Adds port warning for Cockpit and the Registry Console

brennv:


bfallonf:
Thanks @brennv ! I'll do a quick follow-up and include the default port number.

PR #3572, opened 2017-01-25
SDN-related docs fixes

danwinship:
Was about to write some new SDN-related docs but needed to do some fixing first:

 1. We're currently referring to "OpenShift Container Platform SDN", etc, which just sounds terrible. It's "OpenShift SDN" regardless of which version of the software you're using
 2. The CNI-related changes in 3.4 made our network setup much simpler but we never updated the docs to reflect that. (Note: the dia/png change just removes the docker-managed containers from the diagram, since their networking is now handled entirely by docker, not by OpenShift.)
 3. We're constantly changing the exact details of the OVS flows, so we shouldn't be showing specific flows in the troubleshooting doc, or they'll always be wrong

@knobunc PTAL since you wrote a lot of this stuff

PR #3575, opened 2017-01-25
Document new multicast and NetworkPolicy stuff

danwinship:
This documents the new multicast and NetworkPolicy stuff. I wasn't quite sure how to refer to their "Tech Preview" status, so maybe that needs to be tweaked.

Also, there didn't seem to be any good place to put them, and I realized that we were kind of abusing managing_pods.adoc for a lot of not-especially-pod-related stuff, so I split all the networking stuff out of there into a new "managing_networking" file. It seems to work fine?

~~Built on top of #3572; only the last three commits are new here~~


danwinship:
rebased around the egress doc updates.
poke poke, this is supposed to be in 3.5

bfallonf:
@danwinship Thanks! I think this looks good overall.

Is this ok to merge? I might do that and do some followups for CCS style, etc.

bfallonf:
Actually, @danwinship this may have to be squashed before I can merge it.

danwinship:
yes, it's ready. squashed and repushed

bfallonf:
Thanks @danwinship . I'll merge this and do a quick followup.

vikram-redhat:
@bfallonf - when this got split, the anchors were duplicated. Please be sure to check for that when doing a follow up.

PR #3593, opened 2017-01-26
Expand API TOC

brennv:


ahardin-rh:
The REST API area of the documentation is auto-generated from the Origin code. A new guide that will provide end to end use cases for working with the OpenShift REST API is in the works. Thanks!

PR #3596, opened 2017-01-27
Add BestEffort definition

brennv:
Wording from https://kubernetes.io/docs/admin/resourcequota/walkthrough/

### DIFF
--- a/admin_guide/quota.adoc
+++ b/admin_guide/quota.adoc
@@ -138,7 +138,8 @@ the allowed set results in a validation error.
 
 |*BestEffort*
 |Match pods that have best effort quality of service for either `*cpu*` or
-`*memory*`.
+`*memory*`. See the xref:../admin_guide/overcommit.html#qos-classes[Quality of Service Classes] for more on
+committing compute resources.
 
 |*NotBestEffort*
 |Match pods that do not have best effort quality of service for `*cpu*` and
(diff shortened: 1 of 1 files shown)


admin_guide/quota.adoc
```diff
@@ -149,6 +149,10 @@ A *BestEffort* scope restricts a quota to limit the following resources:
 
 - `*pods*`
 
+*BestEffort* scoped pods can take advantage of slack compute resources in their
+namespace, and then require that users make an explicit resource request for
+pods that require a higher quality of service.
```
bparees: this still doesn't really tell me how i create a besteffort pod.  

The other page says that the BestEffort scope will "Match pods that have best effort quality of service for either cpu or memory."

I still don't understand how I set memory or cpu to "best effort"  (the reality, I believe, is that if you set a limit, but no request, for a given resource(eg memory or cpu) that is considered best effort for that resource, but I don't think we state that explicitly anywhere)

brennv: Good point, maybe we need a new section with examples drawing inspiration from https://github.com/kubernetes/community/blob/master/contributors/design-proposals/resource-qos.md#qos-classes

derekwaynecarr: A BestEffort pod can use slack compute resources that are unused on its scheduled node.

A BestEffort pod is a pod that does not have a compute resource requirement for cpu and memory.

derekwaynecarr (changes requested):

admin_guide/quota.adoc
derekwaynecarr: or just link to https://docs.openshift.org/latest/admin_guide/overcommit.html#qos-classes ?

ahardin-rh:
This is an old one, but is it about ready to go? Once this is updated, can you please give it a squash so that I can merge? Thanks!

ahardin-rh:
Continued work in https://github.com/openshift/openshift-docs/pull/5012

PR #3640, opened 2017-02-02
Note node disruption kills bare pods

mmahut:
Clarification about what happens when a pod that is not backed up by a replication controllers is terminated on a node.


pweil- (approved):
LGTM

ahardin-rh:
@mmahut Should this be applied to 3.3 and 3.4? Please tag @openshift/team-documentation before merging next time so that we can do an editorial review and ensure the PR has proper labels and milestones applied. Thanks so much!

PR #3641, opened 2017-02-02
Bug 1412827, Added more information about disabling MAC filtering for the Egress Router

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1412827

### DIFF
--- a/admin_guide/managing_pods.adoc
+++ b/admin_guide/managing_pods.adoc
@@ -158,18 +158,40 @@ specified remote server, using a private source IP address that is not used for
 anything else. The service allow pods to talk to servers that are set up
 to only allow access from whitelisted IP addresses.
 
-[IMPORTANT]
-====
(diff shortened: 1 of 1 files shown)


ahardin-rh:
@danwinship @pravisankar @sferich888 PTAL

danwinship:
I don't actually know the details of what needs to be done, but that all sounds like it's probably right.

Maybe add some generic text to the top of that section. Something like:

The egress router adds a second IP address and MAC address to the node's primary network interface. If you are not running {product-title} on bare metal, you may need to configure your hypervisor or cloud provider to allow the additional address.

ahardin-rh:
@danwinship Thanks. Updated.

danwinship:
lgtm

ahardin-rh:
[rev_history]
|Added more information about disabling MAC filtering for the Egress router.
%

PR #3691, opened 2017-02-07
Update registry cert secret name to match installer

codificat:
The installer uses `registry-certificates` as the secret's name for the certs when securing the registry (see openshift/openshift-ansible#2409).

This updates the docs to use the same name name for the Secret for consistency when this is performed manually.


ahardin-rh:
Sorry for the delay on this! Confirmed with @sdodson that this is still relevant.

ahardin-rh:
No revision history needed

PR #3699, opened 2017-02-08
Documented format for example URL paths

ahardin-rh:


ahardin-rh:
@adellape as discussed today :smile_cat:

adellape (approved):

PR #3709, opened 2017-02-10
Added manual upgrade step for 1.5/3.5: migrating jobs

soltysh:
This is required for 1.5/3.5.

@ahardin-rh ptal

@sdodson fyi


sdodson:
Should the installer just run this with --confirm or should a human be in the loop? Also, I know I asked if this was only 3.4 to 3.5, but we use the same playbooks for 3.4 to 3.5 as we do for 3.5.0 to 3.5.1. Is this safe to re-run?

soltysh:
@sdodson double yes, for `--confirm` (no need to human interaction) and for having it re-run (it's safe, even if run multiple times, it'll do migration just once and the following runs will be no-ops).

adellape:
Closing in favor of https://github.com/openshift/openshift-docs/pull/4091, where I've absorbed this content.

PR #3724, opened 2017-02-10
sudo with systemctl in registry quickstart

ktdreyer:
The other commands in this section use `sudo` - add it here as well, since this step requires root access.


ahardin-rh:
Thanks!

ahardin-rh:
No revision history needed. Atomic only.

PR #3729, opened 2017-02-13
Fixed mistake in quota.adoc

larkly:
requests.storage was entered twice. I assume the intent was to show it as limits.storage instead.


ahardin-rh:
Thank you!

PR #3741, opened 2017-02-13
add timing comment to registry quickstart

ktdreyer:
Users cannot jump immediately from starting the -master service in systemd to seeing running containers with "docker ps", because it takes a couple seconds for a host to pull down the images from Docker Hub.

Add a note to the docs that describes what is going on and how to troubleshoot.


ahardin-rh:
Apologies for the delay!

ahardin-rh:
No revision history. Atomic only.

PR #3742, opened 2017-02-13
Applied minimalism workshop feedback to the Secrets topic

ahardin-rh:
per https://mojo.redhat.com/groups/ccs/blog/2016/12/19/looking-for-a-few-guinea-p-i-mean-volunteers

### DIFF
--- a/dev_guide/secrets.adoc
+++ b/dev_guide/secrets.adoc
@@ -10,15 +10,17 @@
 
 toc::[]
 
-== Overview
+[[dev-guide-secrets-using-secrets]]
+== Using Secrets
+
+This topic discusses important properties of secrets and provides an overview on
+how developers can use them.
 
 The `Secret` object type provides a mechanism to hold sensitive information such
-as passwords, {product-title} client config files, `dockercfg` files, private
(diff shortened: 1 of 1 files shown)


ahardin-rh:
@adellape @bfallonf @bmcelvee @mburke5678 PTAL

mburke5678:
Looks better! 
Did we talk about the use of "it" in the meeting and/or using the nouns like "object," so you would have for ex. Properties of Secret Objects (Properties of Secrets sounds odd). 

Comments:
Line 22: Use of "it" 
Line 22: Passive voice "can be mounted"
Line 65: Plaural? secret/s You must create secrets before the pods that depend on it
Line 65: Use of "it"
Lines 88-90: "An exception to this is when a node is rebooted and the API server must re-read the secret data." An exception to what? Instead of following the prior sentence what should the customer do?

bfallonf:
@ahardin-rh Looks good :star:

ahardin-rh:
@mburke5678 @bfallonf Thanks! Made some edits.

ahardin-rh:
No revision history necessary

PR #3477, opened 2017-01-11
Adding Flannel documentation

danmacpherson:
This is some initial documentation for flannel within OpenShift. This includes:

- Adding the openshift_use_flannel variable to the cluster variables table in the Advanced Installation section
- Adding some more detailed info in the SDN section of the Installation and Configuration guide
- Adding a new section in the Arch Guide about flannel

PR #3743, opened 2017-02-14
Adding flannel arch diagram

danmacpherson:
Also adding variable for flannel nic config


danmacpherson:
Also, have included some follow ups to suggestions in #3477

vikram-redhat:
@knobunc - can you PTAL at the diagram and let us know if any changes are required?

knobunc (approved):
LGTM @openshift/networking PTAL

architecture/additional_concepts/flannel.adoc
```diff
@@ -15,7 +15,9 @@ toc::[]
 
 *flannel* is a virtual networking layer designed specifically for containers. 
 {product-title} can use it for networking containers instead of the default
-software-defined networking (SDN) components.
+software-defined networking (SDN) components. This is useful if running
+{product-title} within a cloud provider platform, such as OpenStack, and you
+want to avoid using dual Open vSwtich SDN on both platforms.
```
danwinship: typo "vSwtich"

oh, you just copied that out of configuring_sdn.adoc, where it is also misspelled...

architecture/additional_concepts/flannel.adoc
```diff
@@ -29,4 +31,9 @@ responsibile for:
...
 Each *flanneld* agent provides this infomation to a centralized *etcd* store so
 other agents on hosts can create an overlay network and route packets to
-any container within the *flannel* network.
\ No newline at end of file
+any container within the *flannel* network.
+
+The following diagram illustrates the architecture and data flow from one
+container to another using a *flannel* network:
+
+image::flannel.png[Flannel Communication]
```
danwinship: This is a diagram of the default flannel configuration, using an overlay network, but our ansible-based flannel install uses flannel in the "host-gw" mode, which is totally different. (The default config does, technically, avoid "dual Open vSwitch" (since flannel doesn't use OVS) but it doesn't avoid double encapsulation of data, which is the actual problem. The "host-gw" mode uses a completely different architecture to avoid that.

danwinship (changes requested):

architecture/additional_concepts/flannel.adoc
vikram-redhat: @danmacpherson - PTAL.

architecture/additional_concepts/flannel.adoc
danmacpherson: Thanks for spotting this. Fixed typo in both places.

architecture/additional_concepts/flannel.adoc
danmacpherson: @danwinship -- Thanks for the feedback. How can we refine the image so that it better represents host-gw mode?

danwinship: I don't know the details of how host-gw sets up networking, other than that it doesn't use an overlay.

danwinship: (The openshift networking team mostly isn't involved with flannel stuff.)

danmacpherson: @danwinship -- from what I've seen, host-gw sets up a series of routes to direct packets, which is why I had the UDP encapsulation to be removed from the original image here:

https://raw.githubusercontent.com/coreos/flannel/master/packet-01.png

So maybe the image is still usable but in the text we need to specify that the default ansible install uses host-gw mode, which requires direct L2 connection between all nodes so that the packets get routed accordingly. What do you think?

danwinship: Hm... I was thinking host-gw didn't use veths or docker0, but maybe that's wrong. But I'm pretty sure there shouldn't be any "flannel0" at least. Anyway, I guess this means you need to find someone who knows more about flannel than me to review it.

danmacpherson: @sferich888 Any chance you could take a look at the diagram?

Also @danwinship I checked my OCP w/flannel lab I'm using to test. When I run an `ip addr`, I can confirm it lists docker0 and veth devices. Not sure about flannel0, but judging by the IP, it seems more like a descriptor for the network range rather than a device. Having said that, I'm okay with removing it.

vikram-redhat: @openshift/networking - can someone have a look at this?

vikram-redhat: @tracyrankin - are you able to help here? Just need someone to review the diagram. Thanks!

rajatchopra (changes requested):

architecture/additional_concepts/flannel.adoc
danmacpherson: Thanks, @rajatchopra ! I'll organize to get the diagram modified per your suggestion.

danmacpherson: @rajatchopra -- I've got the image updated along with the text to accommodate your feedback. I've put the example routes underneath the image, rather than within the image itself.

How does it all look now?

rajatchopra: The flanneld seems to be on the data path of the packet. Maybe connect docker0 directly to the routing table and show flanneld connected to the routing table in a different color connector.
Thanks for working on this, would be really helpful in understanding what's going on underneath.

danmacpherson: @rajatchopra -- modifed the diagram as per your feedback -- the flanneld component should be out of the way of the data path, but still interacting with the routing table. How does this look now?

rajatchopra: Looks good. Thanks.

rajatchopra (approved):

bfallonf:
[rev_history]
|xref:../install_config/configuring_sdn.adoc#install-config-configuring-sdn[Configuring the SDN]
|Added information about the `flannel_interface` variable.
%
|xref:../architecture/additional_concepts/flannel.adoc#architecture-additional-concepts-flannel[Additional Concepts -> Flannel]
|Updated the topic with information on OpenStack installs, running Flannel in *host-gw* mode, and a diagram with route examples. 
%

PR #3763, opened 2017-02-16
Optimize the commands to enable root group access in the Dockerfile

tstjep:
Reduce the number of instructions to set the desired permissions (gid 0, g+rx for files, g+x for directories) in the Dockerfile:

- Use only one RUN statement
- Remove find command and replace it with a +X in the chmod command (X as +x for directories only)

### DIFF
--- a/creating_images/guidelines.adoc
+++ b/creating_images/guidelines.adoc
@@ -281,9 +281,8 @@ to allow users in the root group to access them in the built image:
 
 ====
 ----
-RUN chgrp -R 0 /some/directory
-RUN chmod -R g+rw /some/directory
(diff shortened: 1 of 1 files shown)


ahardin-rh:
@tstjep Thank you!
@bparees PTAL

bparees:
lgtm

ahardin-rh:
Thanks!

ahardin-rh:
[rev_history]
|xref:../creating_images/guidelines.adoc#creating-images-guidelines[Guidelines]
|Updated the commands in the xref:../creating_images/guidelines.adoc#use-uid[Support Arbitrary User IDs] section to enable root group access in the Dockerfile.
%

PR #3765, opened 2017-02-16
Follow-up to PR3529 Deploy metrics using ansible: https://trello.com/c/piaWOfvB/

bmcelvee:
Follow-up edits to https://github.com/openshift/openshift-docs/pull/3529.

### DIFF
--- a/install_config/cluster_metrics.adoc
+++ b/install_config/cluster_metrics.adoc
@@ -19,7 +19,7 @@ exposes metrics that can be collected and stored in back-ends by
 link:https://github.com/GoogleCloudPlatform/heapster[Heapster].
 
 As an {product-title} administrator, you can view a cluster's metrics from all
-containers and components in one user interface.  These metrics are also
+containers and components in one user interface. These metrics are also
 used by xref:../dev_guide/pod_autoscaling.adoc#dev-guide-pod-autoscaling[horizontal pod autoscalers]
 in order to determine when and how to scale.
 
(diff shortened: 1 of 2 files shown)


bmcelvee:
There are a few things from https://github.com/openshift/openshift-docs/pull/3529 that still need to be addressed, but in the meantime I made a few minor edits to the content. 

@jcantrill @ahardin-rh @adellape

install_config/cluster_metrics.adoc
```diff
@@ -66,78 +66,28 @@ address of the node, then Heapster will fail to retrieve any metrics.
...
-The components for cluster metrics must be deployed to the *openshift-infra*
-project. This allows xref:../dev_guide/pod_autoscaling.adoc#dev-guide-pod-autoscaling[horizontal pod
-autoscalers] to discover the Heapster service and use it to retrieve metrics
-that can be used for autoscaling.
-
-All of the following commands in this topic must be executed under the
-*openshift-infra* project. To switch to the *openshift-infra* project:
+An Ansible playbook is available to deploy cluster metrics. You should
+familiarize yourself with the
+xref:../install_config/install/advanced_install.adoc[Advanced Installation and
```
ahardin-rh: We need to have an anchor at the end of the path. In this case, we can use the ID at the top of the page you're pointing to. Also, looks like the topic is actually called Advanced Installation (not Advanced Installation and Configuration):
xref:../install_config/install/advanced_install.adoc#install-config-install-advanced-install[Advanced Installation]

install_config/cluster_metrics.adoc
```diff
@@ -313,236 +308,186 @@ pod is running. Ensure *_/var_* has enough free space to accommodate metrics
...
-$ oc secrets new metrics-deployer nothing=/dev/null
-----
-
-[[deployer-secret-options]]
-==== Deployer Secret Options
-
-The following table contains more advanced configuration options, detailing all
-the secrets which can be used by the deployer:
+The `openshift_metrics` role included with OpenShift Ansible defines the tasks to deploy
+cluster metrics. Following is a list of role variables that can
```
ahardin-rh: The following is a list...

ahardin-rh:
@bmcelvee Just 2 comments from me. Otherwise, looks good!

bmcelvee:
@jcantrill I have a few final questions, then we can finally merge this PR. There were a few comments on #3529 that I wanted to follow up on.

1. Should the following paragraphs be removed (based on Bruno's comments): 
"By default, metrics data is stored for seven days. You can configure this by setting the openshift_metrics_duration variable in your hosts file, which will update the upon deployment metrics.yaml configuration file. After seven days, Cassandra begins to purge the oldest metrics data. Metrics data for deleted pods and projects is not automatically purged; it is only removed once the data is over seven days old."

"The hawkular-metrics.pem value needs to contain the certificate in its .pem format. You may also need to provide the certificate for the Certificate Authority which signed this pem file via the hawkular-metrics-ca.cert secret." 

2. Matt left the following comment, is there a link we need to add: 
"For origin we need to provide a link to where we can download the file if OpenShift ansible is not installed (which is a valid use case for origin installations).

For OCP, the location should be known before hand, why its it <OPENSHIFT_ANSIBLE_DIR>? or is this a resolved to something during the build?" 

Thanks!

jcantrill:
Regarding 1:  He is correct. The deplyoment metrics.yaml is no longer used but the content is relevent.  There is a purge that still happens, its just this value is used to populate part of the cassandra replication controller definition.  It is reasonable to modify the statement to mention nothing about 'deployment metrics.yaml'

Regarding 2: I would disregard.  That is referencing a playbook that must be used in conjunction with a set of other files otherwise it is useless.

bmcelvee:
[rev_history]
|xref:../install_config/cluster_metrics.adoc#install-config-cluster-metrics[Enabling Cluster Metrics]
|Added information about deploying metrics using the ansible role 'openshift_metrics'.
%
|xref:../install_config/install/advanced_install.adoc#install-config-install-advanced-install[Advanced Installation]
|Added information about deploying metrics using the ansible role 'openshift_metrics'.
%

PR #3778, opened 2017-02-17
Added Ceph to RBD references per Trello

mburke5678:


ahardin-rh:
LGTM! :+1:

vikram-redhat:
@childsb - can you PTAL?

mburke5678:
Per @childsb email of 3/10:
"On first inspection this looks Ok to me.  I say lets call it complete and if we notice other occurrences or errors in the docs later we'll fix it then."

PR #3784, opened 2017-02-20
Removed note about forthcoming configuration documentation

ahardin-rh:
Removing the note per discussion with @srampal


ahardin-rh:
Not to complicate things, but after opening this pull request to remove the note box, I think it is best for us to preserve the note in the Contiv topic so that the user knows that the information is not omitted in error and that we are working on it. We can reopen this later if we find it necessary.

PR #3785, opened 2017-02-20
Remove s2i extended build documentation

csrwng:
Removes extended build documentation from the docs because we are deprecating the feature and adds documentation on how this can be achieved using image source and chaining builds.


csrwng:
@bparees ptal 
I wasn't sure where to best put this doc, so I put it in the advanced build operations section. Let me know if you think it can go in a better spot.

dev_guide/builds/advanced_build_operations.adoc
```diff
@@ -122,3 +122,123 @@ endif::[]
 If the specified `NodeSelector` cannot be matched to a node with those labels,
 the build still stay in the `Pending` state indefinitely.
 ====
+
+[[dev-guide-chaining-builds]]
+== Chaining Builds
+
+For compiled languages (Go, C, C++, Java, etc.) the dependencies necessary for compilation might increase the size of the
```
bparees: s/the dependencies necessary for compilation/including the dependencies necessary for compilation in the application image/

dev_guide/builds/advanced_build_operations.adoc
```diff
@@ -122,3 +122,123 @@ endif::[]
...
+image or introduce vulnerabilities that can be exploited.
+
+To avoid these problems, a
+xref:../../architecture/core_concepts/builds_and_image_streams.adoc#source-build[Source-to-Image]
+build can be chained with a
+xref:../../architecture/core_concepts/builds_and_image_streams.adoc#docker-build[Docker]
```
bparees: any 2 build types can be chained together, it doesn't have to be s2i -> docker.

dev_guide/builds/advanced_build_operations.adoc
```diff
@@ -122,3 +122,123 @@ endif::[]
...
+                                       |
+                                       v
+                               +---------------+
+                               | Runtime Image |
+                               | with Compiled |
+                               |    Artifact   |
+                               +---------------+
+----
+
+The first build, which takes source and produces a binary artifact, is a a regular Source-to-Image
```
bparees: s/a a/a/

dev_guide/builds/advanced_build_operations.adoc
```diff
@@ -122,3 +122,123 @@ endif::[]
...
+build. In the following example, a JEE source repository is to build a WAR file that is output to
```
bparees: "a JEE source repository is used to build an image containing a WAR file.  The image is output to(pushed to?) the `artifact-image` ImageStream."

dev_guide/builds/advanced_build_operations.adoc
```diff
@@ -122,3 +122,123 @@ endif::[]
...
+metadata:
+  name: artifact-build
+spec:
+  output:
+    to:
+      kind: ImageStreamTag
+      name: artifact-image
+  source:
+    git:
+      uri: https://github.com/bparees/openshift-jee-sample.git
```
bparees: i think it's time to put a copy of this repo in the openshift org, i don't need my name in the official product docs :)

bparees: i've moved it into openshift.

dev_guide/builds/advanced_build_operations.adoc
```diff
@@ -122,3 +122,123 @@ endif::[]
...
+    sourceStrategy:
+      from:
+        kind: ImageStreamTag
+        name: wildfly:10.1
+        namespace: openshift
+    type: Source
+----
+
+
+The second build uses Image Source and, in this case, an inline Dockerfile to place the binary artifact resulting
```
bparees: link to the imagesource doc

bparees (changes requested):
advanced operations seems like as a good a place as any.

dev_guide/builds/advanced_build_operations.adoc
csrwng: Yes they can, but this is about a specific solution to the problem of capturing binary output from one build and including it in a runtime image. Maybe we should have a different title?

csrwng:
comments addressed

bparees (changes requested):
needs a little more detail around the fact that the runtime image doesn't have/need any compilation tools in it, and pointing out that anytime the first build produces a new image, the second build will automatically be triggered due to the imagechangetrigger (but this is optional), making it effectively act as a single build flow.

csrwng:
@bparees added a paragraph explaining what you said :)

dev_guide/builds/advanced_build_operations.adoc
```diff
@@ -122,3 +122,132 @@ endif::[]
...
+might increase the size of the image or introduce vulnerabilities that can be exploited.
+
+To avoid these problems, two builds can be chained together. One that will produce the compiled artifact, and a second build
+that will place that artifact in a separate image that will run the artifact.
+
+In the following example, a
+xref:../../architecture/core_concepts/builds_and_image_streams.adoc#source-build[Source-to-Image]
+build is combined with a
+xref:../../architecture/core_concepts/builds_and_image_streams.adoc#docker-build[Docker]
+build to compile an artifact that is then placed in a separate runtime image.
```
bparees: "Note: although this example chains an s2i build and a docker build, the first build can use any strategy that will produce an image containing the desired artifacts, and the second build can use any strategy that can consume input content from an image."

(wordsmith as desired).

bparees (changes requested):

dev_guide/builds/advanced_build_operations.adoc
csrwng: done

bparees (approved):

bparees:
lgtm @openshift/team-documentation ptal

adellape:
@csrwng Picking this up via https://github.com/openshift/openshift-docs/pull/3829.

PR #3786, opened 2017-02-21
Doc update for origin PR #12846

louyihua:
This update is corrpeonding to openshift/origin PR #12846.


ahardin-rh:
Looks like the Origin PR on this is still open https://github.com/openshift/origin/pull/12846

louyihua:
Since the probe problem may be resolved through other mechanisms, I'm now closing this PR.

PR #3794, opened 2017-02-21
Fixes for Dedicated branch

vikram-redhat:


vikram-redhat:
I will cherry pick this to various branches

PR #3803, opened 2017-02-21
Change default arp cache size on nodes

pecameron:
Code fix is in 3.6

In OCP clusters with large numbers of routes (greater than the value of
net.ipv4.neigh.default.gc_thresh3, which is 1024 by default) the ARP
cache is not large enough to accommodate for all the entries needed by
the nodes running the router pods.

This change increases the cache size.

bug 1425388
https://bugzilla.redhat.com/show_bug.cgi?id=1425388

Signed-off-by: Phil Cameron <pcameron@redhat.com>

### DIFF
--- a/install_config/router/default_haproxy_router.adoc
+++ b/install_config/router/default_haproxy_router.adoc
@@ -1473,11 +1473,11 @@ accordingly.
 == ARP Cache Tuning for Large-scale Clusters
 
 In {product-title} clusters with large numbers of routes (greater than the
-value of `net.ipv4.neigh.default.gc_thresh3`, which is `1024` by default), you
-must increase the default values of sysctl variables to allow more entries in
-the ARP cache.
+value of `net.ipv4.neigh.default.gc_thresh3`, which is `65536` by default), you
+must increase the default values of sysctl variables on each node in the cluster
+running the router pod to allow more entries in the ARP cache.
 
-The kernel messages would be similar to the following:
+When the problem is occuring, the kernel messages would be similar to the following:
 
 ----
 [ 1738.811139] net_ratelimit: 1045 callbacks suppressed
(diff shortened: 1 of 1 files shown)


pecameron:
@eparis PTAL

install_config/router/default_haproxy_router.adoc
```diff
@@ -1498,18 +1498,45 @@ To verify the actual amount of ARP entries for IPv4, run the following:
...
+# sysctl net.ipv4.neigh.default.gc_thresh1=8192
+# sysctl net.ipv4.neigh.default.gc_thresh2=32768
+# sysctl net.ipv4.neigh.default.gc_thresh3=65536
+----
+
+To make these settings permanent across reboots, until an {product-title} upgrade 
+occurs, edit the following file on each node:
+
+----
+/usr/lib/tuned/atomic-openshift-node-host/tuned.conf
```
eparis: @jeremyeder This can't be right, can it? edit a file in /usr/? What is the right thing here?

pecameron: @eparis @jeremyeder 
# rpm -ql tuned-profiles-atomic-openshift-node
/usr/lib/tuned/atomic-openshift-node-guest
/usr/lib/tuned/atomic-openshift-node-guest/tuned.conf
/usr/lib/tuned/atomic-openshift-node-host
/usr/lib/tuned/atomic-openshift-node-host/tuned.conf
/usr/share/licenses/tuned-profiles-atomic-openshift-node-3.3.0.13
/usr/share/licenses/tuned-profiles-atomic-openshift-node-3.3.0.13/LICENSE
/usr/share/man/man7/tuned-profiles-atomic-openshift-node.7.gz

eparis: `/usr` is for files from a distro.  `/etc` is for files from an admin. On atomic host `/usr` is in fact read-only. We should never document an admin changing a file in `/usr`.  (I'm not saying I know the answer, I just know the not-answer)

pecameron: @eparis 
man tuned.conf
DESCRIPTION
       This  man page documents format of Tuned 2.0 profile definition files. Profile is stored in /etc/tuned/<profile_name>/tuned.conf or in
       /usr/lib/tuned/<profile_name>/tuned.conf file where the /etc/tuned/ directory has higher priority.

eparis: perfect, looks like you found the answer!  `/etc/tuned/<profile_name>/tuned.conf` !

pecameron: @eparis So we should tell the admin to copy /usr/lib/tuned/atomic-openshift-node-host/tuned.conf to /etc/tuned/atomic-openshift-node-host/tuned.conf and then add the lines. Note RHEL relies on files in /usr/lib/tuned/. Also, since the rpm puts the file in /usr/lib/tuned/atomic-openshift-node-host/tuned.conf we need to document the upgrade process to include verifying and maybe updating /etc/tuned/atomic-openshift-node-host/tuned.conf and when deleting the rpm we need to tell people to also delete /etc/tuned/atomic-openshift-node-host/tuned.conf. Failure to delete the file will mean that the RHEL defaults are not restored.

This all sounds very wrong to me.

eparis: https://access.redhat.com/solutions/1305833

This article suggests creating a new profile in /etc which includes the original and then sets just the things it wants set. Updates to the original will get applied, but customization the user makes will never be overwritten (which is what we want)

Any suggestion that a user edit a file in `/usr` is wrong.

jeremyeder:
Internet is down at my house. Will reply in-depth tomorrow. AT&T--

install_config/router/default_haproxy_router.adoc
jeremyeder: This potential drift occurs every time we change content of tuned profiles that we ship.  That is one of the reasons I wrote that kbase article 1305833 :-)

I would suggest the OCP docs link out to the kbase article (tuned is not OCP specific) rather than attempt to re-create any of the content of that kbase.  That should cover customers who need this tuning before 3.6.

Hope that covers all concerns?

eparis: SGTM, unless the article is behind the paywall..

jeremyeder: The kbase article is behind the paywall.  The official docs for custom profiles are here:

https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html-single/Performance_Tuning_Guide/index.html#custom-profiles

pecameron: @eparis @jeremyeder I linked out to the doc. Any other changes needed?

eparis:
it looks great to me. Thank you Phil!

jeremyeder:
ack, thanks.

pecameron:

ahardin-rh:
Looks great. Thank you!

ahardin-rh:
[rev_history]
|xref:../install_config/router/default_haproxy_router.adoc#install-config-router-default-haproxy[Setting up a Router -> Using the Default HAProxy Router]
|Updated the default values for ARP cache size on nodes in the xref:../install_config/router/default_haproxy_router.adoc#deploy-router-arp-cach-tuning-for-large-scale-clusters[ARP Cache Tuning for Large-scale Clusters] section.
%

PR #3830, opened 2017-02-23
Bug 1388906, Added information about disconnected installation of the Atomic Registry

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1388906

### DIFF
--- a/install_config/install/disconnected_install.adoc
+++ b/install_config/install/disconnected_install.adoc
@@ -4,6 +4,7 @@
 {product-version}
 :latest-tag: v3.4.1.5
 :latest-int-tag: 3.4.1
+:latest-registry-console-tag: 3.4
 :data-uri:
 :icons:
 :experimental:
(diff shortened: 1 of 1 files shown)


ahardin-rh:
@adellape PTAL

adellape (changes requested):
Discussed on IRC:

- Can remove the ifdef because this topic is only for OCP anyway.
- Needs a new entity (something like `latest-registry-console-tag`) in the metadata since this image uses a separate set of tags from the core OCP and the logging/metric images, unfortunately:

https://access.redhat.com/containers/#/tags/580a32536017d9328b5792da

ahardin-rh:
@adellape Thanks! Updated.

ahardin-rh:
Note to self: must change the tag for 3.3 when I cherry-pick that in during the weekly release :notebook:

install_config/install/disconnected_install.adoc
```diff
@@ -233,6 +234,15 @@ registry.access.redhat.com/jboss-webserver-3/webserver30-tomcat7-openshift:lates
 registry.access.redhat.com/jboss-webserver-3/webserver30-tomcat7-openshift:1.1
 ----
 
+. If you are using a stand-alone registry, you must pull the *registry-console*
+image.
```
adellape: @ahardin-rh Forgot to say earlier, we should probably also say here "If you are using a stand-alone registry or plan to enable the registry console with the integrated registry". Since that uses the same image too (https://docs.openshift.com/container-platform/3.4/install_config/registry/deploy_registry_existing_clusters.html#registry-console).

adellape (changes requested):

ahardin-rh:
@adellape Updated. Thanks!

ahardin-rh:
Verified by QE in the BZ

ahardin-rh:
[rev_history]
|xref:../install_config/install/disconnected_install.adoc#install-config-install-disconnected-install[nstalling a Cluster -> Disconnected Installation]
|Added information about disconnected installation of the stand-alone registry.
%

PR #3846, opened 2017-02-28
Fix status codes for success

hobti01:
Liveness and readiness probes succeed if the HTTP response code is 200-399, which is the range of HTTP success (2xx) and redirection (3xx).


ahardin-rh:
Thank you!

PR #3853, opened 2017-03-01
add kexec sos psacct packages for support

git001:
The RHEL Support need the informations from kdump sa and sos report.
I would be wise to have this packages already installed on all machines


ahardin-rh:
Thank you!

ahardin-rh:
No revision history needed

PR #3856, opened 2017-03-02
Add the attribute [autofocus] on [Search Documentation] input text

yu8833:
1. When opening https://docs.openshift.com/, the focus should be on the input text. fixed the #3857 .
2. The attribute [autofocus] must have a value.

### DIFF
--- a/index-commercial.html
+++ b/index-commercial.html
@@ -354,7 +354,7 @@ height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
         </li>
         <li>
           <form action="https://help.openshift.com/hc/en-us/search" method="get" id="cse-search-form">
-            <input name="query" id="cse-search-input" class="navbar-search-query form-control" type="text" placeholder="Search" tabindex="1" autocomplete="off" autofocus>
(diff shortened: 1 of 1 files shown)


bfallonf:
Thanks @YuPengZTE . @vikram-redhat who would be best to verify this PR?

ahardin-rh:
@luciddreamz @nhr Can you please verify this change? Thanks!

index-commercial.html
```diff
@@ -354,7 +354,7 @@
         </li>
         <li>
           <form action="https://help.openshift.com/hc/en-us/search" method="get" id="cse-search-form">
-            <input name="query" id="cse-search-input" class="navbar-search-query form-control" type="text" placeholder="Search" tabindex="1" autocomplete="off" autofocus>
+            <input name="query" id="cse-search-input" class="navbar-search-query form-control" type="text" placeholder="Search" tabindex="1" autocomplete="off" autofocus="autofocus">
```
luciddreamz: Remove aufocous from this form input, leave it on the other.

Autofocus automatically sets the focus on a specific form input when the page loads. If you're going to use this for whatever reason, you should only autofocus onto a single form field input (you can't have the cursor focus on two fields).

gaurav-nelson:
fixed via https://github.com/openshift/openshift-docs/commit/54c02ee361c3527dc559a33092a1426869fb4448

👋

Issue #3664, opened 2017-02-05
Typo in Using Router Shards


genadipost:
In [(/enterprise-3.4/install_config/router/default_haproxy_router.adoc#using-router-shards)](https://github.com/openshift/openshift-docs/blob/enterprise-3.4/install_config/router/default_haproxy_router.adoc#using-router-shards)
```In addition, this enables us to create more complex routing rules ala divert high priority traffic to the dedicated finops-router but send the lower priority ones to the devops-router.```

Not sure what ```ala``` means, is it a typo?
If it is, could this be a fix:

```In addition, this enables us to create more complex routing rules that allow to divert high priority traffic to the dedicated finops-router but send the lower priority ones to the devops-router.```

php-coder:
I believe that it's not a typo but a word/phrase that came from the French language and means "in style of" or "like". See also: https://en.wiktionary.org/wiki/a_la

PR #3874, opened 2017-03-06
Style edits to the Using Router Shards section

ahardin-rh:
Resolves https://github.com/openshift/openshift-docs/issues/3664


ahardin-rh:
Minor style edits. No revision history needed.

PR #3907, opened 2017-03-09
Set host-network to false in f5_router.adoc

juanluisvaladas:
host-network is set to true by default. There is no reason to require this, therefore we should set it false.


ahardin-rh:
Thank you!

ahardin-rh:
[rev_history]
|xref:../install_config/router/f5_router.adoc#install-config-router-f5[Setting up a Router -> Using the F5 Router Plug-in]
|Updated the example in the xref:../install_config/router/f5_router.adoc#deploying-the-f5-router[Deploying the F5 Router] section to reflect `host-network` to `false`.
%

PR #3923, opened 2017-03-13
Clarifies that OSD customers can utilize wildcard CNAMES for custom domains on routes

wgordon17:


adellape (approved):
:+1:

adellape:
[rev_history]
|xref:../dev_guide/getting_traffic_into_cluster.adoc#getting-traffic-into-cluster[Getting Traffic into the Cluster]
|Clarified that {product-title} customers can utilize wildcard CNAMEs for custom domains on routes.
%

adellape:
@wgordon17 See above, I think just putting the ifdef right in the rev_history comment here (as you suggested) will work for our purposes.

PR #3925, opened 2017-03-13
Adds comment regarding not supporting password protected key files

wgordon17:
Content was taken from :https://docs.openshift.com/container-platform/3.4/install_config/router/default_haproxy_router.html#using-secured-routes

Since these documents are filtered out of the OpenShift Dedicated docs, it's important to keep this information in a place where Dedicated customers can see it.

### DIFF
--- a/dev_guide/routes.adoc
+++ b/dev_guide/routes.adoc
@@ -114,6 +114,14 @@ spec:
 ----
 ====
 
+Currently, password protected key files are not supported. HAProxy prompts for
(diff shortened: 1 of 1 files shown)


adellape (approved):
:+1:

adellape:
[rev_history]
|xref:../dev_guide/routes.adoc#dev-guide-routes[Routes]
|Added comment regarding not supporting password protected key files to the xref:../dev_guide/routes.adoc#creating-routes[Creating Routes] section.
%

PR #3935, opened 2017-03-14
Bug 1426361 Arranged the containerized install info

bfallonf:
As per: https://bugzilla.redhat.com/show_bug.cgi?id=1426361

An attempt to make the containerized install clearer and obvious.

I'll ask for an ack in the BZ.

### DIFF
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -231,7 +231,7 @@ Topics:
         File: prerequisites
       - Name: Host Preparation
         File: host_preparation
-      - Name: Containerized Components
+      - Name: Installing on Containerized Hosts
         File: rpm_vs_containerized
       - Name: Quick Installation
         File: quick_install
--- a/install_config/install/quick_install.adoc
+++ b/install_config/install/quick_install.adoc
@@ -24,15 +24,21 @@ the `atomic-openshift-installer` command, to install {product-title} across a
 set of hosts. This installer can deploy {product-title} components on targeted
 hosts by either installing RPMs or running containerized services.
 
+[IMPORTANT]
+====
+While RHEL Atomic Host is supported for running containerized {product-title}
+services, the installer is
+xref:../../install_config/install/host_preparation.adoc#software-prerequisites[provided
+by an RPM] and not available by default in RHEL Atomic Host. Therefore, it must
+be run from a Red Hat Enterprise Linux 7 system. The host initiating the
+installation does not need to be intended for inclusion in the {product-title}
+cluster, but it can be.
+====
+
 This installation method is provided to make the installation experience easier
 by xref:running-an-interactive-installation[interactively gathering the data]
 needed to run on each host. The installer is a self-contained wrapper intended
-for usage on a Red Hat Enterprise Linux (RHEL) 7 system. While RHEL Atomic Host
-is supported for running containerized {product-title} services, the installer is
-xref:../../install_config/install/host_preparation.adoc#software-prerequisites[provided
-by an RPM] not available by default in RHEL Atomic Host, and must therefore be
-run from a RHEL 7 system. The host initiating the installation does not need to
-be intended for inclusion in the {product-title} cluster, but it can be.
+for usage on a Red Hat Enterprise Linux (RHEL) 7 system. 
 
 In addition to running xref:running-an-interactive-installation[interactive
 installations] from scratch, the `atomic-openshift-installer` command can also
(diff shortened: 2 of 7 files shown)


sferich888:
@jduncan-rva I know you were looking into this, can you also review.

install_config/install/quick_install.adoc
```diff
@@ -24,15 +24,21 @@ the `atomic-openshift-installer` command, to install {product-title} across a
...
+====
+While RHEL Atomic Host is supported for running containerized {product-title}
+services, the installer is
+xref:../../install_config/install/host_preparation.adoc#software-prerequisites[provided
+by an RPM] and not available by default in RHEL Atomic Host. Therefore, it must
+be run from a Red Hat Enterprise Linux 7 system. The host initiating the
+installation does not need to be intended for inclusion in the {product-title}
+cluster, but it can be.
+====
+
```
sferich888: This seems like an engeering bug. @sdodson do you know if we have any plans to address this?

install_config/install/rpm_vs_containerized.adoc
```diff
@@ -13,24 +13,56 @@ toc::[]
...
+configures services to run within the same user space, while a containerized
+installation installs services using container images and runs separate services
+in individual containers.
 
-- For the xref:../../install_config/install/quick_install.adoc#install-config-install-quick-install[quick installation
-method], you can choose between the RPM or containerized method on a per host
-basis during the interactive installation, or set the values manually in an
+The default method for installing {product-title} on
+ifdef::openshift-origin[]
+Fedora, CentOS, or RHEL
```
sferich888: Why mentioned distributions we don't support? This should just say RHEL. Upstream it can read this but for the OCP docs this needs to only say RHEL.

sferich888 (changes requested):
This seems to work. However I would like others to review. I have mentioned them and hopefully they reply quickly.

install_config/install/quick_install.adoc
sdodson: We'll provide a containerized installer in the future, but we won't start on that until after 3.6.

install_config/install/rpm_vs_containerized.adoc
sdodson: This is inside a conditional for origin, see `ifdef::openshift-origin[]`

jduncan-rva:
I ran this by Eric after catching the question from a customer and not having a good answer. 

The updated docs look like a good step in the right direction, and I'm glad to hear we have a containerized installation/upgrade process on the drawing board.

bfallonf:
OK. If there's no comments or suggestions, then I'll move this to peer review.

@ahardin-rh @adellape @bmcelvee @mburke5678 Any thoughts? I've not added much, but I've moved around some of the info in the hopes to clear up the Atomic Host requirements.

ahardin-rh:
@bfallonf Overall, LGTM. Looks like we just need a rebase :rainbow:

_topic_map.yml
```diff
@@ -231,7 +231,7 @@ Topics:
         File: prerequisites
       - Name: Host Preparation
         File: host_preparation
-      - Name: Containerized Components
+      - Name: Containerized Installation
```
adellape: @bfallonf As you mention in the BZ, there's not a ton of difference when actually performing a Quick or Advanced install for containerized vs RPM-based. I worry that renaming the topic to "Containerized Installation" suggests that this is now a topic that is going to show you how to specifically kick off a "containerized install", when really other than the private registry edge case mentioned under "Required Images" (about those Ansible / environment variables), most of the topic is a discussion about the differences in how these OCP components behave when containerized vs RPM-based.

install_config/install/rpm_vs_containerized.adoc
```diff
@@ -13,24 +13,56 @@ toc::[]
...
+|===
+|Type |RPM  |Containerized
+
+|Installation Method |Packages via `yum` |Container images via `docker`
+|Service Management |`systemd` |`docker` and `systemd` units
+|Operating System | Red Hat Enterprise Linux | Red Hat Enterprise Linux or Red Hat Atomic Host
+|===
+
+[[install-config-install-quick-advanced-install]]
+== Containerized Installs: The Quick and Advanced Methods
```
adellape: @bfallonf Could this title be something like "Installation Methods for Containerized Hosts"? The "Containerized Installs:" part seems superfluous since that's also the (currently, at least) title of the topic, more or less. Really we don't want to make it seem like containerized installs are some special other method, so whatever can help make it clear that quick/advanced can handle both containerized and RPM.

adellape: Also the anchor for this section seems weird.

install_config/install/rpm_vs_containerized.adoc
```diff
@@ -13,24 +13,56 @@ toc::[]
...
+detection of the *_/run/ostree-booted_* file.
+====
+
+The following table outlines the differences between the RPM and Containerized
+methods:
+
+[options="header"]
+|===
+|Type |RPM  |Containerized
+
```
bfallonf: I agree.

install_config/install/rpm_vs_containerized.adoc
bfallonf: Yeah I played around with this a lot. I feel like every iteration didn't explain it well enough.

I'll go with "Installing on Containerized Hosts"

_topic_map.yml
bfallonf: I'll make an overall comment below

bfallonf:
@adellape So the title comments:

I've named the topic "Installing on Containerized Hosts" and the section about the install methods "Install Methods for Containerized Hosts". You can see why I had so much difficulty with the names... I talked this over with @vikram-redhat and @gaurav-nelson , and they agree. I'll merge, cos last day of the sprint, but let me know if you're zealously against it.

bfallonf:
[rev_history]
|xref:../install_config/install/rpm_vs_containerized.adoc#install-config-install-rpm-vs-containerized[Installing a Cluster -> Installing on Containerized Hosts]
|Changes to the Installing on Containerized Hosts topic to promote the information on installing on containerized hosts.  
%

PR #3963, opened 2017-03-17
Fix wrong resource type and name

liangxia:
We are creating pvc here, so the type should be persistentvolumeclaim, and the pvc name should be gluster-claim according to file glusterfs-pvc.yaml which is listed above.

This apply to branch/enterprise-3.3 and above.


ahardin-rh:
Thank you!

ahardin-rh:
No revision history needed

PR #3969, opened 2017-03-17
Bug 1425621 Changed  rolebinding-reader to cluster-reader

bmcelvee:
https://bugzilla.redhat.com/show_bug.cgi?id=1425621


ewolinetz (changes requested):
This should still be rolebinding-reader.
We do not want to give the service account more access than it needs

bmcelvee:
Thanks, @ewolinetz! Closing PR.

PR #3975, opened 2017-03-20
Made statement about support on RHEL 7 inclusive to both listed versions

ahardin-rh:


ahardin-rh:
No revision history needed.

PR #3986, opened 2017-03-21
Minor fixes to the managing projects topic

Miciah:
Reword a few sentences for clarity, delete a redundant example (which is followed by an almost identical example), and add some missing commas.


ahardin-rh:
@Miciah Thank you! :rainbow:

ahardin-rh:
Minor edits. No revision history needed.

PR #3992, opened 2017-03-21
Various fixes to quota documentation

Miciah:
Pull the definition of "terminal state" out of the "pods" entry of the "Object Counts Managed by Quota" table and place the definition prominently above the "Compute Resources Managed by Quota" table above because entries in both tables use the phrase.

Reword some cumbersome sentences in the table to be a little less cumbersome.

Rephrase the note about how a quota's scope limits the resources to which the quota can apply.

Add some missing periods

Use a gerund after "restricts to".

Correct an "and" to "or".

Rephrase the description of what quotas and limits a container may specify. A quota does not "quota"; rather, it "restricts".

Use the present subjunctive in subordinate clauses after "it requires" ("it requires that every X specify", not "specifies", and likewise for "make" versus "makes").

### DIFF
--- a/admin_guide/quota.adoc
+++ b/admin_guide/quota.adoc
@@ -35,6 +35,11 @@ endif::[]
 The following describes the set of compute resources and object types that may be
 managed by a quota.
 
+[NOTE]
+====
+A pod is in a terminal state if `status.phase in (Failed, Succeeded)` is true.
+====
+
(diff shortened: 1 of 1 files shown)


ahardin-rh:
Thank you!

ahardin-rh:
Minor edits. No revision history needed.

PR #3994, opened 2017-03-21
Fix `oc describe limits` example

Miciah:
Add the missing `-n demoproject` option to the `oc describe limits` example to be consistent with the other examples.

Update the example output to reflect the current output.

Refer to the LimitRange object as an "object" rather than as a "document".

### DIFF
--- a/admin_guide/limits.adoc
+++ b/admin_guide/limits.adoc
@@ -65,7 +65,7 @@ spec:
       maxLimitRequestRatio:
         cpu: "10" <14>
 ----
-<1> The name of the limit range document.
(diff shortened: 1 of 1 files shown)


ahardin-rh:
Thank you!

ahardin-rh:
[rev_history]
|xref:../admin_guide/limits.adoc#admin-guide-limits[Setting Limit Ranges]
|Added the missing `-n demoproject `option to the xref:../admin_guide/limits.adoc#viewing-limits[`oc describe` limits example] and updated the command's output.
%

PR #3995, opened 2017-03-21
Document how to configure the PROXY protocol

Miciah:
Document how to configure the HAProxy router to expect incoming connections to use the PROXY protocol and how to configure Amazon ELB to forward requests using the PROXY protocol.

PR #4007, opened 2017-03-21
Follow-up edits to PR#3995

ahardin-rh:
Follow-up to https://github.com/openshift/openshift-docs/pull/3995

Preview build: http://file.rdu.redhat.com/~ahardin/03212017/3995-followup/install_config/router/proxy_protocol.html

### DIFF
--- a/install_config/router/proxy_protocol.adoc
+++ b/install_config/router/proxy_protocol.adoc
@@ -12,18 +12,19 @@ toc::[]
 == Overview
 
 By default, the HAProxy router expects incoming connections to unsecure, edge,
-and re-encrypt routes to use HTTP. However, the router can be configured to
-expect incoming requests to use
+and re-encrypt routes to use HTTP. However, you can configure the router to
+expect incoming requests by using
 link:http://www.haproxy.org/download/1.8/doc/proxy-protocol.txt[the PROXY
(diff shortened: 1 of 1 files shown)


ahardin-rh:
@bfallonf please peer review :pencil2:

install_config/router/proxy_protocol.adoc
```diff
@@ -111,11 +106,14 @@ $ aws elb create-load-balancer --load-balancer-name "$lb" \
...
 }
 ----
-====
-+
-3. Register your router instance or instances with the ELB:
 +
+[NOTE]
 ====
+You must configure all listeners to use the TCP protocol, not the HTTP protocol.
+====
```
bfallonf: I can see the note box above came from before the step, but I'm not sure being after the step is best. Can I suggest:

* Placing the note box above the example with a rewrite to something like "Note that the following example configures all listeners to use the TCP protocol and not the HTTP protocol"

or

* Putting a footnote into the example with something like "Using Protocol=TCP configures all listeners to use the TCP protocol and not the HTTP protocol, which is necessary for these steps to work"

WDYT?

bfallonf:
@ahardin-rh Just one note from me. Otherwise, good job! :star2:

PR #4008, opened 2017-03-22
Add openshift/hello-openshift dependency

juanluisvaladas:
docker.io/openshift/hello-openshift is required to run successfully oc adm diagnostics.
Otherwise the user will not have network diagnostics


juanluisvaladas:
bump

ahardin-rh:
@juanluisvaladas Thanks! LGTM. I have this lined up to be applied to the OCP 3.3+ docs. Please let me know if that is not correct.

Can you please squash your changes into 1 commit? This is necessary for our publishing process. Thanks again and sorry for the delay!

PR #4020, opened 2017-03-24
Update rest API docs to 1.5 levels

smarterclayton:


smarterclayton:
@adellape this was missed - we need to update this in the 3.5 and 1.5 branches (even if post-release)

PR #4023, opened 2017-03-24
Don't list "m" as a valid suffix for memory

Miciah:
In specifying a memory request or limit, the "m" suffix passes validation but is useless because we cannot manage memory at millibyte granularity, so do not list the suffix as allowed.


Miciah:
For reference: https://github.com/kubernetes/kubernetes/issues/28741

ahardin-rh:
@Miciah Thank you!

ahardin-rh:
[rev_history]
|xref:../dev_guide/compute_resources.adoc#dev-guide-compute-resources[Quotas and Limit Ranges]
|Removed "m" as a valid suffix for memory in the xref:../dev_guide/compute_resources.adoc#dev-compute-resources[Compute Resources] section.
%

PR #4044, opened 2017-03-28
require nodeName to match instance name with cloud provider

sjenning:
We had a number of bugs filed over the past year involving customers trying to set the `nodeName` to the hostname (or other things) on nodes with cloud provider integration enabled.  This leads to a myriad of issues involving failed storage mounting, nodes being deleted, etc.

This PR adds an explicit statement that the nodeName should match the instance name when using cloud provider integration.

@derekwaynecarr @sdodson

### DIFF
--- a/install_config/configuring_aws.adoc
+++ b/install_config/configuring_aws.adoc
@@ -106,6 +106,10 @@ kubernetesMasterConfig:
       - "/etc/aws/aws.conf"
 ----
 
+Currently, the `nodeName` *must* match the instance name in AWS in order
+for the cloud provider integration to work properly.  The name must also be
+RFC1123 compliant.
+
 [IMPORTANT]
 ====
 When triggering a containerized installation, only the directories of
--- a/install_config/configuring_gce.adoc
+++ b/install_config/configuring_gce.adoc
@@ -123,7 +123,11 @@ kubeletArguments:
 
 ----
 ====
-+
+
+Currently, the `nodeName` *must* match the instance name in GCE in order
+for the cloud provider integration to work properly.  The name must also be
+RFC1123 compliant.
+
 [IMPORTANT]
 ====
 When triggering a containerized installation, only the directories of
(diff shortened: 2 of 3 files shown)


sdodson:
I believe the SDN still requires that nodeName resolves to an interface that can be used by the SDN, does it not @openshift/networking ?

danwinship:
> I believe the SDN still requires that nodeName resolves to an interface that can be used by the SDN, does it not

Basically; it has to resolve to the IP that other nodes will see when this node sends traffic to them. So, "private" IPs, not "public" ones.

There's some way to override this though; openshift-ansible knows how to configure everything correctly by using `openshift_hostname`/`openshift_public_hostname` and `openshift_ip`/`openshift_public_ip`, so you should probably just look at what it does and make sure the do-it-yourself docs tell people to do the same thing.

sjenning:
@sdodson the instance name is also resolvable to the private IP of the node in all cloud providers i know of.  So this is ok right?

ahardin-rh:
Any update on this? Should this be closed? Thanks!

sjenning:
@ahardin-rh ha, I just looked at this for the first time in months a few hours ago.

We still want this.  In fact, someone contacted me with confusion about it just this week.  Which is what prompted me to look it up again.

https://bugzilla.redhat.com/show_bug.cgi?id=1437399
https://bugzilla.redhat.com/show_bug.cgi?id=1503863

I need to rebase and expand that in addition to the nodeName and instance name matching, the name must be RFC1123 compliant as well.

pravisankar:
https://bugzilla.redhat.com/show_bug.cgi?id=1505266
Related recent bug where hostname did not resolve to private node IP which is needed for openshift SDN to work. 
I will add validation on sdn node to catch this issue early and that will stop running openshift-node service.

ahardin-rh:
@sjenning @pravisankar Thanks! Looks good. We currently release docs on a weekly cadence for 3.4 and later. I assume this work applies to each of those versions and have labeled this as such. Please confirm, then I will get this in. Thanks again!

sjenning:
@ahardin-rh confirmed 3.4 and later

ahardin-rh:
No revision history needed

PR #4066, opened 2017-04-03
Remove no_wdelay from NFS doc

wongma7:
@eparis as requested by email.

no_wdelay is there because the registry requires read-after-write consistency https://github.com/docker/distribution/issues/1176 and we ran into bugs. But this is the generic NFS page, we should not recommend it also for non-registry uses: let's confine the recommendation for no_wdelay to the registry 'known issues' page

### DIFF
--- a/install_config/persistent_storage/persistent_storage_nfs.adoc
+++ b/install_config/persistent_storage/persistent_storage_nfs.adoc
@@ -320,11 +320,9 @@ exported volume on the NFS server should conform to the following conditions:
 - Each export must be:
 +
 ----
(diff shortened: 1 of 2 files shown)


eparis:
LGTM!

vikram-redhat:
@wongma7 - which branches is this valid for? OpenShift 3.4 and less or 3.5 onwards?

eparis:
It can apply to all versions, but if it is just applied to 3.5 and onwards that is fine.

ahardin-rh:
Thank you!

ahardin-rh:
No revision history needed

PR #4068, opened 2017-04-03
F5-router, "Idling applications" feature does not work

pecameron:
OSE   3.5, 3.4, and 3.3

Made a NOTE that this is HAProxy only
(moved last to lines of file into note)

bug 143165
https://bugzilla.redhat.com/show_bug.cgi?id=1431658

### DIFF
--- a/admin_guide/idling_applications.adoc
+++ b/admin_guide/idling_applications.adoc
@@ -67,5 +67,8 @@ Application services become active again when they receive network traffic and
 will be scaled back up their previous state. This includes both traffic to the
 services and traffic passing through routes.
 
-This works when using the default HAProxy router. You will need to configure
-other routers to detect idling to unidle these applications.
+[NOTE]
+====
+Automatic unidling by a router is currently only supported by the default HAProxy router.
+====
+


pecameron:
@knobunc PTAL

admin_guide/idling_applications.adoc
```diff
@@ -16,6 +16,12 @@ toc::[]
 As an {product-title} administrator, you can idle applications in order to
 reduce the consumption of resources. 
 
+[NOTE]
+====
+This works when using the default HAProxy router. You will need to configure
```
knobunc: Erm... you can't configure the F5 to do this.  It requires a code change, and we need to work out how to make it do it (or make it use the service IP... but it doesn't have access to the service network, so it's ugly).

How about "Automatic unidling by a router is currently only supported by the default HAProxy router"

But it really does belong at the bottom since that's the bit that mentions routes.

Should we add a caveat to the F5 section too?  So that people are informed when they pick the F5?

knobunc (changes requested):

admin_guide/idling_applications.adoc
pecameron: OK

pecameron:
@knobunc PTAL

knobunc (approved):
LGTM

Should we also drop a note over in the F5 router information so that you are informed when setting up F5?

pecameron:
@knobunc You still have Changes requested set...

pecameron:
@knobunc Is this ready to merge?

pecameron:
@ahardin-rh Are you the docs person that will handle this?

knobunc:
@pecameron yes, it is ready to merge.  But should we also add a note in the F5 section of the docs noting this limitation?

ahardin-rh:
LGTM!

ahardin-rh:
[rev_history]
|xref:..admin_guide/idling_applications.adoc#admin-guide-idling-applications[Idling Applications]
|Noted that automatic unidling by a router is currently only supported by the default HAProxy router.
%

PR #4087, opened 2017-04-05
Use fewer shell manipulations to get at data

stevekuznetsov:
When grabbing a property from a `systemd` unit file, it is preferrable
to use the built-in `systemctl get-property` instead of using `grep` as
the property may very well span multiple lines, thereby making it hard
to use `grep` to capture it's full value.

Similarly, when inspecting objects in OpenShift, it is preferrable to
use `oc get -o jsonpath='{}'` to index into the actual object rather
than using `grep` on the output of `oc export`.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

### DIFF
--- a/admin_guide/backup_restore.adoc
+++ b/admin_guide/backup_restore.adoc
@@ -174,7 +174,7 @@ To do so, edit the *_/usr/lib/systemd/system/etcd.service_* file, and add
 +
 ----
 # sed -i '/ExecStart/s/"$/  --force-new-cluster"/' /usr/lib/systemd/system/etcd.service
-# cat /usr/lib/systemd/system/etcd.service  | grep ExecStart
+# systemctl show etcd.service --property ExecStart --no-pager
 
 ExecStart=/bin/bash -c "GOMAXPROCS=$(nproc) /usr/bin/etcd --force-new-cluster"
 ----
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
@ahardin-rh was reading through this doc, made some minor edits to make shell interaction a little safer/more idiomatic.

admin_guide/backup_restore.adoc
```diff
@@ -817,16 +817,15 @@ what path to back up by looking at the `*mountPath*` for volumes in the
 . Get the application data `*mountPath*` from the `*deploymentconfig*`:
 +
 ----
-$ oc export dc/jenkins|grep mountPath
-        - mountPath: /var/lib/jenkins
+$ oc get dc/jenkins -o jsonpath='{ .spec.template.spec.containers[?(@.name=="jenkins")].volumeMounts[?(@.name=="jenkins-data")].mountPath }'
```
stevekuznetsov: Looks a little crazy but at the same time you _do_ want to make sure you are looking at the mount path for the correct volume mount on the correct container, as there is no reason we can't add more volume mounts or more containers with volumes in the future to this deployment config.

ahardin-rh:
Thanks so much!

ahardin-rh:
Minor updates to commands. No revision history needed.

Issue #3368, opened 2016-12-12
maven proxy docs should be image specific


bparees:
This doc:
https://docs.openshift.com/container-platform/3.3/install_config/http_proxies.html#using-maven-behind-a-proxy

is referring to behavior that's specific to the xpaas images, it should be part of the xpaas image documentation, not the general product documentation.

bparees:
@bfallonf looks like you introduced the content.

bfallonf:
@bparees I do remember this... I think it was put in when the xpaas docs were still in the openshift docs. I'll work on this. Thanks!

bfallonf:
PR created to comment this out of the OpenShift docs #3386 

Created a BZ for this for xpaas: https://bugzilla.redhat.com/show_bug.cgi?id=1404538

I'll close this issue.

PR #4113, opened 2017-04-10
Bug 1432110 added section on maven behind proxies

bfallonf:
As per: https://bugzilla.redhat.com/show_bug.cgi?id=1432110

@ahardin-rh @adellape @bmcelvee @mburke5678 

Can I get an opinion on this? Originally, this was a section in the OpenShift docs, but there was a issue reported that it shouldn't be, then it was taken out, but then this BZ was reported to put it back in. Instead, I've just put a para telling the reader to check the EAP docs, where the info is now.

Would this be better as it was? The old info is just commented out in the OpenShift docs (just above this new para).

### DIFF
--- a/install_config/http_proxies.adoc
+++ b/install_config/http_proxies.adoc
@@ -279,6 +279,18 @@ MAVEN_ARGS_APPEND=" -s path/to/file"
 ====
 ////
 
+[[using-maven-behind-a-proxy]]
+== Using Maven Behind a Proxy
+
+Using Maven with proxies requires using the `HTTP_PROXY_NONPROXYHOSTS` variable.
+
+See the
+link:https://access.redhat.com/documentation/en-us/red_hat_jboss_middleware_for_openshift/3/html-single/red_hat_jboss_enterprise_application_platform_for_openshift/#configuring_eap_for_openshift[Red
+Hat JBoss Enterprise Application Platform for OpenShift documentation] for
+information about configuring your {product-title} environment for Red Hat
+JBoss Enterprise Application Platform, including the step for setting up Maven
+behind a proxy.
+
 [[s2i-builds]]
 
 == Configuring S2I Builds for Proxies


bfallonf:
@adellape @ahardin-rh @mburke5678 bump

mburke5678:
@bfallonf I don't mind the link to the JBOSS docs, as it appears that the user needs to supply JBOSS information (??). But it is not clear what the "additional step" is; and the link in this PR is different from the link you were asked to add to the docs.

bfallonf:
@mburke5678 I see what you mean. I've aded a little more information to let the reader know why they need to visit that part of the docs.

As for the link, the BZ wanted to link to the list of variables, but that section of the EAP docs talks about how to set it, and little bit of why. I thought it was a better location. The history behind the confusion here is outlined my comment in the BZ.

Let me know what you think?

install_config/http_proxies.adoc
```diff
@@ -279,6 +279,18 @@ MAVEN_ARGS_APPEND=" -s path/to/file"
...
+
+Using Maven with proxies requires using the `HTTP_PROXY_NONPROXYHOSTS` variable.
+
+See the
+link:https://access.redhat.com/documentation/en-us/red_hat_jboss_middleware_for_openshift/3/html-single/red_hat_jboss_enterprise_application_platform_for_openshift/#configuring_eap_for_openshift[Red
+Hat JBoss Enterprise Application Platform for OpenShift documentation] for
+information for how to configure your {product-title} environment for Red Hat
+JBoss Enterprise Application Platform, including the steps for setting up Maven
+behind a proxy.
+
```
mburke5678: Does the user need to do all the steps in the linked section, or just the note that talks about proxies?

adellape:
@bparees Just FYI on this, sort of a follow-up to https://github.com/openshift/openshift-docs/issues/3368.

bfallonf:
@mburke5678 I updated it to be more specific on the step. Basically, s/steps/step. It's the understanding that if the reader is using Maven, then they'll be following these steps anyway, but this is an extra step if they are using an http proxy as well.

bfallonf:
[rev_history]
|xref:../install_config/http_proxies.adoc#architecture-additional-concepts-f5-big-ip[Working with HTTP Proxies]
|Added the xref:../install_config/http_proxies.adoc#using-maven-behind-a-proxy[Using Maven Behind a Proxy] section.
%

PR #4124, opened 2017-04-10
BZ-1433306 Changes options for oc set env example

mburke5678:


mburke5678:
@bmcelvee PTAL

bmcelvee:
There appears to be an extra space before `oc`. Otherwise, LGTM! 🎆

PR #4206, opened 2017-04-20
Image scanning annotations

aweiteka:


aweiteka:
Preview link: https://people.redhat.com/aweiteka/docs/preview/20170510/security/container_content.html

aweiteka:
cc @enoodle @simon3z @lpshikhar @goern @xenserverarmy

simon3z:
@enoodle @pweil- can you review?

xenserverarmy:
A few thoughts....

1. You probably want to have a required field for the scanner version. Providers could have different versions with different capabilities. Free form string feels good there.
2. I'm not clear on the timestamp format and would suggest both human readable and machine readable
3. It might also be worthwhile to have an update timestamp unless the expectation is the timestamp is updated with each "rescan" or update in provider evaluation. If the latter, I'd be prescriptive about it
4. I would like to see a qualityType of  `policy`. Providers with richer policy engines could then use their policy concepts to go beyond simple vulns or license issues.
5. I would probably make the reference attribute a required item. Without it there's no way to validate the data.
6. Suggest changing the description of  `summary` to `summary of issues found`. Maps better to the non-vuln options
7. Suggest changing `score` to something like `items` or `count`. To me `score` has implications which aren't covered in this spec.

goern:
re 1. good Idea
re 2. What is the standard with other openshift object? looks like most of then us `creationTimestamp: 2016-09-08T05:04:46Z`
re 5. is true, is should be required. @aweiteka  do you know if all scan engines provide a human readable summary? if not we should recommend it

@aweiteka should we have something like:
```
Scanning Roles and Responsibilities
Scanning Engine: OSCAP, Blackduck, Twistlock, image-inspector
  Vulnerability database and/or feeds
  Analysis and reporting
  Schedules scans
Policy Engine: CloudForms, Twistlock, Blackduck
  External to platform
  Interprets analysis into policy
  Policy configured by administrator defining what actions to take based on analysis.
  Key integration point between analyzer(s) and platform.
Runtime Platform: OpenShift
  Enforces policy based on binary rules
  Does not interpret analysis
```

somewhere to give a context why we are just talking about the image metadata?

xenserverarmy:
For the summary, my thinking is a link to the scanner UI for that specific report. That way scanners can be as rich as possible without artificially limiting the spec

aweiteka:
> For the summary, my thinking is a link to the scanner UI for that specific report. That way scanners can be as rich as possible without artificially limiting the spec

Reference URL, compliance boolean and summary are all optional and may be used in any combination. Possible examples:

* compliance boolean with reference URL, no summary (e.g. maybe the data is too complex to summarize)
* summary only, no reference URL or compliance (e.g. OpenSCAP report with no parsing or report)
* all data (summary, reference URL, compliance) populated for best UX, IMHO.

aweiteka:
cc @andreasn

enoodle (approved):
Looks good to me

aweiteka:
This was merged in #4432. See #4461 for follow-up edits to address comments 

> You probably want to have a required field for the scanner version. Providers could have different versions with different capabilities. Free form string feels good there.

done #4461

>     I'm not clear on the timestamp format and would suggest both human readable and machine readable

done #4461

>     It might also be worthwhile to have an update timestamp unless the expectation is the timestamp is updated with each "rescan" or update in provider evaluation. If the latter, I'd be prescriptive about it

Valid point. Keeping this simple for now.

>     I would like to see a qualityType of policy. Providers with richer policy engines could then use their policy concepts to go beyond simple vulns or license issues.

done #4461

>     I would probably make the reference attribute a required item. Without it there's no way to validate the data.

done #4461

>     Suggest changing the description of summary to summary of issues found. Maps better to the non-vuln options

done #4461

>     Suggest changing score to something like items or count. To me score has implications which aren't covered in this spec.

done #4461. I went with "data" as type "string" since we have a use case for non-integer scoring data.

PR #4208, opened 2017-04-20
Bug 1437317, Added Deployment Type information

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1437317

### DIFF
--- a/install_config/install/advanced_install.adoc
+++ b/install_config/install/advanced_install.adoc
@@ -631,6 +631,22 @@ openshift_master_named_certificates=[{"certfile": "/root/STAR.openshift.com.crt"
 openshift_master_overwrite_named_certificates=true
 ----
 
+[[advanced-install-deployment-types]]
+=== Configuring Deployment Type
+
+Various defaults used throughout the playbooks and roles in this repository are
+set based on the deployment type configuration (usually defined in an Ansible
+hosts file).
+
+ifdef::openshift-enterprise[]
+Ensure the `deployment_type` parameter in your inventory file is set to
+`openshift-enterprise`.
+endif::[]
+ifdef::openshift-origin[]
+Ensure the `deployment_type` parameter in your inventory file is set to
+`origin`.
+endif::[]
+
 [[advanced-install-cluster-metrics]]
 == Configuring Cluster Metrics
 


ahardin-rh:
@php-coder PTAL

php-coder:
Because I'm not so close to the installer, I'd like ask @sosiouxme and @rhcarvalho (based on their updates in https://github.com/openshift/openshift-ansible/pull/3734 and https://github.com/openshift/openshift-ansible/pull/3799) to review this changes.

install_config/install/advanced_install.adoc
```diff
@@ -626,6 +626,63 @@ openshift_master_named_certificates=[{"certfile": "/root/STAR.openshift.com.crt"
...
+|
+ifdef::openshift-enterprise[]
+`atomic-openshift`
+endif::[]
+ifdef::openshift-origin[]
+`origin`
+endif::[]
+
+|`openshift.common.config_base`
+|*_/etc/origin_*
```
rhcarvalho: Inconsistent formatting (\*\_ here, \` in other places)

install_config/install/advanced_install.adoc
```diff
@@ -626,6 +626,63 @@ openshift_master_named_certificates=[{"certfile": "/root/STAR.openshift.com.crt"
...
+ifdef::openshift-enterprise[]
+Ensure the `openshift_deployment_type` parameter in your inventory file is set to
+`openshift-enterprise`.
+endif::[]
+ifdef::openshift-origin[]
+Ensure the `openshift_deployment_type` parameter in your inventory file is set to
+`origin`.
+endif::[]
+
+.Defaults for Deployment Type
```
rhcarvalho: These variables (defined by the openshift_facts role) are possibly something internal to the installer, not something to be in public docs?
@sdodson

rhcarvalho (commented):
Question: would it make sense to document the rename `deployment_type` -> `openshift_deployment_type`?

FWIW, we still support the old name for backwards compatibility.

(/cc @sosiouxme)

install_config/install/advanced_install.adoc
sdodson: `deployment_type` is definitely expected to be used, it's required actually. I don't know that @sosiouxme's work has made it into a released version so I'd document `deployment_type` in any current documentation and we should switch to `openshift_deployment_type` in 3.6 docs.

install_config/install/advanced_install.adoc
```diff
@@ -626,6 +626,63 @@ openshift_master_named_certificates=[{"certfile": "/root/STAR.openshift.com.crt"
...
+|Image Streams
+|
+ifdef::openshift-enterprise[]
+`rhel`
+endif::[]
+ifdef::openshift-origin[]
+`centos`
+endif::[]
+|===
+
```
sdodson: I do wonder if we should forego all of these details. I wouldn't expect end users to care about the details beyond the fact that the deployment_type defines whether you get an Origin or OCP install. Thoughts?

rhcarvalho: That's what I meant with my [previous comment](https://github.com/openshift/openshift-docs/pull/4208#discussion_r112932287), though I wrote it in the header of this section :-)

I feel those are internal details to the installer code base. They are documented in:
https://github.com/openshift/openshift-ansible/blob/master/DEPLOYMENT_TYPES.md

And the only reason I see for it being there is to assist people developing Ansible roles / code to understand what effect `[openshift_]deployment_type` has on other variables.

install_config/install/advanced_install.adoc
rhcarvalho: `deployment_type` / `openshift_deployment_type` for sure. The "defaults" part is the point for discussion.

install_config/install/advanced_install.adoc
sdodson: For that level of detail they should be looking in openshift-ansible. I'd suggest we remove the whole `Defaults for Deployment Type` section.

ahardin-rh:
@sdodson @rhcarvalho Thanks for your feedback. I updated the PR based on your comments.

Should a create a new PR to address the name change to openshift_deployment_type for OCP 3.6? Thanks again!

rhcarvalho (approved):

rhcarvalho:
> Should a create a new PR to address the name change to openshift_deployment_type for OCP 3.6?

:+1: 

I see only 13 occurrences for `deployment_type` in master :)

sdodson:
@ahardin-rh yeah, that'd be great, thanks.

sdodson (approved):

ahardin-rh:
@sdodson @rhcarvalho Thanks! Name change for OCP 3.6 applied in https://github.com/openshift/openshift-docs/pull/4318

ahardin-rh:
[rev_history]
|xref:../install_config/install/advanced_install.adoc#install-config-install-advanced-install[Installing a Cluster -> Advanced Installation]
|Added a xref:../install_config/install/advanced_install.adoc#advanced-install-deployment-types[Configuring Deployment Type] section.
%

Issue #4210, opened 2017-04-21
`{product-title}` is not substituted in published docs for sdn_troubleshooting.


lpsantil:
**Affected Docs:**
https://github.com/openshift/openshift-docs/blob/master/admin_guide/sdn_troubleshooting.adoc
https://docs.openshift.com/container-platform/3.5/admin_guide/sdn_troubleshooting.html#debugging-steps
https://docs.openshift.com/container-platform/3.4/admin_guide/sdn_troubleshooting.html#debugging-steps

**Issue:**
`{product-title}` is not substituted in published docs.

**Workaround:**

**Resolution:**
Fix substitution.

ahardin-rh:
Thank you! Closing this, as the issue is being addressed in https://github.com/openshift/openshift-docs/pull/4212

PR #4212, opened 2017-04-21
Product name does not render in code blocks

ahardin-rh:
Addresses https://github.com/openshift/openshift-docs/issues/4210

Used OpenShift as the product name, as this code block appears in both Origin and OCP docs.


ahardin-rh:
Minor typo. No revision history needed.

PR #4214, opened 2017-04-21
Created a pointer to Storage section for Online

ahardin-rh:
@tiwillia FYI

- Updated the examples with storage values that are correct for Online
- Added a pointer to the Storage section for the Online docs so users can find restriction info


tiwillia:
Awesome, LGTM :+1:

ahardin-rh:
No revision history needed

PR #4221, opened 2017-04-23
Remove features the integration doesn't have

juanluisvaladas:
F5 router plugin doesn't have feature parity with ha proxy routers.
Doesn't support re-encrypt, doesn't support, iddle pods, etc.

### DIFF
--- a/architecture/additional_concepts/f5_big_ip.adoc
+++ b/architecture/additional_concepts/f5_big_ip.adoc
@@ -27,9 +27,7 @@ xref:../../architecture/core_concepts/routes.adoc#re-encryption-termination[re-e
 xref:../../architecture/core_concepts/routes.adoc#passthrough-termination[passthrough terminated] routes matching on HTTP
 vhost and request path.
 
-The F5 router has feature parity with the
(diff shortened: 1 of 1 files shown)


bfallonf:
Verified in the issue. Thanks @juanluisvaladas . I'll merge this and close that issue.

bfallonf:
Actually, @juanluisvaladas @rajatchopra , any idea when the info changed? As in, which version of the docs should this be in? I've labeled it 3.5 for now, but I'm guessing it goes back further?

juanluisvaladas:
I don't think it has ever been true. 3.0 haproxy routers already supported reencrypt routes.

bfallonf:
Sure. Thanks @juanluisvaladas . I'll label these for the above branches, and update the 3.0 and 3.1 branches directly.

Thanks for all this.

bfallonf:
[rev_history]
|xref:../architecture/additional_concepts/f5_big_ip.adoc#architecture-additional-concepts-f5-big-ip[Additional Concepts -> F5 BIG-IP® Router Plug-in]
|Removed information about integration with HAProxy router.
%

PR #4226, opened 2017-04-24
Removing Dev Preview FAQ

ahardin-rh:


ahardin-rh:
@nhr Can you please help me redirect this page to https://www.openshift.com/pricing/index.html ? Not sure what the process is for these types of requests.

cc @luciddreamz @aheslin

nhr:
It looks like this change will only affect the master branch, which is currently only published for OpenShift Origin (docs.openshift.org rather than docs.openshift.com). Do you need to port this change to other branches as well?

Ultimately I need to know the full list of distro/version combinations will need the redirect.

_topic_map.yml
```diff
@@ -98,8 +98,6 @@ Topics:
   - Name: Administering a Cluster
     File: dedicated_administrators
     Distros: openshift-dedicated
-  - Name: Developer Preview FAQ
-    File: devpreview_faq
     Distros: openshift-online
```
nhr: You need to remove this line as well.

nhr:
I just noticed that you intend this change for Online docs. You also need to remove the 'Distros' line per my review.

ahardin-rh:
@nhr Thank you! This is updated.

PR #4228, opened 2017-04-24
Adding in Online Known Issues + Updated Distro_Map

ahardin-rh:


ahardin-rh:
Discussed with Online team and they do not want to maintain release notes. They only want a topic called Known Issues. https://blog.openshift.com/tag/release/ is their equivalent of release notes.

ahardin-rh:
@tiwillia This is what I have so far. PTAL. 

@aheslin This is put in place instead of Release Notes

tiwillia:
@ahardin-rh is the page called "Release Notes"? I'm not sure I like that name, just because we won't be communicating releases to users, only known issues.

Also, the single bug we have on the list is not externally visible, have we OK'd it with anyone to reveal details of that bug in docs? Do we often include private bugs in release notes?

ahardin-rh:
@tiwillia Sorry, I wasn't clear. The page will be called Known Issues and will live under the About section. I updated the PR to remove the old (unbuilt) Release Notes page to get rid of any confusion. As for the described bug, this is the only one I was given to doc so far. However, if it is private, I can remove it.

tiwillia (approved):
:+1:

ahardin-rh:
@abhgupta Can you please confirm that this BZ should be public?

ahardin-rh:
Discussed with @abhgupta and removed the BZ# and link. The rest is approved.

ahardin-rh:
Today I realized that the About page does not build in the Customer Portal. I discussed this with Tim and as a workaround we are including Known Issues as a section under Release Notes, but Release Notes simply points to the blog and Known Issues section.

ahardin-rh:
Some adjustments made in https://github.com/openshift/openshift-docs/pull/4283

PR #4260, opened 2017-04-26
Improve 3.3 EgressNetworkPolicy docs after bugfix

danwinship:
https://github.com/openshift/ose/pull/716 fixes a longstanding bug in 3.3's EgressNetworkPolicy implementation (relative to 3.4 and later). This updates the docs to match; with the next 3.3 release, using the "0.0.0.0/32" form will result in a warning being logged (though it will still work. But we should encourage people to use the correct form).

I don't know for sure that the version number I indicated there is accurate; https://bugzilla.redhat.com/show_bug.cgi?id=1440886 will track when the fix is going out, I guess. The docs fix can wait until after the OCP fix goes out, I just wanted to file the PR before I forgot about it...

Also, I don't know if you care about using "ifdefs" properly here since this is an OCP-specific branch anyway...

### DIFF
--- a/admin_guide/managing_pods.adoc
+++ b/admin_guide/managing_pods.adoc
@@ -236,13 +236,16 @@ To configure egress policy:
             {
                 "type": "Deny",
                 "to": {
-                    "cidrSelector": "0.0.0.0/32"
+                    "cidrSelector": "0.0.0.0/0"<1>
                 }
             }
         ]
     }
 }
 ----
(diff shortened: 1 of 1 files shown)


danwinship:
Updated the text to reflect the actual version in which this was fixed (3.3.1.25) and re-pushed.

bfallonf:
Thanks @danwinship . This should be ok to merge and it'll appear in the 3.3 branch and no others. Sorry this one took a bit to merge.

PR #4261, opened 2017-04-26
review-dev_guide_for_online

sallyom:
WIP 
@ahardin-rh I'm going to have @Miciah (if you will please) look this over..

### DIFF
--- a/dev_guide/routes.adoc
+++ b/dev_guide/routes.adoc
@@ -38,13 +38,21 @@ image::console_create_route.png["Creating a Route Using the Web Console"]
 
 Using the CLI, the following example creates an unsecured route:
 
+ifndef::openshift-online[]
 ----
 $ oc expose svc/frontend --hostname=www.example.com
 ----
+endif::[]
+ifdef::openshift-online[]
+----
+$ oc expose svc/frontend
+----
+endif::[]
 
 The new route inherits the name from the service unless you specify one
 using the `--name` option.
 
+ifndef::openshift-online[]
 .YAML Definition of the Unsecured Route Created Above
 ====
 [source,yaml]
@@ -60,7 +68,26 @@ spec:
     name: frontend
 ----
 ====
+endif::[]
 
+ifdef::openshift-online[]
+.YAML Definition of the Unsecured Route Created Above
+====
+[source,yaml]
+----
+apiVersion: v1
+kind: Route
+metadata:
+  name: frontend
+spec:
+  to:
+    kind: Service
+    name: frontend
+----
+====
+endif::[]
+
+ifndef::openshift-online[]
 Unsecured routes are the default configuration, and are therefore the simplest
 to set up.
 However,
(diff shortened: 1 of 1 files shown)


dev_guide/routes.adoc
```diff
@@ -60,7 +61,35 @@ spec:
...
+.YAML Definition of the Unsecured Route Created Above
+====
+[source,yaml]
+----
+apiVersion: v1
+kind: Route
+metadata:
+  name: frontend
+spec:
+  host: frontend-projectname.{product-title} {product-version} hostname
```
Miciah: What's up with `.{product-title} {product-version} hostname`? Anyway, `host` should not be specified at all.

dev_guide/routes.adoc
```diff
@@ -60,7 +61,35 @@ spec:
     name: frontend
 ----
 ====
+endif::[]
+ifdef::openshift-online[]
```
Miciah: Could we used a finer-grained ifdef? It's just the command and yaml that need to be conditionalised.

dev_guide/routes.adoc
```diff
@@ -121,7 +150,30 @@ remove a passphrase from a keyfile, you can run:
...
+====
+link:https://en.wikipedia.org/wiki/Transport_Layer_Security[TLS] is the
+replacement of SSL for HTTPS and other encrypted protocols.
+====
+
+----
+$ oc create route edge --service=frontend
+----
+
+====
```
Miciah: Might as well include yaml for an edge route for Online too.

Miciah (changes requested):

dev_guide/routes.adoc
ahardin-rh: In code blocks, {product-title} and {product-version} will not render properly

sallyom:
@Miciah, I've followed your suggestions here, mind checking it again? Thanks.

dev_guide/routes.adoc
```diff
@@ -121,7 +148,47 @@ remove a passphrase from a keyfile, you can run:
...
+    name: frontend
+  tls:
+    termination: edge
+----
+====
+
+====
+endif::[]
 
+ifndef::openshift-online[]
```
Miciah: Would it be clearer to include this with the `ifndef::openshift-online[]` block again, without the intervening `ifdef::openshift-online[]` block?

dev_guide/routes.adoc
```diff
@@ -121,7 +148,47 @@ remove a passphrase from a keyfile, you can run:
...
 ----
+endif::[]
+ifdef::openshift-online[]
+Unsecured routes are the default configuration, and are therefore the simplest
+to set up.
+However,
+xref:../architecture/core_concepts/routes.adoc#secured-routes[secured routes]
+offer security for connections to remain private.  To create a secured HTTPS route
+encrypted with the default certificate for {product-title} {product-version}
+you can use the `create route` command.
```
Miciah: Only the last sentence of the paragraph varies between Online and non-Online, so the ifdeffing could be still a little more fine-grained, but I don't feel strongly about it.

Miciah (approved):
A couple modest suggestions, but looks fine.

ahardin-rh:
No revision history needed

PR #4269, opened 2017-04-26
Various changes for openshift online

tiwillia:
These changes mostly pertain to the developers guide, but there is also a small change in the architecture guide to enable the docs around stateful sets.


tiwillia:
@ahardin-rh fyi

This is probably my last PR, but I know some other folks on the team still have some pending. Let me know if you see any issues with this.

ahardin-rh:
Lots of nested conditionals. @adellape Do you mind giving this a second set of eyes? :eyeglasses:

adellape:
:+1:

PR #4277, opened 2017-04-27
Document new egress-router features

danwinship:
This updates the egress router documentation for
- https://github.com/openshift/origin/pull/13837 (multiple destinations)
- https://github.com/openshift/origin/pull/13586 (HTTP proxy mode)
- https://trello.com/c/mdbZRAgZ (initContainer mode)
- https://trello.com/c/DCDdY7eA (document how to use ConfigMap)

The new features haven't merged yet so the exact syntax in the examples might change and this shouldn't merge until then, but you can review for style / asciidoc issues /etc now.

### DIFF
--- a/admin_guide/managing_networking.adoc
+++ b/admin_guide/managing_networking.adoc
@@ -97,7 +97,7 @@ talk to.
 
 xref:admin-guide-limit-pod-access-egress-router[Router]::
 Using an egress router allows you to create identifiable services to send
-traffic to a specific destination, ensuring an external destination treats
+traffic to certain destinations, ensuring those external destinations treat
 traffic as though it were coming from a known source. This helps with security,
 because it allows you to secure an external database so that only specific pods
 in a namespace can talk to a service (the egress router), which proxies the
@@ -270,12 +270,29 @@ Virtual Switch Ports] and
 link:http://pubs.vmware.com/vsphere-4-esxi-installable-vcenter/index.jsp?topic=/com.vmware.vsphere.esxi_server_config.doc_40_u1/esx_server_config/securing_an_esx_configuration/c_forged_transmissions.html[Forged
 Transmissions] guidance.
 
-[[admin-guide-deploying-an-egress-router-pod]]
-==== Deploying an Egress Router Pod
+[[admin-guide-egress-router-modes]]
+==== Egress Router Modes
 
-. Create a pod configuration using the following:
+The egress router can run in two different modes:
+
+1. A "redirect" mode, where the egress router sets up iptables rules
+to redirect traffic from its own IP address to one or more destination
+IP addresses. Client pods that want to make use of the reserved source
+IP address will need to be modified to connect to the egress router
+rather than connecting directly to the destination IP.
+
+2. An HTTP proxy mode, where the egress router runs as an HTTP proxy.
+This only works for clients talking to HTTP/HTTPS-based services, but
+usually requires fewer changes to the client pods to get them to work.
+(Many programs can be told to use an HTTP proxy just by setting an
+environment variable.)
+
+[[admin-guide-deploying-an-egress-router-pod]]
+==== Deploying a Redirecting Egress Router Pod
+
+. Create a pod configuration such as the following:
 +
-.Example Pod Definition for an Egress Router
+.Example Pod Definition for a simple redirecting Egress Router
 ====
 ----
 apiVersion: v1
@@ -287,7 +304,7 @@ metadata:
   annotations:
     pod.network.openshift.io/assign-macvlan: "true" <1>
 spec:
-  containers:
+  initContainers:
   - name: egress-router
 ifdef::openshift-enterprise[]
     image: registry.access.redhat.com/openshift3/ose-egress-router
@@ -304,8 +321,18 @@ endif::openshift-origin[]
       value: 192.168.12.1
     - name: EGRESS_DESTINATION <4>
       value: 203.0.113.25
+    - name: EGRESS_ROUTER_MODE <5>
+      value: init
+  containers:
+  - name: egress-router-wait
+ifdef::openshift-enterprise[]
+    image: registry.access.redhat.com/openshift3/ose-pod
+endif::openshift-enterprise[]
+ifdef::openshift-origin[]
+    image: openshift/origin-pod
+endif::openshift-origin[]
   nodeSelector:
-    site: springfield-1 <5>
+    site: springfield-1 <6>
 ----
 <1> The `pod.network.openshift.io/assign-macvlan annotation` creates a Macvlan
 network interface on the primary network interface, and then moves it into the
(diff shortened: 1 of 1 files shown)


weliang1 (changes requested):
Better change "Deploying an Egress Router Pod" to "Deploying an Egress Router Redirect Pod"

weliang1 (commented):
"This only works for clients talking to HTTP-based services,"
Should it works for clients talking to HTTPs-based services too?

danwinship:
Repushed without the HTTP proxy docs since that didn't make it in for this devcut. But everything else is merged and this PR is ready to go.

admin_guide/managing_networking.adoc
```diff
@@ -378,6 +401,131 @@ If only some of the nodes in your cluster are capable of claiming the specified
...
+  containers:
+  - name: egress-router-wait
+ifdef::openshift-enterprise[]
+    image: registry.access.redhat.com/openshift3/ose-pod
+endif::openshift-enterprise[]
+ifdef::openshift-origin[]
+    image: openshift/origin-pod
+endif::openshift-origin[]
+----
+<1> This uses the YAML syntax for a multi-line string. In this
```
bmeng: Should we note that only tcp and udp are supported here?

danwinship: updated to clarify the syntax better

vikram-redhat:
[Document: Add an HTTP proxy mode for egress-router](https://trello.com/c/I6Is4843/490-document-add-an-http-proxy-mode-for-egress-router)

danwinship:
re-pushed with the HTTP proxy stuff added back again since that has now merged

bfallonf:
Thanks @danwinship . I'll merge this and do a followup. I'll ping you with any questions if I have them.

knobunc (commented):
LGTM thanks Dan

weliang1 (commented):
We should describe below configuration in the pod to make feature works:
`export http_proxy=http://my-egress-router-service:8080/ `, and `export https_proxy=http://my-egress-router-service:8080/ `

bfallonf:
@weliang1 Sorry, I'm not sure where you're asking to place this. Should this be part of the "Deploying an Egress Router Pod in Redirect Mode" procedure? After the pod is created and before the verification step? In this PR, I mean around line 365.

danwinship:
@bfallonf it should probably go around where it talks about configuring clients to use an http proxy by setting an environment variable.

There are actually two aspects to this: first, we don't mention anywhere that the proxy runs on port 8080. The best way to address that might be to provide an example of a service definition to go with the egress http proxy pod definition. (It would be similar to the service definition in the redirecting egress router case, but needs to specify targetPort 8080 always.)

Second, we ought to give a better hint on how to configure clients to use it. What Weibin suggests isn't correct for *all* clients, but it will work in most cases. We might want to show an example of setting those variables in the `env` section of a client pod.

bfallonf:
@weliang1 @danwinship I think I understand. Thanks for the clarification. I've made a followup to this PR, so it would be great for some feedback there. I'll tag the relevant parts.

admin_guide/managing_networking.adoc
```diff
@@ -378,6 +418,227 @@ If only some of the nodes in your cluster are capable of claiming the specified
...
+      valueFrom:
+        configMapKeyRef:
+          name: egress-routes
+          key: destination
+    - name: EGRESS_ROUTER_MODE
+      value: init
+    ...
+----
+
+Note that the egress router will not automatically update when the
```
sferich888: @derekwaynecarr @bparees not sure who to ask on this but is this a bug (or do I just not understand how configmap and DC/RC's work)? Shouldn't DC/RC's be update with config maps referenced by them change?

admin_guide/managing_networking.adoc
```diff
@@ -378,6 +418,227 @@ If only some of the nodes in your cluster are capable of claiming the specified
...
+ConfigMap changes; you will need to restart the pod to get updates.
+
+[[admin-guide-deploying-an-egress-router-http-proxy-pod]]
+==== Deploying an Egress Router HTTP Proxy Pod
+
+Creating an egress router in HTTP proxy mode is similar to creating a
+redirecting egress router, as described above, but with a slightly
+different specification:
+
+.Example Pod Definition for an Egress Router HTTP Proxy
```
sferich888: Should this say 'HTTP/HTTPS'

admin_guide/managing_networking.adoc
```diff
@@ -378,6 +418,227 @@ If only some of the nodes in your cluster are capable of claiming the specified
...
+`EGRESS_HTTP_PROXY_DESTINATION`. The simplest is to just specify `*`,
+meaning "allow connections to all remote destinations". Other than
+that, each line specifies one group of connections to allow or deny:
+
+- An IP address (eg, `192.168.1.1`) will allow connections to that IP address.
+- A CIDR range (eg, `192.168.1.0/24`) will allow connections to that CIDR range.
+- A hostname (eg, `www.example.com`) will allow proxying to that host.
+- A domain name preceded by `\*.` (eg, `*.example.com`) will allow proxying to that domain and all of its subdomains.
+- A `!` followed by any of the above will deny connections rather than allowing them
+- If the last line is `*`, then anything that hasn't been denied will be allowed. Otherwise, anything that hasn't been allowed will be denied.
```
sferich888: I don't see in here where we talk about how https://bugzilla.redhat.com/show_bug.cgi?id=1435803 is addressed by such a configuration. 

Is the SSL offloaded (to the endpoint) or are we terminating at the egress pod, and re-connecting? If its the latter, what CA and Server certificate is the proxy using?

sferich888 (commented):
@aheslin @vikram-redhat while this is merged, can we get someone to looked at the comments here and work with @danwinship to fill in any missing details.

admin_guide/managing_networking.adoc
bparees: @pmorie should know the details of that.  The DC/RC certainly won't be, but a pod that mounts in a configmap as a volume, or that consumes a configmap as env, is, I think, supposed to get restarted if the configmap changes.  But i can't remember, that might be wishful thinking.

admin_guide/managing_networking.adoc
knobunc: We could... people usually just call them HTTP proxies, even if they handle both.  But there is no harm in making that change.

admin_guide/managing_networking.adoc
knobunc: When you are using a forward HTTP proxy, the client has to be aware of that and talk to the proxy differently.  https://en.wikipedia.org/wiki/HTTP_tunnel#HTTP_CONNECT_tunneling

Basically, the client asks the proxy to open a connection over HTTP and then the client sends the HTTPS traffic through a tunneled connection.

danwinship: (This was clarified over email, but rhbz 1435803 covered two different features, only one of which was implemented in 3.6 and documented here.)

PR #4281, opened 2017-04-27
Removing several pages from Online per content audit

ahardin-rh:
@tiwillia This removes the pages that are identified as not-applicable to Online in the content audit conducted by the Online team.

@aheslin FYI

### DIFF
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -747,6 +747,7 @@ Topics:
     File: configmaps
   - Name: Using Daemonsets
     File: daemonsets
+    Distros: openshift-enterprise,openshift-dedicated,openshift-origin
   - Name: Pod Autoscaling
     File: pod_autoscaling
   - Name: Managing Volumes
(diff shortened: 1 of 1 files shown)


ahardin-rh:
@nhr Can you please help me with redirects here as well? These pages are only being removed from Online and we will need to just redirect to the index page https://docs.openshift.com/online/welcome/index.html

nhr:
You may already know this, but just to be sure: This pull request is against master, which is good, but in order for it to have an effect on online, it also has to be made against the online branch.

That said, I am seeing the following URLs which will need redirects:

* https://docs.openshift.com/online/dev_guide/daemonsets.html
* https://docs.openshift.com/online/dev_guide/jobs.html
* https://docs.openshift.com/online/creating_images/custom.html

But your diff also includes changes to the "Cron Jobs" topic visibility. That's fine for the master branch, but in online that link doesn't exist.

If your plan is to merge master onto the online branch soon, we're fine. Otherwise, it is called "Scheduled Jobs" on the online branch, so your diff for the online branch will need to be slightly different as well. Based on that I will also set up a redirect for:

* https://docs.openshift.com/online/dev_guide/scheduled_jobs.html

Please confirm, and tag me to review the companion PR when you have it ready for the online branch (unless the plan is to replace the online branch with a fresh merge over from master)

PR #4292, opened 2017-05-01
Mark masters as schedulable when they're infra

sdodson:
Somewhere along the way the installer started marking masters as
unschedulable by default but the documentation was not updated to
reflect that. Since all of our examples use the masters as infra nodes
this means that the router and registry won't deploy properly so mark
them as schedulable. We may want to simply move to listing dedicated
infra nodes to simplify this.

### DIFF
--- a/install_config/install/advanced_install.adoc
+++ b/install_config/install/advanced_install.adoc
@@ -507,7 +507,7 @@ environment has the 'region=infra' label.
 ====
 
 [[marking-masters-as-unschedulable-nodes]]
-=== Marking Masters as Unschedulable Nodes
+=== Marking Masters as Schedulable Nodes
 
 Any hosts you designate as masters during the installation process should also
 be configured as nodes by adding them to the *[nodes]* section so that the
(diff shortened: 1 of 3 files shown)


gabemontero (approved):

install_config/install/advanced_install.adoc
```diff
@@ -516,14 +516,16 @@ xref:../../architecture/additional_concepts/networking.adoc#openshift-sdn[{produ
...
 However, in order to ensure that your masters are not burdened with running
-pods, you can make them
+pods, they are by default marked unschedulable.
 xref:../../admin_guide/manage_nodes.adoc#marking-nodes-as-unschedulable-or-schedulable[unschedulable]
-by adding the `*openshift_schedulable=false*` option any node that is also a
-master. For example:
+
+The examples in this document use masters as 'infra' nodes and have been marked
+schedulable ensuring the router and registry are deployed properly. In a production
+environment you should have dedicated infra nodes. For example:
```
sdodson: Eh, maybe I should move the 'For example:' bit around, this reads really poorly.

gabemontero: Good point - I missed this piece earlier.

Yeah, this is going to get tricky though with the other side of the coin from my recent attempts, with the current behavior of trying to define routers on each infra nodes (i.e. router replica count == number of infra nodes) and that failing when the number of desired replicas are not defined (since that seems to be the requirment vs. a best effort "we'll place as many router replicas as we can").

Unless that behavior is mitigated such that the "best effort route placement expectation" returns, I would think the docs have to reconcile
a) let's not burden the masters with running pods in production ... where you then add the router label key/value setting to something other than infra
b) for pre-production, and as these samples below show, you mark the infra nodes as schedulable

I think then both options need to be provided, based on the production vs. non-production distinction.

gabemontero (changes requested):

gabemontero:
Sorry ... not sure what I did to have my comment post twice :-)

adellape:
Closing in favor of https://github.com/openshift/openshift-docs/pull/4317.

PR #4312, opened 2017-05-02
Updated drop-down menu for Online

ahardin-rh:


luciddreamz:
Looks good. You *might* want to flip the order so Online 3 is above Online 2.

ahardin-rh:
@luciddreamz Good call. Updated!

PR #4322, opened 2017-05-03
Added note to restart after modifying identity provider

dutchiechris:
As a new user I modified the identity provider but had no idea how to make it take effect.  Luckily I found this in some training on the web but it should be in these instructions too.


ahardin-rh:
Thank you!

ahardin-rh:
[rev_history]
|xref:../install_config/configuring_authentication.adoc#install-config-configuring-authentication[Configuring Authentication and User Agent]
|Noted that after making changes to an identity provider, you must restart the master service for the changes to take effect.
%

Issue #4348, opened 2017-05-05
Replace oc update by oc replace in examples


php-coder:
I see that at least in one place (https://docs.openshift.com/online/cli_reference/basic_cli_operations.html#update) we're talking about `oc update` which is deprecated:
```console
$ oc update
W0430 17:49:23.614487   18313 cmd.go:372] update is DEPRECATED and will be removed in a future version. Use replace instead.
error: Must specify --filename to replace
See 'oc replace -h' for help and examples.
```

ahardin-rh:
@php-coder Thanks! Do you know when it was deprecated? Would it be correct to remove it from OCP 3.3 and later?

php-coder:
As I can see it's deprecated in Kubernetes from Jun 27 2015 (see: https://github.com/kubernetes/kubernetes/commit/9b3d42c090aeab663bf42e38658a5f88fb4b5272). So, AFIU all our versions are affected.

ahardin-rh:
@php-coder okay, thanks!

ahardin-rh:
https://github.com/openshift/openshift-docs/issues/4350 addresses this

PR #4350, opened 2017-05-05
oc update is now oc replace

ahardin-rh:
This addresses https://github.com/openshift/openshift-docs/issues/4348


ahardin-rh:
No revision history needed

PR #4364, opened 2017-05-08
Online Arch Guide link fix

vikram-redhat:


vikram-redhat:
I have cherry picked it to Online so didn't add that label.

PR #4377, opened 2017-05-09
put in a missing )

goern:
massive addition of a )


ahardin-rh:
@goern Thanks! 

Looks like this is only in a typo in the Origin docs, so no labels or milestone need to be set.

ahardin-rh:
No revision history needed

PR #4389, opened 2017-05-09
Update iptables document for firewalld

danwinship:
@openshift/networking please review for content

### DIFF
--- a/admin_guide/iptables.adoc
+++ b/admin_guide/iptables.adoc
@@ -11,57 +11,65 @@
 toc::[]
 
 == Overview
-There are many system components including {product-title}, containers, and
-software that manage local firewall policies that rely on the kernel iptables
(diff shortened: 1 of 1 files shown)


danwinship:
This applies to master and 3.5

pecameron (commented):
LGTM

bfallonf:
Thanks @danwinship . Let me know if there's any more content you'd like to add here or if it's ok to move forward with this. I can followup with this as part of BZ 1446436.

danwinship:
As I said in email, we might want some additions to advanced_install too, but that could happen as a separate PR (and those details would come from the installer team anyway). So I think you can move forward with this.

bfallonf:
@danwinship Sure thing. I'll merge this and do a followup PR, but also push for feedback, as we discussed over email. Thanks!

Issue #4378, opened 2017-05-09
[online] update custom route hosts restriction


jiri-fiala:
The custom route host template restriction for OpenShift Online needs to be updated; it's still valid for the *Starter* plan, though [there's](https://github.com/openshift/openshift-docs/blame/1912bf58d1e88804816153e21def67a4b739bf5e/dev_guide/routes.adoc#L227) the old Dev Preview template `<route-name>-<namespace>.44fs.preview.openshiftapps.com`, which should be replaced with `<route-name>-<namespace>.1d35.starter-us-east-1.openshiftapps.com`, or removed.

vikram-redhat:
Fixed via: https://github.com/openshift/openshift-docs/pull/4396. Will be in the release next week. Thanks for reporting this @jiri-fiala .

PR #4396, opened 2017-05-10
GH-Issue: 4378: Fixing custom route hostname

vikram-redhat:


vikram-redhat:
This only applies to the online branch. No rev history required. GH: https://github.com/openshift/openshift-docs/issues/4378

PR #4397, opened 2017-05-11
BZ#1449804-Added a note that flannel is only supported for OCP on OSP

gaurav-nelson:
https://bugzilla.redhat.com/show_bug.cgi?id=1449804


ahardin-rh:
Does this need a revision history?

gaurav-nelson:
@ahardin-rh It is very small change and I think it does not need one.

PR #4413, opened 2017-05-12
Fix default metrics_resolution language

lpsantil:
PR for https://github.com/openshift/openshift-docs/issues/4412

PR #4414, opened 2017-05-12
Fix the math

lpsantil:


ahardin-rh:
Looks like this was also addressed in https://github.com/openshift/openshift-docs/pull/4413. If you can submit that PR against master and squash the commits into one commit, we can cherry-pick and publish the changes to the relevant branches. Thanks and sorry for any confusion!

PR #4429, opened 2017-05-16
Correct playbook path for Metrics deployment or Upgrade

shah-zobair:
The playbook path for Metrics deployment or Upgrade should be /usr/share/ansible/openshift-ansible/playbooks/byo/openshift-cluster/openshift-metrics.yml instead of /usr/share/ansible/openshift-ansible/playbooks/common/openshift-cluster/openshift_metrics.yml

### DIFF
--- a/install_config/cluster_metrics.adoc
+++ b/install_config/cluster_metrics.adoc
@@ -502,7 +502,7 @@ The following command sets the Hawkular Metrics route to use
 You must have a persistent volume of sufficient size available.
 
 ----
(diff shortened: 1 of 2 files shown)


ahardin-rh:
@shah-zobair Thank you! Is this relevant for OCP 3.5 only, or earlier versions as well? Thanks again!

shah-zobair:
It's relevant to OCP 3.5 only as previously we used manual upgrade instead
of Ansible playbook.

ahardin-rh:
@shah-zobair Thanks again!

ahardin-rh:
No revision history needed

Issue #4431, opened 2017-05-16
Docs for OCP should point to supported integration matrix


etsauer:
Something like the following should be linked/discoverable from docs.openshift.com. Perhaps in the release notes?

https://access.redhat.com/articles/2176281

bfallonf:
Thanks @etsauer . I've added this to the list of questions to ask yourself when installing and created a PR:

https://github.com/openshift/openshift-docs/pull/4438

Let me know if there's any thoughts. I'll close this issue.

PR #4438, opened 2017-05-18
Issue 4431 Added supported integrations Q to planning

bfallonf:
As per #4431 

@adellape @ahardin-rh @bmcelvee @mburke5678 @gaurav-nelson Peer review plz?

### DIFF
--- a/install_config/install/planning.adoc
+++ b/install_config/install/planning.adoc
@@ -39,6 +39,8 @@ example as a basis for your environment.
  environment, but you might have a preference for a particular method of
  installing, managing, and updating your services.
 
+* _Is my installation supported if integrating with other technologies?_ See the link:https://access.redhat.com/articles/2176281[OpenShift Container Platform Tested Integrations] for a list of tested integrations.
+
 [[installation-methods]]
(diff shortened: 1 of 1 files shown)


ahardin-rh:
LGTM :star2:

mburke5678:
LGTM

etsauer:
LGTM

bfallonf:
Thanks all! I'll merge.

adellape:
@bfallonf Wondering if this whole line should be ifdef'd for just OCP?

bfallonf:
@adellape I guess it depends on if there's a difference between what would work on OCP and what would work on Origin? Or are you asking if this should be in the Origin docs, considering it's upstream?

adellape:
@bfallonf Mainly just that the KBase talks about OCP and its version numbers (3.x) and doesn't mention Origin by name or its versions. Also that the KBase talks about tested/supported things (suggesting GSS), when they don't support Origin.

bfallonf:
@adellape Hmm true. OK, I'll do a followup PR.

PR #4444, opened 2017-05-19
Fixed Defaults for Ceph

this-tech:
The defaults should be used in the example because they work OOB


ahardin-rh:
@donnydavis Thanks! Is this also for OCP 3.5?

donnydavis:
Yes it is also for 3.5

ahardin-rh:
No revision history needed

Issue #4454, opened 2017-05-22
Image registry url in Online docs is incorrect


siamaksade:
The integrated image registry url in the OpenShift Online docs refers to the "preview" url and needs to get updated to the url after GA:
https://docs.openshift.com/online/dev_guide/managing_images.html#accessing-the-internal-registry

Currently
```
$ docker login -u <user_name> -e <email_address> \
    -p <token_value> https://registry.preview.openshift.com
```

Should be
```
$ docker login -u <user_name> -e <email_address> \
    -p <token_value> https://registry.starter-us-east-1.openshift.com
```

ahardin-rh:
Addressed via https://github.com/openshift/openshift-docs/pull/4455 and the link is also fixed on the web console after discussion with the Online team.

PR #4455, opened 2017-05-22
Updated URL

ahardin-rh:
Addresses https://github.com/openshift/openshift-docs/issues/4454

### DIFF
--- a/dev_guide/managing_images.adoc
+++ b/dev_guide/managing_images.adoc
@@ -362,9 +362,8 @@ stream by manually pushing an image], or just to `docker pull` an image
 directly.
 
 ifdef::openshift-online[]
-{product-title} (Next Gen) Developer Preview gives developers a hands-on preview
-of the OpenShift platform in a hosted environment that includes access to an
-internal registry located at *_registry.preview.openshift.com_*.
+OpenShift Online gives developers a hands-on preview of the OpenShift platform
+in a hosted environment that includes access to an internal registry.
 endif::[]
 
 The internal registry authenticates using the same
@@ -396,7 +395,7 @@ ifdef::atomic-registry,openshift-origin,openshift-enterprise,openshift-dedicated
     -p <token_value> <registry_server>:<port>
 endif::[]
 ifdef::openshift-online[]
-    -p <token_value> https://registry.preview.openshift.com
+    -p <token_value> https://registry.<clusterID>.openshift.com
 endif::[]
 ----
 +


ahardin-rh:
@tiwillia PTAL

tiwillia:
@ahardin-rh We might just not want to include a link here. There are several free-tier clusters, meaning there are several internal registries. The one linked is just one of those.

ahardin-rh:
@tiwillia Thanks! Updated with a replaceable instead.

tiwillia:
@ahardin-rh help me out here, whats a replaceable? Do we have a way to replace it with the right URL depending on which cluster the user is a part of?

ahardin-rh:
@tiwillia The example is just updated to not be specific (<registry_URL>), since the path can change. Would the user already know the URL, or would they have to look it up somewhere? If they have to look it up, we can provide guidance here as to how they can do that.

siamaksade:
@ahardin-rh  I don't know of any other way to retrieve the registry URL. Users don't have access to list routes in the _default_ namespace. It might actually make sense to have it somewhere within the web console on e.g. About page.

Alternatively if registry URLs follow a pattern across online clusters, we could specify the pattern: `registry.<plan>-<region>.openshift.com`

We can document then that the `<plan>-<region>` is the same as the one they see in the master URL.

tiwillia:
@abhgupta and I yesterday discussed the possibility of adding this URL into the console `About` page, as mentioned above. The current plan is to add the URL there: https://trello.com/c/zztJk7Lk/787-2-add-registry-url-in-about-cli-page-in-console

Perhaps we can hold off on setting the URL in the docs and instead document the existence of the URL in the console, once its ready?

abhgupta:
@ahardin-rh @timothyh @siamaksade Here are my thoughts.

1. Add a link to the registry URL on the console "About" or "Command Line Tools" page
2. Add documentation to provide the URL format for the registry (with a replaceable for cluster ID) --> `registry.<clusterID>.openshift.com`
3. Perhaps, in the docs, provide a list of cluster region names to cluster ID mappings (eg: Starter US East 1 --> starter-us-east-1)

tiwillia:
@abhgupta that sounds great. We can probably move this PR to complete # 2 in your list for now.

ahardin-rh:
@tiwillia @abhgupta @siamaksade Thanks! I updated the PR, satisfying #2 on @abhgupta's list. 

#3 Is something worth discussing with @aheslin

tiwillia:
LGTM 👍

ahardin-rh:
No revision history needed

PR #4469, opened 2017-05-24
Fix metrics variable names

alikhajeh1:
`request` should be `requests` per https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_metrics/defaults/main.yaml#L9-L10


ahardin-rh:
Thank you!

ahardin-rh:
No revision history needed

Issue #4495, opened 2017-05-26
Nuage SDN doc could do with a tidy up


nstrug:
The Nuage SDN documentation e.g. https://docs.openshift.com/container-platform/3.5/install_config/configuring_nuagesdn.html could do with a tidy up. The page does not explain what Nuage is or why I would want or need to use it in preference to the default SDN. The page also uses undefined and unlinked initialisms (VSP, VSD). In general, the page seems to be written with a target audience that is already very familiar with Nuage in mind, which I think is an incorrect assumption.

ahardin-rh:
@nstrug Thanks! I will start a PR soon to make some improvements. I will try to locate a Nuage SDN expert to provide some guidance.

PR #4497, opened 2017-05-26
Improve Nuage SDN topic

ahardin-rh:
Addresses https://github.com/openshift/openshift-docs/issues/4495


ahardin-rh:
Added minor details; no revision history needed

PR #4498, opened 2017-05-30
added more info to horizontal autoscaler docs

bfallonf:
As per: https://trello.com/c/7vxgSTxF/365-document-custom-metrics-extensible-metric-driven-hpa

@DirectXMan12 I'll tag you in trello, but I'm thinking that all that needs to be added to this card is objects into the table, then some rewriting to get away from the "only CPU utilization" phrase. Let me know if you disagree.

### DIFF
--- a/dev_guide/pod_autoscaling.adoc
+++ b/dev_guide/pod_autoscaling.adoc
@@ -124,36 +124,46 @@ spec:
 
 Use the `oc autoscale` command and specify the maximum number of pods to run at
 any given time. Optionally, you can specify the minimum number of pods and the
-average CPU utilization your pods should target as well, otherwise those are
+average memory utilization your pods should target as well, otherwise those are
 given default values from the {product-title} server.
 
-For example:
+. Enable the memory-based autoscaling by configuring the <file> file:
++
+----
+...
+apiServerArguments:
+  runtime-config:
+  - apis/autoscaling/v2alpha1=true
+...
+----
 
+. Specify the minimum number of pods and the average memory utilization your
+pods should target:
++
 ----
 $ oc autoscale dc/frontend --min 1 --max 10 /
-  --cpu-percent=80 deploymentconfig "hpa-resource-metrics-cpu" autoscaled
+  --cpu-percent=80 deploymentconfig "hpa-resource-metrics-memory" autoscaled
 ----
-
-The above example creates a horizontal pod autoscaler with the following
-definition:
-
++
+This command creates a horizontal pod autoscaler with the following definition:
++
 [source,yaml,options="nowrap"]
 ----
 apiVersion: autoscaling/v2alpha1
 kind: HorizontalPodAutoscaler
 metadata:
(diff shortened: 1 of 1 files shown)


DirectXMan12:
Yeah, this is a bit tricky, because anything beyond CPU utilization requires using something that's alpha in some capacity (even memory -- either autoscaling/v2alpha1 or the corresponding alpha annotations), so I'm not sure how we want to document that.

bfallonf:
@DirectXMan12 We can attach a Technology Preview box to it. We've done this in other topics:

https://docs.openshift.com/container-platform/3.5/admin_guide/managing_networking.html#admin-guide-networking-multicast

Can just say something like what you mention "anything other than CPU utilization is in technology preview". What other objects can be added though? I can add that to the table I've added to here and see how it looks if we like.

DirectXMan12:
did you take a look at https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/ and https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough/ ?  They describe how the new objects work.  The problem is that we don't have the adapter for non-resource-metrics, but that's how the HPA is structured.

bfallonf:
@DirectXMan12 Thanks for the links. I can see how it works, but not too sure what other objects to that table (where, so far, I've added memory). Let me know if you have any hints.

But also, from what you're saying above, should this be documented just yet? Should this be put aside until you've worked on the adapter? Or is that for this current sprint?

DirectXMan12:
> But also, from what you're saying above, should this be documented just yet? But also, from what you're saying above, should this be documented just yet?

Yeah, I'm not entirely certain when that work will be ready to be documented.  We might want to put it aside for the moment (plus, we haven't actually enabled a switch to make the HPA controller call out to the custom metrics APIs yet).

DirectXMan12 (changes requested):
The example in the memory section uses CPU, so that needs to be fixed.  Additionally, we've disabled the HPA v2 API by default, so users will have to opt-in to use it in order to get memory-based autoscaling.

DirectXMan12:
the API should be enablable using the `runtime-config` option under `apiServerArguments`:

```
...
apiServerArguments:
  runtime-config:
  - apis/autoscaling/v2alpha1=true
```

bfallonf:
Thanks @DirectXMan12 . I'm not sure what you're after exactly. Which file is this being added to? And I replaced the "cpu" bits in the procedure with "memory" but should the "--cpu-percent=80" option for the command change as well?

DirectXMan12:
> Thanks @DirectXMan12 . I'm not sure what you're after exactly. Which file is this being added to?

It's a config option in master-config.yaml.

> but should the "--cpu-percent=80" option for the command change as well?

You can't create the autoscaler through a command if you want to do just memory autoscaling.  You have to `kubectl create -f` it.

bfallonf:
Closing this PR in favour of #4817

PR #4502, opened 2017-05-30
Correct link to default inventory file for aggregated logging

ericlake:
The existing link points to the metrics project.


ahardin-rh:
Looks like the link now points to https://docs.openshift.org/latest/install_config/install/advanced_install.html#configuring-ansible

Closing this, but please let me know if we need to reopen it. Thanks!

PR #4503, opened 2017-05-30
Trello: Added topic for projected volumes

mburke5678:
https://trello.com/c/AfFZsOyx/418-document-downward-api-all-in-one-volume-for-secrets-configmap-downward-api

### DIFF
--- a/dev_guide/projected_volumes.adoc
+++ b/dev_guide/projected_volumes.adoc
@@ -1,0 +1,361 @@
+[[dev-guide-projected-volumes]]
+= Projected Volumes
+{product-author}
+{product-version}
+:data-uri:
+:icons:
+:experimental:
+:toc: macro
+:toc-title:
+
+toc::[]
+
+== Overview
+
+A _projected volume_ maps several existing xref:../dev_guide/volumes.adoc#dev-guide-volumes[volume sources] into the same directory.
+
+Currently, the following types of volume sources can be projected:
+
+* xref:secrets.adoc#dev-guide-secrets[Secrets]
+* xref:configmaps.adoc#dev-guide-configmaps[Config Maps]
+* xref:downward_api.adoc#dev-guide-downward-api[Downward API]
+
+[NOTE]
+====
+All sources are required to be in the same namespace as the pod.
+====
+
+Projected volumes can map any combination of these volume sources into a single directory, allowing the user to:
+
+* automatically populate a single volume with the keys from multiple secrets, configmaps, and with downward API information,
+so that I can synthesize a single directory with various sources of information;
+* populate a single volume with the keys from multiple secrets, configmaps, and with downward API information,
+explicitly specifying paths for each item, so that I can have full control over the contents of that volume.
+
+
+== Example Scenarios
+// https://github.com/kubernetes/community/blob/master/contributors/design-proposals/all-in-one-volume.md
+
+The following general scenarios show how you can use projected volumes.
+
+* *ConfigMap, Secrets, Downward API.* Projected volumes allow you to to deploy containers with configuration data that includes passwords.
+An application using these resources could be deploying OpenStack on Kubernetes. The configuration data may need to be assembled differently depending on if the services are going to be used for production or for testing. If a pod is labeled with production or testing, the downward API selector `metadata.labels` can be used to produce the correct OpenStack configs.
+
+* *ConfigMap + Secrets.* Projected volumes allow you to deploy containers involving configuration data and passwords.
+For example, you might execute an Ansible playbook stored as a configmap, with some sensitive encrypted tasks that are decrypted using a vault password file.
+
+* *ConfigMap + Downward API.* Projected volumes allow you to generate a config including the pod name (available via the `metadata.name` selector). This application can then pass the pod name along with requests in order to easily determine the source without using IP tracking.
+
+* *Secrets + Downward API.* Projected volumes allow you to use a secret as a public key to encrypt the namespace of the pod (available via the `metadata.namespace` selector).
+This example allows the operator to use the application to deliver the namespace information securely without using an encrypted transport.
+
+== Example Pod Specifications
+
+The following are examples of pod specifications for creating projected volumes.
+
+[[dev-guide-projected-volumes-example]]
+.Pod with a secret, a downward API, and a configmap
+====
+
+----
+apiVersion: v1
+kind: Pod
+metadata:
+  name: volume-test
+spec:
+  containers:
+  - name: container-test
+    image: busybox
+    volumeMounts: <1>
+    - name: all-in-one
+      mountPath: "/projected-volume"<2>
+      readOnly: true <3>
+  volumes: <4>
+  - name: all-in-one <5>
+    projected:
+      defaultMode: 0400 <6>
+      sources:
+      - secret:
+          name: mysecret <7>
+          items:
+            - key: username
+              path: my-group/my-username <8>
+      - downwardAPI: <9>
+          items:
+            - path: "labels"
+              fieldRef:
+                fieldPath: metadata.labels
+            - path: "cpu_limit"
+              resourceFieldRef:
+                containerName: container-test
+                resource: limits.cpu
+      - configMap: <10>
+          name: myconfigmap
+          items:
+            - key: config
+              path: my-group/my-config
+              mode: 0777 <11>
+----
+
+<1> Add a `volumeMounts` section for each container that needs the secret.
+<2> Specify a path to an unused directory where the secret will appear.
+<3> Set `readOnly` to `true`.
+<4> Add a `volumes` block to list each projected volume source.
+<5> Specify any name for the volume.
+<6> Set the execute permission on the files.
+<7> Add a secret. Enter the name of the secret object. Each secret you want to use must be listed.
+<8> Specify the path to the secrets file under the `mountPath`. Here, the secrets file is in *_/projected-volume/my-group/my-config_*.
+<9> Add a Downward API source.
+<10> Add a ConfigMap source.
+<11> Set the mode for the specific projection
+====
+
+[NOTE]
+====
+If there are multiple containers in the pod, each container needs a `volumeMounts` section, but only one `volumes` section is needed.
+====
+
+
+.Pod with multiple secrets with a non-default permission mode set
+====
+
+----
+apiVersion: v1
+kind: Pod
+metadata:
+  name: volume-test
+spec:
+  containers:
+  - name: container-test
+    image: busybox
+    volumeMounts:
+    - name: all-in-one
+      mountPath: "/projected-volume"
+      readOnly: true
+  volumes:
+  - name: all-in-one
+    projected:
+      defaultMode: 0755
+      sources:
+      - secret:
+          name: mysecret
+          items:
+            - key: username
+              path: my-group/my-username
+      - secret:
+          name: mysecret2
+          items:
+            - key: password
+              path: my-group/my-password
+              mode: 511
+----
+
+====
+
+[NOTE]
+====
+The `defaultMode` can only be specified at the projected level and not for each
+volume source. However, as illustrated above, you can explicitly set the `mode`
+for each individual projection.
+====
+
+== Pathing Considerations
+
+When creating projected volumes, consider the following situations related to the volume file paths.
+
+*Collisions Between Keys when Configured Paths are Identical*:: If you configure any keys with the same path, the pod spec will not be accepted as valid.
+In the following example, the specified path for `mysecret` and `myconfigmap` are the same:
++
+----
+apiVersion: v1
+kind: Pod
+metadata:
+  name: volume-test
+spec:
+  containers:
+  - name: container-test
+    image: busybox
+    volumeMounts:
+    - name: all-in-one
+      mountPath: "/projected-volume"
+      readOnly: true
+  volumes:
+  - name: all-in-one
+    projected:
+      sources:
+      - secret:
+          name: mysecret
+          items:
+            - key: username
+              path: my-group/data
+      - configMap:
+          name: myconfigmap
+          items:
+            - key: config
+              path: my-group/data
+----
+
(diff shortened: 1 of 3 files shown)


mburke5678:
@pmorie @jpeeler Can you take a look? 
Is this what we need from Jesse's PR (https://github.com/openshift/openshift-docs/pull/4503).

dev_guide/projected_volumes.adoc
```diff
@@ -0,0 +1,170 @@
...
+      - secret:
+          name: pass
+----
+
+. Create files containing the secrets:
++
+----
+echo -n "admin" > ./username.txt
+echo -n "1f2d1e2e67df" > ./password.txt
+----
```
jpeeler: Missing the secret creation commands. You could get rid of the above two lines and use the following:

secret.yaml:
```
apiVersion: v1
kind: Secret
metadata:
  name: mysecret
type: Opaque
data:
  pass: MWYyZDFlMmU2N2Rm
  user: YWRtaW4=
---
```
oc create -f secret.yaml.

dev_guide/projected_volumes.adoc
```diff
@@ -0,0 +1,170 @@
...
+oc create -f <filename>
+----
+
+. Create the pod that references the secret:
++
+----
+oc create -f <your_yaml_file>.yaml
+----
+
```
jpeeler: I think the verification of examining inside the container is valuable, otherwise the "tasks" portion of this might should be excluded.

jpeeler:
Jesse's PR based upon my original PR is https://github.com/kubernetes/kubernetes.github.io/pull/3856/. If openshift doesn't have a "tasks" section like Kubernetes, combining them here is fine.

Made a few suggestions @mburke5678 on the commits.

dev_guide/projected_volumes.adoc
mburke5678: @jpeeler Where do the user and pass names come from?
Are they the same user/pass as in 
https://docs.openshift.com/container-platform/3.5/install_config/configuring_authentication.html

jpeeler: The user and pass names are used in https://github.com/kubernetes/kubernetes.github.io/pull/3856/files#diff-92b561a119469ea9519cb2bbe83437d9R21, which I think you left out. The values are made up and can be anything valid (base64 encoded).

mburke5678:
@jpeeler I added the verification section back (just seeing the directory structure is sufficient??) and added more detail to the first example with the secrets, downward API, and configmap volume sources. And I fleshed out the use case procedure more (I used a 3.5 OpenShift system in a pinch). PTAL.

mburke5678:
@chengzhang1016 Can you take a look at the document also? Trying to get this closed in

dev_guide/projected_volumes.adoc
```diff
@@ -145,26 +145,82 @@ spec:
       - secret:
           name: pass
 ----
++
+The `user` and `pass` values are made up and can be anything valid (base64 encoded).
```
jpeeler: I don't think I'd say it quite like this. Perhaps something along the lines of, the user and pass values are for example purposes only. The examples used here are base64 encoded values "user: admin", "pass:1f2d1e2e67df".

dev_guide/projected_volumes.adoc
```diff
@@ -145,26 +145,82 @@ spec:
...
+test-projected-volume   1/1       Running   0          14s
+
+. In another terminal, get a shell to the running Container:
+
+---
+oc exec -it test-projected-volume -- /bin/sh
+----
+
+. In your shell, verify that the `projected-volumes` directory contains your projected sources:
+
```
jpeeler: Ideally you'd do `ls projected-volumes` to show the projections are actually present.

jpeeler:
I left two more comments on older commits, so click the show outdated to read.

dev_guide/projected_volumes.adoc
mburke5678: @jpeeler What should the user see doing ls projected-volumes?

I do not have a 3.6 system at the moment (3.5 gives me an empty projected-volumes directory).

jpeeler: Unfortunately after looking closer, the projections is wrong. It should be like this for the pod spec:

```
apiVersion: v1
kind: Pod
metadata:
  name: test-projected-volume
spec:
  containers:
  - name: test-projected-volume
    image: busybox
    args:
    - sleep
    - "86400"
    volumeMounts:
    - name: all-in-one
      mountPath: "/projected-volume"
      readOnly: true
  volumes:
  - name: all-in-one
    projected:
      sources:
      - secret:
          name: mysecret
```

And then you can check in the container to see:
```
/ # ls projected-volume/
pass  user
```

dev_guide/projected_volumes.adoc
```diff
@@ -0,0 +1,289 @@
...
+. In your shell, verify that the `projected-volumes` directory contains your projected sources:
++
+----
+/ # ls
+bin               home              root              tmp
+dev               proc              run               usr
+etc               projected-volume  sys               var
+----
+
```
chengzhang1016: @jpeeler @mburke5678 
Suggest to add some notice for customer. Such as: Collisions when one path is explicit and the other is automatically projected. You can try to explain below 2 cases:
1. Test collisions when one path is automatically projected, and then another is explicit with same path.
2. Test collisions when one path is explicit, and then another is automatically projected with same path.
There is a TC for your reference: https://polarion.engineering.redhat.com/polarion/redirect/project/OSE/workitem?id=OCP-12395

dev_guide/projected_volumes.adoc
```diff
@@ -0,0 +1,289 @@
...
+              mode: 511
+----
+
+====
+
+[NOTE]
+====
+The `defaultMode` can only be specified at the projected level and not for each
+volume source. However, as illustrated above, you can explicitly set the `mode`
+for each individual projection.
```
mburke5678: I used the configs in a TC as a basis for this change. Please let me know if this is what you expected. I added the defaultMode to both examples in this section.  
OCP-12427 - All-in-one volume: Permission mode work well in whole volume and individual resources.

dev_guide/projected_volumes.adoc
mburke5678: @chengzhang1016 @jpeeler 
I took some of the collision scenarios from Jeff's document here:
https://github.com/kubernetes/community/blob/master/contributors/design-proposals/all-in-one-volume.md

Is this information appropriate for end user docs and address your request here?

Note, I also added "combinations" listed in the document's Analysis section as "Example Scenarios." Does this make sense in explaining more what projected volumes are used for?

dev_guide/projected_volumes.adoc
chengzhang1016: That is OKay for me. Thank you.

dev_guide/projected_volumes.adoc
chengzhang1016: Yes, I think the three collisions should be covered.
```
Collisions between keys when configured paths are identical.
Collisions between keys without configured paths
Collisions when one path is explicit and the other is automatically projected
```

mburke5678:
@chengzhang1016 Did my changes meet your expectations? Can we move the Trello to Accepted?
https://trello.com/c/AfFZsOyx/418-document-downward-api-all-in-one-volume-for-secrets-configmap-downward-api

mburke5678:
@chengzhang1016 accepted the changes via Trello.
" That is OKay for me. Thanks. Accept this card."

mburke5678:

dev_guide/projected_volumes.adoc
```diff
@@ -0,0 +1,368 @@
...
+:icons:
+:experimental:
+:toc: macro
+:toc-title:
+
+toc::[]
+
+== Overview
+
+A _projected volume_ maps several existing xref:volumes.adoc#dev-guide-volumes[volume sources] into the same directory.
```
bfallonf: Full path for the link above?

dev_guide/projected_volumes.adoc
```diff
@@ -0,0 +1,368 @@
...
+====
+All sources are required to be in the same namespace as the pod.
+====
+
+Projected volumes allow the user to map any combination of these volume sources into a single directory
+
+* automatically populate a single volume with the keys from multiple secrets, configmaps, and with downward API information,
+so that I can synthesize a single directory with various sources of information
+* populate a single volume with the keys from multiple secrets, configmaps, and with downward API information,
+explicitly specifying paths for each item, so that I can have full control over the contents of that volume
```
bfallonf: Should there be punctuation at the end of line 28? I'm not sure what to suggest exactly, because the sentence doesn't really run on to the two points above. Is there a way that can be clearer? I'd also suggest full stops after the points.

dev_guide/projected_volumes.adoc
```diff
@@ -0,0 +1,368 @@
...
+
+
+== Example Scenarios
+// https://github.com/kubernetes/community/blob/master/contributors/design-proposals/all-in-one-volume.md
+
+The following general scenarios show how you can use projected volumes.
+
+* *ConfigMap, Secrets, Downward API.* Projected volumes allow you to to deploy containers with configuration data that includes passwords.
+An application using these resources could be deploying OpenStack on Kubernetes. The configuration data may need to be assembled differently depending on if the services are going to be used for production or for testing. If a pod is labeled with production or testing, the downward API selector metadata.labels can be used to produce the correct OpenStack configs.
```
bfallonf: Markup for "metadata.labels"

dev_guide/projected_volumes.adoc
```diff
@@ -0,0 +1,368 @@
...
+
+* *ConfigMap + Secrets.* Projected volumes allow you to deploy containers involving configuration data and passwords.
+For example, you might execute an Ansible playbook stored as a configmap, with some sensitive encrypted tasks that are decrypted using a vault password file.
+
+* *ConfigMap + Downward API.* Projected volumes allow you to generate a config including the pod name (available via the `metadata.name` selector). This application can then pass the pod name along with requests in order to easily determine the source without using IP tracking.
+
+* *Secrets + Downward API.* Projected volumes allow you to use a secret as a public key to encrypt the namespace of the pod (available via the `metadata.namespace` selector).
+This example allows the operator to use the application to deliver the namespace information securely without using an encrypted transport.
+
+## Example Pod Specifications
```
bfallonf: To be consistent, I'd change the ##s above (and there's a few more in the doc) to be ==s.

dev_guide/projected_volumes.adoc
```diff
@@ -0,0 +1,368 @@
...
+      - configMap:
+          name: myconfigmap
+          items:
+            - key: config
+              path: my-group/data
+----
+
+### Collisions between keys without configured paths
+
+The only run-time validation can occur is when all the paths are known at pod creation, similar to the above scenario.
```
bfallonf: Something is missing from the above. I think a "that" after "validation"?

"The only run-time validation that can occur is when all the paths are known at pod creation..."

dev_guide/projected_volumes.adoc
```diff
@@ -0,0 +1,368 @@
...
+====
+
+[NOTE]
+====
+The `defaultMode` can only be specified at the projected level and not for each
+volume source. However, as illustrated above, you can explicitly set the `mode`
+for each individual projection.
+====
+
+## Pathing Considerations
```
bfallonf: OK, realising that the hashes are leftover from markdown now: http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#markdown-compatibility

I'm wondering if this section would be better as a labeled list, though. It's all part of the same thing and might clog up a ToC? At the least, I'd suggest changing the #s to =s and putting caps in the titles.

dev_guide/projected_volumes.adoc
```diff
@@ -0,0 +1,368 @@
...
+### Collisions when one path is explicit and the other is automatically projected
+
+In the event that there is a collision due to a user specified path matching data that is automatically projected,
+the latter resource will overwrite anything preceding it as before
+
+## Configuring a projected volume for a pod
+
+The following example shows how to use a projected volume to mount an existing Secret volume source.
+
+The steps can be used to create a user name and password xref:secrets.adoc#dev-guide-secret[Secrets] from local files. You then create a pod that runs one container, using a projected volume to mount the Secrets into the same shared directory.
```
bfallonf: Full path for the link above?

bfallonf:
@mburke5678 I made some comments. Mainly formatting stuff. Let me know what you think.

dev_guide/projected_volumes.adoc
mburke5678: @bfallonf Changed to labelled lists. Do we have a "proper" formatting style for labelled lists?

bfallonf: Not sure if it's official, but we usually use two ::s after the heading. It come sup like this:

https://docs.openshift.com/container-platform/3.5/admin_guide/sdn_troubleshooting.html#nomenclature

That said, it's just a suggestion, and I see you've put it into dot points, which works.

mburke5678: Might I merge @bfallonf ?

bfallonf: @mburke5678  Sure, merge away.

mburke5678:
[rev_history]
|xref:../dev_guide/projected_volumes.adoc#dev-guide-projected-volumes[Developer Guide -> Projected Volumes]
|Added new section on xref:../dev_guide/projected_volumes.adoc#dev-guide-projected-volumes[Kubernetes projected volumes] to the Developer Guide.
%

adellape:
@mburke5678 The section titles in this could use [unique IDs/anchors](https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#unique-ids).

mburke5678:
See follow-up for IDs https://github.com/openshift/openshift-docs/pull/4601

PR #4506, opened 2017-05-31
Correct incorrect route in oc new-app command

thomasmckay:
The docker-registry route does not exist, only the registry-console route. Changing the example command to use registry-console allows it to run successfully and the registry console UI to be accessed. I cannot confirm if this is correct behavior or that it simply "works".


thomasmckay:
The instructions don't have the creation of the docker-registry route; is that part of a default openshift install? If not, some instructions on this page on how to do that would be welcome.

I am running minishift and was instructed to use the addons feature for registry-route. That appears to work (still testing).

In either case, this PR is incorrect and I will close it after getting feedback on where this should be documented.

Thanks!

thomasmckay:
The registry-route addons is needed.

PR #4514, opened 2017-05-31
Online product name updates

ahardin-rh:


ahardin-rh:
@adellape @bmcelvee @mburke5678 This introduces name changes for when we launch Online pro. We are only calling the product OpenShift Online, but will differentiate between Online Pro and Online Starter when needed.

adellape (commented):
LGTM. We also need to update distro_map to no longer say "Starter", too:

https://github.com/openshift/openshift-docs/blob/master/_distro_map.yml#L15-L24

For Dedicated (which is also versionless for docs), we just have "Latest" there. Could also be "3"... Or "Starter and Pro"... Really it's just going to show up in the breadcrumb at the top of the page on docs.openshift.com. I don't _think_ it has any bearing on Customer Portal title. @vikram-redhat ?

@aheslin WDYT ^ ?

ahardin-rh:
@adellape Thanks! On https://docs.openshift.com/ we use 3 in the drop-down menu, so maybe we should keep consistent? I defer to @aheslin and @vikram-redhat, however

vikram-redhat:
The portal will need a version but I handle that behind the scenes and just use 3 (same as dedicated). Other than that, there is no issue. 

Some of it might depend on whether we need to keep the version 2 docs? If we are only keeping one set of docs we can just go the dedicated route and call it latest. But that is @aheslin 's call :)

aheslin:
We have to keep the Online v2 docs for now since there are support exceptions for it. I'm fine with using version 3.

ahardin-rh:
No revision history needed

PR #4516, opened 2017-06-01
Add cluster capacity documentation.

aveshagarwal:
@bfallonf @derekwaynecarr

### DIFF
--- a/admin_guide/cluster_capacity.adoc
+++ b/admin_guide/cluster_capacity.adoc
@@ -1,0 +1,199 @@
+[[admin-guide-cluster-capacity]]
+= Cluster Capacity Analysis
+{product-author}
+{product-version}
+:data-uri:
+:icons:
+:experimental:
+:toc: macro
+:toc-title:
+
+toc::[]
+
+== Overview
+
+Monitoring available resources in the cluster is very important as operators can increase
+the current resources in time before all of them get exhausted, or can carry different
+steps that lead to increase of available resources. Cluster capacity consists of capacities
+of individual cluster nodes. Capacity covers CPU, memory, disk space and other resources.
+
+Overall remaining allocatable capacity is a rough estimation since it does not assume all
+resources being distributed among nodes. Goal is to analyze remaining allocatable resources
+and estimate available capacity that is still consumable in terms of a number of instances
+of a pod with given requirements that can be scheduled in a cluster.
+
+== Running cluster capacity analysis
+
+Cluster capacity analysis tool can be run either as a stand-alone utility from command line,
+or in a pod inside a openshift cluster.
+
+[admin-guide-cluster-capacity-command-line]
+### Running cluster capacity from command line
+
+To run the tool on command line:
+
+```shell
+$ cluster-capacity --kubeconfig <path-to-kubeconfig> --podspec <path-to-pod-spec>
+```
+
+`--kubeconfig` is path to kubeconfig file to cluster and `--podspec` is path to sample pod spec
+that is used by the tool for estimation. The `podspec` specifies its resource requirements as
+`limits` or `requests`. The cluster capacity tool takes the pod's resource requirements into account
+for its estimation analysis. An example of input pod spec (`pod.yaml`) is given below:
+
+```
+apiVersion: v1
+kind: Pod
+metadata:
+  name: small-pod
+  labels:
+    app: guestbook
+    tier: frontend
+spec:
+  containers:
+  - name: php-redis
+    image: gcr.io/google-samples/gb-frontend:v4
+    imagePullPolicy: Always
+    resources:
+      limits:
+        cpu: 150m
+        memory: 100Mi
+      requests:
+        cpu: 150m
+        memory: 100Mi
+```
+
+
+The output of the cluster capacity command is a number of estimated pods that can be scheduled on the cluster.
+`--verbose` flag can also be provided as part of the above command to output detailed description
+about how many pods can be scheduled on each node in the cluster. An example of verbose output on
+a 2-node cluster is as follows:
+
+```shell
+small-pod pod requirements:
+	- CPU: 150m
+	- Memory: 100Mi
+
+The cluster can schedule 52 instance(s) of the pod small-pod.
+
+Termination reason: Unschedulable: No nodes are available that match all of the following predicates:: Insufficient cpu (2).
+
+Pod distribution among nodes:
+small-pod
+	- 192.168.124.214: 26 instance(s)
+	- 192.168.124.120: 26 instance(s)
+```
+
+[admin-guide-cluster-capacity-job]
+### Running cluster capacity as a Job
+
+Below is an example of a Job spec (`cluster-capacity-job.yaml`) that runs cluster capacity image as a job in a pod.
+Note that input pod spec (`pod.yaml`) to cluster capacity analysis is mounted in a volume using a `ConfigMap`.
+
+```yaml
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: cluster-capacity-job
+spec:
+  parallelism: 1
+  completions: 1
+  template:
+    metadata:
+      name: cluster-capacity-pod
+    spec:
+        containers:
+        - name: cluster-capacity
+          image: openshift/origin-cluster-capacity
+          imagePullPolicy: "Never"
+          volumeMounts:
+          - mountPath: /test-pod
+            name: test-volume
+          env:
+          - name: CC_INCLUSTER
+            value: "true"
+          command:
+          - "/bin/sh"
+          - "-ec"
+          - |
+            /bin/cluster-capacity --podspec=/test-pod/pod.yaml --verbose
+        restartPolicy: "Never"
+        serviceAccountName: cluster-capacity-sa
+        volumes:
+        - name: test-volume
+          configMap:
+            name: cluster-capacity-configmap
+```
+
+Also note that the above pod spec specifies the following environment variable which is required to
+let the cluster capacity command know that it is running inside a cluster as a pod.
+
+```yaml
+env:
+- name: CC_INCLUSTER
+  value: "true"
+```
+
+By default, this pod would be created in the `default` namespace by using `default` service account. However, the above
+pod spec refers to a different service account (`cluster-capacity-sa`) in the `default` namespace created as below:
+
+```
+$ oc create sa cluster-capacity-sa
+```
+
+This service account `cluster-capacity-sa` is assigned a new cluster scope role `cluster-capacity-role` to be able to access
+resources required to run cluster capacity command inside a pod. The role `cluster-capacity-role` is created as:
+
+```shell
+$ cat << EOF| oc create -f -
+kind: ClusterRole
+apiVersion: v1
+metadata:
+  name: cluster-capacity-role
+rules:
+- apiGroups: [""]
+  resources: ["pods", "nodes", "persistentvolumeclaims", "persistentvolumes", "services"]
+  verbs: ["get", "watch", "list"
+EOF
+```
+
+The next step binds the service account `cluster-capacity-sa` to the new cluster role `cluster-capacity-role`:
+
+```shell
+$ oadm policy add-cluster-role-to-user cluster-capacity-role system:serviceaccount:default:cluster-capacity-sa
+```
+
+Please note that in the above Job spec, we used a `ConfigMap` `cluster-capacity-configmap` to mount input
+pod spec file `pod.yaml` into a volume `test-volume` at the path `/test-pod`. The ConfigMap can be created as:
+
+```shell
+$ oc create configmap cluster-capacity-configmap --from-file=pod.yaml=pod.yaml
+```
+
+Please note that the key `pod.yaml` of the ConfigMap is same as pod spec file name, though it is not required.
+By doing this, the input pod spec file can be accessed inside the pod as `/test-pod/pod.yaml`.
+
+Now let's run the above job in an openshift cluster:
+
+```shell
+$ oc create -f cluster-capacity-job.yaml
+job "cluster-capacity-job" created
+```
+
+The logs from the above job can be checked as follows and show the number of pods that can be scheduled in the cluster.
+
+```shell
+$ oc logs jobs/cluster-capacity-job
+small-pod pod requirements:
+        - CPU: 150m
+        - Memory: 100Mi
+
+The cluster can schedule 52 instance(s) of the pod small-pod.
+
+Termination reason: Unschedulable: No nodes are available that match all of the following predicates:: Insufficient cpu (2).
+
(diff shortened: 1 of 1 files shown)


admin_guide/cluster_capacity.adoc
```diff
@@ -0,0 +1,160 @@
...
+```
+
+By default, this pod would be created in the `default` namespace by using `default` service account. However, the above
+pod spec refers to a different service account (`cluster-capacity-sa`) in the `default` namespace created as below:
+
+```
+$ oc create sa cluster-capacity-sa
+```
+
+This service account `cluster-capacity-sa` is given `cluster-admin` cluster scope role below so that the cluster capacity command
```
derekwaynecarr: cluster-admin is way more permission than is required.  Wouldn't cluster-reader work as well?

Ideally we should spec how to create a role with the actual permissions needed.

derekwaynecarr (changes requested):
We should have a more scoped role.  Isn't there a role now for the scheduler ?  Can we not take that and just remove the ability to bind from that role for this tool?

aveshagarwal:
On Thu, Jun 1, 2017 at 12:24 AM Derek Carr <notifications@github.com> wrote:

> *@derekwaynecarr* requested changes on this pull request.
>
> We should have a more scoped role. Isn't there a role now for the
> scheduler ?
>
There is, I will try that tomorrow.

> Can we not take that and just remove the ability to bind from that role
> for this tool?
>

That said, I was also hesitant to bind to cluster-admin role and tried
cluster-readers and node-reader and perhaps some other too, nothing worked.

------------------------------
>
> In admin_guide/cluster_capacity.adoc
> <https://github.com/openshift/openshift-docs/pull/4516#discussion_r119525470>
> :
>
> > +let the cluster capacity command know that it is running inside a cluster as a pod.
> +
> +```yaml
> +env:
> +- name: CC_INCLUSTER
> +  value: "true"
> +```
> +
> +By default, this pod would be created in the `default` namespace by using `default` service account. However, the above
> +pod spec refers to a different service account (`cluster-capacity-sa`) in the `default` namespace created as below:
> +
> +```
> +$ oc create sa cluster-capacity-sa
> +```
> +
> +This service account `cluster-capacity-sa` is given `cluster-admin` cluster scope role below so that the cluster capacity command
>
> cluster-admin is way more permission than is required.
>
Agreed.

> Wouldn't cluster-reader work as well?
>
it didn't.

> Ideally we should spec how to create a role with the actual permissions
> needed.
>
Yeah will update for this in the morning.

bfallonf:
Thanks @aveshagarwal . I'll let you follow up with @derekwaynecarr 's changes then do a followup.

aveshagarwal:
>Isn't there a role now for the scheduler ?
As current system:kube-scheduler is not working, I have added a new role that adds extra (whatever extra  resources were needed by cluster capacity) read permissions needed.

I think now its looks good and much better restricted that unlimited cluster admin role. Let me know your thoughts @derekwaynecarr

admin_guide/cluster_capacity.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+```shell
+$ cat << EOF| oc create -f -
+kind: ClusterRole
+apiVersion: v1
+metadata:
+  name: kube-scheduler-reader-plus
+rules:
+- apiGroups: [""]
+  resources: ["limitranges", "serviceaccounts", "secrets", "namespaces", "resourcequotas"]
```
derekwaynecarr: remind me again why it needs limitranges secrets and quotas?

aveshagarwal: After adding kube-scheduler role to the cluster capacity sa, these resources were throwing error and required to be added when running cluster capacity, so I added them.

derekwaynecarr: i expect us to be able to run the command with scheduler privileges minus bind.

admin_guide/cluster_capacity.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+Pod distribution among nodes:
+small-pod
+	- 192.168.124.214: 26 instance(s)
+	- 192.168.124.120: 26 instance(s)
+
+```
+
+[admin-guide-cluster-capacity-pod]
+### Running cluster capacity in a pod
```
derekwaynecarr: can we update this example to "Running cluster capacity as a Job"?

admin_guide/cluster_capacity.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+Below is an example of a pod spec (`cluster-capacity-pod.yaml`) that runs cluster capacity image as a container. Note that
+input pod spec (`pod.yaml`) to cluster capacity analysis is mounted in a volume using `hostPath`.
+
+```yaml
+apiVersion: v1
+kind: Pod
```
derekwaynecarr: use a job with pod template.

admin_guide/cluster_capacity.adoc
aveshagarwal: sounds good.

admin_guide/cluster_capacity.adoc
aveshagarwal: sounds good.

admin_guide/cluster_capacity.adoc
```diff
@@ -0,0 +1,184 @@
...
+small-pod
+	- 192.168.124.214: 26 instance(s)
+	- 192.168.124.120: 26 instance(s)
+
+```
+
+[admin-guide-cluster-capacity-pod]
+### Running cluster capacity in a pod
+
+Below is an example of a pod spec (`cluster-capacity-pod.yaml`) that runs cluster capacity image as a container. Note that
```
derekwaynecarr: i would like an example that does not require host path.

can we put the pod spec in a config map and reference that instead?

host path requires privileged, and there is no reason this should require that.

aveshagarwal: Yeah sure, I will look into that. I thought this documentation was just "a" way to explain how cluster capacity can be used, because most admin would use it in different ways whatever way they want.

admin_guide/cluster_capacity.adoc
derekwaynecarr: i expect the flow to be the following:

1. call oc create configmap to create a config map with your pod spec
2. create a job that references your configmap
3. call oc logs job/foo to show the result

aveshagarwal: +1

admin_guide/cluster_capacity.adoc
derekwaynecarr: i have learned that people will do what is presented first.  so lets provide the best possible example first.

bfallonf:
@aveshagarwal @derekwaynecarr Can I ask about the status of this card? I can see changes are requested, but is that up to date?

Is there anything I can help with to move this along?

aveshagarwal:
@bfallonf To address the latest concerns raised by @derekwaynecarr, it would require changes in the code. I am looking into that.

bfallonf:
@aveshagarwal Sure, no problems. I'll keep watching this PR and jump in when necessary.

admin_guide/cluster_capacity.adoc
derekwaynecarr: @aveshagarwal -- i think this should fix https://github.com/kubernetes-incubator/cluster-capacity/pull/77

aveshagarwal:
@derekwaynecarr @bfallonf updated this PR. PTAL.

admin_guide/cluster_capacity.adoc
```diff
@@ -0,0 +1,199 @@
...
+:icons:
+:experimental:
+:toc: macro
+:toc-title:
+
+toc::[]
+
+== Overview
+
+Monitoring available resources in the cluster is very important as operators can increase
```
derekwaynecarr: @bfallonf -- if you can take a second pass at this for readability, i think it would be good.  i think this can be condensed.

it is important to monitor available resources on your cluster in order to ensure future pods are able to schedule.  as a cluster is populated with pods, fragmentation will occur.  pods may only support scheduling on particular sets of nodes based on selection and affinity criteria.  as a result, it can be difficult to get an estimate on how many more pods a cluster can schedule.

the cluster capacity tool simulates a sequence of scheduling decisions to determine how many instances of an input pod may schedule on the cluster before it is exhausted of resources in order to provide a more accurate estimation.

admin_guide/cluster_capacity.adoc
```diff
@@ -0,0 +1,199 @@
...
+of individual cluster nodes. Capacity covers CPU, memory, disk space and other resources.
+
+Overall remaining allocatable capacity is a rough estimation since it does not assume all
+resources being distributed among nodes. Goal is to analyze remaining allocatable resources
+and estimate available capacity that is still consumable in terms of a number of instances
+of a pod with given requirements that can be scheduled in a cluster.
+
+== Running cluster capacity analysis
+
+Cluster capacity analysis tool can be run either as a stand-alone utility from command line,
```
derekwaynecarr: from the command line

admin_guide/cluster_capacity.adoc
```diff
@@ -0,0 +1,199 @@
...
+        memory: 100Mi
+```
+
+
+The output of the cluster capacity command is a number of estimated pods that can be scheduled on the cluster.
+`--verbose` flag can also be provided as part of the above command to output detailed description
+about how many pods can be scheduled on each node in the cluster. An example of verbose output on
+a 2-node cluster is as follows:
+
+```shell
```
derekwaynecarr: i would show the exact command used here inside the shell.

admin_guide/cluster_capacity.adoc
```diff
@@ -0,0 +1,199 @@
...
+  parallelism: 1
+  completions: 1
+  template:
+    metadata:
+      name: cluster-capacity-pod
+    spec:
+        containers:
+        - name: cluster-capacity
+          image: openshift/origin-cluster-capacity
+          imagePullPolicy: "Never"
```
derekwaynecarr: this should not be Never

admin_guide/cluster_capacity.adoc
```diff
@@ -0,0 +1,199 @@
...
+Pod distribution among nodes:
+small-pod
+	- 192.168.124.214: 26 instance(s)
+	- 192.168.124.120: 26 instance(s)
+```
+
+[admin-guide-cluster-capacity-job]
+### Running cluster capacity as a Job
+
+Below is an example of a Job spec (`cluster-capacity-job.yaml`) that runs cluster capacity image as a job in a pod.
```
derekwaynecarr: nit: maybe show the oc create configmap command you used to populate it?  it would help people reproduce the steps.

admin_guide/cluster_capacity.adoc
```diff
@@ -0,0 +1,199 @@
...
+        - name: test-volume
+          configMap:
+            name: cluster-capacity-configmap
+```
+
+Also note that the above pod spec specifies the following environment variable which is required to
+let the cluster capacity command know that it is running inside a cluster as a pod.
+
+```yaml
+env:
```
derekwaynecarr: not needed now, but is there a better way we can detect this in future?

for example, could we simplify this in future by looking for `KUBERNETES_SERVICE_HOST` env var or seeing that /var/run/secrets/kubernetes.io/serviceaccount is present?

admin_guide/cluster_capacity.adoc
```diff
@@ -0,0 +1,199 @@
...
+- name: CC_INCLUSTER
+  value: "true"
+```
+
+By default, this pod would be created in the `default` namespace by using `default` service account. However, the above
```
derekwaynecarr: i think we could have a section and title: Add a cluster role to run command

admin_guide/cluster_capacity.adoc
```diff
@@ -0,0 +1,199 @@
...
+
+```shell
+$ oadm policy add-cluster-role-to-user cluster-capacity-role system:serviceaccount:default:cluster-capacity-sa
+```
+
+Please note that in the above Job spec, we used a `ConfigMap` `cluster-capacity-configmap` to mount input
+pod spec file `pod.yaml` into a volume `test-volume` at the path `/test-pod`. The ConfigMap can be created as:
+
+```shell
+$ oc create configmap cluster-capacity-configmap --from-file=pod.yaml=pod.yaml
```
derekwaynecarr: i think we should reverse the order of this doc.

1. create the cluster role
2. create the service account
3. add the role to service account
4. define your pod spec
5. make it available in config map
6. run the job (now that you have done what is needed)

derekwaynecarr:
i have more style nits than technical problems.

@bfallonf - will leave this up to you to fix up more in a follow-on, but technical content is LGTM

derekwaynecarr (approved):

admin_guide/cluster_capacity.adoc
bfallonf: Combined the two suggested intros.

admin_guide/cluster_capacity.adoc
bfallonf: Added 

----
$ cluster-capacity --kubeconfig <path-to-kubeconfig> \
    --podspec <path-to-pod-spec> --verbose
----

admin_guide/cluster_capacity.adoc
bfallonf: Added

----
$ oc create configmap cluster-capacity-job
----

admin_guide/cluster_capacity.adoc
bfallonf: Changed to `Always`

admin_guide/cluster_capacity.adoc
bfallonf: @derekwaynecarr Is this intended for @aveshagarwal ? I'm not sure what you mean.

aveshagarwal: yes for future work, not now.

admin_guide/cluster_capacity.adoc
```diff
@@ -0,0 +1,199 @@
...
+$ cat << EOF| oc create -f -
+kind: ClusterRole
+apiVersion: v1
+metadata:
+  name: cluster-capacity-role
+rules:
+- apiGroups: [""]
+  resources: ["pods", "nodes", "persistentvolumeclaims", "persistentvolumes", "services"]
+  verbs: ["get", "watch", "list"
+EOF
```
chengzhang1016: Should be: verbs: ["get", "watch", "list"]
Miss a "]" in here.

bfallonf: Thanks @chengzhang1016 . I can add that to the followup PR #4610

aveshagarwal:
I am closing this PR as there is another follow up PR https://github.com/openshift/openshift-docs/pull/4610 .

PR #4519, opened 2017-06-01
WIP: Adding advanced installation settings for glusterfs backed registry and persistent storage for new/greenfield installations

screeley44:
This PR is my initial work in progress and focuses on the Advanced Installation instructions - adding 2 new sections for installing GlusterFS backed registry on initial install (greenfield) and/or installing GlusterFS for persistent storage.  For 3.6 this is our main focus...

There are many ways to skin the cat on this, but we should guide the user to a very well defined set of instructions with exact values they need.  Other features might be worth documenting as well, but need to think where we should put those (swapping registry, etc...) and if they are better suited for 3.7 release??

@erinboyd @jarrpa @adellape - take a look

### DIFF
--- a/install_config/install/advanced_install.adoc
+++ b/install_config/install/advanced_install.adoc
@@ -426,6 +426,304 @@ openshift_examples_modify_imagestreams=true
 |Set to `true` if pointing to a registry other than the default. Modifies the image stream location to the value of `*oreg_url*`.
 |===
 
+[[advanced-install-configuring-integrated-glusterfs-backed-registry]]
+=== Configuring an Integrated GlusterFS Backed Registry
+
+ifdef::openshift-enterprise[]
+An integrated GlusterFS Registry can be configured and deployed during the initial installation
+of the cluster to offer a more reliable image registry. link:https://access.redhat.com/documentation/en-us/red_hat_gluster_storage/3.1/html/container-native_storage_for_openshift_container_platform_3.4/[Red Hat Container Native Storage (CNS)] has been integrated into the {product-title} Advanced Installation procedure utilizing containerized GlusterFS storage pods to back the Docker registry.
+endif::[]
+
+ifdef::openshift-origin[]
+An integrated GlusterFS Registry can be configured and deployed during the initial installation
+of the cluster to offer a more reliable image registry. See link:https://github.com/gluster/gluster-kubernetes[Running Containerized GlusterFS in Kubernetes] for additional information on containerized storage using GlusterFS.
+endif::[]
+
+Some additional prerequisites and host preparation are required beyond the normal {product-title} xref:../../install_config/install/prerequisites.adoc#install-config-install-prerequisites[Prerequisites]. 
+
+.Additional Prerequisites and Host Preparation
+[options="header"]
+|===
+
+|Prerequisite |Description
+|`*Storage Nodes*`
+|A minimum of 3 storage nodes must be available to run GlusterFS integrated registry (3-way replication)
+ 
+ Each storage node must have at least 1 raw block device with at least 10GB of free storage
+ifdef::openshift-enterprise[]
+|`*Enable Storage Repo*`
+|rh-gluster-3-for-rhel-7-server-rpms will need to be enabled. 
+
+      `# subscription-manager repos --enable=rh-gluster-3-for-rhel-7-server-rpms`
+endif::[]
+ifdef::openshift-enterprise[]
+|`*Required Software Components*`
+|Install *glusterfs-client* on each storage node. 
+
+      `# yum install glusterfs-client -y`
+
+endif::[]
+ifdef::openshift-origin[]
+|`*Required Software Components*`
+|Install *glusterfs-client* or *glusterfs-fuse* packages (mount.glusterfs) on each storage node. 
+
+endif::[]
+|===
+
+
+In the inventory/hosts file the following will need to be configured
+
+.Integrated GlusterFS Registry Variables and Groups
+[options="header"]
+|===
+
+|Variable |Purpose
+|`*glusterfs_registry*`
+|Placed in the OSEv3:children section and enables the [glusterfs_registry] group
+
+|`*[glusterfs_registry]*`
+| The group of storage nodes that will host the GlusterFS Registry along with the `glusterfs_devices` list parameter
+
+  <hostname or ip> `glusterfs_devices`='[ "(path-to-device)", "(...)" ]'
+
+|`*openshift_hosted_registry_storage_kind=glusterfs*`
+|Placed in the OSEv3:vars section and enables GlusterFS Registry if `[glusterfs_registry]` group exists and `glusterfs_registry` group_name exists
+
+|`*Nodes labeled with region=Infra*`
+|{product-title} will install the integrated docker registry pods and the associated routers on nodes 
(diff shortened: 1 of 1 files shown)


install_config/install/advanced_install.adoc
```diff
@@ -426,6 +426,203 @@ openshift_examples_modify_imagestreams=true
 |Set to `true` if pointing to a registry other than the default. Modifies the image stream location to the value of `*oreg_url*`.
 |===
 
+[[advanced-install-configuring-integrated-glusterfs-backed-registry]]
+=== Configuring an Integrated GlusterFS Backed Registry
+
+An integrated GlusterFS Registry can be configured and deployed during the initial installation
+of the cluster to offer a more reliable image registry. link:https://access.redhat.com/documentation/en-us/red_hat_gluster_storage/3.1/html/container-native_storage_for_openshift_container_platform_3.4/[Red Hat Container Native Storage (CNS)] has been integrated into the {product-title} Advanced Installation procedure utilizing containerized GlusterFS storage pods to back the Docker registry.
```
jarrpa: Don't mention product names (e.g. CNS). This should be as upstream as possible.

jarrpa: Also, missing xref for `Advanced Installation`?

install_config/install/advanced_install.adoc
```diff
@@ -426,6 +426,203 @@ openshift_examples_modify_imagestreams=true
...
+
+.Additional Prerequisites and Host Preparation
+[options="header"]
+|===
+
+|Prerequisite |Description
+|`*Storage Nodes*`
+|A minimum of 3 storage nodes must be available to run GlusterFS integrated registry (3-way replication)
+ 
+ Each storage node must have at least 1 raw block device with at least 10GB of free storage (2GB for the distributed heketi.db and 8GB for registry storage)
```
jarrpa: No need to mention heketi specifically. Just something like `2GB for the volume management database` would do.

install_config/install/advanced_install.adoc
```diff
@@ -426,6 +426,203 @@ openshift_examples_modify_imagestreams=true
...
+
+|`*Enable Storage Repo*`
```
jarrpa: Don't mention repos. Not only should this also apply upstream, the notion of repos doesn't apply to any Atomic OS.

install_config/install/advanced_install.adoc
```diff
@@ -426,6 +426,203 @@ openshift_examples_modify_imagestreams=true
...
+|rh-gluster-3-for-rhel-7-server-rpms will need to be enabled for installation of *heketi-client*. 
+
+      `# subscription-manager repos --enable=rh-gluster-3-for-rhel-7-server-rpms`
+
+|`*Install Components*`
```
jarrpa: Change this to `Required Software` or something like that. As mentioned above, the notion of RPMs doesn't apply to Atomic OSes. You should say something like "The following software/components are required", and them mention the programs directly. In the case of `heketi-cli`, point to the GitHub repo: https://github.com/heketi/heketi/releases .

install_config/install/advanced_install.adoc
```diff
@@ -426,6 +426,203 @@ openshift_examples_modify_imagestreams=true
...
+|Install *glusterfs-client* on each storage node. 
+
+      `# yum install glusterfs-client -y`
```
jarrpa: Only mention the software needed, which is `mount.glusterfs`. You can include a comment that this is found in the `glusterfs-fuse` package on RH OSes and `glusterfs-client` on Debian OSes.

jarrpa (changes requested):
A great start! A few top-level comments:

* The GlusterFS as Persistent Storage section should come first. It should be the central point to discuss pre-reqs and configuration options, with any other sections referencing it as needed.
* The GlusterFS as registry storage should probably actually be a subsection of a `Configuring Cluster Hosted Registry` section. This unfortunately doesn't exist yet, so it should be added. Follow the general format of the `Configuring Cluster Metrics` and `Configuring Cluster Logging` sections, and include a GlusterFS storage example as the first example. :)

install_config/install/advanced_install.adoc
screeley44: @jarrpa - actually this is downstream, so I think it's ok to mention CNS in this case

install_config/install/advanced_install.adoc
screeley44: @jarrpa - same as previous comment, this is actually downstream - so products are relevant

install_config/install/advanced_install.adoc
screeley44: @jarrpa - this doc is in the advanced installation section of the admin guide, maybe I don't need to reference Advanced Installation since that is where the user is currently looking?

install_config/install/advanced_install.adoc
screeley44: ok

install_config/install/advanced_install.adoc
screeley44: per our discussion, I will use the '''ifdef::{product-title}``` to properly document downstream (enterprise) vs upstream (origin)

install_config/install/advanced_install.adoc
screeley44: this might not be needed, verifying and if not, will remove

install_config/install/advanced_install.adoc
screeley44: done

install_config/install/advanced_install.adoc
screeley44: done

jarrpa (changes requested):
My two previous points still stand. :)

adellape:
@screeley44 @jarrpa FYI, I'm picking this up in a new PR to continue edits and address @jarrpa's https://github.com/openshift/openshift-docs/pull/4519#pullrequestreview-41859535 comments. Should be able to xref the new PR and close this one out today, but leaving it open til then.

install_config/install/advanced_install.adoc
adellape: In https://docs.openshift.com/container-platform/3.5/install_config/persistent_storage/persistent_storage_glusterfs.html#gfs-provisioning we say to install `glusterfs-fuse` everywhere. Another spot mentions:

> All OpenShift Container Platform nodes on RHEL systems must have the glusterfs-fuse RPM installed, which should match the version of Red Hat Gluster Storage server running in the containers.

adellape:
> The GlusterFS as Persistent Storage section should come first. It should be the central point to discuss pre-reqs and configuration options, with any other sections referencing it as needed.

@jarrpa Would you have an issue if I created a new GlusterFS subsection in https://docs.openshift.com/container-platform/3.5/install_config/install/prerequisites.html and moved the list of prereqs there? Then both of these two new subsections in `advanced_install.adoc` could just refer back to that?

jarrpa:
@adellape ...hm! I think that would work. Possibly as a subsection of "Persistent Storage"? it would have to cover prereqs for both native and external GlusterFS configurations. They are largely the same, just the external nodes have to have GlusterFS installed on the base OS.

adellape:
Continuing via https://github.com/openshift/openshift-docs/pull/4748.

PR #4526, opened 2017-06-01
Fix up target link

lpsantil:


bfallonf:
Sure. I think the intention was to link to the guide and not the topic itself, but you're right that it's more precise to link to here. I'll merge this. Thanks @lpsantil !

I don't think a review or rev history is needed.

PR #4527, opened 2017-06-02
Add storage recommendations and applications

lpsantil:
This is based, in large part, on information found here [0].

[0] https://docs.google.com/spreadsheets/d/1pTAro4ONaCCVBW3f9B9SxDT95F4dcMP3dJ_D_v_QPjo/edit#gid=0

### DIFF
--- a/scaling_performance/optimizing_storage.adoc
+++ b/scaling_performance/optimizing_storage.adoc
@@ -15,7 +15,7 @@ toc::[]
 
 == General Storage Guidance
 
-The following is a matrix which details the storage backend types {product-title} can be used with implementation examples and vendors.
+The following is a matrix which details the persistent storage backend types {product-title} can be used with, and includes implementation examples and vendors.
 
 === Available Storage Options
 [format="csv",width="70%",cols="3"]
@@ -23,11 +23,11 @@ The following is a matrix which details the storage backend types {product-title
 |=======================
 Type,Examples,Vendors
 hostPath,"hostPath pod mounted volumes.  A specific type of Block Storage.  File paths on the node mounted onto running pods.",Others
-Block,"Block Storage. HDD, SSD, NVMe, Raw Physical Disk, VM Physical Disk, VMDK, LVM, iSCSI, Fibre Channel, Ceph RBD, OpenStack Cinder, Dell/EMC Scale.IO, AWS EBS ^(1)^, GCE Persistent Disk ^(1)^, Azure Disk","RedHat, Seagate, WesternDigital, Toshiba, HGST, Intel, Samsung, OCZ, Plextor, PNY, Kingston, SanDisk, AWS, GCE, VMware, Dell/EMC, HPE, Others"
-File,"Network Attached File Storage.  CNS GlusterFS ^(1)^, GlusterFS ^(1)^, RHEL NFS, NetApp NFS ^(2)^, Vendor NFS, Vendor GlusterFS ^(3)^, Loopback Image over NFS","RedHat, NetApp, AWS, GCE, Others"
+Block,"Block Storage.  VMDK, iSCSI, Fibre Channel, Ceph RBD, OpenStack Cinder, Dell/EMC Scale.IO, AWS EBS ^(1)^, GCE Persistent Disk ^(1)^, Azure Disk","RedHat, Seagate, WesternDigital, Toshiba, HGST, Intel, Samsung, OCZ, Plextor, PNY, Kingston, SanDisk, AWS, Azure, GCE, VMware, Dell/EMC, HPE, Others"
+File,"Network Attached File Storage.  CNS GlusterFS ^(1)^, GlusterFS ^(1)^, RHEL NFS, NetApp NFS ^(2)^, Azure File, Vendor NFS, Vendor GlusterFS ^(3)^, Loopback Image over NFS","RedHat, NetApp, Azure, Others"
 Object,"Object Storage.  Ceph Object Storage, OpenStack Swift,  AWS S3, Google Cloud Storage, Vendor S3 ^(3)^, Vendor Swift ^(3)^","RedHat, AWS, GCE, Others"
 |=======================
-NOTE: ^(1)^ Ceph RBD, OpenStack Cinder, AWS EBS, GCE Persistent Disk support Dynamic PV Provisioning natively in {product-title}.
+NOTE: ^(1)^ Ceph RBD, OpenStack Cinder, GlusterFS, AWS EBS, Azure Disk, GCE Persistent Disk support Dynamic PV Provisioning natively in {product-title}.
 
 NOTE: ^(2)^ NetApp NFS supports Dynamic PV Provisioning when using the Trident Plugin
 
@@ -36,16 +36,16 @@ WARNING: ^(3)^ Vendor GlusterFS, Vendor S3, and Vendor Swift supportability and 
 === Storage Backend Recommendations
 Below is a matrix that summarizes which storage backend types are Recommended, Supported, and Configurable for the given
 {product-title} application and useful attributes.
-[format="csv",width="70%",cols="10"]
+[format="csv",width="70%",cols="9"]
 [options="header]
 |=======================
-Type,Tied to Node?,ROM,RWX,Registry,Scaled Registry,etcd,Metrics,Logging,Apps
-hostPath,Yes,No,No,Not Configurable,Not Configurable,Recommended,Recommended,Recommended,Configurable
-Block,No^(1)^,Yes^(2)^,No,Supported,Configurable,Recommended,Recommended,Recommended,Recommended
-File,No,Yes^(2)^,Yes,Supported,Configurable,Configurable,Configurable,Configurable,Recommended
-Object,No,Yes,Yes,Recommended,Recommended,Not Configurable,Not Configurable,Not Configurable,Recommended^(3)^
+Type,Tied to Node?,ROM,RWX,Registry,Scaled Registry,Metrics,Logging,Apps
+hostPath,Yes,No,No,Configurable,Not Configurable,Recommended,Recommended,Configurable
+Block,No^(1)^,Yes^(2)^,No,Supported,Not Configurable,Recommended,Recommended,Recommended
+File,No,Yes^(2)^,Yes,Supported,Configurable,Configurable,Configurable,Recommended
+Object,No,Yes,Yes,Recommended,Recommended,Not Configurable,Not Configurable,Recommended^(3)^
 |=======================
-NOTE: ^(1)^ This does not apply to HDD, SSD, NVMe, Raw Physical Disk, VM Physical Disk, VMDK, and LVM.  These storage devices are tied to the nodes.
+NOTE: ^(1)^ This does not apply to VMDK.  This storage type is tied to the node.
 
 NOTE: ^(2)^ This does not apply to Physical Disk, VM Physical Disk, VMDK, Loopback over NFS, AWS EBS, and Azure Disk.
 
@@ -56,16 +56,13 @@ NOTE: ^(3)^Object Storage is not used via {product-title}'s PVs.  Apps must inte
 In a non-scaled/HA {product-title} Internal Docker Registry cluster deployment, the preferred Storage Backend is Ojbect Storage
 followed by Block Storage.  The Storage Backend does not need to support ReadWriteMany Access Mode.  The Storage Backend *must
 ensure Read-After-Write-Consistency*.  All Network Attached File Storage Backends are expressly *not recommended* for Internal
-Docker Registry cluster deployment.
+Docker Registry cluster deployment.  While hostPath volumes are configurable for a non-scaled/HA {product-title} Internal Docker Registry cluster deployment, they are not recommended.
 
 ===== Scaled Registry
 In a scaled/HA {product-title} Internal Docker Registry cluster deployment, the preferred Storage Backend is Ojbect Storage.
 The Storage Backend *must support ReadWriteMany* Access Mode and *must ensure Read-After-Write-Consistency*.  Block Storage is
 expressly *not recommended* for a scaled/HA {product-title} Internal Docker Registry cluster deployment.  All Network File
 Storage Backends are expressly *not recommended* for Internal Docker Registry cluster deployment.
-
-===== etcd
-For etcd, the preferred Storage Backend is Block Storage.  For the best etcd reliability, the lowest consistent latency Storage Backend is preferrable.
 
 ===== Metrics
 In {product-title} Hosted Metrics cluster deployment, the preferred Storage Backend is Block Storage.  All Network Attached File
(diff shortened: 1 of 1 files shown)


lpsantil:
@sabre1041 @etsauer @cpitman @sferich888 @raffaelespazzoli @juanluisvaladas @dstockdreher

scaling_performance/optimizing_storage.adoc
```diff
@@ -13,6 +13,81 @@ toc::[]
...
+WARNING: ^(3)^ Vendor GlusterFS, Vendor S3, and Vendor Swift supportability and configurability may vary.
+
+=== Storage Backend Recommendations
+Below is a matrix that summarizes which storage backend types are Recommended, Supported, and Configurable for the given
+{product-title} application and useful attributes.
+[format="csv",width="70%",cols="10"]
+[options="header]
+|=======================
+Type,Tied to Node?,ROM,RWX,Registry,Scaled Registry,etcd,Metrics,Logging,Apps
+hostPath,Yes,No,No,Not Configurable,Not Configurable,Recommended,Recommended,Recommended,Configurable
```
juanluisvaladas: Hostpath is actually configurable for the registry for a single instance. I'd say instead not recommended or not HA.

scaling_performance/optimizing_storage.adoc
```diff
@@ -13,6 +13,81 @@ toc::[]
...
+== General Storage Guidance
+
+The following is a matrix which details the storage backend types {product-title} can be used with implementation examples and vendors.
+
+=== Available Storage Options
+[format="csv",width="70%",cols="3"]
+[options="header]
+|=======================
+Type,Examples,Vendors
+hostPath,"hostPath pod mounted volumes.  A specific type of Block Storage.  File paths on the node mounted onto running pods.",Others
```
juanluisvaladas: I'm not a native speaker but here I understand you assign to the pod a block device instead of passing it a path that is then bound to the pod.

scaling_performance/optimizing_storage.adoc
```diff
@@ -13,6 +13,81 @@ toc::[]
...
+===== etcd
+For etcd, the preferred Storage Backend is Block Storage.  For the best etcd reliability, the lowest consistent latency Storage Backend is preferrable.
+
+===== Metrics
+In {product-title} Hosted Metrics cluster deployment, the preferred Storage Backend is Block Storage.  All Network Attached File
+Storage Backends are expressly *not recommended* for Hosted Metrics cluster deployment.
+
+===== Logging
+In {product-title} Hosted Logging cluster deployment, the preferred Storage Backend is Block Storage.  All Network Attached File
+Storage Backends are expressly *not recommended* for Hosted Logging cluster deployment.
```
juanluisvaladas: Would it make sense to add a reference for NFS storage causing data corruption?

juanluisvaladas (commented):
Very good overall, there are a few things I think could be improved but that's definitely good enough to be merged.

scaling_performance/optimizing_storage.adoc
raffaelespazzoli: I conceptually disagree with putting hostpath at the same level of the other storage type.
as far as storage type goes hostpath is a block storage.
hostpath is a volume type, as a volume type I categorize it as a local storage.
the other relevant volume types in this discussion are
ephemeral storage (empty dir)  which is missing from your document and it's again block storage
and persistent volumes.

I suggest to talk about these concepts in a separate table.

scaling_performance/optimizing_storage.adoc
```diff
@@ -13,6 +13,81 @@ toc::[]
...
+File,"Network Attached File Storage.  CNS GlusterFS ^(1)^, GlusterFS ^(1)^, RHEL NFS, NetApp NFS ^(2)^, Vendor NFS, Vendor GlusterFS ^(3)^, Loopback Image over NFS","RedHat, NetApp, AWS, GCE, Others"
+Object,"Object Storage.  Ceph Object Storage, OpenStack Swift,  AWS S3, Google Cloud Storage, Vendor S3 ^(3)^, Vendor Swift ^(3)^","RedHat, AWS, GCE, Others"
+|=======================
+NOTE: ^(1)^ Ceph RBD, OpenStack Cinder, AWS EBS, GCE Persistent Disk support Dynamic PV Provisioning natively in {product-title}.
+
+NOTE: ^(2)^ NetApp NFS supports Dynamic PV Provisioning when using the Trident Plugin
+
+WARNING: ^(3)^ Vendor GlusterFS, Vendor S3, and Vendor Swift supportability and configurability may vary.
+
+=== Storage Backend Recommendations
```
raffaelespazzoli: as part of the application / storage recommendation discussion, you may want to also introduce this picture that is intended to show the possible combination of storage and compute
https://www.lucidchart.com/documents/edit/80e4f96d-2a84-4cbb-9608-19be6c0545c8

let me know if you have issues opening the link

scaling_performance/optimizing_storage.adoc
```diff
@@ -13,6 +13,81 @@ toc::[]
...
+Below is a matrix that summarizes which storage backend types are Recommended, Supported, and Configurable for the given
+{product-title} application and useful attributes.
+[format="csv",width="70%",cols="10"]
+[options="header]
+|=======================
+Type,Tied to Node?,ROM,RWX,Registry,Scaled Registry,etcd,Metrics,Logging,Apps
```
raffaelespazzoli: scaled registry on block storage? I don't think it works...

raffaelespazzoli: for etcd I assume you don't mean the etcd that openshift uses, maybe we should specify that. If that is so, why is it relevant for this discussion?

raffaelespazzoli: there shouldn't be more than one recommended option per column
logging and metrics should be recommended for block storage not hostpath

raffaelespazzoli: I would remove the app column, because the recommendation is really app dependent

scaling_performance/optimizing_storage.adoc
rootfs: hostPath is not meant for production, it is more a developer's favorite.

scaling_performance/optimizing_storage.adoc
```diff
@@ -13,6 +13,81 @@ toc::[]
...
+The following is a matrix which details the storage backend types {product-title} can be used with implementation examples and vendors.
+
+=== Available Storage Options
+[format="csv",width="70%",cols="3"]
+[options="header]
+|=======================
+Type,Examples,Vendors
+hostPath,"hostPath pod mounted volumes.  A specific type of Block Storage.  File paths on the node mounted onto running pods.",Others
+Block,"Block Storage. HDD, SSD, NVMe, Raw Physical Disk, VM Physical Disk, VMDK, LVM, iSCSI, Fibre Channel, Ceph RBD, OpenStack Cinder, Dell/EMC Scale.IO, AWS EBS ^(1)^, GCE Persistent Disk ^(1)^, Azure Disk","RedHat, Seagate, WesternDigital, Toshiba, HGST, Intel, Samsung, OCZ, Plextor, PNY, Kingston, SanDisk, AWS, GCE, VMware, Dell/EMC, HPE, Others"
+File,"Network Attached File Storage.  CNS GlusterFS ^(1)^, GlusterFS ^(1)^, RHEL NFS, NetApp NFS ^(2)^, Vendor NFS, Vendor GlusterFS ^(3)^, Loopback Image over NFS","RedHat, NetApp, AWS, GCE, Others"
```
rootfs: CephFS (still tech preview from RHS), Azure File (i.e. SMB), OpenStack Manila

scaling_performance/optimizing_storage.adoc
```diff
@@ -13,6 +13,81 @@ toc::[]
...
+Object,"Object Storage.  Ceph Object Storage, OpenStack Swift,  AWS S3, Google Cloud Storage, Vendor S3 ^(3)^, Vendor Swift ^(3)^","RedHat, AWS, GCE, Others"
+|=======================
+NOTE: ^(1)^ Ceph RBD, OpenStack Cinder, AWS EBS, GCE Persistent Disk support Dynamic PV Provisioning natively in {product-title}.
```
rootfs: Also Glusterfs, Azure Disk

scaling_performance/optimizing_storage.adoc
```diff
@@ -13,6 +13,81 @@ toc::[]
...
+
+The following is a matrix which details the storage backend types {product-title} can be used with implementation examples and vendors.
+
+=== Available Storage Options
+[format="csv",width="70%",cols="3"]
+[options="header]
+|=======================
+Type,Examples,Vendors
+hostPath,"hostPath pod mounted volumes.  A specific type of Block Storage.  File paths on the node mounted onto running pods.",Others
+Block,"Block Storage. HDD, SSD, NVMe, Raw Physical Disk, VM Physical Disk, VMDK, LVM, iSCSI, Fibre Channel, Ceph RBD, OpenStack Cinder, Dell/EMC Scale.IO, AWS EBS ^(1)^, GCE Persistent Disk ^(1)^, Azure Disk","RedHat, Seagate, WesternDigital, Toshiba, HGST, Intel, Samsung, OCZ, Plextor, PNY, Kingston, SanDisk, AWS, GCE, VMware, Dell/EMC, HPE, Others"
```
rootfs: Raw devices are not supported yet

scaling_performance/optimizing_storage.adoc
lpsantil: @juanluisvaladas It can be one or the other.  https://kubernetes.io/docs/concepts/storage/volumes/#hostpath

@rootfs Current Hosted Metrics and Logging documentation are (intentionally?) vague here.  Some have implemented those with hostPath when faced with having a cluster that otherwise only has access to NFS storage.  And it is used at scale.  Those implementers have also created labeling and scheduler affinity/anti-affinity rules to support that use-case.  Part of why that was done (probably) was that there was no other obvious guidance.  I'm hope this PR rectifies that for some future implementers.

lpsantil: @raffaelespazzoli Thanks for this.  I intended to only speak to Persistent Storage Types.  I need to speak with you (and others) more about ephemeral storage, because, while it may have benefits I assume its supportability is much different that Persistent Storage Types.  You can always add an entry to the spreadsheet I linked to as well.

scaling_performance/optimizing_storage.adoc
lpsantil: @rootfs I'll remove them.  However, I had a specific request to add VMDK.  Please see spreadsheet.

scaling_performance/optimizing_storage.adoc
lpsantil: @rootfs The guidance I received on CephFS was different.  Please see spreadsheet.   I see Azure File was added to https://docs.openshift.org/latest/architecture/additional_concepts/storage.html#types-of-persistent-volumes but OSP Manila was not.

scaling_performance/optimizing_storage.adoc
lpsantil: @rootfs Thanks!  Good catch.

scaling_performance/optimizing_storage.adoc
lpsantil: @raffaelespazzoli On hostPath, I agree it's a subtype of of Block Storage, and, like you mentioned, I wanted to call it out because it enjoy's its own Volume Type.

scaling_performance/optimizing_storage.adoc
lpsantil: @raffaelespazzoli ACK.

lpsantil: @raffaelespazzoli I did intend to reference the OCP internal etcd. However, you are right.  I will remove etcd from this table.

lpsantil: That's the goal.  However, hostPath is also a subtype of Block Storage.

lpsantil: I almost agree with you.  But want OCP to run apps.  I'd think they'd want to know if ever that a certain storage type is less than recommended or has a caveat.  And there are two in the table as it stands.

scaling_performance/optimizing_storage.adoc
lpsantil: Noted.

scaling_performance/optimizing_storage.adoc
lpsantil: I'll add a note later in the text unless you think it should added somewhere here.

childsb:
References to block storage should be changed to file-on-block since raw block storage isn't currently a feature (but will be sometime in the future).

lpsantil:
@childsb Please, pound on this PR like a rock polisher until it's shiny.

I get that the language is not as specific as it could be and that the apps would only ever see files & file paths.  I've tried to write it at a level where a new(er) OCP administrator would be familiar with.  In a virtualized & contrainerized world, almost everything gets accessed over a network.  But if I said Block Device Storage, you'd understand I wasn't talking about NFS or SMB and you might know LUNs, iSCSI & Fibre Channel, etc.  And if you're an AWS Admin, you'd probably make the connection to EBS.  

Yeah, I get that Object Storage is something that isn't accessed over a PV and so I'm lumping storage types together.  But the Internal Docker Registry can use all of these type and have very different performance characteristics.  What's worse is that it's not even clear that you can use all of these types.  The docs currently say "attach a remote volume" [0].   None the wiser that the first type they have to choose from, NFS [1] is actually not supported.  More confusing is that the very next section in [0] shows how to define an AWS S3 bucket for use with the Internal Docker Registry even though it is not listed as the supported in [1] and it is even the preferred storage type if available.

In short, if you have a better way of organizing the story, I'm all ears and ready to write it that way.  Clearer docs is my goal and the subject is already a minefield. Please also look the internal spreadsheet I've started.


[0] https://docs.openshift.com/container-platform/3.5/install_config/registry/deploy_registry_existing_clusters.html#registry-production-use

[1] https://docs.openshift.com/container-platform/3.5/install_config/registry/deploy_registry_existing_clusters.html#registry-production-use

scaling_performance/optimizing_storage.adoc
lpsantil: I can open the link but I think I need your guidance in understanding some of the details.

lpsantil:
So I found that our docs are using [0] Shared Storage and Block Storage.  Is that a bit better?  Object Storage isn't even mentioned.

[0] https://docs.openshift.com/container-platform/3.5/security/storage.html

scaling_performance/optimizing_storage.adoc
cpitman: How is it block storage? Does OCP modify permissions on host mount storage? If not, then it isn't block.

tmorriso-rh:
@lpsantil Hi, I am following up on this PR. Is this information still valid? From what I can tell, the information just needs to be reviewed and edited. Please advise. Thanks.

tmorriso-rh:
This is now a new PR: https://github.com/openshift/openshift-docs/pull/5623

PR #4536, opened 2017-06-05
Add documentation for build pruning

coreydaley:
Provides supporting documentation for https://github.com/openshift/origin/pull/13788 and https://github.com/openshift/origin/pull/14375

@openshift/devex ptal

PR #4544, opened 2017-06-06
Follow-up to PR4536: Minor edits

bmcelvee:
Minor edits to https://github.com/openshift/openshift-docs/pull/4536

### DIFF
--- a/dev_guide/builds/advanced_build_operations.adoc
+++ b/dev_guide/builds/advanced_build_operations.adoc
@@ -269,7 +269,7 @@ with two stages.
 By default, builds that have completed their lifecycle are persisted indefinitely.
 You can limit the number of previous builds that are retained by supplying a positive
 integer value for `successfulBuildsHistoryLimit` or `failedBuildsHistoryLimit`
-as shown in the below sample build configuration.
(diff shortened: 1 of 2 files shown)


dev_guide/builds/advanced_build_operations.adoc
```diff
@@ -281,18 +281,17 @@ spec:
...
- - a build completes its lifecycle
+ - Updating a build configuration.
+ - A build completes its lifecycle.
 
 Builds are sorted by their creation timestamp with the oldest builds being pruned first.
 
 [NOTE]
 ====
 Administrators can manually prune builds using the xref:../../admin_guide/pruning_resources.adoc#pruning-builds[oadm object pruning command].
 ====
```
adellape: @bmcelvee This note box should be ifdef'd for just OCP, Origin, and (maybe?) Dedicated. This would be a broken link for Online since it doesn't have an `admin_guide`.

Also `oadm` could be marked up in the link.

bmcelvee: Ah yes, thanks @adellape!

PR #4548, opened 2017-06-07
Fix the number of dynamic provision plug-ins that we support

liangxia:
See https://docs.openshift.org/latest/install_config/persistent_storage/dynamically_provisioning_pvs.html#available-dynamically-provisioned-plug-ins


liangxia:
Apply to branch/enterprise-3.6

ahardin-rh:
Thank you!

ahardin-rh:
No revision history needed

PR #4559, opened 2017-06-09
Fix PHP Laravel quickstart name

Miciah:
The template for Laravel was renamed from "laravel-mysql-example" to "laravel-mysql-persistent", so update the example that references the template by the old name.


ahardin-rh:
@Miciah Thanks!

ahardin-rh:
This update only appears in the Online doc (where no revision history is maintained), so no revision history is needed here.

PR #4560, opened 2017-06-09
Document that you may need to install etcd package in order to run

sdodson:
etcdctl from the masters

If your master isn't also an etcd host then it won't be installed there. Otherwise everything is documented correctly.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1372469

### DIFF
--- a/install_config/install/advanced_install.adoc
+++ b/install_config/install/advanced_install.adoc
@@ -1523,6 +1523,13 @@ The default port for the console is `8443`. If this was changed during the insta
 
 If you installed multiple *etcd* hosts:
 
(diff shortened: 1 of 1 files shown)


ahardin-rh:
@sdodson Thanks!

ahardin-rh:
[rev_history]
|xref:../install_config/install/advanced_instal.adoc#install-config-install-advanced-install[Installing a Cluster -> Advanced Installation]
|Added a step to verify that the *etcd* package is installed, If you installed multiple *etcd* hosts.
%

PR #4586, opened 2017-06-15
Followup edits to networking file as per 4277

bfallonf:
As per: 
* https://trello.com/c/I6Is4843/490-document-add-an-http-proxy-mode-for-egress-router
* https://trello.com/c/PvAdc0CK/499-document-support-multiple-destinations-in-egress-router-egress

Will ask for an ack over email.

### DIFF
--- a/admin_guide/managing_networking.adoc
+++ b/admin_guide/managing_networking.adoc
@@ -13,8 +13,9 @@ toc::[]
 
 == Overview
 
-This topic describes the management of the overall xref:../architecture/additional_concepts/networking.adoc#architecture-additional-concepts-networking[cluster network],
-including project isolation and outbound traffic control.
+This topic describes the management of the overall
+xref:../architecture/additional_concepts/networking.adoc#architecture-additional-concepts-networking[cluster
+network], including project isolation and outbound traffic control.
 
 Pod-level networking features, such as per-pod bandwidth limits, are discussed
 in xref:../admin_guide/managing_pods.adoc#admin-guide-manage-pods[Managing
@@ -320,33 +321,33 @@ falling back to filtering MAC addresses in software.
 ====
 
 [[admin-guide-limit-pod-access-important-deployment-considerations]]
-==== Important Deployment Considerations
+*Deployment Considerations*
 
 The Egress router adds a second IP address and MAC address to the node's primary
 network interface. If you are not running {product-title} on bare metal, you may
 need to configure your hypervisor or cloud provider to allow the additional
 address.
 
-*Red Hat OpenStack Platform*
+Red Hat OpenStack Platform::
 
 If you are deploying {product-title} on Red Hat OpenStack Platform, you need to
-whitelist the IP and MAC addresses on your Openstack environment, otherwise
+whitelist the IP and MAC addresses on your OpenStack environment, otherwise
 link:https://access.redhat.com/solutions/2803331[communication will fail]:
-
++
 ----
 neutron port-update $neutron_port_uuid \
   --allowed_address_pairs list=true \
   type=dict mac_address=<mac_address>,ip_address=<ip_address>
 ----
 
-*Red Hat Enterprise Virtualization*
+Red Hat Enterprise Virtualization::
 
 If you are using
 link:https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Virtualization/3.2/html/Administration_Guide/Red_Hat_Enterprise_Virtualization_Manager_configuration_options_explanations_limitations_and_best_practices.html[Red
 Hat Enterprise Virtualization], you should set
 `EnableMACAntiSpoofingFilterRules` to `false`.
 
-*VMware vSphere*
+VMware vSphere::
 
 If you are using VMware vSphere, follow VMware's
 link:http://pubs.vmware.com/vsphere-4-esxi-installable-vcenter/index.jsp?topic=/com.vmware.vsphere.esxi_server_config.doc_40_u1/esx_server_config/securing_an_esx_configuration/c_securing_virtual_switch_ports.html[Securing
@@ -355,29 +356,27 @@ link:http://pubs.vmware.com/vsphere-4-esxi-installable-vcenter/index.jsp?topic=/
 Transmissions] guidance.
 
 [[admin-guide-egress-router-modes]]
-==== Egress Router Modes
+*Egress Router Modes*
 
 The egress router can run in two different modes:
+xref:admin-guide-deploying-an-egress-router-pod[redirect mode] and
+xref:admin-guide-deploying-an-egress-router-http-proxy-pod[HTTP proxy mode].
+Redirect mode works for all services except for HTTP and HTTPS. For HTTP and
+HTTPS services, use HTTP proxy mode.
 
-1. A "redirect" mode, where the egress router sets up iptables rules
-to redirect traffic from its own IP address to one or more destination
-IP addresses. Client pods that want to make use of the reserved source
-IP address will need to be modified to connect to the egress router
-rather than connecting directly to the destination IP.
-
-2. An HTTP proxy mode, where the egress router runs as an HTTP proxy.
-This only works for clients talking to HTTP/HTTPS-based services, but
-usually requires fewer changes to the client pods to get them to work.
-(Many programs can be told to use an HTTP proxy just by setting an
-environment variable.)
 
 [[admin-guide-deploying-an-egress-router-pod]]
-==== Deploying a Redirecting Egress Router Pod
+==== Deploying an Egress Router Pod in Redirect Mode
 
-. Create a pod configuration such as the following:
+In _redirect mode_, the egress router sets up iptables rules to redirect traffic
+from its own IP address to one or more destination IP addresses. Client pods
+that want to make use of the reserved source IP address must be modified to
+connect to the egress router rather than connecting directly to the destination
+IP.
+
+. Create a pod configuration using the following:
 +
-.Example Pod Definition for a simple redirecting Egress Router
-====
+[source, yaml]
 ----
 apiVersion: v1
 kind: Pod
@@ -421,19 +420,18 @@ endif::openshift-origin[]
 <1> The `pod.network.openshift.io/assign-macvlan annotation` creates a Macvlan
 network interface on the primary network interface, and then moves it into the
 pod's network name space before starting the *egress-router* container. Preserve
-the the quotation marks around `"true"`. Omitting them will result in errors.
-<2> An IP address from the physical network that the node itself is on and is
+the quotation marks around `"true"`. Omitting them results in errors.
+<2> IP address from the physical network that the node is on and is
 reserved by the cluster administrator for use by this pod.
-<3> Same value as the default gateway used by the node itself.
+<3> Same value as the default gateway used by the node.
 <4> The external server to direct traffic to. Using this example,
 connections to the pod are redirected to 203.0.113.25, with a source IP address
 of 192.168.12.99.
 <5> This tells the egress router image that it is being deployed as an
-"init container"; older versions of {product-title} (and the egress
+"init container". Previous versions of {product-title} (and the egress
 router image) did not support this mode and had to be run as an
 ordinary container.
-<6> The pod will only be deployed to nodes with the label `site=springfield-1`.
-====
+<6> The pod is only deployed to nodes with the label `site=springfield-1`.
 
 . Create the pod using the above definition:
 +
@@ -449,7 +447,7 @@ oc get pod <pod_name>
 
 . Ensure other pods can find the pod's IP address by creating a service to point to the egress router:
 +
-====
+[source, yaml]
 ----
 apiVersion: v1
 kind: Service
@@ -465,14 +463,12 @@ spec:
   selector:
     name: egress-1
 ----
-====
 +
 Your pods can now connect to this service. Their connections are redirected to
 the corresponding ports on the external server, using the reserved egress IP
 address.
 
-The work of setting up the egress router is done as an "init
-container" by the
+The egress router setup is performed by an "init container" created from the
 ifdef::openshift-enterprise[]
 *openshift3/ose-egress-router*
 endif::openshift-enterprise[]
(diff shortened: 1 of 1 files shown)


bfallonf:
Updated. @danwinship PTAL

admin_guide/managing_networking.adoc
```diff
@@ -306,29 +306,25 @@ link:http://pubs.vmware.com/vsphere-4-esxi-installable-vcenter/index.jsp?topic=/
...
 
-The egress router can run in two different modes:
+The egress router can run in two different modes: xref:admin-guide-deploying-an-egress-router-pod[redirect mode] and xref:admin-guide-deploying-an-egress-router-http-proxy-pod[HTTP proxy mode].
 
-1. A "redirect" mode, where the egress router sets up iptables rules
-to redirect traffic from its own IP address to one or more destination
-IP addresses. Client pods that want to make use of the reserved source
-IP address will need to be modified to connect to the egress router
-rather than connecting directly to the destination IP.
+Use HTTP proxy mode for HTTP and HTTPS services. For all other types of connections, redirect mode is recommended.
```
danwinship: It's not just "recommended"; the HTTP proxy mode won't work at all for non-HTTP-based services

danwinship (commented):
basically lgtm

bfallonf:
Thanks @danwinship . I changed to be more precise.

@gaurav-nelson peer review please?

admin_guide/managing_networking.adoc
```diff
@@ -276,14 +276,14 @@ falling back to filtering MAC addresses in software.
...
 The Egress router adds a second IP address and MAC address to the node's primary
 network interface. If you are not running {product-title} on bare metal, you may
 need to configure your hypervisor or cloud provider to allow the additional
 address.
 
-*Red Hat OpenStack Platform*
+Red Hat OpenStack Platform::
 
 If you are deploying {product-title} on Red Hat OpenStack Platform, you need to
 whitelist the IP and MAC addresses on your Openstack environment, otherwise
```
gaurav-nelson: We should also add `+` before the code block to align it with the description.

```
link:https://access.redhat.com/solutions/2803331[communication will fail]:
+
----
neutron port-update $neutron_port_uuid \

```

admin_guide/managing_networking.adoc
```diff
@@ -377,19 +375,18 @@ endif::openshift-origin[]
 <1> The `pod.network.openshift.io/assign-macvlan annotation` creates a Macvlan
 network interface on the primary network interface, and then moves it into the
 pod's network name space before starting the *egress-router* container. Preserve
-the the quotation marks around `"true"`. Omitting them will result in errors.
-<2> An IP address from the physical network that the node itself is on and is
+the the quotation marks around `"true"`. Omitting them results in errors.
```
gaurav-nelson: remove repeated `the`

admin_guide/managing_networking.adoc
```diff
@@ -461,13 +456,11 @@ source IP address and using the specified gateway, you can specify a
 [[admin-guide-manage-pods-egress-router-multi-destination]]
 ==== Redirecting to Multiple Destinations
 
-In the example above, connections to the egress pod (or its
-corresponding service) on any port will be redirected to a single
-destination IP. It is also possible to configure different destination
-IPs depending on the port:
+In the example above, connections to the egress pod (or its corresponding
```
gaurav-nelson: Do not use to indicate a relative location in a document. Use "previous" or "preceding."

admin_guide/managing_networking.adoc
```diff
@@ -535,14 +527,11 @@ then connections on other ports would simply be rejected.)
...
 
-In the case of a large or frequently-changing set of destination
-mappings, it may be useful to maintain the list externally in a
-ConfigMap, and have the egress router pod read it from there. This
-also has the advantage that project administrators would be able to
-edit the ConfigMap (whereas they may not be able to edit the Pod
-definition directly, since it contains a privileged container).
-
-To do this:
+In the case of a large or frequently-changing set of destination mappings, you
```
gaurav-nelson: s/In the case of/For - BRICE method :smile_cat:

admin_guide/managing_networking.adoc
```diff
@@ -592,18 +583,21 @@ ConfigMap for `EGRESS_DESTINATION` in the environment section:
     ...
 ----
 
-Note that the egress router will not automatically update when the
-ConfigMap changes; you will need to restart the pod to get updates.
+[NOTE]
+====
+The egress router does automatically update when the ConfigMap changes. Restart
```
gaurav-nelson: Maybe `The egress router automatically updates when...`

admin_guide/managing_networking.adoc
```diff
@@ -592,18 +583,21 @@ ConfigMap for `EGRESS_DESTINATION` in the environment section:
...
 
 [[admin-guide-deploying-an-egress-router-http-proxy-pod]]
 ==== Deploying an Egress Router HTTP Proxy Pod
 
-Creating an egress router in HTTP proxy mode is similar to creating a
-redirecting egress router, as described above, but with a slightly
-different specification:
+In _HTTP proxy mode_, the egress router runs as an HTTP proxy. This only works
+for clients talking to HTTP or HTTPS-based services, but usually requires fewer
+changes to the client pods to get them to work. (Many programs can be told to
```
gaurav-nelson: We can remove `Many` without changing the meaning.

admin_guide/managing_networking.adoc
```diff
@@ -660,24 +654,22 @@ redirecting rules.
...
 HTTP proxy container, not with the other environment variables in the
 init container.
-====
 
-You can specify a variety of things for
-`EGRESS_HTTP_PROXY_DESTINATION`. The simplest is to just specify `*`,
-meaning "allow connections to all remote destinations". Other than
-that, each line specifies one group of connections to allow or deny:
+You can specify any of the following for the `EGRESS_HTTP_PROXY_DESTINATION`
+value. However, you can also use `*`, meaning "allow connections to all remote
```
gaurav-nelson: Maybe we can remove the word `however`.

gaurav-nelson:
@bfallonf added few comments. Rest is all A-OK :1st_place_medal:

bfallonf:
Thanks @gaurav-nelson . All done as suggested. Will merge.

bfallonf:
[rev_history]
|Added information about the HTTP proxy and redirect modes for egress services.
%

admin_guide/managing_networking.adoc
danwinship: Oops, no, that should have been "The egress router does NOT automatically update when...". The "not" got lost in the rewording.

bfallonf: Easy update. Thanks @danwinship !

admin_guide/managing_networking.adoc
```diff
@@ -670,24 +666,55 @@ redirecting rules.
...
+  ports:
+  - name: http
+    port: 8080 <1>
+  - name: https
+    port: 443
+  type: ClusterIP
+  selector:
+    name: egress-1
+----
+<1> Ensure the `http` port is always set to `8080`.
```
bfallonf: @weliang1 @danwinship I added the above. It's similar to the service example in the section about redirecting mode.

admin_guide/managing_networking.adoc
```diff
@@ -670,24 +666,55 @@ redirecting rules.
...
++
+[source, yaml]
+----
+    ...
+    env:
+    - name: http_proxy
+      value: http://egress-1:8080/ <1>
+    ...
+----
+<1> The service created in step 2.
```
bfallonf: @weliang1 @danwinship And I've added this as well.

bfallonf:
@weliang1 @danwinship As a followup to the comment in #4277 , I've added the two new sections I've tagged here. Let me know what you think.

admin_guide/managing_networking.adoc
danwinship: I'd call the 8080 port "http-proxy" rather than "http". And there shouldn't be a 443 port

admin_guide/managing_networking.adoc
```diff
@@ -670,24 +666,55 @@ redirecting rules.
...
+----
+<1> Ensure the `http` port is always set to `8080`.
 
-- An IP address (eg, `192.168.1.1`) will allow connections to that IP address.
-- A CIDR range (eg, `192.168.1.0/24`) will allow connections to that CIDR range.
-- A hostname (eg, `www.example.com`) will allow proxying to that host.
-- A domain name preceded by `\*.` (eg, `*.example.com`) will allow proxying to that domain and all of its subdomains.
-- A `!` followed by any of the above will deny connections rather than allowing them
-- If the last line is `*`, then anything that hasn't been denied will be allowed. Otherwise, anything that hasn't been allowed will be denied.
+. In the pod configuration, set the service as a variable:
```
danwinship: "the pod configuration" is ambiguous; you want to set this in the *client* pod (ie, not the egress proxy pod).

Also, this is not accurate for 100% of cases. It needs to say something like "Configure the client pod to use the http proxy. In many cases, this can be done by setting the `http_proxy` and `https_proxy` variables in the pod's environment"

admin_guide/managing_networking.adoc
```diff
@@ -670,24 +666,55 @@ redirecting rules.
...
++
+[source, yaml]
+----
+    ...
+    env:
+    - name: http_proxy
+      value: http://egress-1:8080/ <1>
```
danwinship: also

    - name: https_proxy
      value: http://egress-1:8080/

(the same as above except "https_proxy" instead of "http_proxy")

admin_guide/managing_networking.adoc
bfallonf: @danwinship @weliang1 I like the suggestion, but I'm hesitant to put in "in many cases" but not saying anything about the other cases. Is there something that would work for other cases? Maybe a note box after the step saying something like "This will not work for all cases, because X"?

admin_guide/managing_networking.adoc
bfallonf: @danwinship @weliang1 So, delete the whole https section, or just replace the 443 with 8080? If I'm adding the `https-proxy` variable below, does that mean it's required here as well?

admin_guide/managing_networking.adoc
danwinship: The `http_proxy`/`https_proxy` environment variable is a very common convention, but there's nothing that *requires* apps to use it. In the case of apps/commands that don't use those variables you would probably either need to pass a command-line option or set some option in a config file.

You can probably get away with "in most cases" rather than "in many cases". All you can do in the other cases is refer them to the documentation for whatever tool it is that they're using in the pod.

Maybe start out just saying to use the environment variables, and then afterward have a note about "if this doesn't work then consult the documentation for the software you are running in the pod to see how to configure it to use an HTTP proxy".

admin_guide/managing_networking.adoc
danwinship: just drop the https section:

    ...
    spec:
      ports:
      - name: http-proxy
        port: 8080 <1>
      type: ClusterIP
    ...

The proxy runs on one port, 8080, which handles both http and https proxying. (There are two separate environment variables that need to be specified because those environment variables are a generic thing, not an OpenShift-specific or egress-router-specific thing, and in some other HTTP proxying scenarios, people want to be able to set them to different things. But in our scenario, all users will always want to set both variables to point to 8080 on the egress proxy.)

bfallonf: Cool. Thanks for the info @danwinship

admin_guide/managing_networking.adoc
bfallonf: Thanks @danwinship I've put in a note box with: "Using the `http_proxy` and `https_proxy` environment variables is not necessary for all setups. If the above does not create a working setup, then consult the documentation for the tool or software you are running in the pod."

bfallonf:
Thanks @danwinship @weliang1 . I'll this back to on QA.

bmeng:
The .spec.initContainers.image is a required value for all the pod template. 
eg:
```
apiVersion: v1
kind: Pod
metadata:
  name: egress-1
  labels:
    name: egress-1
  annotations:
    pod.network.openshift.io/assign-macvlan: "true"
spec:
  initContainers:
  - name: egress-router
    image: openshift3/ose-egress-router:$releasever    << required filed in template
    securityContext:
```

Seems they were missed in old commits. But we should add them

danwinship:
> The .spec.initContainers[].image and .spec.containers[].image is a required value for all the pod template.
> ...
> Seems they were missed in old commits.

Oh, yeah, actually this is because the text has:

    ifdef::openshift-enterprise[]
        image: registry.access.redhat.com/openshift3/ose-egress-router
    endif::openshift-enterprise[]
    ifdef::openshift-origin[]
        image: openshift/origin-egress-router
    endif::openshift-origin[]

and so if you read the atomic or dedicated docs (or even https://github.com/openshift/openshift-docs/blob/master/admin_guide/managing_networking.adoc), both ifdefs are false and so you get nothing. I guess there needs to be a fallback "else" clause or something

bfallonf:
Yes, thanks for clearing that up @danwinship . @bmeng please let me know if there's anything else for this PR.

(There was also conflicts, so I fixed those and repushed.)

bmeng:
LGTM

bfallonf:
Thank you @bmeng . I'll merge away.

bfallonf:
[rev_history]
|Added the xref:../admin_guide/managing_networking.adoc#admin-guide-egress-router-modes[Egress Router Modes], xref:../admin_guide/managing_networking.adoc#admin-guide-manage-pods-egress-router-multi-destination[Redirecting to Multiple Destinations], xref:../admin_guide/managing_networking.adoc#admin-guide-manage-pods-egress-router-configmap[Using a ConfigMap to specify EGRESS_DESTINATION], and xref:../admin_guide/managing_networking.adoc#admin-guide-deploying-an-egress-router-pod[Deploying an Egress Router Pod in Redirect Mode] sections, as well as various content changes.
%

PR #4612, opened 2017-06-20
Add generators to be used with oc run command

soltysh:
This adds a section to `oc run` doc explaining different generators that can be used with this command

@adellape ptal


adellape:
Continuing via https://github.com/openshift/openshift-docs/pull/4619, PTAL there.

PR #4626, opened 2017-06-21
Bug 1454622 Added the Configuring the Firewall section

bfallonf:
As per: https://bugzilla.redhat.com/show_bug.cgi?id=1454622

Will ask for technical ack in the BZ.

### DIFF
--- a/install_config/install/advanced_install.adoc
+++ b/install_config/install/advanced_install.adoc
@@ -223,6 +223,9 @@ All].
 validity for etcd CA, peer, server and client certificates. Defaults to `1825`
 (5 years).
 
+|`os_firewall_use_firewalld`
+|Set to `true` to use firewalld instead of the default iptables. Not available on RHEL Atomic Host. See the xref:advanced-install-configuring-firewalls[Configuring the Firewall] section for more information.
+
 |`openshift_master_session_name`
 .4+.^|These variables override defaults for
 xref:../../install_config/configuring_authentication.adoc#session-options[session
@@ -757,6 +760,41 @@ build, defined using the `*BuildDefaults*` admission controller. If
 value; you only need to set this if you want your `git clone` operations to use
 a different value.
 |===
+
+
+ifdef::openshift-enterprise,openshift-origin[]
+[[advanced-install-configuring-firewalls]]
+=== Configuring the Firewall
+
+[IMPORTANT]
+====
+If you are changing the default firewall, ensure that each host in your cluster
+is using the same firewall type to prevent inconsistencies.
+====
+
+[NOTE]
+====
+While iptables is the default firewall, firewalld is recommended for new
+installations.
+====
+
+{product-title} uses iptables as the default firewall, but you can configure
+your cluster to use firewalld during the install process.
+
+Because iptables is the default firewall, {product-title} is designed to have it
+configured automatically. However, iptables rules can break {product-title} if
+not configured correctly. The advantages of firewalld include allowing multiple
+objects to safely share the firewall rules.
+
+To use firewalld as the firewall for an {product-title} installation, add the
+`os_firewall_use_firewalld` variable to the list of configuration variables in
+the Ansible host file at install:
+
+----
+[OSEv3:vars]
+os_firewall_use_firewalld=True
+----
+endif::[]
 
 [[marking-masters-as-unschedulable-nodes]]
 === Configuring Schedulability on Masters


sdodson:
I'd just make sure to mention that firewalld cannot be used on Atomic Host, otherwise looks good to me.

knobunc (changes requested):
The changes present look good.  But it would be nice to say what the advantages / disadvantages are.

iptables: Pros: Always there.  Cons: If something else changes the iptables rules it can break OpenShift unless done carefully (link to the iptables docs)

firewalld: Pros: allows multiple things to safely share the firewall rules.  Cons: Not on Atomic Host, and since it was the old default, existing nodes will need to be changed to enable it properly.

bfallonf:
Thanks @knobunc . I was after a bit more context myself. Not too sure what you mean by "since [firewalld] was the old default". My understanding is that iptables is the default and firewalld is optional. Was this not always the case?

cc: @sdodson

sdodson:
@bfallonf For a while, the unreleased versions of the 3.5 installer defaulted to firewalld but we reverted that change just before release. So at no time has firewalld been the default in a released version.

bfallonf:
@sdodson Thanks for that. Sounds like what I have is best then. @knobunc , please let me know if there's anything else for this PR, or just ack the changes. Thanks, all!

install_config/install/advanced_install.adoc
```diff
@@ -223,6 +223,9 @@ All].
 validity for etcd CA, peer, server and client certificates. Defaults to `1825`
 (5 years).
 
+|`os_firewall_use_firewalld`
+|Set to `true` to use firewalld instead of the default iptables. Not available on RHEL Atomic Host. See the xref:advanced-install-configuring-firewalls[Configuring the Firewall] section for more information.
```
knobunc: This will change soon, they plan to add it to Atomic Host.  Is there a way to track this so we re-check next time AH releases?

install_config/install/advanced_install.adoc
```diff
@@ -617,6 +620,26 @@ value; you only need to set this if you want your `git clone` operations to use
 a different value.
 |===
 
+[[advanced-install-configuring-firewalls]]
+=== Configuring the Firewall
+
+{product-title} uses iptables as the default firewall, but you can switch to use firewalld in the install.
+
+Because iptables is the default filewall, {product-title} is designed to have it
```
knobunc: Careful here... firewalld is the default recommended firewall in RHEL.  How about:

"Because `iptables` the previous default firewall, {product-title} still sets it as the default so adding a node to an existing cluster doesn't cause inconsistencies."

install_config/install/advanced_install.adoc
```diff
@@ -617,6 +620,26 @@ value; you only need to set this if you want your `git clone` operations to use
...
+configured automatically. However, iptables rules can break {product-title} if
+not configured correctly. The advantages of firewalld include allowing multiple
+objects to safely share the firewall rules. However, it is not available on RHEL
```
knobunc: Too many howevers.  Perhaps make this a "Please note that ..."

install_config/install/advanced_install.adoc
```diff
@@ -617,6 +620,26 @@ value; you only need to set this if you want your `git clone` operations to use
...
+Atomic Host.
+
+To use firewalld as the firewall for an {product-title} install, add the
```
knobunc: `firewalld` ?

install_config/install/advanced_install.adoc
```diff
@@ -617,6 +620,26 @@ value; you only need to set this if you want your `git clone` operations to use
...
+`os_firewall_use_firewalld` variable to the list of configuration variables in
+Ansible host file on install:
+
+----
+[OSEv3:vars]
+os_firewall_use_firewalld=True
+----
+
```
knobunc: Can we say that we recommend that new installs use firewalld rather than iptables.

knobunc (changes requested):

install_config/install/advanced_install.adoc
bfallonf: @knobunc This goes against what @sdodson says above: "the unreleased versions of the 3.5 installer defaulted to firewalld but we reverted that change just before release. So at no time has firewalld been the default in a released version."

Am I misunderstanding this? iptables is still the default, right?

install_config/install/advanced_install.adoc
bfallonf: Added a note box.

install_config/install/advanced_install.adoc
bfallonf: @knobunc Sorry, not sure what you're meaning here.

install_config/install/advanced_install.adoc
bfallonf: @knobunc @sdodson Not sure about this one either. If it's not the default, then why recommend it? I'll see what the response above is.

install_config/install/advanced_install.adoc
sdodson: Right, iptables is still the default.

install_config/install/advanced_install.adoc
sdodson: The way this reads now is fine with me, firewalld is not the default, it has advantage X,Y,Z but is not available on atomic host, if you want to use it set this value.

As far as whether or not we should be recommending people enable it, I'm going to defer to the networking team and Ben Breard here. We were told to revert the installer back to using iptables by default because firewalld wasn't in Atomic Host and we had no tooling to migrate people from iptables to firewalld. So I'm not sure if we should be recommending use of firewalld or not.

knobunc: @sdodson: I thought there were two problems:

1) Atomic Host didn't have firewalld (but that will be fixed soon)
2) If you added a node to an existing cluster, the old nodes would have raw iptables, but the new would default to firewalld meaning you would end up with a mixed cluster

I thought 2 was the reason we reverted.  But if you are installing a new cluster it would not be a problem.

@danwinship @dcbw @openshift/networking -- What are your thoughts on this?

install_config/install/advanced_install.adoc
knobunc: Sorry, should firewalld be in backticks?

install_config/install/advanced_install.adoc
knobunc: I'm not sure how my proposal goes against it.  I'm just saying that since `iptables` was the old default, that is still maintained as the default in current versions so that you don't end up with a mixed cluster when adding nodes...

install_config/install/advanced_install.adoc
danwinship: We definitely want to move to a world where everyone is using firewalld; OpenShift+iptables-service is always going to be flaky (and is less shiny-nftables-future-proof too).

From our POV, having most people using firewalld but some still using iptables-service for whatever reason is better than having everyone using iptables-service (because it means fewer people shooting themselves in the foot). But maybe for other teams (installer, docs, etc) it's better to have everyone using the same thing...

sdodson: Can the networking team help provide a migration tooling for migrating people from iptables to firewalld on upgrade?

danwinship: Well, a first step would be to just only worry about people who never changed the iptables-service config from what openshift-ansible originally installed. In that case you don't need to migrate per se, just throw out the old iptables-service config and replace it with a new firewalld config.

For people actually using the service, I'm not sure we want to try to auto-upgrade them, although it would be theoretically possible

install_config/install/advanced_install.adoc
bfallonf: @knobunc I think I understand what you're saying now. "Even though iptables is the default, firewalld is recommended, but OpenShift still uses iptables even though firewalld is recommended, so don't have mixed clusters." I think it was words like "previous default" and "old default" that put me off... I added a admonition to the start of the section:

"If you are changing the default firewall, ensure that each host in your cluster is using the same firewall type to prevent inconsistencies."

I think anything else would verge on saying something that's already there.

install_config/install/advanced_install.adoc
bfallonf: I think in this case, it's using the proper case that's the issue. Changed to "FirewallD".

install_config/install/advanced_install.adoc
bfallonf: @knobunc @sdodson @danwinship Hmm ok. Not too sure what to do here. Would it be best to make this section appear only in the various OpenShift docs, meaning it wouldn't show up in the Atomic docs? And perhaps a small explanation (perhaps with the admonition about not having mixed firewalls in a cluster) mentioning that "FirewallD is recommended, because of <reason>, even though iptables is the default"?

install_config/install/advanced_install.adoc
danwinship: Actually, "firewalld" is correct. ("FirewallD" may have been considered correct at one point; it shows up in the URL http://fedoraproject.org/wiki/FirewallD for instance, but that page text itself uses "firewalld" everywhere, as does the man page and http://firewalld.org/.)

bfallonf: Ah sure. Thanks @danwinship . Your research game seems stronger than mine.

bfallonf:
OK. After chatting with @knobunc about what the best approach would be, I've made some changes.

Pretty much what I suggested above with putting the whole section in ifdefs so that it doesn't show up in the Atomic docs and adding a note box saying firewalld is recommended.

@sdodson @knobunc @danwinship Any more thoughts on this or shall I move it to be reviewed? Thanks for the thoughts, all.

sdodson:
looks fine to me

bfallonf:
Sure thing. Thanks @sdodson . @knobunc , let me know if there's anything else.

In the meantime, @openshift/team-documentation I'd say this is up for peer review, please.

install_config/install/advanced_install.adoc
```diff
@@ -620,6 +623,41 @@ value; you only need to set this if you want your `git clone` operations to use
...
+[IMPORTANT]
+====
+If you are changing the default firewall, ensure that each host in your cluster
+is using the same firewall type to prevent inconsistencies.
+====
+
+[NOTE]
+====
+While iptables is the default firewall, firewalld is recommended for new
+installs.
```
ahardin-rh: s/installs/installations

install_config/install/advanced_install.adoc
```diff
@@ -620,6 +623,41 @@ value; you only need to set this if you want your `git clone` operations to use
...
+
+{product-title} uses iptables as the default firewall, but you can configure
+your cluster to use firewalld during the install process.
+
+Because iptables is the default firewall, {product-title} is designed to have it
+configured automatically. However, iptables rules can break {product-title} if
+not configured correctly. The advantages of firewalld include allowing multiple
+objects to safely share the firewall rules.
+
+To use firewalld as the firewall for an {product-title} install, add the
```
ahardin-rh: s/install/installation

ahardin-rh:
Just a nit from me. Otherwise, looks good! :fireworks:

bfallonf:
Thanks @ahardin-rh . I'll merge this.

bfallonf:
[rev_history]
|xref:../install_config/install/advanced_install.adoc#install-config-install-advanced-install[Installing a Cluster -> Advanced Installation]
|Added the xref:../install_config/install/advanced_install.adoc#advanced-install-configuring-firewalls[Configuring the Firewall] section.
%

PR #4635, opened 2017-06-22
[WIP] Fixed rendering issue in code blocks callouts

gaurav-nelson:
Also want to discuss:
1. Which format to use for code blocks with callouts?

    This: 
    ```
    ----
    code line one <1>
    code line two <2>
    ----
    <1> Description goes here
    <2> Description goes here
    ```

    vs this:
    ```
    ====
    ----
    code line one <1>
    code line two <2>
    ----
    <1> Description goes here
    <2> Description goes here
    ====
    ```
2. Should we include Source Tags (example: `[source, yaml]`) in guidelines as well?

### DIFF
--- a/contributing_to_docs/doc_guidelines.adoc
+++ b/contributing_to_docs/doc_guidelines.adoc
@@ -435,12 +435,15 @@ Without the line spaces, the content is likely to be not parsed correctly.
 
 Use this format when embedding callouts into the code block:
 
+[subs=-callouts]
+....
 ----
 code example 1 <1>
 code example 2 <2>
 ----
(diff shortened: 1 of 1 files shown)


gaurav-nelson:

adellape:
They should just be `----` at this point per [Documentation Guidelines](https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#code-blocks).

A previous version of the guidelines had a distinction between "examples" and "code blocks", but those are outdated so any lingering `====` blocks like that can be removed.

PR #4652, opened 2017-06-26
Introduce docs for etcd migration

ingvagabund:


admin_guide/etcd_migration.adoc
```diff
@@ -0,0 +1,138 @@
...
+It is highly recommended to perform all the steps automatically.
+If any of the etcd members is started too late, the member will not get accepted by the cluster.
+
+== Deployments
+
+The etcd migration is supported over the following deployments:
+
+* rpm based deployment
+* container based deployment
+
```
ingvagabund: @sdodson do/will we support etcd migration for AH?

sdodson: Yeah, we'll need to. We should be able to run this inside the master container. Lets sync up and see if we can get that implemented.

sdodson:
Reading through this, we make no mention of the playbooks. I'll try to find some time to push some commits to the branch to add that but I may not get time. I think this is a case where we should put the playbook front and center and then elaborate on what we're actually doing after that. Making it clear that the playbook handles all of that for you.

sdodson:
Reading through this, we make no mention of the playbooks. I'll try to find some time to push some commits to the branch to add that but I may not get time. I think this is a case where we should put the playbook front and center and then elaborate on what we're actually doing after that. Making it clear that the playbook handles all of that for you.

sdodson:
We're going to hold documenting this process while we test things more extensively.

adellape:
@ingvagabund @sdodson Anything you want to change in here at this point before I start making edits?

sdodson:
Yeah, can I have until the end of day tomorrow to update it? The biggest thing I want to add is how to recover from a failed migration in the unlikely case you have to do so.

sdodson:
Ok, updated to add recovery procedure and to accurately represent the new process.

adellape:
Thanks, moving to https://github.com/openshift/openshift-docs/pull/4652.

PR #4682, opened 2017-06-30
Quota does not support limits.storage

derekwaynecarr:
this is an error in the doc.

quota only supports requests.storage.

/cc @mburke5678

### DIFF
--- a/admin_guide/quota.adoc
+++ b/admin_guide/quota.adoc
@@ -74,10 +74,6 @@ this value.
 |`*limits.memory*`
 |The sum of memory limits across all pods in a non-terminal state cannot exceed
 this value.
-
-|`*limits.storage*`
-|The sum of storage limits across all persistent volume claims cannot
-exceed this value.
 |===
 
 


mburke5678:
@derekwaynecarr This parameter appears back to 3.2, when the topics was introduced. Want to verify that we are making this change to 3.3, 3.4. and 3.5.

derekwaynecarr:
@mburke5678 -- it was changed in error here: https://github.com/openshift/openshift-docs/pull/3729

any version of doc that says we support limits.storage needs updating.

mburke5678:
Removed 3.3 label. Will be addressed through separate PR and Bugzilla
https://bugzilla.redhat.com/show_bug.cgi?id=1466646

mburke5678:
[rev_history]
|xref:../ admin_guide/quota.adoc#admin-guide-quota[Cluster Administration -> Setting Quotas]
|Removed unsupported limits.storage parameter.
%

PR #4712, opened 2017-07-05
Trello associate my namespace (scheduler.alpha.kubernetes.io/node-selector)

mburke5678:


mburke5678:
Not needed. Created in error.

PR #4730, opened 2017-07-07
Removed links to external routers

vikram-redhat:


adellape:
@vikram-redhat Did this get picked to any downstream branches?

PR #4745, opened 2017-07-10
Use oc rollout latest instead of oc deploy --latest

php-coder:
Because `oc deploy --latest` is deprecated

```console
$ oc deploy dc/ruby-ex --latest
Flag --latest has been deprecated, use 'oc rollout latest' instead
Started deployment #2
Use 'oc logs -f dc/ruby-ex' to track its progress.
```

I've updated docs to use `oc rollout latest` instead.

PTAL @openshift/team-documentation @tnozicka @kargakis 
CC @simo5

### DIFF
--- a/dev_guide/deployments/advanced_deployment_strategies.adoc
+++ b/dev_guide/deployments/advanced_deployment_strategies.adoc
@@ -164,7 +164,7 @@ In the editor, add the line `ab-example: "true"` underneath `spec.selector` and
 . Trigger a re-deployment of the first shard to pick up the new labels:
 +
 ----
-$ oc deploy ab-example-a --latest
+$ oc rollout latest dc/ab-example-a
 ----
 
 . Create a service that uses the common label:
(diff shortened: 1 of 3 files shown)


dev_guide/deployments/advanced_deployment_strategies.adoc
```diff
@@ -164,7 +164,7 @@ In the editor, add the line `ab-example: "true"` underneath `spec.selector` and
 . Trigger a re-deployment of the first shard to pick up the new labels:
 +
 ----
-$ oc deploy ab-example-a --latest
+$ oc rollout latest ab-example-a
```
php-coder: It looks a bit strange and I'm not sure whether it works, because here is we're using `ab-example-a` without `dc/` prefix. Let me know if I need to fix it.

kargakis: Should work both ways but I am in favor of adding dc/

php-coder: Updated.

adellape:
@php-coder LGTM, merging. What version was it deprecated in? Labeling it for 3.6+ for now.

php-coder:
> What version was it deprecated in?

@adellape Looks like it was deprecated in https://github.com/openshift/origin/commit/cc7cbc208157f238555d00089b903f4d99119b4a commit. I see that the commit belongs to v1.4.0 and v1.5.0 tags, hence it was deprecated in 3.4

@kargakis is it correct?

PR #4751, opened 2017-07-11
adding a brief note about DNS and egress network policy

JacobTanenbaum:
The solution to [BZ1458849](https://bugzilla.redhat.com/show_bug.cgi?id=1458849) works when using the default ansible install. Adding a note to the section about managing egressnetworkpolicy that the administrator may need to explicitly allow DNS traffic

### DIFF
--- a/admin_guide/managing_networking.adoc
+++ b/admin_guide/managing_networking.adoc
@@ -274,6 +274,17 @@ different, and the egress network policy may not be enforced as expected. In the
 above example, suppose `www.foo.com` resolved to `10.11.12.13` and has a DNS TTL
 of one minute, but was later changed to `20.21.22.23`. {product-title} will then
 take up to one minute to adapt to these changes.
++
+[NOTE]
+====
+The egress firewall always allows pods access to the external interface of the
+node the pod is on for DNS resolution. If your DNS resolution is not handled by
+something on the local node, then you will need to add egress firewall rules
+allowing access to the DNS server's IP addresses if you are using domain names
+in your pods. The xref:../install_config/install/quick_install.adoc#install-config-install-quick-install[default installer]
+sets up a local dnsmasq, so if you are using that setup you will not need to add extra rules.
+====
+
 
 . Use the JSON file to create an EgressNetworkPolicy object:
 +


JacobTanenbaum:
@knobunc @danwinship Does this note the issue appropriately?

admin_guide/managing_networking.adoc
```diff
@@ -275,6 +275,12 @@ above example, suppose `www.foo.com` resolved to `10.11.12.13` and has a DNS TTL
 of one minute, but was later changed to `20.21.22.23`. {product-title} will then
 take up to one minute to adapt to these changes.
 
+[NOTE]
+====
+If you don't use the default install and you include a deny all rule you may need to explicitly allow traffic to your DNS server.
```
knobunc: I'd like this to be expanded a little further to explain why the default install works (i.e. we should say that we always allow access to the external interface of the host the pod is on).  We should also say that they need to allow access to the DNS servers (not the plural) by IP address (just to avoid stupidity).

knobunc (changes requested):

admin_guide/managing_networking.adoc
danwinship: "default install" should probably be a link

admin_guide/managing_networking.adoc
```diff
@@ -275,6 +275,15 @@ above example, suppose `www.foo.com` resolved to `10.11.12.13` and has a DNS TTL
 of one minute, but was later changed to `20.21.22.23`. {product-title} will then
 take up to one minute to adapt to these changes.
 
+[NOTE]
+====
+The default installer always allows pods access to the external interface of the
```
pecameron: @JacobTanenbaum context is not clear to me. It seems that you really just need access to the DNS server's IP address? Is this the installer setting up a firewall rule to allow the pod to access the external interface? Or is it the DENY rule must be last? What is it that we care about in this discussion?

pecameron (commented):
I got confused... (probably just me)

admin_guide/managing_networking.adoc
knobunc: How about:

The egress firewall always allows pods access to the external interface of the node the pod is on. If your DNS resolution is not handled by something on the local node, the you will need to add egress firewall rules allowing access to the DNS server's IP addresses if you are using domain names in your pods.  The default installer sets up a local dnsmasq, so if you are using that setup you will not need to add extra rules.

Please add a link around "default installer" to the right docs for the installer.

bfallonf:
Thanks @JacobTanenbaum . When you say "the default installer", do you mean the [quick install tool](https://docs.openshift.com/container-platform/3.5/install_config/install/quick_install.html), or is something involving the ansible install ([the advanced install](https://docs.openshift.com/container-platform/3.5/install_config/install/advanced_install.html))? I know the quick install is basically a wrapper for the ansible install, so maybe both?

Also, when you say "you will need a rule allowing access to a DNS server's IP", is there somewhere in the docs that shows how to do this that we can link to? It'd be good to have that info there as well.

JacobTanenbaum:
@knobunc @pecameron How does it look? 

@bfallonf I added a reference to the install tool. As to a specific example, this note is within the section on adding egress firewall rules and I am not sure how I would add a general enough example.

admin_guide/managing_networking.adoc
```diff
@@ -275,6 +275,17 @@ above example, suppose `www.foo.com` resolved to `10.11.12.13` and has a DNS TTL
 of one minute, but was later changed to `20.21.22.23`. {product-title} will then
 take up to one minute to adapt to these changes.
 
+[NOTE]
```
bfallonf: @JacobTanenbaum Because the note box you're adding is to do with the step in the procedure above, it needs a + sign to link it to the paragraphs. (See line 267 for an example)

bfallonf:
Thanks @JacobTanenbaum I see what you mean now. And thanks for the link. I added a comment about linking the note box to the para above, but other than that, I think this is fine if everyone is happy with it. I can merge it once that's done, if you agree. Thanks!

JacobTanenbaum:
@bfallonf added the plus sign and squashed the commits

bfallonf:
Thanks @JacobTanenbaum . I'll merge.

bfallonf:
[rev_history]
|Added an admonition about DNS and egress network policy to the xref:../admin_guide/managing_networking.adoc#admin-guide-limit-pod-access-egress[Using an Egress Firewall to Limit Access to External Resources] section.
%

PR #4767, opened 2017-07-12
fixes for Issue#3490

gaurav-nelson:


gaurav-nelson:
@spadgett Can you please check if this looks alright? 
PS: I got your name from GitHub reviewers suggestion. :eyeglasses:

install_config/web_console_customization.adoc
```diff
@@ -48,6 +48,11 @@ assetConfig:
     - ...
 ----
 
+[NOTE]
+====
+JavaScript code from extension scripts is concatenated into a single file called `extensions.js`. Code is then executed in-line, so variable declarations will affect global state. Wrap extensions in a closure, following the Immediately Invoked Function Expression (IIFE) pattern, to ensure that variable declarations do not leak to the window object.
```
spadgett: Agree that we should encourage the use of IIFE, but this would be true regardless of whether the files are concatenated together. So I might just say something like

```
[NOTE]
====
Consider wrapping extension scripts in an Immediately Invoked Function Expression (IIFE).
This ensures that you don't create global variables that conflict with names used by the
web console or other extensions.
====
```

Then provide an example like

```js
(function() {
  // My extension code here...
}());
```

cc @jawnsy @jwforres

install_config/web_console_customization.adoc
```diff
@@ -48,13 +49,29 @@ assetConfig:
     - ...
 ----
 
+[NOTE]
```
gaurav-nelson: Thanks @spadgett I have updated this based on suggestions.

spadgett (approved):

install_config/web_console_customization.adoc
```diff
@@ -48,13 +49,29 @@ assetConfig:
...
+====
+Wrap extension scripts in an Immediately Invoked Function Expression (IIFE).
```
gaurav-nelson: @bfallonf PTAL

bfallonf:
@gaurav-nelson LGTM :sun_with_face:

gaurav-nelson:
Thanks @bfallonf Merging this now.

PR #4773, opened 2017-07-12
Updated support statement for Starter

ahardin-rh:


ahardin-rh:
Showing failures, but builds just fine locally.

PR #4775, opened 2017-07-12
OCP 3.4 Fix

vikram-redhat:


vikram-redhat:
Merging anyways, even though Travis failed as it is a minor fix.

PR #4780, opened 2017-07-12
Fix missing reference tag

wgordon17:
Signed-off-by: Will Gordon <wgordon@redhat.com>


wgordon17:
@adellape Just had a user point this out to us through our Online support channel. Figured I'd pass along the fix

adellape:
@wgordon17 Thanks! Looks like this was only still an issue in this 3.0 branch. Merging.

Issue #1845, opened 2016-03-31
Change "oadm" with "oc adm"


jorgemoralespou:
Now that there is a single oc command with adm capabilities built in, it would be good to update the documentation to reflect it.

adelton:
The pull request https://github.com/openshift/origin/pull/8313 was closed / retracted because @smarterclayton in https://github.com/openshift/origin/pull/8313#issuecomment-203966425 casted his preference

```
   I would prefer oadm in any admin documentation for now.
```

The problem is, packages like https://github.com/openshift/origin/releases/download/v3.6.0-rc.0/openshift-origin-client-tools-v3.6.0-rc.0-98b3d56-linux-64bit.tar.gz or atomic-openshift-clients-3.6.96-1.git.0.381dd63.el7.x86_64.rpm only seem to ship `openshift-origin-client-tools-v3.6.0-rc.0-98b3d56-linux-64bit/oc` / `/usr/bin/oc`, with `oadm` not available.

Is it time to get rid of `oadm` across the project? Can `oadm` do anything that `oc adm` cannot?

adelton:
Ah, so it looks like in http://lists.openshift.redhat.com/openshift-archives/dev/2017-May/msg00013.html we got the green light.

adelton:
Giving it a try with https://github.com/openshift/origin/pull/15202.

adelton:
And https://github.com/openshift/openshift-docs/pull/4785.

adelton:
Could we have this reopened (GitHub does not allow me to do that)?

The change got into openshift/origin via https://github.com/openshift/origin/pull/15202.

But since this issue is against openshift/openshit-docs, we'd still need https://github.com/openshift/openshift-docs/pull/4785 merged, hence this issue should likely still be open.

adelton:
@mfojtik, could you please reopen this issue?

PR #4785, opened 2017-07-14
Use oc adm instead of oadm which might not exist in various installations.

adelton:
Fixes https://github.com/openshift/openshift-docs/issues/1845.

### DIFF
--- a/getting_started/install_openshift.adoc
+++ b/getting_started/install_openshift.adoc
@@ -194,7 +194,7 @@ basic authentication, user access, and routes.
 {product-title} provides two command line utilities to interact with it.
 
 * `oc`: for normal project and application management
-* `oadm`: for administrative tasks
+* `oc adm` or `oadm`: for administrative tasks
 
 Use `oc --help` and `oadm --help` to view all available options.
 
--- a/admin_guide/building_dependency_trees.adoc
+++ b/admin_guide/building_dependency_trees.adoc
@@ -14,7 +14,7 @@ toc::[]
 {product-title} uses xref:../dev_guide/builds/triggering_builds.adoc#image-change-triggers[image
 change triggers] in a `BuildConfig` to detect when an
 xref:../architecture/core_concepts/builds_and_image_streams.adoc#image-stream-tag[image
-stream tag] has been updated. You can use the `oadm build-chain` command to
+stream tag] has been updated. You can use the `oc adm build-chain` command to
 build a dependency tree that identifies which
 xref:../architecture/core_concepts/containers_and_images.adoc#docker-images[images]
 would be affected by updating an image in a specified
@@ -46,13 +46,13 @@ The following table describes common `build-chain` usage and general syntax:
 
 |Build the dependency tree for the *latest* tag in `<image-stream>`.
 |----
-$ oadm build-chain <image-stream>
+$ oc adm build-chain <image-stream>
 ----
 
 |Build the dependency tree for the *v2* tag in DOT format, and visualize it
 using the DOT utility.
 |----
-$ oadm build-chain <image-stream>:v2 \
+$ oc adm build-chain <image-stream>:v2 \
     -o dot \
     \| dot -T svg -o deps.svg
 ----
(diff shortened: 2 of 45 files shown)


vikram-redhat:
@adelton - what version would this apply from? 3.3 onwards?

cc: @adellape

adelton:
No idea, I worked with master. If backporting to older versions is needed likely needs to be decided by someone else.

vikram-redhat:
Ok - maybe @smarterclayton and/or @liggitt ? Should we port this from 3.3 onwards?

adellape:
@vikram-redhat Downloaded `atomic-openshift-clients` from the 3.3 channels and it does not include `oadm`. So this should be relevant from 3.3 and on. I just tried cherry-picking this to the `enterprise-3.3` to see how it would go, and there are a lot of conflicts. I would probably suggest just doing a `s/oadm/oc adm/` on each of the branches individually instead of trying to cherry-pick this everywhere.

adelton:
> I would probably suggest just doing a s/oadm/oc adm/ on each of the branches individually instead of trying to cherry-pick this everywhere.

Please note that not all `git grep '\boadm\b'` occurences are being replaced by this pull request.

I can likely produce similar pull request(s) for older branch(es) but I'd like to have this pull request commented on in master, to avoid redoing those if something about this proposed commit is not going in after all.

adelton:
Rebased on master to fix typo introduced by 6863599afc48ed54bdd9aabb08491921858234db -> af1759344677b554e1c4d05b3f7c34ab2fbb68d2.

adelton:
Rebased on master -> e0f40a9a2b99446450cd073a9679092f341706b8.

adelton:
Rebased on master -> 0d5936c270798ef17b8598bab1919f8148cb6423.

adelton:
@mfojtik, could you please reopen https://github.com/openshift/openshift-docs/issues/1845 and ideally help with review of this PR?

Note that the change to origin itself was already merged via https://github.com/openshift/origin/pull/15202 so at this point this is about docs catching up.

mfojtik:
@smarterclayton was there any reason we keeping the `oadm` ? i guess enough time has passed already that we should switch to `oc adm`...

/approve

adelton:
> @smarterclayton was there any reason we keeping the oadm ? i guess enough time has passed already that we should switch to oc adm...

Yes, @smarterclayton noted that in http://lists.openshift.redhat.com/openshift-archives/dev/2017-May/msg00013.html.

> /approve

Thanks @mfojtik. What else is needed to get this merged / who can merge this?

adelton:
@smarterclayton, @mfojtik, what else is needed to get this merged / who can merge this?

adelton:
Rebased on master -> ccd646e287c0ddb24642644bdb1ffe4ebfc82a01.

php-coder:
PTAL @openshift/team-documentation

vikram-redhat:
@mburke5678 - PTAL.

getting_started/install_openshift.adoc
```diff
@@ -194,7 +194,7 @@ basic authentication, user access, and routes.
 {product-title} provides two command line utilities to interact with it.
 
 * `oc`: for normal project and application management
-* `oadm`: for administrative tasks
+* `oc adm` or `oadm`: for administrative tasks
```
mburke5678: @adelton Should we remove the _or `oadm`_ here and leave only `oc adm`? 

Also, should the following line be:
_Use `oc --help` and `oc adm --help` to view all available options.

adelton: Can we rule out that there might be an environment where `oadm` is available and `oc` is not?

Please note that this pull requests only attempts to actually show that there is `oc adm` way to do things, and prefer that notation. It did not try to convey message that `oadm` is deprecated as it seemed to become in the mean time.

mburke5678: @adelton From what I understand, the oadm is being deprecated in 3.9. Should we take out all oadm references now,since you have gotten so many to this point?

adelton: In 3.9, is it deprecated or completely gone? If it's deprecated but still available (at least in some environments), making note about that with the few `oadm` occurences that are left might give the change better visibility than merely removing them right away.

adelton:
@mburke5678, I have a question about the `branch/*` labels that you've just added.

Do they mean that this change will be automagically backported to those branches? Or are manualy patches needed?

Also, in https://github.com/openshift/openshift-docs/pull/4785#issuecomment-317457952 @adellape mentioned that the change might be needed in 3.3 as well -- should that be added to the list of labels?

PR #4798, opened 2017-07-17
BZ1455824 Maintaing configmaps when upgrading EFK stack

bmcelvee:
https://bugzilla.redhat.com/show_bug.cgi?id=1455824

### DIFF
--- a/install_config/aggregate_logging.adoc
+++ b/install_config/aggregate_logging.adoc
@@ -422,6 +422,28 @@ for deploying Kibana instances.
 for deploying Curator instances.
 |===
 
+
+
+
+[WARNING]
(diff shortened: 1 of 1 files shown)


xiazhao2015:
I believe the doc PR should go to 3.4 according to https://bugzilla.redhat.com/show_bug.cgi?id=1455824#c1

Also the 3.5 ansible parameters will not be used since we've dropped the upgrade support in ansible since logging 3.5. -- @ewolinetz , please correct me if I was wrong, thanks!

bmcelvee:
Thanks @xiazhao2015! I'll close this PR and open one for 3.4.

PR #4805, opened 2017-07-18
Conditionalize v2_v3 link

vikram-redhat:


vikram-redhat:
Have applied this to dedicated and will tag it for next release for the other branches.

PR #4823, opened 2017-07-19
Fixed the default for the RELOAD_INTERVAL (12s -> 5s)

knobunc:
The docs had the wrong default listed.

This is wrong in version 3.1 and all subsequent versions (But I don't mind if it is only fixed in the current release... it's not a severe problem).

### DIFF
--- a/architecture/topics/router_environment_variables.adoc
+++ b/architecture/topics/router_environment_variables.adoc
@@ -59,7 +59,7 @@ will "linger" around for that period. xref:time-units[(TimeUnits)]
 |`STATS_PORT` |  | Port to expose statistics on (if the router implementation supports it).  If not set, stats are not exposed.
 |`STATS_USERNAME` |  | The user name needed to access router stats (if the router implementation supports it).
 |`TEMPLATE_FILE` | `/var/lib/haproxy/conf/custom/` `haproxy-config-custom.template` | The path to the HAProxy template file (in the container image).
-|`RELOAD_INTERVAL` | 12s | The minimum frequency the router is allowed to reload to accept new changes. xref:time-units[(TimeUnits)]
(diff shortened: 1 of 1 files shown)


knobunc:
@openshift/networking PTAL

knobunc:
I checked that we don't set a default in oadm router.  And the default in cmd/infra/router/template.go is 5, which is turned into "5s".

rajatchopra (approved):
LGTM

ahardin-rh:
@knobunc Looks like this needs a rebase. Also, what versions does this apply to?

knobunc:
@bfallonf @ahardin-rh: Rebased.  I've amended the comment at the top to indicate versions.

ahardin-rh:
@knobunc Thanks so much!

@vikram-redhat This is dependent upon https://github.com/openshift/openshift-docs/pull/5298 so I assume we should hold off merging this 🤔

vikram-redhat:
@ahardin-rh - we can merge it, as the changes brought by #5298 are already in master. We should not CP to 3.6 or do the weekly release till #5298 is merged though.

ahardin-rh:
@vikram-redhat Okay, thanks. Merging! 🎆

vikram-redhat:
@ahardin-rh thanks. Since @knobunc has mentioned that this can be applied to the latest version only, I am removing the old labels. This fix will require a separate PR for older versions if needed.

vikram-redhat:
I have added this to the dedicated-3.6-stage branch via this: https://github.com/openshift/openshift-docs/pull/5348, so adding the branch/dedicated label back on.

vikram-redhat:
and online is now done to 3.6 with this PR: https://github.com/openshift/openshift-docs/pull/5350. Adding online label now.

PR #4829, opened 2017-07-20
Moved and reworded section on auto-scaling metric components 

mburke5678:
A section on HPA in the Scaling Cluster Metrics page appears to be a legacy section from 3.3. I removed:
Horizontal Pod Autoscaling
OpenShift Container Platform version 3.3 does not provide Horizontal Pod Autoscaling (HPA) support for metrics pods and scaling metrics pods.

For reference, see: https://docs.openshift.com/container-platform/3.5/scaling_performance/scaling_cluster_metrics.html#cluster-metrics-horizontal-pod-autoscaling

### DIFF
--- a/scaling_performance/scaling_cluster_metrics.adoc
+++ b/scaling_performance/scaling_cluster_metrics.adoc
@@ -22,6 +22,15 @@ autoscalers] in order to determine when and how to scale.
 
 This topic provides information for scaling the metrics components.
 
+ifdef::openshift-enterprise[]
+[[cluster-metrics-horizontal-pod-autoscaling]]
+[NOTE]
+==== 
+Autoscaling the metrics components, such as Hawkular and Heapster, is not supported by {product-title}.
+====
+endif::[]
+
+
 [[metrics-recommendations-for-OCP-version-35]]
 == Recommendations for {product-title} Version 3.5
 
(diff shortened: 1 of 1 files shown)


mburke5678:

bfallonf:
@mburke5678 Makes sense to me. Labeled for 3.4 and 3.5?

mburke5678:
@DirectXMan12 Per your email, I replaced the section, but moved it to the top as a note. It seemed out of place at the bottom and the title (Horizontal Pod Autoscaling) seems confusing as we have a full topic on HPA. Can you take a quick look at the PR and let me know if this makes sense and I can do the same to 3.4.

DirectXMan12:
That looks better.  I'm still not entirely convinced that it won't cause panic when people read the first few words without reading the rest of the sentence.  Maybe just `Autoscaling the metrics components, such as Hawkular and Heapster, is not supported by {product-title}`, otherwise, people might read `{product-title} does not provide HPA support for the deployments and replication controllers` and get confused (because it looks like it's saying that HPA does not support DCs and RCs).

mburke5678:
@openshift/team-documentation Quick change. A paragraph from 3.3 remained in the docs. I removed the version, reworded per Solly, and moved to top of page. This topic is 3.5 only. Will make same change in 3.4.

DirectXMan12 (approved):
LGTM :+1:

mburke5678:
No rev history needed.

PR #4850, opened 2017-07-26
Fixed spacing per aburden

mburke5678:


mburke5678:

Andrew Burden reported via Xchat last evening that the spacing in this topic was wrong:
https://access.redhat.com/documentation/en-us/openshift_container_platform/3.5/html-single/cluster_administration/#admin-guide-manage-authorization-policy

cluster-admindefault role 
admindefault role 

I *think* the missing space was due to the term before the link being *bolded*, which creates an HTML span in the portal docs, and apparently does not add the space before the link.

ahardin-rh:
LGTM

PR #4860, opened 2017-07-27
changes to arbitrary uid guidance

tchughesiv:
Here are a few reasons for using this method over nss_wrapper:

- Faster image build time
  - Nothing extra has to be installed to the base image
- Smaller image (by 50mb+)
  - nss_wrapper & gettext install w/ dependencies
- Permanent user name recognition in an arbitrary uid container (e.g. openshift)
  - container-wide change vs singular process / expiring
    - nss_wrapper only identifies the user properly for a single command/execution
  - Things like systemd, ansible-playbook, & dbus daemon all work as expected with this arbitrary uid solution. Whereas nss_wrapper wasn’t a sufficient method for these commands to function, as they require addtl user lookups/checks.

### DIFF
--- a/creating_images/guidelines.adoc
+++ b/creating_images/guidelines.adoc
@@ -284,8 +284,8 @@ to allow users in the root group to access them in the built image:
 
 ====
 ----
-RUN chgrp -R 0 /some/directory \
-  && chmod -R g+rwX /some/directory
+RUN chgrp -R 0 /some/directory && \
+    chmod -R g=u /some/directory
 ----
 ====
 
@@ -299,45 +299,33 @@ privileged user.
 Because the user ID of the container is generated dynamically, it will not have
 an associated entry in *_/etc/passwd_*. This can cause problems for applications
 that expect to be able to look up their user ID. One way to address this problem
-is to use link:https://cwrap.org/nss_wrapper.html[nss wrapper] and dynamically
-create a *_passwd_* file with the container's user ID as part of the image's
-start script:
+is to dynamically create a *_passwd_* file entry with the container's user ID as part
+of the image's start script. This is what a Dockerfile might include:
 
 ----
-export USER_ID=$(id -u)
-export GROUP_ID=$(id -g)
-envsubst < ${HOME}/passwd.template > /tmp/passwd
-export LD_PRELOAD=/usr/lib64/libnss_wrapper.so
-export NSS_WRAPPER_PASSWD=/tmp/passwd
-export NSS_WRAPPER_GROUP=/etc/group
+RUN chmod g=u /etc/passwd
+ENTRYPOINT [ "uid_entrypoint" ]
+USER 1001
 ----
 
-Where *_passwd.template_* contains:
+Where *_uid_entrypoint_* contains:
 
 ----
-root:x:0:0:root:/root:/bin/bash
-bin:x:1:1:bin:/bin:/sbin/nologin
-daemon:x:2:2:daemon:/sbin:/sbin/nologin
-adm:x:3:4:adm:/var/adm:/sbin/nologin
-lp:x:4:7:lp:/var/spool/lpd:/sbin/nologin
-sync:x:5:0:sync:/sbin:/bin/sync
-shutdown:x:6:0:shutdown:/sbin:/sbin/shutdown
-halt:x:7:0:halt:/sbin:/sbin/halt
-mail:x:8:12:mail:/var/spool/mail:/sbin/nologin
-operator:x:11:0:operator:/root:/sbin/nologin
-games:x:12:100:games:/usr/games:/sbin/nologin
-ftp:x:14:50:FTP User:/var/ftp:/sbin/nologin
-nobody:x:99:99:Nobody:/:/sbin/nologin
-postgres:x:${USER_ID}:${GROUP_ID}:PostgreSQL Server:${HOME}:/bin/bash
+if ! whoami &> /dev/null; then
+  if [ -w /etc/passwd ]; then
+    echo "${USER_NAME:-default}:x:$(id -u):0:${USER_NAME:-default} user:${HOME}:/sbin/nologin" >> /etc/passwd
+  fi
+fi
 ----
 
-Additionally, you must install the *nss_wrapper* and *gettext* packages in your
-image for this to work. The latter provides the `envsubst` command. For
-example you can add this line to your *_Dockerfile_* for yum-based  images:
-
-----
-RUN yum -y install nss_wrapper gettext
-----
+For a complete example of this, see 
+ifdef::openshift-origin[]
+link:https://github.com/RHsyseng/container-rhel-examples/blob/master/starter-arbitrary-uid/Dockerfile.centos7[this Dockerfile]
+endif::[]
+ifdef::openshift-enterprise[]
+link:https://github.com/RHsyseng/container-rhel-examples/blob/master/starter-arbitrary-uid/Dockerfile[this Dockerfile]
+endif::[]
+.
 
 Lastly, the final *USER* declaration in the `Dockerfile` should specify the user
 ID (numeric value) and not the user name. This allows {product-title} to


tchughesiv:
I don't have permissions to apply labels, but i expect these changes could be applied to all openshift versions that default to a restricted scc.

aweiteka:
I like this simpler approach. @bparees could you take a look at this?

creating_images/guidelines.adoc
```diff
@@ -299,44 +299,27 @@ privileged user.
...
-export LD_PRELOAD=/usr/lib64/libnss_wrapper.so
-export NSS_WRAPPER_PASSWD=/tmp/passwd
-export NSS_WRAPPER_GROUP=/etc/group
+ENV USER_NAME=nobody \
+    USER_UID=99
+RUN chown -R ${USER_UID}:0 /some/directory \
+    && chmod -R g=u /some/directory /etc/passwd
+RUN sed "s@${USER_NAME}:x:${USER_UID}:@${USER_NAME}:x:\${USER_ID}:@g" /etc/passwd > /etc/passwd.template
+ENTRYPOINT [ "uid_entrypoint" ]
+USER 99
```
aweiteka: Maybe a pointer to an upstream example? https://github.com/RHsyseng/container-rhel-examples/blob/master/starter-arbitrary-uid/Dockerfile.centos7

aweiteka: Actually, weaving the example into the discussion might be the best thing, "Here's a complete example of this." Point to centos7 for origin, rhel for ocp using ifdef, e.g.

```
ifdef::openshift-origin[]
link:centosdockerfile
endif::[]

ifdef::openshift-enterprise[]
link:rheldockerfile
endif::[]

```

creating_images/guidelines.adoc
```diff
@@ -299,44 +299,27 @@ privileged user.
 Because the user ID of the container is generated dynamically, it will not have
 an associated entry in *_/etc/passwd_*. This can cause problems for applications
 that expect to be able to look up their user ID. One way to address this problem
-is to use link:https://cwrap.org/nss_wrapper.html[nss wrapper] and dynamically
-create a *_passwd_* file with the container's user ID as part of the image's
-start script:
+is to dynamically create a *_passwd_* file with the container's user ID as part
+of the image's `ENTRYPOINT` script. This what a Dockerfile might include:
```
codificat: s/This what/This is what/

creating_images/guidelines.adoc
```diff
@@ -299,44 +299,27 @@ privileged user.
...
-export GROUP_ID=$(id -g)
-envsubst < ${HOME}/passwd.template > /tmp/passwd
-export LD_PRELOAD=/usr/lib64/libnss_wrapper.so
-export NSS_WRAPPER_PASSWD=/tmp/passwd
-export NSS_WRAPPER_GROUP=/etc/group
+ENV USER_NAME=nobody \
+    USER_UID=99
+RUN chown -R ${USER_UID}:0 /some/directory \
+    && chmod -R g=u /some/directory /etc/passwd
+RUN sed "s@${USER_NAME}:x:${USER_UID}:@${USER_NAME}:x:\${USER_ID}:@g" /etc/passwd > /etc/passwd.template
```
bparees: removing the nobody user seems like a mistake.  Why not just concatenate the existing /etc/passwd file?

creating_images/guidelines.adoc
```diff
@@ -299,44 +299,27 @@ privileged user.
...
-
-Additionally, you must install the *nss_wrapper* and *gettext* packages in your
-image for this to work. The latter provides the `envsubst` command. For
-example you can add this line to your *_Dockerfile_* for yum-based  images:
-
-----
-RUN yum -y install nss_wrapper gettext
+if [ -w /etc/passwd ]; then
+  if ! whoami &> /dev/null; then
+    sed "s@${USER_NAME}:x:\${USER_ID}:@${USER_NAME}:x:$(id -u):@g" /etc/passwd.template > /etc/passwd
```
bparees: why use a template?  just generate the new line and concatenate it to the existing /etc/passwd file:
https://github.com/openshift/jenkins/blob/master/2/contrib/jenkins/jenkins-common.sh#L13-L22

codificat (commented):
Just a minor text correction

creating_images/guidelines.adoc
```diff
@@ -299,44 +299,27 @@ privileged user.
...
-export USER_ID=$(id -u)
-export GROUP_ID=$(id -g)
-envsubst < ${HOME}/passwd.template > /tmp/passwd
-export LD_PRELOAD=/usr/lib64/libnss_wrapper.so
-export NSS_WRAPPER_PASSWD=/tmp/passwd
-export NSS_WRAPPER_GROUP=/etc/group
+ENV USER_NAME=nobody \
+    USER_UID=99
+RUN chown -R ${USER_UID}:0 /some/directory \
+    && chmod -R g=u /some/directory /etc/passwd
```
bparees: this chown isn't really helpful. the group change part is, but making uid 99 own everything isn't.

creating_images/guidelines.adoc
bparees: rather than setting the image up to run w/ a default uid that actually exists in /etc/passwd, it's better to set it up to run as a random uid (we typically use 1001) which further proves during testing that things "just work".

bparees (changes requested):

bparees:
no argument w/ the motivation behind this change (thanks @tchughesiv, it was overdue!) but definitely some suggestions on how to make it cleaner.

creating_images/guidelines.adoc
tchughesiv: based on my testing, the chown is necessary so that the image is portable enough that it can access necessary files when run w/ the intended user.  e.g. `docker run` or w/ anyuid scc.

creating_images/guidelines.adoc
tchughesiv: sure... thought about doing this instead, but figured it better to keep the complexity down.  im good w/ adding a line to create a user though if that's preferable.

creating_images/guidelines.adoc
bparees: that may be when you run as user 99 since user 99 is a member of a group, but any user that's not a member of a group will be a member of the root group, which is the whole design here beyond making the files owned by the root group and read/write/executable by it.

So if you also fix your image to not run as 99 by default (but instead run as 1001 or some other user not represented in /etc/group), you will not need to do this.

creating_images/guidelines.adoc
tchughesiv: @bparees  sure... thought about doing that instead, but figured it better to keep the complexity down. im good w/ adding a line to create a user though if that's preferable.

creating_images/guidelines.adoc
tchughesiv: hmm, like add a duplicate line for that user in the passwd file w/ the appropriate uid? or you mean try to do a direct replacement w/o template?  If the latter, i believe that was one of the first things i ever tried w/ this method and there were issues.  I'll definitely give it another go though just to be sure.

bparees: if you look at what i linked, in our jenkins image we just append a new line to the /etc/passwd file for the current container uid.  We skip doing this if and only if the container is being run as the jenkins user (997), but I think you can leave that check out for purposes of this example.  In fact that check isn't even correct since the jenkins user is 998 (the group is 997).

tchughesiv: good point, and we check if the uid is valid first w/ a whoami anyway.

tchughesiv: nice... would def. further simplify things.

creating_images/guidelines.adoc
tchughesiv: def. see your point, but the chown method allows for all use-cases to be successful... regardless of a user's group situation... and in all runtime situations.(restricted/anyuid/docker/etc)
1. where a user's primary group is 0
2. where a user has a non-root primary group

Question is, do we want to force a user belong to the root group for every type of container runtime config? With just a `chgrp`, services defaulting an existing user would then have to add a `usermod` in order for their image to function outside of an arb-uid env right?

bparees: I would assume those files would be untouched, so they'd still be owned+accessible by the service user.  Hence my original point about not changing the user-ownership of the files.

There are basically two cases to cover:
1) you're running the process as the expected user:  things will work because presumably those files were installed and setup to be accessed by that user

2) you're running things as a random user (member of root group): things will work because you've chgrp'd all the files into the root group and given the root group the necessary permissions on the files.

The problematic case would be:

3) the service account is a member of group foo.  The files were previously owned by user foo and group bar, but now they are owned by root. the service can no longer access the files because the service account is not foo and is not a member of the root group.

however again, what you're doing won't solve that either.  only granting "other/world" permissions would.

tchughesiv: fair points... let me run through some tests and come back

tchughesiv:
@bparees @aweiteka @codificat tks for the feedback... requested changes made.

creating_images/guidelines.adoc
```diff
@@ -299,45 +299,35 @@ privileged user.
...
-halt:x:7:0:halt:/sbin:/sbin/halt
-mail:x:8:12:mail:/var/spool/mail:/sbin/nologin
-operator:x:11:0:operator:/root:/sbin/nologin
-games:x:12:100:games:/usr/games:/sbin/nologin
-ftp:x:14:50:FTP User:/var/ftp:/sbin/nologin
-nobody:x:99:99:Nobody:/:/sbin/nologin
-postgres:x:${USER_ID}:${GROUP_ID}:PostgreSQL Server:${HOME}:/bin/bash
+if [ -w /etc/passwd ]; then
+  if ! whoami &> /dev/null; then
+    echo "${USER_NAME}:x:$(id -u):0:${USER_NAME} user:${HOME}:/sbin/nologin" >> /etc/passwd
```
tchughesiv: hmm, thinking about this more @bparees ... do we even need a USER_NAME env var in Dockerfile?  Do we just decide for the user what the user name will be or do we leave in the logic allowing them to specify?  I kind of like the idea of letting someone decide what the user name will be.

tchughesiv: not that it matters a ton... just preference.

bparees: if you're going to let them choose at all, i'd let them choose by setting the env var and having a default.  So no, you don't need it in the dockerfile.

creating_images/guidelines.adoc
```diff
@@ -299,45 +299,34 @@ privileged user.
...
 
 ----
-export USER_ID=$(id -u)
-export GROUP_ID=$(id -g)
-envsubst < ${HOME}/passwd.template > /tmp/passwd
-export LD_PRELOAD=/usr/lib64/libnss_wrapper.so
-export NSS_WRAPPER_PASSWD=/tmp/passwd
-export NSS_WRAPPER_GROUP=/etc/group
+RUN chgrp -R 0 /some/directory \
+    && chmod -R g=u /some/directory /etc/passwd
```
bparees: if you put the && on the previous line it aligns more nicely.

also the -R isn't really needed for a single file chmod.

bparees (changes requested):
two last nits and lgtm.

creating_images/guidelines.adoc
tchughesiv: Agree w/ '&&' on prior line... had just kept how it was in prior docs, will change.

In regards to the -R, that is required for '/some/directory'. Begs the question though, do we repeat the /some/directory `chgrp` and `chmod` since its also shown above? or do we limit this docs section to the single file `chmod g=u /etc/passwd`?  Folks can then click on one of the links below for a full working Dockerfile as a reference for piecing it all together.

bparees: > In regards to the -R, that is required for '/some/directory'. 

hmm, right, forgot you're doing both in that line.  

> Begs the question though, do we repeat the /some/directory chgrp and chmod since its also shown above? or do we limit this docs section to the single file chmod g=u /etc/passwd?

good point, i would vote for not repeating it.  this part of the docs is just about how to get a random uid into /etc/passwd so "whoami" works, not about ensuring that random uid has permission to access particular files/folders (which is what the previous bit covers)

bparees:
lgtm.  @openshift/team-documentation ptal/merge.

bparees:
@openshift/team-documentation bump

bparees (approved):

creating_images/guidelines.adoc
```diff
@@ -299,45 +299,33 @@ privileged user.
...
-Additionally, you must install the *nss_wrapper* and *gettext* packages in your
-image for this to work. The latter provides the `envsubst` command. For
-example you can add this line to your *_Dockerfile_* for yum-based  images:
-
-----
-RUN yum -y install nss_wrapper gettext
-----
+For a complete example of this, see 
+ifdef::openshift-origin[]
+https://github.com/RHsyseng/container-rhel-examples/blob/master/starter-arbitrary-uid/Dockerfile.centos7[this Dockerfile]
```
ahardin-rh: link:https://github.com/RHsyseng/container-rhel-examples/blob/master/starter-arbitrary-uid/Dockerfile.centos7[this Dockerfile]

creating_images/guidelines.adoc
```diff
@@ -299,45 +299,33 @@ privileged user.
...
+endif::[]
+ifdef::openshift-enterprise[]
+https://github.com/RHsyseng/container-rhel-examples/blob/master/starter-arbitrary-uid/Dockerfile[this Dockerfile]
```
ahardin-rh: link:https://github.com/RHsyseng/container-rhel-examples/blob/master/starter-arbitrary-uid/Dockerfile[this Dockerfile]

ahardin-rh:
@tchughesiv For our publishing process, we need to have the changes squashed into one commit. Thanks!

tchughesiv:
@ahardin-rh done

ahardin-rh:
Thank you!

PR #4862, opened 2017-07-28
Bug 1468049 added MTU information to the performance guide

bfallonf:
For: https://bugzilla.redhat.com/show_bug.cgi?id=1468049

cc @openshift/networking @dcbw 

Will send an email for verification.

### DIFF
--- a/scaling_performance/network_optimization.adoc
+++ b/scaling_performance/network_optimization.adoc
@@ -170,6 +170,41 @@ network infrastructure.
 VXLAN-offload does not reduce latency. However, CPU utilization is reduced even
 in latency tests.
 
+[[scaling-performance-optimizing-mtu]]
+=== Optimizing the MTU for Your Network
+
+There are two MTUs (maximum transmission units) that matter: the network
+interface card (NIC) MTU, and the SDN overlay's MTU.
+
+The NIC MTU must be less than or equal to the maximum supported value of the NIC
+of your network. If you are optimizing for throughput, pick the largest possible
+value. If you are optimizing for lowest latency, pick a lower value.
+
+The SDN overlay's MTU must be less than the NIC MTU by 50 bytes at a
+minimum. This accounts for the SDN overlay header. So, on a normal ethernet
+network, set this to 1450. On a jumbo frame ethernet network, set this to 8950.
+
+[NOTE]
+====
+This 50 byte overlay header is relevant to the OpenShift SDN. Other SDN
+solutions might require the value to be more or less.
+====
+
+To configure the MTU, edit the node configuration file at *_/etc/origin/node/node-config.yaml_*, and edit the following:
+
+[source,yaml]
+----
+networkConfig:
+  mtu: 1450 <1>
+  networkPluginName: "redhat/openshift-ovs-subnet" <2>
+----
+<1> Maximum transmission unit (MTU) for the pod overlay network.
+<2> Set to `redhat/openshift-ovs-subnet` for the *ovs-subnet* plug-in,
+`redhat/openshift-ovs-multitenant` for the *ovs-multitenant* plug-in, or
+`redhat/openshift-ovs-networkpolicy` for the *ovs-networkpolicy* plug-in. This
+can also be set to any other CNI-compatible plug-in as well.
+
+
 [[scaling-performance-network-subnetting]]
 == Configuring Network Subnets
 


scaling_performance/network_optimization.adoc
```diff
@@ -170,6 +170,31 @@ network infrastructure.
...
+[[scaling-performance-optimizing-mtu]]
+=== Optimizing the MTU for Your Network
+
+There are two MTUs (maximum transmission units) that matter: the network interface card (NIC) MTU, and the SDN overlay's MTU.
+
+The NIC MTU can be whatever you think is optimal for your network.
+
+However, the SDN overlay's MTU must be less than the NIC MTU by 50 bytes at a
+minimum. On a normal ethernet network, this should be set to around 1450. On a
+jumbo frame ethernet network, this should be around 8950.
```
jeremyeder: I would remove the words "around" from this sentence.  1450 and 8950 are exactly what they should be.

scaling_performance/network_optimization.adoc
```diff
@@ -170,6 +170,31 @@ network infrastructure.
 VXLAN-offload does not reduce latency. However, CPU utilization is reduced even
 in latency tests.
 
+[[scaling-performance-optimizing-mtu]]
+=== Optimizing the MTU for Your Network
+
+There are two MTUs (maximum transmission units) that matter: the network interface card (NIC) MTU, and the SDN overlay's MTU.
+
+The NIC MTU can be whatever you think is optimal for your network.
```
jeremyeder: Can you add some more color here...

* Jumbo frames are 6x the size of typical ethernet frames.
* Jumbo frames reduce the number of interrupts by the same factor (6x), and thus will reduce CPU utilization on both the send and receive sides of a connection.
* This larger frame size will increase network latency due to the properties of the Nagle algorithm  https://en.wikipedia.org/wiki/Nagle%27s_algorithm (enabled by default on Linux), that require the network's minimum segment size to be queued before being dispatched.
* Nagle's algorithm can be disabled by opening a tcp socket with TCP_NODELAY socket option set.

/cc @danwinship @dcbw y'all comfortable with this additional info?

scaling_performance/network_optimization.adoc
```diff
@@ -170,6 +170,31 @@ network infrastructure.
...
+
+However, the SDN overlay's MTU must be less than the NIC MTU by 50 bytes at a
+minimum. On a normal ethernet network, this should be set to around 1450. On a
+jumbo frame ethernet network, this should be around 8950.
+
+To configure the MTU, edit the node configuration file at *_/etc/origin/node/node-config.yaml_* and edit the following:
```
jeremyeder: I would assume that's not strictly enough -- don't we have to restart the kubelet as well?  What about any running pods, don't they have to be restarted? 

I am worried we actually need some helper plays in openshift-ansible to make toggling the MTU safe from a supportability/UX standpoint.

jeremyeder (changes requested):

scaling_performance/network_optimization.adoc
knobunc: We should also document how we choose the default (we take the default interface's MTU and subtract 50 from it).  That is adequate in most cases.

scaling_performance/network_optimization.adoc
knobunc: "The NIC MTU can be whatever you think is optimal for your network."

You can't just pick a number.  It must be lss than or equal to the maximum that all of your NICs on the network support.  But you can pick a lower number for the reasons Jeremy outlined.  Optimizing for throughput?  Pick the largest possible. Optimizing for lowest latency?  Pick something lower.

scaling_performance/network_optimization.adoc
knobunc: And this only applies to the OpenShift SDN stuff... other SDN solutions may have different requirements.

And can we say "to accomodate the SDN overlay header" somewhere to explain where the 50 bytes go.

knobunc (changes requested):
Great start.  Thanks

scaling_performance/network_optimization.adoc
dcbw: These look like the words from my email :)  Which is fine, I should have been more precise.  In general, setting the MTU for your overall network is pretty well covered by other resources, and what ben/jeremy have said.

scaling_performance/network_optimization.adoc
dcbw: Yeah, it's best to set the MTU at cluster standup time rather than adjust it later; you'd need to take the cluster down and then back up to ensure that you don't have glitches related to fragmentation and the MTU.

@knobunc your comment only applies to OpenShift SDN I think, right?

scaling_performance/network_optimization.adoc
```diff
@@ -170,6 +170,31 @@ network infrastructure.
...
+[source,yaml]
+----
+networkConfig:
+  mtu: 1450 <1>
+  networkPluginName: "redhat/openshift-ovs-subnet" <2>
+----
+<1> Maximum transmission unit (MTU) for the pod overlay network
+<2> Set to *redhat/openshift-ovs-subnet* for the *ovs-subnet* plug-in,
+*redhat/openshift-ovs-multitenant* for the *ovs-multitenant* plug-in, or
+*redhat/openshift-ovs-networkpolicy* for the *ovs-networkpolicy* plug-in.
```
dcbw: Or the customer could set the plugin to any other CNI-compatible plugin too...

scaling_performance/network_optimization.adoc
bfallonf: @jeremyeder @knobunc @dcbw I'm not too sure what the jumbo frame info is trying to say. I consulted with a networking guy in the office and he said that it's perhaps not too relevant. Is there any of the info that's specific to OpenShift?

I put the other info in, though.

scaling_performance/network_optimization.adoc
bfallonf: @jeremyeder @knobunc @dcbw Would this be more at home in the [Recommended Install Practices section](https://docs.openshift.com/container-platform/3.5/scaling_performance/install_practices.html) then?

I've put in the other info.

bfallonf:
@jeremyeder has given the thumbs up over email.

scaling_performance/network_optimization.adoc
```diff
@@ -170,6 +170,39 @@ network infrastructure.
...
 in latency tests.
 
+[[scaling-performance-optimizing-mtu]]
+=== Optimizing the MTU for Your Network
+
+There are two MTUs (maximum transmission units) that matter: the network
+interface card (NIC) MTU, and the SDN overlay's MTU.
+
+The NIC MTU must be less than or equal to the maximum supported value of the NIC of your network. If you are optimizing for throughput, pick the largest
+possible value. If you are optimizing for lowest latency, pick a lower value.
```
ahardin-rh: Strange line break? 🤷 If it's nothing, you can just disregard.

scaling_performance/network_optimization.adoc
```diff
@@ -170,6 +170,39 @@ network infrastructure.
...
+
+To configure the MTU, edit the node configuration file at *_/etc/origin/node/node-config.yaml_* and edit the following:
+
+[source,yaml]
+----
+networkConfig:
+  mtu: 1450 <1>
+  networkPluginName: "redhat/openshift-ovs-subnet" <2>
+----
+<1> Maximum transmission unit (MTU) for the pod overlay network
```
ahardin-rh: network.

scaling_performance/network_optimization.adoc
```diff
@@ -170,6 +170,39 @@ network infrastructure.
...
+<2> Set to *redhat/openshift-ovs-subnet* for the *ovs-subnet* plug-in,
+*redhat/openshift-ovs-multitenant* for the *ovs-multitenant* plug-in, or
```
ahardin-rh: `redhat/openshift-ovs-subnet`
    `redhat/openshift-ovs-multitenant`
    `redhat/openshift-ovs-networkpolicy`

scaling_performance/network_optimization.adoc
```diff
@@ -170,6 +170,39 @@ network infrastructure.
...
+*redhat/openshift-ovs-networkpolicy* for the *ovs-networkpolicy* plug-in. This can also be set to any other CNI-compatible plugin as well.
+
```
ahardin-rh: plug-in

scaling_performance/network_optimization.adoc
bfallonf: Hmm just rendered weird. Fixed!

bfallonf:
Thanks @ahardin-rh . All suggestions take in. I'll merge.

bfallonf:
[rev_history]
|xref:../scaling_performance/network_optimization.adoc#scaling-performance-routing-network-optimization[Routing and Network Optimization]
|Added the xref:../scaling_performance/network_optimization.adoc#scaling-performance-optimizing-mtu[Optimizing the MTU for Your Network] section.
%

PR #4857, opened 2017-07-27
Document the web console application launcher

spadgett:

PR #4907, opened 2017-08-01
Follow-up edits to PR#4857

ahardin-rh:
Follow-up to https://github.com/openshift/openshift-docs/pull/4857


ahardin-rh:
Rev history in the original PR

PR #4931, opened 2017-08-03
[WIP]Trello-478 Document Elasticsearch Readiness Probe

bmcelvee:
https://trello.com/c/Ui6lk9Dy/478-document-elasticsearch-kibana-should-have-liveness-readiness-checks

### DIFF
--- a/install_config/aggregate_logging.adoc
+++ b/install_config/aggregate_logging.adoc
@@ -728,6 +728,21 @@ Please note that changes to the configmap might not appear until after redeployi
 the pods. Persisting these changes across deployments can be accomplished by setting
 `openshift_logging_es_allows_cluster_reader` to `true` in the inventory file.
 
+[[elasticsearch-readiness]]
+*Elasticsearch Readiness Probe*
+
+The Elasicsearch readiness probe lets you know when a container is ready to
+begin accepting traffic. Once all of the containers are ready, a pod is then
+considered ready.
+
+To start the Elasicsearch readiness probe, run:
+
+----
+$ /usr/share/elasticsearch/probe/readiness.sh
+----
+
+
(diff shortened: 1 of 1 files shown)


bmcelvee:
@richm Would you mind reviewing this one as well? Thanks!

richm:
It seems odd to document a readiness probe, other than to just say that there is one.  The user should never have to run this manually.  @jcantrill @pweil- do we normally do this?

richm:
In addition, I thought we dropped the Elasticsearch readiness probe for 3.6?  @wozniakjan ?

jcantrill:
I would tend to agree that there should be no need to document it other then it exists.  What useful info can become of this documentation?  Do we believe its necessary to describe what readiness is?  Probably not..

vikram-redhat:
@jcantrill - thanks Jeff and Rich. Should we put something in release notes or just here in the install and config guide?

@bmcelvee @ahardin-rh - FYI in case we need to move this to release notes only.

wozniakjan:
I was under the impression that readiness probe will be disabled for 3.6, have we changed that? If so, are we moving forward with [multi-level ES topology](https://github.com/openshift/openshift-ansible/pull/5001) or [discovery by label](https://github.com/fabric8io/elasticsearch-cloud-kubernetes/pull/98)?

jcantrill:
I would be in favor of not adding any documentation for the readiness probe.  Its an implementation detail and @wozniakjan is correct that it was removed for 3.6 because of the issues it was causing

bmcelvee:
Thanks all! @wozniakjan, your PR was pointed out to me by QE. I'll go ahead and close this PR and move my card accordingly.

PR #4935, opened 2017-08-04
Add replica sets to objects capable of working with IS

soltysh:
@ahardin-rh a followup to your followup ;)
Originally I wanted to mention both replication controllers and replica sets, but I haven't fixed it after copy&paste. You've removed the duplication but we still need that info about replica set. This PR adds it.
Unfortunately, there's no reasonable link for RS, so just clean info must be sufficient.


ahardin-rh:
@soltysh Thank you!

PR #4948, opened 2017-08-08
Updates Backup and Restore (Cluster Backup), adds Registry Certificate Backup topic.

jtudelag:
Adress this BZ: [Docker certs are not mentioned on the backup/restore section](https://bugzilla.redhat.com/show_bug.cgi?id=1464494)

### DIFF
--- a/admin_guide/backup_restore.adoc
+++ b/admin_guide/backup_restore.adoc
@@ -154,8 +154,8 @@ inside the container.
 +
 [NOTE]
 ====
-When working with one or more xref:../install_config/registry/securing_and_exposing_registry#exposing-the-registry[external secured registry], 
-*any host* required to pull or push images, hence to run pods, must trust registry certificates.
+When working with one or more xref:../install_config/registry/securing_and_exposing_registry.adoc#exposing-the-registry[external secured registry], 
+*_any host_* required to pull or push images, in order to run pods, must trust registry certificates.
 ====
 
 [[cluster-restore-single-member-etcd-clusters]]
--- a/admin_guide/backup_restore.adoc
+++ b/admin_guide/backup_restore.adoc
@@ -93,12 +93,18 @@ following sections), which depends on how *etcd* is deployed.
 [[cluster-backup]]
 == Cluster Backup
 
+[[master-backup]]
+=== Master Backup
+
 . Save all the certificates and keys, on each master:
 +
 ----
 # cd /etc/origin/master
 # tar cf /tmp/certs-and-keys-$(hostname).tar *.key *.crt
 ----
+
+[[etcd-backup]]
+=== Etcd Backup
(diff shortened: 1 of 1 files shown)


admin_guide/backup_restore.adoc
```diff
@@ -133,8 +139,24 @@ inside the container.
...
+. Save all the registry certificates, on every master and node host.
++
 ----
+# cd /etc/docker/certs.d/
+# tar cf /tmp/docker-registry-certs-$(hostname).tar * 
+----
++
+[NOTE]
+====
+When working with one or more xref:../install_config/registry/securing_and_exposing_registry#exposing-the-registry[external secured registry],
```
ahardin-rh: xref:../install_config/registry/securing_and_exposing_registry.adoc#exposing-the-registry[external registry]

admin_guide/backup_restore.adoc
```diff
@@ -133,8 +139,24 @@ inside the container.
...
++
 ----
+# cd /etc/docker/certs.d/
+# tar cf /tmp/docker-registry-certs-$(hostname).tar * 
+----
++
+[NOTE]
+====
+When working with one or more xref:../install_config/registry/securing_and_exposing_registry#exposing-the-registry[external secured registry], 
+*any host* required to pull or push images, hence to run pods, must trust registry certificates.
```
ahardin-rh: _any host_

ahardin-rh: s/, hence to run pods,/(in order to run pods)

admin_guide/backup_restore.adoc
jtudelag: Hi @ahardin-rh, Thanks for your correction. 

But in this case, this only makes sense when the external registry is secured (it serves traffic via TLS) https://docs.openshift.com/container-platform/3.3/install_config/registry/securing_and_exposing_registry.html#securing-the-registry.
 
That's why certs need to be backed up. So I would keep the [external secured registry] instead of [external registry].

ahardin-rh: @jtudelag Oops! Yes,  you are correct. We just need to make sure that the path includes .adoc before the #exposing-the-registry.

ahardin-rh:
@jtudelag Are you able to make the edits to move this along? I can also continue the work in another PR if you prefer. Thanks!

jtudelag:
@ahardin-rh sorry, I completely forgot about this... It's updated now ;). Thanks!

ahardin-rh:
@jtudelag I'm so sorry for the delay! I continued the work here in https://github.com/openshift/openshift-docs/pull/5582 (minor edit and squashing the commits) so I will close this PR in favor of that one. I'll get it merged right away. Thanks!

PR #4783, opened 2017-07-13
Use newer label format for AWS cluster id

sdodson:
This applies to 3.6 and newer.

PR #4961, opened 2017-08-09
Use newer label format for AWS cluster id

ahardin-rh:
Rebase of https://github.com/openshift/openshift-docs/pull/4783


andrewklau:
Is this required for an upgrades to 3.6? Also it's not clear what the `xxxx` and `clusterid` represent. Should they be both be eg. `staging` and `staging` or do they serve better purposes?

PR #4963, opened 2017-08-09
Fix unreasonable reserved resource example values

nekop:
The kube-reserved system-reserved exmple specifies extremely high memory 30G and it's inappropriate as an example. Make it reasonable.


tmorriso-rh:
TW note: Applies to 3.2 thru 3.9

vikram-redhat:
Removed Next Release Milestone as work is continuing in #7268.

PR #4980, opened 2017-08-09
Add `rpm` command to reall verify Docker version

stevekuznetsov:
If we want the user to be verifying that the correct version of the
Docker RPM was installed in the preceeding step, we should instruct them
to use `rpm -V` as well as visual confirmation for a more fool-proof
process.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>


adellape:
:+1:

PR #4984, opened 2017-08-09
BUG 1470567 default router and registry created when selector does NOT exist

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1470567

### DIFF
--- a/install_config/install/advanced_install.adoc
+++ b/install_config/install/advanced_install.adoc
@@ -661,8 +661,9 @@ If you are using an image registry other than the default at
 *_/etc/ansible/hosts_* file.
 
 ----
-oreg_url=example.com/openshift3/ose-${component}:${version}
+oreg_url={registry}/openshift3/ose-${component}:${version}
 openshift_examples_modify_imagestreams=true
+openshift_docker_additional_registries={registry}
 ----
 
 .Registry Variables
@@ -675,7 +676,17 @@ openshift_examples_modify_imagestreams=true
 
 |`*openshift_examples_modify_imagestreams*`
 |Set to `true` if pointing to a registry other than the default. Modifies the image stream location to the value of `*oreg_url*`.
+
+|`*openshift_docker_additional_registries*`
+|Specify the additional registry or registries.
 |===
+
+For example:
+----
+oreg_url=example.com/openshift3/ose-${component}:${version}
+openshift_examples_modify_imagestreams=true
+openshift_docker_additional_registries=example.com
+----
 
 [[advanced-install-glusterfs-persistent-storage]]
 === Configuring GlusterFS Persistent Storage
@@ -1081,9 +1092,8 @@ pods. They are set to `region=infra` by default:
 # openshift_registry_selector='region=infra'
 ----
 
-The default router and registry will be automatically deployed during
-installation if nodes exist in the `[nodes]` section that match the selector
-settings. For example:
+The registry and router are only able to run on node hosts with the `region=infra` label. 
+Ensure that at least one node host in your {product-title} environment has the `region=infra` label. For example:
 
 ----
 [nodes]
@@ -1092,14 +1102,25 @@ infra-node1.example.com openshift_node_labels="{'region': 'infra','zone': 'defau
 
 [IMPORTANT]
 ====
-The registry and router are only able to run on node hosts with the
-`region=infra` label. Ensure that at least one node host in your {product-title}
-environment has the `region=infra` label.
+If there is not a node in the [nodes] section that matches the selector settings,
+the default router and registry will be deployed as failed with `Pending` status.
 ====
 
 It is recommended for production environments that you maintain dedicated
 infrastructure nodes where the registry and router pods can run separately from
 pods used for user applications.
+
+If you do not intend to use {product-title} to manage the registry and router, 
+configure the following Ansible settings:
+
+----
+openshift_hosted_manage_registry=false 
+openshift_hosted_manage_router=false
+----
+
+If you are using an image registry other than the default `registry.access.redhat.com`,
+you need to xref:advanced-install-configuring-registry-location[specify the desired registry]
+in the  *_/etc/ansible/hosts_* file.
 
 As described in xref:marking-masters-as-unschedulable-nodes[Configuring
 Schedulability on Masters], master hosts are marked unschedulable by default. If
--- a/_distro_map.yml
+++ b/_distro_map.yml
@@ -62,6 +62,9 @@ openshift-enterprise:
     enterprise-3.5:
       name: '3.5'
       dir: container-platform/3.5
+    enterprise-3.6:
+      name: '3.6'
+      dir: container-platform/3.6
 openshift-dedicated:
   name: OpenShift Dedicated
   author: OpenShift Documentation Project <dev@lists.openshift.redhat.com>
--- a/index-commercial.html
+++ b/index-commercial.html
@@ -426,6 +426,7 @@ height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
 <div class="btn-group">
   <button type="button" class="btn btn-primary dropdown-toggle" data-toggle="dropdown" aria-expanded="false">Select Version<span class="caret"></span></button>
   <ul class="dropdown-menu" role="menu">
+    <li><a href="container-platform/3.6/"><i class="fa fa-arrow-circle-o-right"></i> Container Platform 3.6</a></li>
     <li><a href="container-platform/3.5/"><i class="fa fa-arrow-circle-o-right"></i> Container Platform 3.5</a></li>
     <li><a href="container-platform/3.4/"><i class="fa fa-arrow-circle-o-right"></i> Container Platform 3.4</a></li>
     <li><a href="container-platform/3.3/"><i class="fa fa-arrow-circle-o-right"></i> Container Platform 3.3</a></li>


mburke5678:
@nak3 @sdodson Please take a look. Re: [Bugzilla 1470567](https://bugzilla.redhat.com/show_bug.cgi?id=1470567)

sdodson (approved):

mburke5678:
@zhaozhanqi Please take a look.

install_config/install/advanced_install.adoc
```diff
@@ -663,6 +663,7 @@ If you are using an image registry other than the default at
 ----
 oreg_url=example.com/openshift3/ose-${component}:${version}
 openshift_examples_modify_imagestreams=true
+openshift_hosted_manage_registry=false
```
jianlinliu: openshift_hosted_manage_registry and openshift_hosted_manage_router determine installer would deploy registry/router or not, this option has nothing with "Configuring a Registry Location".  If user are using an image registry other than the default at registry.access.redhat.com, installer would deploy router/registry with 'openshift3/ose-${component}:${version}' image by default into this cluster, where to pull image depends on openshift_docker_additional_registries setting, if user set openshift_docker_additional_registries=example.com in the inventory host file, everything is working well. So openshift_hosted_manage_registry option should be replaced by openshift_docker_additional_registries. Of course, if user wan to deploy registry/router later after installation is completed, openshift_hosted_manage_registry and openshift_hosted_manage_router setting would be meaningful.

install_config/install/advanced_install.adoc
```diff
@@ -1081,9 +1085,8 @@ pods. They are set to `region=infra` by default:
 # openshift_registry_selector='region=infra'
 ----
 
-The default router and registry will be automatically deployed during
-installation if nodes exist in the `[nodes]` section that match the selector
-settings. For example:
+The default router and registry are deployed even though nodes does not exist
```
jianlinliu: Once merged this change, and go through the whole "Configuring Dedicated Infrastructure Nodes" section, that would be really easy to confuse user. This is not a good place to insert the change.
How about move "The default router and registry are deployed even though nodes does not exist in the [nodes] section that match the selector settings." to the following "Important" table, and change the sentence like this:
"The default router and registry are still deployed and failed with pending status even though nodes does not exist in the [nodes] section that match the selector settings"

jianlinliu (commented):
just added some comments

install_config/install/advanced_install.adoc
nak3: I think I understand. It is so strange that `openshift_hosted_manage_registry=false` is explained in the **Configuring a Registry Location** section. Replacing it with `openshift_docker_additional_registries` might be one option, but it would be better to simply delete the `openshift_hosted_manage_registry=false` and its explanation from **Configuring a Registry Location** section.

nak3: Hmm... without `openshift_docker_additional_registries=example.com`, current setting below makes a failure of deploying registry image? If so, `openshift_docker_additional_registries=example.com` would be better to be added in the doc as @jiajliu mentioned (though it is different topic from this PR).

~~~
 oreg_url=example.com/openshift3/ose-${component}:${version}
 openshift_examples_modify_imagestreams=true
~~~

jianlinliu: It would be perfect if openshift_docker_additional_registries replace this line, because if user are using a customized registry, but do not openshift_docker_additional_registries, installer would pull registry/router images from product registry. Maybe that is not the behavior which user has not expected.
Of course, if it is also acceptable to simply delete the openshift_hosted_manage_registry line, at latest, we did not make it worse.

nak3: Ummmm...If `openshift_docker_additional_registries=example.com` is set, the configuration becomes:

~~~
add_registry=--add-registry registry.access.redhat.com --add-registry example.com
~~~

So, as far as `registry.access.redhat.com` is  accessible, I think users cannot change the registry to pull the image of registry and still pull it from `registry.access.redhat.com` (because it is first registry in the line). But, it may be natural that we assume that `registry.access.redhat.com` is blocked or disconnected, when users want to change the location.  But it is our "assumption"... 

Anyway, I think that it is alright to replace the line.

jianlinliu: Interesting, in my testing, openshift_docker_additional_registries=example.com, the docker config file would have such lines which is different with yours:
ADD_REGISTRY='--add-registry example.com --add-registry registry.access.redhat.com'

nak3: Umm, you are correct. `oo_prepend_strings_in_list` should add it as `prepend`. Probably, it is my bad. Let's replace it with `openshift_docker_additional_registries`.

mburke5678:
@jianlinliu @nak3 Can you look at my changes to see if I meet your expectations? I was not totally clear on what you were asking me to change the `openshift_hosted_managed_registry` to.

jianlinliu:
Most of changes look good to me, but after review https://github.com/mburke5678/openshift-docs/blob/d0f6f32452c2bea89606e6583ae8b82df973b2fd/install_config/install/advanced_install.adoc#configuring-dedicated-infrastructure-nodes, seem like the format is messed up. How about this:

**Configuring Dedicated Infrastructure Nodes**
The openshift_router_selector and openshift_registry_selector Ansible settings determine the label selectors used when placing registry and router pods. They are set to region=infra by default:
```
# default selectors for router and registry services
# openshift_router_selector='region=infra'
# openshift_registry_selector='region=infra'
```

The registry and router are only able to run on node hosts with the region=infra label. Ensure that at least one node host in your {product-title} environment has the region=infra label. For example:
```
[nodes]
infra-node1.example.com openshift_node_labels="{'region': 'infra','zone': 'default'}"
```

[IMPORTANT] 

-  The default router and registry are still deployed and failed with pending status even though nodes does not exist in the [nodes] section that match the selector settings.

mburke5678:
@jianlinliu PTAL. I changed the wording of the Important note slightly. Apologies for the formatting error. I thought I made that change yesterday to fix that.

jianlinliu:
LGTM.

mburke5678:

install_config/install/advanced_install.adoc
```diff
@@ -1081,9 +1092,8 @@ pods. They are set to `region=infra` by default:
 # openshift_registry_selector='region=infra'
 ----
 
-The default router and registry will be automatically deployed during
-installation if nodes exist in the `[nodes]` section that match the selector
-settings. For example:
+The registry and router are only able to run on node hosts with the `region=infra label`.
```
bfallonf: The backtick is going one word too far. So: "`region+infra` label"?

install_config/install/advanced_install.adoc
```diff
@@ -1092,15 +1102,26 @@ infra-node1.example.com openshift_node_labels="{'region': 'infra','zone': 'defau
...
+If you do not intend to use {product-title} to manage the registry and router 
+configure the following Ansible settings:
+
+----
+openshift_hosted_manage_registry=false 
+openshift_hosted_manage_router=false
+----
+
+If you are using an image registry other than the default `registry.access.redhat.com`,
+you need to specify the xref:advanced-install-configuring-registry-location[specify the desired registry]
```
bfallonf: Double "specify the". There's one inside the link and outside.

install_config/install/advanced_install.adoc
```diff
@@ -1092,15 +1102,26 @@ infra-node1.example.com openshift_node_labels="{'region': 'infra','zone': 'defau
...
-environment has the `region=infra` label.
+If there is not a node in the [nodes] section that matches the selector settings,
+the default router and registry will be deployed as failed with `Pending` status.
 ====
 
 It is recommended for production environments that you maintain dedicated
 infrastructure nodes where the registry and router pods can run separately from
 pods used for user applications.
 
+If you do not intend to use {product-title} to manage the registry and router
```
bfallonf: Comma after "router" above.

bfallonf:
@mburke5678 A few nits.

mburke5678:
@jianlinliu @nak3 @sdodson Can you confirm if the docs change is for 3.5 and forward? It appears this issue was not in 3.4?

mburke5678:
[rev_history]
|xref:../install_config/install/advanced_install.adoc#install-config-install-advanced-install[Advanced Installation]
|Clarified information about using a non-default registry and router in the xref:..install_config/install/advanced_install.adoc#advanced-install-configuring-registry-location[Configuring a Registry Location] section.
%

mburke5678:
@sdodson @jianlinliu 
Do you have an appropriate `${component}:${version}` that I can use as an example in the file with the
`oreg_url={registry}/openshift3/ose-${component}:${version}` command?

sdodson:
@mburke5678 it should literally be `oreg_url='registry.example.com/openshift3/ose-${component}:${version}' if your registry is registry.example.com. You should not substitute values for ${component} and ${version} openshift will replace those for you at runtime.

mburke5678:
[rev_history]
|xref:install_config/install/advanced_install.adoc#configuring-dedicated-infrastructure-nodes[Advanced Installation]
|Changed wording to correct that a default router and registry will *not* be automatically deployed during installation if nodes exist in the [nodes] section.
%

ahardin-rh:
@mburke5678 The xref needs to have the xref:../ preceding the file path

vikram-redhat:
@mburke5678 - there should only be one revision history per PR. If you need to make an edit to an existing one, simply edit the existing one instead of adding a new one.

PR #4991, opened 2017-08-10
Bug 1463691 Admonition on egress routes for aws

bfallonf:
For: https://bugzilla.redhat.com/show_bug.cgi?id=1463691

cc @danwinship 

Can probably be more specific.

### DIFF
--- a/admin_guide/managing_networking.adoc
+++ b/admin_guide/managing_networking.adoc
@@ -320,6 +320,12 @@ the number of local MAC addresses that the network interface can handle before
 falling back to filtering MAC addresses in software.
 ====
 
+[IMPORTANT]
+====
+Currently, the egress router is not compatible with Amazon AWS due to AWS not
+being compatible with macvlan traffic.
+====
+
 [[admin-guide-limit-pod-access-important-deployment-considerations]]
 *Deployment Considerations*
 


admin_guide/managing_networking.adoc
```diff
@@ -320,6 +320,14 @@ the number of local MAC addresses that the network interface can handle before
 falling back to filtering MAC addresses in software.
 ====
 
+[IMPORTANT]
+====
+Currently, the egress router is not compatible with Amazon AWS due to AWS not
+compatible with macvlan traffic.
```
danwinship: missing word or something. ("not *being* compatible"?)

admin_guide/managing_networking.adoc
```diff
@@ -320,6 +320,14 @@ the number of local MAC addresses that the network interface can handle before
...
+
+See link:https://aws.amazon.com/answers/networking/controlling-vpc-egress-traffic/[the Amazon AWS documentation for alternate methods for egress routing].
```
danwinship: No, nothing there helps OpenShift *users*, who need a solution inside OpenShift (that is aware of the difference between different pods, etc). It just describes different ways that the OpenShift developers might be able to make this feature work on AWS.

bfallonf: @danwinship Sure, I get what you mean. Would you suggest wording it that way instead? So:

"See the Amazon AWS documentation for suggested configurations that don't involve egress routing"

Or is there a more precise thing we could suggest?

danwinship: No, there's nothing you can suggest to users here. Nothing that AWS does is going to help, because OpenShift itself doesn't take advantage of those mechanisms, and there's no way they can take advantage of them from within pods.

danwinship: (So just drop the sentence)

bfallonf:
Thanks for the thoughts @danwinship I've done as suggested.

@openshift/team-documentation peer review please?

mburke5678:
@bfallonf LGTM

bfallonf:
Thanks @mburke5678 . I'll merge.

bfallonf:
[rev_history]
|Added admonition to the xref:../admin_guide/managing_networking.adoc#admin-guide-limit-pod-access-egress-router[Using an Egress Router to Allow External Resources to Recognize Pod Traffic] section about Amazon AWS not working with the egress router. 
%

PR #4999, opened 2017-08-11
Update access mode to supported ones

liangxia:
According to https://docs.openshift.com/container-platform/3.6/architecture/additional_concepts/storage.html#pv-access-modes , GCE persistent disks only supports access mode ReadWriteOnce.


liangxia:
Apply to 3.6 and above.

PR #5024, opened 2017-08-15
Add an upgrade known issues document

sdodson:
Just a stub document for now so we can start cataloging upgrade issues

### DIFF
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -341,6 +341,8 @@ Topics:
         File: blue_green_deployments
       - Name: Operating System Updates and Upgrades
         File: os_upgrades
+      - Name: Known Upgrade Issues
+        File: known_issues
   - Name: Downgrading
     File: downgrade
     Distros: openshift-enterprise
--- a/install_config/upgrading/known_issues.adoc
+++ b/install_config/upgrading/known_issues.adoc
@@ -1,0 +1,194 @@
+[[install-config-upgrading-known-issues]]
+= Overview
+{product-author}
+{product-version}
+:data-uri:
+:icons:
+:experimental:
+:prewrap!:
+
+When upgrading your cluster there are two primary data migrations that take place.
+The first occurs while still running the current version of {product-title}
+and the second happens after the upgrade completes.
+
+If the pre-upgrade migration fails the upgrade will be halted and the admin will
+need to resolve any data consistency problems before attempting the upgrade again.
+If the post-upgrade migration fails the upgrade will complete and the admin should
+investigate the data migration issues as soon as possible. This document will
+catalog known data consistency problems and how to resolve them.
+
+
+== Bug https://bugzilla.redhat.com/show_bug.cgi?id=1435132[1435132] - RoleBindingRestriction object is not deleted after namespace is deleted
+
+Orphaned `rolebindingrestrictions` caused by a bug in the namespace cleanup logic that
+failed to delete them when the namespace was deleted.
+
+=== Error
+
+----
+TASK [Upgrade all storage] *********************************************************************************************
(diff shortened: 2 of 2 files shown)


sdodson:
@enj can you add all known issues you're aware of today?

enj:
@sdodson there are two well know issues:

* Orphaned `rolebindingrestrictions` caused by a bug in the namespace cleanup logic that failed to clean them up when the namespace was deleted.  These can be safely deleted:
    1. Get all `rolebindingrestrictions` across all namespaces
    2. Check to see if the associated namespace exists
    3. Delete the `rolebindingrestriction` if the namespace does not exist
* Orphaned `oauthclientauthorization` caused by the deletion of the associated OAuth client.  These can be safely deleted:
    1. Get all `oauthclientauthorizations`
    2. Check to see if the associated OAuth client exists (can be either an OAuth client or a service account)
    3. Delete the `oauthclientauthorization` if the associated OAuth client does not exist

sdodson:
@enj can you open bugs on each of those if there aren't any and add commits to this PR that includes exact steps for clearing those problems?

sdodson:
https://bugzilla.redhat.com/show_bug.cgi?id=1435132

sdodson:
https://bugzilla.redhat.com/show_bug.cgi?id=1482978

enj:
```bash
# error:     rolebindingrestrictions/match-own-service-accounts -n ywzsg: namespaces "ywzsg" not found

IFS=$'\n'
for rbline in $(oc get rolebindingrestrictions --all-namespaces | tail -n +2); do
	ns=$(echo "$rbline" | awk '{ printf $1; }')
	name=$(echo "$rbline" | awk '{ printf $2; }')
	if ! oc get namespace "$ns"; then
		echo "Bad namespace: $ns"
		(oc create ns "$ns"; sleep 10; oc delete rolebindingrestrictions "$name" -n "$ns"; sleep 10; oc delete namespace "$ns") &
	fi
done
unset IFS

for oa in $(oc get -o name oauthclientauthorization); do
	if [[ "$oa" != *":system:serviceaccount:"* ]]; then
		client=$(echo "$oa" | cut -d : -f 2)
		query="$(oc get oauthclient $client 2>&1)"
		echo "$query" | grep NotFound
		if [ "$?" == "0" ]; then
			echo "Not found: $query"
			oc delete $oa
		fi

	fi
done

for oa in $(oc get -o name oauthclientauthorization); do
	if [[ "$oa" == *":system:serviceaccount:"* ]]; then
		ns=$(echo "$oa" | cut -d : -f 4)
		sa=$(echo "$oa" | cut -d : -f 5)
		# echo "Found: $oa -> ns:$ns  sa: $sa"
		query="$(oc get sa $sa -n $ns 2>&1)"
		echo "$query" | grep "NotFound"
		if [ "$?" == "0" ]; then
			echo "Missing sa: $sa"
			oc delete "$oa"
			echo "    Delete operation: $?"
		fi
		

	fi
done
```

enj:
@sdodson @adellape Can we get my last two comments and the two BZ links added to the PR's contents?

sdodson:
You can push commits to my branch but i'll add them. I don't think copying and pasting from the shell script above is sufficient though.

sdodson:
@enj @adellape ok i'm done with this for now. This would be something we want in master and 3.6.

enj:
@brenton @smarterclayton this is ready minus any changes that @adellape wants to make.  Once this is merged into master / 3.6, we will want to let our support people know so they can spot these problems quickly.

enj (approved):

sdodson:
Squashed changes down.

enj:
This can be closed.

PR #5031, opened 2017-08-16
Updated path to Image Signatures topic

ahardin-rh:
@adellape FYI


ahardin-rh:
No revision history needed

PR #5033, opened 2017-08-16
Add Jobs and Cron Jobs topics

ahardin-rh:


ahardin-rh:
@wgordon17 @tiwillia FYI
The Cron Jobs topic is currently in Online, but a redirect is in place, so users are currently redirected to the index page. I asked @nhr for assistance in removing that. If I recall correctly, the redirect was put in place a while ago during the Online content audit we did.

ahardin-rh:
Cherry-picked to Online via https://github.com/openshift/openshift-docs/pull/5034 so dropping the Online label here. This will be applied to the other branches during the weekly docs release.

PR #5043, opened 2017-08-18
Change memory requests and limits units

makentenza:
Using the example inventory in OCP v3.6 will cause Builds to fail due to an incorrect unit assigned to both requests and limits in memory.

This PR basically changes those units from 'm' to 'Mi' in order to solve this issue.

The following errors can be seen during any build with the existing values:

```docker_sandbox.go:263] NetworkPlugin cni failed on the status hook for pod "xxxxxxxxxx": Unexpected command output nsenter: cannot open : No such file or directory```


sdodson (approved):

ahardin-rh:
This was addressed in https://github.com/openshift/openshift-docs/pull/6767  Thanks!

PR #5046, opened 2017-08-18
Removing excluder un/exclude steps per case 01801897

mburke5678:


mburke5678:
[rev_history]
|xref:../install_config/upgrading/automated_upgrades.adoc#install-config-upgrading-automated-upgrades[Performing Automated In-place Cluster Upgrades]
|Removed the `atomic-openshift-excluder exclude` and `atomic-openshift-excluder include` steps.
%

PR #5029, opened 2017-08-16
Import updated API documentation from origin

jim-minter:
Merge after https://github.com/openshift/origin/pull/15788.  @bparees PTAL.

PR #5050, opened 2017-08-21
add example for using templateinstance API

jim-minter:
@bparees wdyt?

https://trello.com/c/jjZREiDc/899-3-template-instantiate-rest-api-techdebttemplates

### DIFF
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -944,6 +944,8 @@ Dir: rest_api
 Topics:
 - Name: Overview
   File: index
+- Name: Examples
+  File: examples
 - Name: Revision History
   File: revhistory_rest_api
   Distros: openshift-enterprise,openshift-dedicated
--- a/rest_api/examples.adoc
+++ b/rest_api/examples.adoc
@@ -1,0 +1,128 @@
+[[rest-api-examples]]
+= Examples
+{product-author}
+{product-version}
+:data-uri:
+:icons:
+:experimental:
+:toc: macro
+:toc-title:
+
+toc::[]
+{nbsp} +
+This page includes usage examples for OpenShift's REST API.  The examples are
+presented as https://curl.haxx.se/[curl] and
+https://stedolan.github.io/jq/[jq] command calls.  The examples are
+parameterised using environment variables as follows:
+
+[options="header",cols="1,3"]
+|===
+| Environment variable
+| Purpose
+
+| TOKEN
+| Authentication token for OpenShift.  If using X.509 authentication, remove
+lines referencing $TOKEN and provide a client certificate and key instead, e.g.
+via the curl `--cert` and `--key` arguments.
+
+| ENDPOINT
+| TCP endpoint of OpenShift API server, e.g. 127.0.0.1:8443.  Without loss of
+generality, in these examples it is assumed that the API server is presented via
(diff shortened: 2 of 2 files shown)


bparees:
Content wise it looks great!  But I'm debating whether this new general "api examples" section is the best way to structure it.  I'm worried about it becoming a dumping ground.  (and for now it's really lonely).

Might it be better under an "advanced" section/topic under the existing templates content?

@bmcelvee wdyt (last commit only)?

bparees:
(I should note i can see the value in doing it the way you have, too, since it provides a natural location for the  "here's how to invoke apis via curl in general" content).

bmcelvee:
I can see the concern over the API Example section getting cluttered or becoming a dumping ground, but for now my vote is to keep it in it's current location. If the content gets out of hand, we can always reassess. The way you have it does provide a natural location for API examples. And if we move it under the existing templates content, keeping the example content up to date could be a challenge.

bparees:
sounds good, thanks @bmcelvee 

this lgtm, though it's in line behind the other PRs to bring in the new api doc content.

jim-minter:
@ahardin-rh this builds on #5029 and should also be for origin + anything we release subsequently.  Once #5029 merges please will you merge this?

vikram-redhat:
@jim-minter A few questions:

1. Since #5029 already contains commit 5a4ba0f53bb064a18d6b2261e89e706973402a95 which is merged into master (origin), we don't need that commit in this PR. Any particular reason why that commit is here as well? 

2. You mentioned in #5029 that this should only go in origin. Is that correct for this PR as well? I am assuming so, but @bparees has added the labels, so I am not sure. You mentioned that it should go into origin now and anything that is branched in the future. Does that mean it should go into origin now, and the next time it should be in the enterprise docs is when 3.7 comes out (we don't publish separate docs for z stream releases so 3.6.1 won't count, but we can publish it for as 3.6 docs when 3.6.1 is released)? 

3. Finally, everything here does actually apply to dedicated and online?

Thanks!

jim-minter:
Hi @vikram-redhat ,

1- I hoped git would do the right thing and eat the additional commit, but it didn't.  Rebased and repushed, thanks.

2- This and #5029 should go into origin and 3.7, but not 3.6.* because it includes API changes made since 3.6 (and I don't want to backport the document generation to 3.6).

3- I don't know exactly what codebases dedicated and online use.  As soon as each is running a version recent enough to contain /api/docs, we should pull the documentation into the relevant branches of openshift-docs.

vikram-redhat:
Thanks @jim-minter. That makes a lot of sense! I have updated the labels. 

cc: @ahardin-rh. @aburdenthehand

jim-minter:
bump - is anything blocking this from being merged?

vikram-redhat:
@ahardin-rh not sure if @bmcelvee is still looking at this one but if you or Brandi have time tomorrow, could you please look at it? 

Does this require a follow up PR?

ahardin-rh:
Sorry for the delay. I will get this merged, then we can do some minor style edits in a separate PR. Thanks!

ahardin-rh:
[rev_history]
|xref:../ rest_api/examples.adoc#rest-api-examples[REST API Examples]
|Added a new REST API Examples topic.
%

PR #5056, opened 2017-08-22
Define Daemonsets as Technology Preview

ruromero:
In OCP 3.2 Daemonsets were added and not defined as Technology Preview, however they are in later releases.


ruromero:
DaemonSets are fully supported since 3.2, therefore this PR is incorrect

PR #5077, opened 2017-08-25
First create the volume for TEMPLATE_FILE then env

git001:
Due to the fact that

```
oc set env dc/router ...
```

redeploys the router and will crash when the files is not there.

I suggest to first add the custom configmap and **then** set the variable for the TEMPLATE_FILE

### DIFF
--- a/install_config/router/customized_haproxy_router.adoc
+++ b/install_config/router/customized_haproxy_router.adoc
@@ -512,12 +512,12 @@ Using `oc` commands::
 +
 [source,bash]
 ----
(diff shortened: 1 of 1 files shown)


ahardin-rh:
LGTM. @knobunc @pecameron PTAL.

knobunc (approved):
LGTM, thanks for the change.

ahardin-rh:
Thank you!

ahardin-rh:
No revision history needed

PR #5084, opened 2017-08-25
Updating docs with Online Pro custom route information

ahardin-rh:
https://trello.com/c/mG64DZgM/842-docs-configure-your-own-dns-providers-to-work-with-a-custom-route

Preview Build: http://file.rdu.redhat.com/~ahardin/08242017/dns-custom-route/dev_guide/routes.html

@Miciah Here is my first pass at updating the conditional statements in the docs. We still need to add some information on the ELB piece, which I need your guidance on. 

cc @abhgupta

### DIFF
--- a/dev_guide/routes.adoc
+++ b/dev_guide/routes.adoc
@@ -36,23 +36,16 @@ click *Create Route* to define and create a route in your project:
 .Creating a Route Using the Web Console
 image::console_create_route.png["Creating a Route Using the Web Console"]
 
+ifndef::openshift-online[]
 Using the CLI, the following example creates an unsecured route:
 
-ifndef::openshift-online[]
 ----
 $ oc expose svc/frontend --hostname=www.example.com
 ----
-endif::[]
-ifdef::openshift-online[]
-----
-$ oc expose svc/frontend
-----
-endif::[]
 
 The new route inherits the name from the service unless you specify one
 using the `--name` option.
 
-ifndef::openshift-online[]
 .YAML Definition of the Unsecured Route Created Above
 [source,yaml]
 ----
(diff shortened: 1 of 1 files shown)


dev_guide/routes.adoc
```diff
@@ -185,6 +197,16 @@ spec:
 ----
 ====
 
+For {product-title} Pro, use certificate and key files from a CA,then specify
```
Miciah: The long `oc create route edge` example command above (the one with the `--cert`, `--key`, and `--ca-cert` flags) and corresponding yaml block (the one with the `key:`, `certificate:`, and `caCertificate:`) can be included unconditionally now, and they would belong here.

ahardin-rh:
@Miciah Thanks! Updated. Please let me know if I missed anything.

dev_guide/routes.adoc
```diff
@@ -68,14 +57,25 @@ spec:
...
+----
+
+For {product-title} Pro, follow this example, with `--hostname` being optional:
 
+----
+$ oc expose svc/frontend --hostname=www.example.com
+----
+endif::[]
+The new route inherits the name from the service unless you specify one
+using the `--name` option.
```
Miciah: This sentence will show up before the yaml example for Online and after for non-Online, right? It seems like a weird inconsistency, and probably it would be better to put it immediately after the comment, before the yaml, but it's not a huge deal.

dev_guide/routes.adoc
```diff
@@ -188,8 +186,57 @@ spec:
   tls:
     termination: edge
 ----
-====
 
+For {product-title} Pro, you can use your own certificate and key files from a
+CA. You can omit the certificate and key files to use the default certificate.
+However, you can still omit the certificate and key files if you want to use the
+default certificate. With {product-title} Starter, you cannot specify a
```
Miciah: Whoops, these two sentences are redundant:

"You can omit the certificate and key files to use the default certificate."

"However, you can still omit the certificate and key files if you want to use the default certificate."

I'd delete the first and keep the latter (with "However" and "still").

Miciah (changes requested):

PR #5117, opened 2017-08-29
update backward compatibility

deads2k:
This updates our oc skew to one release.  This is coming up now for the groupified APIs and when non-CRUD commands in `oc` can start using just the groupified versions them.

https://github.com/kubernetes/community/blob/master/contributors/design-proposals/versioning.md#supported-releases-and-component-skew indicates a one version compatibility guarantee.  While I wouldn't change capriciously, not all changes can be managed with a reasonable amount of debt.


deads2k:
I'm re-considering my options.

PR #5125, opened 2017-08-29
Added the Cluster Limits topic to the Scaling and Performance Guide

ahardin-rh:
https://trello.com/c/6dLrI7Z2/627-scale37-ocp-create-a-limits-topic-in-openshift-performance-scale-guide

Preview Build: http://file.rdu.redhat.com/~ahardin/08292017/scalability-limits/scaling_performance/cluster_limits.html

### DIFF
--- a/scaling_performance/cluster_limits.adoc
+++ b/scaling_performance/cluster_limits.adoc
@@ -1,0 +1,102 @@
+[[scaling-performance-cluster-limits]]
+= Cluster Limits
+{product-author}
+{product-version}
+:data-uri:
+:icons:
+:experimental:
+:toc: macro
+:toc-title:
+:prewrap!:
+
+toc::[]
+
+== Overview
+
+This topic summarizes the limits for objects in {product-title}.
+
+In most cases, exceeding these thresholds results in lower overall performance.
+It does not necessarily mean that the cluster will fail.
+
+Some of the limits represented in this topic are given for the largest possible
+cluster. For smaller clusters, the limits are proportionally lower.
+
+There are many factors that influence the stated thresholds, including the etcd
+version or storage data format.
+
+[[scaling-performance-current-cluster-limits]]
+== {product-title} 3.7 Cluster Limits
+
+[options="header",cols="3,2"]
+|===
+| Limit Type |Current Limit
+
+| Number of nodes footnoteref:[numberofnodes,Having clusters with more than 5000 nodes is not the best option. Consider splitting into multiple clusters.]
+| 2,000
+
+| Number of pods
+| 120,000
+
+| Number of xref:../admin_guide/manage_nodes.adoc#admin-guide-max-pods-per-node[pods per node]
+| 250
+
+| Number of xref:../admin_guide/manage_nodes.adoc#admin-guide-max-pods-per-node[pods per core]
+| 10 is the default value. The maximum supported value is the number of pods per node.
+
+| Number of namespaces
+| 10,000
+
+| Number of pods per namespace footnoteref:[objectpernamespace,There are
+a number of control loops in the system that need to iterate over all objects
+in a given namespace as a reaction to some changes in state. Having a large
+number of objects of a given type in a single namespace can make those loops
+expensive and slow down processing given state changes.]
+| 15,000
+
+| Number of services footnoteref:[servicesandendpoints,Each service port and each service back-end has a corresponding entry in iptables. The number of back-ends of a given service impact the size of the endpoints objects, which impacts the size of data that is being sent all over the system.]
+| 10,000
+
+| Number of back-ends per service
+| 5,000
+
+| Number of deployments per namespace footnoteref:[objectpernamespace]
+| 20,000
+
+|===
+
+[[scaling-performance-planning-your-environment-according-to-cluster-limits]]
+== Planning Your Environment According to Cluster Limits
+
+[IMPORTANT]
+====
+Oversubscribing the physical resources on a node affects resource guarantees the
+Kubernetes scheduler makes during pod placement. Learn what measures you can
+take to xref:../../admin_guide/overcommit.adoc#disabling-swap-memory[avoid memory swapping].
+====
+
+While
+xref:../install_config/install/planning.adoc#install-config-install-planning[planning
+your environment], determine how many pods are expected to fit per node:
+
+----
+Maximum Pods per Cluster / Expected Pods per Node = Total Number of Nodes
+----
+
+The number of pods expected to fit on a node is dependent on the application
+itself. Consider the application's memory, CPU, and storage requirements.
+
+.Example Scenario
+
+If you want to scope your cluster for 2200 pods per cluster, you would need at
+least nine nodes, assuming that there are 250 maximum pods per node:
+
+----
+2200 / 250 = 8.8
+----
+
+If you increase the number of nodes to 20, then the pod distribution changes to
+110 pods per node:
+
+----
+2200 / 20 = 110
+----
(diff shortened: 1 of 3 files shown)


ahardin-rh:
@jeremyeder Here's my first draft of the new Limits topic. PTAL.

scaling_performance/cluster_limits.adoc
```diff
@@ -0,0 +1,118 @@
...
+| 15,000
+| 50,000
+
+| Number of services footnoteref:[servicesandendpoints,Each service port and each service back-end has a corresponding entry in iptables. The number of back-ends of a given service impact the size of the endpoints objects, which impacts the size of data that is being sent all over the system.]
+| 10,000
+| 100,000
+
+| Number of all services back-ends
+| To be determined
+| 500,000
```
jeremyeder: Please remove all rows that contain "To be determined".  I think we should handle these as research projects and do those as one-offs as more specific requirements are handed down.

jeremyeder: OK, please update accordingly, and I'll take another pass at review.  Thanks!

jeremyeder (changes requested):

ahardin-rh:
@jeremyeder Updated PTAL

scaling_performance/cluster_limits.adoc
```diff
@@ -0,0 +1,102 @@
...
+In most cases, exceeding these thresholds results in lower overall performance.
+It does not necessarily mean that the cluster will fail over.
+
+Some of the limits represented in this topic are given for the largest possible
+cluster. For smaller clusters, the limits are proportionally lower.
+
+There are many factors that influence the stated thresholds, including the etcd
+version or storage data format.
+
+[[scaling-performance-current-cluster-limits]]
```
jeremyeder: One piece of additional feedback is to rename "Current Limits" with the precise version number.  My intent is to update this page with each release with new columns, even if none of the values themselves change.  So "Current" is ambiguous, please make it 3.7.

scaling_performance/cluster_limits.adoc
```diff
@@ -0,0 +1,102 @@
...
+|Total number of all objects footnoteref:[allobjects,There is a limitation on the total number of objects on the system, as this affects etcd and its resource consumption.]
+| 250,000
+
+| Number of nodes footnoteref:[numberofnodes,Having clusters with more than 5000 nodes is not the best option. Consider splitting into multiple clusters.]
+| 2,000
+
+| Number of pods
+| 120,000
+
+| Number of pods per node footnoteref:[podspernodeorcore, The limit for number of pods on a given node is the minimum from the _pod per node_ and _pods per core multiplied by the number of cores of a node_.]
```
jeremyeder: can you just link here to the existing pods-per-core docs?

scaling_performance/cluster_limits.adoc
```diff
@@ -0,0 +1,102 @@
...
+| 250
+
+| Number of pods per core footnoteref:[podspernodeorcore]
+| 10
```
jeremyeder: This is tricky...10 is the default, but it's not the supported maximum.  People can turn it off and still be supported.
It is however an out of the box default that people do hit now and then, so I don't think it should be left off this limits page, just that somehow we need to indicate that it is not the maximum supported number.  The maximum supported will always be pods-per-node.

ahardin-rh:
@jeremyeder Thanks! This is now updated.

ahardin-rh:
@openshift/team-documentation please peer review

jeremyeder:
@ahardin-rh would you mind creating another preview build?

scaling_performance/cluster_limits.adoc
```diff
@@ -0,0 +1,102 @@
...
+your environment], determine how many pods are expected to fit per node:
+
+----
+Maximum Pods per Cluster / Expected Pods per Node = Total Number of Nodes
+----
+
+.Example Scenario
+
+If you want to scope your cluster for 2200 pods per cluster, you would need at
+least 9 nodes, assuming that there are 250 maximum pods per node:
```
bfallonf: If we're not talking about an actual value for a parameter or something, I'd suggest s/9/nine

bfallonf:
@ahardin-rh Just one nit from me.

ahardin-rh:
@jeremyeder [Preview build](http://file.rdu.redhat.com/%7Eahardin/08292017/scalability-limits/scaling_performance/cluster_limits.html) is updated

jeremyeder:
@openshift/scale-performance-review @mffiedler PTAL, preview docs build here:
http://file.rdu.redhat.com/~ahardin/08292017/scalability-limits/scaling_performance/cluster_limits.html

scaling_performance/cluster_limits.adoc
```diff
@@ -0,0 +1,102 @@
...
+:prewrap!:
+
+toc::[]
+
+== Overview
+
+This topic summarizes the limits for objects in {product-title}.
+
+In most cases, exceeding these thresholds results in lower overall performance.
+It does not necessarily mean that the cluster will fail over.
```
mffiedler: I think "fail" would be better than "fail over" which implies some specific HA behaviors.

scaling_performance/cluster_limits.adoc
```diff
@@ -0,0 +1,102 @@
...
+[IMPORTANT]
+====
+Oversubscribing the physical resources on a node affects resource guarantees the
+Kubernetes scheduler makes during pod placement. Learn what measures you can
+take to xref:../../admin_guide/overcommit.adoc#disabling-swap-memory[avoid memory swapping].
+====
+
+While
+xref:../install_config/install/planning.adoc#install-config-install-planning[planning
+your environment], determine how many pods are expected to fit per node:
```
mffiedler: @jeremyeder Should we add a short discussion around how the number of pods expected to fit on a node is dependent on the application itself?  i.e. there is no magic in a pod - you are still responsible for understanding the memory/cpu/storage requirements of the application.   Maybe outside the scope of this section, but still a key part of the "pods per node" thing.

scaling_performance/cluster_limits.adoc
```diff
@@ -0,0 +1,102 @@
...
+version or storage data format.
+
+[[scaling-performance-current-cluster-limits]]
+== {product-title} 3.7 Cluster Limits
+
+[options="header",cols="3,2"]
+|===
+| Limit Type |Current Limit
+
+|Total number of all objects footnoteref:[allobjects,There is a limitation on the total number of objects on the system, as this affects etcd and its resource consumption.]
```
smarterclayton: This really needs to be in context of either number of namespaces (broad) or number of large applications.  I think this is almost meaningless to an end user.

smarterclayton: Tie this to number of namespaces in this comment.  As in "Because namespaces are a unit of tenancy, each namespace has a number of associated etcd objects.  Therefore, total scale will likely be a factor of the number of namespaces as described below" or something.

ahardin-rh:
Thanks, all! This is updated: http://file.rdu.redhat.com/%7Eahardin/08292017/scalability-limits/scaling_performance/cluster_limits.html

scaling_performance/cluster_limits.adoc
jeremyeder: It is definitely meaningless.  I'll vote for dropping it entirely.  Some of the content of this PR came from it's upstream equivalent, and I have already dropped many other meaningless rows that are kept upstream (including rows that just show "TBD").

ahardin-rh: @jeremyeder Thanks. I went ahead and dropped it.

ahardin-rh:
[rev_history]
|xref:../scaling_performance/cluster_limits.adoc#scaling-performance-cluster-limits[Cluster Limits]
|new topic summarizing the limits for objects in {product-title}.
%

PR #5127, opened 2017-08-30
Updated guidelines to include source tags before code blocks

gaurav-nelson:
Updated guidelines to include source tags before code blocks.

### DIFF
--- a/contributing_to_docs/doc_guidelines.adoc
+++ b/contributing_to_docs/doc_guidelines.adoc
@@ -404,16 +404,12 @@ configuration file represent, then use callouts to provide that information.
 
 Follow these general guidelines when using code blocks:
 
-* Do NOT show replaceables within code blocks.
-
 * Do NOT use any markup in code blocks; code blocks generally do not accept any markup.
 
-* Try to use callouts to provide information on what the output represents when required.
-
-For all code blocks, you must include an empty line above a code block.
-
+* For all code blocks, you must include an empty line above a code block.
++
 Acceptable:
-
++
 ....
 Lorem ipsum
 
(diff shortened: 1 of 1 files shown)


gaurav-nelson:

gaurav-nelson:
On line 407 we say: `Do NOT show replaceables within code blocks.`
but on line 483 we have: `To markup command syntax, use the code block and wrap the replaceables in <> with the required command parameters, ` and the give example uses code block with replaceable.

?

bfallonf:
@gaurav-nelson Your changes look fine. Looking at the blame thing in the file, I think the "do not sue code blocks' might be leftover from the beginnings of the repo 3 years ago. Might be good to erase now we've changed our mind about it.

ahardin-rh:
👍 Yeah, we evolved a little since that guideline was introduced. Replaceables in code blocks can be helpful in showing example commands, especially in instances where any actual values are difficult to explain/map to their function.

mburke5678:
@openshift/team-documentation While on the subject, perhaps off topic a bit, but should the no wrap option work for [source,bash,options="nowrap"]? I did not see it working when I tried it recently.
Perhaps, if this is not supported, we could update the guidelines as to when we can use options="nowrap".

adellape:
> While on the subject, perhaps off topic a bit, but should the no wrap option work for [source,bash,options="nowrap"]? I did not see it working when I tried it recently.
> Perhaps, if this is not supported, we could update the guidelines as to when we can use options="nowrap".

@mburke5678 All of our topics have `:prewrap!:` in the top of the file's metadata, which is a global version of `options="nowrap"`:

https://asciidoctor.cn/docs/user-manual/#to-wrap-or-to-scroll

I've noticed that the horizontal scrollbar may not render in local builds, but it should be good when it actually hits docs.openshift.com and the Customer Portal.

contributing_to_docs/doc_guidelines.adoc
```diff
@@ -410,13 +410,18 @@ Follow these general guidelines when using code blocks:
 
 * Try to use callouts to provide information on what the output represents when required.
 
-For all code blocks, you must include an empty line above a code block.
+For all code blocks, you must:
```
adellape: Do we want to make this sound like a hard requirement, or more just that we are now saying explicitly it's OK to use if you want?

gaurav-nelson: Thanks @adellape I'll update the language.

gaurav-nelson:
@adellape Can you please review and merge?

adellape (approved):
LGTM, thanks! (typo in commit message, FWIW... :smile:)

PR #5131, opened 2017-08-30
Trello opaque integer resources

mburke5678:
https://trello.com/c/W4IJHSHi/368-document-psp-opaque-counted-resources-support-techpreview

### DIFF
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -585,9 +585,8 @@ Topics:
 - Name: Allocating Node Resources
   File: allocating_node_resources
   Distros: openshift-origin,openshift-enterprise,atomic-*
-- File: Opaque Integer Resources
-  Name: opaque_integer_resources
-  Distros: openshift-origin,openshift-enterprise,atomic-*
+- Name: Opaque Integer Resources
+  File: opaque_integer_resources
 - Name: Overcommitting
   File: overcommit
   Distros: openshift-origin,openshift-enterprise
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -585,6 +585,9 @@ Topics:
 - Name: Allocating Node Resources
   File: allocating_node_resources
   Distros: openshift-origin,openshift-enterprise,atomic-*
+- File: Opaque Integer Resources
+  Name: opaque_integer_resources
+  Distros: openshift-origin,openshift-enterprise,atomic-*
 - Name: Overcommitting
   File: overcommit
   Distros: openshift-origin,openshift-enterprise
--- a/dev_guide/compute_resources.adoc
+++ b/dev_guide/compute_resources.adoc
@@ -315,95 +315,18 @@ $ oc run nginx --image=nginx --limits=memory=400Mi
 endif::[]
 ====
 
-[[opaque-integer-resources]]
+[[opaque-integer-resources-dev]]
 === Opaque Integer Resources
 
-Opaque integer resources allow cluster operators to advertise new node-level
-resources that would be otherwise unknown to the system. Users can consume these
-resources in pod specifications, just like CPU and memory. The scheduler takes
-care of the resource accounting so that no more than the available amount is
-simultaneously allocated to pods.
+include::admin_guide/opaque_integer_resources.adoc[tag=oir-intro]
 
-[NOTE]
-====
-Opaque integer resources are Alpha currently, and only resource accounting is
-implemented. There is no resource quota or limit range support for these
-resources, and they have no impact on QoS.
-====
+The cluster administrator is usually responsible for creating the resources and making them available. 
+For more information on creating opaque integer resources, see xref:../admin_guide/opaque_integer_resources.adoc#admin-guide-opaque-resources[Opaque Integer Resources] in the Administrator Guide. 
 
-Opaque integer resources are resources that begin with the prefix
-`pod.alpha.kubernetes.io/opaque-int-resource-`. The API server
-restricts quantities of these resources to whole numbers. Examples of
-_valid_ quantities are `3`, `3000m` and `3Ki`. Examples of _invalid_
-quantities are `0.5` and `1500m`.
+To consume an opaque integer resource in a pod, edit the pod to
+include the name of the opaque resource as a key in the `spec.containers[].resources.requests` field.
 
-There are two steps required to use opaque integer resources. First, the cluster
-operator must advertise a per-node opaque resource on one or more nodes. Second,
-users must request the opaque resource in pods.
-
-To advertise a new opaque integer resource, the cluster operator should
-submit a `PATCH` HTTP request to the API server to specify the available
-quantity in the `status.capacity` for a node in the cluster. After this
-operation, the node's `status.capacity` will include a new resource. The
-`status.allocatable` field is updated automatically with the new resource
-asynchronously by the kubelet.
-
-[NOTE]
-====
-Since the scheduler uses the node `status.allocatable` value when evaluating pod
-fitness, there may be a short delay between patching the node capacity with a
-new resource and the first pod that requests the resource to be scheduled on
-that node.
-====
-
-The HTTP request below advertises five `foo` resources on node
-`openshift-node-1`. Note that `~1` is the encoding for the character `/` in the
-patch path. The operation path value in the JSON-Patch is interpreted as a
-JSON-Pointer. For more details, refer to
-link:https://tools.ietf.org/html/rfc6901#section-3[IETF RFC 6901, section 3].
-
-[source,http]
-----
-PATCH /api/v1/nodes/openshift-node-1/status HTTP/1.1
-Accept: application/json
-Content-Type: application/json-patch+json
-Host: openshift-master:8080
-
-[
-  {
-    "op": "add",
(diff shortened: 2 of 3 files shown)


mburke5678:
@aveshagarwal  Can you PTAL at my re-org of your Opaque Integer Resources text?

dev_guide/compute_resources.adoc
```diff
@@ -337,31 +337,38 @@ restricts quantities of these resources to whole numbers. Examples of
...
+They are called _integer resources_ 
+because they must be advertised in integer amounts. 
+
+Some example use cases for opaque integer resources:
+
+* LLC (last-level cache) allocation.
+* GPU devices
+* FPGA devices
+* MICs (Many-Integrated Core, such as Xeon Phi)
+* *Slots* for sharing bandwidth to a parallel file system.
```
aveshagarwal: Any reason, you have Slots in bold text?

dev_guide/compute_resources.adoc
```diff
@@ -404,6 +417,23 @@ spec:
...
+[source,bash]
+----
+Conditions:
+  Type    Status
+  PodScheduled  False
+...
+Events:
+  ...
+  ... Warning   FailedScheduling  pod (oir-demo-2) failed to fit in any node
+fit failure summary on nodes : Insufficient pod.alpha.kubernetes.io/opaque-int-resource-dongle (1)
```
aveshagarwal: It'd better if they have the pod spec mentioned above. As I dont oir-demo-2 and opaque-int-resource-dongle mentioned anywhere above.

aveshagarwal:
@mburke5678 some minor comments, otherwise LGTM.

mburke5678:
@wmengRH Can you take a look at the section on Opaque Integer Resources? I am not sure that QE reviewed the original doc.

dev_guide/compute_resources.adoc
```diff
@@ -404,6 +417,23 @@ spec:
...
+Conditions:
+  Type    Status
+  PodScheduled  False
+...
+Events:
+  ...
+  ... Warning   FailedScheduling  pod (my-pod) failed to fit in any node
+fit failure summary on nodes : Insufficient pod.alpha.kubernetes.io/opaque-int-resource-foo (1)
+----
+
```
wmengRH: the output for OCP is slightly different to Kubernetes. 
```
Events:
  FirstSeen	LastSeen	Count	From			SubObjectPath	Type		Reason			Message
  ---------	--------	-----	----			-------------	--------	------			-------
  14s		0s		6	default-scheduler			Warning		FailedScheduling	No nodes are available that match all of the following predicates:: Insufficient pod.alpha.kubernetes.io/opaque-int-resource-foo (1).
```

wmengRH:
LGTM. Thank you for nice work.

mburke5678:

mburke5678:

dev_guide/compute_resources.adoc
```diff
@@ -331,38 +331,40 @@ implemented. There is no resource quota or limit range support for these
...
-Opaque integer resources are resources that begin with the prefix
-`pod.alpha.kubernetes.io/opaque-int-resource-`. The API server
+Opaque integer resources are called _opaque_ because {product-title} 
+does not know what the resource is, but will schedule a pod on a node 
+only if enough of that resource is available. They are called _integer resources_ 
+because they must be available, or _advertised_, in integer amounts. The API server
 restricts quantities of these resources to whole numbers. Examples of
-_valid_ quantities are `3`, `3000m` and `3Ki`. Examples of _invalid_
-quantities are `0.5` and `1500m`.
+_valid_ quantities are `3`, `3000m` and `3Ki`.
```
bfallonf: Comma after 3000m

dev_guide/compute_resources.adoc
```diff
@@ -331,38 +331,40 @@ implemented. There is no resource quota or limit range support for these
...
-quantity in the `status.capacity` for a node in the cluster. After this
-operation, the node's `status.capacity` will include a new resource. The
-`status.allocatable` field is updated automatically with the new resource
-asynchronously by the kubelet.
+* LLC (last-level cache) allocation.
+* GPU devices
+* FPGA devices
+* MICs (Many-Integrated Core)
+* Slots for sharing bandwidth to a parallel file system.
+// https://github.com/kubernetes/kubernetes/issues/28312
```
bfallonf: I can't tell how the above are use cases. Is there more information that can be put in? They just seem like a list of components, or something similar. But also, I'm not sure what the other acronyms are. This might need to be elaborated on.

dev_guide/compute_resources.adoc
```diff
@@ -331,38 +331,40 @@ implemented. There is no resource quota or limit range support for these
...
-[NOTE]
-====
-Since the scheduler uses the node `status.allocatable` value when evaluating pod
-fitness, there may be a short delay between patching the node capacity with a
-new resource and the first pod that requests the resource to be scheduled on
-that node.
-====
+For example, if a node has 800 GiB of a special kind of disk storage, you could create a name for the special storage, 
+such as `_opaque-int-resource-special-storage_`. You could advertise it in chunks of a certain size, 
+such as 100 GiB. In that case, your node would advertise that it has eight resources of type opaque-int-resource-special-storage.
```
bfallonf: Markup on the last opaque-int-resource-special-storage

dev_guide/compute_resources.adoc
```diff
@@ -331,38 +331,40 @@ implemented. There is no resource quota or limit range support for these
...
-`openshift-node-1`. Note that `~1` is the encoding for the character `/` in the
-patch path. The operation path value in the JSON-Patch is interpreted as a
-JSON-Pointer. For more details, refer to
-link:https://tools.ietf.org/html/rfc6901#section-3[IETF RFC 6901, section 3].
+Opaque integer resource names must begin with the prefix `pod.alpha.kubernetes.io/opaque-int-resource-`. 
 
-[source,http]
+There are two steps required to use opaque integer resources. First, the cluster
+operator must advertise a per-node opaque resource on one or more nodes. Second,
+users must request the opaque resource in pods:
```
bfallonf: Do the above few lines match the following procedure? I think the first part does, but the second is for users. Should the first step be for admins? We're in the dev guide, so there's probably no need to say "users must request X", because if we're talking about OpenShift users "application developers" is probably better, but also it's understood the reader is one of these, so maybe just "Second, request the...".

But if both actions are done by the developer, then I'd suggest just replacing all the above with "To use opaque integer resources:", because otherwise it's superfluous anyway.

bfallonf:
@mburke5678 I made some comments. LMK what you think.

dev_guide/compute_resources.adoc
mburke5678: @bfallonf 
I took that list from a Kubernetes issue in GithHub:
https://github.com/kubernetes/kubernetes/issues/28312

Which was reproduced in this article. The author expanded on the list a bit (but I didn't want to take his words). 
http://www.devoperandi.com/kubernetes-not-just-cpu-and-ram-opaque-integer-resources/

I can pull the list, if you think it inappropriate or not helpful.

dev_guide/compute_resources.adoc
mburke5678: @bfallonf Good question on the roles. I will look into that. 
There is also a task for someone to create that resource so that it can be advertised and consumed by these two steps, which is an admin task (I am told admin types will know how to do this). Any thoughts on where I could put a reference to opaque integer resources where an admin will see it?

mburke5678:
@bfallonf I moved the topic to the adamin guide, as the admin is required to perform a couple of the required tasks. The intro and the step on creating the pod spec are still in the Dev Guide in the Quotas and Limit Ranges topic. How does that look?

Cleaner look here: file.rdu.redhat.com/mburke/trello-opaque-integer-resources/admin_guide/opaque_integer_resources.html  
Can you access this?

_topic_map.yml
```diff
@@ -585,6 +585,9 @@ Topics:
 - Name: Allocating Node Resources
   File: allocating_node_resources
   Distros: openshift-origin,openshift-enterprise,atomic-*
+- File: Opaque Integer Resources
+  Name: opaque_integer_resources
```
bfallonf: switch around "File" and "Name" above cos build issues

dev_guide/compute_resources.adoc
bfallonf: @mburke5678 I think the wording is better.

bfallonf:
@mburke5678 It gives me a "there's nothing to compare" statement in big letters. NP, I can see what you've done in the files changed tab. I think what you've done (putting the content into admin guide but with included section for the dev guide) works well. And with the use case thing, I think what you have here is fine.

Also it's not building, but I made a comment to why.

mburke5678:
[rev_history]
|xref:../admin_guide/opaque_integer_resources.adoc#admin-guide-opaque-resources[Opaque Integer Resources]
|Moved information on opaque integer resources to Administrator Guide
%

vikram-redhat:
@mburke5678 - commits are not squashed.

PR #5132, opened 2017-08-31
Add back tech preview for dedicated and mysql

sspeiche:
The tech preview label is missing from dedicated docs for mysql replication


ahardin-rh:
@sspeiche Thanks!

ahardin-rh:
No revision history needed

Issue #5094, opened 2017-08-26
Incorrect tag for upgrading cluster metrics


ksemaev:
According to https://docs.openshift.org/latest/install_config/upgrading/automated_upgrades.html#automated-upgrading-cluster-metrics
you shoud specify 

> Replace <tag> with v3.6 for the latest version.

This setting gives error after installing:
```
Failed to pull image "docker.io/openshift/origin-metrics-hawkular-metrics:v3.6": rpc error: code = 2 desc = manifest unknown: manifest unknown
Failed to pull image "docker.io/openshift/origin-metrics-heapster:v3.6": rpc error: code = 2 desc = manifest unknown: manifest unknown
```

Repo https://hub.docker.com/r/openshift/origin-metrics-heapster/tags/ does not have this tag. 
Using tag **v3.6.0** fixes situation

gaurav-nelson:
Hello @ksemaev 
Thank you for pointing out this issue. I have made suggested changes and tracking this in PR# #5134.

Closing this, please follow the PR for updates.

PR #5134, opened 2017-08-31
Fixed Incorrect tag for upgrading cluster metrics

gaurav-nelson:
For #5094 

Updated the tag


gaurav-nelson:
Minor updates, no revision history needed.

PR #5163, opened 2017-09-05
Fix access mode typo

liangxia:
Reference https://docs.openshift.org/latest/architecture/additional_concepts/storage.html#pv-access-modes


liangxia:
Apply to branch/enterprise-3.6 and above

PR #5218, opened 2017-09-08
Bug 1467265 Add 'Purge' Option with Uninstall

bmcelvee:
https://bugzilla.redhat.com/show_bug.cgi?id=1467265

@wozniakjan PTAL and can you confirm this belongs in OCP 3.6? Thanks!

### DIFF
--- a/install_config/aggregate_logging.adoc
+++ b/install_config/aggregate_logging.adoc
@@ -428,6 +428,14 @@ for deploying Kibana instances.
 |`openshift_logging_curator_ops_nodeselector`
 |A node selector that specifies which nodes are eligible targets
 for deploying Curator instances.
+
+|`openshift_logging_purge_logging`
+| The common uninstall would keep PVC to prevent unwanted data loss during
+reinstalls. To ensure the Ansible playbook will completely and irreversibly
+remove all logging persistent data including PVC, set
+`openshift_logging_install_logging` to 'false' to trigger uninstallation and
+`openshift_logging_purge_logging` to 'true'. The default is set to 'false'.
+
 |===
 
 


wozniakjan:
@bmcelvee thanks, this is perfect! One little detail, I don't think we are going to have this backported to 3.6. so it should be a new feature introduced in 3.7

bmcelvee:
Thanks, @wozniakjan! @openshift/team-documentation PTAL

install_config/aggregate_logging.adoc
```diff
@@ -420,6 +420,13 @@ for deploying Kibana instances.
 |`openshift_logging_curator_ops_nodeselector`
 |A node selector that specifies which nodes are eligible targets
 for deploying Curator instances.
+
+|`openshift_logging_purge_logging`
+|When `openshift_logging_install_logging` is set to 'false' to trigger
+uninstalation and `openshift_logging_purge_logging` is set to 'true', it will
```
mburke5678: Not sure what "it" is here (but, maybe our end-users will). Same with "trigger uninstallation here and with the `openshift_logging_install_logging` description above. 
I like the way the flag is described in the dev PR as it has a bit of a user story. Maybe we don't need it. 
https://github.com/openshift/openshift-ansible/pull/5189

mburke5678:
@bmcelvee one question and a comment.

install_config/aggregate_logging.adoc
wozniakjan: I meant "it" to refer to the ansible playbook used to install or uninstall logging. Feel free to rewrite the description of the `openshift_logging_purge_logging` in a clearer and more understandable way but I think the dev PR text is unnecessarily long for the docs description. The only reason I was so verbose there is that PR text doesn't have any context while the description in docs is surrounded by rich context.

bmcelvee:
Sorry for the delay, and thank you for your feedback! I changed the wording to include a bit of a user story and add some clarification. Does it look okay?

bmcelvee:
[rev_history]
|Added `openshift_logging_purge_logging` parameter to the
xref:../install_config/aggregate_logging.adoc#aggregate-logging-ansible-variables[Specifying Logging Ansible Variables] section.
%

PR #5219, opened 2017-09-08
Fix bad links

mburke5678:
Fix two bad links in Developer Guide -> Routes to Architecture -> Networking -> Routes. MASTER branch only!


mburke5678:
No rev history needed

vikram-redhat:
No milestone needed if it is not going into any branch.

PR #5229, opened 2017-09-11
Bug 1487360, Clarified that Support needs added for those services to be available

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1487360


ahardin-rh:
@sferich888 PTAL
@abhgupta FYI

ahardin-rh:
@sferich888 Please review

sferich888:
Is there a way we can make this, look more like the quick start, guides and supply screenshots?

ahardin-rh:
@sferich888 I don't think screenshots are necessary for this. The UI is pretty straight forward for this workflow. I vote for simplicity and less screenshots to maintain.

PR #5231, opened 2017-09-11
Bug 1483764, updated Restrictions section

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1483764

### DIFF
--- a/dev_guide/routes.adoc
+++ b/dev_guide/routes.adoc
@@ -54,7 +54,6 @@ using the `--name` option.
 
 ifndef::openshift-online[]
 .YAML Definition of the Unsecured Route Created Above
-====
 [source,yaml]
 ----
 apiVersion: v1
@@ -68,14 +67,11 @@ spec:
     kind: Service
     name: frontend
 ----
-====
-
 <1> For xref:../architecture/networking/routes.adoc#path-based-routes[path-based routing], specify a path component that can be compared against a URL.
 endif::[]
 
 ifdef::openshift-online[]
 .YAML Definition of the Unsecured Route Created Above
-====
 [source,yaml]
 ----
 apiVersion: v1
@@ -87,7 +83,6 @@ spec:
     kind: Service
     name: frontend
 ----
-====
 endif::[]
 
 For information on configuring routes using the CLI, see xref:../architecture/networking/routes.adoc#route-types[Route Types].
@@ -117,7 +112,6 @@ $ oc create route edge --service=frontend \
 ----
 
 .YAML Definition of the Secured Route Created Above
-====
 [source,yaml]
 ----
 apiVersion: v1
@@ -144,7 +138,6 @@ spec:
       [...]
       -----END CERTIFICATE-----
 ----
-====
 
 Currently, password protected key files are not supported. HAProxy prompts for
 a password upon starting and does not have a way to automate this process. To
@@ -174,7 +167,6 @@ $ oc create route edge --service=frontend
 ----
 
 .YAML Definition of the Secured Route Created Above
-====
 [source,yaml]
 ----
 apiVersion: v1
@@ -188,7 +180,6 @@ spec:
   tls:
     termination: edge
 ----
-====
 
 endif::[]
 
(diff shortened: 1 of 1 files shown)


ahardin-rh:
@wgordon17 FYI

@tiwillia PTAL. Also, as part of the related bug, Will suggests that we should also provide guidance on how to create a custom route for Pro users, including basic/generic information for setting up a CNAME DNS record. Can you provide guidance on that bit?

dev_guide/routes.adoc
```diff
@@ -225,14 +225,16 @@ ifdef::openshift-online[]
...
 
-Custom route hosts are not permitted. Instead, the following host template is
-enforced on all user routes:
+Routes are restricted in {product-title} Starter, but are not restricted in
+{product-title} Pro. Custom route hosts are permitted in {product-title} Pro. If
+using {product-title} Starter, the following host template is enforced on all
+user routes:
 
 ----
 <route-name>-<namespace>.1d35.starter-us-east-1.openshiftapps.com
```
wgordon17: This isn't exactly accurate. The "us-east-1" value changes depending on the region the user is provisioned in. I can't think if a good way to generalize this. @luciddreamz Any thoughts?

dev_guide/routes.adoc
```diff
@@ -225,14 +225,16 @@ ifdef::openshift-online[]
...
 ----
 
-Also, custom certificates are not permitted. Only unencrypted routes, edge
+Custom certificates are not permitted. Only unencrypted routes, edge
```
wgordon17: I think this is also incorrect. Custom certificiates *should* work on the Pro tier. @tiwillia Can you confirm?

wgordon17:
@ahardin-rh Something along the lines of

> Once your custom route has been created, you will need to update your DNS provider by creating a CNAME record. Your CNAME record should point your custom domain to the OpenShift Online router as the alias. Your OpenShift Online router is <same per-cluster issue that we have on starter, it will be different for every region>. CNAME records are unable to be set for a naked domain (example.com), and instead requires a subdomain be specified (www.example.com).

@tiwillia Does that sound user friendly and accurate?

luciddreamz:
For reference, here are the current default router shards for every cluster:

- laravel-luciddreamz.1d35.starter-us-east-1.openshiftapps.com
- laravel-luciddreamz.8a09.starter-us-east-2.openshiftapps.com 
- laravel-luciddreamz.a3c1.starter-us-west-1.openshiftapps.com
- laravel-luciddreamz.7e14.starter-us-west-2.openshiftapps.com
- laravel-luciddreamz.193b.starter-ca-central-1.openshiftapps.com
- laravel-luciddreamz.b9ad.pro-us-east-1.openshiftapps.com

We should probably put this in the about page, along with this for the Pro tier:
https://stackoverflow.com/questions/45962980/moving-from-openshift-online-v2-to-openshift-online-v3-alias/45970158#45970158

dev_guide/routes.adoc
tiwillia: Can we provide this as an example, then give the user a method to determine the route themselves?

Users can create a route, and then `oc get route/<route_name>` to see what their external address is.

dev_guide/routes.adoc
tiwillia: @wgordon17 I'm fairly certain we do allow custom certificates in pro tier. When creating a route, custom certificates can be uploaded. This will need to be changed.

tiwillia:
@wgordon17 that sounds fine to me

wgordon17:
@ahardin-rh So, I think this is ready to go, with the updated content. Could you update the PR, and we'll review again?

ahardin-rh:
@wgordon17 @tiwillia Sorry for the delay. This is now updated. PTAL.

Also, the docs team is currently in a merge freeze while we transition to a new team workflow. This should be lifted soon, but just FYI.

dev_guide/routes.adoc
```diff
@@ -225,20 +225,42 @@ ifdef::openshift-online[]
...
+----
+
+Custom certificates are permitted in {product-title} Pro. In {product-title}
+Starter, only unencrypted routes, edge routes using the default certificate, and
+passthrough routes work. Edge routes with custom certificates and re-encrypt
+routes (which necessarily have custom certificates) do not work in
+{product-title} Starter.
 
 These restrictions are enforced in the router. Inspecting the route in the
 console or using `oc get routes` displays the host and certificates specified by
```
Miciah: These two statements are no longer true.  The restriction will be enforced by the API server, which will reject the custom route if it violates restrictions on custom routes (which means there will be no route for `oc get routes` to display).  I can make a separate PR or Bugzilla report if you prefer that over overloading this PR.

dev_guide/routes.adoc
```diff
@@ -225,20 +225,42 @@ ifdef::openshift-online[]
...
-the user. However, custom hosts are not respected, and routes with custom
-certificates do not work.
+the user. However, custom hosts are not respected in {product-title} Starter,
+and routes with custom certificates do not work.
+
+Once your custom route is created, you must update your DNS provider by creating
+a canonical name (CNAME) record. Your CNAME record should point your custom
+domain to the {product-title} router as the alias. Your {product-title} router
+is the same per-cluster issue that i son {product-title} Starter. It is
+different for every region. CNAME records can not be set for a naked domain
```
Miciah: s/i son/is on/

s/can not/cannot/

I'd put the last two sentences about naked domains as a separate paragraph, maybe notated with `[NOTE]`.

I don't understand the sentences, "Your {product-title} router is the same per-cluster issue that i son {product-title} Starter. It is different for every region." I am guessing this could be replaced simply by, "The {product-title} router's domain is different for every cluster."

dev_guide/routes.adoc
Miciah: Correction: There will be a route for `oc get routes` to display, but its status will indicate if the route was restricted.

wgordon17:
I'm also looking to possibly update our custom about.html page, so all the clusters include the route information. It may be easier to just direct users to the about.html page for this information. Stay tuned for that PR's outcome

ahardin-rh:
@Miciah @tiwillia @wgordon17 Updated with the latest changes, just without Will's about.html updates, which we can do in a follow-up PR is you prefer.

dev_guide/routes.adoc
```diff
@@ -225,20 +216,47 @@ ifdef::openshift-online[]
...
 
 These restrictions are enforced in the router. Inspecting the route in the
 console or using `oc get routes` displays the host and certificates specified by
-the user. However, custom hosts are not respected, and routes with custom
-certificates do not work.
+the user. The status of the route indicates if the route is restricted. However,
+custom hosts are not respected in {product-title} Starter, and routes with
+custom certificates do not work.
+
+Once your custom route is created, you must update your DNS provider by creating
```
wgordon17: Once your custom route is created **in {product-title} Pro**

wgordon17 (changes requested):

ahardin-rh:
No revision history needed. These changes will appear in Online only, which does not include revision histories.

PR #5238, opened 2017-09-12
Added info on supported CNI and SDN via kubernetes

bfallonf:
cc @vikram-redhat @aheslin 

A few questions:
* I can get rid of the (tm)s right?
* The original information said this was in 3.5 and 3.6, but Vikram thinks it's just 3.6. Ali, can I get a  verification which version this should go in?
* I'll also take suggestions for better placement, if there is one.

Thanks


adellape:
@bfallonf Confirmed from PM this is applicable for 3.6. I'd recommend leaving the TMs in.

PR #5242, opened 2017-09-12
Updated language around StatefulSets as Technology Preview feature for 3.6

ahardin-rh:
@aheslin FYI


ahardin-rh:
No revision history needed

gaurav-nelson:
@ahardin-rh Since Dedicated is still 3.5 I suggest we remove the Dedicated label from this. @vikram-redhat can you confirm?

ahardin-rh:
@gaurav-nelson Thanks. I am told by the development team that we should not have any mention of TP features in the Dedicated docs at all. Removing the label for this. Then, I will create a follow-up PR to clean out any mentionings of TP features in the Dedicated docs.

PR #5082, opened 2017-08-25
Bug 1463852 - List images required to install Openshift disconnected environment 

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1463852

PR #5248, opened 2017-09-12
disconnected install updates

sosiouxme:
As I discovered while looking at bug https://bugzilla.redhat.com/show_bug.cgi?id=1480195#c12 this topic needed a few updates.

At this time there is no option to not install registry-console, and the install checks will look for its image to be there. So, just include it in the images needed for a disconnected install.

When doing a containerized install, the host components (master, node, ovs, etcd) need to be synchronized as well. This led to some confusing usage of the word "containerized" so that needed some clearing up too. I thought it best to only refer to those host components as "containerized", preferring to refer to the others as "components" or "images". We have host components, "on-cluster infrastructure components" (may be a bit too awkward), builder images, and logging/metrics images.

About builder images; for some reason, previous instructions indicated they were only needed on masters. AFAIK builders can run anywhere so I'm not sure why these were separated out, other than because they're all optional. So I fixed that too.

We could go into further nuances of where only a subset of the images are needed (e.g. infra nodes, master vs node components) but it seemed complicated enough already to me, and disk space is cheap. At least everything is covered; those who want to optimize can figure those things out.

All up for review of course. Questions?

### DIFF
--- a/install_config/install/disconnected_install.adoc
+++ b/install_config/install/disconnected_install.adoc
@@ -186,7 +186,20 @@ To sync the container images:
 $ systemctl start docker
 ----
 
-. Pull all of the required {product-title} containerized components.
+. If you are performing a xref:rpm_vs_containerized.adoc#install-config-install-rpm-vs-containerized[containerized install],
+pull all of the required {product-title} host component images.
+ifdef::openshift-enterprise[]
+Replace `<tag>` with `{latest-tag}` for the latest version.
+endif::[]
++
+----
+# docker pull registry.access.redhat.com/rhel7/etcd
+# docker pull registry.access.redhat.com/openshift3/ose:<tag>
+# docker pull registry.access.redhat.com/openshift3/node:<tag>
+# docker pull registry.access.redhat.com/openshift3/openvswitch:<tag>
+----
+
+. Pull all of the required {product-title} infrastructure component images.
 ifdef::openshift-enterprise[]
 Replace `<tag>` with `{latest-tag}` for the latest version.
 endif::[]
@@ -214,7 +227,7 @@ $ docker pull registry.access.redhat.com/openshift3/node:<tag>
 $ docker pull registry.access.redhat.com/openshift3/openvswitch:<tag>
 ----
 
-. Pull all of the required {product-title} containerized components for the
+. Pull all of the required {product-title} component images for the
 additional centralized log aggregation and metrics aggregation components.
 ifdef::openshift-enterprise[]
 Replace `<tag>` with `{latest-int-tag}` for the latest version.
@@ -316,7 +329,18 @@ $ mkdir </path/to/repos/images>
 $ cd </path/to/repos/images>
 ----
 
-. Export the {product-title} containerized components:
+. If you are performing a xref:rpm_vs_containerized.adoc#install-config-install-rpm-vs-containerized[containerized install],
+export the {product-title} host component images:
++
+----
+# docker save -o ose3-host-images.tar \
+    registry.access.redhat.com/rhel7/etcd \
+    registry.access.redhat.com/openshift3/ose \
+    registry.access.redhat.com/openshift3/node \
+    registry.access.redhat.com/openshift3/openvswitch
+----
+
+. Export the {product-title} infrastructure component images:
 +
 [source, bash]
 ----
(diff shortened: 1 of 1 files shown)


sosiouxme:
@adellape looks like you've worked on this section before.

install_config/install/disconnected_install.adoc
```diff
@@ -171,7 +171,20 @@ To sync the container images:
 # systemctl start docker
 ----
 
-. Pull all of the required {product-title} containerized components.
+. If you are performing a xref:rpm_vs_containerized.adoc[containerized install],
```
adellape: xref needs an explicit anchor for our Customer Portal builds. Top of page for this would be `#install-config-install-rpm-vs-containerized`.

install_config/install/disconnected_install.adoc
```diff
@@ -171,7 +171,20 @@ To sync the container images:
...
+endif::[]
++
+----
+# docker pull registry.access.redhat.com/rhel7/etcd
+# docker pull registry.access.redhat.com/openshift3/ose:<tag>
+# docker pull registry.access.redhat.com/openshift3/node:<tag>
+# docker pull registry.access.redhat.com/openshift3/openvswitch:<tag>
+----
+
+. Pull all of the required {product-title} on-cluster infrastructure component images.
```
adellape: Maybe drop "on-cluster" to just "infrastructure component images"?

install_config/install/disconnected_install.adoc
```diff
@@ -292,7 +297,18 @@ and then transporting them:
 # cd </path/to/repos/images>
 ----
 
-. Export the {product-title} containerized components:
+. If you are performing a xref:rpm_vs_containerized.adoc[containerized install],
```
adellape: Same re: anchor.

install_config/install/disconnected_install.adoc
```diff
@@ -292,7 +297,18 @@ and then transporting them:
...
++
+----
+# docker save -o ose3-host-images.tar \
+    registry.access.redhat.com/rhel7/etcd \
+    registry.access.redhat.com/openshift3/ose \
+    registry.access.redhat.com/openshift3/node \
+    registry.access.redhat.com/openshift3/openvswitch
+----
+
+. Export the {product-title} on-cluster infrastructure component images:
```
adellape: Same re: "on-cluster"?

adellape:
@sosiouxme FYI, some other changes were just merged in this file via https://github.com/openshift/openshift-docs/pull/5082.

install_config/install/disconnected_install.adoc
sosiouxme: Fine with me. I was trying to distinguish between things that run as OpenShift pods on top of a cluster versus those that run directly via docker/runc to bootstrap the cluster itself. I'll leave it to you to come up with something less confusing if you can :)

sosiouxme:
Thanks @adellape - updated and rebased.

sosiouxme:
rebased -- @adellape can this get pulled in before it needs another rebase? :)

kalexand-rh:
@ganhuang, does this change look ok to you? And should it be valid from 3.6 to 3.10? (That's what the bug leads me to believe.

@sosiouxme, do you want to rebase one more time, or do you want me to open a new PR on top of yours?

ganhuang:
@kalexand-rh From my understanding, 3.6-3.9 is fine to cherry pick.

A lot changes made in 3.10. We won't distinguish rpm vs containerized technically. And the image name changed as well (use `ose-control-plane`, `ose-node` instead).

In 3.10, all infra components tend to use container images except node service (could be installed via rpm on RHEL).

vikram-redhat:
Looks like we may have to do a fast turnaround of the disconnected install topic in 3.10 once 3.10 is released as we don't want to hold up the release to wait for these changes.

@kalexand-rh - can you please priortize this PR/changes to the 3.10 disconnected install to make sure disconnected install is updated for 3.10 changes?

vikram-redhat:
@kalexand-rh - and yeah, it might be better to close this PR and create a new one incorporating @sosiouxme's changes.

kalexand-rh:
I've opened two PRs based on this change
One for 3.6-3.9: https://github.com/openshift/openshift-docs/pull/11187
One for master and 3.10: https://github.com/openshift/openshift-docs/pull/11186

I'm closing this PR without merging.

PR #5258, opened 2017-09-13
advanced_install: installer as system container

sosiouxme:
Simplify slightly and also note some things that were not apparent.

I would like to go on to describe using the container from docker, but I thought it would be good to fix up what's there first.

### DIFF
--- a/install_config/install/advanced_install.adoc
+++ b/install_config/install/advanced_install.adoc
@@ -2328,7 +2328,7 @@ concern for the install restarting *docker* on the host.
 ----
 # atomic install --system \
     --storage=ostree \
-    --set INVENTORY_FILE=/path/to/inventory \//<1>
+    --set INVENTORY_FILE=/path/to/inventory \ <1>
 ifdef::openshift-enterprise[]
     registry.access.redhat.com/openshift3/ose-ansible:v3.7
 endif::[]
@@ -2338,11 +2338,12 @@ endif::[]
 ----
 <1> Specify the location on the local host for your inventory file.
 +
-This command initiates the cluster installation using the inventory file specified and the `root` user's
-SSH configuration, logging to the terminal as well as *_/var/log/ansible.log_*. The first time this is executed, the image is imported into
+This command initiates the cluster installation by using the inventory file specified and the `root` user's
+SSH configuration. It logs the output on the terminal and also saves it in the *_/var/log/ansible.log_* file.
+The first time this command is run, the image is imported into
 link:https://access.redhat.com/documentation/en-us/red_hat_satellite/6.2/html/content_management_guide/managing_ostree_content[OSTree]
-storage (system containers use this instead of *docker* daemon storage).
-Later executions re-use the stored image.
+storage (system containers use this rather than *docker* daemon storage).
+On subsequent runs, it reuses the stored image.
 +
 If for any reason the installation fails, before re-running the installer, see
 xref:installer-known-issues[Known Issues] to check for any specific instructions
@@ -2351,22 +2352,22 @@ or workarounds.
 [[running-the-advanced-installation-system-container-other-playbooks]]
 ==== Running Other Playbooks
 
-To run any other playbooks (for example, to run the
-xref:configuring-cluster-pre-install-checks[pre-install checks]
-before proceeding to install) using the containerized installer, you can
-specify a playbook with the `PLAYBOOK_FILE` environment variable. The
-default value is *_/usr/share/ansible/openshift-ansible/playbooks/byo/config.yml_*, which is the main cluster
-installation playbook, but you can set it to the path of another playbook
-inside the container.
+You can use the `PLAYBOOK_FILE` environment variable to specify other playbooks
+you want to run by using the containerized installer. The default value of the `PLAYBOOK_FILE` is
+*_/usr/share/ansible/openshift-ansible/playbooks/byo/config.yml_*, which is the
+main cluster installation playbook, but you can set it to the path of another
+playbook inside the container.
 
-For example:
+For example, to run the
+xref:configuring-cluster-pre-install-checks[pre-install checks] playbook before
+installation, use the following command:
 
 ----
 # atomic install --system \
     --storage=ostree \
     --set INVENTORY_FILE=/path/to/inventory \
-    --set PLAYBOOK_FILE=/usr/share/ansible/openshift-ansible/playbooks/byo/openshift-checks/pre-install.yml \//<1>
-    --set OPTS="-v" \//<2>
+    --set PLAYBOOK_FILE=/usr/share/ansible/openshift-ansible/playbooks/byo/openshift-checks/pre-install.yml \ <1>
+    --set OPTS="-v" \ <2>
 ifdef::openshift-enterprise[]
     registry.access.redhat.com/openshift3/ose-ansible:v3.7
 endif::[]
@@ -2386,9 +2387,8 @@ The installer image can also run as a *docker* container anywhere that *docker* 
 
 [WARNING]
 ====
-This method should not be used to run the installer on one of the hosts being configured,
-as the install typically restarts *docker* on the host, which would disrupt the installer
-container's own execution.
+This method must not be used to run the installer on one of the hosts being configured,
+as the install may restart *docker* on the host, disrupting the installer container execution.
 ====
 
 [NOTE]
@@ -2399,20 +2399,20 @@ run with different entry points and contexts, so runtime parameters are not the 
 
 At a minimum, when running the installer as a *docker* container you must provide:
 
-* SSH key(s) so that Ansible can reach your hosts.
+* SSH key(s), so that Ansible can reach your hosts.
 * An Ansible inventory file.
-* The location of an Ansible playbook to run against that inventory.
+* The location of the Ansible playbook to run against that inventory.
 
 Here is an example of how to run an install via *docker*.
 Note that this must be run by a non-`root` user with access to *docker*.
 
 ----
-$ docker run -t -u `id -u` \
-    -v $HOME/.ssh/id_rsa:/opt/app-root/src/.ssh/id_rsa:Z \
-    -v $HOME/ansible/hosts:/tmp/inventory:Z \
-    -e INVENTORY_FILE=/tmp/inventory \
-    -e PLAYBOOK_FILE=playbooks/byo/config.yml \
-    -e OPTS="-v" \
+$ docker run -t -u `id -u` \ <1>
+    -v $HOME/.ssh/id_rsa:/opt/app-root/src/.ssh/id_rsa:Z \ <2>
+    -v $HOME/ansible/hosts:/tmp/inventory:Z \ <3>
+    -e INVENTORY_FILE=/tmp/inventory \ <3>
+    -e PLAYBOOK_FILE=playbooks/byo/config.yml \ <4>
+    -e OPTS="-v" \ <5>
 ifdef::openshift-enterprise[]
     registry.access.redhat.com/openshift3/ose-ansible:v3.7
 endif::[]
@@ -2420,14 +2420,10 @@ ifdef::openshift-origin[]
     docker.io/openshift/origin-ansible:v3.7
 endif::[]
 ----
-
-The following is a detailed explanation of the options used in the command above.
-
-* `-u `id -u`` makes the container run with the same UID as the current
+<1> `-u `id -u`` makes the container run with the same UID as the current
 user, which allows that user to use the SSH key inside the container (SSH
 private keys are expected to be readable only by their owner).
-
-* `-v $HOME/.ssh/id_rsa:/opt/app-root/src/.ssh/id_rsa:Z` mounts your
+<2> `-v $HOME/.ssh/id_rsa:/opt/app-root/src/.ssh/id_rsa:Z` mounts your
 SSH key (`$HOME/.ssh/id_rsa`) under the container user's `$HOME/.ssh`
 (*_/opt/app-root/src_* is the `$HOME` of the user in the container). If
 you mount the SSH key into a non-standard location you can add an
@@ -2435,24 +2431,23 @@ environment variable with `-e ANSIBLE_PRIVATE_KEY_FILE=/the/mount/point`
 or set `ansible_ssh_private_key_file=/the/mount/point` as a variable in
 the inventory to point Ansible at it.
 +
-Note that the SSH key is mounted with the `:Z` flag: this is
+Note that the SSH key is mounted with the `:Z` flag. This is
 required so that the container can read the SSH key under
-its restricted SELinux context; this also means that your
+its restricted SELinux context. This also means that your
 original SSH key file will be re-labeled to something like
 `system_u:object_r:container_file_t:s0:c113,c247`. For more details
 about `:Z` please check the `docker-run(1)` man page. Keep this in mind
-when providing these volume mount specifications because this could
+when providing these volume mount specifications because this might
 have unexpected consequences: for example, if you mount (and therefore
 re-label) your whole `$HOME/.ssh` directory it will block the host's
 *sshd* from accessing your public keys to login. For this reason you
 may want to use a separate copy of the SSH key (or directory), so that
 the original file labels remain untouched.
-
-* `-v $HOME/ansible/hosts:/tmp/inventory:Z` and `-e INVENTORY_FILE=/tmp/inventory`
+<3> `-v $HOME/ansible/hosts:/tmp/inventory:Z` and `-e INVENTORY_FILE=/tmp/inventory`
 mount a static Ansible inventory file into the container as
 *_/tmp/inventory_* and set the corresponding environment variable to
 point at it. As with the SSH key, the inventory file SELinux labels may
-need to be relabeled via the `:Z` flag to allow reading in the container,
+need to be relabeled by using the `:Z` flag to allow reading in the container,
 depending on the existing label (for files in a user `$HOME` directory
 this is likely to be needed). So again you may prefer to copy the
 inventory to a dedicated location before mounting it.
@@ -2461,14 +2456,12 @@ The inventory file can also be downloaded from a web server if you specify
 the `INVENTORY_URL` environment variable, or generated dynamically using
 `DYNAMIC_SCRIPT_URL` to specify an executable script that provides a
 dynamic inventory.
-
-* `-e PLAYBOOK_FILE=playbooks/byo/config.yml` specifies the playbook
+<4> `-e PLAYBOOK_FILE=playbooks/byo/config.yml` specifies the playbook
 to run (in this example, the BYO installer) as a relative path from the
 top level directory of *openshift-ansible* content. The full path from the
 RPM can also be used, as well as the path to any other playbook file in
 the container.
-
-* `-e OPTS="-v"` supplies arbitrary command line options (in this case,
+<5> `-e OPTS="-v"` supplies arbitrary command line options (in this case,
 `-v` to increase verbosity) to the `ansible-playbook` command that runs
 inside the container.
 
--- a/install_config/install/advanced_install.adoc
+++ b/install_config/install/advanced_install.adoc
@@ -2296,10 +2296,8 @@ If for any reason the installation fails, before re-running the installer, see
 xref:installer-known-issues[Known Issues] to check for any specific
 instructions or workarounds.
 
-[[running-the-advanced-installation-system-container]]
+[[running-the-advanced-installation-containerized]]
 === Running the Containerized Installer
-
-include::install_config/install/advanced_install.adoc[tag=syscontainers_techpreview]
 
 The
 ifdef::openshift-enterprise[]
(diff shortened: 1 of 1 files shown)


install_config/install/advanced_install.adoc
```diff
@@ -2171,9 +2171,10 @@ ifdef::openshift-origin[]
 endif::[]
 image is a containerized version of the {product-title} installer that runs as a
 link:https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux_atomic_host/7/html/managing_containers/running_system_containers[system container]. System containers are stored and run outside of the traditional
-*docker* service. Functionally, using the containerized installer is the same as
-using the traditional RPM-based installer, except it is running in a
-containerized environment instead of directly on the host.
+*docker* service. The installer image provides the same functionality
+as the traditional RPM-based installer, but it runs in a containerized
+environment that provides all of its dependencies rather than being
+installed directly on the host.
```
sosiouxme: This is the only real advantage -- don't have to install a bunch of RPMs on your host. So I think it's worth mentioning.

install_config/install/advanced_install.adoc
```diff
@@ -2186,25 +2187,14 @@ $ docker pull docker.io/openshift/origin-ansible:v3.6
...
+. To use the installer image as a system container, it must be stored in
 link:https://access.redhat.com/documentation/en-us/red_hat_satellite/6.2/html/content_management_guide/managing_ostree_content[OSTree]
-instead of defaulting to *docker* daemon storage. Use the Atomic CLI to import
-the installer image from the local *docker* engine to OSTree storage:
+instead of *docker* daemon storage. As the root user, use the Atomic
+CLI to install the installer image from the local *docker* engine as a
+systemd service using OSTree storage:
 +
 ----
-$ atomic pull --storage ostree \
```
sosiouxme: It turns out this whole step is unnecessary as `atomic install` does it for you.

install_config/install/advanced_install.adoc
```diff
@@ -2216,22 +2206,35 @@ ifdef::openshift-origin[]
 endif::[]
 ----
 <1> Sets the name for the systemd service.
-<2> Specify the location for your inventory file on your local workstation.
+<2> Specify the location for your local inventory file.
```
sosiouxme: This is the only place we referred to local workstation... this can run anywhere that docker and systemd run, not that likely to be a workstation actually.

install_config/install/advanced_install.adoc
```diff
@@ -2216,22 +2206,35 @@ ifdef::openshift-origin[]
...
 
 . Use the `systemctl` command to start the installer service as you would any
-other systemd service. This command initiates the cluster installation:
+other systemd service. This command initiates the cluster installation using
+the inventory file specified above and the root user's SSH configuration:
```
sosiouxme: This is one thing that was not apparent before... ssh config comes from root for this use case (in the docker use case you have to mount it in).

install_config/install/advanced_install.adoc
```diff
@@ -2186,25 +2187,14 @@ $ docker pull docker.io/openshift/origin-ansible:v3.6
 endif::[]
 ----
 
-. The installer system container must be stored in
+. To use the installer image as a system container, it must be stored in
 link:https://access.redhat.com/documentation/en-us/red_hat_satellite/6.2/html/content_management_guide/managing_ostree_content[OSTree]
-instead of defaulting to *docker* daemon storage. Use the Atomic CLI to import
-the installer image from the local *docker* engine to OSTree storage:
+instead of *docker* daemon storage. As the root user, use the Atomic
```
sosiouxme: atomic install and systemctl and journalctl are generally things you have to run as root. Is it customary to use the root prompt in docs?

install_config/install/advanced_install.adoc
```diff
@@ -2216,22 +2206,35 @@ ifdef::openshift-origin[]
...
 . Use the `systemctl` command to start the installer service as you would any
-other systemd service. This command initiates the cluster installation:
+other systemd service. This command initiates the cluster installation using
+the inventory file specified above and the root user's SSH configuration:
 +
 ----
 $ systemctl start openshift-installer
 ----
 +
+There is no output from this command while the installer service is
```
sosiouxme: One potentially confusing thing the first time you run this.

install_config/install/advanced_install.adoc
```diff
@@ -2240,21 +2243,22 @@ $ atomic uninstall openshift-installer
...
 ==== Running Other Playbooks
 
-After you have completed the cluster installation, if you want to later run any
-other playbooks using the containerized installer (for example, cluster upgrade
-playbooks), you can use the `PLAYBOOK_FILE` environment variable. The default
-value is `playbooks/byo/config.yml`, which is the main cluster installation
-playbook, but you can set it to the path of another playbook inside the
-container.
+To run any other playbooks (for example, to run the
+xref:configuring-cluster-pre-install-checks[pre-install checks]
```
sosiouxme: The pre-install checks seemed a more relevant subject for this topic than the upgrade playbooks. Plus I'd really like to let people know they can run checks from this container.

install_config/install/advanced_install.adoc
```diff
@@ -2216,45 +2206,59 @@ ifdef::openshift-origin[]
...
 
 ----
-$ atomic install --system \
+# atomic install --system \
     --storage=ostree \
     --name=openshift-installer \
-    --set INVENTORY_FILE=/etc/ansible/hosts \
-    --set PLAYBOOK_FILE=playbooks/byo/openshift-cluster/upgrades/v3_6/upgrade.yml \//<1>
+    --set INVENTORY_FILE=/path/to/inventory \
+    --set PLAYBOOK_FILE=playbooks/byo/openshift-checks/pre-install.yml \//<1>
```
sosiouxme: actually I have to update this - currently it won't work with a relative path.

install_config/install/advanced_install.adoc
```diff
@@ -2215,57 +2205,72 @@ ifdef::openshift-origin[]
...
 
 ----
-$ atomic install --system \
+# atomic install --system \
     --storage=ostree \
     --name=openshift-installer \
-    --set INVENTORY_FILE=/etc/ansible/hosts \
-    --set PLAYBOOK_FILE=playbooks/byo/openshift-cluster/upgrades/v3_6/upgrade.yml \//<1>
+    --set INVENTORY_FILE=/path/to/inventory \
+    --set PLAYBOOK_FILE=/usr/share/ansible/openshift-ansible/playbooks/byo/openshift-checks/pre-install.yml \//<1>
```
sosiouxme: As currently implemented, you actually need the full path for the playbook. This will change "soon" with updates to the image and we'll be able to go back to relative paths (the full path should always work though).

sosiouxme:
@openshift/team-documentation these are changes for 3.6 and origin ... hopefully by the time 3.7 rolls around we will have an updated atomic with a better workflow and no longer tech preview.

install_config/install/advanced_install.adoc
```diff
@@ -2170,102 +2170,71 @@ ifdef::openshift-origin[]
...
-+
-----
-$ atomic install --system \
+# atomic install --system \
     --storage=ostree \
     --name=openshift-installer \//<1>
     --set INVENTORY_FILE=/path/to/inventory \//<2>
 ifdef::openshift-enterprise[]
-    docker:registry.access.redhat.com/openshift3/ose-ansible:v3.6
+    registry.access.redhat.com/openshift3/ose-ansible:v3.6
```
sosiouxme: @ashcrow @giuseppe check me on this -- it looks like the latest version of atomic on RHEL now has more of the functionality you wanted here.

Previously, it needed to pull through docker, and it would import into ostree every time, and created a regular systemd service (not one-time, and no output, just journald, and service definition was retained after running).

That is still the behavior AFAICS if I specify the image with the `docker:` prefix, however if I leave that off, then it directly imports the image to ostree (no docker even running), creates a one-time service, runs it, and uninstalls it, which I think is the behavior you were wanting to get to.

Just wanted you to check if I'm missing anything here.

install_config/install/advanced_install.adoc
```diff
@@ -2170,102 +2170,71 @@ ifdef::openshift-origin[]
...
-. The installer system container must be stored in
-link:https://access.redhat.com/documentation/en-us/red_hat_satellite/6.2/html/content_management_guide/managing_ostree_content[OSTree]
-instead of defaulting to *docker* daemon storage. Use the Atomic CLI to import
-the installer image from the local *docker* engine to OSTree storage:
+link:https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux_atomic_host/7/html/managing_containers/running_system_containers[system container].
+System containers are stored and run outside of the traditional
+*docker* service. The installer image provides the same functionality
+as the traditional RPM-based installer, but it runs in a containerized
+environment that provides all of its dependencies rather than being
+installed directly on the host.
```
sosiouxme: This is the main advantage -- don't have to install a bunch of RPMs on your host. So I think it's worth mentioning.

install_config/install/advanced_install.adoc
```diff
@@ -2170,102 +2170,71 @@ ifdef::openshift-origin[]
...
-ifdef::openshift-origin[]
-    docker:docker.io/openshift/origin-ansible:v3.6
-endif::[]
-----
-
-. Install the system container so it is set up as a systemd service:
-+
-----
-$ atomic install --system \
+# atomic install --system \
```
sosiouxme: atomic install I think you have to run as root. Is it customary to use the root prompt in docs?

install_config/install/advanced_install.adoc
```diff
@@ -2170,102 +2170,71 @@ ifdef::openshift-origin[]
...
-. Use the `systemctl` command to start the installer service as you would any
-other systemd service. This command initiates the cluster installation:
+<1> Specify the name for the temporary systemd service; can be any name not in use.
+<2> Specify the location on the local host for your inventory file.
 +
-----
-$ systemctl start openshift-installer
-----
+This command initiates the cluster installation using the inventory file specified and the root user's
+SSH configuration, logging to the terminal as well as *_/var/log/ansible.log_*. The first time this is executed, the image will be imported into
```
sosiouxme: Things that were not apparent before... ssh config comes from root for this use case, and where to find logs.

sosiouxme (commented):
Updated per things I discovered about the `atomic install` command just now.

install_config/install/advanced_install.adoc
ashcrow: Agreed

install_config/install/advanced_install.adoc
ashcrow: Correct. It must be run as root.

install_config/install/advanced_install.adoc
ashcrow: > [...]if I leave that off, then it directly imports the image to ostree (no docker even running), creates a one-time service, runs it, and uninstalls it, which I think is the behavior you were wanting to get to.

That is the ``atomic.run`` ``once`` feature that is much cleaner than the original one. Looks like RHEL 7.4 has atomic 1.18.1-3 which has the run once ability. I'm not sure if the docker import is still needed or not with that version. @giuseppe?

ashcrow (commented):
Looks great!

ashcrow:
@sosiouxme thank you for working on this!

install_config/install/advanced_install.adoc
sosiouxme: It looks like with that feature, it doesn't even use the service name provided. I don't think it ever creates a systemd service per se, doesn't seem to send anything to journald. So I'll take out all the discussion of systemd service, which should make this even simpler :)

install_config/install/advanced_install.adoc
giuseppe: yes, it must run as root. "atomic install --system" would fail anyway if not running as root.

sosiouxme:
I added a commit to also discuss running the installer image via Docker. This can be worked on separately if preferred, though they are closely related.

sosiouxme:
@openshift/team-documentation can this PR get some love?

vikram-redhat:
@sosiouxme is this for 3.7?

sosiouxme:
@vikram-redhat it's not specific to 3.7, it's been relevant since it was written two months ago.

install_config/install/advanced_install.adoc
```diff
@@ -2157,115 +2157,181 @@ If for any reason the installation fails, before re-running the installer, see
...
-image is a containerized version of the {product-title} installer that runs as a
-link:https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux_atomic_host/7/html/managing_containers/running_system_containers[system container]. System containers are stored and run outside of the traditional
-*docker* service. Functionally, using the containerized installer is the same as
-using the traditional RPM-based installer, except it is running in a
-containerized environment instead of directly on the host.
+image is a containerized version of the {product-title} installer.
+This installer image provides the same functionality as the RPM-based
+installer, but it runs in a containerized environment that provides all
+of its dependencies rather than being installed directly on the host.
+So, the only requirement to use it is the ability to run a container.
```
nengard: I'd remove the 'So,'

ncbaratta (commented):
LGTM

giuseppe:
@sosiouxme could you please rebase this PR?  It looks like this change didn't get it, and the documentation still reports the old way of using the system container installer.

ashcrow:
👍 we've had some confusion over usage and I think this PR clears things up.

sosiouxme:
@giuseppe @ashcrow sure thing guys.
@vikram-redhat can this get moving now?

vikram-redhat:
@sosiouxme sorry for the delay. @gaurav-nelson - can you take a look at this today? Also, we need an ack from QE to move this along. @xltian - can someone from QE review this and provide approval?

install_config/install/advanced_install.adoc
```diff
@@ -2296,115 +2296,181 @@ If for any reason the installation fails, before re-running the installer, see
...
-. Install the system container so it is set up as a systemd service:
+. As the `root` user, use the Atomic CLI to run the installer as a run-once system container:
 +
 ----
-$ atomic install --system \
+# atomic install --system \
     --storage=ostree \
-    --name=openshift-installer \//<1>
-    --set INVENTORY_FILE=/path/to/inventory \//<2>
+    --set INVENTORY_FILE=/path/to/inventory \//<1>
```
gaurav-nelson: Please change `--set INVENTORY_FILE=/path/to/inventory \//<1>` to `--set INVENTORY_FILE=/path/to/inventory \ <1>`

- remove the `//` before `<1>` 
- multiple instances

install_config/install/advanced_install.adoc
```diff
@@ -2296,115 +2296,181 @@ If for any reason the installation fails, before re-running the installer, see
...
-<2> Specify the location for your inventory file on your local workstation.
-
-. Use the `systemctl` command to start the installer service as you would any
-other systemd service. This command initiates the cluster installation:
+<1> Specify the location on the local host for your inventory file.
 +
-----
-$ systemctl start openshift-installer
-----
+This command initiates the cluster installation using the inventory file specified and the `root` user's
```
gaurav-nelson: s/the cluster installation using the/the cluster installation **by** using the

install_config/install/advanced_install.adoc
```diff
@@ -2296,115 +2296,181 @@ If for any reason the installation fails, before re-running the installer, see
...
+SSH configuration, logging to the terminal as well as *_/var/log/ansible.log_*. The first time this is executed, the image is imported into
```
gaurav-nelson: - s/SSH configuration,/SSH configuration**.**
 - s/logging to the terminal as well as *_/var/log/ansible.log_*./It logs the output on the terminal and also saves it in the *_/var/log/ansible.log_* file.

install_config/install/advanced_install.adoc
```diff
@@ -2296,115 +2296,181 @@ If for any reason the installation fails, before re-running the installer, see
...
+link:https://access.redhat.com/documentation/en-us/red_hat_satellite/6.2/html/content_management_guide/managing_ostree_content[OSTree]
+storage (system containers use this instead of *docker* daemon storage).
+Later executions re-use the stored image.
```
gaurav-nelson: s/Later executions re-use the stored image./On subsequent runs, it reuses the stored image.

xltian: @jianlinliu @ganhuang  Please have someone from QE to review this doc .

install_config/install/advanced_install.adoc
```diff
@@ -2296,115 +2296,181 @@ If for any reason the installation fails, before re-running the installer, see
...
-other systemd service. This command initiates the cluster installation:
+<1> Specify the location on the local host for your inventory file.
 +
-----
-$ systemctl start openshift-installer
-----
+This command initiates the cluster installation using the inventory file specified and the `root` user's
+SSH configuration, logging to the terminal as well as *_/var/log/ansible.log_*. The first time this is executed, the image is imported into
+link:https://access.redhat.com/documentation/en-us/red_hat_satellite/6.2/html/content_management_guide/managing_ostree_content[OSTree]
+storage (system containers use this instead of *docker* daemon storage).
```
gaurav-nelson: s/instead of/rather than

install_config/install/advanced_install.adoc
```diff
@@ -2296,115 +2296,181 @@ If for any reason the installation fails, before re-running the installer, see
...
 
-After you have completed the cluster installation, if you want to later run any
-other playbooks using the containerized installer (for example, cluster upgrade
-playbooks), you can use the `PLAYBOOK_FILE` environment variable. The default
-value is `playbooks/byo/config.yml`, which is the main cluster installation
-playbook, but you can set it to the path of another playbook inside the
-container.
+To run any other playbooks (for example, to run the
+xref:configuring-cluster-pre-install-checks[pre-install checks]
+before proceeding to install) using the containerized installer, you can
```
gaurav-nelson: s/using/**by** using

install_config/install/advanced_install.adoc
```diff
@@ -2296,115 +2296,181 @@ If for any reason the installation fails, before re-running the installer, see
...
 [[running-the-advanced-installation-system-container-other-playbooks]]
 ==== Running Other Playbooks
 
-After you have completed the cluster installation, if you want to later run any
-other playbooks using the containerized installer (for example, cluster upgrade
-playbooks), you can use the `PLAYBOOK_FILE` environment variable. The default
-value is `playbooks/byo/config.yml`, which is the main cluster installation
-playbook, but you can set it to the path of another playbook inside the
-container.
+To run any other playbooks (for example, to run the
```
gaurav-nelson: s/run any other/run other

install_config/install/advanced_install.adoc
gaurav-nelson: s/before proceeding/playbook before proceeding

install_config/install/advanced_install.adoc
gaurav-nelson: From lines 2354 till 2362, replace with:
```
You can use the `PLAYBOOK_FILE` environment variable to specify other playbooks
you want to run by using the containerized installer. The default value of the `PLAYBOOK_FILE` is
*_/usr/share/ansible/openshift-ansible/playbooks/byo/config.yml_*, which is the
main cluster installation playbook, but you can set it to the path of another
playbook inside the container.

For example, to run the
xref:configuring-cluster-pre-install-checks[pre-install checks] playbook before
installation. Use the following command:
```

install_config/install/advanced_install.adoc
```diff
@@ -2296,115 +2296,181 @@ If for any reason the installation fails, before re-running the installer, see
...
 
 ----
-$ atomic install --system \
+# atomic install --system \
     --storage=ostree \
-    --name=openshift-installer \
-    --set INVENTORY_FILE=/etc/ansible/hosts \
-    --set PLAYBOOK_FILE=playbooks/byo/openshift-cluster/upgrades/v3_7/upgrade.yml \//<1>
+    --set INVENTORY_FILE=/path/to/inventory \
+    --set PLAYBOOK_FILE=/usr/share/ansible/openshift-ansible/playbooks/byo/openshift-checks/pre-install.yml \//<1>
```
gaurav-nelson: s///<1>/<1> please remove `//`

I see that these were already there in the original file, but with the recently updated guideline we use <1> directly for callouts

install_config/install/advanced_install.adoc
```diff
@@ -2296,115 +2296,181 @@ If for any reason the installation fails, before re-running the installer, see
...
+    --set OPTS="-v" \//<2>
```
gaurav-nelson: s///<2>/ <2>

install_config/install/advanced_install.adoc
```diff
@@ -2296,115 +2296,181 @@ If for any reason the installation fails, before re-running the installer, see
...
+<2> Set `OPTS` to add command line options to `ansible-playbook`.
+
+[[running-the-advanced-installation-docker]]
+==== Running the Installer as a Docker Container
+
+The installer image can also run as a *docker* container anywhere that *docker* can run.
+
+[WARNING]
+====
+This method should not be used to run the installer on one of the hosts being configured,
```
gaurav-nelson: /s/should/must

install_config/install/advanced_install.adoc
```diff
@@ -2296,115 +2296,181 @@ If for any reason the installation fails, before re-running the installer, see
...
+as the install typically restarts *docker* on the host, which would disrupt the installer
```
gaurav-nelson: s/as/because

install_config/install/advanced_install.adoc
```diff
@@ -2296,115 +2296,181 @@ If for any reason the installation fails, before re-running the installer, see
...
+container's own execution.
```
gaurav-nelson: maybe s/container's own execution/container's execution.

install_config/install/advanced_install.adoc
```diff
@@ -2296,115 +2296,181 @@ If for any reason the installation fails, before re-running the installer, see
...
+====
+
+[NOTE]
+====
+Although this method and the system container method above use the same image, they
+run with different entry points and contexts, so runtime parameters are not the same.
+====
+
+At a minimum, when running the installer as a *docker* container you must provide:
```
gaurav-nelson: s/At a minimum, when running/When running

install_config/install/advanced_install.adoc
```diff
@@ -2296,115 +2296,181 @@ If for any reason the installation fails, before re-running the installer, see
...
+
+* SSH key(s) so that Ansible can reach your hosts.
```
gaurav-nelson: s/SSH key(s) so/SSH key(s)**,**

install_config/install/advanced_install.adoc
```diff
@@ -2296,115 +2296,181 @@ If for any reason the installation fails, before re-running the installer, see
...
+* An Ansible inventory file.
+* The location of an Ansible playbook to run against that inventory.
```
gaurav-nelson: s/an/the

install_config/install/advanced_install.adoc
```diff
@@ -2296,115 +2296,181 @@ If for any reason the installation fails, before re-running the installer, see
...
+
+Here is an example of how to run an install via *docker*.
+Note that this must be run by a non-`root` user with access to *docker*.
+
+----
+$ docker run -t -u `id -u` \
```
gaurav-nelson: Please use callous for explaining this code instead of the list. For example:
```
$ docker run -t -u `id -u` \ <1>
    -v $HOME/.ssh/id_rsa:/opt/app-root/src/.ssh/id_rsa:Z \ <2> ...and so on.
```
and after the code block use the callouts as:
```
<1> `-u `id -u`` makes the container run with the same UID as the current
user, which allows that user to use the SSH key inside the container (SSH
private keys are expected to be readable only by their owner).

<2> `-v $HOME/.ssh/id_rsa:/opt/app-root/src/.ssh/id_rsa:Z` mounts your
SSH key (`$HOME/.ssh/id_rsa`) under the container user's `$HOME/.ssh`
(*_/opt/app-root/src_* is the `$HOME` of the user in the container). If
you mount the SSH key into a non-standard location you can add an
environment variable with `-e ANSIBLE_PRIVATE_KEY_FILE=/the/mount/point`
or set `ansible_ssh_private_key_file=/the/mount/point` as a variable in
the inventory to point Ansible at it.

... and so on.
```

install_config/install/advanced_install.adoc
```diff
@@ -2296,115 +2296,181 @@ If for any reason the installation fails, before re-running the installer, see
...
+
+* `-v $HOME/.ssh/id_rsa:/opt/app-root/src/.ssh/id_rsa:Z` mounts your
+SSH key (`$HOME/.ssh/id_rsa`) under the container user's `$HOME/.ssh`
+(*_/opt/app-root/src_* is the `$HOME` of the user in the container). If
+you mount the SSH key into a non-standard location you can add an
+environment variable with `-e ANSIBLE_PRIVATE_KEY_FILE=/the/mount/point`
+or set `ansible_ssh_private_key_file=/the/mount/point` as a variable in
+the inventory to point Ansible at it.
++
+Note that the SSH key is mounted with the `:Z` flag: this is
```
gaurav-nelson: s/flag:/flag**.**

install_config/install/advanced_install.adoc
```diff
@@ -2296,115 +2296,181 @@ If for any reason the installation fails, before re-running the installer, see
...
+required so that the container can read the SSH key under
+its restricted SELinux context; this also means that your
```
gaurav-nelson: s/SELinux context;/SELinux context**.**

install_config/install/advanced_install.adoc
```diff
@@ -2296,115 +2296,181 @@ If for any reason the installation fails, before re-running the installer, see
...
+original SSH key file will be re-labeled to something like
+`system_u:object_r:container_file_t:s0:c113,c247`. For more details
+about `:Z` please check the `docker-run(1)` man page. Keep this in mind
+when providing these volume mount specifications because this could
```
gaurav-nelson: s/this could/this might

install_config/install/advanced_install.adoc
```diff
@@ -2296,115 +2296,181 @@ If for any reason the installation fails, before re-running the installer, see
...
+re-label) your whole `$HOME/.ssh` directory it will block the host's
+*sshd* from accessing your public keys to login. For this reason you
+may want to use a separate copy of the SSH key (or directory), so that
+the original file labels remain untouched.
+
+* `-v $HOME/ansible/hosts:/tmp/inventory:Z` and `-e INVENTORY_FILE=/tmp/inventory`
+mount a static Ansible inventory file into the container as
+*_/tmp/inventory_* and set the corresponding environment variable to
+point at it. As with the SSH key, the inventory file SELinux labels may
+need to be relabeled via the `:Z` flag to allow reading in the container,
```
gaurav-nelson: s/via/by using

gaurav-nelson:
@sosiouxme Thank you for the PR. I have suggested few updates, not big ones but multiple small ones. 

Then I can merge this one after QE checks.

ganhuang:
cc @wkshi

ashcrow:
@gaurav-nelson @sosiouxme @wkshi thanks guys for working on this!

sosiouxme:
Thanks, I'll fix those up today.

install_config/install/advanced_install.adoc
sosiouxme: This results in some really huge callouts, but I guess that's fine. I think originally there wasn't as clear a correspondence between the lines of command and the list explanation. Now the only real mismatch is that the third callout is about two lines, but I guess that's fine too.

install_config/install/advanced_install.adoc
sosiouxme: Is there a better way to imply that this list is not comprehensive, just sufficient for a minimal use case?

install_config/install/advanced_install.adoc
sosiouxme: It might or might not restart docker depending on what you're doing. There's no reason to get into when it will or won't, but it's also wrong to assert that it always does. How about "may" restart?

install_config/install/advanced_install.adoc
sosiouxme: QE review is a great idea :)

Keep in mind the present PR is directed at v3.6/v3.7 -- playbooks at least will be different for 3.9

sosiouxme:
@gaurav-nelson thanks for the review; I think I've addressed everything except https://github.com/openshift/openshift-docs/pull/5258#pullrequestreview-88353914 where I have a lingering question.

install_config/install/advanced_install.adoc
sosiouxme: Or, I should say: required in every use case, but not sufficient in most.

gaurav-nelson:
Thank you @sosiouxme for the updates, I am waiting for QE review before merging this. After that's done, I will do a follow up PR and then we can work on better explanation for your question.  Thanks again!

wkshi (commented):
Looks good to me.(QE side)

ashcrow:
@gaurav-nelson looks like we got QE approval 🎆 🎆

sosiouxme:
:clap:

adellape:
@gaurav-nelson I'm getting a lot of conflicts trying to pick something else from master to 3.9-stage, and I think it's because this PR should have gotten picked to 3.9 when it was picked to 3.6 and 3.7. I'm gonna try doing that.

PR #5262, opened 2017-09-14
Fixed bad link in templates.adoc

mburke5678:
The _as seen in this example_ link in the _**Uploading a Template**_ section of the **Dev Guide -> Templates** topic goes to [**Core Concerts -> Templates**](https://docs.openshift.com/container-platform/3.4/architecture/core_concepts/templates.html#architecture-core-concepts-templates) which no longer has a template example after a [re-org](https://github.com/openshift/openshift-docs/pull/3334). 
The example was moved to the **Dev Guide -> Templates** topic. I updated the link in the **Dev Guide** topic. 

Also removed the link from _A [template](https://docs.openshift.com/container-platform/3.4/architecture/core_concepts/templates.html#architecture-core-concepts-templates) describes_ as the info on the linked page is limited. No real benefit to linking to that page.


mburke5678:
@ahardin-rh Can you take a look? You reviewed the change where the example was moved. Should we have the example file in both places? Other topics in **Core Concepts** have example YAML.

mburke5678:
@ahardin-rh Bump!

ahardin-rh:
@mburke5678 Sorry for the delay! 😱 This is an interesting one. There's really not a lot of content anymore in Core Concepts -> Templates, but the page still exists so that users can get a definition of what templates are and get directed to the procedure-oriented docs in Developer Guide -> Templates. I would not duplicate the information in 2 places. I think the structure here is still sound, as the Dev Guide topic focuses more on procedure (as it should) and the examples in that topic show examples as it relates to the procedure.
@aheslin Any thoughts here?

mburke5678:
@aheslin  PTAL ^^

mburke5678:
@aheslin Should we add an example template file to [Core Concerts -> Templates](https://docs.openshift.com/container-platform/3.4/architecture/core_concepts/templates.html#architecture-core-concepts-templates)?  Other top-level topics in Core Concepts have example YAML.

aheslin:
It looks like templates are defined in two places right now, core concepts and in the templates section of the Dev Guide. With the way that I've recently heard about how users search for content, I'd get rid of the templates concept in the Arch Guide. I don't know that I'd bother adding a yaml example there.

mburke5678:
@aheslin PTAL

mburke5678:
@nhr Harrison -- Can you set up a redirect from the `architecture/core_concepts/templates.html` topic to `dev_guide/templates.html`. We are removing the Architecture Guide -> Core Concepts -> Templates topic. 
What is the process for doing a redirect? Do you need more info from me?

nhr:
@mburke5678 as of AsciiBinder 0.1.13, you can actually do this yourself in the `_topic_map.yml` file:

Instead of removing the 'templates' entry from the topic map, add an additional parameter:

`Alias: dev_guide/templates`

When AsciiBinder runs it will create a stub file at `/architecture/core_concepts/templates.html` that automatically redirects readers to `/dev_guide/templates.html`.

Jenkins has already been upgraded to AsciiBinder 0.1.13 to support this behavior in production.

PR #5270, opened 2017-09-14
Added information on upgrading to Pro

ahardin-rh:
This should land after October 17.

@tiwillia What page should users go to once they go to manage.openshift.com?

### DIFF
--- a/getting_started/choose_a_plan.adoc
+++ b/getting_started/choose_a_plan.adoc
@@ -47,8 +47,10 @@ image::online_pro_manage_plan.png[manage your plan]
 [[getting-started-upgrading-plan]]
 == Upgrading Your Plan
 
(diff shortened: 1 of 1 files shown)


ahardin-rh:
@tiwillia PTAL - question above

tiwillia:
@ahardin-rh If you go to manage.openshift.com now, you'll hit the Plan Selection page (/register/plan). There, you can select starter or pro plans. This will look exactly the same when manage.openshift.com is updated with pro-tier support, except the button for the 'pro' plan will direct the user through the registration process on the manage.openshift.com application instead of the account.openshift.com application. Users will not see much of a difference at all.

I don't think this PR needs any changes.

ahardin-rh:
No revision history needed

PR #5232, opened 2017-09-11
clarify secret injector behaviour with alternate git URI style

jim-minter:
Hopefully this change is a clarification.  It should go into all branches which have this doc section.

PR #5275, opened 2017-09-15
PR5232-Follow-up Minor Formatting Edits

bmcelvee:
Follow-up to https://github.com/openshift/openshift-docs/pull/5232.


vikram-redhat:
@jim-minter @bmcelvee - I want to confirm that this change belongs in 3.4? This section was only introduced from 3.5 onwards in the docs. 

@bmcelvee - I am not adding this in as part of our weekly release. Depending on the answer to my question above, can you do a follow-up directly into the 3.4 branch if it does indeed need adding in?

bmcelvee:
@vikram-redhat You are correct. I removed the 3.4 label from this PR and the original. Thanks!

PR #5303, opened 2017-09-20
The osm_cluster_network_cidr, openshift_portal_net, and osm_host_subnet_length no longer set by default.

adelton:
This is related to changes from https://bugzilla.redhat.com/show_bug.cgi?id=1451023 and https://github.com/openshift/openshift-ansible/pull/5430.

### DIFF
--- a/install_config/install/advanced_install.adoc
+++ b/install_config/install/advanced_install.adoc
@@ -386,7 +386,7 @@ use the multitenant SDN plug-in.
 |This variable overrides the SDN cluster network CIDR block. This is the network
 from which pod IPs are assigned. This network block should be a private block
 and must not conflict with existing network blocks in your infrastructure to
-which pods, nodes, or the master may require access. Defaults to `10.128.0.0/14`
+which pods, nodes, or the master may require access. Reasonable value is `10.128.0.0/14`
 and cannot be arbitrarily re-configured after deployment, although certain
 changes to it can be made in the
 xref:../../install_config/configuring_sdn.adoc#configuring-the-pod-network-on-masters[SDN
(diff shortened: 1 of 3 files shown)


install_config/install/advanced_install.adoc
```diff
@@ -253,8 +253,8 @@ will be created within the
 xref:../../architecture/networking/sdn.adoc#architecture-additional-concepts-sdn[{product-title}
 SDN]. This network block should be private and must not conflict with any
 existing network blocks in your infrastructure to which pods, nodes, or the
-master may require access to, or the installation will fail. Defaults to
-`172.30.0.0/16`, and cannot be re-configured after deployment. If changing from the default, avoid `172.17.0.0/16`, which the *docker0* network bridge uses by default, or modify the *docker0* network.
+master may require access to, or the installation will fail. Reasonable value is
```
nengard: I'm not sure I understand why this was changed to 'reasonable value is' ... is this not the default? And why is it reasonable?

install_config/install/advanced_install.adoc
```diff
@@ -279,7 +279,7 @@ xref:../../architecture/networking/sdn.adoc#sdn-design-on-masters[SDN
 cluster network] CIDR block. This is the network from which pod IPs are
 assigned. This network block should be a private block and must not conflict
 with existing network blocks in your infrastructure to which pods, nodes, or the
-master may require access. Defaults to `10.128.0.0/14` and cannot be arbitrarily
+master may require access. Reasonable value is `10.128.0.0/14` and cannot be arbitrarily
```
nengard: Same question as above

install_config/install/advanced_install.adoc
```diff
@@ -288,8 +288,8 @@ master configuration].
 |This variable specifies the size of the per host subnet allocated for pod IPs
 by
 xref:../../architecture/networking/sdn.adoc#sdn-design-on-masters[{product-title}
-SDN]. Defaults to `9` which means that a subnet of size /23 is allocated to each
-host; for example, given the default 10.128.0.0/14 cluster network, this will
+SDN]. Reasonable value is `9` which means that a subnet of size /23 is allocated to each
```
nengard: Why reasonable? Do we mean 'suggested'?

install_config/install/advanced_install.adoc
adelton: As https://github.com/openshift/openshift-ansible/pull/5430 indicates, it is no longer the default -- that's the reason behind this docs pull request.

I consider it reasonable because it works reasonably well, one reason being that these were the previous defaults.

install_config/install/advanced_install.adoc
adelton: Dtto.

install_config/install/advanced_install.adoc
adelton: If it was suggested, we could just as well use the initial defaults. Since the defaults are no longer there, OpenShift team likely thinks that those values should no longer be suggested to people as something they should use without modification.

adelton:
Rebased on master -> 4649d7865b3a07061b40f87bf708e009e4a62b2f.

adelton:
Rebased on master -> a5806741c2d79453b2b30b73661420448a42aff0.

vikram-redhat:
@ahardin-rh - can you please move this one as well? It might be best to start a new PR and close this one.

ahardin-rh:
Continuing this work in https://github.com/openshift/openshift-docs/pull/12692

PR #5309, opened 2017-09-20
Use an annotation to provide a route cookie

pecameron:
Route annotation:
router.openshift.io/router.cookie.name
can be used to set a cookie name for the route.

The cookie must pass regex filter "^[a-zA-Z0-9_-][a-zA-Z0-9_-]*"

https://trello.com/c/dZpU26W9
(3) Allow routes to set the cookie names for session stickiness

origin PR 16454
https://github.com/openshift/origin/pull/16454


bfallonf:
@pecameron Thanks! If this is a new feature, would it require a more thorough write-up on how and why to use it? Or does this work with something that's already existing? I can see the corresponding card on the docs board, so I can add this to that for now.

pecameron:
On 09/21/2017 01:43 AM, Brice Fallon-Freeman wrote:
>
> @pecameron <https://github.com/pecameron> Thanks! If this is a new 
> feature, would it require a more thorough write-up on how and why to 
> use it? Or does this work with something that's already existing? I 
> can see the corresponding card on the docs board, so I can add this to 
> that for now.
>
A discussion might be helpful.
By default, as a security precaution, we generate an obscured name for 
the cookie. This permits the user to override the name when a specific 
name is needed by the app. The card, https://trello.com/c/dZpU26W9, and 
referenced bug lay out the use case. The environment variable is there 
if the router admin wishes to set a default cookie name for all cookies 
(overridden by the annotation in routes).

phil

bfallonf:
Thanks @pecameron . I'll merge this, then add the explanation to the docs card for the next sprint.

PR #5104, opened 2017-08-28
Update the current version of Kubernetes

php-coder:
The latest version of OpenShift uses Kubernetes 1.7 for some time already.

PTAL @simo5 @openshift/team-documentation

PR #5344, opened 2017-09-23
Revert "Update the current version of Kubernetes"

vikram-redhat:
Reverts openshift/openshift-docs#5104


vikram-redhat:
See: https://github.com/openshift/openshift-docs/pull/5104#issuecomment-331628207 for an explanation.

PR #5361, opened 2017-09-25
delete misleading sentence about blob storage

bparees:


bparees:

The behavior of whether blobs are stored or not is covered by a subsequent sentence (" By default, all the remote blobs served this way are stored locally for subsequent faster access unless `mirrorpullthrough` is disabled.") and it's misleading to say they are never stored.  (in fact by default they *are* stored).

bmcelvee:
LGTM. This is a minor change, so no Revision History needed.

bparees:
@bmcelvee it's ready for merge.

PR #5365, opened 2017-09-26
Router support for Strict-Transport-Security (hsts)

pecameron:
OSE 3.7

Strict-Transport-Security (hsts) policy (RFC 6797) is a way that the
host can tell clients to always use https requests to the host. It is
controlled by adding the haproxy.router.openshift.io/hsts_header annotation
to the route. When the Strict-Transport-Security response is received
by a client, it observes the policy until:
1) It is updated by another response from the host, or
2) the max-age decrements to 0

The max-age is only updated when the client receives a response
that contains the Strict-Transport-Security header. Otherwise
that the client just decrements max-age to 0.  When hsts policy is
no longer desired for a host set max-age=0 in the annotation rather
than deleting the annotation. When a client makes another request
the response will cause the policy to be discarded.

In the route add the annotation:
metadata:
  annotations:
    haproxy.router.openshift.io/hsts_header: max-age=31536000;includeSubDomains

Bug: 1430035
https://bugzilla.redhat.com/show_bug.cgi?id=1430035

Trello:
https://trello.com/c/H1FhCI1I/452-3-sccfsi-support-hsts-policy

### DIFF
--- a/architecture/networking/routes.adoc
+++ b/architecture/networking/routes.adoc
@@ -601,6 +601,7 @@ For all the items outlined in this section, you can set annotations on the
 |`*haproxy.router.openshift.io/timeout*` | Sets a server-side timeout for the route. xref:time-units[(TimeUnits)] | `ROUTER_DEFAULT_SERVER_TIMEOUT`
 |`*router.openshift.io/haproxy.health.check.interval*`| Sets the interval for the back-end health checks. xref:time-units[(TimeUnits)] | `ROUTER_BACKEND_CHECK_INTERVAL`
 |`*haproxy.router.openshift.io/ip_whitelist*` | Sets a xref:whitelist[whitelist] for the route. |
+|`*haproxy.router.openshift.io/hsts_header*` | Sets a xref:hsts[Strict-Transport-Security] header for the route. |
 |===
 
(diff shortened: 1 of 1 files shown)


pecameron:
@knobunc @bfallonf PTAL

bfallonf:
LGTM. If @knobunc gives the bug thumbs up, I'll merge and do a quick followup.

bfallonf:
[Document: [SCCFSI] Support HSTS Policy](https://trello.com/c/QsGgjanJ/164-document-sccfsi-support-hsts-policy)

zhaozhanqi:
@pecameron 
 Could you give an example or more steps to use this function and how to check the function is taking effected.  And if all kind of routes ( insecre/edge/passthrough/reencrypt) can use this feature? thanks

bfallonf:
The openshift bot above brought this PR to my attention. I created a followup PR in #6407 , which, @zhaozhanqi , I think fulfills your requests. If 6407 needs more work, please let me know. For now, I'll close this.

You can see the published section here: https://docs.openshift.com/container-platform/3.7/admin_guide/managing_networking.html#admin-guide-enabling-hsts

PR #5417, opened 2017-09-29
Bug 1487518 Added two variables to table in aggregated logging

bfallonf:
For: https://bugzilla.redhat.com/show_bug.cgi?id=1487518

### DIFF
--- a/install_config/aggregate_logging.adoc
+++ b/install_config/aggregate_logging.adoc
@@ -29,7 +29,7 @@ The EFK stack is a modified version of the
 https://www.elastic.co/videos/introduction-to-the-elk-stack[ELK stack] and is
 comprised of:
 
-* https://www.elastic.co/products/elasticsearch[Elasticsearch]: An object store where all logs are stored.
+* https://www.elastic.co/products/elasticsearch[Elasticsearch (ES)]: An object store where all logs are stored.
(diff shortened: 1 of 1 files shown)


ahardin-rh:
@bfallonf LGTM. 🎉 My only thought about this page in general, which may be considered outside of the scope of this PR, is that there are inconsistent references to ES and Elasticsearch. Lots of back and forth on what we call it. Perhaps the first usage on the page should be Elasticsearch (ES) and then just ES for the remainder of the topic.

ewolinetz:
:+1:

bfallonf:
@ahardin-rh You're right! I updated the first reference to have an (ES) in there, but it might be too much to change every use. Plus, I'm sure any future updates will mess with it more.

Thanks! I'll merge.

ruromero:
I think it will be very helpful for users to give some hint about the implications of having replicas=0 and that this value is not recommended for production environments. Maybe something simple like this:

> The number of replica shards per primary shard for every new index. Defaults to '0'. A minimum amount replica shards of 1 is advisable for production clusters

bfallonf:
Sure. Thanks @ruromero . I added what you suggested in #5481

bfallonf:
[rev_history]
|Added the `openshift_logging_es_number_of_replicas` and `openshift_logging_es_number_of_shards` variables to the table in xref:../install_config/aggregate_logging.adoc#aggregate-logging-ansible-variables[Specifying Logging Ansible Variables].
%

PR #5480, opened 2017-10-03
remove env var advice from Custom Builder documentation

### LINKED ISSUE #16472: Example custom builder does not adhere to recommendations in documentation
In the [custom builder documentation](https://docs.openshift.com/container-platform/3.6/creating_images/custom.html) it is stated that

> It is recommended that you write your Custom builder image to only read the information provided by the BUILD environment variable, which conforms to an OpenShift Container Platform API version. You can then parse the required information from the build definition JSON instead of relying on the other individual environment variables provided to the builder image. 

But the [example custom builder](https://github.com/openshift/origin/blob/master/images/builder/docker/custom-docker-builder/build.sh) only uses the _not recommended_ environment variables.

Background: Getting variables from a serialized JSON structure are not trivial, so it would be tremendously valuable if the example builder gave examples on how to to that.

jim-minter:
fixes https://github.com/openshift/origin/issues/16472

### DIFF
--- a/creating_images/custom.adoc
+++ b/creating_images/custom.adoc
@@ -85,14 +85,6 @@ enabled in the build configuration (if `*exposeDockerSocket*` was set to
 
 |===
 
-It is recommended that you write your Custom builder image to only read the
-information provided by the `*BUILD*` environment variable, which conforms to an
-{product-title} API version. You can then parse the required information from the
-build definition JSON instead of relying on the other individual environment
-variables provided to the builder image. However, the individual environment
-variables are provided for convenience if you do not want to parse the build
-definition.
-
 [[custom-builder-workflow]]
 == Custom Builder Workflow
 


jim-minter:
for discussion - @bparees?

bparees:
weighed in on this approach in the referenced issue.

i'm not necessarily averse to updating the docs in this way *and* updating the custom builder image to illustrate how to process the BUILD env, but i am not in favor of just saying "there's a BUILD env var, knock yourself out" because i know we're going to keep getting asked by people, "how can i parse that build env var?"

jim-minter:
OCP & Dedicated, starting from 3.9, should be fine - thanks

bmcelvee:
[rev_history]
|xref:../creating_images/custom.adoc#creating-images-custom[Custom Builder]
|Removed the environment variable recommendation from the xref:../creating_images/custom.adoc#custom-builder-image[Custom Builder Image] section.
%

PR #5228, opened 2017-09-11
Bug 1487015, updated Node.js version numbers

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1487015

PR #5486, opened 2017-10-04
Bumped Node.js version numbers

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1487015

Follow-up to https://github.com/openshift/openshift-docs/pull/5228


dinhxuanvu:
From what I see in NodeJS imagestream, the image for 0.10 is still there. However, it is marked as deprecated and it is hidden from view. That's why users can't see it. So we do need to address it in the doc. I would prefer to still have 0.10 and notice that it is deprecated and no longer available to use.

ahardin-rh:
@dinhxuanvu Thanks! I updated the language to reflect that.

dinhxuanvu (approved):
LGTM

ahardin-rh:
No revision history needed

PR #5607, opened 2017-10-09
Bug 1499222 Removed zones parameter info from AWS EBS example

ghost:


ahardin-rh:
@tmorriso-rh Yes, looks like your labeling and milestone are correct. 👍

For the typo of "definition" you can open a separate PR against master and apply it to 3.4, 3.5, and 3.6 since the error goes back that far. However, according to the BZ, the parameter edit here is only applicable to 3.6. So, I recommend creating separate PRs. Hope that makes sense!

tmorriso-rh:
No revision history needed.

PR #5621, opened 2017-10-10
clean up some imagestream doc

bparees:


bparees:
@dmage @miminar @legionus ptal.

bparees:
note that the bit about how to configure certificates for imagestream import will change with https://github.com/openshift/origin/pull/16294 but i'd like to get something in place now and we can updated it once https://github.com/openshift/origin/pull/16294 lands.

(also note that that is the section which I am most in need of your review for accuracy)

dev_guide/managing_images.adoc
```diff
@@ -918,6 +925,19 @@ After the secret is configured, proceed with creating the new image stream or
...
+=== Adding Trusted Certificates for External Registries
+
+If the registry you are importing from is using a certificate which is not signed
+by a standard certificate authority, you will need to explicitly configure the
+system to trust the registry's certificate or signing authority.  This can be 
+done by adding the CA certificate or registry certificate to the host system
+running the registry import controller (typically the master node).
+
+The certificate or CA certificate must be added to `/etc/pki/tls/certs` or `/etc/pki/ca-trust`,
+respectively, on the host system.
```
miminar: It would be worth to mention that `update-ca-trust` needs to be run on RH and similar systems. Whether the restart of marster/nodes is necessary and whether this applies to pods interacting with the registries.

miminar:
LGTM

dev_guide/managing_images.adoc
bparees: This is only for image-import, so i would assume non-master nodes and pods are unaffected. (There is other doc that needs to be done for those things, of course).

As for whether a master restart is required, honestly I don't know, but i think the answer is no, i think the CAs are reread on each client request setup.

bparees: updated w/ the update-ca-trust command

bparees:
@openshift/team-documentation ptal/merge.

dev_guide/managing_images.adoc
```diff
@@ -108,6 +112,9 @@ refreshed (i.e., re-imported) periodically. The period is configured globally at
 system level. See xref:importing-tag-and-image-metadata[Importing Tag and Image
 Metadata] for more details.
 
+The `--reference` flag will create an imagestream tag which is not imported.  The
```
mburke5678: s/which/that

dev_guide/managing_images.adoc
```diff
@@ -918,6 +925,20 @@ After the secret is configured, proceed with creating the new image stream or
 using the `oc import-image` command. During the import process, {product-title}
 will pick up the secrets and provide them to the remote party.
 
+[[trusting-registries]]
+=== Adding Trusted Certificates for External Registries
+
+If the registry you are importing from is using a certificate which is not signed
```
mburke5678: s/which/that

mburke5678:
@bparees Minor nits. Otherwise LGTM

bparees:
@mburke5678 updated and yes.

bparees:
@mburke5678 and 3.7+

mburke5678:
[rev_history]
|xref:../dev_guide/managing_images.adoc#dev-guide-managing-images[Manging Images]
|Clarified wording around Image Streams and added new section on xref:../dev_guide/managing_images.adoc#trusting-registries[Adding Trusted Certificates for External Registries].
%

vikram-redhat:
@mburke5678 - this doesn't have the dedicated and online labels? I have added these, but want to make sure this change applied to dedicated and online as well?

bparees:
yeah it applies everywhere.  at least most of it does.  online users aren't going to be messing w/ the host certificates, but they will use oc tag, for example.

PR #5677, opened 2017-10-12
Fix file path to playbooks

liangxia:
The correct path contains ```playbooks```
https://github.com/openshift/openshift-ansible/blob/master/playbooks/byo/openshift-cluster/openshift-metrics.yml
https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-cluster/openshift_provisioners.yml


liangxia:
Applies to branch/enterprise-3.6 and above.

liangxia:
Latest version already contains correct path, this pr is no longer needed.

PR #5360, opened 2017-09-25
Doc. update for flannel support.

e-minguez:
https://access.redhat.com/documentation/en-us/reference_architectures/2017/html-single/deploying_and_managing_red_hat_openshift_container_platform_3.6_on_red_hat_openstack_platform_10/#flannel_red_hat_openstack_platform_network

Related BZ https://bugzilla.redhat.com/show_bug.cgi?id=1490960

PR #5705, opened 2017-10-13
Follow up to PR 5360

mburke5678:
See https://github.com/openshift/openshift-docs/pull/5360

### DIFF
--- a/install_config/configuring_sdn.adoc
+++ b/install_config/configuring_sdn.adoc
@@ -288,7 +288,7 @@ service neutron-vpn-agent restart
 service neutron-server  restart
 ----
 
-. When creating the {product-title} instances, disable both port security and security
+. When creating the {product-title} instances on Red Hat OpenStack Platform, disable both port security and security
 groups in the ports where the container network flannel interface will be:
 +
 ----
--- a/install_config/configuring_sdn.adoc
+++ b/install_config/configuring_sdn.adoc
@@ -228,7 +228,7 @@ edge load-balancer to containers within OpenShift SDN].
 
 [[using-flannel]]
 == Using Flannel
-As an alternative to the default SDN, {product-title} also provides Ansible
+As an alternate to the default SDN, {product-title} also provides Ansible
 playbooks for installing *flannel*-based networking. This is useful if running
 {product-title} within a cloud provider platform that also relies on SDN, such
 as Red Hat OpenStack Platform, and you want to avoid encapsulating packets twice
@@ -240,15 +240,14 @@ a container from attempting to contact any IP address in the same network
 space. This hinders multi-tenancy because the network cannot be used to isolate
 containers in one application from another.
 
-The key to making the right decision rests between mutli-tenancy isolation vs
-performance. Depending on the priority of each of these, should determine the
-appropriate choice when deciding between OpenShift SDN and flannel internal
-networks.
+Depending on whether you prefer mutli-tenancy isolation or performance, you should determine the
+appropriate choice when deciding between OpenShift SDN (multi-tenancy) and flannel (performance)
+for internal networks.
 
 ifndef::openshift-origin[]
 [IMPORTANT]
 ====
-This is only supported for {product-title} on Red Hat OpenStack Platform.
+Flannel is only supported for {product-title} on Red Hat OpenStack Platform.
 ====
 endif::[]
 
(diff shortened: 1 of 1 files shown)


mburke5678:
@e-minguez Can you take a look to make sure I haven't changed the meaning of anything?

Also, should the `Flannel iptables rules` section be in the `To enable flannel` procedure above it?

install_config/configuring_sdn.adoc
```diff
@@ -240,28 +240,27 @@ a container from attempting to contact any IP address in the same network
...
 [IMPORTANT]
 ====
 The current version of Neutron enforces port security on ports by default. This
 prevents the port from sending or receiving packets with a MAC address
-different from that on the port itself. flannel creates virtual MACs and IP
-addresses and must send and receive packets on the port so port security must
-be disabled on the ports that carry flannel traffic.
+different from that on the port itself. Flannel creates virtual MACs and IP
+addresses and must send and receive packets on the port, so port security must
+be disabled on the ports that carry Flannel traffic.
```
e-minguez: Flannel vs flannel

install_config/configuring_sdn.adoc
```diff
@@ -288,18 +287,21 @@ service neutron-plugin-openvswitch-agent restart
 service neutron-vpn-agent restart
 service neutron-server  restart
 ----
-
-When creating the OpenStack instances, disable both port-security and security
-groups in the ports where the container network Flannel interface will be:
-
++
+When creating the {product-title} instances, disable both port security and security
+groups in the ports where the container network flannel interface will be:
```
e-minguez: flannel vs Flannel

install_config/configuring_sdn.adoc
```diff
@@ -316,15 +318,18 @@ flannel_interface=eth0
 using the `flannel_interface` variable. Without this variable, the
 {product-title} installation uses the default interface.
 
-NOTE: In order to provide custom networking CIDR for pods and services using
-flannel, there is a bugzilla being invested by Red Hat Bugzilla ID 1473858 at
-the time of writing this paper.
+[NOTE]
+====
+Custom networking CIDR for pods and services will be supported in a future release.
```
e-minguez: Custom networking CIDR for pods and services **using flannel** will be supported in a future release.

install_config/configuring_sdn.adoc
```diff
@@ -346,14 +351,17 @@ iptables -t nat -A POSTROUTING -o eth1 -j MASQUERADE
...
 
-NOTE: iptables-save saves all the current "in memory" iptables rules, but, as
-both Docker, Kubernetes and {product-title} creates a lot of iptables rules
-(services, etc.) not designed to be persisted, you will end up with a messy
-/etc/sysconfig/iptables file.
+[NOTE]
+====
+The `iptables-save` command saves all the current _in memory_ iptables rules.
+However, because Docker, Kubernetes and {product-title} create a high number of iptables rules
+(services, etc.) not designed to be persisted, saving these rules can result in a large
```
e-minguez: I'd say in a "bad" file instead large, because the problem is if you persist the k8s services, docker, etc. rules, you will be screwed 🗡

e-minguez (changes requested):

e-minguez:
@mburke5678 I've added a few comments, otherwise LGTM

install_config/configuring_sdn.adoc
mburke5678: In the flannel docs, they use `flannel` with lower case `f` except at the beginning of sentences where they use `Flannel`. I tried to match that.

e-minguez: Line 260 should be lower case then.

mburke5678:
@e-minguez should the Flannel iptables rules section be in the To enable flannel procedure above it? Is that a required process to enable flannel, it seems that it should be in the steps.

e-minguez:
Yes, it is required, but it should be performed after the OCP installation, so it is ok to be moved into the enable flannel procedure as a subsection or something similar.
Thanks

install_config/configuring_sdn.adoc
```diff
@@ -289,19 +288,22 @@ service neutron-vpn-agent restart
 service neutron-server  restart
 ----
 
-When creating the OpenStack instances, disable both port-security and security
-groups in the ports where the container network Flannel interface will be:
-
+. When creating the {product-title} instances, disable both port security and security
```
vikram-redhat: Should this be changed? Our product title is not OpenStack?

e-minguez: Maybe:
"When creating the {product-title} instances in OpenStack,..."

mburke5678:
@e-minguez If the user needs to configure the Ansible playbook before installing OpenShift, we should probably add these steps to the Prerequisites or Host Preparation files before the install?
_Set the following variables in your Ansible inventory file before running the
installation_

https://docs.openshift.com/container-platform/3.6/install_config/install/host_preparation.html

Please let me know what you think so that we can close this PR!

e-minguez:
Currently, the "openshift_use_flannel" ansible variable documentation has a link to the Flannel section on the SDN chapter[1] which I think it is enough.

[1] https://docs.openshift.com/container-platform/3.6/install_config/install/advanced_install.html#configuring-cluster-variables

mburke5678:
Beautiful. Without further objection, I will merge!

mburke5678:
No rev history needed.

vikram-redhat:
@mburke5678 - 4 commits here which should have been squashed into one. Also, this didn't get picked into the right branches once merged into master or the labels applied. Finally, the original dev PR needs to get the to_followup label removed.

vikram-redhat:
I am reverting this as this is causing issues when doing a weekly release due to the issues mentioned in the previous comment from me. @mburke5678 will do a follow up to fix these issues.

Issue #2842, opened 2016-09-14
Backup/Restore should talk more about the relationship between etcd and storage


danmcp:
This page:

https://docs.openshift.org/latest/admin_guide/backup_restore.html

Briefly mentions volume storage.  But it would be helpful to be more elaborate about how the etcd backup still has all the references to the storage volumes.  So when you restore etcd, OpenShift will start launching the previous pods on nodes and reattaching the same storage.  It might be useful to give the analogy that this is really no different than the process of when you remove a node from the cluster and add a new one back in its place.  Anything attached to that node will be reattached to the pods on whatever nodes they get rescheduled to.

ahardin-rh:
Addressed in https://github.com/openshift/openshift-docs/pull/5706

PR #5706, opened 2017-10-13
Added more information about the relationship between etcd and storage

ahardin-rh:
Addresses https://github.com/openshift/openshift-docs/issues/2842


ahardin-rh:
@openshift/team-documentation PTAL. Thoughts on placement of this information?

@danmcp FYI

mburke5678:
@ahardin-rh LGTM Would it make sense to put the note in each of the restore procedures under the restore step? Or, is that to much?

ahardin-rh:
@mburke5678 After mulling it over, I will keep it in one place vs. duplicating the information.

PR #5731, opened 2017-10-16
document explicit openstack block version configuration

### LINKED ISSUE #5730: Cinder auto-detection fails for OpenShift on OpenStack without bs-version set.
In the "Configuring for OpenStack" content an example `/etc/cloud.cfg` is provided. Not mentioned is the fact that on newer OpenStack installs that expose a Cinder endpoint using paths rather than ports this configuration will result in an error along the lines of:

> Failed to attach volume "jenkins2-volume" on node "app-node-1.openshift.staging.xxxx.co" with: BS API version autodetection failed.

To resolve this until such time as OpenShift/Kubernetes properly support Cinder V3 and/or the auto-detection login in the K8S cloud provider is changed it is necessary to force a fallback to V2 using:

```
[BlockStorage]
bs-version=v2
```
This issue arises frequently for users who frequent the #sig-openstack Kubernetes slack channel. If or when OpenShift pulls in https://github.com/kubernetes/kubernetes/pull/53115 this would also resolve the issue.

[1] https://github.com/openshift/openshift-docs/blob/master/install_config/configuring_openstack.adoc
[2] https://docs.openshift.com/container-platform/3.6/install_config/configuring_openstack.html

rootfs:
fix #5730 
@pospispa @xsgordon

### DIFF
--- a/install_config/configuring_openstack.adoc
+++ b/install_config/configuring_openstack.adoc
@@ -32,11 +32,17 @@ region = <OS_REGION_NAME>
 
 [LoadBalancer]
 subnet-id = <UUID of the load balancer subnet>
+[BlockStorage]
+bs-version=v2
 ----
 ====
 
 Consult your OpenStack administrators for values of the `*OS_*` variables, which
 are commonly used in OpenStack configuration.
+
+Currently OpenStack Cinder V3 API is not supported. To resolve this issue or disable auto Cinder API version detection, it is necessary to force Cinder V2 API using `bs-version=v2`.
+
+
 
 [[openstack-configuring-masters]]
 == Configuring {product-title} Masters for OpenStack


pospispa:
@rootfs would it make sense to make the description more detailed like the one below?

Currently OpenStack Cinder V3 API is not supported.
The `bs-version` configuration parameter can be set to the below values:
- `auto`: default value, Cinder API version auto-detection. The auto-detection may fail for some OpenStack configurations.
- `v2`: Cinder V2 API is used.
- `v1`: Cinder V1 API is used.

rootfs:
@pospispa auto detection doesn't always work, so explicit setting to v1 or v2 is a workaround.

pospispa:
> @pospispa auto detection doesn't always work, so explicit setting to v1 or v2 is a workaround.

I agree with that.

xsgordon:
lgtm!

pospispa:
> Does this fix apply to all OCP versions (3.1 - 3.7)?

@rootfs please, will you comment on the above.

kalexand-rh:
I emailed @rootfs, and he said that this change applies to only 3.6 and 3.7.

kalexand-rh:
I've copied this change to the 3.7 branch on https://github.com/openshift/openshift-docs/pull/9742. Closing this PR and merging that one instead.

PR #5745, opened 2017-10-17
Removed outdated browser requirements

ahardin-rh:
@wgordon17 per our recent conversation regarding https://bugzilla.redhat.com/show_bug.cgi?id=1500724


wgordon17:
Weren't we going to try to point them to the release notes because the browser requirements were supposed to be called out there from now on, right?

ahardin-rh:
@wgordon17 Yeah, step 1 is removing the outdated information, then possibly adding the corrected information back in.

@spadgett @jwforres Thoughts on this? This list falls out date quickly. Is this something that we should keep in the docs and, if so, should we move it to release notes for each version? Thanks!

spadgett:
So we're saying the latest version of each browser, which should be OK. Are the supported operating systems wrong?

For iPhone / iPad, I would just say latest iOS version instead of listing specific devices.

ahardin-rh:
@spadgett Okay, thanks. A few follow-up questions:

- Do you think the information belongs here or in the Release Notes? 

- Are these requirements the same for 3.6, 3.5, 3.4, etc.? 

- Should I update the language to be "Fedora 23 and later, Windows 8 and later"?

spadgett:
> Do you think the information belongs here or in the Release Notes?

I don't have a strong opinion. @jwforres ?

> Are these requirements the same for 3.6, 3.5, 3.4, etc.?

At some point we supported IE10, but I'd have to do some investigation to know what release we dropped support.

> Should I update the language to be "Fedora 23 and later, Windows 8 and later"?

Yeah.

jwforres:
@ahardin-rh can we just point to https://access.redhat.com/articles/2176281  the Client Tools section has what browsers were tested for each version of openshift

ahardin-rh:
@jwforres That works for me! Looks like that is for OCP. Is it the same for Online and Dedicated?

ahardin-rh:
@jwforres @spadgett @wgordon17 Updated. 

Preview build for Online/Dedicated, which instructs how to find the cluster's version number:

http://file.rdu.redhat.com/~ahardin/10202017/browser-requirements/architecture/infrastructure_components/web_console.html#browser-requirements

wgordon17:
👍

PR #5764, opened 2017-10-18
Fix typos in "Getting Started"

aaronang:


ahardin-rh:
@aaronang Thanks for your suggestion. However, we go by the IBM Style Guide, which states that check mark should be two words.

PR #5775, opened 2017-10-18
[OCP 3.4] Removed Templates from Architecure TOC

mburke5678:
No rev history needed. Separate PR covered the removal of this file.

PR #5773, opened 2017-10-18
[Online] Removed Templates from Arch TOC

mburke5678:
No rev history needed. Separate PR covered the removal of this file.

PR #5772, opened 2017-10-18
[Dedicated] Removed Templates from Architecure TOC

mburke5678:

PR #5776, opened 2017-10-18
[OCP 3.6] Removed Templates from Architecure TOC

mburke5678:
No rev history needed. Separate PR covered the removal of this file.

### DIFF
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -187,6 +187,7 @@ Topics:
         Distros: openshift-*
       - Name: Templates
         File: templates
+        Alias: dev_guide/templates
         Distros: openshift-*
   - Name: Additional Concepts
     Dir: additional_concepts


mburke5678:
No rev history needed. Separate PR covered the removal of this file.

vikram-redhat:
@mburke5678 - 

1. This should only have been merged to the stage branches, not directly to the prod branch. If it has been directly pushed to the prod branch, then it doesn't need to go in the bucket for next release (because technically, it has been released by being the prod branch). However, note that this causes a problem by merging this before https://github.com/openshift/openshift-docs/pull/5755 has been merged. That PR hasn't merged to prod branches yet and will only merge in next release, but this PR 5776 has been which means that this PR is fixing a problem that is not yet there.

2. If it is for 3.6 and 3.7, then I don't see the a merged PR in the 3.7 stage branch, only the 3.6 stage branch.

I will remove the milestones for this and the others: https://github.com/openshift/openshift-docs/pull/5775 https://github.com/openshift/openshift-docs/pull/5773 https://github.com/openshift/openshift-docs/pull/5772.

PR #5831, opened 2017-10-21
remove extraneous space

dlbewley:
a space was causing `_` to render literally rather than italicize the adjacent string


ahardin-rh:
Thank you!

ahardin-rh:
No revision history needed

PR #5834, opened 2017-10-21
update example externalIP using rfc5737

dlbewley:
the example network from `master-config.yaml` did not match the `externalIP` used by the service below it.
i propose the use of RFC5737 IP addresses since these are meant to simulate IPs external to the cluster.


tmorriso-rh:
LGTM.

tmorriso-rh:
Closing this PR. The work will continue in https://github.com/openshift/openshift-docs/pull/6455.

PR #5890, opened 2017-10-24
[DO NOT MERGE] Alias demonstration

nhr:
@mburke5678 please review


nhr:
You can ignore some of the cruft. The items of main importance:

1. The original topic file (in this case, `dev_guide/getting_traffic_into_cluster`) is removed from the repo.
2. The new topic (in this case, `expose_service/index`) is added to the repo.
3. In the topic map, we add an Alias parameter in the original topic entry that points to the new topic.

Having done this, I tested my changes by doing the following from the root of the `openshift-docs` git repo on my machine:

1. asciibinder clean
2. asciibinder build
3. cd _preview/openshift-enterprise
4. ruby -run -e httpd . -p 3000

In my case, my working branch was called `alias_demonstration`, and so the redirecting URL was:

`http://localhost:3000/alias_demonstration/dev_guide/getting_traffic_into_cluster.html`

Which correctly redirected me to:

`http://localhost:3000/alias_demonstration/expose_service/index.html`

mburke5678:
@nhr  Are you saying that the Alias to a subdirectory will work?

PR #5918, opened 2017-10-24
fix(): List indent

tripledes:
* Without a break line, list is not rendered properly, resulting in
continuous line including '*'


tripledes:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1505780

ahardin-rh:
Thank you!

ahardin-rh:
No revision history needed

PR #5944, opened 2017-10-25
updated comments

arshadkhan39687:


ahardin-rh:
There should be no extra space before the period.

Issue #5946, opened 2017-10-25
Broken links in docs.openshift.org


anjannath:
The following two links are not working.

- https://docs.openshift.org/latest/architecture/core_concepts/routes.html#architecture-core-concepts-routes

- https://docs.openshift.org/latest/architecture/core_concepts/templates.html#architecture-core-concepts-templates

saritha-infosys:
1st link is not working, we can see the info in the below link

https://docs.openshift.org/latest/architecture/core_concepts/index.html

anjannath:
found one more broken link: https://docs.openshift.org/latest/dev_guide/getting_traffic_into_cluster.html#using-ingress-IP-self-service

gaurav-nelson:
Thank you @anjannath for creating this issue, can you please also share these links are broken on which pages?

anjannath:
@gaurav-nelson More information can be found on this issue [openshift/origin#1593](https://github.com/openshift/origin/issues/17165)

gaurav-nelson:
Thank you @anjannath for the updates. This one is already fixed in the current version.

PR #5957, opened 2017-10-25
Fix link #5946 

therrance:
#5946


gaurav-nelson:
Thanks for the PR @Therrance 

This one is already fixed in the latest version.

PR #6016, opened 2017-10-30
typo, which described in fact the opposite of what NoSchedule does

### LINKED ISSUE #5159: Taints and Tolerations - possible incorrect meaning
In the docs @ https://docs.openshift.org/latest/admin_guide/scheduling/taints_tolerations.html
It says under the NoSchedule effect that 

NoSchedule | New pods that do not match the taint are scheduled onto that node.

Based on my reading of https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/ this should read 

NoSchedule | New pods that do not match the taint are not scheduled onto that node.

However please check as my interpretation may be incorrect

clemSeveillac:
Fixes #5159 by documenting correct NoSchedule effect

### DIFF
--- a/admin_guide/scheduling/taints_tolerations.adoc
+++ b/admin_guide/scheduling/taints_tolerations.adoc
@@ -42,7 +42,7 @@ Taints and tolerations consist of a key, value, and effect. An operator allows y
 [cols="2a,3a"]
 !====
 !`NoSchedule`
-!* New pods that do not match the taint are scheduled onto that node.
+!* New pods that do not match the taint are not scheduled onto that node.
 * Existing pods on the node remain.
 !`PreferNoSchedule`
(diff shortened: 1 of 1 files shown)


ahardin-rh:
Thank you!

vikram-redhat:
@ahardin-rh - does this need cherry picking into the right stage branches?

ahardin-rh:
@vikram-redhat That was done in https://github.com/openshift/openshift-docs/pull/6018 and https://github.com/openshift/openshift-docs/pull/6017 but somehow the cross-reference didn't render here 😕

vikram-redhat:
Ah.. I see. Weird.

vikram-redhat:
Hmm.. I see the error and have applied a fix.

PR #6078, opened 2017-11-01
[3.4] There's no hard pruning in the 1.4 release of origin

miminar:


miminar:
This should be opened against enterprise-3.4-stage.

PR #6081, opened 2017-11-01
Fix example autoscale command

ghyde:


tmorriso-rh:
Note to TW: this change applies to OCP 3.2 - 3.9, online, and dedicated.

PR #6127, opened 2017-11-03
Improvements to NetworkPolicy docs

danwinship:
1. Remove lingering reference to the pre-3.6 NetworkPolicy semantics ("opt in to isolation by configuring the Namespace object")
2. Give a few more specific NetworkPolicy examples
3. Be clearer about the 3.6-and-later semantics where pods are isolated if and only if they are selected by some NetworkPolicy
4. Be clearer about combining NetworkPolicies
5. Talk about modifying the default project template to automatically create NetworkPolicies

Feel free to rewrite

@bfallonf

### DIFF
--- a/admin_guide/managing_networking.adoc
+++ b/admin_guide/managing_networking.adoc
@@ -1029,35 +1029,47 @@ xref:../install_config/configuring_sdn.adoc#install-config-configuring-sdn[confi
 to use the *ovs-networkpolicy* plug-in], network isolation is controlled
 entirely by
 link:https://github.com/kubernetes/community/blob/master/contributors/design-proposals/network/network-policy.md[`NetworkPolicy`
-objects]. In particular, by default, all projects are able to access pods in all
-other projects. To isolate a project, opt in to isolation by configuring the
-Namespace object, then create a `NetworkPolicy` object to indicate the allowed
-incoming connections.
+objects]. In particular, by default, all pods in a project are accessible from all sources.
+To isolate one or more pods in a project, you can create `NetworkPolicy` objects in that project
+to indicate the allowed incoming connections. Project administrators can create and delete
+`NetworkPolicy` objects within their own project.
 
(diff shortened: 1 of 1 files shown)


danwinship:
(This is part of https://trello.com/c/aMVNlGpD "Bring NetworkPolicy to GA".)

admin_guide/managing_networking.adoc
```diff
@@ -1029,35 +1029,47 @@ xref:../install_config/configuring_sdn.adoc#install-config-configuring-sdn[confi
...
+spec:
+  podSelector:
+  ingress: []
 ----
-<1> Set to *redhat/openshift-ovs-networkpolicy* for the *ovs-networkpolicy* plug-in
 
-. Configure the pod network policy on each node configuration file at *_/etc/origin/node/node-config.yaml_*:
-+
+Alternatively, you could make pods accept connections from other pods in the same project,
+but reject all other connections from other projects:
```
knobunc: Should we mention that this would give the same behavior as with the multi-tenant plugin?

knobunc (approved):
LGTM, just a minor question.

admin_guide/managing_networking.adoc
danwinship: It's not "the same"; it doesn't have the VNID 0 special case for instance. We could say it's "similar to" multitenant or something like that.

gaurav-nelson:
Merged this one, tracking language edits in https://github.com/openshift/openshift-docs/pull/6223

vikram-redhat:
@danwinship would these changes apply to 3.6 as well? Or just 3.7 and above?

danwinship:
@vikram-redhat This applies to 3.6 too, but not 3.5.

vikram-redhat:
@knobunc - would you happen to know if the Network Policy should have applied to Dedicated? @spurtell - would you know?

PR #6156, opened 2017-11-06
Revert "removed latest-tag and added version to command"

mburke5678:
Reverts openshift/openshift-docs#6146


mburke5678:
Made change to stage branch directly in error. Reverting.

PR #6179, opened 2017-11-06
Add an example to edit the scc non-interactively

dmsimard:
It's useful to be able to edit the scc non-interactively, for example
when wrapping the setup with configuration management.
Let's include an example that shows how to toggle
allowHostDirVolumePlugin to true for use with hostPath volumes.

### DIFF
--- a/admin_guide/manage_scc.adoc
+++ b/admin_guide/manage_scc.adoc
@@ -437,6 +437,19 @@ $ oc edit scc restricted
 
 . Save the changes.
 
+=== Edit the SCCs non-interactively
+
+You can edit the different SCCs non-interactively with the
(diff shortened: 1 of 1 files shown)


php-coder (changes requested):
Thank you for this PR! I like this change!

The only one thing that is bothering me is the size of the SCC doc that is growing. In order to resist to that, I suggest the following:
1) instead of creating another sub-topic, update the existing
2) replace example that is using `oc edit` with `oc patch` command
3) mention that it also possible to use `oc edit` for interactively editing

Thanks!

php-coder:
P.S. Double bonus if you also update the previous section "Modify Cluster Default Behavior" ;)

kalexand-rh:
I made a follow-up PR here: https://github.com/openshift/openshift-docs/pull/11195

PR #6191, opened 2017-11-07
clarify when env variables appear in builds

bparees:


bparees:
@openshift/team-documentation ptal, this behavior has confused users so i'm trying to make it more front and center.  also fixing an extraneous character.

bmcelvee:
LGTM. No Revision History needed.

dev_guide/builds/build_strategies.adoc
```diff
@@ -118,15 +118,16 @@ documentation] for information on how S2I scripts are used.
 
 There are two ways to make environment variables available to the
 xref:../../architecture/core_concepts/builds_and_image_streams.adoc#source-build[source build]
-process and resulting \image: xref:environment-files[environment files] and
-xref:buildconfig-environment[*BuildConfig* environment] values.
+process and resulting image: xref:environment-files[environment files] and
```
vikram-redhat: @bparees @bmcelvee - the build system barfed on the use of the `image:` here. It thought that you were trying to incorporate an image tag :P. I had no idea that it would be a problem, but I guess we should have escaped it.

I have fixed this in the branches and the master.

bparees: hm, well i deleted the \ because it was showing up in the rendered docs.

bmcelvee: How strange. It rendered correctly when I built it locally. Thanks for fixing it, @vikram-redhat!

vikram-redhat: No worries. I should have clarified. This was failing on the portal docs, which is stricter with this stuff.

PR #6205, opened 2017-11-07
Add reset Docker storage per BZ-1501504

mburke5678:
WIP first draft
https://bugzilla.redhat.com/show_bug.cgi?id=1501504

### DIFF
--- a/admin_guide/manage_nodes.adoc
+++ b/admin_guide/manage_nodes.adoc
@@ -487,6 +487,90 @@ default value for `max-pods` is `250`. This means that unless the node has 25
 cores or more, by default, `pods-per-core` will be the limiting factor.
 // end::admin_guide_manage_nodes[]
 
+[[managing-nodes-docker-reset]]
+== Resetting Docker Storage
+
+As you download Docker images and run and delete containers, Docker does not always free up mapped disk space. As a result, over time you can run out of space on a node,
+which might prevent {product-title} from being able to create new pods or cause pod creation to take several minutes.
+
+For example, the following shows pods that are still in the `ContainerCreating` state after six minutes and the events log shows a xref:../dev_guide/events.adoc#events-reference[FailedSync event].
+
+[source,bash]
+----
+$ oc get pod
+NAME                               READY     STATUS              RESTARTS   AGE
+cakephp-mysql-persistent-1-build   0/1       ContainerCreating   0          6m
+mysql-1-9767d                      0/1       ContainerCreating   0          2m
+mysql-1-deploy                     0/1       ContainerCreating   0          6m
+
+$ oc get events
+LASTSEEN   FIRSTSEEN   COUNT     NAME                               KIND                    SUBOBJECT                     TYPE      REASON                         SOURCE                                                 MESSAGE
+6m         6m          1         cakephp-mysql-persistent-1-build   Pod                                                   Normal    Scheduled                      default-scheduler                                      Successfully assigned cakephp-mysql-persistent-1-build to ip-172-31-71-195.us-east-2.compute.internal
+2m         5m          4         cakephp-mysql-persistent-1-build   Pod                                                   Warning   FailedSync                     kubelet, ip-172-31-71-195.us-east-2.compute.internal   Error syncing pod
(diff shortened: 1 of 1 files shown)


mburke5678:
@jupierce PTAL

mburke5678:
@mdshuai 
Can you take a look at the PR as you were the QA contact on the original PR? | Collapse All

admin_guide/manage_nodes.adoc
```diff
@@ -487,6 +487,82 @@ default value for `max-pods` is `250`. This means that unless the node has 25
...
+. Run the following command to shut down Docker and the *atomic-openshift-node* service:
++
+----
+$ systemctl stop docker atomic-openshift-node
+----
+
+. Run the following command to shut down Docker and the the *atomic-openshift-node* service:
++
+----
+$ systemctl stop docker atomic-openshift-node
```
mdshuai: Total same step with above?

mdshuai:
I think you can use the step @jupierce provided.  One thing need update is you can use `oadm cordon/uncordon` to make node as `unschedulable/schedulable`

mburke5678:
@mdshuai We don't refer to oadm cordon in the docs. In the same topic, we use `oadm manage-node <node> --schedulable=false/true` to mark a node as unschedulable.  Is that the correct command?

mburke5678:
@jupierce You code has a line to remove the volume directory:
`rm -rf /var/lib/origin/openshift.local.volumes`

Should we include that?

jupierce:
@sjenning We deleted local volumes during our interactive work on the nodes. Do you want to include it in the docs?

sjenning:
@jupierce yes, I think so.  If we are going to document this, might as well include everything need to get the node back to a pristine storage condition.

mburke5678:
@jupierce @sjenning Is there any negative consequence to deleting the local volumes? Can this be an optional step, if the user wants to retain them? Does stopping Docker make these volumes unneeded as any associated containers would be wiped?

sjenning:
@mburke5678 it clears the local image cache so images, including infra ose-* images, will need to be repulled and might result in slower pod start times while the image store warms back up.

I would consider this to be an "all or nothing" procedure.  Picking a choosing things to delete might result in weird state.  I don't know all the assumptions docker makes.

mburke5678:
@jupierce @mdshuai @sjenning Can these changes roll back to OCP 3.4? 
I am not sure when docker-storage-setup was added.

mburke5678:
[rev_history]
|xref:../admin_guide/manage_nodes.adoc#admin-guide-manage-nodes[Manging Nodes]
|Added new section on xref:../admin_guide/manage_nodes.adoc#managing-nodes-docker-reset[Resetting Docker Storage] to free up space on nodes.
%

mburke5678:

admin_guide/manage_nodes.adoc
```diff
@@ -487,6 +487,90 @@ default value for `max-pods` is `250`. This means that unless the node has 25
...
+----
+
+One solution to this problem is to reset Docker storage to remove artifacts not needed by Docker.
+
+On the node where you want to restart Docker storage:
+
+. Run the following command to mark the node as unschedulable:
++
+----
+oadm manage-node <node> --schedulable=false
```
bmcelvee: $

admin_guide/manage_nodes.adoc
```diff
@@ -487,6 +487,90 @@ default value for `max-pods` is `250`. This means that unless the node has 25
...
+. Run the following command to shut down Docker and the *atomic-openshift-node* service:
++
+----
+$ systemctl stop docker atomic-openshift-node
+----
+
+. Run the following command to remove the local volume directory:
++
+----
+rm -rf /var/lib/origin/openshift.local.volumes
```
bmcelvee: $

bmcelvee:
@mburke5678 The example log is building strangely for me, entirely red text with red highlighting. Is it building that way for you?

Otherwise LGTM! 🍂

mburke5678:
@bmcelvee Thanks. I changed a [source,yaml] to [source,bash] for those example commands. Does that look better?

bmcelvee:
@mburke5678 Yes, better! 😸

vikram-redhat:
@mburke5678 - you had enterprise 3.4 as one of the branches where this change applies, but you haven't picked it to that branch. Also, does it still need the peer review needed label, even though it has been merged?

mburke5678:
@vikram-redhat Done. Thanks.

PR #6225, opened 2017-11-08
Fix mysql links

liangxia:
This should apply to branch/enterprise-3.3 and above.


liangxia:
@openshift/team-documentation  PTAL.

gaurav-nelson:
Thanks for the PR @liangxia LGTM

minor fixes no peer review needed.

gaurav-nelson:
@vikram-redhat Do we need to include this in v3.3 as well?

vikram-redhat:
@gaurav-nelson - if @liangxia says so. But we don't have a stage branch for 3.3, so put this directly in the prod branch for 3.3.

PR #6311, opened 2017-11-10
Fix image: in build_strategies

vikram-redhat:


vikram-redhat:
This has already been picked to all branches as part of weekly release.

PR #6331, opened 2017-11-10
fixed typo in remote-syslog section

sqtran:
The parameter should be REMOVE_TAG_PREFIX, not REMOTE_TAG_PREFIX.

I verified that this is just a documentation bug, the code is correct.


tmorriso-rh:
Work continued in https://github.com/openshift/openshift-docs/pull/7301.

PR #6344, opened 2017-11-13
remove incorrect note about scheduled import

bparees:


bparees:
according to @smarterclayton this note isn't accurate (scheduled import is *enabled* by default).  That matches what i see when i do an openshift start --write-config (the config contains disabled=false).

@soltysh @mfojtik ptal/confirm.

dev_guide/managing_images.adoc
```diff
@@ -826,12 +826,6 @@ spec:
       scheduled: true
 ----
 
-[NOTE]
-====
-By default, `*importPolicy.scheduled*` is disabled and must be enabled in the
```
soltysh: Wait, there are two things mixed here. One is `importPolicy.scheduled` which is set *on an Image Stream*, per-tag to be honest. And second is the master config option `disableScheduledImport` which indeed is `false` by default. I'd suggest changing this description to explicitly mention that the IST's `scheduled` is disabled, and needs to be enabled explicitly.

soltysh (changes requested):

soltysh:
I've just spoke with @bparees on IRC, give the context of this change, this LGTM.

soltysh (approved):

bparees:
@openshift/team-documentation got confirmation from @soltysh on irc, please merge.

bmcelvee:
No revision history needed.

PR #6407, opened 2017-11-17
Added enabling HSTS section

bfallonf:
From: #5365 

@pecameron I created this with info from your PR above, but added some context, moved the info, and gave it an edit.

As @zhaozhanqi asks in the other PR, is this available for all routes (insecure/edge/passthrough/reencrypt)? How can we verify that the route is accepting the feature? If you have any other thoughts, feel free to let me know as well.

cc @knobunc

### DIFF
--- a/admin_guide/managing_networking.adoc
+++ b/admin_guide/managing_networking.adoc
@@ -1228,3 +1228,56 @@ objects:
             name: default
 ...
 ----
+
+[[admin-guide-enabling-hsts]]
+== Enabling HTTP Strict Transport Security
+
+HTTP Strict Transport Security (HSTS) policy is a security enhancement, which
+ensures that only HTTPS traffic is allowed on the host. Any HTTP requests are
+dropped by default. This is useful for ensuring secure interactions with
+websites, or to offer a secure application for the user's benefit.
+
+When HSTS is enabled, HSTS adds a Strict Transport Security header to HTTPS
+responses from the site. You can use the `insecureEdgeTerminationPolicy` value
+in a route to redirect to send HTTP to HTTPS. However, when HSTS is enabled, the
+client changes all requests from the HTTP URL to HTTPS before the request is
+sent, eliminating the need for a redirect. This is not required to be supported
+by the client, and can be disabled by setting `max-age=0`.
+
+[IMPORTANT]
+====
+HSTS works only with secure routes (either edge terminated or re-encrypt). The
+configuration is ineffective on HTTP or passthrough routes.
+====
+
+To enable HSTS to a route, add the `haproxy.router.openshift.io/hsts_header`
+value to the edge terminated or re-encrypt route:
+
+[source,yaml]
+----
+apiVersion: v1
+kind: Route
+metadata:
+  annotations:
+    haproxy.router.openshift.io/hsts_header: max-age=31536000;includeSubDomains;preload
+----
+
+[IMPORTANT]
+====
+Ensure there are no spaces and no other values in the parameters in the `haproxy.router.openshift.io/hsts_header` value. Only `max-age` is required.
+====
+
+The required `max-age` parameter indicates the length of time, in seconds, the
+HSTS policy is in effect for. The client updates `max-age` whenever a response
+with a HSTS header is received from the host. When `max-age` times out, the
+client discards the policy.
+
+The optional `includeSubDomains` parameter tells the client that all subdomains
+of the host are to be treated the same as the host.
+
+If `max-age` is greater than 0, the optional `preload` parameter allows external
+services to include this site in their HSTS preload lists. For example, sites
+such as Google can construct a list of sites that have `preload` set. Browsers
+can then use these lists to determine which sites to only talk to over HTTPS,
+even before they have interacted with the site. Without `preload` set, they need
+to have talked to the site over HTTPS to get the header.
--- a/architecture/networking/routes.adoc
+++ b/architecture/networking/routes.adoc
@@ -604,6 +604,8 @@ and "-". The default is the hashed internal key name for the route. |
 |`*haproxy.router.openshift.io/timeout*` | Sets a server-side timeout for the route. xref:time-units[(TimeUnits)] | `ROUTER_DEFAULT_SERVER_TIMEOUT`
 |`*router.openshift.io/haproxy.health.check.interval*`| Sets the interval for the back-end health checks. xref:time-units[(TimeUnits)] | `ROUTER_BACKEND_CHECK_INTERVAL`
 |`*haproxy.router.openshift.io/ip_whitelist*` | Sets a xref:whitelist[whitelist] for the route. |
+|`*haproxy.router.openshift.io/hsts_header*` | Sets a xref:hsts[Strict-Transport-Security] header for the edge terminated or re-encrypt route. |
+
 |===
 
 .A Route Setting Custom Timeout


admin_guide/managing_networking.adoc
```diff
@@ -1147,3 +1147,50 @@ objects:
       - podSelector: {}
 ...
 ----
+
+[[admin-guide-enabling-hsts]]
+== Enabling HTTP Strict Transport Security
+
+HTTP Strict Transport Security (HSTS) policy is a security enhancement, which
+ensures traffic to a host will always be HTTPS. This is useful for ensuring
```
pecameron: s/ensures traffic will always be HTTPS/requests traffic to the host to be HTTPS/

https tells the client, in a HTTPS response, that future requests should be HTTPS. The client decides whether or not to honor the request. So "ensures" is a little strong.

admin_guide/managing_networking.adoc
```diff
@@ -1147,3 +1147,50 @@ objects:
...
+secure interactions with websites, or to offer secure application for the user's
+benefit.
+
+HSTS adds a Strict Transport Security header to HTTPS responses from the site.
+The client (often a browser) reads the header and from then on enforces the
+policy. All requests to the HTTP URL are changed to HTTPS before the request is
+sent, eliminating the need for a redirect.
+
+When not enabled, any HTTP requests to a host that require HTTPS instead
```
pecameron: This redirect needs to be set up in the route. Its not automatic. By default the request is just dropped. So for this to work reasonably the redirect is set up so that the first time a client requests the target it get a redirect to HTTPS. on the next try, using HTTPS, the client gets the https header in the response. From then on it knows to just send HTTPS.

admin_guide/managing_networking.adoc
```diff
@@ -1147,3 +1147,50 @@ objects:
...
+results in the host generating a redirect to the HTTPS URL, or drops the
+request.
+
+To enable HSTS to a route, add the `haproxy.router.openshift.io/hsts_header`
```
pecameron: This only works on secure routes (edge terminated or re-encrypt). Adding the annotation to http or passthrough doesn't help.

admin_guide/managing_networking.adoc
```diff
@@ -1147,3 +1147,50 @@ objects:
...
+[[admin-guide-enabling-hsts]]
+== Enabling HTTP Strict Transport Security
+
+HTTP Strict Transport Security (HSTS) policy is a security enhancement, which
+ensures traffic to a host will always be HTTPS. This is useful for ensuring
+secure interactions with websites, or to offer secure application for the user's
+benefit.
+
+HSTS adds a Strict Transport Security header to HTTPS responses from the site.
+The client (often a browser) reads the header and from then on enforces the
```
pecameron: The client doesn't have to enforce the policy. Most do enforce it.

pecameron (changes requested):
HSTS is a suggestion from the server that the client should in the future use HTTPS. Clients do not have to respect the request, although most will since it eliminates a possible redirect.

admin_guide/managing_networking.adoc
```diff
@@ -1201,3 +1201,57 @@ objects:
...
+
+[[admin-guide-enabling-hsts]]
+== Enabling HTTP Strict Transport Security
+
+HTTP Strict Transport Security (HSTS) policy is a security enhancement, which
+requests that traffic to the host be HTTPS. This is useful for ensuring secure
+interactions with websites, or to offer secure application for the user's
+benefit.
+
+The default behavior is to drop any HTTP requests that require HTTPS. Enabling
```
knobunc: The default behavior of what?

admin_guide/managing_networking.adoc
```diff
@@ -1201,3 +1201,57 @@ objects:
...
+HSTS lets you set up a redirect so that the first time a client requests the
```
knobunc: Not quite.  All it does is add the header to https responses.  You can set the insecureEdgeTerminationPolicy to redirect to send http to https.

BUT if a browser has received an HSTS header from the site in the past, it will not send http traffic to the site and instead flip to https.  This part is what the change controls.

admin_guide/managing_networking.adoc
```diff
@@ -1201,3 +1201,57 @@ objects:
...
+target it is redirected to HTTPS. Any subsequent attempts will get a HTTPS
+header in the response, meaning future requests will be HTTPS.
+
+HSTS adds a Strict Transport Security header to HTTPS responses from the site.
+The client (often a browser) reads the header and, from then on, can enforce the
+policy. All requests to the HTTP URL are changed to HTTPS before the request is
```
knobunc: Might want to make it clear that the client changes from http to https.

admin_guide/managing_networking.adoc
```diff
@@ -1201,3 +1201,57 @@ objects:
...
+[NOTE]
+====
+To disable HSTS on a route, set `max-age=0`. Do not delete the annotation or
+route.
+====
+
+The optional `includeSubDomains` parameter tells the client that all subdomains
+of the host are to be treated the same as the host.
+
+If `max-age` is greater than 0, the optional `preload` parameter tells the
```
knobunc: Preload tells external services that they can include this site in their HSTS preload lists.  https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security#Preloading_Strict_Transport_Security

So, google keeps track of who has this set, and then builds a list of sites.  Browsers can consume these lists to work out what sites to only talk to over HTTPS without ever having talked to the site before.  Without preload, they need to have talked to the site over https to get the header... an attacker could theoretically proxy a client's traffic to the site to http and strip the header so that the https upgrade never happened.

admin_guide/managing_networking.adoc
```diff
@@ -1201,3 +1201,57 @@ objects:
...
+To enable HSTS to a route, add the `haproxy.router.openshift.io/hsts_header`
+value to the route object:
+
+[source,yaml]
+----
+apiVersion: v1
+kind: Route
+metadata:
+  annotations:
+    haproxy.router.openshift.io/hsts_header: max-age=31536000;includeSubDomains;preload
```
knobunc: Can we talk about the format of the value a little more.  We are very strict in what we accept.  There can be no spaces, an no other parameters.  But the order of the parameters is not fixed, and only max-age is required.

knobunc (changes requested):
Thanks Brice.  Nice first start.

admin_guide/managing_networking.adoc
pecameron: @knobunc The client is not required to support hsts. Most clients do respect it and will send https.

admin_guide/managing_networking.adoc
```diff
@@ -1201,3 +1201,57 @@ objects:
...
+
+HTTP Strict Transport Security (HSTS) policy is a security enhancement, which
+requests that traffic to the host be HTTPS. This is useful for ensuring secure
+interactions with websites, or to offer secure application for the user's
+benefit.
+
+The default behavior is to drop any HTTP requests that require HTTPS. Enabling
+HSTS lets you set up a redirect so that the first time a client requests the
+target it is redirected to HTTPS. Any subsequent attempts will get a HTTPS
+header in the response, meaning future requests will be HTTPS.
```
pecameron: "meaning future requests will be HTTPS", the response may have a max-age=0 which drops the policy for future requests.

admin_guide/managing_networking.adoc
pecameron: Also, the client is not required to support this and max-age=0 turns it off.

admin_guide/managing_networking.adoc
```diff
@@ -1201,3 +1201,57 @@ objects:
...
+sent, eliminating the need for a redirect.
+
+[IMPORTANT]
+====
+HSTS works only with secure routes (either edge terminated or re-encrypt). The
+configuration is ineffective on HTTP or passthrough routes.
+====
+
+To enable HSTS to a route, add the `haproxy.router.openshift.io/hsts_header`
+value to the route object:
```
pecameron: "to the route" -> "to the edge terminated or re-encrypt route.

admin_guide/managing_networking.adoc
pecameron: Also, clients often include a compiled in preload list.

architecture/networking/routes.adoc
```diff
@@ -604,6 +604,8 @@ and "-". The default is the hashed internal key name for the route. |
 |`*haproxy.router.openshift.io/timeout*` | Sets a server-side timeout for the route. xref:time-units[(TimeUnits)] | `ROUTER_DEFAULT_SERVER_TIMEOUT`
 |`*router.openshift.io/haproxy.health.check.interval*`| Sets the interval for the back-end health checks. xref:time-units[(TimeUnits)] | `ROUTER_BACKEND_CHECK_INTERVAL`
 |`*haproxy.router.openshift.io/ip_whitelist*` | Sets a xref:whitelist[whitelist] for the route. |
+|`*haproxy.router.openshift.io/hsts_header*` | Sets a xref:hsts[Strict-Transport-Security] header for the route. |
```
pecameron: "the route" ->> the edge terminated or re-encrypt route

pecameron (changes requested):

admin_guide/managing_networking.adoc
bfallonf: @knobunc @pecameron I've put in that it's OpenShift's default behaviour. Should it be the client's instead?

admin_guide/managing_networking.adoc
bfallonf: @knobunc @pecameron Using the `insecureEdgeTerminationPolicy` value doesn't seem to match up with the rest of the section. Is this an alternative to what's here?

bfallonf: I have:

"You can use the `insecureEdgeTerminationPolicy`
value in a route to redirect to send HTTP to HTTPS. However, when HSTS is
enabled, the client changes all requests from the HTTP URL to HTTPS before the
request is sent, eliminating the need for a redirect."

bfallonf:
Thanks @pecameron @knobunc . I've made some changes. Hopefully this alligns a little more with what you're after. Please let me know what you think, and let me know about the two questions above.

admin_guide/managing_networking.adoc
```diff
@@ -1207,18 +1207,16 @@ objects:
...
-The default behavior is to drop any HTTP requests that require HTTPS. Enabling
-HSTS lets you set up a redirect so that the first time a client requests the
-target it is redirected to HTTPS. Any subsequent attempts will get a HTTPS
-header in the response, meaning future requests will be HTTPS.
-
-HSTS adds a Strict Transport Security header to HTTPS responses from the site.
-The client (often a browser) reads the header and, from then on, can enforce the
-policy. All requests to the HTTP URL are changed to HTTPS before the request is
-sent, eliminating the need for a redirect.
+The {product-title} default behavior is to drop any HTTP requests that require
```
knobunc: Actually, this first sentence doesn't really make sense either.  That behavior is controlled by a separate setting.

admin_guide/managing_networking.adoc
```diff
@@ -1207,18 +1207,16 @@ objects:
...
+HTTPS. When HSTS is enabled, HSTS adds a Strict Transport Security header to
+HTTPS responses from the site. You can use the `insecureEdgeTerminationPolicy`
+value in a route to redirect to send HTTP to HTTPS. However, when HSTS is
+enabled, the client changes all requests from the HTTP URL to HTTPS before the
+request is sent, eliminating the need for a redirect. This is not required to be
+supported by the client, and can be disable by setting `max-age=0`.
```
knobunc: disabled

knobunc (changes requested):

bfallonf:
[rev_history]
|Added the xref:../admin_guide/managing_networking.adoc#admin-guide-enabling-hsts[Enabling HTTP Strict Transport Security] section.
%

bfallonf:
Thanks @knobunc . I've updated this with the suggestions. @openshift/team-documentation peer review plz

admin_guide/managing_networking.adoc
```diff
@@ -1201,3 +1201,57 @@ objects:
             name: default
 ...
 ----
+
+[[admin-guide-enabling-hsts]]
+== Enabling HTTP Strict Transport Security
+
+HTTP Strict Transport Security (HSTS) policy is a security enhancement that
```
gaurav-nelson: Suggestions:
```
HTTP Strict Transport Security (HSTS) policy is a security enhancement which
only allows HTTPS traffic on the host. Any HTTP requests are dropped by
default. This is useful for ensuring secure interactions with websites, or to
offer a secure application for the user's benefit.

When HSTS is enabled, HSTS adds a Strict Transport Security header to HTTPS
responses from the site. You can use...
```

gaurav-nelson:
@bfallonf One suggestion, LGTM otherwise.  :horse:

gaurav-nelson:
Please squash and rebase as well.

PR #6418, opened 2017-11-19
Update builddefaults documentation.

barlik:
Update the documentation with the instructions how to keep the proxy settings, but disable it for the builds.

### DIFF
--- a/install/configuring_inventory_file.adoc
+++ b/install/configuring_inventory_file.adoc
@@ -1212,35 +1212,39 @@ option.
 
 |`openshift_builddefaults_http_proxy`
 |This variable defines the `HTTP_PROXY` environment variable inserted into
-builds using the `BuildDefaults` admission controller. If
-`openshift_http_proxy` is set, this variable will inherit that value; you only
-need to set this if you want your builds to use a different value.
+builds using the `BuildDefaults` admission controller. If you do not define
+this parameter but define the `openshift_http_proxy` parameter, the
+`openshift_http_proxy` value is used. Set the
+`openshift_builddefaults_http_proxy` value to `False` to disable default
+http proxy for builds regardless of the `openshift_http_proxy` value.
 
 |`openshift_builddefaults_https_proxy`
-|This variable defines the `*HTTPS_PROXY*` environment variable inserted into
-builds using the `*BuildDefaults*` admission controller. If
-`*openshift_https_proxy*` is set, this variable will inherit that value; you
-only need to set this if you want your builds to use a different value.
+|This variable defines the `HTTPS_PROXY` environment variable inserted into
+builds using the `BuildDefaults` admission controller. If you do not define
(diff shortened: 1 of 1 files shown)


vikram-redhat:
I am closing this PR as there is not enough context provided around why we need to change it. @barlik please feel free to submit a new one if this change needs to be made.

barlik:
It's difficult to be motivated to create a new PR when the current one was ignored for 10 months and then closed without asking any questions. The context should be evident from the commit diff.
At a customer, we needed to set `openshift_http_proxy`, but not `openshift_builddefaults_http_proxy`. The only way to achieve that is by setting `openshift_builddefaults_http_proxy` to `false`. That was not easy to figure out, hence the documentation update.

vikram-redhat:
@barlik thanks for that added context and I apologize for the delay in getting to this.

@jiajliu - can you please provide QE review?

@kalexand-rh - once we get QE review, can you PTAL?

vikram-redhat:
@barlik - can you please rebase as well?

kalexand-rh:
@jiajliu, @barlik, which versions is this change valid in?

@vikram-redhat, this file's been moved, so, depending on the valid versions, we'll probably need multiple unique PRs before it's over.

kalexand-rh (commented):
I edited one of the definitions so it's clearer. Will you please update them?

barlik:
I've rebased onto master branch and updated the documentation.

@kalexand-rh I've seen the behaviour in release 3.5 and I've just confirmed it with the master branch.

I would also like to point out that I'm not sure that the variable `builddefaults` should be inherited in the first place and setting it to `False` is probably an accidental 'solution' for the problem. So, please can anybody check with the engineers?

kalexand-rh:
Thanks @barlik!

@sdodson, do you have an opinion on these `builddefaults` parameters?

sdodson:
These changes look fine to me. I don't think there would be any changes that would prevent this from working in later releases. Thanks.

kalexand-rh:
Thank you @sdodson!

@jiajliu, will you PTAL or suggest another QE reviewer?

kalexand-rh:
@jianlinliu, will you PTAL?

jianlinliu:
@ganhuang PTAL.

ganhuang:
Tested with `False` set against those parameters, it definitely works.

LGTM. But it would be a little confusing that `false` won't work in such situation. It's be perfect if we could give a note here.

kalexand-rh:
@ganhuang, thanks for confirming this! I think this whole flow is a little confusing. Will you please give me a little more information about what you'd like in the note?

ganhuang:
@kalexand-rh I don't know how to change that words. Not a big deal. I'm fine to get it merged.

kalexand-rh:
Thanks, @ganhuang!

PR #6461, opened 2017-11-22
Add examples of using Secrets for Environment Variables in Build Configs

bparry02:


vikram-redhat:
@mburke5678 - PTAL.

bparry02:
@mburke5678 Thanks for reviewing. In the field we encounter many different cluster versions, so adding it to all applicable versions is really helpful, if possible. My records have it applicable to OCP 3.6 as well.

mburke5678:
@wzheng1 BUMP ^^

wzheng1:
LGTM, thanks! Sorry for late.

mburke5678:
@wzheng1 Thank you for the review!

mburke5678:
@simo5 Can you please take a look at this PR, a change from consulting related to using `valueFrom` to use build environment variables with secrets. There are several places in the topic where `valueFrom` has warnings. Do these apply here?

Jenkins Pipeline strategy does not support valueFrom 
Referencing container resources using valueFrom in build environment variables is not supported

dev_guide/builds/build_environment.adoc
```diff
@@ -51,3 +51,26 @@ variables.
...
+  name: secret-example-bc
+spec:
+  strategy:
+    sourceStrategy:
+      env:
+      - name: MYVAL
+        valueFrom:
+          secretKeyRef:
+            key: myval
+            name: mysecret
```
bparees: @deads2k @liggitt how much does it bother you that someone who can create a build in a namespace, can use this mechanism to access any secret in the namespace?  (because the build controller resolves this secret value on your behalf).

wondering if we need to disable the ability to make these references or at least introduce some sort of sar checking based on the builder service account's permissions.

bparees:
the docs look ok to me but i'd feel better knowing someone has tested this?

bparees:
>  There are several places in the topic where valueFrom has warnings. Do these apply here?

> Jenkins Pipeline strategy does not support valueFrom

applicable but i don't know if you need a secondary warning for it.  The one in the overview should suffice.

> Referencing container resources using valueFrom in build environment variables is not supported

irrelevant here since you're not referencing container resources.

dev_guide/builds/build_environment.adoc
liggitt: this is similar to the power to create a replication controller, deployment, deployment config, etc. the default roles that allow creating builds also allow creating those other workload resources.

bparees: ok, sounds like we're fine then, thanks

mburke5678:
@bparry02 @wzheng1 Can you reply to [Ben's comment above](https://github.com/openshift/openshift-docs/pull/6461#issuecomment-411158149)?

wzheng1:
QE has verified this function and it works as expected, below ones work too：
fieldRef, configMapKeyRef and secretKeyRef

mburke5678:
@bparry02 If you can please squash the commits. I can merge this PR. Which OpenShift versions does this apply to?

mburke5678:
@bparees @liggitt 
Which OpenShift versions does this apply to?

bparees:
3.6 and above.

mburke5678:
@bparry02  If you can please squash the commits.

bparry02:
@mburke5678 commits squashed. Thanks.

PR #6540, opened 2017-11-29
fixed ifdef statement for Online

ahardin-rh:
@vikram-redhat This fixes the issue in the Online docs that Radka caught.


ahardin-rh:
No revision history needed

PR #6562, opened 2017-11-30
Fixed typo

trescst:
I assume that `openshift_enable_service_catalog` should be set to `false` to disable the automatic deployment of the Service Catalog.  This PR fixes that.


adellape:
:+1: Thanks!

PR #6575, opened 2017-11-30
cli ref extend cli: correct plugin name

sosiouxme:
The example yaml actually defined a plugin named `name1`. The fact that it's located in a directory called `great-plugin` is irrelevant. I also though an example usage would help.


ahardin-rh:
@sosiouxme Thank you!

ahardin-rh:
No revision history needed

PR #6596, opened 2017-12-01
Bug 1457945, Clarified default behavior in Garbage Collection docs

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1457945

Preview Build:
http://file.rdu.redhat.com/~ahardin/Dec012017/garbage-collection-bz/admin_guide/garbage_collection.html

### DIFF
--- a/admin_guide/garbage_collection.adoc
+++ b/admin_guide/garbage_collection.adoc
@@ -15,13 +15,20 @@ toc::[]
 The {product-title} node performs two types of garbage collection:
 
 * xref:container-garbage-collection[Container garbage collection]: Removes
-terminated containers. Typically run every minute.
+terminated containers.
 * xref:image-garbage-collection[Image garbage collection]: Removes images not
-referenced by any running pods. Typically run every five minutes.
+referenced by any running pods.
 
 [[container-garbage-collection]]
-
 == Container Garbage Collection
+
+Container garbage collection is enabled by default and happens automatically in
+response to eviction thresholds being reached. The node tries to keep any
+container for any pod accessible from the API. If the pod has been deleted, the
+containers will be as well. Containers are preserved as long the pod is not
+deleted and the eviction threshold is not reached. If the node is under disk
+pressure, it will remove containers and their logs will no longer be accessible
+via `oc logs`.
 
 The policy for container garbage collection is based on three node settings:
 
@@ -32,14 +39,14 @@ The policy for container garbage collection is based on three node settings:
 
 |`*minimum-container-ttl-duration*`
 |The minimum age that a container is eligible for garbage collection. The
-default is *1m* (one minute). Use *0* for no limit. Values for this setting can be
+default is *0*. Use *0* for no limit. Values for this setting can be
 specified using unit suffixes such as *h* for hour, *m* for minutes, *s* for seconds.
 
 |`*maximum-dead-containers-per-container*`
-|The number of instances to retain per pod container. The default is *2*.
+|The number of instances to retain per pod container. The default is *1*.
 
 |`*maximum-dead-containers*`
-|The maximum number of total dead containers in the node. The default is *240*.
+|The maximum number of total dead containers in the node. The default is *-1*, which means unlimited.
 |===
 
 The `*maximum-dead-containers*` setting takes precedence over the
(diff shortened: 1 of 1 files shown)


ahardin-rh:
@sjenning @sdodson PTAL

admin_guide/garbage_collection.adoc
```diff
@@ -15,14 +15,20 @@ toc::[]
...
 
 [[container-garbage-collection]]
-
 == Container Garbage Collection
 
+Container garbage collection is enabled by default and happens automatically in
+response to eviction thresholds being reached. The node tries to keep any pod
+that is still accessible from the API. If the pod has been deleted, the
+containers will be as well. Containers are only deleted when the pod object is
+gone from {product-title}. If the node is under disk pressure, it will remove
```
sjenning: I would say "Containers are preserved as long the pod is not deleted and there an eviction threshold is not reached."

admin_guide/garbage_collection.adoc
```diff
@@ -32,14 +38,14 @@ The policy for container garbage collection is based on three node settings:
...
+default is *0*. Use *0* for no limit. Values for this setting can be
 specified using unit suffixes such as *h* for hour, *m* for minutes, *s* for seconds.
 
 |`*maximum-dead-containers-per-container*`
-|The number of instances to retain per pod container. The default is *2*.
+|The number of instances to retain per pod container. The default is *1*.
 
 |`*maximum-dead-containers*`
-|The maximum number of total dead containers in the node. The default is *240*.
+|The maximum number of total dead containers in the node. The default is *-1*.
```
sjenning: Might indicated that -1 means unlimited

Also all 3 of these flags are deprecated.

propyless: What are the new flags?

sjenning: https://kubernetes.io/docs/tasks/administer-cluster/out-of-resource/#deprecation-of-existing-feature-flags-to-reclaim-disk

ahardin-rh: @sjenning Hmm so if these are all deprecated and have no direct replacements according to that link you shared, I'm not sure what to do. It was indicated that the docs mostly need to have these values updated, but I'm not sure when the flags became deprecated and what versions of our docs they need to be removed from.

sjenning: @ahardin-rh they are deprecated at least as far back as 3.6, possibly 3.5 as well.  @derekwaynecarr do you know when eviction deprecated these flags?

ahardin-rh: @derekwaynecarr Any guidance here? I want to be sure that when we update the docs across versions, we're not including any outdated information, including deprecated flags. Thanks!

vikram-redhat: @sjenning is there anyone else who can answer this other than @derekwaynecarr ?

ahardin-rh: @sjenning @derekwaynecarr PTAL. I'm currently blocked and not sure what is needed here.

admin_guide/garbage_collection.adoc
```diff
@@ -15,14 +15,21 @@ toc::[]
...
 * xref:image-garbage-collection[Image garbage collection]: Removes images not
-referenced by any running pods. Typically run every five minutes.
+referenced by any running pods.
 
 [[container-garbage-collection]]
-
 == Container Garbage Collection
 
+Container garbage collection is enabled by default and happens automatically in
+response to eviction thresholds being reached. The node tries to keep any pod
```
derekwaynecarr: node tries to keep any container for any pod accessible from the API

admin_guide/garbage_collection.adoc
derekwaynecarr: the flags were deprecated in kube 1.4, but still remain through kube 1.10.

we should continue to document them as I do not see them being removed for the next 6 mos.

derekwaynecarr:
one nit, LGTM

ahardin-rh:
Thanks! Updated.

ahardin-rh:
[rev_history]
|xref:../admin_guide/garbage_collection.adoc#admin-guide-garbage-collection[Garbage Collection]
|Added clarifying details about the default behavior of garbage collection.
%

PR #6609, opened 2017-12-05
Rename openshift_hosted_router_selector

juanluisvaladas:
The variable openshift_hosted_router_selector is now openshift_router_selector. In this same
document openshift_router_selector is already used.

Reference:
https://github.com/openshift/openshift-ansible/commit/4b0f5219f4ec3eeb5433fc9f985b1e067b581276

### DIFF
--- a/install_config/install/advanced_install.adoc
+++ b/install_config/install/advanced_install.adoc
@@ -484,7 +484,7 @@ invalid if used. These values override other settings in node configuration
 which may cause invalid configurations. Example usage:
 *{'image-gc-high-threshold': ['90'],'image-gc-low-threshold': ['80']}*.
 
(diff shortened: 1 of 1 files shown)


juanluisvaladas:
bump, @ahardin-rh maybe?

ahardin-rh:
@juanluisvaladas Thank you! I see another instance used in https://docs.openshift.com/container-platform/3.7/install_config/router/default_haproxy_router.html, so I will create a companion PR to address that usage. 

I have this labeled for OCP 3.7 docs and later. Please let me know if this also applies to any earlier versions.

juanluisvaladas:
Hi, only 3.7 and later is correct

ahardin-rh:
No revision history needed.

PR #6020, opened 2017-10-30
Add pages for new OpenShift on OpenStack provider

bogdando:
The new shift-on-stack provider [0] complements and will replace some of
the existing documentation for OCP on OSP 10 reference architecture.
It is planned to be the supported solution in the OSP 13 time frame.

Note that the provider is being moved under the openshift-ansible
repo [1]. Its link must be updated after the documentation moved as
the part of that change request.
"Configuring OpenShift Origin for OpenStack with Ansible" will likely
need some updates as well.

[0] https://github.com/openshift/openshift-ansible-contrib
[1] https://github.com/openshift/openshift-ansible/pull/6039

Changes:

* Clarify on-premises and cloud providers cases in
  the Getting Started, Planning and Advanced Installation pages for
  the openshift-origin/enterprise distros.
* Link documentation pages for the shift-on-stack provider, which
  is currently hosted in-repo docs.
* Add a link and a deprecation note for OCP 2.6 on OSP 10 reference
  architecture based on Heat Templates.

Highlight other cloud offerings as well.
Describe alternative cloud solitions in the planning OpenShift
Enterprise installation pages. For some cases, OpenShift Dedicated
or Online might be a better fit. This makes things more clear for
users from the first docs pages read.

Additionally, highlight the Dedicated alternative, when describing
the cloud hosted infrustracture configuration prerequisites in the
Enterprise advanced installation guide.

Depends-on https://github.com/openshift/openshift-ansible/pull/6039

Signed-off-by: Bogdan Dobrelya <bdobreli@redhat.com>

PR #6614, opened 2017-12-05
added missing endif

mburke5678:
An endif was missing from text added via https://github.com/openshift/openshift-docs/pull/6020


mburke5678:
Just needs to be merged to master. Original PR was merged to master only and picked up by origin

PR #6615, opened 2017-12-05
Add system status badge documentation to web console config

rhamilto:
Words are hard.  ;-)  @jwforres, I kept this in line with what we do for online, but maybe it should be more generic?

### DIFF
--- a/install_config/web_console_customization.adoc
+++ b/install_config/web_console_customization.adoc
@@ -505,6 +505,69 @@ ifdef::openshift-enterprise[]
 endif::[]
 ----
 
+[[system-status-badge]]
+=== System Status Badge
+
+The top navigation bar can also include an optional system status badge in order
+to notify users of system-wide events such as maintenance windows. To make use
+of the existing styles utilizing a yellow warning icon for the badge,
+follow the example below.
+
+. Create the configuration scripts within a file (for example,
+*_systemStatusBadge.js_*):
++
+[source, javascript]
+----
+'use strict';
+
+angular
+  .module('mysystemstatusbadgeextension', ['openshiftConsole'])
+  .run([
+    'extensionRegistry',
+    function(extensionRegistry) {
+      // Replace http://status.example.com/ with your domain
+      var system_status_elem = $('<a href="http://status.example.com/"' +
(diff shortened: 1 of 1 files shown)


jwforres:
I think it would be good to explain what is generally supported, and then we can say this is an example of using statuspage.io to generate the system status

so explain the extension-registry extension point that is supported, and that if they want to use our existing styles they should use such and such markup and give an html snippet

rhamilto:
Take 2.  Attempted to improve the description and opted to make the example very simple so it's just an icon and configurable url.  Better, @jwforres?

jwforres:
I'm fine with that, it does leave out the "5 issues" text but im not sure how helpful that is anyway.

install_config/web_console_customization.adoc
```diff
@@ -505,6 +505,67 @@ ifdef::openshift-enterprise[]
...
+  .run([
+    'extensionRegistry',
+    function(extensionRegistry) {
+      // Replace http://status.example.com/ with your domain
+      var system_status_elem = $('<a href="http://status.example.com/"' +
+      'target="_blank" class="nav-item-iconic system-status"><span title="' +
+      'System Status" class="fa status-icon pficon-warning-triangle-o">' +
+      '</span></a>');
+
+      extensionRegistry
```
jwforres: may want a comment here that says you should only add the extension point if you want the status icon to show up, a typical extension will optionally add this based on the existence of issues after hitting a status API.

rhamilto:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1519738

Attention @ahardin-rh

jwforres:
@ahardin-rh we will want to get it into the 3.7 docs as well i think

rhamilto:
@ahardin-rh, can this be merged so I can move https://bugzilla.redhat.com/show_bug.cgi?id=1519738 to MODIFIED?

ahardin-rh:
@rhamilto Yes, so sorry for the delay. Not sure how I missed this. 😱

rhamilto:
@ahardin-rh, no worries.  Thanks!

ahardin-rh:
Revision history stated in follow-up PR: https://github.com/openshift/openshift-docs/pull/7093

PR #6621, opened 2017-12-05
clarify image-pusher role

### LINKED ISSUE #17341: Role system:image-pusher does not allow to push images.
The clusterrole system:image-pusher does NOT allow to push images.

##### Version
oc version
oc v3.6.173.0.49
kubernetes v1.6.1+5115d708d7
features: Basic-Auth GSSAPI Kerberos SPNEGO

Server https://lb.karmalabs.local:8443
openshift v3.6.173.0.49
kubernetes v1.6.1+5115d708d7

In OCP 3.5 does not work neither.

##### Steps To Reproduce
1. Give a user the role system:image-pusher
oc adm policy add-cluster-role-to-user system:image-pusher "USER"

2. Log into oc with that user
oc login "USER"..

3. Get user token
token=$(oc whoami -t)

4. Log into the registry with Docker cli
docker login <registry> -u "USER" -p $token

5. Try to push an image
docker push...

##### Current Result
"**denied: requested access to the resource is denied**" 

docker push myregistry.example.com/test-push/test-push:latest                                                                                                                                      
The push refers to a repository [myregistry.example.com/test-push/test-push]                                                                                                                         
a11fda45870e: Mounted from test-push/test-push                                                                                                                                                           
**denied: requested access to the resource is denied**

##### Expected Result

The docker push command to ends with no errors ;)

bparees:
fixes https://github.com/openshift/origin/issues/17341

### DIFF
--- a/dev_guide/managing_images.adoc
+++ b/dev_guide/managing_images.adoc
@@ -1052,9 +1052,17 @@ Before performing this procedure, the following must be satisfied:
 
 - The destination project you push to must already exist.
 - The user must be authorized to `{get, update} "imagestream/layers"` in that
-project. The *system:image-pusher* role can be added to a user to provide these
-permissions. If you are a project administrator, then you would also have these
-permissions.
+project. In addition, since the image stream does not already exist, the user 
+must be authorized to `{create} "imagestream"` in that project.  If you are a project 
+administrator, then you would have these permissions.
+
+[NOTE]
+====
+The *system:image-pusher* role does not grant permission to create new image streams,
+only to push images to existing image streams, so it cannot be used to push images
+to image streams that do not yet exist unless additional permissions are also granted to 
+the user.
+====
 
 To create an image stream by manually pushing an image:
 


bparees:

bparees:
applicable to all releases afaik

mburke5678:
LGTM 
The _{create} "imagestream"_ construct looks odd. Think users will understand what that means? (I assume _oc create imagestream_). Maybe we should break out all those commands?? Like the following (which looks bad here, but can be cut and paste into the .adoc file)

`- The user must be authorized to use the following commands in the project:`
+
--
* `oc get imagestream`
* `oc get layers`
* `oc update imagestream`
* `oc update layers` 

In addition, since the image stream does not already exist, the user 
must be authorized to `oc create imagestream` in that project.  If you are a project 
administrator, then you would have these permissions.
--`

bparees:
> The {create} "imagestream" construct looks odd. Think users will understand what that means? (I assume oc create imagestream).

no, it's not oc create imagestream.  It means the ability to perform the create verb on the imagestream resource.  I copied the syntax from what is already there for imagestream/layers. (`{get, update} "imagestream/layers"`)

PR #6662, opened 2017-12-08
Revert advanced audit techpreview drop from 3.7

soltysh:
This needs to land for 3.7, since `audit/v1beta1` is not available there and audit is in tech preview. The tech preview is dropped from 3.8 only. 

@mburke5678 @openshift/team-documentation


soltysh:
The second commit also fixes the formatting which was part of the reverted commit, but most importantly it fixes the order of policy rules. I'll open a master pr shortly with this.

PR #6669, opened 2017-12-09
Update NetworkPolicy objects to use v1 API

danwinship:
In OCP 3.9 we will have kube 1.9, with NetworkPolicy v1, so we should update the examples to use the new API version.

(This change is correct for 3.8 as well if we are doing 3.8 docs. We don't want it for 3.7 or earlier.)


bfallonf:
LGTM. Thanks @danwinship . I'll merge and put into the 3.9 docs (no 3.8).

bfallonf:
[rev_history]
|Updated `NetworkPolicy` examples out of beta.
%

PR #6861, opened 2017-12-18
OCP 3.6 typo fixes

ahardin-rh:


ahardin-rh:
This is a follow-up to https://github.com/openshift/openshift-docs/pull/6851 since the cherry-picking was messy from master.

PR #6892, opened 2017-12-19
Fix getting openshift facts.

jmferrer:
**ansible-playbook /usr/share/ansible/openshift-ansible/roles/openshift_facts/library/openshift_facts.py** returns
```ERROR! Syntax Error while loading YAML.```
because openshift_facts.py is a python program.

This other way returns facts.


ahardin-rh:
Thanks and sorry for the delay. Closing this in favor of https://github.com/openshift/openshift-docs/pull/9139

PR #6966, opened 2018-01-02
openshift_hosted_router_selector is now openshift_router_selector

ahardin-rh:
This is a companion PR to https://github.com/openshift/openshift-docs/pull/6609


ahardin-rh:
No revision history needed

PR #6576, opened 2017-11-30
Added a graph driver feature comparison

ahardin-rh:
https://trello.com/c/NU7G0Kzd/409-5-docs37-ocp-add-a-section-on-docker-graph-driver-feature-comparison

Preview Build:
http://file.rdu.redhat.com/~ahardin/11302017/graph-driver-comparison/scaling_performance/optimizing_storage.html#choosing-a-docker-graph-driver

PR #6973, opened 2018-01-02
Removed unnecessary line in code block

ahardin-rh:
Removed per discussion with @jeremyeder in https://github.com/openshift/openshift-docs/pull/6576


ahardin-rh:
No revision history needed.

PR #6982, opened 2018-01-03
Removed Router Cipher Suite section (OCP3.4)

gaurav-nelson:
For https://bugzilla.redhat.com/show_bug.cgi?id=1524486
- Replaced **Router Cipher Suite** section with a note. (This section was only intended for 3.6 on-ward)


gaurav-nelson:
Minor change no revision history needed.

PR #7009, opened 2018-01-05
Fix tiny typo in other_api_objects.adoc

pabrahamsson:


ahardin-rh:
LGTM. Thank you!

ahardin-rh:
No revision history needed.

vikram-redhat:
@ahardin-rh - this is not in dedicated 3.6 stage. Is that right?

ahardin-rh:
@vikram-redhat Good catch. Not sure how that was skipped, but here's the PR: https://github.com/openshift/openshift-docs/pull/7041

vikram-redhat:
@ahardin-rh - np. I wasn't sure if it was just missed or it didn't apply. In any case, it can get published next week.

PR #7010, opened 2018-01-05
Added new Managing Huge Pages topic

ahardin-rh:
Preview Build: http://file.rdu.redhat.com/~ahardin/01052018/hugepages/scaling_performance/managing_hugepages.html  

https://trello.com/c/steXZKC4/638-5-psap-hugepages-documentation

https://trello.com/c/zGc0NTZE/756-scalehugepages-documentation

### DIFF
--- a/scaling_performance/managing_hugepages.adoc
+++ b/scaling_performance/managing_hugepages.adoc
@@ -1,0 +1,149 @@
+[[scaling-performance-managing-huge-pages]]
+= Managing Huge Pages
+{product-author}
+{product-version}
+:data-uri:
+:icons:
+:experimental:
+:toc: macro
+:toc-title:
+:prewrap!:
+
+toc::[]
+
+
+== What Huge Pages Do
+
+[IMPORTANT]
+====
+Managing huge pages is a Technology Preview feature only.
+ifdef::openshift-enterprise[]
+Technology Preview features are not supported with Red Hat production service
+level agreements (SLAs), might not be functionally complete, and Red Hat does
+not recommend to use them for production. These features provide early access to
+upcoming product features, enabling customers to test functionality and provide
+feedback during the development process.
+
+For more information on Red Hat Technology Preview features support scope, see
+https://access.redhat.com/support/offerings/techpreview/.
+endif::[]
+====
+
+Memory is managed in blocks known as pages. On most systems, a page is 4Ki. 1Mi
+of memory is equal to 256 pages; 1Gi of memory is 256,000 pages, and so on. CPUs
+have a built-in memory management unit that manages a list of these pages in
+hardware. The Translation Lookaside Buffer (TLB) is a small hardware cache of
+virtual-to-physical page mappings. If the virtual address passed in a hardware
+instruction can be found in the TLB, the mapping can be determined quickly. If
+not, a TLB miss occurs, and the system falls back to slower, software-based
+address translation, resulting in performance issues. Since the size of the
+TLB is fixed, the only way to reduce the chance of a TLB miss is to increase the
+page size.
+
+A huge page is a memory page that is larger than 4Ki. On x86_64 architectures,
+there are two common huge page sizes: 2Mi and 1Gi. Sizes vary on other
+architectures. In order to use huge pages, code must be written so that
+applications are aware of them. Transparent Huge Pages (THP) attempt to automate
+the management of huge pages without application knowledge, but they have
+limitations. In particular, they are limited to 2Mi page sizes. THP can lead to
+performance degradation on nodes with high memory utilization or fragmentation
+due to defragmenting efforts of THP, which can lock memory pages. For this
+reason, some applications may be designed to (or recommend) usage of
+pre-allocated huge pages instead of THP.
+
+In {product-title}, applications in a pod can allocate and consume pre-allocated
+huge pages. This topic describes how.
+
+[[huge-pages-prerequisites]]
+== Prerequisites
+
+. Nodes must pre-allocate huge pages in order for the node to report its huge
+ page capacity. A node can only pre-allocate huge pages for a single size.
+
+. `HugePages` must be set to `true` across the system:
++
+----
+# cat /etc/origin/node/node-config.yaml
+...
+kubeletArguments:
+...
+  feature-gates:
+  - HugePages=true
+
+# systemctl restart atomic-openshift-node
+----
+
+. The master API must be enabled for `feature-gates` with `HugePages=true`:
++
+----
+# cat /etc/origin/master/master-config.yaml
+...
+kubernetesMasterConfig:
+  apiServerArguments:
+  ...
+  feature-gates:
+  - HugePages=true
+
+# systemctl restart atomic-openshift-master-api
+----
+
+[[consuming-huge-pages]]
+== Consuming Huge Pages
+
+Huge pages can be consumed via container level resource requirements using the
+resource name `hugepages-<size>`, where size is the most compact binary
+notation using integer values supported on a particular node. For example, if a
+node supports 2048KiB page sizes, it will expose a schedulable resource
+`hugepages-2Mi`. Unlike CPU or memory, huge pages do not support overcommitment.
+
+----
+kind: Pod
+metadata:
+  generateName: hugepages-volume-
+spec:
+  containers:
+  - securityContext:
+      privileged: true
+    image: rhel7:latest
+    command:
+    - sleep
+    - inf
+    name: example
+    volumeMounts:
+    - mountPath: /hugepages
+      name: hugepage
+    resources:
+      limits:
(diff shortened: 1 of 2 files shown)


ahardin-rh:
@jeremyeder PTAL

scaling_performance/managing_hugepages.adoc
```diff
@@ -0,0 +1,87 @@
...
+[[huge-pages-prerequisites]]
+== Prerequisites
+
+. Nodes must pre-allocate huge pages in order for the node to report its huge
+ page capacity. A node can only pre-allocate huge pages for a single size.
+
+. `HugePages` must be set to `true` across the system:
++
+----
+--feature-gates=HugePages=true
```
jeremyeder: This is the kubelet argument format. but in openshift we have the yaml-based kubelet config. please refer back to the cpu manager docs for an example of how to set a feature gate in the node-config.yaml.  Then replace this line with the yaml example.

scaling_performance/managing_hugepages.adoc
```diff
@@ -0,0 +1,87 @@
...
+`hugepages-2Mi`. Unlike CPU or memory, huge pages do not support overcommitment.
+
+----
+apiVersion: v1
+kind: Pod
+metadata:
+  generateName: hugepages-volume-
+spec:
+  containers:
+  - image: fedora:latest
```
jeremyeder: probably make this rhel7

scaling_performance/managing_hugepages.adoc
```diff
@@ -0,0 +1,87 @@
...
+    command:
+    - sleep
```
jeremyeder: the cpu manager examples use the pause pod.  we should probably unify these examples.  i suppose sleep infinity is the easiest since it is in the rhel7 base image.

scaling_performance/managing_hugepages.adoc
```diff
@@ -0,0 +1,87 @@
...
+    - inf
+    name: example
+    volumeMounts:
+    - mountPath: /hugepages
+      name: hugepage
+    resources:
+      limits:
+        hugepages-2Mi: 100Mi
```
jeremyeder: I think we need a call-out box related to this line.  Users of hugepages are used to specifying the amount of memory for hugepages a quantity of hugepages, multiplied by the size of the page.  They do this intuitively and it's been around for decades.

What we are doing here is different. We are allowing the user to specify the amount of memory for hugepages as the exact amount to be allocated.  So there is no "multiplied by the size of the page" step mental calculation.

In other words, given a hugepage size of 2MB, if a user wanted to use 100MB of hugepage-backed RAM for their application, they would allocate 50 hugepages.  Above you see we handle the math for users, allowing them to specify 100MB directly.

scaling_performance/managing_hugepages.adoc
```diff
@@ -0,0 +1,87 @@
...
+----
+
+* Huge page requests must equal the limits. This is the default if limits are
+specified, but requests are not.
+* Huge pages are isolated at a pod scope, container isolation is planned in a
+future iteration.
+* `EmptyDir` volumes backed by huge pages must not consume more huge page memory
+than the pod request.
+* Applications that consume huge pages via `shmget()`` with `SHM_HUGETLB` must run
+with a supplemental group that matches *_proc/sys/vm/hugetlb_shm_group_*.
```
jeremyeder: We probably need an example of this.

jeremyeder (changes requested):

jeremyeder:
I also want to add some context as to why this exists (from the original hugepages design doc for kube).  I will take a stab at that next week.

jeremyeder:
@derekwaynecarr fyi, have you an example of the supplemental group part?

ahardin-rh:
@jeremyeder I made some updates and yoinked some of Derek's content on the purpose. 

Updated Preview Build (You may need to clear cache to see latest changes): http://file.rdu.redhat.com/~ahardin/01052018/hugepages/scaling_performance/managing_hugepages.html

ahardin-rh:
@derekwaynecarr ^^

scaling_performance/managing_hugepages.adoc
```diff
@@ -0,0 +1,126 @@
...
+node supports 2048KiB page sizes, it will expose a schedulable resource
+`hugepages-2Mi`. Unlike CPU or memory, huge pages do not support overcommitment.
+
+----
+apiVersion: v1
+kind: Pod
+metadata:
+  generateName: hugepages-volume-
+spec:
+  containers:
```
derekwaynecarr: add the following:

```
containers:
 securityContext:
  privileged: true
```
I suspect on the majority of machines in the world will report `cat /proc/sys/vm/hugetlb_shm_group` as `0`, so you effectively need to run privileged right now.

ahardin-rh:
@derekwaynecarr Updated with the latest. Thanks for your help!

ahardin-rh:
@jeremyeder @derekwaynecarr Is this good to go for 3.9? Are any further changes needed to the docs? Thanks!

jeremyeder:
lgtm

ahardin-rh:
@wjiangjay Can you please give a QE review? Thanks!

scaling_performance/managing_hugepages.adoc
```diff
@@ -0,0 +1,128 @@
...
+    securityContext:
+      privileged: true
+  - image: rhel7:latest
```
wjiangjay: here should be 
```
kind: Pod
metadata:
  generateName: hugepages-volume-
spec:
  containers:
  - securityContext:
      privileged: true
    image: rhel7:latest
    command:
    - sleep
    - inf
    name: example
    volumeMounts:
    - mountPath: /hugepages
      name: hugepage
    resources:
      limits:
        hugepages-2Mi: 100Mi 
  volumes:
  - name: hugepage
    emptyDir:
      medium: HugePages
```

scaling_performance/managing_hugepages.adoc
```diff
@@ -0,0 +1,128 @@
...
++
+----
+# cat /etc/origin/node/node-config.yaml
+...
+kubeletArguments:
+...
+  feature-gates:
+  - HugePages=true
+
+# systemctl restart atomic-openshift-node
```
wjiangjay: Here also need master-api enabled for feature-gates "HugePages=true"

wjiangjay:
Others LGTM, and do we need the section "how to switch hugepage size"?

ahardin-rh:
@wjiangjay Thanks! Updated.

@jeremyeder @derekwaynecarr Did you want to add information on switching hugepage size per the comment from QE? If so, let me know what details are missing. Thanks!

jeremyeder:
Meaning an additional fragment that demonstrates

hugepages-1Gi: 2Gi

And perhaps a link out to the RHEL docs on it:
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/performance_tuning_guide/s-memory-transhuge

ahardin-rh:
^^ @wjiangjay Can you clarify what is needed in regard to switching hugepage size?

wjiangjay:
@ahardin-rh  
```
Some platforms support multiple huge page sizes.  To allocate huge pages
of a specific size, one must precede the huge pages boot command parameters
with a huge page size selection parameter "hugepagesz=<size>".  <size> must
be specified in bytes with optional scale suffix [kKmMgG].  The default huge
page size may be selected with the "default_hugepagesz=<size>" boot parameter.
```
refer to https://www.kernel.org/doc/Documentation/vm/hugetlbpage.txt

wjiangjay:
@ahardin-rh  comment from @jeremyeder is suitable, just need make a reference to the RHEL docs if we want this.

scaling_performance/managing_hugepages.adoc
wjiangjay: @ahardin-rh  Do not forget this.

ahardin-rh:
Note to self :spiral_notepad: 
I need to add this: https://github.com/redhat-performance/openshift-psap/blob/master/playbooks/roles/cpumanager-hugepages/tasks/main.yaml#L28

ahardin-rh:
@jeremyeder I made some updates PTAL

jeremyeder:
The update around master-config looks good.  I see now that @wjiangjay spotted it 16 days ago :-)

scaling_performance/managing_hugepages.adoc
```diff
@@ -0,0 +1,148 @@
...
+----
+
+. The master API must be enabled for `feature-gates` with `HugePages=true`:
++
+----
+# cat /etc/origin/master/master-config.yaml
+...
+kubeletArguments:
```
wjiangjay: Sorry about that, seems I make you misunderstand, 

here shoud be 
``
kubernetesMasterConfig:
  apiServerArguments:
  ...
  feature-gates:
  - HugePages=true
``

PR #7020, opened 2018-01-07
fix typo for heketi sample topology file name

jeffrey4l:


ahardin-rh:
@jeffrey4l Thank you!

Typo fix. No revision history needed.

PR #7050, opened 2018-01-09
Add note for path-based routes.

imcsk8:
There's no clarification on the docs of how services should behave
for path-based routes.

Fixes: bz#1486953

### DIFF
--- a/architecture/networking/routes.adoc
+++ b/architecture/networking/routes.adoc
@@ -296,8 +296,13 @@ Path based routes specify a path component that can be compared against
 a URL (which requires that the traffic for the route be HTTP based) such
 that multiple routes can be served using the same host name, each with a
 different path. Routers should match routes based on the most specific
-path to the least; however, this depends on the router implementation. The
-following table shows example routes and their accessibility:
+path to the least; however, this depends on the router implementation. 
+The host name and path are passed through to the backend server so it should be
+able to successfully answer requests for them. 
+For example: a request to http://example.com/foo/ that goes to the router will
+result in a pod seeing a request to http://example.com/foo/.
+
+The following table shows example routes and their accessibility:
 
 .Route Availability
 [cols="3*", options="header"]
@@ -343,6 +348,7 @@ Path-based routing is not available when using passthrough TLS, as
 the router does not terminate TLS in that case and cannot read the contents
 of the request.
 ====
+
 
 [[secured-routes]]
 == Secured Routes


architecture/networking/routes.adoc
```diff
@@ -337,13 +337,20 @@ spec:
 <1> The path is the only added attribute for a path-based route.
 ====
 
+[NOTE]
+====
+The service should be able to serve the path specified on the path-based route.
```
pecameron: Not sure why this is needed. The above description lays out the rules with examples. I don't think of a service in this context. The external user provides a fqn/path and, when matched by haproxy, the request ends up on one of the endpoints that the service provided to haproxy for the fqn/path.

pecameron (commented):
I don't think we need this.

architecture/networking/routes.adoc
```diff
@@ -296,8 +296,10 @@ Path based routes specify a path component that can be compared against
 a URL (which requires that the traffic for the route be HTTP based) such
 that multiple routes can be served using the same host name, each with a
 different path. Routers should match routes based on the most specific
-path to the least; however, this depends on the router implementation. The
-following table shows example routes and their accessibility:
+path to the least; however, this depends on the router implementation. 
+The service gets the path on the HTTP headers so it should be prepared
```
pecameron: How about something like:
"The host name and path are passed through to the server."

architecture/networking/routes.adoc
```diff
@@ -296,8 +296,10 @@ Path based routes specify a path component that can be compared against
 a URL (which requires that the traffic for the route be HTTP based) such
 that multiple routes can be served using the same host name, each with a
 different path. Routers should match routes based on the most specific
-path to the least; however, this depends on the router implementation. The
-following table shows example routes and their accessibility:
+path to the least; however, this depends on the router implementation. 
+The host name and path are passed through to the server so it should be
```
knobunc: "server" is ambiguous.  Is "backing server" better?  Or pod?

architecture/networking/routes.adoc
knobunc: Agreed, we can get rid of this.

knobunc (changes requested):

Klaas-:
I do think clarification is in order, as a ose customer I wasn't aware that path based forwarding works like this:
```
spec:
  host: www.example.com
  path: "/test" 
```
forwards requests to myservice:port/test and not myservice:port/

in my head I was more thinking about apache rewrite I guess

architecture/networking/routes.adoc
```diff
@@ -296,8 +296,10 @@ Path based routes specify a path component that can be compared against
 a URL (which requires that the traffic for the route be HTTP based) such
 that multiple routes can be served using the same host name, each with a
 different path. Routers should match routes based on the most specific
-path to the least; however, this depends on the router implementation. The
-following table shows example routes and their accessibility:
+path to the least; however, this depends on the router implementation.
```
pecameron: "depends on the router implementation" so what do we do?

architecture/networking/routes.adoc
```diff
@@ -296,8 +296,10 @@ Path based routes specify a path component that can be compared against
 a URL (which requires that the traffic for the route be HTTP based) such
 that multiple routes can be served using the same host name, each with a
 different path. Routers should match routes based on the most specific
-path to the least; however, this depends on the router implementation. The
-following table shows example routes and their accessibility:
+path to the least; however, this depends on the router implementation. 
+The host name and path are passed through to the server so it should be
+prepared to succesfully answer requests for them.
```
pecameron: "prepared to succesfully answer requests for them." 
wording seems a little clumsy to me. spelling "succesfully" -> successfully

pravisankar:

architecture/networking/routes.adoc
```diff
@@ -296,8 +296,10 @@ Path based routes specify a path component that can be compared against
 a URL (which requires that the traffic for the route be HTTP based) such
 that multiple routes can be served using the same host name, each with a
 different path. Routers should match routes based on the most specific
-path to the least; however, this depends on the router implementation. The
-following table shows example routes and their accessibility:
+path to the least; however, this depends on the router implementation. 
+The host name and path are passed through to the backend server so it should be
+able to successfully answer requests for them.
```
knobunc: Can we make it abundantly clear that a request to http://example.com/foo/ that goes to the router will result in the pod seeing a request to http://example.com/foo/.

Perhaps add:

For example a request to http://example.com/foo/ that goes to the router will result in a pod seeing a request to http://example.com/foo/.

(with the right formatting)

knobunc (changes requested):

architecture/networking/routes.adoc
imcsk8: thanks, i'll add this changes

knobunc (approved):
LGTM.  Thanks

@bfallonf over to you.

bfallonf:
Thanks @imcsk8 . This makes sense. I'll merge.

bfallonf:
And then I was told that we don't cherry pick to online and dedicated branches, so I'll close those without merging. Thanks @gaurav-nelson 

cc @vikram-redhat

PR #7051, opened 2018-01-09
Multicast clarifications

danwinship:
Based on some questions from sales people about using multicast with the networkpolicy plugin.

I wasn't sure how to document the multitenant-specific vs networkpolicy-specific bits. Feel free to reorganize. Also, I wasn't sure if I should state explicitly that the subnet plugin doesn't support multicast at all, and/or that third-party network plugin may or may not support multicast but you'd have to consult the documentation for that plugin to find out.

Besides the multitenant-specific vs networkpolicy-specific changes, I also changed "multicast" to be written with a lowercase "m" when it's not at the start of a sentence, since it's not a proper noun.

### DIFF
--- a/admin_guide/managing_networking.adoc
+++ b/admin_guide/managing_networking.adoc
@@ -984,9 +984,10 @@ At this time, multicast is best used for low bandwidth coordination or service
 discovery and not a high-bandwidth solution.
 ====
 
-Multicast traffic between {product-title} pods is disabled by default. You can
-enable Multicast on a per-project basis by setting an annotation on the
-project's corresponding `netnamespace` object:
+Multicast traffic between {product-title} pods is disabled by default. If you
+are using the *ovs-multitenant* or *ovs-networkpolicy* plugin, you can enable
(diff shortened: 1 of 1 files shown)


vikram-redhat:
@bfallonf - PTAL.

bfallonf:
Thanks @danwinship . This should be OK to merge. Is this for a specific version, or should go back as far as the multicast docs go?

danwinship:
The problem that this is fixing is mostly that the multicast docs were never updated to describe how it worked with ovs-networkpolicy, so it only needs to go back to when ovs-networkpolicy was introduced. (3.5 for Tech Preview or 3.6 for GA).

For 3.6 and 3.5 you'll need to edit out the reference to `kube-service-catalog`. ("To enable multicast in the `default` project, you must also enable it in ~~the `kube-service-catalog` project and~~ all other projects that have been made global.")

kalexand-rh:
Ok, per @danwinship, this change is going back to 3.5. For 3.5 and 3.6, I'll manually CP and remove the reference to "kube-service-catalog"

PR #7060, opened 2018-01-09
Bug 1506654, updated to state that oc exec does not work when accessi…

ahardin-rh:
…ng privileged containers except when the command is executed by a cluster-admin user

https://bugzilla.redhat.com/show_bug.cgi?id=1506654


weherdh:
LGTM

ahardin-rh:
No revision history needed

PR #7071, opened 2018-01-10
Template examples do not have metadata:name field

mburke5678:
@mfojtik @bparees If this is in your area, please advise if these changes are correct:

I was working with the templates topic in the Dev Guide and was getting an error for a few examples:
`The Template "" is invalid: metadata.name: Required value: name or generateName is required`

I added a metadata:name field and the template was created. 

Also, I got a YAML error on one example, and I needed to remove the \ in "{.data['my\.username']}"
```
metadata:
    annotations:
      template.openshift.io/expose-username: "{.data['my\.username']}"
```

### DIFF
--- a/dev_guide/templates.adoc
+++ b/dev_guide/templates.adoc
@@ -503,6 +503,8 @@ Here is an example of a full template with parameter definitions and references:
 ----
 kind: Template
 apiVersion: v1
+metadata:
+  name: my-template
 objects:
   - kind: BuildConfig
     apiVersion: v1
@@ -579,6 +581,8 @@ earlier.
 ----
 kind: "Template"
 apiVersion: "v1"
+metadata:
+  name: my-template
 objects:
   - kind: "Service" <1>
     apiVersion: "v1"
@@ -661,27 +665,31 @@ Unless escaped with a backslash, Kubernetes' JSONPath implementation interprets
 characters such as `.`, `@`, and others as metacharacters, regardless of their
 position in the expression. Therefore, for example, to refer to a `ConfigMap`
 datum named `my.key`, the required JSONPath expression would be
-`{.data['my\.key']}`.
+`{.data['my\.key']}`. Depending on how the JSONPath expression is then written in YAML, 
+an additional backslash might be required, for example `"{.data['my\\.key']}"`.
 ====
 
 The following is an example of different objects' fields being exposed:
 
-====
 [source,yaml]
 ----
 kind: Template
 apiVersion: v1
+metadata:
+  name: my-template
 objects:
 - kind: ConfigMap
   apiVersion: v1
   metadata:
+    name: my-template-config
     annotations:
-      template.openshift.io/expose-username: "{.data['my\.username']}"
+      template.openshift.io/expose-username: "{.data['my\\.username']}"
   data:
     my.username: foo
 - kind: Secret
   apiVersion: v1
   metadata:
+    name: my-template-config-secret
     annotations:
       template.openshift.io/base64-expose-password: "{.data['password']}"
   stringData:
@@ -689,6 +697,7 @@ objects:
 - kind: Service
   apiVersion: v1
   metadata:
+    name: my-template-service
     annotations:
       template.openshift.io/expose-service_ip_port: "{.spec.clusterIP}:{.spec.ports[?(.name==\"web\")].port}"
   spec:
@@ -698,11 +707,12 @@ objects:
 - kind: Route
   apiVersion: v1
   metadata:
+    name: my-template-route
     annotations:
       template.openshift.io/expose-uri: "http://{.spec.host}{.spec.path}"
   spec:
     path: mypath
-====
+----
 
 An example response to a `bind` operation given the above partial template
 follows:
@@ -802,6 +812,8 @@ annotation. Further examples can be found in the OpenShift quickstart templates.
 ----
 kind: Template
 apiVersion: v1
+metadata:
+  name: my-template
 objects:
 - kind: BuildConfig
   apiVersion: v1


dev_guide/templates.adoc
```diff
@@ -671,12 +675,14 @@ The following is an example of different objects' fields being exposed:
...
 apiVersion: v1
+metadata:
+  name: my-template
 objects:
 - kind: ConfigMap
   apiVersion: v1
   metadata:
     annotations:
-      template.openshift.io/expose-username: "{.data['my\.username']}"
+      template.openshift.io/expose-username: "{.data['my.username']}"
```
bparees: i suspect the backslash was intentional/necessary.  @jim-minter ?

bparees: (perhaps it needs to be double escaped, however)

bparees:
need @jim-minter to weigh in on the blacklash, the rest seems ok to me.

mburke5678:
@bparees @jim-minter I get the following error with the \

```
oc create -f template-expose.yaml
error: yaml: line 9: found unknown escape character
```

jim-minter:
Hold this for a little while please.  The backslash is intentional, it is illustrating a comment in the document immediately above it.  If doubled up, the backslash is accepted by `oc create`, but when you get the object back in yaml form via `oc get`, the backslash is single again - so I fear there's a client error (regression?) which needs to be dug out.  I'll take a look in more detail tomorrow.

jim-minter:
No client error (https://blog.plover.com/prog/compiler-error.html), and I've learned a new detail about yaml :grimacing:

From http://www.yaml.org/spec/current.html#id2517668: "escape sequences are only interpreted in double quoted scalars. In all other scalar styles, the “\” character has no special meaning and non-printable characters are not available."

So OpenShift's behaviour here is correct; I got our documentation wrong.

@mburke5678 please could I ask you to:

1) add an additional backslash so that the line in question reads `      template.openshift.io/expose-username: "{.data['my\\.username']}"`

2) append the following text to the paragraph directly above which ends "datum named \`my.key\`, the required JSONPath expression would be \`{.data['my\\.key']}\`.":

Depending on how the JSONPath expression is then written in YAML, an additional backslash may be required, e.g. \`"{.data['my\\\\.key']}"\`.

Thanks!

mburke5678:
@jim-minter @bparees PTAL

jim-minter:
thanks!

mburke5678:
@xiuwang Is this your area? Can you take a look at my changes?

dev_guide/templates.adoc
```diff
@@ -671,12 +676,14 @@ The following is an example of different objects' fields being exposed:
 ----
 kind: Template
 apiVersion: v1
+metadata:
+  name: my-template
 objects:
 - kind: ConfigMap
   apiVersion: v1
   metadata:
```
xiuwang: Need define objects.metadata.name here

mburke5678: @xiuwang PTAL

xiuwang: @mburke5678 Other objects in this template also need add objects.metadata.name

mburke5678: @xiuwang I added an objects.metadata.name` to each object in the template. Is name required? The `oc create -f`appears to have worked without the names.

xiuwang: @mburke5678 No error when 'oc create -f ' without objects.metadata.name, but will prompt error when 'oc new-app <above_template>'
 error: ConfigMap "" is invalid: metadata.name: Required value: name or generateName is required
    error: Secret "" is invalid: metadata.name: Required value: name or generateName is required
    error: Service "" is invalid: metadata.name: Required value: name or generateName is required
    error: Route "" is invalid: [metadata.name: Required value: name or generateName is required, spec.path: Invalid value: "mypath": path must begin with /, spec.to.name: Required value]

jim-minter: @bparees do you view the above as an incompatibility in new-app which we should look at fixing?

mburke5678: @jim-minter @bparees @xiuwang If it matters, I was running the command on a 3.5 system. I can try on 3.7 once I get it installed.

mburke5678: @xiuwang I did update the PR with object names for secret, service, and route.

bparees: @jim-minter i think we should first determine which version of oc client it effects.  "oc create -f" is effectively doing server side validation.  "oc new-app" is effectively doing client side validation, so if you're using a 3.5 client and a 3.7 server, you're getting different validation versions.

jim-minter: More coffee needed.  I don't think there's a bug here.  `oc create <template object>` doesn't verify that objects[*].metadata.name exist (in any case, any could be parameterised, and the parameter could be left empty at process time).  `oc process <template object> | oc create -f -` will error out if any object has no name.  `oc new-app` is correctly effectively doing the latter.

bparees: +1

bparees: (I missed the nuance of what object was actually being created also)

xiuwang: @mburke5678 Thanks all of  you, LGTM now

mburke5678:
[rev_history]
|xref:../dev_guide/templates.adoc#dev-guide-templates[Templates]
|Corrected the template file example in the xref:../dev_guide/templates.adoc#writing-exposing-object-fields[Exposing Object Fields] section.
%

mburke5678:

ahardin-rh:
LGTM! 🎆

PR #7124, opened 2018-01-12
Updated support matrix links

ahardin-rh:
Follow-up work to https://bugzilla.redhat.com/show_bug.cgi?id=1498912

cc @vikram-redhat @aheslin PTAL

Note that the Online and Dedicated support matrices links to https://www.openshift.com/features/containers.html are preserved, as https://access.redhat.com/articles/2176281 is only relevant to OCP.


ahardin-rh:
No revision history needed

PR #7161, opened 2018-01-15
Bug 1518355 Increase Journald Retention Time

bmcelvee:
https://bugzilla.redhat.com/show_bug.cgi?id=1518355


bmcelvee:
Minor change. No revision history needed.

PR #7196, opened 2018-01-16
Fix anchor that was changed to oc adm should be oadm

mburke5678:
Changes two references to oadm to oc adm that should remain oadm via
https://github.com/openshift/openshift-docs/pull/7132


mburke5678:
Implemented https://github.com/openshift/openshift-docs/pull/4785 in 3.6

PR #7210, opened 2018-01-17
Bug 1315850, added supplemental information about SCC

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1315850

### DIFF
--- a/admin_guide/manage_scc.adoc
+++ b/admin_guide/manage_scc.adoc
@@ -39,7 +39,6 @@ endif::openshift-dedicated[]
 
 To get a current list of SCCs:
 
-====
 ----
 $ oc get scc
 
@@ -52,7 +51,6 @@ nonroot            false     []        MustRunAs   MustRunAsNonRoot   RunAsAny  
 privileged         true      [*]       RunAsAny    RunAsAny           RunAsAny    RunAsAny    <none>     false            [*]
 restricted         false     []        MustRunAs   MustRunAsRange     MustRunAs   RunAsAny    <none>     false            [configMap downwardAPI emptyDir persistentVolumeClaim secret]
 ----
-====
 
 [[examining-a-security-context-constraints-object]]
 == Examining a Security Context Constraints Object
@@ -60,7 +58,6 @@ restricted         false     []        MustRunAs   MustRunAsRange     MustRunAs 
 To examine a particular SCC, use `oc get`, `oc describe`, `oc export`, or `oc
 edit`. For example, to examine the *restricted* SCC:
 
-====
 ----
 $ oc describe scc restricted
 Name:					restricted
@@ -94,7 +91,6 @@ Settings:
   Supplemental Groups Strategy: RunAsAny
     Ranges:				<none>
 ----
-====
 
 ifdef::openshift-enterprise,openshift-origin[]
 [NOTE]
@@ -111,7 +107,6 @@ To create a new SCC:
 . Define the SCC in a JSON or YAML file:
 +
 .Security Context Constraint Object Definition
-====
 ----
 kind: SecurityContextConstraints
 apiVersion: v1
@@ -131,7 +126,6 @@ users:
 groups:
 - my-admin-group
 ----
-====
 +
 Optionally, you can add drop capabilities to an SCC by setting the
 `*requiredDropCapabilities*` field with the desired values. Any specified
@@ -139,14 +133,12 @@ capabilities will be dropped from the container. For example, to create an SCC
 with the `*KILL*`, `*MKNOD*`, and `*SYS_CHROOT*` required drop capabilities, add
 the following to the SCC object:
 +
-====
 ----
 requiredDropCapabilities:
 - KILL
 - MKNOD
 - SYS_CHROOT
 ----
-====
 +
 You can see the list of possible values in the
 link:https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities[Docker
@@ -160,25 +152,20 @@ to drop all possible capabilities.
 
 . Then, run `oc create` passing the file to create it:
 +
-====
 ----
 $ oc create -f scc_admin.yaml
 securitycontextconstraints "scc-admin" created
 ----
-====
 
 . Verify that the SCC was created:
 +
-====
 ----
 $ oc get scc scc-admin
 NAME        PRIV      CAPS      SELINUX    RUNASUSER   FSGROUP    SUPGROUP   PRIORITY   READONLYROOTFS   VOLUMES
 scc-admin   true      []        RunAsAny   RunAsAny    RunAsAny   RunAsAny   <none>     false            [awsElasticBlockStore azureDisk azureFile cephFS cinder configMap downwardAPI emptyDir fc flexVolume flocker gcePersistentDisk gitRepo glusterfs iscsi nfs persistentVolumeClaim photonPersistentDisk quobyte rbd secret vsphere]
 ----
-====
 
 [[deleting-security-context-constraints]]
-
 == Deleting Security Context Constraints
 
 To delete an SCC:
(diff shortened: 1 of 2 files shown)


pweil- (approved):
LGTM

stuartchuan:
The changes looks good and  verified with OCP3.1
# openshift version
openshift v3.1.1.11-9-g44fe9ba
kubernetes v1.1.0-origin-1107-g4c8e6f4
etcd 2.1.2

admin_guide/manage_scc.adoc
```diff
@@ -57,10 +57,9 @@ restricted         false     []        MustRunAs   MustRunAsRange     MustRunAs
 [[examining-a-security-context-constraints-object]]
 == Examining a Security Context Constraints Object
 
-To examine a particular SCC, use `oc get`, `oc describe`, `oc export`, or `oc
-edit`. For example, to examine the *restricted* SCC:
+To examine a particular SCC, use `oc get`, `oc describe`, `oc export`, `oc
+explain`, or `oc edit`. For example, to examine the *restricted* SCC:
```
php-coder: `oc explain` isn't related to "particular" SCC. Instead it shows possible properties that all SCC have.

admin_guide/manage_scc.adoc
```diff
@@ -94,7 +93,14 @@ Settings:
...
     Ranges:				<none>
 ----
-====
+
+Existing `SecurityContext` and `PodSecurityContext` variables could be explained
+by `oc explain`:
+
+----
+$ oc explain Pod.spec.securityContext
+$ oc explain Pod.spec.containers.securityContext
```
php-coder: I don't see how it is related and how it could be helpful in the chapter about SCC. If you want to demonstrate the usage of `oc explain` to inspect SCC common properties, than it should be `oc explain SecurityContextConstraints` but IMHO it has less value to admin.

php-coder (changes requested):
The SCC chapter is long enough and I'm unsure that this changes adds something really important.

Could we move it to other doc or maybe improve it somehow else?

ahardin-rh:
@php-coder Understood. I would be okay with moving this information. Do you think we should add it to the CLI Operations section, or elsewhere?
https://docs.openshift.com/container-platform/3.7/cli_reference/basic_cli_operations.html#application-modification-cli-operations

php-coder:
@ahardin-rh While we could add `oc explain` to that page I wouldn't do that. It's rarely used command that doesn't belong to the list of "basic CLI operations". IMHO it's better to provide less details to the new users in order to protect them against confusion.

Looking back to the BZ, I see that the request was to provide a few examples and explain how they works and how SCC affects the process. I'd expect to see the something like this:
1) example of the DeploymentConfig that runs a pod (without explicit `runAsUser`)
1a) explain that it will be processed by one of the available SCCs (99% it will be "restricted" SCC)
1b) the `runAsUser` will be set to a random value
1c) describe how `openshift.io/sa.scc.uid-range` involved into this process
2) modify example to explicitly specify `runAsUser`
2a) mention that it will be rejected by default if user doesn't gave access to SCC that permits any UIDs
2b) make a pod running by granting access to appropriate SCC

At the end mention that the same mechanism is working with SELinux, fsGroup, and SupplementalGroups.

ahardin-rh:
@php-coder Okay, thank you. Yes, it looks like we need to refocus on the original request.

ahardin-rh:
@php-coder Picking this back up. I took my first stab at adding in the examples and dropped the oc explain details, as not to cause confusion. Please let me know if this is what you had in mind. Thanks!

admin_guide/manage_scc.adoc
```diff
@@ -52,15 +51,13 @@ nonroot            false     []        MustRunAs   MustRunAsNonRoot   RunAsAny
 privileged         true      [*]       RunAsAny    RunAsAny           RunAsAny    RunAsAny    <none>     false            [*]
 restricted         false     []        MustRunAs   MustRunAsRange     MustRunAs   RunAsAny    <none>     false            [configMap downwardAPI emptyDir persistentVolumeClaim secret]
 ----
-====
 
 [[examining-a-security-context-constraints-object]]
 == Examining a Security Context Constraints Object
 
-To examine a particular SCC, use `oc get`, `oc describe`, `oc export`, or `oc
+To examine a particular SCC, use `oc get`, `oc describe`, `oc export, or `oc
```
php-coder: Looks like it as unintentional change.

admin_guide/manage_scc.adoc
```diff
@@ -208,6 +195,77 @@ In order to preserve customized SCCs during upgrades, do not edit settings on
...
+----
+apiVersion: v1
+kind: Pod
+...
+fsGroup:
+  type:
+  ranges:
+  - min: 5000
+    max: 6000
+runAsUser: <1>
```
php-coder: Unfortunately, both examples won't work because Pod doesn't have such fields/values. Here is example of Kubernetes documentation about `SecurityContext` that could be used as illustration: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/

admin_guide/manage_scc.adoc
```diff
@@ -208,6 +195,77 @@ In order to preserve customized SCCs during upgrades, do not edit settings on
 the default SCCs other than priority, users, and groups.
 ====
 
+[[example-security-context-constraints]]
+=== Example Security Context Constraints Settings
+
+.Without Explicit runAsUser Setting
```
php-coder: The issue and examples are all around Pod's `SecurityContext` and hence I think it's better to move this chapter to user guide. We still need to mention how SCC could affect it and also says that admin should grant appropriate permissions.

admin_guide/manage_scc.adoc
```diff
@@ -208,6 +195,77 @@ In order to preserve customized SCCs during upgrades, do not edit settings on
...
+  ranges:
+  - min: 5000
+    max: 6000
+----
+<1> When `runAsUser` is not explicitly set, one of the available SCCs (most likely
+ a restricted SCC) will process the configuration. `runAsUser` will be set to a
+ random value.  The admission controller looks for the
+ `openshift.io/sa.scc.uid-range` annotation to populate range fields. See
+ xref:../architecture/additional_concepts/authorization.adoc#understanding-pre-allocated-values-and-security-context-constraints[Understanding
+ Pre-allocated Values and Security Context Constraints] for more information.
```
php-coder: I'd suggest another description for that field:

> When a container (or a pod) doesn't request a user id under which it should be run upon, the effective UID depends on the SCC that admits this pod. Because `restricted` SCC is granted to all authenticated users by default, it will be available to all users and service accounts and will be used in most of the cases. The `restricted` SCC uses `MustRunAsRange` strategy for constraining and defaulting the possible values of the `securityContext.runAsUser` field. As it doesn't provide this range, the admission controller will look for the `openshift.io/sa.scc.uid-range` annotation on the current namespace to populate range fields. At the end, a container will have `runAsUser` equal to a semi-random value (the first value of the range that is hard to predict because every namespace has different ranges).

@ahardin-rh Could you review this description and improve it? The link to "Understanding Pre-allocated Values and Security Context Constraints" also should be in this paragraph.

admin_guide/manage_scc.adoc
```diff
@@ -208,6 +195,77 @@ In order to preserve customized SCCs during upgrades, do not edit settings on
...
+    type: ...
+    level: ...
+supplementalGroups:
+  type: MustRunAs
+  ranges:
+  - min: 5000
+    max: 6000
+----
+<1> If you do not have access to a SCC that permits any UID, the `runAsUser` setting
+will be rejected by default.
```
php-coder: > If you

This could be interpreted wrong..

> to a SCC that permits any UID

and this is also vague..

> will be rejected by default

and here it's better to specify in what circumstances..

I'd suggest the following:

> A container (or a pod) that requests a specific user id, will be accepted by the OpenShift only when a service account (or a user) has been granted access to SCC that allows such user id. SCC may allow arbitrary ids, or an id that falls into a range, or even exactly this user id.

Hm, this also isn't perfect, so, please, help me improve it :)

php-coder:
@ahardin-rh Thank you for this work! I left a couple comments and looks like we'll have more than one iteration before it will be ready to be merged.

I hope that you will help me to improve the wording for the suggested paragraphs :)

admin_guide/manage_scc.adoc
ahardin-rh: @php-coder When you say "user guide" do you mean this section? https://docs.openshift.com/container-platform/3.7/architecture/additional_concepts/authorization.html#security-context-constraints
This particular topic focuses on conceptual information.

php-coder: > When you say "user guide" do you mean this section? 

I meant "any document under `dev_guide` directory" (in contrast to `admin_guide` that is dedicated to admins).

> https://docs.openshift.com/container-platform/3.7/architecture/additional_concepts/authorization.html#security-context-constraints
This particular topic focuses on conceptual information.

Can we create a new topic then? It should be somewhere near how to create a pod.

ahardin-rh: @aheslin Do you have any thoughts on where this content should be placed?

aheslin:
@ahardin-rh I think we can leave the content where it is and follow up with moving the content in a separate PR.

ahardin-rh:
@php-coder @aheslin Thanks!

@php-coder I addressed all of your other comments, but will work to reorganize the content in a separate PR since that work will be part of a larger initiative. Does everything else look okay? Thanks again!

php-coder (changes requested):
Thank you for yet another attempt!

I left a couple comments but also I submitted https://github.com/ahardin-rh/openshift-docs/pull/3 that resolves them. PTAL.

php-coder:
It's LGTM.

@calston It would be great if you quickly look at it to check whether its clear from a user's perspective :) Thanks!

ahardin-rh:
@calston Do you mind reviewing when you get a moment? Thanks!

calston:
Makes good sense 👍

Issue #7254, opened 2018-01-19
Web console walkthrough out of date


spadgett:
This topic was not updated for the new experience in 3.7.

https://docs.openshift.org/latest/getting_started/developers_console.html

cc @jwforres @ahardin-rh

ahardin-rh:
WIP PR: https://github.com/openshift/openshift-docs/pull/7255

ahardin-rh:
Closing, as this is now  outdated and the docs have since been updated and improved

PR #7255, opened 2018-01-19
[WIP]Updates to the Create and Build an Image Using the Web Console topic

ahardin-rh:
This addresses https://github.com/openshift/openshift-docs/issues/7254

@spadgett @jwforres PTAL


spadgett:
What you see in the UI depends on the item you pick. For instance, you'll get builds on the overview for builder images, but you won't get builds for other templates that don't use builders. It might be good to stick to a specific example builder or template so the steps always match what the user sees.

vikram-redhat:
@ahardin-rh - can you please move this along or close if no longer relevant?

ahardin-rh:
This is pretty old, so I am going to close and I can revisit separately if needed.

PR #7256, opened 2018-01-22
Added release notes for the 3.7.23 async release

ahardin-rh:


ahardin-rh:
@aheslin @vikram-redhat FYI, these errata are now in PUSH READY state. I'm just waiting for them to go live.

PR #5049, opened 2017-08-21
Fixes incorrect statement about IPSec regarding throughput.

cjoster:
Documentation falsely claims that the lack of hardware NIC acceleration will lead to increased network throughput when using IPSec.

PR #7269, opened 2018-01-23
Fixed incorrect statement about IPSec throughput

ghost:
@cjoster Please review this follow-up to https://github.com/openshift/openshift-docs/pull/5049.

### DIFF
--- a/scaling_performance/network_optimization.adoc
+++ b/scaling_performance/network_optimization.adoc
@@ -58,8 +58,8 @@ in latency tests.
 [[scaling-performance-optimizing-mtu]]
 === Optimizing the MTU for Your Network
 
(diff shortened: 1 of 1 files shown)


cjoster (changes requested):
The two lines I changed look good, but I do want to raise an objection to "IPSec encryptes traffic at the IP *address* level", which would be better said as the "IP payload level" if, for some reason, "the IP level" (as in the protocol level, or layer 3, isn't sufficient.

mburke5678:
@tmorriso-rh LGTM pending cjoster comment above.

tmorriso-rh:
@cjoster Changed "IP address level" to "IP payload level." Please review. Thanks.

cjoster:
Looks good. Thank you!

PR #7337, opened 2018-01-25
added note on advanced shell privs per BZ

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1369513

### DIFF
--- a/install_config/router/f5_router.adoc
+++ b/install_config/router/f5_router.adoc
@@ -44,6 +44,7 @@ requirements:
 * A F5 host IP with:
 ** Credentials for API access
 ** SSH access via a private key
+* An F5 user with Advanced Shell access
 * A virtual server for HTTP routes:
 ** _HTTP profile_ must be _http_.
 * A virtual server with HTTP profile routes:
@@ -133,7 +134,8 @@ BIG-IP®* host:
 address.
 
 |`--external-host-username`
-|Specifies the *F5 BIG-IP®* user name (typically *admin*).
+|Specifies the *F5 BIG-IP®* user name (typically *admin*). 
+The *F5 BIG-IP* user account must have access to the Advanced Shell (Bash) on the F5 BIG-IP system.
 
 |`--external-host-password`
 |Specifies the *F5 BIG-IP®* password.
@@ -343,4 +345,9 @@ $ oc adm router \
     --external-host-vxlan-gw=10.131.0.5/14
 ----
 +
+[NOTE]
++
+====
+The `external-host-username` is a *F5 BIG-IP* user account with access to the Advanced Shell (Bash) on the F5 BIG-IP system.
+
 The F5 setup is now ready, without the need to set up the ramp node.


mburke5678:
@sferich888 I am not clear on what you mean by "advanced shell" permissions. Privileged SCC? Can you specify what is needed?

mburke5678:
@sferich888 PTAL
Added:
"The *F5 BIG-IQ* user account must have access to the Advanced Shell (Bash) on the BIG-IQ system."

mburke5678:
@bmeng PTAL

bmeng:
@lihongan Please help review the doc.

install_config/router/f5_router.adoc
```diff
@@ -53,6 +53,7 @@ requirements:
 * For edge integration (not recommended):
 ** A working ramp node
 ** A working tunnel to the ramp node
+* An {product-title} user assigned the xref:../../admin_guide/manage_scc.html#listing-security-context-constraints[`privileged` security context constraint]
```
lihongan: we always assign `privileged` scc to a service account like `router`, not an user.

install_config/router/f5_router.adoc
```diff
@@ -133,7 +134,8 @@ BIG-IP®* host:
 address.
 
 |`--external-host-username`
-|Specifies the *F5 BIG-IP®* user name (typically *admin*).
+|Specifies the *F5 BIG-IP®* user name (typically *admin*). 
+The *F5 BIG-IQ* user account must have access to the Advanced Shell (Bash) on the BIG-IQ system.
```
lihongan: what's the `BIP-IQ` mean? should be `*F5 BIG-IP®*` right ?

install_config/router/f5_router.adoc
```diff
@@ -343,4 +345,9 @@ $ oc adm router \
     --external-host-vxlan-gw=10.131.0.5/14
 ----
 +
+[NOTE]
++
+====
+The `external-host-username` is a *F5 BIG-IQ* user account with access to the Advanced Shell (Bash) on the BIG-IQ system.
```
lihongan: same to above, should be `*F5 BIG-IP®*` ?

lihongan:
IMHO, we don't need add the link to a f5 KB article, the user account should be created by F5 BIP-IP admin and F5 admin should have the knowledge about the user account properties (Advance shell).
Actually, when you create a new user via F5 web console, you need click the drop down box called "Terminal Access" and select one of three options: Disabled, **Advanced shell** and tmsh, the second options just we note in this PR.
BTW, the default F5 BIG-IP admin user has `Advanced shell` option as default.

mburke5678:
@lihongan I have made your changes. PTAL

install_config/router/f5_router.adoc
```diff
@@ -133,7 +134,8 @@ BIG-IP®* host:
 address.
 
 |`--external-host-username`
-|Specifies the *F5 BIG-IP®* user name (typically *admin*).
+|Specifies the *F5 BIG-IP®* user name (typically *admin*). 
+The *F5 BIG-IQ* user account must have access to the Advanced Shell (Bash) on the F5 BIG-IP system.
```
lihongan: Seems  here ```The *F5 BIG-IQ* user account``` has not be updated to `BIG-IP`. I checked the F5 doc and found that BIG-IQ is Centralized Management system and can discover, manage and monitor remote BIG-IP devices, but it is not needed for OpenShift and F5 integration, so I think better to use `BIP-IP` here avoiding confusion.

install_config/router/f5_router.adoc
```diff
@@ -343,4 +345,9 @@ $ oc adm router \
     --external-host-vxlan-gw=10.131.0.5/14
 ----
 +
+[NOTE]
++
+====
+The `external-host-username` is a *F5 BIG-IQ* user account with access to the Advanced Shell (Bash) on the F5 BIG-IP system.
```
lihongan: same to above, ```a *F5 BIG-IQ* user account```

install_config/router/f5_router.adoc
mburke5678: @lihongan Thank you for clarifying. I don't think I saw that I was using the wrong term.

mburke5678:
@sferich888 The BZ was filed against 3.2; do we want to port this change all the way back to 3.2?

mburke5678:
[rev_history]
|xref:../install_config/router/f5_router.adoc#install-config-router-f5[Installation -> Using the F5 Router Plug-in]
|Added a note that the F5 user needs Advanced Shell permission to use the xref:../install_config/router/f5_router.adoc#deploying-the-f5-router[F5 Router Plug-in].
%

mburke5678:

install_config/router/f5_router.adoc
```diff
@@ -343,4 +345,9 @@ $ oc adm router \
     --external-host-vxlan-gw=10.131.0.5/14
 ----
 +
+[NOTE]
++
+====
+The `external-host-username` is a *F5 BIG-IP* user account with access to the Advanced Shell (Bash) on the F5 BIG-IP system.
+
```
bfallonf: The markup above is a little off, and the sentence below has been pulled into the note box. line 352 should have another set of =s and perhaps no line space between the NOTE heading and the text. See:

https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#admonitions

bfallonf:
@mburke5678 A comment from me, but looks good otherwise.

lihongan:
LGTM, thank you.

install_config/router/f5_router.adoc
mburke5678: @bfallonf Fixed. Thanks!

PR #7347, opened 2018-01-25
[Closed-Conflicts]Bug 1518474 Update documentation to reflect change to JSON driver as default

bmcelvee:
https://bugzilla.redhat.com/show_bug.cgi?id=1518474

### DIFF
--- a/install_config/aggregate_logging.adoc
+++ b/install_config/aggregate_logging.adoc
@@ -1732,8 +1732,10 @@ $ curl --key /etc/elasticsearch/secret/admin-key \
 [[fluentd-update-source]]
 == Changing the Aggregated Logging Driver
 
-By default, aggregated logging uses the `journald` log driver
-xref:../install_config/install/advanced_install.adoc#configuring-host-variables[unless `json-file` was specified during installation]. You can change the log driver between `journald` and `json-file` as needed.
+By default, aggregated logging uses the `json-file` log driver
+xref:../install_config/install/advanced_install.adoc#configuring-host-variables[unless
+`json-file` was specified during installation]. You can change the log driver
+between `journald` and `json-file` as needed.
 
 [IMPORTANT]
 ====
--- a/install_config/aggregate_logging.adoc
+++ b/install_config/aggregate_logging.adoc
@@ -360,10 +360,6 @@ xref:../install_config/install/advanced_install.adoc#configuring-node-host-label
 |`openshift_logging_fluentd_memory_limit`
 |The memory limit for Fluentd pods.
 
-|`openshift_logging_fluentd_use_journal`
-|Set to `true` if Fluentd should read log entries from Journal. The default is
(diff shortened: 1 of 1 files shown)


install_config/aggregate_logging.adoc
```diff
@@ -855,7 +824,7 @@ still expects to read logs generated using the `json-file` log driver. This
 results in a lack of log ingestion. If this has happened within your logging cluster,
 xref:fluentd-upgrade-source[troubleshoot] it.
 
-Aggregated logging is only supported using the `journald` driver in Docker. See
+Aggregated logging is only supported using the `JSON` driver in Docker. See
```
anpingli: It support both journald and json driver.

install_config/aggregate_logging.adoc
```diff
@@ -320,6 +320,10 @@ xref:../install_config/install/advanced_install.adoc#configuring-node-host-label
 |`openshift_logging_fluentd_memory_limit`
 |The memory limit for Fluentd pods.
 
+|`openshift_logging_fluentd_use_journal`
```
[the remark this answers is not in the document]
portante: It seems @anpingli is right.  As such, this should probably read that along with it being deprecated, the setting is now ignored.

install_config/aggregate_logging.adoc
portante: Perhaps if we rewrite this as, "Aggregated logging is only supported using the `json-file` logging driver with Docker."

portante (changes requested):
Can we take a few minutes to review this together?  It seems like we need to make further edits to this document than what is done here.

install_config/aggregate_logging.adoc
anpingli: @portante Aggregated logging haven't drop support to journal logging driver with Docker.  I think we just suggest using json-file.

portante: @anpingli, we actually don't support logging application logs through journald, even though we have all the code in place.  There are many many issues with it, and while it is a valid option, we need to correct the docs for 3.9 to rescind support before we release.  We can say that the journald logging driver is an option that is still available to users for compatibility, but in 3.10 or later, we'll be removing support for it entirely.

install_config/aggregate_logging.adoc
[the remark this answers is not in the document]
jcantrill: This VAR and any reference to it can be removed completely.  Grep'ing through the code I see no reference to it other then warning and debug messages that we shouldn't use it.

bmcelvee:
@portante that's a good idea. Do you have some time next week?

bmcelvee:
Closing this PR in favor of https://github.com/openshift/openshift-docs/pull/8012.

portante:
@bmcelvee, yes, let's try to meet with @jcantrill as well.

PR #7352, opened 2018-01-26
Fixed playbook path.

jmferrer:
In https://docs.openshift.com/container-platform/3.7/install_config/aggregate_logging.html#aggregate-logging-cleanup var "-e openshift_logging_install_logging=False" is missing but in repo is right.

Is master the right branch to do MR?


ahardin-rh:
@jmferrer Thanks! Yes, the master branch is correct.
@sdodson PTAL

jmferrer:
Update: path is right now so I close this MR.

PR #7360, opened 2018-01-26
Automatic width for columns

git001:
due to the fact that the https://docs.openshift.org/latest/architecture/networking/routes.html#env-variables does not fits into the content width let's calculate the asciidoctor the width for the columns

### DIFF
--- a/architecture/topics/router_environment_variables.adoc
+++ b/architecture/topics/router_environment_variables.adoc
@@ -15,7 +15,7 @@ $ oc set env dc/router ROUTER_SYSLOG_ADDRESS=127.0.0.1 ROUTER_LOG_LEVEL=debug
 ----
 
 .Router Environment Variables
-[cols="2,2,6", options="header"]
+[cols="2,2,6", options="header,autowidth"]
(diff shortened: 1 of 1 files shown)


kalexand-rh:
@git001, I tested this change locally. While adding `autowidth` does improve the spacing of the variable column, it also adds a scrollbar that makes the table even harder to use.

@vikram-redhat, do you have an opinion here?

vikram-redhat:
@gaurav-nelson has been testing this. Gaurav - did we apply this to org docs?

gaurav-nelson:
I have checked the docs for both https://docs.openshift.org/latest/architecture/networking/routes.html#env-variables and https://docs.openshift.com/container-platform/3.9/architecture/networking/routes.html#env-variables

The tables look fine on desktops. I guess this issue is already resolved by https://github.com/openshift/openshift-docs/pull/8340

@git001 Can you please check the mentioned table again?

vikram-redhat:
I am closing this PR. @git001 if the issue persists, please open a new one.

PR #7363, opened 2018-01-26
add docs for cluster role aggregation

deads2k:
Fixes https://github.com/openshift/origin/issues/18232

I've taken a stab at where to include this, but I don't have any strong opinions.  It's lifted from https://kubernetes.io/docs/admin/authorization/rbac/#aggregated-clusterroles .

@adellape I'm not sure who to assign this to. 
@openshift/sig-security

### DIFF
--- a/architecture/additional_concepts/authorization.adoc
+++ b/architecture/additional_concepts/authorization.adoc
@@ -490,6 +490,77 @@ JSON merge on RBAC resources.
 endif::[]
 ifdef::openshift-origin,openshift-enterprise,openshift-dedicated[]
 
+### Aggregated ClusterRoles
+
+As of 1.9, ClusterRoles can be created by combining other ClusterRoles using an `aggregationRule`. The
+permissions of aggregated ClusterRoles are controller-managed, and filled in by unioning the rules of any
+ClusterRole that matches the provided label selector. An example aggregated ClusterRole:
+
+```yaml
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: monitoring
+aggregationRule:
+  clusterRoleSelectors:
+  - matchLabels:
+      rbac.example.com/aggregate-to-monitoring: "true"
+rules: [] # Rules are automatically filled in by the controller manager.
+```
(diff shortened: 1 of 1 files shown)


enj:
LGTM

architecture/additional_concepts/authorization.adoc
```diff
@@ -490,6 +490,77 @@ JSON merge on RBAC resources.
 endif::[]
 ifdef::openshift-origin,openshift-enterprise,openshift-dedicated[]
 
+### Aggregated ClusterRoles
+
+As of 1.9, ClusterRoles can be created by combining other ClusterRoles using an `aggregationRule`. The
```
php-coder: 1.9 -> 3.9

enj:
@kalexand-rh this PR is for https://github.com/openshift/origin/issues/18232

kalexand-rh:
@enj, @deads2k, questions:
Do you have more information about how this was integrated and used in OpenShift?
Was this brought into enterprise versions? (Do you have a Trello card about it?)

architecture/additional_concepts/authorization.adoc
```diff
@@ -490,6 +490,77 @@ JSON merge on RBAC resources.
...
+  labels:
+    rbac.example.com/aggregate-to-monitoring: "true"
+# These rules will be added to the "monitoring" role.
+rules:
+- apiGroups: [""]
+  Resources: ["services", "endpoints", "pods"]
+  verbs: ["get", "list", "watch"]
+```
+
+The default user-facing roles (described below) use ClusterRole aggregation. This lets admins include rules
```
kalexand-rh: Is this true for all of the default roles?
Is there an easy way for a user to ID which ClusterRoles have aggregation?
I'm also curious what the use case is. Do you expect users to set up roles for aggregation or only consume aggregation by using it to grant permissions to objects?

architecture/additional_concepts/authorization.adoc
kalexand-rh: Which product versions is this available in? It can't just be 3.9.

PR #7369, opened 2018-01-29
fix avoid memory swapping link in cluster_limits.adoc

sngchlko:


kalexand-rh:
@vikram-redhat, it looks like you fixed this on link individual branches as part of https://github.com/openshift/openshift-docs/commit/18ae382750189b6fc431ba52d3cf6e5ef31069fb. Does anything else from that change need to make it into master?

vikram-redhat:
@kalexand-rh yes it should get merged to master as well. Otherwise origin docs would still have this broken.

kalexand-rh:
@sngchlko, this change was made on https://github.com/openshift/openshift-docs/pull/9027#. Thanks!

PR #7395, opened 2018-01-29
Bug 1528462 Add Log/Second Technical Limits

bmcelvee:
https://bugzilla.redhat.com/show_bug.cgi?id=1528462

### DIFF
--- a/install_config/aggregate_logging_sizing.adoc
+++ b/install_config/aggregate_logging_sizing.adoc
@@ -192,6 +192,15 @@ journal file corruption.
 link:https://rhn.redhat.com/errata/RHBA-2016-1026.html[This issue is resolved in
 later versions of systemd].
 
+*Fluentd*
+
+As of {product-title} 3.7, with 1KB sized log messages, the logging
+architecture tests have achieved a stable logging rate of 800-900 messages per
+second per node.
+
+Scaling the logging infrastructure to a higher throughput is currently limited
+by the current versions and configurations of the OpenShift logging components.
+
 *Scaling up*
 
 As you scale up your project, the default logging environment might need some


mffiedler:
@portante and @richm should review as well - there are quite a few factors that can affect logging throughput and starting to list them all could be a slippery slope.   Message size, message rates, number of pods logging, elasticsearch cluster size and so on.

richm:
I think it's fair to say 1000 records per second per node is the maximum we have been able to achieve in the lab under "ideal" conditions, and that you will typically see fewer than this with real data in a real production environment.  As far as why - I think we should say as little as possible - perhaps not mentioning the threading aspect - perhaps something like "due to inherent limitations in Fluentd 0.12 and Ruby 2.0, and in the way OpenShift configures Fluentd and in the log processing code used by OpenShift logging."

install_config/aggregate_logging_sizing.adoc
```diff
@@ -192,6 +192,20 @@ journal file corruption.
...
+of what the infrastructure can handle.
+
+For more stability, sending similarly sized logs, at a rate of 800-900
+messages/second on a consistent basis, is recommended.
+
+The limiting factor in scaling the logging infrastructure to a higher throughput
+resides with Fluentd (the log aggregation solution). Currently, the way Fluentd
+operates, the number of threads that can process the logs
+collected is limited, and as a result, have the defined upper bounds. 
+
```
mffiedler: w.r.t. @richm's comment, I suggest replacing the final paragraph with:

Scaling the logging infrastructure to a higher throughput is currently limited by Fluentd 0.12, Ruby 2.0, OpenShift's configuration of Fluentd and the log processing code used in OpenShift.

If that is too geeky, maybe something like "limited by the current versions of the OpenShift logging components and their configuration"

bmcelvee:
Thanks, @mffiedler and @richm! In an effort to help save us from maintaining changing version numbers, I went with the minimalist explanation.

bmcelvee:
@openshift/team-documentation PTAL 🌻

install_config/aggregate_logging_sizing.adoc
```diff
@@ -192,6 +192,18 @@ journal file corruption.
 link:https://rhn.redhat.com/errata/RHBA-2016-1026.html[This issue is resolved in
 later versions of systemd].
 
+*Fluentd*
+
+As of {product-title} 3.7, OpenShift tests its logging architecture with 1KB log
+messages, approximately 1000 messages/second. This is the upper limit
```
ahardin-rh: messages per second

install_config/aggregate_logging_sizing.adoc
```diff
@@ -192,6 +192,18 @@ journal file corruption.
...
+of what the infrastructure can handle.
+
+For more stability, sending similarly sized logs, at a rate of 800-900
+messages/second on a consistent basis, is recommended.
```
ahardin-rh: messages per second

ahardin-rh:
@bmcelvee One nit from me. Otherwise, looks good!

mffiedler (approved):
Looks good to me

install_config/aggregate_logging_sizing.adoc
portante: Can we strike the 1,000 messages/second, and only report 850 messages/second?

Why are reporting "unstable" numbers, as implied by line 201?

portante (changes requested):

install_config/aggregate_logging_sizing.adoc
richm: Ok - how about this:
```
As of {product-title} 3.7, with 1 kilobyte sized log messages, the logging architecture tests have achieved a stable logging rate of 800-900 messages per second per node.
```

portante: Sounds good, except isn't 800-900 messages a second for a specific configuration?  Would faster CPUs, networking, etc. impact this?

If so, we should report what the configuration was where we saw these numbers.

portante (changes requested):

portante:
Is any performance number of any use if you don't know the configuration?  If we are running on faster hardware, how is 850 a number a customer can use to compare to their environment?

These kinds of numbers are dangerous and can mis-represent what we are seeing.

richm:
Then we need to do some investigation to find out how other aspects of OpenShift performance are documented, and be consistent with what is done elsewhere in the OpenShift documentation.

Either that, or just say nothing about logging performance.

bmcelvee:
Sorry for the delay on this. I added Rich's suggested text. Are we okay with leaving the "800-900 messages..." bit, or should we tweak the text to be more vague?

portante:
If we can publish the details of the configuration under which those numbers were achieved, we can leave those numbers in and cite a document which has the details.  If not, then I'd suggest removing the numbers.

richm:
@bmcelvee Can you reach out to the documentation team and find out how other aspects of openshift performance are documented?  The logging documentation should be consistent with that - if there are no performance numbers, then we should omit them from the logging docs.  If there are performance data published for other aspects of openshift, then we should document the logging performance _in the same manner_.

ahardin-rh:
@richm Here are some examples of Performance metrics that we capture for each release:
https://docs.openshift.com/container-platform/3.7/scaling_performance/scaling_cluster_metrics.html
https://docs.openshift.com/container-platform/3.7/scaling_performance/cluster_limits.html

portante:
@ahardin-rh, @jeremyeder, so that documentation seems to be problematic as it does not reference a document which describes the configuration.  How are those numbers meaningful if there is no configuration to compare to?

bmcelvee:
@portante @sjwalter2 Would it be best to leave this in the KB article and not add it to the documentation?

sjwalter2:
I'm not too comfortable just relying on KB. We have real scalability issues when we don't; our best defenses against a user blissfully assuming you can have 10,000 nodes with 1000 pods on each is a documented limit on what we can support. KBase articles are good for "learn more" and "how to fix x", but not good for declaring, officially, what is supported. I recognize that it's hard to pin down quite what is supported, but if we don't have anything in hard numbers it's hard to justify telling someone "No, this won't work". I mean, we CAN say "this won't work" but the first response is "prove it."

portante:
@sjwalter2, the only problem with that is documenting a number means backing it up with a documented configuration used to get that number.  Are we really willing to buy into the work to do that?

sjwalter2:
@portante If time constraints prevent us from fully preparing hard numbers as limits (and all that would entail), we at the very least need a process or formula by which an administrator can guage their cluster's ability to scale. To me, we are unequipped to make those calls just on gut feeling, but rather need a process. Does QE perform scaling tests on the logging stack? If so, what method do they use -- maybe we can document that so that admins can run similar tests and find the boundaries on their environment. And if we document that, can we not document what our results are?

If we're unwilling/unable to document how a user should be scaling their environment, then we need to be willing to document how to understand (once your logging stack has started dropping messages or otherwise underperforming) whether that is caused by you hitting some fundamental limits. So far this has been the role of the KBase, and I'm fine with it staying there if this is the route we choose. Perhaps QE or another team can help us determine what to look for to determine that.

jeremyeder:
I was pinged about this issue.  @portante and @mffiedler own this work, and it is in-scope for 3.10 scale testing.

bmcelvee:
Thanks, @jeremyeder. Since this is in-scope for 3.10, would anyone have an objection to me closing this PR and addressing any changes down the line?

PR #6224, opened 2017-11-08
Clarify NM_CONTROLLED requirement

nekop:
Avoid misunderstanding that some users think NetworkManager is installed but has NM_CONTROLLED=no on network interface configs and DNS is not working.

PR #7404, opened 2018-01-30
Follow-up edits to PR#6224

ahardin-rh:
Follow-up to https://github.com/openshift/openshift-docs/pull/6224


ahardin-rh:
Revision history is in the original PR

PR #7433, opened 2018-01-31
Fix command typo

bmcelvee:


bmcelvee:
No revision history needed.

PR #7440, opened 2018-01-31
Bug 1491127, clarified details about technology preview features

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1491127

@spadgett @jwforres According to the above BZ, is the other information on enabling features in Technology Preview still accurate? Statefulsets does not seem to be disabled by default in the web console, so should we update some of the language there?  I removed mention of what is currently in tech preview, as this would need to be maintained across each release. Thanks!

### DIFF
--- a/install_config/web_console_customization.adoc
+++ b/install_config/web_console_customization.adoc
@@ -1007,8 +1007,6 @@ how to configure HAProxy routers to allow wildcard routes].
 Sometimes features are available in Technology Preview. By default, these
 features are disabled and hidden in the web console.
 
-Currently, there are no web console features in Technology Preview.
-
 To enable a Technology Preview feature:
 
 . Save this script to a file (for example, *_tech-preview.js_*):


spadgett:
It's likely we'll change how this works going forward. Since there are no existing features in tech preview for 3.9, it might be better just to remove that section. (Stateful sets will no longer be tech preview in 3.9.)

ahardin-rh:
@spadgett Thanks! I can remove the section for 3.9 in a separate PR. Is the content as I have it okay for 3.5 through 3.7?

spadgett:
@ahardin-rh LGTM. We did a gap in our docs since we had some tech preview features in 3.6. 

https://github.com/openshift/origin-web-console/blob/enterprise-3.6/app/scripts/constants.js#L88-L94

But removing that line is right since the statement was inaccurate.

ahardin-rh:
@spadgett Okay, thank you!

ahardin-rh:
No revision history needed.

PR #7466, opened 2018-01-31
Removed references to Tech Preview features from Online and Dedicated docs

ahardin-rh:


_topic_map.yml
```diff
@@ -890,7 +890,7 @@ Topics:
   File: openshift_pipeline
 - Name: Cron Jobs
   File: cron_jobs
-  Distros: openshift-enterprise,openshift-dedicated,openshift-origin,openshift-online
+  Distros: openshift-enterprise,openshift-origin
```
ahardin-rh: @vikram-redhat We will need redirects for Online and Dedicated, right?

adellape (approved):
:+1:

_topic_map.yml
ahardin-rh: After discussion, we decided redirects are not needed in this case

PR #7494, opened 2018-02-01
Fix typo in tree graph

pabrahamsson:
Looks like a copy-pasta mistake from tree graph earlier in document.


kalexand-rh:
@adellape, do you agree with this change?

vikram-redhat:
lgtm - merging.

PR #7503, opened 2018-02-02
fix nesting of pipeline and openshift.withCluster in delcarative pipe…

gabemontero:
…line example

@openshift/sig-developer-experience @bmcelvee ptal


bmcelvee:
LGTM! 🌻 To which versions does this fix apply?

bparees:
@bmcelvee applies to all
/lgtm

bmcelvee:
Minor change. No revision history necessary.

PR #7574, opened 2018-02-05
correct image tag v3.7 to v3.6 in OCP 3.6 doc

wmengRH:
fix https://bugzilla.redhat.com/show_bug.cgi?id=1542053


gaurav-nelson:
Thank you @wmengRH I made this point to the stage branch instead.

PR #7605, opened 2018-02-06
Fix link to RHEL Atomic Host container storage documentation

ghost:


vikram-redhat:
The link is good. Just going as far back as 3.6.

vikram-redhat:
Looks like it is already fixed in 3.11

PR #7615, opened 2018-02-06
Replaced DITAA images with PNG files

ahardin-rh:
@vikram-redhat This takes care of replacing the old DITAA images with PNG files from Emma.

I will open separate PRs to get this into the older versions.


vikram-redhat:
Awesome! That was a really quick turnaround! Thank Ashley and Emma!

ahardin-rh:
This is an image format update, so no revision history needed.

PR #7620, opened 2018-02-07
Added footnote on macvlan to specify network when using multiples

bfallonf:
For: https://trello.com/c/DDdu4qqM/723-document-support-selection-of-interface-when-using-macvlan

cc @danwinship PTAL

Or, should there be a new `pod.network.openshift.io/assign-macvlan: "eth1"` thrown in with the current one, instead of replacing it? Say, below it? That's the way it's worded in the dev trello card, but the test cases replace it.

### DIFF
--- a/admin_guide/managing_networking.adoc
+++ b/admin_guide/managing_networking.adoc
@@ -460,10 +460,10 @@ endif::openshift-origin[]
   nodeSelector:
     site: springfield-1 <6>
 ----
-<1> The `pod.network.openshift.io/assign-macvlan annotation` creates a Macvlan
-network interface on the primary network interface, and then moves it into the
-pod's network name space before starting the *egress-router* container. Preserve
-the quotation marks around `"true"`. Omitting them results in errors.
+<1> Creates a Macvlan network interface on the primary network interface, and
+moves it into the pod's network project before starting the *egress-router*
+container. Preserve the quotation marks around `"true"`. Omitting them results
+in errors. To create the Macvlan interface on a network interface other than the primary one, set the annotation value to the name of that interface. For example, `eth1`.
 <2> IP address from the physical network that the node is on and is
 reserved by the cluster administrator for use by this pod.
 <3> Same value as the default gateway used by the node.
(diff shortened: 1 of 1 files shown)


admin_guide/managing_networking.adoc
```diff
@@ -460,10 +460,11 @@ endif::openshift-origin[]
...
 ----
-<1> The `pod.network.openshift.io/assign-macvlan annotation` creates a Macvlan
-network interface on the primary network interface, and then moves it into the
-pod's network name space before starting the *egress-router* container. Preserve
-the quotation marks around `"true"`. Omitting them results in errors.
+<1> Creates a Macvlan network interface on the primary network interface, and moves
+ it into the pod's network name space before starting the *egress-router*
+ container. Set to anything other than "true" to specify the network (if using
+ multiple networks). For example, "eth1" Preserve the quotation marks around
+ `"true"`. Omitting them results in errors.
```
danwinship: Hm. I'd put the "preserve the quotation marks" warning before the discussion of alternate interfaces. (Since it only applies to the value `"true"`; you don't need quotation marks around `eth1`, etc, because those aren't YAML keywords.)

And then something like: To create the Macvlan interface on a network interface other than the primary one, set the annotation value to the name of that interface. For example, "eth1".

admin_guide/managing_networking.adoc
```diff
@@ -559,7 +560,7 @@ metadata:
   labels:
     name: egress-multi
   annotations:
-    pod.network.openshift.io/assign-macvlan: "true"
+    pod.network.openshift.io/assign-macvlan: "true" <1>
```
danwinship: When I said "you don't have to document this much" in the Trello card, what I meant was, we don't expect very many users to use this feature. So I don't think you need to update *all* of the egress-router examples. Just the first one should be fine.

bfallonf:
Thanks @danwinship . Changed to suggestions. If there's anything else, please let me know.

For now, @openshift/team-documentation PTAL.

admin_guide/managing_networking.adoc
```diff
@@ -460,10 +460,10 @@ endif::openshift-origin[]
   nodeSelector:
     site: springfield-1 <6>
 ----
-<1> The `pod.network.openshift.io/assign-macvlan annotation` creates a Macvlan
-network interface on the primary network interface, and then moves it into the
-pod's network name space before starting the *egress-router* container. Preserve
-the quotation marks around `"true"`. Omitting them results in errors.
+<1> Creates a Macvlan network interface on the primary network interface, and
+moves it into the pod's network name space before starting the *egress-router*
```
ahardin-rh: s/name space/project (Maybe? Also I see namespace as one word elsewhere in the docs.)

bfallonf:
Thanks @ahardin-rh . Changed to suggestion. I'll merge.

vikram-redhat:
@bfallonf - forgot to pick this to dedicated 3.9 stage branch? It shouldn't matter now with the new workflow, since we are building of enterprise 3.9 branch.

bfallonf:
@vikram-redhat Hmm yup. Seemed to work out, so good for me :100:

PR #7682, opened 2018-02-09
add glusterfs physical management page

jmferrer:
Added procedure to add physical blocks to containerized glusterfs  cluster.

### DIFF
--- a/admin_guide/manage_glusterfs_physical_devices.adoc
+++ b/admin_guide/manage_glusterfs_physical_devices.adoc
@@ -1,0 +1,69 @@
+[[admin-guide-manage-glusterfs-physical-devices]]
+= Managing Gluster Physical Devices
+{product-author}
+{product-version}
+:data-uri:
+:icons:
+:experimental:
+:toc: macro
+:toc-title:
+:prewrap!:
+
+toc::[]
+
+== Overview
+You can manage
+xref:../architecture/additional_concepts/storage.adoc[gluster storage]
+in your project using the
+xref:../cli_reference/index.adoc#cli-reference-index[CLI].
+
+When you perform glusterfs storage management operations, the CLI creates/deletes/modifies
+xref:../architecture/additional_concepts/storage.adoc#persistent-volume-claims[persistent volume claim objects]
+that are representations of volumes that can be used by pods. The persistent
+volume claim ask for Kubernetes to provision a
+xref:../architecture/additional_concepts/storage.adoc#persistent-volumes[persistent volume object]
+in a storage backend.
+
+[[storage-grow]]
+== Increase available storage
+
+[IMPORTANT]
+====
+Growing the size of physical disks in glusterfs will not increase the available
+space.
+====
+
+To increase the available space:
+* add a new block device in all 3 glusterfs servers
+* add the new block device in **glusterfs_devices** var in inventory file
+* run the installer
+
+
+For example, supose you have this glusterfs section in inventory file:
+
+----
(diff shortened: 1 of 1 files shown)


tmorriso-rh:
@jmferrer Hi Jose, the team is reviewing this PR. What OpenShift release is this for?

@childsb @gnufied @jsafrane Hi All, can someone from the Storage team review this PR?

@jhou1 @liangxia Hi, can you QE review this PR?

jhou1:
@wkshi Can you please review?

wkshi:
QE haven't tested this scenario, @jarrpa could you please help confirm this?

jarrpa:
....this shouldn't work. Has this actually been run successfully??

jmferrer:
Hi. I tested this scenario in 3.7 1 month ago and in 3.9 two weeks ago..

jmferrer:
Last time I used the glusterfs playbook instead of config playbook. Let me change the PR.

jmferrer:
History reviewed. This PR is valid for 3.9 .

jarrpa:
Having finally had a chance to check the code, this does look like it could work. However, the roles was not designed for this and I worry there may be unforeseen consequences. I would definitely want it verified by QE before merging.

@wkshi If you have a look at this, please make sure in your tests that you verify the heketi topology and that dynamic provisioning still works against that Gluster cluster.

wkshi:
@jarrpa @jmferrer I'd like to confirm which release are we going to support this feature? Is there any Trello Card tracking this? Rerun installer playbooks in installed OCP looks not safe in my view.

jmferrer:
@wkshi , ansible playbooks, in theory, must be idempotent. Each time I want to reconfigure the cluster I change the inventory file and launch the installation playbook.

wkshi:
@jmferrer In theory, yes, it should be. But for openshift-ansible, in some case, it's not.

jarrpa:
@jmferrer Yeah, @wkshi is right: they SHOULD be idempotent but it is not a strict requirement. At present, the GlusterFS playbooks are not guaranteed to be idempotent though they may at times behave that way if you're lucky.

There are currently no official timelines to support this, but it is a general goal of mine to get the playbooks to a (mostly) idempotent state such that this would be guaranteed to be valid.

jmferrer:
I **always** use playbooks when I want to change any parameter, from openshift_node_kubelet_args to openshift_hosted_router_replicas because, Ansible makes less mistakes than I. ;-)

I cannot say that playbooks are not idempotent to a customer. I know that programming idempotent playbooks is very difficult some times and some other times is nearly to be impossible. And I know you are overloaded.

But customers need guides to do basic operations like:
* increase gluster cluster space
* backup/restore gluster volumes
* monitoring volumes usage
Without those three guides is very difficult that containerized glusterfs could be accepted in a big customer.

I will use your playbooks as nearly idempotent. In @jarrpa ,  @wkshi and all openshift-ansible project we trust! :-)

jarrpa:
@jmferrer Heh, thanks for the confidence. :)

While I agree that those are crucial features for our project, we do already have relatively large deployments of CNS in enterprise datacenters. We could potentially reach whole new scales with proper automation, but at least we are not doing too bad for now.

Since you agree to use these in a non-idempotent manner, would it make sense to close this PR? I do have some plans to address this functionality, but it will involve some development in the meantime.

tmorriso-rh:
@jmferrer Bump. Please review the comment from @jarrpa about closing this PR. Thanks.

PR #7779, opened 2018-02-13
Bugzilla 1540832: Add local pv changes

ghost:
https://bugzilla.redhat.com/show_bug.cgi?id=1540832


vikram-redhat:
@tmorriso-rh - there are 4 commits in this PR. We should be squashing commits.

PR #7782, opened 2018-02-14
enables travis cache feature

gaurav-nelson:
From https://docs.travis-ci.com/user/caching/#Bundler

`On Ruby and Objective-C projects, installing dependencies via Bundler can make up a large portion of the build duration. Caching the bundle between builds drastically reduces the time a build takes to run.`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: ruby
+cache: bundler
 rvm:
   - 2.3.1
 gemfile: Gemfile


gaurav-nelson:
@nhr @danmcp PTAL
Will this make our builds faster?
I do not know if there are more configuration options we need to configure. I previously used Circle CI cache builds and they really are fast, and that is why I was looking for this feature in Travis CI.

nhr:
Definitely worth a try. I am +1 for merging this and testing it out.

danmcp:
LGTM

gaurav-nelson:
Thanks @nhr @danmcp 

Please merge and test, I am not sure how to test this properly.

gaurav-nelson:
Thanks @nhr ➕ 

Form the logs I think it saved around 24 secs. Not a huge difference but still a difference.
image

PR #7788, opened 2018-02-14
admin-guide:image-policy: allowed registries for import

### LINKED ISSUE #7123: AllowedRegistriesForImport format is undocumented
It's mentioned here:
https://docs.openshift.org/latest/install_config/master_node_configuration.html#master-config-image-policy-config

But no example or discussion is provided to help people understand how to actually craft a proper definition of allowed registries.  They have to go to the api to understand it:
https://github.com/openshift/origin/blob/00c65897e60f396b716b68011d691f5c0675a4c0/pkg/cmd/server/api/v1/types.go#L425-L437

@miminar since you're working on this logic anyway, can you add some appropriate doc?
It should probably be added either in this section:
https://docs.openshift.org/latest/admin_guide/image_policy.html

or as a peer to that same section.

/cc @bmcelvee

miminar:
Documents feature/bugfix implemented/merged in openshift/origin#17783

Resolves #7123

### DIFF
--- a/admin_guide/image_policy.adoc
+++ b/admin_guide/image_policy.adoc
@@ -12,13 +12,100 @@ toc::[]
 
 == Overview
 
-You can control which images are allowed to run on your cluster using the ImagePolicy
-admission plug-in (currently considered beta). It allows you to control:
+You can control which images can be imported, tagged and/or run in a cluster.
+There are two facilities for this purpose.
+
+xref:#image-policy-configuring-the-image-policy-admission-plug-in[*Allowed
+Registries for import*] - Is an image policy configuration that allows to
+restrict image origins to particular set of external registries. This set of
+rules is applied to any image being imported or tagged into any image stream.
+Therefor any image referencing registry not matched by the rule set will be
+rejected.
+
+xref:#image-policy-testing-image-policy-admission-plug-in[*ImagePolicy
+admission plug-in*] - Lets you specify which images are allowed to be run on
+your cluster using. This is a beta feature. It allows you to control:
 
 - *The source of images*: which registries can be used to pull images
-- *Image resolution*: force pods to run with immutable digests to ensure the image does not change due to a re-tag
-- *Container image label restrictions*: force an image to have or not have particular labels
-- *Image annotation restrictions*: force an image in the integrated container registry to have or not have particular annotations
+- *Image resolution*: force pods to run with immutable digests to ensure the
+  image does not change due to a re-tag
+- *Container image label restrictions*: force an image to have or not have
+  particular labels
+- *Image annotation restrictions*: force an image in the integrated container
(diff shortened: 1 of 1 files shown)


miminar:
@bmcelvee can you please take a look?

bmcelvee:
Thanks, @miminar! I'll go ahead and get this merged today and open a follow-up PR with some minor edits. Looks like this should go into 3.9. Is that correct?

miminar:
That's correct, but wait. I just realized I need to mention one or two things there.

bmcelvee:
Sure thing! 🌞

miminar:
Done. Unless @bparees wants to take another look, it's all in your hands :-)

admin_guide/image_policy.adoc
```diff
@@ -12,13 +12,96 @@ toc::[]
...
-- *Container image label restrictions*: force an image to have or not have particular labels
-- *Image annotation restrictions*: force an image in the integrated container registry to have or not have particular annotations
+- *Image resolution*: force pods to run with immutable digests to ensure the
+  image does not change due to a re-tag
+- *Container image label restrictions*: force an image to have or not have
+  particular labels
+- *Image annotation restrictions*: force an image in the integrated container
+  registry to have or not have particular annotations
+
+== Configuring Registires Allowed for Import
```
bparees: registries typo

admin_guide/image_policy.adoc
```diff
@@ -12,13 +12,96 @@ toc::[]
...
+
+Can be achieved in `*_master-config.yaml_*` under
+`imagePolicyConfig:allowedRegistriesForImport` section as demonstrated in the
+following example. By default, the setting is not present, which means, all
+images are allowed.
```
bparees: that's not true is it?  by default you get the registries we hardcoded as defaults for the whitelist.

bparees (changes requested):

admin_guide/image_policy.adoc
miminar: You're right. It's the ansible-installer that doesn't write this policy into `master-config.yaml`. I totally forgot about `openshift start master --write-config` which generates the setting.

I'll open another PR against openshift-ansible to write the policy by default.

miminar: I just removed the *default* part.

bparees:
lgtm

miminar:
Ansible PR: openshift/openshift-ansible#7788

Once merged, I'll do a doc follow-up with amendments for advanced installation..

miminar:
@bmcelvee it's all yours now :bowing_man:

bmcelvee:
Thanks, @miminar! :sunflower: Do I need to wait for the associated PRs to merge, or is it all set?

PR #7822, opened 2018-02-16
[enterprise-3.6] BZ:1544932 Combined fixes

vikram-redhat:


vikram-redhat:
Merging as Travis is stuck with a big backlog.

PR #7840, opened 2018-02-16
Add IMPORTANT note about what to do if migration playbooks fail

sdodson:
We've had a few people try to power through failed migration attempts and it always results in them getting into a bad state.


vikram-redhat:
@sdodson and @adellape - since we don't have this topic in 3.9 anymore, does this need to go into the 3.7 and below docs?

sdodson:
It needs to be in 3.7 and 3.6.

kalexand-rh:
@sdodson, I edited your note on #8837 and will use that PR to put it on 3.7 and 3.6. Thanks!

PR #7855, opened 2018-02-19
Bug 1508672 Bug 1451370 Clarify Content Related to NFS

bmcelvee:
https://bugzilla.redhat.com/show_bug.cgi?id=1508672
https://bugzilla.redhat.com/show_bug.cgi?id=1451370

### DIFF
--- a/install_config/install/advanced_install.adoc
+++ b/install_config/install/advanced_install.adoc
@@ -513,12 +513,12 @@ which may cause invalid configurations. Example usage:
 |`openshift_docker_options`
 a|This variable configures additional `docker` options within
 *_/etc/sysconfig/docker_*, such as options used in
-xref:../../install_config/install/host_preparation.adoc#managing-docker-container-logs[Managing Container Logs]. 
-Use `json-file` or `journald`. The default is `journald`. 
-Example usage: 
+xref:../../install_config/install/host_preparation.adoc#managing-docker-container-logs[Managing Container Logs].
+Use `json-file` or `journald`. The default is `journald`.
+Example usage:
 ----
 "--log-driver json-file --log-opt max-size=1M --log-opt max-file=3"
-"--log-driver journald" 
+"--log-driver journald"
 ----
 Do not use when
 xref:advanced-install-docker-system-container[running `docker` as a system container].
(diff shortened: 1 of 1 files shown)


jcantrill (approved):
cc @portante

portante (approved):
/lgtm

bmcelvee:
It turns out I only changed the metrics section with my first commit, so I made the same changes in the logging section. Though, I did not change anything in the Configuring Registry Storage section (do you know if I should?).

install_config/install/advanced_install.adoc
```diff
@@ -1644,9 +1655,27 @@ be deleted when the Elasticsearch pod terminates.
...
+[[advanced-installation-logging-storage-dynamic]]
+===== Option A: Dynamic
+
+Use the following variable if your {product-title} environment supports
+xref:../../install_config/persistent_storage/dynamically_provisioning_pvs.adoc#install-config-persistent-storage-dynamically-provisioning-pvs[dynamic volume provisioning] for your cloud provider:
+
+----
+[OSEv3:vars]
+
+openshift_logging_storage_kind=dynamic
```
anpingli: I think it is better to use same variables in https://docs.openshift.com/container-platform/3.7/install_config/aggregate_logging.html .  openshift_logging_storage_kind=dynamic -> openshift_logging_es_pvc_dynamic=true

@sdodson  What do you think?

bmcelvee:
@openshift/team-documentation :sunflower:

install_config/install/advanced_install.adoc
```diff
@@ -1644,9 +1655,27 @@ be deleted when the Elasticsearch pod terminates.
...
+----
+
 [discrete]
 [[advanced-installation-logging-storage-nfs-host-group]]
-===== Option A: NFS Host Group
+===== Option B: NFS Host Group
+
+[NOTE]
+====
+The use of NFS is experimental and not supported in {product-title}.
```
ahardin-rh: @bmcelvee I would consider updating each of these boxes to be IMPORTANT boxes since it is calling out supportability.

install_config/install/advanced_install.adoc
bmcelvee: @anpingli, I think you're correct. I went ahead and updated the references. We still have "openshift_logging_storage_kind=nfs" in the NFS section, but think we should leave those as is for the time being. What do you think?

anpingli: @bmcelvee  Agree, The customer may/allow  use  NFS for test purpose.

PR #7856, opened 2018-02-19
BZ1500614 Revision History

bmcelvee:
https://bugzilla.redhat.com/show_bug.cgi?id=1500614


ahardin-rh:
@bmcelvee We need to also update the Full Revision History page when adding revision histories: https://docs.openshift.com/container-platform/3.6/welcome/revhistory_full.html

bmcelvee:
Thanks, @ahardin-rh. Follow-up PR linked :cherry_blossom:

PR #5709, opened 2017-10-13
Updated commands to `oc adm drain node`

ahardin-rh:
Addresses https://github.com/openshift/openshift-docs/issues/5613

@pravisankar Can you please check the syntax here?

PR #7881, opened 2018-02-21
Added oc adm drain options for 3.9

ahardin-rh:
Added per discussion in https://github.com/openshift/openshift-docs/pull/5709#issuecomment-367043544

@pravisankar PTAL

### DIFF
--- a/admin_guide/manage_nodes.adoc
+++ b/admin_guide/manage_nodes.adoc
@@ -27,7 +27,6 @@ xref:../architecture/infrastructure_components/kubernetes_infrastructure.adoc#no
 == Listing Nodes
 To list all nodes that are known to the master:
 
-====
 ----
 $ oc get nodes
 NAME                        STATUS                     AGE
@@ -35,7 +34,6 @@ master.example.com          Ready,SchedulingDisabled   165d
 node1.example.com           Ready                      165d
 node2.example.com           Ready                      165d
 ----
-====
 
 To only list information about a single node, replace `<node>` with the full
 node name:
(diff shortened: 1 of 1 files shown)


admin_guide/manage_nodes.adoc
```diff
@@ -241,7 +241,7 @@ controller] can be evacuated; the replication controllers create new pods on
 other nodes and remove the existing pods from the specified node(s). Bare pods,
 meaning those not backed by a replication controller, are unaffected by default.
 
-To evacuate one or more nodes:
+To evacuate all or selected pods on one or more nodes:
 
 ----
 $ oc adm drain <node1> <node2>
```
pravisankar: We are saying 'evacuate all or selected pods' but we are not telling how to do that using a command.
We should talk about `--pod-selector` here and show that as optional param to the command.
```
 $ oc adm drain <node1> <node2> [--pod-selector=<pod_selector>]
```
I've assumed '[ ]' is used for optional flags if that is not the case use what ever convention is used in other places.

admin_guide/manage_nodes.adoc
```diff
@@ -285,6 +285,19 @@ is drained):
...
+
+----
+$ oc adm drain <node1> <node2>  --dry-run=true
+----
+
+Instead of specifying specific node names (for example, `<node1> <node2>`), you
+can use the `--selector=<node_selector>` option to evacuate pods on selected
+nodes.
+
+You can also use `--pod-selector=<pod>` to filter pods on the node.
```
pravisankar: This can be skipped if we have already mentioned this option before.

pravisankar (changes requested):

ahardin-rh:
@pravisankar Thanks! I made some edits. Please let me know if more changes are necessary. Thanks again!

pravisankar (approved):
LGTM

PR #7896, opened 2018-02-22
Fix non working example

e-minguez:
The nginx example doesn't work as it requires root permissions. Using the openshift/hello-openshift is also better for self promoting openshift :)

### DIFF
--- a/dev_guide/deployments/kubernetes_deployments.adoc
+++ b/dev_guide/deployments/kubernetes_deployments.adoc
@@ -32,22 +32,22 @@ xref:../../architecture/core_concepts/deployments.adoc#replication-controllers[r
 For example, this definition of a Kubernetes deployment creates a replica set to
 bring up three nginx pods:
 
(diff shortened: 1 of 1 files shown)


cooktheryan (approved):

dev_guide/deployments/kubernetes_deployments.adoc
```diff
@@ -32,22 +32,22 @@ xref:../../architecture/core_concepts/deployments.adoc#replication-controllers[r
 For example, this definition of a Kubernetes deployment creates a replica set to
```
ahardin-rh: Looks like the wording here needs to be updated to match the example below.

ahardin-rh:
@kargakis PTAL

ahardin-rh:
@kargakis I still see a reference to nginx just before the example. Once that is updated, I will merge. Thanks!

ahardin-rh:
To get this moving along, I will do a follow-up edit in a separate PR

PR #7906, opened 2018-02-23
Bug 1543648 - add info for adding users and creating groups

bfallonf:
For: https://bugzilla.redhat.com/show_bug.cgi?id=1543648

First draft. More to come. Will do a bit more to see if the "creating a user" stuff is accurate to all the identities providers and see if the customer case is asking for any more info about the rolebindings.

### DIFF
--- a/admin_guide/manage_users.adoc
+++ b/admin_guide/manage_users.adoc
@@ -14,19 +14,64 @@ toc::[]
 [[managing-users-overview]]
 == Overview
 
+A user is an entity that interacts with the {product-title} API. These can be a
+developer for developing applications or an administrator for managing the
+cluster. Users can be assigned to groups, which set the permissions applied to
+all the group's members. For example, you can give API access to a group, which
+give all members of the group API access.
+
 This topic describes the management of
-xref:../architecture/core_concepts/projects_and_users.adoc#architecture-core-concepts-projects-and-users[user] accounts,
-including how new user accounts are created in {product-title} and how they can
-be deleted.
+xref:../architecture/core_concepts/projects_and_users.adoc#architecture-core-concepts-projects-and-users[user]
+accounts, including how new user accounts are created in {product-title} and how
+they can be deleted.
 
 [[managing-users-adding-a-user]]
-== Adding a User
+== Creating a User
 
-After new users log in to {product-title}, an account is created for that user
-per the xref:../install_config/configuring_authentication.adoc#identity-providers[identity
-provider] configured on the master. The cluster administrator can
-xref:../admin_guide/manage_rbac.adoc#admin-guide-manage-rbac[manage the access level of
-each user].
+The process for creating a user depends on the configured
+xref:../install_config/configuring_authentication.adoc#identity-providers[identity
+provider]. By default, {product-title} uses the `DenyAll` identity provider,
+which denies access for all user names and passwords. 
+
+The following process creates a new user, then adds a role to the user:
+
+. Create the user account depending on your identity provider. This can depend
+on the `mappingmethod` used as part of the identity provider configuration. See
+the
+xref:../install_config/configuring_authentication.adoc#mapping-identities-to-users[Mapping
+Identities to Users] section for more information.
+
+. Give the new user the desired role:
++
+----
+# oc create clusterrolebinding <clusterrolebinding_name> /
+  --clusterrole=<role> --user=<user>
+----
++
+Where the `--clusterrole` option is the desired cluster role. For example, to
+give the new user `cluster-admin` privileges, which gives the user access to
+everything within a cluster:
++
+----
+# oc create clusterrolebinding registry-controller /
+  --clusterrole=cluster-admin --user=admin
+----
++
+For an explanation and list of roles, see the
+xref:../architecture/additional_concepts/authorization.adoc#roles[Cluster Roles
+and Local Roles section of the Architecture Guide].
+
+As a cluster administrator, you can also
+xref:../admin_guide/manage_rbac.adoc#admin-guide-manage-rbac[manage the access
+level of each user].
+
+[NOTE]
+====
+Depending on the identity provider, and on the defined group structure, some
+roles may be given to users automatically. See the
+xref:../install_config/syncing_groups_with_ldap.adoc#install-config-syncing-groups-with-ldap[Synching
+groups with LDAP section] for more information.
+====
 
 
 [[managing-users-viewing-user-and-identity-lists]]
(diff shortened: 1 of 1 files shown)


admin_guide/manage_users.adoc
```diff
@@ -14,19 +14,68 @@ toc::[]
...
+The following process is to create a new user as a system administrator, then
+add a role to the user:
+
+. Create the user account:
++
+----
+# touch /path/to/identityproviers/file /
+  <identity_provider> -b /path/to/identityproviers/file /
+  <username> <password>
+----
```
sferich888: This does not make sense as not every authentication provide has a command that follows this syntax!

sferich888: This also does not apply to the AllowAll provider. 

~~~
oc login -u USER
~~~

Is a better example for this.

admin_guide/manage_users.adoc
```diff
@@ -14,19 +14,68 @@ toc::[]
...
++
+For example, to create a user named `admin` with a password of `redhat` using
+the htpasswd identity provider:
++
+----
+# touch /etc/origin/openshift-passwd /
```
sferich888: Show this as 2 commands, not 1; as you don't break them up syntactically (so that they are correct); PS the touch is not needed. 

Why not just link to https://httpd.apache.org/docs/2.4/programs/htpasswd.html to show the syntax for this?

admin_guide/manage_users.adoc
```diff
@@ -14,19 +14,68 @@ toc::[]
...
+  htpasswd -b /etc/origin/openshift-passwd 
+  admin redhat
+----
+
+. Restart the {product-title} master services to enable the changes:
++
+----
+# systemctl restart atomic-openshift-master-api atomic-openshift-master-controllers
+----
```
sferich888: Is the restart needed, if the auth file is the only file that is touched? (I don't think it is - in the context of htpasswd) I know its not needed if users in ldap or the oauth providers are added.

admin_guide/manage_users.adoc
```diff
@@ -14,19 +14,68 @@ toc::[]
...
+
+. Give the new user the desired role:
++
+----
+# oc adm policy add-cluster-role-to-user <role> <user>
+----
```
sferich888: It might be good to link to the ldap sync docs here, to explain how users (in some authentication providers) can be given some roles automatically based on the group structure by the authentication provider.

admin_guide/manage_users.adoc
```diff
@@ -66,6 +115,40 @@ authentication provider after starting to use {product-title}, the user names
 that overlap will not work because of the entries in the identity list, which
 will still point to the old authentication method.
 
+[[admin-guide-managing-users-creating-groups]]
+== Creating Groups
+
+While a user is an entity making requests to {product-title}, users can be
+organized into one or more groups made up from a set of users. Groups are useful
+for managing many users at one time, such as for authorization policies, or to
+grant permissions to multiple users at once.
```
sferich888: We should link to the ldap docs on syncing groups here, to outline that that process is done automatically and that this is only done for the other providers where this is not done automatically.

admin_guide/manage_users.adoc
```diff
@@ -66,6 +115,40 @@ authentication provider after starting to use {product-title}, the user names
 that overlap will not work because of the entries in the identity list, which
 will still point to the old authentication method.
 
+[[admin-guide-managing-users-creating-groups]]
```
sferich888: This section completes the ask for the BZ!

sferich888 (changes requested):
Looks good, however, I have some nits that need to be worked out before we send this to QE for testing.

bfallonf:
@sferich888 Your questions are valid. I had similar ones in the email. I was playing with `oc create user <username>`, but this didn't create a password associated with a username. Does this change depending on the identity provider used? Would `oc login -u <username>` be a better suggested procedure?

I've updated the rest of the process with your suggestions.

admin_guide/manage_users.adoc
```diff
@@ -14,19 +14,57 @@ toc::[]
...
+. Create the user account depending on your identity provider. This can depend
+on the `mappingmethod` used as part of the identity provider configuration. See
+the
+xref:../install_config/configuring_authentication.adoc#mapping-identities-to-users[Mapping
+Identities to Users] section for more information.
+
+. Give the new user the desired role:
++
+----
+# oc adm policy add-cluster-role-to-user <role> <user>
```
deads2k: I think we should prefer `oc create rolebinding`.  @openshift/sig-security probably has an opinion, but for creation I'm hard pressed to think of a reason for ambiguity in the created rolebinding.

simo5: Yes `oc create clusterrolebinding` is preferred, but if `oc adm policy add-cluster-role-to-user` is "better" for some reason then you should definitely suggest the use of the --rolebinding-name option as well.

enj: I agree with @deads2k and @simo5 - we need to start pushing people towards the explicit commands.  Otherwise they end up with the wrong conceptual model for how RBAC works.

bfallonf:
Thanks @deads2k @simo5 @enj . I've updated to your suggestions.

But is the `oc create clusterrolebinding` command in 3.7? It is barely in the docs, and wasn't on my test machine, but is on a workmate's test minishift test machine.

Also, suggesting this be the best method, but not having `oc create user` be the best method for creating a user is...strange. I know creating users is a bit more fiddly. It's also a bit out of scope for this BZ (which was about creating groups and Eric has given the thumbs up), so I might try move that conversation to email.

Please let me know if there's anything more needed or incorrect for this PR.

admin_guide/manage_users.adoc
```diff
@@ -14,19 +14,58 @@ toc::[]
...
+. Create the user account depending on your identity provider. This can depend
+on the `mappingmethod` used as part of the identity provider configuration. See
+the
+xref:../install_config/configuring_authentication.adoc#mapping-identities-to-users[Mapping
+Identities to Users] section for more information.
+
+. Give the new user the desired role:
++
+----
+# oc create clusterrolebinding --clusterrole=<role> --user=<user>
```
deads2k: this requires a clusterrolebinding name

deads2k: > this requires a clusterrolebinding name

it's an arg to the command.  This should have failed on the CLI without it

admin_guide/manage_users.adoc
```diff
@@ -66,6 +105,47 @@ authentication provider after starting to use {product-title}, the user names
...
+{product-title} so that you can configure groups on one place. This presumes
+that information about your users is in an MDAP server. See the
+xref:../install_config/syncing_groups_with_ldap.adoc#install-config-syncing-groups-with-ldap[Synching
+groups with LDAP section] for more information. If you are not using LDAP, you
+can use the following procedure to manually create groups.
+
+To create a new group:
+
+----
+# oc adm groups new <group_name> <user1> <user2>
```
deads2k: @enj @openshift/sig-security this stands out as a command that should be made into `oc create group`

deads2k:
nit on usage, lgtm otherwise

admin_guide/manage_users.adoc
deads2k: > @enj @openshift/sig-security this stands out as a command that should be made into oc create group

the doc is fine.

bfallonf:
Thanks, all. @openshift/team-documentation PTAL

admin_guide/manage_users.adoc
```diff
@@ -14,19 +14,60 @@ toc::[]
 [[managing-users-overview]]
 == Overview
 
+A user is an entity that interacts with the {product-title} API. These can be a developer for developing applications or a administrator for managing the cluster. Users can be assigned to groups, which set the permissions applied to all the group's members. For example, you can give API access to a group, which give all members of the group API access.
+
```
mburke5678: s/a administrator/an administrator

admin_guide/manage_users.adoc
```diff
@@ -14,19 +14,60 @@ toc::[]
...
+provider]. By default, {product-title} uses the `DenyAll` identity provider,
+which denies access for all user names and passwords. 
 
-After new users log in to {product-title}, an account is created for that user
-per the xref:../install_config/configuring_authentication.adoc#identity-providers[identity
-provider] configured on the master. The cluster administrator can
-xref:../admin_guide/manage_rbac.adoc#admin-guide-manage-rbac[manage the access level of
-each user].
+The following process creates a new user, then adds a role to the user:
+
```
mburke5678: Passive. Maybe "Use the following step to create a new user and add a role to the user." ??

mburke5678:
@bfallonf A few nits. Otherwise LGTM

admin_guide/manage_users.adoc
bfallonf: Hmm I think this one is OK. I don't see how it's passive, and the suggestion is similar.

bfallonf:
Thanks @mburke5678 . I'll merge.

PR #7933, opened 2018-02-26
node_affinity: Delete unpaired callout

Miciah:
Delete an unpaired callout number in the "Preferred Node Affinity Rule" example.


ahardin-rh:
@Miciah Thank you!

No revision history needed.

PR #7942, opened 2018-02-26
Removed formatting from conflict

ahardin-rh:


ahardin-rh:
No revision history needed. Formatting fix.

PR #7963, opened 2018-02-28
Do not recommend using the master branch of openshift-ansible.

abutcher:
The release branch version in the example should probably correspond to the latest official release. @adellape is there a variable for that?


adellape:
@abutcher Technically yes (`{product-version}`), though it sadly won't work here for a few reasons, including that it would be rendered verbatim in a codeblock like this. Since it's just the one spot, I think this is fine as-is.

adellape (approved):

PR #7981, opened 2018-03-01
Remove daemonset from RBAC docs

enj:
Users are given rights to daemonset by default in 3.9 so this section is not necessary.

/assign @ahardin-rh


enj:
@openshift/sig-security

ahardin-rh:
LGTM. Thanks!
No revision history needed.

PR #7983, opened 2018-03-01
Removed link to Granting Users Daemonset Permissions as followup to PR#7981

ahardin-rh:
Follow-up to https://github.com/openshift/openshift-docs/pull/7981

Users are given rights to daemonset by default in 3.9 per @enj


ahardin-rh:
No rev history needed

PR #7990, opened 2018-03-01
Groupify OAuth and User objects

liggitt:
Sweeping docs that reference ungrouped API types


liggitt:
cc @openshift/sig-security

enj (approved):

ahardin-rh:
@liggitt Thanks! is this for 3.9?

liggitt:
Yes

ahardin-rh:
Thanks! 
No revision history needed.

PR #8012, opened 2018-03-02
Bug 1518474 Update documentation to reflect change to json-file

bmcelvee:
Replacing https://github.com/openshift/openshift-docs/pull/7347 because of conflicts. 

https://bugzilla.redhat.com/show_bug.cgi?id=1518474

### DIFF
--- a/install_config/aggregate_logging.adoc
+++ b/install_config/aggregate_logging.adoc
@@ -42,23 +42,11 @@ administrators can view all logs, but application developers can only view logs
 for projects they have permission to view. The stack components communicate
 securely.
 
-Aggregated logging is supported using the `json-file` or `journald` driver in Docker.
-The Docker log driver is set to `journald` as the default for all nodes. See
-xref:fluentd-upgrade-source[Updating Fluentd's Log Source After a Docker Log
-Driver Update] for more information about switching between `json-file` and `journald`.
-Fluentd automatically determines which log driver (`journald` or `json-file`) the container runtime is using.
-
-When the log driver is set to `journald`, Fluentd reads journald logs. When set to `json-file` Fluentd reads from *_/var/log/containers_*.
-
-See xref:../install_config/install/host_preparation.adoc#managing-docker-container-logs[Managing
-Docker Container Logs] for information on `json-file` logging driver options
-to manage container logs and prevent filling node disks.
-
-
-[IMPORTANT]
+[NOTE]
 ====
-If Docker log-driver is set to journald, there is no log rate throttling with the `journald` driver.
-As a result, there is a risk of denial-of-service attacks from rogue containers.
+xref:../install_config/install/host_preparation.adoc#managing-docker-container-logs[Managing
+Docker Container Logs] discusses the use of `json-file` logging driver options
+to manage container logs and prevent filling node disks.
 ====
 
 [[aggregate-logging-pre-deployment-configuration]]
@@ -903,7 +891,7 @@ $ token=$(oc whoami -t)
 through the exposed route:
 +
 ----
-$ curl -k -H "Authorization: Bearer $token" -H "X-Proxy-Remote-User: $(oc whoami)" -H "X-Forwarded-For: 127.0.0.1" https://es.example.test/_cat/indices
+$ curl -k -H "Authorization: Bearer $token" -H "X-Proxy-Remote-User: $(oc whoami)" -H "X-Forwarded-For: 127.0.0.1" https://es.example.test/project.my-project.*/_search?q=level:err | python -mjson.tool
 ----
 
 [[aggregated-fluentd]]
@@ -916,66 +904,15 @@ As part of the OpenShift cluster installation, it is recommended that you add th
 Fluentd node selector to the list of persisted
 xref:../install_config/install/advanced_install.adoc#configuring-node-host-labels[node labels].
 
-Fluentd uses `journald` as the system log source. These are log
-messages from the operating system, Docker, and OpenShift. For
-container logs, Fluentd determines which log driver Docker is using,
-`json-file` or `journald`, and automatically reads the logs from that source.
+Fluentd uses `journald` as the system log source. These are log messages from
+the operating system, the container runtime, and OpenShift.
 
-// [[fluentd-use-journald]]
-// *Having Fluentd Use the Systemd Journal as the Log Source*
-
-// By default, Fluentd reads from `journald` and
-// *_/var/log/containers/<container>.log_* for system logs and container logs,
-// respectively. You can instead use the systemd journal as the log source. There
-// are three inventory parameters available:
-
-// [cols="3,7",options="header"]
-// |===
-// |Parameter
-// |Description
-
-// | `journal-source`
-// |The default is empty, so that when using the systemd journal, Fluentd first looks for
-// *_/var/log/journal_*, and if that is not available, uses *_/run/log/journal_*
-// as the journal source. You can specify `journal-source` with an explicit
-// journal path. For example, if you want Fluentd to always read logs
-// from the transient in-memory journal, set `journal-source`=*_/run/log/journal_*.
-
-// | `openshift_logging_journal_read_from_head`
-// |If this setting is `false`, Fluentd starts reading from the end of the journal,
-// ignoring historical logs. If this setting is `true`, Fluentd starts reading
-// logs from the beginning of the journal.
-// |===
-
-[NOTE]
-====
-As of {product-title} 3.3, Fluentd no longer reads historical log files when
-using the JSON file log driver. In situations where clusters have a large number
-of log files and are older than the EFK deployment, this avoids delays when
-pushing the most recent logs into Elasticsearch. Curator deleting logs are
-migrated soon after they are added to Elasticsearch.
-====
-
-[NOTE]
-====
-It may require several minutes, or hours, depending on the size of your
-journal, before any new log entries are available in Elasticsearch, when using
-`openshift_logging_journal_read_from_head=true`.
-====
-
-[WARNING]
-====
-It is highly recommended that you use the default value for `use-journal`. In
-scenarios where upgrading {product-title} changes the Docker log driver, if
-`use-journal=False` is explicitly specified as part of installation, Fluentd
-still expects to read logs generated using the `json-file` log driver. This
-results in a lack of log ingestion. If this has happened within your logging cluster,
-xref:fluentd-upgrade-source[troubleshoot] it.
-
-See
-xref:fluentd-upgrade-source[Updating Fluentd's Log Source After a Docker Log
-Driver Update] for more information.
-====
+Clean installations of {product-title} 3.9 use `json-file` as the default log
+driver, but environments upgraded from {product-title} 3.7 will maintain their
+existing `journald` log driver configuration. It is recommended to use the
+`json-file` log driver. See xref:fluentd-update-source[Changing the Aggregated
+Logging Driver] for instructions to change your existing log driver
+configuration to `json-file`.
 
 [[fluentd-external-log-aggregator]]
 *Configuring Fluentd to Send Logs to an External Log Aggregator*
@@ -1827,16 +1764,16 @@ $ curl --key /etc/elasticsearch/secret/admin-key \
 [[fluentd-update-source]]
 == Changing the Aggregated Logging Driver
 
-By default, aggregated logging uses the `journald` log driver
-xref:../install_config/install/advanced_install.adoc#configuring-host-variables[unless `json-file` was specified during installation]. You can change the log driver between `journald` and `json-file` as needed.
+For aggregated logging, it is recommended to use the `json-file` log driver.
 
 [IMPORTANT]
 ====
-When using the `json-file` driver, ensure that your Docker version is Docker version *docker-1.12.6-55.gitc4618fb.el7_4 now*
-or later.
+When using the `json-file` driver, ensure that you are using Docker
+version *docker-1.12.6-55.gitc4618fb.el7_4 now* or later.
 ====
 
-Fluentd determines the driver Docker is using by checking the *_/etc/docker/daemon.json_* and *_/etc/sysconfig/docker_* files.
+Fluentd determines the driver Docker is using by checking the
+*_/etc/docker/daemon.json_* and *_/etc/sysconfig/docker_* files.
 
 You can determine which driver Docker is using with the `docker info` command:
 
@@ -1846,7 +1783,7 @@ You can determine which driver Docker is using with the `docker info` command:
 Logging Driver: journald
 ----
 
-To change between `json-file` and `journald` after installation:
+To change to `json-file`:
 
 . Modify either the *_/etc/sysconfig/docker_* or *_/etc/docker/daemon.json_* files.
 +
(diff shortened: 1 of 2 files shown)


install_config/aggregate_logging.adoc
```diff
@@ -910,36 +898,16 @@ As part of the OpenShift cluster installation, it is recommended that you add th
...
-
-// | `openshift_logging_journal_read_from_head`
-// |If this setting is `false`, Fluentd starts reading from the end of the journal,
-// ignoring historical logs. If this setting is `true`, Fluentd starts reading
-// logs from the beginning of the journal.
-// |===
+With Fluentd, it is recommended to use `json-file` as the system log source.
+These are log messages from the operating system, Docker, and OpenShift. For
+container logs, Fluentd determines which log driver Docker is using, `json-file`
+or `journald`, and automatically reads the logs from that source.
```
bmcelvee: Does Docker default to `json-file` now? Should we remove this line?

install_config/aggregate_logging.adoc
```diff
@@ -910,36 +898,16 @@ As part of the OpenShift cluster installation, it is recommended that you add th
...
+
+[NOTE]
+====
+As of {product-title} 3.9, it is recommended to use `json-file` as the system
+log source.
+====
```
bmcelvee: This can be changed to:

"As of {product-title} 3.9, support for `Journald` is being deprecated. It is recommended to use `json-file` as the system log source."

install_config/aggregate_logging.adoc
```diff
@@ -910,36 +898,16 @@ As part of the OpenShift cluster installation, it is recommended that you add th
...
 
 [NOTE]
 ====
```
bmcelvee: Is this note still applicable?

bmcelvee: Same with the following note. Is the variable  `openshift_logging_journal_read_from_head=true` still supported?

install_config/aggregate_logging.adoc
```diff
@@ -1839,8 +1807,8 @@ You can determine which driver Docker is using with the `docker info` command:
 Logging Driver: journald
 ----
 
-To change between `json-file` and `journald` after installation:
- 
+To change to `json-file`:
+
 . Modify either the *_/etc/sysconfig/docker_* or *_/etc/docker/daemon.json_* files.
 +
 For example:
```
bmcelvee: Are these steps for ensuring changing to `json-file` correct? It is my understanding that the step to restart docker is not correct. Please provide guidance to correct it. 

Can the "Updating Fluentd's Log Source After a Docker Log Driver Update" section be removed entirely? 

Is `USE_JOURNAL` deprecated?

bmcelvee:
With the changes we are making here, can we update/remove the Throttling Logs in Fluentd section: https://docs.openshift.com/container-platform/3.7/install_config/aggregate_logging.html#aggregated-fluentd?

bmcelvee:
This is the PR I mentioned in the email this morning, @sferich888 and @sdodson.

install_config/aggregate_logging.adoc
sdodson: Docker has always defaulted to json-file but the installer in 3.4-3.7 overwrote that default and set it to journald. Clean installs of 3.9 will use json-file as the default but environments upgraded from 3.7 will maintain their existing journald configuration.

install_config/aggregate_logging.adoc
sdodson: The restart of docker looks fine.

I don't know the defaults of what fluentd needs to do on transition. I feel like @richm said that current versions automatically detect what's in use but I'm not sure.

richm: USE_JOURNAL is deprecated.  Fluentd assumes both /etc/sysconfig/docker and /etc/docker/daemon.json are mounted into the fluentd pod from the host.  Fluentd will look at these config files to see which log driver docker is using.

richm: Fluentd has to be restarted if the docker log driver changes.

richm: But - there is a big caveat here - if you are using json file, and switch to journald, and you still have logs in json files - when fluentd is restarted, fluentd will not be able to lookup kubernetes on those json-file logs, and will not be able to do json parsing on the embedded log field on those logs.  If you are going to change the docker log driver, best to ensure all of the json-file logs are drained first.  Similarly with switching from journald to json-file - if there are any container logs still in the journal, once fluentd is restarted, fluentd will not be able to add kubernetes metadata to those logs, nor will be able to parse the embedded json in the MESSAGE field.

install_config/aggregate_logging.adoc
```diff
@@ -1868,15 +1807,15 @@ systemctl restart docker
 
 . Update the xref:fluentd-upgrade-source[Fluentd log source].
 
+. Restart the Fluentd:
++
+----
+systemctl restart fluentd
```
bmcelvee: What is the correct command to restart Fluentd? I copied the command to restart Docker as a placeholder. 

I found the following in Fluentd docs but don't think it's correct:
"sudo service td-agent restart"

install_config/aggregate_logging.adoc
bmcelvee: Thanks @sdodson and @richm! 

I'm removing instructions for switching to/using `journald`, so the caveat *should* be okay. I removed the `USE_JOURNAL` content from the "Updating Fluentd's Log Source After a Docker Log Driver Update" section, but left the steps for now.

bmcelvee:
@portante and @jcantrill PTAL

install_config/aggregate_logging.adoc
richm: NOTE: Restarting fluentd on many nodes at once will create a huge load on the Kubernetes scheduler, so exercise caution about following the directions below if you have more than a few dozen nodes.

To restart fluentd on *all* nodes, use `oc label node -l logging-infra-fluentd=true --overwrite logging-infra-fluentd=false` to shutdown fluentd on all nodes where it is running, then use `oc get ds -l component=fluentd` to make sure they have all been shutdown, then `oc label node -l logging-infra-fluentd=false --overwrite logging-infra-fluentd=true` to restart fluentd on those nodes.

To restart fluentd on one or a set of nodes:
`oc get nodes -l logging-infra-fluentd=true` to list the nodes where fluentd is running.  Then for each node, use `oc label node $node logging-infra-fluentd-` to unlabel the node and turn off fluentd.  Use `oc get ds -l component=fluentd` to see if those fluentd have been shutdown.  Then for each node, use `oc label node $node logging-infra-fluentd=true` to start up fluentd on those nodes.

install_config/aggregate_logging.adoc
```diff
@@ -910,65 +898,16 @@ As part of the OpenShift cluster installation, it is recommended that you add th
...
-scenarios where upgrading {product-title} changes the Docker log driver, if
-`use-journal=False` is explicitly specified as part of installation, Fluentd
-still expects to read logs generated using the `json-file` log driver. This
-results in a lack of log ingestion. If this has happened within your logging cluster,
-xref:fluentd-upgrade-source[troubleshoot] it.
-
-See
-xref:fluentd-upgrade-source[Updating Fluentd's Log Source After a Docker Log
-Driver Update] for more information.
+As of {product-title} 3.9, support for `journald` is being deprecated. It is
```
jcantrill: @portante @richm @tkatarki I don't think this is EXACTLY correct.  We still support journal right.  The caveat is:

* We ALWAYS gather node logs from journal
* App logs CAN come from journal but there are limitations.  We recommand json

portante: We are deprecating support for the docker `journald` log driver, we are not deprecating support for `journald` itself, as we always collect operational logs from it.  I don't think we should continue to support and verify that we can extract container logs from `journald`.

richm: @portante Then we need to officially deprecate it, and go through the official deprecation procedure.

install_config/aggregate_logging.adoc
bmcelvee: Thanks, @richm! I updated the steps and added a warning about the load on the Kubernetes scheduler.

richm (approved):

install_config/aggregate_logging.adoc
```diff
@@ -916,63 +904,15 @@ As part of the OpenShift cluster installation, it is recommended that you add th
...
-// journal path. For example, if you want Fluentd to always read logs
-// from the transient in-memory journal, set `journal-source`=*_/run/log/journal_*.
-
-// | `openshift_logging_journal_read_from_head`
-// |If this setting is `false`, Fluentd starts reading from the end of the journal,
-// ignoring historical logs. If this setting is `true`, Fluentd starts reading
-// logs from the beginning of the journal.
-// |===
+With Fluentd, it is recommended to use `json-file` as the system log source.
+These are log messages from the operating system, Docker, and OpenShift. For
```
portante: Does it makes sense to keep "system log source"?  This seems like it might confuse folks.

The docker log driver is only about how container logs are delivered, right?

Then the next sentence seems like it needs revision as "These" seems out of place, but also that "operating system" and "Docker", and "OpenShift" logs don't necessarily go through `json-file` log driver.

Perhaps this make sense to other folks?

install_config/aggregate_logging.adoc
portante: If we are clear that logging will use whatever Docker is configured for the logging driver, that seems sufficient, doesn't it?

Can we make this section stronger to cover what @sdodson says that since the installer does not force Docker to use a certain log driver, it will work with whatever Docker is currently using, that the user should be aware that they may have to take specific actions to change the Docker configuration to `json-file`?

install_config/aggregate_logging.adoc
```diff
@@ -1873,16 +1813,67 @@ systemctl restart docker
 ----
```
portante: For the above `systemctl restart docker` command, won't that affect a running OpenShift environment negatively?  Won't the user want to cordon off the node first, make sure pods are drained properly to avoid service disruption?

install_config/install/advanced_install.adoc
```diff
@@ -508,11 +508,10 @@ which may cause invalid configurations. Example usage:
 a|This variable configures additional `docker` options within
 *_/etc/sysconfig/docker_*, such as options used in
 xref:../../install_config/install/host_preparation.adoc#managing-docker-container-logs[Managing Container Logs].
-Use `json-file` or `journald`. The default is `journald`.
+Use `json-file`.
```
portante: How about: "The following example shows the configuration of Docker to use the `json-file` log driver, where Docker will rotate between three (3) 1 MB log files."

portante (changes requested):

install_config/aggregate_logging.adoc
bmcelvee: I removed the line about deprecating support for the `journald` log driver here, since there seemed to be some value in it from the email conversation. I am leaving the notes and content around suggesting the use of the `json-file` log driver.

install_config/aggregate_logging.adoc
bmcelvee: If it makes sense to remove the log source sentence ("Fluentd uses `journald` as the system log source."), I can. Is it important for readers to know that the log source is `journald`? And is the log source still `journald` if the log driver is `json-file`? 

I've looked into this a bit and can't seem to find a definitive answer.

install_config/aggregate_logging.adoc
bmcelvee: Thanks! I incorporated the explanation here and recommended that `json-file` log driver is used. (Will push the changes momentarily).

install_config/aggregate_logging.adoc
richm: How about just this:
```
Fluentd uses journald as the system log source.  These are log messages from the operating system, the container runtime, and OpenShift.
```

bmcelvee: Thanks, @richm! Updating now.

bmcelvee:
Here are the preview builds: 
Aggregating Container Logs - http://file.rdu.redhat.com/~bmcelvee/031518/BZ1518474-new/install_config/aggregate_logging.html
Advanced Installation - http://file.rdu.redhat.com/~bmcelvee/031518/BZ1518474-new/install_config/install/advanced_install.html#configuring-host-variables

install_config/aggregate_logging.adoc
```diff
@@ -1874,15 +1811,66 @@ systemctl restart docker
...
+----
+
+. Restart Fluentd on all nodes:
++
+----
+$ oc label node -l logging-infra-fluentd=false --overwrite logging-infra-fluentd=true
+----
+
 [[fluentd-upgrade-source]]
 == Updating Fluentd's Log Source After a Docker Log Driver Update
```
bmcelvee: Since step 2 of this section, "Updating Fluentd’s Log Source After a Docker Log Driver Update" provides instructions on using `USE_JOURNAL`, should it be removed or deprecated?

richm: `USE_JOURNAL` is ignored.  The step is just to get rid of it, to avoid any future confusion.  A simpler way to get rid of it:
```
oc set env -n logging ds/logging-fluentd USE_JOURNAL-
```

richm:
This currently doesn't work:
```


Using the token previously configured, you should be able access Elasticsearch through the exposed route:

$ curl -k -H "Authorization: Bearer $token" -H "X-Proxy-Remote-User: $(oc whoami)" -H "X-Forwarded-For: 127.0.0.1" https://es.example.test/_cat/indices

```
users do not have permission to view `/_cat/indices`.  Instead, try this:
```
$ curl -k -H "Authorization: Bearer $token" -H "X-Proxy-Remote-User: $(oc whoami)" -H "X-Forwarded-For: 127.0.0.1" https://es.example.test/project.my-project.*/_search?q=level:err | python -mjson.tool
```

richm:
This is not correct:
```


Reducing the Number of Connections from Fluentd to the API Server

With mux, you can deploy N number of mux services, where N is fewer than the number of nodes. Each Fluentd is configured with USE_MUX_CLIENT=1. This tells Fluentd to send the raw logs to mux with no filtering and no Kubernetes metadata filtering, which involves connections to the API server. You can perform all of the processing and Kubernetes metadata filtering with mux.
```
There is no `USE_MUX_CLIENT` nor `openshift_logging_use_mux_client`.  Instead, there is `MUX_CLIENT_MODE` and `openshift_logging_mux_client_mode`.   These must be set in order for fluentd to send logs through mux instead of directly to Elasticsearch.  I don't think we should document `MUX_CLIENT_MODE`, only the ansible parameter `openshift_logging_mux_client_mode`.  See https://github.com/openshift/openshift-ansible/tree/master/roles/openshift_logging#mux---secure_forward-listener-service

richm:
```

openshift_logging_mux_default_namespaces | The default is mux-undefined. The first value in the list is the namespace to use for undefined projects, followed by any additional namespaces to create by default. Usually, you do not need to set this value.
-- | --
```
Should mention that this is an array valued parameter which is a list of namespaces to create, and that users should not usually set this - the default values should usually be used.

bmcelvee:
Thanks, @richm! I fixed the mux content in another BZ (if you'd like I can send it to you), and fixed the Elasticsearch step in this PR.

richm:
lgtm

bmcelvee:
@anpingli PTAL

install_config/aggregate_logging.adoc
```diff
@@ -1875,39 +1812,68 @@ systemctl restart docker
...
+. Restart Fluentd on all nodes:
 +
 ----
-$ oc patch daemonset/logging-fluentd \
-     -p '{"spec":{"template":{"spec":{"containers":[{"name":"fluentd-elasticsearch","env":[{"name": "USE_JOURNAL", "value":""}]}]}}}}'
+$ oc label node -l logging-infra-fluentd=false --overwrite logging-infra-fluentd=true
 ----
 
-. Relabel your nodes to schedule Fluentd deployments:
+. Refresh the Fluentd pod:
```
anpingli: The fluentd pods will be started automatically.   Replace this step with 'Verify Fluentd is on'
oc get pods -l component=fluentd

install_config/aggregate_logging.adoc
```diff
@@ -1875,39 +1812,68 @@ systemctl restart docker
...
+----
+$ oc label node -l logging-infra-fluentd=true --overwrite logging-infra-fluentd=false
+----
 
-. Remove the label from nodes where Fluentd is deployed:
+. Verify Fluentd is off:
 +
 ----
-$ oc label node --all logging-infra-fluentd- <1>
+$ oc get ds -l component=fluentd
```
anpingli: Verify by List pods 
-> oc get pods -l component=fluentd

install_config/aggregate_logging.adoc
anpingli: I can watched logs in /var/log/containers after changed driver from journald to json-file without drained node.  No sure if there are data loss.

richm: @anpingli what will happen in that case is that there may still be old container logs in the journald.  However, once fluentd is restarted and has detected that docker is now using json-file, fluentd will no longer perform kubernetes metadata annotation of those log records from the journal, so they won't get fields such as kubernetes.namespace_id, and will end up in the .orphaned index.

install_config/aggregate_logging.adoc
richm: `oc get ds -l component=fluentd` will give you a summary - which is probably more useful if you have more than a terminal page full of fluentd pods.

bmcelvee:
If it is okay with everyone, I would like to go ahead and merge this PR today. If any addition edits are required, I will open a follow-up PR. Thanks! :sun_with_face:

install_config/aggregate_logging.adoc
```diff
@@ -1875,39 +1812,68 @@ systemctl restart docker
 
 . Update the xref:fluentd-upgrade-source[Fluentd log source].
```
vikram-redhat: Hey @bmcelvee - should this link to fluentd-upgrade-source still stay in this file? It points to the section heading that is just next (and deleted). The other xrefs to fluentd-upgrade-source have been deleted from this file, so I think that may have remained in error. But if we remove it, does the rest of the text still make sense?

bmcelvee: Thanks for catching that @vikram-redhat! I opened a follow-up PR: https://github.com/openshift/openshift-docs/pull/8303. I also clicked through all of the links in the topic to make sure they're correct.

There's one that points to manual upgrades in the following section: 
https://docs.openshift.com/container-platform/3.7/install_config/aggregate_logging.html#aggregate-logging-performing-elasticsearch-maintenance-operations

Should I update it to point to automated upgrades? (The link still works)

vikram-redhat: Thanks @bmcelvee for the fix. I am not sure about the manual upgrades link. I am not sure if that is remaining for 3.9. If it is not remaining, then yeah you should point to the automatic upgrades.

PR #8104, opened 2018-03-12
Add a paragraph about cleaning the registry caches

legionus:
@bparees please review

### DIFF
--- a/admin_guide/pruning_resources.adoc
+++ b/admin_guide/pruning_resources.adoc
@@ -172,8 +172,8 @@ xref:../architecture/additional_concepts/authorization.adoc#roles[cluster role]
 ifdef::openshift-enterprise,openshift-origin[]
 [NOTE]
 ====
-Pruning images removes data from the integrated registry. For this operation to work
-properly, ensure your
+Pruning images removes data from the integrated registry unless `--prune-registry=false`
+is used (see below). For this operation to work properly, ensure your
 xref:../install_config/registry/extended_registry_configuration.adoc#docker-registry-configuration-reference-storage[registry is configured] with `*storage:delete:enabled*` set to *true*.
 ====
 endif::[]
@@ -184,6 +184,27 @@ Pruning images with the `--namespace` flag does not remove images, only image
 streams. Images are non-namespaced resources. Therefore, limiting pruning to a
 particular namespace makes it impossible to calculate their current usage.
 ====
+
+It's important to know that by default the integrated registry caches blobs metadata
+to reduce the number of requests to the storage and increase the speed of processing
+the request. Pruning doesn't update the integrated registry cache. This means that images
+that are pushed after pruning and that contain pruned layers will be broken because
+the pruned layers that have metadata in the cache will not be pushed, resulting in those
+layers being missing after the push is complete. Therefore it is necessary to clear
+the cache after pruning. This can be accomplished by redeploying the registry:
+
+----
+$ oc rollout latest dc/docker-registry
+----
+
+If the integrated registry uses a
+xref:../install_config/registry/extended_registry_configuration.adoc#docker-registry-configuration-reference-redis[redis cache]
+then you need to clean the database manually.
+
+If redeploying the registry after pruning is not acceptable to you,
+then you must
+xref:../install_config/registry/extended_registry_configuration.adoc#docker-registry-configuration-reference-cache[permanently disable the cache]
+instead.
 
 .Prune Images CLI Configuration Options
 [cols="4,8",options="header"]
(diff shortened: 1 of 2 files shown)


admin_guide/pruning_resources.adoc
```diff
@@ -185,6 +185,23 @@ streams. Images are non-namespaced resources. Therefore, limiting pruning to a
 particular namespace makes it impossible to calculate their current usage.
 ====
 
+It's important to know that by default the integrated registry caches blobs metadata
+to reduce the number of requests to the storage and increase the speed of processing
+the request. Pruning doesn't update the integrated registry cache. It means that images
```
bparees: s/It means/This means/

admin_guide/pruning_resources.adoc
```diff
@@ -185,6 +185,23 @@ streams. Images are non-namespaced resources. Therefore, limiting pruning to a
...
+that will be pushed after pruning and that contain pruned layers will be broken.
```
bparees: s/that will be pushed/that are pushed/

s/will be broken/will be broken because the pruned layers that have metadata in the cache will not be pushed, resulting in those layers being missing after the push is complete."

admin_guide/pruning_resources.adoc
```diff
@@ -185,6 +185,23 @@ streams. Images are non-namespaced resources. Therefore, limiting pruning to a
...
+So do this after the pruning is complete. You can run to wipe the cache:
```
bparees: "Therefore it is necessary to clear the cache after pruning.  This can be accomplished by redeploying the registry:"

admin_guide/pruning_resources.adoc
```diff
@@ -185,6 +185,23 @@ streams. Images are non-namespaced resources. Therefore, limiting pruning to a
...
+
+----
+$ oc rollout latest dc/docker-registry
+----
+
+In case that the integrated registry uses a
```
bparees: s/In case that the/If the/

admin_guide/pruning_resources.adoc
```diff
@@ -185,6 +185,23 @@ streams. Images are non-namespaced resources. Therefore, limiting pruning to a
...
+xref:../install_config/registry/extended_registry_configuration.adoc#docker-registry-configuration-reference-redis[redis cache]
+then you need to clean the database manually.
+
+If for some reason this does not work for you, then you need to turn off all caching
+in the integrated registry.
```
bparees: unless we have a reason to expect this to not work, i would just delete this sentence.  If we do have a reason to expect this might not work, then we should be linking to instructions on how to turn off the registry caching.

(and are we suggesting turning it off permanently, or turning it off and then back on?)

bparees (changes requested):

admin_guide/pruning_resources.adoc
legionus: @bparees We do not have an instruction to disable the cache, but for users who can not afford to stop the registry this is the only way out. This option is not mentioned anywhere.

If the user has several registry pods that can not be stopped and if he takes care of the data consistency, then he must permanently turn off the cache. Depending on the storage, the loss in speed can be compensated. 

I believe that we should hide such an important option.

bparees: @legionus ok, i think i misunderstood your "if this does not work for you" statement.  I read it as "if restarting the registry does not clear the cache for some reason, then try disabling the cache"

I see now what you meant is "if redeploying the registry after pruning is not acceptable to you, then you must permanently disable the cache instead".

Can you try to make that clearer?

And if we don't have instructions for how to do it, then it seems like we need to create some, or point to some, because that's going to be the next question a user asks when they read this section.

legionus: > Can you try to make that clearer?

@bparees Sure. Forgive my poor English.

legionus:
@bparees comments addressed.

bparees:
looks great, thanks @legionus!

@openshift/team-documentation ptal/merge

bmcelvee:
@legionus LGTM :snowflake: 

Looks like this only applies to OCP, do you know which versions?

legionus:
> Looks like this only applies to OCP, do you know which versions?

@bparees @bmcelvee What is OCP ?

This documentation is valid from 3.9.

But with some modification this is valid for all versions we have. In older versions, we do not have a simple way to disable the application cache, but it's still possible to speed up the expiration of the data in it.

bparees:
OCP is openshift container platform. The product you work on.

bparees:
Brandi is confirming these docs do not apply to openshift online or openshift dedicated, which they don't, because in those environments customers do not perform administrative tasks like pruning.

bmcelvee:
Thanks, @legionus and @bparees!

vikram-redhat:
@bmcelvee - did this not require a QE review?

PR #8099, opened 2018-03-12
Added documentation on how to enable Ingres Object support

knobunc:
This PR documents how to enable the experimental Ingress Object
support in the OpenShift haproxy-based router.

Please note that this support is experimental and should not be merged
into OSE yet.

PR #8109, opened 2018-03-13
Added info for routes to accept ingress objects

bfallonf:
For #8099 

Thanks, @knobunc in the end, I thought it was better to create a new PR rather than merging and editing. I have some questions, I'll ask them inline.

NOTE: this is not to be cherrypicked into another branch until this is found to be a supported procedure. Ben, do you know when this will be?

### DIFF
--- a/admin_guide/managing_networking.adoc
+++ b/admin_guide/managing_networking.adoc
@@ -144,6 +144,40 @@ $ systemctl restart atomic-openshift-master-api atomic-openshift-master-controll
 
 ----
 
+[[admin-guide-managing-networking-ingress]]
+== Setting up the HAProxy router to accept ingress objects
+
+By default, the HAProxy router only accepts route objects. You can set up the
+HAProxy router as a Kubernetes
+link:https://kubernetes.io/docs/concepts/services-networking/ingress/[Ingress
+Controller] so that it accepts both route and ingress objects. If configured,
+the same rules apply for claiming a domain for a namespace based on the creation
+time of the object. For example, the oldest route winning against any other
+claimants wanting to use the same namespace.
+
+For example, ownership of routes is based on the claims made for that host name with the same namespace. The oldest route wins against any other claims. 
+
+While this makes the router compatible with Kubernetes ingress
+objects, some caveats exist:
+
+* Ingress objects store the keys and certificates in secrets, so the router needs permission to read all secrets in the system.
+* Ingress objects only support edge termination for *https* routes.
+
+To configure an existing router to have Ingress support (assuming the default
+name of `router` for the deployment configuration and the service-account):
+
+. Set the `ROUTER_ENABLE_INGRESS` environment variable to `true`:
++
+----
+$ oc env dc router ROUTER_ENABLE_INGRESS=true`
+----
+
+. Add the cluster-admin role to the router:
++
+----
+$ oc adm policy add-role-to-user cluster-admin router`
+----
+
 [[admin-guide-controlling-egress-traffic]]
 == Controlling Egress Traffic
 
--- a/architecture/topics/router_environment_variables.adoc
+++ b/architecture/topics/router_environment_variables.adoc
@@ -45,6 +45,7 @@ connections (and any time HAProxy is reloaded), the old HAProxy processes
 will "linger" around for that period. xref:time-units[(TimeUnits)]
 |`ROUTER_DENIED_DOMAINS` | | A comma-separated list of domains that the host name in a route can not be part of. No subdomain in the domain can be used either. Overrides option `ROUTER_ALLOWED_DOMAINS`.
 |`ROUTER_ENABLE_COMPRESSION`| | If `true` or `TRUE`, compress responses when possible.
+|`ROUTER_ENABLE_INGRESS`| | If `true` or `TRUE`, look at both Ingress objects and Route objects.
 |`ROUTER_LISTEN_ADDR`| 0.0.0.0:1936 | Sets the listening address for xref:../../install_config/router/default_haproxy_router.adoc#exposing-the-router-metrics[router metrics].
 |`ROUTER_LOG_LEVEL` | warning | The log level to send to the syslog server.
 |`ROUTER_MAX_CONNECTIONS`| 20000 | Maximum number of concurrent connections.


admin_guide/managing_networking.adoc
```diff
@@ -144,6 +144,36 @@ $ systemctl restart atomic-openshift-master-api atomic-openshift-master-controll
 
 ----
 
+[[admin-guide-managing-networking-ingress]]
+== Setting up the HAProxy router to accept ingress objects
+
+By default, the HAProxy router only accepts route objects. You can set up the
```
bfallonf: In order to describe the comparison, I mentioned the router only accepts route objects. Is this accurate?

admin_guide/managing_networking.adoc
```diff
@@ -144,6 +144,36 @@ $ systemctl restart atomic-openshift-master-api atomic-openshift-master-controll
...
+HAProxy router as a Kubernetes
+link:https://kubernetes.io/docs/concepts/services-networking/ingress/[Ingress
+Controller] so that it accepts both route and ingress objects. If configured,
+the same rules apply for claiming a domain for a namespace based on the creation
+time of the object.
```
bfallonf: I'm not sure what the above means. Can this be elaborated on? Is this listed somewhere else in the docs (it mentions "same rules")?

admin_guide/managing_networking.adoc
```diff
@@ -144,6 +144,36 @@ $ systemctl restart atomic-openshift-master-api atomic-openshift-master-controll
...
+
+While this makes the router compatible with Kubernetes ingress
+objects, some caveats exist:
+
+* Ingress objects store the keys and certificates in secrets, so the router needs permission to read all secrets in the system.
+* Ingress objects only support edge termination for `https` routes.
+
+To configure an existing router to have Ingress support (assuming the default name of `router` for the deployment configuration and the ):
```
bfallonf: This sentence trailed off. Do you remember what was supposed to be here?

admin_guide/managing_networking.adoc
knobunc: That is correct.

admin_guide/managing_networking.adoc
knobunc: Hm... the best I can find is a tangential reference in install_config/router/default_haproxy_router.adoc at arond line 1218.

admin_guide/managing_networking.adoc
knobunc: Oops.  "and the service-account"

knobunc (changes requested):

knobunc:
The supported procedure may be https://github.com/openshift/origin/pull/18658 which if we can get consensus should be in 3.10.

bfallonf:
Thanks @knobunc . We can look at this again for 3.10.

In the meantime, let me know if there's anything more. @openshift/team-documentation PTAL

admin_guide/managing_networking.adoc
```diff
@@ -144,6 +144,39 @@ $ systemctl restart atomic-openshift-master-api atomic-openshift-master-controll
...
+
+By default, the HAProxy router only accepts route objects. You can set up the
+HAProxy router as a Kubernetes
+link:https://kubernetes.io/docs/concepts/services-networking/ingress/[Ingress
+Controller] so that it accepts both route and ingress objects. If configured,
+the same rules apply for claiming a domain for a namespace based on the creation
+time of the object. For example, ownership of routes is based on the claims made
+for a host name with the namespace with the oldest route winning against any
+other claimants.
+
```
mburke5678: This sentence seems awkward (or a run on?) because of the two _with the_ phrases. But, I see that the same sentence is in the Install Guide, so maybe I am mistaken.
https://docs.openshift.com/container-platform/3.7/install_config/router/default_haproxy_router.html#using-wildcard-routes.

admin_guide/managing_networking.adoc
```diff
@@ -144,6 +144,39 @@ $ systemctl restart atomic-openshift-master-api atomic-openshift-master-controll
...
+While this makes the router compatible with Kubernetes ingress
+objects, some caveats exist:
+
+* Ingress objects store the keys and certificates in secrets, so the router needs permission to read all secrets in the system.
+* Ingress objects only support edge termination for `https` routes.
+
```
mburke5678: Should the _https_ be bolded with asterisks?

mburke5678:
@bfallonf A couple of nits. Otherwise LGTM

bfallonf:
Thanks @mburke5678 I've made changes as you suggested. I'll merge this.

Note that this is only for Origin as of now, but I'll continuously poke @knobunc until this is ready to be cherry-picked into the OCP branches.

vikram-redhat:
@bfallonf - please make sure to get a QE review before merging.

bfallonf:
@vikram-redhat I merged before I saw your comment...

@bmeng Can I please get a QA of this PR? If there's issues, please let me know. I'll create another PR if needed.

admin_guide/managing_networking.adoc
```diff
@@ -144,6 +144,40 @@ $ systemctl restart atomic-openshift-master-api atomic-openshift-master-controll
...
+. Set the `ROUTER_ENABLE_INGRESS` environment variable to `true`:
++
+----
+$ oc env dc router ROUTER_ENABLE_INGRESS=true`
+----
+
+. Add the cluster-admin role to the router:
++
+----
+$ oc adm policy add-role-to-user cluster-admin router`
```
lihongan: Below two role should be added to sa router, `oc adm policy add-role-to-user cluster-admin router` is invalid.
```
oc adm policy add-cluster-role-to-user cluster-reader system:serviceaccount:default:router
oc adm policy add-cluster-role-to-user system:openshift:controller:service-serving-cert-controller system:serviceaccount:default:router
```

lihongan:
And another issue is how haproxy handle the situation if ingress and route claim the same hostname ?
From my testing, it shows only the first created resource will be served by haproxy.
@knobunc could you please help confirm ? if that true, then we should also write down them here I think.

admin_guide/managing_networking.adoc
bmeng: maybe `oc adm policy add-cluster-role-to-user cluster-reader -z router` also work here.

knobunc:
Yes, internally ingress objects and route objects get turned into the same data structure in the router, so there is no real difference except that an ingress object can turn into multiple objects.  Once that has been done, the same rules apply.  Then the object that with the earliest creation date claims a hostname for a namespace, it doesn't matter if it is a route or ingress.  BUT... since an ingress object could have two hostnames, one hostname can be claimed by the ingress and be active, but the other might not be able to claim, and be ignored.

There are docs in default_haproxy_router.adoc about the claim rules.  But they seem to be under the wildcard information.  It would be good to make them a little more prominent.

knobunc:
@mburke5678 PLEASE NOTE that this is likely only going to be Origin docs.  We have a better solution coming that will supersede all of this for 3.10 so we don't want to support it in OSE for 3.9 just to replace it.

mburke5678:
@bfallonf See Ben's notes ^^

bfallonf:
Thanks @lihongan @bmeng @knobunc .

I've created #8337 as a followup. Please let me know if you have any further thoughts.

PR #5000, opened 2017-08-11
Arch/Networking edits

bfallonf:
PTAL @openshift/team-documentation 

Any thoughts on how I'm going about this are greatly appreciated.

PR #8213, opened 2018-03-19
Add some explanation of hostsubnets that span multiple octets

JacobTanenbaum:
Customer encountered undocumented behaviour regarding the allocation of hostsubnet addresses that span multiple octets. It is well documented in the code itself exposing that information should clear up issues in the future.

Bug [1554939](https://bugzilla.redhat.com/show_bug.cgi?id=1554939)

### DIFF
--- a/architecture/additional_concepts/sdn.adoc
+++ b/architecture/additional_concepts/sdn.adoc
@@ -65,6 +65,15 @@ given node is allocated 510 addresses that it can assign to the containers
 running on it. The size and address range of the cluster network are
 configurable, as is the host subnet size.
 
+[OpenShift SDN]
+====
+In order to make it easeier for humans to follow if a subnet extends into the next higher octet,
+we rotate the bits of the subnet number so subnets with all 0s in the shared octet are allocated
+first. So if the given network is "10.1.0.0/16" hostsubnetlength=6 then 10.1.0.0/26, 10.1.1.0/26,
+through 10.1.255.0/26 are allocated before the system fills in the gaps with 10.1.0.64/26,
+10.1.1.64/26, etc.
+====
+
 Note that OpenShift SDN on a master does not configure the local (master)
 host to have access to any cluster network. Consequently, a master host does not
 have access to pods via the cluster network, unless it is also running as a


JacobTanenbaum:
@knobunc PTAL

architecture/additional_concepts/sdn.adoc
```diff
@@ -65,6 +65,14 @@ given node is allocated 510 addresses that it can assign to the containers
 running on it. The size and address range of the cluster network are
 configurable, as is the host subnet size.
 
+[NOTE]
+====
+If a subnet extends into the next higher octet, we rotate the bits of the subnet number so subnets
```
knobunc: We should probably say that this is a behavior of the openshift SDN plugins (as the next paragraph says)

Also, we should probably give the rationale for this (i.e. we do it so that the SDN ip addresses for the nodes are all the .1 addresses to make it easier for humans to follow).

knobunc (changes requested):

bfallonf:
Thanks @JacobTanenbaum . I can do a followup. Do you know when this was introduced? The BZ is set to 3.6, but should I backport it any earlier?

architecture/additional_concepts/sdn.adoc
```diff
@@ -65,6 +65,15 @@ given node is allocated 510 addresses that it can assign to the containers
 running on it. The size and address range of the cluster network are
 configurable, as is the host subnet size.
 
+[OpenShift SDN]
+====
+In order to make it easeier for humans to follow if a subnet extends into the next higher octet,
```
knobunc: easier

knobunc: follow, (add the ,)

knobunc: no comma after octet

knobunc (changes requested):

bfallonf:
Thanks @JacobTanenbaum . Not sure what happened, but this file shouldn't exist. I took it out in #5000 and made the networking section it's own modular thing. I've moved the file there with #8523 , and deleted the file. Please let me know if there's any thoughts.

I'll close this PR.

PR #8243, opened 2018-03-20
style changes

kealexan:
I created this PR to see how easy it was for a user who is not on the team to create one.


kalexand-rh:
This was a test PR that does not need to be merged.

PR #8253, opened 2018-03-20
bug 1558200 removing redundant text

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1558200


kalexand-rh:
@bmcelvee peer reviewed this one in person, and we agree that it's small enough to not need QE.

PR #8315, opened 2018-03-21
Bug 1552549, removed duplicated systemctl command

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1552549


ahardin-rh:
cherrypick enterprise-3.6

PR #8314, opened 2018-03-21
Added info on the need for port number with openshift_docker_additional_registries

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1558646

PR #8338, opened 2018-03-22
Bug 1558310 Added variables to Configuring a registry location

bfallonf:
For: https://bugzilla.redhat.com/show_bug.cgi?id=1558310

Unfinished, and will fill in the blanks and wait to see if this needs to be in known issues.

### DIFF
--- a/install_config/install/advanced_install.adoc
+++ b/install_config/install/advanced_install.adoc
@@ -801,8 +801,33 @@ openshift_examples_modify_imagestreams=true
 |Set to `true` if pointing to a registry other than the default. Modifies the image stream location to the value of `*oreg_url*`.
 
 |`*openshift_docker_additional_registries*`
-|Specify the additional registry or registries.
-If the registry required to access the registry is other than `80` include the port number required.
+|Specify the additional registry or registries in the form of
+`<address>:<port>`
+
+|`*openshift_cockpit_deployer_prefix*`
+|Specify the URL and path to namespace where the registry-console image is located. 
+ifdef::openshift-origin[]
+Note that the value defaults to `kubernetes`, so the value would be `cockpit/kubernetes:latest`. This differs from several other images that have the `origin-` as the name.
+endif::[]
+ifdef::openshift-enterprise[]
+Note that the value for this must end in `/openshift3` rather than `ose-`, which is the standard for other images. 
+endif::[]
+
+|`*openshift_web_console_prefix*`
+|Specify the prefix for the web console images.
+
+|`*openshift_service_catalog_image_prefix*`
+|Specify the prefix for the service catalog component image.
+
+|`*ansible_service_broker_image_prefix*`
+|Specify the prefix for the ansible service broker component image.
+
+|`*template_service_broker_prefix*`
+|Specify the prefix for the template service broker component image.
+
+|`*openshift_crio_systemcontainer_image_override*`
+|A setting for if you are using CRI-O and if you are using an alternative CRI-O
+system container image from another registry.
 |===
 
 For example:
@@ -810,6 +835,21 @@ For example:
 oreg_url=example.com/openshift3/ose-${component}:${version}
 openshift_examples_modify_imagestreams=true
 openshift_docker_additional_registries=example.com:443
+openshift_crio_systemcontainer_image_override=<registry>/<repo>/<image>:<tag>
+ifdef::openshift-origin[]
+openshift_cockpit_deployer_prefix='registry.example.com/cockpit/'
+openshift_web_console_prefix='registry.example.com/openshift/origin-
+openshift_service_catalog_image_prefix='registry.example.com/openshift/origin-'
+ansible_service_broker_image_prefix='registry.example.com/openshift/origin-'
+template_service_broker_prefix="registry.example.com/openshift/origin-"
+endif::[]
+ifdef::openshift-enterprise[]
+openshift_cockpit_deployer_prefix='registry.example.com/openshift3/'
+openshift_web_console_prefix='registry.example.com/openshift3/ose-
+openshift_service_catalog_image_prefix='registry.example.com/openshift3/ose-'
+ansible_service_broker_image_prefix='registry.example.com/openshift3/ose-'
+template_service_broker_prefix='registry.example.com/openshift3/ose-'
+endif::[]
 ----
 
 [[advanced-install-configuring-docker-route]]


install_config/install/advanced_install.adoc
```diff
@@ -776,15 +776,65 @@ openshift_examples_modify_imagestreams=true
 
 |`*openshift_docker_additional_registries*`
 |Specify the additional registry or registries.
+
+|`*openshift_cockpit_deployer_prefix*`
+|Specify the URL and path to the directory where the image is located.
+
```
jianlinliu: I might say:
Specify the URL and path to namespace where the registry-console image is located.

install_config/install/advanced_install.adoc
```diff
@@ -776,15 +776,65 @@ openshift_examples_modify_imagestreams=true
...
+|`*openshift_web_console_prefix*`
+|The prefix for the component images. For example, with
```
jianlinliu: I might say:
Specify the prefix for the web console images.

install_config/install/advanced_install.adoc
```diff
@@ -776,15 +776,65 @@ openshift_examples_modify_imagestreams=true
...
+ifdef::openshift-origin[]
+`openshift/origin-web-console:v3.9`, set prefix `openshift/origin-`.
```
jianlinliu: How about this:
`openshift/origin-web-console:v3.9`, set prefix `registry.example.com/openshift/origin-`

install_config/install/advanced_install.adoc
```diff
@@ -776,15 +776,65 @@ openshift_examples_modify_imagestreams=true
...
+endif::[]
+ifdef::openshift-enterprise[]
+`openshift3/ose-web-console:v3.9`, set prefix `openshift3/ose-`.
```
jianlinliu: How about this:
`openshift3/ose-web-console:v3.9`, set prefix `registry.example.com/openshift3/ose-`.

install_config/install/advanced_install.adoc
```diff
@@ -776,15 +776,65 @@ openshift_examples_modify_imagestreams=true
...
+endif::[]
+
+|`*openshift_template_service_broker_image_prefix*`
```
jianlinliu: Never use this parameter in openshift-ansible, should be removed.

install_config/install/advanced_install.adoc
```diff
@@ -776,15 +776,65 @@ openshift_examples_modify_imagestreams=true
...
+|
+
+|`*openshift_service_catalog_image_prefix*`
+|For use if you have the service catalog enabled.
```
jianlinliu: How about this:
Specify the prefix for the service catalog component image. For example:
example.com/openshift3/ose-

install_config/install/advanced_install.adoc
```diff
@@ -776,15 +776,65 @@ openshift_examples_modify_imagestreams=true
...
+
+|`*ansible_service_broker_image_prefix*`
```
jianlinliu: How about this:
Specify the prefix for the ansible service broker component image. For example:
example.com/openshift3/ose-

install_config/install/advanced_install.adoc
```diff
@@ -776,15 +776,65 @@ openshift_examples_modify_imagestreams=true
...
+|
+
+|`*template_service_broker_prefix*`
```
jianlinliu: How about this:
Specify the prefix for the template service broker component image. For example:
example.com/openshift3/ose-

install_config/install/advanced_install.adoc
```diff
@@ -776,15 +776,65 @@ openshift_examples_modify_imagestreams=true
...
+endif::[]
+ifdef::openshift-enterprise[]
+template_service_broker_prefix="registry.example.com/ose-"
+endif::[]
+openshift_crio_systemcontainer_image_override=<registry>/<repo>/<image>:<tag>
 ----
 
+[NOTE]
+====
+The `openshift_cockpit_deployer_prefix` and `openshift_web_console_prefix` variables cause a known issue where the
```
jianlinliu: I am not sure what is the known issue here.

install_config/install/advanced_install.adoc
```diff
@@ -776,15 +776,65 @@ openshift_examples_modify_imagestreams=true
...
 openshift_examples_modify_imagestreams=true
 openshift_docker_additional_registries=example.com
+openshift_cockpit_deployer_prefix='registry.example.com/openshift3/'
+ifdef::openshift-origin[]
+`openshift/origin-web-console:v3.9`, set prefix `openshift/origin-`.
+endif::[]
+ifdef::openshift-enterprise[]
+`openshift3/ose-web-console:v3.9`, set prefix `openshift3/ose-`.
+endif::[]
+openshift_template_service_broker_image_prefix
```
jianlinliu: Should be removed.

install_config/install/advanced_install.adoc
```diff
@@ -776,15 +776,65 @@ openshift_examples_modify_imagestreams=true
...
 oreg_url=example.com/openshift3/ose-${component}:${version}
 openshift_examples_modify_imagestreams=true
 openshift_docker_additional_registries=example.com
+openshift_cockpit_deployer_prefix='registry.example.com/openshift3/'
+ifdef::openshift-origin[]
+`openshift/origin-web-console:v3.9`, set prefix `openshift/origin-`.
+endif::[]
+ifdef::openshift-enterprise[]
+`openshift3/ose-web-console:v3.9`, set prefix `openshift3/ose-`.
+endif::[]
```
jianlinliu: should not be something like this?
openshift_web_console_prefix='registry.example.com/openshift3/ose-'

install_config/install/advanced_install.adoc
```diff
@@ -776,15 +776,65 @@ openshift_examples_modify_imagestreams=true
...
+openshift_template_service_broker_image_prefix
+openshift_service_catalog_image_prefix=openshift/origin-
```
jianlinliu: should be something like this:
openshift_service_catalog_image_prefix='registry.example.com/openshift3/ose-'

install_config/install/advanced_install.adoc
```diff
@@ -776,15 +776,65 @@ openshift_examples_modify_imagestreams=true
...
+ansible_service_broker_image_prefix=openshift/
```
jianlinliu: should be something like this:
ansible_service_broker_image_prefix='registry.example.com/openshift3/ose-'

install_config/install/advanced_install.adoc
```diff
@@ -776,15 +776,65 @@ openshift_examples_modify_imagestreams=true
...
+ifdef::openshift-origin[]
+template_service_broker_prefix="registry.example.com/origin-"
+endif::[]
+ifdef::openshift-enterprise[]
+template_service_broker_prefix="registry.example.com/ose-"
```
jianlinliu: it is better to be something like this:
template_service_broker_prefix='registry.example.com/openshift3/ose-'

bfallonf:
Thanks for the examples @jianlinliu ! That was very helpful. I left the note box about the known issues because the BZ associated with that is not yet completed. I will watch that BZ before merging this PR.

install_config/install/advanced_install.adoc
```diff
@@ -798,17 +798,56 @@ openshift_examples_modify_imagestreams=true
...
+openshift_web_console_prefix='registry.example.com/openshift3/ose-
+endif::[]
+openshift_service_catalog_image_prefix='registry.example.com/openshift3/ose-'
+ansible_service_broker_image_prefix='registry.example.com/openshift3/ose-'
+ifdef::openshift-origin[]
+template_service_broker_prefix="registry.example.com/origin-"
+endif::[]
+ifdef::openshift-enterprise[]
+template_service_broker_prefix='registry.example.com/openshift3/ose-'
+endif::[]
```
jianlinliu: If possible, here is only need keep one pair of origin or enterprise version judgment, and define most of prefix there, just like this:
ifdef::openshift-origin[]
openshift_web_console_prefix='registry.example.com/openshift/origin-
openshift_service_catalog_image_prefix='registry.example.com/openshift/origin-'
ansible_service_broker_image_prefix='registry.example.com/openshift/origin-'
template_service_broker_prefix="registry.example.com/openshift/origin-"
endif::[]
ifdef::openshift-enterprise[]
openshift_web_console_prefix='registry.example.com/openshift3/ose-
openshift_service_catalog_image_prefix='registry.example.com/openshift3/ose-'
ansible_service_broker_image_prefix='registry.example.com/openshift3/ose-'
template_service_broker_prefix='registry.example.com/openshift3/ose-'
endif::[]

bfallonf:
Thanks @jianlinliu That is indeed a better way to list them.

I'm not sure if this does need a "known issue" section, so I'll comment in the BZ.

install_config/install/advanced_install.adoc
sosiouxme: ref https://bugzilla.redhat.com/show_bug.cgi?id=1558309 for openshift_cockpit_deployer_prefix -- where the issue is that it doesn't have `ose-` or `origin-` like all the other image prefixes and so if you're not paying very close attention, you're likely to copy all the other ones and end up with an image that doesn't exist. It could be called out in a note like this or just a comment.

not sure what the web console prefix issue is though? that's unrelated to cockpit.

install_config/install/advanced_install.adoc
```diff
@@ -798,17 +798,54 @@ openshift_examples_modify_imagestreams=true
...
+|A setting for if you are using CRI-O and if you are using an alternative CRI-O
+system container image from another registry.
 |===
 
 For example:
 ----
 oreg_url=example.com/openshift3/ose-${component}:${version}
 openshift_examples_modify_imagestreams=true
 openshift_docker_additional_registries=example.com:443
+openshift_cockpit_deployer_prefix='registry.example.com/openshift3/'
```
sosiouxme: I do not know how best to present this image. The names are completely different for origin vs enterprise, and from other similar images.

[Origin](https://github.com/openshift/openshift-ansible/blob/18961e08a744981f0a7726f653edf0c981ff67e6/roles/openshift_hosted_templates/files/v3.9/origin/registry-console.yaml#L103): default is `cockpit/` and to use an alternate registry, should be something like `registry.example.com/cockpit/` unless of course they renamed it to something sane. Note the rest of the image name defaults to `kubernetes` so `cockpit/kubernetes:latest`.

[Enterprise](https://github.com/openshift/openshift-ansible/blob/18961e08a744981f0a7726f653edf0c981ff67e6/roles/openshift_hosted_templates/files/v3.9/enterprise/registry-console.yaml#L103): default is more like usual, so `registry.example.com/openshift3/` would work for alternate registry.

Note there are variables in the template for specifying the rest of the image, and ansible variables that feed into them. I don't know that you want to mention them here. If you get into listing every single variable that can impact image names this will be quite a list.

bfallonf: Hmm right. Thanks @sosiouxme . I've included what you've mentioned here in the table above when it describes the parameter.

install_config/install/advanced_install.adoc
bfallonf: Thanks @sosiouxme . I've updated this info in the table above instead of a note box here.

sferich888:
LGTM

bfallonf:
Thanks @sferich888 . I can see that the BZ about the installer has closed, so I'll assume that nothing needs to be in the known issues. If I'm wrong, please let me know.

In the meantime, @openshift/team-documentation PTAL

install_config/install/advanced_install.adoc
```diff
@@ -798,17 +798,57 @@ openshift_examples_modify_imagestreams=true
 |Set to `true` if pointing to a registry other than the default. Modifies the image stream location to the value of `*oreg_url*`.
 
 |`*openshift_docker_additional_registries*`
-|Specify the additional registry or registries.
-If the registry required to access the registry is other than `80` include the port number required.
+|Specify the additional registry or registries. If the registry required to access the registry is other than `80` include the port number required.
```
mburke5678: @sosiouxme This doesn't look right. It is a change I made in a previous PR [1], but I didn't see it then. Is this is not right, @bfallonf , can you fix? 
_If the registry required to access the registry"
Should this be _If the **port** required to access the registry..._
[1] https://github.com/openshift/openshift-docs/pull/8314/files

install_config/install/advanced_install.adoc
```diff
@@ -798,17 +798,57 @@ openshift_examples_modify_imagestreams=true
...
+
+|`*openshift_cockpit_deployer_prefix*`
+|Specify the URL and path to namespace where the registry-console image is located. 
+ifdef::openshift-origin[]
+Note that the value defaults to `kubernetes`, so the value would be `cockpit/kubernetes:latest`. This differes from several other images that have the `origin-` as the name.
```
mburke5678: s/differes/differs

mburke5678:
@bfallonf One question and one nit. Otherwise LGTM

install_config/install/advanced_install.adoc
bfallonf: @mburke5678 This could make sense. The only thing I am questioning is that it's about "Additional registries", but that doesn't mean it's correct. @sosiouxme thoughts?

sosiouxme: Yes, s/registry/port/ there.

However it may be moot for 3.9+ at this point as all the docker config has moved to the prerequisites.yaml playbook so I'm not sure this goes here any more (might be a nice patch for older versions though).

Also now I doubt the default port is 80. I think it's either 443 (HTTPS for secure communication) or 5000 (docker default). I can't see how it would hurt to be explicit, so I'd probably change the whole line to "Specify the additional registry or registries in the form of <address>:<port>"

bfallonf: @sosiouxme Sure. Changed to "Specify the additional registry or registries in the form of
<address>:<port>"

bfallonf:
This took a while for some reason, but I've updated as per @mburke5678 and @sosiouxme 's suggestions. I'll comment in the BZ to check if all is correct and where it needs to be here.

vikram-redhat:
@bfallonf - anything holding this up (besides the rebase)?

bfallonf:
@vikram-redhat I was waiting for the findings from another BZ. Turns out it was closed at notabug.

Problem is, the new install arrangements are causing issues. It'll be easier to close this and create a new one for 3.10 and another for 3.9 <-.

bfallonf:
Followups: 

#10944 
#10943

PR #8347, opened 2018-03-22
BZ#1501483 - marked memory-based autoscaling as tech preview

gaurav-nelson:
For https://bugzilla.redhat.com/show_bug.cgi?id=1501483

### DIFF
--- a/dev_guide/pod_autoscaling.adoc
+++ b/dev_guide/pod_autoscaling.adoc
@@ -161,6 +161,21 @@ spec:
 [[pod-autoscaling-memory]]
 == Autoscaling for Memory Utilization
 
+[IMPORTANT]
+====
+Autoscaling for Memory Utilization is a Technology Preview feature only.
+ifdef::openshift-enterprise[]
+Technology Preview features are not supported with Red Hat production service
+level agreements (SLAs), might not be functionally complete, and Red Hat does
+not recommend to use them for production. These features provide early access to
+upcoming product features, enabling customers to test functionality and provide
+feedback during the development process.
+
+For more information on Red Hat Technology Preview features support scope, see
+https://access.redhat.com/support/offerings/techpreview/.
+endif::[]
+====
+
 Unlike CPU-based autoscaling, memory-based autoscaling requires specifying the
 autoscaler using YAML instead of using the `oc autoscale` command. Optionally,
 you can specify the minimum number of pods and the average memory utilization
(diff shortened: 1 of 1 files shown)


dev_guide/pod_autoscaling.adoc
```diff
@@ -215,6 +230,11 @@ spec:
 $ oc create -f hpa.yaml
 ----
 
+[NOTE]
+====
+Verify that your application's memory usage does not go down, which is more complicated than it seems.
```
gaurav-nelson: @DirectXMan12 If we want to add a note for complexities we should include what is required? why it is required. 
```
Verify that your application's memory usage does not go down (go down from which value? the one we set in example?)
, which is more complicated than it seems. (why? what goes wrong if we don't do that)
```

gaurav-nelson: @DirectXMan12 PTAL

DirectXMan12: Probably something like

```
In order for memory-based autoscaling to work, memory usage must increase *and* decrease proportionally to replica count.  More specifically, an increase in replica count *must* lead to an overall decrease in memory (working set) usage per pod, on average.  Similarly, a decrease in replica count *must* lead to an overall increase in per-pod memory usage, on average.

It's important to ensure that your application meets these requirements before trying to use memory-base d autoscaling.  Use the OpenShift dashboard (and `oc top` in 3.10+) to verify the memory behavior of your application.

**Comment by gaurav-nelson** [comment]
@openshift/team-documentation PTAL 🆓

File: dev_guide/pod_autoscaling.adoc [hunk 2]
```diff
@@ -215,6 +230,21 @@ spec:
...
+====
+For memory-based autoscaling to work, memory usage must increase and decrease
+proportionally to the replica count. On average:
+
+* An increase in replica count must lead to an overall decrease in memory
+(working set) usage per-pod.
+* A decrease in replica count must lead to an overall increase in per-pod memory
+usage.
+
+Use the OpenShift dashboard to verify the memory behavior of your application
```
mburke5678: What is the OpenShift dashboard? New in 3.9 and I am not aware? Should this be Kibana dashboard?

mburke5678:
@gaurav-nelson A couple of nits.

dev_guide/pod_autoscaling.adoc
```diff
@@ -215,6 +230,21 @@ spec:
...
+and ensure that your application meets these requirements before using
+memory-based autoscaling.
+====
```
mburke5678: Do we need to tell users how to ensure the memory behavior?

dev_guide/pod_autoscaling.adoc
DirectXMan12: s/dashboard/webconsole/

dev_guide/pod_autoscaling.adoc
gaurav-nelson: Thanks @mburke5678, I am updating this to `.. use the web console to check the memory ...` I think this refers to manually increasing the replicas in web console and look at the memory graph.

mburke5678:
@gaurav-nelson LGTM

Issue #8355, opened 2018-03-22
Hyper link typo and wrong version tag in ocp_3_9_release_notes


warmchang:
### Which section(s) is the issue in?
1. Hyper link typo of "experimental-qos-reserved".
2. Wrong version tag of 3.9 release.

### What needs fixing?
1. The xref should be "xref:../admin_guide/overcommit.adoc#configuring-reserve-resources".
2. There is no v3.9.0-rc.0 tag, should be v3.9.0.

PR #8356, opened 2018-03-22
Fix hyper link typo and wrong version tag in ocp_3_9_release_notes

warmchang:
Fix: https://github.com/openshift/openshift-docs/issues/8355

1. The xref should be "xref:../admin_guide/overcommit.adoc#configuring-reserve-resources".
2. There is no v3.9.0-rc.0 tag, should be v3.9.0.


ahardin-rh:
Thank you!

PR #8350, opened 2018-03-22
bug 1558228 linking to firewalld config

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1558228

PR #8358, opened 2018-03-22
bug 1555344 adding SkyDNS context

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1555344

@bfallonf, does this sound ok to you? I didn't see a good topic to link to. [This known issue](https://docs.openshift.com/container-platform/3.9/install_config/storage_examples/gluster_backed_registry.html#backing-docker-registry-with-glusterfs-storage-pod-cannot-resolve) has the most info about changing `dnsmasq`, but I'm not convinced that it's close enough to this scenario.

### DIFF
--- a/install_config/install/prerequisites.adoc
+++ b/install_config/install/prerequisites.adoc
@@ -368,10 +368,11 @@ advanced installation of {product-title} (using Ansible), because this option
 overrides the default IP address set by xref:../../admin_solutions/master_node_config.adoc#node-config-options[`*dnsIP*`].
 
 Instead, allow the installer to configure each node to use *dnsmasq* and forward
-requests to SkyDNS or the external DNS provider. If you do set the
-`*openshift_dns_ip*` option, then it should be set either with a DNS IP that
-queries SkyDNS first, or to the SkyDNS service or endpoint IP (the Kubernetes
-service IP).
+requests to the external DNS provider or SkyDNS, the internal DNS service for
+cluster-wide DNS resolution of internal hostnames for services and pods. If you
+do set the `*openshift_dns_ip*` option, then it should be set either with a DNS
+IP that queries SkyDNS first, or to the SkyDNS service or endpoint IP (the
+Kubernetes service IP).
 ====
 
 To verify that hosts can be resolved by your DNS server:


install_config/install/prerequisites.adoc
```diff
@@ -366,9 +366,11 @@ advanced installation of {product-title} (using Ansible), because this option
 overrides the default IP address set by xref:../../admin_solutions/master_node_config.adoc#node-config-options[`*dnsIP*`].
 
 Instead, allow the installer to configure each node to use *dnsmasq* and forward
-requests to SkyDNS or the external DNS provider. If you do set the
-`*openshift_dns_ip*` option, then it should be set either with a DNS IP that
-queries SkyDNS first, or to the SkyDNS service or endpoint IP (the Kubernetes
+requests to SkyDNS, the internal DNS service that provides cluster-wide DNS
+resolution of internal hostnames for services and pods, or the external DNS
+provider. If you do set the
```
bfallonf: I'm worried the new information has made the "or the external DNS provider" bit a little vague on what it's modifying. Maybe brackets? So:

"Instead, allow the installer to configure each node to use *dnsmasq* and forward request to SkyDNS (the internal DNS service for cluster-wide DNS resolution of internal hostnames for services and pods) or the external DNS provider."

Just a suggestion. Breaks up the sentence at least?

bfallonf:
@kalexand-rh I made a suggestion about how the information is being broken up.

But also, has there been a git thing where the info from #8350 has been pulled in? That content is there too.

install_config/install/prerequisites.adoc
kalexand-rh: IBM style says to avoid parentheses, but that's a good point. How does swapping the order of the DNS types sound to you?

"Instead, allow the installer to configure each node to use dnsmasq and forward request to the external DNS provider or SkyDNS, the internal DNS service for cluster-wide DNS resolution of internal hostnames for services and pods."

kalexand-rh:
I must have started from the wrong branch. :(

install_config/install/prerequisites.adoc
bfallonf: @kalexand-rh Sounds good to me!

bfallonf:
@kalexand-rh Oohh that'd be it. And the other PR merging created the conflict. Let us know if you need help with fixing it all (I say, not knowing your level of Git knowledge).

kalexand-rh:
@bfallonf, thank you! I've used Git for SCM before, so I'm just force pushing over my "learning experiences," where I'm either not following process or learning that the way I used to do things doesn't work anymore. :)

PR #8363, opened 2018-03-22
Documented the route annotation limiting concurrent connections

knobunc:
This adds documentation for haproxy.router.openshift.io/pod-concurrent-connections

Added to OpenShift by https://github.com/openshift/origin/pull/15559

### DIFF
--- a/architecture/networking/routes.adoc
+++ b/architecture/networking/routes.adoc
@@ -604,6 +604,7 @@ For all the items outlined in this section, you can set annotations on the
 |`*haproxy.router.openshift.io/cookie_name*`| Specifies an optional cookie to be used for
 this route. The name must consist of any combination of upper and lower case letters, digits, "_",
 and "-". The default is the hashed internal key name for the route. |
+|`*haproxy.router.openshift.io/pod-concurrent-connections*`| Sets the maximum number of connections that are allowed to a backing pod from a router.  Note: if there are multiple pods, each can have this many connections.  But if you have multiple routers, there is no coordination among them, each may connect this many times. If not set, there is no limit. |
 |`*haproxy.router.openshift.io/rate-limit-connections*`| Setting `true` or `TRUE` to enables rate limiting functionality. |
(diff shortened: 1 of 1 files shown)


bfallonf:
@knobunc LGTM for now. Might require more efforts (ie, a paragraph explaining how to do it and what it means somewhere else) before it goes into the enterprise docs. Judging by the Trello cards, this is for 3.10, right? I can wait until then for any followups.

I'll let @pecameron give his thoughts if he'd like.

architecture/networking/routes.adoc
```diff
@@ -604,6 +604,7 @@ For all the items outlined in this section, you can set annotations on the
 |`*haproxy.router.openshift.io/cookie_name*`| Specifies an optional cookie to be used for
 this route. The name must consist of any combination of upper and lower case letters, digits, "_",
 and "-". The default is the hashed internal key name for the route. |
+|`*haproxy.router.openshift.io/pod-concurrent-connections*`| Sets the maximum number of connections that are allowed to a backing pod from a router.  Note: if there are multiple pods, each can have this many connections.  But if you have multiple routers, there is no coordination among them, each may connect this many times. If not set, there is no limit. |
```
lihongan: I just curious about what's the behaviour if set it to "0" ?  will drop any connections or allow all connections (no limit) ?
I've checked the haproxy doc: https://cbonte.github.io/haproxy-dconv/1.5/configuration.html#5.2-maxconn and it says: 
```
The default value is "0" which means unlimited.
```
so looks "not set" and "set to 0" are equivalent, we should add this to doc.

bfallonf: @knobunc ^ thoughts?

bfallonf:
@knobunc Sorry, one more question I forgot to ask. Does this apply for dedicated and online? It's going into the architecture guide, so I should have asked.

knobunc:
@bfallonf Yes, this applies to dedicated, online, and bare-metal.

bfallonf:
Thanks @knobunc . I'll merge.

PR #8368, opened 2018-03-22
Bug 1542724 Add additional router metrics env var to docs

bmcelvee:
https://bugzilla.redhat.com/show_bug.cgi?id=1542724

### DIFF
--- a/architecture/topics/haproxy_template_router.adoc
+++ b/architecture/topics/haproxy_template_router.adoc
@@ -41,21 +41,25 @@ to provide
 link:https://cbonte.github.io/haproxy-dconv/1.5/configuration.html#9[HAProxy CSV format] metrics, or
 provide no router metrics at all.
 
-The metrics are collected from both the router controller and from HAProxy every 5 seconds.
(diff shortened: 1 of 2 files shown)


vikram-redhat:
@bmcelvee - is this ready for QE/SME/Peer review?

bmcelvee:
@mburke5678 would you mind taking a look at this please? I can't seem to get the link to the Router environment variables working. 

Thanks! :icecream:

bfallonf:
@bmcelvee Do you mean from line 61 in the haproxy file? Because the topics directory is not matching up well with the assembly file? Because of the modularization, it should be able to link to the assembly, but with the full path (because that's where the anchor is). So:

See router xref:../../architecture/networking/routes.adoc#env-variables[route environment variables] section for more information.

Sorry if I'm wrong or if I'm telling you what you already know... We came across this when I first moved the architecture stuff around.

bmcelvee:
Thanks, @bfallonf! That worked. :smile_cat:

PR #8372, opened 2018-03-22
Bug 1511760 Add note to remove secret

bmcelvee:
https://bugzilla.redhat.com/show_bug.cgi?id=1511760

### DIFF
--- a/architecture/service_catalog/index.adoc
+++ b/architecture/service_catalog/index.adoc
@@ -94,6 +94,13 @@ communicate with cluster service brokers and wait for them to perform their
 deprovisioning work. In normal circumstances, this may take several minutes or
 longer depending on the service.
 
+[NOTE]
+====
+If you delete a service binding used by a deployment, you must also remove any
+references to the binding secret from the deployment. Otherwise, the next
+rollout will fail.
+====
+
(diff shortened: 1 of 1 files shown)


bmcelvee:
@spadgett Is this along the lines of what should be added to the docs based on https://bugzilla.redhat.com/show_bug.cgi?id=1511760? Thanks! :four_leaf_clover:

architecture/service_catalog/index.adoc
```diff
@@ -94,6 +94,13 @@ communicate with cluster service brokers and wait for them to perform their
 deprovisioning work. In normal circumstances, this may take several minutes or
 longer depending on the service.
 
+[NOTE]
+====
+If you remove a secret from your deployment environment, you must also remove
+the secret from the deployment environment tab. Otherwise, your rollout will
+fail.
```
spadgett: How about...

"If you delete a service binding used by a deployment, you must also remove any references to the binding secret from the deployment. Otherwise, the next rollout will fail."

bmcelvee: That is much better, thank you. :) I'll update the PR.

spadgett (approved):

PR #8403, opened 2018-03-23
bug 1559296 fixing blue green command errors

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1559296

It looks like this file changed folders in 3.6.

### DIFF
--- a/upgrading/blue_green_deployments.adoc
+++ b/upgrading/blue_green_deployments.adoc
@@ -99,12 +99,11 @@ You must ensure that your current node hosts in production are labeled either
 $ oc get nodes
 ----
 
-. Ensure that all hosts have appropriate node labels. Consider that by default in
-{product-title} 3.9, all master hosts should now also be configured as
-schedulable node hosts (so that they are joined to the
-xref:../architecture/networking/network_plugins.adoc#architecture-additional-concepts-network-plugins[pod network] and can run the web console pod). Additional labels will assist in
-management of clusters, specifically labeling hosts as their types (e.g.,
-`type=master` or `type=node`) should assist management.
+. Ensure that all hosts have appropriate node labels. All master hosts should
+be configured as schedulable node hosts (so that they are joined to the
+xref:../architecture/networking/network_plugins.adoc#architecture-additional-concepts-network-plugins[pod network]
+and can run the web console pod). To improve cluster management, add a label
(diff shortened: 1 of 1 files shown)


upgrading/blue_green_deployments.adoc
```diff
@@ -342,19 +342,21 @@ $ oc delete node --selector=color=blue
 
 .. *Option B* Filter out the masters before running the `delete` command:
 
-... Verify the list of blue node hosts are as expected by running:
+... Verify the list of blue node hosts to delete by running the following
+command. The output of this command includes a list of all node hosts that have 
+the `color=blue` label but do not have the `type=master` label. You can change
```
soltysh: Since the examples from BZ shows that the labels need to exist, I'd propose to change the wording here so that it shows that the value of `type` is different than master, but the label itself must exists, otherwise the command will fail.

kalexand-rh: @soltysh, thanks! I think the section around line 102 might need to be updated a bit, too.

kalexand-rh: @soltysh, I've updated both this section and the section around 102.

kalexand-rh:
(@adellape, `blame` says that I'm changing sections you modified most recently, if you're interested.)

mburke5678:
@kalexand-rh LGTM

vikram-redhat:
@mburke5678 if done with peer review, you should remove the peer review label and put the peer review done label.

upgrading/blue_green_deployments.adoc
```diff
@@ -99,12 +99,12 @@ You must ensure that your current node hosts in production are labeled either
 $ oc get nodes
 ----
 
-. Ensure that all hosts have appropriate node labels. Consider that by default in
+. Ensure that all hosts have appropriate node labels. By default in
 {product-title} 3.9, all master hosts should now also be configured as
```
vikram-redhat: @kalexand-rh be careful when you merge this. It seems to carry over content that is for 3.9 into older branches. 

Should also make sure that this is changed to {product-version}

kalexand-rh:
other PRs:
3.7: https://github.com/openshift/openshift-docs/pull/8497
3.6: #8498 
3.5: #8499

PR #8450, opened 2018-03-27
Removed the Device Plug-in topic from 3.9

ahardin-rh:
Removing this topic from the 3.9 docs per @jeremyeder 

cc @aheslin @vikram-redhat

### DIFF
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -195,7 +195,7 @@ Topics:
   - Name: Admission Controllers
     File: admission_controllers
   - Name: Custom Admission Controllers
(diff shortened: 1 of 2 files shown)


vikram-redhat:
@ahardin-rh @jeremyeder - do we know when this is going to be made live? Are we looking at 3.10 now?

jeremyeder:
@vikram-redhat I've been told that we should not insert vendor-specific stuff into the official product docs.  The trouble is that documenting the device-plugin doesn't make much sense without an example.  We have a stub example in the upstream e2e, and that might be the only thing we can document.

PR #8453, opened 2018-03-27
BZ 1546911 Remove info on requiring manual migration based on your use case

ghost:
https://bugzilla.redhat.com/show_bug.cgi?id=1546911

This change only applies to 3.6.


tmorriso-rh:
@openshift/team-documentation PTAL. Thanks.

mburke5678:
@tmorriso-rh LGTM.

PR #8269, opened 2018-03-20
Updated screenshots for OCP 3.9

ahardin-rh:
https://trello.com/c/4USYdvgB/804-update-screenshots-for-39

PR #8471, opened 2018-03-27
Added missing statefulsets image

ahardin-rh:
Adds image missing from https://github.com/openshift/openshift-docs/pull/8269


ahardin-rh:
@vikram-redhat FYI

PR #8504, opened 2018-03-28
Clarified how haproxy counts the connections

knobunc:
Since each connection to the router also uses an internal connection
from the frontend to the backend, each connection that the user thinks
about really counts for two.  So the user should set the maximum
connections to double the number of external connections that they
want to plan for.

This applies to all versions where ROUTER_MAX_CONNECTIONS has been documented.

@openshift/networking PTAL

### DIFF
--- a/install_config/router/default_haproxy_router.adoc
+++ b/install_config/router/default_haproxy_router.adoc
@@ -295,7 +295,10 @@ $ oc adm router --max-connections=10000   ....
 Edit the xref:../../architecture/networking/routes.adoc#env-variables[`ROUTER_MAX_CONNECTIONS`] environment variable in the router's
 deployment configuration to change the value. The router pods are restarted with
 the new value. If `ROUTER_MAX_CONNECTIONS` is not present, the default value of
-20000, is used.
+20000, is used.    Be aware that each actual connection counts as two connections
+because of the way that HAProxy works.  There is a frontend and internally it
+dispatches to a backend.  So you should set this to double the number of connections
+that you want to accept externally.
 
 [[bind-strict-sni]]
 == HAProxy Strict SNI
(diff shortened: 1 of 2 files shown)


pravisankar (approved):
LGTM

Do we also want to document this caveat in https://github.com/openshift/openshift-docs/blob/master/install_config/router/customized_haproxy_router.adoc ?

vikram-redhat:
@bfallonf - PTAL

rajatchopra:
I didn't know about this doubling fact. It may be more complicated then. Because we send certain connections back to haproxy again on another port 10443 (for re-encrypt as an example). Would that make it a factor of four for certain connections?

 Looks good to me otherwise.

ramr (approved):
LGTM

ramr:
@rajatchopra  should still be double right?  The  connections to the backend are controlled via a `pod-concurrent-connections` annotation. The `ROUTER_MAX_CONNECTIONS` is used for the defaults/global `maxconn` directive and applies to in-bound connections to haproxy.
     *  the client → haproxy frontend `public_ssl`   
     *  `be[_no]_sni` → `fe[_no]_sni`

bfallonf:
@knobunc I think this would be better as a note box. So:
`
[NOTE]
====
A connection includes the frontend and internal backend. This counts as two connections. Be sure to set `ROUTER_MAX_CONNECTIONS` to double than the number of connections you intend to create. 
====
`

Also, should this be included in the router environment variables in the architecture guide for the router_max_connection variable?

https://docs.openshift.com/container-platform/3.9/architecture/networking/routes.html#env-variables

Let me know if it would be easier to fix these with a followup rather than you changing this PR.

bfallonf:
Above edits made in a followup PR #9190

PR #8553, opened 2018-04-03
Static egress IPs are no longer tech preview

danwinship:
The "static per-project egress IPs" feature is no longer tech preview as of 3.9. (This is already mentioned in the release notes.)

(This is against master but it will need to be backported to 3.9 too.)


bfallonf:
Thanks @danwinship Makes sense to me. I'll merge and cherrypick to 3.9.

PR #8559, opened 2018-04-03
Promote CPU manager pinning to supported

mburke5678:
https://trello.com/c/CGNHORQh/831-document-promote-cpu-manager-pinning-to-supported

### DIFF
--- a/scaling_performance/using_cpu_manager.adoc
+++ b/scaling_performance/using_cpu_manager.adoc
@@ -14,21 +14,6 @@ toc::[]
 [[what-cpu-manager-does]]
 == What CPU Manager Does
 
-[IMPORTANT]
-====
-CPU Manager is a Technology Preview feature.
-ifdef::openshift-enterprise[]
-Technology Preview features are not supported with Red Hat production service
-level agreements (SLAs), might not be functionally complete, and Red Hat does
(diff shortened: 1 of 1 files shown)


mburke5678:
@sjenning  @mdshuai 
Does this Trello elevate the entire CPU Manager feature to supported? The term "pinning" in the title makes this not clear to me.

sjenning:
fyi @derekwaynecarr 

The feature really is "Static CPU Policy".  While the end result in CPU pinning, it also repels other pods from the "exclusive" CPUs.  We should probably shouldn't using the word "pinning" and should instead refer to "exclusive cores" vs "shared cores".

mburke5678:
@sjenning @derekwaynecarr The topic doesn't use the term _pinning_. I just want to make sure the entire CPU Manager tool is being promoted. We can certainly work in _exclusive_ vs. _shared_ cores in a follow-up.

sjenning:
sorry, off-topic :) LGTM

mburke5678:
@mdshuai PTAL

mburke5678:
@mdshuai PTAL

mburke5678:
@mdshuai PTAL

mdshuai:
I have missed it, LGTM! Thanks

PR #8574, opened 2018-04-03
Added additional steps

ahardin-rh:
Follow-up to https://bugzilla.redhat.com/show_bug.cgi?id=1543003

### DIFF
--- a/upgrading/automated_upgrades.adoc
+++ b/upgrading/automated_upgrades.adoc
@@ -689,12 +689,15 @@ default value is `storagenode=glusterfs`.
 .. Remove the daemonset label from the node:
 +
 ----
-$ oc label node <node_name> storagenode-
+$ oc label node <node_name> <daemonset_label>-
 ----
 +
 This will cause the GlusterFS pod to terminate on that node.
 
-.. Run the upgrade playbook on the single node where you terminated GlusterFS.
+.. Add an additional label (for example, `type=upgrade`) to the node you want to upgrade.
+
+.. To run the upgrade playbook on the single node where you terminated GlusterFS,
+use `-e openshift_upgrade_nodes_label="type=upgrade"`.
 
 .. When the upgrade completes, relabel the node with the daemonset selector:
 +
@@ -714,6 +717,9 @@ $ for vol in `gluster volume list`; do gluster volume heal $vol info; done
 Ensure all of the volumes are healed and there are no outstanding tasks. The
 `heal info` command lists all pending entries for a given volume's heal process.
 A volume is considered healed when `Number of entries` for that volume is `0`.
+
+.. Remove the upgrade label (for example, `type=upgrade`) and go to the next
+GlusterFS node.
 
 [[special-considerations-for-gcepd]]
 == Special considerations when using gcePD


upgrading/automated_upgrades.adoc
```diff
@@ -702,6 +702,10 @@ This will cause the GlusterFS pod to terminate on that node.
 $ oc label node <node_name> <daemonset_label>
 ----
 
+.. Add an additional label (for example, `type=upgrade`) to the node you want to upgrade.
+
+.. Run the upgrade playbook with `-e openshift_upgrade_nodes_label="type=upgrade"`
+
```
dmoessne: I think we need to change the order here a bit:
- first remove daemonset label  
- then add label to node, e.g. type=upgrade
- after that run the upgrade playbook with  -e openshift_upgrade_nodes_label="type=upgrade"
otherwise we would first upgrade and then add the label which would not do wat we want: updating one, and only one node with that label

So the following would make sense to me

``
    Run oc get daemonset to verify the label found under NODE-SELECTOR. The default value is storagenode=glusterfs.  _(I think that needs to be run in the project where gluster pods are running)_

    Remove the daemonset label from the node:

    $ oc label node <node_name> storagenode-

    This will cause the GlusterFS pod to terminate on that node.

    Add an additional label (for example, type=upgrade) to the node you want to upgrade.

    To run the upgrade playbook on the single node where you terminated GlusterFS use -e openshift_upgrade_nodes_label="type=upgrade"

    When the upgrade completes, relabel the node with the daemonset selector:

    $ oc label node <node_name> <daemonset_label>

    Wait for the GlusterFS pod to respawn and appear.

    oc rsh into the pod and verify all volumes are healed:

    $ oc rsh <GlusterFS_pod_name>
    $ for vol in `gluster volume list`; do gluster volume heal $vol info; done

    Ensure all of the volumes are healed and there are no outstanding tasks. The heal info command lists all pending entries for a given volume’s heal process. A volume is considered healed when Number of entries for that volume is 0.

    Remove the node label (for example, type=upgrade) and go to the next GlusterFS node.

``

fcami (commented):
+1 from me

vikram-redhat:
@jarrpa - would you like to review this as well?

upgrading/automated_upgrades.adoc
jarrpa: Agreed. I would also change `$ oc label node <node_name> storagenode-` to `$ oc label node <node_name> <daemonset_label>-` since the "storagenode" part can be overridden.

jarrpa (changes requested):

ahardin-rh:
@dmoessne @jarrpa @fcami Thank you so much for your help. This is now updated.

upgrading/automated_upgrades.adoc
```diff
@@ -715,6 +718,8 @@ Ensure all of the volumes are healed and there are no outstanding tasks. The
 `heal info` command lists all pending entries for a given volume's heal process.
 A volume is considered healed when `Number of entries` for that volume is `0`.
 
+.. Remove the label (for example, `type=upgrade`) and go to the next GlusterFS node.
```
jarrpa: Change the beginning to `Remove the upgrade label`.

ahardin-rh:
QE verified these changes.

PR #8625, opened 2018-04-06
BZ 1520558 Add note that NFS is not supported for registry storage

ghost:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1520558

### DIFF
--- a/install_config/install/advanced_install.adoc
+++ b/install_config/install/advanced_install.adoc
@@ -939,6 +939,11 @@ installer:
 [[advanced-install-registry-storage-nfs-host-group]]
 ===== Option A: NFS Host Group
 
+[NOTE]
+====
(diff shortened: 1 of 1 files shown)


tmorriso-rh:

mburke5678:
@tmorriso-rh Should the _experimental_ notes go directly under the **Option A: NFS Host Group** and **Option B: External NFS Host** headers like the other _experimental_ notes?
Otherwise LGTM

vikram-redhat:
@mburke5678 that is a good point. We should keep it consistent.

tmorriso-rh:
@mburke5678 Changes made. PTAL. Thanks.

mburke5678:
@tmorriso-rh LGTM!

PR #8628, opened 2018-04-06
Added information about mixed environment upgrades

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1513693

@sdodson This bug was filed against 3.6 docs, so things may have changed in the meantime. Can you please confirm that this is still relevant and, if so, if anything else should be added? I can open a separate PR for 3.6 docs if you agree that these changes should be applied there. Thanks!


sdodson (approved):
/lgtm

PR #8630, opened 2018-04-06
Added information about the corsAllowedOrigins parameter behavior

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1482903

@liggitt PTAL. I notice we don't mention this parameter on this page: https://docs.openshift.com/container-platform/3.9/install_config/master_node_configuration.html, but it probably should be added. 
What is a good description for it, and where on the page should it be placed? I assume it should be part of https://docs.openshift.com/container-platform/3.9/install_config/master_node_configuration.html#master-config-controller-config. Is that correct? I will be doing a larger rework of that page soon, based on https://bugzilla.redhat.com/show_bug.cgi?id=1557178

Thanks!

### DIFF
--- a/architecture/infrastructure_components/web_console.adoc
+++ b/architecture/infrastructure_components/web_console.adoc
@@ -50,6 +50,22 @@ whitelist that host name by specifying the `--cors-allowed-origins` option
 on `openshift start` or from the related
 xref:../../install_config/master_node_configuration.adoc#master-configuration-files[master
 configuration file parameter `corsAllowedOrigins`].
+
+The `corsAllowedOrigins` parameter is controlled by the configuration field. No
+pinning or escaping is done to the value. The following is an example of how you
+can pin a host name and escape dots:
+
+----
+corsAllowedOrigins:
+- (?i)//my\.subdomain\.domain\.com(:|\z)
+----
+
+* The `(?i)` makes it case-insensitive.
+* The `//` pins to the beginning of the domain (and matches the double slash
+following `http:` or `https:`).
+* The `\.` escapes dots in the domain name.
+* The `(:|\z)` matches the end of the domain name `(\z)` or a port separator `(:)`.
+
 endif::openshift-enterprise,openshift-origin[]
 
 ifdef::openshift-origin,openshift-online,openshift-enterprise,openshift-dedicated[]


architecture/infrastructure_components/web_console.adoc
```diff
@@ -50,6 +50,12 @@ whitelist that host name by specifying the `--cors-allowed-origins` option
 on `openshift start` or from the related
 xref:../../install_config/master_node_configuration.adoc#master-configuration-files[master
 configuration file parameter `corsAllowedOrigins`].
+
+[NOTE]
+====
+The `corsAllowedOrigins` parameter is fully under the control of the
+configuration field. No pinning or escaping is done to the value.
+====
```
liggitt: it would probably be good to give an example of how you to pin a hostname and escape dots:

```yaml
corsAllowedOrigins:
- (?i)//my\.subdomain\.domain\.com(:|\z)
```

Explanations:
* `(?i)` makes it case-insensitive
* `//` pins to the beginning of the domain (matches the double slash following http: or https:)
* `\.` escape dots in the domain name
* `(:|\z)` matches the end of the domain name (`\z`) or a port separator (`:`)

ahardin-rh:
@liggitt Thanks! This is now updated. Any thoughts on a description of `corsAllowedOrigins` that I can add to the  Master and Node Configuration topic?

ahardin-rh:
@openshift/team-documentation Please peer review

liggitt:
not really... it's most immediately relevant to the webconsole, but I don't have a good general section it would go under

architecture/infrastructure_components/web_console.adoc
```diff
@@ -50,6 +50,22 @@ whitelist that host name by specifying the `--cors-allowed-origins` option
 on `openshift start` or from the related
 xref:../../install_config/master_node_configuration.adoc#master-configuration-files[master
 configuration file parameter `corsAllowedOrigins`].
+
+The `corsAllowedOrigins` parameter is fully under the control of the
+configuration field. No pinning or escaping is done to the value. Here is an
+example of how you can pin a host name and escape dots:
```
bfallonf: Because "fully" is an abverb, can I sugget: "The `corsAllowedOrigins` parameter is controlled by the configuration field." Though I'm not too sure which field it's talking about. the example below? Also, my suggestion is passive, but I think it's acceptable because we're talking about the parameter, and that should remain being the object.

Also, instead of "Here is..." can I suggest simply: "To pin a host name and escape dots" (Though I'm not too sure if I'm suggesting the right words). Up to you!

ahardin-rh:
QE verified

PR #8652, opened 2018-04-09
update supported versions of wildfly

bparees:


bparees:

v3.10+ only.

bmcelvee:
Thanks, @bparees! LGTM :four_leaf_clover:

PR #8575, opened 2018-04-03
Add --virtual-ip-groups option

imcsk8:
Added grouping option to VRRP configuration.

PR #8659, opened 2018-04-10
Added vrrp to high avail. variables

bfallonf:
Copied over from #8575 so that I can fix some caps and markup.

cc: @imcsk8 @knobunc @pravisankar Please let me know if you have any issues.

### DIFF
--- a/admin_guide/high_availability.adoc
+++ b/admin_guide/high_availability.adoc
@@ -227,7 +227,7 @@ each of the nodes matching the constraints or the label used. This pod runs
 link:http://www.keepalived.org/[*Keepalived*] which uses VRRP (Virtual Router
 Redundancy Protocol) among all the *Keepalived* daemons to ensure that the
 service on the watched port is available, and if it is not, *Keepalived* will
-automatically float the VIPs.
+automatically float the virtual IPs (VIPs).
 
 For production use, make sure to use a `--selector=<label>` with at least two nodes to select
 the nodes. Also, set a `--replicas=<n>` value that matches the
@@ -280,7 +280,7 @@ The list of command options described here are a subset that are relevant to thi
 
 [[virtual-ips]]
 === Virtual IP Addresses
-*Keepalived* manages a set of virtual IP addresses. The administrator must make sure that all these addresses:
+*Keepalived* manages a set of virtual IP addresses (VIPs). The administrator must make sure that all these addresses:
 
 * Are accessible on the configured hosts from outside the cluster.
 * Are not used for any other purpose within the cluster.
@@ -309,7 +309,7 @@ When a check script is not provided, a simple default script is run that tests
 the xref:../install_config/install/prerequisites.adoc#required-ports[TCP
 connection]. This default test is suppressed when the monitor port is `0`.
 
-For each VIP, *keepalived* keeps the state of the node. The VIP on the node may
+For each virtual IPs (VIPs), *keepalived* keeps the state of the node. The VIP on the node may
 be in *MASTER*, *BACKUP*, or *FAULT* state. All VIPs on the node that are not in
 the *FAULT* state participate in the negotiation to decide which will be
 *MASTER* for the VIP. All of the losers enter the *BACKUP* state. When the
@@ -542,6 +542,14 @@ See xref:../admin_guide/high_availability.adoc#ha-vrrp-id-offset[this discussion
 |0
 |See xref:../admin_guide/high_availability.adoc#ha-vrrp-id-offset[VRRP ID Offset] discussion for more details.
 
+|`--virtual-ip-groups`
+|`OPENSHIFT_HA_VIP_GROUPS`
+|
+|The number of groups to create for VRRP. If not set, a group is created for each
+virtual IP range specified with the `--virtual-ips` option. See
+xref:admin-guide-high-availability-configuring-more-than-254[Configuring
+IP failover for more than 254 addresses] for more information.
+
 |`--iptables-chain`
 |`OPENSHIFT_HA_IPTABLES_CHAIN`
 |INPUT
@@ -571,7 +579,7 @@ See xref:../admin_guide/high_availability.adoc#ha-vrrp-id-offset[this discussion
 [[ha-vrrp-id-offset]]
 === VRRP ID Offset
 
-Each ipfailover pod managed by the ipfailover deployment configuration (1 pod per node/replica) runs a *keepalived* daemon. As more ipfailover deployment configurations are configured, more pods are created and more daemons join into the common VRRP negotiation. This negotiation is done by all the *keepalived* daemons and it determines which nodes will service which VIPs.
+Each ipfailover pod managed by the ipfailover deployment configuration (1 pod per node/replica) runs a *keepalived* daemon. As more ipfailover deployment configurations are configured, more pods are created and more daemons join into the common VRRP negotiation. This negotiation is done by all the *keepalived* daemons and it determines which nodes will service which virtual IPs (VIPs).
 
 Internally, *keepalived* assigns a unique vrrp-id to each VIP. The negotiation uses this set of vrrp-ids, when a decision is made, the VIP corresponding to the winning vrrp-id is serviced on the winning node.
 
@@ -579,6 +587,55 @@ Therefore, for every VIP defined in the ipfailover deployment configuration, the
 the list of VIPs.  The vrrp-ids may have values in the range 1..255.
 
 When there are multiple ipfailover deployment configuration care must be taken to specify `--vrrp-id-offset` so that there is room to increase the number of VIPS in the deployment configuration and none of the vrrp-id ranges overlap.
+
+[[admin-guide-high-availability-configuring-more-than-254]]
+=== Configuring IP failover for more than 254 addresses
+
+IP failover management is limited to 254 groups of VIP addresses. By default
+[product-title] assigns one IP address to each group. You can use the
+`virtual-ip-groups` option to change this so multiple IP addresses are in each
+group and define the number of VIP groups available for each VRRP instance when
+xref:configuring-ip-failover[configuring IP failover]. 
+
+Grouping VIPs creates a wider range of allocation of VIPs per VRRP in the case
+of VRRP failover events, and is useful when all hosts in the cluster have access
+to a service locally. For example, when a service is being exposed with an
+`ExternalIP`. 
+
+[NOTE]
+====
+As a rule for failover, do not limit services, such as the router, to
+one specific host. Instead, services should be replicated to each host so that
+in the case of IP failover, the services do not have to be recreated on the new
+host.
+====
+
+[NOTE]
+====
+If you are using {product-title} health checks, the nature of IP failover and
+groups means that all instances in the group will not be checked. For that
+reason,
+link:https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/[the
+Kubernetes health checks] must be used to ensure that services are live.
+====
+
+----
+# oc adm ipfailover <ipfailover_name> --create \
+    ...
+    --virtual-ip-groups=<number_of_ipfailover_groups>
+----
+
+For example, if `--virtual-ip-groups` is set to `3` in an environment with seven
+VIPs, it creates three groups, assigning three VIPs to the first group, and two
+VIPs to the two remaining groups.
+
+[NOTE]
+====
+If the number of groups set by the `--virtual-ip-groups` option is fewer than
+the number of IP addresses set to fail over, the group will contain more than
+one IP address, and all of the addresses will move as a single unit.
+====
+
 
 [[configuring-a-highly-available-service]]
 === Configuring a Highly-available Service


admin_guide/high_availability.adoc
```diff
@@ -542,6 +542,11 @@ See xref:../admin_guide/high_availability.adoc#ha-vrrp-id-offset[this discussion
 |0
 |See xref:../admin_guide/high_availability.adoc#ha-vrrp-id-offset[VRRP ID Offset] discussion for more details.
 
+|`--virtual-ip-groups`
+|`OPENSHIFT_HA_VIP_GROUPS`
+|
+|Number of groups to create for VRRP. If not set, a group is created for each virtual IP given with `--virtual-ips`.
```
knobunc: There is a limit of 254 VRRP IDs in an Ethernet broadcast domain.  By setting this you can control how many VRRP IDs get used for failover.  But if the number of groups is fewer than the number of IP addresses to fail over, a group will contain more than one IP address, and all of those addresses will move together.  

This is useful for cases where all of the machines in the cluster have access to the service locally (i.e. a service is being exposed with an external IP).  BUT it should not be used for services that are local to a particular machine, such as a router (unless that service is actually running on all of the nodes).

knobunc (changes requested):

bfallonf:
@knobunc @imcsk8 I've added a new section that goes into more detail. I do have a few questions, and because there's no more comments in the trello card, I'm wodnering if this has been through QE? The trello card doesn't say much:

https://trello.com/c/fmHCGfhH/401-3-work-out-how-to-do-ipfailover-for-more-than-254-addresses-ingressoperationsscale

cc @bmeng

admin_guide/high_availability.adoc
```diff
@@ -580,6 +585,34 @@ the list of VIPs.  The vrrp-ids may have values in the range 1..255.
...
+# oc adm ipfailover <ipfailover_name> --create \
+    ...
+    --virtual-ip-groups=<number_of_ipfailover_groups>
+----
+
+[NOTE]
+====
+If the number of groups set by the `--virtual-ip-groups` option is fewer than
+the number of IP addresses set to fail over, the group will contain more than
+one IP address, and all of the addresses will move together.
```
bfallonf: Is "move together" the best way to describe the negative effects? Is there a better or more descriptive way to say it?

admin_guide/high_availability.adoc
knobunc: Also you probably should disable health checks on the ipfailover since only one of the items in the group will be checked.  So if you are relying on that for reliability, it will not work and you should not use this capability.  It is best to disable health checks when working with groups.

Basically we would suggest that people only use this with OpenShift services (ExternalIP or NodePort) and disable health checks.

If they are using pods with HostNetworking (like the router), then they need to be cautious when using this and make sure they understand all of the limitations.

admin_guide/high_availability.adoc
knobunc: Move as a single unit?

bfallonf:
Thanks for the extra comments @knobunc . If you or @imcsk8 have any more thoughts please let me know. I've made some changes to the suggestions.

admin_guide/high_availability.adoc
```diff
@@ -580,6 +588,51 @@ the list of VIPs.  The vrrp-ids may have values in the range 1..255.
...
+====
+
+----
+# oc adm ipfailover <ipfailover_name> --create \
+    ...
+    --virtual-ip-groups=<number_of_ipfailover_groups>
+----
+
+For example, if the above is used in a environment with seven VIPs, it creates
+three groups, assigning three VIPs to the first group, and two VIPs to the two
```
pravisankar: Don't you want to hard-code '--virtual-ip-groups=3' on line 620 ? Otherwise this example does not explain why it creates three groups.

bfallonf:
@pravisankar Thanks for that! I missed that it wasn't put together too well. I've changed it to be more specific. If you or @knobunc or @imcsk8 have more thoughts, please let me know.

In the meantime, @openshift/team-documentation PTAL.

admin_guide/high_availability.adoc
```diff
@@ -542,6 +542,14 @@ See xref:../admin_guide/high_availability.adoc#ha-vrrp-id-offset[this discussion
 |0
 |See xref:../admin_guide/high_availability.adoc#ha-vrrp-id-offset[VRRP ID Offset] discussion for more details.
 
+|`--virtual-ip-groups`
+|`OPENSHIFT_HA_VIP_GROUPS`
+|
+|Number of groups to create for VRRP. If not set, a group is created for each
+virtual IP given with `--virtual-ips`. See
+xref:../admin-guide-high-availability-configuring-more-than-254[Configuring
```
ahardin-rh: This path looks incorrect

admin_guide/high_availability.adoc
```diff
@@ -580,6 +588,51 @@ the list of VIPs.  The vrrp-ids may have values in the range 1..255.
 
 When there are multiple ipfailover deployment configuration care must be taken to specify `--vrrp-id-offset` so that there is room to increase the number of VIPS in the deployment configuration and none of the vrrp-id ranges overlap.
 
+[[admin-guide-high-availability-configuring-more-than-254]]
+=== Configuring IP failover for more than 254 addresses
+
+By default, IP failover has a limitation of managing only 254 VIP addresses. You
+can group sets of VIPs for VRRP instances using the `virtual-ip-groups` option
+to define the number of VIP groups available for each VRRP instance when
+cxref:../admin_guide/high_availability.adoc#configuring-ip-failover[configuring
```
ahardin-rh: s/cxref/xref 

The path should be xref:configuring-ip-failover 
(It points to another section in this same file.)

admin_guide/high_availability.adoc
```diff
@@ -580,6 +588,51 @@ the list of VIPs.  The vrrp-ids may have values in the range 1..255.
...
+[NOTE]
+====
+As a rule for failover, services, such as the router, should not be limited to
+one specific host. Instead, services should be replicated to each host so that
+in the case of IP failover, the services do not have to be recreated on the new
+host.
+
+If you are using {product-title} health checks, the nature of IP failover and
+groups means that all instances in the group will not be checked. For that
+reason, the Kubernetes health checks will be more effective.
```
ahardin-rh: s/will be/are

admin_guide/high_availability.adoc
```diff
@@ -580,6 +588,51 @@ the list of VIPs.  The vrrp-ids may have values in the range 1..255.
...
+IP. 
+
+[NOTE]
+====
+As a rule for failover, services, such as the router, should not be limited to
+one specific host. Instead, services should be replicated to each host so that
+in the case of IP failover, the services do not have to be recreated on the new
+host.
+
+If you are using {product-title} health checks, the nature of IP failover and
```
kalexand-rh: Is this really the same note? Should this be a separate limitation?
"If you use {product-title} health checks, do not configure IP failover for groups. Not all instances in the group can be checked for <more detail about what's checked or failing> during failover. You can link:<link to configuring K8 health checks>[configure Kubernetes health checks] for these groups instead."

admin_guide/high_availability.adoc
```diff
@@ -580,6 +588,51 @@ the list of VIPs.  The vrrp-ids may have values in the range 1..255.
...
+====
+
+----
+# oc adm ipfailover <ipfailover_name> --create \
+    ...
+    --virtual-ip-groups=<number_of_ipfailover_groups>
+----
+
+For example, if `--virtual-ip-groups` is set to `3` in an environment with seven
+VIPs, it creates three groups, assigning three VIPs to the first group, and two
```
kalexand-rh: I'd rewrite this as two sentences or s/first group,/first group

admin_guide/high_availability.adoc
```diff
@@ -580,6 +588,51 @@ the list of VIPs.  The vrrp-ids may have values in the range 1..255.
...
+VIPs to the two remaining groups.
+
+[NOTE]
```
kalexand-rh: I don't think this is a note. I think this should be a paragraph before the previous paragraph.

admin_guide/high_availability.adoc
```diff
@@ -580,6 +588,51 @@ the list of VIPs.  The vrrp-ids may have values in the range 1..255.
...
+====
+If the number of groups set by the `--virtual-ip-groups` option is fewer than
+the number of IP addresses set to fail over, the group will contain more than
+one IP address, and all of the addresses will move as a single unit.
+====
+
+Alternatively, you can use the `OPENSHIFT_HA_VIP_GROUPS` environment variable.
```
kalexand-rh: I'd remove this line. And if you have time/inclination, make the relationship between the command line parameters and the environment variables clearer earlier in the topic.

admin_guide/high_availability.adoc
```diff
@@ -542,6 +542,14 @@ See xref:../admin_guide/high_availability.adoc#ha-vrrp-id-offset[this discussion
 |0
 |See xref:../admin_guide/high_availability.adoc#ha-vrrp-id-offset[VRRP ID Offset] discussion for more details.
 
+|`--virtual-ip-groups`
+|`OPENSHIFT_HA_VIP_GROUPS`
+|
+|Number of groups to create for VRRP. If not set, a group is created for each
+virtual IP given with `--virtual-ips`. See
```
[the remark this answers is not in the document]
bfallonf: Changed to "If not set, a group is created for each virtual IP range specified with the `--virtual-ips` option."

admin_guide/high_availability.adoc
```diff
@@ -580,6 +588,51 @@ the list of VIPs.  The vrrp-ids may have values in the range 1..255.
 
 When there are multiple ipfailover deployment configuration care must be taken to specify `--vrrp-id-offset` so that there is room to increase the number of VIPS in the deployment configuration and none of the vrrp-id ranges overlap.
 
+[[admin-guide-high-availability-configuring-more-than-254]]
+=== Configuring IP failover for more than 254 addresses
+
+By default, IP failover has a limitation of managing only 254 VIP addresses. You
```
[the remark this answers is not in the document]
bfallonf: Changed to "By default, IP failover management is limited to 254 VIP addresses."

bfallonf:
Thanks @ahardin-rh @kalexand-rh . I've changed some things, and kept some the same. I'll merge this when I get the thumbs up from eng.

admin_guide/high_availability.adoc
```diff
@@ -580,6 +588,54 @@ the list of VIPs.  The vrrp-ids may have values in the range 1..255.
 
 When there are multiple ipfailover deployment configuration care must be taken to specify `--vrrp-id-offset` so that there is room to increase the number of VIPS in the deployment configuration and none of the vrrp-id ranges overlap.
 
+[[admin-guide-high-availability-configuring-more-than-254]]
+=== Configuring IP failover for more than 254 addresses
+
+By default, IP failover management is limited to 254 VIP addresses. You can
```
knobunc: IP failover management is limited to 254 groups.  By default we assign one IP address to each group.  You can change this so multiple IP addresses are in each group by using ...

admin_guide/high_availability.adoc
```diff
@@ -580,6 +588,54 @@ the list of VIPs.  The vrrp-ids may have values in the range 1..255.
...
+group sets of VIPs for VRRP instances by using the `virtual-ip-groups` option to
+define the number of VIP groups available for each VRRP instance when
+xref:configuring-ip-failover[configuring IP failover]. 
+
+Grouping VIPs creates a wider range of allocation of VIPs per VRRP in the case of VRRP
+failover events, and is useful when all hosts in the cluster have access to a
+service locally. For example, when a service is being exposed with an external
```
knobunc: "ExternalIP" is the field in the service... should we use that terminology?

admin_guide/high_availability.adoc
```diff
@@ -580,6 +588,54 @@ the list of VIPs.  The vrrp-ids may have values in the range 1..255.
 
 When there are multiple ipfailover deployment configuration care must be taken to specify `--vrrp-id-offset` so that there is room to increase the number of VIPS in the deployment configuration and none of the vrrp-id ranges overlap.
```
knobunc: That should be VIPs.

Did we ever define VIP as Virtual IP?

admin_guide/high_availability.adoc
```diff
@@ -580,6 +588,54 @@ the list of VIPs.  The vrrp-ids may have values in the range 1..255.
...
+host.
+====
+
+[NOTE]
+====
+If you are using {product-title} health checks, the nature of IP failover and
+groups means that all instances in the group will not be checked. For that
+reason,
+link:https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/[the
+Kubernetes health checks] are more effective.
```
knobunc: are more effective -> must be used to ensure that services are live.

knobunc (changes requested):
Thanks Brice.

admin_guide/high_availability.adoc
bfallonf: It did in the overview, but for the sake of topic reusability, I've added this in a few times. Good eye.

bfallonf:
Thanks @knobunc . Changes made. This has already gone through review, and I've given minimal changes, so I'll merge.

PR #8668, opened 2018-04-10
Bug 15069999 Add variables for replicas and shards

bmcelvee:
https://bugzilla.redhat.com/show_bug.cgi?id=1506999


bmcelvee:
@openshift/team-documentation PTAL :unicorn:

kalexand-rh:
@bmcelvee, I think this looks ok. It's odd that they're the only parameters in the table that have default values listed, though.

PR #8694, opened 2018-04-11
Added information about logging login failures

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1545116

@soltysh Does this adequately capture what you suggest in the BZ discussion? PTAL. Thanks!

### DIFF
--- a/install_config/master_node_configuration.adoc
+++ b/install_config/master_node_configuration.adoc
@@ -1261,6 +1261,12 @@ rules:
 
   # A catch-all rule to log all other requests at the Metadata level.
   - level: Metadata <1>
+
+  # Log login failures from the web console or CLI. Review the logs and refine your policies.
+  - level: Metadata
+    nonResourceURLs:
+    - /login* <8>
+    - /oauth* <9>
 ----
 <1> There are four possible levels every event can be logged at:
 +
@@ -1284,6 +1290,8 @@ that group.
 <5> A list of groups the rule applies to. An empty list implies every group.
 <6> A list of non-resources URLs the rule applies to.
 <7> A list of namespaces the rule applies to. An empty list implies every namespace.
+<8> Endpoint used by the web console.
+<9> Endpoint used by the CLI.
 
 For more information on advanced audit, see the
 link:https://kubernetes.io/docs/tasks/debug-application-cluster/audit[Kubernetes


install_config/master_node_configuration.adoc
```diff
@@ -1284,6 +1290,8 @@ that group.
 <5> A list of groups the rule applies to. An empty list implies every group.
 <6> A list of non-resources URLs the rule applies to.
 <7> A list of namespaces the rule applies to. An empty list implies every namespace.
+<8> Value used by the web console.
```
soltysh: s/value/endpoint

soltysh (approved):
One nit and you're good to go :)

install_config/master_node_configuration.adoc
```diff
@@ -1261,6 +1261,12 @@ rules:
 
   # A catch-all rule to log all other requests at the Metadata level.
   - level: Metadata <1>
+
+  # Log login failures from the web console or CLI. Turn on the audit, then create policy based on the requested data.
```
kalexand-rh: If both methods are required or encouraged, it should be "webconsole and CLI" 

 "Turn on the audit, then create policy based on the requested data." doesn't fit here. I might change it to a note to review the logs and refine your policies, as required, around line 1187.

install_config/master_node_configuration.adoc
```diff
@@ -1284,6 +1290,8 @@ that group.
 <5> A list of groups the rule applies to. An empty list implies every group.
 <6> A list of non-resources URLs the rule applies to.
 <7> A list of namespaces the rule applies to. An empty list implies every namespace.
+<8> Endpoint used by the web console.
```
kalexand-rh: Is this enough information for a user to construct the endpoints?

ahardin-rh:
@kalexand-rh Thanks for your comments. @soltysh Thoughts on that feedback?

install_config/master_node_configuration.adoc
soltysh: `or` is because you can log either, each rule applies to a different element. Let's be explicit about it.

install_config/master_node_configuration.adoc
soltysh: Not sure, what else we can add here. It's basically an information that this endpoint is where the logging from web console are going through.

ahardin-rh:
QE verified in the BZ

PR #8749, opened 2018-04-16
BZ 1435382 Change service IP addresses for online and dedicated docs

ghost:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1435382

### DIFF
--- a/dev_guide/managing_images.adoc
+++ b/dev_guide/managing_images.adoc
@@ -195,20 +195,15 @@ resource usage caused by old images.
 === Removing Tags from Image Streams
 To remove a tag completely from an image stream run:
 
-====
 ----
 $ oc delete istag/ruby:latest
 ----
-====
 
 or:
 
-====
 ----
 $ oc tag -d ruby:latest
 ----
-====
-
 
 [[referencing-images-in-image-streams]]
 === Referencing Images in Image Streams
@@ -287,7 +282,7 @@ For example, from the *ruby* image stream asking for the `*ImageStreamImage*`
 with the name and ID of *ruby@3a335d7*:
 
 .Definition of an Image Object Retrieved via `ImageStreamImage`
-====
+
 ----
 $ oc export isimage ruby@3a335d7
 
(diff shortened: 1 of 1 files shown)


tmorriso-rh:
@ahardin-rh Can you PTAL at the use of the ifdef/endif statement and the Labels for this PR? Thanks!

dev_guide/managing_images.adoc
```diff
@@ -1089,16 +1090,48 @@ c8a648134623: Pushed
 latest: digest: sha256:be8bc4068b2f60cf274fc216e4caba6aa845fff5fa29139e6e7497bb57e48d67 size: 6273
 ----
 ====
+endif::openshift-enterprise[]
++
+ifdef::openshift-origin,openshift-dedicated[]
```
ahardin-rh: According to Shawn, it looks like openshift-online needs to be included here too

dev_guide/managing_images.adoc
```diff
@@ -1089,16 +1090,48 @@ c8a648134623: Pushed
...
+====
+----
+$ oc get is
+NAME       DOCKER REPO                        TAGS      UPDATED
+my-image   172.30.48.125:5000/test/my-image   latest    3 seconds ago
+----
+====
+endif::openshift-enterprise[]
++
+ifdef::openshift-origin,openshift-dedicated[]
```
ahardin-rh: looks like openshift-online is missing here too

ahardin-rh:
@tmorriso-rh Since this file is shared with OCP, Online, and Dedicated, we should also make sure these changes are labeled for and cherry-picked to OCP 3.7, 3.9, and 3.10. (Dedicated and Online are both currently running on OCP 3.7. Tomorrow, Online will upgrade to 3.9.) So, even though the changes are mostly pertinent to Online and Dedicated, the files are still shared with OCP and so we have to make sure changes are applied in those branches as well. You are correct that no changes to the topic map are needed.

ahardin-rh:
@tmorriso-rh LGTM! :+1:

PR #8766, opened 2018-04-17
[online] to 3.9

vikram-redhat:


vikram-redhat:
@ahardin-rh @aheslin I have left this PR open and not merged it to move the online docs to 3.9. Just hit merge when you have confirmation from Abhishek or Steve. Since this will merge into master the build will start immediately and should be done within 10 minutes to move the online docs to 3.9.

PR #8767, opened 2018-04-17
Bug 1563929 - add a 3.9 upgrade known issues section.

soltysh:
3.9 introduces tighter validation for Service objects. During storage upgrade this might require administrators to invoke manual updates to the erroneous objects.

@openshift/team-documentation this is targeted for 3.9 only.


vikram-redhat:
@soltysh - when you say 3.9 only, do you mean 3.9+ or is this different in 3.10?

@ahardin-rh PTAL

soltysh:
> @soltysh - when you say 3.9 only, do you mean 3.9+ or is this different in 3.10?

@vikram-redhat this is only 3.9 (and all subversions :wink:). 
3.10 will add this back, *only* if needed, so for 3.10 we need to hide that page back, similarly how it was not present in 3.7.

upgrading/upgrading_known_issues.adoc
```diff
@@ -48,181 +45,25 @@ fatal: [test.linux.lan]: FAILED! => {
...
-done
-----
-
-. Run the script while logged in to the CLI as a cluster administrator:
-+
-----
-$ bash oauthclientauthorization.sh
-----
+. Log in to the CLI as a cluster administrator and inspect the `Service` objects
+listed in the error. Edit the `.spec.ports` of the objects removing duplicate
```
ahardin-rh: I would make "Edit..."  its own numbered step, so:

. Log in to the CLI as a cluster administrator and inspect the `Service` objects listed in the error.

. Edit the `.spec.ports` of the objects removing duplicate . . .

ahardin-rh:
@soltysh Thank you! Just one style nit.

Based off of the provided explanation, I think that this is still needing to be applied against the master branch and then cherry-picked to enterprise-3.9 (instead of being applied against enterprise-3.9 only), whether or not it will persist in the 3.10 doc set (which it sounds like it will not). @vikram-redhat Thoughts?

soltysh:
@ahardin-rh comments addressed and opened #8781 against master.

PR #8804, opened 2018-04-19
Add note about --fs-group option

dmage:
https://bugzilla.redhat.com/show_bug.cgi?id=1447249

/cc @bparees @miminar @legionus

### DIFF
--- a/install_config/registry/deploy_registry_existing_clusters.adoc
+++ b/install_config/registry/deploy_registry_existing_clusters.adoc
@@ -79,6 +79,10 @@ To see a full list of options that you can specify when creating the registry:
 ----
 $ oc adm registry --help
 ----
+
+Please note that the value for `--fs-group` must be
+xref:../../install_config/persistent_storage/pod_security_context.adoc#fsgroup[permitted by the SCC]
+used by the registry (typically, the restricted SCC).
 endif::[]
 
 [[registry-daemonset]]


install_config/registry/deploy_registry_existing_clusters.adoc
```diff
@@ -79,6 +79,9 @@ To see a full list of options that you can specify when creating the registry:
 ----
 $ oc adm registry --help
 ----
+
+Please note that the value for the `--fs-group` option should be agreed with
```
bparees: /should be agreed with/must be permitted by the fsgroup setting of the SCC/ 

(which SCC is used?  restricted?  they need to know which one to look at)

bparees (changes requested):

install_config/registry/deploy_registry_existing_clusters.adoc
dmage: In general you don't know which one will be used.

If you use `oc cluster up`, the restricted SCC is applied and in this case you should choose a value from the range:
```
$ oc -n default get ns default -o template --template='{{index .metadata.annotations "openshift.io/sa.scc.supplemental-groups"}}{{"\n"}}'
1000000000/10000
```

Or you can create an SCC with the specific supplemental group and add the service account to it. And this SCC will be applied to the registry automagically if you create the deployment config with this group.

bparees: ok, i'd just hand wave something like `the value for --fs-group must be permitted by the SCC used by the registry (typically, the restricted SCC)`

bparees:
lgtm, @openshift/team-documentation ptal

bmcelvee:
Thanks, @bparees! I will take a look at this in a bit.

bmcelvee:
This applies to OCP 3.6 forward, correct?

bparees:
yes

PR #8812, opened 2018-04-19
[enterprise-3.4] Change v3.4.1 to v3.4 in Syncing Images section

ghost:
@openshift/team-documentation PTAL. The `{latest-int-tag}` pulls in "3.4.1," but the bug reporter states this will not pull in the latest image. I manually changed this to `3.4`. Is this correct? Thanks.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1537101

### DIFF
--- a/install_config/install/disconnected_install.adoc
+++ b/install_config/install/disconnected_install.adoc
@@ -3,7 +3,7 @@
 {product-author}
 {product-version}
 :latest-tag: v3.4.1.44.38
-:latest-int-tag: 3.4.1
+:latest-int-tag: v3.4
 :latest-registry-console-tag: 3.4
 :data-uri:
 :icons:


install_config/install/disconnected_install.adoc
```diff
@@ -193,7 +193,7 @@ will not recycle, potentially causing errors.
 . Pull all of the required {product-title} containerized components for the
 additional centralized log aggregation and metrics aggregation components.
 ifdef::openshift-enterprise[]
-Replace `<tag>` with `{latest-int-tag}` for the latest version.
+Replace `<tag>` with `3.4` for the latest version.
```
ahardin-rh: You can leave the {latest-int-tag} here, but update the value for latest-int-tag: 3.4.1 value at the top of the file, in the file's meta data. I'm looking at the comments in the BZ and it looks like the proposed new value should be v3.4.

PR #8837, opened 2018-04-20
Add IMPORTANT note about what to do if migration playbooks fail

kalexand-rh:
Follow-up to #7840

### DIFF
--- a/upgrading/migrating_etcd.adoc
+++ b/upgrading/migrating_etcd.adoc
@@ -64,6 +64,14 @@ Embedded etcd to External etcd].
 [[etcd-data-migration-automated]]
 == Running the Automated Migration Playbook
 
+[IMPORTANT]
+====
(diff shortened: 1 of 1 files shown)


upgrading/migrating_etcd.adoc
```diff
@@ -64,6 +64,14 @@ Embedded etcd to External etcd].
 [[etcd-data-migration-automated]]
 == Running the Automated Migration Playbook
 
+[IMPORTANT]
+====
+If the migration playbooks fail before the masters are reconfigured to support 
+etcd v3 storage, you must
+xref:etcd-data-migration-recovering[roll back] the migration process.
```
kalexand-rh: @ahardin-rh, I shortened the first link to just [roll back] because it was hard to differentiate the sequential links.

ahardin-rh: Looks good!

PR #7364, opened 2018-01-26
New topic for Kubernetes descheduler

mburke5678:
https://trello.com/c/nulaT5Wp/670-document-implement-basic-descheduler

PR #8853, opened 2018-04-20
Revert "New topic for Kubernetes descheduler"

mburke5678:
Reverts openshift/openshift-docs#7364
The cherry-pick to 3.9 was reverted; but, was not reverted from master.


vikram-redhat:
@mburke5678 - Hey Michael - why was this removed from master? I can understand 3.9 but I believe this should have remained in master as well?

PR #8866, opened 2018-04-24
Add a tool for generating master and node configuration description file

soltysh:
@openshift/team-documentation this tool will allow generating a reasonable tabular documentation for master and node config. Lemme know what you think. 

@derekwaynecarr @mfojtik we've talked about it some time ago, sorry it took me this long :blush: 
Derek, do we want to generate appropriate docs for 3.9 as well?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,8 @@
 all:
 	./_update_rest_api.py
-
 .PHONY: all
+
+master-node-api:
+	# this invocation assumes you have origin's code available in GOPATH
+	go run main.go github.com/openshift/origin/pkg/cmd/server/apis/config/v1>install_config/master_node_configuration_api.adoc
+.PHONY: master-node-api
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -341,6 +341,8 @@ Topics:
     File: uninstalling
 - Name: Master and Node Configuration
   File: master_node_configuration
+- Name: Master and Node Configuration API
+  File: master_node_configuration_api
 - Name: OpenShift Ansible Broker Configuration
   File: oab_broker_configuration
   Distros: openshift-origin,openshift-enterprise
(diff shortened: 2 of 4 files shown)


kalexand-rh:
@vikram-redhat, thoughts?

vikram-redhat:
This is a hard one after the F2F last week. This is clearly reference documentation which will be good for developers, but I don't know if we want to add this to install docs and how to modularize it.

@soltysh - is this for origin only and/or for 3.7? How does the script actually run? Would we need to run this script every few days?

soltysh:
> @soltysh - is this for origin only and/or for 3.7? How does the script actually run? Would we need to run this script every few days?

This should be working on any release. For the actual run see the `Makefile`. It needs running only when API gets updated, which should not happen that often, but running it automatically every X days does make sense.

soltysh:
This doc is more useful for administrator modifying master- and node- configs more than devs.

soltysh:
@openshift/team-documentation do we want to pursuit this one, I think this will be a valuable piece of information for the admins in 3.10.

vikram-redhat:
@soltysh - ok let me figure out where this goes with the team. 

@aheslin - sorry if we talked about this already, but let's have another chat?

vikram-redhat:
@soltysh - sorry still working on this one. Give me a few more days.

vikram-redhat:
@soltysh - sorry Maciej this has taken so long, but in the end we decided to not go ahead with this as it doesn't offer much value to an administrator. Further, there are maintenance issues with this script that we don't have the bandwidth to work on. 

I am going to close this PR, but please feel free to reopen if you have more details as to why it would fit into a good user story for admins/devs.

@aheslin

soltysh:
The original idea was to provide extensive guidance about master config fields. But with the upcoming changes it might not be necessary anymore. We can always return back to this, if needed.

PR #8870, opened 2018-04-24
added restart after yum update per BZ

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1569959


mburke5678:
@jianlinliu -- Can you take a look at this PR? I don't think it will cause a problem to restart after yum update. Just want to make sure.

jianlinliu:
lgtm.

PR #8900, opened 2018-04-25
Brief note about timeouts in the haproxy router

JacobTanenbaum:
Some of the timeouts that we allow cluster administrators to set
interact with others that we expose or in some cases don't expose.
Causing haproxy to actually timeout as the sum of those timeouts

Ex.

setting
ROUTER_SLOWLORIS_HTTP_KEEPALIVE to 10s

because we set "tcp-request inspect-delay" to 5s

ROUTER_SLOWLORIS_HTTP_KEEPALIVE triggers and then the inspect-delay
waits to timeout so the connection closes after 15s

bug 1562244
https://bugzilla.redhat.com/show_bug.cgi?id=1562244

### DIFF
--- a/architecture/topics/router_environment_variables.adoc
+++ b/architecture/topics/router_environment_variables.adoc
@@ -83,6 +83,11 @@ endif::[]
 
 [NOTE]
 ====
+Some of the timeouts that we expose interact casuing the total timeout to be longer then specified.
+====
+
+[NOTE]
+====
 If you want to run multiple routers on the same machine, you must change the
 ports that the router is listening on, `ROUTER_SERVICE_SNI_PORT` and
 `ROUTER_SERVICE_NO_SNI_PORT`. These ports can be anything you want as long as
--- a/architecture/topics/router_environment_variables.adoc
+++ b/architecture/topics/router_environment_variables.adoc
@@ -84,6 +84,9 @@ endif::[]
 [NOTE]
 ====
 Some of the timeouts that we expose interact casuing the total timeout to be longer then specified.
+
+For example:
+ROUTER_SLOWLORIS_HTTP_KEEPALIVE adjusts "timeout http-keep-alive", but haproxy still waits on the "tcp-request inspect-delay" which we set to 5s. So the overall timeout will be ROUTER_SLOWLORIS_HTTP_KEEPALIVE + 5s.
 ====
 
 [NOTE]


JacobTanenbaum:
@bfallonf PTAL

not sure if I should provide a specific example or not

akostadinov:
It's good to have this specific example because it is easy to hit. Also (even more useful) to link to the relevant HAProxy documentation where more detailed information can be found.

architecture/topics/router_environment_variables.adoc
```diff
@@ -81,6 +81,11 @@ endif::[]
 |`ROUTER_CIPHERS` | intermediate  | Specify the set of xref:ciphers[ciphers] supported by bind.
 |===
 
+[NOTE]
+====
+Some of the timeouts that we expose interact casuing the total timeout to be longer then specified.
```
knobunc: Should we say:

e.g. ROUTER_SLOWLORIS_HTTP_KEEPALIVE  adjusts "timeout http-keep-alive" , but haproxy will waits on the "tcp-request inspect-delay" which we set to 5s.

knobunc (changes requested):

knobunc (approved):
/lgtm

akostadinov:
while on it, is `tcp-request inspect-delay` adjustable or is it presently hardcoded?

JacobTanenbaum:
@akostadinov tcp-request inspect-delay is hardcoded

akostadinov:
Thank you. I think we need to make these parameters configurable. 5 seconds is a pretty big value. No need to wait for people with big installations to request that. But this is for another forum.

bfallonf:
@JacobTanenbaum Thanks, this is good info to have in the docs.

I'm concerned people might not see it where it is, but I can't see a better location. Below is a section called Timeouts, that has almost literally no information. Would it be better to group the relevant timeout variable there? I don't mind making these changes in a followup PR if necessary.

JacobTanenbaum:
@bfallonf are you recommending making a section about timeouts? I don't see one

akostadinov:
@JacobTanenbaum , you can refer to the `Note` from the `ROUTER_SLOWLORIS_HTTP_KEEPALIVE` parameter description (such that users notice it). For me this would resolve the immediate issue. I don't mind if options are grouped better if needed though.

bfallonf:
I spoke to @JacobTanenbaum over email that it'd be good to see what grouping together the timeout variables would look like. I'll merge this PR, then make a followup giving that a go. Thanks, all.

Also, I'll label this for 3.9 because that's the version the BZ above is set to. If I'm wrong, please correct me.

PR #8905, opened 2018-04-25
Add ovs-networkpolicy to the list of EgressNetworkPolicy plugins

### LINKED ISSUE #7488: EgressNetworkPolicy is available with ovs-networkpolicy
admin_guide/managing_networking.adoc, in the section "Using an Egress Firewall to Limit Access to External Resources" says:

> You must have the ovs-multitenant plugin enabled in order to limit pod access via egress policy.

But egress policy works with ovs-networkpolicy too. So either it should say "you must have ovs-multitenant or ovs-networkpolicy enabled", or it should say "it doesn't work with ovs-subnet".

knobunc:
The docs needed to have ovs-networkpolicy added to the list of plugins that supportEgressNetworkPolicy.

This should be fixed in 3.9 and beyond.

Fixes https://github.com/openshift/openshift-docs/issues/7488

### DIFF
--- a/admin_guide/managing_networking.adoc
+++ b/admin_guide/managing_networking.adoc
@@ -290,7 +290,7 @@ access to `<project B>`. Or restrict application developers from updating from
 [CAUTION]
 ====
 You must have the
-xref:../install_config/configuring_sdn.adoc#install-config-configuring-sdn[*ovs-multitenant* plugin] enabled in order to limit pod access via egress policy.
+xref:../install_config/configuring_sdn.adoc#install-config-configuring-sdn[*ovs-multitenant* or *ovs-networkpolicy* plugin] enabled in order to limit pod access via egress policy.
 ====
 
 Project administrators can neither create `EgressNetworkPolicy` objects, nor


admin_guide/managing_networking.adoc
```diff
@@ -290,7 +290,7 @@ access to `<project B>`. Or restrict application developers from updating from
 [CAUTION]
 ====
 You must have the
-xref:../install_config/configuring_sdn.adoc#install-config-configuring-sdn[*ovs-multitenant* plugin] enabled in order to limit pod access via egress policy.
+xref:../install_config/configuring_sdn.adoc#install-config-configuring-sdn[*ovs-multitenant* or *ovs-networkpolicy* plugin] enabled in order to limit pod access via egress policy.
```
bfallonf: @knobunc Would this be better to link to parts of the link where it says to enable the plugins? So:

----
You must have the xref:../install_config/configuring_sdn.adoc#migrating-between-sdn-plugins[*ovs-multitenant* or *ovs-networkpolicy*] plugin enabled in order to limit pod access via egress policy.
----

Alternatively, should this link to the current doc where it says "Enabling NetworkPolicy"[1], although there isn't a section for Multitenent it looks like...

[1] https://docs.openshift.com/container-platform/3.9/admin_guide/managing_networking.html#admin-guide-networking-networkpolicy

PR #8908, opened 2018-04-26
Sanitize openshift_master_cluster_hostname

fbac:
openshift_master_cluster_hostname should point to
the internal LB or internal master hostname.
Having the same values in this and in the public URL
leads to installation errors and missconfigurations.


kalexand-rh:
@wanghaoran1988, does this change look ok to you?

wanghaoran1988:
@kalexand-rh LGTM, thanks

PR #8948, opened 2018-04-26
Bug 1557209, Updated openshift_master_oauth_templates

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1557209

### DIFF
--- a/install_config/web_console_customization.adoc
+++ b/install_config/web_console_customization.adoc
@@ -866,7 +866,7 @@ xref:../install_config/install/advanced_install.adoc#advanced-install-configurin
 - xref:changing-the-logout-url[`openshift_master_logout_url`]
 - xref:loading-custom-scripts-and-stylesheets[`openshift_web_console_extension_script_urls`]
 - xref:loading-custom-scripts-and-stylesheets[`openshift_web_console_extension_stylesheet_urls`]
-- xref:customizing-the-login-page[`openshift_master_oauth_template`]
+- xref:customizing-the-login-page[`openshift_master_oauth_templates`]
 - xref:../install_config/cluster_metrics.adoc#install-config-cluster-metrics[`openshift_master_metrics_public_url`]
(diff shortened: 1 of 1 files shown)


install_config/web_console_customization.adoc
```diff
@@ -887,7 +886,6 @@ xref:../install_config/install/advanced_install.adoc#advanced-install-configurin
 
 # Configure a custom login template in the master config
 # See: https://docs.openshift.com/enterprise/latest/install_config/web_console_customization.html#customizing-the-login-page
```
spadgett: The two lines of comments above should be removed as well.

spadgett:
It's not clear to me why this is deprecated. Is it only the installer var deprecated or the entire feature?

cc @simo5 @enj

enj:
> It's not clear to me why this is deprecated. Is it only the installer var deprecated or the entire feature?

@spadgett looks like it was moved around in Ansible: https://github.com/openshift/openshift-ansible/pull/6522

Nothing has changed in OpenShift; the login / provider / error templates continue to work as before.

ahardin-rh:
Extra lines are now removed. Standing by for guidance on an updated example.

ahardin-rh:
@michaelgugino Can you please help with an example?

michaelgugino:
@ahardin-rh 

https://github.com/openshift/openshift-ansible/blob/master/inventory/hosts.example#L97-L98

We try to document most user-accessible variables in hosts.example.

spadgett:
@michaelgugino Can you use that syntax in an inventory file? I didn't think that worked when using INI.

michaelgugino:
@spadgett that's a good point.  I'll get that fixed, and link the PR here.

michaelgugino:
@ahardin-rh @spadgett Updated example here: https://github.com/openshift/openshift-ansible/pull/8460

ahardin-rh:
@spadgett @michaelgugino Thanks! Updated PTAL

spadgett (approved):
/lgtm

ahardin-rh:
QE verified in the BZ

PR #8959, opened 2018-04-27
Bug 1537882, added information about enabling block volume feature gates

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1537882

@screeley44 PTAL. Thanks!

PR #9002, opened 2018-05-02
Follow-up to PR#8959, added link to Configuring for Local Volume

ahardin-rh:
This is a follow-up to https://github.com/openshift/openshift-docs/pull/8959

This is for 3.9 only (not 3.10) per discussion in the original PR


qinpingli:
LGTM

PR #9025, opened 2018-05-03
Update advanced LDAP docs to use BasicAuthPasswordIdentityProvider

enj:
This change moves the docs away from the frontend based
RequestHeaderIdentityProvider to the backend based
BasicAuthPasswordIdentityProvider.  This is because the LDAP docs
assume a password based flow which requires the RequestHeaderIDP
to support form based login.  In 3.3 and earlier (before the grant
approval flow existed), this form based login occurred in a single
step that did not require a session.  The grant flow changes this
flow to require more than one step, necessitating the need for a
session based approach.  Apache does not have a form module that
supports sessions, thus we cannot use the RequestHeaderIDP.  By
moving to the BasicAuthPasswordIDP, we leverage OpenShift's ability
to manage sessions directly.  This changes the network flow from:

RequestHeaderIDP:      User -> Apache -> OpenShift

to:

BasicAuthPasswordIDP:  User -> OpenShift -> Apache

From the perspective of Apache, the entire flow occurs in one step
and thus Apache does not need to maintain a session.  This allows
users to continue to use their investment in SSSD and PAM to have a
robust LDAP setup that supports automatic failover.  There is no
specific requirement on which PAM provider is used; these docs
simply present one specific approach.

This does have the negative side effect of making OpenShift see
user's passwords.  However, any individual that is concerned by this
should instead use the RequestHeaderIdentityProvider with SAML and
GSSAPI to eliminate the need for users to type passwords at all.

This is required for OCP 3.4+

Fixes bug:

Bug 1545548

Caused by fixing:

Bug 1421629
Bug 1434983
Bug 1439221
Bug 1439222

via https://github.com/openshift/origin/pull/13569

Signed-off-by: Monis Khan <mkhan@redhat.com>

/kind bug
/assign @kalexand-rh @simo5 @mrogers950 @stuartchuan 

---

I am pretty sure all of our future detailed examples need to be git repos with images and support for easy configuration changes.  Otherwise we risk all sorts of issues when a user makes a copy paste error with a long set of instructions like these.


install_config/advanced_ldap_configuration/sssd_for_ldap_failover.adoc
```diff
@@ -398,26 +379,22 @@ created. To do so:
...
 ----
 ====
 +
 [NOTE]
 ====
-xref:../advanced_ldap_configuration/configuring_form_based_authentication.adoc#configuring-form-based-authentication[Configuring
-Form-Based Authentication] explains how to add the login URL to support web
-logins.
-
 xref:../advanced_ldap_configuration/configuring_extended_ldap_attributes.adoc#configuring-extended-ldap-attributes[Configuring
```
kalexand-rh: Realistically, would a user ever want to complete the steps in this topic without completing the steps in the next topic?

install_config/advanced_ldap_configuration/sssd_for_ldap_failover.adoc
```diff
@@ -52,8 +52,8 @@ ldap.example.com).
 +
 . Prepare the VMs:
 +
-* *_proxy.example.com_*: A VM to use as the authenticating proxy. This machine must
-have at least SSSD 1.12.0 available, which means a fairly recent operating
+* *_remote-basic.example.com_*: A VM to use as the remote basic authentication server.
+This machine must have at least SSSD 1.12.0 available, which means a fairly recent operating
 system. This topic uses a Red Hat Enterprise Linux 7.2 server for its examples.
 +
 * *_openshift.example.com_*: A VM to use to run {product-title}.
```
kalexand-rh: Would a more realistic use-case be to set up the LDAP failover for an existing production cluster? (I don't see where this VM actually comes into play.)

install_config/advanced_ldap_configuration/configuring_extended_ldap_attributes.adoc
```diff
@@ -14,9 +14,7 @@ toc::[]
 == Overview
 This topic builds upon
 xref:../advanced_ldap_configuration/sssd_for_ldap_failover.adoc#setting-up-for-ldap-failover[Setting up SSSD
-for LDAP Failover] and
-xref:../advanced_ldap_configuration/configuring_form_based_authentication.adoc#configuring-form-based-authentication[Configuring
-Form-Based Authentication] and focuses on configuring extended Lightweight
+for LDAP Failover] and focuses on configuring extended Lightweight
```
kalexand-rh: Why would a cluster admin want to do this?

kalexand-rh:
@enj, we also need to figure out where to tell users this:
`This does have the negative side effect of making OpenShift see
user's passwords. However, any individual that is concerned by this
should instead use the RequestHeaderIdentityProvider with SAML and
GSSAPI to eliminate the need for users to type passwords at all.`

kalexand-rh:
Mo and I discussed this change, and I'm going to edit this on https://github.com/openshift/openshift-docs/pull/9070.

PR #9073, opened 2018-05-06
Introduce new limits for oc builds.

mrsiano:
Based on few vertical scale tests we've made recently.
especially for oc builds vertical, I'm updating this doc accordingly.

### DIFF
--- a/scaling_performance/cluster_limits.adoc
+++ b/scaling_performance/cluster_limits.adoc
@@ -51,6 +51,10 @@ version or storage data format.
 | 10,000
 | 10,000
 
+| Number of builds: Pipeline Strategy
+| N/A
+| 10,000 (Default pod RAM 512Mi)
+
 | Number of pods per namespace footnoteref:[objectpernamespace,There are
 a number of control loops in the system that need to iterate over all objects
 in a given namespace as a reaction to some changes in state. Having a large
@@ -71,7 +75,6 @@ expensive and slow down processing given state changes.]
 | 20,000
 | 20,000
 
-|===
 
 [[scaling-performance-planning-your-environment-according-to-cluster-limits]]
 == Planning Your Environment According to Cluster Limits


mrsiano:
@ahardin-rh @gabemontero @jeremyeder

scaling_performance/cluster_limits.adoc
```diff
@@ -51,6 +51,10 @@ version or storage data format.
 | 10,000
 | 10,000
 
+| Number of builds
+| N/A
+| 10,000
```
gabemontero: If this is based only on the jenkins pipeline strategy build tests we recently did, this feels like a pretty generic claim @mrsiano .... our other build strategies would not influenced by the Jenkins constraints we were investigating.

Or am I missing some additional background here ?

@bparees fyi

bparees: and our other build strategies would be much more subjected to the number of pods the cluster was capable of running since, unlike a pipeline build, "regular" builds need a pod for each build.

mrsiano: right, so stage 1 was to make sure the Jenkins plugin core working as expected, so we narrow down other threats to identify the plugin capabilities.

what do you think about sperate the build capacity to two diffrent aspects as follow:
- Number of builds  runs by jenkins-sync-plugin.
- Number of builds

Ashley, could you suggest how to name it correctly ?

gabemontero: Yep we need separated categories, minimally with the Pipeline Strategy Builds separated out (that is the name for what sync plugin handles).

I don't think we need further separation regarding capacity for the source/docker/custom strategies, but I'll defer to @bparees

bparees: +1

ahardin-rh: Based on Gabe's comments, perhaps we can label these as Pipeline Strategy Builds and Total Builds.

mrsiano:
Once we covered to total number I'll submit another PR

scaling_performance/cluster_limits.adoc
```diff
@@ -51,6 +51,10 @@ version or storage data format.
 | 10,000
 | 10,000
 
+| Number of pipeline strategy builds
```
mffiedler: @mrsiano Is this defined build configurations?  concurrent running builds?  Probably best to qualify it.

mffiedler (changes requested):

scaling_performance/cluster_limits.adoc
gabemontero: Also @mrsiano - for the pipeline stuff, I would be inclined to also provide the following details
a) that the 10K number was achieved with 512Mi container limit and 256Mi JVM heap
b) and that when you doubled those number, you got 20 something K builds  (if I am remembering your findings correctly)

Thoughts?

scaling_performance/cluster_limits.adoc
mrsiano: I believe it already done, thats what we've tested and tune for such a scale thought this plugin effort.

mrsiano:
@gabemontero @ahardin-rh 
since this 10K number referring to default limits which is 512mb as RAM used by the Jenkins pod.
I guess we can submit it as is.

otherwise we need to extend the matrix.

scaling_performance/cluster_limits.adoc
mffiedler: I still think clarification might help.  Is it running builds or defined builds?

mrsiano: the sample-pipline are defined builds right @gabemontero

gabemontero: I wouldn't distinguish as "running vs. defined".

How about I describe the flow for pipeline strategy we employed in detail, and then you all can decide how to condense:
1) `oc start-build` of a pipeline strategy build config will create a build object in the new state
2) the openshift sync plugin running in our jenkins image will monitor for pipeline strategy build objects in the new state, read them into jenkins, and start a corresponding run in Jenkins (where the run is of the pipeline job created in jenkins when the pipeline strategy build config was created)

What @mrsiano measured for us was "how many builds could we reliably read into Jenkins at once, and still properly execute the runs for each new build".

What he discovered was:
a) at our default 512 Mi mem limit for the container, and 256Mi max heap for the JVM, we could read in 10K builds
b) when he doubled those 512/256 setting, we could read in 20K or 23K (somewhere around there)

gabemontero:
@mrsiano @ahardin-rh I think showing the linear growth and we increased memory is important.  I really want both the 10K/512Mi container mem/256 jvm heap finding and 23K/1Gi container mem/512Mi jvm heap finding listed in some fashion.

ahardin-rh:
@mrsiano Yeah, it may be best to create a new table to adequately show growth. If we need to show JVM/pipeline strategy vs other strategy builds, then it may be best to have one table for JVM and one for the others. That's my initial thought on this. If you want to try breaking it out, we can see how that looks and go from there.

mrsiano:
@mffiedler for plain setup how much builds bening populated, we can come up with a number since we gonna use 2 tables.

@ahardin-rh @gabemontero 
by my last commit I called the JVM table as "Number of pipeline strategy builds"
what about name it as ""Number of builds - Jenkins openshift-sync" or something similar

gabemontero:
@mrsiano @ahardin-rh we should stick with `Pipeline Strategy` .... do `Number of builds: Pipeline Strategy` so something along those lines works for me.

By comparison, the other table could be `Number of builds: Source/Docker/Custom Strategy`, assuming the testing shows they have similar numbers.  If that is not true, we break those out into separate tables as needed.

ahardin-rh:
@mrsiano The latest changes LGTM

PR #9107, opened 2018-05-08
Rework of Online docs for Fuse

ahardin-rh:
Preview Build:
http://file.rdu.redhat.com/~ahardin/05082018/fuse-online/getting_started/choose_a_plan.html


ahardin-rh:
Just discussed this with @wgordon17 and we seem to be on the same page. One point of discussion is that, in the console and in the docs, we need to keep consistent in the language that we are using when referencing plans and subscriptions. As we see it, a plan is something that you can get and once you have it, it is a subscription to that service. Not a blocker, but something to keep in mind as this content develops.

ahardin-rh:
@abhgupta PTAL

abhgupta:
The overall approach LGTM. Others can better review the content.

damemi (approved):

yufchang:
Looks good.
 "so you can get up and running quickly. " replaced by "so you can get up and run quickly",right?

ahardin-rh:
@yufchang "up and running" is a common phrase/idiom. The usage is correct here, but perhaps I can reword it.

getting_started/manage_your_plan/fuse_online.adoc
```diff
@@ -0,0 +1,22 @@
...
+:data-uri:
+:icons:
+:experimental:
+:toc: macro
+:toc-title:
+:prewrap!:
+
+[[customize-your-fuse-plan-with-add-ons]]
+== Customize Your Plan
+
```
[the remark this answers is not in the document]
lancejjohnson: @ahardin-rh Changing fuse plans isn't currently available. When it becomes available, I'm curious if the language "Customize" will be correct. With Fuse the user doesn't really have customization options. They will be able to choose from different pre-configured plans. Obviously out of my domain, but I would recommend against "Customize" in the header.

getting_started/manage_your_plan/fuse_online.adoc
```diff
@@ -0,0 +1,22 @@
...
+You can customize your plan with add-ons.
```
lancejjohnson: As I understand the fuse online requirements, this is actually not the case. The user can choose among preconfigured plans but cannot customize them.

getting_started/manage_your_plan/fuse_online.adoc
```diff
@@ -0,0 +1,22 @@
...
+
+. In the *Active Subscriptions* page, click the *Manage Subscription* link.
+
+. Click *Change* by the *Current Plan* heading
+
+. From the *Current Plan* page, click *Small*, *Medium*, or *Large*, then follow
```
lancejjohnson: Just so you're aware, "small", "medium", and "large" are the terms we currently have but I don't know that they have been finalized 100%.

lancejjohnson (commented):
Couple of things.

getting_started/manage_your_plan/fuse_online.adoc
ahardin-rh: @abhgupta When Fuse Online launches at the end of the month, will users be able to change plans yet, or is that coming at a later date? If so, I can just remove the Manage Your Fuse Online Plan page for now. Thanks!

abhgupta:
LGTM

PR #8810, opened 2018-04-19
Removed some sections from Automated Upgrades topic

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1565344

Preview Builds: 
http://file.rdu.redhat.com/~ahardin/04192018/web-console-in-automated/upgrading/automated_upgrades.html

http://file.rdu.redhat.com/~ahardin/04192018/web-console-in-automated/upgrading/manual_upgrades.html

PR #9119, opened 2018-05-09
Removed errant endif

ahardin-rh:
Addresses https://github.com/openshift/openshift-docs/pull/8810#discussion_r186589631


ahardin-rh:
@vikram-redhat FYI

Issue #9106, opened 2018-05-08
include yum-utils as base package


slmagus:
### Which section(s) is the issue in?

Hostprep 

2.3.4. Installing Base Packages Step 1.

### What needs fixing?

Need to include yum-utils for RH7 host system as a required package.
Not having the package generates errors when running the playbook, and as it isn't universally installed on a base RHEL system

vikram-redhat:
@kalexand-rh - can you take a look. I thought we did include this.

kalexand-rh:
@vikram-redhat, that command lists several utils packages, but not yum-utils. I opened a PR to add it: https://github.com/openshift/openshift-docs/pull/9121.

PR #9121, opened 2018-05-09
adding yum-utils to host prep

kalexand-rh:
@jianlinliu, we got a report on issue https://github.com/openshift/openshift-docs/issues/9106 that `yum-utils` was a required package for cluster hosts. Will you please confirm that this addition is correct and that it should apply back to version 3.3?


jianlinliu:
LGTM

kalexand-rh:
Thank you!

PR #9137, opened 2018-05-09
add resources to default imagepolicy list

bparees:


bparees:

goes back as far as this topic has existed.

admin_guide/image_policy.adoc
```diff
@@ -165,7 +165,7 @@ admissionConfig:
 <5> List of annotations to match on the image object's metadata.
 <6> If you are not able to resolve the image, do not fail the pod.
 <7> Array of rules allowing use of image streams in Kubernetes resources. The default configuration
-    allows pods, replication controllers, replica sets, deployments, and jobs
+    allows pods, replicationcontrollers, replicasets, statefulsets, daemonsets, deployments, and jobs
```
kalexand-rh: It looks like both "replicationcontrollers" and "replication controllers" are used in the collection. Do we have a usage preference here?

bparees: i'm not aware but the resource name has no space in it, so i went w/ that.

kalexand-rh: That's a good reason. :) LGTM

kalexand-rh:
I was mistaken - this line was added in 3.5, not 3.3.

Issue #8901, opened 2018-04-25
Deployment resources: explain difference between "limits" and "requests"


ge0ffrey:
The section here:
  https://docs.openshift.com/online/dev_guide/deployments/basic_deployment_operations.html#deployment-resources
doesn't explain the difference between "limits" and "requests" decently. 

Does it make sense to use both at the same time? What's the difference? What's the impact/behaviour? What will determine the amount of memory a pod actually gets?

vikram-redhat:
Hey @ge0ffrey ! Long time since working on the BxMS docs!

@ahardin-rh - PTAL.

ahardin-rh:
@pecameron Can you please help me tighten up the content here, or point me in the right direction?

pecameron:
@ahardin-rh In my reading of this, liimits: is optional and is used for general resource limits. However when a quota is specified (don't know where that is done), requests: is required to set a limit. Maybe pointing to the general discussions of limits and quotas, if they exist, would be helpful. If they don't exist the sections should be written. Depending on the limits, and quotas discussions, it may be useful to separate this into two sections, augmenting each discussion.
FWIW The above is my personal rambling. I have not worked with limits or quotas and don't really know what they do so this may be wildly wrong.

ahardin-rh:
@pecameron Thanks! I opened https://github.com/openshift/openshift-docs/pull/9182 to add a pointer to the appropriate docs. I will work with others to see if the information presented there needs to be expanded to fully answer the questions posed here. :beers:

PR #9182, opened 2018-05-10
Added pointer to Quotas and Limits topic

ahardin-rh:
Addresses https://github.com/openshift/openshift-docs/issues/8901

@miminar Can you please help me review this content, or point me in the right direction? To address the issue above, I added a pointer to the section that discusses requests vs limits, but we may not fully answer these questions that came up:

- Does it make sense to use both at the same time?
- What's the impact/behavior? 
- What will determine the amount of memory a pod actually gets?

Let me know if we should expand the content in https://docs.openshift.com/container-platform/3.9/dev_guide/compute_resources.html#dev-requests-vs-limits

### DIFF
--- a/dev_guide/deployments/basic_deployment_operations.adoc
+++ b/dev_guide/deployments/basic_deployment_operations.adoc
@@ -247,7 +247,6 @@ triggers:
 ----
 ====
 
-
 [[image-change-trigger]]
 === ImageChange Trigger
 //tag::image-change-trig[]
@@ -308,8 +307,6 @@ For more information, see:
 $ oc set triggers --help
 ----
 
-
-
 [[deployment-resources]]
 == Setting Deployment Resources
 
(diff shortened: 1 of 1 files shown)


miminar:
This looks to be decsribed in enough detail here https://docs.openshift.com/online/dev_guide/compute_resources.html#dev-compute-resources. Pointing there would definitely help.

> Does it make sense to use both at the same time?

Yes, requests as the minimum requirements and limits as the maximum boundary.

> What's the impact/behavior?

Again, this is better described in the link I mentioned before. In short, requests play the key role while scheduling - choosing the right node that satisfies the minimum requirements. And limits get applied when the container is running and asks for more.

> What will determine the amount of memory a pod actually gets?

The limits and what is actually availalble on the node where the pod is scheduled. But since there is also storage resource that can be restricted in similar fashion and there can be more in the future, I would avoid describing them all here and instead refer to the appropriate place. Maybe just once example (e.g. CPU) would be enough.

PR #9201, opened 2018-05-11
delete redundant 'also'

wozniakjan:
ptal @bmcelvee


bmcelvee:
LGTM! Thanks for catching this @wozniakjan! :star:

PR #9224, opened 2018-05-14
Removed passthroughs

vikram-redhat:


vikram-redhat:
Have applied this to various branches to fix https://github.com/openshift/openshift-docs/pull/5058#commitcomment-24924304

PR #9248, opened 2018-05-14
manual node certificate signing draft

kalexand-rh:
@sdodson, here's a very rough skeleton of what the manual node signing request docs might look like.  Will you please provide this information or tell me who else would have it?

- the inventory parameter that you set to `false` to enable this manual cert signing
- the list of certs this impacts
- the command to list the cert requests
- the flag to provide the cert
- steps to manually provide updated certs, if any
- which distros this change is valid in (it sounds like an origin and enterprise change, but please confirm)

### DIFF
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -592,6 +592,9 @@ Topics:
 - Name: Pruning Objects
   File: pruning_resources
   Distros: openshift-origin,openshift-enterprise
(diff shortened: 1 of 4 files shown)


kalexand-rh:
We rejected the Trello card that was associated with this PR.  @mburke5678, if anything here might help you with the new card, please take it!

PR #9253, opened 2018-05-15
Bug 1577830 Added important box about changing default SCC

bfallonf:
For: https://bugzilla.redhat.com/show_bug.cgi?id=1577830

Will verify in the BZ

cc @nak3

### DIFF
--- a/admin_guide/manage_scc.adoc
+++ b/admin_guide/manage_scc.adoc
@@ -34,6 +34,13 @@ As an {product-title} cluster administrator, you can list and view details for
 SCCs, but cannot edit or delete the default SCCs.
 endif::openshift-dedicated[]
 
+[IMPORTANT]
+====
+Do not modify the default SCCs. Customizing the default SCCs can lead to issues
+when upgrading. Instead, xref:creating-new-security-context-constraints[create
+new SCCs].
+====
+
 [[listing-security-context-constraints]]
 == Listing Security Context Constraints
 


nak3:
Thank you @bfallonf LGTM

@simo5 sorry for bothering. Do you have any comment or objections?

bfallonf:
[SCCs](https://trello.com/c/SxP4MYi1/212-sccs)

vikram-redhat:
@nak3 - let me know when ok to merge and push to branches. I am guessing it applies to 3.9 onwards.

bfallonf:
Thanks @simo5 . I'll merge.

admin_guide/manage_scc.adoc
```diff
@@ -34,6 +34,13 @@ As an {product-title} cluster administrator, you can list and view details for
 SCCs, but cannot edit or delete the default SCCs.
 endif::openshift-dedicated[]
 
+[IMPORTANT]
+====
+Do not modify the default SCCs. Customizing the default SCCs can lead to issues
+when upgrading. Instead, xref:creating-new-security-context-constraints[create
```
vikram-redhat: @bfallonf - this should have been conditionalized in the openshift-enterprise,openshift-origin ifdef above. This breaks the build in dedicated as it doesn't contain the xref you have linked to.

admin_guide/manage_scc.adoc
```diff
@@ -34,6 +34,13 @@ As an {product-title} cluster administrator, you can list and view details for
 SCCs, but cannot edit or delete the default SCCs.
 endif::openshift-dedicated[]
 
+[IMPORTANT]
+====
+Do not modify the default SCCs. Customizing the default SCCs can lead to issues
```
php-coder: I don't think that this change makes the SCC doc better for the following reasons:

1) we can't deny modifying SCC at all, because otherwise it won't be possible to grant access to SCCs for users and service accounts. In other words, commands like `oc adm policy add-scc-to-user` modifies SCC, particularly its Users/Groups members. So, this phrase is too strong at least and should be rephrased.

2) we [already have a warning](https://docs.openshift.org/latest/admin_guide/manage_scc.html#updating-the-default-security-context-constraints) that is less strict and follows the `oc adm policy reconcile-sccs` logic: users are allowed to modify only User, Groups, and Priority fields alongside with annotations and labels.

3) we have some places that recommends to modify basic SCCs, that should be updated also:
    - https://docs.openshift.org/latest/admin_guide/manage_scc.html#modify-cluster-default-behavior
    - https://docs.openshift.org/latest/admin_guide/manage_scc.html#use-the-hostpath-volume-plugin

php-coder: P.S. I read the original BZ again and I see that request was a bit different: they asked for making the existing warning more visible and clear that most of the SCC fields should be never modified.

nak3: Yes, I originally expected for this PR to update as described in BZ. 

However, when I read this strong statement and commit history of [securitycontextconstraints.go](https://github.com/openshift/origin/commits/master/pkg/cmd/server/bootstrappolicy/securitycontextconstraints.go) (= some capabilities have been dropped/added from/to default SCCs), I started thinking that changing default SCCs may lead to issues.

Having said that, is it possible for us to guarantee that modifying default scc (User/Group/Prioity) will not cause any issues during upgrade? (I think it is not 100% safe and creating and using new SCC is safer than modifying default SCCs.)

nak3: I agree that `oc adm policy add-scc-to-user` and docs you pointed out sounds conflict this statement.

php-coder: > Having said that, is it possible for us to guarantee that modifying default scc (User/Group/Prioity) will not cause any issues during upgrade?

It's safe in most of the cases. But it's possible to harm to the cluster even by altering only these settings :) Let's say, granting `anyuid` or `privileged` SCCs to all authenticated users, may broke some system pods.

php-coder: > we have some places that recommends to modify basic SCCs, that should be updated also

That PR should deal with these recommendations: https://github.com/openshift/openshift-docs/pull/9927

PR #8038, opened 2018-03-06
PR8027-Follow-up Add ifdef statements

bmcelvee:

PR #9262, opened 2018-05-15
Fixed an ifndef error

mburke5678:
I was alerted to this error from a translation question (the _pull-through feature_ phrase was repeated in Origin). 
I made this change to the `ifndef` statement to match how similar `ifdef` statements are formatted. It works in an *asciibinder build* preview. 

@openshift/team-documentation PTAL to make sure this command is now correct.

ps. I noticed after-the-fact that the error was fixed in 3.9 and 3.10. https://github.com/openshift/openshift-docs/pull/8038


kalexand-rh:
It looks right to me. I'm glad it works locally!

PR #9296, opened 2018-05-16
Bug#1576125 - fixes - added note for handling changes in the azure storageclass

gaurav-nelson:
for https://bugzilla.redhat.com/show_bug.cgi?id=1576125


gaurav-nelson:
it is needed in both 3.7 and 3.9 closing this, will create another PR

PR #9297, opened 2018-05-16
Bug#1576125 - fixes - added note for handling changes in the azure storageclass

gaurav-nelson:
For https://bugzilla.redhat.com/show_bug.cgi?id=1576125

### DIFF
--- a/install_config/persistent_storage/dynamically_provisioning_pvs.adoc
+++ b/install_config/persistent_storage/dynamically_provisioning_pvs.adoc
@@ -390,6 +390,17 @@ parameters:
 <1> Azure storage account SKU tier. Default is empty. *Note:* Premium VM can attach both _Standard_LRS_ and _Premium_LRS_ disks, Standard VM can only attach _Standard_LRS_ disks, Managed VM can only attach managed disks, and unmanaged VM can only attach unmanaged disks.
 <2> possible values are `shared` (default), `dedicated`, and `managed`. When `kind` is `shared`, all unmanaged disks are created in a few shared storage accounts in the same resource group as the cluster. When `kind` is `dedicated`, a new dedicated storage account gets created for the new unmanaged disk in the same resource group as the cluster. When `kind` is `managed`, a new managed disk gets created.
 
+[IMPORTANT]
+====
+Azure StorageClass is revised in {product-title} version 3.7. If you upgraded from a previous version, either:
+
+ * specify the property `kind: dedicated` to continue using the Azure StorageClass created before the upgrade. Or,
+ * add the location parameter (for example, `"location": "southcentralus",`)
+ in the *_azure.conf_* file to use the default property `kind: shared`. Doing
+this creates new storage accounts for future use.
+====
+
+
 
 [[change-default-storage-class]]
 == Changing the Default StorageClass
@@ -397,8 +408,7 @@ If you are using GCE and AWS, use the following process to change the default St
 
 . List the StorageClass:
 +
-
-====
+[source,bash]
 ----
 $ oc get storageclass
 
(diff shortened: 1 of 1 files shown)


install_config/persistent_storage/dynamically_provisioning_pvs.adoc
```diff
@@ -390,6 +390,15 @@ parameters:
 <1> Azure storage account SKU tier. Default is empty. *Note:* Premium VM can attach both _Standard_LRS_ and _Premium_LRS_ disks, Standard VM can only attach _Standard_LRS_ disks, Managed VM can only attach managed disks, and unmanaged VM can only attach unmanaged disks.
 <2> possible values are `shared` (default), `dedicated`, and `managed`. When `kind` is `shared`, all unmanaged disks are created in a few shared storage accounts in the same resource group as the cluster. When `kind` is `dedicated`, a new dedicated storage account gets created for the new unmanaged disk in the same resource group as the cluster. When `kind` is `managed`, a new managed disk gets created.
 
+[IMPORTANT]
+====
+Azure StorageClass is revised in {product-title} version 3.7. If you have upgraded from a previous version, either:
```
kalexand-rh: s/have upgraded/upgraded

I'd s/either:/you must take the following actions:

install_config/persistent_storage/dynamically_provisioning_pvs.adoc
```diff
@@ -390,6 +390,15 @@ parameters:
...
+
+ * specify the property `kind: dedicated` to continue using the Azure StorageClass created before the upgrade.
```
kalexand-rh: * To continue using the Azure StorageClass created before the upgrade, specify the property `kind: dedicated`.

install_config/persistent_storage/dynamically_provisioning_pvs.adoc
```diff
@@ -390,6 +390,15 @@ parameters:
...
+ * or, add location in the *_azure.conf_* file to use the default property `kind: shared`. Doing this creates new storage accounts for future use.
```
kalexand-rh: * To create new storage accounts for future use, add the location <of what?> in the *_azure.conf_* file and use the the fault property, `kind: shared`. 

I don't know what location you're using here. And can you not make new storage accounts if you use the old StorageClass?

kalexand-rh (commented):
I have a couple of questions and suggestions.

install_config/persistent_storage/dynamically_provisioning_pvs.adoc
gaurav-nelson: Thanks for the comment, I had the same question and I forgot to tag it here.

install_config/persistent_storage/dynamically_provisioning_pvs.adoc
gaurav-nelson: They could take either one of the two options, but not both. This is my understanding. I'll confirm this with the reporter. Thanks.

gaurav-nelson: @songbird159 Can you confirm?

gaurav-nelson:
@songbird159 Which location we should be adding to the Azure conf file?
```
add location in you azure conf to use the default "kind: Shared" and it will create few new storage accounts and keep using the new created ones.
```

songbird159:
@gaurav-nelson 
The name of the location to create a VM in. An example for it, would be the value 'eastus' it refers to EAST-US.
Reference.
https://docs.microsoft.com/en-us/azure/azure-resource-manager/xplat-cli-azure-resource-manager

gaurav-nelson:
Preview: http://bug1576125-fixes--gnelson-preview.netlify.com/openshift-enterprise/(head%20detached%20at%20fetch_head)/install_config/persistent_storage/dynamically_provisioning_pvs.html#azure-advanced-disk

gaurav-nelson:
Thank You @songbird159 Can you please review the changes?

PR #9319, opened 2018-05-16
bug 1503109 correcting node resource kubelet argument

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1503109

@mdshuai, will you PTAL or suggest another QE reviewer?


mdshuai:
The change LGTM, thanks

PR #9320, opened 2018-05-16
bug 1538391 fixing router version image number

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1538391

@vikram-redhat, are you ok with this change?


kalexand-rh:
Interesting. The file's in all these other branches, but this section's only in master and 3.5. I think it's fair to only correct the change in those two branches.

vikram-redhat:
@kalexand-rh yeah. As this is the release notes for 3.5, the ones in other sections are out of sync or not relevant.

kalexand-rh:
@vikram-redhat, that does make sense. Next time we discuss whether or not we should remove the old release notes from branches, I am fully in favor of ripping old versions out of newer branches. (And I'll put together the PRs to pull them, if you want.)

PR #9324, opened 2018-05-16
bug 1547396 fix disabling self-provisioning steps for 3.7

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1547396
These steps are different in 3.7, before 3.7, and after 3.7.

### DIFF
--- a/admin_guide/managing_projects.adoc
+++ b/admin_guide/managing_projects.adoc
@@ -90,37 +90,67 @@ to all authenticated developers by default.
 
 [[disabling-self-provisioning]]
 === Disabling Self-provisioning
-Removing the `self-provisioners`
-xref:../architecture/additional_concepts/authorization.adoc#roles[cluster role]
-from authenticated user groups will deny permissions for self-provisioning any new projects.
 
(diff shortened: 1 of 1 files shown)


admin_guide/managing_projects.adoc
```diff
@@ -90,39 +90,69 @@ to all authenticated developers by default.
...
+. Log in as a user with
+xref:../architecture/additional_concepts/authorization.adoc#roles[*cluster-admin*]
+privileges.
+. Remove the `self-provisioners`
+xref:../architecture/additional_concepts/authorization.adoc#roles[cluster role]
+from the group.
++
 ----
-Contact your system administrator at projectname@example.com to request a project.
+$ oc adm policy remove-cluster-role-from-group self-provisioner system:authenticated system:authenticated:oauth
```
cucushift: I think only 'system:authenticated:oauth' is enough.

barleyer: I think only 'system:authenticated:oauth' is enough.

PR #9340, opened 2018-05-16
Bug 1477340, Added corsAllowedOrigin to the Master and Node Configuration topic

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1477340

@sdodson @spadgett PTAL

### DIFF
--- a/install_config/master_node_configuration.adoc
+++ b/install_config/master_node_configuration.adoc
@@ -389,6 +389,14 @@ assetConfig:
     requestTimeoutSeconds: 0
 ----
 
+|`*corsAllowedOrigins*`
+|To access the API server from a web application using a different host name, you
+must whitelist that host name by specifying `corsAllowedOrigins` in the
+configuration field or by specifying the `--cors-allowed-origins` option on
+`openshift start`. No pinning or escaping is done to the value. See
+xref:../architecture/infrastructure_components/web_console.adoc#corsAllowedOrigins[Web
+Console] for example usage.
+
 |`*DisabledFeatures*`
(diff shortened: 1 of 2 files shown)


sdodson:
LGTM

ahardin-rh:

install_config/master_node_configuration.adoc
```diff
@@ -389,6 +389,14 @@ assetConfig:
     requestTimeoutSeconds: 0
 ----
 
+|`*corsAllowedOrigins*`
+|To access the API server from a web application using a different host name, you
+must whitelist that host name by specifying `corsAllowedOrigins` in the
+configuration field or by specifying the `--cors-allowed-origins` option on
+`openshift start`. No pinning or escaping is done to the value. See
```
kalexand-rh: I don't know what this means:
No pinning or escaping is done to the value.

Otherwise, LGTM!

spadgett: I believe it's a regular expression, rather than a simple string comparison. So `example.com` would match domains with the text `anotherexample.com` or even `examplescom`, etc. We probably want to call this out explicitly.

cc @liggitt

spadgett: Although I see there's an example if you follow the link.

liggitt: yes, the example added in https://github.com/openshift/openshift-docs/pull/8630/files explains the syntax/meaning

spadgett (approved):
lgtm

ahardin-rh:
QE approved in the BZ

PR #9358, opened 2018-05-17
Fix issue with online release notes

vikram-redhat:


vikram-redhat:
Will CP to enterprise 3.9 branch as online is at that version now.

PR #9372, opened 2018-05-17
Documentation wrong openshift_hosted_router_certificate

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1579348

### DIFF
--- a/install_config/certificate_customization.adoc
+++ b/install_config/certificate_customization.adoc
@@ -34,7 +34,7 @@ xref:../install_config/install/advanced_install.adoc#advanced-install-custom-cer
 ----
 openshift_master_overwrite_named_certificates=true <1>
 openshift_master_named_certificates=[{"certfile": "/path/on/host/to/crt-file", "keyfile": "/path/on/host/to/key-file", "names": ["public-master-host.com"], "cafile": "/path/on/host/to/ca-file"}] <2>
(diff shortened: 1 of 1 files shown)


mburke5678:
@ikke-t @sdodson  I got code from an issue that you worked on that included a bracket with the `openshift_hosted_router_certificate=[{"certfile": "/path/on/host/to/app-crt-file", "keyfile": "/path/on/host/to/app-key-file", "cafile": "/path/on/host/to/app-ca-file"}]` command. 
A customer says that the brackets are not needed, based on upstream: https://github.com/openshift/openshift-ansible/blob/master/inventory/hosts.example#L395
Can you verify which is correct in OpenShift?

mburke5678:
@jianlinliu @ikke-t @sdodson
PTAL

sdodson (approved):
It's fine to remove the brackets. I wouldn't have expected them to cause a problem though.

mburke5678:
@jianlinliu PTAL

jianlinliu:
LGTM.

ikke-t:
The brackets would be needed if it was a list. If it accepts non-lists, then it's fine to remove them, as there is only one item {} there.

mburke5678:
@openshift/team-documentation PTAL tiny change. Just wanted to make sure I didn't miss something

kalexand-rh:
LGTM :+1:

PR #9438, opened 2018-05-19
specify an explicit resolution type in example resolution rules

### LINKED ISSUE #9403: Missing value for ImagePolicy.resolutionRules.policy.

https://docs.openshift.com/container-platform/3.9/admin_guide/image_policy.html#image-policy-configuring-the-image-policy-admission-plug-in

### What needs fixing?

If you use the example ``ImagePolicy`` config example it will fail.

```
F0518 05:10:28.629994   30168 start_allinone.go:99] Server could not start: Couldn't init admission plugin "openshift.io/ImagePolicy": [openshift.io/ImagePolicy.resolutionRules[0].policy: Required value: a policy must be specified for this resource, openshift.io/ImagePolicy.resolutionRules[1].policy: Required value: a policy must be specified for this resource]
```

The error suggests example is missing:

```
ImagePolicy.resolutionRules.policy
```

entries.

Looks like needs to have:

```
policy: AttemptRewrite
```

entry as policy to avoid error.

This needs to be confirmed by someone in the know as it is a guess that is what is required.

bparees:
fixes https://github.com/openshift/openshift-docs/issues/9403

### DIFF
--- a/admin_guide/image_policy.adoc
+++ b/admin_guide/image_policy.adoc
@@ -151,10 +151,12 @@ admissionConfig:
         - targetResource:
             resource: pods
           localNames: true
+          policy: AttemptRewrite
         - targetResource: <8>
             group: batch
             resource: jobs
           localNames: true <9>
+          policy: AttemptRewrite
 
 ----
 <1> Try to resolve images to an immutable image digest and update the image pull specification in the pod.


bparees:
@GrahamDumpleton ptal.  I think it's reasonable to update the sample to be explicit even if we do change the defaulting behavior in my other PR.

bparees:
@GrahamDumpleton does this match what you had to do?

GrahamDumpleton:
Nope, I just set "resolutionRules" to null and it defaulted to something that worked. See: https://github.com/minishift/minishift/issues/2392#issuecomment-390108350

kalexand-rh:
LGTM. 
@bparees, this example goes back to 3.6. Do you agree that the change applies to 3.6-3.10?

bparees:
@kalexand-rh yeah

PR #9490, opened 2018-05-22
Clarify use of # and $ in code blocks

kalexand-rh:

### DIFF
--- a/contributing_to_docs/doc_guidelines.adoc
+++ b/contributing_to_docs/doc_guidelines.adoc
@@ -467,6 +467,13 @@ $ oc get endpoints --all-namespaces --template \
     {{ end }}{{ end }}{{ "\n" }}{{ end }}' | awk '/ 172\.30\./ { print $1 }'
 ----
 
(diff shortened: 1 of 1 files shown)


vikram-redhat:
@kalexand-rh - do we have anything on this in our CCS guidelines? I am guessing this is via the IBM style guide?

kalexand-rh:
@vikram-redhat, it's not in the glossary or the IBM Style guide. (IBM doesn't use either to differentiate commands and instead adds copy widgets so you can easily capture the entire command for easy pasting into your CLI. I once spent two days purging a collection of all the leading dollar signs on commands.)
I was told in our IRC channel that that was the way we were using them and couldn't think of a better place to record it.

vikram-redhat:
@kalexand-rh - great. You should consider submitting it to the CCS Guidelines as well.

vikram-redhat:
In the meantime @openshift/team-documentation any objections?

tmorriso-rh:
LGTM.

PR #9553, opened 2018-05-23
bug BZ1465049 generate router certificates

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1465049

### DIFF
--- a/install_config/redeploying_certificates.adoc
+++ b/install_config/redeploying_certificates.adoc
@@ -588,11 +588,11 @@ annotations to the `router` service, then redeploying the router:
 $ oc project default
 ----
 
-. If your router was initially created on {product-title} 3.1 or earlier, it may
-still be using environment variables to store certificates (which has been
-deprecated in favor of using service serving certificate secret).
+. If your router was initially created on {product-title} 3.1 or earlier, it might
+still use environment variables to store certificates, which has been
+deprecated in favor of using service serving certificate secret.
 
-.. Run the following and look for the
+.. Run the following command and look for the
 `OPENSHIFT_CA_DATA`, `OPENSHIFT_CERT_DATA`, `OPENSHIFT_KEY_DATA` environment
 variables:
 +
@@ -600,7 +600,7 @@ variables:
 $ oc env dc/router --list
 ----
 
-.. If they do not exist, skip this step. If they do, create the following `ClusterRoleBinding`:
+.. If those variables exist, create the following `ClusterRoleBinding`:
 +
 ----
 $ cat <<EOF |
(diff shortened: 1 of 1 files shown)


kalexand-rh:
@zhaozhanqi, will you PTAL or suggest another reviewer?

install_config/redeploying_certificates.adoc
```diff
@@ -623,17 +623,67 @@ EOF
...
+----
+$ oc export secret router-certs > ~/old-router-certs-secret.yaml
+----
+
+. Create a new secret to hold the new certificate and key, and replace the 
+contents of the existing secret:
 +
 ----
-$ oc delete secret router-certs
+$ oc create secret generic router-certs tls.crt=router.pem tls.key=router.key \ <1>
```
zhaozhanqi: Here the command is old way but not correct,  the correct one should be like:
`oc create secret generic router-certs --from-file=tls.crt=custom-router.crt --from-file=tls.key=custom-router.key --type=kubernetes.io/tls -o json | oc replace -f -`

another new way is like:
`oc create secret tls tls-secret --cert=path/to/tls.cert --key=path/to/tls.key`

kalexand-rh: @zhaozhanqi, thanks! Do you agree that the new way should go in 3.9 and later, and the old way should go in 3.5-3.7?

zhaozhanqi: hmm,I'm not sure about that. maybe UI team can help that. 
@xingxingxia  Could you help confirm this:

2:Do you agree that the new command should go in 3.9 and later?
3:Do you agree that this old way should go in 3.5-3.7?

xingxingxia: `oc create secret tls` and `oc secrets new` both exist in 3.3 ~ 3.10. Only since 3.9, `oc secrets new` is deprecated. It is OK to use `oc create secret tls` in the doc.

install_config/redeploying_certificates.adoc
```diff
@@ -623,17 +623,67 @@ EOF
...
+----
+
+. Create a new secret to hold the new certificate and key, and replace the 
+contents of the existing secret:
 +
 ----
-$ oc delete secret router-certs
+$ oc create secret tls router-certs --cert=router.pem \ <1>
+    --key=router.key -o json | \
+    oc replace -f -
```
xingxingxia: `oc create secret tls router-certs --cert=router.pem --key=router.key -o` is enough. Need not `-o json | oc replace -f -`

xingxingxia: Given `oc delete ...` is removed, the command will fail, my result:
```
oc create secret tls ... -o json | oc replace -f -
Error from server (AlreadyExists): secrets ... already exists
```
You should keep `oc delete ...` and remove `-o json | oc replace -f -`:
```
$ oc delete secret router-certs
$ oc create secret tls router-certs ... --key=router.key
```
Or given `oc delete ...` removed, add `--dry-run`:
```
-$ oc delete secret router-certs
+$ oc create secret tls router-certs --cert=router.pem --key=router.key -o json --dry-run | oc replace -f -
```

install_config/redeploying_certificates.adoc
kalexand-rh: Thank you @xingxingxia!  The bug was against 3.5, but I'll send the change back to 3.3.

kalexand-rh:
Thank you both! I've updated the command and will merge it as soon as you confirm. :)

xingxingxia:
The oc create secret part looks good to me. As to other parts, need @zhaozhanqi finally confirm if not yet confirmed

zhaozhanqi:
others LGTM

kalexand-rh:
Thank you both!

@openshift/team-documentation, PTAL. :coffee:

install_config/redeploying_certificates.adoc
```diff
@@ -623,17 +623,67 @@ EOF
...
+    --signer-cert=/etc/origin/master/ca.crt \
+    --signer-key=/etc/origin/master/ca.key \
+    --signer-serial=/etc/origin/master/ca.serial.txt \
+    --hostnames='*.hostnames.for.the.certificate' \
+    --cert=router.crt \
+    --key=router.key \
+
+----
++
+These commands generates the following files:
```
bfallonf: s/generates/generate

Although, you could just as easily say "This generates the following files:"

install_config/redeploying_certificates.adoc
```diff
@@ -623,17 +623,67 @@ EOF
...
++
+----
+$ cat router.crt /etc/origin/master/ca.crt router.key > router.pem
+----
+
+. Back up the old certificate:
++
+----
+$ oc export secret router-certs > ~/old-router-certs-secret.yaml
+----
```
bfallonf: Is there a big reason why you're backing up the old cert? Does this come back later in the procedure? Or is it just to keep it in case. Can we give the why here?

bfallonf:
@kalexand-rh I made some comments. LG otherwise! :tada:

install_config/redeploying_certificates.adoc
kalexand-rh: +1 on the first

I was taught to always qualify "this" and "these" to avoid ambiguity.

install_config/redeploying_certificates.adoc
```diff
@@ -623,17 +623,67 @@ EOF
 oc create -f -
 ----
 +
-Then, run the following to remove the environment variables:
+Then, run the following command to remove the environment variables:
```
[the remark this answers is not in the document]
kalexand-rh: Yup, you're right. I was changing it to complete the phrase. :) If it's its own substep, it still needs to be qualified, but it can be simpler.

install_config/redeploying_certificates.adoc
kalexand-rh: It's not used later - it's just kept in case.

kalexand-rh:
Thank you, @bfallonf!

PR #9584, opened 2018-05-25
Added section to service catalog about outside routes

bfallonf:
For: https://trello.com/c/T2n6DiU2/51-3-document-network-isolation-integration-with-service-linking-related-to-service-catalog-sdndocs

tag @knobunc 

I feel like this is very much off the mark, and the last paragraph is a copy/paste of the sections that I think need to be a procedure, but let me know what you think. The information I had could be quite speculative for now and might be changing? Not sure.

### DIFF
--- a/architecture/service_catalog/index.adoc
+++ b/architecture/service_catalog/index.adoc
@@ -101,6 +101,34 @@ references to the binding secret from the deployment. Otherwise, the next
 rollout will fail.
 ====
 
+ifdef::openshift-enterprise,openshift-origin[]
+[[service-catalog-route-planning]]
+== Service catalog compatibility with services outside of the cluster
+
+[NOTE]
+====
+This section is only relevant to clusters using the `NetworkPolicy` networking
+SDN plug-in.
+====
+
+The difference in networking policies can impact your service broker's
+accessibility. You must ensure that `NetworkPolicy` objects exist in your
+service broker, so that users can have appropriate access to these resources in
+their project or namespace. These `NetworkPolicy` objects are triggered by using
+the service catalog through provisioning or binding.
+
+For example, if the service broker creates a pod inside a user's project, then
+the service broker also needs to create a `NetworkPolicy` object with rules that
+allow access to the pods. Or, if the service catalog creates objects in a
+different namespace in the cluster, then the target project may need
+`NetworkPolicy` rules to allow the source project to talk to it.
+
+See the
+xref:../../admin_guide/managing_networking.adoc#admin-guide-manage-networking[Managing
+networking section of the cluster administrator guide] for more information on
+`NetworkPolicy` objects.
+endif::openshift-enterprise,openshift-origin[]
+
 [[service-catalog-concepts-terminology]]
 == Concepts and Terminology
 


architecture/service_catalog/index.adoc
```diff
@@ -101,6 +101,24 @@ references to the binding secret from the deployment. Otherwise, the next
 rollout will fail.
 ====
 
+[[service-catalog-route-planning]]
+== Service catalog compatibility with services outside of the cluster
+
+As of {product-title} version 3.10, a user can specify a Kubernetes service
```
knobunc: I think you are thinking of the wrong thing.

The service catalog allows you to provision things in your cluster, but it can be a linkage to something outside the cluster too.  If it's outside the cluster then egress rules apply, but they can't be set by anyone other then an admin, so it might be worth mentioning as a note at the end... but not as the main lead-in.  The point of this is that the internal self-hosted services will need network policy rules to be set or people can't access them.

knobunc: Given that this is in the service catalog section, I'm not sure how much of this preamble you need.

architecture/service_catalog/index.adoc
```diff
@@ -101,6 +101,24 @@ references to the binding secret from the deployment. Otherwise, the next
...
+outside of the cluster to be consumed by other users using firewall rules. This
+is up to the user to ensure that the service is able to be consumed by adding
+rules. {product-title} ensures that the network isolation is automatically
+allowing connectivity to occur.
+
+Administrators can use `NetworkPolocy` objects to declare appropriate access to
```
knobunc: This is spot on.  But it's NetworkPolicy not Polocy.

And it only applies when the cluster has the NetworkPolicy plugin installed.

architecture/service_catalog/index.adoc
```diff
@@ -101,6 +101,24 @@ references to the binding secret from the deployment. Otherwise, the next
...
+these resources from a user's project or namespace. For example, from the
+Ansible broker or the Template broker. These `NetworkPolocy` objects are
+triggered by using the service catalog through provisioning or binding.
+
+See the <here> for more information on how to trigger the service catalog.
```
knobunc: This is more about making sure that the service catalog installer remembers to put in any needed network policies to allow access to the service.

architecture/service_catalog/index.adoc
```diff
@@ -101,6 +101,24 @@ references to the binding secret from the deployment. Otherwise, the next
...
+
+<We need to document how to make it possible to access the provisioner from projects (like we do with the router). Then we need to document that the task that is provisioning the service set up appropriate network policy rules to allow access.>
```
knobunc: I'm not sure how thorough we need to be... I think there are three cases we could enumerate:

1) The SC installs things inside the project -- Needs NetPol rules to allow access in the project
2) The SC installs things in a different namespace in the cluster -- The target may need NetPol rules to allow the source project to talk to it
3) The SC enables access to an out-of-cluster service -- Nothing in the cluster needs to be changed.

knobunc (changes requested):

bfallonf:
Thanks @knobunc . I've changed to suggestion. WDYT?

architecture/service_catalog/index.adoc
```diff
@@ -101,6 +101,34 @@ references to the binding secret from the deployment. Otherwise, the next
...
+SDN plug-in.
+====
+
+The difference in networking policies can impact your service broker's
+accessibility. You must ensure that `NetworkPolicy` objects exist in your
+service broker, so that users can have appropriate access to these resources in
+their project or namespace. These `NetworkPolicy` objects are triggered by using
+the service catalog through provisioning or binding.
+
+For example, if the service catalog need to create an object inside of a user's
```
knobunc: service broker?  The catalog doesn't create services, it is just a list of them.

"if the service broker creates a pod inside a user's project, ..."

architecture/service_catalog/index.adoc
```diff
@@ -101,6 +101,34 @@ references to the binding secret from the deployment. Otherwise, the next
...
+project, then `NetworkPolicy` rules needs to allow access to the desired
```
knobunc: ", then the service broker also needs to create a `NetworkPolicy` object with rules that allow access to the pods."

architecture/service_catalog/index.adoc
```diff
@@ -101,6 +101,34 @@ references to the binding secret from the deployment. Otherwise, the next
...
+project. Or, if the service catalog creates objects in a different name space in
```
knobunc: Or, if the service catalog creates objects in a different name space in
the cluster, then the target project may need `NetworkPolicy` rules to allow the source project to talk to it.

knobunc (changes requested):

knobunc (approved):
LGTM

PR #9599, opened 2018-05-25
Bug 1543460, add Securing Builds info to Advanced Installation topic

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1543460


ahardin-rh:
@sjwalter2 PTAL and let me know if this placement makes sense. Is there more that can be added to describe this use case? Thanks!

sjwalter2:
My only concern is that since it's a post-install step, putting it just before the "running the installer" step might be a tad misleading. I'm not sure what the best option here is, though. After all, we want it in this doc (so that people will see it earlier rather than later), but it's not ansible-ized so doesnt fit in before the installation, exactly. Maybe after the "Verifying the Install' section?

Otherwise we could include it as an item in the Configuring Ansible Inventory Files header that just says "this is not possible via ansible but heres the link etc etc"

ahardin-rh:
@sjwalter2 Thanks! Updated :+1:

PR #9632, opened 2018-05-29
Arranged third party networking vendors info

bfallonf:
Arranging third-party vendor content in the Arch guide. 

It's not complete in that the links to and from other directories haven't been fixed, but I want to show what it'll look like, then fix that stuff. This is more about the content from third parties being consistent and equal, so that can happen after this is merged.

Look at the "Available SDN plug-ins" and "Available router plug-ins" files plz.

cc @vikram-redhat 

Preview:
http://file.bne.redhat.com/bfallonf/29_05_2018/third_party_vendors/architecture/networking/networking.html


bfallonf:
@openshift/team-documentation PTAL. Please ignore any small errors and look at the structure. It's about better dividing the router and SDN content into topics, and better classifying third party vendors. Next steps are to create a template around third-party stuff to get them equal in information and layout (hope that makes sense).

kalexand-rh:
I think it looks ok structure-wise. Will you remind me why you're using some bold elements as pseudo-headings?

architecture/networking/network_plugins.adoc
```diff
@@ -50,30 +32,42 @@ Ansible-based installation procedure. See the
 xref:../../architecture/networking/sdn.adoc#architecture-additional-concepts-sdn[OpenShift SDN] section
 for more information.
 
+== Third-Party SDN plug-ins
```
ahardin-rh: I think you can just drop the OpenShift SDN section here. From the Overview, you can make "Software defined network (SDN) plug-ins" a hyperlink to the existing OpenShift SDN topic. Also, for the sake of modular docs, we should probably also rename "Overview", or drop the section altogether, combining the content with the "OpenShift SDN" section.

architecture/networking/assembly_available_router_plugins.adoc
```diff
@@ -0,0 +1,33 @@
+[[architecture-additional-concepts-router-plugins]]
+= Available router plug-ins
+{product-author}
```
ahardin-rh: I'm assuming that the Available Router Plug-ins section will be removed from the Routes topic and incorporated within this topic.

bfallonf: I was thinking about it. But now that you've commented, yes, I think it's best and will do it :)

bfallonf: Oohh I thought the image was different, but it's the same. Yes, definitely will take out.

architecture/networking/network_plugins.adoc
bfallonf: I thought it would be a good idea to highlight that the OpenShift SDN is the default, but it can probably be trimmed down to something like just the link to the section. Thanks for the comment! I'll drop the "Overview" title at the least, and make sure it's more about the topic.

bfallonf:
Thanks @ahardin-rh @kalexand-rh .

And Kathryn, I changed the headings to see what the TOC would look like without those sections. The next step is to edit down those sections anyway, so that content might not make the cut. As I said above, this is all to make the third-party sections more consistent with each other, so we'll see.

bfallonf:
Changes pushed to here, but will hold off on merging.

cc @vikram-redhat

PR #9638, opened 2018-05-29
Fix the OpenShift SDN migration steps

knobunc:
We need to restart openvswitch to clean out the rules before we
restart the node processes.

Fixes bug 1569244 (https://bugzilla.redhat.com/show_bug.cgi?id=1569244)


bfallonf:
Thanks @knobunc !

I can merge, but the BZ is for 3.7, but the target is set to 3.10. Which version should I be cherrypicking this PR to?

kalexand-rh:
@knobunc, which versions do this change apply to?

knobunc:
Whoops.  Sorry!

This should be fixed in 3.10.  It can be back-ported as far back as 3.7 if you want.

bfallonf:
Thanks @knobunc . I'll port back to 3.9. Looks like it's referring to the node service on the master, and I'm guessing that's because of the schedulability change? Let me know if I'm off the mark and it should go back to 3.7.

PR #9671, opened 2018-05-30
[enterprise 3.7] bugs 1444209 and 1531276 - describing and recovering…

kalexand-rh:
… from etcd quorum loss


kalexand-rh:
I did manual file diffs between the source files in master and 3.7. The only difference I saw was a link to the migrating topic, which is now right in this PR.

PR #9681, opened 2018-05-30
add note about using cli in other distros

juanvallejo:
See https://github.com/openshift/origin/issues/19882
and https://github.com/openshift/jenkins-client-plugin/issues/158

### DIFF
--- a/cli_reference/get_started_cli.adoc
+++ b/cli_reference/get_started_cli.adoc
@@ -188,6 +188,16 @@ Customer Portal] as a *_tar.gz_* archive. After logging in with your Red Hat
 account, you must have an active OpenShift Enterprise subscription to access the
 downloads page.
 
+[NOTE]
+===
+If you do not use RHEL or Fedora, ensure that *libc* is installed and on your library path.
+If *libc* is not available, you might see the following error when you run CLI commands:
+
+---
+oc: No such file or directory
+---
+===
+
 https://access.redhat.com/downloads/content/290[*Download the CLI from the Red
 Hat Customer Portal*]
 


gabemontero:

cli_reference/get_started_cli.adoc
```diff
@@ -188,6 +188,16 @@ Customer Portal] as a *_tar.gz_* archive. After logging in with your Red Hat
 account, you must have an active OpenShift Enterprise subscription to access the
 downloads page.
 
+[NOTE]
+===
+For other distributions of Linux, it is recommended to ensure that *libc* is included with
+your distribution. Attempting to use the CLI without libc could result in runtime errors such as:
```
gabemontero: Additionally, it may be in the distribution, but not on the libpath by default ... some sort of reminder to check your libpath would also be good here IMO

gabemontero (changes requested):

cli_reference/get_started_cli.adoc
kalexand-rh: I think it would be better to point out the distro restriction in the note itself. Maybe "If you do not use RHEL or Fedora, ensure that *libc* is installed and on your libpath. If *libc* is not available, you might see the following error when you run CLI commands:"

juanvallejo:
Thanks, comments addressed

cli_reference/get_started_cli.adoc
```diff
@@ -188,6 +188,16 @@ Customer Portal] as a *_tar.gz_* archive. After logging in with your Red Hat
 account, you must have an active OpenShift Enterprise subscription to access the
 downloads page.
 
+[NOTE]
+===
+If you do not use RHEL or Fedora, ensure that *libc* is installed and on your path.
```
gabemontero: on your path -> on your library path

gabemontero (changes requested):
One very minor comment, otherwise lgtm

cli_reference/get_started_cli.adoc
juanvallejo: Thanks, done

kalexand-rh:
@juanvallejo,how far back is this change valid? (I can send it back as far as 3.1, but I need to know if this has always been true.)

gabemontero (approved):
/lgtm

kalexand-rh:
@xingxingxia, how far back is this change valid? I think it might be 3.1.

juanvallejo:
@kalexand-rh 
> how far back is this change valid? (I can send it back as far as 3.1, but I need to know if this has always been true.)

Yes, I believe this applies to all versions

xingxingxia:
Agree

kalexand-rh:
Thank you both! I'm going to merge and cherrypick back to 3.1.

gabemontero:
Ahh ... I just noticed that the new text is in a `ifdef::openshift-enterprise[]` block.... the PR review delta did not go high enough to display that by default.  Only noticed when I tried to look for the new doc on docs.openshift.org.

A good portion of my motivation for engaging with @juanvallejo on this was I have upstream users trying to use `oc` in *atypical* linux distros like alpine or debian.

Would folks be amenable to moving the new text from @juanvallejo to the `ifdef::openshift-origin[]` block?

kalexand-rh:
It sounds like the note has value in both ifdefs and that it should be the one for openshift-origin. If one of you opens the PR, I'll merge/CP it.

gabemontero:
Created https://github.com/openshift/openshift-docs/pull/9799

PR #9685, opened 2018-05-31
added netmask info to dev-guide/traffic section

bfallonf:
For: https://bugzilla.redhat.com/show_bug.cgi?id=1553804

Will verify in the BZ.

### DIFF
--- a/dev_guide/expose_service/expose_internal_ip_service.adoc
+++ b/dev_guide/expose_service/expose_internal_ip_service.adoc
@@ -343,16 +343,16 @@ eth0: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
         ...
 ----
 
-. Add a route between the IP address of the host where the master resides and the gateway IP address of the master host:
+. Add a route between the IP address of the host where the master resides and the gateway IP address of the master host. If using a netmask for a networking route, use the `netmask` option, as well as the netmask to use:
 +
 ----
-$ route add -net <host_ip_address> netmask <netmask> gw <gateway_ip_address> dev <device>
+$ route add -host <host_ip_address> netmask <netmask> gw <gateway_ip_address> dev <device>
 ----
 +
 For example:
 +
 ----
-$ route add -host 10.16.41.22 gw 10.16.41.254 dev eth0
+$ route add -host 10.16.41.22 netmask 255.255.248.0 gw 10.16.41.254 dev eth0
 ----
 +
 The `netstat -nr` command provides the gateway IP address:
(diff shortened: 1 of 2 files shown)


dev_guide/expose_service/expose_internal_ip_service.adoc
```diff
@@ -343,16 +343,16 @@ eth0: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
...
 ----
-$ route add -net <host_ip_address> netmask <netmask> gw <gateway_ip_address> dev <device>
+$ route add -host <host_ip_address> netmask <netmask> gw <gateway_ip_address> dev <device>
 ----
 +
 For example:
 +
 ----
-$ route add -host 10.16.41.22 gw 10.16.41.254 dev eth0
+$ route add -host 10.16.41.22 netmask 255.255.255.0 gw 10.16.41.254 dev eth0
```
zhaozhanqi: line 342: netmask 255.255.248.0 
So I think here should be '255.255.248.0' instead of 255.255.255.0.

bfallonf:
Thanks @zhaozhanqi . I've updated to your suggestion.

zhaozhanqi:
LGTM

dev_guide/expose_service/expose_internal_ip_service.adoc
```diff
@@ -343,16 +343,16 @@ eth0: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
         ...
 ----
 
-. Add a route between the IP address of the host where the master resides and the gateway IP address of the master host:
+. Add a route between the IP address of the host where the master resides and the gateway IP address of the master host. If using a netmask for a networking route, use the `netmask` option, as well as the netmask to use:
```
knobunc: I'd drop this change... the other places don't bother mentioning it.

knobunc (changes requested):
This change is fine with me.

But the whole section is rather pointless.  It documents a usecase that really will never come up.  They really need to work with the network admin to get an ip address (or set of them) that can be routed to a gvien machine.

kalexand-rh (commented):
LGTM

dev_guide/expose_service/expose_internal_ip_service.adoc
bfallonf: @knobunc That's what the BZ is about, so I've kept the change, but extended it to the other examples. Let me know if you disagree.

bfallonf:
Thanks. I'll merge.

PR #9701, opened 2018-05-31
bug 1572654 - adding patch permission to clusterrole

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1572654

@stuartchuan, PTAL


kalexand-rh:
@stuartchuan approved this on the source bug. Merging.

PR #9702, opened 2018-05-31
remove openshift_set_node_ip/openshift_ip docs

sjenning:
xref https://bugzilla.redhat.com/show_bug.cgi?id=1584486

In 3.10, the use of `openshift_set_node_ip` and `openshift_ip` are no longer supported

@smarterclayton @derekwaynecarr @sdodson


ahardin-rh:
LGTM

sdodson:
/lgtm
I fear we'll hear about this being a problem for customers that they can no longer configure which interface to use. We should think about ensuring that we have some method to handle that in the future.

sjenning:
Unfamiliar with docs merge policy.  Who merges this?

ahardin-rh:
@sjenning I can :raised_hand_with_fingers_splayed: 

@vikram-redhat FYI, I don't think this needs QE review. I added it to the 3.10 release notes tracker so that we can call this change out there.

PR #9724, opened 2018-06-01
Wrong yaml snippet of the master-config.yaml file (named certificates nesting)

josgonza-rh:
Fixing wrong `servingInfo` yaml block (Incorrect nesting of the element `namedCertificates`)


ahardin-rh:
Thank you!

PR #9758, opened 2018-06-04
Bug#1583526 - added a note about double-escaping special chars

gaurav-nelson:
For https://bugzilla.redhat.com/show_bug.cgi?id=1583526

### DIFF
--- a/install_config/topics/escaping_special_characters.adoc
+++ b/install_config/topics/escaping_special_characters.adoc
@@ -1,0 +1,18 @@
+////
+Escaping special characters in ansible variables file
+
+This module included in the following assemblies:
+* install_config/configuring_openstack.adoc
+* install_config/install/advanced_install.adoc
+////
+
+[IMPORTANT]
+====
+If a parameter value in the Ansible inventory file contains special characters,
+such as `pass:[#]`, `{` or `}`, you must double-escape the value (that is
+enclose the value in both single and double quotation marks). For example, to
+use `pass:[mypasswordwith###hashsigns]` as a value for the variable
+`openshift_cloudprovider_openstack_password`, declare it as
+`openshift_cloudprovider_openstack_password='"mypasswordwith###hashsigns"'` in
+the Ansible host inventory file.
+====
--- a/install_config/configuring_openstack.adoc
+++ b/install_config/configuring_openstack.adoc
@@ -77,6 +77,8 @@ xref:../install_config/install/advanced_install.adoc#advanced-install-configurin
 - `*openshift_cloudprovider_openstack_region*`
 - `*openshift_cloudprovider_openstack_lb_subnet_id*`
 
+include::install_config/topics/escaping_special_characters.adoc[]
+
 .Example OpenStack Configuration with Ansible
 
 ----
@@ -175,7 +177,7 @@ container. Therefore, *_cloud.conf_* should be in *_/etc/origin/_* instead of
 include::install_config/topics/applying_configuration_changes.adoc[]
 
 [[openstack-configuring-zone-labels]]
-=== Configuring Zone Labels for Dynamically Created OpenStack PVs 
+=== Configuring Zone Labels for Dynamically Created OpenStack PVs
 
 Administrators can configure zone labels for dynamically
 created OpenStack PVs. This option is useful if the OpenStack Cinder zone name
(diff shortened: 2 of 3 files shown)


gaurav-nelson:
@e-minguez PTAL

I have placed the following note:

> If you are using special characters (for example, `pass:[#]`, `{` or `}`) in the
> Ansible host inventory file, you need to double-escape them to make it work. For
> example, to use `pass:[mypasswordwith###hashsigns]` as a value for the variable
> `openshift_cloudprovider_openstack_password`, declare it as
> `openshift_cloudprovider_openstack_password='"mypasswordwith###hashsigns"'` in
> the Ansible host inventory file.

in following sections:

1. advanced_install > configuring-cluster-variables (PREVIEW: https://5b14a790b3127479ad7d3062--gnelson-preview.netlify.com/openshift-enterprise/(head%20detached%20at%20fetch_head)/install_config/install/advanced_install.html#configuring-cluster-variables)
2. configuring_openstack > openstack-configuring-masters-ansible (PREVIEW: https://5b14a790b3127479ad7d3062--gnelson-preview.netlify.com/openshift-enterprise/(head%20detached%20at%20fetch_head)/install_config/configuring_openstack.html#openstack-configuring-masters-ansible)

gaurav-nelson:

install_config/topics/escaping_special_characters.adoc
```diff
@@ -0,0 +1,17 @@
...
+Escaping special characters in ansible variables file
+
+This module included in the following assemblies:
+* install_config/configuring_openstack.adoc
+* install_config/install/advanced_install.adoc
+////
+
+[IMPORTANT]
+====
+If you are using special characters (for example, `pass:[#]`, `{` or `}`) in the
```
e-minguez: The syntax for an ini inventory file is ```pass=whatever``` instead ```pass:whatever```
I'd also use ```pass=mycomplexpa##word``` instead ```pass=[#]```

install_config/topics/escaping_special_characters.adoc
```diff
@@ -0,0 +1,17 @@
...
+Ansible host inventory file, you need to double-escape them to make it work. For
+example, to use `pass:[mypasswordwith###hashsigns]` as a value for the variable
```
e-minguez: Same: ```pass=mypasswordwith###hashsigns```

e-minguez (changes requested):
I've added some comments. Thx!

install_config/topics/escaping_special_characters.adoc
kalexand-rh: I might replace the first sentence: If a parameter value in the Ansible inventory file contains special characters, such as `#`, `{` or `}`, you must enclose the value in both single and double quotation marks.

install_config/topics/escaping_special_characters.adoc
```diff
@@ -0,0 +1,17 @@
+////
```
kalexand-rh: @bfallonf, @vikram-redhat  what do you think of reusing a note like this?  Should we? Should we make a separate folder for notes?

kalexand-rh (commented):
I have a question and a suggestion. This is good information!

install_config/topics/escaping_special_characters.adoc
gaurav-nelson: My bad, the `pass:[...]` format is the asciidoc passthrough variable to render the text correctly.

The final rendered text will looks like this:
> If you are using special characters (for example, `#`, `{` or `}`) in the
> Ansible host inventory file, you need to double-escape them to make it work. For
> example, to use `mypasswordwith###hashsigns` as a value for the variable
> `openshift_cloudprovider_openstack_password`, declare it as
> `openshift_cloudprovider_openstack_password='"mypasswordwith###hashsigns"'` in
> the Ansible host inventory file.

gaurav-nelson:
Thanks @kalexand-rh I have updated the text. However, I have kept the word `double-escaped` as it is a common word which devs understand.

gaurav-nelson:
@e-minguez Can you please check the updates and confirm?
image

install_config/topics/escaping_special_characters.adoc
vikram-redhat: @kalexand-rh isn't that just another module?

e-minguez (approved):
LGTM

gaurav-nelson:
Changes already in 3.10 branch. Not cherry-picking.

PR #9865, opened 2018-06-06
Fix duplicate ids in disabling features

vikram-redhat:


vikram-redhat:
This was merged in the 3.10 branch already, putting it back in master.

cc: @mburke5678

PR #9927, opened 2018-06-08
Don't advice to modify default SCCs

php-coder:
From our experience, modifying default SCCs could lead to the issues during upgrade. While we told users that the only some fields can be safely modified, documentation had places where we editing properties of the "restricted" SCC that shouldn't not be modified as they will be reset to the default values by the `oc adm policy reconcile-sccs` command that is being executed during upgrade. 

This PR corrects documentation and advise to use other approaches instead of editing "restricted" SCC.

Related to https://bugzilla.redhat.com/show_bug.cgi?id=1577830

PTAL @pweil- @liggitt @openshift/sig-security

### DIFF
--- a/admin_guide/manage_scc.adoc
+++ b/admin_guide/manage_scc.adoc
@@ -150,7 +150,7 @@ requiredDropCapabilities:
 You can see the list of possible values in the
 link:https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities[Docker
 documentation].
-
++
 [TIP]
 ====
 Because capabilities are passed to the Docker, you can use a special `ALL` value
@@ -429,56 +429,59 @@ To provide additional capabilities:
 === Modify Cluster Default Behavior
 
 To modify your cluster so that it does not pre-allocate UIDs, allows containers
-to run as any user, and prevents privileged containers:
+to run as any user, but prevents privileged containers, grant access to the
+*anyuid* SCC for everyone:
 
-[NOTE]
-====
-In order to preserve customized SCCs during upgrades, do not edit settings on
-the default SCCs other than priority, users, groups, labels, and annotations.
-====
-
-. Edit the *restricted* SCC:
-+
 ----
- $ oc edit scc restricted
+ $ oc adm policy add-scc-to-group anyuid system:authenticated
 ----
-
-. Change `*runAsUser.Type*` to *RunAsAny*.
-
-. Ensure `*allowPrivilegedContainer*` is set to false.
-
-. Save the changes.
 
 To modify your cluster so that it does not pre-allocate UIDs and does not allow
-containers to run as root:
+containers to run as root, grant access to the *nonroot* SCC for everyone:
 
-. Edit the *restricted* SCC:
-+
 ----
- $ oc edit scc restricted
+ $ oc adm policy add-scc-to-group nonroot system:authenticated
 ----
 
-. Change `*runAsUser.Type*` to *MustRunAsNonRoot*.
+[WARNING]
+====
+Be very careful with any modifications that have a cluster-wide impact. When
+you grant an SCC to all authenticated users, as in the previous example, or
+modify an SCC that applies to all users, such as the *restricted* SCC, it also
+affects Kubernetes and {product-title} components, including the web console
+and integrated docker registry. Changes made with these SCCs can cause these
+components to stop functioning.
 
-. Save the changes.
+Always prefer creating a custom SCC and target it to only specific users or
+groups. This way potential issues are confined to the affected users or groups
+and do not impact critical cluster components.
+====
 
 [[use-the-hostpath-volume-plugin]]
 
 === Use the hostPath Volume Plug-in
 
 To relax the security in your cluster so that pods are allowed to use the
-`hostPath` volume plug-in without granting everyone access to the *privileged*
-SCC:
+`hostPath` volume plug-in without granting everyone access to more privileged
+SCCs such as *privileged*, *hostaccess*, or *hostmount-anyuid*, take the
+following actions:
 
-. Edit the *restricted* SCC:
+. xref:creating-new-security-context-constraints[Create a new SCC] named `hostpath`
+
+. Set the `*allowHostDirVolumePlugin*` parameter to `true` for the new SCC:
 +
 ----
-$ oc edit scc restricted
+$ oc patch scc hostpath -p '{"allowHostDirVolumePlugin": true}'
 ----
 
-. Add `*allowHostDirVolumePlugin: true*`.
+. Grant access to this SCC to all users:
++
+----
+$ oc adm policy add-scc-to-group hostpath system:authenticated
+----
 
-. Save the changes.
+Now, all the pods that request `hostPath` volumes are admitted by the
+`hostpath` SCC.
 
 === Ensure That Admission Attempts to Use a Specific SCC First
 
--- a/admin_guide/seccomp.adoc
+++ b/admin_guide/seccomp.adoc
@@ -102,7 +102,7 @@ The allowable formats of the *seccompProfiles* field include:
 * *unconfined*: unconfined profile, and disables seccomp
 * *localhost/<profile-name>*: the profile installed to the node's local seccomp profile root
 +
-For example, if you are using the default *docker/default* profile, configure the *restricted* SCC with:
+For example, if you are using the default *docker/default* profile, configure your SCC with:
 +
 ----
 seccompProfiles:
@@ -112,7 +112,7 @@ seccompProfiles:
 [[seccomp-configuring-openshift-with-custom-seccomp]]
 ==  Configuring {product-title} for a Custom Seccomp Profile
 
-To ensure pods in your cluster run with a custom profile in the *restricted* SCC:
+To ensure pods in your cluster run with a custom profile:
 
 . Create the seccomp profile in *seccomp-profile-root*.
 
@@ -137,7 +137,7 @@ ifdef::openshift-origin[]
 ----
 endif::[]
 
-. Configure the *restricted* SCC:
+. Configure your SCC:
 +
 ----
 seccompProfiles:


admin_guide/manage_scc.adoc
```diff
@@ -429,56 +429,44 @@ To provide additional capabilities:
...
-. Save the changes.
-
 [[use-the-hostpath-volume-plugin]]
 
 === Use the hostPath Volume Plug-in
 
 To relax the security in your cluster so that pods are allowed to use the
-`hostPath` volume plug-in without granting everyone access to the *privileged*
-SCC:
+`hostPath` volume plug-in without granting everyone access to the too much
```
pweil-: without granting everyone access to more privileged SCCs such as privileged, hostaccess, or hostmount-anyuid.

pweil- (commented):
LGTM with a minor wording change.  Thanks for updating it @php-coder

admin_guide/manage_scc.adoc
```diff
@@ -429,56 +429,44 @@ To provide additional capabilities:
...
-====
-In order to preserve customized SCCs during upgrades, do not edit settings on
-the default SCCs other than priority, users, groups, labels, and annotations.
-====
-
-. Edit the *restricted* SCC:
-+
 ----
- $ oc edit scc restricted
+ $ oc adm policy add-scc-to-group anyuid system:authenticated
```
php-coder: This is not identical to editing `restricted` for the following reasons:
- `anyuid` has priority and will win other SCCs most of the time
- `anyuid` doesn't drop some UID-related capabilities that `restricted` drops
- `anyuid` also allows any `fsGroups`

I agree that granting `anyuid` to everyone could also lead to unexpected results and even break the cluster components in some cases (see https://github.com/openshift/openshift-ansible/pull/8558). But here, I think that it less painful than silent reset to the defaults during upgrade.

Ideally, I'd rather remove this chapter at all, because granting access for `system:authenticated` to any SCC may be a breaking change and IMHO it's a bad practice. Cluster admins should always create custom SCCs and grant access to them only to users/groups which are really need this access.

I decided to keep this chapter because it's about "how to go fast and ignore security best practices", so it's all big hack for someone who doesn't care/doesn't have time to fully understand/just making it work.

admin_guide/manage_scc.adoc
```diff
@@ -429,56 +429,44 @@ To provide additional capabilities:
...
-
 To modify your cluster so that it does not pre-allocate UIDs and does not allow
-containers to run as root:
+containers to run as root, grant access to the *nonroot* SCC for everyone:
 
-. Edit the *restricted* SCC:
-+
 ----
- $ oc edit scc restricted
+ $ oc adm policy add-scc-to-group nonroot system:authenticated
```
php-coder: This is also not identical change, because `nonroot` allows any fsgroups. Also there was a case, when using `nonroot` SCC broke integrated registry. So, It's hard to advise such thing but it the least of evils.

admin_guide/manage_scc.adoc
php-coder: @pweil- Thank you, Paul! You comment was addressed.

admin_guide/manage_scc.adoc
pweil-: add a NOTE section about this?

bfallonf:
Thanks @php-coder . My original PR you link to was a simple solution.

Should we be waiting for someone from the security team to comment before we merge? Your comments about this section seem pretty concerning. Also, my original addition (The important box in the overview) is still there. Should that be deleted?

kalexand-rh:
@enj, @simo5, PTAL

php-coder:
> Also, my original addition (The important box in the overview) is still there. Should that be deleted?

I'd say that it should be modified at least to become a less strict.

admin_guide/manage_scc.adoc
php-coder: @pweil- I've added a WARNING section instead.

PTAL if you find a minute for me :)

admin_guide/manage_scc.adoc
```diff
@@ -429,56 +429,56 @@ To provide additional capabilities:
...
 ----
 
-. Change `*runAsUser.Type*` to *MustRunAsNonRoot*.
-
-. Save the changes.
+[WARNING]
+====
+Be very careful with any modifications that have a cluster-wide impact! When
+SCC is being granted to all authenticated users, as in the example above, it
```
simo5: SCC is being -> an SCC is
(?)

admin_guide/manage_scc.adoc
```diff
@@ -429,56 +429,56 @@ To provide additional capabilities:
...
+will also affect Kubernetes and {product-title} components (a web console, an
+integrated docker registry, etc.) which could lead to their inoperability! The
+same considerations should be taking into account when such SCC is being
```
simo5: taking -> taken

admin_guide/manage_scc.adoc
```diff
@@ -429,56 +429,56 @@ To provide additional capabilities:
...
+modified (for example, the *restricted* SCC is accessible by everyone). Always
```
simo5: accessible by -> granted to ?

admin_guide/manage_scc.adoc
```diff
@@ -429,56 +429,56 @@ To provide additional capabilities:
...
+prefer creating a custom SCC and target it only to specific users or groups. In
+such a case, the impact of the possible breakage will be reduced to a minimum.
```
simo5: In such a case ... -> This way potential issues will be confined to the affected users/groups and will not impact critical cluster components.

simo5 (changes requested):
Minor language nits

admin_guide/manage_scc.adoc
```diff
@@ -429,56 +429,56 @@ To provide additional capabilities:
...
 ----
 
-. Change `*runAsUser.Type*` to *MustRunAsNonRoot*.
-
-. Save the changes.
+[WARNING]
+====
+Be very careful with any modifications that have a cluster-wide impact! When
+SCC is being granted to all authenticated users, as in the example above, it
+will also affect Kubernetes and {product-title} components (a web console, an
```
pweil-: remove 'a' and 'an' from the component list

admin_guide/manage_scc.adoc
pweil-: when such SCC -> when an SCC

pweil- (commented):
LGTM with minor nits.  Thanks!

php-coder:
@simo5 @pweil- Thank you for the review! I've fixed everything that you have pointed out.

admin_guide/manage_scc.adoc
```diff
@@ -429,56 +429,57 @@ To provide additional capabilities:
...
- $ oc edit scc restricted
+ $ oc adm policy add-scc-to-group nonroot system:authenticated
 ----
 
-. Change `*runAsUser.Type*` to *MustRunAsNonRoot*.
-
-. Save the changes.
+[WARNING]
+====
+Be very careful with any modifications that have a cluster-wide impact! When an
```
kalexand-rh: @php-coder, if this is still accurate, I'd use this text instead:

Be very careful with any modifications that have a cluster-wide impact. When you
grant an SCC to all authenticated users, as in the previous example, it also
affects Kubernetes and {product-title} components, including web console and
the integrated docker registry. Changes made with this SCC can cause these
components to stop functioning.

Take the same considerations into account when you modify an SCC, such as
granting the *restricted* SCC to everyone. Always prefer creating a custom
SCC and target it to only specific users or groups. This way potential issues
are confined to the affected users or groups and do not impact critical cluster
components.

kalexand-rh: Also, please take a look at the note at ~37 and make sure you're still happy with it.

admin_guide/manage_scc.adoc
```diff
@@ -429,56 +429,57 @@ To provide additional capabilities:
...
 
 [[use-the-hostpath-volume-plugin]]
 
 === Use the hostPath Volume Plug-in
 
 To relax the security in your cluster so that pods are allowed to use the
-`hostPath` volume plug-in without granting everyone access to the *privileged*
-SCC:
+`hostPath` volume plug-in without granting everyone access to more privileged
+SCCs such as *privileged*, *hostaccess*, or *hostmount-anyuid*.
```
kalexand-rh: s/*hostmount-anyuid*./*hostmount-anyuid*, take the following actions:

admin_guide/manage_scc.adoc
```diff
@@ -429,56 +429,57 @@ To provide additional capabilities:
...
 
-. Edit the *restricted* SCC:
+. Create a new SCC (with name "hostpath", for example)
```
kalexand-rh: s/Create a new SCC (with name "hostpath", for example)/Create a new SCC:

Then please add the command to create the SCC.

admin_guide/manage_scc.adoc
```diff
@@ -429,56 +429,57 @@ To provide additional capabilities:
...
+
+. Set `*allowHostDirVolumePlugin*` to `true`
```
kalexand-rh: s/Set `*allowHostDirVolumePlugin*` to `true`/Set the `*allowHostDirVolumePlugin*` parameter to `true` for the new SCC:

admin_guide/manage_scc.adoc
```diff
@@ -429,56 +429,57 @@ To provide additional capabilities:
...
 +
 ----
-$ oc edit scc restricted
+$ oc patch scc hostpath -p '{"allowHostDirVolumePlugin": true}'
 ----
 
-. Add `*allowHostDirVolumePlugin: true*`.
+. Grant access to this SCC to everyone:
```
kalexand-rh: s/everyone/all users

admin_guide/manage_scc.adoc
```diff
@@ -429,56 +429,57 @@ To provide additional capabilities:
...
++
+----
+$ oc adm policy add-scc-to-group hostpath system:authenticated
+----
 
-. Save the changes.
+Now, all the pods that request `hostPath` volumes, should be admitted by the
```
kalexand-rh: s/volumes, should be/volumes are

kalexand-rh (commented):
@php-coder, I have a few text edits and requests.

admin_guide/manage_scc.adoc
php-coder: I'd prefer to add a link to the chapter about SCC creation rather than providing this command again. WDYT?

kalexand-rh: Unless you expect people to be very familiar with creating an SCC, it's better to present them in the flow so they don't have to jump out. A link will work, though.

admin_guide/manage_scc.adoc
php-coder: > take a look at the note at ~37

I couldn't find this note. Is it a line number?

> Take the same considerations into account when you modify an SCC, such as
granting the restricted SCC to everyone.

This is not what I had in mind initially:

>> The same considerations should be taken into account when an SCC is being modified (for example, the *restricted* SCC is granted to everyone).

I've tried to say that when you're changing `restricted` SCC you have to think twice because this SCC is already granted to all users by default.

@kalexand-rh How I can express this idea then?

kalexand-rh: @php-coder, you're right. The note's at line 33 now:

> Do not modify the default SCCs. Customizing the default SCCs can lead to issues
when upgrading. Instead, xref:creating-new-security-context-constraints[create
new SCCs].

 Ok. How's this?
> Be very careful with any modifications that have a cluster-wide impact. When you
grant an SCC to all authenticated users, as in the previous example, or modify an
SCC that applies to all users, such as the _restricted_ SCC, it also affects
Kubernetes and {product-title} components, including the web console and
integrated docker registry. Changes made with this SCC can cause these
components to stop functioning.

> Always prefer creating a custom SCC and target it to only specific users or
groups. This way potential issues are confined to the affected users or groups
and do not impact critical cluster components.

php-coder:
> you're right. The note's at line 33 now:

@kalexand-rh It's out of the scope of this PR at least. All others comments were addressed. Thank you and PTAL!

admin_guide/manage_scc.adoc
```diff
@@ -429,56 +429,59 @@ To provide additional capabilities:
...
 === Use the hostPath Volume Plug-in
 
 To relax the security in your cluster so that pods are allowed to use the
-`hostPath` volume plug-in without granting everyone access to the *privileged*
-SCC:
+`hostPath` volume plug-in without granting everyone access to more privileged
+SCCs such as *privileged*, *hostaccess*, or *hostmount-anyuid*, take the
+following actions:
+
+. xref:creating-new-security-context-constraints[Create a new SCCs] (with name "hostpath", for example)
```
kalexand-rh: I have two more requests. :)

 s/Create a new SCCs] (with name "hostpath", for example)/[Create a new SCC] named `hostpath`:

Then squash your commits. Thanks!

kalexand-rh:
@barleyer, will you PTAL or suggest another reviewer from QE?

admin_guide/manage_scc.adoc
php-coder: > I have two more requests. :)

No worries; both just have been pushed. Thanks!

admin_guide/manage_scc.adoc
```diff
@@ -429,56 +429,59 @@ To provide additional capabilities:
 === Modify Cluster Default Behavior
 
 To modify your cluster so that it does not pre-allocate UIDs, allows containers
-to run as any user, and prevents privileged containers:
+to run as any user, but prevents privileged containers, grant access to the
+*anyuid* SCC for everyone:
```
mburke5678: @kalexand-rh This feels like a comma splice (before your changes, too!). Not sure how to fix.
"You can configure your cluster to not pre-allocate UIDs, which (?) allows containers to run as any user, but prevents privileged containers.  To configure this, grant access to the *anyuid* SCC for all users."
Something like that to offset  _grant access_ (above it almost sounds like granting access to everyone is another result of preventing pre-allocation.

admin_guide/manage_scc.adoc
```diff
@@ -429,56 +429,59 @@ To provide additional capabilities:
...
+you grant an SCC to all authenticated users, as in the previous example, or
+modify an SCC that applies to all users, such as the *restricted* SCC, it also
+affects Kubernetes and {product-title} components, including the web console
+and integrated docker registry. Changes made with these SCCs can cause these
+components to stop functioning.
+
+Always prefer creating a custom SCC and target it to only specific users or
+groups. This way potential issues are confined to the affected users or groups
+and do not impact critical cluster components.
+====
```
mburke5678: s/Always prefer/Instead or Alternatively

admin_guide/manage_scc.adoc
```diff
@@ -429,56 +429,59 @@ To provide additional capabilities:
...
 
 === Use the hostPath Volume Plug-in
 
 To relax the security in your cluster so that pods are allowed to use the
-`hostPath` volume plug-in without granting everyone access to the *privileged*
-SCC:
+`hostPath` volume plug-in without granting everyone access to more privileged
+SCCs such as *privileged*, *hostaccess*, or *hostmount-anyuid*, take the
+following actions:
+
```
mburke5678: s/take the following actions/perform the following ?

mburke5678:
@kalexand-rh Couple of comments. Otherwise LGTM.

admin_guide/manage_scc.adoc
kalexand-rh: I don't think it's a comma splice, but the initial clause is way overloaded. I think it's trying to say, "To modify your cluster so that *all the things*, grant access to the *anyuid* SCC for everyone."

Maybe it should say:
```
When you grant access to the *anyuid* SCC for everyone, your cluster:
* Does not pre-allocate UIDs
* Allows containers to run as any user
* Prevents privileged containers
```

admin_guide/manage_scc.adoc
kalexand-rh: I'd s/Always prefer creating/It is recommended to create

admin_guide/manage_scc.adoc
kalexand-rh: maybe "perform the following actions"
The style guide talks about "Ambiguous pronoun references," and I think that making the object explicit improves clarity and follows the spirit of that rule. (It also makes it easier to translate, if we add more languages later.)

admin_guide/manage_scc.adoc
mburke5678: @kalexand-rh That looks better!

admin_guide/manage_scc.adoc
mburke5678: That seems to be a common construct in our docs.

bfallonf: "It is" is nearing expletive construction, and no need to say what's recommended when you can just say it. Maybe: 

"Instead, create a custom SCC and target specific users or groups so that potential issues are confined to those affected and do not impact critical cluster components."

barleyer:
LGTM,besides I remember we have a card `Allow SCC Access via RBAC`，thanks.

kalexand-rh:
@php-coder, it looks like this change might be valid back to 3.3. Does that sound right to you?

admin_guide/manage_scc.adoc
kalexand-rh: @bfallonf, "it is recommended" is the preferred upstream/downstream text in [the glossary.](https://doc-stage.usersys.redhat.com/documentation/en-us/ccs_internal_documentation/1.0/html-single/glossary_of_terms_and_conventions_for_product_documentation/#recommend)

kalexand-rh:
Merging. I'll add the final text edits in a new PR.

kalexand-rh:
Slava confirmed via email that this should go back to 3.3. I'll merge to master and then CP back to 3.3 on a PR that includes the last text edits.

php-coder:
@kalexand-rh Thank you!

kalexand-rh:
3.5: https://github.com/openshift/openshift-docs/pull/10168
3.4: https://github.com/openshift/openshift-docs/pull/10169
3.3: https://github.com/openshift/openshift-docs/pull/10170

PR #9989, opened 2018-06-11
bug 1500224 porting custom wildcard cert to 3.6

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1500224

Per @bmeng, I'm adding the wild cert instructions from 3.7 to 3.6.

### DIFF
--- a/install_config/certificate_customization.adoc
+++ b/install_config/certificate_customization.adoc
@@ -103,6 +103,63 @@ the server to pick up the configuration changes.
 
 For the master API or web console, wildcard names are accepted.
 
+
+[[configuring-custom-certificates-wildcard]]
+== Configuring a Custom Wildcard Certificate for the Default Router
+
+You can configure the {product-title} default router with a default wildcard certificate. A default wildcard certificate provides a convenient way for 
+applications that are deployed in {product-title} to use default encryption without needing custom certificates. 
+
+[NOTE]
+====
+Default wildcard certificates are recommended for non-production environments only.
+====
+
+In order to configure a default wildcard certificate, provision a certificate that is valid for `.<app_domain>`, where `<app_domain>` is the value of `openshift_master_default_subdomain` in the xref:../install_config/install/advanced_install.adoc#configuring-ansible[Ansible inventory file], by default `/etc/ansible/hosts`. Once provisioned, place the certificate, key, and ca certificate files on your Ansible host, and add the following line to your Ansible inventory file.
+
+----
+openshift_hosted_router_certificate={"certfile": "/path/to/apps.c1-ocp.myorg.com.crt", "keyfile": "/path/to/apps.c1-ocp.myorg.com.key", "cafile": "/path/to/apps.c1-ocp.myorg.com.ca.crt"}
+----
+
+For example:
+
+----
+openshift_hosted_router_certificate={"certfile": "/home/cloud-user/star-apps.148.251.233.173.nip.io.cert.pem", "keyfile": "/home/cloud-user/star-apps.148.251.233.173.nip.io.key.pem", "cafile": "/home/cloud-user/ca-chain.cert.pem"}
+----
+
+Where the parameter values are: 
+
+* *certfile* is the path to the file that contains the {product-title} router certificate.
+
+* *keyfile* is the path to the file that contains the {product-title} wildcard key.
+
+* *cafile* is the path to the file that contains the root CA for this key and certificate. If an intermediate CA is in use, the file should contain both the intermediate and root CA.
+
+If these certificate files are new to your {product-title} cluster, run the Ansible *_byo/config.yml_* playbook to add these files to the {product-title} configuration files. 
+The playbook adds the certificate files to the *_/etc/origin/master/_* directory.
+
+----
+ifdef::openshift-enterprise[]
+# ansible-playbook [-i /path/to/inventory] \
+    /usr/share/ansible/openshift-ansible/playbooks/byo/config.yml
+endif::[]
+ifdef::openshift-origin[]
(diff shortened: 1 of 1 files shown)


install_config/certificate_customization.adoc
```diff
@@ -103,6 +103,62 @@ the server to pick up the configuration changes.
...
+
+Where the parameter values are: 
+
+* *certfile* is the path to the file that contains the {product-title} router certificate.
+
+* *keyfile* is the path to the file that contains the {product-title} wildcard key.
+
+* *cafile* is the path to the file that contains the root CA for this key and certificate. If an intermediate CA is in use, the file should contain both the intermediate and root CA.
+
+If these certificate files are new to your {product-title} cluster, run the Ansible *_deploy_cluster.yml_* playbook to add these files to the {product-title} configuration files.
```
jianlinliu: I think if only want to add the customized router cert files for an existing cluster, no need to run "deploy_cluster.yml", "playbooks/openshift-hosted/deploy_router.yml" playbook is enough.

install_config/certificate_customization.adoc
```diff
@@ -103,6 +103,62 @@ the server to pick up the configuration changes.
...
+Where the parameter values are: 
+
+* *certfile* is the path to the file that contains the {product-title} router certificate.
+
+* *keyfile* is the path to the file that contains the {product-title} wildcard key.
+
+* *cafile* is the path to the file that contains the root CA for this key and certificate. If an intermediate CA is in use, the file should contain both the intermediate and root CA.
+
+If these certificate files are new to your {product-title} cluster, run the Ansible *_deploy_cluster.yml_* playbook to add these files to the {product-title} configuration files. 
+The playbook adds the certificate files to the *_/etc/origin/master/named_certificates/_* directory.
```
jianlinliu: The specified customized cert files will be uploaded to "/etc/origin/master/", not "/etc/origin/master/named_certificates/".

install_config/certificate_customization.adoc
jianlinliu: BTW, the playbook path is different per version, the above path is only for 3.10.

kalexand-rh: @jianlinliu, does running "deploy_cluster.yml" apply to all versions or just 3.6?

What's the right path for other versions?

install_config/certificate_customization.adoc
kalexand-rh: Is this true in all versions or just 3.6?

install_config/certificate_customization.adoc
jianlinliu: "deploy_cluster.yml" only apply to 3.9 and 3.10; 3.6 and 3.7 are using "playbooks/byo/config.yml"

install_config/certificate_customization.adoc
jianlinliu: I think it is true in all version.

kalexand-rh: Ok. I'm making a new PR to fix this path in 3.9 and 3.10: https://github.com/openshift/openshift-docs/pull/10076

install_config/certificate_customization.adoc
kalexand-rh: I'm a little confused. I've changed the reference to `deploy_cluster.yml` to `/byo/config.yml` in 3.6 and 3.7. Please let me know if you want me to change all the versions to use `playbooks/openshift-hosted/deploy_router.yml` instead.

jianlinliu: If you prefer to use the overall entry playbook, its path should be:
3.9 ~ 3.10: playbooks/deploy_cluster.yml
3.6 ~ 3.7:  playbooks/byo/config.yml
Since 3.9, we introduce standalone plabybook, no need to re-run the overall entry playbook, so for 3.9 ~ 3.10, it is enough to run the standalone playbook - "playbooks/openshift-hosted/deploy_router.yml".
BTW, seen from the PR, I do not know what chnage would applied per different of version, so maybe though you follow my suggest, but I did not get it.

kalexand-rh: Thank you so much!
- I opened a new PR to change 3.9 and 3.10 to use deploy_router.yml here: https://github.com/openshift/openshift-docs/pull/10101
- I fixed this PR for 3.6 and 3.7 to use byo/config.yml.

Is this all correct now?

install_config/certificate_customization.adoc
```diff
@@ -103,6 +103,63 @@ the server to pick up the configuration changes.
...
+
+If xref:../install_config/redeploying_certificates.adoc#redeploying-all-certificates-current-ca[the certificates are not new], 
+for example, you want to change existing certificates or replace expired certificates, run the following playbook:
+
+----
+ansible-playbook /usr/share/ansible/openshift-ansible/playbooks/redeploy-certificates.yml
+----
+
+[NOTE] 
+For this playbook to run, the certificate names must not change. If the certificate names change, rerun the Ansible *_deploy_cluster.yml_* playbook
```
jianlinliu: If this is for 3.6 and 3.7, the playbook should be "byo/config.yml".

kalexand-rh: Thank you! I've fixed the file.

kalexand-rh:
Jianlin approved this change on the source bug. I ran a diff on this section and 3.7, and, except for correcting the playbook paths, the content is identical. I'm going to merge it.

kalexand-rh:
The bot didn't make the change I was expecting, so I closed the PR and opened a new one for 3.7: https://github.com/openshift/openshift-docs/pull/10311

PR #9990, opened 2018-06-11
bug 1500224 listing router certificate parameters

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1500224

@jianlinliu, PTAL

### DIFF
--- a/install_config/install/advanced_install.adoc
+++ b/install_config/install/advanced_install.adoc
@@ -270,6 +270,11 @@ use it.
 .2+.^|These variables are used to configure xref:../../install_config/certificate_customization.adoc#install-config-certificate-customization[custom certificates] which are deployed as part of the installation. See xref:advanced-install-custom-certificates[Configuring Custom Certificates] for more information.
 |`openshift_master_overwrite_named_certificates`
 
+|`openshift_hosted_router_certificate`
+|Provide the location of the 
+xref:../../install_config/certificate_customization.adoc#install-config-certificate-customization[custom certificates]
+for the hosted router.
+
 |`openshift_hosted_registry_cert_expire_days`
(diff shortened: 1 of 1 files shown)


install_config/install/advanced_install.adoc
```diff
@@ -270,6 +270,14 @@ use it.
 .2+.^|These variables are used to configure xref:../../install_config/certificate_customization.adoc#install-config-certificate-customization[custom certificates] which are deployed as part of the installation. See xref:advanced-install-custom-certificates[Configuring Custom Certificates] for more information.
 |`openshift_master_overwrite_named_certificates`
 
+|`openshift_hosted_router_create_certificate`
+|To provide a custom router certificate, set this parameter value to `False`.
```
jianlinliu: According to present code, when providing a custom router cert, not have to set its value to `False`.

kalexand-rh: @jianlinliu, do you ever have to change this parameter value?

kalexand-rh: @jianlinliu, if you never have to change the parameter, do we still need to describe it in this topic?

jianlinliu: At least according to my testing, I never set `openshift_hosted_router_create_certificate` in inventory file, customized router cert files are uploaded correctly. Maybe in some period, there is some existing installer bug, which force user to have to set this parameter to `False`, but now it is not a must. I think we do not need to describe it in this topic.

kalexand-rh:
@openshift/team-documentation, PTAL at this tiny change

bfallonf:
@kalexand-rh LGTM :deciduous_tree:

PR #10023, opened 2018-06-12
Issue10022 - Missing git extension in clone url - file basic_walkthro…

scoulomb:
…ugh.adoc

### DIFF
--- a/getting_started/basic_walkthrough.adoc
+++ b/getting_started/basic_walkthrough.adoc
@@ -47,7 +47,7 @@ image::gs-fork.png[Fork the project]
 application and change to the new directory:
 +
 ----
-$ git clone https://github.com/<your_github_username>/nodejs-ex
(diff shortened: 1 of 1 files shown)


php-coder:
@scoulomb It's not clear why we need this change? It works well without this extension:
```console
$ git clone https://github.com/openshift/nodejs-ex
Cloning into 'nodejs-ex'...
remote: Counting objects: 584, done.
remote: Total 584 (delta 0), reused 0 (delta 0), pack-reused 584
Receiving objects: 100% (584/584), 189.04 KiB | 0 bytes/s, done.
Resolving deltas: 100% (230/230), done.
Checking connectivity... done.
```
In other words, there is no bug that we're trying to fix.

php-coder:
P.S. Even if this change is merged, I don't see a reason for backporting it.

kalexand-rh:
If it's not necessary, then I'll leave it as-is. Thanks @php-coder!

PR #10035, opened 2018-06-12
Slight clarification to multitenant-to-networkpolicy migration doc

danwinship:
If someone follows a link directly to "Migrating from ovs-multitenant to ovs-networkpolicy", they might not notice that it's a subsection of a larger "Migrating Between SDN Plug-ins" section that they also need to read the rest of. This just adds a quick reference to the previous section there. Feel free to rewrite.


bfallonf:
Thanks @danwinship . Good eye noticing this. I can merge this and do a quick followup.

PR #10061, opened 2018-06-12
Bug 1562210 Remove HOSA

bmcelvee:
https://bugzilla.redhat.com/show_bug.cgi?id=1562210


vikram-redhat:
@bmcelvee - John Sanda said in the BZ on 14/Jun that this could go in 3.10. Is that still the case? Anything holding this up?

bmcelvee:
This is just removing content, so it doesn't need peer review.

PR #7770, opened 2018-02-13
Fix incorrect example for downward API

sammarland-rh:
Looking at the upstream Kubernetes docs they use `downwardAPI:`instead of `metadata`.

PR #10130, opened 2018-06-15
Fix incorrect example for downward API

kalexand-rh:
Looking at the upstream Kubernetes docs they use `downwardAPI:`instead of `metadata`.

Updating yaml sample per https://github.com/openshift/openshift-docs/pull/7770

@wjiangjay, how far back is this change valid? I see this yaml file going back to 3.0.

### DIFF
--- a/dev_guide/downward_api.adoc
+++ b/dev_guide/downward_api.adoc
@@ -144,7 +144,7 @@ configure this volume. The plug-in supports the following fields:
 spec:
   volumes:
     - name: podinfo
-      metadata: <1>
+      downwardAPI:: <1>
         items: <2>
           -name: "labels" <3>
            fieldRef:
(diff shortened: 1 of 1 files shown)


kalexand-rh:
@wjiangjay, how far back is this change valid?

wjiangjay:
@mdshuai

mdshuai:
3.0 is metadata, downwardapi first get in 3.1; then in 3.2 [metadata is deprecated: Use downwardAPI instead.](https://github.com/openshift/ose/blob/enterprise-3.2/Godeps/_workspace/src/k8s.io/kubernetes/pkg/api/v1/types.go#L267-L269)

kalexand-rh:
Thank you both! I'll send the change back to 3.2, per your guidance.

kalexand-rh:
QE confirmed this one going back to 3.6.

PR #10138, opened 2018-06-17
[enterprise-3.10] New Travis Build

vikram-redhat:


vikram-redhat:
Merging this as the errors are to do with invalid xrefs, which is what this new build is supposed to find. I will fix the invalid xrefs in another PR.

vikram-redhat:
Follow up: https://github.com/openshift/openshift-docs/pull/10160

PR #10189, opened 2018-06-18
Update docs for controller-manager

gnufied:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1574397

Should apply to both 3.9 and 3.10


gnufied:
cc @tmorriso-rh

PR #10197, opened 2018-06-18
breaking a link to see what happens

kalexand-rh:
@vikram-redhat, I broke a link for you. ;)

### DIFF
--- a/admin_guide/kuryr.adoc
+++ b/admin_guide/kuryr.adoc
@@ -12,7 +12,7 @@
 toc::[]
 
 == Overview
-xref:../../install_config/configuring_kuryrsdn.adoc#install-config-configuring-kuryr-sdn[Kuryr]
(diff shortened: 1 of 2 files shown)


vikram-redhat:
Thanks @kalexand-rh - so the first commit didn't actually fail the build? The second one did, right?

kalexand-rh:
The first one passed, so I added a different sort of broken link. I didn't
hang around to see if it failed, but I assumed it would. :)

vikram-redhat:
@kalexand-rh The first commit didn't trigger an error because that file is not included in any enterprise products :) https://github.com/openshift/openshift-docs/blob/master/_topic_map.yml#L660.

The build doesn't check for origin specific docs because it is targeted for publishing to the customer portal.

Having said that, even if it was included, it would only have triggered a warning, not an error. I am pushing through an update today that should fix that.

kalexand-rh:
Got it! I'll be sure to rebase to pick up your change. :)

PR #10226, opened 2018-06-19
bug 1533347 clarifying time zone format

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1533347

@anpingli, will you PTAL or suggest a different reviewer for this change? Will you please also confirm that this change is valid back to 3.5? (I assume it will be.)


kalexand-rh:
@anpingli, will you PTAL or suggest a different reviewer for this change?

kalexand-rh:
@openshift/team-documentation  PTAL

mburke5678:
@kalexand-rh LGTM

PR #10231, opened 2018-06-19
bug 1511889 fixing ansible logging parameters

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1511889

@anpingli, will you PTAL or suggest a different reviewer?


kalexand-rh:
Anping approved this change in the source bug, and since the change is limited to codeblock updates, I'm going to merge.

PR #10254, opened 2018-06-20
Updated build.py

vikram-redhat:


vikram-redhat:
Merging as the build is timing out. Simple update to the build.py file.

PR #10291, opened 2018-06-20
fixing single-master restore command

kalexand-rh:
@jiajliu, when I confirmed the build of the [original PR](https://github.com/openshift/openshift-docs/pull/10273) for https://bugzilla.redhat.com/show_bug.cgi?id=1567530, I noticed that this command was different for single master versions in 3.3-3.5.

Will you PTAL?


kalexand-rh:
Jia approved this change on the source bug, and since the text change is to a codeblock, I don't think it needs peer review. Merging.

PR #10317, opened 2018-06-21
Bug 1591999 Note - disable Docker image check for disconnected installs

bmcelvee:
https://bugzilla.redhat.com/show_bug.cgi?id=1591999

### DIFF
--- a/install/disconnected_install.adoc
+++ b/install/disconnected_install.adoc
@@ -176,6 +176,14 @@ do
 done
 ----
 
+[NOTE]
+====
+If you are syncing images on each of the hosts, rather than storing them in a
+local repository to pull from, you must disable the Docker image check or
+Skopeo will attempt to pull images even though they are locally cached and
+available.
+====
+
 [[disconnected-syncing-images]]
 === Syncing Images
 


bmcelvee:
@openshift/team-documentation PTAL :smile_cat:

install/disconnected_install.adoc
```diff
@@ -176,6 +176,14 @@ do
 done
 ----
 
+[NOTE]
+====
+If you are syncing images on each of the hosts, rather than storing them in a
+local repository to pull from, you must disable the Docker image check or
+Skopeo will attempt to pull images even though they are locally cached and
```
kalexand-rh: What's Skopeo? Is it important? (I don't see that term anywhere else in the docs.)
How do I disable the check?
I think this info needs to be a step or prereq in the appropriate section. (Does it go in the next section instead of this one?)

bmcelvee: It's an Atomic utility, and honestly I'm on the fence with this one. It was requested as a note here specifically. It's an edge case, so it might not be necessary to document at all.

kalexand-rh: Interesting! If we don't have this flow described somewhere, +1 to not documenting the edge case.

vikram-redhat:
@bmcelvee - the bug was CLOSED as NOTABUG. Do we need to close this PR?

PR #10321, opened 2018-06-21
Bug 1540753 Edited out dnsName value for egress docs

bfallonf:
For: https://bugzilla.redhat.com/show_bug.cgi?id=1540753

Will verify in the BZ.

### DIFF
--- a/admin_guide/managing_networking.adoc
+++ b/admin_guide/managing_networking.adoc
@@ -163,13 +163,13 @@ restrictions on where `EgressNetworkPolicy` can be created:
 * The `default` project (and any other project that has been made global via
 `oc adm pod-network make-projects-global`) cannot have egress policy.
 
-* If you merge two projects together (via `oc adm pod-network join-projects`),
+* If you merge two projects together using `oc adm pod-network join-projects`,
 then you cannot use egress policy in _any_ of the joined projects.
 
-* No project may have more than one egress policy object.
+* No project can have more than one egress policy object.
 
 Violating any of these restrictions results in broken egress policy for the
-project, and may cause all external network traffic to be dropped.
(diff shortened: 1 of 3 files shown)


bmeng:
The description below the json file also contains the dnsName, it should also be removed. 

When the example above is added in a project, it allows traffic to IP range
 `1.2.3.0/24` **and domain name `www.foo.com`**, but denies access to all other
 external IP addresses. (Traffic to other pods is not affected because the policy
only applies to _external_ traffic.)

For the question in bz, just delete the dnsName is ok, do not need to add second cidrSeletor.

admin_guide/managing_networking.adoc
```diff
@@ -215,20 +209,16 @@ To configure egress policy:
...
-all traffic.
+that matches takes effect. If the two rules in the above example were reversed,
+then traffic would not be allowed to `1.2.3.0/24` because the `0.0.0.0/0` rule
+would be checked first, and it would match and deny all traffic.
 +
-Domain name updates are reflected within 30 minutes. In the above example,
-suppose `www.foo.com` resolved to `10.11.12.13`, but later it was changed to
-`20.21.22.23`. Then, {product-title} will take up to 30 minutes to adapt to
-these DNS updates.
+Domain name updates are reflected within 30 minutes. This includes if the values change.
```
gaurav-nelson: How about:
 - `Domain name updates (including IP address change) are reflected within 30 minutes.` OR 
 - `Domain names and IP address changes are reflected within 30 minutes.`

gaurav-nelson:
LGTM ☯️ 
 just one suggestion though.

bfallonf:
Thanks @bmeng I've updated to include your suggestion, and included another edit from another BZ (#10335 ) . I'll merge.

PR #10342, opened 2018-06-22
Bug 1590734 CNS required for rhgs3 images

bmcelvee:
https://bugzilla.redhat.com/show_bug.cgi?id=1590734

### DIFF
--- a/install/disconnected_install.adoc
+++ b/install/disconnected_install.adoc
@@ -253,6 +253,11 @@ $ docker pull registry.access.redhat.com/rhgs3/rhgs-s3-server-rhel7
 +
 [IMPORTANT]
 ====
+For Red Hat support, a Container-Native Storage (CNS) subscription is required for `rhgs3/` images.
+====
++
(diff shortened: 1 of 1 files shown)


install/disconnected_install.adoc
```diff
@@ -253,6 +253,11 @@ $ docker pull registry.access.redhat.com/rhgs3/rhgs-s3-server-rhel7:<tag>
 +
 [IMPORTANT]
 ====
+For Red Hat support, a CNS subscription is required for `rhgs3/` images.
```
kalexand-rh: Are they supported only if you have the subscription, or do you need the subscription to obtain the images at all?

I'd spell out CNS.

kalexand-rh (commented):
I have a question and a suggestion. :cat:

install/disconnected_install.adoc
bmcelvee: Thanks! They're only supported if you have the subscription.

PR #10368, opened 2018-06-22
Bug 1473438 Clarify your_private_key in patch commands

bmcelvee:
https://bugzilla.redhat.com/show_bug.cgi?id=1473438

### DIFF
--- a/install_config/aggregate_logging.adoc
+++ b/install_config/aggregate_logging.adoc
@@ -967,6 +967,12 @@ $ oc patch secrets/logging-fluentd --type=json \
   --patch "[{'op':'add','path':'/data/your_private_key','value':'$(base64 /path/to/your_private_key.pem)'}]"
 ----
 
+[NOTE]
+====
+Replace `your_private_key` with a generic name. This is a link to the JSON path, 
+not a path on your host system
+====
+
 When configuring the external aggregator, it must be able to accept messages
 securely from Fluentd.
 


bmcelvee:
@openshift/team-documentation PTAL :ice_cream:

install_config/aggregate_logging.adoc
```diff
@@ -967,6 +967,12 @@ $ oc patch secrets/logging-fluentd --type=json \
   --patch "[{'op':'add','path':'/data/your_private_key','value':'$(base64 /path/to/your_private_key.pem)'}]"
 ----
 
+[NOTE]
+====
+Replace `your_private_key` with a generic name. This is not a path on your
+host system, but it is the link to the JSON path.
+====
```
mburke5678: @bmcelvee Would this work better as _This is a link to the JSON path, not a path on your host system_?
Do we need to define _JSON path_?

kalexand-rh (commented):
So the change looks fine and exactly matches what the bug's requesting, but I'm so confused.
- The previous paragraph says that the value must match one that's specified somewhere else, not a generic value. Is that a different value?
- Does this advice apply to the `your_ca_cert` path, too?
- I don't know what you mean by "JSON path" or why it should be a generic name.

vikram-redhat:
@bmcelvee - can this be moved along?

PR #10307, opened 2018-06-21
removing content from online

kalexand-rh:
Abhishek Gupta said that the high availability masters section shouldn't be visible in the Online collection.

PR #10387, opened 2018-06-25
removing content from online

kalexand-rh:
From https://github.com/openshift/openshift-docs/pull/10307

### DIFF
--- a/architecture/infrastructure_components/kubernetes_infrastructure.adoc
+++ b/architecture/infrastructure_components/kubernetes_infrastructure.adoc
@@ -86,6 +86,7 @@ balancer of choice.
 endif::[]
 |===
 
+ifdef::openshift-origin,openshift-enterprise,openshift-dedicated[]
 [[high-availability-masters]]
 
 === High Availability Masters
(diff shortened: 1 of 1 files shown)


kalexand-rh:
@vikram-redhat, since online is extremely likely to not use 3.7 again, can I close this PR without merging?

I don't understand these errors. They're not impacted by the section that I changed, and, looking at the local builld, those references are right in dedicated. (I checked "containers" and "architecture-additional-concepts-authorization," and then gave up.)

vikram-redhat:
@kalexand-rh - there is an endif missing in this file which is causing the books to not build properly. The errors are flowing on from there, but not really reflecting the original issue. I am not sure why this wasn't caught in any other branches like master and 3.9. I suspect it could be the result of a bad fix in this file in this branch.

Once you fix it properly nesting the endifs, it should build ok. I added one locally without knowing where it should go in terms of the correct structure of the book, and it builds ok after that. For reference I added it on line 127.

kalexand-rh:
Thank you, @vikram-redhat! I did the cherry-pick over in a test branch, and I see where it opened an ifdef without closing it.

PR #10396, opened 2018-06-25
Discourage reliance on openshift_docker_additional_registries

sdodson:
openshift_docker_additional_registries should not be relied on in order to configure a cluster to use an alternative registry. So remove a few areas where we advise people to configure their cluster in that manner.

https://github.com/openshift/openshift-ansible/pull/8955 is related

### DIFF
--- a/install/configuring_inventory_file.adoc
+++ b/install/configuring_inventory_file.adoc
@@ -244,10 +244,13 @@ For example:
 openshift_docker_additional_registries=example.com:443
 ----
 
+Note: If you need to configure your cluster to use an alternate registry
+set `oreg_url` rather than rely on `openshift_docker_additional_registries`.
+
 |`openshift_docker_insecure_registries`
 |{product-title} adds the specified additional insecure registry or registries to
 the *docker* configuration. For any of these registries, secure sockets layer
-(SSL) is not verified. Also, add these registries to `openshift_docker_additional_registries`.
+(SSL) is not verified.
 
 |`openshift_docker_blocked_registries`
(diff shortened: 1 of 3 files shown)


install/configuring_inventory_file.adoc
```diff
@@ -244,10 +244,13 @@ For example:
 openshift_docker_additional_registries=example.com:443
 ----
 
+Note: If you need to configure your cluster to use an alternate registry it's
+preferred to set `oreg_url` rather than rely on `openshift_docker_additional_registries`.
+
```
mburke5678: s/it's preferred to//

install/disconnected_install.adoc
```diff
@@ -683,6 +682,9 @@ your own registry to be used for Docker search and Docker pull. Use the
 searched. For example, `add_registry=--add-registry registry.access.redhat.com
 --add-registry example.com`.
 
+Note: If you need to configure your cluster to use an alternate registry it's
+preferred to set `oreg_url` rather than rely on `openshift_docker_additional_registries`.
+
```
mburke5678: Same change as above

mburke5678:
@sdodson A couple of nits. Otherwise LGTM.  @adellape These are your files; any thoughts?

install/disconnected_install.adoc
sdodson: updated both.

adellape:
Thanks, continuing via https://github.com/openshift/openshift-docs/pull/10517.

PR #10427, opened 2018-06-26
removing notes about master being unschedulable

kalexand-rh:
@bfallonf, you pointed out on the monster DR PR that in 3.9 and later, master nodes are not unschedulable. Do these changes sound reasonable to you?

### DIFF
--- a/day_two_guide/topics/proc_backing-up-master.adoc
+++ b/day_two_guide/topics/proc_backing-up-master.adoc
@@ -30,8 +30,7 @@ You can customize {product-title} services, such as increasing the log level or
 using proxies. The configuration files are stored in the `/etc/sysconfig`
 directory.
 
-Because the masters are also unschedulable nodes, back up the entire
-`/etc/origin` directory.
+Because the masters are also nodes, back up the entire `/etc/origin` directory.
 
 [discrete]
 == Procedure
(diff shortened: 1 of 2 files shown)


bfallonf:
@kalexand-rh I think so. The only thing I can think of is if the procedures have to happen differently with the new master status (which I doubt). Might be worth checking with @adellape or @sdodson ?

Context: Wondering if the procedures changed in this PR need more changes because masters are schedulable now.

day_two_guide/topics/proc_backing-up-master.adoc
```diff
@@ -30,9 +30,6 @@ You can customize {product-title} services, such as increasing the log level or
 using proxies. The configuration files are stored in the `/etc/sysconfig`
 directory.
 
-Because the masters are also unschedulable nodes, back up the entire
-`/etc/origin` directory.
```
adellape: I think this could just be a change of s/unschedulable/schedulable/, leaving the recommendation to back up all of `/etc/origin/` in.

adellape: Or better yet, s/unschedulable//.

But if the line is taken out, it may also mean making step 1 in the procedure itself more specific.

kalexand-rh: Thanks @adellape!

PR #8859, opened 2018-04-23
Update configuring_aws.adoc

laidback:
Update the document to conform with:
https://github.com/kubernetes-incubator/kube-aws/issues/1025
https://github.com/kubernetes/kubernetes/blob/v1.8.3/pkg/cloudprovider/providers/aws/tags.go#L44-L51

PR #10434, opened 2018-06-26
Update configuring_aws.adoc

kalexand-rh:
Update the document to conform with:
https://github.com/kubernetes-incubator/kube-aws/issues/1025
https://github.com/kubernetes/kubernetes/blob/v1.8.3/pkg/cloudprovider/providers/aws/tags.go#L44-L51

From https://github.com/openshift/openshift-docs/pull/8859


kalexand-rh:
This section isn't in 3.6, so I'm rolling it back to 3.7.

PR #10489, opened 2018-06-28
modules/create_crd_obj: point to CRD creation docs

nikhita:
Fixes three nits:

- a grammatical error.
- points to docs related to creating a CRD, when talking about creating a CR (or else the context is lost, eg about `spec.kind`, `spec.shortnames`, etc).
- fixes `JSON` to `YAML` since the output is described in the YAML format below it.


kalexand-rh:
Thanks @nikhita!

PR #10490, opened 2018-06-28
Remove duplicate words

liangxia:


kalexand-rh:
Thank you, @liangxia! 
@vikram-redhat, should we change the text of the API docs, or do we need to drive changes somewhere else to make them "stick?"

vikram-redhat:
Wow. First there were that many duplicate words? Thank you @liangxia .

Second, the inclusion of the REST API docs makes it hard to CP this to 3.9 or other branches. The REST API guides are supposed to get auto-generated, although Jim Minter is working on updates to it. 

@kalexand-rh - you can merge it to master and then CP to 3.9 and 3.10. Other than that, I am not sure it will work without removing the REST API changes to other earlier branches. Even for 3.10 I am skeptical.

kalexand-rh:
Yes, thanks again @liangxia!
@vikram-redhat, how far back do you want me to try? I'll brace myself for merge errors. ;)

vikram-redhat:
This is a low hanging fruit. I am gonna say let's leave it at this. If you come across these fix them as part of other things.

cc: @aheslin

kalexand-rh:
Sounds good to me.

PR #10510, opened 2018-06-28
heading case, note clarifications

kalexand-rh:


kalexand-rh:
Most of the change is title case to sentence case, but will you please make sure that [this note](https://github.com/openshift/openshift-docs/pull/10510/files#diff-b625c21e153952a3835df02b11e932ccR180) is still good?

kalexand-rh:
@bfallonf, you're right. [Here's the source](https://github.com/openshift/openshift-docs/commit/2c725157cd0b31590d55139addca09d06b90d587). This is the only topic that tries to differentiate between "LDAP groups" and "OpenShift Groups."
- Have you seen usage like this anywhere else?
- Do you think it's fine for me to drop it to lower case and add other clarifications, as required?

bfallonf:
@kalexand-rh Hmm I'm not 100% about any response I could give, but I think it should be OK to give lower case to both. I think there was a time where we gave OpenShift objects caps for whatever reason, and maybe this is leftover from that.

kalexand-rh:
Thank you @bfallonf! It probably is just leftover from that. Thanks for the history and the opinion. :)

Issue #10074, opened 2018-06-13
[enterprise-3.9] Edit suggested in file dev_guide/pod_autoscaling.adoc


nickcmaynard:
### Which section(s) is the issue in?

Pod autoscaling definition.

### What needs fixing?

The HPA definition sample is out of date.  
3.9 uses `apiVersion: autoscaling/v1` with different specification entries.

A valid example:

```yml
apiVersion: autoscaling/v1
kind: HorizontalPodAutoscaler
metadata:
  name: frontend
spec:
  maxReplicas: 10
  minReplicas: 1
  scaleTargetRef:
    apiVersion: apps.openshift.io/v1
    kind: DeploymentConfig
    name: frontend
  targetCPUUtilizationPercentage: 80
```

The pod autoscaling document needs technical oversight and updating for 3.9 validity.

vikram-redhat:
@mburke5678 PTAL.

mburke5678:
@sjenning PTAL

sjenning:
fyi @DirectXMan12 

This is correct.  All references to "autoscaling/v2beta1" should be updated to "autoscaling/v1" for 3.9 and later.

https://docs.openshift.com/container-platform/3.9/dev_guide/pod_autoscaling.html

nickcmaynard:
It was a few weeks ago, but IIRC, I couldn't get the old version to work.  I needed to update it as indicated.  It might be worth testing that @sjenning 's assertion.

mburke5678:
Closing issue. See PR https://github.com/openshift/openshift-docs/pull/10515

PR #10515, opened 2018-06-28
Edit API version in file dev_guide/pod_autoscaling.adoc

mburke5678:
Per https://github.com/openshift/openshift-docs/issues/10074

### DIFF
--- a/dev_guide/pod_autoscaling.adoc
+++ b/dev_guide/pod_autoscaling.adoc
@@ -18,13 +18,6 @@ a replication controller or deployment configuration, based on metrics collected
 from the pods that belong to that replication controller or deployment
 configuration.
 
-ifdef::openshift-enterprise[]
-[NOTE]
-====
-Horizontal pod autoscaling is supported starting in {product-title} Enterprise 3.1.1.
-====
-endif::[]
-
 ifdef::openshift-origin,openshift-enterprise[]
 [[req-for-using-hpas]]
 == Requirements for Using Horizontal Pod Autoscalers
@@ -40,16 +33,18 @@ endif::openshift-origin,openshift-enterprise[]
 The following metrics are supported by horizontal pod autoscalers:
 
 .Metrics
-[cols="3a,8a",options="header"]
+[cols="3a,5a,5a",options="header"]
 |===
 
-|Metric |Description
+|Metric |Description |API version
 
 |CPU utilization
 |Percentage of the xref:../dev_guide/compute_resources.adoc#dev-cpu-requests[requested CPU]
+|`autoscaling/v1`, `autoscaling/v2beta1`
 
 |Memory utilization
 |Percentage of the requested memory.
+|`autoscaling/v2beta1`
 |===
 
 [[hpa-autoscaling]]
@@ -57,7 +52,13 @@ The following metrics are supported by horizontal pod autoscalers:
 
 You can create a horizontal pod autoscaler with the `oc autoscale` command and
 specify the minimum and maximum number of pods you want to run, as well as the
-CPU utilization your pods should target.
+xref:creating-a-hpa[CPU utilization] or xref:pod-autoscaling-memory[memory utilization] your pods should target.
+
+
+[IMPORTANT]
+====
+Autoscaling for Memory Utilization is a Technology Preview feature only.
+====
 
 After a horizontal pod autoscaler is created, it begins attempting to query
 Heapster for metrics on the pods. It may take one to two minutes before Heapster
@@ -97,7 +98,7 @@ deploymentconfig "frontend" autoscaled
 ----
 
 The above example creates a horizontal pod autoscaler with the following
-definition when using version one of the horizontal pod autoscaler:
+definition when using the `autoscaling/v1` version of the horizontal pod autoscaler:
 
 .Horizontal Pod Autoscaler Object Definition
 ====
@@ -111,7 +112,7 @@ spec:
   scaleTargetRef:
     kind: DeploymentConfig <2>
     name: frontend <3>
-    apiVersion: v1 <4>
+    apiVersion: apps/v1 <4>
     subresource: scale
   minReplicas: 1 <5>
   maxReplicas: 10 <6>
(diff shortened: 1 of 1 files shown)


mburke5678:
@sjenning @nickcmaynard 
Changed `autoscaling/v2beta1` to `autoscaling/v1`. 
Changed `spec.metrics.targetAverageUtilization` to `spec.targetAverageUtilization`, per the example in https://github.com/openshift/openshift-docs/issues/10074 

Should `spec.metrics` also be removed?

sjenning:
cc @DirectXMan12

mburke5678:
@DirectXMan12 PTAL

mburke5678:
@mdshuai PTAL

mburke5678:
If we don't hear back from @DirectXMan12 regarding removing `spec.metrics` I can do a follow-up.

mburke5678:
@mdshuai BUMP ^^

dev_guide/pod_autoscaling.adoc
```diff
@@ -324,7 +324,7 @@ kubernetesMasterConfig:
   ...
   apiServerArguments:
     runtime-config:
-    - apis/autoscaling/v2beta1=true
+    - apis/autoscaling/v1=true
```
DirectXMan12: this can't change -- it's specifically referring to the beta version (but the above `v2alpha1` needs to change to `v2beta1`, plus, I think that the beta version is enabled by default now, so this section isn't needed.

DirectXMan12 (changes requested):
Sorry, this dropped of my radar.  What is the purpose of this change?  If the purpose is to only show the stable version of the autoscaling API when possible, that's fine, but you can't stop showing the beta version in the section specifically designed to describe the new beta features.

However, reading the originating issue, it looks like this was spawned out of the idea that the `autoscaling/v2beta1` API group is an older version, which is not true.  Before we proceed with this, I want to make sure everyone's on the same page

dev_guide/pod_autoscaling.adoc
mburke5678: @DirectXMan12 
Per @sjenning in originating issue:
" All references to "autoscaling/v2beta1" should be updated to "autoscaling/v1" for 3.9 and later."
https://github.com/openshift/openshift-docs/issues/10074#issuecomment-401102988

Is this not true?

mburke5678: Related: https://github.com/openshift/openshift-docs/issues/10737 ??

_"I've tried to create an Autoscaler object using the first definition, and I just get the error:
error: unable to recognize "create_tasksapp_autoscaler.yaml": no matches for extensions/, Kind=HorizontalPodAutoscaler_

_If I use oc autoscale, the created object does not use the v2 Autoscaler as specified in the docs. It creates the following object instead:_

_apiVersion: autoscaling/v1_

DirectXMan12: I'm not sure we do a good communicating this, *but* the object is available in both forms, since it's a normal API object.  You just have to fetch it in the correct version (`oc get hpa.v2beta1.autoscaling`).  That being said *most* of the references should probably say `autoscaling/v1`, since it's still the stable version, and then the section discussing the new version should say `autoscaling/v2beta1` (since that's the new version).  I think @sjenning may have misread.

dev_guide/pod_autoscaling.adoc
```diff
@@ -205,7 +205,7 @@ metadata:
   name: hpa-resource-metrics-memory <1>
 spec:
   scaleTargetRef:
-    apiVersion: apps/v1beta1 <2>
+    apiVersion: apps/v2beta1 <2>
```
DirectXMan12: whoops, this shouldn't have changed -- that's a reference to the `apps` API group (but I think it can be `apps/v1` now)

dev_guide/pod_autoscaling.adoc
```diff
@@ -315,8 +315,8 @@ is able to calculate desired scales.
 ** A `True` condition indicates that you need to raise or lower the minimum or maximum replica count in order to scale.
 ** A `False` condition indicates that the requested scaling is allowed.
 
-The horizontal pod autoscaler status conditions are only available with the `v2alpha1` version of the
-autoscaling API by adding the following to your cluster's `master-config.yaml` file:
+The horizontal pod autoscaler status conditions are available with the `v2beta1` version of the
+autoscaling API. Check that `apis/autoscaling/v2beta1=true` is present in your cluster's `master-config.yaml` file:
```
DirectXMan12: this line is no longer necessary

DirectXMan12 (changes requested):

mburke5678:
@mdshuai PTAL

mburke5678:
@DirectXMan12 Is `spec.metrics` in  APIv1? It looks like it was added in v2beta1. If so, does that impact the text above? 
https://docs.openshift.com/container-platform/3.9/dev_guide/pod_autoscaling.html#overview

DirectXMan12:
`spec.metrics` is part of v2beta1. Not sure which specific part you're talking about, but

`when using version two of the horizontal pod autoscaler:` needs to show an actual v2 autoscaler.  Sorry I didn't notice that before.

While we're at it, for 3.11, references to Heapster should be replaced with metrics server, and we should update the link about setting up cluster metrics to mention that you can deploy metrics-server by itself without running the metrics playbook.

mburke5678:
@DirectXMan12 
Solly, each of the sections at the top of the page (Overview, Requirements for Using Horizontal Pod Autoscalers, Supported Metrics, Autoscaling, Autoscaling for CPU Utilization) contain references to metrics. I am wondering if there are any changes to be made if the user goes with v1 over o2beta1

mburke5678:
@mdshuai PTAL

mburke5678:
@DirectXMan12 
Made some significant changes to I hope capture the differences between the use of v1 and v2beta1. See preview:
https://fix-api-version-per-issue-10074--ocpdocs.netlify.com/openshift-enterprise/(head%20detached%20at%20fetch_head)/dev_guide/pod_autoscaling.html

mburke5678:
@mdshuai PTAL

mburke5678:
@nickcmaynard @sjenning @DirectXMan12 @mdshuai 
I am rolling these changes to 3.6 when Autoscaling for Memory Utilization was added. Please let me know if this is not correct.

vikram-redhat:
@mdshuai - PTAL.

cc: @xltian

vikram-redhat:
@mburke5678 have added the 3.11 label.

mdshuai:
The change LGTM, thanks

> @nickcmaynard @sjenning @DirectXMan12 @mdshuai
I am rolling these changes to 3.6 when Autoscaling for Memory Utilization was added. Please let me know if this is not correct.

For the v2 version, `autoscaling/v2beta1` is default enabled at 3.9. In 3.6 we only support `autoscaling/v2alpha1`.

PR #10519, opened 2018-06-28
clarify that reencrypt is supported for registry routes

bparees:


bparees:
3.7+ only.

mburke5678:
@bparees Do we need to explain how to configure reencrypt routes here? (We don't seem to explain this elsewhere in the docs. 
If not, one nit and LGTM

install_config/registry/securing_and_exposing_registry.adoc
```diff
@@ -268,8 +268,7 @@ DNS to the router's IP address.
 +
 [NOTE]
 ====
-Passthrough is currently the only type of route supported for exposing the
-secure registry.
+Reencrypt routes are also supported for exposing the secure registry.
```
mburke5678: s/Reencrypt/Re-encrypt
(we seem to use the hyphenated version)

bparees:
> Do we need to explain how to configure reencrypt routes here? (We don't seem to explain this elsewhere in the docs.

no i don't think this is the right place to cover that, but if it's really something we don't document anywhere, a new bug should be opened for the networking team to add documentation for it.

install_config/registry/securing_and_exposing_registry.adoc
bparees: fixed

mburke5678:
@bparees I found one reference here:
_These self-signed certificates are not recognized by browsers. To mitigate this issue, use a publicly signed certificate, then configure it to re-encrypt traffic with the self-signed certificate._ With steps to use _oc create route reencrypt_
https://docs.openshift.com/container-platform/3.9/day_two_guide/certificate_management.html#change-app-cert-to-ca-signed-cert

bparees:
@mburke5678 so can this merge?

mburke5678:
@bparees We usually ask that a QE resource also review. is there someone you can recommend?

bmcelvee:
@mburke5678 This is such a minor change I don't think it needs QE :smile_cat:

bparees:
for what it's worth this was closed out via 

https://bugzilla.redhat.com/show_bug.cgi?id=1464562
https://github.com/openshift/origin/issues/14249
https://github.com/openshift/origin/pull/15694

mburke5678:
@bparees I just wanted a "second opinion" that reencrypt routes are supported as we are stating. But, if we are all OK, with it, based on the BZ, I can merge away!

mburke5678:
@bparees Do we want to add this info to the 3.7 release notes?
https://github.com/openshift/openshift-docs/pull/10606

bparees:
> @bparees Do we want to add this info to the 3.7 release notes?

i guess

PR #10531, opened 2018-06-29
rm io.openshift.expose-services label

adambkaplan:
Remove the io.openshift.expose-services label from docs.
This is not consumed in any meaningful way by OpenShift.

Bug [1583274](https://bugzilla.redhat.com/show_bug.cgi?id=1583274)


bparees:
lgtm, applicable to all releases.

PR #10539, opened 2018-06-29
bug 1591156 adding cluster role

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1591156

@enj, will you PTAL?
Do we have information about the different rules that we define in the default cluster roles that could help users create their own?

Note for me: be sure to remove the references to cluster role aggregation from 3.7.

### DIFF
--- a/admin_guide/manage_rbac.adoc
+++ b/admin_guide/manage_rbac.adoc
@@ -25,7 +25,7 @@ and manage local roles and bindings.
 endif::[]
 
 [[viewing-roles-and-bindings]]
-== Viewing Roles and Bindings
+== Viewing roles and bindings
 
 xref:../architecture/additional_concepts/authorization.adoc#roles[Roles] can be used to grant
 various levels of access both
@@ -34,12 +34,12 @@ as well as at the
 xref:../architecture/additional_concepts/authorization.adoc#cluster-and-local-rbac[project-scope].
 xref:../architecture/additional_concepts/authentication.adoc#users-and-groups[Users
 and groups] can be associated with, or _bound_ to, multiple roles at the same
-time.  You can view details about the roles and their bindings using the `oc
+time. You can view details about the roles and their bindings using the `oc
 describe` command.
 
 ifdef::openshift-dedicated[]
-Users with the *dedicated-cluster-admin* role can view but not manage cluster roles.  They can manage cluster
-role bindings and manage local roles and bindings.  Users with the *admin*
+Users with the *dedicated-cluster-admin* role can view but not manage cluster roles. They can manage cluster
+role bindings and manage local roles and bindings. Users with the *admin*
 xref:../architecture/additional_concepts/authorization.adoc#roles[default cluster role]
 bound locally can manage roles and bindings in that project.
 endif::[]
@@ -60,16 +60,10 @@ Authorization] section.
 ====
 
 [[viewing-cluster-roles]]
-=== Viewing Cluster Roles
+=== Viewing cluster roles
 ifdef::openshift-enterprise,openshift-origin[]
 To view the cluster roles and their associated rule sets:
 
-----
-$ oc describe clusterrole.rbac
-----
-
-.Viewing Cluster Roles
-[options="nowrap"]
 ----
 $ oc describe clusterrole.rbac
 Name:		admin
@@ -278,17 +272,13 @@ PolicyRule:
 ----
 endif::[]
 
+[[viewing-cluster-bindings]]
+=== Viewing cluster role bindings
+
 To view the current set of cluster role bindings, which show the users and
 groups that are bound to various roles:
 
-----
-$ oc describe clusterrolebinding.rbac
-----
-
 ifdef::openshift-enterprise,openshift-origin[]
-[[viewing-cluster-bindings]]
-.Viewing Cluster Role Bindings
-[options="nowrap"]
 ----
 $ oc describe clusterrolebinding.rbac
 Name:		admin
@@ -493,12 +483,16 @@ Labels:		kubernetes.io/bootstrapping=rbac-defaults
 Annotations:	rbac.authorization.kubernetes.io/autoupdate=true
 
 ...
-
+----
+endif::[]
+ifdef::openshift-dedicated[]
+----
+$ oc describe clusterrolebinding.rbac
 ----
 endif::[]
 
 [[viewing-local-roles-and-bindings]]
-=== Viewing Local Roles and Bindings
+=== Viewing local roles and bindings
 
 All of the
 xref:../architecture/additional_concepts/authorization.adoc#roles[default
@@ -520,9 +514,6 @@ viewing the local role bindings of another project, if the user already has the
 xref:../architecture/additional_concepts/authorization.adoc#roles[default
 cluster role] in it.
 
-[[viewing-local-bindings]]
-.Viewing Local Role Bindings
-[options="nowrap"]
 ----
 $ oc describe rolebinding.rbac -n joe-project
 Name:		admin
@@ -574,7 +565,7 @@ Subjects:
 ----
 
 [[managing-role-bindings]]
-== Managing Role Bindings
+== Managing role bindings
 
 Adding, or _binding_, a
 xref:../architecture/additional_concepts/authorization.adoc#roles[role] to
@@ -587,7 +578,7 @@ When managing a user or group's associated roles for local role bindings using t
 following operations, a project may be specified with the `-n` flag. If it is
 not specified, then the current project is used.
 
-.Local Role Binding Operations
+.Local role binding operations
 [options="header"]
 |===
 
@@ -620,9 +611,9 @@ ifdef::openshift-enterprise,openshift-origin[]
 
 You can also manage cluster role bindings using the following
 operations. The `-n` flag is not used for these operations because
-cluster role bindings uses non-namespaced resources.
+cluster role bindings use non-namespaced resources.
 
-.Cluster Role Binding Operations
+.Cluster role binding operations
 [options="header"]
 |===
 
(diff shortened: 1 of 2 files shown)


enj:
The CLI commands are a better approach for creating cluster roles:

```
$ oc create clusterrole -h
Create a ClusterRole.

Usage:
  oc create clusterrole NAME --verb=verb --resource=resource.group [--resource-name=resourcename] [--dry-run] [flags]

Examples:
  # Create a ClusterRole named "pod-reader" that allows user to perform "get", "watch" and "list" on pods
  oc create clusterrole pod-reader --verb=get,list,watch --resource=pods
  
  # Create a ClusterRole named "pod-reader" with ResourceName specified
  oc create clusterrole pod-reader --verb=get --resource=pods --resource-name=readablepod --resource-name=anotherpod
  
  # Create a ClusterRole named "foo" with API Group specified
  oc create clusterrole foo --verb=get,list,watch --resource=rs.extensions
  
  # Create a ClusterRole named "foo" with SubResource specified
  oc create clusterrole foo --verb=get,list,watch --resource=pods,pods/status
  
  # Create a ClusterRole name "foo" with NonResourceURL specified
  oc create clusterrole "foo" --verb=get --non-resource-url=/logs/*
  
  # Create a ClusterRole name "monitoring" with AggregationRule specified
  oc create clusterrole monitoring --aggregation-rule="rbac.example.com/aggregate-to-monitoring=true"

Options:
      --aggregation-rule=: An aggregation label selector for combining ClusterRoles.
      --allow-missing-template-keys=true: If true, ignore any errors in templates when a field or map key is missing in
the template. Only applies to golang and jsonpath output formats.
      --dry-run=false: If true, only print the object that would be sent, without sending it.
      --non-resource-url=[]: A partial url that user should have access to.
  -o, --output='': Output format. One of:
json|yaml|wide|name|custom-columns=...|custom-columns-file=...|go-template=...|go-template-file=...|jsonpath=...|jsonpath-file=...
See custom columns [http://kubernetes.io/docs/user-guide/kubectl-overview/#custom-columns], golang template
[http://golang.org/pkg/text/template/#pkg-overview] and jsonpath template
[http://kubernetes.io/docs/user-guide/jsonpath].
      --resource=[]: Resource that the rule applies to
      --resource-name=[]: Resource in the white list that the rule applies to, repeat this flag for multiple items
      --save-config=false: If true, the configuration of current object will be saved in its annotation. Otherwise, the
annotation will be unchanged. This flag is useful when you want to perform kubectl apply on this object in the future.
      --template='': Template string or path to template file to use when -o=go-template, -o=go-template-file. The
template format is golang templates [http://golang.org/pkg/text/template/#pkg-overview].
      --validate=false: If true, use a schema to validate the input before sending it
      --verb=[]: Verb that applies to the resources contained in the rule

Use "oc options" for a list of global command-line options (applies to all commands).
```

admin_guide/manage_rbac.adoc
```diff
@@ -762,8 +762,98 @@ To bind the new role to a user, run:
 $ oc adm policy add-role-to-user podview user2 --role-namespace=bob-project -n bob-project
 ----
 
+[[creating-cluster-role]]
+== Creating a cluster role
+
+. Export a cluster role to modify:
++
+----
+oc get clusterrole admin -o yaml > newrole.yaml
```
enj: No YAML, this should follow the same pattern as the "create local role" above.

admin_guide/manage_rbac.adoc
```diff
@@ -762,8 +762,98 @@ To bind the new role to a user, run:
...
+<4> Customize the cluster role rules as required.
+
+. Create the role:
++
+----
+$ oc create -f newrole.yaml
+----
+
+. Grant the permissions in the cluster role to a user:
+.. To grant a user permissions in all namespaces:
```
enj: projects

admin_guide/manage_rbac.adoc
```diff
@@ -762,8 +762,98 @@ To bind the new role to a user, run:
...
++
+----
+$ oadm policy add-cluster-role-to-user <new_cluster_role> <username>
```
enj: oc adm

kalexand-rh:
@xltian, will you PTAL?

xltian:
@stuartchuan  Will work on it.

barleyer:
Should we use `clusterrole.rbac role.rbac clusterrolebinding.rbac rolebinding.rbac` in the docs?

admin_guide/manage_rbac.adoc
```diff
@@ -762,8 +762,158 @@ To bind the new role to a user, run:
...
+----
+
+////
+. Grant the permissions in the cluster role to a user:
+.. To grant a user permissions in all projects:
++
+----
+$ oc adm policy add-cluster-role-to-user <new_cluster_role> <username>
+----
+
```
barleyer: Maybe we should remind that there's also another way to bind a {cluster-}role to users. `oc create {cluster}role{binding}`

kalexand-rh: Is there a different time that a user would want to do this? @enj, what do you think?

kalexand-rh: @enj, when you have a minute, will you PTAL?

barleyer:
From QE，LGTM.

admin_guide/manage_rbac.adoc
kalexand-rh: After a chat with @enj, I think that `add-cluster-role-to-user` is the right way to go.

kalexand-rh:

admin_guide/manage_rbac.adoc
```diff
@@ -622,7 +622,7 @@ You can also manage cluster role bindings using the following
 operations. The `-n` flag is not used for these operations because
 cluster role bindings uses non-namespaced resources.
```
mburke5678: s/uses/use ? Or, is "cluster role bindings" a singular term?

admin_guide/manage_rbac.adoc
```diff
@@ -762,8 +762,158 @@ To bind the new role to a user, run:
...
+[[creating-cluster-role]]
+== Creating a new cluster role
+
+To create a cluster role, run the following command:
+
+----
+$ oc create clusterrole -h ...
+----
+
+The following excerpt from the help of this command describes its usage:
```
mburke5678: This is unusual for our docs, I think. I see it used elsewhere in this topic, but it feels like bad form. Can these be bulleted instead and remove the "excerpt" intro?

admin_guide/manage_rbac.adoc
mburke5678: @kalexand-rh Did you get this comment ^^ just checkin'

admin_guide/manage_rbac.adoc
```diff
@@ -762,8 +762,158 @@ To bind the new role to a user, run:
...
+
+----
+
+[[creating-derivative-cluster-role]]
+== Creating a cluster role from another cluster role
+
+. Export a cluster role to modify:
++
+----
+oc get clusterrole admin -o yaml > newrole.yaml
```
bfallonf: s/oc get.../$ oc get///

admin_guide/manage_rbac.adoc
```diff
@@ -60,15 +60,15 @@ Authorization] section.
...
+=== Viewing cluster roles
 ifdef::openshift-enterprise,openshift-origin[]
 To view the cluster roles and their associated rule sets:
 
 ----
 $ oc describe clusterrole.rbac
 ----
 
-.Viewing Cluster Roles
+.Viewing cluster roles
```
mburke5678: I realize these were already there; but, are they needed? Seem superfluous.

admin_guide/manage_rbac.adoc
bfallonf: @kalexand-rh I'd not put help output into the docs. That's a slippery slope (why not just do it for everything, y'now?). I'd suggest summarising what it can do and maybe use one of the examples.

It might make sense to put somewhere in the docs "You can always see <command> --help for any command for more information", but that might be out of scope for this BZ.

admin_guide/manage_rbac.adoc
```diff
@@ -762,8 +762,158 @@ To bind the new role to a user, run:
...
+annotation will be unchanged. This flag is useful when you want to perform kubectl apply on this object in the future.
+      --template='': Template string or path to template file to use when -o=go-template, -o=go-template-file. The
+template format is golang templates [http://golang.org/pkg/text/template/#pkg-overview].
+      --validate=false: If true, use a schema to validate the input before sending it
+      --verb=[]: Verb that applies to the resources contained in the rule
+
+----
+
+[[creating-derivative-cluster-role]]
+== Creating a cluster role from another cluster role
```
bfallonf: I can't think of a reason why someone would do this. Is this information you have? Can that be included?

mburke5678:
@kalexand-rh A few things. Otherwise LGTM

admin_guide/manage_rbac.adoc
```diff
@@ -762,8 +762,158 @@ To bind the new role to a user, run:
...
+
+. Export a cluster role to modify:
++
+----
+oc get clusterrole admin -o yaml > newrole.yaml
+----
++
+You can export any cluster role, including a
+xref:../architecture/additional_concepts/authorization.adoc#roles[default cluster role],
+to start your new role.
```
bfallonf: I remember a BZ a few moths ago that was about adding a note box saying "Don't change the default roles and instead create new ones from it". Is that relevant here? Should that be noted? (This may invalidate my comment above.)

bfallonf:
@kalexand-rh A few notes from me. One of which I see is echoing @mburke5678 sentimonies above.

admin_guide/manage_rbac.adoc
enj: Remove the YAML please.  The create command is the correct approve.

admin_guide/manage_rbac.adoc
kalexand-rh: @mburke5678, I don't think so. I think that any xrefs can go to the section heading. (Does the team have a set practice for if or when we label commands and examples?)

admin_guide/manage_rbac.adoc
kalexand-rh: @mburke5678 @bfallonf, I'll tidy up both sections. (It is a slippery slope, and I originally had only the next new section.) Thank you!

kalexand-rh:
@mburke5678, @bfallonf, I think this is better. Do you agree?

admin_guide/manage_rbac.adoc
```diff
@@ -574,7 +565,7 @@ Subjects:
 ----
 
 [[managing-role-bindings]]
-== Managing Role Bindings
+== Managing role bindings
```
mburke5678: Did you mean to remove the [[viewing-local-bindings]] on line 523?

admin_guide/manage_rbac.adoc
```diff
@@ -710,60 +701,58 @@ Subjects:
...
-
-  # Create a Role named "foo" with SubResource specified
-  oc create role foo --verb=get,list,watch --resource=pods,pods/status
 
-Options:
-      --dry-run=false: If true, only print the object that would be sent, without sending it.
-      --resource=[]: resource that the rule applies to
-      --resource-name=[]: resource in the white list that the rule applies to, repeat this flag for multiple items
-      --verb=[]: verb that applies to the resources contained in the rule
+. To bind the new role to a user, run the following command:
```
mburke5678: @kalexand-rh Looks much better without the help text. If you wanted to be industrious, you could convert them to properly-formatted examples.

mburke5678:
@kalexand-rh Much better. Couple of comments.

admin_guide/manage_rbac.adoc
kalexand-rh: Yes. The collection had one link to one example, and I think that a link to the nearest section head works as well.

admin_guide/manage_rbac.adoc
kalexand-rh: @mburke5678, I thought about making a table of example commands, but I didn't think they were different enough! Thank you for rechecking this one for me. :)

bfallonf:
@kalexand-rh LGTM :elephant:

kalexand-rh:
:tada: the Travis annotation is back! Merging.

PR #10591, opened 2018-07-03
Bug 1591941 Added footnotes about cidr range for egress_source value

bfallonf:
For: https://bugzilla.redhat.com/show_bug.cgi?id=1591941

Will verify in the BZ.

### DIFF
--- a/admin_guide/managing_networking.adoc
+++ b/admin_guide/managing_networking.adoc
@@ -523,7 +523,7 @@ endif::openshift-origin[]
       privileged: true
     env:
     - name: EGRESS_SOURCE <2>
-      value: 192.168.12.99
+      value: 192.168.12.99/24
     - name: EGRESS_GATEWAY <3>
       value: 192.168.12.1
     - name: EGRESS_DESTINATION <4>
@@ -545,8 +545,12 @@ endif::openshift-origin[]
 moves it into the pod's network project before starting the *egress-router*
 container. Preserve the quotation marks around `"true"`. Omitting them results
 in errors. To create the Macvlan interface on a network interface other than the primary one, set the annotation value to the name of that interface. For example, `eth1`.
-<2> IP address from the physical network that the node is on and is
-reserved by the cluster administrator for use by this pod.
+<2> IP address from the physical network that the node is on and is reserved by the
+cluster administrator for use by this pod. Optionally, you can include the
+subnet length, the `/24` suffix, so that a proper route to the local subnet can
+be set up. If you do not specify a subnet length, then the egress router can
+access only the host specified with the `EGRESS_GATEWAY` variable and no other
+hosts on the subnet.
 <3> Same value as the default gateway used by the node.
 <4> The external server to direct traffic to. Using this example,
 connections to the pod are redirected to 203.0.113.25, with a source IP address
@@ -653,12 +657,12 @@ endif::openshift-origin[]
     securityContext:
       privileged: true
     env:
-    - name: EGRESS_SOURCE
-      value: 192.168.12.99
+    - name: EGRESS_SOURCE <1>
+      value: 192.168.12.99/24
     - name: EGRESS_GATEWAY
       value: 192.168.12.1
-    - name: EGRESS_DESTINATION
-      value: | <1>
+    - name: EGRESS_DESTINATION <2>
+      value: |
         80   tcp 203.0.113.25
         8080 tcp 203.0.113.26 80
         8443 tcp 203.0.113.26 443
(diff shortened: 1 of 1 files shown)


admin_guide/managing_networking.adoc
```diff
@@ -547,7 +547,7 @@ moves it into the pod's network project before starting the *egress-router*
 container. Preserve the quotation marks around `"true"`. Omitting them results
 in errors. To create the Macvlan interface on a network interface other than the primary one, set the annotation value to the name of that interface. For example, `eth1`.
 <2> IP address from the physical network that the node is on and is
-reserved by the cluster administrator for use by this pod.
+reserved by the cluster administrator for use by this pod. Optionally, you can specify a CIDR so that the egress router can access the subnet: `192.168.12.99/24`.
```
danwinship: I think it would probably be better to change the examples to use the CIDR form, because it's not really that hard for the user to do, and it results in fewer surprises.

So change the `EGRESS_SOURCE` value above in the yaml to `192.168.12.99/24`, and then change the footnote here to something like:

> IP address from the physical network that the node is on and is reserved by the cluster administrator for use by this pod. For best results, the IP address should include the subnet length (eg, the `/24` here) so that a proper route to the local subnet can be set up. (If you do not include a subnet length then the egress router will not be able to access other hosts on its subnet, besides the `EGRESS_GATEWAY` host.)

Then in the following examples that don't currently have `EGRESS_SOURCE` footnotes, just add the `/24`, and in the ones that do have `EGRESS_SOURCE` footnotes, copy the same text as you put here

bfallonf:
Thanks, @danwinship . I've updated to your suggestion, with a bit of rewording (mainly to get away from the brackets and to highlight the optional-but-recommended-ness of it). If you have more thoughts, please let me know.

admin_guide/managing_networking.adoc
```diff
@@ -545,8 +545,12 @@ endif::openshift-origin[]
 moves it into the pod's network project before starting the *egress-router*
 container. Preserve the quotation marks around `"true"`. Omitting them results
 in errors. To create the Macvlan interface on a network interface other than the primary one, set the annotation value to the name of that interface. For example, `eth1`.
-<2> IP address from the physical network that the node is on and is
-reserved by the cluster administrator for use by this pod.
+<2> IP address from the physical network that the node is on and is reserved by the
```
kalexand-rh: I'd s/and is reserved/and that is reserved

admin_guide/managing_networking.adoc
```diff
@@ -545,8 +545,12 @@ endif::openshift-origin[]
...
+cluster administrator for use by this pod. Optionally, you can include the
+subnet length (the `/24`) so that a proper route to the local subnet can be set
```
kalexand-rh: To avoid parentheses, s/length (the `/24`)/length, the `/24` suffix, 
s/set up/set

admin_guide/managing_networking.adoc
```diff
@@ -545,8 +545,12 @@ endif::openshift-origin[]
...
+up. If you do not specify a subnet length, then the egress router can access
+only the host specified with the `EGRESS_GATEWAY` variable, and no other hosts
```
kalexand-rh: s/host/host that is
no comma

admin_guide/managing_networking.adoc
```diff
@@ -674,7 +678,13 @@ ifdef::openshift-origin[]
     image: openshift/origin-pod
 endif::openshift-origin[]
 ----
-<1> This uses the YAML syntax for a multi-line string; see below for
+<1> IP address from the physical network that the node is on and is reserved by the
```
kalexand-rh: same here :)

admin_guide/managing_networking.adoc
```diff
@@ -674,7 +678,13 @@ ifdef::openshift-origin[]
...
+cluster administrator for use by this pod. Optionally, you can include the
+subnet length (the `/24`) so that a proper route to the local subnet can be set
+up. If you do not specify a subnet length, then the egress router can access
+only the host specified with the `EGRESS_GATEWAY` variable, and no other hosts
+on the subnet.
+<2> This uses the YAML syntax for a multi-line string; see below for
```
kalexand-rh: I might move the annotation up to the same line as EGRESS_DESTINATION and say something like "\<2> Provide `EGRESS_DESTINATION` values as multi-line YAML string. Use the following formats:
ul goes here"

kalexand-rh (commented):
A couple of style picks and suggestions, including one for a line you didn't actually change (sorry!). This sounds good. :)

Issue #10097, opened 2018-06-14
[enterprise-3.9] Edit suggested in file admin_guide/ipsec.adoc


smarterclayton:
### Which section(s) is the issue in?

Title

### What needs fixing?

Title is unclear.  Should be "Encrypting traffic between nodes with IPSec".  It is not about encrypting hosts.

vikram-redhat:
@tmorriso-rh PTAL.

tmorriso-rh:
Work in progress: https://github.com/openshift/openshift-docs/pull/10597

tmorriso-rh:
Fixed via:

https://github.com/openshift/openshift-docs/pull/10597
https://github.com/openshift/openshift-docs/pull/10754
https://github.com/openshift/openshift-docs/pull/10755

PR #10597, opened 2018-07-03
Change title to encrypting traffic

ghost:
@openshift/team-documentation PTAL. This is a follow-up to https://github.com/openshift/openshift-docs/issues/10097.

### DIFF
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -646,7 +646,7 @@ Topics:
 - Name: Encrypting Data at Datastore Layer
   File: encrypting_data
   Distros: openshift-origin,openshift-enterprise
-- Name: Encrypting Hosts with IPsec
+- Name: Encrypting traffic between nodes with IPsec
   File: ipsec
   Distros: openshift-origin,openshift-enterprise
 - Name: Building Dependency Trees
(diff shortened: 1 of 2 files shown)


mburke5678:
@tmorriso-rh LGTM.
Did we decide to use Title case instead of Sentence case for now? (I seem to recall the convereation, but forget the outcome)
Also, out of curiosity, I see that `scaling_performance/network_optimization.adoc` topic uses IPSec in a couple of places. Is it worth changing it there too?

tmorriso-rh:
@mburke5678 No, it's sentence case. I changed this on the document and in the topic map. Thanks.

bfallonf:
@tmorriso-rh @mburke5678 Yes, sentence case.

Traci, is there a reason why you're deleting the anchors? I believe they can still exist with the discrete anchors. I can see that the `admin-guide-ipsec-opportunistic-group-configuration` anchor has links to it within this file, but you're deleting them as well. Is that intentional?

tmorriso-rh:
@bfallonf I followed the guidance in the template. The templates don't show the discrete anchors used within the file.

kalexand-rh:
This isn't a module, though, so I don't think that guidance from the template applies here. I'd put the anchors back in and remove the discrete tags.

tmorriso-rh:
@vikram-redhat Can you take a look at the "The command "pip install aura.tar.gz" failed and exited with 2 during" error message in the build? Thanks.

PR #10620, opened 2018-07-04
BZ#1548277 - added Azure prerequisites

gaurav-nelson:
For https://bugzilla.redhat.com/show_bug.cgi?id=1548277

### DIFF
--- a/install_config/configuring_azure.adoc
+++ b/install_config/configuring_azure.adoc
@@ -11,25 +11,43 @@
 toc::[]
 
 == Overview
-{product-title} can be configured to access an
-link:https://azure.microsoft.com/en-us/services/storage/disks/[Azure
+{product-title} can be configured to access a
+link:https://azure.microsoft.com/en-us/services/storage/disks/[Microsoft Azure
 infrastructure], including
 xref:../install_config/persistent_storage/persistent_storage_azure.adoc#install-config-persistent-storage-persistent-storage-azure[using
-Azure disk as persistent storage] for application data. After Azure is
+Azure disk as persistent storage] for application data. After Microsoft Azure is
 configured properly, some additional configurations need to be completed on the
 {product-title} hosts.
 
 [[configuring-azure-permissions]]
 == Permissions
-Configuring Azure for {product-title} requires the following role:
+Configuring Microsoft Azure for {product-title} requires the following role:
 
 [cols='1,3']
 |===
 
 | Contributor
-| To create and manage all types of Azure resources.
+| To create and manage all types of Microsoft Azure resources.
 
 |===
+For more information about adding administrator roles, see link:https://docs.microsoft.com/en-us/azure/billing/billing-add-change-azure-subscription-administrator[Add or change Azure subscription administrators].
+
+[[configuring-azure-prerequisites]]
+== Prerequisites
+
+* If you are using Microsoft Azure Disk as a persistent volume on the {product-title} version 3.5 or later, you must enable Azure Cloud Provider.
+* All {product-title} node virtual machines (VMs) running in Microsoft Azure must belong to a single resource group.
+* Microsoft Azure VMs must be named the same as {product-title} nodes and this
+cannot include capital letters.
+* If you plan to use  Azure Managed Disks:
+  ** {product-title} version 3.7 or later is required.
+  ** You must create VMs with Azure Managed Disks.
+* If you plan to use unmanaged disks:
+  ** You must create VMs with unmanaged disks.
+* If you are using a custom DNS configuration for your {product-title} cluster or
+your cluster nodes are in different Microsoft Azure Virtual Networks (VNet), you
+must configure DNS so that each node in the cluster can resolve IP addresses for
+other nodes.
 
 [[azure-configuration-file]]
 == The Azure configuration file


install_config/configuring_azure.adoc
```diff
@@ -11,26 +11,49 @@
...
+resource group.
+* Microsoft Azure VM names:
+  ** Must be same as OpenShift node names, and
+  ** These names must not be included any capital letters.
+* If you plan to use Azure Managed Disks:
+  ** {product-title} version 3.7 or later is required.
+  ** You must create VMs with Azure Managed Disks.
+* If you plan to use unmanaged disks:
+  ** You must create VMs with unmanaged disks.
+* Please be aware of node names. All nodes in OpenShift cluster have to communicate with their names. It means you have to configure DNS to resolve private IP from node names. If you create Azure VNet/VNIC/VM by default, you don't have to configure anything about DNS because Azure default DNS can do that. If you use custom DNS for OpenShift cluster, you have to configure by yourself.
```
gaurav-nelson: @tanaka-takayoshi 

I am not entirely sure what this means, can you help me re-write this?
```
Please be aware of node names. All nodes in OpenShift cluster have to communicate with their names. It means you have to configure DNS to resolve private IP from node names. If you create Azure VNet/VNIC/VM by default, you don't have to configure anything about DNS because Azure default DNS can do that. If you use custom DNS for OpenShift cluster, you have to configure by yourself.
```

install_config/configuring_azure.adoc
```diff
@@ -11,26 +11,49 @@
...
 
+[[configuring-azure-prerequisites]]
+== Prerequisites
+
+* If you are using Microsoft Azure Disk as an OpenShift Persistent Volume for  {product-title} version 3.7 or lower, you must enable Azure Cloud Provider.
++
+[NOTE]
+====
+This is not required if you are using Azure File, Azure Blob storage for docker
+registry, or Azure Active Directory OpenID Connect for master authentication.
```
tanaka-takayoshi: I'm afraid this note is based on my comment. However, I found my comment is worng. Azure File requires Azure Cloud Provider too. Could you remove this note?

tanaka-takayoshi:
I added one comment. It looks good for me for others.

tanaka-takayoshi (changes requested):

gaurav-nelson:
Thanks @tanaka-takayoshi I have removed that note.

install_config/configuring_azure.adoc
```diff
@@ -11,26 +11,45 @@
...
 | Contributor
-| To create and manage all types of Azure resources.
+| To create and manage all types of Microsoft Azure resources.
 
 |===
 
+[[configuring-azure-prerequisites]]
+== Prerequisites
+
+* If you are using Microsoft Azure Disk as an OpenShift Persistent Volume for  {product-title} version 3.7 or lower, you must enable Azure Cloud Provider.
```
weherdh: Actually, Azure disk was started from 3.5, any lower version OCP cannot use Azure disk.

gaurav-nelson:
Thanks @weherdh I have updated the description as follows:
```
If you are using Microsoft Azure Disk as an OpenShift Persistent Volume for  {product-title}'s latest version (or lower upto version 3.5), you must enable Azure Cloud Provider.
```
Can you please verify if this is correct?

gaurav-nelson:
@openshift/team-documentation PTAL 📆

install_config/configuring_azure.adoc
```diff
@@ -11,26 +11,45 @@
 toc::[]
 
 == Overview
-{product-title} can be configured to access an
```
bfallonf: @gaurav-nelson I think instead of deleting the "an", I'd replace with "a". It's still needed.

install_config/configuring_azure.adoc
```diff
@@ -11,26 +11,45 @@
...
 | Contributor
-| To create and manage all types of Azure resources.
+| To create and manage all types of Microsoft Azure resources.
 
 |===
 
+[[configuring-azure-prerequisites]]
+== Prerequisites
+
+* If you are using Microsoft Azure Disk as an OpenShift Persistent Volume for  {product-title}'s latest version (or lower upto version 3.5), you must enable Azure Cloud Provider.
```
bfallonf: I think I'm confused by the above sentence. Maybe:

"If you are using Microsoft Azure Disk as a persistent volume on the {product-title} version 3.5 or higher, you must enable Azure Cloud Provider."

That said, you use "Azure Disk" above without a "Microsoft" preceding it. So I'm not sure what the standard is? Also, does persistent volume need caps and "OpenShift" before it? Also, not sure what "lower upto version 3.5). Hence the above.

install_config/configuring_azure.adoc
```diff
@@ -11,26 +11,45 @@
...
+* All virtual machines (VM) configured as OpenShift nodes, must belong to a single
+resource group.
+* Microsoft Azure VM names:
+  ** Must be same as OpenShift node names, and
```
bfallonf: s/Must be same/Must be the same
s/OpenShift/{product-title}

install_config/configuring_azure.adoc
```diff
@@ -11,26 +11,45 @@
...
+  ** These names must not be included any capital letters.
```
bfallonf: With the "and" on the previous line, I don't think this needs to be two lines. So maybe:

"Microsoft Azure VMs must be named the same as {product-title} nodes and this cannot include capital letters."

install_config/configuring_azure.adoc
bfallonf: EDIT: IBM says use later and not higher for version numbers.

install_config/configuring_azure.adoc
```diff
@@ -11,26 +11,45 @@
...
+* If you plan to use  Azure Managed Disks:
+  ** {product-title} version 3.7 or later is required.
+  ** You must create VMs with Azure Managed Disks.
+* If you plan to use unmanaged disks:
+  ** You must create VMs with unmanaged disks.
```
bfallonf: Could this be one line with the above?

bfallonf:
@gaurav-nelson I made some comments :truck:

install_config/configuring_azure.adoc
gaurav-nelson: Thanks I will update it to `If you are using Microsoft Azure Disk as a persistent volume on the {product-title} version 3.5 or later, you must enable Azure Cloud Provider.` 

Also `you use "Azure Disk" above without a "Microsoft" preceding it.` nope it says `Microsoft Azure Disk` 
Since this is the first instance I am using `Microsoft Azure Disk` and `Azure Disk` for rest of the references.

install_config/configuring_azure.adoc
gaurav-nelson: Yes, I just used it to format it similar to the point above:
image

If this is not ideal I can change it to a single line.

gaurav-nelson:
Thanks @bfallonf for your comments. Addressed those in the latest commit push.

install_config/configuring_azure.adoc
```diff
@@ -11,26 +11,44 @@
...
 xref:../install_config/persistent_storage/persistent_storage_azure.adoc#install-config-persistent-storage-persistent-storage-azure[using
-Azure disk as persistent storage] for application data. After Azure is
+Azure disk as persistent storage] for application data. After Microsoft Azure is
 configured properly, some additional configurations need to be completed on the
 {product-title} hosts.
 
 [[configuring-azure-permissions]]
 == Permissions
-Configuring Azure for {product-title} requires the following role:
+Configuring Microsoft Azure for {product-title} requires the following role:
```
jamesread: Consider linking to this article that describes how to add administrator roles (eg, contributor): https://docs.microsoft.com/en-us/azure/billing/billing-add-change-azure-subscription-administrator .

install_config/configuring_azure.adoc
```diff
@@ -11,26 +11,44 @@
...
-| To create and manage all types of Azure resources.
+| To create and manage all types of Microsoft Azure resources.
 
 |===
 
+[[configuring-azure-prerequisites]]
+== Prerequisites
+
+* If you are using Microsoft Azure Disk as a persistent volume on the {product-title} version 3.5 or later, you must enable Azure Cloud Provider.
+* All virtual machines (VM) configured as OpenShift nodes, must belong to a single
```
jamesread: Consider; 

* All OpenShift node virtual machines (VMs) running in Microsoft Azure must belong to a single resource group.

install_config/configuring_azure.adoc
```diff
@@ -11,26 +11,44 @@
...
+resource group.
+* Microsoft Azure VMs must be named the same as {product-title} nodes and this
```
jamesread: I'm left a little unsure as to what actually needs to be "named". 

Would it be more easily understood if it was something like;

* The {product-title} node hostnames must match the Microsoft Azure Virtual Machine Name [because of ABC]?

jamesread (commented):
Tried to tighten up a couple of parts and make them more likely to understand on the first reading.

jamesread:
Hey, first time I've tried to review something on GitHub, I do apologize if I messed up conventions or did something incorrectly :( Let me know if I should have done something differently, and I hope the comments are useful ;)

vikram-redhat:
Thanks for the feedback @jamesread ! Much appreciated.

install_config/configuring_azure.adoc
weherdh: @gaurav-nelson If we want to enable cloud-provider for Azure, we must make sure this VM name is consist with hostname, otherwise the OCP master and node services are unable to start. We have a bug for that, tell me if you want to know that bug and I will search it for you.

install_config/configuring_azure.adoc
```diff
@@ -11,25 +11,43 @@
...
 
 |===
+For more information about adding administrator roles, see link:https://docs.microsoft.com/en-us/azure/billing/billing-add-change-azure-subscription-administrator[Add or change Azure subscription administrators].
+
+[[configuring-azure-prerequisites]]
+== Prerequisites
+
+* If you are using Microsoft Azure Disk as a persistent volume on the {product-title} version 3.5 or later, you must enable Azure Cloud Provider.
+* All {product-title} node virtual machines (VMs) running in Microsoft Azure must belong to a single resource group.
+* Microsoft Azure VMs must be named the same as {product-title} nodes and this
```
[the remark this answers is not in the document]
jamesread: Yup, I think that's fine :+1: 

I believe this is because OpenShift uses the hostname of the VM to "lookup" the VM in Azure for further configuration. I'm not quite sure where/how this happens though.

bfallonf:
@gaurav-nelson LGTM :house:

PR #10655, opened 2018-07-05
clarify binary input used for local content

bparees:


bparees:
@simo5 does this help?

simo5 (approved):
Thanks a lot, makes it a lot better
/lgtm

bparees:
@bmcelvee ptal. applicable to all releases/platforms.

kalexand-rh (commented):
Do you mean "Local binary inputs?"
Using a forward slash isn't preferred because it's often ambiguous.

kalexand-rh:
That's good to know! I really thought you meant "Local binary."  I'd change it to "Local (Binary)" or maybe  "Binary (Local)" because it avoids the and/or that's implied with the forward slash.

bparees:
>  I'd change it to "Local (Binary)" or maybe "Binary (Local)" because it avoids the and/or that's implied with the forward slash.

i considered that, but that syntax, to me at least, indicates that one is implied by the other, which is not the case here.

What is really needed is `Local (aka Binary) builds`

kalexand-rh:
Our style guide says that parentheses diminish the importance of whatever's included in them, but it doesn't imply equivalence. It's more like you're saying that the reader can ignore whatever's inside them. I think it would be less confusing than using the slash, but IDK what a real user would think. (Personally, I want to change it to "Local" and rewrite the intro paragraph of the Local Source section to make it clear that to use local source you specify a `binary` build. )

bparees:
> Personally, I want to change it to "Local" and rewrite the intro paragraph of the Local Source section to make it clear that to use local source you specify a binary build. )

i would agree but it's become clear that people looking for information on "binary builds" aren't even going to make it to that section when they skim the section titles and only see "Local" in the section title.

(conversely people today looking for information on supplying "local" content were not finding the "binary build" section and realizing it was what they wanted).

Hence my desire to have both terms in the title.  My use of "local/binary" was in fact intended to indicate equivalence/alternative.  e.g. "local or binary inputs".

kalexand-rh:
Then "Local (Binary)" or "Binary (Local)" :+1:

bparees:
updated.

PR #10700, opened 2018-07-06
Added a Known Issue per BZ1485375

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1485375

@bparees @derekwaynecarr PTAL


bparees:
looks reasonable to me but definitely not my area.

ahardin-rh:
QE verified

PR #10740, opened 2018-07-10
document imageimport pem file argument

bparees:
provides doc for the fix to https://bugzilla.redhat.com/show_bug.cgi?id=1592936

### DIFF
--- a/install_config/master_node_configuration.adoc
+++ b/install_config/master_node_configuration.adoc
@@ -660,6 +660,12 @@ permission to create Images or ImageStreamMappings via the API are not affected 
 this policy - typically only administrators or system integrations will have those
 permissions.
 
+|`*AdditionalTrustedCA*`
+|Specified a filepath to a PEM-encoded file listing additional certificate authorities
+that should be trusted during imagestream import.  This file needs to be accessible
+to the API server process.  Depending how your cluster is installed, this may require
+mounting the file into the API server pod.
+
 |`*InternalRegistryHostname*`
 |Sets the hostname for the default internal image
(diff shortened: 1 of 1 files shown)


bparees:
@bmcelvee ptal.  3.11 and above only.

bmcelvee:
Thanks, @bparees! LGTM

bmcelvee:
Thanks, @bparees! I'm sorry about that, I'll get it merged today.

bmcelvee:
Minor change. Does not require QE.

xingxingxia:
Now I read this but seems didn't see this doc in [3.11 doc](https://docs.openshift.com/container-platform/3.11/install_config/master_node_configuration.html#master-config-image-policy-config) ? Anything wrong? (I come here pointed to by [1592936#c29](https://bugzilla.redhat.com/show_bug.cgi?id=1592936#c29))

bparees:
@xingxingxia good question.. @bmcelvee ?   

it is in the origin docs:
https://docs.okd.io/latest/install_config/master_node_configuration.html#master-config-image-policy-config

Note that we've actually addressed this a different way anyway, the api server is going to start mounting the /etc/pki from the host node, so you should not need to use this setting.

bmcelvee:
Looks like this one [somehow] didn't get pulled into the 3.11 branch when it was created. I'm sorry about that.

@bparees Even though it's addressed in a different way, I can still cherry-pick the parameter + description into 3.11. Is that okay with you?

bparees:
@bmcelvee that's fine.

bmcelvee:
Thanks! All set.

PR #10758, opened 2018-07-11
[enterprise-3.6]Fix incorrect example for downward API

kalexand-rh:
From https://github.com/openshift/openshift-docs/pull/10130

@mdshuai, I can't find the relevant Kubernetes docs to confirm that the changes starting at line 190 are correct in version 3.6 and earlier. Will you PTAL?

### DIFF
--- a/dev_guide/downward_api.adoc
+++ b/dev_guide/downward_api.adoc
@@ -144,11 +144,11 @@ configure this volume. The plug-in supports the following fields:
 spec:
   volumes:
     - name: podinfo
-      metadata: <1>
-        items:  <2>
-          - name: "labels" <3>
-            fieldRef:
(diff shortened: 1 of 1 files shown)


kalexand-rh:
@mdshuai, I can't find the relevant Kubernetes docs to confirm that the changes starting at line 190 are correct in version 3.6 and earlier. Will you PTAL?

vikram-redhat:
@mdshuai - PTAL.

cc: @xltian

wjiangjay:
Checked with:
```bash
# openshift version 
openshift v3.6.173.0.128
kubernetes v1.6.1+5115d708d7
etcd 3.2.1
```

And work well, LGTM

PR #10801, opened 2018-07-13
[enterprise-3.6] Update Ceph RBD dynamic example

vikram-redhat:


vikram-redhat:
@tmorriso-rh - please confirm if this does apply to 3.6 as well and if it doesn't, lets roll it back.

tmorriso-rh:
@vikram-redhat The bug was filed against 3.6, so I think it applies.

PR #10818, opened 2018-07-13
Added a section on Changing the Web Console URL Port and Certificates

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1593763

@spadgett @rjhowe PTAL

### DIFF
--- a/install_config/web_console_customization.adoc
+++ b/install_config/web_console_customization.adoc
@@ -11,7 +11,6 @@
 
 toc::[]
 
-ifdef::openshift-origin,openshift-enterprise[]
 == Overview
 
 Administrators can customize the
@@ -232,7 +231,7 @@ xref:loading-custom-scripts-and-stylesheets[Loading Extension Scripts and Styles
 [[changing-the-membership-whitelist]]
 == Customizing the Membership Whitelist
 
-The default whitelist in the membership page shows a subset of cluster roles, such as 
+The default whitelist in the membership page shows a subset of cluster roles, such as
 `admin`, `basic-user`, `edit`, and so on. It also shows custom roles defined within a project.
 
 For example, to add your own set of custom cluster roles to the whitelist:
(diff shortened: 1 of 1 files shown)


install_config/web_console_customization.adoc
```diff
@@ -896,3 +895,28 @@ xref:../install/index.adoc#install-planning[inventory file]:
...
+----
+
+To ensure users are correctly redirected, update the `PublicUrls` for the
+`openshift-web-console configmap`:
+
+----
+$ oc edit configmap/webconsole-config -n openshift-web-console
+----
+
+Update the values for `consolePublicURL` and `masterPublicURL`.
```
spadgett: It's unclear to me if you're changing both master public URL or just console URL. There are probably more changes needed if changing master public URL. If you haven't changed master URL, the `masterPublicURL` property in the config map doesn't need to be updated.

Do we have any existing topics that cover changing master public URL?

spadgett (commented):
cc @deads2k

install_config/web_console_customization.adoc
ahardin-rh: @spadgett I don't see much mentioned about changing the master public URL. It's only mentioned briefly in places like https://docs.openshift.com/container-platform/3.9/install_config/master_node_configuration.html#master-config-oath-authentication-config and https://docs.openshift.com/container-platform/3.9/install_config/master_node_configuration.html#master-config-asset-config

ahardin-rh: @spadgett @deads2k  Any more thoughts?

ahardin-rh: @spadgett @deads2k Bump

ahardin-rh: @deads2k Okay, thanks! This is updated.

ahardin-rh:
@yanpzhan Please provide QE review.
@openshift/team-documentation Please peer review.

kalexand-rh (commented):
Couple of questions and suggestions

install_config/web_console_customization.adoc
```diff
@@ -11,7 +11,6 @@
 
 toc::[]
 
-ifdef::openshift-origin,openshift-enterprise[]
```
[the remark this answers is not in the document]
ahardin-rh: Yeah, not sure why this ifdef (with no endif) was in the doc. This cleans it out. The _topic_map.yml already defines that this topic should be in Origin and OCP docs only, so this like is unnecessary.

rjhowe:
With the new web console being in a pod, in the master-config.yaml the values for masterPublicURL oauthConfig.assetPublicURL and oauthConfig.masterPublicURL need to a the same and the namedcertificate needs to configured for these too, as we get redirect to these URLs for paths /oauth/authorize and /console/oauth.

install_config/web_console_customization.adoc
```diff
@@ -922,3 +921,28 @@ xref:../install/index.adoc#install-planning[inventory file]:
 # See: https://docs.openshift.com/enterprise/latest/install_config/aggregate_logging.html
 #openshift_master_logging_public_url=https://kibana.example.com
 ----
+
+[[changing-the-web-console-url-port-and-certificates]]
+== Changing the Web Console URL Port and Certificates
+
+To ensure your custom certificate is served when users access the web console
+URL, add the certificate and URL to the `NamedCertificates` section of the
```
DanyC97: typo `NamedCertificates` vs `namedCertificates` ?

 as per the [code](https://github.com/openshift/openshift-ansible/blob/006d82982d1dd3117bf9057ae6172a9d16380e0c/roles/openshift_control_plane/templates/master.yaml.v1.j2#L193)

DanyC97 (changes requested):

PR #10828, opened 2018-07-14
[enterprise-3.10] Ephemeral Storage Fixes

vikram-redhat:
Separating this into two commits to make sure that the anchor fixes are applied back to 3.10.


vikram-redhat:
@tmorriso-rh - please make sure to fetch and rebase before you do the followup.

PR #5093, opened 2017-08-26
routes: Fix conditionals for Online

Miciah:
Fix an `ifndef` that is causing "DNS wildcard entry" to be rendered twice, once plain and once hyperlinked.

Conditionalize two sentences that had conditionalized parts for openshift-origin, openshift-enterprise, and openshift-dedicated but nothing for openshift-online, which is producing incomplete sentences in the Online documentation.

Fix conditionals in another sentence with conditionalized parts for openshift-origin, openshift-enterprise, and openshift-dedicated to use the openshift-dedicated text for openshift-online in order to avoid having an incomplete sentence in the Online documentation.

---

@adellape, the first change reverts a change you made in https://github.com/openshift/openshift-docs/pull/4388.  I'm confused why that change was made, but in any case, what we have now is not rendering correctly on https://docs.openshift.com/online/architecture/core_concepts/routes.html.

PR #10853, opened 2018-07-17
PR5093 Followup

vikram-redhat:
Follow up to: https://github.com/openshift/openshift-docs/pull/5093/files


vikram-redhat:
Only picking this for 3.9 and 3.10 branches as the fixes were for online distro.

PR #10864, opened 2018-07-17
Fix flexvolume docs

gnufied:
Fixes flexvolume docs as per latest configuration.

This PR applies to 3.10, 3.11 and master.

cc @tmorriso-rh


tmorriso-rh:
Thanks @gnufied. I am going to request that QE review this so that we can get it merged. After merging, I can create a follow-up PR to change "Flexvolume" to "FlexVolume" and add a few punctuation marks.

@jhou1 @liangxia Please provide QE review. Thanks.

PR #10879, opened 2018-07-18
add some warnings around changing Pod Templates from both Jenkins con…

gabemontero:
…sole and OpenShift resources monitored by sync plugin

Fixes https://github.com/openshift/jenkins-sync-plugin/issues/246

@bparees @bmcelvee ptal

### DIFF
--- a/using_images/other_images/jenkins.adoc
+++ b/using_images/other_images/jenkins.adoc
@@ -435,6 +435,16 @@ control the label field of the Kubernetes plug-in pod template by setting an
 annotation on the image stream or image stream tag object with the key
 `slave-label`. Otherwise, the name is used as the label.
 
+[NOTE]
+====
+If you log into the Jenkins console and make further changes to the Pod Template configuration
+after the Pod Template is created, and the OpenShift Sync plug-in detects that the image
+associated with the ImageStream or ImageStreamTag has changed, it will replace the Pod
+Template and overwrite those configuration changes. No merging of configuration will occur.
+
+Consider the ConfigMap approach if you have more complex configuration needs.
+====
+
 When it finds a ConfigMap with the appropriate label, it assumes that any
 values in the key-value data payload of the ConfigMap contains XML consistent
 with the config format for Jenkins and the Kubernetes plug-in pod templates. A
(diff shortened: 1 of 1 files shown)


using_images/other_images/jenkins.adoc
```diff
@@ -435,6 +435,18 @@ control the label field of the Kubernetes plug-in pod template by setting an
 annotation on the image stream or image stream tag object with the key
 `slave-label`. Otherwise, the name is used as the label.
 
+[NOTE]
+====
+Keep in mind that after the Pod Template is created, if you log into the Jenkins
```
bparees: s/Keep in mind that/  (redundant since it's already going to be marked as a note)

using_images/other_images/jenkins.adoc
```diff
@@ -435,6 +435,18 @@ control the label field of the Kubernetes plug-in pod template by setting an
...
+console and make further changes to the Pod Template configuration, but later
+on the OpenShift Sync plug-in detects that the image associated with the
```
bparees: s/on//

using_images/other_images/jenkins.adoc
```diff
@@ -487,6 +499,15 @@ data:
...
 ----
 
+[NOTE]
+====
+Keep in mind that after the Pod Template is created, if you log into the Jenkins
+console and make further changes to the Pod Template configuration, but later
+on the OpenShift Sync plug-in detects that the ConfigMap has changed in any way,
+it will replace the Pod Template and overwrite those configuration changes.  No
+merging of configuration will occur.
+====
```
bparees: same edits here

bparees (changes requested):

using_images/other_images/jenkins.adoc
```diff
@@ -435,6 +435,18 @@ control the label field of the Kubernetes plug-in pod template by setting an
 annotation on the image stream or image stream tag object with the key
 `slave-label`. Otherwise, the name is used as the label.
 
+[NOTE]
```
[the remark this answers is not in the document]
gabemontero: yes it does

bparees (approved):

gabemontero:
@bmcelvee - while the function around scanning config maps and image streams dates back to 3.6, the behavior as pecisely worded only applies to 3.11

Warnings are applicable to the earlier versions, but the check to replace what is in Jenkins is less stringent ... basically, changes of any sort to the IS or CM on the OpenShift side will cause a replacement of what exists on the Jenkins side.

If you want to backport a modified version to the earlier releases, that is fine, but given the activity with this, I'm not overly concerned if it does not get backported.  I'll defer to you here.

bmcelvee:
Let's do 3.11. 
Since this is a minor change, I don't think it needs QE.

PR #10885, opened 2018-07-18
Correcting the Content-Type header for Patching a BuildConfig

TheDiemer:
Currently the documentation indicates that the `Content-Type` header should be `application/json-patch+json` for Patching a BC, but it really should be `application/merge-patch+json`.

https://bugzilla.redhat.com/show_bug.cgi?id=1602912


vikram-redhat:
Thanks @TheDiemer - assigned the bug to @kalexand-rh .

kalexand-rh:
@wzheng1, will you PTAL?

@TheDiemer, you don't have a version specified. Which versions does this change apply to? (@wzheng1, do you know?)

vikram-redhat:
@TheDiemer - Hey Eric, looking at the raw output swagger for the API (big file ~5MB): https://raw.githubusercontent.com/openshift/origin/master/api/swagger-spec/openshift-openapi-spec.json, I see that both content type headers are still valid (search for "Partially update the specified BuildConfig"). WDYT?

wzheng1:
Sorry, I am not clear which version starts to have this change, @bparees Could you please confirm? What we can confirm is 'Content-Type: application/merge-patch+json' works for OCP 3.10, but 'Content-Type: application/merge-patch+json' doesn't.

bparees:
I don't think this is specific to buildconfigs.  hopefully @openshift/sig-master can tell us when/if the expected/allowed content-type changed.

also @wzheng1 you cited the same content-type both times....  can you restate which one works in 3.10 and which does not?

liggitt:
there are three valid patch content types:

```
JSONPatchType           PatchType = "application/json-patch+json"
MergePatchType          PatchType = "application/merge-patch+json"
StrategicMergePatchType PatchType = "application/strategic-merge-patch+json"
```

the first expects a json patch, which requires an array of patch operations:
```
oc patch ns default --type=json -p='[{"op": "add", "path": "/metadata/annotations", "value": {"test":"test1"}}]'
```

the second expects a merge patch, which overlays the provided json with the existing object:
```
oc patch ns default --type=json -p='{"metadata":{"annotations": {"test":"test2"}}}'
```

the third expects a kubernetes-specific patch format which resembles a merge patch but accepts additional complex directives for merging lists of keyed items.

wzheng1:
@bparees Sorry for the typo. I mean 'Content-Type: application/json-patch+json' not working

wzheng1:
According to @liggitt seems the command format we use is incorrect, let me re-confirm.

kalexand-rh:
@wzheng1, do you have an update on this change?

wzheng1:
Sorry the late notice, both works as below:
$ oc patch ns wzheng123 --type=merge -p='{"metadata":{"annotations": {"test":"test2"}}}'
namespace "wzheng123" patched
$ oc patch ns wzheng123 --type=json -p='[{"op": "add", "path": "/metadata/annotations", "value": {"test":"test1"}}]'
namespace "wzheng123" patched
$ oc version
oc v3.10.14
kubernetes v1.10.0+b81c8f8
features: Basic-Auth GSSAPI Kerberos SPNEGO

PR #10901, opened 2018-07-19
bug 1600173 updating self-provisioner clusterrole name

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1600173

@stuartchuan, will you PTAL?

### DIFF
--- a/admin_guide/managing_projects.adoc
+++ b/admin_guide/managing_projects.adoc
@@ -116,9 +116,7 @@ Subjects:
   Group	system:authenticated:oauth
 ----
 
-. Remove the `self-provisioners`
-xref:../architecture/additional_concepts/authorization.adoc#roles[cluster role binding]
-from the group `system:authenticated:oauth`.
+. Remove the `self-provisioner` cluster role from the group `system:authenticated:oauth`. 
 ** If the `self-provisioners` cluster role binding binds only the
 `self-provisioner` role to the `system:authenticated:oauth` group, run the
 following command:


stuartchuan:
There were never `self-provisioners` cluster role in 3.9, I also checked 3.7 and 3.6 still don't have such cluster role.
```
# oc get clusterrole | grep self-provisioner
self-provisioner
# oc get clusterrolebinding | grep self-provisioner
self-provisioner                                                      /self-provisioner                                                                                                                                      management-infra/management-admin                                                  
self-provisioners                                                     /self-provisioner                                                                                       system:authenticated:oauth                                                                                                        
# openshift version
openshift v3.9.33
kubernetes v1.9.1+a0ce1bc657
etcd 3.2.16
```
Reviewed again 'Disabling Self-provisioning' part, I don't think we need this changes to apply. Only one confused point, will comment in line.

admin_guide/managing_projects.adoc
```diff
@@ -116,18 +116,18 @@ Subjects:
   Group	system:authenticated:oauth
 ----
 
-. Remove the `self-provisioners`
+. Remove the `self-provisioner`
 xref:../architecture/additional_concepts/authorization.adoc#roles[cluster role binding]
 from the group `system:authenticated:oauth`.
```
stuartchuan: There is the only confused point, should changed to `3.Remove the self-provisioner cluster role from the group system:authenticated:oauth.`

admin_guide/manage_rbac.adoc
```diff
@@ -403,7 +403,7 @@ Subjects:
   Group	system:unauthenticated
 
 
-Name:		self-provisioners
+Name:		self-provisioner
```
kalexand-rh: @stuartchuan, based on the command output that you posted, this change is valid:
```
# oc get clusterrole | grep self-provisioner
self-provisioner
````

admin_guide/managing_projects.adoc
kalexand-rh: Got it. Thank you @stuartchuan!

stuartchuan:
Thank you for the updated, it's LGTM now.

kalexand-rh:
I don't think this change needs peer review. Merging.

PR #10914, opened 2018-07-19
Pod Anti Affinity Docs are inconsistent

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1600441

### DIFF
--- a/admin_guide/manage_nodes.adoc
+++ b/admin_guide/manage_nodes.adoc
@@ -415,32 +415,34 @@ anti-affinity can be set to either required or preferred.
 
 Using the `*docker-registry*` pod as an example, the first step in enabling
 this feature is to set the `*scheduler.alpha.kubernetes.io/affinity*` on the
-pod. Since this pod uses a deployment configuration, the most appropriate
-place to add the annotation is to the pod template's metadata.
+pod. 
 
+[source,yaml]
 ----
-$ oc edit dc/docker-registry -o yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: with-pod-antiaffinity
+spec:
+  affinity:
+    podAntiAffinity: <1>
+      preferredDuringSchedulingIgnoredDuringExecution: <2>
+      - weight: 100 <3>
+        podAffinityTerm:
+          labelSelector:
+            matchExpressions:
+            - key: docker-registry <4>
+              operator: In <5>
+              values: 
+              - default
+          topologyKey: kubernetes.io/hostname
+----
 
-...
-  template:
-    metadata:
-      annotations:
-        scheduler.alpha.kubernetes.io/affinity: |
-          {
-            "podAntiAffinity": {
-              "requiredDuringSchedulingIgnoredDuringExecution": [{
-                "labelSelector": {
-                  "matchExpressions": [{
-                    "key": "docker-registry",
-                    "operator": "In",
-                    "values":["default"]
-                  }]
-                },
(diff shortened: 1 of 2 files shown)


mburke5678:
@aveshagarwal 
Removed the DC example from Manage Nodes and added Pod Config from Pod Affinity topic. Is this change appropriate?

aveshagarwal:
@mburke5678 lgtm.

mburke5678:
@mdshuai Please take a look at these changes.

mdshuai:
> @mdshuai Please take a look at these changes.

@wjiangjay will help review it.

wjiangjay:
LGTM

mburke5678:
@aveshagarwal From your comment in the BZ: "The above pod anti-affinity configuration is not valid since 3.5/3.6." Should this change go into 3.5+ or 3.6+?
It appears that there was a change in 3.6. based on the release notes.
https://docs.openshift.com/container-platform/3.6/release_notes/ocp_3_6_release_notes.html#ocp-36-pod-affinity-and-anti-affinity

mburke5678:

admin_guide/manage_nodes.adoc
```diff
@@ -415,32 +415,34 @@ anti-affinity can be set to either required or preferred.
...
+            - key: docker-registry <4>
+              operator: In <5>
+              values: 
+              - default
+          topologyKey: kubernetes.io/hostname
+----
+
+<1> Stanza to configure pod anti-affinity.
+<2> Defines a preferred rule.
+<3> Specifies a weight for a preferred rule. The node that with highest weight is preferred.
```
kalexand-rh: "The node that with highest weight is preferred." doesn't make sense.
I haven't read any of the context, but I think it should say something like "The pod runs on the node with the highest weight" or "The pod applies the rule with the largest weight."

admin_guide/manage_nodes.adoc
```diff
@@ -415,32 +415,34 @@ anti-affinity can be set to either required or preferred.
...
+<4> The key and value (label) that must be matched to apply the rule.
```
kalexand-rh: This annotation confused me until I read annotation 5. I think it should be something like "Description of the pod label that determines when the anti-affinity rule applies. Specify a key and value for the label."
( also might attach it to the line that contains matchExpression or give annotations for both the key and value lines.)

admin_guide/manage_nodes.adoc
```diff
@@ -415,32 +415,34 @@ anti-affinity can be set to either required or preferred.
...
+<5> The operator represents the relationship between the label on the existing pod and the set of values in the `matchExpression` parameters in the specification for the new pod.  Can be `In`, `NotIn`, `Exists`, or `DoesNotExist`.
```
kalexand-rh: There's an extra space before "Can."
s/Can be/Specify

kalexand-rh (commented):
Couple of requests and suggestions. :stars:

admin_guide/manage_nodes.adoc
mburke5678: @kalexand-rh I think the meaning is right. Rules can be required to be followed or followed unless it is not possible (preferred). This setting is only for a preferred rule. If the wording can be better, let me know!

kalexand-rh: After our chat, I think we should revisit this after 3.10

s/that with/with the

PR #10927, opened 2018-07-19
Add services per ns cluster limits for ocp v3.10

chaitanyaenr:
The current services per ns limit is 5000 which is also the limit
in case of upstream kubernetes.


ahardin-rh:
@chaitanyaenr Thanks for this!

PR #10940, opened 2018-07-20
Fix for not sending requestes from master builds

gaurav-nelson:
Check and send API requests only if Travis build is a build for PR.


gaurav-nelson:
@vikram-redhat small fix for preview builds.

PR #10942, opened 2018-07-20
dont send webhooks for merges

gaurav-nelson:
Includes following updates:
- Separate script for sending preview generation request.
- Extra checks in place before sending preview generation request.
- Info messages including why a preview generation request is sent or not sent.

#### What is problematic with the current preview generation process?
Current process works as follows:
1. For every Github PR or master merge send webhook for preview generation.
2. Preview generator then checks for allowed user and builds a preview.

This is resulting in error messages on preview generator which I am subscribed to. The proposed updates will solve all these known errors.

#### What is the suggested flow in this update?
This makes sure that the webhook request is only sent if:
- Travis build is running for a PR (this saves sending request on merges)
- AND PR is from a member of openshift-team-docs (avoid building PRs for every pull request)
- AND Changed files includes at least one .adoc file (avoid build previews for configuration changes)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,4 @@ script:
  - python makeBuild.py
 
 after_success:
- - export USERNAME=${TRAVIS_PULL_REQUEST_SLUG::-15}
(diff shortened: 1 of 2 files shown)


vikram-redhat:
@gaurav-nelson - do you want me to merge it? What's different in this one?

PR #10978, opened 2018-07-20
remove feature gate docs for huge pages

jeremyeder:
hugepages are no longer behind a feature gate.  they are enabled by default as of 3.10.  so we can remove the feature gate instructions.


vikram-redhat:
Thanks @jeremyeder. This looks like a simple enough change. I will merge it and CP to 3.10.

@ahardin-rh @aheslin

ahardin-rh:
@jeremyeder Thanks for this!

PR #10980, opened 2018-07-21
Update name change for CNS and CRS

ghost:
Per Marketing:

- CNS mode will now be referred to as converged mode.
- CRS mode will now be referred to as independent mode.

### DIFF
--- a/_snippets/glusterfs.adoc
+++ b/_snippets/glusterfs.adoc
@@ -19,8 +19,8 @@ How to use this file:
 :gluster-role-link: https://github.com/openshift/openshift-ansible/tree/master/roles/openshift_storage_glusterfs
 ifdef::openshift-enterprise[]
 :gluster: Red Hat Gluster Storage
-:gluster-native: Container-Native Storage
-:gluster-external: Container-Ready Storage
+:gluster-native: converged mode
+:gluster-external: independent mode
 :gluster-install-link: https://access.redhat.com/documentation/en-us/red_hat_gluster_storage/3.3/html/installation_guide/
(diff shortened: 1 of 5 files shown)


tmorriso-rh:
@SaravanaStorageNetwork and @netzzer PTAL to determine if the changes made due the name change make sense.

Cc: @vikram-redhat

vikram-redhat:
@tmorriso-rh this is missing the rest of the changes to the scaling performance guide which didn't have the attribute: https://github.com/tmorriso-rh/openshift-docs/commit/771d7362a0b3c1ae113e1c9694d4cd9ee9ab6cf3#diff-f368171fd950e046a50929f2746793a4

tmorriso-rh:
@vikram-redhat Yes, I still need to add those changes. What about the 3.7 and 3.9 release notes? They don't have the attribute. Do those need to be updated since the name change is for 3.10?

SaravanaStorageNetwork:
I feel both converged and independent mode are misnomer.
Container Native Storage(CNS) conveys better meaning than converged, same  for independent mode.

vikram-redhat:
@SaravanaStorageNetwork - Hey there. This change has come to us from the product marketing and content strategy teams of Gluster product. Can you please take up the issue of naming with them and just provide us the QE review?

anjsupsri:
@SaravanaStorageNetwork, I agree with Vikram. The new rebranded terms for the deployment modes are as follows:
• CNS mode will now be referred to as 'converged mode’
• CRS mode will now be referred to as ‘independent mode'

vikram-redhat:
@tmorriso-rh I have got the ack on this from both Anjana and Ken. Please go ahead and merge and CP.

PR #11044, opened 2018-07-24
bug 1592772 clarifying when to uninstall instead of redeploy

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1592772

I won't be able to use the cherry-pick bot on this one because the file structure is different in 3.9, and the playbook names are different in 3.7.

@sdodson, @jiajliu, will you PTAL?

### DIFF
--- a/install/running_install.adoc
+++ b/install/running_install.adoc
@@ -154,13 +154,14 @@ system can run out of files to open and the playbook will fail.
 
 To run the RPM-based installer:
 
-. Run the *_prequisites.yml_* playbook. This must be run only once before
-deploying a new cluster. Use the following command, specifying `-i` if your
-inventory file located somewhere other than *_/etc/ansible/hosts_*:
+. Run the *_prequisites.yml_* playbook. This playbook installs required software
+packages, if any, and modifies the container runtimes. Unless you need to
+configure the container runtimes, run this playbook only once, before you
+deploy a cluster the first time:
 +
 ----
 ifdef::openshift-enterprise[]
-# ansible-playbook [-i /path/to/inventory] \
+# ansible-playbook [-i /path/to/inventory] \ <1>
     /usr/share/ansible/openshift-ansible/playbooks/prerequisites.yml
 endif::[]
 ifdef::openshift-origin[]
@@ -168,6 +169,8 @@ ifdef::openshift-origin[]
     ~/openshift-ansible/playbooks/prerequisites.yml
 endif::[]
 ----
+<1>  If your inventory file is not in the *_/etc/ansible/hosts_* directory, 
+specify `-i` and the path to the inventory file.
 
 . Run the *_deploy_cluster.yml_* playbook to initiate the cluster installation:
 +
@@ -406,19 +409,24 @@ endif::[]
 The main installation playbook *_{pb-prefix}playbooks/deploy_cluster.yml_* runs a
 set of individual component playbooks in a specific order, and the installer
 reports back at the end what phases you have gone through. If the installation
-fails during a phase, you are notified on the screen along with the errors from
+fails, you are notified which phase failed along with the errors from
 the Ansible run.
 
-After you resolve the issue, rather than run the entire installation over again,
-you can pick up from the failed phase. You must then run each of the remaining
-playbooks in order:
+After you resolve the errors, you can continue installation:
+
+* You can run the remaining individual installation playbooks.
+* If you are installing in a new environment, you can run the
+*_deploy_cluster.yml_* playbook again.
+
+If you want to run only the remaining playbooks, start by running the playbook
+for the phase that failed and then run each of the remaining playbooks in
+order:
 
 ----
 # ansible-playbook [-i /path/to/inventory] <playbook_file_location>
 ----
 
-The following table is sorted in order of when each individual component
-playbook is run:
+The following table is lists the playbooks in the order that they must run:
 
 .Individual Component Playbook Run Order
 [options="header",cols="1,3"]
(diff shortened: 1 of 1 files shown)


install/running_install.adoc
```diff
@@ -154,9 +154,11 @@ system can run out of files to open and the playbook will fail.
...
 To run the RPM-based installer:
 
-. Run the *_prequisites.yml_* playbook. This must be run only once before
-deploying a new cluster. Use the following command, specifying `-i` if your
-inventory file located somewhere other than *_/etc/ansible/hosts_*:
+. Run the *_prequisites.yml_* playbook. This playbook modifies the container
+runtimes and installs required software packages, if any. Unless you need to
+modify the container runtimes, run this playbook only one time, before you
+deploy a cluster the first time. Use the following command, specifying `-i`
+if your inventory file is not in the *_/etc/ansible/hosts_* directory:
```
ganhuang: `/etc/ansible/hosts` should be a file path, not a directory. The original statement is more accurate.

install/running_install.adoc
```diff
@@ -154,9 +154,11 @@ system can run out of files to open and the playbook will fail.
 
 To run the RPM-based installer:
 
-. Run the *_prequisites.yml_* playbook. This must be run only once before
-deploying a new cluster. Use the following command, specifying `-i` if your
-inventory file located somewhere other than *_/etc/ansible/hosts_*:
+. Run the *_prequisites.yml_* playbook. This playbook modifies the container
+runtimes and installs required software packages, if any. Unless you need to
```
ganhuang: The playbook is used to to install required software packages first, then configure the container runtime accordingly. I personally think it would be better to adjust the order. And I prefer to use `configures the container runtimes` instead of `modifies the container runtimes`. 

Not a big deal anyway.

install/running_install.adoc
```diff
@@ -655,9 +658,12 @@ intended. However, this is a transient event and the system does correct itself
 over time. See https://github.com/kubernetes/kubernetes/issues/10030 for
 details.
 
-- On failure of the Ansible installer, you must start from a clean operating
-system installation. If you are using virtual machines, start from a fresh
-image. If you are using bare metal machines, see
+- If the Ansible installer fails and you did not need to change a host name or
+generate new certificates, run the *_deploy_cluster.yml_*
+playbook again. If you changed a host name, generated new certificates, or the
+installer fails again, start over with a clean operating system installation.
```
ganhuang: In addition to start over with a clean operating system installation, it would also be feasible to run the uninstall playbook.

install/running_install.adoc
```diff
@@ -655,9 +658,12 @@ intended. However, this is a transient event and the system does correct itself
 over time. See https://github.com/kubernetes/kubernetes/issues/10030 for
 details.
 
-- On failure of the Ansible installer, you must start from a clean operating
-system installation. If you are using virtual machines, start from a fresh
-image. If you are using bare metal machines, see
+- If the Ansible installer fails and you did not need to change a host name or
```
shlao: "change a host name" is the small part of "sdn configuration", so I think this maybe not accurately.
IMO, It is better to write in this: "If you did not need to change the configuration of SDN"

shlao: line 663, need to modify in the same way.

install/running_install.adoc
```diff
@@ -406,19 +408,20 @@ endif::[]
...
 set of individual component playbooks in a specific order, and the installer
 reports back at the end what phases you have gone through. If the installation
-fails during a phase, you are notified on the screen along with the errors from
+fails, you are notified which phase failed along with the errors from
 the Ansible run.
 
-After you resolve the issue, rather than run the entire installation over again,
-you can pick up from the failed phase. You must then run each of the remaining
-playbooks in order:
+After you resolve the errors, you can either rerun the *_deploy_cluster.yml_*
```
shlao: "you can either rerun the ... ", 
As Scott says, If it is deployed in a new environment, that is ok to re-run the thedeploy_cluster.yml
But if the cluster is already up and it has workload, It is better to  resolve the current errors, then run each of the remaining playbooks in order.

kalexand-rh: @shlao, if you had current workloads, how often would you be installing instead of upgrading?

kalexand-rh:
@ganhuang, @shlao, I made some updates based on your comments. Will you PTAL? Thanks!

install/running_install.adoc
shlao: I don't see the scenario: "it has workload",  @sdodson  Maybe meet this before. 
But add the the prerequisite: “deployed in a new environment“ to "re-run deploy_cluster.yml" is correctly.

kalexand-rh: What's the difference between "a new environment" and "an environment without workloads?"

shlao: "environment without workload“ it means that there exist some idle programs to wait the request.
"a new environment" it is ,common, means that there is nothing except the basic process of OS.

shlao: "environment without workload", for example, the Bank System, in early morning, e.g 2:00 AM
the system without workload, no transactions from the clients.
Beside, the means of the word "environment", include: HW, Software, network and so on.

kalexand-rh: Thanks @shlao! Is this better?

ganhuang:
thanks, lgtm.

kalexand-rh:
@openshift/team-documentation PTAL :star2:

bmcelvee:
LGTM :sun_with_face:

install/running_install.adoc
```diff
@@ -154,20 +154,23 @@ system can run out of files to open and the playbook will fail.
 
 To run the RPM-based installer:
 
-. Run the *_prequisites.yml_* playbook. This must be run only once before
-deploying a new cluster. Use the following command, specifying `-i` if your
-inventory file located somewhere other than *_/etc/ansible/hosts_*:
+. Run the *_prequisites.yml_* playbook. This playbook installs required software
+packages, if any, and modifies the container runtimes. Unless you need to
+configure the container runtimes, run this playbook only one time, before you
```
mburke5678: @kalexand-rh Do we need to explain why the user might need to configure the container runtimes?

install/running_install.adoc
```diff
@@ -154,20 +154,23 @@ system can run out of files to open and the playbook will fail.
...
+deploy a cluster the first time. Run the following command:
```
mburke5678: @kalexand-rh 
Could these two sentences be combined? The _run this playbook_/_run this command_ seems a bit awkward. 
_...run this playbook only one time, before you deploy a cluster the first time using  the following command:_
Or, would that make the sentence too long?

install/running_install.adoc
kalexand-rh: @mburke5678, not unless we usually go into detail about why they'd need to change their Docker settings.

install/running_install.adoc
kalexand-rh: I don't think it's too long, but instead of awkwardness from repetition you get awkwardness from rambling.
I might move the restriction to a note under the command.

sdodson:
@kalexand-rh Late to review but this looks good, thanks.

kalexand-rh:
Thanks, @sdodson! (I would have opened a new PR if you'd found changes.)

PR #11048, opened 2018-07-24
Updates to cluster limits

ahardin-rh:
Updates are per our discussion in today's planning meeting.


ahardin-rh:
@tkatarki @jeremyeder PTAL 

cc @aheslin

PR #11079, opened 2018-07-25
Fixed syntax error and add required fields.

jkroepke:
Since apps/v1, `spec.selector` is required. See https://kubernetes.io/docs/concepts/workloads/controllers/deployment/


vikram-redhat:
@zhouying7780 or @geliu2016  - could you please confirm if this change is acceptable and from what version?

zhouying7780:
Hi, Vikram:

    We support this from the OCP3.3. 





Best Regards,
Ying Zhou.

----- Original Message -----

zhouying7780:
Hi, Vikram:

    We support this from the OCP3.9.

vikram-redhat:
Since this is a small fix, I am merging it to 3.9, 3.10 and 3.11.

PR #11111, opened 2018-07-26
[enterprise-3.10] Go client documentation

chrisnegus:
Cherry picking base go-client documentation into 3.10.


kalexand-rh:
@vikram-redhat, we're moving the existing go-client docs to 3.10 so Chris's changes against master will land correctly.

vikram-redhat:
@kalexand-rh - Was this CPed to 3.10?

PR #11153, opened 2018-07-27
removing a note about tokens and GitLab

### LINKED ISSUE #11144: Remove note about GitLab tokens
per @enj, https://docs.openshift.org/latest/install_config/configuring_authentication.html#GitLab is wrong, the `Using GitLab as an identity provider requires users to get a token using <master>/oauth/token/request to use with command-line tools.` bit does not apply to gitlab

kalexand-rh:
Fixes #11144

@enj, FYI

@stuartchuan, does this sound ok to you?


stuartchuan:
@kalexand-rh yes, it's lgtm.

PR #11161, opened 2018-07-27
MaxLimitRequestsRatio should be positive integral value

ravisantoshgudimetla:
This is for fixing https://bugzilla.redhat.com/show_bug.cgi?id=1597297

/cc @sjenning @mburke5678

### DIFF
--- a/admin_guide/limits.adoc
+++ b/admin_guide/limits.adoc
@@ -194,7 +194,7 @@ CPU constraint.
 If a configuration defines a `maxLimitRequestRatio` value, then any new
 containers must have both a request and limit value. Additionally,
 {product-title} calculates a limit to request ratio by dividing the limit by the
-request.
+request. This value should be a non-negative integer greater than 1.
 
 For example, if a container has `cpu: 500` in the `limit` value, and
 `cpu: 100` in the `request` value, then its limit to request ratio for `cpu` is


admin_guide/limits.adoc
```diff
@@ -194,7 +194,7 @@ CPU constraint.
 If a configuration defines a `maxLimitRequestRatio` value, then any new
 containers must have both a request and limit value. Additionally,
 {product-title} calculates a limit to request ratio by dividing the limit by the
-request.
+request. Please note that this value should be a non-negative integral value greater than 1.
```
mburke5678: @ravisantoshgudimetla Do we need _integral_ here? _...a non-negative value greater than 1._

ravisantoshgudimetla: Yes integral is needed. Please look at bug.

mburke5678: @ravisantoshgudimetla I was uncomfortable with _intregal_ because the more common definition of the word is not related to integers.
But, a quick Google search seems to show that _non-negative integral value_ is a common construct. 
I see _value_ used twice in the sentence now. How about:
_Please note that this value should be a non-negative integer greater than 1._

mburke5678:
@mdshuai PTAL

admin_guide/limits.adoc
```diff
@@ -194,7 +194,7 @@ CPU constraint.
 If a configuration defines a `maxLimitRequestRatio` value, then any new
 containers must have both a request and limit value. Additionally,
 {product-title} calculates a limit to request ratio by dividing the limit by the
-request.
+request. Please note that this value should be a non-negative integer greater than 1.
```
sjenning: remove "Please note".  This is documentation; it's implied :stuck_out_tongue:

sjenning:
lgtm

kalexand-rh:
@wjiangjay, will you PTAL?

wjiangjay:
LGTM

PR #11178, opened 2018-07-30
Quay fixes

chrisnegus:
Added a description of the Red Hat Quay registry to the Registries page.


vikram-redhat:
Not sure why an extra commit is there - Chris and I couldn't figure it out while trying to get this out the door for 3.10. Will merge and CP for 3.10

PR #11184, opened 2018-07-30
Bring the retry topic back and switch to oc rollout

soltysh:
https://github.com/openshift/origin/pull/20463 removes `oc deploy` entirely. This PR updates the last bit of the docs that was using it, although was commented out. I've updated it to use `oc rollout` instead. 

This is 3.11 item.

@adellape ptal


vikram-redhat:
Merging as this doesn't require QE or peer review.

PR #11187, opened 2018-07-30
[enterprise-3.9] disconnected_install: registry-console is not optional

kalexand-rh:
At this time there is no option to not install registry-console, and the
install checks will look for its image to be there. So, just include it
in the images needed for a disconnected install.

See https://bugzilla.redhat.com/show_bug.cgi?id=1480195#c12

When doing a containerized install, the host components (master, node,
ovs, etcd) need to be synchronized as well.

This led to some confusing usage of the word "containerized" so that
needed some clearing up too.

See https://bugzilla.redhat.com/show_bug.cgi?id=1480195#c12

From https://github.com/openshift/openshift-docs/pull/5248. Because of the differences between this change and current master, this change should apply to only 3.9-3.6, not master. From https://github.com/openshift/openshift-docs/pull/5248.

### DIFF
--- a/install_config/install/disconnected_install.adoc
+++ b/install_config/install/disconnected_install.adoc
@@ -188,7 +188,20 @@ To sync the container images:
 $ systemctl start docker
 ----
 
-. Pull all of the required {product-title} containerized components.
+. If you are performing a xref:rpm_vs_containerized.adoc#install-config-install-rpm-vs-containerized[containerized install],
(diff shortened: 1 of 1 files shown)


kalexand-rh:
@ganhuang approved this change on https://github.com/openshift/openshift-docs/pull/5248, so I'm going to merge it to 3.6-3.9.

PR #11202, opened 2018-07-30
[enterprise-3.10] minor fixes for 3.10

vikram-redhat:


vikram-redhat:
This has already been fixed in master.

PR #11215, opened 2018-07-31
Remove the specific version numbers to facilitate maintenance of different branches

warmchang:
Remove the specific version numbers to facilitate maintenance of different branches.

For example, the description of 3.9 on the 3.10 branch is wrong.

/cherrypick enterprise-3.10
/cherrypick enterprise-3.11


kalexand-rh:
@openshift/team-documentation, I think this change makes sense. PTAL

tmorriso-rh:
LGTM.

PR #10361, opened 2018-06-22
updating sysctls for 3.11

kalexand-rh:
Fixes https://github.com/openshift/openshift-docs/issues/10225.
@php-coder, @ingvagabund, PTAL

PR #11232, opened 2018-07-31
updating list of supported sysctls

kalexand-rh:
Per @ingvagabund on https://github.com/openshift/openshift-docs/pull/10361/files#r206516726, *_net.ipv4.tcp_syncookies_* should be in the safe set going back to version 3.4.

@mdshuai, do you agree with this change?


wjiangjay:
Checked and LGTM

PR #11244, opened 2018-07-31
updating some parameter formatting

kalexand-rh:
Making the format of securityContext consistent across the collection.
@mburke5678, any objections?


mburke5678:
@kalexand-rh LGTM

PR #11257, opened 2018-07-31
Remove useless SELinux section of provisioner doc

wongma7:
We can assume openshift install will ensure pods can read/write nfs volumes, so this is irrelevant for readers of this doc. (The install task is here: https://github.com/openshift/openshift-ansible/blob/9c08d4232fadd7a80b01adb981941f809111db3e/roles/openshift_node/tasks/storage_plugins/nfs.yml)


wongma7:
This patch should be copied to all other release branches where this provisioners.adoc exists

Issue #11267, opened 2018-08-01
3.10 Docs do not correctly perform text substitution for product names


lpsantil:
### Which section(s) is the issue in?
https://docs.openshift.com/container-platform/3.10/scaling_performance/optimizing_storage.html#back-end-recommendations

### What needs fixing?
Perform text substitution on `{gluster-native}/{gluster-external}`

lpsantil:
This may affect other areas as well.  3.9 docs are not affected.

kalexand-rh:
@vikram-redhat

tmorriso-rh:
Work in progress: https://github.com/openshift/openshift-docs/pull/11296

PR #11296, opened 2018-08-01
Fix text substitution for product names

ghost:
This is a follow-up to https://github.com/openshift/openshift-docs/issues/11267.


tmorriso-rh:
@lpsantil PTAL. Thanks.

Cc: @vikram-redhat

PR #11347, opened 2018-08-03
Rebrand openshift.org and OpenShift Origin as okd.io and OKD

wgordon17:


vikram-redhat:
@wgordon17 besides the one comment, the only other thing is to change OpenShift Origin in the _distro_map.yml .

vikram-redhat:
@wgordon17 - Ah nvm my comment about _distro_map.yml. I see you had included it.

Would we need to do anything different with the existing redirects?

wgordon17:
@vikram-redhat Fixed the 3.10. Redirects don't need to be touched, I've already added the new route, so this should be ready to go whenever you are! 👍

vikram-redhat:
Great @wgordon17. I will merge.

Issue #10876, opened 2018-07-18
[enterprise-3.9] Edit suggested in file install_config/routing_from_edge_lb.adoc


rearden-steel:
### Which section(s) is the issue in?

On the ramp node:

Set the following variables:

```
# F5_IP=10.3.89.66
# TUNNEL_IP1=10.3.91.216
# TUNNEL_IP2=10.3.91.217 
# CLUSTER_NETWORK=10.128.0.0/1 

```
### What needs fixing?

`# CLUSTER_NETWORK=10.128.0.0/1 
`
->

`# CLUSTER_NETWORK=10.128.0.0/14
`

vikram-redhat:
@bmeng - is this correct?

Doc: https://docs.openshift.com/container-platform/3.9/install_config/routing_from_edge_lb.html

@bfallonf - PTAL and based on @bmeng's response either fix and close or close this issue.

bmeng:
@lihongan Can you help review this issue? The cluster network is incorrect indeed, but I think the whole paragraph of the ramp node needs to be fixed.

lihongan:
Yes, it should be `# CLUSTER_NETWORK=10.128.0.0/14`. 
Here it takes F5 as example to tell how to configure ramp node with ipip tunnel, but it is just demo and cannot be deployed to production since all settings will gone after ramp node (or openvswitch service) restarting. You can add a "Note" before the paragraph if needed.

bfallonf:
Thanks @rearden-steel @lihongan . I've created #11349 for this issue. I have asked a question about the note box I have added, but we can move the conversation to that PR.

I'll close this issue.

PR #11349, opened 2018-08-03
Issue 10876 Fix ramp node config file setting

bfallonf:
For #10876 

@lihongan Can I please ask about the Note I've added? I think it would be better to explain how the settings can become persistent rather than saying it is not something the configuration shows. Do you have that information?

### DIFF
--- a/install_config/routing_from_edge_lb.adoc
+++ b/install_config/routing_from_edge_lb.adoc
@@ -68,7 +68,6 @@ host and a designated ramp node.
 
 . Set the following variables:
 +
-====
 ----
 # F5_IP=10.3.89.66 <1>
 # RAMP_IP=10.3.89.89 <1>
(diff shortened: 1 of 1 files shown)


lihongan:
@bfallonf unfortunately I don't have a solution to make those settings persistent and I think it is meaningless to find one.
 
As I have said that it just take F5 router as example to describe how to create IPIP tunnul on ramp node, and F5 native vxlan intergration is recommended for F5 BIG-IP version 12.x and above, so I can't see other user cases which need the ramp node. From OCP 3.11 even the F5 router also will be deprecated, so maybe we can delete all part of ramp node from the Doc. @rajatchopra what do you think ?

bfallonf:
OK. Thanks @lihongan . It does look like those efforts are being tracked in trello, so we can leave that conversation for there. I'll move forward with this PR.

kalexand-rh (commented):
Couple of questions and suggestions. overall, LGTM

bfallonf:
Thanks, all. I'll merge.

PR #11368, opened 2018-08-03
[WIP] cd into openshfit-ansible for all ansible-playbook

sdodson:
This ensures that we utilize ansible.cfg file from openshift-ansible
which defines critical configuration items.

This removes a fair amount of conditional documentation code that was
essentially
if origin
 ~/openshift-ansible
else
 /usr/share/ansible/openshift-ansible

I feel that while the admin of origin may be operating from a github
checkout when they read `$ cd /usr/share/ansible/openshift-ansible`
they'll be able to figure out that they should change directories to
their checkout.

This replaces all `# ansible-playbook` with `$ ansible-playbook`. You
should never need to be root to run our playbooks.

Related https://bugzilla.redhat.com/show_bug.cgi?id=1458018

/cc @kalexand-rh @adellape

### DIFF
--- a/admin_guide/diagnostics_tool.adoc
+++ b/admin_guide/diagnostics_tool.adoc
@@ -148,7 +148,7 @@ If there are any errors, this diagnostic stores results and retrieved files in a
 [[admin-guide-diagnostics-tool-server-environment]]
 == Running Diagnostics in a Server Environment
 
-An Ansible-deployed cluster provides additional diagnostic benefits for 
+An Ansible-deployed cluster provides additional diagnostic benefits for
 nodes within an {product-title} cluster. These include:
 
 * Master and node configuration is based on a configuration file in a standard
@@ -234,7 +234,7 @@ circumstances, additional system components, such as `docker` or networking
 configurations, can change if their current state differs from the configuration
 in the inventory file. You should run these health checks only if you do not
 expect the inventory file to make any changes to the existing cluster
-configuration. 
+configuration.
 ====
 
 [[admin-guide-diagnostics-tool-ansible-checks]]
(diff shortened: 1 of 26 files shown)


sdodson:
Meh, there's a fair bit of whitespace diff from my editor. Let me know if you want me to clean that up.

admin_guide/diagnostics_tool.adoc
```diff
@@ -347,27 +347,19 @@ command, specify your cluster's inventory file and run the *_health.yml_*
...
 
 ----
-# ansible-playbook -i <inventory_file> \
-ifdef::openshift-enterprise[]
-    /usr/share/ansible/openshift-ansible/playbooks/openshift-checks/health.yml
-endif::[]
-ifdef::openshift-origin[]
-    ~/openshift-ansible/playbooks/openshift-checks/health.yml
-endif::[]
+# cd /usr/share/ansible/openshift-ansible
```
kalexand-rh: I think we should make the instruction to cd into the folder that contains the Ansible playbooks a separate step or include a text direction to "change to the folder that contains the <playbooks>." It would better describe what we're doing and might help avoid confusion for users who are using the git flow

Do you have to be root to `cd` into this folder?

kalexand-rh (commented):
I'm ok with the whitespace trimming.
I think we should be a bit more descriptive about the directory change.

admin_guide/diagnostics_tool.adoc
sdodson: I'm not sure the best way to approach the need to change directories. It's certainly a mess to have to put it everywhere and explain it everywhere but it's also pretty critical to do so.

No, you don't have to be root.

kalexand-rh: If it's critical, then you have to say it each time, and it might need to be a warning instead of just something you mention offhand.

Then we should change those commands, too.

vikram-redhat:
@jiajliu - PTAL

admin_guide/diagnostics_tool.adoc
```diff
@@ -347,27 +347,19 @@ command, specify your cluster's inventory file and run the *_health.yml_*
 playbook:
```
kalexand-rh: line 345:

```
To run the *openshift-ansible* health checks using the `ansible-playbook`
 command, change to the playbook directory, specify your cluster's inventory file, and run the *_health.yml_*
playbook:
```

admin_guide/topics/proc_adding-hosts.adoc
```diff
@@ -122,14 +122,16 @@ other than the default of *_/etc/ansible/hosts_*, specify the location with the
 ** For additional nodes:
```
kalexand-rh: line 119:

```
Change to the playbook directory, and run the *_scaleup.yml_* playbook.
```

admin_guide/node_problem_detector.adoc
```diff
@@ -135,27 +135,28 @@ The Node Problem Detector consumes resources. If you use the Node Problem Detect
...
 
 [NOTE]
 ====
 Because the Node Problem Detector is in Technology Preview, the `openshift_node_problem_detector_install` is set to `false` by default.
 You must manually change the parameter to `true` when installing the Node Problem Detector.
-====  
+====
 
-If xref:admin-guide-node-problem-detector-verify[the Node Problem Detector is not installed], run the *openshift-node-problem-detector/config.yml* playbook to install Node Problem Detector: 
+If xref:admin-guide-node-problem-detector-verify[the Node Problem Detector is not installed], run the *openshift-node-problem-detector/config.yml* playbook to install Node Problem Detector:
```
kalexand-rh:
```
If xref:admin-guide-node-problem-detector-verify[the Node Problem Detector is not installed], change to the playbook directory and run the *openshift-node-problem-detector/config.yml* playbook to install Node Problem Detector:
```

admin_guide/node_problem_detector.adoc
```diff
@@ -380,7 +381,8 @@ openshift_node_problem_detector_state=absent
 +
```
kalexand-rh: line 380:

```
Change to the playbook directory, and run the *config.yml* Ansible playbook:
```

kalexand-rh (commented):
@sdodson, I think we can keep the commands together and call out the need to change to the directory in the text. If it's fair to call it the "playbook directory," I'd make changes like the following examples.

kalexand-rh:
Follow up's here: #12388

PR #11372, opened 2018-08-03
draft of install from available notes

kalexand-rh:
Rough cut of the 4.0 install stubs.

Preview here: http://file.rdu.redhat.com/kalexand/080618/install_4.0/installation/installing-quickly-cloud.html

### DIFF
--- a/modules/about-custom-installation.adoc
+++ b/modules/about-custom-installation.adoc
@@ -1,0 +1,34 @@
+// Module included in the following assemblies:
+//
+// * installation/installing-customizations-cloud.adoc
+
+[id='about-custom-installation-{context}']
+= About the custom installation
+
+The custom installation is divided into four phases:
+
+* Initialization, where you create and customize the installation configuration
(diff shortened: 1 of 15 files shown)


modules/about-custom-installation.adoc
```diff
@@ -0,0 +1,34 @@
+// Module included in the following assemblies:
+//
+// * installation/installing-customizations-cloud.adoc
+
+[id='about-custom-installation-{context}']
+= About the custom installation
+
+The custom installation is divided into four phases:
```
crawford: We are calling these "installation targets" and they are called "config" (or maybe "install-config"), "manifests", "ignition-configs", and "cluster". Nouns make a little more sense here because that is what is being installed. We are working on those definitions in the [installer repo](https://github.com/openshift/installer/pull/100).

kalexand-rh:
This is still very much WIP, but I'm posting the draft to validate the structure.

PR #11383, opened 2018-08-06
fixes version selector for online, okd, and dedicated

gaurav-nelson:
fixes version selector for online, okd, and dedicated


gaurav-nelson:
@vikram-redhat PTAL

gaurav-nelson:
Removed labels, no need to cherry-pick.

PR #11392, opened 2018-08-06
 Mention that docker gc should only be installed on overlay2

sdodson:
Starting in RHEL 7.5 we support 'overlay2' and 'devicemapper' based storage for docker. By default docker will be provisioned with overlay2 storage unless the admin takes special steps to use devicemapper. The dockergc component is only compatible with hosts configured for overlay2.

The release notes are the only place we mention docker gc or crio in general. We should probably consider having the runtimes team submit a section related to crio configuration to the install guide.

/cc @kalexand-rh @mrunalp 

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1568772

### DIFF
--- a/release_notes/ocp_3_7_release_notes.adoc
+++ b/release_notes/ocp_3_7_release_notes.adoc
@@ -118,18 +118,12 @@ detected.
 When CRI-O use is enabled, it is installed alongside `docker`, which currently
 is required to perform build and push operations to the registry. Over time,
 temporary `docker` builds can accumulate on nodes. You can optionally set the
-following to enable garbage collection, which adds a daemonset to clean out the
-builds:
+following parameter to enable garbage collection. You must configure the node
+selector so that the garbage collector runs only on nodes where `docker` is
+configured for `overlay2` storage.
 
 ----
 openshift_crio_enable_docker_gc=true
-----
-
-When enabled, it will run garbage collection on all nodes by default. You can
-also limit the running of the daemonset on specific nodes by setting the
-following:
-
-----
 openshift_crio_docker_gc_node_selector={'runtime': 'cri-o'}
 ----
 
--- a/release_notes/ocp_3_9_release_notes.adoc
+++ b/release_notes/ocp_3_9_release_notes.adoc
@@ -133,18 +133,12 @@ detected.
 When CRI-O use is enabled, it is installed alongside `docker`, which currently
 is required to perform build and push operations to the registry. Over time,
 temporary `docker` builds can accumulate on nodes. You can optionally set the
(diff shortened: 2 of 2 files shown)


kalexand-rh:
@sdodson, who on the runtimes team should we tag for that doc update?

release_notes/ocp_3_7_release_notes.adoc
```diff
@@ -118,18 +118,12 @@ detected.
 When CRI-O use is enabled, it is installed alongside `docker`, which currently
 is required to perform build and push operations to the registry. Over time,
 temporary `docker` builds can accumulate on nodes. You can optionally set the
-following to enable garbage collection, which adds a daemonset to clean out the
-builds:
+following to enable garbage collection. You should make sure to configure the
```
kalexand-rh: I'd say:

following parameter to enable garbage collection. You must configure the node selector so that the garbage collector runs only on nodes where \`docker\` is configured for \`overlay2\` storage.

(in both files, please)

kalexand-rh (changes requested):
I think we also need a note in the admin_guide/garbage_collection.adoc file, possibly at line 32.

Something like
```
[NOTE]
====
You can configure garbage collection only on nodes that use \`overlay2\` storage.
====
```

sdodson:
The Container Garbage Collection section of admin_guide/garbage_collection.adoc refers to garbage collection performed automatically by the kubelet. AFAIK there are no special considerations to be made there.

The `docker_gc` stuff in this PR is specifically related to the condition where you're using CRI-O to run pods but Docker to build container images. In that situation docker_gc is responsible for garbage collecting docker because the kubelet is unaware that docker exists on the host as well and would have container images piling up through the build process.

I'd like @mrunalp to verify what I'm saying is correct however, thanks for making sure we consider this.

release_notes/ocp_3_7_release_notes.adoc
sdodson: ok, updated them.

kalexand-rh:
That makes sense. Thank you for clarifying that!

kalexand-rh (approved):

kalexand-rh:
@jiajliu, will you PTAL?

bfallonf:
@sdodson @kalexand-rh LGTM :1st_place_medal:

mrunalp (approved):
LGTM

kalexand-rh:
@xltian, PTAL

jianlinliu:
LGTM.

kalexand-rh:
@mburke5678 sent some feedback on IRC. Thanks!

PR #11410, opened 2018-08-07
BZ#1570159 - added replicasets

gaurav-nelson:
For https://bugzilla.redhat.com/show_bug.cgi?id=1570159

### DIFF
--- a/architecture/core_concepts/deployments.adoc
+++ b/architecture/core_concepts/deployments.adoc
@@ -70,6 +70,62 @@ spec:
 <4> Labels on the pod should include those from the label selector.
 <5> The maximum name length after expanding any parameters is 63 characters.
 
+[[replica-set]]
+== Replica set
+Similar to a xref:replication-controllers[replication controller], a replica set
+ensures that a specified number of pod replicas are running at any given time.
+The difference between a replica set and a replication controller is that a
+replica set supports set-based selector requirements whereas a replication
+controller only supports equality-based selector requirements.
+
+[NOTE]
+====
+Only use replica sets if you require custom update orchestration or do not
+require updates at all, otherwise, use
+xref:deployments-and-deployment-configurations[Deployments].
+Replica sets can be used independently, but are used by deployments to
+orchestrate pod creation, deletion, and updates. Deployments manage their
+replica sets automatically, provide declarative updates to pods, and do not have
+to manually manage the replica sets that they create.
+====
+A replica set is a core Kubernetes object called `*ReplicaSet*`.
+
+The following is an example `*ReplicaSet*` definition:
+
+[source, yaml]
+----
+apiVersion: apps/v1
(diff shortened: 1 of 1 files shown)


vikram-redhat:
@gaurav-nelson - is this ready to be moved along?

gaurav-nelson:
@bfallonf PTAL

architecture/core_concepts/deployments.adoc
```diff
@@ -70,6 +70,63 @@ spec:
...
+[NOTE]
+====
+It is recommended to use
+xref:deployments-and-deployment-configurations[Deployments] rather than using
+replica sets. Replica sets can be used independently, but it is used by
+deployments to orchestrate pod creation, deletion and updates. When you use
+deployments you do not have to manually manage the replica sets that they
+create. Deployments own and manage their replica sets automatically and provides
+declarative updates to pods. Only use replica sets if you require custom update
+orchestration or do not require updates at all.
```
bfallonf: I think this might be burying the lede. Maybe something like:

----
Only use replica sets if you require custom update orchestration or do not require updates at all, otherwise, use xref:deployments-and-deployment- configurations[Deployments]. Replica sets can be used independently, but are used by deployments to orchestrate pod creation, deletion, and updates. Deployments manage their replica sets automatically, provide declarative updates to pods, and do not have to manually manage the replica sets that they create.
----

Note the grammar and punctuation changes to the sentences as well.

architecture/core_concepts/deployments.adoc
```diff
@@ -70,6 +70,63 @@ spec:
...
+        - containerPort: 8080
+          protocol: TCP
+      restartPolicy: Always
+----
+<1> A label query over a set of resources. The result of `matchLabels` and
+`matchExpressions` are logically conjoined.
+<2> Equality-based selector to specify resources with labels that match the
+selector.
+<3> Set-based selector to filter keys. This selects all resources with key equal
+to `tier` and value equal to `frontend`.
```
bfallonf: Maybe s/This selects all resources with key equal to `tier` and value equal to `frontend`./This Replicaset selects all resources with the key of `tier` with the value of `frontend`. 

Does the `operator` bit mean anything? Should it be "selects all resources with the values of `tier` and `operator` with the value of `frontend`."?

bfallonf:
@gaurav-nelson I made some comments. Let me know if I'm wrong/vague.

architecture/core_concepts/deployments.adoc
gaurav-nelson: @bfallonf Replica set does not select, the selector selects. Operator could be `equal` or `notin` etc. so we can't say `operator` with value.

gaurav-nelson:
Thanks @bfallonf I will update it based on your suggestions. But I am keeping the code examples as it is.

Issue #9889, opened 2018-06-07
[enterprise-3.9] Edit suggested in file install_config/install/advanced_install.adoc


etoews:
### Which section(s) is the issue in?

Configuring the OpenShift Ansible Broker

### What needs fixing?

The `ansible_service_broker_remove` option should be documented. See [this comment](https://github.com/openshift/openshift-ansible/issues/7305#issuecomment-395045769) from @jocelynthode.

PR #11417, opened 2018-08-07
adding parameters that disable OAB and TSB

kalexand-rh:
Fixes https://github.com/openshift/openshift-docs/issues/9889

Per @sdodson in IRC

@jiajliu, will you PTAL?

### DIFF
--- a/install/configuring_inventory_file.adoc
+++ b/install/configuring_inventory_file.adoc
@@ -1720,31 +1720,41 @@ openshift_enable_unsupported_configurations=True
 ----
 
 [[enabling-service-catalog]]
-== Configuring the Service Catalog
+== Customizing Service Catalog Options
 
 The
-xref:../architecture/service_catalog/index.adoc#architecture-additional-concepts-service-catalog[service catalog] is enabled by default during installation. Enabling the service broker
-allows service brokers to be registered with the catalog.
+xref:../architecture/service_catalog/index.adoc#architecture-additional-concepts-service-catalog[service catalog]
+is enabled by default during installation. Enabling the service broker allows
+you to register service brokers with the catalog. When the service catalog is
+enabled, the OpenShift Ansible broker and template service broker are both
+installed as well; see
+xref:configuring-openshift-ansible-broker[Configuring the OpenShift Ansible Broker]
+and xref:configuring-template-service-broker[Configuring the Template Service Broker]
+for more information.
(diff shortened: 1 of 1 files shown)


sdodson:
lgtm, only thing I can think of is that these three components are linked together, it may be worth considering placing the ASB and TSB as subheadings of the Service Catalog heading here.

kalexand-rh:
@jianlinliu PTAL

jianlinliu:
LGTM, the only suggestion is that replace OAB with ASB.

kalexand-rh:
@jianlinliu, OAB is used far more frequently in the docs than ASB. If we need to update all the references, please file a bug.

kalexand-rh:

install/configuring_inventory_file.adoc
```diff
@@ -1734,17 +1734,30 @@ inventory file:
 ----
 openshift_enable_service_catalog=false
 ----
+
+If you disable the service catalog, the OpenShift Ansible broker and template
+service broker are not installed.
 ====
 
 When the service catalog is enabled, the OpenShift Ansible broker and template service broker are both enabled
 as well; see xref:configuring-openshift-ansible-broker[Configuring the OpenShift Ansible Broker] and xref:configuring-template-service-broker[Configuring the Template Service Broker] for more information.
```
bfallonf: In the least, I think what we're adding can be combined with the sentence above the Ansible variable. So:

> [NOTE]
> ====
> To not install the OpenShift Ansible broker and the template service broker, set the service catalog variable in the Ansible inventory file at installation time to `false`:
> 
> ----
> openshift_enable_service_catalog=false
> ----
> ====

But, this section confuses me. It's called "Configuring the service catalog" and the only point is that it's enabled by default, and then there's a thing on how to disable it, but it's in a note box? If anything, I think "Enabling or Disabling the service catalog at installation" might be a better title, and getting rid of the note box.

Sorry, I realise this was like this when you got here...

bfallonf: I think similar comments can be made about the other two sections you've touched below. Up to you how deep you want to go, though.

bfallonf:
@kalexand-rh I've made some comments. Feel free to implement them or throw them out the window. Otherwise, LGTM :door:

kalexand-rh:
Manual CPs for 3.9 and 3.7 because the change needs to go into a different file.

PR #11418, opened 2018-08-07
Added technical note related to Ansible version

ahardin-rh:
Added per discussion with @brenton

PR #11422, opened 2018-08-07
Added OCP 3.6 technical note related to Ansible version

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1605180

3.7 fix: https://github.com/openshift/openshift-docs/pull/11418

### DIFF
--- a/release_notes/ocp_3_6_release_notes.adoc
+++ b/release_notes/ocp_3_6_release_notes.adoc
@@ -1468,6 +1468,18 @@ secrets or configuration maps.
 {product-title} 3.6 introduces the following notable technical changes.
 
 [discrete]
+[[3-6-use-ansible-version-that-ships-with-OCP]]
+=== Use the Ansible Version Shipping with {product-title}
+
+{product-title} 3.6 and 3.7 were developed and tested using Ansible 2.3, which
+ships in the {product-title} channels. Subsequently, the RHEL 7 Extras channel
+added Ansible 2.4, which has known issues with {product-title} 3.6 and 3.7. If
+you experience any problems with the installer, downgrade to Ansible 2.3 by
+running `yum downgrade ansible-2.3\*` See
+link:https://bugzilla.redhat.com/show_bug.cgi?id=1575063[*BZ#1575063*] for
+additional information.
+
+[discrete]
 [[ocp-pci-dss-compliance]]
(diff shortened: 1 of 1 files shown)


ahardin-rh:
@xltian Can you please provide a QE review? Thanks!

release_notes/ocp_3_6_release_notes.adoc
```diff
@@ -1467,6 +1467,18 @@ secrets or configuration maps.
...
+=== Use the Ansible Version Shipping with {product-title}
+
+Use the version of Ansible that ships in the {product-title} channels, version
+2.3, instead of the version that ships with RHEL. Later versions of
+{product-title} use the version from RHEL. However, on {product-title} 3.6, if
+you use Ansible 2.4, you can experience the memory allocation error described in
+link:https://bugzilla.redhat.com/show_bug.cgi?id=1575063[*BZ#1575063*]. To
+downgrade Ansible, disable the Ansible 2.4 channel and run `yum downgrade
+ansible-2.3*`.
+
```
jianlinliu: How about this:
Recommend to use the version of Ansible that ships in the {product-title} channels, version
2.3, instead of the version that ships in RHEL Extras channel. However, on {product-title} 3.6, if
you use Ansible 2.4, you can experience the memory allocation error described in
link:https://bugzilla.redhat.com/show_bug.cgi?id=1575063[*BZ#1575063*]. To
downgrade Ansible, disable the RHEL Extras channel and run `yum downgrade
ansible-2.3*`.

sdodson: Reading this again I can see how it can be confusing, perhaps something like this:

{product-title} 3.6 and 3.7 were developed and tested using Ansible 2.3 which ships in the {product-title} channels. Subsequently the RHEL 7 Extras channel added Ansible 2.4 which has known issues with {product-title} 3.6 and 3.7. If you experience any problems with the
installer please downgrade to Ansible 2.3 by running `yum downgrade ansible-2.3\*` Please see ink:https://bugzilla.redhat.com/show_bug.cgi?id=1575063[*BZ#1575063*] for additional information.


The problem wasn't introduced by us and is simply a side effect of RHEL shipping a newer package than we ship. Memory consumption is one problem we're aware of but there may be others that we just haven't found yet.

jianlinliu:
LGTM.

PR #11430, opened 2018-08-08
bug 1613765 updating openshift_hostname information

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1613765

### DIFF
--- a/install/prerequisites.adoc
+++ b/install/prerequisites.adoc
@@ -397,9 +397,6 @@ the router will be deployed:
 *.cloudapps.example.com. 300 IN  A 192.168.133.2
 ----
 
-When referencing nodes that are not integrated with a cloud provider, use
-the host names that the output of the `hostname` command on each node.
-
 [WARNING]
 ====
 In your *_/etc/resolv.conf_* file on each node host, ensure that the DNS server
@@ -407,6 +404,44 @@ that has the wildcard entry is not listed as a nameserver or that the wildcard
 domain is not listed in the search list. Otherwise, containers managed by
 {product-title} might fail to resolve host names properly.
 ====
+
+[[node-hostnames-prereq]]
+==== Configuring Node Host Names
+
+When you set up a cluster that is not integrated with a cloud provider, you must
+correctly set your nodes' host names. Each node's host name must be resolvable,
+and each node must be able to reach each other node. 
+
+To confirm that a node can reach another node:
+
+. One one node, obtain the host name:
++
+----
+$ hostname
+
+master-1
+----
+
+. On that same node, obtain the fully qualified domain name of the host:
++
+----
+$ hostname -f
+
+master-1.example.com
+----
+
+. From a different node, confirm that you can reach the first node:
++
+----
+$ ping master-1 -c 1
+
+PING master-1.example.com (172.16.122.9) 56(84) bytes of data.
+64 bytes from master-1.example.com (172.16.122.9): icmp_seq=1 ttl=64 time=0.319 ms
+
+--- master-1.example.com ping statistics ---
+1 packets transmitted, 1 received, 0% packet loss, time 0ms
+rtt min/avg/max/mdev = 0.319/0.319/0.319/0.000 ms
+----
 
 [[prereq-network-access]]
 === Network Access Requirements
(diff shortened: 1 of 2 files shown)


install/prerequisites.adoc
```diff
@@ -390,9 +390,8 @@ the router will be deployed:
 *.cloudapps.example.com. 300 IN  A 192.168.133.2
 ----
 
-In almost all cases, when referencing VMs you must use host names, and the host
-names that you use must match the output of the `hostname -f` command on each
-node.
+When referencing VMs, use the host names that the output of the `hostname -f`
```
juanluisvaladas: Why specifically VMs? I don't think there's a difference with physical hosts. "nodes that aren't integrated with a cloud provider" is probably a more precise term

kalexand-rh: The docs specifically refer to VMs in other places: https://docs.openshift.com/container-platform/3.10/install/prerequisites.html#overriding-detected-ip-addresses-host-names

I need to get more info about this change, but I'll try to get feedback from QE on this point. Thanks!

kalexand-rh:
This should still be valid for master, 3.10, and 3.11. There are additional steps that you must take during 3.9-3.10 upgrade if you use this parameter. Those changes are here: https://github.com/openshift/openshift-docs/pull/12539

@jianlinliu, do you think there are any other changes that apply to 3.10 and 3.11?

install/prerequisites.adoc
jianlinliu: Agree, here should not be specifically VMs. 
But the following description is not right. `hostname` == `hostname -f` is not a must or a prerequisites, especially for fresh install. In QE's past testing, in our cluster,  `hostname` is always not equal to `hostname -f`, everything is going well. Of course, if user could assure `hostname` == `hostname -f` would be reduce ratio of hitting some strange issue as much as possible.

jianlinliu:
QE is still verifying related bug for the newly openshift-ansible change, once we completed bug verification, will have a final review here.

install/prerequisites.adoc
```diff
@@ -397,9 +397,8 @@ the router will be deployed:
 *.cloudapps.example.com. 300 IN  A 192.168.133.2
 ----
 
-In almost all cases, when referencing VMs you must use host names, and the host
-names that you use must match the output of the `hostname -f` command on each
-node.
+When referencing nodes that are not integrated with a cloud provider, use
```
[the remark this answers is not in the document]
kalexand-rh: @jianlinliu, I think that's completely fair. Thanks! I'm adding a new commit to capture your suggestions - will you PTAL?

install/prerequisites.adoc
```diff
@@ -408,6 +405,44 @@ domain is not listed in the search list. Otherwise, containers managed by
...
+and each node must be able to reach each other node. 
+
+To confirm that a node can reach another node:
+
+. One one node, obtain the host name:
++
+----
+$ hostname
+
+master-1
```
juanluisvaladas: Since when? We've had **a lot** of issues regarding certificates not having the fqdn as a subject alternative name and connecting to them using the fqdn, or having only the fqdn in the alt names and services connecting to them using only the shortname.

This has been a mess, particuarly with the etcd_hostname variable. Not only it has been problematic, but also the vast majority of our customers have the ```hostname``` as the fqdn

juanluisvaladas: I understand it can return both the short name or the fqdn. I mean this implies hostname **must** return the shortname, and not only this is not necessary but also it's problematic.

We've had a lot of support cases related to having shortname  in a place and fqdn in anohter. I would say we should document hostname returning the fqdn as a best practice.

jianlinliu: Okay, I am fine with document hostname returning the fqdn as a best practice.

kalexand-rh:
@jianlinliu, @juanluisvaladas, is this what you wanted?

juanluisvaladas:
@kalexand-rh yes, thank you very much. I'm pretty sure this will save a lot of trouble in the long run.

kalexand-rh:
Great! I'll wait for @jianlinliu to confirm and get this live tomorrow. Thank you both. :)

jianlinliu:
LGTM.

kalexand-rh:
Thank you! Merging. :)

PR #11432, opened 2018-08-08
bug 1610348 moving openshift-ansible update command

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1610348

@jiajliu, PTAL


kalexand-rh:
This change was approved in the source bug.

PR #11448, opened 2018-08-09
Fix space and unicode quotes in oc annotate command

cben:
1. unicode quotes “ U+201C, ” U+201D will not work in a shell command
2. `oc annotate` takes `name=value` in a single argument, not `name=` and separate `value`.


vikram-redhat:
lgtm. I will merge.

PR #11449, opened 2018-08-09
Switch to oc set env/volume, since oc env/volume is now removed

soltysh:
In openshift/origin#20139 I removed oc env and oc volume, this PR updates the ansible not to use them anymore.

This is 3.11 item.

@adellape ptal


vikram-redhat:
@bmcelvee - PTAL.

bmcelvee:
LGTM :star: 

All associated PRs are merged.

PR #11451, opened 2018-08-09
admin guide - note: not all net.* syscts are namespaced

stlaz:
Related to: https://github.com/openshift/origin/issues/20470

### DIFF
--- a/admin_guide/sysctls.adoc
+++ b/admin_guide/sysctls.adoc
@@ -52,7 +52,19 @@ The following sysctls are known to be namespaced:
 - *_kernel.msg*_*
 - *_kernel.sem_*
 - *_fs.mqueue.*_*
-- *_net.*_*
+
+Additionally, most of the sysctls in the *_net.*_* group are known
+to be namespaced. Their namespace adoption differs based on the kernel
+version and distributor.
+
+To check which *_net.*_* sysctls are namespaced on your system, run the
+following command:
+
+----
+$ podman run --rm -ti docker.io/fedora \
+    /bin/sh -c "dnf install -y findutils && find /proc/sys/ \
+    | grep -e /proc/sys/net"
+----
 
 Sysctls that are not namespaced are called _node-level_ and must be set
 manually by the cluster administrator, either by means of the underlying Linux


kalexand-rh:
@stlaz, you said that you could run something like 
```
$ docker run --rm -ti fedora /bin/sh -c "dnf install -y findutils && find /proc/sys/" | grep -e "\/proc\/sys\/net\/core"
```
to check. Can we make that command generic and explain what you'd modify to suit your os?

admin_guide/sysctls.adoc
```diff
@@ -54,6 +54,13 @@ The following sysctls are known to be namespaced:
 - *_fs.mqueue.*_*
 - *_net.*_*
 
+[NOTE]
```
kalexand-rh: Instead of a note, I'd say something like,

```
Additionally, some of the sysctls in the *_net.*_* group are known to be namespaced.

To confirm which sysctls are namespaced for your kernel, run the following command:

----
$ command with variables
----

In this command:
* <variable> is ____
* <variable> is ____
```

kalexand-rh (commented):
@stlaz, could something like this work?

stlaz:
I'll add the proposed changes. Generally, it's claimed that `net.*` is namespaced, but both the referenced issue and the upstream discussions prove otherwise.

kalexand-rh:
Thank you!

@stuartchuan, will you PTAL?

tmorriso-rh:
LGTM

stuartchuan:
@wjiangjay would you PTAL?

kalexand-rh:
@wjiangjay will you PTAL?

kalexand-rh:
@wjiangjay will you PTAL?

admin_guide/sysctls.adoc
```diff
@@ -52,7 +52,19 @@ The following sysctls are known to be namespaced:
...
+to be namespaced. Their namespace adoption differs based on the kernel
+version and distributor.
+
+To check which *_net.*_* sysctls are namespaced on your system, run the
+following command:
+
+----
+$ podman run --rm -ti fedora \
+    /bin/sh -c "dnf install -y findutils && find /proc/sys/" \
+    | grep -e "/proc/sys/net"
```
wjiangjay: Will got `Error: Failed to synchronize cache for repo 'updates'` with `--net host`.

And better make grep command in the shell command like:
`podman run --net host --rm -ti fedora     /bin/sh -c "dnf install -y findutils && find /proc/sys/ |  grep -e /proc/sys/net"`

wjiangjay:
Sorry for the late

admin_guide/sysctls.adoc
wjiangjay: Another thing is that we better not make any assumption for images,
means better user qualified for container image name:
`podman run --net host --rm -ti docker.io/fedora /bin/sh -c "dnf install -y findutils && find /proc/sys/ | grep -e /proc/sys/net"`

stlaz: You can't use `--net host` as that would make additional sysctls appear.

I don't really see a reason why `--net host` should change anything about the `Error: Failed to synchronize cache for repo 'updates'` message, except for maybe firewall blocking your containers connecting to internets?

I'll add the grep to the `bach -c` command, that's a good improvement, thanks.

kalexand-rh:
@wjiangjay, no worries. Does this change look good to you now?

wjiangjay:
LGTM now

kalexand-rh:
Thanks! I'm applying this between 3.9 and 3.11.

PR #11467, opened 2018-08-09
Clarify s2i Secrets/ConfigMap Behavior

adambkaplan:
When Secrets and ConfigMaps are added to an s2i build, the `destinationDir` is created if it does not exist.

Bug [1613744](https://bugzilla.redhat.com/show_bug.cgi?id=1613744)

### DIFF
--- a/dev_guide/builds/build_inputs.adoc
+++ b/dev_guide/builds/build_inputs.adoc
@@ -853,9 +853,9 @@ respective `destinationDir`. If you left `destinationDir` empty, then the
 secrets are placed in the working directory of the builder image.
 
 The same rule is used when a `destinationDir` is a relative path; the secrets
-are placed in the paths that are relative to the image's working directory. The
-`destinationDir` must exist or an error will occur. No directory paths are
-created during the copy process.
+are placed in the paths that are relative to the image's working directory.
+The final directory in the `destinationDir` path is created if it does not exist in the builder image.
+All preceding directories in the `destinationDir` must exist, or an error will occur.
 
 [NOTE]
 ====


dev_guide/builds/build_inputs.adoc
```diff
@@ -854,8 +854,8 @@ secrets are placed in the working directory of the builder image.
 
 The same rule is used when a `destinationDir` is a relative path; the secrets
 are placed in the paths that are relative to the image's working directory. The
-`destinationDir` must exist or an error will occur. No directory paths are
-created during the copy process.
+parent directory of `destinationDir` must exist or an error will occur. 
+Only the `destinationDir` is created during the copy process.
```
bparees: "parent directory of destinationDir" doesn't sound quite right.

A few scenarios we need to make sure we understand:

1) /etc/nonexistent1/nonexisting2

2) /etc/nonexistent

3) /etc/existing

4) ./existing/nonexisting

5) ./nonexisting1/nonexisting2

6) ./existing/nonexisting

7) ./existing

I suspect what really needs to be said is "only the final path segment of "destinationDir" will be created, the parent tree must exist".  And then we need to confirm what happens if the entire tree(including the final segment) already exists.

xiuwang: absolute paths:
1. -     /etc/nonexistent1/nonexisting2    -- build fail, only creates nonexistent1 dir.
2. -     /etc/nonexistent                            -- will create /etc/nonexistent dir.
3. -     /etc/existing                                  --  only do copying operation.
relative paths:
1.     ./existing/nonexisting1/nonexistin2    -- build fail, only creates nonexistent1 dir.
2.     ./nonexisting                                      --  ./nonexisting created
3.     ./existing                                            -- only copy
If the entire tree(including the final segment) already exists, just copy cm/secret to destinationdir.

dev_guide/builds/build_inputs.adoc
```diff
@@ -853,9 +853,9 @@ respective `destinationDir`. If you left `destinationDir` empty, then the
 secrets are placed in the working directory of the builder image.
 
 The same rule is used when a `destinationDir` is a relative path; the secrets
-are placed in the paths that are relative to the image's working directory. The
-`destinationDir` must exist or an error will occur. No directory paths are
-created during the copy process.
+are placed in the paths that are relative to the image's working directory.
+The `destinationDir` is created if it does not exist in the builder image.
+However, the parent directory of `destinationDir` must exist or an error will occur.
```
bparees: i still find this phrasing confusing.  "destinationDir" is the name of the field, so "./foo/bar" is the destinationDir.  The parent dir is whatever "." is.

Since "foo/bar" won't be created if they don't exist (only "foo" will be created), it sounds like the behavior (for both relative and absolute paths) is:

"If the final directory in the destinationDir path does not exist, it will be created.  If any other directory in the destinationDir path does not exist, the build will fail"

right?

bparees:
lgtm

bmcelvee:
LGTM

@xiuwang do you need to review before we merge?

xiuwang:
LGTM

PR #11484, opened 2018-08-09
Bug 1566438, added suggestion to validate storage before upgrading

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1566438

@sdodson PTAL. Thanks!

PR #11505, opened 2018-08-10
Adding suggestion to validate storage before upgrading 3.10+

ahardin-rh:
Related work: https://github.com/openshift/openshift-docs/pull/11484


ahardin-rh:
QE verified in the BZ

PR #11509, opened 2018-08-10
Added suggestion to validate storage before upgrading for 3.9

ahardin-rh:


ahardin-rh:
QE verified in the BZ

PR #11511, opened 2018-08-12
updated label for tagging resources

### LINKED ISSUE #11504: [enterprise-3.10] Edit suggested in file install_config/configuring_aws.adoc
* Permissions
* Labeling Clusters for AWS

### What needs fixing?
There is inconsistency between each of these sections on what the tags/labels should be. Under the *Permissions* section, the labels are defined as `kubernetes.io/cluster/<name>,Value=<clusterid>`, whereas under *Labeling Clusters for AWS* they are mentioned as `kubernetes.io/cluster/<clusterid>,Value=(owned|shared)`. I find that the latter method of tagging AWS resources works in testing (this method makes the most sense), whereas the former seems redundant (eg: `<clusterid>` and `<name>` in the same tag). I believe this is an accidental conglomeration of both the old and new methods of tagging resources, being that `Key=KubernetesCluster,Value=clusterid` is mentioned in the *Labeling Clusters for AWS* section as being the old/outdated label. If I'm incorrect, then perhaps some further clarification/differentiation would help in the docs.

gaurav-nelson:
fixes #11504

### DIFF
--- a/install_config/configuring_aws.adoc
+++ b/install_config/configuring_aws.adoc
@@ -58,10 +58,10 @@ Configuring AWS for {product-title} requires the following permissions:
 
 [IMPORTANT]
 ====
-* Every master host, node host, and subnet must have the `kubernetes.io/cluster/<name>,Value=<clusterid>` tag.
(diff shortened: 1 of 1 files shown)


gaurav-nelson:
minor changes no reviews needed.

PR #11541, opened 2018-08-14
registry nfs does not require unsupported

bparees:
this behavior was changed in https://github.com/openshift/openshift-ansible/pull/9271

### DIFF
--- a/install/configuring_inventory_file.adoc
+++ b/install/configuring_inventory_file.adoc
@@ -1041,37 +1041,6 @@ openshift_hosted_registry_storage_volume_size=10Gi
 ----
 
 [discrete]
-[[advanced-install-cluster-registry-storage-upgrade-nfs]]
-==== Upgrading or Installing {product-title} with NFS
-
-The use of NFS for the core {product-title} components is not recommended, as NFS
-(and the NFS Protocol) does not provide the proper consistency needed for the
-applications that make up the {product-title} infrastructure.
-
-As a result, the installer and update playbooks require an option to enable the use
-of NFS with core infrastructure components.
-
-----
-# Enable unsupported configurations, things that will yield a partially
-# functioning cluster but would not be supported for production use
-#openshift_enable_unsupported_configurations=false
-----
-
-If you see the following messages when upgrading or installing your cluster,
-then an additional step is required.
-
-----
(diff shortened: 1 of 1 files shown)


bparees:

bmcelvee:
LGTM

Since this is related to a support issue, @dongboyan77 PTAL

install/configuring_inventory_file.adoc
```diff
@@ -1040,37 +1040,6 @@ openshift_hosted_registry_storage_volume_name=registry
 openshift_hosted_registry_storage_volume_size=10Gi
 ----
 
-[discrete]
-[[advanced-install-cluster-registry-storage-upgrade-nfs]]
-==== Upgrading or Installing {product-title} with NFS
-
-The use of NFS for the core {product-title} components is not recommended, as NFS
```
sferich888: Are we testing NFS with the core {product-title} components where we feel its save to remove this? 
Are we going to adjust the notes for logging, metrics, monitoring, etc? If not we need to not simply remove this but adjust this to explicitly call out where we support/recommend NFS (based on testing).

bparees: this section of doc that is being removed is specific to the registry.  nothing has changed about the docs w/ respect to using NFS w/ logging/metrics/etc.  That content still exists (and still results in an unsupported cluster if you use it, i believe, since the ansible logic wasn't changed for those components either).

vikram-redhat:
@bparees since this section was added from 3.9 onwards only, there is nothing to do for 3.7 docs, right?

bparees:
@vikram-redhat so it would seem.

install/configuring_inventory_file.adoc
rootfs: there is at least one known issue [1] with nfs as registry storage, we should at least mention the workaround (i.e. `no_wdelay`) first [2]

1. https://access.redhat.com/documentation/en-us/openshift_container_platform/3.5/html/installation_and_configuration/setting-up-the-registry
2. https://docs.openshift.com/container-platform/3.6/install_config/registry/registry_known_issues.html#known-issue-nfs-image-push-fails

bmcelvee:
Thanks, @vikram-redhat!

PR #11563, opened 2018-08-15
Bump ansible repo to 2.6

sdodson:
OCP 3.11 will require ansible 2.6 so change the repos we tell them to enable.
This change is only relevant to 3.11+


kalexand-rh:
This change seems amazingly uncontentious. It looks like you got all the references that are relevant to 3.11, and I'll merge when Travis finishes. Thanks @sdodson!

PR #11565, opened 2018-08-15
Remove fast-datapath

sdodson:
This is no longer necessary in OCP 3.10+ as those components now run as pods.


sdodson:
/hold
I missed one second.

kalexand-rh:
@jiajliu, do you agree with this change?

jianlinliu:
@shlao PTAL

kalexand-rh:
Great! Thanks @shlao! Merging.

PR #11573, opened 2018-08-16
bug 1614586 adding tech preview OpenStack playbook

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1614586

@brenton, @xltian PTAL  

If we need to include the command, I need to know how it changes in different versions. (This bug was against 3.9, and I assume it needs to roll forward to 3.11.)

### DIFF
--- a/install/running_install.adoc
+++ b/install/running_install.adoc
@@ -403,6 +403,38 @@ ifdef::openshift-origin[]
 endif::[]
 ----
 
+[running-the-advanced-installation-openstack]
+==== Running the Installation Playbook for OpenStack
+
+[IMPORTANT]
+====
+The OpenStack installation playbook is a Technology Preview feature.
+ifdef::openshift-enterprise[]
+Technology Preview features are not supported with Red Hat production service
+level agreements (SLAs), might not be functionally complete, and Red Hat does
+not recommend to use them for production. These features provide early access to
+upcoming product features, enabling customers to test functionality and provide
+feedback during the development process.
+
+For more information on Red Hat Technology Preview features support scope, see
+https://access.redhat.com/support/offerings/techpreview/.
+endif::[]
+====
+
+To install {product-title} on an existing OpenStack installation, use the
+OpenStack playbook. For more information about the playbook, including detailed
+prerequisites, see
+link:https://github.com/openshift/openshift-ansible/tree/master/playbooks/openstack#installation[the OpenStack Provisioning readme file].
+
+To run the playbook, run the following command:
+
+----
+$ ansible-playbook --user openshift \
+  -i openshift-ansible/playbooks/openstack/inventory.py \
+  -i inventory \
+  openshift-ansible/playbooks/openstack/openshift-cluster/provision_install.yml
+----
+
 [[running-the-advanced-installation-individual-components]]
 === Running Individual Component Playbooks
 
--- a/install_config/configuring_openstack.adoc
+++ b/install_config/configuring_openstack.adoc
@@ -168,6 +168,38 @@ container. Therefore, *_cloud.conf_* should be in *_/etc/origin/_* instead of
 *_/etc/_*.
 ====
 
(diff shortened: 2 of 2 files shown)


install/running_install.adoc
```diff
@@ -403,6 +403,32 @@ ifdef::openshift-origin[]
...
 ----
 
+[running-the-advanced-installation-openstack]
+==== Running the Installation Playbook for OpenStack
+
+[IMPORTANT]
+====
+The OpenStack installation playbook is a Technology Preview feature.
+ifdef::openshift-enterprise[]
+Technology Preview features are not supported with Red Hat production service
```
sferich888: They are supported however they get a lower priority. I would stick to language we use about TP in other places (copy a blurb from one of those). 

It looks like we're simply saying: `X is in Technology Preview and is not for production workloads.` in the release notes now. ` and referring to the link your providing in 419/420

kalexand-rh: @sferich888, I did take that language from another TP note, and it matches 6 other random notes I checked. Will you link me to an example of what you were expecting?

install/running_install.adoc
```diff
@@ -403,6 +403,32 @@ ifdef::openshift-origin[]
...
+https://access.redhat.com/support/offerings/techpreview/.
+endif::[]
+====
+
+If you want to install {product-title} on an existing OpenStack installation,
+you can do so by using the playbook for OpenStack. Run the following command:
+
+----
+# rpm -qf /usr/share/ansible/openshift-ansible/playbooks/openstack
+    openshift-ansible-playbooks-3.9.14-1.git.3.c62bc34.el7.noarch
```
brenton: This command only shows the package that owns the playbook directory.  @tomassedovic, could you provide the exact command we should document?  I'm guessing it may be as simple as just running ansible-playbook and passing it in but I'm curious if there are any other required inventory variables.

xltian: @jianlinliu  Let Jianlin to check if it's something OpenShift QE has been covering or if we need to re-direct to OpenStack QE.

jianlinliu: It is better to re-direct this to OpenStack QE.

kalexand-rh: @vikram-redhat, do we have a contact in OpenStack QE?

kalexand-rh:

install/running_install.adoc
```diff
@@ -403,6 +403,38 @@ ifdef::openshift-origin[]
...
+feedback during the development process.
+
+For more information on Red Hat Technology Preview features support scope, see
+https://access.redhat.com/support/offerings/techpreview/.
+endif::[]
+====
+
+If you want to install {product-title} on an existing OpenStack installation,
+you can do so by using the playbook for OpenStack. For more information about
+the playbook, including detailed prerequisites, see
```
mburke5678: @kalexand-rh Can this be tightened up a bit? 
_To install {product-title} on an existing OpenStack installation, use the OpenStack playbook_

install_config/configuring_openstack.adoc
```diff
@@ -168,6 +168,38 @@ container. Therefore, *_cloud.conf_* should be in *_/etc/origin/_* instead of
...
+upcoming product features, enabling customers to test functionality and provide
+feedback during the development process.
+
+For more information on Red Hat Technology Preview features support scope, see
+https://access.redhat.com/support/offerings/techpreview/.
+endif::[]
+====
+
+If you want to install {product-title} on an existing OpenStack installation,
+you can do so by using the playbook for OpenStack. For more information about
```
mburke5678: @kalexand-rh Same as above

mburke5678:
@kalexand-rh A pair of nits. LGTM pending TP note wording.

kalexand-rh:
Thanks @mburke5678. As far as I can tell, I'm using the correct TP note. I'll be happy to update it it as needed.

PR #11585, opened 2018-08-16
Added location where ansible_service_broker_install is set

ahardin-rh:


ahardin-rh:
QE verified in the BZ

PR #11606, opened 2018-08-17
Remove nodeSelector from the default DaemonSet

jsafrane:
By default, CSI drivers should run on all nodes. Add a note that it's
possible to narrow down the nodes in special cases.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1610709


vikram-redhat:
Thanks @tmorriso-rh

liangxia:
Looks good from QE.

PR #11628, opened 2018-08-17
bug 1565363 fixing unschedulable parameter for upgrade

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1565363

@jianlinliu, PTAL

@sdodson, FYI 

Built version here: http://file.rdu.redhat.com/kalexand/081718/BZ1565363/upgrading/automated_upgrades.html

### DIFF
--- a/upgrading/manual_upgrades.adoc
+++ b/upgrading/manual_upgrades.adoc
@@ -286,12 +286,15 @@ IMAGE_VERSION=<tag>
 +
 Replace `<tag>` with `{latest-tag}` for the latest version.
 
-. In previous versions of {product-title}, master hosts were marked unschedulable
+. Modify host labels and parameters.
++
+In previous versions of {product-title}, master hosts were marked unschedulable
 by default by the installer, meaning that no pods could be placed on the hosts.
 Starting with {product-title} 3.9, however, the web console is removed from the
 master services and deployed as a pod on the masters, requiring that masters be
-marked schedulable. This also requires adding labels to certain hosts, which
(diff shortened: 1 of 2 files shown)


sdodson:
The manual upgrade section references ansible variables, we probably want to still tell them to remove those variables from their inventories but I'd add a step that tells them to label their masters, as far as I can tell the docs already tell them to mark them schedulable.

```
Please label your masters with the 'node-role.kubernetes.io/master=true'
`oc label ose3-master.example.com node-role.kubernetes.io/master=true`
```

upgrading/manual_upgrades.adoc
```diff
@@ -286,12 +286,21 @@ IMAGE_VERSION=<tag>
...
 by default by the installer, meaning that no pods could be placed on the hosts.
 Starting with {product-title} 3.9, however, the web console is removed from the
 master services and deployed as a pod on the masters, requiring that masters be
-marked schedulable. This also requires adding labels to certain hosts, which
-assigns _node roles_ (either `master` or `compute`).
+marked schedulable.
++
+During upgrade, if a value is not provided for the `openshift_schedulable`
+parameter, it is set to *true*. If a value is provided for
+`openshift_schedulable`, it is not changed.
```
sdodson: Since they're not running ansible during the manual upgrade I'd just omit this paragraph, you tell them to remove it from the inventory on 319 which accounts for any future ansible use. The `oc adm manage-node <hostname> --schedulable=true` will be the part that actually marks the node schedulable.
Otherwise lgtm

PR #11630, opened 2018-08-17
bug 1616353 image tag parameter requirements

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1616353

@jianlinliu 
- How far back should this guidance apply?
- Does this sound right to you?

### DIFF
--- a/install/disconnected_install.adoc
+++ b/install/disconnected_install.adoc
@@ -190,7 +190,9 @@ $ systemctl start docker
 
 . Pull all of the required {product-title} infrastructure component images.
 ifdef::openshift-enterprise[]
-Replace `<tag>` with `{latest-tag}` for the latest version.
+For the latest version, you must run this command twice to ensure that you have
+all the required images. Run the command one time as shown, and then replace
+`<tag>` with `{latest-tag}` and run the command again.
 endif::[]
 +
 [source, bash]
@@ -244,7 +246,9 @@ will be removed in future releases.
 . Pull all of the required {product-title} component images for the
 additional centralized log aggregation and metrics aggregation components.
 ifdef::openshift-enterprise[]
-Replace `<tag>` with `{latest-int-tag}` for the latest version.
+For the latest version, you must run this command twice to ensure that you have
+all the required images. Run the command one time as shown, and then replace
+`<tag>` with `{latest-tag}` and run the command again.
 endif::[]
 +
 [source, bash]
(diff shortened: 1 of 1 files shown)


jianlinliu:
> How far back should this guidance apply?
only back to 3.10.
> Does this sound right to you?
Most LGTM. But it is better to sync both set of images just like what I said in the bug.

kalexand-rh:
@jianlinliu, thank you! I've updated the topic based on your comments in the bug. Will you please take another look?

install/disconnected_install.adoc
```diff
@@ -190,7 +190,9 @@ $ systemctl start docker
 
 . Pull all of the required {product-title} infrastructure component images.
 ifdef::openshift-enterprise[]
-Replace `<tag>` with `{latest-tag}` for the latest version.
+For the latest version, you must run this command twice to ensure that you have
+all the required images. Run the command one time as shown, and then replace
+`<tag>` with `{latest-tag}` and run the command again.
```
bfallonf: I  think there's confusion between the two tags. The first one is whatever you need it to be, but then you run it again with the literal {latest-tag}? It's not a variable, yes?

Also, I think "For the latest version" isn't compatible with the second part of the sentence. Maybe something like "If you are not using the latest version of images, you must run this command twice. Run the command one time with the desired tag, and replace <tag> with {latest-tag} and run the command again."

bfallonf:
@kalexand-rh I commented with a not. LG otherwise :cactus:

install/disconnected_install.adoc
jianlinliu: Agree, here the statement about <tag> is easy to be confused. About tag, we have two types, one is major release tag, e.g: v3.10, another is some exact container image tag, such as: v3.10.14. In our product registry, v3.10 always is pointing to the latest container image tag, such as: v3.10.19. If user want the exact v3.10.14 image tag, the v3.10 images pulled in the 2nd time actually is not the same as  v3.10.14 what user want, but v3.10.19. So the safest way is pulling all container images with v3.10.14 tag, when sync them to local mirror registry, should push two tags with one container image. Here is some example command about how to do that:
docker pull registry.access.redhat.com/openshift3/ose-pod:v3.10.14
docker tag registry.access.redhat.com/openshift3/ose-pod:v3.10.14 registry.example.com/openshift3/ose-pod:v3.10.14
docker push registry.example.com/openshift3/ose-pod:v3.10.14
docker tag registry.access.redhat.com/openshift3/ose-pod:v3.10.14 registry.example.com/openshift3/ose-pod:v3.10
docker push registry.access.redhat.com/openshift3/ose-pod:v3.10
This also involves "Preparing Images for Export" section, presently sync image is using `scp` and `dock load` command, not using some central mirror registry, that is not easy to do image tag management, which also is mentioned in https://bugzilla.redhat.com/show_bug.cgi?id=1609697, and according to customer case, most of customer is creating some mirror registry to sync images, probably using docker pull/push command to sync images just like what I did in the above example. So that means the whole disconnected install doc is different what customer did and what QE tested. We need re-structured the whole dis-connected install to keep insistent with what QE and customer did.

kalexand-rh: Ok, I definitely misunderstood what you wanted for syncing the tags. I'll get some more information and update this topic again.

kalexand-rh:
Closing in favor of https://github.com/openshift/openshift-docs/pull/12024

PR #11634, opened 2018-08-18
Fix a minor spelling in build and imagestreams doc

chmouel:


vikram-redhat:
Although the issue exists in versions 3.6 and below as well, I am restricting the changes to N-2 docs only.

PR #11669, opened 2018-08-21
BZ#1616161 - added a note that firewalld is not supported on Atomic Host

gaurav-nelson:
For https://bugzilla.redhat.com/show_bug.cgi?id=1616161

### DIFF
--- a/install/configuring_inventory_file.adoc
+++ b/install/configuring_inventory_file.adoc
@@ -1292,8 +1292,9 @@ ifdef::openshift-enterprise,openshift-origin[]
 
 [IMPORTANT]
 ====
-If you are changing the default firewall, ensure that each host in your cluster
+* If you are changing the default firewall, ensure that each host in your cluster
 is using the same firewall type to prevent inconsistencies.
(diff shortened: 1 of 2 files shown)


install/configuring_inventory_file.adoc
```diff
@@ -1316,11 +1316,24 @@ the Ansible host file at install:
...
 [OSEv3:vars]
-os_firewall_use_firewalld=True
+os_firewall_use_firewalld=True <1>
 ----
+<1> Setting this variable to `true` opens the required ports and adds rules to
+the default zone, ensuring that firewalld is configured correctly.
+
+[IMPORTANT]
+====
+*firewalld* is not supported on Atomic Host. But you can use it by supplying an additional `openshift_enable_unsupported_configurations` parameter.
```
kalexand-rh: s/Host. But/Host, but
s/parameter./parameter in the Ansible host file:  (or other correct file)

kalexand-rh (commented):
Couple of picks, but LGTM.

install/configuring_inventory_file.adoc
sdodson: I don't think we should mention the unsupported options here.

PR #11672, opened 2018-08-21
Changed openshift-ansible-utils package name

sergioperez:
I think it just wants to say "openshift-ansible"


kalexand-rh:
I think you're right. Will you open a new PR against the master branch? This change is valid for the master, enterprise-3.10, and enterprise-3.11 branches, so process says that the change needs to be filed against master first.

sergioperez:
Done at #11673 :)

kalexand-rh:
Thank you so much. :)

PR #11673, opened 2018-08-21
Changed openshift-ansible-utils package name

sergioperez:
I think it just wants to say "openshift-ansible"


kalexand-rh:
I agree.

kalexand-rh:
@vikram-redhat, this is a typo. The 3.9 package included *utils*, but the 3.10 doesn't.The host prep topic refers to "openshift-ansible" as the package name. I'm merging without QE.

PR #11704, opened 2018-08-22
Added openshift_service_catalog_image_prefix and openshift_service_catalog_image_version details

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1599976

I will need to open a separate PR against 3.9, since the files were moved around in later versions.

@miheer PTAL


miheer:
It looks good.

ahardin-rh:
QE verified in the BZ

PR #11753, opened 2018-08-24
Fix typos and standardize name - F5 router plug-in

pdp2shirts:
Fix several typos {product_title} to {product-title}
Remove registered trademark symbol from all but first reference in document
standardize naming convention of F5 router plug-in, F5 BIG-IP Controller for OpenShift
added clarifying release and versioning verbiage around F5 router plug-in and the F5 BIG-IP controller for OpenShift.
Also removed references to OpenShift v2

There has been some field and customer confusion that we hope these standard naming conventions will alleviate.

### DIFF
--- a/architecture/topics/f5_big_ip.adoc
+++ b/architecture/topics/f5_big_ip.adoc
@@ -1,5 +1,5 @@
 ////
-F5 BIG-IP® Router plug-in
+F5 BIG-IP® Router Plug-in
 Module included in the following assemblies:
 
 * architecture/networking/assembly_available_router_plugins.adoc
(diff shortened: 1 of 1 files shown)


vikram-redhat:
@gaurav-nelson - PTAL and merge.

novacain1:
@vikram-redhat will we need to also patch master as well here?  Looking for some guidance as I worked with @pdp2shirts on this and want to make sure these clarifications are carried forward in 3.11, etc.  Thanks!

vikram-redhat:
@novacain1 Hey Dave, yeah these should be opened up against the master and then cherry picked to the right branches (3.10, 3.11). If you don't mind closing this one and opening it against the master and then we can do the cherry picking.

@gaurav-nelson - PTAL after that.

kalexand-rh:
Closing in favor of #11917

PR #11761, opened 2018-08-27
Fixes wrong  Welcome page URLs in breadcrumbs

gaurav-nelson:
In current version, the welcome link always points to 3.10 Welcome page. This fixes the issue by updating the DOM with the current URL for selected version.


vikram-redhat:
@gaurav-nelson - have you tested it locally? If yes, then go ahead and merge, as I can't follow the logic :)

PR #11789, opened 2018-08-28
Fixed a typo in the SDN chapter

mburke5678:
I fixed a typo that I found when assisting the Japanese translators, per @danwinship


mburke5678:
Fixing the typo from 3.5 back via manual edits.

PR #11807, opened 2018-08-29
tempaltes: ocp version assignment fix

jiri-fiala:
* replaces the version assignment with the obviously intended comparison
* fixes the search scope, which is currently limited to OCP 3.10 results

Signed-off-by: Jiri Fiala <jfiala@redhat.com>


jiri-fiala:
@vikram-redhat PTAL

PR #11881, opened 2018-08-31
Installing from a container is ok

kalexand-rh:
https://docs.openshift.com/container-platform/3.10/install/running_install.html#running-the-advanced-installation-containerized is still valid.

@sdodson @vrutkovs, PTAL

### DIFF
--- a/release_notes/ocp_3_10_release_notes.adoc
+++ b/release_notes/ocp_3_10_release_notes.adoc
@@ -1785,10 +1785,10 @@ features marked *GA* indicate _General Availability_.
 |Dropped
 |-
 
-|xref:../install/running_install.adoc#running-the-advanced-installation-system-container[System Container for installer and Kubelet]
+|xref:../install/running_install.adoc#running-the-advanced-installation-system-container[Installing from a system container]
 |TP
 |GA
-|Dropped
+|GA
 
 |Hawkular Agent
 |Dropped


release_notes/ocp_3_10_release_notes.adoc
```diff
@@ -1788,7 +1788,7 @@ features marked *GA* indicate _General Availability_.
 |xref:../install/running_install.adoc#running-the-advanced-installation-system-container[System Container for installer and Kubelet]
 |TP
 |GA
-|Dropped
+|GA
```
sdodson: We may need to split these two. We support running the installer via a system container everywhere.
However, running the Kubelet as a system container only happens on Atomic Host. There's no configurability here, you just install and if the target host is atomic host then you get a system container for the kubelet.

kalexand-rh: @sdodson, would it be more accurate to s/System Container for installer and Kubelet/Installing from a system container at line 1788? If you get the kubelet container for free on Atomic Host, was it ever a tech preview feature?

kalexand-rh: @sdodson, will you PTAL?

sdodson: Yes, I think that change would be fine. 
Yes, system container kubelet on atomic host was techpreview and something you previously had to opt into. I'm fine with foregoing any mention of system container kubelet though since there's no configurability.

kalexand-rh:
Thank you @sdodson! I've updated that line and am merging the change.

PR #11894, opened 2018-09-04
added Service Broker sample module and assembly

gaurav-nelson:
Service Broker


vikram-redhat:
@bergerhoffer - PTAL when you get a chance.

PR #11895, opened 2018-09-04
Create sym links for all modules + added a script for future

vikram-redhat:


vikram-redhat:
@openshift/team-documentation - FYI.

Created a symlink from each bucket/book folder to the modules directory. It has two advantages:

1. Brings us in-line with what the other teams are doing.
2. It allows previews within Atom to load.

PR #11909, opened 2018-09-05
Port of vsphere ref arch

bfallonf:
cc @dav1x 

Questions below. I'll do some further edits after that, then put for review.

### DIFF
--- a/install_config/topics/configuring_for_vsphere_registry.adoc
+++ b/install_config/topics/configuring_for_vsphere_registry.adoc
@@ -1,0 +1,97 @@
+////
+Module included in the following assemblies:
+
+install_config/configuring_vsphere.adoc
+////
+
+[id='vsphere-configuring-registry_{context}']
+= Configuring the {product-title} registry for vSphere
+
+
+[source, bash]
+----
+VMFS
+cd /vmfs/volumes/datastore1/
+# VSAN
+cd /vmfs/volumes/vsanDatastore/
+
+# VMFS
+mkdir kubevols # Not needed but good hygiene
+# VSAN
+/usr/lib/vmware/osfs/bin/osfs-mkdir kubevols # Needed
+
+cd kubevols
+
+vmkfstools -c 25G registry.vmdk
+----
+
+== Configuring {product-title} registry for vSphere using Ansible
+
+.Procedure
+
+To configure the Ansible inventory for the registry to use a vSphere volume:
+
+----
+[OSEv3:vars]
+# vSphere Provider Configuration
+openshift_hosted_registry_storage_kind=vsphere <1>
+openshift_hosted_registry_storage_access_modes=['ReadWriteOnce'] <2>
+openshift_hosted_registry_storage_annotations=['volume.beta.kubernetes.io/storage-provisioner: kubernetes.io/vsphere-volume'] <3>
+openshift_hosted_registry_replicas=1 <4>
+----
+
+<1> The storage type.
+<2> vSphere volumes only support `RWO`.
+<4> The annotation for the volume.
+<4> The number of replicas to configure.
+
+== Manually configuring {product-title} registry for vSphere
+
+To use vSphere volume storage, edit the registry’s configuration file and mount to the registry pod.
+
+.Procedure
+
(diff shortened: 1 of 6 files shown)


install_config/topics/configuring_for_vsphere_registry.adoc
```diff
@@ -0,0 +1,96 @@
...
+
+# VMFS
+mkdir kubevols # Not needed but good hygiene
+# VSAN
+/usr/lib/vmware/osfs/bin/osfs-mkdir kubevols # Needed
+
+cd kubevols
+
+vmkfstools -c 25G registry.vmdk
+----
```
bfallonf: @dav1x Any insight into what the steps above are doing?

install_config/topics/configuring_for_vsphere_registry.adoc
```diff
@@ -0,0 +1,96 @@
...
+.Procedure
+
+To configure the Ansible inventory for the registry to use a vSphere volume:
+
+----
+[OSEv3:vars]
+# vSphere Provider Configuration
+openshift_hosted_registry_storage_kind=vsphere <1>
+openshift_hosted_registry_storage_access_modes=['ReadWriteOnce'] <2>
+openshift_hosted_registry_storage_annotations=['volume.beta.kubernetes.io/storage-provisioner: kubernetes.io/vsphere-volume'] <3>
```
bfallonf: Do the square brackets need to be here? As with the line above? There's another part of the doc where they are needed, so it's worth asking. Alternatively, they might be signifying replaceable values here...

install_config/topics/manually_configuring_for_vsphere.adoc
```diff
@@ -0,0 +1,190 @@
...
+
+[IMPORTANT]
+====
+This ensures that the VMDK always presents a consistent UUID to the VM, allowing the disk
+to be mounted properly.
+
+====
++
+For every virtual machine node that will be participating in the cluster:
+VM properties -> VM Options -> Advanced -> Configuration Parameters -> disk.enabledUUID=TRUE
```
bfallonf: I'm guessing the last in this list isn't something in the console UI? Maybe something you set once you get to the "Configuration parameters" page?

install_config/topics/manually_configuring_for_vsphere.adoc
```diff
@@ -0,0 +1,190 @@
...
+export GOVC_INSECURE=1
+----
++
+. Find the Node VM paths:
++
+[source,bash]
+----
+govc ls /datacenter/vm/<vm-folder-name>
+----
+.. Set disk.EnableUUID to true for all VMs:
```
bfallonf: This says to set it to true, but below it is set to `1`. Does it matter? And we seem to use "TRUE" in other places. Can it be either "TRUE" or "1"?

install_config/topics/configuring_for_vsphere_registry.adoc
dav1x: Sure. If a storage class isnt defined for the vsphere cloud provider, the VMDK has to be manually created. The above step is doing that.

install_config/topics/configuring_for_vsphere_registry.adoc
dav1x: They are required. The variables in question are loading into list values.

install_config/topics/manually_configuring_for_vsphere.adoc
dav1x: It could be. I believe I just pulled the verbage from vmware's upstream docs here:
https://vmware.github.io/vsphere-storage-for-kubernetes/documentation/existing.html

install_config/topics/manually_configuring_for_vsphere.adoc
dav1x: I provided two different ways to set the value: using the vcenter interface and govc. We could probably remove the section completely if need be. The upstream docs don't provide this at all as an option.

vikram-redhat:
Closing this in favor of https://github.com/openshift/openshift-docs/pull/12081

PR #11922, opened 2018-09-07
Document admin console inventory variables

spadgett:
Initial documentation of the admin console inventory variables.

/assign @ahardin-rh

### DIFF
--- a/install/configuring_inventory_file.adoc
+++ b/install/configuring_inventory_file.adoc
@@ -1975,4 +1975,45 @@ xref:../install_config/web_console_customization.adoc#install-config-web-console
 |`openshift_web_console_cluster_resource_overrides_enabled`
 |Boolean value indicating if the cluster is configured for overcommit. When `true`, the web console will hide fields for CPU request, CPU limit, and memory request when editing resource limits since these values should be set by the cluster resource override configuration.
 
+|`openshift_web_console_enable_context_selector`
+|Enable the context selector in the web console and admin console mastheads for quickly switching between the two consoles. Defaults to `true` when both consoles are installed.
+
+|===
+
+== Configuring the Admin Console
+
+The admin console is an additional web interface like the web console, but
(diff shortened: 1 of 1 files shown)


spadgett:
@sdodson fyi

ahardin-rh:
@spadgett Thanks for this!
@yapei Can you please provide a QE review? Thank you!

yapei:
Seems we missed following parameters

`openshift_console_branding
openshift_console_documentation_base_url
openshift_console_base_path
openshift_console_image_name
openshift_console_nodeselector
openshift_console_replica_count`

spadgett:
Yeah, I purposely left most of those out because they shouldn't really needed to be changed. There's a good chance those will be reset when we move to an operator for install.

`openshift_console_base_path` should be there, though. Will update.

spadgett:
Updated

ahardin-rh:
Thank you! LGTM

adellape:
@spadgett Can you confirm the name of what we're calling this console, either "Admin Console" or "Cluster Console"? I'm looking at a latest OKD cluster and see "Cluster Console" in the top-left drop-down switcher, just wondering if that's the same for 3.11.

Also, even though we're talking about "admin" tasks, do we expect all users (with or without cluster-admin or similar permissions) to be able to access this new console?

cc @ahardin-rh

spadgett:
> Can you confirm the name of what we're calling this console, either "Admin Console" or "Cluster Console"? I'm looking at a latest OKD cluster and see "Cluster Console" in the top-left drop-down switcher, just wondering if that's the same for 3.11.

The selector will say "Cluster Console." (I'm not a fan of that name, but that's what it will be.)

> Also, even though we're talking about "admin" tasks, do we expect all users (with or without cluster-admin or similar permissions) to be able to access this new console?

Yes, normal users will be able to access the console, and it is still useful for normal users.

PR #11964, opened 2018-09-10
Added 3.11 cluster limits

ahardin-rh:
@jeremyeder Adding in the 3.11 limits. Any changes for 3.11? Thanks!


jeremyeder:
Not yet.  As mentioned in scrum we can follow up with any edits.  Thanks for adding this column.

PR #10772, opened 2018-07-12
Added section for priority and preemption

ravisantoshgudimetla:
Introduce priority and preemption in docs. It may have styling issues. This is copied from upstream documentation.

/cc @sjenning @aveshagarwal @derekwaynecarr @mburke5678 

@aveshagarwal @mburke5678 - Can you help review this?

Addresses #10761

PR #11969, opened 2018-09-11
DO NOT MERGE Priority preemption followup

mburke5678:


mburke5678:
NOTE: This PR is for dev review. It is a follow-up edit to https://github.com/openshift/openshift-docs/pull/10772.
The material will be merged via https://github.com/openshift/openshift-docs/pull/12070, which is modularized. This PR will be closed unmerged.

mburke5678:
@ravisantoshgudimetla Can you PTAL. I made changes to the upstream docs.
I want to make sure I didn't change the meaning of anything. 
I will make any edits in the other PR, https://github.com/openshift/openshift-docs/pull/12070. I think this PR makes it easier to see the changes.

admin_guide/scheduling/priority_preemption.adoc
```diff
@@ -12,67 +12,66 @@ toc::[]
 
 == Overview
 
-Pods can have priority. Priority indicates the importance of a Pod relative to other Pods. If a Pod cannot be scheduled, the scheduler tries to preempt (evict) lower priority Pods to make scheduling of the pending Pod possible.
+Pod priority indicates the importance of a pod relative to other pods. If a pod cannot be scheduled, the scheduler preempts (evicts) lower-priority pods to make scheduling of the pending pod possible.
 
-Priority also affects scheduling order of Pods and out-of-resource eviction ordering on the Node.Pod priority and preemption .
+Pod priority also affects scheduling order of pods and out-of-resource eviction ordering on the node..
```
ravisantoshgudimetla: why `..`  at the  end?

admin_guide/scheduling/priority_preemption.adoc
```diff
@@ -12,67 +12,66 @@ toc::[]
...
+Pod priority indicates the importance of a pod relative to other pods. If a pod cannot be scheduled, the scheduler preempts (evicts) lower-priority pods to make scheduling of the pending pod possible.
 
-Priority also affects scheduling order of Pods and out-of-resource eviction ordering on the Node.Pod priority and preemption .
+Pod priority also affects scheduling order of pods and out-of-resource eviction ordering on the node.. 
 
-== How to use priority and preemption
+== About Pod Priority
 
-To use priority and preemption, follow these steps:
+When pod priority is enabled, scheduler orders pending pods by their priority and a pending pod is placed ahead of other pending pods with lower priority in the scheduling queue. As a result, the higher priority pod may by scheduled sooner that pods with lower priority if its scheduling requirements are met. If such pod cannot be scheduled, scheduler will continue and tries to schedule other lower priority pods.
```
ravisantoshgudimetla: s/pod priority/this feature because, preemption is a scheduler flag whereas pod priority is an field in pod spec and we need both to be enabled.

admin_guide/scheduling/priority_preemption.adoc
```diff
@@ -88,51 +87,77 @@ spec:
...
     image: nginx
     imagePullPolicy: IfNotPresent
-  priorityClassName: high-priority
+  priorityClassName: high-priority <1>
 ----
 
-=== Effect of Pod priority on scheduling order
+<1> Speciffy the priority class to use with this pod. 
 
-When Pod priority is enabled, scheduler orders pending Pods by their priority and a pending Pod is placed ahead of other pending Pods with lower priority in the scheduling queue. As a result, the higher priority Pod may by scheduled sooner that Pods with lower priority if its scheduling requirements are met. If such Pod cannot be scheduled, scheduler will continue and tries to schedule other lower priority Pods.
```
ravisantoshgudimetla: Why remove this section?

admin_guide/scheduling/priority_preemption.adoc
```diff
@@ -88,51 +87,77 @@ spec:
...
+Also, if the scheduler preempts pods on a node and is waiting for termination, if a pod with a higher-priority pod than the waiting pod needs to be scheduled, the scheduler can schedule the higher-priority pod instead. In such a case, the scheduler clears the `nominatedNodeName` of the waiting pod making the pod eligible for another node.
 
-Please note that Pod P is not necessarily scheduled to the “nominated Node”. After victim Pods are preempted, they get their graceful termination period. If another node becomes available while scheduler is waiting for the victim Pods to terminate, scheduler will use the other node to schedule Pod P. As a result nominatedNodeName and nodeName of Pod spec are not always the same. Also, if scheduler preempts Pods on Node N, but then a higher priority Pod than Pod P arrives, scheduler may give Node N to the new higher priority Pod. In such a case, scheduler clears nominatedNodeName of Pod P. By doing this, scheduler makes Pod P eligible to preempt Pods on another Node.
+Preemption does not necessarily remove all lower-priority pods. The scheduler can schedule a pending pod by removing a portion of the lower-priority pods are removed.
```
ravisantoshgudimetla: typo: `The scheduler can schedule a pending pod by removing a portion of the lower-priority pods are removed. `

admin_guide/scheduling/priority_preemption.adoc
```diff
@@ -88,51 +87,77 @@ spec:
...
 
-Please note that Pod P is not necessarily scheduled to the “nominated Node”. After victim Pods are preempted, they get their graceful termination period. If another node becomes available while scheduler is waiting for the victim Pods to terminate, scheduler will use the other node to schedule Pod P. As a result nominatedNodeName and nodeName of Pod spec are not always the same. Also, if scheduler preempts Pods on Node N, but then a higher priority Pod than Pod P arrives, scheduler may give Node N to the new higher priority Pod. In such a case, scheduler clears nominatedNodeName of Pod P. By doing this, scheduler makes Pod P eligible to preempt Pods on another Node.
+Preemption does not necessarily remove all lower-priority pods. The scheduler can schedule a pending pod by removing a portion of the lower-priority pods are removed. 
 
-=== Limitations of preemption
+The schduler consider a node for pod premption only if the pending pod can be scheduled on the node.
```
ravisantoshgudimetla: typo: schduler

admin_guide/scheduling/priority_preemption.adoc
mburke5678: Bad find and replace. Thanks!

admin_guide/scheduling/priority_preemption.adoc
mburke5678: @ravisantoshgudimetla I moved the paragraph higher in the topic. See your comment above ^^

mburke5678:
@ravisantoshgudimetla Thanks for the comments!!

admin_guide/scheduling/priority_preemption.adoc
```diff
@@ -88,51 +87,77 @@ spec:
...
-. There are no other cases of anti-affinity between Pod P and other Pods in the Zone.
-. In order to schedule Pod P on Node N, Pod Q can be preempted, but scheduler does not perform cross-node preemption. So, Pod P will be deemed unschedulable on Node N.
+. Pod Q is running on another node in the same zone as Node N.
+. Pod P has zone-wide anti-affinity with pod Q, meaning Pod P cannot be scheduled in the same zone as Pod Q.
+. There are no other cases of anti-affinity between Pod P and other pods in the zone.
+. In order to schedule Pod P on Node N, the scheduler must preempt Pod Q to remove the pod anti-affinity violation, allowing the scheduler to schedule Pod P on Node N.
+
+Pod Q can be preempted, but scheduler does not perform cross-node preemption. So, Pod P will be deemed unschedulable on Node N. 
+
+== Disabling Priority and Preemption
```
mburke5678: @ravisantoshgudimetla @wjiangjay 
Do we want to include the instructions for disabling the feature? 
1. Feature gates:
`feature-gates.Priority=false` (what is the name to use?) 
2. Edit the scheduler config to set `disablePreemption=true`

admin_guide/scheduling/priority_preemption.adoc
ravisantoshgudimetla: Ohh I missed it, thanks.

admin_guide/scheduling/priority_preemption.adoc
ravisantoshgudimetla: Yes, @mburke5678, that makes sense but we need to be explicit on the scheduler config file for `disablePreemption`. Users may confuse it with scheduler.json file which has predicates and priorities. It should be componentconfig file or update master-config.yaml schedulerArgs section with `disablePreemption=false`.

ravisantoshgudimetla (commented):
@mburke5678 Mostly fine with the changes you made. If you could fix the nits and fine with suggestions made, I can LGTM. Thanks again for helping on the documentation of this important feature.

mburke5678:
@ravisantoshgudimetla A couple of final questions, if I may:

1. The Kubernetes docs say that setting `disablePreemption=false` disables only preemption, and not priority. In this case, the scheduler still respects pod priority, but won't remove pods from nodes in order to schedule them?
2. In the cross-node preemption section, it says that the "scheduler **does not** perform cross-node preemption. So, Pod P will be deemed unschedulable on Node N." Why would the scheduler not perform cross-node preemption?

ravisantoshgudimetla:
> The Kubernetes docs say that setting disablePreemption=false disables only preemption, and not priority. In this case, the scheduler still respects pod priority, but won't remove pods from nodes in order to schedule them?

That is right, that's why I told we have disable both, without that we would see, priority field in the pod spec, when we do a  describe of pod. You're also right that pods won't be removed(preempted), in order to schedule them.

> In the cross-node preemption section, it says that the "scheduler does not perform cross-node preemption. So, Pod P will be deemed unschedulable on Node N." Why would the scheduler not perform cross-node preemption

The problem is performance, say if we want Pod A to be scheduled on Node N, in order to schedule it say if we have to evict some other on other node(an example given was pod anti-affinity at zone level). Having an algorithm with reasonable performance which identifies those pods and evicts them in really difficult, so it was considered out of scope while implementing the feature.

admin_guide/scheduling/priority_preemption.adoc
```diff
@@ -88,51 +87,111 @@ spec:
...
+. In order to schedule Pod P on Node N, the scheduler must preempt Pod Q to remove the pod anti-affinity violation, allowing the scheduler to schedule Pod P on Node N.
 
-If Pod Q were removed from its Node, the Pod anti-affinity violation would be gone, and Pod P could possibly be scheduled on Node N.
+Pod Q can be preempted, but scheduler does not perform cross-node preemption. So, Pod P will be deemed unschedulable on Node N. 
+
+== Disabling Priority and Preemption
+
+To disable pod priority and preemption for a node:
+
+. Modify the node configuration file:
```
ravisantoshgudimetla: This should be master config.

admin_guide/scheduling/priority_preemption.adoc
```diff
@@ -88,51 +87,111 @@ spec:
...
+----
+endif::[]
+ifdef::openshift-origin[]
+----
+# systemctl restart origin-node
+----
+endif::[]
+
+
+To disable the preemption for the cluster:
```
ravisantoshgudimetla: For schedulerArgs section.

ravisantoshgudimetla:
> Does the user need to restart scheduler after configuring the master-config?

Yeah, in our case, if we restart master-api and controllers, we should be good.

mburke5678:
@ravisantoshgudimetla 
> That is right, that's why I told we have disable both, without that we would see, priority field in the pod spec, when we do a describe of pod. You're also right that pods won't be removed(preempted), in order to schedule them.

so, we are back to `disablePreemption=true` in the scheduler configuration file (not scheduler.json).

mburke5678:
@ravisantoshgudimetla A final, final question. 
Are the `System-node-critical` and `System-cluster-critical` classes added to pods during upgrade and install? Or, does the user need to add them?

ravisantoshgudimetla:
> Are the System-node-critical and System-cluster-critical classes added to pods during upgrade and install? Or, does the user need to add them?

If the pods had critical-pod annotation, they would be automatically converted to system-cluster-critical class to ensure backward compatibility

mburke5678:
Trello approved by QE.  https://trello.com/c/SVxMNlhi/944-document-scheduler-system-critical-and-node-critical-priority-classes-should-be-restricted-to-few-namespaces-consuming-priority#action-5b9f09b1502709599695c58d

mburke5678:
@ravisantoshgudimetla @wjiangjay I am going to merge this feature 5/27 EDT through https://github.com/openshift/openshift-docs/pull/12070 unless anyone has any final comments.

Preview of final formatting: https://pod-priority-preemption-reorg--ocpdocs.netlify.com/openshift-enterprise/latest/admin_guide/scheduling/priority_preemption.html
User: stage-user
Pass: zc9$!9S%&0N9hsBVSN42
**Preview is internal use only**

Issue #11983, opened 2018-09-11
[enterprise-3.10] Edit suggested in upgrading/os_upgrades.adoc


davivcgarcia:
### Which section(s) is the issue in?

https://docs.okd.io/3.10/upgrading/os_upgrades.html
https://docs.openshift.com/container-platform/3.10/upgrading/os_upgrades.html

### What needs fixing?

The procedure should describe in details how to proceed with the operating system level updates, making sure that `atomic-openshift-docker-excluder` and `atomic-openshift-excluder` are active to protect sensitive packages.

PR #11984, opened 2018-09-11
fixes openshift/openshift-docs#11983

davivcgarcia:
This pull-request implements the suggestion of #11983.

### DIFF
--- a/upgrading/os_upgrades.adoc
+++ b/upgrading/os_upgrades.adoc
@@ -31,10 +31,23 @@ Use the following to safely upgrade the OS on a host:
 $ oc adm drain <node_name> --force --delete-local-data --ignore-daemonsets
 ----
 
+. In order to protect sensitive packages that should not be updated manually, make
+sure you have applied the proper exclude rules to the hosts being updated.
++
+----
+# atomic-openshift-docker-excluder exclude
+# atomic-openshift-excluder exclude
+----
+
 . Update the host packages, and reboot the host. A reboot ensures that the host is
 running the newest versions, and means that the `docker` and {product-title}
 processes have been restarted, which will force them to check that all of the
 rules in other services are correct.
++
+----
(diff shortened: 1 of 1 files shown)


kalexand-rh:
@davivcgarcia, does this apply to versions before 3.10?

upgrading/os_upgrades.adoc
```diff
@@ -31,11 +31,19 @@ Use the following to safely upgrade the OS on a host:
 $ oc adm drain <node_name> --force --delete-local-data --ignore-daemonsets
 ----
 
-. Update the host packages, and reboot the host. A reboot ensures that the host is
+. Make sure all the sensitive packages are protected by exclude rules, update
```
kalexand-rh: Please add a separate step to enable the excluder rules. I'd restore line 34 to its original state and keep the update and reboot commands in this step.

In the new step, please give a bit more context about why you'd want to do enable these rules.

kalexand-rh (changes requested):
@davivcgarcia, I think you should add a step instead of adding more complexity to this one.

kalexand-rh (approved):
Thanks!

kalexand-rh:
@davivcgarcia, will you please squash your commits?

@jianlinliu, will you PTAL?

davivcgarcia:
@kalexand-rh sorry this mess, I'm still kinda newbie around here. Please let me know if the PR looks good now!

davivcgarcia:
@kalexand-rh answering your previous question, this update should be applicable for all versions.

kalexand-rh:
No worries, @davivcgarcia. It looks much better now. :) 

@jianlinliu, do you agree that we should apply this change going back to 3.6?

jianlinliu:
Yeah, agree to apply this change going back to 3.6

PR #12020, opened 2018-09-13
Remove `remove-cluster-role-from-group` part

josgonza-rh:
Justification: if you run the `oc adm policy` the object ClusterRoleBinding is deleted. As this object is on the OCP bootstrap, next time the api is restarted the object will back to the current configuration so you have to do the previous steps again.

### DIFF
--- a/admin_guide/managing_projects.adoc
+++ b/admin_guide/managing_projects.adoc
@@ -131,13 +131,6 @@ following command:
 $ oc patch clusterrolebinding.rbac self-provisioners -p '{"subjects": null}'
 ----
 +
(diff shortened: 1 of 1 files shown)


barleyer:
Hi,I don't think this PR is necessary，as the document says `If the self-provisioners clusterrolebinding binds the self-provisioner role to more users, groups, or serviceaccounts than the	
system:authenticated:oauth group`，when we run `oc adm policy`,it won't delete clusterrolebinding object,thanks.

barleyer:
It just covers more scenario.

kalexand-rh:
Thanks @barleyer! I'm closing this PR, then.

PR #12023, opened 2018-09-13
Add details for RHBA-2018-2656

ghost:
To be released September 24, 2018

@ahardin-rh PTAL. Did I select the correct Milestone label? Thanks. Do you set any other labels for these updates?


ahardin-rh:
@tmorriso-rh Looks good! You just need a 3.7 label, since that's where we'll cherry-pick it to. Also, I usually leave a note about what date this is shipping.

PR #12033, opened 2018-09-14
Align with github repos move

liangxia:
@openshift/team-documentation


vikram-redhat:
@liangxia do you have a Trello card or some background on this move? Also, will this apply from 3.11, right?

liangxia:
@vikram-redhat  Should apply to all versions.
https://github.com/openshift/origin/issues/19998

PR #12051, opened 2018-09-17
Add necessary API versions

liangxia:
@openshift/team-documentation  Please take a look.

Apply to enterprise-3.7 and above.


kalexand-rh:
Thanks @liangxia! I see no objections. Merging.

PR #12064, opened 2018-09-17
moving ifdef

kalexand-rh:
I think the 3.10 upgrade topic is missing an ifdef. [The topic](https://docs.okd.io/3.10/upgrading/automated_upgrades.html#verifying-the-upgrade) is missing steps compared to 3.9.

Here's how origin would look with this change: http://file.rdu.redhat.com/kalexand/091418/fix-okd-upgrade/upgrading/automated_upgrades.html

@adellape, will you PTAL?


adellape (approved):
:+1:

PR #12154, opened 2018-09-21
draft of 3.11 install updates

kalexand-rh:


install/running_install.adoc
```diff
@@ -165,41 +137,35 @@ ifdef::openshift-enterprise[]
...
-by the `fact_caching_connection` value in the *_/etc/ansible/ansible.cfg_* file. An example of this file is
-shown in xref:../scaling_performance/install_practices.adoc#scaling-performance-install-optimization[Recommended Installation Practices].
-====
+. 
+** If your installation succeeded,
+xref:advanced-verifying-the-installation[verify the installation].
+** If your installation failed, xref:advanced-retrying-installation[retry the installation].
 
 [[running-the-advanced-installation-containerized]]
-=== Running the Containerized Installer
```
kalexand-rh: Is everything in this section still valid?

install/running_install.adoc
```diff
@@ -435,31 +402,78 @@ $ ansible-playbook --user openshift \
...
 
 The main installation playbook *_{pb-prefix}playbooks/deploy_cluster.yml_* runs a
 set of individual component playbooks in a specific order, and the installer
 reports back at the end what phases you have gone through. If the installation
 fails, you are notified which phase failed along with the errors from
 the Ansible run.
 
-After you resolve the errors, you can continue installation:
+////
+[IMPORTANT]
```
kalexand-rh: Is this note still valid?

install/running_install.adoc
```diff
@@ -435,31 +402,78 @@ $ ansible-playbook --user openshift \
...
+endif::[]
+The host initiating the installation does not need to be intended for inclusion
+in the {product-title} cluster, but it can be. Alternatively, a
+xref:running-the-advanced-installation-system-container[containerized version of the installer]
+is available as a system container, which can be run from a RHEL
+Atomic Host system.
+====
+////
+
+[[advanced-retrying-installation]]
```
kalexand-rh: I made a new section that contains some information that didn't fit well in the existing structure. I think having a separate section about how to retry the installer might help. Please let me know if any information needs to be added.

install/running_install.adoc
```diff
@@ -593,117 +607,19 @@ of your etcd hosts in the following:
...
-    /usr/share/ansible/openshift-ansible/playbooks/adhoc/uninstall.yml
-endif::[]
-ifdef::openshift-origin[]
-# ansible-playbook [-i /path/to/file] \
-    ~/openshift-ansible/playbooks/adhoc/uninstall.yml
-endif::[]
-----
-
-[[uninstalling-nodes-advanced]]
-=== Uninstalling Nodes
```
kalexand-rh: I made a new uninstall topic because I think that's a different use case.

install/running_install.adoc
```diff
@@ -593,117 +607,19 @@ of your etcd hosts in the following:
...
 ----
+<1> Provide the load balancer host name listed in the `[lb]` section of your
+inventory file.
 
 You can verify your installation by consulting the
 https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Load_Balancer_Administration/ch-haproxy-setup-VSA.html[HAProxy
 Configuration documentation].
 
-[[optionally-securing-builds]]
-== Optionally Securing Builds
```
kalexand-rh: I don't think this subsection made sense, so I moved it to a prereq.

_topic_map.yml
```diff
@@ -265,23 +265,25 @@ Name: Installing Clusters
...
   File: index
-- Name: System and Environment Requirements
+- Name: System and environment requirements
   File: prerequisites
-- Name: Preparing Your Hosts
+- Name: Preparing your hosts
   File: host_preparation
 - Name: Configuring Your Inventory File
   File: configuring_inventory_file
 - Name: Example Inventory Files
```
kalexand-rh: Will you PTAL at the rendered version of the Example Inventory Files topic? I might just be confused, but I am not sure that all the scenarios are valid with etcd on static pods.

kalexand-rh (commented):
I have a couple of questions.

kalexand-rh:
@xltian, will you PTAL? I bumped the versions and made some text edits for style and structure edits to try to make the topics make more sense. I also had a couple of sections that I'm not sure are still valid - I added comments to those sections.

kalexand-rh:
[Preview build here](http://file.rdu.redhat.com/kalexand/092018/edit_install/install/index.html)

xltian:
@jianlinliu @jiajliu from QE will review

_topic_map.yml
jianlinliu: @gpei PTAL

install/configuring_inventory_file.adoc
```diff
@@ -934,8 +934,8 @@ openshift_hosted_registry_routecertificates= "{'certfile': '<path>/org-cert.pem'
 [[advanced-install-configuring-registry-console]]
 == Configuring the Registry Console
 
-If you are using a Cockpit registry console image other than the default or
-require a specific version of the console, specify the desired registry within
+If you use a Cockpit registry console image other than the default or
```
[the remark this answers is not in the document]
sdodson: Yes, this point can be removed.

_topic_map.yml
gpei: About the `Example Inventory Files` topic

- etcd related inventory file configuration lgtm. We still support both staic pod etcd and rpm etcd in ocp-3.11. When etcd co-located with master, then it would run as static pod on master node, or else etcd would be installed as stand-alone rpm etcd. One thing we could mention here is etcd stand-alone hosts on atomic is no longer supported. 

- Some outdated parameter in the example inventory file "Multiple Masters Using HAProxy Inventory File"
    - openshift_node_kubelet_args is deprecated, could use node config edits instead 
    - openshift_clock_enabled=true is redundant since all the places calling it were set to True by default
```
# apply updated node defaults
openshift_node_kubelet_args={'pods-per-core': ['10'], 'max-pods': ['250'], 'image-gc-high-threshold': ['90'], 'image-gc-low-threshold': ['80']}

# enable ntp on masters to ensure proper failover
openshift_clock_enabled=true
```

kalexand-rh: @gpei, if users wanted to use static etcd pods on masters, what would they need to specify in the inventory file?

Is the static pod the preferred etcd deployment method in 3.10 and 3.11?

When was `openshift_node_kubelet_args` deprecated? I see several references in the collection.

install/running_install.adoc
kalexand-rh: You're right. The section makes it sound like you do it with installation. Thank you!

_topic_map.yml
vrutkovs: >When was openshift_node_kubelet_args deprecated? I see several references in the collection.

In 3.10, during the move to node group configmaps. Instead the node edits should be used:
```
openshift_node_groups=[{'name': 'node-config-all-in-one', 'labels': ['node-role.kubernetes.io/master=true', 'node-role.kubernetes.io/infra=true', 'node-role.kubernetes.io/compute=true'], 'edits': [{ 'key': 'kubeletArguments.pods-per-core','value': ['20']}]}]
```

install/running_install.adoc
wkshi: @kalexand-rh LGTM

install/running_install.adoc
sdodson: They were added by the console and monitoring teams. They can be added to the list or omitted I don't have a strong opinion one way or the other. If they fail then I think the error message is enough for the admin to continue without changes to the documentation.

If we want to add them I can look into the ordering and where to put them.

sdodson: We add playbooks quite frequently, I don't know what criteria we should use to determine whether or not we should document them.

kalexand-rh: @sdodson If they're part of the core installation process, they either need to be added to the list of playbooks at ~470 or we need to remove the table and the "run the rest of the playbooks by hand" method. I'm not sure what drove the inclusion of the "run the rest of the playbooks by hand method," so I can't say which is better.

I'm on the fence about making a table to list all the playbooks that drift in and out and whether they're GA, tech preview, or dev preview/unsupported. (I think it might be a useful addition, but we'd need to talk about a sane method for maintaining it.)

_topic_map.yml
kalexand-rh: I'm making an issue to track any other openshift_node_kubelet_args deprecation changes that aren't included in this PR: https://github.com/openshift/openshift-docs/issues/12200

kalexand-rh:
@gpei, PTAL at https://github.com/openshift/openshift-docs/pull/12154/commits/083151a77c3e41aac85c2451cd9c8dab75e1edaf

install/running_install.adoc
kalexand-rh: I put it back in. Thanks!

kalexand-rh:
@sdodson, @gpei, @jianlinliu, @xltian 

@vikram-redhat asked me to go ahead and merge these changes. I think I've addressed all your feedback so far. If you have more changes or I missed something, please tag me in new comments on the PR so I can open a new one or make a bug and assign it to me.

PR #12161, opened 2018-09-24
Add apiserver qps and burst values recommendation for large clusters

chaitanyaenr:
The apiserver might get overloaded because of low default values for
large and/or dense clusters. This commit adds a recommendation to
avoid this issue.

### DIFF
--- a/scaling_performance/host_practices.adoc
+++ b/scaling_performance/host_practices.adoc
@@ -39,6 +39,21 @@ kubernetesMasterConfig:
     - "1000"
 ----
 
+For large and/or dense clusters, the API server might get overloaded because of the
+default QPS limits. Provided sufficient CPU and memory resources are available on the 
+API server system(s), this issue can be safely alleviated by doubling or quadrupling the 
+default API qps and burst values in the *_/etc/origin/master/master-config.yaml_* file:
+
+----
+masterClients:
+  externalKubernetesClientConnectionOverrides:
+    burst: 1600
+    qps: 800
+  openshiftLoopbackClientConnectionOverrides:
+    burst: 2400
+    qps: 1200
+----
+
 [[scaling-performance-capacity-host-practices-node]]
 == Recommended Practices for {product-title} Node Hosts
 


chaitanyaenr:
@jeremyeder @ashishkamra @mffiedler @ahardin-rh PTAL.

scaling_performance/host_practices.adoc
```diff
@@ -39,6 +39,21 @@ kubernetesMasterConfig:
     - "1000"
 ----
 
+For large and/or dense clusters, the apiserver might get overloaded because of the low
```
jeremyeder: s/ low//

s/default limits/default QPS limits

scaling_performance/host_practices.adoc
```diff
@@ -39,6 +39,21 @@ kubernetesMasterConfig:
     - "1000"
 ----
 
+For large and/or dense clusters, the apiserver might get overloaded because of the low 
+default limits. Provided sufficient CPU and memory resources are available on the 
+apiserver system(s), this issue can be safely alleviated by doubling or quadrupling the 
+default api-qps and api-burst values in the *_/etc/origin/master/master-config.yaml_* file:
```
jeremyeder: Believe these need to be changed in the node-config-master configmap, right?

jeremyeder:
This needs to be ack'd by the master team, as well.  @openshift/sig-master

scaling_performance/host_practices.adoc
deads2k: > Believe these need to be changed in the node-config-master configmap, right?

It depends on which on you mean. The master loops back to itself, so if you ran into rate limiting with the loopback connection, this may be the right spot.

deads2k:
This seems like an ok thing to change. In 4.0 we can re-decide what the default values should be if you like.

scaling_performance/host_practices.adoc
```diff
@@ -39,6 +39,21 @@ kubernetesMasterConfig:
     - "1000"
 ----
```
ahardin-rh: s/apiserver/API server

scaling_performance/host_practices.adoc
ahardin-rh: API `qps` and `burst` values 

Is that accurate?

ahardin-rh:
@chaitanyaenr Is this for 3.11+?

chaitanyaenr:
@ahardin-rh this is for v3.11.

ahardin-rh:
@mffiedler Can you please give a QE review?

mffiedler (approved):

rjhowe:
This value was remove in 3.10 externalKubernetesClientConnectionOverrides

https://github.com/openshift/origin/commit/704f432db592630455b5ceece908b84614dccccc#diff-29dc1a6e13ab4dfdfcbf8bdbbfe0fb6e

chaitanyaenr:
@rjhowe you are right, just setting the qps and burst rates under openshiftLoopbackClientConnectionOverrides is enough instead of both as externalKubernetesClientConnectionOverrides is a dead option. Will update the docs, thanks.

chaitanyaenr:
Just checked the master-config on 3.11 cluster, it has both externalKubernetesClientConnectionOverrides and openshiftLoopbackClientConnectionOverrides options. Maybe we should remove the externalKubernetesClientConnectionOverrides  option from master-config?

PR #12185, opened 2018-09-25
Improve NetworkPolicy documentation

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1594175

### DIFF
--- a/release_notes/ocp_3_7_release_notes.adoc
+++ b/release_notes/ocp_3_7_release_notes.adoc
@@ -568,29 +568,17 @@ Storage to an etcd Node Using PCI Passthrough with OpenStack] is now available.
 ==== Network Policy
 Network Policy is now fully supported in {product-title} 3.7.
 
-Network Policy is an optional plug-in specification of how selections of pods
-are allowed to communicate with each other and other network endpoints. It
+Network Policy is a specification of how groups of pods are allowed to communicate with each other and other network endpoints. It
 provides fine-grained network namespace isolation using labels and port
-specifications.
-
-After installing the Network Policy plug-in, an annotation that flips the
-namespace from `allow all traffic` to `deny all traffic` must first be set on
-the namespace. At that point, `NetworkPolicies` can be created that define what
-traffic to allow. The annotation is as follows:
-
-----
-$ oc annotate namespace ${ns} 'net.beta.kubernetes.io/network-policy={"ingress":{"isolation":"DefaultDeny"}}'
-----
+specifications. 
 
 [NOTE]
 ====
-The annotation is not needed when using the v1 API.
+Network Policy is available only if you use the `ovs-networkpolicy` network plugin.
 ====
 
-The allow-to-red policy specifies "all red pods in namespace `project-a` allow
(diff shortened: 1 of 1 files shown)


mburke5678:
@danwinship PTAL

danwinship:
lgtm

mburke5678:
@bmeng @openshift/team-documentation 
Small change. Please take a look. 

The real change is removing lines 574-538 and adding the note: _Network Policy is available only..._

Change applies to 3.7 Release Notes only.

release_notes/ocp_3_7_release_notes.adoc
```diff
@@ -568,26 +568,16 @@ Storage to an etcd Node Using PCI Passthrough with OpenStack] is now available.
...
-
-----
-$ oc annotate namespace ${ns} 'net.beta.kubernetes.io/network-policy={"ingress":{"isolation":"DefaultDeny"}}'
-----
+specifications. 
 
 [NOTE]
 ====
-The annotation is not needed when using the v1 API.
+Network Policy is available only if you are using the `ovs-networkpolicy` network plugin.
```
kalexand-rh: s/are using/use

release_notes/ocp_3_7_release_notes.adoc
```diff
@@ -568,26 +568,16 @@ Storage to an etcd Node Using PCI Passthrough with OpenStack] is now available.
...
 ====
 
-The allow-to-red policy specifies "all red pods in namespace `project-a` allow
+The *allow-to-red* policy specifies "all red pods in namespace `project-a` allow
```
kalexand-rh: Is this just an example of how to specify a policy, or is *allow-to-red* a new specific policy that's available? If it's an example, I'd add "For example," and explain what a red pod is and how we know it's for `project-a`.

kalexand-rh (commented):
A pick and a question. If you think line 580 makes sense, it's good!

mburke5678:
@danwinship Can you help with @kalexand-rh 's [question above](https://github.com/openshift/openshift-docs/pull/12185#discussion_r220656052)?

release_notes/ocp_3_7_release_notes.adoc
danwinship: Yeah, it should say "for example". The policy in question is just below that paragraph; it didn't show up in this diff but you can see it in the file. "red pods" here means "pods with the label `type=red`", which makes sense in the context of the sample policy. As for "project-a", there's actually nothing in the example that refers to / specifies that. Probably the example was taken from somewhere else where that was explicit. So it should say "in the same namespace as the NetworkPolicy object" rather than "in namespace project-a".

It might make more sense to say:

    For example, the following policy:

        kind: NetworkPolicy
        ...

    specifies ...

mburke5678: @kalexand-rh Good catch!! I hadn't looked at that paragraph but to add the formatting.

vikram-redhat:
@mburke5678 adding label and milestones.

PR #12186, opened 2018-09-25
Added openshift_hosted_router details

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1620117

### DIFF
--- a/install/configuring_inventory_file.adoc
+++ b/install/configuring_inventory_file.adoc
@@ -974,6 +974,34 @@ openshift_hosted_registry_routecertificates= "{'certfile': '<path>/org-cert.pem'
 
 ----
 
+[[install-configuring-router-sharding]]
+== Configuring Router Sharding
+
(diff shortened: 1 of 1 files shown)


ahardin-rh:
@knobunc @sdodson PTAL

sdodson (changes requested):
This looks like it'd provision two routers into the same namespace running on the same ports. wouldn't we need to do something to insure that they're running in different namespaces and on different hosts if they're using the same ports?

Needs review from @openshift/sig-network-edge

Miciah:
The routers should be fine running in the same namespace, and the example specifies two distinct node selectors (`'selector': 'type=router1'` and `'selector': 'type=router2'`), so the routers should be scheduled on different hosts.

sdodson (approved):
Based upon Miciah's feedback, lgtm

ahardin-rh:
QE approval received in the BZ. Fixed minor formatting issue in the in the parameter example.

PR #12190, opened 2018-09-26
update path for imagestreams/templates

sallyom:
@ahardin-rh hi there, can you direct this PR to the right person (if it's not you) thanks!!


vikram-redhat:
@ahardin-rh - This needs to go only in the master branch. I will close this as it requires a rebase since this was submitted but the change should still be ok for master. Can you please submit a new one and merge to master only.

PR #12223, opened 2018-09-27
 Cluster-reader should view all of the openshift logs on kibana

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1624338

### DIFF
--- a/install_config/aggregate_logging.adoc
+++ b/install_config/aggregate_logging.adoc
@@ -461,9 +461,6 @@ these instances on nodes reserved or optimized for running them.
 For example, the selector could be `{"node-type":"infrastructure"}`. At least
 one active node must have this label before Elasticsearch will deploy.
 
-|`openshift_logging_es_ops_allow_cluster_reader`
-|Set to `true` if cluster-reader role is allowed to read operation logs.
-
 |`openshift_logging_es_ops_host`
 |Equivalent to `openshift_logging_es_host` for Ops cluster
 when `openshift_logging_use_ops` is set to `true`.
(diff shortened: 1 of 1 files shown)


mburke5678:
@jcantrill 
> *modify verbiage to state that 'operations users' are those who can satisfy the SubjectAccessReview mentioned in #c.

I am not clear on what the SubjectAccessReview is. Is this an object that the user creates at some point or is it something we create? We don't document SAR at all (save for references in the API guide).

install_config/aggregate_logging.adoc
```diff
@@ -533,7 +530,8 @@ this mode, their saved queries, visualizations, and dashboards are not shared.
 
 You may also set the value `shared_ops`. In this mode, all operations users
 share a Kibana index which allows each operations user to see the same
-queries, visualizations, and dashboards.
+queries, visualizations, and dashboards. Operations users are those users 
+who can satisfy the `SubjectAccessReview`.
```
jcantrill: ... ` oc auth can-i get pod/logs -n default`

jcantrill: @mburke5678 Maybe we drop the ref to `SubjectAccessReview` and just define the command.  It is something they can run to determine there role.

mburke5678: @jcantrill What does the user do if the command shows:
```
#oc auth can-i get pod/logs -n default
no
```
Does the user need `cluster-admin` and `cluster-reader`? You seem unsure in the BZ. What can we tell users to do?

jcantrill: They need an administrator to grant those permissions which is typically some cluster admin role.  I don't know if we want to define the here or not.  It implies elevated permissions

mburke5678:
@jcantrill PTAL

jcantrill (approved):

mburke5678:
@xltian PTAL

xltian:
@QiaolingTang will review from QE.

QiaolingTang:
LGTM

install_config/aggregate_logging.adoc
```diff
@@ -528,12 +525,19 @@ For example, the selector could be `node-type=infrastructure`. At least
...
 share a Kibana index which allows each operations user to see the same
-queries, visualizations, and dashboards.
+queries, visualizations, and dashboards. To determine if you are an operations user:
+
+----
+#oc auth can-i view pod/logs -n default
+yes
+----
+
+If you do not have proper access, contact your cluster administrator.
```
kalexand-rh: s/If you do not have proper access/If you cannot view the pod logs,
IBM Style says:
```
proper: Do not use to mean “correct,” “applicable,” or “appropriate.”
```

kalexand-rh (commented):
One style pick, otherwise LGTM!

PR #12224, opened 2018-09-27
Removed openshift_hostname

bbeaudoin:
Fixes https://github.com/openshift/openshift-ansible/issues/9665

Removed in OpenShift 3.10 as documented in the release notes. Cited as a documentation bug and no longer works as previously documented.


kalexand-rh:
@jianlinliu, any objections?

Issue #12277, opened 2018-10-01
Multiple instances of "* does not match a valid repository ID" when installing OpenShift


brian-avery:
### Which section(s) is the issue in?
Installing Clusters -> Preparing Your Hosts -> Host registration

### What needs fixing?
When installing OpenShift according to the documentation, I am told to attach the OpenShift subscription and then to do:

`subscription-manager repos \
>     --enable="rhel-7-server-rpms" \
>     --enable="rhel-7-server-extras-rpms" \
>     --enable="rhel-7-server-ose-3.10-rpms" \
>     --enable="rhel-7-server-ansible-2.4-rpms"`

However, this results in the following error: 

`Error: 'rhel-7-server-rpms' does not match a valid repository ID. Use "subscription-manager repos --list" to see valid repositories.
Error: 'rhel-7-server-extras-rpms' does not match a valid repository ID. Use "subscription-manager repos --list" to see valid repositories.
Repository 'rhel-7-server-ose-3.10-rpms' is enabled for this system.
Repository 'rhel-7-server-ansible-2.4-rpms' is enabled for this system.`
The current list of repositories to attach is insufficient. Red Hat Enterprise Linux Server is also needed.

PR #12280, opened 2018-10-01
Added missing RHEL subscription reference

brian-avery:
Currently, the step subscription-manager repos \
    --enable="rhel-7-server-rpms" \
    --enable="rhel-7-server-ose-3.11-rpms" \
    --enable="rhel-7-server-ansible-2.6-rpms" fails if the user follows the install guide due to the fact that the Red Hat Enterprise Linux subscription does not get attached. This fixes issue #12277.

### DIFF
--- a/install/host_preparation.adoc
+++ b/install/host_preparation.adoc
@@ -143,6 +143,13 @@ Subscription Manager (RHSM) and attach an active {product-title} subscription.
 # subscription-manager attach --pool=<pool_id>
 ----
 
+List the attached subscriptions:
+----
+# subscription-manager list
+----
+
+OpenShift container platform should now be listed with a status of 'Subscribed'. Red Hat Enterprise Linux Server should also be listed with a status of 'Subscribed'. If it is not, then repeat the two steps above searching for '\*Red Hat Enterprise Linux*' instead of '\*OpenShift*'.
+
 . Disable all yum repositories:
 .. Disable all the enabled RHSM repositories:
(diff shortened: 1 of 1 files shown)


kalexand-rh:
@brian-avery, when did you encounter this issue? The [RHEL install guide](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html-single/installation_guide/#chap-simple-install) has you attach the RHEL subscription, so this step should be addressed during the prereqs.

brian-avery:
@kalexand-rh The Openshift prerequisites have you register your RHEL subscription with your server:
subscription-manager register --username=<user_name> --password=<password>

If the user hasn't already done this then they don't have the RHEL subscription attached and if they have and run this then it will detach it.  Is this an improper assumption?

brian-avery:
@kalexand-rh Looking at the document you attached, I see:  
`subscription-manager register --auto-attach --username=user_name --password=password`

I think if this was added then the repositories would be found correctly without having to have either of the attach commands. Perhaps it would be better to change the register command in the docs to this instead?

brian-avery:
@kalexand-rh any preference between the two? I think using auto-attach would make things cleaner.

brian-avery:
Hi @kalexand-rh Any updates? I run into the same issue on 3.11.

Thanks,
Brian

brian-avery:
It looks like this is fixed in 3.11 now.

PR #12303, opened 2018-10-02
bug 1633647 correcting node names in inventory files

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1633647


kalexand-rh:
@vikram-redhat agrees that this change doesn't need QE, so I'm going to merge.

PR #12309, opened 2018-10-03
 Manually Redeploy Certificates: Fix the overwrite secret command

josgonza-rh:
The command to create the secret must have the `--dry-run` parameter as the secret already exists (you are overwriting it).
If the *dry* parameter is missed an error like the following will be thrown:
```
Error from server (AlreadyExists): secrets "router-certs" already exists
```


kalexand-rh:
@zhaozhanqi, will you PTAL?

zhaozhanqi:
LGTM @kalexand-rh

josgonza-rh:
@kalexand-rh done!

kalexand-rh:
Thank you! Merging. :tada:

PR #12322, opened 2018-10-04
etcd restore for static pods

kalexand-rh:
Draft of the changes for restoring etcd static pods. This change needs to be valid for both 3.10 and 3.11.

### DIFF
--- a/admin_guide/assembly_replace-master-host.adoc
+++ b/admin_guide/assembly_replace-master-host.adoc
@@ -17,7 +17,7 @@ You can replace a failed master host.
 
 First, remove the failed master host from your cluster, and then add a replacement
 master host. If the failed master host ran etcd, scale up etcd by adding etcd
-to the new master host.
+to the new master host. 
 
 [IMPORTANT]
 ====
--- a/admin_guide/assembly_restore-etcd-quorum.adoc
+++ b/admin_guide/assembly_restore-etcd-quorum.adoc
@@ -13,9 +13,13 @@
 toc::[]
 
 
-If you lose etcd quorum, you must back up etcd, take down your etcd cluster,
-and form a new one. You can use one healthy etcd node to form a new cluster, 
-but you must remove all other healthy nodes.
+If you lose etcd quorum, you can restore it.
+
+* If you run etcd on a separate host, you must back up etcd, take down
+your etcd cluster, and form a new one. You can use one healthy etcd node to form
+a new cluster, but you must remove all other healthy nodes.
+* If you run etcd as static pods on your master nodes, you stop the etcd pods,
+create a temporary cluster, and then restart the etcd pods.
 
 [NOTE]
 ====
@@ -46,10 +50,14 @@ cluster is unhealthy
 Note the member IDs and host names of the hosts. You use one of the nodes that
 can be reached to form a new cluster.
 
-include::day_two_guide/topics/proc_backing-up-etcd.adoc[leveloffset=+1]
+== Restoring etcd quorum for separate services
 
-include::day_two_guide/topics/proc_removing-etcd-host.adoc[leveloffset=+1]
+include::day_two_guide/topics/proc_backing-up-etcd.adoc[leveloffset=+2]
 
-include::admin_guide/topics/proc_etcd-quorum-single-node-v2-v3.adoc[leveloffset=+1]
+include::day_two_guide/topics/proc_removing-etcd-host.adoc[leveloffset=+2]
 
-include::day_two_guide/topics/proc_adding-etcd-after-restoring.adoc[leveloffset=+]
+include::admin_guide/topics/proc_etcd-quorum-single-node-v2-v3.adoc[leveloffset=+2]
+
+include::day_two_guide/topics/proc_adding-etcd-after-restoring.adoc[leveloffset=+2]
+
+include::admin_guide/topics/proc_etcd-quorum-static-pod.adoc[leveloffset=+1]
--- a/admin_guide/assembly_restoring-cluster.adoc
+++ b/admin_guide/assembly_restoring-cluster.adoc
@@ -30,6 +30,7 @@ include::admin_guide/topics/proc_restoring-cluster.adoc[leveloffset=+1]
 
 include::day_two_guide/topics/proc_restoring-master.adoc[leveloffset=+1]
 
+
 include::day_two_guide/topics/proc_restoring-node.adoc[leveloffset=+1]
 
 include::day_two_guide/topics/proc_restoring-etcd.adoc[leveloffset=+1]
--- a/day_two_guide/topics/proc_backing-up-etcd.adoc
+++ b/day_two_guide/topics/proc_backing-up-etcd.adoc
@@ -50,15 +50,16 @@ The {product-title} installer creates aliases to avoid typing all the
 flags named `etcdctl2` for etcd v2 tasks and `etcdctl3` for etcd v3 tasks.
 
 However, the `etcdctl3` alias does not provide the full endpoint list to the
-`etcdctl` command, so the `--endpoints` option with all the endpoints must be
-provided.
+`etcdctl` command, so you must specify the `--endpoints` option and list all
+the endpoints.
 ====
 
 Before backing up etcd:
 
-* `etcdctl` binaries should be available or, in containerized installations, the `rhel7/etcd` container should be available
-* Ensure connectivity with the etcd cluster (port 2379/tcp)
-* Ensure the proper certificates to connect to the etcd cluster
+* `etcdctl` binaries must be available or, in containerized installations, the `rhel7/etcd` container must be available.
+* Ensure that the {product-title} API service is running.
+* Ensure connectivity with the etcd cluster (port 2379/tcp).
+* Ensure the proper certificates to connect to the etcd cluster.
 
 ifeval::["{context}" == "environment-backup"]
 . To ensure the etcd cluster is working, check its health.
@@ -186,7 +187,7 @@ joining an existing cluster.
 
 Back up the etcd data:
 
-* If you use the v2 API, take the following actions:
+* If you run etcd on standalone hosts and use the v2 API, take the following actions:
 .. Stop all etcd services:
 +
 ----
@@ -206,7 +207,7 @@ Back up the etcd data:
 ----
 # systemctl start etcd.service
 ----
-* If you use the v3 API, run the following commands:
+* If you run etcd on standalone hosts and use the v3 API, run the following commands:
 +
 [IMPORTANT]
 ====
@@ -247,3 +248,35 @@ any external storage location.
 In the case of an all-in-one cluster, the etcd data directory is located in
 the `/var/lib/origin/openshift.local.etcd` directory.
 --
+* If etcd runs as a static pod, run the following commands:
++
+[IMPORTANT]
+====
+If you use static pods, use the v3 API.
+====
+.. Obtain the etcd endpoint IP address from the static pod manifest:
++
+----
+$ export ETCD_POD_MANIFEST="/etc/origin/node/pods/etcd.yaml"
+$ export ETCD_EP=$(grep https ${ETCD_POD_MANIFEST} | cut -d '/' -f3)
+----
+
+.. Obtain the etcd pod name:
++
+----
+$ oc login -u system:admin
+$ export ETCD_POD=$(oc get pods -n kube-system | grep -o -m 1 '\S*etcd\S*')
+----
+
+.. Take a snapshot of the etcd data in the pod and store it locally:
++
+----
+$ oc project kube-system
+$ oc exec ${ETCD_POD} -c etcd -- /bin/bash -c "ETCDCTL_API=3 etcdctl \
+    --cert /etc/etcd/peer.crt \
+    --key /etc/etcd/peer.key \
+    --cacert /etc/etcd/ca.crt \
+    --endpoints <ETCD_EP> \ <1>
+    snapshot save /etc/etcd/backup/etcd/snapshot.db"
+----
+<1> Specify the etcd endpoint IP address that you obtained.
--- a/day_two_guide/topics/proc_restoring-etcd.adoc
+++ b/day_two_guide/topics/proc_restoring-etcd.adoc
@@ -12,7 +12,7 @@ Module included in the following assemblies:
 = Restoring etcd
 
 The restore procedure for etcd configuration files replaces the appropriate
-files, then restarts the service.
+files, then restarts the service or static pod.
 
 If an etcd host has become corrupted and the `/etc/etcd/etcd.conf` file is lost,
 restore it using:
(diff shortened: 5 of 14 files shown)


admin_guide/assembly_restore-etcd-quorum.adoc
```diff
@@ -13,9 +13,13 @@
...
 
-If you lose etcd quorum, you must back up etcd, take down your etcd cluster,
-and form a new one. You can use one healthy etcd node to form a new cluster, 
-but you must remove all other healthy nodes.
+If you lose etcd quorum, you can restore it.
+
+* If you run etcd as a separate service, you must back up etcd, take down
+your etcd cluster, and form a new one. You can use one healthy etcd node to form
+a new cluster, but you must remove all other healthy nodes.
+* If you run etcd as static pods on your master nodes, you stop the etcd pods,
```
damemi: I believe etcd only runs as a static pod as of 3.10, and am not aware of the user having the option to run as a service

damemi: cc @mfojtik does the user have this option?

mfojtik: yes, there is no option for that

kalexand-rh: I thought @sdodson said that if you already had an etcd cluster and upgraded to 3.10, you kept your current etcd architecture instead of getting converted to static pods. You also have the option of putting etcd on separate nodes and running it as a separate service, there.

Do you agree with those two scenarios? If so, how would you briefly describe them?

sdodson: That's correct, if upgrading from 3.9 with separate etcd hosts those will remain in place. Perhaps s/separate service/separate hosts/ ?

day_two_guide/topics/proc_backing-up-etcd.adoc
```diff
@@ -247,3 +247,35 @@ any external storage location.
...
+----
+$ oc login -u system:admin
+$ export ETCD_POD=$(oc get pods -n kube-system | grep -o -m 1 '\S*etcd\S*')
+----
+
+.. Take a snapshot of the etcd data in the pod and store it locally:
++
+----
+$ oc project kube-system
+$ oc exec ${ETCD_POD} -c etcd -- /bin/bash -c "ETCDCTL_API=3 etcdctl \
```
sdodson: By using oc, this means that this procedure only works when the API is up.

kalexand-rh: @sdodson, is that an unfair assumption? @damemi, are there alternate commands that we can use, or do we need to make "the API server must be running" a prereq for backing up etcd? (If it needs to be a prereq for backing up, I'll need one of you to help me review the situation where we tell users to make an etcd backup and make sure that they're still valid.)

admin_guide/assembly_restore-etcd-quorum.adoc
kalexand-rh: Got it. I think that your suggestion would be perfect if you could run it only on separate nodes, so I added more details at  line 16.

kalexand-rh: Scott clarified more in scrum - if you were running master as a separate service on a master node it **is** converted to a static pod. I'm going to make more updates.

day_two_guide/topics/proc_backing-up-etcd.adoc
sdodson: I guess during normal operations we can assume API availability. But in disaster recovery scenarios it may not be for one reason or another. I'm interested in others feedback here.

damemi: there is the option of doing a `docker ps | grep etcd`, finding the right pod, and doing `docker exec <pod>` instead of `oc exec`

admin_guide/assembly_replace-master-host.adoc
```diff
@@ -17,7 +17,7 @@ You can replace a failed master host.
 
 First, remove the failed master host from your cluster, and then add a replacement
 master host. If the failed master host ran etcd, scale up etcd by adding etcd
-to the new master host.
+to the new master host.
```
kalexand-rh: @sdodson, if you replace a failed master host, is the etcd pod restored automatically, or do you need to scale etcd up on the master separately?

admin_guide/assembly_restoring-cluster.adoc
```diff
@@ -30,6 +30,7 @@ include::admin_guide/topics/proc_restoring-cluster.adoc[leveloffset=+1]
 
 include::day_two_guide/topics/proc_restoring-master.adoc[leveloffset=+1]
 
+
```
kalexand-rh: If the static pod's restored along with the masters, this topic might need rework.

install_config/adding_hosts_to_existing_cluster.adoc
```diff
@@ -90,9 +90,10 @@ instructions.
 . Provision as many new machines as there are masters to replace.
 
 . Add or expand the cluster. for example, if you want to add 3 masters with etcd
-colocated, scale up 3 master nodes or 3 etcd nodes.
+colocated, scale up 3 master nodes.
```
kalexand-rh: Same here - if etcd is added automatically, this process needs to change.

kalexand-rh:
@sdodson, @damemi, I made a few more changes. Will you PTAL and let me know if I've missed anything?

day_two_guide/topics/proc_backing-up-etcd.adoc
geliu2016: @damemi, I think if the ocp env is enabled CRIO, but not docker, then 'docker ps' command is not work also, is it a problem? Currently, I'm prefer to add prereq if API availability is mandatory.

jiajliu:
@geliu2016 @gpei @jianlinliu 
Please help review your part in this doc.

jiajliu:
Upgrade and downgrade part LGTM.

gpei (commented):
In addition, I happen to found our doc is still using `yum update atomic-openshift-utils ` to guide user updating to the latest playbooks, such as  https://github.com/openshift/openshift-docs/blob/master/install_config/adding_hosts_to_existing_cluster.adoc#adding-etcd-hosts-to-existing-cluster , `atomic-openshift-utils` package was on longer provided since 3.10

day_two_guide/topics/proc_backing-up-etcd.adoc
```diff
@@ -247,3 +247,35 @@ any external storage location.
...
+    --cert /etc/etcd/peer.crt \
+    --key /etc/etcd/peer.key \
+    --cacert /etc/etcd/ca.crt \
+    --endpoints <ETCD_EP> \ <1>
+    snapshot save /backup/etcd/snapshot.db"
```
geliu2016: As my understanding，the dir saving snapshot.db could not be specialed feel free, 
because the etcdctl backup will save snapshot.db in etcd pod's inner dir indeedly, if etcd pod crashed, snapshot.db file will lost also, then we could not run etcd restore without snapshot.db, so it have to keep the snapshot.db saved in mount dir which master server's local dir.

dir1 and dir2 is available dir for saving etcd snapshot.db:
vi /etc/origin/node/pods/etcd.yaml:
.....
volumeMounts:
    - mountPath: /etc/etcd/      ====> dir1
      name: master-config
      readOnly: true
    - mountPath: /var/lib/etcd/    ======>dir2
      name: master-data
    workingDir: /var/lib/etcd

@sdodson, pls correct me if there is any misunderstand, thx

day_two_guide/topics/proc_restoring-etcd.adoc
```diff
@@ -252,3 +259,48 @@ ifeval::["{context}" == "downgrade"]
...
++
+----
+$ mv /etc/origin/node/pods/etcd.yaml .
+----
+
+. Clear all old data:
++
+----
+$ rm -rF /var/lib/etcd
+----
```
geliu2016: 'rm -rF' looks like doesn't work, does 'rm -rf' works as well? thx
# rm -rF /var/lib/etcd
rm: invalid option -- 'F'
Try 'rm --help' for more information.

day_two_guide/topics/proc_backing-up-etcd.adoc
sdodson: The master-exec scripts should handle all of that and currently deploy crio and docker specific implementations.

admin_guide/assembly_replace-master-host.adoc
sdodson: etcd would needs to be scaled up as a separate step.

day_two_guide/topics/proc_backing-up-etcd.adoc
sdodson: Right, if the command is running inside the container then you'll want to be careful not to write your backup to a path that's mounted into the container. Such as /var/lib/etcd

install_config/adding_hosts_to_existing_cluster.adoc
kalexand-rh: @gpei, is there a separate process that we should consider for the initial release?

day_two_guide/topics/proc_backing-up-etcd.adoc
sferich888: I think the master-exec stript replacement for this would be: 

```
/usr/local/bin/master-exec etcd etcd ETCDCTL_API=3 etcdctl \ 
```

Instead of changing what were QA'ing we might simply add this as a note; or call out under where we show the snapshot command being run?

day_two_guide/topics/proc_restoring-etcd.adoc
```diff
@@ -252,3 +259,48 @@ ifeval::["{context}" == "downgrade"]
...
++
+You use the etcdctl to recreate the data in the node where you restore the pod.
+
+. Restore the etcd snapshot to the mount path for the etcd pod:
++
+----
+$ export ETCDCTL_API=3
+$ etcdctl snapshot restore /backup/etcd/snapshot.db
```
sferich888: This is called out in other documents (https://github.com/openshift/openshift-docs/pull/12322/files#diff-822c7f3ac6e45a71e6267af9592b9b5cR59) but you might want to link or re-call this out.

sferich888 (changes requested):
Consider the following callouts, and linkage points.

day_two_guide/topics/proc_backing-up-etcd.adoc
kalexand-rh: I think that if @geliu2016 is willing, it would be better to document the one master-exec script method.

install_config/adding_hosts_to_existing_cluster.adoc
kalexand-rh: I added a note at line 95.

sdodson (commented):
all recent changes LGTM

install_config/adding_hosts_to_existing_cluster.adoc
```diff
@@ -92,6 +92,14 @@ instructions.
 . Add or expand the cluster. for example, if you want to add 3 masters with etcd
 colocated, scale up 3 master nodes.
 
+[IMPORTANT]
+====
+In the initial release of {product-title} version 3.11, the *_scaleup.yml_*
+playbook might not scale up etcd if you use CRI-O. This will be fixed in a
```
gpei: The summary of that bug maybe a little confusing, the scale up etcd function is not working in all situations, not only `if you use CRI-O`

install_config/adding_hosts_to_existing_cluster.adoc
gpei: Etcd scale-up using playbook is the same with previous release, no extra separate process needed.

kalexand-rh: Great. Thank you!

day_two_guide/topics/proc_restoring-etcd.adoc
```diff
@@ -252,3 +259,53 @@ ifeval::["{context}" == "downgrade"]
 # journalctl -fu etcd.service
 endif::[]
 ----
+
+== Restoring etcd on a static pod
+
+Before restoring etcd on a static pod:
+* `etcdctl` binaries must be available or, in containerized installations, the `rhel7/etcd` container must be available
```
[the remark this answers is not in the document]
vrutkovs: `etcdctl` in 3.10+ is available in etcd static pod

day_two_guide/topics/proc_restoring-etcd.adoc
```diff
@@ -252,3 +259,53 @@ ifeval::["{context}" == "downgrade"]
 # journalctl -fu etcd.service
 endif::[]
 ----
+
+== Restoring etcd on a static pod
+
```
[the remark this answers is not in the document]
kalexand-rh: @jiajliu, does this update help?

kalexand-rh:
@jiajliu, I made some changes. Will you PTAL?

kalexand-rh:
Thank you @jiajliu! I think I've addressed all of @gpei's comments, so I'm going to squash and merge and follow up on the bugs.

day_two_guide/topics/proc_backing-up-etcd.adoc
geliu2016: @sdodson, I tried '/var/lib/etcd/snapshot.db', it work well for mount to master's local dir, but '/etc/etcd/snapshot.db' fail to mount to master's local dir, is this problem or work as design? thx 
@kalexand-rh, I viewed that you wrote the '/var/lib/etcd/snapshot.db' in doc, so let's waiting for scott's response, if this is work as design, pls update the doc, if it's a problem, I will file a bug for it, and the doc needn't be updated. thx

kalexand-rh: @gpei, Scott says that the '/etc/etcd/snapshot.db' path should be fine.

kalexand-rh: @geliu2016, Scott says that '/etc/etcd/snapshot.db' should work. I'll update it to '/var/lib/etcd/snapshot.db' since that's **actually** working, though: https://github.com/openshift/openshift-docs/pull/12454

geliu2016: @kalexand-rh, ok, I think the doc is ok, and I double confirmed that '/etc/etcd/snapshot.db' doesn't work at least in 3.11, so I will file function bug to scott about this, thanks for your great work on this issue.
@sdodson , I file a bug for this issue as we discussed above, thx
https://bugzilla.redhat.com/show_bug.cgi?id=1638673

kalexand-rh: Thanks! I'm glad that this process is correct again. I'll ping you in all the doc bugs that I think we can close now. :)

PR #12324, opened 2018-10-04
removing 3.10 references

kalexand-rh:
@openshift/team-documentation, Vikram asked me to look through master and see what references to 3.10 should be removed in 3.11. Will you PTAL and see if you agree with these changes?


kalexand-rh:
@vikram-redhat approved this change via email.

PR #12325, opened 2018-10-04
removing 3.10 image references

kalexand-rh:


bmcelvee:
@gabemontero this PR is updating the Jenkins files - PTAL

Thanks!

gabemontero:
Hmmm ... even though the images are deprecated, they still exist ... they just are not longer supported

@bparees is my understanding correct ^^ ?

If so, I don't think we want this update

bparees:
> @bparees is my understanding correct ^^ ?

your understanding is correct (they are deprecated, the images still exist) but i don't see a problem w/ removing the documentation from the 3.11+ version of the docs.

kalexand-rh:
@bparees, are you ok with the proposed change, or would you like to use different language in 3.11?

bparees:
@kalexand-rh it's fine for 3.11.

kalexand-rh:
Thanks @bparees! Merging.

gabemontero:
ok ... i suppose we can always add something back if we get a ton of questions, etc.

PR #12346, opened 2018-10-05
bug 1583029 adding Azure to cloud provider list

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1583029

@jianlinliu, PTAL?

### DIFF
--- a/install_config/cfme/index.adoc
+++ b/install_config/cfme/index.adoc
@@ -75,7 +75,8 @@ following storage class options to back your PVs:
 |NFS somewhere else, like a storage appliance
 
 |Cloud Provider
-|Use automatic storage provisioning from your cloud provider (GCE or AWS)
+|Use automatic storage provisioning from your cloud provider (Google Cloud
+Engine, Amazon Web Services, or Microsoft Azure)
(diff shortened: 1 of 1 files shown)


gpei:
Per https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_management/tasks/validate.yml#L58 , I think it's ok for me

gpei:
After checking the original bug, I just found it was report for ocp-3.9, but in 3.9
https://github.com/openshift/openshift-ansible/blob/release-3.9/roles/openshift_management/tasks/validate.yml#L58
we only support 'aws' and 'gce' as working providers

Azure support for management storage class was added since openshift-ansible-3.10.0-0.22.0
https://github.com/openshift/openshift-ansible/commit/876f5f00d21f88ab73257797a999148d2b643d25

kalexand-rh:
Thank you @gpei! I cloned the bug and resubmitted it as an RFE to get the feature into 3.9. I'm going to merge this change and CP to 3.10 and 3.11 since it's a docs bug in those versions.

PR #12357, opened 2018-10-08
Removing Docker references 

kalexand-rh:


kalexand-rh:
@vikram-redhat, @smarterclayton, PTAL

_topic_map.yml
```diff
@@ -283,7 +283,7 @@ Topics:
 - Name: Disconnected installation
   File: disconnected_install
   Distros: openshift-enterprise
-- Name: InInstalling a stand-alone deployment of OpenShift container registry
+- Name: InInstalling a stand-alone deployment of OpenShift container image registry
```
vikram-redhat: Looks like a long standing typo :)

vikram-redhat:
@kalexand-rh I am happy with it. If you don't hear from anyone else, go ahead and merge for 3.11 release and close the originating PR.

_topic_map.yml
kalexand-rh: Luckily it's just in 3.11, so this PR will cover it. :)

PR #12371, opened 2018-10-09
"oc volume" command is renamed to "oc set volume" command

nekop:
`oc volume` is removed in 3.11.


ahardin-rh:
@nekop Thank you!

PR #12374, opened 2018-10-09
Fix etcd scaleup playbook location

aerdei:


kalexand-rh:
That's the right location: https://github.com/openshift/openshift-ansible/tree/release-3.9/playbooks/openshift-etcd

PR #12375, opened 2018-10-09
Update basic audit log example

aerdei:
When auditConfig is enabled, the logs found in /var/log/audit-ocp.log are different from the ones found in the docs examples. The same scenario was recreated to get the same results with the new log messages.

### DIFF
--- a/install_config/master_node_configuration.adoc
+++ b/install_config/master_node_configuration.adoc
@@ -1133,8 +1133,8 @@ Each audit log contains two entries:
 Example output for user *admin* asking for a list of pods:
 
 ----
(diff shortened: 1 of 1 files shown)


kalexand-rh:
@vikram-redhat?

mburke5678:
@aerdei LGTM. What versions of OpenShiftis this for?

aerdei:
@mburke5678 tested with openshift v3.9.43

soltysh:
That's not quite right, see https://bugzilla.redhat.com/show_bug.cgi?id=1639607 the output became configurable by accident. We need to update the docs about that, not changing the sample output.

PR #12406, opened 2018-10-10
Warning about incomplete download

kalexand-rh:
@vikram-redhat PTAL


kalexand-rh:
VIkram +1'd this in Slack.

Issue #12200, opened 2018-09-27
openshift_node_kubelet_args is deprecated


kalexand-rh:
In 3.10 and 3.11. See https://github.com/openshift/openshift-docs/pull/12154/files#r220841792 and https://github.com/openshift/openshift-docs/pull/12416#issuecomment-430138176

PR #12416, opened 2018-10-10
fixing /byo/ in paths

kalexand-rh:
In 3.9, many playbooks were moved out of the /byo/ folder. This PR is to the fix remaining incorrect references to that folder.

### DIFF
--- a/day_two_guide/topics/pruning_images_and_containers.adoc
+++ b/day_two_guide/topics/pruning_images_and_containers.adoc
@@ -173,7 +173,7 @@ After modifying the `Ansible` variable content, running the installer again
 may configure the new settings and restart the services:
 
 ----
-$ ansible-playbook /usr/share/ansible/openshift-ansible/playbooks/byo/config.yml
+$ ansible-playbook /usr/share/ansible/openshift-ansible/playbooks/deploy_cluster.yml
 ----
 
 WARNING: The `kubeletArguments` section is overwritten with the content of the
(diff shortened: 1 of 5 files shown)


day_two_guide/topics/pruning_images_and_containers.adoc
```diff
@@ -173,7 +173,7 @@ After modifying the `Ansible` variable content, running the installer again
 may configure the new settings and restart the services:
 
 ----
-$ ansible-playbook /usr/share/ansible/openshift-ansible/playbooks/byo/config.yml
+$ ansible-playbook /usr/share/ansible/openshift-ansible/playbooks/openshift-master/config.yml
```
kalexand-rh: Scott says that this should be /playbooks/deploy_cluster.yml

kalexand-rh:
@jianlinliu, will you PTAL?

jianlinliu:
@kalexand-rh the /byo/ path fix looks good. After a glace, I found "openshift_node_kubelet_args" is using in pruning_images_and_containers.adoc, but actually this would not take effect during installation, the kubelet arguments setting have to be configured by configmap since 3.10.

kalexand-rh:
Thank you @jianlinliu! I'll merge this change and move https://github.com/openshift/openshift-docs/issues/12200 up on my priority list to address that. Please let me know if you come across any other issues with the docs.

jianlinliu:
@kalexand-rh the /byo/ path fix looks good. After a glace, I found "openshift_node_kubelet_args" is using in pruning_images_and_containers.adoc, but actually this would not take effect during installation, the kubelet arguments setting have to be configured by configmap since 3.10. Open a new bug to track it - BZ#1639958.

PR #12424, opened 2018-10-11
Correct each storage class name as each component name in example

bysnupy:
The parameter names are duplicated/repeated same one in logging and metrics component sections in example.

I think just one correct parameter is need to each section in each example, and it's too long to display their page layout either.


kalexand-rh:
This change looks reasonable to me. It goes back to 3.7, but I'll need to manually change 3.9 and 3.7 because the text is in a different file. Merging.

PR #12436, opened 2018-10-11
disable ansible 2.4

kalexand-rh:
@jiajliu, @sdodson, I see at least one report on the mailing list that you need to disable ansible-2.4 during upgrade. Does this change sound right to you?

### DIFF
--- a/upgrading/automated_upgrades.adoc
+++ b/upgrading/automated_upgrades.adoc
@@ -179,6 +179,7 @@ repository, if it is not already enabled:
 +
 ----
 # subscription-manager repos --disable="rhel-7-server-ose-3.10-rpms" \
+    --disable="rhel-7-server-ansible-2.4-rpms" \
     --enable="rhel-7-server-ose-3.11-rpms" \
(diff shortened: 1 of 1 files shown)


sdodson (approved):
Not disabling ansible 2.4 repos shouldn't create any harm, but it's fine to do so.
/lgtm

upgrading/automated_upgrades.adoc
```diff
@@ -179,6 +179,7 @@ repository, if it is not already enabled:
 +
 ----
 # subscription-manager repos --disable="rhel-7-server-ose-3.10-rpms" \
+    --disable="rhel-7-server-ansible-2.4-rpms"
```
sdodson: oops, needs trailing \

sdodson (changes requested):

upgrading/automated_upgrades.adoc
kalexand-rh: You're right! Fixed. :)

kalexand-rh:
@jiajliu, are you ok with this change?

PR #12454, opened 2018-10-11
update snapshot directory

kalexand-rh:
Updating directory per @geliu2016


kalexand-rh:
@geliu2016 says that the doc's ok over here, so I'm merging. https://github.com/openshift/openshift-docs/pull/12322#discussion_r224715398

PR #12466, opened 2018-10-12
Bug 1591611 - corrected command sequence for adding new 0Auth app

jhoyt-rh:
PR for correcting command sequence for registering a new application on Git Hub

Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1591611


jhoyt-rh:
Had to make changes due to the UI having changed, but the PR is right.

PR #12473, opened 2018-10-12
Bug 1591611 - added info, corrected commands -- adding new 0Auth app

jhoyt-rh:
Corrected command sequence and added one more line about click button.

GUI has changed, but some info in bug is correct.

Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1591611

@kalexand-rh, PTAL

### DIFF
--- a/install_config/configuring_authentication.adoc
+++ b/install_config/configuring_authentication.adoc
@@ -1411,7 +1411,8 @@ specific GitHub organizations.
 . Register an application:
 ** For GitHub, click https://github.com/settings/profile[Settings] ->
 https://github.com/settings/developers[Developer settings] ->
-https://github.com/settings/applications/new[Register a new OAuth application].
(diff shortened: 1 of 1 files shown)


install_config/configuring_authentication.adoc
```diff
@@ -1411,7 +1411,8 @@ specific GitHub organizations.
 . Register an application:
 ** For GitHub, click https://github.com/settings/profile[Settings] ->
 https://github.com/settings/developers[Developer settings] ->
-https://github.com/settings/applications/new[Register a new OAuth application].
+https://github.com/settings/applications/new[Register a new OAuth application]->
+https://github.com/settings/applications/new[New OAuth App].
```
kalexand-rh: I still don't see the **New OAuth App** button on that GH page.

kalexand-rh (commented):
I don't see that button on GH, but it otherwise looks fine.

PR #12476, opened 2018-10-12
[Enterprise-3.11] Bug 1591611 - added info and corrected command sequence for adding ne…

jhoyt-rh:
From #12466


jhoyt-rh:
@vikram-redhat PTAL

PR #12517, opened 2018-10-17
Update repos orgazation to sclorg

liangxia:
@openshift/team-documentation Please take a look.

It should apply to all enterprise-3.x versions.

### DIFF
--- a/dev_guide/dev_tutorials/ruby_on_rails.adoc
+++ b/dev_guide/dev_tutorials/ruby_on_rails.adoc
@@ -12,7 +12,7 @@ toc::[]
 
 == Overview
 
-Ruby on Rails is a popular web framework written in https://github.com/openshift/mysql/tree/master/5.5[Ruby].
+Ruby on Rails is a popular web framework written in https://github.com/sclorg/mysql-container/tree/master/5.5[Ruby].
 This guide covers using Rails 4 on {product-title}.
 
 [WARNING]
--- a/getting_started/beyond_the_basics.adoc
+++ b/getting_started/beyond_the_basics.adoc
@@ -429,11 +429,11 @@ and try out Quickstart templates for the following languages:
 
 Other images provided by {product-title} include:
 
-* https://github.com/openshift/mysql[MySQL]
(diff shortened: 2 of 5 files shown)


vikram-redhat:
@bmcelvee - PTAL.

vikram-redhat:
@bmcelvee - were you able to look at this?

liangxia:
A simple way to check the pr is valid.

Take the first change as example.
It changes https://github.com/openshift/mysql to https://github.com/sclorg/mysql-container.
You can verify by access the origin/first url, and you will find out that it redirects you to the changed/second one.

bmcelvee:
Thanks, @liangxia! LGTM

bmcelvee:
I can cherry-pick this PR back to 3.6 fairly easily, but with restructuring 3.1-3.5 will present some challenges.

bmcelvee:
I've CP'd back to 3.6 (merged 3.7-3.11). Starting with 3.6 and older there are conflicts with the versions supported with OCP. Should we stop updating 3.6 and earlier or open separate PRs to update the links but not the versions?

liangxia:
> I've CP'd back to 3.6 (merged 3.7-3.11). Starting with 3.6 and older there are conflicts with the versions supported with OCP. Should we stop updating 3.6 and earlier or open separate PRs to update the links but not the versions?

I'll open a separate pr to fix the 3.6 and earlier versions.

vikram-redhat:
@liangxia - thanks. Just FYI - Changes to some of the files in the follow up PRs will have no effect as these are driven from the master branch. For example, _distro_map.yml and anything under templates. Could you please keep the changes only to what was suggested by this PR?

liangxia:
@vikram-redhat  Updated the 3.1 one, does it look good to you ? If yes, I will apply the change to the other ones.

vikram-redhat:
@liangxia - looks good! Thanks

liangxia:
3.2 ~ 3.6 are update, too.

bmcelvee:
Thank you so much, @liangxia!

PR #12043, opened 2018-09-14
Installation Steps for OpenShift 3.10 on OpenStack 13

rlopez133:
I'm creating this PR that includes the steps to incorporate into the OpenShift Docs how to deploy OpenShift 3.10 on OpenStack 13. I've attempted to modularize as much possible into different adocs.

PR #12519, opened 2018-10-17
OpenStack on OCP Ref Arch docs

vikram-redhat:
This is a follow up to fix the build issues in the original PR: #12043 and squash the commits.

This has been QE reviewed by @shlao and peer reviewed by @ahardin-rh . @ahardin-rh if there are more edits required, can you please follow up?

cc: @rlopez133 @aheslin

### DIFF
--- a/install_config/configuring_kuryrsdn.adoc
+++ b/install_config/configuring_kuryrsdn.adoc
@@ -104,7 +104,7 @@ Kuryr-Kubernetes as the network plug-in:
 ----
 
 You must also specify an OpenStack cloud provider as described in the
-xref:configuring_openstack.adoc#openstack-configuring-masters-ansible[OpenStack configuration documentation].
+xref:configuring_openstack.adoc#install-config-configuring-openstack[OpenStack configuration documentation].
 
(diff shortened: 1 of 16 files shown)


vikram-redhat:
@rlopez133 - I don't know if this would apply directly to the 3.11 branch? Can you confirm? If yes, I will CP to that branch as well. At the minimum, I guess we should need to change the comments from the original PR: https://github.com/openshift/openshift-docs/pull/12043#issuecomment-422578125

vikram-redhat:
@ahardin-rh - some substitutions are not working. For example: {rhosp}: https://docs.openshift.com/container-platform/3.10/install_config/configuring_openstack.html. And I think there is an issue with headings as well. Can you please take a look and do a follow up?

rlopez133:
It should be similar for OCP 3.11. Also, versions would need to be fixed for Ansible and OCP if you were to place it in the 3.11 branch.

vikram-redhat:
Thanks @rlopez133. @ahardin-rh  - can you please do a follow up for 3.11?

PR #12538, opened 2018-10-17
Bug 1437137, Added storage management guidance

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1437137

### DIFF
--- a/install/prerequisites.adoc
+++ b/install/prerequisites.adoc
@@ -163,6 +163,51 @@ Kubernetes scheduler makes during pod placement. Learn what measures you can
 take to xref:../admin_guide/overcommit.adoc#disabling-swap-memory[avoid memory swapping].
 ====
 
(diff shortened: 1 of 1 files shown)


mffiedler:
Nit:  Looks like other instances in the doc user CRI-O and CRIO in caps.   I know...I used cri-o in the bz.   Whatever is correct from a docs point of view is fine with me.

Example:  https://docs.openshift.com/container-platform/3.11/install_config/configuring_openstack.html#configuring-a-security-group-openstack

portante:
Are we keeping in mind that RHEL uses `/var/log` for sysstat data (`/var/log/sa`) and systemd uses it for persistent journal data (`/var/log/journal`), while by default rsyslog and other log configurations use `/var/log` in random ways?

mffiedler (approved):

Issue #12409, opened 2018-10-10
oc export is deprecated, docs say otherwise


Askill:
### Which section(s) is the issue in?
https://docs.okd.io/latest/dev_guide/templates.html

### What needs fixing?
oc export ist deprecated accoring to the console output.
oc get --export is not documented.

bysnupy:
FYI, as I remember, `oc get --export` is not same the result with former `oc export` if you want to export `yaml` format from target `objects`.  Additionally  it needs to specify `-o yaml` for yaml format to export it.

Askill:
Thanks fot the tip however that ist what following message suggests, either this or the Documentation should be changed.
![image](https://user-images.githubusercontent.com/16598120/46936969-59fd2400-d060-11e8-875f-41278613ab0d.png)

ahardin-rh:
This was addressed in https://github.com/openshift/openshift-docs/pull/12579

PR #12579, opened 2018-10-21
Replace deprecated export command with new one

bysnupy:
`oc export` command is deprecated since `v3.10`.

Personally I think we should start to take effect new one in documentation at the moment.

At least, we should inform the deprecated status for users at the sections which used `oc export`, if the changes can not update on current documentation.

As far as I know, new `oc get --export` does not implement same feature with old `oc export`, such as `--as-template`, `--output-version` and so on. So we might provide workarounds, such as edit manually something the form, if we can.

Related info:
- issue: 
  - https://github.com/openshift/origin/issues/20813
  - https://github.com/openshift/openshift-docs/issues/12409
- BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1641309

### DIFF
--- a/admin_guide/manage_scc.adoc
+++ b/admin_guide/manage_scc.adoc
@@ -62,7 +62,7 @@ restricted         false     []        MustRunAs   MustRunAsRange     MustRunAs 
 [[examining-a-security-context-constraints-object]]
 == Examining a Security Context Constraints Object
 
-To examine a particular SCC, use `oc get`, `oc describe`, `oc export`, or `oc
+To examine a particular SCC, use `oc get`, `oc describe`, or `oc
 edit`. For example, to examine the *restricted* SCC:
 
 ----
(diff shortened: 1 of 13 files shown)


vikram-redhat:
@ahardin-rh - PTAL.

ahardin-rh:
@bysnupy Thanks for this! Overall, this LGTM. I will do a follow-up PR once this merges to make one or two very small style edits.

@xltian Can you please provide a QE review? Thanks!

xltian:
@xingxingxia  Will review from QE.

xingxingxia:
@zhouying7780 could you help review? I'm on another larger one #12583
@ahardin-rh btw, when you're available pls also use another PR to update https://bugzilla.redhat.com/show_bug.cgi?id=1540890 esp its https://bugzilla.redhat.com/show_bug.cgi?id=1540890#c1
Thanks

zhouying7780:
LGTM.

ahardin-rh:
@xingxingxia Thanks! I assigned https://bugzilla.redhat.com/show_bug.cgi?id=1540890 to myself. (FYI @vikram-redhat )

@zhouying7780 Thanks for the review!

PR #12587, opened 2018-10-22
BZ1633471 updating node removal commands

kalexand-rh:
Update to https://bugzilla.redhat.com/show_bug.cgi?id=1633471


kalexand-rh:
This update was approved on the bug. Merging.

PR #12629, opened 2018-10-24
Fix etcd backup and restore for API v2 and v3

juanluisvaladas:
Docs seem to say that v2 backup does not include v3 data and
vice versa. Both backup all the data, and when restoring, restore
one is enough.


kalexand-rh:
I'm opening a new PR to make some edits and have a bit more discussion: https://github.com/openshift/openshift-docs/pull/12704

PR #12630, opened 2018-10-24
bug 1622044 correcting audit log path

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1622044

@jianlinliu, will you PTAL?


kalexand-rh:
This change was approved on the source bug.

PR #12631, opened 2018-10-24
bug 1639080 clarifying insecure connection override

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1639080

@ericavonb, how does this sound?

@jianlinliu, will you PTAL?

This bug was filed against 3.10 - do you know how far back this change is true?

### DIFF
--- a/install_config/syncing_groups_with_ldap.adoc
+++ b/install_config/syncing_groups_with_ldap.adoc
@@ -61,9 +61,12 @@ necessary to retrieve entries for the sync operation. This value may also be
 provided in an
 xref:../install_config/master_node_configuration.adoc#master-node-configuration-passwords-and-other-data[environment
 variable, external file, or encrypted file].
-<4> When `true`, no TLS connection is made to the server. When `false`, secure
+<4> When `false`, secure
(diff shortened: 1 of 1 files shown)


ericavonb:
@kalexand-rh lgtm!

kalexand-rh:
@stuartchuan, PTAL?

kalexand-rh:
@stuartchuan, the bug says that it's taking effect anyway.

kalexand-rh:
@stuartchuan,  will you PTAL?

stuartchuan:
I still think this is not the root cause fix, just a workaround. I think the bug should be fixed in code but not doc.
@ericavonb what do you think?

kalexand-rh:
@ericavonb, when you have a minute, will you PTAL?

ericavonb:
@kalexand-rh @stuartchuan I think this is the correct behavior (with the docs here) in accord with the standards around ldap urls and tls.

kalexand-rh:
Thanks @ericavonb. @stuartchuan, if this is correct, can I merge this to 3.10 and 3.11?

stuartchuan:
@kalexand-rh it's OK to merge.

kalexand-rh:
Thanks! Merging.

PR #12670, opened 2018-10-26
Fix bug 1643426 , 404 url

liangxia:
https://bugzilla.redhat.com/show_bug.cgi?id=1643426


liangxia:
Applies to enterprise-3.9 and above.

kalexand-rh:
Thank you @liangxia!

PR #12692, opened 2018-10-29
The osm_cluster_network_cidr, openshift_portal_net, and osm_host_subnet_length no longer set by default.

ahardin-rh:


ahardin-rh:
This continues the work in https://github.com/openshift/openshift-docs/pull/5303. A lot has changed since the original PR, but moving this along.

ahardin-rh:
This all boiled down to a word change, so applying to 3.11 only

PR #12695, opened 2018-10-30
Fix bug 1644088

liangxia:
Fix https://bugzilla.redhat.com/show_bug.cgi?id=1644088


liangxia:
Applies to enterprise-3.9 and above.

PR #12705, opened 2018-10-30
Update contributing.adoc

sudeshnas25:


kalexand-rh:
@sudeshnas25, thanks for the PR. If you were to list out each distro, you'd use the current syntax, without the spaces. In this context, the text is correct.

PR #12607, opened 2018-10-23
bug 1640600 updating Apache HTTPD REMOTE_USER definition

kalexand-rh:
From https://bugzilla.redhat.com/show_bug.cgi?id=1640600

@stuartchuan, PTAL?

PR #12711, opened 2018-10-30
updates to Apache config

kalexand-rh:
@enj, FYI

Follow up to https://github.com/openshift/openshift-docs/pull/12607

@stuartchuan, PTAL?

### DIFF
--- a/install_config/configuring_authentication.adoc
+++ b/install_config/configuring_authentication.adoc
@@ -1369,8 +1369,27 @@ negotiate challenge, or both challenges:
     '<redirected challengeURL from step 3 +query>'
 ----
 
-. Confirm that you can log in to the `oc` command line using your Kerberos ticket:
-.. If you have not done so, provide your Kerberos credentials to get a ticket:
+. Test logging into the `oc` command line with and without using a Kerberos ticket:
+.. If you generated a Kerberos ticket by using `kinit`, destroy it:
++
+----
+# kdestroy -c cache_name <1>
+----
+<1> Provide the name of your Kerberos cache.
+.. Log in to the `oc` command line by using your Kerberos credentials:
++
+----
+# oc login
+----
++
+Enter your Kerberos user name and password at the prompt. 
+.. Log out of the `oc` command line:
++
+----
+# oc logout
+----
+.. Use your Kerberos credentials to get a ticket:
++
 ----
 # kinit
 ----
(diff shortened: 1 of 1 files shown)


stuartchuan:
Why does we change the htpasswd to SSPI connection for Apache example?

enj:
> Why does we change the htpasswd to SSPI connection for Apache example?

Moving away from the "hello world" `htpasswd` example to realistic GSSAPI example.  Our examples should focus on configurations that customers would actually use.  The `htpasswd` config is trivial and fully automateable, and has no real connection with actually getting complex configurations like GSSAPI working correctly.

stuartchuan:
As the GSSAPI still is Technology Preview features, does it reasonable to write here?

enj:
GSSAPI on Linux has been fully supported for years.  Only SSPI on Windows is in preview.

enj:
Note also that these docs refer to the server side configuration (and not client side features) which is the same between GSSAPI and SSPI.  The server side request header code has not changed in years.

install_config/configuring_authentication.adoc
```diff
@@ -1118,12 +1118,12 @@ location.
...
 
-. The *mod_auth_form* module is shipped as part of the *mod_session* package that
-is found in the link:https://access.redhat.com/solutions/392003[Optional channel].
+. Obtain the *mod_auth_gssapi* module from the
+link:https://access.redhat.com/solutions/392003[Optional channel].
 Install the following packages:
 +
 ----
-# yum install -y httpd mod_ssl mod_session apr-util-openssl
+# yum install -y httpd mod_ssl mod_session apr-util-openssl mod_auth_gssapi
```
stuartchuan: `httpd` not need here.

stuartchuan: The following ` # systemctl restart httpd` don't need also

install_config/configuring_authentication.adoc
```diff
@@ -1371,19 +1361,29 @@ additional query parameters):
...
+.. Confirm that you can log in to the `oc` command line:
 +
 ----
-#  curl -k -v -u <your_user>:<your_password> \
-    -H 'X-Csrf-Token: 1' '<redirected_challengeURL_from_step_3 +query>'
+# oc login
 ----
++
+If your configuration is correct, you are logged in without entering separate
+credentials.
```
stuartchuan: There should add a different scenario, when no kerberos ticket issued or no krb5 workstation configured on the workstation, should fall back to basic authentication.

kalexand-rh: Ok. @enj, do you have any objections to the new commit?

kalexand-rh:
@enj, @stuartchuan, is this one good to merge?

stuartchuan:
The changes are OK for me now.

PR #12729, opened 2018-10-31
bug 1640881 etcd note

kalexand-rh:
From https://bugzilla.redhat.com/show_bug.cgi?id=1640881

@jianlinliu, will you PTAL? I think this should be valid in 3.10 as well as 3.11.


jianlinliu:
@geliu2016 would take care of this review.

kalexand-rh:
Thanks! @geliu2016 approved this change on the source bug. Merging.

PR #12736, opened 2018-11-01
Fix issues#12730: Correct the hostPath indentation

bysnupy:
Fix issue: https://github.com/openshift/openshift-docs/issues/12730


kalexand-rh:
@jhou1, will you PTAL? If you agree, will you also confirm that it's valid in only 3.10 and 3.11?

jhou1:
@kalexand-rh looks nice, this is also valid for 3.x versions.

kalexand-rh:
Thanks @jhou1! I'll try to send it back to 3.6, then. :)

kalexand-rh:
This particular example was added in 3.10, so that's as far back as it's going. :)

PR #12748, opened 2018-11-01
 Logging Installer forgets to create PV when using NFS

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1639750

### DIFF
--- a/install_config/aggregate_logging.adoc
+++ b/install_config/aggregate_logging.adoc
@@ -724,6 +724,13 @@ named to match the `openshift_logging_es_pvc_prefix` setting, which defaults to
 exists already, it is used; if not, and `openshift_logging_es_pvc_size` has been
 specified, it is created with a request for that size.
 
+If using xref:../install_config/persistent_storage/dynamically_provisioning_pvs.adoc#install-config-persistent-storage-dynamically-provisioning-pvs[dynamically provisioned PVs], the {product-title} logging installer creates PVCs that use the default storage class or the PVC specified with the `openshift_logging_elasticsearch_pvc_storage_class_name` parameter.
+
+If using NFS storage, the {product-title} installer creates the persistent volumes, based on the `openshift_logging_storage_*` parameters 
+and the xref:#deploying-the-efk-stack[{product-title} logging installer] creates PVCs, using the `openshift_logging_es_pvc_` paramters.
+Make sure you specify the correct parameters in order to use persistent volumes with EFK.
+Also set the `openshift_enable_unsupported_configurations=true` parameter in the Ansible inventory file, as the logging installer blocks the installation of NFS with core infrastructure by default.
+
 [WARNING]
 ====
 Using NFS storage as a volume or a persistent volume (or via NAS such as


jcantrill:
> @jcantrill @ewolinetz PTAL
> 
> I added a paragraph:
> _The {product-title} installer creates the persistent volumes, based on the `openshift_logging_storage_*` parameters and the xref:#deploying-the-efk-stack[EFK deployment] creates PVCs, using the openshift_logging_es_pvc_ paramters.
> _Make sure you specify the correct parameters in order to use persistent volumes with EFK._
> 
I don't think this is technically correct as the platform created PV's, not the installer.  The installer may create storage classes which tell the platform what storage options are available, but I don't know enough about these vars or how these vars drive installation to provide more info.  @sdodson Do you have info you may provide regarding these parameters?  I don't recall our team adding them.

> Does this meet your concerns in the BZ? Is this accurate?

ewolinetz:
the installer creates the PVs in the case where its NFS storage. There is a role to create such PVs. In the case of storage classes and dynamic provisioning the platform will create the PV there.

mburke5678:
@ewolinetz @jcantrill 
For the purposes of this PR, is it important to the end user whether the installer or the platform creates the PVs? Or, it is enough to state that OpenShift creates the PVs based on the _openshift_logging_storage_*_ parameters and the PVCs based on the  _openshift_logging_es_pvc parameters_. So be sure to specify both sets of parameters?

install_config/aggregate_logging.adoc
```diff
@@ -724,6 +724,10 @@ named to match the `openshift_logging_es_pvc_prefix` setting, which defaults to
 exists already, it is used; if not, and `openshift_logging_es_pvc_size` has been
 specified, it is created with a request for that size.
 
+The {product-title} installer creates the persistent volumes, based on the `openshift_logging_storage_*` parameters 
+and the xref:#deploying-the-efk-stack[EFK deployment] creates PVCs, using the openshift_logging_es_pvc_ paramters.
```
ewolinetz: instead of `[EFK deployment]` can we use something like `[{product-title} Logging installer]`?

install_config/aggregate_logging.adoc
```diff
@@ -724,6 +724,10 @@ named to match the `openshift_logging_es_pvc_prefix` setting, which defaults to
 exists already, it is used; if not, and `openshift_logging_es_pvc_size` has been
 specified, it is created with a request for that size.
 
+The {product-title} installer creates the persistent volumes, based on the `openshift_logging_storage_*` parameters
```
ewolinetz: installer creates PVs specifically in the case of a NFS backing

ewolinetz:
I think we should have something in there about dynamic provisioning too... otherwise lgtm

mburke5678:
@ewolinetz What do we need to say about dynamic provisioning? I am unfamiliar with storage.

install_config/aggregate_logging.adoc
```diff
@@ -724,6 +724,10 @@ named to match the `openshift_logging_es_pvc_prefix` setting, which defaults to
 exists already, it is used; if not, and `openshift_logging_es_pvc_size` has been
 specified, it is created with a request for that size.
 
+If using NFS storage, the {product-title} installer creates the persistent volumes, based on the `openshift_logging_storage_*` parameters 
+and the xref:#deploying-the-efk-stack[{product-title} logging installer] creates PVCs, using the openshift_logging_es_pvc_ paramters.
+Make sure you specify the correct parameters in order to use persistent volumes with EFK.
```
ewolinetz: Maybe add something above this like:
"If using xref:install_config/persistent_storage/dynamically_provisioning_pvs.adoc#install-config-persistent-storage-dynamically-provisioning-pvs[dynamically provisioned PVs] then {product-title} logging installer will create PVCs that will use the default storage class or the one specified with the `openshift_logging_elasticsearch_pvc_storage_class_name` parameter.

mburke5678:
@xltian PTAL

install_config/aggregate_logging.adoc
```diff
@@ -724,6 +724,12 @@ named to match the `openshift_logging_es_pvc_prefix` setting, which defaults to
...
```
ewolinetz: I think line 729 should be after line 731

mburke5678:
@anpingli PTAL

cc. @wsun1

install_config/aggregate_logging.adoc
```diff
@@ -724,6 +724,12 @@ named to match the `openshift_logging_es_pvc_prefix` setting, which defaults to
 exists already, it is used; if not, and `openshift_logging_es_pvc_size` has been
 specified, it is created with a request for that size.
 
+If using xref:../install_config/persistent_storage/dynamically_provisioning_pvs.adoc#install-config-persistent-storage-dynamically-provisioning-pvs[dynamically provisioned PVs], the {product-title} logging installer creates PVCs that use the default storage class or the one specified with the `openshift_logging_elasticsearch_pvc_storage_class_name` parameter.
+
+If using NFS storage, the {product-title} installer creates the persistent volumes, based on the `openshift_logging_storage_*` parameters
```
anpingli: I think we should mention two points
[1]The logging deploy playbook don't trigger nfs storage allocation. you must run the cluster deployment playbook  /usr/share/ansible/openshift-ansible/playbooks/deploy_cluster.yml
[2] For nfs is unssupported, you have to set openshift_enable_unsupported_configurations=true
 
1) this variable must be set openshift_enable_unsupported_configurations=true.

mburke5678: @jcantrill @anpingli I am moving these suggested changes to a new PR so we can merge this PR independently. 
https://github.com/openshift/openshift-docs/pull/12834

mburke5678:
@anpingli @ewolinetz  
Can we move this PR forward without the changes suggested above in https://github.com/openshift/openshift-docs/pull/12748#discussion_r232602303?

ewolinetz:
@mburke5678 I am alright with moving forward as is.

mburke5678:
@anpingli @ewolinetz 

How far back should we backport these changes?

mburke5678:
@mburke5678 Move changes to 4.0 docs when approved.

mburke5678:

ewolinetz:
@mburke5678 I don't think we need to backport past 3.7 since that is when we started to have the ansible role as it exists currently. But we would need to confirm the variable names as they may have been different then.

PR #12749, opened 2018-11-01
Add missing backslash after registry-console

jataylor-redhat:


kalexand-rh:
Thanks @jataylor-redhat!

PR #12753, opened 2018-11-02
Fix typo in image change trigger

szpak:
It's especially unpleasant as using `imagechange` instead of `imageChange` makes the trigger to be silently ignored on apply.


kalexand-rh:
@wzheng1, this looks reasonable to me, but will you please confirm?

wzheng1:
LGTM, thanks!

kalexand-rh:
Thanks! Merging. :)

Issue #12758, opened 2018-11-02
[enterprise-3.3] Edit suggested in file install_config/install/stand_alone_registry.adoc


schafezp:
### Which section(s) is the issue in?
About OpenShift Container Registry

### What needs fixing?
The link to Atomic Registry, http://www.projectatomic.io/registry , 404s
It could be changed to an archive link or docs introduction
http://www.projectatomic.io/docs/introduction/

bysnupy:
Personally I think only your information would be hard to find where the section is located in. 
So if you can provide more details such as section url and related phrases, it would be more helpful for checking.

schafezp:
https://docs.openshift.com/container-platform/3.3/install_config/install/stand_alone_registry.html

This is the website page with the broken link.

The broken link is in the "About OpenShift Container Registry" section, directly below the second paragraph in the info section next to the blue "I".

The text reads as follows:
"OCR should not be confused with the upstream project Atomic Registry, which is a different implementation using a non-Kubernetes deployment method that leverages systemd and local configuration files to manage services."

In this paragraph "Atomic Registry" is a link to a 404 website.

Does this answer your questions?

bysnupy:
Yes, it does. Thank you for your information.

FYI, I found this article - [Update: Removed links to Atomic Registry as discontinued.](https://www.projectatomic.io/blog/2017/05/oo-standalone-registry/)

PR #12764, opened 2018-11-05
Fix the broken Atomic Registry link

bysnupy:
Related issue: https://github.com/openshift/openshift-docs/issues/12758

I would suggest new hyperlink of `Atomic Registry` instead of the broken one, because original link does not exist no more.

- New link:
[Update: Removed links to Atomic Registry as discontinued.](https://www.projectatomic.io/blog/2017/05/oo-standalone-registry/)
  - I seems good because this page inform users that `Atomic Registry` is not valid and provides more details about it.
- Target version: `v3.3` ~ `v3.7` have the section which was included the broken links.

### DIFF
--- a/install_config/install/stand_alone_registry.adoc
+++ b/install_config/install/stand_alone_registry.adoc
@@ -30,9 +30,9 @@ developer-focused web console and application build and deployment tools.
 
 [NOTE]
 ====
-OCR should not be confused with the upstream project
-link:https://www.projectatomic.io/blog/2017/05/oo-standalone-registry/[Atomic Registry], which is a
-different implementation using a non-Kubernetes deployment method that leverages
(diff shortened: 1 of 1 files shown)


kalexand-rh:
@bysnupy, I'd either remove the note or say:

```
OCR has replaced the upstream
link:https://www.projectatomic.io/blog/2017/05/oo-standalone-registry/[Atomic Registry] project, which was a
different implementation that used a non-Kubernetes deployment method that leveraged
`systemd` and local configuration files to manage services.
```

bysnupy:
@kalexand-rh 
I agree it. It already has been deprecated, and it've been removed from `v3.9` either.

vikram-redhat:
@bysnupy looks good. Merging.

PR #12801, opened 2018-11-06
install updates draft

kalexand-rh:


kalexand-rh:
@crawford, will you PTAL

modules/installation-overview.adoc
```diff
@@ -0,0 +1,130 @@
...
+
+[discrete]
+== Cluster installation process
+
+In supported cloud environment scenarios, {product-title} manages all aspects of
+the cluster, including the operating system itself. Each machine boots with a
+configuration that references resources hosted in the cluster that it joins.
+This configuration allows the cluster to manage itself as updates are applied.
+
+Because each machine in the cluster requires information about the cluster when
```
crawford: I don't think this is as clear about _why_ the bootstrap machine is necessary. There is a line in the installer docs which helps explain this:

> A downside to this approach, however, is that new clusters have no way of starting without external help - every machine in the to-be-created cluster is waiting on the to-be-created cluster.

modules/installation-overview.adoc
```diff
@@ -0,0 +1,130 @@
...
+. The temporary control plane schedules the production control plane to the
+master machines.
+. The temporary control plane shuts down and passes control to the production
+control plane.
+. The bootstrap node injects {product-title} components into the production
+control plane.
+. The installer shuts down the bootstrap node.
+
+The result of this bootstrapping process is a fully running {product-title}
+cluster. The cluster then downloads and configure remaining components
```
crawford: s/configure/configures/

modules/installation-overview.adoc
```diff
@@ -0,0 +1,130 @@
...
+=== Targets
+
+The {product-title} installer operates on the notion of creating and destroying
+targets. Similar to other tools that operate on a graph of dependencies, like
+`make` and `systemd`, each target represents a subset of the dependencies in the
+graph. The main target in the installer creates a cluster, but the other targets
+allow the user to interrupt this process and use or modify the intermediate
+artifacts, such as the the Kubernetes manifests that are installed into the
+cluster. Only the immediate dependencies of a target are written to disk by the
+installer, and you can run the installer multiple times.
```
crawford: s/and/but/

crawford:
I'm still dubious about separating the docs from the code. The more I think about this, the more nervous it makes me. There are subtle changes in meaning that come with this set of docs v.s. the original docs in the installer. I'm worried that things are going to fall out of sync (the installer is still moving really quickly) and I think we could benefit from having professional writers proofing the docs we write as well as contributing their own. I'm also likely the only person from the installer team who is aware of and watching this repo, which is going to be a bottleneck if we expect these docs to be reviewed by both developers and writers.

Maybe this isn't the correct place to raise these concerns, but I at least wanted to write them down. I don't want to hold up this process and I'm fine with this initial pass.

kalexand-rh:
@crawford, the final docs will need to be in this repo in the docs collection. It feeds both https://docs.openshift.com/ and https://access.redhat.com/documentation/en-us/openshift_container_platform/3.11/.

The installer docs can't live in your repo forever.

We're in the early phases of building out the 4.0 collection, and everything in this branch is subject to change, up to complete rewrites, if necessary, as the 4.0 release draws closer.

PR #12805, opened 2018-11-07
remove scheduler.alpha.kubernetes.io/affinity

mburke5678:


mburke5678:
@aveshagarwal PTAL. is this all that is needed to remove `scheduler.alpha.kubernetes.io/affinity`?

aveshagarwal:
@mburke5678 lgtm.

mburke5678:
Removing this text from 3.9+ based on @aveshagarwal email:
"So i think we should remove references to them from our docs for releases after than 1.7 i think."

PR #12815, opened 2018-11-08
Incorrect externalIPNetworkCIDRs example

palonsoro:
Some examples of `externalIPNetworkCIDRs:` at `/etc/origin/master/master-config.yaml` are not correct, as they omit the final `s` and setup a single value instead of an array.


kalexand-rh:
@bmeng, will you PTAL? Do you know how far back this change is valid?

kalexand-rh:
@palonsoro, if you make this change and squash your commits, I'll merge it.

PR #12871, opened 2018-11-15
Add --from-file to create secret

juanluisvaladas:
A customer faced that this command, with --from-file it works fine. According to the docs of oc create secret generic it should include the --from-file.


kalexand-rh:
@soltysh, @zhaozhanqi, PTAL? I think this change is valid back to 3.9. Do you agree?

juanluisvaladas:
I agree with 3.9.

soltysh (approved):
This is correct all the way back to 3.9.
/lgtm

kalexand-rh:
Thanks @soltysh! I'm going to merge. (@zhaozhanqi, if you disagree, I'll be happy to revert.)

PR #12880, opened 2018-11-15
4.0 arch draft

kalexand-rh:
Draft of architecture updates from the arch overview call.

http://file.rdu.redhat.com/kalexand/111618/4.0_arch/architecture/architecture.html


kalexand-rh:
@vikram-redhat, will you PTAL? Do you have a suggestion for someone who'd be willing to do a SME review of this version and look at some graphics markups?

vikram-redhat:
@kalexand-rh - thanks. @derekwaynecarr - would you be the best person to review this for docs?

vikram-redhat:
@smarterclayton and @eparis - tagging you as well for a review.

modules/architecture-overview.adoc
```diff
@@ -0,0 +1,70 @@
+// Module included in the following assemblies:
+//
+// * architecture/architecture.adoc
+
+[id='architecture-overview-{context}']
+= Architecture overview
+
+Like {product-title} v3, {product-title} v4 is a layered system designed to
+expose underlying container images and Kubernetes concepts as accurately as
+possible, with a focus on easy composition of applications by a developer.
```
smarterclayton: We need to touch this up, I’m not sure this is going to be the right way to phrase 4.  I think this is where we reset expectations for customers, and it’s going to be a combination of the core kubernetes mission (application infrastructure), the core 4 mission (automated and self-monitoring, flexible infrastructure), and the openshift story (giving developers tools to evolve their applications under operational oversight).

modules/architecture-overview.adoc
```diff
@@ -0,0 +1,70 @@
...
+For example, after your cluster is running, install Ruby, push code, and add
+MySQL to the app.
+
+A major difference between {product-title} v3 and v4 is that v4 uses Operators
+as both the fundamental unit of the product and an option for easily deploying
+and managing utilities that your apps use.
+
+{product-title} has an Operator-based architecture of smaller, decoupled units
```
smarterclayton: I think I’d focus on selling us as kubernetes first, with a set of standard components that you need  (network, ingress, logging, monitoring), made simple to scale, upgrade, and maintain through a focus on automating operations with operators.

modules/architecture-overview.adoc
```diff
@@ -0,0 +1,70 @@
...
+that work together. It runs on top of a Kubernetes cluster, with data about the
+objects stored in etcd, a reliable clustered key-value store. Those services are
+broken down by function:
+
+* Operators, which run the core {product-title} services.
```
smarterclayton: This feels a bit too much like a list - I would hit the three core layers above and give examples of each.  Operators are a detail, really, even if they underlie everything (more glue).

modules/architecture-overview.adoc
```diff
@@ -0,0 +1,70 @@
...
+applications.
+** Pods and services, which containers use to communicate with each other and
+proxy connections.
+** Projects and users, which provide the space and means for communities to
+organize and manage their content together.
+** Builds and image streams] allow you to
+build working images and react to new images.
+** Deployments, which expand support for the software development and deployment
+lifecycle.
+** Routes, which announce your service to the world.
```
smarterclayton: Ingress and Routes

modules/architecture-overview.adoc
```diff
@@ -0,0 +1,70 @@
...
+** Templates, which allow you to simultaneously create many objects that are
```
smarterclayton: It’s time to deemphasoze templates - we need to describe our catalog of supporting application infrastructure as having three parts - operators (expose apis that automate full component lifecycle like databases), service bindings (consume services running elsewhere), and templates (simple instant examples).

modules/architecture-overview.adoc
```diff
@@ -0,0 +1,70 @@
...
+based on customized parameters.
+* Controllers, which read those REST APIs, apply changes to other objects, and
+report status or write back to the object.
+
```
smarterclayton: This needs to be split out into more of separate section, probably leading with “OpenShift and Kuberneres are API first systems that take application author intent and ensure that intent is maintained, even if hardware or software fails” and then moving all these details into sub bullets.

smarterclayton: (Everything to bottom)

Then we need to go into more detail about the parts we ship and how they work together.  Probably talk about config as API (all management config is available as an API), how we do updates, and how the monitoring stack works with across the board to give admins visibility into the state of their applications.

modules/installation-options.adoc
```diff
@@ -0,0 +1,31 @@
...
+//
+// * architecture/architecture.adoc
+
+[id='installation-options-{context}']
+= Installation options
+
+{product-title} offers two installation options: fully-managed infrastructure and bring your own
+infrastructure. In version 4.0, the fully-managed option allows you to install a
+cluster in Amazon Web Services (AWS) that runs on Red Hat CoreOS nodes. If you want to
+use any other cloud or install your cluster on-premise, use the bring your own
```
smarterclayton: We need to call out that other clouds are coming soon.  This makes us look like we intend to stop at aws and won’t fly in perception.

modules/installation-options.adoc
```diff
@@ -0,0 +1,31 @@
...
+Linux (RHEL) hosts.
+
+The fully-managed option offers full-stack automation to:
+
+* Manage compute
+* Manage operating system (RHCOS)
+* Manage control plane
+* Manage nodes
+
+The bring your own infrastructure option, you have more responsibilities.
```
smarterclayton: Broken sentence

modules/installation-options.adoc
```diff
@@ -0,0 +1,31 @@
...
+You must provide the hosts and update RHEL on them. {product-title} provides:
+
+* Managed control plane
+* Ansible to manage kubelet and runtime
+
+
+With both installation types, installation and upgrade both use a controller
+that constantly reconciles component versions as if it were any other Kubernetes
```
smarterclayton: Not really true, also I would not use controller in our docs if we can use “operator” instead.  We need to nuance this a bit more.

modules/machine-api-overview.adoc
```diff
@@ -0,0 +1,56 @@
...
+//
+// * architecture/architecture.adoc
+
+[id='machine-api-overview-{context}']
+= Machine API overview
+
+On fully-managed {product-title} clusters, the Machine API performs all node
+management actions after the cluster installation finishes. Because of this
+system, {product-title} version 4.0 offers an elastic, dynamic provisioning
+method.
```
smarterclayton: ... on top of public or private cloud infrastructure.

modules/operators-overview.adoc
```diff
@@ -0,0 +1,96 @@
+// Module included in the following assemblies:
+//
+// * architecture/architecture.adoc
+
+[id='operators-overview-{context}']
+= Operators in {product-title}
+
+In {product-title} version 4.0, all cluster functions are divided into a series
+of component Operators.
```
smarterclayton: Platform Operators.  Platform operators manage a particular area of functionality like cluster wide application logging, the management of the Kubernetes control plane, or the machine provisioning system.

modules/operators-overview.adoc
```diff
@@ -0,0 +1,96 @@
...
+
+Operators can manage other Operators within their component. This lends to a
+transparent model: a single Operator usually manages a single binary file.
```
smarterclayton: No, that’s not really true (in fact it’s the opposite).

An operator provides users (admins or developers) with a simple API for expressing the desired cluster functionality.  The operator hides the details of managing the lifecycle of that component.  Operators may manage a single component or tens of components, but the end goal is always to reduce operational burden by automating common actions.

modules/operators-overview.adoc
```diff
@@ -0,0 +1,96 @@
...
+Operators also offer a more granual configuration experience. You configure each component by
+modifying its Operator instead of modifying a global configuration file.
```
smarterclayton: Modifying the API an operator exposes instead of modifying a global configuration file

modules/operators-overview.adoc
```diff
@@ -0,0 +1,96 @@
...
+
+In version 4.0, separate processes and images and pods run to drive the kubernetes
```
smarterclayton: This paragraph is a good focus but the flow seems a little weird.  I would start with

In OpenShift v4 all control plane components are run and managed as applications on the infrastructure to ensure a uniform and consistent management experience.  The control plane ...

modules/node-management.adoc
```diff
@@ -0,0 +1,46 @@
...
+
+[id='node-management-{context}']
+= Node management in fully-managed {product-title}
+
+The fully-managed version of {product-title} version 4.0 integrates management of
+the container operating system and cluster management. Because the cluster manages
+both its updates and updates to RHCOS, {product-title} provides an opinionated
+lifecycle management experience that simplifies the orchestration of node upgrades.
+
+{product-title} employs three DaemonSets and controllers to simplify node management:
```
smarterclayton: Going into details too fast - needs at least 2-3 para and a diagram that summarizes the whole lifecycle.  Also, same comment here about operators vs controllers.  

This is the signature part of v4 and needs a “ELI5” intro and a really strong visual flow so people can grasp it.

The subsequent paras are way too deep, ignition needs to be introduced at a really high level.

Others can provide more context but I might say:

1. Nodes run an operating system designed to be configured and updated as part of a larger system (RHCoS)
2. On boot, nodes have a component Ignition that asks the cluster for the appropriate configuration and software version
3. After boot, the Kubelet starts and joins the node to the cluster
4. The configuration of the machine is managed by the cluster via pods that run on each node and continuously reconcile the configuration, including software version, with the cluster

A bunch more details but would include how an update works, how a config change works, how that affects workloads, and how people are expected to interact with nodes.

We also should put the caveats about rhcos vs rhel on this page.

modules/node-types.adoc
```diff
@@ -0,0 +1,14 @@
+// Module included in the following assemblies:
+//
+// * architecture/architecture.adoc
+
+[id='node-types-{context}']
+= Node types in {product-title}
+
+{product-title} uses bootstrap, master, and worker nodes. The bootstrap node
```
smarterclayton: Haven’t settled on naming yet for worker nodes.  I would however say that we don’t “use ... nodes”, we assign nodes roles within the cluster that define their security boundaries and have specific roles.

It’s unclesr to me that this is the best way to communicate this though - bootstrap nodes are a detail of installation, while master and worker/compute nodes are what an admin deals with every day.  I wouldn’t give them the same level of importance.

This page would be better as a discussion of the roles nodes play (use role instead of type, role is the official term).  Connecting it to the ways people can change nodes (masters today cannot be changed, machines sets mean adding new nodes roles is easy), how security boundaries work, what tools are used to subdivide and group nodes is better.

modules/operators-overview.adoc
```diff
@@ -0,0 +1,96 @@
...
+_vertices_ and the _edges_ that connect them, of component Operators. The edges
+in the graph show which versions you can safely upgrade to. The cluster version
+Operator checks with Cincinnati and determines valid upgrades and upgrade paths
+based on current component versions and information in the graph. If you
+configure it to do so, Cincinnati sends the release artifacts that it needs to
+perform the upgrade to your image registry, and the cluster version Operator
+upgrades your cluster. By accepting automatic updates, you can automatically
+keep your cluster up to date with the most recent compatible components.
+
+To allow Cincinnati to provide only compatible updates, a release verification
```
smarterclayton: Don’t call it Cincinnati.  Say it’s the OpenShift update service.  Cincinnati is more of a code name.

modules/security-overview.adoc
```diff
@@ -0,0 +1,125 @@
+// Module included in the following assemblies:
+//
+// * architecture/architecture.adoc
+
+[id='security-overview-{context}']
+= Security in {product-title}
+
+The {product-title} and Kubernetes APIs authenticate users who present
```
smarterclayton: This is not a great set of content (it’s really unbalanced).  The auth teams need to provide feedback to correct this but it should be talking about our threat model, the different tools we have for security, and provide high level overview of our security subsystems.  This page as it is is not useful.

modules/operators-overview.adoc
```diff
@@ -0,0 +1,96 @@
...
+In version 4.0, separate processes and images and pods run to drive the kubernetes
+API and pod managers. The control plane services run as static pods so they can
+manage normal workloads or processes the same way that they manage disaster
+recovery. Aside from the core control plane components, other services run as 
+normal pods on the cluster, managed by regular Kubernetes constructs. Unlike in the past
+where the `kubelet` could be running as containerized or non-containerized, the `kubelet`
+always runs as a `systemd` process.
+
+
+[id='cluster-version-operator-{context}']
```
smarterclayton: I don’t want to focus on these lower level details until we have an architecture topic on “upgrades” (which covers most of the details).  I would do link outs to the major functional areas and list a few of the core operators and their roles, but this is too focused on “what” and not “why”.  The details below would be maybe a subtopic under specific operators but as is this is too messy.

modules/operators-overview.adoc
```diff
@@ -0,0 +1,96 @@
...
+== The cluster version Operator
+
+The cluster version Operator orchestrates all things.
+
+{product-title} 4.0 introduces several new components that support the cluster
+version Operator, including Cincinnati and Telemetry.
```
smarterclayton: Telemetry should be split out and part of an “observability” high level arch topic that then describes the major subsystems and what they do together (monitoring with Prometheus, alertmanager and grafana, logging, and telemetry).

smarterclayton (changes requested):
The group leads need to get engaged more and help flush out the high level topics.  This is a good start but I expect more input from them to describe the high level areas.  

Can you add a rough “high level topics” tree to this PR so we can comment and discuss what the top sections should be and ensure they match?

smarterclayton:
@derekwaynecarr @eparis I’d like to see us get the high level topics agreed on and help flesh out which topics are under covered.  We need to invest more time with Kathryn and the docs team than we have historically on these docs.

modules/operators-overview.adoc
kalexand-rh: Do you either have a suggestion for a contact who can explain how these fit together or the outline for the user story?  I'm not familiar enough with how monitoring, alerting, grafana, logging, and telemetry fit together to say what that story is, and all of our assemblies need to have one.

kalexand-rh (commented):
@smarterclayton, thank you! I'm working on updates based on your comments.

I'm not sure who the best contacts are for some of your requested changes - will you tag some more people in for me?

modules/node-types.adoc
vikram-redhat: @smarterclayton ^^

kalexand-rh:
@enxebre, @crawford, will you PTAL at this draft architecture overview? @smarterclayton has some great suggestions and requests that I don't have enough information to implement. (If you have other suggestions for stakeholders to review, please tag them in!)

modules/machine-api-overview.adoc
```diff
@@ -0,0 +1,56 @@
...
+link:https://github.com/kubernetes-sigs/cluster-api[Cluster API] project and
+custom {product-title} resources.
+
+The three primary resources are:
+
+`Machines`:: A fundamental unit that describes a node host. A `machine` has a
+class, which describes the types of compute nodes that are offered for different
+cloud platforms. For example, a `machine` type for a worker node on Amazon Web
+Services (AWS) might define a specific machine type and required metadata.
+`MachineClasses`:: A unit that defines a class of `machines` and facilitates
```
enxebre: MachineClassses are not yet integrated with the platform.
MachineDeployments might be landing soon https://github.com/openshift/installer/pull/990. These are mainly to machines what "deployments" to pods. I'll keep this thread  updated.

modules/machine-api-overview.adoc
```diff
@@ -0,0 +1,56 @@
...
+configuration reuse across `machines` of the same class. This unit functions
+like a `StorageClass` for PersistentVolumeClaims.
+`MachineSets`:: Groups of machines. `MachineSets` are to `machines` as
+`ReplicaSets` are to `Pods`. If you need more `machines` or need to scale them down,
+you change the *replicas* field on the `MachineSet` to meet your compute need.
+
+
+The following custom resources add more capabilities to your cluster:
+
+`MachineAutoscaler`:: This resource automatically scales `machines` in
```
enxebre: It's worth clarifying that a `MachineAutoscaler` object would only take any effect after a ClusterAutoscaler object exists. Both ClusterAutoscaler and MachineAutoscaler resources are made available by the ClusterAutoscalerOperator

modules/cloud_installations.adoc
```diff
@@ -19,8 +19,45 @@ cluster. These details include ...
...
+////
+There are individual commands to perform the different actions in cluster creation
+if you want to try to make customizations, but you can run openshift-install
+create cluster to get the default cluster done quick.
+
+$ openshift-install --help
+$ openshift-install create install-config
+$ openshift-install create manifests
+$ openshift-install create ignition-configs
+$ openshift-install create cluster
```
DanyC97: i'd suggest to always educate the users to run using `--dir=` option otherwise issues might appear in case multiple clusters are deployed from same host controller.

@wking fyi

modules/node-management.adoc
DanyC97: @smarterclayton imo i'd add and expand the steps 1 to 8 from https://github.com/derekwaynecarr/openshift-the-easy-way/blob/initial-structure/demo/demo.md . It does create a full picture of what is happening and that will help with self troubleshooting

modules/machine-api-overview.adoc
DanyC97: imo extending and adding more info like https://www.linux.com/blog/event/kubecon/2018/4/extending-kubernetes-cluster-api will clarify a lot of end users who are not familiar with K8 Cluster API

modules/cloud_installations.adoc
wking: Example working [here][1].

[1]: https://github.com/openshift/installer/blame/v0.8.0/README.md#L53-L55

modules/update-service-overview.adoc
```diff
@@ -0,0 +1,38 @@
...
+//
+// * architecture/architecture.adoc
+
+[id='update-service-overview-{context}']
+== The {product-title} update service
+
+The {product-title} update service is the hosted service that provides over the air updates to both 
+{product-title} and RHCOS. It provides a graph, or diagram that contain
+_vertices_ and the _edges_ that connect them, of component Operators. The edges
+in the graph show which versions you can safely upgrade to. The cluster version
```
wking: Before you talk about edges for upgrades, you should probably point out that the vertices are update payloads that specify the intended state of the managed cluster components.

modules/update-service-overview.adoc
```diff
@@ -0,0 +1,38 @@
...
+based on current component versions and information in the graph. If you
+configure it to do so, the {product-title} update service sends the release artifacts that it needs to
+perform the upgrade to your image registry, and the cluster version Operator
+upgrades your cluster. By accepting automatic updates, you can automatically
+keep your cluster up to date with the most recent compatible components.
+
+To allow the {product-title} update service to provide only compatible updates, a release verification
+pipeline exists to drive automation. Each release artifact is verified for
+compatibility with supported cloud platforms and system architectures as well
+as other component packages. After the pipeline confirms the suitability of a
```
wking: You cover testing update payloads in isolation (before we push a node to the update graph), but it would be good to have some words around testing the upgrades (before we push an edge).  CC @steveeJ who actually understands how graph updates work, because I could be getting this wrong ;).

modules/update-service-overview.adoc
```diff
@@ -0,0 +1,38 @@
...
+
+To allow the {product-title} update service to provide only compatible updates, a release verification
+pipeline exists to drive automation. Each release artifact is verified for
+compatibility with supported cloud platforms and system architectures as well
+as other component packages. After the pipeline confirms the suitability of a 
+release, the {product-title} update service can apply the update to your cluster or notify you that it
+is available.
+
+The interaction between the registry and the {product-title} update service is different during
+bootstrap and continuous update modes. When you bootstrap the initial
```
wking: Do we want to talk about bootstrapping here?  It never hits the update service.  Or are their future plans to pull the initial update payload from the update service?

modules/update-service-overview.adoc
```diff
@@ -0,0 +1,38 @@
...
+infrastructure, the cluster version Operator finds 
+the fully qualified image name for the shortname of the images that it needs to 
+apply to the server during installation. It looks at the image stream that it needs
+to apply and renders it to disk. It calls bootkube and waits for a temporary minimal control
+plane to come up and load the cluster version Operator.
+
+During continuous update mode, two controllers run. One continuously updates
+the payload manifests, applies them to the cluster, and outputs the status of
+the controlled rollout of the Operators, whether they are available, upgrading,
+or failed. The second controller constantly checks with to Cincinnati to
```
wking: This is your first mention of Cincinnati by name.  If we want to name it, we should probably name it early in this page.  If we want to leave it generic, we should probably replace Cincinnati with "the {product-title} update service".

And maybe "constantly checks" -> "polls"?  "constantly checks" sounds like a hot spin-lock, which isn't very civilized ;).

modules/machine-api-overview.adoc
```diff
@@ -0,0 +1,56 @@
...
+
+`Machines`:: A fundamental unit that describes a node host. A `machine` has a
+class, which describes the types of compute nodes that are offered for different
+cloud platforms. For example, a `machine` type for a worker node on Amazon Web
+Services (AWS) might define a specific machine type and required metadata.
+`MachineClasses`:: A unit that defines a class of `machines` and facilitates
+configuration reuse across `machines` of the same class. This unit functions
+like a `StorageClass` for PersistentVolumeClaims.
+`MachineSets`:: Groups of machines. `MachineSets` are to `machines` as
+`ReplicaSets` are to `Pods`. If you need more `machines` or need to scale them down,
```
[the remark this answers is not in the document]
wking: Actually, this is fine, since [`MachineDeployments` are a separate thing][1].

[1]: https://github.com/openshift/openshift-docs/pull/12880#discussion_r245325123

modules/update-service-overview.adoc
steveeJ: What's the average reader interested in, i.e. who's the average reader for this document? Might they be interested in [OpenShift 4 Channels][1]? The *suitability* property is determined by the [Release Controller][2].

[1]: https://docs.google.com/document/d/1AeuliIIu2VN-sF8ECpFMygrDoz4D2cZ8B44sZFBw4P8/edit?pli=1#heading=h.xx9vu52n421x

[2]: https://docs.google.com/document/d/1PAHSF86Un6CdG7STX-vIGbrgjVTnVpDOSBplos0U8Rk/edit#heading=h.5k1h50oq8lyi

modules/operators-overview.adoc
```diff
@@ -0,0 +1,65 @@
...
+OLM is a framework that manages Kubernetes-native applications as Operators.
+Instead of managing Kubernetes manifests, it manages Kubernetes Operators.
+OLM manages two classes of Operators, Red Hat Operators and certified Operators.
+
+Some Red Hat Operators drive the cluster functions, like the scheduler and
+problem detectors. Others are provider for you to manage yourself and use in
+your applications, like etcd. {product-title} also offers certified Operators,
+which the community built around the OLM and maintains. These certified Operators
+are traditional applications that are Kubernetes-aware because they are wrapped
+in an Operator.
```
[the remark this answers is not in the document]
kalexand-rh: @wking, I see where you're going, but I don't think that "wrap" would translate well. Does this sound ok to you?
> These certified Operators provide an API layer to traditional applications so you can manage the application through Kubernetes constructs.

modules/update-service-overview.adoc
kalexand-rh: I don't know about future plans. I'll comment this out.

kalexand-rh (commented):
@wking and @enxebre, thank you! I've made some of the easier updates in your comments and added a few more questions.

modules/operators-overview.adoc
```diff
@@ -0,0 +1,65 @@
...
+[id='operators-overview-{context}']
+= Operators in {product-title}
+
+{product-title} v4 uses different classes of Operators to perform cluster 
+operations and run services on the cluster for your applications to use.
+
+[id='platform-operators-{context}']
+== Platform Operators in {product-title}
+
+In {product-title} version 4.0, all cluster functions are divided into a series
```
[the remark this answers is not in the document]
wking: > ... the first is to differentiate the 3.x and 4.x streams, and the second is specifically discussing the 4.0 cluster topology, which I imagine will evolve through the releases.

Do you foresee the Operator approach changing over the life of 4.x?  If we switch to something else, I'd expect that would trigger a bump to OpenShift 5.  Also, down in the `v4` paragraph below you're talking about the `kubelet` being run by systemd instead of as a possibly containerized process, and that seems much more mutable to me than "we're going to use Operators".

modules/machine-api-overview.adoc
```diff
@@ -0,0 +1,56 @@
...
+system, {product-title} version 4.0 offers an elastic, dynamic provisioning
+method.
+
+The Machine API is a combination of primary resources that are based on the upstream
+link:https://github.com/kubernetes-sigs/cluster-api[Cluster API] project and
+custom {product-title} resources.
+
+The three primary resources are:
+
+`Machines`:: A fundamental unit that describes a node host. A `machine` has a
```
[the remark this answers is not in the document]
wking: > Looks like it should be just `Node`.

Then they're overloading :/.  Because from [here][1]:

> A node is a worker machine in Kubernetes, previously known as a **minion**.

Maybe they don't have a generic name for machine/hosts/whatever that is not overloaded?

[1]: https://kubernetes.io/docs/concepts/architecture/nodes/

modules/node-management.adoc
```diff
@@ -0,0 +1,46 @@
+// Module included in the following assemblies:
+//
+// * architecture/architecture.adoc
+
+[id='node-management-{context}']
+= Node management in fully-managed {product-title}
+
+The fully-managed version of {product-title} version 4.0 integrates management of
+the container operating system and cluster management. Because the cluster manages
+both its updates and updates to RHCOS, {product-title} provides an opinionated
```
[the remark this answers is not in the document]
wking: > will you give me a little more context about why the RHCOS updates should be considered a subset of the OCP updates?

There's some more background in openshift/machine-config-operator#183 and openshift/installer#987.  Things are still a bit up in the air, but, as I understand it, RHCOS updates will eventually be triggered by update-payload bumps gated by a Cincinnati graph like all other cluster updates.

modules/node-management.adoc
```diff
@@ -0,0 +1,46 @@
...
+{product-title} employs three DaemonSets and controllers to simplify node management:
+
+* The `machine-config-controller` coordinates machine upgrades.
+* The `machine-config-daemon` DaemonSet is a subset of the Ingition configuration that
+applies the specified machine configuration and controls kubelet configuration.
+* The `machine-config-server` DaemonSet provides the Ignition config to new hosts.
+
+These tools orchestrate operating system updates and configuration changes to
+the hosts by using standard Kuberentes-style constructs. A `machine-config-daemon`
+DaemonSet runs on each compute node in the cluster and watches for changes in
```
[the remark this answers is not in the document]
wking: > @wking, do you know who would know?

Talking this over with @chancez and @robszumski today left me feeling like "machine" is the generic term for a host (e.g. an [AWS instance][1]), a "master" is something (likely a machine) running control-plane components, usually a Kubernetes API server, and a "node" is something (likely a machine) where the cluster can schedule pods, usually via a kubelet.  So in this case, we're scheduling pods, so "node" would be correct.  And with our particular setup, all of the masters are also nodes as well.

We could use something like "compute node" to talk about nodes that are not masters, but we don't need to single those out in this sentence.

Of course, it's also possible that I still have this all wrong ;).

[1]: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Instance.html

modules/operators-overview.adoc
```diff
@@ -0,0 +1,65 @@
...
+managed component's deployment
+
+[id='OLM-operators-{context}']
+== Operators managed by OLM
+
+In addition to the Operators that comprise {product-title}, the Cluster
+Operator Lifecycle Management (OLM) component manages more Operators.
+OLM is a framework that manages Kubernetes-native applications as Operators.
+Instead of managing Kubernetes manifests, it manages Kubernetes Operators.
+OLM manages two classes of Operators, Red Hat Operators and certified Operators.
```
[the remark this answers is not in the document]
adellape: In the latest "Operator Management" UI, they are identified as "Red Hat Operators" and "Certified Operators", each with a subtitle of "Packaged by Red Hat". cc @ecordell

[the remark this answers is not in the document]
adellape: Though I've also heard discussion about "Community Operators", which it looks like "Certified Operators" are maybe incorrectly described as a few lines down. Certified Operators would be ISV / RH partner-provided and maintained, I believe, whereas Community Operators would be more the "community built and maintained" ones. cc @hasbro17

modules/operators-overview.adoc
```diff
@@ -0,0 +1,65 @@
...
+recovery. Aside from the core control plane components, other services run as 
+normal pods on the cluster, managed by regular Kubernetes constructs. Unlike in the past
+where the `kubelet` could be running as containerized or non-containerized, the `kubelet`
+always runs as a `systemd` process.
+
+
+[id='second-level-operators-{context}']
+== Second-level Operators in {product-title}
+
+The cluster version Operator, when we talk about payload manifests, is a
```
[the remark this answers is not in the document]
adellape: Came here to suggest "Cluster Version Operator" as well, and an added "(CVO)", as I feel like I see CVO used elsewhere pretty liberally.

modules/operators-overview.adoc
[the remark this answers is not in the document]
wking: > Does this sound ok to you...

Circling in ;), how about:

> These certified Operators provide an API to manage traditional applications through Kubernetes constructs.

modules/node-management.adoc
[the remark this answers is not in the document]
kalexand-rh: OK. For now, I'm going to change it to "Because the cluster manages
its updates, including updates to Red Hat CoreOS on cluster nodes," because you're right, they're treated the same. I think it's helpful to emphasize that the RHCOS updates are managed here, too. It's helpful context.

modules/operators-overview.adoc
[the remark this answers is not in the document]
kalexand-rh: Yup, and we updated the [guidance](https://github.com/openshift/openshift-docs/commit/9263aded5a5936d7a87386d3d16259aaa902f7e1) in our repo.

modules/node-management.adoc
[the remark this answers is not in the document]
kalexand-rh: @wking, that sounds reasonable, though, so I'll make some updates and see what other feedback I get.

modules/operators-overview.adoc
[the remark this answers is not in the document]
kalexand-rh: @adellape, thanks! I'm making some updates to this draft base on your comments.

modules/update-service-overview.adoc
kalexand-rh: I made a separate issue to document a user story about understanding and configuring updates, and I've added a note to update this section. Thanks!

modules/cloud_installations.adoc
kalexand-rh: I will open another PR to update the install docs, and I plan to include the `--dir=` option there.

modules/installation-options.adoc
kalexand-rh: I could still use a external link to the roadmap, but I've added clarification that more support is coming.

kalexand-rh:
I've created some follow-up issues and am merging per @vikram-redhat.

PR #12905, opened 2018-11-19
ocs 310 additions

cooktheryan:
Addition of links for ocs deployment for the openshift providers that would benefit from having OCS deployed.

@netzzer

### DIFF
--- a/install_config/configuring_aws.adoc
+++ b/install_config/configuring_aws.adoc
@@ -313,3 +313,5 @@ Balancing (ELB) will not be able to create a load balancer.
 ====
 
 After tagging any resources, restart the master services on the master and the node service on all nodes. See the xref:aws-applying-configuration-changes[Applying Configuration Section].
+
+include::install_config/topics/red_hat_openshift_container_storage.adoc[leveloffset=+2]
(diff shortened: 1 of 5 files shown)


kalexand-rh:
@cooktheryan, on a quick scan, this looks a lot like https://github.com/openshift/openshift-docs/pull/12906 . You could do one pr against master, and we could cherrypick it to both 3.10 and 3.11. Will you PTAL at my comments on that PR?

kalexand-rh:
Gotcha. Our if statement usage doesn't cover that use case, and this is small enough usage that I don't think adding a variable makes sense. (Even if we used a variable, we'd still need different PRs to set it up.)

cooktheryan:
@kalexand-rh I just realized I'm not sure my current plan of merging this to master may not work unless we cherry-pick immediately after as both PRs will be modifying the same file on master

kalexand-rh:
@cooktheryan, close this one. I'll CP the other to 3.11 and then manually CP it to 3.10. I'll fix the link before I merge the 3.10 one.

cooktheryan:
@kalexand-rh++

PR #12914, opened 2018-11-20
Add openshift3/ose into OCP infra component images

pichuang:
Signed-off-by: Phil Huang <phil.huang@redhat.com>

### DIFF
--- a/install/disconnected_install.adoc
+++ b/install/disconnected_install.adoc
@@ -199,6 +199,7 @@ $ docker pull registry.redhat.io/openshift3/mariadb-apb:<tag>
 $ docker pull registry.redhat.io/openshift3/mediawiki:<tag>
 $ docker pull registry.redhat.io/openshift3/mediawiki-apb:<tag>
 $ docker pull registry.redhat.io/openshift3/mysql-apb:<tag>
+$ docker pull registry.redhat.io/openshift3/ose:<tag>
 $ docker pull registry.redhat.io/openshift3/ose-ansible:<tag>
 $ docker pull registry.redhat.io/openshift3/ose-ansible-service-broker:<tag>
 $ docker pull registry.redhat.io/openshift3/ose-cli:<tag>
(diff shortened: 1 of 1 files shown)


kalexand-rh:
@jianlinliu, will you PTAL?

jianlinliu:
openshift3/ose is some old name, in 3.11, its name is replaced by "openshift3/ose-control-plane", why we add "openshift3/ose" back?

pichuang:
@jianlinliu  The reason is the installation always pull the image `openshift3/ose`, not `openshift3/ose-control-plane` at OCP 3.11.16

jianlinliu:
@pichuang Really? I am pretty sure we should not use "openshift3/ose" image any more in 3.11, and I also checked openshift-ansible code, did not find any place to refer to it. Could you have any log to prove installer pull the image? Even it is, then it should be some functional bug, but not doc bug.

pichuang:
@jianlinliu 
Yes. I try the installation so many times, and always stop there when it had executed `TASK [container_runtime : Create credentials for oreg_url]`. 

The root cause is the credentials will get the image "openshift3/ose", and the tag name is "latest", not "v3.11.16"

I also checked the source code, and didn't find that too.

Logs:
```
fatal: [master1.ocp.local]: FAILED! => {"attempts": 3, "changed": false, "msg": "time=\"2018-11-20T13:53:45+08:00\" level=fatal msg=\"Error reading manifest latest in quay.pichuang.local/open
shift3/ose: manifest unknown: manifest unknown\" \n", "state": "unknown"}
```

Verify:
```
skopeo inspect docker://internal.registry.local/openshift3/ose
skopeo inspect docker://internal.registry.local/openshift3/ose:v3.11.16
```

jianlinliu:
@pichuang, Good catch. The root cause is at the following lines in roles/openshift_facts/defaults/main.yml.
l_docker_creds_image_dict:
  openshift-enterprise: 'openshift3/ose'
  origin: 'openshift/origin'
l_docker_creds_test_image: "{{ l_docker_creds_image_dict[openshift_deployment_type] }}"

The default value - `'openshift3/ose'` is not set properly, should be some other value, such as: "openshift3/ose-pod:{{openshift_image_tag}}" or "openshift3/ose-control-plane:{{openshift_image_tag}}"

When you are running the install against the internal registry which need authentication, then would hit your issue. Another workaround is set `oreg_test_login=false` to skip the cred check.

Whatever this is a installer bug, but not some doc bug.

jianlinliu:
@sdodson My analysis is right? If yes, could you commit a fix for it.

vikram-redhat:
@pichuang - based on @jianlinliu response I am closing this as this is not a docs bug. Please feel free to reopen if things change.

PR #12969, opened 2018-11-27
Added graph driver content to 4.0

ahardin-rh:
https://jira.coreos.com/browse/OSDOCS-44

Preview Build: http://file.rdu.redhat.com/~ahardin/11272018/graph-drivers-4-0/storage/choosing-a-graph-driver.html

### DIFF
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -64,6 +64,13 @@ Topics:
 - Name: Scaling the cluster monitoring Operator
   File: scaling-cluster-monitoring-operator
 ---
+Name: Storage
+Dir: storage
+Distros: openshift-*
+Topics:
+- Name: Choosing a graph driver
+  File: choosing-a-graph-driver
+---
 Name: Operators
 Dir: operators
 Distros: openshift-*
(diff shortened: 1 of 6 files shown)


ahardin-rh:
@jeremyeder @rhatdan @rhvgoyal Can you please review this content and let me know if it is still accurate for 4.0? Thank you!

ahardin-rh:
@sjug I think you are now the assigned review for this section. I broke the graph driver content out into its own section.

ahardin-rh:
Closing this PR after discussion with @jeremyeder and @sjug 
We're moving away from graph drivers. With CRI-O, there will be no choices in graph driver, it will only be overlay(2) and there will be no storage configuration.

rhatdan:
I am not sure I would make that statement,  We might want to support multiple graphdrivers in the future for using Kata Containers since they work much better on devicemapper/lvm images then on overlay.  And for rootless containers we are using vfs storage.

jeremyeder:
@rhatdan The content as-is addresses concerns that are 3+ years old, and largely irrelevant since defaulting to overlay.  Do you still prefer we keep it, and extend it to include your points above?  Thinking about it from a customer PoV, it strikes me that the existing content might be considered a distraction in the 4.0+ world (these docs would remain present in 3.11 and earlier branches).

rhatdan:
No I don't mind removing it, I just want people to understand that the other graphdrivers are still important.

PR #12971, opened 2018-11-28
Sizing guideline calculus is incorrect

akram:
Sizing guideline calculus is incorrect: We compute the rate in KB per min but then we make the computation for a whole day in seconds.
So the old sizing was 60 times bigger than the required one.

### DIFF
--- a/install_config/aggregate_logging_sizing.adoc
+++ b/install_config/aggregate_logging_sizing.adoc
@@ -286,7 +286,7 @@ Assumptions:
 . Lines per second load on application: 1
 . Raw text data -> JSON
 
-Baseline (256 characters per minute -> 15KB/min)
+Baseline (256 characters per second -> 15KB/min)
 
 [cols="3,4",options="header"]
 |===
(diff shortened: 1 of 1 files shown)


ahardin-rh:
@jeremyeder @jmencak PTAL

jmencak:
Seems to makes sense.  "per minute -> 15KB" should also be adjusted to something like "per minute ~15KB" and count with that.  A bit messy and imprecise to my taste overall and should probably be rewritten.  @mffiedler should probably look at this for sanity since he took over after the original author.

ahardin-rh:
@mffiedler PTAL

mffiedler:
I think the author's original intent was not 256 bytes/min but 256 bytes/sec which is more realistic for logging.   That would make the 15KB/min correct and the original numbers more realistic.     I suggest reverting the calculations and having the change for this PR be:

Baseline (256 characters per second → 15KB/min)

akram:
@mffiedler 
256 bytes/s was already the original assumption, which makes 256x60~=15KB/min
15KB/min is already correct.
What it not correct is 15KB/min x number of seconds of a day 
It should be: 15KB/min x number of minutes of a day

akram:
@mffiedler I just amended my commit and fixed the Baseline unit also

mffiedler:
@akram Thanks!

PR #12979, opened 2018-11-28
Correct command for stopping node service. Fixes #12947

### LINKED ISSUE #12947: migration procedure from redhat/openshift-ovs-subnet to redhat/openshift-ovs-networkpolicy 3.10/3.11
The migration procedure from redhat/openshift-ovs-subnet to redhat/openshift-ovs-networkpolicy is not working as described in the documentation.

https://docs.okd.io/3.11/install_config/configuring_sdn.html#migrating-between-sdn-plugins

### What needs fixing?

Step 2 is asking for restarting master and nodes. This makes sense so far. But step 3 does stop the etcd cluster, kubernetes master controllers and api. So step 4 has no working Master API and does fail.

I don't see any reason for step 3, as the master controller and api already has been restarted and etcd is not affected by changing the network plugin.

Additional steps are required: To change the network policy in the node group config map. Otherwise the sync pod does reset the configuration to the old network plugin. This causes a cluster outage afterwards.

Manually for each node config configmap:
`oc get configmap node-config-compute -n openshift-node`

```
data:
  node-config.yaml: |
    networkConfig:
      networkPluginName: redhat/openshift-ovs-networkpolicy
```

Or execute:

```
# ansible-playbook -i </path/to/inventory/file> \
    ~/openshift-ansible/playbooks/openshift-master/openshift_node_group.yml
```

afcollins:
@openshift/team-documentation

Needed in enterprise-3.10 and enterprise-3.11.

### DIFF
--- a/install_config/configuring_sdn.adoc
+++ b/install_config/configuring_sdn.adoc
@@ -305,11 +305,10 @@ xref:configuring-the-pod-network-on-nodes[nodes] in their configuration files.
 # master-restart controllers
 ----
 
-. Stop the node service on all masters and nodes by removing the pod definition:
+. Stop the node service on all masters and nodes:
 +
 ----
-# mkdir -p /etc/origin/node/pods-stopped
-# mv /etc/origin/node/pods/* /etc/origin/node/pods-stopped/
+# systemctl stop atomic-openshift-node.service
 ----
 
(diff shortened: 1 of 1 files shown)


kalexand-rh:
@bmeng, will you PTAL?

bmeng:
LGTM, thanks.

kalexand-rh:
Thanks! Merging.

PR #12983, opened 2018-11-29
Add -k option to curl to https

nak3:
curl to https without cert usually fail, especially testing phase. This patch replace `-v` with `-k`.

cc @miminar @bmcelvee


bmcelvee:
LGTM

Thanks, @nak3!

nak3:
@bmcelvee @miminar Thank you. Sure, updated.

miminar:
Thanks @nak3. I'd just make it more generic or annotate it as an example for the default installation method. The commands apply only if you use the advanced installer without overriding the default options in the inventory:

```
openshift_hosted_registry_routecertificates= "{'certfile': '<path>/org-cert.pem', 'keyfile': '<path>/org-privkey.pem', 'cafile': '<path>/org-chain.pem'}"
```

PR #12899, opened 2018-11-16
Add build inputs modules

bmcelvee:
Resuming work from: https://github.com/openshift/openshift-docs/pull/11804

PR #12989, opened 2018-11-29
PR12899-follow-up edits to mod content

bmcelvee:
Follow-up to https://github.com/openshift/openshift-docs/pull/12899

### DIFF
--- a/builds/creating-build-inputs.adoc
+++ b/builds/creating-build-inputs.adoc
@@ -2,13 +2,9 @@
 //* assembly/builds
 
 :context: creating-build-inputs
-[id='creating-build-inputs_{context}']
 = Creating build inputs
-{product-author}
-{product-version}
-:data-uri:
-:icons:
-:experimental:
+include::modules/common-attributes.adoc[]
+
 :toc: macro
 :toc-title:
 :prewrap!:
--- a/modules/builds-source-secret-ssh-key-auth.adoc
+++ b/modules/builds-source-secret-ssh-key-auth.adoc
@@ -6,43 +6,36 @@
 // This module can be included from assemblies using the following include statement:
 // include::<path>/builds-source-secret-ssh-key-auth.adoc[leveloffset=+1]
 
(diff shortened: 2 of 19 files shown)


bmcelvee:
We define a “working directory” but don’t tell the user where the “working directory” is--Is it in a POD/Host?

Need to address.

bmcelvee:
> We define a “working directory” but don’t tell the user where the “working directory” is--Is it in a POD/Host?
> 
> Need to address.

Defined in "Adding input secrets and ConfigMaps"?

builds/creating-build-inputs.adoc
```diff
@@ -3,6 +3,7 @@
 
 :context: creating-build-inputs
 [id='creating-build-inputs_{context}']
+include::modules/common-attributes.adoc[]
```
bmcelvee: This doesn't quite work here? Where should it be located in the assembly module?

bmcelvee: ^ fixed

bmcelvee:
@openshift/team-documentation PTAL :snowflake: 

(You can ignore my comments, they're content related and just reminders for myself.)

modules/builds-source-secret-ssh-key-auth.adoc
```diff
@@ -6,35 +6,30 @@
...
 default.
 
-. Prerequisites
+.Prerequisites
 
 * SSH key based authentication requires a private SSH key.
 
 .Procedure
 
+Two files are created: the public key and a corresponding private key (one of
```
kalexand-rh: I think there's another step hidden in here.

kalexand-rh (commented):
Just one comment. If it's building clean, it LGTM.

modules/builds-source-secret-ssh-key-auth.adoc
bmcelvee: Thanks! I'll update in just a bit :smiley_cat:

PR #12990, opened 2018-11-29
Add IAM role for S3 registry config

david-igou:
Currently on the docs the IAM role is a blank codeblock: https://docs.openshift.com/container-platform/3.11/install_config/configuring_aws.html

Here is the correct role, per: https://docs.docker.com/registry/storage-drivers/s3/#s3-permission-scopes


David-Igou:
I accidently attempted to merge my master

PR #12991, opened 2018-11-29
Aws s3 iam role

david-igou:
Currently on the docs the IAM role is a blank codeblock: https://docs.openshift.com/container-platform/3.11/install_config/configuring_aws.html

Here is the correct role, per: https://docs.docker.com/registry/storage-drivers/s3/#s3-permission-scopes

### DIFF
--- a/install_config/topics/configuring_for_aws_registry-proc.adoc
+++ b/install_config/topics/configuring_for_aws_registry-proc.adoc
@@ -29,7 +29,31 @@ in the region of `us-east-1`.
 The default policy
 [source,yaml]
 ----
-
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": [
+        "s3:ListBucket",
+        "s3:GetBucketLocation",
+        "s3:ListBucketMultipartUploads"
+      ],
+      "Resource": "arn:aws:s3:::S3_BUCKET_NAME"
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "s3:PutObject",
+        "s3:GetObject",
+        "s3:DeleteObject",
+        "s3:ListMultipartUploadParts",
+        "s3:AbortMultipartUpload"
+      ],
+      "Resource": "arn:aws:s3:::S3_BUCKET_NAME/*"
+    }
+  ]
+}
 ----
 
 
@@ -57,8 +81,8 @@ openshift_hosted_registry_acceptschema2=true
 openshift_hosted_registry_enforcequota=true
 openshift_hosted_registry_replicas=3
 ----
-<1> The access key for the IAM user.
-<2> The secret key for the IAM user.
+<1> The access key for the IAM user. (Not required with IAM Roles in place)
+<2> The secret key for the IAM user. (Not required with IAM Roles in place)
 <3> The S3 storage bucket name.
 <4> The region in which the bucket exists.
 


David-Igou:
@openshift/team-documentation

I do not have permission to apply labels (Adding this per the instructions of /create_or_edit_content.adoc)

kalexand-rh:

kalexand-rh:
@xltian, will you please suggest a reviewer?

wzheng1:
@bparees Could you please review this doc? For QE's account role, it is ready after we applied it, so we are not sure what kind of role it is required.

install_config/topics/configuring_for_aws_registry-proc.adoc
```diff
@@ -57,8 +81,8 @@ openshift_hosted_registry_acceptschema2=true
 openshift_hosted_registry_enforcequota=true
 openshift_hosted_registry_replicas=3
 ----
-<1> The access key for the IAM user.
-<2> The secret key for the IAM user.
+<1> The access key for the IAM user. (Not required with IAM Roles in place)
+<2> The secret key for the IAM user. (Not required with IAM Roles in place)
```
bparees: I don't understand the not required part here.  If the accesskey + secret aren't supplied here, is the ansible installer supplying the values to the registry in some other way?  what values are used?

/cc @sdodson

David-Igou: No access keys are required if IAM roles are configured and attached to the EC2 instances running the registry. IAM roles allow access to AWS resources without storing access keys on the systems.

bparees: The registry runs inside a pod.  What may or may not be on the ec2 systems is irrelevant.  Are you sure about this?

David-Igou:
```
oc get secret registry-config \
    -o jsonpath='{.data.config\.yml}' -n default | base64 -d
```

Will output the registry config, in the example in the docs:

```
storage:
  delete:
    enabled: true
  cache:
    blobdescriptor: inmemory
  s3:
    accesskey: AKIAJ6VLREDHATSPBUA 
    secretkey: g/8PmTYDQVGssFWWFvfawHpDbZyGkjGNZhbWQpjH 
    region: us-east-1 
    bucket: openshift-registry-storage 
    encrypt: False
    secure: true
    v4auth: true
    rootdirectory: /registry 
    chunksize: "26214400"
```
On my cluster I installed the other day, I did not include these variables, and it reads:

```storage:
  delete:
    enabled: true
  cache:
    blobdescriptor: inmemory
  s3:
    region: us-east-1
    bucket: redacted-bucket-name
    encrypt: false
    secure: true
    v4auth: true
    rootdirectory: /registry
    chunksize: "26214400"
```
per the s3 storage driver section: https://docs.docker.com/registry/storage-drivers/s3/

bparees: Did that registry with no accesskey/secretkey actually work for accessing the bucket?  

Again, the registry is running inside a pod, it does not have access to anything on the ec2 machine that would be providing credentials.  The docker docs you linked presumably assume you're running the registry process directly on the ec2 machine where it can lookup the credentials.

/cc @dmage @legionus perhaps you know more about how this works (or if it works)?

David-Igou: Yes, it successfully accesses the bucket. 

When I `oc rsh` into the bucket and run:
`sh-4.2$ curl http://169.254.169.254/latest/meta-data/iam/security-credentials/`

it returns the name of the IAM role

`curl http://169.254.169.254/latest/meta-data/iam/security-credentials/EC2_S3_ROLE_NAME`

```
{
  "Code" : "Success",
  "LastUpdated" : "2018-12-05redacted",
  "Type" : "AWS-HMAC",
  "AccessKeyId" : "ASDFDSAFREDACTED",
  "SecretAccessKey" : "redacted",
  "Token" : "redacted token",
  "Expiration" : "2018-12-05redacted"
}
```

bparees: ok, I don't really know anything about how this works, so does this mean every pod on the node can access this key/token? (that would be bad)  What makes the registry pod special?

David-Igou: The answer to that is yes, actually. I can't really speak for the validation of this process, I'm just adding a role missing from the docs.

I think the idea is that only Infrastructure nodes should have this policy attached, and will only be running Redhat images. No compute nodes should have this role.

bparees: ok.  seems like a scary configuration choice but i guess the doc is not technically wrong.  thanks

bparees:
lgtm

sdodson:
/lgtm

Not too familiar with AWS policy definition but this aligns with my understanding that if the host has IAM roles sufficient it can operate without actual credentials.

kalexand-rh:
@David-Igou, please squash your commits.

@bparees, @sdodson, since you're both ok with this, I'll merge. Do you object to it going in both 3.10 and 3.11?

David-Igou:
@kalexand-rh done

bparees:
> @bparees, @sdodson, since you're both ok with this, I'll merge. Do you object to it going in both 3.10 and 3.11?

no objection.

kalexand-rh:
Thank you all! Merging.

PR #13000, opened 2018-11-30
osdocs-82 Add build strategy modules

bmcelvee:
https://jira.coreos.com/browse/OSDOCS-82


bmcelvee:
@vikram-redhat - yep! added

bmcelvee:
Next: adding build strategy options modules

PR #13003, opened 2018-12-03
Correct protocol from tcp to udp for a SDN FW rules.

bysnupy:
Fix this BZ: [[DOCS] vxlan port iptables change should be udp not tcp](https://bugzilla.redhat.com/show_bug.cgi?id=1644741)

* Version: v3.11

### DIFF
--- a/install_config/configuring_sdn.adoc
+++ b/install_config/configuring_sdn.adoc
@@ -194,15 +194,16 @@ networkConfig:
   ingressIPNetworkCIDR: 172.29.0.0/16
   networkPluginName: redhat/openshift-ovs-multitenant
   serviceNetworkCIDR: 172.30.0.0/16
-  vxlanPort: 4789 <1>
+  vxlanPort: 4889 <1>
 ----
 <1> Set to the value used by the nodes for the VXLAN Port. It can be an integer between 1-65535. The default value is `4789`.
 
 . Add the new port to the iptables rule on each cluster node:
 +
 ----
-# iptables -A OS_FIREWALL_ALLOW -p tcp -m state --state NEW -m tcp --dport 4889 -j ACCEPT
+# iptables -A OS_FIREWALL_ALLOW -p udp -m state --state NEW -m udp --dport 4889 -j ACCEPT <1>
 ----
+<1> `4889` is the `vxlanPort` value that you set in the master configuration file.
 
 . Restart the master services:
 +


kalexand-rh:
This bug was filed against 3.11. Is it valid in earlier versions?

bysnupy:
@kalexand-rh  No, it's just only valid in `v3.11`, `vxlanPort` can be changed as of `v3.11`.

kalexand-rh:
Thank you @bysnupy!

bmeng:
The iptables change LGTM.

But it is a little confusing that we use the modified port 4889 in step 3 directly.
Will be better mention it in step 2 that we take the port 4889 as an example.

install_config/configuring_sdn.adoc
```diff
@@ -201,7 +201,7 @@ networkConfig:
 . Add the new port to the iptables rule on each cluster node:
 +
 ----
-# iptables -A OS_FIREWALL_ALLOW -p tcp -m state --state NEW -m tcp --dport 4889 -j ACCEPT
+# iptables -A OS_FIREWALL_ALLOW -p udp -m state --state NEW -m udp --dport 4889 -j ACCEPT
```
kalexand-rh: @bysnupy, based on @bmeng's comment, I think this code block might need to look like this. Will you update your PR? (If you don't have time, I will try to open a new PR later this week.)

----
 # iptables -A OS_FIREWALL_ALLOW -p udp -m state --state NEW -m udp --dport 4789 -j ACCEPT <1>
----
<1> `4789` is the `vxlanPort` value that you set in the master configuration file.

bysnupy: @kalexand-rh, thank you for your review, @bmeng's point is appropriated here. We should use the consistent `port`. But this section is `Changing the VXLAN PORT`, and the `vxlan` default `port` is `4789`, so I think we need to change the port in other section as `4889`.

kalexand-rh: Thank you @bysnupy. You're right. :)  I think the format of the codeblock still needs to change to explain the value, though.

bysnupy: @kalexand-rh , If you have a suggestion, could you tell me about it please ?

kalexand-rh: I'm confused. I thought the value that you specified in the master config file needed to match the value that you added to your IPtables rule. If that's true, then I think that:
1. At line 197, s/vxlanPort: 4789/vxlanPort: 4889
2. Replace this command:
---
 # iptables -A OS_FIREWALL_ALLOW -p udp -m state --state NEW -m udp --dport 4889 -j ACCEPT <1>
----
<1> `4889` is the `vxlanPort` value that you set in the master configuration file.

If that's not the case, then can you please tell me where you get the `--dport` value in this command?

bmeng: Yeah, the value `vxlanPort` in the config file should match the `--dport` in the iptables command.

bysnupy: Thank you for review and suggestion @kalexand-rh @bmeng , I've updated along your suggestion.

kalexand-rh:
Thank you @bmeng  and @bysnupy! Please squash your commits so I can merge.

bysnupy:
@kalexand-rh I've squashed the commits.

kalexand-rh:
Thanks @bysnupy. :) Merging

PR #13004, opened 2018-12-03
Fixed missing , on labelPreference example JSON

palonsoro:
A single `,` was missing on labelPreference, which lead to an incorrect JSON.


kalexand-rh:
There are several instances of this text in this file, so if this change is validated by QE, all the code examples need to be updated.

@wjiangjay, will you PTAL?

kalexand-rh:
Thank you @wjiangjay and @palonsoro! Merging.

PR #13016, opened 2018-12-03
adding the auth folder

kalexand-rh:
attempting to save myself some merge conflicts later - I probably should have put all the auth conversion content in the same PR.


kalexand-rh:
Nope, this won't work. Closing.

PR #13017, opened 2018-12-04
Edits in authentication methods

ariordan-redhat:
Specify that using the default registry requires authentication token.
Not in Bugzilla; Ben Breard  requested the change via email:

> I just got pinged on some customer issues with the new registry, and I realize we have a gap on the docs. [https://docs.openshift.com/container-platform/3.11/install_config/configuring_red_hat_registry.html#auth-enabled-registry](https://docs.openshift.com/container-platform/3.11/install_config/configuring_red_hat_registry.html#auth-enabled-registry)
...
> The original text is technically accurate, however, we need openshift installs to use the tokens for a variety of reasons. Naming they're much, much more reliable and will lead to less problems longer term. I can go into details if that's helpful, but I need the docs to tell users to create a token. 

@wzheng1


ariordan-redhat:

kalexand-rh (commented):
Couple of text changes, otherwise LGTM

wzheng1:
@bparees Could you help to review this doc?

bparees:
> @bparees Could you help to review this doc?

aside from @kalexand-rh's comments, it looks fine to me.

PR #13024, opened 2018-12-04
Changed definition of  value for ScalingLimited

mburke5678:


mburke5678:
@sjenning PTAL
This is the change to ScalingLimited that I mentioned in scrum. I believe we had the definition of `false` incorrect for ScalingLimited. I changed value to `true`

ravisantoshgudimetla (approved):
LGTM @mburke5678. Thank you for the PR.

PR #13047, opened 2018-12-05
Add HPE 3PAR Storage object definition

c-snell:


kalexand-rh:
@aheslin, will you PTAL?

Issue #13046, opened 2018-12-05
System-Requirements on Hypervisors missing - do not use overcommit at  hypervisor level 


DanielFroehlich:
### Which section(s) is the issue in?
System-Requirements

### What needs fixing?
There is a sub-section missing on "Hypervisors". There is a requirement missing that the cpu resources should not be overcommitted on hypervisor level.
here is a draft:

OpenShift nodes can run on virtualized on supported hypervisors. Care must be taken to configure the underlying hypervisor / virtualization infrastructure correctly. It is **not recommend to run OpenShift nodes on overcommitted hypervisors**. You should configure the virtualization infrastructure in a way that for all OpenShift virtual machines, the workload on one virtual machine does not affect the processing resources available to the other virtual machines. Guaranteed ressources, i.e. no overcommitment on CPU and RAM ressources is strongly recommended. 

Rational: You would end up with two schedulers (the hypervisor and kubernetes) to fight against each other. This could lead to unstable and erratic behavior of the system.
 
Example: Kubernetes could schedule a new pod onto a node where it _thinks_ there is enough cpu capacity, but _actually_ the node is slow because it does not get enough cpu resources from the hypervisor (due to cpu overcommitment). By the time the hypervisor detects this and assigns more cpu resources, kubernetes could re-schedule the pod to a different node. 

Please see with the documentation of the utilized hypervisor on how to configure accordingly.

mburke5678:
Fixed via https://github.com/openshift/openshift-docs/pull/13050 and https://github.com/openshift/openshift-docs/pull/13427

PR #13050, opened 2018-12-05
Add new sys req for hypervisors: do not overcommit

DanielFroehlich:
Proposed fix for #13046

### DIFF
--- a/install/prerequisites.adoc
+++ b/install/prerequisites.adoc
@@ -268,6 +268,33 @@ OpenShift in a container], add `-e GOMAXPROCS=1` to the `docker run`
 command when launching the server.
 endif::[]
 
+[[configuring-hypervisor-no-overcommit]]
+=== Configuring underlying Hypervisor to guaranteed resources / no overcommit
+
+{product-title} nodes can run virtualized on supported hypervisors. Care must
+be taken to configure the underlying hypervisor / virtualization infrastructure
+correctly. It is **not recommend to run {product-title} nodes on overcommitted
+hypervisors**. You should configure the virtualization infrastructure in a way
+that for all {product-title} virtual machines, the workload on one virtual machine
+does not affect the processing resources available to the other virtual
+machines. Guaranteed resources, i.e. no overcommitment on CPU and RAM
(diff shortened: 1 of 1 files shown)


kalexand-rh:
@sdodson, thoughts?

kalexand-rh:
@sdodson, will you PTAL?

sdodson:
/cc @openshift/sig-pod 

I mean, it's all correct, but I wonder if we really need to state this? Pod team is probably best team to provide any additional feedback.

sdodson (approved):

sjenning:
Yeah, I would think this was understood.

Anyone overcommitting their hypervisors in terms of memory is asking for it.  Overcommitting on CPU would have roughly the same effect as... running on a slower processor (increased scheduling latency and whatnot).

Moreover most people creating virtual machines don't have control off the overcommit policies on the hypervisor.  I'm not sure we need to document this.

DanielFroehlich:
I agree that this is kind of obvious if you think about it. But I had the
discussion with a customer, needed to explain the topic several times and
was then asked "ok, but where is this stated in the documentation?"  Hence
this pull request.

---
Sent from mobile, please excuse typos. Daniel.Froehlich@redhat.com, +49 176
10 24 84 33

Am Do., 24. Jan. 2019, 20:53 hat Seth Jennings <notifications@github.com>
geschrieben:

kalexand-rh:
Scott and I had a chat about this one, and we think that it makes more sense to change https://docs.openshift.com/container-platform/3.11/scaling_performance/optimizing_compute_resources.html to say:
> You can use overcommit procedures so that resources such as CPU and memory are more accessible to the parts of your cluster that need them.
>**IMPORTANT**
>To avoid erratic cluster behavior due to scheduling collisions between the hypervisor and Kubernetes, do not overcommit at the hypervisor level.

kalexand-rh:
Great! I made a new PR with that change: https://github.com/openshift/openshift-docs/pull/13427 Since everyone's in agreement, I'll close this PR and merge it and cherry-pick the new one back to 3.6.

PR #13064, opened 2018-12-06
Added Optimizing compute resources content to 4.0

ahardin-rh:
Sourced from https://docs.openshift.com/container-platform/3.11/scaling_performance/optimizing_compute_resources.html

Preview build; http://file.rdu.redhat.com/~ahardin/12062018/optimizing-compute-resources/scalability_and_performance/optimizing-compute-resources.html

### DIFF
--- a/modules/using-a-pre-deployed-image-to-improve-efficiency.adoc
+++ b/modules/using-a-pre-deployed-image-to-improve-efficiency.adoc
@@ -1,0 +1,38 @@
+// Module included in the following assemblies:
+//
+// scalability_and_performance/optimizing-compute-resources.adoc
+
+[id='using-a-pre-deployed-image-to-improve-efficiency-{context}']
+= Using a pre-deployed image to improve efficiency
+
+You can create a base {product-title} image with a number of tasks built-in to
+improve efficiency, maintain configuration consistency on all node hosts, and
+reduce repetitive tasks. This is known as a pre-deployed image. For example,
+because every node requires the `ose-pod` image in order to run pods, each node
+has to periodically connect to the container image registry in order to pull the
+latest image. This can become problematic when you have 100 nodes attempting
+this at the same time, and can lead to resource contention on the image
+registry, waste of network bandwidth, and increased pod launch times.
+
+.Procedure
+
+To build a pre-deployed image:
+
+. Create an instance of the type and size required.
+
+. Ensure a dedicated storage device is available for CRI-O or local image
+or container storage, separate from any persistent volumes for containers.
+
+. Fully update the system, and ensure CRI-O or Docker is installed.
+
+. Ensure the host has access to all yum repositories.
+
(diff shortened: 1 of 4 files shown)


modules/using-a-pre-deployed-image-to-improve-efficiency.adoc
```diff
@@ -0,0 +1,38 @@
...
+reduce repetitive tasks. This is known as a pre-deployed image. For example,
+because every node requires the `ose-pod` image in order to run pods, each node
+has to periodically connect to the container image registry in order to pull the
+latest image. This can become problematic when you have 100 nodes attempting
+this at the same time, and can lead to resource contention on the image
+registry, waste of network bandwidth, and increased pod launch times.
+
+.Procedure
+
+To build a pre-deployed image:
```
ahardin-rh: @jeremyeder I feel like this procedure is a bit weak. What are specific steps for the user?

modules/using-a-pre-deployed-image-to-improve-efficiency.adoc
```diff
@@ -0,0 +1,38 @@
...
+. Ensure a dedicated storage device is available for CRI-O or local image
+or container storage, separate from any persistent volumes for containers.
+
+. Fully update the system, and ensure CRI-O or Docker is installed.
+
+. Ensure the host has access to all yum repositories.
+
+. Set up thin-provisioned LVM storage.
+
+. Pre-seed your commonly used images (such as the rhel7 base image), as well as
```
ahardin-rh: What is meant by pre-seed? I don't think this will translate well.

ahardin-rh:
@jeremyeder PTAL. I have a few questions on the content. 
I will take care of rebasing when I get your review comments. Thanks!

modules/using-a-pre-deployed-image-to-improve-efficiency.adoc
jeremyeder: pre-pull is probably a better word.  That's what is used upstream https://kubernetes.io/docs/concepts/containers/images/

modules/using-a-pre-deployed-image-to-improve-efficiency.adoc
jeremyeder: It is a bit weak in terms of specifics, but it is accurate.  What I worry about is how RHCOS moves us away from this requirement, and we don't yet have clarity on what the bring-your-own-host (i.e. RHEL) path will look like.  It's possible that this section will still be beneficial.  But until we have 4.0/RHCOS on AWS test results in-hand, it is hard to say on whether we should keep this section.

ahardin-rh: @jeremyeder Thanks! I made some updates. Do you think we should merge this as-is and follow up when we have more details, or do you prefer to keep this open for now?

PR #13087, opened 2018-12-11
porting SSSD assembly

kalexand-rh:
This might be part of https://jira.coreos.com/browse/OSDOCS-149

### DIFF
--- a/modules/ldap-failover-configure-openshift.adoc
+++ b/modules/ldap-failover-configure-openshift.adoc
@@ -1,0 +1,54 @@
+// Module included in the following assemblies:
+//
+// * authentication/configuring-ldap-failover.adoc
+
+[id='sssd-for-ldap-configure-openshift-{context}']
+= Configuring {product-title} to use SSSD as the basic remote authentication server
+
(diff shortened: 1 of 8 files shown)


modules/ldap-failover-configure-openshift.adoc
```diff
@@ -0,0 +1,54 @@
...
+[id='sssd-for-ldap-configure-openshift-{context}']
+= Configuring {product-title} to use SSSD as the basic remote authentication server
+
+Modify the default configuration of your cluster to use the new identity
+provider that you created. Complete the following steps on the first master host
+listed in the Ansible host inventory file.
+
+.Procedure
+
+. Open the *_/etc/origin/master/master-config.yaml_* file.
```
mburke5678: @kalexand-rh Should this reference to master-config be replaced?

kalexand-rh: It should be, but I haven't heard from QE about it. @huffmanca, have you heard anything about this use case from the Auth devs? We can pull the topic if we don't have the new data for it, but last summer @enj said that the info was important.

PR #8124, opened 2018-03-14
Add new topic for expanding PV size

ghost:
@gnufied @qinpingli PTAL and update as needed. This content was pulled from https://github.com/kubernetes/community/pull/657 and https://kubernetes.io/docs/concepts/storage/persistent-volumes/#expanding-persistent-volumes-claims.

Trello card: https://trello.com/c/e0Awnt9f/771-document-resize-perform-cloudprovider-resize-in-controller

PR #13104, opened 2018-12-13
Expanding Persistent Volumes - Removing unnecessary configuration steps.

josgonza-rh:
The configuration regarding the nodes are not necessary.

I've tested it with `Trident` and `GlusterFS` on a Openshift 3.9 cluster.

### DIFF
--- a/dev_guide/expanding_persistent_volumes.adoc
+++ b/dev_guide/expanding_persistent_volumes.adoc
@@ -24,27 +24,7 @@ Apart from that, {product-title} administrators must enable the
 xref:../architecture/additional_concepts/admission_controllers.adoc#architecture-additional-concepts-admission-controllers[Admission Controllers]
 for more information on the `PersistentVolumeClaimResize` admission controller.
 
-To enable the feature gate, set `ExpandPersistentVolumes` to `true` across the system:
-
-. Configure the appropriate xref:../admin_guide/manage_nodes.adoc#modifying-nodes[node configuration map]:
-+
-----
-# oc edit configmaps <name> -n openshift-node
-----
-+
-For example:
-+
-----
-oc edit cm node-config-compute -n openshift-node
-...
-kubeletArguments:
-...
-  feature-gates:
-  - ExpandPersistentVolumes=true
-----
-
-. Enable the `ExpandPersistentVolumes` feature gate on the master API and controller manager:
-+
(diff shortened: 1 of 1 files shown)


kalexand-rh:
@mburke5678, will you PTAL?

dev_guide/expanding_persistent_volumes.adoc
```diff
@@ -58,6 +38,12 @@ kubernetesMasterConfig:
 # master-restart controllers
 ----
 
+Then you should restart the master services
+----
+systemctl restart atomic-openshift-master-*
+----
```
mburke5678: Reminder to update this command for 3.10+

mburke5678:
@qinpingli @childsb 
You looked at the [original PR that added this section](https://github.com/openshift/openshift-docs/pull/8124). I just want to verify that the `ExpandPersistentVolumes` feature gate needs to be set on the master only, and not on nodes.

mburke5678:
@josgonza-rh One change to a sentence. Otherwise LGTM. Want to verify that the feature gate is needed only on master and not nodes.

qinpingli:
@mburke5678 feature gate `ExpandPersistentVolumes` should be set for `apiserver(master)`, `controller(master)`, and `kubelet(node)`.

liangxia:
> The configuration regarding the nodes are not necessary.
> I've tested it with `Trident` and `GlusterFS` on a Openshift 3.9 cluster.

@josgonza-rh @mburke5678 
It may work for GlusterFS without changing the node configurations. 
But it does NOT work for cinder/openstack without node change under my testing (with version openshift v3.9.59).

Also there are specific changes/settings in 3.9, 3.10 and 3.11 ( e.g, 3.9 and 3.10 use different method to change node configuration ), fixing 3.X issues in master branch directly needs extra cautions to not break the other versions.

mburke5678:
@josgonza-rh @qinpingli 
Should we leave the docs as they are to prevent problems?

josgonza-rh:
@mburke5678 I'm agree with you to leave the docs as they are to prevent problems as the current docs are general purpose and my suggestion doesn't work for all the storages (as @liangxia tests said)

mburke5678:
@josgonza-rh Thank you for your input. We do appreciate suggestions for improving the documentation. Feel free to comment any time.

PR #13112, opened 2018-12-14
FIX:BZ1659352:Provide the available parameter sets for config of non-default storageclass

bysnupy:
* Fix: [[DOCS] not working the variables related with non-default storageclass](https://bugzilla.redhat.com/show_bug.cgi?id=1659352)

* Version: `v3.10` and `v3.11`, it seems available at `v3.9` either, but I could not verify the version. 

* Description:
  Because the non-default storageclass configurations are explained under `dynamic` config section, so the `dynamic` option seems requirements of the non-default storageclass config.
  But it's not correct, and it lead wrong parameter sets so it had better to provide the specific available parameter sets based on the installer implementation for suppressing the issues.

### DIFF
--- a/install/configuring_inventory_file.adoc
+++ b/install/configuring_inventory_file.adoc
@@ -1569,7 +1569,14 @@ openshift_metrics_cassandra_storage_type=dynamic
 
 If there are multiple default dynamically provisioned volume types, such as
 gluster-storage and glusterfs-storage-block, you can specify the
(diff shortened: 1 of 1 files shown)


kalexand-rh:
@jhou1, will you PTAL?

jhou1:
I'm unfamiliar with these parameters, @anpingli @jianlinliu can any of you review? Thanks.

jianlinliu:
I did not touch with these parameters either.

kalexand-rh:
Thank you all! Merging.

PR #12977, opened 2018-11-28
draft of Power updates

kalexand-rh:
This is a port of the 3.10 Power doc updates to be merged when Power support is added to 3.11. 

@ahardin-rh, FYI

Preview build: http://file.rdu.redhat.com/kalexand/112918/3.11_power/install/disconnected_install.html

PR #13113, opened 2018-12-14
Revert "draft of Power updates"

kalexand-rh:
Reverts openshift/openshift-docs#12977

I forgot to squash. :-1:


kalexand-rh:
I don't think this will actually fix it. Closing.

PR #13096, opened 2018-12-13
Adding performance data and information about ROUTER_THREADS.

jmencak:

PR #13142, opened 2018-12-17
Adding performance data and information about ROUTER_THREADS.

ahardin-rh:
Continues the work of https://github.com/openshift/openshift-docs/pull/13096

### DIFF
--- a/install_config/monitoring/configuring-openshift-cluster-monitoring.adoc
+++ b/install_config/monitoring/configuring-openshift-cluster-monitoring.adoc
@@ -35,10 +35,10 @@ The {product-title} Ansible `openshift_cluster_monitoring_operator` role configu
 | Enable persistent storage of Alertmanager notifications and silences. This variable is set to `false` by default.
 
 |`openshift_cluster_monitoring_operator_prometheus_storage_class_name`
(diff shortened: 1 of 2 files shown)


mffiedler:
@jmencak It's not part of this doc change, but this sentence from the routing_optimization.adoc doesn't sound right:  

In general, HAProxy can saturate about 5-1000 applications, depending on the technology in use. The number will typically be lower for applications serving only static content.

Should it be HAProxy can service 5-1000 applications?  or support 5-1000 applications?   HAProxy isn't saturating the applications - it becomes saturated by them.

jmencak:
> Should it be HAProxy can service 5-1000 applications? or support 5-1000 applications? HAProxy isn't saturating the applications - it becomes saturated by them.

It depends how you think about this, but what you say is probably more correct, thank you.  @ahardin-rh , any chance you could make this adjustment part of this PR, or do you want me to do a separate PR for this?

ahardin-rh:
@mffiedler @jmencak Thank you! I will make the update.

jmencak:
I believe we discussed this in the past.  How would you phrase/make it better?  If you have a high performance backend server like nginx behind HAProxy serving static content, then it is not 5, but only 1 application.  If you have badly written applications querying a slow DBs, then it could be more than 1000.

jeremyeder:
I would perhaps add something like:

Router performance might be limited by the capabilities/performance of the applications behind it, such as language, static vs dynamic content.

ahardin-rh:
@jeremyeder Thanks! I opened https://github.com/openshift/openshift-docs/pull/13211 to apply your suggestion.

PR #13147, opened 2018-12-18
description of pod default toleration for taint is not correct

mburke5678:


mburke5678:
Closed. Opened in error.

PR #13148, opened 2018-12-18
removed dedicated from topic map

mburke5678:


mburke5678:
Closing. Not needed. Change already in place.

PR #13168, opened 2018-12-19
Changed Elasticsearch replica to Elasticsearch node

mburke5678:
i should read more often "...Ensure that you have the necessary storage for Elasticsearch. Note that each Elasticsearch replica requires its own ..." should be: ".... Elasticsearch *node* requires..."

```Choose a project. Once deployed, the EFK stack collects logs for every project within your OKD cluster. The examples in this section use the default project openshift-logging. The Ansible playbook creates the project for you if it does not already exist. You will only need to create a project if you want to specify a node-selector on it. Otherwise, the openshift-logging role will create a project.```
we should not encourage any other project name then 'openshift-logging'  I'm even certain if its a matter of just setting the var any longer.  We should re write this to just talk about the node selector with the guise that ... we'll create it for you if it doesnt exist, if  for some... additional reading i think we should remove point 6 all together and the info block below it


mburke5678:
Closed. Decided to create new branch off 3.11 for these changes. See: https://github.com/openshift/openshift-docs/pull/13169

PR #13172, opened 2018-12-19
Add info about crio node groups

sdodson:
In 3.10 and later you now need to assign a host to a node group which is configured for crio.

/cc @chrisnegus


ahardin-rh:
@chrisnegus LGTM :+1:

PR #13187, opened 2018-12-20
Marking Masters as Unschedulable Nodes false needs to be False in doc

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1417232

Tiny change. No peer review needed.


mburke5678:
Per @sdodson "False is definitely more correct, without testing it I couldn't tell you if false worked or not and we don't have time for anyone to go back and test this with 3.4."

mburke5678:
@gpei Is this something you can look at?

gpei:
LGTM

PR #13200, opened 2019-01-01
Allowing unsafe sysctls is no longer experimental

hshiina:


stlaz:
@kalexand-rh this is a change made during Kube 1.11 rebase, so this has been there since 3.11 - we must have missed it back then.

kalexand-rh:
Thank you @stlaz!

PR #13219, opened 2019-01-03
OCP 3.10 should reference RHOCS 3.10, rather than 3.11 (bz 1663065)

liangxia:
Fix https://bugzilla.redhat.com/show_bug.cgi?id=1663065

@openshift/team-documentation 

Applies to enterprise-3.10 only.


kalexand-rh:
LGTM!

PR #13239, opened 2019-01-07
Remove extra '*' from master/node config

liangxia:
@openshift/team-documentation Applies to enterprise-3.10 and above.

@weherdh  @jhou1  Could you check if the change looks good ?

You can check how it looks like currently,
https://docs.openshift.com/container-platform/3.11/install_config/configuring_gce.html#manually-configuring-master-hosts-for-gce


jhou1:
Looks good

kalexand-rh:
Thanks! Merging.

PR #13273, opened 2019-01-09
Update CSI documentation to 4.0

jsafrane:
Note to self: re-test when https://github.com/openshift/csi-driver-registrar/pull/5 is available in 4.0 images.

### DIFF
--- a/install_config/persistent_storage/persistent_storage_csi.adoc
+++ b/install_config/persistent_storage/persistent_storage_csi.adoc
@@ -214,7 +214,7 @@ spec:
             - name: socket-dir
               mountPath: /csi
         - name: cinder-driver
-          image: quay.io/jsafrane/cinder-csi-plugin
+          image: k8scloudprovider/cinder-csi-plugin:v0.3.0
           command: [ "/bin/cinder-csi-plugin" ]
           args:
             - "--nodeid=$(NODEID)"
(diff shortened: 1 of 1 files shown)


install_config/persistent_storage/persistent_storage_csi.adoc
```diff
@@ -180,7 +180,7 @@ spec:
       serviceAccount: cinder-csi
       containers:
         - name: csi-attacher
-          image: registry.redhat.io/openshift3/csi-attacher:v3.11
+          image: registry.redhat.io/openshift3/csi-attacher:v4.0
```
liangxia: Are we still using openshift3 ?

jsafrane: Good catch, fixed

vikram-redhat:
@huffmanca just making sure that this is on your radar.

liangxia:
@jsafrane Given we have bug https://bugzilla.redhat.com/show_bug.cgi?id=1665799, could you confirm whether the pr need further change ?

jsafrane:
in bug 1665799 we disabled CSI operator, which means that CSI drivers must  be installed "the old way", i.e. as described in this PR.

huffmanca:
@jsafrane, The master branch is still tracking the 3.x stream. I've created https://github.com/openshift/openshift-docs/pull/14354 to include and update this section for the 4.x docs. Closing this PR, as further discussion should occur in PR 14354.

PR #13290, opened 2019-01-10
Change three to two

jboxman:
@openshift/team-documentation 

Should there be another item in the list or is it actually two?


kalexand-rh:
The xref strategy is different in 4.0, so this content will be deprecated. If it weren't, I'd rewrite the intro to be nonspecific because the section explains four different types of links, not two or three. /shrug

PR #13294, opened 2019-01-11
Fix registry url and a typo

liangxia:
For the registry url, refer to https://docs.openshift.com/container-platform/3.11/install/disconnected_install.html#disconnected-syncing-images . You can see that all other images are using registry.redhat.io except the one updated here.

@openshift/team-documentation 
Apply to enterprise-3.11 ONLY.


kalexand-rh:
Thanks! Merging.

PR #13320, opened 2019-01-15
Fix authby value in the IPsec admin guide

roaldnefs:
Change the `authby` value from `rsasig` to `pubkey` in the IPsec configuration for the opportunistic group.

The combined certificate in the PKCS#12 format created by the `openssl` command result in a `cert key is not rsa type!` error in IPsec.

### DIFF
--- a/admin_guide/ipsec.adoc
+++ b/admin_guide/ipsec.adoc
@@ -154,7 +154,7 @@ conn private
 	# their certificate transmitted via IKE
 	rightca=%same
 	ikev2=insist
-	authby=rsasig
+	authby=pubkey
 	failureshunt=drop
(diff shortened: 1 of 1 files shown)


kalexand-rh:
/ok-to-test

@bmeng, will you PTAL or suggest a different reviewer?

roaldnefs:
The `pubkey` value is not supported by libreswan. Closing this PR for now....

roaldnefs:
For more information check the Red Hat Bugzilla report: [1666988 - authby value in the IPsec as rsasig gives "cert key is not rsa type" error](https://bugzilla.redhat.com/show_bug.cgi?id=1666988).

PR #13324, opened 2019-01-15
Update contrib doc to use a better example for root command

bergerhoffer:
Just updating the example used for a root command to a different one, since you would normally run sudo as a non-root user.

@openshift/team-documentation: A simple internal doc update for your review. Please let me know if I missed setting any labels or fields. Thanks!


kalexand-rh:
You should set the milestone to `Future Release` and add the `Peer Review Needed` label.

PR #13325, opened 2019-01-15
Update referencePolicy.type default

jboxman:
- BZ https://bugzilla.redhat.com/show_bug.cgi?id=1664612
- Fix stacked ifdef that produces undesirable HTML output

### DIFF
--- a/dev_guide/managing_images.adoc
+++ b/dev_guide/managing_images.adoc
@@ -934,33 +934,34 @@ override the image stream annotation.
 ===== Reference Policy
 
 The Reference Policy allows you to specify from where resources that reference
-this image stream tag pulls the image. It is only applicable to remote
-images (those imported from external registries). There are two options to
-choose from, `Local` and `Source`.
+this image stream tag pulls the image. It applies to only images that you import
+from external registries. There are two options to choose from: `Local` and
+`Source`.
 
 The `Source` policy instructs clients to pull directly from the source registry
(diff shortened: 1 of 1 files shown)


dev_guide/managing_images.adoc
```diff
@@ -934,33 +934,34 @@ override the image stream annotation.
 ===== Reference Policy
 
 The Reference Policy allows you to specify from where resources that reference
-this image stream tag pulls the image. It is only applicable to remote
-images (those imported from external registries). There are two options to
-choose from, `Local` and `Source`.
+this image stream tag pulls the image. It only applies to images that are
```
kalexand-rh: s/only applies to/applies to only
s/that are imported/that you import

dev_guide/managing_images.adoc
```diff
@@ -934,33 +934,34 @@ override the image stream annotation.
...
+imported from external registries. There are two options to choose from, `Local`
```
kalexand-rh: s/from,/from:

kalexand-rh (commented):
Couple of picks. If your local builds are rendering the new ifdef stack correctly, it ~~LGTM~~. You should add a line to the 3.10 release notes, too.

kalexand-rh:
@ahardin-rh, the bug asks why this change wasn't recorded in the release notes. Do we usually capture changes to default parameter values?

PR #13327, opened 2019-01-15
Added Network optimization content to 4.0 docs

ahardin-rh:
Preview Build:
http://file.rdu.redhat.com/~ahardin/01152019/4-0-networking-optimization/networking/network-optimization.html

Some questions/notes from an earlier team conversation:

- Is NCR still a thing?

- MTU settings will be different with the new operators, right?

- Optimizing the MTU for your network - this has to account for new automagic MTU PR

- How to edit node configmap for subnet option probably needs to be dropped.

- Subnet configuration will need to be re-written for new installer configs
 
- Is IPSEC still supported?

### DIFF
--- a/modules/configuring-network-subnets.adoc
+++ b/modules/configuring-network-subnets.adoc
@@ -1,0 +1,48 @@
+// Module included in the following assemblies:
+//
+// networking/network-optimization.adoc
+
+[id='Configuring-network-subnets-{context}']
+= Configuring network subnets
+
+All Pods are assigned IPs. This enables Pod to Pod communication and Pod to node
+communication without network address translation (NAT). For a CIDR in the range
+of 10.128.0.0/14, IPs in this range are, by default, assigned to the Pods. If
+you want to change this, you must adjust the `ClusterNetwork`. If you want to
+customize the IPs rolled out to services, then you must adjust the
+`ServiceNetwork`.
+
+.Procedure
+
+To configure a custom IP range, also called a _subnet_, complete the following.
+
+. Run:
++
+----
+$ ./openshift-install --dir=new-install create install-config
+----
+
+. Change to the `new-install` directory:
++
+----
+$ cd new-install
+----
+
+. Edit the `install-config.yaml` file, setting the required fields under
+`networking`:
++
+----
+networking:
+      clusterNetworks:
+        - cidr: 10.128.0.0/14
+          hostSubnetLength: 9
+      machineCIDR: 10.0.0.0/16
+      serviceCIDR: 172.30.0.0/16
+     type: OpenshiftSDN
+----
+
+. Consume the customized `install-config.yaml` file to deploy the cluster:
++
+----
+$ ./openshift-install --dir=new-install create cluster
+----
--- a/modules/optimizing-the-MTU-for-your-network.adoc
+++ b/modules/optimizing-the-MTU-for-your-network.adoc
@@ -1,0 +1,77 @@
+// Module included in the following assemblies:
+//
+// networking/network-optimization.adoc
+
+[id='optimizing-the-mtu-for-your-network-{context}']
+= Optimizing the MTU for your network
+
+There are two important maximum transmission units (MTUs): the network
+interface card (NIC) MTU and the SDN overlay's MTU.
+
+MTUs are now autodetected, so you do not need to adjust MTU settings. When the
+Cluster Network Operator runs, it determines the MTU used by the default route,
+then sets the MTU accordingly. If you want to have a fixed MTU size for your
+workloads, then you must set the MTU explicitly, then set the network
+configuration before installation.
+
+.Procedure
+
+To set the MTU explicitly, complete the following.
+
+. Run:
++
+----
+$ openshift-installer create install-config
+----
+
+. Run the following to generate the required manifests:
++
+----
+$ openshift-installer create manifests
+----
+
+. Edit the `manifests/cluster-network-02-config.yml` file.
+
+. Run the following to use the manifests to create a cluster:
++
+----
+$ openshift-installer create cluster
+----
+
+. The following is the default state of the `cluster-network-02-config.yaml` file.
+Adjust the settings as necessary:
++
+----
+apiVersion: networkoperator.openshift.io/v1
+kind: NetworkConfig
+metadata:
+  creationTimestamp: null
+  name: default
+spec:
+  additionalNetworks: null
+  clusterNetworks:
+  - cidr: 10.128.0.0/14
+    hostSubnetLength: 9
+  defaultNetwork:
+    openshiftSDNConfig:
+      mode: Networkpolicy <1>
+    otherConfig: null
+    type: OpenshiftSDN
+  serviceNetwork: 172.30.0.0/16
+status: {}
+----
++
+<1> `Networkpolicy` is the default.
++
+You can also set the `Multitenant` or `Subnet`, `vxlanPort`, and  `mtu`:
++
+----
+spec:
+  defaultNetwork:
+    type: OpenShiftSDN
+    openshiftSDNConfig:
+      mode: NetworkPolicy
+      vxlanPort: 4789
+      mtu: 1450
+      useExternalOpenvswitch: false
+----
--- a/networking/network-optimization.adoc
+++ b/networking/network-optimization.adoc
@@ -1,0 +1,47 @@
+[id='network-optimization']
+= Network optimization
+include::modules/common-attributes.adoc[]
+:context: networking
+
+toc::[]
+
+{nbsp} +
+The OpenShift SDN uses Open vSwitch, Virtual Extensible LAN (VXLAN) tunnels,
+OpenFlow rules, and iptables. This network can be tuned by using jumbo frames,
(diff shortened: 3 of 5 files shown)


ahardin-rh:
@jeremyeder PTAL

jeremyeder:
@Deepthidharwar

modules/configuring-network-subnets.adoc
```diff
@@ -0,0 +1,38 @@
+// Module included in the following assemblies:
+//
+// networking/network-optimization.adoc
+
+[id='Configuring-network-subnets-{context}']
+= Configuring network subnets
```
ahardin-rh: What does configuring the subnet help the user achieve? What are the benefits and why would you do this?

modules/optimizing-the-MTU-for-your-network.adoc
```diff
@@ -0,0 +1,76 @@
+// Module included in the following assemblies:
+//
+// networking/network-optimization.adoc
+
+[id='optimizing-the-mtu-for-your-network-{context}']
+= Optimizing the MTU for your network
+
```
ahardin-rh: I feel like we're missing a little bit of context in this section as well. Why would you need to set the MTU? What is the use case?

modules/optimizing-the-MTU-for-your-network.adoc
```diff
@@ -0,0 +1,76 @@
...
+To set the MTU explicitly, complete the following.
+
+. Run:
++
+----
+$ openshift-installer create installconfig
+----
+
+. If necessary, set the IP address blocks.
```
ahardin-rh: When is this necessary?

modules/optimizing-the-MTU-for-your-network.adoc
```diff
@@ -0,0 +1,76 @@
...
+
+. Edit the `manifests/cluster-network-02-config.yml` file.
+
+. Run:
++
+----
+$ openshift-installer create cluster
+----
+
+. Set the following options in the `cluster-network-02-config.yaml` file:
```
ahardin-rh: I think we're missing some context on the options here

modules/optimizing-the-MTU-for-your-network.adoc
Deepthidharwar: By default, with the latest automatic detection of MTUs with OCP 4.0 one does not need to tweak MTU settings but in-case one wants to have a fixed MTU size for their workloads then one can do so by tweaking the above settings.

modules/optimizing-the-MTU-for-your-network.adoc
ahardin-rh: @Deepthidharwar When would one need to set the IP address blocks?

Deepthidharwar: Setting the IP address block is same step as setting subnet. So we can look to omit this line .

Deepthidharwar (commented):
One thing to note is invoking Openshift-install command,  Do we need to have it "./openshift-install" or "Openshift-install". The subnet doc has the first option and MTU part has the second option. This needs to be uniform.

ahardin-rh:
@Deepthidharwar Thank you! I made some updates.

ahardin-rh:
@Deepthidharwar Are further changes needed before we merge? Thanks!

Deepthidharwar:
Looks good.

networking/network-optimization.adoc
```diff
@@ -0,0 +1,46 @@
+[id='network-optimization']
+= Network optimization
+include::modules/common-attributes.adoc[]
+:context: networking
+
+toc::[]
+
+The OpenShift SDN uses OpenvSwitch, virtual extensible LAN (VXLAN) tunnels,
```
adellape: Seems like it's "Open vSwitch" on http://www.openvswitch.org, the GitHub repo, etc.

networking/network-optimization.adoc
```diff
@@ -0,0 +1,46 @@
+[id='network-optimization']
+= Network optimization
+include::modules/common-attributes.adoc[]
+:context: networking
+
+toc::[]
+
```
adellape: When I don't have a heading that starts off an assembly, I like to put a `{nbsp} +` on the line right before the first line of text so that it puts a bit of white space between the TOC's horizontal line and the body text. E.g.,:

```
toc::[]

{nbsp} +
The OpenShift SDN...
```

networking/network-optimization.adoc
adellape: "Virtual Extensible LAN", initcaps?

networking/network-optimization.adoc
```diff
@@ -0,0 +1,46 @@
...
+The OpenShift SDN uses OpenvSwitch, virtual extensible LAN (VXLAN) tunnels,
+OpenFlow rules, and iptables. This network can be tuned by using jumbo frames,
+network interface cards (NIC) offloads, multi-queue, and ethtool settings.
+
+VXLAN provides benefits over VLANs, such as an increase in networks from 4096 to
+over 16 million, and layer 2 connectivity across physical networks. This allows
+for all pods behind a service to communicate with each other, even if they are
```
adellape: Throughout:

s/pod/Pod/ 
s/pods/Pods/

networking/network-optimization.adoc
```diff
@@ -0,0 +1,46 @@
...
+inner-packets are subject to normal checksumming rules to guarantee data has not
+been corrupted during transit. Depending on CPU performance, this additional
+processing overhead can cause a reduction in throughput and increased latency
+when compared to traditional, non-overlay networks.
+
+Cloud, VM, and bare metal CPU performance can be capable of handling much more
+than one Gbps network throughput. When using higher bandwidth links such as 10
+or 40 Gbps, reduced performance can occur. This is a known issue in VXLAN-based
+environments and is not specific to containers or {product-title}.
+Any network that relies on VXLAN tunnels will perform similarly because of the
```
adellape: Possibly: s/will perform/performs/

modules/configuring-network-subnets.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+All pods are assigned IPs. This enables pod to pod communication and pod to node
+communication without network address translation (NAT). For a CIDR in the range
+of 10.128.0.0/14, IPs in this range are, by default, assigned to the pods. If
+you want to change this, you must adjust the `ClusterNetwork`. If you want to
+customize the IPs rolled out to services, then you must adjust the
+`ServiceNetwork`.
+
+To configure a custom IP range, also called a _subnet_, complete the following.
```
adellape: Needs a `.Procedure`.

modules/configuring-network-subnets.adoc
```diff
@@ -0,0 +1,45 @@
...
+$ ./openshift-install --dir=new-install create install-config
+----
+
+. Change to the `new-install` directory:
++
+----
+$ cd new-install
+----
+
+. Edit `install-config.yaml` file, setting the required fields under `networking`:
```
adellape: s/Edit/Edit the/

modules/optimizing-the-MTU-for-your-network.adoc
```diff
@@ -0,0 +1,75 @@
...
+configuration before installation.
+
+.Procedure
+
+To set the MTU explicitly, complete the following.
+
+. Run:
++
+----
+$ openshift-installer create install-config
```
adellape: This procedure shows `$ openshift-installer` syntax, while an earlier procedure used `$ ./openshift-installer` syntax. Guessing the earlier one can drop the `./`.

modules/optimizing-the-MTU-for-your-network.adoc
```diff
@@ -0,0 +1,75 @@
...
+  - cidr: 10.128.0.0/14
+    hostSubnetLength: 9
+  defaultNetwork:
+    openshiftSDNConfig:
+      mode: Networkpolicy
+    otherConfig: null
+    type: OpenshiftSDN
+  serviceNetwork: 172.30.0.0/16
+status: {}
+----
```
adellape: It seems like the above is showing the default state, and the next part is showing suggested / optional changes, is that right? Asking because the above doesn't seem to show an `mtu:` setting.

modules/optimizing-the-MTU-for-your-network.adoc
```diff
@@ -0,0 +1,75 @@
...
++
+Under `openshiftSDNConfig`, you can set the mode. `Networkpolicy` is the
```
adellape: The lines about `openshiftSDNConfig` and `Networkpolicy` could maybe be callouts on the above block?

adellape (approved):
Just a few nits / questions. Overall looks cool. :sunglasses:

modules/optimizing-the-MTU-for-your-network.adoc
adellape: So if that's the case, maybe adjust the "Set the following options" line to something that is just saying the following is the default state of the file (and then after that, tell them to make some changes).

ahardin-rh:
@adellape @Deepthidharwar Thanks for the reviews :heart_decoration:

PR #13328, opened 2019-01-15
CNV 1.4 installation guide updates

ousleyp:
New deployment procedure, in progress. Also changed the repository name to reflect 1.4, though I don't know for certain that it's the correct name. Removed mention of 1.3 from Installation Guide only.


ousleyp:
@openshift/team-documentation PTAL ❄️

modules/cnv_install_1_4.adoc
```diff
@@ -0,0 +1,53 @@
...
+. Navigate to the `kubevirt-ansible` directory:
++
+----
+$ cd /usr/share/ansible/kubevirt-ansible
+----
+
+. Edit `cnv.yml` to configure `kubevirt-ansible`:
++
+----
+$ vim vars/cnv.yml
```
simon3z: @ousleyp please do not navigate to `/usr/share/ansible/kubevirt-ansible` and edit the original `vars/cnv.yml` file.
We should probably ask to copy it in the local path (E.g. `cp /usr/share/ansible/kubevirt-ansible/vars/cnv.yml cnv-vars.yml` or something similar). Then it means you have to change also the `vim` and `ansible-playbook` commands accordingly, something like: `ansible-playbook -e @cnv-vars.yml /usr/share/ansible/kubevirt-ansible/playbooks/kubevirt.yml`

modules/cnv_install_1_4.adoc
```diff
@@ -0,0 +1,53 @@
...
+----
++
+// add an example here!
+
+. Ensure that you are authenticated locally to the running OpenShift cluster.
+
+. Launch {ProductName} using the configuration file you edited earlier:
++
+----
+$ ansible-playbook -e @vars/cnv.yml --connection=local playbooks/kubevirt.yml
```
simon3z: @ousleyp please do not document as default practice to use `--connection=local` (cc @nellyc). You can mention it as a possible parameter but I'd like the common practice to be remote (as openshift-ansible).

simon3z (changes requested):

modules/cnv_install_1_4.adoc
dankenigsberg: We clearly should not edit an rpm-shipped file, but why not `cd` to shorten our paths? A user should be encourage to add her changes to a homedir file, and add `-e @~/local-cnv-vars.yml` to `-e @vars/all.yml -e @vars/cnv.yml`

Also, iirc, we should include vars/all.yml in the commandline.

I would make sure that mentioning `~/local-cnv-args.yml` is not on the main procedure. We should be working fine out of the box. Only users with special needs (e.g. off-internet) need to have this file.

modules/cnv_install_1_4.adoc
```diff
@@ -0,0 +1,53 @@
...
+
+The `kubevirt-ansible` RPM contains the latest automation for deploying 
+{ProductName}. This procedure installs all {ProductName} components to your 
+{product-title} cluster. 
+
+This procedure installs the following components:
+
+* {ProductName} core components (KubeVirt)
+* Containerized data importer (CDI) controller
+* Multus and Open vSwitch (OVS) container network interface plug-ins
```
dankenigsberg: Multus, Open vSwitch (OVS) and SR-IOV container network interface plug-ins

modules/cnv_install_1_4.adoc
```diff
@@ -0,0 +1,53 @@
...
+* Updated {product-title} web console
+
+.Prerequisites
```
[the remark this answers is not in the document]
dankenigsberg: yes. facepalm.

modules/cnv_install_1_4.adoc
ousleyp: New proposed steps, based on feedback from @simon3z and @dankenigsberg (let me know how it looks):

- `yum install kubevirt-ansible`
- `cp /usr/share/ansible/kubevirt-ansible/vars/cnv.yml local-cnv-vars.yml`
- `vim local-cnv-vars.yml`
- `cd /usr/share/ansible/kubevirt-ansible`
- `ansible-playbook -e @~/local-cnv-vars.yml -e @vars/all.yml -e @vars/cnv.yml playbooks/kubevirt.yml`
- verify install by navigating to the web UI

Feedback also welcome from @rthallisey (and anyone else).

modules/cnv_install_1_4.adoc
ousleyp: Sure, I will remove that.

modules/cnv_install_1_4.adoc
```diff
@@ -0,0 +1,53 @@
...
+* {product-title} 3.11 cluster, running
+* User with `cluster-admin` privileges
+* xref:../apb_devel/cli_tooling.adoc#apb-devel-cli[APB tool]
```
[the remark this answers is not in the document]
ousleyp: Great, I'll remove the APB tool from the prerequisites. 

@dankenigsberg - should I also remove "User with `cluster-admin` privileges"? @simon3z mentioned via email that I may be able to remove the entire section about creating a cluster admin user. This would leave the 3.11 cluster as the only prerequisite.

modules/cnv_install_1_4.adoc
dankenigsberg: I think that the last `-e` wins, so `@~/local-cnv-vars.yml` should come last. If it is really a *copy* of cnv.yml,  you do not need cnv.yml.

you could say: "in case you need to change anything in cnv.yml, copy it to ~/cnv.yml, modify it and use -e ~/cnv.yml instead of -e vars/cnv.yml"

the important bit is to keep the plain installation short and sweet. I hope that most users would not need to edit cnv.yml.

modules/cnv_install_1_4.adoc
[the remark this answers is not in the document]
dankenigsberg: I think that we *still* need a cluster admin in order to deploy, and we need to `oc login` so we have her credentials. I don't understand how we can drop it.

modules/cnv_install_1_4.adoc
simon3z: > but why not `cd` to shorten our paths?

@dankenigsberg I go by the principle to not leave the session *dirty* at the end of the documented steps.

There may be other documented actions in the next chapters that do not assume you're in `/usr/share/ansible/kubevirt-ansible`.

modules/cnv_install_1_4.adoc
```diff
@@ -0,0 +1,53 @@
...
+* `virtctl` client utility, installed
```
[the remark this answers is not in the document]
lukas-bednar: Yes.

modules/cnv_install_1_4.adoc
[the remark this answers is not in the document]
lukas-bednar: Yes.

modules/cnv_install_1_4.adoc
dankenigsberg: How about proper bash programming with

`pushd <dir>`
do your things
`popd`

modules/cnv_install_1_4.adoc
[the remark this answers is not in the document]
rthallisey: We will still need to login as an admin in order to be able to create Cluster scoped objects.  So I think you can remove the apb sections, but replace it with what Dan mentioned: ```oc login ....``` or a doc line that says something like ```You need to login as admin```

modules/cnv_install_1_4.adoc
ousleyp: Making that change now - thanks Dan.

modules/cnv_install_1_4.adoc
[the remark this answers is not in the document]
ousleyp: No worries :)

modules/cnv_install_1_4.adoc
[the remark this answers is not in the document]
ousleyp: Adding the login command as the first step of the install procedure.

modules/cnv_install_1_4.adoc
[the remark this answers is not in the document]
ousleyp: @dankenigsberg @lukas-bednar Where should I include that?

modules/cnv_install_1_4.adoc
```diff
@@ -0,0 +1,64 @@
...
+
+This procedure installs the following components:
+
+* {ProductName} core components (KubeVirt)
+* Containerized data importer (CDI) controller
+* Multus, Open vSwitch (OVS), and SR-IOV container network interface plug-ins
+* Updated {product-title} web console
+
+.Prerequisites
+* {product-title} 3.11 cluster, running
```
kalexand-rh: I'd say " A running {product-title} 3.11 cluster"

modules/cnv_install_1_4.adoc
```diff
@@ -0,0 +1,64 @@
...
+$ oc login -u system:admin
+----
+
+. Install the `kubevirt-ansible` RPM and its dependencies:
++
+----
+$ yum install kubevirt-ansible
+----
+
+. If your cluster has unique requirements, customize the
```
kalexand-rh: What am I supposed to customize in this file? I'd say something like "If your cluster \<requires or uses\> \<things you can customize here\>, modify the `/usr/share/ansible/kubevirt-ansible/vars/cnv.yml` file."

modules/cnv_install_1_4.adoc
```diff
@@ -0,0 +1,64 @@
...
+$ cp /usr/share/ansible/kubevirt-ansible/vars/cnv.yml local-cnv.yml
+----
+.. Edit the copy:
++
+----
+$ vim local-cnv.yml
+----
+
+. Launch {ProductName}:
+- If you are using the default `cnv.yml` file, run:
```
kalexand-rh: s/- If/** If 
s/are using/use

modules/cnv_install_1_4.adoc
```diff
@@ -0,0 +1,64 @@
...
++
+----
+$ ansible-playbook -e @/usr/share/ansible/kubevirt-ansible/vars/all.yml -e 
+@/usr/share/ansible/kubevirt-ansible/vars/cnv.yml 
+/usr/share/ansible/kubevirt-ansible/playbooks/kubevirt.yml
+----
+- If you copied `cnv.yml` to `~/local-cnv.yml`, run:
```
kalexand-rh: s/- If/** If

modules/cnv_install_1_4.adoc
```diff
@@ -0,0 +1,64 @@
...
++
+----
+$ ansible-playbook -e @/usr/share/ansible/kubevirt-ansible/vars/all.yml -e 
+@~/local-cnv.yml /usr/share/ansible/kubevirt-ansible/playbooks/kubevirt.yml
+----
+
+. Verify the installation by navigating to the web console at 
+`kubevirt-web-ui.your.app.subdomain.host.com`. Log in using your OpenShift
```
kalexand-rh: s/using your OpenShift/by using your {product-title}

modules/cnv_requirements.adoc
```diff
@@ -71,7 +55,7 @@ Container Platform Configuring Clusters Guide] for more information.
 
 === Block volumes
 
-{ProductName} 1.3 supports the use of block volume PVCs. In order to use 
+{ProductName} supports the use of block volume PVCs. In order to use
```
kalexand-rh: s/In order to /To

kalexand-rh (commented):
Couple of picks and a question, but overall LGTM.

modules/cnv_install_1_4.adoc
dankenigsberg: what does `{product-title}` renders into, in this context?
a perquisite of CNV is OCP, not CNV itself.

modules/cnv_install_1_4.adoc
```diff
@@ -0,0 +1,64 @@
...
++
+----
+$ vim local-cnv.yml
+----
+
+. Launch {ProductName}:
+** If you use the default `cnv.yml` file, run:
++
+----
+$ ansible-playbook -e @/usr/share/ansible/kubevirt-ansible/vars/all.yml -e
```
rthallisey: At this point, we should expect that either the user is in `/usr/share/ansible/kubevirt-ansible` or in a local copy of this dir.

rthallisey: For remote installs, we're going to use the inventory file that kubevirt-ansible ships with.  The user can edit it as needed.  So add to every `ansible-playbook` command a `-i inventory`.

rthallisey (changes requested):

modules/cnv_install_1_4.adoc
```diff
@@ -0,0 +1,64 @@
...
+----
+
+. Install the `kubevirt-ansible` RPM and its dependencies:
++
+----
+$ yum install kubevirt-ansible
+----
+
+. If your cluster has unique requirements, customize the 
+`/usr/share/ansible/kubevirt-ansible/vars/cnv.yml` file. Otherwise, skip this
```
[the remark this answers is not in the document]
dankenigsberg: Duplicating code is usually a bad practice. Why would we want to promote it?

The only thing that we can possibly support are few edits to cnv.yml. Modifying roles or playbook should not be documented, nor keeping a possibility-stale copy of them.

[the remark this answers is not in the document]
simon3z: > Duplicating code is usually a bad practice.

Agree 100%, bad practice no doubts.

> The only thing that we can possibly support are few edits to cnv.yml. Modifying roles or playbook should not be documented, nor keeping a possibility-stale copy of them.

+1 :+1:

modules/cnv_install_1_4.adoc
ohadlevy: I wonder, is there an option to use environment variables instead of copying the files?

modules/cnv_install_1_4.adoc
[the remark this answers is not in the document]
ohadlevy: I tend to agree here, if we end up copying something, that's usually a bad practice (e.g. copy of all of kubevirt-ansible).
can you:
* explain the motivation to copy the files? is it just how ansible searches for variables? 
* explain how often does a user end up changing the defaults? 

my motivation is to reduce the permutation of testing / verification to a minimum, if we can be opinionated instead of expose all of the playbook internals, the better. after all, this is our "api" to the user.

modules/cnv_install_1_4.adoc
dankenigsberg: I don't think you can use env variables (it does not show on https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable )

but I'm with you on not copying cnv.yml AT ALL. We should just say:

In order to use a custom repository instead ofregistry.access.redhat.com, add `-e registry_ur=registry.example.com -e cnv_repo_tag=local-repo-tag-for-cnv` to the command line.

To test the still-unsupported SR-IOV add `-e +deploy_sriov_plugin=true`

modules/cnv_install_1_4.adoc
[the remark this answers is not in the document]
rthallisey: @dankenigsberg All the playbooks and roles from `kubevirt.yml`.  This will look for the `kubevirt` role in `./roles` and in `/etc/ansible/roles`.  If you copy only cnv.yml, the roles will be in neither.  Perhaps the best solution here is to install all our roles and playbooks in the default ansible location.

```yaml
....
# Deploy kubevirt
- name: Deploy kubevirt role
  hosts: localhost
  connection: local
  gather_facts: False
  # unset http_proxy. required for running in the CI
  environment:
    http_proxy: ""
  roles:
    - role: kubevirt

# Deploy kubevirt web UI
- import_playbook: kubevirt_web_ui.yml
when: platform == "openshift"
...
```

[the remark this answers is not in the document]
dankenigsberg: 1. I don't want to copy `cnv.yml`. I'd like users to either specify their modification in the command line or write their own variable file.
2. I don't understand why using a `cnv.yml` from my home dir can possibly cause playbooks/kubevirt.yml to change the paths it uses. Are you sure you're solving a real problem? can you share an error log?

modules/cnv_install_1_4.adoc
rthallisey: using `-e xxxx=yyy` has highest precedence;  it's `extra vars`.

dankenigsberg: > using `-e xxxx=yyy` has highest precedence; it's `extra vars`.

correct. that's why I want users to use it to overwrite values in the chipped all.yml and cnv.yml.

modules/cnv_install_1_4.adoc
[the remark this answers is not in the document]
rthallisey: forget it.  My config file was changing the default path. Non-issue.

modules/cnv_install_1_4.adoc
rthallisey: To close the loop on this discussion, I think we all agree that we want to minimize the problem space by lowering the amount of "supported customizations".  Using extra vars will avoid copying any files and give the user the flexibility needed to make kubevirt-anisble work for their environment.  With that being said, can I get an ack from folks on the __minimal__ supported extra vars?
@simon3z @ohadlevy @dankenigsberg 

```
# Container Registry String: docker_prefix:docker_tag
docker_prefix
docker_tag
```

dankenigsberg: This is the complete list. Product management may ask you change it in the future. Even *I* may ask you to include deploy_sriov to this list in a near future. But currently this is it. Only two extra args are supported, and only for offline deployment.

modules/cnv_install_1_4.adoc
```diff
@@ -0,0 +1,64 @@
...
+** If you copied `cnv.yml` to `~/local-cnv.yml`, run:
++
+----
+$ ansible-playbook -e @/usr/share/ansible/kubevirt-ansible/vars/all.yml -e 
+@~/local-cnv.yml /usr/share/ansible/kubevirt-ansible/playbooks/kubevirt.yml
+----
+
+. Verify the installation by navigating to the web console at 
+`kubevirt-web-ui.your.app.subdomain.host.com`. Log in by using your 
+{product-title} credentials.
```
rthallisey: We'll also want to document deletion of CNV.

```yaml
ansible-playbook -e @/usr/share/ansible/kubevirt-ansible/vars/all.yml -e 
@~/local-cnv.yml /usr/share/ansible/kubevirt-ansible/playbooks/kubevirt.yml -e apb_action=deprovision
```

modules/cnv_install_1_4.adoc
rthallisey: Adding one more...
```yaml
# Container Registry String: docker_prefix:docker_tag
docker_prefix
docker_tag

apb_action
```

dankenigsberg: I don't know what `apb_action` does. This should not be a verbatim table, just a plain paragraph telling in order to do X run with `-e apb_action=bla`.

rthallisey: `apb_action=provision` and `apb_acton=deprovision`.  `apb_acton=deprovision` will be needed for deletion.

modules/cnv_install_1_4.adoc
ousleyp: Thanks, Ryan. A few questions about deprovisioning:

1. Does it matter where the command is run?
2. Are there any prerequisites other than having CNV installed in the first place?
3. What's a good way to test that deprovisioning was successful? Is there output that states the success/failure? If so, please send it so I can use it as an example.

dankenigsberg: @lukas-bednar can probably provide the `oc` command line for checking if there is any kubevirt image running on the cluster.

rthallisey: > 1. Does it matter where the command is run?
>
It should be run in the same places as deploy.  That is either in `/usr/share/ansible/kubevirt-ansible` or some place in the user's HOME directory.

> 2. Are there any prerequisites other than having CNV installed in the first place?
>
Having CNV installed is the only prerequisite

> 3. What's a good way to test that deprovisioning was successful? Is there output that states the success/failure? If so, please send it so I can use it as an example.
>
The playbook ran without errors is a success, since it will wait until everything is cleaned up before moving on

modules/cnv_install_1_4.adoc
ousleyp: Adding "-i inventory" to both install and uninstall ansible-playbook commands.  

> At this point, we should expect that either the user is in `/usr/share/ansible/kubevirt-ansible` or in a local copy of this dir.

Is this still the case since we're having users specify any modifications on the command line?

modules/cnv_install_1_4.adoc
ousleyp: @rthallisey @dankenigsberg 

1. Can you clarify if registry_url and cnv_repo_tag are supported args, or only docker_prefix, docker_tag and apb_action? 

2. Can you give an example of why someone would want to change docker_prefix and docker_tag?

Thanks!

rthallisey: > Can you clarify if registry_url and cnv_repo_tag are supported args, or only docker_prefix, docker_tag and apb_action?
>
I would stick with just `docker_prefix`, `docker_tag` and `apb_action` because `registry_url` and `cnv_repo_tag` are both covered by `docker_prefix`;  it simplifies things.

> Can you give an example of why someone would want to change docker_prefix and docker_tag?
>
If a user has an on premise image registry where they hold our images, the install will need to pull images from their internal registry rather than Red Hat's registry.

dankenigsberg: > > Can you clarify if registry_url and cnv_repo_tag are supported args, or only docker_prefix, docker_tag and apb_action?
> 
> I would stick with just `docker_prefix`, `docker_tag` and `apb_action` because `registry_url` and `cnv_repo_tag` are both covered by `docker_prefix`; it simplifies things.

Have you tested your suggestion, @rthallisey ? I see that our code uses `registry_url` directly in several places.

> 
> > Can you give an example of why someone would want to change docker_prefix and docker_tag?
> 
> If a user has an on premise image registry where they hold our images, the install will need to pull images from their internal registry rather than Red Hat's registry.

this example does not explain why a customer in his right mind would like to change `docker_tag` from its sane default `v1.4.0`.

May I repeat my request to start with my text, and only fix errors in it - if spotted? my text desribes what cnv.yml does to differentiate from kubevirt. it is also what QE does to test from brew. Let us write it down and get going.

ousleyp:
https://github.com/openshift/openshift-docs/commit/5153af094b5adbc111ca808979eadc131870d917

https://cnv-new-deploy--ocpdocs.netlify.com/openshift-enterprise/latest/cnv_install/cnv_install.html

@dankenigsberg @simon3z @rthallisey @nellyc 

Please review the latest commit/preview when you're able - thank you all for your help! 

I need further clarification on what needs to be done for remote install, but I incorporated the changes mentioned on this PR as I understood them. Please correct me if I misunderstood anything. :)

modules/cnv_install_1_4.adoc
ousleyp: Done, @dankenigsberg - check out the latest commit and let me know what I got wrong. :)

modules/cnv_install_1_4.adoc
```diff
@@ -17,7 +17,12 @@ This procedure installs the following components:
 * A running {product-title} 3.11 cluster
 * User with *cluster-admin* privileges
 
-// add prereqs for remote installation
+[NOTE]
+====
+If you execute the playbook from a remote host, that host must have the
```
dankenigsberg: I don't see the point of this note. kubevirt-ansible.rpm already requires these packages.

modules/cnv_install_1_4.adoc
```diff
@@ -44,7 +49,7 @@ the command.
 ====
 +
 ----
-$ ansible-playbook -i inventory -e @/usr/share/ansible/kubevirt-ansible/vars/all.yml 
+$ ansible-playbook -i <inventory file> -e @/usr/share/ansible/kubevirt-ansible/vars/all.yml
```
dankenigsberg: I'd prefer to have `inventory_file`. but most importantly, I miss the content of this file. I don't recall if @lukas-bednar provided you an example.

dankenigsberg (changes requested):

modules/cnv_install_1_4.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+.Procedure
+
+. Log into the {product-title} cluster as an admin user:
++
+----
+$ oc login -u system:admin
+----
+
+. Install the `kubevirt-ansible` RPM and its dependencies:
```
dankenigsberg: please start with this, as - if necessary - it would pull in the `oc` client.

dankenigsberg (changes requested):

modules/cnv_install_1_4.adoc
nellyc: we are not supposed to use all.yml, only cnv.yml

also, i believe that we should first cd into /usr/share/ansible/kubevirt-ansible
and execute the command from there in order to not need to be a privileged user

modules/cnv_uninstall_1_4.adoc
```diff
@@ -0,0 +1,39 @@
...
+. Log into the {product-title} cluster as an admin user:
++
+----
+$ oc login -u system:admin
+----
+
+. Uninstall {ProductName}:
++
+----
+$ ansible-playbook -i <inventory file> -e @/usr/share/ansible/kubevirt-ansible/vars/all.yml
```
nellyc: above comment is relevant here as well

modules/cnv_install_1_4.adoc
dankenigsberg: I'm pretty sure we *do* need to specify all.yml. cnv.yml overrides only *some* of the vars there.

dankenigsberg: @nellyc @lukas-bednar don't you use all.yml in QE?

modules/cnv_install_1_4.adoc
rthallisey: The web_ui doesn't use docker_prefix.  I must've missed it in my review.  Also, @dankenigsberg looks like network hardcodes the registry to upstream images.  I think these are both bugs @nellyc.
```yaml
multus_provisioner_repo: "docker.io/nfvpe/multus"
cni_provisioner_repo: "quay.io/schseba/cni-plugins"
ovs_provisioner_repo: "quay.io/kubevirt/ovs-cni-plugin"
```

rthallisey: ignore my comment.  cnv.yaml covers it.

modules/cnv_install_1_4.adoc
rthallisey: Update for the deploy command.
  - Remove all.yml extra_var
  - add apb_action=provision as an extra_var
```diff
-ansible-playbook -i <inventory file> -e @/usr/share/ansible/kubevirt-ansible/vars/all.yml -e @/usr/share/ansible/kubevirt-ansible/vars/cnv.yml  /usr/share/ansible/kubevirt-ansible/playbooks/kubevirt.yml
+ansible-playbook -i <inventory file> -e @/usr/share/ansible/kubevirt-ansible/vars/cnv.yml  /usr/share/ansible/kubevirt-ansible/playbooks/kubevirt.yml -e apb_action=provision
```

dankenigsberg: @ousleyp  please wait for QE approval before dropping all.yml.

modules/cnv_install_1_4.adoc
ousleyp: That makes sense. I'll remove it and switch the order of the "oc login" and package installation commands.

modules/cnv_install_1_4.adoc
dankenigsberg: Dropping `all.yml` is wrong, regardless to that BZ. Why? upstream deployment of KubeVirt includes `all.yml`. `cnv.yml` has downstream-only modifications for it. I expect that most often than not, new variables can be set in all.yml without cnv.yml caring about them. If you drop `all.yml`, cnv.yml would have to become a not-so-precise copy of it. That's a very bad practice - it would quietly break when upstream is modified.

We *may* be able to include `all.yml` from within `cnv.yml` to help shorten the command line. I do not know ansible well enough to answer that.

Adding `apb_action=provision` is functionally harmless, but I don't see the benefit thereof.

modules/cnv_install_1_4.adoc
```diff
@@ -16,26 +16,26 @@ This procedure installs the following components:
...
 ----
 $ oc login -u system:admin
 ----
 
-. Install the `kubevirt-ansible` RPM and its dependencies:
+. Change directories to `/usr/share/ansible/kubevirt-ansible`:
 +
 ----
-$ yum install kubevirt-ansible
+$ cd /usr/share/ansible/kubevirt-ansible
```
dankenigsberg: @simon3z explicitly asked not to leave the reader inside this directory. We can either use full paths as he requested, or use `pushd /usr/share/ansible/kubevirt-ansible` here, and `popd` in the end as I suggested.

Can we please get this doc converged? It feels sad to see @simon3z's comment done and then undone.

modules/cnv_install_1_4.adoc
```diff
@@ -49,9 +49,8 @@ the command.
 ====
 +
 ----
-$ ansible-playbook -i <inventory file> -e @/usr/share/ansible/kubevirt-ansible/vars/all.yml 
--e @/usr/share/ansible/kubevirt-ansible/vars/cnv.yml 
-/usr/share/ansible/kubevirt-ansible/playbooks/kubevirt.yml
+$ ansible-playbook -i <inventory_file> -e @/usr/share/ansible/kubevirt-ansible/vars/cnv.yml
```
dankenigsberg: I asked *not* to drop all.yml before a strict ack from QE that this actually works. no such ack was received, so why is this dropped?

even if an ack is received, I think it is important to keep it.

modules/cnv_uninstall_1_4.adoc
```diff
@@ -23,13 +23,17 @@ This procedure uninstalls the following components:
...
+----
+
 . Uninstall {ProductName}:
 +
 ----
-$ ansible-playbook -i <inventory file> -e @/usr/share/ansible/kubevirt-ansible/vars/all.yml 
--e @/usr/share/ansible/kubevirt-ansible/vars/cnv.yml 
-/usr/share/ansible/kubevirt-ansible/playbooks/kubevirt.yml -e 
-apb_action=deprovision
+$ ansible-playbook -i <inventory_file> -e @/usr/share/ansible/kubevirt-ansible/vars/cnv.yml
```
dankenigsberg: why do you bother to `cd` above if you use the full path here?

dankenigsberg (changes requested):
the content of `inventory_file` is still not explained.

modules/cnv_install_1_4.adoc
dankenigsberg: I just spoke with @nellyc , she told me that as of yesterday @lukas-bednar started running CI without `all.yml`. I still think it's wrong - we should do things upstream first - but you can remove `all.yml`.

modules/cnv_install_1_4.adoc
dankenigsberg: > the content of `inventory_file` is still not explained.

... or at least I cannot see it. You may say:

See "references" for an example `inventory_file` .

rthallisey: I think the problem is that cnv.yml morphed into a patch of all.yml.  Things like `kubevirt_web_ui_branding` and `deploy_sriov_plugin` were duplicated in cnv.yml.  Unfortunately, downstream doesn't have a code review process, so now cnv.yml covers the functionality of all.yml.

modules/cnv_uninstall_1_4.adoc
ousleyp: Nelly had said "also, i believe that we should first cd into /usr/share/ansible/kubevirt-ansible
and execute the command from there in order to not need to be a privileged user" in another comment. That's the only reason I added the `cd` command. But I can shorten those paths and use pushd/popd as you suggested. Will update the PR momentarily.

dankenigsberg: ok. just make sure to do this with `pushd` and finish with `popd` to make @simon3z happy.

ousleyp: Please review the changes, @dankenigsberg: https://github.com/openshift/openshift-docs/pull/13328/commits/6010dcf48c2bb4c3d411f0062cb7045bf28be1d2

modules/cnv_install_1_4.adoc
ousleyp: As mentioned in another comment, here are the latest updates - the user is no longer left in `kubevirt-ansible`: https://github.com/openshift/openshift-docs/pull/13328/commits/6010dcf48c2bb4c3d411f0062cb7045bf28be1d2

dankenigsberg: thanks! I still find it hard to review stuff on github, I'm used to a more linear review.

dankenigsberg (approved):

modules/cnv_install_1_4.adoc
ousleyp: No worries @dankenigsberg, it's extremely hard to keep track of what's being said, especially when force-pushing removes some comments. Thanks for the review!

modules/cnv_uninstall_1_4.adoc
```diff
@@ -0,0 +1,49 @@
+[[uninstall_cnv]]
+=== Uninstalling {ProductName} 1.4
+
+You can uninstall {ProductName} by adding `-e apb_action=deprovision`
```
nellyc: i would replace the 'by adding',
we have apb_action=provision when installing, 
so it should be 'by replacing the apb_action param to...'

ousleyp: Oh, good catch! I'll update that shortly and then move the BZ to ON_QA.

ousleyp: Nelly, how's this? https://github.com/openshift/openshift-docs/pull/13328/commits/1d86405db513624f8b4184dc6d743e85d59ea008

dankenigsberg: /lgtm for what it's worth.

dankenigsberg (approved):

rthallisey (approved):

PR #13346, opened 2019-01-17
[3.11]Correct the report files path about checking certificate expiration.

bysnupy:
* Fix BZ: [[DOCS] Correct the report file path of checking certificate expiration](https://bugzilla.redhat.com/show_bug.cgi?id=1666946)

* Related BZ: [Different users running upgrades from the same host causes conflict over file: /tmp/cert-expiry-report.html](https://bugzilla.redhat.com/show_bug.cgi?id=1612093)

* Version: only `v3.11`

* Description: 
  The report file path is changed to home directory as default value. 
  This change is caused by [this commit](https://github.com/openshift/openshift-ansible/commit/90c456e9098a2b4a7712c2c262a11d17b0111bef#diff-741d642cea9a5220e5564ba3e0c4e61f).
~~~
This commit ensures cert report is placed in user's
home directory + data-time stamp in filename to ensure
multiple runs and multiple users don't overwrite reports
by mistake.
~~~

### DIFF
--- a/install_config/redeploying_certificates.adoc
+++ b/install_config/redeploying_certificates.adoc
@@ -82,16 +82,16 @@ The `openshift_certificate_expiry` role uses the following variables:
 |Generate an HTML report of the expiry check results.
 
 |`openshift_certificate_expiry_html_report_path`
-|`/tmp/cert-expiry-report.html`
-|The full path for saving the HTML report.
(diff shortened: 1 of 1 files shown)


kalexand-rh:
/ok-to-test

@stuartchuan, will you PTAL?

PR #13349, opened 2019-01-17
removing BYO references

kalexand-rh:
I'm removing the assemblies from the 4.0 TOC but retaining the draft source files for now. 

see https://jira.coreos.com/browse/OSDOCS-150


kalexand-rh:
Hi @DanyC97, good questions. Last I heard, 4.0 will be AWS-only, and a wider range of options will be supported very soon. I haven't seen a public roadmap, though, so I don't want to go into too much detail.

PR #13356, opened 2019-01-18
Add link for Intel plugin doc in Example Device section

MCamp859:
Signed-off-by: MCamp859 <maryx.camp@intel.com>
@openshift/team-documentation:  please review.


kalexand-rh:
/ok-to-test

@MCamp859, which versions of OpenShift are compatible with these plug-ins?

@aheslin, FYI

kad:
@kalexand-rh it is compatible with OpenShift releases that are based on Kubernetes 1.11 code, meaning 3.11 and up.

PR #13358, opened 2019-01-21
Fix a command typo

liangxia:
@openshift/team-documentation 

Applies to enterprise-3.2 and above.


ahardin-rh:
@liangxia Thank you! 

Note to team: The 3.2 Release Notes only live in 3.2 docs, so we only need to cherry-pick there. I will search the repo for other instances of this command and update in a separate PR.

PR #13376, opened 2019-01-22
adding a mod docs crash course

kalexand-rh:
@vikram-redhat, PTAL 

After yesterday's call, I thought it might be helpful to add a very brief overview of mod docs to this topic.

PR #13377, opened 2019-01-22
updates to the mod docs guide

kalexand-rh:
@vikram-redhat, PTAL

I updated your mod docs modules based on my understanding of the current guidelines. I think we should use them as sample assemblies and modules in the contribution guide. (See https://github.com/openshift/openshift-docs/pull/13376)


vikram-redhat:
@kalexand-rh - Looks good. Thanks!

PR #13379, opened 2019-01-22
cherrypicking to master

jhoyt-rh:
doing last cherrypick -- to master


jhoyt-rh:
This is for BZ 1637616: https://bugzilla.redhat.com/show_bug.cgi?id=1637616

PR #13380, opened 2019-01-22
Improve page load performance

jboxman:
This PR makes the following changes:

- Relocate <script> tags to end of <body> tag
- Convert commercial masthead to progressive JPEG format

These changes improve the initial page loading experience.


ahardin-rh:
@wgordon17 @luciddreamz Are you the correct contacts to review this proposed change?

jboxman:
@vikram-redhat, these are the changes we discussed last week. Do we have someone that does QE for HTML/JS/CSS changes?

404-commercial.html
```diff
@@ -22,11 +22,6 @@
     <script src="https://assets.openshift.net/content/html5shiv.js"></script>
     <script src="https://assets.openshift.net/content/respond.js"></script>
     <![endif]-->
-    <!-- IE8 requires jQuery and Bootstrap JS to load in head to prevent rendering bugs -->
```
wgordon17: This comment explains why these 3 script tags should not be moved out of the `head` tag. If we're ok with dropping support for IE8 users, this change should be fine

404-community.html
```diff
@@ -28,11 +28,6 @@
     <link href="latest/_images/respond.proxy.gif" id="respond-redirect" rel="respond-redirect">
     <script src="latest/_javascripts/respond.proxy.js" type="text/javascript"></script>
   <![endif]-->
-  <script src="https://assets.openshift.net/content/modernizr.js" type="text/javascript"></script>
```
wgordon17: Same comment as above

wgordon17:
@ahardin-rh The changes seem reasonable, other than the possible IE8 bug. I would recommend deploying this to a staging environment and having someone from the actual QE schedule testing on this to ensure there are no regressions in terms of browser layout or functionality in the browsers that we would want to support

jboxman:
It's worth noting that IE 8 hasn't been receiving security updates for several years now:

https://www.microsoft.com/en-us/windowsforbusiness/end-of-ie-support

Do we have analytics that break down visitors by browser version?

vikram-redhat:
Thanks @jboxman for your PR and @wgordon17 for your review. 

The analytics from Jan 1 2018 to 22 Jan 2019 (so just over a year) show for Internet Explorer users the following versions were used.

11.0 - 95,876(81.48%)
9.0 - 18,950(16.11%)
8.0 - 1,674(1.42%)
10.0 - 595(0.51%)

Based on this I am inclined to say, yes, let's drop support for 8.0 so that the rest of folks can get a better experience.

vikram-redhat:
@wgordon17 - do you still have your stage environment where we could push this to test?

jboxman:
For completeness, there are some additional lines (ignored by other browsers completely) that can be removed if IE 8 is dropped:

```
<!--[if lt IE 9]>
    <script src="https://assets.openshift.net/content/html5shiv.js"></script>
    <script src="https://assets.openshift.net/content/respond.js"></script>
<![endif]-->
```

I can update the PR to remove these as well.

wgordon17:
Hey all, I have staging sites set up now. The biggest thing that sticks out is that Search is broken on the sites. Looks like a Javascript related issue:

+ https://commercial-docs-wgordon.6923.rh-us-east-1.openshiftapps.com/
+ https://community-docs-wgordon.6923.rh-us-east-1.openshiftapps.com/

vikram-redhat:
Thanks for setting these staging sites up @wgordon17 . When @jboxman updates this PR, will these sites get automatically updated?

@jboxman - you could also run asciibinder package from the command line (from the master branch) to replicate and debug any issues. It will give you a local environment to test on.

wgordon17:
@vikram-redhat, the stage is not set up to track anything...but I can manually update it after new commits

jboxman:
@vikram-redhat, I updated the PR and confirmed that the changes work locally. The search query for each flavor includes the expected query args and the modal populates with search results.

The HTML is identical for each flavor, so I factored out the HTML partials. The decision logic is now entirely in JavaScript.

jboxman:
@wgordon17, can you update the staging sites when you have a chance? Thanks!

wgordon17:
@jboxman Updated...and agreed, I can't induce the bug anymore

wgordon17 (approved):
So, I think this looks good. I would still recommend having someone in QE run it through all the browsers we care about on different OS's to ensure nothing else is affected, but I won't block based on that recommendation

vikram-redhat:
@xltian - would we have any bandwidth to test this PR?

xltian:
@yapei  Can your team check this?

yapei:
@vikram-redhat @xltian  @wsun1  We need 1 day to perform the testing and will report result back on Jan 29

vikram-redhat:
Thank you @yapei !

yapei:
@vikram-redhat  Do we need to evaluate performance improvement or only check there's no functional loss?

vikram-redhat:
@yapei only functional loss.

yapei:
Hi @vikram-redhat  We've tested the changes today on servers @wgordon17 created, here I will list all issues we found:

- A JavaScript error is found, it doesn't look like a issue introduced by this PR
`TypeError: null is not an object (evaluating 'document.getElementById("version-selector").value = currentVersion')   - page-loader.js:41`
- "Search" returns "No results found. Try rewording your search." for any filters in OKD 3.11
- "Search" string on docs page streches out of the button on Edge 40.15063.674.0

For commercial docs
- search icon `<span class="fa fa-search">` does not work on the welcome page

vikram-redhat:
@yapei thank you!

@wgordon17 correct me if I am wrong here, but search would only work on the actual site, right?

The last issue with Javascript error is probably because of the same reason as well (not running on the actual site).

jboxman:
For community docs, the error in `_javascripts/page-loader.js` is due to `version-selector` DOM element not existing on the page.

I don't receive this error for commercial and the element in present on the page.

I think this is an easy fix.

jboxman:
@yapei, at what page size does the button text overflow? I'm not seeing the behavior in Edge 15 or Edge 18 on Windows 10:

screen shot 2019-01-30 at 10 44 19 am

jboxman:
Curiously, I'm not sure our search on docs.okd.io is currently working. I tried a search for _cluster_ on both docs.openshift.com and docs.okd.io. Only the former returned results.

```
https://help.openshift.com/search/search_custom.php?q=cluster+url%3A*%5C%2F3.11%5C%2F*&l=docs_cp&si=0

https://help.openshift.com/search/search_custom.php?q=cluster+url%3A*%5C%2F3.11%5C%2F*&l=docs_origin&si=0
```

Here's what I'm getting back using Postman using a URL constructed by from the JavaScript running from `https://docs.okd.io/3.11/welcome/index.html`:

```javascript
{"response":{"version":11.1,"status":0,"q":"pod url:*\\\/3.11\\\/*","query_id":"f25c8a57bf0f4a1eb081cebd56f9e1fe","exec_time":0.02,"query_time":5,"page_size":10,"page_number":1,"record_count":0,"page_count":1}}
```

screen shot 2019-01-30 at 11 31 40 am

jboxman:
It does, however, work for version 3.10 (`https://help.openshift.com/search/search_custom.php?q=pod+url%3A*%5C%2F3.10%5C%2F*&l=docs_origin&si=0`).

So this issue is unrelated to the PR.

jboxman:
I can confirm that the search icon on the docs landing page for the live version of docs.openshift.com does not work.

This can be fixed in a separate PR.

wgordon17:
Interesting, but the "Enter" key does at least still work.

wgordon17:
Also, the staging sites have been rebuilt with the latest code

jboxman:
@yapei, can you repeat the tests again on the staging sites for community and commercial?

Thank you!

yapei:
@jboxman  Sure, will report test result back when finished

SSshahan:
Hi @vikram-redhat,  We've tested latest both staging doc sites,  list all issues we found today:

**For commercial doc**: 
1. (Chrome)click  the links which have no version could be selected on the welcome page, such as 'Dedicated' or 'Online', it will show js error like: 

```javascript
page-loader.js:41 Uncaught TypeError: Cannot set property 'value' of null
    at selectVersion (page-loader.js:41)
    at onload (index.html:52)
```
2.  Wrong powered  logo at the bottom-right corner on the welcome page, It should be Openshift not Openshift Online I think.

**For community doc**:
1. 'search' text overflow still could reproduce on Firefox64 and Edge40, it's strange thing that some browsers may not be reproduced, even for FF64
2. The link of "Red Hat OpenShift Online" on the version selector for landing page is "https://www.openshift.com/", but expect "https://www.openshift.com/products/online/" 

General issue accross the different products sites:
1. the catalog of RestApi page such as https://community-docs-wgordon.6923.rh-us-east-1.openshiftapps.com/latest/rest_api/api/v1.APIResourceList.html and https://community-docs-wgordon.6923.rh-us-east-1.openshiftapps.com/latest/rest_api/apis/v1.APIGroup.html
: need to specify the detail api name in the catalog list.

some of above issues maybe unrelated to the PR, we could report bugs if needed.

jboxman:
For commercial 1), for some reason the JavaScript isn't the latest version in the PR. Instead it's the older version:

```javascript
function selectVersion(currentVersion) {
  document.getElementById("version-selector").value = currentVersion;
  // alert(currentVersion);
}
```

jboxman:
@wgordon17, can you refresh the staging sites? Thanks!

wgordon17:
@jboxman The staging sites have been updated

jboxman:
@wgordon17, I updated `page-loader.js` in ff3121640353c445d6d1865e21a48c184b9a3a86, but for some reason the staging site has the old version without the `null` element test. Can I fix this somehow? Thanks.

wgordon17:
@vikram-redhat A change to the JS, like @jboxman mentioned....does that require cherry-picking to all the other branches as well, in order for it to work on the versioned doc pages?

jboxman:
@vikram-redhat @wgordon17 my suspicion is that the top level HTML files are pulled from master, but other directories and files are probably branch specific. So that would mean that the commit with the JavaScript fix would need to be cherry picked to each branch. (Looking at the URLs, every named or versioned branch has its own `_javascripts/` directory.)

vikram-redhat:
@jboxman @wgordon17 - yes that is right. AsciiBinder builds the javascripts, images and stylesheets individually for each branch: https://github.com/redhataccess/ascii_binder/blob/master/lib/ascii_binder/engine.rb#L364

jboxman:
@wgordon17 looks like I had a typo in the JavaScript. I've fixed it and pushed the change. Can you update the staging sites with the fix? (This may finally be the last time.)

Thanks.

wgordon17:
@jboxman I've updated the staging sites. Please confirm

jboxman:
Oops, merge conflict in the file @wgordon17:

```javascript
function selectVersion(currentVersion) {
<<<<<<< HEAD
  document.getElementById("version-selector").value = currentVersion;
=======
  var el = document.getElementById("version-selector");
  if(el) {
    el.value = currentVersion;
  }
>>>>>>> bb39e91ad... Fix invalid variable reference
  // alert(currentVersion);
}
```

wgordon17:
🤦‍♂️ Thanks @jboxman I've re-done the branches and rebuilt the staging sites. The fix is only cherry-picked to 3.9, 3.10, and 3.11 for you to verify.

jboxman:
@wgordon17, still seeing a syntax error for 3.9-3.11 from a private window using Chrome, for example:

https://commercial-docs-wgordon.6923.rh-us-east-1.openshiftapps.com/container-platform/3.9/welcome/index.html

wgordon17:
Hey @vikram-redhat, I think I'm a bit out of my depth on this one. We might even need @nhr. So, you can see that my branches don't include that merge conflict https://github.com/wgordon17/openshift-docs/blob/master/_javascripts/page-loader.js.

So I'm thinking it's gotta be something that `asciibinder` is doing? The `assemble` script only performs `git checkout --track` on all the remote branches https://github.com/openshift/openshift-docs/blob/master/.s2i/bin/assemble#L23, so I wouldn't think that it's the assemble script creating the merge conflicts. That is why I'm inclined to blame `asciibinder`, but I honestly don't know how to proceed further

nhr:
@wgordon17 I won't be of much help here. It looks like the JS error is related to the version selector in the "breadcrumb" nav element. That was added long after I stepped away from this. As a general comment: the Fedora docs are built with [Antora](https://antora.org) these days, and sooner or later the OpenShift team may want to move in that direction.

nhr:
Looking more at this error, `page-loader.js` in @jboxman 's example clearly contains git diff notations, like the merge wasn't actually performed. Starting at line 40 of that file I am seeing:

`function selectVersion(currentVersion) {
<<<<<<< HEAD
  document.getElementById("version-selector").value = currentVersion;
=======
  var el = document.getElementById("version-selector");
  if(el) {
    el.value = currentVersion;
  }
>>>>>>> bb39e91ad... Fix invalid variable reference
  // alert(currentVersion);
}`

Which is the likely cause of both of the JS errors that I am seeing when I load this in  an incognito tab in Chrome.

wgordon17:
Sorry @jboxman I spent wayyy too long on this to finally discover it's OpenShift's fault. I *had* fixed it, but the secondary builds rely on Apache to build, and for some reason that imagestream disappeared. Long story short, my build fixed it, but never got deployed live...expect it to be fixed shortly! 

Sorry for the false alarm @vikram-redhat and @nhr ;)

wgordon17:
@jboxman It's looking better now! Please verify 👍

jboxman:
@wgordon17 thank you!

jboxman:
Hi @yapei,

Can you take a look at the staging sites again?

* https://commercial-docs-wgordon.6923.rh-us-east-1.openshiftapps.com/
* https://community-docs-wgordon.6923.rh-us-east-1.openshiftapps.com/

Thanks.

vikram-redhat:
> As a general comment: the Fedora docs are built with [Antora](https://antora.org) these days, and sooner or later the OpenShift team may want to move in that direction.

Hey @nhr 

Yup that was the plan for OpenShift docs till we hit a few snags :(

1. 4.0 docs are being rewritten from scratch in a very short time frame with a very small team and the onus to make a whole new infrastructure based on Antora was going to introduce too many bottlenecks and points of failure. So the risk was too high.

2. We lost the folks necessary to make the transition of docs from docs.openshift.com to the customer portal recently. There aren't any folks that can maintain the existing scripts, let alone create new ones if we moved to Antora. 

So unfortunately, till we have some stability and certainty we would need to rely on AsciiBinder.

yapei:
@jboxman We performed another round of testing, this time no issues found, same browser are covered

jboxman:
@vikram-redhat, I think this is ready to go; what do you think? Thanks!

vikram-redhat:
@jboxman Let's do this!

wgordon17:
@vikram-redhat Just remember to be sure to cherry pick the Javascript stuff ;)

jboxman:
The cherry-pick looked onerous, so I finally created a .patch, manually clipped files that didn't apply, and used `git apply` on 3.7, 3.9-3.11.

PR #13381, opened 2019-01-22
Added release notes for OCP 3.11.69

ahardin-rh:
Currently scheduled to ship January 29


ahardin-rh:
@sheriff-rh FYI: these advisories shipped today, so I bumped the release date to today's date and also bumped the latest-tag to the latest version number for 3.11 here. I will do the same with the other advisories and tag you so that you can see examples of how this is done. :fireworks:

PR #13388, opened 2019-01-23
Remove unused atomic pages

jboxman:
These pages haven't been used since `_templates/_page_atomic.html.erb` was removed in 028fed50ba35011acfa0fef694b5e93a02992f16 on 9 May 2018.


jboxman:
@vikram-redhat, these files aren't necessary anymore. It's easier to understand which template is used where if we remove them.

vikram-redhat:
@jboxman overall lgtm. Can you make sure to test it locally first as well? asciibinder package

PR #13413, opened 2019-01-24
[enterprise-3.11] Increment version tag for OCP

jboxman:
This relates to BZ:

- https://bugzilla.redhat.com/show_bug.cgi?id=1666540

### DIFF
--- a/architecture/additional_concepts/authentication.adoc
+++ b/architecture/additional_concepts/authentication.adoc
@@ -689,7 +689,71 @@ responses from `/oauth/authorize`, and how they should be handled:
 |Other  | Other                                                                                                                                            | Fail, optionally surfacing response body to the user
 |===
 
+For example, 
+
+[source, bash]
+----
+$ curl -u <username>:<password>
+'https://<master-address>:8443/oauth/authorize?client_id=openshift-challenging-client&response_type=token' -skv / <1>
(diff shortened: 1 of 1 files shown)


jboxman:
@openshift/team-documentation 

I'm trying to get to the genesis of where  `:latest-int-tag:` and `:latest-registry-console-tag:` originated. We aren't using them now, but they provide an indication that at one time, the disconnected installation needed differing tags.

It looks like in 19883b2a9f1847ecbad853f95fe414d67e461fbb this file first came into being, so these variables predate 12 June 2018. (`git log -p --all -S '{latest-registry-console-tag}' -- install/disconnected_install.adoc`)

This commit (b9364b6b629c0a032d40ef022113cf93c7d0dc2c) shows how one of the now orphaned variables was used. And here is that page: https://docs.openshift.com/container-platform/3.4/install_config/install/disconnected_install.html

So my question is, why would a disconnected installation tag not be the same as the tag for the latest release? For earlier versions of Open Shift, more granularity was necessary, but we aren't doing that now.

jboxman:
Also, using https://github.com/openshift/openshift-docs/pull/13465 now with a proper branch name.

kalexand-rh:
@adellape, PTAL

PR #13425, opened 2019-01-24
osdocs-227 accessing registry modules

bmcelvee:
https://jira.coreos.com/browse/OSDOCS-227 (part 2)

This bucket will require SME review.


bmcelvee:
Need to follow up and update with `oc registry` (e.g., `oc registry login`) and podman commands.

PR #13427, opened 2019-01-24
don't overcommit your hypervisor

kalexand-rh:
From https://github.com/openshift/openshift-docs/pull/13050


kalexand-rh:
@sdodson, @sjenning, and @DanielFroehlich agreed with this change on the source PR, so I'm going to merge the change.

PR #13440, opened 2019-01-25
Fix "include file not found" error

jboxman:
- https://bugzilla.redhat.com/show_bug.cgi?id=1663361

This resolves the following build error:

```
asciidoctor: ERROR: install_config/topics/manually_configuring_for_vsphere.adoc: line 190: include file not found: /Users/jasonb/Self/work/openshift-docs/install_config/topics/install_config/topics/applying_configuration_changes.adoc
```

### DIFF
--- a/install_config/configuring_vsphere.adoc
+++ b/install_config/configuring_vsphere.adoc
@@ -41,6 +41,8 @@ include::install_config/topics/configuring_for_vsphere.adoc[leveloffset=+2]
 
 include::install_config/topics/manually_configuring_for_vsphere.adoc[leveloffset=+2]
 
+include::install_config/topics/applying_configuration_changes.adoc[leveloffset=+2]
+
 include::install_config/topics/configuring_for_vsphere_storage.adoc[leveloffset=+2]
 
 include::install_config/topics/red_hat_openshift_container_storage.adoc[leveloffset=+2]
--- a/install_config/topics/manually_configuring_for_vsphere.adoc
+++ b/install_config/topics/manually_configuring_for_vsphere.adoc
@@ -187,4 +187,3 @@ RFC1123 compliant.
 [[vsphere-applying-configuration-changes]]
 == Applying Configuration Changes
 
-include::applying_configuration_changes.adoc[leveloffset=+2]
--- a/install_config/topics/manually_configuring_for_vsphere.adoc
+++ b/install_config/topics/manually_configuring_for_vsphere.adoc
@@ -187,4 +187,4 @@ RFC1123 compliant.
 [[vsphere-applying-configuration-changes]]
 == Applying Configuration Changes
 
-include::install_config/topics/applying_configuration_changes.adoc[leveloffset=+2]
+include::applying_configuration_changes.adoc[leveloffset=+2]


jboxman:
@openshift/team-documentation 

The CI build fails with this. It does build locally. What's the proper way to include this file?

```
IOError: [Errno 2] No such file or directory: '/home/travis/build/openshift/openshift-docs/applying_configuration_changes.adoc'
The command "python build.py --distro openshift-enterprise --product "OpenShift Container Platform" --version 3 --no-upstream-fetch" exited with 1.
```

mburke5678:
@jboxman That `python build.py` message always seems to show up. The real problem is the line before `IOError: [Errno 2] No such file or directory: '/home/travis/build/openshift/openshift-docs/applying_configuration_changes.adoc'`
Although, I am not sure what the problem there is. The file is used in multiple locations and appears in _Configuring for AWS_ and _Configuring for OpenStack_.
https://docs.openshift.com/container-platform/3.11/install_config/configuring_openstack.html#openstack-applying-configuration-changes

Maybe remove the line from your file and see if the PR passed Travis? At least you would know if that is the problem....

mburke5678:
Does the include need another slash?
`include::/install_config/topics/applying_configuration_changes.adoc[leveloffset=+2]`

vikram-redhat:
The issue here is that the final include is in the wrong place. I think asciibinder is struggling to resolve the include which is in another include:

configuring_vsphere.adoc includes manually_configuring_for_vsphere.adoc which in turn includes applying_configuration_changes.adoc. 

Although this works for customer portal docs (you can check that the issue is not present: https://access.redhat.com/documentation/en-us/openshift_container_platform/3.10/html-single/configuring_clusters/#vsphere-configuring-masters-manual_configuring-for-vsphere) that is because there are different build systems for docs.openshift.com and portal. It works on the customer portal docs because it can resolve these includes correctly, but asciibinder doesn't.

Trying to fix it by removing "install_config/topics" works for asciibinder (as you found out by running asciibinder locally), but fails for the customer portal docs, and that is why this build is failing. It validates builds for customer portal docs not docs.openshift.com (long history..).

Leaving all that aside, and looking at the where this issue is, it looks like that the inclusion of applying_configuration_changes.adoc in the manually_configuring_for_vsphere.adoc is not correct (technically, modularization-wise and content-wise). If you look at where applying_configuration_changes.adoc is included in other places you can see it is included from within the main assembly, and not from another modules (called topics here - again, long history).

So, to fix this properly, we need to:

1. Remove the inclusion of applying_configuration_changes.adoc from manually_configuring_for_vsphere.adoc.
2. Include it just after configuring_for_vsphere.adoc and manually_configuring_for_vsphere.adoc are included in the configuring_vsphere.adoc file (https://github.com/openshift/openshift-docs/blame/master/install_config/configuring_vsphere.adoc#L42). Thus, it should be included after both automatic and manual vsphere are included and then the "applying configuration changes" bit is added.

This will satisfy the requirements of both the customer portal and d.o.c build systems and is correct content wise.

jboxman:
@vikram-redhat, what happens to

```
[[vsphere-applying-configuration-changes]]
== Applying Configuration Changes
```

in `install_config/topics/manually_configuring_for_vsphere.adoc`? After removing the include, that header just dangles there.

mburke5678:
@jboxman Do you have a preview of this latest change?

jboxman:
@mburke5678, I looked locally and the content appears to be in the proper place.

```
open _preview/openshift-enterprise/bz-1663361/install_config/configuring_vsphere.html
```

jboxman:
@vikram-redhat, if this looks good to you, I can merge it and cherry pick.

PR #13446, opened 2019-01-25
updating install drafts

kalexand-rh:
https://jira.coreos.com/browse/OSDOCS-190


_topic_map.yml
```diff
@@ -64,6 +64,19 @@ Topics:
 - Name: Architecture
   File: architecture
 ---
+Name: Installing clusters
+Dir: installation
```
vikram-redhat: @kalexand-rh as per our new buckets, this directory/bucket/folder should be install-RHCOS.

vikram-redhat: Thanks @kalexand-rh - make sure to make that edit in the buckets list.

kalexand-rh: Yup - For 4.0, it could really be installing or installing-AWS. I tagged Ali in the bucket spreadsheet, and I'll open a PR to change the name with the one she prefers.

PR #13468, opened 2019-01-29
updates to the custom install draft

kalexand-rh:
Updates to the customized install for https://jira.coreos.com/browse/OSDOCS-190

Preview build here: http://file.rdu.redhat.com/kalexand/012918/osdocs190/installing-aws/installing-customizations-cloud.html

@wking, will you PTAL? Because we only document supported changes, I've removed some of the information about changing Ignition Configs and Kubernetes objects that was in https://github.com/openshift/installer/pull/1088


modules/installation-configuration-parameters.adoc
```diff
@@ -1,54 +1,110 @@
...
+
+|`machines.platform.aws.type`
+|The EC2 instance type for the control plane machines.
+|Valid link:https://aws.amazon.com/ec2/instance-types/c5/[AWS instance type],
+such as `c5.9xlarge`.
+
+|`machines.platform.aws.zones`
+|The availability zones where the installation program creates machines for this
+pool.
+|Valid AWS availability zone, such as `us-east-1c`.
```
wking: Mention that this is [a YAML sequence][1]?  Although you could probably use "list" or "array" as synonyms, since YAML doesn't distinguish at that level of detail, and the other terms may be more familiar for readers, I dunno.  [JSON uses "array"][2].  Makes it nice and easy, right? :p

[1]: https://yaml.org/spec/1.2/spec.html#sequence//
[2]: https://tools.ietf.org/html/rfc7159#section-5

modules/installation-configuration-parameters.adoc
```diff
@@ -1,54 +1,110 @@
...
+such as `us-east-1`
 
-|`base-domain`
-|The base domain of your cloud provider. This value is used to <do a thing>.
-|A FQDN, such as `dev.example.com`.
-|n/a
+|`platform.aws.userTags`
+|A map of keys and values that the installation program adds as tags to all
+resources that it creates.
+|Key value pairs in the `<key>: <value>` format. For more information about
```
wking: They don't have to have this syntax, they just have to be [YAML mappings][1].  The format you describe is a [block mapping][2], but you could use [flow mappings][3] too.  Probably best to just drop the format sentence and leave it to your earlier "A map..." and the example YAML to get the point across, without wading into YAML syntax.

[1]: https://yaml.org/spec/1.2/spec.html#mapping//
[2]: https://yaml.org/spec/1.2/spec.html#style/block/mapping
[3]: https://yaml.org/spec/1.2/spec.html#id2790832

modules/installation-configuration-parameters.adoc
```diff
@@ -1,54 +1,110 @@
...
+AWS IAM tags, see link:https://docs.aws.amazon.com/IAM/latest/UserGuide/id_tags.html[Tagging IAM Entities]
+in the AWS documentation.
```
wking: Why call out IAM tags here?  If we wanted to link to docs, I'd have expected something generic like [this][1].

[1]: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html

modules/installation-default-aws-config-yaml.adoc
```diff
@@ -0,0 +1,82 @@
+// Module included in the following assemblies:
+//
+// * installing-aws/installing-customizations-cloud.adoc
+
+[id='installation-default-aws-config-yaml-{context}']
+= Default `install-config.yaml` file for AWS
+
+The default `install-config.yaml` file defines only the information that is
+required to deploy your {product-title} cluster. The installation program
```
wking: This is not strictly true.  For example, see [here][1] and surrounding discussion where we decided to include some properties in the wizard-generated `install-config.yaml` for which we have internal defaults.  But the generated `install-config.yaml` certainly defers some configuration to later logic, so I dunno how much you want to discuss vs. glossing over at this stage.  But I thought I'd point it out so you could make an informed decision, whichever way you go.

[1]: https://github.com/openshift/installer/pull/902#discussion_r246956646

modules/installation-initializing.adoc
```diff
@@ -14,20 +14,32 @@ You can customize your installation of {product-title} on a compatible cloud.
...
+creates.
+.. At the prompts, provide the configuration details for your cloud:
+... Select an SSH key to use to access your cluster machines.
+... Select AWS as the platform to target.
+... Select the AWS region to deploy the cluster to.
+... Select the base domain for the Route53 service that you configured for your cluster.
+... Enter a descriptive name for your cluster.
+... Paste the pull secret that you obtained from link:https://try.openshift.com[the OpenShift start page].
+
+. Modify the `./install-config.yaml` file. You can find more information about
```
wking: At this point you have established a placeholder for the asset directory, so you can use `<installation-directory>/install-config.yaml`.  Or leave the directory implicit and use `install-config.yaml`, but `./` resolves to the current working directory, so readers might not recognize an implicit asset directory, look in their working directory, and not find the file.

wking (commented):
I haven't looked at the text you're removing, but here are some suggestions for the text you're adding.  Nothing seems particularly critical, so feel free to pick and choose whatever feedback that resonates ;).

modules/installation-about-custom.adoc
```diff
@@ -1,95 +1,51 @@
...
-For example, if you wanted to create a different number of worker machines, you
-must first run the installation program with the `install-config` target with the
-`openshift-install create install-config` command. After you provide the required
-parameters to the interactive installer, it writes the installation
-configuration into the target directory. You can then modify the installation
-configuration and run the installation program with the `cluster` target with the
-`openshift-install create cluster` command. The installation program reads the install
-configuration from disk, removes it from the target directory, and creates a
-cluster from the provided configuration.
+You can use the {product-title} installation program to customize four levels
```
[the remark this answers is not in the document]
kalexand-rh: Because our [glossary](https://doc-stage.usersys.redhat.com/documentation/en-us/ccs_internal_documentation/1.0/html-single/glossary_of_terms_and_conventions_for_product_documentation/#i) says to not use "the installer" in product docs.

modules/installation-configuration-parameters.adoc
```diff
@@ -1,54 +1,110 @@
...
 |====
-|Parameter|Description|Value format|Default value
+|Parameter|Description|Values
 
-|`provider`
-|The cloud provider to host the cluster.
-a|Enter one of the following values:
+|`machines.platform.aws.iamRoleName`
+|The IAM role assigned to the machines in the pool.
+|
```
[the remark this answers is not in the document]
kalexand-rh: Can I have more information to put in it instead? I couldn't tell if these were custom-generated or if there was a list of standard ones. :)

kalexand-rh:
Thank you @wking! After reading your comments, I have one [immediate request](https://github.com/openshift/openshift-docs/pull/13468#discussion_r252338817). I'll make more updates to address your comments and add more information.

modules/installation-configuration-parameters.adoc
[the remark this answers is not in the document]
wking: > I couldn't tell if these were custom-generated or if there was a list of standard ones. :)

Ah, actually, probably just drop this entry entirely.  We have plans to remove IAM instance roles (openshift/installer#697), they're just currently [blocked on upstream kubelet work][1].

[1]: https://github.com/openshift/installer/pull/697#issuecomment-451582731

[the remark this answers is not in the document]
wking: I've filed openshift/installer#1154, dropping the user-facing knob.  That PR also bumps the install-config version to v1beta2, so you may need to update your examples as well.

modules/installation-about-custom.adoc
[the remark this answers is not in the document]
wking: > Because our glossary says to not use "the installer" in product docs.

Huh.  Do you know if the motivation for that is recorded anywhere?  I'm not sure if I should update the openshift/installer docs or not.

modules/installation-configuration-parameters.adoc
```diff
@@ -1,54 +1,110 @@
...
 
 |`provider`
 |The cloud provider to host the cluster.
-a|Enter one of the following values:
+|`aws`
+
+|`baseDomain`
+|The base domain of your cloud provider. This value is used to create routes
+to your {product-title} cluster components.
+|A fully-qualified domain or subdomain name, such as `dev.example.com`.
```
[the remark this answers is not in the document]
kalexand-rh: That is a tongue twister. So, when we tell users to set up their domain or subdomain, is there extra guidance that we should give them? Like, you might not want to create separate subdomains for your different clusters because the route is a combination of the cluster name and domain?

modules/installation-configuration-parameters.adoc
```diff
@@ -1,54 +1,110 @@
...
+
+|`region`
+|The name of your cluster.
+|A valid AWS region, such as `us-east-1`.
 
-* aws
-* azure
-* openstack
+|`pullSecret`
+|The pull secret that you obtained from link:try.openshift.com[the OpenShift start page].
```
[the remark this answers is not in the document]
kalexand-rh: As long as that link doesn't change for GA, that's totally fair. (For GA, I'd like to change some of those links to the product docs.)

modules/installation-configuration-parameters.adoc
kalexand-rh: Why not both? I'll call it a list and provide that yaml sequence link.

modules/installation-configuration-parameters.adoc
kalexand-rh: Because it's often helpful to provide an example, and in this draft, the sample YAML isn't visible on the same screen.

modules/installation-configuration-parameters.adoc
kalexand-rh: Thank you for the better link! Is the description still acceptable, or is there better information to provide?

modules/installation-default-aws-config-yaml.adoc
kalexand-rh: Well. The description needs to be accurate here. If there's more nuance to the parameters that this file presents and what users can reasonably be expected to do with it, then I need to explain it somewhere. I need more info, though.

Which of the parameters that are in this "default" file but not set by the installation program can a user modify?

modules/installation-initializing.adoc
kalexand-rh: You're right - I changed several references to this file.

kalexand-rh (commented):
Thanks @wking! I've made more updates and have a few more questions and some commentary.

modules/installation-about-custom.adoc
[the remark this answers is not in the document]
kalexand-rh: That decision was made before my time, and I don't see a record for "why" in the word nerds notes.

Eh, I don't think that you have to change the openshift/installer docs because they won't feed into the official GA documentation.

modules/installation-about-custom.adoc
```diff
@@ -1,95 +1,51 @@
...
+
+When you select values for the prompts that the installation program presents,
+you customize {product-title}. You can further modify the cluster platform
+by modifying the `install-config.yaml` file that the installation program
+uses to deploy your cluster. In this file, you can make changes like setting the
+number of machines that the control plane uses, the type of virtual machine
+that the cluster deploys, or the CIDR range for the Kubernetes service network.
+
+It is possible, but not supported, to modify the Kubernetes objects that are injected into the cluster.
+A common modification is additional manifests in the initial installation.
```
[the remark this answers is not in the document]
wking: > Is there a good way for users to identify which Kubernetes objects are safe to change?

You can find out what the CVO manages by inspecting the update payload, but that doesn't cover indirect dependencies (e.g. objects pushed by CVO-managed operators).  And it looks like Kubernetes considered and then [rejected][1] recording creator information.  As long as they stick to namespaces/projects they create, they'll be fine.  For namespaces we create and cluster-scoped objects, I'm not sure what an easy way would be to check whether a change would be supported short of making the edit and seeing if it gets stomped ;).  Maybe @abhinavdahiya or @smarterclayton have some more elegant ideas.

[1]: https://github.com/kubernetes/kubernetes/issues/13306#issuecomment-351297076

modules/installation-configuration-parameters.adoc
wking: > Why not both? I'll call it a list and provide that yaml sequence link.

Fine with me.

modules/installation-configuration-parameters.adoc
wking: Your new "Any valid YAML map, such as..." wording looks fine to me, although I don't seem to be authorized to mark this thread resolved.

modules/installation-configuration-parameters.adoc
wking: > Is the description still acceptable, or is there better information to provide?

Your [current wording][1] looks fine to me.  Was there more information you were fishing for?

[1]: https://github.com/openshift/openshift-docs/commit/3a0157c04c643de1e77c14b6ada661660ade6559#diff-7492465e78e37df8219ae1117519d026R106

kalexand-rh: The link was specifically to IAM tags before, but we made it more general. I wanted to be sure that the description was still fair.

modules/installation-default-aws-config-yaml.adoc
wking: > Which of the parameters that are in this "default" file but not set by the installation program can a user modify?

Users can modify everything in the `install-config.yaml`, and if they choose invalid values (e.g. configuring zero masters), we'll exit with a validation error.  But here are the three classes of parameters:

* Given in the wizard-created file, and...
    * The installer has no default.  For example, `baseDomain` or the cluster name.  The wizard will prompt you for these, but if you supply your own `install-config.yaml` we'll error out if they aren't populated.
    * The installer can generate a default.  For example, machine-pool `replicas`.  These are in the wizard-created file because it's convenient for the implementation, and there didn't seem to be any harm in that approach (more details in the discussion linked from my earlier comment in this thread).  If you want, you can supply your own `install-config.yaml` without them, and the installer will happily inject its built-in defaults.
* Not given in a wizard-created file.  For example, the `rootVolume` machine-pool configuration on AWS.  The installer is able to generate defaults for these, but it does so in an asset that consumes the install-config, and not in the install-config asset itself.

modules/installation-about-custom.adoc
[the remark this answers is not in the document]
wking: > And it looks like Kubernetes considered and then rejected recording creator information.

Although creators can [set `ownerReference` explicitly if they need to][1].  Maybe we just need to push more operators to do that (if they aren't already).

[1]: https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents

modules/installation-configuration-parameters.adoc
wking: > The link was specifically to IAM tags before, but we made it more general. I wanted to be sure that the description was still fair.

Looks like you made the description more general too, so it matches the link.  And `userTags` itself is generic (not IAM-specific), so I think we're all good.

modules/installation-about-custom.adoc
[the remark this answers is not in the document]
kalexand-rh: @jianlinliu, are you testing any modifications to Kubernetes manifests during installation?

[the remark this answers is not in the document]
jianlinliu: Yeah, we had some testing to modify Kubernetes manifests for customized install until some discussion in upstream repo - https://github.com/openshift/installer/pull/1088/files#r250505048. Since that, we lower the testing of any modifications to Kubernetes manifests, and any issue found during testing is not treated as a blocker issue any more. Because QE also thought modification via manifest and ignition is not officially supported. 
I agree with you, if we support it, we have to tell user which Kubernetes objects are safe to change, or a better approach is to prompt the modification to install-config level, we treat everything in install-config is officially supported to allow modification. 
@qinpingli pls follow up this discussion, make sure what we tested is on the same page like what we published on our official doc.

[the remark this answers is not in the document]
kalexand-rh: Thank you @jianlinliu! @qinpingli, please check during the E2E tests and follow up. Per @vikram-redhat, we're merging frequently instead of holding docs for 2/22.  
@sferich888, are we planning to support modifying Kubernetes manifests during installation at 4.0 GA?

PR #13475, opened 2019-01-30
[enterprise-3.11] took out whole release notes book from dedicated and online

vikram-redhat:


vikram-redhat:
@ahardin-rh @aheslin @wgordon17 

I am working to remove them from customer portal docs as well. 

Both online and dedicated docs are now at 3.11.

PR #13483, opened 2019-01-30
BZ-1528464: Added missing command option

bergerhoffer:
@openshift/team-documentation For peer review. In Atom, I have a setting to strip trailing whitespace (as seen in the diff). Let me know if that's not something you guys do and if I should turn that off.

### DIFF
--- a/install_config/aggregate_logging.adoc
+++ b/install_config/aggregate_logging.adoc
@@ -798,19 +798,19 @@ specified as a python compatible dict. For example, `{"node-type":"infra",
 ==== Persistent Elasticsearch Storage ====
 
 By default, the `openshift_logging` Ansible role creates an ephemeral
-deployment in which all data in a pod is lost upon pod restart. 
+deployment in which all data in a pod is lost upon pod restart.
 
 For production environments, each Elasticsearch deployment configuration requires a persistent storage volume. You can specify an existing xref:../architecture/additional_concepts/storage.adoc#persistent-volume-claims[persistent
-volume claim] or allow {product-title} to create one. 
+volume claim] or allow {product-title} to create one.
 
-* *Use existing PVCs.* If you create your own PVCs for the deployment, {product-title} uses those PVCs. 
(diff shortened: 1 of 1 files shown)


bergerhoffer:
@vikram-redhat This issue only affects 3.5, 3.6, and 3.7. I checked this out of enterprise-3.7 and was planning to cherry pick to enterprise-3.6 and enterprise-3.5 once merged. Let me know if this is the right way to do it or not.

vikram-redhat:
@bergerhoffer yes, you should be able to CP to the older branches since this small change is unlikely to cause any merge conflicts.

I am not sure about the trailing space issue. We don't have a standard for that, AFAIK, but @adellape @kalexand-rh might be able to provide more info.

kalexand-rh:
Oh, funny, @sheriff-rh and I were just talking about this. @vikram-redhat, we don't have a standard for dealing with whitespace trailing.

To some, trimming it is a coding best practice because it can introduce syntax errors in some languages, and it's often a best practice to standardize on trimming or not trimming to improve the git blame/commit history. (If not all the contributors deal with whitespace in the same way, git blame is less useful - no one cares if the last change to a line is to trim whitespace.)

I've noticed that PRs from devs often trim whitespace, and while I don't have a strong opinion on it, I think that it would be appropriate to add "trim your trailing whitespace" to the contributors guide. We're close enough to the start of the 4.0 stream to clean up most of it now, and maintaining a useful blame in the 3.x stream will become steadily less important with time.

kalexand-rh (commented):
LGTM

bergerhoffer:
Hi @anpingli, could you please take a look?

anpingli (approved):
LGTM

bergerhoffer:
Merging and cherry picking as peer review and QE review are complete.

bergerhoffer:
Will cherry pick to 3.5 and 3.6 manually since I discovered that the /cherrypick command only works from the master branch.

PR #13489, opened 2019-01-31
CNV 1.4 Web UI update for VM and Templates

aburdenthehand:
16 Procedures and 6 Reference (UI field tables) that covers the 1.4 UI update for CRUD VMs (including underlying Events, Consoles, NICs and Disks) and CRUD VM Templates.

BZ#1666251

(Still need to do an overview concept and reorganisation of the guide re: cli & web ui, but need to see 1.4 operator content first.)

### DIFF
--- a/modules/cnv_creating_template_wizard_web.adoc
+++ b/modules/cnv_creating_template_wizard_web.adoc
@@ -1,0 +1,35 @@
+// Module included in the following assemblies:
+//
+// * cnv_users_guide/cnv_users_guide.adoc
+
+[[cnv-creating-template-wizard-web]]
+== Creating a virtual machine template with the interactive wizard
+
+Virtual machines templates are an easy way to create multiple virtual machines with similar configuration. After a template is created, reference the template when xref:cnv-creating-vm-wizard-web[creating virtual machines].
+
+The web console features an interactive wizard that guides you through xref:cnv-template-wizard-fields-web[*Basic Settings*], xref:cnv-networking-wizard-fields-web[*Networking*], and xref:cnv-storage-wizard-fields-web[*Storage*] screens to simplify the process of creating virtual machine templates. All required fields are marked by a `*`. The wizard prevents you from moving to the next screen until the required fields have been completed.
+
+NICs and storage disks can be created and attached to virtual machines after they have been created. 
+
+.*Bootable Disk*
+
+If either `URL` or `Container` are selected as the *Provision Source* in the xref:cnv-template-wizard-fields-web[*Basic Settings*] screen, a `rootdisk` disk is created and attached to virtual machines as the *Bootable Disk*. You can modify the `rootdisk` but you cannot remove it. 
+
+A *Bootable Disk* is not required for virtual machines provisioned from a *PXE* source if there are no disks attached to the virtual machine. If one or more disks are attached to the virtual machine, you must select one as the *Bootable Disk*. 
+
+.Procedure
+
+. Click *Workloads* -> *Virtual Machine Templates* from the side menu.
+. Click *Create Template* and select *Create with Wizard*. 
+. Fill in all required xref:cnv-template-wizard-fields-web[*Basic Settings*]. 
+. Click *Next* to progress to the xref:cnv-networking-wizard-fields-web[*Networking*] screen. An `nic0` NIC is attached by default. 
+.. (Optional) Click *Create NIC* to create additional NICs. 
+.. (Optional) You can remove any or all NICs by clicking the &#8942; button and selecting *Remove NIC*. Virtual machines created from a template do not need a NIC attached. NICs can be xref:cnv-vm-create-nic-web[created] after a virtual machine has been created. 
+. Click *Next* to progress to the xref:cnv-storage-wizard-fields-web[*Storage*] screen.
+.. (Optional) Click *Create Disk* to create additional disks. These disks can be removed by clicking the &#8942; button and selecting *Remove Disk*. 
+.. (Optional) Click on a disk to modify available fields. Click the &#10003; button to save the update.
+.. (Optional) Click *Attach Disk* to choose an available disk from the *Select Storage* drop-down list.
+. Click *Create Virtual Machine Template >*. The *Results* screen displays the JSON configuration file for the virtual machine template. 
+
+The template is listed in *Workloads* -> *Virtual Machine Templates*. 
+
--- a/modules/cnv_creating_vm_wizard_web.adoc
+++ b/modules/cnv_creating_vm_wizard_web.adoc
@@ -1,0 +1,34 @@
+
+// Module included in the following assemblies:
+//
+// * cnv_users_guide/cnv_users_guide.adoc
+
+[[cnv-creating-vm-wizard-web]]
+== Creating a virtual machine with the interactive wizard
+
+The web console features an interactive wizard that guides you through xref:cnv-vm-wizard-fields-web[*Basic Settings*], xref:cnv-networking-wizard-fields-web[*Networking*], and xref:cnv-storage-wizard-fields-web[*Storage*] screens to simplify the process of creating virtual machines. All required fields are marked by a `*`. The wizard prevents you from moving to the next screen until the required fields have been completed.
+
+NICs and storage disks can be created and attached to virtual machines after they have been created. 
+
+.*Bootable Disk*
+
+If either `URL` or `Container` are selected as the *Provision Source* in the xref:cnv-vm-wizard-fields-web[*Basic Settings*] screen, a `rootdisk` disk is created and attached to the virtual machine as the *Bootable Disk*. You can modify the `rootdisk` but you cannot remove it. 
+
+A *Bootable Disk* is not required for virtual machines provisioned from a *PXE* source if there are no disks attached to the virtual machine. If one or more disks are attached to the virtual machine, you must select one as the *Bootable Disk*. 
+
+.Procedure
+
+. Click *Workloads* -> *Virtual Machines* from the side menu.
+. Click *Create Virtual Machine* and select *Create with Wizard*. 
+. Fill in all required xref:cnv-vm-wizard-fields-web[*Basic Settings*]. Selecting a *Template* automatically fills in these fields. 
+. Click *Next* to progress to the xref:cnv-networking-wizard-fields-web[*Networking*] screen. A `nic0` NIC is attached by default. 
+.. (Optional) Click *Create NIC* to create additional NICs. 
+.. (Optional) You can remove any or all NICs by clicking the &#8942; button and selecting *Remove NIC*. A virtual machine does not need a NIC attached to be created. NICs can be xref:cnv-vm-create-nic-web[created] after the virtual machine has been created. 
+. Click *Next* to progress to the xref:cnv-storage-wizard-fields-web[*Storage*] screen. 
+.. (Optional) Click *Create Disk* to create additional disks. These disks can be removed by clicking the &#8942; button and selecting *Remove Disk*.
+.. (Optional) Click on a disk to modify available fields. Click the &#10003; button to save the update.
+.. (Optional) Click *Attach Disk* to choose an available disk from the *Select Storage* drop-down list.
+. Click *Create Virtual Machine >*. The *Results* screen displays the JSON configuration file for the virtual machine. 
+
+The virtual machine is listed in *Workloads* -> *Virtual Machines*. 
+
--- a/modules/cnv_creating_vm_yaml_web.adoc
+++ b/modules/cnv_creating_vm_yaml_web.adoc
@@ -1,0 +1,27 @@
+// Module included in the following assemblies:
+//
+// * cnv_users_guide/cnv_users_guide.adoc
+
+[[cnv-creating-vm-yaml-web]]
+== Creating a virtual machine using a YAML configuration file
+
+Create a virtual machine by writing or pasting a YAML configuration file in the web console in the *Workloads* -> *Virtual Machines* screen. A valid `example` virtual machine configuration is provided by default whenever you open the YAML edit screen. 
+
+If your YAML configuration is invalid when you click *Create*, an error message indicates the parameter in which the error occurs. Only one error is shown at a time. 
+
+[NOTE]
+====
+Navigating away from the YAML screen while editing cancels any changes to the configuration you have made. 
+====
+
+.Procedure
+
+. Click *Workloads* -> *Virtual Machines* from the side menu.
+. Click *Create Virtual Machine* and select *Create from YAML*.
+. Write or paste your virtual machine configuration in the editable window.
+.. Alternatively, use the `example` virtual machine provided by default in the YAML screen.
+. (Optional) Click *Download* to download the YAML configuration file in its present state.
+. Click *Create* to create the virtual machine.
+
+The virtual machine is listed in *Workloads* -> *Virtual Machines*. 
+
--- a/modules/cnv_delete_template_web.adoc
+++ b/modules/cnv_delete_template_web.adoc
@@ -1,0 +1,18 @@
+// Module included in the following assemblies:
+//
+// * cnv_users_guide/cnv_users_guide.adoc
+
+[[cnv-delete-template-web]]
+== Deleting a virtual machine template
+
+Deleting a virtual machine template permanently removes it from the cluster. 
+
+Delete a virtual machine template using the &#8942; button of the template in the *Workloads* -> *Virtual Machines Templates* list, or using the *Actions* control of the *Virtual Machine Templates Details* screen.
+
+.Procedure
+
+. Click *Workloads* -> *Virtual Machine Templates* from the side menu.
+. Click the &#8942; button of the template to delete and select *Delete Template*.
+.. Alternatively, click the template name to open the *Virtual Machine Template Details* screen and click *Actions* -> *Delete Template*. 
+. In the confirmation pop-up window, click *Delete* to permanently delete the template.
+
--- a/modules/cnv_editing_vm_web.adoc
+++ b/modules/cnv_editing_vm_web.adoc
@@ -1,0 +1,21 @@
+// Module included in the following assemblies:
+//
+// * cnv_users_guide/cnv_users_guide.adoc
+
+[[cnv-editing-vm-web]]
+== Editing a virtual machine 
+
+You can edit some values of a virtual machine in the web console, either by editing the xref:cnv-editing-vm-yaml-web[YAML directly], or from the *Virtual Machine Overview* screen.
+
+When editing from the *Virtual Machine Overview* screen, the virtual machine must be *Off*.
+
+.Procedure
+
+. Click *Workloads* -> *Virtual Machines* from the side menu.
+. Select a Virtual Machine.
+. Click *Edit* to make editable fields available. 
+. You can change the *Flavor* but only to *Custom*, which provides additional fields for *CPU* and *Memory*. 
+. Click *Save*. 
+
+The updated values are shown after the operation is processed. 
+
--- a/modules/cnv_networking_wizard_fields_web.adoc
+++ b/modules/cnv_networking_wizard_fields_web.adoc
@@ -1,0 +1,26 @@
+// Module included in the following assemblies:
+//
+// * cnv_users_guide/cnv_users_guide.adoc
+
+[[cnv-networking-wizard-fields-web]]
+== Networking fields
+|===
+|Name | Description
+
+|Create NIC
+|Create a new NIC for the virtual machine.
+
+|NIC NAME
+|Name for the NIC.
+
+|MAC ADDRESS
+|MAC address for the network interface. If a MAC address is not specified, an ephemeral address is generated for the session. 
+
+|NETWORK CONFIGURATION
+|List of available *NetworkAttachmentDefinition* objects.
+
+|PXE NIC
+|List of PXE-capable networks. Only visible if `PXE` has been selected as the `Provision Source`.
+|===
+
+
--- a/modules/cnv_storage_wizard_fields_web.adoc
+++ b/modules/cnv_storage_wizard_fields_web.adoc
@@ -1,0 +1,28 @@
+// Module included in the following assemblies:
+//cnv_storage_wizard_fields_web.adoc
+// * cnv_users_guide/cnv_users_guide.adoc
+
+[[cnv-storage-wizard-fields-web]]
+== Storage fields
+|===
+|Name | Description
+
+|Create Disk
+|Create a new disk for the virtual machine.
+
+|Attach Disk
+|Select an existing disk, from a list of available PVCs, to attach to the virtual machine. 
+
+|DISK NAME
+|Name of the disk.
+
+|SIZE (GB)
+|Size, in GB, of the disk. 
+
+|STORAGE CLASS
+|Name of the underlying `StorageClass`. 
+
+|Bootable Disk
+|List of available disks from which the virtual machine will boot. This is locked to `rootdisk` if the *Provision Source* of the virtual machine is `URL` or `Container`.
+|=== 
+
--- a/modules/cnv_template_wizard_fields_web.adoc
+++ b/modules/cnv_template_wizard_fields_web.adoc
@@ -1,0 +1,49 @@
+// Module included in the following assemblies:
+//
+// * cnv_users_guide/cnv_users_guide.adoc
+
+[[cnv-template-wizard-fields-web]]
+== Virtual machine template wizard fields
+
+|===
+|Name |Parameter |Description
+
+|Name
+|
+|Name of the virtual machine template. Alphanumeric characters only, up to a maximum of 63. 
+
+|Description
+|
+|Optional description field.
+
+.3+|Provision Source
+|PXE
+|Provision virtual machines from PXE menu. Requires a xref:pxebooting[PXE-capable NIC] in the cluster.
+
+|URL
+|Provision virtual machines from an image available from a *HTTP* or *S3* endpoint. 
+
+|Container
+|Provision virtual machines from a bootable operating system container located in a registry accessible from the cluster. Example: `_kubevirt/cirros-registry-disk-demo_`
+
+|Operating System
+|
+|A list of operating systems available in the cluster. This is the primary operating system for the virtual machine.
+
+|Flavor
+|small, medium, large, tiny, Custom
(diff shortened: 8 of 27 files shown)


modules/cnv_storage_wizard_fields_web.adoc
```diff
@@ -0,0 +1,28 @@
...
+[[cnv-storage-wizard-fields-web]]
+== Storage fields
+|===
+|Name | Description
+
+|Create Disk
+|Create a new disk for the virtual machine.
+
+|Attach Disk
+|Select an existing disk to attach to the virtual machine.
```
jelkosz: Maybe worth mentioning that it lists available PVCs.

modules/cnv_storage_wizard_fields_web.adoc
```diff
@@ -0,0 +1,28 @@
...
+
+|DISK NAME
+|Name of the disk.
+
+|SIZE (GB)
+|Size, in GB, of the disk. 
+
+|STORAGE CLASS
+|`StorageClass` name for the underlying PVC.
```
jelkosz: actually it is really the name of the underlying storage class. Storage class defines where the storage will be living + some characteristics of the storage.

modules/cnv_template_wizard_fields_web.adoc
```diff
@@ -0,0 +1,53 @@
...
+|Provision virtual machines from PXE menu. Requires a PXE-capable NIC in the cluster.
+
+|URL
+|Provision virtual machines from an image available from a *HTTP* or *S3* endpoint. 
+
+|Container
+|Provision virtual machines from a bootable operating system container located in a registry accessible from the cluster. Example: `_kubevirt/cirros-registry-disk-demo_`
+
+|Operating System
+|
```
jelkosz: not sure if this newline here is correct or not...

modules/cnv_template_wizard_fields_web.adoc
```diff
@@ -0,0 +1,53 @@
...
+|A list of operating systems available in the cluster. This is the primary operating system for the virtual machine.
+
+|Flavor
+|small, medium, large, tiny, Custom
+|Presets that determine the amount of CPU and memory allocated to the virtual machines. 
+
+.2+|Workload Profile
```
jelkosz: @MarSik not sure it is a good idea to explain it like this. WDYT?

modules/cnv_template_wizard_fields_web.adoc
```diff
@@ -0,0 +1,53 @@
...
+|generic
+|`iothreads` disabled.
+
+|highperformance
+|`iothreads` are enabled.
+
+|cloud-init
+|
```
jelkosz: again a suspicious enter but not sure if correct or not.

modules/cnv_vm_actions_web.adoc
```diff
@@ -0,0 +1,33 @@
...
+
+|*Stop Virtual Machine*
+|_Running_ or _Other_
+|Stop the virtual machine.
+
+|*Restart Virtual Machine*
+|_Running_ or _Other_
+|Restart the running virtual machine.
+
+|*Migrate Virtual Machine*
```
jelkosz: this needs to be removed for 1.4

modules/cnv_vm_console_web.adoc
```diff
@@ -0,0 +1,16 @@
...
+== Connecting to a virtual machine console
+
+You can connect to the VNC and serial consoles of a running virtual machine from the *Virtual Machine Details* screen of the web console.
+
+.Procedure
+
+. Click *Workloads* -> *Virtual Machines*  from the side menu.
+. Select a Virtual Machine.
+. Click *Consoles*. The VNC console opens by default.
+. Use the *Console* drop-down list to change the console type. Unless the *Disconnect before switching* checkbox is active, you must click *Connect* to open a new console.
```
jelkosz: you don't need to click "connect". The "disconnect before switching" means that when you switch between consoles, the previous always disconnects. E.g.:
1: the "disconnect before switching" is not checked
2: Im on the VNC console now
3: in the listbox switch to serial
4: the VNC console connection stays open and an additional connection will be opened to the serial console. You will be presented in the UI only the serial but on the background the connection to the VNC will be preserved
5: if you now switch back to the VNC, it will be very fast because no reconnect is needed. The disadvantage is that you have more opened connections.

If you have the checkbox checked, the consoles will be connected and disconnected every time you switch. As a result you will have less connections opened but every switch will be slower. But from user PoV not a big difference.

modules/cnv_vm_wizard_fields_web.adoc
```diff
@@ -0,0 +1,57 @@
...
+|Optional description field.
+
+|Template
+|
+|Template from which to create the virtual machine. Selecting a template will automatically complete other fields. 
+
+
+.3+|Provision Source
+|PXE
+|Provision virtual machine from PXE menu. Requires a PXE-capable NIC in the cluster.
```
jelkosz: Maybe worth mentioning that PXE capable nic actually means a normal nick but is associated to a network attachment definition which is backed by an L2 network. e.g. you need to have ovs-cni + multus set up and configured. Im sure it is documented somewhere else so you can link to that part from here.

modules/cnv_vm_wizard_fields_web.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+|URL
+|Provision virtual machine from an image available from a *HTTP* or *S3* endpoint. 
+
+|Container
+|Provision virtual machine from a bootable operating system container located in a registry accessible from the cluster. Example: `_kubevirt/cirros-registry-disk-demo_`
+
+|Operating System
+|
```
jelkosz: again the suspicious enter :)

modules/cnv_vm_wizard_fields_web.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+.2+|Workload Profile
+|generic
+|`iothreads` disabled.
+
+|highperformance
+|`iothreads` are enabled.
+
+|Start virtual machine on creation
+|
```
jelkosz: enter

modules/cnv_vm_wizard_fields_web.adoc
```diff
@@ -0,0 +1,57 @@
...
+|Select this checkbox to automatically start the virtual machine upon creation. 
+
+|cloud-init
+|
```
jelkosz: enter

modules/cnv_template_wizard_fields_web.adoc
MarSik: Yeah, I would be a bit more verbose with less emphasis on details. The highperformance workload variant will do more than just iothreads in the future (once we have the api for stuff). For example see oVirt - the HP option there disables USB, graphical devices, sound cards, enables iothreads and multiqueues..

aburdenthehand: Thanks for the feedback. Have updated it to:
generic: Virtual machines have a general configuration appropriate for standard usage.
highperformance: Virtual machines have a more efficient configuration optimized for high performance loads.

modules/cnv_migrate_vm_web.adoc
```diff
@@ -0,0 +1,18 @@
+// Module included in the following assemblies:
+//
+// * cnv_users_guide/cnv_users_guide.adoc
+
+[[cnv-migrate-vm-web]]
+== Migrating a virtual machine
```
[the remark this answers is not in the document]
aburdenthehand: I've dropped it from the assembly so it won't be published in 1.4, but the migrate_vm file is still included in the MR so that the content doesn't get lost. Unless the migration procedure is likely to change for 2.0, in which case I will delete the module completely.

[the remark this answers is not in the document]
jelkosz: It should stay the same pretty much so it is fine to keep it here if the user will not see it.

modules/cnv_networking_wizard_fields_web.adoc
```diff
@@ -0,0 +1,26 @@
...
+|Name | Description
+
+|Create NIC
+|Create a new NIC for the virtual machine.
+
+|NIC NAME
+|Name for the NIC.
+
+|MAC ADDRESS
+|MAC address for the network interface. If a MAC address is not specified, an ephemeral address is generated for the session.
```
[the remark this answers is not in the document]
jelkosz: @aburdenthehand was talking to Petr offline and he confirmed that this part is correct.

jelkosz:
Thank you @aburdenthehand it looks good now!

alexxa (changes requested):
@aburdenthehand 
I'm afraid to lose comments on some discovered issues, so I'm submitting this Review, but please mind I have not completed it yet. I will need a day or two next week, please.

modules/cnv_template_wizard_fields_web.adoc
MarSik: That would be too vague for me..

I am not a tech writer but I would define generic as good compromise between performance and compatibility for the broadest set of usages. HP can then say that some features and compatibility options were sacrificed in favor of better performance of cpu intensive workloads.

aburdenthehand: @MarSik Appreciate it:
generic: A general configuration that balances performance and compatibility for a broad range of workloads.
highperformance: Some features and compatibility options are sacrificed for better performance of CPU-intensive workloads.
However, once we say 'some features and options', it would be good to direct users to something that goes into detail to what those features and options might be. Do we have a resource like that (even if it's an upstream page)? I gather this is likely to change quickly, but still worth including

modules/cnv_migrate_vm_web.adoc
```diff
@@ -0,0 +1,18 @@
+// Module included in the following assemblies:
+//
+// None for CNV 1.4; topic saved for 2.0
+
+[[cnv-migrate-vm-web]]
+== Migrating a virtual machine
+
+Migrate a running virtual machine to a different node in the cluster. 
+
+Migrate a virtual machine using the &#8942; button of the template in the *Workloads* -> *Virtual Machines* list, or using the xref:cnv_vm_actions_web.adoc#cnv-vm-actions-web[*Actions* control] of the *Virtual Machine Details* screen.
```
alexxa: * There is no "Migrate ..." in xref:cnv_vm_actions_web.adoc#cnv-vm-actions-web[*Actions* control] documented
* There is no "Migrate... " available on the &#8942; button. 

@jelkosz is Migration described in this file possible at all?

aburdenthehand: @alexxa Sorry, I resolved the conversation Tomas and I had about this file: Migrate is not possible, but it was when I wrote this up. I've taken this file out of the assembly so it won't be published. It seems as though the procedure will be the same so it made sense to include the file in the MR for when the feature becomes available.

modules/cnv_storage_wizard_fields_web.adoc
```diff
@@ -0,0 +1,28 @@
...
+|Name of the disk.
+
+|SIZE (GB)
+|Size, in GB, of the disk. 
+
+|STORAGE CLASS
+|Name of the underlying `StorageClass`. 
+
+|Bootable Disk
+|List of available disks from which the virtual machine will boot.
```
alexxa: @jelkosz Please confirm we need to be more specific in [1, 2, 3, 4] or elsewhere. That rootdisk is not only default bootable disk for Container and URL sources, but it's only available option which can't be changed or removed. For PXE one can change disks, but there is always must be one (bootable).

In other words default doesn't mean "(un)changeable". 
And rootdisk is not the same as a bootable disk. [3, 4].

@aburdenthehand  Also I would move some wording of lines [1, 2] and a note "One can't remove a rootdisk" into xref:cnv_storage_wizard_fields_web.adoc#cnv-storage-wizard-fields-web[*Storage*]   

[1] https://github.com/openshift/openshift-docs/pull/13489/files#diff-9ea9eed2e360d410a865899f28c1925bR22
[2] https://github.com/openshift/openshift-docs/pull/13489/files#diff-5b75dac4831be9f2ca87957b91d4aa0aR21
[3] https://github.com/openshift/openshift-docs/pull/13489/files#diff-9ea9eed2e360d410a865899f28c1925bR23
[4] https://github.com/openshift/openshift-docs/pull/13489/files#diff-5b75dac4831be9f2ca87957b91d4aa0aR22

modules/cnv_template_wizard_fields_web.adoc
```diff
@@ -0,0 +1,53 @@
...
+
+[[cnv-template-wizard-fields-web]]
+== virtual machine template wizard fields
+
+|===
+|Name |Parameter |Description
+
+|Name
+|
+|Name of the virtual machine template.
```
alexxa: I would mention what chars can be used in VM naming.

modules/cnv_template_wizard_fields_web.adoc
```diff
@@ -0,0 +1,53 @@
...
+
+|Description
+|
+|Optional description field.
+
+|Namespace
```
alexxa: I don't see Namespace field.

modules/cnv_migrate_vm_web.adoc
alexxa: I would remove it from this MR, and keep it elsewhere then.
Also, [1] should be updated for Migrating feature accordingly.

[1] https://github.com/openshift/openshift-docs/blob/13ccc6480c4fc90aaeb207d19dc5dc657998914d/modules/cnv_vm_actions_web.adoc

modules/cnv_storage_wizard_fields_web.adoc
jelkosz: > That rootdisk is not only default bootable disk for Container and URL sources, but it's only available option which can't be changed or removed. 

indeed, this is the case

> For PXE one can change disks, but there is always must be one (bootable).

For PXE you don't need to have any disks. But if you do, you have to pick one which is the bootable one.

>

modules/cnv_vm_console_web.adoc
```diff
@@ -0,0 +1,21 @@
...
+====
+When the *Disconnect before switching* checkbox is active, the console session is disconnected when you switch between consoles so that only one session is open at a time. Otherwise, the console sessions remain active in the background. 
+====
+
+.Procedure
+
+. Click *Workloads* -> *Virtual Machines*  from the side menu.
+. Select a Virtual Machine.
+. Click *Consoles*. The VNC console opens by default.
+. Use the *Console* drop-down list to change the console type.
```
alexxa: Hm, there is no "Console" drop-down list. "VNC Console" and "Serial Console" are there.

modules/cnv_vm_console_web.adoc
```diff
@@ -0,0 +1,21 @@
...
+// * cnv_users_guide/cnv_users_guide.adoc
+
+[[cnv-vm-console-web]]
+== Connecting to a virtual machine console
+
+You can connect to the VNC and serial consoles of a running virtual machine from the *Virtual Machine Details* screen of the web console.
+
+[NOTE]
+====
+When the *Disconnect before switching* checkbox is active, the console session is disconnected when you switch between consoles so that only one session is open at a time. Otherwise, the console sessions remain active in the background.
```
alexxa: I would move this note lower, in Procedure steps or be more specific.. Because it's not clear where "Disconnect before switching" checkbox is.

modules/cnv_editing_vm_web.adoc
```diff
@@ -0,0 +1,21 @@
...
+You can edit some values of a virtual machine in the web console, either by editing the xref:cnv_editing_vm_yaml_web.adoc#cnv-editing-vm-yaml-web[YAML directly], or from the *Virtual Machine Overview* screen.
+
+When editing from the *Virtual Machine Overview* screen, the virtual machine must be *Off*.
+
+.Procedure
+
+. Click *Workloads* -> *Virtual Machines*  from the side menu.
+. Select a Virtual Machine.
+. Click *Edit* to make editable fields available. 
+. Changing the virtual machine *Flavor* to *Custom* provides additional fields for *CPU* and *Memory*.
```
jelkosz: It would be good to mention here that you can pick only one of the "Custom" flavor or the flavor the VM has been created with. No other flavors are available on edit.
@alexxa FYI

modules/cnv_vm_delete_nic_web.adoc
```diff
@@ -0,0 +1,17 @@
...
+
+Deleting a NIC from a virtual machine detaches and permanently deletes the NIC. 
+
+.Procedure
+
+. Click *Workloads* -> *Virtual Machines* from the side menu.
+. Select a Virtual Machine.
+. Click *Network Interfaces* to display the NICs already attached to the virtual machine.
+. Click the &#8942; button of the NIC you wish to delete and select *Delete*. 
+. Click *Confirm* to detach and delete the NIC.
```
alexxa: Click *Delete*, not *Confirm*.

modules/cnv_vm_events_web.adoc
```diff
@@ -0,0 +1,15 @@
...
+== Viewing the events of a virtual machine
+
+You can view the events for a running a virtual machine from the *Virtual Machine Details* screen of the web console.
+
+.Procedure
+
+. Click *Workloads* -> *Virtual Machines*  from the side menu.
+. Select a Virtual Machine.
+. Click *Events* to view all events for the virtual machine. 
+
```
alexxa: Maybe add a note about Streaming Events and Pausing?

modules/cnv_control_vm_web.adoc
```diff
@@ -0,0 +1,18 @@
+// Module included in the following assemblies:
+//
+// * cnv_users_guide/cnv_users_guide.adoc
+
+[[cnv-control-vm-web]]
+== starting, stopping, and restarting a virtual machine
```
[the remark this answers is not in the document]
aburdenthehand: I wanted to put the specific actions in the title so that it was immediately obvious to anyone scanning the contents page. This guide has grown quite large now and I was worried that it would be buried unless made explicit.

modules/cnv_control_vm_web.adoc
```diff
@@ -0,0 +1,18 @@
...
+
+Depending on the state, virtual machines can be xref:cnv_vm_actions_web.adoc#cnv-vm-actions-web[started, stopped, and restarted] using the &#8942; button of each virtual machine in the *Workloads* -> *Virtual Machines* list.
```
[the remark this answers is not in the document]
aburdenthehand: Same reason as above. It's moot now, though, since the 'Confirm' button for each of the confirmation pop-up windows is now specific to the actions. I'm going to split this into three modules: Starting a VM, Stopping a VM, and Restarting a VM

aburdenthehand:
@openshift/team-documentation - please review.
It's on_qa but has gone through a round of QE feedback.

vikram-redhat:

alexxa:
@aburdenthehand There are 4 comments from me which were not resolved. Will you add changes accordingly, please?

alexxa:
@aburdenthehand Also, within this PR, can you update two first subsections of Reference section below, please? Then, they all will start with a capital letter. 

Reference

- virtual machine wizard fields
- virtual machine template wizard fields
- Cloud-init fields
- Networking fields
- Storage fields
- Virtual machine actions
- Types of storage volumes for virtual machines
- Template: PVC configuration file
- Template: VM configuration file
- Template: VM configuration file (DataVolume)
- Template: DataVolume import configuration file
- Template: DataVolume clone configuration file
- Template: VMI configuration file for PXE booting

modules/cnv_creating_vm_yaml_web.adoc
```diff
@@ -0,0 +1,26 @@
...
+[NOTE]
+====
+Navigating away from the YAML screen while editing cancels any changes to the configuration you have made. 
+====
+
+.Procedure
+
+. Click *Workloads* -> *Virtual Machines*  from the side menu.
+. Click *Create Virtual Machine* and select *Create from YAML*.
+. Write or paste your virtual machine configuration in the editable window.
```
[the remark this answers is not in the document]
aburdenthehand: If you mean the `example` VM when you open the YAML editor, then it's covered in the following substep:
".. Alternatively, use the `example` virtual machine provided by default in the YAML screen."

modules/cnv_delete_template_web.adoc
```diff
@@ -0,0 +1,18 @@
...
+Deleting a virtual machine template permanently removes it from the cluster. 
+
+Delete a virtual machine template using the &#8942; button of the template in the *Workloads* -> *Virtual Machines Templates* list, or using the xref:cnv_vm_actions_web.adoc#cnv-vm-actions-web[*Actions* control] of the *Virtual Machine Templates Details* screen.
+
+.Procedure
+
+. Click *Workloads* -> *Virtual Machine Templates* from the side menu.
+. Click the &#8942; button of the template to delete and select *Delete Template*.
+.. Alternatively, click the template name to open the *Virtual Machine Template Details* screen and click *Actions* -> *Delete Template*. 
+. Click *Confirm* to permanently delete the template.
```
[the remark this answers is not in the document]
aburdenthehand: Thanks, I missed that one.

modules/cnv_vm_events_web.adoc
aburdenthehand: Added lines about pausing and streaming before the procedure.

aburdenthehand:
@alexxa - Done and done. Thanks!

modules/cnv_creating_vm_yaml_web.adoc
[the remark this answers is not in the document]
alexxa: Right, missed it.

alexxa:
@aburdenthehand Looks good to me now. Thanks!

alexxa (approved):

modules/cnv_template_wizard_fields_web.adoc
sheriff-rh: @aburdenthehand  hi, tech writer here. 

I've discussed this with my team, and your suggested `generic` description is great. I suggest also not changing the `highperformance` description. We tend to avoid suggesting sacrifices without in-depth descriptions. 

Everything else looks great in this PR! Will mark peer review done.

sheriff-rh:
Marking as peer review done. As mentioned in the conversation on `modules/cnv_template_wizard_fields_web.adoc` , I recommend changing the `generic` line to your suggestion: 

```
|generic
|A general configuration that balances performance and compatibility for a broad range of workloads.
```

Thanks!

modules/cnv_creating_template_wizard_web.adoc
```diff
@@ -0,0 +1,34 @@
...
+
+Virtual machine templates are an easy way to create multiple virtual machines with similar configuration. After a template is created, reference the template when xref:cnv-creating-vm-wizard-web[creating virtual machines].
+
+The web console features an interactive wizard that guides you through xref:cnv-template-wizard-fields-web[*Basic Settings*], xref:cnv-networking-wizard-fields-web[*Networking*], and xref:cnv-storage-wizard-fields-web[*Storage*] screens to simplify the process of creating virtual machine templates. All required fields are marked by a `*`. The wizard prevents you from moving to the next screen until the required fields have been completed.
+
+NICs and storage disks can be created and attached to virtual machines after they have been created. 
+
+.*Bootable Disk*
+
+If either `URL` or `Container` are selected as the *Provision Source* in the xref:cnv-template-wizard-fields-web[*Basic Settings*] screen, a `rootdisk` disk is created and attached to virtual machines as the *Bootable Disk*. You cannot remove or modify the `rootdisk`.
```
alexxa: @aburdenthehand Hm.. This wording is not correct in this case. One can't changes the Bootable disk, which is rootdisk to something else. But the rootdisk itself can be modified, if the provision source is URL. One can change its size and the Storage class. For Container source, one can modify only a name of the rootdisk, and Bootable disk will refer accordingly.

modules/cnv_creating_vm_wizard_web.adoc
```diff
@@ -0,0 +1,33 @@
...
+[[cnv-creating-vm-wizard-web]]
+== Creating a virtual machine with the interactive wizard
+
+The web console features an interactive wizard that guides you through xref:cnv-vm-wizard-fields-web[*Basic Settings*], xref:cnv-networking-wizard-fields-web[*Networking*], and xref:cnv-storage-wizard-fields-web[*Storage*] screens to simplify the process of creating virtual machines. All required fields are marked by a `*`. The wizard prevents you from moving to the next screen until the required fields have been completed.
+
+NICs and storage disks can be created and attached to virtual machines after they have been created. 
+
+.*Bootable Disk*
+
+If either `URL` or `Container` is selected as the *Provision Source* in the xref:cnv-vm-wizard-fields-web[*Basic Settings*] screen, a `rootdisk` disk is created and attached to the virtual machine as the *Bootable Disk*. You cannot remove or modify the `rootdisk`.
```
alexxa: @aburdenthehand the same here.

alexxa (changes requested):
Sorry, I have to revert my Verification. I have missed one thing. Requested changes are added earlier.

modules/cnv_creating_template_wizard_web.adoc
aburdenthehand: Updated to "You can modify the `rootdisk` but you cannot remove it. "
also added substep to the Storage step in procedure: ".. (Optional) Click on a disk to modify available fields. Click the &#10003; button to save the update."

We could specifically call out the difference in what can be updated between container- and URL-based rootdisks but it makes this para convoluted for what should be straightforward to anyone using the wizard.

modules/cnv_template_wizard_fields_web.adoc
aburdenthehand: Cool beans, updated as suggested. Thank you!

modules/cnv_vm_create_nic_web.adoc
```diff
@@ -0,0 +1,19 @@
...
+
+[[cnv-vm-create-nic-web]]
+== Creating a NIC for a virtual machine
+
+Create and attach additional NICs to a virtual machine from the web console.
+
+.Procedure
+
+. Click *Workloads* -> *Virtual Machines* from the side menu.
+. Select a Virtual Machine.template
```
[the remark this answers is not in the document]
aburdenthehand: Gosh! Thanks for the spot :)
On that, it should be lower-case 'virtual machine', and I notice that's permeated a couple of other procedures so will fix them while I'm here.

aburdenthehand:
All feedback incorporated.
Thanks everybody!

vikram-redhat:
Not to be CPed to 3.11. @aburdenthehand will submit a new one with restructured docs on Monday.

PR #13477, opened 2019-01-30
bug 1622044 corrected auditFilePath examples

sheriff-rh:
Link to bug, see c#8:
https://bugzilla.redhat.com/show_bug.cgi?id=1622044

@gpei PTAL 
@openshift/team-documentation peer review please!

Thanks!

PR #13492, opened 2019-01-31
[enterprise-3.11] bug 1622044 corrected auditFilePath examples

sheriff-rh:
Cherrypicked from https://github.com/openshift/openshift-docs/pull/13477 . 

Resolved conflict.

### DIFF
--- a/install_config/master_node_configuration.adoc
+++ b/install_config/master_node_configuration.adoc
@@ -19,7 +19,7 @@ servers in a development or experimental environment.
 
 However, these arguments are insufficient to describe and control the full set
 of configuration and security options that are necessary in a production
-environment. You must provide those options in the xref:../architecture/infrastructure_components/kubernetes_infrastructure.adoc#master[Master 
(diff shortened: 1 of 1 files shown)


install_config/master_node_configuration.adoc
```diff
@@ -1832,6 +1832,36 @@ xref:../install/configuring_inventory_file.adoc#cluster-variables-table[Cluster
 
 The following examples are excerpts of redirected master log files at various log levels. System information has been removed from these examples.
 
+////
```
vikram-redhat: @sheriff-rh I would check that this log file output is correct for the 3.11 version with the QE team (unless you have already done so.) I can't recall why this was not in there, perhaps we never had this output available for 3.11 at the time this topic was migrated to 3.11.

sheriff-rh: Hey Vikram! I checked with @mburke5678  yesterday and we agreed to squash the output comments. My new commit this morning reflects that change. Thanks!

vikram-redhat: @sheriff-rh - I didn't realize this was commented out section :)

PR #13540, opened 2019-02-05
changed topic map to match bucket name

mburke5678:
Changing the topic-map for the _Cluster Logging_ book to _Logging_ to match the buckets.


mburke5678:
@vikram-redhat PTAL

PR #13564, opened 2019-02-07
Add documentation about experimental registry feature

legionus:
@bparees PTAL


admin_guide/registry_data_check_restore.adoc
```diff
@@ -0,0 +1,112 @@
...
+:experimental:
+:toc: macro
+:toc-title:
+
+toc::[]
+
+[IMPORTANT]
+====
+All that is described here is an experimental functionality and may
+change in the future.
```
bparees: "This section describes functionality that is experimental and may change in the future"

admin_guide/registry_data_check_restore.adoc
```diff
@@ -0,0 +1,112 @@
...
+====
+
+== Overview
+
+Sometimes the data stored in the OpenShift Container Registry may differ
+from the data in the {product-title} cluster's etcd. It means that either
+the {product-title} cluster's database contains references on data that
```
bparees: s/on/to/

admin_guide/registry_data_check_restore.adoc
```diff
@@ -0,0 +1,112 @@
...
+does not exist on the OpenShift Container Registry's storage, or the
+OpenShift Container Registry storage has data that is not referenced in
+the {product-title} cluster's database. This situation may occur as a result
+of an error or accidental deletion.
+
+There are two classes of problems:
+
+. The {product-title} cluster's database contains references to
+non-existent blobs. This usually happens due to storage damage. In this
+case, such imagestrems/images are not possible to use and any pull will
```
bparees: s/imagestrems/imagestreams/

admin_guide/registry_data_check_restore.adoc
```diff
@@ -0,0 +1,112 @@
...
+end with error. The way to fix such images is to push them onto the
```
bparees: s/with error/with an error/

admin_guide/registry_data_check_restore.adoc
```diff
@@ -0,0 +1,112 @@
...
+OpenShift Container Registry again to restore the blobs. The problem is
+to understand which imagestreams were damaged.
+
+. Necessary images were removed from the {product-title} cluster's
+database, but remained in the OpenShift Container Registry storage. This
+may occur due to incorrect pruning settings or due to accidental manual
+removal. In this case, the problem is to restore the references in the
+{product-title} cluster's database in order to prevent blobs from being
+deleted during the hard pruning process.
```
bparees: link to hard pruning doc

admin_guide/registry_data_check_restore.adoc
```diff
@@ -0,0 +1,112 @@
...
+
+== Preparation
+
+Before start, you must give the registry's service account permissions to
```
bparees: s/start/starting/

admin_guide/registry_data_check_restore.adoc
```diff
@@ -0,0 +1,112 @@
...
+change the {product-title} cluster's database:
++
+----
+$ service_account=$(oc get -n default \
+    -o jsonpath=$'system:serviceaccount:{.metadata.namespace}:{.spec.template.spec.serviceAccountName}\n' \
+    dc/docker-registry)
+$ oc adm policy add-cluster-role-to-user cluster-admin ${service_account}
```
bparees: there's no more minimal permission we can grant than cluster-admin?

admin_guide/registry_data_check_restore.adoc
```diff
@@ -0,0 +1,112 @@
...
+    exec -i -t "$(oc -n default get pods -l deploymentconfig=docker-registry \
+    -o jsonpath=$'{.items[0].metadata.name}\n')" \
+    -- /usr/bin/dockerregistry -restore-mode=recover
+----
+
+Unfortunately, imagestream tag names are not stored on the storage.
+Therefore, they can not be restored. Also, if the imagestream tag had a
+history, then it also could not be restored. Tag creation time will also
+be lost. All images from the history will be restored as separate tags.
+The recovered tag names will have the form: `lost-found-<DIGEST>`.
```
bparees: this tag is created in whatever namespace/imagestream the content was originally associated with?

admin_guide/registry_data_check_restore.adoc
```diff
@@ -0,0 +1,112 @@
...
+You can also limit the namespace you want to scan:
++
+----
+$ oc -n default \
+    exec -i -t "$(oc -n default get pods -l deploymentconfig=docker-registry \
+    -o jsonpath=$'{.items[0].metadata.name}\n')" \
+    -- /usr/bin/dockerregistry -restore-mode=check -restore-namespace=myproject
+----
+
+== Restore imagesstreams
```
bparees: /imagesstreams/imagestreams/

bparees (changes requested):
up front you define two problems:

1) missing db metadata
2) missing storage content/blobs

The doc addresses how the tool fixes (1) but it other than a brief mention of repushing content for (2) it doesn't go into much detail.  It might be useful to show sample output from the tool for both scenarios along with some guidance about what the admin needs to do based on that output.

legionus:
> The doc addresses how the tool fixes (1) but it other than a brief mention of repushing content for (2) it doesn't go into much detail.

@bparees Yes. Because problem (2) has no solution. Data lost. We don't know where they were being pushed from. We do not know whether they remained there.

>  It might be useful to show sample output from the tool for both scenarios along with some guidance about what the admin needs to do based on that output.

I can add example of output, but I can't add guidance because then comes the manual work of restoring the names or searching from where the image was pushed. These actions are individual and depend on how the cluster was used.

admin_guide/registry_data_check_restore.adoc
legionus: We need to list all imagestreams/images, get images and use imagestreammapping for any imagestream. I do not know what role it is guaranteed to provide this and is on all versions of the cluster. Maybe cluster-reader is enough, but I'm not sure.

bparees:
> @bparees Yes. Because problem (2) has no solution. Data lost. We don't know where they were being pushed from. We do not know whether they remained there.

> I can add example of output, but I can't add guidance because then comes the manual work of restoring the names or searching from where the image was pushed. These actions are individual and depend on how the cluster was used.

Yup, understood, just looking for something like "given this example output which informs you that image X is missing blobs, you will need to locate a copy of image X and push it to the internal registry as X"

admin_guide/registry_data_check_restore.adoc
bparees: ok, this is fine for now.

admin_guide/registry_data_check_restore.adoc
legionus: @bparees If there is no imagestream, then we will not create it again and return error. If the imagestream already has such tag, we will also return an error.

legionus:
OK. I will add something like that.

admin_guide/registry_data_check_restore.adoc
bparees: ok, might be worth describing that behavior here.

legionus:
@bparees please review again

admin_guide/registry_data_check_restore.adoc
```diff
@@ -0,0 +1,176 @@
...
+Broken imagestream tag "myproject/ubuntu:latest" in position 0: image "sha256:be159ff0e12a38fd2208022484bee14412680727ec992680b66cdead1ba76d19": unknown blob
+----
+
+or if blob does not exist:
+
+----
+Broken imagestream tag "myproject/ubuntu:latest" in position 0: image blob "sha256:d42beb8ded595df5627ad4ef31bf528a6fdbfbd11d82f9023152738d6b05a7fa": unknown blob
+----
+
+If you have images in which contains nonexistent blob, but which are not
```
bparees: "If you have images which reference a non-existent blob, but which are not"

admin_guide/registry_data_check_restore.adoc
```diff
@@ -0,0 +1,176 @@
...
+in any imagestream, then the message will look like:
+
+----
+Broken image "sha256:be159ff0e12a38fd2208022484bee14412680727ec992680b66cdead1ba76d19": unknown blob
+----
+
+If the storage has image that are not in the imagestream, then the
```
bparees: "If the storage has an image that is not in an imagestream, then the"

admin_guide/registry_data_check_restore.adoc
```diff
@@ -0,0 +1,176 @@
...
+message will look like this:
+
+----
+Would add image "sha256:be159ff0e12a38fd2208022484bee14412680727ec992680b66cdead1ba76d19" to imagestream "myproject/ubuntu" with tag "lost-found-be159ff0e12a38fd2208022484bee14412680727ec992680b66cdead1ba76d19"
+----
+
+== Restore
+
+Some data can be recovered. Part of the data is not duplicated and
+therefore their recovery is not possible.
```
bparees: "Some data can be recovered, but in some cases the data is lost and must be restored from an external copy."

admin_guide/registry_data_check_restore.adoc
```diff
@@ -0,0 +1,176 @@
...
+
+Unfortunately, imagestream tag names are not stored on the storage.
+Therefore, they can not be restored. Also, if the imagestream tag had a
+history, then it also could not be restored. Tag creation time will also
+be lost. All images from the history will be restored as separate tags.
+The recovered tag names will have the form: `lost-found-<DIGEST>`. 
+
+=== Restore data on storage
+
+The only way to recover data the OpenShift Container Registry's storage
```
bparees: "The only way to restore data that was lost from the {product-name} Container Registry's storage"

admin_guide/registry_data_check_restore.adoc
```diff
@@ -0,0 +1,176 @@
...
+is to re-push them.
```
bparees: "is to re-push the content."

admin_guide/registry_data_check_restore.adoc
```diff
@@ -0,0 +1,176 @@
...
+
+----
+Broken imagestream tag "myproject/ubuntu:latest" in position 0: image "sha256:be159ff0e12a38fd2208022484bee14412680727ec992680b66cdead1ba76d19": unknown blob
+Broken imagestream tag "myproject/ubuntu:latest" in position 0: image blob "sha256:d42beb8ded595df5627ad4ef31bf528a6fdbfbd11d82f9023152738d6b05a7fa": unknown blob
+Broken image "sha256:be159ff0e12a38fd2208022484bee14412680727ec992680b66cdead1ba76d19": unknown blob
+----
+
+you will need to locate a copy of image
+`sha256:be159ff0e12a38fd2208022484bee14412680727ec992680b66cdead1ba76d19`
+and push it to the OpenShift Container registry. You can get the
```
bparees: Presumably they should push it to the `myproject/ubuntu:latest` imagestream tag?

bparees (changes requested):
a few grammatical cleanups, otherwise this looks terrific, thanks @legionus

admin_guide/registry_data_check_restore.adoc
legionus: @bparees Not necessary. The blobstore is global. You need to restore blob itself. The registry will find blob link by imagestream.
This is the exploitation of architectural bug in docker ))

legionus:
@bparees Re-pushed with cleanups.

admin_guide/registry_data_check_restore.adoc
bparees: ok, fair enough, but that would be a reasonable target for the push, right?  Is there any reason not to use that as the target for the push?

legionus: >  Is there any reason not to use that as the target for the push?

@bparees of course you can use this target for the push. In this case, it is no better than any other.

legionus: I did not mention this target because the image may be under different names in the registry. Any of these names you can use for the re-push.

bparees: ok, since admins/users probably don't understand the subtly of how blobs are linked into repositories, I think it's best to give them a recommendation of where to push it rather than leave them to sort it out.

legionus: @bparees OK. Take a look now.

bparees:
awesome, thanks @legionus 

@bmcelvee can you help get this into both our 3.11 OCP docs, as well as finding a home for it in 4.0?

bmcelvee:
Thanks, @legionus and @bparees! I'll get this merged in to 3.11 (probably tomorrow) and I will convert the content and add it to 4.0 in a separate PR.

bmcelvee:
@bparees - I'm working on recovering this PR

bparees:
thanks @bmcelvee

PR #13568, opened 2019-02-07
Update the monitoring diagram

rh-max:


vikram-redhat:
@rh-max - does this apply to 4.0 as well? Do you need to add it there? Is this waiting on an approval before it can go in to 3.11?

rh-max:
@vikram-redhat 
* Yep, it applies to 4.0 too, and for the current OKD as well.
* So yes, I need to add them. I plan to cherry-pick this to master and to make the same change for 4.0.
* No, I already have the approval for this change for 3.11 and OKD. It is to be merged now.

If you can, please merge it. I don't know when I will be assigned merge permissions.
Thanks for looking into this.

vikram-redhat:
@rh-max ok cool. It is not referenced from anywhere else atm, so I guess you will do a followup and add to master, 3.11 and 4.0 branches. Merging.

vikram-redhat:
@rh-max added in master and merged into 3.11. Might need a separate one for 4.0.

PR #13570, opened 2019-02-07
adding common-attributes to the conventions

kalexand-rh:


kalexand-rh:
@vikram-redhat, any objections? (@adellape, FYI)

PR #13578, opened 2019-02-07
Removing Ansible from Nodes topics

mburke5678:
https://jira.coreos.com/browse/OSDOCS-250

Brute force removal of:
Ansible
playbook
inventory file
openshift-ansible
/etc/ansible/hosts


mburke5678:
Merging. Further changes will be addressed via other Jira's and/or PRs. This issue removed the above terms only.

PR #13598, opened 2019-02-12
Removing references to Docker from Nodes topics

mburke5678:
https://jira.coreos.com/browse/OSDOCS-251


mburke5678:
@sjenning PTAL

modules/nodes-nodes-viewing-listing.adoc
```diff
@@ -112,14 +108,14 @@ System Info:                 <9>
  OS Image:                           OpenShift Enterprise
  Operating System:                   linux
  Architecture:                       amd64
- Container Runtime Version:          docker://1.13.1
+ Container Runtime Version:          1.13.1
```
sjenning: this will be something like `cri-o://1.12.5-6.rhaos4.0.git80d1487.el7` reflecting the crio RPM included in RHCOS

modules/nodes-pods-using-example.adoc
```diff
@@ -41,7 +41,7 @@ spec:
       value: ...
     - name: OPENSHIFT_MASTER
       value: https://master.example.com:8443
-    image: openshift/origin-docker-registry:v0.6.2 <5>
+    image: openshift/origin-registry:v0.6.2 <5>
```
sjenning: need to make sure these images exist.  or just remove this example.

modules/nodes-pods-secrets-certificates-about.adoc
```diff
@@ -20,9 +20,9 @@ masters.
 apiVersion: v1
   kind: Service
   metadata:
-    name: docker-registry
+    name: registry
     annotations:                                                        
-      service.alpha.openshift.io/serving-cert-secret-name: docker-cert<1>
+      service.alpha.openshift.io/serving-cert-secret-name: cert<1>
```
sjenning: `registry-cert`

modules/nodes-pods-using-example.adoc
```diff
@@ -57,7 +57,7 @@ spec:
       readOnly: true
   dnsPolicy: ClusterFirst
   imagePullSecrets:
-  - name: default-dockercfg-at06w
+  - name: default-cfg-at06w
```
sjenning: these might still be called dockercfg secrets.  not sure.

mburke5678: @sjenning 
```
imagePullSecrets:
    - name: default-dockercfg-5zrhb
```

mburke5678:
@sjenning Thanks for the input. Probably should go through and update command output for 4.0 at some point.

modules/nodes-pods-using-example.adoc
sjenning: just confirmed, this _is_ still called dockercfg  to match the upstream secret type `kubernetes.io/dockercfg`

mburke5678: That is what I saw in the console.

PR #13605, opened 2019-02-12
machine health check draft

kalexand-rh:
@ingvagabund, will you PTAL?

### DIFF
--- a/modules/machine-health-checks-about.adoc
+++ b/modules/machine-health-checks-about.adoc
@@ -1,0 +1,28 @@
+// Module included in the following assemblies:
+//
+// * master/deploying-machine-health-checks.adoc
+
+[id='machine-health-checks-about-{context}']
+= About MachineHealthChecks
+
+MachineHealthChecks automatically repair unhealthy Machines in a particular
+MachinePool.
+
+To monitor machine health, you create a resource to define the
+configuration for a controller. You set a condition to check for, such as
+staying in the `NotReady` status for 15 minutes or displaying a permanent condition
+in the node-problem-detector, and a label for the set of machines to monitor.
+
+[NOTE]
+====
+You cannot apply a MachineHealthCheck to a machine with the master role.
+====
+
+The controller that observes a MachineHealthCheck resource checks for the status
+that you defined. If a machine fails the health check, it is automatically deleted
+and a new one is created to take its place. When a machine is deleted, you
+see a `machine deleted` event. To limit disruptive impact of the machine
+deletion, the controller drains and deletes only one node at a time.
+
+
+To stop the check, you remove the resource.
--- a/modules/machine-health-checks-creating.adoc
+++ b/modules/machine-health-checks-creating.adoc
@@ -1,0 +1,24 @@
+// Module included in the following assemblies:
+//
+// * master/deploying-machine-health-checks.adoc
+
+[id='machine-health-checks-creating-{context}']
+= Creating a MachineHealthCheck resource
+
+You can create a MachineHealthCheck resource for all `MachinePools` in your
+cluster except the `master` pool.
+
+.Prerequisites
+
+* Install the `oc` command line and kubectl.
+
+.Procedure
+
+. Create a *_healthcheck.yml_* file that contains the definition of your
+MachineHealthCheck.
+
+. Apply the *_healthcheck.yml_* file to your cluster:
+[source,bash]
+----
+$ kubectl apply -f healcheck.yml
+----
--- a/modules/machine-health-checks-resource.adoc
+++ b/modules/machine-health-checks-resource.adoc
@@ -1,0 +1,67 @@
+// Module included in the following assemblies:
+//
+// * master/deploying-machine-health-checks.adoc
+
+[id='machine-health-checks-resource-{context}']
+= Sample MachineHealthCheck resource
+
+The MachineHealthCheck resource resembles the following YAML file:
+
+.MachineHealthCheck
+[source,yaml]
+----
+apiVersion: healthchecking.openshift.io/v1alpha1
+kind: MachineHealthCheck
+metadata:
+ name: example <1>
+ namespace: example <2>
+Spec:
+  Selector:
+    matchLabels:
+      sigs.k8s.io/cluster-api-cluster: <cluster_name> <3>
+      sigs.k8s.io/cluster-api-machine-role: <label> <4>
+      sigs.k8s.io/cluster-api-machine-type: <label> <4>
+      sigs.k8s.io/cluster-api-machineset: <cluster_name>-<label>-<AWS-zone> <5>
+----
+<1> Specify the name of the MachineHealthCheck to deploy. Include the name of the
+MachinePool to track.
+<2> Specify the namespace to deploy the MachineHealthCheck to.
+<3> Specify the name of your cluster.
+<4> Specify a label for the MachinePool that you want to check.
+<5> Specify the MachineSet to track in `<cluster_name>-<label>-<AWS-zone>`
+format. For example, `prod-node-us-east-1a`.
+
+
+
+////
+
+.MachinePoolHealthCheck
+[source,yaml]
+----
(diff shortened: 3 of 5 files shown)


modules/machine-health-checks-about.adoc
```diff
@@ -0,0 +1,29 @@
...
+
+The controller that observes a `MachineHealthCheck` resource checks for the status
+that you defined. If a machine fails the health check, it is automatically deleted
+and a new one is created to take its place. When a machine is deleted, you
+see <this kind of > event. The controller uses the same logic as the
+link:https://kubernetes.io/docs/concepts/architecture/nodes/#node-controller[Kubernetes node controller]
+to limit disruptive impact of the machine deletion.
+
+
+To stop the check, you remove the resource.
```
kalexand-rh: Is there any more conceptual information that a user might need to know to deploy a machine health check?

modules/machine-health-checks-creating.adoc
```diff
@@ -0,0 +1,24 @@
...
+You can create a `MachineHealthCheck` resource for all `MachinePools` in your
+cluster except the `master` pool.
+
+.Prerequisites
+
+* Install the `oc` command line and kubectl.
+
+.Procedure
+
+. Create a *_healthcheck.yml_* file that contains the definition of your
```
kalexand-rh: Is there a better method for applying a health check?

modules/machine-health-checks-resource.adoc
```diff
@@ -0,0 +1,66 @@
+// Module included in the following assemblies:
+//
+// * master/deploying-machine-health-checks.adoc
+
+[id='machine-health-checks-resource-{context}']
+= Sample `MachineHealthCheck` resource
+
+The `MachineHealthCheck` resource resembles the following YAML file:
+
+.MachineHealthCheck
```
kalexand-rh: Is this a valid sample YAML for the resource?

modules/machine-health-checks-resource.adoc
```diff
@@ -0,0 +1,66 @@
...
+<1> IDK
+<2> IDK
+<3> Enter the name of your cluster.
+<4> Enter a label for the `MachinePool` that you want to check.
+<5> Enter the `MachineSet` to track in `<cluster_name>-<label>-<AWS-zone>`
+format. For example, `prod-node-us-east-1a`.
+
+
+
+////
```
kalexand-rh: I saw both of these resources in a design doc. Are they still valid? If so, can we talk about how a user might employ them?

kalexand-rh (commented):
@ingvagabund, I have a couple of specific questions about how users are supposed to use the machine health check. Will you PTAL?

kalexand-rh:
@enxebre, when you have a chance, will you PTAL?

modules/machine-health-checks-about.adoc
```diff
@@ -0,0 +1,29 @@
+// Module included in the following assemblies:
+//
+// * master/deploying-machine-health-checks.adoc
+
+[id='machine-health-checks-about-{context}']
+= About `MachineHealthChecks`
+
+`MachineHealthChecks` are a specialized implementation of the `MachineAutoscaler`
```
enxebre: `MachineHealthChecks` is unrelated to `MachineAutoscaler`. It's a feature built on top of the machine API

ingvagabund: As @enxebre says, `MachineHealthChecks` is consumed by machine health-check controller, `MachineAutoscaler` by cluster autoscaler operator. Completely independent of each other. `MachineAutoscaler` "just" annotates already existing `MachineSet` object so it can be recognized by the cluster autoscaler. In general, cluster autoscaler removes nodes that are underutilized. Compared to machine health-checker that replaces nodes that are (in a way) broken. It's not guarantied both autoscaler and health checker will not interfere with each other though. Most likely, when a node becomes NotReady and pods running on the node are killed (for any reason), autoscaler will most likely scales up another node. It's just a matter of timing who's gonna by first. In general, machine health-checker will be faster as it's purpose is to fix nodes. Whereas, autoscaler's purpose is to dynamically add/remove nodes based on cluster workload.

modules/machine-health-checks-about.adoc
```diff
@@ -0,0 +1,29 @@
...
+
+[NOTE]
+====
+You cannot apply a `MachineHealthCheck` to a machine with the master role.
+====
+
+The controller that observes a `MachineHealthCheck` resource checks for the status
+that you defined. If a machine fails the health check, it is automatically deleted
+and a new one is created to take its place. When a machine is deleted, you
+see <this kind of > event. The controller uses the same logic as the
```
ingvagabund: `machine deleted` event

modules/machine-health-checks-about.adoc
```diff
@@ -0,0 +1,29 @@
...
+link:https://kubernetes.io/docs/concepts/architecture/nodes/#node-controller[Kubernetes node controller]
+to limit disruptive impact of the machine deletion.
```
ingvagabund: Not sure if this is completely valid assertion. @enxebre WDYT? Recently, we have merged node draining logic into aws actuator machine controller. So by default, when a machine is deleted and the machine is linked to a node, the node is drained (evicted) before it is deleted (e.g. in case pod disruption budget for pods is defined).

This might refer more to a fact that only a single machine is reconciled at a given time.

modules/machine-health-checks-about.adoc
ingvagabund: Just what is described in the docs we have: https://github.com/openshift/machine-api-operator#machine-healthcheck-controller

modules/machine-health-checks-creating.adoc
ingvagabund: There is not right now unless the resource is created by the installer itself.

modules/machine-health-checks-resource.adoc
ingvagabund: There is no strict rule on the matching labels. The idea is to have one machine pool per one machineset. So in case a `MachineHealthCheck` resource is create to guard one machine pool, the matching labels should be exactly the same as machineset matching labels. Though, it's not forbidden to create one `MachineHealthCheck` resource per a set of machine pools. So the matching labels can be more exotic than.

modules/machine-health-checks-about.adoc
enxebre: agreed with @ingvagabund. This is not correct. Today the controller does not have a global vision of all machines hence it does not limit disruptive impact of the machine deletion

modules/machine-health-checks-creating.adoc
kalexand-rh: Do you expect that the resource will be created by the installer itself in the future?

enxebre: nope any time soon. It's a post install feature

modules/machine-health-checks-about.adoc
kalexand-rh: Can you tell me more about the autoscaler and health checker interfering with each other? Is that something that a user should care about? (It sounds like the different processes will eventually self-correct even if they do temporarily collide.)

enxebre: imho this is not something a user should worry about. If they collide in a problematic way, means there's bug we need to fix

modules/machine-health-checks-about.adoc
kalexand-rh: I think I got that information from some data from November. Thank you for pointing out that it's out of date!

Is there anything specific that you think a user would need to know about machine deletion at this point? Just that the node's drained by evicting all the pods before it's deleted?

modules/machine-health-checks-creating.adoc
kalexand-rh: Great! Then if that ever happens, let me know so I can update the docs.

kalexand-rh:
@enxebre, @ingvagabund, I've made updates based on your comments and am going to merge. I'll be happy to make more updates later.

ingvagabund:
Thanks @kalexand-rh :)

PR #13608, opened 2019-02-13
Bugz 1663268: When exposing haproxy metrics, we will need to also update the readiness probe 

ramr:
When exposing haproxy metrics directly out, need to update the readiness probe as the probe paths as they are now served by haproxy. And haproxy just has a single monitor-uri healthz path.

fixes bugz 1663268 (https://bugzilla.redhat.com/show_bug.cgi?id=1663268)

@ariordan-redhat  PTAL Thx

### DIFF
--- a/install_config/router/default_haproxy_router.adoc
+++ b/install_config/router/default_haproxy_router.adoc
@@ -1752,6 +1752,12 @@ $ oc edit dc router
   value: haproxy
 ----
 +
+. Patch the router readiness probe to use the same path as the liveness probe as it is now served by the haproxy router:
++
+----
+$ oc patch dc router -p '"spec": {"template": {"spec": {"containers": [{"name": "router","readinessProbe": {"httpGet": {"path": "/healthz"}}}]}}}'
+----
++
 . Launch the stats window using the following URL in a browser, where the `STATS_PORT` value is `1936` by default:
 +
 ----


vikram-redhat:
This should have gone in master first, but happy to merge and CP back if QE approves. 

@bmeng - can you please look at this request?

ramr:
@vikram-redhat  Thanks. Was under the impression docs for 4.0 would be different but created a PR for master nonetheless. See: https://github.com/openshift/openshift-docs/pull/13622

kalexand-rh:
@ramr, 4.0's building off `enterprise-4.0` until GA, and `master` still feeds the 3.x stream until then.

vikram-redhat:
@ramr  thanks for submitting the new PR. Just to make sure I understand, this change is for 3.11 branch only, right? Not for 4.0? If for 4.0, then it needs to go to the enterprise-4.0 branch. If for 3.11, the new PR should work. It should also work if for both 3.11 and 4.0, and we can confirm with @ariordan-redhat to include it in the 4.0 docs.

ramr:
@vikram-redhat  it is for 3.11 (and I did it against master as that's a roll up of the current change as you had suggested). 
As of now, it is _not_ for 4.0 (we are exposing a lot less from a config standpoint in 4.0, that may change eventually but currently doesn't make sense to put in there). Thanks.

PR #13610, opened 2019-02-13
draft of autoscaler assembly

kalexand-rh:
@bison, will you PTAL?
What will be the right way to modify the autoscaler in a GA cluster?
Is this CRD a good representation of what users need to know to change it?


modules/cluster-autoscaler-crd.adoc
```diff
@@ -0,0 +1,56 @@
...
+      min: 4 <5>
+      max: 256 <6>
+    gpus:
+      - type: nvidia.com/gpu <7>
+        min: 0 <8>
+        max: 16 <9>
+      - type: amd.com/gpu <7>
+        min: 0 <8>
+        max: 4 <9>
+  scaleDown:
```
bison: We've recently added the `unneededTime` option in the `scaleDown` section. This sets the `--scale-down-unneeded-time` option on the autoscaler, which controls the amount of time before an "unneeded" node will be removed. Still need to update our example.

https://github.com/openshift/cluster-autoscaler-operator/pull/37

bison (commented):
This looks like a good start to me! I think we'll need to add something about `MachineAutoscaler` resources as well at some point. In general, the process is:
1. Create a single `ClusterAutoscaler` resource called `default`. This will enable the autoscaler in the cluster. There can only be one, and we only support modifying the fields shown in the example.
1. For each `MachineSet` you wish to scale (this is one per availability zone in AWS) you'll need to create a corresponding `MachineAutoscaler` resource with the min and max values and a reference to the `MachineSet`.

kalexand-rh:
@bison, I made some updates based on your comments and added some information about the MachineAutoscaler. Will you please take another look? Thank you!

modules/machine-autoscaler-crd.adoc
```diff
@@ -0,0 +1,35 @@
...
+  namespace: "openshift-cluster-api"
+spec:
+  minReplicas: 1 <2>
+  maxReplicas: 12 <3>
+  scaleTargetRef:
+    apiVersion: machine.openshift.io/v1alpha1
+    kind: MachineSet
+    name: worker <4>
+----
+<1> Specify the `MachineAutoscaler` name, which is composed of the name of the
```
bison: This is fine, but maybe just a point of clarification. This is actually just the name of the particular MachineAutoscaler you are creating. It can be anything you want, but we should strongly encourage users to simply use the same name as the MachineSet being targeted for autoscaling.

modules/machine-autoscaler-crd.adoc
```diff
@@ -0,0 +1,35 @@
...
+MachineSet to scale and the Amazon Web Services (AWS) zone to scale in.
+The value uses the following format:
+`"<MachineSet_name>-<Amazon_Web_Services-zone>"`.
+<2> Specify the minimum number Machines of the specified type to deploy in the
+specified AWS zone.
+<3> Specify the maxiumum number Machines of the specified type to deploy in the
+specified AWS zone.
+<4> Specify the type of Machine to scale.
```
bison: I would maybe reword this a bit. The `scaleTargetRef` is a reference by name to an existing MachineSet you want to scale. The `apiVersion` and `kind` will almost always be what's in the example for OpenShift users, and the name will be the name of an already existing MachineSet in the cluster.

bison (approved):
Two brief comments about the MachineAutoscaler bit, but looks good to me.

modules/machine-autoscaler-crd.adoc
kalexand-rh: @bison, is there a good way to get the correct  `apiVersion` and `kind` from the CLI, or is it best to go digging through the UI?

bison: Hmmm, well it's whatever the type of the MachineSet you want to scale is, I guess you could describe the existing MachineSet to see it: `oc -n openshift-machine-api describe $name`

For the foreseeable future in OpenShift, it's always going to be `apiVersion: machine.openshift.io/v1beta1` and `kind: MachineSet` though.

modules/machine-autoscaler-crd.adoc
```diff
@@ -0,0 +1,35 @@
...
+apiVersion: "autoscaling.openshift.io/v1alpha1"
+kind: "MachineAutoscaler"
+metadata:
+  name: "worker-us-east-1a" <1>
+  namespace: "openshift-cluster-api"
+spec:
+  minReplicas: 1 <2>
+  maxReplicas: 12 <3>
+  scaleTargetRef:
+    apiVersion: machine.openshift.io/v1alpha1
```
bison: Oops, this may be a type in our example upstream. This should be: `machine.openshift.io/v1beta1` I think. It's no longer alpha.

modules/machine-autoscaler-crd.adoc
kalexand-rh: Thank you so much! If you remember, I'd love a nudge when the `apiVersion` changes so I can update the sample.

kalexand-rh:
I made a few more tweaks based on @bison's comments, and I'm going to merge. Thanks again!

modules/machine-autoscaler-crd.adoc
bison: @kalexand-rh, I think our example is updated now:
https://github.com/openshift/cluster-autoscaler-operator/blob/master/examples/machineautoscaler.yaml

PR #13616, opened 2019-02-13
bug 1670732 need to update master-config.yaml for a cert change

kalexand-rh:
From https://bugzilla.redhat.com/show_bug.cgi?id=1670732


install_config/certificate_customization.adoc
```diff
@@ -343,38 +343,48 @@ For information specific to your load balancer environment, refer to link:https:
...
-<1> Path to a xref:configuring-custom-certificates-master[master API certificate]. 
+<1> Path to a xref:configuring-custom-certificates-master[master API certificate].
 
 . Change to the playbook directory and run the following playbook:
 +
 ----
 ansible-playbook /usr/share/ansible/openshift-ansible/playbooks/redeploy-certificates.yml
 ----
 
+. If you use xref:configuring-custom-certificates[named certificates]:
```
kalexand-rh: @jianlinliu, will you PTAL? (The other changes to the file are whitespace trimming.)

gpei: Hi @kalexand-rh I think we need to add some explanation here that the above steps are actually not working for now, so have to manually update the certificate parameters in master-config.yaml. Or else it looks like all of these steps are required.

kalexand-rh: @gpei, I thought you did have to do both. Scott said that "The playbook will ensure that the certificate files are distributed to the hosts but won't update master-config.yaml." I interpreted that to mean that you had to run the playbook to distribute the certs and then separately add the parameters to the master-config.yaml file.

Is that right?

gpei: Ah yes, I forgot the step to distribute the certs, you're correct.

kalexand-rh:
Thank you @gpei! Merging.

PR #13620, opened 2019-02-13
oc adm cordon vs oc adm manage-node node1 --schedulable=false

mburke5678:
https://jira.coreos.com/browse/OSDOCS-259

### DIFF
--- a/modules/nodes-nodes-working-marking.adoc
+++ b/modules/nodes-nodes-working-marking.adoc
@@ -14,14 +14,16 @@ scheduled on the node. Existing pods on the node are not affected.
 +
 [source,bash]
 ----
-$ oc adm manage-node <node1> <node2> --schedulable=false
+$ oc adm cordon <node>
 ----
 +
 For example:
 +
 [source,bash,options="nowrap"]
 ----
-$ oc adm manage-node node1.example.com --schedulable=false
+$ oc cordon node1.example.com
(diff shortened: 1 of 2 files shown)


modules/nodes-nodes-working-marking.adoc
```diff
@@ -14,14 +14,16 @@ scheduled on the node. Existing pods on the node are not affected.
 +
 [source,bash]
 ----
-$ oc adm manage-node <node1> <node2> --schedulable=false
+$ oc adm cordon <node>
```
mburke5678: @sjenning @smarterclayton It appears through my *arduous* testing that you can only specify one node with the `oc adm cordon` command. Please let me know if this is not the case.

mburke5678:
@sjenning @smarterclayton PTAL

modules/nodes-nodes-working-marking.adoc
sjenning: this appears to be true
```
oc adm cordon node1 node2
error: USAGE: cordon NODE [flags]
See 'oc adm cordon -h' for help and examples.
```
label selector method seems to be busted
```
$ oc adm cordon node --selector=beta.kubernetes.io/arch=amd64
error: cannot specify both a node name and a --selector option
See 'oc adm cordon -h' for help and examples.
```

sjenning:
LGTM

PR #13624, opened 2019-02-13
3.9 bumped latest-tags

sheriff-rh:
This is set to ship on 2019-02-20.


ahardin-rh:
The 3.9.68 advisories are now live :space_invader:  Merging!

PR #13625, opened 2019-02-13
osdocs-157 add samples operator modules

bmcelvee:
https://jira.coreos.com/browse/OSDOCS-157

This is just putting the framework in place for documenting the Samples Operator.


bmcelvee:
This content is not finalized, but I am going ahead and merging so we can continue to iterate on it.

PR #13627, opened 2019-02-13
AUTH-177: Updating Identity Provider instructions.

huffmanca:
AUTH-177: Updating Identity Provider instructions. This updates the instructions to define CRDs and apply them to the OS cluster. In addition, it also updates the HTPasswd section to include details about creating the HTPasswd secret.

This is for OS 4.0.


vikram-redhat:
@openshift/team-documentation PTAL for peer review.

modules/identity-provider-add.adoc
```diff
@@ -27,16 +27,28 @@ users can authenticate.
...
-. Restart the configuration server:
+. Log in to the cluster as a user from your identity provider.
++
+[source,bash]
+----
+$ oc login -u <username> -p <password>
+----
 
-. Log in to the cluster as a user from your identity provider:
+. Confirm that the user logged in successfully, and display the username.
```
kalexand-rh: user name

modules/identity-provider-create-CRD.adoc
```diff
@@ -24,18 +24,30 @@ Resource Definition (CRD) that describes it.
...
+  - name: my_identity_provider <1>
+    challenge: true <2>
+    login: true <3>
+    mappingMethod: claim <4>
+    type: <type> <5>
+    ...
 ----
+<1> A unique name defining the identity provider. This provider name is
+prefixed to provider user names to form an identity name.
+<2> When `true`, unauthenticated token requests from non-web clients (like the
```
kalexand-rh: no parentheses

modules/identity-provider-create-CRD.adoc
```diff
@@ -24,18 +24,30 @@ Resource Definition (CRD) that describes it.
...
-----
+Create a CRD file to describe the identity provider. A generic file displaying
+the structure is below.
 
-. Create the CRD for your identity provider:
-+
-[source,bash]
 ----
-$ command
+apiVersion: config.openshift.io/v1
```
kalexand-rh: Do we need to duplicate a CRD here, too?

modules/identity-provider-htpasswd-secret.adoc
```diff
@@ -0,0 +1,44 @@
+// Module included in the following assemblies:
+//
+// * authentication/identity_providers/configuring-htpasswd-identity-provider.adoc
+
+[id='identity-provider-creating-htpasswd-secret-{context}']
+= Creating the HTPasswd Secret
+
+To use the HTPasswd identity provider, you must define a secret containing the
```
kalexand-rh: s/containing/that contains

modules/identity-provider-htpasswd-secret.adoc
```diff
@@ -0,0 +1,44 @@
...
+metadata:
+  name: htpass-secret <1>
+  namespace: openshift-config
+data:
+  htpasswd: <base64-encoding> <2>
+----
+<1> This secret name is passed to providers to identify the encoding.
+<2> An existing encoding created during the previous step.
+
+. Apply the created CRD to create the secret in {product-title}.
```
kalexand-rh: s/created/

modules/identity-provider-htpasswd-secret.adoc
```diff
@@ -0,0 +1,44 @@
...
+. Create a base64 encoding of the HTPasswd users file.
++
+[source,bash]
+----
+$ base64 </path/to/users.htpasswd>
+----
+
+. Define a Custom Resource Definition (CRD) secret that contains the encoding.
++
+[source]
```
kalexand-rh: I'd [source,yaml]

kalexand-rh (commented):
couple of comments. LGTM.

modules/identity-provider-add.adoc
huffmanca: Incorporated.

modules/identity-provider-create-CRD.adoc
huffmanca: Incorporated. I've also removed parentheses from similar locations in the other CRD files.

modules/identity-provider-htpasswd-secret.adoc
huffmanca: Incorporated.

modules/identity-provider-htpasswd-secret.adoc
huffmanca: Incorporated.

modules/identity-provider-create-CRD.adoc
huffmanca: I'm not certain I understand the question. We could remove this generic CRD, and instead have the instructions just indicate that an IDP specific CRD needs to be defined?

kalexand-rh: That was the intent behind my question. :) After your discussion about using variables in modules, I wonder how many times we could reuse a module like this: https://github.com/openshift/openshift-docs/pull/13610/files#diff-21e6d7d3edc46a41e555b6ee7c74b494R1

huffmanca: I think the file you've referenced makes this largely redundant. I'd much rather point them to the specific OAuth configuration for each IDP and them tell them to apply that configuration instead of having this generic file. I'll remove this file in the next push and point to yours.

modules/identity-provider-htpasswd-secret.adoc
huffmanca: I've incorporated this suggestion into all of the *secret files.

modules/identity-provider-create-CRD.adoc
huffmanca: I've removed the generic CRD file, but there is a potential issue including the `deploying-resource.adoc` file in each of the assemblies. As of right now we can't define the Feature to any multi-word string. Thus the following is acceptable:

> IdentityProvider

But this is not:

> HTPasswd Identity Provider

At the beginning of each assembly we mention that only one user exists, and that a CR must be defined. At the end of the assembly we then demonstrate defining the CR, with an example for each IDP between the two.

Do you think this is sufficient, or should we add something similar to `deploying-resource.adoc` between the two modules?

kalexand-rh: I think it might be sufficient. I think that https://auth-177--ocpdocs.netlify.com/openshift-enterprise/latest/authentication/identity_providers/configuring-htpasswd-identity-provider.html makes sense. I think you should merge it and follow up if you get pushback during QE or dev review.

huffmanca: Sounds great. Thank you for checking!

PR #13628, opened 2019-02-14
Setting encoding for script

vikram-redhat:


vikram-redhat:
Pushing this fix as far back as 3.5, although it could technically go for all branches.

PR #12980, opened 2018-11-29
new assemblies and modules for scheduler

mburke5678:

PR #13652, opened 2019-02-18
Fixed level 3 heading to level 2 per Vikram

mburke5678:
https://github.com/openshift/openshift-docs/pull/12980#commitcomment-32354100


mburke5678:
@vikram-redhat PTAL to make sure this was the change you were looking for.

PR #13663, opened 2019-02-19
3.11.82 release notes and latest-tags

sheriff-rh:
This is shipping 2019-02-20.


jboxman:
@sheriff-rh, thanks for updating the various tags for this!

PR #13665, opened 2019-02-19
OSDOCS-286: Updated output of RBAC commands.

huffmanca:
Updated the RBAC outputs from a 4.0 cluster.

While the majority of outputs render fine locally, the ones under "Viewing cluster roles and bindings" are heinous. I've attempted to add the `options="nowrap"` attribute to this source block in an effort to increase readability. Suggestions for this block would be welcome.

This is for 4.0.

### DIFF
--- a/modules/rbac-viewing-local-roles.adoc
+++ b/modules/rbac-viewing-local-roles.adoc
@@ -14,15 +14,15 @@ You can use the `oc` CLI to view local roles and bindings by using the
 * Obtain permission to view the local roles and bindings:
 
 ifdef::openshift-dedicated[]
-** Users with the *dedicated-cluster-admin* role can view and manage local roles and bindings. 
+** Users with the `dedicated-cluster-admin` role can view and manage local roles and bindings.
 endif::[]
 
 ifdef::openshift-enterprise,openshift-origin[]
-** Users with the *cluster-admin* default cluster role bound cluster-wide can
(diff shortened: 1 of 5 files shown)


huffmanca:
@openshift/team-documentation - requesting peer review.

huffmanca:
After discussion in IRC I've removed the [source,bash] header from the offending source blocks. This allows the horizontal scrolling in the module where it's needed.

The remaining modules were left untouched for now.

modules/rbac-viewing-local-roles.adoc
```diff
@@ -14,7 +14,7 @@ You can use the `oc` CLI to view local roles and bindings by using the
 * Obtain permission to view the local roles and bindings:
 
 ifdef::openshift-dedicated[]
-** Users with the *dedicated-cluster-admin* role can view and manage local roles and bindings. 
+** Users with the *dedicated-cluster-admin* role can view and manage local roles and bindings.
```
kalexand-rh: \`dedicated-cluster-admin\` - we changed the formatting guidelines to almost always prefer just \` \`.

kalexand-rh (commented):
One pick, otherwise LGTM.

modules/rbac-viewing-local-roles.adoc
huffmanca: Thanks for the notice! I changed it here and in the other sections for this topic.

PR #13445, opened 2019-01-25
Operator can add and remove ES nodes

mburke5678:
https://jira.coreos.com/browse/OSDOCS-228

PR #13667, opened 2019-02-19
Remove topic to add/remove ES nodes

mburke5678:
From https://github.com/openshift/openshift-docs/pull/13445#discussion_r257307626
 

> jcantrill: 
> ewolinetz Thoughts here? I think we should exclude it all together as there is no way to modify via CL and I'm not certain how well the EO at this time supports modifying should we advise to go unmanged
> 
> ewolinetz
> +1 on this, we don't want to expose this level of detail to customers. currently it is possible for them to control that, but we intend to take that away and have the operator make these scaling/topology decisions for customers.
>


mburke5678:
@jcantrill @ewolinetz PTAL. Removed the topic on scaling ES nodes from 4.0 per your comments.

modules/efk-logging-elasticsearch-add-remove.adoc
```diff
@@ -1,118 +0,0 @@
-// Module included in the following assemblies:
-//
-// * logging/efk-logging-elasticsearch.adoc
-
-[id='efk-logging-elasticsearch-add-remove_{context}']
-= Changing the number of Elasticsearch nodes
-
-You can scale the number of nodes in Elasticsearch.
-
-For example, if you want to increase redundancy, and use the `FullRedundancy` or `MultipleRedundancy` policy, you can add additional data nodes to increase the number of shard replicas in your cluster.
```
ewolinetz: I think this is fine, I was commenting more on the additional role info which was below.

mburke5678: @ewolinetz We want to leave the info about adding `data` nodes, but not for adding `client` and `master`?

ewolinetz: i dont think we want to have any info about adding new nodes, maybe just scaling up or down. and also the redundancy configuration.

mburke5678: @ewolinetz I think that was what I was going for; perhaps I used incorrect or unclear wording? 
Can I just change the wording from _add a node_ to _scale up your cluster_?
Please see:
https://github.com/openshift/openshift-docs/pull/13667/commits/52d2f522238a2eeea18a71a846b0b74de0421378

modules/efk-logging-elasticsearch-add-remove.adoc
```diff
@@ -32,46 +28,26 @@ nodes:
   - nodeCount: 2 <1>
     nodeSpec:
       resources: {}
-    roles:  <2>
```
ewolinetz: can we leave the roles as they are here since that is currently how CLO will create them?

modules/efk-logging-elasticsearch-add-remove.adoc
```diff
@@ -32,46 +28,26 @@ nodes:
...
-<2> Specify the type of node.
+<2> Specify the `data` role.
 +
-[NOTE]
-====
-You can create 1 or 3 master nodes.
-====
-+
-For example, to scale the data and master nodes:
+For example, to add an additional node:
```
ewolinetz: I think this is what we want to leave off and only encourage users on how to scale up their current deployment.
Thoughts @jcantrill ?

modules/efk-logging-elasticsearch-add-remove.adoc
mburke5678: @ewolinetz  The CLO will create client/data/master nodes when scaling?
roles: 
   - client
    - data
    - master

vikram-redhat:
@mburke5678 merging this. Please make sure to follow up on any edits required.

PR #12900, opened 2018-11-16
Added node tuning operator content

ahardin-rh:
WIP Preview Build: http://file.rdu.redhat.com/~ahardin/Feb182019/node-tuning-operator/scalability_and_performance/using-node-tuning-operator.html

Dev card: https://jira.coreos.com/browse/ATS-185

Docs card: https://jira.coreos.com/browse/OSDOCS-45

PR #13669, opened 2019-02-19
Adding node tuning operator topic from ahardin-rh to Nodes bucket

mburke5678:
https://github.com/openshift/openshift-docs/pull/12900/


mburke5678:
@sjenning FYI. let me know if you think this topic inappropriate for the Nodes topics.

PR #13670, opened 2019-02-19
PR13300-follow-up images and image streams

bmcelvee:
Follow-up to https://github.com/openshift/openshift-docs/pull/13300

PR #13681, opened 2019-02-20
Added Chris's image topics to the map

vikram-redhat:


vikram-redhat:
@bmcelvee Hey Brandi, sorry, I didn't realize this but you had already started to include Chris's content in _topic_map.yml via https://github.com/openshift/openshift-docs/pull/13670/files. I did this today when I realized it wasn't showing up in the 4.0 docs. But at the same time, I removed the xrefs which were causing the builds to fail. Looks like there were other things you were doing in that PR which I haven't touched, so you may still be able to merge yours, but if not, can you close yours?

PR #13706, opened 2019-02-21
added padding after toc

vikram-redhat:


vikram-redhat:
@openshift/team-documentation FYI. This should take away the need for explicit &nbsp; in our assemblies. 

For a demo, see it live on the 4.0 docs :)

PR #13729, opened 2019-02-21
OSDOCS-241: Add the admin CLI commands

bergerhoffer:
@openshift/team-documentation For peer review. This section should apply to all distros except dedicated.

Marking WIP because we're commenting out CLI sections for the first round of reviews, and will need to update the _topic_map.yml once @vikram-redhat comments out all of the other existing sections that don't apply for the first round.

Preview: http://file.rdu.redhat.com/~ahoffer/2019/OSDOCS-241-cli-adm/cli_reference/administrator-cli-commands.html

### DIFF
--- a/modules/cli-administrator-maintenance.adoc
+++ b/modules/cli-administrator-maintenance.adoc
@@ -1,0 +1,30 @@
+// Module included in the following assemblies:
+//
+// * cli_reference/administrator-cli-commands.adoc
+
+[id='cli-maintenance-commands-{context}']
+= Maintenance CLI commands
+
+== migrate
+
+Migrate resources on the cluster to a new version or format depending on the
+subcommand used.
+
+.Example: Perform an update of all stored objects
+----
+$ oc adm migrate storage
+----
+
+.Example: Perform an update of only Pods
+----
+$ oc adm migrate storage --include=pods
+----
+
+== prune
+
+Remove older versions of resources from the server.
+
+.Example: Prune older builds including those whose BuildConfigs no longer exist
+----
+$ oc adm prune builds --orphans
+----
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -443,6 +443,9 @@ Topics:
 #Topics:
 #- Name: Developer CLI commands
 #  File: developer-cli-commands
+#- Name: Administrator CLI commands
+#  File: administrator-cli-commands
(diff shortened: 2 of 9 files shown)


soltysh (approved):
lgtm

bergerhoffer:
Unmarking as WIP since I've updated _topic_map.yml to comment out these new sections.

This is ready for peer review.

modules/cli-administrator-maintenance.adoc
```diff
@@ -0,0 +1,29 @@
+// Module included in the following assemblies:
+//
+// * cli_reference/administrator-cli-commands.adoc
+
+[id='cli-maintenance-commands-{context}']
+= Maintenance CLI commands
+
+== migrate
+
+Migrate resources on the cluster.
```
huffmanca: This is likely due to my ignorance with OpenShift, but can we provide any additional details on what the migration is doing? Is it migrating to a new version?

huffmanca (commented):
Majority of the content looks good, but I did have a question regarding the `migrate` command.

modules/cli-administrator-maintenance.adoc
bergerhoffer: @huffmanca I have to say that I am a bit unclear on this myself. Looking at the subcommands it seems to do something different depending (storage, image-references, etc.).

@soltysh Can you give us a more specific description about what the migrate command does?

soltysh: In the pre-4.0 world this was a required step during every upgrade to run migration to ensure we store data using the required new versions of resources (whenever they were changing versions they were stored in). In the new 4.0 world the commands should be used less frequently, since majority of those actions should be done automatically by the system. So stating that it's migrating to new version is somewhat correct (`oc adm migrate storage` is just for that), but not all commands are doing that. Other such as `oc adm migrate image-refs` is to allow changing image references when migrating images between repositories. `oc adm migrate legacy-hpa` (hopefully not needed) is meant to migrate data from the old format to the new one. Hopefully we'll deprecate these commands soon-ish :wink:

bergerhoffer: Cool thanks for the explanation @soltysh! To give a little more detail then, we'll update it to:

"Migrate resources on the cluster to a new version or format depending on the subcommand used."

soltysh: sgtm

PR #13748, opened 2019-02-22
WIP [enterprise-3.11] Update REST API to 3.11

jboxman:
- Copy files from origin/api/docs


jboxman:
@vikram-redhat, I wanted to see what a merge request for 3.11 REST API docs might look like, so here it is.

On origin:

`git checkout release-3.11`

On openshift-docs:

`git checkout enterprise-3.11`
`git checkout -b thisbranch`
`rsync -avH ../origin/api/docs/ rest_api/`

On first glance it seems correct.

vikram-redhat:
wow - nice work @jboxman. Will look at it in more details later today!

jboxman:
It looks like the issue with _<pre>_ appearing in our PDF output is due to this file:

https://github.com/openshift/origin/blob/release-3.11/tools/genapidocs/apidocs/templates/objectschema.adoc

And this was done by design[0], because the HTML output can include raw HTML. The PDF target seemingly cannot. (Even though it's ultimately created from rendered HTML using a browser rendering backend that creates PDFs.)

So it may be that file needs to be updated to use _asciidoc_ format to replace the current raw HTML.

[0] http://mrhaki.blogspot.com/2014/06/awesome-asciidoc-include-raw-html.html

jboxman:
I was going to send an email, then realized I can just _at_ @vikram-redhat too.

To fix this, I'd need to do the following:

1. Update https://github.com/openshift/origin/pull/22206 and swap the raw HTML in this file[0] for asciidoc markup. It looks like this template calls itself recursively, so this could get interesting.
1. Re-generate the asciidoc files. For 3.x this will require an installed and working OpenShift 3.11 master to query for the API spec. This will require I spin up a working OpenShift 3.11 somewhere. It's possible `minishift` will be sufficient. If not, I may need to try either Quicklab or install OCP on several of our bare-metal hosts.

[0] https://github.com/openshift/origin/blob/release-3.11/tools/genapidocs/apidocs/templates/objectschema.adoc

jboxman:
And this[0] is why the _details_ element had a rogue _open_ attribute. It's valid, but not when raw HTML is parsed by asciidoctor. It works fine in rendered HTML.

[0] https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details

jboxman:
This isn't a viable approach for resolving this, so I'm going to close this PR.

PR #13753, opened 2019-02-23
Remove master-config.yaml from Nodes bucket

mburke5678:


mburke5678:
Much of this is guesswork based on the Kubernetes Admission Controller docs for PodNodeSelector.

mburke5678:
Jira for followup changes: https://jira.coreos.com/browse/OSDOCS-311

mburke5678:
Also: https://github.com/openshift/openshift-docs/pull/13757

modules/nodes-scheduler-node-projects-configuring.adoc
```diff
@@ -5,148 +5,39 @@
...
-** Create a file containing the admission controller information:
+. Create a file containing the admission controller information:
 +
+[source,yaml]
 ----
 podNodeSelectorPluginConfig:
-    clusterDefaultNodeSelector: "k3=v3"
-     ns1: region=west,env=test,infra=fedora,os=fedora
+ clusterDefaultNodeSelector: name-of-node-selector
+ namespace1: name-of-node-selector
```
weinliu: is namespace1  the same as ns1 as below?
I not quite understand the example,  namespace1 and namespace2 have the same value

modules/nodes-scheduler-node-selectors-configuring.adoc
```diff
@@ -29,27 +32,37 @@ spec:
...
-        kind: PodNodeConstraintsConfig
-        nodeSelectorLabelBlacklist:
-          - kubernetes.io/hostname
-          - <label>
-...
+podNodeSelectorPluginConfig:
+ clusterDefaultNodeSelector: name-of-node-selector
+ namespace1: name-of-node-selector
+ namespace2: name-of-node-selector
+----
```
weinliu: Got the same question as below, is this indicating we can have multiple namespaces for the config? Can they be of the same?

modules/nodes-scheduler-node-projects-configuring.adoc
mburke5678: @weinliu Thank yo for your comments. Your pointing out these issues led me to learn that the defaultNodeselector is not yet in 4.1 at this time and not clear if the feature will get in. I commented-out the topic in the _topic.yaml file, hiding it from users until the feature is in; at which time I will work in your comments. 
I believe you are correct that I have this procedure wrong. 
PR for commenting out the feature:
https://github.com/openshift/openshift-docs/pull/14590

PR #13769, opened 2019-02-24
created sym links for monitoring docs and added images

vikram-redhat:


vikram-redhat:
@rh-max - I have added sym links for modules and images and added some missing images as well.

PR #13783, opened 2019-02-25
Document HAProxy behavior when APIServer LB is down, and provide 3rd party recommendations.

acomabon:
Masters HA section needs re-writing, since it does not take into account what happens when the APIserver goes down.
The following PR has been discussed within the organization. Reach out to me privately if need more info.
https://bugzilla.redhat.com/1641896

### DIFF
--- a/architecture/infrastructure_components/kubernetes_infrastructure.adoc
+++ b/architecture/infrastructure_components/kubernetes_infrastructure.adoc
@@ -184,9 +184,6 @@ ifdef::openshift-origin,openshift-enterprise,openshift-dedicated[]
 
 === High Availability Masters
 
-The availability of running applications remains if the master or any of its services fail.
-However, failure of master services reduces the ability of the system to respond to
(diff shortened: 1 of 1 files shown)


kalexand-rh:
/ok-to-test

@acomabon, please squash your commits.

@lihongan, will you PTAL and confirm the applicable versions for me? Based on the bug, I think it might be 3.5-3.11.

acomabon:
@kalexand-rh squashed!

acomabon:
@kalexand-rh @lihongan AFAIU it should be included in all 3.x versions, starting with 3.0, is there any reason why you thing it'd start with 3.5?

lihongan:
The changes are very basic principle for API LB, I think should be included in all versions. Thanks

kalexand-rh:
The bug referenced 3.5, but I'm happy to cherry-pick it back to 3.0. Thanks! Merging.

PR #13786, opened 2019-02-25
The default image prefix should be

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1680564

### DIFF
--- a/modules/efk-logging-configuring-image-about.adoc
+++ b/modules/efk-logging-configuring-image-about.adoc
@@ -5,19 +5,20 @@
 [id='efk-logging-configuring-image-about_{context}']
 = Understanding the cluster logging component image
 
-Each cluster logging component uses an image.
+Each cluster logging component uses an image. You can overwrite the image for each cluster logging component by modifying the appropriate
+environment variable in the `cluster-logging-operator` CR. 
 
 By default, cluster logging uses the following images:
 
-* *ELASTICSEARCH_IMAGE* deploys Elasticsearch, by default `docker.io/openshift/origin-logging-elasticsearch5:latest`.
-* *FLUENTD_IMAGE* deploys Fluentd, by default `docker.io/openshift/origin-logging-fluentd:latest`.
-* *KIBANA_IMAGE* deploys Kibana, by default `docker.io/openshift/origin-logging-kibana5:latest`.
-* *CURATOR_IMAGE* deploys Curator, by default `docker.io/openshift/origin-logging-curator5:latest`.
-* *OAUTH_PROXY_IMAGE* defines OAUTH for {product-title}, by default `docker.io/openshift/oauth-proxy:latest`.
-* *RSYSLOG_IMAGE* deploys Rsyslog, by default `docker.io/viaq/rsyslog:latest`.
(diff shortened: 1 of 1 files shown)


modules/efk-logging-configuring-image-about.adoc
```diff
@@ -9,15 +9,14 @@ Each cluster logging component uses an image.
...
-* *KIBANA_IMAGE* deploys Kibana, by default `docker.io/openshift/origin-logging-kibana5:latest`.
-* *CURATOR_IMAGE* deploys Curator, by default `docker.io/openshift/origin-logging-curator5:latest`.
-* *OAUTH_PROXY_IMAGE* defines OAUTH for {product-title}, by default `docker.io/openshift/oauth-proxy:latest`.
-* *RSYSLOG_IMAGE* deploys Rsyslog, by default `docker.io/viaq/rsyslog:latest`.
+* *ELASTICSEARCH_IMAGE* deploys Elasticsearch, by default `quay.io/openshift/origin-logging-elasticsearch5:latest`.
+* *FLUENTD_IMAGE* deploys Fluentd, by default `quay.io/openshift/origin-logging-fluentd:latest`.
+* *KIBANA_IMAGE* deploys Kibana, by default `quay.io/openshift/origin-logging-kibana5:latest`.
+* *CURATOR_IMAGE* deploys Curator, by default `quay.io/openshift/origin-logging-curator5:latest`.
+* *OAUTH_PROXY_IMAGE* defines OAUTH for {product-title}, by default `quay.io/openshift/origin-oauth-proxy:latest`.
+//* *RSYSLOG_IMAGE* deploys Rsyslog, by default `docker.io/viaq/rsyslog:latest`.
```
ewolinetz: this has nothing to do with choosing to deploy Rsyslog. This is the image that will be used for RSYSLOG when it is deployed. A component being deployed or not is configured by the CR.

Same verbiage for all other IMAGES here.

modules/efk-logging-configuring-image-about.adoc
```diff
@@ -9,15 +9,14 @@ Each cluster logging component uses an image.
...
 * *WATCH_NAMESPACE* allows cluster logging to view all namespaces in the cluster, by default `openshift-logging`.
```
ewolinetz: This is the namespace that the operator will watch for CR events in. It should be set by the installation means (this is typically OLM). For OLM it uses an annotation that it will also add to the deployment which corresponds to the operator group's namespace it is configured for.

modules/efk-logging-configuring-image-about.adoc
mburke5678: @ewolinetz Can the user change the rsyslog image by setting an env var outside the CR, or would we rather they not change it?

ewolinetz: these env vars are only available on the CLO deployment object. they are not available on a CR

mburke5678:
Further changes might be required via: https://github.com/openshift/openshift-docs/pull/13877/files#diff-39a1a250dee13635e22fcd0d176f5763

PR #13792, opened 2019-02-26
Removing all instances of source,bash and adding prompts to oc commands

bergerhoffer:
@openshift/team-documentation For peer review.

Marking as WIP since this should not be merged until after the stage 1 review is complete.

Preview: http://file.rdu.redhat.com/~ahoffer/2019/remove-source-bash/welcome/


vikram-redhat:
@bergerhoffer thanks for this! I checked the display in the customer portal docs to see what it looks like when we don't have [source, bash] and I believe there is no issue: 

https://access.redhat.com/documentation/en-us/openshift_container_platform/3.11/html-single/installing_clusters/#ensuring-host-access

Source: https://github.com/openshift/openshift-docs/blame/enterprise-3.11/install/host_preparation.adoc#L81

bergerhoffer:
Awesome, thanks for checking @vikram-redhat!

modules/nodes-nodes-managing-max-pods-proc.adoc
```diff
@@ -43,8 +43,8 @@ metadata:
 .. If the label is not present, add a key/value pair:
 +
 ----
-oc label machineconfigpool worker custom-kubelet=small-pods
----- 
+# oc label machineconfigpool worker custom-kubelet=small-pods
```
kalexand-rh: Do you need to be root to run the commands in this module?

kalexand-rh (commented):
One question, but LGTM. Do you have a script that checks for duplicate words?

modules/nodes-nodes-managing-max-pods-proc.adoc
bergerhoffer: I doubt it, but the previous CLI commands in the file all used '#' and I wanted to be consistent. I didn't want to change those just in case it was actually purposeful and I just didn't realize why.

bergerhoffer:
@kalexand-rh Kind of. I had started to write an emender test for the EAP repo for duplicate words, common misspellings, and other EAP-specific terms. I never ended up really finishing it or integrating it though.

modules/nodes-nodes-managing-max-pods-proc.adoc
kalexand-rh: That's fair. It's a better question for @mburke5678  for later, then. Our convention is to use # if you need root and $ otherwise. https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#code-blocks

kalexand-rh:
@bergerhoffer, then I won't ask for a copy of it. I'm glad that you have your WIP test to run. :)

vikram-redhat:
@bergerhoffer this can be rebased and merged now.

PR #13806, opened 2019-02-26
Bug 1683072, removed StatefulSets section

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1683072


ahardin-rh:
@bmcelvee @spadgett @benjaminapetersen FYI

benjaminapetersen:
looks good to me, I think these have been in the console long enough they don't need special treatment.

ahardin-rh:
QE verified this change in the BZ

PR #13807, opened 2019-02-26
Bug 1683068, corrected console CRD group

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1683068

@spadgett @benjaminapetersen @bmcelvee FYI

### DIFF
--- a/architecture/customizations.adoc
+++ b/architecture/customizations.adoc
@@ -178,10 +178,11 @@ that are based on many of these CRDs to add more functionality to your
 |
 
 |Console
-|console.openshift.io
-|
+|console.config.openshift.io
+|The top-level configuration for the web console.
 |Namespaced
-|
+|The console CR is created by default with more or less empty values. It honors
+new values. If it is deleted, it recreates automatically.
 
 |ControllerConfig
 |machineconfiguration.openshift.io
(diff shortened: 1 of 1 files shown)


architecture/customizations.adoc
```diff
@@ -178,7 +178,7 @@ that are based on many of these CRDs to add more functionality to your
 |
 
 |Console
-|console.openshift.io
+|config.openshift.io
```
benjaminapetersen: So, this is confusing with all the operators.  I'll dump a little here:

```bash
# there are two configs, both named "cluster"
# all of the operators should follow this pattern:
oc describe console.config.openshift.io cluster
oc describe console.operator.openshift.io cluster

# this is the user-facing top level config that should be documented
console.config.openshift.io 

# this is the operator config, probably does not need to be called out 
console.operator.openshift.io
```
All `configs` for all `operators` are named `cluster` because they are intended to be singletons.

benjaminapetersen: `Can deploy CR`... console CR will be created by default with more or less empty values.  It will honor new values, and if its deleted it will recreate automatically.

benjaminapetersen: `Description` perhaps is `The top level configuration for the web console.`?  That may not be super helpful as you could say `The top level configuration for <component>` for all of them, so perhaps more detail is needed.

ahardin-rh:
QE verified these changes in the BZ

PR #13810, opened 2019-02-26
BZ 1680359: CDI will not import from unsecure registries which are marked

ousleyp:
https://bugzilla.redhat.com/show_bug.cgi?id=1680359

### DIFF
--- a/modules/cnv_importing_vm_disk_to_pvc.adoc
+++ b/modules/cnv_importing_vm_disk_to_pvc.adoc
@@ -7,6 +7,11 @@ controller. When a PVC is created with special
 short-lived import pod that attaches to the PV and downloads the virtual
 disk image into the PV.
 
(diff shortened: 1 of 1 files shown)


modules/cnv_importing_vm_disk_to_pvc.adoc
```diff
@@ -7,6 +7,11 @@ controller. When a PVC is created with special
 short-lived import pod that attaches to the PV and downloads the virtual
 disk image into the PV.
 
+It is not currently possible to import images from insecure registries using
```
kalexand-rh: I'd s/It is not currently possible to import images from insecure registries using/You cannot import images from insecure registries by using

kalexand-rh (commented):
One suggestion, otherwise LGTM

PR #13819, opened 2019-02-26
Modules for Entitled Builds

adambkaplan:
In 4.0, builds which use subscription content will not be able to load entitlements and other configurations from the host node. This provides instruction/tutorials on how to add entitlements to builds.

Original Google doc (Red Hat internal only): https://docs.google.com/document/d/1udIkiF_-R6LUEzIFCnoafu4dFpQnz04zFePFBu7RwTM/edit?usp=sharing

### DIFF
--- a/modules/builds-source-input-satellite-config.adoc
+++ b/modules/builds-source-input-satellite-config.adoc
@@ -1,0 +1,32 @@
+[id='builds-source-input-satellite-config-{context}']
+= Adding satellite configurations to builds
+
+Builds which use Satellite to install content must provide appropriate 
+configurations to obtain content from Satellite repositories.
+
+.Prerequisites
+
+You must provide or create a yum-compatible repository configuration file which
+downloads content from your Satellite instance.
+
+.Procedure
+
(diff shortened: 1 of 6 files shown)


adambkaplan:
@bmcelvee think I have the granular modules we'll need for entitled builds. Key question is how to best tie this all together.

bmcelvee:
@adambkaplan - thank you, this looks great! The way to tie them all together would be an assembly in the Builds folder. Perhaps something like "Running entitled builds".

To add an assembly it would go directly in the Builds folder, and the file/name would go in the topic map; in the file we use includes to build the modules. I'm happy to merge what you have here and open a follow-up PR adding the assembly if you'd like.

adambkaplan:
@bmcelvee sounds like a plan. Found a few nits and squashed my commits.

adambkaplan:
@bmcelvee need me to make any additional updates here?

bmcelvee:
@adambkaplan no, thank you for the bump. I'll go ahead and get these into an assembly in a follow-up PR and tag you on it. Thanks again!

PR #13822, opened 2019-02-26
bug 1683035 pulling install-config.yaml references from quick install

kalexand-rh:
From https://bugzilla.redhat.com/show_bug.cgi?id=1683035

@jianlinliu, PTAL?


kalexand-rh:
This was approved on the source bug. Merging.

PR #13827, opened 2019-02-26
bug 1683131 updating install-config.yaml contents

kalexand-rh:
From https://bugzilla.redhat.com/show_bug.cgi?id=1683131

@jianlinliu, will you PTAL?

PR #13831, opened 2019-02-26
bug 1683026 updating parameter table

kalexand-rh:
From https://bugzilla.redhat.com/show_bug.cgi?id=1683026

@qinpingli, PTAL?

I updated table 1 based on the new parameter values that will be added with https://github.com/openshift/openshift-docs/pull/13827/files.


qinpingli:
LGTM

PR #13835, opened 2019-02-26
OSDOCS-241: Adding installing CLI plugins

bergerhoffer:
@kalexand-rh Would you mind peer reviewing this, since we were chatting earlier today about ending a module when you know there's a next step but can't xref?

@soltysh Can you please review these sections on enabling tab completion and working with plug-ins? I didn't use everything from [here](https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/), so can you make sure I didn't leave out anything critical?

Preview:
* http://file.rdu.redhat.com/~ahoffer/2019/OSDOCS-241-plugins/cli_reference/extending-cli-plugins.html
* http://file.rdu.redhat.com/~ahoffer/2019/OSDOCS-241-plugins/cli_reference/configuring-cli.html

Note that in the source, these new files are commented out for now, since they're not part of the stage 1 review.

### DIFF
--- a/cli_reference/configuring-cli.adoc
+++ b/cli_reference/configuring-cli.adoc
@@ -1,0 +1,9 @@
+[id='cli-configuring-cli']
+= Configuring the CLI
+include::modules/common-attributes.adoc[]
+:context: cli-configuring-cli
+
+toc::[]
+
+// Enabling tab completion
+include::modules/cli-configuring-completion.adoc[leveloffset=+1]
--- a/cli_reference/extending-cli-plugins.adoc
+++ b/cli_reference/extending-cli-plugins.adoc
@@ -1,0 +1,15 @@
+[id='cli-extend-plugins']
+= Extending the CLI with plug-ins
+include::modules/common-attributes.adoc[]
+:context: cli-extend-plugins
+
+toc::[]
+
+You can write and install plug-ins to build on the default `oc` commands,
+allowing you to peform new and more complex tasks with the {product-title} CLI.
+
+// Writing CLI plug-ins
+include::modules/cli-extending-plugins-writing.adoc[leveloffset=+1]
+
+// Installing and using CLI plug-ins
+include::modules/cli-extending-plugins-installing.adoc[leveloffset=+1]
--- a/modules/cli-configuring-completion.adoc
+++ b/modules/cli-configuring-completion.adoc
@@ -1,0 +1,34 @@
+// Module included in the following assemblies:
+//
+// * cli_reference/configuring-cli.adoc
+
+[id='cli-enabling-tab-completion-{context}']
+= Enabling tab completion
+
+After you install the `oc` CLI tool, you can enable tab completion to
+automatically complete `oc` commands or suggest options when you press Tab.
+
+.Prerequisites
+
+* You must have the `oc` CLI tool installed.
+
+.Procedure
+
+The following procedure enables tab completion for Bash.
+
+. Save the Bash completion code to a file.
++
+----
+$ oc completion bash > oc_bash_completion
+----
+
+. Copy the file to `/etc/bash_completion.d/`.
++
+----
+$ sudo cp oc_bash_completion /etc/bash_completion.d/
+----
++
+You can also save the file to a local directory and source it from your
+`.bashrc` file instead.
+
+Tab completion is enabled when you open a new terminal.
--- a/modules/cli-extending-plugins-installing.adoc
+++ b/modules/cli-extending-plugins-installing.adoc
@@ -1,0 +1,66 @@
+// Module included in the following assemblies:
+//
+// * cli_reference/extending-cli-plugins.adoc
+
+[id='cli-installing-plugins-{context}']
+= Installing and using CLI plug-ins
+
+After you write a custom plug-in for the {product-title} CLI, you can install it
+to use the additional functionality that it introduces.
+
+[IMPORTANT]
+====
+OpenShift CLI plug-ins are currently a Technology Preview feature.
+ifdef::openshift-enterprise[]
+Technology Preview features are not supported with Red Hat production service
+level agreements (SLAs), might not be functionally complete, and Red Hat does
+not recommend to use them for production. These features provide early access to
+upcoming product features, enabling customers to test functionality and provide
+feedback during the development process.
+
+See the link:https://access.redhat.com/support/offerings/techpreview/[Red Hat
+Technology Preview features support scope] for more information.
+endif::[]
+====
+
+.Prerequisites
+
+* You must have the `oc` CLI tool installed.
+* You must have a CLI plug-in file that begins with `oc-` or `kubectl-`.
+
+.Procedure
+
+. If necessary, update the plug-in file to be executable.
++
+----
+$ chmod +x <plugin_file>
+----
+. Place the file anywhere in your `PATH`, such as `/usr/local/bin/`.
++
+----
+$ sudo mv <plugin_file> /usr/local/bin/.
+----
+. Run `oc plugin list` to make sure that the plug-in file is listed.
++
+----
+$ oc plugin list
+The following compatible plugins are available:
+
+/usr/local/bin/<plugin_file>
+----
++
+If your plug-in is not listed here, verify that the file name begins with `oc-`
+or `kubectl-`, is executable, and is on your `PATH`.
+. Invoke the new command or option introduced by the plug-in.
++
+For example, if you built and installed the `kubectl-ns` plug-in from the
+ link:https://github.com/kubernetes/sample-cli-plugin[Sample plug-in repository],
+  you can use the following command to view the current namespace.
++
+----
+$ oc ns
+----
++
+Note that the command to invoke the plug-in depends on the plug-in file name.
+For example, a plug-in with the file name of `oc-foo-bar` is invoked by the `oc foo bar`
+command.
--- a/modules/cli-extending-plugins-writing.adoc
+++ b/modules/cli-extending-plugins-writing.adoc
@@ -1,0 +1,72 @@
+// Module included in the following assemblies:
+//
+// * cli_reference/extending-cli-plugins.adoc
+
+[id='cli-writing-plugins-{context}']
+= Writing CLI plug-ins
+
+You can write a plug-in for the {product-title} CLI in any programming language
+or script that allows you to write command-line commands. Note that you can not
+use a plug-in to overwrite an existing `oc` command.
+
+[IMPORTANT]
+====
+OpenShift CLI plug-ins are currently a Technology Preview feature.
+ifdef::openshift-enterprise[]
+Technology Preview features are not supported with Red Hat production service
+level agreements (SLAs), might not be functionally complete, and Red Hat does
+not recommend to use them for production. These features provide early access to
+upcoming product features, enabling customers to test functionality and provide
+feedback during the development process.
+
+See the link:https://access.redhat.com/support/offerings/techpreview/[Red Hat
+Technology Preview features support scope] for more information.
+endif::[]
+====
+
+.Procedure
+
+This procedure creates a simple Bash plug-in that prints a message to the
+terminal when the `oc foo` command is issued.
+
+. Create a file called `oc-foo`.
++
+When naming your plug-in file, keep the following in mind:
+
+* The file must begin with `oc-` or `kubectl-`.
+* The file name determines the command that invokes the plug-in. For example, a
+plug-in with the file name `oc-foo-bar` can be invoked by a command of
+`oc foo bar`. You can also use underscores if you want the command to contain
+dashes. For example, a plug-in with the file name `oc-foo_bar` can be invoked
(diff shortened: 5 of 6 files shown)


modules/cli-extending-plugins-writing.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+echo "I am a plugin named kubectl-foo"
+----
+
+Once you install this plug-in for the {product-title} CLI, it can be invoked
+using the `oc foo` command.
+
+.Additional resources
+
+* Review the link:https://github.com/kubernetes/sample-cli-plugin[Sample plug-in repository]
```
soltysh: Also add https://github.com/kubernetes/cli-runtime/ which is a set of helpers to create plugins in Go

soltysh (approved):
One suggestion, the rest lgtm.

modules/cli-extending-plugins-writing.adoc
bergerhoffer: Thanks @soltysh, I've added that as well.

cli_reference/extending-cli-plugins.adoc
```diff
@@ -0,0 +1,17 @@
+[id='cli-extend-plugins']
+= Extending the CLI with plug-ins
+include::modules/common-attributes.adoc[]
+:context: cli-extend-plugins
+
+toc::[]
+
+You can xref:cli-writing-plugins-{context}[write] and
```
kalexand-rh: I would omit the internal xrefs because of proximity to the TOC.

modules/cli-configuring-completion.adoc
```diff
@@ -0,0 +1,34 @@
+// Module included in the following assemblies:
+//
+// * cli_reference/configuring-cli.adoc
+
+[id='cli-enabling-tab-completion-{context}']
+= Enabling tab completion
+
+Once you have installed the `oc` CLI tool, you can enable tab completion to
```
kalexand-rh: s/Once you have installed/After you install

modules/cli-configuring-completion.adoc
```diff
@@ -0,0 +1,34 @@
...
+. Copy the file to `/etc/bash_completion.d/`.
++
+----
+$ sudo cp oc_bash_completion /etc/bash_completion.d/
+----
++
+You can also save the file to a local directory and source it from your
+`.bashrc` file instead.
+
+Tab completion will be enabled when you open a new terminal.
```
kalexand-rh: s/will be/is

modules/cli-extending-plugins-installing.adoc
```diff
@@ -0,0 +1,66 @@
+// Module included in the following assemblies:
+//
+// * cli_reference/extending-cli-plugins.adoc
+
+[id='cli-installing-plugins-{context}']
+= Installing and using CLI plug-ins
+
+Once you have written a custom plug-in for the {product-title} CLI, you can
```
kalexand-rh: s/Once you have written/After you write

modules/cli-extending-plugins-installing.adoc
```diff
@@ -0,0 +1,66 @@
...
+
+/usr/local/bin/<plugin_file>
+----
++
+If your plug-in is not listed here, verify that the file name begins with `oc-`
+or `kubectl-`, is executable, and is on your `PATH`.
+. Invoke the new command or option introduced by the plug-in.
++
+For example, if you built and installed the `kubectl-ns` plug-in from the
+ link:https://github.com/kubernetes/sample-cli-plugin[Sample plug-in repository],
```
kalexand-rh: We aren't supposed to include GitHub links. @vikram-redhat, will you PTAL?

modules/cli-extending-plugins-installing.adoc
```diff
@@ -0,0 +1,66 @@
...
+
+[id='cli-installing-plugins-{context}']
+= Installing and using CLI plug-ins
+
+Once you have written a custom plug-in for the {product-title} CLI, you can
+install it to use the additional functionality that it introduces.
+
+[IMPORTANT]
+====
+OpenShift CLI plug-ins are currently a Technology Preview feature.
```
kalexand-rh: @ahardin-rh, how did your experiments with a Tech Preview module turn out?

modules/cli-extending-plugins-writing.adoc
```diff
@@ -0,0 +1,72 @@
...
+if [[ "$1" == "config" ]]
+then
+    echo $KUBECONFIG
+    exit 0
+fi
+
+echo "I am a plugin named kubectl-foo"
+----
+
+Once you install this plug-in for the {product-title} CLI, it can be invoked
```
kalexand-rh: s/Once/After

cli_reference/configuring-cli.adoc
```diff
@@ -0,0 +1,9 @@
+[id='cli-configuring-cli']
+= Configuring the CLI
+include::modules/common-attributes.adoc[]
+:context: cli-configuring-cli
+
+toc::[]
+
```
kalexand-rh: Where do you explain how to install the CLI? I made a `install-cli.adoc` module for the install guide with basic instructions, but only the admin who installs a cluster will encounter that. I might make this topic "Installing and configuring the CLI" and include the installation steps and a guide to logging into the command line before you configure it. (Or make another assembly about installing and logging in.)

If you already have a module about installing the cli, will you let me know so we can combine them?

cli_reference/extending-cli-plugins.adoc
```diff
@@ -0,0 +1,17 @@
...
+xref:cli-installing-plugins-{context}[install] plug-ins to build on the default
+`oc` commands, allowing you to peform new and more complex tasks with the
+{product-title} CLI.
+
```
kalexand-rh: I'd make a prereq at the assembly level to install and configure the CLI and xref to the assembly or assemblies.

modules/cli-extending-plugins-writing.adoc
```diff
@@ -0,0 +1,72 @@
...
+
+This procedure creates a simple Bash plug-in that prints a message to the
+terminal when the `oc foo` command is issued.
+
+. Create a file called `oc-foo`.
++
+When naming your plug-in file, keep the following in mind:
+
+* The file must begin with `oc-` or `kubectl-`.
+* The file name determines the command path that will invoke the plug-in. For
```
kalexand-rh: s/will invoke/invoke

Is this a command path or just a command?

modules/cli-extending-plugins-writing.adoc
```diff
@@ -0,0 +1,72 @@
...
+example, a plug-in file name of `oc-foo-bar` can be invoked by a command of
```
kalexand-rh: s/a plug-in file name of/a plug-in with the file name (2 places)

modules/cli-extending-plugins-writing.adoc
```diff
@@ -0,0 +1,72 @@
...
+
+echo "I am a plugin named kubectl-foo"
+----
+
+Once you install this plug-in for the {product-title} CLI, it can be invoked
+using the `oc foo` command.
+
+.Additional resources
+
+* Review the link:https://github.com/kubernetes/sample-cli-plugin[Sample plug-in repository]
```
kalexand-rh: We aren't supposed to link to GitHub. I'd check with Vikram.

kalexand-rh (commented):
I have a couple of picks and suggestions, but it overall LGTM!

modules/cli-extending-plugins-installing.adoc
soltysh: I think in this particular case it's worth mentioning both, since one is a sample cli plugin, and the other is a resource for plugin authors.

cli_reference/configuring-cli.adoc
bergerhoffer: @kalexand-rh I've been adding my sections a bit out of order, so I haven't created that module yet. But there will eventually be a getting started section that will include installing and logging in to the CLI. I'll ping you once I do add this.

modules/cli-extending-plugins-installing.adoc
bergerhoffer: Hmm, I do see that guidance here [1] now, but it doesn't explain why... this is the only site that is called out as unallowed. What's the reasoning @vikram-redhat? Looking through our source, I do see plenty of places where github is referenced, and it seems like it would make sense to do so (unless there's some issue I'm unaware of). In this particular case, I do think that this is valuable information for users if we can provide it.

[1] https://github.com/openshift/openshift-docs/blob/enterprise-4.0/contributing_to_docs/doc_guidelines.adoc#links-to-external-websites

bergerhoffer:
Thanks for the peer review @kalexand-rh ! I've updated the PR with most of the items, though am still waiting to hear back on the GitHub linking question.

cli_reference/extending-cli-plugins.adoc
kalexand-rh: No, I think it's fair to have both.

modules/cli-extending-plugins-installing.adoc
bergerhoffer: @vikram-redhat - @kalexand-rh gave me some more background around where this "guidance" originated from. Let me know your thoughts on it, but maybe this is something we should discuss at the next team meeting.

vikram-redhat: @bergerhoffer We have been phasing out unsupported GitHub links as we go along. For a new docset like 4.0, we need to get buy in from support delivery to make sure they are ok with us adding this link.

The rationale is that if a GitHub link is provided in product docs, it signals to the customers that the content/plugin from that link is Red Hat supported, which is not always the case.

I am happy to add this, if @sferich888 gives his approval, otherwise we need to take it out.

sferich888:
I'm ok with the link in this case, however from a technical point of view I find the example to be less than helpful. 

Poiting to something more useful (or having better client api docs - glares at master/cli team) may help explain the content/concept (of writing a golang based plug-in) better. 

Example: https://github.com/openshift/must-gather can be built (git clone; make), re-named (mv openshift-must-gather oc-mg) and turned into a plug-in (by adding it to your path). This does something more useful than reimplemting an exsisting oc command and can provide use (in collecting data), thus a better example.

However the link you using is the same one upstream kube uses, so it's no better or worse than upstream docs. 

cc: @deads2k (we have to talk about API docs or the lack thereof).

soltysh:
> Example: https://github.com/openshift/must-gather can be built (git clone; make), re-named (mv openshift-must-gather oc-mg) and turned into a plug-in (by adding it to your path). This does something more useful than reimplemting an exsisting oc command and can provide use (in collecting data), thus a better example.

I'll see how far we can go with that and will open a PR if possible and reasonable.

soltysh:
I looked into what @sferich888 proposed wrt to making must-gather a sample plugin by renaming it. It works nicely and perfectly makes sense. Let's go with that in that case. @bergerhoffer mind updating that?

deads2k:
Before we document this, are there any known plugins in the world we really want to support?

bergerhoffer:
@soltysh If you have the basic instructions for using must-gather as a plugin, let me know and I can definitely update the docs to use that instead.

But not sure if anyone wants to weigh in on @deads2k comment first, @sferich888?

> Before we document this, are there any known plugins in the world we really want to support?

Plug-ins are marked as tech preview in this section, so I don't think they're technically supported anyways, at least not in production. So I think it's fine to include in the docs.

sferich888:
@deads2k I like must-gather as a plug-in over the namespace context switch plug-in example (the value is easier to explain). 

Since must-gather is part of the product (an image / part of the payload) it's better for us to show this as an example, over something were never going to support (the namespace context switch plug-in). 

@bergerhoffer (minor nit) the problem with this line of thinking is that, right now your correct, but when this feature is no longer technology preview do we plan to rip out this example, if we do what do, what we replace it with? Why not replace it now with what we want? 

In the end we simply need to define a plug-in that shows how plug-in work, and (imo) shows a user how to develop there own plug-in (which requires good docs/examples on how to use the client apis). @deads2k @soltysh do we have plans for such docs? If not must-gather might be a gateway in to that conversation? At the very least it shows/demonstrates how to interact with kube (over interacting with files - or parts of them).

bergerhoffer:
Understood @sferich888. Just let me know where we want to go with this. I'd like to try to get this in before we do the stage 2 QE review (so I'd need to get it in by next week), so let me know what we want to do and if we can get it nailed down by then. Thanks!

bergerhoffer:
@sferich888 @deads2k Any objections to merging this with the example plug-in as is, and then updating, in a future PR, to a better example? (If we can't nail down a better example this week?)

sferich888:
No merge away.

soltysh:
@sferich888 yeah, we'll need to work on some reasonable example for how to build your own plugin. must-gather might be too over-bloated to be that, imo.

bergerhoffer:
Excellent, merging. If we decide to provide a better example, we can work that in later on. Thanks!

PR #13864, opened 2019-02-27
Added RHEL CoreOS tag to common attributes

mburke5678:
we should set this as an attribute in the common-attributes file and let the team know.

https://bugzilla.redhat.com/show_bug.cgi?id=1683600

### DIFF
--- a/modules/nodes-nodes-viewing-listing.adoc
+++ b/modules/nodes-nodes-viewing-listing.adoc
@@ -114,7 +114,7 @@ System Info:    <9>
  System UUID:                             EC22BF97-A006-4A58-6AF8-0A38DEEA122A
  Boot ID:                                 f24ad37d-2594-46b4-8830-7f7555918325
  Kernel Version:                          3.10.0-957.5.1.el7.x86_64
- OS Image:                                Red Hat CoreOS 4.0
+ OS Image:                                Red Hat Enterprise Linux CoreOS 4.0
  Operating System:                        linux
  Architecture:                            amd64
  Container Runtime Version:               cri-o://1.12.5-2.rhaos4.0.gitd4191df.el7-dev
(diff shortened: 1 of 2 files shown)


mburke5678:
@openshift/team-documentation PTAL. Vikram asked me to add an attribute to common-attributes for RHEL CoreOS. Want to make sure I am doing it right.

kalexand-rh:
@mburke5678, can we have a short and long version? The final names are up in the air, but the last I heard, RHCOS was acceptable to use after the first mention. Maybe something like:

:op-system-first: RHEL CoreOS (RHCOS)
:op-system: RHCOS

mburke5678:
@kalexand-rh PTAL

kalexand-rh:
LGTM!

mburke5678:
@vikram-redhat PTAL

mburke5678:
@vikram-redhat

modules/nodes-nodes-viewing-listing.adoc
```diff
@@ -114,7 +114,7 @@ System Info:    <9>
  System UUID:                             EC22BF97-A006-4A58-6AF8-0A38DEEA122A
  Boot ID:                                 f24ad37d-2594-46b4-8830-7f7555918325
  Kernel Version:                          3.10.0-957.5.1.el7.x86_64
- OS Image:                                Red Hat CoreOS 4.0
+ OS Image:                                RHEL CoreOS 4.0
```
vikram-redhat: @mburke5678 would you need to change this to use the attribute? I am not sure what the official OS name is?

mburke5678: @vikram-redhat That string is in a clode block. The attribute won't work.

PR #13872, opened 2019-02-27
bug 1683710 correcting MachineAutoScaler CRD version

kalexand-rh:
From https://bugzilla.redhat.com/show_bug.cgi?id=1683710

@jhou1, this looks right in the current console. Will you please confirm?


jhou1:
I've confirmed this is correct. Should be `apiVersion: autoscaling.openshift.io/v1alpha1`

PR #13878, opened 2019-02-27
Applied the Tech Preview module

ahardin-rh:
Preview build: http://file.rdu.redhat.com/~ahardin/02272019/TP-node-tuning/scalability_and_performance/using-node-tuning-operator.html#accessing-an-example-node-tuning-operator-specification-node-tuning-operator

### DIFF
--- a/modules/custom-tuning-specification.adoc
+++ b/modules/custom-tuning-specification.adoc
@@ -5,21 +5,6 @@
 [id='custom-tuning-specification-{context}']
 = Custom tuning specification
 
-[IMPORTANT]
-====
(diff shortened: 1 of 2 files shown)


ahardin-rh:
@jmencak I turned the boilerplate "technology preview" language into a reusable module so that we can use it across our whole doc set for all tech preview features. Does calling the feature 
"Custom profiles for custom tuning specification" make sense to you?

Since this Tech Preview call-out sits above the Custom tuning specification section, I want it to be clear that this is associated with that section and not with the Accessing an example Node Tuning Operator specification section.

jmencak:
Thank you.

/lgtm

PR #13884, opened 2019-02-28
Update command output to match OCP 4.0

liangxia:


vikram-redhat:
@mburke5678  PTAL.

mburke5678:
@liangxia @vikram-redhat 
Should we change the version somehow to account for changes? I see: `v1.12.4+4dd65df23d`.
Maybe `v1.12` ?
Or does the version really matter?

vikram-redhat:
@liangxia @mburke5678 - I think we should just preface the output by saying that this is an example. That way, when it does eventually change again in the future, we are correct by saying that this was an example (and that the actual value that uses see might be different.)

liangxia:
@mburke5678 @vikram-redhat  I think it is better that all the command examples are updated once we move forward the product with major versions (E.g, here, we move from 3.x to 4.0, which is not backward compatible). 

I'll leave it to you to decide as you guys are more experienced.

vikram-redhat:
@liangxia Yes that makes sense. Since this is from 3.x to 4.0, we should put the values correct for 4.0 AND also say that this is an example. That way we are covered for both.

@mburke5678 does that work for you?

mburke5678:
@liangxia @vikram-redhat Yes both suggestions are good. I will fix this.

vikram-redhat:
@mburke5678 - can you move this forward?

PR #13892, opened 2019-02-28
bug 1684100 removing the Paris AWS region

kalexand-rh:
From https://bugzilla.redhat.com/show_bug.cgi?id=1684100

@jianlinliu, will you please confirm?


kalexand-rh:

PR #13893, opened 2019-02-28
bug 1683977 correcting set-max-pods name

kalexand-rh:
From https://bugzilla.redhat.com/show_bug.cgi?id=1683977

@liangxia. PTAL?


liangxia:
LGTM

PR #13865, opened 2019-02-27
bug 1677623 added pre-install check descriptions

sheriff-rh:
BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1677623

This PR is for 3.9 and below (to 3.4), adding a descriptor for `openshift_release` in the `pre-install checks` table. 

PR #13866  is for 3.10+

@jiajliu  PTAL at both PRs.

PR #13897, opened 2019-02-28
bug 1677623 added openshift_release pre-install descriptions

sheriff-rh:
Duplicate of #13865 due to commit errors, which is now closed. 

BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1677623

This PR is for 3.9 and below (to 3.4), adding a descriptor for `openshift_release in the `pre-install checks` table, which is in two locations.

@wmengRH PTAL, thanks.

### DIFF
--- a/install/configuring_inventory_file.adoc
+++ b/install/configuring_inventory_file.adoc
@@ -852,6 +852,9 @@ Use the Device Mapper Storage Driver guide].
 available either locally or in at least one of the configured container image
 registries on the host machine.
 
+|`openshift_release`
+|Specifies the generic release of {product-title} for containerized installations. For RPM installations, set a`package_availability` value.
+
 |`package_version`
 |Runs on `yum`-based systems determining if multiple releases of a required
 {product-title} package are available. Having multiple releases of a package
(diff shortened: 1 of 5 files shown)


wmengRH:
LGTM. 
Thanks.

install/configuring_inventory_file.adoc
```diff
@@ -852,6 +852,9 @@ Use the Device Mapper Storage Driver guide].
 available either locally or in at least one of the configured container image
 registries on the host machine.
 
+|`openshift_release`
+|Specifies the generic release of {product-title} for containerized installations. For RPM installations, refer to `package_availability`.
```
kalexand-rh: Do you mean that for RPM installations you need to use `package_availability` instead? If so, I'd do something like s/refer to \`package_availability\`/set a \`package_availability\` value   (two places)

kalexand-rh (commented):
One question. Otherwise LGTM.

install/configuring_inventory_file.adoc
sheriff-rh: I agree, thank you @kalexand-rh !

PR #13911, opened 2019-03-01
[DOCS] should use pod name with  command

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1684390


mburke5678:
Confirmed by jcantrill in Slack: https://coreos.slack.com/archives/CB3HXM2QK/p1551447896064200

PR #13915, opened 2019-03-01
bug 1684384 replacing kubectl with oc

kalexand-rh:
From https://bugzilla.redhat.com/show_bug.cgi?id=1684384

@liangxia, will you PTAL?


liangxia:
LGTM

PR #13917, opened 2019-03-01
bug 1683929 removing cluster.k8s.io CRDs

kalexand-rh:
From https://bugzilla.redhat.com/show_bug.cgi?id=1683929

@sunzhaohua2, I've removed the rows. PTAL?


sunzhaohua2:
LGTM.

PR #13919, opened 2019-03-01
Install Node Problem Detector operator does not require selecting Operator Group anymore

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1684405


mburke5678:
@joelsmith @weinliu PTAL. 
Do we need the step to create the Operator Group any longer?
Prerequisite #2, Create an Operator Group.
https://docs.openshift.com/container-platform/4.0/nodes/nodes/nodes-nodes-problem-detector.html#nodes-nodes-problem-detector-installing_nodes-nodes-problem-detector

Issue #13939, opened 2019-03-02
[enterprise-3.11] Add a guide about node labeling before creating the launch config


bysnupy:
### Which section(s) is the issue in?
* Version: only `v3.11`

* Section: [Creating the Launch Configuration and Auto Scaling group
](https://docs.openshift.com/container-platform/3.11/admin_guide/cluster-autoscaler.html#creating-LC-and-ASG-cluster-auto-scaler-cluster-auto-scaler)
  - `Prerequisites` part

### What needs fixing?

Add a condition in `Prerequisites` about `node labeling`, because the labels can be missing on the `nodes` autoscaled. For example, it leads unexpected result when `EFK` component is using, `logging-fluentd` `pod` are not start on the `node` autoscaled. The `pod` is required `logging-infra-fluentd=true` label in order to start on a `node` and  the label is configured dynamically by `playbooks`, so it's not permanent as default.

PR #13940, opened 2019-03-02
[enterprise-3.11] Add a node labeling guide before creating the launch config

bysnupy:
* Fix: https://github.com/openshift/openshift-docs/issues/13939

* Version: only `v3.11`

* Description:
  If `EFK` and some feature that is depended on `labels` of running `nodes` are using on the `cluster`, it leads unexpected issues, such as not starting expected pods. We would guide about the node labeling requirement kindly before creating launch configuration.


vikram-redhat:
@mburke5678 PTAL.

bysnupy:
@mburke5678 would you PTAL ?

mburke5678:
@bysnupy LGTM!

PR #13941, opened 2019-03-03
Correct the steps for redeploying router certificates

bysnupy:
- Fix: ["redeploy-router-certificates.yml" makes changes to wrong "service serving certificate secrets" annotation](https://bugzilla.redhat.com/show_bug.cgi?id=1672011)

- Version: `v3.3` ~ `v3.11`

- Description:
  If your `router` installed initially with `router-metrics-tls` annotation in the `router Service`, `router-metrics-tls secret` should be `service serving certificates secret`.  In contrast, `router-certs secret` is using as `wild card certificates` and not `service serving certificates secret`.

### DIFF
--- a/install_config/redeploying_certificates.adoc
+++ b/install_config/redeploying_certificates.adoc
@@ -577,13 +577,7 @@ $ oc rollout latest dc/docker-registry
 [[redeploying-router-certificates-manually]]
 ==== Redeploying Router Certificates Manually
 
-When routers are initially deployed, an annotation is added to the router's
-service that automatically creates a
-xref:../dev_guide/secrets.adoc#service-serving-certificate-secrets[service serving certificate secret].
-
-To redeploy router certificates manually, that service serving certificate can
-be triggered to be recreated by deleting the secret, removing and re-adding
-annotations to the `router` service, then redeploying the router:
+To redeploy router certificates manually, you must add new router certificates to a secret named `router-certs`, then redeploy the router:
 
 . Switch to the `default` project for the remainder of these steps:
 +
@@ -688,6 +682,18 @@ $ oc create secret tls router-certs --cert=router.pem \ <1>
 <1> *_router.pem_* is the file that contains the concatenation of the
 certificates that you generated.    
 
+. Redeploy the router:
++
+----
+$ oc rollout latest dc/router
+----
++
+When routers are initially deployed, an annotation is added to the router's
+service that automatically creates a
+xref:../dev_guide/secrets.adoc#service-serving-certificate-secrets[service serving certificate secret] named `router-metrics-tls`.
++
+To redeploy `router-metrics-tls` certificates manually, that service serving certificate can be triggered to be recreated by deleting the secret, removing and re-adding annotations to the router service, then redeploying the `router-metrics-tls` secret:
+
 . Remove the following annotations from the `router` service:
 +
 ----
@@ -696,15 +702,15 @@ $ oc annotate service router \
     service.alpha.openshift.io/serving-cert-signed-by-
 ----
 
+. Remove the existing `router-metrics-tls` secret.
++
+----
+$ oc delete secret router-metrics-tls
+----
+
 . Re-add the annotations:
 +
 ----
 $ oc annotate service router \
-    service.alpha.openshift.io/serving-cert-secret-name=router-certs
-----
-
-. Redeploy the router:
-+
-----
-$ oc rollout latest dc/router
+    service.alpha.openshift.io/serving-cert-secret-name=router-metrics-tls
 ----


vikram-redhat:
@stuartchuan - can you PTAL? 

@huffmanca - FYI. If QE passes it, we should be ok to merge and CP.

install_config/redeploying_certificates.adoc
```diff
@@ -688,6 +682,24 @@ $ oc create secret tls router-certs --cert=router.pem \ <1>
...
++
+----
+$ oc rollout latest dc/router
+----
+
+When routers are initially deployed, an annotation is added to the router's
+service that automatically creates a
+xref:../dev_guide/secrets.adoc#service-serving-certificate-secrets[service serving certificate secret] named `router-metrics-tls`.
+
+To redeploy `router-metrics-tls` certificates manually, that service serving certificate can be triggered to be recreated by deleting the secret, removing and re-adding annotations to the router service, then redeploying the `router-metrics-tls` secret:
```
lihongan: ... can be triggered to be recreated by `removing annotations to the router service,  deleting the secret and re-adding annotations to the router service`, I think the secret should be deleted after removing the annotation, otherwise it will be recreated immediately.

lihongan: Line 708: `service.alpha.openshift.io/serving-cert-signed-by-` this annotation should not be removed.

lihongan: others LGTM, thanks

bysnupy: Thank you for your review and kind correction @lihongan,

You're right. I changed the sequence of the steps as you mentioned. (removing secret step is after removing annotation step) 

And I think it had better to remove `service.alpha.openshift.io/serving-cert-signed-by` annotation at same time. The annotation also added automatically with the current `OpenShift CA` signer value, when it is added `service.alpha.openshift.io/serving-cert-secret-name` again. For example, if new `CA` is deployed before the `router certs` deploying work, it can cause not matching CA issue with new `CA`, because the `service.alpha.openshift.io/serving-cert-signed-by`'s value is old or different with new `CA` signer value.

lihongan: That make sense. Thank you @bysnupy

vikram-redhat:
@stuartchuan - can you confirm if this is ok from QE's point of view?

@xltian

stuartchuan:
@vikram-redhat ,
@lihongan is the QE owner of this part, seems he had reviewed the changes and added lgtm label already.

vikram-redhat:
Thanks @stuartchuan! 

@huffmanca - can you please do a peer review and followup if needed?

bysnupy:
Could you PTAL @lihongan ? ["redeploy-router-certificates.yml" makes changes to wrong "service serving certificate secrets" annotation](https://bugzilla.redhat.com/show_bug.cgi?id=1672011) status is already changed to `VERIFIED`, if some tasks are required, let me know that. thanks.

install_config/redeploying_certificates.adoc
```diff
@@ -688,6 +682,18 @@ $ oc create secret tls router-certs --cert=router.pem \ <1>
 <1> *_router.pem_* is the file that contains the concatenation of the
 certificates that you generated.    
 
+. Redeploy the router:
++
+----
+$ oc rollout latest dc/router
+----
+
```
huffmanca: Need to add a `+` here so that the ordered list doesn't break from the text below.

install_config/redeploying_certificates.adoc
```diff
@@ -688,6 +682,18 @@ $ oc create secret tls router-certs --cert=router.pem \ <1>
...
+When routers are initially deployed, an annotation is added to the router's
+service that automatically creates a
+xref:../dev_guide/secrets.adoc#service-serving-certificate-secrets[service serving certificate secret] named `router-metrics-tls`.
+
```
huffmanca: Same as above, this line needs a `+` added so that the ordered list isn't broken.

huffmanca (changes requested):
I sincerely apologize for the delays in submitting this review. Overall it looks good, just need to tweak two lines so that the numbering isn't broken from the text.

Once these changes are made I will merge and cherry-pick.

install_config/redeploying_certificates.adoc
bysnupy: Thank you for your review ! I've done as mentioned.

install_config/redeploying_certificates.adoc
bysnupy: Thank you, I've fixed.

bysnupy:
@huffmanca I've complete to fix it as mentioned from you. Could you PTAL again ?

huffmanca:
@bysnupy,

Thanks for making these changes. It looks good to me!

huffmanca (approved):

huffmanca:
As the 3.5 through 3.3 cherry-picks failed, these have been manually included under:

https://github.com/openshift/openshift-docs/pull/14430
https://github.com/openshift/openshift-docs/pull/14431
https://github.com/openshift/openshift-docs/pull/14432

PR #13967, opened 2019-03-04
bumped tags for 3.10.119

sheriff-rh:
Ships 13 March 2019.


ahardin-rh:
@sheriff-rh We need the release notes content to be added to the master branch too; the tag bumps are correct being against 3.10 only

PR #13972, opened 2019-03-04
Updating doc guidelines for not using source,bash

bergerhoffer:
@openshift/team-documentation For peer review.

### DIFF
--- a/contributing_to_docs/doc_guidelines.adoc
+++ b/contributing_to_docs/doc_guidelines.adoc
@@ -452,12 +452,13 @@ $ lorem.sh
 Without the line spaces, the content is likely to be not parsed correctly.
 
 * It is recommended to include source tags for the programming language used in the code block to enable syntax highlighting. For example, use the source tags
- `[source, bash]`, `[source, yaml]`, `[source, javascript]`.
+ `[source, yaml]` or `[source, javascript]`.
++
+NOTE: Do not use `[source, bash]` for `oc` commands or any terminal commands, because it does not render properly in some cases.
 +
 ....
 Lorem ipsum
 
-[source, bash]
 ----
 $ lorem.sh
 ----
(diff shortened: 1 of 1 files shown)


vikram-redhat:
@bergerhoffer this would apply to all content, not just 4.0, right?

bergerhoffer:
@vikram-redhat Yes probably, though I only went through and remove source,bash instances from the 4.0 docs.

As for the guidance, I thought that enterprise-4.0 would be replacing master, so that would eventually becoming guidance for all? Or do you want me to submit an update to the master version of the guidance?

vikram-redhat:
Hey @bergerhoffer - yes that is right. enterprise-4.0 will become master, but we will still have guidance for 3.x docs (when 4.0 becomes master, we will move the current master guidance to 3.11, the last 3.x release.)

vikram-redhat:
FWIW, I am happy for us to not remove the source, bash from 3.x docs at this stage, just have it in the guidance for any future work.

bergerhoffer:
@vikram-redhat I opened https://github.com/openshift/openshift-docs/pull/13977 to update the guidance on master for 3.x, if you want to take a look.

contributing_to_docs/doc_guidelines.adoc
```diff
@@ -452,12 +452,13 @@ $ lorem.sh
 Without the line spaces, the content is likely to be not parsed correctly.
 
 * It is recommended to include source tags for the programming language used in the code block to enable syntax highlighting. For example, use the source tags
- `[source, bash]`, `[source, yaml]`, `[source, javascript]`.
+ `[source, yaml]` or `[source, javascript]`.
++
+NOTE: Do not use `[source, bash]` for `oc` commands or any terminal commands, as it does not render properly in some cases, so we leave off for consistency across the docs.
```
kalexand-rh: s/as/because
s/cases, so we leave off for consistency across the docs./cases.

kalexand-rh (commented):
I missed that you had PRs for both streams.

PR #13977, opened 2019-03-05
Updating source,bash guidance in 3.x docs as well

bergerhoffer:
@openshift/team-documentation For peer review. This should just go in master for general guidance.

### DIFF
--- a/contributing_to_docs/doc_guidelines.adoc
+++ b/contributing_to_docs/doc_guidelines.adoc
@@ -444,12 +444,13 @@ $ lorem.sh
 Without the line spaces, the content is likely to be not parsed correctly.
 
 * It is recommended to include source tags for the programming language used in the code block to enable syntax highlighting. For example, use the source tags
- `[source, bash]`, `[source, yaml]`, `[source, javascript]`.
+ `[source, yaml]` or `[source, javascript]`.
++
+NOTE: Do not use `[source, bash]` for `oc` commands or any terminal commands, because it does not render properly in some cases. We will not accept new contributions that contain `[source, bash]`.
 +
 ....
 Lorem ipsum
 
-[source, bash]
 ----
 $ lorem.sh
 ----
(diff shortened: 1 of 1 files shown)


contributing_to_docs/doc_guidelines.adoc
```diff
@@ -444,12 +444,13 @@ $ lorem.sh
 Without the line spaces, the content is likely to be not parsed correctly.
 
 * It is recommended to include source tags for the programming language used in the code block to enable syntax highlighting. For example, use the source tags
- `[source, bash]`, `[source, yaml]`, `[source, javascript]`.
+ `[source, yaml]` or `[source, javascript]`.
++
+NOTE: Do not use `[source, bash]` for `oc` commands or any terminal commands, as it does not render properly in some cases, so we leave off for consistency across the docs.
```
kalexand-rh: s/as/because
s/cases, so we leave off for consistency across the docs./cases. We will not accept new contributions that contain \`[source, bash]\`.

kalexand-rh (commented):
A pick and a suggestion. (Unless you're planning to prune all the 3.x branches, and I don't think you should, it makes all the current inclusions sound like mistakes.)

contributing_to_docs/doc_guidelines.adoc
kalexand-rh: You'll also need to move this into 4.0. (When 4.0 GAs, we're going to deprecate this branch.) For that collection, I'd leave off the last sentence.

bergerhoffer: @kalexand-rh I had opened https://github.com/openshift/openshift-docs/pull/13972 for 4.0, and @vikram-redhat asked me to also include this guidance for 3.x (this PR). I'll update w/ your suggestions here.

And no, I'm not going through and removing all source,bash instances from the 3.x doc set.

PR #12866, opened 2018-11-14
New assembly and modules for About EFK Logging

mburke5678:

PR #13979, opened 2019-03-05
Changes per Rich in PR

mburke5678:
https://github.com/openshift/openshift-docs/pull/12866#issuecomment-444163493

> There are some major changes for 4.0:
>     names of all of the objects are changed - for example logging-es is now elasticsearch; logging-kibana is now kibana; logging-fluentd -> fluentd - for secret names, dc names, ds names, configmap names, etc. etc.
>     if you have a separate OPS cluster, you will not have logging-es and logging-es-ops, you will have elasticsearch-infra (for ops) and elasticsearch-app (for container apps)
>     no more deploymentConfigs - we have switched to Deployments - so for example, anywhere which mentions oc get dc -l component=es will need to change - commands like oc rollout are no longer applicable
>     There will probably be many more changes like this
>


mburke5678:
Merging for HTB. Further changes can be tracked in a followup PR

mburke5678:
@richm PTAL

mburke5678:
@anpingli Please take a look.

mburke5678:
@anpingli Please take a look. I can make any changes in a follow up PR.

PR #13982, opened 2019-03-05
bug 1685346 removing note about routers

kalexand-rh:
From https://bugzilla.redhat.com/show_bug.cgi?id=1685346

@enxebre, @lihongan  does this sound right to you?

### DIFF
--- a/modules/machineset-manually-scaling.adoc
+++ b/modules/machineset-manually-scaling.adoc
@@ -37,17 +37,8 @@ machines to be available.
 +
 [IMPORTANT]
 ====
-The default installation creates only one router. If you scale down your cluster,
-the node that hosts the router pod might be removed. In this case, you temporarily
-cannot access the web console and some other resources. After the router pod
-is redeployed to a different node, service will be restored.
-
-You can check the status of the router pod by running:
-
-----
-$ oc get pod -n openshift-ingress
-----
-
-If there is no router pod, or if it is in the `ContainerCreating` state, wait
-a little longer.
+By default, the {product-title} router pods are deployed on workers.
+Because the router is required to access some cluster resources, including the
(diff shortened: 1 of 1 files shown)


enxebre:
the nodes are draining before deleting the machine so router should never go unavailable

modules/machineset-manually-scaling.adoc
```diff
@@ -34,20 +34,3 @@ $ oc edit <machineset> -n openshift-machine-api
 +
 You can scale the MachineSet up or down. It takes several minutes for the new
 machines to be available.
-+
-[IMPORTANT]
-====
-The default installation creates only one router. If you scale down your cluster,
```
lihongan: Some facts:
1. currently two router pod will be installed by default.
2. router pods are deployed to `worker` node.
3. if you scale down your cluster and no worker nodes, then router pod should be in `pending` status and cannot access the web console.

lihongan: my suggestion as below and please help to refine them
```
[IMPORTANT]
====
The default installation creates two routers and deploys them on worker nodes. If you 
scale down your cluster, please ensure at least one worker node is running. otherwise, 
you may cannot access the web console and some other resources. 

You can check the status of the router pod by running:
----
$ oc get pod -n openshift-ingress
----

kalexand-rh:
@lihongan, will you PTAL at the new note? Because two router pods are deployed by default and the nodes are drained sequentially, I don't think the user will need to check on the router pods. I think it's more important to tell them to not scale to 0 worker nodes. What do you think?

lihongan:
@kalexand-rh yes, agree with you. 
The change looks good, thank you.

kalexand-rh:
Thanks! Merging. :)

PR #13986, opened 2019-03-05
bug in vsphere config

bagnaram:
Should be default-datasture https://github.com/kubernetes/kubernetes/blob/release-1.11/test/e2e/storage/vsphere/config.go

### DIFF
--- a/install_config/topics/manually_configuring_for_vsphere.adoc
+++ b/install_config/topics/manually_configuring_for_vsphere.adoc
@@ -65,7 +65,7 @@ create the file and enter the following:
         server = "10.10.0.2" <10>
         datacenter = "mydatacenter"
         folder = "path/to/vms" <11>
-        datastore = "shared-datastore" <12>
+        default-datastore = "shared-datastore" <12>
         resourcepool-path = "myresourcepoolpath" <13>
 
 [Disk]


kalexand-rh:
/ok-to-test
@cooktheryan, will you PTAL and confirm that this change applies to 3.10 and 3.11?

cooktheryan:
@kalexand-rh will do. Paging our vmware guy @dav1x

mazzystr:
LGTM

dav1x:
The vsphere cloud provider is intentionally backwards compatible from the prior to multiple vCenter configuration. 

Please see here:
https://github.com/kubernetes/kubernetes/blob/release-1.11/pkg/cloudprovider/providers/vsphere/vsphere.go#L129


https://vmware.github.io/vsphere-storage-for-kubernetes/documentation/existing.html#vsphere-cloud-config-file-for-kubernetes-version-18x-or-below

Bearing all this in mind, the documentation should be updated to include the different configurations listed here:
https://vmware.github.io/vsphere-storage-for-kubernetes/documentation/existing.html#vsphere-cloud-config-file-for-kubernetes-version-19x-and-above

The current PR /lgtm ill create an addition PR with the info from above as well. This change can be applied to 3.10 and 3.11

kalexand-rh:
Thank you all! I'll get this change into 3.10 and 3.11 and look for @dav1x's followup.

PR #13993, opened 2019-03-05
OSDOCS-276 upgrade draft

kalexand-rh:
From https://jira.coreos.com/browse/OSDOCS-276


modules/update-service-overview.adoc
```diff
@@ -1,28 +1,35 @@
...
-perform the upgrade to your image registry, and the Cluster Version Operator
-upgrades your cluster. By accepting automatic updates, you can automatically
+The {product-title} update service is the hosted service that provides over the
+air updates to both {product-title} and {op-system-first}. It provides a graph,
+or diagram that contain _vertices_ and the _edges_ that connect them, of
+component Operators. The edges in the graph show which versions you can safely
+upgrade to, and the vertices are update payloads that specify the intended state
+of the managed cluster components.
+
+The Cluster Version Operator (CVO) checks with the {product-title} update
```
abhinavdahiya: would it be useful to somehow imply to the users that CVO is running in _their_ cluster?

modules/update-service-overview.adoc
```diff
@@ -1,28 +1,35 @@
...
+service and determines valid upgrades and upgrade paths based on current
```
abhinavdahiya: _upgrade paths_ are more of the Cincinnati 's job.

modules/update-service-overview.adoc
```diff
@@ -1,28 +1,35 @@
...
+component versions and information in the graph. When you request an upgrade,
```
abhinavdahiya: When an upgrade is requested, CVO just uses the release image as specified by the Update / the graph and upgrades the cluster to the release image. It does not push it to any image registry.

modules/upgrade-cluster-version-parameters.adoc
```diff
@@ -0,0 +1,168 @@
...
+
+|`ClusterVersionSpec.desiredUpdate`
+|Optional. The cluster version that you want. If the current and intended
+version do not match, an upgrade starts. Set only a version from the list of
+recommended updates or your upgrade might fail.
+If an update exists with a version in the `availableUpdates` or history,
+you can specify the version field without specifying the image.
+
+If an upgrade fails, the Operator stops and reports the status of the failing
+component. Setting the `desiredUpdate` value to the previous version starts a
```
abhinavdahiya: It might be useful to mention that roll backs are not verified / supported actions.
/cc @smarterclayton

modules/upgrade-cluster-version-parameters.adoc
```diff
@@ -0,0 +1,168 @@
...
+cluster.
+
+[IMPORTANT]
+====
+The default channel contains stable updates. Do not specify a
+`ClusterVersionSpec.channel` value on production clusters.
+====
+
+|`ClusterVersionSpec.overrides`
+|Optional. A list of overides for components that the CVO manages. Mark
```
abhinavdahiya: It will be useful to mention that this should be used only for debugging purpose and can prevent rollout of upgrades as it allows the cluster to drift to un-supported state.

modules/upgrade-cluster-version-parameters.adoc
```diff
@@ -0,0 +1,168 @@
...
+components as `unmanaged` to prevent the CVO from creating or updating the object.
+
+|`ClusterVersionStatus`
+|The status of available updates and any in-progress updates. The current field
```
abhinavdahiya: There is no current field in the clusterversion.status

modules/upgrade-cluster-version-parameters.adoc
```diff
@@ -0,0 +1,168 @@
...
+`Partial` if the update failed or did not complete.
+
+|`ClusterVersionStatus.observedGeneration`
+|The version of the spec that is synchronized. If this value is not equal to
+the `metadata.generation` value, then the `desired` and `conditions` values
+represent a previous cluster version.
+
+|`ClusterVersionStatus.versionHash`
+|A representation of the content that the cluster uses during upgrade.
+You cannot set this parameter value.
```
abhinavdahiya: all the status fields are owned by the CVO and you cannot change any.
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status

modules/upgrade-cluster-version-parameters.adoc
```diff
@@ -0,0 +1,168 @@
...
+
+|`ClusterVersionStatus.conditions`
+|The condition of the CVO. Conditions are valid for the current `desiredUpdate`
+only when the `metadata.generation` value matches the
+`status.generation` value.
+
+* `Available` means that the upgrade to the `desiredUpdate` value completed.
+* `Progressing` means that an upgrade is in progress.
+* `Failing` means that an update is blocked by a temporary or permanent error.
```
abhinavdahiya: It will be useful to include information on structure of each condition. most important are `Reason`, `Message`
https://github.com/openshift/api/blob/fe0c0ad070ce15d30b0a6dd20adb0ef883efe11a/config/v1/types_cluster_operator.go#L110-L115

modules/upgrade-upgrading.adoc
```diff
@@ -0,0 +1,28 @@
...
+.. If the *UPDATE STATUS* is not *Updates Available*, you cannot upgrade your
+cluster.
+.. The *DESIRED VERSION* indicates the cluster version that your cluster is running
+or is updating to.
+
+. Click *Updates Available*, select a version to update to, and click *Update*.
+The *UPDATE STATUS* changes to `Updating`, and you can review the progress of
+the Operator upgrades on the *Cluster Operators* tab.
+
+
```
abhinavdahiya: @smarterclayton do you docs on how to use `oc adm upgrade` to achieve same results?

modules/upgrade-cluster-version-parameters.adoc
kalexand-rh: Sure. Do you have some guidance for what to do if an upgrade breaks? (I don't see any in the troubleshooting docs in the installer repo.)

modules/upgrade-upgrading.adoc
kalexand-rh: The docs team is having ongoing discussions about whether to prefer the UI or CLI workflows or if we need to document both, as required. If you want the `oc adm upgrade` flow documented, I can start a draft for it. I'll need information about how a GA user should know the valid release images and if it'll differ for OCP and OKD.

kalexand-rh:
Thank you @abhinavdahiya! I made some updates based on your review. If you and/or @smarterclayton want the CLI upgrade path documented, I'm happy to draft it up. I'll need to know how we want both OCP and OKD users to get the right packages.

modules/upgrade-cluster-version-parameters.adoc
```diff
@@ -0,0 +1,179 @@
...
+
+|`ClusterVersionSpec.desiredUpdate`
+|Optional. The cluster version that you want. If the current and intended
+version do not match, an upgrade starts. Set only a version from the list of
+recommended updates or your upgrade might fail.
+If an update exists with a version in the `availableUpdates` or history,
+you can specify the version field without specifying the image.
+
+If an upgrade fails, the Operator stops and reports the status of the failing
+component. Rolling your cluster back to a previous version is not supported.
```
jiajliu: @kalexand-rh @abhinavdahiya Do we plan to state how to restore the cluster when hit a failed/partial upgrade since "Rolling your cluster back to a previous version is not supported" in this doc?

modules/upgrade-upgrading.adoc
jiajliu: The same question here, if we will have some steps on how to do upgrade through CLI.

yapei: Upgrading steps from web console looks good, @jiajliu

modules/upgrade-cluster-version-parameters.adoc
kalexand-rh: If @abhinavdahiya can provide some information, I'm happy to write another topic to describe how to recover from a failed upgrade.

modules/upgrade-cluster-version-parameters.adoc
```diff
@@ -0,0 +1,179 @@
...
+
+|`ClusterVersionSpec.upstream`
+|Optional. Specify the preferred update server. If you do not specify a value,
+the CVO uses the best update server for the cluster and region.
+
+|`ClusterVersionSpec.channel`
+|Optional. Specify the channel to use to apply non-standard updates to the
```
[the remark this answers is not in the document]
kalexand-rh: The comments in the go code say that it's optional: https://github.com/openshift/api/blob/master/config/v1/types_cluster_version.go#L58
@abhinavdahiya, will you please check?

modules/upgrade-cluster-version-parameters.adoc
```diff
@@ -0,0 +1,179 @@
...
+|`ClusterVersionStatus.observedGeneration`
+|The version of the spec that is synchronized. If this value is not equal to
+the `metadata.generation` value, then the `desired` and `conditions` values
+represent a previous cluster version.
+
+|`ClusterVersionStatus.versionHash`
+|A representation of the content that the cluster uses during upgrade.
+You cannot set this parameter value.
+
+|`ClusterVersionStatus.conditions`
```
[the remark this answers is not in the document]
kalexand-rh: @jiajliu, I have information about RetrievedUpdates at [line 168](https://github.com/openshift/openshift-docs/pull/13993/files#diff-b3ac175ac8ca5c90fd2edd5ffa94d528R168). Do I need to move it?

modules/upgrade-cluster-version-parameters.adoc
```diff
@@ -0,0 +1,179 @@
...
+The `ClusterVersionStatus.conditions` value displays both the reason that the
+cluster entered its current condition and a message that provides more
+information about the condition.
+
+|`ClusterVersionStatus.availableUpdates`
+|The list of appropriate updates for the cluster. This list is empty if no
+updates are recommended, the update service is unavailable, or you specified
+an invalid channel.
+
+|`UpdateState`
```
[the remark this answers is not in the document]
kalexand-rh: They were in the [go code comments](https://github.com/openshift/api/blob/master/config/v1/types_cluster_version.go#L122). @abhinavdahiya, where should I move this information?

modules/upgrade-cluster-version-parameters.adoc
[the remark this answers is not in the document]
jiajliu: @kalexand-rh 
I think so, based my test result currently. There is not a signal item named RetrievedUpdates. But depends on the last comment,
> From UpdateState to the last ClusterVersionList, I did not see them shown in clusterversion config.

modules/upgrade-cluster-version-parameters.adoc
```diff
@@ -0,0 +1,179 @@
...
+cluster.
+
+[IMPORTANT]
+====
+The default channel contains stable updates. Do not specify a
+`ClusterVersionSpec.channel` value on production clusters.
+====
+
+|`ClusterVersionSpec.overrides`
+|Optional. A list of overides for components that the CVO manages. Mark
```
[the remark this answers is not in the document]
kalexand-rh: @abhinavdahiya, do you have more information about this parameter that I can include?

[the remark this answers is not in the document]
abhinavdahiya: https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusterversion.md#setting-objects-unmanaged

modules/upgrade-upgrading.adoc
```diff
@@ -0,0 +1,28 @@
...
+// * upgrading/upgrading-aws.adoc
+
+[id='upgrade-upgrading-{context}']
+= Updating a cluster
+
+If updates are available, you can update your cluster from the web console.
+
+.Prerequisites
+
+* Access to the web console as a user with `admin` privileges.
```
sferich888: Should we link to the section that defines how to give a user admin privileges?

modules/upgrade-upgrading.adoc
sferich888: I think we should provide CLI and UI flows (as not all users are alike), your UI users will be UI users, but your CLI users may also us the UI some of the time. 

Starting with the UI is fine, but customers who want more power and control will switch to the CLI as they become familiar and begin automating steps.

modules/update-service-overview.adoc
```diff
@@ -1,28 +1,35 @@
...
-based on current component versions and information in the graph. If you
-configure it to do so, the {product-title} update service sends the release artifacts that it needs to
-perform the upgrade to your image registry, and the Cluster Version Operator
-upgrades your cluster. By accepting automatic updates, you can automatically
+The {product-title} update service is the hosted service that provides over the
+air updates to both {product-title} and {op-system-first}. It provides a graph,
+or diagram that contain _vertices_ and the _edges_ that connect them, of
+component Operators. The edges in the graph show which versions you can safely
+upgrade to, and the vertices are update payloads that specify the intended state
+of the managed cluster components.
```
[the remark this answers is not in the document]
kalexand-rh: @crawford, is there a way to view the update graph?

modules/update-service-overview.adoc
```diff
@@ -1,28 +1,35 @@
...
+
+The Cluster Version Operator (CVO) in your cluster checks with the
+{product-title} update service to see the valid upgrades and upgrade paths based
+on current component versions and information in the graph. When you request an
+upgrade, the {product-title} CVO uses the release image for that upgrade to
+update your cluster. The release artifacts are hosted in Quay as container
```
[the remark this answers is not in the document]
kalexand-rh: I'm making a subtask to have this section reviewed again: https://jira.coreos.com/browse/OSDOCS-346

modules/upgrade-cluster-version-parameters.adoc
```diff
@@ -0,0 +1,179 @@
...
+the CVO uses the best update server for the cluster and region.
+
+|`ClusterVersionSpec.channel`
+|Optional. Specify the channel to use to apply non-standard updates to the
+cluster.
+
+[IMPORTANT]
+====
+The default channel contains stable updates. Do not specify a
+`ClusterVersionSpec.channel` value on production clusters.
```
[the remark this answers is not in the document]
kalexand-rh: If you want one, absolutely. Maybe add something like:
> If you update your cluster from a different channel without explicit direction from Red Hat support, your cluster is no longer supported.

modules/upgrade-cluster-version-parameters.adoc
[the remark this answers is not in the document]
kalexand-rh: @sferich888, can you think of an instance where you'd want users to disable CVO management of the Operators that make up the cluster? If it's not showing up in the list of overrides and we don't want users to use it, it might be better to remove it from the doc.

modules/upgrade-upgrading.adoc
kalexand-rh: I'm adding a link to the assembly - we currently cannot add internal links to individual modules due to tooling constraints.

modules/update-service-overview.adoc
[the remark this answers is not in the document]
crawford: There is currently no easy way to visualize the graph. We will need something eventually.

modules/upgrade-upgrading.adoc
kalexand-rh: That's fair @sferich888. I'm happy to document the CLI flow as soon as update's working as intended.

modules/upgrade-cluster-version-parameters.adoc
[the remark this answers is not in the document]
abhinavdahiya: > From UpdateState to the last ClusterVersionList, I did not see them shown in clusterversion config

WDY mean by "I did not see them", are you talking about the clusterversion object in the cluster. The object you view in the cluster is serialized form of `ClusterVersion`. Fields that are missing could be effect of `omitempty`..

modules/upgrade-cluster-version-parameters.adoc
```diff
@@ -0,0 +1,179 @@
...
+cannot modify them.
+
+|`ClusterVersionStatus.desired`
+|The version that the cluster is reconciling towards. In clusters that have not
+fully initialized, this value displays the available image or tag version data.
+
+|`ClusterVersionStatus.history`
+|A list of the most recent versions applied to the cluster. This value might be
+empty during cluster initialization and is updated when a new update is applied.
+The list is ordered by date, where the newest update is first in the list.
```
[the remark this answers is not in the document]
abhinavdahiya: "This value might be empty during cluster initialization and is updated when a new update is applied."

The API binding is the CVO updates this list when it applies an update. not when cluster is initialized. CVO **might** update the list when it decides installation is complete for initial cluster, but that's not API boundable... hopefully that makes sense.

modules/update-service-overview.adoc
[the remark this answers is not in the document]
kalexand-rh: https://jira.coreos.com/browse/CORS-1029

modules/upgrade-cluster-version-parameters.adoc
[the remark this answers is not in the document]
sferich888: I can't but that doesn't mean we shouldn't document the options. 

It means we need to put red flags around this saying not to use it.

modules/upgrade-cluster-version-parameters.adoc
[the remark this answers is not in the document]
jiajliu: > WDY mean by "I did not see them", are you talking about the clusterversion object in the cluster. The object you view in the cluster is serialized form of `ClusterVersion`. Fields that are missing could be effect of `omitempty`..

Yes, I'm talking about the clusterversion object in the cluster, which can be read by users directly. I thought it was about the clusterversion object. And take "UpdateState" for example, 
'''
Whether or not cluster updates succeeded.
CompletedUpdate means that an update succeeded.
PartialUpdate means that an update is in progress or did not finish.
'''
According to this description, it should be CompletedUpdate or PartialUpdate after a cluster update. It will mislead users to check whether upgrade succeed or not through "UpdateState" item, but they can not find it in clusterversion object. 
This was the reason why I raised them out. If this section "ClusterVersion parameters" is not about the clusterversion object, but the API used by CVO. Then @kalexand-rh @abhinavdahiya please can ignore all my comments about "ClusterVersion parameters".

modules/upgrade-cluster-version-parameters.adoc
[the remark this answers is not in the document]
jiajliu: Yes. I got your mean. But just as my following comments(https://github.com/openshift/openshift-docs/pull/13993/#discussion_r267608179). I stand at an end user side through clusterversion object to see the "ClusterVersion parameters". But this section is from API side to guide DEV how to make the CVO manage the clusterversion. That's the key point.

modules/upgrade-cluster-version-parameters.adoc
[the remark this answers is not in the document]
jiajliu: > The object you view in the cluster is serialized form of ClusterVersion. Fields that are missing could be effect of omitempty..
@abhinavdahiya ^^ Do u think it's better to add a note about what fields would be empty to avoid the misunderstand like "UpdateState"? Or should we give another section to describe the parameters in clusterversion object when introduce the upgrade process through CLI? Currently, QE as users much depends on the CV to check the upgrade status such as succeed or failed? failed for what reason? So this doc may be very important to help users on these.

modules/upgrade-cluster-version-definition.adoc
```diff
@@ -0,0 +1,77 @@
+// Module included in the following assemblies:
```
kalexand-rh: I propose removing the table of parameters and providing an annotated YAML file instead. @jiajliu, @abhinavdahiya, what do you think?

modules/upgrade-upgrading-cli.adoc
```diff
@@ -0,0 +1,138 @@
+// Module included in the following assemblies:
+//
+// * upgrading/upgrading-cluster.adoc
+
+[id='upgrade-upgrading-web-{context}']
```
kalexand-rh: @jiajliu, after the default update server works the way it's supposed to, will the CLI process look like this?

modules/upgrade-cluster-version-definition.adoc
abhinavdahiya: I can see how that will be useful to the end users. +1

kalexand-rh: Great! Then I'll pull the table of parameters. I'm happy to add more details to enhance the update instructions if we see gaps. (I'll update the yaml and command outputs when the default upgrade server's working and I can run through all the steps.)

jiajliu: I went through the new section about cluster-version at https://github.com/openshift/openshift-docs/blob/ec09b3754165373dbdb8cff44a94c6b557577eae/modules/upgrade-cluster-version-definition.adoc.
lgtm. but one more question is that if you miss some parts of status in the yaml file? 
'''
<--snip-->
    desired:
      image: registry.svc.ci.openshift.org/ocp/release@sha256:06fdf60795e98ceb70ca82550ef2ec3fc1005f27ac3581f46cbebbfcc0b2e25d
      version: 4.0.0-0.nightly-2019-03-28-030453
    history:
    - completionTime: 2019-03-28T05:59:50Z
      image: registry.svc.ci.openshift.org/ocp/release@sha256:06fdf60795e98ceb70ca82550ef2ec3fc1005f27ac3581f46cbebbfcc0b2e25d
      startedTime: 2019-03-28T05:45:07Z
      state: Completed
      version: 4.0.0-0.nightly-2019-03-28-030453
    observedGeneration: 1
    versionHash: wKTJ5oTXq48=
'''

modules/upgrade-upgrading-cli.adoc
jiajliu: @kalexand-rh 
Could i ask if this doc is targeted for customer? or just a internal version? Cause I noticed it mostly came from doc[1] wrote to help QE do pre-release test(non-production). I'm not sure if the steps in the doc should be an official/supported way for customer doing update. So if this "upgrade draft" doc is for customer, it's better to ask dev to give an official document about it.

Such as following part, afaik, for production cluster, upstream is not needed to be changed and this internal fake upstream should not be show in the doc.
```
$ curl https://openshift-release.svc.ci.openshift.org/graph
```
And if it's for clustomer, I think following make users confused.
> For non-production clusters, you can modify the specification that describes the cluster version state that you want:

This is incomplete for customer to check if the cluster is updated or not. It should include many componants here. But if just have a basic check, maybe run "oc get clusterversion" is enough.
> After the update completes, you can confirm that the Cluster Version Operator is using the new image:

[1] https://docs.google.com/document/d/1CfsJ3wSOtHZybKFV2CF-7t3tZiEsFFjP5bCztkXb94U/edit#

modules/upgrade-cluster-version-definition.adoc
```diff
@@ -0,0 +1,77 @@
...
+cluster. If you do not change the value, the CVO uses the default channel.
++
+[IMPORTANT]
+====
+The default channel contains stable updates. Do not modify the
+`ClusterVersionSpec.channel` value on production clusters. If you update your
+cluster from a different channel without explicit direction from Red Hat
+support, your cluster is no longer supported.
+====
+<2>  A list of overides for components that the CVO manages. Mark
```
steveeJ: s/overide/override

modules/upgrade-cluster-version-definition.adoc
kalexand-rh: After a discussion with Clayton, I'm removing the Cluster Version yaml. There's not a good use case to edit in 4.1.

modules/update-service-overview.adoc
```diff
@@ -1,7 +1,7 @@
 // Module included in the following assemblies:
 //
 // * architecture/architecture.adoc
-// * upgrading/upgrading-cluster.adoc
+// * updating/updating-cluster.adoc
```
kalexand-rh: @abhinavdahiya, does this conceptual information about updates still sound right to you, or does it need to change?

modules/upgrade-cluster-version-definition.adoc
jiajliu: > After a discussion with Clayton, I'm removing the Cluster Version yaml. There's not a good use case to edit in 4.1.

https://github.com/openshift/openshift-docs/blob/40bb56d720556cf3db30d088d39abcad1e25efd1/modules/upgrade-cluster-version-definition.adoc
I notice this section still existed. And several issues:
1. no "namespaces" in metadata?
2. "channel" should be "stable-4.1"
3. there is another condition "RetrievedUpdates"
  - lastTransitionTime: 2019-05-22T07:13:26Z
    status: "True"
    type: RetrievedUpdates

modules/upgrade-upgrading-cli.adoc
```diff
@@ -0,0 +1,147 @@
...
+
+[IMPORTANT]
+====
+If you installed the OpenShift 4 Beta Drop 5, you can update to {product-title}
+{product-version}.
+====
+
+.Prerequisites
+
+* Install the OpenShift Command-line Interface (CLI), commonly known as `oc`.
```
jiajliu: Need install/update cli to target upgraded version.

modules/upgrade-upgrading-cli.adoc
```diff
@@ -0,0 +1,147 @@
...
++
+----
+$ oc get clusterversion
+
+NAME      VERSION     AVAILABLE   PROGRESSING   SINCE     STATUS
+version   4.1.0-rc.4   True        False         24h       Cluster version is 4.1.0-rc.4
+----
++
+If your `VERSION` is `4.1.0-rc.4`, you installed OpenShift 4 Beta Drop 5, and
+you can update to {product-title} {product-version}.
```
jiajliu: I got "beta drap 5 should be rc.3" from mail "Announcing OpenShift 4.1 - Beta 5 [4.1.0-rc.3]. please double confirm about the rc-beta relation.

modules/upgrade-upgrading-cli.adoc
```diff
@@ -0,0 +1,147 @@
...
+----
+
+. View the available updates and note the version number of the update that
+you want to apply:
++
+----
+$ oc adm upgrade
+
+Cluster version is 4.0.0-0.nightly-2019-01-25-200832
+Updates:
```
jiajliu: The payload version in your example may confuse users due to there are mixed prerelease(nightly) and release(rc) version. And they can not be mixed according to current update process. Could u update them to with either prerelease or release payload.

modules/upgrade-upgrading-cli.adoc
```diff
@@ -0,0 +1,147 @@
...
+  {
+    "completionTime": "2019-05-20T15:47:21Z",
+    "image": "registry.svc.ci.openshift.org/ocp/ocp-release@sha256:6f4cf2db7e63c4dba54496a72b83fec22c49293b520ff0cdb78f1e38b23f1ccb",
+    "startedTime": "2019-05-20T15:23:29Z",
+    "state": "Completed",
+    "verified": false,
+    "version": "4.1.0-rc.4"
+  }
+]
+----
```
jiajliu: Refer to https://github.com/openshift/openshift-docs/pull/13993/files#r286777865

modules/upgrade-upgrading-cli.adoc
```diff
@@ -0,0 +1,147 @@
...
+
+. After the update completes, you can confirm that the cluster version has
+updated to the new version:
++
+----
+$ oc get clusterversion
+
+NAME      VERSION     AVAILABLE   PROGRESSING   SINCE     STATUS
+version   4.1.0       True        False         2m        Cluster version is 4.1.0
+----
```
jiajliu: Refer to https://github.com/openshift/openshift-docs/pull/13993/files#r286777865

modules/updating-clear.adoc
```diff
@@ -0,0 +1,13 @@
+// Module included in the following assemblies:
+//
+// * upgrading/upgrading-cluster-cli.adoc
+
+[id="updating-clear_{context}"]
+= Recovering when an update fails before it is applied
+
+If an update fails before it is applied, such as when the version that you
```
jianlinliu: How to define "before it is applied", this is very important for customer's understanding here.
QE try upgrade to a version that can not be found, oc client would throw error directly, no any history for this operation in clusterversion. `oc adm upgrade clear` would take no any effect.
QE also try upgrade to a version that is not signed, upgrade failed upon image validation with "the image may not be safe to use" error.  `oc adm upgrade clear` can not help me clear the error, would take no any effect either.
From the above scenario, I did not see what the command is used for.

modules/upgrade-upgrading-cli.adoc
kalexand-rh: @wking, do you have suggestions about getting better command output for upgrades before GA? I never see anything helpful in the stable channel.

modules/updating-clear.adoc
kalexand-rh: @smarterclayton, will you PTAL? I need to clarify when to run  `oc adm upgrade clear` to cancel a not-yet-started update. It's not working the way I expected.

updating/updating-cluster-cli.adoc
```diff
@@ -0,0 +1,26 @@
...
+====
+During the {product-title} beta period, update functionality is limited in the
+following ways:
+
+* You cannot update between different beta drops, such as from OpenShift 4 Beta
+Drop 3 to OpenShift 4 Beta Drop 4.
+* The ability to update from a beta drop of {product-title} to a full version of
+{product-title} is not guaranteed.
+* After you install OpenShift 4 Beta Drop 4, you can update your cluster to
+one of the available nightly builds.
```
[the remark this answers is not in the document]
kalexand-rh: The note's just there for the beta, and I will remove it for GA. The last I heard was that you can update from Beta 7 (not 5 anymore - I need to update the note) to the GA version by using the stable channel.

modules/upgrade-upgrading-cli.adoc
wking: `stable-4.0` has some edges if you want to launch a [4.0.0-0.9 cluster][1] to see them.  Currently no 4.1 edges:

```console
$ curl -sH 'Accept:application/json' https://api.openshift.com/api/upgrades_info/v1/graph?channel=stable-4.1 | jq .
{
  "nodes": [],
  "edges": []
}
$ curl -sH 'Accept:application/json' https://api.openshift.com/api/upgrades_info/v1/graph?channel=prerelease-4.1 | jq .edges
[]
```

We're supposed to have a 4.1 upgrade at some point leading into GA, but we keep making breaking changes without operator migration support.  Maybe there will be an rc.5->rc.6 edge in `prerelease-4.1` that you can use in the next few days.

[1]: https://github.com/openshift/installer/releases/tag/v0.16.1

updating/updating-cluster-cli.adoc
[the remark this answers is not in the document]
jianlinliu: If Beta 7 is our final beta build, then we should on the same page, only support Beta 7 to GA upgrade.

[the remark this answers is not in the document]
jianlinliu: About 'nightly build', even this is for beta user, but as a customer, I do not think they know anything about nightly build. Nightly build should be only internal usage, am I wrong?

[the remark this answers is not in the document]
jianlinliu: > "You cannot update between different beta drops, such as from OpenShift 4 Beta Drop 3 to OpenShift 4 Beta Drop 4"

> "After you install OpenShift 4 Beta Drop 4, you can update your cluster to one of the available nightly builds."

Per my understanding, the 1st sentence was saying upgrade between all different versions of beta is not supported, beta 3 -> beta4 is not supported, beta 5 -> beta 6 is not supported either. While the 2nd sentence was saying user can do any upgrade since beta 4. They seem like conflicting.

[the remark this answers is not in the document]
wking: > Nightly build should be only internal testing usage, am I wrong?

There's talk of getting some/all nightlies out where customers can pull them, but yeah, at the moment there is no external access to nightlies.

modules/upgrade-upgrading-cli.adoc
kalexand-rh: You're right. And the last email I see is "Announcing OpenShift 4.1 - Beta 7 [4.1.0-rc.5]." In the replies, Sebastian Jug says that Beta 7 is still using rc.4. That matches the RHCOS bits here: https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.1/

I'm updating the PR for Beta 7 and rc.4.

modules/upgrade-cluster-version-definition.adoc
kalexand-rh: You're right. The file is still present, but it's not used in the assembly any more. It doesn't need to be published for 4.1, but we might need it for upgrades for 4.2. I updated the file.

modules/updating-clear.adoc
kalexand-rh: I'm removing this topic from the TOC until we can get feedback from Clayton about how to fix it.
I'm not deleting the file because I want to merge it and then fix it and add it to the TOC later.

kalexand-rh:
@jiajliu, will you PTAL at the last commit? 

A built version of the latest proposed changes is here: http://file.rdu.redhat.com/kalexand/052919/osdocs276/updating/updating-cluster.html

The output is the best information I have with the beta. If it looks close enough, I would like to merge this change and update the output again the day of GA based on the final Beta version.

kalexand-rh:
Merging per @vikram-redhat. I have a follow-up task to update the command output to reflect a Beta 7 to GA update as soon as the info is available.

PR #13995, opened 2019-03-06
Replaced glusterfs-registry for glusterfs_registry

luigiaparicio:
Under "Using the Installer" section
The right name for that group is with an underscore


vikram-redhat:
@jhou1  - could you confirm?

jhou1:
LGTM

PR #14026, opened 2019-03-07
BZ-1684487: Added note on how the files were generated

bergerhoffer:

Preview: http://file.rdu.redhat.com/~ahoffer/2019/BZ-1684487/applications/operator_sdk/osdk-getting-started.html#managing-memcached-operator-using-olm_osdk-getting-started

### DIFF
--- a/modules/managing-memcached-operator-using-olm.adoc
+++ b/modules/managing-memcached-operator-using-olm.adoc
@@ -64,6 +64,9 @@ $ kubectl apply -f deploy/rbac.yaml
 $ kubectl apply -f deploy/crd.yaml
 ----
 +
+NOTE: The `rbac.yaml` and `crd.yaml` files were generated into the `deploy/`
+directory by the Operator SDK when you built the Memcached Operator.
++
 Because the OLM creates Operators in a particular namespace when a manifest is
 applied, administrators can leverage the native Kubernetes RBAC permission model
 to restrict which users are allowed to install Operators.


modules/managing-memcached-operator-using-olm.adoc
```diff
@@ -64,6 +64,10 @@ $ kubectl apply -f deploy/rbac.yaml
 $ kubectl apply -f deploy/crd.yaml
 ----
 +
+NOTE: The `rbac.yaml` and `crd.yaml` files were generated into the `deploy/`
+directory by the Operator SDK in the previous section when the Memcached
```
kalexand-rh: I'd consider removing "in the previous section."

kalexand-rh (commented):
LGTM!

modules/managing-memcached-operator-using-olm.adoc
bergerhoffer: Yeah I wasn't sure about that, but I wasn't sure how else to make it clear that they needed to complete that other section (without linking :) ). And also earlier in this file it had already made a reference to "the previous section". wdyt?

kalexand-rh: I think that the prereq to have built the memcached operator (line 23) covers it. Assemblies are meant to be 1:1 mappings to user stories, so I think it's fair to assume that the user will have completed the previous sections, especially when it's reinforced by the prereqs. (I'm not a fan of the other "previous section" reference, either, for what it's worth.)

bergerhoffer: Yeah, it's just that this whole BZ was because it was felt like someone might come right into this section without really completing the above section. But I'll go ahead and remove the "previous section" reference. Thanks.

bergerhoffer: @kalexand-rh I removed it and also reworded to "you built" so that hopefully makes it more clear that the user actually had to do something else. Let me know if it looks okay now.

kalexand-rh: That sounds good to me. Thank you!

jianzhangbjz:
LGTM, thanks!

PR #14028, opened 2019-03-07
CNV-928: Masquerade mode and Known Issue

ousleyp:
https://jira.coreos.com/browse/CNV-928

### DIFF
--- a/cnv_users_guide/cnv_users_guide.adoc
+++ b/cnv_users_guide/cnv_users_guide.adoc
@@ -54,6 +54,10 @@ include::modules/cnv_openshift_commands.adoc[leveloffset=+2]
 include::modules/cnv_virtctl_commands.adoc[leveloffset=+2]
 include::modules/cnv_ensure_correct_project.adoc[leveloffset=+2]
 
+[[cnv_userguide_configuring_networking]]
+=== Configuring networking for virtual machines
+include::modules/cnv_configuring_masquerade.adoc[leveloffset=+2]
+
 [[cnv_userguide_importing_vms_openshift]]
 === Importing and uploading virtual machines and disk images
 include::modules/cnv_upload_virtctl.adoc[leveloffset=+2]
--- a/modules/cnv_configuring_masquerade.adoc
+++ b/modules/cnv_configuring_masquerade.adoc
@@ -1,0 +1,35 @@
+[[configuring-masquerade]]
+=== Configuring masquerade mode
+
+You can use masquerade mode to hide a virtual machine's outgoing traffic behind
+the pod IP address. Masquerade mode uses Network Address Translation (NAT) to
+connect virtual machines to the pod network backend through a Linux bridge.
+
+Enable masquerade mode and allow traffic to enter the virtual machine by
+editing your virtual machine configuration file.
+
+.Prerequisites
+
+* The virtual machine must be configured to use DHCP to acquire IPv4 addresses.
+
+.Procedure
+
+. Edit the `interfaces` spec of your virtual machine configuration file:
++
+[source,yaml]
+----
+kind: VM
+spec:
+  domain:
+    devices:
+      interfaces:
+        - name: red
+          masquerade: {} <1>
+          ports:
+            - port: 80 <2>
+  networks:
+  - name: red
+    pod: {}
+----
+<1> Connect using masquerade mode
+<2> Allow incoming traffic on port 80
--- a/cnv_release_notes/cnv_release_notes.adoc
+++ b/cnv_release_notes/cnv_release_notes.adoc
@@ -143,6 +143,9 @@ pod automatically restarts, and the kubelet reports the correct number of
 available KVM devices.
 ====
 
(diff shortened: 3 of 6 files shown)


SchSeba (approved):
LGTM thanks!

ousleyp:
@openshift/team-documentation PTAL, thanks!!

cnv_users_guide/cnv_users_guide.adoc
```diff
@@ -54,6 +54,10 @@ include::modules/cnv_openshift_commands.adoc[leveloffset=+2]
 include::modules/cnv_virtctl_commands.adoc[leveloffset=+2]
 include::modules/cnv_ensure_correct_project.adoc[leveloffset=+2]
 
+[[cnv_userguide_configuring_networking]]
+=== Configuring networking for virtual machines
```
kalexand-rh: Do you need headings in both the assembly and the module?

modules/cnv_configuring_masquerade.adoc
```diff
@@ -0,0 +1,35 @@
+[[configuring-masquerade]]
```
kalexand-rh: You'll eventually need to make your module headings reusable and set all the initial headings to H1.

modules/cnv_configuring_masquerade.adoc
```diff
@@ -0,0 +1,35 @@
...
+Enable masquerade mode and allow traffic to enter the virtual machine by
+editing your virtual machine configuration file.
+
+.Prerequisites
+
+* The virtual machine must be configured to use DHCP to acquire IPv4 addresses.
+
+.Procedure
+
+. Edit the `interfaces` spec of your virtual machine configuration file:
```
kalexand-rh: A user should know how to find this, right?

modules/cnv_configuring_masquerade.adoc
```diff
@@ -0,0 +1,35 @@
...
+      interfaces:
+        - name: red
+          masquerade: {} <1>
+          ports:
+            - port: 80 <2>
+  networks:
+  - name: red
+    pod: {}
+----
+<1> connect using masquerade mode
```
kalexand-rh: Initial caps, please

Are both of these values static, or can you provide more information about how a user should set them?

kalexand-rh (commented):
Couple of picks and questions, but it overall LGTM.

cnv_users_guide/cnv_users_guide.adoc
ousleyp: I wasn't sure where it should go in the overall assembly, so I created a separate networking section. It's not necessarily permanent, though, because we're planning to restructure the guides for CNV 2.0.

modules/cnv_configuring_masquerade.adoc
ousleyp: Yup, I'm aware that we're lagging behind in that respect. We'll probably do all of them at once.

modules/cnv_configuring_masquerade.adoc
ousleyp: The file? Yes.

modules/cnv_configuring_masquerade.adoc
ousleyp: I'll fix the capitalization - oops. Thanks for pointing that out :)

I think those values are both static, but I'll try to find out more.

Thanks for the review, Kathryn!

modules/cnv_configuring_masquerade.adoc
```diff
@@ -0,0 +1,35 @@
...
+      interfaces:
+        - name: red
+          masquerade: {} <1>
+          ports:
+            - port: 80 <2>
+  networks:
+  - name: red
+    pod: {}
+----
+<1> Connect using masquerade mode
```
myakove: I think using the word 'connect' is misleading, I prefer interface type here.

dankenigsberg: interface type is something different: virito, e1000, etc. so I wouldn't use it here.

We sometime use "bind", but I don't really know why @myakove find "connect" misleading.

ousleyp: @myakove The "Connect using masquerade mode" was taken straight from the upstream docs, IIRC. I will leave it as-is for now, based on the response from @dankenigsberg - but let me know if you can clarify further.

@myakove Are there any other concerns from a QE perspective? I'm not sure how things get marked "Verified" in JIRA since the next status is "Done." I'm eager to merge this PR since it needs to go in the 1.4 docs. :) Thanks for your help.

myakove:
Ok be me, I can live with connect :).

modules/cnv_configuring_masquerade.adoc
ousleyp: @myakove Just saw that you created a sub-task, so ignore my question about how things get marked "Verified." I assume that when you've finished testing this story, you will mark that sub-task "Done" which will indicate to me that I can merge the PR. Let me know if that's incorrect. Thanks!

alexxa (approved):

PR #14035, opened 2019-03-08
Do we need Deleting nodes from a cluster in OpenShift 4

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1685360


mburke5678:
@jhou1 Can you take a look at this PR or assign to another QE person?

PR #14054, opened 2019-03-11
Migrating Users and Roles content to Authentication.

huffmanca:
Moving the assemblies from `Users and Roles` into `Authentication` per discussion with Ali and Vikram.

This is for OS 4.0.


huffmanca:
@openshift/team-documentation , would someone mind reviewing this? Essentially seeing if these topics are in the correct place, or if they should be adjusted to a different location under Authentication.

kalexand-rh (commented):
LGTM

PR #14055, opened 2019-03-11
replacing _{context} with -{context}

kalexand-rh:
@vikram-redhat


kalexand-rh:
We agreed to this in the meeting, and Travis is happy, so I'm merging.

PR #14058, opened 2019-03-11
the description of scheduler policy-configmap's project is conflicting

mburke5678:


mburke5678:
@sjenning PTAL or assign to a team member.
I addressed the concern in the BZ and update the scheduler docs to match 4.0.
The major changes are the new `policy.cfg` file that the user creates/edits and the new format for the ` policy.cfg` section in the generated configmap.

mburke5678:
@lyman9966 Please take a look

lyman9966:
@mburke5678 ,
in file "modules/nodes-scheduler-default-creating.adoc", line 61, should be: 
name: scheduler-policy

PR #14063, opened 2019-03-12
bug 1622044 amended auditFilePath directories

sheriff-rh:
bug 1622044:
https://bugzilla.redhat.com/show_bug.cgi?id=1622044

Previously ack'ed by QA, merging.


sheriff-rh:
Correction - bug was only for 3.10+; removing 3.9 label.

PR #14080, opened 2019-03-13
Removed stray <<HEAD from previous merge error

mburke5678:


mburke5678:
QE not required. Formatting change only.

PR #14095, opened 2019-03-13
Updates to build strategies and creating images

bmcelvee:
Moving modules to more appropriate locations. Modules will need editing/updates. 

@bparees - FYI


bmcelvee:
This PR is only moving content and cleaning up formatting it went through stage 1 reviews. No QE required.

PR #14101, opened 2019-03-13
bug 1685731 - broken link to reference example

bgaydosrh:
Fixing bug https://bugzilla.redhat.com/show_bug.cgi?id=1685731


ousleyp:
LGTM

vikram-redhat:
@bgaydosrh I mentioned it in our chat but will leave here for reference as well. Make sure to only merge if the build passes, not before. If the build fails, you should fix the cause and then resubmit.

cc: @ousleyp

PR #14141, opened 2019-03-15
[enterprise-3.11] Correct a command to label "logging-infra-fluentd=true"

bysnupy:
- Partial Fix: [ [Docs] three improvements requests to "Adding Hosts to an Existing Cluster"](https://bugzilla.redhat.com/show_bug.cgi?id=1669929)

- Version: `v3.11`

- Description:
  `logging-infra-fluentd` node label is required when you installed a `logging` component, and usage is not correct, the correct command is as follows. Refer [Updating labels on nodes
](https://docs.openshift.com/container-platform/3.11/admin_guide/manage_nodes.html#updating-labels-on-nodes) for more details.
~~~
# oc label node/new-node.example.com logging-infra-fluentd=true
~~~

### DIFF
--- a/admin_guide/topics/proc_adding-hosts.adoc
+++ b/admin_guide/topics/proc_adding-hosts.adoc
@@ -135,10 +135,10 @@ $ ansible-playbook [-i /path/to/file] \
     playbooks/openshift-master/scaleup.yml
 ----
 +
-. Set the node label to `logging-infra-fluentd: "true"`.
+. Set the node label to `logging-infra-fluentd=true`, if you deployed the EFK stack in your cluster.
 +
 ----
-# oc label node/new-node.example.com logging-infra-fluentd: "true"
+# oc label node/new-node.example.com logging-infra-fluentd=true
 ----
 
 . After the playbook runs,


kalexand-rh:
@mburke5678, will you PTAL?

admin_guide/topics/proc_adding-hosts.adoc
```diff
@@ -135,10 +135,10 @@ $ ansible-playbook [-i /path/to/file] \
     playbooks/openshift-master/scaleup.yml
 ----
 +
-. Set the node label to `logging-infra-fluentd: "true"`.
+. Set the node label to `logging-infra-fluentd=true`, if you installed a logging component on an existing cluster.
```
mburke5678: We call logging the "EFK" stack elsewhere in the docs. Could you use that terminology?  I was confused at first by "a logging component."
Something like:
s/if you installed a logging component on an existing cluster/if you deployed the EFK stack in your cluster.

mburke5678:
@bysnupy Thanks for the feedback. One small suggestion. Otherwise it looks good to me.

bysnupy:
Thank you for your review @mburke5678 ! I've modified as you mentioned. It's a good point, I'll use EFK stack term from now when pointing logging component in our docs. :)

admin_guide/topics/proc_adding-hosts.adoc
bysnupy: I've modified as your point. Thanks !!

vikram-redhat:
@mburke5678 can this be merged now?

PR #14163, opened 2019-03-18
CNV heading levels

aburdenthehand:
Fixing some of the legacy heading issues we've had since the conversion from kbase:
- all modules now include "Module included in the following assemblies" prefix
- all modules now single-level heading, properly offset in the assemblies
- sentence case across all titles
- removed old content no longer included in assemblies

@ousleyp - Could you please provide sanity check? Will squash after review. 
Local build:
http://file.bne.redhat.com/aburden/CNV-docs-preview/heading_levels/
@vikram-redhat - I believe you still need to tag this for me.


kalexand-rh:
@aburdenthehand, if you're fixing the levels, will you fix the anchor id format, too? They should all take the form `[id="<description>-{context}"]`.
You might be able to do two find/replaces on your modules folder to replace `[[` with `[id="` and `]]` with `-{context}"]` to get ~90% there.

ousleyp:
Yay, @aburdenthehand, this is awesome! Thank you *so much* for jumping on this. 

I agree with Kathryn that now would be a good time to fix the anchor ids. I glanced over the PR and it LGTM so far, but I can give it another review after the anchor ids have been fixed. 

I also just merged two PRs into 3.11 so there are two new files to pull in (cnv_viewing_guest_ip_addresses.adoc and cnv_configuring_masquerade.adoc), and one new section in the User's Guide (it currently reads "[[cnv_userguide_configuring_networking]]
=== Configuring networking for virtual machines"). The masquerade module is totally old-style. The guest IP addresses module has the right anchor format and the "module included in assemblies" comment but it doesn't have the right header level (sorry about that!). 

Let me know if I can help further.

vikram-redhat:
@aburdenthehand - regarding your last commit, it is better to add the context now then realize later that you can't use them because you haven't added it. There is no issue with adding context now, and it will make it easier for you to migrate to mod docs when you do move them to 4.0.

aburdenthehand:
@kalexand-rh @ousleyp 
Anchor IDs are all updated. Adding '-{context}' across the board broke a bunch of xrefs and made me wonder if it doesn't make more sense to add that suffix to the anchor when it's needed, as no CNV module is currently re-used, so I've returned all IDs to '[id="<description>"]'.

If that's a bad idea, I'll revert that final commit (8460d48) to bring back the '-{context}' suffix (easier to do now than later) and go ahead and update all the necessary xrefs.

aburdenthehand:
@vikram-redhat - sorry, replied without seeing your message. 
If that's the case then I will rollback that last commit to return the '-{context}' suffix to all CNV anchors and update the xrefs that will break.

kalexand-rh:
@aburdenthehand, you shouldn't have any xrefs in modules.

aburdenthehand:
After discussing various impacts with @vikram-redhat, I have rolled this back so that it is just an update to bring the header levels in line. The domino effect of updating the anchors meant either updating all the xrefs (only valid for 1.4, and then dumped for 2.0+), or just dumping them now (which negatively impacts the current navigability of the 1.4 docs). 

I created another branch which retains the complete anchor update and the removal of all the xrefs, which we can use for 2.0 after 1.4 gets locked away.

aburdenthehand:
minor update relating to #14195

kalexand-rh:
Thanks @aburdenthehand!

kalexand-rh:
@vikram-redhat, @aburdenthehand, is this ready to merge?

aburdenthehand:
@kalexand-rh, yes please

PR #14165, opened 2019-03-18
replacing ' in anchor ids

kalexand-rh:


applications/pruning-objects.adoc
```diff
@@ -30,5 +30,5 @@ include::modules/pruning-hard-pruning-registry.adoc[leveloffset=+1]
 include::modules/pruning-cronjobs.adoc[leveloffset=+1]
 .Additional resources
 - xref:../nodes/jobs/nodes-nodes-jobs.adoc#nodes-nodes-jobs_nodes-nodes-jobs[Running tasks in pods using jobs]
-- xref:../applications/quotas/quotas-setting-across-multiple-projects.adoc#setting-quotas-across-multiple-projects[Resource quotas across multiple projects]
+- xref:../applications/quotas/quotas-setting-across-multiple-projects.adoc[Resource quotas across multiple projects]
```
vikram-redhat: @kalexand-rh  this is removing an anchor id?

vikram-redhat:
@kalexand-rh - not sure why but this PR seems to be removing anchor ids as well from some xrefs. Is that intentional? Otherwise, it builds correctly locally for all distros. 

Besides the rebase that you might have to do :)

kalexand-rh:
@vikram-redhat, that was intentional. If I leave in the anchor ids, the build fails. I tested a few of the links that I stripped the anchor ids from, and they still work as expected locally.

PR #14175, opened 2019-03-19
[enterprise-3.11] Correct typo and package name for openshift-ansible

bysnupy:
- Version: `v3.11`

- Description:
  `openshift-ansible` can be updated without `*`, because `openshift-ansible` will be installed with dependencies. And `yum` is required root permission, so `$` prompt is wrong, it should be `#`.


bysnupy:
I will open again this pr to master branch.

PR #14177, opened 2019-03-19
[enterprise-3.11] Correct typo and package name for openshift-ansible

bysnupy:
- Version: `v3.11`

- Description:
  `openshift-ansible` package can be updated without specifying `*`, because it will install with dependencies.
   And `yum` is required `root` permission for installing package, so prompt should be `#`.


kalexand-rh:
/ok-to-test

@gpei, will you please take a look?

gpei:
@kalexand-rh LGTM :+1:

PR #14187, opened 2019-03-20
updating menu click syntax highlighting

kalexand-rh:

### DIFF
--- a/contributing_to_docs/doc_guidelines.adoc
+++ b/contributing_to_docs/doc_guidelines.adoc
@@ -725,16 +725,17 @@ a|Use the following command to roll back a Deployment, specifying the Deployment
 `oc rollback <deployment>`
 
 |Use single asterisks for web console / GUI items (menus, buttons, page titles, etc.).
+Use two characters to form the arrow in a series of menu items, $$->$$.
 
 a|Choose $$*Cluster Console*$$ from the list.
 
-Navigate to the $$*Operators -> Catalog Sources*$$ page.
+Navigate to the $$*Operators* -> *Catalog Sources*$$ page.
 
 Click $$*Create Subscription*$$.
 
 a|Choose *Cluster Console* from the list.
 
-Navigate to the *Operators -> Catalog Sources* page.
+Navigate to the *Operators* -> *Catalog Sources* page.
 
 Click *Create Subscription*.
 


mburke5678:
@kalexand-rh Is the `$$` an escape character to retain the `*` in the text? If so, LGTM

bergerhoffer:
@kalexand-rh Thank you for this! I've seen a lot of places where the asterisks only surround the outside of the whole path and that's bothered me :). Also, at least the guidance already has this, but I've seen places that use an arrow character instead of "->", so we might want to point this out as well.

kalexand-rh:
@mburke5678, you're right. I think that I need to escape more characters now. (If it doesn't look right when I merge, I'll drop in another fix.)

@bergerhoffer, me, too! I've seen a mix of use, but I think that enclosing the UI elements in asterisks individually will help us update the docs when the UI changes. I added more guidance for the arrow.

contributing_to_docs/doc_guidelines.adoc
```diff
@@ -725,16 +725,17 @@ a|Use the following command to roll back a Deployment, specifying the Deployment
 `oc rollback <deployment>`
 
 |Use single asterisks for web console / GUI items (menus, buttons, page titles, etc.).
+Use two characters to form the arrow in a series of menu items, $$->.
```
bergerhoffer: The "$$" is rendering here. You need to add "$$" after the arrow characters.

contributing_to_docs/doc_guidelines.adoc
```diff
@@ -725,16 +725,17 @@ a|Use the following command to roll back a Deployment, specifying the Deployment
...
 
 a|Choose $$*Cluster Console*$$ from the list.
 
-Navigate to the $$*Operators -> Catalog Sources*$$ page.
+Navigate to the $$*Operators* $$-> $$*Catalog Sources*$$ page.
```
bergerhoffer: Actually, adding the $$'s in the middle there is causing the arrow to render. You should only need the $$'s surrounding the entire thing:

```
"$$*Operators* -> *Catalog Sources*$$"
```

bergerhoffer (commented):
The recent updates caused just a few issues.

PR #14195, opened 2019-03-21
fixing the build

kalexand-rh:
https://github.com/openshift/openshift-docs/pull/14101 broke the build, so I'm patching it.

@aburdenthehand, will you PTAL at this anchor as part of your work on https://github.com/openshift/openshift-docs/pull/14163?


aburdenthehand:
@kalexand-rh Thanks for that. Updated #14163

PR #14208, opened 2019-03-22
bug 1690748 Revert "bug 1684100 removing the Paris AWS region"

kalexand-rh:
This reverts commit 58689b54a3c168c79c71b45224d89841da0a5c02.

Adding Paris region back per https://bugzilla.redhat.com/show_bug.cgi?id=1690748


kalexand-rh:
This change was approved on the source bug.

PR #14210, opened 2019-03-22
Add Kibana Proxy settings per LOG-276

mburke5678:
https://jira.coreos.com/browse/LOG-276

### DIFF
--- a/modules/efk-logging-kibana-limits.adoc
+++ b/modules/efk-logging-kibana-limits.adoc
@@ -9,8 +9,8 @@ Each component specification allows for adjustments to both the CPU and memory l
 
 .Prerequisite
 
-If needed, get the name of the Cluster Logging Custom Resource in the `openshift-logging` project:
-
+* If needed, get the name of the Cluster Logging Custom Resource from the `openshift-logging` project:
++
 ----
 $ oc get ClusterLogging
 NAME       AGE
@@ -28,22 +28,30 @@ $ oc edit ClusterLogging instance
 apiVersion: "logging.openshift.io/v1"
 kind: "ClusterLogging"
 metadata:
-  name: "clusterlogging"
+  name: "instance"
 
 ....
 
 spec:
-visualization:
-   type: "kibana"
-   kibana:
-     replicas: 1
-     resources: <1>
-      limits:
-        cpu:
-     memory:
-       requests:
-        cpu:
-        memory:
-----
+    visualization:
+      type: "kibana"
+      kibana:
+        replicas:
+      resources:  <1>
+        limits:
+          memory: 1Gi
+        requests:
+          cpu: 500m
+          memory: 1Gi
+      proxy:  <2>
+        resources:
+          limits:
+            memory: 100Mi
+          requests:
+            cpu: 100m
+            memory: 100Mi
 
-<1> Specify the CPU and memory limits as needed.
+---- 
+
+<1> Specify the CPU and memory limits to allocate for each node.
+<2> Specify the CPU and memory limits to allocate to the Kibana proxy.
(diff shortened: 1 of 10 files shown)


mburke5678:
@anpingli @jcantrill Please take a look. 
Do we need to define what a Kibana proxy is? (If so, what is a Kibana proxy?). Is it optional?

jcantrill:
> @anpingli @jcantrill Please take a look.
> Do we need to define what a Kibana proxy is? (If so, what is a Kibana proxy?). Is it optional?

It is not optional.  I'm indifferent if we should define it.  It's basically the thing which controls access to Kibana defering auth to openshift.  This allows us to provide an auth mechanism without needing to modify Kibana.

modules/efk-logging-kibana-limits.adoc
```diff
@@ -5,18 +5,20 @@
 [id="efk-logging-kibana-limits-{context}"]
 = Configure Kibana CPU and memory limits
 
-Each component specification allows for adjustments to both the CPU and memory limits. 
+Each component specification allows for adjustments to both the CPU and memory limits. You should not need to manually adjust these values as the Cluster Logging Operator sets values sufficient for your environment.
```
jcantrill: I would refrain from making this statement as the Operator yet is unable to make decisions about the deployment related to the characteristics of the cluster.  We set defaults based on experience from previous releases but admins may need to adjust them

modules/efk-logging-kibana-limits.adoc
```diff
@@ -5,18 +5,20 @@
...
 
 .Prerequisite
 
-If needed, get the name of the Cluster Logging Custom Resource from the `openshift-logging` project:
-
+* If needed, get the name of the Cluster Logging Custom Resource from the `openshift-logging` project:
```
jcantrill: When https://github.com/openshift/cluster-logging-operator/pull/123 merges the resource will be named 'instance'

modules/efk-logging-kibana-limits.adoc
```diff
@@ -5,18 +5,20 @@
...
++
 ----
 $ oc get ClusterLogging
 NAME       AGE
 instance   112m
 ----
 
+* Set cluster logging to the unmanaged state.
```
jcantrill: This is not needed in order to adjust the proxy settings.

modules/efk-logging-kibana-scaling.adoc
```diff
@@ -39,16 +39,7 @@ spec:
...
-              cpu:
-              memory:
-            requests:
-              cpu:
-              memory:
 ---- 
 
 <1> Specify the number of Kibana nodes.
-<2> Specify the CPU and memory limits for each node.
+
```
jcantrill: I don't expect the need to adjust the proxy but you could leave some note here, but it's not per node but for the proxy.

mburke5678: @jcantrill I had removed this line from the file and created a new file for memory requests. 
https://github.com/openshift/openshift-docs/pull/14210/files#diff-54aee4af720c50efd3731f628998fb35R59

modules/efk-logging-kibana-limits.adoc
mburke5678: @jcantrill What about for memory and CPU requests for Kibana (not the proxy)?

mburke5678:
@anpingli PTAL

mburke5678:
@anpingli Please take a look.

modules/efk-logging-kibana-limits.adoc
anpingli: @mburke5678  The change is in Cluster Logging Custom Resource, the cluster logging must be  in Managed state. 

The other Part LGTM

PR #14215, opened 2019-03-23
[enterprise-3.11] auditFilePath value is set to wrong path

bysnupy:
- Fix: [[DOCS] auditFilePath value is set to wrong path](https://bugzilla.redhat.com/show_bug.cgi?id=1622044)

- Version: `v3.11`

- Description:
  As of `v3.10`, `openshift` `apiserver` is running as static pod which is configured with limited mount point, such as `/var/lib/origin`, not `/var/log`.


vikram-redhat:
@sheriff-rh - PTAL.

bysnupy:
Could you PTAL @sheriff-rh ?

sheriff-rh:
@bysnupy  Hi - my sincerest apologies; I somehow missed this PR. I'm taking care of this today. Your changes look valid, so I'll merge them straight away to 3.10+ . Thanks for your input!

PR #14216, opened 2019-03-24
openshift_logging_master_public_url is no longer used

hshiina:
openshift_logging_master_public_url got unused along with
introducing oauth-proxy in 3.11.


kalexand-rh:
@anpingli, will you PTAL and confirm that this change is valid in only 3.11?

kalexand-rh:
@anpingli, will you PTAL and confirm that this change is valid in only 3.11?

anpingli:
@kalexand-rh  Yes, it is valid for v3.11

kalexand-rh:
Thank you @anpingli! Merging.

PR #14222, opened 2019-03-25
 beta drop update note

kalexand-rh:


updating/updating-cluster.adoc
```diff
@@ -7,6 +7,18 @@ toc::[]
...
 
+[IMPORTANT]
+====
+During the {product-title} beta period, update functionality is limited:
+
+* You cannot update between different beta drops, such as from OpenShift 4 Beta
+Drop 2 to OpenShift 4 Beta Drop 3.
+* The ability to update from a beta drop of {product-title} to a full version of
+is not guaranteed.
+* After you install OpenShift 4 Beta Drop 3, you can update your to the
```
vikram-redhat: typo, "update your ...."

kalexand-rh:
@abhinavdahiya, does this note sound right for the beta 3 upgrade restrictions?

updating/updating-cluster.adoc
```diff
@@ -7,6 +7,18 @@ toc::[]
...
+[IMPORTANT]
+====
+During the {product-title} beta period, update functionality is limited:
+
+* You cannot update between different beta drops, such as from OpenShift 4 Beta
+Drop 2 to OpenShift 4 Beta Drop 3.
+* The ability to update from a beta drop of {product-title} to a full version of
+is not guaranteed.
+* After you install OpenShift 4 Beta Drop 3, you can update your cluster to the
+available nightly builds.
```
abhinavdahiya: `available nightly builds` ?
@sdodson do you know if it is any nightly builds after beta3 for some specific builds ?
also how many night builds all / first x / just next one ?

sdodson: I believe it will be a set of specifically tested builds.

sdodson: From the customer's perspective they wouldn't know that, they'll only see the updates we want them to.

kalexand-rh: Right. I though I heard that it was a set of three validated daily builds, so from a customer's perspective, it'll look like nightly builds. I'm happy to change the text if you think something else would be less confusing for both internal and external audiences.

sdodson: I think it's fine as written.

kalexand-rh:

updating/updating-cluster.adoc
```diff
@@ -7,6 +7,18 @@ toc::[]
 
 You can update, or upgrade, an {product-title} cluster.
 
+[IMPORTANT]
+====
+During the {product-title} beta period, update functionality is limited:
```
bergerhoffer: * This mentions beta without a version. Should it say `During the {product-title} 4 beta period` ?
* Consider leading in to the bullets like "update functionality has the following limitations" or "update functionality is limited in the following ways"

updating/updating-cluster.adoc
```diff
@@ -7,6 +7,18 @@ toc::[]
...
+
+* You cannot update between different beta drops, such as from OpenShift 4 Beta
+Drop 2 to OpenShift 4 Beta Drop 3.
+* The ability to update from a beta drop of {product-title} to a full version of
```
bergerhoffer: This drops off. I assume should either be "a full version of {product-title}" or just remove the "of".

updating/updating-cluster.adoc
bergerhoffer: "cluster" is singular, but "builds" is plural. Suggest changing it to "you can update your cluster to one of the
available nightly builds."

kalexand-rh:
Thanks @bergerhoffer!

PR #14084, opened 2019-03-13
WIP Network operator

ariordan-redhat:
@openshift/team-documentation For peer review, please.

PR #14224, opened 2019-03-25
WIP Networking Operator - OSDOCS-177

jboxman:
- Based on https://github.com/openshift/openshift-docs/pull/14084
- https://jira.coreos.com/browse/OSDOCS-177


jboxman:
I'm using this comment https://github.com/openshift/openshift-docs/pull/14084#pullrequestreview-214001660 as the basis for changes to the content from https://github.com/openshift/openshift-docs/pull/14084.

jboxman:
It looks like d853fcd5981daa2a42dee551c0ae773c8b2c4ea9 intersects with some of this content.

danwinship:
> The preview will be availble shortly at:
> 
>     * https://issue/OSDOCS-177--ocpdocs.netlify.com/

what's the correct URL here?

modules/nw-networkoperator-crd.adoc
```diff
@@ -0,0 +1,99 @@
...
+apiVersion: networkoperator.openshift.io/v1
+kind: NetworkConfig
+spec:
+  clusterNetworks:
+  - cidr: 10.128.0.0/14 <1>
+    hostSubnetLength: 9 <2>
+  serviceNetwork: 172.30.0.0/16 <3>
+  defaultNetwork: <4>
+    type: OpenShiftSDN <5>
+    openshiftSDNConfig:
```
danwinship: you probably need a footnote on the `openshiftSDNConfig` line, clarifying that this section needs to be specified if (and only if) you want to override any of the OpenShift SDN configuration

modules/nw-networkoperator-crd.adoc
```diff
@@ -0,0 +1,99 @@
...
+`config.openshift.io` object.
+<4> `defaultnetwork`: The header for the parameters that specify the
+default pod network.
+<5> `NetworkType`: The Network plugin that is to be deployed. `OpenShiftSDN` is
+the only plugin supported in {product-name} {product-version}. This parameter is
+set in the `config.openshift.io` object.
+<6> `mode`: Configures the isolation mode for `OpenShiftSDN`. The allowed values
+are `Multitenant`, `Subnet`, or `NetworkPolicy`. The default value is
+`NetworkPolicy`.
+<7> `mtu`: The Maximum Transmission Unit (MTU) for the VXLAN overlay network.
```
danwinship: Don't expand the acronym. Nobody ever uses the full phrase (except when expanding it because the author's style guide says that they have to).

modules/nw-networkoperator-crd.adoc
```diff
@@ -0,0 +1,99 @@
...
+The default value is the MTU of the node that the cluster-network-operator is first
+run on, minus 50 bytes for overhead. If the nodes in your cluster do not have
+the same MTU, then you must set this explicitly.
```
danwinship: Talking about "the node that the CNO is first run on" may be too much info. "This is normally configured automatically, but if the nodes in your cluster do not all have the same MTU, then you must set this explicitly, to 50 less than the smallest node MTU value."

modules/nw-networkoperator-crd.adoc
```diff
@@ -0,0 +1,99 @@
...
+<8> `VXLANPort`: The port to use for all VXLAN packets. The default value is 4789.
```
danwinship: "You should not need to change this unless you are running in a virtualized environment where some nodes are already part of another VXLAN network."

modules/nw-networkoperator-crd.adoc
```diff
@@ -0,0 +1,99 @@
...
+<9> `DeployKubeProxy` specifies whether a stand-alone `kube-proxy` should
```
danwinship: There's no need to document this here; this only needs to be set when using a third-party network plugin, and we don't document how to configure third-party network plugins here.

modules/nw-networkoperator-crd.adoc
```diff
@@ -0,0 +1,99 @@
...
+//Currently only `OpenShiftSDN` consumes this.
+//+optional
+<11> `bindAddress`: The IP address for `kube-proxy` to serve on. The default value is
+`0.0.0.0`.
+<12> `iptablesSyncPeriod` The refresh period for `iptables` rules. The default value
+is `30s`.
+<13> `proxyArguments`: This section contains additional arguments to pass to the
+`kube-proxy` process.
+
+//Use External openvswitch If true, then assume the nodes already have a running
```
danwinship: This is another "you only need to do this if X" situation, but I forget what "X" is

jboxman:
@danwinship, thanks for the comments. I'm reviewing them now.

modules/nw-modifying-nwoperator-config-startup.adoc
```diff
@@ -0,0 +1,149 @@
...
+----
+$ openshift-install create manifests
+----
++
+The installer creates a `manifests` directory that contains `.yaml` files.
+Two of these files describe the configuration that the Network Operator needs.
++
+* `cluster-network-01-crd.yml` contains the Custom Resource Definitions (CRD)
+for the `Network.config.openshift.io` and
+`NetworkConfig.networkoperator.openshift.io` Custom Resources (CRs).
```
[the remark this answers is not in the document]
jboxman: Okay, that was my impression when I looked at it. I'll omit discussion of it.

modules/nw-modifying-nwoperator-config-startup.adoc
```diff
@@ -0,0 +1,149 @@
...
+// * networking/configuring-network-operator.adoc
+
+[id="modifying-nwoperator-config-startup-{context}"]
+= Modifying advanced network configuration parameters
+
+You can modify the advanced network configuration parameters in the
+`NetworkConfig.networkoperator.openshift.io` Custom Resource (CR) object
+before you start the cluster.
+This procedure uses the `install-config.yaml` file that was generated
+when you modified the basic network configuration.
```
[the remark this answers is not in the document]
jboxman: > You can just start with create manifests and then edit them there.

@danwinship, do you mean I must start with creating an `install-config.yaml` with the installer, and then proceed to create the manifests?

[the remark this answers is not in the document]
danwinship: doh! that should have said "you *can't* just start with `create manifests`"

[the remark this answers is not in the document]
danwinship: So, eg, to change `serviceNetwork`, you have to `create install-config`, edit it there, and then `create manifests`. If you skip the install-config step, you'll end up with a broken cluster.

jboxman:
@openshift/team-documentation 

PTAL

This PR adds content for working with the Network Operator at installation time and at runtime. This Operator can be configured via the `install-config.yaml`, but further configuration must be done by modifying a manifest file. The parameters from `install-config.yaml` appear in both configs and those provided in `install-config.yaml` supersede identical ones provided in the manifest CR, so this complicates explanation.

@kalexand-rh, does it make sense to try to incorporate one or both levels of configuration for this Operator into the customized AWS installation assembly?

Thanks.

fyi @vikram-redhat

jboxman:
I need to update this PR to modify the procedure for the Operator configuration. It is unnecessary to duplicate the fields for networking that are specified during the basic custom install.

kalexand-rh:
@jboxman, I might have more time to look tomorrow, but I think that you need just one assembly that explains how to customize the networking operator and install the cluster.

modules/network-config-cr.adoc
```diff
@@ -0,0 +1,42 @@
+// Module included in the following assemblies:
+//
+// * networking/configuring-network-operator.adoc
+[id="network-default-cr-{context}"]
+= Cluster network configuration custom resource
+
+The basic configuration for the `Network Operator` is set in the `Network`
```
squeed: A bit of terminology: the Kubernetes object is "Network.config.openshift.io". "Nework" is the object name. "config.openshift.io" is the group. These happen to be defined as CRDs but that's not important for end-users to know.

So I would word this as something like:

The cluster's network configuration is set in the "Network.config.openshift.io" object. All fields in the "Spec" subresource are mandatory.

The object is automatically created by the openshift-installer.

modules/network-modify-config.adoc
```diff
@@ -0,0 +1,79 @@
+// Module included in the following assemblies:
+// * networking/configuring-network-operator.adoc
+
+[id="network-modifying-config-{context}"]
+= Modifying network configuration in a running cluster
+
+The Network Operator reconciles the state of the cluster against a desired
+configuration.
+
+You can modify only `kubeProxy` configuration parameters in a running cluster.
```
squeed: You can also create new additionalNetworks

modules/network-modify-config.adoc
```diff
@@ -0,0 +1,79 @@
...
+* `oc` command installed.
+* Logged in to a running cluster.
+
+.Procedure
+
+. Use the following command to edit the `Network.operator.openshift.io` Custom
+Resource (CR):
++
+----
+$ oc edit network
```
squeed: `oc edit network.operator.openshift.io cluster`

modules/network-modify-config.adoc
```diff
@@ -0,0 +1,79 @@
...
+[source,yaml]
+----
+# Please edit the object below. Lines beginning with a '#' will be ignored,
+# and an empty file will abort the edit. If an error occurs while saving this file will be
+# reopened with the relevant failures.
+#
+apiVersion: config.openshift.io/v1
+kind: Network
+metadata:
+  creationTimestamp: 2019-04-02T17:19:32Z
```
squeed: remove the creationTimestamp, generation, resourceVersion, selfLink, and uid - those are internal k8s fields.

modules/network-modify-config.adoc
```diff
@@ -0,0 +1,79 @@
...
+  - cidr: 10.128.0.0/14
+    hostPrefix: 23
+  clusterNetworkMTU: 8951
+  networkType: OpenShiftSDN
+  serviceNetwork:
+  - 172.30.0.0/16
+----
+
+. Add the following code to the `spec` field in the
+`NetworkConfig.networkoperator.openshift.io` CR:
```
squeed: Network.operator.openshift.io

modules/network-modify-config.adoc
```diff
@@ -0,0 +1,79 @@
...
+  kubeProxyConfig:
+   iptablesSyncPeriod: 30s
+   bindAddress: 0.0.0.0
+   proxyArguments:
+   - "iptables-min-sync-period": ["30s"]
+----
+
+. Save the file and quit the text editor.
+
+The Network Operator checks the syntax when you quit the file. If your
```
squeed: OC / kubectl checks the syntax.

modules/network-modify-config.adoc
```diff
@@ -0,0 +1,79 @@
...
+modifications contain a syntax error, the editor will re-open the file and
+display an error message at the top.
```
squeed: It would be good to add a bit about checking the operator status to see if your changes were accepted.

modules/network-modifying-operator-install-config.adoc
```diff
@@ -0,0 +1,144 @@
...
+    hostPrefix: 23
+  networkType: OpenShiftSDN
+  serviceNetwork:
+  - 172.30.0.0/16
+status: {}
+----
++
+The Network Operator takes its configuration from the
+`cluster-network-02-config.yml` file when the cluster is created. Both the
+`Network.config.openshift.io` CR and the `Network.operator.openshift.io` CR must
```
squeed: This isn't the case - the files can be named anything.

modules/network-modifying-operator-install-config.adoc
```diff
@@ -0,0 +1,144 @@
+// Module included in the following assemblies:
```
squeed: It would be good to give some motivation - that there are some operator-specific parameters that are not covered as part of the Network.config.openshift.io object.

modules/network-modifying-operator-install-config.adoc
danwinship: But `cluster-network-02-config.yml` is already created by the installer.

However, AFAICT when testing, this paragraph is explicitly wrong: the installer assumes that each file contains only a single object, and so the cluster config and operator config need to be in separate files

modules/network-operator-cr.adoc
```diff
@@ -0,0 +1,94 @@
...
+  deployKubeProxy: false
+  kubeProxyConfig: <10>
+    bindAddress: 0.0.0.0 <11>
+    iptablesSyncPeriod: 30s <12>
+    proxyArguments:
+      iptables-min-sync-period: <13>
+      - 30s
+----
+
+<1> `cidr`: A block of IP addresses from which pod IPs are allocated.
```
kalexand-rh: None of the other sections of the doc use this format for codeblock annotations. Remove or relocate the parameter definitions in all of these annotation sets.

modules/network-config-cr.adoc
```diff
@@ -0,0 +1,42 @@
...
+configuration in a `config.openshift.io` Custom Resource (CR) object. All fields
```
kalexand-rh: Please check all the instances of `config.openshift.io`. Based on one of of @squeed's comments, I'm not sure that you're using this right.

modules/network-operator-cr.adoc
```diff
@@ -0,0 +1,94 @@
...
+{product-title} uses the Classless Inter-Domain Routing (CIDR) standard to
+allocate IP addresses. The 'OpenShiftSDN' network plugin supports multiple
+ClusterNetworks. The address blocks for multiple ClusterNetworks must not
+overlap. This parameter is set in the `config.openshift.io` object.
```
kalexand-rh: Is `config.openshift.io` right through here? Don't you set the value in this CR?

modules/network-operator-cr.adoc
```diff
@@ -0,0 +1,94 @@
...
+<2> `hostPrefix`:
```
kalexand-rh: Missing description.

modules/network-operator-cr.adoc
```diff
@@ -0,0 +1,94 @@
...
+<6> `openshiftSDNConfig`: Specify only if you want to override part of the
+{product-title} SDN configuration.
+<7> `mode`: Configures the isolation mode for `OpenShiftSDN`. The allowed values
+are `Multitenant`, `Subnet`, or `NetworkPolicy`. The default value is
+`NetworkPolicy`.
+<8> `mtu`: MTU for the VXLAN overlay network. This is normally configured
+automatically, but if the nodes in your cluster do not all use the same MTU,
+then you must set this explicitly to 50 less than the smallest node MTU value.
+<9> `VXLANPort`: The port to use for all VXLAN packets. The default value is
+4789. You should not need to change this unless you are running in a virtualized
```
kalexand-rh: Rewrite to remove "should."

modules/network-operator-cr.adoc
```diff
@@ -0,0 +1,94 @@
...
+environment where some nodes are already part of another VXLAN network.
+<10> `kubeProxyConfig`: The parameters in this section specify the
+`kube-proxy` configuration. If you do not specify the parameter values, the
+Network Operator chooses sensible defaults.
```
kalexand-rh: s/chooses sensible defaults/applies the default parameter values.
What's a sensible default? When do I know if I want something other than a sensible default? Isn't this one of the only values that can be changed after installation? You should point that out.

modules/network-operator-cr.adoc
```diff
@@ -0,0 +1,94 @@
...
+//Currently only `OpenShiftSDN` consumes this.
+//+optional
+<11> `bindAddress`: The IP address for `kube-proxy` to serve on. The default
+value is `0.0.0.0`.
+<12> `iptablesSyncPeriod`: The refresh period for `iptables` rules. The default
+value is `30s`. Valid suffixes include `s`, `m`, and `h` and are described in
+the link:https://golang.org/pkg/time/#ParseDuration[Go time package]
+documentation.
+<13> `iptables-min-sync-period`: The minimum duration before refreshing
+`iptables` rules. This ensures that the refresh does not happen too frequently.
```
kalexand-rh: s/This/This rule or duration

> Avoid sentences in which a pronoun can refer to more than one noun or in which a pronoun refers to an unstated noun.

modules/network-operator-cr.adoc
```diff
@@ -0,0 +1,94 @@
...
+Valid suffixes include `s`, `m`, and `h` and are described in the
+link:https://golang.org/pkg/time/#ParseDuration[Go time package]
+
+Use the following command to view the `Network.operator.openshift.io` CR object
+in a running cluster.
+[source]
```
kalexand-rh: remove [source]

modules/network-view-config.adoc
```diff
@@ -0,0 +1,87 @@
...
+// * networking/configuring-network-operator.adoc
+
+[id="network-view-config-{context}"]
+= Viewing the network configuration in a running cluster
+
+// Add explanation of value for this content
+
+.Prerequisites
+
+* Start a cluster and log in using your credentials.
```
kalexand-rh: Which permissions do I need?

modules/nw-nwop-config-OpenShiftSDN.adoc
```diff
@@ -0,0 +1,36 @@
+// Deprecated
```
kalexand-rh: If it's deprecated, why are you adding it?

modules/nw-nwop-config-kubeproxy.adoc
```diff
@@ -0,0 +1,32 @@
+// Deprecated
```
kalexand-rh: If it's deprecated, why are you adding it?

kalexand-rh (commented):
@danwinship, I have a couple of questions for you mixed in my review comments.

_topic_map.yml
```diff
@@ -67,6 +67,8 @@ Topics:
     File: installing-aws-default
   - Name: Installing a cluster on AWS with customizations
     File: installing-aws-customizations
+  - Name: Installing a cluster on AWS with network customizations
+    File: installing-aws-network-customizations
```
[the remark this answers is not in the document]
danwinship: Virtually all of the network customizations must be made at install time and can't be changed afterward.

modules/network-modifying-operator-install-config.adoc
danwinship: You need to put the `config.openshift.io` object and the `operator.openshift.io` object in separate files, rather than concatenating them together into a single file like it says here.

kalexand-rh: Are there any limitations on the filenames?

_topic_map.yml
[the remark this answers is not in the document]
kalexand-rh: That's fair! Do you think most users will want to make these changes?

modules/network-modifying-operator-install-config.adoc
danwinship: The `operator.openshift.io` object has to be in a file that comes alphabetically after `cluster-network-01-crd.yml`. Other than that, no. `cluster-network-03-config.yml` would be a fine name for it.

_topic_map.yml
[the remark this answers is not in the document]
danwinship: "Most"? No

Actually, I guess for people installing on AWS, changing the network defaults will be fairly uncommon. (People installing on bare metal or VMware or OpenStack are more likely to have problems with the default values.)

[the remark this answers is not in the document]
kalexand-rh: Thanks! Then I think it does need to be a separate assembly. We're isolating different installation user stories in separate assemblies so that users have an easier time finding the information that they need. (If you'd said that most users would want to tinker with the network settings in AWS, then I would vote to combine the assemblies.)

modules/network-modify-config.adoc
jboxman: What is `additionalNetworks`?

jboxman: @kalexand-rh, when you say separate assemblies, at what level of abstraction? Does this mean the `network-modify-config.adoc` file is split into two different files, one for each use case (`additionalNetworks` and `kubeProxy`), and then included in the original assembly?

Or does this mean creating second level assemblies and nesting each of these in `networking/configuring-network-operator.adoc`?

Or some other approach?

Thanks.

modules/network-modify-config.adoc
jboxman: @openshift/team-documentation, are we stripping out internal fields from `metadata` in general? I included it because it's verbatim output from a command.

modules/network-modify-config.adoc
jboxman: Thanks, I missed that one!

modules/network-modify-config.adoc
jboxman: @squeed, how do I check the operator status?

modules/nw-nwop-config-kubeproxy.adoc
jboxman: I didn't want to lose it. The file is a carry over from the previous PR. I'm going to delete it from the PR before merging.

modules/nw-nwop-config-OpenShiftSDN.adoc
jboxman: I didn't want to lose it. The file is a carry over from the previous PR. I'm going to delete it from the PR before merging.

modules/nw-modifying-nwoperator-config-startup.adoc
```diff
@@ -0,0 +1,149 @@
...
+[source,yaml]
+----
+
+---
+apiVersion: networkoperator.openshift.io/v1
+kind: NetworkConfig
+metadata:
+  name: cluster
+spec:
+  additionalNetworks: null
```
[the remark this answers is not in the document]
jboxman: What are valid values for the `additionalNetworks` parameter?

modules/network-modifying-operator-install-config.adoc
```diff
@@ -0,0 +1,144 @@
...
+  - cidr: 10.128.0.0/14
+    hostPrefix: 23
+  defaultNetwork:
+    type: OpenShiftSDN
+  serviceNetwork: 172.30.0.0/16
+  status: {}
+----
+<1> YAML separator to allow both CRs to exist in the same YAML file.
+
+. Edit the `cluster-network-02-config.yml` to customize your cluster network
```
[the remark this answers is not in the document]
jboxman: Missed that instance, thanks!

modules/network-modifying-operator-install-config.adoc
```diff
@@ -0,0 +1,144 @@
...
+  - 172.30.0.0/16
+status: {}
+----
++
+The Network Operator takes its configuration from the
+`cluster-network-02-config.yml` file when the cluster is created. Both the
+`Network.config.openshift.io` CR and the `Network.operator.openshift.io` CR must
+be stored in `cluster-network-02-config.yml`.
+
+. Open `cluster-network-02-config.yml` in a text editor.
```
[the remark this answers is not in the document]
jboxman: I think the updated process will involve creating a new file with the proper name and adding only the parameters not already present in `cluster-network-02-config.yml`. Much simplified to explain.

modules/network-operator-cr.adoc
jboxman: When you say _this format_, are you referring to "`parameter`: _description_" or something else?

modules/network-operator-cr.adoc
jboxman: Thanks!

modules/network-operator-cr.adoc
jboxman: The unusual aspect of this particular Operator is that much of its configuration is specified at installation time in `Network.config.openshift.io` and is subsequently made available to the Operator and available in `Network.operator.openshift.io`.

modules/nw-networkoperator-crd.adoc
jboxman: @danwinship, are you referring to `DeployKubeProxy` here?

danwinship: yes

modules/network-modify-config.adoc
kalexand-rh: None of these. If they're separate use cases that users are unlikely to need to address together, I think thatthey should be described in different assemblies in the networking folder.

jboxman: Thanks! I hadn't considered that approach.

kalexand-rh: Sure! I think we had most of the discussions about how to map a single user goal/story to a single assembly before you joined the team. The "1 user goal = 1 assembly" idea is also what's driving the need to have a separate installation assembly to cover networking.

jboxman:
@kalexand-rh, so in `modules/installation-launching-installer.adoc`, do I need to update the context for the directory callout? I'm including it in `installing/installing_aws/installing-aws-network-customizations.adoc`.

```
ifeval::["{context}" == "install-customizations-cloud"]
Specify the location of your customized `./install-config.yaml` file.
endif::[]
ifeval::["{context}" == "installing-aws-default"]
<1> Specify the directory name to store the files that the installation program
creates.
...
endif::[]
```
Also, does the first conditional need a &lt;1&gt;?

modules/nw-networkoperator-crd.adoc
jboxman: @danwinship, so does that mean that only `vxlanPort` and `mtu` are currently parameters that someone can modify in the Network Operator config?

modules/network-modifying-operator-install-config.adoc
```diff
@@ -0,0 +1,144 @@
...
+before continuing.
+====
+
+.Prerequisites
+
+* The `install-config.yaml` configuration file for the cluster you are creating.
+
+.Procedure
+
+. Use the following command to create manifests:
```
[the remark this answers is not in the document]
jboxman: I confirmed they are indeed thoroughly consumed.

modules/nw-networkoperator-crd.adoc
danwinship: I was just saying that `useExternalOpenvswitch` exists for a specific use case, but I don't know what that use case is. All the other fields are still relevant (or not), as discussed elsewhere here.

modules/network-modify-config.adoc
squeed: `oc get clusteroperator network`

modules/network-modifying-operator-install-config.adoc
squeed: For example, if you want to use Multitenant mode, change the MTU, or the vxlan port, you would need to create the operator configuration manually.

kalexand-rh:
> @kalexand-rh, so in `modules/installation-launching-installer.adoc`, do I need to update the context for the directory callout? I'm including it in `installing/installing_aws/installing-aws-network-customizations.adoc`.

@jboxman, why do you need to change the context? Based on your current assembly, you're creating the `install-config.yaml` file at this point, so the directory is correct. It's `modules/network-modifying-install-config.adoc` that needs to change.
- Remove the command to generate the file, because you don't need it.
- Remove the `modules/network-modifying-install-config.adoc` and update step 2 in `modules/installation-launching-installer.adoc` to also cover updating the network config section of the yaml.

jboxman:
> > @kalexand-rh, so in `modules/installation-launching-installer.adoc`, do I need to update the context for the directory callout? I'm including it in `installing/installing_aws/installing-aws-network-customizations.adoc`.
> 
> @jboxman, why do you need to change the context? Based on your current assembly, you're creating the `install-config.yaml` file at this point, so the directory is correct. It's `modules/network-modifying-install-config.adoc` that needs to change.

This:

https://snag.gy/tCwPib.jpg

The callout is missing when included in `installing/installing_aws/installing-aws-customizations.adoc`.

kalexand-rh:
Ah! @jboxman, yes, it should be. Sorry.

modules/network-modify-config.adoc
jboxman: @squeed, what are valid parameters for `additionalNetworks`?

squeed: @jboxman they are any of the extra networks provided by Multus. The only ones supported in 4.1 are SRIOV and RawCNI.

modules/network-operator-cr.adoc
```diff
@@ -0,0 +1,142 @@
...
+----
+apiVersion: operator.openshift.io/v1
+kind: Network
+spec:
+  clusterNetwork: <1>
+  - cidr: 10.128.0.0/14
+    hostPrefix: 23
+  serviceNetwork: 172.30.0.0/16 <1>
+  defaultNetwork:
+    type: OpenShiftSDN <1>
```
pecameron: This is also in the install-config. What happens if they don't match? What happens if they are missing?

modules/network-operator-cr.adoc
```diff
@@ -0,0 +1,142 @@
...
+      mtu: 1450 <4>
+      vxlanPort: 4789 <5>
+  kubeProxyConfig: <6>
+    bindAddress: 0.0.0.0 <7>
+    iptablesSyncPeriod: 30s <8>
+    proxyArguments:
+      iptables-min-sync-period: <9>
+      - 30s
+----
+<1> Specified in the `install-config.yaml` file.
```
pecameron: @squeed So why is it here?

modules/network-operator-cr.adoc
```diff
@@ -0,0 +1,142 @@
...
+<2> Specify only if you want to override part of the {product-title} SDN
+configuration.
+
+<3> Configures the isolation mode for `OpenShiftSDN`. The allowed values are
+`Multitenant`, `Subnet`, or `NetworkPolicy`. The default value is
+`NetworkPolicy`.
+
+<4> MTU for the VXLAN overlay network. This is normally configured
+automatically, but if the nodes in your cluster do not all use the same MTU,
+then you must set this explicitly to 50 less than the smallest node MTU value.
```
pecameron: To allow for the vxlan tunnel header.

modules/network-operator-cr.adoc
```diff
@@ -0,0 +1,142 @@
...
+endif::install[]
+
+ifdef::not-install[]
+----
+apiVersion: operator.openshift.io/v1
+kind: Network
+spec:
+  clusterNetwork:
+  - cidr: 10.128.0.0/14 <1>
+    hostPrefix: 23 <1>
```
pecameron: Should have some example with multiple cidrs. Also, the examples can use different values for the fields to show what can be done.

modules/network-operator-cr.adoc
```diff
@@ -0,0 +1,142 @@
...
+      - 30s
+----
+<1> Specified when the cluster is installed. It cannot be modified.
+
+<2> Specify only if you want to override part of the {product-title} SDN
+configuration.
+
+<3> Configures the isolation mode for `OpenShiftSDN`. The allowed values are
+`Multitenant`, `Subnet`, or `NetworkPolicy`. The default value is
+`NetworkPolicy`.
```
pecameron: @dcbw, @squeed can isolation mode be changed?

modules/network-operator-cr.adoc
```diff
@@ -0,0 +1,142 @@
...
+
+<4> MTU for the VXLAN overlay network. This is normally configured
```
pecameron: VXLAN keeps coming up. We should talk about it and tell why we care about it.

networking/configuring-network-operator.adoc
```diff
@@ -0,0 +1,35 @@
+[id="configuring-network-operator"]
+= Configuring the network at installation time
+include::modules/common-attributes.adoc[]
+:context: configuring-network-operator
+
+toc::[]
+
+The Network Operator installs and upgrades the networking components on an
```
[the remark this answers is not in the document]
pecameron: People say networking, but what does it really mean in a cluster? What does it do? What are the cluster features that are supported by networking. How does it work or change with the various options?

Reading this tells people that already know about how networking in 3.x works how to set things up in 4.x. The assumption is that you already know cluster networking and you are just interested in the control knobs to set things up properly.

I think that we should describe in a general way how networking works in the cluster. So you are an administrator and you want to insert a cluster into your network. What do you need to think about? How do you go about setting it all up? 

In a common case, the cluster is there to handle user requests for services. The user usually has a URL for the service and (often) through a browser sends a request. The external to the cluster network delivers the request to a router in the cluster and the router passes along to the app for service. The network manages passing the packets around the cluster. It makes sure packets get to where they need to be and  don't end up in the wrong place. 

Cluster networking manages a IPv4 network that connects the pods and services in the cluster. Each node is assigned a hostPrefix size set of IP addresses from one of the network cidrs. Pods that are created on that node are assigned an ip address form the nodes' pool....
and so it goes. The features and control knobs should be introduced in the context of how they are used probably from the cluster admin's perspective.

just my 2 cents

pecameron (commented):
There is a lot of good stuff here. I made some notes as I went through the PR.
In general, I would like a high level discussion from the admin's perspective that introduces networking and the various options and control knobs. Basically get the admin to think about cluster networking the way we want it used. I also personally like links to detail discussions from the general discussions. There are a lot of examples, which is good. The examples can be used to expose more of the features. I think that there is one network cidr and one service cidr. Also many of the options are not seen in the examples.

pecameron:
@jboxman The network operator does upgrades and monitors changes in networking. I don't think those features were mentioned.

jboxman:
@pecameron, thanks for taking a look at this! I completely agree about the conceptual information and expanding the set of examples.

modules/installation-aws-config-yaml.adoc
```diff
@@ -43,12 +51,57 @@ compute: <2>
 metadata:
   name: test-cluster <1>
 networking:
-  clusterNetworks:
+  clusterNetwork:
```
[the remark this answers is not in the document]
jboxman: There's a table in another file that precedes this configuration file in the rendered HTML that explains what each of these parameters is for.

jboxman:
I've created a new PR based on the comments from the past 30 days: https://github.com/openshift/openshift-docs/pull/14572

Thanks.

PR #14235, opened 2019-03-26
Add note about USER and PROJECT_DOMAIN_NAME support

luis5tb:


vikram-redhat:
@luis5tb for both 3.10 and 3.11?

PR #14241, opened 2019-03-26
AWS UPI draft

kalexand-rh:
From https://jira.coreos.com/browse/OSDOCS-323

http://file.rdu.redhat.com/kalexand/041719/osdocs323/installing/installing_aws_upi/installing-aws-upi.html

### DIFF
--- a/installing/installing_aws/installing-aws-customizations.adoc
+++ b/installing/installing_aws/installing-aws-customizations.adoc
@@ -15,8 +15,6 @@ to host the cluster.
 
 include::modules/installation-overview.adoc[leveloffset=+1]
 
-include::modules/installation-clouds.adoc[leveloffset=+1]
-
 include::modules/installation-provide-credentials.adoc[leveloffset=+1]
 
 include::modules/installation-obtaining-installer.adoc[leveloffset=+1]
--- a/installing/installing_aws_upi/installing-aws-upi.adoc
+++ b/installing/installing_aws_upi/installing-aws-upi.adoc
@@ -1,0 +1,78 @@
+[id="installing-aws-upi"]
+= Installing a cluster on AWS using CloudFormation templates
+include::modules/common-attributes.adoc[]
+:context: installing-aws-upi
+
+toc::[]
+
+In {product-title} version {product-version}, you can install a
+cluster on Amazon Web Services (AWS) using infrastructure that you provide.
+
+One way to create this infrastructure is to use the provided
+CloudFormation templates. You can modify the templates to customize your
+infrastructure or use the information that they contain to create AWS objects
+according to your company's policies.
+
+.Prerequisites
+
+* xref:../../installing/installing_aws/installing-aws-account.adoc[Configure an AWS account]
+to host the cluster.
+
+include::modules/installation-overview.adoc[leveloffset=+1]
+
+include::modules/installation-aws-upi-requirements.adoc[leveloffset=+1]
+
+include::modules/installation-aws-permissions.adoc[leveloffset=+2]
+
+include::modules/installation-obtaining-installer.adoc[leveloffset=+1]
+
+include::modules/installation-provide-credentials.adoc[leveloffset=+1]
+
+include::modules/installation-generate-aws-upi.adoc[leveloffset=+1]
+
+include::modules/installation-extracting-infraid.adoc[leveloffset=+1]
+
+include::modules/installation-creating-aws-vpc.adoc[leveloffset=+1]
+
+include::modules/installation-cloudformation-vpc.adoc[leveloffset=+2]
+
+include::modules/installation-creating-aws-dns.adoc[leveloffset=+1]
+
+include::modules/installation-cloudformation-dns.adoc[leveloffset=+2]
+
+include::modules/installation-creating-aws-security.adoc[leveloffset=+1]
+
+include::modules/installation-cloudformation-security.adoc[leveloffset=+2]
+
+include::modules/installation-aws-upi-rhcos-ami.adoc[leveloffset=+1]
+
+include::modules/installation-creating-aws-bootstrap.adoc[leveloffset=+1]
+
+include::modules/installation-cloudformation-bootstrap.adoc[leveloffset=+2]
+
+include::modules/installation-creating-aws-control-plane.adoc[leveloffset=+1]
+
+include::modules/installation-cloudformation-control-plane.adoc[leveloffset=+2]
+
+include::modules/installation-aws-upi-bootstrap.adoc[leveloffset=+1]
+
+////
+[id="installing-workers-aws-upi"]
+== Creating worker nodes
+
+You can either manually create worker nodes or use a MachineSet to create worker
+nodes after the cluster deploys. If you use a MachineSet to create and maintain
+the workers, you can allow the cluster to manage them. This allows you to easily
+scale, manage, and upgrade your workers.
+////
+
+
+include::modules/installation-creating-aws-worker.adoc[leveloffset=+2]
+
+include::modules/installation-cloudformation-worker.adoc[leveloffset=+3]
+
+include::modules/cli-install.adoc[leveloffset=+1]
+
+include::modules/cli-logging-in-kubeadmin.adoc[leveloffset=+1]
+
+include::modules/installation-aws-upi-installation.adoc[leveloffset=+1]
--- a/modules/installation-aws-upi-bootstrap.adoc
+++ b/modules/installation-aws-upi-bootstrap.adoc
@@ -1,0 +1,34 @@
+// Module included in the following assemblies:
+//
+// * installing/installing_aws_upi/installing-aws-upi.adoc
+
+[id="installation-aws-upi-bootstrap-{context}"]
+= Initializing the bootstrap node on AWS UPI
+
+After you create all of the required infrastructure in Amazon Web Services (AWS),
+you can install the cluster.
+
+.Prerequisites
+
+* Configure an AWS account.
+* Generate the Ignition config files for your cluster.
+* Create and configure a VPC and assocated subnets in AWS.
+* Create and configure DNS, load balancers, and listeners in AWS.
+* Create control plane and compute roles.
+* Create the bootstrap machine.
+* Create the control plane machines.
+* If you plan to manually manage the worker machines, create the worker machines.
+
+.Procedure
+
+. Change to the directory that contains the installation program and run the
+following command:
++
+----
+$ ./openshift-install wait-for bootstrap-complete --dir=<installation-directory> <1>
+----
+<1> Specify the directory name that you generated the `install-config.yaml` file for
+your cluster in.
++
+If the command exits without a `FATAL` warning, your production control plane
+has initialized.
--- a/modules/installation-aws-upi-installation.adoc
+++ b/modules/installation-aws-upi-installation.adoc
@@ -1,0 +1,90 @@
+// Module included in the following assemblies:
+//
+// * installing/installing_aws_upi/installing-aws-upi.adoc
+
+[id="installation-aws-upi-installation-{context}"]
+= Completing an AWS UPI installation
+
+After you start the {product-title} installation on Amazon Web Service (AWS)
+user-provisioned infrastructure, remove the bootstrap node, reconcile the default
+Machine and MachineSet definitions, and delete unused nodes
+
+.Prerequisites
+
+* Deploy the bootstrap node for an {product-title} cluster on user-provisioned AWS infrastructure.
+* Install the `oc` CLI and log in.
+
+.Procedure
+
+. Delete the bootstrap resources. If you used the CloudFormation template,
+link:https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-console-delete-stack.html[delete its stack]:
++
+----
+$ aws cloudformation delete-stack --stack-name <name> <1>
+----
+<1> `<name>` is the name of your bootstrap stack.
+
+////
+. View the list of machines in the `openshift-machine-api` namespace:
++
+----
+$ oc get machines --namespace openshift-machine-api
+NAME                                    INSTANCE   STATE     TYPE        REGION      ZONE         AGE
+test-tkh7l-master-0                                       m4.xlarge   us-east-2   us-east-2a   9m22s
+test-tkh7l-master-1                                       m4.xlarge   us-east-2   us-east-2b   9m22s
+test-tkh7l-master-2                                       m4.xlarge   us-east-2   us-east-2c   9m21s
+test-tkh7l-worker-us-east-2a-qjcxq                        m4.large    us-east-2   us-east-2a   8m6s
+test-tkh7l-worker-us-east-2b-nq8zs                        m4.large    us-east-2   us-east-2b   8m6s
+test-tkh7l-worker-us-east-2c-ww6c6                        m4.large    us-east-2   us-east-2c   8m7s
+----
++
+Note the `NAME` of each node. Because you manually deployed control plane
+nodes, the master machines are not controlled by the Machine API. Similarly,
+the worker machines are not backed by AWS instances on your subnet. You delete
+each of these machines.
+
+. Delete each of the listed machines:
++
+----
+$ oc delete machine --namespace openshift-machine-api <node_name> <1>
+machine.machine.openshift.io "<node_name>" deleted
+----
+<1> Specify the name of a master or worker node to delete.
+////
+
+. Review the pending certificate signing requests (CSRs) and ensure that the
+requests are for the machines that you added to the cluster:
++
+----
+$ oc get csr
+
+NAME        AGE     REQUESTOR                                                                   CONDITION
+csr-8b2br   15m     system:serviceaccount:openshift-machine-config-operator:node-bootstrapper   Approved,Issued
+csr-8vnps   15m     system:serviceaccount:openshift-machine-config-operator:node-bootstrapper   Approved,Issued
+csr-b96j4   25s     system:node:ip-10-0-52-215.us-east-2.compute.internal                       Approved,Issued
+csr-bfd72   5m26s   system:node:ip-10-0-50-126.us-east-2.compute.internal                       Pending
+csr-c57lv   5m26s   system:node:ip-10-0-95-157.us-east-2.compute.internal                       Pending
+----
+
+. Approve the CSRs for your cluster machines.
+** To approve them individually, run the following command for each valid
+CSR:
++
+----
+$ oc adm certificate approve <csr_name> <1>
+----
+<1> `<csr_name>` is the name of a CSR from the list of current CSRs.
+
+** If all the CSRs are valid, approve them all by running the following
+command:
++
+----
+$ oc get csr -ojson | jq -r '.items[] | select(.status == {} ) | .metadata.name' | xargs oc adm certificate approve
+----
+
+. Complete the cluster installation:
++
+----
+$ ./openshift-install wait-for install-complete
+INFO Waiting up to 30m0s for the cluster to initialize...
+----
--- a/modules/installation-aws-upi-requirements.adoc
+++ b/modules/installation-aws-upi-requirements.adoc
@@ -1,0 +1,523 @@
+// Module included in the following assemblies:
+//
+// * installing/installing_aws_upi/installing-aws-upi.adoc
+
+[id="installation-aws-upi-requirements-{context}"]
+= Required AWS UPI infrastructure components
+
+To install {product-title} on user-provisioned infrastructure in
+Amazon Web Services (AWS), you must manually create both the machines and their
+supporting infrastructure.
+
+You can use the provided CloudFormation templates to create this infrastructure,
+you can manually create the components, or you can reuse existing infrastructure
+that meets the cluster requirements. Review the CloudFormation templates
+for more details about how the components interrelate.
+
+[id="installation-aws-upi-cluster-machines-{context}"]
+== Cluster machines
+
+You need `AWS::EC2::Instance` objects for the following machines:
+
+* A bootstrap machine. This machine is required during installation, but you
+can remove it after your cluster deploys.
+* At least three control plane machines. The control plane machines are not
+governed by a MachineSet.
+* Compute machines. You can create these worker machines manually during the
+cluster installation process.
+
+////
+You can also create and control them by using a MachineSet after your
+control plane initializes and you can access the cluster API by using the `oc`
+command line interface.
+////
+
+You can use the following instance types for the cluster machines:
+
+.Valid instance types for machines
+
+[cols="2a,2a,2a,2a",options="header"]
+|===
+
+|Instance type
+|Bootstrap
+|Control plane
+|Compute
+
+|`i3.large`
+|x
+|
+|
+
+| `m4.large`
+|
+|
+|x
+
+| `m4.xlarge`
+|
+|x
+|x
+
+| `m4.2xlarge`
+|
+|x
+|x
+
+| `m4.4xlarge`
+|
+|x
+|x
+
+| `m4.8xlarge`
+|
+|x
+|x
+
+| `m4.10xlarge`
+|
+|x
+|x
+
+| `m4.16xlarge`
+|
+|x
+|x
+
+| `c4.large`
+|
+|
+|x
+
+| `c4.xlarge`
+|
+|
+|x
+
+| `c4.2xlarge`
+|
+|x
+|x
+
+| `c4.4xlarge`
+|
+|x
+|x
+
+| `c4.8xlarge`
+|
+|x
+|x
+
+| `r4.large`
+|
+|
+|x
+
+| `r4.xlarge`
+|
+|x
+|x
+
+| `r4.2xlarge`
+|
+|x
+|x
+
+| `r4.4xlarge`
+|
+|x
+|x
+
+| `r4.8xlarge`
+|
+|x
+|x
+
+| `r4.16xlarge`
+|
+|x
+|x
+
+|===
+
+[id="installation-aws-upi-other-infrastructure-{context}"]
+== Other infrastructure components
+
+* A VPC
+* DNS entries
+* Load balancers and listeners
+* A Route53 zone
+* Security groups
+* IAM roles
+* S3 buckets
+
+.Required VPC components
+
+You need to provide a suitable VPC and subnets that allow communication to your
+machines.
+
+[cols="2a,7a,3a,3a",options="header"]
+|===
+
+|Component
+|AWS type
+2+|Description
+
+|VPC
+|* `AWS::EC2::VPC`
+* `AWS::EC2::VPCEndpoint`
+2+|You must provide a public VPC for the cluster to use. The VPC requires an
+endpoint that references the route tables for each subnet.
+
+|Public subnets
+|* `AWS::EC2::Subnet`
+* `AWS::EC2::SubnetNetworkAclAssociation`
+2+|Your VPC must have public subnets for between 1 and 3 availability zones
+and associate them with appropriate ingress rules.
+
+|Internet gateway
+|
+* `AWS::EC2::InternetGateway`
+* `AWS::EC2::VPCGatewayAttachment`
+* `AWS::EC2::RouteTable`
+* `AWS::EC2::Route`
+* `PublicSubnetRouteTableAssociation`
+* `AWS::EC2::NatGateway`
+* `AWS::EC2::EIP`
+2+|You must have a public internet gateway, with public routes, attached to the
+VPC. Each public subnet must also be attached to the route and have a NAT
+gateway and EIP address.
+
+.7+|Network access control
+.7+| * `AWS::EC2::NetworkAcl`
+* `AWS::EC2::NetworkAclEntry`
+2+|You must allow the VPC to access the following ports:
+h|Port
+h|Reason
+
+|`80`
+|Inbound HTTP traffic
+
+|`443`
+|Inbound HTTPS traffic
+
+|`22`
+|Inbound SSH traffic
+
+|`1024` - `65535`
+|Inbound ephemeral traffic
+
+|`0` - `65535`
+|Outbound ephemeral traffic
+
+
+|Private subnets
+|* `AWS::EC2::Subnet`
+* `AWS::EC2::RouteTable`
+* `AWS::EC2::SubnetRouteTableAssociation`
+2+|Your VPC can have a private subnets. The provided CloudFormation templates
+can create private subnets for between 1 and 3 availability zones.
+If you use private subnets, you must provide appropriate routes and tables
+for them.
+
+|===
+
+
+.Required DNS and load balancing components
+
+Your DNS and load balancer configuration needs to use a public hosted zone and
+can use a private hosted zone similar to the IPI installation method. You must
+create a DNS entry that resolves to your load balancer. An entry for
+`api.<cluster_name>.<domain>` must point to the external load balancer, and an
+entry for `api-int.<cluster_name>.<domain>` must point to the internal load
+balancer.
+
+The cluster also requires load balancers and listeners for port 6443, which are
+required for the Kubernetes API and its extensions, and port 22623, which are
+required for the Ignition config files for new machines. The targets will be the
+master nodes. Port 6443 must be accessible to both clients external to the
+cluster and nodes within the cluster. Port 22623 must be accessible to nodes
+within the cluster.
+
+
+[cols="2a,2a,8a",options="header"]
+|===
+
+|Component
+|AWS type
+|Description
+
+|DNS
+|`AWS::Route53::HostedZone`
+|The hosted zone for your internal DNS.
+
+|etcd record sets
+|`AWS::Route53::RecordSet`
+|The registration records for etcd for your control plane machines.
+
+|Public load balancer
+|`AWS::ElasticLoadBalancingV2::LoadBalancer`
+|The load balancer for your public subnets.
+
+|External API server record
+|`AWS::Route53::RecordSetGroup`
+|Alias records for the external API server.
+
+|External listener
+|`AWS::ElasticLoadBalancingV2::Listener`
+|A listener on port 6443 for the external load balancer.
+
+|External target group
+|`AWS::ElasticLoadBalancingV2::TargetGroup`
+|The target group for the external load balancer.
+
+|Private load balancer
+|`AWS::ElasticLoadBalancingV2::LoadBalancer`
+|The load balancer for your private subnets.
+
+|Internal API server record
+|`AWS::Route53::RecordSetGroup`
+|Alias records for the internal API server.
+
+|Internal listener
+|`AWS::ElasticLoadBalancingV2::Listener`
+|A listener on port 26443 for the internal load balancer.
+
+|Internal target group
+|`AWS::ElasticLoadBalancingV2::TargetGroup`
+|The target group for the Internal load balancer.
+
+|Internal listener
+|`AWS::ElasticLoadBalancingV2::Listener`
+|A listener on port 26443 for the internal load balancer.
+
+|Internal target group
+|`AWS::ElasticLoadBalancingV2::TargetGroup`
+|The target group for the Internal load balancer.
+
+|===
+
+.Security groups
+
+The control plane and worker machines require access to the following ports:
+
+[cols="2a,2a,2a,2a",options="header"]
+|===
+
+|Group
+|Type
+|IP Protocol
+|Port range
+
+
+.4+|MasterSecurityGroup
+.4+|`AWS::EC2::SecurityGroup`
+|`icmp`
+|`0`
+
+|`tcp`
+|`22`
+
+|`tcp`
+|`6443`
+
+|`tcp`
+|`22623`
+
+.2+|WorkerSecurityGroup
+.2+|`AWS::EC2::SecurityGroup`
+|`icmp`
+|`0`
+
+|`tcp`
+|`22`
+
+
+.2+|BootstrapSecurityGroup
+.2+|`AWS::EC2::SecurityGroup`
+
+|`tcp`
+|`22`
+
+|`tcp`
+|`19531`
+
+|===
+
+.Control plane ingress
+
+The control plane machines require the following ingress groups. Each ingress group is
+a `AWS::EC2::SecurityGroupIngress` resource.
+
+[cols="2a,5a,2a,2a",options="header"]
+|===
+
+|Ingress group
+|Description
+|IP protocol
+|Port range
+
+
+|`MasterIngressEtcd`
+|etcd
+|`tcp`
+|`2379`- `2380`
+
+|`MasterIngressVxlan`
+|Vxlan packets
+|`udp`
+|`4789`
+
+|`MasterIngressWorkerVxlan`
+|Vxlan packets
+|`udp`
+|`4789`
+
+|`MasterIngressInternal`
+|Internal cluster communication
+|`tcp`
+|`9000` - `9999`
+
+|`MasterIngressWorkerInternal`
+|Internal cluster communication
+|`tcp`
+|`9000` - `9999`
+
+|`MasterIngressKube`
+|Kubernetes kubelet, scheduler and controller manager
+|`tcp`
+|`10250` - `10259`
+
+|`MasterIngressWorkerKube`
+|Kubernetes kubelet, scheduler and controller manager
+|`tcp`
+|`10250` - `10259`
+
+|`MasterIngressIngressServices`
+|Kubernetes ingress services
+|`tcp`
+|`30000` - `32767`
+
+|`MasterIngressWorkerIngressServices`
+|Kubernetes ingress services
+|`tcp`
+|`30000` - `32767`
+
+|===
+
+
+.Worker ingress
+
+The worker machines require the following ingress groups. Each ingress group is
+a `AWS::EC2::SecurityGroupIngress` resource.
+
+[cols="2a,5a,2a,2a",options="header"]
+|===
+
+|Ingress group
+|Description
+|IP protocol
+|Port range
+
+
+|`WorkerIngressVxlan`
+|Vxlan packets
+|`udp`
+|`4789`
+
+|`WorkerIngressWorkerVxlan`
+|Vxlan packets
+|`udp`
+|`4789`
+
+|`WorkerIngressInternal`
+|Internal cluster communication
+|`tcp`
+|`9000` - `9999`
+
+|`WorkerIngressWorkerInternal`
+|Internal cluster communication
+|`tcp`
+|`9000` - `9999`
+
+|`WorkerIngressKube`
+|Kubernetes kubelet, scheduler and controller manager
+|`tcp`
+|`10250`
+
+|`WorkerIngressWorkerKube`
+|Kubernetes kubelet, scheduler and controller manager
+|`tcp`
+|`10250`
+
+|`WorkerIngressIngressServices`
+|Kubernetes ingress services
+|`tcp`
+|`30000` - `32767`
+
+|`WorkerIngressWorkerIngressServices`
+|Kubernetes ingress services
+|`tcp`
+|`30000` - `32767`
+
+|===
+
+
+.Roles and instance profiles
+
+You must grant the machines permissions in AWS. The provided CloudFormation
+templates grant the machines permission the following `AWS::IAM::Role` objects
+and provide a `AWS::IAM::InstanceProfile` for each set of roles. If you do
+not use the templates, you can grant the machines the following broad permissions
+or the following individual permissions.
+
+[cols="2a,2a,2a,2a",options="header"]
+|===
+
+|Role
+|Effect
+|Action
+|Resource
+
+.4+|Master
+|`Allow`
+|`ec2:*`
+|`*`
+
+|`Allow`
+|`elasticloadbalancing:*`
+|`*`
+
+|`Allow`
+|`iam:PassRole`
+|`*`
+
+|`Allow`
+|`s3:GetObject`
+|`*`
+
+|Worker
+|`Allow`
+|`ec2:Describe*`
+|`*`
+
+
+.3+|Bootstrap
+|`Allow`
+|`ec2:Describe*`
+|`*`
+
+|`Allow`
+|`ec2:AttachVolume`
+|`*`
+
+|`Allow`
+|`ec2:DetachVolume`
+|`*`
+
+|`Allow`
+|`s3:GetObject`
+|`*`
+
+|===
--- a/modules/installation-aws-upi-rhcos-ami.adoc
+++ b/modules/installation-aws-upi-rhcos-ami.adoc
@@ -1,0 +1,66 @@
+// Module included in the following assemblies:
+//
+// * installing/installing_aws_upi/installing-aws-upi.adoc
+
+[id="installation-aws-upi-rhcos-ami-{context}"]
+= {op-system} AMIs for the AWS UPI infrastructure
+
+You must use a valid {op-system-first} AMI for your Amazon Web Services
+(AWS) zone for your {product-title} nodes.
+
+.{op-system} AMIs
+
+[cols="2a,2a",options="header"]
+|===
+
+|AWS zone
+|AWS AMI
+
+|`ap-northeast-1`
+|`ami-06153e8d5cf577a81`
+
+|`ap-northeast-2`
+|`ami-0b75c8dbcc07363b9`
+
+|`ap-south-1`
+|`ami-08a07f507cd20aa23`
+
+|`ap-southeast-1`
+|`ami-030cd605c9d3eec03`
+
+|`ap-southeast-2`
+|`ami-0140da25e0be718b2`
+
+|`ca-central-1`
+|`ami-027d77f6170fde18b`
+
+|`eu-central-1`
+|`ami-089595a0f4d60a839`
+
+|`eu-west-1`
+|`ami-063ce7cafbc2a16df`
+
+|`eu-west-2`
+|`ami-06619611f27b0103a`
+
+|`eu-west-3`
+|`ami-0ec28a762270a0d8f`
+
+|`sa-east-1`
+|`ami-0845523ad10519ac4`
+
+|`us-east-1`
+|`ami-0678c4c4a53cd4680`
+
+|`us-east-2`
+|`ami-07e0e0e0035b5a3fe`
+
+|`us-west-1`
+|`ami-0c58a195b65f5250f`
+
+|`us-west-2`
+|`ami-01e2758deb3135fb0`
+
+|===
+
+
--- a/modules/installation-cloudformation-control-plane.adoc
+++ b/modules/installation-cloudformation-control-plane.adoc
@@ -1,0 +1,374 @@
+// Module included in the following assemblies:
+//
+// * installing/installing_aws_upi/installing-aws-upi.adoc
+
+[id="installation-cloudformation-control-plane-{context}"]
+= CloudFormation template for control plane machines
+
+You can use the following CloudFormation template to deploy the control plane
+machines that you need for your {product-title} cluster.
+
+[source,yaml]
+----
+AWSTemplateFormatVersion: 2010-09-09
+Description: Template for Openshift Cluster UPI Node Launch (EC2 master instances)
+
+Parameters:
+  InfrastructureName:
+    AllowedPattern: ^([a-zA-Z][a-zA-Z0-9\-]{0,31})$
+    MaxLength: 32
+    MinLength: 1
+    ConstraintDescription: Infrastructure name must be alphanumeric, start with a letter and a maximum of 32 characters
+    Description: A short, unique cluster ID used to tag nodes for the kubelet cloud provider.
+    Type: String
+  RhcosAmi:
+    Description: Current RHEL CoreOS AMI to use for boostrap
+    Type: AWS::EC2::Image::Id
+  AutoRegisterDNS:
+    Default: "yes"
+    AllowedValues:
+    - "yes"
+    - "no"
+    Description: Do you want to invoke DNS etcd registration (requires Hosted Zone info provided)?
+    Type: String
+  PrivateHostedZoneId:
+    Description: The Route53 private zone ID to register the etcd targets with (e.g Z21IXYZABCZ2A4)
+    Type: String
+  PrivateHostedZoneName:
+    Description: The Route53 zone to register the targets with (No trailing dot - e.g. cluster.mycorp.com)
+    Type: String
+  Master0Subnet:
+    Description: The subnets (recommend private) to launch the master nodes into
+    Type: AWS::EC2::Subnet::Id
+  Master1Subnet:
+    Description: The subnets (recommend private) to launch the master nodes into
+    Type: AWS::EC2::Subnet::Id
+  Master2Subnet:
+    Description: The subnets (recommend private) to launch the master nodes into
+    Type: AWS::EC2::Subnet::Id
+  MasterSecurityGroupId:
+    Description: The master security group ID to associate with master nodes.
+    Type: AWS::EC2::SecurityGroup::Id
+  IgnitionLocation:
+    Default: https://api-int.$CLUSTER_NAME.$DOMAIN:22623/config/master
+    Description: Location to fetch bootstrap ignition from. (Recommend to use the autocreated ignition config location.)
+    Type: String
+  CertificateAuthorities:
+    Default: data:text/plain;charset=utf-8;base64,ABC...xYz==
+    Description: Base64 encoded certificate authority string to use.
+    Type: String
+  MasterInstanceProfileName:
+    Description: IAM profile to associate with master nodes.
+    Type: String
+  MasterInstanceType:
+    Default: m4.xlarge
+    Type: String
+    AllowedValues:
+    - "m4.xlarge"
+    - "m4.2xlarge"
+    - "m4.4xlarge"
+    - "m4.8xlarge"
+    - "m4.10xlarge"
+    - "m4.16xlarge"
+    - "c4.2xlarge"
+    - "c4.4xlarge"
+    - "c4.8xlarge"
+    - "r4.xlarge"
+    - "r4.2xlarge"
+    - "r4.4xlarge"
+    - "r4.8xlarge"
+    - "r4.16xlarge"
+  AutoRegisterELB:
+    Default: "yes"
+    AllowedValues:
+    - "yes"
+    - "no"
+    Description: Do you want to invoke NLB registration (requires Lambda ARN parameter to be supplied)?
+    Type: String
+  RegisterNlbIpTargetsLambdaArn:
+    Description: ARN for NLB IP target registration lambda (from cluster_infra_upi.yaml; otherwise select "no" for AutoRegisterELB)
+    Type: String
+  ExternalApiTargetGroupArn:
+    Description: ARN for external API load balancer target group (from cluster_infra_upi.yaml; otherwise select "no" for AutoRegisterELB)
+    Type: String
+  InternalApiTargetGroupArn:
+    Description: ARN for internal API load balancer target group (from cluster_infra_upi.yaml; otherwise select "no" for AutoRegisterELB)
+    Type: String
+  InternalServiceTargetGroupArn:
+    Description: ARN for internal service load balancer target group (from cluster_infra_upi.yaml; otherwise select "no" for AutoRegisterELB)
+    Type: String
+
+Metadata:
+  AWS::CloudFormation::Interface:
+    ParameterGroups:
+    - Label:
+        default: "Host Information"
+      Parameters:
+      - MasterInstanceType
+      - RhcosAmi
+      - IgnitionLocation
+      - CertificateAuthorities
+      - MasterSecurityGroupId
+      - MasterInstanceProfileName
+    - Label:
+        default: "Network Configuration"
+      Parameters:
+      - VpcId
+      - AllowedBootstrapSshCidr
+      - Master0Subnet
+      - Master1Subnet
+      - Master2Subnet
+    - Label:
+        default: "DNS"
+      Parameters:
+      - AutoRegisterDNS
+      - PrivateHostedZoneName
+      - PrivateHostedZoneId
+    - Label:
+        default: "Load Balancer Automation"
+      Parameters:
+      - AutoRegisterELB
+      - RegisterNlbIpTargetsLambdaArn
+      - ExternalApiTargetGroupArn
+      - InternalApiTargetGroupArn
+      - InternalServiceTargetGroupArn
+    ParameterLabels:
+      VpcId:
+        default: "VPC ID"
+      Master0Subnet:
+        default: "Master-0 Subnet"
+      Master1Subnet:
+        default: "Master-1 Subnet"
+      Master2Subnet:
+        default: "Master-2 Subnet"
+      MasterInstanceType:
+        default: "Master Instance Type"
+      MasterInstanceProfileName:
+        default: "Master Instance Profile Name"
+      RhcosAmi:
+        default: "RHEL CoreOS AMI ID"
+      BootstrapIgnitionLocation:
+        default: "Master Ignition Source"
+      CertificateAuthorities:
+        default: "Ignition CA String"
+      MasterSecurityGroupId:
+        default: "Master Security Group ID"
+      AutoRegisterDNS:
+        default: "Use Provided DNS Automation"
+      AutoRegisterELB:
+        default: "Use Provided ELB Automation"
+      PrivateHostedZoneName:
+        default: "Private Hosted Zone Name"
+      PrivateHostedZoneId:
+        default: "Private Hosted Zone ID"
+
+Conditions:
+  DoRegistration: !Equals ["yes", !Ref AutoRegisterELB]
+  DoDns: !Equals ["yes", !Ref AutoRegisterDNS]
+
+Resources:
+  Master0:
+    Type: AWS::EC2::Instance
+    Properties:
+      ImageId: !Ref RhcosAmi
+      IamInstanceProfile: !Ref MasterInstanceProfileName
+      InstanceType: !Ref MasterInstanceType
+      NetworkInterfaces:
+      - AssociatePublicIpAddress: "false"
+        DeviceIndex: "0"
+        GroupSet:
+        - !Ref "MasterSecurityGroupId"
+        SubnetId: !Ref "Master0Subnet"
+      UserData:
+        Fn::Base64: !Sub
+        - '{"ignition":{"config":{"append":[{"source":"${SOURCE}","verification":{}}]},"security":{"tls":{"certificateAuthorities":[{"source":"${CA_BUNDLE}","verification":{}}]}},"timeouts":{},"version":"2.2.0"},"networkd":{},"passwd":{},"storage":{},"systemd":{}}'
+        - {
+          SOURCE: !Ref IgnitionLocation,
+          CA_BUNDLE: !Ref CertificateAuthorities,
+        }
+      Tags:
+      - Key: !Join ["", ["kubernetes.io/cluster/", !Ref InfrastructureName]]
+        Value: "shared"
+
+  RegisterMaster0:
+    Condition: DoRegistration
+    Type: Custom::NLBRegister
+    Properties:
+      ServiceToken: !Ref RegisterNlbIpTargetsLambdaArn
+      TargetArn: !Ref ExternalApiTargetGroupArn
+      TargetIp: !GetAtt Master0.PrivateIp
+
+  RegisterMaster0InternalApiTarget:
+    Condition: DoRegistration
+    Type: Custom::NLBRegister
+    Properties:
+      ServiceToken: !Ref RegisterNlbIpTargetsLambdaArn
+      TargetArn: !Ref InternalApiTargetGroupArn
+      TargetIp: !GetAtt Master0.PrivateIp
+
+  RegisterMaster0InternalServiceTarget:
+    Condition: DoRegistration
+    Type: Custom::NLBRegister
+    Properties:
+      ServiceToken: !Ref RegisterNlbIpTargetsLambdaArn
+      TargetArn: !Ref InternalServiceTargetGroupArn
+      TargetIp: !GetAtt Master0.PrivateIp
+
+  Master1:
+    Type: AWS::EC2::Instance
+    Properties:
+      ImageId: !Ref RhcosAmi
+      IamInstanceProfile: !Ref MasterInstanceProfileName
+      InstanceType: !Ref MasterInstanceType
+      NetworkInterfaces:
+      - AssociatePublicIpAddress: "false"
+        DeviceIndex: "0"
+        GroupSet:
+        - !Ref "MasterSecurityGroupId"
+        SubnetId: !Ref "Master1Subnet"
+      UserData:
+        Fn::Base64: !Sub
+        - '{"ignition":{"config":{"append":[{"source":"${SOURCE}","verification":{}}]},"security":{"tls":{"certificateAuthorities":[{"source":"${CA_BUNDLE}","verification":{}}]}},"timeouts":{},"version":"2.2.0"},"networkd":{},"passwd":{},"storage":{},"systemd":{}}'
+        - {
+          SOURCE: !Ref IgnitionLocation,
+          CA_BUNDLE: !Ref CertificateAuthorities,
+        }
+      Tags:
+      - Key: !Join ["", ["kubernetes.io/cluster/", !Ref InfrastructureName]]
+        Value: "shared"
+
+  RegisterMaster1:
+    Condition: DoRegistration
+    Type: Custom::NLBRegister
+    Properties:
+      ServiceToken: !Ref RegisterNlbIpTargetsLambdaArn
+      TargetArn: !Ref ExternalApiTargetGroupArn
+      TargetIp: !GetAtt Master1.PrivateIp
+
+  RegisterMaster1InternalApiTarget:
+    Condition: DoRegistration
+    Type: Custom::NLBRegister
+    Properties:
+      ServiceToken: !Ref RegisterNlbIpTargetsLambdaArn
+      TargetArn: !Ref InternalApiTargetGroupArn
+      TargetIp: !GetAtt Master1.PrivateIp
+
+  RegisterMaster1InternalServiceTarget:
+    Condition: DoRegistration
+    Type: Custom::NLBRegister
+    Properties:
+      ServiceToken: !Ref RegisterNlbIpTargetsLambdaArn
+      TargetArn: !Ref InternalServiceTargetGroupArn
+      TargetIp: !GetAtt Master1.PrivateIp
+
+  Master2:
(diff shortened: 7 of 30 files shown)


modules/installation-aws-upi-installation.adoc
```diff
@@ -0,0 +1,164 @@
...
+          userDataSecret:
+            name: worker-user-data
+      versions:
+        kubelet: ""
+status:
+  fullyLabeledReplicas: 1
+  observedGeneration: 1
+  replicas: 1
+----
+<1> I think you change the RHCOS AMI here.
```
kalexand-rh: @cuppett, do you know if there's a good RHCOS AMI that we should use?

Do you have more details about the values that a user would need to provide to update these MachineSets?

cuppett: The IAM instance profile name comes as an output from the 03_cluster_security.yaml file. The subnet is created as part of 01_vpc.yaml; however, it doesn't have a name tag to filter on. They can change the specification to ID and use the exact ID or add a name. We could also add a generated name tag to help minimize the edit needed. The same is true for the security group.

For the AMI, we build a different one each day. I usually just query the AWS console to grab a recent one. @wking may have a more prescriptive method.

wking: > For the AMI...

See [here][1] and [here][2] for scripted approaches to figuring out your AMI.  And see [here][3] for possible future work.

[1]: https://github.com/openshift/installer/pull/1423#issuecomment-479348663
[2]: https://github.com/openshift/installer/pull/1168#issuecomment-459501811
[3]: https://github.com/openshift/installer/issues/1399

kalexand-rh: Thank you! I'll watch the RFE because that will be useful.

modules/installation-aws-upi-installation.adoc
```diff
@@ -0,0 +1,187 @@
...
+status:
+  fullyLabeledReplicas: 1
+  observedGeneration: 1
+  replicas: 1
+----
+<1> Specify the {op-system-first} AMI to use for your worker nodes.
+<2> Specify the name of the worker security group that you created.
+<3> Specify the subnet for the machines.
+
+.. Replace the `securityGroups` stanza with one that specifies the name of the
```
kalexand-rh: @cuppett, do you have the format for the securityGroup and subnet stanzas that use names instead of tags?

kalexand-rh:
@wking, will you PTAL at this AWS UPI draft? (I know that we're scrubbing UPI from the docs, but I'm waiting for updates from the open JIRA to try to fix this terminology.)\

@jianlinliu, will you PTAL?

modules/installation-aws-upi-installation.adoc
cuppett: Here is one (lifted from a BZ):

```
      securityGroups:
      - filters:
        - name: tag:Name
          values:
          - worker-sg-jialiuuuu1
      subnet:
        filters:
        - name: tag:Name
          values:
          - jialiuuuu1-xbhm2-private-us-east-2b
```
There should also be a way to specify this using the always-there subnet or security group ID (instead of a tag of Name), but I'm not sure where these filters' formats are defined well. We should find them or a spot for them.

installing/installing_aws_upi/installing-aws-upi.adoc
```diff
@@ -0,0 +1,81 @@
...
+[id="installing-workers-aws-upi"]
+== Creating worker nodes
+
+You can either manually create worker nodes before you install the cluster or
+use a MachineSet to create worker nodes after the cluster deploys.
+
+[WARNING]
+====
+It is better to use a MachineSet to allow the cluster to create and manage
+worker nodes than to manually create them.
```
wking: Judgemental?  If we are going to claim this, we should back it up, e.g. by talking about how easy it is to let the cluster automatically roll out newer RHCOS or recover from lost machines or some such.

modules/installation-aws-upi-bootstrap.adoc
```diff
@@ -0,0 +1,34 @@
...
+[id="installation-aws-upi-bootstrap-{context}"]
+= Initializing the bootstrap node on AWS UPI
+
+After you create all of the required infrastructure in Amazon Web Services (AWS),
+you can install the cluster.
+
+.Prerequisites
+
+* Configure an AWS account.
+* Generate the ignition config files for your cluster.
```
wking: "ignition" -> "Ignition"

modules/installation-aws-upi-bootstrap.adoc
```diff
@@ -0,0 +1,34 @@
...
+* Create and configure a VPC and assocated subnets in AWS.
+* Create and configure DNS, load balancers, and listeners in AWS.
+* Create control plane and worker roles.
```
wking: This seems split between the new "control-plane / compute" and the old "master / worker".  Since this is at the header level, maybe use both of the new terms?  `Create control-plane and compute roles.`

modules/installation-aws-upi-bootstrap.adoc
```diff
@@ -0,0 +1,34 @@
...
+* Create the bootstrap node.
```
wking: The bootstrap machine/instance/whatever is not a node in the Kubernetes sense, since it is not available to the control-plane for scheduling Pods.  Can we say "bootstrap machine" or "bootstrap host" or something more generic?  I feel like the "machine" wording has come up before, so feel free to mark this resolved if the conclusion was "'node' is the best we can do" ;).

modules/installation-aws-upi-bootstrap.adoc
```diff
@@ -0,0 +1,34 @@
...
+* Create the control plane nodes.
+* If you plan to manually manage the worker machines, create the worker machines.
+
+.Procedure
+
+. Change to the directory that contains the installation program and run the
+following command:
++
+----
+$ ./openshift-install user-provided-infrastructure bootstrap-complete --dir=<installation-directory> <1>
```
wking: This is stale since openshift/installer#1552 with more changes in the pipe in openshift/installer#1568.  I'm not sure which version you're targeting in these docs; you may want to wait until openshift/installer#1568 settles and then pivot to wherever it ended up.

nit: I also prefer setting `--dir` right after `openshift-install` to help underline its position as an `openshift-install` option.  You *can* set it later, but I don't want users stopping to think "does this other subcommand support `--dir` or not?", because you can always set it.

modules/installation-aws-upi-bootstrap.adoc
```diff
@@ -0,0 +1,34 @@
...
+----
+<1> Specify the directory name that you generated the `install-config.yaml` file
+and ignition config files for your cluster.
++
+After a message displays that contains `bootstrap-complete`, your bootstrap node
+has initialized.
```
wking: Folks should be considering exit codes, not logged messages.  We may change logged messages later, but "exited zero" is always going to be a stable success marker.  If you want a more-stable log marker, you might say "If the command exits without a FATAL warning, your production control-plane has initialized" (or something, I'm leaning on the openshift/installer#1568 wording).

modules/installation-aws-upi-installation.adoc
```diff
@@ -0,0 +1,187 @@
...
+[id="installation-aws-upi-cleanup-{context}"]
+= Completing an AWS UPI installation
+
+After you start the {product-title} installation on Amazon Web Service (AWS)
+user-provisioned infrastructure, remove the bootstrap node, reconcile the default
+Machine and MachineSet definitions, and delete unused nodes
+
+.Prerequisites
+
+* Deploy the bootstrap node for a{product-title} cluster on user-provisioned AWS infrastructure.
```
wking: nit: missing space in `a{product-title}`?

modules/installation-aws-upi-installation.adoc
wking: The `subnet` entry is defined [here][1].  For UPI, it's probably easier to just set the subnet [ID or ARN][2].

[1]: https://github.com/openshift/cluster-api-provider-aws/blob/d19c6e58a9b2c5597a27f2f6ea4bd4a46f43c194/pkg/apis/awsproviderconfig/v1beta1/awsmachineproviderconfig_types.go#L130-L131
[2]: https://github.com/openshift/cluster-api-provider-aws/blob/d19c6e58a9b2c5597a27f2f6ea4bd4a46f43c194/pkg/apis/awsproviderconfig/v1beta1/awsmachineproviderconfig_types.go#L216-L226

modules/installation-aws-upi-requirements.adoc
```diff
@@ -0,0 +1,399 @@
...
+* DNS entries
+* Load balancers and listeners
+* A Route53 zone
+* Security groups
+* IAM roles
+* S3 buckets
+
+.Required VPC components
+
+You need to create a VPC and subnets that allow communication to your machines.
```
wking: Isn't one of the UPI drivers that you can recycle existing infra?  You don't need to *create* a VPC, you just need to *have* one.  Can we rephrase this and other parts of this doc to make that distinction ("If you don't have an existing VPC, you can create one with our handy CloudFormation config or whatever you want to use instead...").

modules/installation-aws-upi-requirements.adoc
```diff
@@ -0,0 +1,399 @@
...
+|Inbound HTTP traffic
+
+|`443`
+|Inbound HTTPS traffic
+
+|`22`
+|Inbound SSH traffic
+
+|`1024` - `65535`
+|Inbound ephemeral traffic
```
wking: This is what the current installer CloudFormation templates have, but I don't know that such a wide range is actually required.  I'll see about getting better installer-side docs for these rules, and ping you once we have something.

modules/installation-aws-upi-requirements.adoc
```diff
@@ -0,0 +1,399 @@
...
+
+|`0` - `65535`
+|Outbound ephemeral traffic
+
+
+|Private subnets
+|* `AWS::EC2::Subnet`
+* `AWS::EC2::RouteTable`
+* `AWS::EC2::SubnetRouteTableAssociation`
+2+|Your VPC can create a private subnets for between 1 and 3 availability zones.
```
wking: More subnets are fine too, they just aren't supported by the current CloudFormation templates.

modules/installation-aws-upi-requirements.adoc
```diff
@@ -0,0 +1,399 @@
...
+|Kubernetes ingress services
+|`tcp`
+|`30000` - `32767`
+
+|===
+
+
+.Roles and instance profiles.
+
+The machines require the following `AWS::IAM::Role` objects. You must also
```
wking: The permissions you have below are overly broad.  You can make them more specific (see the list [here][1]), but even that list is overly broad.  So regardless of what you do to the values below, you probably want to soften the "require" wording here.

[1]: https://github.com/openshift/installer/blob/27b24b8250293db120464095960f3b42cfaa94f5/pkg/asset/installconfig/aws/permissions.go

modules/installation-creating-aws-dns.adoc
```diff
@@ -0,0 +1,123 @@
...
+{product-title} cluster to use. The easiest way to create these components is
+to modify the provided CloudFormation template.
+
+If you need to, you can run the template multiple times within a single VPC.
+
+[NOTE]
+====
+If you do not use the provided CloudFormation template to create your AWS
+infrastructure, you must review the provided information and manually create
+the infrastructure. You might need to consult Red Hat support to ensure that
```
wking: How will they know?  Ideally, the test for UPI config is "I tried to launch a cluster on my infra, and it came up".  If that fails, it's on us to provide useful debugging/log-gathering tools to help identify the violated assumption and repair.  But that's going to be an incremental thing with no "now we've fixed it" moment.  Can we just reword this to say "try it and see, and fall back to support with your tool-gathered logs if you get stuck"?

modules/installation-overview.adoc
```diff
@@ -7,50 +7,148 @@
...
+that you prepare and maintain.
+
+These two basic types of {product-title} clusters are
+Installer Provisioned Infrastructure (IPI) clusters and
+and User Provisioned Infrastructure (UPI) clusters.
+
+You use the same installation program to deploy both types of clusters. The main
+assets generated by the installation program are the Ignition configurations for the
+bootstrap, master, and worker machines. With these three configurations and correctly
+configured infrastructure, you can start an {product-title} cluster.
```
wking: heh, I just had a `{product-article} {product-title}` knee-jerk ;).

modules/installation-overview.adoc
```diff
@@ -7,50 +7,148 @@
 [id="installation-overview-{context}"]
 = {product-title} installation overview
 
-You use the {product-title} installer to create {product-title} clusters in
-//a variety of environments. The same installer allows you to install
-//{product-title} on your own infrastructure or in
+The {product-title} 4 installation program offers you flexibility. You can use the installation
```
wking: 4 -> `{product-version}` or `{product-major-version}` or some such (here and below)?

modules/installation-overview.adoc
```diff
@@ -7,50 +7,148 @@
...
+
+[IMPORTANT]
+====
+It is possible to modify Kubernetes and the Ignition Configs that control
+the underlying {op-system-first} operating system during installation. However,
+no validation is available to confirm the suitability of any modifications that
+you make to these objects. If you modify these objects, you might render
+your cluster non-functional. Because of this risk, modifying Kubernetes and
+Ignition Configs is not supported unless you are following documented procedures
+or instructed to do so by Red Hat support.
```
wking: nit: "are following" doesn't match "instructed" for the "or".  Maybe "were instructed"?

modules/installation-overview.adoc
```diff
@@ -7,50 +7,148 @@
...
 
+[discrete]
+=== UPI cluster installation process
+
+You can also install {product-title} on infrastructure that you provide. You use the
+installation program to generate the assets that you need to provision the
+cluster infrastructure, create the cluster infrastructure, and then deploy
+the cluster to the infrastructure that you provided.
+
+With UPI clusters, you must manage and maintain the cluster nodes yourself.
```
wking: "nodes" -> "resources", since you're on the hook for subnets and load balancers and route tables, etc., etc.  Maybe list a few explicitly here?  We want this to sound as unappealing and tedious as it's going to be so folks are more likely to make their lives easier by taking the IPI route ;).

modules/installation-aws-upi-bootstrap.adoc
```diff
@@ -0,0 +1,34 @@
+// Module included in the following assemblies:
+//
+// * installing/installing_aws_upi/installing-aws-upi.adoc
+
+[id="installation-aws-upi-bootstrap-{context}"]
+= Initializing the bootstrap node on AWS UPI
```
jianlinliu: I think this step should happen prior to adding compute/worker nodes, because it is related to "bootstrap".

modules/installation-aws-upi-installation.adoc
```diff
@@ -0,0 +1,187 @@
...
+
+. Delete the bootstrap resources. If you used the CloudFormation template,
+link:https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-console-delete-stack.html[delete its stack]:
++
+----
+$ aws cloudformation delete-stack --stack-name <name> <1>
+----
+<1> `<name>` is the name of your bootstrap stack.
+
+. View the list of machines in the `openshift-machine-api` namespace:
```
jianlinliu: Seem like the following steps is doing https://github.com/openshift/installer/blob/master/docs/user/aws/install_upi.md#option-1-dynamic-compute-using-machine-api. These steps are not working well, refer to https://bugzilla.redhat.com/show_bug.cgi?id=1697968.

modules/installation-creating-aws-worker.adoc
```diff
@@ -0,0 +1,166 @@
...
+====
+If you do not use the provided CloudFormation template to create your worker
+nodes, you must review the provided information and manually create
+the infrastructure. You might need to consult Red Hat support to ensure that
+your cluster initializes correctly.
+====
+
+[WARNING]
+====
+It is better to use a MachineSet to allow the cluster to create and manage
```
jianlinliu: It is better to have a link to that section how to use machineset to provision new worker node.

modules/installation-aws-upi-bootstrap.adoc
kalexand-rh: @wking, we're targeting beta4 for this drop. If you expect openshift/installer#1568 to land before that drop, I'll pivot to that.

modules/installation-aws-upi-requirements.adoc
```diff
@@ -0,0 +1,399 @@
...
+
+Create `AWS::EC2::Instance` objects for the following machines:
+
+* A bootstrap node. This machine is required during installation, but you
+can remove it after your cluster deploys.
+* At least three control plane machines. The control plane machines are not
+governed by a machineset.
+* Worker nodes. You can create these manually during the cluster installation
+process or you can create and control them by using a MachineSet after
+you create a bootstrap node.
```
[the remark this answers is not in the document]
kalexand-rh: Yes. "You can also create and control them by using a MachineSet after your control plane initializes and you can access the cluster API by using the `oc` command line interface."

modules/installation-creating-aws-dns.adoc
```diff
@@ -0,0 +1,123 @@
...
+  },
+  {
+    "ParameterKey": "VpcId", <11>
+    "ParameterValue": "vpc-055h956b1e6572be9" <12>
+  }
+]
+----
+<1> A short, representative cluster name to use for hostnames, etc.
+<2> Specify the cluster name that you used when you generated the
+`install-config.yaml` file.
```
skordas: cluster name is also used when user decide to generate only ignition-configs files.

modules/installation-creating-aws-dns.adoc
```diff
@@ -0,0 +1,123 @@
...
+<3> The Route53 public zone ID to register the targets with.
+<4> Specify the Route53 public zone ID, which as a format similar to
+`Z21IXYZABCZ2A4`. You can obtain this value from the AWS console.
+<5> The Route53 zone to register the targets with.
+<6> Specify the Route53 base domain that you used when you generated the
+`install-config.yaml` file. Do not include the trailing period (.) that is
```
skordas: the same like in line 68

modules/installation-creating-aws-security.adoc
```diff
@@ -0,0 +1,101 @@
...
+    "ParameterValue": "subnet-0dfe1c9412a54d352" <6>
+  },
+  {
+    "ParameterKey": "VpcId", <7>
+    "ParameterValue": "vpc-055h956b1e6572be9" <8>
+  }
+]
+----
+<1> A short, representative cluster name to use for hostnames, etc.
+<2> Specify the cluster name that you used when you generated the
```
skordas: cluster name is also used when user decide to generate only ignition-configs files.

modules/installation-creating-aws-bootstrap.adoc
```diff
@@ -0,0 +1,182 @@
...
+  {
+    "ParameterKey": "InternalServiceTargetGroupArn", <23>
+    "ParameterValue": "arn:aws:elasticloadbalancing:<zone>:269733383066:targetgroup/<dns_stack>-Inter-1INRIWR8L3627/4cd9e764bb418f47" <24>
+  }
+]
+
+----
+<1> A short, representative cluster name to use for hostnames, etc.
+<2> Specify the cluster name that you used when you generated the
+`install-config.yaml` file.
```
skordas: cluster name is also used when user decide to generate only ignition-configs files.

modules/installation-creating-aws-bootstrap.adoc
```diff
@@ -0,0 +1,182 @@
...
+You must enter the command on a single line.
+====
++
+----
+$ aws cloudformation create-stack --stack-name <name> <1>
+     --template-body file://<template>.yaml <2>
+     --parameters file:///<parameters>.json <3>
+     --capabilities CAPABILITY_NAMED_IAM
+----
+<1> `<name>` is the name for the CloudFormation stack, such as `cluster-dns`.
```
skordas: 'cluster-dns' -> `cluster-bootstrap'

modules/installation-creating-aws-control-plane.adoc
```diff
@@ -0,0 +1,211 @@
...
+    "ParameterValue": "arn:aws:elasticloadbalancing:<zone>:269733383066:targetgroup/<dns_stack>-Inter-1Q3IAOSC8ZNNM/8e85701acf15632d" <34>
+  },
+  {
+    "ParameterKey": "InternalServiceTargetGroupArn", <35>
+    "ParameterValue": "arn:aws:elasticloadbalancing:<zone>:269733383066:targetgroup/<dns_stack>-Inter-1INRIWR8L3627/4cd9e764bb418f47" <36>
+  }
+]
+----
+<1> A short, representative cluster name to use for hostnames, etc.
+<2> Specify the cluster name that you used when you generated the
```
skordas: cluster name is also used when user decide to generate only ignition-configs files.

modules/installation-overview.adoc
kalexand-rh: Without a second auxiliary verb, the "are" applies to instructed, too, but I'll make it explicit for clarity. Thanks!

modules/installation-creating-aws-worker.adoc
```diff
@@ -0,0 +1,166 @@
...
+  },
+  {
+    "ParameterKey": "InternalServiceTargetGroupArn", <35>
+    "ParameterValue": "1" <36>
+  }
+]
+----
+<1> A short, representative cluster name to use for hostnames, etc.
+<2> Specify the cluster name that you used when you generated the
+`install-config.yaml` file.
```
skordas: cluster name is also used when user decide to generate only ignition-configs files.

modules/installation-creating-aws-worker.adoc
```diff
@@ -0,0 +1,166 @@
...
+* `r4.16xlarge`
+
+. Launch the template:
++
+----
+$ aws cloudformation create-stack --stack-name <name> \ <1>
+     --template-body file://<template>.yaml \ <2>
+     --parameters file:///<parameters>.json <3>
+----
+<1> `<name>` is the name for the CloudFormation stack, such as `cluster-dns`.
```
skordas: `cluster-dns` -> `cluster-worker`

modules/installation-creating-aws-control-plane.adoc
```diff
@@ -0,0 +1,211 @@
...
+====
+You must enter the command on a single line.
+====
++
+----
+$ aws cloudformation create-stack --stack-name <name> <1>
+     --template-body file://<template>.yaml <2>
+     --parameters file:///<parameters>.json <3>
+----
+<1> `<name>` is the name for the CloudFormation stack, such as `cluster-dns`.
```
skordas: `cluster-dns` -> `cluster-master`

modules/installation-aws-upi-requirements.adoc
kalexand-rh: I'll rewrite the into and link to the fine-grained permissions. I'm opening a new issue to update the table of permissions.

modules/installation-overview.adoc
```diff
@@ -7,50 +7,148 @@
...
 is a temporary control plane that launches the rest of the cluster and then is
-destroyed.
-
-The main assets generated by the installation program are the Ignition configurations for the
-bootstrap, master, and worker machines. With these three configurations and correctly
-configured infrastructure, you can start an {product-title} cluster.
+destroyed. Both UPI and IPI clusters use the same bootstrap process to initialize the
+cluster.
 
 Bootstrapping a cluster involves the following steps:
```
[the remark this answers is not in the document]
kalexand-rh: @jianlinliu, that's an excellent point. I'm adding "(Requires manual intervention for UPI)"  to the bootstrap steps that need it. Please let me know if this looks better now.

installing/installing_aws_upi/installing-aws-upi.adoc
```diff
@@ -0,0 +1,81 @@
...
+
+include::modules/installation-overview.adoc[leveloffset=+1]
+
+include::modules/installation-aws-upi-requirements.adoc[leveloffset=+1]
+
+include::modules/installation-obtaining-installer.adoc[leveloffset=+1]
+
+include::modules/installation-provide-credentials.adoc[leveloffset=+1]
+
+include::modules/installation-initializing.adoc[leveloffset=+1]
```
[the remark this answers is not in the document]
kalexand-rh: I'm removing the modules that @skordas called out and moving the UPI-relevant info to the topic about creating Ignition configs.

installing/installing_aws_upi/installing-aws-upi.adoc
```diff
@@ -0,0 +1,81 @@
...
+include::modules/installation-cloudformation-bootstrap.adoc[leveloffset=+2]
+
+include::modules/installation-creating-aws-control-plane.adoc[leveloffset=+1]
+
+include::modules/installation-cloudformation-control-plane.adoc[leveloffset=+2]
+
+[id="installing-workers-aws-upi"]
+== Creating worker nodes
+
+You can either manually create worker nodes before you install the cluster or
```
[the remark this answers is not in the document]
kalexand-rh: I think that this is still before you install the cluster. You haven't bootstrapped the control plane yet.

modules/installation-aws-upi-bootstrap.adoc
kalexand-rh: @wking, it has, and I think that if "machine" is really the right term, then it can stay. (I'm lumping it in the same category as "desired" - the style guide says "no," but Kubernetes trumps it.)

modules/installation-aws-upi-bootstrap.adoc
kalexand-rh: "bootstrap-complete" won

wking: > If you expect openshift/installer#1568 to land...

We've abandoned that one :p, so you should be clear to adapt this line to the already-landed e openshift/installer#1552 now.

modules/installation-aws-upi-installation.adoc
kalexand-rh: I think they're working on it, but I'll track the bug and follow up when it's resolved.

installing/installing_aws_upi/installing-aws-upi.adoc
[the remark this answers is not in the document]
wking: If you create compute nodes before the cluster, they will just sit waiting for the cluster to come up (either waiting for it to serve them Ignition configs, or for the kubelet to be able to attach to the control plane).  So you *can* create them before, but there's not much point.  Probably either drop timing wording entirely, or make it "after" so folks don't wonder why their compute node is hung.

modules/installation-creating-aws-control-plane.adoc
```diff
@@ -0,0 +1,211 @@
...
+* Configure an AWS account.
+* Generate the ignition config files for your cluster.
+* Create and configure a VPC and assocated subnets in AWS.
+* Create and configure DNS, load balancers, and listeners in AWS.
+* Create control plane and worker roles.
+* Create the bootstrap node.
+
+.Procedure
+
+. Upload the master ignition config file to the S3 bucket that you created for
```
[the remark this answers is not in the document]
wking: > ... seem like not, user could specify https://api.<cluster_name>.<domain_name>:22623/config/master for the ignition file.

Yeah, for IPI clusters, the only Ignition config that ends up on S3 is the bootstrap Ignition config.

installing/installing_aws_upi/installing-aws-upi.adoc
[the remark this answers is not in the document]
kalexand-rh: I'm moving the worker node creation to after the bootstrap completes and changing the wording.

modules/installation-creating-aws-control-plane.adoc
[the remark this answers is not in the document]
kalexand-rh: For IPI or UPI?

modules/installation-creating-aws-worker.adoc
kalexand-rh: I'm going to create a new module based on https://docs.openshift.com/container-platform/4.0/machine_management/creating-machineset.html to include or link to.

kalexand-rh:
I've pushed updates for all of the resolved comments. Thank you all! I'm going to follow up on the remaining comments and will be happy to make more changes.

modules/installation-aws-upi-installation.adoc
```diff
@@ -0,0 +1,187 @@
...
+----
+$ oc get machines --namespace openshift-machine-api
+NAME                                 INSTANCE              STATE     TYPE         REGION      ZONE         AGE
+test-tkh7l-worker-us-east-2a-hxlqn   i-0e7f3a52b2919471e   pending   m4.4xlarge   us-east-2   us-east-2a   3s
+----
+
+. Complete the cluster installation:
++
+----
+$ bin/openshift-install wait-for finish
```
jianlinliu: Per latest installer sub-command, should replace 'finish' with 'cluster-ready'
`
./openshift-install version
./openshift-install v4.1.0-201904101232-dirty
built from commit 741d4db20a2d673eb79cf6dbd076001c9cb65753
release image registry.svc.ci.openshift.org/ocp/release:4.0.0-0.nightly-2019-04-10-182914

 ./openshift-install wait-for --help
<--snip-->
Available Commands:
  bootstrap-complete Wait until cluster bootstrapping has completed
  cluster-ready      Wait until the cluster is ready
<--snip-->
`

kalexand-rh: Thank you! I've updated the command.

modules/installation-aws-upi-rhcos-ami.adoc
```diff
@@ -0,0 +1,66 @@
...
+//
+// * installing/installing_aws_upi/installing-aws-upi.adoc
+
+[id="installation-aws-upi-rhcos-ami-{context}"]
+= {op-system} AMIs for the AWS UPI infrastructure
+
+You must use a valid {op-system-first} AMI for your Amazon Web Services
+(AWS) zone for your {product-title} nodes.
+
+.{op-system} AMIs
```
kalexand-rh: #forum-coreos says that these AMIs are out of date and to use https://github.com/openshift/installer/pull/1612/files

modules/installation-aws-upi-installation.adoc
```diff
@@ -0,0 +1,206 @@
...
+. View the machines in the `openshift-machine-api` namespace and confirm that
+they are launching:
++
+----
+$ oc get machines --namespace openshift-machine-api
+NAME                                 INSTANCE              STATE     TYPE         REGION      ZONE         AGE
+test-tkh7l-worker-us-east-2a-hxlqn   i-0e7f3a52b2919471e   pending   m4.4xlarge   us-east-2   us-east-2a   3s
+----
+
+. Review the pending certificate signing requests (CSRs) and ensure that the
```
kalexand-rh: @jianlinliu, re: https://bugzilla.redhat.com/show_bug.cgi?id=1699293

kalexand-rh:
@jianlinliu, @skordas, I've made all of the changes that I'm aware of except for addressing https://bugzilla.redhat.com/show_bug.cgi?id=1697968. Will you PTAL and let me know what other changes need to be made?

@openshift/team-documentation, will you please review?

modules/installation-creating-aws-bootstrap.adoc
```diff
@@ -0,0 +1,182 @@
...
+    "ParameterValue": "arn:aws:elasticloadbalancing:<zone>:269733383066:targetgroup/<dns_stack>-Inter-1Q3IAOSC8ZNNM/8e85701acf15632d" <22>
+  },
+  {
+    "ParameterKey": "InternalServiceTargetGroupArn", <23>
+    "ParameterValue": "arn:aws:elasticloadbalancing:<zone>:269733383066:targetgroup/<dns_stack>-Inter-1INRIWR8L3627/4cd9e764bb418f47" <24>
+  }
+]
+
+----
+<1> A short, representative cluster name to use for hostnames, etc.
```
ahardin-rh: host names (two words per IBM style)

modules/installation-creating-aws-bootstrap.adoc
```diff
@@ -0,0 +1,182 @@
...
+parameters JSON file.
+
+. Confirm that the template components exist:
++
+----
+$ aws cloudformation describe-stacks --stack-name <name>
+----
++
+After the `StackStatus` displays `CREATE_COMPLETE`, the output displays values
+for the following parameters. You need to provide these parameter values to
```
ahardin-rh: s/need to/must

modules/installation-creating-aws-control-plane.adoc
```diff
@@ -0,0 +1,206 @@
...
+    "ParameterKey": "InternalApiTargetGroupArn", <33>
+    "ParameterValue": "arn:aws:elasticloadbalancing:<zone>:269733383066:targetgroup/<dns_stack>-Inter-1Q3IAOSC8ZNNM/8e85701acf15632d" <34>
+  },
+  {
+    "ParameterKey": "InternalServiceTargetGroupArn", <35>
+    "ParameterValue": "arn:aws:elasticloadbalancing:<zone>:269733383066:targetgroup/<dns_stack>-Inter-1INRIWR8L3627/4cd9e764bb418f47" <36>
+  }
+]
+----
+<1> A short, representative cluster name to use for hostnames, etc.
```
ahardin-rh: host names (likely in other spots too)

modules/installation-creating-aws-control-plane.adoc
```diff
@@ -0,0 +1,206 @@
...
+* `r4.4xlarge`
+* `r4.8xlarge`
+* `r4.16xlarge`
+<27> Whether or not to register a network load balancer (NLB).
+<28> Specify `yes` or `no`. If you specify `yes`, you must provide a Lambda
+Amazon Resource Name (ARN) value.
+<29> The ARN for NLB IP target registration lambda group.
+<30> Specify the `RegisterNlbIpTargetsLambda` value from the output of the CloudFormation template for DNS
+and load balancing.
+<31> The ARN for external API load balancer target group
```
ahardin-rh: End each definition in a period, for consistency

modules/installation-creating-aws-dns.adoc
```diff
@@ -0,0 +1,132 @@
...
+parameters JSON file.
+
+. Confirm that the template components exist:
++
+----
+$ aws cloudformation describe-stacks --stack-name <name>
+----
++
+After the `StackStatus` displays `CREATE_COMPLETE`, the output displays values
+for the following parameters. You need to provide these parameter values to
```
ahardin-rh: s/need to/must

modules/installation-creating-aws-security.adoc
```diff
@@ -0,0 +1,101 @@
...
+    "ParameterKey": "PrivateSubnets", <5>
+    "ParameterValue": "subnet-0dfe1c9412a54d352" <6>
+  },
+  {
+    "ParameterKey": "VpcId", <7>
+    "ParameterValue": "vpc-055h956b1e6572be9" <8>
+  }
+]
+----
+<1> A short, representative cluster name to use for hostnames, etc.
```
ahardin-rh: host names

modules/installation-creating-aws-security.adoc
```diff
@@ -0,0 +1,101 @@
...
+<2> Specify the cluster name that you used when you generated the Ignition
```
ahardin-rh: Ignition configuration files?

modules/installation-creating-aws-security.adoc
```diff
@@ -0,0 +1,101 @@
...
+parameters JSON file.
+
+. Confirm that the template components exist:
++
+----
+$ aws cloudformation describe-stacks --stack-name <name>
+----
++
+After the `StackStatus` displays `CREATE_COMPLETE`, the output displays values
+for the following parameters. You need to provide these parameter values to
```
ahardin-rh: s/need to/must

modules/installation-creating-aws-vpc.adoc
```diff
@@ -0,0 +1,88 @@
...
+
+. Confirm that the template components exist:
++
+----
+$ aws cloudformation describe-stacks --stack-name <name>
+----
++
+After the `StackStatus` displays `CREATE_COMPLETE`, the output displays values
+for the following parameters. You need to provide these parameter values to
+the other CloudFormation templates that you run to create your cluster:
```
ahardin-rh: s/need to/must

modules/installation-creating-aws-worker.adoc
```diff
@@ -0,0 +1,155 @@
...
+    "ParameterKey": "InternalApiTargetGroupArn", <33>
+    "ParameterValue": "1" <34>
+  },
+  {
+    "ParameterKey": "InternalServiceTargetGroupArn", <35>
+    "ParameterValue": "1" <36>
+  }
+]
+----
+<1> A short, representative cluster name to use for hostnames, etc.
```
ahardin-rh: host names

modules/installation-creating-aws-worker.adoc
```diff
@@ -0,0 +1,155 @@
...
+<2> Specify the cluster name that you used when you generated the Ignition
+configs for the cluster.
```
ahardin-rh: Ignition configuration files?

modules/installation-machines-bare-metal.adoc
```diff
@@ -0,0 +1,99 @@
...
+for your cluster.
+
+.Procedure
+
+. Obtain the {op-system-first} images.
+
+. Use the the ignition config files that you generated by using the installation
+program to create the required bootstrap, worker, and master nodes.
+
+. Monitor for bootstrap-complete
```
ahardin-rh: Maybe reword this step to say that you should watch for the *bootstrap-complete* success message to confirm that the bootstrap process completed?

modules/installation-overview.adoc
```diff
@@ -7,57 +7,160 @@
...
+components
+* Cluster registration, which associates the cluster identity to your Red Hat
+account to facilitate the gathering of usage metrics
+
+In {product-title} 4, the installation program is a Go binary file that performs a
+series of file transformations on a set of assets. The way you interact with the
+installation program differs depending on your installation type.
+
+* When you use IPI, you delegate the infrastructure bootstrapping and provisioning
+to the installation program
```
ahardin-rh: possible strange line breaks within these two bulleted paragraphs

modules/installation-overview.adoc
```diff
@@ -7,57 +7,160 @@
...
+bootstrap node. You must run the installation program to generate the required
+configuration files and then run the installation program again to deploy a cluster
+to your configured infrastructure.
+
+You use three sets of files during installation: an installation configuration
+file, Kubernetes manifests, and Ignition configurations for your machine types.
+
+[IMPORTANT]
+====
+It is possible to modify Kubernetes and the Ignition Configs that control
```
ahardin-rh: Ignition configuration files

modules/installation-overview.adoc
```diff
@@ -7,57 +7,160 @@
...
+the underlying {op-system-first} operating system during installation. However,
+no validation is available to confirm the suitability of any modifications that
+you make to these objects. If you modify these objects, you might render
+your cluster non-functional. Because of this risk, modifying Kubernetes and
+Ignition Configs is not supported unless you are following documented procedures
```
ahardin-rh: Ignition configuration files

modules/installation-overview.adoc
```diff
@@ -7,57 +7,160 @@
...
+or are instructed to do so by Red Hat support.
+====
+
+The installation configuration file is transformed into Kubernetes manifests, and
+then the manifests are wrapped into Ignition configurations. The installation program uses
+these Ignition configurations to create the cluster.
+
+The install configuration files are all pruned when you run the installation program,
```
ahardin-rh: s/install/installation

modules/installation-aws-upi-bootstrap.adoc
```diff
@@ -0,0 +1,34 @@
...
+
+After you create all of the required infrastructure in Amazon Web Services (AWS),
+you can install the cluster.
+
+.Prerequisites
+
+* Configure an AWS account.
+* Generate the Ignition config files for your cluster.
```
ahardin-rh: We tend to spell out "configuration files" instead of just saying "config". I see this usage throughout the files in this PR, so we should probably pick a way. I don't see any guidance for our team in how we'll treat this in 4.x docs, unless I missed it somewhere.

kalexand-rh: I know. :/ The references in the CoreOS docs are to "Ignition configs" more often than not, and neither Chris nor I have ever heard anyone refer to them as "Ignition configuration." I think we might need another exception, like the always-lowercase "etcd."

modules/installation-machines-bare-metal.adoc
kalexand-rh: This module shouldn't be in this PR, so I'm fixing this on the bare metal PR. Thanks!

modules/installation-aws-upi-bootstrap.adoc
kalexand-rh: I had a chat with Vikram, and I'm adding "Ignition config" to the glossary. https://github.com/openshift/openshift-docs/pull/14461

mazzystr:
@kalexand-rh  Is there a URL to view these docs rendered in stage?  I tried `https://docs.openshift.com/container-platform/4.0/` but I don't see the updates.

kalexand-rh:
@mazzystr, it looks like Netlify isn't updating the preview draft anymore. Please use this link: http://file.rdu.redhat.com/kalexand/041719/osdocs323/installing/installing_aws_upi/installing-aws-upi.html

mazzystr:
@kalexand-rh I'm in Solution Engineering under Itamar and Vijay.  Our team has written many OCP on <provider> reference architectures and the Day 2 Section.  I'm the teams SME for OCP on AWS.  If you don't mind I'm going to peer review your pull.  :)

kalexand-rh:
@mazzystr, thank you so much! I'd love to hear your feedback so that I can improve the context. (I also have a PR up for bare metal and will have one up for vSphere soon.)

modules/installation-creating-aws-worker.adoc
```diff
@@ -0,0 +1,159 @@
...
+* `r4.xlarge`
+* `r4.2xlarge`
+* `r4.4xlarge`
+* `r4.8xlarge`
+* `r4.16xlarge`
+
+. Launch the template:
++
+----
+$ aws cloudformation create-stack --stack-name <name> \ <1>
```
mazzystr: The order of events is confusing here.  Can we bring the example template up above executing the `aws cloudformation create-stack ...`?   Maybe at minimum create a hyperlink to the example template?  Customers will drive serially through this procedure and execute the `aws cloudformation create-stack ...` before the `<template>.yaml` file is created.

modules/installation-extracting-infraid.adoc
```diff
@@ -0,0 +1,31 @@
+// Module included in the following assemblies:
```
mazzystr: Shouldn't this file be called ...-clusterid or ...-clustername?

modules/machineset-manually-scaling.adoc
```diff
@@ -11,7 +11,7 @@ manually scale the MachineSet.
 .Prerequisites
 
 * Install an {product-title} cluster and the `oc` command line.
-* Log into `oc` as a user with <some permission>.
+* Log into `oc` as a user with `cluster-admin` permission.
```
mazzystr: Out of scope for UPI on AWS.  This should be a separate PR.

modules/installation-creating-aws-worker-machineset.adoc
```diff
@@ -0,0 +1,150 @@
+// Module included in the following assemblies:
```
mazzystr: We're no longer in aws land so how about a filename of installation-creating-ocp-worker-machineset.adoc

modules/machineset-creating.adoc
```diff
@@ -6,13 +6,15 @@
 = Creating a MachineSet
 
 You can create more MachineSets. Because the MachineSet definition contains
-details that are specific to the AWS region that the cluster is deployed in,
-you copy an existing MachineSet from your cluster and modify it.
+details that are specific to the Amazon Web Services (AWS) region that the
+cluster is deployed in, you copy an existing MachineSet from your cluster and
+modify it.
```
mazzystr: Out of scope for UPI on AWS.  This should be a separate PR.

modules/machineset-creating.adoc
```diff
@@ -6,13 +6,15 @@
...
 
 .Prerequisites
 
 * Deploy an {product-title} cluster.
-* Install the `oc` command line and log in as a user with <some permissions>.
+* Install the `oc` command line and log in as a user with `cluster-admin`
+permission.
```
mazzystr: Out of scope for UPI on AWS.  This should be a separate PR.

mazzystr (commented):
I think that's about it for now.  I'm going to run through the entire thing again this afternoon to ensure that all commands actually work.  :)

installing/installing_aws/installing-aws-customizations.adoc
```diff
@@ -15,8 +15,6 @@ to host the cluster.
 
 include::modules/installation-overview.adoc[leveloffset=+1]
 
-include::modules/installation-clouds.adoc[leveloffset=+1]
-
 include::modules/installation-provide-credentials.adoc[leveloffset=+1]
```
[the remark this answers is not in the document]
kalexand-rh: @mazzystr, will you try that link again? I'm happy to add that bullet point.

modules/installation-aws-upi-rhcos-ami.adoc
```diff
@@ -0,0 +1,66 @@
+// Module included in the following assemblies:
```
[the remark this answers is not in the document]
kalexand-rh: @wking, what do you think of this approach? Do you think this is fair until you get [this](https://github.com/openshift/installer/issues/1399) RFE about obtaining images in?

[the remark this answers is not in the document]
wking: > This has risk of directing the customer to deploy buggy versions of OCP 4.x.

Folks using a stale bootimage AMI might be exposed to bugs in their bootstrap machine, but that's only up for a few dozen minutes, so the risk seems minimal.  Once the cluster is up, the machine-config daemon running on all Linux nodes will detect RHCOS and pivot to the release-image specified OSContainer (the machine-os-content referenced by the release image), so again, the window of potential buggy-RHCOS exposure is small.

I'm agnostic on sourcing these AMIs until we can pull the bootimage metadata from the release image itself (e.g. openshift/installer#1286).  The installer uses a hard-coded list of AMIs, so I'm not concerned about the docs taking the same approach.  But the owner ID and `rhcos-4*` descriptions are probably going to be stable as well, so I don't have a problem if the docs want to go that way either.

modules/installation-creating-aws-vpc.adoc
```diff
@@ -0,0 +1,88 @@
...
++
+[IMPORTANT]
+====
+You must enter the command on a single line.
+====
++
+----
+$ aws cloudformation create-stack --stack-name <name> <1>
+     --template-body file://<template>.yaml <2>
+     --parameters file:///<parameters>.json <3>
```
mazzystr: The prior template-body  file location uses // for present working directory.  The <parameters>.json file should follow suit.

mazzystr (changes requested):

modules/installation-creating-aws-bootstrap.adoc
```diff
@@ -0,0 +1,183 @@
...
+
+. Launch the template:
++
+[IMPORTANT]
+====
+You must enter the command on a single line.
+====
++
+----
+$ aws cloudformation create-stack --stack-name <name> <1>
```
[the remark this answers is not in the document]
kalexand-rh: We can't put links in individual modules. It's a known issue that we hope we can address by the mod docs guidelines at some point.

For each of these files, I'm moving the step to create the template after the step to create the parameter file and adding instruction to copy the file from the template, which should be less than one "screen" away from the step.

modules/installation-creating-aws-worker-machineset.adoc
kalexand-rh: Depending on what Trevor decides, I might be pulling this file from the PR. For now, it's a UPI-only task, and I'll update the filename.

modules/installation-extracting-infraid.adoc
kalexand-rh: It's called "infraID" when you extract it from the metadata and "InfrastructureName" in the AWS templates.

modules/machineset-creating.adoc
kalexand-rh: Depending on how we land with the "pull MachineSets from UPI" or "fix MachineSets after UPI install" debate, this might be in scope.

modules/installation-creating-aws-dns.adoc
```diff
@@ -0,0 +1,136 @@
...
+computer.
+
+. Create a JSON file that contains the parameter values that the template
+requires:
++
+[source,json]
+----
+[
+  {
+    "ParameterKey": "ClusterName", <1>
```
mazzystr: What is the difference between ClusterName and InfrastructureName?  Are these two values different than `jq -r .infraID /some/dir/openshift-abcxyz/metadata.json`?

modules/installation-creating-aws-dns.adoc
```diff
@@ -0,0 +1,136 @@
...
+    "ParameterValue": "mycluster" <2>
+  },
+  {
+    "ParameterKey": "InfrastructureName", <3>
```
mazzystr: What is the difference between ClusterName and InfrastructureName?  Are these two values different than `jq -r .infraID /some/dir/openshift-abcxyz/metadata.json`?

modules/installation-creating-aws-dns.adoc
```diff
@@ -0,0 +1,136 @@
...
+  {
+    "ParameterKey": "HostedZoneId", <5>
+    "ParameterValue": "Z21IXYZABCZ2A4" <6>
+  },
+  {
+    "ParameterKey": "HostedZoneName", <7>
+    "ParameterValue": "example.com" <8>
+  },
+  {
+    "ParameterKey": "PublicSubnets", <9>
```
mazzystr: This is bad news. We should not make customers stop this procedure to reach into AWS to retrieve resource ids. AWS has a procedure for cross-stack communication without a relationship. The link is here. We should remove this parameter.

Example:
"SubnetId" : { "Fn::ImportValue" : {"Fn::Sub": "${NetworkStackName}-SubnetID" } }

modules/installation-creating-aws-dns.adoc
```diff
@@ -0,0 +1,136 @@
...
+    "ParameterValue": "subnet-09c2adf329727bc59" <10>
+  },
+  {
+    "ParameterKey": "PrivateSubnets", <11>
```
mazzystr: This is bad news. We should not make customers stop this procedure to reach into AWS to retrieve resource ids. AWS has a procedure for cross-stack communication without a relationship. The link is here. We should remove this parameter.

Example:
"SubnetId" : { "Fn::ImportValue" : {"Fn::Sub": "${NetworkStackName}-SubnetID" } }

modules/installation-creating-aws-dns.adoc
```diff
@@ -0,0 +1,136 @@
...
+    "ParameterValue": "subnet-0dfe1c9412a54d352" <12>
+  },
+  {
+    "ParameterKey": "VpcId", <13>
```
mazzystr: This is bad news. We should not make customers stop this procedure to reach into AWS to retrieve resource ids. AWS has a procedure for cross-stack communication without a relationship. The link is here. We should remove this parameter.

Example:
"SubnetId" : { "Fn::ImportValue" : {"Fn::Sub": "${NetworkStackName}-SubnetID" } }

modules/installation-creating-aws-dns.adoc
```diff
@@ -0,0 +1,136 @@
...
+    "ParameterValue": "vpc-055h956b1e6572be9" <14>
+  }
```
mazzystr: Please add Parameter VpcStackName

modules/installation-cloudformation-dns.adoc
```diff
@@ -0,0 +1,385 @@
...
+  PublicSubnets:
+    Description: The internet-facing subnets
+    Type: List<AWS::EC2::Subnet::Id>
+  PrivateSubnets:
+    Description: The internal subnets
+    Type: List<AWS::EC2::Subnet::Id>
+  VpcId:
+    Description: The VPC created resources will belong.
+    Type: AWS::EC2::VPC::Id
+
```
mazzystr: Please add parameter VpcStackName

mazzystr (changes requested):

kalexand-rh:
@cuppett, will you PTAL at the feedback that @mazzystr left about the CloudFormation templates in https://github.com/openshift/openshift-docs/pull/14241#pullrequestreview-227961871?

modules/installation-cloudformation-dns.adoc
```diff
@@ -0,0 +1,385 @@
...
+objects and load balancers that you need for your {product-title} cluster.
+
+
+[source,yaml]
+----
+AWSTemplateFormatVersion: 2010-09-09
+Description: Template for Openshift Cluster UPI Network Elements (Route53 & LBs)
+
+Parameters:
+  ClusterName:
```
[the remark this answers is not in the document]
cuppett: ClusterName is the value you type into the openshift-installer "chris". InfrastructureName has the generated suffix, "chris-xyz12"

modules/installation-cloudformation-dns.adoc
```diff
@@ -0,0 +1,385 @@
...
+    AllowedPattern: ^([a-zA-Z][a-zA-Z0-9\-]{0,31})$
+    MaxLength: 32
+    MinLength: 1
+    ConstraintDescription: Cluster name must be alphanumeric, start with a letter and a maximum of 32 characters
+    Description: A short, representative cluster name to use for hostnames, etc.
+    Type: String
+  InfrastructureName:
```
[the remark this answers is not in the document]
cuppett: Same as above.

modules/installation-cloudformation-dns.adoc
```diff
@@ -0,0 +1,385 @@
...
+    Description: A short, unique cluster ID used to tag cloud resources and identify items owned/used by the cluster.
+    Type: String
+  HostedZoneId:
+    Description: The Route53 public zone ID to register the targets with (e.g Z21IXYZABCZ2A4)
+    Type: String
+  HostedZoneName:
+    Description: The Route53 zone to register the targets with (No trailing dot - e.g. mycorp.com)
+    Type: String
+    Default: "example.com"
+  PublicSubnets:
```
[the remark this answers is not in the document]
cuppett: These stacks do help feed each other, but they are not pre-requisites for each other. They are not required to use 01_vpc and may have an existing VPC. Also, stack linking like this is still extremely fragile in AWS and updates to stacks linked this way do not cascade changes as you'd hope.

modules/installation-cloudformation-dns.adoc
```diff
@@ -0,0 +1,385 @@
...
+    Description: The internet-facing subnets
+    Type: List<AWS::EC2::Subnet::Id>
+  PrivateSubnets:
```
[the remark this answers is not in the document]
cuppett: Same as above.

modules/installation-cloudformation-dns.adoc
```diff
@@ -0,0 +1,385 @@
...
+    Description: The internal subnets
+    Type: List<AWS::EC2::Subnet::Id>
+  VpcId:
```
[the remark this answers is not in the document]
cuppett: Same as above.

modules/installation-creating-aws-dns.adoc
cuppett: Same as above.

modules/installation-creating-aws-dns.adoc
cuppett: Same as above.

modules/installation-creating-aws-dns.adoc
cuppett: It's not needed in the template and similar comments as above.

modules/installation-cloudformation-dns.adoc
cuppett: Same as above.

cuppett (commented):
> @cuppett, will you PTAL at the feedback that @mazzystr left about the CloudFormation templates in [#14241 (review)](https://github.com/openshift/openshift-docs/pull/14241#pullrequestreview-227961871)?

Done.

modules/installation-creating-aws-dns.adoc
cuppett: Same as above.

modules/installation-cloudformation-control-plane.adoc
```diff
@@ -0,0 +1,386 @@
...
+    Description: The subnets (recommend private) to launch the master nodes into
+    Type: AWS::EC2::Subnet::Id
+  MasterSecurityGroupId:
+    Description: The master security group ID to associate with master nodes.
+    Type: AWS::EC2::SecurityGroup::Id
+  IgnitionLocation:
+    Default: https://api.$CLUSTER_NAME.$DOMAIN:22623/config/master
+    Description: Location to fetch bootstrap ignition from. (Recommend to use the autocreated ignition config location.)
+    Type: String
+  CertificateAuthorities:
```
mazzystr: Certificate Authority x.509 certificate is stored unencrypted and available to all users including AWS support.  This absolutely cannot happen.

We have to think of another way.

mazzystr (changes requested):

cuppett:
That makes sense. Can make this a **** kind of field. Can also do it with
SSM parameter. However today it goes in UserData and does within the
template too, so I am not sure what it buys us...

Steve
Mobile

mazzystr:
An option would be to scp it manually to the instance.  UserData can have bash `! while -f ca.pem; do sleep 5; done` loop to pause until the file is present.

cuppett:
Not unique to AWS UPI. All the installs will have this issue. See email.
Get a BZ going. Good catch, BTW.

Steve
Mobile

modules/installation-cloudformation-control-plane.adoc
abhinavdahiya: > Certificate Authority x.509 certificate is stored unencrypted and available to all IAM users in the account including AWS support. This absolutely cannot happen.

this is the trust certificate to trust the server that will serve the ignition. what's the problem with this being available to other users...?

wking: Just to underline @abhinavdahiya's point, this is the certificate, not the key.  The keys get pushed up to S3, but aren't exposed in the CloudFormation template.  I don't see a problem with leaking certificates.

mazzystr:
Folks we are officially bugzilla'd - BZ [1701408](https://bugzilla.redhat.com/show_bug.cgi?id=1701408)

jianlinliu:
@kalexand-rh Should we have something like `delete cluster` section for upi-on-aws?

modules/installation-cloudformation-control-plane.adoc
mazzystr: Keep in mind the cert is in the wild in at least two places.  1) S3 ... since the bucket and object both have no encryption for data at rest the data is replicated throughout the AWS S3 ecosystem. It is unknown who has access to this ecosystem.  2) CloudFormation ... The data is exposed as a parameter value.  The data is replicated throughout the AWS CloudFormation backend.  It is unknown who has access to this ecosystem.

I'm less concerned about the IAM users of the account having access than AWS world wide support possibly having access.

kalexand-rh:
@jianlinliu, eventually. It's a different user story from installation, though. I have a [JIRA](https://jira.coreos.com/browse/OSDOCS-405) to figure out all of the processes that will be different between UPI and IPI.

modules/installation-cloudformation-control-plane.adoc
wking: > I'm less concerned about the IAM users of the account having access than AWS world wide support possibly having access.

But again, what is your concern with a public certificate at all?  Say a malicious Mallory acquires the cert.  What are you concerned about him doing with it.  He can't impersonate the Ignition-config server unless he *also* has the key, so leaking certs doesn't seem like a problem to me.

modules/installation-cloudformation-control-plane.adoc
```diff
@@ -0,0 +1,386 @@
...
+    Default: data:text/plain;charset=utf-8;base64,ABC...xYz==
+    Description: Base64 encoded certificate authority string to use.
+    Type: String
+  MasterInstanceProfileName:
+    Description: IAM profile to associate with master nodes.
+    Type: String
+  MasterInstanceType:
+    Default: m4.xlarge
+    Type: String
+    AllowedValues:
```
mazzystr: Latest generation of EC2 instances aren't allowed (m5, c5, r5).  It might be interesting to allow AMD based instances as well.

modules/installation-creating-aws-control-plane.adoc
```diff
@@ -0,0 +1,208 @@
...
+`https://api.<cluster_name>.<domain_name>:22623/config/master`.
+<21> The base64 encoded certificate authority string to use.
+<22> Specify the value from the `master.ign` file that is in the installation
+directory. This value is the long string with the format
+`data:text/plain;charset=utf-8;base64,ABC...xYz==`.
+<23> The IAM profile to associate with master nodes.
+<24> Specify the `MasterInstanceProfile` parameter value from the output of
+the CloudFormation template for the security group and roles.
+<25> The type of AWS instance to use for the control plane machines.
+<26> Allowed values:
```
mazzystr: Latest generation of EC2 instances aren't allowed (m5, c5, r5).  It might be interesting to allow AMD based instances as well.

modules/installation-cloudformation-control-plane.adoc
cuppett: Correct. There are still outstanding issues with nitro-based instance classes.

https://bugzilla.redhat.com/show_bug.cgi?id=1658237

https://github.com/coreos/bugs/issues/2371

cc @gnufied

modules/installation-creating-aws-control-plane.adoc
cuppett: Same as above.

modules/installation-aws-upi-bootstrap.adoc
```diff
@@ -0,0 +1,34 @@
...
+* Create the control plane machines.
+* If you plan to manually manage the worker machines, create the worker machines.
+
+.Procedure
+
+. Change to the directory that contains the installation program and run the
+following command:
++
+----
+$ ./openshift-install wait-for bootstrap-complete --dir=<installation-directory> <1>
```
mazzystr: Incorrect command...
`$ bin/openshift-install wait-for bootstrap-complete --dir=/tmp/openshift/mycluster
FATA[0001] Error executing openshift-install: unknown command "wait-for" for "openshift-install"`

Maybe it should be...
`bin/openshift-install user-provided-infrastructure bootstrap-complete --dir=/tmp/openshift/mycluster`

modules/installation-cloudformation-control-plane.adoc
wking: The installer currently falls...forward to m5 for regions that lack m4.  So if we don't support m5, we'll need to adjust that.

wking: https://github.com/openshift/installer/pull/1163#discussion_r257803519

modules/installation-aws-upi-bootstrap.adoc
wking: Your installer is too old, it's `wait-for` now.

modules/installation-cloudformation-dns.adoc
```diff
@@ -0,0 +1,385 @@
...
+  InternalApiLoadBalancerName:
+    Description: Full name of the Internal API load balancer created.
+    Value: !GetAtt IntApiElb.LoadBalancerFullName
+  ApiServerDnsName:
+    Description: Full hostname of the API server - Needed for ignition configs
+    Value: !Join [".", ["api", !Ref ClusterName, !Ref HostedZoneName]]
+  RegisterNlbIpTargetsLambda:
+    Description: Lambda ARN useful to help register/deregister IP targets for these load balancers
+    Value: !GetAtt RegisterNlbIpTargets.Arn
+  ExternalApiTargetGroupArn:
```
mazzystr: Can we remove Arn from the resource name?  It will help with searching when trying to find values for parameters for the next stack.

modules/installation-cloudformation-dns.adoc
```diff
@@ -0,0 +1,385 @@
...
+    Description: ARN of External API target group
+    Value: !Ref ExternalApiTargetGroup
+  InternalApiTargetGroupArn:
```
mazzystr: Can we remove Arn from the resource name?  It will help with searching when trying to find values for parameters for the next stack.

modules/installation-cloudformation-dns.adoc
```diff
@@ -0,0 +1,385 @@
...
+    Description: ARN of Internal API target group
+    Value: !Ref InternalApiTargetGroup
+  InternalServiceTargetGroupArn:
```
mazzystr: Can we remove Arn from the resource name?  It will help with searching when trying to find values for parameters for the next stack.

modules/installation-cloudformation-control-plane.adoc
gnufied: Actually, The most serious of NVMe issues are fixed in - https://bugzilla.redhat.com/show_bug.cgi?id=1632440 . Our SRE team has confirmed that after deploying fixed kernel they did not receive any "stuck in attach/detach" issue.  

What @cuppett linked above though is  a separate customer running their own Openshift version on AWS.  That bug however is more weird. The steps to reproduce - https://bugzilla.redhat.com/show_bug.cgi?id=1658237#c46 . So basically to reproduce this bug, you have to reboot the node while detach is in progress.  Our Kernel team says - this is a problem with how EBS sends ACPI events to kernel. In practice I think this is less of a problem. We also have some fixes in Openshift to ensure if a node does get stuck with volumes in "attaching/detaching" state, it is removed from scheduling and is tainted.  It must be then drained, rebooted and taint manually removed.

I think we should make sure that - we are running fixed Kernel with proper tuning in RHCOS and start supporting M5/C5 instance types.

cuppett: That is excellent news! We could enable the latest generation instances in the templates.

@abhinavdahiya @crawford Do we think m5 should be our default? The latest generation would be my recommendation in general and continues the trend of improving cost/performance. A downside is they support a total of 28 EBS and ENA attachments where the previous generations could support up to 40 EBS attachments [1].

[1]: https://docs.aws.amazon.com/AWSEC2/latest/WindowsGuide/volume_limits.html

installing/installing_aws/installing-aws-customizations.adoc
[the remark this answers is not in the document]
mazzystr: https://access.redhat.com/documentation/en-us/reference_architectures/2018/html-single/deploying_and_managing_openshift_3.9_on_amazon_web_services/index#client_testing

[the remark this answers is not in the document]
kalexand-rh: Thank you @mazzystr! I've added the test.

modules/installation-aws-upi-bootstrap.adoc
wking: Try [beta4](https://mirror.openshift.com/pub/openshift-v4/clients/ocp/4.1.0-rc.0/).

modules/installation-cloudformation-control-plane.adoc
bartle-stripe: @gnufied It doesn't appear that https://bugzilla.redhat.com/show_bug.cgi?id=1632440 is publicly viewable; could you share some details or open up access to that bug?

modules/installation-cloudformation-control-plane.adoc
mazzystr: Since #1203 is changing the s3 block_public_* things to true I'm less unhappy about this.

modules/installation-creating-aws-control-plane.adoc
```diff
@@ -0,0 +1,208 @@
...
+CloudFormation template for DNS and load balancing.
+<15> A subnet, preferably private, to launch the control plane machines on.
+<16> Specify a subnet from the `PrivateSubnets` value from the output of the
+CloudFormation template for DNS and load balancing.
+<17> The master security group ID to associate with master nodes.
+<18> Specify the `MasterSecurityGroupId` value from the output of the
+CloudFormation template for the security group and roles.
+<19> The location to fetch control plane Ignition config file from.
+<20> Specify the generated Ignition config file location,
+`https://api.<cluster_name>.<domain_name>:22623/config/master`.
```
jianlinliu: s/api/api-int/

modules/installation-creating-aws-worker.adoc
```diff
@@ -0,0 +1,165 @@
...
+<4> Specify an `AWS::EC2::Image::Id` value.
+<5> A subnet, preferably private, to launch the worker nodes on.
+<6> Specify a subnet from the `PrivateSubnets` value from the output of the
+CloudFormation template for DNS and load balancing.
+<7> The worker security group ID to associate with worker nodes.
+<8> Specify the `WorkerSecurityGroupId` value from the output of the
+CloudFormation template for the security group and roles.
+<9> The location to fetch bootstrap Ignition config file from.
+<10> Specify the generated Ignition config location,
+`https://api.<cluster_name>.<domain_name>:22623/config/worker`.
```
jianlinliu: s/api/api-int/

modules/installation-creating-aws-worker.adoc
```diff
@@ -0,0 +1,165 @@
...
+    "ParameterKey": "WorkerSubnet", <5>
+    "ParameterValue": "yes" <6>
+  },
+  {
+    "ParameterKey": "WorkerSecurityGroupId", <7>
+    "ParameterValue": "sg-042ebc4dd65c71d67" <8>
+  },
+  {
+    "ParameterKey": "IgnitionLocation", <9>
+    "ParameterValue": "https://api.<cluster_name>.<domain_name>:22623/config/worker" <10>
```
jianlinliu: s/api/api-int/

modules/installation-creating-aws-control-plane.adoc
```diff
@@ -0,0 +1,208 @@
...
+    "ParameterKey": "Master2Subnet", <15>
+    "ParameterValue": "subnet-0dfe1c9412a54d352" <16>
+  },
+  {
+    "ParameterKey": "MasterSecurityGroupId", <17>
+    "ParameterValue": "sg-0a53cd1337f15u2ee" <18>
+  },
+  {
+    "ParameterKey": "IgnitionLocation", <19>
+    "ParameterValue": "https://api.<cluster_name>.<domain_name>:22623/config/master" <20>
```
jianlinliu: s/api/api-int/

modules/installation-creating-upi-worker-machineset.adoc
```diff
@@ -0,0 +1,144 @@
...
+            creationTimestamp: null
+          placement:
+            availabilityZone: us-east-2a
+            region: us-east-2
+          publicIp: null
+          securityGroups:
+          - filters:
+            - name: tag:Name
+              values:
+              -  test-tkh7l-worker-sg <2>
```
kalexand-rh: @wking, do lines 95 and 117 look right to you?

wking: yup, `{infraName}-{role}-{sg}` matches [this][1].

[1]: https://github.com/openshift/installer/blob/6e8d74889800a841a6ed197c6ec62ea08734fdea/upi/aws/cloudformation/03_cluster_security.yaml#L52-L76

jianlinliu: Are your sure here? Per my test result, https://bugzilla.redhat.com/show_bug.cgi?id=1697968#c11, security group would be created with '{role}-{sg}-{infraName}' as 'GroupName' and empty 'Name'. While machine api is expecting '{infraName}-{role}-{sg}' as 'Name'. Do I miss something?

modules/installation-generate-aws-upi.adoc
```diff
@@ -71,7 +71,7 @@ you can safely ignore this warning.
 .. Remove the files that define the control plane machines:
 +
 ----
-$ rm -f openshift/99_openshift-cluster-api_master-machines-*.yaml
+$ rm -f openshift/99_openshift-cluster-api_master-*.yaml
```
jianlinliu: Per my latest test result, https://bugzilla.redhat.com/show_bug.cgi?id=1698207, I think this change need to be reverted.

kalexand-rh:
Per @abhinavdahiya, I've removed the steps for creating a cluster that uses cluster-managed workers.

kalexand-rh:
@skordas, you said that you got this process to work yesterday, and I made a few updates based on your feedback. Are there any other changes that I need to add now, or does this look ok to you?

modules/installation-creating-aws-bootstrap.adoc
```diff
@@ -0,0 +1,201 @@
...
+Ignition config files for your cluster are served from an S3 bucket. If you
+choose to serve the files from another location, you must modify the templates.
+====
++
+[NOTE]
+====
+The bootstrap Ignition config file does contain secrets, like X.509 keys. If
+uploading the Ignition config file to an S3 bucket does not provide adequate
+security, you can enable an S3 bucket policy to allow only certain users,
+such as the OpenShift IAM user, to access objects that the bucket contains.
```
wking: "If uploading to an S3 bucket is not secure enough, tighten down your S3 bucket" doesn't feel quite right.  How about:

> The bootstrap Ignition config file does contain secrets, like X.509 keys.  The steps below provide basic security for the S3 bucket.  To provide additional security, you can enable an S3 bucket policy to allow only certain users, such as the OpenShift IAM user, to access objects that the bucket contains.  And you can avoid S3 entirely and serve your bootstrap Ignition config from any address reachable from the bootstrap machine.

I'm also fine not enumerating possible S3 tightenings, and saying something like:

> The bootstrap Ignition config file does contain secrets, like X.509 keys.  The steps below provide basic security for the S3 bucket.  If you have security concerns with these steps, you can configure a more-secure S3 bucket or avoid S3 entirely and serve your bootstrap Ignition config from any address reachable from the bootstrap machine.

kalexand-rh: @wking, please check again?

wking: f674f0e20fd6 looks good to me.

kalexand-rh:
QE just approved https://bugzilla.redhat.com/show_bug.cgi?id=1697968, which is the last bug that they considered a blocker for this PR. Per @vikram-redhat, that's enough to merge. Please file bugs for any outstanding issues.

Merging.

modules/installation-aws-upi-requirements.adoc
wking: I'd forgotten to follow up here, but openshift/installer#2296 removed these blocks from the installer's CloudFormation stacks.

PR #14248, opened 2019-03-27
Edits to Configuring network subnets

ahardin-rh:
@sjug Thanks for raising this issue. PTAL.

### DIFF
--- a/modules/configuring-network-subnets.adoc
+++ b/modules/configuring-network-subnets.adoc
@@ -33,12 +33,12 @@ $ cd new-install
 +
 ----
 networking:
-      clusterNetworks:
+      clusterNetwork:
         - cidr: 10.128.0.0/14
-          hostSubnetLength: 9
+          hostPrefix: 23
       machineCIDR: 10.0.0.0/16
-      serviceCIDR: 172.30.0.0/16
-     type: OpenshiftSDN
+      serviceNetwork: 172.30.0.0/16
+      networkType: OpenshiftSDN
 ----
 
 . Consume the customized `install-config.yaml` file to deploy the cluster:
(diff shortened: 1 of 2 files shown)


modules/configuring-network-subnets.adoc
```diff
@@ -33,12 +33,12 @@ $ cd new-install
...
-      clusterNetworks:
+      clusterNetwork:
         - cidr: 10.128.0.0/14
-          hostSubnetLength: 9
+          hostPrefix: 9
       machineCIDR: 10.0.0.0/16
-      serviceCIDR: 172.30.0.0/16
-     type: OpenshiftSDN
+      serviceNetwork: 172.30.0.0/16
+     networkType: OpenshiftSDN
```
sjug: One more space indent (should be at same depth as `clusterNetwork`, `machineCIDR` and `serviceNetwork`)

ahardin-rh: @sjug Fixed!

sjug:
I'm just wondering if we should add the `apiVersion: v1beta4` for this snippet as well?

The previous fields are from `v1beta3` and are now deprecated. If we just keep the docs up to date then we assume that it's the latest API version?

ahardin-rh:
@sjug We need to keep the docs up to date. If the API version changes so frequently, then do we want to state it explicitly here?

sjug:
Good point, lets not change/add it. I was just thinking it gives people a reference point of where in the yaml they are, but they can figure it out.

jboxman:
Also, `hostPrefix` should be `23`: https://github.com/openshift/openshift-docs/pull/14224#discussion_r270504697

ahardin-rh:
@jboxman Good catch. Updated! Merging!

PR #14263, opened 2019-03-28
Changing Operator name back so that it is correct for betas

bergerhoffer:

The name will eventually be changed to "OpenShift Ansible Service Broker Operator", but not until closer to GA. Since we don't want to confuse beta users, am putting the name back for now so that the steps are currently correct.


kalexand-rh (commented):
It LGTM. If it makes sense, you can just revert this commit before GA.

bergerhoffer:
Thanks @kalexand-rh. I'd probably rather not revert - I'd usually save a revert for something that's a true error. Plus keeping history as accurate as possible.

Merging.

PR #14269, opened 2019-03-29
Update the list of cluster configuration resources.

sanchezl:


architecture/customizations.adoc
```diff
@@ -25,27 +25,49 @@ Custom Resource Definitions (CRDs) that are deployed on your cluster.
 |===
 |Configuration resource |Description
 
+|APIServer
+|Configure the API server.
```
deads2k: Kubernetes API server I'd think.

architecture/customizations.adoc
```diff
@@ -25,27 +25,49 @@ Custom Resource Definitions (CRDs) that are deployed on your cluster.
...
 |Network
-|See
+|Configure the cluster networks.
 
 |OAuth
-|See
+|Configure the integrated OAuth server.
+
+|Project
+|Configure how projects are created.
```
deads2k: @sanchezl this one is also ours to doc, you should make sure the jira doc issue has subtasks for it

sanchezl: Opened https://jira.coreos.com/browse/MSTR-356

architecture/customizations.adoc
sanchezl: Updated

deads2k:
@kalexand-rh if we're going to keep this page, you're trying to stay up to date with this list here: https://github.com/openshift/api/blob/master/config/v1/register.go#L32-L64 

@sanchezl you're missing a couple.  Skip ClusterOperator and ClusterVersion

deads2k:
Once that's update, perhaps merge this and expand on the APIServer in a different pull.

sanchezl:
@deads2k I was missing Scheduler, this list now matches.

kalexand-rh:
@deads2k, this list is supposed to match what's showing up in the *Administration* > *Cluster Settings* > *Global Configuration* page. Neither the PR nor the current docs match today's build. Do you know when this list will be stable?

PR #14277, opened 2019-03-29
Additional configuration section for Kubernetes Static Pods

bagnaram:
This section addresses configuration to the control plane static pods


vikram-redhat:
@bagnaram  is this intended for 4.1?

bagnaram:
This targets any release of openshift that utilizes static pods for the control plane. I believe that is >= 3.10

vikram-redhat:
@bagnaram could you squash your commits?

@jiajliu  would you able to confirm that this is correct?

install_config/master_node_configuration.adoc
```diff
@@ -58,6 +58,38 @@ is authoritative. The only way to successfully run only two masters is if you
 install etcd on hosts other than the masters.
 ====
 
+[[master-node-config-kubernetes-infrastructure]]
+== Configuring Kubernetes Infrastructure
+
+Master API pods are defined as static pods by the   
+link:https://docs.okd.io/container-platform/3.11/architecture/infrastructure_components/kubernetes_infrastructure.html#overview[Kubernetes Infrastructure]. In order to make changes to the `api`, `conrollers`, or `etcd` pods, the static pod definitions will need to be changed. These are located in `/etc/origin/node/pods/` on each of the _master hosts_.
```
vikram-redhat: use xref instead of a link

bagnaram: OK thank you for pointing that

PR #14304, opened 2019-04-01
Bump the pods per namespace limit to 25k

chaitanyaenr:
We tested the pods per namespace limit on a 3.11 cluster and found
out that there are no abnormal spikes in the resource usage of control
plane as we increase the number of pods per ns to 25k provided the cluster
has sufficient CPU, memory and disk resources to meet the application
needs. The time taken to create the pods is not exactly linear but it doesn’t 
vary much.

### DIFF
--- a/scaling_performance/cluster_limits.adoc
+++ b/scaling_performance/cluster_limits.adoc
@@ -69,11 +69,13 @@ It does not necessarily mean that the cluster will fail.
 a number of control loops in the system that need to iterate over all objects
 in a given namespace as a reaction to some changes in state. Having a large
 number of objects of a given type in a single namespace can make those loops
-expensive and slow down processing given state changes.]
+expensive and slow down processing given state changes. The limit 
+assumes that the system has enough CPU, memory, and disk to satifty the 
+application requirements.]
 | 3,000
 | 3,000
 | 3,000
-| 3,000
+| 25,000
 
 | Number of services footnoteref:[servicesandendpoints,Each service port and each service back-end has a corresponding entry in iptables. The number of back-ends of a given service impact the size of the endpoints objects, which impacts the size of data that is being sent all over the system.]
 | 10,000
@@ -109,6 +111,12 @@ expensive and slow down processing given state changes.]
 Oversubscribing the physical resources on a node affects resource guarantees the
 Kubernetes scheduler makes during pod placement. Learn what measures you can
 take to xref:../admin_guide/overcommit.adoc#disabling-swap-memory[avoid memory swapping].
+====
+
+[IMPORTANT]
+====
+Some of the limits are streched only in a single dimension, so the they might
+vary when a lot of objects are running on the cluster.
 ====
 
 While


chaitanyaenr:
@ahardin-rh @jeremyeder @akrzos  @ashishkamra @mffiedler PTAL.

scaling_performance/cluster_limits.adoc
```diff
@@ -111,6 +113,12 @@ Kubernetes scheduler makes during pod placement. Learn what measures you can
 take to xref:../admin_guide/overcommit.adoc#disabling-swap-memory[avoid memory swapping].
 ====
 
+[IMPORTANT]
+====
+Some of the limits are streched only in a single dimension, so the they might
+vary when lot of objects are running on the cluster.
```
ahardin-rh: when a lot of

jeremyeder:
what was the configuration used to create 25,000 pods?

chaitanyaenr:
Cluster Setup:
OCP puddle: v3.11.16
Etcd version: 3.2.0
Operating System: RHEL-7.5
Docker version: 1.13.1
Docker storage driver: overlay2
Master/Etcd flavor: r4.4xlarge ( 16 VCPUs, 122GB memory )
Node flavor: m4.large ( 2 vCPUs, 8GB memory )
Pod image:  gcr.io/google_containers/pause-amd64:3.0
Nodes: 250

Cluster loader Config:
[](https://github.com/openshift/svt/blob/master/openshift_scalability/config/cluster-limits-pods-per-namespace.yaml)

We ran the test in iterations ( 3k, 6k, 12k, 24k, 48k pods ) and here is a link to the doc with the results:
[](https://docs.google.com/document/d/1rM1-rIC28UWct3X0dvKqPsyVavr_8jHQbj2Y_iZG8ls/edit?usp=sharing)

ahardin-rh:
@chaitanyaenr Once you apply the suggested edits and squash the changes, I can merge this. If you prefer for me to apply the changes, just let me know and I can take care of it. Thanks!

chaitanyaenr:
@ahardin-rh updated the PR to reflect the suggested changes. PTAL.

scaling_performance/cluster_limits.adoc
```diff
@@ -69,11 +69,13 @@ It does not necessarily mean that the cluster will fail.
 a number of control loops in the system that need to iterate over all objects
 in a given namespace as a reaction to some changes in state. Having a large
 number of objects of a given type in a single namespace can make those loops
-expensive and slow down processing given state changes.]
-| 3,000
+expensive and slow down processing given state changes. The limit 
+assumes that the system has enough CPU, memory and disk, to satifty the
```
ahardin-rh: CPU, memory, and disk to satisfy

ahardin-rh:
@chaitanyaenr Just one minor comment and we should be good to go :+1:

scaling_performance/cluster_limits.adoc
ashishkamra: "so the they" -> "so they"
"objects running on the cluster" - is there a better way to phrase it? objects don't really run on the cluster.

ashishkamra:
@ahardin-rh Unrelated to this commit, at the beginning of this doc we have the following sentence
"These limits are based on on the the largest possible cluster". can you remove the extra "on the" ?

ahardin-rh:
@ashishkamra Good catch. I will do some follow-up edits in a separate PR. Merging this one. Thanks!

chaitanyaenr:
Thanks for the review.

PR #14324, opened 2019-04-02
Added information about configuring faster storage for etcd

ahardin-rh:
@sjug PTAL at this in context of the existing installation/customizing install-config.yaml content in place for 4.1 docs.

### DIFF
--- a/modules/installation-aws-config-yaml.adoc
+++ b/modules/installation-aws-config-yaml.adoc
@@ -26,6 +26,10 @@ controlPlane: <2>
       zones:
       - us-west-2a
       - us-west-2b
+      rootVolume:
+        iops: 4000
+        size: 500
+        type: io1
       type: m5.xlarge
   replicas: 3
 compute: <2>
@@ -33,10 +37,10 @@ compute: <2>
   platform:
     aws:
       rootVolume:
-        iops: 4000
+        iops: 2000
         size: 500
-        type: io1
-      type: c5.9xlarge
+        type: io1 <4>
+      type: c5.4xlarge
       zones:
       - us-west-2c
   replicas: 5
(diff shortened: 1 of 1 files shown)


sjug:
/lgtm 

Subtle but valid.

ahardin-rh:
@mffiedler Can you please provide a QE review?
@kalexand-rh Since we discussed this a little last week, can you please provide a peer review?

kalexand-rh (commented):
LGTM

modules/installation-aws-config-yaml.adoc
```diff
@@ -62,4 +66,6 @@ sshKey: ssh-ed25519 AAAA... <3>
 <2> If you do not provide these parameters and values, the installation program
 provides the default value.
 <3> You can optionally provide the `sshKey` value that you use to access the
-machines in your cluster.
\ No newline at end of file
+machines in your cluster.
+<4> To configure faster storage for etcd, especially for larger clusters, set the
+storage type as `io1` and set `iops size` to least `500`.
```
mffiedler: iops and size are separate fields - iops=speed in I/O operations/second and size=how big disk is (Gb).   I suggest "To configure faster storage for etcd, especially for larger clusters, set the storage type as `io1` and set `iops` to 4000."

mffiedler (changes requested):

mffiedler:
@ahardin-rh @sjug Where did 4000 come from for iops?  In recent master-vert tests, 2000 iops would be more than sufficient.     Not opposed to using  it as an example, but a 500Gb 4000 iops disk on AWS would be pretty expensive.

sjug:
Thanks @mffiedler, that's from me, pulled it from the [installer docs]Vhttps://github.com/openshift/installer/blob/master/docs/user/aws/customization.md).

mffiedler:
Ok,  it might sound nitpicky, but (because I am cheapskate) I would recommend changing the example to something more modest.   Those 5 c5.9xlarge instances with that disk would be $5K/month.    I suggest changing the instance size to c5.4xlarge and iops to 2000, if there are no objections.

sjug:
Sounds good to me

ahardin-rh:
@sjug @mffiedler Thanks! Updated. Merging! :rocket:

PR #14351, opened 2019-04-04
CNV-127: Bridging (and PXE booting)

ousleyp:
https://jira.coreos.com/browse/CNV-127

### DIFF
--- a/modules/cnv-connecting-resource-bridge-network.adoc
+++ b/modules/cnv-connecting-resource-bridge-network.adoc
@@ -1,0 +1,89 @@
+// Module included in the following assemblies:
+//
+// * cnv/cnv_users_guide/cnv-attaching-vm-multiple-networks.adoc
+
+[id="connecting-resource-bridge-network-{context}"]
+= Connecting a resource to a bridge-based network
+
+As a network administrator, you can configure a NetworkAttachmentDefinition
+of type `cnv-bridge` to provide Layer-2 networking to Pods and virtual machines.
+
+.Prerequisites
+
+* {ProductName} 2.0 or newer
+* A Linux bridge must be configured and attached to the correct
+Network Interface Card on every node.
+* If you use VLANs, `vlan_filtering` must be enabled on the bridge.
+* The NIC must be tagged to all relevant VLANs.
+** For example: `bridge vlan add dev bond0 vid 1-4095 master`
+
+.Procedure
+
+. Create a new file for the NetworkAttachmentDefinition in any local directory.
+The file must have the following contents, modified to match your
+configuration:
++
+[source,yaml]
+----
+apiVersion: "k8s.cni.cncf.io/v1"
+kind: NetworkAttachmentDefinition
+metadata:
+  name: a-bridge-network
+  annotations:
+    k8s.v1.cni.cncf.io/resourceName: bridge-cni.network.kubevirt.io/br0 <1>
+spec:
+  config: '{
+    "cniVersion": "0.3.0",
+    "name": "a-bridge-network", <2>
+    "type": "cnv-bridge", <3>
+    "bridge": "br0", <4>
+    "isGateway": true,
+    "ipam": {}
+}'
+----
+<1> If you add this annotation to your NetworkAttachmentDefinition, your VMIs
+will only run on nodes that have the `br0` bridge connected.
+<2> The `name` value is part of the annotation you will use in the next step.
+<3> The actual name of the Container Network Interface (CNI) plug-in that provides
+the network for this NetworkAttachmentDefinition. Do not change this field unless
+you want to use a different CNI.
+<4> You must substitute the actual name of the bridge, if it is not `br0`.
+
+. Edit the configuration file of a resource that you want to connect to the
+bridge network:
++
+[source,yaml]
+----
+apiVersion: v1
+kind: Pod
+metadata:
+  name: example-pod
+  annotations:
+    k8s.v1.cni.cncf.io/networks: a-bridge-network <1>
+spec:
+...
+----
+<1> You must substitute the actual `name` value from the
+NetworkAttachmentDefinition.
++
+In this example, the NetworkAttachmentDefinition and Pod are in the same
+namespace.
++
+To specify a different namespace, use the following syntax:
++
+[source,yaml]
+----
+...
+  annotations:
+    k8s.v1.cni.cncf.io/networks: <namespace>/a-bridge-network
+...
+----
+
+. Apply the configuration file to the resource:
++
+----
+$ oc create -f <local/path/to/network-attachment-definition.yaml>
+----
+
+
+
--- a/modules/cnv-networking-glossary.adoc
+++ b/modules/cnv-networking-glossary.adoc
@@ -1,0 +1,30 @@
+// Module included in the following assemblies:
+//
+// * cnv/cnv_users_guide/cnv-attaching-vm-multiple-networks.adoc
+// * cnv/cnv_users_guide/cnv-configuring-pxe-booting.adoc
+
+[id="cnv-networking-glossary-{context}"]
+= {ProductName} networking glossary
+
+{ProductName} provides advanced networking functionality by using custom
+resources and plug-ins.
+
+The following terms are used throughout {ProductName} documentation:
+
+Container Network Interface (CNI):: a link:https://www.cncf.io/[Cloud Native Computing Foundation]
+project, focused on container network connectivity. {ProductName} uses CNI
+plug-ins to build upon the basic Kubernetes networking functionality.
+
+Multus:: a "meta" CNI plug-in that allows multiple CNIs to exist so that a Pod or
+virtual machine can use the interfaces it needs.
+
+Custom Resource Definition (CRD):: a link:https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/[Kubernetes]
+API resource that allows you to define custom resources, or an object defined by
+using the CRD API resource.
+
+NetworkAttachmentDefinition:: a CRD introduced by the Multus project that
+allows you to attach Pods, VMs, and VMIs to one or more networks.
+
+Preboot eXecution Environment (PXE):: an interface that enables an administrator
+to boot a client machine from a server over the network. Network booting allows
+you to remotely load operating systems and other software onto the client.
--- a/modules/cnv-pxe-booting-with-mac-address.adoc
+++ b/modules/cnv-pxe-booting-with-mac-address.adoc
@@ -1,0 +1,154 @@
+// Module included in the following assemblies:
+//
+// * cnv/cnv_users_guide/cnv-configuring-pxe-booting.adoc
+
+[id="cnv-pxe-booting-with-mac-address-{context}"]
+= PXE booting with a specified MAC address
+
+As an administrator, you can boot a client over the network by first creating a
+NetworkAttachmentDefinition object for your PXE network. Then, reference
+the NetworkAttachmentDefinition in your virtual machine instance configuration
+file before you start the VMI. You can also specify a MAC address in the VMI
+configuration file, if required by the PXE server.
+
+.Prerequisites
+
+* A Linux bridge must be connected.
+* The PXE server must be connected to the same VLAN as the bridge.
+
+.Procedure
+
+. Configure a PXE network on the cluster:
+
+.. Create the NetworkAttachmentDefinition file for PXE network `pxe-net-conf`:
++
+[source,yaml]
+----
+apiVersion: "k8s.cni.cncf.io/v1"
+kind: NetworkAttachmentDefinition
+metadata:
+  name: pxe-net-conf
+spec:
+  config: '{
+      "cniVersion": "0.3.1",
+      "name": "pxe-net-conf",
+      "plugins": [
+        {
+          "type": "cnv-bridge",
+          "bridge": "br1",
+          "ipam": {}
+        },
+        {
+          "type": "cnv-tuning" <1>
+        }
+      ]
+    }'
+----
+<1> The `cnv-tuning` plug-in provides support for custom MAC addresses.
++
+[NOTE]
+====
+The VMI will be attached to the bridge `br1` through an access port with the
+requested VLAN.
+====
+
+. Create the NetworkAttachmentDefinition object by using the file you created
+in the previous step:
++
+----
+$ oc create -f pxe-net-conf.yaml
+----
+
+. Edit the virtual machine instance configuration file to include the
+details of the interface and network.
+
+.. Specify the network and MAC address, if required by the PXE server.
+If the MAC address is not specified, a value is assigned
+automatically. However, note that at this time, MAC addresses
+assigned automatically are not persistent.
++
+Ensure that `bootOrder` is set to `1` so that the interface boots first.
+In this example, the interface is connected to a network called
+`<pxe-net>`:
++
+----
+interfaces:
+- masquerade: {}
+  name: default
+- bridge: {}
+  name: pxe-net
+  macAddress: de:00:00:00:00:de
+  bootOrder: 1
+----
++
+[NOTE]
+====
+Boot order is global for interfaces and disks.
+====
+
+.. Assign a boot device number to the disk to ensure proper booting
+after operating system provisioning.
++
+Set the disk `bootOrder` value to `2`:
++
+----
+devices:
+  disks:
+  - disk:
+      bus: virtio
+    name: containerdisk
+    bootOrder: 2
+----
+
+.. Specify that the network is connected to the previously created
+NetworkAttachmentDefinition. In this scenario, `<pxe-net>` is connected
+to the NetworkAttachmentDefinition called `<pxe-net-conf>`:
++
+----
+networks:
+- name: default
+  pod: {}
+- name: pxe-net
+  multus:
+    networkName: pxe-net-conf
+----
+
+. Create the virtual machine instance:
++
+----
+$ oc create -f vmi-pxe-boot.yaml
+  virtualmachineinstance.kubevirt.io "vmi-pxe-boot" created
+----
+
+. Wait for the virtual machine instance to run:
++
+----
+$ oc get vmi vmi-pxe-boot -o yaml | grep -i phase
+  phase: Running
+----
+
+. View the virtual machine instance using VNC:
++
+----
+$ virtctl vnc vmi-pxe-boot
+----
+
+. Watch the boot screen to verify that the PXE boot is successful.
+
+. Log in to the VMI:
++
+----
+$ virtctl console vmi-pxe-boot
+----
+
+. Verify the interfaces and MAC address on the VM, and that the interface
+connected to the bridge has the specified MAC address. In this
+case, we used `eth1` for the PXE boot, without an IP address. The other
+interface, `eth0`, got an IP address from {product-title}.
++
+----
+$ ip addr
+...
+3. eth1: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN group default qlen 1000
+   link/ether de:00:00:00:00:de brd ff:ff:ff:ff:ff:ff
+----
--- a/modules/cnv-template-vmi-pxe-config.adoc
+++ b/modules/cnv-template-vmi-pxe-config.adoc
@@ -1,0 +1,58 @@
+// Module included in the following assemblies:
+//
+// * cnv/cnv_users_guide/cnv-configuring-pxe-booting.adoc
+
+[id="cnv-pxe-vmi-template-{context}"]
+= Template: VMI configuration file for PXE booting
+
+*vmi-pxe-boot.yaml*
+[source,yaml]
+----
+apiVersion: kubevirt.io/v1alpha3
+kind: VirtualMachineInstance
+metadata:
+  creationTimestamp: null
+  labels:
+    special: vmi-pxe-boot
+  name: vmi-pxe-boot
+spec:
+  domain:
+    devices:
+      disks:
+      - disk:
+          bus: virtio
+        name: containerdisk
+        bootOrder: 2
+      - disk:
+          bus: virtio
+        name: cloudinitdisk
+      interfaces:
+      - masquerade: {}
+        name: default
+      - bridge: {}
+        name: pxe-net
+        macAddress: de:00:00:00:00:de
+        bootOrder: 1
+    machine:
+      type: ""
+    resources:
+      requests:
+        memory: 1024M
+  networks:
+  - name: default
+    pod: {}
+  - multus:
+      networkName: pxe-net-conf
+    name: pxe-net
+  terminationGracePeriodSeconds: 0
+  volumes:
+  - name: containerdisk
+    containerDisk:
+      image: kubevirt/fedora-cloud-container-disk-demo
+  - cloudInitNoCloud:
+      userData: |
+        #!/bin/bash
+        echo "fedora" | passwd fedora --stdin
+    name: cloudinitdisk
+status: {}
+----
--- a/modules/cnv-vm-create-nic-web.adoc
+++ b/modules/cnv-vm-create-nic-web.adoc
@@ -1,0 +1,18 @@
+// Module included in the following assemblies:
+//
+// * cnv/cnv_users_guide/cnv-attaching-vm-multiple-networks.adoc
+
+[id="cnv-vm-create-nic-web-{context}"]
+= Creating a NIC for a virtual machine
+
+Create and attach additional NICs to a virtual machine from the web console.
+
+.Procedure
+
+. Click *Workloads* -> *Virtual Machines* from the side menu.
+. Select a virtual machine template.
+. Click *Network Interfaces* to display the NICs already attached to the virtual machine.
+. Click *Create NIC* to create a new slot in the list.
+. Fill in the *NAME*, *NETWORK*, *MAC ADDRESS*, and *BINDING METHOD* for the new NIC.
+. Click the *&#10003;* button to save and attach the NIC to the virtual machine.
+
--- a/cnv/cnv_users_guide/cnv-attaching-vm-multiple-networks.adoc
+++ b/cnv/cnv_users_guide/cnv-attaching-vm-multiple-networks.adoc
@@ -1,0 +1,30 @@
+[id="attaching-to-multiple-networks"]
+= Attaching a virtual machine to multiple networks
+include::modules/cnv-document-attributes.adoc[]
+:context: cnv-attaching-multiple-networks
+toc::[]
+
+{ProductName} provides Layer-2 networking capabilities that allow you to connect
+virtual machines to multiple networks. You can import virtual machines with
+existing workloads that depend on access to multiple interfaces. You can also
+configure a PXE network so that you can boot machines over the network.
+
+To get started, a network administrator configures a NetworkAttachmentDefinition
+of type `cnv-bridge`. Then, users can attach Pods, VMIs, and VMs to the
+bridge network. From the {ProductName} web console, you can create a vNIC
+that refers to the bridge network.
+
+include::modules/cnv-networking-glossary.adoc[leveloffset=+1]
+
+include::modules/cnv-connecting-resource-bridge-network.adoc[leveloffset=+1]
+
+[NOTE]
+====
+When defining the vNIC in the next section, ensure that the *NETWORK* value is
+the bridge network name from the NetworkAttachmentDefinition you created
+in the previous section.
+====
+
+include::modules/cnv-vm-create-nic-web.adoc[leveloffset=+1]
+
+include::modules/cnv-networking-wizard-fields-web.adoc[leveloffset=+1]
--- a/cnv/cnv_users_guide/cnv-configuring-pxe-booting.adoc
+++ b/cnv/cnv_users_guide/cnv-configuring-pxe-booting.adoc
@@ -1,0 +1,22 @@
+[id="configuring-pxe-booting"]
+= Configuring PXE booting for virtual machines
+include::modules/cnv-document-attributes.adoc[]
+:context: pxe-booting
+toc::[]
+
+PXE booting, or network booting, is available in {ProductName}.
+Network booting allows a computer to boot and load an
+operating system or other program without requiring a locally attached
+storage device. For example, you can use it to choose your desired OS
+image from a PXE server when deploying a new host.
+
+.Prerequisites
+
+* A Linux bridge must be xref:../../cnv/cnv_users_guide/cnv-attaching-vm-multiple-networks.adoc#connecting-resource-bridge-network-cnv-attaching-multiple-networks[connected].
+* The PXE server must be connected to the same VLAN as the bridge.
+
+include::modules/cnv-networking-glossary.adoc[leveloffset=+1]
+
+include::modules/cnv-pxe-booting-with-mac-address.adoc[leveloffset=+1]
+
+include::modules/cnv-template-vmi-pxe-config.adoc[leveloffset=+1]
--- a/modules/cnv-networking-wizard-fields-web.adoc
+++ b/modules/cnv-networking-wizard-fields-web.adoc
@@ -1,0 +1,34 @@
+// Module included in the following assemblies:
+//
+// * cnv/cnv_users_guide/cnv-attaching-vm-multiple-networks.adoc
+
+[id="cnv-networking-wizard-fields-web-{context}"]
+= Networking fields
+|===
+|Name | Description
+
+|Create NIC
+|Create a new NIC for the virtual machine.
+
+|NIC NAME
+|Name for the NIC.
+
+|MAC ADDRESS
+|MAC address for the network interface. If a MAC address is not specified, an
+ephemeral address is generated for the session.
+
+|NETWORK CONFIGURATION
+|List of available NetworkAttachmentDefinition objects.
+
+|BINDING METHOD
+|List of available binding methods. For the default Pod network, `masquerade`
+is the only recommended binding method. For secondary networks, use the `bridge`
+binding method. The `masquerade` method is not supported for non-default
+networks.
+
+|PXE NIC
+|List of PXE-capable networks. Only visible if `PXE` has been selected as the
+`Provision Source`.
+|===
+
+


ousleyp:
@vikram-redhat Hey Vikram, I added the cnv_document_attributes.adoc file to the modules folder, and included it in these two assemblies, but I'm not seeing any variable substitution. What'd I do wrong? :)

cnv/cnv_attaching_resource_multiple_networks.adoc
```diff
@@ -0,0 +1,28 @@
+[id="attaching-to-multiple-networks"]
+= Attaching a resource to multiple networks
```
dankenigsberg: I find the term "resource" too generic and vague. Would you replace it with "virtual machine"?

cnv/cnv_attaching_resource_multiple_networks.adoc
```diff
@@ -0,0 +1,28 @@
...
+:context: cnv-attaching-multiple-networks
+toc::[]
+
+{ProductName} provides Layer-2 networking capabilities that allow you to connect
+virtual machines to multiple networks. You can import virtual machines with
+existing workloads that depend on access to multiple interfaces. You can also
+configure a PXE network so that you can boot machines over the network.
+
+To get started, a network administrator configures a NetworkAttachmentDefinition
+of type `cnv-bridge`. Then, users can attach Pods and other resources to the
```
dankenigsberg: how about going specific?
"Pods and other resources" - >  Pods, VMIs and VMs

modules/cnv_connecting_resource_bridge_network.adoc
```diff
@@ -0,0 +1,78 @@
+// Module included in the following assemblies:
+//
+// * cnv/cnv_attaching_resource_multiple_networks.adoc
+
+[id="connecting-resource-bridge-network-{context}"]
+= Connecting a resource to a bridge network
```
dankenigsberg: English: would it not be nicer to say

> bridge-based network?

modules/cnv_connecting_resource_bridge_network.adoc
```diff
@@ -0,0 +1,78 @@
...
+
+As a network administrator, you can configure a NetworkAttachmentDefinition
+of type `cnv-bridge` to provide Layer-2 networking to Pods and virtual machines.
+
+.Prerequisites
+
+* {ProductName} 2.0 or newer
+* A Linux bridge must be configured on every host.
```
dankenigsberg: host -> node

what can I do, that's the k8s term.

dankenigsberg: > every node ... and attached to the correct Network Interface Card on that node.

connecting to hardware is an important prerequisite.

modules/cnv_connecting_resource_bridge_network.adoc
```diff
@@ -0,0 +1,78 @@
...
+* `vlan_filtering` must be enabled on the bridge.
```
dankenigsberg: I'd use "should" instead of "must".

The world is not going to end if they forget it; they would simply be unable to use vlans.

modules/cnv_connecting_resource_bridge_network.adoc
```diff
@@ -0,0 +1,78 @@
...
+
+.Procedure
+
+. Create a new file for the NetworkAttachmentDefinition. This file can be stored
+locally in any directory.
++
+----
+$ touch network-attachment-definition.yaml
```
dankenigsberg: the `touch` stage seems superfluous to me. Can we just merge it into the Edit bullet point, E.g

> Create a new file for the NetworkAttachmentDefinition with your favorite. This file can be stored
locally in any directory.

dankenigsberg (changes requested):
only a partial review. @myakove @SchSeba can you continue it?

cnv/cnv_attaching_resource_multiple_networks.adoc
ousleyp: Yes! That's what I originally had, but since you can also attach pods and VMIs to multiple networks, I wasn't sure if it was accurate enough. Probably better to be too specific than too vague, though.

modules/cnv_connecting_resource_bridge_network.adoc
ousleyp: Not sure which sounds better, to be honest. I will change it to bridge-based network though, because that seems more accurate.

modules/cnv_connecting_resource_bridge_network.adoc
ousleyp: "Should" is a word we tend to avoid in docs because it is ambiguous in meaning. I'm going to say "vlan_filtering must be enabled on the bridge if you want to use VLANs." :)

dankenigsberg: I'd rather you keep the simple shorter "must".

ousleyp: Well, you brought up a good point that it isn't really a "must." So now I'm not sure. :) 

I agree that "vlan_filtering must be enabled on the bridge if you want to use VLANs" is not the most graceful way to phrase it... and it's in the passive voice (though that seems natural  to me in a prerequisites section)

More options:
"To use VLANs, enable vlan_filtering on the bridge."
"If you want to use VLANs, enable `vlan_filtering` on the bridge."
"`vlan_filtering` must be enabled on the bridge."
(and endless iterations thereof.)

paging @kalexand-rh - thoughts?

modules/cnv-connecting-resource-bridge-network.adoc
```diff
@@ -0,0 +1,74 @@
...
+[source,yaml]
+----
+apiVersion: "k8s.cni.cncf.io/v1"
+kind: NetworkAttachmentDefinition
+metadata:
+  name: a-bridge-network
+spec:
+  config: '{
+    "cniVersion": "0.3.0",
+    "name": "a-bridge-network", <1>
```
SchSeba: maybe the number before the `,` something like this `"name": "a-bridge-network" <1>,`

modules/cnv-connecting-resource-bridge-network.adoc
```diff
@@ -0,0 +1,74 @@
...
+    "isGateway": true,
+    "ipam": {
+      "type": "host-local",
+      "subnet": "192.168.5.0/24",
+      "dataDir": "/mnt/cluster-ipam"
+    }
+}'
+----
+<1> The `name` value is part of the annotation you will use in the next step.
+<2> The actual name of the Container Network Interface that provides the network
```
SchSeba: `Container Network Interface Plugin`

modules/cnv-connecting-resource-bridge-network.adoc
```diff
@@ -0,0 +1,74 @@
...
+for this NetworkAttachmentDefinition. Do not change this field unless you want
+to use a different CNI.
```
SchSeba: CNI -> Container Network Interface or please add in the previews comment `(CNI)`

modules/cnv-connecting-resource-bridge-network.adoc
```diff
@@ -0,0 +1,74 @@
...
+metadata:
+  name: a-bridge-network
+spec:
+  config: '{
+    "cniVersion": "0.3.0",
+    "name": "a-bridge-network", <1>
+    "type": "cnv-bridge", <2>
+    "bridge": "br0", <3>
+    "isGateway": true,
+    "ipam": {
```
SchSeba: Please remove the IPAM for L2 networks `"ipam": {}` also please take a look on this issue https://github.com/kubevirt/kubevirt/issues/2163

for Pod that want to have a L3 bridge they need `"ipam": { "type": "dhcp"} but for virtual machines leave the ipam empty so the vm will ask for the dhcp and not the CNI plugin.
if we don't do it the dhcp release and renew will not work.

BTW please don't document the IPAM type host-local this plugins is not for a multi node environment.
the only way for the host-local to be multi node is if the `/mnt/cluster-ipam` folder is mounted and shared by all the nodes but them we can get a rise condition

modules/cnv-connecting-resource-bridge-network.adoc
```diff
@@ -0,0 +1,74 @@
...
+bridge network:
++
+[source,yaml]
+----
+apiVersion: v1
+kind: Pod
+metadata:
+  name: example-pod
+  annotations:
+    k8s.v1.cni.cncf.io/networks: a-bridge-network <1>
```
SchSeba: please add that this way the network attachment definition and the pod needs to be on the same project(namespace).

If not the user need to specify the name `<namespace>/<network attachment definition>`

modules/cnv-pxe-booting-with-mac-address.adoc
```diff
@@ -0,0 +1,136 @@
...
+----
+apiVersion: "k8s.cni.cncf.io/v1"
+kind: NetworkAttachmentDefinition
+metadata:
+  name: pxe-net-conf
+spec:
+  config: '{
+      "cniVersion": "0.3.1",
+      "type": "cnv-bridge",
+      "bridge": "br1"
```
SchSeba: please add here after the bridge `"ipam":{}`

modules/cnv-pxe-booting-with-mac-address.adoc
```diff
@@ -0,0 +1,136 @@
...
+    }'
+----
++
+[NOTE]
+====
+In this example, the VMI will be attached through a trunk port
```
SchSeba: this is not a trunk port. the port on the host here need to be configured as access.

if you want the pod to be trunk you need to and the vlan to the network attachment definition CRD

modules/cnv-pxe-booting-with-mac-address.adoc
```diff
@@ -0,0 +1,136 @@
...
+In this example, the interface is connected to a network called
+`<pxe-net>`:
++
+----
+interfaces:
+- masquerade: {}
+  name: default
+- bridge: {}
+  name: pxe-net
+  macAddress: de:00:00:00:00:de
```
SchSeba: the pxe-net CRD use the bridge plugin that doesn't support for mac address you need to chain it with the cnv-tuning plugin like this

```yaml
{
      "cniVersion": "0.3.1",
      "name": "bridge-tuning",
      "plugins": [
        {
          "type": "cnv-bridge",
          "bridge": "br-tuning",
          "ipam": {}
        },
        {
          "type": "cnv-tuning"
        }
      ]
    }
```

SchSeba (changes requested):
looks good I just have a few comments

modules/cnv-connecting-resource-bridge-network.adoc
```diff
@@ -0,0 +1,74 @@
...
+
+As a network administrator, you can configure a NetworkAttachmentDefinition
+of type `cnv-bridge` to provide Layer-2 networking to Pods and virtual machines.
+
+.Prerequisites
+
+* {ProductName} 2.0 or newer
+* A Linux bridge must be configured and attached to the correct
+Network Interface Card on every node.
+* `vlan_filtering` must be enabled on the bridge if you want to use VLANs.
```
SchSeba: the user also need to add this line for every vlan

`
bridge vlan add dev <external interface> vid <vlan-id> master
`
example:
`
bridge vlan add dev bond0 vid 2 master
`

dankenigsberg:
I was not aware (or conveniently forgot) that we have to tag our uplink. It
is not reasonable to ask the end admin to do that. Isn't there a way to
trunk everything into the uplink?

SchSeba:
@dankenigsberg there is no way to mark the interface as trunk a work around can be to add a range something like this
`
bridge vlan add dev bond0 vid 1-4098 master
`

dankenigsberg:
I hope Hangbin can confirm that this is the best way to define a trunk port.

We should do this in Ignition, right where we create the bridge and enable
vlan filtering.

modules/cnv-connecting-resource-bridge-network.adoc
ousleyp: I think I prefer it the way it is. I think it looks a bit cleaner and is still understandable with the example that comes next. @kalexand-rh, are there guidelines about callouts that I should be following? The 4.0 contrib guide doesn't explicitly specify if they have to go at the end of the line or not.

modules/cnv-connecting-resource-bridge-network.adoc
ousleyp: Thanks for the explanation! I have very minimal networking knowledge so I appreciate it. :) I'm changing it to just `"ipam": {}` for this YAML.

modules/cnv-connecting-resource-bridge-network.adoc
ousleyp: Good to know! Adding a note about that now.

modules/cnv-pxe-booting-with-mac-address.adoc
ousleyp: Ah, that was left over from the 1.4 docs. I wasn't sure if it still applied. Do you think I should just remove that note, or should I replace it with something?

modules/cnv-connecting-resource-bridge-network.adoc
ousleyp: Can you explain this in plainer language? Should this be a step in the procedure, a prerequisite, or its own procedure? Thanks.

modules/cnv-pxe-booting-with-mac-address.adoc
ousleyp: Oh wow, thanks for the heads up. Is this documented upstream anywhere? I can't find it. 

Please let me know if the updated version is correct.

dankenigsberg:
On Mon, Apr 08, 2019 at 01:51:27PM +0300, Dan Kenigsberg wrote:
> I hope Hangbin can confirm that this is the best way to define a trunk port.

Yes, the work around Sebastian pointed is the best way.

Thanks
Hangbin

modules/cnv-connecting-resource-bridge-network.adoc
kalexand-rh: It's not explicitly called out anywhere, but after a little testing, I think it's an asciidoc limitation - the annotations can only go at the end of the line or they won't render.

modules/cnv_connecting_resource_bridge_network.adoc
kalexand-rh: Where do you use VLANs? I don't see where you specify them in the code block. Since it's a prereq, the bridge should already be enabled before you start this topic. Maybe "If you use VLANs, `vlan_filtering` must be enabled on the bridge."

+1 to not using should. From our style guide:
> :exclamation:  should v
Always prefer a more direct verb or way of saying something. Omit “should,” or use “must” or “might,” as
appropriate.

modules/cnv-pxe-booting-with-mac-address.adoc
SchSeba: Hi you can take a look in this PR https://github.com/kubevirt/user-guide/pull/220

modules/cnv-connecting-resource-bridge-network.adoc
SchSeba: this should be here with the vlan_filtering.

what this line does it's allow pod/vms connected to the bridge with a specific vlan to exit the host using the host interface connected to the bridge.

modules/cnv-pxe-booting-with-mac-address.adoc
SchSeba: I think you need to replace it with 
`the VMI will be attached through a access port with the requested vlan to the bridge <br1>`

ousleyp: @SchSeba Does this seem sufficient? "In this example, the VMI will be attached to the bridge `br1` through an access port." 

I'm not sure if we need to mention the VLAN here. Does the admin need to configure it (and the access port) themselves? VLANs are not mentioned in this file at all, so please let me know if I need to add a prerequisite or if this is something that happens "under the hood."

ousleyp:
> @dankenigsberg there is no way to mark the interface as trunk a work around can be to add a range something like this
> `bridge vlan add dev bond0 vid 1-4098 master`

@SchSeba asked me to add this to the prerequisites (in the cnv-connecting-resource-bridge-network.adoc procedure), under/alongside the `vlan_filtering` item. But because it's a command that the admin needs to run, and they're only seeing it now, that makes me think it should be step 1 of the procedure. 

@kalexand-rh: I'm not sure if I can even add a command as a prerequisite. Thoughts? It doesn't seem like something that should be its own procedure, though, because it's general sysadmin stuff and not anything specific to CNV.

Thanks :)

modules/cnv-pxe-booting-with-mac-address.adoc
ousleyp: Thanks @SchSeba ! Can you confirm that this looks good to you?

`apiVersion: "k8s.cni.cncf.io/v1"
kind: NetworkAttachmentDefinition
metadata:
  name: pxe-net-conf
spec:
  config: '{
      "cniVersion": "0.3.1",
      "name": "pxe-net-conf",
      "plugins": [
        {
          "type": "cnv-bridge",
          "bridge": "br1",
          "ipam": {}
        },
        {
          "type": "cnv-tuning" <1>
        }
      ]
    }'`

dankenigsberg:
> > @dankenigsberg there is no way to mark the interface as trunk a work around can be to add a range something like this
> > `bridge vlan add dev bond0 vid 1-4098 master`
> 
> @SchSeba asked me to add this to the prerequisites (in the cnv-connecting-resource-bridge-network.adoc procedure), under/alongside the `vlan_filtering` item. But because it's a command that the admin needs to run, and they're only seeing it now, that makes me think it should be step 1 of the procedure.

Please put it in a prerequisite, since in OCP4 admins cannot log into a host and run commands. This is something that has to be pre-prepared, most probably via Ignition.

modules/cnv-pxe-booting-with-mac-address.adoc
SchSeba: looks good :)

modules/cnv-pxe-booting-with-mac-address.adoc
SchSeba: @ousleyp I think we need to document it this why.

VM -> port (vlan mode) -> linux-bridge -> bond0 (host interface port connected in trunk mode) -> network switch

kalexand-rh:
> @kalexand-rh: I'm not sure if I can even add a command as a prerequisite. Thoughts? It doesn't seem like something that should be its own procedure, though, because it's general sysadmin stuff and not anything specific to CNV.
> Please put it in a prerequisite, since in OCP4 admins cannot log into a host and run commands. This is something that has to be pre-prepared, most probably via Ignition.

It would be better in a separate module, but I'm not sure where it should fit in the flow. If I'm interpreting @dankenigsberg  correctly, you need to use special VMs for this process. If that's right, then I might slide the prereq into the assembly, make (or update) an assembly about making the VMs for this use case, and link to the new assembly from this assembly's prereqs.

dankenigsberg:
@kalexand-rh I don't know which special VMs you refer to.

All we ask is to add a prerequisite.
* You must have a bridge
* The bridge is connected to an uplink such as bond0
* The bridge has vlan_filtering on
* The uplink has all vlans tagged.

All of this is goings to be created by the kni-installer. It can also be done by an operator, or an admin via Ignition. We should not delve into that.

kalexand-rh:
Ah! @dankenigsberg, I though that the admin had to manually complete the configuration. If it's all just true prereqs, then it's fine to list them like that.

modules/cnv-connecting-resource-bridge-network.adoc
```diff
@@ -0,0 +1,83 @@
...
+
+As a network administrator, you can configure a NetworkAttachmentDefinition
+of type `cnv-bridge` to provide Layer-2 networking to Pods and virtual machines.
+
+.Prerequisites
+
+* {ProductName} 2.0 or newer
+* A Linux bridge must be configured and attached to the correct
+Network Interface Card on every node.
+* If you use VLANs, `vlan_filtering` must be enabled on the bridge.
```
dankenigsberg: we need to add here

`vlan_filtering` must be enabled on the bridge and the NIC should be tagged to all relevant VLANs.

dankenigsberg (changes requested):

_topic_map.yml
```diff
@@ -541,3 +541,12 @@ Topics:
...
   File: application-monitoring
+---
+Name: Container-native Virtualization
+Dir: cnv
+Distros: openshift-enterprise
+Topics:
+- Name: Attaching a virtual machine to multiple networks
+  File: cnv-attaching-vm-multiple-networks
+- Name: Configuring PXE booting for virtual machines
+  File: cnv-configuring-pxe-booting
```
SchSeba: small nit here please add a newline

SchSeba (changes requested):

modules/cnv-pxe-booting-with-mac-address.adoc
ousleyp: The note now reads "The VMI will be attached to the bridge `br1` through an access port with the requested VLAN." @SchSeba does that make sense?

modules/cnv-connecting-resource-bridge-network.adoc
ousleyp: @dankenigsberg  I added "* The NIC must be tagged to all relevant VLANs." as another prerequisite in cnv-connecting-resource-bridge-network.adoc. Is that sufficient or do we need to add the actual command "bridge vlan add dev bond0 vid 1-4098 master"?

I also made a couple of other minor changes in https://github.com/openshift/openshift-docs/pull/14351/commits/53f3f7e09d33683882862d62e7d7d9fe55f7644e

One question - do I also need to add additional prerequisites in cnv-pxe-booting-with-mac-address.adoc? 

Let me know if there is anything else I'm missing. I'd like to pass this to QE soon. Thanks. :)

_topic_map.yml
ousleyp: Done in https://github.com/openshift/openshift-docs/pull/14351/commits/53f3f7e09d33683882862d62e7d7d9fe55f7644e

modules/cnv-connecting-resource-bridge-network.adoc
dankenigsberg: @ousleyp "bridge vlan add dev bond0 vid 1-4098 master" is quite an arcane command. Please add it as an "e.g."

dankenigsberg: (I cannot review cnv-pxe-booting-with-mac-address.adoc now)

ousleyp: @dankenigsberg thanks! It now reads "For example: `bridge vlan add dev bond0 vid 1-4098 master`"

> (I cannot review cnv-pxe-booting-with-mac-address.adoc now)

No rush. Let me know when you get a chance to review the rest.

dankenigsberg: omg, how come do I see my own typo only when you copy it? VIDs are limited by `4095`. please replace 4098 with 4095.

modules/cnv-pxe-booting-with-mac-address.adoc
```diff
@@ -0,0 +1,147 @@
...
+PXE booting, or network booting, is supported in {ProductName}.
+Network booting allows a computer to boot and load an
+operating system or other program without requiring a locally attached
+storage device. For example, you can use it to choose your desired OS
+image from a PXE server when deploying a new host.
+
+.Prerequisites
+
+* A bridge must be configured on all nodes. The bridge must be connected to
+an interface that allows PXE booting.
```
dankenigsberg: yes, the same vlan-related prerequisites apply here, too. (If the user would like to boot from a pxe-enabled VLAN). I think that it would be best to refer to the former prereq:

* A Linux bridge must be connected, as detailed in connecting-resource-bridge-network-{context}
* A PXE server is connected to the same VLAN of the bridge.

dankenigsberg (changes requested):

modules/cnv-pxe-booting-with-mac-address.adoc
SchSeba: Yes thanks!

modules/cnv-connecting-resource-bridge-network.adoc
ousleyp: Fixed. :)

modules/cnv-pxe-booting-with-mac-address.adoc
ousleyp: @dankenigsberg: I added something similar, including a link. Unfortunately, I can't test the link yet (the path will be different than it is on my local build) but please let me know what you think of the rest when you get the chance. thanks :)

And if this is resolved, let me know if you think we can move it to QE Review.

dankenigsberg (approved):
I'd like to see this version merged.

ousleyp:
> I'd like to see this version merged.

Great! I'm going to squash the commits now and seek peer review and QE.

cnv/cnv-attaching-vm-multiple-networks.adoc
```diff
@@ -0,0 +1,28 @@
...
+toc::[]
+
+{ProductName} provides Layer-2 networking capabilities that allow you to connect
+virtual machines to multiple networks. You can import virtual machines with
+existing workloads that depend on access to multiple interfaces. You can also
+configure a PXE network so that you can boot machines over the network.
+
+To get started, a network administrator configures a NetworkAttachmentDefinition
+of type `cnv-bridge`. Then, users can attach Pods, VMIs, and VMs to the
+bridge network. And from the {ProductName} web console, you can create a vNIC
```
ahardin-rh: network. From the

cnv/cnv-attaching-vm-multiple-networks.adoc
```diff
@@ -0,0 +1,28 @@
...
+that refers to the bridge network.
+
+include::modules/cnv-networking-glossary.adoc[leveloffset=+1]
+
+include::modules/cnv-connecting-resource-bridge-network.adoc[leveloffset=+1]
+
+[CAUTION]
+====
+In the next section, make sure to define a vNIC that refers to the bridge
```
ahardin-rh: s/make sure to/you must

modules/cnv-connecting-resource-bridge-network.adoc
```diff
@@ -0,0 +1,85 @@
...
+apiVersion: v1
+kind: Pod
+metadata:
+  name: example-pod
+  annotations:
+    k8s.v1.cni.cncf.io/networks: a-bridge-network <1>
+spec:
+...
+----
+<1> Make sure to substitute the actual `name` value from the
```
ahardin-rh: s/Make sure to/You must

modules/cnv-networking-glossary.adoc
```diff
@@ -0,0 +1,26 @@
+// Module included in the following assemblies:
+//
+// * cnv/cnv-attaching-vm-multiple-networks.adoc
+// * cnv/cnv-configuring-pxe-booting.adoc
+
+[id="cnv-networking-glossary-{context}"]
+= {ProductName} networking glossary
+
+{ProductName} provides advanced networking functionality by using custom
+resources and plugins.
```
ahardin-rh: s/plugins/plug-ins

modules/cnv-networking-glossary.adoc
```diff
@@ -0,0 +1,26 @@
...
+
+The following terms are used throughout {ProductName} documentation:
+
+Container Network Interface (CNI):: a link:https://www.cncf.io/[Cloud Native Computing Foundation]
+project, focused on container network connectivity. {ProductName} uses CNI
+plugins to build upon the basic Kubernetes networking functionality.
```
ahardin-rh: s/plugins/plug-ins

modules/cnv-networking-glossary.adoc
```diff
@@ -0,0 +1,26 @@
...
+
+Multus:: a "meta" CNI plugin that allows multiple CNIs to exist so that a Pod or
```
ahardin-rh: s/plugin/plug-in

modules/cnv-networking-glossary.adoc
```diff
@@ -0,0 +1,26 @@
...
+virtual machine can use the interfaces it needs.
+
+Custom Resource Definition (CRD):: a link:https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/[Kubernetes]
+API resource that allows you to define custom resources, or an object defined by
+using the CRD API resource.
+
+NetworkAttachmentDefinition:: a CRD introduced by the Multus project that
+allows you to attach Pods, VMs and VMIs to one or more networks.
```
ahardin-rh: VMs, and VMIs  

(Oxford comma)

modules/cnv-pxe-booting-with-mac-address.adoc
```diff
@@ -0,0 +1,149 @@
+// Module included in the following assemblies:
+//
+// * cnv/cnv-configuring-pxe-booting.adoc
+
+[id="cnv-pxe-booting-with-mac-address-{context}"]
+= PXE booting with a specified MAC address
+
+PXE booting, or network booting, is supported in {ProductName}.
```
ahardin-rh: Does PXE need defined here or somewhere in this set of topics? I'm not familiar with the acronym and had to look it up. I would consider spelling out the acronymn in the first usage here and then using PXE thereafter.

modules/cnv-pxe-booting-with-mac-address.adoc
```diff
@@ -0,0 +1,149 @@
...
+          "bridge": "br1",
+          "ipam": {}
+        },
+        {
+          "type": "cnv-tuning" <1>
+        }
+      ]
+    }'
+----
+<1> The `cnv-tuning` plugin provides support for custom MAC addresses.
```
ahardin-rh: s/plugin/plug-in

modules/cnv-vm-create-nic-web.adoc
```diff
@@ -0,0 +1,18 @@
...
+Create and attach additional NICs to a virtual machine from the web console.
+
+.Procedure
+
+. Click *Workloads* -> *Virtual Machines* from the side menu.
+. Select a virtual machine template.
+. Click *Network Interfaces* to display the NICs already attached to the virtual machine.
+. Click *Create NIC* to create a new slot in the list.
+. Fill in the *NAME*, *NETWORK*, and *MAC ADDRESS* for the new NIC.
+. Click the &#10003; button to save and attach the NIC to the virtual machine.
```
ahardin-rh: &#10003; should probably also be in bold since it's a button name

modules/cnv-pxe-booting-with-mac-address.adoc
ousleyp: I could add an entry to the glossary module. How is this, @ahardin-rh ?

Preboot eXecution Environment (PXE):: an interface that enables an administrator
to boot a client machine from a server over the network. Network booting allows
you to remotely load operating systems and other software onto the client.

cnv/cnv-attaching-vm-multiple-networks.adoc
```diff
@@ -0,0 +1,28 @@
...
+existing workloads that depend on access to multiple interfaces. You can also
+configure a PXE network so that you can boot machines over the network.
+
+To get started, a network administrator configures a NetworkAttachmentDefinition
+of type `cnv-bridge`. Then, users can attach Pods, VMIs, and VMs to the
+bridge network. From the {ProductName} web console, you can create a vNIC
+that refers to the bridge network.
+
+include::modules/cnv-networking-glossary.adoc[leveloffset=+1]
+
```
yossisegev: Dead link (404 not found)

cnv/cnv-attaching-vm-multiple-networks.adoc
```diff
@@ -0,0 +1,28 @@
...
+virtual machines to multiple networks. You can import virtual machines with
+existing workloads that depend on access to multiple interfaces. You can also
+configure a PXE network so that you can boot machines over the network.
+
+To get started, a network administrator configures a NetworkAttachmentDefinition
+of type `cnv-bridge`. Then, users can attach Pods, VMIs, and VMs to the
+bridge network. From the {ProductName} web console, you can create a vNIC
+that refers to the bridge network.
+
+include::modules/cnv-networking-glossary.adoc[leveloffset=+1]
```
yossisegev: Dead link (404 not found)

cnv/cnv-configuring-pxe-booting.adoc
```diff
@@ -0,0 +1,11 @@
+[id="configuring-pxe-booting"]
+= Configuring PXE booting for virtual machines
+include::modules/cnv_document_attributes.adoc[]
+:context: pxe-booting
+toc::[]
+
+include::modules/cnv-networking-glossary.adoc[leveloffset=+1]
+
+include::modules/cnv-pxe-booting-with-mac-address.adoc[leveloffset=+1]
```
yossisegev: I followed the actual [link](https://github.com/openshift/openshift-docs/blob/master/modules/cnv_pxebooting.adoc) .
It explains how to create a bridge using ovs-vsctl, which we are not supposed to use anymore (use nmcli or "ip link" instead).

ousleyp: > I followed the actual [link](https://github.com/openshift/openshift-docs/blob/master/modules/cnv_pxebooting.adoc) .

Hey, I'm confused about where you're seeing this. cnv_pxebooting.adoc is the old file (present in the CNV 1.4 docs) that is being replaced by cnv-pxe-booting-with-mac-address.adoc in CNV 2.0.

yossisegev: Hey Pan
I can't find cnv-pxe-booting-with-mac-address.adoc, neither under modules nor anywhere else in the repo.

cnv/cnv-attaching-vm-multiple-networks.adoc
yossisegev: No such file exists.

cnv/cnv-attaching-vm-multiple-networks.adoc
```diff
@@ -0,0 +1,28 @@
...
+
+include::modules/cnv-connecting-resource-bridge-network.adoc[leveloffset=+1]
```
yossisegev: File doesn't exist.

cnv/cnv-attaching-vm-multiple-networks.adoc
```diff
@@ -0,0 +1,28 @@
...
+
+[CAUTION]
+====
+In the next section, you must define a vNIC that refers to the bridge
+network by selecting the NetworkAttachmentDefinition object you created in
+the previous section.
+====
+
+include::modules/cnv-vm-create-nic-web.adoc[leveloffset=+1]
```
yossisegev: Actual file name is cnv_vm_create_nic_web.adoc (underscores instead of hypens).

cnv/cnv-attaching-vm-multiple-networks.adoc
```diff
@@ -0,0 +1,28 @@
...
+To get started, a network administrator configures a NetworkAttachmentDefinition
+of type `cnv-bridge`. Then, users can attach Pods, VMIs, and VMs to the
+bridge network. From the {ProductName} web console, you can create a vNIC
+that refers to the bridge network.
+
+include::modules/cnv-networking-glossary.adoc[leveloffset=+1]
+
+include::modules/cnv-connecting-resource-bridge-network.adoc[leveloffset=+1]
+
+[CAUTION]
```
yossisegev: Isn't "Caution" a bit harsh for this comment?

cnv/cnv-attaching-vm-multiple-networks.adoc
ousleyp: Hey, that was the file name previously (in 1.4) but for 2.0, we are switching to hyphens instead of underscores where possible. This is what we should have done all along, but we missed that the first time. The file in this PR has hyphens, so it is the correct file name.

cnv/cnv-configuring-pxe-booting.adoc
ousleyp: Hey, it sounds like you are searching within the enterprise-4.1 branch (or the master branch?) but I'm not sure why. All of the files I reference in the assemblies are here in this PR. Does that help?

cnv/cnv-attaching-vm-multiple-networks.adoc
dankenigsberg: I cannot stand github, but you can find the file in the "File changed" tab. The link to it https://github.com/openshift/openshift-docs/pull/14351/files#diff-df6e4689984821cbe8be31fb4871b266 would expire if Pan squashes the commit in the future.

cnv/cnv-attaching-vm-multiple-networks.adoc
ousleyp: Yeah, I think it would be fine as a [NOTE]. I'll change it.

ousleyp: Changed here: https://github.com/openshift/openshift-docs/pull/14351/commits/76f810354b4a29b3177ff0173d55e3894bdfaa6a

modules/cnv-pxe-booting-with-mac-address.adoc
ahardin-rh: @ousleyp Looks great to me. Thanks!

cnv/cnv-configuring-pxe-booting.adoc
yossisegev: Sorry, my bad. I didn't realize that the referenced files are also part of this PR, so I was looking for them in the branch master branch.

modules/cnv-connecting-resource-bridge-network.adoc
```diff
@@ -0,0 +1,85 @@
...
+  config: '{
+    "cniVersion": "0.3.0",
+    "name": "a-bridge-network", <1>
+    "type": "cnv-bridge", <2>
+    "bridge": "br0", <3>
+    "isGateway": true,
+    "ipam": {}
+}'
+----
+<1> The `name` value is part of the annotation you will use in the next step.
```
yossisegev: There is inconsistency between the callout pointer (hope I'm using a correct and comprehensible terminology) and the reference callout itself. For example:
This is how the callout pointer looks:
`"name": "a-bridge-network", (1)`
and this is how the referenced callout looks:
`i. The name value is part of the annotation you will use in the next step.`
I would expect the index in the callout to be also "(1)", and not "i".
I'm not an expert in tech-writing and documentation, but I think it doesn't look good.

modules/cnv-connecting-resource-bridge-network.adoc
```diff
@@ -0,0 +1,85 @@
...
+you want to use a different CNI.
+<3> You must substitute the actual name of the bridge, if it is not `br0`.
+
+. Edit the configuration file of a resource that you want to connect to the
+bridge network:
++
+[source,yaml]
+----
+apiVersion: v1
+kind: Pod
```
yossisegev: Don't we prefer to use a _VirtualMachine_ yaml for our examples, instead of _Pod_? This is a CNV document.

modules/cnv-vm-create-nic-web.adoc
```diff
@@ -0,0 +1,18 @@
+// Module included in the following assemblies:
```
yossisegev: Didn't test the contents of this file, due to issue in accessing web UI.
I  configured the NIC via CLI ("oc exec... ip link add...").

modules/cnv-connecting-resource-bridge-network.adoc
ousleyp: Ah, are you rendering this in Github? It shouldn't look like that. 

I just uploaded a local build which is a good approximation of what it will look like when published: http://file.bos.redhat.com/pousley/042919/cnv-127/cnv/cnv-attaching-vm-multiple-networks.html

modules/cnv-vm-create-nic-web.adoc
```diff
@@ -0,0 +1,18 @@
...
+
+Create and attach additional NICs to a virtual machine from the web console.
+
+.Procedure
+
+. Click *Workloads* -> *Virtual Machines* from the side menu.
+. Select a virtual machine template.
+. Click *Network Interfaces* to display the NICs already attached to the virtual machine.
+. Click *Create NIC* to create a new slot in the list.
+. Fill in the *NAME*, *NETWORK*, and *MAC ADDRESS* for the new NIC.
```
yossisegev: "NETWORK" must be in the name of the one defined in the NetworkAttachmentDefinition (which is described in modules/cnv-connecting-resource-bridge-network.adoc).

modules/cnv-connecting-resource-bridge-network.adoc
yossisegev: OK, it does look great on your upload.
You can resolve this conversation (I can't find how to resolve conversations myself).

modules/cnv-pxe-booting-with-mac-address.adoc
```diff
@@ -0,0 +1,149 @@
...
+----
+devices:
+  disks:
+  - disk:
+      bus: virtio
+    name: containerdisk
+    bootOrder: 2
+----
+
+.. Specify that the network is connected to the previously created
```
yossisegev: The part of creating the NetworkAttachmentDefinition is missing. The file was composed in line 25, but the NetworkAttachmentDefinition, which derives from this file, was never created.
oc create -f pxe-net-conf.yaml

modules/cnv-pxe-booting-with-mac-address.adoc
```diff
@@ -0,0 +1,149 @@
...
+
+[id="cnv-pxe-booting-with-mac-address-{context}"]
+= PXE booting with a specified MAC address
+
+PXE booting, or network booting, is supported in {ProductName}.
+Network booting allows a computer to boot and load an
+operating system or other program without requiring a locally attached
+storage device. For example, you can use it to choose your desired OS
+image from a PXE server when deploying a new host.
+
```
yossisegev: I'm not sure, but it might be worth adding a comment, saying that this doc only instructs how to set up the PXE client side, not the server side, which is not (necessarily) related to CNV.

modules/cnv-connecting-resource-bridge-network.adoc
ousleyp: Good question. We do usually focus on VMs and VMIs in our examples, but I think it's good to show a variety of options. @dankenigsberg what do you think?

modules/cnv-vm-create-nic-web.adoc
ousleyp: Following modular documentation principles, the module itself should not contain references to the cnv-connecting-resource-bridge-network file. A module must stand on its own. :) 

In the assembly (cnv-attaching-vm-multiple-networks.adoc), there is already the following note before the cnv-vm-create-nic-web.adoc module:

[NOTE]
====
In the next section, you must define a vNIC that refers to the bridge
network by selecting the NetworkAttachmentDefinition object you created in
the previous section.
====

Does that seem sufficient to you?

modules/cnv-pxe-booting-with-mac-address.adoc
ousleyp: Aha! Excellent catch. I can add `oc create -f pxe-net-conf.yaml` after line 55, if that sounds good to you.

modules/cnv-vm-create-nic-web.adoc
yossisegev: I don't think this note is sufficient. I think it should be explicitly noted here, in order to be fully clear.
You don't have to reference the cnv-connecting-resource-bridge-network file, but rather add a note very similar to the one you presented here. Something like:
```
[NOTE]
====
NETWORK must refer to the bridge network name from the
NetworkAttachmentDefinition object you created in
the previous section.
====
```

modules/cnv-pxe-booting-with-mac-address.adoc
yossisegev: Sounds like the perfect place for this addition.

modules/cnv-vm-create-nic-web.adoc
ousleyp: Unfortunately, I can't even say "the previous section" because this module could potentially be used anywhere, and it has to make sense no matter where it is used. 

If you think the current phrasing is unclear, I can change the existing note so that it specifically mentions *NETWORK* like the NIC module does. Do you think that would help?

This is the same link I provided earlier, but I'm pasting it here so you can see where the note goes (above "Creating a NIC for a virtual machine"): http://file.bos.redhat.com/pousley/042919/cnv-127/cnv/cnv-attaching-vm-multiple-networks.html

modules/cnv-template-vmi-pxe-config.adoc
```diff
@@ -0,0 +1,58 @@
+// Module included in the following assemblies:
```
yossisegev: Something is invalid with this template, but I still haven't figured out what.
When deploying it, after creating the NetworkAttachmentDefinition from cnv-pxe-booting-with-mac-address.adoc, the VMI creation fails with this error:
`  Warning  SyncFailed        0s (x5 over 0s)  virt-handler, working-pvsnt-worker-0-j84d6  server error. command SyncVMI failed: "preparing the pod network failed: Link not found"`
I tried another template, which uses the same NetworkAttachmentDefinition, and it works fine.
The one I used is [here](http://pastebin.test.redhat.com/759323), and I created it by running the command:
`oc process -f vm-template-fedora-multus-mac.yaml -p NAME=cnv-127-vm -p MULTUS_NETWORK=pxe-net-conf | oc create -f -`

modules/cnv-vm-create-nic-web.adoc
yossisegev: Yes, good idea. I agree that if you specify NETWORK in this note, that would clarify what the user should assign for NETWORK.

ousleyp: Great :) I changed it in https://github.com/openshift/openshift-docs/pull/14351/commits/590c5e7c9d207998764b3ad260d8d3a2848da5ce - let me know if the new note works for you. Thanks!

modules/cnv-pxe-booting-with-mac-address.adoc
ousleyp: Added it in https://github.com/openshift/openshift-docs/pull/14351/commits/590c5e7c9d207998764b3ad260d8d3a2848da5ce

modules/cnv-vm-create-nic-web.adoc
yossisegev: Great! Thanks for the modification.

modules/cnv-pxe-booting-with-mac-address.adoc
yossisegev: Right in place, thanks!

cnv/cnv-attaching-vm-multiple-networks.adoc
```diff
@@ -20,9 +20,9 @@ include::modules/cnv-connecting-resource-bridge-network.adoc[leveloffset=+1]
 
 [NOTE]
 ====
-In the next section, you must define a vNIC that refers to the bridge
-network by selecting the NetworkAttachmentDefinition object you created in
-the previous section.
+When defining the vNIC in the next section, ensure that the *NETWORK* value is
```
ahardin-rh: s/*NETWORK*/`NETWORK`

cnv/cnv-configuring-pxe-booting.adoc
```diff
@@ -4,6 +4,19 @@ include::modules/cnv_document_attributes.adoc[]
...
+Network booting allows a computer to boot and load an
+operating system or other program without requiring a locally attached
+storage device. For example, you can use it to choose your desired OS
+image from a PXE server when deploying a new host.
+
+.Prerequisites
+
+//note: the xref below needs to be changed after the CNV buckets are set up.
+
+* A Linux bridge must be connected, as described xref:../../cnv/cnv-attaching-vm-multiple-networks.adoc#connecting-resource-bridge-network-cnv-attaching-multiple-networks[here].
```
ahardin-rh: We avoid "here" and "click here" statements. I would just make the word "connected" the link. The user will see it's a link and intuit that they can click there for more information.

cnv/cnv-attaching-vm-multiple-networks.adoc
ousleyp: @ahardin-rh Thanks! Does the syntax in cnv-vm-create-nic-web.adoc need to be fixed also?

modules/cnv-template-vmi-pxe-config.adoc
ousleyp: Hey, thanks for the heads up. I'm not sure what changes to make here, and I'm not familiar with using that command to create a template. I know that we need the template to use `masquerade` so at the very least, that would need to be changed from the one you provided...

@dankenigsberg can you weigh in on this?

cnv/cnv-attaching-vm-multiple-networks.adoc
ahardin-rh: @ousleyp Hmm it looks like that procedure pertains to the web console and I interpreted *NAME*, *NETWORK*, and *MAC ADDRESS* as fields in the UI. If that's the case, the way it is there is correct. If NETWORK is a variable set in a file and not used in the context of the UI, then we would place it in backticks, per 
https://github.com/ahardin-rh/openshift-docs/blob/enterprise-4.1/contributing_to_docs/doc_guidelines.adoc#quick-markup-reference

ousleyp: @ahardin-rh Ok, cool - NETWORK here refers to the UI field that will come up in the cnv-vm-create-nic-web.adoc module. I think that means I can keep it in *s in the assembly as well. Thanks for your help!!

modules/cnv-connecting-resource-bridge-network.adoc
dankenigsberg: I think that VM would be better, but I would not loose sleep over it. I can live with Pod in this single occurance.

modules/cnv-connecting-resource-bridge-network.adoc
```diff
@@ -0,0 +1,85 @@
...
+metadata:
+  name: example-pod
+  annotations:
+    k8s.v1.cni.cncf.io/networks: a-bridge-network <1>
```
mmirecki: The pod/vmi can also be scheduled on a node where the bridge used in a NetworkAttachmentDefinition is available.

To mark the nodes with available bridges the bridge marker daemon (https://github.com/kubevirt/bridge-marker/) must be running. The daemon can be installed using the Network Addons Operator (https://github.com/kubevirt/cluster-network-addons-operator).
In order for a vmi to be scheduled on available nodes, the NetworkAttachmentDefinition must have an annotation marking that it requires the given bridge: 
  metadata.annotations:     k8s.v1.cni.cncf.io/resourceName: bridge-cni.network.kubevirt.io/<bridge name>
For example, to mark vmi's with a NetworkAttachmentDefinition to to be scheduled only on nodes where a bridge named "br1" is present, the network attachment defintion would look as follows:
[source,yaml]
----
apiVersion: "k8s.cni.cncf.io/v1"
kind: NetworkAttachmentDefinition
metadata:
  name: a-bridge-network
  annotations:
    k8s.v1.cni.cncf.io/resourceName: bridge-cni.network.kubevirt.io/br1
spec:
  config: '{
    "cniVersion": "0.3.0",
    "name": "a-bridge-network", 
    "type": "cnv-bridge",
    "bridge": "br0", 
    "isGateway": true,
    "ipam": {}
}'
----

To mark a pod as requiring the bridge, the pod would need to have a resouce specified, for example:

[source,yaml]
----
apiVersion: v1
kind: Pod
metadata:
  name: a-name
  namespace: default
spec:
  containers:
  - name: a-name
    image: docker.io/...
    resources:
      limits:
        bridge.network.kubevirt.io/zbr: "1"
      requests:
        bridge.network.kubevirt.io/zbr: "1"
...
----------------------------------

modules/cnv-connecting-resource-bridge-network.adoc
```diff
@@ -0,0 +1,85 @@
...
+. Create a new file for the NetworkAttachmentDefinition in any local directory.
+The file must have the following contents, modified to match your
+configuration:
++
+[source,yaml]
+----
+apiVersion: "k8s.cni.cncf.io/v1"
+kind: NetworkAttachmentDefinition
+metadata:
+  name: a-bridge-network
```
dankenigsberg: @ousleyp can you add here
```
  annotations:
    k8s.v1.cni.cncf.io/resourceName: bridge-cni.network.kubevirt.io/br0 <1>
```
and a note

<1> if this annotation is added, VMs would be scheduled only on hosts that have this bridge available.

We can mention that this is implemented by the bridge-marker, but I'd rather we do not include @mmirecki's text about its installation as it should be working out of the box.

@ousleyp I don't manage to directly reply to your question https://github.com/openshift/openshift-docs/pull/14351#discussion_r281359810 (I never got the hang of github :-/ )
We can live without this text, but it would be great to have it included.

modules/cnv-connecting-resource-bridge-network.adoc
ousleyp: Hey Marcin, thanks for the review and information! Can you clarify if the bridge marker daemon (and network addons operator) are "supported" in downstream CNV 2.0? I hadn't heard about those until now. 

@dankenigsberg - is adding info about this required for CNV-127?

(FYI, I can't directly tag you, Marcin - I think it's because you aren't in the OpenShift org)

dankenigsberg: oh now I can comment here. @ousleyp you can learn more about the bridge marker in https://jira.coreos.com/browse/CNV-701 but I don't think it is required. all that we need to say: "if you wanna use it, add the annotation, and your VMs would run only on hosts that have this bridge connected."

ousleyp: @dankenigsberg thanks for clarifying! How's this? https://github.com/openshift/openshift-docs/pull/14351/commits/fa6734e7bd7ba0f4159bb55e0eb26d62ef91e3e9

modules/cnv-connecting-resource-bridge-network.adoc
```diff
@@ -0,0 +1,89 @@
...
+  annotations:
+    k8s.v1.cni.cncf.io/resourceName: bridge-cni.network.kubevirt.io/br0 <1>
```
mmirecki: LGTM

modules/cnv-template-vmi-pxe-config.adoc
dankenigsberg: Sorry for the lack of response here. We were trying to understand the bug.

Per Sebastian, it is a bug in kubemacpool.

Please include the template as it is. If we have a bug there, we'd fix it.

Let's have this pr merged promptly.

PR #14357, opened 2019-04-05
Removing Descheduler and Node Problem Detector from 4.1

mburke5678:
Removing Descheduler and Node Problem Detector from 4.1 per Tushar email from 4/5/19


mburke5678:
@sjenning @vikram-redhat 
I removed all the NPD and Descheduler topics. Commented out the assemblies in the `_topic_map.yaml`, which removes the links in the left-pane TOC.

PR #14366, opened 2019-04-08
trying to fix portal format

kalexand-rh:


kalexand-rh:
@ahardin-rh, will you PTAL?

ahardin-rh:
@kalexand-rh LGTM! Thank you for the fix!

kalexand-rh:
Thank you @ahardin-rh!

(@vikram-redhat, this really is just for 3.11 because the Power stuff isn't in master.)

PR #14370, opened 2019-04-08
EO - ClusterLogging Status messages when cluster logging is not functional

mburke5678:
https://jira.coreos.com/browse/LOG-339

### DIFF
--- a/logging/config/efk-logging-elasticsearch.adoc
+++ b/logging/config/efk-logging-elasticsearch.adoc
@@ -19,7 +19,7 @@ or higher memory. Each Elasticsearch node can operate with a lower memory settin
 
 [NOTE]
 ====
-Procedures in this topic require your cluster to be in an unmanaged state. For more information, see xref:../logging/efk-logging-management.adoc#efk-logging-management[Changing the cluster logging management state].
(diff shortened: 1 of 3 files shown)


mburke5678:
@jcantrill @anpingli I would appreciate any thoughts on how best to present this information. What do the users need to know about the various status information they can get?

jcantrill:
@ewolinetz can you provide the high level list of status items.  Maybe we should not identify any of them specifically and just tell user's how to see the operator status?

ewolinetz:
We will expose any events from the scheduler specific to a pod such as:
 * Container Waiting (for both the ES and Proxy containers)
 * Container Terminated (for both the ES and Proxy containers)
 * Pod unschedulable

We will also display conditions regarding the ES node storage. If the threshold setting is enabled (which is by default), we will update the conditions when the low and high watermark has been met.

All of this will be available in the `Status.Nodes` section of the `Elasticsearch` object.

mburke5678:
@QiaolingTang Please take a look. I am going to merge for beta 3 and can follow changes in a follow up.

mburke5678:
Follow-up to move information into their own files.
https://github.com/openshift/openshift-docs/pull/14370

mburke5678:
Reviewed via https://github.com/openshift/openshift-docs/pull/14456

PR #14375, opened 2019-04-09
bug 1695134 amended upgrade steps

sheriff-rh:
Bug:
https://bugzilla.redhat.com/show_bug.cgi?id=1695134

@openshift/team-documentation PTAL - I suspect someone can craft better phrasing than I can here. 

@jiajliu  please take a look as well.


kalexand-rh:
@ahardin-rh, will you PTAL?

upgrading/automated_upgrades.adoc
```diff
@@ -580,6 +580,15 @@ to ensure you use the latest upgrade playbooks.
...
 
+. Add the `openshift_image_tag` variable to your inventory file. Set its value
+to the full release version, for example `3.11.72`, which you can obtain from
+the errata.
++
+[IMPORTANT]
+====
+You must define the `openshift_image_tag` variable before you upgrade {product-title}.
+====
+
```
jianlinliu: From reference bugs:
https://bugzilla.redhat.com/show_bug.cgi?id=1684142
https://bugzilla.redhat.com/show_bug.cgi?id=1684315

I think it is better to make openshift_image_tag mandatory *only* when run some minor version upgrade, such as: from v3.10.m to v3.10.n, or from v3.11.m to v3.11.n (n > m). If  user run 3.10 to 3.11 upgrade, I do not think need this mandatory setting. So it is better to add such pre-condition scenario description for this warning.

upgrading/automated_upgrades.adoc
```diff
@@ -580,6 +580,15 @@ to ensure you use the latest upgrade playbooks.
 . Ensure the `openshift_deployment_type` parameter in your inventory file is set
 to `openshift-enterprise`.
 
+. Add the `openshift_image_tag` variable to your inventory file. Set its value
+to the full release version, for example `3.11.72`, which you can obtain from
```
jianlinliu: I think the tag version should have 'v' prefix, such as: 'v3.11.72'

wmengRH: 'v3.11.x' is correct. 
Thanks.
take real released image tags might be better, such as v3.11.88, v3.10.127

upgrading/automated_upgrades.adoc
palonsoro: This is not mandatory setting for a 3.10-->3.11 upgrade, in what refers that an actual upgrade from 3.10 to 3.11 happens.

However, if an upgrade is attempted in phases, there is some chance that some nodes get components in a 3.11 minor and others get them in a different minor (if a errata is released between one phase and the next one). Not sure if this minor detail deserves to be explicitly warned.

upgrading/automated_upgrades.adoc
sheriff-rh: I've changed this to v3.11.98 (which released today).

upgrading/automated_upgrades.adoc
jianlinliu: Personally I am not very opposed to such change, but if avoid such situation, as best practice, it is better to set openshift_pkg_version together with openshift_image_tag just like https://bugzilla.redhat.com/show_bug.cgi?id=1575931#c3.

palonsoro: Yes, in order to have 100% consistency, both variables should have explicit values pointing to the same minor

jianlinliu: And one more suggestion, here "must" seem like a bit aggressive, after all, not every situation need set "openshift_image_tag" and "openshift_pkg_version", sound "better" is good. If could explain why should set both as a best practice, that would be perfect.

palonsoro: Well, not every situation requires those variables but there is actually a situation where you must set `openshift_image_tag`: A minor upgrade. Maybe the word "must" can be replaced with something softer, but it is important that users understand `openshift_image_tag` has to be specified, at least, for minor upgrades.

sheriff-rh: We have a section on this page that says the `openshift_image_tag` and `openshift_pkg_version` need to be updated at the same time: https://docs.openshift.com/container-platform/3.11/upgrading/automated_upgrades.html#special-considerations-for-mixed-environments 

Do we need to just make this more prominent instead?

jianlinliu: The section in the link is for 'Special considerations for mixed environments', according to this new PR, seem like setting both 'openshift_image_tag' and 'openshift_pkg_version' is some common operation, so here is my proposal:
1. Remove these newly added lines in this PR.
2. Polish 'Special considerations for mixed environments' section (including the section title) to reflect setting both 'openshift_image_tag' and 'openshift_pkg_version' as a best practice.

sheriff-rh: This sounds fair to me. @palonsoro  do you agree?

palonsoro: Sorry for the late response.

What I don't see clear from this is that it is not only a best practice: it is mandatory in some cases. If we are going to proceed this way, it should be made clear that it is mandatory for minor upgrades or, at least, warn explicitly that some components **won't** be updated properly if the guideline is not followed.

sheriff-rh: @jianlinliu  any further suggestions? I feel like we are at an impasse here. Any suggestions to resolve this?

jianlinliu: I agree @palonsoro 's statement, and not conflicts with mine. I updated my proposal a bit:
1. Remove these newly added lines in this PR.
2. Polish 'Special considerations for mixed environments' section (including the section title) to reflect setting both 'openshift_image_tag' and 'openshift_pkg_version' as a best practice for major upgrades. And setting both become mandatory for minor upgrades.

PR #13272, opened 2019-01-09
Manila volumes provisioning update for 4.0

tsmetana:
There is a new Manila provisioner to be released in OCP-4.0: Here's my attempt to update the documentation accordingly.

PR #14376, opened 2019-04-09
OSDOCS-395: Include the Manila PV.

huffmanca:
Included the 3.x OpenStack Manila instructions in 4.x. This also includes the changes highlighted in [PR 13272](https://github.com/openshift/openshift-docs/pull/13272/).

@tsmetana,

I've reworked the content to use our current processes and standards, but left the technical bits unchanged. I did have a couple of questions regarding them, though:

1. The current instructions never have the user create or apply any of the files in the procedure. Should we wrap these in an `oc create -f << EOF` block? If so, is there a specific namespace we need to specify?
2. The apiVersion in the Deployment block (near the end of the assembly) is `extensions/v1beta1`. Should this be changed to be a non-beta version?
3. When creating the Secret (step 4) we ask the user to supply several base64 encoded values. Would it make sense to use the `oc create secret` command so that the base64 encoding is done for the user? Something like:

`oc create secret generic manila-secret -n default --from-literal=os-authURL=$VALUE1 --from-literal=os-userName=$VALUE2 ...`

### DIFF
--- a/modules/persistent-storage-manila-install.adoc
+++ b/modules/persistent-storage-manila-install.adoc
@@ -1,0 +1,150 @@
+// Module included in the following assemblies:
+//
+// * storage/persistent-storage/persistent-storage-manila.adoc
+
+[id="persistent-storage-manila-install-{context}"]
+= Installing the external provisioner
+
+To use OpenStack Manila persistent storage you must install
+and configure an external provisioner in the {product-title}
+cluster.
+
+The external provisioner is distributed as a container image
+and can be run in the {product-title} cluster as usual.
+
+.Procedure
+
+. Create a service account:
++
+[source,yaml]
+----
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: manila-provisioner-runner
+----
+
+. Create a ClusterRole:
++ 
+[source,yaml]
+----
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: manila-provisioner-role
+rules:
+  - apiGroups: [""]
+    resources: ["persistentvolumes", "endpoints"]
+    verbs: ["get", "list", "watch", "create", "delete", "update"]
+  - apiGroups: [""]
+    resources: ["persistentvolumeclaims"]
+    verbs: ["get", "list", "watch", "update"]
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["storageclasses"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["events"]
+    verbs: ["list", "watch", "create", "update", "patch"]
+  - apiGroups: ["v1"]
+    resources: ["secrets"]
+    verbs: ["get", "list"]
+----
+
+. Bind the rules via ClusterRoleBinding:
++ 
+[source,yaml]
+----
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: manila-provisioner
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: manila-provisioner-role
+subjects:
+- kind: ServiceAccount
+  name: manila-provisioner-runner
+  namespace: default
+----
+
+. Create a new secret:
++ 
+[source,yaml]
+----
+apiVersion: v1
+kind: Secret
+metadata:
+  name: manila-secret <1>
+  namespace: default <2>
+data:
+  os-authURL: <base64 encoded OpenStack Keystone URL>
+  os-userName: <base64 encoded Manila username>
+  os-password: <base64 encoded password>
+  os-projectName: <base64 encoded OpenStack project (tenant) name>
+  os-domainName: <base64 encoded OpenStack Manila service domain>
+  os-region: <base64 encoded OpenStack region>
+----
+<1> The secret name will be referenced by the Manila volume's
+StorageClass.
+<2> The secret namespace will be referenced by the Manila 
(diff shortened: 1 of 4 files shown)


tsmetana:
Thank you. This looks good.

>     1. The current instructions never have the user create or apply any of the files in the procedure. Should we wrap these in an `oc create -f << EOF` block? If so, is there a specific namespace we need to specify?

I'd leave this up to the user: there is a "Understanding persistent storage" section that explains the object types and what they mean... Also this started as a rewrite of existing doc so we should aim for consistency and I randomly checked some other provisioning piesces (iSCSI, block volume) and they seem to follow this "example only" convention too.

>     2. The apiVersion in the Deployment block (near the end of the assembly) is `extensions/v1beta1`. Should this be changed to be a non-beta version?

Yes. Deployment kind lives in several API versions, but `apps/v1` is the stable one. Good catch.

>     3. When creating the Secret (step 4) we ask the user to supply several base64 encoded values. Would it make sense to use the `oc create secret` command so that the base64 encoding is done for the user? Something like:
> 
> 
> `oc create secret generic manila-secret -n default --from-literal=os-authURL=$VALUE1 --from-literal=os-userName=$VALUE2 ...`

I don't do this, but that doesn't mean I'm right: if this is how the secrets are being created in other parts of the documentation, then yes. Again: I'd aim for consistency.

modules/persistent-storage-manila-install.adoc
```diff
@@ -0,0 +1,150 @@
...
+is optional, and defaults to `default`.
+<7> Default NFS client for the share exported. This field is optional,
+and is only used for the `NFS` protocol. Defaults to `0.0.0.0`. 
+
+. Start the provisioner itself. The following example uses a Deployment:
+
+[source, yaml]
+----
+kind: Deployment
+apiVersion: extensions/v1beta1
```
tsmetana: As you mentioned this should be `apps/v1`.

tsmetana (changes requested):

modules/persistent-storage-manila-install.adoc
huffmanca: Corrected.

modules/persistent-storage-manila-install.adoc
```diff
@@ -0,0 +1,150 @@
...
+This field is required.
+<5> Name of the secret object containing OpenStack credentials.
+This field is required.
+<6> Namespace of the OpenStack credentials secret object. This field
+is optional, and defaults to `default`.
+<7> Default NFS client for the share exported. This field is optional,
+and is only used for the `NFS` protocol. Defaults to `0.0.0.0`. 
+
+. Start the provisioner itself. The following example uses a Deployment:
+
```
bergerhoffer: Need a + here so that it's indented properly with step 6

modules/persistent-storage-manila-usage.adoc
```diff
@@ -0,0 +1,34 @@
...
+delete and unexport the OpenStack Manila share.
+
+.Prerequisites
+
+* The OpenStack Manila external provisioner must be installed.
+
+.Procedure
+
+* Create a PersistentVolumeClaim using the corresponding
+StorageClass
```
bergerhoffer: Need a period.

modules/persistent-storage-manila-usage.adoc
```diff
@@ -0,0 +1,34 @@
...
+
+[source,yaml]
```
bergerhoffer: Same thing about what I mentioned to you on IRC, eventually you might want to look at actually adding in the CLI commands instead of just the YAML.

modules/persistent-storage-manila-usage.adoc
```diff
@@ -0,0 +1,34 @@
...
+
+.Prerequisites
+
+* The OpenStack Manila external provisioner must be installed.
+
+.Procedure
+
+* Create a PersistentVolumeClaim using the corresponding
+StorageClass
+
```
bergerhoffer: Add a + here so that it's indented properly with "step 1"

storage/persistent-storage/persistent-storage-manila.adoc
```diff
@@ -0,0 +1,28 @@
...
+:context: persistent-storage-manila
+
+toc::[]
+
+{product-title} is capable of provisioning PVs using the
+link:https://wiki.openstack.org/wiki/Manila[OpenStack Manila] shared 
+file system service.
+
+It is assumed the OpenStack Manila service has been correctly set up and is
+accessible from the {product-title} cluster. Only the NFS share types can be
```
bergerhoffer: Is "share types" plural correct here, or should it be type singular?

modules/persistent-storage-manila-install.adoc
huffmanca: Corrected.

modules/persistent-storage-manila-usage.adoc
huffmanca: Corrected.

storage/persistent-storage/persistent-storage-manila.adoc
huffmanca: Corrected this to be singular. Good catch!

modules/persistent-storage-manila-usage.adoc
huffmanca: Corrected.

modules/persistent-storage-manila-usage.adoc
huffmanca: Agreed. We'll need to make certain that these are consistent, and I think we should change these to apply the created YAML files. We'll need to go through all of them simultaneously, though.

PR #14377, opened 2019-04-09
Bug BZ1442565 -- Do not use NFS storage in advanced install

jhoyt-rh:
Edits are in this file:
https://docs.openshift.com/container-platform/3.5/install_config/install/advanced_install.html#advanced-installation-logging-storage

Bug is here:
https://bugzilla.redhat.com/show_bug.cgi?id=1442565

### DIFF
--- a/install_config/install/advanced_install.adoc
+++ b/install_config/install/advanced_install.adoc
@@ -1311,11 +1311,19 @@ not set, then cluster logging data is stored in an `emptyDir` volume, which will
 be deleted when the Elasticsearch pod terminates.
 
 There are three options for enabling cluster logging storage when using the
-advanced install:
+advanced installation; however, for production environments, you should use only block storage.
+Do not attempt to use NFS storage. NFS storage is described below only to indicate how it might be
+used if you were not using block storage, or you were not in a production environment.
+
 
 [discrete]
 [[advanced-installation-logging-storage-nfs-host-group]]
 ===== Option A: NFS Host Group
+
+[NOTE]
+====
+Do not attempt to use NFS storage in a production environment.
+====
 
 When the following variables are set, an NFS volume is created during an
 advanced install with path *_<nfs_directory>/<volume_name>_* on the host within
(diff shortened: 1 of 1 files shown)


jhoyt-rh:
Asking someone in core OpenShift ccs for peer review and to add all labels, etc. to this PR. Assignee is not part of OpenShift core group, but of OpenShift-Aligned Docs

install_config/install/advanced_install.adoc
```diff
@@ -1601,7 +1614,7 @@ Infrastructure].
 ====
 
 Note the following when using the `native` HA method:
-
+Note
```
bergerhoffer: Was this purposeful? It looks like it might be an accidental add

install_config/install/advanced_install.adoc
```diff
@@ -1311,12 +1311,20 @@ not set, then cluster logging data is stored in an `emptyDir` volume, which will
...
+used if you were not using block storage, or you were not in a production environment.
+
 
 [discrete]
 [[advanced-installation-logging-storage-nfs-host-group]]
 ===== Option A: NFS Host Group
 
+[NOTE]
+====
+We recommend that you do not attempt to use NFS storage in a production environment.
```
bergerhoffer: I know that "we recommend" vs. "it is recommended" vs. "Red Hat recommends" vs. avoiding recommend at all is a big discussion topic and a little hazy guidance, but I definitely think we should avoid "we recommend", since we really never use "we" in the docs.

I'd suggest to either say "It is recommended", since that is already used a few places in this file, or just say "Do not", like how you worded it in the above paragraph.

Same for the admonition below.

install_config/install/advanced_install.adoc
jhoyt-rh: Removed it. Thanks.

install_config/install/advanced_install.adoc
jhoyt-rh: Changed to "Do not use NFS...." Thanks.

jhoyt-rh: Changed in both places. Thanks.

jhoyt-rh:
Awaiting QE review

jhou1:
@liangxia PTAL

liangxia:
LGTM (from QE)

PR #14378, opened 2019-04-09
OSDOCS-387: Included iSCSI on AWS note.

huffmanca:
Included iSCSI on AWS note, indicating that the default security policy must be updated.

This is for OS 4.x.

### DIFF
--- a/storage/persistent-storage/persistent-storage-iscsi.adoc
+++ b/storage/persistent-storage/persistent-storage-iscsi.adoc
@@ -21,6 +21,13 @@ High-availability of storage in the infrastructure is left to the underlying
 storage provider.
 ====
 
+[IMPORTANT]
+====
+When you use iSCI on Amazon Web Services, you must update the default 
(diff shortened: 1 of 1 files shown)


storage/persistent-storage/persistent-storage-iscsi.adoc
```diff
@@ -21,6 +21,13 @@ High-availability of storage in the infrastructure is left to the underlying
 storage provider.
 ====
 
+[IMPORTANT]
+====
+When using iSCSI on AWS the default security policy must be updated to
```
kalexand-rh: Maybe
> When you use iSCI on Amazon Web Services, you must update the default security policy to include TCP traffic between nodes on the iSCSI ports. By default, they are ports \`860\` and \`3260\`.

huffmanca: Updated the text to use your suggestion!

PR #14379, opened 2019-04-09
Adding "restart" to virtctl commands

ousleyp:
CNV 1.4 docs erroneously stated that you could not use `virtctl restart` to restart a VM. This PR aims to fix this in the two places where available `virtctl` commands are listed.

Regarding QE: there is not a BZ for this issue. I emailed the cnv-devel mailing list, and Marcin Franczyk/Israel Pinto/Zhe Peng validated that the command works as expected in CNV 1.4.

### DIFF
--- a/modules/cnv_controlling_vms.adoc
+++ b/modules/cnv_controlling_vms.adoc
@@ -5,15 +5,14 @@
 [[controlvm]]
 = Controlling virtual machines
 
-Virtual machines can be started and stopped, depending on the
-current state of the virtual machine. The option to restart VMs
-is available in the Web Console only.
+You can start, stop, or restart a virtual machine, depending on its current
+state.
 
 Use the *virtctl* client utility to change the state of the virtual
 machine, open virtual console sessions with the virtual
 machines, and expose virtual machine ports as services.
 
-The *virtctl* syntax is: `virtctl <action> <VM-name>`
+The *virtctl* syntax is: `virtctl <action> <vm_name> <options>`
 
 You can only control objects in the project you are currently working
 in, unless you specify the `-n <project_name>` option.
@@ -22,6 +21,10 @@ Examples:
 
 ----
 $ virtctl start example-vm
+----
+
+----
+$ virtctl restart example-vm
 ----
 
 ----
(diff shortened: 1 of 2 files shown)


modules/cnv_controlling_vms.adoc
```diff
@@ -5,15 +5,14 @@
...
 = Controlling virtual machines
 
-Virtual machines can be started and stopped, depending on the
-current state of the virtual machine. The option to restart VMs
-is available in the Web Console only.
+You can start, stop, or restart a virtual machine, depending on its current
+state.
 
-Use the *virtctl* client utility to change the state of the virtual
+Use the `virtctl` client utility to change the state of the virtual
```
kalexand-rh: In 3.x, system packages get **s, so this was right before.

modules/cnv_controlling_vms.adoc
```diff
@@ -5,15 +5,14 @@
...
 machine, open virtual console sessions with the virtual
 machines, and expose virtual machine ports as services.
 
-The *virtctl* syntax is: `virtctl <action> <VM-name>`
+The `virtctl` syntax is: `virtctl <action> <vm_name> [flags]`
```
kalexand-rh: Why [flags]? I think it should be \<options\>.

modules/cnv_virtctl_commands.adoc
```diff
@@ -5,7 +5,9 @@
 [[virtctl-commands]]
 = Virtctl commands
 
-The *virtctl* client is a command-line utility for managing {ProductName} resources. The following table contains the *virtctl* commands used throughout this document.
+The `virtctl` client is a command-line utility for managing {ProductName}
+resources. The following table contains the `virtctl` commands used throughout
```
kalexand-rh: throughout the

kalexand-rh (commented):
Couple of picks, largely LGTM

modules/cnv_controlling_vms.adoc
ousleyp: Ah, I got [flags] from the virtctl --help output. I can definitely change it to <options> if that's what we use.

modules/cnv_controlling_vms.adoc
ousleyp: Haha, whoops. I will change it right back.

kalexand-rh:
This looks good now. Since QE's approved it, I'm merging.

PR #14382, opened 2019-04-10
OSDOCS-63: Including initial draft of certificate details.

huffmanca:
This includes the details on how OpenShift creates and uses certificates. In addition, it includes how to specify custom certificates on the OpenShift server as needed.

This is for OS 4.x.

As this is a WIP, the following is still needed:

* Understand how infra/platform certs work in 4.1, are created, rotated, replaced, revoked
* Include details on adding API server named certificate

@mrogers950 ,

I've added you as a reviewer for SME reviews as they're applicable.

### DIFF
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -152,6 +152,15 @@ Topics:
     File: configuring-google-identity-provider
   - Name: Configuring an OpenID Connect identity provider
     File: configuring-oidc-identity-provider
+- Name: Configuring certificates
+  Dir: certificates
+  Topics:
+  - Name: Replacing the default ingress certificate
+    File: replacing-default-ingress-certificate
+  - Name: Adding API server certificates
+    File: api-server
+  - Name: Securing service traffic using service serving certificates
+    File: service-serving-certificate
 - Name: Using RBAC to define and apply permissions
   File: using-rbac
   Distros: openshift-enterprise,openshift-origin
--- a/authentication/certificates/api-server.adoc
+++ b/authentication/certificates/api-server.adoc
@@ -1,0 +1,15 @@
+[id="api-server-certificates"]
+= Adding API server certificates
+include::modules/common-attributes.adoc[]
+:context: api-server-certificates
+
+toc::[]
+
+The default API server certificate is issued by an internal {product-title}
+cluster CA. Clients outside of the cluster will not be able to verify the
+API server's certificate by default. This certificate can be replaced
+by one that is issued by a CA that clients trust.
+
+include::modules/customize-certificates-api-add-default.adoc[leveloffset=+1]
+
+include::modules/customize-certificates-api-add-named.adoc[leveloffset=+1]
--- a/authentication/certificates/service-serving-certificate.adoc
+++ b/authentication/certificates/service-serving-certificate.adoc
@@ -1,0 +1,24 @@
+[id="add-service-serving"]
+= Securing service traffic using service serving certificate secrets
+include::modules/common-attributes.adoc[]
+:context: service-serving-certificate
+
+toc::[]
+
+include::modules/customize-certificates-understanding-service-serving.adoc[leveloffset=+1]
+
+[IMPORTANT]
+====
+The service CA certificate is automatically rotated during upgrades, but
+must be
+xref:../../authentication/certificates/service-serving-certificate.adoc#manually-rotate-service-ca_{context}[manually rotated]
+in the event that the cluster is not upgraded.
+====
+
+include::modules/customize-certificates-add-service-serving.adoc[leveloffset=+1]
+
+include::modules/customize-certificates-add-service-serving-configmap.adoc[leveloffset=+1]
+
+include::modules/customize-certificates-rotate-service-serving.adoc[leveloffset=+1]
+
+include::modules/customize-certificates-manually-rotate-service-ca.adoc[leveloffset=+1]
--- a/modules/customize-certificates-add-service-serving-configmap.adoc
+++ b/modules/customize-certificates-add-service-serving-configmap.adoc
@@ -1,0 +1,57 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/service-serving-certificate.adoc
+
+[id="add-service-certificate-configmap_{context}"]
+= Add a service certificate to a ConfigMap
+
+A Pod can access the service CA certificate by mounting a ConfigMap that
+is annotated with `service.beta.openshift.io/inject-cabundle=true`.
+Once annotated, the cluster automatically injects the service CA
+certificate into the `service-ca.crt` key on the ConfigMap. Access to
+this CA certificate allows TLS clients to verify connections to
+services using service serving certificates.
+
+[IMPORTANT]
+====
+After adding this annotation to a ConfigMap all existing data in it is
+deleted. It is recommended to use a separate ConfigMap to contain the
+`service-ca.crt`, instead of using the same ConfigMap that stores your
+Pod's configuration.
+====
+
+.Procedure
+
+. Annotate the ConfigMap with `service.beta.openshift.io/inject-cabundle=true`.
++
+----
+$ oc annotate configmap <configmap-name> \//<1>
+     service.beta.openshift.io/inject-cabundle=true
+----
+<1> Replace `<configmap-name>` with the name of the ConfigMap to annotate.
++
+[NOTE]
+====
+Explicitly referencing the `service-ca.crt` key in a volumeMount will
+prevent a Pod from starting until the ConfigMap has been injected with
+the CA bundle.
+====
++
+For instance, to annotate the ConfigMap `foo` the following command would be
+used:
++
+----
+$ oc annotate configmap foo service.beta.openshift.io/inject-cabundle=true
+----
+
+. View the ConfigMap to ensure the certificate has been generated. This
+appears as a `service-ca.crt` in the YAML output.
++
+----
+$ oc get configmap <configmap-name> -o yaml
+apiVersion: v1
+data:
+  service-ca.crt: |
+    -----BEGIN CERTIFICATE-----
+...
+----
--- a/modules/customize-certificates-add-service-serving.adoc
+++ b/modules/customize-certificates-add-service-serving.adoc
@@ -1,0 +1,53 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/service-serving-certificate.adoc
+
+[id="add-service-certificate_{context}"]
+= Add a service certificate
+
+To secure communication to your service, generate a
+signed serving certificate and key pair into a secret in the same
+namespace as the service.
+
+[IMPORTANT]
+====
+The generated certificate is only valid for the internal service DNS name
+`<service.name>.<service.namespace>.svc`, and are only valid for
+internal communications.
+====
+
+.Prerequisites:
+
+* You must have a service defined.
+
+.Procedure
+
+. Annotate the service with `service.beta.openshift.io/serving-cert-secret-name`.
++
+----
+$ oc annotate service <service-name> \//<1>
+     service.beta.openshift.io/serving-cert-secret-name=<secret-name> //<2>
+----
+<1> Replace `<service-name>` with the name of the service to secure.
+<2> `<secret-name>` will be the name of the generated secret containing the
+certificate and key pair. For convenience, it is recommended that this
+be the same as `<service-name>`.
++
+For instance, use the following command to annotate the service `foo`:
++
+----
+$ oc annotate service foo service.beta.openshift.io/serving-cert-secret-name=foo
+----
+
+. Examine the service to confirm the annotations are present.
++
+----
+$ oc describe service <service-name>
+...
+Annotations:              service.beta.openshift.io/serving-cert-secret-name: <service-name>
+                          service.beta.openshift.io/serving-cert-signed-by: openshift-service-serving-signer@1556850837
+...
+----
+
+. After the cluster generates a secret for your service, your PodSpec can
+mount it, and the Pod will run after it becomes available.
--- a/modules/customize-certificates-api-add-default.adoc
+++ b/modules/customize-certificates-api-add-default.adoc
@@ -1,0 +1,56 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/api-server.adoc
+
+[id="add-default-api-server_{context}"]
+= Add an API server default certificate
+
+To allow clients outside the cluster to validate the API server's
+certificate, you can replace the default certificate
+with one that is issued by a public or organizational CA.
+
+.Prerequisites
+
+* You must have a valid certificate and key in the PEM format.
+
+.Procedure
+
+. Create a secret that contains the certificate and key in the
+`openshift-config` namespace.
++
+----
+$ oc create secret tls <certificate> \//<1>
+     --cert=</path/to/cert.crt> \//<2>
+     --key=</path/to/cert.key> \//<3>
+     -n openshift-config
+----
+<1> `<certificate>` is the name of the secret that will contain
+the certificate.
+<2> `</path/to/cert.crt>` is the path to the certificate on your
+local file system.
+<3> `</path/to/cert.key>` is the path to the private key associated
+with this certificate.
+
+. Update the API server to reference the created secret.
++
+----
+$ oc patch apiserver cluster \
+     --type=merge -p \
+     '{"spec": {"servingCerts": {"defaultServingCertificate":
+     {"name": "<certificate>"}}}}' <1>
+----
+<1> Replace `<certificate>` with the name used for the secret in
+the previous step.
+
+. Examine the `apiserver/cluster` object and confirm the secret is now
+referenced.
++
+----
+$ oc describe apiserver cluster
+...
+Spec:
+  Serving Certs:
+    Default Serving Certificate:
+      Name:  <certificate>
+...
+----
--- a/modules/customize-certificates-api-add-named.adoc
+++ b/modules/customize-certificates-api-add-named.adoc
@@ -1,0 +1,68 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/api-server.adoc
+
+[id="add-named-api-server_{context}"]
+= Add an API server named certificate
+
+The default API server certificate is issued by an internal {product-title}
+cluster CA. You can add additional certificates to the API server to send
+based on the client's requested URL, such as when a reverse proxy or
+load balancer is used.
+
+.Prerequisites
+
+* You must have the certificate and key, in the PEM format, for the
+client's URL.
+* The certificate must be issued for the URL used by the client to
+reach the API server.
+* The certificate must have the `subjectAltName` extension for the URL.
+
+.Procedure
+
+. Create a secret that contains the certificate and key in the
+`openshift-config` namespace.
++
+----
+$ oc create secret tls <certificate> \//<1>
+     --cert=</path/to/cert.crt> \//<2>
+     --key=</path/to/cert.key> \//<3>
+     -n openshift-config
+----
+<1> `<certificate>` is the name of the secret that will contain
+the certificate.
+<2> `</path/to/cert.crt>` is the path to the certificate on your
+local file system.
+<3> `</path/to/cert.key>` is the path to the private key associated
+with this certificate.
+
+. Update the API server to reference the created secret.
++
+----
+$ oc patch apiserver cluster \
+     --type=merge -p \
+     '{"spec":{"servingCerts": {"namedCertificates":
+     [{"names": ["<hostname>"], //<1>
+     "servingCertificate": {"name": "<certificate>"}}]}}}' <2>
+----
+<1> Replace `<hostname>` with the hostname that the API server
+should provide the certificate for.
+<2> Replace `<certificate>` with the name used for the secret in
+the previous step.
+
+. Examine the `apiserver/cluster` object and confirm the secret is now
+referenced.
++
+----
+$ oc describe apiserver cluster
+...
+Spec:
+  Serving Certs:
+    Named Certificates:
+      Names:
+        <hostname>
+      Serving Certificate:
+        Name:  <certificate>
+...
+----
+
--- a/modules/customize-certificates-manually-rotate-service-ca.adoc
+++ b/modules/customize-certificates-manually-rotate-service-ca.adoc
@@ -1,0 +1,59 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/service-signing-certificates.adoc
+
+[id="manually-rotate-service-ca_{context}"]
+= Manually rotate the service CA certificate
+
+The service CA is valid for one year after {product-title} is
+installed. If you have a cluster that has been running
+for one year without upgrades, then you must manually
+refresh the service CA.
+
+This process is automatically performed during an upgrade.
+Follow these steps only if you do not update your cluster.
+
+.Prerequisites
+
+* You must be logged in as a cluster admin.
+
+.Procedure
+
+. View the expiration date of the current service CA certificate by
+using the following command.
++
+----
+$ oc get secrets/signing-key -n openshift-service-ca \
+     -o template='{{index .data "tls.crt"}}' \
+     | base64 -d \
+     | openssl x509 -noout -enddate
+----
++
+If this certificate expires before a scheduled upgrade, then proceed
+through the following steps. Otherwise, there is no need to perform
+the rotation at this time.
+
+. Manually rotate the service CA. This process generates a new service CA
+which will be used to sign the new service certificates.
++
+----
+$ oc delete secret/signing-key -n openshift-service-ca
+----
+
+. To apply the new certificates to all services, restart all the Pods
+in your cluster. This command ensures that all services use the
+updated certificates.
++
+----
+$ for I in $(oc get ns -o jsonpath='{range .items[*]} {.metadata.name}{"\n"} {end}'); \
+      do oc delete pods --all -n $I; \
+      sleep 1; \
+      done
+----
++
+[WARNING]
+====
+This command will cause a service interruption, as it goes through and
+deletes every running pod in every namespace. These pods will automatically
+restart after they are deleted.
+====
--- a/modules/customize-certificates-replace-default-router.adoc
+++ b/modules/customize-certificates-replace-default-router.adoc
@@ -1,0 +1,47 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/replacing-default-ingress-certificate.adoc
+
+[id="replacing-default-ingress_{context}"]
+= Replacing the default ingress certificate
+
+You can replace the default ingress certificate for all
+applications under the `.apps` subdomain. After you replace
+the certificate, all applications, including the web console
+and CLI, will have encryption provided by specified certificate.
+
+.Prerequisites
+
+* You must have a wildcard certificate and its private key,
+both in the PEM format, for use.
+* The certificate must have a `subjectAltName` extension of
+`*apps.<clustername>.<domain>`.
+
+.Procedure
+
+. Create a secret that contains the wildcard certificate and key:
++
+----
+$ oc create secret tls <certificate> \//<1>
+     --cert=</path/to/cert.crt> \//<2>
+     --key=</path/to/cert.key> \//<3>
+     -n openshift-ingress
+----
+<1> `<certificate>` is the name of the secret that will contain
+the certificate and private key.
+<2> `</path/to/cert.crt>` is the path to the certificate on your
+local file system.
+<3> `</path/to/cert.key>` is the path to the private key associated
+with this certificate.
+
+. Update the ingress controller configuration with the newly created
+secret:
++
+----
+$ oc patch ingresscontroller.operator default \
+     --type=merge -p \
+     '{"spec":{"defaultCertificate": {"name": "<certificate>"}}}' \//<1>
+     -n openshift-ingress-operator
+----
+<1> Replace `<certificate>` with the name used for the secret in
+the previous step.
--- a/modules/customize-certificates-rotate-service-serving.adoc
+++ b/modules/customize-certificates-rotate-service-serving.adoc
@@ -1,0 +1,47 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/service-serving-certificate.adoc
+
+[id="rotate-service-serving_{context}"]
+= Manually rotate the generated service certificate
+
+You can rotate the service certificate by deleting the
+associated secret. Deleting the secret results in a new one
+being automatically created, resulting in a new certificate.
+
+.Prerequisites
+
+* A secret containing the certificate and key pair must
+have been generated for the service.
+
+.Procedure
+
+. Examine the service to determine the secret containing the
+certificate. This is found in the `serving-cert-secret-name`
+annotation, as seen below.
++
+----
+$ oc describe service <service-name>
+...
+service.beta.openshift.io/serving-cert-secret-name: <secret>
+...
+----
+
+. Delete the generated secret for the service. This process
+will automatically recreate the secret.
++
+----
+$ oc delete secret <secret> //<1>
+----
+<1> Replace `<secret>` with the name of the secret from the previous
+step.
+
+. Confirm that the certificate has been recreated
+by obtaining the new secret and examining the `AGE`.
++
+----
+$ oc get secret <service-name>
+
+NAME              TYPE                DATA   AGE
+<service.name>    kubernetes.io/tls   2      1s
+----
--- a/modules/customize-certificates-understanding-default-router.adoc
+++ b/modules/customize-certificates-understanding-default-router.adoc
@@ -1,0 +1,19 @@
+// Module included in the following assemblies:
+//
+// authentication/certificates/replacing-default-ingress-certificate.adoc
+
+[id="understanding-default-ingress_{context}"]
+= Understanding the default ingress certificate
+
+By default {product-title} uses the Ingress Operator to
+create an internal CA and issue a wildcard certificate that is valid for
+applications under the `.apps` sub-domain. Both the web console and CLI
+use this certificate as well.
+
+The internal infrastructure CA certificates are self-signed.
+While this process might be perceived as bad practice by some security or
+PKI teams, any risk here is minimal. The only clients that implicitly
+trust these certificates are other components within the cluster.
+Replacing the default wildcard certificate with one that is issued by a
+public or organizational CA will allow external clients to connect
+securely to applications running under the `.apps` sub-domain.
--- a/modules/customize-certificates-understanding-service-serving.adoc
+++ b/modules/customize-certificates-understanding-service-serving.adoc
@@ -1,0 +1,19 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/service-serving-certificate.adoc
+
+[id="understanding-service-serving_{context}"]
+= Understanding service serving certificates
+
+Service serving certificates are intended to support complex
+middleware applications that require encryption. These certificates are
+issued as TLS web server certificates.
+
+The `service-ca` controller uses the `x509.SHA256WithRSA` signature
+algorithm to generate service certificates.
+
+The generated certificate and key are in PEM format, stored in `tls.crt`
+and `tls.key` respectively, within a created secret. The
+certificate and key are automatically replaced when they get close to
+expiration. The service CA certificate, which signs the service
+certificates, is only valid for one year after {product-title} is installed.
--- a/authentication/certificates/replacing-default-ingress-certificate.adoc
+++ b/authentication/certificates/replacing-default-ingress-certificate.adoc
@@ -1,0 +1,10 @@
+[id="replacing-default-ingress"]
+= Replacing the default ingress certificate
+include::modules/common-attributes.adoc[]
+:context: replacing-default-ingress
+
+toc::[]
+
+include::modules/customize-certificates-understanding-default-router.adoc[leveloffset=+1]
+
+include::modules/customize-certificates-replace-default-router.adoc[leveloffset=+1]


huffmanca:
@mrogers950 This should be ready for review. A link to the updated preview was sent privately.

_topic_map.yml
```diff
@@ -262,6 +262,15 @@ Topics:
     File: configuring-google-identity-provider
   - Name: Configuring an OpenID Connect identity provider
     File: configuring-oidc-identity-provider
+- Name: Configuring certificates
+  Dir: certificates
+  Topics:
+  - Name: Replacing the default router certificate
```
vikram-redhat: Aren't we calling router 'ingress' now? @bmcelvee can you confirm?

huffmanca: You're correct. I spoke with her about this Friday afternoon. I'll update these references.

huffmanca: These instances have been renamed to `ingress`.

modules/customize-certificates-api-add-default.adoc
```diff
@@ -0,0 +1,29 @@
...
+`openshift-config` namespace.
++
+----
+$ oc create secret tls <certificate> --cert=</path/to/cert.crt> --key=</path/to/cert.key> -n openshift-config
+----
+
+. Update the API server to reference the created secret.
++
+----
+$ oc patch apiserver cluster --type=merge -p "{\"spec\": {\"servingCerts\": {\"defaultServingCertificate\": {\"name\": \"<certificate>\"}}}}"
```
sferich888: What namespace are you doing this in?

modules/customize-certificates-manually-rotate-service-ca.adoc
```diff
@@ -0,0 +1,39 @@
...
++
+----
+$ oc delete secret/signing-key -n openshift-service-ca
+----
+
+. Restart all Pods. This refreshes the created service 
+certificates, ensuring that all services use the updated certificates.
++
+----
+$ oc delete pods --all --all-namespaces
```
sferich888: This seems dangerous! Should we not do this more systematic?

sferich888 (changes requested):

modules/customize-certificates-add-service-serving.adoc
```diff
@@ -0,0 +1,27 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/service-serving-certificate.adoc
+
+[id="add-service-certificate-{context}"]
+= Add a service certificate
+
+To secure communication to your service, have the cluster generate a 
+signed serving certificate/key pair into a secret in your namespace. To 
+do this, set the `service.alpha.openshift.io/serving-cert-secret-name`
```
mrogers950: service.beta now

modules/customize-certificates-api-add-default.adoc
```diff
@@ -0,0 +1,29 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/api-server.adoc
+
+[id="add-default-api-server-{context}"]
+= Add an API server default certificate
+
+The default API server certificate is issued by an internal {product-title} 
+cluster CA. This certificate can be replaced so that any API clients, such
+as `oc`, will use the defined certificate.
```
mrogers950: The reason to replace the API server cert is so that external clients can verify it, they don't technically "use" the server cert, so this text is incorrect. We should try to convey that in the default case, clients will not be able to verify the server cert. So the purpose of this is to replace the API server certificate by one that is signed by a CA that your external clients have access to in their trust stores. Something like:

"The default API server certificate is issued by an internal {product-title} 
cluster CA. Clients outside of the cluster will not be able to verify the API server's certificate by default.  This certificate can be replaced by one that is issued by a CA that clients trust."

modules/customize-certificates-api-add-named.adoc
```diff
@@ -0,0 +1,30 @@
...
+[id="add-named-api-server-{context}"]
+= Add an API server named certificate
+
+The default API server certificate is issued by an internal {product-title} 
+cluster CA. Additional certificates may be added to the API server to send
+based on the client's requested URL.
+
+.Prerequisites
+
+* You must have the certificate and key for the client's URL.
```
mrogers950: Might benefit from a little more detail:

The cert and key files need to be in PEM format.
The cert needs to be issued for the URL that is used by the client to reach the API server. This means it needs a SAN extension (subjectAltName) for the URL.

modules/customize-certificates-manually-rotate-service-ca.adoc
```diff
@@ -0,0 +1,39 @@
...
+$ oc get secrets/signing-key -n openshift-service-ca -o template='{{index .data "tls.crt"}}' | base64 -d | openssl x509 -noout -enddate
+----
+
+. Manually rotate the service CA.
++
+----
+$ oc delete secret/signing-key -n openshift-service-ca
+----
+
+. Restart all Pods. This refreshes the created service
```
mrogers950: Might need a little clarification. This reads as if this triggers the actual certificate refresh. It's actually refreshing the pods so that they pick up the newly rotated serving certs.

modules/customize-certificates-replace-default-router.adoc
```diff
@@ -0,0 +1,29 @@
...
+[id="replacing-default-ingress-{context}"]
+= Replacing the default ingress certificate
+
+You can replace the default ingress certificate for all
+applications under the `.apps` subdomain. This will update
+the certificate used for the web console and CLI.
+
+.Prerequisites
+
+* You must have a wildcard certificate for use.
```
mrogers950: cert/key in PEM format.
A wildcard certificate that is issued for (SAN) for *.apps.<domain>.

modules/customize-certificates-add-service-serving.adoc
```diff
@@ -0,0 +1,27 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/service-serving-certificate.adoc
```
mrogers950: We need to also document that process for annotating a configmap to get the CA bundle. This can be taken from the service-ca-operator repo README.

modules/customize-certificates-add-service-serving.adoc
mrogers950: this goes for the rest of the service annotations we mention in these docs

modules/customize-certificates-api-add-default.adoc
huffmanca: The `oc patch` command for the ingress controller specifies a namespace. The apiserver does not require one. I've tested this both without a namespace and with one, and received confirmation that one is not required.

modules/customize-certificates-manually-rotate-service-ca.adoc
huffmanca: After speaking with engineering I've changed this command to be a for loop that goes across all namespaces and sleeps for a second before proceeding to the next. I've also added a warning beneath it to indicate that this will cause a service interruption.

modules/customize-certificates-add-service-serving.adoc
huffmanca: All of these references have been updated to `service.beta.openshift...`.

modules/customize-certificates-api-add-default.adoc
huffmanca: This text has been updated to use the recommendation.

modules/customize-certificates-api-add-named.adoc
huffmanca: I've made these changes for the `API server named certificate` section. Does the default certificate require the PEM format as well?

modules/customize-certificates-manually-rotate-service-ca.adoc
huffmanca: This section has been updated, and now reads as:

> Restart all Pods. This refreshes all Pods so that they receive the newly
> created service certificates created by the recently rotated service CA. 
> This command ensures that all services use the updated certificates.

modules/customize-certificates-replace-default-router.adoc
huffmanca: These prerequisites have been included.

modules/customize-certificates-add-service-serving.adoc
huffmanca: These instructions have been included immediately below the instructions on adding the service certificates.

huffmanca:
@openshift/team-documentation ,

Is it possible for someone to peer review this one?

modules/customize-certificates-add-service-serving.adoc
```diff
@@ -0,0 +1,27 @@
...
+
+To secure communication to your service, have the cluster generate a 
+signed serving certificate and key pair into a secret in your namespace. To 
+do this, set the `service.beta.openshift.io/serving-cert-secret-name` 
+annotation on your service with the value set to the name you want to use 
+for your secret.
+
+.Procedure
+
+* Annotate the service with `service.beta.openshift.io/serving-cert-secret-name`.
```
kalexand-rh: I am confused. Should this step be 
```
* Set the name of the secret that your service uses:
+
----
$ oc annotate service <service-name> service.beta.openshift.io/serving-cert-secret-name=<service-name> <1>
----
<1> For `<service-name>`, specify the name of the secret for your service.
...
```

Is there a good way to validate that this worked?

modules/customize-certificates-add-service-serving.adoc
```diff
@@ -0,0 +1,27 @@
...
++
+----
+$ oc annotate service <service-name> service.beta.openshift.io/serving-cert-secret-name=<service-name>
+----
++
+For instance, to annotate the service `foo` the following command would be
```
kalexand-rh: s/the following command would be used:/use the following command:

>! would v
Use only to refer to an unfulfilled condition or hypothesis and to express the future in the past... Do
not use as an auxiliary verb...

modules/customize-certificates-understanding-service-serving.adoc
```diff
@@ -0,0 +1,23 @@
...
+mount it and the Pod will run as it becomes available. The generated
+certificate will be valid for the internal service DNS name 
+`<service.name>.<service.namespace>.svc`.
+
+The generated certificate and key are in PEM format, stored in `tls.crt` 
+and `tls.key` respectively, within the created secret. The 
+certificate and key pair is automatically replaced when it gets close to 
+expiration. 
+
+The signature algorithm in use is `x509.SHA256WithRSA`.
```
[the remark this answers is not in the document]
kalexand-rh: I'd move this line earlier in the procedure so the bit about expiration is closer to the note. I'd also make some mention of the expiration period in the concept in case it gets reused without the note.

authentication/certificates/service-serving-certificate.adoc
```diff
@@ -0,0 +1,27 @@
+[id="add-service-serving"]
+= Service serving certificate secrets
```
kalexand-rh: This is titled like a pure conceptual assembly but includes procedures. Please change the title.

I think this assembly should be broken into two, one about setting and one about manually rotating. Please check and make sure that you have one user story per assembly.

modules/customize-certificates-view-service-serving.adoc
```diff
@@ -0,0 +1,35 @@
+// Module included in the following assembly:
+//
+// * authentication/certificates/service-serving-certificate.adoc
+
+[id="view-service-serving-{context}"]
+= View the expiration date of the generated service certificate
+
+While the certificate and key pair is automatically replaced when
+it gets close to expiration, it is possible to view the expiration
```
kalexand-rh: ~~it is possible to view the expiration date within the secret.~~ you can view the secret's expiration date.

Why would I want to do this?

modules/customize-certificates-view-service-serving.adoc
```diff
@@ -0,0 +1,35 @@
...
+date within the secret.
+
+.Prerequisites
+
+* A secret containing the certificate and key pair must
```
kalexand-rh: s/must have been generated/must exist

modules/customize-certificates-rotate-service-serving.adoc
```diff
@@ -0,0 +1,29 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/service-serving-certificate.adoc
+
+[id="rotate-service-serving-{context}"]
+= Manually rotate the generated service certificate
+
```
kalexand-rh: You need an intro, maybe, "If you /<reason why you'd do this\>, you must manually rotate the generated service certificate before it expires. \<dire warning about what happens if you don't.\>"

modules/customize-certificates-rotate-service-serving.adoc
```diff
@@ -0,0 +1,29 @@
...
+* A secret containing the certificate and key pair must 
+have been generated for the service.
+
+.Procedure
+
+. Delete the generated secret for the service:
++
+----
+$ oc delete secret <service.name>
```
kalexand-rh: This needs an annotation.

I'd also add the step to determine what the right secret is, if possible.

modules/customize-certificates-rotate-service-serving.adoc
```diff
@@ -0,0 +1,29 @@
...
+----
```
kalexand-rh: I'd also add a note after the command that the secret's automatically regenerated under the same name. (Or whatever's really happening here.)

modules/customize-certificates-manually-rotate-service-ca.adoc
```diff
@@ -0,0 +1,50 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/service-signing-certificates.adoc
+
+[id="manually-rotate-service-ca-{context}"]
+= Manually rotate the service CA certificate
+
+The service CA is valid for one year after {product-title} is 
+installed. If you have a cluster that has been running
+for one year without upgrades, then you will need to manually
```
kalexand-rh: s/will need to /must

modules/customize-certificates-manually-rotate-service-ca.adoc
```diff
@@ -0,0 +1,50 @@
...
+refresh the service CA.
+
+This process is automatically performed during an upgrade,
+and should only be followed for clusters that do not perform
```
kalexand-rh: s/upgrade, and should only be followed for/upgrade. Follow these steps only if you do not update your cluster.

modules/customize-certificates-manually-rotate-service-ca.adoc
```diff
@@ -0,0 +1,50 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/service-signing-certificates.adoc
+
+[id="manually-rotate-service-ca-{context}"]
+= Manually rotate the service CA certificate
```
kalexand-rh: This really needs to be part of a separate assembly about manually rotating the certs, and I think that some of this info might need to be at the assembly level.

I think you might need a little more info about the CA certificate in general. This is the only place that it's mentioned, and I'm, again, slightly confused about its importance.

modules/customize-certificates-manually-rotate-service-ca.adoc
```diff
@@ -0,0 +1,50 @@
...
+for one year without upgrades, then you will need to manually
+refresh the service CA.
+
+This process is automatically performed during an upgrade,
+and should only be followed for clusters that do not perform
+upgrades.
+
+.Prerequisites
+
+* You must be logged in as a cluster admin.
```
kalexand-rh: Is this really the only prereq?

modules/customize-certificates-manually-rotate-service-ca.adoc
```diff
@@ -0,0 +1,50 @@
...
+
+.Procedure
+
+. View the expiration date of the current service CA certificate by
+using the following command. If this certificate expires
```
kalexand-rh: Move the second sentence after the command.

modules/customize-certificates-manually-rotate-service-ca.adoc
```diff
@@ -0,0 +1,50 @@
...
+created service certificates created by the recently rotated service CA. 
+This command ensures that all services use the updated certificates.
++
+----
+$ for I in $(oc get ns -o jsonpath='{range .items[*]} {.metadata.name}{"\n"} {end}'); do oc delete pods --all -n $I; sleep 1; done
+----
++
+[WARNING]
+====
+This command will cause a service interruption, as it goes through and
```
kalexand-rh: > Running this command causes a service interruption because it deletes every running Pod in every namespace.

And then they automatically restart, yes?

modules/customize-certificates-manually-rotate-service-ca.adoc
```diff
@@ -0,0 +1,50 @@
...
++
+----
+$ oc get secrets/signing-key -n openshift-service-ca -o template='{{index .data "tls.crt"}}' | base64 -d | openssl x509 -noout -enddate
+----
+
+. Manually rotate the service CA.
++
+----
+$ oc delete secret/signing-key -n openshift-service-ca
+----
```
kalexand-rh: Add a paragraph about rotating the service CA creating the new certificates.

modules/customize-certificates-manually-rotate-service-ca.adoc
```diff
@@ -0,0 +1,50 @@
...
+
+. Restart all Pods. This refreshes all Pods so that they receive the newly
```
kalexand-rh: > To apply the new certificates to all services, restart all the Pods in your cluster.

modules/customize-certificates-manually-rotate-service-ca.adoc
```diff
@@ -0,0 +1,50 @@
...
++
+----
+$ for I in $(oc get ns -o jsonpath='{range .items[*]} {.metadata.name}{"\n"} {end}'); do oc delete pods --all -n $I; sleep 1; done
+----
++
+[WARNING]
+====
+This command will cause a service interruption, as it goes through and
+deletes every running pod in every namespace.
+====
```
kalexand-rh: Is there another logical command to put here? Like watching the Pods come back online or checking to see if they have the right cert?

kalexand-rh (commented):
I have some questions and suggestions.

modules/customize-certificates-add-service-serving-configmap.adoc
```diff
@@ -0,0 +1,49 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/service-serving-certificate.adoc
+
+[id="add-service-certificate-configmap-{context}"]
+= Add a service certificate to a ConfigMap
+
+To secure communication to services that consume a ConfigMap,
```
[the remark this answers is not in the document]
huffmanca: I've adjusted this to be `use` instead of consume.

modules/customize-certificates-add-service-serving-configmap.adoc
```diff
@@ -0,0 +1,49 @@
...
+[id="add-service-certificate-configmap-{context}"]
+= Add a service certificate to a ConfigMap
+
+To secure communication to services that consume a ConfigMap, 
+have the cluster generate a signed serving certificate and key pair into the
+ConfigMap. To do this, set the 
+`service.beta.openshift.io/inject-cabundle=true` 
+annotation on your ConfigMap. A certificate will be generated, which allows
+consumers to trust the `service-ca.crt` in their TLS client configuration,
+allowing connections to services that utilize these service sharing
```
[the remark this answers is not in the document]
huffmanca: Updated this to use the recommendations.

modules/customize-certificates-add-service-serving-configmap.adoc
```diff
@@ -0,0 +1,49 @@
...
+
+[id="add-service-certificate-configmap-{context}"]
+= Add a service certificate to a ConfigMap
+
+To secure communication to services that consume a ConfigMap, 
+have the cluster generate a signed serving certificate and key pair into the
+ConfigMap. To do this, set the 
+`service.beta.openshift.io/inject-cabundle=true` 
+annotation on your ConfigMap. A certificate will be generated, which allows
+consumers to trust the `service-ca.crt` in their TLS client configuration,
```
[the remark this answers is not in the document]
huffmanca: My understanding is that this would be anything that uses the ConfigMap. This could be a Pod, a controller, or anything else that references the ConfigMap. I'm leaving this one as `consumers` for now.

I've added `generated` before `service-ca.crt` to indicate this certificate is created and does not need to be supplied. Hopefully the procedure below also indicates this, indicating that the `service-ca.crt` has been added to the ConfigMap by the previous step.

modules/customize-certificates-add-service-serving-configmap.adoc
```diff
@@ -0,0 +1,49 @@
...
+allowing connections to services that utilize these service sharing 
+certificates.
+
+.Procedure
+
+. Annotate the ConfigMap with `service.beta.openshift.io/inject-cabundle=true`.
++
+----
+$ oc annotate configmap <configmap-name> service.beta.openshift.io/inject-cabundle=true
```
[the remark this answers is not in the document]
huffmanca: It's not required. If the ConfigMap is in a separate namespace then it must be specified, but otherwise the command works without issue.

modules/customize-certificates-understanding-default-router.adoc
```diff
@@ -0,0 +1,18 @@
+// Module included in the following assemblies:
+//
+// authentication/certificates/replacing-default-ingress-certificate.adoc
+
+[id="understanding-default-ingress-{context}"]
+= Understanding the default ingress certificate
+
+By default {product-title} uses the Ingress Controller Operator to 
+create an internal CA and issue a wildcard certificate that is valid for 
+applications under the `.apps` sub-domain. This includes the web console
```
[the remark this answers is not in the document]
huffmanca: I've reworded this sentence to be:

> Both the web console and CLI use this certificate as well.

modules/customize-certificates-understanding-default-router.adoc
```diff
@@ -0,0 +1,18 @@
...
+= Understanding the default ingress certificate
+
+By default {product-title} uses the Ingress Controller Operator to 
+create an internal CA and issue a wildcard certificate that is valid for 
+applications under the `.apps` sub-domain. This includes the web console 
+and CLI. 
+
+The internal infrastructure certificates are self-signed, 
+which might be perceived as bad practice by some security or PKI teams. 
+Any risk here is minimal, as the only clients that trust these
```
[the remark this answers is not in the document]
huffmanca: Reworded.

modules/customize-certificates-understanding-default-router.adoc
```diff
@@ -0,0 +1,18 @@
...
+[id="understanding-default-ingress-{context}"]
+= Understanding the default ingress certificate
+
+By default {product-title} uses the Ingress Controller Operator to 
+create an internal CA and issue a wildcard certificate that is valid for 
+applications under the `.apps` sub-domain. This includes the web console 
+and CLI. 
+
+The internal infrastructure certificates are self-signed, 
+which might be perceived as bad practice by some security or PKI teams.
```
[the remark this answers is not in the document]
huffmanca: I've left this in its current place, but I've reworded the sentences so that the bad is immediately countered by why it isn't an issue in the same sentence.

modules/customize-certificates-understanding-default-router.adoc
```diff
@@ -0,0 +1,18 @@
...
+By default {product-title} uses the Ingress Controller Operator to 
+create an internal CA and issue a wildcard certificate that is valid for 
+applications under the `.apps` sub-domain. This includes the web console 
+and CLI. 
+
+The internal infrastructure certificates are self-signed, 
+which might be perceived as bad practice by some security or PKI teams. 
+Any risk here is minimal, as the only clients that trust these 
+certificates are other components within the cluster. If the default
+wildcard certificate is replaced, then all external users and systems
```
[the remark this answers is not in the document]
huffmanca: I've left the first suggestion alone, as this is the only possible wildcard certificate. I've incorporated the second suggestion.

modules/customize-certificates-replace-default-router.adoc
```diff
@@ -0,0 +1,30 @@
...
+the certificate used for the web console and CLI.
+
+.Prerequisites
+
+* You must have a wildcard certificate, in the PEM format, for use. 
+* The certificate must have a `subjectAltName` extension of `*.apps.`.
+
+.Procedure
+
+. Create a secret containing the wildcard certificate and key:
```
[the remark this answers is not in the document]
huffmanca: Incorporated.

modules/customize-certificates-replace-default-router.adoc
```diff
@@ -0,0 +1,30 @@
...
+[id="replacing-default-ingress-{context}"]
+= Replacing the default ingress certificate
+
+You can replace the default ingress certificate for all
+applications under the `.apps` subdomain. This will update
+the certificate used for the web console and CLI.
+
+.Prerequisites
+
+* You must have a wildcard certificate, in the PEM format, for use.
```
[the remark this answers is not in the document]
huffmanca: Matt mentioned the certificate requirements, both of which are included here.

modules/customize-certificates-replace-default-router.adoc
```diff
@@ -0,0 +1,30 @@
...
+
+* You must have a wildcard certificate, in the PEM format, for use. 
+* The certificate must have a `subjectAltName` extension of `*.apps.`.
+
+.Procedure
+
+. Create a secret containing the wildcard certificate and key:
++
+----
+$ oc create secret tls <certificate> --cert=</path/to/cert.crt> --key=</path/to/cert.key> -n openshift-ingress
```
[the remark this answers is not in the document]
huffmanca: This command has been placed on multiple lines and annotated.

modules/customize-certificates-replace-default-router.adoc
```diff
@@ -0,0 +1,30 @@
...
+----
+
+. Update the Ingress controller configuration with the newly created
+secret:
++
+----
+$ oc patch ingresscontroller.operator cluster --type=merge -p "{\"spec\":{\"defaultCertificate\":{\”name\”: \"<certificate>\"}}}" -n openshift-ingress-operator
```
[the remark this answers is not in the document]
huffmanca: This command has been placed on multiple lines and annotated.

modules/customize-certificates-manually-rotate-service-ca.adoc
huffmanca: The command tells you as each pod is restarted. I'm going to leave this as is for now, as if you've executed the steps in order then nothing else is required.

modules/customize-certificates-manually-rotate-service-ca.adoc
huffmanca: Updated to use the recommendation.

modules/customize-certificates-manually-rotate-service-ca.adoc
huffmanca: I've included an extra sentence here indicating that it generates a new service CA which is used to sign the new service certificates.

modules/customize-certificates-manually-rotate-service-ca.adoc
huffmanca: Included this in an additional sentence.

modules/customize-certificates-manually-rotate-service-ca.adoc
huffmanca: Moved to after the command.

modules/customize-certificates-add-service-serving-configmap.adoc
```diff
@@ -0,0 +1,49 @@
...
+
+To secure communication to services that consume a ConfigMap, 
+have the cluster generate a signed serving certificate and key pair into the
+ConfigMap. To do this, set the 
+`service.beta.openshift.io/inject-cabundle=true` 
+annotation on your ConfigMap. A certificate will be generated, which allows
+consumers to trust the `service-ca.crt` in their TLS client configuration,
+allowing connections to services that utilize these service sharing 
+certificates.
+
```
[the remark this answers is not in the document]
huffmanca: I'm leaving this off, as the section header indicates that this is for adding the certificate to ConfigMaps.

modules/customize-certificates-api-add-default.adoc
```diff
@@ -0,0 +1,30 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/api-server.adoc
+
+[id="add-default-api-server-{context}"]
+= Add an API server default certificate
+
+The default API server certificate is issued by an internal {product-title}
```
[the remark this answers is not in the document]
huffmanca: I've moved this precursor information into the assembly and reworded the first sentence.

modules/customize-certificates-api-add-default.adoc
```diff
@@ -0,0 +1,30 @@
...
+= Add an API server default certificate
+
+The default API server certificate is issued by an internal {product-title} 
+cluster CA. Clients outside of the cluster will not be able to verify the 
+API server's certificate by default. This certificate can be replaced 
+by one that is issued by a CA that clients trust.
+
+.Prerequisites
+
+* You must have a valid certificate and key.
```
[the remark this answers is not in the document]
huffmanca: I've updated this to include that it must be in the PEM format. It's any certificate clients already trust.

modules/customize-certificates-api-add-default.adoc
```diff
@@ -0,0 +1,30 @@
...
++
+----
+$ oc create secret tls <certificate> --cert=</path/to/cert.crt> --key=</path/to/cert.key> -n openshift-config
+----
+
+. Update the API server to reference the created secret.
++
+----
+$ oc patch apiserver cluster --type=merge -p "{\"spec\": {\"servingCerts\": {\"defaultServingCertificate\": {\"name\": \"<certificate>\"}}}}"
+----
```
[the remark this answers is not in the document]
huffmanca: I've included an extra step that examines the API server and shows the certificate.

authentication/certificates/api-server.adoc
```diff
@@ -0,0 +1,10 @@
+[id="api-server-certificates"]
+= Adding API server certificates
+include::modules/common-attributes.adoc[]
+:context: api-server-certificates
+
+toc::[]
+
+include::modules/customize-certificates-api-add-default.adoc[leveloffset=+1]
+
+include::modules/customize-certificates-api-add-named.adoc[leveloffset=+1]
```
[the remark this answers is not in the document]
huffmanca: I consider these part of the same user story - securing your API server for your environment. The first, replacing the default router, is going to be for any generic CA that clients already trust. Named certificates are going to be primarily used when external clients access the API server using a different hostname and there are different certificates trusted.

I've expanded the description of the second module to indicate when it should be used.

modules/customize-certificates-api-add-named.adoc
```diff
@@ -0,0 +1,35 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/api-server.adoc
+
+[id="add-named-api-server-{context}"]
+= Add an API server named certificate
+
+The default API server certificate is issued by an internal {product-title} 
+cluster CA. Additional certificates may be added to the API server to send
```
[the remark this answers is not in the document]
huffmanca: Corrected.

modules/customize-certificates-api-add-named.adoc
```diff
@@ -0,0 +1,35 @@
...
+.Procedure
+
+. Create a secret that contains the certificate and key in the 
+`openshift-config` namespace.
++
+----
+$ oc create secret tls <certificate> --cert=</path/to/cert.crt> --key=</path/to/cert.key> -n openshift-config
+----
+
+. Update the API server to reference the created secret. This must include
```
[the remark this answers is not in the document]
huffmanca: I've annotated this command.

modules/customize-certificates-understanding-service-serving.adoc
```diff
@@ -0,0 +1,23 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/service-serving-certificate.adoc
+
+[id="understanding-service-serving-{context}"]
+= Understanding service serving certificate secrets
+
+Service serving certificate secrets are intended to support complex 
+middleware applications that need out-of-the-box certificates. It has
```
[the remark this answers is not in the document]
huffmanca: Updated this section.

modules/customize-certificates-understanding-service-serving.adoc
```diff
@@ -0,0 +1,23 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/service-serving-certificate.adoc
+
+[id="understanding-service-serving-{context}"]
+= Understanding service serving certificate secrets
+
+Service serving certificate secrets are intended to support complex 
+middleware applications that need out-of-the-box certificates. It has 
+the same settings as the server certificates generated by the
```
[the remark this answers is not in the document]
huffmanca: Updated this section to use the recommendations.

modules/customize-certificates-understanding-service-serving.adoc
```diff
@@ -0,0 +1,23 @@
...
+
+[id="understanding-service-serving-{context}"]
+= Understanding service serving certificate secrets
+
+Service serving certificate secrets are intended to support complex 
+middleware applications that need out-of-the-box certificates. It has 
+the same settings as the server certificates generated by the 
+administrator tooling for nodes and masters.
+
+Once the cluster generates a secret for your service, your PodSpec can
```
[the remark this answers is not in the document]
huffmanca: Corrected.

modules/customize-certificates-understanding-service-serving.adoc
```diff
@@ -0,0 +1,23 @@
...
+mount it and the Pod will run as it becomes available. The generated
+certificate will be valid for the internal service DNS name
```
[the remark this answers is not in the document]
huffmanca: I've adjusted the first part.

I've clarified a bit on the DNS name to indicate that this only used for internal traffic. The point is that these certificates can't be used externally.

modules/customize-certificates-understanding-service-serving.adoc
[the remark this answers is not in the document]
huffmanca: I've reworked this content using the suggestions above. The expiration is now listed in the module, and the note only refers to the automatic upgrade and manual rotation.

modules/customize-certificates-understanding-service-serving.adoc
```diff
@@ -0,0 +1,23 @@
...
+[id="understanding-service-serving-{context}"]
+= Understanding service serving certificate secrets
+
+Service serving certificate secrets are intended to support complex 
+middleware applications that need out-of-the-box certificates. It has 
+the same settings as the server certificates generated by the 
+administrator tooling for nodes and masters.
+
+Once the cluster generates a secret for your service, your PodSpec can
+mount it and the Pod will run as it becomes available. The generated
```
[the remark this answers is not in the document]
huffmanca: I've added the comma. I'm leaving the second sentence as is for now.

modules/customize-certificates-add-service-serving.adoc
```diff
@@ -0,0 +1,27 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/service-serving-certificate.adoc
+
+[id="add-service-certificate-{context}"]
+= Add a service certificate
+
+To secure communication to your service, have the cluster generate a
```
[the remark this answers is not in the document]
huffmanca: Incorporated.

modules/customize-certificates-add-service-serving.adoc
```diff
@@ -0,0 +1,27 @@
...
+
+[id="add-service-certificate-{context}"]
+= Add a service certificate
+
+To secure communication to your service, have the cluster generate a 
+signed serving certificate and key pair into a secret in your namespace. To 
+do this, set the `service.beta.openshift.io/serving-cert-secret-name` 
+annotation on your service with the value set to the name you want to use 
+for your secret.
+
```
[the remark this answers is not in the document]
huffmanca: I've added the service as a prerequisite. I think we should have an overarching discussion regarding implied prerequisites at some point.

Once the service is defined, even if it contains nothing useful, you can proceed with the rest of the annotations.

modules/customize-certificates-add-service-serving.adoc
huffmanca: I've annotated this command, and clarified what each of these replaceables mean. I've also added a way of describing the service to view the annotations.

modules/customize-certificates-add-service-serving.adoc
huffmanca: Corrected.

modules/customize-certificates-add-service-serving.adoc
```diff
@@ -0,0 +1,27 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/service-serving-certificate.adoc
+
+[id="add-service-certificate-{context}"]
+= Add a service certificate
+
+To secure communication to your service, have the cluster generate a 
+signed serving certificate and key pair into a secret in your namespace. To
```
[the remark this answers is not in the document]
huffmanca: This sentence has been removed, and the command has been annotated.

modules/customize-certificates-understanding-service-serving.adoc
```diff
@@ -0,0 +1,23 @@
...
+certificate will be valid for the internal service DNS name 
+`<service.name>.<service.namespace>.svc`.
+
+The generated certificate and key are in PEM format, stored in `tls.crt` 
+and `tls.key` respectively, within the created secret. The 
+certificate and key pair is automatically replaced when it gets close to
```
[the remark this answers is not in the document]
huffmanca: This was intended to be a "certificate and key" "pair", meaning the "pair" is singular. I've removed the word "pair" to be consistent with the rest of the paragraph and made this plural.

modules/customize-certificates-manually-rotate-service-ca.adoc
huffmanca: Correct. The service CA certificate is automatically generated, and this section is only relevant if you use service certificates, as it's the service CA certificate that lasts for a year. I don't see any other prerequisites to complete this module.

modules/customize-certificates-manually-rotate-service-ca.adoc
huffmanca: I'd prefer to keep this in one assembly. You only need to perform this process if you utilize service certificates. These certificates are exclusively used for internal cluster communication.

The service CA certificate signs all of the service certificates, and isn't configured or touched by the customer at all. The only thing they really should know is that if they use service certificates then this **must** be rotated after a year **if and only if** they don't do an upgrade within a year. I strongly suspect less than 5% of customers will ever need to perform this, but it's something they should be aware of if they intend to use service certificates. Thus its inclusion here.

modules/customize-certificates-manually-rotate-service-ca.adoc
huffmanca: Updated to use the suggestion.

modules/customize-certificates-manually-rotate-service-ca.adoc
huffmanca: Corrected.

modules/customize-certificates-view-service-serving.adoc
huffmanca: I think you're right on this module. Originally I was thinking it was so you could view the expiration date of the certificate, but the only expiration date that matters is the service CA, which is handled in a separate module.

I'm deleting this file and removing the reference.

modules/customize-certificates-view-service-serving.adoc
huffmanca: Per above, I'm deleting this file and removing the reference.

modules/customize-certificates-add-service-serving-configmap.adoc
```diff
@@ -0,0 +1,49 @@
...
+. Annotate the ConfigMap with `service.beta.openshift.io/inject-cabundle=true`.
++
+----
+$ oc annotate configmap <configmap-name> service.beta.openshift.io/inject-cabundle=true
+----
++
+[NOTE]
+====
+Explicitly referencing the `service-ca.crt` key in a volumeMount will
```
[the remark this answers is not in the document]
huffmanca: If the ConfigMap's certificate is referenced by a Pod's volumeMount, then the Pod will not start until ConfigMap has the necessary certificate. I think it's important to note here so that customers know if they're performing these steps it may cause a delay in Pod creation/start. This should happen automatically and without issue to the end user.

authentication/certificates/service-serving-certificate.adoc
huffmanca: I've adjusted the title to be `Securing service traffic using service serving certificate secrets`. The goal here is to secure inter-cluster traffic between services.

While an argument could be made about this being two user stories, I think this is really one story - securing service traffic. If you secure service traffic, then you must be aware of the service CA certificate's expiration and how to rotate it if needed. As mentioned in another comment, I strongly suspect few customers will ever need those steps, but I don't believe it belongs in its own section.

I'm leaving this as one assembly for now.

modules/customize-certificates-replace-default-router.adoc
```diff
@@ -0,0 +1,30 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/replacing-default-ingress-certificate.adoc
+
+[id="replacing-default-ingress-{context}"]
+= Replacing the default ingress certificate
+
+You can replace the default ingress certificate for all
+applications under the `.apps` subdomain. This will update
```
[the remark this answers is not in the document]
huffmanca: I've reworded this section slightly.

modules/customize-certificates-api-add-named.adoc
```diff
@@ -0,0 +1,35 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/api-server.adoc
+
+[id="add-named-api-server-{context}"]
+= Add an API server named certificate
+
+The default API server certificate is issued by an internal {product-title} 
+cluster CA. Additional certificates may be added to the API server to send
+based on the client's requested URL.
```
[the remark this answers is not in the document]
huffmanca: The client could be any utility or program that access the OpenShift API; I'm leaving it as the generic version for now.

modules/customize-certificates-rotate-service-serving.adoc
huffmanca: I've included a bit more information about what happens when it occurs. I don't have a strong "why" as of now, but I can include this in a separate PR once it's available.

modules/customize-certificates-rotate-service-serving.adoc
huffmanca: I've included an annotation and the steps for determining the secret.

modules/customize-certificates-rotate-service-serving.adoc
huffmanca: I've included a sentence in the description of this command to indicate this is automatically recreated. This is also found in the description and in the following step as well.

modules/customize-certificates-add-service-serving-configmap.adoc
```diff
@@ -0,0 +1,48 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/service-serving-certificate.adoc
+
+[id="add-service-certificate-configmap-{context}"]
+= Add a service certificate to a ConfigMap
+
+To secure communication to services that use a ConfigMap, 
+have the cluster generate a signed serving certificate and key pair into the
```
mrogers950: This is incorrect. (the wording also implies that the service uses the configmap) 
What you are injecting into the configMap is just the CA certificate. The purpose for this is so that a pod seperate from a service can get access to the CA certificate, in order to do TLS against a Service that is using a service serving cert.

modules/customize-certificates-add-service-serving-configmap.adoc
```diff
@@ -0,0 +1,48 @@
...
+ConfigMap. To do this, set the 
+`service.beta.openshift.io/inject-cabundle=true` 
+annotation on your ConfigMap. A certificate will be generated, which allows
```
mrogers950: The CA is not generated - the configmap is injected with the current service CA certificate (it is generated elsewhere automatically, though this is not something to document).

modules/customize-certificates-add-service-serving-configmap.adoc
```diff
@@ -0,0 +1,48 @@
...
+. View the ConfigMap to ensure the certificate has been generated. This 
+appears as a `service-ca.crt` in the YAML output.
++
+----
+$ oc get configmap <configmap-name> -o yaml
+apiVersion: v1
+data:
+  service-ca.crt: |
+    -----BEGIN CERTIFICATE-----
+...
```
mrogers950: We need to mention that the configmap object that you use for this purpose should not be shared with other configmap data. IE, if you have one configmap that has all your pod's config items in multiple keys, don't annotate it for injection because the other data will be deleted. So in this case use a second configmap exclusively for the service CA bundle.

modules/customize-certificates-add-service-serving.adoc
mrogers950: "your namespace" -> "same namespace as the service"

modules/customize-certificates-understanding-service-serving.adoc
```diff
@@ -0,0 +1,23 @@
...
+//
+// * authentication/certificates/service-serving-certificate.adoc
+
+[id="understanding-service-serving-{context}"]
+= Understanding service serving certificate secrets
+
+Service serving certificate secrets are intended to support complex 
+middleware applications that need out-of-the-box certificates. It has 
+the same settings as the server certificates generated by the 
+administrator tooling for nodes and masters.
```
mrogers950: I'm not sure if this comment about administrator tooling still applies for 4.0. 

It would be better for this last sentence to say that the service serving certificates are issued as TLS Web server certificates.

modules/customize-certificates-replace-default-router.adoc
```diff
@@ -0,0 +1,46 @@
...
+
+You can replace the default ingress certificate for all
+applications under the `.apps` subdomain. After you replace
+the certificate, all applications, including the web console
+and CLI, will have encryption provided by specified certificate.
+
+.Prerequisites
+
+* You must have a wildcard certificate, in the PEM format, for use. 
+* The certificate must have a `subjectAltName` extension of `*.apps.`.
```
mrogers950: the subjectAltName needs to be for *apps.<clustername>.<domain>

modules/customize-certificates-replace-default-router.adoc
```diff
@@ -0,0 +1,46 @@
...
+= Replacing the default ingress certificate
+
+You can replace the default ingress certificate for all
+applications under the `.apps` subdomain. After you replace
+the certificate, all applications, including the web console
+and CLI, will have encryption provided by specified certificate.
+
+.Prerequisites
+
+* You must have a wildcard certificate, in the PEM format, for use.
```
mrogers950: wildcard certificate and its private key, both in the PEM format

modules/customize-certificates-replace-default-router.adoc
```diff
@@ -0,0 +1,46 @@
...
+. Create a secret that contains the wildcard certificate and key:
++
+----
+$ oc create secret tls <certificate> \//<1>
+     --cert=</path/to/cert.crt> \//<2>
+     --key=</path/to/cert.key> \//<3>
+     -n openshift-ingress
+----
+<1> `<certificate>` is the name of the secret that will contain
+the certificate.
```
mrogers950: contain the certificate and private key

modules/customize-certificates-understanding-default-router.adoc
```diff
@@ -0,0 +1,18 @@
...
+
+[id="understanding-default-ingress-{context}"]
+= Understanding the default ingress certificate
+
+By default {product-title} uses the Ingress Controller Operator to 
+create an internal CA and issue a wildcard certificate that is valid for 
+applications under the `.apps` sub-domain. Both the web console and CLI
+use this certificate as well. 
+
+The internal infrastructure certificates are self-signed.
```
mrogers950: Only the internal infrastructure *CA* certificates are self-signed.

modules/customize-certificates-understanding-default-router.adoc
```diff
@@ -0,0 +1,18 @@
...
+= Understanding the default ingress certificate
+
+By default {product-title} uses the Ingress Controller Operator to 
+create an internal CA and issue a wildcard certificate that is valid for 
+applications under the `.apps` sub-domain. Both the web console and CLI
+use this certificate as well. 
+
+The internal infrastructure certificates are self-signed. 
+While this process might be perceived as bad practice by some security or 
+PKI teams, any risk here is minimal. The only clients that trust these
```
mrogers950: "The only clients that *implicitly* trust these certificates... "

modules/customize-certificates-understanding-default-router.adoc
```diff
@@ -0,0 +1,18 @@
...
+create an internal CA and issue a wildcard certificate that is valid for 
+applications under the `.apps` sub-domain. Both the web console and CLI
+use this certificate as well. 
+
+The internal infrastructure certificates are self-signed. 
+While this process might be perceived as bad practice by some security or 
+PKI teams, any risk here is minimal. The only clients that trust these 
+certificates are other components within the cluster. If you replace 
+the default wildcard certificate, then all external users and systems
+will use the newly defined certificate.
```
mrogers950: Saying "external users and systems will *use*" is incorrect. It should say something more along the lines of:

"Replacing the default wildcard certificate with one that is issued by a public or organizational CA will allow external clients to connect securely to applications running under the apps sub-domain"

modules/customize-certificates-understanding-service-serving.adoc
```diff
@@ -0,0 +1,26 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/service-serving-certificate.adoc
+
+[id="understanding-service-serving-{context}"]
+= Understanding service serving certificate secrets
+
+Service serving certificate secrets are intended to support complex 
+middleware applications that require encryption. These certificates have 
+the same settings as the server certificates that are generated by the
```
mrogers950: I think my previous comment on this one got eaten when I refreshed the PR..
We should just say that the server certificates are issued as a TLS Web Server certificate.

 I don't think that the part about administrator tooling is relevant for 4.

modules/customize-certificates-add-service-serving-configmap.adoc
huffmanca: I've updated the introductory text to read:

> A Pod can obtain a reference to the service CA certificate, so that the Pod 
> can securely communicate with a service. To accomplish this communication, 
> have the cluster insert the service CA certificate into the ConfigMap by
> adding the  
> `service.beta.openshift.io/inject-cabundle=true` 
> annotation on your ConfigMap. 
> 
> The service CA certificate is stored in the ConfigMap as `service-ca.crt`.

modules/customize-certificates-add-service-serving-configmap.adoc
huffmanca: This has been corrected.

modules/customize-certificates-add-service-serving-configmap.adoc
huffmanca: I've added an important block below the introduction that says this.

modules/customize-certificates-add-service-serving.adoc
huffmanca: Updated to use the recommended verbiage.

modules/customize-certificates-understanding-service-serving.adoc
huffmanca: The last sentence has been reworded to just say:

> These certificates are issued as TLS web server certificates.

modules/customize-certificates-replace-default-router.adoc
huffmanca: Updated.

modules/customize-certificates-replace-default-router.adoc
huffmanca: Updated to use this verbiage.

modules/customize-certificates-replace-default-router.adoc
huffmanca: Updated to use this verbiage.

modules/customize-certificates-understanding-default-router.adoc
huffmanca: Included this detail.

modules/customize-certificates-understanding-default-router.adoc
huffmanca: Added.

modules/customize-certificates-understanding-default-router.adoc
huffmanca: Updated the last sentence to use this sentence.

modules/customize-certificates-understanding-service-serving.adoc
huffmanca: I don't think the comment was eaten! This feedback has been included.

modules/customize-certificates-api-add-default.adoc
```diff
@@ -0,0 +1,56 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/api-server.adoc
+
+[id="add-default-api-server-{context}"]
+= Add an API server default certificate
+
+To allow clients outside the cluster to validate the API server's
+certificate, you can replace the default certificate a CA
```
mrogers950: "To allow clients outside of the cluster to validate the API server's certificate, you can replace the default certificate with one that is issued by a public or organizational CA."

huffmanca: Updated to use this text.

modules/customize-certificates-replace-default-router.adoc
```diff
@@ -0,0 +1,48 @@
...
+<2> `</path/to/cert.crt>` is the path to the certificate on your
+local file system.
+<3> `</path/to/cert.key>` is the path to the private key associated
+with this certificate.
+
+. Update the ingress controller configuration with the newly created
+secret:
++
+----
+$ oc patch ingresscontroller.operator cluster \
```
lihongan: should be `oc patch ingresscontroller.operator default`, because the default ingresscontroller has been renamed from `cluster` to `default`.

huffmanca: This has been corrected.

huffmanca:
@lihongan ,

The change has been incorporated. Let me know if this one looks good to merge, or if there are any other issues to address.

lihongan:
@huffmanca thanks, the part of `Replacing the default ingress certificate` looks good to me now.
@stuartchuan @xingxingxia Could you please help review other two parts?

xingxingxia:
> @stuartchuan @xingxingxia Could you please help review other two parts?

@stuartchuan Hi, per https://bugzilla.redhat.com/show_bug.cgi?id=1685704#c27 and [Slack confirmation with Dev](https://coreos.slack.com/archives/CB48XQ4KZ/p1557226705404400?thread_ts=1557219567.402800&cid=CB48XQ4KZ), seems other parts are auth team related. Could you check when free? Let me know if objection.

stuartchuan:
The auth related parts are OK for me.

stlaz (commented):
I've managed to go through 2/3 chapters, the content seemed correct, I have a few general comments that could make the chapters a bit more easier to read.

I also had a brief look into the "Securing service traffic using service serving certificate secrets" chapter, it feels a bit more too heavy, also, I do think it should be more about "Securing service traffic using service serving certificates" rather than the secrets, which are but an implementation detail of this intent.

modules/customize-certificates-manually-rotate-service-ca.adoc
```diff
@@ -0,0 +1,59 @@
...
+----
+$ oc delete secret/signing-key -n openshift-service-ca
+----
+
+. To apply the new certificates to all services, restart all the Pods 
+in your cluster. This command ensures that all services use the 
+updated certificates.
++
+----
+$ for I in $(oc get ns -o jsonpath='{range .items[*]} {.metadata.name}{"\n"} {end}'); \
```
stlaz: Won't this by any chance irreversibly kill all the pods that are not associated to any replica controller/set?

modules/customize-certificates-api-add-named.adoc
```diff
@@ -0,0 +1,68 @@
...
+----
+$ oc patch apiserver cluster \
+     --type=merge -p \
+     "{\"spec\": {\"servingCerts\": {\"namedCertificates\": \
+     [{\"names\": [\"<hostname>\"], \"servingCertificate\": \//<1>
+     {\"name\": \"<certificate>\"}}]}}}" //<2>
+----
+<1> Replace `<hostname>` with the hostname that the API server
+should provide the certificate for.
+<1> Replace `<certificate>` with the name used for the secret in
```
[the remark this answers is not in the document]
huffmanca: Corrected.

modules/customize-certificates-api-add-named.adoc
```diff
@@ -0,0 +1,68 @@
+// Module included in the following assemblies:
+//
+// * authentication/certificates/api-server.adoc
+
+[id="add-named-api-server-{context}"]
+= Add an API server named certificate
+
+The default API server certificate is issued by an internal {product-title}
```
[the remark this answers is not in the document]
huffmanca: I've adjusted my editor to spot these issues and removed them from the files in this PR.

modules/customize-certificates-api-add-default.adoc
```diff
@@ -0,0 +1,56 @@
...
+----
+$ oc patch apiserver cluster \
+     --type=merge -p \
+     "{\"spec\": {\"servingCerts\": {\"defaultServingCertificate\": \
+     {\"name\": \"<certificate>\"}}}}" //<1>
+----
+<1> Replace `<certificate>` with the name used for the secret in
+the previous step.
+
+. Examine the `cluster` API server and confirm the secret is now
```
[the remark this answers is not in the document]
huffmanca: Corrected.

modules/customize-certificates-api-add-named.adoc
```diff
@@ -0,0 +1,68 @@
...
+local file system.
+<3> `</path/to/cert.key>` is the path to the private key associated
+with this certificate.
+
+. Update the API server to reference the created secret. 
++
+----
+$ oc patch apiserver cluster \
+     --type=merge -p \
+     "{\"spec\": {\"servingCerts\": {\"namedCertificates\": \
```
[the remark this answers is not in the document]
huffmanca: I've tried this with single quotations, and removing the backslashes, and get:

> Error from server: Invalid JSON Patch

I've tried editing this twice just to be certain there were no typos, and it seems to be invalid.

modules/customize-certificates-api-add-named.adoc
```diff
@@ -0,0 +1,68 @@
...
+     [{\"names\": [\"<hostname>\"], \"servingCertificate\": \//<1>
+     {\"name\": \"<certificate>\"}}]}}}" //<2>
+----
+<1> Replace `<hostname>` with the hostname that the API server
+should provide the certificate for.
+<1> Replace `<certificate>` with the name used for the secret in
+the previous step.
+
+. Examine the `cluster` API server and confirm the secret is now
```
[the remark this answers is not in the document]
huffmanca: I've updated this to use `apiserver/cluster`.

modules/customize-certificates-understanding-default-router.adoc
```diff
@@ -0,0 +1,19 @@
+// Module included in the following assemblies:
+//
+// authentication/certificates/replacing-default-ingress-certificate.adoc
+
+[id="understanding-default-ingress-{context}"]
+= Understanding the default ingress certificate
+
+By default {product-title} uses the Ingress Controller Operator to
```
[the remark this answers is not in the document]
huffmanca: There were discussions regarding this before I left, and it seems that the team has finalized on "Ingress Operator" (specifically when the operator is in use), or just "ingress" otherwise. I've fixed it here, and will open up a separate BZ to make these references consistent across the docs.

modules/customize-certificates-replace-default-router.adoc
```diff
@@ -0,0 +1,48 @@
...
+<3> `</path/to/cert.key>` is the path to the private key associated
+with this certificate.
+
+. Update the ingress controller configuration with the newly created
+secret:
++
+----
+$ oc patch ingresscontroller.operator default \
+     --type=merge -p \
+     "{\"spec\":{\"defaultCertificate\": \
```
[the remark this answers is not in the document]
huffmanca: Same as above - my tests using a single quotation here have resulted in an error.

modules/customize-certificates-add-service-serving-configmap.adoc
```diff
@@ -0,0 +1,56 @@
...
+A Pod can access the service CA certificate by mounting a ConfigMap that 
+is annotated with `service.beta.openshift.io/inject-cabundle=true`. 
+Once annotated, the cluster automatically injects the service CA 
+certificate into the `service-ca.crt` key on the ConfigMap. Access to 
+this CA certificate allows TLS clients to verify connections to 
+services using service serving certificates.
+
+[IMPORTANT]
+====
+After adding this annotation to a ConfigMap all existing data is deleted. It
```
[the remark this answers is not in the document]
huffmanca: Corrected.

authentication/certificates/service-serving-certificate.adoc
```diff
@@ -0,0 +1,24 @@
+[id="add-service-serving"]
+= Securing service traffic using service serving certificate secrets
+include::modules/common-attributes.adoc[]
+:context: service-serving-certificate
+
+toc::[]
+
+include::modules/customize-certificates-understanding-service-serving.adoc[leveloffset=+1]
+
+[IMPORTANT]
```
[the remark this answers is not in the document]
huffmanca: I would love for it to be there, but this is a dirty hack because we can't include xrefs inside of modules. I think a link to this section is required, and the only way to do it is to include it in the assembly instead of the module. :(

modules/customize-certificates-manually-rotate-service-ca.adoc
mrogers950: If it's not associated with a replica set it would otherwise still be backed by a deployment, though

modules/customize-certificates-replace-default-router.adoc
```diff
@@ -0,0 +1,48 @@
...
+     {\”name\”: \"<certificate>\"}}}" \//<1>
```
mrogers950: this \”name\” has the different double quotes that are not accepted if you copy and paste this line. I think my google doc had this and I fixed it later. they should all be " and not ”. I think this is why you saw the error. So try this for the conversion to single quotes:

``` '{"spec":{"defaultCertificate":  {"name": "<certificate>"}}}' ```

File: modules/customize-certificates-understanding-service-serving.adoc [hunk 83]
```diff
@@ -0,0 +1,25 @@
...
+
+[id="understanding-service-serving-{context}"]
+= Understanding service serving certificate secrets
+
+Service serving certificate secrets are intended to support complex 
+middleware applications that require encryption. These certificates are
+issued as TLS web server certificates.
+
+After the cluster generates a secret for your service, your PodSpec can
+mount it, and the Pod will run as it becomes available. The generated
```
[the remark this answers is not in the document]
huffmanca: Adjusted this to be `after`, instead of `as` or `once`. During peer reviews the word `once` has been discouraged.

authentication/certificates/service-serving-certificate.adoc
```diff
@@ -0,0 +1,24 @@
+[id="add-service-serving"]
+= Securing service traffic using service serving certificate secrets
```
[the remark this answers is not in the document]
huffmanca: I've removed `secret` from the chapter title and introductory string.

modules/customize-certificates-replace-default-router.adoc
huffmanca: That works! I'll update the sections listed above.

modules/customize-certificates-replace-default-router.adoc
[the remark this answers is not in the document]
huffmanca: This has been updated to use single quotations, using the command provided by Matt.

modules/customize-certificates-api-add-named.adoc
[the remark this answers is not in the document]
huffmanca: I've gotten this to work with single quotations; however, it cannot include backslashes as line separators in the patch string. For instance, the following works:

> oc patch apiserver cluster \
>     --type=merge -p \
>     '{"spec":{"servingCerts": {"namedCertificates":
>     [{"names": ["<hostname>"], 
>     "servingCertificate": {"name": "<certificate>"}}]}}}' 

This does not:

> oc patch apiserver cluster \
>     --type=merge -p \
>     '{"spec":{"servingCerts": {"namedCertificates": \
>     [{"names": ["<hostname>"], \
>     "servingCertificate": {"name": "<certificate>"}}]}}}' 

I've included the top section in the documentation. While it looks slightly different than other commands, users should be able to paste this in without error.

modules/customize-certificates-manually-rotate-service-ca.adoc
huffmanca: Resolving this conversation for now.

modules/customize-certificates-api-add-default.adoc
```diff
@@ -0,0 +1,56 @@
...
+local file system.
+<3> `</path/to/cert.key>` is the path to the private key associated
+with this certificate.
+
+. Update the API server to reference the created secret.
++
+----
+$ oc patch apiserver cluster \
+     --type=merge -p \
+     "{\"spec\": {\"servingCerts\": {\"defaultServingCertificate\": \
```
stlaz: Could you please also use single-quotes to wrap the string here as well?

modules/customize-certificates-api-add-named.adoc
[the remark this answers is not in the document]
stlaz: Thanks. When I do `git show`, my terminal shows those trailing whitespaces with bright red, which hurts my eyes :slightly_smiling_face:

modules/customize-certificates-manually-rotate-service-ca.adoc
stlaz: What I meant by "replicasets" were even those created by deployments internally, meaning - you can always run a pod by itself. It may not be all that useful, but there can always be some pods that run just like that, can't they?

authentication/certificates/service-serving-certificate.adoc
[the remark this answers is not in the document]
stlaz: That makes me a sad panda! Thanks for the explanation :slightly_smiling_face:

modules/customize-certificates-api-add-default.adoc
huffmanca: Thanks for pointing this instance out - I've changed this to use single quotations as well. Again, I haven't found a way to get this to work with the `\` between lines, but you can copy and paste the command into the terminal and it will work.

modules/customize-certificates-understanding-service-serving.adoc
```diff
@@ -0,0 +1,25 @@
...
+// * authentication/certificates/service-serving-certificate.adoc
+
+[id="understanding-service-serving-{context}"]
+= Understanding service serving certificate secrets
+
+Service serving certificate secrets are intended to support complex 
+middleware applications that require encryption. These certificates are
+issued as TLS web server certificates.
+
+After the cluster generates a secret for your service, your PodSpec can
```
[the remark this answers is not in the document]
huffmanca: I moved this paragraph to the following procedural section, and broke it into two pieces. Now the bit about the valid certificate namespaces is in the introduction, and the information about a PodSpec mounting it is the final step of the procedure.

huffmanca:
@stlaz ,

Thank you for the review! The feedback should be incorporated. Let me know if you have any further thoughts on this PR.

huffmanca:
I'm going to go ahead and merge this content as QE has validated it. If there are additional issues, then we can file BZs to get those addressed.

PR #14383, opened 2019-04-10
Adds ingress controller scaling

danehans:
Adds steps needed for scaling an ingress controller. This PR also renames the associated assembly filename and contents.

@openshift/sig-network-edge PTAL

### DIFF
--- a/modules/nw-scaling-ingress-controller.adoc
+++ b/modules/nw-scaling-ingress-controller.adoc
@@ -8,86 +8,33 @@
 An ingress controller can be manually scaled up or down based on
 requirements such as the need to increase throughput. `oc`
 commands are used to scale the `IngressController` resource.
+The following procedure provides an example for scaling up the
+default `IngressController`.
 
 .Procedure
-. View the current number of replicas of the ingress controller resource.
+. View the current number of available replicas for the default `IngressController`.
 +
 ----
-$ name="foo"  <1>
-$ oc get ingresscontroller/$name -n openshift-ingress-operator -o yaml
-apiVersion: operator.openshift.io/v1
-kind: IngressController
-metadata:
-  creationTimestamp: 2019-04-09T16:11:52Z
-  finalizers:
-  - ingress.openshift.io/ingress-controller
-  generation: 2
-  name: $name
-  namespace: openshift-ingress-operator
-  resourceVersion: "1187245"
-  selfLink: /apis/operator.openshift.io/v1/namespaces/openshift-ingress-operator/ingresscontrollers/$name
-  uid: 303a2234-5ae2-11e9-a3f2-0adb90d769be
-spec:
-  replicas: 2
-status:
-  availableReplicas: 2
-  conditions:
-  - lastTransitionTime: 2019-04-09T16:12:27Z
-    status: "True"
-    type: Available
-  domain: apps.example.com
-  endpointPublishingStrategy:
-    type: LoadBalancerService
-  selector: ingress.operator.openshift.io/ingress-controller-deployment=$name
+$ oc get -n openshift-ingress-operator ingresscontrollers/default -o jsonpath='{$.status.availableReplicas}'
+
 ----
-<1> Replace "foo" with the name of the ingress controller you would like to scale.
 
-The `replicas` field represents the desired number of pods, while `availableReplicas`
-represents the number of a running pods. If an ingress controller is created without
-specifying `replicas`, the field will not appear in the output and a default value
-of `2` will be used.
-
-. Scale the ingress controller to the desired number of replicas using
-the `oc patch` command.
+. Scale the default `IngressController` to the desired number of replicas using
+the `oc patch` command. The following example scales the default `IngressController`
+to 3 replicas.
 +
 ----
-$ count="3"  <2>
-$ oc patch -n openshift-ingress-operator ingresscontroller/$name \
---patch '{"spec":{"replicas": $count}}' --type=merge
+$ oc patch -n openshift-ingress-operator ingresscontroller/default --patch '{"spec":{"replicas": 3}}' --type=merge
 ----
-<2> Replace "3" with the number of desired ingress controller replicas.
 
-. Verify the ingress controller has been updated with the desired number of replicas.
+. Verify the default `IngressController` has been updated with the desired
+number of replicas.
 +
 ----
-$ oc get ingresscontroller/$name -n openshift-ingress-operator -o yaml
-apiVersion: operator.openshift.io/v1
-kind: IngressController
-metadata:
-  creationTimestamp: 2019-04-09T16:11:52Z
-  finalizers:
-  - ingress.openshift.io/ingress-controller
-  generation: 2
-  name: $name
-  namespace: openshift-ingress-operator
-  resourceVersion: "1187246"
-  selfLink: /apis/operator.openshift.io/v1/namespaces/openshift-ingress-operator/ingresscontrollers/$name
-  uid: 303a2234-5ae2-11e9-a3f2-0adb90d769be
-spec:
-  replicas: 3
-status:
-  availableReplicas: 3
-  conditions:
-  - lastTransitionTime: 2019-04-09T16:12:27Z
-    status: "True"
-    type: Available
-  domain: apps.example.com
-  endpointPublishingStrategy:
-    type: LoadBalancerService
-  selector: ingress.operator.openshift.io/ingress-controller-deployment=$name
+$ oc get -n openshift-ingress-operator ingresscontrollers/default -o jsonpath='{$.status.availableReplicas}'
 ----
 
 [NOTE]
 ====
-Scaling is not an immediate action as it takes time to create the desired number of replicas.
+Scaling is not an immediate action, as it takes time to create the desired number of replicas.
 ====
--- a/modules/nw-scaling-ingress-controller.adoc
+++ b/modules/nw-scaling-ingress-controller.adoc
@@ -1,0 +1,93 @@
+// Module filename: nw-scaling-ingress-controller.adoc
+// Module included in the following assemblies:
+// * networking/ingress-controller-configuration.adoc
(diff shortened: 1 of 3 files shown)


danehans:
cc @bmcelvee

modules/nw-scaling-ingress-controller.adoc
```diff
@@ -0,0 +1,93 @@
+// Module filename: nw-scaling-ingress-controller.adoc
+// Module included in the following assemblies:
+// * networking/ingress-controller-configuration.adoc
+
+[id="nw-ingress-controller-configuration-{context}"]
+= Scaling an ingress controller
+
+An ingress controller can be manually scaled up or down based on
```
ironcladlou: What's the convention for referring to API resources like this? For example:

* "ingress controller"
* `ingresscontroller`

modules/nw-scaling-ingress-controller.adoc
```diff
@@ -0,0 +1,93 @@
...
+requirements such as the need to increase throughput. `oc`
+commands are used to scale the `IngressController` resource.
```
ironcladlou: When referring to the resource name like this, should we use camel casing? All the REST API and CLI tooling seem to refer to resource names in lowercase.

modules/nw-scaling-ingress-controller.adoc
Miciah: I've been using ```an `IngressController` resource``` for referring to a concrete API object and "an ingress controller" to refer to an ingress controller in abstract.

modules/nw-scaling-ingress-controller.adoc
```diff
@@ -0,0 +1,93 @@
...
+  uid: 303a2234-5ae2-11e9-a3f2-0adb90d769be
+spec:
+  replicas: 3
+status:
+  availableReplicas: 3
+  conditions:
+  - lastTransitionTime: 2019-04-09T16:12:27Z
+    status: "True"
+    type: Available
+  domain: apps.example.com
```
Miciah: What domain should we use in examples? I see various domains used throughout the documentation, and it would be nice to be consistent.

modules/nw-scaling-ingress-controller.adoc
```diff
@@ -0,0 +1,93 @@
...
+
+An ingress controller can be manually scaled up or down based on
+requirements such as the need to increase throughput. `oc`
+commands are used to scale the `IngressController` resource.
+
+.Procedure
+. View the current number of replicas of the ingress controller resource.
++
+----
+$ name="foo"  <1>
```
ironcladlou: Would it be simpler to make the example in terms of the `default` ingresscontroller to avoid the local and references to the local?

modules/nw-scaling-ingress-controller.adoc
```diff
@@ -0,0 +1,93 @@
...
+specifying `replicas`, the field will not appear in the output and a default value
+of `2` will be used.
+
+. Scale the ingress controller to the desired number of replicas using
+the `oc patch` command.
++
+----
+$ count="3"  <2>
+$ oc patch -n openshift-ingress-operator ingresscontroller/$name \
+--patch '{"spec":{"replicas": $count}}' --type=merge
```
Miciah: `$count` will not be expanded within single quotes.

modules/nw-scaling-ingress-controller.adoc
```diff
@@ -0,0 +1,93 @@
...
+An ingress controller can be manually scaled up or down based on
+requirements such as the need to increase throughput. `oc`
+commands are used to scale the `IngressController` resource.
+
+.Procedure
+. View the current number of replicas of the ingress controller resource.
++
+----
+$ name="foo"  <1>
+$ oc get ingresscontroller/$name -n openshift-ingress-operator -o yaml
```
ironcladlou: Asking because I don't have the answer and I'm using this PR to inform my own work:

Is this level of detail useful? For example, just the relevant info could be obtained through:

```
$ oc get -n openshift-ingress-operator ingresscontrollers/default -o jsonpath='{$.status.availableReplicas}'
```

modules/nw-scaling-ingress-controller.adoc
```diff
@@ -0,0 +1,93 @@
...
+  endpointPublishingStrategy:
+    type: LoadBalancerService
+  selector: ingress.operator.openshift.io/ingress-controller-deployment=$name
+----
+<1> Replace "foo" with the name of the ingress controller you would like to scale.
+
+The `replicas` field represents the desired number of pods, while `availableReplicas`
+represents the number of a running pods. If an ingress controller is created without
+specifying `replicas`, the field will not appear in the output and a default value
+of `2` will be used.
```
ironcladlou: the description of availableReplicas is not quite precise... from the API docs:

> availableReplicas is number of observed available replicas according to the ingress controller deployment.

modules/nw-scaling-ingress-controller.adoc
```diff
@@ -0,0 +1,93 @@
...
+  conditions:
+  - lastTransitionTime: 2019-04-09T16:12:27Z
+    status: "True"
+    type: Available
+  domain: apps.example.com
+  endpointPublishingStrategy:
+    type: LoadBalancerService
+  selector: ingress.operator.openshift.io/ingress-controller-deployment=$name
+----
+<1> Replace "foo" with the name of the ingress controller you would like to scale.
```
ironcladlou: Unnecessary note if we talk about the `default`

modules/nw-scaling-ingress-controller.adoc
```diff
@@ -0,0 +1,93 @@
...
+
+. Scale the ingress controller to the desired number of replicas using
+the `oc patch` command.
++
+----
+$ count="3"  <2>
+$ oc patch -n openshift-ingress-operator ingresscontroller/$name \
+--patch '{"spec":{"replicas": $count}}' --type=merge
+----
+<2> Replace "3" with the number of desired ingress controller replicas.
```
ironcladlou: You probably see where I'm going with this, but an alternative would be:

```
$ oc patch -n openshift-ingress-operator ingresscontroller/default \
	--patch '{"spec":{"replicas": 3}}' --type=merge
```

Another advantage is that it's a copy/pastable one-liner

networking/ingress-controller-configuration.adoc
```diff
@@ -1,9 +1,9 @@
-// Assembly filename: configuring-routing.adoc
-// Explains Router configuration.
-[id="configuring-routing"]
-= Configuring routing
+// Assembly filename: ingress-controller-configuration.adoc
```
kalexand-rh: Please also update the filename in` _topic_map.yaml`. I think that will resolve the build failure.

modules/nw-scaling-ingress-controller.adoc
danehans: I believe upstream takes the same approach that @Miciah describes. I will update the doc accordingly.

modules/nw-scaling-ingress-controller.adoc
danehans: IMO we should use the API object `kind` value to be consistent with upstream. For example, [DaemonSet](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/) and [Deployment](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/).

@bmcelvee is there a docs standard we should follow?

modules/nw-scaling-ingress-controller.adoc
danehans: @bmcelvee is there a docs standard we should follow?

modules/nw-scaling-ingress-controller.adoc
danehans: Done

modules/nw-scaling-ingress-controller.adoc
danehans: I've simplified the doc and removed this content. IMO we can address the difference between the 2 fields in a separate doc and keep this focused on the tactical steps. Let me know if you disagree and I can reincorporate the content.

modules/nw-scaling-ingress-controller.adoc
danehans: Removed

modules/nw-scaling-ingress-controller.adoc
danehans: Thanks for the heads-up. Removed.

modules/nw-scaling-ingress-controller.adoc
danehans: Thanks, I agree. Removed.

networking/ingress-controller-configuration.adoc
danehans: Done. Thanks!

modules/nw-scaling-ingress-controller.adoc
```diff
@@ -0,0 +1,43 @@
...
+commands are used to scale the `IngressController` resource.
+The following procedure provides an example for scaling up the
+default `IngressController`.
+
+.Procedure
+. View the current number of available replicas for the default `IngressController`.
++
+----
+$ oc get -n openshift-ingress-operator ingresscontrollers/default \
+-o jsonpath='{$.status.availableReplicas}'
```
ironcladlou: Not suggesting a change here — just for my own reference going forward to be consistent — what's a good wrapping convention? e.g. wrap at 80 characters?

danehans: IMO we should wrap at 80 characters. With that said, code blocks seem to vary from one project to the next.

@bmcelvee I do not see a standard approach to line wrapping taken by other docs in the repo. Is there a docs standard we should follow?

ironcladlou:
Thanks!
/lgtm

danehans:
@ironcladlou thanks for the review.

bmcelvee:
LGTM - merging

PR #14384, opened 2019-04-10
Fix incorrect deployment syntax

### LINKED ISSUE #14372: [enterprise-3.11] Edit suggested in file dev_guide/dev_tutorials/openshift_pipeline.adoc
The Jenkinsfile -> deploy stage

### What needs fixing?
In the deploy section, the given example will not result in a deployment:
def rm = openshift.selector("dc", templateName).rollout()
Selector.rollout() returns a RolloutManager on which you can call .latest() to trigger a new deployment. You can find the docs here, although it is an image so I can't link directly to it.  It is the final section at the bottom of the page: https://github.com/openshift/jenkins-client-plugin

Please update the line to read:
def rm = openshift.selector("dc", templateName).rollout().latest()

kkoller:
Calling .rollout() on a selector returns a RolloutManager but does not actually trigger a deployment.  You must call .latest() on the RolloutManager. You can find the docs here, although it is an image so I can't link directly to it. It is the final section at the bottom of the page: https://github.com/openshift/jenkins-client-plugin

This commit fixes #14372

### DIFF
--- a/dev_guide/dev_tutorials/openshift_pipeline.adoc
+++ b/dev_guide/dev_tutorials/openshift_pipeline.adoc
@@ -171,7 +171,7 @@ pipeline {
         script {
             openshift.withCluster() {
                 openshift.withProject() {
-                  def rm = openshift.selector("dc", templateName).rollout()
+                  def rm = openshift.selector("dc", templateName).rollout().latest()
                   timeout(5) { <9>
                     openshift.selector("dc", templateName).related('pods').untilEach(1) {
                       return (it.object().status.phase == "Running")


kalexand-rh:
/ok-to-test

@kkoller, thank you! 

@vikram-redhat, do you have a suggestion for a technical reviewer?

This example is present back to 3.9, so it might be valid past 3.11.

vikram-redhat:
@zhouying7780 are you able to confirm this?

zhouying7780:
@kkoller, thank you!   LGTM.

PR #14437, opened 2019-04-12
In CNV 2.0 we no longer need to recreate the uploadproxy route

aglitke:
An issue has been fixed and the CNV operator creates the correct route during deployment so the user no longer needs to perform the mitigation steps outlined in the documentation.

This PR is split into two commits:
1. Import `modules/cnv_upload_virtctl.adoc` from master
2. Remove the no longer required text

### DIFF
--- a/modules/cnv_upload_virtctl.adoc
+++ b/modules/cnv_upload_virtctl.adoc
@@ -1,0 +1,62 @@
+// Module included in the following assemblies:
+//
+// * cnv_users_guide/cnv_users_guide.adoc
+
+[[upload-vmdisk-virtctl]]
+= Uploading a local disk image to a new PVC
+
+You can use `virtctl image-upload` to upload a virtual machine disk image from
+a client machine to your {product-title} cluster. This creates a PVC that can be
+associated with a virtual machine after the upload has completed.
+
+.Prerequisites
+
+* A virtual machine disk image, in RAW or QCOW2 format. It can be compressed
+using *xz* or *gzip*.
+* *kubevirt-virtctl* must be installed on the client machine.
+* The client machine must be xref:../install_config/router/default_haproxy_router.html#overview[configured]
+to trust the OpenShift router's certificate.
+
+.Procedure
+
+. Identify the following items:
+* File location of the VM disk image you want to upload
+* Name and size desired for the resulting PVC
+
+. Remove the existing passthrough route:
++
+----
+$ oc delete route -n cdi cdi-uploadproxy-route
+----
+
+. Create a secured route using re-encryption termination:
++
+----
+$ oc get secret -n cdi cdi-upload-proxy-ca-key -o=jsonpath="{.data['tls\.crt']}" | base64 -d > ca.pem
+
+$ oc create route reencrypt cdi-uploadproxy-route -n cdi --service=cdi-uploadproxy --dest-ca-cert=ca.pem
+----
+
+. Use the `virtctl image-upload` command to upload your VM image,
+making sure to include your chosen parameters. For example:
++
+----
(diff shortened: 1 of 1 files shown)


aglitke:
@ousleyp Here is the new PR against the proper branch,

ousleyp:
@aglitke Thanks for doing this :) The change LGTM. We do need QE to sign off on it before merging. I think that Ying Cui is the right person to tag, but I can't seem to tag her here. Thoughts?

ousleyp:
@nellyc I want to get this merged, but I haven't heard from QE yet. Is someone available to validate the steps of this procedure? 

(After we merge this, the CNV docs team will create a PR to modularize this content, so there is no relevant preview build.)

Thanks!

modules/cnv_upload_virtctl.adoc
```diff
@@ -0,0 +1,48 @@
...
+
+. Identify the following items:
+* File location of the VM disk image you want to upload
+* Name and size desired for the resulting PVC
+
+. Use the `virtctl image-upload` command to upload your VM image,
+making sure to include your chosen parameters. For example:
++
+----
+$ virtctl image-upload --uploadproxy-url=https://$(oc get route cdi-uploadproxy-route -n cdi -o=jsonpath='{.status.ingress[0].host}') --pvc-name=upload-fedora-pvc --pvc-size=10Gi --image-path=/images/fedora28.qcow2
```
codificat: In an environment where CDI has been deployed through the HCO I see two differences:

- The upload route is in the `kubevirt-hyperconverged` namespace, not in the `cdi` namespace
- The route name is `cdi-uploadproxy`, not `cdi-uploadproxy-route`

So this command as-is wouldn't work on an environment deployed through HCO.

I believe that the idea in that case would be that image-upload would discover that through a `CDIConfig` —which isn't there in the env where I tested, but that's a different story— but if that's the case it should be noted.

Also, for readability: the command does get quite long. Maybe split it in multiple lines? And maybe separate the route discovery by creating a var and using it in the `virtctl` invocation?

modules/cnv_upload_virtctl.adoc
```diff
@@ -0,0 +1,48 @@
...
+----
++
+[CAUTION]
+====
+To allow insecure server connections when using HTTPS, use the `--insecure`
+parameter. Be aware that when you use the `--insecure` flag, the authenticity of
+the upload endpoint is *not* verified.
+====
+
```
codificat: Maybe mention additional/optional options? Like `--access-mode`... or maybe `-h` to get help?

codificat (changes requested):
I believe that some changes are needed so that this works on environments deployed through the HCO

modules/cnv_upload_virtctl.adoc
codificat: Oh, and another comment about the route: finding it (the `oc get route...` command) requires certain privileges, i.e. your user must be able to get routes in the cdi/hco namespace. Maybe this permission should be granted out of the box, but in my test env it wasn't. So it's either a bug or this should probably be documented here.

aglitke: We have automatic ingress detection so we can simplify this dramatically by removing the --uploadproxy-url parameter completely.

modules/cnv_upload_virtctl.adoc
aglitke: I don't want to overly broaden the scope of this PR.  --help is a standard argument so I think we can expect users to know it.  As for documenting --access-mode and others, we would need to open a new PR and this late in the game it will probably need to wait.

ousleyp:
My PR that incorporated @aglitke's patches has been merged: https://github.com/openshift/openshift-docs/pull/15505

I think this one can be closed.

PR #14446, opened 2019-04-12
BZ-1699185: Removed reference of deleting default storageclass.

huffmanca:
BZ-1699185: Removed reference of deleting default storageclass.

This is for OS 4.x.


chao007:
LGTM, thanks for testing and updating the doc.

huffmanca:
@chao007,

This has been updated to remove the references to GCE.

PR #14447, opened 2019-04-13
BZ1694774: Included module to define a cluster admin.

huffmanca:
Included a module in the RBAC section that includes an example of defining a cluster admin.

This is for OS 4.x.

### DIFF
--- a/modules/rbac-creating-cluster-admin.adoc
+++ b/modules/rbac-creating-cluster-admin.adoc
@@ -1,0 +1,22 @@
+// Module included in the following assemblies:
+//
+// * authentication/using-rbac.adoc
+
+[id="creating-cluster-admin-{context}"]
+= Creating a cluster admin
+
+The `cluster-admin` role is required to perform administrator
+level tasks on the {product-title} cluster, such as modifying
+cluster resources. 
+
+.Prerequisites 
+
+* You must have created a user to define as the cluster admin.
+
+.Procedure
+
+* Define the user as a cluster admin:
++
+----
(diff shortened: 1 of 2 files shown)


modules/rbac-creating-cluster-admin.adoc
```diff
@@ -0,0 +1,23 @@
+// Module included in the following assemblies:
+//
+// * authentication/using-rbac.adoc
+
+ifdef::openshift-enterprise,openshift-origin[]
```
bergerhoffer: Why is this line here? There's not a matching ending one. And if it is required, I'd think we'd want to put the conditional in the assembly? wdyt?

modules/rbac-creating-cluster-admin.adoc
```diff
@@ -0,0 +1,23 @@
...
+[id="creating-cluster-admin-{context}"]
+= Example of creating a cluster admin
```
bergerhoffer: Is there a reason that this title starts with "Example"? Could it just be "Creating a cluster admin"?

huffmanca: You're right - this flows better as starting with `Creating`. Incorporated this change.

modules/rbac-creating-cluster-admin.adoc
huffmanca: It was an oversight from the base file that I used. It seems that conditionals are split between modules and assemblies, and one of the modules I started this from included it.

I've removed the conditional from the module. The other sections on defining user- and cluster- roles don't have a similar conditional, so I believe it's safe to keep out of the assembly.

PR #14457, opened 2019-04-15
Updated the Invalid formatting section

mburke5678:


mburke5678:
@vikram-redhat PTAL

vikram-redhat:
@mburke5678 I need to confirm if this works on the customer portal.

PR #14461, opened 2019-04-16
adding Ignition config to the glossary

kalexand-rh:
@cnegus, @vikram-redhat, will you PTAL?


kalexand-rh:
Chris sent me some feedback through IRC, and I've updated the PR.

PR #14477, opened 2019-04-16
BZ-1699270: Updating tech preview feature notice for Block Volumes

huffmanca:
BZ-1699270: Updating tech preview feature notice for Block Volumes.

Note that due to our modularization guidelines we cannot link to the section on enabling TP features. Instead, I have included the module in the assembly. From discussions with other writers this is the encouraged method of reusing content.

This is for OS 4.x.


weherdh:
LGTM, thanks.

PR #14481, opened 2019-04-16
Migrate automatic egress IPs - OSDOCS-380

jboxman:
- OSDOCS-380

### DIFF
--- a/modules/nw-about-automatic-egress-ips.adoc
+++ b/modules/nw-about-automatic-egress-ips.adoc
@@ -1,0 +1,27 @@
+// Module included in the following assemblies:
+//
+// * networking/enabling-automatic-egress-ips.adoc
+
+[id="about-automatic-egress-ips-{context}"]
+= About automatic egress IPs
+
+You can assign egress IP addresses to namespaces by setting the `egressIPs`
+field of the `NetNamespace` resource. With fully automatic egress IPs, you can
+set the `egressCIDRs` field of each node's `HostSubnet` resource to indicate the
+range of egress IP addresses that can be hosted. Namespaces that have requested
+egress IP addresses are matched with nodes that are able to host those egress IP
+addresses, then the egress IP addresses are assigned to those nodes.
+
+High availability is automatic. If a node hosting egress IP addresses is
+unreachable and there are nodes that are able to host those egress IP addresses,
+based on the `egressCIDR` values of the `HostSubnet` resources, then the egress
+IP addresses will move to a new node. When the original egress IP address node
+comes back online, the egress IP addresses automatically move to balance egress
+IP addresses across nodes.
+
+[IMPORTANT]
+====
+You cannot use manually assigned and automatically assigned egress IP addresses
+on the same nodes or with the same IP address ranges.
+====
+
--- a/modules/nw-enabling-automatic-egress-ips.adoc
+++ b/modules/nw-enabling-automatic-egress-ips.adoc
@@ -1,0 +1,64 @@
+// Module included in the following assemblies:
+//
+// * networking/enabling-automatic-egress-ips.adoc
+
+[id="enabling-automatic-egress-ips-{context}"]
+= Enabling automatic egress IPs for a namespace
+
+In {product-title} you can enable automatic egress IPs for a specific namespace
+across one or more nodes.
+
+.Prerequisites
+
+* Install the OpenShift Command-line Interface (CLI), commonly known as `oc`.
+* You must be logged in to the cluster with the `cluster-admin` role.
+
+.Procedure
+
+. Update the `NetNamespace` with the egress IP address using the following JSON:
++
+----
+ $ oc patch netnamespace <project_name> \ <1>
+  -p '{
+    "egressIPs": [
+      "<ip_address>" <2>
+      ]
+    }'
+----
+<1> Specify the name of the project.
+<2> Specify an egress IP address.
++
+For example, to assign `project1` to an IP address of 192.168.1.100 and
+`project2` to an IP address of 192.168.1.101:
++
+----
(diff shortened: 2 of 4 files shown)


jboxman:
@openshift/team-documentation, PTAL. Thanks.

modules/nw-about-automatic-egress-ips.adoc
```diff
@@ -0,0 +1,27 @@
...
+
+You can assign egress IP addresses to namespaces by setting the `egressIPs`
+field to the `NetNamespace` resource. With fully automatic egress IPs, you can
+set the `egressCIDRs` field of each node's `HostSubnet` resource to indicate the
+range of egress IP addresses that can be hosted. Namespaces that have requested
+egress IP addresses are matched with nodes that are able to host those egress IP
+addresses, then the egress IP addresses are assigned to those nodes.
+
+High availability is automatic. If a node hosting egress IP addresses
+goes down and there are nodes that are able to host those egress IP addresses,
```
ahardin-rh: I'm wondering if we can use alternate wording for "goes down" here, as I'm not sure it will translate well.

modules/nw-enabling-automatic-egress-ips.adoc
```diff
@@ -0,0 +1,56 @@
...
+[id="enabling-automatic-egress-ips-{context}"]
+= Enabling automatic egress IPs for a namespace
+
+In {product-title} you can enable automatic egress IPs for a specific namespace
+across one or more nodes.
+
+.Prerequisites
+
+* The `oc` CLI program must be installed.
+* Logged in to the cluster with the `cluster-admin` role.
```
ahardin-rh: s/Logged in/You must be logged in

ahardin-rh:
@jboxman Just some minor comments from me. Otherwise, looks good! :rocket:

jboxman:
@ahardin-rh thank you!

jboxman:
Hi @squeed @pecameron @danwinship 

Does `egressIPs` take only one IP address or are multiple IPs valid?

Thanks.

modules/nw-about-automatic-egress-ips.adoc
```diff
@@ -0,0 +1,27 @@
+// Module included in the following assemblies:
+//
+// * networking/enabling-automatic-egress-ips.adoc
+
+[id="about-automatic-egress-ips-{context}"]
+= About automatic egress IPs
+
+You can assign egress IP addresses to namespaces by setting the `egressIPs`
+field to the `NetNamespace` resource. With fully automatic egress IPs, you can
```
danwinship: "the `egress IPs` field **of** the `NetNamespace` resource". Or maybe "on".

danwinship: I don't know how the different pieces of the 4.1 docs fit together, but it seems like you're mentioning "fully-automatic egress IPs" here without having explained what that means. (It's possible the 3.x docs were bad about this as well.)

(There are "manually-assigned" (or "semi-automatic") egress IPs, where the administrator explicitly picks which egress IPs will be on which nodes, and so has to set `egressIPs` on both the `NetNamespace` and the `HostSubnet` objects. And then there are "automatically-assigned" or "fully-automatic" egress IPs, where the administrator sets `egressIPs` on `NetNamespace` but sets `egressCIDRs` on the `HostSubnet`, and then OpenShift figures out the assignment of specific egress IPs to nodes.)

The manual/semi-automatic mode is older and less useful, so it makes sense for the docs to primarily focus on the fully-automatic mode, but we need to document the semi-automatic mode too.

danwinship:
> Does `egressIPs` take only one IP address or are multiple IPs valid?

In fully-automatic mode, you can only assign a single `egressIP` value to a given `NetNamespace`.

When using manually-assigned egress IPs, you can do high-availability by assigning multiple `egressIP` values to a namespace, and then assigning each of those IPs to a different node. So if you did:

    oc patch netnamespace project1 -p '{"egressIPs": ["192.168.1.100", "192.168.1.101"]}'
    oc patch hostsubnet node1 -p '{"egressIPs": ["192.168.1.100"]}'
    oc patch hostsubnet node2 -p '{"egressIPs": ["192.168.1.101"]}'

Then `project1` would use the IP `192.168.1.100` on `node1` by default, but if `node1` went down, then it would switch to using the IP `192.168.1.101` on `node2`.

modules/nw-about-automatic-egress-ips.adoc
jboxman: I have a separate PR for the manually assigned approach (https://github.com/openshift/openshift-docs/pull/14555). It may make more sense for the content for both PRs to be in a single assembly, with the fully automatic content followed by the manually assigned content.

All of this was derived from the monolithic network configuration page for OCP 3.x.

jboxman:
@danwinship, the manually assigned approach is covered in https://github.com/openshift/openshift-docs/pull/14555.

modules/nw-about-automatic-egress-ips.adoc
jboxman: @danwinship, what is the `NetNamespace` and what is the `HostSubnet`? It seems like a namespace itself may be synonymous with a _project_, so is the `NetNamespace` related to the network configuration for a project? And the `HostSubnet` relates to nodes available in the cluster?

Thanks.

danwinship: Yeah, `NetNamespace` is the object that OpenShift SDN uses to store its own per-namespace data, and `HostSubnet` is the object it uses to store its own per-node data. So the `NetNamespace` will have the same name as the namespace/project you're using an egress IP for, and the `HostSubnet` will have the same name as the node you're assigning the egress IP to.

jboxman:
I'm going to cherry-pick this into https://github.com/openshift/openshift-docs/pull/14555 and create the unified content there.

PR #14515, opened 2019-04-17
osdocs-168 Ingress Operator

bmcelvee:
https://jira.coreos.com/browse/OSDOCS-168

### DIFF
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -310,13 +310,17 @@ Topics:
   File: understanding-networking
 - Name: Multiple networks
   File: managing_multinetworking
-- Name: Understanding the DNS Operator in OpenShift Container Platform
-  File: dns-operator
-- Name: Ingress controller configuration
-  File: ingress-controller-configuration
 - Name: Using multicast
   File: using-multicast
-  Distros: openshift-origin, openshift-enterprise
+- Name: Understanding the DNS Operator in OpenShift Container Platform
+  File: dns-operator
+- Name: Ingress
+  Dir: ingress
+  Topics:
+  - Name: Ingress Operator in OpenShift Container Platform
+    File: configuring-ingress-operator
+  - Name: Ingress controller configuration
+    File: ingress-controller-configuration
 ---
 Name: Registry
 Dir: registry
--- a/modules/nw-creating-secured-routes.adoc
+++ b/modules/nw-creating-secured-routes.adoc
@@ -5,12 +5,10 @@
 [id="nw-creating-secure-routes-{context}"]
 = Creating secured edge terminated routes
 
-You can create a secure edge terminated route where TLS
-termination occurs on the router before the traffic is proxied to the
-destination.
-The secure edge terminated route specifies the TLS certificate
-and key information.
-The TLS certificate is served by the router front end.
+You can create a secure edge terminated route where TLS termination occurs on
+the ingress controller before the traffic is proxied to the destination. The
+secure edge terminated route specifies the TLS certificate and key information.
+The TLS certificate is served by the ingress controller front end.
 
 .Procedure
 . Create a private key for the edge secured route.
@@ -82,8 +80,8 @@ spec:
       -----END CERTIFICATE-----
 ----
 
-. Verify that the DNS entry for `www.example.test` points to the router
-instance(s) and that the route to your domain is available.
+. Verify that the DNS entry for `www.example.test` points to the ingress
+controller endpoint and that the route to your domain is available.
 +
 The example below uses curl and a local resolver to simulate the
 DNS lookup:
@@ -92,4 +90,5 @@ DNS lookup:
 $ routerip="4.1.1.1"  <1>
 $ curl -k --resolve www.example.test:443:$routerip https://www.example.test/
 ----
-<1> Replace the IP address with the IP address of one of your router instances.
+<1> Replace the IP address with the IP address or addresses at which the ingress
+controller endpoint is published.
--- a/modules/nw-ingress-controller-status.adoc
+++ b/modules/nw-ingress-controller-status.adoc
@@ -1,0 +1,16 @@
+// Module included in the following assemblies:
+//
+// * ingress/configure-ingress-operator.adoc
+
+[id="nw-ingress-controller-status-{context}"]
+= View ingress controller status
+
+Your can view the status of a particular ingress controller.
+
+.Procedure
+
+* View the status of an ingress controller:
++
+----
+$ oc describe --namespace=openshift-ingress-operator ingresscontroller/<name>
+----
--- a/modules/nw-ingress-operator-logs.adoc
+++ b/modules/nw-ingress-operator-logs.adoc
@@ -1,0 +1,16 @@
+// Module included in the following assemblies:
+//
+// * ingress/configure-ingress-operator.adoc
+
+[id="nw-ingress-operator-logs-{context}"]
+= View ingress controller logs
+
+You can view your ingress controller logs.
+
+.Procedure
+
+* View your ingress controller logs:
++
+----
+$ oc logs --namespace=openshift-ingress-operator deployments/ingress-operator
+----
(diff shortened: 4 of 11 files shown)


bmcelvee:
@ironcladlou, @Miciah, and @danehans - With this PR, I'm putting some framework in place for the operator docs (from the Ingress Operator repo), put the Ingress folder as a sub-folder in Networking (rearraged the files too), and replaced the term "router" with "ingress controller". 

Does that sound okay to you all? Thanks!

bmcelvee:
Are the procedures included here the same as for https://jira.coreos.com/browse/OSDOCS-355? If so, I'll bump them to troubleshooting. If not, could you point me in the right direction?

bmcelvee:

bmcelvee:

_topic_map.yml
```diff
@@ -305,8 +305,13 @@ Topics:
   File: understanding-networking
 - Name: Multiple networks
   File: managing_multinetworking
-- Name: Ingress controller configuration
-  File: ingress-controller-configuration
+- Name: Ingress
+  Dir: ingress
+  Topics:
+  - Name: Ingress Operator in OpenShift Container Platform
+    File: configuring-ingress-operator
```
danehans: Why is the same filename referenced by both topic names?

modules/nw-creating-secured-routes.adoc
```diff
@@ -82,8 +80,8 @@ spec:
       -----END CERTIFICATE-----
 ----
 
-. Verify that the DNS entry for `www.example.test` points to the router
-instance(s) and that the route to your domain is available.
+. Verify that the DNS entry for `www.example.test` points to the ingress
+controller instance(s) and that the route to your domain is available.
```
danehans: @openshift/sig-network-edge Instead of `ingress controller instance(s)`, should this be the ingress controller service ip?

modules/nw-creating-secured-routes.adoc
```diff
@@ -92,4 +90,5 @@ DNS lookup:
 $ routerip="4.1.1.1"  <1>
 $ curl -k --resolve www.example.test:443:$routerip https://www.example.test/
 ----
-<1> Replace the IP address with the IP address of one of your router instances.
+<1> Replace the IP address with the IP address of one of your ingress controller
+instances.
```
danehans: Same question as above.

modules/nw-creating-secured-routes.adoc
Miciah: If this document applies generally irrespective of endpoint publishing strategy type, we cannot refer to the service.  On a cloud environment, it will be the `LoadBalancer`-type `Service`, but on UPI it will likely be an external load-balancer configured by the user.  How about "ingress controller endpoint"?

modules/nw-ingress-operator-logs.adoc
```diff
@@ -0,0 +1,16 @@
+// Module included in the following assemblies:
+//
+// * ingress/configure-ingress-operator.adoc
+
+[id="nw-ingress-operator-logs-{context}"]
+= View Ingress Operator logs
+
+You can view your Ingress Operator logs.
```
Miciah: Using "your" for the operator feels odd: the platform deploys and manages an instance of the operator.  This stands in contrast to ingress controllers, where  the user can create one or more `IngressController` resources, so "your ingress controller" feels less odd.  I realize this is pretty subjective, so feel free to disregard.

networking/ingress/configuring-ingress-operator.adoc
```diff
@@ -0,0 +1,35 @@
...
+= Ingress Operator in {product-title}
+include::modules/common-attributes.adoc[]
+:context: configuring-ingress
+
+toc::[]
+
+The Ingress Operator is an OpenShift component that enables external access to
+cluster services by configuring Ingress Controllers. You can use the Ingress
+Operator to route traffic by specifying OpenShift Route and Kubernetes Ingress
+resources.
```
Miciah: Should that be ```OpenShift `Route` and Kubernetes `Ingress` resources``` since it is referring to API resource kinds?

modules/nw-ingress-controller-status.adoc
```diff
@@ -0,0 +1,16 @@
+// Module included in the following assemblies:
+//
+// * ingress/configure-ingress-operator.adoc
+
+[id="nw-ingress-controller-status-{context}"]
+= View Ingress controller status
+
+Your can view the status of a particular Ingress controller.
```
Miciah: The capitalization is odd—is there a reason to capitalize "Ingress" here?

_topic_map.yml
bmcelvee: Wonky rebase. Fixing, thanks!

modules/nw-creating-secured-routes.adoc
bmcelvee: Thanks! Updating

modules/nw-creating-secured-routes.adoc
bmcelvee: Also updating

modules/nw-ingress-controller-status.adoc
bmcelvee: No, probably just got in the habit since we're capitalizing it for the operator. Fixing, thanks!

modules/nw-ingress-operator-logs.adoc
bmcelvee: That makes sense, thanks!

networking/ingress/configuring-ingress-operator.adoc
bmcelvee: Yep - thanks!

modules/nw-ingress-view.adoc
```diff
@@ -0,0 +1,21 @@
+// Module included in the following assemblies:
+//
+// * ingress/configure-ingress-operator.adoc
+
+[id="nw-ingress-view-{context}"]
+= View the default ingress controller
+
+The Ingress Operator is a core feature of OpenShift and is enabled out of the
```
huffmanca: Should `OpenShift` be {product-title} here?

modules/nw-ingress-view.adoc
```diff
@@ -0,0 +1,21 @@
...
+box.
+
+Every new OpenShift installation has an `ingresscontroller` named default. It
```
huffmanca: Same question as above regarding `OpenShift` here.

huffmanca (commented):
I think we need some more conceptual information; however, it doesn't have to be included as part of this PR. I have just a couple of trivial questions regarding the current content.

Thanks for pulling this together!

modules/nw-ingress-view.adoc
bmcelvee: It should, will fix - thanks!

_topic_map.yml
ironcladlou: This is referencing a file at `networking/configuring-ingress-operator.adoc` but the actual files are nested under `networking/ingress` which seems to break all the includes and cause the topic to be excluded from the build.

jboxman: Does the subdirectory include symlinks? For example, `installing/installing_aws/` symlinks to `../modules` and `../images`:

```
s -1 installing/installing_aws/
modules
images
installing-aws-account.adoc
installing-aws-customizations.adoc
installing-aws-default.adoc
uninstalling-cluster-aws.adoc
```

ironcladlou: No. Offtopic, but we really should find a way to kill the symlink pattern. It's actively hostile to tooling which would greatly improve the experience of iterating on docs.

bmcelvee: Thanks for pointing that out, we can look into the symlink pattern for sure. For now the file structure is set up correctly for asciibinder (at least I'm able to build it successfully). I'm going to go ahead and merge the PR to get it in for review, but we can and certainly will continue to iterate on it.

bmcelvee:
Merging PR to get it in for review. It does need more conceptual material, which we will continue to iterate on. Thanks!

PR #14523, opened 2019-04-18
osdocs324 - draft of vSphere UPI

kalexand-rh:
Because this assembly uses some of the bare metal UPI modules, I rebased the branch for osdocs325 onto this branch. I expect to merge the bare metal PR first and fix the commits on this branch.

https://jira.coreos.com/browse/OSDOCS-324

Built version here: http://file.rdu.redhat.com/kalexand/0418/osdocs324/installing/installing_vsphere/installing-vsphere.html


kalexand-rh:
@staebler, will you PTAL at the draft doc?

modules/installation-requirements-upi.adoc
```diff
@@ -0,0 +1,74 @@
...
+install the cluster.
+====
+
+The bootstrap and control plane machines must use {op-system-first} as the operating system.
+
+[id="network-connectivity-{context}"]
+== Network connectivity requirements
+
+All the {op-system-first} machines require network in `initramfs` during boot
+to fetch Ignition config from the Machine Config Server. During the initia
```
staebler: `initial` instead of `initia`

modules/installation-vsphere-config-yaml.adoc
```diff
@@ -0,0 +1,48 @@
...
+
+You can customize the `install-config.yaml` file to specify more details about
+your {product-title} cluster's platform or modify the values of the required
+parameters.
+
+[source,yaml]
+----
+apiVersion: v1beta4
+baseDomain: example.com <1>
+compute:
```
staebler: The `compute` field is neither used not needed for a vSphere install.

modules/installation-vsphere-config-yaml.adoc
```diff
@@ -0,0 +1,48 @@
...
+<4> The fully-qualified host name or IP address of the vCenter server.
+<5> The name of the user for accessing the server. This user must have at least
+the roles and privileges that are required for
+link:https://vmware.github.io/vsphere-storage-for-kubernetes/documentation/vcp-roles.html[dynamic persistent volume provisioning]
+in vSphere.
+<6> The password associated with the vSphere user.
+<7> The vSphere datacenter.
+<8> The default vSphere datastore to use.
+<9> The pull secret that provides components in the cluster access to images for {product-title} components.
+<10> The default SSH key for the `core` user.
```
staebler: It is not clear to me from reading this what the core user is. Is there documentation elsewhere that describes that each OpenShift machine has a core user that uses this SSH key? Also, why is it the "default" SSH key instead of "the" SSH key?

modules/installation-vsphere-config-yaml.adoc
```diff
@@ -0,0 +1,48 @@
+// Module included in the following assemblies:
```
staebler: This is showing up in the doc as
```
/ Module included in the following assemblies:
```

modules/installation-vsphere-machines.adoc
```diff
@@ -0,0 +1,135 @@
...
+.Procedure
+
+////
+This will work for GA.
+. Obtain the {op-system} images from the
+link:https://access.redhat.com/downloads[Product Downloads] page on the Red
+Hat customer portal.
+////
+
+. Convert the bootstrap, master, and worker Ingintion conf files to Base64
```
staebler: Replace `conf` with `config`.

modules/installation-vsphere-machines.adoc
```diff
@@ -0,0 +1,135 @@
...
+encoding.
+
+. Upload the Base64-encoded bootstrap Ignition config file that the
+installation program created to your HTTP server. Note the URLs of this file.
++
+You must host the bootstrap Ignition config file because it is too large to
```
staebler: too large to fit in a vApp property

modules/installation-vsphere-machines.adoc
```diff
@@ -0,0 +1,135 @@
...
+  "networkd": {},
+  "passwd": {},
+  "storage": {},
+  "systemd": {}
+}
+----
+<1> Specify the URL of the bootstrap Ignition config file that you hosted.
++
+When you create the Virtual Machine (VM) for the bootstrap machine, you reference
+this this Ignition config file.
```
staebler: remove duplicate `this`

modules/installation-vsphere-machines.adoc
```diff
@@ -0,0 +1,135 @@
...
+
+. Convert the temporary bootstrap file to Base64 encoding.
+
+. Obtain the {op-system} OVA image:
++
+----
+$ curl -J -L -O https://releases-art-rhcos.svc.ci.openshift.org/art/storage/releases/ootpa/410.8.20190412.1/rhcos-410.8.20190412.1-vmware.ova
```
staebler: Is this the final recommended way for a user to obtain the OVA? Or do we still have plans to update this a URL that will yield the latest OVA?

modules/installation-vsphere-machines.adoc
```diff
@@ -0,0 +1,135 @@
...
+.. From the *Hosts and Clusters* tab, right-click your cluster's name and
+click *Deploy OVF Template*.
+.. On the *Select an OVF* tab, specify the name of the {op-system} OVA file
+that you downloaded.
+.. On the *Select a name and folder* tab, set a *Virtual machine name*, such
+as {op-system}, and click the name of the your vSphere cluster.
+.. On the *Select a computer resource* tab, click the name of the your vSphere
+cluster.
+.. On the *Select storage* tab, configure the storage options for your VM.
+For example, select *Thin Provision* if you plan to supply
```
staebler: I think that Thin Provision should be our recommendation. It does not depend on whether the user is going to supply additional storage.

modules/installation-vsphere-machines.adoc
```diff
@@ -0,0 +1,135 @@
...
+additional storage to the cluster instead of using storage on the machines, and
+select the datastore that you specified in your `install-config.yaml` file.
```
staebler: I recommend a separate sentence for the text about selecting the datastore.

modules/installation-vsphere-machines.adoc
```diff
@@ -0,0 +1,135 @@
...
+.. On the *Select network* tab, specify the the network that you configured
+for the cluster, if available.
+.. If you plan to use the same template for all cluster machine types, do not
+specify values on the *Customize template* tab.
+
+. After the template deploys, deploy a VM for a machine in the cluster.
+.. Right-click the template's name and click *Clone* -> *Clone to Virtual Machine*.
+.. On the *Select a name and folder* tab, specify a name for the VM. Include the
+machine type in the name, such as `control-plane-0` or `compute-1`.
```
staebler: There is no requirement that the VM names include the machine type. OpenShift does not care what the VM names are.

modules/installation-vsphere-machines.adoc
```diff
@@ -0,0 +1,135 @@
...
+====
+You must specify the same host for each machine in your cluster.
+====
+.. Optionally, on the *Select storage* tab, customize the storage options.
+.. On the *Select clone options*, select
+*Customize this virtual machine's hardware*.
+.. On the *Customize hardware* tab, click *VM Options* -> *Advanced* -> *Edit Configuration*.
+.. On the *Configuration Parameters* window, click *Add Configuration Params*
+and define the following parameter names and values:
+*** `guestinfo.ignition.config.data`: Specify the name of the Ignition config
```
staebler: The value of this vApp property needs to be the actual base64-encoded text of the Ignition config. It should not be the name of a file.

modules/installation-vsphere-machines.adoc
```diff
@@ -0,0 +1,135 @@
...
+file for this machine type.
+*** `guestinfo.ignition.config.data.encoding`: Specify `base64`.
+*** `enable_disk_uuid`: Specify `true`.
+.. Review the contents of the *Virtual Hardware* panel of the
```
staebler: Separate this into 2 sentences.

modules/installation-installing-bare-metal.adoc
```diff
@@ -0,0 +1,71 @@
...
+$ openshift-install --dir <installation_directory> wait-for bootstrap-complete
+INFO Waiting up to 30m0s for the Kubernetes API at https://api.test.example.com:6443...
+INFO API v1.12.4+c53f462 up
+INFO Waiting up to 30m0s for the bootstrap-complete event...
+----
++
+The command succeeds when the Kubernetes APIServer signals that it has been
+bootstrapped on the control plane machines.
+
+. Remove the bootstrap machine from the load balancer.
```
staebler: We should include text here (or as a subsequent step) that informs the user that they can destroy the bootstrap VM at this point.

modules/installation-operators-config.adoc
```diff
@@ -0,0 +1,23 @@
...
+After the control plane initializes, you must immediately configure some
+Operators so that they all become available.
+
+.Prerequisites
+
+* Your control plane has initialized.
+
+.Procedure
+
+. Watch the cluster components come online:
```
staebler: Is this the recommended approach being given for other UPI installations? I woud have expected us to tell the user to use `wait-for install-complete` first and only look at clusteroperators if there was a problem?

kalexand-rh: Yes, you do. There are a couple of Operators that are expected to not come online at this point, like registry, and I needed a way to watch for them.

kalexand-rh:
@gpei, will you please review?

gpei:
@jiajliu can you help on this?

jiajliu:
@staebler @kalexand-rh 
I noticed in this pr, the steps about upi on vmware was not quite the same with [1]&[2]. Especially about the infras on vmware and aws(such as dns, vm,network etc.) which created by terraform/manually. If this is expected that Terraform is not a must step for the setup of upi/vmware?

[1] https://github.com/openshift/installer/blob/master/upi/vsphere/README.md
[2] https://github.com/openshift/installer/blob/master/docs/user/vsphere/install_upi.md

kalexand-rh:
@jiajliu, my understanding was that the terraform steps were not required and that as long as the user sets up the components correctly, they can do it however they usually configure them. It's more like bare metal in that respect.

staebler:
@jiajliu Kathryn is correct: Terraform is not a required part of the vSphere installation. We use terraform mostly because we stand up and teardown many clusters (particularly to run CI jobs). A customer that is only creating a single cluster may find it easier to do the steps to set up their infrastructure manually or some other way.

installing/installing_vsphere/installing-vsphere.adoc
```diff
@@ -0,0 +1,47 @@
+[id="installing-vsphere"]
+= Installing a cluster on vSphere
+include::modules/common-attributes.adoc[]
+:context: installing-vsphere
+
+toc::[]
+
+In {product-title} version {product-version}, you can install a cluster on
```
dav1x: Most clusters won't be a single host. I would reword this.

modules/installation-vsphere-machines.adoc
```diff
@@ -0,0 +1,135 @@
...
+*Customize this virtual machine's hardware*.
+.. On the *Customize hardware* tab, click *VM Options* -> *Advanced* -> *Edit Configuration*.
+.. On the *Configuration Parameters* window, click *Add Configuration Params*
+and define the following parameter names and values:
+*** `guestinfo.ignition.config.data`: Paste the contents of the base64-encoded
+Ignition config file for this machine type.
+*** `guestinfo.ignition.config.data.encoding`: Specify `base64`.
+*** `enable_disk_uuid`: Specify `true`.
+.. Review the contents of the *Virtual Hardware* panel of the
+*Customize hardware* tab, modify the specified values as required. Ensure that
```
dav1x: Should we list a best practice guide for creating the VMs? There are options like setting eager thick disk and high latency sensitivity that will benefit performance greatly.

dav1x (approved):

installing/installing_vsphere/installing-vsphere.adoc
kalexand-rh: What would I need to change to support multiple hosts? @staebler said that you had to do some things differently, but I don't have the info. (I do have an issue to add the information.)

modules/installation-vsphere-machines.adoc
kalexand-rh: Can you tell me more about the requirements? @staebler said that it might be better to use thin disks to decrease space utilization, but if another way is better, I'm happy to change it.

staebler: My recommendation for thin provision comes from https://coreos.com/os/docs/latest/booting-on-vmware.html. It may be outdated.

installing/installing_vsphere/installing-vsphere.adoc
staebler: @dav1x Are you recommended that we just do not mention using a single host here? There install docs are not describing how a customer would install onto multiple hosts, but such an installation is possible.

Something like the following?
```
you can install a cluster on vSphere infrastructure that you provision.
```

dav1x: @staebler That looks great. A usual source for provisioning would be a cluster (vmware level not OCP).

modules/installation-vsphere-machines.adoc
dav1x: It should be fine. There were a few issues where RH and VMware support collaborated on performance issues. The solution was setting high latency on VMs used in the OCP cluster. 

Here is the option in terraform:
https://www.terraform.io/docs/providers/vsphere/r/virtual_machine.html#latency_sensitivity

This could also be manually tuned as per: 
https://access.redhat.com/documentation/en-us/reference_architectures/2018/html-single/deploying_and_managing_openshift_3.9_on_vmware_vsphere/index#prior_to_ansible_installation

jiajliu:
@kalexand-rh 
http://file.rdu.redhat.com/kalexand/0418/osdocs324/installing/installing_vsphere/installing-vsphere.html#installation-registry-storage-non-production-installing-vsphere
'''
$ oc patch configs.imageregistry.operator.openshift.io cluster --type merge --patch '{"spec":{"storage":{"filesystem":{"volumeSource": {"emptyDir":{}}}}}}'
'''
should be 
'''
oc patch configs.imageregistry.operator.openshift.io cluster --type merge --patch '{"spec":{"storage":{"emptyDir":{}}}}'
'''

installing/installing_vsphere/installing-vsphere.adoc
kalexand-rh: @dav1x, @staebler, can you tell me more about what a user would need to do to install on multiple hosts?

dav1x: @kalexand-rh , nothing special would need to be done. The UPI assigns VMs to a resource pool, resource pools are tied to vsphere clusters (or single hosts).

modules/installation-vsphere-machines.adoc
kalexand-rh: Probably not, but I have [an issue](https://jira.coreos.com/browse/OSDOCS-399) to confirm download/file locations before GA.

modules/installation-vsphere-machines.adoc
```diff
@@ -0,0 +1,135 @@
...
+.. On the *Select a name and folder* tab, specify a name for the VM. You might
+include the machine type in the name, such as `control-plane-0` or `compute-1`.
+.. On the *Select a name and folder* tab, select the name of the folder that
+you created for the cluster.
+.. On the *Select a compute resource* tab, select the name of a host in your
+datacenter.
++
+[IMPORTANT]
+====
+You must specify the same host for each machine in your cluster.
```
kalexand-rh: @dav1x, per the discussion elsewhere on this thread, I'm removing this note to deploy all the VMs in the cluster on the same host.

kalexand-rh:
Thank you @jiajliu! I've rebuilt the preview docs.

modules/installation-vsphere-machines.adoc
kalexand-rh: I'm adding the step to set high latency. Please let me know if you want me to remove it.

installing/installing_vsphere/installing-vsphere.adoc
```diff
@@ -0,0 +1,47 @@
...
+vSphere infrastructure that you provision.
+
+include::modules/installation-overview.adoc[leveloffset=+1]
```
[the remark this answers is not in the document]
kalexand-rh: I double-checked the source file, and it's right in the source.

modules/installation-vsphere-config-yaml.adoc
```diff
@@ -0,0 +1,44 @@
...
+[id="installation-vsphere-config-yaml-{context}"]
+= Sample `install-config.yaml` file for vSphere
+
+You can customize the `install-config.yaml` file to specify more details about
+your {product-title} cluster's platform or modify the values of the required
+parameters.
+
+[source,yaml]
+----
+apiVersion: v1beta4
```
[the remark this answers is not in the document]
kalexand-rh: @staebler, do we need ` networking.machineCIDR`? It's in the link that Jianlin provided but not in the [installer docs](https://github.com/openshift/installer/blob/master/docs/user/vsphere/install_upi.md#setting-up-install-config-for-installer).

modules/installation-vsphere-machines.adoc
```diff
@@ -0,0 +1,130 @@
...
+.Procedure
+
+////
+This will work for GA.
+. Obtain the {op-system} images from the
+link:https://access.redhat.com/downloads[Product Downloads] page on the Red
+Hat customer portal.
+////
+
+. Convert the bootstrap, master, and worker Ignition config files to Base64
```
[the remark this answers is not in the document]
kalexand-rh: I added a way to do this on Linux. If you have a better method, I'm happy to update the example.

kalexand-rh:
@jianlinliu, I made updates based on your comments. Thank you!

modules/installation-vsphere-config-yaml.adoc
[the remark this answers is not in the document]
staebler: No, the user does not need to set `networking.machineCIDR`. That field is used to establish an IP address for the machines that the installer creates. Since the installer is not responsible for creating any machines in a UPI installation, the installer does not need to know the machine CIDR.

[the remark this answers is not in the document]
staebler: I don't know if it is better to imply to the user that they should set the machine CIDR, even though the installer will not use it, to maintain consistency with IPI installations.

modules/installation-dns-upi.adoc
```diff
@@ -1,13 +1,15 @@
...
-The following DNS records are required for an {product-title} cluster. In each
-record, `<cluster_name>` is the cluster name and `<base_domain>` is the cluster
-base domain that you specify in the `install-config.yaml` file.
+The following DNS records are required for a user provisioned infrastructure
+(UPI) {product-title} cluster. In each record, `<cluster_name>` is the cluster
+name and `<base_domain>` is the cluster base domain that you specify in the
+`install-config.yaml` file.
 
 .Required DNS records
 [cols="1a,3a,5a",options="header"]
```
jianlinliu: I feel like "etcd-server-ssl._tcp.<cluster_name>.<base_domain>" is wrong, am I right? Because I saw "_etcd-server-ssl._tcp.<cluster_name>.<base_domain>" in a IPI install.
And 'control-panel-0.<cluster_name>.<base_domain>', 'compute-1.<cluster_name>.<base_domain>' is not required here? Pls help confirm.

modules/installation-vsphere-machines.adoc
```diff
@@ -0,0 +1,150 @@
...
+Before you install a user provisioned infrastructure (UPI) cluster on vSphere,
+you need to create {op-system} machines on a single host for it to use.
+
+.Prerequisites
+
+* Obtain the Ignition config files for your cluster.
+* Have access to an HTTP server that you can access from your computer and that
+the machines that you create can access.
+
+.Procedure
```
jianlinliu: Does anyone really try such procedure step by step strictly. 
Today I tried the whole day, never had a successful install. I am using vcsa-qe.vmware.devcluster.openshift.com cluster for testing. Seem like bootstrap never get ready, I can not login bootstrap either, I have no way to debug where is wrong? I am not sure it is caused by missed some steps in this doc, or my understanding for the procedure is wrong. Anyone could provide help here? If anyone want to check my cluster, log in https://vcsa-qe.vmware.devcluster.openshift.com, and switch to "VMs and Templates inventory view", check 'jialiu-vsphere1' folder for all my instances. My validation is totally is blocked here.

jianlinliu: I also opened a bug for that:
https://bugzilla.redhat.com/show_bug.cgi?id=1703803

modules/installation-vsphere-config-yaml.adoc
jiajliu: @staebler 
About the replicas in install-config.yaml, I noticed only master field shown in the doc here.
1. Then how installer decide the number of computer nodes?
2. And how installer decide the number of master nodes?just respect to replicas in install-config.yaml?
I found when the replicas in install-config is not equal with the number of hosts created by terraform, the master node number will respect to the actual hosts number. Is this expected?
```
# cat .openshift_install_state.json |grep replicas
                "replicas": 3,
                    "replicas": 3,
# cat terraform.tfvars|grep control_plane_count
control_plane_count = 4
// must match the value of control_plane_count.
# oc get node
NAME              STATUS                     ROLES    AGE   VERSION
compute-0         Ready                      worker   15m   v1.13.4+27a00af64
compute-1         Ready                      worker   15m   v1.13.4+27a00af64
compute-2         Ready                      worker   16m   v1.13.4+27a00af64
control-plane-0   Ready                      master   15m   v1.13.4+27a00af64
control-plane-1   Ready,SchedulingDisabled   master   15m   v1.13.4+0e917197e
control-plane-2   Ready,SchedulingDisabled   master   15m   v1.13.4+0e917197e
control-plane-3   Ready                      master   15m   v1.13.4+27a00af64
```

modules/installation-dns-upi.adoc
staebler: Yes, the etcd SRV record should have an underscore at the beginning.

staebler: Machines internal to the cluster do need to be able to resolve the hostnames of the other machines in the cluster.

modules/installation-vsphere-machines.adoc
```diff
@@ -0,0 +1,150 @@
...
+For example, if you use a Linux operating system, you can use the `base64`
+command to encode the files.
++
+----
+$ base64 <path_to_installation_directory>/bootstrap.ign > <path_to_installation_directory>/bootstrap.64
+$ base64 <path_to_installation_directory>/master.ign > <path_to_installation_directory>/master.64
+$ base64 <path_to_installation_directory>/worker.ign > <path_to_installation_directory>/worker.64
+----
+
+. Upload the Base64-encoded bootstrap Ignition config file that the
```
staebler: The user should not upload the base64-encoded bootstrap Ignition config file. The user should upload the plain-text config file.

modules/installation-vsphere-machines.adoc
staebler: https://github.com/openshift/openshift-docs/pull/14523#discussion_r279456130

modules/installation-vsphere-config-yaml.adoc
staebler: >     1. Then how installer decide the number of computer nodes?
The installer does not decide the number of compute nodes. The end user decides that. The installer does not care how many compute nodes there are. The installer is not created the compute nodes.

>     2. And how installer decide the number of master nodes?just respect to replicas in install-config.yaml?
The user is required to specify the number of control plane nodes in the install-config.yaml. This is required so that the appropriate Endpoints can be created for etcd.

modules/installation-vsphere-config-yaml.adoc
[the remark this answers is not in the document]
kalexand-rh: If it's not necessary, then I don't think we need to include it. I don't know what I'd tell users to specify for the value.

modules/installation-dns-upi.adoc
kalexand-rh: @staebler, is there a better way to get this point across for this use case?

modules/installation-vsphere-machines.adoc
kalexand-rh: Sorry, my notes said "Host the bootstrap after you base64 encode it." I've updated the docs to not base64 encode the bootstrap.ign file.

modules/installation-dns-upi.adoc
staebler: I think it is implied at https://github.com/openshift/openshift-docs/blame/6869218df46b2a59a9a1f97cbf460fb93205f4fb/modules/installation-network-upi.adoc#L73. But maybe we could call it out explicitly there. I can't think of language to use that is great. Perhaps something like the following.
```
Each machine must be able to resolve the hostnames of the other machines.
```

modules/installation-vsphere-config-yaml.adoc
jiajliu: To be more clear, I want to ask, even end user "decide" the number of control plane nodes and create all the hosts, if that means, installer will according to actual hosts to schedule the control plane to the master machines, but not the replicas in config. The replicas just decide the number of etcd endpoints. Is this right?

staebler: Yes, for UPI the replicas only dictate the number of etcd endpoints.

jiajliu: @kalexand-rh 
Thx for @staebler's confirm. Here suggest to give a brief explanation the "replicas" in install-config sample [1], that "the number of etcd endpoints", which is not quite the same with the config in ipi.
[1] http://file.rdu.redhat.com/kalexand/0418/osdocs324/installing/installing_vsphere/installing-vsphere.html#installation-vsphere-config-yaml-installing-vsphere

staebler: I think that would only confuse things. The user *should* be setting the replicas to the number of control plane machines that they will be creating.

modules/installation-vsphere-machines.adoc
jianlinliu: Tried, still not working. leave comments in bugzillar.
I have one question, is there such similar setting (e.g: guestinfo.ignition.config.data:) when deploying cluster via Terraform? I did not see such setting in a cluster deployed via Terraform. Do I miss something? If indeed no such setting, how to inject ignition files into instances when launching via Terraform. Any hint?

modules/installation-vsphere-machines.adoc
```diff
@@ -0,0 +1,142 @@
...
+.. Optionally, on the *Select storage* tab, customize the storage options.
+.. On the *Select clone options*, select
+*Customize this virtual machine's hardware*.
+.. On the *Customize hardware* tab, click *VM Options* -> *Advanced*.
+*** From the *Latency Sensitivity* list, select *High*.
+*** Click *Edit Configuration*, and on the *Configuration Parameters* window,
+click *Add Configuration Params*. Define the following parameter names and values:
+**** `guestinfo.ignition.config.data`: Paste the contents of the base64-encoded
+Ignition config file for this machine type.
+**** `guestinfo.ignition.config.data.encoding`: Specify `base64`.
```
staebler: As an alternate, the user can set these properties after doing the clone. On the VM, open the Configure tab. Select Settings/vApp Options from the left menu. In the Properties section, the two guestinfo properties will be listed. The user can select a property and set its value by pressing SET VALUE. This may be easier for the user than having to manually enter the name of the property.

modules/installation-vsphere-machines.adoc
staebler: The same properties are set when launched by terraform. See https://github.com/openshift/installer/blob/bc8db97a005785e6524ff0e504d2c4d0c651dc94/upi/vsphere/machine/main.tf#L45-L50.

modules/installation-vsphere-machines.adoc
kalexand-rh: @staebler, yes. If users set these values after the VM is cloned, do they also need to set the disk_uuid?

staebler: The `disk.EnableUUID` still needs to be set. That can be set either before or after the clone: Whichever the user prefers.

modules/installation-vsphere-machines.adoc
jianlinliu: @staebler yeah, I saw such code, but do you where can I review the vapp setting from vsphere client web console?

staebler: > @staebler yeah, I saw such code, but do you where can I review the vapp setting from vsphere client web console?

If you set the vApp property after the clone (which is how terraform does it), then you can see the property under the Configure tab of the VM. See https://github.com/openshift/openshift-docs/pull/14523#discussion_r279769953. However, if you set the property using the Configuration Parameters, the values will not show up in the vApp properties on the Configure tab.

staebler: But, to be clear, I have tested setting the vApp properties both ways (ie, from Configuration Parameters and from the Configure tab). It works both ways.

staebler: Screenshot_2019-04-30 vSphere - bootstrap-0 - vApp Options

staebler: Screenshot_2019-04-30 vSphere - compute-0 - vApp Options

modules/installation-dns-upi.adoc
jianlinliu: Per https://github.com/openshift/installer/blob/master/upi/vsphere/machine/ignition.tf#L71, hostname in the machine would be set by terraform. While in this manual doc, there is no any place to instruction how to set the machine's hostname. In my testing, I launched master-0, maser-1, and worker-0 and so on, but all machines' hostname is 'localhost'. How to resolve that?

modules/installation-vsphere-machines.adoc
```diff
@@ -0,0 +1,142 @@
...
+**** `disk.EnableUUID`: Specify `true`.
```
jianlinliu: s/true/TRUE/

modules/installation-vsphere-machines.adoc
```diff
@@ -0,0 +1,142 @@
...
+.. In the *Virtual Hardware* panel of the
+*Customize hardware* tab, modify the specified values as required. Ensure that
+the amount of RAM, CPU, and disk storage meets the minimum requirements for the
+machine type.
```
jianlinliu: And after 'Add Configuration Params', in the follow wizard - 'Customize vApp properties', user need input 'guestinfo.ignition.config.data.encoding' and 'guestinfo.ignition.config.data.encoding' again. This is really important. We need mention the step here.

staebler: @jianlinliu That is not necessary. The user needs to either specify the properties from the Configuration Parameters or from the Configure tab. They do not need to do both.

modules/installation-dns-upi.adoc
staebler: See https://github.com/openshift/installer/blob/master/docs/user/vsphere/install_upi.md#hostname

kalexand-rh: I'm making a follow-up issue to add information about the DHCP nodes and working with Matthew to update the prereqs to require DHCP and recommend using DHCP for long-term node management.

modules/installation-vsphere-machines.adoc
kalexand-rh: @staebler, @jianlinliu  should we consider it to be a usability enhancement to add the the parameters to the VMs after initialization instead of setting them earlier? I can change the docs to just add them here.

modules/installation-dhcp-upi.adoc
```diff
@@ -0,0 +1,19 @@
+// Module included in the following assemblies:
+//
+// * installing/installing_bare_metal/installing-bare-metal.adoc
+// * installing/installing_vsphere/installing-vsphere.adoc
+
+[id="installation-dhcp-upi-{context}"]
```
kalexand-rh: @staebler, does this section sound right to you?

modules/installation-vsphere-machines.adoc
staebler: I find it to be a moderately better user experience to use the Configure/vApp Properties UI. However, since the user already has to go to the Advanced/Configuration Properties UI to enable disk UUID, I don't have a strong opinion about which way we should guide the user. I do think that we should only guide them one way and not both.

modules/installation-dhcp-upi.adoc
staebler: The last sentence seems to me to be a toss away. If we are going to mention that the machines can be configured to use static IP addresses, then we should also give them some guidance on how to do that. Otherwise, let's recommend the use of DHCP and leave it at that.

kalexand-rh: I have a follow-up JIRA to add the non-DHCP path later.

modules/installation-vsphere-machines.adoc
kalexand-rh: I agree that we should only describe one set of actions, and if we didn't need to go enable disk UUID, this method would win. I vote to leave it as-is for now.

jianlinliu:
Completed my 1st round of review, except DNS issue for machine hostname issue, most looks good to me.

kalexand-rh:
Thank you, @jianlinliu! I'm going to merge this PR and keep your bug open in case there's a doc change I can make now to help with the DNS issue.

PR #14524, opened 2019-04-18
BUG BZ1695719 -- removed outdated note

jhoyt-rh:
Verified with Ashley this only requires removing the outdated note about limitation on x86_64 architecture


jhoyt-rh:
QE review and peer review not needed, since removal of an outdated note was all that was required, and Vikram said no QE review was needed.

PR #14529, opened 2019-04-18
cluteringress is now ingresscontroller

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1698146

### DIFF
--- a/modules/infrastructure-moving-router.adoc
+++ b/modules/infrastructure-moving-router.adoc
@@ -14,10 +14,10 @@ is displayed to a worker node.
 
 .Procedure
 
-. View the `clusteringress` Custom Resource for the router Operator:
+. View the `IngressController` Custom Resource for the router Operator:
 +
 ----
-$ oc get clusteringress default -n openshift-ingress-operator -o yaml <1>
+$ oc get ingresscontroller default -n openshift-ingress-operator -o yaml <1>
 ----
 <1> The router is managed by an Operator that is named
 `openshift-ingress-operator`, and its Pod is in the `openshift-ingress-operator`
@@ -28,7 +28,7 @@ The command output resembles the following text:
 [source,yaml]
 ----
 apiVersion: ingress.openshift.io/v1alpha1
-kind: ClusterIngress
+kind: IngressController
 metadata:
   creationTimestamp: 2019-01-28T17:23:39Z
   finalizers:
@@ -37,7 +37,7 @@ metadata:
   name: default
   namespace: openshift-ingress-operator
   resourceVersion: "1294295"
-  selfLink: /apis/ingress.openshift.io/v1alpha1/namespaces/openshift-ingress-operator/clusteringresses/default
+  selfLink: /apis/ingress.openshift.io/v1alpha1/namespaces/openshift-ingress-operator/ingresscontrolleres/default
   uid: 73ff7bfd-2321-11e9-8ff2-026a37856868
 spec:
   defaultCertificateSecret: null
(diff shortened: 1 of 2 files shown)


modules/infrastructure-moving-router.adoc
```diff
@@ -14,10 +14,10 @@ is displayed to a worker node.
 
 .Procedure
 
-. View the `clusteringress` Custom Resource for the router Operator:
+. View the `IngressController` Custom Resource for the router Operator:
 +
 ----
-$ oc get clusteringress default -n openshift-ingress-operator -o yaml <1>
+$ oc get ingresscontroller default -n openshift-ingress-operator -o yaml <1>
```
kalexand-rh: @Miciah, the CR for the router's changed, and I'm not sure if the instructions for moving it to an infrastructure node is right anymore. Will you PTAL?

Miciah: I'm confused by the change.  It looks like you started updating the existing example and then added a another version that supersedes the old version.  You can delete https://github.com/openshift/openshift-docs/blob/0947083c5fd006e36ba64c11f2f36203d3446d29/modules/infrastructure-moving-router.adoc#L30-L57 (which has the updated `kind` but incorrect `selfLink` and old `spiVersion`, `finalizers`, `spec`, and `status`) and keep https://github.com/openshift/openshift-docs/blob/0947083c5fd006e36ba64c11f2f36203d3446d29/modules/infrastructure-moving-router.adoc#L59-L81 (which is correct), and then that example will be correct.

See my other comments for a couple other minor issues.  Also, the `oc get pod` output at the end of the document is funny—I don't get `IP`, `NODE`, or `NOMINATED NODE` columns when I run the command.  With those changes, the instructions should be fine.

modules/infrastructure-moving-router.adoc
```diff
@@ -14,10 +14,10 @@ is displayed to a worker node.
...
 ----
 <1> The router is managed by an Operator that is named
 `openshift-ingress-operator`, and its Pod is in the `openshift-ingress-operator`
```
Miciah: This is not quite right.  It is true that the router (which is being renamed to "ingress controller"—Brandi has another PR for this renaming) is managed by the Ingress Operator, and the operator's pod is in the `openshift-ingress-operator` project, but I wouldn't say that the operator is named `openshift-ingress-operator`.  Anyway, I don't think that the operator's name or namespace are relevant in describing this command, which is getting the ingress controller named `default`, which is in the `openshift-ingress-operator` namespace.

modules/infrastructure-moving-router.adoc
```diff
@@ -55,14 +55,38 @@ spec:
...
-. Edit the `clusteringress` resource and change the `nodeSelector` to use the
+. Edit the `ingresscontroller` resource and change the `nodeSelector` to use the
 `infra` label:
 +
 ----
-$ oc edit clusteringress default -n openshift-ingress-operator -o yaml
+$ oc edit ingresscontroller default -n openshift-ingress-operator -o yaml
 ----
 +
 Update the `nodeSelector` stanza to reference the `infra` label as shown:
```
Miciah: "Update" is incorrect because the default `IngressController` resource does not have any `nodeSelector` stanza.  "Add" would be better, and better still would be to indicate further that the `nodeSelector` stanza needs to be added under the `spec` stanza:
```yaml
spec:
  nodeSelector:
    matchLabels:
      node-role.kubernetes.io/infra: ""
```

Miciah: Sorry, the above yaml is incorrect—`nodeSelector` needs to be under `nodePlacement`.  The correct yaml is as follows:
```yaml
spec:
  nodePlacement:
    nodeSelector:
      matchLabels:
        node-role.kubernetes.io/infra: ""
```

danehans:
@kalexand-rh I didn't realize you were working on this content. I created similar docs in #14589

modules/infrastructure-moving-router.adoc
```diff
@@ -81,9 +75,9 @@ Update the `nodeSelector` stanza to reference the `infra` label as shown:
 ----
 $ oc get pod -n openshift-ingress
```
lihongan: Just adding `-o wide` to this command or cannot see those fields like `IP`, `NODE`. Since following steps need to ensure the pod is running on the `infra` node, so we should keep the old output.

modules/infrastructure-moving-router.adoc
```diff
@@ -81,9 +75,9 @@ Update the `nodeSelector` stanza to reference the `infra` label as shown:
...
 
-NAME                              READY     STATUS        RESTARTS   AGE       IP           NODE                           NOMINATED NODE
-router-default-86798b4b5d-bdlvd   1/1       Running       0          28s       10.130.2.4   ip-10-0-217-226.ec2.internal   <none>
-router-default-955d875f4-255g8    0/1       Terminating   0          19h       10.129.2.4   ip-10-0-148-172.ec2.internal   <none>
+NAME                              READY     STATUS        RESTARTS   AGE
+router-default-86798b4b5d-bdlvd   1/1       Running       0          28s
+router-default-955d875f4-255g8    0/1       Terminating   0          19h
```
lihongan: Above three lines can be reverted if using command `oc get pod -n openshift-ingress -o wide`.
And in latest 4.1 build the output like:
```
$ oc -n openshift-ingress get pod -o wide
NAME                              READY   STATUS    RESTARTS   AGE     IP            NODE                                               NOMINATED NODE   READINESS GATES
router-default-66c797c795-7ls7x   1/1     Running   0          4h27m   10.128.2.12   ip-172-31-159-23.ap-northeast-2.compute.internal   <none>           <none>
router-default-66c797c795-jj9k6   1/1     Running   0          4h27m   10.128.2.10   ip-172-31-159-23.ap-northeast-2.compute.internal   <none>           <none>
```
Please note one extra column named `READINESS GATES` is added.

kalexand-rh: Thank you @lihongan! I've updated the command. Will you please check?

lihongan: LGTM.  Thank you @kalexand-rh

kalexand-rh:
Thank you @lihongan and @Miciah! Merging.

PR #14533, opened 2019-04-18
[Istio] Incorrect namespace in docs and name no longer has to be istio-installation

brian-avery:
@geekspertise @JStickler Can you please take a look?


brian-avery:
This fixes Bug 1701393

JStickler:
LGTM

Issue #14536, opened 2019-04-20
[enterprise-3.9] Improve cert generation documentation to include DNS


hexfusion:
### Which section(s) is the issue in?

admin_guide/assembly_restoring-cluster.html#manually-adding-etcd-host_admin-restore-cluster

### What needs fixing?

We outline ENV variables to generate new certs but the example only shows defining `IP:` in SAN. Most clusters will also use `DNS:`. We should adjust the example to include the DNS usage.

existing

```
export NEW_ETCD_HOSTNAME="*etcd0.example.com*"
export NEW_ETCD_IP="192.168.55.21"

export CN=$NEW_ETCD_HOSTNAME
export SAN="IP:${NEW_ETCD_IP}"
export PREFIX="/etc/etcd/generated_certs/etcd-$CN/"
export OPENSSLCFG="/etc/etcd/ca/openssl.cnf"
```

proposed 

```
export NEW_ETCD_HOSTNAME="*etcd0.example.com*"
export NEW_ETCD_IP="192.168.55.21"

export CN=$NEW_ETCD_HOSTNAME
export SAN="IP:${NEW_ETCD_IP}, DNS:${NEW_ETCD_HOSTNAME}"
export PREFIX="/etc/etcd/generated_certs/etcd-$CN/"
export OPENSSLCFG="/etc/etcd/ca/openssl.cnf"
```

/assign

kalexand-rh:
We merged this change, so I'm closing the issue. Thanks @hexfusion!

PR #14537, opened 2019-04-20
Update the SAN to HOSTNAME of the etcd member

vwalek:
Updating the SAN to reflects also the DNS name of the host.

Fix for https://bugzilla.redhat.com/show_bug.cgi?id=1701705

### DIFF
--- a/day_two_guide/topics/proc_scaling-etcd-manual.adoc
+++ b/day_two_guide/topics/proc_scaling-etcd-manual.adoc
@@ -30,7 +30,7 @@ export NEW_ETCD_HOSTNAME="*etcd0.example.com*"
 export NEW_ETCD_IP="192.168.55.21"
 
 export CN=$NEW_ETCD_HOSTNAME
-export SAN="IP:${NEW_ETCD_IP}"
+export SAN="IP:${NEW_ETCD_IP}, DNS:${NEW_ETCD_HOSTNAME}"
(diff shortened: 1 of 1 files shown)


vwalek:
@openshift/team-documentation

vikram-redhat:
@jianlinliu - can you please confirm this?

hexfusion:
/lgtm
Fixes #14536 
Thanks @vwalek

hexfusion:
This should be backported to all 3.x not just 3.11

vikram-redhat:
Going ahead and merging this.

vikram-redhat:
@jianlinliu - can you confirm it applies to previous releases as well?

jianlinliu:
@geliu2016 PTAL.

geliu2016:
@jianlinliu LGTM

vikram-redhat:
@geliu2016 - can you confirm if this applies in older versions?

hexfusion:
These docs bit us in the wild, the DNS in SAN is a fairly common configuration and not one that would be version specific to OpenShift. I would ask that we backport this to any version where we give the openssl indications to gen certs as customers will use the examples directly. An extra unused DNS in SAN is better than an etcd cluster that doesn't reach quorum.

vikram-redhat:
Done till 3.7. We are not supporting working on docs before that version.

PR #14539, opened 2019-04-22
Description of the routes/custom-host resource

mlbiam:
After digging through github issues to try to solve https://github.com/openshift/origin/issues/22612 I found this needed to be added to my rbac policy.  This resource does not appear in any documentation post 3.6.


kalexand-rh:

vikram-redhat:
Thanks @mlbiam - we are looking to redo the REST API guides post 4.1. We will keep this on the backburner till we are ready to make this change.

PR #14549, opened 2019-04-22
bug 1688487 moving monitoring

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1688487

@liangxia, will you PTAL?

### DIFF
--- a/modules/infrastructure-moving-monitoring.adoc
+++ b/modules/infrastructure-moving-monitoring.adoc
@@ -6,17 +6,12 @@
 = Moving the monitoring solution
 
 
-By default, a Prometheus, Grafana, and AlertManager stack is deployed to
+By default, the Prometheus Cluster Monitoring stack, which contains Prometheus,
+Grafana, and AlertManager, is deployed to
 provide cluster monitoring. It is managed by the Cluster Monitoring Operator,
 and to move its components to different machines, you apply a custom ConfigMap.
 
 .Procedure
-
-. Optionally, review the default ConfigMap for the monitoring stack:
-+
-----
-$ oc get configmap cluster-monitoring-config -n openshift-monitoring
-----
 
 . Save the following ConfigMap definition as the
 `cluster-monitoring-configmap.yaml` file:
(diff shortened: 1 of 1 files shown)


liangxia:
@juzhao will review the changes.

modules/infrastructure-moving-monitoring.adoc
```diff
@@ -5,10 +5,65 @@
...
-The monitoring solution uses Cluster Version Operator (CVO) to create the
-ConfigMap that the monitoring Operator uses to determine how to deploy its
-resources. Because it uses the CVO and users cannot modify the CVO, you cannot
-change where the Operator deploys resources.
-====
\ No newline at end of file
+
+By default, a Prometheus, Grafana, and AlertManager stack is deployed to
+provide cluster monitoring. It is managed by the Cluster Monitoring Operator,
+and to move its components to different machines, you apply a custom ConfigMap.
```
juzhao: I think it should be
change "stack is" to "stacks are"
"It is managed" to "they are managed"
"you apply" to "you should apply"

modules/infrastructure-moving-monitoring.adoc
```diff
@@ -5,10 +5,65 @@
...
+
+.Procedure
+
+. Optionally, review the default ConfigMap for the monitoring stack:
++
+----
+$ oc get configmap cluster-monitoring-config -n openshift-monitoring
```
juzhao: no need for this step, cluster-monitoring-config configmap is not exist when the cluster is UP, we could create the configmap by the following yaml file or edit the cluster-monitoring-config configmap if it is created

modules/infrastructure-moving-monitoring.adoc
```diff
@@ -5,10 +5,65 @@
...
+        node-role.kubernetes.io/infra: ""
+----
++
+Running this ConfigMap forces the components of the monitoring stack to redeploy
+to infrastructure nodes.
+
+. Apply the new ConfigMap:
++
+----
+$ oc create cluster-monitoring-configmap.yaml
```
juzhao: should be "oc create -f cluster-monitoring-configmap.yaml"

Note: we assume the cluster-monitoring-config configmap is not exist

modules/infrastructure-moving-monitoring.adoc
```diff
@@ -5,10 +5,65 @@
...
+----
+
+. Watch the monitoring Pods move to the new machines:
++
+----
+$ watch 'oc get pod -n openshift-monitoring'
```
juzhao: 'oc get pod -n openshift-monitoring -o wide'
would be better

modules/infrastructure-moving-monitoring.adoc
kalexand-rh: It's one stack that contains multiple components, so it's right to use singular here. I'll make that clearer.

kalexand-rh:
This bug was approved on the PR. Merging.

PR #14551, opened 2019-04-22
Removed mistaken insecure-flag from VirtualCenter section (because it…

palonsoro:
Removed `insecure-flag` references from `VirtualCenter` section examples, as this can only be configured on `Global` section.


vikram-redhat:
@palonsoro Hi Pablo - this PR was closed due to our migration to the 4.1 platform. Are you able to open the same PR against the master-3 branch? I will then priortize it.

PR #14557, opened 2019-04-22
BZ-1699212: Included instructions for expanding persistent volumes.

huffmanca:
BZ-1699212: Included instructions for expanding persistent volumes.

This is for OS 4.x, and is a conversion of the 3.x content.

### DIFF
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -361,6 +361,8 @@ Topics:
     File: persistent-storage-csi
   - Name: Persistent storage using OpenStack Manila
     File: persistent-storage-manila
+- Name: Expanding persistent volumes
+  File: expanding-persistent-volumes
 - Name: Dynamic provisioning
   File: dynamic-provisioning
 ---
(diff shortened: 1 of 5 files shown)


modules/storage-expanding-glusterfs-pvc.adoc
```diff
@@ -0,0 +1,33 @@
+// Module included in the following assemblies:
+//
+// * storage/expanding-persistent-volumes.adoc
+
+[id="expanding-glusterfs-pvc-{context}"]
+= Expanding GlusterFS-based Persistent Volume Claims
```
liangxia: Gluster is not supported in 4.1, perhaps using AWS EBS as example.

huffmanca: I've removed this section entirely. I'm not certain having an EBS example would be beneficial, as it would be the exact steps as the section above, just with slightly different names?

If you think it adds value I can include it here.

liangxia:
@chao007  ( who is the feature owner from QE ) will review the detail.

chao007:
The update LGTM

PR #14562, opened 2019-04-23
Added procedure for configuring identity providers via the console

ahardin-rh:
https://jira.coreos.com/browse/PROD-832

@spadgett PTAL

@huffmanca I am proposing that we add this to each of the assemblies you created for configuring specific identity providers. I think we can re-use this module across all of those assemblies by keeping this generic. Thoughts?

### DIFF
--- a/modules/identity-provider-configuring-using-web-console.adoc
+++ b/modules/identity-provider-configuring-using-web-console.adoc
@@ -1,0 +1,26 @@
+// Module included in the following assemblies:
+//
+//* authentication/identity_providers/configuring-htpasswd-identity-provider.adoc
+//* authentication/identity_providers/configuring-oidc-identity-provider.adoc
+
+[id="identity-provider-configuring-using-the-web-console-{context}"]
+= Configuring identity providers using the web console
+
+Configure your identity provider (IDP) through the web console instead of the CLI.
+
+.Prerequisites
+
+* You must be logged in to the web console as a cluster administrator.
+
+.Procedure
+
+. Navigate to *Administration* -> *Cluster Settings*.
+. Under the *Global Configuration* tab, click *OAuth*.
+. Under the *Identity Providers* section, select your identity provider from the
+*Add* drop-down menu.
+
+[NOTE]
+====
+You can specify multiple IDPs through the web console without overwriting
+existing IDPs.
+====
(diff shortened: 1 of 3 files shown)


huffmanca:
@ahardin-rh,

This looks clean and straightforward. I agree I prefer keeping it generic, that way it can be added across multiple assemblies.

On today's built I only see options for HTPasswd and OpenID Connect. Is the web console expected to include every IDP? If not, then it should only be added to the relevant assemblies.

In addition, can I specify multiple IDPs through the web console? If I specify a second one, will it erase the first? If so, we should add a note here saying that you must configure these through the CLI to have multiple IDPs.

spadgett:
> On today's built I only see options for HTPasswd and OpenID Connect. Is the web console expected to include every IDP? If not, then it should only be added to the relevant assemblies.

Those are the only two in 4.1, but we're adding a bunch more in 4.2.

> In addition, can I specify multiple IDPs through the web console? If I specify a second one, will it erase the first? If so, we should add a note here saying that you must configure these through the CLI to have multiple IDPs.

You can add multiple in the console. They won't overwrite the existing IDPs.

spadgett (approved):
lgtm, agree with keeping it simple 👍

modules/identity-provider-configuring-using-web-console.adoc
```diff
@@ -0,0 +1,27 @@
...
+//* authentication/identity_providers/configuring-oidc-identity-provider.adoc
+
+[id="identity-provider-configuring-using-the-web-console-{context}"]
+= Configuring identity providers using the web console
+
+Configure your identity provider through the web console instead of the CLI.
+
+.Prerequisites
+
+* You must be logged in to the web console.
```
spadgett: ...as a cluster administrator?

ahardin-rh:
@huffmanca @spadgett Thanks for the reviews! This is now updated and I applied the module to the HTPasswd and OpenID Connect assemblies.

ahardin-rh:
@xiaocwan Can you please provide a QE review?

xiaocwan:
@ahardin-rh LGTM. Agree to keep it simple and generic. Only there might be a little confused for users about the steps in web console doc, seems here could be a little hint or description such as your above discussion -- the "Add" action will not overwrite the existed IDPs, etc. This is only from my point of view. HDYT?

ahardin-rh:
@xiaocwan Thank you! I added a note to clarify that you can specify multiple IDPs through the web console without overwriting existing IDPs. I'm going to merge this to get it in, but please let me know if further changes are needed.

PR #14573, opened 2019-04-23
Update usage for OpenShift CLI

jboxman:
Update usage based on our recent discussion.

### DIFF
--- a/contributing_to_docs/term_glossary.adoc
+++ b/contributing_to_docs/term_glossary.adoc
@@ -285,9 +285,15 @@ allowed for use across all distribution documentation:
 ''''
 === OpenShift CLI
 
-Usage: OpenShift CLI
+Usage: OpenShift CLI (`oc`)
 
 The `oc` tool is the command line interface of OpenShift 3 and 4.
+
+When referencing as a prerequisite for a procedure module, use the following
+construction: Install the OpenShift Command-line Interface (CLI), commonly known
+as `oc`.
+
+When referencing in overview text, use: OpenShift CLI (`oc`).
 
 ''''
 === OpenShift master


jboxman:
@openshift/team-documentation PTAL, thanks.

contributing_to_docs/term_glossary.adoc
```diff
@@ -285,10 +285,16 @@ allowed for use across all distribution documentation:
 ''''
 === OpenShift CLI
 
-Usage: OpenShift CLI
+Usage: OpenShift CLI (`oc`)
 
 The `oc` tool is the command line interface of OpenShift 3 and 4.
 
+When referencing as a prerequisite for a procedure module, use the following
+construction: Install the OpenShift Command-line Interface (CLI), commonly known
```
bergerhoffer: Do we need to put a prerequisite of having the CLI installed in front of every single procedure that uses a CLI command? We certainly don't right now. We should determine when to or not to (if not always).

And if we determine that we should always, then I still think that this is overkill: "Install the OpenShift Command-line Interface (CLI), commonly known as `oc`", as opposed to just "Install the OpenShift CLI (`oc`)" as the prerequisite. Maybe spelling it out more makes sense in places where we know people are just getting started, but it seems a bit much to use for every procedure.

huffmanca: I agree that this seems like overkill if we have to add this into every assembly where the `oc` command is used. To me it's the same as saying that we require a running OpenShift cluster before running commands.

I think on the sections that predominantly focus on the CLI, such as some of the CLI reference bucket, it makes sense. Throughout most of the other guides it feels like overkill, though.

mburke5678: +1
Installting the CLI is part of the installation, not a specific prerequisite to the task in a given topic.

kalexand-rh: I agree that installing the CLI is part of installation, but then only the person installing OpenShift has the CLI installed, not the person completing that specific task. I think it's fair to assume that the cluster is up but not fair to assume that each user has the CLI installed and the correct permissions to perform an action.

kalexand-rh: I think that we should pair this prereq with another prereq that states the permission you need to run the commands in the procedure.

vikram-redhat:
@jboxman - this can be merged or were you thinking of making more changes?

jboxman:
@vikram-redhat, I think everyone has been too busy to make any further suggestions. I haven't had time to decide what changes to make based on the feedback already provided.

vikram-redhat:
@jboxman I don't think we will get more input on this PR. Did you want to either follow up or close?

jboxman:
Based on https://github.com/openshift/openshift-docs/pull/16791#discussion_r331615156 I'm going to merge this; I'd like to revisit the observation about communicating what the relevant security role a user needs. That's important and using consistent language about that would be helpful.

jboxman:
@vikram-redhat, how many branches are we committing the contributing guides to? Just master or each 4.x branch? Thanks.

vikram-redhat:
@jboxman the contribution guides are only in master, so you may need to move this there.

PR #14574, opened 2019-04-23
Update Device Manager and Plugins for 4.1

mburke5678:
Per @sjenning in Slack: "any docs on how to install a device plugin will need to be rewritten for 4.x."
https://coreos.slack.com/archives/GBF4GNEFM/p1555946550029500

cc @vikaschoudhary16

### DIFF
--- a/modules/nodes-pods-plugins-about.adoc
+++ b/modules/nodes-pods-plugins-about.adoc
@@ -17,7 +17,7 @@ Containers are supported by individual vendors.
 ====
 
 A device plug-in is a gRPC service running on the nodes (external to
-`atomic-openshift-node.service`) that is responsible for managing specific
+the `kubelet`) that is responsible for managing specific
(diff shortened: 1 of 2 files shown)


modules/nodes-pods-plugins-about.adoc
```diff
@@ -17,7 +17,7 @@ Containers are supported by individual vendors.
 ====
 
 A device plug-in is a gRPC service running on the nodes (external to
-`atomic-openshift-node.service`) that is responsible for managing specific
+the `kubelet-config-controller`) that is responsible for managing specific
```
sjenning: this should just be `kubelet`

sjenning:
just one comment, then LGTM

mburke5678:
@weinliu Is this something you can review from the QE side?

weinliu:
The doc LGTM

PR #14578, opened 2019-04-24
Remove service mesh from topic map for 3.10 branch.

JStickler:
TP 1 and TP2 were supported on OpenShift 3.10, but we've since had eight more TP releases on OCP 3.11.  We don't want additional customers installing Service Mesh on 3.10, so we'd like to remove the guide from the 3.10 pages.

@vikram-redhat  @geekspertise


JStickler:
The build is failing because the 3.11 release notes reference the Service Mesh docs.  But this PR shouldn't be merged with the 3.11 branch, and the 3.10 release notes do not mention the Service Mesh TP.

kalexand-rh:
If this change applies to only 3.10, I think you should open it against the 3.10 branch instead of master.

PR #14585, opened 2019-04-25
[WIP] Add a note for preferred Neutron backend in OpenStack

eurijon:
Until https://bugzilla.redhat.com/show_bug.cgi?id=169104 is fixed
using ML2/OVN backend is not recommended for Openshift on Openstack
deployments with Kuryr SDN.


install_config/topics/ocp_osp_enabling_octavia.adoc
```diff
@@ -111,3 +111,10 @@ backend is ML2/OVS. There is no need for modifications if the backend is
 ML2/OVN.
 ====
 
+[NOTE]
+====
+If Kuryr SDN is used, ML2/OVS is the preferred Neutron backend at the moment.
```
huffmanca: We should remove `at the moment.` If the recommendation changes then the docs should be updated. So what about:

> ...ML2/OVS is the preferred Neutron backend.

install_config/topics/ocp_osp_enabling_octavia.adoc
```diff
@@ -111,3 +111,10 @@ backend is ML2/OVS. There is no need for modifications if the backend is
...
+ML2/OVN is also supported but due to an issue that impacts the performance
```
huffmanca: I'd reword this sentence. If the creation doesn't timeout, then what about:

> ML2/OVN is also supported. Container creation may take longer than expected due to performance issues.

If container creation eventually does timeout, then I would say:

While ML2/OVN is also supported, there are known issues that can result in a timeout during container creation.

huffmanca (changes requested):
A couple of rewording recommendations.

Also, which versions is this for? Only 3.11, or previous versions as well?

vikram-redhat:
Wait on @mcurry-rh's ack before merging.

install_config/topics/ocp_osp_enabling_octavia.adoc
eurijon: Agree

install_config/topics/ocp_osp_enabling_octavia.adoc
eurijon: Agree

eurijon:
@huffmanca sorry I didn't specify the versions, it is for OCP 3.10 and 3.11
Thank you

PR #14587, opened 2019-04-25
Prevent overwriting the default value of "openshift_master_admission_plugin_config"

bysnupy:
- Version: `v3.10` and `v3.11`

- Description: 
  We should warn about overwriting the `default value` by configured value in order to prevent unexpected result and behavior by missing default value of `admission plugin` config.
If you configure `openshift_master_admission_plugin_config` in inventory file during installation, then `default value` is overwritten.

- Related BZ: [Setting openshift_master_admission_plugin_config overwrites default ImagePolicyConfig](https://bugzilla.redhat.com/show_bug.cgi?id=1702149)

### DIFF
--- a/install/configuring_inventory_file.adoc
+++ b/install/configuring_inventory_file.adoc
@@ -124,6 +124,12 @@ a|This variable sets the parameter and arbitrary JSON values as per the requirem
 ----
 openshift_master_admission_plugin_config={"ClusterResourceOverride":{"configuration":{"apiVersion":"v1","kind":"ClusterResourceOverrideConfig","memoryRequestToLimitPercent":"25","cpuRequestToLimitPercent":"25","limitCPUToMemoryPercent":"200"}}}
 ----
+In this value, `openshift_master_admission_plugin_config={"openshift.io/ImagePolicy":{"configuration":{"apiVersion":"v1","executionRules":[{"matchImageAnnotations":[{"key":"images.openshift.io/deny-execution","value":"true"}],"name":"execution-denied","onResources":[{"resource":"pods"},{"resource":"builds"}],"reject":true,"skipOnResolutionFailure":true}],"kind":"ImagePolicyConfig"}}}` is the default parameter value.
+
+[IMPORTANT]
+====
+You must include the default `openshift_master_admission_plugin_config` value even if you need to add a custom setting.
+====
 
 |`openshift_master_audit_config`
 |This variable enables API service auditing. See


install/configuring_inventory_file.adoc
```diff
@@ -125,6 +125,12 @@ a|This variable sets the parameter and arbitrary JSON values as per the requirem
 openshift_master_admission_plugin_config={"ClusterResourceOverride":{"configuration":{"apiVersion":"v1","kind":"ClusterResourceOverrideConfig","memoryRequestToLimitPercent":"25","cpuRequestToLimitPercent":"25","limitCPUToMemoryPercent":"200"}}}
 ----
 
+[IMPORTANT]
+====
+To prevent overwriting the default value of this parameter,
```
kalexand-rh: Do you mean that if you need to customize this parameter value, that you must specify the default value in addition to your custom value?

What's the default value? If I'm interpreting this issue correctly, I think that we should show both the default and custom values together.
```
----
openshift_master_admission_plugin_config=<both default and current examples together>
-----
In this value, `<default part>` is the default parameter value.

[IMPORTANT]
====
You must include the default `openshift_master_admission_plugin_config` value even if you need to add a custom setting.
====

bysnupy: > Do you mean that if you need to customize this parameter value, that you must specify the default value in addition to your custom value?

Yes, I do. 

> What's the default value? If I'm interpreting this issue correctly, I think that we should show both the default and custom values together.

You're right. It need to provide default value kindly. I thought just simple warning is enough to this issue.

The default value of `openshift_master_admission_plugin_config` is as follows.
Refer the original YAML [here](https://github.com/openshift/openshift-ansible/blob/release-3.11/roles/openshift_master_facts/defaults/main.yml) for more details.
~~~
openshift_master_admission_plugin_config={"openshift.io/ImagePolicy":{"configuration":{"apiVersion":"v1","executionRules":[{"matchImageAnnotations":[{"key":"images.openshift.io/deny-execution","value":"true"}],"name":"execution-denied","onResources":[{"resource":"pods"},{"resource":"builds"}],"reject":true,"skipOnResolutionFailure":true}],"kind":"ImagePolicyConfig"}}}
~~~

bysnupy: Thank you for your review and suggestion ! @kalexand-rh  !
I've changed them through your idea.

kalexand-rh:
@jianlinliu, this is 3.x, but will you PTAL when you can?

jianlinliu:
LGTM.

PR #14589, opened 2019-04-25
Adds ingress nodeselector and tolerations examples

danehans:
Example procedures for configuring ingress controller `nodeSelector` and `tolerations` features.

### DIFF
--- a/modules/nw-ingress-controller-nodeselector.adoc
+++ b/modules/nw-ingress-controller-nodeselector.adoc
@@ -1,0 +1,78 @@
+// Module filename: nw-ingress-controller-nodeselector.adoc
+// Module included in the following assemblies:
+// * networking/ingress-controller-configuration.adoc
+
+[id="nw-ingress-controller-configuration-{context}"]
+= Assigning ingress controller pods to nodes
+
+By default, ingress controller pods are scheduled to worker nodes by using nodeSelectors.
+nodeSelector is the simplest form of node selection constraint and is detailed
+xref:nodes-scheduler-node-selectors-about.adoc[here].
+If the default node selection does not suffice, the IngressController resource supports nodeSelectors
+to constrain ingress controller pods.
+
+.Example Procedure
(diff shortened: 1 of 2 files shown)


danehans:
@kalexand-rh @bmcelvee I was unaware of https://github.com/openshift/openshift-docs/pull/14529 when I was in the middle of this PR. I figured that I would complete the doc and we can figure out how to proceed. I tried linking to content from other assemblies using `xref`, please double-check if my syntax is correct.

bmcelvee:
Thanks, @danehans! :)  @kalexand-rh and I are going to sync up tomorrow to review and make sure we'e clear and not duplicating content.

modules/nw-ingress-controller-nodeselector.adoc
```diff
@@ -0,0 +1,78 @@
...
+xref:nodes-scheduler-node-selectors-about.adoc[here].
+If the default node selection does not suffice, the IngressController resource supports nodeSelectors
+to constrain ingress controller pods.
+
+.Example Procedure
+. One or more labels can be attached to a node. Run `oc get nodes` to get the node names of your cluster. Select one or
+more nodes to attach a label to. Run the following command, replacing `<node-name>` with your desired node name:
++
+----
+$ oc label nodes <node-name> <label-key>=<label-value>
```
kalexand-rh: @danehans, @ingvagabund confirmed that "we do not reconcile labels from a node to a machine, only from a machine to a node."

I can check to see if I can add more context to the module that I'm updating, but I think that we should use the  method that I already have that adds the label to the MachineSet instead.

bmcelvee: Do you all think it's valid to remove the "Assigning ingress controller pods to nodes" module and merge the "Ingress controller taints and tolerations" module?

Thanks!

vikram-redhat:
@bmcelvee @kalexand-rh can this be moved along?

PR #14591, opened 2019-04-25
couple of updates based on beta 4

kalexand-rh:
I installed beta 4 and made a few changes to the installation topic.

### DIFF
--- a/modules/installation-obtaining-installer.adoc
+++ b/modules/installation-obtaining-installer.adoc
@@ -14,7 +14,7 @@ computer.
 
 .Prerequisites
 
-* You must install the cluster from a computer that uses OS X or Linux.
+* You must install the cluster from a computer that uses Linux or macOS.
 * You need 300 MB of local disk space to download the installation media.
 
 .Procedure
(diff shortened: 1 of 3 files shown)


kalexand-rh:
@wking, when you have a minute, do you have any objection to these text changes?

modules/installation-obtaining-installer.adoc
```diff
@@ -23,23 +23,24 @@ computer.
...
+. Download the installation program for your operating system and place the file
+in the directory where you will store the installation configuration files.
 +
 [IMPORTANT]
 ====
 The installation program creates several files on the computer that you use to
-install your cluster. While you can delete the installation program after you
-install a cluster, you must keep the files that the installation program
-creates.
+install your cluster. You must keep both the installation program and the files
```
wking: Well, you can always re-download the installer.  But yeah, keeping it is probably easier.

modules/installation-obtaining-installer.adoc
```diff
@@ -23,23 +23,24 @@ computer.
...
+that the installation program creates after you finish installing the cluster.
 ====
 
+////
 . Set the installation program to be executable:
```
wking: No need to comment this out; just remove it.  The old way is not coming back.

modules/installation-obtaining-installer.adoc
kalexand-rh: If we're sure that people can find the right versions to successfully uninstall or that any version of the installer will work, I'm happy to change this.

PR #14594, opened 2019-04-26
addressing latest curator changes

josefkarasek:


logging/config/efk-logging-curator.adoc
```diff
@@ -5,109 +5,18 @@ include::modules/common-attributes.adoc[]
 
 toc::[]
 
-The Elasticsearch Curator tool performs scheduled maintenance operations on a per-project basis. Curator performs actions daily based on its configuration. Only one Curator Pod is
-recommended per Elasticsearch cluster.
+The Elasticsearch Curator tool performs scheduled maintenance operations
+on a per-project basis. Curator performs actions daily based on its configuration.
+Only one Curator Pod is recommended per Elasticsearch cluster.
```
mburke5678: Does Curator perform operations *daily*? User can set the schedule to run weekly, monthly, etc.

modules/efk-logging-curator-delete-index.adoc
```diff
@@ -9,11 +9,11 @@
 This file can be expanded when more functions are supported. Only delete_indices currently. When more functions are added, we can use include::modules/efk-logging-curator-actions.adoc[leveloffset=+1] for a strict delete index files topic, if needed
 ////
 
-You can configure Curator to delete Elasticsearch data.
+Periodically delete Elasticsearch data based on retention settings.
```
mburke5678: Here the user isn't actually doing the deleting (like `oc delete pod`), but setting how Curator does the deleting? If so, _You can configure Curator to delete..._ is more accurate. Maybe:
_You can configure Curator to delete Elasticsearch data based on retention settings._

logging/config/efk-logging-curator.adoc
josefkarasek: @mburke5678 This is a really good point. In ansible we only exposed `hour` and `minute`. Not the rest.

@ewolinetz @jcantrill CR change for 4.2 to be more in line with 3.x?
For now I'd say we document that users should run curator daily, therefore only modify run_hour and run_minute.

ewolinetz: i'm not sure if that's what we'd want to do. is there really a reason we would let users configure how often curator runs vs what we have as our default now? (which is to run multiple times a day)

mburke5678:
@anpingli Is this something you can review? Documenting Curator for 4.2. Thank you in advance!

anpingli (commented):
LGTM

PR #14595, opened 2019-04-26
Add note about not persisting node labels

mburke5678:
Per [Jan Chaloupka in Slack](https://coreos.slack.com/archives/CDDEUQV8X/p1556289956332400?thread_ts=1556287795.324900&cid=CDDEUQV8X):
_...if you leave a very short note saying "node labels are not persisted after a node is deleted even though the node is backup by a machine". That might do the trick._

### DIFF
--- a/modules/nodes-nodes-working-updating.adoc
+++ b/modules/nodes-nodes-working-updating.adoc
@@ -7,6 +7,12 @@
 
 You can update any label on a node.
 
+[NOTE]
+====
+Node labels are not persisted after a node is deleted even if the node is backed up by a Machine.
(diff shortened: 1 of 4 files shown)


mburke5678:
@kalexand-rh PTAL

modules/nodes-nodes-working-updating.adoc
```diff
@@ -7,6 +7,11 @@
 
 You can update any label on a node.
 
+[NOTE]
+====
+Node labels are not persisted after a node is deleted even if the node is backed up by a Machine.
```
kalexand-rh: I might add something like, "If you need to ensure that the labels persist if the Machine is replaced, add the labels to the MachineSet that controls the Machine instead."

PR #14602, opened 2019-04-26
Remove service mesh manuals from 3.10 topic map/3.10 branch.

JStickler:
TP 1 and TP2 were supported on OpenShift 3.10, but we've since had eight more TP releases on OCP 3.11. We don't want additional customers installing Service Mesh on 3.10, so we'd like to remove the guide from the 3.10 pages.

@vikram-redhat @kalexand-rh @geekspertise

Replaces #14578, which was incorrectly created against the master branch instead of against the 3.10 branch.


kalexand-rh:
This LGTM. Merging.

PR #14609, opened 2019-04-29
Adding note about managed/unmanaged CLO/EO

mburke5678:
Added note that if you set the Elasticsearch operator to unmanaged and leave the Cluster Logging Operator managed, the CLO will overwrite the EO. Based on scrum discussion on 4/29


mburke5678:
@jcantrill @richm @ewolinetz It appears that we do not explicitly tell users to set the EO to unmanged. All of the topics on configuring Elasticsearch are done via the Cluster Logging CR or other methods not related to CRs. We do tell users that they can set EO to unmanged, with a new warning about a managed CLO clobbering the unmanged EO CR. 
Do we want to remove the notion that users can set EO to unmanaged?

PR #14622, opened 2019-04-29
Mention that `masquerade` mode for CNV default network is recommended

booxter:


booxter:
@SchSeba

SchSeba (commented):
The changes looks good to me!

But I think we should open the PR pointing the openshift:enterprise-4.1 branch like this PR
https://github.com/openshift/openshift-docs/pull/14351

@ousleyp can you help us? what is the correct branch for this PR?

ousleyp:
> The changes looks good to me!
> 
> But I think we should open the PR pointing the openshift:enterprise-4.1 branch like this PR
> #14351
> 
> @ousleyp can you help us? what is the correct branch for this PR?

Yes, anything for CNV 2.0 needs to have the base branch enterprise-4.1. You are correct :)

Taking a quick look at this, though, I think it should be a release note instead of a note inside of the masquerade procedure. Not sure if it should be a Known Issue or a New/Changed feature (maybe both?).

@booxter Can you please close this PR and comment on https://github.com/openshift/openshift-docs/pull/14603 with your suggested text? I will tweak it and add it to the release notes when I get the chance. 

(feel free to add additional comments with anything else that should go in the release notes, while you're at it)

Thank you!!

booxter:
@ousleyp I am happy to but I don't necessarily agree that release notes is the way to go. This preference is not for the time being but is meant to be the way for the future. (Ideally, we would forbid `bridge` for default networks, and we have plans to do so, but not for CNV 2.0 due to timing constraints.) So keeping it captured in release notes for a particular release seems incorrect.

booxter:
@ousleyp also, please forgive me for the silly question, but what's the procedure with branches in this repo? Don't we (first) merge patches into master, then maybe cherry-pick them to release branches? Why should we contribute directly to a release branch? Will the changes be back-merged into master at some point?

ousleyp:
> @ousleyp I am happy to but I don't necessarily agree that release notes is the way to go. This preference is not for the time being but is meant to be the way for the future. (Ideally, we would forbid `bridge` for default networks, and we have plans to do so, but not for CNV 2.0 due to timing constraints.) So keeping it captured in release notes for a particular release seems incorrect.

My main concern is that a customer who needs to know that masquerade mode is the recommended mode will not necessarily look at the masquerade procedure itself first. And since it is a significant technical change (we had documented bridge mode in the past, though we did go back and change it to masquerade after the release of 1.4), I think it should be highlighted in the release notes under a new/changed features or technical changes section. We can briefly explain that masquerade mode does not have the limitations of bridge mode, and I don't think we even have to include the BZ to get the point across. We can just summarize the limitations of bridge mode, if necessary. Then, when bridge mode is forbidden, we can make sure to update the release notes again and make our warning in the user's guide a bit firmer. :)

Note that masquerade mode is *not* mentioned in the 1.4 release notes.

After the release note is squared away, I think that it's fine to include a note in the masquerade procedure, too, since it'll be the same going forward. But I don't think we should link to a BZ outside of the release notes (with few exceptions). Maybe it could just say something like "Using `masquerade` mode is the recommended way to connect virtual machines to the default Pod network."

> @ousleyp also, please forgive me for the silly question, but what's the procedure with branches in this repo? Don't we (first) merge patches into master, then maybe cherry-pick them to release branches? Why should we contribute directly to a release branch? Will the changes be back-merged into master at some point?

For 2.0, use enterprise-4.1, but for 1.4, you can use master (in that case, we would cherrypick to enterprise-3.11). Normally we do use master and cherrypick, but there is some added complexity right now due to the modularization effort - a massive change in our doc structure between OCP versions 3.11 and 4.1. CNV 2.0 aligns with OCP 4.1. IIUC, we will eventually use master again. @vikram-redhat - do you have anything to add/correct here?

I hope this makes sense - sorry for the novel! Let me know what your thoughts are. And thanks for helping with the docs :)

booxter:
@ousleyp I think what I hear is that there is value in both having a release note AND the note about recommended way (perhaps without linking to a bug). Thanks for explanation, I will update / propose a new PR that would cover both points above to the correct branch tomorrow.

vikram-redhat:
>  @vikram-redhat - do you have anything to add/correct here?

No that is all correct :)

booxter:
@vikram-redhat I am trying to understand how I can propose the change to enterprise-4.1 branch. Cherry-picks don't work because files that they expand are not present in the branch. For example, this commit with the initial docs for CNV is not there: 0176e8d85a58837c7c8b686b7ab25c67b954b861

Perhaps I miss something, or I should start with backporting all the docs into the branch but this seems weird. Aren't there CNV docs for enterprise-4.1 already? (I see two files only that match *cnv* file naming pattern.)

Please advise.

vikram-redhat:
> @vikram-redhat I am trying to understand how I can propose the change to enterprise-4.1 branch. Cherry-picks don't work because files that they expand are not present in the branch. For example, this commit with the initial docs for CNV is not there: [0176e8d](https://github.com/openshift/openshift-docs/commit/0176e8d85a58837c7c8b686b7ab25c67b954b861)
> 

That is right. That commit is for the old structure for 3.11, for the new structure for 4.1, which has just been added: https://github.com/openshift/openshift-docs/pull/14624, you will need to wait till @ousleyp (or other members of the CNV docs team) has added that file in from 3.11 to 4.1.

> Perhaps I miss something, or I should start with backporting all the docs into the branch but this seems weird. Aren't there CNV docs for enterprise-4.1 already? (I see two files only that match _cnv_ file naming pattern.)
> 
> Please advise.

booxter:
/hold

@ousleyp does CNV docs team have plans to backport the new structure to 4.1 in the near future?

booxter:
@ousleyp could you please chime in here? It's unclear to me what's the plan re: CNV 2.0 docs right now.

ousleyp:
> @ousleyp could you please chime in here? It's unclear to me what's the plan re: CNV 2.0 docs right now.

Oops, sorry - I didn't see your previous tag. As far as I know, we have no plans to backport the 3.11 files in a batch. For every user story, we're manually moving the required files from 3.11 to 4.1 and making changes to those files at the same time so that they fulfill modular documentation requirements.

This is the best reference for 4.1 contribution info: https://github.com/openshift/openshift-docs/blob/enterprise-4.1/contributing_to_docs/doc_guidelines.adoc

HTH!

booxter:
This is covered by another PR.

PR #14397, opened 2019-04-10
ingress: Document the default certificate

Miciah:
Very much a work in progress.

This PR resolves [OSDOCS-356](https://jira.coreos.com/browse/OSDOCS-356).

* `modules/ingress-architecture-default-certificate.adoc`: New file.  Define the default wildcard certificate.
* `modules/ingress-creating-a-reencrypt-route-with-a-custom-certificate.adoc`: New file.  Document how to create a reencrypt route with a custom certificate.
* `modules/ingress-creating-an-edge-route-with-a-custom-certificate.adoc`: New file.  Document how to create an edge route with a custom certificate.
* `modules/ingress-setting-a-custom-default-certificate.adoc`: New file.  Document how to set a custom default certificate.

PR #14623, opened 2019-04-29
PR14397 follow-up put default certificate modules in an assembly

bmcelvee:
Follow-up to https://github.com/openshift/openshift-docs/pull/14397


bmcelvee:
@Miciah - here's the follow-up PR. Would you mind please taking a look, thanks! :)

modules/nw-ingress-setting-a-custom-default-certificate.adoc
```diff
@@ -21,19 +24,24 @@ default   10m
 
 .Procedure
 
-The following assumes that the custom certificate/key pair are in the `tls.crt` and `tls.key` files in the current working directory.  Substitute the actual path names for `tls.crt` and `tls.key`.  You also may substitute another name for the `custom-certs-default` when creating the `Secret` resource and referencing it in the `IngressController` CR.
+The following assumes that the custom certificate and key pair are in the
+`tls.crt` and `tls.key` files in the current working directory. Substitute the
+actual path names for `tls.crt` and `tls.key`. You also may substitute another
+name for the `custom-certs-default` when creating the `Secret` resource and
```
Miciah: The "the" should be deleted (my mistake).

networking/ingress/configuring-default-certificate.adoc
```diff
@@ -0,0 +1,16 @@
...
+
+The following sections describe how to create a re-encrypt route with a custom
+certificate, how to create an edge route with a custom certificate, and how to
+set a custom default certificate.
+
+include::modules/nw-ingress-creating-a-reencrypt-route-with-a-custom-certificate.adoc[leveloffset=+1]
+
+include::modules/nw-ingress-creating-an-edge-route-with-a-custom-certificate.adoc[leveloffset=+1]
+
+include::modules/nw-ingress-setting-a-custom-default-certificate.adoc[leveloffset=+1]
```
Miciah: The first two modules are more for developers (in v3, it would in the "Developer Guide") whereas the third is an administrative task (in v3, "Configuring Clusters" or possibly "Day Two"), but we are not making that distinction in the v4 documentation, right?

I would still put the `nw-ingress-setting-a-custom-default-certificate.adoc` first since it describes configuring the default, which pretty much any administrator will want to do, and the other two modules describe how to use custom certificates instead of the default.  Does that make sense?

bmcelvee: That makes sense. I rearranged the order, and added some context for admin versus dev tasks.

bmcelvee:
Just made some minor follow-up edits.

@lihongan - I'm going ahead and merging this content for the review. It'll be viewable in the rendered docs. Thanks!

PR #14627, opened 2019-04-30
updating autoscaler api versions

kalexand-rh:
per https://github.com/openshift/cluster-autoscaler-operator/tree/master/examples

@vikram-redhat, can I merge?

### DIFF
--- a/modules/cluster-autoscaler-cr.adoc
+++ b/modules/cluster-autoscaler-cr.adoc
@@ -11,7 +11,7 @@ values for the ClusterAutoscaler.
 
 [source,yaml]
 ----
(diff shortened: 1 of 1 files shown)


vikram-redhat:
@kalexand-rh - would this resource definition need changing as well? https://docs.openshift.com/container-platform/4.1/machine_management/applying-autoscaling.html#machine-autoscaler-cr-applying-autoscaling

kalexand-rh:
@vikram-redhat, not according to the samples in that link.

vikram-redhat:
Yikes. I thought that they would have wanted to keep them the same.

kalexand-rh:
¯\_(ツ)_/¯
I'll check again before GA.

PR #3916, opened 2017-03-10
add docs for maximum pods per node

sjenning:
xref
https://bugzilla.redhat.com/show_bug.cgi?id=1430484
https://bugzilla.redhat.com/show_bug.cgi?id=1376585

@derekwaynecarr

PR #14631, opened 2019-04-30
Node section : podsPerCore/maxPods example and descriptive text don't match

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1704249


mburke5678:
@mffiedler PTAL

Does the line that follows the code snippet make sense?
_In the above example, the default value for `podsPerCore` is `10` and the
default value for `maxPods` is `250`. This means that unless the node has 25
cores or more, by default, `podsPerCore` will be the limiting factor._

Should it be _**10** cores or more_?

This paragraph came from @sjenning via https://github.com/openshift/openshift-docs/pull/3916/files

mffiedler (approved):
@mburke5678 Yes, that looks great

PR #14638, opened 2019-05-01
BZ1698640 -- set values for 2 parameters

jhoyt-rh:
Bug **BZ1698640** is here: https://bugzilla.redhat.com/show_bug.cgi?id=1698640

Reported in Rel 3.9; also applies to 3.10 and 3.11, and will need to be cherrypicked to them as well.

### DIFF
--- a/upgrading/automated_upgrades.adoc
+++ b/upgrading/automated_upgrades.adoc
@@ -546,11 +546,25 @@ OpenShift upgrades] for further guidance.
 [[special-considerations-for-mixed-environments]]
 ==== Special considerations for mixed environments
 
-Before you upgrade a mixed environment, for example one with Red Hat Enterprise
+Before you upgrade a mixed environment, such as one with Red Hat Enterprise
 Linux (RHEL) and RHEL Atomic Host, set values in the inventory file
 for both the `openshift_pkg_version` and `openshift_image_tag` parameters.
-Setting these variables ensures that all nodes in your cluster run the same
+Setting these values ensures that all nodes in your cluster run the same
 version of {product-title}.
+
+For example, to upgrade from {product-title}
+3.9 to {product-title} 3.10, set the following parameters and values:
+
+----
+openshift_pkg_version=-3.10.16
+openshift_image_tag=v3.10.16
+----
+
+[NOTE]
+====
+These parameters can also be present in other, non-mixed, environments.
+====
+
 
 [[special-considerations-for-gcepd]]
 ==== Special considerations when using gcePD


upgrading/automated_upgrades.adoc
```diff
@@ -546,12 +546,26 @@ OpenShift upgrades] for further guidance.
...
-Before you upgrade a mixed environment, for example one with Red Hat Enterprise
+Before you upgrade a mixed environment, such as one with Red Hat Enterprise
 Linux (RHEL) and RHEL Atomic Host, set values in the inventory file
 for both the `openshift_pkg_version` and `openshift_image_tag` parameters.
-Setting these variables ensures that all nodes in your cluster run the same
+Setting these values ensures that all nodes in your cluster run the same
 version of {product-title}.
 
+For example, to upgrade from {product-title}
+3.9 to {product-title} 3.10, you would set the values as follows:
```
kalexand-rh: s/you would set the values as follows/set the following parameters and values:

>!  would v
Use only to refer to an unfulfilled condition or hypothesis and to express the future in the past... Do
not use as an auxiliary verb...

upgrading/automated_upgrades.adoc
```diff
@@ -546,12 +546,26 @@ OpenShift upgrades] for further guidance.
...
+
+----
+openshift_pkg_version=v3.10.16
+openshift_image_tag=v3.10.16
+----
+
+[NOTE]
+====
+These options may also be present in other, non-mixed, environments.
```
kalexand-rh: s/options may/parameters might

> ! may v
Use only to indicate permission. Otherwise, use “can” or “might.”
“May” is overused in technical information and should in most instances be changed to “can” or “might.”

kalexand-rh (commented):
Two style picks. Otherwise LGTM.

upgrading/automated_upgrades.adoc
jhoyt-rh: Done. Thanks.

upgrading/automated_upgrades.adoc
jhoyt-rh: Done. Thanks.

jhoyt-rh:
Discussion with QE and Reporter resolved the format of the version parameter. 

Reporter had opened an additional bug [BZ1706991] (https://bugzilla.redhat.com/show_bug.cgi?id=1706991) asking that the format be changed in order to be consistent with other version formats in OpenShift. The engineer explained this was not going to be done because for some OpenShift customers "a unique package version or image would need to be specified, such as hotfixes or private registries, which may not match the simple "v3.x.x" pattern. 

That explanation is not included with this bug's documentation because that bug is not listed as a Documentation bug -- per Vikram's instructions. If the reporter wants this to be added to the docs, he will need to open a Doc bug.

kalexand-rh:
@jhoyt-rh, only 3.11 cherrypicked cleanly. Please open new PRs for 3.10 and 3.9. If you tag me, I'll merge them.

PR #14639, opened 2019-05-01
BZ-1684357: Updates to resource quotas section

bergerhoffer:


bergerhoffer:
Merging as these updates were okayed in the BZ.

PR #14642, opened 2019-05-01
Bug 1704489, removed Optimizing compute resources assembly

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1704489

Removing this section after discussion with the team. The Node Tuning Operator topic covers this.


ahardin-rh:
No QE needed since this is a content removal

PR #14651, opened 2019-05-02
Users cannot modify the replicas of prometheus instances of the cluster monitoring stack

ahardin-rh:
Per team discussion with Frederic Branczyk  and Ashish Kamra

@sjug PTAL


ahardin-rh:
Content removal; no QE review necessary.

PR #14672, opened 2019-05-02
Moving CNV document attributes file into 4.1

ousleyp:
@vikram-redhat - I think we need to include this file in our assemblies as we did in 1.4. Please merge it (quickly, if possible), or let me know if there's a reason we shouldn't use this file in 2.0. Thanks!

### DIFF
--- a/modules/cnv-document-attributes.adoc
+++ b/modules/cnv-document-attributes.adoc
@@ -1,0 +1,34 @@
+// Standard document attributes to be used in the documentation
+//
+// The following are shared by all documents:
+:toc:
+:toclevels: 4
+:experimental:
+//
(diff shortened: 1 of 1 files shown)


vikram-redhat:
@ousleyp - need to change the file name to use dashes instead of underscores.

ousleyp:
> @ousleyp - need to change the file name to use dashes instead of underscores.

Done. I think there are only a few assemblies where we'll need to change the filename, so that shouldn't be a big deal.

ousleyp:
@bgaydosrh @aburdenthehand - FYI, now that this has been merged, any assemblies for 4.1 need to include cnv-document-attributes.adoc, NOT cnv_document_attributes.adoc. :)

PR #14681, opened 2019-05-03
Machines

enxebre:
Machine API revisit:
- Modify machine topics ordering 
- Drop duplicated overview from infra seciton
- Update creating machine set section
- Update machine API overview

### DIFF
--- a/machine_management/creating-infrastructure-machinesets.adoc
+++ b/machine_management/creating-infrastructure-machinesets.adoc
@@ -13,8 +13,6 @@ that are required to run the environment.
 
 include::modules/infrastructure-components.adoc[leveloffset=+1]
 
-include::modules/machine-api-overview.adoc[leveloffset=+1]
-
 [id="creating-infrastructure-machinesets-production"]
 == Creating infrastructure MachineSets for production environments
 
--- a/modules/machineset-creating.adoc
+++ b/modules/machineset-creating.adoc
@@ -5,11 +5,6 @@
 [id="machineset-creating-{context}"]
 = Creating a MachineSet
 
-You can create more MachineSets. Because the MachineSet definition contains
-details that are specific to the Amazon Web Services (AWS) region that the
-cluster is deployed in, you copy an existing MachineSet from your cluster and
-modify it.
-
 .Prerequisites
 
 * Deploy an {product-title} cluster.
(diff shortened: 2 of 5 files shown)


machine_management/creating-infrastructure-machinesets.adoc
```diff
@@ -13,8 +13,6 @@ that are required to run the environment.
 
 include::modules/infrastructure-components.adoc[leveloffset=+1]
 
-include::modules/machine-api-overview.adoc[leveloffset=+1]
-
 [id="creating-infrastructure-machinesets-production"]
```
enxebre: Do we need `creating-infrastructure-machinesets.adoc` section at all?

>You can create a MachineSet to host only infrastructure components.
You apply specific Kubernetes labels to these Machines and then
update the infrastructure components to run on only those Machines.These
infrastructure nodes are not counted toward the total number of subscriptions
that are required to run the environment.

Is taht statement true?
cc @derekwaynecarr @eparis

enxebre:
cc @derekwaynecarr @kalexand-rh @frobware

machine_management/creating-infrastructure-machinesets.adoc
kalexand-rh: @enxebre, last I heard it was true. @derekwaynecarr, will you PTAL?

Making *a* new machineset and making new machinesets for your cluster infra are different user goals, so they need to be separate in our new doc structure.

kalexand-rh:
@sunzhaohua2, will you PTAL?

modules/machineset-creating.adoc
```diff
@@ -18,194 +13,97 @@ permission.
...
-        machine.openshift.io/cluster-api-machine-type: <machine_label>
-        machine.openshift.io/cluster-api-machineset: <cluster_name>-<machine_label>-<AWS-availability-zone> <1>
-...
-----
-<1> Ensure that both the `machine.openshift.io/cluster-api-machineset` parameter values
-match the `name` that you defined in the `metadata` section.
+        machine.openshift.io/cluster-api-cluster: <clusterID>
+        machine.openshift.io/cluster-api-machine-role: <role>
+        machine.openshift.io/cluster-api-machine-type: <role>
+        machine.openshift.io/cluster-api-machineset: <clusterID>-worker-us-east-1a
```
sunzhaohua2: Should be `machine.openshift.io/cluster-api-machineset: <clusterID>-<role>-us-east-1a`

sunzhaohua2:
Others LGTM, thanks!

kalexand-rh:
I merged the follow-up PR. Thanks @enxebre!

machine_management/creating-infrastructure-machinesets.adoc
derekwaynecarr: there isnt anything "special" about an infrastructure machine set.  its a machine set like any other.  i would orient discussion around infrastructure specific hosts with those components that would run on those hosts.

kalexand-rh: Thanks, Derek! Since it's a different user goal, setting up an infrastructure machineset is in a different assembly, but the steps to create the machineset are identical. The infrastructure assembly additionally contains the specific steps to move workloads to the infrastructure machines. https://docs.openshift.com/container-platform/4.1/machine_management/creating-infrastructure-machinesets.html

PR #14572, opened 2019-04-23
Network Operator - OSDOCS-177

jboxman:
Successor to https://github.com/openshift/openshift-docs/pull/14224.

- Removes unused files
- Renames files

PR #14705, opened 2019-05-03
Add Kubernetes network proxy configuration content

jboxman:
Revise https://github.com/openshift/openshift-docs/pull/14572.

This assembly now focuses exclusively on kube-proxy. None of the other configuration parameters can be set post-installation, save `additionalNetworks`, which needs its own assembly.


jboxman:
@openshift/team-documentation PTAL? Thank you!

modules/nw-kube-proxy-about.adoc
```diff
@@ -0,0 +1,9 @@
+// Module included in the following assemblies:
+// * networking/configuring-kubeproxy.adoc
+
+[id="about-kubeproxy-{context}"]
+= About kube-proxy
+
+The Kubernetes network proxy (kube-proxy) runs on each node. It maintains
```
kalexand-rh: Would you need to reuse this module? I think it might be fair to put it in the assembly itself.

modules/nw-kube-proxy-config.adoc
```diff
@@ -0,0 +1,41 @@
...
+
+[id="modifying-kubeproxy-config-{context}"]
+= kube-proxy configuration parameters
+
+You can modify the following `kubeProxyConfig` parameters:
+
+.Parameters
+[cols=".^2,.^5,.^3a",options="header"]
+|====
+|Parameter|Description|Values
```
kalexand-rh: I'd add a column for the default value and clearly break out the description, available values, and default values into the appropriate columns. The information isn't evenly distributed across the table as-is. Please double-check that your descriptions are actually just descriptions instead of the available values.

modules/nw-kube-proxy-config.adoc
```diff
@@ -0,0 +1,41 @@
...
+
+|`bindAddress`
+|The IP address for `kube-proxy` to serve on.
+|The default value is `0.0.0.0`.
```
kalexand-rh: Are there any restrictions for this address or info about how a user's supposed to know the appropriate value?

modules/nw-kube-proxy-config.adoc
```diff
@@ -0,0 +1,41 @@
...
+
+|`proxyArguments.iptables-min-sync-period`
+|The minimum duration before refreshing `iptables` rules. This parameter ensures
+that the refresh does not happen too frequently. Valid suffixes include `s`,
+`m`, and `h` and are described in the
+link:https://golang.org/pkg/time/#ParseDuration[Go time package]
+|`30s`
+
+|`proxyArguments.proxy-mode`
+|The implmentation for synchronizing network rules.
```
kalexand-rh: implementation

Please provide a little more info about these parameter values. I'm not sure why I'd pick one over another.

modules/nw-kube-proxy-config.adoc
```diff
@@ -0,0 +1,41 @@
...
+|The `proxy-mode` can be either `iptables` or `userspace`. The default value is
+`iptables`.
+
+|====
+
+For a complete list of the parameters that `proxyArguments` supports refer to
```
kalexand-rh: supports.

kalexand-rh: So, are you saying that you can use all of the available proxyArgument parameters? That's not clear.

If you can use all of the proxyArguments, consider splitting the table so that you can provide information about the proxyArgument parameters as a whole, including the link, and then provide examples. If these are required or most important, make that clear.

If they're not all available, why do you need this link?

modules/nw-kube-proxy-configuring.adoc
```diff
@@ -0,0 +1,88 @@
...
+editor. If your modifications contain a syntax error, the editor opens the file
+and displays an error message.
+
+. Run the following command to confirm the configuration update:
++
+----
+$ oc get networks.operator.openshift.io -o yaml
+----
++
+The command will return output similar to the following example:
```
kalexand-rh: s/will return/returns

modules/nw-kube-proxy-mode.adoc
```diff
@@ -0,0 +1,18 @@
+// Module included in the following assemblies:
+// * networking/configuring-kubeproxy.adoc
+
+[id="modifying-kubeproxy-proxy-mode-{context}"]
+= About service proxy mode
+
+{product-title} has two different implementations of the service-routing
+infrastructure. The default implementation is entirely iptables-based, and uses
```
kalexand-rh: no comma

modules/nw-kube-proxy-mode.adoc
```diff
@@ -0,0 +1,18 @@
...
+probabilistic iptables rewriting rules to distribute incoming service
+connections between the endpoint Pods. The older implementation uses a user
+space process to accept incoming connections and then proxy traffic between the
+client and one of the endpoint Pods.
+
+The iptables-based implementation is much more efficient, but it requires that
+all endpoints are always able to accept connections; the user space
```
kalexand-rh: s/connections; the/connections. The
The sentence is too long.

modules/nw-kube-proxy-mode.adoc
```diff
@@ -0,0 +1,18 @@
...
+implementation is slower, but can try multiple endpoints in turn until it finds
```
kalexand-rh: no comma

modules/nw-kube-proxy-sync.adoc
```diff
@@ -0,0 +1,14 @@
+// Module included in the following assemblies:
+// * networking/configuring-kubeproxy.adoc
+
+[id="modifying-kubeproxy-sync-{context}"]
+= About iptables rules synchronization
+
+The synchronization period determines how frequently the Kubernetes network
+proxy (kube-proxy) syncs the iptables rules on a node.
+
+A sync is triggered when either of the following occurs:
```
kalexand-rh: s/A sync is triggered/A sync begins
^ Use a simple tense. I question if "triggered" will translate well.

s/following/following events

modules/nw-kube-proxy-sync.adoc
```diff
@@ -0,0 +1,14 @@
...
+
+* An event occurs, such as service or endpoint is added or removed from the
```
kalexand-rh: added to

jboxman:
@kalexand-rh, thanks!

modules/nw-kube-proxy-config.adoc
jboxman: @kalexand-rh, do you mean a separate table that covers only `proxyArguments`?

@squeed, @danwinship: Do you know which `proxyArguments` we want to explicitly expose in the docs? Right now we have `proxy-mode` and `iptables-min-sync-period`. By my count there are _40_ parameters for kube-proxy.

Thanks!

kalexand-rh: @boxman, that depends. If you just have the two, you could complicate the table to cover them. If you need to cover all 40, which I hope you don't need to, use a separate table.

danwinship: We want to document whatever we documented in 3.x. That might be just proxy-mode and iptables-min-sync-period?

I don't think we need to point to the "complete list" of parameters. While it's true that people *can* override all of those options, we definitely don't actually *want* them to...

kalexand-rh: @danwinship, which ones do you *want* them to override? Just the ones from 3.x? If you *can* do it, but we don't want them to, we should include a disclaimer that "while other proxyArguments are available, only changing the specified parameters is supported" or whatever the restriction is.

danwinship: I think changing any of the options is technically supported, it's just not recommended. There are a few that most customers would never need to use, but a few have legitimate use cases for. But there are also some that are mostly just useful for shooting yourself in the foot with, so we don't want to be actively pointing people toward them.

squeed: Realistically the only one I'd like to see documented is "iptables-min-sync-period". We don't support changing `proxy-mode` anymore. Also, we will be moving away kube-proxy args and start using the kube-proxy configuration object.

modules/nw-kube-proxy-configuring.adoc
```diff
@@ -43,12 +43,45 @@ The syntax is validated by the `oc` command when you save the file and exit the
...
+    name: cluster
+  spec:
+    clusterNetwork:
+    - cidr: 10.128.0.0/14
+      hostPrefix: 23
+    defaultNetwork:
+      type: OpenShiftSDN
+    deployKubeProxy: false
+    kubeProxyConfig:
+      bindAddress: 0.0.0.0
```
zhaozhanqi: deployKubeProxy: false
bindAddress:0.0.0.0 
those two are also need to remove.

modules/nw-kube-proxy-config.adoc
jboxman: Removing reference entirely.

modules/nw-kube-proxy-config.adoc
jboxman: No longer exposing this parameter.

modules/nw-kube-proxy-config.adoc
jboxman: No longer exposing this parameter.

modules/nw-kube-proxy-configuring.adoc
jboxman: Removed.

jboxman:
I'll merge this after https://github.com/openshift/openshift-docs/pull/14936 is merged so the CNO topic doesn't completely disappear for some period of time.

jboxman:
Verified on bug. I'm going to merge this and update the TOC tree afterwards.

PR #14711, opened 2019-05-06
Removing unnecessary file

jboxman:
- The content lives in modules/nw-configuring-route-timeouts.adoc


kalexand-rh (commented):
LGTM, then!

PR #14727, opened 2019-05-06
Move networking optimization content to installation topic

ahardin-rh:


ahardin-rh:
@kalexand-rh FYI; I am discussing this change with the Autotuning & Scalability (ATS) team. Please let me know if you have any concerns.

modules/installation-aws-config-yaml.adoc
```diff
@@ -82,3 +82,8 @@ provides the default value.
 machines in your cluster.
 <4> To configure faster storage for etcd, especially for larger clusters, set the
 storage type as `io1` and set `iops` to `2000`.
+<5> For a CIDR in the range of 10.128.0.0/14, IPs in this range are, by default,
```
kalexand-rh: s/IPs/IP addresses

modules/installation-aws-config-yaml.adoc
```diff
@@ -82,3 +82,8 @@ provides the default value.
...
+ assigned to the Pods. If you want to change this, you must adjust the
```
kalexand-rh: What would you change this to, and how would a user know what to change to get what they wanted?

modules/installation-aws-config-yaml.adoc
```diff
@@ -82,3 +82,8 @@ provides the default value.
...
+ `ClusterNetwork`.
+ <6> To customize the IPs rolled out to services, then you must adjust the
```
kalexand-rh: s/IPs/IP addresses

Can we have another sentence or two of context here?

kalexand-rh (commented):
I think this is the right place to put this information but that a little more context might help.

modules/installation-aws-config-yaml.adoc
ahardin-rh: @Deepthidharwar Any suggestions?

jboxman:
This conceptual module is useful, I think.

The procedure itself is somewhat of a duplicate of [Installing a cluster on AWS with network customizations](https://docs.openshift.com/container-platform/4.1/installing/installing_aws/installing-aws-network-customizations.html).

You can customize your service network and your cluster network by modifying the `installation-config.yaml`.

More specifics on [clusterNetwork and serviceNetwork](https://docs.openshift.com/container-platform/4.1/installing/installing_aws/installing-aws-network-customizations.html#network-customization-config-yaml-installing-aws-network-customizations) are on that page as well.

Does it make sense to discuss all of the potential customizations to `installation-config.yaml` in the same place and cover advanced network configuration separately?

When changing network configuration parameters, the ones specified in `installation-config.yaml` are the source of truth for the advanced configuration, so anyone wanting to modify those parameters and make additional changes that cannot be done in that file _must_ first customize that file, and then create manifests using the installer to make changes unsupported in `installation-config.yaml`.

So advanced network customization needs the content for regular customization, too.

ahardin-rh:
@jboxman Yeah, the purpose of this PR is to remove the networking optimization content from the performance and scalability section and place the meat of the content into the customizations to installation-config.yaml content. :+1:

ahardin-rh:
@ashishkamra @Deepthidharwar PTAL

Deepthidharwar:
I agree with #jboxman, almost all the of network optimisation content is part of network customisation.  We need it that way for 4.0
We have ability to modify CIDRs for pod IP, service IPs as part of install-config.yaml and we also have VXLAN, port and mode as part of Network CR. So it is good to remove it out and avoid duplication.

modules/installation-aws-config-yaml.adoc
Deepthidharwar: It's basically a  block of IP addresses one reserves, from which Pod IPs are allocated. If one wants Pod IPs to be different from what is reserved by default, then one needs to modify this.

modules/installation-aws-config-yaml.adoc
Deepthidharwar: A block of IP addresses for services. OpenShiftSDN allows only one serviceNetwork block. The address block must not overlap with any other network block. 
Given that, this is all a part of install procedure currently and where much documented, we do not want to duplicate these steps.

ahardin-rh:
@jboxman @Deepthidharwar Ah, I now see what you mean. Sorry, I think I was looking in a wrong file/outdated build. :woman_facepalming: I agree and updated the PR. Deepthi pointed out that install-cluster.yaml should probably be install-config.yaml in one instance, so I updated that here. PTAL. Thanks!

PR #14737, opened 2019-05-07
A NetworkPolicy object can be modified without admin

jboxman:
- https://bugzilla.redhat.com/show_bug.cgi?id=1707280


jboxman:
Verified in the bug. Merging.

PR #14743, opened 2019-05-07
Use proper command prompt symbol in example

jboxman:
- https://bugzilla.redhat.com/show_bug.cgi?id=1707224


jboxman:
Verified in the bug. Merging.

PR #14744, opened 2019-05-07
NetworkPolicy supports combined podSelector and namespaceSelector

jboxman:
- https://bugzilla.redhat.com/show_bug.cgi?id=1707265

### DIFF
--- a/modules/nw-networkpolicy-about.adoc
+++ b/modules/nw-networkpolicy-about.adoc
@@ -11,9 +11,8 @@ the OpenShift SDN plug-in for the Kubernetes Container Network Interface (CNI).
 
 [NOTE]
 ====
-The `v1` NetworkPolicy features are available only in {product-title}. This
-means that egress policy types, IPBlock, and combining `podSelector` and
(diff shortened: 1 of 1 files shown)


jboxman:
@openshift/team-documentation, is there a better way to phrase this? Or does it make sense to list only v1 features we support instead?

modules/nw-networkpolicy-about.adoc
```diff
@@ -11,9 +11,8 @@ the OpenShift SDN plug-in for the Kubernetes Container Network Interface (CNI).
 
 [NOTE]
 ====
-The `v1` NetworkPolicy features are available only in {product-title}. This
-means that egress policy types, IPBlock, and combining `podSelector` and
-`namespaceSelector` are not available in {product-title}.
+The `v1` NetworkPolicy features are available in {product-title} except for
```
kalexand-rh: What v1 network policies? I assume it's the Kubernetes policies, but it's not clear. Please be more explicit.

kalexand-rh (commented):
One request. Otherwise LGTM

PR #14759, opened 2019-05-08
bug 1707678 merging networking requirement PRs

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1707678

@jianlinliu, I've merged the topics. Please take a look?


modules/installation-network-upi.adoc
```diff
@@ -4,7 +4,20 @@
 // * installing/installing_vsphere/installing-vsphere.adoc
 
 [id="installation-network-upi-{context}"]
-= UPI network topology requirements
+= UPI networking requirements
```
jianlinliu: Personally I prefer to merge "UPI networking requirements" and "Connectivity between machines" as one section, their context is closely relative. Is that reasonable?

kalexand-rh: Sure! I moved the table to this section. Is this better?

PR #14554, opened 2019-04-22
bug 1669995 you need the metadata.json file to uninstall

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1669995

@halfsquatch, will you PTAL?

PR #14760, opened 2019-05-08
install/uninstall updates

kalexand-rh:
I am not confident that https://bugzilla.redhat.com/show_bug.cgi?id=1683019 will be fixed in 4.1, so I'm pulling over some text improvements from that bug that are valid in 4.1.

@wking, @jianlinliu, will you PTAL?


kalexand-rh:
Merging with https://github.com/openshift/openshift-docs/pull/14554

PR #14762, opened 2019-05-08
sdn: document how to allow access from ingress

squeed:
Now that the ingress namespace is labeled appropriately, we can select for it.

### DIFF
--- a/modules/nw-networkpolicy-about.adoc
+++ b/modules/nw-networkpolicy-about.adoc
@@ -47,9 +47,6 @@ spec:
   podSelector:
   ingress: []
 ----
-////
-
-// Blocked on https://github.com/openshift/cluster-ingress-operator/pull/218
 
 * Only allow connections from the {product-title} ingress router:
 +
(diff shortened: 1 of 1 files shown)


squeed:
@jboxman PTAL

jboxman:
@squeed LGTM, thanks!

jboxman:
@zhaozhanqi, can you take a look at the `NetworkPolicy` object for ingress in this PR and confirm it is correct? Thanks!

zhaozhanqi:
@squeed , I remembered the openshift-ingress namespaces had been the label `name=openshift-ingress`, why we still need to update to `network.openshift.io/policy-group: ingress`?

squeed:
@zhaozhanqi good question - If we tell customers to use "name: openshift-ingress`, then we can never change the name. By using a different label, it gives us flexibility in the futurel

zhaozhanqi:
@squeed ok, thanks.  then LGTM

PR #14764, opened 2019-05-08
Add cluster limits for OpenShift 4.1

chaitanyaenr:
This commit:
- Updates the limits page for OpenShift 4.1 based on the large scale
  test run.
- Adds a note about the single dimentional analysis we used for
  determining/validating some of the limits.

### DIFF
--- a/modules/how-to-plan-your-environment-according-to-cluster-limits.adoc
+++ b/modules/how-to-plan-your-environment-according-to-cluster-limits.adoc
@@ -12,6 +12,12 @@ Kubernetes scheduler makes during pod placement. Learn what measures you can
 take to avoid memory swapping.
 ====
 
+[IMPORTANT]
+====
+Some of the limits are stretched only in a single dimension. They will vary
+when many objects are running on the cluster.
+====
+
 While planning your environment, determine how many pods are expected to fit per
 node:
 
--- a/modules/openshift-cluster-limits.adoc
+++ b/modules/openshift-cluster-limits.adoc
@@ -13,71 +13,72 @@
 | 2,000
 | 2,000
 | 2,000
-|TBD
+| 250
 
 | Number of pods footnoteref:[numberofpods,The pod count displayed here is the number of test pods. The actual number of pods depends on the application’s memory, CPU, and storage requirements.]
 | 120,000
 | 150,000
 | 150,000
-|TBD
+| 150,000
 
 | Number of pods per node
 | 250
 | 250
 | 250
-|TBD
+| 250
 
 | Number of pods per core
 | 10 is the default value. The maximum supported value is the number of pods per node.
 | There is no default value. The maximum supported value is the number of pods per node.
 | There is no default value. The maximum supported value is the number of pods per node.
-|TBD
+| There is no default value. The maximum supported value is the number of pods per node.
 
-| Number of namespaces
+| Number of namespaces footnoteref:[numberofnamepaces, When there are a large number of active projects, etcd may suffer from poor performance if the keyspace grows excessively large and exceeds the space quota. Periodic maintenance of etcd, including defragmentaion, is highly recommended to free etcd storage.]
 | 10,000
 | 10,000
 | 10,000
-|TBD
+| 8,000
 
 | Number of builds: Pipeline Strategy
 | 10,000 (Default pod RAM 512 Mi)
 | 10,000 (Default pod RAM 512 Mi)
 | 10,000 (Default pod RAM 512 Mi)
-|TBD
+| 10,000 (Default pod RAM 512 Mi)
 
 | Number of pods per namespace footnoteref:[objectpernamespace,There are
 a number of control loops in the system that must iterate over all objects
 in a given namespace as a reaction to some changes in state. Having a large
 number of objects of a given type in a single namespace can make those loops
-expensive and slow down processing given state changes.]
+expensive and slow down processing given state changes. The limit assumes that 
+the system has enough CPU, memory, and disk to satisfy the application requirements.]
 | 3,000
 | 3,000
-| 3,000
-|TBD
+| 25,000
+| 25,000
 
 | Number of services footnoteref:[servicesandendpoints,Each service port and each service back-end has a corresponding entry in iptables. The number of back-ends of a given service impact the size of the endpoints objects, which impacts the size of data that is being sent all over the system.]
 | 10,000
 | 10,000
 | 10,000
-|TBD
+| 10,000
 
 | Number of services per namespace
 | N/A
 | 5,000
 | 5,000
-|TBD
+| 5,000
 
 | Number of back-ends per service
 | 5,000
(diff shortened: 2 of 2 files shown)


chaitanyaenr:
@ahardin-rh @akrzos @ashishkamra @mffiedler PTAL.

modules/how-to-plan-your-environment-according-to-cluster-limits.adoc
```diff
@@ -12,6 +12,11 @@ Kubernetes scheduler makes during pod placement. Learn what measures you can
 take to avoid memory swapping.
 ====
 
+====
```
ahardin-rh: Add [IMPORTANT] or [NOTE] before the first ==== to make it  a note box. You could also make this message part of the IMPORTANT note box just before it to avoid having 2 boxes in a row.

ahardin-rh:
@chaitanyaenr Just one comment from me. Otherwise, looks good! :+1:

chaitanyaenr:
We hit some etcd issues at 9000 projects when running the cluster limits tests to validate the namespaces per cluster limit (10,000). Here are the links to the filed BZ's - https://bugzilla.redhat.com/show_bug.cgi?id=1703581, https://bugzilla.redhat.com/show_bug.cgi?id=1706103. One of the issues was that the etcd space quota was too low and we were able to get past that after running defragmentation and bumping up the quota. The other issue was that the cluster went down when one of the etcd was down and also etcd was using lots of system resources at the same time. 

We tested this on the beta4 build and I believe that the recent builds have the fixes for most of the issues.

chaitanyaenr:
@ahardin-rh added [Important] before the recommendation like you suggested, thanks for the review.

modules/how-to-plan-your-environment-according-to-cluster-limits.adoc
```diff
@@ -12,6 +12,12 @@ Kubernetes scheduler makes during pod placement. Learn what measures you can
 take to avoid memory swapping.
 ====
 
+[IMPORTANT]
+====
+Some of the limits are stretched only in a single dimension, so they might vary
+when a lot of objects are running on the cluster.
```
akrzos: I think the verbage here should change:
```
Some of the limits are stretched only in a single dimension, so they might vary
when a lot of objects are running on the cluster.
```
to
```
Some of the limits are stretched only in a single dimension.  They will vary
when many objects are running on the cluster.
```

akrzos:
> We hit some etcd issues at 9000 projects when running the cluster limits tests to validate the namespaces per cluster limit (10,000). Here are the links to the filed BZ's - https://bugzilla.redhat.com/show_bug.cgi?id=1703581, https://bugzilla.redhat.com/show_bug.cgi?id=1706103. One of the issues was that the etcd space quota was too low and we were able to get past that after running defragmentation and bumping up the quota. The other issue was that the cluster went down when one of the etcd was down and also etcd was using lots of system resources at the same time.
> 
> We tested this on the beta4 build and I believe that the recent builds have the fixes for most of the issues.

I think we need to drop the namespace limit to what was achieved although the build did not have the fixes that might make this achievable.

ahardin-rh:
@vikram-redhat This maps to https://jira.coreos.com/browse/OSDOCS-42 

@chaitanyaenr Once we get QE approval from @mffiedler we can get this merged in. Thanks!

chaitanyaenr:
Bumped down the namespaces per cluster limit to 8000 until we test the previous limit on a large scale cluster again with the fixed build.

modules/openshift-cluster-limits.adoc
```diff
@@ -13,71 +13,72 @@
 | 2,000
 | 2,000
 | 2,000
-|TBD
+| 250
```
warmchang: Is it really 250? Not 2500?

chaitanyaenr: @warmchang we lowered the value to 250 as it's the maximum node scale we tested for OCP 4.1. We can bump it up to 2000 if and when we get an opportunity to run tests at a higher node count.

warmchang: @chaitanyaenr Got it, thank you for reply!

ahardin-rh:
@mffiedler PTAL

modules/openshift-cluster-limits.adoc
```diff
@@ -13,71 +13,72 @@
...
 
 | Number of pods per core
 | 10 is the default value. The maximum supported value is the number of pods per node.
 | There is no default value. The maximum supported value is the number of pods per node.
 | There is no default value. The maximum supported value is the number of pods per node.
-|TBD
+| There is no default value. The maximum supported value is the number of pods per node.
 
-| Number of namespaces
+| Number of namespaces footnoteref:[numberofnamepaces, When there are large number of active projects with objects, etcd may suffer from poor performance if the keyspace grows excessively large and exceeds the space quota. Periodic maintaince of etcd including defragmentaion is highly recommended to free the space for the backend.]
```
mffiedler: nit:  "When there are a large number of active projects" 
spelling:  "maintenance"   Suggested:  "Periodic maintenance of etcd, including defragmentation, is highly recommended to free storage in etcd"

mffiedler (changes requested):

modules/openshift-cluster-limits.adoc
chaitanyaenr: Fixed, thanks.

modules/openshift-cluster-limits.adoc
```diff
@@ -13,71 +13,72 @@
...
 
 | Number of pods per core
 | 10 is the default value. The maximum supported value is the number of pods per node.
 | There is no default value. The maximum supported value is the number of pods per node.
 | There is no default value. The maximum supported value is the number of pods per node.
-|TBD
+| There is no default value. The maximum supported value is the number of pods per node.
 
-| Number of namespaces
+| Number of namespaces footnoteref:[numberofnamepaces, When there are large number of active projects, etcd may suffer from poor performance if the keyspace grows excessively large and exceeds the space quota. Periodic maintenance of etcd, including defragmentaion, is highly recommended to free etcd storage.]
```
mffiedler: I think you missed my first one:   s/When there are large number/When there are a large number/

modules/openshift-cluster-limits.adoc
ahardin-rh: Thanks! Merging! If we need further edits, we can add them in a separate PR.

ahardin-rh: Spoke too soon. See Mike's latest comment.

mffiedler (approved):
/lgtm

modules/openshift-cluster-limits.adoc
chaitanyaenr: Fixed, thanks.

chaitanyaenr:
Thanks for the review.

PR #14769, opened 2019-05-09
Use --type=merge when running oc patch command

jboxman:
- https://bugzilla.redhat.com/show_bug.cgi?id=1707744


jboxman:
Verified in the bug. Merging.

PR #14786, opened 2019-05-09
Simplify custom ingress certificate procedure

ironcladlou:
Use `oc patch` to simplify the procedure. Make the rollout note more
visible.

### DIFF
--- a/modules/nw-ingress-setting-a-custom-default-certificate.adoc
+++ b/modules/nw-ingress-setting-a-custom-default-certificate.adoc
@@ -18,7 +18,7 @@ ingress domain.
 * You must have an `IngressController` CR. You may use the default one:
 +
 ----
-$ oc -n openshift-ingress-operator get ingresscontrollers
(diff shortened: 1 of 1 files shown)


Miciah:
What do you think of keeping the yaml source for illustrative purposes ("If you do an `oc get`, the resource definition should now look like this:")? If you grep for "should look", you can see other instances of the pattern of giving a command and then showing the yaml definition of the result.

ironcladlou:
>What do you think of keeping the yaml source for illustrative purposes ("If you do an oc get, the resource definition should now look like this:")? If you grep for "should look", you can see other instances of the pattern of giving a command and then showing the yaml definition of the result.

Implemented in 6b4bd6636, PTAL

bmcelvee:
@lihongan PTAL, thanks!

lihongan:
LGTM - thanks

bmcelvee:
Thanks!

@ironcladlou - would you mind squashing the two commits into one? Then I can merge. Thanks!

PR #14704, opened 2019-05-03
Bug 1704653 remove routing from edge load balancers content

bmcelvee:
https://bugzilla.redhat.com/show_bug.cgi?id=1704653

PR #14788, opened 2019-05-09
Remove routing-from-edge-load-balancer

Miciah:
Remove outdated content on routing directly from a load balancer to pods.  OCP 4.1 only runs on RHEL CoreOS, so it is not possible to install OCP on a load balancer.

Follow-up to https://github.com/openshift/openshift-docs/pull/14704.

* `_topic_map.yml`: Delete entry for `routing-from-edge-load-balancer`.
* `modules/nw-include-load-balancer.adoc`:
* `networking/ingress/routing-from-edge-load-balancer.adoc`: Delete files.


bmcelvee:
Thanks, @Miciah! Since this is just removing outdated content I don't think it needs QE review. I'll go ahead and merge.

PR #14789, opened 2019-05-09
NetworkPolicy objects should use apiVersion: networking.k8s.io/v1

jboxman:
- https://bugzilla.redhat.com/show_bug.cgi?id=1707289


jboxman:
Verified in bug. Merging.

PR #14790, opened 2019-05-09
OSD/OSO/ARO does _not_ support all those plugins.

wgordon17:
It's cleaner to just remove


wgordon17:
@ahardin-rh Would you mind merging this please?

PR #14796, opened 2019-05-09
Update storage.adoc for OSD/ARO 

anjalitelang:
Removing reference of GCP/GCE disks from Dedicated docs


anjalitelang:
@ahardin-rh Please review

PR #14801, opened 2019-05-09
Re-add NetworkPolicy to the docs for OSD

wgordon17:


wgordon17:
@ahardin-rh Would you mind reviewing? I'm pretty sure we enabled these docs in the past, but it looks like at some point we must've lost them

admin_guide/managing_networking.adoc
```diff
@@ -1364,6 +1364,8 @@ over multicast even when they can't communicate over unicast.)
 . Multicast packets sent by a pod in one project will never be delivered to pods
 in any other project, even if there are `NetworkPolicy` objects allowing
 communication between the to projects.
+endif::openshift-enterprise,openshift-origin[]
```
ahardin-rh: @wgordon17 Looks like this is tripping up the build because it's missing its opening ifdef::openshift-enterprise,openshift-origin[]

ahardin-rh: @wgordon17 Looks like this is tripping up the build because it's missing its opening ifdef::openshift-enterprise,openshift-origin[]

wgordon17: I think I found this weird dangling `endif` at https://github.com/openshift/openshift-docs/pull/14801/files#diff-b579ab33a365f76f12b0225de1d4234bL1247. Not really sure how that was there despite the other stuff not showing up, but lets see what happens without it 👍

wgordon17:
@ahardin-rh Ok, it looks like I got it fixed 👍 But apparently I don't get a preview since I'm not a team member 😞

ahardin-rh:
@wgordon17 The preview bot must be down. LGTM though. Do you mind squashing the 3 commits into one so that I can cleanly cherrypick the change once it merges? Thank you!

wgordon17:
@ahardin-rh done, thanks 👍

wgordon17:
Btw, I don't think it's down, I just don't think it likes me

`wgordon17 is not a team member of @openshift/team-documentation, not building a preview.`

Looks like I'd have to be added to https://github.com/openshift/openshift-docs/blob/master/autopreview.sh#L4 in order for it build previews for me 🤷‍♂

PR #14802, opened 2019-05-09
osdocs389 disabling hyperthreading

kalexand-rh:
https://jira.coreos.com/browse/OSDOCS-389

@abhinavdahiya, will you PTAL?

I don't see where you'd set this parameter for bare metal or vsphere. And I don't know why anyone would want to disable this setting.

### DIFF
--- a/modules/installation-aws-config-yaml.adoc
+++ b/modules/installation-aws-config-yaml.adoc
@@ -27,7 +27,8 @@ This sample YAML file is provided for reference only. You must obtain your
 apiVersion: v1
 baseDomain: example.com <1>
 controlPlane: <2>
-  name: master <3>
+- hyperthreading: Enabled <3> <4>
+  name: master
   platform:
     aws:
       zones:
@@ -37,16 +38,17 @@ controlPlane: <2>
         iops: 4000
         size: 500
         type: io1
-      type: m5.xlarge
+      type: m5.xlarge <4>
   replicas: 3
 compute: <2>
-- name: worker <3>
+  hyperthreading: Enabled <3>
+  name: worker
   platform:
     aws:
       rootVolume:
         iops: 2000
         size: 500
-        type: io1 <4>
+        type: io1 <5>
       type: c5.4xlarge
       zones:
       - us-west-2c
@@ -73,7 +75,7 @@ platform:
       adminContact: jdoe
       costCenter: 7536
 pullSecret: '{"auths": ...}' <1>
-sshKey: ssh-ed25519 AAAA... <5>
+sshKey: ssh-ed25519 AAAA... <6>
 ----
 <1> Required. The installation program prompts you for this value.
 <2> If you do not provide these parameters and values, the installation program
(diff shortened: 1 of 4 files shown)


modules/installation-aws-config-yaml.adoc
```diff
@@ -27,7 +27,8 @@ This sample YAML file is provided for reference only. You must obtain your
 apiVersion: v1beta3
 baseDomain: example.com <1>
 controlPlane: <2>
-- name: master
+- hyperthreading: Enabled
```
abhinavdahiya: hyperthreading is Enabled by default and we shouldn't require all install-config.yaml examples to include that.
only the disabled hyperthreading example must include this field being set.

kalexand-rh: @abhinavdahiya, I sent you a longer explanation in a Slack DM. We're handling all of the valid changes to the `install-config.yaml` file at once for each installation type. The installation assemblies are supposed to cover all of the installation tasks, and since this setting has to be set during installation, we can't put it somewhere else.

rpattath: @kalexand-rh When HT is disabled aws instance should be configured to have larger instance size as in:

hyperthreading: Enabled
  name: worker
  platform: {}
  replicas: 2
controlPlane:
  hyperthreading: Disabled
  name: master
  platform:
    aws:
      type: m5.2xlarge
  replicas: 2

modules/installation-vsphere-config-yaml.adoc
```diff
@@ -13,35 +13,53 @@ parameters.
...
+`hyperthreading`. By default, simultaneous multithreading is enabled
+to increase the performance of your machines' cores. You can disable it by
+setting the parameter value to `Disabled`. If you disable simultanous
+multithreading in some cluster machines, you must disable it in all cluster
+machines.
++
+[IMPORTANT]
+====
+If you disable simultaneous multithreading, ensure that your capacity planning
+accounts for the dramatically decreased machine performance.
```
rpattath: For HT disabled scenarion there would be a memory and CPU requirement of at least

num_cpus: "8"
memory: "32000"
This configuration worked for vsphere.

rpattath (approved):

kalexand-rh:
@rpattath, thank you for the comments! I've made updates based on them in the annotations about hyperthreading. Does this change look ok to you now?

kalexand-rh:
@rpattath, I'm taking your earlier approval as final approval. If you see more issues with the content, I'll open a follow-up PR. Thanks!

modules/installation-aws-config-yaml.adoc
```diff
@@ -85,9 +87,23 @@ array. Because the sections have different definitions, the first line of the
 single machine pool, it is possible that future versions of {product-title}
 will support defining multiple compute pools during installation. Only one
 control plane pool is used.
-<4> To configure faster storage for etcd, especially for larger clusters, set the
+<4> Whether to enable or disable simultaneous multithreading, or
+`hyperthreading`. By default, simultaneous multithreading is enabled
+to increase the performance of your machines' cores. You can disable it by
+setting the parameter value to `Disabled`. If you disable simultanous
+multithreading in some cluster machines, you must disable it in all cluster
+machines.
```
wking: Do you know the source of this all-or-none condition?  I can't see why it couldn't be disabled on a per-machine-pool basis.

kalexand-rh: Before 4.1 GA, Abhinav said "machine-config-operator doesn't support upgrading such a change {after installation} now" in a DM. If you know that this is wrong because of you can disable it on a per-machine-pool-basis after installation in 4.2 or want to expose creating more machines at installation, please file a bug.

PR #14806, opened 2019-05-09
Add a link to AMD GPU device plugin

y2kenny:


vikram-redhat:
@y2kenny can you give more details about this request?

y2kenny:
This is just to add another example of device-plugin available to OpenShift.

y2kenny:
Are there any issue with this commit?

vikram-redhat:
@y2kenny is this supported by Red Hat? @sferich888?

sferich888:
No it's not (supported by Red Hat - the plug-in that is, Red Hat doesn't have a supported plug-in). With that said I Red Hat supports the capability of using such a plug-in.

This link is like line 51 and is an example of plug-in a user could use.

PR #14809, opened 2019-05-10
BZ1698198--remove oc label command to remove daemonset label

jhoyt-rh:


jhoyt-rh:
Closing this PR due to it being the BZ being a long time in review, and the original branch is not current/available. I am opening a new PR for the latest change, which includes the name of the playbook and the xrref.

PR #14436, opened 2019-04-12
doc oc adm must-gather

sanchezl:
Add new `oc adm must-gather` administrative command.

PR #14766, opened 2019-05-08
osdocs392 bootstrap gather draft

kalexand-rh:
https://jira.coreos.com/browse/OSDOCS-392

built draft here: http://file.rdu.redhat.com/kalexand/050819/osdocs392/installing/installing-gather-logs.html

I added the ssh key and agent steps to all the install topics as an optional process for production clusters, eg: http://file.rdu.redhat.com/kalexand/050819/osdocs392/installing/installing_aws/installing-aws-default.html#ssh-agent-using-installing-aws-default

@jstuever, will you PTAL?

PR #14811, opened 2019-05-10
Adding must-gather CLI command

bergerhoffer:
This PR replaces #14436.

### DIFF
--- a/modules/cli-administrator-cluster-management.adoc
+++ b/modules/cli-administrator-cluster-management.adoc
@@ -5,6 +5,15 @@
 [id="cli-cluster-management-commands-{context}"]
 = Cluster management CLI commands
 
+== must-gather
+
+Bulk collect data about the current state of your cluster to debug issues.
+
+.Example: Gather debugging information
+----
+$ oc adm must-gather
+----
+
 == top
 
 Show usage statistics of resources on the server.


bergerhoffer:
@xingxingxia Could you please review this addition?

modules/cli-administrator-cluster-management.adoc
```diff
@@ -5,6 +5,15 @@
 [id="cli-cluster-management-commands-{context}"]
 = Cluster management CLI commands
 
+== must-gather
+
+Gather debugging information.
```
kalexand-rh: I just put in a [PR](https://github.com/openshift/openshift-docs/pull/14766) about the log gathering command to use during a failed installation. Since there are different log gathering commands for different stages of cluster existence, do you think it's appropriate to point out the difference?

kalexand-rh (commented):
I have a question, but it LGTM

modules/cli-administrator-cluster-management.adoc
sferich888: Yes

bergerhoffer: @kalexand-rh Good question. Agree that this could definitely use a better description, I just pulled in what was submitted in the other PR for now. Ideally we wanted to add a section on using must-gather, but we aren't going to get to it by GA since I wasn't able to get the SME input that I needed in time.

@sferich888 Would this be an appropriate summary to use here for must-gather, or do you have a better suggestion?

```
Bulk collect data about the current state of your cluster to debug issues.
```

sferich888: Yes

xingxingxia:
LGTM.

PR #14820, opened 2019-05-12
BZ-1708731: Removed unconventional markup

bergerhoffer:
@adellape Would you mind peer reviewing this? It's a request to remove the [.lead] markup, which we don't really seem to use like this elsewhere in the docs.


adellape (commented):
:+1:

PR #14822, opened 2019-05-13
Prevent -- substitution in command

jboxman:
- https://bugzilla.redhat.com/show_bug.cgi?id=1707535


vikram-redhat:
@jboxman - use \-- to escape.

jboxman:
@vikram-redhat, it was originally

```
[command]`oc exec -it <pod_name> -- ip a`
```

Do you mean using `----` or something else?

jboxman:
Verified in the bug. Merging.

PR #14293, opened 2019-03-29
Update feature gates for 4.0 new design

mburke5678:
https://gist.github.com/rphillips/4545b9ad23c1ffcaef6b63f282177e18
https://github.com/deads2k/api/blob/044292b2cd938dfaf777bcab375d661fd798c373/config/v1/types_feature.go

PR #14853, opened 2019-05-14
bug 1709729 enabling feature gates for machinehealthchecks

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1709729

@mburke5678, I broke one of your assemblies in half. If you're not ok with the changes I'm proposing, I can revert them. 

@sunzhaohua2, will you PTAL? 

preview build: http://file.rdu.redhat.com/kalexand/051419/BZ1709729/machine_management/deploying-machine-health-checks.html

### DIFF
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -572,7 +572,10 @@ Topics:
   - Name: Configuring your cluster to place pods on overcommited nodes
     File: nodes-cluster-overcommit
     Distros: openshift-enterprise,openshift-origin
-  - Name: Disabling features using feature gates
+  - Name: Enabling features using FeatureGates
+    File: nodes-cluster-enabling-features
+    Distros: openshift-enterprise,openshift-origin
+  - Name: Disabling features using FeatureGates
     File: nodes-cluster-disabling-features
     Distros: openshift-enterprise,openshift-origin
 ---
--- a/machine_management/deploying-machine-health-checks.adoc
+++ b/machine_management/deploying-machine-health-checks.adoc
@@ -12,12 +12,12 @@ include::modules/technology-preview.adoc[leveloffset=+1]
 
 .Prerequistes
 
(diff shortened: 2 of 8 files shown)


mburke5678:
@kalexand-rh Why would Disabling Feature Gates not follow Enabling in the same topic?

kalexand-rh:
@mburke5678, they're different user goals, so they should be different assemblies. This would be like putting installing and uninstalling in the same assembly - you do them at different times for different reasons.

sunzhaohua2:
@kalexand-rh I think different features have different FeatureGate name, for machinehealthcheck it is machine-api. We should point this out to know how to enable machinehealthcheck.
```
kind: FeatureGate
metadata:
  name: machine-api
spec:
  featureSet: "TechPreviewNoUpgrade"
```
```
$ oc get featuregate
NAME          AGE
cluster       45h
machine-api   18h
```

kalexand-rh:
https://github.com/openshift/openshift-docs/pull/14293#issuecomment-487084397 makes it sound like all of the tech preview features are supposed to be controlled by the same gate.

@enxebre, @ingvagabund, @rphillips, will you confirm that the machinehealthcheck should be controlled by a separate feature gate than the rest of the TP features?

ingvagabund:
@kalexand-rh to my knowledge, there is only a single feature gate for all the features. So either all the extra features are enabled or all are disabled. Including `machinehealthcheck`.

ingvagabund:
@kalexand-rh at the time of feature gating machinehealthcheck I was not aware all the extra features needs to gated by a specific `cluster` feature gate. Though, based on https://github.com/openshift/openshift-docs/pull/14853#issuecomment-492486276 it looks like there is one.

ingvagabund:
Addressing our case in https://github.com/openshift/machine-api-operator/pull/318 (hoping it will get merged).

kalexand-rh:
Jan's PR got merged, and Zhaohua Sun approved the change on the bug. Merging.

PR #14859, opened 2019-05-15
OSDOCS-429: Adding disaster recovery docs

bergerhoffer:
Preview: http://file.rdu.redhat.com/~ahoffer/2019/disaster-recovery/disaster_recovery/scenario-1-infra-recovery.html


bergerhoffer:
@tnozicka Can you please review: http://file.rdu.redhat.com/~ahoffer/2019/disaster-recovery/disaster_recovery/scenario-3-expired-certs.html

@hexfusion Can you please review: http://file.rdu.redhat.com/~ahoffer/2019/disaster-recovery/disaster_recovery/scenario-1-infra-recovery.html

bergerhoffer:
@tnozicka @hexfusion @patrickdillon Docs have been updated with the latest changes for all 3 scenarios, if you can please review. Thanks!

vikram-redhat:
@bergerhoffer - @geliu2016 pointed me to this bug that affects these docs: https://bugzilla.redhat.com/show_bug.cgi?id=1711879

modules/dr-recover-lost-control-plane-hosts.adoc
```diff
@@ -0,0 +1,278 @@
...
+... Verify that the new master host has been added to the etcd member list.
+
+.... Access the original master and connect to the running etcd container.
++
+----
+[core@ip-10-0-143-125 ~] id=$(sudo crictl ps --name etcd-member | awk 'FNR==2{ print $1}') && sudo crictl exec -it $id /bin/sh
+----
+
+.... Export variables needed for connecting to etcd.
++
```
hexfusion: I know the prompt changes but should we remind the user that this is run inside the container?

modules/dr-recover-lost-control-plane-hosts.adoc
```diff
@@ -0,0 +1,278 @@
...
+----
+sh-4.2# export ETCDCTL_API=3 ETCDCTL_CACERT=/etc/ssl/etcd/ca.crt ETCDCTL_CERT=$(find /etc/ssl/ -name *peer*crt) ETCDCTL_KEY=$(find /etc/ssl/ -name *peer*key)
+----
++
+.... Execute `etcdctl member list` and verify that the new member is listed.
++
```
hexfusion: As pointed out by QE `etcdctl member list -w table` is more correct given the output

hexfusion: Also, we should note that it could take a few minutes for the member to start in which case you will see output saying the member has not yet started. This is expected but should take no longer than 10 mins. If problem persists review container logs to understand why it is not starting.

`|  cbe982c74cbb42f | not started |  |  https://10.0.156.255:2379 |`

hexfusion (commented):
looks good added a few notes, will keep digging

modules/dr-recover-lost-control-plane-hosts.adoc
```diff
@@ -0,0 +1,278 @@
...
+..starting kube-apiserver-pod.yaml
+..starting etcd-member.yaml
+Starting kubelet..
+----
++
+Once the `etcd-snapshot-restore.sh` script completes, your cluster should now have a single member
+etcd cluster, and API services will begin restarting. This might take up to 15 minutes.
++
+In a terminal that has access to the cluster, run the following command to verify that it is ready:
++
```
hexfusion: We might want to add here that unless the other etcd members are shutdown it is possible that they will try to connect to the new cluster we just created and spam logs with errors like.

```
2019-05-20 15:33:17.648445 E | rafthttp: request cluster ID mismatch (got 9f5f9f05e4d43b7f want 807ae3bffc8d69ca)
```

This is expected and comes from members from the old cluster 9f5f9f05e4d43b7f trying to connect to 807ae3bffc8d69ca. To resolve make sure the old cluster members are shutdown.

hexfusion: Folks see that and they think it is a problem with the new cluster when in fact new cluster is fine but old members still have this peer listed.

modules/dr-restoring-cluster-state.adoc
```diff
@@ -0,0 +1,77 @@
...
+
+. Verify that the Machine Configs have been applied using the following command.
++
+----
+$ oc get machineconfigpool
+NAME     CONFIG                                             UPDATED   UPDATING
+master   rendered-master-50e7e00374e80b767fcc922bdfbc522b   True      False
+----
+
+. Repeat these steps on your other master hosts.
```
patrickdillon: `Repeat these steps...` should be moved earlier, to before `Wait for the API to come back up`.

hexfusion:
> @bergerhoffer - @geliu2016 pointed me to this bug that affects these docs: https://bugzilla.redhat.com/show_bug.cgi?id=1711879

@vikram-redhat we have a pending update that will resolve this.

modules/dr-recover-lost-control-plane-hosts.adoc
bergerhoffer: @hexfusion When you say "unless the other etcd members are shutdown", do you mean the old failed members that we're replacing?

hexfusion: Correct, for whatever reason, if one or more of these etcd instances is still running or even crashlooping, users will see logs like above. In which case they should explicitly stop etcd-member static pod on that node to resolve..

bergerhoffer:
@hexfusion @patrickdillon Thanks for reviewing, doc is updated w/ the feedback (and some from google docs) if you can take another look.

modules/dr-recover-lost-control-plane-hosts.adoc
```diff
@@ -0,0 +1,301 @@
...
+----
+[core@ip-10-0-143-125 ~]$ source /run/etcd/environment; export INITIAL_CLUSTER="etcd-member-$(hostname).$(hostname -d)=https://${ETCD_DNS_NAME}:2380"
+----
+
+.. Run the `etcd-snapshot-restore.sh` script.
++
+Pass in two parameters to the `etcd-snapshot-restore.sh` script: the path to a backed up etcd snapshot file and list of members, which is defined by the `INITIAL_CLUSTER` variable.
++
+// TODO: Is this snapshot file still optional? Wouldn't the parameters be out of order then if you didn't include it? Question out to Sam.
++
```
hexfusion: Well in some cases the local db might be your only option. I agree that default should be an explicit snapshot path. If they want to use the local db they can handle that on their own.

hexfusion: So to be clear the snapshot path should NOT be optional.

modules/dr-recover-lost-control-plane-hosts.adoc
```diff
@@ -0,0 +1,301 @@
...
+Username: kubeadmin
+Password:
+Login successful.
+----
+
+... Obtain a local copy of the pull secret used to install the cluster.
++
+----
+[core@ip-10-0-143-125 ~]$ oc get secret pull-secret -n openshift-config -o jsonpath={.data.'\.dockerconfigjson'} | base64 -d > config.json
+----
```
hexfusion: As noted recently in google doc  and in other examples the pull secret is already on disk so we can skip this part and ref the file on disk.

--registry-config='/var/lib/kubelet/config.json'

modules/dr-recover-lost-control-plane-hosts.adoc
```diff
@@ -0,0 +1,301 @@
...
+Username: kubeadmin
+Password:
+Login successful.
+----
+
+... Obtain a local copy of the pull secret used to install the cluster.
++
+----
+[core@ip-10-0-156-255 ~]$ oc get secret pull-secret -n openshift-config -o jsonpath={.data.'\.dockerconfigjson'} | base64 -d > config.json
+----
```
hexfusion: same as above this can be removed.

--registry-config='/var/lib/kubelet/config.json'

modules/dr-recover-lost-control-plane-hosts.adoc
```diff
@@ -0,0 +1,301 @@
...
+
+... Export two environment variables that are required by the `etcd-member-recover.sh` script.
++
+----
+[core@ip-10-0-156-255 ~]$ export SETUP_ETCD_ENVIRONMENT=$(oc adm release info --image-for setup-etcd-environment --registry-config=./config.json)
+----
++
+----
+[core@ip-10-0-156-255 ~]$ export KUBE_CLIENT_AGENT=$(oc adm release info --image-for kube-client-agent --registry-config=./config.json)
+----
```
hexfusion: --registry-config='/var/lib/kubelet/config.json'

modules/dr-recover-lost-control-plane-hosts.adoc
```diff
@@ -0,0 +1,301 @@
...
+In a terminal that has access to the cluster, run the following command:
++
+----
+$ oc get node ip-10-0-143-125.us-east-2.compute.internal -o  jsonpath='{.status.addresses[?(@.type=="InternalIP")].address}{"\n"}'
+10.0.143.125
+----
+
+.. From the AWS console, edit the etcd-0 Route 53 record in the private DNS zone and set the value to the new private IP address. See link:https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resource-record-sets-editing.html[Editing Records] in the AWS documentation for instructions.
+
+.. Repeat these steps to update the etcd-1 and etcd-2 entries as appropriate to the current addresses.
```
bergerhoffer: @hexfusion Trying to wrap up our convo earlier on this. So if we assume in this procedure that you had just the one master that you restored and are replacing the others, is this step still appropriate here or not? Per item 2 in https://bugzilla.redhat.com/show_bug.cgi?id=1711891

If so, do we need to call out here that this step is only appropriate in that case (and not if you're just replacing a single master, but the majority of your masters were fine)?

modules/dr-recover-lost-control-plane-hosts.adoc
bergerhoffer: So if the recommended way is to use a saved snapshot, should we go ahead and list that as a prereq (like how we do for scenario 2 here [1]).

[1] http://file.rdu.redhat.com/~ahoffer/2019/disaster-recovery/disaster_recovery/scenario-2-restoring-cluster-state.html

modules/dr-recover-lost-control-plane-hosts.adoc
hexfusion: In this scenario, we are restoring from majority failure which would assume in a 3 node cluster 2 nodes are NLA. In which case we would need to update DNS to properly reflect new master nodes once they are up. I would assume because the remaining node is still alive that its DNS is OK. After the new machines are created we need to make sure to update the A records to properly reflect.

When all new masters are online and before they are added to the cluster A records should be verified/updated for each new. master node.

hexfusion: It does not hurt to verify all DNS A records.

modules/dr-recover-lost-control-plane-hosts.adoc
hexfusion: yes

bergerhoffer:
@hexfusion PR is updated with the latest, can you look at the updates?

hexfusion (approved):
Overall this looks great, thanks for the hard work and working with us through all of the iterations.

/lgtm

modules/dr-recover-expired-control-plane-certs.adoc
```diff
@@ -0,0 +1,78 @@
...
+
+.Prerequisites
+
+* SSH access to master hosts.
+
+.Procedure
+
+. Access a master host with an expired certificate as the root user.
+
+. Obtain the `cluster-kube-apiserver-operator` image information for a release.
```
tnozicka: s/information/reference

modules/dr-recover-expired-control-plane-certs.adoc
```diff
@@ -0,0 +1,78 @@
...
+. Create a recovery API server.
++
+----
+# podman run -it --network=host -v /etc/kubernetes/:/etc/kubernetes/:Z --entrypoint=/usr/bin/cluster-kube-apiserver-operator "${KAO_IMAGE}" recovery-apiserver create
+----
+
+. Run the `export KUBECONFIG` command from the output of the above command, which is needed for the `oc` commands later in this procedure.
++
+----
+# export KUBECONFIG=/<path_to_kubeconfig>/admin.kubeconfig
```
tnozicka: s/<path_to_kubeconfig>/<path_to_recovery_kubeconfig>

modules/dr-recover-expired-control-plane-certs.adoc
```diff
@@ -0,0 +1,78 @@
...
+----
+
+. Wait for the recovery API server to come up.
++
+----
+# until oc --config=/etc/kubernetes/static-pod-resources/recovery-kube-apiserver-pod/admin.kubeconfig get namespace kube-system 2>/dev/null 1>&2; do echo 'Waiting for recovery apiserver to come up.'; sleep 1; done
```
tnozicka: `--config=/etc/kubernetes/static-pod-resources/recovery-kube-apiserver-pod/admin.kubeconfig` this is not needed now that we export KUBECONFIG variable

modules/dr-recover-expired-control-plane-certs.adoc
```diff
@@ -0,0 +1,78 @@
...
+----
+
+. Run the `regenerate-certificates` command. It fixes the certificates in the API, overwrites the old certificates on the local drive, and restarts static Pods to pick them up.
++
+----
+# podman run -it --network=host -v /etc/kubernetes/:/etc/kubernetes/:Z --entrypoint=/usr/bin/cluster-kube-apiserver-operator "${KAO_IMAGE}" regenerate-certificates
+----
++
+When the control plane starts on this node, it will reinstall itself on the other nodes because the kubelet is connected to API servers using an internal load balancer.
```
tnozicka: this doesn't sound well together with the next paragraph.

modules/dr-recover-expired-control-plane-certs.adoc
```diff
@@ -0,0 +1,78 @@
...
+
+. After the certificates are fixed in the API, use the following commands to force new rollouts for the control plane.
```
tnozicka: probably move `it will reinstall itself on the other nodes because the kubelet is connected to API servers using an internal load balancer.` here

tnozicka (commented):
control plane recovery looks good over all, few nits

modules/dr-recover-expired-control-plane-certs.adoc
```diff
@@ -0,0 +1,78 @@
...
+* SSH access to master hosts.
+
+.Procedure
+
+. Access a master host with an expired certificate as the root user.
+
+. Obtain the `cluster-kube-apiserver-operator` image information for a release.
++
+----
+# KAO_IMAGE=$( oc adm release info --registry-config='/var/lib/kubelet/config.json' <release_image> --image-for=cluster-kube-apiserver-operator ) <1>
```
tnozicka: probably make RELEASE_IMAGE a variable and put the description there, it will be easier to copy paste

```
RELEASE_IMAGE=<release_image>

KAO_IMAGE=$( oc adm release info --registry-config='/var/lib/kubelet/config.json' "${RELEASE_IMAGE}" --image-for=cluster-kube-apiserver-operator )
```

modules/dr-recover-expired-control-plane-certs.adoc
```diff
@@ -0,0 +1,78 @@
...
+----
++
+<1> An example value for `<release_image>` is `quay.io/openshift-release-dev/ocp-release:4.1.0-rc.0`.
```
tnozicka: might be best to leave the rc part out of it

bergerhoffer: @tnozicka if i take out the "-rc", should I also take out the ".0" after it? So the version is just "4.1.0"?

bergerhoffer:
Thanks for the feedback @tnozicka - updates are in the PR if you can double check them.

bergerhoffer:
@openshift/team-documentation Can I get a peer review started of this please?

modules/dr-recover-expired-control-plane-certs.adoc
```diff
@@ -0,0 +1,168 @@
...
+set -eou pipefail
+
+# context
+intapi=$(oc get infrastructures.config.openshift.io cluster -o "jsonpath={.status.apiServerURL}")
+context="$(oc config current-context)"
+# cluster
+cluster="$(oc config view -o "jsonpath={.contexts[?(@.name==\"$context\")].context.cluster}")"
+server="$(oc config view -o "jsonpath={.clusters[?(@.name==\"$cluster\")].cluster.server}")"
+# token
+ca_crt_data="$(oc get secret -n openshift-machine-config-operator node-bootstrapper-token -o "jsonpath={.data.ca\.crt}" | base64 --decode)"
```
tnozicka: @rphillips is this a long lived token?

modules/dr-recover-expired-control-plane-certs.adoc
```diff
@@ -0,0 +1,168 @@
...
+
+. Wait for the recovery API server to come up.
++
+----
+# until oc get namespace kube-system 2>/dev/null 1>&2; do echo 'Waiting for recovery apiserver to come up.'; sleep 1; done
+----
+
+. Create a bootstrap kubeconfig with a valid user.
+
+.. Create a file called `restore_kubeconfig.sh` with the following contents.
```
tnozicka: I'd assume we need to recover `/etc/kubernetes/ca.crt` but for that we need to run `regenerate-certs` first.

modules/dr-recover-expired-control-plane-certs.adoc
rphillips: I'm not sure

@deads2k do you know if this token is long lived?

modules/dr-recover-expired-control-plane-certs.adoc
rphillips: True. @bergerhoffer this means that #10 and #11 would move up to step 7.

bergerhoffer: Moved steps 10 and 11 to above the current step 7.

bergerhoffer:
Hi @geliu2016, @xingxingxia - could someone QE review these disaster recovery docs? Thanks!

modules/dr-restoring-cluster-state.adoc
```diff
@@ -0,0 +1,116 @@
...
++
+----
+sh-4.2#  etcdctl member list -w table
+
++------------------+---------+------------------------------------------+------------------------------------------------------------------+---------------------------+
+|        ID        | STATUS  |                   NAME                   |                            PEER ADDRS                            |       CLIENT ADDRS        |
++------------------+---------+------------------------------------------+------------------------------------------------------------------+---------------------------+
+| 29e461db6be4eaaa | started | etcd-member-ip-10-0-164-170.ec2.internal | https://etcd-2.clustername.devcluster.openshift.com:2380 | https://10.0.164.170:2379 |
+|  cbe982c74cbb42f | started | etcd-member-ip-10-0-143-125.ec2.internal | https://etcd-0.clustername.openshift.com:2380 | https://10.0.143.125:2379 |
+| a752f80bcb0da3e8 | started |   etcd-member-ip-10-0-156-2.ec2.internal | https://etcd-1.clustername.devcluster.openshift.com:2380 |   https://10.0.156.2:2379 |
```
hexfusion: https://etcd-0.clustername.openshift.com:2380 

should be 

https://etcd-0.clustername.devcluster.openshift.com:2380

hexfusion:
Small nit above otherwise LGTM for etcd sections

modules/dr-restoring-cluster-state.adoc
bergerhoffer: Thanks, updated.

geliu2016:
@bergerhoffer , it seems that this bug have not be fix in doc: https://bugzilla.redhat.com/show_bug.cgi?id=1713219#c5, and we're verifying doc on baremetal and vsphere cluster, so we may close it after test done on all support platform.

bergerhoffer:
Great thanks @geliu2016. I've reached out to @hexfusion and @tnozicka to help me get the right steps added to the doc for approving the CSRs. I will hopefully update this PR tomorrow to fix that BZ.

Let me know if you find any other issues with doc during your testing. Thanks!

geliu2016:
This doc is work well to cover aws platform after bug fixed:  https://bugzilla.redhat.com/show_bug.cgi?id=1713219#c5, so we may close it at first, and if there is any special issue on other platform, Dev may have options to adjust the script or doc, so there is not block issue for doc processing. thx

modules/dr-recover-expired-control-plane-certs.adoc
```diff
@@ -0,0 +1,166 @@
...
+.. On a master host, stop the kubelet.
++
+----
+# systemctl stop kubelet
+----
+
+.. Delete stale kubelet data.
++
+----
+# rm -rf /var/lib/kubelet/*
```
tnozicka: We can't delete `/var/lib/kubelet/config.json`, if the recovery fails we need this to recover.

bergerhoffer: @tnozicka What do you recommend we do instead here?

bergerhoffer: @sjenning Suggested the following, so i've made the updates:

```
# rm -rf /var/lib/kubelet/pki /var/lib/kubelet/kubeconfig
```

modules/dr-recover-lost-control-plane-hosts.adoc
```diff
@@ -0,0 +1,305 @@
...
++
+This procedure assumes that you have copied a snapshot file called `snapshot.db` to the `/home/core/` directory of your master host.
+
+.. Access the remaining master host.
+
+.. Set the `INITIAL_CLUSTER` variable to the list of members in the format of `<name>=<url>`. This variable will be passed to the restore script, and in this procedure, it is assumed that there is only a single member at this time.
++
+----
+[core@ip-10-0-143-125 ~]$ source /run/etcd/environment; export INITIAL_CLUSTER="etcd-member-$(hostname).$(hostname -d)=https://${ETCD_DNS_NAME}:2380"
+----
```
hexfusion: @bergerhoffer We need to make this more generic in bare metal and vsphere we can not assume hostname will give us what is expected.

Let's change this to a more generic example

[core@ip-10-0-143-125 ~]$ export INITIAL_CLUSTER="etcd-member-ip-10-0-143-125=https://etcd-0.clustername.devcluster.openshift.com:2380"

hexfusion: We could note the general format as
export INITIAL_CLUSTER="etcd-member-<hostname>.<domain>=https://<etcd-dns-name>:2380"

bergerhoffer:
Got the go ahead to merge this. If we need to add additional steps for anything else, we can do so in a separate PR.

modules/dr-recover-expired-control-plane-certs.adoc
```diff
@@ -0,0 +1,189 @@
+// Module included in the following assemblies:
+//
+// * disaster_recovery/scenario-3-expired-certs.adoc
+
+[id="dr-scenario-3-recovering-expired-certs_{context}"]
+= Recovering from expired control plane certificates
+
+Follow this procedure to recover from a situation where your control plane certificates have expired.
```
kalexand-rh: Maybe "You can generate new control plane certificates if yours expired."

modules/dr-recover-expired-control-plane-certs.adoc
```diff
@@ -0,0 +1,189 @@
...
+.Procedure
+
+. Access a master host with an expired certificate as the root user.
+
+. Obtain the `cluster-kube-apiserver-operator` image reference for a release.
++
+----
+# RELEASE_IMAGE=<release_image> <1>
+----
+<1> An example value for `<release_image>` is `quay.io/openshift-release-dev/ocp-release:4.1.0`.
```
kalexand-rh: How do I know how to get this value?

modules/dr-recover-expired-control-plane-certs.adoc
```diff
@@ -0,0 +1,189 @@
...
++
+----
+# KAO_IMAGE=$( oc adm release info --registry-config='/var/lib/kubelet/config.json' "${RELEASE_IMAGE}" --image-for=cluster-kube-apiserver-operator )
```
kalexand-rh: Does this command belong to another step?

modules/dr-recover-expired-control-plane-certs.adoc
```diff
@@ -0,0 +1,189 @@
...
+# podman pull --authfile=/var/lib/kubelet/config.json "${KAO_IMAGE}"
+----
+
+. Create a recovery API server.
++
+----
+# podman run -it --network=host -v /etc/kubernetes/:/etc/kubernetes/:Z --entrypoint=/usr/bin/cluster-kube-apiserver-operator "${KAO_IMAGE}" recovery-apiserver create
+----
+
+. Run the `export KUBECONFIG` command from the output of the above command, which is needed for the `oc` commands later in this procedure.
```
kalexand-rh: I'm not sure what part of the previous command I need to use in this one.
Not "above"
> ! above adj
Do not use to indicate a relative location in a document, as in “the above restrictions.” Use “previous” or
“preceding.”

I'd remove the second clause and add something like "You must export the `KUBECONFIG` to run `oc` commands in the following steps."

modules/dr-recover-expired-control-plane-certs.adoc
```diff
@@ -0,0 +1,189 @@
...
++
+----
+# export KUBECONFIG=/<path_to_recovery_kubeconfig>/admin.kubeconfig
+----
+
+. Wait for the recovery API server to come up.
```
kalexand-rh: I'd s/come up/become available

modules/dr-recover-expired-control-plane-certs.adoc
```diff
@@ -0,0 +1,189 @@
...
++
+----
+# until oc get namespace kube-system 2>/dev/null 1>&2; do echo 'Waiting for recovery apiserver to come up.'; sleep 1; done
+----
+
+. Run the `regenerate-certificates` command. It fixes the certificates in the API, overwrites the old certificates on the local drive, and restarts static Pods to pick them up.
```
kalexand-rh: I'd move the second sentence after the command.

modules/dr-recover-expired-control-plane-certs.adoc
```diff
@@ -0,0 +1,189 @@
...
++
+----
+# podman run -it --network=host -v /etc/kubernetes/:/etc/kubernetes/:Z --entrypoint=/usr/bin/cluster-kube-apiserver-operator "${KAO_IMAGE}" regenerate-certificates
+----
+
+. After the certificates are fixed in the API, use the following commands to force new rollouts for the control plane. It will reinstall itself on the other nodes because the kubelet is connected to API servers using an internal load balancer.
```
kalexand-rh: I'd move the second sentence after the command.

modules/dr-recover-expired-control-plane-certs.adoc
```diff
@@ -0,0 +1,189 @@
...
++
+----
+# oc patch kubecontrollermanager cluster -p='{"spec": {"forceRedeploymentReason": "recovery-'"$( date --rfc-3339=ns )"'"}}' --type=merge
+----
++
+----
+# oc patch kubescheduler cluster -p='{"spec": {"forceRedeploymentReason": "recovery-'"$( date --rfc-3339=ns )"'"}}' --type=merge
+----
+
+. Create a bootstrap kubeconfig with a valid user.
```
kalexand-rh: How do I know if I have a valid user?

modules/dr-recover-expired-control-plane-certs.adoc
```diff
@@ -0,0 +1,189 @@
...
+cat "$KUBECONFIG"
+----
+
+.. Make the script executable.
++
+----
+# chmod +x restore_kubeconfig.sh
+----
+
+.. Execute the script and save the output to a file called `kubeconfig`.
```
kalexand-rh: Not "execute"

modules/dr-recover-expired-control-plane-certs.adoc
```diff
@@ -0,0 +1,189 @@
...
++
+----
+# ./restore_kubeconfig.sh > kubeconfig
+----
+
+.. Copy the `kubeconfig` file to all master hosts and move it to `/etc/kubernetes/kubeconfig`.
```
kalexand-rh: s/move it to `/etc/kubernetes/kubeconfig`./move it to the `/etc/kubernetes/kubeconfig` directory.

modules/dr-recover-expired-control-plane-certs.adoc
```diff
@@ -0,0 +1,189 @@
...
+----
+# systemctl start kubelet
+----
+
+. Approve the pending `node-bootstrapper` certificates signing requests (CSRs).
+
+.. Get the list of current CSRs.
++
+----
+# oc get csr
```
kalexand-rh: Do you need to run these commands as root? If not, s/#/$

modules/dr-recover-expired-control-plane-certs.adoc
```diff
@@ -0,0 +1,189 @@
...
+----
+
+.. Review the details of a CSR to verify it is valid.
```
kalexand-rh: s/verify it/verify that it

modules/dr-recover-expired-control-plane-certs.adoc
```diff
@@ -0,0 +1,189 @@
...
+
+.. Approve each valid CSR.
++
+----
+# oc adm certificate approve <csr_name>
+----
++
+Be sure to approve all pending `node-bootstrapper` CSRs.
+
+. Destroy the recovery API server because it is no longer needed.
```
kalexand-rh: ~because it is no longer needed~

modules/dr-recover-expired-control-plane-certs.adoc
```diff
@@ -0,0 +1,189 @@
...
++
+----
+# podman run -it --network=host -v /etc/kubernetes/:/etc/kubernetes/:Z --entrypoint=/usr/bin/cluster-kube-apiserver-operator "${KAO_IMAGE}" recovery-apiserver destroy
+----
++
+Wait for the control plane to restart and pick up the new certificates. This might take up to 10 minutes.
```
kalexand-rh: How do you know if it's restarted? This might need to be a separate step.

s/This/This process

kalexand-rh (commented):
@bergerhoffer, sorry the review's so late, but here are some suggestions.

PR #14867, opened 2019-05-15
BZ-1707397: Removed deny OAuth grant option.

huffmanca:
Removed the `deny` OAuth grant option.

This is for OS 4.x.


stuartchuan:
The changes LGTM.

Issue #14753, opened 2019-05-08
[enterprise-4.1] Edit suggested in file authentication/identity_providers/configuring-github-identity-provider.adoc


cblecker:
### Which section(s) is the issue in?

Creating the Secret / Creating a ConfigMap

### What needs fixing?

The command calls it out, but the docs should be clearer that both these items should go in the `openshift-config` namespace. It could also be clearer that a configmap is only needed for the certificate bundle, and only if you're using a GitHub Enterprise install with a non-publicly trusted root.

vikram-redhat:
Tracking: https://bugzilla.redhat.com/show_bug.cgi?id=1707637

PR #14873, opened 2019-05-15
BZ-1707637: Incorporated minor fixes.

huffmanca:
Included fixes mentioned in https://github.com/openshift/openshift-docs/issues/14753 and https://github.com/openshift/openshift-docs/issues/14753 .

@cblecker - would you be able to review these changes? I noticed that the GitHub callback URL no longer contains the port, should this be included?

This is for OS 4.x.

### DIFF
--- a/modules/identity-provider-registering-github.adoc
+++ b/modules/identity-provider-registering-github.adoc
@@ -17,19 +17,20 @@ https://github.com/settings/applications/new[Register a new OAuth application].
 ** For GitHub Enterprise, go to your GitHub Enterprise home page and then click
 *Settings -> Developer settings -> Register a new application*.
 . Enter an application name, for example `My OpenShift Install`.
-. Enter a homepage URL, such as `https://myapiserver.com:8443`.
+. Enter a homepage URL, such as 
+`https://console-openshift-console.apps.example-openshift-cluster.com/`.
 . Optionally, enter an application description.
 . Enter the authorization callback URL, where the end of the URL contains the
 identity provider `name`:
 +
 ----
-<apiserver>/oauth2callback/<identityProviderName>
+https://oauth-openshift.apps.<cluster-name>.<cluster-domain>/oauth2callback/<idp-provider-name>
 ----
 +
 For example:
 +
 ----
-https://myapiserver.com:8443/oauth2callback/github/
+https://oauth-openshift.apps.example-openshift-cluster.com/oauth2callback/github/
 ----
 . Click *Register application*. GitHub provides a Client ID and a Client Secret.
 You need these values to complete the identity provider configuration.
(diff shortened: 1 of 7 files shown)


cblecker:
This looks great!

/lgtm

cblecker:
Oh, and specifically on the port, it's no longer needed, as it's on the https default 443 port. We're hitting an apps URL via the router, rather than the apiserver directly.

modules/identity-provider-registering-github.adoc
```diff
@@ -17,19 +17,20 @@ https://github.com/settings/applications/new[Register a new OAuth application].
...
-. Enter a homepage URL, such as `https://myapiserver.com:8443`.
+. Enter a homepage URL, such as 
+`https://console-openshift-console.apps.example-openshift-cluster.com/`.
 . Optionally, enter an application description.
 . Enter the authorization callback URL, where the end of the URL contains the
 identity provider `name`:
 +
 ----
-<apiserver>/oauth2callback/<identityProviderName>
+https://openshift-authentication-openshift-authentication.apps.<clusterDomain>/oauth2callback/<identityProviderName>/
```
stuartchuan: Should be https://oauth-openshift.apps.<cluster-name>.<cluster-domain>/oauth2callback/<idp-provider-name>

modules/identity-provider-registering-github.adoc
```diff
@@ -17,19 +17,20 @@ https://github.com/settings/applications/new[Register a new OAuth application].
...
 ----
 +
 For example:
 +
 ----
-https://myapiserver.com:8443/oauth2callback/github/
+https://openshift-authentication-openshift-authentication.apps.example-openshift-cluster.com/oauth2callback/github/
```
stuartchuan: Then also need update as above comment.

stuartchuan:
https://github.com/openshift/openshift-docs/pull/14873/files#diff-5940016fd1c3c83f1e5af4975883eb49L28 the secret still need to be created in openshift-config namespace.

stuartchuan:
Also the redirect URI  for google, gitlab  and OpenID configuration all need update to `https://oauth-openshift.apps.<cluster-name>.<cluster-domain>/oauth2callback/<idp-provider-name>`

modules/identity-provider-registering-github.adoc
huffmanca: Corrected to use the given format.

modules/identity-provider-registering-github.adoc
huffmanca: I've used the above format, but kept it as an explicit URL instead of using replaceables.

cblecker:
Thanks for the catch @stuartchuan. I was able to confirm on my test cluster that this is correct.

huffmanca:
The `clientID` description for these CRs has been updated to use the new callback URL. This should be ready for another review.

stuartchuan:
Thanks for the update, it's lgtm now.

PR #14795, opened 2019-05-09
Specifying more than 3 nodes on Cluster Logging CR creates invalid Elasticsearch CR

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1706134

PR #14881, opened 2019-05-16
How to scaling your Elasticsearch cluster

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1700247

### DIFF
--- a/modules/efk-logging-deploy-subscription.adoc
+++ b/modules/efk-logging-deploy-subscription.adoc
@@ -315,7 +315,7 @@ However, an unmanaged deployment does not receive updates until the cluster logg
 [NOTE]
 +
 ====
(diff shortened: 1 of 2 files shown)


mburke5678:
@anpingli 
1. I added the note from https://github.com/openshift/openshift-docs/pull/14795 about scaling ES nodes using the CR instance, which you recently reviewed. The limitation is the nodeCount must be 1-3 has been removed. The maximum ES master nodes is 3; if you specify more than 3, the additional nodes are data only. Also updated the example code and specifying 5 nodes.

2. We are not documenting that users can modify the CRDs. 

Let me know if the addresses your concerns.

Thank you!!

Michael

anpingli (commented):
LGTM

PR #14891, opened 2019-05-16
bug 1710427 specify number of CRS

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1710427

### DIFF
--- a/modules/installation-approve-csrs.adoc
+++ b/modules/installation-approve-csrs.adoc
@@ -9,8 +9,8 @@
 [id="installation-approve-csrs_{context}"]
 = Approving the CSRs for your machines
 
-When you add machines to a cluster, you must approve the pending certificate
-signing requests (CSRs) for the machines that you added.
+When you add machines to a cluster, you must approve two pending certificates
+signing request (CSRs) for each machine that you added.
 
 .Prerequisites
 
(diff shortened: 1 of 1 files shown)


modules/installation-approve-csrs.adoc
```diff
@@ -19,20 +19,34 @@ signing requests (CSRs) for the machines that you added.
...
+<2> A server request CSR.
++
+In this example, two machines are joining the cluster.
+
+. After all of the pending CSRs for the machines you added are in `Pending`
+status, approve the CSRs for your cluster machines:
++
+[NOTE]
+====
+Because the CSRs rotate automatically, approve your CSRs within half an hour
```
abhinavdahiya: cc @sjenning @rphillips

abhinavdahiya: When trying to get a certificate (client/server), kubelet (node) creates a CSR and waits for it to be Approved and Issued. If the CSR is not approved for _some time_ (Pod team can provide more guidance on what this period is), kubelet (node) creates a new request for certificate with new CSR.

So there's a chance that there might be more CSR objects in Pending for the same purpose.

I'm hoping Pod Team (cc @sjenning ) can provide information on the exact behavior.

rphillips: the _some time_ the kubelet will wait for a certificate to be issues is 1 hr.

kalexand-rh: Thanks! I've updated the note to say that it takes an hour for the certs to rotate and that if you wait longer than that, you might have to approve more certs.

kalexand-rh:

PR #14453, opened 2019-04-15
Garbage Collection fka Nodes docs feedback 4

mburke5678:
Comments per @sunilcio 
https://bugzilla.redhat.com/show_bug.cgi?id=1688637#c11

PR #14893, opened 2019-05-16
Garbage collection configure container and image GC one file

mburke5678:
"one more thing we need to also include in doc is that if we want to configure both image garbage collection and container garbage collection, then we need to combine parameters for both in single kubeletConfig"

https://github.com/openshift/openshift-docs/pull/14453#issuecomment-493147977

### DIFF
--- a/modules/nodes-nodes-garbage-collection-configuring.adoc
+++ b/modules/nodes-nodes-garbage-collection-configuring.adoc
@@ -6,7 +6,20 @@
 [id="nodes-nodes-garbage-collection-configuring_{context}"]
 = Configuring garbage collection for containers and images
 
-As an administrator, you can configure how {product-title} performs garbage collection.
+As an administrator, you can configure how {product-title} performs garbage collection by creating a `kubeletConfig` object for each Machine Config Pool.
+
+[NOTE]
+====
+{product-title} supports only one `kubeletConfig` object for each Machine Config Pool.
+====
+
+You can configure any combination of the following:
+
+* soft eviction for containers
+* hard eviction for containers
+* eviction for images 
+
+For soft container eviction you can also configure a grace period before eviction.
 
 .Prerequisites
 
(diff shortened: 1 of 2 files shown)


mburke5678:
@sunilcio @rphillips PTAL. I combined the separate steps to configure container and image GC into one step and included all GC parameters, as in [Sunil's example in the other PR](https://github.com/openshift/openshift-docs/pull/14453#issuecomment-493147977).

modules/nodes-nodes-garbage-collection-configuring.adoc
```diff
@@ -43,42 +56,54 @@ $ oc label machineconfigpool worker custom-kubelet=small-pods
...
-    - nodefs.available=1m30s
-    - nodefs.inodesFree=1m30s
-    - imagefs.available=1m30s
-    - imagefs.inodesFree=1m30s
+    evictionSoft: <3>
+    - memory.available: "<100Mi" <4>
+    - nodefs.available: "<10%"
+    - nodefs.inodesFree: "<5%"
+    - imagefs.available: "<15%"
+    - imagefs.inodesFree: "<10%"
```
sunilcio: @mburke5678 we need to remove all `<` (less than) operator for evictionSoft list options as it is creates syntax error and kubelet fails to start. Please refer my example kubeletConfig yaml https://github.com/openshift/openshift-docs/pull/14453#issuecomment-493147977. Also we need to remove `-` (hyphen) from the list as this also does not work.

mburke5678: @sunilcio Sorry. Should have just copied your example instead of trying to fix mine. Wanted to show the less than option, not knowing it doesn't work.

modules/nodes-nodes-resources-configuring-about.adoc
```diff
@@ -100,7 +100,7 @@ The following is an example to illustrate the impact of node allocatable for mem
 * Node capacity is `32Gi`
 * --kube-reserved is `2Gi`
 * --system-reserved is `1Gi`
-* --eviction-hard is set to `<100Mi`.
+* --eviction-hard is set to `100Mi`.
```
mburke5678: @sunilcio I changed this instance also.

mburke5678:
@sunilcio PTAL. I am asked to get this closed by the end of the week.

sunilcio:
@mburke5678 LGTM. Just re-add the label <4> & <5> to new options as I see the old ones are being removed.

PR #14918, opened 2019-05-17
bug 1711207 clarifying compute/control plane types

kalexand-rh:
Draft of https://bugzilla.redhat.com/show_bug.cgi?id=1711207

@morenod, @wking, will you PTAL?

### DIFF
--- a/modules/installation-aws-config-yaml.adoc
+++ b/modules/installation-aws-config-yaml.adoc
@@ -24,10 +24,10 @@ This sample YAML file is provided for reference only. You must obtain your
 
 [source,yaml]
 ----
-apiVersion: v1beta3
+apiVersion: v1
 baseDomain: example.com <1>
 controlPlane: <2>
-- name: master
+  name: master <3>
   platform:
     aws:
       zones:
@@ -40,7 +40,7 @@ controlPlane: <2>
       type: m5.xlarge
   replicas: 3
 compute: <2>
-- name: worker
+- name: worker <3>
   platform:
     aws:
       rootVolume:
@@ -73,12 +73,19 @@ platform:
       adminContact: jdoe
       costCenter: 7536
 pullSecret: '{"auths": ...}' <1>
-sshKey: ssh-ed25519 AAAA... <3>
+sshKey: ssh-ed25519 AAAA... <5>
 ----
 <1> Required. The installation program prompts you for this value.
 <2> If you do not provide these parameters and values, the installation program
 provides the default value.
-<3> You can optionally provide the `sshKey` value that you use to access the
-machines in your cluster.
+<3> The `controlPlane` section is a singlet, and the `compute` section is an
+array. Because the sections have different definitions, the first line of the
+`compute` section must begin with a hyphen, `-`, and the first line of the
+`controlPlane` section must not. Although both sections currently define a
+single machine pool, it is possible that future versions of {product-title}
+will support defining multiple compute pools during installation. Only one
+control plane pool is used.
 <4> To configure faster storage for etcd, especially for larger clusters, set the
 storage type as `io1` and set `iops` to `2000`.
+<5> You can optionally provide the `sshKey` value that you use to access the
+machines in your cluster.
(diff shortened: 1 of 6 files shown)


modules/installation-aws-config-yaml.adoc
```diff
@@ -73,12 +73,19 @@ platform:
...
 pullSecret: '{"auths": ...}' <1>
-sshKey: ssh-ed25519 AAAA... <3>
+sshKey: ssh-ed25519 AAAA... <5>
 ----
 <1> Required. The installation program prompts you for this value.
 <2> If you do not provide these parameters and values, the installation program
 provides the default value.
-<3> You can optionally provide the `sshKey` value that you use to access the
-machines in your cluster.
+<3> The `controlplane` section is a singlet, and the `compute` section is an
```
wking: `controlplane` -> `controlPlane` throughout

modules/installation-aws-config-yaml.adoc
```diff
@@ -73,12 +73,19 @@ platform:
...
+array. Because the sections have different definitions, the first line of the
+`compute` section must begin with a hyphen, `-`, and the first line of the
+`controlplane` section must not. Although both sections currently define a
+single machine pool, it is possible that future versions of {product-title}
+will support defining multiple compute pools during installation. Only one
+control plane pool is used in the current system design.
```
wking: I dunno if we need the "current design" wiggle, but ask the master team to confirm.

modules/installation-aws-config-yaml.adoc
```diff
@@ -73,12 +73,19 @@ platform:
...
 <4> To configure faster storage for etcd, especially for larger clusters, set the
 storage type as `io1` and set `iops` to `2000`.
+<5> You can optionally provide the `sshKey` value that you use to access the
```
wking: Strengthen to "you need to set this if you want to debug several classes of install failures" or some such.

modules/installation-aws-config-yaml.adoc
kalexand-rh: I'll just take out the wiggle.

modules/installation-aws-config-yaml.adoc
kalexand-rh: I'll add it to https://github.com/openshift/openshift-docs/pull/14766/.

modules/installation-bare-metal-config-yaml.adoc
```diff
@@ -14,39 +14,46 @@ parameters.
...
+<5> The cluster name.
+<6> A block of IP addresses from which pod IPs are allocated.
+<7> The subnet prefix length to assign to each individual node. For example, if
 `hostPrefix` is set to `23`, then each node is assigned a `/23` subnet out of
 the given `cidr`, which allows for 510 (2^(32 - 23) - 2) pod IPs addresses.
-<7> The IP address pool to use for service IP addresses. You can enter only
+<8> The IP address pool to use for service IP addresses. You can enter only
 one IP address pool.
-<8> The pull secret that provides components in the cluster access to images for
+<9> The pull secret that provides components in the cluster access to images for
```
DanyC97: @kalexand-rh  @wking my 0.02 $: for folks who are new on v4 or those coming from v3, i think will be useful to reword it to explicitly say what this pullSecret is, something like 

```
The pull secret that provides components in the cluster access to container images for {product-title} components
```
i think will do.

I'd even go and maybe extend the pullSecret glossary to say that is nothing but the login credentials to a docker private registry which is holding the container images

wking: "images" -> "container images" is fine with me.  And that's the main pull-secret goal, but it also authorizes you to telemetry, and maybe other things I'm not aware of.  So in anything as global as a glossary, I'd probably say that it's about authenticating your cluster as belonging to you for services provided by the included authorities, including quay.io for serving container images.

kalexand-rh: @wking, I updated the pull secret description. What do you think?

wking: Looks great

DanyC97: looks great, thanks @kalexand-rh @wking !

kalexand-rh:

PR #14926, opened 2019-05-17
osdocs326 install concept updates

kalexand-rh:
https://jira.coreos.com/browse/OSDOCS-326

@abhinavdahiya, @wking, Alex, Scott, and Derek had some articles about UPI and IPI installations. I've changed a bit of the conceptual information about installations to match.

Will you PTAL and let me know what you think? I'm curious if there's more areas you think that users need info on before they install.


modules/installation-overview.adoc
```diff
@@ -26,18 +26,40 @@ These two basic types of {product-title} clusters are frequently called
 installer-provisioned infrastructure clusters and user-provisioned
 infrastructure clusters.
 
+Both types of clusters have the following characteristics:
+
+* Highly available infrastructure with no single points of failure
```
wking: Well, you could ask for a single control-plane node, or some other non-HA thing, with either approach.  Maybe append "possible" or some such?

modules/installation-overview.adoc
```diff
@@ -145,12 +164,16 @@ installation program to generate the assets that you require to provision the
...
+If you do not use infrastructure that the installation program provisioned, you must manage
 and maintain the cluster resources yourself, including:
 
 * The control plane and compute machines that make up the cluster
 * Load balancers
 * Cluster networking, including the DNS records and required subnets
+* Storage for the cluster infrastructure and applications
+
+If your cluster uses user-provisioned infrastructure, you have the option of
+adding RHEL worker nodes to your cluster.
```
wking: Can't you do this with an installer-provisioned cluster too?

kalexand-rh: @mtnbikenc, in the [BYO RHEL compute](https://docs.openshift.com/container-platform/4.1/machine_management/adding-rhel-compute.html) topic, we say: 
> In OpenShift Container Platform, you can add Red Hat Enterprise Linux (RHEL) compute, or worker, machines to a user-provisioned infrastructure cluster. You can use RHEL as the operating system on only compute machines.

Is that right?

modules/installation-overview.adoc
kalexand-rh: The documented minimum number of control plane nodes is 3. Please file a bug and get PM to agree if you want me to admit otherwise. :smile_cat:

wking: Uh, maybe you decide to run three control-plane VMs on a single physical host?  Stretching, but I'd rather emphasize HA-by-default, and avoid claiming unqualified HA regardless of anything admins do to shoot themselves in the foot ;)

kalexand-rh: Lol, I'll add a note to the UPI topic to say that you should use different physical hosts for the machines, too.

kalexand-rh:
Jianlin approved this PR in the JIRA. @wking, do you see any other issues before I merge?

modules/installation-overview.adoc
wking: > Lol, I'll add a note to the UPI topic to say that you should use different physical hosts for the machines, too.

I'm less concerned about specific holes than I am about the fact that, no matter how hard we try, there will probably still be holes ;).  So I'd rather have some wiggle in blanket HA claims.

kalexand-rh: Is this enough wiggle room?
> Highly available infrastructure with no single points of failure is available
by default

wking: yup, works for me :)

PR #14932, opened 2019-05-19
BZ-1711689: Updated PV access modes.

huffmanca:
Updated the access mode table to comment out any storage options that aren't included as of 4.1.

This is for OS 4.x.

### DIFF
--- a/modules/storage-persistent-storage-pv.adoc
+++ b/modules/storage-persistent-storage-pv.adoc
@@ -126,18 +126,19 @@ the Pods that use these volumes are deleted.
 |===
 |Volume Plug-in  |ReadWriteOnce  |ReadOnlyMany  |ReadWriteMany
 |AWS EBS  | ✅ | - |  -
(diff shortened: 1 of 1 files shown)


modules/storage-persistent-storage-pv.adoc
```diff
@@ -128,16 +128,17 @@ the Pods that use these volumes are deleted.
 |AWS EBS  | ✅ | - |  -
 |Azure File | ✅ | ✅ | ✅
 |Azure Disk | ✅ | - | -
```
liangxia: Azure (both file and disk) are not supported in OCP 4.1

huffmanca: Both of these lines have been commented out.

huffmanca:
@liangxia ,

The Azure lines have been commented out; this should be ready for rereview.

liangxia:
LGTM ( from QE )

PR #14934, opened 2019-05-19
BZ-1693420: Removed unnecessary secret references.

huffmanca:
Examining the current identity providers there are some unnecessary references to secrets. For instance, the RequestHeader IDP cannot configure a secret, but the instructions are in the assembly. This has been removed.

This is for OS 4.x.


stuartchuan:
Yes, the changes lgtm.

PR #14939, opened 2019-05-20
RHCOS URLS for beta6

kalexand-rh:
@jianlinliu, since beta 6 was released, I updated the image URLs to match https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.1/4.1.0-rc.4/.

Will you PTAL?

### DIFF
--- a/modules/installation-aws-user-infra-rhcos-ami.adoc
+++ b/modules/installation-aws-user-infra-rhcos-ami.adoc
@@ -17,49 +17,49 @@ You must use a valid {op-system-first} AMI for your Amazon Web Services
 |AWS AMI
 
 |`ap-northeast-1`
-|`ami-06153e8d5cf577a81`
+|`ami-0725cf3917ea24b16`
 
 |`ap-northeast-2`
-|`ami-0b75c8dbcc07363b9`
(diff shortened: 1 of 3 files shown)


jianlinliu:
I think final beta6 build should be rc.5, some fix is not merged into rc.4, so rc.5 came out.

kalexand-rh:
Figures. I don't see rc5 posted. I'll check again my afternoon and update the links.

kalexand-rh:
I still don't see new bits.

jianlinliu:
Ah, rc.5 is for beta7, pls check "Announcing OpenShift 4.1 - Beta 7 [4.1.0-rc.5]" email in aos-devel, seem like this PR need some polish to keep up-to-date.

kalexand-rh:
From that thread:
> There are no changes for the RHCOS assets as the latest there is still 4.1.0-rc.4?

> right, there are no changes to rhcos assets.

We don't explicitly mention the beta version in this PR, so I think it's right.
@jianlinliu, I'm going to merge this PR, but if there are problems with it, I am happy to submit another one.

PR #13726, opened 2019-02-21
Added 4.1 release notes

ahardin-rh:
Preview Build: http://file.rdu.redhat.com/~ahardin/05072019/4-0-release-notes/release_notes/ocp-4-1-release-notes.html

PR #14941, opened 2019-05-20
OpenShift Ansible Service Broker is not installed by default in 4.1

ahardin-rh:
@jianzhangbjz These edits are per your comments in https://github.com/openshift/openshift-docs/pull/13726


jianzhangbjz:
LGTM, thanks!

PR #14948, opened 2019-05-20
Replace use of snip with dots

jboxman:


jboxman:
@openshift/team-documentation switched to dots

mburke5678:
@jboxman LGTM

PR #14955, opened 2019-05-21
[enterprise-3.11] Remove  deprecated "Recycle" reclaim policy term

bysnupy:
- Version: only `v3.11`

- Description:
  `Recycle` reclaim policy is deprecated as of `v3.6`.
  Until `v3.10`, there is a alert message about `recycler is deprecated`.
  But it's removed now. So we should remove this deprecated term either for not being confusing.

- Related Information:
  [Openshift Volume Recycler Now Deprecated](https://docs.openshift.com/container-platform/3.6/release_notes/ocp_3_6_release_notes.html)
~~~
Openshift Volume Recycler is being deprecated. Anyone using recycler should use dynamic provision and volume deletion instead.
~~~

### DIFF
--- a/architecture/additional_concepts/storage.adoc
+++ b/architecture/additional_concepts/storage.adoc
@@ -123,7 +123,7 @@ to policy.
 === Reclaim volumes
 
 The reclaim policy of a `PersistentVolume` tells the cluster what to do with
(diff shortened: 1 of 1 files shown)


kalexand-rh:
@vikram-redhat, this seems reasonable. I'm going to merge.

PR #14977, opened 2019-05-21
Bug 1712245: Known issue with build env vars

adambkaplan:
Builds which use shell substitution to populate an environment variable may fail.

https://bugzilla.redhat.com/show_bug.cgi?id=1712245


bmcelvee:
Thanks, @adambkaplan! LGTM

Minor change, no QE need

PR #14978, opened 2019-05-21
clarify custom build w/ buildah title

bparees:
buildah isn't adding a docker socket, it's replacing the need for a docker socket.


bmcelvee:
Thanks! LGTM 

Minor change, no QE needed.

PR #14982, opened 2019-05-21
PR14978-followup

bmcelvee:
Changed topic map reference

https://github.com/openshift/openshift-docs/pull/14978


bmcelvee:
Minor change, no QE needed

PR #14983, opened 2019-05-21
Update configuring_authentication.adoc

csheremeta:
Fixed typo caused by strange parsing of underscores in string


vikram-redhat:
Fixed via: https://github.com/openshift/openshift-docs/pull/15080

PR #14985, opened 2019-05-21
BZ-1712232: Updated list of supported PVs.

huffmanca:
Updated list of supported PVs; this now includes NFS, vSphere, Fibre Channel and HostPath.

This is for OS 4.x.


liangxia:
LGTM ( from QE )

PR #14989, opened 2019-05-21
Document build and image configuration resources

bmcelvee:


_topic_map.yml
```diff
@@ -328,6 +331,9 @@ Topics:
   File: create-images
 - Name: Managing imagestreams
   File: image-streams-manage
+- Name: Image controller configuration resources
```
bmcelvee: Is controller correct here?

_topic_map.yml
```diff
@@ -312,6 +312,9 @@ Topics:
 - Name: Securing builds by strategy
   File: securing-builds-by-strategy
   Distros: openshift-enterprise,openshift-origin
+- Name: Build controllerconfiguration resources
+  File: build-configuration
```
bmcelvee: Is controller correct here?

modules/images-configuration-parameters.adoc
```diff
@@ -0,0 +1,53 @@
...
+
+|`externalRegistryHostnames`
+|Provides the hostnames for the default external image registry. The external
+hostname should be set only when the image registry is exposed externally. The
+first value is used in 'publicDockerImageRepository' field in ImageStreams.
+The value must be in `hostname[:port]` format.
+
+|`additionalTrustedCA`
+|A reference to a ConfigMap containing additional CAs that should be trusted
+during ImageStream import, pod image pull, and `openshift-image-registry`
```
bmcelvee: Should these be more along the lines of: 

`ImageStream import`, `pod image pull`, and `openshift-image-registry pullthrough`

bmcelvee:
@bparees here's the first draft of the config content, please take a look when you get a chance

thanks!

_topic_map.yml
bmcelvee: Also, should this move to the Registry bucket?

bparees: no i think images is the right section for it.

bparees: > Is controller correct here?

no, it's just "image configuration resources" i guess.  it controls the configuration of how images get consumed on the cluster, primarily.

modules/builds-configuration-file.adoc
```diff
@@ -0,0 +1,69 @@
+// Module included in the following assemblies:
+//
+// * builds/build-configuration.adoc
+
+[id="builds-configuration-file_{context}"]
+= Configuring build settings
+
+You can configure build settings by editing the
+`build.config.openshift.io/cluster` file.
```
bparees: s/file/resource/

modules/builds-configuration-file.adoc
```diff
@@ -0,0 +1,69 @@
...
+
+.Procedure
+
+* Edit the `build.config.openshift.io/cluster` file:
```
bparees: s/file/resource/

modules/builds-configuration-file.adoc
```diff
@@ -0,0 +1,69 @@
...
++
+----
+$ oc edit build.config.openshift.io/cluster
+----
++
+The following is an example `build.config.openshift.io/cluster` file:
```
bparees: s/file/resource/

modules/builds-configuration-file.adoc
```diff
@@ -0,0 +1,69 @@
...
+  buildOverrides:
+    imageLabels:
+    - name: labelkey
+      value: labelvalue
+    nodeSelector:
+      selectorkey: selectorvalue
+    tolerations:
+    - effect: NoSchedule
+      key: node-role.kubernetes.io/builds
+operator: Exists
```
bparees: would be nice if we can annotate this w/ footnotes that give a little more detail on each field

modules/builds-configuration-parameters.adoc
```diff
@@ -0,0 +1,77 @@
...
+
+The `build.config.openshift.io/cluster` resource offers the following
+configuration parameters.
+
+[cols="3a,8a",options="header"]
+|===
+|Parameter |Description
+
+|`spec`
+|`BuildSpec` holds user-settable values for the build controller configuration.
```
bparees: don't need to doc spec itself, it's just a container for the other fields.

modules/builds-configuration-parameters.adoc
```diff
@@ -0,0 +1,77 @@
...
+|`additionalTrustedCA`
+|A reference to a ConfigMap containing additional CAs that should be trusted for
+image pushes and pulls during builds. The namespace for this ConfigMap is
+`openshift-config`.
+
+|`buildDefaults`
+|Controls the default information for builds.
+
+|`buildOverrides`
+|Controls override settings for builds.
```
bparees: can we find a way to structure this so it's obvious which fields are part of "buildDefaults" and which are part of "buildOverrides"?

modules/builds-configuration-parameters.adoc
```diff
@@ -0,0 +1,77 @@
...
+
+|`items`
+|Standard object's metadata.
+
+
+|===
+
+[NOTE]
+====
+`spec` is the only required parameter. All other parameters are optional.
```
bparees: i don't believe spec is required either.

modules/images-configuration-file.adoc
```diff
@@ -0,0 +1,47 @@
+// Module included in the following assemblies:
+//
+// * openshift_images/image-configuration.adoc
+
+[id="images-configuration-file_{context}"]
+= Configuring image registry settings
```
bparees: i don't love this title because people will probably assume it means the cluster registry when in reality there are fields here related to both external registries (CAs, allowed registries, etc)  as well as the cluster registry (internalRegistryHostname, externalRegistryHostname)

modules/images-configuration-file.adoc
```diff
@@ -0,0 +1,47 @@
...
+      insecure: false
+  additionalTrustedCA:
+    name: myconfigmap
+  registrySources:
+    insecureRegistries:
+    - insecure.com
+    blockedRegistries:
+    - untrusted.com
+status:
+  internalRegistryHostname: image-registry.openshift-image-registry.svc:5000
```
bparees: would be good to annotate w/ footnotes

modules/images-configuration-parameters.adoc
```diff
@@ -0,0 +1,85 @@
...
+
+The `image.config.openshift.io/cluster` resource offers the following
+configuration parameters.
+
+[cols="3a,8a",options="header"]
+|===
+|Parameter |Description
+
+|`metadata`
+|Standard object's metadata.
```
bparees: don't need to mention

modules/images-configuration-parameters.adoc
```diff
@@ -0,0 +1,85 @@
...
+
+|`spec`
+|Holds user-settable values for configuration.
+
+|`status`
+|Holds observed values from the cluster.
```
bparees: unlike what i said for the build object, maybe there is value in making it clear that the spec subsection is for user editing and the status subsection is for observed settings.  But they are subsections, not fields themselves, effectively.

modules/images-configuration-parameters.adoc
bparees: i guess

modules/images-configuration-parameters.adoc
```diff
@@ -0,0 +1,85 @@
...
+and that you want applications to be able to import from. Users with permission
+to create images or `ImageStreamMappings` from the API are not affected by this
+policy. Typically only cluster administrators will have the appropriate permissions.
+
+|`additionalTrustedCA`
+|A reference to a ConfigMap containing additional CAs that should be trusted
+during ImageStream import, pod image pull, and `openshift-image-registry`
+pullthrough.
+
+The namespace for this ConfigMap is `openshift-config`.
```
bparees: "The format of the configmap is to use the registry hostname as the key, and the base64-encoded certificate as the value, for each additional registry CA to trust"

modules/images-configuration-parameters.adoc
```diff
@@ -0,0 +1,85 @@
...
+
+|`registrySources`
+|Contains configuration that determines how the container runtime should treat
+individual registries when accessing images for build sand pods. For instance,
```
bparees: s/build sand pods/builds and pods/
(and if that was copied from our api doc then it needs to be fixed there too)

modules/images-configuration-parameters.adoc
```diff
@@ -0,0 +1,85 @@
...
+|Registries which do not have a valid SSL certificate or only support HTTP
+connections.
+
+|`blockedRegistries`
+|Whitelisted for image pull and push actions. All other registries are blocked.
+
+Only one of `blockedRegistries` or `allowedRegistries` may be set
+
+|`allowedRegistries`
+|Whitelisted for image pull and push actions. All other registries are blocked.
```
bparees: these items should be nested under registrySources

modules/images-configuration-parameters.adoc
bmcelvee: That was just my typo. :/ It's "builds+pods" in the API docs.

modules/builds-configuration-file.adoc
bmcelvee: Added

modules/images-configuration-file.adoc
bmcelvee: Added

bmcelvee:

bmcelvee:
@openshift/team-documentation please peer review, thanks!

modules/builds-configuration-file.adoc
```diff
@@ -0,0 +1,92 @@
...
+should be trusted for image pushes and pulls during builds. The namespace for
+this ConfigMap is `openshift-config`.
+<3> `buildDefaults`: Controls the default information for builds.
+<4> `defaultProxy`: Contains the default proxy settings for all build operations,
+including image pull or push and source download.
+<5> `env`: A set of default environment variables that are applied to the build if
+the specified variables do not exist on the build.
+<6> `gitProxy`: Contains the proxy settings for Git operations only. If set, this
+overrides any Proxy settings for all Git commands, such as `git clone`.
+<7> `imageLabels`: A list of docker labels that are applied to the resulting image.
```
ahardin-rh: Should these instances of "docker" be replaced with "container"?

modules/builds-configuration-parameters.adoc
```diff
@@ -0,0 +1,68 @@
...
+
+`gitProxy`: Contains the proxy settings for Git operations only. If set, this
+overrides any Proxy settings for all Git commands, such as `git clone`.
+
+Values that are not set here are inherited from DefaultProxy.
+
+`env`: A set of default environment variables that are applied to the build if
+the specified variables do not exist on the build.
+
+`imageLabels`: A list of docker labels that are applied to the resulting image.
```
ahardin-rh: Same here. Should instances of "docker" be replaced by "container" or something similar?

modules/images-configuration-file.adoc
```diff
@@ -0,0 +1,69 @@
...
+    blockedRegistries:<6>
+    - untrusted.com
+status:
+  internalRegistryHostname: image-registry.openshift-image-registry.svc:5000
+----
+<1> `Image`: Holds cluster-wide information about how to handle images. The
+canonical name is `cluster`.
+<2> `allowedRegistriesForImport`: Limits the container image registries from which
+normal users may import images. Set this list to the registries that you trust
+to contain valid Docker images, and that you want applications to be able to
```
ahardin-rh: s/Docker images/container images

ahardin-rh:
@bmcelvee Just a few questions regarding the usage of docker and Docker vs. container image. :+1:

modules/builds-configuration-file.adoc
```diff
@@ -0,0 +1,92 @@
...
+      value: labelvalue
+    nodeSelector:<11>
+      selectorkey: selectorvalue
+    tolerations:<12>
+    - effect: NoSchedule
+      key: node-role.kubernetes.io/builds
+operator: Exists
+----
+<1> `Build`: Holds cluster-wide information on how to handle builds. The
+canonical name is `cluster`.
```
bparees: it's not just canonical.....that's the only valid name.  if you name it anything else, it will have no effect.

modules/builds-configuration-file.adoc
```diff
@@ -0,0 +1,92 @@
...
+<2> `additionalTrustedCA`: A reference to a ConfigMap containing additional CAs that
+should be trusted for image pushes and pulls during builds. The namespace for
+this ConfigMap is `openshift-config`.
+<3> `buildDefaults`: Controls the default information for builds.
+<4> `defaultProxy`: Contains the default proxy settings for all build operations,
+including image pull or push and source download.
+<5> `env`: A set of default environment variables that are applied to the build if
+the specified variables do not exist on the build.
+<6> `gitProxy`: Contains the proxy settings for Git operations only. If set, this
+overrides any Proxy settings for all Git commands, such as `git clone`.
```
bparees: i guess this also overrides the defaultProxy, right @adambkaplan ?

modules/builds-configuration-file.adoc
bparees: i'd go with `image` not `container` but sure.  (even though "a list of image labels that are applied to the resulting image" feels redundant, it's more accurate than calling it a container)

modules/builds-configuration-parameters.adoc
bparees: same answer i gave above :)

bparees: maybe the best thing is to just remove docker.  "a list of labels that are applied to the resulting image".  removes the redundancy.

modules/images-configuration-file.adoc
```diff
@@ -0,0 +1,69 @@
...
+  registrySources:<4>
+    insecureRegistries:<5>
+    - insecure.com
+    blockedRegistries:<6>
+    - untrusted.com
+status:
+  internalRegistryHostname: image-registry.openshift-image-registry.svc:5000
+----
+<1> `Image`: Holds cluster-wide information about how to handle images. The
+canonical name is `cluster`.
```
bparees: same common on this being a required name, not canonical.

modules/images-configuration-file.adoc
```diff
@@ -0,0 +1,69 @@
...
+`openshift-image-registry pullthrough`. The namespace for this ConfigMap is
+`openshift-config`. The format of the ConfigMap is to use the registry hostname
+as the key, and the base64-encoded certificate as the value, for each additional
+registry CA to trust.
+<4> `registrySources`: Contains configuration that determines how the container
+runtime should treat individual registries when accessing images for builds and
+pods. For instance, whether or not to allow insecure access. It does not contain
+configuration for the internal cluster registry.
+<5> `insecureRegistries`: Registries which do not have a valid SSL certificate or
+only support HTTP connections.
```
bparees: s/SSL/TLS/

modules/images-configuration-file.adoc
```diff
@@ -0,0 +1,69 @@
...
+<6> `blockedRegistries`: Whitelisted for image pull and push actions. All other
+registries are blocked.
```
bparees: this looks wrong.  blockedRegistries are the blacklisted set, not the whitelisted set.

my example didn't include both "blockdRegistries" and "allowedRegistries" because you can't set both at the same time.  but i think you pasted the description of allowedRegistries here.

bparees (changes requested):

modules/builds-configuration-file.adoc
bmcelvee: @bparees should we replace docker labels and docker images with container labels and container images? And if so, should the API docs be updated reflect the change? Thanks!

bparees: `docker label` -> `image label`
`docker image` -> `image` or `container image`

and yes, api docs should be updated too.  this will all make @rhatdan very happy.

but for this particular sentence i'd just change it to "a list of labels that are applied to the resulting image" to avoid the redundancy.

modules/images-configuration-file.adoc
bmcelvee: Thanks! I thought it looked strange. This and the SSL -> TLS might also need to be updated in the API docs.

bmcelvee: Nevermind, just the SSL -> TLS bit.

vikram-redhat:
@bmcelvee ready for QE review?

bmcelvee:
@vikram-redhat yes it is ready for QE review, thanks! 

@wewang58 Please take a look, thanks!

modules/builds-configuration-file.adoc
```diff
@@ -0,0 +1,92 @@
...
+    release.openshift.io/create-only: "true"
+  creationTimestamp: "2019-05-17T13:44:26Z"
+  generation: 2
+  name: cluster
+  resourceVersion: "107233"
+  selfLink: /apis/config.openshift.io/v1/builds/cluster
+  uid: e2e9cc14-78a9-11e9-b92b-06d6c7da38dc
+spec:
+  additionalTrustedCA:<2>
+    name: myconfigmap
```
adambkaplan: @bmcelvee `additionalTrustedCA` should be deprecated and not referenced in the official docs. It doesn't do anything in 4.1.

Adding CAs should be done via `image.config.openshift.io/cluster`

modules/builds-configuration-file.adoc
```diff
@@ -0,0 +1,92 @@
...
+    nodeSelector:<11>
+      selectorkey: selectorvalue
+    tolerations:<12>
+    - effect: NoSchedule
+      key: node-role.kubernetes.io/builds
+operator: Exists
+----
+<1> `Build`: Holds cluster-wide information on how to handle builds. The
+canonical, and only valid name is `cluster`.
+<2> `additionalTrustedCA`: A reference to a ConfigMap containing additional CAs that
```
adambkaplan: Ditto - remove this reference.

modules/builds-configuration-file.adoc
adambkaplan: @bparees no- this only sets the proxy for the git clone init container.

modules/builds-configuration-parameters.adoc
```diff
@@ -0,0 +1,68 @@
...
+|Holds cluster-wide information on how to handle builds. The canonical, and only
+valid name is `cluster`.
+
+`spec`: Holds user-settable values for the build controller
+configuration.
+
+|`BuildSpec`
+|`additionalTrustedCA`: A reference to a ConfigMap containing additional CAs that
+should be trusted for image pushes and pulls during builds. The namespace for
+this ConfigMap is `openshift-config`.
```
adambkaplan: Remove `addtionalTrustedCA` as it doesn't do anything.

modules/images-configuration-file.adoc
```diff
@@ -0,0 +1,69 @@
...
+canonical, and only valid name is `cluster`.
+<2> `allowedRegistriesForImport`: Limits the container image registries from which
+normal users may import images. Set this list to the registries that you trust
+to contain valid images, and that you want applications to be able to
+import from. Users with permission to create images or `ImageStreamMappings`
+from the API are not affected by this policy. Typically only cluster
+administrators will have the appropriate permissions.
+<3> `additionalTrustedCA`: A reference to a ConfigMap containing additional CAs that
+should be trusted during `ImageStream import`, `pod image pull`, and
+`openshift-image-registry pullthrough`. The namespace for this ConfigMap is
```
adambkaplan: Also include builds in this list

modules/images-configuration-parameters.adoc
```diff
@@ -0,0 +1,82 @@
...
+|`allowedRegistriesForImport`: Limits the container image registries from which
+normal users may import images. Set this list to the registries that you trust
+to contain valid images, and that you want applications to be able to
+import from. Users with permission to create images or `ImageStreamMappings`
+from the API are not affected by this policy. Typically only cluster
+administrators will have the appropriate permissions.
+
+`additionalTrustedCA`: A reference to a ConfigMap containing additional CAs that
+should be trusted during `ImageStream import`, `pod image pull`, and
+`openshift-image-registry pullthrough`.
```
adambkaplan: Also include builds in this list

adambkaplan (changes requested):
@bmcelvee looks like we missed an item during our api review.

The `additionalTrustedCA` field in the build config spec object doesn't do anything. Configuring `additionalTrustedCA`s is only done on the `image.config.openshift.io/cluster` object.

It is unfortunately too late to do anything for 4.1.0 - the best we can do is mark this field as deprecated in 4.2 (_maybe_ backport the deprecation note to 4.1.z)  :man_shrugging:

bmcelvee:
Going ahead and merging. We can make any additional edits in follow-up PRs.

modules/builds-configuration-file.adoc
bparees: @adambkaplan I phrased it poorly but what i meant is "if you have a defaultProxy set and then you also set this, this value takes precedence(overrides) for git clone operations".

(and if you only have defaultProxy set, it is used for git clone).

it might be worth stating that clearly so people understand the relationship between defaultProxy and gitProxy (assuming i'm remembering it correctly, i think we went through some iterations on it)

PR #14994, opened 2019-05-22
Build changes and fixes

vikram-redhat:
@openshift/team-documentation @openshift/team-aligned-docs FYI.

This fixes the current issue with the build where some links were not getting checked by Travis. This was due to our recent move to use "" for anchor ids instead of ''. The build script was just not equipped to handle that and the links were not getting properly checked.

I have fixed the script and fixed several anchor ids. 

Note that our guidance on xrefs hasn't changed. When using an xref you *must* link to an anchor id, not just the page/assembly, even if it is to the top of the page/assembly (https://github.com/openshift/openshift-docs/blob/enterprise-4.1/contributing_to_docs/doc_guidelines.adoc#links-to-internal-content).


mburke5678:
Thank you!!

PR #15003, opened 2019-05-22
UDP services not supported with Kuryr SDN edit

maxwelldb:
Edit for #14680 

Need a review from @openshift/team-documentation

PR #15008, opened 2019-05-22
UDP services not supported with Kuryr SDN edit - 3.11

maxwelldb:
Manual duplication of #15003 for 3.11.


maxwelldb:
Let's do this again after I rebase it locally. :)

PR #15010, opened 2019-05-22
Update TOC for networking

jboxman:
@openshift/team-documentation, first attempt at refreshing the TOC tree for _networking_. Items under `openshift-sdn` are specific to OpenShift SDN CNI plug-in and aren't applicable to any future vendor CNI plug-ins.

I need to add _About OpenShift SDN_ with updated content from 3.11.

I don't know if Ingress/DNS need to be listed higher/first.

@bmcelvee, fyi


kalexand-rh (commented):
I'm not familiar enough with the content to say if it's an organizational improvement, but I p. sure that you are. LGTM.

PR #15015, opened 2019-05-22
BZ1698640 mixed environment upgrades

jhoyt-rh:
For bug BZ1698640. This fix applies only to release 3.9, because the fix was covered in enterprise-3.10 and enterprise-3.11 by [PR 14638](https://github.com/openshift/openshift-docs/pull/14638). 

This is a separate fix because the file structure had changed after enterprise-3.9, and  it was not able to be cherrypicked when enterprise-3.11 and enterprise-3.10 fixes were applied. My instructions were to create a separate PR for this bug's enterprise-3.9 fix.

### DIFF
--- a/upgrading/automated_upgrades.adoc
+++ b/upgrading/automated_upgrades.adoc
@@ -719,12 +719,24 @@ OpenShift upgrades] for further guidance.
 [[special-considerations-for-mixed-environments]]
 == Special Considerations for Mixed Environments
 
-Mixed environment upgrades (for example, those with Red Hat Enterprise Linux and
-Red Hat Enterprise Linux Atomic Host) require setting both
-`openshift_pkg_version` and `openshift_image_tag`. In mixed environments,  if
-you only specify `openshift_pkg_version`, then that number is used for the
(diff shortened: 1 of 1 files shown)


jhoyt-rh:
@vikram-redhat Will you PTAL and add labels, milestones, etc. This is for Enterprise-3.9 only. Thank you.

vikram-redhat:
It still has some lines from https://github.com/openshift/openshift-docs/pull/14809/files#diff-40f3b73686ab98d0e3b12958011ae7bfL734

jhoyt-rh:
Regarding vikram-redhat comment about still retaining lines not associated with this bug -- I have removed them, returning that section of the file to its previous content. That material will be covered in another bug [BZ1698198](https://bugzilla.redhat.com/show_bug.cgi?id=1698198) with [PR](https://github.com/openshift/openshift-docs/pull/14809).

kalexand-rh:
Thanks @jhoyt-rh! The text matches the change in the original PR now, so I'm going to merge.

PR #15017, opened 2019-05-23
telemetry

kalexand-rh:


kalexand-rh:
@reestr, will you PTAL a look at the way we're talking about Telemetry? We've gotten some confusing information about it and don't want to pass on the confusion to customers. :)

The other place we talk about it is here: https://github.com/openshift/openshift-docs/blob/3a171d826bc8fb7065a49701461056ed8daaab48/modules/telemetry-consequences-of-disabling-telemetry.adoc

kalexand-rh:
Tim and I fixed this PR on a call. After Max's telemetry docs go live, the install docs need to link to them. I have a task on the last-minute epic on our board to do this.

PR #15019, opened 2019-05-23
fix(network-connectivity-requirements): Typo [15014]

adityahiran:
@openshift/team-documentation


adityahiran:
Will create a new one

PR #15027, opened 2019-05-24
Add curl -k option as installer won't install OpenShift CA cert into …

nekop:
…OS in 3.10 or later.

The curl command needs `-k` insecure option otherwise it failed with TLS error in 3.10+.


kalexand-rh:
@xingxingxia, will you PTAL?

xingxingxia:
LGTM

PR #15034, opened 2019-05-24
bug 1713474 m5 instances

kalexand-rh:
@wking, re: https://bugzilla.redhat.com/show_bug.cgi?id=1713474

### DIFF
--- a/modules/installation-aws-limits.adoc
+++ b/modules/installation-aws-limits.adoc
@@ -36,9 +36,9 @@ different instance type, review your account limits to ensure that your cluster
 can deploy the machines that you need.
 
 In most regions, the bootstrap and worker machines uses an `m4.large` machines
-and the master machines use `m4.xlarge` machines. If these machine types
-are not available in the region that you select, `m5.large` and `m5.xlarge`
-machines are used instead.
+and the master machines use `m4.xlarge` instances. In some regions, including
+all regions that do not support these instance types, `m5.large` and `m5.xlarge`
+instances are used instead.
 
 |Elastic IPs (EIPs)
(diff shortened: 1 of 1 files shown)


modules/installation-aws-limits.adoc
```diff
@@ -36,9 +36,10 @@ different instance type, review your account limits to ensure that your cluster
 can deploy the machines that you need.
 
 In most regions, the bootstrap and worker machines uses an `m4.large` machines
-and the master machines use `m4.xlarge` machines. If these machine types
-are not available in the region that you select, `m5.large` and `m5.xlarge`
-machines are used instead.
+and the master machines use `m4.xlarge` instances. In some regions, including
+all regions that do not support these instance types, `m5.large` and `m5.xlarge`
+instances are used instead. In future releases, `m5.large` and `m5.xlarge`
+instances will be used by default in more regions.
```
wking: While a change to m5 defaults [is likely][1], I don't see a need to commit to it here.  Maybe just "Future releases may have different defaults" or some such.  For example, we may become more memory-efficient and be able to use smaller control-plane types, or decide that control-plane or compute machines on most clusters would be better served by some non-M type, etc.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1710981

kalexand-rh: Or I'll just pull the sentence. :)

kalexand-rh:
@wking, instead of saying anything about what we might do with the instances in the future, I pulled the sentence. When that bug merges, I'll update the docs. Please let me know if we make any major changes here.

PR #15035, opened 2019-05-24
Updated anchor ID format

ahardin-rh:
@vikram-redhat @sheriff-rh FYI


sheriff-rh:
Ah okay this makes a lot more sense. Thanks @ahardin-rh!

PR #15042, opened 2019-05-28
bug 1713638 adding vSphere version

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1713638

preview build here: http://file.rdu.redhat.com/kalexand/052819/BZ1713638/installing/installing_vsphere/installing-vsphere.html#installation-overview-installing-vsphere

@jianlinliu, will you PTAL?

### DIFF
--- a/modules/installation-vsphere-infrastructure.adoc
+++ b/modules/installation-vsphere-infrastructure.adoc
@@ -1,0 +1,9 @@
+// Module included in the following assemblies:
+//
+// * installing/installing_vsphere/installing-vsphere.adoc
+
+[id="installation-vsphere-infrastructure_{context}"]
+= VMware vSphere infrastructure requirements
+
+You must install the {product-title} cluster on a VMware vSphere version 6.5 or
+6.7 instance.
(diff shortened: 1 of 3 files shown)


modules/installation-vsphere-infrastructure.adoc
```diff
@@ -0,0 +1,9 @@
+// Module included in the following assemblies:
+//
+// * installing/installing_vsphere/installing-vsphere.adoc
+
+[id="installation-vsphere-infrastructure_{context}"]
+= VMware vSphere infrastructure requirements
+
+You must install the {product-title} cluster on a VMware vSphere instance that
+is version 6.5 or later.
```
jianlinliu: Hmm, "later"? I think this is a bit aggressive, and QE only run 4.1 testing on 6.5, but not 6.5+, here is the detailed info for vsphere version:
vCenter: 6.5.0-10964411
Hypervisor: VMware ESXi, 6.5.0, 7967591

kalexand-rh: @dav1x, @reestr, the email thread said that while the original validation was completed on 6.5 that you were going to complete more testing on 6.7. How would you like that to be reflected in the docs?

kalexand-rh:
We figured out the right wording, and Jianlin approved the change on the bug. Merging.

PR #15044, opened 2019-05-28
bug 1714539 syncing cloudformation templates

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1714539

@jianlinliu, will you PTAL?


kalexand-rh:
Jianlin approved this change on the source bug. Merging.

PR #15047, opened 2019-05-28
BZ-1714182: Removed TP notice from NFS.

huffmanca:
Removed technology preview notice from NFS. I also examined the other storage sections, and don't see any other references to this feature being TP.

This is for OS 4.x.


liangxia:
LGTM ( from QE )

PR #15052, opened 2019-05-28
Fix Cluster Autoscaler Example

stencell:
- Correct wording for <11>
  - This example is for scaling down or deleting nodes...not deploying.
- Change examples in scaleDown section
  - Using 10s for all of these is probably not practical, even though it is only an example. Having such artificially low values would causing scale down actions to happen too quickly


kalexand-rh:
@sunzhaohua2, will you PTAL?

jhou1:
Looks good!

PR #15054, opened 2019-05-28
Add information about allowed VXLAN ports on AWS

jboxman:
This is useful because the AWS SG is defined such that only a specific port range is valid for this configuration.

@openshift/team-documentation, PTAL. Thanks!

### DIFF
--- a/modules/nw-operator-cr.adoc
+++ b/modules/nw-operator-cr.adoc
@@ -77,6 +77,11 @@ then you must set this explicitly to 50 less than the smallest node MTU value.
 <5> The port to use for all VXLAN packets. The default value is `4789`. If you
 are running in a virtualized environment with existing nodes that are part of
 another VXLAN network then you might be required to change this.
+ifdef::install[]
++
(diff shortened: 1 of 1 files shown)


jboxman:
@zhaozhanqi can you PTAL? Thanks!

modules/nw-operator-cr.adoc
```diff
@@ -77,6 +77,11 @@ then you must set this explicitly to 50 less than the smallest node MTU value.
 <5> The port to use for all VXLAN packets. The default value is `4789`. If you
 are running in a virtualized environment with existing nodes that are part of
 another VXLAN network then you might be required to change this.
+ifdef::install[]
++
+On AWS, you can select an alternate port for the VXLAN between port `9000` and
```
kalexand-rh: Amazon Web Services

Isn't the network operator defined for only AWS right now?

Is the default value acceptable on AWS? It's not in the range that you can specify.

kalexand-rh (commented):
Couple of questions. If the default value works on AWS, it LGTM.

zhaozhanqi:
LGTM

PR #15056, opened 2019-05-29
WIP: Cnv migration

apinnick:


vikram-redhat:
@ousleyp PTAL.

ousleyp:
Hi @apinnick - I'm happy to help with a review soon. First, can you please clarify something? It looks like the files here are all still CNV 1.4-style modules, but for docs published in the OCP 4.1 collection, they need to be made into assemblies (one assembly per user story). The modules that the assemblies include also need to be updated to reflect the modular docs standards (anchor ID, header level). Are you planning to do the modularization part of this work in this PR, a different PR, or am I missing something?

In case you need them, the OCP 4.1 guidelines can be found here: https://github.com/openshift/openshift-docs/blob/enterprise-4.1/contributing_to_docs/doc_guidelines.adoc

Thanks!

apinnick:
Good point! When I created the module, the CNV guide hadn't yet been moved over, so there was no convenient framework. I'll create an assembly.

apinnick:
On second thought, I think the assembly should be created by OpenShift writers, so let's not review it until more of the CNV users guide has been moved over.

PR #15058, opened 2019-05-29
bug 1713125 adding pxe booting

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1713125

preview build here: http://file.rdu.redhat.com/kalexand/052919/BZ1713125/installing/installing_bare_metal/installing-bare-metal.html#installation-upi-machines-pxe-installing-bare-metal

@morenod PTAL?


kalexand-rh:
David approved this change on the source bug. Merging.

PR #15059, opened 2019-05-29
bug 1715118 updating parameter value

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1715118

@mazzystr, PTAL?

### DIFF
--- a/modules/installation-creating-aws-bootstrap.adoc
+++ b/modules/installation-creating-aws-bootstrap.adoc
@@ -78,11 +78,11 @@ requires:
 [
   {
     "ParameterKey": "InfrastructureName", <1>
-    "ParameterValue": "mycluster-vw9j6" <2>
+    "ParameterValue": "mycluster-<random_string>" <2>
   },
   {
     "ParameterKey": "RhcosAmi", <3>
-    "ParameterValue": "ami-07f604c5f18a1e7a9" <4>
+    "ParameterValue": "ami-<random_string>" <4>
   },
   {
     "ParameterKey": "AllowedBootstrapSshCidr", <5>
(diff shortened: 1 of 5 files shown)


modules/installation-creating-aws-bootstrap.adoc
```diff
@@ -110,19 +110,19 @@ requires:
   },
   {
     "ParameterKey": "RegisterNlbIpTargetsLambdaArn", <17>
-    "ParameterValue": "arn:aws:lambda:<zone>:269733483066:function:<dns_stack>-RegisterNlbIpTargets-NHEKB9IHN5XZ" <18>
+    "ParameterValue": "arn:aws:lambda:<zone>:<account_number>:function:<dns_stack>-RegisterNlbIpTargets-NHEKB9IHN5XZ" <18>
```
mazzystr: zone needs to be changed to region

I don't know what <dns_stack> is.  There's no bit in the documentation describing what should go in this position.  I can guess from what actually gets deployed is it's <INFRASTRUCTURE NAME>-<STACK NAME>-RegisterNlbIpTargets-<RANDOM HASH>
This makes the example look obnoxious ...
`"ParameterValue": "arn:aws:lambda:<region>:<account_number>:function:<infrastructure name>-<stack name>-RegisterNlbIpTargets-<random hash>"`
It's verbose but clearer than before.  The excessive length is because the resource names are too long.  <Stack name> and <infrastructure name> doesn't need to be in the resource name.  The random hash is going to prevent name conflict.  They're all tagged correctly so well searchable.

AWS uses the following generic examples for Lamda ...
```
arn:aws:lambda:region:account-id:function:function-name
arn:aws:lambda:region:account-id:function:function-name:alias-name
arn:aws:lambda:region:account-id:function:function-name:version
arn:aws:lambda:region:account-id:event-source-mappings:event-source-mapping-id
```

This is taken from [here](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html).  Then web browser search for `AWS Lambda`

kalexand-rh: @mazzystr, I'm making some updates based on you feedback. Thanks! Will you PTAL?

mazzystr:
LGTM!  Thank you!

kalexand-rh:
Thanks Chris! Merging.

modules/installation-creating-aws-worker.adoc
```diff
@@ -41,19 +41,19 @@ template requires:
...
+    "ParameterValue": "mycluster-<random_string>" <2>
   },
   {
     "ParameterKey": "RhcosAmi", <3>
-    "ParameterValue": "ami-07f604c5f18a1e7a9" <4>
+    "ParameterValue": "ami-<random_string>" <4>
   },
   {
     "ParameterKey": "WorkerSubnet", <5>
     "ParameterValue": "yes" <6>
```
mazzystr: Sorry ... found another problem. This parameter key should be called PrivateSubnet. The Value should be sg-<random> string. The parameter should be modified in the template yaml also @wking

kalexand-rh: https://github.com/openshift/openshift-docs/pull/15086

PR #15073, opened 2019-05-30
Administrator CLI commands: sample upgrade version should be 4.1.1

vrutkovs:
Avoid mentioning 4.0 here


kalexand-rh:
@vikram-redhat, this looks reasonable.
@jiajliu, FYI

PR #15075, opened 2019-05-30
Administrator CLI commands: use 4.1 example in release example

vrutkovs:


kalexand-rh:
@vrutkovs, I think we should show the delta between the last beta version and GA.

vrutkovs:
@kalexand-rh good idea, updated

kalexand-rh:
Thanks @vrutkovs!

@mffiedler, will you please +1? This looks right to me.

mffiedler (approved):
/lgtm

kalexand-rh:
Thank you both! Merging.

PR #15084, opened 2019-05-30
BZ-1715395: Updated block volume support.

huffmanca:
Updated the list of block volume support. I've removed the previous sentence regarding only manually provisioned volumes being supported, and added a table where it used to be.

This is for OS 4.x.


liangxia:
LGTM ( from QE )

PR #15086, opened 2019-05-30
more CF updates

kalexand-rh:
For after https://github.com/openshift/installer/pull/1794/ merges

@mazzystr, PTAL

### DIFF
--- a/modules/installation-creating-aws-worker.adoc
+++ b/modules/installation-creating-aws-worker.adoc
@@ -13,6 +13,12 @@ You can create worker nodes in Amazon Web Services (AWS) for your cluster to use
 The easiest way to manually create these nodes is to modify the provided
 CloudFormation template.
 
+[IMPORTANT]
+====
+The CloudFormation template creates a stack that represents one worker machine.
+You must create a stack for each worker machine.
+====
+
 [NOTE]
 ====
 If you do not use the provided CloudFormation template to create your worker
@@ -48,8 +54,8 @@ template requires:
     "ParameterValue": "ami-<random_string>" <4>
   },
   {
-    "ParameterKey": "WorkerSubnet", <5>
-    "ParameterValue": "yes" <6>
+    "ParameterKey": "Subnet", <5>
+    "ParameterValue": "subnet-<random_string>" <6>
   },
   {
     "ParameterKey": "WorkerSecurityGroupId", <7>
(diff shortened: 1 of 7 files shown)


mazzystr (approved):

mazzystr:
LGTM!  Thank you!

modules/installation-creating-aws-worker.adoc
```diff
@@ -48,8 +54,8 @@ template requires:
     "ParameterValue": "ami-<random_string>" <4>
   },
   {
-    "ParameterKey": "WorkerSubnet", <5>
-    "ParameterValue": "yes" <6>
+    "ParameterKey": "PrivateSubnet", <5>
+    "ParameterValue": "sg-<random_string>" <6>
```
jianlinliu: Copy/paste error? Obviously here should be "Subnet" as "ParameterKey" per CF template change.

kalexand-rh: Thank you, @jianlinliu! The dev source changed again after I edited this file. I am fixing this error.

jianlinliu: A minor suggestion, seen from the "sg-<random_string>", looks like "ParameterValue" is asking some security group name for "Subnet", if possible, pls make the example value for "ParameterValue" more meaningful, but not confuse user.

kalexand-rh: You're right. I changed it to "subnet-<random-string>." Do you see any other changes?

modules/installation-creating-aws-worker.adoc
```diff
@@ -80,8 +86,7 @@ file metadata, which has the format `<cluster-name>-<random-string>`.
 <3> Current {op-system-first} AMI to use for the worker nodes.
 <4> Specify an `AWS::EC2::Image::Id` value.
 <5> A subnet, preferably private, to launch the worker nodes on.
-<6> Specify a subnet from the `PrivateSubnets` value from the output of the
-CloudFormation template for DNS and load balancing.
+<6> Specify a subnet for your VPC.
```
jianlinliu: I think the original description is more clear and reasonable. Why we need such change here?

modules/installation-creating-aws-worker.adoc
jianlinliu: lgtm now.

modules/installation-creating-aws-worker.adoc
kalexand-rh: Good point. I've restored the original comment.

kalexand-rh:
Merging.

PR #15092, opened 2019-05-31
Added new welcome index based on Chris's work

vikram-redhat:


vikram-redhat:
A majority of the links are not in the correct format, and I will be fixing them in the next few days.

PR #15097, opened 2019-05-31
web console draft

kalexand-rh:
built draft here: http://file.rdu.redhat.com/kalexand/053119/console/web-console/configuring-web-console.html

@spadgett, @benjaminapetersen, will you PTAL?

### DIFF
--- a/modules/web-console-configuration.adoc
+++ b/modules/web-console-configuration.adoc
@@ -33,4 +33,5 @@ If you do not specify a value, the user returns to the login page for the
 web console. Specifying a `logoutRedirect` URL allows your users to perform
 single log out (SLO) through the identity provider to destroy their single
 sign-on session.
-<2> The web console URL.
+<2> The web console URL. You cannot modify this parameter value. If you do, the
+cluster resets it to the default value.
--- a/modules/web-console-configuration.adoc
+++ b/modules/web-console-configuration.adoc
@@ -1,0 +1,41 @@
+// Module included in the following assemblies:
+//
+// * web_console/configuring-web-console.adoc
+
+[id="web-console-configuration_{context}"]
+= Configuring the web console
+
+You can configure image registry settings by editing the
+`image.config.openshift.io/cluster` resource.
+
+* Edit the `consoles.operator.openshift.io resource:
++
+----
+$ oc edit consoles.operator.openshift.io
+----
++
+The following example displays the parameters from this resource that you can
+modify:
++
+[source,yaml]
+----
+apiVersion: config.openshift.io/v1
+kind: Console
+metadata:
+  name: cluster
+spec:
+  managementState: Managed <1>
+  authentication:
+    logoutRedirect: "" <2>
+status:
+  consoleURL: "" <3>
+----
+<1> Specify the state of the console, `Managed` or `Disabled`. If the value is
+`Managed`, then the web console is enabled. Set it to `Disabled` to disable
+the console.
+<2> Specify the URL of the page to load when a user logs out of the web console.
+If you do not specify a value, the user returns to the login page for the
+web console. Specifying a `logoutRedirect` URL allows your users to perform
+single log out (SLO) through the identity provider to destroy their single
+sign-on session.
+<3> The web console URL. Do not modify this value.
--- a/modules/web-console-configuration.adoc
+++ b/modules/web-console-configuration.adoc
@@ -5,17 +5,16 @@
 [id="web-console-configuration_{context}"]
 = Configuring the web console
 
-You can configure image registry settings by editing the
-`image.config.openshift.io/cluster` resource.
+You can configure the web console settings by editing the
+`console.config.openshift.io` resource.
 
-* Edit the `consoles.operator.openshift.io resource:
+* Edit the `console.config.openshift.io` resource:
 +
 ----
-$ oc edit consoles.operator.openshift.io
+$ oc edit console.config.openshift.io cluster
 ----
 +
-The following example displays the parameters from this resource that you can
-modify:
+The following example displays the sample resource definition for the console:
 +
 [source,yaml]
 ----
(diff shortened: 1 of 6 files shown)


modules/web-console-configuration.adoc
```diff
@@ -0,0 +1,41 @@
...
+  managementState: Managed <1>
+  authentication:
+    logoutRedirect: "" <2>
+status:
+  consoleURL: "" <3>
+----
+<1> Specify the state of the console, `Managed` or `Disabled`. If the value is
+`Managed`, then the web console is enabled. Set it to `Disabled` to disable
+the console.
+<2> Specify the URL of the page to load when a user logs out of the web console.
```
kalexand-rh: @benjaminapetersen, is this right?
@huffmanca, have you heard about this SSO requirement?

modules/web-console-configuration.adoc
```diff
@@ -0,0 +1,41 @@
+// Module included in the following assemblies:
+//
+// * web_console/configuring-web-console.adoc
+
+[id="web-console-configuration_{context}"]
+= Configuring the web console
+
+You can configure image registry settings by editing the
```
benjaminapetersen: `image registry` copy/paste error?

modules/web-console-configuration.adoc
```diff
@@ -0,0 +1,41 @@
...
+`image.config.openshift.io/cluster` resource.
```
benjaminapetersen: So there are 2 different configs that can affect the console, you have them combined into one config here:

```bash
oc edit console.config.openshift.io cluster
```
for the top level config (this is the user facing config):

```yaml
kind: Console
metadata:
  name: cluster
spec: 
  authentication:
    # you can configure this for single sign-out
    # this comment explains the feature:
    #  - https://github.com/openshift/api/blob/master/config/v1/types_console.go#L44
    logoutRedirect:  https://www.somewhere.com/idp-logout-redirect
status:
  # we report this to you for your convenience. You can't change it,
  # it will just be put back to the correct value.
  consoleURL: https://console-openshift-console.apps.<your-cluster>.com
```

modules/web-console-configuration.adoc
```diff
@@ -0,0 +1,41 @@
...
+----
+<1> Specify the state of the console, `Managed` or `Disabled`. If the value is
+`Managed`, then the web console is enabled. Set it to `Disabled` to disable
+the console.
+<2> Specify the URL of the page to load when a user logs out of the web console.
+If you do not specify a value, the user returns to the login page for the
+web console. Specifying a `logoutRedirect` URL allows your users to perform
+single log out (SLO) through the identity provider to destroy their single
+sign-on session.
+<3> The web console URL. Do not modify this value.
```
benjaminapetersen: I don't know that you have to say `Do not modify this value` as it will get fixed if a user tampers.  Expressing that this is the canonical URL of the web console for the cluster is probably sufficient.

spadgett:
@bparees fyi, I know you were asking about this

modules/web-console-configuration.adoc
kalexand-rh: @benjaminapetersen, thank you very much for that clarification! I added a separate topic to disable to console. Will you PTAL?

modules/web-console-configuration.adoc
```diff
@@ -0,0 +1,36 @@
...
+    logoutRedirect: "" <1>
+status:
+  consoleURL: "" <2>
+----
+<1> Specify the URL of the page to load when a user logs out of the web console.
+If you do not specify a value, the user returns to the login page for the
+web console. Specifying a `logoutRedirect` URL allows your users to perform
+single log out (SLO) through the identity provider to destroy their single
+sign-on session.
+<2> The web console URL.
```
spadgett: I think it's still worth saying the URL can't be changed. The value won't be honored if the user tries to set this, but I'd still mention it to avoid any surprises.

spadgett (commented):
Looks good, just one comment. Thanks!

modules/web-console-configuration.adoc
kalexand-rh: Fair enough. I'll tell them not to change it. Thanks!

kalexand-rh:
@vikram-redhat, PTAL. I think that the console devs are happy, but I'm not sure who to tag for QE.

modules/web-console-configuration.adoc
```diff
@@ -0,0 +1,37 @@
...
+spec:
+  authentication:
+    logoutRedirect: "" <1>
+status:
+  consoleURL: "" <2>
+----
+<1> Specify the URL of the page to load when a user logs out of the web console.
+If you do not specify a value, the user returns to the login page for the
+web console. Specifying a `logoutRedirect` URL allows your users to perform
+single log out (SLO) through the identity provider to destroy their single
```
benjaminapetersen: nit:  I believe it is typically "single logout" vs "single log out".

modules/web-console-configuration.adoc
benjaminapetersen: Yup! looks good.

spadgett (approved):
lgtm

modules/web-console-configuration.adoc
kalexand-rh: Fixed. Thanks!

vikram-redhat:
@xltian we need an exception for this one for 4.1 as well. I think @yapei might be the correct associate to confirm.

yapei:
@vikram-redhat  Do we need add customizing `brand` and `documentationBaseURL` ?
```
$ oc edit console.operator.openshift.io cluster
spec:
  customization:
    brand: okd
    documentationBaseURL: https://docs.okd.io/latest/ 
```

bparees:
@spadgett thanks... can the loglevel or number of instances of the console be configured via that config object or another one?  how about loglevel of the operator itself?

Those things might belong elsewhere but are also docs we should have.

vikram-redhat:
@yapei thanks. We probably do, but they can come at a later date. If the text looks correct for what is there, are you able to approve it?

yapei:
> @yapei thanks. We probably do, but they can come at a later date. If the text looks correct for what is there, are you able to approve it?

Ok, also the current doc looks good from QE's point of view

vikram-redhat:
> Ok, also the current doc looks good from QE's point of view

Thanks! Merging this. If there are any other changes required, we can revisit in a new PR.

spadgett:
> @spadgett thanks... can the loglevel or number of instances of the console be configured via that config object or another one? how about loglevel of the operator itself?
> 
> Those things might belong elsewhere but are also docs we should have.

Replicas and log level are not configurable today. (@benjaminapetersen keep me honest.) We're looking at improving logging in the operator in 4.2 [1]. It would be good to wire through logging options for console itself in 4.2, although there's not much that's logged. The console backend for the most part is really simple.

[1] https://github.com/openshift/console-operator/pull/227

benjaminapetersen:
looks good to me!

PR #15119, opened 2019-06-03
bug BZ1716427 note about patch failing

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1716427


staebler:
Do we care to add any context for why the patch may fail? I don't know what the general approach is to giving information such as that in installation instructions.

kalexand-rh:
@jianlinliu, will you PTAL?

jianlinliu:
LGTM.

kalexand-rh:
Thanks! Merging.

PR #15125, opened 2019-06-03
Only OpenShift SDN currently supports modifying kube-proxy

jboxman:
- See https://github.com/openshift/api/blob/release-4.1/operator/v1/types_network.go#L70


jboxman:
@vikram-redhat, this moves kube-proxy to OpenShift SDN sub-topic, as no other CNI plug-ins currently support configuring it.

PR #15168, opened 2019-06-04
fixing RHCOS URL and clarifying image versions

kalexand-rh:
From a Slack discussion with @tbielawa, @wking, and @sosiouxme.

### DIFF
--- a/modules/installation-user-infra-machines-iso.adoc
+++ b/modules/installation-user-infra-machines-iso.adoc
@@ -17,16 +17,8 @@ the machines that you create can access.
 
 .Procedure
 
-////
-This will work for GA.
-. Obtain the {op-system} images from the
-link:https://access.redhat.com/downloads[Product Downloads] page on the Red
-Hat customer portal.
-////
-
 . Upload the control plane, compute, and bootstrap Ignition config files that the
 installation program created to your HTTP server. Note the URLs of these files.
-
 
 . Obtain the {op-system} images that are required for your preferred method
 of installing operating system instances from the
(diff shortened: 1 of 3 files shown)


modules/installation-user-infra-machines-iso.adoc
```diff
@@ -17,33 +17,30 @@ the machines that you create can access.
...
 link:https://access.redhat.com/downloads[Product Downloads] page on the Red
 Hat customer portal or the
 link:https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.1/[{op-system} image mirror]
 page.
 +
+[IMPORTANT]
+====
+The {op-system} images might not change with every release of {product-title}.
+You must download images with the same version number, and that version must be
+less than or equal to the {product-title} version that you install. Use the image versions
```
wking: You should use the largest version that is <= your target release, not any version that is <= your target release.

modules/installation-user-infra-machines-iso.adoc
```diff
@@ -17,33 +17,30 @@ the machines that you create can access.
...
 of installing operating system instances from the
 link:https://access.redhat.com/downloads[Product Downloads] page on the Red
 Hat customer portal or the
 link:https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.1/[{op-system} image mirror]
 page.
 +
+[IMPORTANT]
+====
+The {op-system} images might not change with every release of {product-title}.
+You must download images with the same version number, and that version must be
```
sosiouxme: I think I know what you mean by "You must download images with the same version number" (for the things you're downloading, all should have the same number) however I think that will be confusing in juxtaposition with the next bit. Also they're looking in a versioned directory, and they're only going to download 1 or 2 things so I don't think they'll get confused about the versions of those things matching each other. Best not to say anything?
```suggestion
You must download images with the highest version that is
```

kalexand-rh: @sosiouxme, @wking, thanks! Is this acceptable?

modules/installation-user-infra-machines-iso.adoc
```diff
@@ -17,33 +17,30 @@ the machines that you create can access.
...
 
-** ISO: `rhcos-<version>-<date>-installer.iso`
-** Compressed metal BIOS: `rhcos-<version>-<date>-metal-bios.raw.gz`
-** Compressed metal UEFI: `rhcos-<version>-<date>-metal-uefi.raw.gz`
-
-. Obtain the {op-system} images that are required for your preferred method
-of installing operating system instances.
+** ISO: `rhcos-<version>-installer.iso`
+** Compressed metal BIOS: `rhcos-<version>-metal-bios.raw.gz`
+** Compressed metal UEFI: `rhcos-<version>-metal-uefi.raw.gz`
```
sosiouxme: It's technically `rhcos-<version>-<architecture>-metal-uefi.raw.gz` (e.g. `rhcos-4.1.0-x86_64-metal-uefi.raw.gz`) and similar. This should be obvious to anyone who goes and looks at what's there, so I'm not really concerned about it if you want to roll in version+arch into `<version>`, just wanted to make sure you noticed that.

kalexand-rh: I fixed it. If we eventually add Power or z support, it'll be an important designation.

sosiouxme:
Thanks
/lgtm

kalexand-rh:
@jianlinliu, the image names changed yesterday. Do you have any objections?

jianlinliu:
lgtm

kalexand-rh:
Thank you! Merging.

PR #13585, opened 2019-02-08
Fix typos and copy and paste errors

nekop:

PR #15173, opened 2019-06-04
Fix typos and copy and paste errors

kalexand-rh:
From https://github.com/openshift/openshift-docs/pull/13585

@mrogers950, @stuartchuan, @barleyer, will you PTAL and confirm which 3.x versions this change should apply to?


vikram-redhat:
@mrogers950 can you please take a look?

@stuartchuan can you confirm which versions it should apply to?

barleyer:
@lihongan could you take a look?

lihongan:
LGTM, thank you. 
I think the change should be applied to all 3.x versions, but seems the chapter has less content in old version so not all changes can be applied to 3.6 and before.

kalexand-rh:
Thank you, @lihongan! I'll make sure that the change is applied back to 3.6 and try to put it on earlier branches.

kalexand-rh:
I checked, and this change isn't valid before 3.6.

PR #15174, opened 2019-06-04
BZ1676399 - Upgrade documentation if oreg_url changes

kalexand-rh:
Added an additional playbook that needs to ran if `oreg_url`
is change for the upgrade.

https://bugzilla.redhat.com/show_bug.cgi?id=1676399

From https://github.com/openshift/openshift-docs/pull/14253

@jiajliu, will you please confirm?

### DIFF
--- a/upgrading/automated_upgrades.adoc
+++ b/upgrading/automated_upgrades.adoc
@@ -600,6 +600,15 @@ Otherwise, the service is restarted on HA masters, but the systems do not reboot
 xref:../install/configuring_inventory_file.adoc#configuring-cluster-variables[Configuring
 Cluster Variables] for details.
 
+. If you modified the `oreg_url` parameter to change the cluster image registry location,
+you must run the `imageconfig` playbook to update the image location:
++
+----
+$ cd /usr/share/ansible/openshift-ansible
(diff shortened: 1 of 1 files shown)


kalexand-rh:
@jiajliu, will you PTAL?

kalexand-rh:
@jiajliu, will you PTAL?

kalexand-rh:
@xltian, will you please help me find the right reviewer?

jiajliu:
@kalexand-rh 
Sorry to miss your information in the pr. It's would be better to track it in the bug so we can track them more timely:)
The description LGTM except the sequence to run the playbook. According to my test result(https://bugzilla.redhat.com/show_bug.cgi?id=1676399#c14), users need to run the imageconfig.yml playbook before run upgrade playbook if they change oreg_url before upgrade, or else, they will hit the issue in the bug when run upgrade playbook directly.

kalexand-rh:
Hi @jiajliu! I'll keep that in mind next time - the bug was closed and was not mine, so I thought the PR would be better. I moved running the imageconfig.yml playbook to before you run either upgrade playbook, so this one should be good now. Thank you!

jiajliu:
LGTM now

PR #15213, opened 2019-06-05
moving install concept

kalexand-rh:
Jake Lucky, the PM for the OpenShift Cluster Manager, pinged me on Slack and said that the install concept at the start of each installation assembly made the topics seem intimidatingly long, even if they weren't.

I've removed the concept from each assembly and added a link to the arch guide topic instead.

@openshift/team-documentation, what do you think?

preview build here: http://file.rdu.redhat.com/kalexand/060519/install_concept/installing/installing_vsphere/installing-vsphere.html#installation-overview-installing-vsphere

### DIFF
--- a/installing/installing_aws/installing-aws-customizations.adoc
+++ b/installing/installing_aws/installing-aws-customizations.adoc
@@ -12,12 +12,13 @@ some parameters in the `install-config.yaml` file before you install the cluster
 
 .Prerequisites
 
+* Review details about the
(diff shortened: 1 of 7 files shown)


vikram-redhat:
I would check in with Ali as well before merging.

adellape:
I enjoy this.

kalexand-rh:
I asked Ali about this PR on Slack, and she said that it looked fine. We don't want users to refuse to scroll through the installation topics.

PR #15214, opened 2019-06-05
osdocs-441 adding registry storage info to install

kalexand-rh:
https://jira.coreos.com/browse/OSDOCS-441

Built draft here: http://file.rdu.redhat.com/kalexand/060519/osdocs441/installing/installing_vsphere/installing-vsphere.html#installation-overview-installing-vsphere

@huffmanca, @bmcelvee, will you PTAL?

### DIFF
--- a/modules/installation-registry-storage-non-production.adoc
+++ b/modules/installation-registry-storage-non-production.adoc
@@ -5,7 +5,7 @@
 // * installing/installing_vsphere/installing-vsphere.adoc
 
 [id="installation-registry-storage-non-production_{context}"]
-= Providing non-production storage to the image registry
+= Configuring storage for the image registry in non-production clusters
 
 You must configure storage for the image registry Operator. For non-production
 clusters, you can set the image registry to an empty directory.
--- a/modules/installation-registry-storage-non-production.adoc
+++ b/modules/installation-registry-storage-non-production.adoc
@@ -1,10 +1,11 @@
 // Module included in the following assemblies:
 //
+// * installing/installing_aws_user_infra/installing-aws-user-infra.adoc
 // * installing/installing_bare_metal/installing-bare-metal.adoc
 // * installing/installing_vsphere/installing-vsphere.adoc
 
 [id="installation-registry-storage-non-production_{context}"]
-= Provide storage to the image registry
+= Providing non-production storage to the image registry
 
 You must configure storage for the image registry Operator. For non-production
 clusters, you can set the image registry to an empty directory.
--- a/modules/registry-configuring-storage-aws-user-infra.adoc
+++ b/modules/registry-configuring-storage-aws-user-infra.adoc
@@ -1,6 +1,7 @@
 // Module included in the following assemblies:
 //
 //* architecture/installation-.adoc
+// * installing/installing_aws_user_infra/installing-aws-user-infra.adoc
 
 [id="registry-configuring-storage-aws-user-infra_{context}"]
 = Configuring registry storage for AWS with user-provisioned infrastructure
@@ -21,7 +22,7 @@ procedure.
 Use the following procedure if the Registry Operator cannot create an S3 bucket
 and automatically configure storage.
 
-. Create a S3 bucket:
+. Create an S3 bucket:
 +
 ----
 $ oc create configs.imageregistry.operator.openshift.io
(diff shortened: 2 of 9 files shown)


bmcelvee:
LGTM. I'll hold off on switching the label until Christian has a chance to take a look.

modules/installation-registry-storage-non-production.adoc
```diff
@@ -1,10 +1,11 @@
 // Module included in the following assemblies:
 //
+// * installing/installing_aws_user_infra/installing-aws-user-infra.adoc
 // * installing/installing_bare_metal/installing-bare-metal.adoc
 // * installing/installing_vsphere/installing-vsphere.adoc
 
 [id="installation-registry-storage-non-production_{context}"]
-= Provide storage to the image registry
+= Providing non-production storage to the image registry
```
huffmanca: This title and section seem to be about configuring an image registry to emptydir for non-production clusters. As such, the wording of the title is a little confusing to me at present. Would it make sense to rename this to be something similar to:

> Configuring storage for the image registry in non-production clusters

I think the `non-production storage` bit is what's bothering me.

huffmanca: Or possibly:
> Configuring an empty directory for non-production clusters

huffmanca (commented):
One nitpick. I think it's fine if you want to merge it as is, though. The rest looks good to me!

modules/installation-registry-storage-non-production.adoc
kalexand-rh: I like the way the first one sounds. I'll update the file. Thank you!

kalexand-rh:
@jianlinliu, will you PTAL? I moved some of the relevant options for configuring registry storage into the installation assemblies. I didn't modify the methods that were already approved. Do you agree with this change?

modules/registry-configuring-storage-baremetal.adoc
```diff
@@ -24,7 +25,7 @@ the `configs.imageregistry/cluster` resource.
 . Verify you do not have a registry pod:
 +
 ----
-$ oc get pods -n configs.imageregistry.operator.openshift.io/cluster 
+$ oc get pods -n configs.imageregistry.operator.openshift.io/cluster
```
jianlinliu: Obviously the above command is using wrong syntax. 
Here you want to verify if a image registry config - 'cluster' is existing for the following modification, but not pod. am I right?
If you want to verify registry existing or not, should be:
`$ oc get po -n openshift-image-registry`

modules/registry-configuring-storage-vsphere.adoc
```diff
@@ -1,6 +1,7 @@
 // Module included in the following assemblies:
 //
 //* architecture/installation-.adoc
+// * installing/installing_vsphere/installing-vsphere.adoc
 
 [id="registry-configuring-storage-vsphere_{context}"]
 = Configuring registry storage for vSphere
```
jianlinliu: The same issue like baremetal image registry configuration.
Obviously the above command is using wrong syntax. 
`
$ oc get pod -n configs.imageregistry.operator.openshift.io/cluster
invalid namespace "configs.imageregistry.operator.openshift.io/cluster": [may not contain '/']
`
Here you want to verify if a image registry config - 'cluster' is existing for the following modification, but not pod. am I right?
If you want to verify registry existing or not, should be:
`$ oc get po -n openshift-image-registry`

modules/registry-configuring-storage-aws-user-infra.adoc
```diff
@@ -21,7 +22,7 @@ procedure.
 Use the following procedure if the Registry Operator cannot create an S3 bucket
 and automatically configure storage.
 
-. Create a S3 bucket:
+. Create an S3 bucket:
 +
 ----
 $ oc create configs.imageregistry.operator.openshift.io
```
jianlinliu: I do not think this command is used for creating an s3. s3 should be created manually by user in this case from aws cli or web console, isn't it?
And even this command can not be executed successfully.
`$ oc create configs.imageregistry.operator.openshift.io
Error: required flag(s) "filename" not set`

kalexand-rh: Brandi says that it's fine to remove this command.

kalexand-rh:
@jianlinliu, thanks! I updated the commands per your comments. Brandi says that the original commands were approved by QE earlier in the release, but if yours are right now, to use them.

jianlinliu:
LGTM.

kalexand-rh:
Thanks! Merging.

PR #15218, opened 2019-06-05
Enable caching for content per Google PageSpeed Insights recommendation

jboxman:
So I've been thinking about this for a few months now. Currently, we don't provide an `Expires` header. An optimal strategy will require additional research and discussion, but I wanted to open this now anyway.

```
$ curl -I https://docs.openshift.com/

HTTP/1.1 200 OK
Date: Wed, 05 Jun 2019 18:17:01 GMT
Server: Apache/2.4.34 (Red Hat) OpenSSL/1.0.2k-fips
Last-Modified: Wed, 05 Jun 2019 14:41:49 GMT
ETag: "3c62-58a9498ecc540"
Accept-Ranges: bytes
Content-Length: 15458
Vary: Accept-Encoding
X-UA-Compatible: IE=Edge,chrome=1
Content-Type: text/html; charset=utf-8
Set-Cookie: 0831e0291a1cfb29457385c56afd1928=897d79a9d7e69f42f5e4e045077169e5; path=/; HttpOnly; Secure
Cache-control: private
```

### DIFF
--- a/.s2i/httpd-cfg/01-commercial.conf
+++ b/.s2i/httpd-cfg/01-commercial.conf
@@ -248,6 +248,36 @@ AddType text/vtt                            vtt
 </IfModule>
 
 # ----------------------------------------------------------------------
+# Specify cache retension policy
+# ----------------------------------------------------------------------
+
+<IfModule mod_expires.c>
+  ExpiresActive On
+  ExpiresDefault "access plus 1 seconds"
+  ExpiresByType image/x-icon "access plus 2592000 seconds"
+  ExpiresByType image/jpeg "access plus 2592000 seconds"
+  ExpiresByType image/png "access plus 2592000 seconds"
+  ExpiresByType image/gif "access plus 2592000 seconds"
+  ExpiresByType text/css "access plus 604800 seconds"
+  ExpiresByType text/javascript "access plus 216000 seconds"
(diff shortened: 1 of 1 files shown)


vikram-redhat:
This is great. Yes, we should do this. The only thing I would remove is text/html for which we should not have any caching (due to fires which require immediate updates.)

vikram-redhat:
Also poor vikram-raj is probably not sure why he got tagged :)

jboxman:
@vikram-redhat I blame auto-complete :)

jboxman:
@vikram-redhat if Apache is going away, I need to reformulate this.

vikram-redhat:
@jboxman yes this will need to change. I can't recall why we didn't merge it already though?

jboxman:
@vikram-redhat, you wanted it to leave the behavior for caching HTML pages alone so users receive updates as soon as possible.

vikram-redhat:
@jboxman correct, so it just means we take the text/html part out and then merge the rest.

jboxman:
@vikram-redhat, true.

jboxman:
@vikram-redhat, did we ever move to Nginx?

vikram-redhat:
@jboxman no we didn't :) Merge away.

PR #15225, opened 2019-06-05
Global replace: "comprised of" -> "composed of"

maxwelldb:
As discussed--any thoughts, @kalexand-rh ?


kalexand-rh (commented):
LGTM. Thanks @maxwelldb

PR #15246, opened 2019-06-06
BZ-1684357: Removing section from 4.1 since it's not configurable

bergerhoffer:
Preview: http://file.rdu.redhat.com/~ahoffer/2019/BZ-1684357-1a/applications/quotas/quotas-setting-per-project.html


bergerhoffer:
Hi @xingxingxia.  @deads2k said that this is not configurable in 4.1, so we should remove this section from the 4.1 docs. I've done so in this PR, can you please take a look?

xingxingxia:
PR LGTM, thanks

PR #15251, opened 2019-06-06
BZ-1717447: Indicated that no default strategy is provided.

huffmanca:
Updated the OAuth grant information to indicate that no default strategy is provided. Clients must specify their grant strategy.

This is for OS 4.x.


barleyer:
From QE:
/lgtm

PR #15268, opened 2019-06-07
Add required ports for Prometheus to required ports section

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1693194

### DIFF
--- a/install/prerequisites.adoc
+++ b/install/prerequisites.adoc
@@ -592,6 +592,8 @@ post back status, to receive tasks, and so on.
 | *10250*
 |TCP
 |The master proxies to node hosts via the Kubelet for `oc` commands.
+This port must to be allowed from masters and infra nodes to any master and node.
+For metrics, the source must be the infra nodes.
 
 | *10010*
 |TCP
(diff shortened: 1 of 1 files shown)


mburke5678:
@sreber84 PTAL

mburke5678:
If we need to apply changes to 3.7 and prior, I will create a follow-up PR, as the install docs structure changed in 3.9+

install/prerequisites.adoc
```diff
@@ -729,16 +729,45 @@ It depends on the target host of the deployment and uses the computed value of
...
 infrastructure node so the members of the Elasticsearch cluster might communicate
 with each other.
+
+| *9090*
+| TCP 
+| For Prometheus web console use.
+
+| *9100*
+| TCP
+| For the Prometheus Node-Exporter, which exports hardware and operating system metrics.
```
palonsoro: Port `9100` should be allowed from infra nodes to any master and node. It may be worth to tell this explicitly

install/prerequisites.adoc
```diff
@@ -729,16 +729,45 @@ It depends on the target host of the deployment and uses the computed value of
...
+
+| *8443*
+| TCP
+| For node hosts to communicate to the master API, for the node hosts to
+post back status, to receive tasks, and so on.
+
+| *10250*
```
palonsoro: This port needs to be allowed from masters (already explained above) and infra nodes to any master and node. Although it is said above that it is explicitly needed from masters, it would be good to also explain that for the metrics to work, the needed source is the infra nodes.

install/prerequisites.adoc
```diff
@@ -729,16 +729,45 @@ It depends on the target host of the deployment and uses the computed value of
...
+| TCP 
+| For the Kubernetes cAdvisor, a container resource usage and performance analysis agent.
+
+| *8444*
+| TCP 
+| |Port that the controller service listens on.
+
+| *1936*
```
palonsoro: Again, this port must be allowed from infra nodes to infra nodes (or,  concretely, to those nodes hosting the routers) if prometheus metrics are enabled on routers.

mburke5678:
@palonsoro Please take a look. I am not as familiar with firewall rules as I probably should be. Any advice on how to make this table more useful is greatly appreciated.

mburke5678:
@palonsoro Please take a look. I am not as familiar with firewall rules as I probably should be. Any advice on how to make this table more useful is greatly appreciated.

mburke5678:
@palonsoro Please take a look. I am not as familiar with firewall rules as I probably should be. Any advice on how to make this table more useful is greatly appreciated.

mburke5678:
@palonsoro Please take a look. I am not as familiar with firewall rules as I probably should be. Any advice on how to make this table more useful is greatly appreciated.

mburke5678:
@palonsoro Please take a look. I am not as familiar with firewall rules as I probably should be. Any advice on how to make this table more useful is greatly appreciated.

palonsoro:
I see the changes that now specify the concrete sources and destinations. I find them good.

What I still think that needs a change is to have them separated from logging ports. This is not logging, it is metrics.

palonsoro:
Please disregard my latest observation. I hadn't seen the title change. With the title change, it looks good to me.

mburke5678:
@juzhao Please take a look

install/prerequisites.adoc
```diff
@@ -724,23 +726,61 @@ It depends on the target host of the deployment and uses the computed value of
...
 | *9300*
 |TCP
-|For Elasticsearch inter-cluster use. Required to be internally open on any
+| For Elasticsearch inter-cluster use. Required to be internally open on any
 infrastructure node so the members of the Elasticsearch cluster might communicate
 with each other.
+
+| *9090*
+| TCP 
+| For Prometheus web console use.
```
juzhao: 9090 is also used to reload the prometheus configuration, so should be
For Prometheus API and web console use.

mburke5678: @juzhao I made the change as suggested.

PR #15271, opened 2019-06-08
Included initial draft of a script to search for unused modules.

huffmanca:
This is the first attempt at creating a script to display any modules that aren't currently being included in assemblies. It functions as follows

* Building a list of the files under "modules/"
* Building a list of assemblies, consisting of all other files ending in ".adoc".
* Searching through each assembly for any lines with "include::", and removing these entries from the list of modules.

You can run it by passing the path to the "openshift-docs" repo on your local filesystem. I've included both the built binary, along with the source. For instance, to run it from the `$REPO/scripts` directory:

$ ./find_unused ../

@kalexand-rh ,

Tagging you for review (if you get time), since you'd mentioned that this might be a useful addition.

### DIFF
--- a/_unused_topics/README
+++ b/_unused_topics/README
@@ -1,0 +1,2 @@
+Placeholder file. Any modules that are not included will be placed here
+by the `scripts/find_unused.py` script.
--- a/scripts/find_unused.py
+++ b/scripts/find_unused.py
@@ -1,0 +1,98 @@
+#!/usr/bin/env python
+
+"""
+This script goes through and searches the OpenShift docs repository for any modules
+that are not included in assemblies.
+
+__author__      = "Christian Huffman"
+__version__     = "1 July, 2019"
+"""
+
+from os.path import join
+import argparse
+import os
+import shutil
+import sys
+
+class FindUnused():
+    unused_directory = "_unused_topics"     # Edit this to adjust the target path.
+    modules_directory = "/modules"          # Edit this to adjust the modules subdirectory
+    
+    def __init__(self, args):
+        self.args = args
+    
+    def process(self):
+        """
+        This function is responsible for obtaining the list of modules, the list of assemblies,
+        and then either moving or displaying any modules not being included.
+        """
+        modules = {}
+        assemblies = []
+        # Build up a dictionary from the elements in the modules subdirectory, storing their path as the value.
+        for root, directories, files in os.walk(self.args.path + self.modules_directory):
+            for filename in files:
+                modules[filename] = os.path.join(root,filename)
+        # Since modules can also include other modules, we include them in the list of assemblies.
+        for root, directories, files in os.walk(self.args.path):
+            for filename in files:
+                if filename.endswith(".adoc"):
+                    assemblies.append(os.path.join(root,filename))
+        remaining_modules = self.check_assemblies(assemblies,modules)
+        # Determine if we should move the files or simply print the list
+        if self.args.move:
+            self.move_files(self.args.path,remaining_modules)
+        else:
+            for filename in remaining_modules:
+                print remaining_modules[filename]
+    
+    def check_assemblies(self,assemblies,modules):
+        """
+        Because we can't modify dictionaries while they're being iterated over, we're
+        forced to create a new list to collect any files that are found. Once this process
+        completes, we then go through the original list and remove the found files.
+        
+        This process leaves us with a dictionary containing the unused files and their paths,
+        which is returned.
+        """
+        included_modules = {}
+        for filename in assemblies:
+            currentfile = open(filename,"r")
+            for line in currentfile.readlines():
+                for key in modules:
+                    if key in line:
+                        included_modules[key] = ""
+            currentfile.close()
+        for key in included_modules:
+            del modules[key]
+        return modules
+    
+    def move_files(self,root,remaining_modules):
+        """
+        Goes through each path in the dictionary and moves it to "root/self.unused_directory/"
+        Once finished it prints out the total number of files moved.
+        """
+        try:
+            for key in remaining_modules:
+                os.rename(remaining_modules[key],root + "/" + self.unused_directory + "/" + key)
+            print "Moved %d files" % len(remaining_modules)
+        except OSError as err:
+            print("Unable to move files: {0}".format(err))
+            print("Target destination: " + self.unused_directory)
+    
+    
+def main():
+  usage = """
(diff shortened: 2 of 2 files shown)


kalexand-rh (commented):
Thanks, @huffmanca! This LGTM.

kalexand-rh:
@vikram-redhat, PTAL?

vikram-redhat:
Thanks @huffmanca - this makes sense. The logic and code look flawless. I have two points to make.

1. I caution against removing the modules that are not in any assembly. Both the CNV and Service Mesh initiatives are migrating their content to the 4.1 repo, and are building modules first. You will get a lot of false positives. It might be better to print a list of these modules instead of deleting them so that the person running the script can make decisions based on that.

2. I am worried that this was written in go as opposed to a shell script or python, the language of choice for our scripting adventures :)

I have nothing against go (languages are just a means to an end), but I am concerned that any writer on the team will need to invest in installing the compiler/executable for go. This is not an issue per se, but to maintain consistency, we should probably make this a python script.

I apologize, I had no idea you were doing this and it might seem like a wasted effort. Hopefully, the logic can be easily transferred.

kalexand-rh:
1. It doesn't delete them. It just outputs the list unused modules.

```
[kalexand@localhost scripts]$ ./find_unused ../
modules/sb-uninstall-asb.adoc
modules/efk-logging-exported-fields-collectd.adoc
modules/monitoring-dead-mans-switch-pagerduty.adoc
modules/nodes-containers-using-ssh.adoc
modules/monitoring-enabling-dynamically-provisioned-storage.adoc
modules/monitoring-update-and-compatibility-guarantees.adoc
modules/security-overview.adoc
modules/efk-logging-kibana-console-launch.adoc
modules/sb-about-template-service-broker.adoc
modules/nodes-nodes-resources-configuring-viewing.adoc
modules/nodes-pods-daemonsets-pods.adoc
modules/efk-logging-deploying-storage-considerations.adoc
modules/builds-output-image-digest.adoc
modules/architecture-new-content.adoc
modules/nodes-scheduler-taints-tolerations-examples.adoc
modules/efk-logging-elasticsearch-admin.adoc
modules/monitoring-setting-persistent-storage-size.adoc
modules/service-accounts-managing-secrets.adoc
modules/adding-new-devices.adoc
modules/configuration-resource-overview.adoc
modules/upgrade-cluster-version-definition.adoc
modules/monitoring-configuring-etcd-monitoring.adoc
modules/sb-about-ansible-service-broker.adoc
modules/nodes-containers-using-about.adoc
modules/sb-install-service-catalog.adoc
modules/identity-provider-configuring-apache-request-header.adoc
modules/efk-logging-exported-fields-docker.adoc
modules/nodes-pods-disabling-features-about.adoc
modules/efk-logging-curator-yaml.adoc
modules/cnv-accessing-rdp-console.adoc
modules/monitoring-full-list-of-configuration-variables.adoc
modules/efk-logging-elasticsearch-persistent-storage-about.adoc
modules/efk-logging-fluentd-collector.adoc
modules/completing-installation.adoc
modules/running-modified-installation.adoc
modules/deploying-local-provisioner.adoc
modules/nodes-cluster-overcommit-node-memory.adoc
modules/monitoring-dead-mans-switch.adoc
modules/cnv-vm-rdp-console-web.adoc
modules/nodes-nodes-audit-log-advanced.adoc
modules/mounting-local-volumes.adoc
modules/identity-provider-provisioning-user-lookup-mapping.adoc
modules/monitoring-enabling-persistent-storage.adoc
modules/machine-configs-and-pools.adoc
modules/modules
modules/efk-logging-uninstall-efk-ops.adoc
modules/monitoring-monitoring-overview.adoc
modules/sb-uninstall-tsb.adoc
modules/installation-about-custom.adoc
modules/configuration-resource-configure.adoc
modules/sb-about-service-catalog.adoc
modules/nodes-cluster-disabling-features-list.adoc
modules/security-context-constraints-restore-defaults.adoc
modules/images-create-metadata-define.adoc
modules/nodes-nodes-working-adding.adoc
modules/configuring-local-provisioner.adoc
modules/efk-logging-fluentd-log-rotation.adoc
modules/understanding-workers-masters.adoc
modules/installation-creating-worker-machineset.adoc
modules/nodes-scheduler-node-antiaffinity-configuring.adoc
modules/rbac-updating-policy-definitions.adoc
modules/con-pod-reset-policy.adoc
modules/pod-using-a-different-service-account.adoc
modules/monitoring-grouping-alerts.adoc
modules/service-accounts-adding-secrets.adoc
modules/cnv-ansible-modules.adoc
modules/identity-provider-create-CR.adoc
modules/nodes-scheduler-taints-tolerations-seconds.adoc
modules/nodes-pods-priority-examples.adoc
modules/exploring-cvo.adoc
modules/understanding-installation.adoc
modules/looking-inside-nodes.adoc
[kalexand@localhost scripts]$ git status
On branch PR15271
nothing to commit, working tree clean
```
2. OpenShift 4 is written in go, and I think that F30 installs the required go dependencies by default. (I do not recall specifically installing go packages after I had to reinstall.)

huffmanca:
Python would be my first choice for something like this, and rewriting it in Python shouldn't take very long; I chose to do it in Go primarily as a learning exercise, as OpenShift is written in Go.

If you'd prefer it in Python let me know and I can replace it.

As Kathryn indicated, it doesn't actually delete any files - just lists them. I don't think a script should ever delete the files automatically. At most, which I personally wouldn't recommend, it should move them into a separate directory that can be manually curated/cleaned at a later date.

jboxman:
This is actually ideal because you can pipe it to `xargs` and then handle deletion that way using `git rm` as the pipe target. Because `git`, it's always possible to revert the change.

huffmanca:
@vikram-redhat ,

I went ahead and updated this to use python instead of Go. In addition, I've added the `-m` flag which will move the files automatically, instead of requiring a user to pipe them in. Without any flags it prints the list out.

The code should be documented and fairly straightforward. It does take slightly longer than the Go version, but it's still less than 5 seconds on my machine.

I also included the `_unused_topics/` directory, along with a README, because the directory must exist when moving the files.

Let me know your thoughts, or what further changes should be made.

vikram-redhat:
This works flawlessly, both the listing and moving. Great work. When you are ready, ok to merge. Are you also able to:

1. Send out an email to the team about this tool.
2. Add a section in the manual about the usage of this tool.

Periodically, we should run this tool to move modules into the unused modules folder. I am hoping that the email will start the conversation around that.

PR #15287, opened 2019-06-10
Fixes custom build Docker Hub link

wgordon17:


wgordon17:
This should probably be cherry-picked back to other versions that reference this same link

kalexand-rh:
@vikram-redhat, the new link's right, and it's valid all the way to 3.0.

PR #15289, opened 2019-06-10
Update cluster-entitlements.adoc

crmarques:


kalexand-rh:
@crmarques, thanks for pointing this out! You need to open changes against master first, though. I'm fixing this issue on #15293.

PR #15311, opened 2019-06-10
Clarify correct usage of automatic egress IP assignments

jboxman:
- https://bugzilla.redhat.com/show_bug.cgi?id=1709254


jboxman:
QE verified in the BZ.

PR #15312, opened 2019-06-10
osdocs404 multiple vsphere hosts:

kalexand-rh:
https://jira.coreos.com/browse/OSDOCS-404

@staebler, will you PTAL?

built draft here: http://file.rdu.redhat.com/kalexand/061019/osdocs404/installing/install_config/vsphere-hosts.html


modules/vsphere-multiple-hosts.adoc
```diff
@@ -0,0 +1,53 @@
...
+kind: ConfigMap
+metadata:
+  creationTimestamp: null
+  name: cloud-provider-config
+  namespace: openshift-config
+----
+<1> Add more `[VirtualCenter]` sections to define the additional hosts for your
+cluster. See the
+link:https://vmware.github.io/vsphere-storage-for-kubernetes/documentation/existing.html[vSphere Cloud Provider Configuration]
+page in the VMware documentation.
```
kalexand-rh: What next? How long does it take for the changes to take effect?

modules/vsphere-multiple-hosts.adoc
```diff
@@ -0,0 +1,53 @@
...
+link:https://vmware.github.io/vsphere-storage-for-kubernetes/documentation/existing.html[vSphere Cloud Provider Configuration]
+and can provide the data that is required for the `VirtualCenter` hosts that you
+add.
+
+.Procedure
+
+. Edit the `cloud-provider-config` ConfigMap:
++
+----
+$ oc edit configmap/cloud-provider-config
```
staebler: This needs a `-n openshift-config` to get the configmap from the openshift-config namespace.

modules/vsphere-multiple-hosts.adoc
staebler: This will take on the order of minutes to fully take effect. The user should probably watched `oc get co` to wait for all of the operators to become available and not progressing. Specifically, I know that `kube-apiserver` and `kube-controller-manager` will roll out new versions of the pods that they manage.

modules/vsphere-multiple-hosts.adoc
```diff
@@ -0,0 +1,89 @@
...
+
+kind: ConfigMap
+metadata:
+  creationTimestamp: null
+  name: cloud-provider-config
+  namespace: openshift-config
+----
+<1> Add more `[VirtualCenter]` sections to define the additional hosts for your
+cluster. See the
+link:https://vmware.github.io/vsphere-storage-for-kubernetes/documentation/existing.html[vSphere Cloud Provider Configuration]
```
jianlinliu: Per the vmware doc, should we also update 'datacenters' parameter in 'Global' section?

jianlinliu: Per vmware doc example, should be something like:
datacenters = "dc1, dc2"
Because I have no such multiple host vsphere cluster to validate it, actually I am not 100% sure my above statement is correct or not, just a question from me. Maybe need dev @staebler confirm.

jiajliu:
@staebler @kalexand-rh 
Following up the steps seems not enough to add one more node on another vcenter into the an existed cluster. I have several try but still failed. So I tracked my steps in https://bugzilla.redhat.com/show_bug.cgi?id=1734276.

PR #15314, opened 2019-06-10
update RemovePending description based on upgrade from removed/unmana…

gabemontero:
…ged fix


bmcelvee:
Thanks! LGTM

Minor change, no QE needed.

PR #15344, opened 2019-06-11
Updates for BZ1688784

lamek:
Clarified the specific resources that can be encrypted.

https://bugzilla.redhat.com/show_bug.cgi?id=1688784

### DIFF
--- a/admin_guide/encrypting_data.adoc
+++ b/admin_guide/encrypting_data.adoc
@@ -11,8 +11,7 @@ toc::[]
 == Overview
 
 This topic reviews how to enable and configure encryption of secret data at the
-datastore layer. While the examples use the `secrets` resource, any resource can be
-encrypted, such as `configmaps`.
+datastore layer. While the examples use the `secrets` resource, both `secrets` and `configmaps` resources can be encrypted.
 
 [IMPORTANT]
 ====


lamek:
@openshift/team-documentation ready for peer review.

lamek:
@soltysh can you help confirm that only secrets and configmaps can be encrypted. Thanks.
@vikram-redhat how many versions back should I port this?

kalexand-rh (commented):
LGTM

vikram-redhat:
@lamek - till 3.9.

soltysh:
> @soltysh can you help confirm that only secrets and configmaps can be encrypted. Thanks.

@enj you'll know better, I guess.

enj:
/hold

Any resource can be encrypted.

lamek:
Thanks @enj. The attached bug says custom resources created from a CRD cannot be encrypted. (https://bugzilla.redhat.com/show_bug.cgi?id=1686452)
Would the following be a fair distinction:
> While the examples use the `secrets` resource, any built-in resources can be encrypted. Custom resources created from CRDs can not be encrypted.

enj:
> Thanks @enj. The attached bug says custom resources created from a CRD cannot be encrypted. (https://bugzilla.redhat.com/show_bug.cgi?id=1686452)
> Would the following be a fair distinction:
> 
> > While the examples use the `secrets` resource, any built-in resources can be encrypted. Custom resources created from CRDs can not be encrypted.

Custom resources created from CRDs **can** be encrypted.  Reading through the code, I see no reason for CRDs to not support encryption.  @stts can confirm.

sttts:
@enj CRDs use the same etcd storage layer as native resources. I see no reason why it should not work for CRDs.

lamek:
@hexfusion can you PTAL? Per your comments in the related BZ (https://bugzilla.redhat.com/show_bug.cgi?id=1686452) CustomResources created from CRDs **can not** be encrypted.

hexfusion:
I was going by documentation, my bad for not digging further. defer to @sttts @enj.

sttts:
I verified: it does not work with CRDs.

Reason: the encryption is not wired into the kube-apiserver storage factory, which is not used in the apiextensions-apiserver.

enj:
> I verified: it does not work with CRDs.
> 
> Reason: the encryption is not wired into the kube-apiserver storage factory, which is not used in the apiextensions-apiserver.

Not sure exactly what you mean, but this sounds like a bug to me.

sttts:
@enj One "not" too much. The encryption is ~~not~~ wired into the kube-apiserver storage factory. apiextensions-apiserver uses the "normal" RESTOptions getter without all the crazyness in the kube-apiserver one (for historic reasons).

Yes, we can call that a bug.

enj:
Opened https://bugzilla.redhat.com/show_bug.cgi?id=1726347 to track.

PR #15352, opened 2019-06-11
Add sections on setting cluster and project node selectors

mburke5678:
https://jira.coreos.com/browse/POD-94
https://jira.coreos.com/browse/POD-119

### DIFF
--- a/modules/nodes-scheduler-node-selectors-cluster.adoc
+++ b/modules/nodes-scheduler-node-selectors-cluster.adoc
@@ -1,0 +1,91 @@
+// Module included in the following assemblies:
+//
+// * nodes/nodes-scheduler-node-selector.adoc
+
+[id="nodes-scheduler-node-selectors-cluster_{context}"]
+= Creating default cluster-wide node selectors  
+
+You can use default node selectors on pods together with labels on nodes to constrain all pods created in a cluster to specific nodes.
+
+With cluster node selectors, when you create a pod in that cluster, {product-title} adds the appropriate `<key>:<value>` and schedules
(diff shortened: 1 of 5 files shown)


mburke5678:
@ravisantoshgudimetla PTAL

modules/nodes-scheduler-node-selectors-cluster.adoc
```diff
@@ -0,0 +1,89 @@
...
+spec:
+  nodeSelector:
+    region: west
+----
+
+If the project where you are creating the pod has a project node selector, that selector takes preference over a cluster node selector.
+
+.Procedure
+
+To add a default node selector to a project:
```
ravisantoshgudimetla: This is across cluster not per project

mburke5678:
@sunilcio Is this an issue you can review?

sunilcio:
@mburke5678 let me check now.

mburke5678:
@sunilcio Any thoughts on this? I know we are all busy. But I would like to get this merged in. It seems like an important feature that we missed.

sunilcio:
@mburke5678 I followed the steps in doc and it works correctly. LGTM.
I would recommend to also add statement from https://bugzilla.redhat.com/show_bug.cgi?id=1699460#c12 to something like below in Step 1 (Creating default cluster-wide node selectors)
`After making the change to the defaultNodeSelector in the cluster Scheduler config resource, wait for the kube-apiserver pods to redeploy.  This can take several minutes. Until the kube-apiserver pods redeploy, the defaultNodeSelector will not have effect.`

mburke5678:

PR #15355, opened 2019-06-12
bug BZ1691470 adding uninstall notes

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1691470

@mtnbikenc @csheremeta, PTAL? @brenton, FYI

### DIFF
--- a/install/uninstalling_cluster.adoc
+++ b/install/uninstalling_cluster.adoc
@@ -31,6 +31,32 @@ installed by Ansible, including:
 The playbook deletes content for any hosts defined in the inventory file
 that you specify when running the playbook.
 
+[IMPORTANT]
+====
+Before you uninstall your cluster, review the following list of scenarios and
+make sure that uninstalling is the best option:
+
+* If your installation process failed and you want to continue the process,
+you can
+xref:../install/running_install.adoc#advanced-retrying-installation[retry the installation].
+The installation playbooks are designed so that if they fail to install your
+cluster, you can run them again without needing to uninstall the cluster.
+* If you want to restart a failed installation from the beginning, you can
+uninstall the {product-title} hosts in your cluster by running the
+*_uninstall.yml_* playbook, as described in the following section. This playbook
+only uninstalls the {product-title} assets for the most recent version that you
+installed.
+* If you must change the host names or certificate names, you must recreate
+your certificates before retrying installation by running the *_uninstall.yml_*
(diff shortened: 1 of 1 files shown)


csheremeta:
Hey Katherine,

Thank you so much for adding this text!

kalexand-rh:
Sure thing, Candace! Let me know if you want any changes.

kalexand-rh:

install/uninstalling_cluster.adoc
```diff
@@ -31,6 +31,31 @@ installed by Ansible, including:
 The playbook deletes content for any hosts defined in the inventory file
 that you specify when running the playbook.
 
+[IMPORTANT]
+====
+Before you uninstall your cluster, make sure that uninstalling is the best
```
huffmanca: It might be implied, because this sentence immediately precedes a list, but perhaps it's better to say something like the following:

> Before you uninstall your cluster, examine the following items to ensure an uninstall is necessary.

install/uninstalling_cluster.adoc
```diff
@@ -31,6 +31,31 @@ installed by Ansible, including:
...
+option.
+
+* If your installation process failed and you want to continue the process,
+you can
+xref:../install/running_install.adoc#advanced-retrying-installation[retry the installation].
+The installation playbooks are designed so that if they fail to install your
+cluster, you can run them again without needing to uninstall the cluster.
+* If you want to restart a failed installation from the beginning, you can
```
huffmanca: I don't understand how this is different from the default uninstall procedure in the immediately following section? Do you run the uninstall without the `-i /path/to/file` so that no inventory file is used?

I suspect our customers, who are more familiar with the 3.x method, will be able to understand the differences between this and the default uninstall, but it might make sense to spell them out here.

huffmanca (commented):
A couple of minor tweaks, but overall this looks good!

install/uninstalling_cluster.adoc
csheremeta: I think this was meant to specify that it's ok to use the uninstall playbook IFF the customer experienced a failed installation and wants to start the installation over using the exact same setup, minor/asynchronous versions, etc. This is really the only time the uninstall playbook is ok to run (imvho)

kalexand-rh: Yes. I'm going to add "as described in the following section" to try to clarify this part. Thanks!

PR #15361, opened 2019-06-12
BZ-1719359: Commenting out Manila references.

huffmanca:
Removed the persistent storage using OpenStack Manila. Searching the rest of the storage sections I don't see any other references, so this should be it.

This is for OS 4.x.


liangxia:
LGTM (from QE)

PR #15370, opened 2019-06-12
Bug 1718163 Update Exposing Registry Content

bmcelvee:
https://bugzilla.redhat.com/show_bug.cgi?id=1718163


bmcelvee:
@adambkaplan - here's the WIP PR

modules/registry-exposing-secure-registry-manually.adoc
```diff
@@ -15,116 +15,44 @@ images using the route host.
...
-[NOTE]
-====
-Re-encrypt routes are also supported for exposing the secure registry.
-====
+You can expose the registry insecurely or securely:
 
-.. If it does not exist, create the route with the `oc create route passthrough`
-command, specifying the registry as the route’s service. By default, the name of
-the created route is the same as the service name:
+. Using `DefaultRoute`, which is insecure:
```
adambkaplan: From testing - the default route is in fact secured, but is self-signed. Therefore in most use cases users logging into the registry will get an x509 certificate error.

modules/registry-exposing-secure-registry-manually.adoc
```diff
@@ -15,116 +15,46 @@ images using the route host.
...
-[NOTE]
-====
-Re-encrypt routes are also supported for exposing the secure registry.
-====
+. Expose the router using `DefaultRoute`, which is insecure:
 
-.. If it does not exist, create the route with the `oc create route passthrough`
-command, specifying the registry as the route’s service. By default, the name of
-the created route is the same as the service name:
+.. Set `DefaultRoute` to `True`.
```
adambkaplan: Working command:
```
$ oc patch config.imageregistry.operator.openshift.io/cluster --patch '{"spec":{"defaultRoute":true}}' --type=merge
```

modules/registry-exposing-secure-registry-manually.adoc
```diff
@@ -15,116 +15,46 @@ images using the route host.
...
-... Get the *podman-registry* service details:
+.. Log in insecurely with Podman:
 +
 ----
-$ oc get svc
-NAME              CLUSTER_IP       EXTERNAL_IP   PORT(S)                 SELECTOR                  AGE
-podman-registry   172.30.69.167    <none>        5000/TCP                podman-registry=default   4h
-kubernetes        172.30.0.1       <none>        443/TCP,53/UDP,53/TCP   <none>                    4h
-router            172.30.172.132   <none>        80/TCP                  router=router             4h
+$ podman login --insecure default-route.image-registry...
```
adambkaplan: Script to log in:
```
HOST=$(oc get route default-route -n openshift-image-registry --template='{{ .spec.host }}')
podman login -u $(oc whoami) -p $(oc whoami -t) --tls-verify=false $HOST
```

modules/registry-exposing-secure-registry-manually.adoc
```diff
@@ -15,116 +15,46 @@ images using the route host.
...
-----
+You can expose the registry insecurely or securely:
+
+OR
 
-. Log in to the registry using the information from securing the registry.
-However, this time point to the host name used in the route rather than your
-service IP. When logging in to a secured and exposed registry, make sure you
-specify the registry in the `podman login` command:
+. Using Custom Routes, which is secure:
```
adambkaplan: We can drop "is secure" - adding your own route lets admins provide their own TLS config

modules/registry-exposing-secure-registry-manually.adoc
```diff
@@ -15,116 +15,46 @@ images using the route host.
...
 ----
-# podman login -e user@company.com \
-    -u f83j5h6 \
-    -p Ju1PeM47R0B92Lk3AZp-bWJSck2F7aGCiZ66aFGZrs2 \
-    <host>
+$ oc create secret tls public-route-tls
+  - n image-registry\
+  -- cert=/path/to/tls.crt\
+  -- key=/path/to/tls.key\
+//  -- from-file=tls.cacrt=/path/to/tls.cacrt\
```
adambkaplan: From testing - passing `--from-file` won't work.

@openshift/openshift-team-network-edge I noticed that OpenShift routes let you define a `caCertificate`, but Ingress objects do not (just the tls cert and key). What is the distinction?

registry/securing-exposing-registry.adoc
```diff
@@ -7,11 +7,8 @@ toc::[]
 ifdef::openshift-origin,openshift-enterprise,openshift-dedicated[]
 
 By default, the {product-title} registry is secured during cluster
-installation so that it serves traffic via TLS. A passthrough route is also
-created by default to expose the service externally.
+installation so that it serves traffic via TLS.
```
adambkaplan:
```suggestion
installation so that it serves traffic via TLS. 
Unlike previous versions of OpenShift, the registry is not exposed outside of the cluster at installation time.
```

modules/registry-exposing-secure-registry-manually.adoc
adambkaplan: Also worth noting - this part of the doc is commented out due to https://bugzilla.redhat.com/show_bug.cgi?id=1719965

Miciah: > @openshift/openshift-team-network-edge I noticed that OpenShift routes let you define a `caCertificate`, but Ingress objects do not (just the tls cert and key). What is the distinction?

The `Route` resource's `spec.tls.certificate` field specifies the server certificate whereas the `spec.tls.caCertificate` field specifies a CA certificate.  However, it is possible to create a PEM bundle that contains both the server certificate and a CA certificate and then specify this PEM bundle as the value for `spec.tls.certificate`.  That is, the `spec.tls.caCertificate` field is not really necessary to specify CA certificates.

Similar is true of the `Ingress` resource: you can create a `Secret` containing a PEM bundle that contains both the server certificate and a CA certificate and specify the name of this `Secret` as the value for an `Ingress` resource's `spec.tls[*].secretName` field.

(Technical details: OpenShift translates each `Ingress` resource and its associated `Secret` resources into one or more `Route` resources.  The contents of a `Secret` referenced by an `Ingress` resource's `spec.tls[*].secretName` field are put into a generated `Route` resource's `spec.tls.certificate` field.  Because a `Route` resource's `spec.tls.certificate` field can specify a PEM bundle, so too can an `Ingress` resource's `spec.tls[*].secretName` field.)

adambkaplan: @Miciah Thanks for the input. It seems in the docs the recommended procedure around TLS is to have the certificate be signed by the CA in a single PEM bundle [1]. I think we will want to continue that approach to make things consistent for admins.

I have two follow-up questions:
1. If a cluster admin sets up custom default certs on the ingress operator, will these default certs  apply to both `Ingress` objects and `Route` objects?
2. In the long run does it make sense for the registry operator to create `Ingress` objects, rather than `Route` objects?

/cc @dmage 

[1] https://docs.openshift.com/container-platform/4.1/networking/ingress-operator.html#nw-ingress-setting-a-custom-default-certificate_configuring-ingress

Miciah: > * If a cluster admin sets up custom default certs on the ingress operator, will these default certs  apply to both `Ingress` objects and `Route` objects?

Yes, because `Ingress` resources get translated into `Route` resources, and it the ingress controller (AKA router) handles all `Route` resources the same way irrespective of how they were created.

> * In the long run does it make sense for the registry operator to create `Ingress` objects, rather than `Route` objects?

We have no plans to deprecate the `Route` resource.  In general, OpenShift operators and users can use whichever resource is more convenient.  @ironcladlou, do you have anything to add to that?

(I did notice that the openshift-controller-manager operator, which starts the ingress-to-route controller, is at the same run level as the ingress operator, which means that ingress-to-route _might_ start after the ingress operator, in which case creating an `Ingress` and waiting for any corresponding `Route` resources to get created _might_ take longer than directly creating a new `Route` on a new cluster, but I don't think that's worth worrying about, and I only mention it for completeness.)

modules/registry-exposing-secure-registry-manually.adoc
```diff
@@ -14,117 +14,50 @@ images using the route host.
...
-----
+////
+
+OR
 
-. Log in to the registry using the information from securing the registry.
-However, this time point to the host name used in the route rather than your
-service IP. When logging in to a secured and exposed registry, make sure you
-specify the registry in the `podman login` command:
+. Expose the router using Custom Routes:
```
wzheng1: There is a bug about this part:  https://bugzilla.redhat.com/show_bug.cgi?id=1720270, so I suggest to remove this part, at least using secret part.

adambkaplan: I believe this is commented out (won't appear when the HTML is rendered).
Hopefully the fix will land in a z-stream soon and we'll be able to restore this section.

modules/registry-exposing-secure-registry-manually.adoc
```diff
@@ -14,117 +14,50 @@ images using the route host.
...
-+
-[NOTE]
-====
-Re-encrypt routes are also supported for exposing the secure registry.
-====
-
-.. If it does not exist, create the route with the `oc create route passthrough`
-command, specifying the registry as the route’s service. By default, the name of
-the created route is the same as the service name:
+. Expose the router using `DefaultRoute`, which is insecure:
```
adambkaplan:
```suggestion
. Expose the router using `DefaultRoute`:
```

The default route is in fact secured. It may not be _trusted_.

modules/registry-exposing-secure-registry-manually.adoc
```diff
@@ -14,117 +14,50 @@ images using the route host.
...
-$ oc get svc
-NAME              CLUSTER_IP       EXTERNAL_IP   PORT(S)                 SELECTOR                  AGE
-podman-registry   172.30.69.167    <none>        5000/TCP                podman-registry=default   4h
-kubernetes        172.30.0.1       <none>        443/TCP,53/UDP,53/TCP   <none>                    4h
-router            172.30.172.132   <none>        80/TCP                  router=router             4h
+$ oc patch config.imageregistry.operator.openshift.io/cluster --patch '{"spec":{"defaultRoute":true}}' --type=merge
 ----
 
-... Create the route:
+.. Log in insecurely with Podman:
```
adambkaplan:
```suggestion
.. Log in with Podman:
```

modules/registry-exposing-secure-registry-manually.adoc
```diff
@@ -14,117 +14,50 @@ images using the route host.
...
 +
 ----
-$ oc create route passthrough    \
-    --service=podman-registry    \//<1>
-    --hostname=<host>
-route "podman-registry" created     <2>
+HOST=$(oc get route default-route -n openshift-image-registry --template='{{ .spec.host }}')
+podman login -u $(oc whoami) -p $(oc whoami -t) --tls-verify=false $HOST
```
adambkaplan:
```suggestion
podman login -u $(oc whoami) -p $(oc whoami -t) --tls-verify=false $HOST <1>
----
<1> `--tls-verify=false` is needed if the cluster's default certificate for routes is untrusted. You can set a custom, trusted certificate as the default certificate via the Ingress Operator: https://docs.openshift.com/container-platform/4.1/networking/ingress-operator.html#nw-ingress-setting-a-custom-default-certificate_configuring-ingress
```

modules/registry-exposing-secure-registry-manually.adoc
wzheng1: OK, get it! Thanks!

wzheng1:
Can this pr be merged and change bug https://bugzilla.redhat.com/show_bug.cgi?id=1718163 to ON_QA? so that QE can verify it.

adambkaplan:
/hold

This still needs work. I also need to verify the podman login procedure works out of the box.

wzheng1:
QE use below steps to docker login with default route:
a.Get wildcard route ca
$oc get secret router-certs-default -n openshift-ingress -o yaml
$echo <tls.crt_content> | base64 -d > ca.crt
b. Trust ca in your client platform
$sudo cp ca.crt  /etc/pki/ca-trust/source/anchors/<default_route>.crt
#sudo update-ca-trust enable
c. Reload docker
#sudo systemctl daemon-reload
#sudo systemctl restart docker

modules/registry-exposing-secure-registry-manually.adoc
bmcelvee: That's correct, this section is still commented out.

bmcelvee:
@adambkaplan I made some updates, and we can look at it again early next week to get it merged. Thanks!

bmcelvee:

modules/registry-exposing-secure-registry-manually.adoc
adambkaplan: @bmcelvee fix is now verified, we can un-comment this section

modules/registry-exposing-secure-registry-manually.adoc
adambkaplan: Circling back on this - @bmcelvee you can remove the `--from-file`.

Creating the secret should be an _optional_ step - otherwise the route will use the default TLS configuration from the ingress operator.

modules/registry-exposing-secure-registry-manually.adoc
```diff
@@ -14,117 +14,53 @@ images using the route host.
...
-Status: Image is up to date for <host>/test/busybox:latest
-
-$ oc get imagestreams -n test
-NAME      PODMAN REPO                       TAGS      UPDATED
-busybox   172.30.11.215:5000/test/busybox   latest    2 seconds ago
+spec:
+  routes:
+    - name: public-routes
+      hostname: myregistry.mycorp.organization
+      secret name: public-route-tls
```
adambkaplan: This should be called out as optional - only set `secretName` if the customer is providing a custom TLS config for the registry's route.

```suggestion
      secretName: public-route-tls
```

adambkaplan (changes requested):
almost there

modules/registry-exposing-secure-registry-manually.adoc
```diff
@@ -5,126 +5,64 @@
...
+certificate with the xref:../networking/ingress-operator.adoc#configuring-ingress[Ingress Operator].
 
-. Log in to the registry using the information from securing the registry.
-However, this time point to the host name used in the route rather than your
-service IP. When logging in to a secured and exposed registry, make sure you
-specify the registry in the `podman login` command:
+
+Expose the registry using custom routes:
+
+. Create a secret:
```
adambkaplan:
```suggestion
. Optional - create a secret with your route's TLS keys:
```

adambkaplan (changes requested):
One minor change, otherwise looks good

modules/registry-exposing-secure-registry-manually.adoc
bmcelvee: Thanks! I added "Create a secret with your route's TLS keys" and have in a line following the command a statement that it's optional. Is that okay?

adambkaplan (approved):
/lgtm

modules/registry-exposing-secure-registry-manually.adoc
```diff
@@ -5,126 +5,64 @@
...
+
+. Create a secret with with your route's TLS keys:
 +
 ----
-# podman login -e user@company.com \
-    -u f83j5h6 \
-    -p Ju1PeM47R0B92Lk3AZp-bWJSck2F7aGCiZ66aFGZrs2 \
-    <host>
+$ oc create secret tls public-route-tls
+  - n image-registry\
```
adellape: Formatting/spaces fixes:

```
$ oc create secret tls public-route-tls \
    -n image-registry \
    --cert=</path/to/tls.crt> \
    --key=</path/to/tls.key> \
```

modules/registry-exposing-secure-registry-manually.adoc
```diff
@@ -5,126 +5,64 @@
...
+  -- cert=/path/to/tls.crt\
+  -- key=/path/to/tls.key\
 ----
-
-. You can now tag and push images using the route host. For example, to tag and
-push a `busybox` image in a project called `test`:
++
+This step is optional. if you do not create a secret, the route will use the
```
adellape: s/if/If/

adellape: s/will use/uses/

registry/securing-exposing-registry.adoc
```diff
@@ -6,12 +6,11 @@ toc::[]
 
 ifdef::openshift-origin,openshift-enterprise,openshift-dedicated[]
 
-By default, the {product-title} registry is secured during cluster
-installation so that it serves traffic via TLS. A passthrough route is also
-created by default to expose the service externally.
+By default, the {product-title} registry is secured during cluster installation
+so that it serves traffic through TLS. Unlike previous versions of
+{product-title}, the registry is not exposed outside of the cluster at the time
+of installation.
```
adellape: Do we know if this is true for OSD (since the ifdef includes `openshift-dedicated`)?

modules/registry-exposing-secure-registry-manually.adoc
```diff
@@ -5,126 +5,64 @@
...
-registry from outside the cluster using the route address, and to tag and push
-images using the route host.
+Instead of logging in to the {product-title} registry from within the cluster,
+you can gain external access to it by exposing it with a route. This allows you
+to log in to the registry from outside the cluster using the route address, and
+to tag and push images using the route host.
 
 .Prerequisites:
 
 * The following prerequisites are automatically preformed:
```
adellape: Preformed or performed?

adellape (changes requested):
Overall LGTM, couple minor things.

registry/securing-exposing-registry.adoc
bmcelvee: No, I'm not sure. But I'm going to go ahead and remove the ifdefs for now since they're from legacy content.

Thanks!

adellape (approved):

PR #15371, opened 2019-06-12
BZ-1719455: Removed references to Ceph.

huffmanca:
Removed references to Ceph in the 4.1 docs, as these are not yet supported.

This is for OS 4.x.


liangxia:
LGTM (from QE)

PR #15377, opened 2019-06-13
BZ-1720147 Launch config in AWS has a limit hard-coded in local-stored templates, but not when uploaded

ghost:
https://bugzilla.redhat.com/show_bug.cgi?id=1720147


kalexand-rh:
@DHF-DaHernan, thanks for opening this PR! It looks like you picked up a bunch of other changes, though. Will you please rebase against master and see if that fixes all of the duplicate changes?

PR #15381, opened 2019-06-13
bug 1720279 fixing array format

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1720279

@mazzystr, FYI

@jianlinliu, will you PTAL?


kalexand-rh:
Jianlin verified this change on the bug. Merging.

PR #15388, opened 2019-06-13
removed install/ directory

sheriff-rh:
This file should not have existed in 3.9 prior, and was pulled in from a previous commit. 

@kalexand-rh  PTAL. Set peer review needed.


kalexand-rh (commented):
LGTM!

PR #15402, opened 2019-06-13
Update quantity of identity provider parameters

jboxman:
@openshift/team-documentation, it was mentioned in an email that we state that four parameters are listed, but we only list two. Should it be _four_ or _two_ or just _several_ as I propose here?

Thanks.

### DIFF
--- a/modules/identity-provider-parameters.adoc
+++ b/modules/identity-provider-parameters.adoc
@@ -5,7 +5,7 @@
 [id="identity-provider-parameters_{context}"]
 = Identity provider parameters
 
-Four parameters are common to all identity providers:
+The following parameters are common to all identity providers:
 
 [cols="2a,8a",options="header"]
 |===


modules/identity-provider-parameters.adoc
```diff
@@ -5,7 +5,7 @@
 [id="identity-provider-parameters_{context}"]
 = Identity provider parameters
 
-Four parameters are common to all identity providers:
+Several parameters are common to all identity providers:
```
mburke5678: How about _The following paramters_? 
Wonder what happened to the other two?

huffmanca: I like `The following parameters`.

I believe that `claim` and `lookup` were removed, as these could be deduced by the engineering team depending on the identity provider; so these exist, but are not configured by the customer and are instead inferred.

PR #15414, opened 2019-06-14
BZ-1720471: Corrected reclaimation example.

huffmanca:
Corrected reclamation statement to not use unsupported volumes.

This is for OS 4.x.


liangxia:
LGTM (from QE)

PR #15425, opened 2019-06-14
Pod/Container can steal cpu/memory resources if cluster admin did not set minimum (not maximum) limitrange

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1644512

### DIFF
--- a/admin_guide/limits.adoc
+++ b/admin_guide/limits.adoc
@@ -26,6 +26,10 @@ enumerated constraints, then the resource is rejected. If the resource does not
 set an explicit value, and if the constraint supports a default value, then the
 default value is applied to the resource.
 
+For CPU and Memory limits, if you specify a `max` value, but do not specify a
+`min` limit, the resource can consume CPU/memory resources greater than `max` value`.
+
+
 ifdef::openshift-origin,openshift-enterprise[]
 You can specify limits and requests for ephemeral
 storage by using the ephemeral storage technology preview. This feature is
@@ -90,13 +94,19 @@ containers.
 <3> The maximum amount of memory that a pod can request on a node across all
 containers.
 <4> The minimum amount of CPU that a pod can request on a node across all
-containers.
+containers. Not setting a `min` value or setting `0` is unlimited allowing the 
+pod to consume more than than  the `max` value.
 <5> The minimum amount of memory that a pod can request on a node across all
-containers.
+containers. Not setting a `min` value or setting `0` is unlimited allowing the 
+pod to consume more than than  the `max` value.
 <6> The maximum amount of CPU that a single container in a pod can request.
 <7> The maximum amount of memory that a single container in a pod can request.
 <8> The minimum amount of CPU that a single container in a pod can request.
+Not setting a `min` value or setting `0` is unlimited allowing the 
+pod to consume more than than  the `max` value.
 <9> The minimum amount of memory that a single container in a pod can request.
+Not setting a `min` value or setting `0` is unlimited allowing the 
+pod to consume more than than  the `max` value.
 <10> The default amount of CPU that a container will be limited to use if not
 specified.
 <11> The default amount of memory that a container will be limited to use if not specified.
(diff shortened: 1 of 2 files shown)


mburke5678:
@sjenning @nak3  PTAL
This issue is related only to the CPU and memory settings in LimitRange objects only and not other CPU/Memory settings?

admin_guide/limits.adoc
```diff
@@ -172,7 +189,8 @@ Per container, the following must hold true if specified:
 (optional)
 
 If the configuration defines a `min` CPU, then the request value must be greater
-than the CPU value. A limit value does not need to be specified.
+than the CPU value. Not setting a `min` value or setting `0` is unlimited 
+allowing the pod to consume more than than  the `max` value.
```
nak3: `than than` must be a typo.

admin_guide/limits.adoc
```diff
@@ -224,17 +242,18 @@ Across all containers in a pod, the following must hold true:
 
 |`*Min*`
 |`Min[resource]` less than or equal to `container.resources.requests[resource]`
-(required) less than or equal to `container.resources.limits[resource]`
-(optional)
+(required) less than or equal to `container.resources.limits[resource]`.
+Not setting a `min` value or setting `0` is unlimited allowing the 
+pod to consume more than than  the `max` value.
```
nak3: Same with above `than than`.

nak3:
Thank you @mburke5678 

>  For CPU and Memory limits, if you specify a `max` value, but do not specify a
> `min` limit, the resource can consume CPU/memory resources greater than `max` value`.

This issue needs following two conditions:

- a. Admin does not specify `min` limit or does specify `0` in `Limitrange`.
- b. Users specify `0` in `resources.limits.{cpu,memory}` in `Pod`.

The fix does not mention about b, but do you have any idea how we can explain about b? (I am wondering that current fix is also OK, but if you have any good idea, please let me know.)

nak3:
> Which OpenShift versions does this apply to? 3.7+? 4.1?

All OpenShift and Kuberntes versions have this issue.

nak3:
lgtm, but defer to @sjenning

mburke5678:
Is it the case that setting `min: cpu:` and `min:memory` to `0` in a pod spec mean unlimited and allow the pod to consume more than `max` in all cases, or only when a LimitRange object exist? 
Either way, it might be worth mention in the docs. 
I did a quick search on `cpu:` and found 70 references in 22 files, not all of which are about setting the `cpu` and `memory` values. 
Just want a second opinion on whether we should add such a note in all situations where the user is setting these values for a pod; or mention once to the admin and let the admin tell users?

mburke5678:
@sjenning 
Is the LimitRange object in 4.x? It doesn't seem we brought that topic forward from 3.11 docs.

mburke5678:
@sjenning 
Is it the case that setting min: cpu: and min:memory to 0 in a pod spec mean unlimited and allow the pod to consume more than max in all cases, or only when a LimitRange object exist?
Either way, it might be worth mention in the docs.
I did a quick search on cpu: and found 70 references in 22 files, not all of which are about setting the cpu and memory values.
Just want a second opinion on whether we should add such a note in all situations where the user is setting these values for a pod; or mention once to the admin and let the admin tell users?

mburke5678:
@sjenning PTAL

sjenning:
> Is the LimitRange object in 4.x? It doesn't seem we brought that topic forward from 3.11 docs.

yes

> Is it the case that setting min: cpu: and min:memory to 0 in a pod spec mean unlimited and allow the pod to consume more than max in all cases, or only when a LimitRange object exist?

setting them both to 0 means the pod is best effort and can consume as much cpu and memory as the system makes available best effort pods (cpu scraps and memory not used by guaranteed pods).  The limit range can 1) not exist 2) exist but have no min 3) exist and have the min set to 0 and this is possible.

LGTM

PR #15438, opened 2019-06-17
BZ-1720781: Updated API examples to use oc get.

huffmanca:
Adjusted the API examples to use `oc get $RESOURCE -o yaml` instead of `oc describe $RESOURCE`.

This is for OS 4.x.


zhouying7780:
LGTM

PR #15474, opened 2019-06-19
Docker builds are not supported in ARO

thekad:
This warning should be displayed on ARO as well.

### DIFF
--- a/dev_guide/builds/build_strategies.adoc
+++ b/dev_guide/builds/build_strategies.adoc
@@ -12,7 +12,7 @@
 toc::[]
 
 
-ifdef::openshift-online[]
+ifdef::openshift-online,openshift-aro[]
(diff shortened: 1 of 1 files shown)


kalexand-rh:
@thekad, please set this PR against the `master-3` branch.

kalexand-rh:
Please rebase against `master-3`.

thekad:
@kalexand-rh yep, should've done it from my clone instead of trusting GH's UI, should be OK now down to a single commit.

kwoodson:
This looks good to me.  I do not see another reference to the openshift-aro product/project but syntactically looks correct.

kalexand-rh:
Thank you both! I double-checked, and in the 3.x stream, `openshift-aro` is correct.

PR #15479, opened 2019-06-20
updating CloudFormation templates

kalexand-rh:
https://github.com/openshift/installer/pull/1712/files merged, so I'm syncing the templates. I confirmed with @cuppett that these changes should be present in both 4.1 and 4.2.

@jianlinliu, will you PTAL?


jianlinliu:
LGTM.

PR #15483, opened 2019-06-20
osdocs-398 draft of cluster/machineautoscaling updates

kalexand-rh:
https://jira.coreos.com/browse/OSDOCS-398

@bison, @frobware, will you PTAL?

### DIFF
--- a/machine_management/applying-autoscaling.adoc
+++ b/machine_management/applying-autoscaling.adoc
@@ -30,12 +30,23 @@ include::modules/cluster-autoscaler-cr.adoc[leveloffset=+2]
 :FeatureName: ClusterAutoscaler
 include::modules/deploying-resource.adoc[leveloffset=+2]
 
+.Next steps
+
+* After you configure the ClusterAutoscaler, you must configure at least one
+MachineAutoscaler.
+
 [id="configuring-machineautoscaler"]
 == Configuring the MachineAutoscalers
 
-After you deploy the ClusterAutoscaler, you can
-deploy MachineAutoscaler resources for each of the machine types in your
-cluster to manage deployments of individual machines.
+After you deploy the ClusterAutoscaler,
+deploy MachineAutoscaler resources that reference the MachineSets that are used
+to scale the cluster.
+
+[IMPORTANT]
+====
+You must deploy at least one MachineAutoscaler resource after you deploy the
+ClusterAutoscaler resource.
+====
 
 [NOTE]
 ====
@@ -53,4 +64,3 @@ include::modules/deploying-resource.adoc[leveloffset=+2]
 
 * For more information about pod priority, see
 xref:../nodes/pods/nodes-pods-priority.adoc#nodes-pods-priority[Including pod priority in pod scheduling decisions in {product-title}].
-
--- a/modules/cluster-autoscaler-cr.adoc
+++ b/modules/cluster-autoscaler-cr.adoc
@@ -32,12 +32,12 @@ spec:
       - type: amd.com/gpu <7>
         min: 0 <8>
         max: 4 <9>
-  scaleDown:
-    enabled: true <10>
-    delayAfterAdd: 10m <11>
-    delayAfterDelete: 5m <12>
-    delayAfterFailure: 30s <13>
-    unneededTime: 60s <14>
+  scaleDown: <10>
+    enabled: true <11>
+    delayAfterAdd: 10m <12>
+    delayAfterDelete: 5m <13>
+    delayAfterFailure: 30s <14>
+    unneededTime: 60s <15>
 ----
 <1> Specify the priority that a pod must exceed to cause the ClusterAutoscaler
 to deploy additional nodes. Enter a 32-bit integer value. The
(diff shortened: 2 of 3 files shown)


modules/cluster-autoscaler-cr.adoc
```diff
@@ -48,16 +48,19 @@ you assign to each pod.
...
+value of `10m` is used.
 <12> Specify the period, in minutes, to wait before deleting a node after
-a node has recently been _deleted_.
+a node has recently been _deleted_. If you do not specify a value, the default
+value of `10s` is used.
 <13> Specify the period, in seconds, to wait before deleting a node after
-a scale down failure occured.
+a scale down failure occurred. If you do not specify a value, the default
+value of `3m` is used.
 <14> Specify the period, in seconds, before an unnecessary node is eligible
```
kalexand-rh: What's the default value for `unneededTime`?

bison: Default is 10 minutes.

modules/cluster-autoscaler-cr.adoc
```diff
@@ -48,16 +48,19 @@ you assign to each pod.
...
-<7> Specify the type of GPU node to deploy. Only `nvidia.com/gpu` and `amd.com/gpu`
-are valid types.
+<7> Optionally, specify the type of GPU node to deploy. Only
+`nvidia.com/gpu` and `amd.com/gpu` are valid types.
 <8> Specify the minimum number of GPUs to deploy.
 <9> Specify the maximum number of GPUs to deploy.
 <10> Specify whether the ClusterAutoscaler can remove unnecessary nodes.
-<11> Specify the period, in minutes, to wait before deleting a node after
-a node has recently been _added_.
+<11> Optionally, specify the period, in minutes, to wait before deleting a node after
```
bison: The values don't actually have to be in minutes, though in practice these usually would be. All of the time values in this section are just strings representing durations. So, any of the following are valid: `1s`, `1m`, `1m1s`, `1.1m`, `1h1m` etc. Anything supported by golang's [ParseDuration](https://golang.org/pkg/time/#ParseDuration).

modules/machine-autoscaler-about.adoc
```diff
@@ -9,4 +9,14 @@ The MachineAutoscaler adjusts the number of Machines in the MachineSets that you
 deploy in an {product-title} cluster. You can scale both the default `worker`
 MachineSet and any other MachineSets that you create. The MachineAutoscaler
 makes more Machines when the cluster runs out of resources to support more
-deployments.
\ No newline at end of file
+deployments. The MachineAutoscaler also reconciles any changes that you make to
```
bison: > The MachineAutoscaler also reconciles any changes that you make to the MachineSets, such as changing the minimum or maximum number of instances.

I think this is a good addition, but just my $0.02, I might reword to make it clear that the thing being modified is the MachineAutoscaler since users should not be modifying MachineSets directly. Maybe something along the lines of:

"Any changes to the values in MachineAutoscaler resources are immediately applied to the MachineSet they target."

bison:
I think this looks great. If it helps, I think the biggest thing we need to clarify is that scaling the cluster is at minimum a two step process. The first step being the creation of the ClusterAutoscaler resource, and the second step being creating one or more MachineAutoscaler resources. I think some people were confused and thought you only had to do one or the other.

So the bit added to the MachineAutoscaler doc here helps, but does it also need to be mentioned directly in the ClusterAutoscaler section?

kalexand-rh:
@bison, thank you for the feedback! Will you please take another look? All of the smaller files render into a single topic, and the rendered version is here: http://file.rdu.redhat.com/kalexand/062419/osdocs398/machine_management/applying-autoscaling.html

kalexand-rh:
@sunzhaohua2, will you PTAL?

sunzhaohua2:
LGTM, thank you!

kalexand-rh:
Thank you!

@bison, if you don't see any other changes, I'm going to merge.

modules/machine-autoscaler-about.adoc
```diff
@@ -9,4 +9,15 @@ The MachineAutoscaler adjusts the number of Machines in the MachineSets that you
...
+the minimum or maximum number of instances, are immediately applied
+to the MachineSet they target.
+
+[IMPORTANT]
+====
+You must deploy a MachineAutoscaler for the ClusterAutoscaler to scale your
+machines. The ClusterAutoscaler uses the annotations on MachineSets that the
+MachineAutoscaler sets to determine the resources that it can scale. If you
+define a ClusterAutoscaler without also defining MachineAutoscalers, the
+ClusterAutoscaler will never modify your cluster.
```
frobware: s/never/scale?

machine_management/applying-autoscaling.adoc
```diff
@@ -30,13 +30,24 @@ include::modules/cluster-autoscaler-cr.adoc[leveloffset=+2]
...
+.Next steps
+
+* After you configure the ClusterAutoscaler, you must configure at least one
+MachineAutoscaler.
+
 [id="configuring-machineautoscaler"]
 == Configuring the MachineAutoscalers
 
-After you deploy the ClusterAutoscaler, you can
+After you deploy the ClusterAutoscaler,
```
frobware: Not sure what we're trying to convey here. Maybe:

- After you deploy the ClusterAutoscaler, deploy MachineAutoscaler resources referencing the MachineSet that will be used to scale the cluster.

Dunno - opinions? @bison

modules/cluster-autoscaler-cr.adoc
```diff
@@ -48,16 +48,22 @@ you assign to each pod.
...
-<10> Specify whether the ClusterAutoscaler can remove unnecessary nodes.
-<11> Specify the period, in minutes, to wait before deleting a node after
-a node has recently been _added_.
-<12> Specify the period, in minutes, to wait before deleting a node after
-a node has recently been _deleted_.
-<13> Specify the period, in seconds, to wait before deleting a node after
-a scale down failure occured.
-<14> Specify the period, in seconds, before an unnecessary node is eligible
-for deletion.
+<10> In this section, you can specify the periods in by using any valid
```
frobware: "... specify the periods by using ..."

frobware: Though given the reference is to pkg/time/ParseDuration I wonder whether we should s/period/duration.

machine_management/applying-autoscaling.adoc
bison: Yeah, that might be a bit clearer.

modules/cluster-autoscaler-cr.adoc
kalexand-rh: They're synonyms, so normally I'd agree, but rewriting the rest of the statements to use "duration" instead of "period" make them a bit clunky. I think that "the period to wait for each action by using" might be most consistent with the subsequent content.

kalexand-rh:
@frobware, I've addressed your comments. If you're satisfied, I'll merge.

frobware:
> @frobware, I've addressed your comments. If you're satisfied, I'll merge.

/lgtm

PR #15484, opened 2019-06-20
Changed vm.overcommit to 1

mburke5678:
The `vm.overcommit_memory = 0` parameter in 4.x should be `vm.overcommit_memory = 1` as overcommit is enabled by default in 4.x

### DIFF
--- a/modules/nodes-cluster-overcommit-configure-nodes.adoc
+++ b/modules/nodes-cluster-overcommit-configure-nodes.adoc
@@ -27,24 +27,17 @@ by setting the `vm.panic_on_oom` parameter to `0`. A setting of 0 instructs the
 kernel to call oom_killer in an Out of Memory (OOM) condition, which kills
 processes based on priority
 
-You can view the current setting by running the following commands on your node:
+You can view the current setting by running the following commands on your nodes:
 
 ----
 $ sysctl -a |grep commit
 
-vm.overcommit_memory = 0
+vm.overcommit_memory = 1
 ----
 
 ----
 $ sysctl -a |grep panic
 vm.panic_on_oom = 0
-----
-
-You can change these settings using:
-
-----
-$ sysctl -w vm.overcommit_memory=1
-$ sysctl -w vm.panic_on_oom=0
 ----
 
 [NOTE]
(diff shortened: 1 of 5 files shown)


modules/nodes-cluster-overcommit-configure-nodes.adoc
```diff
@@ -43,8 +43,8 @@ vm.panic_on_oom = 0
 You can change these settings using:
 
 ----
-$ sysctl -w vm.overcommit_memory=1
-$ sysctl -w vm.panic_on_oom=0
+$ sysctl -w vm.overcommit_memory=0
+$ sysctl -w vm.panic_on_oom=1
```
rphillips: panic_on_oom should be set to 0... 0 does not cause the kernel to core.

rphillips (changes requested):

modules/nodes-cluster-overcommit-configure-nodes.adoc
mburke5678: Would there be a situation where the user might change to 1?

rphillips: `panic_on_oom=1` is not likely with k8s nodes since it causes the kernel to panic and halt.

rphillips (approved):

mburke5678:
@sunzhaohua2 Is this an issue you can review? A simple change to a parameter from `0` to `1` to show that overcommit is enabled by default for 4.1.

sunzhaohua2:
It's Node team's feature, @sunilcio will review it.

modules/nodes-cluster-overcommit-node-disable.adoc
```diff
@@ -0,0 +1,17 @@
...
+// * nodes/nodes-cluster-overcommit.adoc
+
+[id="nodes-cluster-overcommit-node-disable_{context}"]
+= Disabling overcommitment for a node
+
+When enabled, overcommitment can be disabled on each node.
+
+.Procedure
+
+To disable overcommitment in a project run the following command on a node:
```
sunilcio: @mburke5678 I guess here we need to mention node instead of project as we are disabling it on node.

PR #15485, opened 2019-06-20
bug 1714606 update DNS/hostname requirements

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1714606

### DIFF
--- a/modules/installation-dns-user-infra.adoc
+++ b/modules/installation-dns-user-infra.adoc
@@ -29,9 +29,12 @@ external to the cluster and from all the nodes within the cluster.
 |This DNS record must point to the load balancer
 for the control plane machines. This record must be resolvable from all the
 nodes within the cluster.
-
-|
-|
+[IMPORTANT]
+====
+The API server must be able to resolve the worker nodes by the host names
+that are recorded in Kubernetes. If it cannot resolve the node names, proxied
+API calls can fail, and you cannot retrieve information from Pods.
+====
 
 |Routes
 |`*.apps.<cluster_name>.<base_domain>`
--- a/modules/installation-dns-user-infra.adoc
+++ b/modules/installation-dns-user-infra.adoc
@@ -33,7 +33,7 @@ nodes within the cluster.
 ====
 The API server must be able to resolve the worker nodes by the host names
 that are recorded in Kubernetes. If it cannot resolve the node names, proxied
-API calls can fail, and you cannot retrieve information from Pods.
+API calls can fail, and you cannot retrieve logs from Pods.
 ====
 
 |Routes
--- a/modules/installation-dns-user-infra.adoc
+++ b/modules/installation-dns-user-infra.adoc
@@ -30,6 +30,9 @@ external to the cluster and from all the nodes within the cluster.
 for the control plane machines. This record must be resolvable from all the
 nodes within the cluster.
 
+|
+|
+
 |Routes
 |`*.apps.<cluster_name>.<base_domain>`
 |A wildcard DNS record that points to the load balancer that targets the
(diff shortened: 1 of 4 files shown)


modules/installation-dns-user-infra.adoc
```diff
@@ -30,6 +30,9 @@ external to the cluster and from all the nodes within the cluster.
 for the control plane machines. This record must be resolvable from all the
 nodes within the cluster.
 
+|
```
kalexand-rh: @brenton, @soltysh, is there specific guidance that we can provide in this DNS table to address the domain/host issue in the bug?

soltysh: I'll defer to @ironcladlou since he owns that area.

modules/installation-network-user-infra.adoc
```diff
@@ -14,7 +14,9 @@ establish a network connection to download their Ignition config files.
 
 It is recommended to use the DHCP server to manage the machines for the cluster
 long-term. Ensure that the DHCP server is configured to provide persistent IP
-addresses and host names to the cluster machines.
+addresses and host names to the cluster machines. DHCP must also configure
+a default DNS search zone. The Kubernetess API server must be able to resolve
```
brenton: I probably wouldn't state we have a requirement for a default DNS search zone.  The requirement is simply that all machines in the cluster can resolve each other.  In the case where API servers and worker nodes are in different zones then a default DNS search is one way to handle the problem.   Another way is simply having the hosts fully qualified names recorded in their node object.  The problem hit in testing was that Kubernetes didn't know the FQDNs so when the API Servers tried to proxy requests it didn't know how to perform the correct DNS lookup.

modules/installation-dns-user-infra.adoc
brenton: I'd probably add a note here that explains the importance of the API Server being able to resolve the worker nodes by the hostname recorded in Kubernetes.  You could mention that subtle problems with proxied API calls can happen otherwise.  You could show how using 'oc' to fetch logs from a pod would exhibit the problem.

kalexand-rh:
@brenton, are the change in the last acceptable?

modules/installation-dns-user-infra.adoc
```diff
@@ -29,6 +29,12 @@ external to the cluster and from all the nodes within the cluster.
 |This DNS record must point to the load balancer
 for the control plane machines. This record must be resolvable from all the
 nodes within the cluster.
+[IMPORTANT]
+====
+The API server must be able to resolve the worker nodes by the host names
+that are recorded in Kubernetes. If it cannot resolve the node names, proxied
+API calls can fail, and you cannot retrieve information from Pods.
```
brenton: Instead of saying "information" you could simply say "logs".

modules/installation-network-user-infra.adoc
```diff
@@ -16,6 +16,12 @@ It is recommended to use the DHCP server to manage the machines for the cluster
 long-term. Ensure that the DHCP server is configured to provide persistent IP
 addresses and host names to the cluster machines.
 
+The Kubernetes API server must be able to resolve the node names of the cluster
+machines. If the API servers and worker nodes are in different zones, you can
+configure a default DNS search zone to allow the API server to resolve the
+node names. The API server can also resolve the host names if you configure each
+host to record its fully-qualified domain name in its node object.
```
brenton: Instead of "The API server can also resolve the host names..." I might consider something like, "Another acceptable approach is simply to always refer to hosts by their fully-qualified domain name in both the node object and all DNS requests".

brenton:
:100: 

/lgtm

kalexand-rh:
QE approved this change on the bug, so I'm going to merge. Thanks!

PR #15496, opened 2019-06-21
Updated vSphere support matrix details

mylesagray:
Updated the vSphere support matrix to include details about the currently supported storage integration (in-tree VCP driver) and it's supported vSphere versions.

CSI support to be added later.

### DIFF
--- a/modules/installation-vsphere-infrastructure.adoc
+++ b/modules/installation-vsphere-infrastructure.adoc
@@ -5,22 +5,21 @@
 [id="installation-vsphere-infrastructure_{context}"]
 = VMware vSphere infrastructure requirements
 
-You must install the {product-title} cluster on a VMware vSphere version 6.5U2
+You must install the {product-title} cluster on a VMware vSphere version 6.5
 or 6.7U2 or later instance.
 
-VMware recommends using vSphere Version 6.7U2 or later with your {product-title}
+VMware recommends using vSphere Version 6.5 or later with your {product-title}
 cluster.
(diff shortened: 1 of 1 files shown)


modules/installation-vsphere-infrastructure.adoc
```diff
@@ -5,22 +5,21 @@
...
 = VMware vSphere infrastructure requirements
 
-You must install the {product-title} cluster on a VMware vSphere version 6.5U2
+You must install the {product-title} cluster on a VMware vSphere version 6.5
 or 6.7U2 or later instance.
 
-VMware recommends using vSphere Version 6.7U2 or later with your {product-title}
+VMware recommends using vSphere Version 6.5 or later with your {product-title}
 cluster.
 vSphere 6.7U2 includes:
```
kalexand-rh: @mylesagray, should this be 6.5 as well?

@katherinedube, will you PTAL?

katherinedube: 6.7U2 would still be a requirement for NSX-T SDN even if you remove vSAN CSI. So I think it should probably stay as 6.7U2. I'll let Myles confirm though.

mylesagray: Agree with @katherinedube - Let's leave it at 6.7 U2 but remove CSI reference.

kalexand-rh:
@mylesagray, will you squash your commits?

mylesagray:
@kalexand-rh - squashed

PR #15497, opened 2019-06-21
[enterprise-3.11] Correct reboot condition more clear

bysnupy:
- Version: `v3.11` 

- Description:
   Reboot does not happen by default during upgrade, but it's confusing like always rebooting after upgrade. [Upgrading to the latest OpenShift Container Platform release](https://docs.openshift.com/container-platform/3.11/upgrading/automated_upgrades.html#upgrading-ocp) is as follows. You can see reboot does not be required by default.

~~~
To enable rolling, full system restarts of the hosts, set the openshift_rolling_restart_mode parameter in your inventory file to system. Otherwise, the service is restarted on HA masters, but the systems do not reboot. See Configuring Cluster Variables for details.
~~~

### DIFF
--- a/upgrading/automated_upgrades.adoc
+++ b/upgrading/automated_upgrades.adoc
@@ -674,7 +674,7 @@ xref:customizing-node-upgrades[Customizing Node Upgrades], continue running the
 *_upgrade_nodes.yml_* playbook until all nodes are upgraded.
 // tag::automated_upgrade_after_reboot[]
 
-. After all master and node upgrades have completed, reboot all hosts.
+. After all master and node upgrades have completed, you can reboot all hosts if you required.
 
 . If you use aggregated logging, xref:upgrading-efk-logging-stack[upgrade the EFK logging stack].
 


upgrading/automated_upgrades.adoc
```diff
@@ -674,7 +674,7 @@ xref:customizing-node-upgrades[Customizing Node Upgrades], continue running the
 *_upgrade_nodes.yml_* playbook until all nodes are upgraded.
 // tag::automated_upgrade_after_reboot[]
 
-. After all master and node upgrades have completed, reboot all hosts.
+. After all master and node upgrades have completed, you can reboot all hosts if you required.
```
kalexand-rh: Why would you need to reboot?

bysnupy: If only openshift upgrade does not be required reboot, but some cases may be required. For instance, a user add some tasks using [Ansible hooks for upgrades](https://docs.openshift.com/container-platform/3.11/upgrading/automated_upgrades.html#upgrade-hooks) which is required host reboot for his upgrade playbooks. So I could not express "no reboot for upgrade" due to some irregular ones.

vikram-redhat:
@kalexand-rh - can we merge this or close?

kalexand-rh:
@jianlinliu, for 3.x, do you agree that it's not always required to reboot your hosts after upgrade and that you only need to if some other dependency, like a webhook, requires it? 

@vikram-redhat, I think we'd need a follow-up to provide exactly why you'd need to reboot, or we'll get more questions.

jianlinliu:
> for 3.x, do you agree that it's not always required to reboot your hosts after upgrade and that you only need to if some other dependency, like a webhook, requires it?

Agree, reboot is not a MUST sometimes. But how to distinguish when reboot is required after host update? E.g: I remember some months ago, dnsmasq was updated in RHEL Channel, host was upgrading to latest version, user have to reboot host to make it working. Or else, install/upgrade would failed. But user can not tell when or whether need reboot host.

kalexand-rh:
I think it would make a worse user experience to run into unexpected issues because they didn't reboot than to reboot when they didn't really need to, so I think we should close this PR. @jianlinliu, @bysnupy, @vikram-redhat, do you agree?

vikram-redhat:
@kalexand-rh yes, works for me.

bysnupy:
@kalexand-rh yes, no problem. Thanks.

kalexand-rh:
Thanks!

PR #15504, opened 2019-06-21
BZ-1696310: Clarified post-installation usage.

huffmanca:
Cleaned up instructions regarding using this section on an existing cluster. I also added the `ansible-playbook` command, as this procedure looked off having only a single step.

This is for OS 3.x.


jianlinliu:
LGTM.

PR #15526, opened 2019-06-24
Playbook to just perform logging cert rotation

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1710424

### DIFF
--- a/install_config/aggregate_logging.adoc
+++ b/install_config/aggregate_logging.adoc
@@ -2344,47 +2344,28 @@ $ curl --key /etc/elasticsearch/secret/admin-key \
 [[fluentd-redeploy-certs]]
 == Redeploying EFK Certificates
 
-You can redeploy EFK certificates, if needed.
+You can use an Ansible playbook to perform a certificate rotation for the EFK stack without needing to run the install/upgrade playbook.
+
+This playbook deletes the current certificate files, generates new EFK certificates, updates certificate secrets, and restarts Kibana and Elasticsearch 
+to force those components to read in the updated certificates.
 
 To redeploy EFK certificates:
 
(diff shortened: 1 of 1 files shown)


mburke5678:
@ewolinetz PTAL

install_config/aggregate_logging.adoc
```diff
@@ -2344,47 +2344,28 @@ $ curl --key /etc/elasticsearch/secret/admin-key \
 [[fluentd-redeploy-certs]]
 == Redeploying EFK Certificates
 
-You can redeploy EFK certificates, if needed.
+You can use an Ansible playbook to perform a certificate rotation for the EFK stack without needing to run the install/upgrade playbook.
+
+This playbook deletes the current certificate files, generates new EFK certificates, updates certificate secrets, and restarts components
```
ewolinetz: so, the playbook doesn't restart every component -- just kibana and elasticsearch. Not sure if we want to point that out.

fluentd will reload the certificates upon failing to connect to elasticsearch and a curator pod is created based on the cronjob schedule.

ewolinetz:
question about clarity for readers, otherwise lgtm

install_config/aggregate_logging.adoc
mburke5678: Thanks for the clarification. I took the DocText in th eBZ too literally, it appears!

mburke5678:
@anpingli Small addition to the docs for an Ansible playbook to redeploy EFK certificates. Please take a look. 
Do you know which versions this applies to? We update back to 3.7.

anpingli:
LGTM,  It is for 3.11 only.

PR #15527, opened 2019-06-24
OSDOCS-349 Update glossary reference from router to Ingress

bmcelvee:
https://jira.coreos.com/browse/OSDOCS-349

### DIFF
--- a/contributing_to_docs/term_glossary.adoc
+++ b/contributing_to_docs/term_glossary.adoc
@@ -213,6 +213,25 @@ The file that Ignition uses to configure Red Hat Enterprise Linux CoreOS during
 operating system initialization. The installation program generates different
 Ignition config files to initialize bootstrap, master, and worker nodes.
 
+=== Ingress
+
+Usage: Ingress
+
+API object that allows developers to expose Services through an HTTP(S) aware
+load balancing and proxy layer via a public DNS entry. The Ingress resource may
+further specify TLS options and a certificate, or specify a public CNAME that
+the OpenShift Ingress Controller should also accept for HTTP and HTTPS traffic.
+An administrator typically configures their Ingress Controller to be visible
+outside the cluster firewall, and may also add additional security, caching, or
+traffic controls on the Service content.
+
+=== Ingress Controller
+
+Usage: Ingress Controller(s)
+
+A resource that forwards traffic to endpoints of Services. The Ingress Controller
+replaces router from {product-title} 3 and earlier.
+
 == K
 
 ''''
@@ -392,11 +411,11 @@ Usage: Route(s)
 
 OpenShift-specific API object that allows developers to expose Services through
 an HTTP(S) aware load balancing and proxy layer via a public DNS entry. The
-Route may further specify TLS options and a certificate, or specify a public
-CNAME that the OpenShift router should also accept for HTTP and HTTPS traffic.
-An administrator typically configures their router to be visible outside the
-cluster firewall, and may also add additional security, caching, or traffic
-controls on the Service content.
+route may further specify TLS options and a certificate, or specify a public
+CNAME that the OpenShift Ingress Controller should also accept for HTTP and
+HTTPS traffic. An administrator typically configures their Ingress Controller to
+be visible outside the cluster firewall, and may also add additional security,
+caching, or traffic controls on the Service content.
 
 == S
 


bmcelvee:
@Miciah would you mind taking a look please? Just want to make sure the definition is still correct. Thanks!

contributing_to_docs/term_glossary.adoc
```diff
@@ -386,17 +386,17 @@ Pod are running at all times. Used by DeploymentConfigs.
 Do not use: replication controller(s)
 
 ''''
-=== Route
+=== Ingress
 
-Usage: Route(s)
+Usage: Ingress, Ingress Controller(s)
 
 OpenShift-specific API object that allows developers to expose Services through
```
Miciah: `Ingress` is not OpenShift-specific; it comes from Kubernetes.

contributing_to_docs/term_glossary.adoc
```diff
@@ -386,17 +386,17 @@ Pod are running at all times. Used by DeploymentConfigs.
...
 an HTTP(S) aware load balancing and proxy layer via a public DNS entry. The
-Route may further specify TLS options and a certificate, or specify a public
-CNAME that the OpenShift router should also accept for HTTP and HTTPS traffic.
-An administrator typically configures their router to be visible outside the
-cluster firewall, and may also add additional security, caching, or traffic
-controls on the Service content.
+Ingress Controller may further specify TLS options and a certificate, or specify
```
Miciah: ```Ingress Controller```→``` `Ingress` resource ``` on this line.  (Technically, the `Ingress` resource may specify _the name of a secret that contains_ the certificate, whereas the `Route` resource specifies the certificate itself, but it is probably safe to gloss over that detail.)

contributing_to_docs/term_glossary.adoc
```diff
@@ -386,17 +386,17 @@ Pod are running at all times. Used by DeploymentConfigs.
...
+a public CNAME that the OpenShift Ingress Controller should also accept for HTTP
+and HTTPS traffic. An administrator typically configures their router to be
```
Miciah: "router"→"Ingress Controller".

contributing_to_docs/term_glossary.adoc
```diff
@@ -386,17 +386,17 @@ Pod are running at all times. Used by DeploymentConfigs.
 Do not use: replication controller(s)
 
 ''''
-=== Route
+=== Ingress
```
Miciah: The entry for "Ingress" needs to be moved to maintain alphabetization.

Miciah (commented):
The `Route` resource is not going away, so we should document both the `Route` resource and the `Ingress` resource.  However, we are moving away from the "router" terminology for the component, which we now instead refer to as the "ingress controller".  It may be helpful to have an entry for "router" too though, if only to refer the reader to the new name.  So I suggest the following entries:

* `Ingress` resource (the new entry, with some changes I suggest in comments).
* `Route` resource (the old entry).
* The ingress controller.
* The router ("Old name for the ingress controller, q.v.").

bmcelvee:
Thanks so much, @Miciah! I separated the Ingress and Route resources. Would it be okay to add an entry for Ingress Controller where we state that it's replacing router?

Miciah:
> Would it be okay to add an entry for Ingress Controller where we state that it's replacing router?

That would make sense to me.  Also, we should distinguish between an Ingress Controller and an `Ingress` resource: a developer creates an `Ingress` resource to expose a `Service`, and the Ingress Controller is what actually forwards traffic to the `Service` (or to be more technically accurate, it forwards traffic to the endpoints of the `Service`).

bmcelvee:
@openshift/team-documentation PTAL, thanks! :)

contributing_to_docs/term_glossary.adoc
```diff
@@ -213,6 +213,25 @@ The file that Ignition uses to configure Red Hat Enterprise Linux CoreOS during
...
+the OpenShift Ingress Controller should also accept for HTTP and HTTPS traffic.
+An administrator typically configures their Ingress Controller to be visible
+outside the cluster firewall, and may also add additional security, caching, or
+traffic controls on the Service content.
+
+=== Ingress Controller
+
+Usage: Ingress Controller(s)
+
+A resource that forwards traffic to endpoints of Services. Ingress Controller
```
mburke5678: To be *really* picky, s/Ingress Controller/An Ingress Controller

mburke5678:
@bmcelvee A tiny nit that you can ignore, being this an internal reference doc. LGTM

bmcelvee:
Thank you, @mburke5678!

PR #15576, opened 2019-06-26
adding kebab to cnv attributes

ousleyp:
The `:kebab:` attribute was not in `cnv-document-attributes.adoc`, so I'm adding it. I'm also removing the `include::modules/common-attributes.adoc` from the CNV uninstall assembly, so that it doesn't depend on OCP.


kalexand-rh:
@aburdenthehand, @bgaydosrh, FYI

PR #15599, opened 2019-06-26
Update new machineset docs

mburke5678:
When i was playing with creating machinesets, I kept missing variables that needed to be changed from <variable>.  I put a call-out on each instance. Also, I bumped the _check an existing MachineSet from your cluster_ step to sub-steps, as they aren't separate steps, really. Just steps to get the information for Step 1.

### DIFF
--- a/modules/machineset-creating.adoc
+++ b/modules/machineset-creating.adoc
@@ -23,10 +23,8 @@ as shown, and is named `<file_name>.yaml`.
 +
 Ensure that you set the `<clusterID>` and `<role>` parameter values.
 
-. If you are not sure about which value to set for an specific field, you can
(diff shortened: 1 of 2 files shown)


mburke5678:
@kalexand-rh Any interest in doing something like this PR? 
When i was playing with creating machinesets, I kept missing variables that needed to be changed from `<variable`. Thought I would put a call-out on each instance. 
Also, bumped the _check an existing MachineSet from your cluster_ step to sub-steps, as they aren't separate steps, rally. Just steps to get the information for Step 1.

mburke5678:
@enxebre PTAL

mburke5678:
@sunzhaohua2 Is this something you can review?

PR #15600, opened 2019-06-26
Editing CLO section of Infra MachineSet topc

mburke5678:
Updating the section on creating an Infra machineset to add how to move cluster logging to an Infra node.

### DIFF
--- a/modules/infrastructure-moving-logging.adoc
+++ b/modules/infrastructure-moving-logging.adoc
@@ -3,10 +3,85 @@
 // * machine_management/creating-infrastructure-machinesets.adoc
 
 [id="infrastructure-moving-logging_{context}"]
-= Moving the logging aggregation solution
+= Moving the cluster logging resources
 
-[IMPORTANT]
+You can configure the Cluster Logging Operator to deploy the pods 
+for any or all of the Cluster Logging components, Elasticsearch, Kibana, and Curator to different nodes. 
+You cannot move the Cluster Logging Operator pod from its installed location. 
+
+For example, you can move the Elasticsearch pods to a separate node because of 
+high CPU, memory, and disk requirements.
+
+[NOTE]
 ====
-Cluster logging in {product-title} is not installed by default.
-You currently cannot move the logging component to another MachineSet.
+You should set your MachineSet to use at least 6 replicas.
 ====
+
+.Prerequisites
+
+* Cluster logging and Elasticsearch must be installed. These features are not installed by default.
+
+* If needed, get the name of the Cluster Logging Custom Resource in the openshift-logging project:
++
+----
+$ oc get ClusterLogging
+NAME       AGE
+instance   112m
+----
+
+.Procedure
+
+. Edit the Cluster Logging Custom Resource:
++
+----
+apiVersion: logging.openshift.io/v1
+kind: ClusterLogging
+
+....
+
+spec:
+  collection:
+    logs:
+      fluentd:
+        resources: null
+      rsyslog:
+        resources: null
+      type: fluentd
+  curation:
+    curator:
+      nodeSelector: <1>
+          node-role.kubernetes.io/infra: ''
+      resources: null
+      schedule: 30 3 * * *
+    type: curator
+  logStore:
+    elasticsearch:
+      nodeCount: 3
+      nodeSelector: <1>
+          node-role.kubernetes.io/infra: ''
+      redundancyPolicy: SingleRedundancy
+      resources:
+        limits:
+          cpu: 500m
+          memory: 4Gi
+        requests:
+          cpu: 500m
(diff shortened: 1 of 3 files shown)


modules/infrastructure-moving-logging.adoc
```diff
@@ -3,10 +3,84 @@
...
+Fluentd, to different nodes. You cannot move the Cluster Logging Operator pod from its installed location. 
+
+[NOTE]
 ====
-Cluster logging in {product-title} is not installed by default.
-You currently cannot move the logging component to another MachineSet.
+You should set your MachineSet to use at least 6 replicas.
 ====
+
+.Prerequisite
```
kalexand-rh: Prerequisites

modules/infrastructure-moving-logging.adoc
```diff
@@ -3,10 +3,84 @@
...
+
+* Cluster logging and Elasticsearch must be installed. These features are not installed by default.
```
kalexand-rh: Logging

modules/infrastructure-moving-logging.adoc
```diff
@@ -3,10 +3,84 @@
...
+      proxy:
+        resources: null
+      replicas: 1
+      resources: null
+    type: kibana
+
+....
+
+----
+<1> Add a `nodeSelector` parameter with the appropriate `<key>: <value>` pair to the component you want to move.
```
kalexand-rh: Will you add an example that references the `node-role.kubernetes.io/infra: ""` `nodeSelector`?

kalexand-rh (commented):
Two picks and a request. Will you tag QE?

mburke5678:
@kalexand-rh I will have QE take a look. Wanted to make sure the material fit into the topic properly first. Thanks for looking!

kalexand-rh:
@mburke5678, I'm happy to do it! If you add the example that I asked for, I think it'll fit in just fine.

modules/infrastructure-moving-logging.adoc
mburke5678: We have been using `cluster logging` in lowercase when talking about the feature in general.

modules/infrastructure-moving-logging.adoc
mburke5678: Is the `node-role.kubernetes.io/infra: ""` node selector required to your knowledge? I couldn't get it to work with that format. I tried the simple `tag=value` format, which was how QE tested the feature, and it worked.

kalexand-rh: If it's not, then we should check the other node selectors.

mburke5678:
@anpingli Can you please take a look? 
Also, @kalexand-rh [suggested I should make the node selectors in the custom resource `node-role.kubernetes.io/infra: ""`](https://github.com/openshift/openshift-docs/pull/15600#discussion_r298237466).
When I changed them from `es: yes` to `node-role.kubernetes.io/infra: ""` I got 6 new fluentd pods on the right node (the original 6 Fluentd pods remain, even after I remove the `es: yes` selector from the nodes). 
The Elasticsearch pods did not move {even after I remove the `es: yes` selector from the nodes)/
The Kibana pod moved the new nodes.

anpingli:
@mburke5678  I think the goal is to place the ES pods into specified nodes, as ES pods require huge CPU/Memory/Disk resources.    so it is suggested adding nodeSelector for ES.   it is optional to add nodeSelector for kibana/cronjob.   It is not suggested to add nodeSeletor for damonset, as we expected the fluentd/rsyslog locate in each node

mburke5678:
@anpingli I removed the node selector from the `spec.collector.logs.fuentd` section. And added a short note at the top "you can move the Elasticsearch pods to a separate node because of 
high CPU, memory, and disk requirements."

anpingli:
LGTM

mburke5678:
@ewolinetz Please take a look at this procedure to move the cluster logging components to new nodes.

modules/infrastructure-moving-logging.adoc
```diff
@@ -3,10 +3,85 @@
...
-Cluster logging in {product-title} is not installed by default.
-You currently cannot move the logging component to another MachineSet.
+You should set your MachineSet to use at least 6 replicas.
 ====
+
+.Prerequisites
+
+* Cluster logging and Elasticsearch must be installed. These features are not installed by default.
+
+* If needed, get the name of the Cluster Logging Custom Resource in the openshift-logging project:
```
ewolinetz: the clusterlogging CR will always need to be named `instance` fyi. if it isn't the operator will print out a message and not take action on it.

ewolinetz:
small point about the name of the CLO watched resource. Non-blocker though, lgtm.

PR #15608, opened 2019-06-26
BZ-1724311: Updated assemblies to include references.

huffmanca:
Updated the `understanding identity provider` assembly to include references to both RBAC and removing the default kubeadmin user.

This is for OS 4.x.


huffmanca:
Merging this one. It's a simple addition that doesn't require QE or peer review.

PR #15309, opened 2019-06-10
add pipeline env recommendation based on recent customer interaction

bmcelvee:
Recovered from https://github.com/openshift/openshift-docs/pull/14558

PR #15618, opened 2019-06-26
PR15309 follow-up minor edits

bmcelvee:
Follow-up edits to https://github.com/openshift/openshift-docs/pull/15309


bmcelvee:
Running into issues with out cherrypick process. Manually cherrypicked and pushed to 3.6, 3.7, 3.9, 3.10, and 3.11. Only 3.7 and 3.11 are showing up.

PR #15642, opened 2019-06-28
Update apiserver.doc

sanchezl:
Remove the instructions on replacing the default certificate because if followed, as written, will also replace the internal serving certificates and leave the cluster in a degraded state.

Following the instructions for adding a named certificate would be the appropriate way to override the certificate used to serve traffic over the external `api.{cluster}.{domain}` name.

### DIFF
--- a/authentication/certificates/api-server.adoc
+++ b/authentication/certificates/api-server.adoc
@@ -10,6 +10,4 @@ cluster CA. Clients outside of the cluster will not be able to verify the
 API server's certificate by default. This certificate can be replaced
 by one that is issued by a CA that clients trust.
 
-include::modules/customize-certificates-api-add-default.adoc[leveloffset=+1]
-
 include::modules/customize-certificates-api-add-named.adoc[leveloffset=+1]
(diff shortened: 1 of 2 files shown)


kalexand-rh:
@sanchezl, does this apply to 4.1, too? If so, please target the PR to `master` and rebase.

sanchezl:
/hold
for rebase

sanchezl:
/hold cancel
rebased

deads2k:
> @sanchezl, does this apply to 4.1, too? If so, please target the PR to `master` and rebase.

@kalexand-rh we're going ahead with this related to https://bugzilla.redhat.com/show_bug.cgi?id=1720770 .  Is there another page we need to delete too?  The setting is going away entirely.

huffmanca:
This LGTM. There should be no other sections that discuss modifying the API server's certificate. We discuss how the authentication occurs in a couple of other places, but we never patch the apiserver anywhere else.

huffmanca:
Confirmed that we're ready to proceed on this. Merging and cherry-picking.

PR #15644, opened 2019-06-28
Add master node sizing recommendations

chaitanyaenr:
This commit adds information to the Scalabilty and Performance doc about the
size of master nodes based on the data collected during the control plane
density test runs at different node scale. This will help customers understand
the master node requirements depending on the number of nodes in the cluster.

### DIFF
--- a/modules/master-node-sizing.adoc
+++ b/modules/master-node-sizing.adoc
@@ -1,0 +1,36 @@
+// Module included in the following assemblies:
+//
+// * scalability_and_performance/recommended-host-practices.adoc
+
+[id="master-node-sizing_{context}"]
+=  Master node sizing
+
+The master node resource requirements depend on the number of nodes in the cluster. The following master node size recommendations are based on the results of control plane density focused testing.
+
+[options="header",cols="3*"]
+|===
+| Number of worker nodes |CPU cores |Memory (GB)
+
+| 25
+| 4
+| 16
+
+| 100
+| 8
+| 32
+
+| 250
+| 16
+| 64
+
+|===
+
+[IMPORTANT]
+====
+Because you cannot modify the master node size in a running {product-title} {product-version} cluster, you must estimate your total node count and use the suggested master size during installation.
+====
+
+[NOTE]
+====
+In {product-title} {product-version}, half of a CPU core (500 millicore) is now reserved by the system by default compared to {product-title} 3.11 and previous versions. The sizes are determined taking that into consideration.
+====
--- a/scalability_and_performance/recommended-host-practices.adoc
+++ b/scalability_and_performance/recommended-host-practices.adoc
@@ -9,6 +9,8 @@ This topic provides recommended host practices for {product-title}.
 
 include::modules/recommended-node-host-practices.adoc[leveloffset=+1]
 
+include::modules/master-node-sizing.adoc[leveloffset=+1]
+
 include::modules/create-a-kubeletconfig-crd-to-edit-kubelet-parameters.adoc[leveloffset=+1]
 
 include::modules/recommended-etcd-practices.adoc[leveloffset=+1]


chaitanyaenr:
@jeremyeder @akrzos @ashishkamra @mffiedler PTAL.

modules/master-node-sizing.adoc
```diff
@@ -0,0 +1,30 @@
...
+Control plane density focused tests were run at different node scale to determine the master node resource requirements depending on the number of nodes in the cluster. The recommended sizes are
+----
+25 nodes: 4 CPU cores and 16GB memory
+100 nodes: 8 CPU cores and 32GB memory
+250 nodes: 16 CPU cores and 64GB memory
+----
+
+====
+[NOTE]
+In {product-title} {product-version}, half of a CPU core (500 millicore) is now reserved by the system by default compared to {product-title} 3.11 and previous versions. The sizes are determined taking that into consideration.
```
chaitanyaenr: 1 core is reserved by the system in the case of OpenShift dedicated like @jeremyeder mentioned, should we mention that here or just in the OpenShift dedicated docs?

kalexand-rh:

modules/master-node-sizing.adoc
```diff
@@ -0,0 +1,39 @@
...
+|===
+
+[NOTE]
+====
+In {product-title} {product-version}, half of a CPU core (500 millicore) is now reserved by the system by default compared to {product-title} 3.11 and previous versions. The sizes are determined taking that into consideration.
+====
+
+[NOTE]
+====
+M5 instances offer better performance with an upgraded CPU, AWS Elastic Network Adapter for faster networking speed and NVMe storage for increased disk reads and write speeds when compared to M4 instances.
```
jmencak: Gone are the days of cloud-provider neutral docs?  I remember I faced a similar issue when documenting [HAProxy performance](https://docs.openshift.com/container-platform/4.1/scalability_and_performance/routing-optimization.html#baseline-router-performance_routing-optimization) on a "typical" cloud provider instance.  We went for something like 4vCPU/16GB RAM in the end.  Are we going to maintain similar docs for GKE/Azure?  Is it possible to keep this a bit more neutral?

chaitanyaenr: Thanks for the feedback @jmencak. The idea here is to make it easy for users to know what instance types to use depending on the cloud provider along with resources needed. It's just AWS for now but we need to add instance sizes recommendations for other cloud providers as we move forward.

Thinking this through, it's better to keep it a bit neutral I guess.

vikram-redhat:
@chaitanyaenr when ready to merge let us know.

@kalexand-rh this looks ok to me. Normally, it would be handled by @ahardin-rh. Since this is a small one, please just do a peer review and merge when @chaitanyaenr and @mffiedler give ack.

modules/master-node-sizing.adoc
```diff
@@ -0,0 +1,39 @@
...
+| 250
+| 16
+| 64
+| m5.4xlarge
+
+|===
+
+[IMPORTANT]
+====
+In {product-title} {product-version}, the master node size cannot be modified on a running cluster. It's highly recommended to estimate the node count and use the suggested sizes.
```
akrzos: I would suggest changing "it's" to "it is"

modules/master-node-sizing.adoc
chaitanyaenr: @vikram-redhat thoughts on this? Should we add cloud provider specific stuff to the doc or just stick with neutral docs?

modules/master-node-sizing.adoc
```diff
@@ -0,0 +1,36 @@
+// Module included in the following assemblies:
+//
+// * scalability_and_performance/recommended-host-practices.adoc
+
+[id="master-node-sizing_{context}"]
+=  Master node sizing
+
+Control plane density focused tests were run at different node scale to determine the master node resource requirements depending on the number of nodes in the cluster. The recommended sizes are
```
kalexand-rh: I'd replace this line with the following text:

>  The master node resource requirements depend on the number of nodes in the cluster. The following master node size recommendations are based on the results of control plane density focused testing.

Is this the number of master nodes or the number of total nodes?  If so, I'd clarify that in the table.

modules/master-node-sizing.adoc
```diff
@@ -0,0 +1,36 @@
...
+
+| 250
+| 16
+| 64
+
+|===
+
+[IMPORTANT]
+====
+In {product-title} {product-version}, the master node size cannot be modified on a running cluster. It is highly recommended to estimate the node count and use the suggested sizes.
```
kalexand-rh: Maybe 
> Because you cannot modify the master node size in a running {product-title} {product-version} cluster, you must estimate your total node count and use the suggested master size during installation.

kalexand-rh (commented):
I have a couple of text suggestions, but this looks fine.

For what it's worth, I think it would be helpful to provide the instance sizes for different cloud providers. I'd wait until after all the AWS regions use M5 instances by default, though.

chaitanyaenr:
@kalexand-rh thanks for the feedback, updated the PR to address the comments. PTAL.

kalexand-rh:
I think it looks good! @mffiedler, @chaitanyaenr, let me know when you're ready to merge.

mffiedler (approved):
/lgtm

chaitanyaenr:
@kalexand-rh the PR is ready to merge. Thanks.

modules/master-node-sizing.adoc
vikram-redhat: In an ideal world: cloud provider specific. In reality, we may have to stick with neutral docs. 

What we could do as we have a better idea on how popular each cloud provider is with our customer base, we could look at moving to the cloud provider specific docs and adding each provider as the demand rises from the field.

PR #15665, opened 2019-06-28
bug 1725079 adding bare metal install context

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1725079


modules/installation-bare-metal-config-yaml.adoc
```diff
@@ -56,23 +56,32 @@ machines.
...
+you, so you must set the number of workers to `0` in this file. Remember that you
 must manually deploy at least one compute, or worker, machine in the cluster.
 <5> The number of control plane machines that you add to the cluster. Because
 the cluster uses this values as the number of etcd endpoints in the cluster, the
 value must match the number of control plane machines that you deploy.
-<6> The cluster name.
-<7> A block of IP addresses from which pod IPs are allocated. This block must
-not overlap with existing physical networks.
+<6> The cluster name that you specified in your DNS records.
+<7> A block of IP addresses from which Pod IP addresses are allocated. This block must
```
kalexand-rh: @dcbw, do these parameter descriptions sound right to you?

modules/installation-registry-storage-non-production.adoc
```diff
@@ -8,7 +8,8 @@
 = Configuring storage for the image registry in non-production clusters
 
 You must configure storage for the image registry Operator. For non-production
-clusters, you can set the image registry to an empty directory.
+clusters, you can set the image registry to an empty directory. If you do so,
+then you cannot store images.
```
dmage: If you do so, all images will be lost after the registry gets restarted.

kalexand-rh: Thank you @dmage! I've updated the text.

PR #15669, opened 2019-07-01
search: avoid iframe

jiri-fiala:
This replaces the current iframe search used in the modal window when searching via the input field above docs topic selection on the left hand side. The new style can be checked [here](http://httpd-test-pro.b9ad.pro-us-east-1.openshiftapps.com/container-platform/3.11/welcome/) (only the 3.11 OCP variant). Avoiding the iframe fixes rendering on iOS devices, which [ignore the iframe height](https://stackoverflow.com/questions/34320046/iframe-height-issues-on-ios-mobile-safari), thus prevent normal scrolling, with the current style that assumes limited iframe height in the modal window.

The main page search (that searches all the doc variants and is not rendered in a modal window) is adjusted in the same manner. This universal search style can be checked [here](http://httpd-test-pro.b9ad.pro-us-east-1.openshiftapps.com/index.html).

The `master` and `master-3` branch differs in regards of `_templates/_search_*` files presence, but the `_templates/_page_openshift.html.erb` is there in both cases (slightly different though; the `master-3` version has [this Beta notice](https://github.com/openshift/openshift-docs/blob/master-3/_templates/_page_openshift.html.erb#L110)). @vikram-redhat, please let me know if a separate `master-3` style PR is required.

### DIFF
--- a/_javascripts/hc-search.js
+++ b/_javascripts/hc-search.js
@@ -1,51 +1,105 @@
 function hcSearchCategory(label, version) {
 // optional version filters search results for a single specific product version
-// currently can be used with OCP and Origin only
+// currently can be used with OCP and OKD docs only
 
-  var modalSearch = document.getElementById("hc-search-modal");
-  var searchBtn = document.getElementById("hc-search-btn");
-  var closeModal = document.getElementById("hc-modal-close");
-  var searchResults = document.getElementById("hc-search-results");
-  var query = document.getElementById("hc-search-input");
+  // elements used repeatedly:
+  var modalSearch = $("#hc-search-modal");
+  var searchBtn = $("#hc-search-btn");
+  var closeModal = $("#hc-modal-close");
+  var query = $("#hc-search-input");
 
   // pressing enter in the input = search btn click
-  query.addEventListener("keyup", function(event) {
+  query.keyup( function(event) {
     event.preventDefault();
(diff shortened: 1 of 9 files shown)


vikram-redhat:
Thanks @jiri-fiala. In our limited testing (thanks @huffmanca!) this works great. 

The _page_openshift.html.erb should be the same in both branches and that beta notice is a left over after the migration to 4.1. I think once we merge this to master, we can CP it to master-3 (resolve any merge conflicts if they arise).

jboxman:
On my browser, the top of the div is getting clipped by the banner:

Screen Shot 2019-07-01 at 9 57 37 AM

jiri-fiala:
@jboxman, good catch, certain viewport sizes caused the modal to overlap the header, while the header has a fancy `z-index: 9999;`, much more than the modal had set, so I've pushed further on that front with the modal itself. The [preview](http://httpd-test-pro.b9ad.pro-us-east-1.openshiftapps.com/container-platform/3.11/welcome/) is updated.

vikram-redhat:
@jiri-fiala  ok to merge this?

jiri-fiala:
Fine by me.
@wgordon17, as you've reported the scrolling problem initially, could you please make one last sanity check?

wgordon17:
Still working well for me on iOS
LGTM 👍

PR #15682, opened 2019-07-01
Removing prereq to get name of Cluster Logging CR

mburke5678:
Per @ewolinetz in https://github.com/openshift/openshift-docs/pull/15600#discussion_r299093432:

`the clusterlogging CR will always need to be named instance fyi. if it isn't the operator will print out a message and not take action on it.`

The prerequisite to obtain the custom resource name is no longer relevant.


ewolinetz:
lgtm

PR #15698, opened 2019-07-02
bumped latest tags to 3.9.85

sheriff-rh:
Ships 2019-07-02.


vikram-redhat:
The errata is still marked as in push so holding off merging till that is done.

PR #15759, opened 2019-07-08
CNV - QE update for Networking user stories

aburdenthehand:
CNV 2.0 docs are being reviewed for QE. 
This PR resolves the feedback for the Networking user stories.

### DIFF
--- a/cnv/cnv_users_guide/cnv-using-the-default-pod-network-with-cnv.adoc
+++ b/cnv/cnv_users_guide/cnv-using-the-default-pod-network-with-cnv.adoc
@@ -26,6 +26,6 @@ include::modules/cnv-networking-wizard-fields-web.adoc[leveloffset=+2]
 
 == Configuration file examples
 
-include::modules/cnv-template-vmi-pxe-config.adoc[leveloffset=+2]
+include::modules/cnv-template-vm-config.adoc[leveloffset=+2]
 
-include::modules/cnv-template-windows-vmi.adoc[leveloffset=+2]
+nclude::modules/cnv-template-windows-vmi.adoc[leveloffset=+2]
--- a/modules/cnv-configuring-masquerade-mode-cli.adoc
+++ b/modules/cnv-configuring-masquerade-mode-cli.adoc
@@ -14,7 +14,8 @@ editing your virtual machine configuration file.
 
 .Prerequisites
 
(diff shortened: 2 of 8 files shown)


modules/cnv-template-vmi-config-with-masquerade.adoc
```diff
@@ -0,0 +1,46 @@
+// Module included in the following assemblies:
```
yossisegev: This is an invalid spec, you can't create a VMI out of it.
For example - it doesn't contain name field.

yossisegev (changes requested):

yossisegev:
### ***** IMPORTANT *****
We don't use masquerade in CNV 2.0, due to https://bugzilla.redhat.com/show_bug.cgi?id=1725848.
Please replace the use of "masquerade", which appears in the docs themselves and in some VM examples, with "bridge".
Thanks!

modules/cnv-template-vmi-config-with-masquerade.adoc
aburdenthehand: @yossisegev I've updated it. Is that better?

yossisegev: @aburdenthehand trying to create a VM using this spec now fails on another error, at least in my env:
`Error from server (InternalError): error when creating "STDIN": Internal error occurred: failed calling webhook "mutatevirtualmachines.example.com": Post https://kubemacpool-service.kubemacpool-system.svc:443/mutate-virtualmachines?timeout=30s: stream error: stream ID 67; INTERNAL_ERROR`

Did you try deploying this spec on your env? If you did and it succeeded, then it's probably something in my env, and you can use this spec.
@SchSeba can you please take a look? It looks like something related to kubemacpool.
If you want to check it on my env - it's the same cluster which detais I sent you in my previous email (the one about masquerade). Thanks!

aburdenthehand: @yossisegev
Right you are. I was able to test this in a borrowed environment. I've updated it again - this time taking the default config provided by the web console (Create via YAML), modified with the details in the previous example.
Using this config, I was able to create the VM and run it successfully.

yossisegev: @aburdenthehand 
I have just tested the spec in modules/cnv-template-vm-config.adoc, and verified it creates a VM successfully.

PR #15806, opened 2019-07-09
add base documentation for all operator+config resources

bparees:

PR #15816, opened 2019-07-10
Removing references to resources that cannot be configured and are no…

bergerhoffer:
…t needed for viewing

### DIFF
--- a/installing/install_config/customizations.adoc
+++ b/installing/install_config/customizations.adoc
@@ -119,12 +119,6 @@ that are based on many of these CRDs to add more functionality to your
 |Namespaced
 |
 
-|AuthenticationOperatorConfig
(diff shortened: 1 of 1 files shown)


bergerhoffer:
@kalexand-rh These items I'm removing are just in a commented out section, but I wanted to give you a heads up because @deads2k mentions [here](https://github.com/openshift/openshift-docs/pull/15806/files#r302046103) that we shouldn't be documenting these resources. Just in case this section was intended to eventually make it into the docs.

kalexand-rh:
Thanks for letting me know, @bergerhoffer! This section might eventually make it into the docs, so I'm glad that you put in the fix.

kalexand-rh (commented):
LGTM

PR #15818, opened 2019-07-10
samples operator resource is v1, not v1alpha1

bparees:


bparees:
@gabemontero @bmcelvee ptal

should be cherry-picked back to 4.1 also.

bparees:
replaced with https://github.com/openshift/openshift-docs/pull/15821

PR #15840, opened 2019-07-11
Moved VMware import into Create VMs assembly, added QA updates.

apinnick:


vikram-redhat:

modules/cnv-importing-vmware-vm.adoc
```diff
@@ -1,39 +1,51 @@
...
+include::modules/technology-preview.adoc[leveloffset=+1]
+:!FeatureName:
 
 .Prerequisites
 
 * The VMware virtual machine is powered off.
 * The VMware Virtual Disk Development Kit (VDDK) has been uploaded to the namespace in which you are importing the virtual machine.
 
-.Procedure
+include::cnv-uploading-vddk.adoc[leveloffset=+1]
```
bergerhoffer: According to mod docs rules, modules should not include other modules like this. 

There might be another way to handle it, but since it seems like there are 3 procedures in this section, I'd have each of those be their own separate module, which is included in the assembly. And then in the assembly before these module inclusions, you can have some intro if needed. (Or you can put the prereqs in each module, if appropriate).

But since these are procedures, you do need the ".Procedure" heading in front of each of the set of steps.

modules/cnv-importing-vmware-vm.adoc
```diff
@@ -56,10 +68,12 @@ To view the progress of the import:
...
-. In the *Overview* tab, click *ANNOTATIONS* for the *v2vConversionProgress* value. 
+. In the *Overview* tab, click *ANNOTATIONS* for the *v2vConversionProgress* value.
 
 If an error occurs, you can check the conversion pod log:
 
 . Click *Workloads* -> *Pods* from the side menu.
 . Click the conversion pod, *kubevirt-v2v-conversion-_<id>_*, to view the pod details.
 . Click *Logs* to view the conversion pod log.
+
+include::cnv-updating-imported-vm-network-name.adoc[leveloffset=+1]
```
bergerhoffer: Same comment here about not including a module from another module.

modules/cnv-importing-vmware-vm.adoc
```diff
@@ -1,39 +1,51 @@
...
 +
 * If you select *Connect to New Instance*, provide values in the following fields:
 
 ** *vCenter Hostname*
 ** *vCenter User Name*
 ** *vCenter Password*
 +
-Optionally select *Save as New vCenter Instance* and click the 
-*Check* or *Check and Save* button. The wizard connects to the vCenter instance.
+Click *Check and Save*. The wizard connects to the vCenter instance and saves the credentials.
```
bergerhoffer: I think it was already like this, but I think the indentation/lack of bullet is odd here. It's indented like it belongs under the third field of "vCenter Password", when it is actually another actionable step. Seems like it should have a bullet.

modules/cnv-updating-imported-vm-network-name.adoc
```diff
@@ -0,0 +1,31 @@
...
+
+Update the imported virtual machine's NIC name to conform to {ProductName} naming conventions.
+
+. Log in to the imported virtual machine.
+. Go to the */etc/sysconfig/network-scripts* directory.
+. Rename the network configuration file with the new NIC name (a single NIC is called `eth0`):
++
+[options="nowrap" subs="+quotes,verbatim"]
+----
+$ mv _original_NIC_ ifcfg-eth__x__
```
bergerhoffer: Is this "x" supposed to be a replaceable value/number? If so, you might want to make that more apparent. Without seeing the source here, I wouldn't have been able to tell that it was italicized in the built docs.

I'm not sure how it would look here, but we typically do replaceable values as <replaceable_value_name>. I don't know what the appropriate description to use here is, but whatever it is, you could do ```$ mv _original_NIC_ ifcfg-eth<replaceable_value_name>```. And you could also add a callout here to explain, and perhaps add your note above about a single NIC being called `eth0`.

modules/cnv-uploading-vddk.adoc
```diff
@@ -1,11 +1,10 @@
 // Module included in the following assemblies:
 //
-// * cnv/cnv_users_guide/cnv-importing-vm.adoc
+// * modules/cnv-importing-vmware-vm.adoc
+
 [id="cnv-uploading-vddk-{context}"]
 = Uploading the VMware Virtual Disk Development Kit
 
-.Procedure
```
bergerhoffer: This should stay (see above comment)

modules/cnv-updating-imported-vm-network-name.adoc
apinnick: `x` is a replaceable value here. I do not think this requires a callout or `<replaceable_value_name>` because the value is obvious to an admin. It's going to be eth0 or eth1 or eth2, etc., depending on how many NICs the VM has. This is a standard network device naming convention.

modules/cnv-importing-vmware-vm.adoc
apinnick: If you think an option requires a separate bullet, I think this should be turned into a numbered subprocedure. Will change.

modules/cnv-importing-vmware-vm.adoc
apinnick: If I put the intro into the assembly, the heading will also have to be there. OK, if that's the only solution.

PR #15598, opened 2019-06-26
CNV-1617: Clone VM

ousleyp:
Modularizing clone VM/disk procedures for CNV 2.0.

Jira: https://jira.coreos.com/browse/CNV-1617

PR #15865, opened 2019-07-12
Resolving BZ1705663 for master-3 content

bgaydosrh:
Fixing these two minor content error according to BZ1705663: https://bugzilla.redhat.com/show_bug.cgi?id=1705663

This fix should be labeled Enterprise-3.11 and targeted for CNV 1.4.

@openshift/team-documentation please peer review.
Thanks!
Bob


modules/cnv_cloning_pvc_datavolume.adoc
```diff
@@ -40,7 +40,7 @@ spec:
       - ReadWriteOnce
     resources:
       requests:
-storage: 500Mi
+        storage: 2gi
```
ousleyp: The indentation LGTM, but I don't think you needed to change the value to `storage: 2gi`. In 2.0, we still have `storage: 500Mi` for this example (see https://github.com/openshift/openshift-docs/pull/15598/files#diff-e25406d973bae44a7fcc4e45c697f2b6). I think it should be capitalized `Gi` if you're going to change it.

ousleyp:
One comment. Otherwise, LGTM.

PR #13988, opened 2019-03-05
OSDOCS-321-DEVEXP-239 Running entitled builds

bmcelvee:
https://jira.coreos.com/browse/DEVEXP-239
https://jira.coreos.com/browse/OSDOCS-321

Follow-up to https://github.com/openshift/openshift-docs/pull/13819.

PR #15869, opened 2019-07-12
Bug 1722922 - Add running entitled builds to topic map

bmcelvee:
https://bugzilla.redhat.com/show_bug.cgi?id=1722922

Added the content in https://github.com/openshift/openshift-docs/pull/13988, just didn't get it into the topic map.

### DIFF
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -329,6 +329,9 @@ Topics:
 - Name: Performing advanced builds
   File: advanced-build-operations
   Distros: openshift-enterprise,openshift-origin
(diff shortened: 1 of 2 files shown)


adambkaplan (approved):
/lgtm

bmcelvee:
Thanks, @adambkaplan! Since this has your approval, and it's just adding the previously reviewed content to the topic map I don't think it requires QE and peer review. 

I will open a follow-up PR to address the removal of, "Layers are always squashed in OpenShift Container Platform Branch Build.", in 4.2.

PR #15937, opened 2019-07-18
add clarification for OLM operator version syntax

jpeeler:
It must be semantic version compliant.

BZ: 1724742


adellape (approved):
LGTM 👍

PR #15947, opened 2019-07-19
fix mdap typo

neal-timpe:
changed mdap to ldap. I did this already for master and 3.9 and 10 and 11, but I need a separate pr for this older version. @openshift/team-documentation


kalexand-rh:
@neal-timpe, does this change apply to any earlier versions, or just 3.7?

neal-timpe:
No, I just checked them. This is the first version where the typo is introduced.

PR #15956, opened 2019-07-19
CNV-1589: Using Ansible modules with CNV (kubevirt_vm module used as example)

ousleyp:
Jira: https://jira.coreos.com/browse/CNV-1589


modules/cnv-automating-virtual-machine-creation-with-ansible.adoc
```diff
@@ -0,0 +1,98 @@
...
+(...)
+TASK [Create my first VM] ************************************************************************
+changed: [localhost]
+
+PLAY RECAP ********************************************************************************************************
+localhost                  : ok=2    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
+----
+
+. If you did not include `state: running` in your playbook file, you can boot
+the virtual machine by running the playbook again:
```
mmazur: I think it could be made a little more explicit, that booting the virtual machine by running the playbook again is possible only after adding that `state: running` line.

mmazur (approved):

modules/cnv-example-ansible-playbook-creating-vms.adoc
```diff
@@ -0,0 +1,40 @@
...
+You can use the `kubevirt_vm` Ansible Playbook to automate virtual machine
+creation.
+
+The following YAML file is an example of the `kubevirt_vm` playbook. It includes
+sample values that you must replace with your own information if you run the
+playbook.
+
+[source,yaml]
+----
+---
```
huffmanca: Is the `---` necessary, or can the user assume they'll need to include their own stanza?

cnv/cnv_users_guide/cnv-automating-management-tasks.adoc
```diff
@@ -0,0 +1,19 @@
...
+You can automate {ProductName} management tasks by using Red Hat Ansible
+Automation. Learn the basics by using an Ansible Playbook to create a new
+virtual machine.
+
+include::modules/cnv-about-red-hat-ansible-automation.adoc[leveloffset=+1]
+
+include::modules/cnv-automating-virtual-machine-creation-with-ansible.adoc[leveloffset=+1]
+
+To verify that the virtual machine was created, try to
+xref:../../cnv/cnv_users_guide/cnv-accessing-vm-consoles.adoc#cnv-accessing-vm-consoles[access the VM console].
```
huffmanca: It seems that the `#cnv-accessing-vm-consoles` anchor doesn't exist on this page. Did you mean `#cnv-accessing-vm-consoles-cli` ?

huffmanca (commented):
Honestly this looks great! I have a couple of minor questions, but it could likely be merged as is.

Thank you!

modules/cnv-example-ansible-playbook-creating-vms.adoc
ousleyp: IIUC, every Ansible Playbook starts with `---`, and this example is of the complete playbook (unlike the other example which is just a snippet).

mmazur: Yes, a whole playbook needs to start with the `---`.

modules/cnv-automating-virtual-machine-creation-with-ansible.adoc
ousleyp: Hey - the source material says "Changing this behavior requires specifying state: running as one of the module’s parameters when creating a new VM. **Or we can get vm1 to boot by running the first playbook one more time**, since this time the task will be interpreted as attempting to change the state of an existing VM to running, which is what we want." 

I interpreted this to mean that the playbook could remain unchanged and it would still start the VM. Am I misreading it? Thanks for your help, @mmazur

cnv/cnv_users_guide/cnv-automating-management-tasks.adoc
ousleyp: @huffmanca I actually meant to link to the top of the page (the assembly :context:). Is that okay as-is? If not, I'm happy to change it.

modules/cnv-automating-virtual-machine-creation-with-ansible.adoc
mmazur: If you take a look at the [examples for that blog post](https://github.com/kubevirt/ansible-kubevirt-modules/tree/master/examples/blog/part1):

1. the first one basically just does `state: running`,
2. while the second one actually defines the vm.

And when I wrote that _we can get vm1 to boot by running the **first** playbook one more time_, that means the one with `state: running`. Re–running the **second** example (the one with the vm spec) will do nothing, unless you've added `state: running` to it first.

So with your text describing only a single playbook, it needs to be made explicitly clear, that to make the vm actually boot, the playbook being run needs to contain `state: running`. As in, something like this:

_6. If you did not include `state: running` in your playbook file, you can boot the virtual machine by adding that line to the playbook file and running the playbook again:_

ousleyp: Ah, thanks so much for clarifying! That makes much more sense. I will make that change shortly.

PR #15959, opened 2019-07-21
bz1724028 remove callout ref

bobfuru:
Removed startingCSV callout 3 listing. (Previous PR removed callout 3 and marker, but not ref listing. Also note that this issue was fixed in previous PR for 4.2 but not 4.1, which is why this PR is being submitted.)

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1724028

@openshift/team-documentation


bergerhoffer:
This LGTM!

FYI @vikram-redhat I'm removing the "enterprise-4.2" label since @bobfuru says that this change has already been applied to that branch. I confirmed that it's only master and enterprise-4.1 that need to remove this callout.

PR #15963, opened 2019-07-22
CNV - Importing VMs structural change

aburdenthehand:
Moving the importing procedures to their own assemblies for consistency. 
I've included the wizard fields reference tables at the end of the assembly, even though they are relevant specifically for the *Importing the VMware virtual machine or template* procedure. The concern was that the reference material then overshadowed the next procedure. To overcome this I've included an xref that points to the reference tables. 
It is an imperfect solution, however it is only temporary since this procedure will become more user-friendly.

Not a part of this PR but the importing procedures should be moved so they follow 'Creating VM'.
@vikram-redhat - this will involve a change to the_topic_map WIP

Build link: http://file.bne.redhat.com/aburden/CNV-docs-preview/cnv-importing-vmware-assembly/cnv/cnv_users_guide/cnv-importing-vmware-vm.html

### DIFF
--- a/cnv/cnv_users_guide/cnv-create-vms.adoc
+++ b/cnv/cnv_users_guide/cnv-create-vms.adoc
@@ -29,22 +29,3 @@ API Reference] for a definitive list of virtual machine settings.
 
 include::modules/cnv-vm-storage-volume-types.adoc[leveloffset=+1]
 
-[id="cnv-importing-vmware-vm_{context}"]
-== Importing a VMware virtual machine or template with the virtual machine wizard
-
-You can import a VMware virtual machine or template. If you import a template, a virtual machine based on the template is created.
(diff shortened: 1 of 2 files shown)


apinnick:
@aburdenthehand 

No problem. The reference tables are also relevant for creating a VM with the web console, but I have no problems with moving the VM import modules to their own assembly if you don't mind the fact that all that reference chapters are now repeated in the VM Import assembly. I had originally moved the VM Import into the Create VMs assembly to avoid having to repeat the reference modules.

kalexand-rh:
@aburdenthehand, rebase and this seems fine.

aburdenthehand:
@kalexand-rh and @apinnick - Thanks for taking a look. Rebased.

PR #15967, opened 2019-07-22
Kubernetes Feature Flag doc is confusing and incorrect

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1722240

This doc confuses me. It has 1 more feature listed than we manage (it lists MachineHealthCheck, which I believe we enable in 4.2, not 4.1.) It also doesn't really distinguish the nuances. By default we enable any Kubernetes feature gates which are beta. We disable any kubenretes feature gates which are for alpha features. By default we enable, by default a number of feature gates which are not beta/ga. Those include:
ExperimentalCriticalPodAnnotation
RotateKubeletServerCertificate
SupportPodPidsLimit

There is 1 feature gate which is enabled by default in kube but which we explicitly disable:
LocalStorageCapacityIsolation

If a user were to enable the 'tech preview no upgrade' features they will also get, but will likely lose the ability to upgrade that cluter:
CSIBlockVolume

I think the doc needs to be rewritten to explain those 3 categories of feature flags and to make it clear that setting tech preview no upgrade only affects the singe additional feature gate.

### DIFF
--- a/modules/feature-gate-features.adoc
+++ b/modules/feature-gate-features.adoc
@@ -5,25 +5,39 @@
 [id="feature-gate-features_{context}"]
 = Features that are affected by FeatureGates
 
-The following features are affected by FeatureGates:
+The following Technology Preview features included in {product-title}:
 
 [options="header"]
 |===
-| FeatureGate| Description
-
-|`CSIBlockVolume`
-|Enables external CSI drivers to implement raw block volume support.
+| FeatureGate| Description| Default
 
 |`ExperimentalCriticalPodAnnotation`
 |Enables annotating specific Pods as critical so that their scheduling is guaranteed.
-
-|`MachineHealthCheck`
-|Enables automatically repairing damaged machines in a machine pool.
+|True
 
 |`RotateKubeletServerCertificate`
 |Enables the rotation of the server TLS certificate on the cluster.
+|True
 
 |`SupportPodPidsLimit`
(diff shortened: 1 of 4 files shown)


mburke5678:
@deads2k PTAL

@eparis requested some changes to the Feature Gates doc for 4.1, which I will adapt for 4.2/

deads2k:
I think you need to diff against https://github.com/openshift/api/blob/release-4.1/config/v1/types_feature.go#L68-L90

mburke5678:
@deads2k Does this suggest that `MachineHealthCheck` is not in 4.1 or is that feature not controlled by a feature gate?

mburke5678:
@deads2k PTAL @ https://github.com/openshift/openshift-docs/pull/15967#issuecomment-514608476

mburke5678:
@xltian Do you know if MachineHealthChecks is in 4.1? I was told that the feature is in 4.1 and turned on with a feature gate. But David Eads pointed at some code that seems to suggest it is not. Thank you!!

jhou1:
MachineHealthCheck is in 4.1 and is disabled by default. It could be turned on with feature gate.

mburke5678:
@jhou1 Thank you for confirming. Would you be able to review the changes in this PR?

jhou1:
@mburke5678 it looks good!

mburke5678:
Feature Gates updates for 4.2: https://github.com/openshift/openshift-docs/pull/16159

PR #15974, opened 2019-07-22
Document Issues with swapping log collectors after deployment

mburke5678:
We are assembling a list of differences between Fluentd and Rsyslog as log collectors and the ramifications of switching between the two.

https://jira.coreos.com/browse/OSDOCS-624
https://jira.coreos.com/browse/LOG-20

### DIFF
--- a/modules/efk-logging-collector-fluentd-v-rsyslog.adoc
+++ b/modules/efk-logging-collector-fluentd-v-rsyslog.adoc
@@ -19,4 +19,8 @@ Some of the noteable differences between the two collectors are:
 
 * remote syslog
 
+When switching between the two collectors can result in:
 
+* A gap in messages between scale down of one collector and scale up of the other when reading from the tail.
+
+* Duplicate messages between scale down of one collector and scale up of the other when reading from the head.
--- a/modules/efk-logging-collector-fluentd-v-rsyslog.adoc
+++ b/modules/efk-logging-collector-fluentd-v-rsyslog.adoc
@@ -24,3 +24,13 @@ When switching between the two collectors can result in:
 * A gap in messages between scale down of one collector and scale up of the other when reading from the tail.
 
 * Duplicate messages between scale down of one collector and scale up of the other when reading from the head.
++
+By default, Rsyslog reads from the head of the container log files, and from the start of the journal. Switching to Rsyslog 
+will lead to many duplicated messages, as well as Rsyslog taking a long time to record the most current log entries.
++
+You can configure Rsyslog to skip older entries in the journal and start reading from the end by setting the `RSYSLOG_JOURNAL_READ_FROM_TAIL`
+parameter in the Rsyslog daemonset:
++
+----
+oc set env ds/rsyslog RSYSLOG_JOURNAL_READ_FROM_TAIL=true
+----
--- a/modules/efk-logging-collector-fluentd-v-rsyslog.adoc
+++ b/modules/efk-logging-collector-fluentd-v-rsyslog.adoc
@@ -5,7 +5,12 @@
 [id="efk-logging-collector-fluentd-v-rsyslog_{context}"]
 = About Fluentd and Rsyslog
 
-You can choose between Fluentd and Rsyslog as the logging collector. 
+In {product-title} {product-version}, you can choose between Fluentd and Rsyslog as the logging collector. 
+
+[NOTE]
+====
+We are currently evaluating alternative log collectors in order to offer the best collection tool for cluster logging.  
+====
 
 Some of the noteable differences between the two collectors are:
 
@@ -27,7 +32,7 @@ You can work around this issue using one of the following methods:
 ----
 oc set env ds/rsyslog RSYSLOG_K8S_CACHE_ENTRY_TTL=60
 ----
-
+go
 When switching between the two collectors can result in:
 
 * A gap in messages between scale down of one collector and scale up of the other when reading from the tail.
--- a/modules/efk-logging-collector-fluentd-v-rsyslog.adoc
+++ b/modules/efk-logging-collector-fluentd-v-rsyslog.adoc
@@ -17,7 +17,16 @@ For Rsyslog, you can edit the Rsyslog configmap to add support for Syslog log fo
 
 * The Event Router is not supported for the Rsyslog log collector.
 
-* remote syslog
+* Due to a known Rsyslog issue, if you delete a project and create a new project using the same name before Rsyslog removes the cache entry, Rsyslog sends logs to the exsiting index. 
++
+You can work around this issue using one of the following methods:
++
+** Restart Rsyslog after deleting the namespace and before creating another namespace with the same name.
+** Adjust the cache settings for link:https://www.rsyslog.com/doc/v8-stable/configuration/modules/mmkubernetes.html#cacheexpireinterval[cache expiration] and link:https://www.rsyslog.com/doc/v8-stable/configuration/modules/mmkubernetes.html#cacheentryttl[cache entry TTL] so that cache entries are deleted more frequently.  By default, TTL is 1 hour. For example, use the following command to set the TTL to 1 minute: 
++
+----
+oc set env ds/rsyslog RSYSLOG_K8S_CACHE_ENTRY_TTL=60
+----
 
 When switching between the two collectors can result in:
 
--- a/modules/efk-logging-collector-fluentd-v-rsyslog.adoc
+++ b/modules/efk-logging-collector-fluentd-v-rsyslog.adoc
@@ -9,9 +9,9 @@ You can choose between Fluentd and Rsyslog as the logging collector.
 
 Some of the noteable differences between the two collectors are:
 
-* No throttling in Rsyslog. With Fluentd, an administrator can reduce the rate at which the logs are read in by Fluentd before being processed. This feature is not currently available with Rsyslog.
+* No throttling in Rsyslog. With Fluentd, an administrator can reduce the rate at which the logs are read in by Fluentd before being processed. Log throttling is not currently available with Rsyslog.
 
-* No _built-in_ support for remote syslog forwarding in rsyslog. With Fluentd, you can use the `secure-forward` plug-in to send a copy of its logs to an external log aggregator. This feature is not currently available with Rsyslog.
+* No _built-in_ support for remote syslog forwarding in rsyslog. With Fluentd, you can use the `secure-forward` plug-in to send a copy of its logs to an external log aggregator. Remote syslog forwarding is not currently available with Rsyslog.
 +
 For Rsyslog, you can edit the Rsyslog configmap to add support for Syslog log forwarding using the *omfwd* module, see link:https://www.rsyslog.com/doc/v8-stable/configuration/modules/omfwd.html[]. To send logs to a different Rsyslog instance, you can the *omrelp* module, see link:https://www.rsyslog.com/doc/v8-stable/configuration/modules/omrelp.html[]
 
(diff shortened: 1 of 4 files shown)


mburke5678:
@nhosoi @richm I have started a document to capture the appropriate details about the various functionality differences and associated ramifications to switching, based on the scrum discussion this morning. Please feel free to add any thoughts here so we can get this blocking issue moving.

mburke5678:
https://jira.coreos.com/browse/LOG-173

modules/efk-logging-collector-fluentd-v-rsyslog.adoc
```diff
@@ -0,0 +1,26 @@
...
+// * logging/efk-logging-fluentd.adoc
+
+[id="efk-logging-collector-fluentd-v-rsyslog_{context}"]
+= About Fluentd and Rsyslog
+
+You can choose between Fluentd and Rsyslog as the logging collector. 
+
+Some of the noteable differences between the two collectors are:
+
+* Daching differences.
```
richm: Caching

modules/efk-logging-collector-fluentd-v-rsyslog.adoc
```diff
@@ -0,0 +1,26 @@
...
+
+* Deleted namespace
```
richm: I think these two items should be deleted.  The rsyslog Kubernetes metadata caching now, by default, works exactly like the fluentd Kubernetes metadata caching, and the parameters should be documented elsewhere (if any, if we have documented the fluentd Kubernetes metadata cache parameters).

modules/efk-logging-collector-fluentd-v-rsyslog.adoc
```diff
@@ -0,0 +1,26 @@
...
+
+* No throttling in Rsyslog. With Fluentd, an administrator can reduce the rate at which the logs are read in by Fluentd before being processed. This feature is not currently available with Rsyslog.
+
+* No secure forward in Rsyslog. With Fluentd, you can use the Fluentd the `secure-forward` plug-in to send a copy of its logs to an external log aggregator. This feature is not currently available with Rsyslog.
+
+* remote syslog
```
richm: Although we do not have "built-in" support for remote syslog forwarding in rsyslog (like we have in fluentd via the `REMOTE_SYSLOG_*` env. var. settings), users can edit the rsyslog configmap to add support for syslog log forwarding using the `omfwd` module (https://www.rsyslog.com/doc/v8-stable/configuration/modules/omfwd.html).  If sending logs to another rsyslog, users can also use the better `omrelp` module (https://www.rsyslog.com/doc/v8-stable/configuration/modules/omrelp.html)

modules/efk-logging-collector-fluentd-v-rsyslog.adoc
```diff
@@ -0,0 +1,26 @@
...
+
+When switching between the two collectors can result in:
+
+* A gap in messages between scale down of one collector and scale up of the other when reading from the tail.
+
+* Duplicate messages between scale down of one collector and scale up of the other when reading from the head.
```
richm: By default, rsyslog reads from the head of the container log files, and from the start of the journal, which will lead to many duplicated messages, as well as rsyslog taking a long time to "catch up" to the most current log entries.

Not sure if we should document it here, but probably in the same place where we document the comparable fluentd parameters but the env. var. `RSYSLOG_JOURNAL_READ_FROM_TAIL` (https://github.com/openshift/cluster-logging-operator/blob/master/files/rsyslog/10-viaq-modules.conf#L7) can be set `RSYSLOG_JOURNAL_READ_FROM_TAIL=true` to make rsyslog skip older entries in the journal and start reading from the end (which will lead to record loss).  Similarly for the container log files, use `RSYSLOG_FILE_READ_FROM_TAIL=true` (https://github.com/openshift/cluster-logging-operator/blob/master/files/rsyslog/60-mmk8s.conf#L11) to skip older container log records and start reading from the end of the log files (which will lead to record loss).

modules/efk-logging-collector-fluentd-v-rsyslog.adoc
mburke5678: @richm 
Can we define what the caching differences are?

modules/efk-logging-collector-fluentd-v-rsyslog.adoc
```diff
@@ -9,13 +9,13 @@ You can choose between Fluentd and Rsyslog as the logging collector.
...
 Some of the noteable differences between the two collectors are:
 
-* Daching differences.
-
-* Deleted namespace
-
 * No throttling in Rsyslog. With Fluentd, an administrator can reduce the rate at which the logs are read in by Fluentd before being processed. This feature is not currently available with Rsyslog.
 
-* No secure forward in Rsyslog. With Fluentd, you can use the Fluentd the `secure-forward` plug-in to send a copy of its logs to an external log aggregator. This feature is not currently available with Rsyslog.
+* No _built-in_ support for remote syslog forwarding in rsyslog. With Fluentd, you can use the Fluentd the `secure-forward` plug-in to send a copy of its logs to an external log aggregator. This feature is not currently available with Rsyslog.
```
richm: change `you can use the Fluentd the` to `you can use the`

modules/efk-logging-collector-fluentd-v-rsyslog.adoc
mburke5678: We don't have anywhere we list the env vars that can be set. We have a topic that links to https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/README.md. 
I can add a link to https://github.com/openshift/origin-aggregated-logging/blob/master/rsyslog/README.md. 

For this issue ^^ I can add a line in this doc that describes the READ_FROM_TAIL variable.
```
You can configure Rsyslog to skip older entries in the journal and start reading from the end by setting the `RSYSLOG_JOURNAL_READ_FROM_TAIL`
parameter in the Rsyslog daemonset:
+
----
oc set env ds/rsyslog RSYSLOG_JOURNAL_READ_FROM_TAIL=true
----
```
Seems like the place where you would want to know about this parameter??

richm: ok

richm:
@QiaolingTang ptal

modules/efk-logging-collector-fluentd-v-rsyslog.adoc
```diff
@@ -0,0 +1,36 @@
...
+
+Some of the noteable differences between the two collectors are:
+
+* No throttling in Rsyslog. With Fluentd, an administrator can reduce the rate at which the logs are read in by Fluentd before being processed. This feature is not currently available with Rsyslog.
+
+* No _built-in_ support for remote syslog forwarding in rsyslog. With Fluentd, you can use the `secure-forward` plug-in to send a copy of its logs to an external log aggregator. This feature is not currently available with Rsyslog.
++
+For Rsyslog, you can edit the Rsyslog configmap to add support for Syslog log forwarding using the *omfwd* module, see link:https://www.rsyslog.com/doc/v8-stable/configuration/modules/omfwd.html[]. To send logs to a different Rsyslog instance, you can the *omrelp* module, see link:https://www.rsyslog.com/doc/v8-stable/configuration/modules/omrelp.html[]
+
+* The Event Router is not supported for the Rsyslog log collector.
```
QiaolingTang: I thought there is no relationship between event router and log collectors. Why do you document `The Event Router is not supported for the Rsyslog log collector.` here?
I've tested in my env before, and the event logs collected by event router all could be parsed correctly by rsyslog.

modules/efk-logging-collector-fluentd-v-rsyslog.adoc
```diff
@@ -0,0 +1,36 @@
...
+
+* remote syslog
+
+When switching between the two collectors can result in:
+
+* A gap in messages between scale down of one collector and scale up of the other when reading from the tail.
+
+* Duplicate messages between scale down of one collector and scale up of the other when reading from the head.
++
+By default, Rsyslog reads from the head of the container log files, and from the start of the journal. Switching to Rsyslog
```
QiaolingTang: Seems only documented about the issues and solutions when changing from fluentd to rsyslog, no information about changing from rsyslog to fluentd.

modules/efk-logging-collector-fluentd-v-rsyslog.adoc
richm: Are event router records formatted in exactly the same way with fluentd as with rsyslog?  I think not.  I think that's what the problem is - rsyslog will collect those logs, but rsyslog will not format the logs in the same way as fluentd.

modules/efk-logging-collector-fluentd-v-rsyslog.adoc
richm: You will have the same problems going from rsyslog to fluentd.  Either you configure fluentd to read from the beginning of the journald and log files, which means a lot of duplicate records, or you configure fluentd to read from the tail of the journal and log files, which means missing records.

modules/efk-logging-collector-fluentd-v-rsyslog.adoc
QiaolingTang: I'm not sure. Since there is a bug about rsyslog parsing logs, I'm not able to check it now. I'll update my test result here once I could. Thank you.

modules/efk-logging-collector-fluentd-v-rsyslog.adoc
mburke5678: @richm Is there a way in Openshift to set whether Fluentd reads from head or tail? I see there are Fluentd plugins to do this. 
In the past, it seems we had a `JOURNAL_READ_FROM_TAIL` environment variable to configure this.

richm: For container logs, use `CONTAINER_READ_FROM_HEAD=false` to read from the tail - by default it will read from the head.  For the journal, use `JOURNAL_READ_FROM_HEAD=true` to read from the head - by default it will read from the tail.

modules/efk-logging-collector-fluentd-v-rsyslog.adoc
QiaolingTang: @mburke5678  Yes.

@richm  I compared the event logs collected by fluentd and rsyslog,  here is the logs: http://pastebin.test.redhat.com/786553.
If ignore the order of some objects, seems no difference.

richm: Then it looks like fluentd is broken - it doesn't seem to be processing the events correctly - it should "move" the events underneath the top level `kubernetes` field - https://github.com/ViaQ/fluent-plugin-viaq_data_model/blob/master/lib/fluent/plugin/filter_viaq_data_model.rb#L452 - `@process_kubernetes_events` is `true` by default.  Not sure what's going on here . . .

QiaolingTang: Do you think this is a bug? I checked with 4.1 and 3.11:https://privatebin-it-iso.int.open.paas.redhat.com/?27170295f76a88ac#ccyZoBERqvkyPB+WZkPEBXbMKIgZLjwfYhb8aHt5y/g=, the event logs collected by fluentd are parsed to the same format as it in 4.2.

richm: Yes - it is a bug

mburke5678:
@sichvoge requested information that we are evaluating using one log collector and the ability to switch log collectors might be removed at some point. He promised to speak with marketing to get the messaging right,
https://coreos.slack.com/archives/GGUR75P60/p1566048251302300?thread_ts=1565977930.296100&cid=GGUR75P60

vikram-redhat:
@mburke5678 can this be moved along?

mburke5678:
@vikram-redhat This PR would only be needed if we add Rsyslog as a log collector. It is not clear if this will ever happen. Please dispose of as you see fit.

richm:
I recommend getting rid of any/all references to rsyslog

vikram-redhat:
@mburke5678 in that case, I am closing this PR.

PR #15975, opened 2019-07-22
Replace Elasticsearch Operator install

mburke5678:
In 4.1, the Elasticsearch Operator could not be installed via the web console. In 4.2 that functionality has returned and the user can use the web console to install the EO.


mburke5678:
@anpingli PTAL.

anpingli (approved):
LGTM

PR #15988, opened 2019-07-23
Fixing typo in bare metal installation

jan-zmeskal:
Removing redundant word from installation guide.


kalexand-rh:
Thanks @jan-zmeskal!

PR #15999, opened 2019-07-23
Add network customization for vSphere, GCP, bare-metal installs

jboxman:
So, it's possible to configure the CNO for all installation types, not just AWS.

@kalexand-rh, what are you thoughts on placement for this content? Ignition configuration happens prior to creating manifests, so I placed the modules for configuring the CNO by editing manifests after. Unlike the AWS install, I notice for vSphere the example file doesn't include any networking parameters; Is that not configurable for vSphere?

This probably needs deeper investigation but I wanted to start this.

Thanks!

### DIFF
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -110,6 +110,8 @@ Topics:
     File: installing-gcp-default
   - Name: Installing a cluster on GCP with customizations
     File: installing-gcp-customizations
+  - Name: Installing a cluster on GCP with network customizations
+    File: installing-gcp-network-customizations
   - Name: Uninstalling a cluster on GCP
     File: uninstalling-cluster-gcp
 - Name: Installing on user-provisioned AWS
@@ -127,6 +129,8 @@ Topics:
   Topics:
   - Name: Installing a cluster on bare metal
     File: installing-bare-metal
+  - Name: Installing a cluster on bare metal with network customizations
+    File: installing-bare-metal-network-customizations
 - Name: Installing on OpenStack
   Dir: installing_openstack
   Topics:
@@ -145,6 +149,8 @@ Topics:
   Topics:
   - Name: Installing a cluster on vSphere
     File: installing-vsphere
+  - Name: Installing a cluster on vSphere with network customizations
+    File: installing-vsphere-network-customizations
 - Name: Installing in restricted networks
   Dir: installing_restricted_networks
   Topics:
--- a/installing/installing_bare_metal/installing-bare-metal-network-customizations.adoc
+++ b/installing/installing_bare_metal/installing-bare-metal-network-customizations.adoc
@@ -1,0 +1,90 @@
+[id="installing-bare-metal-network-customizations"]
+= Installing a cluster on bare metal with network customizations
+include::modules/common-attributes.adoc[]
+:context: installing-bare-metal-network-customizations
+
+toc::[]
+
+In {product-title} version {product-version}, you can install a cluster on bare
+metal infrastructure that you provision with customized network configuration
+options. By customizing your network configuration, your cluster can coexist
+with existing IP address allocations in your environment and integrate with
+existing MTU and VXLAN configurations.
+
+You must set most of the network configuration parameters during installation,
+and you can modify only `kubeProxy` configuration parameters in a running
+cluster.
+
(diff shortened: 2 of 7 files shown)


kalexand-rh:
@jboxman, that sounds reasonable, and I can take a closer look when you're closer to merging.

@dav1x, do you have opinions about adding networking parameters for vSphere?
> Unlike the AWS install, I notice for vSphere the example file doesn't include any networking parameters; Is that not configurable for vSphere?

jboxman:
@kalexand-rh, I've updated this. The big outstanding question is what `networking` parameters, if any, are available in the `install-config.yaml` for bare-metal and vSphere respectively. Based on that, I'll need to customize or not include the networking parameters table in each assembly.

kalexand-rh:
@jboxman, all of the parameters that I am aware of are in the `install-config.yaml` examples already. You might want to check with your devs.

jboxman:
As near as I can tell, the `Networking` parameter is always acceptable; where or not it has any effect for a given platform I don't know so far.

https://github.com/openshift/installer/blob/4edffaead4e6f4d4e539daa1cf7ea3444fe69c49/pkg/types/installconfig.go#L56

jboxman:
@kalexand-rh, I've updated the PR and fixed the build error. The previews are here:

- [bare-metal](https://inst-nw-customization-vsphere-baremetal--ocpdocs.netlify.com/openshift-enterprise/latest/installing/installing_bare_metal/installing-bare-metal-network-customizations.html)
- [vSphere](https://inst-nw-customization-vsphere-baremetal--ocpdocs.netlify.com/openshift-enterprise/latest/installing/installing_vsphere/installing-vsphere-network-customizations.html)

kalexand-rh:
@jboxman, it seems reasonable to me and is consistent with our other approach to this configuration. @abhinavdahiya, is there any other context that you think might be helpful in helping users know when they'd want to use the network operator in a bare metal or vSphere installation instead of using the standard options?

jboxman:
@kalexand-rh, what's the next step for this PR? Thanks!

kalexand-rh:
@jboxman, I'd probably push for a deeper dev review again before sending it to QE for verification.

kalexand-rh (commented):
It LGTM, pending other reviews.

jboxman:
@kalexand-rh, not sure who to ask from engineering as this cuts across concerns.

@squeed @danwinship - The CNO can be customized during install for all supported platforms  in 4.2, yes?

Thanks.

squeed:
Correct - the CNO is platform independent. If possible, it would be nice to see it removed from the platform-specific documentation.

danwinship:
Minor clarification: `networking.clusterNetwork`, `networking.serviceNetwork`, and `networking.networkType` are CNO configuration and are platform-independent (although `Kuryr` is only a valid value for `networkType` when using OpenStack). `networking.machineCIDR` is not CNO config, and is not allowed/effective on some platforms.

dav1x:
@kalexand-rh Sorry for the late response. I think the nsx-t plugin would be the SDN plugin specific to vSphere. But, from my understanding they are working on an operator for installation and configuration of NSX-T.  I'm not really an NSX expert unfortunately.

jboxman:
@anuragthehatter, this PR adds installation network customization for GCP, vSphere, and bare metal. Can you confirm whether the content is accurate or suggest someone who can?

Thanks!

Previews:

- [GCP](http://file.bos.redhat.com/jboxman/20191106/installing/installing_gcp/installing-gcp-network-customizations.html)
- [vSphere](http://file.bos.redhat.com/jboxman/20191106/installing/installing_vsphere/installing-vsphere-network-customizations.html)
- [Bare metal](http://file.bos.redhat.com/jboxman/20191106/installing/installing_bare_metal/installing-bare-metal-network-customizations.html)

jboxman:
@kalexand-rh, latest iteration of this, which adds GCP. Once this is merged, I can remove the conditionals and TP notice in a 4.3-only PR for OVN GA in 4.3 on (hopefully) all platforms.

anuragthehatter:
Looks good to me. Thanks for incorporating it.
/lgtm

jboxman:
@anuragthehatter, thanks!

PR #16005, opened 2019-07-24
Revert "Revert "Replace Elasticsearch Operator install""

mburke5678:
Reverts openshift/openshift-docs#15993

I reverted https://github.com/openshift/openshift-docs/pull/15975 in error via openshift/openshift-docs#15993 due to versioning confusion. This is to revert the revert.


mburke5678:
@vikram-redhat PTAL

PR #16012, opened 2019-07-24
Some title name fixes for SM and CNV

vikram-redhat:
@JStickler @ousleyp @geekspertise @aburdenthehand @bgaydosrh 

Doing a brute force fix to this as it seems the customer portal doesn't like the attribute in the Header.

### DIFF
--- a/cnv/cnv_install/cnv-about-cnv.adoc
+++ b/cnv/cnv_install/cnv-about-cnv.adoc
@@ -1,5 +1,5 @@
 [id="cnv-about-cnv"]
-= About {ProductName}
+= About Container-native Virtualization
 include::modules/cnv-document-attributes.adoc[]
 :context: cnv-about-cnv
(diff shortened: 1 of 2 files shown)


ousleyp:
Hey @vikram-redhat,

Is including variables in headers still problematic or has the issue been fixed? 

Context: I realized that the PR I just had merged included a couple of files with {CNVProductName} in the heading/title. I opened a small PR to fix it after vaguely remembering *this* PR, but then I did a quick search and found some other instances (both assemblies and modules).

I haven't personally noticed any issues on the Portal. For example, [this assembly](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.1/html/container-native_virtualization/container-native-virtualization-installation#preparing-cluster-for-cnv) has both `{product-title}` and `{CNVProductName}` in the header. Am I missing something?

Thanks!

PR #16017, opened 2019-07-24
4.2 installation stubs

kalexand-rh:
I don't want to fight merge conflicts in the installation TOC, so I created stubs for the stories that are associated with the Azure, GCP, and disconnected epics and added them to the TOC. I'll uncomment them as I complete drafts of the assemblies.

@openshift/team-documentation, any objections?


kalexand-rh:
I'm taking the +1s from @mburke5678 and @vikram-redhat as blessing. Merging.

PR #16028, opened 2019-07-25
support improved undefined field handling

mburke5678:
https://jira.coreos.com/browse/LOG-372
https://jira.coreos.com/browse/OSDOCS-646

### DIFF
--- a/modules/efk-logging-fluentd-undefined.adoc
+++ b/modules/efk-logging-fluentd-undefined.adoc
@@ -1,0 +1,145 @@
+// Module included in the following assemblies:
+//
+// * logging/efk-logging-fluentd.adoc
+
+[id="efk-logging-fluentd-undefined_{context}"]
+= Configuring how the log collector normalizes logs 
+
+Cluster Logging uses a specific data model, like a database schema, to store log records and their metadata in the logging store. There are some restrictions on the data:
+
+* There must be a `"message"` field containing the actual log message.
+* There must be a `"@timestamp"` field containing the log record timestamp in RFC 3339 format, preferably millisecond or better resolution.
+* There must be a `"level"` field with the log level, such as `err`, `info`, `unknown`, and so forth.
+
+[NOTE]
+====
+For more information on the data model, see link:https://github.com/openshift/origin-aggregated-logging/blob/master/docs/com.redhat.viaq-openshift-project.asciidoc[Exported Fields]. 
+====
+
+Because of these requirements, conflicts and inconsistencies can arise with log data collected from different subsystems.  
+
+For example, if you use the `MERGE_JSON_LOG` feature (`MERGE_JSON_LOG=true`), it can be extremely useful to have your applications log their output in JSON, and have the log collector automatically parse and index the data in Elasticsearch. However, this leads to several problems, including:
+
+* field names can be empty, or contain characters that are illegal in Elasticsearch; 
+* different applications in the same namespace might output the same field name with different value data types; 
+* applications might emit too many fields; 
+* fields may conflict with the cluster logging built-in fields.
+
+You can configure how cluster logging treats fields from disparate sources by editing the log collector daemonset, Fluentd or Rsyslog, and setting environment variables in the table below. 
+
+// from https://github.com/ViaQ/fluent-plugin-viaq_data_model/commit/8b5ef11cedec4c372b2cb082afc7f9cc08473654
+
+* Undefined fields. One of the problems with log data from disparate systems is that some fields might be unknown to the ViaQ data model.
+Such fields are called _undefined_. ViaQ requires all top-level fields to be defined and described. 
++
+Use the parameters to configure how {product-title} moves any undefined fields under a top-level field called `undefined` to avoid conflicting with the _well known_ ViaQ 
+top-level fields. You can add undefined fields to the top-level fields and move others to an `undefined` container.
++
+You can also replace special characters in undefined fields and convert undefined fields to their JSON string representation. 
+Coverting to JSON string preserves the structure of the value, so that you can retrieve the value later and convert it back to a map or an array.
++
+** Simple scalar values like numbers and booleans are changed to a quoted string. For example: `10` becomes `"10"`, `3.1415` becomes `"3.1415"``, `false` becomes `"false"`.
++
+** Map/dict values and array values are converted to their JSON string representation: `"mapfield":{"key":"value"}` becomes `"mapfield":"{\"key\":\"value\"}"` and `"arrayfield":[1,2,"three"]` becomes `"arrayfield":"[1,2,\"three\"]"`.
+
+* Defined fields. You can also configure which defined fields appear in the top levels of the logs. 
++
+[[default-fields]]
+The default top-level fields, defined through the `CDM_DEFAULT_KEEP_FIELDS` parameter, are `CEE`, `time`, `@timestamp`, `aushape`, `ci_job`, `collectd`, `docker`, `fedora-ci`, 
+`file`, `foreman`, `geoip`, `hostname`, `ipaddr4`, `ipaddr6`, `kubernetes`, `level`, `message`, `namespace_name`, `namespace_uuid`,
+`offset`, `openstack`, `ovirt`, `pid`, `pipeline_metadata`, `rsyslog`, `service`, `systemd`, `tags`, `testcase`, `tlog`, `viaq_msg_id`.
++
+Any fields not included in `${CDM_DEFAULT_KEEP_FIELDS}` or `${CDM_EXTRA_KEEP_FIELDS}` are moved to `${CDM_UNDEFINED_NAME}` if `CDM_USE_UNDEFINED` is `true`.
++
+[NOTE]
+====
+The `CDM_DEFAULT_KEEP_FIELDS` parameter is for only advanced users, or if you are instructed to do so by Red Hat support.
+==== 
+
+* Empty fields. You can determine which empty fields to retain from disparate logs.
+
+.Environment parameters for log normalization
+[cols="2,4a,2", options="header"]
+|===
+|Parameters  |Definition  |Example
+
+|`CDM_EXTRA_KEEP_FIELDS`
+|Specify an extra set of defined fields to be kept at the top level of the logs in addition to the `CDM_DEFAULT_KEEP_FIELDS`. The default is "".
+|`CDM_EXTRA_KEEP_FIELDS="broker"`
+
+|`CDM_KEEP_EMPTY_FIELDS`
+|Specify fields to retain even if empty in CSV format. Empty defined fields not specified are dropped. The default is "message", keep empty messages.
+|`CDM_KEEP_EMPTY_FIELDS="message"`
+
+|`CDM_USE_UNDEFINED`
+|Set to `true` to move undefined fields to the `undefined` top level field. The default is `false`. If `true`, values in `CDM_DEFAULT_KEEP_FIELDS`
+and `CDM_EXTRA_KEEP_FIELDS` are not moved to `undefined`.
+|`CDM_USE_UNDEFINED=true`
+
+|`CDM_UNDEFINED_NAME`
+|Specify a name for the undefined top level field if using `CDM_USE_UNDEFINED`. The default is`undefined`.
+Enabled only when `CDM_USE_UNDEFINED` is `true`.
+|`CDM_UNDEFINED_NAME="undef"`
+
+|`CDM_UNDEFINED_MAX_NUM_FIELDS`
+a|If the number of undefined fields is greater than this number, all undefined fields are converted to their JSON string representation 
+and stored in the `CDM_UNDEFINED_NAME` field. If the record contains more than this value of undefined fields, no further processing takes place on these fields. 
+Instead, the fields will be converted to a single string JSON value, stored in the top-level `CDM_UNDEFINED_NAME` field.  Keeping the default of `-1` allows for an unlimited number of undefined fields, which is not recommended.
+[NOTE]
+====
+This parameter is honored even if `CDM_USE_UNDEFINED` is false.
+====
+|`CDM_UNDEFINED_MAX_NUM_FIELDS=4`
+
+|`CDM_UNDEFINED_TO_STRING`
+|Set to `true` to convert all undefined fields to their JSON string representation. The default is `false`.
+|`CDM_UNDEFINED_TO_STRING=true`
+
+|`CDM_UNDEFINED_DOT_REPLACE_CHAR`
+|Specify a character to use in place of a dot character '.' in an undefined field. `MERGE_JSON_LOG` must be `true`. The default is `UNUSED`. 
+If you set the `MERGE_JSON_LOG` parameter to `true`, see the Note below.
+|`CDM_UNDEFINED_DOT_REPLACE_CHAR="_"`
+
+|===
+
+[NOTE]
+====
+If you set the `MERGE_JSON_LOG` parameter in the log collector daemonset and `CDM_UNDEFINED_TO_STRING` environment variables to true, you might receive an Elasticsearch *400* error. 
+The error occurs because when`MERGE_JSON_LOG=true`, the log collector adds fields with data types other than string. When you set `CDM_UNDEFINED_TO_STRING=true`, 
+the log collector attempts to add those fields as a string value resulting in the Elasticsearch 400 error. The error clears when the log collector rolls over 
+the indices for the next day’s logs
+
+When the log collector rolls over the indices, it creates a brand new index. The field definitions are updated and you will not get the *400* error.
+====
+
+.Procedure
+
+Use the `CDM_*` parameters to configure undefined and empty field processing.
+
+. Configure how to process fields, as needed:
+
+.. Specify the fields to move using `CDM_EXTRA_KEEP_FIELDS`.
+
+.. Specify any empty fields to retain in the `CDM_KEEP_EMPTY_FIELDS` parameter in CSV format.
+
+. Configure how to process undefined fields, as needed:
+
+.. Set `CDM_USE_UNDEFINED` to `true` to move undefined fields to the top-level `undefined` field:
+
+.. Specify a name for the undefined fields using the `CDM_UNDEFINED_NAME` parameter.
+
+.. Set `CDM_UNDEFINED_MAX_NUM_FIELDS` to a value other than the default `-1`, to set an upper bound on the number of undefined fields 
+in a single record.  
+
+. Specify `CDM_UNDEFINED_DOT_REPLACE_CHAR` to change any dot `.` characters in an undefined field name to another character.
+For example, if `CDM_UNDEFINED_DOT_REPLACE_CHAR=@@@` and there is a field named `foo.bar.baz` the field is transformed into `foo@@@bar@@@baz`.
+
+. Set `UNDEFINED_TO_STRING` to `true` to convert undefined fields to their JSON string representation.  
+
+[NOTE]
+====
+If you configure the `CDM_UNDEFINED_TO_STRING` or `CDM_UNDEFINED_MAX_NUM_FIELDS` parameters, you use the `CDM_UNDEFINED_NAME` to change the undefined field name. This field is needed because `CDM_UNDEFINED_TO_STRING` or `CDM_UNDEFINED_MAX_NUM_FIELDS` could change the value type of the undefined field. When `CDM_UNDEFINED_TO_STRING` or `CDM_UNDEFINED_MAX_NUM_FIELDS` is set to true and there are more undefined fields in a log, the value type becomes `string`. Elasticsearch stops accepting records if the value type is changed, for example, from JSON to JSON string.
+
+For example, when `CDM_UNDEFINED_TO_STRING` is `false` or `CDM_UNDEFINED_MAX_NUM_FIELDS` is the default, `-1`, the value type of the undefined field is `json`. If you change `CDM_UNDEFINED_MAX_NUM_FIELDS` to to a value other than default and there are more undefined fields in a log, the value type becomes `string` (json string). Elasticsearch stops accepting records if the value type is changed.
+====
+
--- a/logging/config/efk-logging-fluentd.adoc
+++ b/logging/config/efk-logging-fluentd.adoc
@@ -21,7 +21,6 @@ For more information, see xref:../../logging/config/efk-logging-management.adoc#
 // assemblies.
 
 
-
 include::modules/efk-logging-fluentd-pod-location.adoc[leveloffset=+1]
 
 include::modules/efk-logging-fluentd-log-viewing.adoc[leveloffset=+1]
@@ -44,5 +43,7 @@ include::modules/efk-logging-fluentd-external.adoc[leveloffset=+1]
 include::modules/efk-logging-fluentd-throttling.adoc[leveloffset=+1]
 
 include::modules/efk-logging-fluentd-json.adoc[leveloffset=+1]
+
+include::modules/efk-logging-fluentd-undefined.adoc[leveloffset=+1]
 
 include::modules/efk-logging-fluentd-envvar.adoc[leveloffset=+1]


mburke5678:
@nhosoi @richm Please take a look at this PR addressing the [undefined fields issue for Fluentd and Rsyslog](https://jira.coreos.com/browse/LOG-372).
I  want to name the topic something other than _undefined fields_, as the environment parameters seem to affect fields other than just undefined. There are parameters to affect undefined, defined, and empty fields. Is this correct?

richm:
> There are parameters to affect undefined, defined, and empty fields. Is this correct?

in the table in the first comment of https://jira.coreos.com/browse/LOG-372 we do not want to document USE_MMEXTERNAL, SKIP_EMPTY, and UNDEFINED_DEBUG.  Those should be considered advanced settings, only for use by authorized Red Hat personnel.  All of the other parameters are common to both rsyslog and fluentd and have the same effect in rsyslog and fluentd.

mburke5678:
@richm I saw your comment in the Jira from yesterday ( From the perspective of a user, they don't need to know about SKIP_EMPTY or USE_MMEXTERNAL) but wasn't sure if this was the final decision, so I left them here. 
Does not documenting these variables conflict with @nhosoi PR on error messaging? https://github.com/openshift/cluster-logging-operator/pull/221
If the error messages talk about setting and unsetting `SKIP_EMPTY` and `USE_MMEXTERNAL` but not describing these variables in the docs be confusing to users who get these messages? 
Maybe I can find some way of describing these fields while telling users they will be set accordingly based on the `CDM_*` fields?

richm:
> @richm I saw your comment in the Jira from yesterday ( From the perspective of a user, they don't need to know about SKIP_EMPTY or USE_MMEXTERNAL) but wasn't sure if this was the final decision, so I left them here.

I think not documenting SKIP_EMPTY and USE_MMEXTERNAL is the right thing to do.

> Does not documenting these variables conflict with @nhosoi PR on error messaging? [openshift/cluster-logging-operator#221](https://github.com/openshift/cluster-logging-operator/pull/221)

No.  An ordinary user who doesn't use SKIP_EMPTY or USE_MMEXTERNAL will never see these messages.
An "advanced" user who does use SKIP_EMPTY or USE_MMEXTERNAL will see these messages and hopefully find them useful.

> If the error messages talk about setting and unsetting `SKIP_EMPTY` and `USE_MMEXTERNAL` but not describing these variables in the docs be confusing to users who get these messages?

See above.

> Maybe I can find some way of describing these fields while telling users they will be set accordingly based on the `CDM_*` fields?

I don't think we need to document SKIP_EMPTY and USE_MMEXTERNAL

mburke5678:
@richm So, how will the "advanced" see `SKIP_EMPTY` in the Rsyslog daemonset and know if he/she wants to use it? Particularly, if the adventurous user sets it to true thinking it is the proper way to remove empty fields (which could result in the error messages).  
_The Rsyslog daemonset contains the `SKIP_EMPTY` environment variable, which is `fasle` by default. If set to `true`, Openshift uses the Rsyslog *mmnormalize* plugin to remove empty fields from the record. Undefined field processing is not performed and the `CDM_` parameters are ignored.  This feature should be used by advanced users or you are or are instructed to do so by Red Hat support._

richm:
> @richm So, how will the "advanced" see `SKIP_EMPTY` in the Rsyslog daemonset and know if he/she wants to use it?

The "advanced" user will already know that it is there (by looking at the source code, e.g.)

> Particularly, if the adventurous user sets it to true thinking it is the proper way to remove empty fields (which could result in the error messages).

The "advanced" user will already know what effect a particular setting will have.  The reminder is only there in case the "advanced" user is playing around with the settings and e.g. forgot to remove or unset USE_MMEXTERNAL.

> _The Rsyslog daemonset contains the `SKIP_EMPTY` environment variable, which is `fasle` by default. If set to `true`, Openshift uses the Rsyslog _mmnormalize_ plugin to remove empty fields from the record. Undefined field processing is not performed and the `CDM_` parameters are ignored. This feature should be used by advanced users or you are or are instructed to do so by Red Hat support._

I would prefer not to document it at all.

mburke5678:
@QiaolingTang Please take a look.

modules/efk-logging-fluentd-undefined.adoc
```diff
@@ -5,22 +5,18 @@
 [id="efk-logging-fluentd-undefined_{context}"]
 = Configuring how the log collector normalizes logs 
 
-Cluster Logging uses the link:https://github.com/ViaQ[ViaQ common data model] to deal with problem that arise from conflicts 
-and inconsistencies in log data collected from different subsystems. {product title} allows you to configure  
+Cluster Logging uses the link:https://github.com/ViaQ/fluent-plugin-viaq_data_model[ViaQ common data model] to deal with problem that arise from conflicts
```
richm: Fluentd uses that plugin.  Rsyslog uses several different ways.

modules/efk-logging-fluentd-undefined.adoc
```diff
@@ -5,22 +5,18 @@
 [id="efk-logging-fluentd-undefined_{context}"]
 = Configuring how the log collector normalizes logs
```
richm: This section is primarily about how to deal with problems that arise when using `MERGE_JSON_LOG=true`.  It can be extremely useful to have your applications log their output in JSON, and have OpenShift logging automatically parse and index the data in Kibana, but this leads to several problems: field names may be empty, or contain characters that are illegal in Elasticsearch; different applications in the same namespace may output the same field name with different value data types; applications may emit too many fields; fields may conflict with the OpenShift logging built-in fields, and more.

modules/efk-logging-fluentd-undefined.adoc
```diff
@@ -0,0 +1,121 @@
...
++
+You can also convert undefined fields to JSON format and replace special characters.
+
+* Defined fields. You can also configure which defined fields appear in the top levels of the logs. 
++
+[[default-fields]]
+The default top-level fields are `CEE,time`, `@timestamp`, `aushape`, `ci_job`, `collectd`, `docker`, `fedora-ci`, 
+`file`, `foreman`, `geoip`, `hostname`, `ipaddr4`, `ipaddr6`, `kubernetes`, `level`, `message`, `namespace_name`, `namespace_uuid`,
+`offset`, `openstack`, `ovirt`, `pid`, `pipeline_metadata`, `rsyslog`, `service`, `systemd`, `tags`, `testcase`, `tlog`, `viaq_msg_id`.
+You can use the `CDM_DEFAULT_KEEP_FIELDS` parameter to add additional defaults and the `CDM_EXTRA_KEEP_FIELDS` parameter to include
```
QiaolingTang: I thought we'd better not to document about changing the `CDM_DEFAULT_KEEP_FIELDS`.  If users want to keep a field out of undefined fields, using `CDM_EXTRA_KEEP_FIELDS` is enough.

modules/efk-logging-fluentd-undefined.adoc
```diff
@@ -0,0 +1,121 @@
...
+* Defined fields. You can also configure which defined fields appear in the top levels of the logs. 
++
+[[default-fields]]
+The default top-level fields are `CEE,time`, `@timestamp`, `aushape`, `ci_job`, `collectd`, `docker`, `fedora-ci`, 
+`file`, `foreman`, `geoip`, `hostname`, `ipaddr4`, `ipaddr6`, `kubernetes`, `level`, `message`, `namespace_name`, `namespace_uuid`,
+`offset`, `openstack`, `ovirt`, `pid`, `pipeline_metadata`, `rsyslog`, `service`, `systemd`, `tags`, `testcase`, `tlog`, `viaq_msg_id`.
+You can use the `CDM_DEFAULT_KEEP_FIELDS` parameter to add additional defaults and the `CDM_EXTRA_KEEP_FIELDS` parameter to include 
+additional non-default fields at the top level.
++
+Any defined fields not included in `${CDM_DEFAULT_KEEP_FIELDS}` or `${CDM_EXTRA_KEEP_FIELDS}` are moved to `${CDM_UNDEFINED_NAME}` if `CDM_USE_UNDEFINED` is `true`.
```
QiaolingTang: Suggestion: remove `defined` in `Any defined fields not included in`

modules/efk-logging-fluentd-undefined.adoc
```diff
@@ -0,0 +1,121 @@
...
+[cols="2,4,2", options="header"]
+|===
+|Parameters  |Definition  |Example
+
+|CDM_DEFAULT_KEEP_FIELDS
+|Specify a set of defined fields add as default top-level fields. 
+|See the xref:default-fields[list of defaults].
+	-
+|CDM_EXTRA_KEEP_FIELDS
+|Specify an extra set of defined fields to be kept in addition to the `CDM_DEFAULT_KEEP_FIELDS`. The default is "".
```
QiaolingTang: I thought we may change this description to "Specify an extra set of fields to be kept in the top level of json in addition to the `CDM_DEFAULT_KEEP_FIELDS` when `CDM_USE_UNDEFINED` is "true". " or something like this.

modules/efk-logging-fluentd-undefined.adoc
```diff
@@ -0,0 +1,121 @@
...
+|CDM_EXTRA_KEEP_FIELDS="broker"
+
+|CDM_KEEP_EMPTY_FIELDS
+|Specify defined fields to retain even if empty. Empty defined fields not specified are dropped. The default is "".
```
QiaolingTang: Suggestion: remove `defined`.

modules/efk-logging-fluentd-undefined.adoc
```diff
@@ -0,0 +1,121 @@
...
+|CDM_UNDEFINED_NAME
+|Specify a name for the undefined top level field if using `CDM_USE_UNDEFINED`. The default is`undefined`.
+Enabled only when `CDM_USE_UNDEFINED` is `true`.
+|CDM_UNDEFINED_NAME="undef"
+
+|CDM_UNDEFINED_MAX_NUM_FIELDS
+|If the number of undefined fields is greater than this number, all undefined fields are converted to their JSON string representation 
+and stored in the `CDM_UNDEFINED_NAME` field. If the record contains more than this value of undefined fields, no further processing takes place on these fields. 
+Instead, the fields will be converted to a single string JSON value, stored in the top-level `CDM_UNDEFINED_NAME` field. Keeping the default of `-1`allows for an unlimited number of undefined fields, which is not recommended.
+Enabled only when `CDM_USE_UNDEFINED` is `true`.
```
QiaolingTang: According to https://jira.coreos.com/browse/LOG-372?focusedCommentId=114310&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-114310, thought this line should be removed.

modules/efk-logging-fluentd-undefined.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+.Procedure
+
+Use the `CDM_*` parameters to configure undefined and empty field processing.
+
+. Configure how to process defined fields, as needed:
+
+.. Specify the defined fields to move using the `CDM_DEFAULT_KEEP_FIELDS` and `CDM_EXTRA_KEEP_FIELDS`.
+
+.. Set `CDM_KEEP_EMPTY_FIELDS` to `true` to retain empty defined fields.
```
QiaolingTang: Set `CDM_KEEP_EMPTY_FIELDS` to `true` may cause some misunderstanding here, because the value of `CDM_KEEP_EMPTY_FIELDS` is in the CSV format(the same type as the `CDM_DEFAULT_KEEP_FIELDS` and `CDM_EXTRA_KEEP_FIELDS`), and the user may think `true` is a boolean value.

For example:
CDM_KEEP_EMPTY_FIELDS="loggerName"

modules/efk-logging-fluentd-undefined.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+. Configure how to process undefined fields, as needed:
+
+.. Set `CDM_USE_UNDEFINED` to `true` to move undefined fields:
```
QiaolingTang: Set `CDM_USE_UNDEFINED` to `true` to move undefined fields to the undefined property.

modules/efk-logging-fluentd-undefined.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+.. Specify a name for the undefined fields using the `CDM_UNDEFINED_NAME` parameter.
+
+.. Set `CDM_UNDEFINED_MAX_NUM_FIELDS` to a value other than the default `-1`, to set an upper bound on the number of undefined fields 
+in a single record.  
+
+. Set `UNDEFINED_TO_STRING` to `true` to convert undefined fields to JSON format.
```
QiaolingTang: The value is converted to the json string when set `UNDEFINED_TO_STRING` to `true`.

For example, if not set `UNDEFINED_TO_STRING` , we may see a field's value like:
 "Number": 10,
10 is an integer here
but if set `UNDEFINED_TO_STRING` to `true`, then the value should be:
"Number": "10",
"10" is a string here

modules/efk-logging-fluentd-undefined.adoc
```diff
@@ -0,0 +1,121 @@
...
+.. Set `CDM_USE_UNDEFINED` to `true` to move undefined fields:
+
+.. Specify a name for the undefined fields using the `CDM_UNDEFINED_NAME` parameter.
+
+.. Set `CDM_UNDEFINED_MAX_NUM_FIELDS` to a value other than the default `-1`, to set an upper bound on the number of undefined fields 
+in a single record.  
+
+. Set `UNDEFINED_TO_STRING` to `true` to convert undefined fields to JSON format. 
+
+. Specify `CDM_UNDEFINED_DOT_REPLACE_CHAR` to change any dot `.` characters in an undefined field name to another character.
```
QiaolingTang: Seems we only support "_", not sure if we could convert "." to other characters. 

@richm Do we support convert "." to other characters except "_" ?

richm: Specify `CDM_UNDEFINED_DOT_REPLACE_CHAR=string` where `string` is the string value you want to use to replace the `.` in the field name.  For example, if `CDM_UNDEFINED_DOT_REPLACE_CHAR=@@@` and there is a field named `foo.bar.baz` then the field will be transformed into `foo@@@bar@@@baz`

modules/efk-logging-fluentd-undefined.adoc
richm: To be specific - `UNDEFINED_TO_STRING` will cause the value to be transformed to its JSON string representation.  For simple scalar values like numbers and booleans, this just means the value will be a quoted string - 10 -> "10", 3.1415 -> "3.1415", false -> "false".

For map/dict values and array values, they are converted to their JSON string representation: `"mapfield":{"key":"value"}` becomes `"mapfield":"{\"key\":\"value\"}"` and `"arrayfield":[1,2,"three"]` becomes `"arrayfield":"[1,2,\"three\"]"` - this at least preserves the structure of the value, so that you can retrieve the value later and convert it back to a map or an array.

modules/efk-logging-fluentd-undefined.adoc
richm: use the word "field" instead of "property" in this context

modules/efk-logging-fluentd-undefined.adoc
richm: yes, sounds good - you are correct - `CDM_DEFAULT_KEEP_FIELDS` is one of those "advanced" settings that regular users do not need to know about

mburke5678: I added a note that `CDM_DEFAULT_KEEP_FIELDS` is for advanced users. I would like to define the parameter as we mention the parameter in other places (although the name might be descriptive enough).
```
The default top-level fields, defined through the `CDM_DEFAULT_KEEP_FIELDS` parameter, are `CEE,time`, `@timestamp`, `aushape`, `ci_job`, `collectd`, `docker`, `fedora-ci`, 
`file`, `foreman`, `geoip`, `hostname`, `ipaddr4`, `ipaddr6`, `kubernetes`, `level`, `message`, `namespace_name`, `namespace_uuid`,
`offset`, `openstack`, `ovirt`, `pid`, `pipeline_metadata`, `rsyslog`, `service`, `systemd`, `tags`, `testcase`, `tlog`, `viaq_msg_id`.
+
+
[NOTE]
====
The `CDM_DEFAULT_KEEP_FIELDS` parameter is for only advanced users or you are instructed to do so by Red Hat support.
==== 
```

modules/efk-logging-fluentd-undefined.adoc
mburke5678: Sorry. I missed Rich's message on this from yesterday.

modules/efk-logging-fluentd-undefined.adoc
richm: The `CDM_DEFAULT_KEEP_FIELDS` parameter is for only advanced users, or if you are instructed to use it by Red Hat support.

modules/efk-logging-fluentd-undefined.adoc
mburke5678: _Set `CDM_USE_UNDEFINED` to `true` to move undefined fields to the top-level `undefined` field_ ?

modules/efk-logging-fluentd-undefined.adoc
mburke5678: I borrowed the _or you are..._ phrasing from elsewhere in the docs. But, I like this ^^ better.

modules/efk-logging-fluentd-undefined.adoc
```diff
@@ -0,0 +1,121 @@
...
++
+Use the parameters to configure how {product-title} moves any undefined fields under a top-level field called `undefined` to avoid conflicting with the _well known_ ViaQ 
+top-level fields. You can add undefined fields to the top-level fields and move others to an `undefined` container.
++
+You can also convert undefined fields to JSON format and replace special characters.
+
+* Defined fields. You can also configure which defined fields appear in the top levels of the logs. 
++
+[[default-fields]]
+The default top-level fields are `CEE,time`, `@timestamp`, `aushape`, `ci_job`, `collectd`, `docker`, `fedora-ci`,
```
nhosoi: `CEE,time` looks one word...  Should they be separated?

modules/efk-logging-fluentd-undefined.adoc
mburke5678: Outdated

modules/efk-logging-fluentd-undefined.adoc
mburke5678: Outdated

modules/efk-logging-fluentd-undefined.adoc
mburke5678: Outdated

modules/efk-logging-fluentd-undefined.adoc
mburke5678: Outdated

modules/efk-logging-fluentd-undefined.adoc
mburke5678: Outdated

modules/efk-logging-fluentd-undefined.adoc
nhosoi: Rather, can we add a note that this [1] is executed even if `CDM_USE_UNDEFINED` is false?  That'd reduce the confusion (as I had... :)

[1]  "the fields will be converted to a single string JSON value, stored in the top-level `CDM_UNDEFINED_NAME` field"

modules/efk-logging-fluentd-undefined.adoc
mburke5678: Outdated

mburke5678:
@QiaolingTang @richm @nhosoi Thank you for your comments. I think I got them all in. Please let me know what else I can change/improve/correct.

modules/efk-logging-fluentd-undefined.adoc
```diff
@@ -0,0 +1,132 @@
...
+//
+// * logging/efk-logging-fluentd.adoc
+
+[id="efk-logging-fluentd-undefined_{context}"]
+= Configuring how the log collector normalizes logs 
+
+Cluster Logging uses the link:https://github.com/ViaQ/fluent-plugin-viaq_data_model[ViaQ common data model] to deal with problem that arise from conflicts 
+and inconsistencies in log data collected from different subsystems.   
+
+If you use the `MERGE_JSON_LOG` feature (MERGE_JSON_LOG=true), it can be extremely useful to have your applications log their output in JSON, and have cluster logging automatically parse and index the data in Kibana. However, this leads to several problems, including:
```
QiaolingTang: Thought you might want to say "and have log collectors automatically parse and index the data in Elasticsearch" here.

mburke5678: @QiaolingTang Thank you. I made the change.

mburke5678:
@QiaolingTang What is the QE status of this PR?

modules/efk-logging-fluentd-undefined.adoc
```diff
@@ -0,0 +1,132 @@
...
+
+.. Set `CDM_USE_UNDEFINED` to `true` to move undefined fields to the top-level `undefined` field:
+
+.. Specify a name for the undefined fields using the `CDM_UNDEFINED_NAME` parameter.
+
+.. Set `CDM_UNDEFINED_MAX_NUM_FIELDS` to a value other than the default `-1`, to set an upper bound on the number of undefined fields 
+in a single record.  
+
+. Set `UNDEFINED_TO_STRING` to `true` to convert undefined fields to their JSON string representation.  
+For example, if `CDM_UNDEFINED_DOT_REPLACE_CHAR=@@@` and there is a field named `foo.bar.baz` the field is transformed into `foo@@@bar@@@baz`.
```
QiaolingTang: @mburke5678 I think this line should be under the Line 124.

modules/efk-logging-fluentd-undefined.adoc
```diff
@@ -0,0 +1,132 @@
...
+//
+// * logging/efk-logging-fluentd.adoc
+
+[id="efk-logging-fluentd-undefined_{context}"]
+= Configuring how the log collector normalizes logs 
+
+Cluster Logging uses the link:https://github.com/ViaQ/fluent-plugin-viaq_data_model[ViaQ common data model] to deal with problem that arise from conflicts 
+and inconsistencies in log data collected from different subsystems.   
+
+If you use the `MERGE_JSON_LOG` feature (MERGE_JSON_LOG=true), it can be extremely useful to have your applications log their output in JSON, and have the log collector automatically parse and index the data in Kibana. However, this leads to several problems, including:
```
QiaolingTang: @mburke5678  `and have the log collector automatically parse and index the data in Kibana` please change `Kibana` to `Elasticsearch`

modules/efk-logging-fluentd-undefined.adoc
mburke5678: @nhosoi Added a note. Does this meet your expectations? 
5af4d98f06a472449bbc865b3586e2f55278dcd9

modules/efk-logging-fluentd-undefined.adoc
mburke5678: @QiaolingTang Sorry. I missed this comment earlier. I think I got all the comments in.

modules/efk-logging-fluentd-undefined.adoc
mburke5678: @richm Is there a good link to use for info on ViaQ? That is what is needed here. What is the ViaQ data model?

modules/efk-logging-fluentd-undefined.adoc
```diff
@@ -0,0 +1,136 @@
+// Module included in the following assemblies:
+//
+// * logging/efk-logging-fluentd.adoc
+
+[id="efk-logging-fluentd-undefined_{context}"]
+= Configuring how the log collector normalizes logs 
+
+Cluster Logging uses the link:https://github.com/ViaQ[ViaQ common data model] to deal with problem that arise from conflicts 
+and inconsistencies in log data collected from different subsystems.
```
richm: Probably shouldn't mention that specific plugin for that specific collector.  Instead let's say something like this:
```
Cluster Logging uses a certain data model ([1],[2]), like a database schema, to store log records and their metadata in the logging store.  There are some restrictions on the data:

* there must be a `"message"` field containing the actual log message
* there must be a `"@timestamp"` field containing the log record timestamp in rfc3339 (preferably millisecond or better resolution) format
* there must be a `"level"` field with the log level e.g. `err`, `info`, `unknown`, etc.

Cluster Logging attempts to parse and normalize the log data, but there are some cases where it cannot, especially if you use the `MERGE_JSON_LOG` feature.
```
[1],[2] are footnotes or some sort of links to https://github.com/openshift/origin-aggregated-logging/blob/master/docs/com.redhat.viaq-openshift-operations.asciidoc and https://github.com/openshift/origin-aggregated-logging/blob/master/docs/com.redhat.viaq-openshift-project.asciidoc

Or perhaps we already document the fields available for searching and listing in Kibana in our docs?  If so, then just link that here.

mburke5678: Thanks!  1a00924166992d44b90ddeb28f3f887a1fb351e9

mburke5678:
@QiaolingTang Can I merge this or do you have more concerns?

modules/efk-logging-fluentd-undefined.adoc
```diff
@@ -0,0 +1,146 @@
...
+[cols="2,4,2", options="header"]
+|===
+|Parameters  |Definition  |Example
+
+|`CDM_EXTRA_KEEP_FIELDS`
+|Specify an extra set of defined fields to be kept at the top level of the logs in addition to the `CDM_DEFAULT_KEEP_FIELDS`. The default is "".
+|`CDM_EXTRA_KEEP_FIELDS="broker"`
+
+|`CDM_KEEP_EMPTY_FIELDS`
+|Specify fields to retain even if empty in CSV format. Empty defined fields not specified are dropped. The default is "".
```
QiaolingTang: @mburke5678 We have a default value `message`, see https://github.com/openshift/cluster-logging-operator/blob/master/pkg/k8shandler/fluentd.go#L239 and https://github.com/openshift/cluster-logging-operator/blob/master/pkg/k8shandler/rsyslog.go#L366

modules/efk-logging-fluentd-undefined.adoc
```diff
@@ -0,0 +1,146 @@
...
+
+. Set `UNDEFINED_TO_STRING` to `true` to convert undefined fields to their JSON string representation.  
+
+[NOTE]
+====
+If you configure the `CDM_UNDEFINED_TO_STRING` or `CDM_UNDEFINED_MAX_NUM_FIELDS` parameters, you use the `CDM_UNDEFINED_NAME` to change the undefined field name. This field is needed because `CDM_UNDEFINED_TO_STRING` or `CDM_UNDEFINED_MAX_NUM_FIELDS` could change the value type of the undefined field. When `CDM_UNDEFINED_TO_STRING` or `CDM_UNDEFINED_MAX_NUM_FIELDS` is set to true and there are more undefined fields in a log, the value type becomes `string`. Elasticsearch stops accepting records if the value type is changed, for example, from JSON to JSON string. When `CDM_UNDEFINED_TO_STRING` or `CDM_UNDEFINED_MAX_NUM_FIELDS` are `false`, the value type of the undefined field is JSON.
+
+For example, when `CDM_UNDEFINED_TO_STRING` or `CDM_UNDEFINED_MAX_NUM_FIELDS` are set to false, the value type is `json`. If you change `CDM_UNDEFINED_MAX_NUM_FIELDS` to `true` and there are more undefined fields in a log, the value type becomes `string` (json string). Elasticsearch stops accepting records if the value type is changed.
```
QiaolingTang: Please note the value of `CDM_UNDEFINED_MAX_NUM_FIELDS` should be a number, and could not be `true` or `false`

mburke5678: @QiaolingTang 
Should this state: 
_For example, when `CDM_UNDEFINED_TO_STRING` is `false` or `CDM_UNDEFINED_MAX_NUM_FIELDS` is default, `-1`, the value type is `json`. If you change `CDM_UNDEFINED_MAX_NUM_FIELDS` to to a value other than default...

mburke5678:
Moving to 4.3. Reverting changes added to 4.2

https://github.com/openshift/openshift-docs/pull/16597

PR #16029, opened 2019-07-26
Add multitenant isolation using NetworkPolicy objects

jboxman:
- https://bugzilla.redhat.com/show_bug.cgi?id=1719936

Updated to include new project template module directly.

~This is awkward, because we already have a module that explains how to create a default new project template for the cluster, but that needs to be done so that NetworkPolicy objects can be added to it. But as I understand it, I can't link to that from a module, so I put the entire contents in the assembly.~

~I'm open to suggestions on alternate approaches for this.~

### DIFF
--- a/modules/nw-networkpolicy-project-defaults.adoc
+++ b/modules/nw-networkpolicy-project-defaults.adoc
@@ -1,0 +1,80 @@
+// Module included in the following assemblies:
+//
+// * networking/configuring-networkpolicy.adoc
+
+[id="networkpolicy-adding-network-policy-objects-to-the-new-project-template_{context}"]
+= Adding network policy objects to the new project template
+
+As a cluster administrator, you can add network policy objects to the default template for new projects.
+{product-title} will automatically create all the NetworkPolicy CRs specified in the template in the project.
+
+.Prerequisites
+
+* A cluster using the OpenShift SDN network plug-in with `mode: NetworkPolicy`
+set. This mode is the default for OpenShift SDN.
+* Install the OpenShift Command-line Interface (CLI), commonly known as `oc`.
+* You must log in to the cluster with a user with `cluster-admin` privileges.
+* You must have created a custom default project template for new projects.
+
+.Procedure
+
+. Edit the default template for a new project by running the following command:
++
+----
+$ oc edit template <project_template> -n openshift-config
+----
++
+Replace `<project_template>` with the name of the default template that you
+configured for your cluster. The default template name is `project-request`.
+
+. In the template, add each NetworkPolicy object as an element to the `objects`
+parameter. The `objects` parameter accepts a collection of one or more objects.
++
+In the following example, the `objects` parameter collection includes several
+NetworkPolicy objects:
++
+[source,yaml]
+----
+objects:
+- apiVersion: networking.k8s.io/v1
+  kind: NetworkPolicy
+  metadata:
+    name: allow-from-same-namespace
+  spec:
+    podSelector:
+    ingress:
+    - from:
+      - podSelector: {}
+- apiVersion: networking.k8s.io/v1
+  kind: NetworkPolicy
+  metadata:
+    name: allow-from-openshift-ingress
+  spec:
+    ingress:
+    - from:
+      - namespaceSelector:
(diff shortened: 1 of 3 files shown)


huffmanca (commented):
Overall this looks good, just a few suggestions about including specific examples vs generic ones.

jboxman:
@huffmanca, you found this before I anticipated anyone would look; There's actually one additional policy file that is necessary, for metrics. So there will probably be two, but possibly more. I've been trying to track down what else may be required or optionally useful to include.

jboxman:
Hi @zhaozhanqi,

Can you PTAL at this? It adds instructions for emulating _multitenant_ mode using NetworkPolicy objects and creating default NetworkPolicy objects in new projects.

Thanks.

networking/configuring-networkpolicy.adoc
```diff
@@ -14,3 +14,65 @@ include::modules/nw-networkpolicy-create.adoc[leveloffset=+1]
...
+* Install the OpenShift Command-line Interface (CLI), commonly known as `oc`.
+* You must log in to the cluster with a user with `cluster-admin` privileges.
+* You must xref:../applications/projects/configuring-project-creation.adoc#modifying-template-for-new-projects_configuring-project-creation[create a custom template for new projects].
+
+.Procedure
+
+. Edit the default template for a new project by running the following command:
++
+----
+$ oc edit template <project_template> -n openshift-config
```
zhaozhanqi: seems the default template is not exist by default.  we need to create it via
`oc adm create-bootstrap-project-template -o yaml > template.yaml`
`oc create -f template.yaml -n openshift-config`
Then we can edit the template
@xingxingxia  please correct me if I'm wrong.

networking/configuring-networkpolicy.adoc
```diff
@@ -14,3 +14,65 @@ include::modules/nw-networkpolicy-create.adoc[leveloffset=+1]
...
+    ingress:
+    - from:
+      - namespaceSelector:
+          matchLabels:
+            network.openshift.io/policy-group: ingress
+    podSelector: {}
+    policyTypes:
+    - Ingress
+...
+----
```
zhaozhanqi: Better add one step to check if the new created project has the default networkpolicy,like 
```
oc new-project test
oc get networkpolicy
NAME                           POD-SELECTOR   AGE
allow-from-openshift-ingress   <none>         11m
allow-from-same-namespace      <none>         11m
```

networking/configuring-networkpolicy.adoc
xingxingxia: @zhouying7780 could you help look at it when having time?

jboxman: Oh, that step is actually included as a prerequisite: _You must create a custom template for new projects._ If the user follows that link, the procedure has steps for creating a default template for projects.

networking/configuring-networkpolicy.adoc
jboxman: @zhaozhanqi that's a great idea, I'll add that as a step!

jboxman:
I may need to reorganize this content for clarity around creating a new project template.

jboxman:
@kalexand-rh, I updated this to include the dependent module directly.

kalexand-rh:
@jboxman, I think this looks good for now.

modules/nw-networkpolicy-project-defaults.adoc
```diff
@@ -0,0 +1,80 @@
+// Module included in the following assemblies:
+//
+// * networking/configuring-networkpolicy.adoc
+
+[id="networkpolicy-new-project_{context}"]
```
kalexand-rh: I'd make the anchor ID match the module title.

networking/configuring-networkpolicy.adoc
jboxman: I reworked the content so the prerequisite is more visible.

jboxman:
@zhaozhanqi, I've incorporated your suggestions. Let me know what you think. Thanks!

zhaozhanqi:
LGTM

PR #16035, opened 2019-07-26
Updated TP note to align with published standard.

huffmanca:
Updated TP note to align with published standard.

This is for OS 4.x.


kalexand-rh (commented):
LGTM. Thank you, @huffmanca!

PR #16044, opened 2019-07-26
PR13988 and PR15869 follow-up change squash layers note

bmcelvee:
Change requested in https://github.com/openshift/openshift-docs/pull/13988, based on https://jira.coreos.com/browse/RUN-405 and https://github.com/openshift/builder/pull/35.

@adambkaplan and @wewang58

### DIFF
--- a/modules/builds-strategy-docker-squash-layers.adoc
+++ b/modules/builds-strategy-docker-squash-layers.adoc
@@ -15,6 +15,5 @@ all instructions into a single layer on top of the base image.
 ----
 strategy:
   dockerStrategy:
-    imageOptimizationPolicy: SkipLayers <1>
+    imageOptimizationPolicy: SkipLayers
 ----
-<1> Layers are always squashed in {product-title} {product-version}.


modules/builds-strategy-docker-squash-layers.adoc
```diff
@@ -17,4 +17,4 @@ strategy:
   dockerStrategy:
     imageOptimizationPolicy: SkipLayers <1>
 ----
-<1> Layers are always squashed in {product-title} {product-version}.
+<1> Builds will keep their layers by default.
```
adambkaplan: Personally I would just get rid of this footnote, since we say the same thing above:

> Docker builds normally create a layer representing each instruction in a
`Dockerfile`. Setting the `imageOptimizationPolicy` to `SkipLayers` will merge
all instructions into a single layer on top of the base image.

adambkaplan (changes requested):

modules/builds-strategy-docker-squash-layers.adoc
bmcelvee: Easy enough!

bmcelvee:
Since this is such a small change (just removing the footnote), I don't think it needs peer review.

PR #16063, opened 2019-07-30
SRVCOM-490 documentation for Knative services

abrennan89:
Adding documentation for getting started with Knative services, configuring domains and deploying serverless applications.

### DIFF
--- a/modules/configuring-knative-services-domains.adoc
+++ b/modules/configuring-knative-services-domains.adoc
@@ -1,0 +1,40 @@
+// Module included in the following assemblies:
+//
+// * serverless/getting-started-knative-services.adoc
+
+[id="configuring-knative-services-domains_{context}"]
+= Configuring domains for Knative services
+
+The Red Hat OpenShift Serverless Operator will attempt to find the public domain that is used by the cluster and use that same domain for Knative Services. Having a domain configured makes it possible to reach services in your cluster via HTTP over the internet directly, for example, using the `curl` command.
+
+Alternatively, you can configure a different domain for Knative services or even configure a different domain for different Knative services based on labels. Once you have completed this configuration, you can create a template for the domain configuration.
+
+.Procedure
+. To change the configuration for Knative services, you must edit the `config-domain` ConfigMap in the `knative-serving` namespace using `oc edit`.
+
+   # oc edit cm config-domain -n knative-serving
+
+. Copy the sample ConfigMap and modify the relevant fields.
++
+[source,yaml]
+----
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config-domain
+  namespace: knative-serving
+  labels:
+    serving.knative.dev/release: devel
+data:
+  example.com: |
+----
++
+Here is an example of the configuration for the domain `example.com`, which will be used for routes with the label `app` set to `example`.
++
+[source,yaml]
+----
+data:
+  example.com: |
+    selector:
+      app: example
+----
--- a/modules/domain-templates-knative-services.adoc
+++ b/modules/domain-templates-knative-services.adoc
@@ -1,0 +1,36 @@
+// Module included in the following assemblies:
+//
+// * serverless/knative-services.adoc
+
+[id="domain-templates-knative-services_{context}"]
+= Defining a template for the configured domain
+
+.Procedure
+. Define a `domainTemplate`.
++
+`domainTemplate` specifies the golang text template string to use when constructing the Knative service's DNS name.
++
+The default value is `{{.Name}}.{{.Namespace}}.{{.Domain}}`.
++
+An example of this DNS name would look like `myapp.mynamespace.example.com` for an application called `myapp` with a namespace called `mynamespace`, and `example.com` as the domain.
++
+- You will need to change this value if extra levels in the domain name cause issues. For example, if the wildcard certificate’s domain only supports a single level domain name.
+
+- To resolve these types of issues, use the format `{{.Name}}-{{.Namespace}}.{{.Domain}}` or remove the Namespace entirely from the template.
+
+- When choosing a new value, remember that using characters such as dashes in service or namespace names can cause conflicts and should be avoided.
++
+NOTE: It is strongly recommended to keep the namespace part of the template to avoid domain name clashes.
++
+
+. Define a `tagTemplate`.
++
+`tagTemplate` specifies the golang text template string to use when constructing the DNS name for tags within traffic blocks.
++
+Tags are similar to subroutes that enable traffic blocks to specific revisions.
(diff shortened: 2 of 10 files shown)


modules/configuring-knative-services-domains.adoc
```diff
@@ -0,0 +1,34 @@
...
+      serving.knative.dev/release: devel
+  data:
+    example.com: |
+
+.. Here is an example of configuration for the domain `example.org`, which will be used for routes with the label `app` set to `example`.
+
+  data:
+    example.org: |
+  selector:
+      app: example
```
mgencur: I think the indentation for the selector is not correct here. It should be:
```
data:
  example.org: |
    selector:
      app: example
```
There can be multiple domains defined and the next domain will be at the same level as example.org.

mgencur (commented):
hi @abrennan89 generally looks good. Just a small nit.

modules/configuring-knative-services-domains.adoc
abrennan89: Thanks Martin! Should be fixed in the new commit.

modules/domain-templates-knative-services.adoc
```diff
@@ -0,0 +1,35 @@
...
+NOTE: It is strongly recommended to keep the namespace part of the template to avoid domain name clashes.
+
+. Define a `tagTemplate`.
+
+`tagTemplate` specifies the golang text template string to use when constructing the DNS name for tags within traffic blocks.
+
+Tags are like subroutes which enables traffic blocks to target a specific revision.
+Tags are used along with the `domainTemplate` to determine the full URL for the tag in the format `{{.Name}}-{{.Tag}}`.
+
+For example, if your service is called "myapp", the namespace is "mynamespace", the domain is "example.com" and the tag you want is "latest", the generated host will be `myapp-latest.mynamespace.example.com`.
```
huffmanca: Use backticks for literals instead of quotations.

modules/domain-templates-knative-services.adoc
```diff
@@ -0,0 +1,35 @@
+// Module included in the following assemblies:
+//
+// * serverless/knative-services.adoc
+
+[id="domain-templates-knative-services_{context}"]
+= Defining a template for the configured domain
+
+.Procedure
+. Define a `domainTemplate`.
```
huffmanca: This note isn't specific to the line, but applies to the entire file. Because you use spacing for indentation, the ordered list breaks. You have two steps labeled as step 1.

I'd strongly recommend using `+` to connect steps in a procedure, and wrapping source blocks in `----` to explicitly mark their beginning and ending. This should also help maintenance should these sections need to change in the future.

huffmanca (changes requested):
The content largely looks good, but I did request a few changes surrounding formatting.

As mentioned in one of the comments, I would strongly recommend not using spacing for code blocks. Instead, explicitly wrap these in `----`. It prevents parsing/indentation issues and makes your modules easier to maintain moving forward.

Another common issue that I saw (and attempted to highlight) - all CLI commands should begin with either a $ or #. 

Outside of these issues it looks good! Feel free to ping me directly if you have any questions on any of these comments.

abrennan89:
@huffmanca the notes should all be addressed now :) lmk if there's anything else, otherwise this is good to merge I think (still commented out of the yml for now until release).

abrennan89:
@huffmanca the comments should all be addressed now, let me know if there's anything else and thank you for the great feedback so far :slightly_smiling_face: I think once it has been peer reviewed it's ready to merge (still commented out of the yml until release).

modules/domain-templates-knative-services.adoc
```diff
@@ -0,0 +1,36 @@
...
++
+`domainTemplate` specifies the golang text template string to use when constructing the Knative service's DNS name.
++
+The default value is `{{.Name}}.{{.Namespace}}.{{.Domain}}`.
++
+Ax example of this DNS name would look like `myapp.mynamespace.example.com` for an application called `myapp` with a namespace called `mynamespace`, and `example.com` as the domain.
```
abrennan89: Fix typo @abrennan89

modules/configuring-knative-services-domains.adoc
```diff
@@ -0,0 +1,38 @@
...
+
+Alternatively, you can configure a different domain for Knative services or even configure a different domain for different Knative services based on labels. Once you have completed this configuration, you can create a template for the domain configuration.
+
+.Procedure
+. To change the configuration for Knative services, you must edit the `config-domain` ConfigMap in the `knative-serving` namespace using `oc edit`.
+
+   # oc edit cm config-domain -n knative-serving
+
+. Copy the sample ConfigMap and modify the relevant fields.
+
```
huffmanca: You need a `+` here to identify the source block as part of this command.

I'd also include `[source,yaml]` before this block.

modules/configuring-knative-services-domains.adoc
```diff
@@ -0,0 +1,38 @@
...
+kind: ConfigMap
+metadata:
+  name: config-domain
+  namespace: knative-serving
+  labels:
+    serving.knative.dev/release: devel
+data:
+  example.com: |
+----
+
```
huffmanca: Include a `+` here to keep the text part of the indentation level.

modules/configuring-knative-services-domains.adoc
```diff
@@ -0,0 +1,38 @@
...
+Here is an example of the configuration for the domain `example.com`, which will be used for routes with the label `app` set to `example`.
+
```
huffmanca: Include a `+` here to keep the example part of the indentation level.

This should also have a `[source,yaml]` block before it.

modules/configuring-knative-services-domains.adoc
```diff
@@ -0,0 +1,38 @@
...
+----
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config-domain
+  namespace: knative-serving
+  labels:
+    serving.knative.dev/release: devel
+data:
+  example.com: |
```
huffmanca: Is the `|` needed here?

modules/configuring-knative-services-domains.adoc
```diff
@@ -0,0 +1,38 @@
...
+----
+
+Here is an example of the configuration for the domain `example.com`, which will be used for routes with the label `app` set to `example`.
+
+----
+data:
+  example.com: |
```
huffmanca: Same question as above - is the `|` needed?

modules/configuring-knative-services-domains.adoc
abrennan89: yes it's part of the syntax apparently for yaml

PR #16077, opened 2019-07-31
Moved the Viewing cluster logging to its own topic

mburke5678:
Moved the Viewing cluster logging to its own topic with new assembly


mburke5678:
@sichvoge requested information that we are evaluating using one log collector and the ability to switch log collectors might be removed at some point. He promised to speak with marketing to get the messaging right,
https://coreos.slack.com/archives/GGUR75P60/p1566048251302300?thread_ts=1565977930.296100&cid=GGUR75P60

PR #16085, opened 2019-07-31
bug  1724684 clarifying AWS key/token type

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1724684

@openshift/team-documentation, PTAL. It's the same note in multiple places.

### DIFF
--- a/modules/installation-aws-iam-user.adoc
+++ b/modules/installation-aws-iam-user.adoc
@@ -39,5 +39,13 @@ individual components, so the same credentials are used by all components.
 `AdministratorAccess` policy.
 
 . Record the access key ID and secret access key values. You must use these
-values when you configure your local machine to run the installation program, and this is
-the only time that you can obtain them.
+values when you configure your local machine to run the installation program.
++
+[IMPORTANT]
+====
+You cannot use a temporary session token that you generated while using a
+multi-factor authentication device to authenticate to AWS when you deploy a
+cluster. The cluster continues to use your current AWS credentials to
+create AWS resources for the entire life of the cluster, so you must
+use key-based, long-lived credentials.
+====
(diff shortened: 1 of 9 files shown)


modules/installation-aws-iam-user.adoc
```diff
@@ -32,6 +32,17 @@ AWS permissions and attach it to the user, this is not the preferred option.
 The cluster will not have the ability to grant additional credentials to
 individual components, so the same credentials are used by all components.
 ====
++
+[IMPORTANT]
```
huffmanca: Reading this section I don't see where a token would be used here, although it's quite possibly my own ignorance in this regard. :) Checking the link on creating an IAM user in AWS I don't see reference to tokens, either.

Are we certain this section is relevant for creating an IAM user?

modules/installation-launching-installer.adoc
```diff
@@ -68,6 +68,17 @@ process uses to the installation program.
 .. If you do not have an AWS profile stored on your computer, enter the AWS
 access key ID and secret access key for the user that you configured to run the
 installation program.
++
+[IMPORTANT]
+====
+You must not use a temporary session token that you generated while using a
```
huffmanca: Would it be better to put this in the prerequisites where we discuss obtaining the access token?

huffmanca (commented):
I have no issues with the content, just a couple questions regarding its location.

modules/installation-aws-iam-user.adoc
kalexand-rh: I think it needs to be mentioned in this module, but it's more relevant in a later step.

modules/installation-launching-installer.adoc
kalexand-rh: It's not this access token. :)
I wish that they hadn't made the installer smart enough to generate your credentials for you - creating the credentials was briefly a separate step in its own module, and that's where it would make most sense. Supplying your AWS creds is not technically a prereq at all. I think your suggestion's good, though, so I'm moving this information to the assembly prereqs.

kalexand-rh:
@huffmanca +1'd the edits in Slack, and the original change was approved on the source bug. I'll let QE know that there were some edits, but I think this is fair to merge.

PR #16095, opened 2019-08-01
bug 1725079 updates based on more feedback

kalexand-rh:
From more comments on https://bugzilla.redhat.com/show_bug.cgi?id=1725079

### DIFF
--- a/modules/installation-bare-metal-config-yaml.adoc
+++ b/modules/installation-bare-metal-config-yaml.adoc
@@ -31,9 +31,9 @@ networking:
   serviceNetwork: <9>
   - 172.30.0.0/16
 platform:
-  none: {}
-pullSecret: '{"auths": ...}' <10>
-sshKey: 'ssh-ed25519 AAAA...' <11>
+  none: {} <10>
+pullSecret: '{"auths": ...}' <11>
+sshKey: 'ssh-ed25519 AAAA...' <12>
 ----
 <1> The base domain of the cluster. All DNS records must be sub-domains of this
 base and include the cluster name.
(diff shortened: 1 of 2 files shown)


mburke5678:
@kalexand-rh LGTM

kalexand-rh:
Thank you, @mburke5678! I'm going to make sure that QE's ok with the clarifications before I merge.

modules/installation-bare-metal-config-yaml.adoc
```diff
@@ -57,7 +57,8 @@ If you disable simultaneous multithreading, ensure that your capacity planning
 accounts for the dramatically decreased machine performance.
 ====
 <4> This parameter controls the number of workers that the cluster deploys for
-you, so you must set the number of workers to `0` in this file. Remember that you
+you. Because the cluster cannot provision workers for you,
+you must set the number of workers to `0` in this file. Remember that you
```
djuran: I still find this wording confusing. I understand the cluster doesn't provision the hardware but it still deploys OCP to any worker nodes you boot with the worker ignite file.

modules/installation-bare-metal-config-yaml.adoc
```diff
@@ -56,9 +56,11 @@ machines.
 If you disable simultaneous multithreading, ensure that your capacity planning
 accounts for the dramatically decreased machine performance.
 ====
-<4> This parameter controls the number of workers that the cluster deploys for
-you, so you must set the number of workers to `0` in this file. Remember that you
-must manually deploy at least one compute, or worker, machine in the cluster.
+<4> This parameter controls the number of workers that the cluster creates
```
kalexand-rh: @djuran, is this clearer?

djuran: How about. But please check if the below text is correct. I'm really new to this topic...

With the bare-metal scenario, the hardware is outside the control of the openshift controller and this parameter should be set to 0. 
With that said, the installer will still deploy worker nodes on the hardware using the worker ingition file.

kalexand-rh: No, you have to manually deploy the worker nodes by using the Ignition config that you generated by using the installation program. The cluster recognizes the workers that you deploy instead of deploying them in any way.

PR #16097, opened 2019-08-01
Add instructions to isolate compute workload

jkupferer:
The current instructions to add infra nodes is not complete without also having instructions describing how to isolate other compute workload to prevent it from running on the new infra nodes.

This PR adds documentation describing how to implement compute nodes with workload isolation using node selectors. While similar functionality could be achieved with taints and tolerations, it is significantly easier to implement in a secure manner with node-selectors. Current OpenShift administrators are also likely to be familiar and comfortable with how to manage node selectors.

### DIFF
--- a/machine_management/creating-infrastructure-machinesets.adoc
+++ b/machine_management/creating-infrastructure-machinesets.adoc
@@ -21,14 +21,21 @@ new MachineSet.
 include::modules/infrastructure-components.adoc[leveloffset=+1]
 
 [id="creating-infrastructure-machinesets-production"]
-== Creating infrastructure MachineSets for production environments
+== Creating infrastructure and compute MachineSets for production environments
 
 In a production deployment, deploy at least three MachineSets to hold
 infrastructure components. Both the logging aggregation solution and
 the service mesh deploy Elasticsearch, and Elasticsearch requires three
 instances that are installed on different nodes. For high availability, install
 deploy these nodes to different availability zones. Since you need different
-MachineSets for each availability zone, create at least three MachineSets.
+MachineSets for each availability zone, create at least three MachineSets with
+role "infra".
+
+Compute nodes are required to isolate workload that should not use the
+infrastructure nodes becaus the default worker machinesets do not produce node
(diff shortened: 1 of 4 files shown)


vikram-redhat:
Related to: https://bugzilla.redhat.com/show_bug.cgi?id=1733230 which I have closed as notabug because we do document taints.

If using node selectors is easier and is supported, we should add this to the docs. 

@kalexand-rh would you have the bandwidth to review? We can then get QE to confirm the described steps work.

kalexand-rh:
@vikram-redhat, I have some priority issues that I need to address, but I will try to move this one forward this week.

jkupferer:
Regarding, https://bugzilla.redhat.com/show_bug.cgi?id=1733230

Taints and toleration, these do not sufficiently isolate workload as currently users can add tolerations for taints including the ability to bypass restrictions and run app workload on the masters:

See: https://bugzilla.redhat.com/show_bug.cgi?id=1729242

jkupferer:
Fixed conflicts, poking this PR again...

vikram-redhat:
@jkupferer sorry for the delay.

@kalexand-rh any chance you could look at this?

vikram-redhat:
@jkupferer do you mind rebasing?

jkupferer:
@vikram-redhat, rebase complete.

jkupferer:
A topic that may require some discussion is the use of labels on machines and machinesets:
```
    machine.openshift.io/cluster-api-machine-role: <role>
    machine.openshift.io/cluster-api-machine-type: <role>
```

The use of these labels is based on the documentation found here: https://github.com/openshift/machine-api-operator

These labels are also set by machines created by the installer, for example:
https://github.com/openshift/installer/blob/master/pkg/asset/machines/openstack/machinesets.go#L51-L55
https://github.com/openshift/installer/blob/master/pkg/asset/machines/aws/machines.go#L61-L65

It isn't clear from those documentation what the specific purpose of these labels are, though it is clearly useful to label machinesets by role in some manner. If these specific labels are not appropriate, some other labels should be shown.

PR #16098, opened 2019-08-01
Fix release notes table for TP features

bobcallaway:
The data in the table appears in incorrect columns due to an extraneous | in the adoc


kalexand-rh:
Thanks @bobcallaway! I built the docs locally and confirmed that the table does look right now. Merging.

PR #16110, opened 2019-08-02
Changing kebab to use png

vikram-redhat:
@openshift/team-documentation @openshift/team-aligned-docs - FYI.

SVG doesn't render correctly on the customer portal.


jboxman:
Does the integration not copy the SVG file?

vikram-redhat:
@jboxman it does, but the portal displays it as enormous rectangles instead.

PR #16123, opened 2019-08-05
BZ-1737533: Added warning to hostnetwork SCC

bergerhoffer:

### DIFF
--- a/modules/security-context-constraints-about.adoc
+++ b/modules/security-context-constraints-about.adoc
@@ -47,6 +47,13 @@ The cluster contains eight default SCCs:
 * `hostaccess`
 * `hostmount-anyuid`
 * `hostnetwork`
++
+[WARNING]
+====
+If additional workloads are run on master hosts, use caution when providing
+access to `hostnetwork`. A workload that runs `hostnetwork` on a master host is
+effectively root on the cluster and must be trusted accordingly.
+====
 * `node-exporter`
 * `nonroot`
 * `privileged`


bergerhoffer:
@ravisantoshgudimetla Can you please review? I rearranged the wording - please let me know if we need to adjust it. Thanks!

Preview:
https://bz-1722775--ocpdocs.netlify.com/openshift-enterprise/latest/authentication/managing-security-context-constraints.html#security-context-constraints-about_configuring-internal-oauth

mburke5678:
@bergerhoffer LGTM

bergerhoffer:
Thanks @mburke5678!

modules/security-context-constraints-about.adoc
```diff
@@ -47,6 +47,13 @@ The cluster contains eight default SCCs:
 * `hostaccess`
 * `hostmount-anyuid`
 * `hostnetwork`
++
+[WARNING]
+====
+Use caution when allowing workloads onto master nodes when the `hostnetwork` SCC
```
ravisantoshgudimetla: This looks good to me. Perhaps we can add that `master nodes have instance-based credentials` for clarification

/cc @derekwaynecarr

bergerhoffer: Thanks @ravisantoshgudimetla I tried to work that in, wdyt of option 1:

```
Use caution when allowing workloads onto master nodes when the `hostnetwork` SCC
is granted. The entire cluster can be compromised if some pods are allowed to
land onto nodes that have instance-based credentials, such as master nodes.
```

or option 2:

```
Use caution when allowing workloads onto master nodes, which have instance-based
credentials, when the `hostnetwork` SCC is granted. The entire cluster can be
compromised if some pods are allowed to land onto nodes that have instance-based
credentials.
```

ravisantoshgudimetla: I'd prefer option 1. We should also mention that `the future release of OpenShift container platform, will move away from instance based credentials`

bergerhoffer: Thanks @ravisantoshgudimetla, I've updated with option 1. 

As for moving away from instance-based credentials, I think it'd be best to just update the docs once we actually do move away from them, as opposed to mentioning that now, if that's okay.

ravisantoshgudimetla: I think we should state the intent clearly. I don't know if it's a standard practice to not talk about things happening in future. If yes, let me know.

bergerhoffer: Hi @ravisantoshgudimetla. Yes we do typically try to avoid mentioning future predictions in the docs. 

For all we know, that decision might change or get pushed out, and it can be a maintenance issue in the docs. I've consulted a few other writers and confirmed that this is the approach we normally take. Let me know if you have any other questions on this!

derekwaynecarr: I recommend the following:

`If additional workloads are run on control plane hosts, use caution when providing access to host network.  A workload that runs host network on a control plane host is effectively root on the cluster and must be trusted accordingly.`

I think this is the safest recommendation independent of underlying Iaas.

bergerhoffer: Thanks @derekwaynecarr! I've update the warning with this wording. Can you double check that it looks okay? CC @ravisantoshgudimetla too.

modules/security-context-constraints-about.adoc
```diff
@@ -47,6 +47,13 @@ The cluster contains eight default SCCs:
 * `hostaccess`
 * `hostmount-anyuid`
 * `hostnetwork`
++
+[WARNING]
+====
+If additional workloads are run on master hosts, use caution when providing
+access to `hostnetwork`. A workload that runs `hostnetwork` on a master host is
+effectively root on the cluster and must be trusted accordingly.
```
ravisantoshgudimetla: This LGTM
/lgtm
@derekwaynecarr mind doing final review?

PR #16124, opened 2019-08-05
Bug 1731177 adding procedural documentation for router sharding

bmcelvee:
https://bugzilla.redhat.com/show_bug.cgi?id=1731177

### DIFF
--- a/modules/nw-ingress-sharding-namespace-labels.adoc
+++ b/modules/nw-ingress-sharding-namespace-labels.adoc
@@ -1,0 +1,54 @@
+// Module included in the following assemblies:
+//
+// * configuring_ingress_cluster_traffic/configuring-ingress-cluster-traffic-ingress-controller.adoc
+// * ingress-operator.adoc
+
+[id="nw-ingress-sharding-namespace-labels_{context}"]
+= Configuring ingress controller sharding by using namespace labels
+
+Ingress Controller sharding by using namespace labels means that the Ingress
+Controller serves any route in any namespace that is selected by the namespace
+selector.
+
+Ingress Controller sharding is useful when balancing incoming traffic load among
+a set of Ingress Controllers and when isolating traffic to a specific Ingress
+Controller. For example, company A goes to one Ingress Controller and company B
+to another.
+
+.Procedure
+
+. Edit the `router-internal.yaml` file:
++
+----
+# cat router-internal.yaml
+apiVersion: v1
+items:
+- apiVersion: operator.openshift.io/v1
+  kind: IngressController
+  metadata:
+    name: sharded
+    namespace: openshift-ingress-operator
+  spec:
+    domain: <apps-sharded.basedomain.example.net>
+    nodePlacement:
+      nodeSelector:
+        matchLabels:
+          node-role.kubernetes.io/worker: ""
+    routeSelector:
+      matchLabels:
+        type: sharded
+  status: {}
+kind: List
+metadata:
+  resourceVersion: ""
+  selfLink: ""
+----
+
+. Apply the Ingress Controller `router-internal.yaml` file:
++
+----
+# oc apply -f router-internal.yaml
+----
++
+The Ingress Controller selects routes in any namespace that is selected by the
+namespace selector that have the label `type: sharded`.
--- a/modules/nw-ingress-sharding-route-labels.adoc
+++ b/modules/nw-ingress-sharding-route-labels.adoc
@@ -1,0 +1,54 @@
+// Module included in the following assemblies:
+//
+// * configuring_ingress_cluster_traffic/configuring-ingress-cluster-traffic-ingress-controller.adoc
+// * ingress-operator.adoc
+
+[id="nw-ingress-sharding-route-labels_{context}"]
+= Configuring ingress controller sharding by using route labels
+
+Ingress Controller sharding by using route labels means that the the Ingress
+Controller serves any route in any namespace that is selected by the route
+selector.
+
+Ingress Controller sharding is useful when balancing incoming traffic load among
+a set of Ingress Controllers and when isolating traffic to a specific Ingress
+Controller. For example, company A goes to one Ingress Controller and company B
+to another.
+
+.Procedure
+
+. Edit the `router-internal.yaml` file:
++
+----
+# cat router-internal.yaml
+apiVersion: v1
+items:
+- apiVersion: operator.openshift.io/v1
+  kind: IngressController
+  metadata:
+    name: sharded
+    namespace: openshift-ingress-operator
+  spec:
+    domain: <apps-sharded.basedomain.example.net>
+    nodePlacement:
+      nodeSelector:
+        matchLabels:
+          node-role.kubernetes.io/worker: ""
+    routeSelector:
+      matchLabels:
+        type: sharded
+  status: {}
+kind: List
+metadata:
+  resourceVersion: ""
+  selfLink: ""
+----
+
+. Apply the Ingress Controller `router-internal.yaml` file:
++
+----
+# oc apply -f router-internal.yaml
+----
++
+The Ingress Controller selects routes in any namespace that have the label
+`type: sharded`.
--- a/modules/nw-using-ingress-and-routes.adoc
+++ b/modules/nw-using-ingress-and-routes.adoc
@@ -1,6 +1,6 @@
 // Module included in the following assemblies:
 //
-// * ingress/getting-traffic-cluster.adoc
+// * ingress/configuring-ingress-cluster-traffic-ingress-controller.adoc
 
 [id="nw-using-ingress-and-routes_{context}"]
 = Using Ingress Controllers and routes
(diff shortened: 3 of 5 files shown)


bmcelvee:
I pulled instructions from the blog and kbase article (linked in the BZ). Would the user generally want to update an existing .yaml or create a new one. As if I've included the procedure for updating an existing .yaml file. Additionally, if updating an existing file, would the user need to run `oc get` for the file? Thanks!

@Miciah @rcarrata PTAL

modules/nw-ingress-sharding.adoc
```diff
@@ -0,0 +1,53 @@
...
+Ingress Controllers that have permission to view all of the labels in all
+projects can select routes to admit based on the labels. This is called Ingress
+Controller sharding. This is useful when balancing incoming traffic load among a
+set of Ingress Controllers and when isolating traffic to a specific Ingress
+Controller. For example, company A goes to one Ingress Controller and company B
+to another.
+
+.Procedure
+
+. Edit the `router-internal.yaml`:
```
bergerhoffer: Either add "file" at the end or remove the "the".

modules/nw-ingress-sharding.adoc
```diff
@@ -0,0 +1,53 @@
...
++
+----
+# cat router-internal.yaml
```
bergerhoffer: Is this file already supposed to exist or are they supposed to create it with the following contents? Because all the "cat" command is doing is just printing the contents of the file. Maybe be a little more specific on what they're supposed to do?

modules/nw-ingress-sharding.adoc
```diff
@@ -0,0 +1,53 @@
...
+apiVersion: v1
+items:
+- apiVersion: operator.openshift.io/v1
+  kind: IngressController
+  metadata:
+    name: sharded
+    namespace: openshift-ingress-operator
+  spec:
+    domain: <apps-sharded.basedomain.example.net>
```
bergerhoffer: Not critical, but maybe a callout or two would be helpful here so users know what in here they need to configure for their specific needs?

bergerhoffer (commented):
Just a few comments/questions!

modules/nw-ingress-sharding.adoc
```diff
@@ -0,0 +1,53 @@
+// Module included in the following assemblies:
+//
+// * configuring_ingress_cluster_traffic/configuring-ingress-cluster-traffic-ingress-controller.adoc
+// * ingress-operator.adoc
+
+[id="nw-ingress-sharding_{context}"]
+= Configuring ingress controller sharding
+
+Ingress Controllers that have permission to view all of the labels in all
+projects can select routes to admit based on the labels. This is called Ingress
```
Miciah: The first sentence doesn't make sense: (1) the ingress controller needs to be able to list certain resources, not just their labels (whatever that even means); (2) the resources that an ingress controller needs to have permission to list are namespaces, services, endpoints, and routes; and (3) any ingress controller that the ingress operator creates has permissions to list all such resources in the cluster, so it is unnecessary and probably confusing to begin by talking about permissions.

modules/nw-ingress-sharding.adoc
```diff
@@ -0,0 +1,53 @@
...
+  selfLink: ""
+----
+
+. Apply the Ingress Controller `router-internal.yaml` file:
++
+----
+# oc apply -f router-internal.yaml
+----
++
+The Ingress Operator creates a `LoadBalancerService` type and a `ClusterIPService` type.
```
Miciah: The example `resource-internal.yaml` file defines a `LoadBalancerService`-type ingress controller, for which the operator creates a `LoadBalancer`-type Service resource and a `ClusterIP`-type Service.  However, that seems like extraneous detail; more to the point, the ingress controller will select routes (in any namespace) that have the label `type: sharded`.

Miciah (commented):
There are two types of sharding that the ingress operator supports: (1) sharding based on namespace labels (the ingress controller serves any route in any namespace that is selected by the namespace selector) and sharding based on route labels (the ingress controller serves any route, in any namespace, that is selected by the route selector).  The `router-internal.yaml` file defines an ingress controller that shards by route labels.  We should describe both options.  (Technically, one can create an ingress controller that has both a route selector and a namespace selector, so these two types of sharding are not mutually exclusive.)

bmcelvee:
@lihongan please take a look, thanks!

modules/nw-ingress-sharding-namespace-labels.adoc
```diff
@@ -0,0 +1,56 @@
...
+items:
+- apiVersion: operator.openshift.io/v1
+  kind: IngressController
+  metadata:
+    name: sharded
+    namespace: openshift-ingress-operator
+  spec:
+    domain: <apps-sharded.basedomain.example.net>
+    endpointPublishingStrategy:
+      type: LoadBalancerService
```
lihongan: I don't think we can configure `endpointPublishingStrategy` here, the ingress operator will handle it based on different platforms. Please delete the two lines.

modules/nw-ingress-sharding-route-labels.adoc
```diff
@@ -0,0 +1,56 @@
...
+items:
+- apiVersion: operator.openshift.io/v1
+  kind: IngressController
+  metadata:
+    name: sharded
+    namespace: openshift-ingress-operator
+  spec:
+    domain: <apps-sharded.basedomain.example.net>
+    endpointPublishingStrategy:
+      type: LoadBalancerService
```
lihongan: Same to previous one, the two lines can be removed.

bmcelvee:
Thanks, @lihongan! I removed those lines and am going to go ahead and merge. If we need any additional changes, I can address those in a follow-up PR.

modules/nw-ingress-sharding-namespace-labels.adoc
```diff
@@ -0,0 +1,54 @@
...
+  metadata:
+    name: sharded
+    namespace: openshift-ingress-operator
+  spec:
+    domain: <apps-sharded.basedomain.example.net>
+    nodePlacement:
+      nodeSelector:
+        matchLabels:
+          node-role.kubernetes.io/worker: ""
+    routeSelector:
```
Miciah: `routeSelector` should be `namespaceSelector` in this example.  https://github.com/openshift/openshift-docs/pull/16648 fixes the error.

modules/nw-ingress-sharding-namespace-labels.adoc
Miciah: The user _can_ specify `endpointPublishingStrategy`, but it is better to omit it, as you suggested, so that the operator will use an appropriate default.

Miciah: Sorry, I was assuming that the procedure was creating a new ingresscontroller, but it uses `oc apply`, which can create a new ingresscontroller _or_ update an existing one.

When creating a new ingresscontroller, `spec.endpointPublishingStrategy` can be specified (or left unspecified, to express a desire for the default value for the platform), and its value will determine the effective endpoint publishing strategy.  If the example is creating a new ingresscontroller, it is better to omit `spec.endpointPublishingStrategy`, as we both said earlier, so that the operator will use the platform default.  (A user who knows better can always add it to the example yaml.)

When updating an existing ingresscontroller, specifying `spec.endpointPublishingStrategy` does not change the effective endpoint publishing strategy (`spec.endpointPublishingStrategy` will have the specified value after the update, but `status.endpointPublishingStrategy`, which shows the effective value, will remain unchanged).  This is also true if `spec.endpointPublishingStrategy` is not specified—after the update, the ingresscontroller will not have `spec.endpointPublishingStrategy` set, but it will still have `status.endpointPublishingStrategy` set.  Having different values in `spec.endpointPublishingStrategy` and in `status.endpointPublishingStrategy` could be confusing.  Having `status.endpointPublishingStrategy` but not `spec.endpointPublishingStrategy` set could be confusing if the user set it when creating the ingresscontroller to override the default, but this scenario is less likely to be confusing, I think, so omitting `spec.endpointPublishingStrategy` from the example makes sense in this case too.

In either case, this procedure should clearly state whether it is creating a new ingresscontroller or updating an existing one (documenting both would be fine too).  For creating a new one, I would suggest using `oc create`, and for updating an existing one, I would suggest using `oc patch`, e.g.:
```shell
oc -n openshift-ingress-operator patch ingresscontroller/sharded --type=merge --patch='{"spec":{"routeSelector":{"matchLabels":{"type":"sharded"}}}}'
```
or
```shell
oc -n openshift-ingress-operator patch ingresscontroller/sharded --type=merge --patch='{"spec":{"nodeSelector":{"matchLabels":{"type":"sharded"}}}}'
```
Sharding can be disabled on an existing ingresscontroller by specifying `null`:
```shell
oc -n openshift-ingress-operator patch ingresscontroller/sharded --type=merge --patch='{"spec":{"routeSelector":null}}'
```
or
```shell
oc -n openshift-ingress-operator patch ingresscontroller/sharded --type=merge --patch='{"spec":{"nodeSelector":null}}'
```

PR #16163, opened 2019-08-07
Surface VMware MAC address allocation range

jboxman:
- https://bugzilla.redhat.com/show_bug.cgi?id=1730103

### DIFF
--- a/modules/installation-network-user-infra.adoc
+++ b/modules/installation-network-user-infra.adoc
@@ -113,3 +113,20 @@ A working configuration for the Ingress router is required for an
 {product-title} cluster. You must configure the Ingress router after the control
 plane initializes.
 ====
+
+ifeval::["{context}" == "installing-vsphere"]
+[discrete]
+== Ethernet adaptor hardware address requirements
+
+When provisioning VMs for the cluster, the ethernet interfaces configured for
+each VM must use a MAC address from the VMware Organizationally Unique
+Identifier (OUI) allocation ranges:
+
+* `00:05:69:00:00:00` to `00:05:69:FF:FF:FF`
+* `00:0c:29:00:00:00` to `00:0c:29:FF:FF:FF`
+* `00:1c:14:00:00:00` to `00:1c:14:FF:FF:FF`
+* `00:50:56:00:00:00` to `00:50:56:FF:FF:FF`
+
+If a MAC address outside the VMware OUI is used, the cluster installation will
+not succeed.
+endif::[]


jboxman:
@kalexand-rh, does placing the VMware required ethernet hardware address content here make sense?

modules/installation-network-user-infra.adoc
```diff
@@ -111,3 +111,12 @@ A working configuration for the Ingress router is required for an
 {product-title} cluster. You must configure the Ingress router after the control
 plane initializes.
 ====
+
+[discrete]
+== Ethernet adaptor hardware address requirements
```
kalexand-rh: @jboxman, this module's used in the bare metal assembly, too. You need to conditionalize this for the vSphere assembly or make it generic.

modules/installation-network-user-infra.adoc
```diff
@@ -111,3 +111,12 @@ A working configuration for the Ingress router is required for an
...
+
+When provisioning VMs for the cluster, the ethernet interfaces configured for
```
kalexand-rh: @dav1x, does this sound right to you, and would it be more meaningful in the section about creating VMs instead of setting up the network interfaces? (If it's more meaningful here, I think we need to clarify if the ethernet interfaces are really for each VM or if they're for a larger component.)

jboxman:
@kalexand-rh, I added a conditional so the content is only exposed for vSphere.

dav1x:
@kalexand-rh This looks great! Thanks for the update. The cloud provider searches for the following OUI ranges beyond just 00:50:56, perhaps we should add them as well:

https://github.com/kubernetes/legacy-cloud-providers/blob/master/vsphere/vsphere.go#L550-L555

```
		"00:05:69": true,
		"00:0c:29": true,
		"00:1c:14": true,
		"00:50:56": true,
```

modules/installation-network-user-infra.adoc
dav1x: This sounds right.

dav1x:
> @dav1x, I've added the additional ranges. Is `...:3F:FF:FF` is the proper range end for the other three?

`...FF:FF:FF`

jboxman:
@dav1x @kalexand-rh 

I updated the MAC address range endings.

jboxman:
@dav1x @kalexand-rh 

If everyone is comfortable with this change, I can merge it. Let me know, thanks!

dav1x:
LGTM! Let it fly

PR #16167, opened 2019-08-07
Add install-config recovery note for bug 1728865

maxwelldb:
WIP. 

https://bugzilla.redhat.com/show_bug.cgi?id=1728865

### DIFF
--- a/modules/installation-initializing.adoc
+++ b/modules/installation-initializing.adoc
@@ -62,6 +62,7 @@ it to install multiple clusters.
 +
 [IMPORTANT]
 ====
-The `install-config.yaml` file is consumed during the installation process. If
-you want to reuse the file, you must back it up now.
+The `install-config.yaml` file is consumed during the installation process.
+If you want to reuse the file, back it up now. If you lose the file, you can recover it
+by running `openshift-install create install-config`.
 ====


maxwelldb (commented):
@bostrt Thought on clarifying this language. How about: 

>The `install-config.yaml` file is consumed during the installation process.
If you want to reuse the file, back it up now. If you lose the file, you can recover it
by running `openshift-install create install-config`.

bostrt:
Looks good to me :)

maxwelldb:
@kalexand-rh How do you feel about this addition to the install docs?

modules/installation-initializing.adoc
```diff
@@ -62,6 +62,7 @@ it to install multiple clusters.
 +
 [IMPORTANT]
 ====
-The `install-config.yaml` file is consumed during the installation process. If
-you want to reuse the file, you must back it up now.
+The `install-config.yaml` file is consumed during the installation process.
+If you want to reuse the file, back it up now. If you lose the file, you can recover it
```
kalexand-rh: Honestly, I wouldn't add this information. This isn't the point where a user would actually need to know this command, and we've told them twice to back it up. I'd defer the bug until we add a troubleshooting section. (@vikram-redhat, FYI)

maxwelldb: I concur.

vikram-redhat: The note is already there, so I guess there is nothing to do for this bug?

kalexand-rh: Not at this time. If you're collecting issues for a troubleshooting section, this would be a lovely addition to that list.

maxwelldb: @vikram-redhat What's the best way to track a future troubleshooting entry, then? I close this after that's tracked elsewhere.

maxwelldb: @vikram-redhat ^^^ ?

maxwelldb:
Won't merge. This is for a troubleshooting guide.

PR #16175, opened 2019-08-08
DEVEXP-388: Adding note on S3 bucket security configuration during UPI.

ricardomaraschini:
Adds a WARNING refering user to AWS documentation regarding S3 bucket security configuration. During IPI these security measures are automatically taken.

### DIFF
--- a/modules/registry-configuring-storage-aws-user-infra.adoc
+++ b/modules/registry-configuring-storage-aws-user-infra.adoc
@@ -36,3 +36,9 @@ storage:
     bucket: <bucket-name>
     region: <region-name>
 ----
+
+[WARNING]
+====
+To secure your registry images in AWS, link:https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-publicaccessblockconfiguration.html[block public access]
+to the S3 bucket.
+====


modules/registry-configuring-storage-aws-user-infra.adoc
```diff
@@ -36,3 +36,11 @@ storage:
     bucket: <bucket-name>
     region: <region-name>
 ----
+
+[WARNING]
+====
```
bmcelvee: I would suggest removing the link to the blog, and simplifying the warning as follows: 

[WARNING]
====
To secure your registry images in AWS, link:https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-publicaccessblockconfiguration.html[block public access]
to the S3 bucket. 
====

bmcelvee:
Thanks, @ricardomaraschini! I left a suggestion in-line. According to the Jira, this looks like it's only applicable to 4.2. Is that correct?

modules/registry-configuring-storage-aws-user-infra.adoc
ricardomaraschini: Done.

ricardomaraschini:
@bmcelvee I would say it applies to all versions. Let me check and i will get back to you.

ricardomaraschini:
@bmcelvee Just 4.2, though not bad to add to 4.1 docs as well as lots of customers set up the registry on their own.

bmcelvee:
Thank you! I'll go ahead and merge + cherrypick to 4.1 and 4.2.

This is a minor change (addition of a warning) and should not require QE review.

PR #16176, opened 2019-08-08
CNV - kubevirt-hyperconverged namespace changed to openshift-cnv

aburdenthehand:
For [CNV-2474](https://jira.coreos.com/browse/CNV-2474)
`kubevirt-hyperconverged` namespace/project replaced with `openshift-cnv` in the CNV docs. 
This change does not impact the catalog, custom resource name, or the UI buttons for these objects.

### DIFF
--- a/cnv/cnv_install/uninstalling-container-native-virtualization.adoc
+++ b/cnv/cnv_install/uninstalling-container-native-virtualization.adoc
@@ -19,7 +19,7 @@ include::modules/cnv-deleting-hco-subscription.adoc[leveloffset=+1]
 
 [NOTE]
 ====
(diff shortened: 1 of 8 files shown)


rthallisey:
👍 +1

alexxa:
@aburdenthehand I don't see any cnv target version on this PR or that Jira ticket. What version is this PR for? We shouldn't change release docs 2.0. It's 2.1 change. And for testing of upgrade from 2.0 to 2.1 we will change ns of 2.0 deployment, but just for testing, AFAIK.

aburdenthehand:
@alexxa - This is a 2.1 change. It will be cherrypicked to openshift 4.2 and not 4.1 so will not affect 2.0 docs.

PR #16181, opened 2019-08-08
Update Ignition spec documentation link

sdodson:


sdodson:
I'm not certain if we should be linking here or not. But if we do we should make an effort to maintain the content without moving it.

modules/ignition-config-viewing.adoc
```diff
@@ -81,7 +81,7 @@ config from the bootstrap machine:
 Here are a few things you can learn from the `bootstrap.ign` file: +
 
 * Format: The format of the file is defined in the
-https://github.com/coreos/ignition/tree/spec2x[Ignition config spec].
+https://github.com/coreos/ignition/blob/spec2x/doc/configuration-v2_3.md[Ignition config spec].
```
kalexand-rh: @sferich888, do you want us to stabilize this link or remove it?

ajeddeloh: Isn't Openshift using spec 2.2.0? This should link to whichever version Openshift generates.

sdodson: https://github.com/openshift/openshift-docs/blame/master/modules/rhcos-about.adoc#L124 mentions 2.3

ajeddeloh: Gotcha. MCO is still using 2.2.0 which means the generated configs will all be 2.2.0. Ignition can handle either just fine.

sferich888: @chrisnegus don't we have docs in product we could use here?

chrisnegus: @sferich888 If you are asking if we publish the whole Ignition configuration spec in the docs, we don't as far as I know. That's why I added the link. If the version is too hard to keep up with, we could drop the link or provide a more generic one. Maybe have them run a "oc" command to see the current version, then look up the spec themselves on the Ignition github site.

sferich888: I don't think we want to push people to GitHub for these docs, can/should we create ignition docs for OpenShift?

chrisnegus: @sferich888 I don't know. I was simply trying to promote a level of understanding of how cool Ignition is. To go beyond that, I'd expect we would want a clearer message on how and to what extent we want to encourage people to modify Ignition configs themselves.

sferich888: I agree and if we don't have that we should keep ignition as a black box to our users. Thus no need for the link

sdodson: I came across this because I was looking at the level of Ignition understanding the documentation assumes and conveys. At this point it seems fine to me to consider it a black box.

kalexand-rh: > ... modifying Kubernetes and Ignition config files is not supported unless you are following documented procedures or are instructed to do so by Red Hat support.

When we get to the point where we need to treat Ignition as more than a black box, we'll need to update that language in the installation concepts and be clear about the changes that are and aren't supported.

chrisnegus: @kalexand-rh That text looks good.

chrisnegus (commented):
I look forward to a larger discussion of how we want to document ignition.

vikram-redhat:
@chrisnegus @sdodson can we merge this?

sdodson:
> @chrisnegus @sdodson can we merge this?

My opinion is yes.

kalexand-rh:
https://github.com/openshift/openshift-docs/pull/16181#discussion_r312300955

Eric said no, though.

vikram-redhat:
Closing as per https://github.com/openshift/openshift-docs/pull/16181#issuecomment-550400610

PR #16183, opened 2019-08-08
Add content for EgressNetworkPolicy

jboxman:
- OSDOCS-648

Bring content forward from 3.11 for `EgressNetworkPolicy`. This applies only to OpenShift SDN ~as OVN uses the `NetworkPolicy` CR instead for this functionality~.

### DIFF
--- a/modules/nw-egressnetworkpolicy-about.adoc
+++ b/modules/nw-egressnetworkpolicy-about.adoc
@@ -1,0 +1,72 @@
+// Module included in the following assemblies:
+//
+// * networking/openshift-sdn/configuring-egress-firewall.adoc
+
+[id="nw-egressnetworkpolicy-about_{context}"]
+= How an egress firewall works in a project
+
+As an {product-title} cluster administrator, you can use an _egress firewall_ to
+limit the external hosts that some or all Pods can access from within the
+cluster. An egress firewall supports the following scenarios:
+
+- A Pod can only connect to internal hosts and cannot initiate connections to
+the public Internet.
+- A Pod can only connect to the public Internet and cannot initiate connections
+to internal hosts that are outside the {product-title} cluster.
+- A Pod cannot reach specified internal subnets or hosts outside the {product-title} cluster.
+- A Pod can connect to only specific external hosts.
+
+You configure an egress firewall policy by creating an EgressNetworkPolicy Custom Resource (CR) object and specifying an IP address range in CIDR format or by specifying a DNS name.
+For example, you can allow one project access to a specified IP range but deny the same access to a different project. Or you can restrict application developers from updating from Python pip mirrors, and force updates to come only from approved sources.
+
+[IMPORTANT]
+====
+You must have OpenShift SDN configured to use either the network policy or multitenant modes to configure egress firewall policy.
+
+If you use network policy mode, egress policy is compatible with only one policy per namespace and will not work with projects that share a network, such as global projects.
+====
+
+[CAUTION]
+====
+Egress firewall rules do not apply to traffic that goes through routers. Any user with permission to create a Route CR object can bypass egress network policy rules by creating a route that points to a forbidden destination.
+====
+
+[id="limitations-of-an-egress-firewall_{context}"]
+== Limitations of an egress firewall
+
+An egress firewall has the following limitations:
+
+* No project can have more than one EgressNetworkPolicy object.
+
+* The `default` project cannot use egress network policy.
+
+* When using the OpenShift SDN network provider in multitenant mode, the following limitations apply:
+
+  - Global projects cannot use an egress firewall. You can make a project global by using the `oc adm pod-network make-projects-global` command.
+
+  - Projects merged by using the `oc adm pod-network join-projects` command cannot use an egress firewall in any of the joined projects.
+
+Violating any of these restrictions results in broken egress network policy for the project, and may cause all external network traffic to be dropped.
+
+[id="policy-rule-order_{context}"]
+== Matching order for egress network policy rules
+
+The egress network policy rules are evaluated in the order that they are defined, from first to last. The first rule that matches an egress connection from a Pod applies. Any subsequent rules are ignored for that connection.
+
+[id="domain-name-server-resolution_{context}"]
+== How Domain Name Server (DNS) resolution works
+
+If you use DNS names in any of your egress firewall policy rules, proper resolution of the domain names is subject to the following restrictions:
+
+* Domain name updates are polled based on the TTL (time to live) value of the domain returned by the local non-authoritative servers.
+
+* The Pod must resolve the domain from the same local name servers when necessary. Otherwise the IP addresses for the domain known by the egress firewall controller and the Pod can be different. If the IP addresses for a host name differ, the egress firewall might not be enforced consistently.
+
+* Because the egress firewall controller and Pods asynchronously poll the same local name server, the Pod might obtain the updated IP address before the egress controller does, which causes a race condition. Due to this current limitation, domain name usage in EgressNetworkPolicy objects is only recommended for domains with infrequent IP address changes.
+
+[NOTE]
+====
+The egress firewall always allows Pods access to the external interface of the node that the Pod is on for DNS resolution.
+
+If you use domain names in your egress firewall policy and your DNS resolution is not handled by a DNS server on the local node, then you must add egress firewall rules that allow access to your DNS server’s IP addresses. if you are using domain names in your Pods.
+====
--- a/modules/nw-egressnetworkpolicy-create.adoc
+++ b/modules/nw-egressnetworkpolicy-create.adoc
@@ -1,0 +1,42 @@
+// Module included in the following assemblies:
+//
+// * networking/openshift-sdn/configuring-egress-firewall.adoc
+
+[id="nw-networkpolicy-create_{context}"]
+= Creating an egress firewall policy object
+
+As a {product-title} cluster administrator, you can create an egress firewall policy object for a project.
+
+[IMPORTANT]
+====
+If the project already has an EgressNetworkPolicy object defined, you must edit the existing policy to make changes to the egress firewall rules.
+====
+
+.Prerequisites
+
+* A cluster that uses the OpenShift SDN network provider plug-in.
+* Install the OpenShift Command-line Interface (CLI), commonly known as `oc`.
+* You must log in to the cluster as a cluster administrator.
+
+.Procedure
+
+. Create a policy rule:
+.. Create a `<policy-name>.yaml` file where `<policy-name>` describes the egress
+policy rules.
+.. In the file you created, define an egress policy object.
+
+. Enter the following command to create the policy object:
++
+----
+$ oc create -f <policy-name>.yaml -n <project>
+----
++
+In the following example, a new EgressNetworkPolicy object is created in a
+project named `project1`:
++
+----
+$ oc create -f default-rules.yaml -n project1
+egressnetworkpolicy.network.openshift.io/default-rules created
+----
++
+. Optional: Save the `<policy-name>.yaml` so that you can make changes later.
--- a/modules/nw-egressnetworkpolicy-delete.adoc
+++ b/modules/nw-egressnetworkpolicy-delete.adoc
@@ -1,0 +1,31 @@
+// Module included in the following assemblies:
+//
+// * networking/openshift-sdn/removing-egress-firewall.adoc
+
+[id="nw-egressnetworkpolicy-delete_{context}"]
+
+= Removing an EgressNetworkPolicy object
+
+As a {product-title} cluster administrator, you can remove an egress firewall from a project.
+
+.Prerequisites
+
+* A cluster using the OpenShift SDN network plug-in.
+* Install the OpenShift Command-line Interface (CLI), commonly known as `oc`.
+* You must log in to the cluster as a cluster administrator.
+
+.Procedure
+
+To remove an egress network policy object for a project, complete the following steps:
+
+. Find the name of the EgressNetworkPolicy object for the project. Replace `<project>` with the name of the project.
++
+----
+$ oc get -n <project> egressnetworkpolicy
+----
+
+. Enter the following command to delete the EgressNetworkPolicy object. Replace `<project>` with the name of the project and `<name>` with the name of the object.
++
+----
+$ oc delete -n <project> egressnetworkpolicy <name>
+----
--- a/modules/nw-egressnetworkpolicy-object.adoc
+++ b/modules/nw-egressnetworkpolicy-object.adoc
@@ -1,0 +1,70 @@
+// Module included in the following assemblies:
+//
+// * networking/openshift-sdn/configuring-egress-firewall.adoc
+
+[id="nw-egressnetworkpolicy-object_{context}"]
+= EgressNetworkPolicy custom resource (CR) object
+
+The following YAML describes an EgressNetworkPolicy CR object:
+
+[source,yaml]
+----
+kind: EgressNetworkPolicy
+apiVersion: v1
+metadata:
+  name: <name> <1>
+spec:
+  egress: <2>
+    ...
+----
+<1> Specify a `name` for your egress firewall policy.
+
+<2> Specify a collection of one or more egress network policy rules as described in the following section.
+
+[id="egressnetworkpolicy-rules_{context}"]
+== EgressNetworkPolicy rules
+
+The following YAML describes an egress firewall rule object. The `egress` key expects an array of one or more objects.
+
+[source,yaml]
+----
+egress:
+- type: <type> <1>
+  to: <2>
+    cidrSelector: <cidr> <3>
+    dnsName: <dns-name> <4>
+----
+<1> Specify the type of rule. The value must be either `Allow` or `Deny`.
+
+<2> Specify a value for either the `cidrSelector` key or the `dnsName` key for the rule. You cannot use both keys in a rule.
+
+<3> Specify an IP address range in CIDR format.
+
+<4> Specify a domain name.
+
+[id="egressnetworkpolicy-example_{context}"]
+== Example EgressNetworkPolicy CR object
+
+The following example defines several egress firewall policy rules:
+
+[source,yaml]
+----
+kind: EgressNetworkPolicy
+apiVersion: v1
+metadata:
+  name: default-rules <1>
+spec:
+  egress: <2>
+  - type: Allow
+    to:
+      cidrSelector: 1.2.3.0/24
+  - type: Allow
+    to:
+      dnsName: www.example.com
+  - type: Deny
+    to:
+      cidrSelector: 0.0.0.0/0
+----
+<1> The name for the policy object.
+
+<2> A collection of egress firewall policy rule objects.
--- a/modules/nw-egressnetworkpolicy-view.adoc
+++ b/modules/nw-egressnetworkpolicy-view.adoc
@@ -1,0 +1,33 @@
+// Module included in the following assemblies:
+//
+// networking/openshift-sdn/configuring-egressnetworkpolicy.adoc
+
+[id="nw-egressnetworkpolicy-view_{context}"]
+
+= Viewing NetworkPolicy objects
+
+You can list the EgressNetworkPolicy objects in your cluster.
+
+.Prerequisites
+
+* A cluster using the OpenShift SDN network plug-in.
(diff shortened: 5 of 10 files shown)


jboxman:
Also:

> Weibin Liang 16:13
jboxman below testing notes are what i got when testing v3.x, not sure they are still in v4.1, danwinship can confirm it,
• You must have the ovs-multitenant or network-policy plug-in enabled in order to limit pod access via egress policy. The default project (and any other global namespace) cannot have egress policy.
   • The policy will not effect the traffic between pods.
   • One project only create one policy.

jboxman:
Hi @weliang1, can you validate the content for this PR or let me know who I can ask? Thanks!

* [Configuring an egress firewall for a project](https://osdocs-648--ocpdocs.netlify.com/openshift-enterprise/latest/networking/openshift-sdn/configuring-egress-firewall.html)
* [Editing an egress firewall for a project](https://osdocs-648--ocpdocs.netlify.com/openshift-enterprise/latest/networking/openshift-sdn/editing-egress-firewall.html)
* [Removing an egress firewall for a project](https://osdocs-648--ocpdocs.netlify.com/openshift-enterprise/latest/networking/openshift-sdn/removing-egress-firewall.html)

modules/nw-egressnetworkpolicy-about.adoc
```diff
@@ -0,0 +1,70 @@
+// Module included in the following assemblies:
+//
+// * networking/openshift-sdn/configuring-egress-firewall.adoc
+
+[id="nw-egressnetworkpolicy-about_{context}"]
+= How an egress firewall works in a project
+
+As an {product-title} cluster administrator, you can use an _egress firewall_ to
+limit the external IP addresses that some or all Pods can access from within the
```
knobunc: You can filter by dns name too... so perhaps "external hosts" rather than IP addresses?

modules/nw-egressnetworkpolicy-about.adoc
```diff
@@ -0,0 +1,70 @@
...
+cluster. Egress firewall policy supports the following scenarios:
+
+- A Pod can only connect to internal hosts, and cannot initiate connections to
+the public Internet.
+- A Pod can only connect to the public Internet, and cannot initiate connections
+to internal hosts that are outside the {product-title} cluster.
+- A Pod cannot reach specified internal subnets or hosts outside the {product-title} cluster that should be unreachable.
```
knobunc: I suspect another common case will be:

- A pod can only connect to certain specified outside hosts, and cannot initiate connections to any other external hosts.

modules/nw-egressnetworkpolicy-about.adoc
```diff
@@ -0,0 +1,70 @@
...
+
+An egress firewall policy is configured by creating an EgressNetworkPolicy Custom Resource (CR) object and specifying an IP address range in CIDR format or by specifying a DNS name.
+For example, you can allow one project access to a specified IP range but deny the same access to a different project, or you can restrict application developers from updating from Python pip mirrors, and force updates to come only from approved sources.
+
+[IMPORTANT]
+====
+You must have OpenShift SDN configured to use either the network policy or multitenant modes to configure egress firewall policy.
+
+If you are using network policy mode, egress policy is compatible with only one policy per project, and will not work with projects that share a network, such as global projects.
```
knobunc: "only one policy per project" is that egress policy or network policy?  I assume egress.

modules/nw-egressnetworkpolicy-about.adoc
```diff
@@ -0,0 +1,70 @@
...
+====
+
+[CAUTION]
+====
+Exposing services by creating routes will ignore egress network policies.
```
knobunc: "by creating routes" with custom, external, endpoints "will ignore ..."

The trick is that you could make a service without a selector (https://kubernetes.io/docs/concepts/services-networking/service/#services-without-selectors) and then create endpoint objects for it with external IP addresses.

Then you hit the service, and, because if the iptables rules kube-proxy installs, the kernel magically sends the traffic to the external resource, bypassing the egress firewall.

The advice is to disable creation of routes (or perhaps endpoint objects instead, if that's possible) where that's a concern.

modules/nw-egressnetworkpolicy-about.adoc
weliang1: Tested both multitenet and network-policy mode in 4.3.0-rc.3, more than one egress policies can be created per project.

modules/nw-egressnetworkpolicy-about.adoc
```diff
@@ -0,0 +1,70 @@
...
+Egress network policy service endpoint filtering is done at the node `kube-proxy`.
+When the router is involved, `kube-proxy` is bypassed and egress network policy enforcement is not applied.
+====
+
+[id="limitations-of-egress-firewall"]
+== Limitations of egress firewall
+
+An egress firewall has the following limitations:
+
+* No project may have more than one EgressNetworkPolicy object.
```
weliang1: Tested both multitenet and network-policy mode in 4.3.0-rc.3, more than one egress network policies can be created per project.

juanluisvaladas: That's unexpected behavior... Can you please provide a yaml with the objects?

weliang1: @juanluisvaladas  Test log including policy yaml files: http://pastebin.test.redhat.com/831123

juanluisvaladas: Definitely it's not the correct behavior, I think the documentation should be merged anyway...

modules/nw-egressnetworkpolicy-about.adoc
danwinship: @weliang1 you can *create* more than one EgressNetworkPolicy, but if you do, the expected behavior is that the sdn pods will log an error and all egress traffic from that namespace will be dropped. There ought to be QE tests confirming this.
(We maybe ought to think about having an admission controller so we can actually stop people from creating the conflicting policies...)

danwinship: (ie, to clarify, the docs should say you can only create one ENP per namespace. We don't enforce that at object-creation time, so people can screw up their cluster by creating more than one.)

modules/nw-egressnetworkpolicy-about.adoc
danwinship: "routes" here refers to `Route` objects. You can't actually hack around ENP with selectorless services because we (openshift-sdn) control kube-proxy and can stop it from creating iptables rules for any service that would allow bypassing ENP. However, we don't control the routers, so if you make a `Route` that points to an external IP, you can use that bypass ENP.

modules/nw-egressnetworkpolicy-about.adoc
danwinship: Clarified above: we don't enforce this at object creation time, so you can create multiple ENPs, but it won't do what you wanted it to (and will cause the sdn pods to spew errors)

modules/nw-egressnetworkpolicy-about.adoc
jboxman: I'll need to find a way to reword this, so it makes more sense. As written, I'm not sure what we were trying to express.

jboxman: @danwinship, how about:

> Egress network policy rules do not apply to service endpoint traffic. An Ingress Controller handles traffic for a service, bypassing the egress firewall.

modules/nw-egressnetworkpolicy-about.adoc
```diff
@@ -0,0 +1,72 @@
...
+
+As an {product-title} cluster administrator, you can use an _egress firewall_ to
+limit the external hosts that some or all Pods can access from within the
+cluster. Egress firewall policy supports the following scenarios:
+
+- A Pod can only connect to internal hosts, and cannot initiate connections to
+the public Internet.
+- A Pod can only connect to the public Internet, and cannot initiate connections
+to internal hosts that are outside the {product-title} cluster.
+- A Pod cannot reach specified internal subnets or hosts outside the {product-title} cluster that should be unreachable.
```
kalexand-rh: ~that should be unreachable~

modules/nw-egressnetworkpolicy-about.adoc
```diff
@@ -0,0 +1,72 @@
...
+- A Pod can connect to only specific external hosts.
+
+An egress firewall policy is configured by creating an EgressNetworkPolicy Custom Resource (CR) object and specifying an IP address range in CIDR format or by specifying a DNS name.
```
kalexand-rh: I might make this sentence active.

modules/nw-egressnetworkpolicy-about.adoc
```diff
@@ -0,0 +1,72 @@
...
+For example, you can allow one project access to a specified IP range but deny the same access to a different project, or you can restrict application developers from updating from Python pip mirrors, and force updates to come only from approved sources.
```
kalexand-rh: I might split this sentence.

modules/nw-egressnetworkpolicy-about.adoc
```diff
@@ -0,0 +1,72 @@
...
+
+[IMPORTANT]
+====
+You must have OpenShift SDN configured to use either the network policy or multitenant modes to configure egress firewall policy.
+
+If you are using network policy mode, egress policy is compatible with only one policy per namespace, and will not work with projects that share a network, such as global projects.
```
kalexand-rh: s/are using/use

s/namespace,/namespace

modules/nw-egressnetworkpolicy-about.adoc
```diff
@@ -0,0 +1,72 @@
...
+====
+
+[CAUTION]
+====
+Egress network policy rules do not apply to service endpoint traffic. An Ingress Controller handles traffic for a service, bypassing the egress firewall.
+====
+
+[id="limitations-of-an-egress-firewall"]
```
kalexand-rh: add `_{context}` to all of the anchors for the headings in the modules.

modules/nw-egressnetworkpolicy-about.adoc
```diff
@@ -0,0 +1,72 @@
...
+== Limitations of an egress firewall
+
+An egress firewall has the following limitations:
+
+* No project can have more than one EgressNetworkPolicy object.
+
+* The `default` project cannot use egress network policy.
+
+* When using the OpenShift SDN network provider in multitenant mode the following limitations apply:
```
kalexand-rh: mode,

modules/nw-egressnetworkpolicy-about.adoc
```diff
@@ -0,0 +1,72 @@
...
+
+  - Any project that has been made global via `oc adm pod-network make-projects-global` cannot use egress network policy.
```
kalexand-rh: Maybe

> Global projects cannot use egress network policy. You can make a project global by using the `oc adm pod-network make-projects-global` command.

modules/nw-egressnetworkpolicy-about.adoc
```diff
@@ -0,0 +1,72 @@
+// Module included in the following assemblies:
+//
+// * networking/openshift-sdn/configuring-egress-firewall.adoc
+
+[id="nw-egressnetworkpolicy-about_{context}"]
+= How an egress firewall works in a project
+
+As an {product-title} cluster administrator, you can use an _egress firewall_ to
+limit the external hosts that some or all Pods can access from within the
+cluster. Egress firewall policy supports the following scenarios:
```
kalexand-rh: Is it the "egress firewall" or "egress network?" I feel like you need an article before egress.

modules/nw-egressnetworkpolicy-about.adoc
```diff
@@ -0,0 +1,72 @@
...
+
+* No project can have more than one EgressNetworkPolicy object.
+
+* The `default` project cannot use egress network policy.
+
+* When using the OpenShift SDN network provider in multitenant mode the following limitations apply:
+
+  - Any project that has been made global via `oc adm pod-network make-projects-global` cannot use egress network policy.
+
+  - Projects merged together via `oc adm pod-network join-projects` cannot use egress network policy in any of the joined projects.
```
kalexand-rh: s/together via  `oc adm pod-network join-projects`/ by using the  `oc adm pod-network join-projects` command

modules/nw-egressnetworkpolicy-about.adoc
```diff
@@ -0,0 +1,72 @@
...
+
+Violating any of these restrictions results in broken egress network policy for the project, and may cause all external network traffic to be dropped.
+
+[id="policy-rule-order"]
```
kalexand-rh: _{context}

modules/nw-egressnetworkpolicy-about.adoc
```diff
@@ -0,0 +1,72 @@
...
+== Matching order for egress network policy rules
+
+The egress network policy rules are evaluated in the order that they are defined, from first to last. The first rule that matches an egress connection from a Pod applies. Any subsequent rules are ignored for that connection.
+
+[id="domain-name-server-resolution"]
+== How Domain Name Server (DNS) resolution works
+
+If you use DNS names in any of your egress network policy rules, proper resolution of the domain names is subject to the following:
```
kalexand-rh: following restrictions? following guidelines? You need a noun.

modules/nw-egressnetworkpolicy-about.adoc
```diff
@@ -0,0 +1,72 @@
...
+
+* Domain name updates are polled based on the TTL (time to live) value of the domain returned by the local non-authoritative servers.
+
+* The Pod should resolve the domain from the same local name servers when necessary, otherwise the IP addresses for the domain known by the egress network policy controller and the Pod will be different, and the egress network policy may not be enforced as expected.
```
kalexand-rh: s/should/must

This sentence is too long.

s/may/might

modules/nw-egressnetworkpolicy-about.adoc
```diff
@@ -0,0 +1,72 @@
...
+
+* The egress network policy controller and Pods asynchronously poll the same local name server. There can be a race condition where the Pod may get the updated IP before the egress controller. Due to this current limitation, domain name usage in EgressNetworkPolicy objects is only recommended for domains with infrequent IP address changes.
```
kalexand-rh: > ! there is, there are v
Avoid at the beginning of an independent clause. In most cases rewrite to remove the phrase.

I'd combine the first two sentences, like

> Because the egress network policy controller and Pods asynchronously poll the same local name server, the Pod might obtain the updated IP address before the egress controller does, which causes a race condition.

s/may get/gets

modules/nw-egressnetworkpolicy-about.adoc
```diff
@@ -0,0 +1,72 @@
...
+
+[NOTE]
+====
+The egress firewall always allows Pods access to the external interface of the node that the Pod is on for DNS resolution.
+If your DNS resolution is not handled by something on the local node, then you will need to add egress firewall rules allowing access to the DNS server’s IP addresses if you are using domain names in your Pods.
```
kalexand-rh: Can we be more specific than "something?"

s/will need to/must
s/allowing/that allow

Wait, does this only apply if your DNS resolution is not handled by something on the local node and if you are using domain names in your Pods? If so, put both of these restrictions in the first phrase.

modules/nw-egressnetworkpolicy-create.adoc
```diff
@@ -0,0 +1,42 @@
...
+As a {product-title} cluster administrator, you can create an egress firewall policy object for a project.
+
+[IMPORTANT]
+====
+If the project already has an EgressNetworkPolicy object defined, you must edit the existing policy to make changes to the egress firewall rules.
+====
+
+.Prerequisites
+
+* A cluster using the OpenShift SDN network provider plug-in.
```
kalexand-rh: s/using/that uses

modules/nw-egressnetworkpolicy-create.adoc
```diff
@@ -0,0 +1,42 @@
...
+* Install the OpenShift Command-line Interface (CLI), commonly known as `oc`.
+* You must log in to the cluster as a cluster administrator.
+
+.Procedure
+
+. Create a policy rule:
+.. Create a `<policy-name>.yaml` file where `<policy-name>` describes the egress
+policy rules.
+.. In the file you just created define an egress policy object.
```
kalexand-rh: s/just created/created,

modules/nw-egressnetworkpolicy-create.adoc
```diff
@@ -0,0 +1,42 @@
...
++
+In the following example, a new EgressNetworkPolicy object is created in a
+project named `project1`:
++
+----
+$ oc create -f default-rules.yaml -n project1
+egressnetworkpolicy.network.openshift.io/default-rules created
+----
++
+. Optionally, save the `<policy-name>.yaml` so that you can make changes later.
```
kalexand-rh: s/Optionally,/Optional:

> If a step is optional, begin the step with the word Optional. If your authoring tool provides a
semantic element to generate the word Optional, use that element.
Example
```
1. Optional: Click Set password. In the Password field, type a password.
```

modules/nw-egressnetworkpolicy-view.adoc
```diff
@@ -0,0 +1,33 @@
...
+.Procedure
+
+* To view the names of the EgressNetworkPolicy objects defined in your cluster,
+enter the following command:
++
+----
+$ oc get egressnetworkpolicy -o go-template='{{range .items}}{{.metadata.name}}{{end}}'
+----
+
+* Optionally, you can inspect a specific policy object by entering the following
```
kalexand-rh: Optional:

networking/openshift-sdn/configuring-egress-firewall.adoc
```diff
@@ -0,0 +1,13 @@
+[id="configuring-egress-network-policy"]
```
kalexand-rh: Please make your anchor match your assembly name.

networking/openshift-sdn/configuring-egress-firewall.adoc
```diff
@@ -0,0 +1,13 @@
...
+= Configuring egress network policy to control access to external IP addresses
+include::modules/common-attributes.adoc[]
+:context: configuring-egress-network-policy
+
+toc::[]
+
```
kalexand-rh: Please add an intro sentence or two.

networking/openshift-sdn/removing-egress-firewall.adoc
```diff
@@ -0,0 +1,11 @@
+[id="removing-egress-network-policy"]
+= Removing egress network policy from a project
+include::modules/common-attributes.adoc[]
+:context: removing-egress-network-policy
+
+toc::[]
+
+As a cluster administrator, you can remove an egress firewall from a project. This removes all restrictions on network traffic from the project leaving the {product-title} cluster.
```
kalexand-rh: I'd s/. This/ to

s/leaving/that leaves

modules/nw-egressnetworkpolicy-delete.adoc
```diff
@@ -0,0 +1,32 @@
+// Module included in the following assemblies:
+//
+// * networking/openshift-sdn/removing-egress-firewall.adoc
+
+[id="nw-egressnetworkpolicy-edit_{context}"]
```
kalexand-rh: Update the ID to match the module title.

modules/nw-egressnetworkpolicy-object.adoc
```diff
@@ -0,0 +1,70 @@
+// Module included in the following assemblies:
+//
+// * networking/openshift-sdn/configuring-egress-firewall.adoc
+
+[id="egressnetworkpolicy-cr-object_{context}"]
```
kalexand-rh: Update the anchor to match the module title

kalexand-rh (commented):
I have some suggestions and requests, but this is looking good.

jboxman:
@kalexand-rh, that was a lot to read through, thanks!

modules/nw-egressnetworkpolicy-about.adoc
danwinship: No, ENPs *do* apply to service endpoints. They don't apply to *Routes*.

Egress network policy rules do not apply to traffic that goes through Routers. A user who can create `Route`s can bypass Egress network policy rules by creating a route that points to a forbidden destination.

jboxman: @danwinship, great, thanks!

PR #16193, opened 2019-08-09
State that Grafana and its dashboards are read-only

rh-max:


rh-max:
Hi @bmcelvee could you please have a look at this one?

bmcelvee:
Hey @rh-max - the change LGTM, but the PR needs a rebase. To which versions of 3 does this change apply?

Also before we merge, I don't see the install_config -> monitoring folder and files in the topic map. We'll need to add those to make sure they build. 

Thanks!

rh-max:
Hey @bmcelvee , thanks for the review! Rebased. Regarding the `_topic_map.yml`, it's the 3.x docs, so all monitoring docs are under this entry:
`- Name: Prometheus Cluster Monitoring`
`  File: prometheus_cluster_monitoring`
`  Distros: openshift-origin,openshift-enterprise`

PR #16197, opened 2019-08-09
Bug 1714844 Clarify storage type and replica number

bmcelvee:
https://bugzilla.redhat.com/show_bug.cgi?id=1714844

### DIFF
--- a/modules/registry-configuring-storage-baremetal.adoc
+++ b/modules/registry-configuring-storage-baremetal.adoc
@@ -14,7 +14,7 @@ registry to use storage.
 * Cluster administrator permissions.
 * A cluster on bare metal.
 * A provisioned persistent volume (PV) with `ReadWriteMany` access mode, such as
-NFS.
+`NFS`.
 * Must have "100Gi" capacity.
 
 .Procedure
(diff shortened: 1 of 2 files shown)


bmcelvee:
@openshift/team-documentation please peer review

I might need to take the content out of the [Note] format, and am open to suggestions. Thanks!

modules/registry-configuring-storage-baremetal.adoc
```diff
@@ -57,4 +57,16 @@ $ oc get clusteroperator image-registry
 //type: Degraded
 //----
 
-​​
+[Note]
```
bergerhoffer: If you do keep this in a note, this has to be "NOTE" all caps, otherwise it doesn't render quite right when built. (Same with the below module).

modules/registry-configuring-storage-baremetal.adoc
```diff
@@ -57,4 +57,16 @@ $ oc get clusteroperator image-registry
...
+=====
+If the storage type is `emptyDIR`, the replica number can not be greater than `1`.
+If the storage type is NFS, and you want to scale up the registry Pod by setting
```
bergerhoffer: Should "NFS" be shown as a literal value like how `emptyDIR` is? Not sure if it's `nfs` or whatnot.

bergerhoffer (commented):
I don't have a strong opinion on whether this should be in a NOTE or not. But I do wonder whether it would make sense to instead be in with a step? I'm not sure exactly where it makes sense because I don't see anything about replica numbers in here, but is it talking about the storage type listed in step 2? Or is this elsewhere?

bmcelvee:
Thanks so much! I'm going to leave the note for now (corrected the formatting, silly mistake). I think I'm going to be reworking these modules soon anyway. :) 

Going ahead with the merge and cherrypick.

PR #16216, opened 2019-08-13
modules/dr-recover-lost-control-plane-hosts: update setup-etcd-environment

hexfusion:
This PR updates the location of `setup-etcd-environment` image. This used to be its own image but is now part of the `machine-config-operator` [1]. I also prefixed all oc commands with`sudo` because core user does not have access to `/var/lib/kubelet/config.json`

```
$ ls -la /var/lib/kubelet/config.json
-rw-------. 1 root root 120 Aug 12 23:50 /var/lib/kubelet/config.json
```

[1] https://github.com/openshift/machine-config-operator/pull/1028

### DIFF
--- a/modules/dr-recover-lost-control-plane-hosts.adoc
+++ b/modules/dr-recover-lost-control-plane-hosts.adoc
@@ -172,7 +172,7 @@ Login successful.
 ... Obtain the pull specification for the `kube-etcd-signer-server` image.
 +
 ----
-[core@ip-10-0-143-125 ~]$ export KUBE_ETCD_SIGNER_SERVER=$(oc adm release info --image-for kube-etcd-signer-server --registry-config=/var/lib/kubelet/config.json)
+[core@ip-10-0-143-125 ~]$ export KUBE_ETCD_SIGNER_SERVER=$(sudo oc adm release info --image-for kube-etcd-signer-server --registry-config=/var/lib/kubelet/config.json)
 ----
 
 ... Run the `tokenize-signer.sh` script.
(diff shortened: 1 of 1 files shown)


alaypatel07 (commented):
Is sudo required on other `oc get` calls? For instance, `2.i.a) oc get csr`

retroflexer:
> I also prefixed all oc commands with`sudo` because core user does not have access to `/var/lib/kubelet/config.json`
> 
> ```
> $ ls -la /var/lib/kubelet/config.json
> -rw-------. 1 root root 120 Aug 12 23:50 /var/lib/kubelet/config.json
> ```

Who creates this file? Why doesn't it have at least read permissions for group and others? If there is a way to avoid using `sudo` we should do it so.

hexfusion:
> Is sudo required on other `oc get` calls? For instance, `2.i.a) oc get csr`

nope will update.

hexfusion:
> > I also prefixed all oc commands with`sudo` because core user does not have access to `/var/lib/kubelet/config.json`
> > ```
> > $ ls -la /var/lib/kubelet/config.json
> > -rw-------. 1 root root 120 Aug 12 23:50 /var/lib/kubelet/config.json
> > ```
> 
> Who creates this file? Why doesn't it have at least read permissions for group and others? If there is a way to avoid using `sudo` we should do it so.

In general, I would agree but this is locked down for security purposes.
https://github.com/openshift/machine-config-operator/pull/827

bergerhoffer (approved):

PR #16221, opened 2019-08-13
Bug 1740684: Disable firewalld service during adding RHEL worker

bysnupy:
* Version: v4.1
* Reference:
  - [Unless RHEL worker disable firewalld.service, "oc rsh/exec/logs" does not work](https://bugzilla.redhat.com/show_bug.cgi?id=1740684)
  - https://github.com/openshift/openshift-ansible/pull/11824#issuecomment-520830918

* Description:
  Add a configuration step to disable and stop `firewalld` service to suppress 10250  port communication trouble.


modules/rhel-preparing-node.adoc
```diff
@@ -70,3 +70,10 @@ Note that this might take a few minutes if you have a large number of available
     --enable="rhel-7-server-extras-rpms" \
     --enable="rhel-7-server-ose-4.1-rpms"
 ----
+
+. If firewalld has been started on the host, disable and stop the service:
++
+----
+# systemctl disable firewalld.service
+# systemctl stop firewalld.service
+----
```
kalexand-rh: Is firewalld enabled by default? If so, I might say something like
```
. Stop and disable firewalld on the host:
+
----
# systemctl disable firewalld.service
# systemctl stop firewalld.service
----
+
[NOTE]
====
You must not enable firewalld later. If you do, you cannot access {product-title} logs on the worker.
====
```

@mtnbikenc, what do you think?

mtnbikenc: We should also mention the scaleup and upgrade playbooks will ensure firewalld is disabled.

kalexand-rh: Are you backporting this change to 4.1?

bysnupy: @kalexand-rh I've updated both scaleup and upgrade playbooks sections it's mentioned.

modules/rhel-compute-updating.adoc
```diff
@@ -18,6 +18,17 @@ Because the RHEL machines require assets that are generated by the cluster to
 complete the update process, you must update the cluster before you update
 the RHEL compute machines in it.
 ====
+* Stop and disable firewalld on the host:
```
kalexand-rh: @bysnupy, please move this step to after the Procedure heading and make it start with a "." instead of an "*".

kalexand-rh:
@wmengRH, will you PTAL?

wmengRH:
LGTM.

modules/rhel-compute-updating.adoc
bysnupy: @kalexand-rh , thank you for your review. I've modified as you mentioned.

kalexand-rh:
Thank you both! Merging.

PR #16280, opened 2019-08-16
Add notes to example inventory files about NFS DNS-1123 requirements

csheremeta:
https://bugzilla.redhat.com/show_bug.cgi?id=1703095

### DIFF
--- a/install/configuring_inventory_file.adoc
+++ b/install/configuring_inventory_file.adoc
@@ -1208,6 +1208,9 @@ be *_/exports/registry_*:
 ----
 [OSEv3:vars]
 
+# nfs_directory must conform to DNS-1123 subdomain must consist of lower case
+# alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character
+
 openshift_hosted_registry_storage_kind=nfs
 openshift_hosted_registry_storage_access_modes=['ReadWriteMany']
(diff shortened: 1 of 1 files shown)


vikram-redhat:
I am not sure how I agree with having the same warning again and again on the same page.

@openshift/team-documentation thoughts?

vikram-redhat:
@openshift/team-documentation *bump*

kalexand-rh:
I think that it's fine to have the note on the same page because they're in different contexts, and the user might not see it if we put it in only the same scenario.

install/configuring_inventory_file.adoc
```diff
@@ -1208,6 +1208,9 @@ be *_/exports/registry_*:
 ----
 [OSEv3:vars]
 
+# nfs_directory must conform to DNS-1123 subdomain must consist of lower case
```
kalexand-rh: Maybe:
> The `openshift_hosted_registry_storage_nfs_directory` must match the DNS-1123 subdomain name. Its name must consist of lower case lower case alphanumeric characters, dashes ('-)', or periods ('.') and must start and end with an alphanumeric character

vikram-redhat:
I am merging this. If any changes need to be made for the language, they can come at a later time.

Issue #1424, opened 2016-01-14
group sync needs to indicate that userNameAttributes needs to match LDAP configuration


thoraxe:
In the base LDAP configuration for OpenShift you have something like:

``` YAML
...
      attributes:
        id:
        - dn
        email:
        - mail
        name:
        - cn
        preferredUsername:
        - uid
```

And then in the group sync command, there is an element called userNameAttributes.

The example LDAP configuration in the documentation uses "uid" as the preferredUsername, but then the sync examples (https://docs.openshift.org/latest/install_config/syncing_groups_with_ldap.html) use "mail" as the userNameAttributes.

If you sync groups with "mail" but your userids are "uid", you end up with a mismatch, and OpenShift won't properly associate the users with the group. For example, if your group contains a username of `foo@example.com` but the real username in the system is just `foo`, even if `foo`'s group is assigned rights to a project, foo won't see them.

nnachefski:
+1.  i ran into this when doing group sync to an IPA server.  I changed the user attribute to uid and everything started working.  

thanks!
-Nick

mburke5678:
@pweil- We are culling through old issues and I am wondering about this one. Do we want to change the examples in the LDAP Example 5, 7, 10, and 13 use `userNameAttributes: [ mail ]`? 
If so, can you suggest someone who might have proper examples we can use?

pweil-:
@enj can you suggest some examples here?

eitchugo:
Sorry about necro'ing this issue, but I got into this problem recently: IMHO the 'mail' isn't a common attribute used here.

@mburke5678 I think the best approach is to replace lines in all examples from:

```
userNameAttributes: [ mail ]
```
to 
```
userNameAttributes: [ uid ]
```

Since we have this recommendation on the same page (using `enterprise-3.11` as reference) that needs to change as well:

```
Defines which attributes on an LDAP user entry will be used, in order, as its OpenShift Container Platform user name. The first attribute with a non-empty value is used. This should match your PreferredUsername setting for your LDAPPasswordIdentityProvider. The attribute to use as the name of the user in the OpenShift Container Platform group record. mail or sAMAccountName are preferred choices in most installations.
```

And from: https://docs.openshift.com/container-platform/3.11/install_config/configuring_authentication.html#LDAPPasswordIdentityProvider

The example is using `uid` as the `PreferredUsername` setting, which is actually a common setup (instead of mail).

If you agree with that, I can create a Pull Request. Let me know :)

[...]

PR #16287, opened 2019-08-16
group sync needs to indicate that userNameAttributes needs to match LDAP configuration

mburke5678:
The example LDAP configuration in the documentation uses "uid" as the preferredUsername, but then the sync examples (https://docs.openshift.org/latest/install_config/syncing_groups_with_ldap.html) use "mail" as the userNameAttributes.

...there was a mismatch between the two examples....the two examples need to match
https://github.com/openshift/openshift-docs/issues/1424


mburke5678:
@thoraxe PTAL

mburke5678:
@stuartchuan Is this an issue you can review?

PR #16302, opened 2019-08-19
Corrected example for in_forward for external logger

mburke5678:
@richm Tiny change. I think I was missing the Host address in the configuration sample for sending Fluentd logs to an external aggregator using `in_forward` for Fluentd 0.12


richm (approved):
/lgtm

PR #16309, opened 2019-08-19
removing reference to a singlet in the install-config

kalexand-rh:
@jboxman, will you PTAL?

### DIFF
--- a/modules/installation-aws-config-yaml.adoc
+++ b/modules/installation-aws-config-yaml.adoc
@@ -81,13 +81,13 @@ sshKey: ssh-ed25519 AAAA... <6>
 <1> Required. The installation program prompts you for this value.
 <2> If you do not provide these parameters and values, the installation program
 provides the default value.
-<3> The `controlPlane` section is a singlet, and the `compute` section is an
-array. Because the sections have different definitions, the first line of the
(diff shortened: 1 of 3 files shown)


jboxman:
@kalexand-rh 

It doesn't sound like from the annotation text that we have an expectation that a user is familiar with YAML at all, although that seems to be the preferred format for CRs in 4.x. YAML is basically a human readable format for complex data structures. So one parameter requires/accepts an object, the other an array of objects. That's how I've always thought about it and what's happening under the hood as it were.

I'm thinking something along these lines:

> The `controlPlane` parameter expects an object that describes the configuration for the control plane pool. The `compute` parameter expects an array of objects with each object describing the configuration for the compute pools. It is possible that future versions of OpenShift Container Platform will support defining multiple compute pools during installation. Only one control plane pool is used.

The machine pool configuration object is probably the same golang type definition, but I'd need to check the source to be sure.

(Looks like the YAML spec defines a YAML node as either scalar, mapping, or collection, so there's that terminology as well.)

kalexand-rh:
@jboxman, I don't think that's clear enough to explain why you must include the hyphen in one of the definitions but not the other. (I got a bug earlier in the release asking for clarification on that point.)

(Yes, that's why Trevor and I picked "mapping" for the change.)

PR #16312, opened 2019-08-19
Pod/Container can steal cpu/memory resources if cluster admin did not set minimum (not maximum) limitrange

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1644512


mburke5678:
See https://github.com/openshift/openshift-docs/pull/15425 for discussion and approval

PR #16319, opened 2019-08-19
SRVCOM-488: OpenShift Serverless Installation 

nebelsprachen:
- branch/enterprise-4.1 
- branch/enterprise-4.2

### DIFF
--- a/modules/serverless-cluster-sizing-requirements.adoc
+++ b/modules/serverless-cluster-sizing-requirements.adoc
@@ -1,0 +1,24 @@
+// Module included in the following assemblies:
+//
+// serverless/installing-openshift-serverless.adoc
+
+[id="cluster-size-requirements_context"]
+= Cluster size requirements
+
+The cluster must be sized appropriately to ensure that OpenShift Serverless can run correctly. You can use the MachineSet API to manually scale your cluster up to the desired size.
+
+An OpenShift cluster with 10 CPUs and 40 GB memory is the minimum requirement for getting started with your first serverless application. This usually means you must scale up one of the default MachineSets by two additional machines. 
+
+[NOTE]
+====
+For this configuration, the requirements depend on the deployed applications. By default, each pod requests ~400m of CPU and recommendations are based on this value. In the given recommendation, an application can scale up to 10 replicas. Lowering the actual CPU request of the application further pushes the boundary.
+====
+
+[NOTE]
+====
+The numbers given only relate to the pool of worker machines of the OpenShift cluster. Master nodes are not used for general scheduling and are omitted.
+====
+
+For more advanced use-cases, such as using OpenShift logging, monitoring, metering, and tracing, you must deploy more resources. Recommended requirements for such use-cases are 24 vCPUs and 96GB of memory.
+
+
--- a/modules/serverless-installing-knative-serving.adoc
+++ b/modules/serverless-installing-knative-serving.adoc
@@ -1,0 +1,50 @@
+// Module included in the following assemblies:
+//
+// serverless/installing-openshift-serverless.adoc
+
+[id="installing-knative-serving_{context}"]
+= Installing Knative Serving
+
+You must create a `KnativeServing` object to install Knative Serving using the OpenShift Serverless Operator.
+
+[source,yaml]
+.Sample `serving.yaml`
+----
+apiVersion: v1
+kind: Namespace
+metadata:
+ name: knative-serving
+---
+apiVersion: serving.knative.dev/v1alpha1
+kind: KnativeServing
+metadata:
+ name: knative-serving
+ namespace: knative-serving
+----
+
+
+.Prerequisite
+* An account with cluster administrator access.
+* Installed OpenShift Serverless Operator.
+
+
+.Procedure
+. Copy the sample YAML file into `serving.yaml` and apply it using:
++
+----
+$ oc apply -f serving.yaml
+----
++
+. Verify the installation is complete by using the command:
++
+----
+$ oc get knativeserving/knative-serving -n knative-serving --template='{{range .status.conditions}}{{printf "%s=%s\n" .type .status}}{{end}}'
+----
++
+Results should be similar to:
++
+----
+DeploymentsAvailable=True
+InstallSucceeded=True
+Ready=True
+----
--- a/modules/serverless-installing-service-mesh-control-plane.adoc
+++ b/modules/serverless-installing-service-mesh-control-plane.adoc
@@ -1,0 +1,92 @@
+// Module included in the following assemblies:
+//
+// serverless/installing-openshift-serverless.adoc
+
+[id="installing-service-mesh-control-plane_{context}"]
+= Installing the ServiceMeshControlPlane
+
+Service Mesh is comprised of a data plane and a control plane. After you install the ServiceMesh operator, you can install the control plane. The control plane manages and configures the sidecar proxies to enforce policies and collect telemetry. The following procedure installs a version of the ServiceMesh control plane that acts as an ingress to your applications.
+
+[NOTE] 
+====
+You must install the control plane into the `istio-system` namespace. Other namespaces are currently not supported.
+====
+
+[source,yaml]
+.Sample servicemesh-installation.yaml
+----
+apiVersion: maistra.io/v1
+kind: ServiceMeshControlPlane
+metadata:
+ name: basic-install
+ namespace: istio-system <1>
+spec:
+ istio:
+   global:
+     multitenant: true
+     proxy:
+       autoInject: disabled
+     omitSidecarInjectorConfigMap: true
+     disablePolicyChecks: false
+   istio_cni:
+     enabled: true
+   gateways:
+     istio-ingressgateway:
+       autoscaleEnabled: false
+       type: LoadBalancer
+     istio-egressgateway:
+       enabled: false
+     cluster-local-gateway:
+       autoscaleEnabled: false <1>
+       enabled: true
+       labels:
+         app: cluster-local-gateway
+         istio: cluster-local-gateway
+       ports:
+         - name: status-port
+           port: 15020
+         - name: http2
+           port: 80
+           targetPort: 8080
+         - name: https
+           port: 443
+   mixer:
+     enabled: false
+     policy:
+       enabled: false
+     telemetry:
+       enabled: falsely 
+   pilot:
+     autoscaleEnabled: false
+     sidecar: false
+   kiali:
+     enabled: false 
+   tracing:
+     enabled: false
+   prometheus:
+     enabled: false
+   grafana:
+     enabled: false
+   sidecarInjectorWebhook:
+     enabled: false
+----
+<1> Autoscaling is disabled in this version. This release is not intended for production use.
+
+[NOTE]
+====
+Running ServiceMesh with a sidecar injection enabled with OpenShift Serverless is currently not recommended. 
+====
+
+.Prerequisite
+* An account with cluster administrator access.
+* The ServiceMesh operator must be installed. 
+
+.Procedure
+. Copy the sample YAML file into a `smcp.yaml` file.
+. Apply the YAML file using the command:
++
+----
+$ oc apply -f smcp.yaml
+----
+
+
--- a/modules/serverless-installing-service-mesh-member-roll.adoc
+++ b/modules/serverless-installing-service-mesh-member-roll.adoc
@@ -1,0 +1,45 @@
+// Module included in the following assemblies:
+//
+// serverless/installing-openshift-serverless.adoc
+
+[id="installing-service-mesh-member-roll_{context}"]
+= Installing a ServiceMeshMemberRoll
+
+You must have a Service Mesh Member Roll for the control plane namespace, if the Service Mesh is configured for multi-tenancy. For applications to use the deployed control plane and ingress, their namespaces must be part of a member roll. 
+
+A multi-tenant control plane installation only affects namespaces configured as part of the Service Mesh. You must specify the namespaces associated with the Service Mesh in a `ServiceMeshMemberRoll` resource located in the same namespace as the `ServiceMeshControlPlane` resource and name it `default`.
+
+.ServiceMeshMemberRoll Custom Resource Example
+----
+apiVersion: maistra.io/v1
+kind: ServiceMeshMemberRoll
+metadata:
+ name: default
+ Namespace: istio-system
+spec:
+ members:<1>
+ - knative-serving
+ - mynamespace
+----
+<1> A list of namespaces that should be joined into the Service Mesh
+
+
+.Prerequisites
+* Installed Service Mesh Operator.
+* A custom resource file that defines the parameters of your Red Hat OpenShift Service Mesh control plane.
+
+
+.Procedure
+. Create a YAML file that replicates the ServiceMeshMemberRoll Custom Resource sample.
+. Configure the YAML file to include relevant namespaces.
++
+[NOTE] 
+====
+Add all namespaces to which you want to deploy serverless applications. Ensure you retain the `knative-serving` namespace in the member roll.
+====
++ 
+. Copy the configured YAML into a file `smmr.yaml` and apply it using:
++
+----
+$ oc apply -f smmr.yaml
+----
--- a/serverless/installing-openshift-serverless.adoc
+++ b/serverless/installing-openshift-serverless.adoc
@@ -1,0 +1,59 @@
+[id="installing-openshift-serverless"]
+= Installing OpenShift Serverless
+include::modules/common-attributes.adoc[]
+:context: installing-openshift-serverless
+
+toc::[]
+
+// Cluster sizing requirements
+include::modules/serverless-cluster-sizing-requirements.adoc[leveloffset=+1]
+// Add or remove an instance of a machine in a MachineSet
+include::modules/machineset-manually-scaling.adoc[leveloffset=+2]
+
+
+// Installing Service Mesh
+include::modules/serverless-installing-service-mesh-no-kiali-jaeger.adoc[leveloffset=+1]
+
+
+// Installing Service Mesh Control Plane
+include::modules/serverless-installing-service-mesh-control-plane.adoc[leveloffset=+2]
+// Verifying the control plane installation
+include::modules/ossm-control-plane-verify.adoc[leveloffset=+2]
+
+
+// Installing Service Mesh Member Roll
+include::modules/serverless-installing-service-mesh-member-roll.adoc[leveloffset=+2]
+
+
+// Installing Serverless Operator
+[id="installing-serverless-operator_{context}"]
+== Installing the OpenShift Serverless Operator
+
+The OpenShift Serverless Operator can be installed using the {product-title} instructions for installing Operators.
+
+You can install the OpenShift Serverless Operator in the host cluster by following the {product-title} instructions on installing an Operator.
+
+[NOTE]
+====
+The Serverless Operator only works for OpenShift versions 4.1.13 and later.
+====
+
+For details, see the {product-title} documentation on xref:../applications/operators/olm-adding-operators-to-cluster.adoc[adding Operators to a cluster].
+
+
+[id="deleting-openshift-serverless-operator_{context}"]
+== Deleting the OpenShift Serverless Operator
+
+You can remove the OpenShift Serverless Operator from the host cluster by following the {product-title} instructions on deleting an Operator.
+
+For details, see the {product-title} documentation on xref:../applications/operators/olm-deleting-operators-from-cluster.adoc[deleting Operators from a cluster].
+
+
+// Installing Knative Serving
+include::modules/serverless-installing-knative-serving.adoc[leveloffset=+1]
+
+
+
+
+
+
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -1010,6 +1010,9 @@ Topics:
 #  File: understanding-serverless
 #- Name: OpenShift Serverless product architecture
 #  File: serverless-architecture
+### Installing OpenShift Serverless
+#- Name: Installing OpenShift Serverless
+#  File: installing-openshift-serverless
 ### Knative services
 #- Name: Getting started with Knative services
 #  File: getting-started-knative-services
--- a/modules/machineset-manually-scaling.adoc
+++ b/modules/machineset-manually-scaling.adoc
@@ -5,14 +5,14 @@
 [id="machineset-manually-scaling_{context}"]
 = Scaling a MachineSet manually
 
-If you must add or remove an instance of a machine in a MachineSet, you can
-manually scale the MachineSet.
+If you must add or remove an instance of a machine in a MachineSet, you can manually scale the MachineSet.
 
 .Prerequisites
 
 * Install an {product-title} cluster and the `oc` command line.
 * Log into `oc` as a user with `cluster-admin` permission.
 
+.Procedure
 . View the MachineSets that are in the cluster:
 +
 ----
--- a/modules/ossm-control-plane-verify.adoc
+++ b/modules/ossm-control-plane-verify.adoc
@@ -8,12 +8,12 @@
 
 Verify that the control plane installation completed successfully.
 
-.Procedure
-
 [NOTE]
 ====
 The name of the resource is `istio-installation`.
 ====
+
+.Procedure
 
 . Run this command to determine if the Operator finished deploying the control plane:
 +
--- a/modules/ossm-operator-install.adoc
+++ b/modules/ossm-operator-install.adoc
@@ -5,14 +5,14 @@
 [id="ossm-operator-installation_{context}"]
 = Installing the Operators
 
-The {ProductShortName} installation process introduces an Operator to manage the installation of the control plane within the `istio-operator` namespace. This Operator defines and monitors a custom resource related to the deployment, update, and deletion of the control plane.
+The Service Mesh installation process introduces an Operator to manage the installation of the control plane within the `istio-operator` namespace. This Operator defines and monitors a custom resource related to the deployment, update, and deletion of the control plane.
 
-Starting with {ProductName} {ProductVersion}, you must install the Jaeger Operator and the Kiali Operator before the {ProductName} Operator can install the control plane.
+You must install the Jaeger Operator and the Kiali Operator before the Service Mesh Operator can install the control plane.
 
 
 [id="ossm-operator-install-jaeger_{context}"]
 == Installing the Jaeger Operator
-You must install the Jaeger Operator for the {ProductName} Operator to install the control plane.
+You must install the Jaeger Operator for the Service Mesh Operator to install the control plane.
 
 .Prerequisites
 * An account with cluster administrator access.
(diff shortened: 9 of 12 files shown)


modules/serverless-cluster-sizing-requirements.adoc
```diff
@@ -0,0 +1,21 @@
...
+= Cluster size requirements
+
+Determine how many nodes and pods you require for your OpenShift Container Platform cluster. Cluster scalability correlates to the number of pods in a cluster environment. That number influences the other numbers in your setup.
+
+// Scalability, planning requirements, and limits
+include:scalability_and_performance/planning-your-environment-according-to-object-limits.adoc[leveloffset=+1]
+
+Your cluster must be sized appropriately to ensure the instructions work properly. To do so, you can use the MachineSet API to manually scale your cluster up to the desired size.
+
+An Openshift cluster with 16 CPUs and 40 GB memory is the minimum requirement for getting started with your first serverless application. This usually means you will have to scale up one of the default MachineSets by two additional machines.
```
markusthoemmes:
```suggestion
An Openshift cluster with 10 CPUs and 40 GB memory is the minimum requirement for getting started with your first serverless application. This usually means you will have to scale up one of the default MachineSets by two additional machines.
```

We corrected that in the doc already.

modules/serverless-cluster-sizing-requirements.adoc
```diff
@@ -0,0 +1,21 @@
...
+[id="cluster-size-requirements_context"]
+= Cluster size requirements
+
+Determine how many nodes and pods you require for your OpenShift Container Platform cluster. Cluster scalability correlates to the number of pods in a cluster environment. That number influences the other numbers in your setup.
+
+// Scalability, planning requirements, and limits
+include:scalability_and_performance/planning-your-environment-according-to-object-limits.adoc[leveloffset=+1]
+
+Your cluster must be sized appropriately to ensure the instructions work properly. To do so, you can use the MachineSet API to manually scale your cluster up to the desired size.
+
```
markusthoemmes: Add a note here:

> Note: All numbers given only relate to the pool of worker machines of the Openshift cluster. Master nodes are not used for general scheduling and thus are omitted.

Adjusted in the doc already.

modules/serverless-installing-knative-serving.adoc
```diff
@@ -0,0 +1,48 @@
...
+
+[id="installing-knative-serving_{context}"]
+= Installing Knative Serving
+
+You need to create a `KnativeServing` object to install Knative Serving using the Openshift Serverless Operator.
+
+
+[IMPORTANT]
+====
+This sample YAML file is provided for reference.
```
markusthoemmes: What does this tell me as a user?

modules/serverless-installing-service-mesh-control-plane.adoc
```diff
@@ -0,0 +1,103 @@
...
+     enabled: false
+   prometheus:
+     enabled: false
+   grafana:
+     enabled: false
+   sidecarInjectorWebhook:
+     enabled: false
+----
+<1> Required. The installation program prompts you for this value.
+<2> Autoscaling is disabled in this version, as this is currently not intended for production use.
```
markusthoemmes: Maybe slightly reword this to sound more like: This is a minimum version to kick the tires. Please adjust these values to your production needs as necessary?

modules/serverless-installing-service-mesh-control-plane.adoc
```diff
@@ -0,0 +1,103 @@
...
+   tracing:
+     enabled: false
+   prometheus:
+     enabled: false
+   grafana:
+     enabled: false
+   sidecarInjectorWebhook:
+     enabled: false
+----
+<1> Required. The installation program prompts you for this value.
```
markusthoemmes: Which installation program? This seems like a duplicate to the note below?

markusthoemmes (commented):
Left a few comments, I think a preview is needed in this case to see what the includes result in.

modules/serverless-cluster-sizing-requirements.adoc
```diff
@@ -0,0 +1,21 @@
+// Module included in the following assemblies:
+//
+// serverless/installing-openshift-serverless.adoc
+
+
+[id="cluster-size-requirements_context"]
+= Cluster size requirements
+
+Determine how many nodes and pods you require for your OpenShift Container Platform cluster. Cluster scalability correlates to the number of pods in a cluster environment. That number influences the other numbers in your setup.
```
abrennan89: "You must determine"

abrennan89: Ensure that {product-title} variable is being used instead of manually writing "OpenShift Container Platform" each time

abrennan89: "Cluster scalability correlates to the number of pods in a cluster environment. That number influences the other numbers in your setup." I don't think it's clear enough what this means. Please look at rewriting this sentence so it explains better.

modules/serverless-cluster-sizing-requirements.adoc
```diff
@@ -0,0 +1,21 @@
...
+
+// Scalability, planning requirements, and limits
+include:scalability_and_performance/planning-your-environment-according-to-object-limits.adoc[leveloffset=+1]
```
abrennan89: It looks like this include is inside a module - shouldn't  modules/serverless-cluster-sizing-requirements.adoc  then be an assembly instead?

modules/serverless-cluster-sizing-requirements.adoc
```diff
@@ -0,0 +1,21 @@
...
+
+Your cluster must be sized appropriately to ensure the instructions work properly. To do so, you can use the MachineSet API to manually scale your cluster up to the desired size.
```
abrennan89: We should add some details about what MachineSet API is or a link to existing docs etc; to me it's confusing to just mention this but not have a procedure or some explanation?

modules/serverless-cluster-sizing-requirements.adoc
```diff
@@ -0,0 +1,21 @@
...
+
+An Openshift cluster with 16 CPUs and 40 GB memory is the minimum requirement for getting started with your first serverless application. This usually means you will have to scale up one of the default MachineSets by two additional machines.
+
+For more advanced use-cases, such as using Openshift Logging, Monitoring, Metering and Tracing, you will need to deploy more resources. Recommended requirements for such use-cases are 24 vCPUs and 96GB of memory.
+
+// Add or remove an instance of a machine in a MachineSet
```
abrennan89: There isn't enough content here about what machinesets are IMHO, maybe add a link to some context info or add something about how these work regarding serverless?

modules/serverless-installing-knative-serving.adoc
abrennan89: Should probably be "This sample yaml shows the minimum configuration to create a KnativeServing object:" or something like that

modules/serverless-installing-knative-serving.adoc
```diff
@@ -0,0 +1,48 @@
...
+
+
+.Procedure
+. Copy the sample YAML file into `serving.yaml` and apply it using:
++
+----
+oc apply -f serving.yaml
+----
++
+. Verify the installation is complete by using the command:
```
abrennan89: What should the output be to verify against?

modules/serverless-installing-service-mesh-control-plane.adoc
```diff
@@ -0,0 +1,103 @@
+// Module included in the following assemblies:
+//
+// serverless/installing-openshift-serverless.adoc
+
+= Installing the ServiceMeshControlPlane
+
+After you install the ServiceMesh operator, you can install the control plane. The following procedure installs a version of ServiceMesh that acts as an ingress to your applications.
+
+[IMPORTANT]
+====
```
abrennan89: again, explain exactly what the sample is for, there's no context

modules/serverless-installing-service-mesh-control-plane.adoc
abrennan89: This looks like it refers to the namespace. I'd remove the 'installation program' bit and just call out the fields that are required.
I'm not sure there's a need to include that whole huge code sample - does this not exist in service mesh docs or somewhere else already?

modules/serverless-installing-service-mesh-control-plane.adoc
```diff
@@ -0,0 +1,103 @@
...
+
+
+====
+[NOTE]
+Running ServiceMesh with a sidecar injection enabled with Openshift Serverless is currently not recommended. 
+====
+
+====
+[NOTE] 
+You need to install the control plane into the `istio-system` namespace. Other namespaces are currently not supported.
```
abrennan89: Move this note to earlier in the section

modules/serverless-installing-service-mesh-control-plane.adoc
```diff
@@ -0,0 +1,103 @@
...
+====
+
+
+.Prerequisite
+* An account with cluster administrator access.
+* The ServiceMesh operator must be installed. 
+
+
+.Procedure
+. Copy the sample YAML file into a `smcp.yaml` file.
```
abrennan89: what sample yaml file? please put it here instead so it makes sense

modules/serverless-installing-service-mesh-member-roll.adoc
```diff
@@ -0,0 +1,45 @@
+// Module included in the following assemblies:
+//
+// serverless/installing-openshift-serverless.adoc
+
+[id="installing-service-mesh-member-roll_{context}"]
+= Installing a ServiceMeshMemberRoll
+
+For applications to make use of the deployed control plane and ingress, their namespaces need to be part of a member roll.
```
abrennan89: "need to be added to a member roll"

modules/serverless-installing-service-mesh-member-roll.adoc
```diff
@@ -0,0 +1,45 @@
+// Module included in the following assemblies:
+//
+// serverless/installing-openshift-serverless.adoc
+
+[id="installing-service-mesh-member-roll_{context}"]
+= Installing a ServiceMeshMemberRoll
+
+For applications to make use of the deployed control plane and ingress, their namespaces need to be part of a member roll. 
+
+.ServiceMeshMemberRoll Custom Resource Example
```
abrennan89: again, no context, what is this example for, explain it

modules/serverless-installing-service-mesh-member-roll.adoc
```diff
@@ -0,0 +1,45 @@
...
+A multi-tenant control plane installation only affects namespaces configured as part of the Service Mesh. You must specify the namespaces associated with the Service Mesh in a `ServiceMeshMemberRoll` resource located in the same namespace as the `ServiceMeshControlPlane` resource and name it `default`.
+
+
+.Prerequisites
+* An installed, verified Service Mesh Operator.
+* A custom resource file that defines the parameters of your Red Hat OpenShift Service Mesh control plane.
+
+
+.Procedure
+. Create a YAML file that replicates the ServiceMeshMemberRoll Custom Resource sample.
```
abrennan89: show all the examples here with the actual procedure steps to give context, random examples with no steps don't have enough context

modules/serverless-installing-service-mesh.adoc
```diff
@@ -0,0 +1,45 @@
...
+====
+
+== Installing ServiceMesh without Kiali and Jaeger Operators
+
+.Prerequisites
+* An account with cluster administrator access.
+
+
+.Procedure
+. Log into your OpenShift Container Platform installation as a cluster administrator.
```
abrennan89: {product-title}

modules/serverless-installing-service-mesh.adoc
```diff
@@ -0,0 +1,45 @@
...
+. Run the following commands to install:
```
abrennan89: "Run the following commands to install Service Mesh:"

modules/serverless-installing-the-serverless-operator.adoc
```diff
@@ -0,0 +1,10 @@
+// Module included in the following assemblies:
+//
+// serverless/installing-openshift-serverless.adoc
+
+= Installing the OpenShift Serverless Operator
+
+The OpenShift Serverless Operator can be installed by using the operator indexed in the OpenShift OperatorHub.
```
abrennan89: "The OpenShift Serverless Operator can be installed by using the {product-title} instructions on installing an Operator."

Include CLI and web console modules?

abrennan89 (changes requested):

modules/serverless-installing-service-mesh-control-plane.adoc
markusthoemmes: Not in this form, it's actually the recommended and guaranteed-to-work well way to install ServiceMesh for the serverless setting.

markusthoemmes: We could however pick apart and call out the "important" bits (i.e. disabling the actual mesh injection)

modules/serverless-installing-service-mesh-member-roll.adoc
joaedwar: @abrennan89  I don't understand what you mean by this. Would you like me to include the YAML file in the procedure? Looking at OpenShift docs examples, it seems they have the YAML examples outside of the procedures

abrennan89: If you're including the example outside the procedure, I'd make sure it's explained with some context about what individual fields are required for basically.
E.g. in the service mesh docs here https://docs.openshift.com/container-platform/4.1/service_mesh/service_mesh_install/installing-ossm.html there are tables which explain what the fields do.
I'm not sure we need to explain every field in a yaml example because some of it is probably assumed knowledge or elsewhere in docs, like at the link above that you could point to, but I think it'd be better to try to call out or draw attention to what's actually different for serverless.

LMK if that makes sense. @markusthoemmes any thoughts? I think you mentioned something similar in another comment re calling out the important fields in sample yaml?

serverless/installing-service-mesh-control-plane.adoc
```diff
@@ -0,0 +1,97 @@
...
+       enabled: true
+       labels:
+         app: cluster-local-gateway
+         istio: cluster-local-gateway
+       ports:
+         - name: status-port
+           port: 15020
+         - name: http2
+           port: 80
+           targetPort: 80
```
markusthoemmes: This should now say `targetPort: 8080` (changed in the GDoc)

serverless/installing-serverless-operator.adoc
```diff
@@ -0,0 +1,10 @@
+// Module included in the following assemblies:
+//
+// serverless/installing-openshift-serverless.adoc
+
+= Installing the OpenShift Serverless Operator
+
+The OpenShift Serverless Operator can be installed by using the operator indexed in the OpenShift OperatorHub.
```
abrennan89: s/installed by using the operator indexed in the OpenShift OperatorHub. /can be installed using the {product-title} instructions for installing Operators.

Although I would recommend adding this to the assembly instead so that you can just xref to the whole operators assembly rather than have to inc. modules. The CLI commands for example for installing Operators are not in the module below.

You can see what I mean in the KubeFed docs where I've used this: https://github.com/openshift/openshift-docs/blob/master/kubefed/installing-kubefed.adoc

serverless/installing-service-mesh-control-plane.adoc
```diff
@@ -0,0 +1,97 @@
...
+* An account with cluster administrator access.
+* The ServiceMesh operator must be installed. 
+
+
+.Procedure
+. Copy the sample YAML file into a `smcp.yaml` file.
+. Apply the YAML file using the command:
++
+----
+ oc apply -f smcp.yaml
```
abrennan89: $ oc apply....

abrennan89 (changes requested):
Some additional feedback

modules/serverless-installing-service-mesh-member-roll.adoc
```diff
@@ -0,0 +1,45 @@
+// Module included in the following assemblies:
+//
+// serverless/installing-openshift-serverless.adoc
+
+[id="installing-service-mesh-member-roll_{context}"]
+= Installing a ServiceMeshMemberRoll
+
+If Service Mesh is configured for multi-tenancy, you should have a Service Mesh Member Roll for the control plane namespace. For applications to make use of the deployed control plane and ingress, their namespaces need to be part of a member roll.
```
alanfx: AFAIK, I'm not sure you can configure Service Mesh without multi-tenancy any more? I guess we'd have to verify this with the Istio team, but that is my understanding.

joaedwar: @alanfx Should the 'If" be removed and this made into an imperative ( i.e You must have a Service Mesh Member Roll...)

serverless/installing-openshift-serverless.adoc
```diff
@@ -0,0 +1,17 @@
...
+:context: installing-openshift-serverless
+
+toc::[]
+
+include:serverless/cluster-sizing-requirements.adoc[leveloffset=+1]
+include:serverless/installing-service-mesh.adoc[leveloffset=+1]
+include:serverless/installing-service-mesh-control-plane.adoc[leveloffset=+1]
+include:modules/serverless-installing-service-mesh-member-roll.adoc[leveloffset=+1]
+include:modules/serverless-installing-knative-serving.adoc[leveloffset+=1]
+include:serverless/installing-serverless-operator.adoc[leveloffset+=1]
```
adellape: All of these `include:`s should have 2 colons: `include::`.

adellape: Same for all other instances of `include:` in other files in this PR.

serverless/installing-service-mesh-control-plane.adoc
```diff
@@ -0,0 +1,97 @@
...
+//
+// serverless/installing-openshift-serverless.adoc
+
+= Installing the ServiceMeshControlPlane
+
+Service Mesh is comprised of a data plane and a control plane. After you install the ServiceMesh operator, you can install the control plane. The control plane manages and configures the sidecar proxies to enforce policies and collect telemetry. The following procedure installs a version of the ServiceMesh control plane that acts as an ingress to your applications.
+
+====
+[NOTE] 
+You need to install the control plane into the `istio-system` namespace. Other namespaces are currently not supported.
```
adellape: s/need to/must/

modules/serverless-installing-knative-serving.adoc
```diff
@@ -0,0 +1,43 @@
+// Module included in the following assemblies:
+//
+// serverless/installing-openshift-serverless.adoc
+
+[id="installing-knative-serving_{context}"]
+= Installing Knative Serving
+
+You need to create a `KnativeServing` object to install Knative Serving using the Openshift Serverless Operator.
```
adellape: There are a lot of instances throughout the PR of `Openshift` (lowercase `s`). They should all be `OpenShift`.

serverless/cluster-sizing-requirements.adoc
```diff
@@ -0,0 +1,16 @@
...
+= Cluster size requirements
+
+The cluster must be sized appropriately to ensure the instructions work properly. To do so, you can use the MachineSet API to manually scale your cluster up to the desired size.
+
+An Openshift cluster with 10 CPUs and 40 GB memory is the minimum requirement for getting started with your first serverless application. This usually means you will have to scale up one of the default MachineSets by two additional machines.
+
+For more advanced use-cases, such as using Openshift Logging, Monitoring, Metering and Tracing, you will need to deploy more resources. Recommended requirements for such use-cases are 24 vCPUs and 96GB of memory.
+
+// Add or remove an instance of a machine in a MachineSet
+include:modules/machineset-manually-scaling.adoc[leveloffset=+1]
```
adellape: This module is missing the `.Procedure` after the `.Prerequisites` list.

serverless/installing-service-mesh.adoc
```diff
@@ -0,0 +1,17 @@
+// Module included in the following assemblies:
+//
+// serverless/installing-openshift-serverless.adoc
+
+[id=installing-service-mesh-operator_{context}]
+= Installing the ServiceMesh operator
+
+Openshift Serverless requires an installed version of ServiceMesh.
+
+include:modules/ossm-operator-install.adoc[leveloffset=+1]
```
adellape: The `{ProductShortName}` in this module doesn't seem to be rendering properly.

image

adellape: Same for a number of `{ProductName}` instances.

serverless/installing-service-mesh.adoc
```diff
@@ -0,0 +1,17 @@
...
+
+====
+[NOTE] 
+ServiceMesh documentation requests the installation of Kiali and Jaeger Operators. Kiali and Jaeger operators do not need to be installed unless you plan to use them. They are disabled in the default settings proposed below.
+====
```
adellape: Should be in this form:

```
[NOTE]
====
ServiceMesh documentation requests the installation of Kiali and Jaeger Operators. Kiali and Jaeger operators do not need to be installed unless you plan to use them. They are disabled in the default settings proposed below.
====
```

Note that there also appears to be a whitespace after `[NOTE] ` above; should be removed.

adellape: There's a few other instances of this out-of-order `[NOTE`] and `====` blocks.

serverless/installing-service-mesh-control-plane.adoc
```diff
@@ -0,0 +1,97 @@
...
+. Copy the sample YAML file into a `smcp.yaml` file.
+. Apply the YAML file using the command:
++
+----
+ oc apply -f smcp.yaml
+----
+
+
+// Verifying the control plane installation
+include:modules/ossm-control-plane-verify.adoc[leveloffset=+1]
```
adellape: This module has this spot in it:

```
.Procedure

[NOTE]
====
The name of the resource is `istio-installation`.
====
```

This renders funny:

image

Suggest moving the note admonition either before the `.Procedure` or elsewhere in a step.

serverless/installing-serverless-operator.adoc
adellape: :+1:

adellape (changes requested):
After the `include::`s are fixed (see inline comments), we can get a preview doc build going for everyone to review.

serverless/cluster-sizing-requirements.adoc
```diff
@@ -0,0 +1,23 @@
...
+
+[id="cluster-size-requirements_context"]
+= Cluster size requirements
+include::modules/common-attributes.adoc[]
+:context: cluster-size-requirements
+
+The cluster must be sized appropriately to ensure that OpenShift Serverless can run correctly. You can use the MachineSet API to manually scale your cluster up to the desired size.
+
+An OpenShift cluster with 10 CPUs and 40 GB memory is the minimum requirement for getting started with your first serverless application. This usually means you will have to scale up one of the default MachineSets by two additional machines.
+
```
mgencur: I think this section is now missing some parts of the updated installation document:

Note: The requirements are greatly depending on the deployed applications. By default, each pod requests ~400m of CPU and the recommendation is based on that. With the given recommendation, such an application can scale up to 10 replicas. Lowering the actual CPU request of the application further pushes that boundary.

For more advanced full-fledged use-cases (for example making use of Monitoring, Metering and Tracing) you’ll need to deploy more resources according to the respective documentation. A good starting point are 24 vCPUs and 96GB of memory.

For OpenShift Logging, there are specific CPU and memory requirements. The Elasticsearch node needs 16G of memory for both memory requests and CPU limits. You might need to add additional nodes to the OpenShift Container Platform cluster to run with the recommended or higher memory. When running the OpenShift Logging on a cluster in AWS the recommended instance type is m4.2xlarge (8 vCPUs, 32 GiB Mem).

serverless/installing-service-mesh-control-plane.adoc
```diff
@@ -0,0 +1,96 @@
...
+   grafana:
+     enabled: false
+   sidecarInjectorWebhook:
+     enabled: false
+----
+<1> Autoscaling is disabled in this version. This release is not intended for production use.
+
+[NOTE]
+====
+Running ServiceMesh with a sidecar injection enabled with OpenShift Serverless is currently not recommended.
```
mgencur: Just a note: The service mesh team has a fix for the blocker issue that prevented us from running with sidecar injection. We might need to revisit what we want to use when the fix lands and testing shows that it works.

mgencur:
Looks good to me now.
QE Approved

modules/serverless-installing-knative-serving.adoc
```diff
@@ -0,0 +1,50 @@
...
+. Copy the sample YAML file into `serving.yaml` and apply it using:
++
+----
+$ oc apply -f serving.yaml
+----
++
+. Verify the installation is complete by using the command:
++
+----
+$ oc get knativeserving/knative-serving -n knative-serving --template='{{range .status.conditions}}{{printf "%s=%s\n" .type .status}}{{end}'
```
mgencur: This command is missing one curly bracket at the end. Should be ```oc get knativeserving/knative-serving -n knative-serving --template='{{range .status.conditions}}{{printf "%s=%s\n" .type .status}}{{end}}'```

joaedwar: Got it !! Thanks Martin

mgencur:
QE Approved

bbrowning (changes requested):
Lots of great stuff here - thanks! I've left a few comments with some screenshots of the rendered preview to see where some things are bleeding into the rendered page that shouldn't be visible.

More generally, regarding Service Mesh, it appears all our install docs there are pointing to upstream Maistra/Jaeger/Kiali bits and not the actual productized Service Mesh operators. Also, we reference Maistra TP12 but for our Serverless TP1 release this will need to be the Service Mesh GA release.

We also explicitly disable Kiali, Jaeger, and sidecar injection in our `ServiceMeshControlPlane` example. But the docs walk the user through installing the Jaeger and Kiali operators, which makes me think we should delete that part. Perhaps that's where the unused `modules/serverless-installing-service-mesh-no-kiali-jaeger.adoc` should be included here instead of the full Service Mesh install docs?

joaedwar:
> Lots of great stuff here - thanks! I've left a few comments with some screenshots of the rendered preview to see where some things are bleeding into the rendered page that shouldn't be visible.
> 
> More generally, regarding Service Mesh, it appears all our install docs there are pointing to upstream Maistra/Jaeger/Kiali bits and not the actual productized Service Mesh operators. Also, we reference Maistra TP12 but for our Serverless TP1 release this will need to be the Service Mesh GA release.
> 
> We also explicitly disable Kiali, Jaeger, and sidecar injection in our `ServiceMeshControlPlane` example. But the docs walk the user through installing the Jaeger and Kiali operators, which makes me think we should delete that part. Perhaps that's where the unused `modules/serverless-installing-service-mesh-no-kiali-jaeger.adoc` should be included here instead of the full Service Mesh install docs?

@bbrowning you're absolutely right the modules/serverless-installing-service-mesh-no-kiali-jaeger.adoc should be included to point the user away from installing Kiali, Jeager, and sidecar injection

adellape (approved):
Structural changes look good! LGTM, just a couple minor comments in-line.

PR #16365, opened 2019-08-22
Configuration docs for Kiali.

JStickler:
First pass at configuration docs for Kiali.

@brianredbeard, @mwringe, @gbaufake, @mattmahoneyrh, @hhovsepy


JStickler:
Note to reviewers, click OpenShift Container Platform, but don't select a version in the preview.  Just scroll down and click Service Mesh in the menu.  The content for this PR is under Day Two in the navigation.

mwringe:
We don't want the user to modify the Kiali CR directly. The Kiali CR is created and managed by the Maistra operator.

If a user modified the Kiali CR directly then their changes will be overwritten when the Maistra operator reconciles the Kiali CR.

PR #16367, opened 2019-08-23
CNV - Combining the VM and VM Template wizard fields modules

aburdenthehand:
This PR has no JIRA or bug tracker but has three improvements:
- Combining the 'Virtual machine wizard' and the 'Virtual machine template wizard' field modules so that there is a single entity to update [1]. This is shared across three assemblies. 
- Decomissioned the old template wizard fields module.
- Ensured the 'wizard field' modules accurately report the assemblies in which they are included.

This was something we wanted for the 2.0 (4.1) release but was not logistically possible at the time. 

[1] The 'Virtual machine wizard' includes two fields that are not in the 'Virtual machine template wizard'. An additional difference is that the 'Virtual machine template wizard' referred to 'virtual machines' plural whereas the 'Virtual machine wizard' refers to 'virtual machines' singular, because templates can be used to create multiple virtual machines. There was no elegant workaround for this so I left all reference as 'virtual machine' singular.


adellape (approved):
:+1: 	Looks good locally.

adellape:
/enterprise-4.2

PR #16374, opened 2019-08-23
osdocs-632 custom Azure installation

kalexand-rh:
Draft for https://jira.coreos.com/browse/OSDOCS-632 informed by https://godoc.org/github.com/openshift/installer/pkg/types/azure.

### DIFF
--- a/modules/installation-azure-config-yaml.adoc
+++ b/modules/installation-azure-config-yaml.adoc
@@ -1,17 +1,17 @@
 // Module included in the following assemblies:
 //
-// * installing/installing_aws/installing-aws-customizations.adoc
-// * installing/installing_aws/installing-aws-network-customizations.adoc
+// * installing/installing_azure/installing-azure-customizations.adoc
+// * installing/installing_azure/installing-azure-network-customizations.adoc
 
-ifeval::["{context}" == "installing-aws-network-customizations"]
+ifeval::["{context}" == "installing-azure-network-customizations"]
 :with-networking:
 endif::[]
-ifeval::["{context}" != "installing-aws-network-customizations"]
+ifeval::["{context}" != "installing-azure-network-customizations"]
 :without-networking:
 endif::[]
 
-[id="installation-aws-config-yaml_{context}"]
-= Sample customized `install-config.yaml` file for AWS
+[id="installation-azure-config-yaml_{context}"]
+= Sample customized `install-config.yaml` file for Azure
 
 You can customize the `install-config.yaml` file to specify more details about
 your {product-title} cluster's platform or modify the values of the required
@@ -31,29 +31,20 @@ controlPlane: <2>
   hyperthreading: Enabled <3> <4>
   name: master
   platform:
-    aws:
-      zones:
-      - us-west-2a
-      - us-west-2b
-      rootVolume:
-        iops: 4000
-        size: 500
-        type: io1
-      type: m5.xlarge <4>
+    azure:
+      osDisk:
+        diskSizeGB: 512 <5>
+      type: Standard_DS4_v2
   replicas: 3
 compute: <2>
 - hyperthreading: Enabled <3>
   name: worker
   platform:
-    aws:
-      rootVolume:
-        iops: 2000
-        size: 500
-        type: io1 <5>
-      type: c5.4xlarge
-      zones:
-      - us-west-2c
-  replicas: 3
+    azure:
+      osDisk:
+        diskSizeGB: 512 <5>
+      type: Standard_DS4_v
+  replicas: 5
 metadata:
   name: test-cluster <1>
 ifdef::without-networking[]
@@ -70,13 +61,11 @@ endif::[]
   serviceNetwork:
   - 172.30.0.0/16
 platform:
-  aws:
-    region: us-west-2 <1>
-    userTags:
-      adminContact: jdoe
-      costCenter: 7536
+  azure:
+    region: centralus <1>
+    baseDomainResourceGroupName: resource-group <6>
 pullSecret: '{"auths": ...}' <1>
-sshKey: ssh-ed25519 AAAA... <6>
+sshKey: ssh-ed25519 AAAA... <7>
 ----
 <1> Required. The installation program prompts you for this value.
 <2> If you do not provide these parameters and values, the installation program
@@ -91,7 +80,7 @@ installation. Only one control plane pool is used.
 <4> Whether to enable or disable simultaneous multithreading, or
 `hyperthreading`. By default, simultaneous multithreading is enabled
 to increase the performance of your machines' cores. You can disable it by
-setting the parameter value to `Disabled`. If you disable simultanous
+setting the parameter value to `Disabled`. If you disable simultaneous
 multithreading in some cluster machines, you must disable it in all cluster
 machines.
 +
@@ -102,9 +91,12 @@ accounts for the dramatically decreased machine performance. Use larger
 instance types, such as `m4.2xlarge` or `m5.2xlarge`, for your machines if you
 disable simultaneous multithreading.
 ====
-<5> To configure faster storage for etcd, especially for larger clusters, set the
-storage type as `io1` and set `iops` to `2000`.
-<6> You can optionally provide the `sshKey` value that you use to access the
+<5> You can specify the size of the disk to use in GB.
+//To configure faster storage for etcd, especially for larger clusters, set the
+//storage type as `io1` and set `iops` to `2000`.
+<6> Specify the name of the resource group that contains the DNS zone for your
+base domain.
+<7> You can optionally provide the `sshKey` value that you use to access the
 machines in your cluster.
 +
 [NOTE]
--- a/modules/installation-configuration-parameters.adoc
+++ b/modules/installation-configuration-parameters.adoc
@@ -2,14 +2,24 @@
 //
 // * installing/installing_aws/installing-aws-customizations.adoc
 // * installing/installing_aws/installing-aws-network-customizations.adoc
-// Consider also adding the installation-configuration-parameters.adoc module.
+// * installing/installing_azure/installing-azure-customizations.adoc
 
+ifeval::["{context}" == "install-customizations-cloud"]
+:aws:
+endif::[]
+ifeval::["{context}" == "installing-aws-network-customizations"]
+:aws:
+endif::[]
+ifeval::["{context}" == "installing-azure-customizations"]
+:azure:
+endif::[]
 
 [id="installation-configuration-parameters_{context}"]
 = Installation configuration parameters
 
 Before you deploy an {product-title} cluster, you provide parameter values to
-describe your Amazon Web Services (AWS) account and optionally customize your
+describe your account on the cloud platform that hosts your cluster
+and optionally customize your
 cluster's platform. When you create the `install-config.yaml` installation
 configuration file, you provide values for the required parameters through the
 command line. If you customize your cluster, you can modify the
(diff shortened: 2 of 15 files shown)


modules/installation-azure-config-yaml.adoc
```diff
@@ -0,0 +1,115 @@
...
+
+[source,yaml]
+----
+apiVersion: v1
+baseDomain: example.com <1>
+controlPlane: <2>
+  hyperthreading: Enabled <3> <4>
+  name: master
+  platform:
+    azure:
```
kalexand-rh: @sjug, do you have information about optimizing etcd for Azure? You provided the review for Ashley for the AWS equivalent in 4.1.

kalexand-rh:
@abhinavdahiya, @jhixson74, aside from the customization for etcd that I tagged @sjug on, am I on the right track with the custom parameters and yaml for Azure?

modules/installation-azure-config-yaml.adoc
```diff
@@ -0,0 +1,115 @@
...
+      zones:
```
abhinavdahiya: does not match the field docs on https://godoc.org/github.com/openshift/installer/pkg/types/azure

modules/installation-azure-config-yaml.adoc
```diff
@@ -0,0 +1,115 @@
...
+      - us-west-2a
+      - us-west-2b
+      rootVolume:
```
abhinavdahiya: this field doesn't exist see https://godoc.org/github.com/openshift/installer/pkg/types/azure

modules/installation-azure-config-yaml.adoc
```diff
@@ -0,0 +1,115 @@
...
+        iops: 4000
+        size: 500
+        type: io1
+      type: m5.xlarge <4>
```
abhinavdahiya: Azure has different types.

modules/installation-configuration-parameters.adoc
```diff
@@ -174,3 +181,19 @@ For more information about AWS tags,
...
 in the AWS documentation.
 |====
+
+
+.Optional Azure parameters
+[cols=".^2,.^3a,.^3a",options="header"]
+|====
+|Parameter|Description|Values
+
+|`platform.azure.baseDomainResourceGroupName`
```
abhinavdahiya: this is not optional.

modules/installation-configuration-parameters.adoc
```diff
@@ -174,3 +181,19 @@ For more information about AWS tags,
...
+|The name of the resource group that contains the DNS zone for your base domain.
+|String, for example `production_cluster`
+
+|`platform.azure.region`
```
abhinavdahiya: this is not optional

modules/installation-launching-installer-custom.adoc
```diff
@@ -0,0 +1,51 @@
...
+* Configure an account with the cloud provider to host your cluster.
+* Obtain the {product-title} installation program and the pull secret for your
+cluster.
+
+.Procedure
+
+. Run the installation program:
++
+----
+$ ./openshift-install create cluster --dir=<installation_directory> \ <1>
```
abhinavdahiya: why do we tell people to `--log-level debug` we should let them see what's expected with default which is info.

modules/installation-azure-config-yaml.adoc
kalexand-rh: Is there an equivalent modification that you can make in Azure, or is this customization not valid/required in Azure?

modules/installation-launching-installer-custom.adoc
kalexand-rh: It looks like the change came in after the review of the architecture content.

Do you want line 31 to read
```
    --log-level info <2>
```
or something else?

modules/installation-azure-config-yaml.adoc
kalexand-rh: It looks like it's [osDisk](https://github.com/openshift/installer/pull/2254/files#diff-42170ba9a400c7c4eb4884f5f069040fR6)

modules/installation-configuration-parameters.adoc
```diff
@@ -174,3 +181,27 @@ For more information about AWS tags,
...
+
+
+.Additional Azure parameters
+[cols=".^2,.^3a,.^3a",options="header"]
+|====
+|Parameter|Description|Values
+
+|`machines.platform.azure.type`
+|The Azure VM instance type.
+|Any valid VM type.
```
kalexand-rh: @jhixson74, re: https://github.com/openshift/installer/pull/2254#issuecomment-527235540, if we have limits on the types of VMs that are acceptable, I want to list them here.

modules/installation-launching-installer-custom.adoc
jhixson74: I'm assuming @abhinavdahiya means to just leave it off. When log level isn't specified, the default is info.

modules/installation-configuration-parameters.adoc
jhixson74: Azure can only run Windows & Linux. The specific versions are listed here: https://docs.microsoft.com/en-us/azure-stack/operator/azure-stack-supported-os?view=azs-1908

modules/installation-launching-installer-custom.adoc
kalexand-rh: I'm opening a separate PR to remove this option: https://github.com/openshift/openshift-docs/pull/16556

This will make it easier for me to apply the change to 4.1 as well.

modules/installation-configuration-parameters.adoc
kalexand-rh: I'm adding that link.

modules/installation-azure-config-yaml.adoc
sjug: Currently the only machinepool exposed instance configs are zone, disk size and instance type. So only instance type will potentially affect the etcd performance. It doesn't look like Azure instances decouple the storage and the rest of the instance type, so the AWS recommendations don't translate over exactly.

The disk specifications of the `Standard_DS5_v2` instance is closest to what we recommend for AWS.

modules/installation-configuration-parameters.adoc
```diff
@@ -174,3 +192,41 @@ For more information about AWS tags,
...
+
+|`platform.azure.region`
+|The name of the Azure region that hosts your cluster.
+|Any valid region name.
+
+|====
+
+[NOTE]
+====
+You cannot customize
```
e-tienne: Hello, can you please share more details about this? We have tests configured to use this custom configuration and I have also confirmed I can use the setting with the latest build. 

Example:
```json
compute:
- hyperthreading: Enabled
  name: worker
  platform:
    azure:
      osDisk:
        diskSizeGB: 128
      type: Standard_D4s_v3
      zones:
      - "1"
```

kalexand-rh: Let's see what happens with https://github.com/openshift/installer/pull/2371?

kalexand-rh:

modules/installation-azure-config-yaml.adoc
```diff
@@ -0,0 +1,107 @@
...
+= Sample customized `install-config.yaml` file for Azure
+
+You can customize the `install-config.yaml` file to specify more details about
+your {product-title} cluster's platform or modify the values of the required
+parameters.
+
+[IMPORTANT]
+====
+This sample YAML file is provided for reference only. You must obtain your
+`install-config.yaml` file by using the installation program and modify it.
```
bobfuru: s/modify/modifying

modules/installation-azure-config-yaml.adoc
```diff
@@ -0,0 +1,107 @@
...
+sequence of mappings. To meet the requirements of the different data structures,
+the first line of the `compute` section must begin with a hyphen, `-`, and the
+first line of the `controlPlane` section must not. Although both sections
+currently define a single machine pool, it is possible that future versions
+of {product-title} will support defining multiple compute pools during
+installation. Only one control plane pool is used.
+<4> Whether to enable or disable simultaneous multithreading, or
+`hyperthreading`. By default, simultaneous multithreading is enabled
+to increase the performance of your machines' cores. You can disable it by
+setting the parameter value to `Disabled`. If you disable simultanous
```
bobfuru: s/simultanous/simultaneous

modules/installation-azure-config-yaml.adoc
```diff
@@ -0,0 +1,107 @@
...
+//To configure faster storage for etcd, especially for larger clusters, set the
+//storage type as `io1` and set `iops` to `2000`.
+<6> Specify the name of the resource group that contains the DNS zone for your
+base domain.
+<7> You can optionally provide the `sshKey` value that you use to access the
+machines in your cluster.
++
+[NOTE]
+====
+For production {product-title} clusters on which you want to perform installation
```
bobfuru: For better readability, I suggest, "For production {product-title} clusters you want to perform installation debugging or disaster recovery on, you must provide..."

modules/installation-azure-config-yaml.adoc
```diff
@@ -0,0 +1,107 @@
...
+debugging or disaster recovery, you must provide an SSH key that your `ssh-agent`
+process uses to the installation program.
```
bobfuru: Should it be "uses to _access (or authenticate to)_ the installation program"?

modules/installation-configuration-parameters.adoc
```diff
@@ -174,3 +192,41 @@ For more information about AWS tags,
...
+
+|`machines.platform.azure.type`
+|The Azure VM instance type.
+|VMs that use Windows or Linux as the operating system. See the
+link:https://docs.microsoft.com/en-us/azure-stack/operator/azure-stack-supported-os?view=azs-1908[Guest operating systems supported on Azure Stack]
+in the Azure documentation.
+
+`machines.platform.azure.osDisk.diskSizeGB`
+|The Azure disk size for the VM.
+|Integer that represents the size of the disk in GB, for example `512`. The
```
bobfuru: s/GB, for example `512`./GB. For example, `512`.

modules/installation-configuration-parameters.adoc
```diff
@@ -2,14 +2,24 @@
...
+:azure:
+endif::[]
 
 [id="installation-configuration-parameters_{context}"]
 = Installation configuration parameters
 
 Before you deploy an {product-title} cluster, you provide parameter values to
-describe your Amazon Web Services (AWS) account and optionally customize your
+describe your account on the cloud platform that hosts your cluster
+and optionally customize your
```
bobfuru: s/and optionally customize/and optionally to customize

modules/installation-configuration-parameters.adoc
```diff
@@ -102,6 +113,32 @@ accounts for the dramatically decreased machine performance.
 ====
 |`Enabled` or `Disabled`
 
+|`compute.replicas`
+|The number of compute, or worker, machines to provision.
+|A positive integer greater than or equal to `2`. The default value is `3`.
+
+|`controlPlane.hyperthreading`
+|Whether to enable or disable simultaneous multithreading, or `hyperthreading`
```
bobfuru: s/`hyperthreading` on/`hyperthreading`, on

bobfuru (commented):
Minor edits; otherwise, LGTM.

modules/installation-azure-config-yaml.adoc
kalexand-rh: No, you don't use it to access the installation program. You have to provide it to the installation program so that it can seed your machines with it.

modules/installation-configuration-parameters.adoc
kalexand-rh: @bobfuru, what's your source? The only guidance I can find in the IBM Style guide is that you need to use "for example." It does not provide a suggested punctuation and generally avoids creating a fragment.

modules/installation-azure-config-yaml.adoc
bobfuru: I see. It might make more sense to say that "you must provide the installation program with an SSH key that your `ssh-agent` process uses." Or not, if that changes the meaning. Pretty minor, so probably fine like it is.

modules/installation-configuration-parameters.adoc
bobfuru: "For example" should be followed by a comma. As written, it's also a bit of a run-on. Would also be okay to write "...in GB (for example, `512`)." Although I try to avoid parentheses that aren't part of code.
I can't find a stated guideline in IBM or other guides, though most of thehir own refs use it at the start of a sentence.
But as written, it is short, to the point, and clear, so if that's a convention we use, then ignore my ^^ comments altogether. :)

modules/installation-configuration-parameters.adoc
e-tienne: This has been merged, but I do not the availability zones in your example.

vikram-redhat:
@e-tienne besides the comment here (https://github.com/openshift/openshift-docs/pull/16374#discussion_r328820058), is this ok from QE's POV?

e-tienne:
LGTM

PR #16387, opened 2019-08-26
Bug BZ1725931 Certificate Background Info

jhoyt-rh:
For BUG BZ1725931: For Closed Loop Escalation Customer Support Case 02374412: Customer wants all information in the docs in one place. Initially, request was to modify https://docs.openshift.com/container-platform/3.11/install_config/configuring_authentication.html with additional, background information. 

However, to include everything from our docs in one place, we would have to redesign a number of books, which we are not going to do at this time, especially going back to release 3.11.

As a compromise, I'm suggesting adding the 292 new lines shown below as introductory/preparatory material to the certificate_customization.adoc file instead. 

Note: there are many cross-references in the BZ from Red Hat TAM Silvio Perez Torres to other files he sent the customer, and those could be added as links in the configuring_authentication.adoc file if it is determined to do so. I have not added all those references here.

silvio.perez@redhat.com, will you PTAL at this file, and see what you recommend for further information required, especially in the certificate_customization file mentioned above.

### DIFF
--- a/install_config/certificate_customization.adoc
+++ b/install_config/certificate_customization.adoc
@@ -1,5 +1,5 @@
 [[install-config-certificate-customization]]
-= Configuring Custom Certificates
+= Managing and Configuring Custom Certificates
 {product-author}
 {product-version}
 :data-uri:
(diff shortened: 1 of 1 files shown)


vikram-redhat:
The associated bug is closed. I am closing this PR. If need be, please open a new one.

PR #16392, opened 2019-08-26
[WIP] PROD-615-proxy-enablement

bmcelvee:
https://jira.coreos.com/browse/OSDOCS-468

### DIFF
--- a/installing/install_config/enable-cluster-wide-proxy.adoc
+++ b/installing/install_config/enable-cluster-wide-proxy.adoc
@@ -1,0 +1,29 @@
+[id="enable-cluster-wide-proxy"]
+= Enable cluster-wide proxy
+include::modules/common-attributes.adoc[]
+:context: enable-cluster-wide-proxy
+
+toc::[]
+
+The `Proxy` object is used to manage cluster-wide egress proxy. When a cluster is
+installed or upgraded without proxy configured, a proxy object is still
+generated but it will have a nil spec/status. For example:
+
+----
+$ oc get proxy/cluster -o yaml
+apiVersion: config.openshift.io/v1
+kind: Proxy
+metadata:
+  creationTimestamp: "2019-08-21T22:36:49Z"
+  generation: 2
+  name: cluster
+  resourceVersion: "24913"
+  selfLink: /apis/config.openshift.io/v1/proxies/cluster
+  uid: <user ID string>
+spec:
+  trustedCA:
+    name: ""
+status:
+----
+
+include::modules/installation-proxy-configure-object.adoc[leveloffset=+1]
--- a/modules/installation-proxy-configure-object.adoc
+++ b/modules/installation-proxy-configure-object.adoc
@@ -1,0 +1,50 @@
+// Module included in the following assemblies:
+//
+// * installing/insall_config/enable-cluster-wide-proxy.adoc
+
+[id="installation-proxy-configure-object_{context}"]
+= Configure the Proxy object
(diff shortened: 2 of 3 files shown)


danehans:
I think it would be beneficial to add a proxy section under networking that leverages the details of [this doc](https://gist.github.com/danehans/2c7321711d10ae100c986ef75bc6d213#file-proxy_doc-md). I think it's important for people to understand what the controllers do, which operator runs the controller, how the CA injection stuff works, etc..

modules/installation-proxy-configure-object.adoc
```diff
@@ -0,0 +1,50 @@
...
+kind: Proxy
+metadata:
+  creationTimestamp: "2019-08-21T22:36:49Z"
+  generation: 2
+  name: cluster
+  resourceVersion: "24913"
+  selfLink: /apis/config.openshift.io/v1/proxies/cluster
+  uid: <user ID string>
+spec:
+  httpProxy: http://admin:admin@35.196.128.173:3128
```
danehans: Please change so the doc is not referencing our test proxy.

modules/installation-proxy-configure-object.adoc
```diff
@@ -0,0 +1,50 @@
...
+  httpsProxy: https://admin:admin@35.231.5.161:3128
```
danehans: Please change so the doc is not referencing our test proxy.

modules/installation-proxy-configure-object.adoc
```diff
@@ -0,0 +1,50 @@
...
+  noProxy: example.com
+  readinessEndpoints:
+  - http://www.google.com
+  - https://www.google.com
+  trustedCA:
+    name: user-ca-bundle
+status:
+  httpProxy: http://admin:admin@35.196.128.173:3128
```
danehans: Please change so the doc is not referencing our test proxy.

modules/installation-proxy-configure-object.adoc
```diff
@@ -0,0 +1,50 @@
...
+  httpsProxy: https://admin:admin@35.231.5.161:3128
```
danehans: Please change so the doc is not referencing our test proxy.

danehans:
Due to https://bugzilla.redhat.com/show_bug.cgi?id=1743174, a docs note is needed stating: The install config `machineCIDR` directive is used to bypass proxy for internal cluster connections. For deployments such as bare-metal that do not use `machineCIDR`, one of the following steps should be completed for proxy to function properly:

1. Cluster node IP's should be assigned from `machineCIDR`.
2. `machineCIDR` should be updated to reflect the CIDR being used for cluster nodes
3. Add cluster node IPs or CIDR to `noProxy` of proxy "cluster".

installing/install_config/enable-cluster-wide-proxy.adoc
```diff
@@ -0,0 +1,29 @@
...
+  name: cluster
+  resourceVersion: "24913"
+  selfLink: /apis/config.openshift.io/v1/proxies/cluster
+  uid: <user ID string>
+spec:
+  trustedCA:
+    name: ""
+status:
+----
+
```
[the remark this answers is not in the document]
kalexand-rh: I think we'll need a separate assembly to cover the installation with proxy piece. @bmcelvee, I might have time to make that next week.

bergerhoffer:
Closing this PR in favor of #16562.

PR #16401, opened 2019-08-27
adding jq package prereq

kalexand-rh:
@wmengRH, will you PTAL? Since we use jq in the command at line 84, I think we need to make that package a prereq for these steps.


wmengRH:
LGTM.

PR #16422, opened 2019-08-28
CR support for setting tolerations on logging components

mburke5678:
https://jira.coreos.com/browse/OSDOCS-511


mburke5678:
@ewolinetz @anpingli Please take a look. The changes aren't as big as indicated. There is repetition in each topic, such as the procedure.

mburke5678:
@anpingli PTAL

mburke5678:
@anpingli Please take a look.

modules/efk-logging-elasticsearch-tolerations.adoc
```diff
@@ -0,0 +1,67 @@
...
+  operator: "Exists"
+----
+
+.Prerequisites
+
+* Cluster logging and Elasticsearch must be installed.
+
+.Procedure
+
+. Use the following command to add a taint to a node where you want the Elasticsearch pods scheduled:
```
ewolinetz: i'm not sure if this is the proper use of tainting nodes...
we would want to add a toleration if we needed to ensure that the component's pod can be deployed no matter what

ewolinetz:
other than how we're describing to use this feature LGTM.
I don't think tainting nodes is the proper usage for this with regards to scheduling pods in certain locations -- that is more a nodeSelector IMO

modules/efk-logging-elasticsearch-tolerations.adoc
mburke5678: Changed to: _Use the following command to add taints and tolerations for cluster logging_
I added to the intro paragraph the use case in the Jira:
_You can use taints and tolerations to ensure that cluster logging pods 
run on specific nodes and that no other workload can run on those nodes._

modules/efk-logging-curator-tolerations.adoc
```diff
@@ -0,0 +1,57 @@
+// Module included in the following assemblies:
+//
+// * logging/efk-logging-curator.adoc
+
+[id="efk-logging-curator-tolerations_{context}"]
+= Using tolerations to control the Curator pod placement
+
+You can ensure which node the Curator pod runs on
```
anpingli: it looks not very clear. Something like "You can control on which node the Curator pod runs using tolerations on the pod" sounds clear.

modules/efk-logging-fluentd-tolerations.adoc
```diff
@@ -0,0 +1,67 @@
+// Module included in the following assemblies:
+//
+// * logging/efk-logging-fluentd.adoc
+
+[id="efk-logging-fluentd-tolerations_{context}"]
+= Using tolerations to control the log collector pod placement
+
+You can ensure which nodes the logging collector pods run on
```
anpingli: @mburke5678 I don't think we need  taint and tolerations to controll the collection pods. @jcantrill  How do you think?

modules/efk-logging-kibana-tolerations.adoc
```diff
@@ -0,0 +1,58 @@
+// Module included in the following assemblies:
+//
+// * logging/efk-logging-kibana.adoc
+
+[id="efk-logging-kibana-tolerations_{context}"]
+= Using tolerations to control the Kibana pod placement
+
+You can ensure which nodes the Kibana pod runs on
```
anpingli: it looks not very clear. Something like "You can control on which node the kibana pod runs using tolerations on the pod" sounds clear

modules/efk-logging-fluentd-tolerations.adoc
mburke5678: @ewolinetz Any thoughts ^^

mburke5678:

modules/efk-logging-fluentd-tolerations.adoc
jcantrill: IMO, collectors are supposed to run on every node.  Isn't part of taints and tolerations also to ensure the pod doesn't get evicted for things like memory and cpu pressures? If so then this would be a way to guard against those situations

mburke5678: Should we specify this in the doc:
"pods that tolerate the taint without specifying tolerationSeconds in their toleration specification remain bound forever"
https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/#taint-based-evictions

mburke5678: @anpingli PTAL at the comments above. 
Also, you can see the conversations in Slack:
https://coreos.slack.com/archives/GGUR75P60/p1568903576072500
https://coreos.slack.com/archives/GGUR75P60/p1568903649073100

mburke5678:
Closed in error dealing with `efk` to `cluster` file name changes.

PR #16453, opened 2019-08-30
Remove the outdated compatibility section

rh-max:
@bmcelvee Could you take a look please? If it looks fine, we should also cherrypick it into master-3.


vikram-redhat:
@bmcelvee bump

vikram-redhat:
@rh-max should this first go into master-3?

rh-max:
@vikram-redhat Hmm, I didn't think it matters with cherry-picking, as long as the change does end up in both branches? But I'll default to master-3 for my new PRs (I also just tried to change base of this PR, and it's not worth the trouble).

vikram-redhat:
@rh-max yes that is correct, you can CP to and from any branch :)

PR #16508, opened 2019-09-05
CNV-OCP monitoring/telemetry modularization for associated products

aburdenthehand:
This was deferred from the CNV 2.0 release.
Separating introduction content from OCP into modules to allow for content re-use with associated products. Using conditionals so that the now-modules do not include the (otherwise redundant) titles for the original OpenShift assemblies.

### DIFF
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -1009,7 +1009,7 @@ Topics:
     File: cnv-logs
   - Name: Viewing events
     File: cnv-events
-  - Name: OpenShift cluster monitoring
+  - Name: OpenShift cluster monitoring, logging, and Telemetry
     File: cnv-openshift-cluster-monitoring
 - Name: Container-native Virtualization 2.0 release notes
   Dir: cnv_release_notes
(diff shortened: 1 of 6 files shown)


aburdenthehand:
@rh-max Can you please review:
OCP docs:
* https://cnv-ocp-monitoring--ocpdocs.netlify.com/openshift-enterprise/latest/telemetry/about-telemetry.html
* https://cnv-ocp-monitoring--ocpdocs.netlify.com/openshift-enterprise/latest/monitoring/cluster-monitoring/about-cluster-monitoring.html
CNV docs:
* https://cnv-ocp-monitoring--ocpdocs.netlify.com/openshift-enterprise/latest/cnv/cnv_users_guide/cnv-openshift-cluster-monitoring.html

For the CNV side the mismatch of titles is a bit jarring. What are your thoughts on naming the Telemetry module 'About Telemetry'?

rh-max:
@aburdenthehand  Looks very good! I only have two notes:
* Sure, do change to "About Telemetry" (be sure to change filenames and `include` statements too). I'd also consider changing the titles of "Cluster logging" and maybe even "CLI troubleshooting and debugging commands".
* You left the `:ocp-monitoring:` context. Unless we use it later, how about undefining it using `:ocp-monitoring!:`? The same goes for `:ocp-telemetry:`.
* Good idea on including a note on LOC 6-8 in those modules!

aburdenthehand:
@rh-max thanks mate
- Still working on the "Cluster logging" part and will raise a separate PR for it. Otherwise updated as suggested
- Good call. Updated as suggested.
Ready to be cherrypicked to 4.1 and 4.2

sheriff-rh (approved):

sheriff-rh:
Peer reviewed - Looks good! Merging and picking to 4.1 and 4.2.

PR #16513, opened 2019-09-05
improve node selector docs

mburke5678:
Changing the node selector doc to move labeling the node first.

### DIFF
--- a/modules/nodes-scheduler-node-selectors-pod.adoc
+++ b/modules/nodes-scheduler-node-selectors-pod.adoc
@@ -7,44 +7,53 @@
 
 You can use node selector labels on pods to control where the pod is scheduled.
 
-You then add labels to a specific nodes where you want the pods scheduled or to the MachineSet that controls the nodes. 
+With node selectors, {product-title} schedules the pods on nodes that contain matching labels. 
 
 You can add labels to a node or MachineConfig, but the labels will not persist if the node or machine goes down. 
 Adding the label to the MachineSet ensures that new nodes or machines will have the label.
 
-.Procedure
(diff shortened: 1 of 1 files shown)


mburke5678:
@sunilcio You have looked at node selector docs before. Can you take a look at this small change?

If so, do you know why I cannot add a node selector to a pod in 4.1?
I am using a cluster-bot system with a recent 4.1 CI build.
```
$ oc whoami
system:admin
```
I tried editing a pod spec to add a node selector but was not allowed to make the change:
````
# pods "router-default-65b79fd87b-kpqtd" was not valid:
# * spec: Forbidden: pod updates may not change fields other than `spec.containers[*].image`, `spec.initContainers[*].image`, `spec.activeDeadlineSeconds` or `spec.tolerations` (only additions to existing tolerations)
````
Similar result in the web console as `kube:admin`

mburke5678:
@sunilcio You have looked at node selector docs before. Can you take a look at this small change?

sunilcio:
@mburke5678 I do not think we can edit nodeSelector in pod spec after it is already scheduled and  created. In the above error is reporting what can be changed.
@sjenning could you please check if this is expected?

mburke5678:
@sunilcio According to @joelsmith node selectors for existing scheduled pods need to be added to the pod's controlling  (owner) object. 
See my changes https://github.com/openshift/openshift-docs/pull/16513/commits/a64c55dd5a70a170fbeca015f5787c96c58d186f

vikram-redhat:
@sunilcio could you confirm if you are happy with this change?

sunilcio:
@vikram-redhat LGTM, thanks.

PR #16545, opened 2019-09-09
SRVCOM-546: Edits to install docs

nebelsprachen:
Labels:

* branch/enterprise-4.1
* branch/enterprise-4.2
------------------------------------------
* Minor edits
* Point install docs to Service Mesh install docs


serverless/installing-openshift-serverless.adoc
```diff
@@ -12,8 +12,10 @@ include::modules/machineset-manually-scaling.adoc[leveloffset=+2]
 
 
 // Installing Service Mesh
-include::modules/serverless-installing-service-mesh-no-kiali-jaeger.adoc[leveloffset=+1]
+// Commented out for this release. May apply for future releases 
+// include::modules/serverless-installing-service-mesh-no-kiali-jaeger.adoc[leveloffset=+1]
 
+OpenShift Serverless requires an installed version of xref:../modules/ossm-operatorhub-install.adoc[Service Mesh].
```
jboxman: This `xref` doesn't build locally. I think the link needs to be to the correct assembly instead. Maybe:

> OpenShift Serverless requires an installed version of xref:../service_mesh/service_mesh_install/preparing-ossm-installation.adoc#preparing-ossm-installation[Service Mesh].

joaedwar:
@mgencur Is this what you had in mind?

serverless/installing-openshift-serverless.adoc
```diff
@@ -12,8 +12,10 @@ include::modules/machineset-manually-scaling.adoc[leveloffset=+2]
 
 
 // Installing Service Mesh
-include::modules/serverless-installing-service-mesh-no-kiali-jaeger.adoc[leveloffset=+1]
+// Commented out for this release. May apply for future releases 
+// include::modules/serverless-installing-service-mesh-no-kiali-jaeger.adoc[leveloffset=+1]
 
+OpenShift Serverless requires an installed version of xref:../service_mesh/service_mesh_install/preparing-ossm-installation.adoc#preparing-ossm-installation[Service Mesh].
```
mgencur: This document points to the whole Service Mesh installation which includes installing ServiceMeshControlPlan and ServiceMeshMemberRoll. But we also include these steps in our documentation. 
I'd say it would be better to include just what is necessary from the Service Mesh docs and that is installing their operators: https://github.com/openshift/openshift-docs/blob/master/modules/ossm-operatorhub-install.adoc
And since we're deleting the file which had a note about creating the istio-system namespace and this step is required, I'd insert this step in https://github.com/openshift/openshift-docs/blob/master/modules/serverless-installing-service-mesh-control-plane.adoc just before the last step which applies the smcp.yaml

mgencur:
LGTM

modules/serverless-installing-service-mesh-control-plane.adoc
```diff
@@ -84,11 +88,22 @@ Running ServiceMesh with a sidecar injection enabled with OpenShift Serverless i
 * The ServiceMesh operator must be installed. 
 
 .Procedure
+. Log into your {product-title} installation as a cluster administrator.
```
ahardin-rh: s/Log into/Log in to

ahardin-rh:
Just one minor comment from me. Otherwise, looks great!

PR #16547, opened 2019-09-10
Fix typo in Serverless Service Mesh config

mgencur:


mgencur:
It would be also good to remove the remaining <1> in the config so that it's easy to copy&paste this config directly. The good way to do this is to have the file committed separately and referenced via link, and then have the example including these bullets as well. A good example is the RedHat developer tutorials: https://redhat-developer-demos.github.io/knative-tutorial/knative-tutorial-basics/0.7.x/05-eventing/eventing-src-svc.html#eventing-create-event-source

PR #16556, opened 2019-09-10
removing the --log-level debug option

kalexand-rh:
@abhinavdahiya, based your [comments](https://github.com/openshift/openshift-docs/pull/16374#discussion_r317220248) on the Azure draft, would you like this option removed from all of the installation docs?

### DIFF
--- a/modules/installation-aws-user-infra-bootstrap.adoc
+++ b/modules/installation-aws-user-infra-bootstrap.adoc
@@ -25,12 +25,10 @@ you can install the cluster.
 following command:
 +
 ----
-$ ./openshift-install wait-for bootstrap-complete --dir=<installation_directory> \ <1>
-    --log-level debug <2>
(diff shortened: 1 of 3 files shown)


abhinavdahiya:
So 2 things..
1) making sure most users are running at `info` level because that's designed for best UX.
2) informing users that they can indeed set log-level to something other than `info`, like `debug | info | warn | error` .

So maybe switching the docs to set `--log-level info <2>`  and then telling them `<2> you can set the log levels to these other values.... and info is the default one if you skip that flag` ..

if that becomes too troublesome, just dropping it like we do here is fine with me. :)

kalexand-rh:
@jianlinliu, will you PTAL or tag someone else from QE?

jianlinliu:
LGTM

kalexand-rh:

mburke5678:
@kalexand-rh LGTM

PR #16562, opened 2019-09-10
PROD-615-proxy-enablement

bergerhoffer:
This PR will replace #16392.

### DIFF
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -310,6 +310,8 @@ Topics:
     File: configuring-ingress-cluster-traffic-service-external-ip
   - Name: Configuring ingress cluster traffic using a NodePort
     File: configuring-ingress-cluster-traffic-nodeport
+- Name: Configuring the cluster-wide proxy
+  File: enable-cluster-wide-proxy
 ---
 Name: Storage
 Dir: storage
--- a/networking/enable-cluster-wide-proxy.adoc
+++ b/networking/enable-cluster-wide-proxy.adoc
@@ -1,0 +1,17 @@
+[id="enable-cluster-wide-proxy"]
+= Configuring the cluster-wide proxy
+include::modules/common-attributes.adoc[]
+:context: config-cluster-wide-proxy
+
+toc::[]
+
+Production environments can deny direct access to the Internet and instead have
+an HTTP or HTTPS proxy available. You can configure {product-title} to use a
+proxy by
+xref:../networking/enable-cluster-wide-proxy.adoc#nw-proxy-configure-object_config-cluster-wide-proxy[modifying the Proxy object for existing clusters]
+or by configuring the proxy settings in the `install-config.yaml` file for new
+clusters.
+
+include::modules/nw-proxy-configure-object.adoc[leveloffset=+1]
+
+include::modules/nw-proxy-remove.adoc[leveloffset=+1]
--- a/modules/nw-proxy-configure-object.adoc
+++ b/modules/nw-proxy-configure-object.adoc
@@ -1,0 +1,107 @@
+// Module included in the following assemblies:
+//
+// * networking/enable-cluster-wide-proxy.adoc
+
+[id="nw-proxy-configure-object_{context}"]
+= Enabling the cluster-wide proxy
+
+The Proxy object is used to manage the cluster-wide egress proxy. When a cluster is
+installed or upgraded without the proxy configured, a Proxy object is still
+generated but it will have a nil `spec`. For example:
+
+[source,yaml]
+----
+apiVersion: config.openshift.io/v1
+kind: Proxy
+metadata:
+  name: cluster
+spec:
+  trustedCA:
+    name: ""
+status:
+----
+
+A cluster administrator can configure the proxy for {product-title} by modifying
+this `cluster` Proxy object.
+
+NOTE: Only the Proxy object named `cluster` is supported, and no additional
+proxies can be created.
+
+.Prerequisites
+
+* Cluster administrator permissions
+* {product-title} `oc` CLI tool installed
+
+.Procedure
+
+. Create a ConfigMap that contains any additional CA certificates required for
+proxying HTTPS connections.
++
+NOTE: You can skip this step if the proxy’s identity certificate is signed by an
+authority from the RHCOS trust bundle.
+
+.. Create a file called `user-ca-bundle.yaml` with the following contents, and provide the values of your PEM-encoded certificates:
++
+[source,yaml]
+----
+apiVersion: v1
+data:
+  ca-bundle.crt: | <1>
+    <MY_PEM_ENCODED_CERTS> <2>
+kind: ConfigMap
+metadata:
+  name: user-ca-bundle <3>
+  namespace: openshift-config <4>
+----
+<1> This data key must be named `ca-bundle.crt`.
+<2> One or more PEM-encoded X.509 certificates used to sign the proxy's
+identity certificate.
+<3> The ConfigMap name that will be referenced from the Proxy object.
+<4> The ConfigMap must be in the `openshift-config` namespace.
+
+.. Create the ConfigMap from this file:
++
+----
+$ oc create -f user-ca-bundle.yaml
+----
+
+. Use the `oc edit` command to modify the Proxy object:
++
+----
+$ oc edit proxy/cluster
+----
+
+. Configure the necessary fields for the proxy:
++
+[source,yaml]
+----
+apiVersion: config.openshift.io/v1
+kind: Proxy
+metadata:
+  name: cluster
+spec:
+  httpProxy: http://<username>:<pswd>@<ip>:<port> <1>
+  httpsProxy: https://<username>:<pswd>@<ip>:<port> <2>
+  noProxy: example.com <3>
+  readinessEndpoints:
+  - http://www.google.com <4>
+  - https://www.google.com
+  trustedCA:
+    name: user-ca-bundle <5>
+----
+<1> A proxy URL to use for creating HTTP connections outside the cluster.
+<2> A proxy URL to use for creating HTTPS connections outside the cluster. If
+this is not specified, then `httpProxy` is used for both HTTP and HTTPS
+connections.
+<3> A comma-separated list of destination domain names, domains, IP addresses or
+other network CIDRs to exclude proxying. Preface a domain with `.` to include
+all subdomains of that domain. Use `*` to bypass proxy for all destinations.
+<4> One or more URLs external to the cluster to use to perform a readiness check
+before writing the `httpProxy` and `httpsProxy` values to status.
+<5> A reference to the ConfigMap in the `openshift-config` namespace that
+contains additional CA certificates required for proxying HTTPS connections.
+Note that the ConfigMap must already exist before referencing it here. This
+field is required unless the proxy's identity certificate is signed by an
+authority from the RHCOS trust bundle.
+
+. Save the file to apply the changes.
--- a/modules/nw-proxy-remove.adoc
+++ b/modules/nw-proxy-remove.adoc
@@ -1,0 +1,36 @@
+// Module included in the following assemblies:
+//
+// * networking/enable-cluster-wide-proxy.adoc
+
+[id="nw-proxy-remove_{context}"]
+= Removing the cluster-wide proxy
+
+The `cluster` Proxy object cannot be deleted. To remove the proxy from a cluster,
+remove all `spec` fields from the Proxy object.
+
+.Prerequisites
+
+* Cluster administrator permissions
+* {product-title} `oc` CLI tool installed
+
+.Procedure
+
+. Use the `oc edit` command to modify the proxy:
++
+----
+$ oc edit proxy/cluster
+----
+
+. Remove all `spec` fields from the Proxy object. For example:
++
+[source,yaml]
+----
+apiVersion: config.openshift.io/v1
+kind: Proxy
+metadata:
+  name: cluster
+spec: {}
+status: {}
+----
+
+. Save the file to apply the changes.


bergerhoffer:
@danehans Can you please take a look?

installing/install_config/enable-cluster-wide-proxy.adoc
```diff
@@ -0,0 +1,16 @@
+[id="enable-cluster-wide-proxy"]
+= Configuring cluster-wide proxy
+include::modules/common-attributes.adoc[]
+:context: enable-cluster-wide-proxy
+
+toc::[]
+
+Production environments can deny direct access to the Internet and instead have
```
bergerhoffer: @danehans I pulled this paragraph from the 3.11 docs, but wasn't sure if we needed to change the "as simple as setting standard environment variables in configuration or JSON files" part for 4.2. Suggestions?

bergerhoffer:

_topic_map.yml
```diff
@@ -142,6 +142,8 @@ Topics:
     Distros: openshift-enterprise,openshift-origin
   - Name: Configuring your firewall
     File: configuring-firewall
+  - Name: Configuring cluster-wide proxy
```
jboxman: _Configuring cluster-wide proxy support_ or _Configuring the cluster-wide proxy_?

modules/installation-proxy-configure-object.adoc
```diff
@@ -0,0 +1,121 @@
+// Module included in the following assemblies:
+//
+// * installing/insall_config/enable-cluster-wide-proxy.adoc
+
+[id="installation-proxy-configure-object_{context}"]
+= Enabling cluster-wide proxy
```
jboxman: s/Enabling cluster-wide proxy/Enabling a cluster-wide proxy/

modules/installation-proxy-remove.adoc
```diff
@@ -0,0 +1,43 @@
+// Module included in the following assemblies:
+//
+// * installing/insall_config/enable-cluster-wide-proxy.adoc
+
+[id="installation-proxy-remove_{context}"]
+= Removing proxy from the cluster
```
jboxman: s/Removing proxy from the cluster/Removing the cluster-wide proxy/

modules/installation-proxy-remove.adoc
```diff
@@ -0,0 +1,43 @@
...
+$ oc edit proxy/cluster
+----
+
+. Remove all `spec` fields from the proxy object. For example:
++
+[source,yaml]
+----
+apiVersion: config.openshift.io/v1
+kind: Proxy
+metadata:
```
jboxman: In all my examples, I've removed the metadata specific to the running cluster, usually leaving just this:

```
metadata:
  name: cluster
```

installing/install_config/enable-cluster-wide-proxy.adoc
```diff
@@ -0,0 +1,16 @@
...
+an HTTP or HTTPS proxy available. Configuring {product-title} to use these
+proxies can be as simple as setting standard environment variables in
```
jboxman: s/these proxies/a cluster-wide proxy/

modules/installation-proxy-configure-object.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+The proxy object is used to manage cluster-wide egress proxy. When a cluster is
+installed or upgraded without proxy configured, a proxy object is still
```
jboxman: s/without proxy/without a proxy/

modules/installation-proxy-configure-object.adoc
```diff
@@ -0,0 +1,121 @@
...
+generated but it will have a nil spec and status. For example:
```
jboxman: s/spec/`spec`/
s/status/`status`/

Because of how the k8s API works, the `status` parameter provides the current state of the resource. I don't think a user can ever directly configure the `status` parameter, only the `spec` parameter. The `spec` parameter is where a user configures the expected state of the system, and `status` reports the current state.

So you may not need to mention anything about it specifically.

Also, in this example, the `spec` is not `nil`. It's a mapping that's basically:

```json
{
  "trustedCA": {
    "name": ""
  }
}
```
So the `name` parameter is set to the empty string.

modules/installation-proxy-configure-object.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+[source,yaml]
+----
+apiVersion: config.openshift.io/v1
+kind: Proxy
+metadata:
```
jboxman: I would condense to:

```
metadata:
  name: cluster
```

modules/installation-proxy-remove.adoc
```diff
@@ -0,0 +1,43 @@
...
+
+.Procedure
+
+. Use the `oc edit` command to modify the proxy:
++
+----
+$ oc edit proxy/cluster
+----
+
+. Remove all `spec` fields from the proxy object. For example:
```
jboxman: So I think the final object is simply:

```
kind: Proxy
metadata:
  name: cluster
spec:
  trustedCA:
    name: ""
```

modules/installation-proxy-configure-object.adoc
```diff
@@ -0,0 +1,121 @@
...
+spec:
+  httpProxy: http://<username>:<pswd>@<ip>:<port> <1>
+  httpsProxy: https://<username>:<pswd>@<ip>:<port> <2>
+  noProxy: example.com <3>
+  readinessEndpoints:
+  - http://www.google.com <4>
+  - https://www.google.com
+  trustedCA:
+    name: user-ca-bundle <5>
+status:
```
jboxman: I don't think you can configure the `status` parameter. It informs on the current status of the resource as the cluster API sees it.

modules/installation-proxy-configure-object.adoc
```diff
@@ -0,0 +1,121 @@
...
+  resourceVersion: "24913"
+  selfLink: /apis/config.openshift.io/v1/proxies/cluster
+  uid: 2a344b01-c464-11e9-b3f3-025de4b59c38
+spec:
+  trustedCA:
+    name: ""
+status:
+----
+
+A cluster administrator can configure proxy for {product-title} by modifying
```
jboxman: s/can configure proxy/can configure a proxy/

modules/installation-proxy-configure-object.adoc
```diff
@@ -0,0 +1,121 @@
...
+this `cluster` proxy object.
+
+NOTE: Only the proxy object named `cluster` is supported, and no additional
+proxies can be created.
+
+.Prerequisites
```
jboxman: I've been using:

```
* Install the OpenShift Command-line Interface (CLI), commonly known as `oc`.
* You must log in to the cluster as a cluster administrator.
```

Did we ever decide upon a common style for these frequently used prerequisites?

modules/installation-proxy-configure-object.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+* Cluster administrator permissions
+* {product-title} `oc` CLI tool installed
+
+.Procedure
+
+. Use the `oc edit` command to modify the proxy:
```
jboxman: s/proxy/proxy object/

modules/installation-proxy-configure-object.adoc
```diff
@@ -0,0 +1,121 @@
...
+----
+<1> One or more PEM-encoded X.509 certificates used to sign the proxy's
+identity certificate.
+<2> The ConfigMap name that will be referenced from the proxy object.
+<3> The ConfigMap must be in the `openshift-config` namespace.
+
+.. Create the ConfigMap from this file:
++
+----
+$ oc create -f user-ca-bundle.yaml
```
jboxman: The user never explicitly creates `user-ca-bundle.yaml`.

modules/installation-proxy-configure-object.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+* Cluster administrator permissions
+* {product-title} `oc` CLI tool installed
+
+.Procedure
+
+. Use the `oc edit` command to modify the proxy:
++
+----
+$ oc edit proxy/cluster
```
jboxman: I think the `oc edit` step can come after the ConfigMap step; If the user doesn't need additional CA certificates, that step can still be skipped. Otherwise, the user needs to open another terminal, because for the ConfigMap the original terminal now has an editor open.

modules/installation-proxy-configure-object.adoc
```diff
@@ -0,0 +1,121 @@
...
+  - http://www.google.com <4>
+  - https://www.google.com
+  trustedCA:
+    name: user-ca-bundle <5>
+status:
+  httpProxy: http://<username>:<pswd>@<ip>:<port>
+  httpsProxy: https://<username>:<pswd>@<ip>:<port>
+  noProxy: 10.0.0.0/16,10.128.0.0/14,127.0.0.1,169.254.169.254,172.30.0.0/16,api-int.cluster.example.com,api.cluster.example.com,etcd-0.cluster.example.com,etcd-1.cluster.example.com,etcd-2.cluster.example.com,example.com,localhost
+----
+<1> A proxy URL to use for creating HTTP connections outside the cluster.
```
jboxman: I'd probably

_A proxy URL to use when making HTTP connections with hosts outside the cluster._

jboxman:
@bergerhoffer, I took a look. It's late enough I'm hesitant to label _Peer review completed_ though.

_topic_map.yml
danehans: +1 for _Configuring the cluster-wide proxy_

installing/install_config/enable-cluster-wide-proxy.adoc
danehans: @bergerhoffer this is not completely accurate:
```
Configuring {product-title} to use these proxies can be as simple as setting standard environment variables in configuration or JSON files.
```
I would change to something like:
```
Configuring {product-title} to use these proxies can be as simple as modifying the `Proxy` object for existing clusters or configuring proxy settings in `install-config.yaml` for new clusters.
```

modules/installation-proxy-configure-object.adoc
```diff
@@ -0,0 +1,121 @@
...
+authority from the RHCOS trust bundle.
+
+.. Create a file with the following contents and the values of your PEM-encoded
+certificates:
++
+[source,yaml]
+----
+apiVersion: v1
+data:
+  ca-bundle.crt: |
```
danehans: @bergerhoffer you may want to mention that the data key must be named `ca-bundle.crt`.

modules/installation-proxy-configure-object.adoc
danehans: @bergerhoffer I would remove status and all the status fields from this example.

modules/installation-proxy-configure-object.adoc
```diff
@@ -0,0 +1,121 @@
...
+----
+
+. Configure the necessary fields for the proxy:
++
+[source,yaml]
+----
+apiVersion: config.openshift.io/v1
+kind: Proxy
+metadata:
+  creationTimestamp: "2019-08-21T22:36:49Z"
```
danehans: I would remove all the metadata fields except `name:` from this example.

modules/installation-proxy-configure-object.adoc
```diff
@@ -0,0 +1,121 @@
...
+<3> The ConfigMap must be in the `openshift-config` namespace.
+
+.. Create the ConfigMap from this file:
++
+----
+$ oc create -f user-ca-bundle.yaml
+----
+
+. Configure the necessary fields for the proxy:
++
```
danehans: @bergerhoffer I suggest adding the command used to modify the default proxy:
```
$ oc edit proxy/cluster
```

modules/installation-proxy-remove.adoc
```diff
@@ -0,0 +1,43 @@
+// Module included in the following assemblies:
+//
+// * installing/insall_config/enable-cluster-wide-proxy.adoc
+
+[id="installation-proxy-remove_{context}"]
+= Removing proxy from the cluster
+
+The `cluster` proxy object cannot be deleted. To remove proxy from a cluster,
+remove all `spec` fields from the proxy object.
+
```
danehans: Since `spec` can not be removed, I suggest providing an example like so:
```
$ oc get proxy/cluster -o yaml
kind: Proxy
metadata:
  name: cluster
spec: {}
status: {}
```

modules/installation-proxy-remove.adoc
danehans: Example should be:
```
$ oc get proxy/cluster -o yaml
kind: Proxy
metadata:
  name: cluster
spec: {}
status: {}
```

@jboxman I have an open question with the installer team regarding why this field is being set when spec should be `spec: {}`. It's harmless that the field is set to an empty string, but may cause confusion and docs inaccuracies.

modules/installation-proxy-remove.adoc
danehans: @bergerhoffer as @jboxman mentions, drop all the metadata fields except `name: cluster`.

danehans:
@bergerhoffer after reviewing the rendered docs, my biggest concern is the documentation lives under the install section and all the docs in this PR are related to config'ing a proxy post install. The install config (`install-config.yaml`) is used for installing a cluster with proxy. I cover the install details in slides 12-14 in the presentation I provided you.

modules/installation-proxy-configure-object.adoc
bergerhoffer: Oops yep, that should be below, nice catch!

bergerhoffer:
@danehans Thank you for the review! Yes I agree. This is where these docs were initially put, so I hadn't moved them yet, but I do think that this location is probably more appropriate for installing a cluster with a proxy (unless there are different steps per provider).

Do you have any suggestions on where it might make most sense to put this. Under networking?

modules/installation-proxy-configure-object.adoc
bergerhoffer: I don't think so, I've just been consistent when I used them. And honestly I hate the "commonly known as `oc`" one. We did agree awhile back that we can call it the "OpenShift CLI (`oc`)", and I prefer that because it is descriptive and concise. I don't have a strong opinion on the other. This one actually came from Brandi since this PR builds on hers, but I had usually been saying something like "Access to the cluster as a user with the `cluster-admin` role."

modules/installation-proxy-configure-object.adoc
bergerhoffer: I agree, removing the "status" mention.

bergerhoffer:
This has been moved under Networking, new preview link is https://osdocs-468--ocpdocs.netlify.com/openshift-enterprise/latest/networking/enable-cluster-wide-proxy.html

modules/installation-proxy-remove.adoc
jboxman: That makes a lot more sense! Thanks!

networking/enable-cluster-wide-proxy.adoc
```diff
@@ -0,0 +1,17 @@
...
+include::modules/common-attributes.adoc[]
+:context: config-cluster-wide-proxy
+
+toc::[]
+
+Production environments can deny direct access to the Internet and instead have
+an HTTP or HTTPS proxy available. You can configure {product-title} to use a
+proxy by
+xref:../networking/enable-cluster-wide-proxy.adoc#nw-proxy-configure-object_config-cluster-wide-proxy[modifying the Proxy object for existing clusters]
+or by configuring the proxy settings in the `install-config.yaml` file for new
```
gpei: hi @bergerhoffer do we have related doc PR for this?  Guiding customer how to configure proxy settings in the `install-config.yaml` is also very important.

PR #16594, opened 2019-09-11
OSDOCS-485: Updated CSI instructions for OCP 4.2

huffmanca:
Updating the CSI details for the OCP 4.2 release:

- CSI is moving from technical preview to GA, so the technical preview note has been dropped.
- We do not ship any drivers or sidecars. I've removed the section on deploying the CSI drivers, and included a mention in the IMPORTANT block at the beginning to reference the CSI driver details.

This is for OCP 4.2.


vikram-redhat:
@liangxia please review.

liangxia:
The change looks good to me.

As I'm just back from a long leave, I will let the QE owner @chao007 to have a look, too.

vikram-redhat:
@chao007 is this PR ok from QE's side?

chao007:
@vikram-redhat  I submit one issue https://bugzilla.redhat.com/show_bug.cgi?id=1756836 here, but I can still see it in this pr.
Will it be removed later?
Others are OK.

liangxia:
> @vikram-redhat I submit one issue https://bugzilla.redhat.com/show_bug.cgi?id=1756836 here, but I can still see it in this pr.
> Will it be removed later?

The issue is fixed in https://github.com/openshift/openshift-docs/pull/16977
And I can see from web page,
OpenShift Container Platform does not ship with any CSI drivers.


> Others are OK.

So all are OK now.

PR #16596, opened 2019-09-11
OSDOCS-687: Included FibreChannel details.

huffmanca:
Included FibreChannel details. This is for OCP 4.1 and 4.2, and is a port of the 3.x documentation.

### DIFF
--- a/modules/persistent-storage-fibre-provisioning.adoc
+++ b/modules/persistent-storage-fibre-provisioning.adoc
@@ -1,0 +1,48 @@
+// Module included in the following assemblies:
+//
+// * storage/persistent-storage/persistent-storage-fibre.adoc
+
+[id="provisioning-fibre_{context}"]
+= Provisioning
+To provision Fibre Channel volumes using the PersistentVolume API
+the following must be available:
+
+* The `targetWWNs` (array of Fibre Channel target's World Wide
+Names).
+* A valid LUN number.
+* The filesystem type.
+
+A PersistentVolume and a LUN have a one-to-one mapping between them.
+
+.Prerequisites
+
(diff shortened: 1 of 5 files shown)


modules/persistent-storage-fibre-provisioning.adoc
```diff
@@ -0,0 +1,43 @@
+// Module included in the following assemblies:
+//
+// * storage/persistent-storage/persistent-storage-fibre.adoc
+
+[id="provisioning-fibre_{context}"]
+= Provisioning
+All that is required for Fibre Channel persistent
```
ahardin-rh: Maybe these can be an unordered list?

storage/persistent-storage/persistent-storage-fibre.adoc
```diff
@@ -0,0 +1,35 @@
...
+resources without having any knowledge of the underlying infrastructure.
+PersistentVolumes are not bound to a single project or namespace; they can be
+shared across the {product-title} cluster.
+PersistentVolumeClaims are specific to a project or namespace and can be
+requested by users.
+
+
+[IMPORTANT]
+====
+High-availability of storage in the infrastructure is left to the underlying
```
ahardin-rh: I think you can drop the hyphenization; we can cross-check other instances of this

ahardin-rh (commented):
Just a few comments from me. Otherwise, looks great!🎉

storage/persistent-storage/persistent-storage-fibre.adoc
huffmanca: Good call! It seems that `high availability` is vastly more common. I've made the change here.

liangxia:
LGTM (from QE)

PR #16607, opened 2019-09-12
updating bare metal network parameter

kalexand-rh:
Per https://github.com/openshift/installer/pull/2227/files, `clusterNetworks` should be `clusterNetwork` in the `install-config.yaml` files. This is the only instance that has not been updated.

@morenod, will you PTAL? I think this change is valid for 4.1 and 4.2.


kalexand-rh:
@morenod, will you PTAL and confirm if this change is valid for 4.1 and 4.2?

morenod:
lgtm

PR #16631, opened 2019-09-13
Removing KubeFed for 4.2

abrennan89:
KubeFed had been planned for tech preview in 4.2, but this is no longer happening.
Removing all modules and entries related to KubeFed for now.


sheriff-rh (approved):
Works for me!

PR #16635, opened 2019-09-13
OSDOCS-640: Adding docs for configuring proxy during installation

bergerhoffer:
Preview: https://osdocs-640--ocpdocs.netlify.com/openshift-enterprise/latest/installing/installing_aws/installing-aws-customizations.html#installation-configure-proxy_install-customizations-cloud


modules/installation-configure-proxy.adoc
```diff
@@ -0,0 +1,67 @@
...
+
+. Edit your `install-config.yaml` file and add the proxy settings. For example:
++
+[source,yaml]
+----
+apiVersion: v1
+baseDomain: my.domain.com
+proxy:
+  httpProxy: http://<username>:<pswd>@<ip>:<port> <1>
+  httpsProxy: https://<username>:<pswd>@<ip>:<port> <2>
```
danehans: You may want to mention that the url scheme can be either `http` or https`.

modules/installation-configure-proxy.adoc
```diff
@@ -0,0 +1,67 @@
...
+.Procedure
+
+. Edit your `install-config.yaml` file and add the proxy settings. For example:
++
+[source,yaml]
+----
+apiVersion: v1
+baseDomain: my.domain.com
+proxy:
+  httpProxy: http://<username>:<pswd>@<ip>:<port> <1>
```
danehans: You may want to mention that the url scheme must be `http`.

modules/installation-configure-proxy.adoc
```diff
@@ -0,0 +1,67 @@
+// Module included in the following assemblies:
+//
+// * installing/installing_aws/installing-aws-customizations.adoc
+// * installing/installing_aws_user_infra/installing-aws-user-infra.adoc
+// * installing/installing_bare_metal/installing-bare-metal.adoc
+// * installing/installing_vsphere/installing-vsphere.adoc
+// TODO: gcp
+
+// Questions:
+// - Can they provide multiple certs in additionalTrustBundle?
```
danehans: Yes and would be typical. Most proxy's will be issued an id cert that a) was not signed by an authority from the RHCOS trust bundle and b) that is not issued by a CA, but instead an intermediate CA.

modules/installation-configure-proxy.adoc
```diff
@@ -0,0 +1,67 @@
...
+// - Is it true that the trustedCA field is filled in? Slides say "trusted-ca-bundle",  but ConfigMap created is user-ca-bundle
```
danehans: Yes, a proxy is created with `trustedCA` referencing configmap `trusted-ca-bundle` in namespace `openshift-config-managed` when `additionalTrustBundle` is specified in install-config. The installer will create the proxy `cluster` and configmap `user-ca-bundle` resources. `trusted-ca-bundle` configmap is generated by ClusterNetwork Operator at run time by validating configmap `user-ca-bundle`, merging the contents with the RHCOS trust bundle and producing configmap `trusted-ca-bundle`. Take a look at the slide for additional detail.

bergerhoffer:
Thanks for the feedback @danehans! I've adjusted the field descriptions if you can take another look. Thanks!

installing/installing_bare_metal/installing-bare-metal.adoc
```diff
@@ -49,6 +49,8 @@ include::modules/installation-initializing-manual.adoc[leveloffset=+1]
 
 include::modules/installation-bare-metal-config-yaml.adoc[leveloffset=+2]
 
+include::modules/installation-configure-proxy.adoc[leveloffset=+2]
```
danehans: Take a look at slide 20 from the presentation that I shared with you regarding bare-metal installs. The cluster network operator uses `machineCIDR` from install-config for noProxy status. Node IP's must be from machineCIDR for bare-metal installs. Otherwise, the node IP's/CIDR must be added manually to `noProxy` of the proxy object or install-config. For example:

```
$ oc get cm/cluster-config-v1 -n kube-system -o yaml | grep machineCIDR
      machineCIDR: 10.0.0.0/16
$ oc get proxy/cluster -o yaml | grep 10.0.0.0/16
  noProxy: .cluster.local,.svc,.us-west-2.compute.internal,10.0.0.0/16,10.128.0.0/14,127.0.0.1,169.254.169.254,172.30.0.0/16,api-int.eric-proxy.devcluster.openshift.com,api.eric-proxy.devcluster.openshift.com,apps.eric-proxy.devcluster.openshift.com,etcd-0.eric-proxy.devcluster.openshift.com,etcd-1.eric-proxy.devcluster.openshift.com,etcd-2.eric-proxy.devcluster.openshift.com,localhost
```

modules/installation-configure-proxy.adoc
```diff
@@ -0,0 +1,76 @@
...
+<1> A proxy URL to use for creating HTTP connections outside the cluster. The
+URL scheme must be `http`.
+<2> A proxy URL to use for creating HTTPS connections outside the cluster. If
+this is not specified, then `httpProxy` is used for both HTTP and HTTPS
+connections. The URL scheme can either be `http` or `https`.
+<3> A comma-separated list of destination domain names, domains, IP addresses or
+other network CIDRs to exclude proxying. Preface a domain with `.` to include
+all subdomains of that domain. Use `*` to bypass proxy for all destinations.
+<4> If provided, the installer generates a ConfigMap named `user-ca-bundle` in
+the `openshift-config` namespace that contains the one or more additional CA
```
danehans: s/the one/one/

modules/installation-configure-proxy.adoc
```diff
@@ -0,0 +1,76 @@
...
+  httpsProxy: https://<username>:<pswd>@<ip>:<port> <2>
+  noProxy: example.com <3>
+additionalTrustBundle: | <4>
+    -----BEGIN CERTIFICATE-----
+      <MY_TRUSTED_CA_CERT>
+    -----END CERTIFICATE-----
+...
+----
+<1> A proxy URL to use for creating HTTP connections outside the cluster. The
+URL scheme must be `http`.
```
danehans: Note that `httpProxy` is also used for https traffic if no `httpsProxy` is provided.

modules/installation-configure-proxy.adoc
```diff
@@ -0,0 +1,76 @@
...
+certificates required for proxying HTTPS connections. The Cluster Network
+Operator then creates a `trusted-ca-bundle` ConfigMap merging these contents
+with the RHCOS trust bundle, and this ConfigMap is referenced in the Proxy
+object's `trustedCA` field. The `additionalTrustBundle` field is required unless
+the proxy's identity certificate is signed by an authority from the RHCOS trust
+bundle.
++
+[NOTE]
+====
+The installer does not support the `readinessEndpoints` proxy field.
```
danehans: s/`readinessEndpoints` proxy field/proxy `readinessEndpoints` field/

modules/installation-configure-proxy.adoc
wking: I see @danehans asking for this restriction [above][1], but I'm curious about why.  My understanding is that you can use a HTTPS URI for `httpProxy` (e.g. to avoid leaking your proxy-connection creds), although you might bump into containers/image#699 if the proxy doesn't support any of the encryption options currently compiled into containers/image (and that would go for `httpsProxy` too).

[1]: https://github.com/openshift/openshift-docs/pull/16635/files#r324255631

modules/installation-configure-proxy.adoc
```diff
@@ -0,0 +1,76 @@
...
+<3> A comma-separated list of destination domain names, domains, IP addresses or
+other network CIDRs to exclude proxying. Preface a domain with `.` to include
+all subdomains of that domain. Use `*` to bypass proxy for all destinations.
+<4> If provided, the installer generates a ConfigMap named `user-ca-bundle` in
+the `openshift-config` namespace that contains the one or more additional CA
+certificates required for proxying HTTPS connections. The Cluster Network
+Operator then creates a `trusted-ca-bundle` ConfigMap merging these contents
+with the RHCOS trust bundle, and this ConfigMap is referenced in the Proxy
+object's `trustedCA` field. The `additionalTrustBundle` field is required unless
+the proxy's identity certificate is signed by an authority from the RHCOS trust
```
wking: You also only need this (for proxy connections) if you're using an `https` proxy URI.  No proxy cert to include if all your proxy connections are `http`.

bergerhoffer:
Preview bot isn't working, so posted a build to the file share site. The "Configuring the cluster-wide proxy during installation" file has been included in all appropriate provider/install types. Links to each install section are below:

- [AWS](http://file.rdu.redhat.com/~ahoffer/2019/OSDOCS-640/installing/installing_aws/installing-aws-customizations.html)
- [AWS UPI](http://file.rdu.redhat.com/~ahoffer/2019/OSDOCS-640/installing/installing_aws_user_infra/installing-aws-user-infra.html) *
- [Azure](http://file.rdu.redhat.com/~ahoffer/2019/OSDOCS-640/installing/installing_azure/installing-azure-customizations.html)
- [GCP](http://file.rdu.redhat.com/~ahoffer/2019/OSDOCS-640/installing/installing_gcp/installing-gcp-customizations.html)
- [Bare metal](http://file.rdu.redhat.com/~ahoffer/2019/OSDOCS-640/installing/installing_bare_metal/installing-bare-metal.html)
- [vSphere](http://file.rdu.redhat.com/~ahoffer/2019/OSDOCS-640/installing/installing_vsphere/installing-vsphere.html)
- [Restricted network AWS](http://file.rdu.redhat.com/~ahoffer/2019/OSDOCS-640/installing/installing_restricted_networks/installing-restricted-networks-aws.html) *
- [Restricted network bare metal](http://file.rdu.redhat.com/~ahoffer/2019/OSDOCS-640/installing/installing_restricted_networks/installing-restricted-networks-bare-metal.html)
- [Restricted network vSphere](http://file.rdu.redhat.com/~ahoffer/2019/OSDOCS-640/installing/installing_restricted_networks/installing-restricted-networks-vsphere.html)

\* These two sections restructured all content under the "Creating the installation files for AWS" heading (broke up a single procedure into two procedures and inserted configuring proxy procedure in between). Needs to be reviewed carefully.

installing/installing_restricted_networks/installing-disconnected.adoc
```diff
@@ -32,6 +32,9 @@ include::modules/installation-configuration-parameters.adoc[leveloffset=+2]
 
 include::modules/installation-azure-config-yaml.adoc[leveloffset=+2]
 
+// TODO: If this assembly is going to be for restricted network Azure install, whenever it is included, need to confirm whether proxy is supported before including the following file:
```
kalexand-rh: Restricted network Azure isn't in scope for 4.2, and I think I need to remove this file.

installing/installing_restricted_networks/installing-restricted-networks-aws.adoc
```diff
@@ -67,6 +67,12 @@ include::modules/ssh-agent-using.adoc[leveloffset=+1]
 
 include::modules/installation-generate-aws-user-infra.adoc[leveloffset=+1]
 
+include::modules/installation-generate-aws-user-infra-install-config.adoc[leveloffset=+2]
+
+include::modules/installation-configure-proxy.adoc[leveloffset=+2]
+
+include::modules/installation-generate-aws-user-infra-ignition.adoc[leveloffset=+2]
+
 // After the proxy change merges, I need to put it in and emphasize that you
```
kalexand-rh: You can remove this note, too.

modules/installation-configure-proxy.adoc
```diff
@@ -0,0 +1,83 @@
+// Module included in the following assemblies:
+//
+// * installing/installing_aws/installing-aws-customizations.adoc
+// * installing/installing_aws_user_infra/installing-aws-user-infra.adoc
+// * installing/installing_bare_metal/installing-bare-metal.adoc
+// * installing/installing_vsphere/installing-vsphere.adoc
+// TODO: gcp
```
kalexand-rh: Whenever you have time, please update the list of assemblies that includes this module.

modules/installation-configure-proxy.adoc
```diff
@@ -0,0 +1,83 @@
...
+
+Production environments can deny direct access to the Internet and instead have
+an HTTP or HTTPS proxy available. You can configure a new {product-title}
+cluster to use a proxy by configuring the proxy settings in the
+`install-config.yaml` file.
+
+ifdef::bare-metal[]
+[NOTE]
+====
+For bare metal installations, node IP addresses should be assigned from the `install-config.yaml` file's `networking.machineCIDR` field. If they are not, then the node IP addresses must be manually added to the `install-config.yaml` file's `proxy.noProxy` field.
```
kalexand-rh: s/should/must

modules/installation-configure-proxy.adoc
```diff
@@ -0,0 +1,83 @@
...
+----
+<1> A proxy URL to use for creating HTTP connections outside the cluster. The
+URL scheme must be `http`.
+<2> A proxy URL to use for creating HTTPS connections outside the cluster. If
+this is not specified, then `httpProxy` is used for both HTTP and HTTPS
+connections. The URL scheme can either be `http` or `https`.
+<3> A comma-separated list of destination domain names, domains, IP addresses or
+other network CIDRs to exclude proxying. Preface a domain with `.` to include
+all subdomains of that domain. Use `*` to bypass proxy for all destinations.
+<4> If provided, the installer generates a ConfigMap named `user-ca-bundle` in
```
kalexand-rh: @wking, how does the additionalTrustBundle work with proxy and disconnected? Can you provide multiple certs, or do you need to make sure that the proxy and your mirror repo use the same certs? Is there a specific format for multiple certs?

modules/installation-configure-proxy.adoc
```diff
@@ -0,0 +1,83 @@
...
+additionalTrustBundle: | <4>
+    -----BEGIN CERTIFICATE-----
+      <MY_TRUSTED_CA_CERT>
+    -----END CERTIFICATE-----
+...
+----
+<1> A proxy URL to use for creating HTTP connections outside the cluster. The
+URL scheme must be `http`.
+<2> A proxy URL to use for creating HTTPS connections outside the cluster. If
+this is not specified, then `httpProxy` is used for both HTTP and HTTPS
```
kalexand-rh: s/this/this parameter value

modules/installation-configure-proxy.adoc
```diff
@@ -0,0 +1,83 @@
...
+connections. The URL scheme can either be `http` or `https`.
+<3> A comma-separated list of destination domain names, domains, IP addresses or
```
kalexand-rh: s/addresses/addresses,

modules/installation-configure-proxy.adoc
kalexand-rh: s/installer/installation program
s/named/that is named

modules/installation-configure-proxy.adoc
```diff
@@ -0,0 +1,83 @@
...
+other network CIDRs to exclude proxying. Preface a domain with `.` to include
+all subdomains of that domain. Use `*` to bypass proxy for all destinations.
+<4> If provided, the installer generates a ConfigMap named `user-ca-bundle` in
+the `openshift-config` namespace that contains one or more additional CA
+certificates required for proxying HTTPS connections. The Cluster Network
```
kalexand-rh: s/required/that are required

modules/installation-configure-proxy.adoc
```diff
@@ -0,0 +1,83 @@
...
+Operator then creates a `trusted-ca-bundle` ConfigMap merging these contents
```
kalexand-rh: s/merging/that merges

modules/installation-configure-proxy.adoc
```diff
@@ -0,0 +1,83 @@
...
+with the RHCOS trust bundle, and this ConfigMap is referenced in the Proxy
```
kalexand-rh: s/RHCOS/{op-system-first}

modules/installation-configure-proxy.adoc
```diff
@@ -0,0 +1,83 @@
...
+object's `trustedCA` field. The `additionalTrustBundle` field is required unless
+the proxy's identity certificate is signed by an authority from the RHCOS trust
```
kalexand-rh: s/RHCOS/{op-system}

modules/installation-configure-proxy.adoc
```diff
@@ -0,0 +1,83 @@
...
+bundle.
++
+[NOTE]
+====
+The installer does not support the proxy `readinessEndpoints` field.
```
kalexand-rh: s/installer/installation program

modules/installation-configure-proxy.adoc
```diff
@@ -0,0 +1,83 @@
...
+====
+
+. Save the file and reference it when installing {product-title}.
+
+The installer creates a cluster-wide proxy named `cluster` with the proxy
```
kalexand-rh: s/installer/installation program
s/named/that is named
s/with/that uses

kalexand-rh (commented):
I have a couple of picks, but I think that this looks good. I have one new question that I think @wking can answer.

modules/installation-configure-proxy.adoc
bergerhoffer: @kalexand-rh So they technically don't have to, so I didn't think that "must" was appropriate here. Because if they aren't, then that's when they have to fill out the noProxy field. Do you have a different suggestion other than "should" that you'd prefer, that isn't must?

installing/installing_restricted_networks/installing-disconnected.adoc
bergerhoffer: @kalexand-rh Yeah, I heard it wasn't yet, but I wasn't sure if you were just prepping with this file for whenever it will be, that's why I added the note.

installing/installing_restricted_networks/installing-restricted-networks-aws.adoc
bergerhoffer: @kalexand-rh I can remove it, but just wanted to make sure we didn't need to add anything for the "emphasize that you must configure a proxy..." as the note says?

modules/installation-configure-proxy.adoc
bergerhoffer: Oooo I wasn't aware of this attribute, thanks!

modules/installation-configure-proxy.adoc
kalexand-rh: @bergerhoffer, maybe:
```
For bare metal installations, if you do not assign node IP addresses from the range that is specified in the `networking.machineCIDR` value in the `install-config.yaml` file, you must include them in the `proxy.noProxy` parameter value in the `install-config.yaml` file. 
```

modules/installation-configure-proxy.adoc
```diff
@@ -0,0 +1,87 @@
...
+
+. Edit your `install-config.yaml` file and add the proxy settings. For example:
++
+[source,yaml]
+----
+apiVersion: v1
+baseDomain: my.domain.com
+proxy:
+  httpProxy: http://<username>:<pswd>@<ip>:<port> <1>
+  httpsProxy: https://<username>:<pswd>@<ip>:<port> <2>
```
danehans: Due to https://bugzilla.redhat.com/show_bug.cgi?id=1750650, a `https` url scheme is not supported in 4.2. Therefore, the example should be:
`httpsProxy: http://<username>:<pswd>@<ip>:<port> <2>`. It's my understanding that a release note is or has been created for this caveat. You may want to reference the release note in this doc or make sure user's are aware of the limitation.

modules/installation-configure-proxy.adoc
```diff
@@ -0,0 +1,87 @@
...
+    -----BEGIN CERTIFICATE-----
+      <MY_TRUSTED_CA_CERT>
+    -----END CERTIFICATE-----
+...
+----
+<1> A proxy URL to use for creating HTTP connections outside the cluster. The
+URL scheme must be `http`.
+<2> A proxy URL to use for creating HTTPS connections outside the cluster. If
+this field is not specified, then `httpProxy` is used for both HTTP and HTTPS
+connections. The URL scheme can either be `http` or `https`.
```
danehans: See https://github.com/openshift/openshift-docs/pull/16635/files#r327846123 regarding:
```
The URL scheme can either be `http` or `https`
```

modules/installation-generate-aws-user-infra-ignition.adoc
```diff
@@ -0,0 +1,81 @@
...
+certificates that expire after 24 hours. You must complete your cluster
+installation and keep the cluster running for 24 hours in a non-degraded state
+to ensure that the first certificate rotation has finished.
+====
+
+.Prerequisites
+
+* Obtain the {product-title} installation program.
+ifdef::restricted[]
+For a restricted network installation, these files are on your bastion host.
```
danehans: If restricted network is not going to be supported in 4.2, you should consider removing this.

modules/installation-generate-aws-user-infra-ignition.adoc
```diff
@@ -0,0 +1,81 @@
...
+endif::restricted[]
+* Create the `install-config.yaml` installation configuration file.
+
+.Procedure
+
+. Remove the Kubernetes manifest files for the control plane machines. By
+removing these files, you prevent the cluster from automatically generating
+control plane machines.
+.. Generate the Kubernetes manifests for the cluster:
```
danehans: I'm a bit confused by this paragraph. First it talks about removing manifests, then it states to generate the manifests.

modules/installation-generate-aws-user-infra-install-config.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+Generate and customize the installation configuration file that the
+installation program needs to deploy your cluster.
+
+.Prerequisites
+
+* Obtain the {product-title} installation program and the pull secret for your
+cluster.
+ifdef::restricted[]
+For a restricted network installation, these files are on your bastion host.
```
danehans: Same as https://github.com/openshift/openshift-docs/pull/16635/files#r327847634

modules/installation-generate-aws-user-infra-install-config.adoc
```diff
@@ -0,0 +1,121 @@
...
+... If you do not have an AWS profile stored on your computer, enter the AWS
+access key ID and secret access key for the user that you configured to run the
+installation program.
+... Select the AWS region to deploy the cluster to.
+... Select the base domain for the Route53 service that you configured for your cluster.
+... Enter a descriptive name for your cluster.
+... Paste the pull secret that you obtained from the
+link:https://cloud.redhat.com/openshift/install[OpenShift Infrastructure Providers] page.
+
+. Edit the `install-config.yaml` file to set the number of compute, or worker,
```
danehans: @wking correct me if I'm wrong, but this section should be removed due to https://bugzilla.redhat.com/show_bug.cgi?id=1755073.

modules/installation-generate-aws-user-infra-install-config.adoc
```diff
@@ -0,0 +1,121 @@
...
+[source,yaml]
+----
+compute:
+- hyperthreading: Enabled
+  name: worker
+  platform: {}
+  replicas: 0
+----
+
+ifdef::restricted[]
```
danehans: Same as https://github.com/openshift/openshift-docs/pull/16635/files#r327847634

danehans:
@bergerhoffer this is coming along nicely. I provided a few comments. This PR covers proxy at install time. Will a separate PR address configuring `Proxy` for day 2?

danehans:
@bergerhoffer UPI-based proxy installs require a newer ami from [4.1](https://docs.openshift.com/container-platform/4.1/installing/installing_aws_user_infra/installing-aws-user-infra.html#installation-aws-user-infra-rhcos-ami_installing-aws-user-infra). Are the AMI references being updated as part of the general 4.2 docs? cc: @wking @ewolinetz

modules/installation-configure-proxy.adoc
wking: Multi-cert format in https://github.com/openshift/installer/blob/master/docs/user/customization.md#additional-trust-bundle (basically just concat certs)

wking:
> Are the AMI references being updated...

If we haven't yet, 4.2 should definitely bump to https://github.com/openshift/installer/blob/release-4.2/data/data/rhcos.json

modules/installation-configure-proxy.adoc
bergerhoffer: Looks like the WIP release notes PR has mentioned this as a known issue, if you want to make sure this is sufficient: http://file.rdu.redhat.com/~ahardin/09232019/OCP-4-2-release-notes/release_notes/ocp-4-2-release-notes.html#ocp-4-2-known-issues

I'll update the example here.

modules/installation-configure-proxy.adoc
bergerhoffer: Let me know if this wording works:

```
The URL scheme must be `http`; `https` is currently not supported.
```

modules/installation-generate-aws-user-infra-ignition.adoc
bergerhoffer: @kalexand-rh can confirm for me, but the docs for these restricted network are in for 4.2, so they should be supported. Also, this sentence will only appear in the restricted docs, so if they don't go out for some reason, then this won't appear.

modules/installation-configure-proxy.adoc
bergerhoffer: I'll also update the example and this wording where we use these same descriptions when configuring the proxy post-install.

modules/installation-generate-aws-user-infra-ignition.adoc
wking: > First it talks about removing manifests, then it states to generate the manifests.

You: "Installer, make your usual manifests." Installer: "Here you go" You: "Forget about these.  The rest look good.  Carry on."

However we want to explain that workflow works for me.

modules/installation-generate-aws-user-infra-install-config.adoc
wking: I think so, but prob wait until https://github.com/openshift/installer/pull/2402 lands or is otherwise sorted.

modules/installation-generate-aws-user-infra-ignition.adoc
bergerhoffer: This was pulled from the original module that I broke up, but I agree that it is confusing.

@kalexand-rh - I don't have the background to know why the steps are the way that they are. Do you have any reword suggestions that would make this clearer, or know who might be good to ask on this?

bergerhoffer: Sorry I missed your comment @wking before I added mine - thanks for the explanation! 

@kalexand-rh let's look at this together tomorrow.

bergerhoffer:
@vikram-redhat Do you know if it's on anyone's plate to update the list of AMIs for 4.2 [1] to [2] as suggested? 

[1] https://docs.openshift.com/container-platform/4.2/installing/installing_aws_user_infra/installing-aws-user-infra.html#installation-aws-user-infra-rhcos-ami_installing-aws-user-infra
[2] https://github.com/openshift/installer/blob/release-4.2/data/data/rhcos.json

bergerhoffer:
@danehans I've made a few updates. I'm going to go ahead and merge this PR because we have a round of reviews over the next few days that I'd like this to be included for. But we can continue to refine these docs as necessary in a separate PR.

When you ask about day 2 configuration of the proxy, are you talking about enabling the proxy on an existing cluster, which was added by PR #16562 (and can be viewed [here](https://docs.openshift.com/container-platform/4.2/networking/enable-cluster-wide-proxy.html)). Or are there some other configuration docs that we're missing?

modules/nw-proxy-configure-object.adoc
```diff
@@ -81,18 +81,20 @@ metadata:
   name: cluster
 spec:
   httpProxy: http://<username>:<pswd>@<ip>:<port> <1>
-  httpsProxy: https://<username>:<pswd>@<ip>:<port> <2>
+  httpsProxy: http://<username>:<pswd>@<ip>:<port> <2>
```
bergerhoffer: @danehans Please verify that the changes to this file are appropriate in response to your comments on the install docs that use these similar examples and descriptions. It can be previewed [here](https://osdocs-640--ocpdocs.netlify.com/openshift-enterprise/latest/networking/enable-cluster-wide-proxy.html).

vikram-redhat:
> @vikram-redhat Do you know if it's on anyone's plate to update the list of AMIs for 4.2 [1] to [2] as suggested?
> 
> [1] https://docs.openshift.com/container-platform/4.2/installing/installing_aws_user_infra/installing-aws-user-infra.html#installation-aws-user-infra-rhcos-ami_installing-aws-user-infra
> [2] https://github.com/openshift/installer/blob/release-4.2/data/data/rhcos.json

@kalexand-rh would you want to add this to the last minute tasks?

installing/installing_aws/installing-aws-customizations.adoc
```diff
@@ -44,6 +44,8 @@ include::modules/installation-configuration-parameters.adoc[leveloffset=+2]
 
 include::modules/installation-aws-config-yaml.adoc[leveloffset=+2]
 
+include::modules/installation-configure-proxy.adoc[leveloffset=+2]
```
gpei: I have a doubt whether we should add it here. As I know, only UPI way support proxy setting until IPI supports using a preconfigured VPC. This section is for IPI-on-AWS.

installing/installing_azure/installing-azure-customizations.adoc
```diff
@@ -32,6 +32,8 @@ include::modules/installation-configuration-parameters.adoc[leveloffset=+2]
 
 include::modules/installation-azure-config-yaml.adoc[leveloffset=+2]
 
+include::modules/installation-configure-proxy.adoc[leveloffset=+2]
```
gpei: Same here, we only have IPI-on-Azure in 4.2

installing/installing_gcp/installing-gcp-customizations.adoc
```diff
@@ -32,6 +32,8 @@ include::modules/installation-configuration-parameters.adoc[leveloffset=+2]
 
 include::modules/installation-gcp-config-yaml.adoc[leveloffset=+2]
 
+include::modules/installation-configure-proxy.adoc[leveloffset=+2]
+
```
gpei: Same here, this section is for IPI-on-GCP

modules/installation-configure-proxy.adoc
gpei: I once had a try with using https URI for `httpProxy`, the network-operator will complain about it.

installing/installing_aws/installing-aws-customizations.adoc
wking: You _can_ use a proxy on AWS IPI, the proxy just has to live outside the VPC and still be reachable by the cluster.  This is how CI works now.  Not as nice as what's in flight with openshift/release#4719, but still something.

modules/installation-configure-proxy.adoc
wking: What error message did you see?

installing/installing_aws/installing-aws-customizations.adoc
bergerhoffer: @gpei Are you suggesting that this should not be included here for AWS IPI (and Azure and GCP)? I had confirmed with @katherinedube  the appropriate supported list here: https://jira.coreos.com/browse/OSDOCS-640?focusedCommentId=130922&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-130922

katherinedube: As Trevor mentioned, until we support deploying to a pre-existing VPC (planned for the next release), the proxy must live outside the VPC and be reachable by the nodes. While I agree it's not ideal, I'm not sure we should be saying it doesn't work unless it *really* doesn't work.

gpei: yeah, I understand that we _can_ use proxy for IPI install. But QE was told `UPI must be used for proxy testing` in 4.2, so all proxy related testing were scheduled with UPI installs in QE's 4.2 test matrix.

wking: All proxy CI to date has been IPI (with a pivot to UPI still in flight).

installing/installing_gcp/installing-gcp-customizations.adoc
gpei: And for proxy support on GCP, we still have a testblocker bug https://bugzilla.redhat.com/show_bug.cgi?id=1753930, which was targeted to 4.3.

wking: openshift/installer#2405 is in the merge queue, which would unblock a 4.2 backport.  But yeah, restricting proxy GCP to UPI in the meantime would be a good, conservative choice.

modules/installation-configure-proxy.adoc
gpei: It's a quite long time ago, so I just spin up a new cluster with such setting. 
In network-operator pod log, I could see such message, also a clear prompt.
```
2019/09/27 02:42:21 Updated ClusterOperator with conditions:
- lastTransitionTime: "2019-09-27T02:42:21Z"
  message: The configuration is invalid for proxy 'cluster' (httpProxy requires a
    'http' URI scheme). Use 'oc edit proxy.config.openshift.io cluster' to fix.
  reason: InvalidProxyConfig
```

wking: Followed up [here][1], where that check landed.

[1]: https://github.com/openshift/cluster-network-operator/commit/095cd157b3a2f8764f7ee090d09a3a92cf177ef8#r35259609

installing/installing_aws/installing-aws-customizations.adoc
katherinedube: I just checked through the OpenShift QE test matrix google doc and went through all the comments. I traced it back to a discussion Daneyon had with Xiaoli to remove the testing of ipi installs (at the time that it might have been the belief), but after ipi was tested, it wasn't the case. However, I doesn't look like anyone told QE of this. Everything including docs, CEE enablement, what we've told customers/field, etc. mentions that ipi will work. Can we get QE to do some testing on ipi installs?

installing/installing_gcp/installing-gcp-customizations.adoc
wking: The 4.3/master changes all landed.  Hopefully we'll have the 4.2 backports soon.

installing/installing_aws/installing-aws-customizations.adoc
danehans: As @wking mentions, IPI can _technically_ be used to install a proxy-enabled cluster. However, do we believe this is how any customer would use the feature? If so, I'm interested in better understanding the use case. We took this approach for phase 1 of proxy CI out of necessity and with the understanding that phase 2 (UPI-based) is the real way of setting up a proxy-enabled cluster. If IPI proxy is included in the docs, we must make sure the reader understands that components are proxied only because they honor proxy env var settings and can simply bypass proxying by unsetting these env vars.

danehans: > I traced it back to a discussion Daneyon had with Xiaoli to remove the testing of ipi installs (at the time that it might have been the belief), but after ipi was tested, it wasn't the case.

@katherinedube IPI was being used for testing as a stop-gap until UPI CI was complete. It was identified early on by @derekwaynecarr that UPI is required due to the lack of existing VPC support for IPI. As I mention in my above comment, IPI can technically work, but I don't see the requirements for IPI proxy meeting real world use cases.

installing/installing_gcp/installing-gcp-customizations.adoc
danehans: PRs for https://bugzilla.redhat.com/show_bug.cgi?id=1753930 have merged.

installing/installing_aws/installing-aws-customizations.adoc
katherinedube: Just to be clear, even if we're going to say upi support is required for proxy this should only effect public cloud provider ipi deployments. Meaning I don't know this is really a generic ipi problem across all of our ipi providers like OSP. As it is today in the text matrix, it's all ipi providers - so nothing is being tested with OSP (not sure this should have been excluded.) Otherwise, why couldn't a public cloud provider use proxy with ipi? Could you provider some additional details as to why it wouldn't be practical? Thanks!

jianlinliu: I totally agree with @danehans 's statement.
IPI's prerequisite is having internet connectivity. That means, cluster itself already have internet connection, why still need proxy? Even proxy is enabled, how to validate all components and operators are really getting outside via proxy but not via its own internet connectivity? 
In QE's testing, we dropped internet connectivity, and enable proxy, we found the following bugs:
Bug 1753467 - [proxy] no proxy is set for kube-controller-manager

That indicates if we do not drop internet connectivity for proxy testing, our test result is NOT persuadable, or else, we would miss the above bugs. But in the whole IPI install process, we have no way to drop internet connectivity.

wking: > In QE's testing, we dropped internet connectivity, and enable proxy, we found...

We definitely want to support proxy on blackholed UPI; I don't think anyone opposes that.  Once we have bring-your-own-VPC, we all agree we want to support proxy there too.  And testing in *those* environments makes me reasonably confident that at least OpenShift components will respect a voluntary proxy.  The only contention seems to be whether we want to support proxy for installer-provided-everything.  And the pushback seems to be mostly around "who wants this?".  But I don't see anything about that case that would be a support burden, so I'd rather not add confusion by walling it off as an unsupported niche.  Maybe customers want a proxy for MitM snooping on our egress (and they are also convinced from our blackhole support that we are playing nice).  Or maybe they want a proxy for some other reason.  But I much prefer the optics of "you may not care about this on IPI" to "we don't support this (suggesting it does not work reliably) on IPI".  From a QE angle, we don't test all possible config permutations on all platforms.  I'm fine leaving IPI+proxy covered just in CI, and relying on QE in the blackholed cases to turn up bugs like they're already doing.

jianlinliu: >  From a QE angle, we don't test all possible config permutations on all platforms.
Yeah, that is my concern. QE did not run proxy testing in a blackholed IPI (due to in 4.2 IPI, we have no way to create a blackholed env), but only proxy testing in a blackholed UPI.

I like "we don't support this (suggesting it does not work reliably) on IPI".

installing/installing_bare_metal/installing-bare-metal.adoc
bergerhoffer: @gpei @danehans I've asked a question in the BZ here if either of you have input as well: https://bugzilla.redhat.com/show_bug.cgi?id=1752832#c2

Basically just wanting to confirm if bare metal is still the only place this note needs to be during installation, and then for scaling workers after installation, does that include both manual scaling and autoscaling? And if so for autoscaling, how would that work?

installing/installing_gcp/installing-gcp-customizations.adoc
gpei: https://bugzilla.redhat.com/show_bug.cgi?id=1753930 was already verified with 4.3 build. 
But if the 4.2 backports were not merged into 4.2.0, customer still have the problem when 4.2.0 is shipped.

bergerhoffer: FYI - with PR #17080 we're limiting support to UPI installs only, so support is only for UPI for GCP now.

@danehans Were the PRs for https://bugzilla.redhat.com/show_bug.cgi?id=1753930 backported to 4.2, or is this still a problem in 4.2 for GCP UPI? 

If not, you mentioned there was a workaround of adding 'metadata.google.internal.' to noProxy - let me know if this is something that we need to document or not.

gpei: @bergerhoffer The cloned bug for 4.2 https://bugzilla.redhat.com/show_bug.cgi?id=1759245 has been verified now, it's not an issue for 4.2.0 GCP UPI anymore, thanks!

therevoman:
My client is using a transparent proxy..  no url to put into http(s)Proxy..  however any outbound traffic needs a CA in the chain...   additionalTrustedCA only sort of works for installation and post-install tasks.

bergerhoffer:
@danehans Can you address @therevoman's situation above?

PR #16642, opened 2019-09-13
bz 1699836 added note for custom CA deployment

sheriff-rh:
Bug:
https://bugzilla.redhat.com/show_bug.cgi?id=1699836

@jiajliu  PTAL.

### DIFF
--- a/install_config/redeploying_certificates.adoc
+++ b/install_config/redeploying_certificates.adoc
@@ -683,6 +683,13 @@ This chain can contain more than one certificate if you use intermediate CAs.
 ----
 $ cat router.crt /etc/origin/master/ca.crt router.key > router.pem
 ----
++
+[NOTE]
+====
+This step is only valid if you are using a certificate signed by the OpenShift
+CA. If a custom certificate is used, a file with the correct CA chain should be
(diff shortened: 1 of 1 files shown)


install_config/redeploying_certificates.adoc
```diff
@@ -683,6 +683,13 @@ This chain can contain more than one certificate if you use intermediate CAs.
 ----
 $ cat router.crt /etc/origin/master/ca.crt router.key > router.pem
 ----
++
+[NOTE]
+====
+This step is only valid if you are using a certificate signed by the OpenShift
+CA. If a custom certificate is used, a file with the correct CA chain should be
+used instead.
```
palonsoro: I think it would be better to explicitly say "a file with the correct CA chain should be used insteaed of /etc/origin/master/ca.crt" . 

With current writing, a novice user may think that the file to be replaced with "a file with the correct CA chain" is the whole resulting router.pem

sheriff-rh: I agree. I added the context.

sheriff-rh:
@gpei please take a look.

gpei:
LGTM

PR #16648, opened 2019-09-13
Fix example for ingress sharding by namespace

Miciah:
Follow-up to https://github.com/openshift/openshift-docs/pull/16124.

* `modules/nw-ingress-sharding-namespace-labels.adoc`: Change `routeSelector` to `namespaceSelector`.


lihongan:
Good catch! Thank you, Miciah.

PR #16649, opened 2019-09-13
fixing up broken metering links

lamek:


lamek:
Moving these changes to an existing PR.

PR #16650, opened 2019-09-14
BZ1752212: Marking iSCSI as tech preview.

huffmanca:
In a storage standup it was mentioned that iSCSI should be marked as Technology Preview in OCP 4.x. This PR marks the assembly as TP.

This is for OCP 4.x.


liangxia:
@tsmetana @jsafrane  Could you confirm ?

jsafrane:
Yes, tech preview is a good goal. The code is almost the same as 3.11, where it was GA, but with the number of issues fixed there & block support we did not want to have high expectations in 4.x.

liangxia:
LGTM ( from QE )

PR #16655, opened 2019-09-16
BZ-1729195: Indicated pods must be privileged for block volumes.

huffmanca:
Indicated pods must be privileged for block volumes.

This is for OCP 4.x.


vikram-redhat:
@liangxia could you confirm if this change is correct?

liangxia:
LGTM

PR #16662, opened 2019-09-16
Documentation "Configuring systemd-journald" lists incorrect steps

mburke5678:
According to @richm _there are a few problems with that doc_, including a step to restart Rsyslog, which is not in 4.x.
https://github.com/openshift/openshift-docs/pull/16597#discussion_r323850430

Also: https://bugzilla.redhat.com/show_bug.cgi?id=1751289

I removed the topic from 4.2 and 4.1. This PR is to fix the topic and place it back. 
PR removing for 4.1: https://github.com/openshift/openshift-docs/pull/16676

PR #16676, opened 2019-09-17
Removing logging systemd topic

mburke5678:
According to @richm _there are a few problems with that doc_, including a step to restart Rsyslog, which is not in 4.x.
https://github.com/openshift/openshift-docs/pull/16597#discussion_r323850430

Removing the topic from 4.1 and created [PR is to fix the topic and place it back](https://github.com/openshift/openshift-docs/pull/16662).


vikram-redhat:
@mburke5678 this doesn't require a QE review? I am removing the label, but please let me know if you think otherwise.

PR #16678, opened 2019-09-17
osdocs-626 preparing for disconnected installation

kalexand-rh:
https://jira.coreos.com/browse/OSDOCS-626

### DIFF
--- a/installing/installing_restricted_networks/installing-restricted-networks-preparations.adoc
+++ b/installing/installing_restricted_networks/installing-restricted-networks-preparations.adoc
@@ -1,0 +1,45 @@
+[id="installing-restricted-networks-preparations"]
+= Creating a mirror registry for installation in a restricted network
+include::modules/common-attributes.adoc[]
+:context: installing-restricted-networks-preparations
+
+toc::[]
+
+Before you install a cluster on infrastructure that you provision in a
+restricted network, you must create a mirror registry.
+
+[IMPORTANT]
+====
+You must have access to the internet to obtain the data that populates the mirror
+repository. In this procedure, you place the mirror registry on a bastion host
+that has access to both your network and the internet. If you do not have access
+to a bastion host, use the method that best fits your restrictions to bring the
+contents of the mirror registry into your restricted network.
+====
+
+include::modules/installation-about-mirror-registry.adoc[leveloffset=+1]
+
+[id="installing-preparing-bastion"]
+== Preparing the bastion host
+
+Before you create the mirror registry, you must prepare the bastion host.
+
+include::modules/cli-install.adoc[leveloffset=+2]
+
+include::modules/installation-creating-mirror-registry.adoc[leveloffset=+1]
+
+include::modules/installation-local-registry-pull-secret.adoc[leveloffset=+1]
+
+//include::modules/installation-adding-registry-pull-secret.adoc[leveloffset=+1]
+
+include::modules/installation-mirror-repository.adoc[leveloffset=+1]
+
+////
+Need to fix these links after the other PR merges.
+.Next steps
+
+* Install a cluster on infrastructure that you provision, such as
+xref:../installing/installing_vsphere/installing-vsphere.adoc#installing-vsphere[VMware vSphere]
+or
+xref:../installing/installing_bare_metal/installing-bare-metal.adoc#installing-bare-metal[bare metal].
+////
--- a/modules/installation-adding-registry-pull-secret.adoc
+++ b/modules/installation-adding-registry-pull-secret.adoc
@@ -1,0 +1,112 @@
+// Module included in the following assemblies:
+//
+// * TBD
+
+[id="installation-adding-registry-pull-secret_{context}"]
+= Adding the registry to your pull secret
+
+Modify your the pull secret for your {product-title} cluster to describe
+your local registry before you install an {product-title} cluster in a
+restricted network.
+
+.Prerequisites
+
+* You configured a mirror registry to use in your restricted network.
+
+.Procedure
+
+Complete the following steps on the bastion host:
+
+. Download your `registry.redhat.io` pull secret from the
+link:https://cloud.redhat.com/openshift/install[OpenShift Infrastructure Providers]
+page.
+
+. Generate the base64-encoded user name and password or token for your mirror
+registry:
++
+----
+$ echo -n '<user_name>:<password>' | base64 -w0 <1>
+
+BGVtbYk3ZHAtqXs=
+----
+<1> For `<user_name>` and `<password>`, specify the user name and password that
+you configured for your registry.
+
+. Make a copy of your pull secret in JSON format:
++
+----
+$ cat ./pull-secret.text | jq .  > <path>/<pull-secret-file><1>
+----
+<1> Specify the path to the folder to store the pull secret in and a name for
+the JSON file that you create.
++
+The contents of the file resemble the following example:
++
+----
+{
+  "auths": {
+    "cloud.openshift.com": {
+      "auth": "b3BlbnNo...",
+      "email": "you@example.com"
+    },
+    "quay.io": {
+      "auth": "b3BlbnNo...",
+      "email": "you@example.com"
+    },
+    "registry.connect.redhat.com": {
+      "auth": "NTE3Njg5Nj...",
+      "email": "you@example.com"
+    },
+    "registry.redhat.io": {
+      "auth": "NTE3Njg5Nj...",
+      "email": "you@example.com"
+    }
+  }
+}
+----
+
+. Edit the new file and add a section that describes your registry to it:
++
+----
+  "auths": {
+...
+    "<local_registry_host_name>:<local_registry_host_port>": { <1>
+      "auth": "<credentials>", <2>
+      "email": "you@example.com"
+  },
+...
+----
+<1> For `bastion_host_name`, specify the registry domain name
+that you specified in your certificate, and for `<local_registry_host_port>`,
+specify the port that your mirror registry uses to serve content.
+<2> For `<credentials>`, specify the base64-encoded user name and password for
+the mirror registry that you generated.
++
+The file resembles the following example:
++
+----
+{
+  "auths": {
+    "cloud.openshift.com": {
+      "auth": "b3BlbnNo...",
+      "email": "you@example.com"
+    },
+    "quay.io": {
+      "auth": "b3BlbnNo...",
+      "email": "you@example.com"
+    },
+    "registry.connect.redhat.com": {
+      "auth": "NTE3Njg5Nj...",
+      "email": "you@example.com"
+    },
+    "<local_registry_host_name>:<local_registry_host_port>": {
+      "auth": "<credentials>",
+      "email": "you@example.com"
+    },
+    "registry.redhat.io": {
+      "auth": "NTE3Njg5Nj...",
+      "email": "you@example.com"
+    }
+  }
+}
+----
--- a/modules/installation-creating-mirror-registry.adoc
+++ b/modules/installation-creating-mirror-registry.adoc
@@ -1,0 +1,149 @@
+// Module included in the following assemblies:
+//
+// * installing/installing_restricted_networks/installing-restricted-networks-preparations.adoc
+
+ifeval::["{context}" == "installing-restricted-networks-preparations"]
+:restricted:
+endif::[]
+
+[id="installation-creating-mirror-registry_{context}"]
+= Creating a mirror registry
+
+Create a registry to host the mirrored content that you require for installing
+{product-title}.
+ifdef::restricted[]
+For installation in a restricted network, you must place the mirror on your
+bastion host.
+endif::restricted[]
+
+[NOTE]
+====
+The following procedure creates a simple registry that stores data in the
+`/opt/registry` folder and runs in a `podman` container. You can use a different
+registry solution, such as
+link:https://access.redhat.com/documentation/en-us/red_hat_quay/3/html-single/manage_red_hat_quay/index#repo-mirroring-in-red-hat-quay[Red Hat Quay].
+Review the following procedure to ensure that your registry functions
+correctly.
+====
+
+.Prerequisites
+
+* You have a Red Hat Enterprise Linux (RHEL) server on your network to use
+as the registry host.
+* The registry host can access the internet.
+
+.Procedure
+
+ifdef::restricted[]
+On the bastion host, take the following actions:
+endif::restricted[]
+
+. Install the required packages:
++
+----
+# yum -y install podman httpd httpd-tools jq
+----
++
+The `podman` package provides the container package that you run the registry
+in. The `httpd` and `httpd-tools` packages provide the `htpasswd` utility, which
+you use to create users. The `jq` package improves the display of JSON output
+on your command line.
+
+. Create folders for the registry:
++
+----
+# mkdir -p /opt/registry/{auth,certs,data}
+----
++
+These folders are mounted inside the registry container.
+
+. Provide a certificate for the registry. If you do not have an existing, trusted
+certificate authority, you can generate a self-signed certificate:
++
+----
+$ cd /opt/registry/certs
+# openssl req -newkey rsa:4096 -nodes -sha256 -keyout domain.key -x509 -days 365 -out domain.crt
+----
++
+At the prompts, provide the required values for the certificate:
+[horizontal]
+Country Name (2 letter code):: Specify the two-letter ISO country code for your location.
+See the link:https://www.iso.org/iso-3166-country-codes.html[ISO 3166 country codes]
+standard.
+State or Province Name (full name):: Enter the full name of your state or province.
+Locality Name (eg, city):: Enter the name of your city.
+Organization Name (eg, company):: Enter your company name.
+Organizational Unit Name (eg, section):: Enter your department name.
+Common Name (eg, your name or your server's hostname):: Enter the host name for
+the registry host. Ensure that your hostname is in DNS and that it resolves to
+the expected IP address.
+Email Address:: Enter your email address.
+For more information, see the
+link:https://www.openssl.org/docs/man1.1.1/man1/req.html[req] description in the
+OpenSSL documentation.
+
+. Generate a user name and a password for your registry that uses the `bcrpt` format:
++
+----
+# htpasswd -bBc /opt/registry/auth/htpasswd <user_name> <password> <1>
+----
+<1> Replace `<user_name>` and `<password>` with a user name and a password.
+
+. Create the `mirror-registry` container to host your registry:
++
+----
+# podman run --name mirror-registry -p 5000:<local_registry_host_port> \ <1>
+     -v /opt/registry/data:/var/lib/registry:z \
+     -v /opt/registry/auth:/auth:z \
+     -e "REGISTRY_AUTH=htpasswd" \
+     -e "REGISTRY_AUTH_HTPASSWD_REALM=Registry Realm" \
+     -e REGISTRY_AUTH_HTPASSWD_PATH=/auth/htpasswd \
+     -v /opt/registry/certs:/certs:z \
+     -e REGISTRY_HTTP_TLS_CERTIFICATE=/certs/domain.crt \
+     -e REGISTRY_HTTP_TLS_KEY=/certs/domain.key \
+     docker.io/library/registry:2
+----
+<1> For `<local_registry_host_port>`, specify the port that your mirror registry
+uses to serve content.
+
+. Open the required ports for your registry:
++
+----
+# firewall-cmd --add-port=<local_registry_host_port>/tcp --zone=internal --permanent <1>
+# firewall-cmd --add-port=<local_registry_host_port>/tcp --zone=public   --permanent <1>
+# firewall-cmd --reload
+----
+<1> For `<local_registry_host_port>`, specify the port that your mirror registry
+uses to serve content.
+
+. Add the self-signed certificate to your list of trusted certificates:
++
+----
+# cp /opt/registry/certs/domain.crt /etc/pki/ca-trust/source/anchors/
+# update-ca-trust
+----
++
+You must trust your certificate to log in to your registry during the mirror process.
+
+. Confirm that the registry is available:
++
+----
+$ curl -u <user_name>:<password> -k https://<local_registry_host_name>:<local_registry_host_port>/v2/_catalog <1>
+
+{"repositories":[]}
+----
+<1> For `<user_name>` and `<password>`, specify the user name and password
+for your registry. For `<local_registry_host_name>`, specify the registry domain name
+that you specified in your certificate, such as `registry.example.com`. For
+`<local_registry_host_port>`, specify the port that your mirror registry uses to
+serve content.
++
+If the command output displays an empty repository, your registry is available.
+
+////
+. To stop the registry::
++
+----
+# podman stop mirror-registry
+----
+////
--- a/modules/installation-mirror-repository.adoc
+++ b/modules/installation-mirror-repository.adoc
@@ -1,0 +1,96 @@
+// Module included in the following assemblies:
+//
+// * installing/installing_restricted_networks/installing-restricted-networks-preparations.adoc
+
+[id="installation-mirror-repository_{context}"]
+= Mirroring the {product-title} image repository
+
+Mirror the {product-title} image repository to use during cluster installation
+or upgrade.
+
+.Prerequisites
+
+* You configured a mirror registry to use in your restricted network and
+can access the certificate and credentials that you configured.
+* You downloaded the pull secret from the
+link:https://cloud.redhat.com/openshift/install[OpenShift Infrastructure Providers]
+page and modified it to include authentication to your mirror repository.
+
+.Procedure
+
+Complete the following steps on the bastion host:
+
+. Review the
+link:https://access.redhat.com/downloads/content/290/[{product-title} downloads page]
+to determine the version of {product-title} that you want to install.
+
+. Set the required environment variables:
++
+----
+$ export OCP_RELEASE=<release_version> <1>
+$ export LOCAL_REGISTRY='<local_registry_host_name>:<local_registry_host_port>' <2>
+$ export LOCAL_REPOSITORY='<repository_name>' <3>
+$ export PRODUCT_REPO='openshift-release-dev' <4>
+$ export LOCAL_SECRET_JSON='<path_to_pull_secret>' <5>
+$ export RELEASE_NAME="ocp-release" <6>
+----
+<1> For `<release_version>`, specify the version number of {product-title} to
+install, such as `4.2.0`.
+<2> For `<local_registry_host_name>`, specify the registry domain name for your mirror
+repository, and for `<local_registry_host_port>`, specify the port that it
+serves content on.
+<3> For `<repository_name>`, specify the name of the repository to create in your
+registry, such as `ocp4/openshift4`.
+<4> The repository to mirror. For a production release, you must specify
+`openshift-release-dev`.
+<5> For `<path_to_pull_secret>`, specify the absolute path to and file name of
+the pull secret for your mirror registry that you created.
+<6> The release mirror. For a production release, you must specify
+`ocp-release`.
+
+. Mirror the repository:
++
+----
+$ oc adm -a ${LOCAL_SECRET_JSON} release mirror \
+     --from=quay.io/${UPSTREAM_REPO}/${RELEASE_NAME}:${OCP_RELEASE} \
+     --to=${LOCAL_REGISTRY}/${LOCAL_REPOSITORY} \
+     --to-release-image=${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}:${OCP_RELEASE}
+----
++
+This command pulls the release information as a digest, and its output includes
+text that resembles the following sample:
++
+----
+...
+Success
+Update image:  <local_registry_host_name>:<local_registry_host_port>/<local_registry>/<local_repository>:<release_version>
+Mirror prefix: <local_registry_host_name>:<local_registry_host_port>/<local_registry>/<local_repository>
+
+To use the new mirrored repository to install, add the following section to the install-config.yaml:
+
+imageContentSources:
+- mirrors:
+  - <local_registry_host_name>:<local_registry_host_port>/<repository_name>/release
+  source: quay.io/openshift-release-dev/ocp-release
+- mirrors:
+  - <local_registry_host_name>:<local_registry_host_port>/<repository_name>/release
+  source: quay.io/openshift-release-dev/ocp-v4.0-art-dev
+...
+----
+
+. Record the `imageContentSources` section from the output of the previous
+command. This information is required during {product-title} installation.
+
+. To create the installation program that is based on the content that you
+mirrored, extract it and pin it to the release:
++
+----
+$ oc adm release extract --command=openshift-install "${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}:${OCP_RELEASE}"
+----
++
+[IMPORTANT]
+====
+To ensure that you use the correct images for the version of {product-title}
+that you selected, you must extract the installation program from the mirrored
+content.
+====
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -107,7 +107,7 @@ Topics:
 - Name: Installing on GCP
   Dir: installing_gcp
   Topics:
-  - Name: Configuring an GCP account
+  - Name: Configuring a GCP account
     File: installing-gcp-account
   - Name: Installing a cluster quickly on GCP
     File: installing-gcp-default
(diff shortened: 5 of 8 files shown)


kalexand-rh:
@umohnani8, will you PTAL?

installing/installing_disconnected/installing-disconnected-preparations.adoc
```diff
@@ -1,18 +1,36 @@
...
 
 Before you install a cluster on infrastructure that you provision in a
-disconnected environment, you must prepare the environment.
+disconnected environment, you must create a mirror registry.
 
-//include::modules/cluster-entitlements.adoc[leveloffset=+1]
+.Prerequisites
 
+* You must have access to a bastion host that can access both the internet and
+your network.
```
wking: You don't need this to be a bastion host.  You can set up the registry outside, shut it down, move it into your disconnected network, and start it back up.  Or bring images in via other pathways.  Having a registry with access to both sides is just the easiest approach to setting up a mirror.

installing/installing_disconnected/installing-disconnected-preparations.adoc
```diff
@@ -1,18 +1,36 @@
...
+
+include::modules/installation-creating-mirror-registry.adoc[leveloffset=+1]
+
+include::modules/installation-adding-registry-pull-secret.adoc[leveloffset=+1]
+
+include::modules/installation-mirror-repository.adoc[leveloffset=+1]
 
 .Next steps
 
+//I need to figure out how to sequence the disconnected and proxy prep.
```
wking: These aren't inherently ordered, you just need to setup both at some point and feed their respective settings into the `install-config.yaml`.  The only wrinkle is that if each need `additionalTrustBundle` entries, you'd need to append the second entry, but if only one or the other needs additional trust, you need to create a new `additionalTrustBundle` instead of appending.  That's not all that complicated in practice, but might require some careful wording to avoid confusion.

modules/installation-about-disconnected.adoc
```diff
@@ -0,0 +1,61 @@
...
+In {product-title} {version}, you can perform an installation that does not
+require an active connection to the internet to obtain software components. You
+complete a disconnected installation on only infrastructure that you provision,
+not infrastructure that the installer provisions, so your platform selection is
+limited.
+// maybe point out that you can follow the bare metal installation rules on supported hardware and link to the matrix
+
+If you choose to perform a disconnected installation on a cloud platform, you
+still require access to its cloud APIs and will still require internet access
+to use your cluster. Depending on your network, you might require less internet
```
wking: > ... and will still require internet access to use your cluster

Can you unpack this?  I expected access to cloud APIs to be all you'd need.

modules/installation-about-disconnected.adoc
```diff
@@ -0,0 +1,61 @@
...
+access for an installation on bare metal hardware or on VMware vSphere.
+
+To complete a disconnected installation, you must create a repository that
+mirrors the contents of the {product-title} registry and contains the
+installation media. You can create this mirror on a bastion host, which can
+access both the internet and your closed network, or by using other methods
+that meet your restrictions.
```
wking: Ah, here you're softening your [earlier bastion requirement][1].  I'd move some of this wording up to the prereq section, so folks don't say "well, we can't use a bastion" and give up before getting down this far ;).

[1]: https://github.com/openshift/openshift-docs/pull/16678/#discussion_r325790509

modules/installation-about-disconnected.adoc
```diff
@@ -0,0 +1,61 @@
...
+
+[IMPORTANT]
+====
+Because disconnected installations are always on user-provisioned infrastructure,
+always complete a standard user-provisioned infrastructure installation before
+you attempt a disconnected installation.
```
wking: You could do a disconnected vSphere install, right?  And there's also at least some work on [installer-provisioned metal][1], although I'm not sure if that's supported in 4.2 or not.  So I think we want to soften this, but I'm not sure what we want the new wording to be ;).

[1]: https://github.com/openshift/installer/blob/master/docs/user/metal/install_ipi.md

modules/installation-about-disconnected.adoc
```diff
@@ -0,0 +1,61 @@
...
+====
+
+[id="installation-disconnected-limits{context}"]
+== Additional limits
+
+Disconnected clusters have the following additional limitations and restrictions:
+
+* Cluster settings always displays errors because the cluster cannot determine
```
wking: "Cluster settings" is not very specific.  Maybe "The ClusterVersion status will include `[Unable to retrieve available updates][0]` unless you update its [`spec.upstream`][1]."?  And hmm, maybe we want to patch the ClusterVersion object to empty that property, since it's optional.  Not sure if the best place to patch it is clobbering [this][2] in a multi-part install or to patch the object post-install.  Probably more reliable to patch post-install, since then you wouldn't be impacted in the unlikely event that the installer renames its generated manifest file.  But then there will be a window of time before the patch when you will still see the error.  Or maybe clearing the `upstream` field will [still give you errors][3].  @abhinavdahiya, thoughts about the best-practices config for "I don't want to upgrade this cluster ever"?

[0]: https://github.com/openshift/cluster-version-operator/blob/0725bd53c732c8e3a490d1080f03edbe7e579b3a/pkg/cvo/availableupdates.go#L165
[1]: https://github.com/openshift/api/blob/cdbe64fb0c9184ffe47d24d0ea490705a92d9fd8/config/v1/types_cluster_version.go#L54-L58
[2]: https://github.com/openshift/installer/blob/d9a9648cf2330d467cca9f2988846d031464125e/data/data/manifests/bootkube/cvo-overrides.yaml.template#L7
[3]: https://github.com/openshift/cluster-version-operator/blob/0725bd53c732c8e3a490d1080f03edbe7e579b3a/pkg/cvo/availableupdates.go#L192

modules/installation-about-disconnected.adoc
```diff
@@ -0,0 +1,61 @@
...
+if updates are available.
+* The authentication Operator might randomly fail.
```
wking: I'd leave this out until we get more specifics and evidence for it.  As it stands, I'm not sure what is going on or how sever this issue is.

modules/installation-about-disconnected.adoc
```diff
@@ -0,0 +1,61 @@
...
+* Because you cannot access the required ImageStreamTags, you cannot use the
+contents of the Developer Catalog even if it contains items.
```
wking: You could mirror these too.  We should talk to samples and OLM folks about cleanly workflows for that.  We might want to wave our hands around it here, although patching streams run by the cluster-version-operator or samples-operator requires [additional configuration][1] so your edits don't get stomped.  Once [rhbz#1741391][2] gets fixed, you won't need to patch the CVO-managed image streams anymore, although you will still need to patch any samples-operator-managed image streams (and possibly OLM stuff) because only by-digest pullspecs will fall back to mirrors automatically.

[1]: https://github.com/openshift/release/pull/4585/commits/75de00dca094ad41790187b1902b3d327b5401fd#diff-e9ff14653190666556f490397d19c553R205-R238
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1741391

modules/installation-about-disconnected.adoc
```diff
@@ -0,0 +1,61 @@
...
+* The `TelemeterClientDown` and `Watchdog` alerts from the monitoring Operator
```
wking: `Watchdog` always fires, that's the point ;).  I dunno if it's worth calling out here.

Does `TelemeterClientDown` still happen when you [explicitly disable telemetry][1]?  That's what I'd expect folks to do for disconnected clusters, although they should obviously be aware of [the caveats][2].

[1]: https://docs.openshift.com/container-platform/4.1/telemetry/opting-out-of-telemetry.html#disabling-telemetry_telemetry
[2]: https://docs.openshift.com/container-platform/4.1/telemetry/opting-out-of-telemetry.html#telemetry-consequences-of-disabling-telemetry_telemetry

modules/installation-about-disconnected.adoc
```diff
@@ -0,0 +1,61 @@
...
+== About the mirror repository
+
+The key component that is require to complete a disconnected installation is
+the mirror repository. This repository contains both the installation media
+and the images that the cluster requires.
+
+Because of the way that {product-title} verifies integrity for the release
+payload, the image references in your local registry are identical to the ones
+that are hosted on
+link:
```
wking: FIXME?  "link:" -> "quay.io."?  "on link:" -> "by Red Hat."?

modules/installation-about-disconnected.adoc
```diff
@@ -0,0 +1,61 @@
...
+During the bootstrapping process of installation, the images must have the same
+names no matter which repository they are pulled from.
```
wking: "names" -> "digests"?

modules/installation-about-disconnected.adoc
```diff
@@ -0,0 +1,61 @@
...
+To ensure that the release payload is identical, you mirror the repository.
```
wking: "repository" -> "images"?  Or "you mirror the images to your local repository"?

modules/installation-about-disconnected.adoc
```diff
@@ -0,0 +1,61 @@
...
+
+You must specify separate mirrors for both the images and the release package itself.
```
wking: You can use the same mirror for both the release image and the secondard images referenced from the release image (which is what [the installer docs][1] do).

[1]: https://github.com/openshift/installer/blob/master/docs/user/customization.md#image-content-sources

modules/installation-adding-registry-pull-secret.adoc
```diff
@@ -0,0 +1,114 @@
...
+  }
+}
+----
+
+. Edit the new file and add a section that describes your registry to it:
++
+----
+  "auths": {
+...
+    "<bastion_host_name>:5000": { <1>
```
wking: Doesn't have to be `:5000`, your local registry could serve HTTPS on the default 443.  You could use `<local_registry_authority>` to lean on [the URI nomenclature][1] or `<local_registry_host_port>` because we don't actually want the authority's optional `userinfo "@"` here.

[1]: https://tools.ietf.org/html/rfc3986#section-3.2

modules/installation-adding-registry-pull-secret.adoc
```diff
@@ -0,0 +1,114 @@
...
+      "email": "you@example.com"
+    },
+    "registry.redhat.io": {
+      "auth": "NTE3Njg5Nj...",
+      "email": "you@example.com"
+    }
+  }
+}
+----
+//This is wrong - you need one more curly brace at the end.
```
wking: FIXME?

modules/installation-adding-registry-pull-secret.adoc
```diff
@@ -0,0 +1,114 @@
...
+    },
+    "registry.redhat.io": {
+      "auth": "NTE3Njg5Nj...",
+      "email": "you@example.com"
+    }
+  }
+}
+----
+
+. Edit the new file and add a section that describes your registry to it:
```
wking: Do we care about the cloud.redhat.com pull-secret entries?  You'll need those to create the mirror, and to submit telemetry.  So maybe for setups where you had a proxy.  But for a truly disconnected install I don't see how the cluster would actually use them, which would allow simpler docs like:

> Create a pull secret containing only your local registry authentication information...

modules/installation-mirror-repository.adoc
```diff
@@ -0,0 +1,101 @@
...
+link:https://cloud.redhat.com/openshift/install[OpenShift Infrastructure Providers]
+page and modified it to include authentication to your mirror repository.
+
+.Procedure
+
+Complete the following steps on the bastion host:
+
+. Review the
+link:https://quay.io/repository/openshift-release-dev/ocp-release?tab=tags[Quay.io]
+repository to determine the version of {product-title} that you want to install.
```
wking: Hmm.  Is this the UX we want?  I'd expect folks would want to use the most-recent release in their channel, which you can get with:

```console
$ curl -sH 'Accept:application/json' 'https://api.openshift.com/api/upgrades_info/v1/graph?channel=stable-4.1' | jq -r '.nodes | sort_by(.version | sub(".*[.]"; "") | tonumber)[-1] | .version + "\t" + .payload'
4.1.16	quay.io/openshift-release-dev/ocp-release@sha256:61ed953962d43cae388cb3c544b4cac358d4675076c2fc0befb236209d5116f7
```

replacing `stable-4.1` with your favorite channel.  Or use `[]` instead of `[-1]` to get a list of all the choices.  And we could do that filtering in JavaScript if we wanted something live in the served docs.  Or are we not allowed to rely on JavaScript or something?

Lacking JavaScript, I'd expect there would be a customer-portal link we could point folks at with a list of supported versions.

modules/installation-mirror-repository.adoc
```diff
@@ -0,0 +1,101 @@
...
+
+. Set the required environment variables:
++
+----
+$ export OCP_RELEASE=<release_version> <1>
+$ export AIRGAP_REG='<bastion_host_name>:5000' <2>
+$ export AIRGAP_REPO='<repo_name>' <3>
+$ export PRODUCT_REPO='openshift-release-dev' <4>
+$ export AIRGAP_SECRET_JSON='<path_to_pull_secret>${HOME}/pull-secret-2.json' <5>
+$ export OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE=${AIRGAP_REG}/${AIRGAP_REPO}:${OCP_RELEASE} <6>
```
wking: We don't want this; more on what we should use instead in a later comment.

modules/installation-mirror-repository.adoc
```diff
@@ -0,0 +1,101 @@
...
+$ export RELEASE_NAME="ocp-release" <7>
+----
+<1> For `<release_version>`, specify the version number of {product-title} to
+install, such as `4.2.0`.
+<2> For `<bastion_host_name>`, specify the registry domain name for your mirror
+repository.
+<3> For `<repo_name>`, specify the name of the repository to create in your
+registry. Use a descriptive name, such as `ocp4/openshift4`.
```
wking: nit: "Use a descriptive name" seems overly prescriptive.  Maybe just "...in your registry, such as `ocp4/openshift4`."

modules/installation-mirror-repository.adoc
```diff
@@ -0,0 +1,101 @@
...
+<4> The repository to mirror. For a production installation, you must specify
```
wking: "installation" -> "release".  Mirroring manipulates release images.  We haven't gotten as far as installation yet.

modules/installation-adding-registry-pull-secret.adoc
```diff
@@ -0,0 +1,114 @@
...
+
+BGVtbYk3ZHAtqXs=
+----
+<1> For `<user_name>` and `<password>`, specify the user name and password that
+you configured for your registry.
+
+. Make a copy of your pull secret:
++
+----
+$ cat ./pull-secret.text | jq .  > <path>/pull-secret-2.json <1>
```
wking: if the first was `.text`, the second can just be `pull-secret.json`.  Dropping the `-2` here will also decouple subsequent docs from the number of steps taken when generating this file.

modules/installation-mirror-repository.adoc
```diff
@@ -0,0 +1,101 @@
...
+repository to determine the version of {product-title} that you want to install.
+
+. Set the required environment variables:
++
+----
+$ export OCP_RELEASE=<release_version> <1>
+$ export AIRGAP_REG='<bastion_host_name>:5000' <2>
+$ export AIRGAP_REPO='<repo_name>' <3>
+$ export PRODUCT_REPO='openshift-release-dev' <4>
+$ export AIRGAP_SECRET_JSON='<path_to_pull_secret>${HOME}/pull-secret-2.json' <5>
```
wking: nit: this seems to mix `<placeholder>` with an example value.  For the other variables, you seem to just use `<placeholder>` and then give example values down below under `<1>` and similar.

modules/installation-mirror-repository.adoc
```diff
@@ -0,0 +1,101 @@
...
+     --to=${AIRGAP_REG}/${AIRGAP_REPO} \
+     --to-release-image=${AIRGAP_REG}/${AIRGAP_REPO}:${OCP_RELEASE}
+----
++
+This command pulls the release information as a digest. The command output
+resembles the following text:
++
+----
+Success
+Update image:  <bastion_host_name>:5000/ocp/release:<release_version>
```
wking: If you want to fully `<placeholder>` this, it would be `<airgap_registry>/<airgap_repo>:<release_version>`.

modules/installation-mirror-repository.adoc
```diff
@@ -0,0 +1,101 @@
...
+. Review the
+link:https://quay.io/repository/openshift-release-dev/ocp-release?tab=tags[Quay.io]
+repository to determine the version of {product-title} that you want to install.
+
+. Set the required environment variables:
++
+----
+$ export OCP_RELEASE=<release_version> <1>
+$ export AIRGAP_REG='<bastion_host_name>:5000' <2>
+$ export AIRGAP_REPO='<repo_name>' <3>
```
wking: Suggest dropping `AIRGAP_`, because the fact that `oc` can reach both Red Hat and the mirror means the mirror itself is probably not airgapped.  [Earlier][1] I floated "local registry" or similar, so maybe `LOCAL_REGISTRY` and `LOCAL_REPO` if you wanted to follow that pattern?

[1]: https://github.com/openshift/openshift-docs/pull/16678#discussion_r325831243

modules/installation-mirror-repository.adoc
```diff
@@ -0,0 +1,101 @@
...
+imageContentSources:
+- mirrors:
+  - <bastion_host_name>:5000/<repo_name>/release
+  source: quay.io/openshift-release-dev/ocp-release
+- mirrors:
+  - <bastion_host_name>:5000/<repo_name>/release
+  source: registry.svc.ci.openshift.org/ocp/release
+
+
+To use the new mirrored repository for upgrades, use the following to create an ImageContentSourcePolicy:
```
wking: This part isn't useful for install, so I'd elide it.  We just need the output through the `imageContentSources` block above.

modules/installation-mirror-repository.adoc
```diff
@@ -0,0 +1,101 @@
...
+  - mirrors:
+    - <bastion_host_name>:5000/ocp/release
+    source: quay.io/openshift-release-dev/ocp-release
+  - mirrors:
+    - <bastion_host_name>:5000/ocp/release
+    source: registry.svc.ci.openshift.org/ocp/release
+----
+
+. Record the `imageContentSources` section from the output of the previous
+command. This information is required during {product-title} installation.
```
wking: And here at the bottom I'd run:

```command
oc adm release extract --command=openshift-install "${AIRGAP_REG}/${AIRGAP_REPO}:${OCP_RELEASE}"
```

(or whatever the variables composing the mirror target end up being).  Then you can use `./openshift-install ...` to run the extracted installer without having to set `OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE` and we can drop all the docs for this internal, test-only environment variable.

installing/installing_disconnected/installing-disconnected-preparations.adoc
kalexand-rh: I'm making a follow-up JIRA to review this sequencing after the proxy content merges: https://jira.coreos.com/browse/OSDOCS-693

modules/installation-about-disconnected.adoc
kalexand-rh: @wking, I might have misunderstood, but don't you need internet access to use the cloud APIs?

modules/installation-about-disconnected.adoc
kalexand-rh: vSphere is technically user-provisioned infrastructure, right?

modules/installation-adding-registry-pull-secret.adoc
kalexand-rh: I'd fixed it and missed the note. :/

modules/installation-adding-registry-pull-secret.adoc
kalexand-rh: Probably not. I'll make a separate module with the simpler workflow.

modules/installation-mirror-repository.adoc
kalexand-rh: Do you have that customer portal link?

modules/installation-adding-registry-pull-secret.adoc
kalexand-rh: @wking, that's fair. Do you have any other guidance for a suggested port? I'd be more likely to use `<local_registry_host_port>` so that I can say to open it [here](https://github.com/openshift/openshift-docs/pull/16678/files#diff-7d4591e9ba80df3ebf307fb24525011bR100).

modules/installation-mirror-repository.adoc
kalexand-rh: That's because the instructions specified a specific value for the pull secret file - you pointed out an improvement to it in a previous comment. I'm adjusting all of the pull secret file names to be generic.

modules/installation-about-disconnected.adoc
kalexand-rh: I'm not sure. I'll remove the restriction.

modules/installation-about-disconnected.adoc
kalexand-rh: I'll soften this a little for now and see if more changes need to be made after the OLM disconnected docs land.

modules/installation-about-disconnected.adoc
kalexand-rh: Can you tell me more about this patch? Do we want users to patch this field? Can I just tell them that there's an error?

modules/installation-about-disconnected.adoc
wking: AWS populates routes to some of its endpoints, so you can reach them from inside the VPC, even if you've blocked all "external" traffic.  Or something like that.  Some things (like IAM) don't get this treatment, so the cred operator would want a way to get out to AWS's external IAM endpoint.

modules/installation-about-disconnected.adoc
wking: Telling them ClusterVersion will report an error (more specific than what you have now) is fine, but if it was my cluster I'd patch to remove it ;).

modules/installation-adding-registry-pull-secret.adoc
wking: 5000 seems popular for "I'm running some dev thing".  For production, I'd expect a registry running on 443, but I don't think these docs are where we want to walk folks through production registries.

modules/installation-about-disconnected.adoc
kalexand-rh: @wking, do you have information about patching `spec.upstream` somewhere?

modules/installation-adding-registry-pull-secret.adoc
kalexand-rh: @wking, we need to provide enough information that they could take this example and configure a production registry. I'll make it generic.

Is there any other information that would be valid in a production environment that we don't have yet?

kalexand-rh:
@wking, I made updates based on your excellent feedback. Will you PTAL?

modules/installation-about-disconnected.adoc
wking: Should be:

```console
$ oc patch clusterversion version --type json -p '{"op": "remove", "path": "/spec/upstream"}'
```
But let me sanity check, and I'll report back.

modules/installation-adding-registry-pull-secret.adoc
wking: Note still here?  Do you have some local work you haven't pushed up?

modules/installation-adding-registry-pull-secret.adoc
wking: > Is there any other information that would be valid in a production environment that we don't have yet?

If you mirror to an existing, production registry, you probably just need the `oc adm release mirror ...` step and `imageContentSources` block.  You won't need anything for `additionalTrustBundle` to support the mirror registry, because it will already have a cert signed by a RHCOS-trusted CA.  If the mirrored images are public on the mirror registry, you won't need a mirror token in your pull secret.  I dunno, there's not all that much else going on...

modules/installation-mirror-repository.adoc
wking: https://access.redhat.com/downloads/content/290/

modules/installation-about-disconnected.adoc
wking: So for "I want to watch for updates in a different location":

```console
$ oc patch clusterversion version --type json -p '[{"op": "add", "path": "/spec/upstream", "value": "https://example.com/your/new/location"}]'
```

And for "I don't want to check for upgrades, I'll either never upgrade (!) or manually upgrade (e.g. via `oc adm upgrade ...`)":

```console
$ oc patch clusterversion version --type json -p '[{"op": "remove", "path": "/spec/channel"}, {"op": "remove", "path": "/spec/upstream"}]'
```

in which case ClusterVersion will have a `status.conditions` entry like:

```yaml
  - lastTransitionTime: "2019-09-21T22:36:52Z"
    message: The update channel has not been configured.
    reason: NoChannel
    status: "False"
    type: RetrievedUpdates
```

which seems appropriate.

modules/installation-mirror-repository.adoc
```diff
@@ -0,0 +1,87 @@
...
+- mirrors:
+  - <bastion_host_name>:<local_registry_host_port>/<repository_name>/release
+  source: registry.svc.ci.openshift.org/<local_registry>/<local_repository>
+...
+----
+
+. Record the `imageContentSources` section from the output of the previous
+command. This information is required during {product-title} installation.
+
+. Create the release package from the content that you mirrored:
```
wking: "Create the release package" -> "Extract the installation program"

wking: And maybe "to pin it to the mirrored release" or some such?  We don't want folks thinking, "oh, I already have an `openshift-install`" and skipping this step.

modules/installation-about-restricted-network.adoc
```diff
@@ -0,0 +1,63 @@
...
+error.
+//* The authentication Operator might randomly fail.
+* By default, you cannot use the contents of the Developer Catalog because
+ you cannot access the required ImageStreamTags.
+//* The `TelemeterClientDown` and `Watchdog` alerts from the monitoring Operator always display.
+
+[id="installation-restricted-network-mirror_{context}"]
+== About the mirror repository
+
+The key component that is require to complete an installation in a restricted
```
wking: nit: "require" -> "required".  Or maybe "The key component that is required to complete an installation" -> "The key to installation" or some such.

modules/installation-adding-registry-pull-secret.adoc
```diff
@@ -0,0 +1,112 @@
...
+  "auths": {
+...
+    "<bastion_host_name>:<local_registry_host_port>": { <1>
+      "auth": "<credentials>", <2>
+      "email": "you@example.com"
+  },
+...
+----
+<1> For `bastion_host_name`, specify the registry domain name
+that you specified in your certificate, and for `<local_registry_host_port`,
```
wking: nit: `<local_registry_host_port` -> `<local_registry_host_port>`.  And you can leave off the `:port` if you're using the standard HTTP(S) of 80/443, although it may not be worth calling that out.

modules/installation-creating-mirror-registry.adoc
```diff
@@ -0,0 +1,139 @@
...
++
+----
+# htpasswd -bBc /opt/registry/auth/htpasswd <user_name> <password> <1>
+----
+<1> Replace `<user_name>` and `<password>` with a user name and a password.
+
+. Create the `poc-registry` container to host your registry:
++
+----
+# podman run --name poc-registry -p <local_registry_host_port>:<local_registry_host_port> \ <1>
```
wking: The internal port here needs to be 5000, because that's what `docker.io/library/registry:2` is listening on inside the container.  External port can be whatever.  So this should be `-p 5000:<local_registry_host_port>`.

wking: Also, maybe`poc-registry`->`mirror-registry` or something?  With these docs, this is no longer a proof of concept ;).

modules/installation-mirror-repository.adoc
```diff
@@ -0,0 +1,87 @@
...
+
+. Review the
+link:https://quay.io/repository/openshift-release-dev/ocp-release?tab=tags[Quay.io]
+repository to determine the version of {product-title} that you want to install.
+
+. Set the required environment variables:
++
+----
+$ export OCP_RELEASE=<release_version> <1>
+$ export LOCAL_REGISTRY='<bastion_host_name>:<local_registry_host_port>' <2>
```
wking: You don't have to mirror to a bastion; you could be mirroring to your corporate proxy or whatever for auditing/performance.  I think we should use `<local_registry_host_name>` or some such instead of `<bastion_host_name>`.  I guess it depends on how tightly we want to couple `modules/installation-mirror-repository.adoc` to `modules/installation-creating-mirror-registry.adoc` vs. allowing for pre-existing mirror registries.

modules/installation-mirror-repository.adoc
```diff
@@ -0,0 +1,87 @@
...
+     --from=quay.io/${UPSTREAM_REPO}/${RELEASE_NAME}:${OCP_RELEASE} \
+     --to=${LOCAL_REGISTRY}/${LOCAL_REPOSITORY} \
+     --to-release-image=${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}:${OCP_RELEASE}
+----
++
+This command pulls the release information as a digest. The command output
+resembles the following text:
++
+----
+Success
```
wking: There's a bunch of output before about what's getting mirrored.  Maybe "The command output resembles the following text" -> "the command output includes text like the following" and then either begin and end with ellipses or without ellipses.

modules/installation-mirror-repository.adoc
```diff
@@ -0,0 +1,87 @@
...
+
+To use the new mirrored repository to install, add the following section to the install-config.yaml:
+
+imageContentSources:
+- mirrors:
+  - <bastion_host_name>:<local_registry_host_port>/<repository_name>/release
+  source: quay.io/openshift-release/ocp-release
+- mirrors:
+  - <bastion_host_name>:<local_registry_host_port>/<repository_name>/release
+  source: registry.svc.ci.openshift.org/<local_registry>/<local_repository>
```
wking: This source is what you'd see when mirroring from an internal CI release.  For production 4.2 releases the sources should be `quay.io/openshift-release-dev/ocp-release` (for the release image) and `quay.io/openshift-release-dev/ocp-v4.0-art-dev` (for the images that the release image references).

modules/installation-about-disconnected.adoc
wking: Although it looks like we may want to file something around NoChannel for a console update, because `/settings/cluster` currently puts up a moderately frightening stop sign and doesn't expose the reason or message (unless I'm just missing something about the UI):

Screenshot from 2019-09-21 16-21-12

modules/installation-mirror-repository.adoc
kalexand-rh: I'm adding some conditionals to limit "bastion" to the restricted network topic and generally refer to it as the "registry host" instead.

modules/installation-about-restricted-network.adoc
kalexand-rh: I'd take "The key to installation, but because we mean "key" as an adjective here, we have to make the nouns explicit. It's a requirement in our style guide that helps with translation and improves clarity for non-native English speakers.

modules/installation-about-disconnected.adoc
kalexand-rh: Did you file a bug about that console issue? I think that more text would certainly help there.

I'm making a [follow-up issue ](https://jira.coreos.com/browse/OSDOCS-695)to describe patching the CVO. I don't see any modifications to it in our current doc set, and with our docs workflow, it will be easier to get reviews on the proposed change on a separate PR.

modules/installation-adding-registry-pull-secret.adoc
kalexand-rh: I'm glad that it's that easy.

kalexand-rh:

bobfuru (commented):
A couple of comments and otherwise LGTM!

installing/installing_restricted_networks/installing-restricted-networks-preparations.adoc
```diff
@@ -0,0 +1,45 @@
+[id="installing-restricted-networks-preparations"]
+= Creating a mirror registry for installation in a restricted network
```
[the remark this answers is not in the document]
kalexand-rh: I'll clarify that in line 8.

PR #16680, opened 2019-09-17
Removed configuring knative services section

abrennan89:
This config gets overwritten by the operator, so I am removing it for now. We plan to add a new section on configuring serving correctly once we have details for this.


jboxman:
lgtm

PR #16687, opened 2019-09-17
BZ-1752553: Indicated that the default SC is Operator-owned.

huffmanca:
Added notes to indicate that the default StorageClass is owned by the ClusterStorageOperator, and changes, beyond adjusting the annotations or labels, should not be made.

This is for OCP 4.x.


liangxia:
LGTM (from QE)

PR #16688, opened 2019-09-17
BZ-1743397: Included block detailing not to use vSphere storage for registry storage.

huffmanca:
Included block detailing not to use vSphere storage for registry storage. vSphere storage does not support ReadWriteMany, which is a requirement for the image registry storage.

This is for OCP 4.x.


wzheng1:
LGTM

PR #16691, opened 2019-09-17
suggested update to machineset definition for labels

kalexand-rh:
Per Jake Lucky in Slack.

### DIFF
--- a/modules/machineset-yaml-aws.adoc
+++ b/modules/machineset-yaml-aws.adoc
@@ -28,6 +28,8 @@ spec:
   selector:
     matchLabels:
       machine.openshift.io/cluster-api-cluster: <infrastructureID> <1>
+      machine.openshift.io/cluster-api-machine-role: <role> <3>
+      machine.openshift.io/cluster-api-machine-type: <role> <3>
       machine.openshift.io/cluster-api-machineset: <infrastructureID>-<role>-<zone> <2>
   template:
     metadata:
@@ -39,7 +41,7 @@ spec:
     spec:
       metadata:
         labels:
-          node-role.kubernetes.io/<role>: "" <3>
+          node-role.kubernetes.io/<role>: "true" <2>
       providerSpec:
         value:
           ami:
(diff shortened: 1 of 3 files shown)


kalexand-rh:
@sunzhaohua2, will you PTAL?

kalexand-rh:
@jhou1, will you PTAL?

modules/machineset-yaml-aws.adoc
```diff
@@ -39,7 +41,7 @@ spec:
     spec:
       metadata:
         labels:
-          node-role.kubernetes.io/<role>: "" <3>
+          node-role.kubernetes.io/<role>: "true" <2>
```
jhou1: This is the node label to add. So `<2>` should be `<3>`

modules/machineset-yaml-aws.adoc
```diff
@@ -28,6 +28,8 @@ spec:
   selector:
     matchLabels:
       machine.openshift.io/cluster-api-cluster: <infrastructureID> <1>
+      machine.openshift.io/cluster-api-machine-role: <role> <3>
+      machine.openshift.io/cluster-api-machine-type: <role> <3>
```
jhou1: `<3>` should be `<2>` for above 2 lines

modules/machineset-yaml-aws.adoc
```diff
@@ -84,6 +86,6 @@ command:
 ----
 $ oc get -o jsonpath='{.status.infrastructureName}{"\n"}' infrastructure cluster
 ----
-<2> Specify the infrastructure ID, node label, and zone.
+<2> Specify the infrastructure ID, node label, and zone. Use a descriptive name for the node label, such as `infra` for infrastructure machines.
```
jhou1: Use a descriptive name for the **machine** label,

modules/machineset-yaml-azure.adoc
```diff
@@ -88,7 +88,7 @@ command:
 ----
 $ oc get -o jsonpath='{.status.infrastructureName}{"\n"}' infrastructure cluster
 ----
-<2> Specify the node label to add.
+<2> Specify the node label to add, such as `infra` for infrastructure machines.
```
jhou1: such as `infra` for infrastructure **nodes**.

jhou1:
Hi @kalexand-rh, in a machineSet, the labels nested like the following will be applied to the nodes.

```
spec:
  template:
...
    spec:
      metadata:
        labels:
          node-role.kubernetes.io/<role>: "true"
```

modules/machineset-yaml-aws.adoc
jhou1: Actually no need to update here, if we want the machines to be labeled, we should update 

```
spec:
  template:
    metadata:
      labels:
```

PR #16776, opened 2019-09-20
Removed Scaling the Prometheus Components section

ahardin-rh:
https://jira.coreos.com/browse/OSDOCS-692


ahardin-rh:
Confirmed these changes with @sjug and @ashishkamra. No QE necessary, as this is a content removal.

PR #16780, opened 2019-09-20
adding usage examples

lamek:


sheriff-rh (approved):
Looks great!

lamek:
@pruan-rht This is ready for QE review.

pruan-rht:
@lamek, looks good.  Do we really need all these examples for the various reports?  I think it might be an overkill.  Perhaps one example of hourly, daily, runImmediately, and aggregated reports would be sufficient.

lamek:
@pruan-rht I'm honestly not sure yet. I think we need more customer feedback. You make a great point. Having fewer examples with better annotations would make the page easier to read, but having a lot of them like this demonstrates everything you can do with metering, and makes it very easy to copy/paste.

Unless you feel strongly, I'm going to merge with the current number of reports. I think post release we will focus a long in the "using" section, and use feedback from customer to add more examples and more context.

pruan-rht:
@lamek currently the examples only focused on two types of reports, `hourly` and `daily`.  There are also `RunImmediately`, `cron` and report aggregation.

pruan-rht:
@lamek my main concern is information overload.  If we want to be through about various reports, I think it's more helpful to the end user to demonstrate the various types of the report YAML rather than just focusing on just two types.  IMO, one example os each possible reporting formats would be sufficient.

lamek:
@pruan-rht I did some significant updates and removed a lot of the examples. PTAL. The hourly, daily, and aggregate daily are all part of the first section. The next sections include a one-time report and using a cron period.
https://metering-chargeback-showback--ocpdocs.netlify.com/openshift-enterprise/latest/metering/metering-usage-examples.html

pruan-rht:
@lamek lgtm

PR #16788, opened 2019-09-23
osdocs-477 Azure machinesets

kalexand-rh:
https://jira.coreos.com/browse/OSDOCS-477

@enxebre, will you PTAL?

### DIFF
--- a/modules/machineset-yaml-azure.adoc
+++ b/modules/machineset-yaml-azure.adoc
@@ -1,0 +1,93 @@
+// Module included in the following assemblies:
+//
+// * machine_management/creating-infrastructure-machinesets.adoc
+// * machine_management/creating-machineset-azure.adoc
+
+[id="machineset-yaml-azure_{context}"]
+=  Sample YAML for a MachineSet Custom Resource on Azure
+
+This sample YAML defines a MachineSet that runs in the `centralus1`
+Microsoft Azure region and creates nodes that are labeled with
+`node-role.kubernetes.io/<role>: ""`
+
+In this sample, `<infrastructureID>` is the infrastructure ID label that is
+based on the cluster ID that you set when you provisioned
+the cluster, and `<role>` is the node label to add.
+
+[source,yaml]
+----
+apiVersion: machine.openshift.io/v1beta1
+kind: MachineSet
+metadata:
+  labels:
+    machine.openshift.io/cluster-api-cluster: <infrastructureID> <1>
+    machine.openshift.io/cluster-api-machine-role: <role> <2>
+    machine.openshift.io/cluster-api-machine-type: <role> <2>
+  name: <infrastructureID>-<role>-<region> <3>
+  namespace: openshift-machine-api
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      machine.openshift.io/cluster-api-cluster: <infrastructureID> <1>
+      machine.openshift.io/cluster-api-machineset: <infrastructureID>-<role>-<region> <3>
+  template:
+    metadata:
+      creationTimestamp: null
+      labels:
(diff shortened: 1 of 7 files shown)


modules/machineset-yaml-azure.adoc
```diff
@@ -0,0 +1,91 @@
...
+      providerSpec:
+        value:
+          apiVersion: azureproviderconfig.openshift.io/v1beta1
+          credentialsSecret:
+            name: azure-cloud-credentials
+            namespace: openshift-machine-api
+          image:
+            offer: ""
+            publisher: ""
+            resourceID: /resourceGroups/<infrastructureID>-rg/providers/Microsoft.Compute/images/<infrastructureID>
```
kalexand-rh: Is this value static for all of the machinesets in your cluster?

modules/machineset-yaml-azure.adoc
```diff
@@ -0,0 +1,91 @@
...
+            sku: ""
+            version: ""
+          internalLoadBalancer: ""
+          kind: AzureMachineProviderSpec
+          location: centralus
+          managedIdentity: <infrastructureID>-identity <1>
```
kalexand-rh: Is this right for a new machineset?

modules/machineset-yaml-azure.adoc
```diff
@@ -0,0 +1,91 @@
...
+            managedDisk:
+              storageAccountType: Premium_LRS
+            osType: Linux
+          publicIP: false
+          publicLoadBalancer: ""
+          resourceGroup: <infrastructureID>-rg <1>
+          sshPrivateKey: ""
+          sshPublicKey: ""
+          subnet: <infrastructureID>-<role>-subnet <1> <2>
+          userDataSecret:
```
kalexand-rh: For a new machineset, how do you need to change the subnet and userDataSecret parameter values?

kalexand-rh (commented):
@enxebre, I have a couple of questions. Will you PTAL?

I'm also curious if anything else needs to change when you create a new machineset.

kalexand-rh:

modules/machineset-yaml-azure.adoc
enxebre: Not necessarily but most of the times it will. This is the os image, like the `ami` in aws https://github.com/openshift/openshift-docs/pull/16788/files#diff-70576f5487cf72d64bf5875f88d7eb0dR46

modules/machineset-yaml-azure.adoc
enxebre: the `userDataSecret` should be the same than the other worker machineSets unless you create yourself a custom one

modules/machineset-yaml-azure.adoc
kalexand-rh: Is it common enough to be able to reuse the same resourceID that I don't need to provide a description of when you'd change it?

modules/machineset-yaml-azure.adoc
```diff
@@ -0,0 +1,91 @@
...
+            name: <role>-user-data <2>
+          vmSize: Standard_D2s_v3
+          vnet: <infrastructureID>-vnet <1>
+          zone: "1"
+----
+<1> Specify the the infrastructure ID that is based on the cluster ID that
```
bobfuru: d/the

bobfuru (commented):
Minor edits (extra "the"); otherwise, LGTM!

kalexand-rh:
Per Vikram, I'm merging without QE. I'll open a follow-up PR to address more comments.
@sunzhaohua2, PTAL?

vikram-redhat:
@jhou has confirmed this has passed QE.

PR #16791, opened 2019-09-23
CNV 2.1 upgrade + install docs

ousleyp:
Note: The upgrade docs depend on/overlap with the installation guide updates for 2.1. For now, I'm bundling all the updates in this one PR, but if that turns out to be counterproductive, I'll split them up.

Relevant Jiras: CNV-2278, CNV-2473, CNV-2581

**Preview builds**

Upgrade: http://file.bos.redhat.com/pousley/100319/cnv-2.1-upgrade/cnv/cnv_install/upgrading-container-native-virtualization.html

Install: http://file.bos.redhat.com/pousley/100319/cnv-2.1-upgrade/cnv/cnv_install/installing-container-native-virtualization.html

Uninstall: http://file.bos.redhat.com/pousley/100319/cnv-2.1-upgrade/cnv/cnv_install/uninstalling-container-native-virtualization.html

PTAL, @openshift/team-documentation - thanks!

Please cherrypick to enterprise-4.2 only.

### DIFF
--- a/cnv/cnv_install/upgrading-container-native-virtualization.adoc
+++ b/cnv/cnv_install/upgrading-container-native-virtualization.adoc
@@ -1,0 +1,19 @@
+[id="upgrading-container-native-virtualization"]
+= Upgrading container-native virtualization
+include::modules/cnv-document-attributes.adoc[]
+:context: upgrading-container-native-virtualization
+toc::[]
+
+You enable automatic updates during {CNVProductName} xref:../../cnv/cnv_install/installing-container-native-virtualization.adoc#cnv-subscribing-to-hco-catalog_installing-container-native-virtualization[installation].
+Learn what to expect and how you can check the status of an update in progress.
+
+:FeatureName: {CNVProductNameStart}
+include::modules/technology-preview.adoc[leveloffset=+1]
+
+include::modules/cnv-about-upgrading-cnv.adoc[leveloffset=+1]
+
+include::modules/cnv-monitoring-upgrade-status.adoc[leveloffset=+1]
+
+.Additional information
+
+* xref:../../applications/operators/olm-understanding-olm.adoc#olm-csv_olm-understanding-olm[ClusterServiceVersions (CSVs)]
--- a/modules/cnv-about-upgrading-cnv.adoc
+++ b/modules/cnv-about-upgrading-cnv.adoc
@@ -1,0 +1,38 @@
+// Module included in the following assemblies:
+//
+// * cnv/cnv_install/upgrading-container-native-virtualization.adoc
+
+[id="cnv-about-upgrading-cnv_{context}"]
+= About upgrading container-native virtualization
+
+If you enabled automatic updates when you installed {CNVProductName}, you
+receive updates as they become available.
+
+.Additional information
+
+* In {CNVProductName} version 2.1, only _z-stream_ updates are available.
+For example, {CNVProductName} 2.1.0 -> {CNVProductName} 2.1.1
+
+* Updates are delivered via the _Marketplace Operator_, which is deployed
+during {product-title} installation. The Marketplace Operator makes
+external Operators available to your cluster.
+
+* Upgrading does not interrupt virtual machine workloads.
+** Virtual machine Pods are not restarted or migrated during an upgrade. If you
+need to update the `virt-launcher` Pod, you must restart or live migrate the
+virtual machine.
++
+[NOTE]
+====
+Each virtual machine has a `virt-launcher` Pod that runs the virtual machine
+instance. The `virt-launcher` Pod runs an instance of `libvirt`, which is
+used to manage the virtual machine process.
+====
+
+* Upgrading does not interrupt network connections.
+
+* DataVolumes and their associated PersistentVolumeClaims are preserved during
+upgrade.
+
+* The amount of time an update takes to complete depends on your network
+connection. Most automatic updates complete within fifteen minutes.
--- a/modules/cnv-monitoring-upgrade-status.adoc
+++ b/modules/cnv-monitoring-upgrade-status.adoc
@@ -1,0 +1,48 @@
+// Module included in the following assemblies:
+//
+// * cnv/cnv_install/upgrading-container-native-virtualization.adoc
+
+[id="cnv-monitoring-upgrade-status_{context}"]
+= Monitoring upgrade status
+
+The best way to monitor {CNVProductName} upgrade status is to watch the
+ClusterServiceVersion (CSV) `PHASE`.
+
+.Prerequisites
+
+* Access to the cluster as a user with the `cluster-admin` role.
+* Install the OpenShift Command-line Interface (CLI), commonly known as `oc`.
+
+.Procedure
+
+. Run the following command:
++
+----
+$ oc get csv
+----
+
+. Review the output, checking the `PHASE` field. For example:
++
+----
+VERSION  REPLACES                                        PHASE
+2.1.1    kubevirt-hyperconverged-operator.v2.1.0         Installing
+2.1.0                                                    Replacing
+----
+
+. Optional: Monitor the aggregated status of all {CNVProductName} component
+`Conditions` by running the following command:
++
+----
+$ oc get hco -n openshift-cnv hyperconverged-cluster \
+-o=jsonpath='{range .status.conditions[*]}{.type}{"\t"}{.status}{"\t"}{.message}{"\n"}{end}'
+----
++
+A successful upgrade results in the following output:
++
+----
+ReconcileComplete  True  Reconcile completed successfully
+Available          True  Reconcile completed successfully
+Progressing        False Reconcile completed successfully
+Degraded           False Reconcile completed successfully
+Upgradeable        True  Reconcile completed successfully
+----
--- a/modules/cnv-preparing-to-install.adoc
+++ b/modules/cnv-preparing-to-install.adoc
@@ -5,48 +5,18 @@
 [id="cnv-preparing-to-install_{context}"]
 = Preparing to install container-native virtualization
 
-Before deploying {CNVProductName}:
-
-* Create a namespace called `openshift-cnv`.
-* Create `OperatorGroup` and `CatalogSource` Custom Resource Definition objects
-(CRDs) in the `openshift-cnv` namespace.
+Before deploying {CNVProductName}, create a namespace that is named
+`openshift-cnv`.
 
 .Prerequisites
 
-* {product-title} 4.1
 * User with `cluster-admin` privileges
-* The {product-title} Command-line Interface (CLI), commonly known as `oc`
 
 .Procedure
 
-. Create the `openshift-cnv` namespace by running the following
+* Create the `openshift-cnv` namespace by running the following
 command:
 +
 ----
-$ oc new-project openshift-cnv
-----
-
-. Create the `OperatorGroup` and `CatalogSource` in the
-`openshift-cnv` namespace by running the following command:
-+
-----
-cat <<EOF | oc apply -f -
-apiVersion: operators.coreos.com/v1alpha2
-kind: OperatorGroup
-metadata:
-  name: hco-operatorgroup
-  namespace: openshift-cnv
----
-apiVersion: operators.coreos.com/v1alpha1
-kind: CatalogSource
-metadata:
-  name: hco-catalogsource
-  namespace: openshift-operator-lifecycle-manager
-  imagePullPolicy: Always
-spec:
-  sourceType: grpc
-  image: registry.redhat.io/container-native-virtualization/hco-bundle-registry:v2.0.0
-  displayName: KubeVirt HyperConverged
-  publisher: Red Hat
-EOF
+$ oc create ns openshift-cnv
 ----
--- a/modules/cnv-subscribing-to-hco-catalog.adoc
+++ b/modules/cnv-subscribing-to-hco-catalog.adoc
@@ -12,28 +12,37 @@ namespace access to the {CNVProductName} Operators.
 
 .Prerequisites
 
-* `OperatorGroup` and `CatalogSource` Custom Resource Definition objects (CRDs),
-both created in the `openshift-cnv` namespace
+* Create a namespace that is named `openshift-cnv`.
 
 .Procedure
 
-. Open a browser window and navigate to the {product-title} web console.
+. Open a browser window and log in to the {product-title} web console.
 
-. Navigate in the web console to the *Operators → OperatorHub* page.
+. Navigate to the *Operators* → *OperatorHub* page.
 
-. Scroll or type a keyword into the *Filter by keyword* box to find the KubeVirt
-HyperConverged Cluster Operator, then click on it.
+. Locate the *KubeVirt HyperConverged Cluster Operator* and then select it.
 
 . Read the information about the Operator and click *Install*.
 
 . On the *Create Operator Subscription* page:
-.. Select one of the following:
-*** *All namespaces on the cluster (default)* installs the Operator in the default
+.. Select *A specific namespace on the cluster* from the *Installation Mode*
+list and choose the `openshift-cnv` namespace.
++
+[WARNING]
+====
+* *All namespaces on the cluster (default)* installs the Operator in the default
 `openshift-operators` namespace to watch and be made available to all namespaces
-in the cluster. This option is not always available.
-*** *A specific namespace on the cluster* allows you to choose a specific, single
+in the cluster. This option is not supported for use with {CNVProductName}.
+You must only install the Operator in the `openshift-cnv` namespace.
+* *A specific namespace on the cluster* allows you to choose a specific, single
 namespace in which to install the Operator. The Operator will only watch and be
 made available for use in this single namespace.
+====
+.. Select *2.1* from the list of available *Update Channel* options.
+.. For *Approval Strategy*, ensure that *Automatic*, which is the default value,
+is selected.
+{CNVProductNameStart} automatically updates when a new z-stream release is
+available.
 
 . Click *Subscribe* to make the Operator available to the selected namespaces on
 this {product-title} cluster.
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -1057,6 +1057,8 @@ Topics:
     File: installing-container-native-virtualization
   - Name: Installing the `virtctl` client
     File: cnv-installing-virtctl
+  - Name: Upgrading container-native virtualization
+    File: upgrading-container-native-virtualization
   - Name: Uninstalling container-native virtualization
     File: uninstalling-container-native-virtualization
 - Name: Container-native virtualization user's guide
--- a/cnv/cnv_install/installing-container-native-virtualization.adoc
+++ b/cnv/cnv_install/installing-container-native-virtualization.adoc
@@ -7,19 +7,14 @@ toc::[]
 Install {CNVProductName} to add virtualization functionality to your {product-title}
 cluster.
 
-Before you deploy {CNVProductName}, you must create two Custom Resource
-Definition (CRD) objects:
-
-* `kind: OperatorGroup`
-* `kind: CatalogSource`
-
-You can create both objects by running a single command.
-
-To finish installing {CNVProductName}, use the {product-title} 4.1
+You can use the {product-title} 4.2
 xref:../../web-console/web-console.adoc#web-console-overview_web-console[web console]
 to subscribe to and deploy the {CNVProductName} Operators.
 
-:FeatureName: {CNVProductName}
+.Prerequisites
+* {product-title} 4.2
+
+:FeatureName: {CNVProductNameStart}
 include::modules/technology-preview.adoc[leveloffset=+1]
 
 include::modules/cnv-preparing-to-install.adoc[leveloffset=+1]
--- a/cnv/cnv_install/uninstalling-container-native-virtualization.adoc
+++ b/cnv/cnv_install/uninstalling-container-native-virtualization.adoc
@@ -4,14 +4,12 @@ include::modules/cnv-document-attributes.adoc[]
 :context: uninstalling-container-native-virtualization
 toc::[]
 
-You can uninstall {CNVProductName} by using the {product-title} 4.1
+You can uninstall {CNVProductName} by using the {product-title}
 xref:../../web-console/web-console.adoc#web-console-overview_web-console[web console].
-First, delete the custom resource you created during deployment. Then, delete
-the *KubeVirt HyperConverged Cluster Operator* catalog subscription.
 
 .Prerequisites
 
-* {CNVProductNameStart} 2.0
+* {CNVProductNameStart} 2.1
 
 include::modules/cnv-deleting-custom-resource.adoc[leveloffset=+1]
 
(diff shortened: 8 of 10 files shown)


ousleyp:
Attn @nellyc @rthallisey - I wrote notes in the relevant Jiras but this is ready for review. I am concerned that I might not have fully understood the google docs that were provided, so please let me know if anything is inaccurate. Thanks!

cnv/cnv_install/upgrading-container-native-virtualization.adoc
```diff
@@ -0,0 +1,19 @@
+[id="upgrading-container-native-virtualization"]
+= Upgrading {CNVProductName}
+include::modules/cnv-document-attributes.adoc[]
+:context: upgrading-container-native-virtualization
+toc::[]
+
+You enable automatic updates during {CNVProductName} xref:../../cnv/cnv_install/installing-container-native-virtualization.adoc#cnv-subscribing-to-hco-catalog_installing-container-native-virtualization[installation].
```
stu-gott: I'm uncertain of the phrasing here. How about "Automatic updates are enabled during installation"? or "can be" if it's optional.

modules/cnv-about-upgrading-cnv.adoc
```diff
@@ -0,0 +1,31 @@
...
+** For example: {CNVProductName} 2.1.0 -> {CNVProductName} 2.1.1
+
+* Updates are delivered via the _Marketplace Operator_, which is deployed
+during {product-title} installation. The Marketplace Operator makes
+external Operators available to your cluster.
+
+* Upgrading does not interrupt virtual machine workloads.
+** Because workloads are uninterrupted, virtual machine Pods are not restarted
+or migrated during an upgrade. If you need to update the `virt-launcher` Pod,
+you must restart or live migrate it.
```
stu-gott: you must restart or live migrate the virtual machine.

modules/cnv-about-upgrading-cnv.adoc
```diff
@@ -0,0 +1,31 @@
...
+
+* Upgrading does not interrupt network connections.
+
+* DataVolumes and their backing PersistentVolumeClaims are preserved during
```
stu-gott: Suggest: DataVolumes and their associated PersistentVolumeClaims

stu-gott (commented):
Nice job!

modules/cnv-about-upgrading-cnv.adoc
ousleyp: Good catch, thanks! Fixed.

modules/cnv-about-upgrading-cnv.adoc
ousleyp: Sounds good to me. I made the change. Thanks!

cnv/cnv_install/upgrading-container-native-virtualization.adoc
ousleyp: It does sound strange, but a lot of our content includes "You <verb> <the thing>" phrasing. We try to avoid the passive voice, and "can" is sometimes confusing/unclear.

I am discussing this via email with Steve G and Ryan, because I think we don't *want* it to be optional, but it currently is optional in the UI. I'll leave this comment open and update the phrasing when we come to a conclusion. :)

(Thanks for the review, @stu-gott!)

modules/cnv-subscribing-to-hco-catalog.adoc
```diff
@@ -12,28 +12,36 @@ namespace access to the {CNVProductName} Operators.
...
 
-* `OperatorGroup` and `CatalogSource` Custom Resource Definition objects (CRDs),
-both created in the `openshift-cnv` namespace
+* Create a project called `openshift-cnv`.
 
 .Procedure
 
 . Open a browser window and navigate to the {product-title} web console.
 
 . Navigate in the web console to the *Operators → OperatorHub* page.
```
alexxa: I would drop "in the web console" and replace to:  Navigate to the *Operators → OperatorHub* page. or In the left menu navigate to the ...

modules/cnv-subscribing-to-hco-catalog.adoc
```diff
@@ -12,28 +12,36 @@ namespace access to the {CNVProductName} Operators.
 
 .Prerequisites
 
-* `OperatorGroup` and `CatalogSource` Custom Resource Definition objects (CRDs),
-both created in the `openshift-cnv` namespace
+* Create a project called `openshift-cnv`.
 
 .Procedure
 
 . Open a browser window and navigate to the {product-title} web console.
```
alexxa: s/navigate to /log in

modules/cnv-subscribing-to-hco-catalog.adoc
```diff
@@ -12,28 +12,36 @@ namespace access to the {CNVProductName} Operators.
...
 
 . Navigate in the web console to the *Operators → OperatorHub* page.
 
-. Scroll or type a keyword into the *Filter by keyword* box to find the KubeVirt
-HyperConverged Cluster Operator, then click on it.
+. Scroll or type a keyword into the *Filter by keyword* box to find the
+{CNVProductNameStart} Operator, then click on it.
```
alexxa: why a capital case for CNVProductName here? It's not a beginning of the sentence. For example, a paragraph above, before "Prerequisites" has it with low case + Operator.

modules/cnv-subscribing-to-hco-catalog.adoc
```diff
@@ -12,28 +12,36 @@ namespace access to the {CNVProductName} Operators.
...
 . On the *Create Operator Subscription* page:
-.. Select one of the following:
+.. Select *A specific namespace on the cluster* from the *Installation Mode*
+list and choose the `openshift-cnv` namespace.
++
+[NOTE]
+====
 *** *All namespaces on the cluster (default)* installs the Operator in the default
 `openshift-operators` namespace to watch and be made available to all namespaces
 in the cluster. This option is not always available.
```
alexxa: This option is not always available or supported maybe? Please double check with @rthallisey maybe.

modules/cnv-subscribing-to-hco-catalog.adoc
```diff
@@ -12,28 +12,36 @@ namespace access to the {CNVProductName} Operators.
...
 *** *A specific namespace on the cluster* allows you to choose a specific, single
 namespace in which to install the Operator. The Operator will only watch and be
 made available for use in this single namespace.
+====
+.. Select the 2.1 *Update Channel* option.
```
alexxa: @rthallisey now there is only "alpha" channel. Will it be changed to "2.1"?

modules/cnv-subscribing-to-hco-catalog.adoc
```diff
@@ -12,28 +12,36 @@ namespace access to the {CNVProductName} Operators.
...
+.. Select an *Approval Strategy* option:
+*** *Automatic* is recommended. If you select this option, {CNVProductName}
+automatically updates when a new z-stream release is available.
```
alexxa: I would also mention how it works with "Manual" option.

modules/cnv-about-upgrading-cnv.adoc
```diff
@@ -0,0 +1,31 @@
...
+* In {CNVProductName} version 2.1, only _z-stream_ updates are available.
+** For example: {CNVProductName} 2.1.0 -> {CNVProductName} 2.1.1
+
+* Updates are delivered via the _Marketplace Operator_, which is deployed
+during {product-title} installation. The Marketplace Operator makes
+external Operators available to your cluster.
+
+* Upgrading does not interrupt virtual machine workloads.
+** Because workloads are uninterrupted, virtual machine Pods are not restarted
+or migrated during an upgrade. If you need to update the `virt-launcher` Pod,
```
slintes: Just curious, do people know what the `virt-launcher` pod is (the pod which actually runs the VMI)?
But I also have no better idea to describe this :)

modules/cnv-preparing-to-install.adoc
```diff
@@ -5,17 +5,11 @@
 [id="cnv-preparing-to-install_{context}"]
 = Preparing to install {CNVProductName}
 
-Before deploying {CNVProductName}:
-
-* Create a namespace called `openshift-cnv`.
-* Create `OperatorGroup` and `CatalogSource` Custom Resource Definition objects
-(CRDs) in the `openshift-cnv` namespace.
+Before deploying {CNVProductName}, create a namespace called `openshift-cnv`.
```
slintes: IMHO this page is obsolete. The creation of the namespace / project is also mentioned on the `Subscribing to the KubeVirt HyperConverged Cluster Operator catalog` page. The prerequisites and procedure from this page can go there as well. WDYT?

alexxa (changes requested):

modules/cnv-subscribing-to-hco-catalog.adoc
ousleyp: Ah, good catch! I think I misinterpreted this guideline: https://github.com/ousleyp/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#operator-name-capitalization

I missed the "defer to the product's capitalization style" part (not to mention that I wasn't consistent about this). I will change it back to lower case.

modules/cnv-about-upgrading-cnv.adoc
ousleyp: I agree it's a good idea to add a little more context here, so I'll try to include your brief description. Thanks for catching that!

modules/cnv-preparing-to-install.adoc
ousleyp: It is just a single command at this point, but I'm not sure if we should be switching between the terminal and the UI in the same procedure. I can't have two procedures in one module (both of these files that you referred to are modules, and they are arranged in an assembly that encompasses the entire user story). Thus, I think I might keep this in, at least until it becomes possible to create "openshift-*" namespaces from the UI. I'll consider it further, though. Thanks!

modules/cnv-subscribing-to-hco-catalog.adoc
ousleyp: Good catch, that was leftover from before the first step existed.

modules/cnv-subscribing-to-hco-catalog.adoc
ousleyp: Good call, @alexxa 
@rthallisey do you think it makes sense to change this to just say:
"This option is not supported for use with container-native virtualization." ? I don't think the "This option is not always available" part is very helpful anyway.

modules/cnv-subscribing-to-hco-catalog.adoc
ousleyp: My impression is that "Manual" is not actually supported. I'm waiting on @rthallisey 's response to an email from Steve G. It sounded like the way the options are displayed in the UI might be changed so that automatic is the default. I'll rephrase this when I hear more.

modules/cnv-about-upgrading-cnv.adoc
ousleyp: @slintes Do you think this is accurate?
"Each virtual machine has a `virt-launcher` Pod that runs the virtual machine
instance. The `virt-launcher` Pod contains an instance of `libvirt`, which is
used to manage the virtual machine process."

slintes: I'd prefer `The virt-launcher Pod *runs* an instance...`, otherwise :+1:

modules/cnv-subscribing-to-hco-catalog.adoc
```diff
@@ -12,28 +12,36 @@ namespace access to the {CNVProductName} Operators.
 
 .Prerequisites
 
-* `OperatorGroup` and `CatalogSource` Custom Resource Definition objects (CRDs),
-both created in the `openshift-cnv` namespace
+* Create a project called `openshift-cnv`.
```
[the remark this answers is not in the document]
slintes: Are you sure? To my best knowledge in Openshift we talk about projects. Technically they are the same though. I would align this with the rest of the Openshift docs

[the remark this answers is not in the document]
alexxa: yes, there is a [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1732932)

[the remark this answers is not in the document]
slintes: oh, interesting, thx for clarification :+1:

modules/cnv-subscribing-to-hco-catalog.adoc
rthallisey: I think we want to limit the deploy to `openshift-cnv` since it's how QE tested and some of CNV's operator are singletons.

modules/cnv-subscribing-to-hco-catalog.adoc
rthallisey: We don't need to mention Manual, but rather something like "use the default option". 

`Use the default upgrade policy, *Automatic*`

modules/cnv-about-upgrading-cnv.adoc
ousleyp: Sounds good, I'll make that change now.

modules/cnv-subscribing-to-hco-catalog.adoc
ousleyp: @rthallisey Okay, so then the way it is currently phrased seems correct. The step instructs the user to "Select *A specific namespace on the cluster* from the *Installation Mode*
list and choose the `openshift-cnv` namespace." It describes the two options but says that the "all namespaces" option is not supported for use with CNV. 

Or did you mean that the other installation mode option is going to be removed from the UI?

modules/cnv-subscribing-to-hco-catalog.adoc
```diff
@@ -12,28 +12,36 @@ namespace access to the {CNVProductName} Operators.
...
-. Open a browser window and navigate to the {product-title} web console.
+. Open a browser window and log in to the {product-title} web console.
 
-. Navigate in the web console to the *Operators → OperatorHub* page.
+. Navigate to the *Operators → OperatorHub* page.
 
-. Scroll or type a keyword into the *Filter by keyword* box to find the KubeVirt
-HyperConverged Cluster Operator, then click on it.
+. Scroll or type a keyword into the *Filter by keyword* box to find the
+{CNVProductName} Operator, then click on it.
```
alexxa: s/{CNVProductName}/KubeVirt HyperConverged Cluster
why it was replaces  wise verse at first place?
image

alexxa (changes requested):
@ousleyp please see a few more suggestions. Then it should be good to merge.

modules/cnv-about-upgrading-cnv.adoc
slintes: Not sure if I understand...?
IMHO it's clear that "uninterrupted" is in the context of upgrades only.
But removing that part also works.
I don't have a preference.

modules/cnv-monitoring-upgrade-status.adoc
```diff
@@ -0,0 +1,43 @@
...
++
+----
+VERSION  REPLACES                                  PHASE
+2.0.99                                             Installing
+2.1.0    kubevirt-hyperconverged-operator.v2.0.99  Pending
+----
++
+[WARNING]
+====
+`PHASE` represents the system's best guess based on available information.
```
rthallisey: `Conditions` represent the system's best guess based on available information.

modules/cnv-monitoring-upgrade-status.adoc
```diff
@@ -0,0 +1,43 @@
...
+====
+
+. (Optional) Monitor the status of all {CNVProductName} components by running
```
rthallisey: Monitor the aggregated status of all {CNVProductName} component `Conditions` by running

modules/cnv-subscribing-to-hco-catalog.adoc
ousleyp: Good question. I could have sworn that it was the Container-native Virtualization Operator in the material @rthallisey provided, but looking at it now I just see Kubevirt HyperConverged Cluster Operator. Will fix.

modules/cnv-about-upgrading-cnv.adoc
ousleyp: @alexxa I like your suggestion to remove "Because workloads are uninterrupted." I think the phrase is somewhat confusing in the context of that sentence, and the intended meaning is already conveyed through the bullet point levels.

ousleyp:
@alexxa  I made the changes, but I'm having some issues with getting the preview build to work (the upgrade changes aren't appearing for some reason). I'll let you know when it is functional, but in the meantime, feel free to look at the latest commit. Thanks!

alexxa (approved):

modules/cnv-subscribing-to-hco-catalog.adoc
rthallisey: I would say to only document the specific namespace method, so it's clear that's the path we want to steer the user.

kalexand-rh (commented):
I have a few picks and suggestions, but it's looking good.

modules/cnv-monitoring-upgrade-status.adoc
```diff
@@ -0,0 +1,54 @@
...
+A successful upgrade results in the following output:
++
+----
+ReconcileComplete  True  Reconcile completed successfully
+Available          True  Reconcile completed successfully
+Progressing        False Reconcile completed successfully
+Degraded           False Reconcile completed successfully
+Upgradeable        True  Reconcile completed successfully
+----
+
```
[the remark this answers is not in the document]
alexxa: may it stay as it's now. We give that command to monitor the status of the upgrade. The given output will be displayed only when the upgrade is finished, as the statement above says.

modules/cnv-subscribing-to-hco-catalog.adoc
```diff
@@ -12,28 +12,36 @@ namespace access to the {CNVProductName} Operators.
...
-. Scroll or type a keyword into the *Filter by keyword* box to find the KubeVirt
-HyperConverged Cluster Operator, then click on it.
+. Scroll or type a keyword into the *Filter by keyword* box to find the
+*KubeVirt HyperConverged Cluster Operator*, then click on it.
 
 . Read the information about the Operator and click *Install*.
 
 . On the *Create Operator Subscription* page:
-.. Select one of the following:
+.. Select *A specific namespace on the cluster* from the *Installation Mode*
```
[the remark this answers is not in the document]
alexxa: There are two of them.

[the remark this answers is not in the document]
kalexand-rh: Thanks! Then you need to present the two options in an unordered list and make it clear that you do one of them.

modules/cnv-monitoring-upgrade-status.adoc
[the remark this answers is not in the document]
kalexand-rh: As long as the note's nested the right way, yes.

modules/cnv-monitoring-upgrade-status.adoc
```diff
@@ -0,0 +1,54 @@
...
+[id="cnv-monitoring-upgrade-status_{context}"]
+= Monitoring upgrade status
+
+The best way to monitor {CNVProductName} upgrade status is to watch the
+ClusterServiceVersion (CSV) `PHASE`.
+
+.Prerequisites
+
+* Access to the cluster as a user with the `cluster-admin` role.
+* The {product-title} CLI (`oc`) installed.
```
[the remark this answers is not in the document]
ousleyp: @kalexand-rh I'm not quite sure what you mean -- should I use the longer construction here? The PR comments make it seem like that is still up in the air.

modules/cnv-monitoring-upgrade-status.adoc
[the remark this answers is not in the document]
ousleyp: I actually intended for it to *not* be connected to the command, because `PHASE` refers to the first part of the procedure, and `Conditions` refers to the second, optional part. Do you think that makes sense?

modules/cnv-monitoring-upgrade-status.adoc
```diff
@@ -0,0 +1,54 @@
...
+ReconcileComplete  True  Reconcile completed successfully
+Available          True  Reconcile completed successfully
+Progressing        False Reconcile completed successfully
+Degraded           False Reconcile completed successfully
+Upgradeable        True  Reconcile completed successfully
+----
+
+[NOTE]
+====
+`PHASE` and `Conditions` both represent the system's best guess based on
```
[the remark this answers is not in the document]
ousleyp: `PHASE` is called out in the first part of the procedure. I think that `Conditions` needs to be added to the sample output for the last command.

modules/cnv-subscribing-to-hco-catalog.adoc
```diff
@@ -12,28 +12,36 @@ namespace access to the {CNVProductName} Operators.
...
+list and choose the `openshift-cnv` namespace.
++
+[NOTE]
+====
```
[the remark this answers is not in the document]
ousleyp: It seems right to me. The note is aligned with step 5a.

modules/cnv-subscribing-to-hco-catalog.adoc
```diff
@@ -12,28 +12,36 @@ namespace access to the {CNVProductName} Operators.
...
 *** *All namespaces on the cluster (default)* installs the Operator in the default
 `openshift-operators` namespace to watch and be made available to all namespaces
-in the cluster. This option is not always available.
+in the cluster. This option is not supported for use with {CNVProductName}.
 *** *A specific namespace on the cluster* allows you to choose a specific, single
 namespace in which to install the Operator. The Operator will only watch and be
 made available for use in this single namespace.
+====
+.. Select the *2.1* *Update Channel* option.
```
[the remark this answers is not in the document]
ousleyp: This is actually supposed to be more like "Select *2.1* from the list of available *Update Channel* options."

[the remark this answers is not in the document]
ousleyp: I think there's only one option, which is why I originally went with that (admittedly awkward) construction.

modules/cnv-subscribing-to-hco-catalog.adoc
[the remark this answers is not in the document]
ousleyp: I'm a bit confused. My understanding is that we don't want the customer to think that they could choose either option -- they need to pick "A specific namespace on the cluster." I presented both options in the note in order to explain them/give context, because both options do appear in the UI. Is that sufficient or should I rewrite the step?

modules/cnv-monitoring-upgrade-status.adoc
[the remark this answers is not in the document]
kalexand-rh: It's used in 21 places in the repo now. I'd use it, and if we pick different language, it'll get updated with the others.

modules/cnv-monitoring-upgrade-status.adoc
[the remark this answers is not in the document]
kalexand-rh: After getting a little more information, I think this note needs to be removed.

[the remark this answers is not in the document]
alexxa: >> I don't see either PHASE or Conditions called out.
Phase and Conditions are not things to be called out, it's like a variable displaying the status of the things. Phase title is displayed by `oc get csv` command (value is Installing/Replacing/ Succeeded, ..), and conditions by that complicated yaml parsing command. 
>> After getting a little more information, I think this note needs to be removed.
Bring it back please. Currently, we need to stress, that e.g. even if the Phase reports "Succeeded", it may happen that not all things went/work well, therefore it's "best guess based on... "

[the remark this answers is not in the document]
ousleyp: Opened PR #17095 to address this.

PR #16796, opened 2019-09-23
Fixed command for finding OAuth namespace route.

cajosc:


kalexand-rh:
@stuartchuan, will you PTAL and confirm that this is valid in 4.1 and 4.2?

stuartchuan:
Yes, it's valid in both 4.1 and 4.2.

PR #16818, opened 2019-09-23
fixing 'log into'

kalexand-rh:
@bobfuru, PTAL

### DIFF
--- a/logging/config/cluster-logging-configuring.adoc
+++ b/logging/config/cluster-logging-configuring.adoc
@@ -41,11 +41,11 @@ spec:
       redundancyPolicy: SingleRedundancy
       resources:
         limits:
-          cpu: 
-          memory: 
+          cpu:
+          memory:
         requests:
-          cpu: 
-          memory: 
+          cpu:
+          memory:
       storage: {}
     type: elasticsearch
   managementState: Managed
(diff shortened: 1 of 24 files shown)


logging/config/cluster-logging-configuring.adoc
```diff
@@ -59,7 +59,7 @@ spec:
 ----
 You can configure the following for cluster logging:
 
-* You can place cluster logging into an unmanaged state that allows an administrator to assume full control of individual
+* You can place cluster logging in to an unmanaged state that allows an administrator to assume full control of individual
```
bobfuru: I don't think "into" needs to be separated in this case. Depends on whether it is referring to "cluster logging" or "cluster log in" as a concept, if that makes sense.

modules/ignition-config-viewing.adoc
```diff
@@ -97,7 +97,7 @@ into data URLs, which tends to make the content a bit clumsy to read. (Use the
  on NFS that configures that service, you would just add an NFS configuration
  file, which would then be started by the init process when the system comes up.
 * users: A user named core is created, with your ssh key assigned to that user.
-This will allow you to log into the cluster with that user name and your
+This allow you to log in to the cluster with that user name and your
```
bobfuru: s/allow/allows

bobfuru (commented):
Couple of comments and LGTM!

logging/config/cluster-logging-configuring.adoc
kalexand-rh: You're right - it doesn't need to be separated.

PR #16849, opened 2019-09-24
fix broken links

mburke5678:
Removed bad link and removed `relfileprefix` from various Nodes files.


mburke5678:
@vikram-redhat FYI

PR #16883, opened 2019-09-25
[WIP] CNV-2593 Fencing

aburdenthehand:
New assembly and modules to cover the basic information for fencing in cnv 2.1

Cherrypick to enterprise 4.2 only.

### DIFF
--- a/modules/cnv-about-fencing.adoc
+++ b/modules/cnv-about-fencing.adoc
@@ -1,4 +1,3 @@
-
 // Module included in the following assemblies:
 //
 // * cnv/cnv_users_guide/cnv-fencing-nodes.adoc
--- a/modules/cnv-about-fencing.adoc
+++ b/modules/cnv-about-fencing.adoc
@@ -22,7 +22,7 @@ The fencing feature introduces three new controllers, which are deployed by the
 
 * `MachineDisruptionBudget` controller
 * `MachineHealthCheck` controller
-* `MachineRemedation` controller
+* `MachineRemediation` controller
 
 The `MachineDisruptionBudget` controller creates a `MachineDisruptionBudget` object 
 on each node to monitor the health of the nodes and tracks the number of 
@@ -32,11 +32,11 @@ the number of healthy or unhealthy nodes are not within the conditions set in th
 `MachineDisruptionBudget` object on the node.
 
 The `MachineHealthCheck` controller monitors node health. If a node becomes 
-unhealthy for longer than 5 minutes, the controller creates a `MachineRemedation` 
+unhealthy for longer than 5 minutes, the controller creates a `MachineRemediation` 
 Custom Resource (CR) for the node, provided it is permitted by the 
 `MachineDisruptionBudget` controller.
 
-`MachineRemedation` controller monitors the `MachineRemedation` CRs and applies 
+`MachineRemediation` controller monitors the `MachineRemediation` CRs and applies 
 the remediation strategy specified in the CR. 
 
 
--- a/modules/cnv-about-fencing.adoc
+++ b/modules/cnv-about-fencing.adoc
@@ -16,25 +16,20 @@ Fencing the node ensures that the workloads are no longer running on the
 unhealthy node, protecting the persistent storage. Once fencing is complete, 
 workloads can be safely replaced on other nodes.
 
-The fencing feature introduces three new controllers, which are deployed by the 
+The fencing feature introduces new controllers, which are deployed by the 
 `MachineRemediation` Operator:
 
-* `MachineDisruptionBudget` controller
+* `MaxUnhealthy` controller
 * `MachineHealthCheck` controller
-* `MachineRemediation` controller
+// * `MachineRemediation` controller - delayed to 4.4
 
-The `MachineDisruptionBudget` controller creates a `MachineDisruptionBudget` object 
-for each node to track the number of healthy nodes in the cluster,
-the desired number of healthy nodes in the cluster, and how many disruptions are
-allowed. This controller prevents node deletion if the number of healthy or unhealthy nodes are not within the conditions set in the `MachineDisruptionBudget` object on 
-the node.
+The `MaxUnhealthy` tracks the number of unhealthy nodes in the cluster. 
+This controller prevents node deletion if the number of unhealthy nodes are not within the allowed range.
 
-The `MachineHealthCheck` controller monitors node health. If a node becomes 
-unhealthy for longer than 5 minutes, the controller creates a `MachineRemediation` 
-Custom Resource (CR) for the node, provided it is permitted by the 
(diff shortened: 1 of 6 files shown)


modules/cnv-about-fencing.adoc
```diff
@@ -0,0 +1,41 @@
...
+workloads can be safely replaced on other nodes.
+
+The fencing feature introduces three new controllers, which are deployed by the 
+`MachineRemediation` Operator:
+
+* `MachineDisruptionBudget` controller
+* `MachineHealthCheck` controller
+* `MachineRemediation` controller
+
+The `MachineDisruptionBudget` controller creates a `MachineDisruptionBudget` object
```
beekhof: Technically the `MachineDisruptionBudget` objects are created by the admin. There is also a default one created by the MachineRemediationOperator

modules/cnv-about-fencing.adoc
```diff
@@ -0,0 +1,41 @@
...
+
+The fencing feature introduces three new controllers, which are deployed by the 
+`MachineRemediation` Operator:
+
+* `MachineDisruptionBudget` controller
+* `MachineHealthCheck` controller
+* `MachineRemediation` controller
+
+The `MachineDisruptionBudget` controller creates a `MachineDisruptionBudget` object 
+on each node to monitor the health of the nodes and tracks the number of
```
beekhof: s/on/for/

beekhof: To be pedantic, MDB is only involved in the tracking of how many healthy/desired nodes there are. It is MachineHealthCheck that monitors node health.

modules/cnv-monitoring-fencing-events.adoc
```diff
@@ -0,0 +1,47 @@
+// Module included in the following assemblies:
+//
+// * cnv/cnv_users_guide/cnv-fencing-nodes.adoc
+
+[id="cnv-monitoring-fencing-events_{context}"]
+
+= Monitoring fencing events
```
beekhof: In 4.3 it should be possible to look for k8s events instead.
    `oc get events -n openshift-machine-api --field-selector involvedObject.kind=Machine`
See https://asciinema.org/a/ZYkbsUyXciEwPe1ZUlsifObhK, you can see the full list of events at the 1:40 mark.

vikram-redhat:
@aburdenthehand did this not make it for CNV 2.1?

modules/cnv-about-fencing.adoc
```diff
@@ -0,0 +1,35 @@
...
+[id="cnv-about-fencing_{context}"]
+
+= About fencing nodes
+
+Fencing is the act of protecting shared resources in order to safely recover 
+from node failure.
+
+In the case of an unhealthy node, Pods and virtual machines that depend on 
+persistent storage cannot be replaced on another node until it can be confirmed 
+that the original Pod or virtual machine is no longer accessing the storage.
```
beekhof: If we dropped the reference to virtual machines, the remainder would be completely appropriate to include in the main OCP docs.  Maybe we should do that?

modules/cnv-about-fencing.adoc
```diff
@@ -0,0 +1,35 @@
+// Module included in the following assemblies:
+//
+// * cnv/cnv_users_guide/cnv-fencing-nodes.adoc
+
+[id="cnv-about-fencing_{context}"]
+
+= About fencing nodes
+
+Fencing is the act of protecting shared resources in order to safely recover 
+from node failure.
```
beekhof: Perhaps:  Fencing is the act of putting a node into a state where it can be known not to be accessing shared resources,   in order to make use of them elsewhere after a node level failure.

modules/cnv-about-fencing.adoc
```diff
@@ -0,0 +1,35 @@
...
+
+In the case of an unhealthy node, Pods and virtual machines that depend on 
+persistent storage cannot be replaced on another node until it can be confirmed 
+that the original Pod or virtual machine is no longer accessing the storage. 
+Fencing the node ensures that the workloads are no longer running on the 
+unhealthy node, protecting the persistent storage. Once fencing is complete, 
+workloads can be safely replaced on other nodes.
+
+The fencing feature introduces new controllers, which are deployed by the 
+`MachineRemediation` Operator:
```
beekhof: s/MachineRemediation Operator/Machine API Operator/

modules/cnv-about-fencing.adoc
```diff
@@ -0,0 +1,35 @@
...
+
+* `MaxUnhealthy` controller
```
beekhof: MaxUnhealthy is a field on the Machine HealthCheck CRD, not a separate controller.  Sorry for any confusion

modules/cnv-about-fencing.adoc
```diff
@@ -0,0 +1,35 @@
...
+* `MachineHealthCheck` controller
+// * `MachineRemediation` controller - delayed to 4.4
+
+The `MaxUnhealthy` tracks the number of unhealthy nodes in the cluster.
```
beekhof: Everything he is correct apart from it happening as part of the MachineHealthCheck controller

modules/cnv-disabling-fencing-on-node.adoc
```diff
@@ -0,0 +1,21 @@
+// Module included in the following assemblies:
+//
+// * cnv/cnv_users_guide/cnv-fencing-nodes.adoc
+
+[id="cnv-disabling-fencing-on-node_{context}"]
```
beekhof: Unfortunately it seems this capability did not make it into 4.3, but I'm working to confirm that

modules/cnv-enabling-fencing-on-node.adoc
```diff
@@ -0,0 +1,19 @@
+// Module included in the following assemblies:
+//
+// * cnv/cnv_users_guide/cnv-fencing-nodes.adoc
+
+[id="cnv-enabling-fencing-on-node_{context}"]
```
beekhof: As above :(

beekhof (changes requested):

beekhof:
One more thing, in moving the implementation from CNV to OCP 4.3, we also lost the ability for fencing to be enabled by default. So we'll need to cover that too :-(

modules/cnv-about-fencing.adoc
aburdenthehand: Completely agree. Once we've nailed down the content we can use conditionals to only include references to virtual resources when used in a CNV context.

modules/cnv-disabling-fencing-on-node.adoc
aburdenthehand: Alrighty, I'll comment these two out in the assembly for now

aburdenthehand:
This PR is no longer relevant

PR #16894, opened 2019-09-26
OSDOCS-628 updating restricted network installation

kalexand-rh:
Draft for https://jira.coreos.com/browse/OSDOCS-628

@wking, PTAL?

### DIFF
--- a/modules/installation-adding-registry-pull-secret.adoc
+++ b/modules/installation-adding-registry-pull-secret.adoc
@@ -17,21 +17,30 @@ restricted network.
 
 Complete the following steps on the bastion host:
 
-. Download your `registry.redhat.io` pull secret from the
-link:https://cloud.redhat.com/openshift/install/pull-secret[Pull Secret] page on the {cloud-redhat-com} site.
-
-. Generate the base64-encoded user name and password or token for your mirror
-registry:
+. Generate the pull secret for your registry:
 +
 ----
-$ echo -n '<user_name>:<password>' | base64 -w0 <1>
-
-BGVtbYk3ZHAtqXs=
+$ podman login --authfile ~/pullsecret_config.json <local_registry_host_name>:<local_registry_host_port> <1>
 ----
-<1> For `<user_name>` and `<password>`, specify the user name and password that
-you configured for your registry.
+<1> For `<local_registry_host_name>`, specify the registry domain name
+for your mirror registry, such as `registry.example.com`. For
+`<local_registry_host_port>`, specify the port that your mirror registry uses to
+serve content.
++
+Provide your credentials for the mirror registry at the prompts.
 
-. Make a copy of your pull secret in JSON format:
+. View the pull secret that you created and record the pull secret value:
++
+----
+# cat ~/pullsecret_config.json
+
+{ "auths": { "<local_registry_host_name>:<local_registry_host_port>": { "auth": "ZHVtbXk6ZHVtbXk=" } } }
+----
+
+. Download your pull secret from the
+link:https://cloud.redhat.com/openshift/install/pull-secret[Pull Secret] page on the {cloud-redhat-com} site.
+
+. Make a copy of the {cloud-redhat-com} link:https://cloud.redhat.com/openshift/install/pull-secret[Pull Secret] that you downloaded in JSON format:
 +
 ----
 $ cat ./pull-secret.text | jq .  > <path>/<pull-secret-file><1>
@@ -64,10 +73,10 @@ The contents of the file resemble the following example:
 }
 ----
 
-. Edit the new file and add a section that describes your registry to it:
+. Edit the {cloud-redhat-com} link:https://cloud.redhat.com/openshift/install/pull-secret[Pull Secret] file and add a section that describes your registry to it:
 +
 ----
-  "auths": {
+  "auths": { <1>
 ...
     "<local_registry_host_name>:<local_registry_host_port>": { <1>
       "auth": "<credentials>", <2>
@@ -75,11 +84,7 @@ The contents of the file resemble the following example:
   },
 ...
 ----
-<1> For `bastion_host_name`, specify the registry domain name
-that you specified in your certificate, and for `<local_registry_host_port>`,
-specify the port that your mirror registry uses to serve content.
-<2> For `<credentials>`, specify the base64-encoded user name and password for
-the mirror registry that you generated.
+<1> Paste the contents of the `pullsecret_config.json` file that you created.
 +
 The file resembles the following example:
 +
--- a/modules/installation-mirror-repository.adoc
+++ b/modules/installation-mirror-repository.adoc
@@ -129,9 +129,9 @@ that you mirrored content for.
 .. Build a signature URI:
 +
 ----
-$ URI="https://mirror.openshift.com/pub/openshift-v4/signatures/openshift/release/${DIGEST/:/=}/<signature_name>" <1>
+$ URI="https://mirror.openshift.com/pub/openshift-v4/signatures/openshift/release/${DIGEST/:/=}/signature-<n>" <1>
 ----
-<1> Specify the name of the signature to store.
+<1> Specify the name of the signature to store. The signature name must be in the format of `signature-<n>`, where `<n>` is a number. If you have multiple signatures, the signature that includes the lowest number in its name is applied first, and other signatures are applied in increasing numerical order if one fails.
 
 .. Download the signature:
 +
--- a/modules/installation-mirror-repository.adoc
+++ b/modules/installation-mirror-repository.adoc
@@ -1,12 +1,22 @@
 // Module included in the following assemblies:
 //
 // * installing/installing_restricted_networks/installing-restricted-networks-preparations.adoc
+// * updating/updating-restricted-network.adoc
 
+ifeval::["{context}" == "installing-restricted-networks-preparations"]
+:restricted:
+:install:
+endif::[]
+ifeval::["{context}" == "updating-restricted-network"]
+:restricted:
+:update:
+endif::[]
+
+ifdef::install[]
 [id="installation-mirror-repository_{context}"]
 = Mirroring the {product-title} image repository
 
-Mirror the {product-title} image repository to use during cluster installation
-or upgrade.
+Mirror the {product-title} image repository to use during cluster installation.
 
 .Prerequisites
 
@@ -14,6 +24,21 @@ or upgrade.
 can access the certificate and credentials that you configured.
 * You downloaded the pull secret from the
 link:https://cloud.redhat.com/openshift/install/pull-secret[Pull Secret] page on the {cloud-redhat-com} site and modified it to include authentication to your mirror repository.
+endif::install[]
+
+ifdef::update[]
+[id="update-mirror-image-repository_{context}"]
+= Update the contents of the {product-title} image repository
+
+Update the contents of the image repository that hosts the mirrored content that
+you require for installing {product-title}. You must update the mirror registry
+to update {product-title} to a new version.
+
+.Prerequisites
+
+* You have access to the mirror registry that you used to store the images that
+you used to install {product-title}.
+endif::update[]
 
 .Procedure
 
@@ -21,7 +46,13 @@ Complete the following steps on the bastion host:
 
 . Review the
 link:https://access.redhat.com/downloads/content/290/[{product-title} downloads page]
-to determine the version of {product-title} that you want to install.
+to determine the version of {product-title} that you want to
+ifdef::install[]
+install.
+endif::install[]
+ifdef::update[]
+update to.
+endif::update[]
 
 . Set the required environment variables:
 +
@@ -34,7 +65,11 @@ $ export LOCAL_SECRET_JSON='<path_to_pull_secret>' <5>
 $ export RELEASE_NAME="ocp-release" <6>
 ----
 <1> For `<release_version>`, specify the version number of {product-title} to
-install, such as `4.2.0`.
+install, such as `4.2.1`.
+ifdef::update[]
+When you update {product-title}, you must specify a version number that is
+higher than the version that is installed.
+endif::update[]
 <2> For `<local_registry_host_name>`, specify the registry domain name for your mirror
 repository, and for `<local_registry_host_port>`, specify the port that it
 serves content on.
@@ -57,11 +92,103 @@ $ oc adm -a ${LOCAL_SECRET_JSON} release mirror \
 ----
 +
 This command pulls the release information as a digest, and its output includes
-the `imageContentSources` data that you require when you install your cluster.
+text that resembles the following sample:
 
-. Record the entire `imageContentSources` section from the output of the previous
-command. The information about your mirrors is unique to your mirrored repository, and you must add the `imageContentSources` section to the `install-config.yaml` file during installation.
+. Record the
+ifdef::install[]
+`imageContentSources`
+endif::install[]
+ifdef::update[]
+`ImageContentSourcePolicy`
+endif::update[]
+section from the output of the previous
+command. This information is required
+ifdef::install[]
+during {product-title} installation.
+endif::install[]
+ifdef::update[]
+when you update your {product-title} cluster.
+endif::update[]
 
+. Build the signature for the content that you mirrored and verify that it is
+signed by an official Red Hat key:
+.. Set the digest for the version to
+ifdef::install[]
+install:
+endif::install[]
+ifdef::update[]
+update to:
+endif::update[]
++
+----
+$ DIGEST="$(oc adm release info quay.io/${PRODUCT_REPO}/${RELEASE_NAME}:${OCP_RELEASE} | sed -n 's/Pull From: .*@//p')" <1>
+----
+<1> For `<release_version>`, specify the version number of {product-title} to
+that you mirrored content for.
+
+.. Build a signature URI:
++
+----
+$ URI="https://mirror.openshift.com/pub/openshift-v4/signatures/openshift/release/${DIGEST/:/=}/<signature_name>" <1>
+----
+<1> Specify the name of the signature to store.
+
+.. Download the signature:
++
+----
+$ wget "${URI}"
+----
+
+.. Download the keys that were used to sign the signature:
++
+----
+$ wget https://www.redhat.com/security/data/f21541eb.txt
+$ wget https://www.redhat.com/security/data/fd431d51.txt
+----
+
+.. Import the keys:
++
+----
+$ gpg --no-default-keyring --keyring ./temp.keyring --import ./f21541eb.txt
+$ gpg --no-default-keyring --keyring ./temp.keyring --import ./fd431d51.txt
+----
+
+.. Verify the signature:
++
+----
+$ gpg --no-default-keyring --keyring ./temp.keyring --verify <signature_name> <1>
+gpg: Signature made Tue 24 Sep 2019 09:38:24 AM PDT using RSA key ID F21541EB
+gpg: Good signature from "Red Hat, Inc. (beta key 2) <security@redhat.com>"
+gpg:                 aka "Mark Cox Internal RSA 4096 test key <mjc@redhat.com>"
+gpg: WARNING: This key is not certified with a trusted signature!
+gpg:          There is no indication that the signature belongs to the owner.
+Primary key fingerprint: B08B 659E E86A F623 BC90  E8DB 938A 80CA F215 41EB
+----
+<1> Specify the name of the signature that you stored.
+
+.. Confirm that the `Primary key fingerprint` value from the signature output is
+is listed on the
+link:https://access.redhat.com/security/team/key[Product Signing Keys]
+page on the Red Hat Customer Portal.
+
+. Mirror the repository:
++
+----
+$ oc adm -a ${LOCAL_SECRET_JSON} release mirror \
+     --from=quay.io/${PRODUCT_REPO}/${RELEASE_NAME}:${OCP_RELEASE} \
+     --to=${LOCAL_REGISTRY}/${LOCAL_REPOSITORY} \
+     --to-release-image=${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}:${OCP_RELEASE}
+----
++
+This command pulls the release information as a digest, and its output includes the
+ifdef::install[]
+`imageContentSources` data that you require when you install your cluster.
+endif::install[]
+ifdef::update[]
+`ImageContentSourcePolicy` data that you require when you update your cluster.
+endif::update[]
+
+ifdef::install[]
 . To create the installation program that is based on the content that you
 mirrored, extract it and pin it to the release:
 +
@@ -75,3 +202,4 @@ To ensure that you use the correct images for the version of {product-title}
 that you selected, you must extract the installation program from the mirrored
 content.
 ====
+endif::install[]
--- a/modules/installation-mirror-repository.adoc
+++ b/modules/installation-mirror-repository.adoc
@@ -131,7 +131,7 @@ that you mirrored content for.
 ----
 $ URI="https://mirror.openshift.com/pub/openshift-v4/signatures/openshift/release/${DIGEST/:/=}/signature-<n>" <1>
 ----
-<1> Specify the name of the signature to store. The signature name must be in the format of `signature-<n>`, where `<n>` is a number. If you have multiple signatures, the signature that includes the lowest number in its name is applied first, and other signatures are applied in increasing numerical order if one fails.
+<1> Specify the name of the signature to store. The signature name must be in the format of `signature-<n>`, where `<n>` is a sequential series of numbers that starts with `1`. The signature verification process checks all valid signatures by incrementing `<n>` until it finds a valid signature or the requested `signature-<n>` does not exist. 
 
 .. Download the signature:
 +
@@ -164,12 +164,14 @@ gpg: WARNING: This key is not certified with a trusted signature!
 gpg:          There is no indication that the signature belongs to the owner.
 Primary key fingerprint: B08B 659E E86A F623 BC90  E8DB 938A 80CA F215 41EB
 ----
-<1> Specify the name of the signature that you stored.
+<1> Specify the name of the signature that you stored, which is in the format `signature-<n>`.
 
 .. Confirm that the `Primary key fingerprint` value from the signature output is
 is listed on the
 link:https://access.redhat.com/security/team/key[Product Signing Keys]
 page on the Red Hat Customer Portal.
+
+.. If you cannot verify the signature, repeat this process to create another one. Name the new signature `signature-<n+1>`, where `<n>` is the number that you used in the last signature name.
 
 . Mirror the repository:
 +
--- a/modules/update-restricted-network-cli.adoc
+++ b/modules/update-restricted-network-cli.adoc
@@ -1,0 +1,164 @@
+// Module included in the following assemblies:
+//
+// * updating/updating-restricted-network.adoc
+
+[id="update-restricted-network-cli_{context}"]
+= Update the restricted network cluster
+
+After you update the contents of the mirror registry, you can update your
+restricted network cluster.
+
+.Prerequisites
+
+* You have access to the mirror registry that you used to store the images that
+you used to install {product-title}.
+* You have the `ImageContentSourcePolicy` information for the content that you
+mirrored.
+* The the OpenShift Command-line Interface (CLI), commonly known as `oc`, is
+installed.
+* The `jq` package is installed.
+
+.Procedure
+
+Complete the following steps on the bastion host:
+
+. Review the `ImageContentSourcePolicy` resource for your cluster:
++
+----
+$ oc get ImageContentSourcePolicy -o yaml
+
+apiVersion: operator.openshift.io/v1alpha1
+kind: ImageContentSourcePolicy
+metadata:
+  name: example
+spec:
+  repositoryDigestMirrors: <1>
+  - mirrors:
+    - <local_registry_host_name>:<local_registry_host_port>/<repository_name>/release
+    source: quay.io/openshift-release-dev/ocp-release
+  - mirrors:
+    - <local_registry_host_name>:<local_registry_host_port>/<repository_name>/release
+    source: quay.io/openshift-release-dev/ocp-v4.0-art-dev
+----
+<1> The `repositoryDigestMirrors` section must contain the `mirrors` data that
+is in the output of the `oc adm mirror release` command from when you mirrored
+the images for this release.
+
+. If the `repositoryDigestMirrors` section of the `ImageContentSourcePolicy` does not contain the `mirrors` for this release, create another `ImageContentSourcePolicy` resource:
+.. Record the name of the current `ImageContentSourcePolicy`. You will delete this resource after you complete the update.
+.. Define the `ImageContentSourcePolicy` resource for your cluster:
++
+----
+$ cat test.yaml
+apiVersion: operator.openshift.io/v1alpha1 <1>
+kind: ImageContentSourcePolicy
+metadata:
+  name: example
+spec:
+  repositoryDigestMirrors:
+  - mirrors:
+    - <local_registry_host_name>:<local_registry_host_port>/<repository_name>/release
+    source: quay.io/openshift-release-dev/ocp-release
+  - mirrors:
+    - <local_registry_host_name>:<local_registry_host_port>/<repository_name>/release
+    source: quay.io/openshift-release-dev/ocp-v4.0-art-dev
+----
+<1> Specify the entire specification that was shown in the output of the
+`oc adm mirror release` command that mirrored the images for this release.
+.. Create the `ImageContentSourcePolicy` resource that you defined:
++
+----
+$ oc create -f test.yaml
+----
+.. Wait a few minutes, and then confirm that the new `ImageContentSoucePolicy` is in effect by reviewing the node status:
++
+----
+$ oc describe node | grep machineconfig
+                    machineconfiguration.openshift.io/currentConfig: rendered-master-9194b32791755030dcce887c66024113
+                    machineconfiguration.openshift.io/desiredConfig: rendered-master-9194b32791755030dcce887c66024113
+                    machineconfiguration.openshift.io/reason:
+                    machineconfiguration.openshift.io/state: Done
+                    machineconfiguration.openshift.io/currentConfig: rendered-worker-a0f873210ec70fb117835a9c9527db29
+                    machineconfiguration.openshift.io/desiredConfig: rendered-worker-a0f873210ec70fb117835a9c9527db29
+                    machineconfiguration.openshift.io/state: Done
+                    machineconfiguration.openshift.io/currentConfig: rendered-master-9194b32791755030dcce887c66024113
+                    machineconfiguration.openshift.io/desiredConfig: rendered-master-9194b32791755030dcce887c66024113
+                    machineconfiguration.openshift.io/reason:
+                    machineconfiguration.openshift.io/state: Done
+                    machineconfiguration.openshift.io/currentConfig: rendered-worker-a0f873210ec70fb117835a9c9527db29
+                    machineconfiguration.openshift.io/desiredConfig: rendered-worker-a0f873210ec70fb117835a9c9527db29
+                    machineconfiguration.openshift.io/state: Done
+                    machineconfiguration.openshift.io/currentConfig: rendered-worker-a0f873210ec70fb117835a9c9527db29
+                    machineconfiguration.openshift.io/desiredConfig: rendered-worker-a0f873210ec70fb117835a9c9527db29
+                    machineconfiguration.openshift.io/state: Done
+                    machineconfiguration.openshift.io/currentConfig: rendered-master-9194b32791755030dcce887c66024113
+                    machineconfiguration.openshift.io/desiredConfig: rendered-master-9194b32791755030dcce887c66024113
+                    machineconfiguration.openshift.io/state: Done
+
+----
++
+For each machine, the `machineconfiguration.openshift.io/state` must be `Done` before you begin the update.
+
+. Start the cluster update:
++
+----
+$ oc adm upgrade --to-image <local_registry_host_name>:<local_registry_host_port>/ocp/release:<release_version> <1>
+----
+<1> For `<local_registry_host_name>` and `<local_registry_host_port>`, specify
+the values that describe your mirror registry host name and port. For
+`<release_version>`, specify the version that you want to upgrade to.
+
+. Review the cluster version status history to monitor the status of the update.
+It might take some time for all the objects to finish updating.
++
+----
+$ oc get clusterversion -o json|jq ".items[0].status.history"
+
+[
+  {
+    "completionTime": null,
+    "image": "quay.io/openshift-release-dev/ocp-release@sha256:9c5f0df8b192a0d7b46cd5f6a4da2289c155fd5302dec7954f8f06c878160b8b",
+    "startedTime": "2019-06-19T20:30:50Z",
+    "state": "Partial",
+    "verified": true,
+    "version": "4.2.1"
+  },
+  {
+    "completionTime": "2019-06-19T20:30:50Z",
+    "image": "quay.io/openshift-release-dev/ocp-release@sha256:b8307ac0f3ec4ac86c3f3b52846425205022da52c16f56ec31cbe428501001d6",
+    "startedTime": "2019-06-19T17:38:10Z",
+    "state": "Completed",
+    "verified": false,
+    "version": "4.2.0"
+  }
+]
+----
++
+The history contains a list of the most recent versions applied to the cluster.
+This value is updated when the CVO applies an update. The list is ordered by
+date, where the newest update is first in the list. Updates in the history have
+state `Completed` if the rollout completed and `Partial` if the update failed
+or did not complete.
++
+[IMPORTANT]
+====
+If an upgrade fails, the cluster version reports a condition that explains the
+failure. Contact Red Hat support.
+====
+
+. After the update completes, you can confirm that the cluster version has
+updated to the new version:
++
+----
+$ oc get clusterversion
+
+NAME      VERSION     AVAILABLE   PROGRESSING   SINCE     STATUS
+version   4.2.1       True        False         2m        Cluster version is 4.2.1
+----
+
+. Optional: Delete the original `ImageContentSourcePolicy` resource, if one existed:
++
+----
+$ oc delete ImageContentSourcePolicy <name> <1>
+----
+<1> For `<name>`, specify the name of the original `ImageContentSourcePolicy` resource.
--- a/modules/update-restricted-network-cli.adoc
+++ b/modules/update-restricted-network-cli.adoc
@@ -102,7 +102,7 @@ For each machine, the `machineconfiguration.openshift.io/state` must be `Done` b
 . Start the cluster update:
 +
 ----
-$ oc adm upgrade --to-image <local_registry_host_name>:<local_registry_host_port>/ocp/release:<release_version> <1>
+$ oc adm upgrade --allow-explicit-upgrade --to-image <local_registry_host_name>:<local_registry_host_port>/ocp/release@${DIGEST} <1>
 ----
 <1> For `<local_registry_host_name>` and `<local_registry_host_port>`, specify
 the values that describe your mirror registry host name and port. For
--- a/updating/updating-restricted-network.adoc
+++ b/updating/updating-restricted-network.adoc
@@ -1,0 +1,25 @@
+[id="updating-restricted-network"]
+= Updating a restricted network cluster
+include::modules/common-attributes.adoc[]
+:context: updating-restricted-network
+
+toc::[]
+
+You can update, or upgrade, an {product-title} cluster that you installed in a
+restricted network by using the web console.
+
+.Prerequisites
+
+* Access to the cluster as a user with `admin` privileges.
+See xref:../authentication/using-rbac.adoc[Using RBAC to define and apply permissions].
+* Access to the mirror registry that you used to store the images
+that you used to install {product-title}.
+
+include::modules/update-service-overview.adoc[leveloffset=+1]
+
+//This file has the right if statements to apply to upgrade and install.
+include::modules/installation-mirror-repository.adoc[leveloffset=+1]
+
+//You can only update through the CLI for disconnected, for now, but using the
+//web console is still the best user experience in all other instances.
+include::modules/update-restricted-network-cli.adoc[leveloffset=+1]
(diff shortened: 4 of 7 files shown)


modules/installation-mirror-repository.adoc
```diff
@@ -35,7 +67,11 @@ $ export LOCAL_SECRET_JSON='<path_to_pull_secret>' <5>
 $ export RELEASE_NAME="ocp-release" <6>
 ----
 <1> For `<release_version>`, specify the version number of {product-title} to
-install, such as `4.2.0`.
+install, such as `4.2.1`.
+ifdef::update[]
+When you update {product-title}, you must specify a version number that is
+higher than the version that is installed.
```
wking: I think we might support rollbacks from 4.y.z to 4.y.(z-1), although you should probably get someone with more authority to confirm that.  Also, I'm not sure all increases are supported.  For example, we no longer recommend upgrading to 4.1.12, so even though 4.1.11->4.1.12 is an increase, it may be an unsupported upgrade.

modules/installation-mirror-repository.adoc
```diff
@@ -76,11 +113,84 @@ imageContentSources:
...
+  name: example
+spec:
+  repositoryDigestMirrors:
+  - mirrors:
+    - <local_registry_host_name>:<local_registry_host_port>/<repository_name>/release
+    source: quay.io/openshift-release-dev/ocp-release
+  - mirrors:
+    - <local_registry_host_name>:<local_registry_host_port>/<repository_name>/release
+    source: quay.io/openshift-release-dev/ocp-v4.0-art-dev
+...
```
wking: Nit, you could have this ellipsis land outside the `endif::update[]` and be shared between the install and update cases.

modules/installation-mirror-repository.adoc
```diff
@@ -66,6 +102,7 @@ Success
 Update image:  <local_registry_host_name>:<local_registry_host_port>/<local_registry>/<local_repository>:<release_version>
 Mirror prefix: <local_registry_host_name>:<local_registry_host_port>/<local_registry>/<local_repository>
 
+ifdef::install[]
 To use the new mirrored repository to install, add the following section to the install-config.yaml:
 
 imageContentSources:
```
wking: nit, do you want a leading `----` before this literal block?

modules/installation-mirror-repository.adoc
```diff
@@ -76,11 +113,84 @@ imageContentSources:
...
+
+. Build the signature for the content that you mirrored and verify that it is
+signed by an official Red Hat key:
+.. Set the digest for the version to install to:
++
+----
+$ DIGEST="$(oc adm release info quay.io/openshift-release-dev/ocp-release:<release_version> | sed -n 's/Pull From: //p')" <1>
+----
+<1> For `<release_version>`, specify the version number of {product-title} to
+that you mirrored content for.
```
wking: You want to verify the signature before going through the trouble of mirroring, and then use a by-digest pullspec for the mirror command instead of the `<release_version>` tag.

modules/update-restricted-network-cli.adoc
```diff
@@ -0,0 +1,106 @@
+// Module included in the following assemblies:
+//
+// * updating/updating-restricted-network.adoc
+
+[id="update-restricted-network-cli_{context}"]
+= Update the restricted network cluster
+
+Verify the sig
```
wking: nit: sig -> signature?  And maybe more stuff to make this a sentence? :p

modules/update-restricted-network-cli.adoc
```diff
@@ -0,0 +1,106 @@
...
+
+.Prerequisites
+
+* You have access to the mirror registry that you used to store the images that
+you used to install {product-title}.
+* You have the `ImageContentSourcePolicy` information for the
```
wking: nit: incomplete sentence

modules/update-restricted-network-cli.adoc
```diff
@@ -0,0 +1,106 @@
...
+* The the OpenShift Command-line Interface (CLI), commonly known as `oc`, is
+installed.
+* The `jq` package is installed.
+
+.Procedure
+
+Complete the following steps on the bastion host:
+
+. Create an `ImageContentSourcePolicy` resource for your cluster:
```
wking: You should already have an ImageContentSourcePolicy.  If you mirror from the same repos to the same mirror repo, you won't need to update it.  Basically, folks should run `oc adm release mirror ...` and then merge the `ImageContentSourcePolicy` it suggests with the values that are already in their cluster, and only creating a fresh ImageContentSourcePolicy if none already exists.

modules/update-restricted-network-cli.adoc
```diff
@@ -0,0 +1,106 @@
...
+spec:
+  repositoryDigestMirrors:
+  - mirrors:
+    - <local_registry_host_name>:<local_registry_host_port>/<repository_name>/release
+    source: quay.io/openshift-release-dev/ocp-release
+  - mirrors:
+    - <local_registry_host_name>:<local_registry_host_port>/<repository_name>/release
+    source: quay.io/openshift-release-dev/ocp-v4.0-art-dev
+----
+<1> Specify the entire specification that was showin in the output of the
```
wking: nit: "showin" -> "shown in"

modules/update-restricted-network-cli.adoc
```diff
@@ -0,0 +1,106 @@
...
+----
+
+. Start the cluster update:
++
+----
+$ oc adm upgrade --force --to-image <local_registry_host_name>:<local_registry_host_port>/ocp/release:<release_version> <1>
+----
+<1> For `<local_registry_host_name>` and `<local_registry_host_port>`, specify
+the values that describe your mirror registry host name and port. For
+`<release_version>`, specify the version that you want to upgrade to.
```
wking: if you trust your mirror, this is fine.  If you aren't sure you can trust your mirror, you will want to use a by-digest pullspec here with the digest whose signature you verified earlier.

modules/update-restricted-network-cli.adoc
```diff
@@ -0,0 +1,106 @@
...
++
+The history contains a list of the most recent versions applied to the cluster.
+This value is updated when the CVO applies an update. The list is ordered by
+date, where the newest update is first in the list. Updates in the history have
+state `Completed` if the rollout completed and `Partial` if the update failed
+or did not complete.
++
+[IMPORTANT]
+====
+If an upgrade fails, the Operator stops and reports the status of the failing
```
wking: The cluster-version operator aborts that particular upgrade round, and starts over, attempting to apply it again.  Maybe just say:

> If an upgrade fails, ClusterVersion will include a condition explaining why.

modules/installation-mirror-repository.adoc
kalexand-rh: I made a new PR based on the discussion on Slack about downgrades: https://github.com/openshift/openshift-docs/pull/16914
I'm moving this info to the previous step and saying "valid version." (I don't see 4.1.12 in the product link, so if they actually follow that step, they should pick something usable.)

modules/installation-mirror-repository.adoc
kalexand-rh: Git is hiding the `----` for this codeblock.

modules/update-restricted-network-cli.adoc
kalexand-rh: @wking, when is the ImageContentSourcePolicy created? I don't see an explicit step to do it. Is the policy created when you drop the imageContentSources data into your `install-config.yaml` during installation?

If you can share the command to edit the existing policy, I'm happy to document that instead.

wking: > when is the ImageContentSourcePolicy created?

The installer creates it if you set `imageContentSources`.  Otherwise, you'd have to create it from scratch.

vikram-redhat:
@kalexand-rh ready for QE?

modules/update-restricted-network-cli.adoc
```diff
@@ -0,0 +1,129 @@
...
+is in the output of the `oc adm mirror release` command from when you mirrored
+the images for this release.
+
+. If the `repositoryDigestMirrors` section of the `ImageContentSourcePolicy`
+does not contain the `mirrors` for this release, add them to that section.
+
+. Start the cluster update:
++
+----
+$ oc adm upgrade --force --to-image <local_registry_host_name>:<local_registry_host_port>/ocp/release:<release_digest> <1>
```
jianlinliu: Do we really want release_digest instead of release_version? Per my past testing, release_digest does not work. Refer to https://jira.coreos.com/browse/CORS-1105?focusedCommentId=121819&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-121819. Since this discussion, QE always use release_version for testing, but not release_digest!

modules/update-restricted-network-cli.adoc
```diff
@@ -0,0 +1,129 @@
...
+* The `jq` package is installed.
+
+.Procedure
+
+Complete the following steps on the bastion host:
+
+. Review the `ImageContentSourcePolicy` resource for your cluster:
++
+----
+$ oc edit ImageContentSourcePolicy
```
jianlinliu: If just review, I think `oc get ImageContentSourcePolicy -o yaml` will be more save. In QE's testing, we suggest to create new ImageContentSourcePolicy, but not edit old ImageContentSourcePolicy, That would be more easier for track and management. 
There is some discussion in https://jira.coreos.com/browse/CORS-1105?focusedCommentId=115606&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-115606.

modules/update-restricted-network-cli.adoc
```diff
@@ -0,0 +1,129 @@
...
+NAME      VERSION     AVAILABLE   PROGRESSING   SINCE     STATUS
+version   4.2.1       True        False         2m        Cluster version is 4.2.1
+----
+
+. If you modified the contents of the `repositoryDigestMirrors` section of the
+`ImageContentSourcePolicy` and are no longer using the old mirrors, you can
+remove them:
++
+----
+$ oc edit ImageContentSourcePolicy
```
jianlinliu: If we are managing ImageContentSourcePolicy by creating a new object, there we should use `oc delete ImageContentSourcePolicy xxxx` to remove old ones.

kalexand-rh:
@jianlinliu, I think I've captured all of your updates. I'd rather publish what you've been testing. Will you PTAL?

modules/installation-mirror-repository.adoc
```diff
@@ -48,13 +87,51 @@ the pull secret for your mirror registry that you created.
 <6> The release mirror. For a production release, you must specify
 `ocp-release`.
 
+. Build the signature for the content that you plan to mirror and verify that it
+is signed by an official Red Hat key:
+.. Set the digest for the version to install to:
++
+----
+$ DIGEST="$(oc adm release info quay.io/${PRODUCT_REPO}/${RELEASE_NAME}:${OCP_RELEASE} | sed -n 's/Pull From: //p')" <1>
```
jianlinliu: I do not think this command would get a correct DIGEST.
`
$ oc adm release info quay.io/openshift-release-dev/ocp-release:4.2.0-rc.2 | sed -n 's/Pull From: //p'

quay.io/openshift-release-dev/ocp-release@sha256:9a1a38ce4d5fa79c785801eca3679a759bede9f40081b9e1b9f0cca2baa673a
`

The DIGEST should be 'sha256:9a1a38ce4d5fa79c785801eca3679a759bede9f40081b9e1b9f0cca2baa673a2', right?

modules/installation-mirror-repository.adoc
```diff
@@ -48,13 +87,51 @@ the pull secret for your mirror registry that you created.
...
+.. Download the signature:
++
+----
+$ wget "${URI}"
+----
+
+.. Verify the signature:
++
+----
+$ gpg --verify <signature_name> <1>
```
jianlinliu: When I run this command, I get this:
`
$ gpg --verify signature-1
gpg: Signature made Tue 08 Oct 2019 10:12:39 AM EDT using RSA key ID F21541EB
gpg: Can't check signature: No public key
`
Seem like here miss some steps.

modules/installation-mirror-repository.adoc
```diff
@@ -48,13 +87,51 @@ the pull secret for your mirror registry that you created.
...
+page on the Customer Portal.
+
 . Mirror the repository:
 +
 ----
 $ oc adm -a ${LOCAL_SECRET_JSON} release mirror \
      --from=quay.io/${PRODUCT_REPO}/${RELEASE_NAME}:${OCP_RELEASE} \
      --to=${LOCAL_REGISTRY}/${LOCAL_REPOSITORY} \
-     --to-release-image=${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}:${OCP_RELEASE}
+     --to-release-image=${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}:${DIGEST}
```
jianlinliu: In the following cluster upgrade command, we decide to use release_version, then here we should NOT mirror release payload image with DIGEST

modules/installation-mirror-repository.adoc
```diff
@@ -75,17 +153,48 @@ imageContentSources:
...
+when you update your {product-title} cluster.
+endif::update[]
 
+ifdef::install[]
 . To create the installation program that is based on the content that you
 mirrored, extract it and pin it to the release:
 +
 ----
-$ oc adm release extract --command=openshift-install "${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}:${OCP_RELEASE}"
+$ oc adm release extract --command=openshift-install "${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}:${DIGEST}"
```
jianlinliu: If we decide to use release_version, here should revert 'DIGEST' back.

modules/update-restricted-network-cli.adoc
```diff
@@ -0,0 +1,168 @@
...
+
+Complete the following steps on the bastion host:
+
+. View any existing `ImageContentSourcePolicy` resources for your cluster:
++
+----
+$ oc get ImageContentSourcePolicy -o yaml
+----
+
+. If the previous command returned an `ImageContentSourcePolicy` resource,
```
jianlinliu: At this step, upgrade is not started yet, before your cluster is totally upgraded, I do not suggest to remove the old ones. It maybe risky.

modules/installation-mirror-repository.adoc
wking: Ah, right.  Could use `s/Pull From: .*@//p` to jump straight to the digest

modules/installation-mirror-repository.adoc
wking: Ah, prob need to import at least beta2 from the release key page.

modules/installation-mirror-repository.adoc
wking: Yeah, we need a tag here to keep the mirrored release from being garbage collected.

modules/installation-mirror-repository.adoc
wking: We want `@${DIGEST}` here to guard against mirror-registry mutation.

modules/update-restricted-network-cli.adoc
wking: Yeah.  Union here.  Prune later, if you want, certainly after upgrade completes.

modules/update-restricted-network-cli.adoc
```diff
@@ -0,0 +1,168 @@
...
+    source: quay.io/openshift-release-dev/ocp-v4.0-art-dev
+----
+<1> The `repositoryDigestMirrors` section must contain the `mirrors` data that
+is in the output of the `oc adm mirror release` command from when you mirrored
+the images for this release.
+
+. If the `repositoryDigestMirrors` section of the `ImageContentSourcePolicy`
+does not contain the `mirrors` for this release, add them to that section.
+////
+. Start the cluster update:
```
jianlinliu: If new ImageContentSourcePolicy is created, we need do some more check before start the cluster update. 
The new ImageContentSourcePolicy settings will be deployed to each node and shortly start using the mirrored repository for requests to the source repository.

To check that the mirrored configuration worked, go to one of your nodes. For example:
`
$ oc get node

NAME                         STATUS ROLES AGE VERSION

ip-10-0-137-44.ec2.internal  Ready worker 7m v1.14.6+90fadebfa

ip-10-0-138-148.ec2.internal Ready          master 11m v1.14.6+90fadebfa

ip-10-0-139-122.ec2.internal Ready          master 11m v1.14.6+90fadebfa

ip-10-0-147-35.ec2.internal  Ready,SchedulingDisabled       worker 7m v1.14.6+90fadebfa

ip-10-0-153-12.ec2.internal  Ready worker 7m v1.14.6+90fadebfa

ip-10-0-154-10.ec2.internal  Ready master 11m v1.14.6+90fadebfa

`

You can see that scheduling on each worker node is disabled as the change is being applied.

If no ImageContentSourcePolicy resource is created, we could skip this check.

modules/installation-mirror-repository.adoc
wking: https://access.redhat.com/security/team/key links to https://access.redhat.com/articles/3530471 as "a manual process has been established".  Maybe just lean on that for downloading a key and verifying with that?  For releases, the signatures are so far all by the beta2 key, which is https://www.redhat.com/security/data/f21541eb.txt .

modules/update-restricted-network-cli.adoc
wking: > If new ImageContentSourcePolicy is created, we need do some more check before start the cluster update...

Ugh.  What if you miss a SchedulingDisabled?  This really needs to be represented in `status` on some object so we can use a level-check instead of SchedulingDisabled edge checks.

jianlinliu: Hmm, I am not sure what unexpected behaviour will happen when upgrade is started but new mirror registry configuration is not laid down onto node yet. Maybe upgrade will reconcile until new mirror registry configuration is totally laid down. I think waiting for new ImageContentSourcePolicy setting onto nodes before upgrade would be safer.

wking: I agree that upgrading after mirror propagation is safer.  I just wish we had something better than hoping to catch all the SchedulingDisabled as they flit past.

jianlinliu: okay. If use other way instead of SchedulingDisabled checks, I would suggest to use `oc describe node | grep machineconfig` to ensure each node is in 'Done' state, that means new ImageContentSourcePolicy mirror setting is applied to node via MCO, WDYT?

wking: > ...to ensure each node is in 'Done' state...

But presumably it was in Done before you added the new ImageContentSourcePolicy.  I don't know what sort of delay there is before you start seeing node-side effects.  And there's [no `status` on ImageContentSourcePolicy][1]...  I'll see if I can get something more clear, but checking for `Done` status on the nodes after a brief wait seems like the best we have at the moment.

[1]: https://github.com/openshift/api/blob/9525304a0adb725ab4a4a54539a1a6bf6cc343d3/operator/v1alpha1/types_image_content_source_policy.go#L9-L20

kalexand-rh:
@wking, @jianlinliu, I think I've captured all of your changes. Will you please let me know what else needs to change?

modules/update-restricted-network-cli.adoc
```diff
@@ -0,0 +1,186 @@
...
+updated to the new version:
++
+----
+$ oc get clusterversion
+
+NAME      VERSION     AVAILABLE   PROGRESSING   SINCE     STATUS
+version   4.2.1       True        False         2m        Cluster version is 4.2.1
+----
+
+. Delete the original `ImageContentSourcePolicy` resource, if one existed:
```
jianlinliu: This should be some optional operation. Not a must.

modules/update-restricted-network-cli.adoc
```diff
@@ -0,0 +1,186 @@
...
++
+----
+$ oc delete ImageContentSourcePolicy <name> <1>
+----
+<1> For `<name>`, specify the name of the original `ImageContentSourcePolicy` resource.
+
+. If you modified the contents of the `repositoryDigestMirrors` section of the
```
jianlinliu: The part is already covered by "oc delete ImageContentSourcePolicy" command in the above section, should be removed.

modules/update-restricted-network-cli.adoc
```diff
@@ -0,0 +1,186 @@
...
+                    machineconfiguration.openshift.io/state: Done
+
+----
++
+For each machine, the `machineconfiguration.openshift.io/state` must be `Done` before you begin the update.
+
+. Start the cluster update:
++
+----
+$ oc adm upgrade --force --to-image <local_registry_host_name>:<local_registry_host_port>/ocp/release:<release_version> <1>
```
jianlinliu: For a product release, I think here no need --force option.

modules/update-restricted-network-cli.adoc
```diff
@@ -0,0 +1,186 @@
...
+    source: quay.io/openshift-release-dev/ocp-v4.0-art-dev
+----
+<1> Specify the entire specification that was showin in the output of the
+`oc adm mirror release` command that mirrored the images for this release.
+.. Create the `ImageContentSourcePolicy` resource that you defined:
++
+----
+$ oc create -f test.yaml
+----
+.. Confirm that the new `ImageContentSoucePolicy` is in effect by reviewing the node status:
```
jianlinliu: How about adding some words like - "Wait for some mins, then confirm that the new `ImageContentSoucePolicy` is in effect"

modules/installation-mirror-repository.adoc
jianlinliu: I try to follow the procedure mentioned in the doc, but still failed, any idea?
`
$ wget https://www.redhat.com/security/data/f21541eb.txt
$ gpg --no-default-keyring --keyring ./temp.keyring --import ./f21541eb.txt
gpg: /root/.gnupg/trustdb.gpg: trustdb created
gpg: key F21541EB: public key "Red Hat, Inc. (beta key 2) <security@redhat.com>" imported
gpg: Total number processed: 1
gpg:               imported: 1  (RSA: 1)
$ gpg --verify signature-1
gpg: Signature made Tue 08 Oct 2019 10:12:39 AM EDT using RSA key ID F21541EB
gpg: Can't check signature: No public key
`

modules/installation-mirror-repository.adoc
```diff
@@ -48,13 +87,51 @@ the pull secret for your mirror registry that you created.
 <6> The release mirror. For a production release, you must specify
```
jianlinliu: For the following lines:
> <5> For <path_to_pull_secret>, specify the absolute path to and file name of the pull secret for your mirror registry that you created.

The file not only need the pull secret for your mirror registry that you created, but also the pull secret from the OpenShift Infrastructure Providers page. So the description here is not accurate. The description in 'Prerequisites' of disconnected install doc maybe more clear:

> You downloaded the pull secret from the OpenShift Infrastructure Providers page and modified it to include authentication to your mirror repository.

kalexand-rh: https://docs.openshift.com/container-platform/4.2/installing/installing_restricted_networks/installing-restricted-networks-preparations.html#installation-local-registry-pull-secret_installing-restricted-networks-preparations
So, should this section be replaced with this section instead? https://github.com/openshift/openshift-docs/pull/16678/files#diff-caf7c18be6cf05b5acd8148e443eb3e2R1

wking: For disconnected clusters, we'll need different authorities in the pull secret for different steps (some previous discussion [here][1]).  For mirroring from Quay to the mirror registry (which may be what this step is?), you'll need `quay.io` and your mirror token.  For launching the cluster from your mirror registry, you'll only need the mirror token.  For pushing telemetry from the cluster (only when you have network access to us, possibly via a proxy), you'll need ... `cloud.openshift.com`?  Etc.  But it may be easier to just have a secret with the mirror token and the cloud.openshift.com-sourced tokens, and use that everywhere without trying to break it down to minimal sets for each action.

[1]: https://github.com/openshift/openshift-docs/pull/16678#discussion_r325832486

jianlinliu: https://docs.openshift.com/container-platform/4.2/installing/installing_restricted_networks/installing-restricted-networks-preparations.html#installation-local-registry-pull-secret_installing-restricted-networks-preparations looks good to me, Should NOT be replaced.  Just like what wking explained, At this step, user is going to mirror release image, so need auth for both Quay and mirror registry token.

kalexand-rh: @jianlinliu, doesn't that just give you the mirror registry token? I'm proposing changing to https://osdocs628--ocpdocs.netlify.com/openshift-enterprise/latest/installing/installing_restricted_networks/installing-restricted-networks-preparations.html#installation-adding-registry-pull-secret_installing-restricted-networks-preparations, which has the same step to generate the mirror registry token and then add it to the pull secret that you get from the OpenShift Infrastructure Providers page.

modules/installation-adding-registry-pull-secret.adoc
```diff
@@ -17,22 +17,31 @@ restricted network.
...
+# cat ~/pullsecret_config.json
 
-BGVtbYk3ZHAtqXs=
+{ "auths": { "<local_registry_host_name>:<local_registry_host_port>": { "auth": "ZHVtbXk6ZHVtbXk=" } } }
 ----
-<1> For `<user_name>` and `<password>`, specify the user name and password that
-you configured for your registry.
 
-. Make a copy of your pull secret in JSON format:
+. Download your `registry.redhat.io` pull secret from the
```
jianlinliu: Why we name the pull secret `registry.redhat.io`? The pull secret including several auths for different registries, registry.redhat.io is only one of them. Personally I think this is easy to confuse user.

modules/installation-mirror-repository.adoc
jianlinliu: Oops, I missed your new change. Thanks for pointing the url to me. Most looks good to me except `registry.redhat.io` pull secret name looks like a bit weird.

modules/installation-adding-registry-pull-secret.adoc
kalexand-rh: To differentiate it from the other pull secret. Will you please take a look at the new commit and let me know if it's ok now?

jianlinliu: LGTM.

modules/installation-mirror-repository.adoc
kalexand-rh: @wking, will you PTAL?

cc: @abhinavdahiya since Trevor's not working on install work this sprint

kalexand-rh:
@jianlinliu, is this PR ready to merge?

jianlinliu:
> @jianlinliu, is this PR ready to merge?

Most looks good to me. But seem like https://github.com/openshift/openshift-docs/pull/16894#discussion_r333812008 still have no any progress yet.

modules/installation-mirror-repository.adoc
vikram-redhat: @wking @abhinavdahiya *bump*.

modules/installation-mirror-repository.adoc
```diff
@@ -48,13 +87,51 @@ the pull secret for your mirror registry that you created.
...
+page on the Customer Portal.
+
 . Mirror the repository:
 +
 ----
 $ oc adm -a ${LOCAL_SECRET_JSON} release mirror \
      --from=quay.io/${PRODUCT_REPO}/${RELEASE_NAME}:${OCP_RELEASE} \
      --to=${LOCAL_REGISTRY}/${LOCAL_REPOSITORY} \
-     --to-release-image=${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}:${OCP_RELEASE}
+     --to-release-image=${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}@${OCP_RELEASE}
```
brandisher: In this line, `@${OCP_RELEASE}` needs to be `@${DIGEST}`.  `@4.2.0` wouldn't be a valid format but `@sha256:abcdefg12345` would be.

wking: You can't set a target digest.  Why is `@4.2.0` invalid?

wking: ah, I think this need to be `:${OCP_RELEASE}` (`:` for tags).

modules/installation-mirror-repository.adoc
wking: Looks like the issue with @jianlinliu's earlier attempt was 4.2.2's signature-1 coming from release 2 instead of beta 2:

```console
$ DIGEST="$(oc adm release info quay.io/openshift-release-dev/ocp-release:4.2.2 | sed -n 's/Pull From: .*@//p')"
$ echo "${DIGEST}"
sha256:dc782b44cac3d59101904cc5da2b9d8bdb90e55a07814df50ea7a13071b0f5f0
$ URI="https://mirror.openshift.com/pub/openshift-v4/signatures/openshift/release/${DIGEST/:/=}/signature-1"
$ echo "${URI}"
https://mirror.openshift.com/pub/openshift-v4/signatures/openshift/release/sha256=dc782b44cac3d59101904cc5da2b9d8bdb90e55a07814df50ea7a13071b0f5f0/signature-1
$ wget https://www.redhat.com/security/data/fd431d51.txt
$ gpg --no-default-keyring --keyring ./temp.keyring --import fd431d51.txt
$ gpg --no-default-keyring --keyring ./temp.keyring --verify signature-1
gpg: Signature made Fri 25 Oct 2019 08:47:11 AM PDT using RSA key ID FD431D51
gpg: Good signature from "Red Hat, Inc. (release key 2) <security@redhat.com>"
gpg: WARNING: This key is not certified with a trusted signature!
gpg:          There is no indication that the signature belongs to the owner.
Primary key fingerprint: 567E 347A D004 4ADE 55BA  8A5F 199E 2F91 FD43 1D51
```

wking: So if we're going to document which keys to import, probably best to say "both f21541eb and fd431d51" to cover our bases.

jianlinliu: @kalexand-rh Great. @wking's steps works well now. pls update those steps in our doc.

modules/installation-mirror-repository.adoc
jianlinliu: Yeah. Good catch. Should `s/@/:/`

kalexand-rh:
Thank you all!

@jianlinliu, I've added the steps to download the import the key and corrected the tag. Please let me know if anything else needs to change.

jianlinliu:
> Thank you all!
> 
> @jianlinliu, I've added the steps to download the import the key and corrected the tag. Please let me know if anything else needs to change.

@kalexand-rh really close to LGTM. Only need one amendment:
Replace
`$ gpg --verify <signature_name>`
with
`$ gpg --no-default-keyring --keyring ./temp.keyring --verify <signature_name>`

modules/update-restricted-network-cli.adoc
duritong: it seems that for disconnected installations you still need `--force` at least with 4.2.0

modules/update-restricted-network-cli.adoc
```diff
@@ -0,0 +1,164 @@
...
+                    machineconfiguration.openshift.io/state: Done
+
+----
++
+For each machine, the `machineconfiguration.openshift.io/state` must be `Done` before you begin the update.
+
+. Start the cluster update:
++
+----
+$ oc adm upgrade --to-image <local_registry_host_name>:<local_registry_host_port>/ocp/release:<release_version> <1>
```
duritong: This should be

`oc adm upgrade --force --to-image <local_registry_host_name>:<local_registry_host_port>/ocp-release-dev/ocp-release:<release_version> `

kalexand-rh: @jianlinliu, will you please double-check?

kalexand-rh:

modules/update-restricted-network-cli.adoc
wking: since github.com/openshift/oc#111 (4.2 backport), folks should be using `--allow-explicit-upgrade` or `--allow-unsafe-upgrade` instead of a bare `--force`.  But the signature check should only be fatal for by-tag pullspecs like this.  We should be using `@${DIGEST}`, to make sure we're upgrading to the digest we just verified locally.  Once we do that, we may not need any forcing.  Or we may still need `--allow-explicit-upgrade`.

modules/installation-mirror-repository.adoc
```diff
@@ -57,11 +92,103 @@ $ oc adm -a ${LOCAL_SECRET_JSON} release mirror \
...
+Primary key fingerprint: B08B 659E E86A F623 BC90  E8DB 938A 80CA F215 41EB
+----
+<1> Specify the name of the signature that you stored.
 
-. Record the entire `imageContentSources` section from the output of the previous
-command. The information about your mirrors is unique to your mirrored repository, and you must add the `imageContentSources` section to the `install-config.yaml` file during installation.
+.. Confirm that the `Primary key fingerprint` value from the signature output is
+is listed on the
+link:https://access.redhat.com/security/team/key[Product Signing Keys]
+page on the Customer Portal.
```
ahardin-rh: s/Customer Portal/Red Hat Customer Portal

updating/updating-restricted-network.adoc
```diff
@@ -0,0 +1,25 @@
+[id="updating-restricted-network"]
+= Updating a restricted network cluster
+include::modules/common-attributes.adoc[]
+:context: updating-restricted-network
+
+toc::[]
+
+You can update, or upgrade, an {product-title} cluster that you installed in a
+restricted network to by using the web console.
```
ahardin-rh: s/to by/by

modules/installation-mirror-repository.adoc
```diff
@@ -57,11 +92,103 @@ $ oc adm -a ${LOCAL_SECRET_JSON} release mirror \
...
+----
+<1> For `<release_version>`, specify the version number of {product-title} to
+that you mirrored content for.
+
+.. Build a signature URI:
++
+----
+$ URI="https://mirror.openshift.com/pub/openshift-v4/signatures/openshift/release/${DIGEST/:/=}/<signature_name>" <1>
+----
+<1> Specify the name of the signature to store.
```
wking: Do we want to give folks instructions on constructing this name?  It should be `signature-1`, falling back to `signature-2`, etc. until you either get an acceptable signature ("good sig by Red Hat's f21541eb; the release is fine") or a 404 and give up ("I guess this release isn't signed by anyone I trust").

ahardin-rh:
Only 2 small suggestions from me. Looks good!

modules/update-restricted-network-cli.adoc
kalexand-rh: @wking, where else do we need to swap to @${DIGEST}?

modules/installation-mirror-repository.adoc
kalexand-rh: I think I'm missing something. Isn't this just the name of the signature that you're storing? What would you be falling back to?

modules/update-restricted-network-cli.adoc
wking: We _could_ use `@${DIGEST}` in the [`--from` argument][1].  That doesn't matter as much, because if you use the current `:${OCP_RELEASE}` and quay.io decides to MitM you (unlikely), then you're still protected by using the digest you verified here, and the cluster just fails the install with "I couldn't find that digest in your mirror, and I couldn't reach the upstream registry to find it there".  That's annoying, but not as annoying as "I'm running a compromised cluster" ;).  And tags are nice and readable, so maybe folks want to pass a by-tag pullspec to `--from` and take that slim denial-of-service risk.  I don't have a preference either way.

[1]: https://github.com/openshift/openshift-docs/pull/16894/files#diff-ae82fa130b6ca45233c1ccc96dce28ebR178

modules/installation-mirror-repository.adoc
wking: Docs for signature-lookup here and here. Yes, <signature_name> is going to be the local filename, but you can't pick any filename, it needs to follow signature-<index> to match a signature that exists on our signature mirrors.

kalexand-rh:
@wking, I've attempted to address your most recent feedback. Thank you! Will you please take another look?

modules/installation-mirror-repository.adoc
```diff
@@ -129,9 +129,9 @@ that you mirrored content for.
 .. Build a signature URI:
 +
 ----
-$ URI="https://mirror.openshift.com/pub/openshift-v4/signatures/openshift/release/${DIGEST/:/=}/<signature_name>" <1>
+$ URI="https://mirror.openshift.com/pub/openshift-v4/signatures/openshift/release/${DIGEST/:/=}/signature-<n>" <1>
 ----
-<1> Specify the name of the signature to store.
+<1> Specify the name of the signature to store. The signature name must be in the format of `signature-<n>`, where `<n>` is a number. If you have multiple signatures, the signature that includes the lowest number in its name is applied first, and other signatures are applied in increasing numerical order if one fails.
```
wking: nit: maybe:
> <1> Specify the name of the signature to store. The signature name must be in the format of `signature-<n>`, where `<n>` is a number beginning with 1. If a signature is invalid, increment `<n>` until you find a valid signature or the requested `signature-<n>` does not exist. 
Or some such.  That adds "start at 1" and drop "is applied first", because this is more about "has anyone I trust signed this image?" and not about "who did something first?".

kalexand-rh: > ... If a signature is invalid, increment `<n>` until you find a valid signature or the requested `signature-<n>` does not exist.

Where do these other certificates come from? Is this a hidden step to say "If you get to `$ gpg --no-default-keyring --keyring ./temp.keyring --verify <signature_name>` and that doesn't work, then name the next one `signature_n+1`, and it'll just run through all the `signature_n` files in that directory until it finds a functioning one?

wking: > Where do these other certificates come from?

We don't have any at the moment, but may grow some during a future key-rotation or some such.  In that case, we'd have `signature-1` by the new key and `signature-2` by the old key (order insignificant).

> Is this a hidden step...

I was hoping that these docs made it explicit, but yeah "keep going until you find a satisfactory sig or run out of options" is the intended behavior.  Hopefully we can teach `oc` to do this soon...

kalexand-rh: > > Where do these other certificates come from?
> 
> We don't have any at the moment, but may grow some during a future key-rotation or some such. In that case, we'd have `signature-1` by the new key and `signature-2` by the old key (order insignificant).

"We" meaning OCP devs or "we" meaning the person making these certs?

How and why would they have more than one, and why would they be checking multiples at this point? If I'm reading the process right, they generate a very specific certificate and then check it. Is it actually relevant that you can check multiples until you find a valid one?

> 
> > Is this a hidden step...
> 
> I was hoping that these docs made it explicit, but yeah "keep going until you find a satisfactory sig or run out of options" is the intended behavior. Hopefully we can teach `oc` to do this soon...
That's what I'm trying to figure out. I'm not seeing a point in the current flow where a user would have a failed certificate or multiple certificates. There's not an explicit step to make another one. I'm trying to figure out if I need to another step around line 173 and what it would say or if this is a cool detail that's not helpful in this flow.

wking: > ... or "we" meaning the person making these certs?

This one.

> Is it actually relevant that you can check multiples until you find a valid one?

Not at the moment, but I'd be nervous if our manual-verification docs only supported `signature-1` when all the software implementations were set up to walk available sigs.  But maybe that would be fine and we'll have `oc` tooling for this by the time it matters?

> I'm trying to figure out if I need to another step around line 173...

Yeah, if you think the wording here isn't enough, you'd want a "if the sig doesn't check out, increment `signature-<index>` , return to defining `URI`, and try again.

modules/installation-mirror-repository.adoc
```diff
@@ -57,11 +92,103 @@ $ oc adm -a ${LOCAL_SECRET_JSON} release mirror \
...
++
+----
+$ gpg --no-default-keyring --keyring ./temp.keyring --import ./f21541eb.txt
+$ gpg --no-default-keyring --keyring ./temp.keyring --import ./fd431d51.txt
+----
+
+.. Verify the signature:
++
+----
+$ gpg --no-default-keyring --keyring ./temp.keyring --verify <signature_name> <1>
```
wking: Just noticed that this is checking for a valid sig, but not checking the signed content (so a single valid sig could be applied to multiple images).  I'll get instructions for checking content when I get in.

modules/installation-mirror-repository.adoc
kalexand-rh: > Yeah, if you think the wording here isn't enough, you'd want a "if the sig doesn't check out, increment `signature-<index>` , return to defining `URI`, and try again.

Yup. I'm adding a commit with another step. (If I took your suggestion as-is, I would expect to see a bug about the note not being clear enough and having to add the step later.) Thank you for working through this with me, and please let me know if you see more changes after the update.

modules/installation-mirror-repository.adoc
wking: Oh, so what we want is `--decrypt`, not `--verify`.

```console
$ gpg --no-default-keyring --keyring ./temp.keyring --decrypt signature-1 
{"critical": {"image": {"docker-manifest-digest": "sha256:a7e97365d16d8d920fedd3684b018b780337e069deb1dd8500e866c0d6110334"}, "type": "atomic container signature", "identity": {"docker-reference": "quay.io/openshift-release-dev/ocp-release:4.1.20"}}}gpg: Signature made Tue 15 Oct 2019 02:21:18 PM PDT using RSA key ID F21541EB
gpg: Good signature from "Red Hat, Inc. (beta key 2) <security@redhat.com>"
gpg: WARNING: This key is not certified with a trusted signature!
gpg:          There is no indication that the signature belongs to the owner.
Primary key fingerprint: B08B 659E E86A F623 BC90  E8DB 938A 80CA F215 41EB
```

We want:

```console
$ gpg --no-default-keyring --keyring ./temp.keyring --decrypt signature-1 | jq -r '.critical.image["docker-manifest-digest"]'
gpg: Signature made Tue 15 Oct 2019 02:21:18 PM PDT using RSA key ID F21541EB
gpg: Good signature from "Red Hat, Inc. (beta key 2) <security@redhat.com>"
gpg: WARNING: This key is not certified with a trusted signature!
gpg:          There is no indication that the signature belongs to the owner.
Primary key fingerprint: B08B 659E E86A F623 BC90  E8DB 938A 80CA F215 41EB
sha256:a7e97365d16d8d920fedd3684b018b780337e069deb1dd8500e866c0d6110334
```

to match `${DIGEST}`.  Docs on the full structure [here][1], altough I dunno how formal we want to get.  I'll work on `oc` tooling for this next week so I'm ok cutting some corners in the docs here in the meantime, as long as we're confirming that the expected digest is in the signed content.

[1]: https://github.com/containers/image/blob/8c5dd5b6e7c0253cf2f0f30c8ecd8d5e35cf03c2/docs/containers-signature.5.md#json-data-format

mtrmac: Not even that.

This code, which resolves `${DIGEST}` from the `:${OCP_RELEASE}` tag above, and then uses `:${OCP_RELEASE}` again in `oc adm release mirror`, is **INSECURE** and **MUST NOT** be used; the registry can maliciously modify the `:${OCP_RELEASE}` tag in the meantime. (Alternatively, if you trust the registry never to do that, and to always point at legitimate OCP images, you don’t need to bother with signatures, by definition.)

mtrmac: > Docs on the full structure [here](https://github.com/containers/image/blob/8c5dd5b6e7c0253cf2f0f30c8ecd8d5e35cf03c2/docs/containers-signature.5.md#json-data-format), altough I dunno how formal we want to get. I'll work on `oc` tooling for this next week

_Please_ use `c/image/copy.Image`, to do copying with signature verification integrated, if at all possible. Feel free to ping me anytime if you need help with that.

I don’t want copy&pasted partial implementations of that format all over OCP if I can help it.

modules/update-restricted-network-cli.adoc
mtrmac: Ah, true, if `oc adm upgrade` does use `${DIGEST}`, it is probably safe enough for `oc adm mirror` not to do so. Still, I think it’s a bad habit to use a the tag that may change after verification, even if it is safe in this particular case —  that reference can easily be copy&pasted elsewhere by someone who is not aware of this review discussion and the intricacies of this particular situation.

LalatenduMohanty:
@wking @kalexand-rh Can merge this asap.  So that we can PRS for the enhancements  we are trying to do in the workflow.

shawn-hurley:
@kalexand-rh @LalatenduMohanty @sdodson wondering if we have this doc for 4.4?

LalatenduMohanty:
> @kalexand-rh @LalatenduMohanty @sdodson wondering if we have this doc for 4.4?

Yeah, I am trying to get it in for 4.4

rheinzma:
I tried the procedure in this pull request with 4.3.8 in an air gapped environment (no proxy). The node running the "oc adm mirror" command has access to both, quay.io as well as my local registry mirror from which the installation was performed.

Actually the procedure did not work, as after performing

```
oc adm upgrade --allow-explicit-upgrade --to-image ${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}@${DIGEST}
```

The version operator reported

```
NAME      VERSION   AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.3.8     True        True          41m     Unable to apply x.x.x.x:443/ocp4/openshift4@sha256:a414f6308db72f88e9d2e95018f0cc4db71c6b12b2ec0f44587488f0a16efc42: the image may not be safe to use
```

When I ran with --force  it worked:

```
oc adm upgrade --allow-explicit-upgrade --force --to-image ${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}@${DIGEST}
```

Some other notes regarding the PR: 

**Section "Update the contents of the OpenShift Container Platform image repository"**

- 2: comment 1 - "<release_version>" now needs a platform (e.g. OCP_RELEASE=4.3.8-x86_64) also
- 3+4: We mirror the registry in Step (3) as well as (6) which is redundant and a little risky, as we should FIRST verify the signature and THEN mirror the image right ? So (3) and (4) has no value in this command chain as it is repeated in step (6).
- 5b: "signature-n" needs explaination, as I had to "guess" what to do here.

**Section "Update the restricted network cluster"**

- 3: This should reuse the same variables as above, to make it easier for the user

```
oc adm upgrade --allow-explicit-upgrade --to-image ${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}@${DIGEST}
```

LalatenduMohanty:
FYI, we created a separate PR for disconnected upgrades and it is merged now . https://github.com/openshift/openshift-docs/pull/21993

kalexand-rh:
> FYI, we created a separate PR for disconnected upgrades and it is merged now . #21993

The method in this PR is obsolete, and I don't see any context that would still be helpful. Closing

PR #16895, opened 2019-09-26
other available 4.2 update changes

kalexand-rh:
@wking, does this sound right to you?

### DIFF
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -171,9 +171,11 @@ Name: Updating clusters
 Dir: updating
 Distros: openshift-origin,openshift-enterprise
 Topics:
-- Name: Updating a cluster to a minor version from the web console
+- Name: Updating a cluster between minor versions
+  File: updating-cluster-between-minor
+- Name: Updating a cluster within a minor version from the web console
   File: updating-cluster
-- Name: Updating a cluster to a minor version by using the CLI
+- Name: Updating a cluster within a minor version by using the CLI
   File: updating-cluster-cli
 - Name: Updating a cluster that includes RHEL compute machines
   File: updating-cluster-rhel-compute
--- a/modules/understanding-upgrade-channels.adoc
+++ b/modules/understanding-upgrade-channels.adoc
@@ -1,11 +1,12 @@
 // Module included in the following assemblies:
 //
-// updating/updating-cluster.adoc
-// updating/updating-cluster-cli.adoc
-// updating/updating-cluster-rhel-compute.adoc
-// updating/updating-disconnected-cluster.adoc
+// * updating/updating-cluster.adoc
+// * updating/updating-cluster-between-minor.adoc
+// * updating/updating-cluster-cli.adoc
+// * updating/updating-cluster-rhel-compute.adoc
+// * updating/updating-disconnected-cluster.adoc
 
-[id="understanding_upgrade_channels_{context}"]
+[id="understanding-upgrade-channels_{context}"]
 = Understanding {product-title} upgrade channels
 
 In {product-title} 4.1, Red Hat introduced the concept of upgrade channels for
@@ -19,6 +20,4 @@ have no impact on the version of the cluster you install; the
 `openshift-install` binary for a given patch level of {product-title} always
 installs that patch level.
 
-See link:https://access.redhat.com/articles/4495171[OpenShift 4.2 Upgrades
-phased roll out] for more information on the types of updates and upgrade
-channels.
+See link:https://access.redhat.com/articles/4495171[OpenShift 4.2 Upgrades phased roll out] for more information on the types of updates and upgrade channels.
--- a/modules/update-upgrading-cli.adoc
+++ b/modules/update-upgrading-cli.adoc
@@ -46,7 +46,7 @@ $ oc get clusterversion -o json|jq ".items[0].spec"
 +
 [IMPORTANT]
 ====
-For production clusters, you must subscribe to the `stable-4.2` channel.
+For production clusters, you must subscribe to a `stable-*` channel.
 ====
 
 . View the available updates and note the version number of the update that
--- a/modules/update-upgrading-web.adoc
+++ b/modules/update-upgrading-web.adoc
@@ -1,6 +1,14 @@
 // Module included in the following assemblies:
 //
 // * updating/updating-cluster.adoc
+// * updating/updating-cluster-between-minor.adoc
+
+ifeval::["{context}" == "updating-cluster"]
+:within:
+endif::[]
+ifeval::["{context}" == "updating-cluster-between-minor"]
+:between:
+endif::[]
 
 [id="update-upgrading-web_{context}"]
 = Updating a cluster by using the web console
@@ -8,7 +16,7 @@
 If updates are available, you can update your cluster from the web console.
 
 You can find information about available {product-title} advisories and updates
-link:https://access.redhat.com/downloads/content/290/ver=4.1/rhel---7/4.1.0/x86_64/product-errata[in the errata section]
+link:https://access.redhat.com/downloads/content/290/ver=4.2/rhel---7/4.2.0/x86_64/product-errata[in the errata section]
 of the Customer Portal.
 
 .Prerequisites
@@ -19,17 +27,42 @@ of the Customer Portal.
 
 . From the web console, click *Administration* > *Cluster Settings* and review
 the contents of the *Overview* tab.
-.. For production clusters, ensure that the *CHANNEL* is set to `stable-4.2`.
+. For production clusters, ensure that the *CHANNEL* is set to the correct channel for
+ifdef::within[]
+the version that you want to update to,
+endif::within[]
+ifdef::between[]
+your current minor version,
+endif::between[]
+such as `stable-{product-version}`.
 +
 [IMPORTANT]
 ====
-For production clusters, you must subscribe to the `stable-4.2` channel.
+For production clusters, you must subscribe to a stable-* or fast-* channel.
 ====
-.. If the *UPDATE STATUS* is not *Updates Available*, you cannot upgrade your
+** If the *UPDATE STATUS* is not *Updates Available*, you cannot upgrade your
 cluster.
-.. The *DESIRED VERSION* indicates the cluster version that your cluster is running
+** The *DESIRED VERSION* indicates the cluster version that your cluster is running
 or is updating to.
 
-. Click *Updates Available*, select a version to update to, and click *Update*.
+. Click *Updates Available*, select
+ifdef::within[]
+a version to update to,
+endif::within[]
+ifdef::between[]
+the highest available version
+endif::between[]
+and click *Update*.
 The *UPDATE STATUS* changes to `Updating`, and you can review the progress of
 the Operator upgrades on the *Cluster Operators* tab.
+
+ifdef::between[]
+. After the update completes and the Cluster Version Operator refreshes the available updates, check if more updates are available in your current channel.
++
+--
+** If updates are available, continue to perform updates in the current channel until you can no longer update.
+** If no updates are available, change the *CHANNEL* to the stable-* or fast-* channel for the next minor version, and update to the version that you want in that channel.
+--
++
+You might need to perform several intermediate updates until you reach the version that you want.
+endif::between[]
--- a/updating/updating-cluster-between-minor.adoc
+++ b/updating/updating-cluster-between-minor.adoc
@@ -1,0 +1,24 @@
+[id="updating-cluster-between-minor"]
+= Updating a cluster between minor versions
+include::modules/common-attributes.adoc[]
+:context: updating-cluster-between-minor
+
+toc::[]
+
+You can update, or upgrade, an {product-title} cluster between minor versions.
+
+[NOTE]
+====
+Because of the difficulty of changing update channels by using `oc`, use the web console to change the update channel. It is recommended to complete the update process within the web console. You can follow the steps in xref:../updating/updating-cluster-cli.adoc#updating-cluster-cli[Updating a cluster within a minor version by using the CLI] to complete the update after you change to a {product-version} channel.
+====
+
+.Prerequisites
+
+* Access to the cluster as a user with `admin` privileges.
+See xref:../authentication/using-rbac.adoc[Using RBAC to define and apply permissions].
+
+include::modules/update-service-overview.adoc[leveloffset=+1]
+
+include::modules/understanding-upgrade-channels.adoc[leveloffset=+1]
+
+include::modules/update-upgrading-web.adoc[leveloffset=+1]
--- a/updating/updating-cluster-cli.adoc
+++ b/updating/updating-cluster-cli.adoc
@@ -1,12 +1,11 @@
 [id="updating-cluster-cli"]
-= Updating a cluster to a minor version by using the CLI
+= Updating a cluster within a minor version by using the CLI
 include::modules/common-attributes.adoc[]
 :context: updating-cluster-cli
 
 toc::[]
 
-You can update, or upgrade, an {product-title} cluster by using the
-OpenShift CLI (`oc`).
+You can update, or upgrade, an {product-title} cluster within a minor version by using the OpenShift CLI (`oc`).
 
 include::modules/update-service-overview.adoc[leveloffset=+1]
 
--- a/modules/update-service-overview.adoc
+++ b/modules/update-service-overview.adoc
@@ -33,6 +33,11 @@ architectures as well as other component packages. After the pipeline confirms
 the suitability of a release, the {product-title} update service notifies you
 that it is available.
 
+[IMPORTANT]
+====
+Because the update service displays all valid updates, you must not force an update to a version that the update service does not display.
+====
+
 ////
 The interaction between the registry and the {product-title} update service is different during
 bootstrap and continuous update modes. When you bootstrap the initial
--- a/updating/updating-cluster.adoc
+++ b/updating/updating-cluster.adoc
@@ -1,12 +1,11 @@
 [id="updating-cluster"]
-= Updating a cluster to a minor version from the web console
+= Updating a cluster within a minor version from the web console
 include::modules/common-attributes.adoc[]
 :context: updating-cluster
 
 toc::[]
 
-You can update, or upgrade, an {product-title} cluster to a minor version by
-using the web console.
+You can update, or upgrade, an {product-title} cluster by using the web console.
 
 .Prerequisites
 


modules/update-upgrading-cli.adoc
```diff
@@ -46,7 +46,7 @@ $ oc get clusterversion -o json|jq ".items[0].spec"
 +
 [IMPORTANT]
 ====
-For production clusters, you must subscribe to the `stable-4.1` channel.
+For production clusters, you must subscribe to a `stable` channel.
```
wking: `stable-*`?  Or otherwise make explicit that `stable-` is the required prefix.

modules/update-upgrading-web.adoc
```diff
@@ -5,7 +5,8 @@
 [id="update-upgrading-web_{context}"]
 = Updating a cluster by using the web console
 
-If updates are available, you can update your cluster from the web console.
+If updates are available, you can update your cluster from the web console. You
+can perform both major and minor updates from the web console.
```
wking: Major (e.g. 4.9.10->5.0.0), minor (e.g. 4.1.9->4.2.3), and patch (e.g. 4.1.9->4.1.11)?

modules/update-upgrading-web.adoc
```diff
@@ -19,11 +20,12 @@ of the Customer Portal.
 
 . From the web console, click *Administration* > *Cluster Settings* and review
 the contents of the *Overview* tab.
-.. For production clusters, ensure that the *CHANNEL* is set to `stable-4.1`.
+.. For production clusters, ensure that the *CHANNEL* is set to the correct
+`stable-<major_version>` channel for the version that you want to update to.
```
wking: Is `{major.minor_version}` a thing?  Would be nice if you didn't have to bump this manually, but could get it automatically expanded to match whatever flavor the docs were being built for.

modules/update-upgrading-web.adoc
wking: And in case we are disagreeing on terms, my major.minor.patch terms are from https://semver.org/spec/v2.0.0.html#summary

updating/updating-cluster-cli.adoc
```diff
@@ -8,6 +8,11 @@ toc::[]
 You can update, or upgrade, an {product-title} cluster by using the
 OpenShift CLI (`oc`).
 
+[IMPORTANT]
+====
+You can update your cluster to only a different minor version by using `oc`.
```
wking: You can tell your cluster to upgrade to whatever version using `oc adm upgrade ...`.  But telling it to take upgrades that aren't in our official Cincinnati graph (e.g. https://api.openshift.com/api/upgrades_info/v1/graph?channel=stable-4.1 ) may be unsupported.  I'm not clear on the official support policy in this space; maybe @crawford knows?

crawford: It's unsupported. We really need to teach the CVO to do local validation as well so it's less easy to make that mistake.

wking: > We really need to teach the CVO to do local validation as well so it's less easy to make that mistake.

The CVO may already refuse non graph upgrades without `--force`.  But I was previously not clear on whether that override voided support for the cluster.  Even more reason to track edge removal in Git ;).

modules/update-upgrading-web.adoc
kalexand-rh: Your terminology is better than mine.

modules/update-upgrading-web.adoc
kalexand-rh: It's not. I think that it's a good idea though, so I'm going to bring it through our internal change process.

updating/updating-cluster-cli.adoc
kalexand-rh: I'll clarify that you can only update within the current minor version, but I'm not going to mention that they can make an unsupported update by using `--force`.

wking: > I'll clarify that you can only update within the current minor version...

Until we start recommending 4.1->4.2 upgrades.

> ... but I'm not going to mention that they can make an unsupported update by using `--force`.

But disconnected folks will have to do this, because the CVO cannot pull their signatures to validate them.

On the other hand, without `--force` it should be impossible to make an unsupported upgrade.

kalexand-rh: > > I'll clarify that you can only update within the current minor version...
> 
> Until we start recommending 4.1->4.2 upgrades.

Clayton said that the best user experience was to upgrade 4.1 > 4.2 updates from the web console, so that's the only path I'm going to document.
> 
> > ... but I'm not going to mention that they can make an unsupported update by using `--force`.
> 
> But disconnected folks will have to do this, because the CVO cannot pull their signatures to validate them.
> 
> On the other hand, without `--force` it should be impossible to make an unsupported upgrade.

The disconnected update path will have its own separate topic after all my in-flight PRs merge.

updating/updating-cluster-cli.adoc
```diff
@@ -10,7 +10,8 @@ OpenShift CLI (`oc`).
 
 [IMPORTANT]
 ====
-You can update your cluster to only a different minor version by using `oc`.
+You can update your cluster to only a patch version within your current minor
+version by using `oc`.
```
wking: Without `--force`, you have the same choices as in the web UI, and you can see those choices with `oc adm upgrade`.  From `oc adm upgrade --help`:

> If no arguments are passed the command will retrieve the current version info and display whether an upgrade is in progress or whether any errors might prevent an upgrade, as well as show the suggested updates available to the cluster.

kalexand-rh:
@jiajliu, will you PTAL?

kalexand-rh:
@wking, will you PTAL at the latest commit? I addressed some of your feedback in Slack and added a topic about updating between minor versions.

modules/update-upgrading-web.adoc
```diff
@@ -1,17 +1,22 @@
...
 
-If updates are available, you can update your cluster from the web console. You
-can update from one minor version to another, for example from version 4.1.17 to
-version 4.2.0, or to a patch within the minor version, for example from version
-4.1.16 to version 4.1.17.
+If updates are available, you can update your cluster from the web console.
 
 You can find information about available {product-title} advisories and updates
-link:https://access.redhat.com/downloads/content/290/ver=4.1/rhel---7/4.1.0/x86_64/product-errata[in the errata section]
+link:https://access.redhat.com/downloads/content/290/ver=4.1/rhel---7/4.2.0/x86_64/product-errata[in the errata section]
```
wking: I dunno about `rhel---7`; when I visit https://access.redhat.com/downloads/content/290/ver=4.1/, I'm redirected to https://access.redhat.com/downloads/content/290/ver=4.1/rhel---8/4.1.18/x86_64/product-software .

Also, I'm not wild about pinning to 4.2.0, will that only show errata fixed by 4.2.0->4.2.1?  Or will it show 4.2.0->4.2.{latest}?

And if we do pin to something in 4.2.z, you'll want `ver=4.2`, not `ver=4.1`.

modules/update-upgrading-web.adoc
```diff
@@ -22,18 +27,39 @@ of the Customer Portal.
...
+** The *DESIRED VERSION* indicates the cluster version that your cluster is running
 or is updating to.
 
-. Click *Updates Available*, select a version to update to, and click *Update*.
+. Click *Updates Available*, select
+ifdef::within[]
+a version to update to,
+endif::within[]
+ifdef::between[]
+the highest available version
```
wking: One hop does not guarantee you get to a good jumping-off point.  For example, the current stable-4.1 graph is:

```console
$ curl -sH 'Accept:application/json' 'https://api.openshift.com/api/upgrades_info/v1/graph?channel=stable-4.1' | ~/src/openshift/cincinnati/hack/graph.sh | dot -Tpng >graph.png
```

using [`graph.sh`][1] to generate:

graph

With 4.1.18 as a jumping-off point (as we're currently planning), folks on 4.2.0 would have to go 4.2.0->4.2.14->4.1.18 or similar.  Anyone still running the guarded 4.1.10 would have to go 4.1.10->4.1.11->4.1.18 or similar.  We probably want to say "and keep upgrading until there are no recommended upgrades left" somewhere for this use-case.

[1]: https://github.com/openshift/cincinnati/blob/master/hack/graph.sh

modules/update-upgrading-web.adoc
```diff
@@ -22,18 +27,39 @@ of the Customer Portal.
...
+endif::between[]
+and click *Update*.
+The *UPDATE STATUS* changes to `Updating`, and you can review the progress of
+the Operator upgrades on the *Cluster Operators* tab.
+
+ifdef::between[]
+. After the update completes, change the *CHANNEL* to the `stable-*` channel for the next minor version.
```
wking: "...and wait until the cluster-version operator refreshes the upgrade graph."?  It would be nice if we were watching the ClusterVersion object and triggering a fresh fetch when the channel changed (and maybe we are), but even then, it might take some time to pull the graph, or the network could be down, or whatever.  I think it's worth saying something to hint that into the reader's conciousness even if we don't take a big dive into all possible complications ;).

modules/update-upgrading-web.adoc
```diff
@@ -22,18 +27,39 @@ of the Customer Portal.
...
+
+. Click *Updates Available*, select the version that you want to update to, and click *Update*.
 The *UPDATE STATUS* changes to `Updating`, and you can review the progress of
 the Operator upgrades on the *Cluster Operators* tab.
```
wking: This sentence is not between-specific.  Maybe it's best to leave the between-minor docs out of this module so it can stay focused on executing a single upgrade, and then have a different module or leading section on between-minors that says "to upgrade between minors, upgrade to the most recent release in your current channel by repeating {the upgrade proceedure} until you no longer have any available updates.  Then change your channel to the new `stable-*` and perform the available update."  Or something...

modules/update-upgrading-web.adoc
kalexand-rh: The errata page shows all of the changes for the minor release.

modules/update-upgrading-web.adoc
kalexand-rh: Is there a better way to get that graph and see what updates are available? "and keep upgrading until there are no recommended upgrades left" isn't the best user experience.

wking: Not at the moment, but maybe in the mid-term.  Eventual UX will be "turn on auto-upgrades, set your target channel, and we'll route you through the graph".

kalexand-rh: When that's about to happen, please let me know.

updating/updating-cluster-cli.adoc
kalexand-rh: Did Clayton change his mind about the web UI being the best user experience for updates between versions? I thought that you couldn't change the channel.

modules/update-upgrading-web.adoc
kalexand-rh: How's this?

updating/updating-cluster-cli.adoc
wking: > Did Clayton change his mind about the web UI being the best user experience for updates between versions?

I doubt it ;).

> I thought that you couldn't change the channel.

You can `oc patch ...` like the examples [here][1].

[1]: https://github.com/openshift/openshift-docs/pull/16678#discussion_r326876856

kalexand-rh:
@jiajliu, will you PTAL?

_topic_map.yml
```diff
@@ -166,9 +166,11 @@ Name: Updating clusters
 Dir: updating
 Distros: openshift-origin,openshift-enterprise
 Topics:
-- Name: Updating a cluster to a minor version from the web console
+- Name: Updating a cluster between minor versions
+  File: updating-cluster-between-minor
```
jiajliu: i think this definition about minor version and within a minor version need to be unified.
in v4.1(https://docs.openshift.com/container-platform/4.1/updating/updating-cluster.html), minor version upgrade means 4.1.0-4.1.2,
but in this pr, seems minor version upgrade refer to 4.1-4.2.

jiajliu:
> @jiajliu, will you PTAL?

@kalexand-rh sorry for missing your msg, just add two comments in the pr.

updating/updating-cluster-between-minor.adoc
```diff
@@ -0,0 +1,22 @@
...
+:context: updating-cluster-between-minor
+
+toc::[]
+
+You can update, or upgrade, an {product-title} cluster between minor versions.
+
+[IMPORTANT]
+====
+You must update your cluster to a new minor version by using the web console.
+====
```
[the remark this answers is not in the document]
kalexand-rh: I've gotten conflicting messages about whether or not it's required to use `--force` from the CLI. After update between 4.1 and 4.2 is working as intended, I'm happy to revisit.

There's an ongoing discussion about how the update from 4.1.z to 4.2 is supposed to work on the sbr-shift-list, and I've asked Eric to let me know when there's a plan to move forward.

_topic_map.yml
kalexand-rh: I'll open a separate PR to fix that language in 4.1.

updating/updating-cluster-between-minor.adoc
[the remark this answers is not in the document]
sferich888: We don't want customers forcing upgrades, we want them using the graphs we provide. 
That said we need to document to customers how to use the force option, and clearly explain that its not a recommended decision to do this.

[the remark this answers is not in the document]
sferich888: Customers can update from cli or web console, the web console (or our Cincinnati graph are the ways you know what acceptable upgrades are).

updating/updating-cluster-cli.adoc
sferich888: The UI is the best user experience because the cli doesn't have the tools to list/show you acceptable versions (maybe we should fix that)?

updating/updating-cluster-between-minor.adoc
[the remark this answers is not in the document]
kalexand-rh: I'm happy to add some conceptual material and warnings about the existence of the --force option and emphasize the need to use one of the supported options in the graph. We're only adding steps that we actually want users to take. (If I add steps about how to use the --force option, you know people are going to follow them.)

updating/updating-cluster-cli.adoc
kalexand-rh: I think that might be a good RFE. Along with the ability to easily retrieve and update the channels in the graph.

wking: > The UI is the best user experience because the cli doesn't have the tools to list/show you acceptable versions...

`oc adm upgrade` should list recommend targets from the graph (via ClusterVersion.status`).

kalexand-rh: > [kalexand@localhost ~]$ oc adm upgrade
Cluster version is 4.1.20

> No updates available. You may force an upgrade to a specific release image, but doing so may not be supported and result in downtime or data loss.

When is update going to be working as intended?

wking: > When is update going to be working as intended?

No recommended upgrades from 4.1.20 is appropriate now.  What targets where you expecting to see?  You can test the `oc` functionality by running against 4.1.18, which should show 4.1.20 if you are in stable-4.1 or prerelease-4.1.

kalexand-rh: Ah, then we're still stuck because the question's about updating between minor versions not within a minor version.

updating/updating-cluster-between-minor.adoc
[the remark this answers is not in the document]
kalexand-rh: > Can i understand this "user can only upgrade to a minor version through web console"? why not cli supported in minor version upgrade?
> and for a 4.1.z version, how user can upgrade the cluster to 4.2? i ask this because i did not see any available update for both 4.1.18 and 4.1.20 when set channel to stable-4.2?

@jiajliu, Trevor and I talked about this for awhile. It's because you need to change channel from the `stable-4.1` to a 4.2 channel to upgrade, and there's not a good way to do that in the CLI. Instead of changing channel in the UI and then swapping to the CLI to finish, we're just staying in the UI.

_topic_map.yml
kalexand-rh: Here's the 4.1 only PR to change these titles: https://github.com/openshift/openshift-docs/pull/17631

kalexand-rh:
@jiajliu, will you PTAL? Engineering wants this change to go live on Tuesday, and I'd like to address any of your concerns on Monday, if possible.

modules/update-upgrading-web.adoc
```diff
@@ -19,17 +27,41 @@ of the Customer Portal.
...
+endif::between[]
+and click *Update*.
 The *UPDATE STATUS* changes to `Updating`, and you can review the progress of
 the Operator upgrades on the *Cluster Operators* tab.
+
+ifdef::between[]
+. After the update completes and the Cluster Version Operator refreshes the available updates, check if more updates are available in your current channel.
+--
+** If updates are available, continue to perform updates in the current channel until you can no longer update.
+** If no updates are available, change the *CHANNEL* to the `stable-*` or `fast-*` channel for the next minor version, and update to the version that you want in that channel.
```
vikram-redhat: @kalexand-rh - the build is failing here because it is getting confused with the multiple * on the same line (it's an old known bug).

Suggest overriding our guidelines to not have backticks for 'stable-* or fast-*' bit.

updating/updating-cluster-between-minor.adoc
```diff
@@ -0,0 +1,24 @@
...
+include::modules/common-attributes.adoc[]
+:context: updating-cluster-between-minor
+
+toc::[]
+
+You can update, or upgrade, an {product-title} cluster between minor versions.
+
+[NOTE]
+====
+Because of the difficulty of changing update channels by using `oc`, you use the web console to change the update channel. It is recommended to complete the update process within the web console. You can follow the steps in xref:../updating/updating-cluster-cli.adoc[Updating a cluster within a minor version by using the CLI] to complete the update after you change to a {product-version} channel.
```
vikram-redhat: @kalexand-rh missing xref.

kalexand-rh:
@jiajliu, will you PTAL? I think that this content will be valid with tomorrow's errata release, and I want to get it live as soon as possible.

kalexand-rh:

modules/understanding-upgrade-channels.adoc
```diff
@@ -1,11 +1,12 @@
...
-// updating/updating-cluster-rhel-compute.adoc
-// updating/updating-disconnected-cluster.adoc
+// * updating/updating-cluster.adoc
+// * updating/updating-cluster-between-minor.adoc
+// * updating/updating-cluster-cli.adoc
+// * updating/updating-cluster-rhel-compute.adoc
+// * updating/updating-disconnected-cluster.adoc
 
-[id="understanding_upgrade_channels_{context}"]
+[id="understanding-upgrade-channels_{context}"]
```
bergerhoffer: Not critical, just wondering whether we try to avoid changing anchor links in a version after it's been released.

modules/update-upgrading-web.adoc
```diff
@@ -19,17 +27,41 @@ of the Customer Portal.
...
 the contents of the *Overview* tab.
-.. For production clusters, ensure that the *CHANNEL* is set to `stable-4.2`.
+. For production clusters, ensure that the *CHANNEL* is set to the correct channel for
+ifdef::within[]
+the version that you want to update to,
+endif::within[]
+ifdef::between[]
+your current minor version,
+endif::between[]
+such as `stable-{product-version}`
```
bergerhoffer: Missing a period.

modules/update-upgrading-web.adoc
```diff
@@ -19,17 +27,41 @@ of the Customer Portal.
...
+ifdef::between[]
+the highest available version
+endif::between[]
+and click *Update*.
 The *UPDATE STATUS* changes to `Updating`, and you can review the progress of
 the Operator upgrades on the *Cluster Operators* tab.
+
+ifdef::between[]
+. After the update completes and the Cluster Version Operator refreshes the available updates, check if more updates are available in your current channel.
+--
```
bergerhoffer: I'm not sure what these dashes are for, but the bullets below are incorrect (dark and not indented), and the "+" above the last sentence in the module is rendering and not indenting with step 4.

updating/updating-cluster-between-minor.adoc
```diff
@@ -0,0 +1,24 @@
...
+include::modules/common-attributes.adoc[]
+:context: updating-cluster-between-minor
+
+toc::[]
+
+You can update, or upgrade, an {product-title} cluster between minor versions.
+
+[NOTE]
+====
+Because of the difficulty of changing update channels by using `oc`, you use the web console to change the update channel. It is recommended to complete the update process within the web console. You can follow the steps in xref:../updating/updating-cluster-cli.adoc#updating-cluster-cli[Updating a cluster within a minor version by using the CLI] to complete the update after you change to a {product-version} channel.
```
bergerhoffer: "you use the web console" sounds weird to me. I would drop "you" and just say "use the web console"

bergerhoffer (commented):
Just a few minor things

modules/understanding-upgrade-channels.adoc
kalexand-rh: It's an error that I regret not catching during peer review. There was a lot of drama in the mod docs repo over how much using malformed anchor IDS can break automation. There are also no other links to this content in the collection.

updating/updating-cluster-between-minor.adoc
kalexand-rh: I think there was a "must" in there at one point. Thanks!

mffiedler:
/lgtm
@jiajliu I reviewed this and tested the steps and approved it.   If you have any further comments you can put them here or open a new bz.

mffiedler (approved):

kalexand-rh:
Thanks! Merging.

PR #16905, opened 2019-09-27
Make 'make' work with python3

stlaz:
Python 2 packages are getting removed from Fedora so this script would stop working soon (hopefully, python2 should die already).

### DIFF
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -22,1164 +22,1117 @@
 # topic groups and topics on the main page.
 
 ---
-Name: Welcome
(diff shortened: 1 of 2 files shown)


vikram-redhat:
@stlaz we are probably going to be doing away with the _update_rest_api script soon for the generation of the REST API guide. This PR doesn't add anything that is immediately required but puts the burden of verifying the changes to the topic_map on us, the work for which I am not sure I can allocate currently, so I am closing it. If this is causing you or your team issues, please let me know why offline.

PR #16906, opened 2019-09-27
Update on "Updating and deleting SCCs"

stlaz:
SCCs should not be edited with the intention to add access to
users/serviceaccounts. Instead, RBAC should be used.

### DIFF
--- a/modules/security-context-constraints-command-reference.adoc
+++ b/modules/security-context-constraints-command-reference.adoc
@@ -96,20 +96,26 @@ the default SCCs.
 [id="deleting-security-context-constraints_{context}"]
 == Deleting an SCC
 
+[WARNING]
+====
+Avoid updates and removals of the default SCCs. These are recreated on a
+cluster restart.
+====
+
 To delete an SCC:
 
 ----
 $ oc delete scc <scc_name>
 ----
 
-[NOTE]
-====
-If you delete a default SCC, it will regenerate when you restart the cluster.
-====
 
 [id="updating-security-context-constraints_{context}"]
-
 == Updating an SCC
+[WARNING]
+====
+Avoid updates and removals of the default SCCs. These are recreated on a
+cluster restart.
+====
 
 To update an existing SCC:
 
@@ -119,8 +125,8 @@ $ oc edit scc <scc_name>
 
 [NOTE]
 ====
-To preserve customized SCCs during upgrades, do not edit settings on
-the default SCCs.
-//other than priority, users, groups, labels, and annotations.
+To allow access to use an SCC, do not update any of its fields. Instead, refer to
+xref:../authentication/managing-security-context-constraints.html#role-based-access-to-ssc_configuring-internal-oauth[Role-based access to Security Context Constraints]
 ====
+
 endif::openshift-enterprise,openshift-origin[]


modules/security-context-constraints-command-reference.adoc
```diff
@@ -93,24 +93,21 @@ the default SCCs.
...
 ----
 $ oc delete scc <scc_name>
 ----
 
-[NOTE]
-====
-If you delete a default SCC, it will regenerate when you restart the cluster.
-====
-
-[id="updating-security-context-constraints_{context}"]
```
kalexand-rh: Restore this subsection.

modules/security-context-constraints-command-reference.adoc
```diff
@@ -93,24 +93,21 @@ the default SCCs.
 //other than priority, users, groups, labels, and annotations.
 ====
 
-[id="deleting-security-context-constraints_{context}"]
-== Deleting an SCC
+[id="updating-deleting-security-context-constraints_{context}"]
+== Updating and deleting an SCC
```
kalexand-rh: This should still be separate sections.

modules/security-context-constraints-command-reference.adoc
```diff
@@ -93,24 +93,21 @@ the default SCCs.
...
+
+[WARNING]
+====
+Please avoid updates and removals of the default SCCs. These will be regenerated
```
kalexand-rh:
```suggestion
Do not remove the default SCCs. They are recreated
```

If you don't want them to do it, then you need to say so. "Please" is not allowed by our style guide.

modules/security-context-constraints-command-reference.adoc
```diff
@@ -93,24 +93,21 @@ the default SCCs.
...
 
-[NOTE]
-====
-If you delete a default SCC, it will regenerate when you restart the cluster.
-====
-
-[id="updating-security-context-constraints_{context}"]
-
-== Updating an SCC
-
```
kalexand-rh: Add another note here.

kalexand-rh (commented):
@stlaz, is this change valid for 4.1 and 4.2?

stlaz:
@kalexand-rh thank you for your review, hopefully all comments were addressed.

This change is valid for all 4.x versions

modules/security-context-constraints-command-reference.adoc
```diff
@@ -119,8 +125,8 @@ $ oc edit scc <scc_name>
 
 [NOTE]
 ====
-To preserve customized SCCs during upgrades, do not edit settings on
-the default SCCs.
-//other than priority, users, groups, labels, and annotations.
+To allow access to use an SCC, do not update any of its fields. Instead, refer to
+xref:../authentication/managing-security-context-constraints.html#role-based-access-to-ssc_configuring-internal-oauth[Role-based access to Security Context Constraints]
```
kalexand-rh: You can't have an xref in a module.

stlaz: I was wondering about that (even though it worked when I tried but I had no idea whether what I am doing is right or not).

How do you refer to other subsections, then?

stlaz: @kalexand-rh is there a way to refer to other subsections, then? The above seem to have worked, but I understand why it shouldn't be allowed from the logical perspective.

kalexand-rh: @stlaz, it does render correctly, but it only does so in the context of this particular assembly. The modules need to be able to stand alone without links to any other content so that we can use any of them in any other context that's required. The "no xrefs in modules" ban is CCS-wide, and it's a ban that's resisted many, many arguments against them. I will be happy when it's lifted.

The best mitigation strategy is to add an xref from one assembly to another, but that won't work in this case. @huffmanca's on vacation, but I'm going to check with him about his original intent with this assembly and see what the best strategy is.

stlaz: @kalexand-rh thanks for the heads-up, did you and @huffmanca reach any conclusion about how to do this properly? Is it possible to do now?

vikram-redhat:
@bergerhoffer - are you able to work with @stlaz to see if it is possible to either move this or close if no longer relevant?

stlaz:
/close
let's just close it for now

PR #16907, opened 2019-09-27
travis build machinery - move to python3

stlaz:
I have no idea what it is doing but no functionality should have been changed.

Got rid of few antipatterns on the way - unused code, unused code in comments, module-level executed code, and
```
reload(sys)
sys.setdefaultencoding('utf8')
```
is icky - python3 will take care of that.


stlaz:
There's a bug in the unmaintained `aura` package which tries to split `bytes` by `string`. It would be good to move away from that package for the time being.

vikram-redhat:
I have no way to verify that the updated aura code will produce content that can be verified to be compatible with ccutil, our build publication system for customer portal docs. Thank you for submission, but I am going to reject it.

stlaz:
I don’t believe you checked the change at all. Also, why don’t you just submit the script’s output to your tooling to verify?

ernestask:
It really would just be spending some more of your time to verify the validity of the changes - it’s what maintainers do every day. Closing just because you are unwilling to do that is rather sad.

PR #16909, opened 2019-09-27
clarify the state of db replication support

bparees:


bparees:
/assign @bmcelvee 
i don't think we have this doc in 4.x, but if we do it should be fixed there also..... this can be cherry-picked back as far as we want, and the okd docs should also be updated (not branch sure what those are built off of these days?)

bparees:
@bmcelvee open for merge?

bmcelvee:
@bparees yes, thanks. I'll go ahead and merge. Should this also be applied to master-3 (if so, I can cherrypick from this PR)?

bparees:
> Should this also be applied to master-3 (if so, I can cherrypick from this PR)?

is master-3 the 3.x docs?  if so then yes...  this change is applicable to all of 4.x and 3.x.

bmcelvee:
Yeah, master-3 is the 3.x master branch. I'll go ahead and cherrypick to master-3 and assess how best to apply the changes to the rest of 3.x.

PR #16910, opened 2019-09-27
adding Azure name restrictions

kalexand-rh:
Per @wking

### DIFF
--- a/modules/installation-initializing.adoc
+++ b/modules/installation-initializing.adoc
@@ -112,11 +112,23 @@ and compute nodes.
 sub-domains of this base and will also include the cluster name.
 endif::osp[]
 ... Enter a descriptive name for your cluster.
+ifdef::azure[]
++
+[IMPORTANT]
+====
+All Azure resources that are available through public endpoints are subject to
+resource name restrictions, and you cannot create resources that use certain
+terms. For a list of terms that Azure restricts, see
+link:https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-reserved-resource-name[Resolve reserved resource name errors]
+in the Azure documentation.
+====
+endif::azure[]
 ifdef::gcp[]
 If you provide a name that is longer
 than 6 characters, only the first 6 characters will be used in the infrastructure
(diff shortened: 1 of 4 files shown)


wking (approved):
Looks good to me.  We might want to call this out (again?) closer to the install-config section where users are picking names.  And maybe near route-adding docs or other things that lead to public resources with this restriction.  But as long as we mention this once somewhere, I'm happy.

kalexand-rh:
@wking, I'll drop it again near the install config. If you tell me which routes you mean, I can add it there, too.

wking:
[These routes][1].

[1]: https://github.com/openshift/openshift-docs/tree/ca69ca64e9b84e24e2514e2d401ba44f6e2af453/networking/routes

kalexand-rh:
I added a note to the "Secured routes" topic and both Azure installation topics.

wking (approved):

kalexand-rh:

modules/installation-initializing.adoc
```diff
@@ -112,6 +112,17 @@ and compute nodes.
 sub-domains of this base and will also include the cluster name.
 endif::osp[]
 ... Enter a descriptive name for your cluster.
+ifdef::azure[]
++
+[IMPORTANT]
+====
+All Azure resources that are available through public endpoints are subject to
+resource name restrictions, and you cannot create resources that use certain
```
mburke5678: Can you remove the comma? I think it would read better without.
s/restrictions, and /restrictions and 

I like Azure's wording: _cannot use reserved words or trademarks in the name_.

All Azure resources that are available through public endpoints cannot use reserved words or trademarks in the name.

kalexand-rh: No, removing the comma's grammatically incorrect.

I am wary of removing the user from the note. I think Azure's working makes more sense in their context: presented after the user's encountered the error and in a place where it makes sense to imply that all of the trademarks are Microsoft trademarks.

kalexand-rh:
@e-tienne, PTAL?

PR #16916, opened 2019-09-27
Bug 1738476 Add using samples content to restricted network install

bmcelvee:
https://bugzilla.redhat.com/show_bug.cgi?id=1738476

### DIFF
--- a/modules/installation-restricted-network-samples.adoc
+++ b/modules/installation-restricted-network-samples.adoc
@@ -1,0 +1,74 @@
+// Module included in the following assemblies:
+//
+// * installing/installing_restricted_networks/installing-restricted-networks-preparations.adoc
+
+[id="installation-restricted-network-samples_{context}"]
+= Using sample imagestreams in a restricted network installation
+
+Most imagestreams in the OpenShift namespace managed by the Samples Operator
+point to images located in the Red Hat registry at link:registry.redhat.io[registry.redhat.io]. Mirroring
+will not apply to these imagestreams.
+
+The `jenkins`, `jenkins-agent-maven`, and `jenkins-agent-nodejs` imagestreams do
+come from the install payload and are managed by the Samples Operator, so no
+further mirroring procedures are needed for those imagestreams.
+
+[NOTE]
+====
+The `cli`, `installer`, `must-gather`, and `tests` imagestreams, while
+part of the install payload, are not managed by the Samples Operator. These are
+not addressed in this procedure.
+====
+
+.Prerequisites
+* Access to the cluster as a user with the `cluster-admin` role.
+* Create a pull secret for your mirror registry.
+
+.Procedure
+
+. Mirror images from link:registry.redhat.io[registry.redhat.io] associated with any imagestreams you need
+in the restricted network environment into one of the defined mirrors:
++
+----
+$ oc image mirror myregistry.com/myimage:latest myregistry.com/myimage:stable
+----
++
+. Add the required trusted CAs for the mirror in the cluster’s image
+configuration object:
++
+----
+$ oc create configmap registry-config --from-file=$path/ca.crt -n openshift-config
+$ oc patch image.config.openshift.io/cluster --patch '{"spec":{"additionalTrustedCA":{"name":"registry-config"}}}' --type=merge
+----
++
+//(meaning https://github.com/openshift/api/blob/master/config/v1/types_image.go#L46 … add ref to associated doc)
++
+. Update the `samplesRegistry` field in the Samples Operator configuration object
+to contain the `hostname` portion of the mirror location defined in the mirror
+configuration:
++
+----
+$ oc get configs.samples.operator.openshift.io -n openshift-cluster-samples-operator
+----
++
+[NOTE]
+====
+This is required because the imagestream import process does not use the mirror or search mechanism at this time.
+====
++
+. Add any imagestreams that are not mirrored into the `skippedImagestreams` field
+of the Samples Operator configuration object. Or if you do not want to support
+any of the sample imagestreams, set the Samples Operator to `Removed` in the
+Samples Operator configuration object.
++
+[NOTE]
+====
+Any unmirrored imagestreams that are not skipped, or if the Samples Operator is
+not changed to `Removed`, will result in the Samples Operator reporting a
+`Degraded` status two hours after the imagestream imports start failing.
+====
++
+Many of the templates in the OpenShift namespace
+reference the imagestreams. So using `Removed` to purge both the imagestreams
+and templates will eliminate the possibility of attempts to use them if they
+are not functional because of any missing imagestreams.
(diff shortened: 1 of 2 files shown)


modules/installation-restricted-network-samples.adoc
```diff
@@ -0,0 +1,58 @@
...
+not addressed in this procedure. 
+====
+
+.Prerequisites
+* Cluster administrator privileges
+
+.Procedure
+
+. Mirror images from registry.redhat.io associated with any imagestreams you need
+in the restricted network environment into one of the defined CRI-O mirrors.
```
bmcelvee: I'm not actually sure this is documented anywhere. I will need to track it down, and if not we will need to add the procedure.

modules/installation-restricted-network-samples.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+. Add the required trusted CAs for the CRI-O mirror in the cluster’s image
+configuration object.
+
+//(meaning https://github.com/openshift/api/blob/master/config/v1/types_image.go#L46 … add ref to associated doc)
+
+. Update the `samplesRegistry` field in the Samples Operator configuration object
```
xiuwang: Need keep paragraph indent different with last level paragraph

modules/installation-restricted-network-samples.adoc
```diff
@@ -0,0 +1,58 @@
...
+.Prerequisites
+* Cluster administrator privileges
+
+.Procedure
+
+. Mirror images from registry.redhat.io associated with any imagestreams you need
+in the restricted network environment into one of the defined CRI-O mirrors.
+
+. Add the required trusted CAs for the CRI-O mirror in the cluster’s image
+configuration object.
```
xiuwang: Better add cmd as example:
`oc create configmap  registry-config --from-file=$path/ca.crt  -n openshift-config`
`oc patch image.config.openshift.io/cluster --patch '{"spec":{"additionalTrustedCA":{"name":"registry-config"}}}' --type=merge`

modules/installation-restricted-network-samples.adoc
```diff
@@ -0,0 +1,63 @@
+// Module included in the following assemblies:
+//
+// * installing/installing_restricted_networks/installing-restricted-networks-preparations.adoc
+
+[id="installation-restricted-network-samples_{context}"]
+= Using sample imagestreams in a restricted network installation
+
+Most imagestreams in the OpenShift namespace managed by the Samples Operator
+point to images located in the Red Hat registry at link:registry.redhat.io. Mirroring
```
bmcelvee: Need to fix link.

modules/installation-restricted-network-samples.adoc
bmcelvee: @gabemontero is the CRI-O mirror concept the same as creating a mirror registry?

https://docs.openshift.com/container-platform/4.2/installing/installing_restricted_networks/installing-restricted-networks-preparations.html#installation-creating-mirror-registry_installing-restricted-networks-preparations

gabemontero: certainly move off of `CRI-O mirrors` to whatever is the more acceptable terminology as you see fit

modules/installation-restricted-network-samples.adoc
```diff
@@ -0,0 +1,62 @@
...
+.Prerequisites
+* Cluster administrator privileges
+
+.Procedure
+
+. Mirror images from link:registry.redhat.io[registry.redhat.io] associated with any imagestreams you need
+in the restricted network environment into one of the defined mirrors.
++
+. Add the required trusted CAs for the mirror in the cluster’s image
+configuration object:
```
gabemontero: @bmcelvee since the original google doc .... in addition to the CAs. the mirrored registry quite possibly will have required auth associated with it ... we ran into the same thing with the builds in disconnected environment support.

With that, I believe the user will need to add a pull secret to the openshift namespace with the mirrored registries credentials.  That pull secret should be automatically leveraged when the OCM initiates image imports for the image streams they have mirrored.

@bparees @mtrmac FYI ^^ in case I've misinterpreted something.

bmcelvee: Will this be sufficient to cover the pull secret requirement?

https://bz1738476--ocpdocs.netlify.com/openshift-enterprise/latest/installing/installing_restricted_networks/installing-restricted-networks-preparations.html#installation-local-registry-pull-secret_installing-restricted-networks-preparations

It's a module included in the assembly that this module is going into.

bparees: I would expect the user to provide a pullsecret to the install, that can pull from their mirror (because all the infra pods are going to have to pull their image from there).

So the pull-secret the samples operator sees and copies to `openshift` should already have the necessary creds for this.

bparees: (but certainly if for some reason they mirror the sample content to a different registry than what they mirrored the OCP images themselves to, they would need to setup the additional secret in the `openshift` namespace)

gabemontero: If they put the registry.redhat.io content in the same mirror as the install payload, I would agree @bparees   

Is there not the notion of creating more than one mirror registry (I'm asking more than speaking from knowledge)?  ... separate from the mirror used for the install payload ... I see `Copy image to multiple locations` from `oc help image mirror`.

bparees: > Is there not the notion of creating more than one mirror registry

yeah, you've anticipated my second comment above.

bparees: @gabemontero @bmcelvee this ties back to my "pull secrets docs" email.  If we had good documentation on how to setup pullsecrets for use by the samples operator, you could just point there and say "make sure the samples operator has creds to import from this mirror registry (which it may already thanks to the pull-secret you used at install-time)"

bparees: https://jira.coreos.com/browse/OSDOCS-715

bmcelvee: I checked in with Gabe, and the "Creating a pull secret for your mirror registry" procedure in the same assembly that this procedure is being included in is sufficient (at least for now). I also updated the prerequisites for this procedure accordingly. 

For OSDOCS-715, I bumped it up in priority so I can get to it sooner.

modules/installation-restricted-network-samples.adoc
bmcelvee: It's building correctly now, thanks!

modules/installation-restricted-network-samples.adoc
bmcelvee: Added, thanks!

bmcelvee:
@openshift/team-documentation please peer review (this one isn't terribly large). Thanks!

bmcelvee:
@xiuwang I know you've already taken a look at this PR, but would you mind giving it one more review? Thanks so much!

bobfuru:
LGTM

PR #16944, opened 2019-09-30
SREP-1943 included accessing the registry for dedicated

sheriff-rh:
JIRA: 
https://jira.coreos.com/browse/SREP-1943

Preview build: 
http://file.rdu.redhat.com/~antaylor/093019/srep-1943/srep-1943/registry/accessing-the-registry.html

@jewzaam please take a look.

### DIFF
--- a/modules/registry-accessing-directly.adoc
+++ b/modules/registry-accessing-directly.adoc
@@ -11,6 +11,7 @@ You can access the registry from inside the cluster.
 
 Access the registry from the cluster by using internal routes:
 
+ifdef::openshift-enterprise,openshift-origin[]
 . Access the node by getting the node's address:
 +
 ----
@@ -18,12 +19,21 @@ $ oc get nodes
 $ oc debug nodes/<node_address>
 ----
 +
+endif::[]
+
 . Log in to the container image registry by using your access token:
 +
+ifdef::openshift-enterprise,openshift-origin[]
 ----
 $ oc login -u kubeadmin -p <password_from_install_log>
 $ podman login -u kubeadmin -p $(oc whoami -t) image-registry.openshift-image-registry.svc:5000
 ----
+endif::[]
+ifdef::openshift-dedicated[]
+----
+$ podman login -u $(oc whoami) -p $(oc whoami -t) $(oc -n openshift-image-registry get route default-route -o jsonpath='{.spec.host}')
+----
+endif::[]
 +
 You should see a message confirming login, such as:
 +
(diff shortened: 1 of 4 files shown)


modules/registry-accessing-directly.adoc
```diff
@@ -80,9 +90,17 @@ $ podman pull name.io/image
...
 ----
 $ podman tag name.io/image image-registry.openshift-image-registry.svc:5000/openshift/image
 ----
+endif::[
+ifdef::openshift-dedicated[]
+----
+$ podman tag name.io/image $(oc whoami -t) $(oc -n openshift-image-registry get route default-route -o jsonpath='{.spec.host}')/openshift/image
+----
+endif::[]
+]
```
jewzaam:
```suggestion
```

modules/registry-accessing-directly.adoc
```diff
@@ -80,9 +90,17 @@ $ podman pull name.io/image
 The project name must appear in this pull specification for {product-title} to
 correctly place and later access the image in the registry:
 +
+ifdef::openshift-enterprise,openshift-origin[]
 ----
 $ podman tag name.io/image image-registry.openshift-image-registry.svc:5000/openshift/image
 ----
+endif::[
```
jewzaam:
```suggestion
endif::[]
```

jewzaam (changes requested):
Push section is missing in the rendered docs, I expect because of the brackets in the wrong place.  See suggestions, hopefully that fixes it.

modules/registry-accessing-directly.adoc
sheriff-rh: Thanks @jewzaam I couldn't figure out what was holding it up. Technically the build should have failed for that.... :P

I updated the preview build as well: 
http://file.rdu.redhat.com/~antaylor/093019/srep-1943/srep-1943/registry/accessing-the-registry.html

jewzaam (changes requested):
In reviewing the rest of the doc, noticed things that dedicated-admins cannot do:
- Viewing registry logs: can't access logs for pods in core namespaces
- Accessing registry metrics: cannot directly access the metrics as documented

On the metrics, it looks like these are not being pulled into prometheus.  This should be the way customers access the data.  The existing documentation isn't clear on what secret to pull data from.  When I try to hit the endpoint with curl from inside the pod I get no data back.  When I try to hit it via the route we've created (and are using) it times out.

I will open a bug for these issues and tag you.  For now we should remove these sections from the dedicated-admins documentation.

Side note, we don't currently permit push for dedicated-admins, not sure what is blocking that.  Will open a JIRA.  We can leave the docs as you've updated and circle back if we find we can't allow push for some reason.

jewzaam:
Related: https://bugzilla.redhat.com/show_bug.cgi?id=1757866

sheriff-rh:
Reviewed final changes with @wgordon17  and am merging at this time. Will pick to 4.3 and prior.

PR #16945, opened 2019-10-01
CNV - closing context 

aburdenthehand:
When the VM/VM template content was merged with conditionals, I didn't close the context tags in the assemblies. This fixes that error in three assemblies. 

Cherrypick to enterprise-4.1 and enterprise-4.2


kalexand-rh:
If this works for you, it LGTM. Merging.

PR #16410, opened 2019-08-27
Added 4.2 release notes

ahardin-rh:
Preview Build: http://file.rdu.redhat.com/~ahardin/09232019/OCP-4-2-release-notes/release_notes/ocp-4-2-release-notes.html

PR #16963, opened 2019-10-01
Describe configuring an internal load balancer for cloud

jboxman:
- https://github.com/openshift/openshift-docs/pull/16410#discussion_r327634216

Describe how to configure the Ingress Controller to use an internal load balancer on cloud providers.

### DIFF
--- a/modules/nw-ingress-setting-internal-lb.adoc
+++ b/modules/nw-ingress-setting-internal-lb.adoc
@@ -1,0 +1,66 @@
+// Module included in the following assemblies:
+//
+// * networking/ingress-operator.adoc
+
+[id="nw-ingress-setting-internal-lb_{context}"]
+= Configuring an Ingress Controller to use an internal load balancer
+
+When creating an Ingress Controller on cloud platforms, the Ingress Controller is published by a public cloud load balancer by default.
+As an administrator, you can create an Ingress Controller that uses an internal cloud load balancer.
+
+You can configure the `default` Ingress Controller for your cluster to be internal by deleting and recreating it.
+
+[WARNING]
+====
+If your cloud provider is Azure, you must have at least one public load balancer that points to your nodes.
+If you do not, all of your nodes will lose egress connectivity to the Internet.
+====
+
+[IMPORTANT]
+====
+If you want to change the `scope` for an `IngressController` object, you must delete and then recreate that `IngressController` object. You cannot change the `.spec.endpointPublishingStrategy.loadBalancer.scope` parameter after the Custom Resource (CR) is created.
+====
+
+See the link:https://kubernetes.io/docs/concepts/services-networking/service/#internal-load-balancer[Kubernetes Services documentation]
+for implementation details.
+
+.Prerequisites
+
+* Install the OpenShift Command-line Interface (CLI), commonly known as `oc`.
+* Log in as a user with `cluster-admin` privileges.
+
+.Procedure
+
+. Create an `IngressController` Custom Resource (CR) in a file named `<name>-ingress-controller.yaml`, such as in the following example:
++
+[source,yaml]
+----
+apiVersion: operator.openshift.io/v1
+kind: IngressController
+metadata:
+  namespace: openshift-ingress-operator
+  name: <name> <1>
+spec:
+  domain: <domain> <2>
+  endpointPublishingStrategy:
+    type: LoadBalancerService
+    loadBalancer:
+      scope: Internal <3>
+----
+<1> Replace `<name>` with a name for the `IngressController` object.
+<2> Specify the `domain` for the application published by the controller.
+If the `name` for the controller is `default` and you do not specify the `domain` parameter, the default cluster domain is used.
+<3> Specify a value of `Internal` to use an internal load balancer.
+
+. Create the Ingress Controller defined in the previous step by running the following command:
++
+----
+$ oc create -f <name>-ingress-controller.yaml <1>
+----
+<1> Replace `<name>` with the name of the `IngressController` object.
+
+. Optional: Confirm that the Ingress Controller was created by running the following command:
++
+----
+$ oc --all-namespaces=true get ingresscontrollers
+----
--- a/modules/nw-ingress-sharding-namespace-labels.adoc
+++ b/modules/nw-ingress-sharding-namespace-labels.adoc
@@ -4,7 +4,7 @@
 // * ingress-operator.adoc
 
 [id="nw-ingress-sharding-namespace-labels_{context}"]
-= Configuring ingress controller sharding by using namespace labels
+= Configuring Ingress Controller sharding by using namespace labels
 
 Ingress Controller sharding by using namespace labels means that the Ingress
 Controller serves any route in any namespace that is selected by the namespace
--- a/modules/nw-ingress-sharding-route-labels.adoc
+++ b/modules/nw-ingress-sharding-route-labels.adoc
@@ -1,10 +1,10 @@
 // Module included in the following assemblies:
 //
 // * configuring_ingress_cluster_traffic/configuring-ingress-cluster-traffic-ingress-controller.adoc
-// * ingress-operator.adoc
+// * networking/ingress-operator.adoc
 
 [id="nw-ingress-sharding-route-labels_{context}"]
-= Configuring ingress controller sharding by using route labels
+= Configuring Ingress Controller sharding by using route labels
 
 Ingress Controller sharding by using route labels means that the Ingress
 Controller serves any route in any namespace that is selected by the route
--- a/networking/ingress-operator.adoc
+++ b/networking/ingress-operator.adoc
@@ -7,7 +7,7 @@ toc::[]
 
 The Ingress Operator implements the `ingresscontroller` API and is the
 component responsible for enabling external access to {product-title}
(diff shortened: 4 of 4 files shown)


modules/nw-ingress-setting-internal-lb.adoc
```diff
@@ -0,0 +1,30 @@
...
+  namespace: openshift-ingress-operator
+  name: internal
+spec:
+  endpointPublishingStrategy:
+    type: LoadBalancerService
+    loadBalancer:
+      scope: Internal
+----
+
+See the link:https://kubernetes.io/docs/concepts/services-networking/#internal-load-balancer.[Kubernetes Services documentation]
```
Miciah: The link has a spurious "." in it.  Also, the documentation has apparently moved and is now at the following URL: https://kubernetes.io/docs/concepts/services-networking/service/#internal-load-balancer

jboxman:
@Miciah, thanks. This was more of a point of a departure than a finished PR. I have a few other items to complete before I can revisit this and finalize it for a review.

jboxman:
@Miciah, I'm not yet familiar with the Ingress Operator or controllers. What does "_the IngressController is published by a public cloud load balancer by default_" mean? What does it mean to get published?

Thanks.

Miciah:
> What does "_the IngressController is published by a public cloud load balancer by default_" mean? What does it mean to get published?

It means "made available" or "exposed"; "published" doesn't have a technical meaning beyond that one that we give it in the API definition (https://godoc.org/github.com/openshift/api/operator/v1#EndpointPublishingStrategy).  To rephrase the statement that you quoted, the Ingress Operator will create a cloud load balancer through which clients can connect to the IngressController unless the IngressController specifies a different endpoint publishing strategy.

modules/nw-ingress-setting-internal-lb.adoc
```diff
@@ -0,0 +1,57 @@
...
+for implementation details.
+
+.Prerequisites
+
+* Install the OpenShift Command-line Interface (CLI), commonly known as `oc`.
+* Log in as a user with `cluster-admin` privileges.
+
+.Procedure
+
+. Create an `IngressController` CR in a file named `<name>-ingress-controller.yaml`, such as in the following example:
```
lamek: Do we have to spell out "CR" the first time we introduce it?

lamek (commented):
One small question, otherwise LGTM.

modules/nw-ingress-setting-internal-lb.adoc
jboxman: @lamek yeah, good call!

jboxman:
@Miciah, I've updated this; How does it look? Thanks!

modules/nw-ingress-setting-internal-lb.adoc
```diff
@@ -0,0 +1,57 @@
+// Module included in the following assemblies:
+//
+// * networking/ingress-operator.adoc
+
+[id="nw-ingress-setting-internal-lb_{context}"]
+= Configuring an Ingress Controller to use an internal load balancer
+
+When creating an Ingress Controller on cloud platforms, the Ingress Controller is published by a public cloud load balancer by default.
+As an administrator, you can create an Ingress Controller that uses internal cloud load balancers.
```
Miciah: "internal cloud load balancers" → "an internal cloud load balancer" (one Ingress Controller has one cloud load balancer).

Miciah (commented):
I suggested one minor correction, but otherwise looks good.

modules/nw-ingress-setting-internal-lb.adoc
```diff
@@ -0,0 +1,57 @@
...
+----
+$ oc create -f <name>-ingress-controller.yaml <1>
+----
+<1> Replace `<name>` with the name of the `IngressController` object.
+
+. Optional: Confirm that the Ingress Controller was created by running the following command:
++
+----
+$ oc --all-namespaces=true get ingresscontrollers
+----
```
kalexand-rh: @jboxman, this restriction came up on the Group G arch call:
> 4.2.0 warning on switching ingress to Internal can cause all worker nodes to lose internet connectivity. The only way to prevent that would be to have at least one public k8s service even if it doesn’t do anything 

I think that we might need a step to confirm that you have such a service.

Miciah: I'm surprised to learn there is such a restriction, and it seems like a very serious defect.  Is there an associated Bugzilla report?

kalexand-rh: @abhinavdahiya, was there a bug for the need to have at least one public k8s service if you switch the ingress controller to private?

jboxman:
@kalexand-rh, does the work you're doing supersede this PR? If so, should I close it or finalize it?

Thanks!

kalexand-rh:
@jboxman, please finalize it. I'll incorporate the content in the change that I'm working on.

modules/nw-ingress-setting-internal-lb.adoc
Miciah: It should be noted that the restriction applies specifically to Azure, not to other cloud platforms.  (It has already been noted in the Group G arch call, but I'm repeating it here for the benefit of anyone who learns of the issue from seeing this thread.)

modules/nw-ingress-setting-internal-lb.adoc
```diff
@@ -0,0 +1,57 @@
...
+. Create an `IngressController` Custom Resource (CR) in a file named `<name>-ingress-controller.yaml`, such as in the following example:
++
+[source,yaml]
+----
+apiVersion: operator.openshift.io/v1
+kind: IngressController
+metadata:
+  namespace: openshift-ingress-operator
+  name: <name> <1>
+spec:
```
ironcladlou: `.spec.domain` is required unless we're talking about the `default` resource (which can omit `.spec.domain` and claim the default domain).

modules/nw-ingress-setting-internal-lb.adoc
ironcladlou: I'm going to talk with @Miciah and @abhinavdahiya and @wking  about how we solve this transparently. I think we should treat the issue as a blocker bug that we'll solve before release and remove the note here entirely.

ironcladlou: Sounds like the installer already sets up egress for private clusters in 4.3 by installing a public LB, so there's nothing to note here after all?

abhinavdahiya: the installer only sets up egress by creating a dummy public standard load balancer, when users request Internal clusters.

so for public cluster users, that want to make their ingress private as Day-2, there should be warning that, all nodes of the cluster will loose egress to internet in case there is no public LB pointing to those nodes...

jboxman: @abhinavdahiya, is that for _any_ public cloud provider user, or just for Azure? As I'm still trying to internalize this, my first pass at this is simply the following:

> If your cloud provider is Azure, you must have at least one public load balancer that points to your nodes.
If you do not, all of your nodes will lose egress connectivity to the Internet.

But if I include this as part of the procedure, is there a way to confirm if there is at least one public load-balancer point to all nodes? Doesn't having a public load-balancer defeat the purpose of private ingress?

Thanks.

cc @ironcladlou @Miciah

jboxman: Hi @ironcladlou @Miciah 

Any thoughts on this ^^?

Thanks.

bmcelvee:
@jboxman - just in case you see this, I'm going to grab this PR first thing in the morning and get it merged. Thanks!

modules/nw-ingress-setting-internal-lb.adoc
bmcelvee: @ironcladlou I'm helping out to get this PR merged and just want to verify - this is the only update this PR needs? Thanks!

(I'll also likely have follow-up questions related to "User defined IngressControllers at installation".)

ironcladlou:
>Configuring an Ingress Controller to use an internal load balancer

>You can configure the `default` Ingress Controller for your cluster to be internal by deleting and recreating it.

Would it be clearer to represent these separately?

The procedure for _Configuring an Ingress Controller to use an internal load balancer_ is similar to what's already here, except we can remove all the notes about the default ingresscontroller, as we're talking about creating additional ingress controllers that live alongside the default. 

The procedure for _configure the `default` Ingress Controller for your cluster to be internal by deleting and recreating it_ in its most basic form could be:

```
oc replace --force --wait --filename - <<EOF
apiVersion: operator.openshift.io/v1
kind: IngressController
metadata:
  namespace: openshift-ingress-operator
  name: default
spec:
  endpointPublishingStrategy:
    type: LoadBalancerService
    loadBalancer:
      scope: Internal
EOF
```

jboxman:
@bmcelvee, thanks! It's been floating in the ether for a while now.

I believe there's similar content that is trapped in our release notes for 4.2 that has been requested to come out.

bmcelvee:
Merging this PR with the `to_followup` label and opening the follow-up PR. Will link when it's open.

bmcelvee:
Link to follow-up PR: https://github.com/openshift/openshift-docs/pull/18764.

PR #16976, opened 2019-10-01
bug 1744072 correcting UPI autoscaling

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1744072

@enxebre, will you PTAL?

### DIFF
--- a/machine_management/applying-autoscaling.adoc
+++ b/machine_management/applying-autoscaling.adoc
@@ -9,6 +9,12 @@ Applying autoscaling to an {product-title} cluster involves deploying a
 ClusterAutoscaler and then deploying MachineAutoscalers for each Machine type
 in your cluster.
 
+[IMPORTANT]
+====
+You can configure the ClusterAutoscaler only in clusters that use
+installer-provisioned infrastructure.
+====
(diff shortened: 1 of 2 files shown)


machine_management/applying-autoscaling.adoc
```diff
@@ -9,6 +9,12 @@ Applying autoscaling to an {product-title} cluster involves deploying a
 ClusterAutoscaler and then deploying MachineAutoscalers for each Machine type
 in your cluster.
 
+[IMPORTANT]
+====
+You can configure the ClusterAutoscaler only in clusters that use
```
enxebre: Is this doc generic for upi and ipi? If so may be mention you can configure  autoscaling in UPI if the machine API is operational

kalexand-rh: We don't currently document using the machineAPI to create machines in a UPI cluster. There was some disagreement about whether or not to include this method for AWS. If it's something that we want customers to use, please let me know so that I can get an issue to add that doc in our queue.

jhou1:
LGTM

PR #16986, opened 2019-10-01
bug 1756106 correcting Azure service principal commands

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1756106

@e-tienne, will you PTAL?


e-tienne:
This PR LGTM regarding BZ 1756106

Note: I have noticed that the current document linked has non standardized Azure keys: ( "tenantID" and "appID" in the document instead of "tenantID" and "appID")

PR #16989, opened 2019-10-01
Added Accessing hosts topic

ahardin-rh:
Per https://bugzilla.redhat.com/show_bug.cgi?id=1753034

Preview build: http://file.rdu.redhat.com/~ahardin/10242019/accessing-hosts-on-aws/networking/accessing-hosts.html

### DIFF
--- a/modules/accessing-hosts-on-aws.adoc
+++ b/modules/accessing-hosts-on-aws.adoc
@@ -1,0 +1,49 @@
+// Module included in the following assemblies:
+//
+// * networking/accessing-hosts.adoc
+
+[id="accessing-hosts-on-aws_{context}"]
+= Accessing hosts on Amazon Web Services (AWS) after a full-stack automation (IPI) installation
+
+The {product-title} installer does not create any public IP addresses for any of
+the Amazon Elastic Compute Cloud (Amazon EC2) instances that it provisions for
+your {product-title} cluster. In order to be able to SSH to your {product-title}
+hosts, you must follow this procedure.
+
+.Procedure
+
+. Create a security group that allows SSH access into the virtual private cloud
+(VPC) created by the `openshift-install` command.
+
+. Create an Amazon EC2 instance on one of the public subnets the installer
+created.
+
+. Associate a public IP address with the Amazon EC2 instance that you created.
++
+Unlike with the {product-title} installation, you should associate the Amazon
+EC2 instance you created with an SSH keypair. It does not matter what operating
+system you choose for this instance, as it will simply serve as an SSH bastion
+to bridge the internet into your {product-title} cluster's VPC. The Amazon
+Machine Image (AMI) you use does matter. With Red Hat Enterprise Linux CoreOS,
+for example, you can provide keys via Ignition, like the installer does.
+
+. Once you provisioned your Amazon EC2 instance and can SSH into it, you must add
+the SSH key that you associated with your {product-title} installation. This key
+can be different from the key for the bastion instance, but does not have to be.
++
+[NOTE]
+====
+Direct SSH access is only recommended for disaster recovery. When the Kubernetes
+API is responsive, run privileged pods instead.
+====
+
+. Run `oc get nodes`, inspect the output, and choose one of the nodes that is a
+master. The host name looks similar to `ip-10-0-1-163.ec2.internal`.
+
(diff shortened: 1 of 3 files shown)


ahardin-rh:
@sferich888 @wking Can you please take a look at my first pass at adding these docs in? Thanks!

sferich888:
This is specific to IPI installs; UPI installs on AWS likely have to use similar information, and that needs to be denoted somewhere. 

We also need to include similar info for the other platforms we'll be supporting.

modules/accessing-hosts-on-aws.adoc
```diff
@@ -0,0 +1,41 @@
...
+. Create a security group that allows SSH access into the virtual private cloud
+(VPC) created by the `openshift-install` command.
+
+. Create an Amazon EC2 instance inside the VPC that the installer created or on
+one of the public subnets the installer created.
+
+. Associate a public IP address with the Amazon EC2 instance that you created.
++
+Unlike with the {product-title} installation, you must associate the Amazon EC2
+instance you created with an SSH keypair. It does not matter what operating
```
wking: Depends on which AMI you use.  With RHCOS, you can provide keys via Ignition, like the installer does.  So we should at least soften the "must" or more clearly scope it to a particular AMI family.

modules/accessing-hosts-on-aws.adoc
```diff
@@ -0,0 +1,41 @@
...
+system you choose for this instance, as it will simply serve as an SSH bastion
+to bridge the internet into your {product-title} cluster's VPC.
+
+. Once you provisioned your Amazon EC2 instance and can SSH into it, you must add
+the SSH key that you associated with your {product-title} installation, not the
+key for the bastion instance.
```
wking: These could be different keys, but tgey don't have to be.

modules/accessing-hosts-on-aws.adoc
```diff
@@ -0,0 +1,41 @@
...
+
+. Run `oc get nodes`, inspect the output, and choose one of the nodes that is a
```
wking: Somewhere we want to say that direct SSH access is only recommended for disaster recover.  When the Kubernetes API is responsive, as you're assuming here, folks should run (privileged) pods instead.

ahardin-rh:
@sferich888 @wking Thanks for your feedback! I made some updates. PTAL.

Preview build: http://file.rdu.redhat.com/~ahardin/10242019/accessing-hosts-on-aws/networking/accessing-hosts.html

ahardin-rh:
@sferich888 @wking Can you please review the latest changes? http://file.rdu.redhat.com/~ahardin/10242019/accessing-hosts-on-aws/networking/accessing-hosts.html Thanks!

sheriff-rh (approved):
Looks great, Ashley!

ahardin-rh:
@rbbratta Can you please provide a QE review or point me to the correct contact who can help? Thanks!

ahardin-rh:
@lihongan Are you the correct contact to provide QE review? Thank you!

modules/accessing-hosts-on-aws.adoc
ahardin-rh: Thanks! I added a note

ahardin-rh:
@xltian Can you please help me track down a correct QE contact for this? Thank you!

xltian:
@jianlinliu  will take a look from QE.

modules/accessing-hosts-on-aws.adoc
```diff
@@ -0,0 +1,49 @@
...
+the Amazon Elastic Compute Cloud (Amazon EC2) instances that it provisions for
+your {product-title} cluster. In order to be able to SSH to your {product-title}
+hosts, you must follow this procedure.
+
+.Procedure
+
+. Create a security group that allows SSH access into the virtual private cloud
+(VPC) created by the `openshift-install` command.
+
+. Create an Amazon EC2 instance inside the VPC that the installer created or on
```
jianlinliu: If want to user can ssh login bastion host, I think `Create an Amazon EC2 instance on one of the public subnets the installer created` will be more accurate, no need other option, such as: `inside the VPC that the installer created`.

ahardin-rh: Thank you!

PR #16995, opened 2019-10-02
CNV - Temporarily commenting out VM count in telemetry

aburdenthehand:
BZ#1755915 (VM count is missing at the prometheus) is targetted for 2.1.1 z-stream. Commenting out the VM count in telemetry until it's fixed.

cherrypick to enterprise-4.2


vikram-redhat:
@aburdenthehand do you need to unset the variable cnv-cluster at the end of the module?

PR #17013, opened 2019-10-02
updating Azure machine types

kalexand-rh:
Per https://github.com/openshift/installer/blob/master/pkg/types/azure/defaults/machines.go

@e-tienne, thank you for pointing this out! Will you please confirm?

### DIFF
--- a/modules/installation-azure-limits.adoc
+++ b/modules/installation-azure-limits.adoc
@@ -25,9 +25,9 @@ ability to install and run {product-title} clusters.
 |Component |Number of components required by default| Default Azure limit |Description
 
 |vCPU
-|22
+|34
 |20 per region
-|A default cluster requires 22 vCPUs, so you must increase the account limit.
+|A default cluster requires 34 vCPUs, so you must increase the account limit.
 
 By default, each cluster creates the following instances:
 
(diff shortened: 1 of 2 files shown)


modules/installation-azure-limits.adoc
```diff
@@ -25,19 +25,20 @@ ability to install and run {product-title} clusters.
...
 * Three control plane machines
 * Three compute machines
 
-Because the bootstrap and control plane machines use `Standard_D4s_v3` virtual
-machines, which use 4 vCPUs, and the worker machines use `Standard_D2s_v3`
-virtual machines, which use 2 vCPUs, a default cluster requires 22 vCPUs.
+Because the bootstrap machine uses `Standard_D4s_v3` machines, which use 4 vCPUS,
+the control plane machines use `Standard_D8s_v3` virtual
+machines, which use 8 vCPUs, and the worker machines use `Standard_D2s_v3`
+virtual machines, which use 4 vCPUs, a default cluster requires 32 vCPUs.
```
e-tienne: For the installation to be successful, we also need to count the bootstrap vCPUs
bootstrap = 4 x 1, control = 8 x 3, workers = 4 x 2 

32 is the total vCPUS needed after a successful cluster installation, but 36 is needed to create the cluster.

kalexand-rh: I double-checked the Azure chart, and `Standard_D2s_v3` machines use 2 cores. To align it with the previous note (one bootstrap, three compute, three control machines):
bootstrap = 4 x 1 machine = 4
control = 8 x 3 machines = 24
workers = 2 x 3 machines = 6

e-tienne: Good catch. I've also put the minimum recommended number of workers instead of the default.
Thanks for fixing it!

LGTM

e-tienne:
LGTM

PR #17015, opened 2019-10-02
Bz1756837 - Remove CSI deployment example

bobfuru:
[BZ 1756837](https://bugzilla.redhat.com/show_bug.cgi?id=1756837)
The example deployment YAML file was removed in 4.2 because we don't actually support any of the drivers yet. (For reference: [PR 16594](https://github.com/openshift/openshift-docs/pull/16594)). 
I have amended BZ 1756837 to apply to release/version 4.1 and to remove CSI deployment example to align with 4.2.
@chao007 PTAL


bobfuru:
@openshift/team-documentation Please review and merge to 4.1 only

vikram-redhat:
@bobfuru instead of deleting the module, we should just remove the reference to it from the assembly.

PR #17016, opened 2019-10-02
bug 1738456 setting the control plane as unschedulable

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1738456

@wking, are you ok with the way I translated this change?

### DIFF
--- a/modules/installation-user-infra-generate-k8s-manifest-ignition.adoc
+++ b/modules/installation-user-infra-generate-k8s-manifest-ignition.adoc
@@ -1,14 +1,26 @@
 // Module included in the following assemblies:
 //
 // * installing/installing_aws_user_infra/installing-aws-user-infra.adoc
-// * installing/installing_restricted_networks/installing-restricted-networks-aws.adoc
+// * installing/installing_bare_metal/installing-bare-metal.adoc
 // * installing/installing_gcp_user_infra/installing-gcp-user-infra.adoc
+// * installing/installing_restricted_networks/installing-restricted-networks-aws.adoc
+// * installing/installing_restricted_networks/installing-restricted-networks-bare-metal.adoc
+// * installing/installing_restricted_networks/installing-restricted-networks-vsphere.adoc
+// * installing/installing_vsphere/installing-vsphere.adoc
 
-ifeval::["{context}" == "installing-restricted-networks-aws"]
-:restricted:
-endif::[]
+
 ifeval::["{context}" == "installing-aws-user-infra"]
 :aws:
+endif::[]
+ifeval::["{context}" == "installing-restricted-networks-aws"]
+:aws:
+:restricted:
+endif::[]
+ifeval::["{context}" == "installing-restricted-networks-vsphere"]
+:restricted:
+endif::[]
+ifeval::["{context}" == "installing-restricted-networks-bare-metal"]
+:restricted:
 endif::[]
 ifeval::["{context}" == "installing-gcp-user-infra"]
 :gcp:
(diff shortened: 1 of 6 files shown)


wking (approved):
Looks good to me.  Needs to go into GCP UPI docs too, but maybe we don't have those yet.

kalexand-rh:
@adellape, please pick up this change in your draft of the GCP UPI docs.

kalexand-rh:

modules/installation-generate-aws-user-infra-ignition.adoc
```diff
@@ -2,6 +2,13 @@
 //
 // * installing/installing_aws_user_infra/installing-aws-user-infra.adoc
```
bergerhoffer: Can you update this list of assemblies?

modules/installation-generate-aws-user-infra-ignition.adoc
```diff
@@ -2,6 +2,13 @@
...
 
+ifeval::["{context}" == "installing-aws-user-infra"]
```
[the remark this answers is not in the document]
kalexand-rh: Because just because a file's originally created with one use in mind doesn't mean it won't be picked up somewhere else, and with my available information, that step's only necessary for AWS.

adellape:
Just merged https://github.com/openshift/openshift-docs/pull/17043 which is likely the conflict culprit here.

modules/installation-user-infra-generate-k8s-manifest-ignition.adoc
```diff
@@ -53,47 +63,47 @@ contains the `install-config.yaml` file you created.
 Because you create your own compute machines later in the installation process,
 you can safely ignore this warning.
 
+ifdef::aws,gcp[]
```
kalexand-rh: @jianlinliu, we didn't have a step to remove these manifest files for vSphere or bare metal.

modules/installation-user-infra-generate-k8s-manifest-ignition.adoc
```diff
@@ -53,47 +63,47 @@ contains the `install-config.yaml` file you created.
...
 Because you create and manage the worker machines yourself, you do not need
 to initialize these machines.
+endif::aws,gcp[]
 
-. Optional: Currently, emptying the compute pools makes control plane machines
-schedulable. However, due to a
-link:https://github.com/kubernetes/kubernetes/issues/65618[Kubernetes limitation],
-router Pods running on control plane machines will not be reachable by the
-ingress load balancer.
+. Modify the `manifests/cluster-scheduler-02-config.yml` Kubernetes manifest file to prevent Pods from being scheduled on the control plane machines:
```
kalexand-rh: @jianlinliu, in the original GCP PR, this was marked as an optional step. Will you please double-confirm that we always need to change this setting on GCP?

kalexand-rh: https://github.com/openshift/openshift-docs/pull/17043

modules/installation-user-infra-generate-k8s-manifest-ignition.adoc
kalexand-rh: After I took another look at the files that I'd need to change to apply this step to vSphere and bare metal, I realized that it would be easier to do it on this PR. For these methods, the [4.1 steps (eg vSphere)](https://docs.openshift.com/container-platform/4.1/installing/installing_vsphere/installing-vsphere.html#installation-generate-ignition-configs_installing-vsphere) skip generating the manifests and just have you generate the Ignition configs. With this change, you need to generate the manifests too ([proposed 4.2 change](https://bz1738456--ocpdocs.netlify.com/openshift-enterprise/latest/installing/installing_vsphere/installing-vsphere.html#installation-user-infra-generate-k8s-manifest-ignition_installing-vsphere)).

The way the conditionals are set up, the 4.2 docs still won't make you remove the control plane or compute manifests.

modules/installation-user-infra-generate-k8s-manifest-ignition.adoc
jianlinliu: Per my test result, I do not think it is a optional step when we are guiding user set worker replica to 0. Now I think everywhere keep consistent now, it is a MUST.

kalexand-rh: @adellape, FYI ^

kalexand-rh:
Jianlin approved this change on the bug, so I'm going to merge. Thanks!

PR #17071, opened 2019-10-04
whitelist updates for 4.2

kalexand-rh:


cuppett (commented):
This provides a majority of cluster content. We need to figure out a way to identify registry.redhat.io and quay.io as providing the core cluster content.

modules/configuring-firewall.adoc
```diff
@@ -5,16 +5,99 @@
...
+.2+|AWS
+
+|`route53.amazonaws.com`
+|Allows the Cluster Ingress Operator to creates wildcard alias records for routes
+
+|`amazonaws.com`
+|Required to access AWS services and resources, including s3, iam, and `quay-registry.s3`
+
+|GCP
+|`www.googleapis.com/dns/v1/projects`
```
cuppett: GCP would be *.googleapis.com and we should point here: https://cloud.google.com/endpoints/

modules/configuring-firewall.adoc
```diff
@@ -5,16 +5,99 @@
...
+|cluster-ingress-operator: Creates wildcard alias records for routes.
+
+.3+|Azure
```
cuppett: For Azure it is management.azure.com and we should point here: https://docs.microsoft.com/en-us/rest/api/azure/

modules/configuring-firewall.adoc
```diff
@@ -5,16 +5,99 @@
...
++
+[cols="3,4",options="header"]
+|===
+|URL | Function
+
+|`mirror.openshift.com`
+|Required to access mirrored installation content and images
+
+|`storage.googleapis.com`
+|Something
```
cuppett: These two lines can go. They would be covered by *.googleapis.com above.

modules/configuring-firewall.adoc
```diff
@@ -5,16 +5,99 @@
...
+
+|`cloudfront.net`
```
cuppett: This should be *.cloudfront.net. It's used for edge delivery of Quay.io images. I've asked the Quay SaaS team to work to make this first party in the future.

modules/configuring-firewall.adoc
```diff
@@ -5,16 +5,99 @@
...
-api.access.redhat.com:443
-infogw.api.openshift.com:443
-----
+[cols="3,4",options="header"]
+|===
+|URL | Function
+|`registry.redhat.io`
+|Provides core container images
+
+|`quay.io`
```
cuppett: Change this to *.quay.io to allow region failover and the future cdn.quay.io (see below comments about first party image serving).

modules/configuring-firewall.adoc
```diff
@@ -5,16 +5,99 @@
...
+|`cloudfront.net`
+|Required by the Quay CDN
+
+|`https://api.openshift.com/api/upgrades_info/v1/graph`
+|Required to check if updates are available for the cluster
+
+|`art-rhcos-ci.s3.amazonaws.com`
+|Required to download {op-system-first} images
+
+|`https://api.openshift.com/`
```
cuppett: Lose the "https" in all spots to stay consistent with the URL formats.

modules/configuring-firewall.adoc
```diff
@@ -5,16 +5,99 @@
...
+|`mirror.openshift.com`
+|Required to access mirrored installation content and images
+
+|`storage.googleapis.com`
+|Something
+
+|`cloudfront.net`
+|Required by the Quay CDN
+
+|`https://api.openshift.com/api/upgrades_info/v1/graph`
```
cuppett: Lose the "https://" in all spots to stay consistent with the URL formats.

modules/configuring-firewall.adoc
```diff
@@ -5,16 +5,99 @@
...
+|Required to check if updates are available for the cluster
+
+|`art-rhcos-ci.s3.amazonaws.com`
+|Required to download {op-system-first} images
+
+|`https://api.openshift.com/`
+|Required for your cluster token
+
+|`https://cloud.redhat.com/openshift/`
```
cuppett: Lose the "https://" in all spots to stay consistent with the URL formats.

modules/configuring-firewall.adoc
```diff
@@ -5,16 +5,99 @@
...
+|===
+|Cloud |URL |Function
+
+.2+|AWS
+
+|`route53.amazonaws.com`
+|Allows the Cluster Ingress Operator to creates wildcard alias records for routes
+
+|`amazonaws.com`
+|Required to access AWS services and resources, including s3, iam, and `quay-registry.s3`
```
cuppett: For Amazon (and the other clouds), let's point to their docs and provide a general guideline. Here it would be *.amazonaws.com as one entry where they need to look up the specific endpoints to allow for their region and send them here:

https://docs.aws.amazon.com/general/latest/gr/rande.html

modules/configuring-firewall.adoc
```diff
@@ -5,16 +5,99 @@
...
+
+. Whitelist any site that provides resources for a language or framework that your builds require.
+
+. If you do not disable Telemetry, you must grant access to the following host names and ports to access Red Hat Insights:
++
+[cols="3,4",options="header"]
+|===
+|URL | Function
+
+|`cert-api.access.redhat.com:443`
```
cuppett: Lose the :443

modules/configuring-firewall.adoc
```diff
@@ -5,16 +5,99 @@
...
+|Required for telemetry
+
+|`api.access.redhat.com:443`
```
cuppett: Lose the :443

modules/configuring-firewall.adoc
```diff
@@ -5,16 +5,99 @@
...
+|Required for telemetry
+
+|`infogw.api.openshift.com:443`
```
cuppett: Lose the :443

cuppett (changes requested):
Hopefully these suggestions make sense. I'll clean up the Google doc to match.

modules/configuring-firewall.adoc
sdodson: I wonder if we should special case the quay-registry.s3.amazonaws.com because that's not specific to AWS implementation but to Quay's implementation?

cuppett: Yes. We absolutely should leave the quay-registry.s3 URL entry separate and distinct. It's included further down and you can see the net of all recommended changes in the Google Sheet.

bergerhoffer:
@kalexand-rh, I spoke with @danehans and he said that the list in step 4 is missing the instance metadata endpoint (169.254.169.254).

Docs on it for each provider are: 

* https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html
* https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service
* https://cloud.google.com/compute/docs/storing-retrieving-metadata

Can you incorporate this? @cuppett can you confirm that this is okay to add?

modules/configuring-firewall.adoc
```diff
@@ -17,7 +17,7 @@ Before you install {product-title}, you must configure your firewall to grant ac
 |`registry.redhat.io`
 |Provides core container images
 
-|`quay.io`
+|`*.quay.io`
 |Provides OLM content
```
cuppett: This needs to change to "Provides core container images" also.

modules/configuring-firewall.adoc
```diff
@@ -48,27 +48,18 @@ Before you install {product-title}, you must configure your firewall to grant ac
...
-.2+|AWS
-
-|`route53.amazonaws.com`
-|Allows the Cluster Ingress Operator to creates wildcard alias records for routes
-
-|`amazonaws.com`
-|Required to access AWS services and resources, including s3, iam, and `quay-registry.s3`
+|AWS
+|`*.amazonaws.com`
+|Required to access AWS services and resources, including s3, iam, and `quay-registry.s3`. Review the link:https://docs.aws.amazon.com/general/latest/gr/rande.html[AWS Service Endpoints] in the AWS documentation to determine the exact endpoints to allow for the regions that you use.
```
cuppett: I would lose the ", including s3, iam, and `quay-registry.s3` bit. quay-registry is called out as a separate row. Also, identifying just S3 and IAM is not all the services we use. Those are identified in the AWS portion of the docs if they want to get real specific on endpoints.

modules/configuring-firewall.adoc
```diff
@@ -48,27 +48,18 @@ Before you install {product-title}, you must configure your firewall to grant ac
...
 
 |GCP
-|`www.googleapis.com/dns/v1/projects`
-|cluster-ingress-operator: Creates wildcard alias records for routes.
-
-.3+|Azure
-|`management.azure.com/subscriptions/<subscriptionId>/resourceGroups/<resourceGroupName>/providers/Microsoft.Network/dnsZones`, where `<subscriptionId>` and `<resourceGroupName>` are the subscription ID and resource group name that you use for your cluster
-|Allows the Cluster Ingress Operator to creates wildcard alias records for routes
+|`*.googleapis.com`
+|Allows the Cluster Ingress Operator to creates wildcard alias records for routes. Review link:https://cloud.google.com/endpoints/[Cloud Endpoints] in the GCP documentation to determine the endpoints to allow for your APIs.
```
cuppett: Again, lose "Allows the Cluster Ingress Operator to creates wildcard alias records for routes." We use these endpoints for all parts of the installation and talk to many services for lots of different reasons (not just ingress).

modules/configuring-firewall.adoc
```diff
@@ -48,27 +48,18 @@ Before you install {product-title}, you must configure your firewall to grant ac
...
 
-|`management.azure.com/subscriptions/<subscriptionId>/resourceGroups/<resourceGroupName>/providers/Microsoft.Network/privateDnsZones`, where `<subscriptionId>` and `<resourceGroupName>` are the subscription ID and resource group name that you use for your cluster
-|Allows the Cluster Ingress Operator to creates wildcard alias records for routes
+|Azure
+|`management.azure.com`
+|Allows the Cluster Ingress Operator to creates wildcard alias records for routes. Review the link:https://docs.microsoft.com/en-us/rest/api/azure/[Azure REST API Reference] in the Azure documentation to determine the endpoints to allow for your APIs.
```
cuppett: Same here. Lose the ingress description.

modules/configuring-firewall.adoc
```diff
@@ -80,22 +71,20 @@ Before you install {product-title}, you must configure your firewall to grant ac
...
-|`storage.googleapis.com`
-|Something
 
-|`cloudfront.net`
-|Required by the Quay CDN
+|`*.cloudfront.net`
+|Required by the Quay CDN to provide edge delivery of Quay.io images
 
-|`https://api.openshift.com/api/upgrades_info/v1/graph`
+|`api.openshift.com/api/upgrades_info/v1/graph`
```
cuppett: Lose the end part of the URL here: "/api/upgrades_info/v1/graph"

modules/configuring-firewall.adoc
```diff
@@ -80,22 +71,20 @@ Before you install {product-title}, you must configure your firewall to grant ac
...
 |Required to check if updates are available for the cluster
 
 |`art-rhcos-ci.s3.amazonaws.com`
 |Required to download {op-system-first} images
 
-|`https://api.openshift.com/`
+|`api.openshift.com/`
```
cuppett: Lose the trailing /

modules/configuring-firewall.adoc
```diff
@@ -80,22 +71,20 @@ Before you install {product-title}, you must configure your firewall to grant ac
...
 |Required for your cluster token
 
-|`https://cloud.redhat.com/openshift/`
+|`cloud.redhat.com/openshift/`
```
cuppett: Lose the trailing /

modules/configuring-firewall.adoc
```diff
@@ -80,22 +71,20 @@ Before you install {product-title}, you must configure your firewall to grant ac
...
 |Required for your cluster token
 
 |`quay-registry.s3.amazonaws.com`
```
cuppett: You may want to move this one up beside redhat registry and *.cloudfront.net. It's the same kind of description "core container images".

cuppett (commented):
Looking good. Last couple tweaks to tighten it all up.

cuppett:
> @kalexand-rh, I spoke with @danehans and he said that the list in step 4 is missing the instance metadata endpoint (169.254.169.254).
> 
> Docs on it for each provider are:
> 
> * https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html
> * https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service
> * https://cloud.google.com/compute/docs/storing-retrieving-metadata
> 
> Can you incorporate this? @cuppett can you confirm that this is okay to add?

I'm not sure this is needed in the whitelist. These are supposed to be in the NOPROXY list by default with the cloud providers enabled. So an admin shouldn't need to know about it for the actual proxy. @danehans ?

kalexand-rh:
Thanks @cuppett! I've incorporated all of your suggestions.

danehans:
I was unaware of the context of the firewall list. I thought it was a general list and not proxy specific. @cuppett is correct, the metadata ip and gcp metadata dns name (gcp deployments only) are generated by default to `proxy.status.noProxy` for cloud deployments. Here is the current default noProxy list (AWS deploy):

```
.cluster.local,.svc,.${REGION}.compute.internal,10.0.0.0/16,10.128.0.0/14,127.0.0.1,169.254.169.254,172.30.0.0/16,api-int.${CLUSTER_NAME}.${BASE_DOMAIN},etcd-0.${CLUSTER_NAME}.${BASE_DOMAIN},etcd-1.${CLUSTER_NAME}.${BASE_DOMAIN},etcd-2.${CLUSTER_NAME}.${BASE_DOMAIN},localhost
```

The following `noProxy` values are derived from the install config:

1. `172.30.0.0/16` is from `serviceNetwork`
2. `10.0.0.0/16` is from `machineCIDR`
3. `10.128.0.0/14` is from `clusterNetwork` cidr

bergerhoffer:
@kalexand-rh Keep me posted on the status of when this can get merged as I have a proxy update to make that will depend on this content being added. Thanks!

kalexand-rh:
@jianlinliu, will you PTAL?

modules/configuring-firewall.adoc
```diff
@@ -5,16 +5,87 @@
...
+|===
+|URL | Function
+|`registry.redhat.io`
+|Provides core container images
+
+|`*.quay.io`
+|Provides core container images
+
+|`sso.redhat.com`
+|The `https://cloud.redhat.com/openshift` site uses authentication from `sso.redhat.com`
```
wking: You don't need `registry.redhat.io`, `*.quay.io` or `sso.redhat.com` for some restricted-network use cases (e.g. you've turned off telemetry and are pulling images from a mirror registry).

modules/configuring-firewall.adoc
```diff
@@ -5,16 +5,87 @@
...
+|===
+
+. Whitelist any site that provides resources for a language or framework that your builds require.
```
wking: Do we want a "such as..." example to make this more concrete.  Like some NPM registry thing for Node?

jianlinliu:
Most looks good to me. 
Should we add *.apps.<CLUSTER-NAME>.<DOMAIN> into this whitelist?

modules/configuring-firewall.adoc
kalexand-rh: > Most looks good to me.
> Should we add *.apps.. into this whitelist?

@ironcladlou, what do you think?

ironcladlou: All default routes (e.g. console, auth, prometheus) sit behind that domain. Also, keep in mind the user can configure the ingress wildcard domain prefix at install-time (but *.apps is the default). Does that help? I'm not 100% up to speed on the context here. @danehans WDYT?

jianlinliu:
Per QE's test result:
`
$ grep CONNECT access.log |awk '{print $7}' |sort | uniq
accounts.google.com:443
api.openshift.com:443
cloudresourcemanager.googleapis.com:443
d3uo42mtx6z2cr.cloudfront.net:443
iam.googleapis.com:443
oauth2.googleapis.com:443
oauth-openshift.apps.qe-gpei-42p.qe.gcp.devcluster.openshift.com:443
quay.io:443
registry.redhat.io:443
registry.svc.ci.openshift.org:443
serviceusage.googleapis.com:443
storage.googleapis.com:443
www.googleapis.com:443

`
Seem like `*.apps` and `accounts.google.com`

gpei:
I spinned up 4.2 clusters on AWS&GCP with proxy enabled, then I could see the cluster was trying to connect to the following URL in proxy access log during installation, this also could be used as a reference IMO. `accounts.google.com` and `oauth-openshift.apps.{cluster.domain}` is the two URL missing in current doc that I can see.

* On GCP
```
accounts.google.com 
api.openshift.com 
cloudresourcemanager.googleapis.com 
d3uo42mtx6z2cr.cloudfront.net 
iam.googleapis.com 
oauth2.googleapis.com 
oauth-openshift.apps.qe-gpei-42p.qe.gcp.devcluster.openshift.com 
quay.io 
registry.redhat.io 
serviceusage.googleapis.com 
storage.googleapis.com 
www.googleapis.com 
```
* On AWS
```
api.openshift.com 
d3uo42mtx6z2cr.cloudfront.net 
ec2.us-east-2.amazonaws.com 
iam.amazonaws.com 
oauth-openshift.apps.qe-gpei-420.qe.devcluster.openshift.com 
qe-gpei-420-lpt7d-image-registry-us-east-2-rrbwlywqxqbmfqrujos.s3.us-east-2.amazonaws.com 
quay.io 
registry.redhat.io 
```

kalexand-rh:
Thanks! I've added the two URLs and gotten Stephen to approve the new descriptions. I'm going to merge, and if more changes are required, I'll open a follow-up PR.

kalexand-rh:

modules/configuring-firewall.adoc
```diff
@@ -5,16 +5,93 @@
...
+
+. If you do not disable Telemetry, you must grant access to the following host names and ports to access Red Hat Insights:
```
bergerhoffer: Remove "and ports" since it doesn't call out any ports? Or just say URLs?

modules/configuring-firewall.adoc
```diff
@@ -5,16 +5,93 @@
...
++
+[cols="3,4",options="header"]
+|===
+|URL | Function
+
+|`cert-api.access.redhat.com`
+|Required for telemetry
```
bergerhoffer: telemetry -> Telemetry

x2 instances below

modules/configuring-firewall.adoc
```diff
@@ -5,16 +5,93 @@
...
+
+|`api.access.redhat.com`
+|Required for telemetry
+
+|`infogw.api.openshift.com`
+|Required for telemetry
+|===
+
+. If you use Amazon Web Services (AWS), Microsoft Azure, or Google Cloud Platform (GCP) to host your cluster, you must grant access URLs that provide the cloud provider API and DNS for that cloud:
```
bergerhoffer: access -> access to the

modules/configuring-firewall.adoc
```diff
@@ -5,16 +5,93 @@
...
 
-Before you install {product-title}, you must configure your firewall to access
-Red Hat Insights.
+Before you install {product-title}, you must configure your firewall to grant access to the sites that {product-title} requires.
 
 .Procedure
 
-* Set your firewall to allow the following host names and ports on the outgoing
-network firewall:
+. Whitelist the following registry URLS:
```
bergerhoffer: URLS -> URLs

modules/configuring-firewall.adoc
```diff
@@ -5,16 +5,93 @@
...
+|AWS
+|`*.amazonaws.com`
+|Required to access AWS services and resources. Review the link:https://docs.aws.amazon.com/general/latest/gr/rande.html[AWS Service Endpoints] in the AWS documentation to determine the exact endpoints to allow for the regions that you use.
+
+.2+|GCP
+|`*.googleapis.com`
+|Required to access GCP services and resources. Review link:https://cloud.google.com/endpoints/[Cloud Endpoints] in the GCP documentation to determine the endpoints to allow for your APIs.
+
+|`accounts.google.com`
+| Required to access your GCP account
```
bergerhoffer: Add period.

modules/configuring-firewall.adoc
```diff
@@ -5,16 +5,93 @@
...
+|URL | Function
+
+|`mirror.openshift.com`
+|Required to access mirrored installation content and images
+
+|`*.cloudfront.net`
+|Required by the Quay CDN to provide edge delivery of Quay.io images
+
+|`*.apps`
+|Required to access the default cluster routes unless you set an ingress wildcard during installation.
```
bergerhoffer: This has a period but the others in the table don't. Be consistent either with or without. And maybe look at being consistent across tables in this module.

PR #17075, opened 2019-10-06
Removed VMware VM import

apinnick:
@vikram-redhat Please tag with openshift-docs for review. Thanks.

VMware VM import section removed for CNV 2.1.0.


vikram-redhat:
@apinnick hey there Avital - you don't need to delete the modules and the assembly. Simply removing them from the topic map will ensure they don't appear in the published docs. As I understand it, this is going to be reintroduced with 2.1.1 so best to keep the files in, but just not have them in the topic map.

apinnick:
> @apinnick hey there Avital - you don't need to delete the modules and the assembly. Simply removing the pointer to the assembly from the topic map will ensure they don't appear in the published docs. As I understand it, this is going to be reintroduced with 2.1.1 so best to keep the files in, but just not have them in the topic map.

Ah -- OK. I wasn't sure whether it was OK to leave in modules that wouldn't be used. I'll close this PR and create a new one. Thanks.

PR #17076, opened 2019-10-06
Removed VMware VM import from topic_map.yml

apinnick:
Removed VMware VM import for CNV 2.1.0

@vikram-redhat Ready for review


vikram-redhat:
LGTM

PR #17079, opened 2019-10-06
Add'l params for OSP. BZ 1758114.

maxwelldb:
Adding details for OSP. See https://bugzilla.redhat.com/show_bug.cgi?id=1758114

@morenod Hi! Does this resolve what you found in the BZ ticket?

### DIFF
--- a/modules/installation-configuration-parameters.adoc
+++ b/modules/installation-configuration-parameters.adoc
@@ -63,12 +63,12 @@ uses the `<metadata.name>.<baseDomain>` format.
 |`controlPlane.platform`
 |The cloud provider to host the control plane machines. This parameter value
 must match the `compute.platform` parameter value.
-|`aws`, `azure`, or `gcp`
+|`aws`, `azure`, `gcp`, or, for OpenStack, `{}`
 
 |`compute.platform`
 |The cloud provider to host the worker machines. This parameter value
 must match the `controlPlane.platform` parameter value.
-|`aws`, `azure`, `osp` or `gcp`
+|`aws`, `azure`, `gcp`, `openstack`
 
 |`metadata.name`
 |The name of your cluster.
@@ -79,8 +79,8 @@ endif::osp[]
 
 |`platform.<platform>.region`
 |The region to deploy your cluster in.
-|A valid region for your cloud, such as `us-east-1` for AWS or `centralus`
-for Azure.
+|A valid region for your cloud, such as `us-east-1` for AWS, `centralus`
+for Azure, or `regionone` for OpenStack.
 
 |`pullSecret`
 |The pull secret that you obtained from the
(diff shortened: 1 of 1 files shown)


modules/installation-configuration-parameters.adoc
```diff
@@ -63,12 +63,12 @@ uses the `<metadata.name>.<baseDomain>` format.
 |`controlPlane.platform`
 |The cloud provider to host the control plane machines. This parameter value
 must match the `compute.platform` parameter value.
-|`aws`, `azure`, or `gcp`
+|`aws`, `azure`, `gcp`, or, for OpenStack, `{}`
```
kalexand-rh: This change invalidates the descriptions for these parameters, at least for OpenStack.

maxwelldb: Fair point. 

@morenod Does the OSP value here definitely not match as the others do?

morenod: when compute.platform is empty (nothing to say about it), value is {}.
when it has content, value is the name of the platform.

this is valid for aws and openstack (I cannot confirm the rest of the platforms as I havent tested them)

maxwelldb: @kalexand-rh So, it seems like controlPlane.platform and compute.platform can also have {} as a platform-independent value. Can you confirm this?

kalexand-rh: > when compute.platform is empty (nothing to say about it), value is {}.
> when it has content, value is the name of the platform.

@abhinavdahiya, can you comment on the intended usage of `{}` as a `compute.platform` value in the `install-config.yaml` file? I thought it was intended to blank out the other platform variables and that the `compute.platform` and `controlPlane.platform` parameter values needed to match.

abhinavdahiya: > https://github.com/openshift/installer/blob/master/docs/user/openstack/customization.md#machine-pools

it seems like openstack does have compute level properties.

abhinavdahiya: ex: https://github.com/openshift/installer/blob/master/docs/user/openstack/customization.md#custom-machine-pools

maxwelldb: @abhinavdahiya So is it the case that these two fields don't necessarily have to match like the docs currently state?

maxwelldb:
@morenod Does this PR now resolve your BZ ticket?

morenod:
@maxwelldb lgtm

maxwelldb:
Thanks!

maxwelldb:
@bergerhoffer @lamek @bobfuru Can I get an :+1: / :-1: on this whenever you have a moment? David's both the reporter and QE contact, so this should be good up to that point.

bergerhoffer (commented):
LGTM

Issue #12227, opened 2018-09-28
Edit suggested in file install_config/master_node_configuration.adoc


mhutter:
### Which section(s) is the issue in?

"Node Configuration Files".

### What needs fixing?

"Example 1" incorrectly states that `iptablesSyncPeriod` defaults to `5s`. This was changed back in 3.3 to `30s`: https://bugzilla.redhat.com/show_bug.cgi?id=1367199

### Affected

probably affects all versions and variants back to 3.3

PR #17093, opened 2019-10-07
Fix incorrect default value for iptablesSyncPeriod

mhutter:
I know 3.9 is EOL soon, but we still have some clusters that use it and this now bit me repeatedly 🙈 

Relates: #12227

### DIFF
--- a/install_config/master_node_configuration.adoc
+++ b/install_config/master_node_configuration.adoc
@@ -1449,7 +1449,7 @@ dockerConfig:
 imageConfig:
   format: openshift/origin-${component}:${version}
   latest: false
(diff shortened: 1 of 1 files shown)


jboxman:
@weliang1, can you confirm this change for OCP 3.9 (and earlier)?

mhutter:
@jboxman this was changed back in 3.3: https://bugzilla.redhat.com/show_bug.cgi?id=1367199#c9

jboxman:
@vikram-redhat, did we stop builds for 3.10 and below? If so, this is no longer fixable.

vikram-redhat:
Yeah 3.10 and below won't be fixed.

PR #17094, opened 2019-10-07
enterprise-3.11/3.10: Clean up node-config documentation

mhutter:
Remove duplications.

Relates: #12227

### DIFF
--- a/admin_guide/manage_nodes.adoc
+++ b/admin_guide/manage_nodes.adoc
@@ -500,7 +500,6 @@ To make configuration changes to an existing node, edit the appropriate configur
 watches for changes in the configuration maps. During installation, the sync pods are created by using _sync Daemonsets_,
 and a *_/etc/origin/node/node-config.yaml_* file, where the node configuration parameters reside, is added to each node. When a sync pod
 detects configuration map change, it updates the *_node-config.yaml_* on all nodes in that node group and restarts the appropriate nodes.
-// end::node-configmap[]
 
 ----
 $ oc get cm -n openshift-node
@@ -513,6 +512,7 @@ node-config-master-infra   1         1d
 ----
 
 .Sample configuration map for the *node-config-compute* group
+[source,yaml]
 ----
 apiVersion: v1
 authConfig:      <1>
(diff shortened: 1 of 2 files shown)


vikram-redhat:
@mburke5678 PTAL. 

This probably needs to go into master-3 first.

mhutter:
Ah, missed that. Ping me if I should reopen a PR against `master-3`

admin_guide/manage_nodes.adoc
```diff
@@ -537,6 +537,8 @@ node-config-master-infra   1         1d
 ----
 
 .Sample configuration map for the *node-config-compute* group
+====
```
mburke5678: @mhutter We have been removing the `====` around code examples for some time. Adding these here would be the only place in the file that has them. Unless you have a specific reason for adding `====` that I am not aware of.

admin_guide/manage_nodes.adoc
```diff
@@ -595,10 +597,7 @@ volumeConfig:
   localQuota:
     perFSGroup: null    <8>
 volumeDirectory: /var/lib/origin/openshift.local.volumes
-
-Events:  <none>
```
mburke5678: @mhutter Why remove the `Events:  <none>`? Is that not in the ConfigMap?

install_config/master_node_configuration.adoc
```diff
@@ -1442,68 +1442,6 @@ To modify a node in your cluster, update the xref:../admin_guide/manage_nodes.ad
 Do not manually edit the `node-config.yaml` file.
 ====
 
-////
```
mburke5678: @mhutter What if instead of just removing the `node-config.yaml` we replace it with a sample node ConfigMap from https://docs.openshift.com/container-platform/3.11/admin_guide/manage_nodes.html#modifying-nodes?

See https://github.com/openshift/openshift-docs/pull/17158

admin_guide/manage_nodes.adoc
mhutter: No, it is not. It's part of the output of an `oc describe ...` command.

admin_guide/manage_nodes.adoc
mhutter: Ah, no I have not :)

install_config/master_node_configuration.adoc
mhutter: Would work as well, as long as the duplication is eliminated (so now only one example can get outdated :) )

mburke5678: @mhutter Please take a look at my PR. https://github.com/openshift/openshift-docs/pull/17158

If you do not have time, I can merge this PR for you.

mburke5678:
@mhutter I can merge this one if you can squash the commits. If not, I can merge mine.

mhutter:
Rebased onto current `master-3` and squashed commits

PR #17095, opened 2019-10-07
Note about phase/conditions being "best guess"

ousleyp:
Based on @alexxa's recommendation, I'm adding a note back into the CNV upgrade docs that I had written and subsequently removed before merging PR #16791.

Cherrypick to enterprise-4.2 only, thanks!

cc @kalexand-rh

### DIFF
--- a/modules/cnv-monitoring-upgrade-status.adoc
+++ b/modules/cnv-monitoring-upgrade-status.adoc
@@ -6,7 +6,14 @@
 = Monitoring upgrade status
 
 The best way to monitor {CNVProductName} upgrade status is to watch the
-ClusterServiceVersion (CSV) `PHASE`.
+ClusterServiceVersion (CSV) `PHASE`. You can also monitor the CSV conditions
+in the web console or by running the command provided here.
++
+[NOTE]
+====
+The `PHASE` and conditions values are approximations that are based on
+available information.
+====
 
 .Prerequisites
 
@@ -30,7 +37,7 @@ VERSION  REPLACES                                        PHASE
 ----
 
 . Optional: Monitor the aggregated status of all {CNVProductName} component
-`Conditions` by running the following command:
+conditions by running the following command:
 +
 ----
 $ oc get hco -n openshift-cnv hyperconverged-cluster \


modules/cnv-monitoring-upgrade-status.adoc
```diff
@@ -6,7 +6,15 @@
 = Monitoring upgrade status
 
 The best way to monitor {CNVProductName} upgrade status is to watch the
-ClusterServiceVersion (CSV) `PHASE`.
+ClusterServiceVersion (CSV) `PHASE`. You can also monitor the CSV `Conditions`
+in the web console or by running the command provided here.
++
+[NOTE]
+====
+`PHASE` and `Conditions` both represent the system's best guess based on
```
kalexand-rh: I'm objecting to the note mainly because the system can't guess.

> Do not attribute human characteristics to inanimate objects, which are incapable of human
behavior and emotions.

I'd probably say "The `PHASE` and `Conditions` values are approximations that are based on available information."

Please add the heading values at line 50.

alexxa: There is no heading values. That command on lines 44-45 parses a big yaml where all data are in rows, and prints out status, type and message in a summary like table.

kalexand-rh: If they're not labeled as `Conditions` in the output, then remove the backticks and uppercase format. I think it's clear enough from the context what the conditions are, but if it's not a specific parameter, don't format it like one.

alexxa: Just to be sure, it's clear... `Conditions` is a specific parameter in that parsed yaml file, which is also displayed in UI view. It's just with that command we don't print it names of parameters, but their values only:

```
[root@cnv-qe-03 ~]# oc describe hco -n openshift-cnv hyperconverged-cluster | grep -A 5 -B 5 Condition
  Generation:          1
  Resource Version:    125110
  Self Link:           /apis/hco.kubevirt.io/v1alpha1/namespaces/openshift-cnv/hyperconvergeds/hyperconverged-cluster
  UID:                 d30eec8a-e9cd-11e9-b923-525400d67956
Status:
  Conditions:
    Last Heartbeat Time:   2019-10-08T13:23:18Z
    Last Transition Time:  2019-10-08T13:16:23Z
    Message:               Reconcile completed successfully
    Reason:                ReconcileCompleted
    Status:                True

``` 
Screenshot from 2019-10-07 13-37-09

kalexand-rh: ... then I think it's a bug and that the parameter labels should be printed in the output.

alexxa: it seems I fail to explain myself =)
it's a command (line 44-45) which says print me `{.type}{"\t"}{.status}{"\t"}{.message}` values only, no those labels or that it belongs to `{range .status.conditions[*]}`, in a table view, instead of many lines for each CNV component.

alexxa: anyway, I just wanted to be clear, that Conditions label exists. I'm fine if that note will not capitalize it or backtick

kalexand-rh: Thank you! That is clearer. I'm glad that you're fine with removing the styling from the conditions because I still think that's the best thing to do for now.

ousleyp: Thanks for the discussion, @kalexand-rh and @alexxa. I rephrased it as suggested and removed the "conditions" styling in a few places. 

@rthallisey Let me know if you have any concerns before we merge. Thanks!

PR #17114, opened 2019-10-08
removing semantic versioning

kalexand-rh:
Per https://github.com/openshift/installer/pull/2442/files

@wking, will you please +1 this change for 4.1 and 4.2?


wking (approved):
This is porting openshift/installer#2442, and as described there, this should have happened before 4.1.0 went GA, when ART started building installers instead of me ;).

kalexand-rh:
@vikram-redhat, do you mind if I merge without QE?

vikram-redhat:
@kalexand-rh all good. I realize that it is unlikely we will need to reuse **this** file going forward, but I would have kept it still as a module instead of deleting it and then let the unused module script clean it up (but of course, removing the include for it.)

kalexand-rh:
If that's the plan for all module pruning going forward, I think we need to add more nuance to the docs manual.

PR #17155, opened 2019-10-09
BZ1756296: Added information about default uid and gid.

huffmanca:
Included information about the considerations of Azure File, specifically that the uid will not necessarily match the Pod's.

This is for OCP 4.2.

### DIFF
--- a/modules/dynamic-provisioning-azure-file-considerations.adoc
+++ b/modules/dynamic-provisioning-azure-file-considerations.adoc
@@ -1,0 +1,40 @@
+// Module included in the following assemblies:
+//
+// storage/persistent-storage/persistent-storage-azure-file.adoc
+
+[id="azure-file-considerations_{context}"]
+= Considerations when using Azure File
+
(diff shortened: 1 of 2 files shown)


bmcelvee:
LGTM

modules/dynamic-provisioning-azure-file-considerations.adoc
```diff
@@ -0,0 +1,40 @@
+// Module included in the following assemblies:
+//
+// storage/persistent-storage/persistent-storage-azure-file.adoc
+
+[id="azure-file-considerations_{context}"]
+= Considerations when using Azure File
```
liangxia: The line is render as "CONSIDERATIONS WHEN USING AZURE FILE" does not look correct.

liangxia:
LGTM except the render issue.

vikram-redhat:
Thanks @liangxia. @huffmanca can fix the render issue later tomorrow.

PR #17196, opened 2019-10-10
modules/installation-bootstrap-gather: --master for each machine

wking:
This is what we do in CI since openshift/release@3013859bdc (openshift/release#4734), so we know it works.  Comma-delimited machines won't work, as described in that release commit.  I'd be surprised if space-delimited worked.

### DIFF
--- a/modules/installation-bootstrap-gather.adoc
+++ b/modules/installation-bootstrap-gather.adoc
@@ -49,15 +49,20 @@ command:
 ----
 $ ./openshift-install gather bootstrap --dir=<directory> \ <1>
     --bootstrap <bootstrap_address> \ <2>
-    --master "<master_address> <master_address> <master_address>" <3>
+    --master <master_1_address> \ <3>
+    --master <master_2_address> \ <3>
+    --master <master_3_address>" <3>
 ----
 <1> `installation_directory` is the directory you stored the {product-title}
 definition files that the installation program creates.
 <2> `<bootstrap_address>` is the fully-qualified domain name or IP address of
 the cluster's bootstrap machine.
-<3> `<master_address>` is the fully-qualified domain name or IP address of a
-control plane, or master, machine in your cluster. Specify a space-delimited
-list that contains all the control plane machines in your cluster.
+<3> For each control plane, or master, machine in your cluster, replace `<master_*_address>` with its fully-qualified domain name or IP address.
++
+[NOTE]
+====
+A default cluster contains three control plane machines. List all of your control plane machines as shown, no matter how many your cluster uses.
+====
 --
 +
 The command output resembles the following example:


wking:
Or maybe it works accidentally now because we have poor quoting/validation ;).

wking:
Pushed an additional commit to catch up with openshift/installer#1822.

modules/installation-bootstrap-gather.adoc
```diff
@@ -49,54 +49,22 @@ command:
 ----
 $ ./openshift-install gather bootstrap --dir=<directory> \ <1>
     --bootstrap <bootstrap_address> \ <2>
-    --master "<master_address> <master_address> <master_address>" <3>
+    --master <master1_address> --master <master2_address> --master <master3_address>" <3>
```
kalexand-rh:
```suggestion
    --master <master_1_address> \ <3>
    --master <master_2_address> \ <3>
    --master <master_3_address>" <3>
```

modules/installation-bootstrap-gather.adoc
```diff
@@ -49,54 +49,22 @@ command:
...
 ----
 <1> `installation_directory` is the directory you stored the {product-title}
 definition files that the installation program creates.
 <2> `<bootstrap_address>` is the fully-qualified domain name or IP address of
 the cluster's bootstrap machine.
-<3> `<master_address>` is the fully-qualified domain name or IP address of a
-control plane, or master, machine in your cluster. Specify a space-delimited
-list that contains all the control plane machines in your cluster.
+<3> `<master*_address>` are the fully-qualified domain names or IP addresses of
```
kalexand-rh:
```suggestion
<3> For each control plane, or master, machine in your cluster, replace `<master_*_address>` with its fully-qualified domain names or IP addresses.
```

kalexand-rh (commented):
Thank you for the update! I might change it up a bit.

wking:
Squashed in your suggestions.

kalexand-rh:
@wking, you'll eventually need to squash to one commit.

@shlao, will you PTAL?

wking:
> you'll eventually need to squash to one commit.

Because of the different backport depths, I just spun the second commit off into its own PR (#17651).  This one can go back to 4.1 and 4.2.

kalexand-rh:
@shlao, will you PTAL?

shlao:
@ahardin-rh  it's good, but In the big Corps, there are, maybe, more than 3 masters.
Can we use this output?
```
 --master <master_3_address>" (3)
... ...
 --master <master_N_address>" (N)
```

modules/installation-bootstrap-gather.adoc
```diff
@@ -49,15 +49,16 @@ command:
...
 ----
 <1> `installation_directory` is the directory you stored the {product-title}
 definition files that the installation program creates.
 <2> `<bootstrap_address>` is the fully-qualified domain name or IP address of
 the cluster's bootstrap machine.
-<3> `<master_address>` is the fully-qualified domain name or IP address of a
-control plane, or master, machine in your cluster. Specify a space-delimited
-list that contains all the control plane machines in your cluster.
+<3> For each control plane, or master, machine in your cluster, replace `<master_*_address>` with its fully-qualified domain names or IP addresses.
+control plane, or master, machines in your cluster.
```
kalexand-rh: This line needs to be removed.

kalexand-rh:
> @ahardin-rh it's good, but In big Corps, there are, maybe, more than 3 masters.
> Can we use this output?
> 
> ```
>  --master <master_3_address>" (3)
> ... ...
>  --master <master_N_address>" (N)
> ```

 I think that that would make the command harder to use, but we can be more explicit in the note. @wking, what do you think of the commit that I added?

wking:
I've pushed 0b1e9890c -> 518234b7e, softening @kalexand-rh's wording a bit to drop the "must" and allowing for clusters with fewer control plane machines (we may not support one or two control plane machines, but the gather logic doesn't care about that).

modules/installation-bootstrap-gather.adoc
```diff
@@ -49,15 +49,20 @@ command:
...
 <2> `<bootstrap_address>` is the fully-qualified domain name or IP address of
 the cluster's bootstrap machine.
-<3> `<master_address>` is the fully-qualified domain name or IP address of a
-control plane, or master, machine in your cluster. Specify a space-delimited
-list that contains all the control plane machines in your cluster.
+<3> For each control plane, or master, machine in your cluster, replace `<master_*_address>` with its fully-qualified domain name or IP address.
++
+[NOTE]
+====
+A default cluster contains three control plane machines. List all of your control plane machines, even if you aren't using three like the example above.
```
kalexand-rh: I think the "must" could have stayed.

```suggestion
A default cluster contains three control plane machines. List all of your control plane machines as shown, no matter how many your cluster uses.
```

Our style guide prohibits both contractions and the use of "above" like that.

wking: Sounds good :)  I can squash back down to one commit when I get back to my laptop.

kalexand-rh: Great! Thank you, Trevor!

wking: Squashed and rebased onto the current master with 8a32a0925 -> 1e49f6876.

kalexand-rh:
I think this will work. Thank you!

PR #17232, opened 2019-10-11
BZ1760250 Add GCE sample definition

bobfuru:
[BZ 1760250](https://bugzilla.redhat.com/show_bug.cgi?id=1760250)
I'm not sure that callout 2 and the note at the end are necessary, or if we should only list `replication-type` as `none` and not document the `regional-pd` value at all.

@liangxia @huffmanca PTAL

### DIFF
--- a/modules/dynamic-provisioning-gce-definition.adoc
+++ b/modules/dynamic-provisioning-gce-definition.adoc
@@ -15,19 +15,6 @@ metadata:
 provisioner: kubernetes.io/gce-pd
 parameters:
   type: pd-standard  <1>
-  zone: us-central1-a  <2>
-  zones: us-central1-a, us-central1-b, us-east1-b  <3>
-  fsType: ext4 <4>
+  replication-type: none
 ----
 <1> Select either `pd-standard` or `pd-ssd`. The default is `pd-ssd`.
-<2> GCE zone. If no zone is specified, volumes are generally round-robined
-across all active zones where the {product-title} cluster has a node. 
-Zone and zones parameters must not be used at the same time.
-<3> A comma-separated list of GCE zone(s). If no zone is specified, 
-volumes are generally round-robined across all active zones where the 
-{product-title} cluster has a node. Zone and zones parameters must not 
-be used at the same time.
-<4> File system that is created on dynamically provisioned volumes. This 
-value is copied to the `fsType` field of dynamically provisioned 
-persistent volumes and the file system is created when the volume is 
-mounted for the first time. The default value is `ext4`.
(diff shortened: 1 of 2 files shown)


huffmanca:
@bobfuru ,

I'm inclined to leave the annotation off of `replication-type`, considering that the only other option for this attribute requires deprecated parameters.

Considering that the default value for this parameter is the only one that doesn't require additional, deprecated, parameters, I would delete the line entirely and let it default to `none`.

liangxia:
The parameter `zone` and `zones` are removed there (I'm OK with the remove here), but it still exist for [AWS](https://docs.openshift.com/container-platform/4.2/storage/dynamic-provisioning.html#aws-definition_dynamic-provisioning). Shall we report a bug to request removing that as well ?

liangxia:
Also, I'm OK leaving the `replication-type` there as our default storage class on GCP have it, which can explain why it is set to `none`.

```
$ oc version
Client Version: v4.2.0
Server Version: 4.2.0-rc.5
Kubernetes Version: v1.14.6+2e5ed54

$ oc get sc standard -o yaml
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    storageclass.kubernetes.io/is-default-class: "true"
  creationTimestamp: "2019-10-14T06:25:58Z"
  name: standard
  ownerReferences:
  - apiVersion: v1
    kind: clusteroperator
    name: storage
    uid: 9020ec49-ee4a-11e9-a432-42010a000005
  resourceVersion: "8129"
  selfLink: /apis/storage.k8s.io/v1/storageclasses/standard
  uid: 7c3164a9-ee4b-11e9-9a6b-42010a000003
parameters:
  replication-type: none
  type: pd-standard
provisioner: kubernetes.io/gce-pd
reclaimPolicy: Delete
volumeBindingMode: WaitForFirstConsumer

```

bobfuru:
> @bobfuru ,
> 
> I'm inclined to leave the annotation off of `replication-type`, considering that the only other option for this attribute requires deprecated parameters.
> 
> Considering that the default value for this parameter is the only one that doesn't require additional, deprecated, parameters, I would delete the line entirely and let it default to `none`.

I agree and had the same inclination. Will remove the annotation and leave it as `replication-type: none`. Thanks.

bobfuru:
> The parameter `zone` and `zones` are removed there (I'm OK with the remove here), but it still exist for [AWS](https://docs.openshift.com/container-platform/4.2/storage/dynamic-provisioning.html#aws-definition_dynamic-provisioning). Shall we report a bug to request removing that as well ?

Yes, good catch. I've opened [BZ 1761524](https://bugzilla.redhat.com/show_bug.cgi?id=1761524) to get it removed from AWS EBS sample.

PR #17233, opened 2019-10-11
Added information about upgrade channels / phased rollout + migration tool links

ahardin-rh:
Per discussion in the program meeting. Adding content to the Upgrading topics, as well as the 4.2 Release Notes.

Preview Build: http://file.rdu.redhat.com/~ahardin/10112019/phased-rollout/updating/updating-cluster.html

cc @aheslin @kalexand-rh @vikram-redhat @sferich888 @reestr

### DIFF
--- a/modules/understanding-upgrade-channels.adoc
+++ b/modules/understanding-upgrade-channels.adoc
@@ -1,0 +1,24 @@
+// Module included in the following assemblies:
+//
+// updating/updating-cluster.adoc
+// updating/updating-cluster-cli.adoc
+// updating/updating-cluster-rhel-compute.adoc
+// updating/updating-disconnected-cluster.adoc
+
+[id="understanding_upgrade_channels_{context}"]
+= Understanding {product-title} upgrade channels
+
+In {product-title} 4.1, Red Hat introduced the concept of upgrade channels for
+recommending the appropriate upgrade versions to your cluster. Upgrade channels
+separate upgrade strategies and also are used to control the cadence of updates.
+Channels are tied to a minor version of {product-title}. For instance,
+{product-title} {product-version} channels will never include an upgrade to a
+4.3 release. This ensures administrators make an explicit decision to upgrade to
+the next minor version of {product-title}. Channels only control updates and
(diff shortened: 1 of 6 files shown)


modules/understanding-upgrade-channels.adoc
```diff
@@ -0,0 +1,24 @@
...
+separate upgrade strategies and also are used to control the cadence of updates.
+Channels are tied to a minor version of {product-title}. For instance,
+{product-title} {product-version} channels will never include an upgrade to a
+4.3 release. This ensures administrators make an explicit decision to upgrade to
+the next minor version of {product-title}. Channels only control updates and
+have no impact on the version of the cluster you install; the
+`openshift-install` binary for a given patch level of {product-title} always
+installs that patch level.
+
+See link:https://access.redhat.com/articles/4495171[OpenShift 4.2 Upgrades
```
smarterclayton: Is there a reason we didn't include the full description?  Why go to an article?

kalexand-rh (commented):
LGTM! I'm going to make a couple of changes in my in-flight update PR.

modules/understanding-upgrade-channels.adoc
ahardin-rh: @smarterclayton From the discussions I was following, it sounded like the content could continue to evolve and that we could potentially add in more types. I didn't want to duplicate any information and effort and thought pointing to the article could help with future docs maintenance while that is all sorted out. Let me know if that's a misunderstanding on my part. If so, we can pull the content into our doc set instead. Thanks!

ahardin-rh:
Thanks! Merging this to get it into GA. If we have to re-structure anything, I will open a new PR.

modules/understanding-upgrade-channels.adoc
smarterclayton: That's fine, we can always do it later.  I would like to see some of the discussion in the doc in the long run, but yes, it may still be evolving a bit.

kalexand-rh: @smarterclayton, is that KB stable enough to roll into the docs next week?

PR #17235, opened 2019-10-12
BZ1761067: Corrected local storage output example.

huffmanca:
This corrects the output of the `local-storage` namespace.

This is for OCP 4.2.


liangxia:
LGTM (from QE)

PR #17281, opened 2019-10-14
Update procedure for removing an additional network

jboxman:
- https://bugzilla.redhat.com/show_bug.cgi?id=1758295

### DIFF
--- a/modules/nw-multus-delete-network.adoc
+++ b/modules/nw-multus-delete-network.adoc
@@ -8,6 +8,13 @@ As a cluster administrator, you can remove an additional network from your
 {product-title} cluster. The additional network is not removed from any Pods it
 is attached to.
 
+[IMPORTANT]
+====
+In {product-title} 4.2.0 you must manually delete the additional network CR after removing it from the Cluster Network Operator configuration.
+This will be fixed in a future release.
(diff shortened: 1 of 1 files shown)


modules/nw-multus-delete-network.adoc
```diff
@@ -8,6 +8,13 @@ As a cluster administrator, you can remove an additional network from your
 {product-title} cluster. The additional network is not removed from any Pods it
 is attached to.
 
+[IMPORTANT]
+====
+In {product-title} 4.2.0 you must manually delete the additional network CR after removing it from the Cluster Network Operator configuration.
```
bobfuru: Are we certain this will apply only to this z-stream release (`4.2.0`), rather than just `4.2`?

bobfuru (commented):
One question but LGTM otherwise!

modules/nw-multus-delete-network.adoc
jboxman: @bobfuru, it's targeted to be fixed in 4.3.0 and back ported to 4.2.1.

weliang1 (commented):
LGTM

jboxman:
@weliang1 thank you!

PR #16985, opened 2019-10-01
osdocs-690 entitlement updates

kalexand-rh:
https://jira.coreos.com/browse/OSDOCS-690

@wking, @reestr, PTAL?

This PR also fixes https://bugzilla.redhat.com/show_bug.cgi?id=1756513

PR #17304, opened 2019-10-15
OpenStack -> RHOSP in ShiftStack IPI docs

maxwelldb:
Based on discussion and resulting PM clarification here: https://github.com/openshift/openshift-docs/pull/16985#discussion_r334549189

Related to https://jira.coreos.com/browse/OSDOCS-470

For 4.2.

### DIFF
--- a/modules/installation-osp-about-kuryr.adoc
+++ b/modules/installation-osp-about-kuryr.adoc
@@ -5,18 +5,18 @@
 [id="installation-osp-about-kuryr_{context}"]
 = About Kuryr
 
-Kuryr is a container network interface (CNI) plug-in that uses Neutron and Octavia to provide networking to Pods and Services. It is primarily designed for {product-title} clusters that run on OpenStack instances.
+Kuryr is a container network interface (CNI) plug-in that uses Neutron and Octavia to provide networking to Pods and Services. It is primarily designed for {product-title} clusters that run on Red Hat OpenStack Platform (RHOSP) instances.
 
-Kuryr improves network performance by connecting Pods to the OpenStack SDN. It also provides connections between Pods and OpenStack instances.
+Kuryr improves network performance by connecting Pods to the OpenStack SDN. It also provides connections between Pods and RHOSP instances.
 
 [IMPORTANT]
 ====
-Kuryr is recommended for deployments on encapsulated OpenStack tenant networks to avoid double encapsulation. For example, on a deployment of an encapsulated OpenShift SDN over an OpenStack network.
+Kuryr is recommended for deployments on encapsulated RHOSP-tenant networks to avoid double encapsulation. For example, on a deployment of an encapsulated OpenShift SDN over a RHOSP network.
 
 The default OpenShift SDN is recommended if:
 
 * You use provider networks or tenant VLANs.
-* The deployment will use many Services on a few hypervisors. Each Service creates an Octavia Amphora virtual machine in OpenStack that hosts a required load balancer.
+* The deployment will use many Services on a few hypervisors. Each Service creates an Octavia Amphora virtual machine in RHOSP that hosts a required load balancer.
 * You need UDP Services.
 ====
 
--- a/modules/installation-osp-accessing-api.adoc
+++ b/modules/installation-osp-accessing-api.adoc
@@ -7,6 +7,6 @@
 [id="installation-osp-accessing-api_{context}"]
 = Enabling access to the environment
 
-At deployment, all {product-title} machines are created in an OpenStack tenant network. Therefore, they are not accessible directly in most OpenStack deployments.
+At deployment, all {product-title} machines are created in a Red Hat OpenStack Platform (RHOSP)-tenant network. Therefore, they are not accessible directly in most RHOSP deployments.
 
 You can configure the {product-title} API to be accessible either with or without floating IP addresses.
--- a/modules/installation-osp-troubleshooting.adoc
+++ b/modules/installation-osp-troubleshooting.adoc
@@ -1,6 +1,6 @@
 // Module included in the following assemblies:
 //
-// * TBD
+// * n/a
 
 [id="installation-osp-customizing_{context}"]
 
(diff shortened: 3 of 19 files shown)


maxwelldb:
@bergerhoffer @lamek @bobfuru This needs a look and merge. L size, but it's all just name changes.

maxwelldb:
And yes, "a RHOSP" might read a little strangely, but it seems to be more commonly used in the OSP docs than "an RHOSP," so I'm sticking with it for now.

modules/installation-osp-about-kuryr.adoc
```diff
@@ -5,18 +5,18 @@
...
-Kuryr is a container network interface (CNI) plug-in that uses Neutron and Octavia to provide networking to Pods and Services. It is primarily designed for {product-title} clusters that run on OpenStack instances.
+Kuryr is a container network interface (CNI) plug-in that uses Neutron and Octavia to provide networking to Pods and Services. It is primarily designed for {product-title} clusters that run on Red Hat OpenStack Platform (RHOSP) instances.
 
-Kuryr improves network performance by connecting Pods to the OpenStack SDN. It also provides connections between Pods and OpenStack instances.
+Kuryr improves network performance by connecting Pods to the OpenStack SDN. It also provides connections between Pods and RHOSP instances.
 
 [IMPORTANT]
 ====
-Kuryr is recommended for deployments on encapsulated OpenStack tenant networks to avoid double encapsulation. For example, on a deployment of an encapsulated OpenShift SDN over an OpenStack network.
+Kuryr is recommended for deployments on encapsulated RHOSP tenant networks to avoid double encapsulation. For example, on a deployment of an encapsulated OpenShift SDN over a RHOSP network.
```
bobfuru: s/RHOSP tenant/RHOSP-tenant (assuming this is similar to a "single-tenant network")

modules/installation-osp-accessing-api.adoc
```diff
@@ -7,6 +7,6 @@
 [id="installation-osp-accessing-api_{context}"]
 = Enabling access to the environment
 
-At deployment, all {product-title} machines are created in an OpenStack tenant network. Therefore, they are not accessible directly in most OpenStack deployments.
+At deployment, all {product-title} machines are created in a Red Hat OpenStack Platform (RHOSP) tenant network. Therefore, they are not accessible directly in most RHOSP deployments.
```
bobfuru: s/(RHOSP) tenant/(RHOSP)-tenant

modules/installation-osp-troubleshooting.adoc
```diff
@@ -11,12 +11,12 @@
 
 Unfortunately, there will always be some cases where {product-title} fails to install properly. In these events, it is helpful to understand the likely failure modes as well as how to troubleshoot the failure.
 
-This document discusses some troubleshooting options for OpenStack based
+This document discusses some troubleshooting options for RHOSP based
```
bobfuru: s/RHOSP based/RHOSP-based

bobfuru (commented):
A few very small suggestions.

modules/installation-osp-troubleshooting.adoc
maxwelldb: This module isn't likely to see the light of day, but :+1:

bobfuru:
LGTM!

bergerhoffer:
@maxwelldb Yes "a RHOSP" does sound wrong to me. Where did you look? I just opened a few RHOSP docs and didn't see any references at all to either "a RHOSP" or "an RHOSP".

maxwelldb:
@bergerhoffer I found it in OSP Certification docs, e.g. https://access.redhat.com/documentation/en-us/red_hat_openstack_certification/1.0/html/red_hat_openstack_application_and_vnf_workflow_guide/publishing-a-rhosp-application-certification. 

The mainline OSP docs seem to avoid the abbreviation almost entirely, though they also frequently violate the product naming guidelines, e.g. using "OpenStack Platform." You can find the occasional "an RHOSP" in there, though.

AFAIK, IBM guidelines say to stick with "a" if it's pronounced something like "rossp" and "an" if it's an initialism. I assumed from the frequency of "a RHOSP" vs. "an RHOSP" on redhat.com that it's not the latter.

bergerhoffer:
@maxwelldb It's not a critical issue, though I still think it sounds bad. I checked with the OpenStack DPM who says that their guidance is to typically write out the full name (he says they have ~900 instances written out and only a handful of RHOSP in the latest.

For our purposes, I think we should just write the full name out every time, especially for as infrequently as it is used. We should probably add things like this as attributes in the common-attributes.adoc file.

maxwelldb:
@bergerhoffer I don't think writing it out every time is a good move. The phrase would appear dozens of times per assembly, and it's kind of cumbersome as it is--especially since it's frequently paired with {product-title}. 

I think it would be good to get the OSP DPM to align with the master product and solution naming list: https://docs.google.com/spreadsheets/d/1DLS_lS3VKidgZIvcLmLp9BoiqptkvqHWfe1D5FD2kfk/edit#gid=1987148185

It would be good to clarify the guidance that's there, and then we could use the result of that to update common-attributes.adoc.

bergerhoffer:
@maxwelldb The master naming spreadsheet says that we can use RHOCP in technical documentation, but we don't.

maxwelldb:
@bergerhoffer And that's something that I'm discussing with Vikram later today, as seeing the list got me thinking. I don't think the multiple sets of guidelines floating around--formal and not--are helpful. 

Those guidelines do allow RHOSP as a short name, so that's what I used.

bergerhoffer:
@maxwelldb Sure. If the OSP docs don't use those guidelines, then it probably shouldn't say that in the spreadsheet for other teams to use. Same with RHOCP.

Either way, they should also add guidance on the "a" vs. "an", and the pronunciation of them. But if you do go with RHOSP in our docs, then I think you should use "an" unless we know for sure that they want people saying rossp.

maxwelldb:
@bergerhoffer I can ask around to see who can correct that. Is Chuck Copello the OSP DPM? 

Aside from that, it's just:

>Choose the indefinite article _a_ or _an_ according to how the abbreviation, word, or number is spoken. Use _a_ before an abbreviation, word, or number that begins with a consonant sound. Use "an" before an abbreviation, word, or number that begins with a vowel sound.

According to page 10 of the latest IBM Style Guide. I'd imagine that the OpenStack folks might be forced to declare a preference that hasn't yet been decided.

bergerhoffer:
Yep to Chuck. And about the "hasn't yet been decided" part, that's why I'm suggesting writing it out in our docs, so we're not making that decision. And because we don't know the answer to how it's spoken, if we have to make a choice, I would think the safest option would be to assume that it is spoken spelled out R-H-O-S-P.

bergerhoffer:
@maxwelldb Or is there any way that we can just write around it? Using "the" instead of "a" or "an", etc.?

maxwelldb:
@bergerhoffer Chuck said that they're currently in the midst of a style consistency review that'll solve "a" vs. "an," but to use "an RHCOS" for now.

I'll make a PR to update for that and the common attributes file presently.

PR #17328, opened 2019-10-15
Removed OCS and updated raw block storage.

huffmanca:
Removes OCS from the OCP 4.2 release notes, and rewords the details surrounding raw block volume support.

Note the ClusterStorageOperator has been present since 4.1, so this has also been removed.

This is for OCS 4.2.


ahardin-rh:
LGTM!

PR #17331, opened 2019-10-15
Update logging secure forward

mburke5678:
Per slack convo with @richm 
https://coreos.slack.com/archives/CB3HXM2QK/p1571171589181800?thread_ts=1571167864.174500&cid=CB3HXM2QK

* Added a full `secure_forward` section from the `fluentd` configmap
* Added a step to update the `secure_forward` with the cert file and private key
* Added an example of the secret patch command with the same cert file and private key in new step.

### DIFF
--- a/modules/efk-logging-fluentd-external.adoc
+++ b/modules/efk-logging-fluentd-external.adoc
@@ -25,16 +25,41 @@ For Rsyslog, you can edit the Rsyslog configmap to add support for Syslog log fo
 The logging deployment provides a `secure-forward.conf` section in the Fluentd configmap
 for configuring the external aggregator:
 
+.Sample `secure-forward.conf` section
+[source,yaml]
+----
+ secure-forward.conf: |
+    # <store>
+    # @type secure_forward
+
+    # self_hostname ${hostname}
+    # shared_key <SECRET_STRING>
+
+    # secure yes
+    # enable_strict_verification yes
+
+    # ca_cert_path /etc/fluent/keys/your_ca_cert
+    # ca_private_key_path /etc/fluent/keys/your_private_key
+      # for private CA secret key
+    # ca_private_key_passphrase passphrase
+
+    # <server>
+      # or IP
+    #   host server.fqdn.example.com
+----
+
 .Procedure
 
 To send a copy of Fluentd logs to an external log aggregator:
 
 . Edit the `secure-forward.conf` section of the Fluentd configuration map:
 +
-.Sample `secure-forward.conf` section
 ----
 $ oc edit configmap/fluentd -n openshift-logging
-
+----
++
+[source,yaml]
+----
 <store>
   @type forward
   <server> <1>
@@ -51,10 +76,45 @@ $ oc edit configmap/fluentd -n openshift-logging
 ----
 <1> Enter the name, host, and port for your external Fluentd server.
 
+. Add the path to your certificate and private key to the `secure-forward.conf` section:
++
+[source,yaml]
+----
+  secure-forward.conf: |
+    # <store>
+    # @type secure_forward
+
+    # self_hostname ${hostname}
+    # shared_key <SECRET_STRING>
+
+    secure yes
+    # enable_strict_verification yes
+
+    ca_cert_path /etc/fluent/keys/ca.crt <1>
+    ca_private_key_path /etc/fluent/keys/privkey.pem <2>
+      # for private CA secret key
+    # ca_private_key_passphrase passphrase
+
+    # <server>
+      # or IP
+    #   host server.fqdn.example.com
+    #   port 24284
+    # </server>
+    # <server>
+      # ip address to connect
+    #   host 203.0.113.8
+      # specify hostlabel for FQDN verification if ipaddress is used for host
+    #   hostlabel server.fqdn.example.com
+    # </server>
+    # </store>
+----
+<1> Add the path to your certificate.
+<2> Add the path to your private key.
+ 
 . Add certificates to be used in `secure-forward.conf` to the existing
 secret that is mounted on the Fluentd pods. The `your_ca_cert` and
 `your_private_key` values must match what is specified in `secure-forward.conf`
-in `configmap/logging-fluentd`:
+in `configmap/fluentd`:
 +
 ----
 $ oc patch secrets/fluentd --type=json \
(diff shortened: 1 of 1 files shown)


richm:
While you're in here, remove the Note about rsyslog.  In fact, these directions only apply for doing Fluentd -> Fluentd forwarding, since only Fluentd supports the `forward`/`secure_forward` protocol.

modules/efk-logging-fluentd-external.adoc
```diff
@@ -51,10 +76,45 @@ $ oc edit configmap/fluentd -n openshift-logging
...
+    # </server>
+    # <server>
+      # ip address to connect
+    #   host 203.0.113.8
+      # specify hostlabel for FQDN verification if ipaddress is used for host
+    #   hostlabel server.fqdn.example.com
+    # </server>
+    # </store>
+----
+<1> Add the path to your certificate.
```
richm: your CA certificate

modules/efk-logging-fluentd-external.adoc
```diff
@@ -51,10 +76,45 @@ $ oc edit configmap/fluentd -n openshift-logging
...
+    # @type secure_forward
+
+    # self_hostname ${hostname}
+    # shared_key <SECRET_STRING>
+
+    secure yes
+    # enable_strict_verification yes
+
+    ca_cert_path /etc/fluent/keys/ca.crt <1>
+    ca_private_key_path /etc/fluent/keys/privkey.pem <2>
```
richm: if you use mTLS (aka client cert auth with priv key) you will need to specify both the client cert and key using https://docs.fluentd.org/output/forward#tls_client_cert_path and https://docs.fluentd.org/output/forward#tls_client_private_key_path

richm: See also https://docs.fluentd.org/output/forward#how-to-connect-to-a-tls-ssl-enabled-server
that is, you can use tls without using tls_client_cert_path and tls_client_private_key_path - however, in order to use tls_client_cert_path and tls_client_private_key_path, you must also use tls

modules/efk-logging-fluentd-external.adoc
```diff
@@ -51,10 +76,45 @@ $ oc edit configmap/fluentd -n openshift-logging
...
+    # <server>
+      # ip address to connect
+    #   host 203.0.113.8
+      # specify hostlabel for FQDN verification if ipaddress is used for host
+    #   hostlabel server.fqdn.example.com
+    # </server>
+    # </store>
+----
+<1> Add the path to your certificate.
+<2> Add the path to your private key.
```
richm: probably omit this and say something like "If you want to use mTLS (client cert auth), see the https://docs.fluentd.org/output/forward page for information about client cert and key parameters and other settings.".

mburke5678:
@anpingli Please take a look. Thank you!!

vikram-redhat:
@mburke5678 should this be in master?

mburke5678:
@richm Do these changes apply to 4.2 and 4.3?

richm:
It is confusing.  There are a lot of references to using the `@type secure_forward` plugin, and to using the secure_forward feature of fluentd 0.12, which is only applicable to 3.11.

What is the intention of this documentation for 4.x?  Do we need to have any references to "This is how you do it if you use secure_forward and/or fluentd 0.12" in our 4.x docs?

mburke5678:
@richm I would think we want to document the _appropriate_ version of the plugin for 4.x. Do we not have that here?

modules/efk-logging-fluentd-external.adoc
```diff
@@ -17,24 +17,44 @@ ifdef::openshift-enterprise[]
...
 The logging deployment provides a `secure-forward.conf` section in the Fluentd configmap
-for configuring the external aggregator:
+in the `openshift-logging` project for configuring the external aggregator:
+
+.Sample `secure-forward.conf` section
+[source,yaml]
+----
+ secure-forward.conf: |
+    # <store>
+    # @type secure_forward
```
richm: This sample config is for 3.11.  Use https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/configs.d/user/secure-forward.conf as the sample config for 4.x

modules/efk-logging-fluentd-external.adoc
```diff
@@ -51,10 +71,46 @@ $ oc edit configmap/fluentd -n openshift-logging
 ----
 <1> Enter the name, host, and port for your external Fluentd server.
 
+. Add the path to your certificate and private key to the `secure-forward.conf` section:
++
+[source,yaml]
+----
+  secure-forward.conf: |
+    # <store>
+    # @type secure_forward
```
richm: This config is for 3.11.  Use https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/configs.d/user/secure-forward.conf as the sample config for 4.x

richm:
Also, the final section starting with "The following is an example of a in_forward configuration for Fluentd 0.12:" - this is for _output_ not _input_. I think we should just remove this section entirely, and let the user use the external fluentd documentation for reference.  If we need to have an example config then use
```
<source>
  @type forward

  <security>
    self_hostname ${HOSTNAME}
    shared_key the-shared-key-used-with-client
  </security>
  bind 0.0.0.0
  port 24284
</source>
```
for fluentd 1.x and
```
<source>
  @type secure_forward

  self_hostname ${HOSTNAME}
  bind 0.0.0.0
  port 24284

  shared_key the-shared-key-used-with-client

  secure no
</source>
```
for fluentd 0.12

modules/efk-logging-fluentd-external.adoc
mburke5678: The sample config in the docs is what is in the 4.2 Fluentd configmap. The user should replace the default `secure-forward.conf` in the configmap with the config in the link?

richm: > The sample config in the docs is what is in the 4.2 Fluentd configmap

Ah, ok.  That is wrong.  We need to fix that.   https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/configs.d/user/secure-forward.conf is what should be used there instead.

mburke5678:
@anpingli Is this an issue you can review?

anpingli:
I'd like use the external fluentd documentation for reference https://docs.fluentd.org/output/forward

mburke5678:
I am proposing PR https://github.com/openshift/openshift-docs/pull/17558 in favor of the current PR, as I think I have a better understanding of how to use the forward plugins. PTAL

PR #17349, opened 2019-10-16
sync _template

vikram-redhat:


vikram-redhat:
git cherrypick enterprise-3.10

PR #17363, opened 2019-10-16
Added Metering docs for OpenShift Serverless

abrennan89:


serverless/metering-serverless.adoc
```diff
@@ -0,0 +1,62 @@
...
+You can then check the report as shown in the following example:
+```
+$ kubectl get report
+
+NAME                        QUERY                       SCHEDULE   RUNNING    FAILED   LAST REPORT TIME       AGE
+knative-service-cpu-usage   knative-service-cpu-usage              Finished            2019-06-30T23:59:59Z   10h
+```
+
+// Viewing metering reports
+include::modules/metering-viewing-report-results.adoc[leveloffset=+1]
```
mgencur: The preview has one occurence of `{product-tile} ` which doesn't resolve (you can see it when looking at the preview)

mgencur: The last example output in the included file will look differently for Knative. Please look at the tutorials from @rhuss and see they're a little different.

serverless/metering-serverless.adoc
```diff
@@ -0,0 +1,62 @@
...
+The following `ReportDataSources` are available for use with Metering and the {ServerlessProductName} Knative Serving component:
+
+`knative-service-cpu-usage`:: Accumulated CPU seconds used per Knative service over the report time period.
+`knative-service-memory-usage`:: Average Memory consumption per Knative service over the report time period.
+
+These `ReportDataSources` are found in the `datasources` directory.
+You can apply them by using the following command:
+
+```
+$ oc apply -f datasources/
```
mgencur: Where does the datasources directory come from? We should mention that users should checkout certain repo. But maybe it would be better to include the example yaml files directly (not a whole directory).

mgencur (commented):
hi @abrennan89 it's on a good way. We just need to polish a few things. Added some comments below.

serverless/metering-serverless.adoc
abrennan89: I will just rewrite the content in this case then, if our version is different we can't really re-use the content.

serverless/metering-serverless.adoc
abrennan89: @rhuss would this be the https://github.com/rhuss/knative-reports repo?
Are the two yaml files here the only two expected for now?

serverless/metering-serverless.adoc
abrennan89: Should be fixed now @mgencur - I removed this section and replaced it

serverless/metering-serverless.adoc
abrennan89: Used this repo and called out the yaml files for now - please lmk if there's a different repo to point to or any further changes. Thanks.

markusthoemmes: I feel like we should inline them. Or if we don't inline them we should move the repo to somewhere more "official" like the openshift-knative organisation.

mgencur: I must agree with @markusthoemmes , usually the examples are included in our docs directly, not as reference to external github repos.

abrennan89: That makes sense, thanks for the input!

abrennan89: I think I misunderstood you originally @mgencur sorry about that.
Should I just add a step to say "create X yaml file by copying the sample" and then include the whole file for each yaml?
I'd imagine we need to do the same for `queries` then too :)

mgencur: Hmm, I'd leave out any words about "copying" anything, we can just include every single yaml file in our docs, including `queries` . Overall, there are not too many of them.

abrennan89:
@mgencur @markusthoemmes thanks for the feedback, you should see the change inc. the yaml files now (preview: https://srvks-204--ocpdocs.netlify.com/openshift-enterprise/latest/serverless/serverless-metering.html) - is there anything else we should add / change? Thanks.

modules/serverless-metering-datasources.adoc
```diff
@@ -0,0 +1,71 @@
+// Module included in the following assemblies:
+// serverless-metering.adoc
+
+[id="datasources-metering-serverless_{context}"]
+= Datasources for Knative Serving metering
+The following `ReportDataSources` are available for use with metering and the {ServerlessProductName} Knative Serving component.
```
markusthoemmes: instead of "are available" maybe we should signify that these are examples of how it can be used with metering?

modules/serverless-metering-datasources.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+[id="knative-service-cpu-usage-ds_{context}"]
+== knative-service-cpu-usage
+This datasource provides the accumulated CPU seconds used per Knative service over the report time period.
+
+.Sample yaml
```
markusthoemmes:
```suggestion
.YAML
```

modules/serverless-metering-datasources.adoc
```diff
@@ -0,0 +1,71 @@
...
+            group_left(label_serving_knative_dev_service, label_serving_knative_dev_revision)
+            kube_pod_labels{label_serving_knative_dev_service!=""}
+          )
+----
+
+[id="knative-service-memory-usage-ds_{context}"]
+== knative-service-memory-usage
+This datasource provides the average Memory consumption per Knative service over the report time period.
+
+.Sample yaml
```
markusthoemmes:
```suggestion
.YAML
```

modules/serverless-metering-queries.adoc
```diff
@@ -0,0 +1,127 @@
+// Module included in the following assemblies:
+// serverless-metering.adoc
+
+[id="queries-metering-serverless_{context}"]
+= Queries for Knative Serving metering
+The following `ReportQuery` resources are available for use with metering and the {ServerlessProductName} Knative Serving component.
```
markusthoemmes: We might want to signify here that these reference the example DataSources from above. In fact I'm wondering if we can fold both ReportQueries into just one object querying both CPU and RAM. @rhuss maybe knows. The two ReportQueries are almost identical so that might seem confusing.

serverless/serverless-metering.adoc
```diff
@@ -0,0 +1,71 @@
...
+For information about installing metering on {product-title}, see xref:../metering/metering-installing-metering.adoc[Installing Metering].
+
+include::modules/serverless-metering-datasources.adoc[leveloffset=+1]
+include::modules/serverless-metering-queries.adoc[leveloffset=+1]
+
+[id="reports-metering-serverless_{context}"]
+== Running a metering report for Knative Serving
+You can run metering reports against Knative Serving by creating `Report` resources.
+You can find example reports located in the `reports` directory.
+Before you run a report, you must modify the input parameter within the `Report` resource to specify the start and end dates of the reporting period.
```
markusthoemmes: Do we need to inline this "Report" resource as well? This seems to still reference the repository.

serverless/serverless-metering.adoc
```diff
@@ -0,0 +1,71 @@
...
+You can then check the report as shown in the following example:
+----
+$ kubectl get report
+
+NAME                        QUERY                       SCHEDULE   RUNNING    FAILED   LAST REPORT TIME       AGE
+knative-service-cpu-usage   knative-service-cpu-usage              Finished            2019-06-30T23:59:59Z   10h
+----
+
+[id="reports-accessing-serverless_{context}"]
+== Accessing a metering report for Knative Serving
```
markusthoemmes: I'm not sure this should be in our docs. Explaining to users how to get to certain reports isn't the job of our docs I think.

modules/serverless-metering-queries.adoc
rhuss: I don't think its possible as the 'value' column holding the report metrics (cpu or storage) has different types (`bytes_seconds` for memory, `cpu_core_seconds` for CPU) and I don't know if this can be parameterized as input parameters for the report itself.

And even if it could, it puts then more burden on the user to provide these parameters in the actual `Report` instance which calls this `ReportQuery`. So in this case, I think the duplication is ok.

modules/serverless-metering-reports.adoc
```diff
@@ -0,0 +1,54 @@
...
+
+NAME                        QUERY                       SCHEDULE   RUNNING    FAILED   LAST REPORT TIME       AGE
+knative-service-cpu-usage   knative-service-cpu-usage              Finished            2019-06-30T23:59:59Z   10h
+----
+
+
+[id="knative-service-cpu-usage-report_{context}"]
+== knative-service-cpu-usage-report
+
+.YAML file
```
markusthoemmes: Should these move further up possibly? I would've expected them below the paragraph explaining the parameters and before applying the YAML.

modules/serverless-metering-reports.adoc
```diff
@@ -0,0 +1,55 @@
...
+[source, yaml]
+----
+apiVersion: metering.openshift.io/v1alpha1
+kind: Report
+metadata:
+  name: knative-service-cpu-usage
+spec:
+  reportingStart: '2019-06-01T00:00:00Z'
```
rhuss:
```suggestion
  reportingStart: '2019-06-01T00:00:00Z' <1>
```

modules/serverless-metering-reports.adoc
```diff
@@ -0,0 +1,55 @@
...
+  reportingEnd: '2019-06-30T23:59:59Z'
```
rhuss:
```suggestion
  reportingEnd: '2019-06-30T23:59:59Z' <2>
```

modules/serverless-metering-reports.adoc
```diff
@@ -0,0 +1,55 @@
...
+  query: knative-service-cpu-usage
```
rhuss:
```suggestion
  query: knative-service-cpu-usage <3>
```

modules/serverless-metering-reports.adoc
```diff
@@ -0,0 +1,55 @@
...
+runImmediately: true
+----
```
rhuss:
```suggestion
----
<1> Start date of the report, in ISO 8601 format
<2> End date of the report, in ISO 8601 format
<3> Either `knative-service-cpu-usage` for CPU usage report or `knative-service-memory-usage` for a memory usage report
```

rhuss: So, I would merge the call to the report to one section, and removing the second level headers (which imo should have been third level headers anyway). The only difference is explained in the `<3>` explanation.

rhuss (commented):
@abrennan89 I made some suggestions and actually I would merge the section where to apply reports to one (see the callout <3> in the suggestion)

abrennan89:
@rhuss @markusthoemmes I think that should be all the changes now. Lmk if there's anything else and thank you for all the feedback :)

mgencur:
hi @abrennan89 , it looks good to me! Just one nit - we should use the openshift-metering namespace instead of metering. So we might need to remove the note there about using `metering` namespace.

mgencur:
LGTM. QE Approved

sheriff-rh (approved):
LGTM!

vikram-redhat:
@abrennan89 @sheriff-rh will this need to be CPed to 4.3 as well?

serverless/serverless-metering.adoc
```diff
@@ -0,0 +1,17 @@
+[id="metering-serverless"]
+= Using metering with {ServerlessProductName}
+include::modules/common-attributes.adoc[]
+include::modules/serverless-document-attributes.adoc[]
+:context: metering-serverless
+
+As an {product-title} cluster administrator, you can use metering to analyze what is happening in your {ServerlessProductName} cluster.
+
+For more information about metering on {product-title}, see xref:../metering/metering-about-metering.adoc[About metering].
```
vikram-redhat: @abrennan89 missing anchor id. All xrefs must have the anchor id as well.

serverless/serverless-metering.adoc
```diff
@@ -0,0 +1,17 @@
...
+
+[id="installing-metering-serverless_{context}"]
+== Installing metering
+For information about installing metering on {product-title}, see xref:../metering/metering-installing-metering.adoc[Installing Metering].
```
vikram-redhat: Same - missing anchor id.

abrennan89:
Apologies @vikram-redhat ! Yes this should be for 4.3 also I'd imagine until we hear otherwise. Also, I've added an additional PR #17890 to fix the anchor IDs.
CC @sheriff-rh

PR #17373, opened 2019-10-16
updates for BZ1746768

lamek:
Replacing `oc-apply` with `oc-replace` when adding an identity provider.

### DIFF
--- a/modules/identity-provider-add.adoc
+++ b/modules/identity-provider-add.adoc
@@ -31,6 +31,11 @@ users can authenticate.
 ----
 $ oc apply -f </path/to/CR>
 ----
++
+[NOTE]
+====
+If a CR does not exist, `oc apply` creates a new CR and might trigger the following warning: `Warning: oc apply should be used on resources created by either oc create --save-config or oc apply`. In this case you can safely ignore this warning.
+====
 
 . Log in to the cluster as a user from your identity provider, entering the
 password when prompted.


modules/identity-provider-add.adoc
```diff
@@ -29,7 +29,7 @@ users can authenticate.
 . Apply the defined CR:
 +
 ----
-$ oc apply -f </path/to/CR>
+$ oc replace-f </path/to/CR>
```
bergerhoffer: Need a space between "replace" and "-f"

bergerhoffer (commented):
Just one thing

modules/identity-provider-add.adoc
lamek: :running_woman: So fast. Thank you!

zhouying7780:
It's better to still use `oc apply` instead of `oc replace` , because `oc apply` will create the resource when that not exist, but `oc replace` does not have this function and will  meet error.
We could add a tip to explain the "warning message" is correct. 
When resource created by `oc create --save-config` or `oc apply` we will save resource config in the annotations, when use `oc apply` with resource without the annotations, then will meet the "warning message".

lamek:
@zhouying7780 Thanks. I updated this change to be a note per your suggestion. Could you PTAL once more and :+1: if you're good with this?
Link to preview - https://bz1746768-oc-apply-to-oc-replace--ocpdocs.netlify.com/openshift-enterprise/latest/authentication/identity_providers/configuring-htpasswd-identity-provider.html#add-identity-provider_configuring-htpasswd-identity-provider

zhouying7780:
LGTM , thanks

PR #17384, opened 2019-10-16
closing more ifevals

kalexand-rh:
@vikram-redhat, PTAL?


jboxman:
@kalexand-rh, for the network modules I see you're unsetting the specific value. Why not unset the variable entirely? It's less verbose.

https://asciidoctor.org/docs/user-manual/#using-attributes-set-assign-and-reference

kalexand-rh:
@jboxman, because checking to see **what** the variables were wasn't part of my process - I was looking for modules that defined and then didn't undefine them. I've updated them all. Thanks

vikram-redhat:
Thanks @kalexand-rh.

PR #17323, opened 2019-10-15
Indefinite article change for RHOSP abbreviation

maxwelldb:
Additionally:

* Adding OSP to common-atributes

Context: https://github.com/openshift/openshift-docs/pull/17304

For 4.2.

PR #17388, opened 2019-10-16
Revoke "RHOSP" initialism treatment

maxwelldb:
4.2+

Context: 

* https://github.com/openshift/openshift-docs/pull/17304
* https://github.com/openshift/openshift-docs/pull/17323

From the OSP DPM today:

> We had a larger conversation around RHOSP this morning, with overwhelming support for "a RHOSP...".
> 
> I'll send a note to the writer team to ensure we are consistent when we write RHOSP. We do use the full name a lot...

@bergerhoffer Would you like to take this, since you're pretty familiar with the issue at this point? :)

### DIFF
--- a/modules/installation-initializing.adoc
+++ b/modules/installation-initializing.adoc
@@ -120,7 +120,7 @@ ifdef::osp[]
 .. Select *openstack* as the platform to target.
 .. Specify the {rh-openstack-first} external network name to use for installing the cluster.
 .. Specify the Floating IP address to use for external access to the OpenShift API.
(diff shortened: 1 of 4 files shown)


maxwelldb:
@bobfuru Actually, you have the background on this, too, if you want it.

bergerhoffer (commented):
Sorry for the delay @maxwelldb! As much as I'd prefer "an", I'm fine with honoring the team's decision. Thanks, this LGTM!

maxwelldb:
Thanks!

PR #16890, opened 2019-09-26
[WIP] Ocp 3 4 migration

apinnick:
Formatted doc preview: 
http://file.tlv.redhat.com/~apinnick/ocp_3_4_migration/migration/migrating-openshift-3-to-4.html

Google docs draft:
https://docs.google.com/document/d/1MU_AanBKvEAUUhVku9s1CPX0UrZmdVXJURN6Tn-eCgE/edit#heading=h.i7jdrb4rw5qy

PR #17415, opened 2019-10-16
OCP 3 to 4 migration

ahardin-rh:
Continuation of https://github.com/openshift/openshift-docs/pull/16890 t be released in 4.2.z stream next week


modules/migration-adding-replication-repository-to-cam.adoc
```diff
@@ -0,0 +1,28 @@
...
+.Procedure
+
+. In the *Replication repositories* section, click *Add replication repository*.
+
+. In the *Replication repository* window, fill in the following fields:
+
+* *Replication repository name*
+* *S3 bucket name*
+* *S3 bucket region*: Required for AWS S3 if the bucket region is not *us-east-1*. Optional for a generic S3 repository.
+* *S3 endpoint*: Required for a generic S3 repository. This is the URL of the S3 service, not the bucket, for example, `http://_<minio-gpte-minio.apps.cluster.com>_`.
```
jwmatthews: We will want to update this to reference an example NooBaa endpoint instead of minio

I am working on other updates to show an example of how to setup the S3 pre-reqs.
Will share later today.

migration/migrating-openshift-3-to-4.adoc
```diff
@@ -0,0 +1,92 @@
...
+You can migrate data to any storage class that is available on the target cluster, for example, from Red Hat Gluster Storage or NFS storage to Red Hat Ceph Storage.
+
+Optionally, you can migrate the {product-title} 3.7 (and later) control plane settings to {product-title} 4.x with the Control Plane Migration Assistant (CPMA). See xref:migration-understanding-cpma_{context}[].
+
+.Prerequisites
+
+* You must have `cluster-admin` privileges.
+* You must have `podman` installed.
+* The source cluster(s) must be {product-title} 3.7 or later.
+* The target cluster must be {product-title} 3.9 or later.
```
jwmatthews: Suggest we change to:

"The target cluster is recommended to be the latest released version of {product-title}."


For background...
This will technically work on 3.7 or later for both source and destination.
Recommendation is customer migrate to latest released OpenShift for the destination.

migration/migrating-openshift-3-to-4.adoc
```diff
@@ -0,0 +1,92 @@
...
+* The replication repository object storage must support the S3 API, for example, Red Hat NooBaa Storage, AWS S3, and generic S3 storage providers.
```
jwmatthews: I assume we need to change "Red Hat NooBaa Storage".

Upstream:  NooBaa
Downstream: MCG

For this immediate release we will use upstream NooBaa as an example, after OCS4 releases we will change to leverage an operator for MCG.

I will ask product management for their recommendation of what to call this now.

jwmatthews: Confirmed with product management.

Let's change to:

"The replication repository is an object store endpoint that needs to support the S3 API, for example, NooBaa, AWS S3, and generic S3 storage providers."

migration/migrating-openshift-3-to-4.adoc
```diff
@@ -0,0 +1,92 @@
...
+
+You can migrate data to any storage class that is available on the target cluster, for example, from Red Hat Gluster Storage or NFS storage to Red Hat Ceph Storage.
+
+Optionally, you can migrate the {product-title} 3.7 (and later) control plane settings to {product-title} 4.x with the Control Plane Migration Assistant (CPMA). See xref:migration-understanding-cpma_{context}[].
+
+.Prerequisites
+
+* You must have `cluster-admin` privileges.
+* You must have `podman` installed.
+* The source cluster(s) must be {product-title} 3.7 or later.
```
jwmatthews: Add an additional prerequisite

"The migration controller must be able to communicate with both the source and destination clusters, we recommend the migration controller be installed on the destination cluster."

modules/migration-configuring-cors-3.adoc
```diff
@@ -0,0 +1,47 @@
...
+To access the API server of an {product-title} 3 cluster from a web application (in this case, the Cluster Application Migration tool) on a different host, you must configure cross-origin resource sharing by adding the CAM host name to the master configuration file.
+
+.Procedure
+
+. Log in to the {product-title} 3 cluster.
+. Add the CAM host name to the `corsAllowedOrigins` stanza in the *_/etc/origin/master/master-config.yaml_* configuration file:
++
+----
+corsAllowedOrigins:
+- (?i)//migration-openshift-migration\.apps\.cluster\.com(:|\z)
```
eriknelson: I saw some folks very confused by this, thinking literally `(?i)//migration-openshift-migration\.apps\.cluster\.com(:|\z)` is the string that they should use.

I think prior to "Configuring cross-origin resource sharing on OpenShift Container Platform 3 clusters", we need a section that explains how to get the UI's Origin, and we need to make it clear that the docs will continue with `(?i)//migration-openshift-migration\.apps\.cluster\.com(:|\z)` as an example Origin, but this value will be different for every user.

```md
You enable communication with the CAM by configuring cross-origin resource sharing on OpenShift Container Platform 3 and 4 (if the OpenShift Container Platform 4 cluster is not hosting the CAM).

# Obtaining the UI's Origin
First, you must obtain the UI's Origin value to be whitelisted across all of the clusters involved in a migration.

1) You must log into the cluster that is hosting the web UI.
2) The following command will export the encoded `corsAllowedOrigin` for your cluster: `oc get -n openshift-migration route/migration -o go-template='(?i}//{{ .spec.host }}(:|\z){{ println }}' | sed 's,\.,\\.,g'`
3) This output string should be used where an allowed origin must be whitelisted. As an example, we'll use `(?i)//migration-openshift-migration\.apps\.cluster\.com(:|\z)`, but this will differ for your cluster.

```

eriknelson (changes requested):

migration/migrating-openshift-3-to-4.adoc
sferich888: I suggest we leave this as is. We should treat these situations with a support exception. 
This will be confusing if you look at this with the lifecycle.

migration/migrating-openshift-3-to-4.adoc
sferich888: Do we need to call this out as "upstream" and possibly include a note / call out (warning) about support not being provided for upstream projects?

sferich888 (changes requested):

migration/migrating-openshift-3-to-4.adoc
jwmatthews: @sferich888 when you say leave as-is, I just want to confirm.

Right now doc says:  "The target cluster must be {product-title} 3.9 or later."

I think it should say:
  "The target cluster is recommended to be the latest released version of {product-title}."


Then we can address 3->3 migrations on a case-by-case basis.  Do you agree?

XinRedhat:
Reviewed, LGTM.

apinnick:
I want to make some updates. Can I close this PR and create my own branchy? Can the comments be resolved?

modules/migration-adding-replication-repository-to-cam.adoc
apinnick: Changed URL to `http://s3-noobaa.apps.cluster.com`

migration/migrating-openshift-3-to-4.adoc
apinnick: Changed `Red Hat NooBaa Storage` to `NooBaa`.

migration/migrating-openshift-3-to-4.adoc
apinnick: That's a recommendation, not a prerequisite. The default installation is on the target cluster, so I'm not sure it's necessary to mention this. The user would have to go out of their way to install it on another cluster.

modules/migration-configuring-cors-3.adoc
apinnick: I've added the steps for obtaining the CAM host URL and marked `(?i)//migration-openshift-migration\.apps\.cluster\.com(:|\z)` with a callout saying that it needs to be replaced with the user's CAM host URL.

apinnick:
New doc preview: http://file.tlv.redhat.com/~apinnick/ocp-migration-new/migration/migrating-openshift-3-to-4.html

New PR: https://github.com/openshift/openshift-docs/pull/17516

ahardin-rh:
@apinnick Looks like you addressed the comments in your new PR, so I will close this PR in favor if your new one.

PR #17424, opened 2019-10-17
bug 1731441 must keep cluster running for 24 hours

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1731441

This note should appear in all of the installation assemblies after the installation program finishes or you run the `wait-for install-complete` command.

@gpei, will you PTAL?


kalexand-rh:
I'm applying an existing note in more places, and this change was approved on the bug. Merging.

PR #17430, opened 2019-10-17
add example selector syntax

jeremyeder:


ahardin-rh:
@jeremyeder Thank you!

PR #17451, opened 2019-10-18
modules/rhcos-about.adoc: fix links related to RHCOS upgrades

miabbott:
We don't want users going to projectatomic.io, so just link directly
to the `rpm-ostree` GitHub page instead.

Add link to MCO docs about how OS upgrades are handled.

### DIFF
--- a/modules/rhcos-about.adoc
+++ b/modules/rhcos-about.adoc
@@ -65,10 +65,8 @@ images. You can use the crictl CLI tool to work with containers and pods from t
 CRI-O container engine. While direct use of these tools in {op-system} is
 discouraged, you can use them for debugging purposes.
 
-{op-system} features transactional upgrades and rollbacks using the
-http://www.projectatomic.io/docs/os-updates/[rpm]
-http://www.projectatomic.io/docs/os-updates/[-ostree]
-upgrade system. Updates are delivered via container images and are part of the
+{op-system} features transactional upgrades using the `rpm-ostree` system.
+ Updates are delivered via container images and are part of the
  OpenShift update process. When deployed, the container image is pulled,
  extracted, and written to disk, then the bootloader is modified to boot into
  the new version. The machine will reboot into the update in a rolling manner to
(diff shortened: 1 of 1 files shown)


modules/rhcos-about.adoc
```diff
@@ -66,9 +66,8 @@ CRI-O container engine. While direct use of these tools in {op-system} is
 discouraged, you can use them for debugging purposes.
 
 {op-system} features transactional upgrades and rollbacks using the
-http://www.projectatomic.io/docs/os-updates/[rpm]
-http://www.projectatomic.io/docs/os-updates/[-ostree]
-upgrade system. Updates are delivered via container images and are part of the
+ https://github.com/coreos/rpm-ostree[rpm-ostree]
```
vikram-redhat: @sferich888 ok with this link?

sferich888:
We don't want people rolling back updates.
This in principal should not be included.

vikram-redhat:
@miabbott if you could remove that link and rephrase based on @sferich888's comment, I will merge it.

modules/rhcos-about.adoc
```diff
@@ -84,8 +83,9 @@ For {op-system} systems, the layout of the `rpm-ostree` file system has the
 * `/var/lib/containers` is the graph storage location for storing container
  images.
 
-In {product-title}, the Machine Config Operator handles operating
-system upgrades. Instead of upgrading individual packages, as is done with yum
+In {product-title}, the Machine Config Operator handles
```
kalexand-rh: @sferich888, will you please review this link as well?

kalexand-rh: @vikram-redhat, you probably shouldn't merge it with the second link. It would be better to port the content and get it approved in context.

miabbott:
Pushed a new commit with the links removed and touched up some of the text around upgrades

kalexand-rh:
@vikram-redhat, I think this looks good. Do you need any further review before I merge?

vikram-redhat:
@kalexand-rh merge away.

PR #17452, opened 2019-10-18
Note that elasticsearch CPU request should be set appropriately for heavy log loads

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1759318

Describe the issue: 
It should be noted in the CPU/memory limit section that ElasticSearch uses the number of availalbe processors to determine thread pool size at initialization, so you will need to set the CPU request to something that makes sense if you require more logging performance.
This means setting the elasticsearch pod CPU request low but the limit high will not give expected results.
Ex. set CPU request to 1 but limit to 32, the cpu usage on those pods will never get that high because Elasticsearch only created threads for 1 CPU and performance will be essentially the same as 1 cpu request and limit.

Suggestions for improvement:
Make a note of this for anyone that requires more performant logging.

### DIFF
--- a/modules/cluster-logging-elasticsearch-limits.adoc
+++ b/modules/cluster-logging-elasticsearch-limits.adoc
@@ -37,8 +37,27 @@ spec:
             cpu: "4000m"
             memory: "4Gi"
           requests:
-            cpu: "100m"
+            cpu: "4000m"
             memory: "1Gi"
 ----
 <1> Specify the CPU and memory limits as needed. If you leave these values blank,
 the Elasticsearch Operator sets default values that should be sufficient for most deployments.
++
+If you adjust the amount of Elasticsearch CPU and memory, you must change both the request value and the limit value. 
++
+For example:
++
+[source,yaml]
+----
+      resources:
+        limits:
+          cpu: "8"
+          memory: "32Gi"
+        requests:
+          cpu: "8"
+          memory: "32Gi"
+----
++
+Kubernetes generally adheres the node CPU configuration and DOES not allow Elasticsearch to use the specified limits. 
+Setting the same value for the `requests` and `limits` ensures that Elasticseach can use the CPU and memory you want, assuming the node has the CPU and memory available. 
+


mburke5678:
@richm @jcantrill Can you help with this one? Above my pay grade. 
The request is for better wording around Elasticsearch CPU limits and requests for high performance logging.
Is this information correct?
Do we want to advertise the information in this way?

richm:
Yes.  The point is that, if you want to increase the amount of Elasticsearch CPU and memory, you must change both the request value and the limit value.  For example, if you set in your clusterlogging CR something like this:
```
...
      resources:
        limits:
          cpu: 8
          memory: 32Gi
        requests:
          cpu: 1
          memory: 16Gi
```
Thinking that Elasticsearch will use up to 8, it probably won't ever use more than 1, because Kubernetes will probably limit the number of cpus, depending on the node cpu budget.  So if you want to make sure that Elasticsearch will use all 8 cpus, you have to set the requests value the same as the limit value:
```
      resources:
        limits:
          cpu: 8
          memory: 32Gi
        requests:
          cpu: 8
          memory: 32Gi
```
That will tell Kubernetes that you are serious, that Elasticsearch _must_ have that much cpu and memory.  Of course, you will need to make sure the node actually has that much cpu and memory available.

richm:
lgtm

jcantrill (approved):
lgtm

mburke5678:
@anpingli Please take a look. Thank you!

FYI @RH-ematysek

anpingli (commented):
LGTM

mburke5678:

modules/cluster-logging-elasticsearch-limits.adoc
```diff
@@ -37,8 +37,28 @@ spec:
...
             memory: "4Gi"
           requests:
-            cpu: "100m"
+            cpu: "4000m"
             memory: "1Gi"
 ----
 <1> Specify the CPU and memory limits as needed. If you leave these values blank,
 the Elasticsearch Operator sets default values that should be sufficient for most deployments.
++
+If you need to adjust the amount of Elasticsearch CPU and memory, you must change both the request value and the limit value.
```
kalexand-rh: > ! need to v
Avoid using “need to” to refer to a requirement that a product or procedure imposes. Use “must” or, in
constructions in which “must” is not idiomatically correct, use “have to” or “required to.”

Maybe ~If you need~

modules/cluster-logging-elasticsearch-limits.adoc
```diff
@@ -37,8 +37,28 @@ spec:
...
 <1> Specify the CPU and memory limits as needed. If you leave these values blank,
 the Elasticsearch Operator sets default values that should be sufficient for most deployments.
++
+If you need to adjust the amount of Elasticsearch CPU and memory, you must change both the request value and the limit value. 
++
+For example:
++
+[source,yaml]
+----
+      resources:
```
kalexand-rh: Why is the syntax different here an in the previous YAML snippet? Can you omit the "" like this?

kalexand-rh (commented):
A pick and a question, then it LGTM

PR #17495, opened 2019-10-21
BZ-1762342: Adding how to obtain your cluster ID for Support

bergerhoffer:


bergerhoffer:
@sferich888 Can you review this to make sure the steps look okay for obtaining your cluster ID? Thanks!

* http://file.rdu.redhat.com/~ahoffer/2019/BZ-1762342-cluster-id/support/getting-support.html
* http://file.rdu.redhat.com/~ahoffer/2019/BZ-1762342-cluster-id/support/gathering-cluster-data.html

bergerhoffer:
@lamek @bobfuru For peer review please!

modules/support.adoc
```diff
@@ -17,9 +17,10 @@ ifdef::openshift-enterprise,openshift-dedicated[]
 
 [NOTE]
 ====
-When submitting a support case, it is often helpful to gather information about
-your cluster to provide to Red Hat Support to aid in troubleshooting. This data
-can be gathered using the `oc adm must-gather` command.
+When submitting a support case, it is recommended to provide the following information about your cluster to Red Hat Support to aid in troubleshooting:
```
bobfuru: I don't think you need "it is recommended to" here. Suggest removing.

bobfuru (commented):
One small suggestion and otherwise LGTM.

modules/support.adoc
bergerhoffer: I've gone back and forth on this, but one thing is that it's not always possible to do this (like if the installation didn't actually complete). So I'm not sure that I want to make it as definite as instructing them, without providing the caveats, so I think I'll leave this as is for now.

zhouying7780:
LGTM, thanks.

Issue #17472, opened 2019-10-21
[enterprise-4.2] Edit suggested in file logging/cluster-logging.adoc


mmusaji:
### Which section(s) is the issue in?
https://docs.openshift.com/container-platform/4.2/logging/cluster-logging.html#cluster-logging-about_cluster-logging

### What needs fixing?
3rd and 4th paragraphs are duplicates so one of them can be deleted.

bmillemathias:
looking to file modules/cluster-logging-about.adoc I see
```
{product-title} cluster administrators can deploy cluster logging using a few CLI commands and the {product-title}.
web console to install the Elasticsearch Operator and Cluster Logging Operator. When the operators are installed,.
create a Cluster Logging Custom Resource (CR) to schedule cluster logging pods and other resources necessary.
to support cluster logging. The operators are responsible for deploying, upgrading, and maintaining cluster logging.

ifndef::cnv-logging[]
{product-title} cluster administrators can deploy cluster logging using a few
CLI commands and the {product-title}  web console to install the Elasticsearch
Operator and Cluster Logging Operator. When the operators are installed, create
a Cluster Logging Custom Resource (CR) to schedule cluster logging pods and
other resources necessary to support cluster logging. The operators are
responsible for deploying, upgrading, and maintaining cluster logging.
```

I don't know adoc but there is so conditional include it seems.

mburke5678:
@mmusaji It appears that this problem, possibly due to erroneous conditionals, was corrected in the PR at @bmillemathias submitted. I will close this issue. 
If you still see the problem, feel free to reopen the issue and tag me.

PR #17500, opened 2019-10-21
Remove duplicate paragraph about logging

bmillemathias:
Fix issue #17472

### DIFF
--- a/modules/cluster-logging-about.adoc
+++ b/modules/cluster-logging-about.adoc
@@ -27,11 +27,6 @@ It collects all node and container logs and writes them to link:https://www.elas
 link:https://www.elastic.co/guide/en/kibana/current/introduction.html[Kibana] is the centralized, web UI 
 where users and administrators can create rich visualizations and dashboards with the aggregated data.
 
-{product-title} cluster administrators can deploy cluster logging using a few CLI commands and the {product-title} 
-web console to install the Elasticsearch Operator and Cluster Logging Operator. When the operators are installed, 
-create a Cluster Logging Custom Resource (CR) to schedule cluster logging pods and other resources necessary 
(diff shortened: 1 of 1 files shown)


vikram-redhat:
@mburke5678 @aburdenthehand - is this just an error due to forgetting to conditionalize the preceding text? Or that text does need removing?

We should also move this to master once we know.

jboxman:
I couldn't make heads nor tails of it. The duplicate text is deep within the second conditional, so it should _not_ appear stacked like it does. The only conclusion I can come to is that _Asciidoctor_ is injecting the first conditional with that para after the conditional with the second instance of the para; otherwise, the ordering in the rendered document is impossible to achieve.

So this may be a bug in _Asciidoctor_ itself.

I made no attempts to render it to confirm this hypothesis, however.

aburdenthehand:
@jboxman I'm inclined to agree. 
It's not an unresolved tag from an earlier module (tested by removing all prior content).
It seems to be the way asciibinder parses the conditionals. I thought it might be because there are two ifndef::cnv-logging[] in a single ifdef::openshift*[] conditional but I ruled that out by closing the idef::openshift conditional and opening a second one for this particular para - and the problem persisted. 

The best I've been able to get so far is to change the second ifndef::cnv-logging[] (the one surrounding the problem para) into an ifdef::!cnv-logging[] conditional. This stops the para repeating in the openshift logging assembly, but does not stop this para being built in the cnv assembly. 
Will come back to this in the afternoon.

aburdenthehand:
Thanks for the fix @bmillemathias!
Jason and I went off on a tangent not realising that this para was duplicated during a cherrypick to the 4.2 branch, to which you had already provided the answer. There is no weirdness here, just a bad CP from master.

PR #17513, opened 2019-10-22
BZ-1731015: Adding docs for replacing a single etcd member

bergerhoffer:
Preview: http://file.rdu.redhat.com/~ahoffer/2019/BZ-1731015/backup_and_restore/replacing-failed-master.html

### DIFF
--- a/backup_and_restore/replacing-failed-master.adoc
+++ b/backup_and_restore/replacing-failed-master.adoc
@@ -1,0 +1,27 @@
+[id="replacing-failed-master-host"]
+= Replacing a failed master host
+include::modules/common-attributes.adoc[]
+:context: replacing-failed-master-host
+
+toc::[]
+
+This document describes the process to replace a single etcd member. This procedure assumes that there is still an etcd quorum in the cluster.
+
+[NOTE]
+====
+If you have lost the majority of your master hosts, leading to etcd quorum loss, then you must follow the disaster recovery procedure to xref:../backup_and_restore/disaster_recovery/scenario-1-infra-recovery.adoc#dr-infrastructure-recovery[recover from lost master hosts] instead of this procedure.
+
+If the TLS certificates are not valid on the member being replaced, then you must follow the procedure to xref:../backup_and_restore/disaster_recovery/scenario-3-expired-certs.adoc#dr-recovering-expired-certs[recover from expired control plane certificates] instead of this procedure.
+====
+
+To replace a single master host:
+
+* xref:../backup_and_restore/replacing-failed-master.adoc#restore-remove-failed-master_replacing-failed-master-host[Remove the member from the etcd cluster].
+
+* xref:../backup_and_restore/replacing-failed-master.adoc#restore-add-master_replacing-failed-master-host[Add the member back to the etcd cluster].
+
+// Removing a failed master host from the cluster
+include::modules/restore-remove-failed-master.adoc[leveloffset=+1]
+
+// Adding a master host back to the cluster
+include::modules/restore-add-master.adoc[leveloffset=+1]
--- a/modules/dr-recover-lost-control-plane-hosts.adoc
+++ b/modules/dr-recover-lost-control-plane-hosts.adoc
@@ -5,27 +5,17 @@
 [id="dr-scenario-1-recover-master-hosts_{context}"]
 = Recovering from lost master hosts
 
-////
-TODO:
- - Add a final result
-////
-
-Follow these steps to recover from the loss of one or more master hosts.
+Follow these steps to recover from the loss of the majority of master hosts, leading to etcd quorum loss.
 
 .Prerequisites
 
 * Access to the cluster as a user with the `cluster-admin` role.
 * SSH access to a remaining master host.
-* A backed up etcd snapshot, if you are recovering a loss of a majority of masters.
+* A backed-up etcd snapshot.
 
 .Procedure
 
 . Restore etcd quorum on the remaining master.
-+
-[NOTE]
-====
-This step is only necessary if you have had a majority of your masters fail. You can skip this step if you have a majority of your masters still available.
-====
 
 .. Copy the etcd snapshot file to the remaining master host.
 +
@@ -290,7 +280,7 @@ sh-4.2#  etcdctl member list -w table
 +------------------+---------+------------------------------------------+----------------------------------------------------------------+---------------------------+
 ----
 +
-Note that it may take up to 10 minutes for the new member to start.
+It may take up to 10 minutes for the new member to start.
 
 ... Repeat these steps to add your other new master host until you have achieved full etcd membership.
 
@@ -301,5 +291,3 @@ In a terminal that has access to the cluster, run the following command:
 ----
 $ oc delete pod -n openshift-config etcd-signer
 ----
-
-// TODO: Add a final statement of what the result of all this should be.
--- a/modules/restore-add-master.adoc
+++ b/modules/restore-add-master.adoc
@@ -1,0 +1,99 @@
+// Module included in the following assemblies:
+//
+// * backup_and_restore/replacing-failed-master.adoc
+
+[id="restore-add-master_{context}"]
+= Adding a master host back to the etcd cluster
+
+Follow these steps to add a master host back to the etcd cluster. This procedure assumes that you previously removed the master host from the cluster and that its etcd dependencies, such as TLS certificates and DNS, are valid.
+
+.Prerequisites
+
+* Access to the cluster as a user with the `cluster-admin` role.
+* SSH access to the master host to add to the etcd cluster.
+* The IP address of an existing active etcd member.
+
+.Procedure
+
+. Access the master host to add to the etcd cluster.
++
+[IMPORTANT]
+====
+You must run this procedure on the master host that is being added to the etcd cluster.
+====
+
+. Run the `etcd-member-add.sh` script and pass in two parameters:
+
+** the IP address of an existing etcd member
+** the name of the etcd member to add
+
++
+----
+[core@ip-10-0-147-172 ~]$ sudo -E etcd-member-add.sh 10.0.128.73 etcd-member-ip-10-0-147-172.us-east-2.compute.internal
+Downloading etcdctl binary..
+etcdctl version: 3.3.10
+API version: 3.3
+etcd-member.yaml found in ./assets/backup/
+etcd.conf backup upready exists ./assets/backup/etcd.conf
+Stopping etcd..
+Waiting for etcd-member to stop
+etcd data-dir backup found ./assets/backup/etcd..
+Updating etcd membership..
+Removing etcd data_dir /var/lib/etcd..
+
+ETCD_NAME="etcd-member-ip-10-0-147-172.us-east-2.compute.internal"
+ETCD_INITIAL_CLUSTER="etcd-member-ip-10-0-147-172.us-east-2.compute.internal=https://etcd-1.clustername.devcluster.openshift.com:2380,etcd-member-ip-10-0-171-108.us-east-2.compute.internal=https://etcd-2.clustername.devcluster.openshift.com:2380,etcd-member-ip-10-0-128-73.us-east-2.compute.internal=https://etcd-0.clustername.devcluster.openshift.com:2380"
+ETCD_INITIAL_ADVERTISE_PEER_URLS="https://etcd-1.clustername.devcluster.openshift.com:2380"
+ETCD_INITIAL_CLUSTER_STATE="existing"'
+Member  1e42c7070decd39 added to cluster 6e25bab1bb556673
+Starting etcd..
+----
+
+. Verify that the new member is in the list of Pods associated with etcd and that its status is `Running`:
++
+----
+[core@ip-10-0-147-172 ~]$ oc get pods -n openshift-etcd
+NAME                                                     READY   STATUS    RESTARTS   AGE
+etcd-member-ip-10-0-128-73.us-east-2.compute.internal    2/2     Running   0          15h
+etcd-member-ip-10-0-147-172.us-east-2.compute.internal   2/2     Running   7          122m
+etcd-member-ip-10-0-171-108.us-east-2.compute.internal   2/2     Running   0          15h
+----
+
+. Verify that the etcd member has been successfully added to the etcd cluster:
+
+.. Connect to the running etcd container:
++
+----
+[core@ip-10-0-147-172 ~] id=$(sudo crictl ps --name etcd-member | awk 'FNR==2{ print $1}') && sudo crictl exec -it $id /bin/sh
+----
+
+.. In the etcd container, export the variables needed for connecting to etcd:
++
+----
+sh-4.2# export ETCDCTL_API=3 ETCDCTL_CACERT=/etc/ssl/etcd/ca.crt ETCDCTL_CERT=$(find /etc/ssl/ -name *peer*crt) ETCDCTL_KEY=$(find /etc/ssl/ -name *peer*key)
+----
+
+.. In the etcd container, execute `etcdctl member list` and verify that the new member is listed:
++
+----
+sh-4.2#  etcdctl member list -w table
+
++------------------+---------+------------------------------------------+------------------------------------------------------------------+---------------------------+
+|        ID        | STATUS  |                   NAME                   |                            PEER ADDRS                            |       CLIENT ADDRS        |
++------------------+---------+------------------------------------------+------------------------------------------------------------------+---------------------------+
+| 29e461db6be4eaaa | started | etcd-member-ip-10-0-128-73.us-east-2.compute.internal | https://etcd-2.clustername.devcluster.openshift.com:2380 | https://10.0.128.73:2379 |
+|  cbe982c74cbb42f | started | etcd-member-ip-10-0-147-172.us-east-2.compute.internal | https://etcd-0.clustername.devcluster.openshift.com:2380 | https://10.0.147.172:2379 |
+| a752f80bcb0da3e8 | started |   etcd-member-ip-10-0-171-108.us-east-2.compute.internal | https://etcd-1.clustername.devcluster.openshift.com:2380 |   https://10.0.171.108:2379 |
++------------------+---------+------------------------------------------+------------------------------------------------------------------+---------------------------+
+----
++
+It may take up to 10 minutes for the new member to start.
+
+.. In the etcd container, execute `etcdctl endpoint health` and verify that the new member is healthy:
++
+----
+sh-4.2# etcdctl endpoint health --cluster
+https://10.0.128.73:2379 is healthy: successfully committed proposal: took = 4.5576ms
+https://10.0.147.172:2379 is healthy: successfully committed proposal: took = 5.1521ms
+https://10.0.171.108:2379 is healthy: successfully committed proposal: took = 4.2631ms
+----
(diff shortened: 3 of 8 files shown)


modules/restore-add-master.adoc
```diff
@@ -0,0 +1,77 @@
...
+.Procedure
+
+. Access the master host to add to the etcd cluster.
+
+. Run the `etcd-member-add.sh` script.
++
+Pass in two parameters to the `etcd-member-add.sh` script: IP address of an existing etcd member and the name of the etcd member to add.
++
+----
+[core@ip-10-0-128-73 ~]$ sudo -E etcd-member-add.sh 10.0.128.73 \ etcd-member-ip-10-0-147-172.us-east-2.compute.internal
```
bergerhoffer: @retroflexer Can you verify that this prompt of `[core@ip-10-0-128-73 ~]` is okay here (and two other instances in the PR)?

Your draft had used `[core@ip-10-0-143-125 ~]`, but that one never shows up in the example member lists used, so I chose one that did.

modules/restore-add-master.adoc
```diff
@@ -0,0 +1,77 @@
...
+
+ETCD_NAME="etcd-member-ip-10-0-147-172.us-east-2.compute.internal"
+ETCD_INITIAL_CLUSTER="etcd-member-ip-10-0-147-172.us-east-2.compute.internal=https://etcd-1.clustername.devcluster.openshift.com:2380,etcd-member-ip-10-0-171-108.us-east-2.compute.internal=https://etcd-2.clustername.devcluster.openshift.com:2380,etcd-member-ip-10-0-128-73.us-east-2.compute.internal=https://etcd-0.clustername.devcluster.openshift.com:2380"
+ETCD_INITIAL_ADVERTISE_PEER_URLS="https://etcd-1.clustername.devcluster.openshift.com:2380"
+ETCD_INITIAL_CLUSTER_STATE="existing"'
+Member  1e42c7070decd39 added to cluster 6e25bab1bb556673
+Starting etcd..
+----
++
+// TODO: Didn't include the pod list verification - okay to skip?
```
bergerhoffer: @retroflexer I didn't include the step to verify the list of pods, because I figured that the next step takes care of the verification. Let me know if you do think that it is a critical to add.

modules/restore-remove-failed-master.adoc
```diff
@@ -0,0 +1,58 @@
...
+$ sudo -E etcd-member-remove.sh etcd-member-ip-10-0-147-172.us-east-2.compute.internal
+Downloading etcdctl binary..
+etcdctl version: 3.3.10
+API version: 3.3
+etcd client certs already backed up and available ./assets/backup/
+Member 23e4736df4451b32 removed from cluster 6e25bab1bb556673
+etcd member etcd-member-ip-10-0-147-172.us-east-2.compute.internal with 23e4736df4451b32 successfully removed..
+----
++
+// TODO: Mention to use 'oc get pods -n openshift-etcd' to get name? Or not necessarily?
```
bergerhoffer: @retroflexer In your draft, you first mentioned to check "oc get pods -n openshift-etcd" to see the pod list. But you said that the etcd member name wasn't necessarily the same as the etcd name, so I didn't include that. Is that okay to leave off then?

backup_and_restore/replacing-failed-master.adoc
```diff
@@ -0,0 +1,33 @@
...
+
+// Removing a failed master host from the cluster
+include::modules/restore-remove-failed-master.adoc[leveloffset=+1]
+
+// Adding a master host back to the cluster
+include::modules/restore-add-master.adoc[leveloffset=+1]
+
+.Next steps
+
+// TODO: Does this need to be done before or after adding the member back to the cluster?
```
bergerhoffer: @retroflexer If the certs aren't valid, does the DR scenario to recover from expired certs need to be done BEFORE or AFTER adding the member back to the etcd cluster?

backup_and_restore/replacing-failed-master.adoc
```diff
@@ -0,0 +1,33 @@
...
+If you have lost the majority of your master hosts, leading to etcd quorum loss, then you must follow the disaster recovery procedure to xref:../backup_and_restore/disaster_recovery/scenario-1-infra-recovery.adoc#dr-infrastructure-recovery[recover from lost master hosts].
+====
+
+To replace a single master host:
+
+* xref:../backup_and_restore/replacing-failed-master.adoc#restore-remove-failed-master_replacing-failed-master-host[Remove the member from the etcd cluster].
+
+* xref:../backup_and_restore/replacing-failed-master.adoc#restore-add-master_replacing-failed-master-host[Add the member back to the etcd cluster].
+
+// TODO: Confirm. But how can you use this to scale up if the procedure assumes that you're adding a member that had previously been removed?
```
bergerhoffer: @retroflexer You mentioned that this procedure was only for adding back members that had been previously removed. Can you give me details on why this is the case? Is it just because it needs to have valid certs (and DNS entries as you mentioned in your comment of "If it is a brand new member, you will not have the requisite certificates and DNS entries etc. So, it perforce has to be a previous member.") If someone had taken care of the DNS entries/certs on a brand new host, could they use this procedure to add it to the etcd cluster?

modules/restore-add-master.adoc
```diff
@@ -0,0 +1,77 @@
+// Module included in the following assemblies:
+//
+// * backup_and_restore/replacing-failed-master.adoc
+
+[id="restore-add-master_{context}"]
+= Adding a master host back to the etcd cluster
+
+Follow these steps to add a master host back to the etcd cluster. This procedure assumes that you previously removed the master host from the cluster and that its TLS certificates are valid.
```
bergerhoffer: @retroflexer - @hexfusion commented in your draft about other etcd dependencies (such as valid DNS entries). 

Should we call out both of these? Or are there others where we should say something more generic about the etcd dependencies and use these two as examples?

bergerhoffer (commented):
@retroflexer I have a few questions for you please, thanks!

FYI @hexfusion

backup_and_restore/replacing-failed-master.adoc
retroflexer: If the certs aren't valid, we recommend them to use recover script.

`recover` script will take care of adding the certs as well as adding the member back to the etcd cluster.

modules/restore-add-master.adoc
retroflexer: No. Actually this is very important. We need to emphasize that the add command should be run on the same master node of the member being added. The manifests and the environment needed for the new member is generated on its master node. 

 So, it should have been:
[core@ip-`10-0-147-172` ~]

backup_and_restore/replacing-failed-master.adoc
retroflexer: Yeah. It requires certificates and DNS entries. If they can take care of certs and DNS, then, we could use the replacing procedure.

retroflexer: But trying to stay out of trouble by explicitly asking them to use it for adding back a previously failed member.

modules/restore-add-master.adoc
retroflexer: Yeah. etcd dependencies and use the two as examples.

modules/restore-add-master.adoc
retroflexer: `member list` is different from ` oc get pods -n openshift-etcd`. We need to verify both.

modules/restore-add-master.adoc
retroflexer: Having the prompt as `[core@ip-10-0-128-73 ~]` is definitely misleading and wrong.

modules/restore-add-master.adoc
hexfusion: To add some context: member list will only show if the member has actually started and joined the cluster in the past. Even if now the member is not running it would still report `started` vs `unstarted`. To verify the health of the members we can do `etcdctl endpoint health --cluster` in which it should return healthy for all 3 members. I think it is reasonable to also verify the pods are running as Suresh pointed out.

hexfusion:
In general, the comments here makes sense to me defer final lgtm to @retroflexer.

retroflexer:
Thanks, Sam (@hexfusion) for your time. 

I will take it from here. Appreciate your time while you should be off taking care of the new born!

modules/restore-add-master.adoc
hexfusion: ^^ This should be a big note. If they run this script on the wrong master they could lose quorum.

bergerhoffer: Okay, I see. So for adding back, they need to be on the host being added, so the prompt here should be `[core@ip-10-0-147-172 ~]`.

But what about for when removing, you don't to be on the host you're removing, do you? That's why I had chosen `[core@ip-10-0-128-73 ~]` there. Let me know about this one.

retroflexer: Yup. remove can be performed from any of the hosts. 

But the BIG NOTE should alert about the `add`. Add must be performed from its own master node, as the files that are generated as part of the procedure are needed to be on the master node for the member to start successfully.

backup_and_restore/replacing-failed-master.adoc
bergerhoffer: @retroflexer Then I think we probably should remove this line about scaling up/down

bergerhoffer:
@retroflexer I've updated based on this feedback. Let me know how it looks now, thanks!

https://bz-1731015--ocpdocs.netlify.com/openshift-enterprise/latest/backup_and_restore/replacing-failed-master.html

retroflexer:
> @retroflexer I've updated based on this feedback. Let me know how it looks now, thanks!
> 
> https://bz-1731015--ocpdocs.netlify.com/openshift-enterprise/latest/backup_and_restore/replacing-failed-master.html

A few minor changes:
1. Can you make the prompt `[core@ip-10-0-147-172 ~]$` in the add section in all commands on host (commands with core@).

2. Remove the backslash (\) in this command: 
```[core@ip-10-0-147-172 ~]$ sudo -E etcd-member-add.sh 10.0.128.73 \ etcd-member-ip-10-0-147-172.us-east-2.compute.internal```

3. What do you think about adding additional command at the end to check the health, as Sam suggested: 
```To verify the health of the members we can do etcdctl endpoint health --cluster```

**Comment by bergerhoffer** [comment]
@bobfuru @lamek For peer review please!

File: modules/dr-recover-lost-control-plane-hosts.adoc [hunk 7]
```diff
@@ -5,27 +5,17 @@
...
-
-Follow these steps to recover from the loss of one or more master hosts.
+Follow these steps to recover from the loss of the majority of master hosts, leading to etcd quorum loss.
 
 .Prerequisites
 
 * Access to the cluster as a user with the `cluster-admin` role.
 * SSH access to a remaining master host.
-* A backed up etcd snapshot, if you are recovering a loss of a majority of masters.
+* A backed up etcd snapshot.
```
bobfuru: s/backed up/backed-up/

modules/restore-add-master.adoc
```diff
@@ -0,0 +1,101 @@
+// Module included in the following assemblies:
+//
+// * backup_and_restore/replacing-failed-master.adoc
+
+[id="restore-add-master_{context}"]
+= Adding a master host back to the etcd cluster
+
+Follow these steps to add a master host back to the etcd cluster. This procedure assumes that you previously removed the master host from the cluster and that its etcd dependencies are valid, such as TLS certificates and DNS.
```
bobfuru: Do you mean:
s/etcd dependencies are valid, such as TLS certificates and DNS/etcd dependencies, such as TLS certificates and DNS, are valid/
?

modules/restore-add-master.adoc
```diff
@@ -0,0 +1,101 @@
...
+* SSH access to the master host to add to the etcd cluster.
+* The IP address of an existing active etcd member.
+
+.Procedure
+
+. Access the master host to add to the etcd cluster.
++
+[IMPORTANT]
+====
+This procedure _must_ be run on the master host being added to the etcd cluster.
```
bobfuru: Because it is already called out in an important note:
s/_must_/must/

Also, "must be run" sounds odd to me. Maybe try "must run", "must be executed", or, "You must run this procedure on the master host that is being added to the etcd cluster."

modules/restore-add-master.adoc
```diff
@@ -0,0 +1,101 @@
...
+====
+
+. Run the `etcd-member-add.sh` script:
++
+Pass in two parameters to the `etcd-member-add.sh` script: IP address of an existing etcd member and the name of the etcd member to add.
```
bobfuru: /: IP address/: the IP address/

Also, there are almost two steps here. Might read better if you split the second sentence into two bullets. Something like:

2. Run the `etcd-member-add.sh` script, where you will pass in two parameters:

- the IP address of an existing etcd member
- the name of the etcd member to add

modules/restore-add-master.adoc
```diff
@@ -0,0 +1,101 @@
...
++
+----
+[core@ip-10-0-147-172 ~]$ sudo -E etcd-member-add.sh 10.0.128.73 etcd-member-ip-10-0-147-172.us-east-2.compute.internal
```
bobfuru: Do you need the `[core@ip-10-0-147-172 ~]` prompt here (as well as steps 3-4)?

modules/restore-add-master.adoc
```diff
@@ -0,0 +1,101 @@
...
+
+ETCD_NAME="etcd-member-ip-10-0-147-172.us-east-2.compute.internal"
+ETCD_INITIAL_CLUSTER="etcd-member-ip-10-0-147-172.us-east-2.compute.internal=https://etcd-1.clustername.devcluster.openshift.com:2380,etcd-member-ip-10-0-171-108.us-east-2.compute.internal=https://etcd-2.clustername.devcluster.openshift.com:2380,etcd-member-ip-10-0-128-73.us-east-2.compute.internal=https://etcd-0.clustername.devcluster.openshift.com:2380"
+ETCD_INITIAL_ADVERTISE_PEER_URLS="https://etcd-1.clustername.devcluster.openshift.com:2380"
+ETCD_INITIAL_CLUSTER_STATE="existing"'
+Member  1e42c7070decd39 added to cluster 6e25bab1bb556673
+Starting etcd..
+----
+
+. Verify that the new member is in the list of Pods associated with etcd and is actively running:
```
bobfuru: "and is actively running" 
It's not clear to me what the subject is here ("and _what_ is actively running"), the new member?

modules/restore-add-master.adoc
```diff
@@ -0,0 +1,101 @@
...
+
+. Verify that the etcd member has been successfully added to the etcd cluster:
+
+.. Connect to the running etcd container:
++
+----
+[core@ip-10-0-147-172 ~] id=$(sudo crictl ps --name etcd-member | awk 'FNR==2{ print $1}') && sudo crictl exec -it $id /bin/sh
+----
+
+.. In the etcd container, export variables needed for connecting to etcd:
```
bobfuru: s/variables/the variables/

modules/restore-add-master.adoc
```diff
@@ -0,0 +1,101 @@
...
+|        ID        | STATUS  |                   NAME                   |                            PEER ADDRS                            |       CLIENT ADDRS        |
++------------------+---------+------------------------------------------+------------------------------------------------------------------+---------------------------+
+| 29e461db6be4eaaa | started | etcd-member-ip-10-0-128-73.us-east-2.compute.internal | https://etcd-2.clustername.devcluster.openshift.com:2380 | https://10.0.128.73:2379 |
+|  cbe982c74cbb42f | started | etcd-member-ip-10-0-147-172.us-east-2.compute.internal | https://etcd-0.clustername.devcluster.openshift.com:2380 | https://10.0.147.172:2379 |
+| a752f80bcb0da3e8 | started |   etcd-member-ip-10-0-171-108.us-east-2.compute.internal | https://etcd-1.clustername.devcluster.openshift.com:2380 |   https://10.0.171.108:2379 |
++------------------+---------+------------------------------------------+------------------------------------------------------------------+---------------------------+
+----
++
+Note that it may take up to 10 minutes for the new member to start.
```
bobfuru: Should this be in a note block? I can only find one other instance in 4.x docs where we use "note that" in paragraph.

modules/restore-add-master.adoc
```diff
@@ -0,0 +1,101 @@
...
+
+.. In the etcd container, execute `etcdctl endpoint status` and verify that the new member is listed and healthy:
```
bobfuru: Will users understand what constitutes "healthy" without additional explanation?

modules/restore-remove-failed-master.adoc
```diff
@@ -0,0 +1,66 @@
...
+* SSH access to an active master host.
+
+.Procedure
+
+. Access an active master host.
+
+. View the list of Pods associated with etcd:
++
+----
+[core@ip-10-0-128-73 ~]$ oc get pods -n openshift-etcd
```
bobfuru: Same question as above, do we need the command prompt before `$` in these steps?

modules/restore-remove-failed-master.adoc
```diff
@@ -0,0 +1,66 @@
...
+
+. Verify that the etcd member has been successfully removed from the cluster:
+
+.. Connect to the running etcd container:
++
+----
+[core@ip-10-0-128-73 ~] id=$(sudo crictl ps --name etcd-member | awk 'FNR==2{ print $1}') && sudo crictl exec -it $id /bin/sh
+----
+
+.. In the etcd container, export variables needed for connecting to etcd:
```
bobfuru: s/export variables/export the variables/

bobfuru (commented):
Minor edits and a few questions but otherwise LGTM!

modules/restore-add-master.adoc
bergerhoffer: Yeah actually we do, to be clear about which host you are on

modules/restore-add-master.adoc
bergerhoffer: I actually do think we use this a decent bit, but I think in this case I can remove the "Note that" and just start with "It".

backup_and_restore/replacing-failed-master.adoc
```diff
@@ -0,0 +1,27 @@
...
+include::modules/common-attributes.adoc[]
+:context: replacing-failed-master-host
+
+toc::[]
+
+This document describes the process to replace a single etcd member. This procedure assumes that there is still an etcd quorum in the cluster.
+
+[NOTE]
+====
+If you have lost the majority of your master hosts, leading to etcd quorum loss, then you must instead follow the disaster recovery procedure to xref:../backup_and_restore/disaster_recovery/scenario-1-infra-recovery.adoc#dr-infrastructure-recovery[recover from lost master hosts].
```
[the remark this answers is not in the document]
bergerhoffer: I'll reword and see if it's clearer.

modules/restore-add-master.adoc
bergerhoffer: @retroflexer Can you confirm if there's anything in this output that would indicate to a user that the member is not healthy?

retroflexer: You could add the following command to check the health additionally:
```
sh-4.2# etcdctl endpoint health --cluster
https://10.0.128.73:2379 is healthy: successfully committed proposal: took = 4.5576ms
https://10.0.147.172:2379 is healthy: successfully committed proposal: took = 5.1521ms
https://10.0.171.108:2379 is healthy: successfully committed proposal: took = 4.2631ms
```

PR #17554, opened 2019-10-23
CNV-1204 edit vm templates

aburdenthehand:
It's possible to edit virtual machine template details in the web console. This process is the same as editing VMs in the web console. 

cherrypick to enterprise-4.3 only


aburdenthehand:
The procedure and information for editing VMs and VM templates in the UI is expected to remain identical (and has changed for CNV 2.2), so I combined the content into a single module, separating it with conditionals.

ahardin-rh:
Looks good!

PR #17566, opened 2019-10-23
adding GCP restricted networks

kalexand-rh:
@gpei, will you PTAL?

### DIFF
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -137,9 +137,11 @@ Topics:
   - Name: Installing a cluster on GCP into an existing VPC
     File: installing-gcp-vpc
   - Name: Installing a private cluster on GCP
-    File: installing-gcp-private    
+    File: installing-gcp-private
   - Name: Installing a cluster on GCP using Deployment Manager templates
     File: installing-gcp-user-infra
+  - Name: Restricted network GCP installation
+    File: installing-restricted-networks-gcp
   - Name: Uninstalling a cluster on GCP
     File: uninstalling-cluster-gcp
 - Name: Installing on bare metal
@@ -767,7 +769,7 @@ Topics:
 - Name: Modifying a MachineSet
   File: modifying-machineset
 - Name: Deleting a Machine
-  File: deleting-machine  
+  File: deleting-machine
 - Name: Applying autoscaling to a cluster
   File: applying-autoscaling
 - Name: Creating infrastructure MachineSets
--- a/installing/installing_gcp/installing-restricted-networks-gcp.adoc
+++ b/installing/installing_gcp/installing-restricted-networks-gcp.adoc
@@ -1,12 +1,18 @@
-[id="installing-gcp-user-infra"]
-= Installing a cluster on GCP using Deployment Manager templates
+[id="installing-restricted-networks-gcp"]
+= Installing a cluster on GCP in a restricted network
 include::modules/common-attributes.adoc[]
-:context: installing-gcp-user-infra
+:context: installing-restricted-networks-gcp
 
 toc::[]
 
 In {product-title} version {product-version}, you can install a cluster on
-Google Cloud Platform (GCP) that uses infrastructure that you provide.
+Google Cloud Platform (GCP) that uses infrastructure that you provide and an internal mirror of the installation release content.
+
+[IMPORTANT]
+====
+While you can install an {product-title} cluster by using mirrored installation
+release content, your cluster still requires internet access to use the GCP APIs.
+====
 
 The steps for performing a user-provided infrastructure install are outlined here. Several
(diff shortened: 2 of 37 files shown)


_topic_map.yml
```diff
@@ -110,6 +110,8 @@ Topics:
     File: installing-gcp-default
   - Name: Installing a cluster on GCP with customizations
     File: installing-gcp-customizations
+  - Name: Restricted network GCP installation
+    File: installing-restricted-networks-gcp
```
gpei: So why this section was moved to IPI-on-GCP, from my understanding, it should be placed where it was - `Installing in restricted networks `

installing/installing_gcp/installing-restricted-networks-gcp.adoc
```diff
@@ -0,0 +1,113 @@
...
+include::modules/installation-gcp-service-account.adoc[leveloffset=+2]
+include::modules/installation-gcp-permissions.adoc[leveloffset=+3]
+include::modules/installation-gcp-regions.adoc[leveloffset=+2]
+include::modules/installation-gcp-install-cli.adoc[leveloffset=+2]
+
+include::modules/installation-user-infra-generate.adoc[leveloffset=+1]
+
+include::modules/installation-initializing.adoc[leveloffset=+2]
+
+include::modules/installation-configure-proxy.adoc[leveloffset=+2]
```
gpei: Do we really need `proxy` added in this doc? Once proxy enabled, that means the whole cluster would have the access to Internet, then we don't need mirror registry any more. 
Unless customer have to use a mirror registry in their deployment and also need proxy to access some websites.

And for a cluster have no access to the Internet on GCP, I remember we need to ensure several addresses(to access some google services) must be accessible for the cluster when without proxy, I could sort them out later if we decide not to involve proxy.
@jianlinliu wdyt?

_topic_map.yml
kalexand-rh: I'm working on a change to reorder the installation table of contents based on cloud provider. I that change to merge before this one does.

kalexand-rh:
@gpei, @jianlinliu, my apologies for the delay. I've removed the section about configuring a proxy. What other changes do I need to make?

gpei:
@kalexand-rh we may also need to document several URL should be available for the cluster in restricted network, but I see we already have https://docs.openshift.com/container-platform/4.2/installing/install_config/configuring-firewall.html , this Whitelist seems to be enough.

kalexand-rh:
@gpei, should everyone using a restricted cluster always whitelist the sites in that topic?

gpei:
@kalexand-rh For restricted cluster on GCP, `*.googleapis.com` and `accounts.google.com` should be in the whitelist.

kalexand-rh:
@gpei, I've updated the prereq to cover the requirement. Thank you!

kalexand-rh:
@gpei, will you confirm that this is ready to merge?

gpei:
@kalexand-rh lgtm now

kalexand-rh:

installing/installing_gcp/installing-gcp-user-infra.adoc
```diff
@@ -14,7 +14,20 @@ completing these steps or to help model your own. You are also free to create
...
 example.
 
-[id="csr-management-aws_{context}"]
+.Prerequisites
+
+* Review details about the
+xref:../../architecture/architecture-installation.adoc#architecture-installation[{product-title} installation and update]
+processes.
+* If you use a firewall and plan to use telemetry, you must
+xref:../../installing/install_config/configuring-firewall.adoc#configuring-firewall[configure it to allow the sites] that your cluster requires access to.
```
mburke5678: s/it/the firewall

installing/installing_gcp/installing-restricted-networks-gcp.adoc
```diff
@@ -0,0 +1,106 @@
+[id="installing-restricted-networks-gcp"]
+= Installing a cluster on GCP in a restricted network
+include::modules/common-attributes.adoc[]
+:context: installing-restricted-networks-gcp
+
+toc::[]
+
+In {product-title} version {product-version}, you can install a cluster on
+Google Cloud Platform (GCP) by using infrastructure that you provide and an internal mirror of the installation release content.
```
mburke5678: s/by using/using ?

kalexand-rh: > ! using n
To avoid ambiguity, replace this gerund with either “by using” or “that uses.”

mburke5678: In that case, would `that uses` be better?

mburke5678:
A couple of nits. Otherwise LGTM

PR #17569, opened 2019-10-23
extract the release image on a machine with internet

kalexand-rh:
Per @umohnani8 in Slack.

(@wking, FYI)

@jianlinliu, will you PTAL?

### DIFF
--- a/modules/installation-mirror-repository.adoc
+++ b/modules/installation-mirror-repository.adoc
@@ -74,4 +74,6 @@ $ oc adm release extract --command=openshift-install "${LOCAL_REGISTRY}/${LOCAL_
 To ensure that you use the correct images for the version of {product-title}
 that you selected, you must extract the installation program from the mirrored
 content.
+
+You must perform this step on a machine with an active internet connection.
 ====
--- a/installing/installing_restricted_networks/installing-restricted-networks-preparations.adoc
+++ b/installing/installing_restricted_networks/installing-restricted-networks-preparations.adoc
@@ -27,9 +27,9 @@ include::modules/cli-installing-cli.adoc[leveloffset=+2]
 
 include::modules/installation-creating-mirror-registry.adoc[leveloffset=+1]
 
-include::modules/installation-local-registry-pull-secret.adoc[leveloffset=+1]
+//include::modules/installation-local-registry-pull-secret.adoc[leveloffset=+1]
 
-//include::modules/installation-adding-registry-pull-secret.adoc[leveloffset=+1]
+include::modules/installation-adding-registry-pull-secret.adoc[leveloffset=+1]
 
 include::modules/installation-mirror-repository.adoc[leveloffset=+1]
 
--- a/modules/installation-obtaining-installer.adoc
+++ b/modules/installation-obtaining-installer.adoc
@@ -17,8 +17,13 @@
 [id="installation-obtaining-installer_{context}"]
 = Obtaining the installation program
 
-Before you install {product-title}, download the installation file on a local
-computer.
+Before you install {product-title}, download the installation file on
+ifdef::restricted[]
+the bastion host.
+endif::restricted[]
+ifndef::restricted[]
+a local computer.
+endif::restricted[]
 
 .Prerequisites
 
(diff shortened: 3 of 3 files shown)


jianlinliu:
Refer to https://jira.coreos.com/browse/CORS-1107?focusedCommentId=114277&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-114277, 

> all clusters that are created by using `OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE` are unsupported.

kalexand-rh:
Thank you for reminding me @jianlinliu, @umohnani8, do you have another suggestion for getting your customer back on track?

umohnani8:
@kalexand-rh yeah we are just working on confirming the detail about the ENV variable. I will comment back here in a bit. But the `oc adm release extract` command is definitely wrong and not needed in the disconnected case.

wking:
Using the official installer without alteration is what we want to have work.  Testing in openshift/release#5567.  But until we get that green, extracting from the mirrored release like the master docs have is certainly more official than using the environment variable override.

kalexand-rh:
@wking, let me know when that merges so that we can account for it in doc. @umohnani8, I'm not going to merge this change as-is, but if there's additional information to present instead, please let me know so I can amend this PR.

kalexand-rh:
I'm going to close this PR. @umohnani8, please let me know if you have more information that you'd like me to incorporate.

umohnani8:
@kalexand-rh can we reopen this. We confirmed that this works as expected. We need to remove the `oc adm release extract` command from the steps.

modules/installation-mirror-repository.adoc
```diff
@@ -33,6 +33,7 @@ $ export LOCAL_REPOSITORY='<repository_name>' <3>
 $ export PRODUCT_REPO='openshift-release-dev' <4>
 $ export LOCAL_SECRET_JSON='<path_to_pull_secret>' <5>
 $ export RELEASE_NAME="ocp-release" <6>
+$ export OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE=${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}:${OCP_RELEASE}`
```
umohnani8: please remove this line

kalexand-rh:
@umohnani8, is it supported now? I hesitate to bring a change in that both Abhinav and Jianlin have declined.

umohnani8:
Yeah, we don't want to add the `OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE` into the docs, so that should be removed.
But the `oc adm release extract` command should either be removed or moved up so that users know they have to run that on a machine that has access to the internet and then they can copy the artifacts over to the machine that is disconnected.
@wking @abhinavdahiya @christianh814 please correct me if I am mistaken.

umohnani8:
LGTM
@wking PTAL

kalexand-rh:
Trevor says that this looks ok to him. @jianlinliu, are you ok with this change?

jianlinliu:
> But the oc adm release extract command should either be removed or moved up so that users know they have to run that on a machine that has access to the internet and then they can copy the artifacts over to the machine that is disconnected.

Actually I still can not figure out why user have to extract installer on a machine that has access to the internet. I agree `Mirror the repository` should be run on a machine that has access to the internet  (In original doc, `Complete the following steps on the bastion host` already indicates that machine has access to the internet), but I can not agree extract installer have to be run on a machine that has access to the internet. Once release repository is mirrored onto mirror registry, any other disconnected machine can extract installer from the mirrored release repository, the connectivity between mirror registry and disconnected machine should be available, right? That is why we call the mirror registry machine as a bastion host. Do I miss something?  Based on the PR is harmless for original process, I do not object to this PR too much.

kalexand-rh:
@jianlinliu, that's about where I am with this PR - I think that the existing direction to perform the steps on the bastion host was enough, but adding a reminder in a different location is harmless.

modules/installation-mirror-repository.adoc
```diff
@@ -74,4 +74,6 @@ $ oc adm release extract --command=openshift-install "${LOCAL_REGISTRY}/${LOCAL_
 To ensure that you use the correct images for the version of {product-title}
 that you selected, you must extract the installation program from the mirrored
 content.
+
+You must perform this step on a machine with an active internet connection.
```
bobfuru: Consider combining this with the previous sentence into a single admonition: "...you must extract the installation program from the mirrored content using a machine with an active internet connection."

bobfuru:
One small comment; otherwise, LGTM!

modules/installation-mirror-repository.adoc
kalexand-rh: If the major point of the PR wasn't "people are missing that you need to do this on a machine with internet," I'd take that improvement.

jianlinliu:
Today someone reported similar issue - https://bugzilla.redhat.com/show_bug.cgi?id=1777890. I reproduced it, refer to that bug for more details. So now this PR sound reasonable for me now.

PR #17580, opened 2019-10-24
bug 176848 reference supported regions in the Azure install topics

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1764848


kalexand-rh:

maxwelldb:
I could see nixing "supported" in the text that follows the links, but looks copacetic to me. :+1:

PR #17320, opened 2019-10-15
masterClientConnectionOverrides set in node-config.yaml is ignored

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1691046

PR #17585, opened 2019-10-24
Add qps burst

mburke5678:
@gpei asked for a change in https://github.com/openshift/openshift-docs/pull/17320

_I noticed we have a specialized section https://docs.openshift.com/container-platform/3.11/admin_guide/manage_nodes.html#configuring-node-resources about how to add parameter to "kubeletArguments", can we add the expected kube-api-qps and kube-api-burst by following the above doc?_


mburke5678:
@gpei PTAL

gpei:
LGTM

PR #17587, opened 2019-10-24
Fix ingress controller sharding example when using namespace labels

sushilsuresh:
The ingress controller sharding example is using the wrong selector.
Instead of using namespaceSelector for selecting based on namespace labels, the example is using routeSelector.


kalexand-rh:
@bmcelvee, who should provide QE for this suggested change?

Miciah:
The example was fixed in #16648.

bmcelvee:
Thanks, @Miciah and @sushilsuresh! Since this was already fixed, I'll close this PR.

PR #17603, opened 2019-10-24
BZ1747486 - adding hostPath volume doc

bobfuru:
[BZ 1747486](https://bugzilla.redhat.com/show_bug.cgi?id=1747486)
Added hostPath config example, based on upstream docs.
@sferich888 and @liangxia PTAL

### DIFF
--- a/modules/persistent-storage-hostpath-pod.adoc
+++ b/modules/persistent-storage-hostpath-pod.adoc
@@ -1,0 +1,41 @@
+// Module included in the following assemblies:
+//
+// * storage/persistent-storage/persistent-storage-hostpath.adoc
+
+[id="persistent-storage-hostpath-pod_{context}"]
+= Mounting the hostPath share in a privileged Pod
+
+After the PersistentVolumeClaim has been created, it can be used inside by an application. The following example demonstrates mounting this share inside of a Pod.
+
+.Prerequisites
+* A PersistentVolumeClaim exists that is mapped to the underlying hostPath share.
+
+.Procedure
+
+* Create a privileged Pod that mounts the existing PersistentVolumeClaim:
++
+[source,yaml]
+----
+apiVersion: v1
+kind: Pod
+metadata:
+  name: pod-name <1>
+spec:
+  containers:
+    ...
+    securityContext:
+      privileged: true <2>
+    volumeMounts:
+    - mountPath: /data <3>
+      name: hostpath-privileged
+  ...
+  securityContext: {}
+  volumes:
+    - name: hostpath-privileged
+      persistentVolumeClaim:
+        claimName: task-pvc-volume <4>
+----
+<1> The name of the Pod.
+<2> The Pod must run as privileged to access the node's storage.
+<3> The path to mount the hostPath share inside the privileged Pod.
+<4> The name of the PersistentVolumeClaim that has been previously created.
--- a/modules/persistent-storage-hostpath-static-provisioning.adoc
+++ b/modules/persistent-storage-hostpath-static-provisioning.adoc
@@ -1,0 +1,64 @@
+// Module included in the following assemblies:
+//
+// * storage/persistent-storage/persistent-storage-vsphere.adoc
+
+[id="vsphere-static-provisioning_{context}"]
+= Statically provisioning hostPath volumes
+
+A Pod that uses a hostPath volume must be referenced by manual (static) provisioning.
+
+.Procedure
+
+. Define the persistent volume (PV). Create a file, `pv.yaml`, with the PersistentVolume object definition:
++
+[source,yaml]
+----
+  apiVersion: v1
+  kind: PersistentVolume
+  metadata:
+    name: task-pv-volume <1>
+    labels:
+      type: local
+  spec:
+    storageClassName: manual <2>
+    capacity:
+      storage: 5Gi
+    accessModes:
+      - ReadWriteOnce <3>
+    persistentVolumeReclaimPolicy: Retain
+    hostPath:
+      path: "/mnt/data" <4>
+----
+<1> The name of the volume. This name is how it is identified by PersistentVolumeClaims or Pods.
+<2> Used to bind PersistentVolumeClaim requests to this PersistentVolume.
+<3> The volume can be mounted as `read-write` by a single node.
+<4> The configuration file specifies that the volume is at `/mnt/data` on the cluster’s node.
+
+. Create the PV from the file:
++
+----
+$ oc create -f pv.yaml
+----
+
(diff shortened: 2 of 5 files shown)


liangxia:
LGTM.
Need to CP to 4.1, too.

bobfuru:
@sferich888 Bump, PTAL.

modules/persistent-storage-hostpath-mounting.adoc
```diff
@@ -0,0 +1,35 @@
...
+// * storage/persistent-storage/persistent-storage-hostpath.adoc
+
+[id="persistent-storage-hostpath-mounting_{context}"]
+= Mounting with hostPath
+
+{product-title} supports hostPath mounting for development and testing on a single-node cluster.
+
+In a production cluster, you would not use hostPath. Instead, a cluster administrator would provision a network resource, such as a GCE Persistent Disk volume, an NFS share, or an Amazon EBS volume.
+
+Cluster administrators can also use StorageClasses to set up dynamic provisioning.
```
sferich888: Do we need to show how to set up the storage class for this?

bobfuru: @sferich888 I'm investigating. hostPath dynamic provisioning is not supported, so my understanding is that storage class may not be needed at all.

bobfuru:
@liangxia 
@huffmanca 
PTAL - I have refactored this assembly to give the user instructions for configuring persistent storage using hostPath. Confirmed with Dev that dynamic provisioning is not supported. Updates to this PR are intended to give the user instructions for defining the PV and PVC, as well as calling out that Pods must be run as privileged. Also following up with PM about support level.

modules/persistent-storage-hostpath-static-provisioning.adoc
```diff
@@ -0,0 +1,64 @@
...
+kind: PersistentVolumeClaim
+metadata:
+  name: task-pvc-volume
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
+  storageClassName: slow
```
liangxia: The PVC's storageClassName mismatch with PV's storageClassName.

liangxia:
Other than the storageClassName mismatch, LGTM.

chao007:
Can we add a pod file here?
For hostpath volume, we need to use privileged pod.

modules/persistent-storage-hostpath-static-provisioning.adoc
bobfuru: Updated so they are both using `manual` now.

bobfuru:
> Can we add a pod file here?
> For hostpath volume, we need to use privileged pod.

Yes, good suggestion. I have added `modules/persistent-storage-hostpath-pod.adoc` to this commit.

bobfuru:
@liangxia or @chao007 
PTAL at revised version with added Pod file

@sferich888 
Can I also get an ack that this looks good to you?

modules/persistent-storage-hostpath-pod.adoc
```diff
@@ -0,0 +1,41 @@
...
+//
+// * storage/persistent-storage/persistent-storage-hostpath.adoc
+
+[id="persistent-storage-hostpath-pod_{context}"]
+= Mounting the hostPath share in a privileged Pod
+
+After the PersistentVolumeClaim has been created, it can be used inside by an application. The following example demonstrates mounting this share inside of a Pod.
+
+.Prerequisites
+* A PersistentVolumeClaim exists that is mapped to the underlying hostPath share.
```
sferich888: Should we link to docs on creating this pvc?

modules/persistent-storage-hostpath-static-provisioning.adoc
```diff
@@ -0,0 +1,64 @@
+// Module included in the following assemblies:
+//
+// * storage/persistent-storage/persistent-storage-vsphere.adoc
+
+[id="vsphere-static-provisioning_{context}"]
+= Statically provisioning hostPath volumes
```
sferich888: Don't we list this in https://github.com/openshift/openshift-docs/pull/17603/files#diff-1c65246cb57e25d53ab01b23a5d211de

modules/persistent-storage-hostpath-pod.adoc
sferich888: /lgtm after we consider this

chao007:
> > Can we add a pod file here?
> > For hostpath volume, we need to use privileged pod.
> 
> Yes, good suggestion. I have added `modules/persistent-storage-hostpath-pod.adoc` to this commit.

Thanks.

modules/persistent-storage-hostpath-pod.adoc
bobfuru: The steps for creating this PVC are actually described in the previous module (modules/persistent-storage-hostpath-static-provisioning.adoc). But unfortunately, our toolchain prevents the use of module-to-module xrefs. And because each 4.x module is written to serve as standalone content so that it may be reused anywhere else in our docs, we also can't refer to "the previous step." So, I am leaving it as is.

bobfuru:
@huffmanca PTAL - acks received from QE and Support. Thanks!

modules/persistent-storage-hostpath-pod.adoc
```diff
@@ -0,0 +1,41 @@
...
+      name: hostpath-privileged
+  ...
+  securityContext: {}
+  volumes:
+    - name: hostpath-privileged
+      persistentVolumeClaim:
+        claimName: task-pvc-volume <4>
+----
+<1> The name of the Pod.
+<2> The Pod must run as privileged to grant access from mounted storage to data across Pods in the same node.
```
huffmanca: This sentence is a bit wordy. I think we can just say `The Pod must run as privileged to access the node's storage.`

huffmanca (commented):
Only a couple of minor things. The rest looks good to me!

storage/persistent-storage/persistent-storage-hostpath.adoc
```diff
@@ -0,0 +1,17 @@
+[id="persistent-storage-using-hostpath"]
+= Persistent storage using hostPath
+include::modules/common-attributes.adoc[]
+:context: persistent-storage-hostpath
+
+toc::[]
+
+A _hostPath_ volume in an {product-title} cluster mounts a file or directory from the host node’s filesystem into your Pod. Most Pods will not need a hostPath volume, but it does offer powerful protection should an application require it.
```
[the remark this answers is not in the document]
bobfuru: I see what you mean. Original wording was "powerful escape hatch" so was looking for something more translatable but "protection" is not the right choice. Changing it to say that "it does offer a quick option for testing should an application require it."

PR #17608, opened 2019-10-24
BZ 1762140 - added cinder RWO

bobfuru:
Added Cinder to table (with RWO)
[BZ 1762140](https://bugzilla.redhat.com/show_bug.cgi?id=1762140)
@liangxia PTAL


liangxia:
The change looks good to me.

BUT we should apply it to 4.2 and 4.3 (4.1 does not support OpenStack)

PR #17641, opened 2019-10-25
BZ1622018 - add StorageClass default note master-3

bobfuru:
[BZ 1622018](https://bugzilla.redhat.com/show_bug.cgi?id=1622018)
Adds note about StorageClass default to 3.10, 3.11. 
@liangxia PTAL

### DIFF
--- a/install_config/persistent_storage/dynamically_provisioning_pvs.adoc
+++ b/install_config/persistent_storage/dynamically_provisioning_pvs.adoc
@@ -525,6 +525,11 @@ $ oc patch storageclass standard -p '{"metadata": {"annotations": \
     {"storageclass.kubernetes.io/is-default-class": "true"}}}'
 ----
 
+[NOTE]
+====
+If more than one StorageClass is marked as default, a PVC can only be created if the `storageClassName` is explicitly specified. Therefore, only one StorageClass should be set as the default.
+====
+
 . Verify the changes:
 +
 [source,bash]


install_config/persistent_storage/dynamically_provisioning_pvs.adoc
```diff
@@ -525,6 +525,11 @@ $ oc patch storageclass standard -p '{"metadata": {"annotations": \
     {"storageclass.kubernetes.io/is-default-class": "true"}}}'
 ----
 
+[NOTE]
+====
+Only one StorageClass can be set as default. If more than one default is marked, a PVC cannot be created without explicitly specifying the `storageClassName`.
```
liangxia: 1. s/Only one StorageClass can be set as default/Only one StorageClass should be set as default/
2. A PVC still can be created when there are more than on default storage class. It is that no PV and backend volume dynamic provisioned to bind with it.
So we need to re-organized the sentences.

bobfuru: Trying to capture comment and K8s docs to make this more clear. Not sure if I understood this correctly:  "...no PV and backend volume dynamic provisioned to bind with it."

Revised: "If more than one StorageClass is marked as default, a PV cannot be dynamically provisioned. With more than one default, a PVC can only be created if the `storageClassName` is explicitly specified. Therefore, only one StorageClass should be set as the default."

Another option is to shorten to: "If more than one StorageClass is marked as default, a PVC can only be created if the `storageClassName` is explicitly specified. Therefore, only one StorageClass should be set as the default."

liangxia: @bobfuru The shorten one looks great.

bobfuru:

lamek (commented):
LGTM. One suggestion: you can get rid of the word "therefore" if you swap the order of the sentences.

> Only one StorageClass should be set as the default. If more than one StorageClass is marked as default, a PVC can only be created if the `storageClassName` is explicitly specified.

bobfuru:
> LGTM. One suggestion: you can get rid of the word "therefore" if you swap the order of the sentences.
> 
> > Only one StorageClass should be set as the default. If more than one StorageClass is marked as default, a PVC can only be created if the `storageClassName` is explicitly specified.

Thanks. I went back and forth on this, but it's important in this case that the best practice be named first. I'm going to leave it in this case.

PR #17671, opened 2019-10-28
BZ1761703 | Adding new storage - Cinder assembly

bobfuru:
[BZ 1761703](https://bugzilla.redhat.com/show_bug.cgi?id=1761703) - Adding assembly for persistent storage using OpenStack Cinder. This is for 4.2+.
@liangxia PTAL

### DIFF
--- a/modules/persistent-storage-cinder-creating-pv.adoc
+++ b/modules/persistent-storage-cinder-creating-pv.adoc
@@ -1,0 +1,47 @@
+// Module included in the following assemblies:
+//
+// * storage/persistent_storage/persistent-storage-cinder.adoc
+
+[id="persistent-storage-cinder-creating-pv_{context}"]
+= Creating the persistent volume
+
+You must define your persistent volume (PV) in an object definition before creating
+it in {product-title}:
+
+.Procedure
+
+. Save your object definition to a file.
++
+.cinder-persistentvolume.yaml
+[source,yaml]
+----
+apiVersion: "v1"
+kind: "PersistentVolume"
+metadata:
+  name: "pv0001" <1>
+spec:
+  capacity:
+    storage: "5Gi" <2>
+  accessModes:
+    - "ReadWriteOnce"
+  cinder: <3>
+    fsType: "ext3" <4>
+    volumeID: "f37a03aa-6212-4c62-a805-9ce139fab180" <5>
+----
+<1> The name of the volume that is used by persistent volume claims or pods.
+<2> The amount of storage allocated to this volume.
+<3> Indicates `cinder` for OpenStack Cinder volumes.
+<4> The file system that is created when the volume is mounted for the first time.
+<5> The Cinder volume to use.
++
+[IMPORTANT]
+====
+Do not change the `*fstype*` parameter value after the volume is formatted and
+provisioned. Changing this value can result in data loss and Pod failure.
+====
+
+. Create the object definition file you saved in the previous step.
++
+----
+$ oc create -f cinder-persistentvolume.yaml
+----
--- a/modules/persistent-storage-cinder-provisioning.adoc
+++ b/modules/persistent-storage-cinder-provisioning.adoc
@@ -1,0 +1,13 @@
+// Module included in the following assemblies:
+//
+// * storage/persistent_storage/persistent-storage-cinder.adoc
+
+[id="persistent-storage-cinder-provisioning_{context}"]
+= Manual provisioning with Cinder
+
+Storage must exist in the underlying infrastructure before it can be mounted as a volume in {product-title}.
+
+.Prerequisites
+
+* {product-title} configured for OpenStack
+* Cinder volume ID
--- a/modules/persistent-storage-cinder-volume-security.adoc
+++ b/modules/persistent-storage-cinder-volume-security.adoc
@@ -1,0 +1,58 @@
+// Module included in the following assemblies:
+//
+// * storage/persistent_storage/persistent-storage-cinder.adoc
+
+[id="persistent-storage-cinder-volume-security_{context}"]
+= Cinder volume security
+
+If you use Cinder PVs in your application, configure security for their
+deployment configurations.
+
+.Prerequisite
+- An SCC must be created that uses the appropriate `*fsGroup*` strategy.
+
+.Procedure
+
+. Create a service account and add it to the SCC:
++
+----
+$ oc create serviceaccount <service_account>
+$ oc adm policy add-scc-to-user <new_scc> -z <service_account> -n <project>
+----
+
+. In your application's deployment configuration, provide the service account
+name and `securityContext`:
++
+[source,yaml]
+----
+apiVersion: v1
+kind: ReplicationController
+metadata:
+  name: frontend-1
+spec:
+  replicas: 1  <1>
+  selector:    <2>
+    name: frontend
+  template:    <3>
+    metadata:
+      labels:  <4>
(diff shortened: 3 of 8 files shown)


liangxia:
@chao007 from QE will review.

modules/persistent-storage-cinder-creating-pv.adoc
```diff
@@ -0,0 +1,41 @@
...
+= Creating the Persistent Volume
+
+You must define your persistent volume (PV) in an object definition before creating
+it in {product-title}:
+
+.Procedure
+
+. Save your object definition to a file.
++
+.azure-advanced-disk-storageclass.yaml
```
chao007: File name is started with "azure"

bobfuru: @chao007 Can I get an ack please that this PR looks good from QE?

chao007: Yes. Thanks for update

storage/persistent-storage/persistent-storage-cinder.adoc
```diff
@@ -0,0 +1,22 @@
+[id="persistent-storage-cinder"]
+= Persistent storage using Cinder
+include::modules/common-attributes.adoc[]
+:context: persistent-storage-cinder
+
+toc::[]
+
+<product-title} supports OpenStack Cinder. You can provision your {product-title} cluster with persistent storage using link:https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/8/html-single/architecture_guide/index#comp-cinder[OpenStack Cinder]. Some familiarity with Kubernetes and OpenStack is assumed.
```
huffmanca: **<product-title}** should be **{product-title}**

huffmanca: I'd also move the link into an **Additional references** section to be consistent with the other assemblies.

modules/persistent-storage-cinder-about.adoc
```diff
@@ -0,0 +1,11 @@
+// Module included in the following assemblies:
+//
+// * storage/persistent_storage/persistent-storage-cinder.adoc
+
+[id="persistent-storage-cinder-about_{context}"]
+= Overview
```
huffmanca: Is this module used? I don't see it included anywhere.

modules/persistent-storage-cinder-creating-pv.adoc
```diff
@@ -0,0 +1,41 @@
...
+
+[id="persistent-storage-cinder-creating-pv_{context}"]
+= Creating the Persistent Volume
+
+You must define your persistent volume (PV) in an object definition before creating
+it in {product-title}:
+
+.Procedure
+
+. Save your object definition to a file.
```
huffmanca: This is currently a procedure with only one step. I'd recommend either adding a step afterward that uses `oc create -f <file>` to create the saved file, or changing this to an unordered list.

modules/persistent-storage-cinder-creating-pv.adoc
```diff
@@ -0,0 +1,41 @@
...
+    storage: "5Gi" <2>
+  accessModes:
+    - "ReadWriteOnce"
+  cinder: <3>
+    fsType: "ext3" <4>
+    volumeID: "f37a03aa-6212-4c62-a805-9ce139fab180" <5>
+----
+<1> The name of the volume that is used by persistent volume claims or pods.
+<2> The amount of storage allocated to this volume.
+<3> The volume type.
```
huffmanca: I think both references to **<3>** can be removed, as `cinder` is the only applicable value for these volumes.

Alternatively, you can enhance this description by saying that this must be `cinder` for OpenStack Cinder volumes?

modules/persistent-storage-cinder-creating-pv.adoc
```diff
@@ -0,0 +1,41 @@
...
+<4> Filesystem-type to mount.
```
huffmanca: Filesystem -> file system.

I'd rewrite this to say:

> The file system that is created when the volume is mounted for the first time.

modules/persistent-storage-cinder-creating-pv.adoc
huffmanca: It looks like we're defining a PersistentVolume here. I'd update the header of this block to be `.azure-advanced-disk-persistentvolume.yaml`.

modules/persistent-storage-cinder-provisioning.adoc
```diff
@@ -0,0 +1,14 @@
...
+[id="persistent-storage-cinder-provisioning_{context}"]
+= Dynamic provisioning with Cinder
+
+Storage must exist in the underlying infrastructure before it can be mounted as a volume in {product-title}.
+
+.Prerequisites
+
+* {product-title} configured for OpenStack
+* Cinder volume ID
+* `*PersistentVolume*` API.
```
huffmanca: This API is provided by OpenShift, and not something a customer would ever have to install or configure. It's more or less an implementation detail, and can be safely removed.

modules/persistent-storage-cinder-provisioning.adoc
```diff
@@ -0,0 +1,14 @@
+// Module included in the following assemblies:
+//
+// * storage/persistent_storage/persistent-storage-cinder.adoc
+
+[id="persistent-storage-cinder-provisioning_{context}"]
+= Dynamic provisioning with Cinder
```
huffmanca: I believe the following steps all detail manual provisioning, not dynamic, of volumes.

modules/persistent-storage-cinder-volume-security.adoc
```diff
@@ -0,0 +1,55 @@
...
+//
+// * storage/persistent_storage/persistent-storage-cinder.adoc
+
+[id="persistent-storage-cinder-volume-security_{context}"]
+= Cinder volume security
+
+If you use Cinder PVs in your application, configure security for their
+deployment configurations.
+
+. Create an SCC that uses the appropriate `*fsGroup*` strategy.
```
huffmanca: If there aren't instructions on how to accomplish this step we could move it to a Prerequisite.

huffmanca (commented):
A few comments, but overall it looks good. You'll also need to update a couple of additional lists. Check the following files as well:

> modules/storage-persistent-storage-pv.adoc (two spots, one for `Types of PVs`, and another for `Supported access modes for PVs`. Looks like the second has already been updated).
> storage/dynamic-provisioning.adoc (I ported the modules/dynamic-provisioning-cinder-definition.adoc over from 3.x, but it needs to be uncommented here for dynamic provisioning).

modules/persistent-storage-cinder-provisioning.adoc
bobfuru: Changing header to "Manual provisioning with Cinder"

modules/persistent-storage-cinder-volume-security.adoc
bobfuru: Moved to Prereq as suggested.

bobfuru (commented):
@huffmanca Applied your feedback, do you mind taking one more look? Thanks!

modules/persistent-storage-cinder-creating-pv.adoc
huffmanca: It's in the IBM Style Guide to use it as two words, thus the suggestion here. I've been guilty of writing it as one, though. :)

PR #17672, opened 2019-10-28
updates for BZ1638901

lamek:


install/configuring_inventory_file.adoc
```diff
@@ -353,7 +353,7 @@ environments, even those that have another version installed.
 
 |`openshift_master_default_subdomain`
 |This variable overrides the default subdomain to use for exposed
-xref:../architecture/networking/routes.adoc#architecture-core-concepts-routes[routes].
+xref:../architecture/networking/routes.adoc#architecture-core-concepts-routes[routes]. The value for this variable must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character.
```
bergerhoffer: Yeah this is a bit complicated. It's okay as is, but I'd suggest to break it up into two sentences. And maybe put `-` in backticks instead of quotes?

```
The value for this variable must consist only of lower case alphanumeric characters or dashes (`-`). It must also start and end with an alphanumeric character.
```
or 

```
The value for this variable must consist only of lower case alphanumeric characters or dashes (`-`). It may not start or end with a dash (`-`).
```

bergerhoffer (commented):
It's fine, but added some other possible suggestions to review.

Also don't forget to set a milestone of next release.

install/configuring_inventory_file.adoc
lamek: Got it. I broke it into 2 sentences. Slight nuance is that you must begin with a letter a-z (alphabetic), but you can end with a letter or number.

lamek:
For posterity - https://bugzilla.redhat.com/show_bug.cgi?id=1638901

PR #17678, opened 2019-10-28
10/28 feedback from Nelly wrt 2.1 release notes

ousleyp:
This came up in an email so there is no associated bug. 

Please cherrypick to enterprise-4.2 only. Thanks!


nellyc (approved):
LGTM

might be worth to create a revert PR once this is merged, so we will not forget to put it back with the 2.1.1 release in 2-3 weeks

lamek:
LGTM

bergerhoffer:
After chatting with @ousleyp, this should also be carried forward to 4.3. Cherry picking.

PR #17687, opened 2019-10-28
BZ-1743299: Clarifying how to find what service accounts an SCC is ap…

bergerhoffer:
…plied to

The PR for the 4.x version of !17686.

### DIFF
--- a/modules/security-context-constraints-command-reference.adoc
+++ b/modules/security-context-constraints-command-reference.adoc
@@ -48,16 +48,17 @@ restricted         false   []     MustRunAs   MustRunAsRange     MustRunAs   Run
 [id="examining-a-security-context-constraints-object_{context}"]
 == Examining an SCC
 
-To examine a particular SCC, use `oc get`, `oc describe`, or `oc
-edit`. For example, to examine the `restricted` SCC:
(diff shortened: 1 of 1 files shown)


modules/security-context-constraints-command-reference.adoc
```diff
@@ -84,6 +85,8 @@ Settings:
   Supplemental Groups Strategy: RunAsAny
     Ranges:				<none>
 ----
+<1> Lists which users and service accounts that the SCC is applied to.
```
bobfuru: In this case (and the one below), "that" sounds odd to me.
s/Lists which users and service accounts that the SCC is applied to./Lists which users and which service accounts the SCC is applied to.

modules/security-context-constraints-command-reference.adoc
```diff
@@ -84,6 +85,8 @@ Settings:
...
+<2> Lists which groups that the SCC is applied to.
```
bobfuru: similar to above
s/groups that the/groups the/

bobfuru (commented):
A couple small suggestions (take or leave).

bergerhoffer:
@barleyer Can you please review this? This to update the 4.1-4.3 docs.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1743299

PR #17688, opened 2019-10-28
removing mirror command output:

kalexand-rh:


kalexand-rh:
@jianlinliu, will you PTAL?

jianlinliu:
Looks good to me, though I am not sure why we remove these mirror command output.

kalexand-rh:
@jianlinliu, enough customers were trying to substitute data for the variables instead of using their output that engineering asked for it to be scrubbed.

PR #17694, opened 2019-10-28
Update versions in Doc Guidelines

codyhoag:
@vikram-redhat can you confirm these versions, and merge if correct? I updated based on the [_distro_map.yml](https://github.com/openshift/openshift-docs/blob/master/_distro_map.yml) file. Thanks!

### DIFF
--- a/contributing_to_docs/doc_guidelines.adoc
+++ b/contributing_to_docs/doc_guidelines.adoc
@@ -198,11 +198,11 @@ possible values for `{product-title}` and `{product-version}`.
 
 |`openshift-origin`
 |OKD
-|1.0, 1.1, 1.2, 1.3, 1.4, 1.5, 3.6, 3.7, 3.9, 3.10, 3.11, 4.0, Latest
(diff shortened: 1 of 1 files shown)


vikram-redhat:
LGTM

vikram-redhat:
Doesn't require any CP.

codyhoag:
@vikram-redhat do we ignore the contributing guides on legacy branches (e.g., [here](https://github.com/openshift/openshift-docs/blob/enterprise-4.1/contributing_to_docs/doc_guidelines.adoc#product-name--version))? I had marked for legacy branches because the versions were off in those too. I see where it makes sense to only maintain the `master` version; just wanted to confirm for future reference.

vikram-redhat:
@codyhoag correct. Those legacy guidelines shouldn't be linked from anywhere else. If you find them linked, go ahead and delete. Just the master is correct (we could probably delete them from the branches).

codyhoag:
@vikram-redhat got it; yep, makes sense.

PR #17698, opened 2019-10-29
added online pro and starter

vikram-redhat:


vikram-redhat:
@sheriff-rh @wgordon17 FYI

PR #17716, opened 2019-10-29
modules/*proxy*: httpsProxy supports https:// too

wking:
Drop the wording from aa39bf33b4 (#16635), because the network operator [explicitly supports both][1] the `http` and `https` schemes since openshift/cluster-network-operator@42dbcf8955 (openshift/cluster-network-operator#245), which landed before release-4.2 split off from the network operator's master.  This should be backported to 4.2 as well.

CC @bergerhoffer, @danehans

[1]: https://github.com/openshift/cluster-network-operator/blob/42dbcf89559c543b656c3f3eb72e46a4e954b16d/pkg/controller/proxyconfig/validation.go#L37-L38

### DIFF
--- a/modules/installation-configure-proxy.adoc
+++ b/modules/installation-configure-proxy.adoc
@@ -50,7 +50,7 @@ apiVersion: v1
 baseDomain: my.domain.com
 proxy:
   httpProxy: http://<username>:<pswd>@<ip>:<port> <1>
-  httpsProxy: http://<username>:<pswd>@<ip>:<port> <2>
+  httpsProxy: https://<username>:<pswd>@<ip>:<port> <2>
   noProxy: example.com <3>
 additionalTrustBundle: | <4>
     -----BEGIN CERTIFICATE-----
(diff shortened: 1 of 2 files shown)


wking:
And I'm [still][1] [not clear][2] on why you can't use `https` for `httpProxy`, but at least that documented restriction is reflected in the current network-operator validation.

[1]: https://github.com/openshift/cluster-network-operator/commit/095cd157b3a2f8764f7ee090d09a3a92cf177ef8#r35259609
[2]: https://github.com/openshift/openshift-docs/pull/16635/files#r325784121

wking:
Also discussion [here][1].  It's not clear to me if/where the "old Git's don't like HTTPS proxies" limitation is tracked outside of [the Bugzilla][2], but I think calling it out in as a build limitation (for some builds?  All builds?) makes more sense than saying that it's completely unsupported (maybe folks are running clusters where HTTPS proxies are important but which won't use builds?).

[1]: https://github.com/openshift/openshift-docs/pull/16635#discussion_r327846123
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1750650

wking:
[Here][1] are the release docs for the limitation.  Not sure if/how we want to reference those from this master-branch PR.  Thoughts?

[1]: https://github.com/openshift/openshift-docs/pull/16972/files#diff-e21937773545bc689525f2a41cb1cab6R997

kalexand-rh:
@danehans, will you PTAL?

vikram-redhat:
@danehans BUMP.

@wking can you rebase?

wking:
#25438 landed the softening, so we don't need this PR anymore.

/close

PR #17733, opened 2019-10-30
Bug 1757866: Fixed documentation on fetching image-registry metrics.

ricardomaraschini:
Old sample was wrong, showing metrics that do not exist  on version 4.1.

### DIFF
--- a/modules/registry-accessing-metrics.adoc
+++ b/modules/registry-accessing-metrics.adoc
@@ -25,19 +25,28 @@ or using the cluster role.
 . Run a metrics query, for example:
 +
 ----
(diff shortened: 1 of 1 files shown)


bmcelvee:
Thanks, @ricardomaraschini! LGTM. Since this is the same for 4.1 and 4.2 (and I'm assuming it's the same for 4.3) I'm going to cherrypick this change to master and 4.2 and close https://github.com/openshift/openshift-docs/pull/17734. 

@vikram-redhat I don't think this requires QE review, but wanted to check with you first. Thanks!

vikram-redhat:
@bmcelvee yes works for me.

PR #17742, opened 2019-10-30
Changing to UI-based namespace creation flow

ousleyp:
cc @alexxa @nellyc @rthallisey 

Updating the namespace creation flow for CNV 2.1.

(for enterprise-4.2 and 4.3)

### DIFF
--- a/modules/cnv-preparing-to-install.adoc
+++ b/modules/cnv-preparing-to-install.adoc
@@ -14,9 +14,11 @@ Before deploying {CNVProductName}, create a namespace that is named
 
 .Procedure
 
-* Create the `openshift-cnv` namespace by running the following
-command:
-+
-----
-$ oc create ns openshift-cnv
-----
+. From the {product-title} web console, navigate to the
+*Administration* -> *Namespaces* page.
+
+. Click *Create Namespace*.
+
+. In the *Name* field, type `openshift-cnv`.
+
+. Click *Create*.


modules/cnv-preparing-to-install.adoc
```diff
@@ -14,9 +14,11 @@ Before deploying {CNVProductName}, create a namespace that is named
...
 .Procedure
 
-* Create the `openshift-cnv` namespace by running the following
-command:
-+
-----
-$ oc create ns openshift-cnv
-----
\ No newline at end of file
+* From the {product-title} web console, navigate to the
```
bergerhoffer: These should be numbered bullets since there is more than one step now.

bergerhoffer (commented):
Just one thing, and then a question on the version(s) this applies for. Says only 4.2, but should this also be carried forward for 4.3?

If it is for 4.2 only, then it should be submitted directly to enterprise-4.2 (and not master, because master is the latest and greatest, like 4.3)

alexxa:
@bergerhoffer I would merge it to 4.2 only. There is a BZ on OCP which is supposed to be fixed in 4.3 release and then creating NS procedure will be removed completely from CNV docs.

ousleyp:
> @bergerhoffer I would merge it to 4.2 only. There is a BZ on OCP which is supposed to be fixed in 4.3 release and then creating NS procedure will be removed completely from CNV docs.

@alexxa, actually I just realized it should be for both 4.2 and 4.3 to align with the OCP docs process. We can still definitely remove it for 4.3, so don't worry about that. :)

PR #17749, opened 2019-10-30
[WIP] Reverting the change from PR 17678

ousleyp:
As discussed in https://github.com/openshift/openshift-docs/pull/17678, this should be merged around the time of the CNV 2.1.1 release. cc @nellyc


ousleyp:
@nellyc Should we bring back the mention of VMware VMs?

PR #17760, opened 2019-10-30
BZ1767460: Removing the step to add the cluster-admin role.

huffmanca:
Removing the instruction to add the cluster-admin role to the LocalStorage Operator. This was needed for 4.3 clusters due to https://bugzilla.redhat.com/show_bug.cgi?id=1766856 ; however, I've confirmed that this step is not required for 4.2 clusters.

This is for OCP 4.2 and 4.3.


bobfuru:
@liangxia PTAL, thanks!

PR #17780, opened 2019-10-31
Removing notes added for BZ 1767167

ousleyp:
This PR should be merged when [BZ 1767167](https://bugzilla.redhat.com/show_bug.cgi?id=1767167) is fixed. The fix is expected to go out with the next z-stream release after 2.1.

These notes were originally added in https://github.com/openshift/openshift-docs/pull/17773.

attn: @nellyc @alexxa


aburdenthehand:
LGTM
Cherrypick to enterprise-4.3 only

PR #17793, opened 2019-10-31
Pod disruption budgets can prevent nodes being drained

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1763798

### DIFF
--- a/modules/nodes-pods-pod-disruption-configuring.adoc
+++ b/modules/nodes-pods-pod-disruption-configuring.adoc
@@ -22,16 +22,39 @@ kind: PodDisruptionBudget
 metadata:
   name: my-pdb
 spec:
-  selector:  <2>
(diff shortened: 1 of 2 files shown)


modules/nodes-pods-pod-disruption-configuring.adoc
```diff
@@ -22,16 +22,38 @@ kind: PodDisruptionBudget
...
+----
++
+[source,yaml]
+----
+apiVersion: policy/v1beta1 <1>
+kind: PodDisruptionBudget
+metadata:
+  name: my-pdb
+spec:
+  maxUnavailable:  <3>
```
rphillips: this example looks wrong...

```
spec:
  maxUnavailable: [some count]
  selector:
    matchLabels:
      foo: bar
```

PR #17711, opened 2019-10-29
Include Online sections

wgordon17:

PR #17795, opened 2019-10-31
Update OpenShift Online redirects

wgordon17:


wgordon17:
@sheriff-rh Are you able to build a preview of this for `online/starter`? We obviously can't test the redirects with that preview, but we should at least be able to see the "Get Support" change 👍🏻

wgordon17:
@vikram-redhat Regex update on the `online/` redirects...because right now we're not catching redirects for things like https://docs.openshift.com/online/cli_reference/get_started_cli.html. Hopefully this fixes that

sheriff-rh:
Here ya go @wgordon17 : 
http://file.rdu.redhat.com/~antaylor/patch-7/support/getting-support.html

If more changes are committed, shoot me a message (or ping me again here) and i'll update the build preview.

wgordon17:
Cool, I think that looks good! Based on @vikram-redhat's previous PR (#17711) could you please tag for 4.2 and 4.3, and then merge if you don't see any issues?

PR #17799, opened 2019-10-31
Math to calculate Logging storage doesn't match example

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1759329

### DIFF
--- a/install_config/aggregate_logging_sizing.adoc
+++ b/install_config/aggregate_logging_sizing.adoc
@@ -244,7 +244,7 @@ Baseline (256 characters per second -> 15KB/min)
 1 kibana
 1 curator
 1 fluentd
-| 6 pods total: 90000 x 1440 = 128,6 MB/day
+| 6 pods total: 90000 x 1440 = 129,6 MB/day
 
 |3 es
 1 kibana


mburke5678:
@jcantrill @richm @ewolinetz can anyone help with this BZ? 

https://bugzilla.redhat.com/show_bug.cgi?id=1759329
```
Document URL: 
https://docs.openshift.com/container-platform/3.11/install_config/aggregate_logging_sizing.html#install-config-aggregate-logging-sizing-guidelines-storage

Section Number and Name: 
Configuring Clusters
Aggregate Logging Sizing Guidelines
Storage Considerations

Describe the issue: 
The documentation provides an example chart of how much storage is used by the logging cluster per day with a specific layout in mind (3 es, 1 kibana, 1 curator, and 11 fluentd handles ~345.6 MB per day) and then provides a set of equations to help the user calculate the amount of storage needed for their Own cluster.

Suggestions for improvement: 
Clarify the math and/or provide more thorough and accurate examples.
```

ewolinetz:
I'm currently reading through the docs... and they are confusing to follow. I'll take a stab it

install_config/aggregate_logging_sizing.adoc
```diff
@@ -244,7 +244,7 @@ Baseline (256 characters per second -> 15KB/min)
 1 kibana
 1 curator
 1 fluentd
-| 6 pods total: 90000 x 1440 = 128,6 MB/day
```
ewolinetz: I'm not sure where `90000` came from.. but `1440` is `min/day`.. so if we go with the assumption of `256 B/sec` that gives us `15360 B/min`.

If we do `15360 * 6` we get `92160` which is more than `90000` but also this is incorrect. We only have a single `fluentd` in this example... so it should be `15360 x 1440 = 22118400 B/day` -> about `21,1 MB/day`. (We only care about Fluentd since it is the only pod there actually collecting/processing logs).

This table doesn't take the limitation of a single Fluentd into consideration, but I'm not sure that matters in this case since we should actually estimate high. But that does make me think we don't care about the number of collectors when calculating but rather the number of pods that we're collecting logs for.

If we knew the limitation of fluentd's ingestion then we could also make a recommendation on the number of fluentd required for a given number of nodes (then customers could try to balance number of nodes vs number of pods with logs etc) -- but this is likely outside the scope of this document.

ewolinetz: thoughts @jcantrill @richm @lukas-vlcek ?

mburke5678: @jcantrill @richm @lukas-vlcek ^^

richm: We do know something from the latest 4.2 performance/scale testing:
```
Fluentd was again able to reliably and consistently index 2500 1KB messages per second from a single node.  This is consistent with 4.1 results
```
That works out to be `2500 * 86400 = 216M` records per day, and `2500 * 86400 * 1024 = 221GB` of data per day.  And this is per node.  You'll likely have to tune up Elasticsearch quite a bit: maxed out memory - 62Gi, 16 cpu cores or more, fast local disks, and possibly some other tuning.

PR #17840, opened 2019-11-04
4.2 Release Notes modified for dedicated

sheriff-rh:
Changes based on: https://docs.google.com/document/d/1th2emY24A7kgvTxUM5PUEd5-x0gzUH-EPEGyjjI1EFM/edit#

@wgordon17 PTAL
@patrickstrick  @vikram-redhat FYI
@ahardin-rh  would you mind peer reviewing? 
Local build: http://file.rdu.redhat.com/~antaylor/110419/4.2dedicatedRNs/release_notes/ocp-4-2-release-notes.html

### DIFF
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -39,12 +39,13 @@ Topics:
 ---
 Name: Release notes
 Dir: release_notes
-Distros: openshift-enterprise
+Distros: openshift-enterprise,openshift-dedicated
 Topics:
 - Name: OpenShift Container Platform 4.2 release notes
   File: ocp-4-2-release-notes
 - Name: Versioning policy
   File: versioning-policy
+  Distros: openshift-enterprise
 ---
 Name: Architecture
 Dir: architecture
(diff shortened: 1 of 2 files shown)


ahardin-rh:
If I recall correctly, we previously omitted the bug list and Tech Preview section because they are OCP-centric. Right now, the only new features and enhancements we highlight here are for Dev experience and Web console. I would expect there to be more, but I defer to Will.

wgordon17:
@ahardin-rh Yeah, good call wrt TP section. @sheriff-rh can you please remove that section? Otherwise LGTM

I think the bug list is still beneficial for OSD customers to see what was fixed by the upgrade...in case they were waiting on something.

wgordon17:
Also, any change we can get the left nav bar to agree with the topic header?

sheriff-rh:
@wgordon17  I removed the TP table and that is reflected in the latest version of the build. I tried to change the title for the release notes page to `{product-title}`, which is populated per build version, but that broke things. I'll get with @ahardin-rh  and figure out a reasonable solution and title.

wgordon17:
Awesome 👍🏻 Thanks!

_topic_map.yml
```diff
@@ -39,12 +39,13 @@ Topics:
 ---
 Name: Release notes
 Dir: release_notes
-Distros: openshift-enterprise
+Distros: openshift-enterprise, openshift-dedicated
```
adellape: @sheriff-rh The space here might be breaking it.

sheriff-rh: good catch @adellape  ! just updated.

sheriff-rh:
Closed in favor of #17888 .

PR #17887, opened 2019-11-05
OSDOCS-729_731: Update block volume support table for Cinder, iSCSI

bobfuru:
In OCP 4.3, Cinder is moving to tech preview ([OSDOCS-729](https://jira.coreos.com/browse/OSDOCS-729)) and iSCSI is moving to GA ([OSDOCS-731](https://jira.coreos.com/browse/OSDOCS-731)). 

This PR modifies the [Block volume support](https://docs.openshift.com/container-platform/4.2/storage/understanding-persistent-storage.html#block-volume-support_understanding-persistent-storage) table accordingly. (Cinder - manual/dynamic provisioning, not fully supported; iSCSI - manual provisioning only, full support.)

### DIFF
--- a/modules/storage-persistent-storage-block-volume.adoc
+++ b/modules/storage-persistent-storage-block-volume.adoc
@@ -29,14 +29,11 @@ The following table displays which volume plug-ins support block volumes.
 |Volume Plug-in  |Manually provisioned  |Dynamically provisioned |Fully supported
 |AWS EBS  | ✅ | ✅ | ✅
 |Azure Disk | ✅ | ✅ | ✅
-
+|Cinder | ✅ | ✅ |
 |Fibre Channel | ✅ | |
-
 |GCP | ✅ | ✅ | ✅
 |HostPath | | |
-
-|iSCSI | ✅ | |
-
+|iSCSI | ✅ | | ✅
 |Local Volumes | ✅ | ✅ | ✅
 |NFS | | |
 |VMware vSphere  | ✅ | ✅ | ✅


bobfuru:
@liangxia and @bertinatto PTAL 

This PR documents work being done in [STOR-234](https://jira.coreos.com/browse/STOR-234) (iSCSI) and [STOR-235](https://jira.coreos.com/browse/STOR-235) (Cinder).

modules/storage-persistent-storage-block-volume.adoc
```diff
@@ -29,13 +29,13 @@ The following table displays which volume plug-ins support block volumes.
...
 |Azure Disk | ✅ | ✅ | ✅
-
+|Cinder | ✅ | |
 |Fibre Channel | ✅ | |
 
 |GCP | ✅ | ✅ | ✅
 |HostPath | | |
 
-|iSCSI | ✅ | |
+|iSCSI | ✅ | ✅ | ✅
```
liangxia: Dynamic provision for iSCSI is not supported in OCP.

bobfuru: > Dynamic provision for iSCSI is not supported in OCP.

@bertinatto @liangxia The three options in the docs table are `Manually provisioned`, `Dynamically provisioned`, and `Fully supported`. Is it correct then to label iSCSI as `Fully supported` if it does not support dynamic provisioning?

bertinatto: I'm not sure what `Fully supported` means, but from the notes in this doc I think that, if unchecked, it means Tech Preview. If that's the case, both Cinder and iSCSI are fully supported in 4.3.

Other than that, I think Cinder supports both manual and dynamic [0] provisioning. And as @liangxia pointed out, iSCSI supports only manual provisioning.

[0] https://docs.okd.io/latest/install_config/persistent_storage/persistent_storage_cinder.html

bobfuru: Thanks! I have removed `Dynamically provisioned` from the `iSCSI` row. Also, all three options are now included in the `Cinder` row.

@huffmanca Can you confirm that `Fully supported`, as @bertinatto describes, is correct (checked = GA)?

bobfuru: @liangxia PTAL again to triple-check I have the right boxes checked, thanks!

liangxia: @bobfuru  As stated in [PROD-1901](https://jira.coreos.com/browse/PROD-1901),
We need to move the support levels as follows:
    Raw Block with iSCSI to GA
    Raw Block with Cinder to Tech Preview

bertinatto: That's right, I made a mistake in https://github.com/openshift/openshift-docs/pull/17887#discussion_r343170784, Cinder is Tech Preview. Sorry.

bobfuru: Thanks for the catch!

I removed the `Fully supported` checkmark from Cinder row, indicating that it is Tech Preview. Left iSCSI as is, indicating that it supports manual provisioning, does not support dynamic provisioning, and is fully supported (GA).

adellape (approved):
:+1:

PR #16004, opened 2019-07-23
VPA: view Pod sizing recommendations

mburke5678:
Moved out of 4.2

https://jira.coreos.com/browse/OSDOCS-501
https://jira.coreos.com/browse/PROD-1182

PR #17926, opened 2019-11-06
4.5 Vertical Pod Autoscaler

mburke5678:
New PR for https://github.com/openshift/openshift-docs/pull/16004

https://jira.coreos.com/browse/OSDOCS-501
https://jira.coreos.com/browse/PROD-1182
https://issues.redhat.com/browse/OCPNODE-212

### DIFF
--- a/modules/nodes-pods-vertical-autoscaler-about.adoc
+++ b/modules/nodes-pods-vertical-autoscaler-about.adoc
@@ -1,0 +1,23 @@
+// Module included in the following assemblies:
+//
+// * nodes/nodes-vertical-autoscaler.adoc
+
+[id="nodes-pods-vertical-autoscaler-about_{context}"]
+= About the Vertical Pod Autoscaler Operator
+
+The Vertical Pod Autoscaler Operator (VPA) is implemented as an API resource and a custom resource (CR). The CR determines the actions the Vertical Pod Autoscaler Operator should take with the Pods associated with a specific workload object, such as a Daemonset, ReplicationController, and so forth. 
+
+The VPA automatically computes historic and current CPU and memory usage for the containers in those Pods and can use this data to automatically re-deploy Pods with optimized resource limits and requests to ensure that these Pods are operating efficiently at all times. When re-deploying Pods, the VPA honors any Pod Disruption Budget set for applications. If you do not want the VPA to automatically re-deploy Pods, you can use this resource information to manually update the Pods as needed.
+
+When configured to update Pods automatically, the VPA reduces resources for Pods that are requesting more resources then they are using and increase resources for Pods that are not requesting enough. 
+
+For example, if you have a pod that uses 50% of the CPU but only requests 10%, the VPA determines that the Pod is consuming more CPU than requested and restarts the Pods with higher resources. 
+
+For developers, the VPA helps ensure their Pods stay up during periods of high demand by scheduling Pods onto nodes so that appropriate resources are available for each Pod. 
+
+Administrators can use the VPA to better utilize cluster resources, such as preventing Pods from reserving more CPU resources than needed. The VPA monitors the resources that workloads are actually using and adjusts the resource requirements so capacity is available to other workloads. The VPA also maintains the ratios between limits and requests that are specified in initial container configuration. 
+
+[NOTE]
+====
+If you stop running the VPA or delete a specific VPA CR in your cluster, the resource requests for the Pods already modified by the VPA do not change. Any new Pods get the resources defined in the workload object, not the previous recommendations made by the VPA.
+====
--- a/modules/nodes-pods-vertical-autoscaler-configuring.adoc
+++ b/modules/nodes-pods-vertical-autoscaler-configuring.adoc
@@ -1,0 +1,102 @@
+// Module included in the following assemblies:
+//
+// * nodes/nodes-vertical-autoscaler.adoc
+
+[id="nodes-pods-vertical-autoscaler-configuring_{context}"]
+= Using the Vertical Pod Autoscaler Operator
+
+You can use the Vertical Pod Autoscaler Operator (VPA) by creating a VPA custom resource (CR). The CR indicates which Pods it should analyze and determines the actions the VPA should take with those Pods. 
+
+.Procedure
+
+To create a VPA CR for a specific workload object:
+
+. Change to the project where the workload object you want to scale is located.
+
+.. Create a VPA CR YAML file:
++
+[source, yaml]
+----
+apiVersion: autoscaling.k8s.io/v1
+kind: VerticalPodAutoscaler
+metadata:
+  name: vpa-recommender
+spec:
+  targetRef:
+    apiVersion: "apps/v1"
+    kind:       Deployment <1>
+    name:       frontend <2>
+  updatePolicy:
+    updateMode: "Auto" <3>
+  resourcePolicy: <4>
+    containerPolicies:
+    - containerName: my-opt-sidecar
+      mode: "Off"
+----
+<1> Specify the type of workload object you want this VPA to manage: `Deployment`, `StatefulSet`, `Job`, `DaemonSet`, `ReplicaSet`, or `ReplicationController`.
+<2> Specify the name of an existing workload object you want this VPA to manage.
+<3> Specify the VPA mode:
+* `auto` to automatically apply the recommended resources on Pods associated with the controller. The VPA terminates existing Pods and creates new Pods with the recommended resource limits and requests.
+* `recreate` to automatically apply the recommended resources on Pods associated with the workload object. The VPA terminates existing Pods and creates new Pods with the recommended resource limits and requests. The `recreate` mode should be used rarely, only if you need to ensure that the pods are restarted whenever the resource request changes. 
+* `initial` to automatically apply the recommended resources when Pods associated with the workload object are created. The VPA does not update the Pods as it learns new resource recommendations.
+* `off` to only generate resource recommendations for the Pods associated with the workload object. The VPA does not update the Pods as it learns new resource recommendations and does not apply the recommendations to new Pods. 
+<4> Optional. Specify the containers you want to opt-out and set the mode to `Off`.
+
+
+.. Create the VPA CR:
++
+----
+$ oc create -f <file-name>.yaml
+----
++
+After a few moments, the VPA learns the resource usage of the containers in the Pods associated with the workload object.
++
+You can view the VPA recommendations using the following command:
+
+----
+$ oc get vpa <vpa-name> --output yaml
+----
++
+The output shows the recommendations for CPU and memory requests, similar to the following:
++
+----
+...
+status:
+
+...
+
+  recommendation:
+    containerRecommendations:
+    - containerName: frontend
+      lowerBound: <1>
+        cpu: 25m
+        memory: 262144k
+      target: <2>
+        cpu: 25m
+        memory: 262144k
+      uncappedTarget: <3>
+        cpu: 25m
+        memory: 262144k
+      upperBound: <4>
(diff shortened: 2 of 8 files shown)


mburke5678:
Per Greg B in Jira _this won't be in OpenShift 4.3 _.

https://jira.coreos.com/browse/PROD-1182?focusedCommentId=142416&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-142416

mburke5678:
@joelsmith Can you please take a look at your earliest convenience? We have docs freeze on June 4 and I would like to get VPA done and into the 4.5 docs. 

I have a couple of questions:

1. Should we mention the `recreate` updateMode? It is the same as `auto`, but the upstream docs suggest to use `recreate` should be used _rarely_.

2. I am under the impression that the VPA works on pods created by a specific Deployment or ReplicaSet. Does it work on all pods in a project, regardless of the controller? I had one case where I lunched pods using a second Deployment and the VPA seems to have launched the pods with the recommended resources. A third Deployment was not launched with the recommended resources, but the VPA calculated recommendations.

```
    containerRecommendations:
    - containerName: hello-world
      lowerBound:
        cpu: 10m
        memory: "87381333"
      target:
        cpu: 11m
        memory: "87381333"
      uncappedTarget:
        cpu: 11m
        memory: "87381333"
      upperBound:
        cpu: 183m
        memory: 191500k
    - containerName: hello-whirrled
      lowerBound:
        cpu: 10m
        memory: "87381333"
      target:
        cpu: 11m
        memory: "87381333"
      uncappedTarget:
        cpu: 11m
        memory: "87381333"
      upperBound:
        cpu: 227m
        memory: "238349315"
    - containerName: hullo-world
      lowerBound:
        cpu: 8m
        memory: "87381333"
      target:
        cpu: 11m
        memory: "87381333"
      uncappedTarget:
        cpu: 11m
        memory: "87381333"
      upperBound:
        cpu: 1991m
        memory: 2081500k
----

$ oc get Deployments
NAME             READY   UP-TO-DATE   AVAILABLE   AGE
hello-whirrled   2/2     2            2           32m
hello-world      2/2     2            2           34m
hullo-world      2/2     2            2           6m46s

```
The `hello-whirrled` Deployment is not in the VPA CR, but it launched with recommended values:
```
  containers:
  - image: quay.io/openshift-knative/knative-eventing-sources-event-display:v0.13.2
    imagePullPolicy: IfNotPresent
    name: hello-whirrled
    ports:
    - containerPort: 8080
      protocol: TCP
    resources:
      limits:
        cpu: 24m
        memory: 625Mi
      requests:
        cpu: 12m
        memory: 131072k
```

modules/nodes-pods-vertical-autoscaler-configuring.adoc
```diff
@@ -0,0 +1,49 @@
...
+// * nodes/nodes-vertical-autoscaler.adoc
+
+[id="nodes-pods-vertical-autoscaler-configuring_{context}"]
+= Using the Vertical Pod Autoscaler
+
+You can use the Vertical Pod Autoscaler Operator (VPA) by creating a Vertical Pod Autoscaler Custom Resource (CR). The CR determines which Deployment or ReplicaSet it should analyze and determines the actions the VPA should take with the Pods associated with that Deployment or ReplicaSet. 
+
+.Procedure
+
+To create a VPA Custom Resource for a specific controller, either a Deployment or ReplicaSet:
```
joelsmith, line 12: I haven't tested it, but I'm pretty sure VPA can work with pods from any controller (Deployment and ReplicaSet, but also DaemonSet, CronJob, DeploymentConfig, ReplicationController, etc.) so I don't know whether to call that out, but I don't think we want to make it sound like it's limited to rs and deployments.

I'll do some testing to make sure.

joelsmith (changes requested):
Looking really good overall. Just a few minor things to address. Thanks!

modules/nodes-pods-vertical-autoscaler-configuring.adoc
mburke5678: @joelsmith I tried each of the object types in the console under Workloads that spawn pods: Deployment, Deployment Config, Statefulset, Job, DaemonSet, ReplicaSet, or ReplicationController. I used the recommender. All generated recommendations except for Job (which makes sense as the job completed before the VPA could get its info, I assume) and DeplotmentConfig. For the DC, and I tried a few, I got:
```
message: 'Cannot read targetRef. Reason: Unhandled targetRef apps/v1 / DeploymentConfig
      / example, last error no matches for kind "DeploymentConfig" in group "apps"'
```
Does this message make sense?

mburke5678:
@weinliu PTAL

modules/nodes-pods-vertical-autoscaler-configuring.adoc
```diff
@@ -0,0 +1,49 @@
...
+kind: VerticalPodAutoscaler
+metadata:
+  name: vpa-recommender
+spec:
+  targetRef:
+    apiVersion: "apps/v1"
+    kind:       Deployment <1>
+    name:       frontend <2>
+  updatePolicy:
+    updateMode: "Auto" <3>
```
weinliu, line 30: @joelsmith  Do we get `updateMode` supported? I did no see this in the demo

modules/nodes-pods-vertical-autoscaler-install.adoc
```diff
@@ -0,0 +1,52 @@
...
+// * nodes/nodes-vertical-autoscaler.adoc
+
+[id="nodes-pods-vertical-autoscaler-install_{context}"]
+= Installing the Vertical Pod Autoscaler
+
+You can use the {product-title} console to install the Vertical Pod Autoscaler (VPA) Operator.
+
+.Procedure
+
+. In the {product-title} web console, click *Catalog* -> *OperatorHub*.
```
weinliu, line 12: I do not see Calalog on a 4.6 env  now

joelsmith: Me either. On my 4.5 cluster, it's *Operators* -> *OperatorHub*. @mburke5678 could you double check whether your test cluster says "Catalog" or "Operators" for the console section?

modules/nodes-pods-vertical-autoscaler-configuring.adoc
joelsmith: Yes, it is supported. I left it out of the demo because it is short, but we should probably have some test cases to verify that it works. Thanks for covering it in the docs @mburke5678

joelsmith (approved):
lgtm

mburke5678:
@weinliu PTAL

modules/nodes-pods-vertical-autoscaler-configuring.adoc
mburke5678: @weinliu Can you comment on which controllers the VPA works with?

weinliu: Deployment verified to be working

weinliu:
LGTM, great to go

modules/nodes-pods-vertical-autoscaler-about.adoc
```diff
@@ -0,0 +1,21 @@
+// Module included in the following assemblies:
+//
+// * nodes/nodes-vertical-autoscaler.adoc
+
+[id="nodes-pods-vertical-autoscaler-about_{context}"]
+= About the Vertical Pod Autoscaler
+
+A Vertical Pod Autoscaler (VPA) uses a Mutating Admission Webhook to automatically compute historic and current CPU and memory usage for the containers in Pods associated with specific Pod controllers, such as Deployments, StatefulSets, Jobs, DaemonSets, ReplicaSets, or ReplicationControllers. You can configure a VPA to use the data it learns to automatically re-deploy Pods with optimized resource limits and requests to ensure that these Pods are operating efficiently at all times. When re-deploying Pods, a VPA honors any Pod Disruption Budget set for your applications. If you do not want a VPA to automatically re-deploy Pods, you can use this resource information to manually update the Pods as needed.
```
joelsmith, line 8: I don't think that the webhook should be mentioned here.  Maybe instead, it could say:

> The Vertical Pod Autoscaler (VPA) monitors pod metrics to automatically ...

Looking at upstream docs for HPA, they do a nice job of explaining the relationship between the controllers and the API object (which they call a resource). See https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/ for what I'm talking about. I like what they say here:

> The Horizontal Pod Autoscaler is implemented as a Kubernetes API resource and a controller. The resource determines the behavior of the controller. The controller periodically adjusts the number of replicas in a replication controller or deployment to match the observed average CPU utilization to the target specified by user.

So they define the HPA as both things, which I think makes a lot of sense. So maybe we should look at the VPA similarly, with the VPA being the combination of controllers and the object together.

mburke5678: @joelsmith 
_The Vertical Pod Autoscaler Operator (VPA) is implemented as an API resource and a custom resource (CR) object. The CR determines the actions the VPA should take with the Pods associated with a specific workload object, such as a Daemonset, ReplicationController, and so forth._ 

Also, since the `auto` and `recreate` modes essentially do the same thing currently, does it make sense to list both options? Especially where the Kubernetes doc suggests that `recreate` **should be used rarely**.  Should we remove the `recreate` option?
@rphillips

modules/nodes-pods-vertical-autoscaler-configuring.adoc
```diff
@@ -0,0 +1,101 @@
...
+  resourcePolicy: <4>
+    containerPolicies:
+    - containerName: my-opt-sidecar
+      mode: "Off"
+----
+<1> Specify the type of workload object you want this VPA to manage: `Deployment`, `StatefulSet`, `Job`, `DaemonSet`, `ReplicaSet`, or `ReplicationController`.
+<2> Specify the name of an existing workload object you want this VPA to manage.
+<3> Specify the VPA mode:
+* `auto` to automatically apply the recommended resources on Pods associated with the controller. The VPA terminates existing Pods and creates new Pods with the recommended resource limits and requests.
```
lamek, line 39: I didn't know a bullet list worked inside callout explanations. Looks great.

modules/nodes-pods-vertical-autoscaler-about.adoc
```diff
@@ -0,0 +1,23 @@
...
+The Vertical Pod Autoscaler Operator (VPA) is implemented as an API resource and a custom resource (CR). The CR determines the actions the Vertical Pod Autoscaler Operator should take with the Pods associated with a specific workload object, such as a Daemonset, ReplicationController, and so forth. 
+
+The VPA automatically computes historic and current CPU and memory usage for the containers in those Pods and can use this data to automatically re-deploy Pods with optimized resource limits and requests to ensure that these Pods are operating efficiently at all times. When re-deploying Pods, the VPA honors any Pod Disruption Budget set for applications. If you do not want the VPA to automatically re-deploy Pods, you can use this resource information to manually update the Pods as needed.
+
+When configured to update Pods automatically, the VPA reduces resources for Pods that are requesting more resources then they are using and increase resources for Pods that are not requesting enough. 
+
+For example, if you have a pod that uses 50% of the CPU but only requests 10%. The VPA determines that the Pod is consuming more CPU than requested and restarts the Pods with higher resources.
```
lamek, line 14: Double check - did you want this to be a single sentence?

lamek (commented):
This looks really good. At I didn't understand why "Using the VPA" was a separate section from "About using VPA", but after reading through I can see how one goes into the execution of using the CR, and the other covers all the possible options you could use inside the CR.

1 small nit and LGTM.

PR #17931, opened 2019-11-06
Add 3.11 migration documentation

mburke5678:
https://jira.coreos.com/browse/LOG-549

### DIFF
--- a/upgrading/automated_upgrades.adoc
+++ b/upgrading/automated_upgrades.adoc
@@ -735,6 +735,71 @@ upgrade the component.
 To upgrade an existing EFK logging stack deployment, you review your parameters
 and run the *_openshift-logging/config.yml_* playbook.
 
+The EFK upgrade also upgrades Elasticsearch from version 2 to version 5. For important information on changes in Elasticsearch 5, you should review the link:https://www.elastic.co/guide/en/elasticsearch/reference/5.5/breaking-changes-5.0.html[Elasticsearch breaking changes].
+
+It is important to note that Elasticsearch 5 has some significant changes to the index structures. Previously, Elasticsearch permitted a dot character, `.`, in field names.
+In version 5, Elasticsearch interprets any dot in an Elasticsearch field name as nested structure. If you have a field with a dot, the string after the dot is interpreted as the type of field, leading to mapping conflicts during the upgrade. 
+
+To help identify potential conflicts, {product-title} provides a script that examines your Elasticsearch fields to determine if any fields contain a dot in the name.
+
+For example, the following fields were allowed in Elasticsearch 2:
+
+----
+{
+  "field": 123  // "field" is of type number
+}
+
+// Any dot in field name is treated as any other valid character in the field name.
+// It is just part of the field name.
+{
+  "field.name": "Bob"  // "field.name" is of type String 
+----
+
+In Elasticsearch 5 and higher the `field` string would become the field and the `name` string would become a type for the field: 
+
+----
+{
+  "field": 123  // "field" is of type number
+}
+
+// Any dot in field name is always interpreted as nested structure.
+{
+  "field" : {  // "field" is of type Object
+    "name": "Bob"  // "name" is of type String
+  }
+}
+----
+
+Upgrading in this case would result in the `field` field having two different types, which is not permitted. 
+
+If you need to keep these conflicting indices, you need to reindex the data and change the documents to get rid of conflicting data structure. For more information, see link:https://www.elastic.co/guide/en/elasticsearch/reference/2.4/dots-in-names.html#_upgrading_fields_with_dots_to_5_x[Upgrading fields with dots to 5.x]. 
+
+[[upgrading-efk-logging-stack-script]]
+==== Determining if fields have dots in field names
+
+You can run the following script to determine if your indices contain any fields with a dot in the name.
+
+[NOTE]
+====
+The following command uses the `jq` JSON processor to get directly at the necessary data. Red Hat Enterprise Linux (RHEL), depending on version, might not provide a package for `jq`. You might need to install this from external sources, or unsupported locations.
+====
+
+----
+oc exec -c elasticsearch -n $LOGGING_NS $pod -- es_util --query='_mapping?pretty&filter_path=**.mappings.*.properties' \
+  | jq '.[].mappings[].properties | keys' \
+  | jq .[] \
+  | egrep -e "\."
+----
+
+The upgrade path depends on whether the indices xref:upgrading-efk-logging-stack-dots[have fields with dots] or xref:upgrading-efk-logging-stack-nodots[do not have fields with dots]. 
+
+[[upgrading-efk-logging-stack-dots]]
+==== Upgrading if fields have dots in field names
+
+If the xref:upgrading-efk-logging-stack-script[script above] indicates your indices contain fields with a dot in the name, use the following steps to correct this issue and upgrade. 
+
+To upgrade your EFK stack:
+
 . Review how to
 xref:../install_config/aggregate_logging.adoc#aggregate-logging-ansible-variables[specify logging Ansible variables]
 and update your Ansible inventory file to at least set the
@@ -756,19 +821,78 @@ You can set the `openshift_logging_elasticsearch_replace_configmap` parameter to
 ConfigMap can cause the upgrade to fail. The default is set to `false`. For more information, see the parameter 
 in xref:../install_config/aggregate_logging.adoc#aggregate-logging-ansible-variables[specify logging Ansible variables].
 
-. If your Fluentd `DeploymentConfig` and `DaemonSet` parameters for the EFK
-components set `imagePullPolicy` to `IfNotPresent`, set them to `Always`:
+. Dechedule your Fluentd pods to stop data ingestion and ensure the cluster state does not change.
++
+For example, you can change the node selector in Fluentd pods to one that does not match any nodes.
 +
 ----
-        image: <image_name>:<vX.Y>
-        imagePullPolicy: Always
+oc patch daemonset logging-fluentd -p '{"spec": {"template": {"spec": {"nodeSelector": {"non-existing": "true"}}}}}'
 ----
+
+. Perform an link: https://www.elastic.co/guide/en/elasticsearch/reference/2.4/indices-flush.html[Elasticsearch Index flush] on all relevant indices. The flush process persists all logs from memory to disk, which prevents log loss when Elasticsearch is shutdown during the upgrade.
+
+. Perform an online or offline backup:
+
+* Perform an link:https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-snapshots.html[online backup] of specific Elasticsearch indices the entire cluster.
+
+* Perform an offline backup: 
+
+.. Scale down all Elasticsearch DaemonSets to `0`:
++
+----
+$ oc scale dc <name> -n openshift-logging --replicas=0
+----
+
+.. Backup external persistent volumes using the appropriate method for your organization.
+
+. For any file name with a dot character, you need to take one of the following actions before upgrading:
++
+* Deleting the indices. This is the better approach to avoid mapping conflicts during the upgrade.
+* Reindexing the data and changing the documents to get rid of conflicting data structure. This method retain the data.   
+For information on potential mapping conflicts see link:https://www.elastic.co/guide/en/elasticsearch/reference/5.5/breaking_50_mapping_changes.html[Mappging changes] in the Elasticsearch documentation.
+
+. Repeat the on-line or offline backup.
 
 . Run the *_openshift-logging/config.yml_* playbook according to the
 xref:../install_config/aggregate_logging.adoc#deploying-the-efk-stack[deploying
 the EFK stack] instructions to complete the logging upgrade. You run the
 installation playbook for the new {product-title} version to upgrade the
 logging deployment.
+
+[[upgrading-efk-logging-stack-nodots]]
+==== Upgrading if fields do not have dots
+
+If the xref:upgrading-efk-logging-stack-script[script above] indicates your indices do not contain fields with a dot in the name, use the following steps to upgrade. 
+
+. Optionally, dechedule your Fluentd pods and scale down your Elasticsearch pods to stop data ingestion and ensure the cluster state does not change.
++
+For example, you can change the node selector in Fluentd pods to one that does not match any nodes.
++
+----
+oc patch daemonset logging-fluentd -p '{"spec": {"template": {"spec": {"nodeSelector": {"non-existing": "true"}}}}}'
+----
+
+. Optionally, perform and online or offline backup:
+
+* Perform an link:https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-snapshots.html[online backup] of specific Elasticsearch indices the entire cluster.
+
+* Perform an offline backup: 
+
+.. Scale down all Elasticsearch DeploymentConfigs to `0`:
++
+----
+$ oc scale dc <name> -n openshift-sdn --replicas=0
+----
+
+.. Backup external persistent volumes using the appropriate method for your organization.
+
+. Run the *_openshift-logging/config.yml_* playbook according to the
+xref:../install_config/aggregate_logging.adoc#deploying-the-efk-stack[deploying
+the EFK stack] instructions to complete the logging upgrade. You run the
+installation playbook for the new {product-title} version to upgrade the
+logging deployment.
+
+. Optionally, use the link:https://www.elastic.co/guide/en/elasticsearch/reference/5.5/modules-snapshots.html[Elasticsearch restore module] to restore your Elasticsearch indices from the snapshot.
 
 [[upgrading-cluster-metrics]]
 === Upgrading cluster metrics


upgrading/automated_upgrades.adoc
```diff
@@ -756,6 +756,25 @@ You can set the `openshift_logging_elasticsearch_replace_configmap` parameter to
 ConfigMap can cause the upgrade to fail. The default is set to `false`. For more information, see the parameter 
 in xref:../install_config/aggregate_logging.adoc#aggregate-logging-ansible-variables[specify logging Ansible variables].
 
+. Back up disks.
```
mburke5678: @jcantrill What needs to be backed up? Later versions of the ES docs talk about backing up your data using snapshots. Is that what's needed here?

upgrading/automated_upgrades.adoc
```diff
@@ -756,6 +756,25 @@ You can set the `openshift_logging_elasticsearch_replace_configmap` parameter to
...
+
+. Review the link:https://www.elastic.co/guide/en/elasticsearch/reference/5.5/breaking-changes-5.0.html[Elasticsearch breaking changes].
+The breaking changes contains important information you need to be aware of when migrating your application to Elasticsearch 5.0.
+
+. Run the following command to list all fields in the cluster that have a dot character in the name, such as *foo.bar*.
+For more context see https://bugzilla.redhat.com/show_bug.cgi?id=1666141#c5
++
```
mburke5678: @jcantrill We were recently asked by Eric Rich to not use JQ commands in the docs as RHEL 7 doesn't support JQ.
https://bugzilla.redhat.com/show_bug.cgi?id=1764726#c6

upgrading/automated_upgrades.adoc
```diff
@@ -756,6 +756,25 @@ You can set the `openshift_logging_elasticsearch_replace_configmap` parameter to
...
+----
+oc exec -c elasticsearch -n $LOGGING_NS $pod -- es_util --query='_mapping?pretty&filter_path=**.mappings.*.properties' \
+  | jq '.[].mappings[].properties | keys' \
+  | jq .[] \
+  | egrep -e "\."
+----
++
+If the script returns any file name with a dot character, you must....
```
mburke5678: @jcantrill What do the users need to do if they have files with dots in the name?

upgrading/automated_upgrades.adoc
jcantrill: @mburke5678  I would itemize this as 'Backup persistent volumes'.  Not sure if we have general PV procedure for doing thath.

@lukas-vlcek we should probably scale down ES prior to upgrade should we not?  This would allow proper backup of the persistentvolumes.  I don't think would advise backup via snapshots unless you know otherwise.

upgrading/automated_upgrades.adoc
jcantrill: We'll have to figure out the python equivalent

upgrading/automated_upgrades.adoc
jcantrill: @lukas-vlcek They need to?  delete those indices?

lukas-vlcek: @mburke5678 @jcantrill If they can delete them then this is the best approach to make sure they will not run into mapping conflicts when upgrading to ES 5.x. If, however, they need to keep this data then they need to investigate if these fields will cause mapping conflict (the script does not tell you that now).

upgrading/automated_upgrades.adoc
lukas-vlcek: What is recommended instead of `jq`? Python script?

upgrading/automated_upgrades.adoc
mburke5678: Do we want to recommend the ES Migration Helper plugin? Will this help with our issue?
https://github.com/elastic/elasticsearch-migration/tree/2.x

lukas-vlcek: @mburke5678 Probably not, this is ES plugin so it would have to packaged and supported by us. Moreover, this is a __site_ plugin which is an extra web UI interface and we would have to provide a way how customers could connect to this via web browser. Right now you can not access ES directly via browser (this would mean you have to add/configure proxy, setup user privs, ... probaby not worth it atm).

upgrading/automated_upgrades.adoc
jcantrill: > What is recommended instead of `jq`? Python script?

yes.  piping through a python script. not fun

upgrading/automated_upgrades.adoc
mburke5678: @lukas-vlcek The mapping conflict we are concerned with here is dots in the name? 
If so, the solution would be to change the dot to another character, perhaps an underline? 
`file.name` -> `file_name` as one possible solution?

lukas-vlcek: @mburke5678 It is not that easy.

Changing field name means you break all saved queries, charts and dashboards relying on the original field name (plus you also need to handle conflict if there are already both fields "field.name" and "field_name"; then what are you going to rename the "field.name" to?).
And you also change structure and semantics of indexed document. It is not just technical detail.

What is going on here is that there is a change between ES 2x and ES 5x in how dots in field names are interpreted. In ES 2x it was possible to have the following documents indexed into the same index:

```javascript
{
  "field": 123  // "field" is of type number
}
```
```javascript
// Any dot in field name is treaded as any other valid character in the field name.
// It is just part of the field name.
{
  "field.name": "Bob"  // "field.name" is of type String
}
```
In ES 5x things look a bit different:

```javascript
{
  "field": 123  // "field" is of type number
}
```
```javascript
// Any dot in field name is always interpreted as nested structure.
{
  "field" : {  // "field" is of type Object
    "name": "Bob"  // "name" is of type String
  }
}
```
And this is conflicting because field "field" can not be of type number and Object at the same time. Note that the conflict happens only if there is ask to store both the documents in the same index.

What our script does is that it will tell you about the "field.name" field, and that's it. This is a signal telling us that there is possibility to run into mapping conflict when upgrading to ES 5x.

What users then need to do is go and investigate if there is also ES 2x mapping for just the "field" field with different type than Object (like in our case there is "field" of type number). In such case this index can not be upgraded to ES 5x automatically.

lukas-vlcek: @mburke5678 In any way, we shall point users to Elastic docs about [upgrading indices with dots in field names](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/dots-in-names.html#_upgrading_fields_with_dots_to_5_x). This does however does not tell users what to do if they run into conflicting mapping. The truth is that if they can not or do not want delete such indices they need to reindex the data and change the documents to get rid of conflicting data structure. This is always specific.
What we could do it provide some simple example. WDYT?

upgrading/automated_upgrades.adoc
lukas-vlcek: @jcantrill First, I would point to this [Elasticsearch Snapshot](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-snapshots.html) documentation page where it highlights the following warning:

> You cannot back up an Elasticsearch cluster by simply taking a copy of the data directories of all of its nodes. Elasticsearch may be making changes to the contents of its data directories while it is running; copying its data directories cannot be expected to capture a consistent picture of their contents. If you try to restore a cluster from such a backup, it may fail and report corruption and/or missing files. Alternatively, it may appear to have succeeded though it silently lost some of its data. The only reliable way to back up a cluster is by using the snapshot and restore functionality.

If we want to do a backup of individual folder on data nodes then keep in mind that node data folders also contain complete cluster state. This means that if you scale down to the point where cluster becomes yellow (it is not possible to allocate replicas) before backup then this is also what user will get when restoring from this backup. Or let's put it this way, if user scales down to 3 nodes then he/she has to restore into three nodes as well. But still this sounds quite fragile to me.

IMO we shall support proper Snapshot and Restore API.

jcantrill: What we do in future is not going to be relevant to this documentation. We are strictly writing to address the 2.x to 5.x upgrade.  We addressed this issue when opened by using AWS backups.  I believe we need to advise users to bring the entire system down, snapshot, upgrade recognizing they may lose logs in during upgrade.

upgrading/automated_upgrades.adoc
mburke5678: @lukas-vlcek Does the script look only for `field.name` or any `field.*` fields?

mburke5678:
@lukas-vlcek I added your examples from above and more wording around the issue.

upgrading/automated_upgrades.adoc
lukas-vlcek: @jcantrill Ok, in that case I would suggest scaling fluentd pods down first, then calling [Index flush](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/indices-flush.html) on all relevant indices (\*) to make sure all logs are persisted from memory to disk and they do not get lost when ES is shutdown.
(\*) By relevant indices I mean primarily indices holding logs, but I think we can call it on all indices, including SG or Kibana, should there be any update in the flux...

upgrading/automated_upgrades.adoc
lukas-vlcek: @mburke5678 It looks for any field name containing the dot. No matter where the dot is or what is before or after the dot.

upgrading/automated_upgrades.adoc
jcantrill: @sferich888 Can you advise JQ alternative here?  I'm not found of figuring out python to do something that another tool already does well

upgrading/automated_upgrades.adoc
```diff
@@ -756,6 +798,27 @@ You can set the `openshift_logging_elasticsearch_replace_configmap` parameter to
...
+. Run the following script to determine if you have a `field.name` field in your indices.
++
+----
+oc exec -c elasticsearch -n $LOGGING_NS $pod -- es_util --query='_mapping?pretty&filter_path=**.mappings.*.properties' \
+  | jq '.[].mappings[].properties | keys' \
+  | jq .[] \
+  | egrep -e "\."
+----
++
+If the script returns any file name with a dot character, you need to take action before upgrading:
```
jcantrill: `...any key name...`

upgrading/automated_upgrades.adoc
mburke5678: @lukas-vlcek Added a step between scaling down and snapshotting:

_Perform an Elasticsearch Index flush on all relevant indices. The flush process persists all logs from memory to disk, which prevents log loss when Elasticsearch is shutdown during the upgrade._

mburke5678: The user should scale down the fluentd pods? @jcantrill suggests the user `scale down ES`.

upgrading/automated_upgrades.adoc
mburke5678: @lukas-vlcek Changed to:
_To help identify potential conflicts, OpenShift provides a script that examines your Elasticsearch fields to determine if any fields contain a dot in the name._

upgrading/automated_upgrades.adoc
jcantrill: scaling down fluent will stop data ingestion but scaling down ES will also ensure the cluster state is not changing so you could say snapshot the AWS volumes

upgrading/automated_upgrades.adoc
sferich888: @jcantrill I agree, the request here is mainly for of command output parsing. Because your using exec to pull things out of ES the context of the original request doesn't quite apply/work here. 

Thus JQ is probably the best tool, for this (I just wish it was a dependency we could resolve). We just need to clearly call out that this command uses JQ, as an example to get directly at the necessary data and it's dependency can't be meet by RHEL. Please add a note or call out to this effect, and we can consider this a exception to the BZ request. 

I would also request that we not pipe to JQ multiple times and then pipe to grep/egrep. This overcomplicates things. 

Something like the following is better: 

`| jq ''.[].mappings[].properties | keys[] | select(., contains("\."))'` @jcantrill can we see if a command improvement in this area is possible? I think I got the syntax her correct but I'm not 100% sure, this my example may need some tweaking.

mburke5678: @sferich888 Not sure what you mean by _dependency can't be meet by RHEL_.

Note:
The following command uses the *jq* JSON processor to get directly at the necessary data. RHEL cannot meet *jq*'s dependencies.

mburke5678:
@anpingli Please take a look. Thank you.

upgrading/automated_upgrades.adoc
sferich888: Red Hat doesn't provide a `jq` package, for users to use. Thus, users are forced, if we use `jq` in our docs to find the command, install it and then use our docs. 

This is why were trying to phase out the use of the tool in our docs.

mburke5678: @sferich888 Is this strong enough? 
```
"The following command uses the *jq* JSON processor to get directly at the necessary data. {product-title} does not provide a *jq* package. 
To use this command, you will need to download and install *jq*."
```

sferich888: Product title is RHEL, but yes this works. 
You might consider: 

{product-title}, depending on version, may not provide a package for `jq`. You may need to install this from external sources, or unsupported locations.

mburke5678:
@anpingli Please take a look. Thank you.

upgrading/automated_upgrades.adoc
```diff
@@ -756,6 +797,34 @@ You can set the `openshift_logging_elasticsearch_replace_configmap` parameter to
 ConfigMap can cause the upgrade to fail. The default is set to `false`. For more information, see the parameter 
 in xref:../install_config/aggregate_logging.adoc#aggregate-logging-ansible-variables[specify logging Ansible variables].
 
+. Scale down your Fluentd and Elasticsearch pods to stop data ingestion and ensure the cluster state does not change.
+
+. Perform an link:https://www.elastic.co/guide/en/elasticsearch/reference/2.4/indices-flush.html[Elasticsearch Index flush] on all relevant indices. The flush process persists all logs from memory to disk, which prevents log loss when Elasticsearch is shutdown during the upgrade.
+
+. Use the link:https://www.elastic.co/guide/en/elasticsearch/reference/5.5/modules-snapshots.html[Elasticsearch snapshot module] to create a snapshot of your AWS volumes.
```
anpingli: @mburke5678  'snapshot of your AWS volumes'? Is this document for AWS platform only?

upgrading/automated_upgrades.adoc
```diff
@@ -756,6 +797,34 @@ You can set the `openshift_logging_elasticsearch_replace_configmap` parameter to
...
+
+. Run the following script to determine if your indices contain any fields with a dot in the name.
```
anpingli: Which scripts?  Could you tell me how to open the whole document? I  can only view the diff part?

upgrading/automated_upgrades.adoc
```diff
@@ -756,6 +797,34 @@ You can set the `openshift_logging_elasticsearch_replace_configmap` parameter to
 ConfigMap can cause the upgrade to fail. The default is set to `false`. For more information, see the parameter 
 in xref:../install_config/aggregate_logging.adoc#aggregate-logging-ansible-variables[specify logging Ansible variables].
 
+. Scale down your Fluentd and Elasticsearch pods to stop data ingestion and ensure the cluster state does not change.
```
anpingli: No replicas for damonset controller pods. Is 'scale down' the proper word ?  shall we use deschuler the damonset pods by either modify the nodeSelector in ds/logging-fluentd or update the node label.

anpingli: Are we expected the user to these steps prior openshift-logging playbooks?

upgrading/automated_upgrades.adoc
mburke5678: @anpingli This was from @jcantrill comment:
_scaling down fluent will stop data ingestion but scaling down ES will also ensure the cluster state is not changing so you could say snapshot the AWS volumes_
https://github.com/openshift/openshift-docs/pull/17931#discussion_r348196409

What should the user snapshot? Elasticsearch volumes?

upgrading/automated_upgrades.adoc
mburke5678: @anpingli For full preview, see:
http://file.rdu.redhat.com/~mburke/logging-311-upgrade/upgrading/automated_upgrades.html#upgrading-efk-logging-stack

upgrading/automated_upgrades.adoc
mburke5678: @jcantrill @lukas-vlcek Please comment on @anpingli question on scale down. 

Anping, the scale down/descheduling is done before running playbooks.

mburke5678: Anping, the scale down/descheduling is done before running playbooks.

upgrading/automated_upgrades.adoc
anpingli: @mburke5678 Thanks, I found the script at http://file.rdu.redhat.com/~mburke/logging-311-upgrade/upgrading/automated_upgrades.html#upgrading-efk-logging-stack

upgrading/automated_upgrades.adoc
anpingli: @mburke5678  We need to split this sentence to two steps. 1) de-scheduler fluentd 2) scale down elasticsearch.   and the step 'scale down elasticsearch' should be after step 7.
@jcantrill  @lukas-vlcek I think we need to explain the 'scale down Elasticsearch' .  I think that means 'set dc/elasticsearchs replicas number=0' rather than elasicsearch cluster 'scale down'

anpingli:
Add conditon or drop step 10.
10. Use the Elasticsearch restore module to restore your Elasticsearch indices from the snapshot.

upgrading/automated_upgrades.adoc
jcantrill: It is not possible to 'scale down' a daemonset.  Essentially you have to modify the node selector such that it won't mache any nodes.

@anpingli correct, we will need to set each ES DS replica to 0

lukas-vlcek (commented):
Looking at the http://file.rdu.redhat.com/~mburke/logging-311-upgrade/upgrading/automated_upgrades.html#upgrading-efk-logging-stack I have the following comments:

- The term "scale down ES pods" is confusing to me. Does it really mean stopping all the nodes? Or doing the real scale down (i.e. decreasing number of running ES nodes?). I think we really mean stopping them and not scaling them down.

- When we ask users to perform flush and snapshot we can not tell them to turn off all the nodes before. Both flush and snapshot requires ES nodes up and running.

- Step 7 should be performed only if any data (i.e. index) was changed in step 6. Isn't it?

upgrading/automated_upgrades.adoc
lukas-vlcek: @mburke5678 @jcantrill There are two different snapshots:

First, Elasticsearch has its own [Snapshot&Restore API](https://www.elastic.co/guide/en/elasticsearch/reference/5.5/modules-snapshots.html) and you tell it which indices you want to snapshot to repository (mounted storage). ES has to be running.

Second, there is [AWS snapshot](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSSnapshots.html) and this is completely orthogonal to ES. However, to make sure the snapshot is consistent ES must be stopped. You need to snapshot all the data folders (it holds both the index data and cluster metadata).

I think Jeff was talking about the second scenario, in which case we should remove link to ES Snapshot&Restore API from the doc.

mburke5678:
@anpingli @lukas-vlcek 
Should the workflow be:
Flush
Snapshot
Fix fields with a dot
Optionally snapshot again if any changes
Deschedule Fluentd (how is this done? RemoveDuplicates?)
Scale dc elasticsearch  --replicas=0 -or- shutdown ES (how is this done?)
Run playbook

Preview: http://file.rdu.redhat.com/~mburke/logging-311-upgrade/upgrading/automated_upgrades.html#upgrading-efk-logging-stack

upgrading/automated_upgrades.adoc
mburke5678: @jcantrill ^^

lukas-vlcek:
I do not think we are talking about ES native Snapshot. Which means we do not perform snapshot before all ES nodes are turned off. So this snapshot will be called only after `scale dc elasticsearch --replicas=0`.
@jcantrill Do you have different idea?

upgrading/automated_upgrades.adoc
jcantrill: We have no experience in Elasticsearch snapshots and are advising user's to back up the actual disk in any way that supports their platform, AWS snapshots is one of them.  I would expect there to be interruption in logging where user's bring down their entire ES cluster, backup disk, migrate

mburke5678: "To protect your  Elasticsearch data, you should back up your data using your standard policies" Something like that?

sferich888: What is QE testing, as part of our backup plan in this area? 

@xltian ?

anpingli: @sferich888  That is an offline backup.  Logging QE didn't test the backup/restore in our testing.

anpingli:
@lukas-vlcek  Shall we only backup the dot indices?  https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-snapshots.html

anpingli:
@mburke5678, I think that is exception following.
if have dot indices.
      1. Deschedule Fluentd 
      2. Flush
      3. [A] Online backup ( backup individual index or  the entire cluster https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-snapshots.html)
      3. [B] Offline backup ( scale all dc elasticsearch --replicas=0;  backup external persistent volume; scale up all dc elasticsearch --replicas=1 )
       4. Fix fields with a dot
       5. [Optional ] backup storage again (Online or offline)
  else:
          [Optional ] backup storage

    Run playbook to upgraded logging
    [Optional]  If failed:  restore snapshot  or fix the error.

mburke5678:
@anpingli @lukas-vlcek @jcantrill After the upgrade do the back ups needs to be restored or are the back ups only to have a copy of the data in case of a problem after upgrade?

mburke5678:
@anpingli I performed a modified version of your steps to update (actually reinstalling the same version of cluster logging).
1. Descheduler Fluentd (ES node went into error)
2. Scaled ES daemonset to 0
3. Scaled ES daemonset to 1
4. Run playbook>
The playbook got stuck on 
```
RUNNING HANDLER [openshift_logging_elasticsearch : oc_obj] *********************************************************************************************************
Tuesday 07 January 2020  15:31:12 -0500 (0:00:00.484)       0:02:33.971 ******* 
FAILED - RETRYING: Waiting for logging-es-data-master-rk5fs6d4 to finish scaling up (60 retries left).
``` 
The playbook finished finished but the ES pod is still in ERROR.

mburke5678:
@anpingli @lukas-vlcek I updated the docs to better match [Anping's suggested workflow](https://github.com/openshift/openshift-docs/pull/17931#issuecomment-571451297) in the latest commit.

Preview: http://file.rdu.redhat.com/~mburke/logging-311-upgrade/upgrading/automated_upgrades.html#upgrading-efk-logging-stack

anpingli (commented):
It is optional. the playbooks can be running when DeploymentConfigs to 0.
5.c  Scale up all Elasticsearch DeploymentConfigs to 1:

It is not necessary, the playbook will enabled the nodeSelector.
Post 8: oc patch daemonset logging-fluentd -p '{"spec": {"template": {"spec": {"nodeSelector": {"logging-infra-fluentd": "true"}}}}}'

mburke5678:
@anpingli I removed those two steps.

mburke5678:
@anpingli What do we need to complete this PR?

anpingli:
I sugget to remove 
5.c Scale up all Elasticsearch DeploymentConfigs to 1
8: oc patch daemonset logging-fluentd -p '{"spec": {"template": {"spec": {"nodeSelector": {"logging-infra-fluentd": "true"}}}}}'

mburke5678:
@anpingli I removed those two steps last week when you asked. Is this what you expected?

https://github.com/openshift/openshift-docs/pull/17931/commits/692a25404dab8ccb0a7e5ffaf489c2c17917c35a

anpingli:
LGTM now.

upgrading/automated_upgrades.adoc
```diff
@@ -735,6 +735,71 @@ upgrade the component.
 To upgrade an existing EFK logging stack deployment, you review your parameters
 and run the *_openshift-logging/config.yml_* playbook.
 
+The EFK upgrade also upgrades Elasticsearch from the 2.x version to the 5.x version. For important information on changes in Elasticsearch 5.x, you should review the link:https://www.elastic.co/guide/en/elasticsearch/reference/5.5/breaking-changes-5.0.html[Elasticsearch breaking changes].
```
sheriff-rh: I recommend changing from `[version].x` syntax in our documentation, even though I see that ES uses it in theirs - maybe try `version 5 and higher` or on line 758 `In Elasticsearch 5 the field...`

upgrading/automated_upgrades.adoc
```diff
@@ -735,6 +735,71 @@ upgrade the component.
...
+{
+  "field" : {  // "field" is of type Object
+    "name": "Bob"  // "name" is of type String
+  }
+}
+----
+
+Upgrading in this case would result in the `field` field having two different types, which is not permitted. 
+
+If you need to keep these conflicting indices, you need to reindex the data and change the documents to get rid of conflicting data structure. For more information, see link:https://www.elastic.co/guide/en/elasticsearch/reference/2.4/dots-in-names.html#_upgrading_fields_with_dots_to_5_x[Upgrading fields with dots to 5.x].
```
sheriff-rh: It's okay to use `5.x` here because you're referencing a link that is titled `5.x`. Hope this makes sense!

sheriff-rh (commented):
Just one nit about using `[version].x` in our own words and documentation, otherwise looks good!

PR #17981, opened 2019-11-07
Fix hostpath index and merge conflict

bobfuru:
Carries over work from [17603](https://github.com/openshift/openshift-docs/pull/17603), which had a merge conflict from CP to 4.1.


bobfuru:
https://docs.openshift.com/container-platform/4.1/storage/persistent-storage/persistent-storage-hostpath.html

PR #18020, opened 2019-11-08
Remove snapshots in 4.3

bobfuru:
@liangxia PTAL 
We added Snapshots to 4.1 & 4.2 docs, with a Deprecated note in 4.2. This PR removes Snapshots in master, CP to 4.3 only. SME (Jan) has approved.


liangxia:
LGTM

PR #18042, opened 2019-11-11
added supports networkpolicy

neal-timpe:
Please cherry pick this PR to the 4.2 and 4.3 branch.

https://bugzilla.redhat.com/show_bug.cgi?id=1770538
https://issues.jboss.org/projects/OSSMDOC/issues/OSSMDOC-13?filter=allissues

The BZ says that the current 4.2 docs say, "Red Hat OpenShift Service Mesh is only suited for OpenShift Container Platform Software Defined Networking (SDN) configured as a flat network with no external providers." And that's no longer true.

I've changed that to say Red Hat OpenShift Service Mesh supports NetworkPolicy.

### DIFF
--- a/modules/ossm-supported-configurations.adoc
+++ b/modules/ossm-supported-configurations.adoc
@@ -17,7 +17,6 @@ OpenShift Online and OpenShift Dedicated are not supported for {ProductName} {Pr
 
 * The deployment must be contained to a single {product-title} cluster that is not federated.
 * This release of {ProductName} is only available on {product-title} x86_64.
-* {ProductName} is only suited for {product-title} Software Defined Networking (SDN) configured as a flat network with no external providers.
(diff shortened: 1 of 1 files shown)


dgn (approved):

rcernich:
Hey @neal-timpe, it actually supports flat network, network policy, and multitenant networking configurations.

jboxman:
And OpenShift SDN supports all those modes, so maybe there's no need for a sentence describing supported network isolation modes at all? (As there's no isolation mode that doesn't work with this.)

neal-timpe:
@rcernich @dgn Do you agree with @jboxman? I think removing this sentence might be a good solution too.

rcernich:
Correct.

dgn:
Makes sense

neal-timpe:
@openshift/team-documentation Please take a look an cherry pick this PR to the 4.2 and 4.3 branch.

adellape (approved):
:+1:

Issue #18041, opened 2019-11-11
[enterprise-3.11] Edit suggested in file release_notes/ocp_3_11_release_notes.adoc


zhigangwang315:
### Which section(s) is the issue in?

https://docs.openshift.com/container-platform/3.11/release_notes/ocp_3_11_release_notes.html#ocp-3-11-153-images

### What needs fixing?
The path to container images is wrong

The path in this doc is correct
https://docs.openshift.com/container-platform/3.11/install/disconnected_install.html#disconnected-syncing-images

codyhoag:
@zhigangwang315 just to confirm, are you suggesting the registry URL defined as `registry.access.redhat.com` be updated to `registry.redhat.io`?

zhigangwang315:
A customer tried to download the container images listed at the end of the 3.11 release notes:

https://docs.openshift.com/container-platform/3.11/release_notes/ocp_3_11_release_notes.html
The first container image on the list does not exist:


sudo docker pull registry.access.redhat.com/openshift3/aos3-installation-container-v3.11.153-3<http://registry.access.redhat.com/openshift3/aos3-installation-container-v3.11.153-3>

Using default tag: latest

Error response from daemon: repository registry.access.redhat.com/openshift3/aos3-installation-container-v3.11.153-3<http://registry.access.redhat.com/openshift3/aos3-installation-container-v3.11.153-3> not found


When we point him to the doc below to download image using version tag, it works.

https://docs.openshift.com/container-platform/3.11/install/disconnected_install.html#disconnected-syncing-images

So someone need to verify the path in the release notes.


Thanks.

Zhigang


________________________________

codyhoag:
@zhigangwang315 I have requested the image names for 3.11.153 from the release team. I will update the doc once I've received them.

codyhoag:
@zhigangwang315 I've updated the release notes with the appropriate image names in #18074. This update should be reflected in the published docs soon. Thanks for reporting!

PR #18074, opened 2019-11-12
Red Hat image name updates in release notes

codyhoag:
@ahardin-rh @sheriff-rh can either of you confirm this change? This was reported in #18041

This was documented as a change in the 3.11 release notes ([here](https://docs.openshift.com/container-platform/3.11/release_notes/ocp_3_11_release_notes.html#ocp-311-accessing-and-configuring-red-hat-registry)). Let me know if this was intentional, since the change was made in the same release. Thanks!


sheriff-rh:
I'm unclear what the issue is. Typically, we put the image list in the bug advisories in 3.x releases, like [this](https://access.redhat.com/errata/RHBA-2019:2824). For this particular release, we put the image list in the release notes.  It's just a list of images available for 3.11.153, it does not conflict with the disconnected installation docs referenced in the issue. 

I suppose we could make it less confusing by moving the image list to a KCS like we do with [4.x](https://access.redhat.com/solutions/4562241)? Then we change the [release notes](https://docs.openshift.com/container-platform/4.1/release_notes/ocp-4-1-release-notes.html#ocp-4-1-23) to point to the KCS.

codyhoag:
@sheriff-rh from what I gathered on first pass, the registry URL changed from `registry.access.redhat.com` &rarr; `registry.redhat.io` in 3.11, and the request is to update the registry URL for accessing those images as such. But perhaps this goes deeper than I initially realized.

codyhoag:
I'll confirm with reporter

codyhoag:
The reporter has informed me that the first few images in the release notes do not exist. I have confirmed by attempting a `docker pull` that they aren't reachable. I'll investigate further.

codyhoag:
Confirmed **Build Name** &rarr; **Image Name** change with RCM team. Acquired image names from Thiago.

@sheriff-rh can you review and, if all looks good, merge too? Thanks!

sheriff-rh:
Looks good! Merging.

PR #18078, opened 2019-11-12
PROD-2140: Describe autofill clusterID in a support case

bobfuru:
@sferich888 PTAL - this PR relates to [OSDOCS-764](https://jira.coreos.com/browse/OSDOCS-764), which is to document changes introduced in [PROD-2140](https://jira.coreos.com/browse/PROD-2140).

I used terminology *Open Support Case* or *Report Bug* in the action to describe both prerelease and GA versions.

### DIFF
--- a/modules/support-get-cluster-id.adoc
+++ b/modules/support-get-cluster-id.adoc
@@ -5,21 +5,31 @@
 [id="support-get-cluster-id_{context}"]
 = Obtaining your cluster ID
 
-When providing information to Red Hat Support, it is helpful to provide the unique identifier for your cluster. You can obtain your cluster ID by using the OpenShift CLI (`oc`) or by using the {product-title} web console.
+When providing information to Red Hat Support, it is helpful to provide the unique identifier for your cluster. You can have your cluster ID autofilled by using the {product-title} web console. You can also manually obtain your cluster ID by using the web console or the OpenShift CLI (`oc`).
 
 .Prerequisites
 
 * Access to the cluster as a user with the `cluster-admin` role.
-* The {product-title} CLI (`oc`) installed.
+* Access to the web console or the OpenShift CLI (`oc`) installed.
 
 .Procedure
+ifdef::openshift-enterprise,openshift-dedicated[]
+* To open a support case and have your cluster ID autofilled using the web console:
+.. From the toolbar, navigate to *(?) Help* -> *Open Support Case*.
+.. The 'Cluster ID' value is autofilled.
+endif::[]
+ifdef::openshift-origin[]
+* To open a bug and have your cluster ID autofilled using the web console:
+.. From the toolbar, navigate to *(?) Help* -> *Report Bug*.
+.. The 'Cluster ID' value is autofilled after you click `Submit Bug`.
+endif::[]
+
+* To manually obtain your cluster ID using the web console:
+.. Navigate to *Home* -> *Dashboards* -> *Overview*.
+.. The value is available in the *Cluster ID* field of the *Details* section.
 
 * To obtain your cluster ID using the OpenShift CLI (`oc`), run the following command:
 +
 ----
 $ oc get clusterversion -o jsonpath='{.items[].spec.clusterID}{"\n"}'
 ----
-
-* To obtain your cluster ID using the {product-title} web console:
-.. Navigate to *Home* -> *Dashboards* -> *Overview*.
-.. The value is available in the *Cluster ID* field of the *Details* section.
--- a/modules/support.adoc
+++ b/modules/support.adoc
@@ -20,7 +20,7 @@ ifdef::openshift-enterprise,openshift-dedicated[]
 When submitting a support case, it is recommended to provide the following information about your cluster to Red Hat Support to aid in troubleshooting:
 
 * Data gathered using the `oc adm must-gather` command
-* The unique cluster ID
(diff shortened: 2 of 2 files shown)


modules/support-get-cluster-id.adoc
```diff
@@ -5,21 +5,25 @@
...
 .Prerequisites
 
 * Access to the cluster as a user with the `cluster-admin` role.
-* The {product-title} CLI (`oc`) installed.
+* Access to the web console or the OpenShift CLI (`oc`) installed.
 
 .Procedure
 
+* To open a support case and have your cluster ID autofilled using the web console:
+.. From the toolbar, navigate to *(?) Help* -> *Open Support Case* or *Report Bug*.
```
sferich888: You should only see *Open Support Case*. 
You should only see, *Report Bug* if this is OKD.

modules/support.adoc
```diff
@@ -20,7 +20,7 @@ ifdef::openshift-enterprise,openshift-dedicated[]
 When submitting a support case, it is recommended to provide the following information about your cluster to Red Hat Support to aid in troubleshooting:
 
 * Data gathered using the `oc adm must-gather` command
-* The unique cluster ID
+* The unique cluster ID. (This value is autofilled when you navigate to *(?) Help* -> *Open Support Case* or *Report Bug*.)
```
sferich888: This 'additonal text' should probably be in a call out or not.

bobfuru: I agree, though lines 20-23 are already in an `important note` callout. Also, this note is already conditionalized for Enterprise and Dedicated, so I removed the `Report Bug` ref.

modules/support-get-cluster-id.adoc
bobfuru: Added conditionals

bobfuru:
@sferich888 Can I please get an ack that you are good with changes?

bobfuru:
@yapei PTAL, thanks!

modules/support-get-cluster-id.adoc
```diff
@@ -5,21 +5,30 @@
...
 
 .Prerequisites
 
 * Access to the cluster as a user with the `cluster-admin` role.
-* The {product-title} CLI (`oc`) installed.
+* Access to the web console or the OpenShift CLI (`oc`) installed.
 
 .Procedure
 
+* To open a support case and have your cluster ID autofilled using the web console:
```
sferich888: Support Case here need to be conditionalized.

bobfuru: Ah, good point. I updated conditionals so "support case" shows for Enterprise and "bug" shows for Origin.

modules/support-get-cluster-id.adoc
```diff
@@ -5,21 +5,31 @@
...
+* Access to the web console or the OpenShift CLI (`oc`) installed.
 
 .Procedure
+ifdef::openshift-enterprise,openshift-dedicated[]
+* To open a support case and have your cluster ID autofilled using the web console:
+.. From the toolbar, navigate to *(?) Help* -> *Open Support Case*.
+.. The 'Cluster ID' value is autofilled.
+endif::[]
+ifdef::openshift-origin[]
+* To open a bug and have your cluster ID autofilled using the web console:
```
bobfuru: @sferich888 Should I remove "and have your cluster ID autofilled"? I just tested 4.3 prerelease, and while it opens a BZ, I don't see where it autofills a clusterID.

modules/support-get-cluster-id.adoc
```diff
@@ -5,21 +5,31 @@
...
+.. From the toolbar, navigate to *(?) Help* -> *Report Bug*.
+.. The 'Cluster ID' value is autofilled.
```
bobfuru: ^^ Same as comment above. Not sure clusterID is autofilled in BZ.

sferich888: It's put into the bugzilla environment, field. What Bz did you test with?

modules/support-get-cluster-id.adoc
sferich888: No it auto populates a field that's not typically shown. It's an advanced field if I'm not mistakin

bobfuru: I see it now. It isn't showing in the BZ create bug stage. Once submitted, it does autofill (along with version) in the `Environment` field.

modules/support-get-cluster-id.adoc
bobfuru: Got it. I'm going to update to say that it is autofilled after you click `Submit bug`. Thanks!

bobfuru:
@sferich888 I am using a conditional that shows "Support Case" for `openshift-enterprise` and `openshift-dedicated`. I am using "Report Bug" for `openshift-origin` branch. 

Can you please confirm this is correct in terms of downstream distro support level?

sferich888:
Open support case should also work with online, but I think you have it right.

yapei:
LGTM from QE's point of view

bobfuru:
> Open support case should also work with online, but I think you have it right.

Thanks, Eric. Online doesn't have anything in the "[Support](https://docs.openshift.com/online/starter/support/getting-support.html)" section of docs where it would make sense to include this info, so I'll leave it this way.

bobfuru:
@openshift/team-documentation PTAL. This is for 4.3 and has been approved by SME and QE.

ahardin-rh:
Looks good!

PR #18079, opened 2019-11-12
OSDOCS-770 AWS existing VPC

kalexand-rh:
https://jira.coreos.com/browse/OSDOCS-770

I'm checking with the networking team to see if they have opinions about positioning this against the existing custom networking docs before I fold it into an assembly.

@abhinavdahiya, will you PTAL?

### DIFF
--- a/installing/installing_aws/installing-aws-vpc.adoc
+++ b/installing/installing_aws/installing-aws-vpc.adoc
@@ -1,0 +1,60 @@
+[id="installing-aws-vpc"]
+= Installing a cluster on AWS to an existing VPC
+include::modules/common-attributes.adoc[]
+:context: installing-aws-vpc
+
+toc::[]
+
+In {product-title} version {product-version}, you can install a cluster into an existing Amazon Virtual Private Cloud (VPC) on Amazon Web Services (AWS). The installation program provisions the rest of the required infrastructure, which you can further customize. To customize the installation, you modify
+parameters in the `install-config.yaml` file before you install the cluster.
+
+.Prerequisites
+
+* Review details about the
+xref:../../architecture/architecture-installation.adoc#architecture-installation[{product-title} installation and update]
+processes.
+* xref:../../installing/installing_aws/installing-aws-account.adoc#installing-aws-account[Configure an AWS account]
+to host the cluster.
++
+[IMPORTANT]
+====
+If you have an AWS profile stored on your computer, it must not use a temporary
+session token that you generated while using a multi-factor authentication
+device. The cluster continues to use your current AWS credentials to create
+AWS resources for the entire life of the cluster, so you must use long-lived
+credentials. To generate appropriate keys, see
+link:https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_access-keys.html[Managing Access Keys for IAM Users]
+in the AWS documentation. You can supply the keys when you run the installation
+program.
+====
+* If you use a firewall, you must
+xref:../../installing/install_config/configuring-firewall.adoc#configuring-firewall[configure it to allow the sites] that your cluster requires access to.
+
+include::modules/installation-custom-aws-vpc.adoc[leveloffset=+1]
+
+include::modules/cluster-entitlements.adoc[leveloffset=+1]
+
+include::modules/ssh-agent-using.adoc[leveloffset=+1]
+
+include::modules/installation-obtaining-installer.adoc[leveloffset=+1]
+
+include::modules/installation-initializing.adoc[leveloffset=+1]
+
+include::modules/installation-configuration-parameters.adoc[leveloffset=+2]
+
+include::modules/installation-aws-config-yaml.adoc[leveloffset=+2]
+
+// Removing; Proxy not supported for AWS IPI for 4.2
+// include::modules/installation-configure-proxy.adoc[leveloffset=+2]
+
+include::modules/installation-launching-installer.adoc[leveloffset=+1]
+
+include::modules/cli-installing-cli.adoc[leveloffset=+1]
+
+include::modules/cli-logging-in-kubeadmin.adoc[leveloffset=+1]
+
+.Next steps
+
+* xref:../../installing/install_config/customizations.adoc#customizations[Customize your cluster].
+* If necessary, you can
+xref:../../support/remote_health_monitoring/opting-out-of-remote-health-reporting.adoc#opting-out-remote-health-reporting_opting-out-remote-health-reporting[opt out of remote health reporting].
--- a/modules/installation-configuration-parameters.adoc
+++ b/modules/installation-configuration-parameters.adoc
@@ -2,6 +2,7 @@
 //
 // * installing/installing_aws/installing-aws-customizations.adoc
 // * installing/installing_aws/installing-aws-network-customizations.adoc
+// * installing/installing_aws/installing-aws-vpc.adoc
 // * installing/installing_azure/installing-azure-customizations.adoc
 // * installing/installing_azure/installing-azure-network-customizations.adoc
 // * installing/installing_azure/installing-azure-vnet.adoc
@@ -14,6 +15,9 @@ ifeval::["{context}" == "installing-aws-customizations"]
 :aws:
 endif::[]
 ifeval::["{context}" == "installing-aws-network-customizations"]
+:aws:
+endif::[]
+ifeval::["{context}" == "installing-aws-vpc"]
 :aws:
 endif::[]
 ifeval::["{context}" == "installing-azure-customizations"]
@@ -191,8 +195,7 @@ such as `io1`.
 such as `c5.9xlarge`.
 
 |`compute.platform.aws.zones`
-|The availability zones where the installation program creates machines for the
-compute MachinePool.
+|The availability zones where the installation program creates machines for the compute MachinePool. If you provide your own VPC, you must provide a subnet in that availability zone.
 |A list of valid AWS availability zones, such as `us-east-1c`, in a
 link:https://yaml.org/spec/1.2/spec.html#sequence//[YAML sequence].
 
@@ -224,6 +227,11 @@ resources that it creates.
 For more information about AWS tags,
 see link:https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html[Tagging Your Amazon EC2 Resources]
 in the AWS documentation.
+
+|`platform.aws.subnets`
+|If you provide the VPC instead of allowing the installation program to create the VPC for you, specify the subnets for the cluster to use. The subnets must be part of the same `machineCIDR` range that you specify. For a standard cluster, specify a public and a private subnet for each availability zone. For a private cluster, specify a private subnet for each availability zone.
+|Valid subnet range.
+
 |====
 endif::aws[]
 
@@ -348,6 +356,9 @@ ifeval::["{context}" == "installing-aws-customizations"]
 :!aws:
 endif::[]
 ifeval::["{context}" == "installing-aws-network-customizations"]
+:!aws:
+endif::[]
+ifeval::["{context}" == "installing-aws-vpc"]
 :!aws:
 endif::[]
 ifeval::["{context}" == "installing-azure-customizations"]
--- a/modules/installation-custom-aws-vpc.adoc
+++ b/modules/installation-custom-aws-vpc.adoc
@@ -1,0 +1,136 @@
+// Module included in the following assemblies:
+//
+// * installing/installing_aws/installing-aws-vpc.adoc
+
+[id="installation-custom-aws-vpc_{context}"]
+= About using a custom VPC
+
+In {product-title} {product-version}, you can deploy a cluster into existing subnets in an existing Amazon Virtual Private Cloud (VPC) in Amazon Web Services (AWS). By deploying {product-title} into an existing AWS VPC, you might be able to avoid limit constraints in new accounts or more easily abide by the operational constraints that your company's guidelines set. If you cannot obtain the infrastructure creation permissions that are required to create the VPC yourself, use this installation option.
+
+Because the installation program cannot know what other components are also in your existing subnets, it cannot choose subnet CIDRs and so forth on your behalf. You must configure networking for the subnets that you install your cluster to yourself.
+
+[id="installation-custom-aws-vpc-requirements_{context}"]
+== Requirements for using your VPC
+
+The installation program no longer creates the following components:
+
+* Internet gateways
+* NAT gateways
+* Subnets
+* Route tables
+* VPCs
+* VPC DHCP options
+* VPC endpoints
+
+If you use a custom VPC, you must correctly configure it and its subnets for the installation program and the cluster to use. The installation program cannot subdivide network ranges for the cluster to use, set route tables for the subnets, or set VPC options like DHCP, so you must do so before you install the cluster.
+
+Your VPC must meet the following characteristics:
+
+* The VPC's CIDR block must contain the `Networking.MachineCIDR` range, which is the IP address pool for cluster machines.
+* The VPC must not use the `kubernetes.io/cluster/.*: owned` tag.
+* You must enable the `enableDnsSupport` and `enableDnsHostnames` attributes in your VPC so that the cluster can use the Route53 zones that are attached to the VPC to resolve cluster’s internal DNS records. See link:https://docs.aws.amazon.com/vpc/latest/userguide/vpc-dns.html#vpc-dns-support[DNS Support in Your VPC] in the AWS documentation.
+
+If you use a cluster with public access, you must create a public and a private subnet for each availability zone that your cluster uses. The installation program modifies your subnets to add the `kubernetes.io/cluster/.*: shared` tag, so your subnets must have at least one free tag slot available for it. Review the current link:https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html#tag-restrictions[Tag Restrictions] in the AWS documentation to ensure that the installation program can add a tag to each subnet that you specify.
+
+.Required VPC components
+
+You must provide a suitable VPC and subnets that allow communication to your
+machines.
+
+[cols="2a,7a,3a,3a",options="header"]
+|===
+
+|Component
+|AWS type
+2+|Description
+
+|VPC
+|* `AWS::EC2::VPC`
+* `AWS::EC2::VPCEndpoint`
+2+|You must provide a public VPC for the cluster to use. The VPC uses an
+endpoint that references the route tables for each subnet to improve communication with the registry that is hosted in S3.
+
+|Public subnets
+|* `AWS::EC2::Subnet`
+* `AWS::EC2::SubnetNetworkAclAssociation`
+2+|Your VPC must have public subnets for between 1 and 3 availability zones
+and associate them with appropriate Ingress rules.
+
+|Internet gateway
+|
+* `AWS::EC2::InternetGateway`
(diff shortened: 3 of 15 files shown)


modules/installation-about-custom-aws-vpcs.adoc
```diff
@@ -0,0 +1,64 @@
+// Module included in the following assemblies:
+//
+// * installing/
+
+[id="installation-about-custom-aws-vpcs_{context}"]
+= About using a custom VPC
+
+In {product-title} {product-version}, you can deploy a cluster into an existing VPC in Amazon Web Services (AWS). If you do, you must also use existing subnets within the VPC and routing rules.
```
abhinavdahiya: > If you do, you must also use existing subnets within the VPC and routing rules.

hmm, might need to be a little more focused :thinking:

modules/installation-about-custom-aws-vpcs.adoc
```diff
@@ -0,0 +1,64 @@
...
+
+By deploying {product-title} into an existing AWS VPC, you might be able to avoid limit constraints in new accounts or more easily abide by the operational constraints that your company's guidelines set. This is a good option to use if you cannot obtain the infrastructure creation permissions that are required to create the VPC yourself.
+
+[id="installation-about-custom-aws-vpcs-requirements_{context}"]
+== Requirements for using your VPC
+
+The installation program will no longer create the following components:
+* Internet gateways (`aws_internet_gateway`)
```
abhinavdahiya: drop all the terraform things in brackets.

modules/installation-about-custom-aws-vpcs.adoc
kalexand-rh: Do you have a specific suggestion to refocus this line?

modules/installation-about-custom-aws-vpcs.adoc
```diff
@@ -0,0 +1,64 @@
...
+= About using a custom VPC
+
+In {product-title} {product-version}, you can deploy a cluster into an existing VPC in Amazon Web Services (AWS). If you do, you must also use existing subnets within the VPC and routing rules.
+
+By deploying {product-title} into an existing AWS VPC, you might be able to avoid limit constraints in new accounts or more easily abide by the operational constraints that your company's guidelines set. This is a good option to use if you cannot obtain the infrastructure creation permissions that are required to create the VPC yourself.
+
+[id="installation-about-custom-aws-vpcs-requirements_{context}"]
+== Requirements for using your VPC
+
+The installation program will no longer create the following components:
```
abhinavdahiya: hmm maybe designing it in a way that describe that installer would be doing X to provide Y, but now the user needs to make sure Y is still possible by doing X or Z as per their choice...

usually the installer creates `Internet gateways`, `NAT gateways`, `Route tables` for subnets to make sure the cluster has connectivity to the Internet.

In bring your VPC case, the user needs to make sure the subnets for the cluster are similarly configured to provide connectivity to the Internet.

modules/installation-about-custom-aws-vpcs.adoc
```diff
@@ -0,0 +1,64 @@
...
+
+Your VPC must meet the following characteristics:
+
+* The VPC’s CIDR block must contain the `Networking.MachineCIDR` range, which is the IP address pool for cluster machines.
+* The VPC must not use the `kubernetes.io/cluster/.*: owned` tag. The installation program modifies your subnets to add the `kubernetes.io/cluster/.*: shared` tag. Your subnets must have at least one free tag slot available for it. Review the current link:https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html#tag-restrictions[Tag Restrictions] in the AWS documentation to ensure that the installation program can add a tag to each subnet that you specify.
+* You must enable the `enableDnsSupport` and `enableDnsHostnames` attributes in your VPC so that the cluster can use the Route53 zones that are attached to the VPC to resolve cluster’s internal DNS records. See link:https://docs.aws.amazon.com/vpc/latest/userguide/vpc-dns.html#vpc-dns-support[DNS Support in Your VPC] in the AWS documentation.
+
+
+If you us a cluster with public access, you must create a public and a private subnet for each availability zone that your cluster uses. No other service can use these subnets.
```
abhinavdahiya: > If you us a cluster with public access

`us` -> `need` I think

> No other service can use these subnets.

the goal is to share, so this proabably won't fly and we don't need this.

kalexand-rh: s/us/use

I thought that the subnets needed to be reserved for OCP itself. Maybe s/service/service external to the cluster?

modules/installation-about-custom-aws-vpcs.adoc
kalexand-rh: Do you have an up-to-date description of the preferred configuration?

modules/installation-about-custom-aws-vpcs.adoc
```diff
@@ -0,0 +1,64 @@
...
+To ensure that the subnets that you provide to the installation program are suitable, it confirms the following data:
+
+* All the subnets that you specify exist.
+* You provide private subnets.
+* The subnet CIDRs belong to the machine CIDR that you specified.
+* Unless you use a private cluster, you must provide exactly one public and private subnet for each availability zone. Each availability zone contains no more than one public and one private subnet.
+* You provide a public subnet for each private subnet availability zone. Machines will not be provisioned in availability zones that you do not provide private subnets for.
+* If you are not using a private cluster, you provided a public subnet for each availability zone.
+
+If you destroy a cluster that uses an existing VPC, the VPC is not deleted. When you remove the last {product-title} cluster from a VPC, the `kubernetes.io/cluster/.*: shared` tag is removed from the VPC as well.
```
abhinavdahiya: we don't tag the VPC, only the subnets.

modules/installation-configuration-parameters.adoc
```diff
@@ -175,8 +175,7 @@ such as `io1`.
 such as `c5.9xlarge`.
 
 |`compute.platform.aws.zones`
-|The availability zones where the installation program creates machines for the
-compute MachinePool.
+|The availability zones where the installation program creates machines for the compute MachinePool. If you provide your own VPC, you must specify an availability zone that contains subnets.
```
abhinavdahiya: > you must specify an availability zone that contains subnets.

`you must provide a subnet in that availability zone`?

modules/installation-configuration-parameters.adoc
```diff
@@ -208,6 +207,11 @@ resources that it creates.
 For more information about AWS tags,
 see link:https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html[Tagging Your Amazon EC2 Resources]
 in the AWS documentation.
+
+|`platform.aws.subnets`
+|If you provide the VPC instead of allowing the installation program to create the VPC for you, specify the subnets for the cluster to use. The subnets must be part of the same machineCIDR range that you specify and not used by any other process. For a standard cluster, specify a public and a private subnet for each availability zone. For a private cluster, specify a private subnet for each availability zone.
```
abhinavdahiya: > and not used by any other process.

I don't think that's true.

kalexand-rh: So it's ok for a process outside of the cluster to try to use the subnet range?

modules/installation-about-custom-aws-vpcs.adoc
wking: Instead of making this "you must also..." sounding like "this means you have to do more stuff" maybe we can make subnets part of the "you're here because you want to do this" portion with:

> In {product-title} {product-version}, you can deploy a cluster into existing subnets in an existing VPC in Amazon Web Services (AWS).  By deploying {product-title} into an existing AWS VPC, you might be able to avoid limit constraints in new accounts or more easily abide by the operational constraints that your company's guidelines set. This is a good option to use if you cannot obtain the infrastructure creation permissions that are required to create the VPC yourself.  Because the installer doesn't know what other components are also in your existing subnets, it cannot chose subnet CIDRs and so forth on your behalf, so you must configure networking for the provided subnets yourself.

The wording in my suggesting is still a bit glib, but maybe something in that order?  I'm also fine with the lead-in sentences as they stand in the PR now ;).

modules/installation-about-custom-aws-vpcs.adoc
wking: [The UPI VPC template][1] is the preferred configuration.

[1]: https://docs.openshift.com/container-platform/4.2/installing/installing_aws_user_infra/installing-aws-user-infra.html#installation-cloudformation-vpc_installing-aws-user-infra

modules/installation-about-custom-aws-vpcs.adoc
```diff
@@ -0,0 +1,63 @@
...
+* Route tables
+* VPCs
+* VPC DHCP options
+* VPC endpoints
+
+If you use a custom VPC, you must correctly configure it and its subnets for the installation program and the cluster to use. The installation program cannot subdivide network ranges for the cluster to use, set route tables for the subnets, or set VPC options like DHCP, so you must do so before you install the cluster.
+
+Your VPC must meet the following characteristics:
+
+* The VPC’s CIDR block must contain the `Networking.MachineCIDR` range, which is the IP address pool for cluster machines.
```
wking: nit: do we have a smart/curly apostrophe policy?

modules/installation-about-custom-aws-vpcs.adoc
```diff
@@ -0,0 +1,63 @@
...
+* The VPC must not use the `kubernetes.io/cluster/.*: owned` tag.
```
wking: Do we enforce this?  I don't see a problem with allowing this.  For example, the user may want to install into pre-existing subnets in a VPC they are giving completely to the cluster because they want to set it up with an egress proxy.

modules/installation-about-custom-aws-vpcs.adoc
```diff
@@ -0,0 +1,63 @@
...
+
+If you us a cluster with public access, you must create a public and a private subnet for each availability zone that your cluster uses. You must reserve these subnets for use by the {product-title} cluster. The installation program modifies your subnets to add the `kubernetes.io/cluster/.*: shared` tag, so your subnets must have at least one free tag slot available for it. Review the current link:https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html#tag-restrictions[Tag Restrictions] in the AWS documentation to ensure that the installation program can add a tag to each subnet that you specify.
+
+To ensure that the subnets that you provide to the installation program are suitable, it confirms the following data:
+
+* All the subnets that you specify exist.
+* You provide private subnets.
+* The subnet CIDRs belong to the machine CIDR that you specified.
+* Unless you use a private cluster, you must provide exactly one public and private subnet for each availability zone. Each availability zone contains no more than one public and one private subnet.
```
wking: "Unless you use a private cluster" -> "If you are using a public cluster" (and also two entries below), to avoid having to think through the negative case?

modules/installation-about-custom-aws-vpcs.adoc
```diff
@@ -0,0 +1,63 @@
...
+* You provide a public subnet for each private subnet availability zone. Machines will not be provisioned in availability zones that you do not provide private subnets for.
+* If you are not using a private cluster, you provided a public subnet for each availability zone.
+
+If you destroy a cluster that uses an existing VPC, the VPC is not deleted. When you remove the{product-title} cluster from a VPC, the `kubernetes.io/cluster/.*: shared` tag is removed from the subnets that it used.
```
wking: nit: `the{product-title}` -> `the {product-title}`

modules/installation-about-custom-aws-vpcs.adoc
```diff
@@ -0,0 +1,63 @@
...
+
+The AWS credentials that you use when you create your cluster do not need the networking permissions that are required to make VPCs and networking core components within the VPC, such as subnets, routing tables, internet gateways, NAT, and VPN. You still need permission to make the application resources that the machines within the cluster require, such as ELBs, security groups, S3 buckets, and nodes.
+
+[id="installation-about-custom-aws-vpcs-isolation_{context}"]
+== Isolation between clusters
+
+If you deploy {product-title} to an existing network, the isolation of cluster services is reduced in the following ways:
+
+* You can install multiple {product-title} clusters in the same VPC.
+* ICMP ingress is allowed to entire network.
```
wking: `to entire network` -> `from the entire network`, because we're talking about ingress?

modules/installation-configuration-parameters.adoc
wking: > So it's ok for a process outside of the cluster to try to use the subnet range?

Yup, you can share the subnets just like the VPC can be shared.  You may not want to share the CIDR, because tools would then be unable to route to both the subnet with that CIDR and whatever the other thing was.  But whatever, that's up to the user to decide if they would need simultaneous routing to both places or not.

modules/installation-about-custom-aws-vpcs.adoc
kalexand-rh: I'm not sure if we do - it was called out in the design doc. @abhinavdahiya, will you PTAL?

kalexand-rh:
@wking, I've made some updates. Will you PTAL?

modules/installation-custom-aws-vpc.adoc
```diff
@@ -0,0 +1,136 @@
+// Module included in the following assemblies:
+//
+// * installing/installing_aws/installing-aws-vpc.adoc
+
+[id="installation-custom-aws-vpc_{context}"]
+= About using a custom VPC
+
+In {product-title} {product-version}, you can deploy a cluster into existing subnets in an existing VPC in Amazon Web Services (AWS). By deploying {product-title} into an existing AWS VPC, you might be able to avoid limit constraints in new accounts or more easily abide by the operational constraints that your company's guidelines set. If you cannot obtain the infrastructure creation permissions that are required to create the VPC yourself, use this installation option.
+
+Because the installation program cannot know what other components are also in your existing subnets, it cannot chose subnet CIDRs and so forth on your behalf. You must configure networking for the subnets that you install your cluster to yourself.
```
abhinavdahiya: `chose` -> `choose` ??

modules/installation-custom-aws-vpc.adoc
```diff
@@ -0,0 +1,136 @@
...
+
+* The VPC's CIDR block must contain the `Networking.MachineCIDR` range, which is the IP address pool for cluster machines.
+* The VPC must not use the `kubernetes.io/cluster/.*: owned` tag.
+* You must enable the `enableDnsSupport` and `enableDnsHostnames` attributes in your VPC so that the cluster can use the Route53 zones that are attached to the VPC to resolve cluster’s internal DNS records. See link:https://docs.aws.amazon.com/vpc/latest/userguide/vpc-dns.html#vpc-dns-support[DNS Support in Your VPC] in the AWS documentation.
+
+If you use a cluster with public access, you must create a public and a private subnet for each availability zone that your cluster uses. You must reserve these subnets for use by the {product-title} cluster. The installation program modifies your subnets to add the `kubernetes.io/cluster/.*: shared` tag, so your subnets must have at least one free tag slot available for it. Review the current link:https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html#tag-restrictions[Tag Restrictions] in the AWS documentation to ensure that the installation program can add a tag to each subnet that you specify.
```
abhinavdahiya: > You must reserve these subnets for use by the {product-title} cluster.

hmm. what's the intention for this section?

modules/installation-custom-aws-vpc.adoc
```diff
@@ -0,0 +1,136 @@
...
+[id="installation-custom-aws-vpc-validation_{context}"]
+== VPC validation
+
+To ensure that the subnets that you provide to the installation program are suitable, it confirms the following data:
+
+* All the subnets that you specify exist.
+* You provide private subnets.
+* The subnet CIDRs belong to the machine CIDR that you specified.
+* You provide subnets for each availability zone. Each availability zone contains no more than one public and one private subnet. If you use a private cluster, provide only a private subnet for each availability zone. Otherwise, provide exactly one public and private subnet for each availability zone.
+* You provide a private subnet for each public subnet availability zone. Machines are not provisioned in availability zones that you do not provide private subnets for.
```
abhinavdahiya: `You provide a private subnet for each public subnet availability zone`

since private subnets are always required.. `You provide a public subnet for each private subnet availability zone` ??

modules/installation-custom-aws-vpc.adoc
kalexand-rh: I removed the other reference to reserving the subnets and missed this one. Thank you!

kalexand-rh:
I think that I've addressed all of Abhinav's requests. (Please let me know if you have more changes!)

@jianlinliu, will you PTAL or suggest a different reviewer?

jianlinliu:
Most looks to me. One concern, from customer perspective, when I am reading the whole doc, I am curious what is outstanding difference in this section compare with other section. e.g: espeically compare with "installing-aws-customizations.html#installation-aws-config-yaml_installing-aws-customizations" and  "installing-aws-customizations.html#installation-configuration-parameters_installing-aws-customizations". I understand this is modularizing doc, so that be easy to assemble different topic doc. So I do not object to this modular doc because I have not better idea yet. Just maybe bring some potential improvement if you have some better idea.

kalexand-rh:

kalexand-rh:
@jianlinliu, I've added some conditional statements that hide the subnet option from the general customization topic.

installing/installing_aws/installing-aws-vpc.adoc
```diff
@@ -0,0 +1,61 @@
+[id="installing-aws-vpc"]
+= Installing a cluster on AWS to an existing VPC
+include::modules/common-attributes.adoc[]
+:context: installing-aws-vpc
+
+toc::[]
+
+In {product-title} version {product-version}, you can install a cluster into an existing VPC on
```
bergerhoffer: You never spell out VPC anywhere. I'd suggest spelling out at least once at the beginning of the assembly.

bergerhoffer (commented):
Just a few comments and suggestions!

modules/installation-custom-aws-vpc.adoc
```diff
@@ -0,0 +1,136 @@
+// Module included in the following assemblies:
+//
+// * installing/installing_aws/installing-aws-vpc.adoc
+
+[id="installation-custom-aws-vpc_{context}"]
+= About using a custom VPC
+
+In {product-title} {product-version}, you can deploy a cluster into existing subnets in an existing VPC in Amazon Web Services (AWS). By deploying {product-title} into an existing AWS VPC, you might be able to avoid limit constraints in new accounts or more easily abide by the operational constraints that your company's guidelines set. If you cannot obtain the infrastructure creation permissions that are required to create the VPC yourself, use this installation option.
```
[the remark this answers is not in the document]
kalexand-rh: @bergerhoffer, I want to keep both because you're trying to avoid the fact that your default, or carefully curated corporate, account is limited. (You usually don't have enough space in your account to make all the cluster components, and it's apparently common for cloud admins to not let you make networking components.)

Would "avoid service limit constraints" scan better?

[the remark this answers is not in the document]
bergerhoffer: @kalexand-rh oh i see, i didn't realize that "limit constraints" was what it actually is there. It sounded like a double word to me. If it works for you, "avoid service limit constraints" would be much clearer to me.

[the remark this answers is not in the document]
kalexand-rh: It does indeed! Thank you for pointing this out. I'll update all of the VPC PRs. :)

PR #18093, opened 2019-11-13
Tweaks for RHOSP vs. others. BZ 17617109.

maxwelldb:
Context: https://bugzilla.redhat.com/show_bug.cgi?id=1767109

This PR:
* Adds a conditional statement so that 'Configuring a cloud platform account' isn't a prereq in RHOSP assemblies
* Changes the first line of the run installer module to be more platform-agnostic.

### DIFF
--- a/modules/installation-launching-installer.adoc
+++ b/modules/installation-launching-installer.adoc
@@ -60,7 +60,7 @@ endif::[]
 [id="installation-launching-installer_{context}"]
 = Deploy the cluster
 
-You can install {product-title} on a compatible cloud.
+You can install {product-title} on a compatible cloud platform.
 
 [IMPORTANT]
 ====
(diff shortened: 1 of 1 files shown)


modules/installation-launching-installer.adoc
```diff
@@ -69,7 +69,8 @@ You can run the `create cluster` command of the installation program only once,
 
 .Prerequisites
 
-* Configure an account with the cloud platform that hosts your cluster.
+ifndef::osp[* Configure an account with the cloud platform that hosts your cluster.]
+
```
kalexand-rh: Don't you need to do some credential work on OSP, too? Otherwise, this looks reasonable to me.

kalexand-rh (commented):
LGTM

kalexand-rh:
I spotchecked the way that the if statements render, and it looks fine to me.

PR #18096, opened 2019-11-13
Fixing instances of {ProductName}; adding the word Platform

ousleyp:
- There were a few instances where `{ProductName}` was still used instead of `{CNVProductName}`; Aiko pointed these out in the localization review
- Red Hat Ansible Automation is now called Red Hat Ansible Automation Platform

There is no BZ associated with these changes.
They should be cherrypicked to enterprise-4.2 and enterprise-4.3.


sheriff-rh (approved):
Looks good!

PR #18130, opened 2019-11-15
templates: pass OSO version when searching

jiri-fiala:
This small change includes the version filter for OSO docs searching (similarly as it's already in place with the other docs), so only relevant Starter or Pro results are to be displayed.
Also, it updates the search endpoint.

The search functionality can be previewed here:
[OSO Starter](http://commercial-docs-pro.b9ad.pro-us-east-1.openshiftapps.com/online/starter/welcome/index.html)
[OSO Pro](http://commercial-docs-pro.b9ad.pro-us-east-1.openshiftapps.com/online/pro/welcome/index.html)

### DIFF
--- a/_javascripts/hc-search.js
+++ b/_javascripts/hc-search.js
@@ -51,7 +51,7 @@ function hcsearch(searchParams) {
   var hcSearchResult = $("#hc-search-results");
 
   // the "searchprovider" is to return a JSON response in the expected format
-  var searchprovider = "https://help.openshift.com/search/search_custom.php";
+  var searchprovider = "https://search.help.openshift.com/json";
   var searchReq = { "q" : searchParams.q + searchParams.urlFilter,
                     "l" : searchParams.label,
                     "si" : searchParams.si }  // q = query, l = label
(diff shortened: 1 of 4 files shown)


vikram-redhat:
@jiri-fiala Hey - does this need to go into older branches? search-commercial.html only exists in the 4.x stream.

jiri-fiala:
Hi @vikram-redhat, I don't think the `search-commercial.html` needs to go to any older v3 branches. It's the "universal" search, that mixes results from all docs and is on the front page (without selecting any product/version).

However, it looks like I made a typo in that file. This is fixed in https://github.com/openshift/openshift-docs/pull/18149 (rebased against the current `master`). It also fixes the label field name (which was not apparent on my stage at all due to working URL filters.)

Currently, it looks like the new endpoint is used (so the new `hc_search.js` is served), but URL filtering is not working for OSO yet - an older `_page_openshift.html.erb` template appears to be used to generate documentation pages. I think that the template should be merged into 3.11 branch too, according to the distro map.
I can open separate PRs in `master` and `master-3` branches, if that helps with the cherry picking. Or maybe against specific enterprise branches, if those diverged from `master` and `master-3` too much? Please let me know what's needed.

vikram-redhat:
Thanks @jiri-fiala. There should be no need to separate _page_openshift.html.erb out for 3.x and 4.x and I will CP them in by tomorrow.

PR #18156, opened 2019-11-18
OSP: instruct users to uncompress the artifact

miabbott:
We've received multiple reports from users who did not uncompress the
OpenStack artifact before trying to use it in their cluster.  Let's
explicitly instruct them to uncompress the artifact before having them
try to use it in a cluster.

### DIFF
--- a/modules/installation-osp-creating-image.adoc
+++ b/modules/installation-osp-creating-image.adoc
@@ -19,7 +19,12 @@ The {product-title} installation program requires that a {op-system-first} image
 
 . Under *Version*, select version 4.2.0 for RHEL 8.
 
-. Download the _Red Hat Enterprise Linux CoreOS - OpenStack Image (QCOW)_.
+. Download the _Red Hat Enterprise Linux CoreOS - OpenStack Image (QCOW)_ and decompress the image.
++
+[NOTE]
+====
+You must decompress the OpenStack image before the cluster can use it.
+====
 
 . From the image that you downloaded, create an image that is named `rhcos` in your cluster by using the {rh-openstack} CLI:
 +


maxwelldb:
@miabbott Can you confirm that these customers are retrieving the artifact from the portal downloads page that's linked in these docs?

@morenod As I recall, this problem is related to the outcome of this bug: https://bugzilla.redhat.com/show_bug.cgi?id=1763803#c6

Can we verify what the safest recommendation is here?

modules/installation-osp-creating-image.adoc
```diff
@@ -21,6 +21,8 @@ The {product-title} installation program requires that a {op-system-first} image
 
 . Download the _Red Hat Enterprise Linux CoreOS - OpenStack Image (QCOW)_.
 
+. After downloading the file, use `gunzip` to uncompress the file before trying to create an image in OpenStack.
```
maxwelldb: >Use `gunzip` to decompress the image.

Alternatively, combine this step with step 1:

>Download the _Red Hat Enterprise Linux CoreOS - OpenStack Image (QCOW)_ and extract the image.

modules/installation-osp-creating-image.adoc
```diff
@@ -21,6 +21,8 @@ The {product-title} installation program requires that a {op-system-first} image
...
+
 . From the image that you downloaded, create an image that is named `rhcos` in your cluster by using the {rh-openstack} CLI:
```
maxwelldb: "downloaded" -> "extracted"/"decompressed"

maxwelldb (changes requested):

miabbott:
> @miabbott Can you confirm that these customers are retrieving the artifact from the portal downloads page that's linked in these docs?

See https://bugzilla.redhat.com/show_bug.cgi?id=1764856 as an example.

> @morenod As I recall, this problem is related to the outcome of this bug: https://bugzilla.redhat.com/show_bug.cgi?id=1763803#c6
> 
> Can we verify what the safest recommendation is here?

Unfortunately there is going to be some slight differences between 4.2 and 4.3. 

 In 4.2, some of the artifacts were compressed, but lacked the `.gz` file extension.  This caused more grief than anything, so for 4.3 we made the choice to have *ALL* compressed artifacts include the `.gz` extension.

Thus, this change is to reflect that the OpenStack artifact is in fact compressed (it should ship with the `.gz` extension) and users have to decompress it before it can be usable.  I didn't want this to be omitted from the docs, since I've seen reports in Slack about folks not being aware they need to decompress the OpenStack artifact before trying to use it.

maxwelldb:
> > @miabbott Can you confirm that these customers are retrieving the artifact from the portal downloads page that's linked in these docs?
> 
> See https://bugzilla.redhat.com/show_bug.cgi?id=1764856 as an example.
> 
> > @morenod As I recall, this problem is related to the outcome of this bug: https://bugzilla.redhat.com/show_bug.cgi?id=1763803#c6
> > Can we verify what the safest recommendation is here?
> 
> Unfortunately there is going to be some slight differences between 4.2 and 4.3.
> 
> In 4.2, some of the artifacts were compressed, but lacked the `.gz` file extension. This caused more grief than anything, so for 4.3 we made the choice to have _ALL_ compressed artifacts include the `.gz` extension.
> 
> Thus, this change is to reflect that the OpenStack artifact is in fact compressed (it should ship with the `.gz` extension) and users have to decompress it before it can be usable. I didn't want this to be omitted from the docs, since I've seen reports in Slack about folks not being aware they need to decompress the OpenStack artifact before trying to use it.

@miabbott IIRC, the customer groups RHCOS images by major version only. Does it follow that this change should apply to both 4.2 and 4.3, then?

miabbott:
> @miabbott IIRC, the customer groups RHCOS images by major version only. Does it follow that this change should apply to both 4.2 and 4.3, then?

Yes, in both versions the user will have to decompress the artifact before it can be used in the OpenStack cluster.  The difference I was trying to highlight was that the 4.3 artifacts will have the `.gz` extension if they are compressed, whereas the 4.2 artifacts may be missing the extension.

modules/installation-osp-creating-image.adoc
```diff
@@ -19,7 +19,12 @@ The {product-title} installation program requires that a {op-system-first} image
 
 . Under *Version*, select version 4.2.0 for RHEL 8.
 
-. Download the _Red Hat Enterprise Linux CoreOS - OpenStack Image (QCOW)_.
+. Download the _Red Hat Enterprise Linux CoreOS - OpenStack Image (QCOW)_ and decompress the image..
```
maxwelldb: Doubled periods at the end of line 22.

modules/installation-osp-creating-image.adoc
```diff
@@ -19,7 +19,12 @@ The {product-title} installation program requires that a {op-system-first} image
...
++
+[NOTE]
+====
+The OpenStack image must be decompressed before being used by the cluster.
```
maxwelldb: "You must decompress the OpenStack image before the cluster can use it."

Otherwise, :+1:

maxwelldb (changes requested):

maxwelldb (approved):
Thanks! Just waiting on a nod from QE before merging.

morenod (approved):
lgtm

kalexand-rh:
Thank you, @morenod! Since Max says that this change was just pending QE, I'm going to merge it.

PR #17772, opened 2019-10-31
Enhance fluentd info

mijicherian:
image

PR #18162, opened 2019-11-18
Adding Kibana index mode to 4.x

mburke5678:
After discovering that the section on Kibana index mode configuration did not convert from 3.11 to 4.x, I created this section to add the missing info.  

https://github.com/openshift/openshift-docs/pull/17772#discussion_r346873590


mburke5678:
However @ jcantrill suggested not adding the info. Closing.

PR #18178, opened 2019-11-19
Replacing example CR to About Cluster Logging

mburke5678:
Replacing CR in _About the Cluster Logging Custom Resource Definition_.
https://docs.openshift.com/container-platform/4.2/logging/cluster-logging.html#cluster-logging-configuring-crd_cluster-logging

### DIFF
--- a/modules/cluster-logging-about-crd.adoc
+++ b/modules/cluster-logging-about-crd.adoc
@@ -3,19 +3,60 @@
 // * logging/cluster-logging.adoc
 
 [id="cluster-logging-configuring-crd_{context}"]
-= About the Cluster Logging Custom Resource Definition
+= About the Cluster Logging Custom Resource
 
-The Cluster Logging Operator Custom Resource Definition (CRD) defines a complete cluster logging deployment
-that includes all the components of the logging stack to collect, store and visualize logs.
-
-You should never have to modify this CRD. To make changes to your deployment, create and modify a specific Custom Resource (CR).
+To make changes to your cluster logging deployment, create and modify the Cluster Logging Custom Resource (CR).
 Instructions for creating or modifying a CR are provided in this documentation as appropriate.
 
 The following is an example of a typical Custom Resource for cluster logging.
 
-[id="cluster-logging-configuring-about-sample_{context}"]
+[id="efk-logging-configuring-about-sample_{context}"]
 .Sample Cluster Logging CR
 ifdef::openshift-dedicated[]
+[source,yaml]
+----
+apiVersion: "logging.openshift.io/v1"
+kind: "ClusterLogging"
+metadata:
+  name: "instance"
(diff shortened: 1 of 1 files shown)


mburke5678:
@sheriff-rh Do you recall why you added a `ifdef::openshift-dedicated[]` tag here?

https://github.com/openshift/openshift-docs/commit/f5a93ffd72a828762dd16c4d9d377bbb89c00a12#diff-2e73b7f1b268e41025eec8ac0ed41fcdR18

jboxman:
Relatedly, this is a CRD: https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#create-a-customresourcedefinition

So it's an example CR, rather than a CRD, in this PR. Somewhat confusingly, a CRD appears to be implemented using a CR, but it's through the `apiVersion: apiextensions.k8s.io/v1` API in k8s.

mburke5678:
@jboxman You are correct. As I recall, we wanted to put the CRDs in the docs until we realized that the user edits the CRs (and the CRDs can be huge), and it seemed better to use the CR (I could be remembering this).

jboxman:
@mburke5678, what passes as the CRD will actually be in the API docs, as a table of parameters and values.

sheriff-rh:
> @sheriff-rh Do you recall why you added a `ifdef::openshift-dedicated[]` tag here?
> 
> [f5a93ff#diff-2e73b7f1b268e41025eec8ac0ed41fcdR18](https://github.com/openshift/openshift-docs/commit/f5a93ffd72a828762dd16c4d9d377bbb89c00a12#diff-2e73b7f1b268e41025eec8ac0ed41fcdR18)

Hey @mburke5678 ! There is a separate yaml for dedicated vs enterprise, since dedicated is a very tightly controlled configuration. Hope this helps!

mburke5678:
@sheriff-rh I see it now. The Enterprise YAML isn't in Master (in cluster-logging-configuring-crd.adoc), likely because of the file name change (??).  I copied the two CRs from the 4.1 version and added them to the version in master. Can you take a quick look?
https://raw.githubusercontent.com/openshift/openshift-docs/afb4f5f82b324da4ee89493b5f11edd703149f3f/modules/cluster-logging-about-crd.adoc

sheriff-rh:
> @sheriff-rh I see it now. The Enterprise YAML isn't in Master (in cluster-logging-configuring-crd.adoc), likely because of the file name change (??). I copied the two CRs from the 4.1 version and added them to the version in master. Can you take a quick look?
> https://raw.githubusercontent.com/openshift/openshift-docs/afb4f5f82b324da4ee89493b5f11edd703149f3f/modules/cluster-logging-about-crd.adoc

Looks great @mburke5678 , thanks for checking!

PR #18182, opened 2019-11-19
Adding note that you must manually create Ingress DNS records if prox…

bergerhoffer:
…y is enabled

### DIFF
--- a/modules/installation-configure-proxy.adoc
+++ b/modules/installation-configure-proxy.adoc
@@ -8,6 +8,15 @@
 // * installing/installing_restricted_networks/installing-restricted-networks-vsphere.adoc
 // * installing/installing_vsphere/installing-vsphere.adoc
 
+ifeval::["{context}" == "installing-gcp-user-infra"]
+:gcp-upi:
+endif::[]
+ifeval::["{context}" == "installing-restricted-networks-aws"]
+:aws-restricted-upi:
+endif::[]
+ifeval::["{context}" == "installing-aws-user-infra"]
+:aws-upi:
+endif::[]
 ifeval::["{context}" == "installing-bare-metal"]
 :bare-metal:
 endif::[]
@@ -94,6 +103,22 @@ Only the Proxy object named `cluster` is supported, and no additional
 proxies can be created.
 ====
 
+ifdef::aws-upi,aws-restricted-upi,gcp-upi[]
+
+When the cluster-wide proxy is enabled, you must manually create the Ingress DNS records.
+
+// TODO: xref: https://docs.openshift.com/container-platform/4.2/installing/installing_aws_user_infra/installing-aws-user-infra.html#installation-create-ingress-dns-records_installing-aws-user-infra
+endif::aws-upi,aws-restricted-upi,gcp-upi[]
+
+ifeval::["{context}" == "installing-gcp-user-infra"]
+:!gcp-upi:
+endif::[]
+ifeval::["{context}" == "installing-restricted-networks-aws"]
+:!aws-restricted-upi:
+endif::[]
+ifeval::["{context}" == "installing-aws-user-infra"]
+:!aws-upi:
+endif::[]
 ifeval::["{context}" == "installing-bare-metal"]
 :!bare-metal:
 endif::[]
--- a/modules/installation-create-ingress-dns-records.adoc
+++ b/modules/installation-create-ingress-dns-records.adoc
@@ -6,7 +6,7 @@
 [id="installation-create-ingress-dns-records_{context}"]
 = Creating the Ingress DNS Records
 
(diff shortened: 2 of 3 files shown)


bergerhoffer:
@danehans can you please review this update? I put the note both in the proxy docs, and in the section on creating the Ingress DNS records.

For example, here is the preview for the AWS UPI docs:

* https://proxy-dns--ocpdocs.netlify.com/openshift-enterprise/latest/installing/installing_aws_user_infra/installing-aws-user-infra.html#installation-configure-proxy_installing-aws-user-infra
* https://proxy-dns--ocpdocs.netlify.com/openshift-enterprise/latest/installing/installing_aws_user_infra/installing-aws-user-infra.html#installation-create-ingress-dns-records_installing-aws-user-infra

modules/installation-create-ingress-dns-records.adoc
```diff
@@ -6,7 +6,7 @@
 [id="installation-create-ingress-dns-records_{context}"]
 = Creating the Ingress DNS Records
 
-If you removed the DNS Zone configuration, manually create DNS records that point to the Ingress load balancer.
+If you removed the DNS Zone configuration, or if you enabled the cluster-wide proxy, manually create DNS records that point to the Ingress load balancer.
```
danehans: s/Ingress load balancer/hostname of the router load balancer service/. You may want to provide an example on how to get the Ingress load-balancer. Here's an example for the default router:
```
$ oc get svc/router-default -n openshift-ingress -o yaml | grep hostname
    - hostname: a4a37c40e0af011eabdc302eee56a296-792674788.us-west-2.elb.amazonaws.com
```

modules/installation-configure-proxy.adoc
```diff
@@ -94,6 +103,22 @@ Only the Proxy object named `cluster` is supported, and no additional
 proxies can be created.
 ====
 
+ifdef::aws-upi,aws-restricted-upi,gcp-upi[]
+
+When the cluster-wide proxy is enabled, you must manually create the Ingress DNS records.
+
+// TODO: xref: https://docs.openshift.com/container-platform/4.2/installing/installing_aws_user_infra/installing-aws-user-infra.html#installation-create-ingress-dns-records_installing-aws-user-infra
```
danehans: If the TODO stays, please add some context for others to understand.

modules/installation-create-ingress-dns-records.adoc
bergerhoffer: @danehans What would someone need to know the host name for? (I don't see anything in this procedure that uses it).

modules/installation-configure-proxy.adoc
bergerhoffer: Unfortunately we can't currently link between modules. This is just a placeholder if we ever can in the future, and I would use this link in the above sentence.

modules/installation-configure-proxy.adoc
```diff
@@ -94,6 +103,22 @@ Only the Proxy object named `cluster` is supported, and no additional
 proxies can be created.
 ====
 
+ifdef::aws-upi,aws-restricted-upi,gcp-upi[]
```
kalexand-rh: If you expect that the same statement will always be the same for all UPI cloud clusters, please use the same context key for each cloud. It'll make it slightly easier to add and maintain other UPI clouds, and I know that several are in the queue.

modules/installation-create-ingress-dns-records.adoc
```diff
@@ -6,7 +6,7 @@
 [id="installation-create-ingress-dns-records_{context}"]
 = Creating the Ingress DNS Records
 
-If you removed the DNS Zone configuration, manually create DNS records that point to the Ingress load balancer.
+If you removed the DNS Zone configuration, or if you enabled the cluster-wide proxy, manually create DNS records that point to the host name of the router load balancer service.
```
kalexand-rh: I'd say "If you removed the DNS Zone configuration or enabled the cluster-wide proxy," to place equal importance on both options.

modules/installation-gcp-user-infra-adding-ingress.adoc
```diff
@@ -6,8 +6,8 @@
 = Optional: Adding the ingress DNS records
 
 If you removed the DNS Zone configuration when creating Kubernetes manifests and
-generating Ignition configs, you must manually create DNS records that point at
-the ingress load balancer. You can create either a wildcard
+generating Ignition configs, or if you enabled the cluster-wide proxy, you must manually create DNS records that point to
```
kalexand-rh: Same here.

kalexand-rh (commented):
I have a request and a suggestion, but this LGTM.

vikram-redhat:
@bergerhoffer can this be updated or closed if no longer required?

PR #18206, opened 2019-11-20
osdocs771 AWS private clusters

kalexand-rh:
https://jira.coreos.com/browse/OSDOCS-771

@abhinavdahiya, I'm working on the modules for installing private clusters. Will you PTAL at the new content and let me know what else is required? I'm going to make the BYO VPC steps a requirement for this story, but is there anything that I'm missing?

### DIFF
--- a/modules/private-clusters-about-aws.adoc
+++ b/modules/private-clusters-about-aws.adoc
@@ -1,0 +1,26 @@
+// Module included in the following assemblies:
+//
+// * installing/installing_aws/installing-aws-private.adoc
+
+[id="private-clusters-about-aws_{context}"]
+= Private clusters in AWS
+
+To create a private cluster on Amazon Web Services (AWS), you must provide an existing private VPC and subnets to host the cluster. The installation program must also be able to resolve the DNS records that the cluster requires. The installation program configures the Ingress Operator and API server for access from only the private network.
+
+The cluster still requires access to Internet to access the AWS APIs.
+
+The following items are not required or created when you install a private cluster:
+
+* Public subnets
+* Public load balancers, which support public ingress
+* A public Route 53 Zone that matches the `baseDomain` for the cluster
+
+The installation program does use the `baseDomain` that you specify to create a private Route 53 Zone and the required records for the cluster. The cluster is configured so that the Operators do not create public records for the cluster and all cluster machines are placed in the private subnets that you specify.
+
+[id="private-clusters-limitations-aws_{context}"]
+== Limitations
+
+The ability to add public functionality to a private cluster is limited.
+
+* You cannot make the Kubernetes API endpoints public after installation without taking additional actions, including creating public subnets in the VPC for each availablity zone in use, creating a public load balancer, and configuring the control plane security groups to allow traffic from Internet on 6443 (Kubernetes API port).
+* If you use a public Service type load balancer, you must tag a public subnet in each availability zone with `kubernetes.io/cluster/<cluster-infra-id>: shared` so that AWS can use them to create public load balancers.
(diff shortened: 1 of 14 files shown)


modules/private-clusters-about-aws.adoc
```diff
@@ -0,0 +1,26 @@
+// Module included in the following assemblies:
+//
+// * installing/
+
+[id="private-clusters-about-aws_{context}"]
+= Private clusters in AWS
+
+To create a private cluster on Amazon Web Services (AWS), you must provide an existing private VPC and subnets to host the cluster. The installation program must also be able to resolve the DNS records that the cluster requires. The installation program configures the Ingress Operator and API server for only internal traffic.
```
abhinavdahiya: > only internal traffic

Maybe replace it with something closer to, ` for access from the private network only`

abhinavdahiya:
lgtm

kalexand-rh:
@jianlinliu, will you PTAL? I cherry-picked the existing VPC change onto this PR so that I could build the assembly, so many of the changes are duplicated from that PR. 

The content that's new for this PR is:
installing/installing_aws/installing-aws-private.adoc
modules/private-clusters-default.adoc 
modules/private-clusters-about-aws.adoc 
modules/installation-aws-config-yaml.adoc, line 102 and its note
modules/installation-configuration-parameters.adoc, line 138

jianlinliu:
Hmm, personally I think using existing VPC and private cluster are two different scenarios. Of course, user can combine the two scenarios for different purpose. E.g: if user want a private cluster, user even can run openshift-install binary in their laptop, but eventually installation will report a fake negative result to fail the whole installation, because installer can not access the cluster api and *.apps endpoint outside the vpc, but actually the cluster is already provisioned successfully. To make the whole progress more friendly, maybe we could suggest user firstly create a VPC, create a bastion host in this VPC, download installer binary onto this bastion host, run the whole installation on this bastion host to create a private cluster. From my understanding, an existing VPC is not a MUST from technical perspective for a private cluster. I also go through the whole preview, seem like we did not emphasise why we must have an existing VPC, did not mention where installation should be happening either (should be in a bastion host). 
For existing VPC, some customers maybe indeed want create multiple clusters facing internet in one shared VPC, nothing with private cluster.
So here we combine the two different scenarios into one topic, probably confuse customers.

kalexand-rh:
@jianlinliu, the enhancement makes it sound like the VPC is required: https://github.com/openshift/enhancements/blob/master/enhancements/installer/aws-internal-clusters.md#proposal

jianlinliu:
Yeah, from customer use experience, to be more friendly, pre-existing VPC can be treated as a requirement.
My last comment is suggesting we still need a separate section for BYO VPC. I thought this PR will deprecate https://github.com/openshift/openshift-docs/pull/18729, seem like not. If that, I am okay with an existing VPC is required.
If that, the only comments is:

> seem like we did not emphasise why we must have an existing VPC, did not mention where installation should be happening either (should be in a bastion host).

modules/private-clusters-default.adoc
```diff
@@ -0,0 +1,15 @@
+// Module included in the following assemblies:
+//
+// * installing/installing_aws/installing-aws-private.adoc
+// * installing/installing_gcp/installing-gcp-private.adoc
+
+[id="private-clusters-default_{context}"]
+= Private clusters
+
+If your environment does not require an external internet connection, you can deploy a private {product-title} cluster that does not expose external endpoints. Private clusters are accessible from only an internal network and are not visible to the Internet.
```
kalexand-rh: @jianlinliu, does this module address your concerns about context for why you need your own network and using the bastion host?

kalexand-rh:

sheriff-rh (approved):
Excellent work! All changes look good to me.

modules/private-clusters-default.adoc
jianlinliu: LGTM.

PR #18207, opened 2019-11-20
Bug 1712525, Configuring a custom PKI

codyhoag:
https://bugzilla.redhat.com/show_bug.cgi?id=1712525

Reorganized from previous submission: #18004

### DIFF
--- a/networking/configuring-a-custom-pki.adoc
+++ b/networking/configuring-a-custom-pki.adoc
@@ -1,0 +1,47 @@
+[id="configuring-a-custom-pki"]
+= Configuring a custom PKI
+include::modules/common-attributes.adoc[]
+:context: configuring-a-custom-pki
+
+toc::[]
+
+Some platform components, such as the web console, use Routes for communication and
+must trust other components' certificates to interact with them. If
+you are using a custom public key infrastructure (PKI), you must configure it so
+its privately signed CA certificates are recognized across the cluster.
+
+You can leverage the Proxy API to add cluster-wide trusted CA certificates. You
+must do this either during installation or at runtime.
+
+* During _installation_, xref:installation-configure-proxy_{context}[configure the cluster-wide proxy]. You must define your
+privately signed CA certificates in the `install-config.yaml` file's
+`additionalTrustBundle` setting.
++
+The installation program generates a ConfigMap that is named `user-ca-bundle`
+that contains the additional CA certificates you defined. The Cluster Network
+Operator then creates a `trusted-ca-bundle` ConfigMap that merges these CA
+certificates with the {op-system-first} trust bundle; this ConfigMap is
+referenced in the Proxy object's `trustedCA` field.
+
+* At _runtime_, xref:nw-proxy-configure-object_{context}[modify the default Proxy object to include your privately signed
+CA certificates] (part of cluster's proxy enablement workflow). This involves
+creating a ConfigMap that contains the privately signed CA certificates that
+should be trusted by the cluster, and then modifying the proxy resource with the
+`trustedCA` referencing the privately signed certificates' ConfigMap.
+
+[NOTE]
+====
+The installer configuration's `additionalTrustBundle` field and the proxy
+resource's `trustedCA` field are used to manage the cluster-wide trust bundle;
+`additionalTrustBundle` is used at install time and the proxy's `trustedCA` is
+used at runtime.
+
+The `trustedCA` field is a reference to a `ConfigMap` containing the custom
+certificate and key pair used by the cluster component.
+====
+
+include::modules/installation-configure-proxy.adoc[leveloffset=+1]
+
+include::modules/nw-proxy-configure-object.adoc[leveloffset=+1]
+
+include::modules/certificate-injection-using-operators.adoc[leveloffset=+1]
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -387,6 +387,8 @@ Topics:
     File: configuring-ingress-cluster-traffic-nodeport
 - Name: Configuring the cluster-wide proxy
   File: enable-cluster-wide-proxy
+- Name: Configuring a custom PKI
+  File: configuring-a-custom-pki
 ---
 Name: Storage
 Dir: storage
--- a/builds/setting-up-trusted-ca.adoc
+++ b/builds/setting-up-trusted-ca.adoc
@@ -22,3 +22,4 @@ include::modules/configmap-adding-ca.adoc[leveloffset=+1]
 
 * link:https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#create-a-configmap[Create a ConfigMap]
 * link:https://kubectl.docs.kubernetes.io/pages/app_management/secrets_and_configmaps.html[Secrets and ConfigMaps]
+* xref:../networking/configuring-a-custom-pki.adoc#configuring-a-custom-pki[Configuring a custom PKI]
--- a/modules/certificate-injection-using-operators.adoc
+++ b/modules/certificate-injection-using-operators.adoc
@@ -1,0 +1,27 @@
+// Module included in the following assemblies:
+//
+// * networking/configuring-a-custom-pki.adoc
+
+[id="certificate-injection-using-operators_{context}"]
+= Certificate injection using Operators
+
+Once your custom CA certificate is added to the cluster via ConfigMap, the
+Cluster Network Operator merges the user-provided and system CA certificates
+into a single bundle and injects the merged bundle into the Operator requesting
+the trust bundle injection.
+
+Operators request this injection by creating an empty ConfigMap with the
+following label:
+
+[source,yaml]
+----
+config.openshift.io/inject-trusted-cabundle="true"
+----
+
+The Operator mounts this ConfigMap into the container's local trust store.
+
+[NOTE]
+====
+Adding a trusted CA certificate is only needed if the certificate is not
+included in the {op-system-first} trust bundle.
+====
--- a/modules/installation-configure-proxy.adoc
+++ b/modules/installation-configure-proxy.adoc
@@ -7,6 +7,7 @@
 // * installing/installing_restricted_networks/installing-restricted-networks-bare-metal.adoc
 // * installing/installing_restricted_networks/installing-restricted-networks-vsphere.adoc
 // * installing/installing_vsphere/installing-vsphere.adoc
+// * networking/configuring-a-custom-pki.adoc
 
 ifeval::["{context}" == "installing-bare-metal"]
 :bare-metal:
--- a/modules/nw-ingress-setting-a-custom-default-certificate.adoc
+++ b/modules/nw-ingress-setting-a-custom-default-certificate.adoc
@@ -1,6 +1,6 @@
 // Module included in the following assemblies:
 //
-// * ingress/ingress-operator.adoc
+// * networking/ingress-operator.adoc
 
 [id="nw-ingress-setting-a-custom-default-certificate_{context}"]
 = Setting a custom default certificate
@@ -12,8 +12,10 @@ custom resource (CR).
 .Prerequisites
 
 * You must have a certificate/key pair in PEM-encoded files, where the
-certificate is signed by a trusted certificate authority and valid for the
-Ingress domain.
+certificate is signed by a trusted certificate authority or by a private trusted
+certificate authority that you have configured.
+
+* Your certificate is valid for the Ingress domain.
 
 * You must have an `IngressController` CR. You may use the default one:
 +
@@ -72,4 +74,4 @@ map[name:custom-certs-default]
 The certificate secret name should match the value used to update the CR.
 
 Once the `IngressController` CR has been modified, the Ingress Operator
-will update the Ingress Controller's deployment to use the custom certificate.
+updates the Ingress Controller's deployment to use the custom certificate.
--- a/modules/nw-proxy-configure-object.adoc
+++ b/modules/nw-proxy-configure-object.adoc
@@ -1,5 +1,6 @@
 // Module included in the following assemblies:
 //
+// * networking/configuring-a-custom-pki.adoc
 // * networking/enable-cluster-wide-proxy.adoc
 
 [id="nw-proxy-configure-object_{context}"]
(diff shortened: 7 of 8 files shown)


codyhoag:
*Configuring a custom PKI* assembly build:

http://file.rdu.redhat.com/~choag/112019/ingress-cert/networking/configuring-a-custom-pki.html

codyhoag:
@danehans can you review the new docs I've created on *Configuring a custom PKI*?

modules/about-certificate-management.adoc
```diff
@@ -0,0 +1,26 @@
...
+
+Operators request this injection by creating an empty ConfigMap with the
+following label:
+
+[source,yaml]
+----
+config.openshift.io/inject-trusted-cabundle="true"
+----
+
+The operator mounts this ConfigMap into the container's local trust store.
```
codyhoag: @danehans perhaps we should list the supported operators here? What would those be?

danehans: I think maintaining the list will be very difficult. Any operator that makes a call external to the cluster requires proxy support. Note that ingress operator will be on this list when https://jira.coreos.com/browse/NE-182 is done.

modules/about-certificate-management.adoc
```diff
@@ -0,0 +1,26 @@
+// Module included in the following assemblies:
+//
+// * networking/configuring-a-custom-pki.adoc
+
+[id="about-certificate-management_{context}"]
+= About certificate management
+
+Once a custom certificate is set, the network operator merges the user-provided
+and system CA certificates into a single bundle and injects the merged bundle
```
danehans: What is meant by "Once a custom certificate is set". It might be better to be explicit stating how a custom ca cert is added to the cluster (i.e. configmap). I would also consider breaking this into 2 or more sentences.

networking/configuring-a-custom-pki.adoc
```diff
@@ -0,0 +1,56 @@
...
+:context: configuring-a-custom-pki
+
+toc::[]
+
+Some platform components (i.e., web console) use Routes for communication and,
+therefore, must trust other components' certificates to interact with them. If
+you're using a custom public key infrastructure (PKI), you must configure it so
+its privately signed CA certificates are recognized across the cluster.
+
+You can leverage the Proxy API to configure your custom PKI. You must do this
```
danehans: s/You can leverage the Proxy API to configure your custom PKI/You can leverage the Proxy API to add cluster-wide trusted CA certificates.

networking/configuring-a-custom-pki.adoc
```diff
@@ -0,0 +1,56 @@
...
+either during installation *or* at runtime.
+
+* During *installation*, configure the cluster-wide proxy. You must define your
+privately signed CA certificates in the `install-config.yaml` file's
+`additionalTrustBundle` setting.
++
+The installation program generates a ConfigMap that is named `user-ca-bundle`
+that contains the additional CA certificates you defined. The Cluster Network
+Operator then creates a `trusted-ca-bundle` ConfigMap that merges these contents
```
danehans: s/these contents/these CA certificates/

networking/configuring-a-custom-pki.adoc
```diff
@@ -0,0 +1,56 @@
...
+* At *runtime*, enable the cluster-wide proxy. This involves creating a
+ConfigMap that contains the privately signed CA certificates that should be
+trusted by the cluster, and then modifying the proxy resource with the
+`trustedCA` referencing the privately signed certificates' ConfigMap.
+
+[NOTE]
+====
+The installer configuration's `additionalTrustBundle` field and the proxy
+resource's `trustedCA` field are used to manage the cluster-wide trust bundle;
+`additionalTrustBundle` is used at install time and `trustedCA` is used at run
```
danehans: s/and `trustedCA`/and `trustedCA` of `proxy`/

networking/configuring-a-custom-pki.adoc
```diff
@@ -0,0 +1,56 @@
...
+privately signed CA certificates in the `install-config.yaml` file's
+`additionalTrustBundle` setting.
++
+The installation program generates a ConfigMap that is named `user-ca-bundle`
+that contains the additional CA certificates you defined. The Cluster Network
+Operator then creates a `trusted-ca-bundle` ConfigMap that merges these contents
+with the {op-system-first} trust bundle; this ConfigMap is referenced in the
+Proxy object's `trustedCA` field.
+
+* At *runtime*, enable the cluster-wide proxy. This involves creating a
```
danehans: Regarding "enable the cluster-wide proxy", 4.2 and greater clusters will always have a proxy object named "cluster". If proxy is not enabled at install time, the object will have a nil spec and status. A user is not necessarily enabling proxy. Instead, the `trustedCA` field of proxy "cluster" is used to add additional cluster-wide trusted CA's.

modules/about-certificate-management.adoc
codyhoag: Good point; I'll leave the list out due to maintainability issues. Just to confirm, the proxy API is only valid for 4.2+, correct?

networking/configuring-a-custom-pki.adoc
```diff
@@ -0,0 +1,56 @@
...
+include::modules/nw-proxy-configure-object.adoc[leveloffset=+1]
+
+== Checkpoint
+
+You have defined your privately signed CA certificates in one of two ways:
+
+* In the `install-config.yaml` file's `additionalTrustBundle` field
+  (installation).
+
+* In the `user-ca-bundle.yaml` file's `ca-bundle.crt` field (runtime). Then you
```
danehans: It might be helpful to show examples of the install and runtime configs:

__Runtime Config__

1. configmap config:
```
apiVersion: v1
data:
  ca-bundle.crt: |
    -----BEGIN CERTIFICATE-----
  <SNIP>
    -----END CERTIFICATE-----
kind: ConfigMap
metadata:
  name: user-ca-bundle
  namespace: openshift-config
```

2. proxy config:
```
kind: Proxy
metadata:
  name: cluster
spec:
  trustedCA:
    name: user-ca-bundle
status:
```

__ Install time__
```
$ cat install-config.yaml 
<SNIP>
additionalTrustBundle: |
    -----BEGIN CERTIFICATE-----
<SNIP>
    -----END CERTIFICATE-----
```

codyhoag: The config examples are actually spelled out in the modules I included in the assembly:
- [Runtime](http://file.rdu.redhat.com/~choag/112019/ingress-cert/networking/configuring-a-custom-pki.html#nw-proxy-configure-object_configuring-a-custom-pki)
- [Install](http://file.rdu.redhat.com/~choag/112019/ingress-cert/networking/configuring-a-custom-pki.html#installation-configure-proxy_configuring-a-custom-pki)

I added the *Checkpoint* to specifically call out the fields directly related to the additional CA certs. I included the additional proxy information because that seems important to the context of the proxy API we're leveraging here. Does that make sense?

codyhoag: @danehans Or are you suggesting I call those examples out *again* in the checkpoint?

danehans: > I added the Checkpoint to specifically call out the fields directly related to the additional CA certs. I included the additional proxy information because that seems important to the context of the proxy API we're leveraging here. Does that make sense?

OK, I missed that.

codyhoag: no worries; thanks for reviewing!

codyhoag:
@lihongan can you review this content? Thanks!

lihongan:
LGTM, thank you!

codyhoag:
@bergerhoffer can you review to ensure I followed all the rules? :slightly_smiling_face: This is a reorg of #18004. Thanks!

networking/configuring-a-custom-pki.adoc
```diff
@@ -0,0 +1,57 @@
+[id="configuring-a-custom-pki"]
+= Configuring a custom PKI
+include::modules/common-attributes.adoc[]
+:context: configuring-a-custom-pki
+
+toc::[]
+
+Some platform components (i.e., web console) use Routes for communication and,
```
bergerhoffer: Avoid i.e. Suggest something like "Some platform components, such as the web console, use..."

networking/configuring-a-custom-pki.adoc
```diff
@@ -0,0 +1,57 @@
...
+therefore, must trust other components' certificates to interact with them. If
```
bergerhoffer: I don't think "therefore" is necessary. Suggest just "communication and must..."

networking/configuring-a-custom-pki.adoc
```diff
@@ -0,0 +1,57 @@
...
+you're using a custom public key infrastructure (PKI), you must configure it so
```
bergerhoffer: you're -> you are

networking/configuring-a-custom-pki.adoc
```diff
@@ -0,0 +1,57 @@
...
+its privately signed CA certificates are recognized across the cluster.
+
+You can leverage the Proxy API to add cluster-wide trusted CA certificates. You
+must do this either during installation *or* at runtime.
```
bergerhoffer: Not necessary to emphasize "or".

Also, asterisks are only used for GUI items. If you do need to emphasize, should be italics.

modules/about-certificate-management.adoc
```diff
@@ -0,0 +1,27 @@
...
+
+Operators request this injection by creating an empty ConfigMap with the
+following label:
+
+[source,yaml]
+----
+config.openshift.io/inject-trusted-cabundle="true"
+----
+
+The operator mounts this ConfigMap into the container's local trust store.
```
bergerhoffer: operator -> Operator

networking/configuring-a-custom-pki.adoc
```diff
@@ -0,0 +1,57 @@
...
+The installer configuration's `additionalTrustBundle` field and the proxy
+resource's `trustedCA` field are used to manage the cluster-wide trust bundle;
+`additionalTrustBundle` is used at install time and the proxy's `trustedCA` is
+used at runtime.
+
+The `trustedCA` field is a reference to a `ConfigMap` containing the custom
+certificate and key pair used by the cluster component.
+====
+
+include::modules/installation-configure-proxy.adoc[leveloffset=+1]
```
bergerhoffer: So, the way this is laid out now kind of makes it look like you need to do both of these procedures in the modules.

I'd consider linking to each of these from earlier in this assembly where you talk about configuring during installation or at runtime.

networking/configuring-a-custom-pki.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+include::modules/nw-proxy-configure-object.adoc[leveloffset=+1]
+
+== Checkpoint
```
bergerhoffer: I'd argue that this Checkpoint section isn't really adding any value, and I don't think we normally do something like this.

I have seen places that have something like "Verifying <x>" that verifies something you did earlier, but it usually has actual steps to check something.

But for your situation, I'd think this just wasn't necessary and that the lead-in from the assembly should be sufficient, if you link to these.

modules/about-certificate-management.adoc
```diff
@@ -0,0 +1,27 @@
+// Module included in the following assemblies:
+//
+// * networking/configuring-a-custom-pki.adoc
+
+[id="about-certificate-management_{context}"]
+= About certificate management
```
bergerhoffer: Just a question on whether this title is actually descriptive of what this section is discussing, or if there's something more specific it could say? It seems a little broad as is.

bergerhoffer (commented):
Just a few suggestions to consider!

networking/configuring-a-custom-pki.adoc
codyhoag: @bergerhoffer good to know. Does this also apply to `e.g.`, or is that one allowed?

networking/configuring-a-custom-pki.adoc
codyhoag: @bergerhoffer yeah, I was struggling with that. I was unsure whether it was best practice to link to a module from within its own assembly. I can't think of a reason why that wouldn't work, since we know the module will always be available from within the article.

I'll go ahead and provide links; I agree that would be much better.

networking/configuring-a-custom-pki.adoc
codyhoag: @bergerhoffer makes sense. My intention was to call out the specific places where we were adding private certs, since some parts of the included modules, while necessary, were more about the proxy set up and less about the cert specification.

I agree it's a bit out of place compared to what I've seen in other docs. I'll remove if you think the current setup is clear enough. :+1:

modules/about-certificate-management.adoc
codyhoag: @bergerhoffer what do you think about *Certificate injection using Operators* instead?

networking/configuring-a-custom-pki.adoc
bergerhoffer: Yep, also should avoid e.g. - IBM style guide says to use "for example" instead.

networking/configuring-a-custom-pki.adoc
bergerhoffer: Just so that I'm clear, the procedure modules you included for install/configure the proxy.. the user still needs to follow the entire procedures, correct?

Wanted to make sure that they don't only need to follow just a piece of the procedures (because then I'd suggest making your own steps instead of reusing these).

codyhoag: Yes, they need the entire procedure modules, which go through setting up the proxy. So although the cert specifications are the highlighted part of the process, the proxy must still be configured also.

modules/about-certificate-management.adoc
bergerhoffer: Possibly. It does seem informational so I don't mind the About part, though usually I"m used to seeing the concept modules like this more at the top of an assembly as opposed to at the end.

I'm not positive, but I'm wondering if it would make more sense coupled with the information you have at the beginning of the assembly? Just something to think about, I'm fine with however you think it's best presented.

PR #18245, opened 2019-11-21
osdocs-772 GCP private clusters

kalexand-rh:
https://jira.coreos.com/browse/OSDOCS-772

@patrickdillon,  will you PTAL at the module changes for a GCP private cluster? I'm planning on including the BYO VPC information in the assembly that will contain this change, since it should be identical.

### DIFF
--- a/installing/installing_gcp/installing-gcp-private.adoc
+++ b/installing/installing_gcp/installing-gcp-private.adoc
@@ -1,0 +1,53 @@
+[id="installing-gcp-private"]
+= Installing a private cluster on GCP
+include::modules/common-attributes.adoc[]
+:context: installing-gcp-private
+
+toc::[]
+
+In {product-title} version {product-version}, you can install a private cluster into an existing VPC on Google Cloud Platform (GCP). The installation program provisions the rest of the required infrastructure, which you can further customize. To customize the installation, you modify
+parameters in the `install-config.yaml` file before you install the cluster.
+
+.Prerequisites
+
+* Review details about the
+xref:../../architecture/architecture-installation.adoc#architecture-installation[{product-title} installation and update]
+processes.
+* xref:../../installing/installing_gcp/installing-gcp-account.adoc#installing-gcp-account[Configure a GCP account]
+to host the cluster.
+* If you use a firewall, you must
+xref:../../installing/install_config/configuring-firewall.adoc#configuring-firewall[configure it to allow the sites] that your cluster requires access to.
+
+include::modules/private-clusters-default.adoc[leveloffset=+1]
+
+include::modules/private-clusters-about-gcp.adoc[leveloffset=+2]
+
+include::modules/cluster-entitlements.adoc[leveloffset=+1]
+
+//uncomment after GCP VPC merges
+//include::modules/installation-about-custom-gcp-vpc.adoc[leveloffset=+1]
+
+include::modules/ssh-agent-using.adoc[leveloffset=+1]
+
+include::modules/installation-obtaining-installer.adoc[leveloffset=+1]
+
+include::modules/installation-initializing.adoc[leveloffset=+1]
+
+include::modules/installation-configuration-parameters.adoc[leveloffset=+2]
+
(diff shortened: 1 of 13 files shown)


modules/private-clusters-about-gcp.adoc
```diff
@@ -0,0 +1,40 @@
+// Module included in the following assemblies:
+//
+// * installing/
+
+[id="private-clusters-about-gcp_{context}"]
+= Private clusters in GCP
+
+To create a private cluster on Google Cloud Platform (GCP), you must provide an existing private VPC and subnets to host the cluster. The installation program must also be able to resolve the DNS records that the cluster requires. The installation program configures the Ingress Operator and API server for only internal traffic.
```
patrickdillon: > To create a private cluster on Google Cloud Platform (GCP), you must provide an existing private VPC and subnets to host the cluster.

I think the installer may create private networks if none are provided. Users would then need to vpn to the installer-created network.

patrickdillon:
cc @jstuever

modules/private-clusters-about-gcp.adoc
jstuever: The network and subnets must be provided by the user as there is no way for the installer to automatically access the cluster api in a private network it just provisioned (and therefore wait-for bootstrap-complete will always fail).

modules/private-clusters-about-gcp.adoc
```diff
@@ -0,0 +1,40 @@
...
+
+The following items are not required or created when you install a private cluster:
+
+* Public subnets
+* Public network load balancers, which support public ingress
+* A public DNS zone that matches the `baseDomain` for the cluster
+
+The installation program does use the `baseDomain` that you specify to create a private DNS zone and the required records for the cluster. The cluster is configured so that the Operators do not create public records for the cluster and all cluster machines are placed in the private subnets that you specify.
+
+Because it is not possible to limit access to external load balancers based on source tags, the private cluster uses internal load balancers to allow access to internal instances instead of network load balancers.
```
jstuever: I believe we now use internal load balancers in GCP for both private and public clusters. The difference is that only public clusters create the external network load balancers.

kalexand-rh:
@wmengRH and @jiajliu, after you take a look at the shared VPC docs, will you PTAL here? There's some information that's introduced on the shared VPC docs that will be included in this assembly, but all of the process information should be here.

jiajliu:
@kalexand-rh Sorry for missing your msg. May @gpei can help on it.

installing/installing_gcp/installing-gcp-private.adoc
```diff
@@ -0,0 +1,53 @@
...
+include::modules/common-attributes.adoc[]
+:context: installing-gcp-private
+
+toc::[]
+
+In {product-title} version {product-version}, you can install a private cluster into an existing VPC on Google Cloud Platform (GCP). The installation program provisions the rest of the required infrastructure, which you can further customize. To customize the installation, you modify
+parameters in the `install-config.yaml` file before you install the cluster.
+
+.Prerequisites
+
```
shellyyang1989: If your workstation on which installation program runs is outside of  VPC network, you cannot validate the cluster status cause the cluster has only internal access. It's better to create a bastion host within the VPC network and run installation program on it.

kalexand-rh: @shellyyang1989, @jianlinliu, do you have information about the bastion host that's required?

jianlinliu: I think some PR about bastion host already in progress:
https://github.com/openshift/openshift-docs/pull/16989/files

kalexand-rh: I had a chat with Jeremiah, and he thinks that the bastion host isn't necessarily the way to go. We agreed that it's going to be better to talk about the access that he host that you install the private cluster on needs instead of proscribing a bastion, at least for now. Jeremiah used a bastion but thinks that a VPN or other existing access will be more likely.

kalexand-rh:

sheriff-rh (approved):
Looks great! Keep it up!

PR #18261, opened 2019-11-22
osdocs-774 private Azure

kalexand-rh:
https://jira.coreos.com/browse/OSDOCS-774

@jhixson74, @abhinavdahiya, will you  PTAL at the module changes for a Azure private cluster? I'm planning on including the BYO VPC information in the assembly that will contain this change, since it should be identical.

### DIFF
--- a/modules/private-clusters-about-azure.adoc
+++ b/modules/private-clusters-about-azure.adoc
@@ -1,0 +1,35 @@
+// Module included in the following assemblies:
+//
+// * installing/
+
+[id="private-clusters-about-azure_{context}"]
+= Private clusters in Azure
+
+To create a private cluster on Microsoft Azure, you must provide an existing private VNet and subnets to host the cluster. The installation program must also be able to resolve the DNS records that the cluster requires. The installation program configures the Ingress Operator and API server for only internal traffic.
+
+Depending how your network connects to the private VNET, you might need to use a DNS forwarder in order to resolve the cluster's private DNS records. The cluster's machines use `168.63.129.16` internally for DNS resolution. For more information, see link:https://docs.microsoft.com/en-us/azure/dns/private-dns-overview[What is Azure Private DNS?] and link:https://docs.microsoft.com/en-us/azure/virtual-network/what-is-ip-address-168-63-129-16[What is IP address 168.63.129.16?] in the Azure documentation.
+
+The cluster still requires access to Internet to access the Azure APIs.
+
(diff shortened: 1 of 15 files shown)


modules/private-clusters-about-azure.adoc
```diff
@@ -0,0 +1,34 @@
...
+To create a private cluster on Microsoft Azure, you must provide an existing private VNet and subnets to host the cluster. The installation program must also be able to resolve the DNS records that the cluster requires. The installation program configures the Ingress Operator and API server for only internal traffic.
+
+The cluster still requires access to Internet to access the Azure APIs.
+
+The following items are not required or created when you install a private cluster:
+
+* A `BaseDomainResourceGroup`, since the cluster does not create public records
+* Public IP addresses
+* Public DNS records
+* Public load balancers
```
jhixson74: This isn't exactly true. Ideally we want this, but because of limitations in Azure, we do create a public load balancer with a dummy rule. Azure does not allow outbound traffic without this, unfortunately.

kalexand-rh: Can you tell me a little more about the dummy rule?

modules/private-clusters-about-azure.adoc
```diff
@@ -0,0 +1,34 @@
+// Module included in the following assemblies:
+//
+// * installing/
+
+[id="private-clusters-about-azure_{context}"]
+= Private clusters in Azure
+
+To create a private cluster on Microsoft Azure, you must provide an existing private VNet and subnets to host the cluster. The installation program must also be able to resolve the DNS records that the cluster requires. The installation program configures the Ingress Operator and API server for only internal traffic.
```
e-tienne: Depending how your network connects to the private VNET, you might need to use a DNS forwarder in order to resolve the cluster's private DNS records. The cluster's machines use 168.63.129.16 internally for DNS resolution.

https://docs.microsoft.com/en-us/azure/dns/private-dns-overview
https://docs.microsoft.com/en-us/azure/virtual-network/what-is-ip-address-168-63-129-16

kalexand-rh:
@e-tienne, I've pushed an update to provide that data. What else do you see?

kalexand-rh:
@e-tienne, will you please take another look?

e-tienne:
LGTM

kalexand-rh:

sheriff-rh (approved):
Looks great!

PR #13877, opened 2019-02-27
 Edits to Logging 4.0 topics per team review 

mburke5678:
First round of edits from review. Smaller, easier changes.
https://docs.google.com/document/d/1-E1z-Z_CvIC7u4ggQFfaH-9aF_p7lIGsdOLD7Zynj9g/edit?ts=5c759927#heading=h.c7x2dmk5jv1x

Also: https://bugzilla.redhat.com/show_bug.cgi?id=1683040

PR #16759, opened 2019-09-19
Change cluster logging file name prefix from `efk-logging` to `cluster-logging`

mburke5678:
Changing the prefix on the file names of the cluster logging docs from `efk-logging` to `cluster-logging` to remove the EFK concept.
Also changed the associated context and ID strings.

PR #18267, opened 2019-11-24
modules/*logging*: Consistently 'oc edit ClusterLogging instance' before YAML

wking:
The edit command is not YAML itself, so consistently place it in its own block before the YAML block to get proper syntax highlighting (instead of highlighting the shell command as invalid YAML).  The command first appeared in the YAML blocks in 98e58b8c8a (#13877).  There was some subequent cleanup in aa33f74bc7 (#15682), with the command being moved out of YAML blocks in:

* `modules/efk-logging-elasticsearch-limits.adoc`
* `modules/efk-logging-fluentd-collector.adoc`
* `modules/efk-logging-kibana-limits.adoc`

aa33f74bc7 also copied out the `edit` command, but forgot to remove it from the YAML blocks in:

* `modules/efk-logging-fluentd-limits.adoc`
* `modules/efk-logging-kibana-scaling.adoc`
* `modules/efk-logging-management-state-changing.adoc`

Hopefully after this commit we're consistent in all of those locations ;).

CC @mburke5678

### DIFF
--- a/modules/cluster-logging-collector-limits.adoc
+++ b/modules/cluster-logging-collector-limits.adoc
@@ -17,8 +17,6 @@ $ oc edit ClusterLogging instance
 +
 [source,yaml]
 ----
-$ oc edit ClusterLogging instance
-
 apiVersion: "logging.openshift.io/v1"
 kind: "ClusterLogging"
 metadata:
@@ -39,29 +37,3 @@ spec:
             memory: 1Gi
 ----
 <1> Specify the CPU and memory limits and requests as needed. The values shown are the default values.
-
-////
-[source,yaml]
-----
-$ oc edit ClusterLogging instance
-
-apiVersion: "logging.openshift.io/v1"
-kind: "ClusterLogging"
-metadata:
-  name: "instance"
-
-....
-
-spec:
-  collection:
-    logs:
-      rsyslog:
-        resources:
-          limits: <1>
-            memory: 358Mi
-          requests:
-            cpu: 100m
-            memory: 358Mi
-----
-<1> Specify the CPU and memory limits and requests as needed. The values shown are the default values.
-////
--- a/modules/cluster-logging-configuring-node-selector.adoc
+++ b/modules/cluster-logging-configuring-node-selector.adoc
@@ -9,12 +9,14 @@ Each component specification allows the component to target a specific node.
 
 .Procedure
 
-Edit the Cluster Logging Custom Resource (CR) in the `openshift-logging` project:
-
-[source,yaml]
+. Edit the Cluster Logging Custom Resource (CR) in the `openshift-logging` project:
++
 ----
 $ oc edit ClusterLogging instance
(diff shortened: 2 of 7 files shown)


wking:
Probably needs to get picked back all the way to 4.1, but I haven't checked to see if it applies cleanly.

modules/cluster-logging-collector-collector.adoc
```diff
@@ -57,4 +57,3 @@ nodeSpec:
       type: "fluentd" <1>
 ----
 <1> Set the log collector to `fluentd`.
-////
```
mburke5678: Are you removing this `////` on purpose?

This file is not currently used. While doing clean-up, would you mind removing the `logging/cluster-logging-collector.adoc` from _Modules included in_?

mburke5678:
@wking Would you mind fixing the edit command in https://docs.openshift.com/container-platform/4.2/logging/config/cluster-logging-management.html#cluster-logging-management-state-changing-es_cluster-logging-curator while you are in there?

mburke5678:
@wking Thanks for catching and fixing these!

modules/cluster-logging-collector-collector.adoc
wking: > Are you removing this //// on purpose?

Yeah, no need to keep comments around in adoc ;).  If the defaults need to change at some point, that should be tracked in a bug or RFE or whatever.

> Also, this file is not currently used.

Done with #18284.  Looks like there's some other unused logging content as well, e.g. nobody consuming `modules/cluster-logging-collector-undefined.adoc`, but I've left that alone because I don't really understand how these adoc files are wired together.

wking:
> Would you mind fixing the edit command in https://docs.openshift.com/container-platform/4.2/logging/config/cluster-logging-management.html#cluster-logging-management-state-changing-es_cluster-logging-curator while you are in there?

Done with b2731a5c4 -> e3774c5c5, although since that's fixing an issue from #16759, it may make it harder to pick this back onto earlier releases.

modules/cluster-logging-collector-collector.adoc
mburke5678: I hid that whole chunk as at the time the option to switch the logging collector to Rsyslog was still in consideration. It appears that is not the case now. If we remove the `////` at the end, all the material after Line 13 should come out also.

wking:
Example of the broken shell-is-not-valid-YAML syntax highlighting this PR should fix:

Screenshot from 2019-11-25 11-55-46

from [here][1].

[1]: https://docs.openshift.com/container-platform/4.2/logging/config/cluster-logging-management.html#cluster-logging-management-state-changing-es_cluster-logging-curator

wking:
Rebased onto master with e3774c5c5 -> 0d4f9d87d now that #18284 has landed.

mburke5678:
@wking LGTM

PR #18283, opened 2019-11-25
Fix of azure load balancer.

Makdaam:
Based on https://github.com/Azure/OpenShift/issues/108


kalexand-rh:
@vikram-redhat, do you have a suggestion for QE?

Makdaam:
@kalexand-rh The annotation here is a boolean, as described in the documentation. https://docs.microsoft.com/en-us/azure/aks/load-balancer-standard
The mistake is an copy-paste error from OSD where it refers to a similar, but different annotation in Amazon Cloud, which accepts a CIDR address, and not "true".

PR #18311, opened 2019-11-26
clarifying RHEL node entitlements

kalexand-rh:
@vikram-redhat, PTAL?


vikram-redhat:
@reestr FYI. Ping me if you want to know more about why we removed this.

PR #18316, opened 2019-11-26
Add FlexVolume assembly to 4.x docs

bobfuru:
[BZ 1702576](https://bugzilla.redhat.com/show_bug.cgi?id=1702576)
@gnufied PTAL

### DIFF
--- a/modules/persistent-storage-flexvolume-drivers.adoc
+++ b/modules/persistent-storage-flexvolume-drivers.adoc
@@ -1,0 +1,13 @@
+// Module included in the following assemblies:
+//
+// storage/persistent-storage/persistent-storage-flexvolume.adoc
+
+[id="flexvolume-drivers_{context}"]
+= About FlexVolume drivers
+
+A FlexVolume driver is an executable file that resides in a well-defined directory on all nodes in the cluster. {product-title} calls the FlexVolume driver whenever it needs to mount or unmount a volume represented by a `PersistentVolume` with `flexVolume` as the source.
+
+[IMPORTANT]
+====
+Attach and detach operations are not supported in {product-title} for FlexVolume.
+====
--- a/modules/persistent-storage-flexvolume-installing.adoc
+++ b/modules/persistent-storage-flexvolume-installing.adoc
@@ -1,0 +1,65 @@
+// Module included in the following assemblies:
+//
+// storage/persistent-storage/persistent-storage-flexvolume.adoc
+
+[id="flexvolume-installing_{context}"]
+
+= Installing FlexVolume drivers
+
+FlexVolume drivers that are used to extend {product-title} are executed only on the node. To implement FlexVolumes, a list of operations to call and the installation path are all that is required.
+
+.Prerequisites
+
+* FlexVolume drivers must implement these operations:
+
+`init`::
+Initializes the driver. It is called during initialization of all nodes.
+
+** Arguments: none
+** Executed on: node
+** Expected output: default JSON
+
+`mount`::
+Mounts a volume to directory. This can include anything that is necessary to mount the volume, including finding the device and then mounting the device.
+
+** Arguments: `<mount-dir>` `<json>`
+** Executed on: node
+** Expected output: default JSON
+
+`unmount`::
+Unmounts a volume from a directory. This can include anything that is necessary to clean up the volume after unmounting.
+
+** Arguments: `<mount-dir>`
+** Executed on: node
+** Expected output: default JSON
+
+`mountdevice`::
+Mounts a volume's device to a directory where individual Pods can then bind mount.
+
+This call-out does not pass "secrets" specified in the FlexVolume spec. If your driver requires secrets, do not implement this call-out.
+
+** Arguments: `<mount-dir>` `<json>`
+** Executed on: node
+** Expected output: default JSON
+
+`unmountdevice`::
+Unmounts a volume's device from a directory.
+
+** Arguments: `<mount-dir>`
+** Executed on: node
+** Expected output: default JSON
+
+* All other operations should return JSON with `{"status": "Not supported"}` and exit code `1`.
+
+.Procedure
+
+To install the FlexVolume driver:
+
+. Ensure that the executable file exists on all nodes in the cluster.
+
+. Place the executable file at the volume plug-in path:
+*_/etc/kubernetes/kubelet-plugins/volume/exec/<vendor>~<driver>/<driver>_*.
+
+For example, to install the FlexVolume driver for the storage `foo`, place the
+executable file at:
+*_/etc/kubernetes/kubelet-plugins/volume/exec/openshift.com~foo/foo_*.
--- a/modules/storage-expanding-flexvolume.adoc
+++ b/modules/storage-expanding-flexvolume.adoc
@@ -1,0 +1,34 @@
+// Module included in the following assemblies
+//
+// * storage/expanding-persistent-volumes.adoc
+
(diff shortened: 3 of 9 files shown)


modules/persistent-storage-flexvolume-installing.adoc
```diff
@@ -0,0 +1,59 @@
...
+Unmounts a volume from a directory. This can include anything that is
+necessary to clean up the volume after unmounting, such as detaching the volume
+from the node.
+
+** Arguments: `<mount-dir>`
+** Executed on: node
+** Expected output: default JSON
+
+* All other operations should return JSON with `{"status": "Not supported"}` and exit code `1`.
+
```
gnufied: Flexvolume's api documentation is kinda incomplete in Openshift. This documentation does not document `mountdevice` and `unmountdevice` api calls. In addition we do not document resizing support for flexvolume plugins. 

A more complete documentation is available on - https://github.com/kubernetes/community/blob/master/contributors/devel/sig-storage/flexvolume.md and https://github.com/kubernetes/community/blob/master/contributors/design-proposals/storage/grow-flexvolume-size.md

modules/persistent-storage-flexvolume-installing.adoc
```diff
@@ -0,0 +1,59 @@
...
+. Ensure that the executable file exists on all masters and nodes in the cluster.
+
+. Place the executable file at the volume plug-in path:
+*_/etc/kubernetes/kubelet-plugins/volume/exec/<vendor>~<driver>/<driver>_*.
+
+For example, to install the FlexVolume driver for the storage `foo`, place the
+executable file at:
+*_/etc/kubernetes/kubelet-plugins/volume/exec/openshift.com~foo/foo_*.
+
+To install the FlexVolume driver on Atomic hosts:
```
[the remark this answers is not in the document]
bobfuru: Removed procedure for installing on Atomic hosts

modules/persistent-storage-flexvolume-installing.adoc
bobfuru: Thanks for the info and links. 

- I didn't include `mountdevice` and `unmountdevice` before because they were documented in "FlexVolume drivers with master-initiated attach/detach" section and I was not including attach/detach info. I've added those calls in the doc assembly. 

- I'm currently working on getting FlexVolume expansion content added (WIP).

modules/persistent-storage-flexvolume-installing.adoc
```diff
@@ -0,0 +1,65 @@
...
+* All other operations should return JSON with `{"status": "Not supported"}` and exit code `1`.
+
+.Procedure
+
+To install the FlexVolume driver:
+
+. Ensure that the executable file exists on all masters and nodes in the cluster.
+
+. Place the executable file at the volume plug-in path:
+*_/etc/kubernetes/kubelet-plugins/volume/exec/<vendor>~<driver>/<driver>_*.
```
bobfuru: @gnufied  I believe these install paths were updated since this: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-storage/flexvolume.md#prerequisites
Can you confirm that this install path and the one below [52] are correct?

modules/persistent-storage-flexvolume-installing.adoc
bobfuru: @gnufied I've finished adding FlexVol content from the upstream links you shared. Tried to place content where it made most sense (e.g., FlexVol expansion under the "Expanding PVs" section). 
PTAL - thanks!

bobfuru:
@liangxia PTAL (FYI, this PR removes info about attach/detach, which is not supported in OpenShift for FlexVolume)

liangxia:
Sorry, busy on bugs these days. And this requires fully testing on OCP 4.2 and 4.3 which cost time. Plan to work on it later this week.

modules/storage-expanding-flexvolume.adoc
```diff
@@ -0,0 +1,35 @@
...
+
+Similar to other volume types, FlexVolume volumes can also be expanded when in use by a Pod.
+
+.Prerequisites
+
+* The underlying volume driver supports resize.
+* The driver is set with the `RequiresFSResize` capability to `true`.
+* Dynamic provisioning is used.
+* The controlling StorageClass has `allowVolumeExpansion` set to `true`.
+* The `PersistentVolumeClaimResize` admission plugin is enabled (optional).
```
gnufied: This is enabled by default in Openshift and hence I would drop the line about admission plugin from documentation.

modules/persistent-storage-flexvolume-drivers.adoc
```diff
@@ -0,0 +1,13 @@
+// Module included in the following assemblies:
+//
+// storage/persistent-storage/persistent-storage-flexvolume.adoc
+
+[id="flexvolume-drivers_{context}"]
+= About FlexVolume drivers
+
+A FlexVolume driver is an executable file that resides in a well-defined directory on all machines in the cluster, including both masters and nodes. {product-title} calls the FlexVolume driver whenever it needs to mount or unmount a volume represented by a `PersistentVolume` with `flexVolume` as the source.
+
```
gnufied: Openshift does not support installing flex driver on master nodes.

modules/storage-expanding-flexvolume.adoc
```diff
@@ -0,0 +1,35 @@
...
+
+.Procedure
+
+* To use resizing in the FlexVolume plugin, you must implement the `ExpandableVolumePlugin` interface using these methods:
+
+`ExpandVolumeDevice`::
```
gnufied: Since Openshift does not support installation of flexvolume plugins on master nodes - this documentation is not accurate.

```
To use resizing in the FlexVolume plugin, you must implement the `NodeExpandableVolumePlugin` interface using these methods:

RequiresFSResize() bool::
Which indicates that the volume supports expansion on the node.

`NodeExpand`:
This function will internally call `expandfs` on flex volume
```

Since Openshift doed not support installation of flexvolume plugins on master nodes, Openshift does not support control-plane expansion of flexvolume.

modules/persistent-storage-flexvolume-drivers.adoc
bobfuru: Okay, changing it to read, "...n a well-defined directory on all nodes in the cluster."

modules/storage-expanding-flexvolume.adoc
bobfuru: @gnufied If I understand correctly, when you say "OpenShift does not support control-plane expansion of flexvolume" means it does not support the text only in the first method (`ExpandVolumeDevice`). Or do you mean the other two methods (`RequiresFSResize` and `ExpandFS`) are also not supported?

modules/persistent-storage-flexvolume-installing.adoc
```diff
@@ -0,0 +1,65 @@
...
+** Executed on: node
+** Expected output: default JSON
+
+* All other operations should return JSON with `{"status": "Not supported"}` and exit code `1`.
+
+.Procedure
+
+To install the FlexVolume driver:
+
+. Ensure that the executable file exists on all masters and nodes in the cluster.
```
liangxia: `on all masters and nodes` ?
We stated earlier that `FlexVolume drivers that are used to extend {product-title} are executed only on the node.`

bobfuru: Removed ref to `masters` here. 
@liangxia With this change, can I please get a LGTM from you?

PR #18317, opened 2019-11-26
Add Virtual Function (VF) flags for SR-IOV Network Operator

jboxman:
- https://jira.coreos.com/browse/OSDOCS-828

Original PR: https://github.com/pliurh/openshift-docs/pull/4

### DIFF
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -343,7 +343,7 @@ Topics:
     File: configuring-ipvlan
   - Name: Configuring a host-device network
     File: configuring-host-device
-  - Name: Configuring SR-IOV
+  - Name: Configuring an SR-IOV network
     File: configuring-sr-iov
   - Name: Editing an additional network
     File: edit-additional-network
--- a/modules/nw-sriov-configuring-device.adoc
+++ b/modules/nw-sriov-configuring-device.adoc
@@ -33,7 +33,7 @@ apiVersion: sriovnetwork.openshift.io/v1
 kind: SriovNetworkNodePolicy
 metadata:
   name: <name> <1>
-  namespace: sriov-network-operator <2>
+  namespace: openshift-sriov-network-operator <2>
 spec:
   resourceName: <sriov_resource_name> <3>
   nodeSelector:
@@ -55,18 +55,18 @@ spec:
 <4> Specify the node selector to select which node to be configured. User can choose to label the nodes manually or with tools like Kubernetes Node Feature Discovery.
 Only SR-IOV network devices on selected nodes will be configured. The SR-IOV
 CNI plug-in and device plug-in will be only deployed on selected nodes.
-<5> Specify an integer value between `0` and `99`. A larger number gets lower priority, so a priority of `99` is lower than a priority of `10`.
-<6> Specify a value for the maximum transmission unit (MTU) of the virtual function. The value for MTU must be in the range from `1` to `9000`. If you do not need to specify the MTU, specify a value of `''`. 
+<5> Optional. Specify an integer value between `0` and `99`. A larger number gets lower priority, so a priority of `99` is lower than a priority of `10`. The default value is `99`.
+<6> Optional. Specify a value for the maximum transmission unit (MTU) of the virtual function. The maximum MTU value can vary for different NIC models.
 <7> Specify the number of the virtual functions (VF) to create for the SR-IOV physical network device. For an Intel Network Interface Card (NIC), the number of VFs cannot be larger than the total VFs supported by the device. For a Mellanox NIC, the number of VFs cannot be larger than `128`.
 <8> The `nicSelector` mapping selects the Ethernet device for the Operator to configure. You do not need to specify values for all the parameters. It is recommended to identify the Ethernet adapter with enough precision to minimize the possibility of selecting an Ethernet device unintentionally.
 If you specify `rootDevices`, you must also specify a value for `vendor`, `deviceID`, or `pfName`.
 If you specify both `pfName` and `rootDevices` at the same time, ensure that they point to an identical device.
-<9> Specify the vendor hex code of the SR-IOV network device. The only allowed values are either `8086` or `15b3`.
-<10> Specify the device hex code of SR-IOV network device. The only allowed values are `158b`, `1015`, `1017`.
-<11> The parameter accepts an array of one or more physical function (PF) names for the Ethernet device.
+<9> Optional. Specify the vendor hex code of the SR-IOV network device. The only allowed values are either `8086` or `15b3`.
+<10> Optional. Specify the device hex code of SR-IOV network device. The only allowed values are `158b`, `1015`, `1017`.
+<11> Optional. The parameter accepts an array of one or more physical function (PF) names for the Ethernet device.
 <12> The parameter accepts an array of one or more PCI bus addresses for the physical function of the Ethernet device. Provide the address in the following format: `0000:02:00.1`.
-<13> Specify the driver type for the virtual functions. You can specify one of the following values: `netdevice` or `vfio-pci`. The default value is `netdevice`.
-<14> Specify whether to enable RDMA mode. The default value is `false`. Only RDMA over Converged Ethernet (RoCE) mode is supported on Mellanox Ethernet adapters.
+<13> Optional. Specify the driver type for the virtual functions. You can specify one of the following values: `netdevice` or `vfio-pci`. The default value is `netdevice`.
+<14> Optional. Specify whether to enable RDMA mode. The default value is `false`. Only RDMA over Converged Ethernet (RoCE) mode is supported on Mellanox Ethernet adapters.
 +
 [NOTE]
 =====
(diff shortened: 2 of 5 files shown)


jboxman:
Need to take care, because of this: https://yaml.org/type/bool.html

jboxman:
@pliurh, this is an edited version of https://github.com/pliurh/openshift-docs/pull/4 for the docs.

_topic_map.yml
```diff
@@ -343,7 +343,7 @@ Topics:
     File: configuring-ipvlan
   - Name: Configuring a host-device network
     File: configuring-host-device
-  - Name: Configuring SR-IOV
+  - Name: Configuring an SR-IOV network
```
zshi-redhat: nit: shall this be `a SR-IOV`?

modules/nw-sriov-configuring-device.adoc
```diff
@@ -55,18 +55,18 @@ spec:
...
-<11> The parameter accepts an array of one or more physical function (PF) names for the Ethernet device.
+<9> Optional. Specify the vendor hex code of the SR-IOV network device. The only allowed values are either `8086` or `15b3`.
+<10> Optional. Specify the device hex code of SR-IOV network device. The only allowed values are `158b`, `1015`, `1017`.
+<11> Optional. The parameter accepts an array of one or more physical function (PF) names for the Ethernet device.
 <12> The parameter accepts an array of one or more PCI bus addresses for the physical function of the Ethernet device. Provide the address in the following format: `0000:02:00.1`.
-<13> Specify the driver type for the virtual functions. You can specify one of the following values: `netdevice` or `vfio-pci`. The default value is `netdevice`.
-<14> Specify whether to enable RDMA mode. The default value is `false`. Only RDMA over Converged Ethernet (RoCE) mode is supported on Mellanox Ethernet adapters.
+<13> Optional. Specify the driver type for the virtual functions. You can specify one of the following values: `netdevice` or `vfio-pci`. The default value is `netdevice`.
```
zshi-redhat: Shall we add a note here: `For mellanox card to work in dpdk mode, use netdevice driver type`

modules/nw-sriov-network-attachment.adoc
```diff
@@ -28,20 +34,41 @@ apiVersion: sriovnetwork.openshift.io/v1
...
+<3> Optional. Replace `<target_namespace>` with the namespace where the NetworkAttachmentDefinition CR will be created. The default value is `openshift-sriov-network-operator`.
+<4> Optional. Specify a configuration object for the ipam CNI plug-in as a YAML block scalar. The plug-in manages IP address assignment for the attachment definition.
+<5> Optional. Replace `<vlan>` with a Virtual LAN (VLAN) ID for the additional network. The integer value must be from `0` to `4095`. The default value is `0`.
 <6> Replace `<sriov_resource_name>` with the value for the `.spec.resourceName` parameter from the SriovNetworkNodePolicy CR that defines the SR-IOV hardware for this additional network.
+<7> Optional. Replace `<link_state>` with the link state of Virtual Function (VF). Allowed value are `enable`, `disable` and `auto`.
+<8> Optional. Replace `<max_tx_rate>` with a maximum transmission rate, in Mbps, for the VF.
+<9> Optional. Replace `<min_tx_rate>` with a minimum transmission rate, in Mbps, for the VF. This value should always be less than or equal to Maximum transmission rate.
++
+NOTE: Intel NICs do not support the `minTxRate` parameter.
```
zshi-redhat: Might be good to link the note to corresponding [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1772847) so user knows the consequence (this will lead to config daemon failure ).

modules/nw-sriov-network-attachment.adoc
```diff
@@ -28,20 +34,41 @@ apiVersion: sriovnetwork.openshift.io/v1
...
     ...
   vlan: <vlan> <5>
   resourceName: <sriov_resource_name> <6>
+  linkState: <link_state> <7>
+  maxTxRate: <max_tx_rate> <8>
+  minTxRate: <mix_rx_rate> <9>
+  vlanQoS: <vlan_qos> <10>
+  spoofChk: "<spoof_check>" <11>
+  trust: "<trust_vf>" <12>
+  capabilities: <capabilities> <13>
```
zshi-redhat: Shall the format of `capabilities` be consistent with `ipam` ?

modules/nw-sriov-network-attachment.adoc
```diff
@@ -28,20 +34,41 @@ apiVersion: sriovnetwork.openshift.io/v1
...
++
+NOTE: Intel NICs do not support the `minTxRate` parameter.
+<10> Optional. Replace `<vlan_qos>` with an IEEE 802.1p priority level for the VF. The default value is `0`.
+<11> Optional. Replace `<spoof_check>` with the spoof check mode of the VF. The allowed values are the strings `"on"` and `"off"`.
++
+IMPORTANT: You must enclose the value you specify in quotes or the CR will be rejected by the SR-IOV Network Operator.
+<12> Optional. Replace `<trust_vf>` with the trust mode of the VF. The allowed values are the strings `"on"` and `"off"`.
++
+IMPORTANT: You must enclose the value you specify in quotes or the CR will be rejected by the SR-IOV Network Operator.
+<13> Optional. Replace `<capabilities>` with the capabilities to configure for this network. You can specify `"{ "ips": true }"` to enable IP address support or `"{ "mac": true }"` to enable MAC address support.
```
jboxman: @pliurh, is it correct to say `capabilities` enables IP address or MAC address support? Can it be both? And if this is enabled, what does that actually mean? Why would a user want to enable one of these capabilities? What is the behavior if neither is enabled?

Thanks!

modules/nw-sriov-network-attachment.adoc
```diff
@@ -28,20 +34,41 @@ apiVersion: sriovnetwork.openshift.io/v1
...
+  namespace: openshift-sriov-network-operator <2>
 spec:
   networkNamespace: <target_namespace> <3>
   ipam: |- <4>
     ...
   vlan: <vlan> <5>
   resourceName: <sriov_resource_name> <6>
+  linkState: <link_state> <7>
+  maxTxRate: <max_tx_rate> <8>
+  minTxRate: <mix_rx_rate> <9>
```
vpickard: This should be <min_tx_rate>

modules/nw-sriov-network-attachment.adoc
vpickard: Agree with Zenghui's comment about linking to existing but.

vpickard (changes requested):

modules/nw-sriov-network-attachment.adoc
pliurh: You can have both. With MAC or IP capabilities`enabled` means you can specify the IP and/or MAC address in the pod manifests. If neither is enabled, then CNI will give you an automatic generated IP and MAC address. It's a new feature of Multus. There shall be a separated document to descript this in detail. I think it is better to have a link to that doc. @dougbtv Could you share the PR for this feature?

jboxman: @pliurh, there's a tooling limitation that currently prevents us from having any `xref` in modules. So we can't actually link to that content directly.

If it's modularized, we could include it in the SR-IOV Network Operator assembly.

zshi-redhat: > @pliurh, is it correct to say `capabilities` enables IP address or MAC address support? 

`capabilities` enables static assignment of IP and MAC address. 
For static IP support, static ipam should be used (e.g. use static ipam type in SR-IOV net-attach-def).
For static MAC support in SR-IOV case, SR-IOV CNI needs to have the ability to accept the given MAC address and set it to actual device.

Can it be both? 

Yes, it can have both MAC and IP capabilities defined in `capabilities` field.

And if this is enabled, what does that actually mean? 

If it's enabled, Multus will parse it and call libcni to replace the capabilites with MAC and IP addresses defined in Pod Annotation(by user), then pass those addresses to delegated CNI via CNI rumtimeConfig. SR-IOV CNI or static IPAM plugin should(already) support CNI rumtimeConfig.

Why would a user want to enable one of these capabilities? 

User wants to configure static IP or/and MAC addresses.

What is the behavior if neither is enabled?

If neither is enabled, then static MAC or/and IP configured in Pod annotation will not be applied to pod interface.
(This is not 100% true, as we support legacy CNI args configuration in SRIOV and IPAM plugin, which will appy static MAC and IP to interface if they are passed down as CNI args, however, this mode will be deprecated in the future releases.)

> 
> Thanks!

jboxman:
Hi @lamek,

@vikram-redhat suggested I hand this off to you to wrap up so that I can finalize a demo of the API docs generator this week.

Thanks!

lamek:
@jboxman I got it.

lamek (commented):
I CPed this to a new PR here:
https://github.com/openshift/openshift-docs/pull/19000

lamek:
@jboxman per my comment above I'm going to close out this PR.

PR #18326, opened 2019-11-27
fixed path

apinnick:
For 4.2+. Fixed a broken xref in a doc

PR #18331, opened 2019-11-27
[enterprise-4.3] fixed path

kalexand-rh:
From https://github.com/openshift/openshift-docs/pull/18326/


kalexand-rh:
The CP bot worked, even though it threw an error, so I don't need this PR.

PR #18345, opened 2019-11-27
[WIP] CNV chapter restructure

aburdenthehand:
Taking advantage of US holidays to restructure the CNV repo now that it allows for 3rd level nesting.

Note: This is WIP and has not yet updated xrefs or module metadata


_topic_map.yml
```diff
@@ -1182,84 +1166,112 @@ Topics:
...
+      File: cnv-logs
+    - Name: Viewing events
+      File: cnv-events
+      Name: Viewing cluster information
+      File: cnv-using-dashboard-to-get-cluster-info
+    - Name: OpenShift cluster monitoring, logging, and Telemetry
+      File: cnv-openshift-cluster-monitoring
+    - Name: Collecting container-native virtualization data for Red Hat Support
+      File: cnv-collecting-cnv-data
 - Name: Container-native virtualization 2.1 release notes
```
kalexand-rh: I'd probably move the release notes assembly out of this subdirectory and to the top of the CNV bucket, just before installation.

_topic_map.yml
```diff
@@ -1166,12 +1156,6 @@ Topics:
 #### VIRTUAL MACHINE CHESS SALAD (silly name to highlight that the commented out assemblies need to be checked against merged filenams)
```
kalexand-rh: Is the "Container-native virtualization user's guide" folder still helpful? I think it might be more helpful to move all of its contents directly into the CNV "bucket" and set up other logical groupings under the main header. I'd try to keep only the assemblies that have no similar assemblies at the top level, something like:

Container-native virtualization
- Release notes
> Installation
> Virtual machines
     > Creating virtual machines
     > Virtual machine networking
     ....
> Live migration
> Node maintenance
....
- Using the CLI tools

_topic_map.yml
```diff
@@ -1166,12 +1156,6 @@ Topics:
...
   - Name: Creating virtual machines
     File: cnv-create-vms
-# Importing virtual machines
-  - Name: TLS certificates for DataVolume imports
-    File: cnv-tls-certificates-for-dv-imports
-  - Name: Importing virtual machine images with DataVolumes
-    File: cnv-importing-virtual-machine-images-datavolumes
-# VM CHESS SALAD cont'd
   - Name: Editing virtual machines
```
kalexand-rh: Can you create another logical folder for some of these single VM topics?

kalexand-rh (commented):
I have a couple of suggestions, and I think you need to rebase.

_topic_map.yml
aburdenthehand: Sounds good.

_topic_map.yml
aburdenthehand: That is an inspired suggestion. I created #18366 as a branch off this one to see how they compare and I think it's a great improvement over this one.

PR #18367, opened 2019-11-28
WIP Add note regarding network policy and host network routers

jboxman:
- https://bugzilla.redhat.com/show_bug.cgi?id=1768608
- https://bugzilla.redhat.com/show_bug.cgi?id=1764220

So, the cliff notes version is simply this:

If you enable network policy for a project, only traffic that is allowed by the policy can reach the Pod. However, this is not exactly the case for the ingress router. One some platforms (other than AWS, GCP, Azure), if you define a network policy in a project, traffic from the router _cannot_ reach any Pods in that project.

So the goal is to convey to a user that, if you are using a cluster installed on a platform other than AWS, GCP, and Azure, if you have a project with Pods that must be reachable from the ingress router, it is not possible to define network policy objects in that project.

(Specifically, if the router is installed on the _host network_, network policy cannot allow traffic from it.)

### DIFF
--- a/modules/nw-networkpolicy-about.adoc
+++ b/modules/nw-networkpolicy-about.adoc
@@ -6,10 +6,11 @@
 
 = About network policy
 
-In a cluster using a Kubernetes Container Network Interface (CNI) plug-in that
-supports NetworkPolicy, network isolation is controlled entirely by
-NetworkPolicy objects. In {product-title} {product-version}, OpenShift SDN
-supports using NetworkPolicy in its default network isolation mode.
+In a cluster using a Kubernetes Container Network Interface (CNI) Pod network
+provider plug-in that supports NetworkPolicy, network isolation is controlled
+entirely by NetworkPolicy objects. In {product-title} {product-version},
+OpenShift SDN supports using NetworkPolicy in its default network isolation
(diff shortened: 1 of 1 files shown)


danwinship:
> it is not possible to define network policy objects in that project.

It's not quite that harsh. As long as you create a policy allowing access to the relevant pod from all sources, then you can have whatever other policies you want in the namespace. (eg, you can have an "allow only from same namespace" policy that applies to everyone, and then add an "allow from everyone" policy that applies only to the pod targeted to the router.)

> (Specifically, if the router is installed on the _host network_, network policy cannot allow traffic from it.)

Well, you can still do the hack documented in the 3.x docs of adding a label to the `default` namespace and then allowing from namespaces with that label. This has the side effect of also allowing hostNetwork traffic.

Additionally, in 4.3 you could now create an `ipBlock` policy allowing traffic from the cluster's `machineCIDR`. That's actually not a bad approach... (We'd need to get QE to confirm that this works though.)

jboxman:
This is superseded by https://github.com/openshift/openshift-docs/pull/19360.

I'll create a Jira issue to track documenting the `ipBlock` approach for 4.3+.

PR #18393, opened 2019-12-02
dedicated logging corrections

sheriff-rh:
Preview build: http://file.rdu.redhat.com/~antaylor/dedicatedlogging/logging/dedicated-cluster-deploying.html 
@wgordon17  please take a look.

### DIFF
--- a/modules/dedicated-cluster-install-deploy.adoc
+++ b/modules/dedicated-cluster-install-deploy.adoc
@@ -6,7 +6,7 @@
 
 = Installing the Cluster Logging and Elasticsearch Operators
 
-You can use the {product-title} console to install cluster logging, by deploying,
+You can use the {product-title} console to install cluster logging by deploying instances of
 the Cluster Logging and Elasticsearch Operators.  The Cluster Logging Operator
 creates and manages the components of the logging stack.  The Elasticsearch Operator
 creates and manages the Elasticsearch cluster used by cluster logging.
@@ -14,24 +14,31 @@ creates and manages the Elasticsearch cluster used by cluster logging.
 [NOTE]
 ====
 The {product-title} cluster logging solution requires that you install both the
-Cluster Logging Operator and Elasticsearch Operator.
+Cluster Logging Operator and Elasticsearch Operator. When you deploy an instance
+of the Cluster Logging Operator, it also deploys an instance of the Elasticsearch
+Operator.
 ====
 
-.Prerequisites
+Your OpenShift Dedicated cluster includes 600 GiB of persistent storage that is
+exclusively available for deploying Elasticsearch for cluster logging.
 
-Ensure that you have the necessary persistent storage for Elasticsearch. Note that each Elasticsearch node
-requires its own storage volume.
-
-Elasticsearch is a memory-intensive application. Each Elasticsearch node needs 16G of memory for both memory requests and limits.
-The initial set of {product-title} nodes might not be large enough to support the Elasticsearch cluster. You must add additional nodes to the
-{product-title} cluster to run with the recommended or higher memory. Each Elasticsearch node can operate with a lower
-memory setting though this is not recommended for production deployments.
+Elasticsearch is a memory-intensive application. Each Elasticsearch node needs
+8G of memory for both memory requests and limits. Each Elasticsearch node can
+operate with a lower memory setting, though this is not recommended for
+production deployments.
 
 .Procedure
 
-. Install the Elasticsearch Operator via the OperatorHub. See: xref:../operators/olm-adding-operators-to-cluster.adoc[Adding Operators to a cluster].
+. Install the Elasticsearch Operator from the OperatorHub:
 
-. Install the Cluster Logging Operator using the {product-title} web console for best results:
+.. In the {product-title} web console, click *Operators* -> *OperatorHub*.
+
+.. Choose *Elasticsearch* from the list of available Operators, and click *Install*.
+
+.. On the *Create Operator Subscription* page, under *A specific namespace on the cluster* select *openshift-logging*.
+Then, click *Subscribe*.
+
+. Install the Cluster Logging Operator from the OperatorHub:
 
 .. In the {product-title} web console, click *Operators* -> *OperatorHub*.
 
@@ -44,10 +51,8 @@ Then, click *Subscribe*.
 
 .. Switch to the *Operators* → *Installed Operators* page.
 
-.. Ensure that *Cluster Logging* is listed in the *openshift-logging* project with a *Status* of *InstallSucceeded*.
-
-.. Ensure that *Elasticsearch Operator* is listed in the *openshift-operator-redhat* project with a *Status* of *InstallSucceeded*.
-The Elasticsearch Operator is copied to all other projects.
+.. Ensure that *Cluster Logging* and *Elasticsearch* Operators are listed in the
+*openshift-logging* project with a *Status* of *InstallSucceeded*.
 +
 [NOTE]
 ====
@@ -59,29 +64,19 @@ If either operator does not appear as installed, to troubleshoot further:
 +
 * Switch to the *Operators* → *Installed Operators* page and inspect
 the *Status* column for any errors or failures.
-* Switch to the *Workloads* → *Pods* page and check the logs in any Pods in the
-`openshift-logging` and `openshift-operators-redhat` projects that are reporting issues.
+* Switch to the *Workloads* → *Pods* page and check the logs in each Pod in the
+`openshift-logging` project that is reporting issues.
 
-. Create a cluster logging instance:
+. Create and deploy a cluster logging instance:
 
-.. Switch to the *Administration* -> *Custom Resource Definitions* page.
+.. Switch to the *Operators* → *Installed Operators* page.
 
-.. On the *Custom Resource Definitions* page, click *ClusterLogging*.
+.. Click the installed *Cluster Logging* Operator.
 
-.. On the *Custom Resource Definition Overview* page, select *View Instances* from the *Actions* menu.
-
-.. On the *Cluster Loggings* page, click *Create Cluster Logging*.
+.. Under the *Overview* tab, click *Create Instance* . Paste the following YAML
+definition into the window that displays.
 +
-You might have to refresh the page to load the data.
-
-.. In the YAML, replace the code with the following:
-+
-[NOTE]
-====
-This default cluster logging configuration should support a wide array of environments. Review the topics on tuning and
-configuring the cluster logging components for information on modifications you can make to your cluster logging cluster.
-====
-+
+.Cluster Logging Custom Resource (CR)
 [source,yaml]
 ----
 apiVersion: "logging.openshift.io/v1"
@@ -124,12 +119,12 @@ spec:
         node-role.kubernetes.io/worker: ""
 ----
 
-.. Click *Create*. This creates the Cluster Logging Custom Resource and Elasticsearch Custom Resource, which you
-can edit to make changes to your cluster logging cluster.
+.. Click *Create* to deploy the logging instance, which creates the Cluster
+Logging and Elasticsearch Custom Resources.
 
-. Verify the install:
+. Verify that the Pods for the Cluster Logging instance deployed:
 
-.. Switch to the *Workloads* -> *Pods* page.
+.. Switch to the *Workloads* → *Pods* page.
 
 .. Select the *openshift-logging* project.
 +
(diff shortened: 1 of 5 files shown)


modules/dedicated-cluster-install-deploy.adoc
```diff
@@ -19,19 +19,27 @@ Cluster Logging Operator and Elasticsearch Operator.
 
 .Prerequisites
 
-Ensure that you have the necessary persistent storage for Elasticsearch. Note that each Elasticsearch node
-requires its own storage volume.
+You are provided with a maximum of 600GiB of persistent storage exclusively
```
kalexand-rh: The prereqs should be in an ordered list.

maybe

```
* Create storage volumes for each Elasticsearch node. Your {product-title} cluster includes 600 GiB of persistent storage that is exclusively for deploying Elasticsearch for cluster logging.
```

Do these exist by default? Does a user need to create them before they start? Do they know how many nodes to create up front?

modules/dedicated-cluster-install-deploy.adoc
```diff
@@ -19,19 +19,27 @@ Cluster Logging Operator and Elasticsearch Operator.
...
+available for cluster logging related Elasticsearch deployments. Note that each
+Elasticsearch node requires its own storage volume.
 
-Elasticsearch is a memory-intensive application. Each Elasticsearch node needs 16G of memory for both memory requests and limits.
-The initial set of {product-title} nodes might not be large enough to support the Elasticsearch cluster. You must add additional nodes to the
-{product-title} cluster to run with the recommended or higher memory. Each Elasticsearch node can operate with a lower
-memory setting though this is not recommended for production deployments.
+Elasticsearch is a memory-intensive application. Each Elasticsearch node needs
```
kalexand-rh: maybe
```
* Allocate enough memory for each Elasticsearch node to manage both memory requests and limits. For production environments, allocate 8 GiB of memory. You can allocate as little as <state_minimum_value> for non-production environments.
```

Will a user know how to do this? How low can you go with the memory?

modules/dedicated-cluster-install-deploy.adoc
```diff
@@ -19,19 +19,27 @@ Cluster Logging Operator and Elasticsearch Operator.
...
+8G of memory for both memory requests and limits. Each Elasticsearch node can
+operate with a lower memory setting though this is not recommended for
+production deployments.
 
 .Procedure
 
-. Install the Elasticsearch Operator via the OperatorHub. See: xref:../operators/olm-adding-operators-to-cluster.adoc[Adding Operators to a cluster].
+. Install the Elasticsearch Operator via the OperatorHub:
```
kalexand-rh: s/via/from

> ! via prep
Use this term, which means “by way of,” to refer to routing data, such as in “Messages are routed via
SMTP.” In other situations, try to replace “via” with one of the following synonyms: “across,” along,” “by,”
“from,” “on,” or “through.”

modules/dedicated-cluster-install-deploy.adoc
```diff
@@ -19,19 +19,27 @@ Cluster Logging Operator and Elasticsearch Operator.
...
 
-. Install the Cluster Logging Operator using the {product-title} web console for best results:
+.. In the {product-title} web console, click *Operators* -> *OperatorHub*.
+
+.. Choose  *Elasticsearch* from the list of available Operators, and click *Install*.
```
kalexand-rh: There's an extra space here.

modules/dedicated-cluster-install-deploy.adoc
```diff
@@ -19,19 +19,27 @@ Cluster Logging Operator and Elasticsearch Operator.
...
+
+.. On the *Create Operator Subscription* page, under *A specific namespace on the cluster* select *openshift-logging*.
```
kalexand-rh: Does this namespace exist by default? I don't see instructions to make it in the assembly.

modules/dedicated-cluster-install-deploy.adoc
```diff
@@ -19,19 +19,27 @@ Cluster Logging Operator and Elasticsearch Operator.
...
+Then, click *Subscribe*.
+
+. Install the Cluster Logging Operator via the OperatorHub:
```
kalexand-rh: s/via/from

modules/dedicated-cluster-install-deploy.adoc
```diff
@@ -60,28 +66,18 @@ If either operator does not appear as installed, to troubleshoot further:
 * Switch to the *Operators* → *Installed Operators* page and inspect
 the *Status* column for any errors or failures.
 * Switch to the *Workloads* → *Pods* page and check the logs in any Pods in the
```
kalexand-rh: s/any Pods/each Pod

modules/dedicated-cluster-install-deploy.adoc
```diff
@@ -60,28 +66,18 @@ If either operator does not appear as installed, to troubleshoot further:
...
+.. Click the installed *Cluster Logging* Operator.
 
-.. In the YAML, replace the code with the following:
-+
-[NOTE]
-====
-This default cluster logging configuration should support a wide array of environments. Review the topics on tuning and
-configuring the cluster logging components for information on modifications you can make to your cluster logging cluster.
-====
+.. Under the *Overview* tab, click *Create Instance* . You should now copy and
```
kalexand-rh: maybe 
```
.. Under the *Overview* tab, click *Create Instance*. Paste the following YAML definition into the window that displays.
```

>! should v
Always prefer a more direct verb or way of saying something. Omit “should,” or use “must” or “might,” as
appropriate.

>! below adj
Do not use to indicate a relative location in a document, as in “the information below.”

modules/dedicated-cluster-install-deploy.adoc
```diff
@@ -146,3 +141,6 @@ You should see several pods for cluster logging, Elasticsearch, Fluentd, and Kib
 * fluentd-pb2f8
 * fluentd-zqgqx
 * kibana-7fb4fd4cc9-bvt4p
+
+. Access the Cluster Logging interface, *Kibana*, via the *Monitoring* →
```
kalexand-rh: s/via/from

modules/dedicated-cluster-install-deploy.adoc
```diff
@@ -124,12 +120,11 @@ spec:
         node-role.kubernetes.io/worker: ""
 ----
 
-.. Click *Create*. This creates the Cluster Logging Custom Resource and Elasticsearch Custom Resource, which you
-can edit to make changes to your cluster logging cluster.
+.. Click *Create*. This creates the Cluster Logging and Elasticsearch Custom Resources.
```
kalexand-rh: maybe 
```
.. Click *Create* to deploy the logging instance, which creates the Cluster Logging and Elasticsearch Custom Resources that it uses.
```

modules/dedicated-cluster-install-deploy.adoc
```diff
@@ -6,7 +6,7 @@
 
 = Installing the Cluster Logging and Elasticsearch Operators
 
-You can use the {product-title} console to install cluster logging, by deploying,
+You can use the {product-title} console to install cluster logging by deploying
```
kalexand-rh: I'd s/deploying/deploying instances of

modules/dedicated-cluster-install-deploy.adoc
```diff
@@ -60,28 +66,18 @@ If either operator does not appear as installed, to troubleshoot further:
 * Switch to the *Operators* → *Installed Operators* page and inspect
 the *Status* column for any errors or failures.
 * Switch to the *Workloads* → *Pods* page and check the logs in any Pods in the
-`openshift-logging` and `openshift-operators-redhat` projects that are reporting issues.
+`openshift-logging` project that is reporting issues.
 
 . Create a cluster logging instance:
```
kalexand-rh: I might s/Create/Deploy

I might also clarify that the logging operator call the elastic search operators when you deploy the logging instance.

modules/dedicated-cluster-install-deploy.adoc
```diff
@@ -19,19 +19,27 @@ Cluster Logging Operator and Elasticsearch Operator.
```
kalexand-rh: I might change the previous note to explain that deploying the logging operator also deploys the elasticsearch operator.

modules/dedicated-cluster-install-deploy.adoc
```diff
@@ -124,12 +120,11 @@ spec:
         node-role.kubernetes.io/worker: ""
 ----
 
-.. Click *Create*. This creates the Cluster Logging Custom Resource and Elasticsearch Custom Resource, which you
-can edit to make changes to your cluster logging cluster.
+.. Click *Create*. This creates the Cluster Logging and Elasticsearch Custom Resources.
 
 . Verify the install:
```
kalexand-rh: Maybe "Verify that the Pods for the Cluster Logging instance deployed:"

kalexand-rh (commented):
I have a couple of picks and suggestions. This is looking good.

modules/dedicated-cluster-install-deploy.adoc
sheriff-rh: Honestly, I think we should just do away with the `.Prerequisites`. Nothing in there is a prereq, now that I'm looking at it. I think it could be a continuation of the statement above the note. This is just informational text.  What do you think?

modules/dedicated-cluster-install-deploy.adoc
sheriff-rh: The 8gb is determined by the YAML, which is not really recommended to alter. I haven't heard of a lower setting, actually. Good point.

modules/dedicated-cluster-install-deploy.adoc
kalexand-rh: It depends on if the user needs to actually do anything to prepare these items or if it's mostly FYI. If it's FYI, I agree that removing the prereqs section would be better.

modules/dedicated-cluster-install-deploy.adoc
sheriff-rh: Agreed. Done!

modules/dedicated-cluster-install-deploy.adoc
sheriff-rh: Got it! Thanks

modules/dedicated-cluster-install-deploy.adoc
sheriff-rh: Correct - great question!

modules/dedicated-cluster-install-deploy.adoc
sheriff-rh: Yep, fixed.

modules/dedicated-cluster-install-deploy.adoc
sheriff-rh: Agreed, fixed.

modules/dedicated-cluster-install-deploy.adoc
sheriff-rh: This is fantastic, thank you.

modules/dedicated-cluster-install-deploy.adoc
sheriff-rh: Fixed!

modules/dedicated-cluster-install-deploy.adoc
sheriff-rh: Excellent! Added.

modules/dedicated-cluster-install-deploy.adoc
sheriff-rh: Yep. Added.

modules/dedicated-cluster-install-deploy.adoc
sheriff-rh: How about Create and deploy? That seems like an accurate description of the steps that follow. As for the logging/elasticsearch, there is a note on line 16 that calls out this requirement, and step 1 and 2 are install both Operators.

modules/dedicated-cluster-install-deploy.adoc
sheriff-rh: Agreed! Added.

modules/dedicated-cluster-install-deploy.adoc
kalexand-rh: That's good to know!

modules/dedicated-cluster-install-deploy.adoc
kalexand-rh: I think that if you clarify the note at line 16 that "Create and deploy" would be fine. I think it would be helpful to be explicit about one calling the other.

modules/dedicated-cluster-install-deploy.adoc
wgordon17: Just FYI

PR #18397, opened 2019-12-03
Update contributing.adoc

olga-mir:
Update OpenShift distros list to include `openshift-aro`


kalexand-rh:
@vikram-redhat, PTAL?

PR #18425, opened 2019-12-03
OSDOCS-788 TLS Configuration

bmcelvee:
https://jira.coreos.com/browse/OSDOCS-788

### DIFF
--- a/modules/nw-ingress-controller-configuration-parameters.adoc
+++ b/modules/nw-ingress-controller-configuration-parameters.adoc
@@ -80,15 +80,26 @@ nodePlacement:
 ----
 ====
 
-//|`tlsSecurityProfile`
-//|`tlsSecurityProfile` specifies settings for TLS connections for Ingress controllers.
-//
-//If not set, the default is based on the `apiservers.config.openshift.io/cluster` resource.
-//
-//When using the `Old`, `Intermediate`, and `Modern` profile types, the effective profile configuration is subject to change between releases. For example, given a specification to use the `Intermediate` profile deployed on release `X.Y.Z`, an upgrade to release `X.Y.Z+1` may cause a new profile configuration to be applied to the Ingress controller, resulting in a rollout.
-//
-//The minimum TLS version for Ingress controllers is `1.1`, and the maximum TLS version is `1.2`.  An implication of this restriction is that the Modern TLS profile type cannot be used because it requires TLS `1.3`.
-//
+|`tlsSecurityProfile`
+|`tlsSecurityProfile` specifies settings for TLS connections for Ingress controllers.
+
+If not set, the default value is based on the `apiservers.config.openshift.io/cluster` resource.
+
+When using the `Old`, `Intermediate`, and `Modern` profile types, the effective profile configuration is subject to change between releases. For example, given a specification to use the `Intermediate` profile deployed on release `X.Y.Z`, an upgrade to release `X.Y.Z+1` may cause a new profile configuration to be applied to the Ingress controller, resulting in a rollout.
+
+The minimum TLS version for Ingress controllers is `1.1`, and the maximum TLS version is `1.2`.
+
+[IMPORTANT]
+====
+The HAProxy Ingress controller image does not support TLS `1.3` and because the `Modern` profile requires TLS `1.3`, it is not supported. The Ingress Operator converts the `Modern` profile to `Intermediate`.
+
+The Ingress Operator also converts the TLS `1.0` of an `Old` or `Custom` profile to `1.1`, and TLS `1.3` of a `Custom` profile to `1.2`.
+====
+
+[NOTE]
+====
+Ciphers and the minimum TLS version of the configured security profile are reflected in the `TLSProfile` status.
+====
 
 |===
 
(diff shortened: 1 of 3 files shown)


modules/nw-ingress-controller-configuration-parameters.adoc
```diff
@@ -80,15 +80,21 @@ nodePlacement:
...
+
+If not set, the default value is based on the `apiservers.config.openshift.io/cluster` resource.
+
+When using the `Old`, `Intermediate`, and `Modern` profile types, the effective profile configuration is subject to change between releases. For example, given a specification to use the `Intermediate` profile deployed on release `X.Y.Z`, an upgrade to release `X.Y.Z+1` may cause a new profile configuration to be applied to the Ingress controller, resulting in a rollout.
+
+The minimum TLS version for Ingress controllers is `1.1`, and the maximum TLS version is `1.2`.
+
+[IMPORTANT]
+====
+The `Modern` profile is not supported because it requires TLS `1.3`. The Ingress Operator converts the `Modern` profile to `Intermediate`.
```
danehans: > The `Modern` profile is not supported because it requires TLS `1.3`

We may want to provide additional information here. The current HAProxy ingress controller (i.e. router) image does not support TLS 1.3.

danehans:
I suggest making a note that ciphers and minimum tls version of the configured security profile are reflected in [status](https://github.com/openshift/api/blob/master/operator/v1/types_ingress.go#L356-L358). I think it also makes sense to include `Custom` profile docs. Additional information on the `Custom` profile can be found [here](https://github.com/openshift/api/blob/master/config/v1/types_tlssecurityprofile.go#L104-L117).

modules/nw-ingress-controller-configuration-parameters.adoc
bmcelvee: Thanks! Added.

bmcelvee:
I added a module for the TLS security profiles since there was a fair amount of information that needed to be included.

bmcelvee:
@lihongan please take a look when you get a chance, thanks!

modules/nw-ingress-controller-tls-profiles.adoc
```diff
@@ -0,0 +1,56 @@
...
+
+.Example `Custom` profile
+[source]
+----
+ciphers:
+   - ECDHE-ECDSA-CHACHA20-POLY1305
+   - ECDHE-RSA-CHACHA20-POLY1305
+   - ECDHE-RSA-AES128-GCM-SHA256
+   - ECDHE-ECDSA-AES128-GCM-SHA256
+minTLSVersion: TLSv1.1
```
lihongan: better to describe fully configuration in `spec` for `Custom` profile. e.g.
```
spec:
  tlsSecurityProfile:
    type: Custom
    custom:
      ciphers:
        - ECDHE-ECDSA-AES128-GCM-SHA256
        - ECDHE-RSA-AES128-GCM-SHA256
      minTLSVersion: VersionTLS11
```

modules/nw-ingress-controller-tls-profiles.adoc
```diff
@@ -0,0 +1,56 @@
...
+[discrete]
+== Old
+
+[id="nw-ingress-controller-tls-profiles-intermediate{context}"]
+[discrete]
+== Intermediate
+
+[id="nw-ingress-controller-tls-profiles-modern{context}"]
+[discrete]
+== Modern
```
lihongan: For `Old` `Intermediate` `Modern` can we have a configuration example like:
```
spec:
  tlsSecurityProfile:
    type: Modern
```

bmcelvee:
@openshift/team-documentation Please peer review, thanks!

modules/nw-ingress-controller-tls-profiles.adoc
```diff
@@ -0,0 +1,61 @@
...
+spec:
+  tlsSecurityProfile:
+    type: Modern
+----
+
+The `Custom` profile is a user-defined TLS security profile.
+
+[WARNING]
+====
+You must be careful using a `Custom` profile as invalid configurations can cause problems.
```
bergerhoffer: Per IBM, shouldn't use "as" to mean "because".

> Do not use to mean “because” or “while.”

Suggest also adding a comma:

```
 You must be careful using a `Custom` profile, because invalid configurations can cause problems.
```

modules/nw-ingress-controller-tls-profiles.adoc
```diff
@@ -0,0 +1,61 @@
...
+====
+
+.Example `Custom` profile
+[source]
```
bergerhoffer: Add yaml

modules/nw-ingress-controller-tls-profiles.adoc
```diff
@@ -0,0 +1,61 @@
...
+----
+
+The `Custom` profile is a user-defined TLS security profile.
+
+[WARNING]
+====
+You must be careful using a `Custom` profile as invalid configurations can cause problems.
+====
+
+.Example `Custom` profile
```
bergerhoffer: Consider making this title consistent with the others.

```
 .Sample `Custom` profile configuration
```

bergerhoffer (commented):
Just a few minor things.

bmcelvee:
Thanks, @bergerhoffer and all!

PR #18427, opened 2019-12-03
Provide an example to mirror images

radudd:
The example shows how to mirror an image in the internal registry from one tag to a different one. It would be useful to show an example how to mirror a sample image from external registry to the internal one.


radudd:
Additionally, provide a patch command to update the `samplesRegistry` parameter

kevchu3 (commented):
This worked for me.  The current version of the docs don't provide a working command to update the samplesRegistry field so this commit would be helpful.

vikram-redhat:
@bmcelvee can you take a look?

PR #18431, opened 2019-12-04
OSDOCS-471: Adding etcd encryption docs

bergerhoffer:
Preview: http://file.rdu.redhat.com/~ahoffer/2019/OSDOCS-471/authentication/encrypting-etcd.html

### DIFF
--- a/authentication/encrypting-etcd.adoc
+++ b/authentication/encrypting-etcd.adoc
@@ -1,0 +1,15 @@
+[id="encrypting-etcd"]
+= Encrypting etcd data
+include::modules/common-attributes.adoc[]
+:context: encrypting-etcd
+
+toc::[]
+
+// About etcd encryption
+include::modules/about-etcd-encryption.adoc[leveloffset=+1]
+
+// Enabling etcd encryption
+include::modules/enabling-etcd-encryption.adoc[leveloffset=+1]
+
+// Disabling etcd encryption
+include::modules/disabling-etcd-encryption.adoc[leveloffset=+1]
--- a/modules/about-etcd-encryption.adoc
+++ b/modules/about-etcd-encryption.adoc
@@ -1,0 +1,18 @@
+// Module included in the following assemblies:
+//
+// * authentication/encrypting-etcd.adoc
+
+[id="about-etcd_{context}"]
+= About etcd encryption
+
+By default, etcd data is not encrypted in {product-title}. You can enable etcd encryption for your cluster to provide an additional layer of data security. For example, it protects the loss of sensitive data if an etcd backup is exposed to the incorrect parties.
+
+When you enable etcd encryption, the following OpenShift API server and Kubernetes API server resources are encrypted:
+
+* Secrets
+* ConfigMaps
+* Routes
+* OAuth access tokens
+* OAuth authorize tokens
+
+When you enable etcd encryption, encryption keys are created. These keys are rotated on a weekly basis. You must have these keys in order to restore from an etcd backup.
--- a/modules/disabling-etcd-encryption.adoc
+++ b/modules/disabling-etcd-encryption.adoc
@@ -1,0 +1,65 @@
+// Module included in the following assemblies:
+//
+// * authentication/encrypting-etcd.adoc
+
+[id="disabling-etcd-encryption_{context}"]
+= Disabling etcd encryption
+
+You can disable encryption of etcd data in your cluster.
+
+.Prerequisites
+
+* Access to the cluster as a user with the `cluster-admin` role.
+
+.Procedure
+
+. Modify the API server object:
++
+----
+$ oc edit apiserver
+----
+
+. Set the `encryption` field type to `identity`:
++
+[source,yaml]
+----
+spec:
+  encryption:
+    type: identity
+----
+
+. Save the file to apply the changes.
++
+The decryption process starts. It can take 20 minutes or longer for this process to complete, depending on the size of your cluster.
+
+. Verify that etcd decryption was successful.
+
+.. Review the `Encrypted` status condition for the OpenShift API server to verify that its resources were successfully decrypted:
++
+----
+$ oc get openshiftapiserver -o=jsonpath='{range .items[0].status.conditions[?(@.type=="Encrypted")]}{.reason}{"\n"}{.message}{"\n"}'
+----
++
+The output shows `DecryptionCompleted` upon successful decryption:
++
+----
+DecryptionCompleted
+Encryption mode set to identity and everything is decrypted
+----
++
+If the output shows `DecryptionInProgress`, this means that decryption is still in progress. Wait a few minutes and try again.
+
+.. Review the `Encrypted` status condition for the Kubernetes API server to verify that its resources were successfully decrypted:
++
+----
+$ oc get kubeapiserver -o=jsonpath='{range .items[0].status.conditions[?(@.type=="Encrypted")]}{.reason}{"\n"}{.message}{"\n"}'
+----
++
+The output shows `DecryptionCompleted` upon successful decryption:
++
+----
+DecryptionCompleted
+Encryption mode set to identity and everything is decrypted
+----
++
+If the output shows `DecryptionInProgress`, this means that decryption is still in progress. Wait a few minutes and try again.
--- a/modules/enabling-etcd-encryption.adoc
+++ b/modules/enabling-etcd-encryption.adoc
@@ -1,0 +1,70 @@
+// Module included in the following assemblies:
+//
+// * authentication/encrypting-etcd.adoc
+
+[id="enabling-etcd-encryption_{context}"]
+= Enabling etcd encryption
+
+You can enable etcd encryption to encrypt sensitive resources in your cluster.
+
+// [WARNING]
+// ====
+// Do not back up etcd until the encryption process is complete.
+// ====
+
+.Prerequisites
+
+* Access to the cluster as a user with the `cluster-admin` role.
+
+.Procedure
+
+. Modify the API server object:
++
+----
+$ oc edit apiserver
+----
+
+. Set the `encryption` field type to `aescbc`:
++
+[source,yaml]
+----
+spec:
+  encryption:
+    type: aescbc
+----
+
+. Save the file to apply the changes.
++
+The encryption process starts. It can take 20 minutes or longer for this process to complete, depending on the size of your cluster.
+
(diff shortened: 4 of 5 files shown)


bergerhoffer:
@sttts Can you please review?

Preview: http://file.rdu.redhat.com/~ahoffer/2019/OSDOCS-471/authentication/encrypting-etcd.html

modules/disabling-etcd-encryption.adoc
```diff
@@ -0,0 +1,65 @@
...
+. Save the file to apply the changes.
++
+The decryption process starts. It may take 20 minutes or longer for this process to complete, depending on the size of your cluster.
+
+. Verify that etcd decryption was successful.
+
+.. Verify that the status of the `Encrypted` status condition is `DecryptionCompleted` for the OpenShift API server:
++
+----
+$ oc get openshiftapiserver -o=jsonpath='{range .items[0].status.conditions[?(@.type=="Encrypted")]}{.reason}{"\n"}{.message}{"\n"}'
```
bergerhoffer: @sttts I did the best I could with what output to show here, but let me know if you think we should show more fields, or just do `oc get openshiftapiserver -o yaml` and have them manually look for the Encrypted condition.

sttts: Looks fine. I see you have the black belt in JSON paths :+1:

modules/disabling-etcd-encryption.adoc
```diff
@@ -0,0 +1,65 @@
...
++
+If the output shows `DecryptionInProgress`, this means that decryption is still in progress. Wait a few minutes and try again.
+
+.. Verify that the status of the `Encrypted` status condition is `DecryptionCompleted` for the Kubernetes API server:
++
+----
+$ oc get kubeapiserver -o=jsonpath='{range .items[0].status.conditions[?(@.type=="Encrypted")]}{.reason}{"\n"}{.message}{"\n"}'
+----
++
+Successful encryption results in the following output:
```
sttts: decryption

modules/enabling-etcd-encryption.adoc
```diff
@@ -0,0 +1,70 @@
...
+// * authentication/encrypting-etcd.adoc
+
+[id="enabling-etcd_{context}"]
+= Enabling etcd encryption
+
+You can enable etcd encryption to encrypt sensitive resources in your cluster.
+
+[WARNING]
+====
+Do not back up etcd until the encryption process is complete.
```
sttts: actually this should work (@xingxingxia I assume you test this). Of course the backup then is not fully encrypted.

modules/disabling-etcd-encryption.adoc
```diff
@@ -0,0 +1,65 @@
...
+[source,yaml]
+----
+spec:
+  encryption:
+    type: identity
+----
+
+. Save the file to apply the changes.
++
+The decryption process starts. It may take 20 minutes or longer for this process to complete, depending on the size of your cluster.
```
xingxingxia: I'd like to document https://bugzilla.redhat.com/show_bug.cgi?id=1775554 more detailed, e.g. time for revisions of restarts is needed, time for real migration is then consumed depending on the size of your cluster.

modules/enabling-etcd-encryption.adoc
xingxingxia: Yes, it should work. This is moved to Etcd Component and pending Etcd QE Ge Liu to verify it https://bugzilla.redhat.com/show_bug.cgi?id=1776811 which needs additional Doc in DR doc section too.

xingxingxia:
Except above comment, this PR LGTM.

modules/disabling-etcd-encryption.adoc
sttts: :+1:

modules/enabling-etcd-encryption.adoc
bergerhoffer: So users don't need to wait until encryption is complete to take a backup? Both upon initial enabling and during the weekly rotation?  CC @retroflexer and @geliu2016 

I'm going to remove this warning for now and merge this PR. And we can follow up on this in the PR when we add update DR.

retroflexer: Thanks @xingxingxia for testing it again. The config with the failed state is active for next couple of days. @sttts and anyone else to look at it?

bergerhoffer:

modules/enabling-etcd-encryption.adoc
sttts: @retroflexer we don't have anybody in US timezone for that topic. We need access to the cluster tomorrow (EU) morning. I also commented on the BZ to get more logs and the encryption config.

modules/about-etcd-encryption.adoc
```diff
@@ -0,0 +1,18 @@
+// Module included in the following assemblies:
+//
+// * authentication/encrypting-etcd.adoc
+
+[id="about-etcd_{context}"]
```
kalexand-rh: s/about-etcd/about-tetcd-encryption

For [modules](https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#module-file-metadata), the anchor needs to match the file name.

Please update the other modules, too.

authentication/encrypting-etcd.adoc
```diff
@@ -0,0 +1,15 @@
...
+
+toc::[]
+
+// About etcd encryption
+include::modules/about-etcd-encryption.adoc[leveloffset=+1]
+
+// Enabling etcd encryption
+include::modules/enabling-etcd-encryption.adoc[leveloffset=+1]
+
+// Disabling etcd encryption
```
kalexand-rh: I think that the disabling module should go in a separate assembly because they're not part of the same user goal.

modules/about-etcd-encryption.adoc
```diff
@@ -0,0 +1,18 @@
...
+= About etcd encryption
```
kalexand-rh: On a recent minimalism call, Ingrid suggested removing "about" from modules titles because it doesn't add much value.

modules/about-etcd-encryption.adoc
```diff
@@ -0,0 +1,18 @@
...
+
+By default, etcd data is not encrypted in {product-title}. You can enable etcd encryption for your cluster to provide an additional layer of protection against data leakage. For example, it protects the loss of sensitive data if an etcd backup is exposed to the incorrect parties.
```
kalexand-rh: "data leakage" is a bit jargony and will probably not translate well. I might just say that it provide an additional later of data security.

modules/disabling-etcd-encryption.adoc
kalexand-rh: s/may/can

modules/disabling-etcd-encryption.adoc
```diff
@@ -0,0 +1,65 @@
...
+
+. Verify that etcd decryption was successful.
+
+.. Verify that the status of the `Encrypted` status condition is `DecryptionCompleted` for the OpenShift API server:
```
kalexand-rh: I might say "Verify that the OpenShift API server decrypted:" here and consider adding more details after the command.

modules/disabling-etcd-encryption.adoc
```diff
@@ -0,0 +1,65 @@
...
+Successful decryption results in the following output:
++
+----
+DecryptionCompleted
+Encryption mode set to identity and everything is decrypted
+----
++
+If the output shows `DecryptionInProgress`, this means that decryption is still in progress. Wait a few minutes and try again.
+
+.. Verify that the status of the `Encrypted` status condition is `DecryptionCompleted` for the Kubernetes API server:
```
kalexand-rh: Same suggestion here.

modules/enabling-etcd-encryption.adoc
```diff
@@ -0,0 +1,70 @@
...
+----
+
+. Set the `encryption` field type to `aescbc`:
++
+[source,yaml]
+----
+spec:
+  encryption:
+    type: aescbc
+----
```
kalexand-rh: If you have the information handy, I might add a sentence about why `aescbc` and a statement about whether or not there are other encryption options.

modules/enabling-etcd-encryption.adoc
```diff
@@ -0,0 +1,70 @@
...
+
+. Save the file to apply the changes.
++
+The encryption process starts. It may take 20 minutes or longer for this process to complete, depending on the size of your cluster.
```
kalexand-rh: s/may/can

modules/enabling-etcd-encryption.adoc
```diff
@@ -0,0 +1,70 @@
...
+
+. Verify that etcd encryption was successful.
+
+.. Verify that the status of the `Encrypted` status condition is `EncryptionCompleted` for the OpenShift API server:
```
kalexand-rh: Same suggestion from the decryption module for both these steps.

kalexand-rh (commented):
I have a couple of suggestions and requests, but this looks great.

modules/about-etcd-encryption.adoc
bergerhoffer: I did hear that, but I don't necessarily agree with it. I think it helps to make it clear that this is overview/conceptual information. Just putting "etcd encryption" would look odd to me. Unless we make a team decision to remove these instances, I'd like to keep this as is now.

kalexand-rh: I'll add it to the team meeting agenda.

modules/enabling-etcd-encryption.adoc
bergerhoffer: Good suggestion. Will add this in my next PR as I'll want that reviewed (and want to merge this PR now).

modules/enabling-etcd-encryption.adoc
bergerhoffer: @sttts @xingxingxia So what message do we want to present to users here? So even if it works to take a backup in the middle of the initial encryption process, if they do it and the backup isn't fully encrypted, I'm guessing that's something that we would want to warn them of.

What do you say to adding a warning along the lines of `Do not back up etcd until the encryption process is complete or the backup might not be fully encrypted`. Or `initial encryption process` if the concern is only for the first time encryption is enabled? wdyt?

Are there any concerns with taking a backup in the middle of a weekly rotation, or is that handled okay with the multiple versions kept of the keys?

PR #18432, opened 2019-12-04
 This is misleading and the wording could be improved. It is possible to also add RHEL worker nodes

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1774039

### DIFF
--- a/modules/installation-overview.adoc
+++ b/modules/installation-overview.adoc
@@ -172,6 +172,11 @@ the cluster, including the operating system itself. Each machine boots with a
 configuration that references resources hosted in the cluster that it joins.
 This configuration allows the cluster to manage itself as updates are applied.
 
+If your cluster uses a machine-provisioned infrastructure, you have the option of
+adding RHEL worker machines to your cluster. Note that the RHEL nodes are not 
+managed by MachineSets and you must manage and update the underlying operating system
+on these nodes.
+
 [discrete]
 === The installation process with user-provisioned infrastructure
 
--- a/machine_management/adding-rhel-compute.adoc
+++ b/machine_management/adding-rhel-compute.adoc
@@ -6,7 +6,7 @@ include::modules/common-attributes.adoc[]
 toc::[]
 
 In {product-title}, you can add Red Hat Enterprise Linux (RHEL) compute, or
-worker, machines to a user-provisioned infrastructure cluster. You can use RHEL
+worker, machines to a cluster. You can use RHEL
 as the operating system on only compute machines.
 
 include::modules/rhel-compute-overview.adoc[leveloffset=+1]
--- a/modules/rhel-compute-overview.adoc
+++ b/modules/rhel-compute-overview.adoc
@@ -6,9 +6,9 @@
 [id="rhel-compute-overview_{context}"]
 = About adding RHEL compute nodes to a cluster
 
-In {product-title} {product-version}, you have the option of using Red Hat Enterprise Linux (RHEL) machines as compute, or worker, machines in your cluster if you use a user-provisioned infrastructure installation. You must use {op-system-first} machines for the control plane, or master, machines in your cluster.
+In {product-title} {product-version}, you have the option of using Red Hat Enterprise Linux (RHEL) machines as compute, or worker, machines in your cluster. You must use {op-system-first} machines for the control plane, or master, machines in your cluster.
 
-As with all installations that use user-provisioned infrastructure, if you choose to use RHEL compute machines in your cluster, you take responsibility for all operating system life cycle management and maintenance, including performing system updates, applying patches, and completing all other required tasks.
+If you choose to use RHEL compute machines in your cluster, you take responsibility for all operating system life cycle management and maintenance, including performing system updates, applying patches, and completing all other required tasks.
 
 [IMPORTANT]
 ====


modules/installation-overview.adoc
```diff
@@ -172,6 +172,11 @@ the cluster, including the operating system itself. Each machine boots with a
 configuration that references resources hosted in the cluster that it joins.
 This configuration allows the cluster to manage itself as updates are applied.
 
+If your cluster uses a machine-provisioned infrastructure, you have the option of
```
kalexand-rh: Please close this PR. RHEL nodes are only allowed on UPI clusters.
> In OpenShift Container Platform 4.2, you have the option of using Red Hat Enterprise Linux (RHEL) machines as compute, or worker, machines in your cluster if you use a user-provisioned infrastructure installation. 

https://docs.openshift.com/container-platform/4.2/machine_management/adding-rhel-compute.html

mburke5678:
@sdodson PTAL

@xltian Can you review these changes or assign to someone who can? Thank you!

sdodson:
Everything works whether it's IPI or UPI. I know this is something that we'd talked about at length in the past and I'm sure @kalexand-rh is correct that we'd said UPI only at one point but I don't recall a final decision that was the only place it'd be supported.

xltian:
@gpei to review from QE.

gpei:
RHEL worker is only officially supported in UPI when QE start the testing on it since 4.1, so RHEL worker in IPI is not in QE's OS Test Matrix for OCP 4.1/4.2/4.3 testing. Sorry, I can't ack this change.

mburke5678:
@gpei Thank you very much for your clarification.

cc. @vikram-redhat This is the PR I asked you about in Slack PM

katherinedube:
Adding RHEL worker nodes to an existing cluster has always been a Day 2 operation. How you create that cluster is not dependent on the deployment method used. Meaning there's never been a technical restriction for using either user-provisioned or install-provision deployment method whenever someone wants add RHEL nodes to their cluster. If I recall, all the original RHEL worker node CI jobs were ran on OpenShift clusters deployed using the installer-provisioned infrastructure deployment method. 

More importantly, we need to start encouraging more customers to adopt the installer-provisioned infrastructure deployment method and not artificially keep forcing them into having to create all the infrastructure objects themselves if they want to perform operations such as use RHEL nodes. This is also the reason we've been extending installer-provisioned use cases by adding support for shared VPCs and private clusters in the next release of OpenShift. 

If we feel it needs testing, then let's get it tested but let's not block this for no tangible reason.

sdodson:
> If I recall, all the original RHEL worker node CI jobs were ran on OpenShift clusters deployed using the installer-provisioned infrastructure deployment method.

They still do, we make use of the machine api and machine sets as the most expedient way to generate new RHEL worker hosts to add to the cluster.

I agree that joining RHEL7 Workers have no dependency on whether or not the cluster is an IPI or UPI installation.

gpei:
1. The 4.3 Test Configuration Matrix was already decided in 4.3 initial stage, we don't have RHEL worker in IPI install proposed there. QE have a full test schedule until 4.3 GA based on the features to be published, so I'm afraid we don't have much chance to arrange such test request suddenly added in this period. This is not a simple verification, all the RHEL worker tests we covered in UPI, we also have to guarantee they were tested in IPI, also need to cover the features that are not supported in UPI but in IPI.

2. As installer QE team, we have to test against all platforms that IPI supported from scratch, including: IPI on AWS, IPI on Azure, IPI on GCP, IPI on OSP13, IPI on OSP15, also need to cover the RHEL node support on private cluster, disconnected install, enable proxy etc.
Here's some of the bugs QE found during testing on RHEL node in UPI, I' not sure these scenarios could be covered by upstream CI job.
Bug 1777388 - cloud-utils-growpart install conflicts with gce-disk-expand in google managed image when scale up rhel node
Bug 1745875 - Failed to add RHEL worker behind proxy
Bug 1744456 - [disconnected] rhel node worker can not be installed in a disconnected env

3. Due to the limited resource of installer QE team(we only have 2 people familiar with RHEL node scale-up feature), we have to test this in automation way for all platform support. In QE's automation job, we'll spin up a bastion host inside the VPC for forwarding the ssh connection, since the RHEL workers will be created in private subnet. And this work only could be done in 4.3 IPI which has BYO-VPC support, as to 4.1/4.2, RHEL worker in IPI install couldn't be realized in QE's automation framework for now. 

4. When QE start the research on adding RHEL worker in UPI-on-AWS, it cost almost two weeks to make a successful scale-up. I'm not sure how much time we'll spend on researching this in IPI, but I think at least one month is needed to finish the RHEL worker scale-up testing on all the platforms.

5. In addition to installation testing, QE also need to perform functional testing on the RHEL node to make sure all the necessary functions are working well on RHEL worker just like on RHCOS.
Here's two RHEL bugs QE found in Containers and MCO component
Bug 1708605 - container runtime error after upgrade RHEL 7 worker node
Bug 1708602 - Got error from machine-config-daemon when the rhel worker is adding to the cluster but cluster is applying new rendered machineconfig

IMO we couldn't simply say RHEL worker in IPI install is supported in the official doc before we fully tested it.

vikram-redhat:
If it hasn't been tested, we can't doc it. @mburke5678 let's close this PR till QE has had a chance to verify that the process works.

@aheslin @sferich888 @reestr

PR #18433, opened 2019-12-04
bug 1751223 RHEL minimum versions

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1751223

@wmengRH, will you PTAL?


wmengRH:
LGTM

PR #18445, opened 2019-12-04
CNV-3186: hostpath provisioner

ousleyp:
Jira: https://jira.coreos.com/browse/CNV-3186
Preview build: http://file.bos.redhat.com/pousley/010620/hostpath-cdi/cnv/cnv_users_guide/cnv-configuring-local-storage-for-vms.html

Please cherrypick to enterprise-4.3 (for CNV 2.2)

### DIFF
--- a/modules/cnv-about-hostpath-provisioner.adoc
+++ b/modules/cnv-about-hostpath-provisioner.adoc
@@ -1,0 +1,26 @@
+// Module included in the following assemblies:
+//
+// * cnv/cnv_users_guide/cnv-configuring-local-storage-for-vms.adoc
+
+[id="cnv-about-hostpath-provisioner_{context}"]
+= About the hostpath provisioner
+
+The hostpath provisioner is a local storage provisioner designed for
+{CNVProductName}. If you want to configure local storage for
+virtual machines, you must enable the hostpath provisioner first.
+
+When you install the {CNVProductName} Operator, the hostpath provisioner Operator
+is automatically installed. To use it, you must:
+
+* Configure SELinux:
+** If you use Red Hat Enterprise Linux CoreOS 8 workers, you must create a MachineConfig
+object on each node.
+** Otherwise, apply the SELinux label `container_file_t` to the PersistentVolume (PV) backing
+directory on each node.
+* Create a HostPathProvisioner custom resource.
+* Create a StorageClass object for the hostpath provisioner.
+
+The hostpath provisioner Operator deploys the provisioner as a _DaemonSet_ on each
+node when you create its custom resource. In the custom resource file, you specify
+the backing directory for the PersistentVolumes that the hostpath provisioner
+creates.
--- a/modules/cnv-creating-storage-class.adoc
+++ b/modules/cnv-creating-storage-class.adoc
@@ -1,0 +1,51 @@
+// Module included in the following assemblies:
+//
+// * cnv/cnv_users_guide/cnv-configuring-local-storage-for-vms.adoc
+
+[id="cnv-creating-storage-class_{context}"]
+= Creating a StorageClass object
+
+When you create a StorageClass object, you set parameters that affect the
+dynamic provisioning of PersistentVolumes (PVs) that belong to that storage class. +
+[NOTE]
+====
+You cannot update a StorageClass object's parameters after you create it.
+====
+
+.Procedure
+
+. Create a YAML file for defining the storage class. For example:
++
+----
+$ touch storageclass.yaml
+----
+
+. Edit the file. For example:
++
+[source,yaml]
+----
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: hostpath-provisioner <1>
+provisioner: kubevirt.io/hostpath-provisioner
+reclaimPolicy: Delete <2>
+volumeBindingMode: WaitForFirstConsumer <3>
+----
+<1> You can optionally rename the storage class by changing this value.
+<2> The two possible `reclaimPolicy` values are `Delete` and `Retain`. If you
+do not specify a value, the storage class defaults to `Delete`.
+<3> The `volumeBindingMode` value determines when dynamic provisioning and volume
+binding occur. Specify `WaitForFirstConsumer` to delay the binding and provisioning
+of a PV until after a Pod that uses the PersistentVolumeClaim (PVC)
+is created. This ensures that the PV meets the Pod's scheduling requirements.
+
+. Create the StorageClass object:
++
+----
+$ oc create -f storageclass.yaml
+----
+
+.Additional information
+
+* link:https://kubernetes.io/docs/concepts/storage/storage-classes/[Storage Classes]
--- a/modules/cnv-using-hostpath-provisioner.adoc
+++ b/modules/cnv-using-hostpath-provisioner.adoc
@@ -1,0 +1,67 @@
+// Module included in the following assemblies:
+//
+// * cnv/cnv_users_guide/cnv-configuring-local-storage-for-vms.adoc
+
+[id="cnv-using-hostpath-provisioner_{context}"]
+= Using the hostpath provisioner to enable local storage
+
+To deploy the hostpath provisioner and enable your virtual machines to use local
+storage, first create a HostPathProvisioner custom resource.
+
+.Prerequisites
+
+* Create a backing directory on each node for the PersistentVolumes (PVs)
+that the hostpath provisioner creates.
+
+* Apply the SELinux context `container_file_t` to the PV
+backing directory on each node. For example:
++
+----
+$ sudo chcon -t container_file_t -R </path/to/backing/directory>
+----
++
+[NOTE]
+====
+If you use Red Hat Enterprise Linux CoreOS 8 workers, you must configure SELinux
+by using a MachineConfig manifest instead.
+====
+
+.Procedure
+
+. Create the HostPathProvisioner custom resource file. For example:
++
+----
+$ touch hostpathprovisioner_cr.yaml
+----
+
+. Edit the file, ensuring that the `spec.pathConfig.path` value is the directory
+where you want the hostpath provisioner to create PVs. For example:
++
+[source,yaml]
+----
+apiVersion: hostpathprovisioner.kubevirt.io/v1alpha1
+kind: HostPathProvisioner
+metadata:
+  name: hostpath-provisioner
+spec:
+  imagePullPolicy: IfNotPresent
+  pathConfig:
+    path: "</path/to/backing/directory>" <1>
+    useNamingPrefix: "false" <2>
+----
+<1> Specify the backing directory where you want the provisioner to create PVs.
+<2> Change this value to `true` if you want to use the name of the PersistentVolumeClaim (PVC)
+that is bound to the created PV as the prefix of the directory name.
++
+[NOTE]
+====
+If you did not create the backing directory, the provisioner attempts to create
+it for you. If you did not apply the `container_file_t` SELinux context, this can cause
+`Permission denied` errors.
+====
+
+. Create the custom resource in the `openshift-cnv` namespace:
++
+----
+$ oc create -f hostpathprovisioner_cr.yaml -n openshift-cnv
+----
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -1310,6 +1310,8 @@ Topics:
     File: cnv-uploading-local-disk-images-virtctl
   - Name: Uploading a local disk image to a block storage DataVolume
     File: cnv-uploading-local-disk-images-block
+  - Name: Configuring local storage for virtual machines
+    File: cnv-configuring-local-storage-for-vms
   - Name: Expanding virtual storage by adding blank disk images
     File: cnv-expanding-virtual-storage-with-blank-disk-images
   - Name: Preparing CDI scratch space
--- a/cnv/cnv_users_guide/cnv-configuring-local-storage-for-vms.adoc
+++ b/cnv/cnv_users_guide/cnv-configuring-local-storage-for-vms.adoc
@@ -1,0 +1,17 @@
+[id="cnv-configuring-local-storage-for-vms"]
+= Configuring local storage for virtual machines
+include::modules/cnv-document-attributes.adoc[]
+:context: cnv-configuring-local-storage-for-vms
+toc::[]
+
+
+You can configure local storage for your virtual machines by using the hostpath
+provisioner feature.
+
+include::modules/cnv-about-hostpath-provisioner.adoc[leveloffset=+1]
+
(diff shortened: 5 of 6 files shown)


modules/cnv-using-hostpath-provisioner.adoc
```diff
@@ -0,0 +1,57 @@
...
+  imagePullPolicy: IfNotPresent
+  pathConfig:
+    path: "</path/to/backing/directory>"
+    useNamingPrefix: "false"
+----
++
+[NOTE]
+====
+If you did not create the backing directory, the provisioner will attempt to create
+it for you. This can cause `Permission denied` errors.
```
awels: it can cause permission denied errors IF the user forgets to run the SELinux command. There will be some additional text with regards to a RHCOS 8 worker, as those require a MachineConfig to setup the SELinux. The yaml for an RHCOS worker looks like this:
```yaml
apiVersion: machineconfiguration.openshift.io/v1
kind: MachineConfig
metadata:
  name: 50-set-selinux-for-hostpath-provisioner
  labels:
    machineconfiguration.openshift.io/role: worker
spec:
  config:
    ignition:
      version: 2.2.0
    systemd:
      units:
        - contents: |
            [Unit]
            Description=Set SELinux chcon for hostpath provisioner
            Before=kubelet.service

            [Service]
            ExecStart=/usr/bin/chcon -Rt container_file_t /var/hpvolumes

            [Install]
            WantedBy=multi-user.target
          enabled: true
          name: hostpath-provisioner.service
```

modules/cnv-about-hostpath-provisioner.adoc
```diff
@@ -0,0 +1,18 @@
...
+[id="cnv-about-hostpath-provisioner_{context}"]
+= About the hostpath provisioner
+
+The hostpath provisioner is a local storage provisioner designed for
+{CNVProductName}. If you want to configure local storage for
+virtual machines, you must enable the hostpath provisioner first.
+
+The Operator for the hostpath provisioner is installed by default with the
+{CNVProductName} Operator. The hostpath provisioner Operator deploys the
+provisioner itself, and it creates the required `StorageClass` object.
```
awels: It does not create the storage class object either, since there are 2 possible options for the binding mode, we decided to not manage the storage class with the operator. So the user will have to create the storage class as well. https://github.com/kubevirt/hostpath-provisioner-operator#storage-class has some more information on this.

ousleyp: @awels Oh, I thought that users running CNV 2.2 would have to be running Kubernetes >= 1.12 since they would not be running OCP 3.x. In November you had said that the hostpath provisioner operator creates the storage class for you; can you confirm that I should add info about the two different options?

Thanks!

modules/cnv-using-hostpath-provisioner.adoc
ousleyp: @awels I did not make changes related to the StorageClass yet because of my outstanding question, but I did make the other changes if you'd like to review them :)

modules/cnv-about-hostpath-provisioner.adoc
awels: Well technically there is nothing stopping people from running cnv 2.2 on OS 3.11, so that is why we had to give them the option. We don't recommend it, but its not impossible.

awels: So in November this was the case, but after some thought, we decided that we should also support 3.11 in case people wanted to run the hostpath provisioner on OS 3.11. Also other provisioners usually don't have the operator manage the storage class. And last but not least, we wanted to give the user the option to name the storage class differently from the default.

So I would add a section similar to the CR, where you have an example yaml and point out the things they can customize (the volumeBindingMode and name).

ousleyp: @awels Thanks for explaining the reasoning behind that decision. :) I will include info about setting the storage class binding mode to `WaitForFirstConsumer`, but because we really only recommend CNV 2.2 on OCP 4+ (if not 4.3), I am going to omit the info about the `Immediate` binding mode. I also ran this by @djfjeff and he agrees that we should only include info relevant to OCP 4+.

awels: Agreed, I just wanted to make you aware of the reasoning and options so you could write the best possible documentation.

ousleyp: Hi @awels and @alitke - I have made some changes that I would like you to review (mostly adding info about creating the storage class). Here's the latest preview build: http://file.bos.redhat.com/pousley/121719/hostpath-cdi/cnv/cnv_users_guide/cnv-configuring-local-storage-for-vms.html

And here's the commit with most of the changes: https://github.com/openshift/openshift-docs/pull/18445/commits/b0c76c831e37b7b4bb98edb019a5c5a0b3391dfe

Please let me know if you think the info is in the right order, in particular, and if there are any inaccuracies. There are some references to kubevirt (API versions, etc.) so please let me know if any of those need to be replaced with downstream versions (and what the replacements should be).

Thanks so much!

awels:
Very nice, looks good to me.

ousleyp:
> Very nice, looks good to me.

Thank you @awels, I'm glad to hear that :) I'll set the Jira to QE Review.

modules/cnv-creating-storage-class.adoc
```diff
@@ -0,0 +1,46 @@
...
+----
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: storage-class-name
+provisioner: <path/to/provisioner>
+reclaimPolicy: Delete <1>
+volumeBindingMode: WaitForFirstConsumer
+----
+<1> The two possible `reclaimPolicy` values are `Delete` and `Retain`. If you
```
qwang1: "reclaimPolicy" is noted here, "name" and "volumeBindingMode" are noted on the "Example: StorageClass YAML for the hostpath provisioner". Why don't we gather these tips in the same place?

modules/cnv-example-storage-class-yaml.adoc
```diff
@@ -0,0 +1,31 @@
...
+provisioner: kubevirt.io/hostpath-provisioner
+reclaimPolicy: Delete
+volumeBindingMode: WaitForFirstConsumer <2>
+----
+<1> You can optionally rename the storage class by changing this value.
+<2> The `volumeBindingMode` value determines when dynamic provisioning and volume
+binding occur. Specify `WaitForFirstConsumer` to delay the binding and provisioning
+of a PersistentVolume (PV) until after a Pod that uses the PersistentVolumeClaim (PVC)
+is created. This ensures that the PV meets the Pod's scheduling requirements.
+The `WaitForFirstConsumer` setting is not available in {product-title} 3.x.
```
qwang1: Should "Immediate" volumeBindingMode be mentioned and make a distinction with "WaitForFirstConsumer"?

ousleyp: @qwang1 Hi Qixuan, I intentionally didn't include `Immediate` because IIUC it is only used with OpenShift 3.x, and we don't recommend using CNV 2.2 with OCP 3.x.

modules/cnv-creating-storage-class.adoc
ousleyp: Great point :) I am updating the PR now.

ousleyp: @qwang1 Here's the preview of the updated section: http://file.bos.redhat.com/pousley/010220/hostpath-cdi/cnv/cnv_users_guide/cnv-configuring-local-storage-for-vms.html#cnv-example-storage-class-yaml_cnv-configuring-local-storage-for-vms

I'm glad you caught that!

modules/cnv-example-storage-class-yaml.adoc
qwang1: @ousleyp Thanks, good to know

modules/cnv-about-hostpath-provisioner.adoc
```diff
@@ -0,0 +1,23 @@
...
+= About the hostpath provisioner
+
+The hostpath provisioner is a local storage provisioner designed for
+{CNVProductName}. If you want to configure local storage for
+virtual machines, you must enable the hostpath provisioner first.
+
+When you install the {CNVProductName} Operator, the hostpath provisioner Operator
+is automatically installed. To use it, you must:
+
+* apply the SELinux label `container_file_t` to the PersistentVolume (PV) backing
```
aburdenthehand: Nit: These bullets should have tops and tails (caps and fullstop)

modules/cnv-about-hostpath-provisioner.adoc
```diff
@@ -0,0 +1,23 @@
...
+directory on each node
+* create a `HostPathProvisioner` custom resource
+* create a `StorageClass` object for the provisioner in the `openshift-cnv` namespace
+
+The hostpath provisioner Operator deploys the provisioner as a `DaemonSet` on each
```
aburdenthehand: I don't think DaemonSet needs markup. 
https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/

modules/cnv-example-storage-class-file.adoc
```diff
@@ -0,0 +1,33 @@
...
+volumeBindingMode: WaitForFirstConsumer <3>
+----
+<1> You can optionally rename the storage class by changing this value.
+<2> The two possible `reclaimPolicy` values are `Delete` and `Retain`. If you
+do not specify a value, the storage class defaults to `Delete`.
+<3> The `volumeBindingMode` value determines when dynamic provisioning and volume
+binding occur. Specify `WaitForFirstConsumer` to delay the binding and provisioning
+of a PersistentVolume (PV) until after a Pod that uses the PersistentVolumeClaim (PVC)
+is created. This ensures that the PV meets the Pod's scheduling requirements.
+The `WaitForFirstConsumer` setting is not available in {product-title} 3.x.
```
aburdenthehand: Not sure if this is relevant. Can we install CNV 2.2 on OpenShift 3.x?

modules/cnv-example-cr-hostpath-provisioner.adoc
```diff
@@ -0,0 +1,30 @@
...
+kind: HostPathProvisioner
+metadata:
+  name: hostpath-provisioner
+spec:
+  imagePullPolicy: IfNotPresent
+  pathConfig:
+    path: "</path/to/backing/directory>" <1>
+    useNamingPrefix: "false" <2>
+----
+<1> Specify the backing directory where you want the provisioner to create PersistentVolumes (PV).
```
aburdenthehand: I think if you add these to the example in the procedure itself we don't need the example at all. It seems to be the only difference.

modules/cnv-example-storage-class-file.adoc
```diff
@@ -0,0 +1,33 @@
...
+----
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: hostpath-provisioner <1>
+provisioner: kubevirt.io/hostpath-provisioner
+reclaimPolicy: Delete <2>
+volumeBindingMode: WaitForFirstConsumer <3>
+----
+<1> You can optionally rename the storage class by changing this value.
```
aburdenthehand: As above (add these to the example in procedure)

aburdenthehand:
Some nits, and some structural changes (as discusses) but otherwise LGTM!

ousleyp:
@qwang1 Hi Qixuan, please review the new draft when you can. I made the MachineConfig section into a procedure and restructured the doc to remove the redundancies. 

You can see the changes in this commit (I won't squash it yet): https://github.com/openshift/openshift-docs/pull/18445/commits/28263b50766ed4aa83fe44899d631339449fdab1

Preview Build: http://file.bos.redhat.com/pousley/010620/hostpath-cdi/cnv/cnv_users_guide/cnv-configuring-local-storage-for-vms.html

Thank you!

modules/cnv-creating-storage-class.adoc
```diff
@@ -0,0 +1,51 @@
...
+do not specify a value, the storage class defaults to `Delete`.
+<3> The `volumeBindingMode` value determines when dynamic provisioning and volume
+binding occur. Specify `WaitForFirstConsumer` to delay the binding and provisioning
+of a PV until after a Pod that uses the PersistentVolumeClaim (PVC)
+is created. This ensures that the PV meets the Pod's scheduling requirements.
+
+. Create the StorageClass object:
++
+----
+$ oc create -f storageclass.yaml -n <namespace>
```
qwang1: 'StorageClass' object is cluster scoped, it doesn't need namespace.

modules/cnv-about-hostpath-provisioner.adoc
```diff
@@ -0,0 +1,26 @@
...
+When you install the {CNVProductName} Operator, the hostpath provisioner Operator
+is automatically installed. To use it, you must:
+
+* Configure SELinux:
+** If you use Red Hat Enterprise Linux CoreOS 8 workers, you must create a MachineConfig
+object on each node.
+** Otherwise, apply the SELinux label `container_file_t` to the PersistentVolume (PV) backing
+directory on each node.
+* Create a HostPathProvisioner custom resource.
+* Create a StorageClass object for the provisioner in the `openshift-cnv` namespace.
```
qwang1: 'StorageClass' object is cluster scoped, it doesn't need namespace.

modules/cnv-creating-storage-class.adoc
```diff
@@ -0,0 +1,51 @@
...
++
+[source,yaml]
+----
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: hostpath-provisioner <1>
+provisioner: kubevirt.io/hostpath-provisioner
+reclaimPolicy: Delete <2>
+volumeBindingMode: WaitForFirstConsumer <3>
```
maya-r: we probably need to add `volumeMode: Filesystem` here.
[HCO installs a configmap](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/5e2b026673244c92cbd148badbf00c9c82001578/pkg/controller/hyperconverged/hyperconverged_controller.go#L1045-L1074) causing the default volumeMode to be Block in some cases*.

This is one place where we can override this user interface default.

*"Bare metal platform"

awels: If unspecified the default is FileSystem, so no need to specify unless you want something besides the default.

ousleyp: Based on @awels comment, I am resolving this without changes.

modules/cnv-creating-storage-class.adoc
ousleyp: Fixed.

modules/cnv-about-hostpath-provisioner.adoc
ousleyp: Fixed.

qwang1:
LGTM，thanks Pan！

PR #18458, opened 2019-12-05
bug 1763799 updating oc adm release extract command

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1763799


kalexand-rh:

PR #18466, opened 2019-12-05
BZ1776095: Update prereq important note

bobfuru:
As described in [BZ 1776095](https://bugzilla.redhat.com/show_bug.cgi?id=1776095), this has been reviewed and approved by customer and TAM.
@bmcelvee PTAL for peer review
Thanks!

### DIFF
--- a/modules/registry-configuring-storage-vsphere.adoc
+++ b/modules/registry-configuring-storage-vsphere.adoc
@@ -14,13 +14,11 @@ registry to use storage.
 
 * Cluster administrator permissions.
 * A cluster on VMware vSphere.
-* A provisioned persistent volume (PV) with `ReadWriteMany` access mode, such as
-`NFS`.
+* A provisioned persistent volume (PV) with `ReadWriteMany` access mode.
 +
 [IMPORTANT]
 ====
-vSphere volumes do not support the `ReadWriteMany` access mode. You must use
-a different storage backend, such as `NFS`, to configure the registry storage.
+vSphere volumes do not support the `ReadWriteMany` access mode. You must use a different storage backend to configure the registry storage.
 ====
 +
 * Must have "100Gi" capacity.


bobfuru:
[BZ 1776095](https://bugzilla.redhat.com/show_bug.cgi?id=1776095)
@wewang58 PTAL for QE Review. Thanks!

wzheng1:
I don't think registry will support RWO accessmode: https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/storage/pvc/pvc.go#L113-L119

bobfuru:
> https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/storage/pvc/pvc.go#L113-L119

My understanding is that: 
RWO is the only supported access mode ;
AND to make it work, you must set number of replicas to `1` ;
AND you must manually create PV/PVC with RWO ;
ELSE you must use a different storage backend (not vSphere volume).

^^ This was suggested workaround by customer and TAM. If this works, I'll edit the note to more clearly state, such as:

[IMPORTANT]
====
vSphere volumes only support `ReadWriteOnce` access mode. To enable, you must set the number of replicas to `1`. Also, you must create the PV and persistent volume claim (PVC) manually with `ReadWriteOnce` access mode.
Otherwise, you must use a different storage backend to configure the registry storage.
====

@wzheng1 Will the above work?

bmcelvee:
@dmage or @coreydaley would you mind giving us some guidance on this one? Thanks!

dmage:
When the deployment for the registry gets changed, the controller first deploys a new version and then shutdowns old pods. So even if you have only one replica, you still need RWX access mode. And the image registry operator rejects volume claims that don't provide RWX.

bobfuru:
> When the deployment for the registry gets changed, the controller first deploys a new version and then shutdowns old pods. So even if you have only one replica, you still need RWX access mode. And the image registry operator rejects volume claims that don't provide RWX.

Thanks for your input, @dmage! I've updated the [BZ 1776095](https://bugzilla.redhat.com/show_bug.cgi?id=1776095). We'll plan to leave the note in that states vSphere volume requires RWX.

bmcelvee:
LGTM

bmcelvee (approved):

bobfuru:
No doc changes needed, as this ended up being an RFE for Eng. (https://bugzilla.redhat.com/show_bug.cgi?id=1782580) Closing PR.

PR #18470, opened 2019-12-05
bz-1780401 ETCD restore is not clear causing mistakes during the proc…

vwalek:
the etcdctl should be installed with etcd rpm package

### DIFF
--- a/day_two_guide/topics/proc_restoring-etcd.adoc
+++ b/day_two_guide/topics/proc_restoring-etcd.adoc
@@ -273,7 +273,18 @@ Before restoring etcd on a static pod:
 * `etcdctl` binaries must be available or, in containerized installations,
 the `rhel7/etcd` container must be available.
 +
-You can obtain etcd by running the following commands:
+You can install the `etcdctl` binary with the etcd package by running the following commands:
++
+----
+# yum install etcd  
+----
+The package also installs the systemd service. Disable and mask the service as it shouldn't start when etcd runs in static pod.
++
+----
+# systemctl disable etcd.service 
+# systemctl mask etcd.service
+----
+You can obtain etcd binary directly from upstream by running the following commands.
 +
 ----
 $ git clone https://github.com/coreos/etcd.git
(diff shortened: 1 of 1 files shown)


day_two_guide/topics/proc_restoring-etcd.adoc
```diff
@@ -273,14 +273,17 @@ Before restoring etcd on a static pod:
...
 +
-You can obtain etcd by running the following commands:
+You can install the `etcdctl` binary with the etcd package by running the following commands:
 +
 ----
-$ git clone https://github.com/coreos/etcd.git
-$ cd etcd
-$ ./build
+# yum install etcd  
+----
```
kalexand-rh: You need to add a line with a "+" on it, like line 277, so this will render correctly.

day_two_guide/topics/proc_restoring-etcd.adoc
```diff
@@ -273,14 +273,17 @@ Before restoring etcd on a static pod:
...
+The package also installs the systemd service. Disable and mask the service as it shouldn't start when etcd runs in static pod.
```
kalexand-rh: s/as it shouldn't start/because it is not required

kalexand-rh (commented):
@vwalek, will you make two small changes and squash?

kalexand-rh:
@geliu2016, will you PTAL?

geliu2016:
@kalexand-rh LGTM, thx

kalexand-rh:
Thank you @geliu2016!

@vwalek, if make two small changes and squash, I'll merge.

day_two_guide/topics/proc_restoring-etcd.adoc
vwalek: yeah, but we won't to make sure that the service is not started manually by mistake. it is not required, but if not masked it can be started. The disable is to make sure it will not start if the system is rebooted. It is more for safety.

vwalek:
I will open new PR for that.

PR #18481, opened 2019-12-06
Improve visibility of support tools

adambkaplan:
Move Support topic to be below the Updating topics.

Hanging or failing upgrades is the most frequent scenario where running `oc adm must-gather` is useful. Moving "Support" to be below this section makes it easier for customers and the community to get help by running the must-gather tool by placing this info in a more prominent location.

This section is currently at the very bottom of our topic links, which makes it easy for individuals to miss it. There have been instances where developers who work on OpenShift have not been aware of this tool.

### DIFF
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -203,6 +203,26 @@ Topics:
 # - Name: Troubleshooting an update
 #  File: updating-troubleshooting
 ---
+Name: Support
+Dir: support
+Distros: openshift-enterprise,openshift-online,openshift-dedicated
+Topics:
+- Name: Getting support
+  File: getting-support
+- Name: Gathering data about your cluster
+  File: gathering-cluster-data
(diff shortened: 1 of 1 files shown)


kalexand-rh:
@aheslin, PTAL?

vikram-redhat:
Interesting idea. @sferich888 - thoughts?

Also: @bergerhoffer - I vaguely recall discussing this placement, but I can't remember why we settled here.

sferich888:
Ok with me. Let's get PMs approval too.

ahardin-rh (approved):
LGTM

bmcelvee (approved):

bergerhoffer:
No objections from me on moving if it will help our users find it more easily when they need it.

@vikram-redhat Should we go ahead and move it all the way back to 4.1?

vikram-redhat:
@bergerhoffer yeah works for me.

PR #17824, opened 2019-11-01
CNV_3044 Add Disk to VM

bgaydosrh:
Created new module for adding disks to VMs, conditionalized to reuse for both add to VM and VM templates.

Peer review requested from @aburdenthehand to ensure my conditionals and associated logic and syntax is valid.

Also content questions regarding GUI and associated functionality will be sent to Adam Litke and Tomas Jellinek.

Label **Peer Review Needed** and **enterprise-4.3**.

Many thanks!

Bob

PR #18483, opened 2019-12-06
CNV_3044 Add Disk to VM

bmcelvee:
CP from https://github.com/openshift/openshift-docs/pull/17824. @bgaydosrh - FYI

### DIFF
--- a/modules/cnv-add-disk-to-vm.adoc
+++ b/modules/cnv-add-disk-to-vm.adoc
@@ -1,0 +1,45 @@
+// Module included in the following assemblies:
+//
+// * cnv/cnv_users_guide/cnv-edit-vms.adoc
+// * cnv/cnv_users_guide/cnv-editing-vm-template.adoc
+
+// Establishing conditionals so content can be re-used for editing VMs
+// and VM templates.
+
+ifeval::["{context}" == "cnv-edit-vms"]
+:cnv-vm-add-disk:
+:object: virtual machine
+:object-gui: Virtual Machines
+endif::[]
+
+ifeval::["{context}" == "cnv-editing-vm-template"]
+:cnv-vm-add-disk-template:
+:object: virtual machine template
+:object-gui: Virtual Machine Templates
+endif::[]
+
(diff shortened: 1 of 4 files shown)


bmcelvee:
I'm not sure where the error is with this one.

bmcelvee:
@vikram-redhat - this cherrypick is giving us a bit of trouble. Would you mind helping us troubleshoot?

modules/cnv-add-disk-to-vm.adoc
```diff
@@ -0,0 +1,45 @@
...
+// and VM templates.
+
+// ifeval::["{context}" == "cnv-edit-vms"]
+// :cnv-vm-add-disk:
+// :object: virtual machine
+// :object-gui: Virtual Machines
+// endif::[]
+
+ifeval::["{context}" == "cnv-editing-vm-template"]
+:cnv-vm-add-disk-template:
```
aburdenthehand: This ifdef is declared but not used (I'm pretty sure asciibinder dislikes that). We can remove this line.

modules/cnv-add-disk-to-vm.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+// :object!:
+// :object-gui!:
+
+// ifeval::["{context}" == "cnv-edit-vms"]
+// :cnv-vm-add-disk!:
+// endif::[]
+
+ifeval::["{context}" == "cnv-editing-vm-template"]
+:cnv-vm-template!:
```
aburdenthehand: We can get rid of this too.
It also doesn't match what was declared earlier (and we're getting rid of that anyway)

modules/cnv-add-disk-to-vm.adoc
bmcelvee: Thanks, @aburdenthehand! I'll give it a go for both of your suggestions :)

bmcelvee:
Andrew dug into this one a bit more and it sounds like the best path forward is to close this PR and make the changes on a PR specifically for 4.2. Thanks so much for looking into it!

PR #18492, opened 2019-12-06
added ossm images

neal-timpe:
These images seem to have been left out by mistake and are resulting in broken images in the ossm docs.


JStickler:
Addresses BZ 1780340.

LGTM.

PR #18505, opened 2019-12-08
Bug 1780519: [DOCS] Unclear "node" term for restarting "atomic-openshift-node.service"

bysnupy:
* Version: v3.11

* Bugzilla: [[DOCS] Unclear "node" term for restarting "atomic-openshift-node.service"](Reference: https://bugzilla.redhat.com/show_bug.cgi?id=1780519)

* Description:
  "restarts the appropriate nodes" can misunderstand as "restart node hosts", so added "atomic-openshift-node.service" for suppressing confused.

### DIFF
--- a/admin_guide/manage_nodes.adoc
+++ b/admin_guide/manage_nodes.adoc
@@ -523,7 +523,7 @@ During installation, {product-title} creates a configmap in the *openshift-node*
 To make configuration changes to an existing node, edit the appropriate configuration map. A _sync pod_ on each node
 watches for changes in the configuration maps. During installation, the sync pods are created by using _sync Daemonsets_,
 and a *_/etc/origin/node/node-config.yaml_* file, where the node configuration parameters reside, is added to each node. When a sync pod
-detects configuration map change, it updates the *_node-config.yaml_* on all nodes in that node group and restarts the appropriate nodes.
+detects configuration map change, it updates the *_node-config.yaml_* on all nodes in that node group and restarts the appropriate nodes `atomic-openshift-node.service`.
 
 ----
 $ oc get cm -n openshift-node


admin_guide/manage_nodes.adoc
```diff
@@ -523,7 +523,7 @@ During installation, {product-title} creates a configmap in the *openshift-node*
 To make configuration changes to an existing node, edit the appropriate configuration map. A _sync pod_ on each node
 watches for changes in the configuration maps. During installation, the sync pods are created by using _sync Daemonsets_,
 and a *_/etc/origin/node/node-config.yaml_* file, where the node configuration parameters reside, is added to each node. When a sync pod
-detects configuration map change, it updates the *_node-config.yaml_* on all nodes in that node group and restarts the appropriate nodes.
+detects configuration map change, it updates the *_node-config.yaml_* on all nodes in that node group and restarts the appropriate nodes `atomic-openshift-node.service`.
```
kalexand-rh: Does the service restart on all of the nodes in the node group? If so, I'd say "it both updates the \*_node-config.yaml\_* file and restarts the \`atomic-openshift-node.service\` process on all nodes in that node group.

If it doesn't, I'd s/restarts the appropriate nodes \`atomic-openshift-node.service\`/restarts the \`atomic-openshift-node.service\` process on the nodes that require it.

If we know exactly which subset of nodes need to be restarted, I might replace  "that require it" with a description of those nodes.

bysnupy: > Does the service restart on all of the nodes in the node group?

@kalexand-rh  Yes, it does. So I think "it both updates the *_node-config.yaml_* file and restarts the `atomic-openshift-node.service` process on all nodes in that node group." is appropriate on it.

kalexand-rh: @bysnupy, is updating a node group configmap different than updating the node-config.yaml file?

bysnupy: @kalexand-rh 
> is updating a node group configmap different than updating the node-config.yaml file?

It's same each other, but we warn not to update node-config.yaml directly due to overwriting as node group configmap.

1. update a node group configmap
2. updated node-config.yaml  as changed configurations by a sync pod
3. restarted atomic-openshift-node.service

vikram-redhat:
Closing as changes were applied through https://github.com/openshift/openshift-docs/pull/19174

PR #18510, opened 2019-12-09
modules/upi: Document that RHCOS is RHEL8

cgwalters:
So people know it has the same requirements/limits.

### DIFF
--- a/modules/installation-requirements-user-infra.adoc
+++ b/modules/installation-requirements-user-infra.adoc
@@ -48,6 +48,9 @@ endif::ibm-z[]
 The bootstrap, control plane, and compute machines must use the {op-system-first} as the
 operating system.
 
+Note that {op-system} is based on Red Hat Enterprise Linux 8 and inherits all of its hardware certifications and requirements.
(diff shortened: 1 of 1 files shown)


modules/installation-requirements-user-infra.adoc
```diff
@@ -80,6 +80,9 @@ Each cluster machine must meet the following minimum requirements:
 
 |===
 
+Note that {op-system} is RHEL 8, and inherits all hardware certifications and requirements.
```
kalexand-rh:
```suggestion
Note that {op-system} is based on Red Hat Enterprise Linux 8 and inherits all of its hardware certifications and requirements.
```

modules/installation-requirements-user-infra.adoc
```diff
@@ -80,6 +80,9 @@ Each cluster machine must meet the following minimum requirements:
...
+See the link:https://access.redhat.com/articles/rhel-limits[Red Hat Enterprise Linux technology capabilities and limits ] for more information.
```
kalexand-rh:
```suggestion
See link:https://access.redhat.com/articles/rhel-limits[Red Hat Enterprise Linux technology capabilities and limits].
```

PR #18523, opened 2019-12-09
adding release notes link

kalexand-rh:
Adding a link to the release notes from the cluster customization page. (And doing it backwards so that everything goes live.)

### DIFF
--- a/installing/install_config/customizations.adoc
+++ b/installing/install_config/customizations.adoc
@@ -10,6 +10,11 @@ _configuration resources_
 //and _Custom Resources_
 are available.
 
+[NOTE]
+====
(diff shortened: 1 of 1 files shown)


vikram-redhat:
Hey @kalexand-rh - this probably didn't need to go into master (not sure). In any case, I have undone this with https://github.com/openshift/openshift-docs/pull/18635/files#diff-162300a9be635b34c5f6aeef7585d971R15, with a bogus anchor id. Let me know if I have got this wrong.

kalexand-rh:
Sorry, Vikram. Thank you for the bogus anchor ID! Is there a better strategy to use in case an issue like this one comes up again?

PR #17685, opened 2019-10-28
Bump OCP 4.3 to cri-o 1.16

mburke5678:
https://jira.coreos.com/browse/OSDOCS-735

PR #18530, opened 2019-12-10
Reverting node output changes and just keeping version bump

bergerhoffer:
Reverting update to this file from #17685 and just keeping the version bump change in the output.


bergerhoffer:
@mburke5678 Can you review please?

mburke5678:
@bergerhoffer LGTM

Issue #17796, opened 2019-10-31
OCP 4.3 Release Notes Tracker


ahardin-rh:
Please leave comments here for anything that should be highlighted in the 4.3 release notes. Thank you! 

rendered draft of 4.3 release notes: https://docs.openshift.com/container-platform/4.3/release_notes/ocp-4-3-release-notes.html

tnozicka:
apps/v1beta1, apps/v1beta2, extensions/v1beta1 (except ingresses) are deprecated in 4.3 (were deprecated upstream for some time) and will be removed in 4.4.  Kubernetes 1.16 (in OCP 4.3) has already disabled serving those, but we have re-enabled it back for 4.3 release.

ref https://jira.coreos.com/browse/WRKLDS-16

According to @smarterclayton we should write a section in docs with "master list of deprecations per release somewhere" and reference it from release notes. cc @bergerhoffer

mburke5678:
Kubernetes is upgraded to version 1.16 from version 1.14.
[Kubernetes change log for 1.15](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.15.md#kubernetes-v115-release-notes)
[Kubernetes change log for 1.16](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.16.md#kubernetes-v1160-release-notes)

bobfuru:
**RAW BLOCK VOLUME SUPPORT** (`ocp-4-3-storage-devices`)
iSCSI raw block volumes are now fully supported with OpenShift Container Platform 4.3.

Raw block volumes using Cinder are now in Technology Preview.

In `ocp-4-3-technology-preview` table: 

- "Raw Block with Cinder" moves to TP in 4.3
- "Raw Block with iSCSI" moves from TP (4.2) to GA (4.3)

jboxman:
OVN SDN is GA in 4.3. (Was TP.)

vikram-redhat:
Auto rotation of certs coming: https://jira.coreos.com/browse/PROD-970?focusedCommentId=144768&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-144768

bobfuru:
Open a Support Case from the {product-title} web console.
https://jira.coreos.com/browse/OSDOCS-764

s-urbaniak:
[...]

PR #18542, opened 2019-12-10
Add etcd data encryption info to release notes

codyhoag:
Gathered from issue tracker #17796

@bergerhoffer can you verify this addition? I mostly extracted info from your tagged 4.3 article to give a brief summary.

### DIFF
--- a/release_notes/ocp-4-3-release-notes.adoc
+++ b/release_notes/ocp-4-3-release-notes.adoc
@@ -77,6 +77,35 @@ to customers based on data from the Red Hat Service Reliability Engineering
 (SRE) team, you might not immediately see notification in the web console that
 updates from version 4.2.z to 4.3 are available at initial release.
 
+[id="ocp-4-3-security"]
+=== Security
+
+[id="ocp-4-3-cert-rotation"]
+==== Automatic rotation of certificates
+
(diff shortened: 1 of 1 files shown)


release_notes/ocp-4-3-release-notes.adoc
```diff
@@ -77,6 +77,29 @@ to customers based on data from the Red Hat Service Reliability Engineering
 (SRE) team, you might not immediately see notification in the web console that
 updates from version 4.2.z to 4.3 are available at initial release.
 
+[id="ocp-4-3-authentication"]
+=== Authentication
```
bergerhoffer: I would consider making this a "Security" heading (I looked back and saw we had that heading in 4.1). And then maybe you could consider moving the item under "Certificates" below into this section as well. wdyt?

bergerhoffer (commented):
Looks good! Just one suggestion to consider.

release_notes/ocp-4-3-release-notes.adoc
codyhoag: I'm glad you brought that up. I was on the fence about what section to categorize this under :sweat_smile:.  I think **Security** makes a lot of sense; I'll move forward with making that change. Thanks!

PR #18543, opened 2019-12-10
Add proc for using tolerations on local-storage-operator Pods

bobfuru:
[BZ 1781233](https://bugzilla.redhat.com/show_bug.cgi?id=1781233)
Borrowing some content from https://docs.openshift.com/container-platform/4.2/logging/config/cluster-logging-kibana.html#cluster-logging-kibana-tolerations_cluster-logging-kibana to explain taint terminology and usage. 

Doc update to be merged to master and CP to enterprise-4.3.

### DIFF
--- a/modules/persistent-storage-local-tolerations.adoc
+++ b/modules/persistent-storage-local-tolerations.adoc
@@ -1,0 +1,48 @@
+// Module included in the following assemblies:
+//
+// storage/persistent-storage/persistent-storage-local.adoc
+
+[id="local-tolerations_{context}"]
+= Using tolerations with Local Storage Operator Pods
+
+Taints can be applied to nodes to prevent them from running general workloads. To allow the Local Storage Operator to use tainted nodes, you must add tolerations to the Pod or DaemonSet definition. This allows the created resources to run on these tainted nodes.
+
+You apply tolerations to the Local Storage Operator Pod through the LocalVolume resource
+and apply taints to a node through the node specification. A taint on a node instructs the node to repel all Pods that do not tolerate the taint. Using a specific taint that is not on other Pods ensures that the Local Storage Operator Pod can also run on that node.
+
+[IMPORTANT]
+====
+Taints and tolerations consist of a key, value, and effect. As an argument, it is expressed as `key=value:effect`. An operator allows you to leave one of these parameters empty.
+====
+
+.Prerequisites
+
+* The Local Storage Operator is installed.
+
+* Local disks are attached to {product-title} nodes with a taint.
+
+* Tainted nodes are expected to provision local storage.
+
+.Procedure
+To configure local volumes for scheduling on tainted nodes:
+
+. Modify the YAML file that defines the Pod and add the `LocalVolume` spec, as shown in the following example:
++
+[source, yaml]
+----
+  apiVersion: "local.storage.openshift.io/v1"
+  kind: "LocalVolume"
+  metadata:
+    name: "local-disks"
+    namespace: "local-storage"
+  spec:
+    tolerations:
+      - key: localstorage <1>
+        operator: Equal <2>
+        value: "localstorage" <3>
+----
+<1> Specify the key that you added to the node.
+<2> Specify the `Equal` operator to require the `key`/`value` parameters to match. If operator is 'Exists`, the system checks that the key exists and ignores the value. If operator is `Equal`, then the key and value must match.
+<3> Specify the value `local` of the tainted node.
+
+The defined tolerations will be passed to the resulting DaemonSets, allowing the diskmaker and provisioner Pods to be created for nodes that contain the specified taints.
--- a/storage/persistent-storage/persistent-storage-local.adoc
+++ b/storage/persistent-storage/persistent-storage-local.adoc
@@ -28,6 +28,8 @@ include::modules/persistent-storage-local-pvc.adoc[leveloffset=+1]
 
 include::modules/persistent-storage-local-pod.adoc[leveloffset=+1]
 
+include::modules/persistent-storage-local-tolerations.adoc[leveloffset=+1]
+
 == Deleting the Local Storage Operator's resources
 
 include::modules/persistent-storage-local-removing-devices.adoc[leveloffset=+2]


bobfuru:
@huffmanca and @liangxia PTAL

modules/persistent-storage-local-tolerations.adoc
```diff
@@ -0,0 +1,43 @@
...
+You apply tolerations to the Local Storage Operator Pod through the LocalVolume Custom Resource (CR)
+and apply taints to a node through the node specification. A taint on a node is a `key:value pair` that
+instructs the node to repel all Pods that do not tolerate the taint. Using a specific `key:value` pair
+that is not on other Pods ensures that only the Local Storage Operator Pod can run on that node.
+
+.Prerequisites
+
+* The Local Storage Operator is installed.
+
+* Local disks are attached to the {product title} nodes.
```
huffmanca: I think we'd either want to include a Prerequisite saying that a node has been tainted that the Local Storage Operator needs to run on, or modify this line to say that:

> Local disks are attached to {product-title} nodes with a taint, and these nodes are expected to provision local storage.

modules/persistent-storage-local-tolerations.adoc
```diff
@@ -0,0 +1,43 @@
...
+//
+// storage/persistent-storage/persistent-storage-local.adoc
+
+[id="local-tolerations_{context}"]
+= Using tolerations with Local Storage Operator Pods
+
+You can control which nodes the Local Storage Operator Pod runs on and prevent
+other workloads from using those nodes by using tolerations on the Pods.
+
+You apply tolerations to the Local Storage Operator Pod through the LocalVolume Custom Resource (CR)
```
huffmanca: Replace `Custom Resource (CR)` with `resource`. This may be a preference, but `Custom Resource (CR)` is an implementation detail that the user doesn't need to know.

modules/persistent-storage-local-tolerations.adoc
```diff
@@ -0,0 +1,43 @@
+// Module included in the following assemblies:
+//
+// storage/persistent-storage/persistent-storage-local.adoc
+
+[id="local-tolerations_{context}"]
+= Using tolerations with Local Storage Operator Pods
+
+You can control which nodes the Local Storage Operator Pod runs on and prevent
+other workloads from using those nodes by using tolerations on the Pods.
```
huffmanca: This isn't entirely correct. You'd prevent workloads from running on nodes by adding taint values to the nodes. You can then schedule Pods to run those nodes by adding Tolerations to the Pod or DaemonSet definition.

I think what we might want to say here is:

> Taints can be applied to nodes to prevent them from running general workloads. To allow the Local Storage Operator to use these tainted nodes you must define tolerations that allow the created resources to run on these nodes.

modules/persistent-storage-local-tolerations.adoc
```diff
@@ -0,0 +1,43 @@
...
+
+You apply tolerations to the Local Storage Operator Pod through the LocalVolume Custom Resource (CR)
+and apply taints to a node through the node specification. A taint on a node is a `key:value pair` that
+instructs the node to repel all Pods that do not tolerate the taint. Using a specific `key:value` pair
+that is not on other Pods ensures that only the Local Storage Operator Pod can run on that node.
```
huffmanca: I believe the customer would have some workload that they want to reserve from general use. If that workload has local storage, then they need resources from the LSO to be created for their workload.

This is a nitpicky phrasing issue, because the goal isn't that **only** the LSO can run on the node. The goal is that the LSO can **also** run on the node, consuming local storage for whatever workload they're actually running.

modules/persistent-storage-local-tolerations.adoc
```diff
@@ -0,0 +1,43 @@
...
+
+.Prerequisites
+
+* The Local Storage Operator is installed.
+
+* Local disks are attached to the {product title} nodes.
+
+.Procedure
+To allow the LocalVolume CR to select tainted nodes for use:
```
huffmanca: I'd reword this to remove CR:

> To configure local volumes for scheduling on tainted nodes:

modules/persistent-storage-local-tolerations.adoc
```diff
@@ -0,0 +1,43 @@
...
+  apiVersion: "local.storage.openshift.io/v1"
+  kind: "LocalVolume"
+  metadata:
+    name: "local-disks"
+    namespace: "local-storage"
+  spec:
+    tolerations:
+      - key: localstorage <1>
+        operator: Equal <2>
+        value: "local"
```
huffmanca: Since you've added annotations for the first two, you should probably add it for the third as well.

Note that if the operator is `Exists`, then it simply checks that the key exists and ignores the value. If the operator is `Equal`, then the key **and** value must match. It might be worth a mention here?

modules/persistent-storage-local-tolerations.adoc
bobfuru: Amended second bullet and added third bullet in Prereqs.

modules/persistent-storage-local-tolerations.adoc
bobfuru: This comment explains taints and tolerations better than anything else I've read: 

> You'd prevent workloads from running on nodes by adding taint values to the nodes. You can then schedule Pods to run those nodes by adding Tolerations to the Pod or DaemonSet definition.

So I combined that with your suggested text.

modules/persistent-storage-local-tolerations.adoc
bobfuru: Text updated to match bolding (removed "only" and added "also").

modules/persistent-storage-local-tolerations.adoc
bobfuru: Agree. Added the third annotation and a note to #2.

bobfuru (commented):
@huffmanca I've made a few comments and applied your suggestions. Thanks for your feedback!

modules/persistent-storage-local-tolerations.adoc
```diff
@@ -0,0 +1,45 @@
...
+//
+// storage/persistent-storage/persistent-storage-local.adoc
+
+[id="local-tolerations_{context}"]
+= Using tolerations with Local Storage Operator Pods
+
+Taints can be applied to nodes to prevent them from running general workloads. To allow the Local Storage Operator to use tainted nodes, you must add tolerations to the Pod or DaemonSet definition. This allows the created resources to run on these tainted nodes.
+
+You apply tolerations to the Local Storage Operator Pod through the LocalVolume resource
+and apply taints to a node through the node specification. A taint on a node is a `key:value pair` that
```
liangxia: From `oc adm taint --help`,
A taint consists of a key, value, and effect. As an argument here, it is expressed as key=value:effect.

bobfuru: From https://docs.openshift.com/container-platform/4.2/nodes/scheduling/nodes-scheduler-taints-tolerations.html:
An operator ('Equal' in the following example) allows you to leave one of these parameters empty.

I added this clarification as a note just before the prereqs.

bmcelvee:
LGTM!

modules/persistent-storage-local-tolerations.adoc
liangxia: User will met https://bugzilla.redhat.com/show_bug.cgi?id=1781520

bobfuru: > User will met https://bugzilla.redhat.com/show_bug.cgi?id=1781520

@liangxia I see that this ^^ bug was closed ("will be fixed in master"). So, does this documentation as written look good from QE perspective?

liangxia: Fix in master(currently for 4.4) means it will not fix in 4.3

bobfuru: @liangxia I updated key:value in `tolerations` spec to match, as explained in <2> annotation. So, will it work in QE if it uses the `Equal` operator (as described in the note, "An operator allows you to leave one of these parameters empty")?

Based on BZ1781520, I don't understand if all three (key-value-effect) are required, even with Equal/Exists operator. If so, what do you suggest for doc changes?

liangxia: *  A taint consists of a key, value, and effect. As an argument here, it is expressed as key=value:effect.
  *  The key must begin with a letter or number, and may contain letters, numbers, hyphens, dots, and underscores, up to
253 characters.
  *  Optionally, the key can begin with a DNS subdomain prefix and a single '/', like example.com/my-app
  *  The value is optional. If given, it must begin with a letter or number, and may contain letters, numbers, hyphens,
dots, and underscores, up to  63 characters.
  *  The effect must be NoSchedule, PreferNoSchedule or NoExecute.
  *  Currently taint can only apply to node.

I would like we state here that the taint is "key=value:effect", so customer does not forgot adding the effect. Or else, customer will met the confusing error message describe in above bug.

liangxia: And QE confirmed that `Equal` operator does work as stated below in <2> annotation.

bobfuru: Thanks, Liang. This makes more sense now. I made a couple of small tweaks to the wording so that it states `key=value:effect`. PTAL and confirm that it looks good.

PR #16338, opened 2019-08-21
Remove VM provisioning vars from official docs

dav1x:
This removes the mention of the provisioning Ansible variables for vSphere. Provisioning is covered under an unsupport appendix section of the vsphere reference architecture for OCP 3.10 and 3.11.

As per BZ# https://bugzilla.redhat.com/show_bug.cgi?id=1730003

PR #18569, opened 2019-12-11
BZ-1730003 Remove vSphere VM deployment variables

codyhoag:
https://bugzilla.redhat.com/show_bug.cgi?id=1730003

Supersedes #16338

### DIFF
--- a/install_config/topics/configuring_for_vsphere.adoc
+++ b/install_config/topics/configuring_for_vsphere.adoc
@@ -73,13 +73,6 @@ openshift_deployment_type=openshift-enterprise
 oreg_url=registry.access.redhat.com/openshift3/ose-${component}:${version}
 openshift_examples_modify_imagestreams=true
 
-VM deployment
-openshift_cloudprovider_vsphere_template="rhel75-template"
-openshift_cloudprovider_vsphere_vm_network="VM Network"
-openshift_cloudprovider_vsphere_vm_netmask="255.255.254.0"
-openshift_cloudprovider_vsphere_vm_gateway="10.x.y.254"
-openshift_cloudprovider_vsphere_vm_dns="10.x.y.5"
-
 # vSphere Cloud provider
 openshift_cloudprovider_kind=vsphere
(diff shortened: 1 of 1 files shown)


codyhoag:
@openshift/team-documentation can you review? Thanks!

kalexand-rh:
@katherinedube, @sferich888, are you both ok with this change?

install_config/topics/configuring_for_vsphere.adoc
```diff
@@ -186,3 +179,7 @@ app-2.example.com openshift_node_group_name="node-config-compute"
 default container image registry, specify the credentials for that account. See
 xref:../install_config/configuring_red_hat_registry.html#install-config-configuring-red-hat-registry[Accessing and Configuring the Red Hat Registry].
 
+[NOTE]
+====
+link:https://access.redhat.com/documentation/en-us/reference_architectures/2019/html-single/deploying_and_managing_openshift_3.11_on_a_vmware_software-defined_data_center/index#deploying_a_vsphere_vm_environment_optional[Deploying a vSphere VM environment] is not officially supported by Red Hat, but it can be configured.
```
kalexand-rh: We don't usually doc things that aren't supported, so removing the option is the right thing to do, but I think that Katherine or Eric should weigh in on the link.

katherinedube: After reading through the Bugzilla, I'm good with removing this as well. Otherwise, Maria Bracho is the PM for the VMware. She should at least be made aware of the change.

kalexand-rh: I missed that Bugzilla had logged me out, but with the rest of the conversation, it does make sense.

codyhoag: @katherinedube just so we're on the same page, do suggest we also remove the link? Or is it OK to link out to the reference architecture article?

katherinedube: I think the link is fine. As long as we remove the unsupported vm deployment section.

PR #18595, opened 2019-12-12
Fix how to create new alertmanager.yaml secret

duritong:
The way how it was described didn't work. A) There is no file config.yaml in the game here
and B) if you do create secret --from-file it will automatically add the base64 encoded content
of the file to data with the filename it read from.

This is broken in 4.2 configs as well, should be backported there.

### DIFF
--- a/modules/monitoring-applying-custom-alertmanager-configuration.adoc
+++ b/modules/monitoring-applying-custom-alertmanager-configuration.adoc
@@ -23,31 +23,29 @@ $ oc -n openshift-monitoring get secret alertmanager-main --template='{{ index .
 +
 [source,yaml]
 ----
-data:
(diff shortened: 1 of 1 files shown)


brancz:
lgtm

vikram-redhat:
@rh-max - can you please take a look?

rh-max:
@duritong Thank you for the fix!
@vikram-redhat I'll rely on @brancz 's ACK of the changes, so lgtm. Could you please merge this to master and cherry-pick to enterprise-4.2 and enterprise-4.3? Thanks.

vikram-redhat:
@rh-max  Waiting on QE ack.

juzhao:
LGTM

PR #18611, opened 2019-12-13
Fix pruning groups example code block

aivaraslaimikis:
Code block was missing additional dash in beginning and in the end of block.


vikram-redhat:
Thanks @aivaraslaimikis!

PR #18612, opened 2019-12-13
Fix prune groups sync config option name

aivaraslaimikis:
Two examples of group pruning is using not existing pruner option:
--sync-file
From oc cli and other examples it should be:
--sync-config

Verified on oc v3.11.154


aivaraslaimikis:
@openshift/team-documentation

vikram-redhat:
@xingxingxia can you verify this change?

xingxingxia:
LGTM. CC'ing @barleyer if having any more comment.

kalexand-rh:
Merging.

Issue #18444, opened 2019-12-04
[enterprise-4.2] Edit suggested in file authentication/using-rbac.adoc


ChetHosey:
### Which section(s) is the issue in?

https://docs.openshift.com/container-platform/4.2/authentication/using-rbac.html#evaluating-authorization_using-rbac

### What needs fixing?

Under "Bindings", I'm not sure what this statement is trying to say:

> Project administrators can visualize roles, including a matrix of the verbs and resources each are associated using the CLI to view local roles and bindings.

It took me a while to realize that this probably means "...each are associated *with* using the CLI...". Even with that correction it's a bit awkward.

Something like the following might be more clear:

> Project administrators can use the CLI to view local role and bindings, including a matrix of the verbs and resources each are associated with.

That still doesn't explain *how* to use the CLI for that purpose. It's further described in https://docs.openshift.com/container-platform/4.2/authentication/using-rbac.html#viewing-local-roles_using-rbac. So it might be helpful to reference that section using a link, if at all possible, so the reader knows where to go for details.

ahardin-rh:
@ChetHosey Thank you for your feedback! This is now addressed in https://github.com/openshift/openshift-docs/pull/18763

PR #18763, opened 2019-12-19
Issue 18444, clarifed wording about viewing roles in the CLI

ahardin-rh:
https://github.com/openshift/openshift-docs/issues/18444


ahardin-rh:
Minor update. No reviews needed.

PR #18787, opened 2019-12-20
OSDOCS-796: Add description of metrics exported by Machine Config Daemon

rh-max:
OSDOCS: https://issues.redhat.com/browse/OSDOCS-796

### DIFF
--- a/modules/machine-config-daemon-metrics.adoc
+++ b/modules/machine-config-daemon-metrics.adoc
@@ -49,11 +49,15 @@ While some entries contain commands for getting specific logs, the most comprehe
 |Logs errors encountered during pivot. *
 |Pivot errors might prevent OS upgrades from proceeding. For further investigation, run these commands to access the node and see its logs:
 
-`$ oc debug node <node>`
+`$ oc debug node/<node_name>`
 
-`$ oc logs -f -n openshift-machine-config-operator machine-config-daemon-<hash> -c machine-config-daemon`
+`$ chroot /host`
 
 `$ journalctl -u pivot.service`
+
+You can also run:
+
+`$ oc logs -f -n openshift-machine-config-operator machine-config-daemon-<hash> -c machine-config-daemon`
 
 |mcd_state
 |[]string{"state", "reason"}
@@ -67,7 +71,9 @@ While some entries contain commands for getting specific logs, the most comprehe
 |Logs kubelet health failures.  *
 |This is expected to be empty, with failure count of 0. If failure count exceeds 2, the error indicating threshold is exceeded. This indicates a possible issue with the health of the kubelet. For further investigation, run these commands to access the node and see its logs:
 
-`$ oc debug node <node>`
+`$ oc debug node/<node_name>`
+
+`$ chroot /host`
 
 `$ journalctl -u kubelet`
 
--- a/modules/machine-config-daemon-metrics.adoc
+++ b/modules/machine-config-daemon-metrics.adoc
@@ -40,24 +40,34 @@ While some entries contain commands for getting specific logs, the most comprehe
 |mcd_drain*
 |{"drain_time", "err"}
 |Logs errors received during failed drain. *
-|While drains might need multiple tries to succeed, terminal failed drains prevent updates from proceeding. The `drain_time` metric, which shows how much time the drain took, might help with troubleshooting. For further investigation, see the logs by running:
+|While drains might need multiple tries to succeed, terminal failed drains prevent updates from proceeding. The `drain_time` metric, which shows how much time the drain took, might help with troubleshooting.
+
+For further investigation, see the logs by running:
 
 `$ oc logs -f -n openshift-machine-config-operator machine-config-daemon-<hash> -c machine-config-daemon`
 
 |mcd_pivot_err*
 |[]string{"pivot_target", "err"}
 |Logs errors encountered during pivot. *
-|Pivot errors might prevent OS upgrades from proceeding. For further investigation, run these commands to access the node and see its logs:
+|Pivot errors might prevent OS upgrades from proceeding.
+
+For further investigation, run the following commands to access the node and see its logs.
+
+To access the node:
 
 `$ oc debug node/<node_name>`
+
+Once on the node, show the logs:
 
 `$ chroot /host`
 
 `$ journalctl -u pivot.service`
 
-You can also run:
+Alternatively, you can run:
 
 `$ oc logs -f -n openshift-machine-config-operator machine-config-daemon-<hash> -c machine-config-daemon`
+
+Unlike `journalctl`, this command only shows the logs from the `machine-config-daemon` container and not all the logs from the node.
 
 |mcd_state
 |[]string{"state", "reason"}
@@ -69,9 +79,15 @@ You can also run:
 |mcd_kubelet_state*
 |[]string{"err"}
 |Logs kubelet health failures.  *
-|This is expected to be empty, with failure count of 0. If failure count exceeds 2, the error indicating threshold is exceeded. This indicates a possible issue with the health of the kubelet. For further investigation, run these commands to access the node and see its logs:
+|This is expected to be empty, with failure count of 0. If failure count exceeds 2, the error indicating threshold is exceeded. This indicates a possible issue with the health of the kubelet.
+
+For further investigation, run the following commands to access the node and see its logs.
+
+To access the node:
 
 `$ oc debug node/<node_name>`
+
+Once on the node, show the logs:
 
 `$ chroot /host`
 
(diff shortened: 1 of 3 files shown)


juzhao:
@mnguyen @miabbott
Is machine-config daemon owned by you? If so, please help to review the doc

kikisdeliveryservice:
@juzhao @mike-nguyen  @miabbott 

I owned this epic for MCO and worked with @rh-max on these docs. If you have any questions, LMK.

juzhao:
> @juzhao @mike-nguyen @miabbott
> 
> I owned this epic for MCO and worked with @rh-max on these docs. If you have any questions, LMK.

Thanks, please help to review the docs

vikram-redhat:
> @juzhao @mike-nguyen @miabbott
> 
> I owned this epic for MCO and worked with @rh-max on these docs. If you have any questions, LMK.

Are there any updates on this? Is this doc ok from QE point of view?

juzhao:
@kikisdeliveryservice is the right person to review the doc, please reply to @vikram-redhat

modules/machine-config-daemon-metrics.adoc
```diff
@@ -0,0 +1,77 @@
...
+`$ oc logs -f -n openshift-machine-config-operator machine-config-daemon-<hash> -c machine-config-daemon`
+
+|mcd_pivot_err*
+|[]string{"pivot_target", "err"}
+|Logs errors encountered during pivot. *
+|Pivot errors might prevent OS upgrades from proceeding. For further investigation, see the logs by running these commands:
+
+`$ oc logs -f -n openshift-machine-config-operator machine-config-daemon-<hash> -c machine-config-daemon`
+
+`$ journalctl -u pivot.service`
```
miabbott: It should be noted that this command should be run on the nodes themselves after using something like `oc debug node <node>`

modules/machine-config-daemon-metrics.adoc
```diff
@@ -0,0 +1,77 @@
...
+|For further investigation, see the logs by running:
+
+`$ oc logs -f -n openshift-machine-config-operator machine-config-daemon-<hash> -c machine-config-daemon`
+
+|mcd_kubelet_state*
+|[]string{"err"}
+|Logs kubelet health failures.  *
+|This is expected to be empty, with failure count of 0. If failure count exceeds 2, the error indicating threshold is exceeded. This indicates a possible issue with the health of the kubelet. For further investigation, see the logs by running:
+
+`$ journalctl -u kubelet`
```
miabbott: Same comment as above; use `oc debug node <node>` to access the node first

mike-nguyen:
Agreed with Micah's comments.  Otherwise LGTM.

rh-max:
Thanks @miabbott , implemented the feedback.
To clarify, @kikisdeliveryservice has reviewed this document before the PR was even created. Here, I was just asking for a QE review.
Thanks @kikisdeliveryservice , @miabbott , @juzhao , @mike-nguyen for your work on this! On its way to peer review & merging now.

modules/machine-config-daemon-metrics.adoc
```diff
@@ -0,0 +1,81 @@
+// Module included in the following assemblies:
+//
+// * nodes/nodes/nodes-nodes-machine-config-daemon-metrics.adoc
+
+[id="machine-config-daemon-metrics_{context}"]
+= Machine Config Daemon metrics
+
+Beginning {product-title} 4.3, the Machine Config Daemon provides a set of metrics. These metrics can be accessed using the Prometheus Cluster Monitoring stack.
+
+The following table describes this set of metrics. Note that:
```
sheriff-rh: I recommend a `[NOTE]` here, instead of a bulleted list.

sheriff-rh (approved):
Looks great! Just one suggestion.

modules/machine-config-daemon-metrics.adoc
```diff
@@ -0,0 +1,81 @@
...
+|While drains might need multiple tries to succeed, terminal failed drains prevent updates from proceeding. The `drain_time` metric, which shows how much time the drain took, might help with troubleshooting. For further investigation, see the logs by running:
+
+`$ oc logs -f -n openshift-machine-config-operator machine-config-daemon-<hash> -c machine-config-daemon`
+
+|mcd_pivot_err*
+|[]string{"pivot_target", "err"}
+|Logs errors encountered during pivot. *
+|Pivot errors might prevent OS upgrades from proceeding. For further investigation, run these commands to access the node and see its logs:
+
+`$ oc debug node <node>`
```
miabbott: This is out of order.  If we do `oc debug node` and then try to do `oc logs` I don't think that will work.

The explicit commands would look like:

```
$ debug node/ip-10-0-139-75.us-west-2.compute.internal
$ chroot /host
$ journalctl -u pivot.service
```

This connects to the node, `chroot` to the node's filesystem, then run's the `journalctl` command on the node.

modules/machine-config-daemon-metrics.adoc
```diff
@@ -0,0 +1,81 @@
...
+|For further investigation, see the logs by running:
+
+`$ oc logs -f -n openshift-machine-config-operator machine-config-daemon-<hash> -c machine-config-daemon`
+
+|mcd_kubelet_state*
+|[]string{"err"}
+|Logs kubelet health failures.  *
+|This is expected to be empty, with failure count of 0. If failure count exceeds 2, the error indicating threshold is exceeded. This indicates a possible issue with the health of the kubelet. For further investigation, run these commands to access the node and see its logs:
+
+`$ oc debug node <node>`
```
miabbott: This is in the correct order, but we could make it more explicit with the addition of the `chroot /host` command as noted above.

modules/machine-config-daemon-metrics.adoc
rh-max: Thanks @sheriff-rh . I'll rather keep it like this, since those details closely pertain to the table, and are not a sort of side note.

rh-max: Changed my mind and going with your suggestion. Substituted two bullet points with two [NOTE]s.

modules/machine-config-daemon-metrics.adoc
rh-max: Changed accordingly. So how to describe the difference between logs from `oc logs` and from `journalctl`? I want to be more explicit than just "You can also run:".

sheriff-rh (approved):
LGTM, thanks Max! Shoot me a message if you need a merge.

modules/machine-config-daemon-metrics.adoc
miabbott: The `oc logs` command is only going to fetch the logs from the particular container in the pod (`machine-config-daemon`).  It's something that you want to run from your workstation, not from a node on the cluster.

The `oc debug node` route gets you on the host itself and using `journalctl` is going to give you the full contents of the journal on the node.  You could find the `machine-config-daemon` logs in that journal, but they will be mixed in with the rest of the node logs and therefore harder to sift through.

kikisdeliveryservice: @rh-max @miabbott one way to make this clear would be to use a different shell prompt for the on-node commands so oc debug node stays the same but instead of using `$` for the chroot and journalctl commands you can use something else like `%` or `#` to show that it's a different level than `$`

kikisdeliveryservice: so it would look like:
```
$ oc debug node/ip-10-0-139-75.us-west-2.compute.internal
% chroot /host
% journalctl -u pivot.service
```
You can also run
`$ oc logs -f ....`

miabbott: @kikisdeliveryservice has the right idea....using different prompts to indicate on your workstation vs on the node itself.  thanks Kirsten!

kikisdeliveryservice: @rh-max wdyt? ^^^

rh-max: Thanks @miabbott and @kikisdeliveryservice ! I've applied your feedback and improved the structure a bit. Does the result look good to you?
(We can't just opt into using the `%` for shell prompt in the docs, so I've separated the "before login" and "after login" steps with words.)

mike-nguyen: The user can also run `oc debug node/<node> -- chroot /host journalctl -u pivot.service`.  With this one liner we won't need to distinguish between the different prompts.

vikram-redhat:
@rh-max is this ready for merge?

modules/machine-config-daemon-metrics.adoc
rh-max: Thanks @mike-nguyen , I changed the commands to your simplified command. Since the feedback is implemented and I hear no NOACKs, I'm asking @sheriff-rh to merge this.

rh-max:
Hey Andrew @sheriff-rh , could you please merge this one and cherry-pick it into `4.3`? Thanks.

vikram-redhat:
@rh-max multiple commits.

rh-max:
@vikram-redhat Sorry, forgot to squash.

PR #18800, opened 2019-12-21
fixing travis

vikram-redhat:


vikram-redhat:
@openshift/team-documentation This fixes the issue with Travis not reporting errors sometimes. Now, not only will all errors be reported, but more critical errors (like a missing file) will stop the build and the error will be reported straight away without further processing. Less critical errors will be reported together.

For more background: https://issues.redhat.com/browse/OSDOCS-867

PR #18806, opened 2019-12-24
cnv-docs: request users to set monitoring ns label

danielBelenky:
In CNV, we deploy a set of PrometheusRules wich supposed to be picked by
openshift's monitoring-operator. For that process to work properly, the
user has to label the namespace in which he deploys CNV.

Signed-off-by: Daniel Belenky <dbelenky@redhat.com>


danielBelenky:
setting hold because there might be an automated solution

fabiand:
Yes, let's investigate the automated solution.

vikram-redhat:
@danielBelenky is this still needed? If not can we close it?

@aburdenthehand

PR #18812, opened 2019-12-29
Pythonto3

vikram-redhat:


vikram-redhat:
@openshift/team-documentation - FYI. Updated scripts to use Python 3. No further action required. Still have to do some changes to get rid of build warnings but this should see us through.

It is done in conjunction with https://github.com/openshift/openshift-docs/pull/18800 and some other changes internally. 

Other changes to make it compatible with portal are also in place. Let me know if you have any questions.

Issue #18268, opened 2019-11-25
[online] Edit suggested in file getting_started/access_your_services.adoc


erjosito:
### Which section(s) is the issue in?

Create an Internal-Only TCP Service

### What needs fixing?
```
$ oc expose dc httpd-example --type=LoadBalancer --name=internal-lb --dry-run -o yaml | awk '1;/metadata:/{ print "  annotations:\n    service.beta.kubernetes.io/azure-load-balancer-internal: 0.0.0.0/0" }' | oc create -f -service/internal-lb exposed
```

should probably be (according to the annotation for internal load balancers described in [internal-lb](https://docs.microsoft.com/bs-latn-ba/azure/aks/internal-lb) )

```
$ oc expose dc httpd-example --type=LoadBalancer --name=internal-lb --dry-run -o yaml | awk '1;/metadata:/{ print "  annotations:\n    service.beta.kubernetes.io/azure-load-balancer-internal: \"true\"" }' | oc create -f -service/internal-lb exposed
```

sheriff-rh:
Hi Jose, I have made the corrections as you have suggested. If the changes are satisfactory to you, I will merge them later today: #18665

erjosito:
@sheriff-rh Looks good! Thanks!

sheriff-rh:
Changes have been merged, I will mark this issue as closed. 

Jose, thank you for helping us keep our docs in shape! Have a great holiday and 2020!

PR #18825, opened 2020-01-03
Fix OSD LoadBalancer internal-only command

wgordon17:
@sheriff-rh Can I get this merged please 🙏

### DIFF
--- a/modules/dedicated-exposing-TCP-services.adoc
+++ b/modules/dedicated-exposing-TCP-services.adoc
@@ -50,10 +50,8 @@ You can alternatively expose your applications internally only, enabling access
 only through AWS VPC Peering or a VPN connection.
 
 ----
-$ oc expose dc httpd-example --type=LoadBalancer --name=internal-lb --dry-run -o
-yaml | awk '1;/metadata:/{ print "  annotations:\n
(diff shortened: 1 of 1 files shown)


maxwelldb:
@sheriff-rh I'm merging this after discussion with Will. Could you verify that it's all good when you have a moment?

wgordon17:
@sheriff-rh It looks like change was introduced by this issue: #18268 

The issue incorrectly used the `azure` annotation. However the implication that we should be using `"true"` instead of `0.0.0.0/0` seems valid based on [K8s docs](https://kubernetes.io/docs/concepts/services-networking/service/#internal-load-balancer) (AWS tab). I have tested this with `0.0.0.0/0` and it seems that it still works with this value, this is likely something that is or will be deprecated by K8s at some point. Please update this to use `"true"` at some point to ensure we're inline with upstream K8s.

So it would look like 

```bash
$ oc expose dc httpd-example --type=LoadBalancer --name=internal-lb --dry-run -o yaml | awk '1;/metadata:/{ print "  annotations:\n    service.beta.kubernetes.io/aws-load-balancer-internal: \"true\"" }' | oc create -f -
service/internal-lb created
```

PR #18851, opened 2020-01-06
Bug 1745889, Online users cannot have docker.io as a private registry

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1745889


sheriff-rh (approved):
Thanks!

PR #18858, opened 2020-01-06
Clarify requirement for using openshift-operators-redhat namespace in logging setup

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1788255

### DIFF
--- a/modules/cluster-logging-deploy-subscription.adoc
+++ b/modules/cluster-logging-deploy-subscription.adoc
@@ -56,10 +56,17 @@ metadata:
   annotations:
     openshift.io/node-selector: ""
   labels:
-    openshift.io/cluster-logging: "true"
-    openshift.io/cluster-monitoring: "true"
+    openshift.io/cluster-monitoring: "true" <2>
 ----
-<1> You must specify the `openshift-operators-redhat` namespace.
+<1> You must specify the `openshift-operators-redhat` namespace. To prevent 
+possible conflicts with metrics, you should configure the Prometheus Cluster 
+Monitoring stack to scrape metrics from the `openshift-operators-redhat` 
+namespace and not the `openshift-operators` namespace. The `openshift-operators` 
+namespace can contain community operators, which are untrusted and could publish 
+a metric with the same name as an {product-title} metric, which would cause 
+conflicts. 
+<2> You must specify this label as shown to ensure that cluster monitoring 
+scrapes the `openshift-operators-redhat` namespace.
 
 .. Run the following command to create the namespace:
 +


mburke5678:
@bparees @bostrt PTAL

modules/cluster-logging-deploy-subscription.adoc
```diff
@@ -59,7 +59,13 @@ metadata:
...
     openshift.io/cluster-monitoring: "true"
 ----
-<1> You must specify the `openshift-operators-redhat` namespace.
+<1> You must specify the `openshift-operators-redhat` namespace. To prevent 
+possible conflicts with metrics, you should configure the Prometheus Cluster 
+Monitoring stack to scrape metrics from the `openshift-operators-redhat` 
+namespace and not the `openshift-operators` namespace. The `openshift-operators` 
+namespace can contain community operators, which are untrusted and could publish 
+a metric with the same name as an {product-title} metric, which would cause 
+conflicts.
```
bparees: guessing people will ask "how do you configure cluster monitoring to scrape X and not Y" to which the answer is "via labels on the namespace" (the labels that are seen here).

maybe worth a `<2>` on the cluster-monitoring labels to explain that that is the purpose of the label.

mburke5678: @bparees I added an item for the `openshift.io/cluster-monitoring: "true"` label. 
What does the `openshift.io/cluster-logging: "true"` label do?

See d7beed4a363ab9cb5054782554ca009422167ad6

bparees: I actually do not know..  @jcantrill ?

only the cluster-monitoring label should be needed for metric scraping.

jcantrill: > I actually do not know.. @jcantrill ?
> 
`git blame` shows it in a dev deployment script as part of https://github.com/openshift/elasticsearch-operator/pull/76  I know of no use for the label.  I would suggest dropping it from the docs. 

> only the cluster-monitoring label should be needed for metric scraping.

mburke5678:
Removed the `openshift.io/cluster-logging: "true"` labels from the EO and CLO project examples per https://github.com/openshift/openshift-docs/pull/18858#discussion_r364752329

modules/cluster-logging-deploy-subscription.adoc
```diff
@@ -56,10 +56,17 @@ metadata:
...
+    openshift.io/cluster-monitoring: "true" <2>
+----
+<1> You must specify the `openshift-operators-redhat` namespace. To prevent 
+possible conflicts with metrics, you should configure the Prometheus Cluster 
+Monitoring stack to scrape metrics from the `openshift-operators-redhat` 
+namespace and not the `openshift-operators` namespace. The `openshift-operators` 
+namespace can contain community operators, which are untrusted and could publish 
+a metric with the same name as an {product-title} metric, which would cause 
+conflicts. 
+<2> You must specify this labels as shown to ensure that cluster monitoring
```
bparees:
```suggestion
<2> You must specify this label as shown to ensure that cluster monitoring 
```

modules/cluster-logging-deploy-subscription.adoc
```diff
@@ -56,10 +56,17 @@ metadata:
...
+----
+<1> You must specify the `openshift-operators-redhat` namespace. To prevent 
+possible conflicts with metrics, you should configure the Prometheus Cluster 
+Monitoring stack to scrape metrics from the `openshift-operators-redhat` 
+namespace and not the `openshift-operators` namespace. The `openshift-operators` 
+namespace can contain community operators, which are untrusted and could publish 
+a metric with the same name as an {product-title} metric, which would cause 
+conflicts. 
+<2> You must specify this labels as shown to ensure that cluster monitoring 
+scrapes only from the `openshift-operators-redhat` namespace.
```
bparees:
```suggestion
scrapes the `openshift-operators-redhat` namespace.
```

bparees:
nits and lgtm

mburke5678:
@anpingli PTAL. I am most concerned with removing the `openshift.io/cluster-logging: "true"` parameter from the EO and CLO project examples. I want to ask if you have concerns with removing this parameter.

anpingli:
@mburke5678  Let me confirmed with a round function test. Will update the result later.

mburke5678:
@anpingli I was able to complete the install without those parameters. That doesn't mean that logging worked properly when I did. 
Else, if you know what that parameter is for, we can leave it in. The decision to remove the parameters from the docs arose when no one knew what the parameter is for.

anpingli:
@bparees  Which type of metrics do we want monitoring to scrape? why not use the ServiceMonitor CRD?

bparees:
> @bparees Which type of metrics do we want monitoring to scrape? why not use the ServiceMonitor CRD?

i assume there is a servicemonitor CRD being created.  But for cluster-monitoring to scrape a namespace, the namespace must be labeled appropriately, which is primarily what this docs PR is about.

i'm not sure what metrics logging reports, offhand.

anpingli (commented):
LGTM

PR #18863, opened 2020-01-07
Update securing_builds doc with warning on networking rule bypass

sfowl:
Proposed warning only applies to OpenShift 3.x, not OpenShift 4.

### DIFF
--- a/admin_guide/securing_builds.adoc
+++ b/admin_guide/securing_builds.adoc
@@ -27,6 +27,13 @@ xref:../architecture/core_concepts/builds_and_image_streams.adoc#docker-build[Do
 build] permission should also be granted with caution as a vulnerability in the Docker build
 logic could result in a privileges being granted on the host node.
 
+[WARNING]
+====
+During Docker and custom builds, actions performed by the Docker daemon are privileged and
+occur in the host network namespace. Such actions bypass configured networking rules
+including those defined by EgressNetworkPolicy objects and static egress IP addresses.
+====
+
 By default, all users that can create builds are granted permission to use the
 Docker and Source-to-Image build strategies. Users with xref:../architecture/additional_concepts/authorization.adoc#roles[*cluster-admin*]
 privileges can enable the Custom build strategy, as referenced in the xref:restricting-build-strategies-to-a-user-globally[Restricting Build Strategies to a User Globally]


admin_guide/securing_builds.adoc
```diff
@@ -27,6 +27,13 @@ xref:../architecture/core_concepts/builds_and_image_streams.adoc#docker-build[Do
 build] permission should also be granted with caution as a vulnerability in the Docker build
 logic could result in a privileges being granted on the host node.
 
+[WARNING]
+====
+During builds, actions performed by the Docker daemon are privileged and occur in the host
+network namespace. Such actions bypass configured networking rules including
```
juanluisvaladas: The warning is completely fine but perhaps we should be more explicit saying this only affects docker builds and potentially custom builds to avoid customers being concerned about non real problems.
Could  we replace "During builds" with "During docker and custom builds"?

sfowl: Thanks, fixed.

juanluisvaladas (approved):
lgtm

vikram-redhat:
Needs to be CP'ed to master-3. Not to be merged till QE has acked it.

kalexand-rh (commented):
LGTM, pending QE approval.

admin_guide/securing_builds.adoc
```diff
@@ -27,6 +27,13 @@ xref:../architecture/core_concepts/builds_and_image_streams.adoc#docker-build[Do
 build] permission should also be granted with caution as a vulnerability in the Docker build
 logic could result in a privileges being granted on the host node.
 
+[WARNING]
+====
+During Docker and custom builds, actions performed by the Docker daemon are privileged and
+occur in the host network namespace. Such actions bypass configured networking rules
+including EgressNetworkPolicies and static egress IP addresses.
```
jboxman: Would this make more sense as:

> networking rules including egress network policy...

Or

> networking rules including those defined by EgressNetworkPolicy objects...

jboxman:
@sfowl, can you squash the commit and force-push? Thanks!

admin_guide/securing_builds.adoc
sfowl: Thanks, went with the latter.

sfowl:
> @sfowl, can you squash the commit and force-push? Thanks!

Yep, done.

PR #18871, opened 2020-01-07
Fix netpol documentation for IngressControllers with hostSubnet

juanluisvaladas:


juanluisvaladas:
I have **not** compiled this, so please do check that I'm not breaking the format.
Fixes BZ https://bugzilla.redhat.com/show_bug.cgi?id=1768608 and needs a backport to all the way down to 4.2.

Perhaps 4.1 could use it too but I haven't verified if 4.1 actually has netid: 0 for the default project.

jboxman:
@juanluisvaladas, thanks for opening this! I had found the various discussions around it somewhat vexing, and would never have derived this clean solution from those discussions as my working knowledge of OCP network policy isn't deep enough.

jboxman:
I'm going to wrap this one up with https://github.com/openshift/openshift-docs/pull/19360.

PR #18918, opened 2020-01-09
ARO: Add link to Azure Monitor documentation

olga-mir:
Microsoft already has documentation on Azure Monitor for containers including on ARO clusters. We do not wan't to duplicate it, but we need to make our customers aware that this feature is available to them and provide links to that documentation.


kalexand-rh:
@vikram-redhat, who should provide a review?

Makdaam:
/lgtm
The doc change better reflects what's going on. The first version was written when the Azure Monitor for ARO was still in development.

kalexand-rh:
Ok! Then I'll merge it. Thanks. :)

PR #18924, opened 2020-01-09
Remove cluster logging step to get channel version

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1784929

### DIFF
--- a/modules/cluster-logging-deploy-subscription.adoc
+++ b/modules/cluster-logging-deploy-subscription.adoc
@@ -127,14 +127,6 @@ For example:
 $ oc create -f eo-og.yaml
 ----
 
-.. Use the following command to get the `channel` value required for the next step.
-+
-----
-$ oc get packagemanifest elasticsearch-operator -n openshift-marketplace -o jsonpath='{.status.channels[].name}'
-
-4.3
-----
-
 .. Create a Subscription object YAML file (for example, `eo-sub.yaml`) to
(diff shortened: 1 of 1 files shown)


mburke5678:
@jcantrill @bparees PTAL. 

I will update 4.3 with the 4.3 channel after merge.

modules/cluster-logging-deploy-subscription.adoc
```diff
@@ -154,7 +146,7 @@ spec:
   name: "elasticsearch-operator"
 ----
 <1> You must specify the `openshift-operators-redhat` namespace.
-<2> Specify the `.status.channels[].name` value from the previous step.
+<2> Specify channel version `4.2`.
```
bparees: so we're going to have to keep updating this doc w/ every release?  can we somehow make this more generic so we don't have that problem?

mburke5678: Is there a plan to get the Elasticserch Operator to install using the web console? If users could install via the web console all of these steps become unnecessary.

bparees: i thought there was some plan to have cluster logging installation automatically pull in elastic via a dependency process.

jcantrill: afaik its possible to do everything from the console though not necessarily through the OLM bits.

PR #18936, opened 2020-01-10
CNV - Unsetting variable cleanup (4.2)

aburdenthehand:
Wrapping the unsetting of variables in the conditionals that they were set in.

This is for 4.2 repo only because these assemblies include a module that has an identical file name but different content in the master branch.


sheriff-rh (approved):
LGTM

PR #18938, opened 2020-01-10
Remove ipi/upi dichotomy for machine management capabilities

enxebre:
Usage of the machine API does NOT mandate that the customer used IPI. The only requirement is that the customer delegates the ability to create lifecycle machines to the product.


enxebre:
cc @kalexand-rh

kalexand-rh:
@jhou1, will you PTAL?

PR #18328, opened 2019-11-27
Bug 1774377: [DOCS] No mention for how to upgrade RHEL compute node from v4.1 to v4.2

bysnupy:
- Reference: [[DOCS] No mention for how to upgrade RHEL compute node from v4.1 to v4.2](https://bugzilla.redhat.com/show_bug.cgi?id=1774377)
- Version: 4.2
- Description: If a RHEL compute node upgrade from 4.1 to 4.2, you should switch the required repositories as new version.

PR #18990, opened 2020-01-13
updating RHEL packages for 4.3

kalexand-rh:
4.3 follow-up to https://github.com/openshift/openshift-docs/pull/18328

@gpei, is this enough for the RHEL packages for 4.3, or does the Ansible version need to change, too?


gpei:
It should be enough, no ansible version change needed here, we're still requiring `ansible >= 2.8.2` in 4.3. Thanks for the update!

kalexand-rh:
Thank you, @gpei!

PR #18998, opened 2020-01-13
Remove Greylock and Splunk from log forwarding

mburke5678:
Removing Splunk and Graylog from into to log forwarding topic. These are future endpoints based on What New in OpenShift 4.3 presentation 1/13.


mburke5678:
@sichvoge @jcantrill Should we remove Splunk and Greylock from the log forwarding docs? Christian suggested these will be supported in the future?

jcantrill (approved):

PR #18974, opened 2020-01-11
cnv subscribing: only openshift-cnv must be used

dankenigsberg:
The warning text seemed to suggest that any namespace may be chosen for
cnv. This is not the case; openshift-cnv must be typed in.

Signed-off-by: Dan Kenigsberg <danken@redhat.com>

PR #19004, opened 2020-01-13
PR based on "cnv subscribing: only openshift-cnv must be used"

ousleyp:
* See discussion in https://github.com/openshift/openshift-docs/pull/18974
* Cherrypick to enterprise-4.2 and enterprise-4.3


ahardin-rh:
Looks good!

PR #19021, opened 2020-01-14
Failed migration plan does not save user's changes to default settings

apinnick:
For 4.3.


jortel:
```
If a migration fails, the migration plan reverts to its default settings and does not save any changes you may have made. You must manually roll back the migration, delete the migration plan, and create a new migration plan. (BZ#1784899)
```

To be more specific: instead of  `does not save any changes you may have made` I recommend we something like `may not retain user selected settings for PVs associated with quiesced pods`.

apinnick:
> ```
> If a migration fails, the migration plan reverts to its default settings and does not save any changes you may have made. You must manually roll back the migration, delete the migration plan, and create a new migration plan. (BZ#1784899)
> ```
> 
> To be more specific: instead of `does not save any changes you may have made` I recommend we something like `may not retain user selected settings for PVs associated with quiesced pods`.

Are all pods quiesced automatically or are you talking about an option that the user would choose? In IMS (CloudForms), quiescing applications is the default and the user can choose not to quiesce. Does the user have this choice when creating an OpenShift migration plan?

If it's an option, then I can rewrite the sentence. If it's not an option, the detail that the pods are quiesced is unnecessary, but I can add that changed PV settings are lost. 

I phrased it as "any changes" because I wasn't sure whether other settings were affected. Are other settings affected by failed migration?

> Also, after the scaled down pods are scaled back up, the plan can be edited to re-do selections for the PVs. The plan does not need to be deleted/created.

@jwmatthews told me that the plan has to be deleted and recreated. John?

apinnick:
Re: quiescing pods -- I see that is an option when creating migration plan. I'll rewrite that.

I still need John's input about deleting/recreating plan.

jwmatthews:
@apinnick  (cc @jortel @eriknelson )

During a migration user may choose to do a 'quiesce' or not, it is optional.
In the UI, this is a checkbox someone would check on/off when doing the migration, from API it'd be a boolean option.

The issue Jeff described is that if user chose a quiesce, and after doing quiesce the migration failed....then we are in a spot that the MigPlan no longer contains the info the user previous had for the PV choices.   

Technically the MigPlan could be edited to fix after following Jeff's steps.  I wasn't sure if the WebUI could handle this (from API makes sense edit of MigPlan is sufficient as Jeff wrote).   

Unless @jortel writes back and confirms that UI is good on just editing, let's go with the instructions of delete the MigPlan and recreate for now.   I think this is safer instructions to give a user since we have some concerns about edit of plan in UI not handling this correctly.

apinnick:
@jwmatthews Please review updated version.

jwmatthews (approved):
Technical approval

sheriff-rh (approved):

jortel:
On Thu, 2020-01-16 at 01:01 +0000, John Matthews wrote:
> @apinnick  (cc @jortel @eriknelson )
> 
> During a migration user may choose to do a 'quiesce' or not, it is
> optional.
> 
> In the UI, this is a checkbox someone would check on/off when doing
> the migration, from API it'd be a boolean option.
> 
> The issue Jeff described is that if user chose a quiesce, and after
> doing quiesce the migration failed....then we are in a spot that the
> MigPlan no longer contains the info the user previous had for the PV
> choices.
> 
> Technically the MigPlan could be edited to fix after following Jeff's
> steps.  I wasn't sure if the WebUI could handle this (from API makes
> sense edit of MigPlan is sufficient as Jeff wrote).
> 
> Unless @jortel writes back and confirms that UI is good on just
> editing, let's go with the instructions of delete the MigPlan and
> recreate for now.   I think this is safer instructions to give a user
> since we have some concerns about edit of plan in UI not handling
> this correctly.

Yes, the UI supports editing a plan which runs through the wizard
(again) including the PV settings.

PR #19034, opened 2020-01-14
splitting scalability module

kalexand-rh:
@ahardin-rh, I think this module needs to be broken up. Do you have any objections?


ahardin-rh:
@kalexand-rh No objections here! Looks great to me.

kalexand-rh:
Thank you!

PR #19053, opened 2020-01-15
update migration from 4.2 for 4.3 GA

apinnick:


apinnick:
@bergerhoffer Please review and merge. This is for 4.3. Thanks!

migration/migrating-4_2-4/migrating-application-workloads-4_2-to-4.adoc
```diff
@@ -1,12 +1,12 @@
...
-= Migrating application workloads between {product-title} 4.2 clusters
+= Migrating application workloads from {product-title} 4.2 to {product-version}
 include::modules/common-attributes.adoc[]
 :context: migrating-4_2-4_x
 :migrating-4_2-4_x:
 
 toc::[]
 
-You can migrate application workloads between {product-title} 4.2 clusters with the Cluster Application Migration (CAM) tool. The CAM tool enables you to control the migration and to minimize application downtime.
+You can migrate application workloads between {product-title} 4.2 to {product-version} with the Cluster Application Migration (CAM) tool. The CAM tool enables you to control the migration and to minimize application downtime.
```
bergerhoffer: Should "between" be "from" now here?

bergerhoffer (commented):
One inline comment, but when looking through the migration docs, I still see some references to 4.2 in this assembly that I'm guessing also need to be changed to {product-version}:

* https://migration-4_3--ocpdocs.netlify.com/openshift-enterprise/latest/migration/migrating-3-4/migrating-with-cpma.html

migration/migrating-4_2-4/migrating-application-workloads-4_2-to-4.adoc
apinnick: Yes! Good catch. Thought I'd fixed that. Will do

apinnick:
> One inline comment, but when looking through the migration docs, I still see some references to 4.2 in this assembly that I'm guessing also need to be changed to {product-version}:
> 
>     * https://migration-4_3--ocpdocs.netlify.com/openshift-enterprise/latest/migration/migrating-3-4/migrating-with-cpma.html

Ah, yes. I didn't look at CPMA because it's only for 3-4 migration. Thanks. Will fix.

bergerhoffer:
LGTM now, thanks!

PR #19069, opened 2020-01-15
Fix double words

codyhoag:
Double words were also removed in previous release branches:

#19066 (3.11)
#19067 (4.1)
#19068 (4.2)


sheriff-rh (approved):
Great catch!

PR #19075, opened 2020-01-15
Add note that collector must be defined for LF

mburke5678:
https://coreos.slack.com/archives/GGUR75P60/p1579121644086600


modules/cluster-logging-log-forwarding-configure.adoc
```diff
@@ -30,9 +30,37 @@ metadata:
     clusterlogging.openshift.io/logforwardingtechpreview: enabled <1>
   name: "instance"
   namespace: "openshift-logging"
+spec:
```
jcantrill: You really only need the collection block.  you can omit the rest

PR #18944, opened 2020-01-10
Move KB article "OpenShift 4.2 Upgrades phased roll out" contents int…

jottofar:
Improve channel documentation. First step is to incorporate the KB article https://access.redhat.com/articles/4495171 directly into docs rather than referencing and use {product-version} docs variable for easier maintenance.

PR #19088, opened 2020-01-16
Mod channel doc

jottofar:
Continue improvements to upgrade channel documentation:
- renamed section
- restructured for better flow and readability
- incorporated https://github.com/openshift/openshift-docs/pull/18944#discussion_r365286052
- other changes will be discussed in PR

### DIFF
--- a/modules/understanding-upgrade-channels.adoc
+++ b/modules/understanding-upgrade-channels.adoc
@@ -7,89 +7,85 @@
 // * updating/updating-disconnected-cluster.adoc
 
 [id="understanding-upgrade-channels_{context}"]
-= Understanding {product-title} upgrade channels
+= {product-title} Upgrade Channels and Releases
 
-In {product-title} 4.1, Red Hat introduced the concept of upgrade channels for
-recommending the appropriate upgrade versions to your cluster. Upgrade channels
-separate upgrade strategies and also are used to control the cadence of updates.
-Channels are tied to a minor version of {product-title}. For instance,
-{product-title} {product-version} channels will never include an upgrade to a
-4.5 release. This ensures administrators make an explicit decision to upgrade to
-the next minor version of {product-title}. Channels only control updates and
-have no impact on the version of the cluster you install; the
-`openshift-install` binary for a given patch level of {product-title} always
-installs that patch level.
+In {product-title} 4.1, Red Hat introduced the concept of channels for
+recommending the appropriate release versions for cluster upgrade. By controlling
+the pace of upgrades, these upgrade channels allow users to choose an upgrade
+strategy. Upgrade channels are tied to a minor version of
+{product-title}. For instance, {product-title} {product-version}
+upgrade channels will never include an upgrade to a 4.5 release. This ensures
+administrators make an explicit decision to upgrade to the next minor version of
+{product-title}. Upgrade channels only control release selection and have
+no impact on the version of the cluster you install; the `openshift-install`
+binary for a given patch level of {product-title} always installs that patch level.
 
-{product-title} {product-version}, which includes the upgrade from the previous 4.3
-release, has three upgrade channels to choose from:
+{product-title} {product-version} has the following upgrade channels to choose from:
 
 * candidate-{product-version}
 * fast-{product-version}
 * stable-{product-version}
 
-The upgrade channels contain two types of updates:
+[discrete]
+== candidate-{product-version} Channel
 
-.  General Availability Software (or GA) - These versions of {product-title} are
-fully supported and are considered production quality. You may upgrade to the general
-availability release from either of the fast and stable channels.
+The candidate-{product-version} channel will contain candidate builds for a z-stream
+({product-version}.z) release.
+Release candidates contain all the features of the product but they are not supported and
+should only be used to test feature acceptance and assist in qualifying the next version
+of {product-title}.
+A release candidate is any build (e.g. 4.3.0-rc.3, 4.3.0) that is available in the candidate
+channel.
+After a version lands in the candidate channel, it goes through more quality checks and if
+it meets the quality standard it is promoted to fast-{product-version} or stable-{product-version}
+channels.
+So if a given release is available in the candidate-{product-version} channel and also in the fast-{product-version}
+or stable-{product-version} channels, it is a Red Hat supported version.
+Additionally candidate-{product-version} may include dead end releases from which there are no or ever
+be recommended upgrades.
 
-.  Release Candidate Software (or RC) - These versions of {product-title} are
-representative of the eventual general availability release and are available only
-in the candidate-{product-version} channel. The release candidate will contain all
-the features of the product. You are allowed to upgrade from a release candidate to
-another release candidate and to upgrade from a previous minor version of
-{product-title} to the current release candidate. Release candidate builds are not
-supported by Red Hat and you will not be able to upgrade from a release candidate to
-the general availability release of {product-title}. Candidates should be used to
-test feature acceptance and assist in qualifying the next version of
-{product-title} in your infrastructure.
-+
+The candidate-{product-version} channel also allows upgrading from a previous minor version of
+{product-title} to {product-version}.
+
 [NOTE]
 ====
-Release candidates differ from the nightly builds found on try.openshift.com. You
-cannot upgrade nightly builds to nightly builds. Nightly builds are available for
-early access to features but are not upgradable or supported.
+Release candidates differ from the nightly builds found on try.openshift.com. Nightly
+builds are available for early access to features but updating to or from nightly
+builds is neither recommended nor supported. Nightly builds are not avilable in
+any upgrade channel.
 ====
 
-For GA versions, the fast and stable channels present a choice between receiving
-updates as soon as they are available or allowing Red Hat to control the rollout of
-those updates.
+[discrete]
+== fast-{product-version} Channel
+
+The fast-{product-version} channel is updated with new {product-version} patch
+versions as soon as Red Hat declares the given patch as a general availability
+release. As such, these releases are fully supported, are production quality and have
+performed well while available as a release candidate in the candidate-{product-version}
+channel from where they were promoted. Some time after a release appears in
+fast-{product-version} it will also appear in stable-{product-version}. Releases will
+never appear in stable-{product-version} before they appear in fast-{product-version}.
+
+The fast-{product-version} channel also allows upgrading from a previous minor version of
+{product-title}.
 
 [discrete]
-== fast-{product-version}
+== stable-{product-version} Channel
 
-The fast channel is updated with new {product-version} patch versions as soon as Red
-Hat declares they are generally available. Use this channel if you wish to receive
-updates as soon as they are available or for your pre-production environments when
-participating in the connected customer program. This channel will contain all
-z-stream ({product-version}.z) updates but will not suggest upgrades to the next
-minor release (4.5.z) when the next minor release is available.
+Like the fast-{product-version} channel, the stable-{product-version} channel will
+only contain releases that have been declared general availability are therefore
+fully supported. However the stable-{product-version} channel will gradually roll out
+releases to customers based on data from our SRE teams, support services, and
+pre-production and production environments that participate in our connected customer
+program, rather than being immediately available as they are in the fast-{product-version}
+channel. For patch and CVE fixes this delay can range from several hours to a day
+and allows an extra period of assessment in how the software performs.
 
-[discrete]
-== stable-{product-version}
-
-The stable channel will contain updates on a time delay as they are gradually
-rolled out to customers based on data from our SRE teams, support services, and
-pre-production and production environments that participate in our connected
-customer program, rather than being immediately available as they are in the fast
-channel. For patch and CVE fixes this can range from several hours to a day and
-allows an extra period of assessment in how the software performs. If issues are
-detected during rollout, upgrades to that version may be blocked in both the fast
-and stable channels, and a new version may be introduced that will be the new
-preferred upgrade target.
-
-Customers can improve this process by configuring pre-production systems on the
-fast channel, production systems on the stable channel, and participating in Red
-Hat’s connected customer program - this allows Red Hat to observe the impact of
-updates on your specific hardware and software configurations. Future releases may
-improve or alter the pace at which updates move from the fast to the stable channels.
-
-If issues are discovered with an upgrade between patch levels, Red Hat may withdraw
-that suggested upgrade for affected versions. A newer patch would become available
-in the appropriate channels and be suggested for upgrade.
+The stable-{product-version} channel also allows upgrading from a previous minor version of
+{product-title}.
 
 [discrete]
-== Upgrade version paths
+== Upgrade Version Paths
 
 {product-title} maintains an upgrade recommendation service that understands the
 version of {product-title} you have installed as well as the path to take within
@@ -101,19 +97,48 @@ following in the fast-{product-version} channel:
 * {product-version}.3
 * {product-version}.4
 
-The service only recommends upgrades that have been tested and have no known issues.
+The service only recommends upgrades that have been tested and have no serious issues.
 If you are on {product-version}.1 and {product-title} is allowing you to select
 {product-version}.4, then it is safe for you to go from .{product-version}.1 to .{product-version}.4. Likewise,
 the absence of {product-version}.2 may be due to a CVE that was fixed in {product-version}.3 and Red Hat no
-longer suggests upgrading to a known vulnerable version. If an issue is found that
-results in a new version being retracted from the recommendations, Red Hat will
-release a new version that is capable of upgrading from all necessary versions,
-including the retracted version.
+longer suggests upgrading to a known vulnerable version.
+
+Update stability depends on your channel. The presence of an update recommendation in
+the candidate-{product-version} channel does not imply that the update is supported.
+It means that no serious issues have been found with the update yet, but there may
+not be significant traffic through the update to suggest stability. The presence of
+an update recommendation in the fast-{product-version} or stable-{product-version}
+channels is a declaration that the update is fully supported while it is in the
+channel. While releases will never be removed from a channel, update recommendations
+which exhibit serious issues will be removed from all channels. Updates initiated
+after the update recommendation has been removed may not be supported.
+
+Red Hat will eventually provide supported update paths from any supported (fast-{product-version}
+or stable-{product-version}) release to the latest release in {product-version}.z,
+although there may be delays while safe paths away from troubled releases are
+constructed and verified.
 
 [discrete]
-== Disconnected clusters
+== Fast and Stable Channel Usage and Strategies
 
-Customers which have chosen to not be connected to Red Hat and are curating their
+The fast-{product-version} and stable-{product-version} channels present a choice between receiving
+general availability releases as soon as they are available or allowing Red Hat to
+control the rollout of those updates. If issues are detected during rollout or at a
+later time, upgrades to that version may be blocked in both the fast-{product-version} and
+stable-{product-version} channels, and a new version may be introduced that will be the new
+preferred upgrade target.
+
+Customers can improve this process by configuring pre-production systems on the
+fast-{product-version} channel, production systems on the stable-{product-version} channel,
+and participating in Red Hat’s xref:../support/remote_health_monitoring/about-remote-health-monitoring.adoc[connected customer program]. This program allows Red
+Hat to observe the impact of updates on your specific hardware and software
+configurations. Future releases may improve or alter the pace at which updates move
+from the fast-{product-version} to the stable-{product-version} channel.
+
+[discrete]
+== Restricted Network Clusters
+
+Customers who have chosen to not be connected to Red Hat and are controlling their
 own {product-title} container image content manually should consult the Red Hat
 errata associated with product releases and note any comments impacting upgrades.
 During upgrade the user interface may caution about switching between these versions
@@ -121,10 +146,20 @@ and it is up to the customer to ensure they have correctly selected the appropri
 version before bypassing those cautions.
 
 [discrete]
-== Switching between channels
+== Switching Between Channels
 
-It is supported for customers to switch between the fast and stable channel at any
-time. Channels only offer suggested upgrades, and will never suggest a dangerous
-upgrade. If you switch to the candidate channel after installing from a GA version,
-you will see a warning the current version is not recognized, and you can safely
-switch back to a GA channel.
+It is supported for customers to switch from the stable-{product-version} channel to
+the fast-{product-version} at any time. Customers can also switch to the
+candidate-{product-version} at any time but should be aware that some releases in the
+candidate-{product-version} channel may be release candidates and therefore not
+supported. A customer can switch from candidate-{product-version} to fast-{product-version}
+if their current release is a general availability release. Customers can always
+switch from fast-{product-version} to stable-{product-version}, although there may
+be a delay of up to a day if their current release was recently promoted to
+fast-{product-stable} while they wait for the release to be promoted to
+stable-{product-version}. If you change to a channel that does not include your
+current release, an alert will fire, no updates will be recommended, and you can
+safely change back to your original channel at any point.
+
+Please refer to the <<candidate-{product-version} Channel>> section above to understand the
+distinction between a release candidate and a general avilability release.


modules/understanding-upgrade-channels.adoc
```diff
@@ -7,17 +7,129 @@
...
+
+GA release versions are only availble in fast-{product-version} or stable-{product-version}.
+
+More details about each channel and their usage are in the following sections.
+
+= candidate-{product-version}
+
+The candidate channel allows upgrading from an RC version to a later RC version and
+upgrading from a previous minor version of {product-title} to the current RC version.
+You cannot upgrade from an RC release to a GA release of {product-title}.
```
jottofar: This is misleading, because you may also be able to update to GA from an RC release in the candidate channel.  See https://github.com/openshift/openshift-docs/pull/18944#discussion_r365289359

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,17 +7,129 @@
...
+and it is up to the customer to ensure they have correctly selected the appropriate
+version before bypassing those cautions.
+
+= Switching Between Channels
+
+It is supported for customers to switch between the fast and stable channel at any
+time. Channels only offer suggested upgrades, and will never suggest a dangerous
+upgrade. If you switch to the candidate channel after installing from a GA version,
+you will see a warning the current version is not recognized, and you can safely
+switch back to a GA channel.
```
jottofar: This is not true, all of our GA releases are also in the candidate channel. Ref https://github.com/openshift/openshift-docs/pull/18944#discussion_r365310079

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,79 +7,89 @@
 // * updating/updating-disconnected-cluster.adoc
 
 [id="understanding-upgrade-channels_{context}"]
-= Understanding {product-title} upgrade channels
+= {product-title} Upgrade Channels and Update Types
```
wking: I like this, because all of our docs should be about understanding.  But the file name also has `understanding-`.  Can we fix that too?  @kalexand-rh, is there history/policy we're missing?

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,79 +7,89 @@
...
-In {product-title} 4.1, Red Hat introduced the concept of upgrade channels for
-recommending the appropriate upgrade versions to your cluster. Upgrade channels
-separate upgrade strategies and also are used to control the cadence of updates.
-Channels are tied to a minor version of {product-title}. For instance,
-{product-title} {product-version} channels will never include an upgrade to a
+In {product-title} 4.1, Red Hat introduced the concept of Upgrade Channels for
+recommending the appropriate upgrade versions to your cluster. Upgrade Channels
+separate upgrade strategies and are also used to control the cadence of updates.
+Upgrade Channels are tied to a minor version of {product-title}. For instance,
+{product-title} {product-version} Upgrade Channels will never include an upgrade to a
```
wking: I personally prefer a single "update channels" (or "upgrade channels"), followed by "channels".  But whatever, folks will understand either way.

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,79 +7,89 @@
...
 4.5 release. This ensures administrators make an explicit decision to upgrade to
-the next minor version of {product-title}. Channels only control updates and
+the next minor version of {product-title}. Upgrade Channels only control updates and
 have no impact on the version of the cluster you install; the
 `openshift-install` binary for a given patch level of {product-title} always
 installs that patch level.
 
-{product-title} {product-version}, which includes the upgrade from the previous 4.3
-release, has three upgrade channels to choose from:
+The updates within an Upgrade Channel are designated as one of two types. These
```
wking: I'd rather call the nodes "releases" and not "updates".  If we only ever cut a single release (with no update edges), it would still be a release, but it wouldn't update anything.

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,79 +7,89 @@
...
+Update Types are defined as:
 
-* candidate-{product-version}
-* fast-{product-version}
-* stable-{product-version}
+.  General Availability (GA) Software - {product-title} GA releases are fully
+supported and production quality.
+
+.  Release Candidate (RC) Software - {product-title} RC releases are not supported
```
wking: "RC releases" == "Release Candidate releases", which stutters.  I'd rather spell out "Release candidates" and use "candidates" for short.  "RCs" in a pinch, but I'm generally anti-acronym and don't think "RCs" is worth the reduction in characters ;)

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,79 +7,89 @@
...
-supported by Red Hat and you will not be able to upgrade from a release candidate to
-the general availability release of {product-title}. Candidates should be used to
-test feature acceptance and assist in qualifying the next version of
-{product-title} in your infrastructure.
 +                                                                                    
 [NOTE]                                                                               
 ==== 
-Release candidates differ from the nightly builds found on try.openshift.com. You
+RC releases differ from the nightly builds found on try.openshift.com. You
 cannot upgrade nightly builds to nightly builds. Nightly builds are available for
```
wking: You might be able to update between nightlies, or you may not.  How about:

> Updating to or from nightly builds is neither recommended nor supported.

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,79 +7,89 @@
...
 ==== 
 
-For GA versions, the fast and stable channels present a choice between receiving
-updates as soon as they are available or allowing Red Hat to control the rollout of
-those updates.
+{product-title} {product-version} has three Upgrade Channels to choose from:
+
+* candidate-{product-version}
+* fast-{product-version}
+* stable-{product-version}
```
wking: This is different from the period-marked lists above (e.g. GA vs. RC).  Is it valid asciidoc?

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,79 +7,89 @@
...
-Release candidates differ from the nightly builds found on try.openshift.com. You
+RC releases differ from the nightly builds found on try.openshift.com. You
 cannot upgrade nightly builds to nightly builds. Nightly builds are available for
 early access to features but are not upgradable or supported.
 ==== 
 
-For GA versions, the fast and stable channels present a choice between receiving
-updates as soon as they are available or allowing Red Hat to control the rollout of
-those updates.
+{product-title} {product-version} has three Upgrade Channels to choose from:
```
wking: "at least the following"?  Folks can count 'em, and it leaves room for more and one less thing that could go stale.

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,79 +7,89 @@
...
+
+* candidate-{product-version}
+* fast-{product-version}
+* stable-{product-version}
+
+RC release versions are only available in candidate-{product-version}.
+
+GA release versions are only availble in fast-{product-version} or stable-{product-version}.
```
wking: I think this may be confusing.  Actually, I think the whole RC vs. GA thing may be confusing.  Trying to clarify:

* We cut releases with names.
* At the moment, sometimes those names are 4.y.z-0.nightly-YYYY-MM-DD-HHMMSS (nightlies), 4.y.z-rc.n (RCs?), and 4.y.z (GAs?).
* When they're cut, nightlies are nightlies.  They are never going to end up in any channel.
* When they're cut, RCs are not in any channel.  If we like them, we add them to the candidate channel and they become release candidates.  We will never add them to fast/stable.
* When they're cut, GAs are not in any channel.  If we like them, we add them to the candidate channel and they become release candidates.  If we still like them, we add them to fast/stable and they become GA releases.  But a release image is not automatically GA because we happened to name it 4.2.0 (or whatever).

Can we drop the whole "types of updates" thing and just talk about channels?

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,79 +7,89 @@
...
+
+More details about each channel and their usage are in the following sections.
```
wking: Do we need this line?  Without it, folks would see the candidate header, and figure it out?  Or maybe the entries in the channel list above^ should link down to the channel-specific subsections below?  I dunno how that works in asciidoc, but if it sounds appealing I'm sure we can find out.

modules/understanding-upgrade-channels.adoc
wking: Suggested rewording:

> The candidate-{product-version} channel is the first {product-version} channel to which releases are promoted.  If a release performs well as a candidates [would be nice to link to Telemetry-reporting section here], it is later promoted to fast-{product-version} [would be nice to link to the fast section here], but never removed from candidate-{product-version}.  candidate-{product-version} may also contain {product-version-minus-one} releases which are being considered as jumping-off points for {product-version-minus-one} to {product-version} updates.  candidate-{product-version} will never contain nightly releases (which will never belong to any channel) or releases from {product-version-plus-one} or later.

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,79 +7,89 @@
...
-z-stream ({product-version}.z) updates but will not suggest upgrades to the next
-minor release (4.5.z) when the next minor release is available.
+The fast channel, containing only GA releases, is updated with new
+{product-version} patch versions as soon as Red Hat declares they are generally
+available. Use this channel if you wish to receive updates as soon as they are
+available or for your pre-production environments when participating in the connected
+customer program. This channel will contain all z-stream ({product-version}.z)
+updates but will not suggest upgrades to the next minor release (4.5.z) when the next
+minor release is available. This channel also includes the upgrade from the previous
+4.3 release.
```
wking: Rewording ideas; take anything that sounds useful:

> The fast-{product-version} channel is the second {product-version} channel to which releases are promoted if they have performed well in candidate-{product-version} [would be nice to link to the candidate section here].  Membership in fast-{product-version} is a declaration that the release is supported.  fast-{product-version} will never contain {product-version}.z-rc.n releases.  Some time after a release appears in fast-{product-version}, it will also appear in stable-{product-version} [would be nice to link to the stable section here].  Releases will never appear in stable-{product-version} before they appear in fast-{product-version}.

wking: I don't think we need to talk about previous/next minor here, because we already talked about that in the candidate section.

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,79 +7,89 @@
...
-detected during rollout, upgrades to that version may be blocked in both the fast
-and stable channels, and a new version may be introduced that will be the new
-preferred upgrade target.
+The stable channel, containing only GA releases, will contain updates on a time delay
+as they are gradually rolled out to customers based on data from our SRE teams,
+support services, and pre-production and production environments that participate in
+our connected customer program, rather than being immediately available as they are
+in the fast channel. For patch and CVE fixes this can range from several hours to a
+day and allows an extra period of assessment in how the software performs. This
+channel also includes the upgrade from the previous 4.3 release.
```
wking: Rewording ideas; take anything that sounds useful:

> The stable-{product-version} channel is the {product-version} channel to which releases are promoted after they have been promoted to fast-{product-version} [would be nice to link to the fast section here]. Membership in stable-{product-version} is, like membership in fast-{product-version}, a declaration that the release is supported. Releases will enter stable-{product-version} with a delay after they are promoted to fast-{product-version}.  For patch and CVE fixes, the delay can range from several hours to a day and allows an extra period of assessment in how the software performs.

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,79 +7,89 @@
...
+
+= Fast and Stable Channel Usage and Strategies
+
+The fast-{product-version} and stable-{product-version} channels present a choice between receiving
+GA updates as soon as they are available or allowing Red Hat to control the rollout
```
wking: s/GA/supported/

wking: and s/updates/releases/

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,79 +7,89 @@
...
-preferred upgrade target.
+The stable channel, containing only GA releases, will contain updates on a time delay
+as they are gradually rolled out to customers based on data from our SRE teams,
+support services, and pre-production and production environments that participate in
+our connected customer program, rather than being immediately available as they are
+in the fast channel. For patch and CVE fixes this can range from several hours to a
+day and allows an extra period of assessment in how the software performs. This
+channel also includes the upgrade from the previous 4.3 release.
+
+= Fast and Stable Channel Usage and Strategies
```
wking: This section should go below the *Upgrade Version Paths* section, because you need to understand how updates are supposed to be handled before you can decide between fast/stable.  The content above this about the channels themselves only talks about relesases, and hasn't yet said anything about updates.

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,79 +7,89 @@
...
 
 Customers can improve this process by configuring pre-production systems on the
 fast channel, production systems on the stable channel, and participating in Red
-Hat’s connected customer program - this allows Red Hat to observe the impact of
-updates on your specific hardware and software configurations. Future releases may
+Hat’s connected customer program. This program allows Red Hat to observe the impact
+of updates on your specific hardware and software configurations. Future releases may
 improve or alter the pace at which updates move from the fast to the stable channels.
 
 If issues are discovered with an upgrade between patch levels, Red Hat may withdraw
```
wking: Rewording ideas for the *Upgrade Version Paths* section below:

> {product-title} maintains an upgrade recommendation service that understands the version of {product-title} you have installed as well as releases to which you may safely update.  You can imagine seeing the following in the fast-{product-version} channel:
>
> * {product-version}.0
> * {product-version}.1
> * {product-version}.3
> * {product-version}.4
>
> The service only recommends upgrades that have been tested and have no serious issues.  If you are on {product-version}.1 and {product-title} is allowing you to select {product-version}.4, then it is safe for you to go from .{product-version}.1 to .{product-version}.4. Likewise, the absence of {product-version}.2 may be due to a CVE that was fixed in {product-version}.3 and Red Hat no longer suggests upgrading to a known vulnerable version.
>
> Update stability depends on your channel.  The presence of an update recommendation in the candidate-{product-version} channel does not imply that the update is supported.  It means that no serious issues have been found with the update yet, but there may not be significant traffic through the update to suggest stability.  The presence of an update recommendation in the fast-{product-version} or stable-{product-version} channels is a declaration that the update is supported while it is in the channel.  Like releases in the stable-{product-version} channel, update recommendations in the stable-{product-version} channel will be promoted from fast-{product-version} with a delay that can range from can range from several hours to a day.  While releases will never be removed from a channel, update recommendations which exhibit serious issues will be removed from all channels.  Updates initiated after the update recommendation has been removed may not be supported.
>
> Red Hat will eventually provide supported update paths from any supported (fast-{product-version} or stable-{product-version}) release to the latest release in {product-version}.z, although there may be delays while safe paths away from troubled releases are constructed and verified.

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,79 +7,89 @@
...
+be blocked in both the fast and stable channels, and a new version may be introduced
+that will be the new preferred upgrade target.
 
 Customers can improve this process by configuring pre-production systems on the
 fast channel, production systems on the stable channel, and participating in Red
-Hat’s connected customer program - this allows Red Hat to observe the impact of
-updates on your specific hardware and software configurations. Future releases may
+Hat’s connected customer program. This program allows Red Hat to observe the impact
+of updates on your specific hardware and software configurations. Future releases may
 improve or alter the pace at which updates move from the fast to the stable channels.
```
wking: Rewording ideas for the *Switching Between Channels* section below:

> It is supported for customers to switch from the stable-{product-version} channel to fast-{product-versoin} at any time.  Customers may also switch to the candidate-{product-version} at any time without waiving support, but subsequent updates in the candidate-{product-version} channel may waive support (e.g. if that update was not also in the fast-{product-version} channel when it was initiated).  Customers may be able to switch from candidate-{product-version} to fast-{product-version}, depending on whether their current release is supported or not.  Customers will always be able to switch from fast-{product-version} to stable-{product-version}, although there may be a delay of up to a day if their release was recently promoted to fast-{product-stable} while they wait for the release to be promoted to stable-{product-version} for their cluster.  If you change to a channel that does not include your current release, an alert will fire, no updates will be recommended, and you can safely change back to your original channel at any point.

modules/understanding-upgrade-channels.adoc
jottofar: I don't feel strongly either way but what I was attempting to do was identify Upgrade Channel to the reader (customer) as a proper noun and then have it become familiar since Upgrade Channels have become a central and all-important feature of OpenShift and our upgrade strategy.

modules/understanding-upgrade-channels.adoc
jottofar: This is for a numbered list. Based on your below comment about "counting them" which I agree with, I'll change this to a simple bullet list.

modules/understanding-upgrade-channels.adoc
LalatenduMohanty: I do not have a personal preference either. However we have to make sure customer/reader does not get confused between the two. So if we have been using `upgrade` word then we should just keep using it.

modules/understanding-upgrade-channels.adoc
LalatenduMohanty: I do not think this is very confusing. Can this be improved further? of course. But this is a way better improvement to what we had previously and I will be happy to take this in and then gradually improve.

modules/understanding-upgrade-channels.adoc
LalatenduMohanty: Couple of comments
1. We should not use `release candidates` instead of `RC`. Reading `RC` might not be user friendly to may folks. 
2. We just merge 
```
{product-title} {product-version} has three Upgrade Channels to choose from:

* candidate-{product-version}
* fast-{product-version}
* stable-{product-version}
```
with this section. I do not think having information in two places going to improve the reading experience but on the contrary it forces the reader to scroll up again and remember what he/she read before.

LalatenduMohanty: Actually I find @wking's suggested comment has a lot of information which would confuse folks. The initial sections should be quick read for people then we want to extra information we should add them in the later part of the document somehow.

modules/understanding-upgrade-channels.adoc
LalatenduMohanty: We should not say `or for your pre-production environments when participating in the connected
customer program`. Because customer should take that decision. Once we have phased rollout with autoupdate, hopefully they will see the value in it and out for stable channel.

LalatenduMohanty: @wking 's sugegstion `The fast-{product-version} channel is the {product-version} channel to which releases are promoted if they have performed well in candidate-{product-version} [would be nice to link to the candidate and Telemetry sections here]. Membership in fast-{product-version} is a declaration that the release is supported. ` should not be here. As by now we should have clearly mentioned that fast and stable are supported channels.

modules/understanding-upgrade-channels.adoc
LalatenduMohanty: For customers and business `fast` and `stable` gets the releases at the same time. Technically we should not differentiate between them with respect to publishing them in to respective channels. If we do not agree on this we should ask the release-admins for clarification.

modules/understanding-upgrade-channels.adoc
LalatenduMohanty: Using `may be` for certain situations is a bad idea IMO. We should be very clear about what is possible and which is not.  For example `Customers may be able to switch from candidate-{product-version} to fast-{product-version}, depending on whether their current release is supported or not.` is not something we want to communicate as part of the document text. May be we can add a section which talks about these corner case exceptions.

modules/understanding-upgrade-channels.adoc
LalatenduMohanty: Some of these information are also present previously, we should not repeat these as it confuses the reader


> Update stability depends on your channel. The presence of an update recommendation in the candidate-{product-version} channel does not imply that the update is supported. It means that no serious issues have been found with the update yet, but there may not be significant traffic through the update to suggest stability. The presence of an update recommendation in the fast-{product-version} or stable-{product-version} channels is a declaration that the update is supported while it is in the channel. Like releases in the stable-{product-version} channel, update recommendations in the stable-{product-version} channel will be promoted from fast-{product-version} with a delay that can range from can range from several hours to a day. While releases will never be removed from a channel, update recommendations which exhibit serious issues will be removed from all channels. Updates initiated after the update recommendation has been removed may not be supported.

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,17 +7,129 @@
...
+
+* {product-version}.0
+* {product-version}.1
+* {product-version}.3
+* {product-version}.4
+
+The service only recommends upgrades that have been tested and have no known issues.
+If you are on {product-version}.1 and {product-title} is allowing you to select
+{product-version}.4, then it is safe for you to go from .{product-version}.1 to .{product-version}.4. Likewise,
+the absence of {product-version}.2 may be due to a CVE that was fixed in {product-version}.3 and Red Hat no
```
LalatenduMohanty: Just not CVE, there might be some bugs which were discovered and applicable to some clusters

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,17 +7,129 @@
...
+longer suggests upgrading to a known vulnerable version. If an issue is found that
+results in a new version being retracted from the recommendations, Red Hat will
+release a new version that is capable of upgrading from all necessary versions,
+including the retracted version.
+
+= Disconnected Clusters
+
+Customers which have chosen to not be connected to Red Hat and are curating their
+own {product-title} container image content manually should consult the Red Hat
```
LalatenduMohanty: I am not sure of `curating` is the correct word here. Somehow it gives an impression that customers can create  own container image. But that's not supported right?

modules/understanding-upgrade-channels.adoc
LalatenduMohanty: AFAIK we should not suggest customers to switch between candidate and stable/fast. Candidate is only for testing which QE also tests.

LalatenduMohanty (changes requested):

modules/understanding-upgrade-channels.adoc
wking: Anything available in fast/stable will also be available in candidate, so I don't like "GA release versions are only availble in fast..." (also "availble" typo).  But sure, I'm fine punting and coming back to this later.

modules/understanding-upgrade-channels.adoc
wking: > The initial sections should be quick read for people then we want to extra information we should add them in the later part of the document somehow.

Maybe "this is what life in `stable-{product-version}` will be like" (90+% of admins) with later sections for fast and candidate?

modules/understanding-upgrade-channels.adoc
wking: > Technically we should not differentiate between them with respect to timing of publishing them in to respective channels.

Maybe releases at the same time, but update recommendation edges are delayed in stable for phased rollouts.

modules/understanding-upgrade-channels.adoc
wking: > May be we can add a section which talks about these corner case exceptions.

I think the *Switching Between Channels* section is that corner-case section.

modules/understanding-upgrade-channels.adoc
wking: > Some of these information are also present previously...

I would rather lead with update discussion, and only talk about release nodes incidentally.

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,17 +7,156 @@
...
+supported. A customer can switch from candidate-{product-version} to fast-{product-version}
+if their current release is a general availability release. Customers can always
+switch from fast-{product-version} to stable-{product-version}, although there may
+be a delay of up to a day if their current release was recently promoted to
+fast-{product-stable} while they wait for the release to be promoted to
+stable-{product-version}. If you change to a channel that does not include your
+current release, an alert will fire, no updates will be recommended, and you can
+safely change back to your original channel at any point.
+
+Please refer to candidate-{product-version} section above to undertsand the
```
jottofar: @kalexand-rh how do I link to a section in the same adoc file?

jottofar:
@LalatenduMohanty @wking this push contains changes resulting from prior comments and what we discussed in our call yesterday. It does not consider @wking comments late Friday.

LalatenduMohanty (changes requested):

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,17 +7,156 @@
...
-Channels are tied to a minor version of {product-title}. For instance,
-{product-title} {product-version} channels will never include an upgrade to a
-4.4 release. This ensures administrators make an explicit decision to upgrade to
-the next minor version of {product-title}. Channels only control updates and
-have no impact on the version of the cluster you install; the
+= {product-title} Upgrade Channels and Releases
+
+In {product-title} 4.1, Red Hat introduced the concept of Upgrade Channels for
+recommending the appropriate release versions for cluster upgrade. Upgrade Channels
+separate upgrade strategies and are also used to control the cadence of upgrades.
```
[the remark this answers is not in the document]
jottofar: How about:
In OCP 4.1, Red Hat introduced the concept of channels for recommending the appropriate release versions for cluster upgrade. By controlling the pace of upgrades, these Upgrade Channels allow users to choose an upgrade strategy.

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,17 +7,156 @@
...
-See link:https://access.redhat.com/articles/4495171[OpenShift 4.2 Upgrades phased roll out] for more information on the types of updates and upgrade channels.
+{product-title} {product-version} has the following Upgrade Channels to choose from:
+
+.  candidate-{product-version}
+.  fast-{product-version}
+.  stable-{product-version}
+
+= candidate-{product-version}
+
+The candidate-{product-version} channel will contain all z-stream ({product-version}.z)
```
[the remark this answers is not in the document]
jottofar: I'm sure this was stated in our discussion, probably by @wking. Perhaps it needs to be qualified or softened?

vikram-redhat:
@jottofar is this ready for QE review?

@sferich888 ok from your POV?

@kalexand-rh FYI.

modules/understanding-upgrade-channels.adoc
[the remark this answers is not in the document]
jottofar: @LalatenduMohanty have a look at responses to your most recent comments

jottofar:
@wking may still have comments on most recent draft and then I will push up any resulting changes.

jottofar:
@wking may still have comments on most recent draft and then I will push up any resulting changes.

LalatenduMohanty:
@jottofar I will take another look and add my final comments.

jottofar:
@wking can you take a final look as well so we can get this wrapped up.

modules/understanding-upgrade-channels.adoc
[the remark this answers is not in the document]
LalatenduMohanty: Sounds fine. Please use small case with s/`Upgrade Channels`/`upgrade channels`/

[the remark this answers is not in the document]
jottofar: My intent here is to get users to think of Upgrade Channels as  a real "thing" with a name and therefore I use a proper noun. Henceforth in the documentation I can refer to it using its name. "channels" is generic but we're describing a very specific thing here - a set of channels used to do upgrades. Anything important should have a name. If we don't like Upgrade Channels we should come up with something else.

modules/understanding-upgrade-channels.adoc
wking: > AFAIK we should not suggest customers to switch between candidate and stable/fast. Candidate is only for testing which QE also tests.

I have no problem with suggesting folks move in stable directions.  For example, if you launch a 4.3.0-rc.0 cluster, change to candidate-4.3, update to 4.3.0, change to stable-4.3, and update in stable from there.  I'm not clear on whether we'd support that cluster, or if it would be tainted by at one point running the unsupported 4.3.0-rc.0.  But I'd rather address whether clusters can come in and out of support explicitly, instead of dodging by not discussion channel changes out of candidate.

modules/understanding-upgrade-channels.adoc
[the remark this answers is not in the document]
wking: > ..."channels" is generic...

In this section, "channels" is only ever going to be update channels.  I [don't feel a need][1] to make these proper nouns, but don't have a problem with proper nouns either.

> If we don't like Upgrade Channels we should come up with something else.

I'm [fine][1] with any case permutations on "upgrade channels" ;).

[1]: https://github.com/openshift/openshift-docs/pull/19088#discussion_r367762433

modules/understanding-upgrade-channels.adoc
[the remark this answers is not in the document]
wking: > @LalatenduMohanty have a look at responses to your most recent comments

Which comments were these?  Was it related to this sentence about z-stream releases?

> Is this correct?

The previous text made false claims about changing from fast/stable to the candidate channel (see [this thread][1]).  I want to make it clear that you can always change your cluster along stable-{v} -> fast-{v} -> candidate-{v} without having to worry about the "current version is not recognized" warning.  That is distinct from changing your cluster along candidate-{v} -> fast-{v} -> stable-{v}, where you might get the "current version is not recognized" warning.  Another constraint is that we never remove a release from a channel, so if at some point in the past the release was in stable-{v}, you can set your cluster to any of stable-{v}, fast-{v}, or candidate-{v} without risking the "current version is not recognized" warning.  And I'm fine if any or none of my explanation here makes it into the docs suggested by this PR, as long as the old false claim gets removed.

[1]: https://github.com/openshift/openshift-docs/pull/18944#discussion_r365310079

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,17 +7,156 @@
...
+releases. If a given release is only available in the candidate-{product-version}
+channel and not available in either the fast-{product-version} or stable-{product-version}
+channels, it is considered a release candidate. Although release candidates contain
+all the features of the product, they are not supported and should only be used to
+test feature acceptance and assist in qualifying the next version of 
+{product-title} in your infrastructure. Additionally, a release candidate can only be
+upgraded to another release candidate version and, if one does not exist, is not
```
wking: You don't define "release candidate version".  I think the main issue this sentence is trying to convey is that we will happily leave dead-ends in the release candidate graph, while fast/stable have support-level guarantees about us (eventually) providing exits from any release in that channel to get you to the current channel tip.  That's a somewhat awkward point to address here, because we haven't talked about fast/stable yet.  I'd rather talk about stable, then fast, then candidate, because we expect most admins to only care about stable.  But if you want a narrower pivot, maybe reword this sentence to:

> Additionally, the candidate-{product-version} may include dead end releases from which there are never recommended upgrades.

or some such?

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,17 +7,156 @@
...
+upgradable.
+
+The candidate-{product-version} channel also allows upgrading from a previous minor version of
+{product-title} to the current release candidate version.
```
wking: "the current release candidate version" -> "{product-version}".

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,17 +7,156 @@
...
+builds are available for early access to features but updating to or from nightly
+builds is neither recommended nor supported. Nightly builds are not avilable in
+any Upgrade Channels.
+==== 
+
+= fast-{product-version}
+
+The fast-{product-version} channel is updated with new {product-version} patch
+versions as soon as Red Hat declares the given patch as a general availability
+release. As such, these releases are fully supported, are production quality and have
```
wking: nit: do we have a [serial comma][1] docs policy?

[1]: https://en.wikipedia.org/wiki/Serial_comma

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,17 +7,156 @@
...
+performed well while available as a release candidate in the candidate-{product-version}
+channel from where they were promoted. Some time after a release appears in
+fast-{product-version}, and therefore after the release has had further use, it will
```
wking: > ...and therefore after the release has had further use...

Can we drop this phrase?  With phased stable rollouts, most of the stable clusters would benefit from earlier stable clusters having explored the update.  But it's possible to have zero fast clusters on the source release and then the first stable cluster in a phased release is getting prodded off the edge with nothing more than the information we had from candidate.

modules/understanding-upgrade-channels.adoc
wking: Currently this example reads like "there was a CVE in {product-version}.2, so we dropped it from fast-{product-version}".  But we drop edges, not releases, when we discover new issues with existing releases.  We skip releases when we have build trouble (e.g. openshift/cincinnati-graph-data#33).  I'm fine talking about either or both of those, but I think it would help if this example graph also included edges.  Visual graphs are nice, like [the SVGs here][1], but I'd also be ok with:

```asciidoc
.  {product-version}.0, with recommended updates to {product-version}.3 and {product-version}.4
.  {product-version}.1, with recommended updates to {product-version}.4
.  {product-version}.3, with recommended updates to {product-version}.4
.  {product-version}.4
```

where an issue had been discovered with {product-version}.1 that lead us to pull its incoming edges.  Then we cut {product-version}.3 with improvements on {product-version}.0, but no safe {product-version}.1 -> {product-version}.3 update.  Then we cut {product-version}.4 with a safe update to get everyone off {product-version}.1.


[1]: https://github.com/openshift/cincinnati/blob/85c62120e5fc6779ed48540c9b06bf17e7652165/docs/design/openshift.md#channels

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,17 +7,156 @@
...
+
+Update stability depends on your channel. The presence of an update recommendation in
+the candidate-{product-version} channel does not imply that the update is supported.
+It means that no serious issues have been found with the update yet, but there may
+not be significant traffic through the update to suggest stability. The presence of
+an update recommendation in the fast-{product-version} or stable-{product-version}
+channels is a declaration that the update is fully supported while it is in the
+channel. Like releases in the stable-{product-version} channel, update
+recommendations in the stable-{product-version} channel will be promoted from
+fast-{product-version} with a delay that can range from several hours to a day.
```
wking: I think this sentence can be dropped, since it's covered above in the stable-{product-version} section and doesn't speak to the stability of releases or update recommendations.

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,17 +7,156 @@
...
+While releases will never be removed from a channel, update recommendations which
+exhibit serious issues will be removed from all channels. Updates initiated after
+the update recommendation has been removed may not be supported.
```
wking: I don't have any suggested wording to tighten this down (and it matches what I recommended [over here][1]), but the "may" is wishy.  My expectation is that you're not supported if you initiate an update after we pull the edge from our graphs, but that pinning down that time is going to be hard.  For example the CVO polls every 5m or some such, so if your CVO fetches the graph at 12:00Z, we pull the edge at 12:01Z, and you initiate the update based on your cached graph at 12:03Z, we are not going to hang you out to dry.  If you have to sneakernet graphs into a restricted-network environment, higher latency is still reasonable.  But if you initiate the update over a week after we pulled the edge and your cluster breaks, you're probably on your own.  Do we want to set an official cap on latency?  Are we ok with the wishy "may"?

[1]: https://github.com/openshift/openshift-docs/pull/19088/#discussion_r367776894

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,17 +7,156 @@
...
+Customers can improve this process by configuring pre-production systems on the
+fast-{product-version} channel, production systems on the stable-{product-version} channel,
+and participating in Red Hat’s connected customer program. This program allows Red
+Hat to observe the impact of updates on your specific hardware and software
+configurations. Future releases may improve or alter the pace at which updates move
+from the fast-{product-version} to the stable-{product-version} channel.
+
+If issues are discovered with a release between patch levels, Red Hat may withdraw
+that suggested upgrade path for affected release versions. A newer release would
+become available in the appropriate channels and be suggested as an upgrade path.
```
wking: This paragraph overlaps with the earlier "If issues are detected during rollout..." line and the even earlier "While releases will never be removed from a channel, update recommendations which exhibit serious issues will be removed from all channels." paragraph.  I think we can drop this one.

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,17 +7,156 @@
...
+
+= Disconnected Clusters
```
wking: nit: "Disconnected" -> "Restricted-network" to match our existing docs (you can have clusters without access to Red Hat's Cincinnati that are not completely disconnected).

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,17 +7,156 @@
...
+
+Customers which have chosen to not be connected to Red Hat and are curating their
```
wking: nit: "Customers which" -> "Customers who"?

modules/understanding-upgrade-channels.adoc
wking: > Somehow it gives an impression that customers can create own container image. But that's not supported right?

Yeah, although I think we want "container image content manually" -> "update recommendations", because what they're owning is the update graph, not the release images.

I don't have strong feelings on "curating" vs. support, although it does imply that the official Red Hat graph is something that customers might want to polish.  While that is occasionally true, any polishing should happen [in the graph-data repo][1], where everyone can submit PRs, customers included.  What restricted-network users are doing is owning maintenance on their update graph, which gives all the power, responsibility, and headache that forking a project usually gives.  "curating" -> "maintaining" or some such might do a better job of getting that balance across.  But I'm personally fine leaving "curating" if folks like it more.

[1]: https://github.com/openshift/cincinnati-graph-data/

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,17 +7,156 @@
...
+own {product-title} container image content manually should consult the Red Hat
+errata associated with product releases and note any comments impacting upgrades.
+During upgrade the user interface may caution about switching between these versions
+and it is up to the customer to ensure they have correctly selected the appropriate
+version before bypassing those cautions.
+
+= Switching Between Channels
+
+It is supported for customers to switch from the stable-{product-version} channel to
+the fast-{product-versoin} at any time. Customers can also switch to the
```
wking: nit: "versoin" -> "version"

modules/understanding-upgrade-channels.adoc
wking: nit: "undertsand" -> "understand".

> how do I link to a section in the same adoc file?

[Looks like][1] (not sure how stable that anchor will be) `<<${ANCHOR},${TEXT}>>`, so with a BlockId on the header:

```asciidoc
[[candidate-channel]]
= candidate-{product-version}
```

and then consuming text here like:

```asciidoc
Please refer to <<candidate-channel,the candidate-{product-version} section>> to understand...
```

It's also possible that you can just use `<<candidate-channel>>` and get appropriate text filled in automatically.

Although I see stuff like `[id="understanding-upgrade-channels_{context}"]`, and don't know why we'd be using that instead of `[[...]]` BlockIds (they should [be equivalent][2]).

[1]: http://asciidoc.org/userguide.html#X41
[2]: http://asciidoc.org/userguide.html#_block_identifier

modules/understanding-upgrade-channels.adoc
jottofar: @kalexand-rh do we have a [serial comma](https://en.wikipedia.org/wiki/Serial_comma) docs policy?

modules/understanding-upgrade-channels.adoc
jottofar: This one?
"Like releases in the stable-{product-version} channel, update
recommendations in the stable-{product-version} channel will be promoted from
fast-{product-version} with a delay that can range from several hours to a day."

modules/understanding-upgrade-channels.adoc
jottofar: I'm leaving this alone for now. There is a lot of discretion in deciding whether to support a given customer in a given situation at a given time so "may" is not an altogether bad chioce.

modules/understanding-upgrade-channels.adoc
wking: Fixed in c3ce8cd.

modules/understanding-upgrade-channels.adoc
wking: Fixed in c3ce8cd.

modules/understanding-upgrade-channels.adoc
wking: Fixed in c3ce8cd.

modules/understanding-upgrade-channels.adoc
wking: Fixed in c3ce8cd.

modules/understanding-upgrade-channels.adoc
wking: > This one?

Yup.

jottofar:
@LalatenduMohanty @wking this is ready for review for incorporation of your latest comments. If you feel we're done for this round please provide lgtm.

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,89 +7,80 @@
...
-{product-title} {product-version}, which includes the upgrade from the previous 4.3
-release, has three upgrade channels to choose from:
+= {product-title} Upgrade Channels and Releases
+
+In {product-title} 4.1, Red Hat introduced the concept of channels for
+recommending the appropriate release versions for cluster upgrade. By controlling
+the pace of upgrades, these upgrade channels allow users to choose an upgrade
+strategy. Upgrade channels are tied to a minor version of
+{product-title}. For instance, {product-title} {product-version}
+upgrade channels will never include an upgrade to a 4.5 release. This ensures
```
LalatenduMohanty: The hard coding of 4.5 looks wrong here. This is will be wrong when this document gets published for 4.4 . Also we include previous version in channels e.g. https://github.com/openshift/cincinnati-graph-data/blob/master/channels/fast-4.3.yaml but not future minor versions (Y releases)

LalatenduMohanty (changes requested):
Will continue rest of the review tomorrow

modules/understanding-upgrade-channels.adoc
jottofar: This, as written, is for 4.4. So it's saying, for example, that stable-4.4 will never include an upgrade to 4.5. We checked with docs and, unfortunately, currently they could not come up with a way to implement variables like 'product-version-previous' or 'product-version-next'  so we're left with having to hard-code them.

LalatenduMohanty: Sorry I gave an wrong example. I meant 4.6.  I am guessing the need for implementing `product-version-next` has not came yet. @kalexand-rh Whats your suggestion on this?

modules/understanding-upgrade-channels.adoc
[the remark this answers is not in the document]
LalatenduMohanty: @jottofar we should modified the text as `The candidate-{product-version} channel will contain candidate builds for  all z-stream ({product-version}.z) releases`

LalatenduMohanty (changes requested):
The PR looks very good to me now except for a small nitpick. Once that is fixed I can mark it lgtm

modules/understanding-upgrade-channels.adoc
jottofar: We discussed in other PR, answer here: https://github.com/openshift/openshift-docs/pull/18944#discussion_r366816103

jottofar:
> The PR looks very good to me now except for a small nitpick. Once that is fixed I can mark it lgtm

Done

LalatenduMohanty (approved):
/lgtm

LalatenduMohanty:
> @jottofar is this ready for QE review?
> 
> @sferich888 ok from your POV?
> 
> @kalexand-rh FYI.

It has the `lgtm` , can we please ask the  QE to take a look.

wking:
weekly bump :innocent:

vikram-redhat:
Are we waiting for a QE ack on this?

@xltian can you request someone take a look.

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,89 +7,80 @@
...
+The candidate-{product-version} channel will contain candidate builds for all z-stream
+({product-version}.z) releases. If a given release is only available in the candidate-{product-version}
+channel and not available in either the fast-{product-version} or stable-{product-version}
+channels, it is considered a release candidate. Although release candidates contain
+all the features of the product, they are not supported and should only be used to
 test feature acceptance and assist in qualifying the next version of
-{product-title} in your infrastructure.
-+
+{product-title} in your infrastructure. Additionally, since a release candidate
+can only be upgraded to another release candidate, candidate-{product-version} may include
```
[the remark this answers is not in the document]
jottofar: /cc @LalatenduMohanty @wking This refers only to what would be available in our upgrade channels - candidate-4.x, fast-4.x, stable-4.x - and as far as I know this has not changed.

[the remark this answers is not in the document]
jiajliu: yes, i also mean in the upgrade channels. I catch 4.3.0-rc.3 to 4.3.0 upgrade path with following command.
```
curl -sH 'Accept:application/json' 'https://api.openshift.com/api/upgrades_info/v1/graph?channel=candidate-4.3' | /home/liujia/work/cincinnati/hack/graph.sh | dot -Tpng >graph.png
```
Also, i checked it in https://github.com/openshift/cincinnati-graph-data/blob/master/channels/candidate-4.3.yaml, am i mis-understanding "a release candidate can only be upgraded to another release candidate"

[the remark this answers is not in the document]
LalatenduMohanty: The statement `a release candidate can only be upgraded to another release candidate` is true. When builds gets releases in candidate channel, they are release candidate that point of time. However if we find it stable enough  it gets promoted to fast and stable channel which contains supported versions. I guess the current text confuses the user because of the previous sentence `If a given release is only available in the candidate-{product-version} channel and not available in either the fast-{product-version} or stable-{product-version}, it is considered a release candidate. `  Because every build when introduced in to candidate channel is a release candidate and then later we decide if we are going to support it by moving to fast and stable. let me know if that makes sense.

modules/understanding-upgrade-channels.adoc
jottofar: Removed sentence

modules/understanding-upgrade-channels.adoc
jottofar: No guidance provided so left as is.

modules/understanding-upgrade-channels.adoc
jottofar: done

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,17 +7,156 @@
...
-recommending the appropriate upgrade versions to your cluster. Upgrade channels
-separate upgrade strategies and also are used to control the cadence of updates.
-Channels are tied to a minor version of {product-title}. For instance,
-{product-title} {product-version} channels will never include an upgrade to a
-4.4 release. This ensures administrators make an explicit decision to upgrade to
-the next minor version of {product-title}. Channels only control updates and
-have no impact on the version of the cluster you install; the
+= {product-title} Upgrade Channels and Releases
+
+In {product-title} 4.1, Red Hat introduced the concept of Upgrade Channels for
```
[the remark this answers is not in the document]
jottofar: done

modules/understanding-upgrade-channels.adoc
jottofar: done

modules/understanding-upgrade-channels.adoc
jottofar: done

modules/understanding-upgrade-channels.adoc
jottofar: done

modules/understanding-upgrade-channels.adoc
jottofar: done

LalatenduMohanty (approved):
/lgtm

jottofar:
Can we please get this merged.

/cc @xltian  @jiajliu

jianlinliu:
jiajliu take sick leave, I will take over this review. 

LGTM.

kalexand-rh:
I hope that jiajliu is better and back from sick leave soon!

I'm going to merge the change and open a follow-up PR to fix a few style issues.

PR #19090, opened 2020-01-16
iSCSI fix in TP tracker table

bobfuru:
This comment isn't quite correct: https://github.com/openshift/openshift-docs/issues/17796#issuecomment-572644371

iSCSI raw block was GA in 4.2, while iSCSI persistent storage was TP. Both are GA in 4.3. So this PR creates separate rows in the Tech Preview Tracker table of 4.3 release notes.


sheriff-rh (approved):
Looks great! Thanks, Bob!

PR #19112, opened 2020-01-16
OSDOCS-856 updating source links to 4.3

kalexand-rh:
From https://issues.redhat.com/browse/OSDOCS-856

I expect that these are the links that we'll need for 4.3.


kalexand-rh:
One of the links is live. One is not. @vikram-redhat, if these aren't both ready tomorrow, I'm going to split the PR.

PR #19113, opened 2020-01-17
bz-1780401 ETCD restore using upstream binary with incorrect version

vwalek:
originally opened here - https://github.com/openshift/openshift-docs/pull/18470

due mistake I deleted the branch before merge.

Please check again and comment.

### DIFF
--- a/day_two_guide/topics/proc_restoring-etcd.adoc
+++ b/day_two_guide/topics/proc_restoring-etcd.adoc
@@ -273,14 +273,18 @@ Before restoring etcd on a static pod:
 * `etcdctl` binaries must be available or, in containerized installations,
 the `rhel7/etcd` container must be available.
 +
-You can obtain etcd by running the following commands:
+You can install the `etcdctl` binary with the etcd package by running the following commands:
 +
 ----
-$ git clone https://github.com/coreos/etcd.git
-$ cd etcd
-$ ./build
+# yum install etcd
 ----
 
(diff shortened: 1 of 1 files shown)


day_two_guide/topics/proc_restoring-etcd.adoc
```diff
@@ -273,14 +273,18 @@ Before restoring etcd on a static pod:
...
+You can install the `etcdctl` binary with the etcd package by running the following commands:
 +
 ----
-$ git clone https://github.com/coreos/etcd.git
-$ cd etcd
-$ ./build
+# yum install etcd
 ----
 
+The package also installs the systemd service. Disable and mask the service as it shouldn't run as a systemd service when etcd runs in static pod.
```
vwalek: yeah, but we won't to make sure that the service is not started manually by mistake. it is not required, but if not masked it can be started. The disable is to make sure it will not start if the system is rebooted. It is more for safety.

kalexand-rh:
```suggestion
The package also installs the systemd service. Disable and mask the service so that it does not run as a systemd service when etcd runs in static pod. By disabling and masking the service, you ensure that you do not accidentally start it and prevent it from automatically restarting when you reboot the system.
```

@vwalek, will you make an update similar to this one? It's better to explain than to just say that you shouldn't do a thing.

vwalek: @kalexand-rh done :) Please review

kalexand-rh (commented):
LGTM!

kalexand-rh:
Thank you @vwalek! QE approved the original PR, so I'm going to merge.

PR #19118, opened 2020-01-17
CNV - BZ#1761501 - VM count added to telemetry

aburdenthehand:
Uncommenting telemetry VM count line now that it's been acked for 2.2
This is conditionalised for CNV use only. 

Cherrypick to enterprise-4.3


ousleyp:
LGTM. Merging after discussion with @alexxa

PR #19123, opened 2020-01-17
Updated Customer Admin definition

sheriff-rh:
@patrickstrick  PTAL
@vikram-redhat should this be cherrypicked beyond 3.11? Dedicated 3 is on 3.11 only. Thanks!


ahardin-rh:
LGTM!
@sheriff-rh Yes, I would apply it to 4.x docs too so that when Dedicated gets bumped up, all of the relevant content comes with it.

patrickstrick:
@sheriff-rh @ahardin-rh The Dedicated Portal and so also the Customer Admin role only applies to OSD v3. OSD 4 uses OpenShift Cluster Manager instead.

patrickstrick:
The text changes LGTM

PR #19145, opened 2020-01-17
bug 1791994 - FCP restriction on z in 4.2

kalexand-rh:
From https://bugzilla.redhat.com/show_bug.cgi?id=1791994

Per the bug, this change applies to only 4.2.


kalexand-rh:
@dbenoit17, will you PTAL?

mburke5678:
@kalexand-rh LGTM

PR #19147, opened 2020-01-17
WIP - CNV fencing (KNIP1160)

aburdenthehand:
Placeholder from 2.2 for the Fencing content for KNIP1160
One assembly and one new module


vikram-redhat:
@aburdenthehand - what is the status with this?

aburdenthehand:
This was a RHHI feature that involved CNV but was pushed from 2.1 to a subsequent release, and then again. I left it open for when we needed it but since it's been 6 months the feature is likely to have changed by now anyway. We may as well close it. I have the notes, in any case, if they're needed in the future.

PR #19148, opened 2020-01-17
Add dpdk and rdma examples

lamek:


lamek:
Examples of dpdk and rdma for the Multus team.
From here - https://github.com/zshi-redhat/openshift-docs/commit/baa394a92d9ad682603ed08dbbd70aa6b6725b4a

PR #19153, opened 2020-01-18
bug 1777206 Fix the process how to add interfaces

lamek:
Currently multus-cni only supports to add interface to pod at creation.
Once created pod, multus-cni does not modify interfaces.
This change fix the document to align it, i.e. use 'oc create -f' command,
not 'oc edit'.


ahardin-rh:
Looks great!

PR #19160, opened 2020-01-20
Remove cluster-api-cluster label on MHC

enxebre:
There's no reason to include this label on a MHC to filter machines, dropping it to avoid confusion.


kalexand-rh:
@jhou1, will you PTAL?

jhou1:
Looks good.

kalexand-rh:
Thank you both!

PR #19170, opened 2020-01-20
Hide daemon set section for OpenShift 3.11 ARO

nilsanderselde:
Hide the daemon set section in the OpenShift 3.11 ARO docs. It is not possible in OpenShift 3.11 ARO for a user with the `osa-customer-admin` role to patch a namespace to disable the default project-wide node selector as described on https://access.redhat.com/documentation/en-us/openshift_container_platform/3.11/html/developer_guide/dev-guide-daemonsets


vikram-redhat:
@lamek FYI

PR #19171, opened 2020-01-20
Fix some incorrect or unclear wording

nilsanderselde:
Clarify a few things in the docs.

- `oc adm policy remove-user` does not remove users; it removes roles from users
- `oc adm policy remove-group` does not remove groups; it removes roles from groups.
- The use of the word "retain" when talking about `--rolebinding-name=` is confusing. "Retain" implies that a name that already existed will be preserved and carried forward, but the name being assigned is new (being currently created by the user).


vikram-redhat:
xxia@redhat.com - can you confirm this please?

PR #19192, opened 2020-01-21
Adding warning about partially encrypted backup

bergerhoffer:
Preview: https://encryption-backup-warning--ocpdocs.netlify.com/openshift-enterprise/latest/authentication/encrypting-etcd.html#enabling-etcd-encryption_encrypting-etcd

### DIFF
--- a/modules/enabling-etcd-encryption.adoc
+++ b/modules/enabling-etcd-encryption.adoc
@@ -7,10 +7,10 @@
 
 You can enable etcd encryption to encrypt sensitive resources in your cluster.
 
-// [WARNING]
-// ====
-// Do not back up etcd until the encryption process is complete.
-// ====
+[WARNING]
+====
+It is not recommended to take a backup of etcd until the initial encryption process is complete. If the encryption process has not completed, the backup might be only partially encrypted.
+====
 
 .Prerequisites
 


modules/enabling-etcd-encryption.adoc
```diff
@@ -7,10 +7,10 @@
 
 You can enable etcd encryption to encrypt sensitive resources in your cluster.
 
-// [WARNING]
-// ====
-// Do not back up etcd until the encryption process is complete.
-// ====
+[WARNING]
+====
+It is not recommended to take a backup of etcd until the initial encryption process is complete, otherwise the backup might be only partially encrypted.
```
jboxman: Could you say: _If you backup the etcd database before the initial encryption process is complete, then the backup might be only partially encrypted._ Or similar?

jboxman (approved):
One thought, otherwise LGTM.

modules/enabling-etcd-encryption.adoc
bergerhoffer: @jboxman possibly, but I'd like to work in that it's not recommended to do so. Thoughts?

jboxman: @bergerhoffer, possibly:

> It is not recommended to take a backup of the etcd database until the initial encryption process is complete. If the encryption process is not completed, the backup might be only partially encrypted.

bergerhoffer: +1 thanks @jboxman. I do think that breaking it up into two sentences will be better for translation.

bergerhoffer:
@geliu2016 Can you take a look and make sure that the wording around this warning is accurate? Thanks!

geliu2016:
LGTM, thx

PR #19198, opened 2020-01-21
Adding Health Checks to Applications book

mburke5678:
Moved the container health checks topics from Nodes -> Containers to Applications per @thoraxe


mburke5678:
@thoraxe PTAL

PR #19231, opened 2020-01-22
Adding a warning for BZ1787745

lamek:
https://bugzilla.redhat.com/show_bug.cgi?id=1787745

### DIFF
--- a/metering/metering-about-installing-metering.adoc
+++ b/metering/metering-about-installing-metering.adoc
@@ -7,6 +7,13 @@ toc::[]
 
 Before installing metering into your cluster, review the following sections.
 
+[WARNING]
+====
+Installing Metering on an Openshift FIPS-enabled cluster with TLS enabled will fail while attempting to generate TLS certificates for the user.
+
+See link:https://bugzilla.redhat.com/show_bug.cgi?id=1787745[BZ#1787745] for details.
+====
+
 include::modules/metering-about-install.adoc[leveloffset=+1]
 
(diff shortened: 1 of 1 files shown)


lamek:
@timflannagan1 Can you take a quick look at the language here? I shorted the description in the JIRA story, but otherwise the same.

metering/metering-about-installing-metering.adoc
```diff
@@ -7,6 +7,13 @@ toc::[]
 
 Before installing metering into your cluster, review the following sections.
 
+[WARNING]
+====
+Installing Metering on an Openshift FIPS-enabled cluster fails during the execution of the `images/metering-ansible-operator/roles/meteringconfig/tasks/configure_root_ca.yml` task file.
```
timflannagan1: For context, we enable TLS/authentication between the Metering pod components by default. In the case where using this default behavior (i.e. not explicitly disabling TLS via `tls.enabled: false`), will execution will attempting to generate TLS certificates.

I think we may want to stray away from specifying the path to the task file as the BZ should have that information, and just point out that deploying Metering will fail on a FIPS-enabled cluster when TLS is enabled.

timflannagan1: I think the wording on this can be improved, but I was thinking of something along the lines of:

Installing Metering on an Openshift FIPS-enabled cluster with TLS enabled will fail while attempting to generate TLS certificates for the user.

jboxman:
@lamek, LGTM.

PR #19232, opened 2020-01-22
Added release notes for Serverless tech preview 1.4.0

abrennan89:


modules/serverless-rn-1-4-0.adoc
```diff
@@ -0,0 +1,19 @@
...
+//
+// * serverless/release-notes.adoc
+
+[id="serverless-rn-1-4-0_{context}"]
+= Release Notes for Red Hat {ServerlessProductName} Technology Preview 1.4.0
+
+[id="new-features-1-4-0_{context}"]
+== New features
+* {ServerlessProductName} 1.4.0 is available on {product-title} 4.2 and newer versions.
+* {ServerlessProductName} has been updated to use Knative Serving 0.11.0.
```
markusthoemmes:
```suggestion
* {ServerlessProductName} has been updated to use Knative Serving 0.11.1.
```

modules/serverless-rn-1-4-0.adoc
```diff
@@ -0,0 +1,19 @@
...
+* {ServerlessProductName} has been updated to use Knative Client (`kn` CLI) 0.11.0.
+* Knative Client is now available for download through the *Command Line Tools* page in the {product-title} web console.
+
+[id="fixed-issues-1-4-0_{context}"]
+== Fixed issues
```
markusthoemmes: We could add here:

```
- Fixed a bug where Knative Services could not be reached from outside of the installed ServiceMesh on their cluster-local address
```

https://issues.redhat.com/browse/SRVKS-378 for reference of that bug.

modules/serverless-rn-1-4-0.adoc
```diff
@@ -0,0 +1,19 @@
...
+
```
markusthoemmes: Another one to add:

```
- Fixed a bug where removing configuration entries from the KnativeServing resource did not properly restore the defaults.
```

https://issues.redhat.com/browse/SRVKS-241 for reference.

modules/serverless-rn-1-4-0.adoc
```diff
@@ -0,0 +1,19 @@
...
+[id="serverless-rn-1-4-0_{context}"]
+= Release Notes for Red Hat {ServerlessProductName} Technology Preview 1.4.0
+
+[id="new-features-1-4-0_{context}"]
+== New features
+* {ServerlessProductName} 1.4.0 is available on {product-title} 4.2 and newer versions.
+* {ServerlessProductName} has been updated to use Knative Serving 0.11.0.
+* {ServerlessProductName} has been updated to use Knative Client (`kn` CLI) 0.11.0.
+* Knative Client is now available for download through the *Command Line Tools* page in the {product-title} web console.
+
```
markusthoemmes: Let's please add a section that states that we've migrated our Knative Serving CR's APIgroup from `serving.knative.dev` to `operator.knative.dev`.

We can write it like this:

```
Note:
The KnativeServing object's api group has changed in this release from "serving.knative.dev" to 
"operator.knative.dev". This migration is done automatically when installing the operator. Please 
adjust all your scripts to reflect the new api group as shown in the "Installing Knative Serving" section.
```

I'll do a review of the existing docs and send a PR that moves existing occurrences over.

markusthoemmes (commented):
Added a few things, sorry for being a little late to get to this.

modules/serverless-rn-1-4-0.adoc
abrennan89: See Slack thread about this one. I'm not sure we need to add it.

modules/serverless-rn-1-4-0.adoc
```diff
@@ -0,0 +1,27 @@
...
+* {ServerlessProductName} 1.4.0 is available on {product-title} 4.2 and newer versions.
+* {ServerlessProductName} has been updated to use Knative Serving 0.11.1.
+* {ServerlessProductName} has been updated to use Knative Client (`kn` CLI) 0.11.0.
+* {ServerlessProductName} has been updated to use Knative Serving Operator 0.11.1.
+* Knative Client is now available for download through the *Command Line Tools* page in the {product-title} web console.
+* The `KnativeServing` object's API group has changed in this release from `serving.knative.dev` to `operator.knative.dev`. This migration is completed automatically when {ServerlessOperatorName} 1.4.0 is installed.
++
+[IMPORTANT]
+====
+You will need to adjust all of your scripts to reflect the new API group. The {ServerlessProductName} installation instructions have been updated to provide details on how to complete this adjustment.
```
markusthoemmes:
```suggestion
You will need to adjust all of your scripts to reflect the new API group. The {ServerlessProductName} installation instructions have been updated to the new API group as well.

If you're upgrading from {ServerlessProductName} 1.3.0 the operator will create a `KnativeServing` CR in the new API group for you. This CR will be a mirror of the `KnativeServing` CR in the old group, that was used in 1.3.0.
If you need to stick to the old group, you can use the old CR as before. Note though that it's deprecated and it will eventually be removed.
To upgrade to the new group, make sure all scripts and applications that rely on the old API group are updated to the new group. Once that is done, the old CR can be removed and the new one used for management.

Note: While the old CR is still there, changes to the new CR will be reconciled away by the operator. While the old CR is still active, all changes need to be made to it. Once you removed it, you can use the new CR as you used the old one in the past.
```

I'm sure this sounds awful, but I hope it gets across what we need to document here and what the users need to know 😂 

Essentially the migration path now looks like this:

1. Install the new version of the operators.
2.a. An installed system keeps working as before. You will have both an old and a new resource.
2.b. You can install the system freshly with the old resource and it still works as before. You will have both an old and a new resource.
2.c. You can install the system with the new CR. You will have only that new resource.
3. If you had both old and a new resource, you now have time to adjust your scripts etc. Once you've done that, you can remove the old resource without repercussions and continue to use just the new resource.

Including the note on that changes to the new CR will be undone while the old CR has not been deleted.

abrennan89: Questions:
- why would someone need to stick to the old CR? It seems redundant to say this rather than just have everyone upgrade properly unless there's a reason.

I've made some updates based on not including mention of using the old CR for now. If there's a good use case for adding this lmk!

markusthoemmes: I wanted to signify that the old CR still works and one essentially can use the lifespan of this entire version to do the switchover. That allows for graceful upgrades for consuming systems that depend on the old CR. You can‘t upgrade both versions at once without a conflict so you need to upgrade Serverless first and then until you upgrade the consuming system rely on the old CR to fully work.

modules/serverless-rn-1-4-0.adoc
```diff
@@ -0,0 +1,33 @@
...
+* {ServerlessProductName} has been updated to use Knative Serving Operator 0.11.1.
+* Knative Client is now available for download through the *Command Line Tools* page in the {product-title} web console.
+* The `KnativeServing` object's API group has changed in this release from `serving.knative.dev` to `operator.knative.dev`. You will need to adjust any of your scripts or applications that rely on the old API group to use the new API group. The {ServerlessProductName} installation instructions have been updated to use the new API group.
++
+When upgrading from {ServerlessProductName} 1.3.0 to 1.4.0, the {ServerlessOperatorName} will create a `KnativeServing` custom resource (CR) in the new API group for you. This CR will be a mirror of the `KnativeServing` CR in the old group that was used in {ServerlessProductName} 1.3.0.
++
+If you need to keep using the old group temporarily, you can use the old CR as before. However, this CR is deprecated and will eventually be removed.
++
+Once you have updated references to the new API group, you can remove any older CR versions and use the newly deployed `KnativeServing` CR instead.
```
markusthoemmes: We'll need to tweak this a bit as the behavior has changed:

> Once you have updated references to the new API group, you'll want to remove the old CR version. To safely do this without downtime, remove the owner reference from the newly deployed `KnativeServing` CR using `oc edit knativeserving.operator.knative.dev knative-serving -n knative-serving`. After the owner reference has been removed, you can safely remove any older CR versions and start using the new one.

modules/serverless-rn-1-4-0.adoc
```diff
@@ -0,0 +1,33 @@
...
+====
+If a previous version of the CR exists, changes to the new CR will be overwritten by the {ServerlessOperatorName}. While the old CR is still active, all changes need to be made to that CR.
+====
+
+[id="fixed-issues-1-4-0_{context}"]
+== Fixed issues
+* Connecting to a private, cluster local Knative Service from a namespace that was not part of the `knative-serving-ingress` Service Mesh was failing on `i/o timeout`. This issue is now fixed.
+
+[id="known-issues-1-4-0_{context}"]
+== Known issues
```
markusthoemmes: Let's add

- Unqualified usages of `knativeserving` in `oc` commands (like `oc get knativeserving -n knative-serving`) may no longer work because of the aforementioned migration to a new API group. Use the explicit fully-qualified format (like `oc get knativeserving.operator.knative.dev -n knative-serving`) instead.

https://issues.redhat.com/browse/SRVKS-414 for reference.

modules/serverless-rn-1-4-0.adoc
```diff
@@ -0,0 +1,55 @@
...
+* {ServerlessProductName} has been updated to use Knative Serving Operator 0.11.1.
+* Knative Client is now available for download through the *Command Line Tools* page in the {product-title} web console.
+* The `KnativeServing` object's API group has changed in this release from `serving.knative.dev` to `operator.knative.dev`. You will need to adjust any of your scripts or applications that rely on the old API group to use the new API group. The {ServerlessProductName} installation instructions have been updated to use the new API group.
++
+When upgrading from {ServerlessProductName} 1.3.0 to 1.4.0, the {ServerlessOperatorName} will create a `KnativeServing` custom resource (CR) in the new API group for you. This CR will be a mirror of the `KnativeServing` CR in the old group that was used in {ServerlessProductName} 1.3.0.
++
+If you need to keep using the old group temporarily, you can use the old CR as before. However, this CR is deprecated and will eventually be removed.
++
+Once you have updated references to the new API group, you can remove any older CR versions and use the newly deployed `KnativeServing` CR instead.
```
mgencur: If the user removes the old CR it will uninstall the whole Serverless. IMO, we need to put the text from the note below directly to this paragraph so that nobody can miss that.

mgencur (changes requested):
Looks good. Just one comment regarding the old vs new API

modules/serverless-rn-1-4-0.adoc
```diff
@@ -0,0 +1,56 @@
...
+[IMPORTANT]
+====
+If a previous version of the CR exists, changes to the new CR will be overwritten by the {ServerlessOperatorName}. While the old CR is still active, all changes need to be made to that CR.
+====
+
+[id="fixed-issues-1-4-0_{context}"]
+== Fixed issues
+* Connecting to a private, cluster local Knative Service from a namespace that was not part of the `knative-serving-ingress` Service Mesh was failing on `i/o timeout`. This issue is now fixed.
+
+* The `container_name` and `pod_name` metric labels were removed in {product-tile} 4.3. The documentation has been updated to use the new `container` and `pod` metric labels instead. If you are using metering with Serverless on {product-tile} 4.3 or later, you must update your Prometheus queries according to the current version of the Serverless metering documentation.
```
markusthoemmes:
```suggestion
* The `container_name` and `pod_name` metric labels were removed in {product-title} 4.3. The documentation has been updated to use the new `container` and `pod` metric labels instead. If you are using metering with Serverless on {product-title} 4.3 or later, you must update your Prometheus queries according to the current version of the Serverless metering documentation.
```

This doesn't render properly.

modules/serverless-rn-1-4-0.adoc
```diff
@@ -0,0 +1,56 @@
...
+oc get knativeserving -n knative-serving
+----
++
+Use the explicit fully-qualified format instead. For example:
++
+----
+oc get knativeserving.operator.knative.dev -n knative-serving
+----
+
+* The first Knative Service created in a namespace can take over 10 seconds to become ready, waiting in `IngressNotConfigured` state. This is due to the time taken for the namespace to be added to the `ServiceMeshMemberRoll`.
```
markusthoemmes: This has been known in 1.3.0 already. We even documented it in "Deploying a serverless application"

> The first time that a Knative service is created in a namespace, that  namespace will automatically receive a new networking configuration.  This might cause the initial service to take longer than is usually  required for a service to become ready.

modules/serverless-rn-1-4-0.adoc
abrennan89: Oops, thanks Markus

alanfx:
QE Aproved, @abrennan89

modules/serverless-rn-1-4-0.adoc
```diff
@@ -0,0 +1,54 @@
...
+[id="serverless-rn-1-4-0_{context}"]
+= Release Notes for Red Hat {ServerlessProductName} Technology Preview 1.4.0
+
+[id="new-features-1-4-0_{context}"]
+== New features
+* {ServerlessProductName} 1.4.0 is available on {product-title} 4.2 and newer versions.
+* {ServerlessProductName} has been updated to use Knative Serving 0.11.1.
+* {ServerlessProductName} has been updated to use Knative Client (`kn` CLI) 0.11.0.
+* {ServerlessProductName} has been updated to use Knative Serving Operator 0.11.1.
+* Knative Client is now available for download through the *Command Line Tools* page in the {product-title} web console.
```
bmcelvee: Just a thought (not super important at all), the client abbreviation can be used or included here since it was introduced in the previous bullet.

modules/serverless-rn-1-4-0.adoc
```diff
@@ -0,0 +1,54 @@
...
+* The `KnativeServing` object's API group has changed in this release from `serving.knative.dev` to `operator.knative.dev`. You will need to adjust any of your scripts or applications that rely on the old API group to use the new API group. The {ServerlessProductName} installation instructions have been updated to use the new API group.
++
+When upgrading from {ServerlessProductName} 1.3.0 to 1.4.0, the {ServerlessOperatorName} will create a `KnativeServing` custom resource (CR) in the new API group for you. This CR will be a mirror of the `KnativeServing` CR in the old group that was used in {ServerlessProductName} 1.3.0.
++
+If you need to keep using the old group temporarily, you can use the old CR as before. However, this CR is deprecated and will eventually be removed.
++
+Once you have updated references to the new API group, you can remove any older CR versions and use the newly deployed `KnativeServing` CR instead. To safely do this without downtime, remove the owner reference from the newly deployed `KnativeServing` CR using:
++
+----
+ oc edit knativeserving.operator.knative.dev knative-serving -n knative-serving
```
bmcelvee: s/ oc edit.../$ oc edit...

If acceptable, this would apply to the other code examples in this section

modules/serverless-rn-1-4-0.adoc
```diff
@@ -0,0 +1,54 @@
...
+oc get knativeserving -n knative-serving
+----
++
+Use the explicit fully-qualified format instead. For example:
++
+----
+oc get knativeserving.operator.knative.dev -n knative-serving
+----
+
+* {product-title} scale from zero latency causes a delay of around 10 seconds when creating pods. This is a current {product-title} limitation.
```
bmcelvee: s/around/approximately

bmcelvee:
Just a few suggestions, otherwise LGTM! :)

abrennan89:
branch/enterprise-4.2
branch/enterprise-4.3

abrennan89:
Change applies for OCP 4.2+

vikram-redhat:
Release notes should go directly in the branch, not to master.

vikram-redhat:
@abrennan89 not CPing to 4.2 just yet. Let me know if it should be in there as well. I will need to work with you to resolve the conflicts.

PR #19237, opened 2020-01-22
Remove duplicated etcd server info

codyhoag:
@kalexand-rh can you confirm? It seems this info was duplicated outside of the table, so I removed it.


kalexand-rh:
@codyhoag, you're right. LGTM

PR #19239, opened 2020-01-22
adding vpc and private clusters to the install type chart

kalexand-rh:


ahardin-rh:
LGTM!

PR #18917, opened 2020-01-09
modules/installation-aws-user-infra-rhcos-ami: Bump to RHCOS 43.81.201912131630.0

wking:
Following up on #18396 and catching up with openshift/installer@3544350cc (openshift/installer#2777).

Generated with:

```console
$ date --iso=m --utc
2020-01-09T00:57+0000
$ git fetch origin
$ git --no-pager log -1 --oneline origin/release-4.3 -- data/data/rhcos.json
3544350cc (origin/pr/2777) BZ 1775388: rhcos: Bump to 43.81.201912131630.0
$ git cat-file -p 3544350cc:data/data/rhcos.json | jq -r '.amis | to_entries | sort_by(.key)[] | "\n|`" + .key + "`\n|`" + .value.hvm + "`"'
```

and pasting the output into the module doc.

PR #19244, opened 2020-01-22
updating RHCOS AMIs

kalexand-rh:
Trevor updated the AMIs on https://github.com/openshift/openshift-docs/pull/18917, but it looks like they've shifted again.

Updates per current status of https://github.com/openshift/installer/blob/release-4.3/data/data/rhcos.json

@jianlinliu, will you PTAL?


jianlinliu:
LGTM.

PR #19246, opened 2020-01-22
OSDOCS-878: Adding docs for auto rotation of service CA

bergerhoffer:
Preview: https://osdocs-878-auto-rotate--ocpdocs.netlify.com/openshift-enterprise/latest/authentication/certificates/service-serving-certificate.html#understanding-service-serving_service-serving-certificate

### DIFF
--- a/modules/customize-certificates-manually-rotate-service-ca.adoc
+++ b/modules/customize-certificates-manually-rotate-service-ca.adoc
@@ -5,8 +5,14 @@
 [id="manually-rotate-service-ca_{context}"]
 = Manually rotate the service CA certificate
 
-The service CA is valid for one year after {product-title} is
-installed. Follow these steps to manually refresh the service CA before the expiration date.
+The service CA is valid for 26 months and is automatically refreshed when there is less than six months validity left.
+
+If necessary, you can manually refresh the service CA by using the following procedure.
+
+[WARNING]
+====
+A manually-rotated service CA does not maintain trust with the previous service CA. You might experience a temporary service disruption until the Pods in the cluster are restarted, which ensures that Pods are using service serving certificates issued by the new service CA.
+====
 
 .Prerequisites
 
--- a/modules/customize-certificates-understanding-service-serving.adoc
+++ b/modules/customize-certificates-understanding-service-serving.adoc
@@ -15,5 +15,18 @@ algorithm to generate service certificates.
 The generated certificate and key are in PEM format, stored in `tls.crt`
 and `tls.key` respectively, within a created secret. The
 certificate and key are automatically replaced when they get close to
-expiration. The service CA certificate, which signs the service
-certificates, is only valid for one year after {product-title} is installed.
+expiration.
+
+The service CA certificate, which issues the service certificates, is valid for 26 months and is automatically rotated when there is less than six months validity left. After rotation, the previous service CA configuration is still trusted until its expiration. This allows a grace period for all affected services to refresh their key material before the expiration. If you do not upgrade your cluster during this grace period, which restarts services and refreshes their key material, you might need to manually restart services to avoid failures after the previous service CA expires.
+
+[NOTE]
+====
+You can use the following command to manually restart all Pods in the cluster. Be aware that running this command causes a service interruption, because it deletes every running Pod in every namespace. These Pods will automatically restart after they are deleted.
+
+----
+$ for I in $(oc get ns -o jsonpath='{range .items[*]} {.metadata.name}{"\n"} {end}'); \
+      do oc delete pods --all -n $I; \
+      sleep 1; \
+      done
+----
+====
--- a/modules/update-upgrading-cli.adoc
+++ b/modules/update-upgrading-cli.adoc
@@ -145,3 +145,19 @@ $ oc get clusterversion
 NAME      VERSION     AVAILABLE   PROGRESSING   SINCE     STATUS
 version   4.1.2       True        False         2m        Cluster version is 4.1.2
 ----
+
+. If you are upgrading to this release from {product-title} 4.3.3 or earlier, you must restart all Pods after the upgrade is complete. You can do this using the following command:
++
+----
+$ for I in $(oc get ns -o jsonpath='{range .items[*]} {.metadata.name}{"\n"} {end}'); \
+      do oc delete pods --all -n $I; \
+      sleep 1; \
+      done
+----
++
+[NOTE]
+====
+Restarting all Pods is required because the service CA is automatically rotated as of {product-title} 4.3.5. The service CA is rotated during the upgrade and a restart is required afterward to ensure that all services are using the new service CA before the previous service CA expires.
+
+After this one-time manual restart, subsequent upgrades and rotations will ensure restart before the service CA expires without requiring manual intervention.
+====
--- a/modules/update-upgrading-web.adoc
+++ b/modules/update-upgrading-web.adoc
@@ -63,6 +63,38 @@ and click *Update*.
 The *UPDATE STATUS* changes to `Updating`, and you can review the progress of
 the Operator upgrades on the *Cluster Operators* tab.
 
+. If you are upgrading to this release from {product-title}
+ifdef::within[]
+4.3.3 or earlier,
+endif::within[]
+ifdef::between[]
+// 4.2.16 or earlier,
+4.2,
+endif::between[]
+you must restart all Pods after the upgrade is complete. You can do this using the following command, which requires the OpenShift CLI (`oc`):
++
+----
+$ for I in $(oc get ns -o jsonpath='{range .items[*]} {.metadata.name}{"\n"} {end}'); \
+      do oc delete pods --all -n $I; \
+      sleep 1; \
+      done
+----
++
+[NOTE]
+====
+Restarting all Pods is required because the service CA is automatically rotated as of {product-title}
+ifdef::within[]
+4.3.5.
+endif::within[]
+ifdef::between[]
+// 4.2.18 and 4.3.5.
+4.3.5.
+endif::between[]
+The service CA is rotated during the upgrade and a restart is required afterward to ensure that all services are using the new service CA before the previous service CA expires.
+
+After this one-time manual restart, subsequent upgrades and rotations will ensure restart before the service CA expires without requiring manual intervention.
+====
+
 ifdef::between[]
 . After the update completes and the Cluster Version Operator refreshes the available updates, check if more updates are available in your current channel.
 +
--- a/updating/updating-cluster-between-minor.adoc
+++ b/updating/updating-cluster-between-minor.adoc
@@ -18,6 +18,21 @@ Because of the difficulty of changing update channels by using `oc`, use the web
 See xref:../authentication/using-rbac.adoc[Using RBAC to define and apply permissions].
 * Have a recent xref:../backup_and_restore/backing-up-etcd.adoc#backup-etcd[etcd backup] in case your upgrade fails and you must xref:../backup_and_restore/disaster_recovery/scenario-2-restoring-cluster-state.adoc#dr-restoring-cluster-state[restore your cluster to a previous state].
 
+// The following note will need to stay until (and be updated for) the 4.4 docs
+
+// Assuming next z-streams:
+// from {product-title} 4.1.31 or earlier; as of {product-title} 4.1.34
+// from {product-title} 4.2.16 or earlier; as of {product-title} 4.2.18 and 4.3.5
+
+[IMPORTANT]
+====
+If you are upgrading to this release from {product-title} 4.2, you must restart all Pods after the upgrade is complete.
+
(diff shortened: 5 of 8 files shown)


marun:
Need to update validity duration of the CA from one year to 14 months, but otherwise lgtm.

bergerhoffer:
This PR must be held until the auto rotation feature is released in the 4.3.1 z-stream

modules/customize-certificates-manually-rotate-service-ca.adoc
```diff
@@ -5,8 +5,14 @@
 [id="manually-rotate-service-ca_{context}"]
 = Manually rotate the service CA certificate
 
-The service CA is valid for one year after {product-title} is
-installed. Follow these steps to manually refresh the service CA before the expiration date.
+The service CA is valid for 14 months and is automatically refreshed when there is less than six months validity left.
+
+If necessary, you can manually refresh the service CA using the following procedure.
```
kalexand-rh: s/using/by using

modules/customize-certificates-understanding-service-serving.adoc
```diff
@@ -15,5 +15,6 @@ algorithm to generate service certificates.
 The generated certificate and key are in PEM format, stored in `tls.crt`
 and `tls.key` respectively, within a created secret. The
 certificate and key are automatically replaced when they get close to
-expiration. The service CA certificate, which signs the service
-certificates, is only valid for one year after {product-title} is installed.
+expiration.
+
+The service CA certificate, which signs the service certificates, is valid for 14 months and is automatically rotated when there is less than six months validity left. After rotation, the previous service CA configuration is still trusted until its expiration. This allows a grace period for all affected services to refresh their key material before the expiration. If you do not upgrade your cluster during this grace period, you might need to manually restart services to avoid failures once the previous service CA expires.
```
kalexand-rh: s/once/after

modules/customize-certificates-manually-rotate-service-ca.adoc
```diff
@@ -5,8 +5,14 @@
...
+
+[WARNING]
+====
+Manual rotation does not maintain trust with the previous service CA. This might result in service disruption until all affected services are refreshed to use the new service CA.
```
kalexand-rh: I think it might be clearer if the focus was more on the trust between the CAs instead of the rotation process. I might say something like:
> A manually-rotated service CA does not maintain trust with the previous service CA. You might experience a temporary service disruption until the Pods in the server restart, which applies the new service CA to the services.

kalexand-rh (commented):
I have a few picks, but it LGTM.

modules/customize-certificates-manually-rotate-service-ca.adoc
marun: Note that the service CA 'issues' service serving certificates. The last sentence would therefore more correctly be `until the pods in the cluster are restarted, which ensures that pods are using service serving certificates issued by the new CA`.

modules/customize-certificates-understanding-service-serving.adoc
```diff
@@ -15,5 +15,6 @@ algorithm to generate service certificates.
 The generated certificate and key are in PEM format, stored in `tls.crt`
 and `tls.key` respectively, within a created secret. The
 certificate and key are automatically replaced when they get close to
-expiration. The service CA certificate, which signs the service
-certificates, is only valid for one year after {product-title} is installed.
+expiration.
+
+The service CA certificate, which signs the service certificates, is valid for 14 months and is automatically rotated when there is less than six months validity left. After rotation, the previous service CA configuration is still trusted until its expiration. This allows a grace period for all affected services to refresh their key material before the expiration. If you do not upgrade your cluster during this grace period, you might need to manually restart services to avoid failures after the previous service CA expires.
```
kalexand-rh: Is "key material" the best phrase?

kalexand-rh (commented):
Consider adding the step to restart the Pods and a reminder about the necessary versions after the upgrade steps. I think that the note makes sense up front, but it would be a kindness to remind them when they actually need to do it.

modules/customize-certificates-understanding-service-serving.adoc
marun: key material == certificates issued by the CA and the CA bundle used to verify those certificates

marun: signs -> issues

marun: maybe mention that upgrading restarts services and thus refreshes their key material?

Also, not sure what editor you're using but it would be easier to review if this block was broken up into separate lines.

bergerhoffer: Thanks, will update.

And we removed hard wrapping from our guidelines a bit ago. The minor inconvenience during reviews was outweighed by by other benefits to not hard wrapping our doc text.

bergerhoffer:
@barleyer Can you please review these updates? I know we're still waiting on final sign-off of the feature from QA, but I'm on PTO this Friday and want to be ready if this is going out in 4.3.1 on Monday.

Right now, I'm assuming that this is making it for the next z-streams: 4.3.1, 4.2.18, and 4.1.34. But will need confirmation on these once this is decided.

Also, this doc update will be applied back to 4.1 and 4.2, and I'll adjust the versions mentioned in the [IMPORTANT] notes accordingly depending on the version.

bergerhoffer:
@sferich888 Do you think we should add a step to the update procedures (for example here [1]) for manually restarting pods? This would only need to be done once, when upgrading for the first time to a release that supports auto rotation, like for 4.3, only if you're upgrading from 4.3.0 to 4.3.1 or later. Details are in this important note here at the top of the page before you get started [2]. Let me know what you think, thanks!

[1] https://osdocs-878-auto-rotate--ocpdocs.netlify.com/openshift-enterprise/latest/updating/updating-cluster.html#update-upgrading-web_updating-cluster
[2] https://osdocs-878-auto-rotate--ocpdocs.netlify.com/openshift-enterprise/latest/updating/updating-cluster.html

modules/customize-certificates-understanding-service-serving.adoc
```diff
@@ -15,5 +15,6 @@ algorithm to generate service certificates.
 The generated certificate and key are in PEM format, stored in `tls.crt`
 and `tls.key` respectively, within a created secret. The
 certificate and key are automatically replaced when they get close to
-expiration. The service CA certificate, which signs the service
-certificates, is only valid for one year after {product-title} is installed.
+expiration.
+
+The service CA certificate, which issues the service certificates, is valid for 26 months and is automatically rotated when there is less than six months validity left. After rotation, the previous service CA configuration is still trusted until its expiration. This allows a grace period for all affected services to refresh their key material before the expiration. If you do not upgrade your cluster during this grace period, which restarts services and refreshes their key material, you might need to manually restart services to avoid failures after the previous service CA expires.
```
sferich888: We need to link to https://docs.openshift.com/container-platform/4.3/authentication/certificates/service-serving-certificate.html#manually-rotate-service-ca_service-serving-certificate directly. 

Telling people the process they need to follow as part of this call out is critical.

updating/updating-cluster-between-minor.adoc
```diff
@@ -18,6 +18,21 @@ Because of the difficulty of changing update channels by using `oc`, use the web
...
+// The following note will need to stay until (and be updated for) the 4.4 docs
+
+// Assuming next z-streams:
+// from {product-title} 4.1.31 or earlier; as of {product-title} 4.1.34
+// from {product-title} 4.2.16 or earlier; as of {product-title} 4.2.18
+
+[IMPORTANT]
+====
+If you are upgrading to this release from {product-title} 4.2.16 or earlier, you must restart all Pods after the upgrade is complete.
+
```
barleyer: As Eric suggested,we better show the step of restarting pods,like:
```
$ for I in $(oc get ns -o jsonpath='{range .items[*]} {.metadata.name}{"\n"} {end}'); \
      do oc delete pods --all -n $I; \
      sleep 1; \
      done
```

barleyer:
One commemts, but otherwise lgtm.

modules/customize-certificates-understanding-service-serving.adoc
bergerhoffer: @sferich888 Unfortunately I can't link right now due to our current guidelines.

And @marun can confirm, but they shouldn't need to follow that whole procedure (since rotation is handled during the upgrade), but would just need to do step 3 to restart the pods. 

But I can add that step's info right here too.

updating/updating-cluster-between-minor.adoc
bergerhoffer: @sferich888 So like the other place, I can add this step right here in the callout, but do you think it would be better placed to actually make it a step at the end of the update procedure, (like here [1]). Or do you think that people will catch this note at the top with the step?

[1] https://osdocs-878-auto-rotate--ocpdocs.netlify.com/openshift-enterprise/latest/updating/updating-cluster-between-minor.html#update-upgrading-web_updating-cluster-between-minor

modules/customize-certificates-understanding-service-serving.adoc
```diff
@@ -15,5 +15,18 @@ algorithm to generate service certificates.
...
+
+[NOTE]
+====
+You can use the following command to manually restart all Pods in the cluster. Be aware that this command will cause a service interruption, as it goes through and deletes every running Pod in every namespace. These Pods will automatically restart after they are deleted.
```
kalexand-rh: I'd
s/this command will cause/running this command causes
s/as it goes through and deletes/because it deletes

modules/update-upgrading-cli.adoc
```diff
@@ -145,3 +145,19 @@ $ oc get clusterversion
...
+----
+$ for I in $(oc get ns -o jsonpath='{range .items[*]} {.metadata.name}{"\n"} {end}'); \
+      do oc delete pods --all -n $I; \
+      sleep 1; \
+      done
+----
++
+[NOTE]
+====
+This is required because the service CA is automatically rotated as of {product-title} 4.3.1. The service CA is rotated during the upgrade and a restart is required afterward to ensure that all services are using the new service CA before the previous service CA expires.
```
kalexand-rh: I'd s/This/Restarting all Pods

modules/update-upgrading-web.adoc
```diff
@@ -63,6 +63,36 @@ and click *Update*.
...
 the Operator upgrades on the *Cluster Operators* tab.
 
+. If you are upgrading to this release from {product-title}
+ifdef::within[]
+4.3.0,
+endif::within[]
+ifdef::between[]
+4.2.16 or earlier,
+endif::between[]
+you must restart all Pods after the upgrade is complete. You can do this using the following command:
```
kalexand-rh: Since this is in the web console section, I'd point out that you need `oc` to complete this step.

kalexand-rh (commented):
I think adding the loop will help!

bergerhoffer:
@sferich888 @barleyer I've added the step to restart pods to the update procedures. Can you please take a look?

For example:
https://osdocs-878-auto-rotate--ocpdocs.netlify.com/openshift-enterprise/latest/updating/updating-cluster-between-minor.html#update-upgrading-web_updating-cluster-between-minor

Though a question about this particular one @sferich888 - it's for updating using the web console, but we're throwing in an oc command to restart the pods. Is this okay, or is there an easy way to restart all pods from the console?

modules/customize-certificates-understanding-service-serving.adoc
sferich888: Ok this works for me; however, we should include the other module here vs duplicating content?

sferich888:
> Though a question about this particular one @sferich888 - it's for updating using the web console, but we're throwing in an oc command to restart the pods. Is this okay, or is there an easy way to restart all pods from the console?

I'm not aware of a way to restart everything from the console, and most people wouldn't be able too. You would need cluster admin permissions.

modules/update-upgrading-web.adoc
```diff
@@ -63,6 +63,36 @@ and click *Update*.
...
+4.2.16 or earlier,
+endif::between[]
+you must restart all Pods after the upgrade is complete. You can do this using the following command, which requires the OpenShift CLI (`oc`):
++
+----
+$ for I in $(oc get ns -o jsonpath='{range .items[*]} {.metadata.name}{"\n"} {end}'); \
+      do oc delete pods --all -n $I; \
+      sleep 1; \
+      done
+----
```
barleyer: Nice one.

bergerhoffer:
FYI @barleyer @marun Updated the versions in preparation for going out in 4.3.3 on Monday.

bergerhoffer:
FYI @marun @barleyer - 4.3.4 is being dropped so this is now scheduled for 4.3.5 as far as I'm aware. I've updated this PR accordingly.

bergerhoffer:
Merging as 4.3.5 just went out.

PR #19260, opened 2020-01-23
cnv-bz17931571 - remove IPAM from NAD examples

aburdenthehand:
The suggested fix in BZ#1793571 is to drop the 'ipam' line from the NAD examples.
Removed from two modules.

cherrypick to enterprise-4.3

### DIFF
--- a/modules/cnv-creating-bridge-nad-cli.adoc
+++ b/modules/cnv-creating-bridge-nad-cli.adoc
@@ -38,7 +38,6 @@ spec:
       {
         "type": "cnv-bridge", <2>
         "bridge": "br0", <3>
-        "ipam": {}
       },
       {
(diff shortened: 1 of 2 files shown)


modules/cnv-creating-bridge-nad-cli.adoc
```diff
@@ -38,7 +38,6 @@ spec:
       {
         "type": "cnv-bridge", <2>
         "bridge": "br0", <3>
```
phoracek: I would drop the comma at the end. Not sure if the current state is invalid, but without comma it would be consistent with the other JSONs if nothing else.

phoracek (approved):

ousleyp:
LGTM. Merging after discussing with @alexxa

PR #19282, opened 2020-01-24
Changing wording per Christian H in slack

mburke5678:
one note on the following:
Log forwarding does not provide secure storage for audit logs. You are responsible to ensure the endpoint is compliant with your organizational and governmental regulations and is properly secured. OpenShift Container Platform cluster logging does not comply with those regulations.
Can we replace “Log forwarding” with “OpenShift Logging”. We don’t provide any security mechanisms with the Elasticsearch solution shipped as a storage for OpenShift Logging and that’s why we would recommend using an appropriate storage as you said, and why we don’t support collecting audit logs by default but only through the Log Forwarding feature.

https://coreos.slack.com/archives/GGUR75P60/p1579860545100800

### DIFF
--- a/modules/cluster-logging-log-forwarding-about.adoc
+++ b/modules/cluster-logging-log-forwarding-about.adoc
@@ -27,7 +27,7 @@ A pipeline associates the type of data to an output. A type of data you can forw
 
 Note the following:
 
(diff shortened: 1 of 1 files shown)


mburke5678:
@sichvoge PTAL

PR #19289, opened 2020-01-24
BZ#1794683 Remove warning for required public CA signing of certs

codyhoag:
https://bugzilla.redhat.com/show_bug.cgi?id=1794683

### DIFF
--- a/modules/nw-ingress-setting-a-custom-default-certificate.adoc
+++ b/modules/nw-ingress-setting-a-custom-default-certificate.adoc
@@ -13,7 +13,7 @@ custom resource (CR).
 
 * You must have a certificate/key pair in PEM-encoded files, where the
 certificate is signed by a trusted certificate authority or by a private trusted
-certificate authority that you have configured.
+certificate authority that you configured in a custom PKI.
 
 * Your certificate is valid for the Ingress domain.
 
@@ -24,13 +24,6 @@ $ oc --namespace openshift-ingress-operator get ingresscontrollers
 NAME      AGE
 default   10m
 ----
-
-[WARNING]
-====
-If the default certificate is replaced, it *must* be signed by a public
-certificate authority already included in the CA bundle as provided by the
-container userspace.
-====
 
 [NOTE]
 ====


codyhoag:
@lihongan can you confirm? Thanks!

danehans:
Instead of completely removing the warning, it may be better to update it stating something like:

" If the cert is not signed by a CA in the RHCOS trust bundle, use `additionalTrustBundle` at install or `spec.trustedCA` from object proxy/cluster at runtime to add the CA cert."

__Note:__ See additional details [here](https://docs.openshift.com/container-platform/4.3/networking/enable-cluster-wide-proxy.html#nw-proxy-configure-object_config-cluster-wide-proxy) for configuring  `spec.trustedCA`.

codyhoag:
@danehans what do you think about making one of our prerequisites for this module more informative? A few lines above this warning we have a generic prereq statement:
> ...the certificate is signed by a trusted certificate authority or by a private trusted certificate authority you configured.

Instead, I propose we mention configuring the custom PKI for a private cert:
> ...the certificate is signed by a trusted certificate authority or by a private trusted certificate authority you configured in a custom PKI.

We have the [Configuring a Custom PKI](https://docs.openshift.com/container-platform/4.3/networking/configuring-a-custom-pki.html) docs linked in the assembly, so they can reference the info you suggested there. What do you think?

lihongan:
LGTM. Thanks.
Yes, the label `branch/enterprise-4.1` should be removed.

codyhoag:
I'm going to move forward with this PR since we've addressed the main concern of removing an inaccurate note. If we decide additional info would be helpful, I'll add it in a separate PR.

modules/nw-ingress-setting-a-custom-default-certificate.adoc
```diff
@@ -13,7 +13,7 @@ custom resource (CR).
 
 * You must have a certificate/key pair in PEM-encoded files, where the
 certificate is signed by a trusted certificate authority or by a private trusted
-certificate authority that you have configured.
+certificate authority you configured in a custom PKI.
```
mburke5678: @codyhoag I believe when we discussed writing for translation at one point we suggested using words such as `that` in place: 
`certificate authority that you configured in a custom PKI.`

mburke5678:
@codyhoag One nit. Otherwise LGTM

modules/nw-ingress-setting-a-custom-default-certificate.adoc
codyhoag: @mburke5678 ah, got it! That makes sense; I'll add it back.

danehans:
@codyhoag I see you cherry-picked to 4.2. [This](https://issues.redhat.com/browse/MON-884) issue must be marked done for the fix to be backported to 4.2. Should the docs backport wait until then?

codyhoag:
Thanks, @danehans I'm reverting this back in #19483.

PR #19291, opened 2020-01-24
CronJob: fix English of leading sentence

dankenigsberg:
I think that the existing syntax is incorrect, and try to explain CronJob by itself.


jboxman:
Seems valid.

- https://docs.openshift.com/container-platform/4.5/rest_api/workloads_apis/cronjob-batch-v1beta1.html
- https://docs.openshift.com/container-platform/4.5/rest_api/workloads_apis/job-batch-v1.html

PR #19304, opened 2020-01-24
fixing the ordering to align with other install guides

sferich888:
@rvanderp3 please review


rvanderp3 (approved):

kalexand-rh (commented):
I think this change looks good, but I think we need it in https://github.com/openshift/openshift-docs/blob/master/installing/installing_bare_metal/installing-bare-metal-network-customizations.adoc, too.

sferich888:
@kalexand-rh I pushed another update to get that other file too.

kalexand-rh:
Thank you for the update! Merging.

Issue #18036, opened 2019-11-11
[enterprise-4.2] Edit suggested in file registry/configuring-registry-storage/configuring-registry-storage-baremetal.adoc


jcordes73:
### Which section(s) is the issue in?
https://docs.openshift.com/container-platform/4.2/registry/configuring-registry-storage/configuring-registry-storage-baremetal.html#registry-configuring-storage-baremetal_configuring-registry-storage-baremetal

### What needs fixing?
NFS shouldn't be recommended or caveats should be outlined as in https://docs.openshift.com/container-platform/4.2/scalability_and_performance/optimizing-storage.html#specific-application-storage-recommendations.

vikram-redhat:
@codyhoag might just be a matter of copying over that warning to these docs.

codyhoag:
@jcordes73 Thanks for bringing this up. Since the suggestion for using NFS was simply removed for these docs in 4.3, I've removed the suggestion for 4.1/4.2. I've sent this for review:

#19305 
#19306

codyhoag:
The NFS suggestion has been replaced with OCS for OCP 4.2+:

#19306 
#19665

PR #19305, opened 2020-01-24
[enterprise-4.1] Replace NFS suggestion with OCS

codyhoag:
Suggested in #18036.

PR #19306, opened 2020-01-24
[enterprise-4.2] Replace NFS suggestion with OCS

codyhoag:
Suggested in #18036.

### DIFF
--- a/modules/registry-configuring-storage-baremetal.adoc
+++ b/modules/registry-configuring-storage-baremetal.adoc
@@ -15,8 +15,7 @@ registry to use storage.
 
 * Cluster administrator permissions.
 * A cluster on bare metal.
-* A provisioned persistent volume (PV) with `ReadWriteMany` access mode, such as
-`NFS`.
+* A provisioned persistent volume (PV) with `ReadWriteMany` access mode, such as Red Hat OpenShift Container Storage.
 * Must have "100Gi" capacity.
 
 .Procedure


codyhoag:
@duanwei33 can you confirm that we should avoid recommending NFS in this context?

I also sent this change for 4.1: #19305

codyhoag:
@duanwei33 can you review this small change when you have a moment. Let me know if I should reach out to someone else. Thanks!

duanwei33:
/LGTM
Sorry for the delay, I missed the last reminder. @codyhoag

bmcelvee:
LGTM, though I'm a little surprised this is being changed again. @dmage or @coreydaley  would you mind also approving? Thanks!

dmage:
If we don't like NFS, we should suggest another viable solution. What else can we recommend in 4.2 that works better than NFS?

codyhoag:
@dmage do you recommend I reach out to someone to ask what would be a good solution to replace NFS in this context?

codyhoag:
FYI, 4.3 docs have already removed the NFS suggestion in this context. If we'd like to add an alternate solution, I'll need to update the 4.3 doc too.

dmage:
In our documentation we [have only 3 options](https://docs.openshift.com/container-platform/4.3/storage/understanding-persistent-storage.html#pv-access-modes_understanding-persistent-storage): Azure File, NFS, and Red Hat OpenShift Container Storage. Azure is not suitable for baremetal clusters, NFS can be problematic. Can we recommend OCS?

codyhoag:
@duanwei33 can you take a second look at this? We've decided to recommend OCS *instead* of NFS. Do you agree?

I've made this change for 4.1 (#19305) and 4.3 (#19665) as well. Let me know if this doesn't make sense in those versions. Thanks

duanwei33:
/LGTM
I also let @liangxia to see if he has any comments.

PR #19312, opened 2020-01-24
Privileged pods not required for hostpath PV[C]s

aglitke:
When reviewing the documentation about hostpath volumes I noticed that
the section claims that Pods must be privileged in order to run with
PVCs that are bound to statically provisioned hostpath PVs.  This is not
correct.  Unprivileged pods can use these volumes as well.

I suspect there was some confusion between this usage model and the
deprecated inline volume specification for pods.  In that model the path
on the host could be specified in the Pod spec and this would obviously
create a security risk so this was restricted to privileged pods only.
In the case of a statically provisioned hostpath PV that is bound to a
PVC, only a cluster admin may create such a PV so there is no danger in
exposing a sensitive host path to an unprivileged user.

Signed-off-by: Adam Litke <alitke@redhat.com>

### DIFF
--- a/modules/persistent-storage-hostpath-pod.adoc
+++ b/modules/persistent-storage-hostpath-pod.adoc
@@ -3,7 +3,7 @@
 // * storage/persistent-storage/persistent-storage-hostpath.adoc
 
 [id="persistent-storage-hostpath-pod_{context}"]
-= Mounting the hostPath share in a privileged Pod
+= Mounting the hostPath share in a Pod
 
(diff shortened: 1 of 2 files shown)


bobfuru:
@huffmanca You were helping me test hostPath on 11/1/19. For some reason, you were only able to get hostPath storage to work with static provisioning when you made the pod run as privileged. So we added a note that it was required. 

Can you please help me in reviewing this PR when you get a chance? Thanks!

bobfuru (approved):
@huffmanca PTAL - Can we remove references to "privileged" for hostPath in this assembly?

PR #19318, opened 2020-01-25
removing and updating hard-coded versions for 2.2 upgrade docs

ousleyp:
- No BZ.
- For enterprise-4.3 and enterprise-4.4 only.
- Preview build: http://file.bos.redhat.com/pousley/012520/hard-coded-versions/cnv/cnv_install/upgrading-container-native-virtualization.html

@aburdenthehand Please review and merge if you can. Thanks!

### DIFF
--- a/modules/cnv-about-upgrading-cnv.adoc
+++ b/modules/cnv-about-upgrading-cnv.adoc
@@ -8,14 +8,18 @@
 If you enabled automatic updates when you installed {CNVProductName}, you
 receive updates as they become available.
 
-.Additional information
+=== How {CNVProductName} upgrades work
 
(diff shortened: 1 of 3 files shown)


modules/cnv-about-upgrading-cnv.adoc
```diff
@@ -10,8 +10,7 @@ receive updates as they become available.
 
 .Additional information
```
kalexand-rh: I'd probably change this heading to a statement that introduces the list. (And if there's no good way to do that, I'd break up the list in such a way that you could.)

kalexand-rh (commented):
One suggestion, otherwise it LGTM.

PR #19319, opened 2020-01-26
Added manual rollback procedure and updated known issues.

apinnick:
For 4.2 and 4.3. 
Manual rollback procedure for quiesced applications.


xinredhat (commented):
LGTM

migration/migrating-3-4/troubleshooting.adoc
```diff
@@ -8,7 +8,15 @@ toc::[]
 
 You can view the migration Custom Resources (CRs) and download logs to troubleshoot a failed migration.
 
+You can manually roll back a failed migration.
+
+[NOTE]
+====
+Manual rollback is only required if an application has been quiesced, to prevent data corruption. This procedure is not required if the application was not quiesced, because the original application is still running on the source cluster.
```
kalexand-rh: Can you s/quiesced/paused and not lose any meaning? If so, I'd do it in all places in the PR. (It looks like you're removing the only place it's currently used.)

I'd move "to prevent data corruption" to the start of the sentence.

modules/migration-manual-rollback.adoc
```diff
@@ -0,0 +1,90 @@
+// Module included in the following assemblies:
+// migration/migrating-3-4/troubleshooting.adoc
+// migration/migrating-4-4/troubleshooting.adoc
+// migration/migrating-4-4/troubleshooting.adoc
+[id='migration-manual-rollback_{context}']
+= Manual rollback
```
kalexand-rh: Procedures should start with gerunds, so I'd title it "Manually rolling back a migration"

modules/migration-manual-rollback.adoc
```diff
@@ -0,0 +1,90 @@
...
+. On the target cluster, switch to the migrated project:
++
+----
+$ oc project <project>
+----
+
+. Get the deployed resources:
++
+----
+$ oc get all
```
kalexand-rh: Does this really get all resources now? There was a 3.x issue last week saying that it didn't.

modules/migration-manual-rollback.adoc
```diff
@@ -0,0 +1,90 @@
...
+----
+
+. Delete each deployed resource type or scale it to `0`:
```
kalexand-rh: Please present only the best one. If there are different ones that users should do for different reasons, then present them as options in an unordered list with the reasons why you'd want to do one over the other.

modules/migration-manual-rollback.adoc
```diff
@@ -0,0 +1,90 @@
...
++
+----
+$ oc delete <resource_type>
+$ oc scale --replicas=0 <resource_type>/<resource_name>
+----
++
+You must ensure that no applications are running on the target cluster and accessing data on the PVC.
+
+. To stop a DaemonSet without deleting it, update the `nodeSelector` in the YAML file:
```
kalexand-rh: I think that this step might need to be a parallel option with the previous step or might need to be presented first. (I'm assuming that the DaemonSets are listed in the output of `oc get all`, and I don't want users to delete itesm that they might want to just stop.)

modules/migration-manual-rollback.adoc
```diff
@@ -0,0 +1,90 @@
...
+    metadata:
+      labels:
+        name: hello-daemonset
+    spec:
+      nodeSelector:
+        role: worker <1>
+----
+<1> Specify a `nodeSelector` value that does not exist on any node.
+
+. During migration, the reclaim policy for bound PVs is `Retain`, to ensure that data is not lost when an application is removed from the source cluster. You can update the PV's reclaim policy:
```
kalexand-rh: Can you move the action to the start of the step? Maybe something like, "Update each PV's reclaim policy so that the unnecessary data is removed. During migration, the reclaim policy for bound PVs is `Retain`, to ensure that data is not lost when an application is removed from the source cluster. You can remove these PVs during rollback."

modules/migration-manual-rollback.adoc
```diff
@@ -0,0 +1,90 @@
...
+<1> Specify `Recycle` or `Delete`.
+
+. On the source cluster, switch to the migrated project and get its deployed resources:
++
+----
+$ oc project <project>
+$ oc get all
+----
+
+. Start one or more replicas of the deployed resources:
```
kalexand-rh: Do you need to restart each deployed resource? If so, please s/of the deployed resources/of each deployed resource

kalexand-rh (commented):
I have a couple of suggestions, but this is looking good.

migration/migrating-3-4/troubleshooting.adoc
apinnick: I'm thinking that "stopped" might be a better word. I just did a search of the openshift docs and "stopped" seems to be the preferred word for pods (rather than pausing, halting, or quiescing). The UI itself isn't very helpful because the check box says "halt transactions". In a way, that's true, but the transactions between the pods and storage are halted by stopping the pods, so it would be accurate to say that the application is stopped. 

`I'd move "to prevent data corruption" to the start of the sentence.`  To the start of which sentence? The first sentence in the note or the sentence preceding the note? I think it might make sense to move it to the sentence preceding the note.

modules/migration-manual-rollback.adoc
apinnick: The command is being run on a 4.3 (or latest) cluster. 

Even if it doesn't actually get all the resources, that isn't critical. What's important is to kill anything that could be writing to the PV from the target cluster.

modules/migration-manual-rollback.adoc
apinnick: They're different options. I was trying to save space because it seemed silly to say "Delete resources if you want to delete them" and "Scale resources to 0 if you don't want to delete them." Will do.

migration/migrating-3-4/troubleshooting.adoc
kalexand-rh: Then please use "stopped" and consider filing a UI bug to make that check box text more helpful. :)

I think that moving it to the start of the note or the sentence before the note would work. The meaning's slightly different depending on where you put it, so please pick the one that is best.

modules/migration-manual-rollback.adoc
apinnick: On second thought, who cares? They can all be deleted. If a failed migration is being rolled back, it doesn't matter whether they're deleted or scaled to 0. :-)

modules/migration-manual-rollback.adoc
kalexand-rh: Thanks for the clarification!

modules/migration-manual-rollback.adoc
apinnick: No, that's not a problem. DaemonSets aren't so easy to kill. A DaemonSet would have to be deleted with `kubectl delete <daemonset>`. They won't delete it by accident.

modules/migration-manual-rollback.adoc
apinnick: Sounds good.

modules/migration-manual-rollback.adoc
apinnick: I would assume so. Will do.

apinnick:
@kalexand-rh  
I made the fixes and rewrote the intro sentences. Please review. Thanks!

migration/migrating-3-4/troubleshooting.adoc
apinnick: You're right. That check box has been bothering me for ages and I keep putting off creating that bug. Will do!

Issue #19339, opened 2020-01-27
Technology Preview Tracker invalid on OCP 4.3 release notes


brian-avery:
### Which section(s) is the issue in?
In the `Table 1. Technology Preview Tracker broken on OCP 4.3` table on https://docs.openshift.com/container-platform/4.3/release_notes/ocp-4-3-release-notes.html#ocp-4-3-technology-preview, the OCP 4.1 column lists several components as TP that were actually GA in 4.1. For example, `OpenShift Service Mesh` and `Operator Lifecycle Manager`

### What needs fixing?
The components that are GA in OCP 4.1 should be marked as GA.

codyhoag:
Thanks, @brian-avery! Processing this update now: #19341

codyhoag:
Merged. This update will be published soon.

PR #19341, opened 2020-01-27
Update 4.1 TP statuses in release notes

codyhoag:
Suggested in #19339. 

@sheriff-rh can you provide a sanity check? It appears that some of the 4.1 statuses mimicked 3.11 statuses that were there in our 4.2 RNs (likely from the copy/paste).

I did a diff and updated the features that I noticed were not correctly carried over.


sheriff-rh (approved):
Looks great, thanks for taking care of this @codyhoag !

PR #19345, opened 2020-01-28
Add static IP to 4.3 installs (OSDOCS-727)

chrisnegus:
This PR is meant to resolve [OSDOCS-727](https://issues.redhat.com/browse/OSDOCS-727). It describes how to do ISO installs in environments without access to DHCP, so IP address information needs to be set statically.  I was going to add a section on setting static IP addresses from a PXE server. But since that would require DHCP, that wouldn't fly.

### DIFF
--- a/modules/installation-user-infra-machines-iso.adoc
+++ b/modules/installation-user-infra-machines-iso.adoc
@@ -59,10 +59,19 @@ coreos.inst=yes
 coreos.inst.install_dev=sda <1>
 coreos.inst.image_url=<bare_metal_image_URL> <2>
 coreos.inst.ignition_url=http://example.com/config.ign <3>
+ip=<dhcp or static IP address> <4>
 ----
 <1> Specify the block device of the system to install to.
 <2> Specify the URL of the RAW image that you uploaded to your server.
 <3> Specify the URL of the Ignition config file for this machine type.
+<4> Set `ip=dhcp` or set an individual static IP address (ip=) and DNS server (nameserver=) on each node.
+For example, setting
+`ip=10.10.10.2::10.10.10.254:255.255.255.0:core0.example.com:enp1s0:none nameserver=4.4.4.41` sets:
+* Node's IP address to 10.10.10.2
+* Gateway address to 10.10.10.254
+* Netmask to 255.255.255.0
+* Hostname to core0.example.com
+* The DNS server address to 4.4.4.41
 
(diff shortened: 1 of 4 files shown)


kalexand-rh (commented):
The values in your static IP example still need to be enclosed in backticks, but it otherwise LGTM!

modules/installation-user-infra-machines-iso.adoc
```diff
@@ -59,10 +59,19 @@ coreos.inst=yes
 coreos.inst.install_dev=sda <1>
 coreos.inst.image_url=<bare_metal_image_URL> <2>
 coreos.inst.ignition_url=http://example.com/config.ign <3>
+ip=<dhcp or static IP address> <4>
 ----
 <1> Specify the block device of the system to install to.
 <2> Specify the URL of the RAW image that you uploaded to your server.
 <3> Specify the URL of the Ignition config file for this machine type.
+<4> Set `ip=dhcp` or set an individual static IP address (ip=) and DNS server (nameserver=) on each node.
```
miabbott: Some advanced notes that may be unnecesary:

- You can specify multiple `ip=` lines if your system has multiple network interfaces.
- You can mix DHCP and static IP configurations when your system has multiple network interfaces, but you must specify the interface used.  (i.e. `ip=dhcp:enp1s0 ip=10.10.10.2::10.10.10.254:255.255.255.0:core0.example.com:enp1s1:none`)
- You can specify multiple instances of `nameserver=` to provide alternate DNS servers.

miabbott (approved):
Left a note about some advanced configurations that we've seen from customers, but they shouldn't block merging

modules/installation-user-infra-machines-iso.adoc
kalexand-rh: @miabbott, if you use multiple DNS servers, do you use a comma-separated list to specify them?

miabbott: Nope, just separate `nameserver=` entries, for example:

`ip=10.10.10.2::10.10.10.254:255.255.255.0:core0.example.com:enp2s0:none nameserver=4.4.4.41 nameserver=1.1.1.1 nameserver=8.8.8.8`

kalexand-rh: Thank you! I'm going to merge Chris's PR and open a follow-up to capture your notes.

PR #19349, opened 2020-01-28
fix link

vikram-redhat:
https://github.com/openshift/openshift-docs/pull/19295

@bobfuru I apologise. This is the correct path. It should traverse up twice and then back down using the actual folder names. I confused it by incorrectly fixing it the first time.


bobfuru:
> #19295
> 
> @bobfuru I apologise. This is the correct path. It should traverse up twice and then back down using the actual folder names. I confused it by incorrectly fixing it the first time.

Ah, yes. That makes sense now about using the folder names. Thanks!

PR #19365, opened 2020-01-28
modules/configuring-firewall: Document release signature sources

wking:
Reflecting [this configuration][1], which tells the cluster-version operator to pull signatures from Google Storage and/or Red Hat's mirrors ([the search order is an internal detail][2]).

[1]: https://github.com/openshift/cluster-update-keys/blob/cca4ce696383e70ae669e770bd63265a9540b721/manifests.rhel/0000_90_cluster-update-keys_configmap.yaml#L4-L5
[2]: https://github.com/openshift/cluster-version-operator/blob/54faf6fad0d4dfa7c2a7953076f608d018577fd1/pkg/verify/configmap.go#L33-L48

### DIFF
--- a/modules/configuring-firewall.adoc
+++ b/modules/configuring-firewall.adoc
@@ -75,23 +75,26 @@ Before you install {product-title}, you must configure your firewall to grant ac
 |URL | Function
 
 |`mirror.openshift.com`
-|Required to access mirrored installation content and images
+|Required to access mirrored installation content and images. This site is also a source of release image signatures, although the Cluster Version Operator needs only a single functioning source.
+
+|`storage.googleapis.com/openshift-release`
+|A source of release image signatures, although the Cluster Version Operator needs only a single functioning source.
 
 |`*.apps.<cluster_name>.<base_domain>`
-|Required to access the default cluster routes unless you set an ingress wildcard during installation
+|Required to access the default cluster routes unless you set an ingress wildcard during installation.
 
 |`quay-registry.s3.amazonaws.com`
-|Required to access Quay image content in AWS
+|Required to access Quay image content in AWS.
 
 |`api.openshift.com`
-|Required to check if updates are available for the cluster
+|Required to check if updates are available for the cluster.
 
 |`art-rhcos-ci.s3.amazonaws.com`
-|Required to download {op-system-first} images
+|Required to download {op-system-first} images.
 
 |`api.openshift.com`
-|Required for your cluster token
+|Required for your cluster token.
 
 |`cloud.redhat.com/openshift`
-|Required for your cluster token
+|Required for your cluster token.
 |===


kalexand-rh:
@gpei, will you PTAL?

gpei:
@kalexand-rh lgtm

modules/configuring-firewall.adoc
```diff
@@ -75,7 +75,10 @@ Before you install {product-title}, you must configure your firewall to grant ac
 |URL | Function
 
 |`mirror.openshift.com`
-|Required to access mirrored installation content and images
+|Required to access mirrored installation content and images.  Also a source of release image signatures, although the cluster-version operator needs only a single functioning source.
```
kalexand-rh:
```suggestion
|Required to access mirrored installation content and images. Although the Cluster Version Operator requires only a single functioning content source, it also is a source of release image signatures.
```

modules/configuring-firewall.adoc
```diff
@@ -75,7 +75,10 @@ Before you install {product-title}, you must configure your firewall to grant ac
...
+
+|`storage.googleapis.com/openshift-release`
+|A source of release image signatures, although the cluster-version operator needs only a single functioning source.
```
kalexand-rh:
```suggestion
|Although the Cluster Version Operator requires only a single functioning content source, it also is a source of release image signatures.
```

kalexand-rh (commented):
We call it the `Cluster Version Operator`.

Since you're adding a sentence in the table, each of the equivalent entries needs a period. (Let me know if you want me to do this in a follow-up PR.)

modules/configuring-firewall.adoc
wking: Let's sort out the wording in [this thread][1], and then come back to this entry.

[1]: https://github.com/openshift/openshift-docs/pull/19365#discussion_r374189442

modules/configuring-firewall.adoc
wking: I merged your suggestion, and then took a closer look while rebasing onto master.  I've since walked back most of the changes, but kept your "Cluster Version Operator".  My main concern with your:

> Although the Cluster Version Operator requires only a single functioning content source, it also is a source of release image signatures.

is that "it" reads as "the Cluster Version Operator", when it's the mirror that's the source of signatures.  Points I want to get across in this entry:

* `mirror.openshift.com` is a source of installation content and images (covered in existing wording that predates this PR).
* `mirror.openshift.com` is a source of release image signatures (new in this PR).
* The CVO only needs a single, functioning source of signatures.  So for this purpose you need `mirror.openshift.com` or `storage.googleapis.com/openshift-release`, but you don't *need* both (it's fine to allow both if you want).  (new in this PR)

I'm certainly open to other phrasing attempts that cover all three points, if you still think my current wording is unclear.

wking:
> Since you're adding a sentence in the table, each of the equivalent entries needs a period.

I've pushed c513fe5f7 -> 8e7a945a3 adding the trailing periods.

acomabon:
@kalexand-rh any chance of having this merged soon?

vikram-redhat:
@wking can this be merged after rebase? @kalexand-rh if you are ok with it? I see QE has acked it.

PR #19422, opened 2020-01-31
SRVKS-409 Serverless metering updates for k8s 1.16 (>=OCP4.3) metrics

maschmid:
https://issues.redhat.com/browse/SRVKS-409 

The pod_name and container_name metric labels have been renamed to pod and container, respectively in k8s 1.16 (since OCP 4.3)

see also https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.16.md#removed-metrics

(This needs to also be backported to OCP 4.3 and later)

### DIFF
--- a/modules/serverless-metering-datasources.adoc
+++ b/modules/serverless-metering-datasources.adoc
@@ -24,7 +24,7 @@ spec:
              label_serving_knative_dev_service,
              label_serving_knative_dev_revision)
           (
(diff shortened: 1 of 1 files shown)


abrennan89 (approved):
Looks good! I'll add this to the PRs for review/merge for the 1.4.0 release.
I guess we can consider this QE approved since the change comes from QE :smile: thanks @maschmid 

Q: Do we need to add something to the release notes about this change?

kalexand-rh (commented):
LGTM

PR #17896, opened 2019-11-05
WIP - Disconnected round two, with file system

smarterclayton:
4.3 contains changes that simplify working with credentials and true
disconnected. Updating the docs (draft) to capture the differences
and deemphasize the custom bastion, and encourage using your own.

More details about what is required from a mirror registry.

Try to split out the bits of steps for disconnected and connected
mirroring. There are connected mirror use cases.

This is still heavily draft, and has no catalog changes, but a
vehicle for discussion.

PR #19435, opened 2020-02-01
*: Rename bastion -> mirror host

wking:
This is part of decoupling our docs from the assumption that a bastion will exist.  For installation, we just care that there is a registry with the mirrored content that the cluster can pull from.  Having that
registry on a bastion makes creating the mirrored images easier, but as our existing docs hint, there are other ways you can get the mirrored images in to your mirror repository.  This commit borrows its motivation and some of its wording from @smarterclayton's [here][1].

Also push the port information into `<mirror_registry>`, because we want to make it clear that we don't require a port for these entries.  The previous content was not consistent about whether placeholder were
referenced with or without their wrapping `<>`.  In my new text, I've included the `<>` to make it more clear that they are part of the placeholder, and not literal characters that should remain after placeholder substitution.

CC @kalexand-rh.  There's a lot of duplicate-ish content in here.  Any ideas for DRYing it up?

[1]: https://github.com/openshift/openshift-docs/pull/17896#discussion_r342723419

### DIFF
--- a/installing/installing_aws/installing-restricted-networks-aws.adoc
+++ b/installing/installing_aws/installing-restricted-networks-aws.adoc
@@ -27,7 +27,7 @@ according to your company's policies.
 +
 [IMPORTANT]
 ====
-Because the installation media is on the mirror host, you may use that computer
+Because the installation media is on the mirror host, you can use that computer
 to complete all installation steps.
 ====
 * Review details about the
@@ -121,7 +121,7 @@ include::modules/installation-creating-aws-worker.adoc[leveloffset=+2]
 
 include::modules/installation-cloudformation-worker.adoc[leveloffset=+3]
 
-//You may install the CLI on the mirror host.
+//You can install the CLI on the mirror host.
 
 include::modules/cli-logging-in-kubeadmin.adoc[leveloffset=+1]
 
--- a/installing/installing_aws/installing-restricted-networks-aws.adoc
+++ b/installing/installing_aws/installing-restricted-networks-aws.adoc
@@ -22,12 +22,12 @@ according to your company's policies.
 
 .Prerequisites
 
-* xref:../../installing/install_config/installing-restricted-networks-preparations.adoc#installing-restricted-networks-preparations[Create a mirror registry on your bastion host]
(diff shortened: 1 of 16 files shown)


installing/installing_aws/installing-restricted-networks-aws.adoc
```diff
@@ -22,12 +22,12 @@ according to your company's policies.
...
 .Prerequisites
 
-* xref:../../installing/install_config/installing-restricted-networks-preparations.adoc#installing-restricted-networks-preparations[Create a mirror registry on your bastion host]
+* xref:../../installing/install_config/installing-restricted-networks-preparations.adoc#installing-restricted-networks-preparations[Create a mirror registry on your mirror host]
  and obtain the `imageContentSources` data for your version of {product-title}.
 +
 [IMPORTANT]
 ====
-Because the installation media is on the bastion host, use that computer
+Because the installation media is on the mirror host, you may use that computer
```
kalexand-rh: I get that you're trying to step back from the bastion, but do you have a particular workflow in mind? "May implies permission, and that's not really what we mean here.

I'd either ~you may~ or, if you want to back it off more and still want them to use the installer that they generated from the mirror repo, say something like "Because the installation media is on the mirror host, the following steps assume that you use that computer to complete all installation steps. If you cannot complete the {product-title} installation from the mirror host, use the method that best fits your restrictions to ensure that you use the installation program that you generated from the mirrored content."

kalexand-rh (commented):
Our style guide forbids the use of "may," and that's really the only issue that i see. Most of the repetition is either due to the fact that there's a separate assembly for each disconnected installation or the fact that there are separate yaml samples for each installation type. (If I see a good opportunity to clean up the latter, I will.)

kalexand-rh:
@wking, will you PTAL at https://github.com/openshift/openshift-docs/pull/19435/commits/6973df0919bb14e3839e68dd4bb84af6ae22afc0? I rebased and made a few style edits.

kalexand-rh:
@jianlinliu,will you PTAL?

@wking, I want to merge this as soon as Jianlin blesses it so I can resolve any conflicts with Clayton's PR.

jianlinliu:
LGTM

wking:
Don't block on me.  I can always file further suggestions in the off chance I notice anything post-merge.

PR #19439, opened 2020-02-01
fix CNV dir structure

vikram-redhat:
@aburdenthehand @ousleyp


vikram-redhat:
Not merging just yet, cos we probably need to setup redirects from the incorrect path to new ones. Will update this PR with the redirects.

vikram-redhat:
Have added redirects as well and this should be ready to merge once we have an update on the search frequency. @ousleyp @aburdenthehand @bgaydosrh @lmandavi  FYI.

PR #19447, opened 2020-02-03
comments about folder names

kalexand-rh:
@vikram-redhat, PTAL?


vikram-redhat:
Thanks @kalexand-rh this looks good and is explicit. Merge away.

PR #19453, opened 2020-02-03
BZ-1667744: Fix repo name for power-9 reposync

andymcc:


kalexand-rh:
Barry Donahue approved this change through DM, so I'm going to merge it.

PR #19461, opened 2020-02-03
Ensure product name is capitalized

jboxman:
@vikram-redhat, I notice the name for these shows up as _OpenShift Online <lowercase>_ in the docs but on the product page, the entire name is capitalized.

https://www.openshift.com/products/online/

### DIFF
--- a/_distro_map.yml
+++ b/_distro_map.yml
@@ -44,10 +44,10 @@ openshift-online:
   site_url: https://docs.openshift.com/
   branches:
     enterprise-3.11:
-      name: 'pro'
+      name: 'Pro'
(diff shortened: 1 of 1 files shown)


vikram-redhat:
@ahardin-rh do you recall if we wanted them to be lower case in the docs?

ahardin-rh:
@vikram-redhat No, I don't think so. I think it should be capitalized. Thanks @jboxman

jboxman:
@vikram-redhat, is _master_ the proper place for this to merge and work?

vikram-redhat:
@jboxman Yes that is correct.

PR #19483, opened 2020-02-03
[enterprise-4.2] Remove new docs on configuring custom PKI using Proxy API

codyhoag:
Not yet backported to 4.2 (4.2 progress tracked [here](https://issues.redhat.com/browse/MON-884)). Only available in 4.3 (which is already published).


codyhoag:
@danehans can you give a final thumbs up that this should be removed from 4.2 docs until it's backported?

Preview for setting custom default certificate: http://file.rdu.redhat.com/~choag/02032020/custom-pki-revert/networking/ingress-operator.html#nw-ingress-setting-a-custom-default-certificate_configuring-ingress

(Configuring a custom PKI assembly is removed.)

modules/nw-ingress-setting-a-custom-default-certificate.adoc
```diff
@@ -25,6 +23,13 @@ NAME      AGE
 default   10m
 ----
 
+[WARNING]
```
kalexand-rh: Is this warning an elaboration of the first prereq? If so, I'd make it active and attach it to that prereq.

(Will the user know how to make sure the CA bundle is provided by the container userspace? (Also, userspace is only used once in the collection. Is there another way to convey this idea?)

kalexand-rh (commented):
Two questions, and it LGTM

codyhoag:
This has been approved via email. The ability to privately sign certs was confirmed in 4.2.x, but it only worked in certain cases. The [official backport](https://issues.redhat.com/browse/MON-884) is still in progress; we'll add this content back when the backport has finished.

PR #19498, opened 2020-02-04
updates for BZ1728350

lamek:


lamek:
@timflannagan1 PTAL.

modules/metering-store-data-in-s3.adoc
```diff
@@ -119,3 +119,5 @@ If you left `spec.storage.hive.s3.createBucket` set to `true`, or unset, then yo
     ]
 }
 ----
+
+Metering does not manage or delete any s3 bucket data. When uninstalling metering, any s3 buckets created for metering storage must be manually cleaned up.
```
timflannagan1: I think we should move this higher up to avoid the situation where a user skips over some key information.

timflannagan1: I think we should also specify that this applies to any s3 bucket used for metering and not just buckets created by Metering, i.e. "..., any s3 buckets used to store metering data must be manually cleaned up."

lamek: Sounds good. I made some updates per your feedback.

modules/metering-store-data-in-s3.adoc
```diff
@@ -5,6 +5,8 @@
 [id="metering-store-data-in-s3_{context}"]
 = Storing data in Amazon S3
 
+Metering can use an existing Amazon S3 bucket or create a new bucket for storage. Metering does not manage or delete any S3 bucket data. When uninstalling metering, any S3 buckets used to store metering data must be manually cleaned up.
```
kalexand-rh: Can you uninstall metering? There's no topic for it, and this information would be far more helpful in an uninstalling topic.

~new~

Consider refocusing the last sentence on the fact that you must manage the S3 data yourself and using uninstalling as an example.

kalexand-rh (commented):
I'd consider rewriting the module to better match the procedure template, and I think that documenting how to uninstall the metering service would help address the root cause.

modules/metering-store-data-in-s3.adoc
lamek: Great suggestion. We were discussing the same thing on the metering team. I'm going to do a separate PR for a section on "Deleting Reports". I'll add in a section for "Uninstalling Metering."

PR #19518, opened 2020-02-04
Minimize words for procedure steps

jboxman:
- Clean up 'the following command'

@openshift/team-documentation, so this came up in a recent minimalism session. I have many other files like this, but don't want to proceed unless it actually seems logical to omit instances of this. The consensus was this text is unnecessary when introducing a command to execute.

### DIFF
--- a/modules/nw-cluster-network-operator.adoc
+++ b/modules/nw-cluster-network-operator.adoc
@@ -14,7 +14,7 @@ using a DaemonSet.
 The Cluster Network Operator is deployed during installation as a Kubernetes
 `Deployment`.
 
-. Run the following command to view the Deployment status:
+. View the Deployment status:
 +
 ----
 $ oc get -n openshift-network-operator deployment/network-operator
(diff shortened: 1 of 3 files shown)


kalexand-rh:
This is an IBM Style thing. All of the CLI examples, which start on page 189, either include "the following command" or the name of the command itself.

@maxwelldb, do you remember if this style choice was to improve accessibility?

maxwelldb:
I don't recall it being specifically to do with accessibility, but hey, the overlap between usability and accessibility is pretty heavy. Seems like the kind of touchstone that could be useful if you're scanning or relying on a screenreader. 

It is _probably_ a boon for translators, too. 

ISG does also explicitly say on 185 of 7 - _Commands, parameters, and options in instructions_: 

> To instruct users to enter commands, use _type_, _enter_, or _issue_. Use one term consistently throughout documentation. 

Additionally, related to this PR:

> Do not use _run_ to refer to specifying a command on a command line or at a command prompt. However, you can use _run_ to refer to the general use of a command.

jboxman:
@kalexand-rh @maxwelldb 

That's a lot to think about, thanks.

jboxman:
Based on the feedback from the last meeting, I'm going to close this and retain the language.

PR #19525, opened 2020-02-05
improved docs for openshift 3.11

CrisMon-01:
on some distribution of oc 3.11 there is no oc whoami --token command, you had to use -t


CrisMon-01 (commented):
more precision

jboxman:
I'm going to fix this with https://github.com/openshift/openshift-docs/pull/20394.

jboxman:
Fixed, thanks!

PR #19538, opened 2020-02-05
quick spelling updates for metering

lamek:


lamek:
@timflannagan1 please let me know if you saw anything else. Thanks.

modules/metering-cluster-usage-examples.adoc
```diff
@@ -24,4 +24,4 @@ spec:
 <1> To say organized, remember to change the name of your Report if you change any of the other values.
```
timflannagan1: s/say/stay

timflannagan1: This also needs to be updated in the "Measure cluster utilization using cron expressions" section too.

timflannagan1:
In "Measure cluster utilization using cron expressions":

You can also use cron expressions when configuring the period of your repots. The following report measure cluster utilization by looking at CPU utilization from 9am-5pm every weekday.

s/repots/reports

PR #19568, opened 2020-02-06
Change the apiVersion to apps/v1

rh-max:


lilic (approved):
Thanks! 👍 

/lgtm

rh-max:
@soltysh Changing apiVersion per your request. The other parts of the docs will do the same shortly.

LiliC:
Thanks for spotting it @soltysh :)

rh-max:
@LiliC Thank you!
@juzhao Could you please review the change?

juzhao:
LGTM if the change is for 4.4 only, 4.3 can still use "extensions/v1beta1"

rh-max:
@juzhao According to @LiliC and @soltysh, while `extensions/v1beta1` _can_ be used in 4.3, it is deprecated and should be avoided.

LiliC:
4.3 should also follow best practices and migrate from using the extensions/v1beta1, so we should avoid mentioning or encouraging users to use it. There are even some ideas of adding alerts to 4.3 that would fire if you are using that api. So switching to both 4.3 and 4.4 is the right thing to do here.

modules/monitoring-configuring-monitoring-for-an-application.adoc
```diff
@@ -19,7 +19,7 @@ This procedure shows, on an example, how an application developer can deploy an
 +
 [source,yaml]
 ----
-apiVersion: extensions/v1beta1
```
LiliC: I get the following error when trying to apply this file via the UI in a 4.3 cluster:

> An error occurred
> Error "Required value" for field "spec.selector".

lilic (changes requested):

modules/monitoring-configuring-monitoring-for-an-application.adoc
LiliC: We need to also update the very last thing to replace it with:

> oc port-forward -n openshift-user-workload-monitoring svc/prometheus-operated 9090

we did not include namespace.

LiliC: There was a bug in the app can you also change the image to: `quay.io/brancz/prometheus-example-app:v0.2.0`

This is my final deployment:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: example-app
  namespace: default
spec:
  selector:
    matchLabels:
      app: example-app
  replicas: 1
  template:
    metadata:
      labels:
        app: example-app
    spec:
      containers:
      - name: example-app
        image: quay.io/brancz/prometheus-example-app:v0.2.0
        ports:
        - name: web
          containerPort: 8080
---
```

soltysh (approved):
lgtm, except the comment from Lili

lilic (approved):
👍 
/lgtm

rh-max:
Thanks @soltysh for reporting, thanks @LiliC and @juzhao for help.
@bmcelvee, could you

bmcelvee:
LGTM - before merging, I can't seem to find where these are building in the docs. Would you mind linking preview builds?

Thanks! :)

lilic (commented):
/lgtm 

👍

rh-max:
@bmcelvee The only updates that go to the published docs are in`modules/monitoring-deploying-a-sample-service.adoc`, as built here https://monitoring-change-apiversion--ocpdocs.netlify.com/openshift-enterprise/latest/monitoring/monitoring-your-own-services.html#deploying-a-sample-service_monitoring-your-own-services (please re-check, contains two total fixes now). The rest of the updates went into the docs that might be published in a future release, but are currently not included and so are not available for a preview.

PR #19597, opened 2020-02-07
[odo] update nodejs references to latest versions

yselkowitz:
RHOAR Node.js images are no longer supported, and RHSCL Node.js 8 is deprecated.

CC: @lholmquist

### DIFF
--- a/modules/templates-creating-from-console.adoc
+++ b/modules/templates-creating-from-console.adoc
@@ -27,16 +27,16 @@ metadata:
   name: "ruby"
   creationTimestamp: null
 spec:
-  dockerImageRepository: "registry.redhat.io/openshift3/ruby-20-rhel7"
+  dockerImageRepository: "registry.redhat.io/rhscl/ruby-26-rhel7"
   tags:
     -
-      name: "2.0"
+      name: "2.6"
       annotations:
-        description: "Build and run Ruby 2.0 applications"
+        description: "Build and run Ruby 2.6 applications"
         iconClass: "icon-ruby"
         tags: "builder,ruby" <1>
-        supports: "ruby:2.0,ruby"
-        version: "2.0"
+        supports: "ruby:2.6,ruby"
+        version: "2.6"
 ----
 <1> Including *builder* here ensures this `ImageStreamTag` appears in the
 web console as a builder.


lholmquist (approved):

kalexand-rh:
@vikram-redhat, who should review this?

vikram-redhat:
@Preeticp PTAL

Preeticp:
Thank you @vikram-redhat will follow up on this.

Preeticp:
@boczkowska can you or Amit please verify this and provide your comments?

cli_reference/openshift_developer_cli/understanding-odo.adoc
```diff
@@ -45,11 +45,9 @@ In {odo-title}, services are provisioned from the OpenShift Service Catalog and
 |====
 |Node.js|Java
 |`centos/nodejs-8-centos7`|`redhat-openjdk-18/openjdk18-openshift`
-|`rhoar-nodejs/nodejs-8`|`openjdk/openjdk-11-rhel8`
-|`rhoar-nodejs/nodejs-10`|`openjdk/openjdk-11-rhel7`
+|`rhscl/nodejs-10-rhel7`|`openjdk/openjdk-11-rhel8`
+|`rhscl/nodejs-12-rhel7`|`openjdk/openjdk-11-rhel7`
```
amitkrout: We are still using the name space ```rhoar-nodejs``` in our actual odo code - https://github.com/openshift/odo/blob/425d77670264efc502e02aed28a5665546bfefd8/pkg/catalog/catalog.go#L282-L290. Can you please share the supported reference to this change.

@kadel If this is the case then we need to update the image registry in our odo code base too. PTAL

lholmquist: This was the blog post that sort of talks about the 1 node.js from redhat,  https://developers.redhat.com/blog/2020/01/07/red-hat-support-for-node-js/

I noticed in that code that there are also references to bucharestgold/node images.  We've actually changed the org name to nodeshift,  but we've actually replaced those community images with the ubi based ones from rhscl.   looks like that should be another PR :)

lholmquist: Also,  related to my last comment,  Node 8 went EOL at the end of 2019,  so those references should be removed and Node 12 can be added

yselkowitz:
/hold
This requires similar changes in odo itself, per earlier comments.

girishramnani:
I tried the odo catalog command on `3.11` openshift cluster and that still has nodejs 8 image stream. We are still supporting 3.11` cluster in odo and so I woulds suggest to **not** remove node 8 support from the documentation.

girishramnani:
opened an issue to track the task required on odo side https://github.com/openshift/odo/issues/2603

yselkowitz:
> I tried the odo catalog command on `3.11` openshift cluster and that still has nodejs 8 image stream. We are still supporting 3.11` cluster in odo and so I would suggest to **not** remove node 8 support from the documentation.

nodejs-8-rhel7 as well as both versions of rhoar-nodejs are no longer supported.  While OpenShift does not update imagestreams in stable branches, that does not mean we should encourage further use of unsupported versions.

PR #19617, opened 2020-02-07
OSDOCS-884 document DNS forwarding

bmcelvee:
https://issues.redhat.com/browse/OSDOCS-884

### DIFF
--- a/modules/nw-dns-forward.adoc
+++ b/modules/nw-dns-forward.adoc
@@ -1,0 +1,98 @@
+// Module included in the following assemblies:
+//
+// * networking/dns-operator.adoc
+
+[id="nw-dns-forward_{context}"]
+= Using DNS forwarding
+
+You can use DNS forwarding to override the forwarding configuration identified in `etc/resolv.conf` on a per-zone basis by specifying which name server should be used for a given zone.
+
+.Procedure
+
+. Modify the DNS Operator object named `default`:
++
+----
+$ oc edit dns.operator/default
+----
++
+This allows the Operator to create and update the ConfigMap named `dns-default` with additional server configuration blocks based on `Server`. If none of the servers has a zone that matches the query, then name resolution falls back to the name servers that are specified in `/etc/resolv.conf`.
++
+.Sample DNS
+[source,yaml]
+----
+apiVersion: operator.openshift.io/v1
+kind: DNS
+metadata:
+  name: default
+spec:
+  servers:
+  - name: foo-server <1>
+    zones: <2>
+      - foo.com
+    forwardPlugin:
+      upstreams: <3>
+        - 1.1.1.1
+        - 2.2.2.2:5353
+  - name: bar-server
+    zones:
+      - bar.com
+      - example.com
+    forwardPlugin:
+      upstreams:
+        - 3.3.3.3
+        - 4.4.4.4:5454
+----
+<1> `name` must comply with the `rfc6335` service name syntax.
+<2> `zones` must conform to the definition of a `subdomain` in `rfc1123`. The cluster domain, `cluster.local`, is an invalid `subdomain` for `zones`.
+<3> A maximum of 15 `upstreams` is allowed per `forwardPlugin`.
++
+[NOTE]
+====
+If `servers` is undefined or invalid, the ConfigMap only contains the default server.
+====
++
+. View the ConfigMap:
++
+----
+$ oc get configmap/dns-default -n openshift-dns -o yaml
+----
++
+.Sample DNS ConfigMap based on previous sample DNS
+[source,yaml]
+----
+apiVersion: v1
+data:
+  Corefile: |
+    foo.com:5353 {
+        forward . 1.1.1.1 2.2.2.2:5353
+    }
+    bar.com:5353 example.com:5353 {
+        forward . 3.3.3.3 4.4.4.4:5454 <1>
+    }
+    .:5353 {
+        errors
+        health
+        kubernetes cluster.local in-addr.arpa ip6.arpa {
+            pods insecure
+            upstream
+            fallthrough in-addr.arpa ip6.arpa
+        }
+        prometheus :9153
+        forward . /etc/resolv.conf {
+            policy sequential
+        }
+        cache 30
+        reload
+    }
+kind: ConfigMap
+metadata:
+  labels:
+    dns.operator.openshift.io/owning-dns: default
+  name: dns-default
+  namespace: openshift-dns
+----
+<1> Changes to the `forwardPlugin` triggers a rolling update of the CoreDNS DaemonSet.
+
+.Additional resources
+
+* For more information on DNS forwarding, see the link:https://coredns.io/plugins/forward/[CoreDNS forward documentation].
--- a/modules/nw-dns-view.adoc
+++ b/modules/nw-dns-view.adoc
@@ -5,8 +5,7 @@
 [id="nw-dns-view_{context}"]
 = View the default DNS
 
-Every new {product-title} installation has a `dns.operator` named `default`. It
-cannot be customized, replaced, or supplemented with additional `dnses`.
+Every new {product-title} installation has a `dns.operator` named `default`.
 
 .Procedure
 
@@ -38,8 +37,3 @@ use the `oc get` command:
 $ oc get networks.config/cluster -o jsonpath='{$.status.serviceNetwork}'
 [172.30.0.0/16]
 ----
-
-[NOTE]
-====
-Configuration of the CoreDNS Corefile or Kubernetes plugin is not supported.
-====
--- a/networking/dns-operator.adoc
+++ b/networking/dns-operator.adoc
@@ -13,6 +13,8 @@ include::modules/nw-dns-operator.adoc[leveloffset=+1]
 
 include::modules/nw-dns-view.adoc[leveloffset=+1]
 
+include::modules/nw-dns-forward.adoc[leveloffset=+1]
+
 include::modules/nw-dns-operator-status.adoc[leveloffset=+1]
 
 include::modules/nw-dns-operator-logs.adoc[leveloffset=+1]


bmcelvee:
This may also need to go in 4.3, but I will verify before adding the label.

bmcelvee:
Preview build: https://osdocs-884-document-dns-forwarding--ocpdocs.netlify.com/openshift-enterprise/latest/networking/dns-operator.html#nw-dns-forward_dns-operator

bmcelvee:
@ironcladlou, @danehans, or @Miciah - would one of you mind giving this PR a review? Thanks!

modules/nw-dns-forward.adoc
```diff
@@ -0,0 +1,97 @@
...
+You can use forwarding with OpenShift DNS to resolve name queries for your other internal devices using the DNS servers in your data center.
+
+[NOTE]
+====
+The DNS Operator prepends the configuration of `Servers` to `configmap/dns-default`, instead of the entire ConfigMap being statically defined.
+====
+
+.Prerequisites
+
+* AWS user-provisioned infrastructure
```
Miciah: DNS forwarding works on all platforms.

modules/nw-dns-forward.adoc
```diff
@@ -0,0 +1,97 @@
...
+
+.Procedure
+
+. Remove the default ConfigMap asset and code from the `manifests.pkg` file.
```
Miciah: What is the `manifests.pkg` file?  I think this step can be dropped.

modules/nw-dns-forward.adoc
```diff
@@ -0,0 +1,97 @@
...
++
+. Modify the DNS ConfigMap to create a ConfigMap with additional server configuration blocks based on `Server`. Each `Server` has a forwarding configuration and uses `/etc/resolv.conf` as a resolver of last resort.
```
Miciah: The user modifies the `dns.operator` object named `default`; the operator creates and updates the `ConfigMap` named `dns-default`.  The phrase "last resort" seems a little strong; if none of the servers has a zone that matches the query, then name resolution falls back to the name servers that are specified in `/etc/resolv.conf`.

modules/nw-dns-forward.adoc
```diff
@@ -0,0 +1,97 @@
...
+// * dns/dns-operator.adoc
+
+[id="nw-dns-forward_{context}"]
+= Using DNS forwarding on user-provisioned infrastructure in AWS
+
+You can use forwarding with OpenShift DNS to resolve name queries for your other internal devices using the DNS servers in your data center.
+
+[NOTE]
+====
+The DNS Operator prepends the configuration of `Servers` to `configmap/dns-default`, instead of the entire ConfigMap being statically defined.
```
Miciah: I'm not sure what this note is saying.  Is this referring to falling back to the name servers in `/etc/resolv.conf`?

modules/nw-dns-forward.adoc
```diff
@@ -0,0 +1,97 @@
+// Module included in the following assemblies:
+//
+// * dns/dns-operator.adoc
+
+[id="nw-dns-forward_{context}"]
+= Using DNS forwarding on user-provisioned infrastructure in AWS
```
Miciah: This is not specific to AWS.

modules/nw-dns-forward.adoc
```diff
@@ -0,0 +1,97 @@
...
+
+You can use forwarding with OpenShift DNS to resolve name queries for your other internal devices using the DNS servers in your data center.
```
Miciah: By default, the cluster DNS service will resolve internal cluster names and forward requests for external names to the name servers that are specified in `/etc/resolv.conf`.  The purpose of the DNS forwarding API is to allow overriding the forwarding configuration on a per-zone basis, by specifying which name server should be used for a given zone.

modules/nw-dns-forward.adoc
bmcelvee: Thanks, @Miciah! I leaned on the enhancement doc to doc DNS forwarding, but it seems like it was out of date. Updates are coming.

modules/nw-dns-forward.adoc
bmcelvee: Would it be appropriate to introduce DNS forwarding as a condensed version of your explanation, .e.g.,:

"You can use DNS forwarding to override the forwarding configuration identified in `etc/resolv.conf` on a per-zone basis by specifying which name server should be used for a given zone."

Or would you mind if I use your explanation?

Miciah: I think that looks better, yeah.

modules/nw-dns-forward.adoc
```diff
@@ -0,0 +1,86 @@
...
+// * dns/dns-operator.adoc
+
+[id="nw-dns-forward_{context}"]
+= Using DNS forwarding
+
+You can use DNS forwarding to override the forwarding configuration identified in `etc/resolv.conf` on a per-zone basis by specifying which name server should be used for a given zone.
+
+.Procedure
+
+* Modify the DNS Operator object `name: default`. This allows the Operator to create and update the ConfigMap named `dns-default` with additional server configuration blocks based on `Server`. If none of the servers has a zone that matches the query, then name resolution falls back to the name servers that are specified in `/etc/resolv.conf`.
```
Miciah: Sorry if I my earlier comment wasn't clear; there is an object that has the name "default" and the kind "DNS" in the "operator.openshift.io" group.  It can be referenced in `oc` commands as `dnses.operator.openshift.io/default` or just `dnses.operator/default` (either ".operator" or "operator.openshift.io" is needed because there is also a "DNS" kind in the "config.openshift.io" group).  So this line should have ``` named `default` ``` rather than ``` `name: default` ```.

bmcelvee: Oh, thank you!

bmcelvee:
@lihongan PTAL, thanks!

modules/nw-dns-forward.adoc
```diff
@@ -0,0 +1,86 @@
...
+* Modify the DNS Operator object named `default`. This allows the Operator to create and update the ConfigMap named `dns-default` with additional server configuration blocks based on `Server`. If none of the servers has a zone that matches the query, then name resolution falls back to the name servers that are specified in `/etc/resolv.conf`.
++
+[NOTE]
+====
+If `Server` is undefined or invalid, the ConfigMap only contains the default server.
+====
++
+.Sample DNS
+[source,YAML]
+----
```
lihongan: better to add the line `$ oc edit dns.operator/default` here.

modules/nw-dns-forward.adoc
```diff
@@ -0,0 +1,86 @@
...
+        - 3.3.3.3
+        - 4.4.4.4:5454
+----
+<1> `Name` must comply with the `rfc6335` service name syntax.
+<2> `Zones` must conform to the definition of a `subdomain` in `rfc1123`. The cluster domain, `cluster.local`, is an invalid `subdomain` for `Zones`.
+<3> A maximum of 15 `Upstreams` is allowed per `ForwardPlugin`.
++
+.Sample DNS ConfigMap based on previous sample DNS
+[source,YAML]
+----
```
lihongan: adding the line `$ oc get configmap/dns-default -n openshift-dns -o yaml`

lihongan:
@bmcelvee just two minor comments, others looks good. thank you.

modules/nw-dns-forward.adoc
bmcelvee: Added!

modules/nw-dns-forward.adoc
bmcelvee: Added!

bmcelvee:

modules/nw-dns-forward.adoc
```diff
@@ -0,0 +1,97 @@
+// Module included in the following assemblies:
+//
+// * dns/dns-operator.adoc
```
bergerhoffer: Should be `networking/dns-operator.adoc` I assume?

modules/nw-dns-forward.adoc
```diff
@@ -0,0 +1,97 @@
...
+You can use DNS forwarding to override the forwarding configuration identified in `etc/resolv.conf` on a per-zone basis by specifying which name server should be used for a given zone.
+
+.Procedure
+
+. Modify the DNS Operator object named `default`:
++
+----
+$ oc edit dns.operator/default
+----
+.. This allows the Operator to create and update the ConfigMap named `dns-default` with additional server configuration blocks based on `Server`. If none of the servers has a zone that matches the query, then name resolution falls back to the name servers that are specified in `/etc/resolv.conf`.
```
bergerhoffer: Consider whether or not this needs to be a step (1a) or if it can just be included as text under step 1. Kind of like how step 2 does it. Just good to be consistent, and since there's only 1 substep there, probably not needed.

modules/nw-dns-forward.adoc
```diff
@@ -0,0 +1,97 @@
...
+        - 3.3.3.3
+        - 4.4.4.4:5454
+----
+<1> `Name` must comply with the `rfc6335` service name syntax.
+<2> `Zones` must conform to the definition of a `subdomain` in `rfc1123`. The cluster domain, `cluster.local`, is an invalid `subdomain` for `Zones`.
+<3> A maximum of 15 `Upstreams` is allowed per `ForwardPlugin`.
++
+[NOTE]
+====
+If `Server` is undefined or invalid, the ConfigMap only contains the default server.
```
bergerhoffer: Just wondering what this `Server` item references. I don't see anything in the example yaml that has a capital `Server`.

Same question in the `Server` reference in the step above.

modules/nw-dns-forward.adoc
```diff
@@ -0,0 +1,97 @@
...
+  - name: bar-server
+    zones:
+      - bar.com
+      - example.com
+    forwardPlugin:
+      upstreams:
+        - 3.3.3.3
+        - 4.4.4.4:5454
+----
+<1> `Name` must comply with the `rfc6335` service name syntax.
```
bergerhoffer: You have all of these fields capitalized (Name, Zones, Upstreams, etc.) but they're not capitalized in the yaml. Suggest lowering.

modules/nw-dns-view.adoc
```diff
@@ -5,8 +5,9 @@
 [id="nw-dns-view_{context}"]
 = View the default DNS
 
-Every new {product-title} installation has a `dns.operator` named `default`. It
-cannot be customized, replaced, or supplemented with additional `dnses`.
+Every new {product-title} installation has a `dns.operator` named `default`.
+
+// It cannot be customized, replaced, or supplemented with additional `dnses`.
```
bergerhoffer: Not critical but just wondering if this needed to be kept or not.

modules/nw-dns-forward.adoc
```diff
@@ -0,0 +1,97 @@
...
+
+. Modify the DNS Operator object named `default`:
++
+----
+$ oc edit dns.operator/default
+----
+.. This allows the Operator to create and update the ConfigMap named `dns-default` with additional server configuration blocks based on `Server`. If none of the servers has a zone that matches the query, then name resolution falls back to the name servers that are specified in `/etc/resolv.conf`.
++
+.Sample DNS
+[source,YAML]
```
bergerhoffer: Not sure if it renders the same capitalized, but suggest `source,yaml` to be consistent throughout our doc set.

bergerhoffer (commented):
Just a few comments.

bmcelvee:
Changed the milestone from future release to next release because the feature was backported.

PR #19618, opened 2020-02-07
Add info about machine delete policies

chaitanyaenr:
This commit:
- Adds information about the available machineset delete
policies which are Random, Newest and Oldest as well the default
policy used i.e Random. The priority of the machines to be deleted
can also be set by adding the machine.openshift.io/cluster-api-delete-machine
annotation to the machines.
- Adds information about using custom/pet machinesets for use cases which
needs the services to run on particular nodes and would want the controller
to ignore them during scale down.

### DIFF
--- a/modules/machineset-manually-scaling.adoc
+++ b/modules/machineset-manually-scaling.adoc
@@ -36,7 +36,26 @@ $ oc edit machineset <machineset> -n openshift-machine-api
 You can scale the MachineSet up or down. It takes several minutes for the new
 machines to be available.
 +
+
+= The MachineSet delete policy
+`Random`, `Newest`, and `Oldest` are the three supported options. The default is `Random`, meaning  that random machines are chosen and deleted when scaling MachineSets down. The delete policy can be set according to the use case by modifying the particular MachineSet:
+
+----
+
+spec:
+  deletePolicy: <delete policy>
+  replicas: <desired replica count>
+
+----
+
+Specific machines can also be priortized for deletion by adding the annotation `machine.openshift.io/cluster-api-delete-machine` to the machine of interest, regardless of the delete policy.
+
 [IMPORTANT]
 ====
 By default, the {product-title} router pods are deployed on workers. Because the router is required to access some cluster resources, including the web console, do not scale the worker MachineSet to `0` unless you first relocate the router pods.
+====
+
+[NOTE]
+====
+Custom MachineSets can be used for use cases requiring that services run on specific nodes and that those services are ignored by the controller when the worker MachineSets are scaling down. This prevents service disruption.
 ====


chaitanyaenr:
cc: @jtaleric.

modules/machineset-manually-scaling.adoc
```diff
@@ -36,7 +36,27 @@ $ oc edit machineset <machineset> -n openshift-machine-api
 You can scale the MachineSet up or down. It takes several minutes for the new
 machines to be available.
 +
+
+= MachineSet delete policy
+
+MachineSet delete policy defines how priority is assigned to nodes to delete when downscaling a MachineSet. There are three supported options - Random, Newest and Oldest. It default to Random which means that random machines are picked and deleted when downscaling MachineSets. The delete policy can be set according to the use case by modifying the particular MachineSet:
```
ahardin-rh: The MachineSet delete policy

    `Random`, `Newest`, and `Oldest` are the three supported options. The default is `Random`, meaning  that random machines are chosen and deleted when scaling MachineSets down.

modules/machineset-manually-scaling.adoc
```diff
@@ -36,7 +36,27 @@ $ oc edit machineset <machineset> -n openshift-machine-api
...
+
+----
+
+spec:
+  deletePolicy: <delete policy>
+  replicas: <desired replica count>
+
+----
+
+Specific machines can also be priortized for deletion by adding the annotation `machine.openshift.io/cluster-api-delete-machine` to the machine of interest regardless of the delete policy.
```
ahardin-rh: , regardless

modules/machineset-manually-scaling.adoc
```diff
@@ -36,7 +36,27 @@ $ oc edit machineset <machineset> -n openshift-machine-api
...
+
 [IMPORTANT]
 ====
 By default, the {product-title} router pods are deployed on workers. Because the router is required to access some cluster resources, including the web console, do not scale the worker MachineSet to `0` unless you first relocate the router pods.
 ====
+
+[NOTE]
+====
+Custom/Pet MachineSets can be used for the use cases which needs running services on particular nodes and have a need to be ignored by the controller during downscaling the MachineSets. This will prevent the services from getting disrupted as the scaling down the worker MachineSets would not affect it.
```
ahardin-rh: for use cases requiring that services run on specific nodes and that those services are ignored by the controller when the MachineSets are scaling down. This prevents service disruption.

chaitanyaenr:
@ahardin-rh thanks for the review, updated the PR with the requested changes. PTAL.

ahardin-rh:
@chaitanyaenr LGTM. This is good to go once we get a QE review. Thanks!

modules/machineset-manually-scaling.adoc
```diff
@@ -36,7 +36,26 @@ $ oc edit machineset <machineset> -n openshift-machine-api
...
+Specific machines can also be priortized for deletion by adding the annotation `machine.openshift.io/cluster-api-delete-machine` to the machine of interest, regardless of the delete policy.
+
 [IMPORTANT]
 ====
 By default, the {product-title} router pods are deployed on workers. Because the router is required to access some cluster resources, including the web console, do not scale the worker MachineSet to `0` unless you first relocate the router pods.
 ====
+
+[NOTE]
+====
+Custom/Pet MachineSets can be used for use cases requiring that services run on specific nodes and that those services are ignored by the controller when the MachineSets are scaling down. This prevents service disruption.
```
enxebre: I don't understand this statement? there's no such a thing like a "pet" machineSet. Any machineSet consider machines "cattle". Your workloads availability is ensured via your pod disruption budget.

mffiedler: Pet is not a good term.   Custom is probably better.   The use case is when you require a class of cloud instances which are different from your cattle.   e.g. My cattle need 2 vCPU and 8Gi, so the majority of my MachineSets use that instance type.    To get a different class of cloud instan ces for my Elastic clusters (8 vCPU, 32Gi), I need a machineset that spins up m5.2xlarge.   AFAIK the only way to address this use case today is with custom machinesets.

ahardin-rh: @chaitanyaenr Once you update the language here, we can get this moved forward. Thanks!

chaitanyaenr: @ahardin-rh updated, thanks.

PR #19675, opened 2020-02-11
bug 1781064 clarifying max nodes

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1781064

@jhou1, will you PTAL?

### DIFF
--- a/modules/cluster-autoscaler-about.adoc
+++ b/modules/cluster-autoscaler-about.adoc
@@ -11,38 +11,43 @@ provide infrastructure management that does not rely on objects of a specific
 cloud provider. The ClusterAutoscaler has a cluster scope, and is not associated
 with a particular namespace.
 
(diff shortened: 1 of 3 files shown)


jhou1:
@miyadav Will you PTAL?

miyadav:
The changed text doesn't address the customer issue which is mentioned in
the link [1] as it seems it was  interpreted as the keys(maxReplicas) in
the yaml  will be used directly for scaling the machines  , whereas it is
not for scaling but rather limit it once a scaling decision has been made
based upon workload.

Can we  put description for maxReplicas/min like : "The limit to the
max/min number Machines of the specified type to be deployed in the
specified AWS zone once the cluster autoscaler has decided for scaling the
cluster.
[1] https://access.redhat.com/solutions/4457721

kalexand-rh:
@miyadav, will you please take another look?

PR #19691, opened 2020-02-12
[enterprise-4.2] Added manual rollback procedure and updated known issues.

kalexand-rh:
From https://github.com/openshift/openshift-docs/pull/19319


modules/migration-known-issues.adoc
```diff
@@ -22,3 +22,4 @@ These annotations preserve the UID range, ensuring that the containers retain th
 * Cluster-scoped resources, including Cluster Role Bindings and Security Context Constraints, are not yet handled by the CAM. If your applications require cluster-scoped resources, you must create them manually on the target cluster. (link:https://bugzilla.redhat.com/show_bug.cgi?id=1759804[*BZ#1759804*])
 * Incorrect source cluster storage class is displayed when creating the migration plan. (link:https://bugzilla.redhat.com/show_bug.cgi?id=1777869[*BZ#1777869*])
 * If a cluster in the CAM web console becomes inaccessible, it blocks attempts to close open migration plans. (link:https://bugzilla.redhat.com/show_bug.cgi?id=1758269[*BZ#1758269*])
+* If a migration fails, the migration plan does not retain custom PV settings for quiesced pods. You must manually roll back the migration, delete the migration plan, and create a new migration plan with your PV settings. (link:https://bugzilla.redhat.com/show_bug.cgi?id=1784899[*BZ#1784899*])
```
kalexand-rh: @apinnick, I checked the bug, and it does apply for 4.2. If there's a reason that you don't want it in the release notes, please let me know so we can pull it.

apinnick: @kalexand-rh  Sorry, but I'm confused. I'm not familiar with your release note procedures. It's probably not a big deal, but is there any reason why I wouldn't want it in the release notes?

kalexand-rh: @apinnick, not that I'm aware of. I think it made most sense to include it.

PR #19693, opened 2020-02-12
updating Z versions per support

kalexand-rh:


modules/installation-requirements-user-infra.adoc
```diff
@@ -115,7 +115,7 @@ ifdef::ibm-z[]
 
 You can install {product-title} version {product-version} on the following IBM hardware:
 
-* IBM Z, version z13 and later and z13s and later
+* IBM Z, version 13, 14, or 15
```
sferich888: nit: version should be plural

kalexand-rh: Thank you! I have fixed that nit.

PR #19722, opened 2020-02-13
bug 1801528 correction machine deletion command

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1801528

@shellyyang1989, will you please confirm? Also, this bug was filed against 4.2, so I assume that it's valid for 4.3 and 4.4 as well. Is it also valid in 4.1?


shellyyang1989:
LGTM. It's valid for 4.1, 4.2, 4.3 and 4.4. Thanks

kalexand-rh:
Thank you! Merging.

PR #19736, opened 2020-02-13
Create dummy Release Notes for 4.4 on master

codyhoag:
@vikram-redhat I recall us having to create a dummy Release Notes doc on master to successfully build. I've created a 4.4 version for `master`, in anticipation for creating the real doc on `enterprise-4.4`. Can you take a look and see if this looks good? I plan on creating the doc stub on `enterprise-4.4` once `master` is in order. Thanks!


codyhoag:
@vikram-redhat Perhaps a cherrypick from this to `enterprise-4.4` is in order too. Then we can simply begin updating the Release Notes doc on the versioned branch. I'll add the tag and we can remove it if need be.

vikram-redhat:
Yes, I believe we keep one in the master just for validation of links etc, in the rare case we need to link to them. And changes go directly in the 4.4 branch. Thanks for creating it. Merge and CP away!

codyhoag:
Manual cherrypick, with updates for 4.4-specific release notes sent in #19742.

PR #19787, opened 2020-02-18
Cnv module metadata update 

aburdenthehand:
Restructuring the repo means a lot of changes to module metadata. This PR updates all that precious metadata, as well as correcting a few additional things along the way:
- all cnv metadata now employs `// * <assembly_name> (previously an inconsistent inclusion of asterix)
- all metadata now has an empty line immediately following the metadata (important for running scripts like this)
- all modules used in CNV now have completely up-to-date metadata (some had been neglected)

In addition, two modules jumped out:
* modules/cnv-accessing-vmi-web.adoc 
(Not used in any assembly)

* modules/gathering-data-specific-features.adoc
(seems to be cnv-specific but no cnv prefix? Also, declares dependency for something that we use regardless? maybe hangover for when it was used in ocp docs?)

### DIFF
--- a/modules/cnv-add-nic-to-vm.adoc
+++ b/modules/cnv-add-nic-to-vm.adoc
@@ -1,10 +1,7 @@
 // Module included in the following assemblies:
 //
-// * cnv/cnv_users_guide/cnv-edit-vms.adoc
-// * cnv/cnv_users_guide/cnv-editing-vm-template
-//
-// Establishing conditionals so content can be re-used for editing VMs
-// and VM templates.
+// * cnv/cnv_virtual_machines/cnv-edit-vms.adoc
+// * cnv/cnv_vm_templates/cnv-editing-vm-template.adoc
 
 ifeval::["{context}" == "cnv-edit-vms"]
 :object: virtual machine
--- a/modules/cnv-monitoring-upgrade-status.adoc
+++ b/modules/cnv-monitoring-upgrade-status.adoc
@@ -1,6 +1,6 @@
 // Module included in the following assemblies:
 //
-// * cnv/cnv_install/upgrading-container-native-virtualization.adoc
+// * cnv/upgrading-container-native-virtualization.adoc
 
 [id="cnv-monitoring-upgrade-status_{context}"]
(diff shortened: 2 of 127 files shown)


modules/cnv-monitoring-upgrade-status.adoc
```diff
@@ -52,4 +52,4 @@ Available          True  Reconcile completed successfully
 Progressing        False Reconcile completed successfully
 Degraded           False Reconcile completed successfully
 Upgradeable        True  Reconcile completed successfully
-----
\ No newline at end of file
+----
```
aburdenthehand: Not sure what git is picking up here. This code block hasn't changed from master

modules/cnv-add-nic-to-vm.adoc
```diff
@@ -1,10 +1,7 @@
 // Module included in the following assemblies:
 //
-// * cnv/cnv_users_guide/cnv-edit-vms.adoc
-// * cnv/cnv_users_guide/cnv-editing-vm-template
-//
-// Establishing conditionals so content can be re-used for editing VMs
```
aburdenthehand: Git makes it look like this comment has been removed, but it's still there in the file.

ousleyp:
> * modules/gathering-data-specific-features.adoc
>   (seems to be cnv-specific but no cnv prefix? Also, declares dependency for something that we use regardless? maybe hangover for when it was used in ocp docs?)

So, this might not make sense now, but when I created this module for CNV, it came to light that you would be able to use the same procedure to gather data from different features. I think that at the time, CNV was the only feature you could use it for (maybe?) but more were planned, so CNV just served as a very convenient example. Let me know if you think I should change this.

aburdenthehand:
@ousleyp nah yeah that makes sense. We can keep it as it is - do we still need the note on dependency?

ousleyp:
> @ousleyp nah yeah that makes sense. We can keep it as it is - do we still need the note on dependency?

It's still meant to be usable for OCP and they don't usually include our doc attributes, so I think so?

ousleyp (approved):
LGTM!

PR #19796, opened 2020-02-18
Clarify SAML support

stlaz:
The current wording may be interpreted as if there was no support
for SAML-providing identity providers, which is not true as in
such cases, the RequestHeader IdP can be used if a login proxy
is properly configured.


bergerhoffer:
LGTM

PR #19841, opened 2020-02-20
updated to unset variables

vikram-redhat:
@apinnick @bergerhoffer FYI

Content was rendering incorrectly on the customer portal and this fixes it (I will forward the email thread about it). Can you review and make sure that the unset variables are in the correct positions.

Also, I am only going to CP to 4.3 and 4.4, but feel free to CP to earlier branches if needed.


apinnick:
> @apinnick @bergerhoffer FYI
> 
> Content was rendering incorrectly on the customer portal and this fixes it (I will forward the email thread about it). Can you review and make sure that the unset variables are in the correct positions.
> 
> Also, I am only going to CP to 4.3 and 4.4, but feel free to CP to earlier branches if needed.

@vikram-redhat Thanks. These look OK.

apinnick:
@vikram-redhat Should I create a new PR for 4.2? Or did you fix the cherrypick?

PR #19844, opened 2020-02-20
Fix navigation path and CRD format in creating-custom-notification-banners

dmvolod:


ahardin-rh (commented):
LGTM. Thank you! We'll just need to squash the 2 commits into 1 commit before we can merge. Thanks again!

PR #19845, opened 2020-02-20
Fix yaml indention in adding-a-custom-logo

dmvolod:


ahardin-rh:
@dmvolod Is this change targeting both 4.3 and 4.4, or 4.4 only? As with the other PR, before we can merge, can you please squash these 2 commits down to 1 commit? Thank you!

dmvolod:
@ahardin-rh I checked this against 4.3 and will check on 4.4 but looks like this is fix for both.
I will rebase on master when one of the gitignore adds will be merged

ahardin-rh (approved):
:+1:

PR #19848, opened 2020-02-20
Fix navigation path in creating-custom-links

dmvolod:


ahardin-rh:
Confirmed in 4.4 cluster; no formal QE review necessary

Issue #19758, opened 2020-02-14
[enterprise-4.2] Edit suggested in file authentication/certificates/replacing-default-ingress-certificate.adoc


skamphax0r:
### Which section(s) is the issue in?

Replacing the default ingress certificate

### What needs fixing?

in https://access.redhat.com/documentation/en-us/openshift_container_platform/4.2/html/networking/configuring-ingress#nw-ingress-setting-a-custom-default-certificate_configuring-ingress there is a warning box about using non publicly signed CAs

in this doc there is not - depending on which doc the customer uses, this may cause confusion and support tickets.

codyhoag:
Good call! This is now being tracked in #19856. Thanks!

PR #19856, opened 2020-02-20
Clarify steps are intended for adding custom CA

codyhoag:
Suggested in #19758

### DIFF
--- a/modules/customize-certificates-replace-default-router.adoc
+++ b/modules/customize-certificates-replace-default-router.adoc
@@ -14,6 +14,13 @@ and CLI, will have encryption provided by specified certificate.
 
 * You must have a wildcard certificate and its private key,
 both in the PEM format, for use.
++
+[WARNING]
+====
+The new certificate *must* be signed by a public certificate authority already
+included in the CA bundle as provided by the container userspace.
+====
+
 * The certificate must have a `subjectAltName` extension of
 `*.apps.<clustername>.<domain>`.
 
--- a/modules/customize-certificates-replace-default-router.adoc
+++ b/modules/customize-certificates-replace-default-router.adoc
@@ -8,17 +8,19 @@
 You can replace the default ingress certificate for all
 applications under the `.apps` subdomain. After you replace
 the certificate, all applications, including the web console
-and CLI, will have encryption provided by specified certificate.
+and CLI, will have encryption provided by the specified certificate.
 
 .Prerequisites
 
 * You must have a wildcard certificate and its private key,
 both in the PEM format, for use.
 +
-[WARNING]
+[NOTE]
 ====
-The new certificate *must* be signed by a public certificate authority already
-included in the CA bundle as provided by the container userspace.
+This procedure assumes your replacement default ingress certificate is not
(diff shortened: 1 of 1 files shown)


codyhoag:
@danehans can you confirm this new note? It's a close copy from the warning we have [here](https://docs.openshift.com/container-platform/4.2/networking/ingress-operator.html#nw-ingress-setting-a-custom-default-certificate_configuring-ingress). Note this is for **4.2**. Thanks!

codyhoag:
@danehans can you review this new note? It's a close copy from the warning we have [here](https://docs.openshift.com/container-platform/4.2/networking/ingress-operator.html#nw-ingress-setting-a-custom-default-certificate_configuring-ingress). Note this is for **4.2**. Thanks!

modules/customize-certificates-replace-default-router.adoc
```diff
@@ -14,6 +14,13 @@ and CLI, will have encryption provided by specified certificate.
...
 both in the PEM format, for use.
++
+[WARNING]
+====
+The new certificate *must* be signed by a public certificate authority already
+included in the CA bundle as provided by the container userspace.
+====
+
 * The certificate must have a `subjectAltName` extension of
 `*.apps.<clustername>.<domain>`.
```
danehans: In essence, the wildcard cert SAN must match the cluster's ingress domain:
```
$ oc get ingress.config/cluster -o yaml
kind: Ingress
metadata:
  name: cluster
spec:
  domain: apps.<clustername>.<domain>
status: {}
```

The SAN for the above ingress domain would be `*.apps.<clustername>.<domain>`

modules/customize-certificates-replace-default-router.adoc
```diff
@@ -14,6 +14,13 @@ and CLI, will have encryption provided by specified certificate.
 
 * You must have a wildcard certificate and its private key,
 both in the PEM format, for use.
++
+[WARNING]
+====
+The new certificate *must* be signed by a public certificate authority already
+included in the CA bundle as provided by the container userspace.
```
danehans: The cert does not need to be signed by a public ca. The documented procedures are for adding a custom, i.e. non-public, ca. If the cert is signed by a public ca, then only the secret and ingresscontroller steps are required.

codyhoag: @danehans thank you for clarifying that. After a careful second reread, I see what you mean. Since this tripped up an external user and myself, it's probably helpful to make this overly obvious; I've modified this PR to add a note for pointing this fact out so it's not overlooked. Can you confirm?

codyhoag:
@danehans can you confirm the note I edited based on your prior feedback? Thanks!

codyhoag:
Since the original point that spurred this PR is no longer relevant (warning no longer exists [here](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.2/html/networking/configuring-ingress#nw-ingress-setting-a-custom-default-certificate_configuring-ingress) because using non-publicly signed CAs is available again using [this method](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.2/html/networking/configuring-a-custom-pki)), I'm closing this PR.

PR #19867, opened 2020-02-20
Routing

neal-timpe:
I’m happiest with the routing module and the assembly right now. The ingress and egress module need work, but I have a plan. When I get done the organization of the topics may change.

OSSMDOC-15 
OSSMDOC-27 
OSSMDOC-28 
OSSMDOC-30

This should be cherrypicked to the 4.3, 4.4, and 4.5 branch.

### DIFF
--- a/service_mesh/service_mesh_user_guide/ossm-traffic-manage.adoc
+++ b/service_mesh/service_mesh_user_guide/ossm-traffic-manage.adoc
@@ -1,0 +1,16 @@
+[id="ossm-routing-traffic"]
+= Traffic management
+include::modules/ossm-document-attributes.adoc[]
+:context: routing-traffic
+
+toc::[]
+
+You can control the flow of traffic and API calls between services in  {ProductName}. For example, some services in your service mesh may need to to communicate within the mesh and others may need to be hidden. Manage the traffic to hide specific backend services, expose services, create testing or versioning deployments, or add a security layer on a set of services. 
+
+This guide references the Bookinfo sample application to provide examples of routing in an example application. Install the xref:../../service_mesh/service_mesh_day_two/ossm-example-bookinfo.adoc#ossm-bookinfo-tutorial[Bookinfo application] to learn how these routing examples work. 
+
+include::modules/ossm-routing.adoc[leveloffset=+1]
+
+include::modules/ossm-routing-ingress.adoc[leveloffset=+1]
+
+include::modules/ossm-routing-bookinfo-example.adoc[leveloffset=+1]
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -1411,6 +1411,11 @@ Topics:
     File: ossm-tutorial-jaeger-tracing
   - Name: Automatic route creation
     File: ossm-auto-route
+- Name: Service mesh user guide
+  Dir: service_mesh_user_guide
+  Topics:
+  - Name: Traffic management
+    File: ossm-traffic-manage
 #  - Name: Grafana tutorial
 #    File: ossm-tutorial-grafana
 #  - Name: Prometheus tutorial
--- a/modules/ossm-routing-bookinfo-example.adoc
+++ b/modules/ossm-routing-bookinfo-example.adoc
@@ -1,0 +1,158 @@
+// Module included in the following assemblies:
+//
+// * service_mesh/service_mesh_user_guide/ossm-traffic-manage.adoc
+
+[id="ossm-routing-bookinfo_{context}"]
+= Routing example using the bookinfo application
+
+The {ProductShortName} Bookinfo sample application consists of four separate microservices, each with multiple versions. Three different versions, one of the microservices called `reviews`, have been deployed and are running concurrently.
+
+.Prerequisites:
+
+* Deploy the Bookinfo sample application to work with the following examples.
+
+.About this task
+
+To illustrate the problem this causes, access the bookinfo app `/product page` in a browser and refresh several times.
+
+Sometimes the book review output contains star ratings and other times it does not. Without an explicit default service version to route to, {ProductShortName} routes requests to all available versions one after the other.
+
+This tutorial helps you apply rules that route all traffic to `v1` (version 1) of the microservices. Later, you can apply a rule to route traffic based on the value of an HTTP request header.
+
+[id="ossm-routing-bookinfo-applying_{context}"]
+== Applying a virtual service
+
+To route to one version only, apply virtual services that set the default version for the micro-services. In the following example, the virtual service routes all traffic to `v1` of each micro-service
+
+1.  Run the following command to apply the virtual services:
+
+    $ oc apply -f https://raw.githubusercontent.com/Maistra/istio/maistra-1.1/samples/bookinfo/networking/virtual-service-all-v1.yaml
++
+1. To test the command was successful, display the defined routes with the following command:
++
+    $ oc get virtualservices -o yaml
++    
+That command returns the following YAML file. 
++
+[source,yaml]
+----
+apiVersion: networking.istio.io/v1alpha3
+kind: VirtualService
+metadata:
+  name: details
+  ...
+spec:
+  hosts:
+  - details
+  http:
+  - route:
+    - destination:
+      host: details
+      subset: v1
+---
+apiVersion: networking.istio.io/v1alpha3
+kind: VirtualService
+metadata:
+  name: productpage
+  ...
+spec:
+  gateways:
+  - bookinfo-gateway
+  - mesh
+  hosts:
+  - productpage
+  http:
+  - route:
+    - destination:
+      host: productpage
+      subset: v1
+---
+apiVersion: networking.istio.io/v1alpha3
+kind: VirtualService
+metadata:
+  name: ratings
+  ...
+spec:
+  hosts:
+  - ratings
+  http:
+  - route:
+    - destination:
+      host: ratings
+      subset: v1
+---
+apiVersion: networking.istio.io/v1alpha3
+kind: VirtualService
+metadata:
+  name: reviews
+  ...
+spec:
+  hosts:
+  - reviews
+  http:
+  - route:
+    - destination:
+      host: reviews
+      subset: v1
+----
++
+You have configured {ProductShortName} to route to the `v1` version of the Bookinfo microservices, most importantly the `reviews` service version 1.
+
+[id="ossm-routing-bookinfo-test_{context}"]
+== Test the new routing configuration
+
+You can easily test the new configuration by once again refreshing the `/productpage` of the Bookinfo app.
+
+1.  Open the Bookinfo site in your browser. The URL is `http://$GATEWAY_URL/productpage`, where `$GATEWAY_URL` is the External IP address of the ingress.
++
+The reviews part of the page displays with no rating stars, no matter how many times you refresh. This is because you configured {ProductShortName} to route all traffic for the reviews service to the version `reviews:v1` and this version of the service does not access the star ratings service.
++
+Your service mesh now routes traffic to one version of a service.
+
+[id="ossm-routing-bookinfo-route_{context}"]
+== Route based on user identity
+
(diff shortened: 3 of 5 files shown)


tvieira (commented):
swap `kubectl` with `oc` commands.

modules/ossm-uguide-bookinfo-example.adoc
```diff
@@ -0,0 +1,189 @@
...
+
+[id="ossm-userguide-routing-bookinfo-applying_{context}"]
+== Applying a virtual service
+
+To route to one version only, apply virtual services that set the default version for the micro-services. In the following example, the virtual service routes all traffic to `v1` of each micro-service
+
+1.  Run the following command to apply the virtual services:
+
+    ----
+    $ oc apply -f https://raw.githubusercontent.com/istio/istio/master/samples/bookinfo/networking/virtual-service-all-v1.yaml
```
jwendell: I think that instead of refer to Istio upstream (plus master branch) we should refer to maistra and the current branch (e.g. https://raw.githubusercontent.com/Maistra/istio/maistra-1.1/samples/bookinfo/networking/virtual-service-all-v1.yaml)

modules/ossm-uguide-routing.adoc
```diff
@@ -0,0 +1,264 @@
...
+The following example shows a possible gateway configuration for external HTTPS ingress traffic:
+
+----
+apiVersion: networking.istio.io/v1alpha3
+kind: Gateway
+metadata:
+  name: ext-host-gwy
+spec:
+  selector:
+    app: my-gateway-controller
```
jwendell: I know this is just an example, but if the user copies and pastes this code, it won't work as is, because this gateway controller does not exist out of the box.

So, I suggest changing this line to: `istio: ingressgateway # use istio default controller`, as seen here: https://github.com/istio/istio/blob/release-1.4/samples/bookinfo/networking/bookinfo-gateway.yaml#L7

modules/ossm-uguide-routing-ingress.adoc
```diff
@@ -0,0 +1,228 @@
...
+// * service_mesh/service_mesh_user_guide/ossm-traffic-manage.adoc
+
+[id="ossm-routing-ingress_{context}"]
+= Managing ingress traffic
+
+In a service mesh, you can use a {ProductName} gateway to enable {ProductShortName} features such as monitoring and route rules to be applied to traffic entering the cluster.
+
+You can configure {ProductShortName} to expose a service outside of the service mesh using an {ProductShortName} gateway.
+
+.Before you begin
```
jwendell: I don't think we need this `Before you begin` section. We can (1) assume OSSM is installed and (2) user doesn't need to change to istio directory.

modules/ossm-uguide-routing-ingress.adoc
```diff
@@ -0,0 +1,228 @@
...
+
+----
+$ export INGRESS_PORT=$(oc -n istio-system get service istio-ingressgateway -o jsonpath='{.spec.ports[?(@.name=="http2")].nodePort}')
+----
+
+----
+$ export SECURE_INGRESS_PORT=$(oc -n istio-system get service istio-ingressgateway -o jsonpath='{.spec.ports[?(@.name=="https")].nodePort}')
+----
+
+Setting the ingress IP depends on the cluster provider:
```
jwendell: Do we want to supply information for GKE (do we support it?), Minikube (we definitely don't support it), Docker for desktop (I'm pretty sure we don't support it)?

modules/ossm-uguide-routing-ingress.adoc
```diff
@@ -0,0 +1,228 @@
...
+1.  Create an {ProductShortName} `Gateway`:
+
+    $ oc apply -f - <<EOF
+    apiVersion: networking.istio.io/v1alpha3
+    kind: Gateway
+    metadata:
+      name: httpbin-gateway
+    spec:
+      selector:
+        istio: ingressgateway # use {ProductShortName} default gateway implementation
```
jwendell: This `{ProductShortName}` is not expanding, it's being rendered as raw `{ProductShortName}`. You can just cut it off.

modules/ossm-uguide-routing-ingress.adoc
```diff
@@ -0,0 +1,228 @@
...
+            host: httpbin
+    EOF
++
+Those commands create a virtual service configuration for the `httpbin` service containing two route rules that allow traffic for paths `/status` and `/delay`.
++
+The gateways list specifies that only requests through your `httpbin-gateway` are allowed. All other external requests are rejected with a 404 response.
++    
+Internal requests from other services in the mesh are not subject to these rules but instead will default to round-robin routing. To apply these rules to internal calls as well, you can add the special value `mesh` to the list of `gateways`. Since the internal hostname for the service is probabaly different, for example `httpbin.default.svc.cluster.local`, from the external one, you must also add it to the `hosts` list.
++
+1.  Access the `httpbin` service with cURL:
```
jwendell: You seem to be instructing the user to run and verify a command but this won't work because there were no previous instructions to deploy this httpbin application. This tutorial only tells the user to deploy the bookinfo one.

modules/ossm-uguide-bookinfo-example.adoc
```diff
@@ -0,0 +1,152 @@
...
+        - destination:
+            host: reviews
+            subset: v1
++
+You have configured {ProductShortName} to route to the `v1` version of the Bookinfo microservices, most importantly the `reviews` service version 1.
+
+[id="ossm-userguide-routing-bookinfo-test_{context}"]
+== Test the new routing configuration
+
+You can easily test the new configuration by once again refreshing the `/product page` of the Bookinfo app.
```
jwendell: `/productpage`

jwendell (changes requested):

modules/ossm-uguide-routing-ingress.adoc
neal-timpe: That's a good point. I'll work on that.

neal-timpe: I decided to remove that example rather than add the instructions for installing the httpbin sample. I think this pr is big enough as is. I can add the httpbin sample or other items in a later release.

jwendell (approved):

tvieira:
LGTM

tvieira (approved):

service_mesh/service_mesh_user_guide/ossm-traffic-manage.adoc
```diff
@@ -0,0 +1,16 @@
...
+
+toc::[]
+
+You can control the flow of traffic and API calls between services in  {ProductName}. For example, some services in your service mesh may need to to communicate within the mesh and others may need to be hidden. Manage the traffic to hide specific backend services, expose services, create testing or versioning deployments, or add a security layer on a set of services. 
+
+This guide references the Bookinfo sample application to provide examples of routing in an example application. Install the xref:../../service_mesh/service_mesh_day_two/ossm-example-bookinfo.adoc#ossm-bookinfo-tutorial[Bookinfo application] to learn how these routing examples work. 
+
+include::modules/ossm-uguide-routing.adoc[leveloffset=+1]
+
+include::modules/ossm-uguide-routing-ingress.adoc[leveloffset=+1]
```
JStickler, line 14: Could we put the Ingress topic before the Routing topic?

modules/ossm-uguide-routing-ingress.adoc
```diff
@@ -0,0 +1,67 @@
...
+
+That command returns the `NAME`, `TYPE`, `CLUSTER-IP`, `EXTERNAL-IP`,    `PORT(S)`, and `AGE` of each item in your namespace. 
+
+If the `EXTERNAL-IP` value is set, your environment has an external load balancer that you can use for the ingress gateway. 
+
+If the `EXTERNAL-IP` value is `<none>`, or perpetually `<pending>`, your environment does not provide an external load balancer for the ingress gateway. You can access the gateway using the service's https://kubernetes.io/docs/concepts/services-networking/service/#nodeport[node port].
+
+Choose the instructions for your environment:
+
+.External load balancer
```
JStickler, line 29: Configuring routing with a load balancer.

modules/ossm-uguide-routing-ingress.adoc
```diff
@@ -0,0 +1,67 @@
...
+
+In some environments, the load balancer may be exposed using a host name instead of an IP address. For that case, the ingress gateway's `EXTERNAL-IP` value is not be an IP address. Instead, it's a host name, and the previous command fails to set the `INGRESS_HOST` environment variable.
+
+Use the following command to correct the `INGRESS_HOST` value:
+
+----
+$ export INGRESS_HOST=$(oc -n istio-system get service istio-ingressgateway -o jsonpath='{.status.loadBalancer.ingress[0].hostname}')
+----
+
+.Node port
```
JStickler, line 55: Configuring routing without a load balancer.

modules/ossm-uguide-routing.adoc
```diff
@@ -0,0 +1,264 @@
...
+EOF
+----
+
+[id="ossm-routing-config-v-host_{context}"]
+== Configuring your virtual host
+
+The following sections explain each field in the YAML file and explain how you can create a virtual host in a virtual service.
+
+[id="ossm-routing-hosts_{context}"]
+=== The hosts field
```
JStickler, line 58: Configuring the hosts field

modules/ossm-uguide-routing.adoc
```diff
@@ -0,0 +1,264 @@
...
+
+The virtual service hostname can be an IP address, a DNS name, or, depending on the platform, a short name that resolves to a fully qualified domain name.
+
+----
+hosts:
+- reviews
+----
+
+[id="ossm-routing-routing-rules_{context}"]
+=== Routing rules
```
JStickler, line 70: Configuring routing rules in virtual services

modules/ossm-uguide-routing.adoc
```diff
@@ -0,0 +1,264 @@
...
+[id="ossm-routing-lb_{context}"]
+==== Load balancing options 
+
+By default, {ProductName} uses a round-robin load balancing policy, where each service instance in the instance pool gets a request in turn. {ProductName} also supports the following models, which you can specify in destination rules for requests to a particular service or service subset.
+
+* Random: Requests are forwarded at random to instances in the pool.
+* Weighted: Requests are forwarded to instances in the pool according to a specific percentage.
+* Least requests: Requests are forwarded to instances with the least number of requests.
+
+.Destination example
```
JStickler, line 110: Destination rule example (because it's not just the destination)

modules/ossm-uguide-routing-ingress.adoc
```diff
@@ -0,0 +1,67 @@
+// Module included in the following assemblies:
+//
+// * service_mesh/service_mesh_user_guide/ossm-traffic-manage.adoc
+
+[id="ossm-routing-ingress_{context}"]
+= Managing ingress traffic
+
+In a service mesh, you can use a {ProductName} gateway to enable {ProductShortName} features such as monitoring and route rules to be applied to traffic entering the cluster.
```
JStickler, line 8: Suggestion
In Red Hat OpenShift Service Mesh, you use the Ingress Gateway to enable Service Mesh features such as monitoring, security, and route rules to be applied to traffic entering the cluster.  Traffic coming into the mesh must come in via the Ingress Gateway.
You configure Service Mesh to expose a service outside of the service mesh using an Service Mesh gateway.

neal-timpe: >Traffic coming into the mesh must come in via the Ingress Gateway.

I think this is true, but I remember now talking to Rob. Traffic does come into the mesh through the Ingress gateway, but it does that by default. Configuring the service outside the service mesh is a non-default special case and that's why it's lower down. It's only used by a subset of users.

I'm making the second paragraph part of the first to make that more clear.

JStickler (approved):
LGTM

PR #19886, opened 2020-02-20
VMware vSphere installation and Time configuration

jcpowermac:
Recently there has been an increase of support cases
directly related to ESXi time misconfiguration.
This PR addresses that issue providing a notice
of the potential problem and a link to VMware documentation
to resolve.

### DIFF
--- a/modules/installation-vsphere-infrastructure.adoc
+++ b/modules/installation-vsphere-infrastructure.adoc
@@ -24,3 +24,8 @@ clusters are subject to the following restrictions:
 
 If you use a vSphere version 6.5 instance, consider upgrading to 6.7U2 before
 you install {product-title}.
+
+[IMPORTANT]
+====
+You must ensure that the time on your ESXi hosts is synchronized before you install {product-title}. See link:https://docs.vmware.com/en/VMware-vSphere/6.7/com.vmware.vsphere.vcenterhost.doc/GUID-8756D419-A878-4AE0-9183-C6D5A91A8FB1.html[Edit Time Configuration for a Host] in the VMware documentation.
+====


jcpowermac:
@kalexand-rh not sure this is enough information or if you feel this is in the correct location.
Please let me know and I will make changes.  Thanks!

modules/installation-vsphere-infrastructure.adoc
```diff
@@ -24,3 +24,9 @@ clusters are subject to the following restrictions:
 
 If you use a vSphere version 6.5 instance, consider upgrading to 6.7U2 before
 you install {product-title}.
+
+[IMPORTANT]
+====
+Time configuration of ESXi hosts is critical for a successful installation of
+OpenShift Container Platform.  See link:https://docs.vmware.com/en/VMware-vSphere/6.7/com.vmware.vsphere.vcenterhost.doc/GUID-8756D419-A878-4AE0-9183-C6D5A91A8FB1.html[Edit Time Configuration for a Host] in the VMware documentation.
```
kalexand-rh:
```suggestion
{product-title}. See link:https://docs.vmware.com/en/VMware-vSphere/6.7/com.vmware.vsphere.vcenterhost.doc/GUID-8756D419-A878-4AE0-9183-C6D5A91A8FB1.html[Edit Time Configuration for a Host] in the VMware documentation.
```

modules/installation-vsphere-infrastructure.adoc
```diff
@@ -24,3 +24,9 @@ clusters are subject to the following restrictions:
 
 If you use a vSphere version 6.5 instance, consider upgrading to 6.7U2 before
 you install {product-title}.
+
+[IMPORTANT]
+====
+Time configuration of ESXi hosts is critical for a successful installation of
```
kalexand-rh: Do the hosts just need to be in sync, or is it better to enable the NTP client? I might say "You must ensure that the time on your ESXi hosts is synchronized before you install {product-title}."

kalexand-rh (commented):
I have a question, and there's a variable for the product name.

Which versions is this change valid for?

modules/installation-vsphere-infrastructure.adoc
jcpowermac: @kalexand-rh 
Sorry I didn't see the comments until now.  I will fix the {product-title}
It should be all supported versions.

modules/installation-vsphere-infrastructure.adoc
jcpowermac: @kalexand-rh I think what you wrote sounds good.  I will update.
In regards to your other question I think NTP should always be enabled but I am sure there is some enivornment where that is not possible and will need to manually set.  As long as the time between the ESXi hosts and the machine that is running `openshift-installer` are in sync there shouldn't be a problem.

kalexand-rh:
Thank you @jcpowermac! @jiajliu, will you PTAL and confirm that this change sounds reasonable for 4.1+?

jiajliu:
LGTM. I think it's about vmware time configuration, so it's ok for all ocp versions.

kalexand-rh:
Thank you both! Merging.

PR #19911, opened 2020-02-21
BZ1797036 Add optional step for configuring NFS PV in reg storage

bobfuru:
[BZ1797036](https://bugzilla.redhat.com/show_bug.cgi?id=1797036)
4.2+

[PREVIEW BUILD LINK](https://bz1797036--ocpdocs.netlify.com/openshift-enterprise/latest/installing/installing_vsphere/installing-restricted-networks-vsphere.html#registry-configuring-storage-vsphere_installing-restricted-networks-vsphere)

It really only makes sense to include this optional step after step 3 ("Check the registry configuration"). The customer request was: 

> We'd like to have kept the SC (`thin`) as the default class (storageclass.kubernetes.io/is-default-class=true) and are going to add a non-default SC (` nfs01`), and set it as new image-registry, .
> In this case, how do we edit PVC ?
> The reason why we would not like to change default SC, is that if the default class changes, and then the image registry should need to be rebuilt due to that. That meant changing the default SC after the service lanuched, and it will have negative impact to existing users

### DIFF
--- a/modules/registry-configuring-storage-vsphere.adoc
+++ b/modules/registry-configuring-storage-vsphere.adoc
@@ -64,6 +64,77 @@ storage:
 Leave the `claim` field blank to allow the automatic creation of an
 `image-registry-storage` PVC.
 
+. Optional: Add a new storage class to a PV:
+.. Create the PV:
++
+----
+$ oc create -f -
+----
++
+[source,yaml]
+----
+
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: image-registry-pv
+spec:
+  accessModes:
+    ReadWriteMany
+    capacity:
+      storage: 100Gi
+  nfs:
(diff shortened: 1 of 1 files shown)


bobfuru:
@chao007 PTAL

bobfuru:
@bmcelvee or @kalexand-rh Could one of you recommend a Dev SME who could take a quick look at this?

chao007:
LGTM

bobfuru:
Adds an optional step 4 for adding NFS storage class. 
Preview build: https://bz1797036--ocpdocs.netlify.com/openshift-enterprise/latest/installing/installing_vsphere/installing-restricted-networks-vsphere.html#registry-configuring-storage-vsphere_installing-restricted-networks-vsphere

lamek:
LGTM.

PR #19929, opened 2020-02-24
[enterprise-4.2] updated to unset variables

vikram-redhat:
https://github.com/openshift/openshift-docs/pull/19841


vikram-redhat:
@apinnick FYI 

Do you know if this should be CPed to 4.1 as well?

apinnick:
@vikram-redhat No, it doesn't look like it needs to be CPed to 4.1.

PR #19937, opened 2020-02-24
removed xref from dedicated module

sheriff-rh:
Swapped an `xref` for a `link`. Preview build here: 
http://file.rdu.redhat.com/~antaylor/xrefremoval/xrefremoval/getting_started/accessing-your-services.html

@vikram-redhat  PTAL.

### DIFF
--- a/modules/dedicated-creating-your-cluster.adoc
+++ b/modules/dedicated-creating-your-cluster.adoc
@@ -26,7 +26,6 @@ following are the default ranges available to use:
 .. Pod CIDR: 10.128.0.0/14
 
 . Add your Identity provider by clicking the *Add OAuth Configuration* link.
-Further configuration details may be found xref:../authentication/dedicated-understanding-authentication.adoc[here].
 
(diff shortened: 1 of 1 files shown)


modules/dedicated-creating-your-cluster.adoc
```diff
@@ -7,7 +7,7 @@
 
 To create your {product-title} cluster:
 
-. Log in to link:https://cloud.redhat.com/openshift[cloud.redhat.com/openshift].
+. Log in to *https://cloud.redhat.com/openshift*.
```
bergerhoffer: I don't know if we have it documented, but I do think it's a recommended practice to use the `link:` prefix for external links. One reason I can think of is that if you want an easy way to find links (for example, for broken link checking, etc.) the `link:` is an easy way to find them.

If you look at other places that link to cloud.redhat.com/openshift in our docs, they do use `link:`, and apparently there's an attribute of `cloud-redhat-com` that can be used for the title of it (Red Hat OpenShift Cluster Manager).

bergerhoffer (commented):
Just one suggestion

PR #16884, opened 2019-09-25
Add Using Imagestreams with Kubernetes Resources assembly

bmcelvee:
Per slack request

PR #19946, opened 2020-02-24
New Add Using Imagestreams with Kubernetes Resources assembly

bmcelvee:
This PR will likely replace https://github.com/openshift/openshift-docs/pull/16884.

### DIFF
--- a/modules/images-managing-images-enabling-imagestreams-kubernetes.adoc
+++ b/modules/images-managing-images-enabling-imagestreams-kubernetes.adoc
@@ -1,0 +1,100 @@
+// Module included in the following assemblies:
+//
+// * openshift_images/managing-images/using-imagestreams-with-kubernetes-resources.adoc
+
+
+[id="images-managing-images-enabling-imagestreams-kubernetes_{context}"]
+= Enabling imagestreams with Kubernetes resources
+
+When using imagestreams with Kubernetes resources, you can only reference imagestreams that reside in the
+same project as the resource. The imagestream reference must consist of a single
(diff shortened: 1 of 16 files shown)


bmcelvee:
@vikram-redhat would you mind helping me troubleshoot this build error? I brought content over from on older PR in the hopes of replacing that one, but I'm still having build errors. 

I made some minor organizational changes, and corrected several issues with an existing related file.

modules/images-managing-images-enabling-imagestreams-kubernetes.adoc
```diff
@@ -0,0 +1,100 @@
...
++
+When enabled, the behavior is enabled for all tags within the imagestream.
+
+. Then you can query the imagestreams and see if the option is set:
++
+----
+$ oc set image-lookup imagestream --list
+----
+
+////
```
bmcelvee: Will address commented out content and fix numbering

vikram-redhat:
Hey @bmcelvee - because you are moving some of these files from their original locations, you will need to do a search through the rest of the docs that reference these files. The current error is simply saying that because you have moved the ../openshift_images/image-streams-manage.adoc file to ../openshift_images/managing_imagestreams/image-streams-manage.adoc a reference to that file from the builds book is broken: https://github.com/openshift/openshift-docs/blame/master/builds/running-entitled-builds.adoc#L35

To fix the build, you will need to sweep and change these references at all locations other than the files you are touching. Also, reminder to add these in the redirects register so that when this goes live, I can setup the redirects.

PR #19958, opened 2020-02-24
Remove xref from module

bmcelvee:


bmcelvee:
Removed xref and updated assembly usage. No reviews needed.

PR #20007, opened 2020-02-25
BZ#1807121 - NAD requires a 'name' for the configuration

aburdenthehand:
Minor change to the NAD configuration to add a 'name' field and annotation (shuffling the others down).

I checked on our 2.2 environment and this also has an impact there.

cherrypick to enterprise-4.3 and enterprise-4.4


alexxa (approved):
Looks good to me. Thanks.

phoracek (changes requested):
We will need the same fix for the PXE docs.

phoracek (approved):
I was too quick to comment. PXE is fine already.

Issue #19820, opened 2020-02-19
[enterprise-4.2] Edit suggested in file logging/config/cluster-logging-external.adoc


aelicieguo:
### Which section(s) is the issue in?
Configuring log collector to send logs to an external syslog server

There are no info on configuration for sending logs to external syslog server with UDP

### What needs fixing?
add REMOTE_SYSLOG_TYPE

defaults to syslog_buffered, declare to syslog for sending messages with UDP

mburke5678:
Thanks for pointing this out. The parameter was added to the code recently and the docs were not updated. Follow my changes in https://github.com/openshift/openshift-docs/pull/20008

PR #20008, opened 2020-02-25
Issues: Edit suggested in file logging/config/cluster-logging-external.adoc

mburke5678:
https://github.com/openshift/openshift-docs/issues/19820

This is for 4.1 and 4.2 only. Not for master or 4.3.


mburke5678:
Used wording from https://bugzilla.redhat.com/show_bug.cgi?id=1780298 verified by @anpingli 
Anping, can you please make sure that this parameter is available in 4.1 and 4.2?
Is this the code: https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/generate_syslog_config.rb#L27
Added by @nhosoi in https://github.com/openshift/origin-aggregated-logging/pull/1659/files

PR #20031, opened 2020-02-26
Fix YAML examples and Customizing CLI modules

dmvolod:


dmvolod:
@ahardin-rh , could you please to take a look at this. Thank you.

Issue #19919, opened 2020-02-21
[enterprise-4.3] Edit suggested in file networking/accessing-hosts.adoc


peterclauterbach:
### Which section(s) is the issue in?
Title
Accessing hosts on Amazon Web Services (AWS) after a full-stack automation (IPI) installation

### What needs fixing?
My understanding is that we do not use IPI or UPI acronyms in external docs.

jboxman:
@kalexand-rh, is it _installer-provisioned_ now?

kalexand-rh:
@jboxman, _installer-provisioned_ is correct. IPI and UPI should not be used anywhere in the collection.

codyhoag:
Fixed in #20038

PR #20038, opened 2020-02-26
Always installer-provisioned

jboxman:
- https://github.com/openshift/openshift-docs/issues/19919

### DIFF
--- a/modules/accessing-hosts-on-aws.adoc
+++ b/modules/accessing-hosts-on-aws.adoc
@@ -3,7 +3,7 @@
 // * networking/accessing-hosts.adoc
 
 [id="accessing-hosts-on-aws_{context}"]
-= Accessing hosts on Amazon Web Services (AWS) after a full-stack automation (IPI) installation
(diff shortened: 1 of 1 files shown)


jboxman:
@kalexand-rh, how's this? It reads kind of strange with _installer installation_.

modules/accessing-hosts-on-aws.adoc
```diff
@@ -3,7 +3,7 @@
 // * networking/accessing-hosts.adoc
 
 [id="accessing-hosts-on-aws_{context}"]
-= Accessing hosts on Amazon Web Services (AWS) after a full-stack automation (IPI) installation
+= Accessing hosts on Amazon Web Services (AWS) after a full-stack automation installer-provisioned installation
```
kalexand-rh: I might say "Accessing hosts on Amazon Web Services in an installer-provisioned infrastructure cluster"

I don't see AWS in the body of the module, so you don't need to define the initialism.

PR #20053, opened 2020-02-26
OSDOCS-893: Updating DR procedures for the etcd cluster operator

bergerhoffer:
Preview: https://osdocs-893-etcd-op--ocpdocs.netlify.com/openshift-enterprise/latest/backup_and_restore/disaster_recovery/scenario-2-restoring-cluster-state.html


modules/backup-etcd.adoc
```diff
@@ -21,6 +21,15 @@ You should only save a backup from a single master host. You do not need a backu
...
 ----
 $ sudo /usr/local/bin/etcd-snapshot-backup.sh ./assets/backup
+1bf371f1b5a483927cd01bb593b0e12cff406eb8d7d0acf4ab079c36a0abd3f7
+etcdctl version: 3.3.18
+API version: 3.3
+found latest kube-apiserver-pod: /etc/kubernetes/static-pod-resources/kube-apiserver-pod-7
+found latest kube-controller-manager-pod: /etc/kubernetes/static-pod-resources/kube-controller-manager-pod-8
+found latest kube-scheduler-pod: /etc/kubernetes/static-pod-resources/kube-scheduler-pod-6
+found latest etcd-pod: /etc/kubernetes/static-pod-resources/etcd-pod-2
+Snapshot saved at /var/home/core/assets/backup/snapshot_2020-03-18_220218.db
```
bergerhoffer: @retroflexer I replaced the instances of `myback99` with `assets/backup` in this output, as that's what we use in this example command. Please make sure that this is correct.

modules/backup-etcd.adoc
```diff
@@ -21,6 +21,15 @@ You should only save a backup from a single master host. You do not need a backu
 +
 ----
 $ sudo /usr/local/bin/etcd-snapshot-backup.sh ./assets/backup
```
bergerhoffer: @retroflexer We say in step 1 to access the host as the root user, then proceed to use `sudo` in this command. Should I just remote "as the root user" from step 1 and keep `sudo` in this command? We do use `sudo` for most of the restore procedures so it is probably best to keep that if that works.

c1505:
The current 4.3 docs are not working for me for restore https://docs.openshift.com/container-platform/4.3/backup_and_restore/disaster_recovery/scenario-2-restoring-cluster-state.html#dr-restoring-cluster-state.  Is this the way it should be for 4.3 too or is this just for 4.4 and forward?

retroflexer:
No. The recovery for 4.3 is entirely different than the recovery for 4.4.
In 4.4, we take advantage of the presence of cluster-etcd-operator, which
is not present in 4.3.

retroflexer:
No. The recovery for 4.3 is entirely different than the recovery for 4.4. In 4.4, we take advantage of the presence of cluster-etcd-operator, which is not present in 4.3.

modules/backup-etcd.adoc
retroflexer: Yes. We don't want them to access the system as root. That's why we include sudo in the commands assuming they are running as non-root user.

modules/backup-etcd.adoc
retroflexer: OK. That's fine. Keep it consistent.

bergerhoffer:

https://osdocs-893-etcd-op--ocpdocs.netlify.com/openshift-enterprise/latest/backup_and_restore/disaster_recovery/scenario-2-restoring-cluster-state.html

bergerhoffer:
@geliu2016 Would you be able to review these updates, and run through testing of a restore using this procedure? Thank you!

https://osdocs-893-etcd-op--ocpdocs.netlify.com/openshift-enterprise/latest/backup_and_restore/disaster_recovery/scenario-2-restoring-cluster-state.html

bmcelvee:
LGTM! 

Just a note - there's one instance of "{product-title} 4.4" instead of "{product-title} {product-version}", but it shouldn't be an issue since it's referring to 4.4 specifically.

bergerhoffer:
Thanks for the review @bmcelvee! Yeah in that case, it needs to hard code 4.4 here, because if carried forward to 4.5, that would be untrue to say "As of OCP 4.5". But hopefully in 4.5 we'll just be yanking that section anyways :).

modules/dr-restoring-cluster-state.adoc
```diff
@@ -5,117 +5,218 @@
...
 
-.Procedure
+.. Access a master host that is not the recovery host.
 
-. Prepare each master host in your cluster to be restored.
+.. Move the existing etcd Pod file out of the Kubelet manifest directory:
 +
-You should run the restore script on all of your master hosts within a short period of time so that the cluster members come up at about the same time and form a quorum. For this reason, it is recommended to stage each master host in a separate terminal, so that the restore script can then be started quickly on each.
+----
+[core@ip-10-0-154-194 ~]$ sudo mv /etc/kubernetes/manifests/etcd.yaml /tmp
```
geliu2016, line 45: The exact file is /etc/kubernetes/manifests/etcd-pod.yaml.

modules/dr-restoring-cluster-state.adoc
```diff
@@ -5,117 +5,218 @@
...
+[core@ip-10-0-154-194 ~]$ sudo crictl ps | grep etcd
+----
++
+The output of this command should be empty.
 
-.. Access the master host.
+.. Move the existing Kubernetes API server Pod file out of the Kubelet manifest directory:
++
+----
+[core@ip-10-0-154-194 ~]$ sudo mv /etc/kubernetes/manifests/kube-apiserver.yaml /tmp
```
geliu2016, line 59: The exact file is: /etc/kubernetes/manifests/kube-apiserver-pod.yaml

geliu2016:
> @geliu2016 Would you be able to review these updates, and run through testing of a restore using this procedure? Thank you!
> 
> https://osdocs-893-etcd-op--ocpdocs.netlify.com/openshift-enterprise/latest/backup_and_restore/disaster_recovery/scenario-2-restoring-cluster-state.html

Reviewed and run through all steps, there are two file name err have been marked in comments, the executed result is pass without err. thx

geliu2016:
@bergerhoffer Reviewed and run through all steps, there are two file name err have been marked in comments, the executed result is pass without err. thx

bergerhoffer:
Thanks for the review @geliu2016! I've updated those file names. Let me know whether this looks okay now or if you have any other feedback!

@retroflexer Let me know if there is anything else outstanding for this procedure, or if we're good to merge this in. Thanks!

retroflexer:
Looks good to me. Thanks for the good work, @bergerhoffer.

geliu2016:
LGTM，thx

PR #20063, opened 2020-02-27
added webscale

vikram-redhat:


vikram-redhat:
@ahardin-rh I have added webscale distro. Once this is merged, we will have the option to conditionalize content with the openshift-webscale tag. We will need to edit the topic_map for inclusion/exclusion of content as the case might be. Please work with John and Stephen to help them along.

PR #20065, opened 2020-02-27
use a working example as node selectors

duritong:
In the section how to create infra nodes, we set the node label
to the following: node-role.kubernetes.io/infra: ""

Please backbort this to other versions


duritong:
Yeah this was merged with #21486

PR #20088, opened 2020-02-27
Clarifying statement

bergerhoffer:
This is the update for 4.4. Will backport as far as possible but am guessing previous releases will touch different files and need separate PRs.

This PR is a result of feedback that the sentences saying "compute, or worker, machines" is confusing.

### DIFF
--- a/machine_management/more-rhel-compute.adoc
+++ b/machine_management/more-rhel-compute.adoc
@@ -6,7 +6,7 @@ include::modules/common-attributes.adoc[]
 toc::[]
 
 If your {product-title} cluster already includes Red Hat Enterprise Linux (RHEL)
-compute, or worker, machines, you can add more RHEL compute machines to it.
+compute machines, which are also known as worker machines, you can add more RHEL compute machines to it.
 
 include::modules/rhel-compute-overview.adoc[leveloffset=+1]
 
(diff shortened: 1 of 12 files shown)


kalexand-rh:
@abhinavdahiya, PTAL?

mburke5678:
@bergerhoffer LGTM

machine_management/more-rhel-compute.adoc
```diff
@@ -6,7 +6,7 @@ include::modules/common-attributes.adoc[]
 toc::[]
 
 If your {product-title} cluster already includes Red Hat Enterprise Linux (RHEL)
-compute, or worker, machines, you can add more RHEL compute machines to it.
+worker machines, you can add more RHEL compute machines to it.
```
abhinavdahiya: i was under the assumption that we were moving away from `worker` and more towards `compute` ?

bergerhoffer: Thanks for the feedback @abhinavdahiya. Instead of choosing worker over compute, I've instead changed the wording to be clearer (which was the intent of this PR). Let me know if these changes look okay to you - especially the one the one here [1] which is slightly different in that it mentions replicas.

[1] https://github.com/openshift/openshift-docs/pull/20088/files#diff-b81c8ab7e12a1648ebab26d2f023f928R61-R62

mburke5678:
@bergerhoffer LGTM

bergerhoffer:
Merging to get this fix in for Marco.

PR #20089, opened 2020-02-27
bz1744543 amended etcd restoration steps

sheriff-rh:
Removed all etcd v2 references to avoid confusion. Clarified several steps per BZ below. 

Bug: 
https://bugzilla.redhat.com/show_bug.cgi?id=1744543

Build preview: 
http://file.rdu.redhat.com/~antaylor/bz1744543/day_two_guide/host_level_tasks.html#day-two-guide-etcd-backup

### DIFF
--- a/day_two_guide/topics/con_scaling-etcd.adoc
+++ b/day_two_guide/topics/con_scaling-etcd.adoc
@@ -19,7 +19,7 @@ number of members.
 
 Having a cluster with an odd number of etcd hosts can account for fault
 tolerance. Having an odd number of etcd hosts does not change the number needed
-for a quorum but increases the tolerance for failure. For example, with a 
+for a quorum but increases the tolerance for failure. For example, with a
 cluster of three members, quorum is two, which leaves a failure tolerance of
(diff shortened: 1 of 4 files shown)


geliu2016:
LGTM, thanks

sburke1967:
Hi Andrew,
Step 6 of Restoring etcd data should be removed as you do not repeat this process on all the etc nodes, this is only performed on the first one to re-create a single mode cluster: http://file.rdu.redhat.com/~antaylor/bz1744543/day_two_guide/host_level_tasks.html#restoring-etcd_deprecating-etcd

Step 10 - After the first instance is running, you can add the remaining peers back into the cluster.
This should be the process for re-adding the remaining nodes back into the cluster, not just stating that it can now be done, as per https://docs.openshift.com/container-platform/3.11/admin_guide/assembly_restore-etcd-quorum.html#adding-etcd-after-restoring_restore-etcd-quorum

Cheers

Sean

sheriff-rh:
The build preview has been updated to reflect the latest changes. Thanks @sburke1967 !

PR #20095, opened 2020-02-28
Revert "Edit suggested in file nodes/scheduling/nodes-scheduler-default.ado"

mburke5678:
Reverts openshift/openshift-docs#20094


mburke5678:
Merged from local master-3 in error.

PR #20108, opened 2020-02-28
Edits to log forwarding per email

mburke5678:
Per email and Slack: https://coreos.slack.com/archives/GGUR75P60/p1582896919255000

"I saw a mistake in the yaml file. Type and name should be switched. This is also wrong in the example in the docs. "

### DIFF
--- a/modules/cluster-logging-log-forwarding-configure.adoc
+++ b/modules/cluster-logging-log-forwarding-configure.adoc
@@ -11,7 +11,7 @@ If you enable Log Forwarding, you should define a pipeline all for three source 
 
 .Procedure
 
-To configure the Log Forwarding feature:
+To configure the log forwarding feature:
 
 . Edit the Cluster Logging Custom Resource (CR) in the `openshift-logging` project:
 +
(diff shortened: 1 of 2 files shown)


mburke5678:
@sichvoge @jcantrill  PTAL
I flipped the `name` and `type` parameters in the log forwarding CR, as I think the customer, Dennis Hergaarden, was suggesting. 
I also saw in his email that he needed to disable log forwarding and delete the log forward CR in order to make edits in the CR file. I added the steps here too.

mburke5678:
@sichvoge @jcantrill PTAL

modules/cluster-logging-log-forwarding-configure.adoc
```diff
@@ -96,12 +96,72 @@ spec:
...
-
++
 [NOTE]
 ====
 If you set `disableDefaultForwarding: true` you must configure a pipeline and output for all three types of logs, application, infrastructure, and audit. If you do not specify a pipeline and output for a log type, those logs are not stored and will be lost.
 ==== 
 
+== Updating the Log Forwarding Custom Resource
+
+If you make changes to the log forwarding outputs or pipelines, you need to disable log forwarding, edit the CR, then enable log forwarding again.
```
jcantrill: This really should be unnecessary.  If you have to do this then there is a bug in the code.

anpingli (commented):
https://secureforward.offcluster.com:9200  --> https://secureforward.offcluster.com: 24224.   (Most of time, that is a fluentd receiver. The default port is 24224)

anpingli (commented):
The other part LGTM

modules/cluster-logging-log-forwarding-configure.adoc
anpingli: @mburke5678  Yes, It is unnecessary.  The change can be apply without disbale log forwarding. if the forward didn't take effect,  we will filed a bug to trace it.

mburke5678: @anpingli Thanks. I think I misunderstood a comment you made in Slack. I will follow up with https://github.com/openshift/openshift-docs/pull/20675 to remove this section,

Issue #19640, opened 2020-02-10
[enterprise-4.2] Edit suggested in file web-console/customizing-the-web-console.adoc


tdaleibm:
### Which section(s) is the issue in?
Page: OCP 4.2 Customizing the web console in OpenShift Container Platform. Section: Adding a custom logo and product name
https://docs.openshift.com/container-platform/4.2/web-console/customizing-the-web-console.html

### What needs fixing?
Yaml example indentation in step 2 is off on line "customProductName: My Console" and subsequent lines

jboxman:
@tdaleibm, thanks for reporting this!

PR #20119, opened 2020-02-28
Fix indentation

jboxman:
- https://github.com/openshift/openshift-docs/issues/19640


jboxman:
@ahardin-rh, I think this is correct for 4.2.

PR #20155, opened 2020-03-02
Update SR-IOV marketplace channel jsonpath

jboxman:
- https://github.com/openshift/openshift-docs/pull/20060

### DIFF
--- a/modules/nw-sriov-installing-operator.adoc
+++ b/modules/nw-sriov-installing-operator.adoc
@@ -65,9 +65,9 @@ $ oc create -f sriov-operatorgroup.yaml
 step.
 +
 ----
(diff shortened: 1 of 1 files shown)


jboxman:
@zshi-redhat, does this look correct? Thanks!

jboxman:
@zshi-redhat, I think this also needs to be merged?

zshi-redhat:
> @zshi-redhat, I think this also needs to be merged?

@jboxman hi sorry for the late reply. Yes this needs to be merged in release-4.2 branch.

jboxman:
@zshi-redhat, cool.

I didn't for this, but at some point we can probably replace every instance of _4.x_ with `{product-version}`, which always yields the major and minor version of OpenShift for each branch.

PR #20228, opened 2020-03-04
Remove incorrect prefix from ShiftStack params

maxwelldb:
For [BZ 1804076](https://bugzilla.redhat.com/show_bug.cgi?id=1804076)


maxwelldb:
@openshift/team-documentation Can someone take a look at this when they have a sec? Super small.

bergerhoffer (commented):
The actual change looks okay to me, but I assume that this change needs to be made not just in 4.2, but in 4.2, 4.3, and 4.4?

If so, then the PR should be made against master, and then cherry picked to 4.4, 4.3, and 4.2.

maxwelldb:
That's what I get for working sick. I'll recreate this against master.

Issue #19992, opened 2020-02-25
[enterprise-4.3] Edit suggested in file installing/install_config/configuring-private-cluster.adoc


seferovic:
### Which section(s) is the issue in?

After you install OpenShift Container Platform version 4.2 cluster, you can set some of its core components to be private.

### What needs fixing?

Version number 4.2 -> 4.3

codyhoag:
Thanks! Fixed and merged here: #20229. This should be published soon.

PR #20229, opened 2020-03-04
Use token for version

codyhoag:
Suggested in #19992

### DIFF
--- a/installing/install_config/configuring-private-cluster.adoc
+++ b/installing/install_config/configuring-private-cluster.adoc
@@ -5,7 +5,7 @@ include::modules/common-attributes.adoc[]
 
 toc::[]
 
-After you install {product-title} version 4.2 cluster, you can set some of its core components to be private.
+After you install an {product-title} version {product-version} cluster, you can set some of its core components to be private.
 
 [IMPORTANT]
 ====


installing/install_config/configuring-private-cluster.adoc
```diff
@@ -5,7 +5,7 @@ include::modules/common-attributes.adoc[]
 
 toc::[]
 
-After you install {product-title} version 4.2 cluster, you can set some of its core components to be private.
+After you install {product-title} version {product-version} cluster, you can set some of its core components to be private.
```
bergerhoffer: The replace is fine. And I know it was like this before that, but it reads weird if you want to fix it while you're in there.

I'd either put "an" in front of `{product-title} `, or remove "cluster". So it either says:

```
After you install an OpenShift Container Platform version 4.4 cluster,
or
After you install OpenShift Container Platform version 4.4,
```

bergerhoffer (commented):
Your update LGTM. But added a suggestion for an improvement on the existing text.

installing/install_config/configuring-private-cluster.adoc
codyhoag: Yep! Better to improve it while I'm in there. I'll fix the flow. Thanks!

bergerhoffer: Thanks @codyhoag!

Issue #19599, opened 2020-02-07
[enterprise-3.11] Edit suggested in file dev_guide/environment_variables.adoc


syedriko:
### Which section(s) is the issue in?
 Developer Guide/Managing Environment Variables 

### What needs fixing?
finds environment variables ... whose names begin with RAILS_ ...
$ env | grep RAILS_ | oc set env rc/r1 -e -

The regexp needs a ^ to match names that begin with RAILS_ and not just contain it: grep ^RAILS_

codyhoag:
Fixed and merged in #20245. This should be published soon. Thanks!

PR #20245, opened 2020-03-04
Proper regexp for variables that *begin* with RAILS_

codyhoag:
Suggested in #19599


codyhoag:
@wewang58 PTAL? Thanks!

codyhoag:
Due to this being a one char regex addition that passed QE, a peer review isn't necessary.

PR #20271, opened 2020-03-05
BZ#1797841 Note for recommended master node usage

codyhoag:
https://bugzilla.redhat.com/show_bug.cgi?id=1797841

### DIFF
--- a/modules/architecture-machine-roles.adoc
+++ b/modules/architecture-machine-roles.adoc
@@ -44,11 +44,18 @@ control the Kubernetes cluster. In {product-title}, the master machines are
 the control plane. They contain more
 than just the Kubernetes services for managing the {product-title} cluster.
 Because all of the machines with the control plane role are master machines,
-and the terms "master" and "control plane" are used interchangeably to describe
+the terms _master_ and _control plane_ are used interchangeably to describe
 them. Instead of being grouped into a
 MachineSet, master machines are defined by a series of standalone machine API
 resources. Extra controls apply to master machines to prevent you from deleting
 all master machines and breaking your cluster.
+
+[NOTE]
+====
+Use three master nodes. Although you can theoretically
+use any number of master nodes, the number is constrained by etcd quorum due to
+master static Pods and etcd static Pods working on the same hosts.
+====
 
 Services that fall under the Kubernetes category on the master include the
 API server, etcd, controller manager server, and HAProxy services.


codyhoag:
@xingxingxia PTAL? Thanks!

xingxingxia:
@codyhoag Good job! LGTM Thank you!

modules/architecture-machine-roles.adoc
```diff
@@ -44,12 +44,19 @@ control the Kubernetes cluster. In {product-title}, the master machines are
...
-and the terms "master" and "control plane" are used interchangeably to describe
+the terms _master_ and _control plane_ are used interchangeably to describe
 them. Instead of being grouped into a
 MachineSet, master machines are defined by a series of standalone machine API
 resources. Extra controls apply to master machines to prevent you from deleting
 all master machines and breaking your cluster.
 
+[NOTE]
+====
+It is recommended to only use three master nodes. Although you can theoretically
```
xingxingxia: > It is recommended to only use three master nodes
Adding `only` is not suggested. Better to remove the word `only`, because users may want to create large size clusters.

modules/architecture-machine-roles.adoc
```diff
@@ -44,12 +44,19 @@ control the Kubernetes cluster. In {product-title}, the master machines are
...
+use any number of master nodes, the number is constrained by etcd quorum due to
```
xingxingxia: Not sure whether appropriate to link external https://github.com/etcd-io/etcd/blob/master/Documentation/faq.md#why-an-odd-number-of-cluster-members here. If yes, could link it.

xingxingxia:
@codyhoag LGTM except minor comments, pls check, thanks!

modules/architecture-machine-roles.adoc
codyhoag: Makes sense; I'll remove it

modules/architecture-machine-roles.adoc
codyhoag: @xingxingxia we cannot link to Github links in the official docs. Is there any additional info from that link that you think should be included here?

xingxingxia: Currently short statement is OK

codyhoag:
Minor comments from QE addressed. This is now Verified (within BZ).

modules/architecture-machine-roles.adoc
```diff
@@ -44,12 +44,19 @@ control the Kubernetes cluster. In {product-title}, the master machines are
...
-and the terms "master" and "control plane" are used interchangeably to describe
+the terms _master_ and _control plane_ are used interchangeably to describe
 them. Instead of being grouped into a
 MachineSet, master machines are defined by a series of standalone machine API
 resources. Extra controls apply to master machines to prevent you from deleting
 all master machines and breaking your cluster.
 
+[NOTE]
+====
+It is recommended to use three master nodes. Although you can theoretically
```
ahardin-rh: s/It is recommended to use/Use

ahardin-rh:
Looks good! Just one suggestion from me.

PR #20275, opened 2020-03-05
removing openshift.io/cluster-logging: 'true'

mburke5678:
Per @sichvoge

### DIFF
--- a/modules/cluster-logging-deploy-clo-cli.adoc
+++ b/modules/cluster-logging-deploy-clo-cli.adoc
@@ -24,7 +24,6 @@ metadata:
   annotations:
     openshift.io/node-selector: ""
   labels:
-    openshift.io/cluster-logging: "true"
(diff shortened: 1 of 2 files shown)


mburke5678:
@QiaolingTang Can you review this small change?

QiaolingTang:
@mburke5678 will these changes be in 4.1~4.3 ?

mburke5678:
@QiaolingTang As far as I understand. @sichvoge please let us know if that is not true.

sichvoge:
As far as I understood, that annotation had no effect from the get go. But I don't see this as urgent to "backport". It's ok if we only do a test on a newer version of OCP (e.g. next is 4.4).

mburke5678:
@sichvoge @QiaolingTang I cherrypicked to 4.3 and 4.4. If you think we should back port to 4.2 and 4.1, let me know.

PR #20278, opened 2020-03-05
Remove Throttling topic from logging

mburke5678:
Per @jcantrill throttling Fluentd has never worked in 4.x and should be removed.


vikram-redhat:
@mburke5678 this should go through the change control for features already documented being removed.

mburke5678:
@jcantrill @sichvoge @xltian @sferich888 @reestr 
Please take a look and let me know if anyone has concerns with removing this topic from the docs..

sferich888:
@jcantrill why are we not treating this as a gap between 3.x and 4.x?

xltian:
@anpingli  Can you please review and confirm that  if it's true " throttling Fluentd has never worked in 4.x and should be removed."?

jcantrill (approved):

anpingli (commented):
LGTM

jcantrill:
> @jcantrill why are we not treating this as a gap between 3.x and 4.x?

This is a gap between 3 and 4 with no good user experience since its introduction. We additionally do not have a great understanding of what it means to set a given value and how that translates into actually restricting the log throughput.

This feature will be re-visited after the introduction of GA ClusterLogForwarder in 4.5 which will properly surface API to expose it.  Also, there is work by the logging team to introduce changes to the underlying kubelet to add a container log backpressure policy to control rogue containers and keep from overwhelming the collector and logging system in general

sichvoge:
Agree with Jeff. We are working on more appropriate mechanisms to actually solve customers problems rather than having a semi-working feature.

mburke5678:
@sferich888 Should we remove this information from the docs in light of @sichvoge comment that they are working on a new solution?

sferich888:
If it's not working we should remove it from the docs.

sichvoge:
@mburke5678 For what versions will you remove it? Should we do it for all 4? cc/ @jcantrill

mburke5678:
@sichvoge @jcantrill I can remove from 4.2+ without approval. 4.1 is officially closed for changes. if the concern is high enough, I can ask for an exception.

jcantrill:
> @mburke5678 For what versions will you remove it? Should we do it for all 4? cc/ @jcantrill

4.3+ because the run.sh script otherwise did not change in 4.2 and it would still work though I am unable to attest to it being properly tested and verified

mburke5678:
> the run.sh script otherwise did not change in 4.2 and it would still work though I am unable to attest to it being properly tested and verified

@anpingli Can you verify?

PR #20291, opened 2020-03-06
Remove ControllerLeaseTTL reference from 3.11

fmartingr:
# Summary
Removed `ControllerLeaseTTL` reference from the documentation.

# Reason
- `ControllerLeaseTTL` was removed on [origin#19189](https://github.com/openshift/origin/pull/19189) and has been gone from 3.10.0 onward.

# Notes
- I wanted to update the documentation with the new `Election` field but I couldn't find the `ControllerTTL` reference it mentions. ([Source](https://github.com/openshift/origin/blob/v3.10.0/pkg/cmd/server/apis/config/types.go#L1425-L1437))
- It seems that my auto clean has removed some dangling whitespaces from the documentation as well. If that's a problem I can revert those changes.


vikram-redhat:
@xltian do you know who can confirm this for us?

PR #20305, opened 2020-03-06
Errata update for 4.3.5 release

codyhoag:
Ships 2020-03-09


codyhoag:
@bergerhoffer I included your note on service CA auto-rotation. Can you take a quick look to make sure it looks good? Thanks!

bergerhoffer:
@codyhoag It looks perfect, thanks! I just confirmed that it is actually making it for 4..3.5, so we should be good this time!

And just a reminder that all z-streams going forward should have had this note as well if possible. Thanks again for all your help with this!

PR #20328, opened 2020-03-09
Pipelines structure, attribute file, and folders and symlinks for pipelines

Preeticp:
This PR:
- Adds a commented out structure of docs for OpenShift pipeline in the OpenShift Pipeline bucket.
- Moves Pipelines bucket to be on the same level as Builds bucket instead of inside the Builds bucket.
- Adds an attribute file for OpenShift Pipelines
- Adds a tkn_cli bucket under the CLI Tools bucket
- Adds openshift_pipelines folder with symlinks
- Adds tkn_cli folder and symlinks

Note the modifications in the topic.yml are commented out as we are still DP and are likely to go TP by the end of this month.


Preeticp:
@bmcelvee can you please take a look at this PR. It creates 2 buckets for Pipeline docs, one below the Builds bucket and the other one for Pipelines cli `tkn` in the CLI reference bucket. I am asking you cause you were designated as our buddy for the initial integration of Pipeline docs.
Initially, I thought the build failed because the sections were commented out, but now I think it is cause we do not have actual assemblies in place, is that correct?

_topic_map.yml
```diff
@@ -1235,6 +1245,14 @@ Topics:
   Topics:
   - Name: Getting started with Helm on OpenShift Container Platform
     File: getting-started-with-helm-on-openshift-container-platform
+#- Name: tkn CLI
```
bmcelvee: I think these should go in the CLI tools bucket.

bmcelvee: Ignore that content - you've already got it there. Sorry!

cli_reference/tkn_cli/images
```diff
@@ -0,0 +1 @@
+../images
```
bmcelvee: The symlinks might be causing the build failure. Have you tried commenting them out? Or removing them and building locally?

Preeticp: I will try that, thank you!

Preeticp:
@bmcelvee request a review and merge, please. I checked with Vikram regarding the build failure, it was cause of the separator not being commented out. I have corrected that and have added the symlinks back.

bmcelvee:
LGTM! Glad he was able to help figure it out! 

This applies to 4.4, correct? I just want to be sure before cherry picking it over.

Preeticp:
Thank you Brandi, afaik, this will be an async release, as in, the docs will be published only once they go TP. They are aiming to do TP release before the month end, but it could be a bit delayed too. Hence I have commented out this section. I will confirm this again with the team and get back to you as soon as I have their opinion.

Preeticp:
Hi @bmcelvee I just got confirmation from Siamak, since TP would be happening before the 4.4 GA date we would want to push the docs in 4.3

bmcelvee:
Sounds good, thanks! I'm adding the 4.3 label and cherry picking to both 4.3 and 4.4.

PR #20334, opened 2020-03-09
Add documentation for multiple indetity providers order

vareti:
Add documentation to clear things on the order in which IDPs are checked if more than one of them is configured.


vareti:
not much activity. closing for now

/close

PR #20363, opened 2020-03-09
Include namespace+pod selector network policy

jboxman:
- https://bugzilla.redhat.com/show_bug.cgi?id=1757221

### DIFF
--- a/modules/nw-networkpolicy-about.adoc
+++ b/modules/nw-networkpolicy-about.adoc
@@ -100,7 +100,7 @@ spec:
 * Only allow HTTP and HTTPS traffic based on Pod labels:
 +
 To enable only HTTP and HTTPS access to the Pods with a specific label
-(`role=frontend` in following example), add a NetworkPolicy object similar to:
+(`role=frontend` in following example), add a NetworkPolicy object similar to the following:
 +
 [source,yaml]
 ----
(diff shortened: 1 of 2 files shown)


modules/nw-networkpolicy-about.adoc
```diff
@@ -120,6 +120,31 @@ spec:
...
+* Accept connections by using both namespace and Pod selectors:
++
+To match network traffic by combining namespace and Pod selectors, you can use a NetworkPolicy object similar to the following:
++
+[source,yaml]
+----
+kind: NetworkPolicy
+apiVersion: extensions/v1beta1
+metadata:
+  name: allow-pod-and-namespace-both
```
anuragthehatter: snake style would be preferred here ```allow_pod_and_namespace_both```

modules/nw-networkpolicy-about.adoc
```diff
@@ -120,6 +120,31 @@ spec:
...
 ----
 
+* Accept connections by using both namespace and Pod selectors:
++
+To match network traffic by combining namespace and Pod selectors, you can use a NetworkPolicy object similar to the following:
++
+[source,yaml]
+----
+kind: NetworkPolicy
+apiVersion: extensions/v1beta1
```
anuragthehatter: apiVersion is now ```networking.k8s.io/v1```. This also needs to be updated in all releases and current one.

modules/nw-networkpolicy-about.adoc
jboxman: @anuragthehatter, is there an OpenShift convention for this? The dashes originated with the OCP 3.x docs that I ported for 4.1.

anuragthehatter: @jboxman Thats fine. I thought you referred from the bug description where i pasted this object. I am not sure of any convention though.

modules/nw-networkpolicy-about.adoc
jboxman: @anuragthehatter, do you mean all releases of OCP?

modules/nw-networkpolicy-about.adoc
jboxman: @anuragthehatter, we've been using `-` somewhat consistently throughout the docs in YAML. I found 572 instances of a `-` in a `name` key across 283 files in `master`. Ideally, I'd like to keep using `-` in the examples for `NetworkPolicy`.

modules/nw-networkpolicy-about.adoc
anuragthehatter: @jboxman I noticed all earlier releases already have "networking.k8s.io/v1" so that seems good. Just want to make sure we don't have "extensions/v1beta1" somewhere

jboxman:
@anuragthehatter, good to merge then?

Thanks.

PR #20369, opened 2020-03-10
Update REST APIs

jboxman:
This would have the latest APIs.


vikram-redhat:
@jboxman can this be merged?

jboxman:
@vikram-redhat, it's lacking any of the improvements made subsequently. I may be able to finalize those fixes this weekend, and then update this PR. I need to confirm that the fix for the Prometheus spec produces somewhat useable output; could also release with the huge 200 attribute table if necessary.

At times, there isn't enough water for all the fires.

vikram-redhat:
> At times, there isn't enough water for all the fires.

:D

jboxman:
Closing in favor of https://github.com/openshift/openshift-docs/pull/22214.

Issue #20254, opened 2020-03-05
[enterprise-4.3] Edit suggested in file metering/configuring_metering/metering-configure-persistent-storage.adoc


filiy:
### Which section(s) is the issue in?

Under the Storing data in Microsoft Azure. In the example under "You can use the following command to create the secret."

### What needs fixing?

oc create secret -n openshift-metering generic your-azure-secret --from-literal=azure-storage-account-name=your-storage-account-name --from-literal=azure-secret-access-key-your-secret-key

Should be:

oc create secret -n openshift-metering generic your-azure-secret --from-literal=azure-storage-account-name=your-storage-account-name --from-literal=azure-secret-access-key=your-secret-key

vikram-redhat:
@maximsvistunov PTAL

vikram-redhat:
https://docs.openshift.com/container-platform/4.3/metering/configuring_metering/metering-configure-persistent-storage.html#metering-store-data-in-azure_metering-configure-persistent-storage

rh-max:
Thanks @filiy for reporting! Addressed in https://github.com/openshift/openshift-docs/pull/20370.
@vikram-redhat , could you please merge #20370 and cherry-pick it into `enterprise-4.3` and `enterprise-4.2`? It's a simple fix and needs no further review.

PR #20370, opened 2020-03-10
Fix a command

rh-max:
This addresses https://github.com/openshift/openshift-docs/issues/20254.


rh-max:
Thanks @vikram-redhat !

PR #20394, opened 2020-03-10
Prefer short flag for whoami -t

jboxman:
@vikram-redhat, probably oldest cherry-pick in the entire repo. This was fixed by a commit in 2016, but apparently never made it into _enterprise-3.11_.


jboxman:
Verified on Slack.

PR #20398, opened 2020-03-11
OSDOCS-957: Updating restart requirements due to auto rotation for 4.4

bergerhoffer:
https://issues.redhat.com/browse/OSDOCS-957

Preview: https://osdocs-878-44--ocpdocs.netlify.com/openshift-enterprise/latest/updating/updating-cluster-between-minor.html

### DIFF
--- a/modules/update-upgrading-cli.adoc
+++ b/modules/update-upgrading-cli.adoc
@@ -145,19 +145,3 @@ $ oc get clusterversion
 NAME      VERSION     AVAILABLE   PROGRESSING   SINCE     STATUS
 version   4.1.2       True        False         2m        Cluster version is 4.1.2
 ----
-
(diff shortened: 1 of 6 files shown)


bergerhoffer:
@marun @barleyer Can you both please review these updates for 4.4? 

I've removed the notes about having to restart from the sections about upgrading within a minor version (like 4.4.0 -> 4.4.1) as that should no longer be necessary since 4.4.0 will have auto rotation already. And I have updated the notes in the sections about upgrading between minor versions (like 4.3.1 -> 4.4.0) since that still might be required if they're upgrading from a release prior to 4.3.5.

Let me know if you see any issues, thanks!

bergerhoffer:

adellape (approved):
:+1:

PR #20406, opened 2020-03-11
Reintroduce GA'd features in TP table

codyhoag:
@vikram-redhat based on our discussion, I've re-added the long standing GA features to the TP table I had initially removed for the 4.4 release notes.

@sheriff-rh FYI: for precautionary reasons, we're going to keep the long standing GA features in the TP table for now.


sheriff-rh:
Thanks for taking charge on this @codyhoag !

PR #20440, opened 2020-03-12
Build: Clarify behavior of linked pull/push secrets

adambkaplan:
Pull and push secrets do not need to be linked to the builder service account.
A build works best if the pull/push secrets are added to the namespace and
referenced directly in the BuildConfig.

[DEVEXP-510](https://issues.redhat.com/browse/DEVEXP-510)


adambkaplan:
@bmcelvee FYI linking a pull/push secret to the `builder` service account lets it be considered for "automatic" pull secrets. I.e. a customer does not need to add the `pullSecret` or `pushSecret` field to the `BuildConfig`.

Personally I prefer to discourage that behavior - it is not GitOps friendly and makes it difficult to debug a build that is failing because it can't pull or push an image.

bmcelvee:
Thanks, @adambkaplan! Does this apply to every version of 4.x?

I'm going to label it 4.4 and 4.5 for now. 

No QE review required.

adambkaplan:
@bmcelvee I've broadened this to any secret linking to the builder service account.

Per @deads2k a pod doesn't need a secret linked to a service account for it to be used. This has been the case since we released 4.1. Builds don't need secrets linked as an image pull secret because image pulls happen from within the build container, not from the kubelet.

The only advantage to linking the secret today is that it lets builds use a pull secret without it being referenced in the `BuildConfig`. @bparees @openshift/openshift-team-developer-experience I'm inclined to discourage this behavior. I recall we had at least one BZ where it was hard to debug a build because we couldn't identify which pull secret was used to obtain an image - I don't recall offhand if it was limited to builds or imagestream import.

adambkaplan:
Marking work in progress so we can discuss this a bit further.

bparees:
> The only advantage to linking the secret today is that it lets builds use a pull secret without it being referenced in the BuildConfig. 

that's a pretty big advantage

> @bparees @openshift/openshift-team-developer-experience I'm inclined to discourage this behavior. 

what behavior do you want to discourage?  defining BCs w/o explicitly specifying the pullsecret to use?  I think we need to continue to support/except that type of usage.  For 99% of builds that just push to an imagestream in the Build's own namespace, it works great and expecting people to take an additional step doesn't make sense.

If we want to change anything here, I think we'd want to change the actual behavior to not require linking to the SA (buildcontroller just iterates all the secrets in the NS, looking for a match).  But of course that means we might start picking up more secrets than we used to, and that might include secrets which look valid for the registry but don't actually work.


So imho the potential resolutions here are:

A) change the build controller behavior (in which case this doc can be deleted since the linking step is no longer necessary)

or

B) more clearly document the build controller behavior so users understand why the linking step is necessary/what it accomplishes. (among other things it no longer has anything to do w/ making it mountable within the pod).

but i don't think we should just delete the documentation of the linking step and expect people to just always set a pullsecret in their buildconfig.  It makes setting up a new build too burdensome.

adambkaplan:
> B) more clearly document the build controller behavior so users understand why the linking step is necessary/what it accomplishes. (among other things it no longer has anything to do w/ making it mountable within the pod).

I think this is the right thing to do here. I can move the deleted content to a new section that describes the feature.

modules/builds-docker-credentials-private-registries.adoc
```diff
@@ -87,6 +79,14 @@ the build configuration:
 $ oc set build-secret --push bc/sample-build dockerhub
 ----
 
+. You can also link the push secret to the *builder* service account instead of specifying the
+`pushSecret` field. Push secrets linked to the *builder* service account can be automatically added
+to any build in the namespace.
```
adambkaplan: @bparees @bmcelvee Moved the secret linking as an optional step in the procedure. Please let me know if this phrasing makes sense.

bparees: it doesn't.

push secrets don't have to be linked at all.  we only search the build's SA for linked secrets for pulling.  We don't do any logic to search for push secrets, either we use what we found for pull, or we use the pushsecret you explicitly specified (which, again, does not need to be linked to the SA anymore)

modules/builds-docker-credentials-private-registries.adoc
```diff
@@ -115,3 +115,11 @@ This example uses `pullSecret` in a Source build, but it is also applicable
 in Docker and Custom builds.
 ====
 endif::[]
+
+. You can also link the pull secret to the *builder* service account instead of specifying the
+`pullSecret` field. Pull secrets linked to the *builder* service account can be automatically added
+to any build in the namespace.
```
bparees: this is more or less accurate but i'd reword it.

"can be automatically added" is too vague.

what you need to convey is the behavior which is:  if an explicit pullsecret is not specified on the build, the system will look at secrets linked to the build's SA and determine if any of them contain auth entries that reference the registry that is the target of the pull and, if it finds one, will use it.

Note also that there's no requirement the build is using the "builder" SA, that's just a default, so that needs to be made clear also.  If they are using some other SA, they need to link the secrets to that other SA.

modules/builds-docker-credentials-private-registries.adoc
bparees: > We don't do any logic to search for push secrets

I should qualify this.....  as far as i know we don't.  my memory on this isn't great, so i could be wrong, but i don't think we do.

adambkaplan: The relevant function is `resolveImageSecretAsReference`. We call it in the following locations:

1. Resolve the push secret: https://github.com/openshift/openshift-controller-manager/blob/master/pkg/build/controller/build/build_controller.go#L1113-L1122
2. Resolve the pull secret: https://github.com/openshift/openshift-controller-manager/blob/master/pkg/build/controller/build/build_controller.go#L1144-L1162
3. Resolve the pull secret for image sources: https://github.com/openshift/openshift-controller-manager/blob/master/pkg/build/controller/build/build_controller.go#L1165-L1176

So it looks like I should also add similar text for the image sources.

bparees: ok, i stand corrected..in that case my comments from the other section apply here as well in terms of better describing the actual flow that happens to pick a secret

adambkaplan:
@bparees @bmcelvee updated the language around linked pull/push secrets for builds. Also added similar statements to the build image sources section. PTAL.

modules/builds-docker-credentials-private-registries.adoc
```diff
@@ -115,3 +116,18 @@ This example uses `pullSecret` in a Source build, but it is also applicable
...
+automatically added to the build if the secret contains a credential that matches the repository
+hosting the build's input image.
++
+----
+$ oc secrets link builder dockerhub
+----
+
+[NOTE]
+====
+In some circumstances Docker strategy builds which use `Binary` source may not have pull secrets
```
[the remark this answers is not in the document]
gabemontero: It does go through that path I noted ... I had to debug / diagnose this as part of triaging a recent BZ from QE, https://bugzilla.redhat.com/show_bug.cgi?id=1810920#c4 , that is suppose to be addressed by this PR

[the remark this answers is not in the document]
bparees: ok, that's not specific to binary build.  You could have the same situation if you had a git repo that contained your Dockerfile, and did not specify the "From" image in your buildconfig.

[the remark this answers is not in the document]
bparees: (you also have the same situation if you inline a Dockerfile in the buildconfig, i believe, because i don't think we parse your dockerfile and try to extract the From image to choose a secret, as far as i recall)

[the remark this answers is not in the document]
bparees: basically pull secret matching only works if you explicitly tell us, in your buildconfig, via the From field, what image your build needs.  If it's implicit (defined in the Dockerfile) it doesn't work.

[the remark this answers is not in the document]
gabemontero: sound goods ... circling back to the line we are commenting on here, I don't think this note as is sufficiently covers minimally the situation wrt binary build ... and if there is a broader scope, either the updated note should address that, or the note should be nuked in lieu a more generic statement along those lines somewhere here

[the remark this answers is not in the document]
bparees: agree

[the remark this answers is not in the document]
adambkaplan: I'll update the note that for any `Docker` strategy build the buildConfig must have the `from:` field populated. This is a given for all `Source` strategy builds.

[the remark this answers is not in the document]
gabemontero: Going back to https://bugzilla.redhat.com/show_bug.cgi?id=1810920 I see that QE created their BC by running `oc  new-build --binary --name prune --to prune:latest`

Seems a bit iffy that they'll realize they need to cross reference this note with the resulting BC from our CLI:

```
apiVersion: build.openshift.io/v1
kind: BuildConfig
metadata:
  annotations:
    openshift.io/generated-by: OpenShiftNewBuild
  creationTimestamp: "2020-03-12T07:24:38Z"
  labels:
    build: prune
  name: prune
  namespace: wzheng1
  resourceVersion: "129030"
  selfLink: /apis/build.openshift.io/v1/namespaces/wzheng1/buildconfigs/prune
  uid: 9be8bd7c-58d7-4b29-8aef-9cc8ebfbfa6a
spec:
  failedBuildsHistoryLimit: 5
  nodeSelector: null
  output:
    to:
      kind: ImageStreamTag
      name: prune:latest
  postCommit: {}
  resources: {}
  runPolicy: Serial
  source:
    binary: {}
    type: Binary
  strategy:
    dockerStrategy: {}
    type: Docker
  successfulBuildsHistoryLimit: 5
  triggers:
  - github:
      secret: pDP4dNb7u7DBAn6KCOmy
    type: GitHub
  - generic:
      secret: YVMKtFuDF90Frys48P4u
    type: Generic
status:
  lastVersion: 1
```

At first blush, seems like 
- perhaps we note at least short term that resulting binary BCs from `oc new-build` (and I'm guessing `oc new-app` as well) may need manual adjustment per this new note
- it may not be quick fix, but do we entertain augmenting the CLI to fill in that field (based on fetching and parsing of the Dockerfile it finds on the local system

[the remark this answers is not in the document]
bparees: > it may not be quick fix, but do we entertain augmenting the CLI to fill in that field (based on fetching and parsing of the Dockerfile it finds on the local system

I think just doc'ing the limitation + implications is sufficient.

modules/builds-docker-credentials-private-registries.adoc
```diff
@@ -115,3 +116,18 @@ This example uses `pullSecret` in a Source build, but it is also applicable
...
++
+----
+$ oc secrets link builder dockerhub
+----
+
+[NOTE]
+====
+You must specify a `from` image in the `BuildConfig` spec to take advantage of this feature.
+Docker strategy builds generated by `oc new-build` or `oc new-app` may not do this in some situations.
+====
```
adambkaplan, line 133: @bparees @gabemontero Updated this note per your feedback. PTAL.

bparees:
lgtm.

modules/builds-docker-credentials-private-registries.adoc
gabemontero: note looks good

modules/builds-docker-credentials-private-registries.adoc
```diff
@@ -87,6 +79,15 @@ the build configuration:
...
 
+. You can also link the push secret to the service account used by the build instead of specifying
+the `pushSecret` field. By default builds use the *builder* service account. The push secret is
+automatically added to the build if the secret contains a credential that matches the repository
+hosting the build's output image.
++
+----
+$ oc secrets link builder dockerhub
+----
+
```
gabemontero, line 90: should the you must specify a from note be placed here as well ?

modules/builds-image-source.adoc
```diff
@@ -55,6 +55,15 @@ source:
 <5> The location of the file to be copied out of the referenced image.
 <6> An optional secret provided if credentials are needed to access the input image.
 
+If an input image requires a pull secret, you may optionally link the pull secret to the service
```
bmcelvee, line 58: Suggested rewrite: 

Optionally, if an input image requires a pull secret you can link the pull secret to the service account used by the build.

bmcelvee:
@adambkaplan - I just had a couple of suggestions that I don't mind adding in a follow-up PR. I initially noted that this PR doesn't need QE review, but if you'd like I can request it anyway. Thanks again!

modules/builds-docker-credentials-private-registries.adoc
adambkaplan: I don't think so - you always need to specify the output image for push. If no output image is specified, image push doesn't happen 😄

adambkaplan:
@bmcelvee @bparees @gabemontero updated & squashed

modules/builds-docker-credentials-private-registries.adoc
gabemontero: sounds good

gabemontero:
LGTM

will let @bmcelvee or @bparees apply actual label

bmcelvee:
Merging now, and opening a follow-up PR to fix some formatting. Will link it here as soon as it's available.

PR #20454, opened 2020-03-13
The default network provider for OKD is OVNKubernetes

jboxman:
OVNKubernetes is the default network provider for OKD.


vrutkovs (approved):

ahardin-rh:
Looks good!

PR #20105, opened 2020-02-28
update configuring sriov doc

zenghui-shi:

PR #20456, opened 2020-03-13
Sriov with topics

jboxman:
Split the SR-IOV topics out into assemblies.

### DIFF
--- a/networking/hardware_networks/about-sriov.adoc
+++ b/networking/hardware_networks/about-sriov.adoc
@@ -1,9 +1,11 @@
-[id="configuring-sr-iov"]
-= Configuring an additional network for SR-IOV
+[id="about-sriov"]
+= About Single Root I/O Virtualization (SR-IOV) hardware networks
 include::modules/common-attributes.adoc[]
-:context: configuring-sr-iov
+:context: about-sriov
 
 toc::[]
+
+You can use Single Root I/O Virtualization (SR-IOV) network devices with additional networks on your {product-title} cluster for high performance applications.
 
 [id="about-using-sr-iov-hardware"]
 == About SR-IOV hardware on {product-title}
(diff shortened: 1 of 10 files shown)


jboxman:
Need to include this content, from the updated assembly, as well: https://github.com/openshift/openshift-docs/pull/20105

jboxman:
@zshi-redhat @fepan 

This separates the SR-IOV sections into topics in the table of contents under _Networking -> Hardware networks_.

Does that make sense? Or is a name other than _hardware networks_ more appropriate?

Thanks!

zshi-redhat:
> @zshi-redhat @fepan
> 
> This separates the SR-IOV sections into topics in the table of contents under _Networking -> Hardware networks_.
> 
> Does that make sense? Or is a name other than _hardware networks_ more appropriate?

Yes, I think this is a reasonable move as SR-IOV doc is too large to be held in multi-network section and it has its own path for installation and configuration via operator which is different from other multi-network examples (installed by default).

Hardare networks should provide enough information to distinguish with other sections like OpenShiftSDN etc and it allows us to add more hardware related features in the future such as FPGA, Infiniband. If we want to be more specific about the feature, the alternative could be SR-IOV networks.

> 
> Thanks!

networking/hardware_networks/configuring-sriov-device.adoc
```diff
@@ -0,0 +1,16 @@
...
+include::modules/common-attributes.adoc[]
+:context: about-sriov
+
+toc::[]
+
+You can configure a Single Root I/O Virtualization (SR-IOV) device in your cluster.
+
+include::modules/nw-sriov-configuring-device.adoc[leveloffset=+1]
+
+include::modules/nw-sriov-add-pod-runtimeconfig.adoc[leveloffset=+1]
```
zshi-redhat: Please move `include::modules/nw-sriov-add-pod-runtimeconfig.adoc` to `networking/hardware_networks/configuring-sriov-net-attach.adoc` file and put it after `include::modules/nw-multus-ipam-object.adoc`.

The content in `nw-sriov-add-pod-runtimeconfig` is mainly about how to configure sriov net-attach-def with static MAC and IP addresses.

zenghui-shi (commented):
few comments about including files.

networking/hardware_networks/using-dpdk-and-rdma.adoc
```diff
@@ -0,0 +1,37 @@
...
+:context: about-sriov
+
+toc::[]
+
+You can use Single Root I/O Virtualization (SR-IOV) network hardware with the Data Plane Development Kit (DPDK) and with remote direct memory access (RDMA).
+
+[id="examples-of-using-virtual-functions"]
+== Examples of using virtual functions in DPDK and RDMA modes
+
+[IMPORTANT]
```
ahardin-rh: Could you reuse the technology-preview.adoc here instead?

ahardin-rh:
Just one question from me. Looks good!

jboxman:
cc @lamek - I think this is going to happen in time for OCP next.

jboxman:
@zshi-redhat, possibly SR-IOV networks, then. I would expect that other hardware networks might have a similarly large number of topics? An Operator, configuration of operator, configuration of network devices, and so on?

networking/hardware_networks/using-dpdk-and-rdma.adoc
jboxman: @ahardin-rh, probably; It's originally contributed content and it may be someone pasted it in there, rather than including it, maybe because it was originally in a module.

I'll see if I can fix it.

networking/hardware_networks/about-sriov.adoc
```diff
@@ -63,57 +65,13 @@ can be disabled by editting default SriovOperatorConfig CR.
 ====
 
 include::modules/nw-sriov-supported-devices.adoc[leveloffset=+2]
-
 include::modules/nw-sriov-device-discovery.adoc[leveloffset=+2]
```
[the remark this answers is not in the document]
jboxman: @zshi-redhat, done!

zshi-redhat:
> @zshi-redhat, possibly SR-IOV networks, then. I would expect that other hardware networks might have a similarly large number of topics? An Operator, configuration of operator, configuration of network devices, and so on?

Yes, you're right Jason.There will be other types of hardware be supported in the future, our current plan is to put their SR-IOV related functions be supported in the same SR-IOV Operator project. Example could be using SR-IOV operator for infiniband or FPGA SR-IOV devices.

I cannot exclude the possibility to support other non-SR-IOV hardware devices in the future releases, if there were separate operators for these non-SR-IOV devices, perhaps they shall also be put in separate section in the doc?

jboxman:
(self) Need to fix:

```
ID "discover-sr-iov-devices_about-sriov" is duplicated in the source content
build/master.xml fails to validate
```

jboxman:
@zshi-redhat, it may be best to work through that if and when it arises; I don't want to prematurely optimize this. Such content could also fit under _hardware networks_, perhaps with an introductory topic that explains each of the major hardware types. It's certainly workable within a single directory.

In the meantime, I think I'll merge this as is.

PR #20460, opened 2020-03-13
add jenkins http proxy environment variables setting

jkhelil:


modules/images-other-jenkins-env-var.adoc
```diff
@@ -131,4 +131,13 @@ project. This variable must be set before Jenkins starts the first time for it
 to have an effect.
 |Default Maven agent image in Jenkins server:
 `image-registry.openshift-image-registry.svc:5000/openshift/jenkins-agent-maven:latest`
+
+|`http_proxy`,
+`https_proxy`
+`no_proxy`
+|Configure the http/https proxy on jenkins.
+These values could be set from the cluster wide proxy object.
+|oc set env dc/jenkins http_proxy=$(oc get proxy cluster -o jsonpath='{.status.httpProxy}') \
```
akram: Move this sentence into the next column: 
`These values could be set from the cluster wide proxy object.`

and add backquotes around the command line example

jkhelil: fixed

akram:
@bmcelvee can you PTAL ?
/lgtm

PR #20487, opened 2020-03-16
Bug 1813819: [4.2] Remove known issue for pod sysctls in 4.2 as it exist in 4.1 only

liangxia:
https://bugzilla.redhat.com/show_bug.cgi?id=1813819

Known issue for pod sysctls is in OCP 4.1, which is wrongly applied to OCP 4.2 in 4.2 release note.

Apply to enterprise-4.2 only. @openshift/team-documentation

### DIFF
--- a/release_notes/ocp-4-2-release-notes.adoc
+++ b/release_notes/ocp-4-2-release-notes.adoc
@@ -1297,7 +1297,7 @@ indicate that the feature is removed from the release or deprecated.
 |Pod sysctls
 |GA
 |GA
-|GA. See xref:ocp-4-2-known-issues[Known issues] for current limitations.
+|GA
 
 |Central Audit
 |GA


kalexand-rh:
@vikram-redhat, do you have an opinion about linking to the 4.1 release notes from the 4.2 release notes?

vikram-redhat:
Ideally, we shouldn't be creating dependencies like this between release notes. I am confused about this linking though. Why is the dependency noted in the 4.1 docs if the issue exists in 4.2 as well? Wouldn't it be better to add that limitation to the 4.2 release notes docs as well?

liangxia:
The issue exist for version 4.1, and we have that noted for version 4.1 in 4.1 release note/branch.
While in branch 4.2, we have support scope for both 3.11, 4.1 and 4.2, but it is wrongly noted for version 4.2 in 4.2 release note/branch.
The PR move the note from version 4.2 to version 4.1 in 4.2 release note/branch, so it is in 4.2 branch, but fix the issue for version 4.1

liangxia:
@kalexand-rh @vikram-redhat  Trying to fix the issue in a much simpler way. As the known issue only exist for version 4.1, now the fix simply remove the wrongly apply note from version 4.2 in 4.2 release note.

Issue #20480, opened 2020-03-14
[enterprise-4.3] Edit suggested in file logging/cluster-logging-deploying.adoc


janehojy:
### Which section(s) is the issue in?

Deploying cluster logging > Install the Elasticsearch Operator using the CLI 

### What needs fixing?

Under section 2, **Install the Elasticsearch Operator by creating the following objects:**

c. Create a Subscription object YAML file (for example, eo-sub.yaml) to subscribe a Namespace to an Operator.

Example subscription
```
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  name: "elasticsearch-operator"
  namespace: "openshift-operators-redhat" 
spec:
  channel: "4.3" 
  installPlanApproval: "Automatic"
  source: "redhat-operators"
  sourceNamespace: "openshift-marketplace"
  name: "elasticsearch-operator"
```

it should be stated clearly that `source` has to match the name used in the CatalogSource object.

Especially for disconnected clusters, most people would have named it `my-operator-catalog` and not `redhat-operators`, if they followed the documentation [here](https://docs.openshift.com/container-platform/4.3/operators/olm-restricted-networks.html#olm-restricted-networks-operatorhub_olm-restricted-networks) which says:

```
4. Create a CatalogSource object that references your catalog image.
a. Modify the following to your specifications and save it as a catalogsource.yaml file:

> apiVersion: operators.coreos.com/v1alpha1
> kind: CatalogSource
> metadata:
>   name: my-operator-catalog
>   namespace: openshift-marketplace
> spec:
>   sourceType: grpc
>   image: <registry_host_name>:<port>/olm/redhat-operators:v1 
>   displayName: My Operator Catalog
>   publisher: grpc
```

In which case, `source: "redhat-operators"` in the eo-sub.yaml file would not work for them. It has to be `source: my-operator-catalog`.

janehojy:
Same goes for the clo-sub.yaml file if [installing the Cluster Logging Operator using the CLI](https://docs.openshift.com/container-platform/4.3/logging/cluster-logging-deploying.html#cluster-logging-deploy-clo-cli_cluster-logging-deploying).

```
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  name: cluster-logging
  namespace: openshift-logging 
spec:
  channel: "4.3" 
  name: cluster-logging
[...]

PR #20504, opened 2020-03-16
Logging install subscription source to match CSC name

mburke5678:
https://github.com/openshift/openshift-docs/issues/20480

### DIFF
--- a/modules/cluster-logging-deploy-clo-cli.adoc
+++ b/modules/cluster-logging-deploy-clo-cli.adoc
@@ -82,11 +82,13 @@ metadata:
 spec:
   channel: "4.3" <2>
   name: cluster-logging 
-  source: redhat-operators 
+  source: redhat-operators <3>
   sourceNamespace: openshift-marketplace
 ----
 <1> You must specify the `openshift-logging` Namespace.
 <2> Specify `4.3` as the channel.
+<3> Specify `redhat-operators`. If your {product-title} cluster is installed on a restricted network, also known as a disconnected cluster, 
+specify the name of the CatalogSource object you created when you configured the Operator Lifecycle Manager (OLM).
 
 .. Create the Subscription object:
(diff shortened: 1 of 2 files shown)


modules/cluster-logging-deploy-clo-cli.adoc
```diff
@@ -82,11 +82,13 @@ metadata:
 spec:
   channel: "4.3" <2>
   name: cluster-logging 
-  source: redhat-operators 
+  source: redhat-operators <3>
   sourceNamespace: openshift-marketplace
 ----
 <1> You must specify the `openshift-logging` Namespace.
 <2> Specify `4.3` as the channel.
+<3> Specify `redhat-operators`. If your {product-title} cluster is installed on a restricted network, also known as a disconnected cluster,
```
jcantrill: why should this be different then `openshift-operators-redhat`?  The fact the environment is disconnected seems immaterial.  The namespace `openshift-operators-redhat` was chosen as an alternate to the original global namesapce of 'openshift-operators' so the prometheus stack could be enabled to scrape metrics for global redhat operators

mburke5678: This is the catalog source in the Subscription object. In a disconnected install, the source is different than a connected install and the ConfigSource object points to that source. Thus, the cluster logging Subscription must point to the ConfigSource object to get to the source.

**Build the Operator catalog image:**
https://docs.openshift.com/container-platform/4.3/operators/olm-restricted-networks.html#olm-building-operator-catalog-image_olm-restricted-networks
**Configuring OperatorHub for restricted networks**
https://docs.openshift.com/container-platform/4.3/operators/olm-restricted-networks.html#olm-restricted-networks-operatorhub_olm-restricted-networks

QiaolingTang:
lgtm

Issue #18027, opened 2019-11-08
[enterprise-4.2] Edit suggested in file installing/installing_restricted_networks/installing-restricted-networks-vsphere.adoc


mlcassatt:
### Which section(s) is the issue in?
User-provisioned DNS requirements

### What needs fixing?
etcd

etcd-<index>.<cluster_name>.<base_domain>

OpenShift Container Platform requires DNS records for each etcd instance to point to the control plane machines that host the instances. The etcd instances are differentiated by <index> values, which start with 0 and end with n-1, where n is the number of control plane machines in the cluster. The DNS record must resolve to an unicast IPv4 address for the control plane machine, and the records must be resolvable from all the nodes in the cluster.

_etcd-server-ssl._tcp.<cluster_name>.<base_domain>

For each control plane machine, OpenShift Container Platform also requires a SRV DNS record for etcd server on that machine with priority 0, weight 10 and port 2380. A cluster that uses three control plane machines requires the following records:

# _service._proto.name.                            TTL    class SRV priority weight port target.
_etcd-server-ssl._tcp.<cluster_name>.<base_domain>  86400 IN    SRV 0        10     2380 etcd-0.<cluster_name>.<base_domain>.
_etcd-server-ssl._tcp.<cluster_name>.<base_domain>  86400 IN    SRV 0        10     2380 etcd-1.<cluster_name>.<base_domain>.
_etcd-server-ssl._tcp.<cluster_name>.<base_domain>  86400 IN    SRV 0        10     2380 etcd-2.<cluster_name>.<base_domain>.

########
Should be (missing periods at the end of record name in both instructions and example):
########

etcd

etcd-<index>.<cluster_name>.<base_domain>.

OpenShift Container Platform requires DNS records for each etcd instance to point to the control plane machines that host the instances. The etcd instances are differentiated by <index> values, which start with 0 and end with n-1, where n is the number of control plane machines in the cluster. The DNS record must resolve to an unicast IPv4 address for the control plane machine, and the records must be resolvable from all the nodes in the cluster.

_etcd-server-ssl._tcp.<cluster_name>.<base_domain>.

For each control plane machine, OpenShift Container Platform also requires a SRV DNS record for etcd server on that machine with priority 0, weight 10 and port 2380. A cluster that uses three control plane machines requires the following records:

# _service._proto.name.                            TTL    class SRV priority weight port target.
[...]

PR #20505, opened 2020-03-16
Adjust DNS records to include terminal period

jboxman:
- https://bugzilla.redhat.com/show_bug.cgi?id=1770960

ref: https://github.com/openshift/openshift-docs/issues/18027

### DIFF
--- a/modules/installation-dns-user-infra.adoc
+++ b/modules/installation-dns-user-infra.adoc
@@ -15,7 +15,7 @@
 The following DNS records are required for an {product-title} cluster that uses
 user-provisioned infrastructure. In each record, `<cluster_name>` is the cluster
 name and `<base_domain>` is the cluster base domain that you specify in the
-`install-config.yaml` file.
+`install-config.yaml` file. A complete DNS record takes the form: `<component>.<cluster_name>.<base_domain>.`.
 
 .Required DNS records
 [cols="1a,5a,3a",options="header"]
@@ -26,12 +26,12 @@ name and `<base_domain>` is the cluster base domain that you specify in the
 |Description
 
 .2+a|Kubernetes API
-|`api.<cluster_name>.<base_domain>`
+|`api.<cluster_name>.<base_domain>.`
 |This DNS A/AAAA or CNAME record must point to the load balancer
 for the control plane machines. This record must be resolvable by both clients
 external to the cluster and from all the nodes within the cluster.
 
-|`api-int.<cluster_name>.<base_domain>`
+|`api-int.<cluster_name>.<base_domain>.`
 |This DNS A/AAAA or CNAME record must point to the load balancer
 for the control plane machines. This record must be resolvable from all the
 nodes within the cluster.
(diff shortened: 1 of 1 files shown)


jboxman:
Preview: http://file.bos.redhat.com/jboxman/20200319/bz-1770960/installing/installing_vsphere/installing-vsphere.html#installation-dns-user-infra_installing-vsphere

modules/installation-dns-user-infra.adoc
```diff
@@ -67,9 +67,9 @@ following records:
 [options="nowrap"]
 ----
 # _service._proto.name.                            TTL    class SRV priority weight port target.
-_etcd-server-ssl._tcp.<cluster_name>.<base_domain>  86400 IN    SRV 0        10     2380 etcd-0.<cluster_name>.<base_domain>.
-_etcd-server-ssl._tcp.<cluster_name>.<base_domain>  86400 IN    SRV 0        10     2380 etcd-1.<cluster_name>.<base_domain>.
-_etcd-server-ssl._tcp.<cluster_name>.<base_domain>  86400 IN    SRV 0        10     2380 etcd-2.<cluster_name>.<base_domain>.
+_etcd-server-ssl._tcp.<cluster_name>.<base_domain>.  86400 IN    SRV 0        10     2380 etcd-0.<cluster_name>.<base_domain>.
```
zhaozhanqi: I do not think we need the `.` at the end of `etcd-0.<cluster_name>.<base_domain>.`
@lihongan ^^

jboxman: @zhaozhanqi, does vSphere have some kind of DNS service? Where are you creating the DNS record?

I want to be sure that the issue isn't one of different DNS services expecting different input; Maybe one requires the terminal period, and another rejects it or accepts input both with and without the terminal period.

zhaozhanqi: we are using AWS route53 for vsphere cluster testing @jiajliu ^^ could you help confirm if I'm correct for above DNS records.  thanks

modules/installation-dns-user-infra.adoc
```diff
@@ -43,22 +43,22 @@ API calls can fail, and you cannot retrieve logs from Pods.
 ====
 
 |Routes
-|`*.apps.<cluster_name>.<base_domain>`
+|`*.apps.<cluster_name>.<base_domain>.`
```
lihongan: The update is correct, the `.` at the end of domain is required.

zhaozhanqi: yes, the name should have `.` at the end. but seems the `target` part of etcd SRV record `etcd-0.<cluster_name>.<base_domain>.` should not contain `.` at the end.
`_etcd-server-ssl._tcp.<cluster_name>.<base_domain>.  86400 IN    SRV 0        10     2380 etcd-0.<cluster_name>.<base_domain>.`

lihongan: yes, seems not required for target part.

jboxman:
@zhaozhanqi @lihongan 

Updated.

mburke5678:
@jboxman LGTM

PR #20517, opened 2020-03-17
Fixing backticks that were rendering

bergerhoffer:


_topic_map.yml
```diff
@@ -1345,7 +1345,7 @@ Topics:
     File: preparing-cluster-for-cnv
   - Name: Installing container-native virtualization
     File: installing-container-native-virtualization
-  - Name: Installing the `virtctl` client
+  - Name: Installing the virtctl client
```
bergerhoffer: @ousleyp @aburdenthehand Backticks don't get replaced properly in the TOC and they're rendering, so I'm removing. Let me know if you have any concerns with that, thanks!

ousleyp: Good catch, thanks!

modules/nodes-cluster-resource-configure-jdk.adoc
```diff
@@ -91,13 +91,13 @@ that the right settings are being passed to the right JVM.
 
 The `JAVA_TOOL_OPTIONS` environment variable is always respected by the OpenJDK,
 and values specified in `JAVA_TOOL_OPTIONS` will be overridden by other options
-specified on the JVM command line. By default,  to ensure that these options are 
+specified on the JVM command line. By default,  to ensure that these options are
```
jboxman: _By default,  to_ extra white space

jboxman (approved):
Did find one double spacing between two words, but otherwise LGTM.

modules/nodes-cluster-resource-configure-jdk.adoc
bergerhoffer: Sure, happy to fix that as well.

PR #20532, opened 2020-03-17
quick ifdef update

lamek:
@sheriff-rh Quick ifdef update we discussed. PTAL.


sheriff-rh:
yep, nailed it! good work @lamek !

sheriff-rh (approved):
LGTM!

Issue #20541, opened 2020-03-18
[enterprise-4.3] Edit suggested in file logging/cluster-logging-uninstall.adoc


m3adow:
### Which section(s) is the issue in?

In the second point of the "Uninstalling cluster logging from OpenShift Container Platform" section.

### What needs fixing?

There is a strange symbol at the end of the PVC deletion shell command, `￼` which has to be removed.

codyhoag:
@m3adow I didn't see any strange symbols on my end. Can you provide any additional info (or screenshot) describing this issue?

This is a screenshot from https://docs.openshift.com/container-platform/4.3/logging/cluster-logging-uninstall.html when I viewed it using Chrome:

![Screenshot from 2020-03-18 10-02-50](https://user-images.githubusercontent.com/1738558/76968786-d7852d80-68ff-11ea-9615-bbe4b92b87d7.png)

m3adow:
I'm using Firefox, this is my view:

![image](https://user-images.githubusercontent.com/1872304/76972777-5c3b7000-692f-11ea-898f-25d5161efa73.png)


I tested it with my Firefox installation, a Private Browsing Window and a Selenium installation of Firefox, so I assume it's a Firefox issue.

codyhoag:
Very weird. My Firefox browser doesn't show that character either. There is whitespace at the end of that command. I'll remove it and republish and we'll see if that fixes it.

codyhoag:
@m3adow the fix should be published now. Can you confirm the strange character is removed?

https://docs.openshift.com/container-platform/4.3/logging/cluster-logging-uninstall.html

m3adow:
Looks good to me.

PR #20550, opened 2020-03-18
Potential special char in command

codyhoag:
Suggested in #20541


codyhoag:
Simple whitespace/special char removal. No peer review necessary.

PR #20599, opened 2020-03-19
Added broker injection to kn trigger docs

abrennan89:


dsimansk:
I can't make up my mind if there should be explicit mention of that annotation value `knative-eventing-injection: enabled` or not. Personally I like the details, but it's pretty low-level one. :)

Except the minor detail above, It's good from my point of view :+1:

serverless/knative_eventing/serverless-kn-trigger.adoc
```diff
@@ -18,8 +18,25 @@ Before you can use triggers, you will need:
...
+.Procedure
+
+To create a trigger, enter the following command:
+----
+$ kn trigger create <TRIGGER-NAME> --broker <BROKER-NAME> --filter <KEY=VALUE> --sink <SINK>
+----
+
+To create a trigger and also create the `default` broker using broker injection, enter the following command:
+----
+$ kn trigger create <TRIGGER-NAME> --broker --inject-broker --filter <KEY=VALUE> --sink <SINK>
```
matzew: :+1:

matzew (approved):
/lgtm
/approve

serverless/knative_eventing/serverless-kn-trigger.adoc
maschmid: This would fail with " Invalid value: "--inject-broker"" ,  because the --broker expects the name of the broker, If we want to use the default, we need to omit the "--broker" flag altogether.

kn trigger create <TRIGGER-NAME> --inject-broker --filter <KEY=VALUE> --sink <SINK>

maschmid: or, alternatively, include the "--broker default" as well:

kn trigger create --broker default --inject-broker --filter <KEY=VALUE> --sink

maschmid:
QE approved

abrennan89:
Changes apply for 4.3 and 4.4

modules/serverless-creating-broker-admin.adoc
```diff
@@ -3,17 +3,22 @@
...
 
 If you have cluster administrator permissions, you can create a broker automatically by annotating a namespace.
 
 .Prerequisites
 * Knative Eventing installed.
 * Cluster administrator permissions for {product-title}.
 
+[NOTE]
+====
+If you do not have cluster administrator permissions, you must create the broker using the manual method outlined in xref:../modules/serverless-creating-broker.adoc#serverless-creating-broker[Creating a broker manually].
```
ahardin-rh: We cannot include xrefs in modules 🙁

abrennan89: OOF, forgot about this. No worries, I can remove it since I mention this in the assembly now anyway. Good catch and thank you!

ahardin-rh:
@abrennan89 Just one comment from me. Everything else looks good!

PR #20616, opened 2020-03-20
Fix style for options that are not compulsory

jboxman:
So IBM says:

> Indicating optional and conditional steps
If a step is optional, begin the step with the word Optional. If your authoring tool provides a semantic element to generate the word Optional, use that element.

> Example
> 1. Optional: Click Set password. In the Password field, type a password.

Do we care about a period versus a colon for this? Do we care about _optional_ versus _optionally_?

And do we want to offset this semantically, with italics or bold? (We do no, currently.)

### DIFF
--- a/modules/cluster-logging-collector-fluentd.adoc
+++ b/modules/cluster-logging-collector-fluentd.adoc
@@ -191,7 +191,7 @@ For example:
 $ oc create configmap secure-forward --from-file=secure-forward.conf -n openshift-logging
 ----
 
-. Optionally, import any secrets required for the receiver:
+. Optional: Import any secrets required for the receiver:
 +
 ----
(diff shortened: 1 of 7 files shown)


kalexand-rh:
> So IBM says:
> 
> > Indicating optional and conditional steps
> > If a step is optional, begin the step with the word Optional. If your authoring tool provides a semantic element to generate the word Optional, use that element.
> 
> > Example
> > 
> > 1. Optional: Click Set password. In the Password field, type a password.
> 
> Do we care about a period versus a colon for this? Do we care about _optional_ versus _optionally_?

We should be consistent where we can.
> 
> And do we want to offset this semantically, with italics or bold? (We do no, currently.)

No. That would just add more overhead.

kalexand-rh (commented):
LGTM

jboxman:
@mburke5678 @apinnick @maximsvistunov 

FYI, I replaced instances of _Optionally, _ with _Optional: _ per IBM style guide for what appear to be procedures. Let me know if that isn't correct for your modules.

Thanks!

PR #20617, opened 2020-03-20
Ensure procedure steps are sequential

jboxman:
The procedures were missing some `+` in critical places.


ahardin-rh:
Looks great!

PR #20631, opened 2020-03-23
cnv: avoid documenting cnv-bridge-cni

dankenigsberg:
As of OpenShift 4.3, the standard bridge-cni and tuning-cni are support
vlan filtering and static mac address, respectfully. We should direct
user to use these standard CNIs and not the CNV-crafted ones.

Signed-off-by: Dan Kenigsberg <danken@redhat.com>

### DIFF
--- a/modules/cnv-creating-bridge-nad-cli.adoc
+++ b/modules/cnv-creating-bridge-nad-cli.adoc
@@ -6,7 +6,7 @@
 = Creating a Linux bridge NetworkAttachmentDefinition in the CLI
 
 As a network administrator, you can configure a NetworkAttachmentDefinition
-of type `cnv-bridge` to provide Layer-2 networking to Pods and virtual machines.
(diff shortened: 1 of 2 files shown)


dankenigsberg:
@duyanyan would you please keep me honest here, verifying that there is no need to use the cnv-crafted CNIs?

dankenigsberg:
@phoracek please approve this PR (or reject it if you think it is wrong)

ousleyp:
LGTM. I will merge if @phoracek and/or @duyanyan approve :)

ousleyp (approved):

phoracek:
If we want to switch to OpenShift's bridge and tuning plugins, we have to fully honor their API which is not based around NetworkAttachmentDefinition, but around their Network operator. That would require changes in our UI and in docs.

I'm in favor of switching to OCP's CNIs, but this PR would be at the end of a bigger effort. We should not merge this at this point.

PR #20640, opened 2020-03-23
issue 18392: vCPU limit, not instance limit

### LINKED ISSUE #18392: [enterprise-4.2] Edit suggested in modules/installation-aws-limits.adoc
The "Instance limits" row needs to be replaced with a "vCPU limits" row based on an AWS change to limit allocation.

See https://github.com/openshift/installer/issues/2736

Because the change appears to be valid in all accounts, it should apply to 4.1+.

kalexand-rh:
Fixes #18392

@jianlinliu, will you PTAL?

### DIFF
--- a/modules/installation-aws-limits.adoc
+++ b/modules/installation-aws-limits.adoc
@@ -16,29 +16,28 @@ to request additional resources for you AWS account.
 The following table summarizes the AWS components whose limits can impact your
 ability to install and run {product-title} clusters.
 
-
 [cols="2a,3a,3a,8a",options="header"]
 |===
 |Component |Number of clusters available by default| Default AWS limit |Description
 
-|Instance Limits
-|Varies
-|Varies
+|vCPU limits
+|0
+|1 vCPU
 |By default, each cluster creates the following instances:
 
 * One bootstrap machine, which is removed after installation
 * Three master nodes
 * Three worker nodes
 
-These instance type counts are within a new account's default limit. To deploy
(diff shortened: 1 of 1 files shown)


modules/installation-aws-limits.adoc
```diff
@@ -16,30 +16,29 @@ to request additional resources for you AWS account.
...
 [cols="2a,3a,3a,8a",options="header"]
 |===
 |Component |Number of clusters available by default| Default AWS limit |Description
 
-|Instance Limits
-|Varies
-|Varies
+|vCPU limits
+|0
+|1
```
jianlinliu: Here "Number of clusters available by default" - 0, "Default AWS limit" is right? Or just some place holder? I think we should input some meaningful value here.

modules/installation-aws-limits.adoc
```diff
@@ -16,30 +16,29 @@ to request additional resources for you AWS account.
...
 [cols="2a,3a,3a,8a",options="header"]
 |===
 |Component |Number of clusters available by default| Default AWS limit |Description
 
-|Instance Limits
-|Varies
-|Varies
+|vCPU limits
+|0
+|1 vCPU
```
jianlinliu: Here "Number of clusters available by default" - 0, "Default AWS limit" - 1, but the following description is saying "Each default cluster requires 20 vCPU", and I do not think default aws limit only allow 1 vCPU.

PR #20641, opened 2020-03-23
FIXME: modules/customize-certificates-replace-default-router: Include egress CAs

wking:
Explain that you may not be able to use *just* the ingress CA(s) for the proxy's trusted CAs.

CC @Miciah

### DIFF
--- a/modules/customize-certificates-replace-default-router.adoc
+++ b/modules/customize-certificates-replace-default-router.adoc
@@ -19,7 +19,7 @@ both in the PEM format, for use.
 
 .Procedure
 
-. Create a ConfigMap that includes the certificate authority used to signed the new certificate:
+. Create a ConfigMap that includes the certificate authority used to sign the new certificate:
 +
 ----
 $ oc create configmap custom-ca \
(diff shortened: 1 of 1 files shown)


modules/customize-certificates-replace-default-router.adoc
```diff
@@ -19,14 +19,16 @@ both in the PEM format, for use.
...
+. Create a ConfigMap that includes the certificate authority used to sign the new certificate:
 +
 ----
 $ oc create configmap custom-ca \
      --from-file=ca-bundle.crt=</path/to/example-ca.crt> \//<1>
      -n openshift-config
 ----
 <1> `</path/to/example-ca.crt>` is the path to the certificate authority file on your local file system.
++
+Including the ingress certificate authority supports cluster components like OAuth reaching other cluster components via public routes. If you are also using a proxy for egress (FIXME: link to https://docs.openshift.com/container-platform/4.3/networking/enable-cluster-wide-proxy.html ), the `ca-bundle.crt` value should include both the ingress and egress CAs.
```
kalexand-rh: @wking, you can't add a link here. You could add a link in the parent assembly in a prereq.
```suggestion
Including the ingress certificate authority supports cluster components like OAuth reaching other cluster components via public routes. If you configured a cluster-wide proxy for egress, the `ca-bundle.crt` value must include both the ingress and egress CAs.
```

PR #20665, opened 2020-03-24
Remove Tenant Driven Snapshotting from TP Tracker Table

bobfuru:
[BZ1813156](https://bugzilla.redhat.com/show_bug.cgi?id=1813156)
This was a holdover from 3.x docs and is not accurate. It is also being removed from newer 4.x Release Notes versions.

@liangxia PTAL

### DIFF
--- a/release_notes/ocp-4-1-release-notes.adoc
+++ b/release_notes/ocp-4-1-release-notes.adoc
@@ -267,9 +267,7 @@ xref:../applications/operators/olm-understanding-operatorhub.adoc#olm-understand
 [id="ocp-4-1-storage"]
 === Storage
 
-Storage support in {product-title} 4.1 is the same as {product-title} 3.11 with
-the exception of the following available in Technology Preview: EFS (CSI Driver
-handled via Amazon), Manila provisioner/operator, and Snapshot.
+Storage support in {product-title} 4.1 is the same as {product-title} 3.11.
 
 [id="ocp-4-1-scale"]
 === Scale
@@ -558,10 +556,6 @@ features marked *GA* indicate _General Availability_.
 |GA* footnoteref:[disclaimer, Features marked with `*` indicate delivery in a z-stream patch.]
 |GA
 
-|Tenant Driven Snapshotting
-|TP
-|TP
-
 |`oc` CLI Plug-ins
 |TP
 |TP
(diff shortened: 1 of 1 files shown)


jboxman:
@bobfuru LGTM

liangxia:
@qinpingli  PTAL

qinpingli:
Hi @bobfuru ,

There is still some description about snapshot: https://github.com/openshift/openshift-docs/blob/0b0ceec80b1ede5650f029786e4595a3a00f1862/release_notes/ocp-4-1-release-notes.adoc#storage

I checked with the team, seems there is no difference from OCP 3.11 for end users.

If so, the content about `manila` and `efs` need to be removed from the table `Technology Preview Tracker` too. Do you like we create a new bug for this or update in the patch?

bobfuru:
> Hi @bobfuru ,
> 
> There is still some description about snapshot: https://github.com/openshift/openshift-docs/blob/0b0ceec80b1ede5650f029786e4595a3a00f1862/release_notes/ocp-4-1-release-notes.adoc#storage
> 
> I checked with the team, seems there is no difference from OCP 3.11 for end users.
> 
> If so, the content about `manila` and `efs` need to be removed from the table `Technology Preview Tracker` too. Do you like we create a new bug for this or update in the patch?

Hi @qinpingli - Yes, could you please file this as a new bug against 4.1 and I will take a look? Thanks.

bobfuru:
> Hi @bobfuru ,
> 
> There is still some description about snapshot: https://github.com/openshift/openshift-docs/blob/0b0ceec80b1ede5650f029786e4595a3a00f1862/release_notes/ocp-4-1-release-notes.adoc#storage
> 
> I checked with the team, seems there is no difference from OCP 3.11 for end users.
> 
> If so, the content about `manila` and `efs` need to be removed from the table `Technology Preview Tracker` too. Do you like we create a new bug for this or update in the patch?

Hi @qinpingli - good catch. You are correct that there is no difference between 3.11 and 4.1. What actually should happen is that the Storage description shouldn't mention EFS, Manila, or Snapshots as it did because there was no change between 3.11 and 4.1. So, I've removed that part of the description.

I have left EFS and Manila in the tracker table because they were TP in both 3.11 and 4.1, so those rows should be correct. 

As with 4.2 and 4.3 branches, I did remove "Tenant Driven Snapshotting" from the 4.1 table to avoid confusion. (Since 3.11, this feature has been referred to as "PV Snapshots", which is deprecated in 4.2 and will be introduced as "CSI Snapshots" in a later release).

Hope that all makes sense! If so, please respond with lgtm. Thanks!

qinpingli:
Hi @bobfuru
Checked the table again, it's good, thx.

PR #20673, opened 2020-03-24
OSDOCS-939 three-node clusters

bmcelvee:
https://issues.redhat.com/browse/OSDOCS-939

### DIFF
--- a/installing/installing_aws/installing-aws-user-infra.adoc
+++ b/installing/installing_aws/installing-aws-user-infra.adoc
@@ -59,6 +59,8 @@ include::modules/installation-generate-aws-user-infra-install-config.adoc[levelo
 
 include::modules/installation-configure-proxy.adoc[leveloffset=+2]
 
+//include::modules/installation-three-node-cluster.adoc[leveloffset=+2]
+
 include::modules/installation-user-infra-generate-k8s-manifest-ignition.adoc[leveloffset=+2]
 
 include::modules/installation-extracting-infraid.adoc[leveloffset=+1]
--- a/modules/installation-three-node-cluster.adoc
+++ b/modules/installation-three-node-cluster.adoc
@@ -1,0 +1,29 @@
+// Module included in the following assemblies:
+//
+// * installing/installing_aws/installing-aws-user-infra.adoc [Eventually]
+// * installing/installing_azure/installing-azure-user-infra.adoc [Eventually]
+// * installing/installing_gcp/installing-gcp-user-infra.adoc [Eventually]
+// * installing/installing_gcp/installing-restricted-networks-gcp.adoc [Eventually]
+// * installing/installing_bare_metal/installing-bare-metal.adoc
+// * installing/installing_aws/installing-restricted-networks-aws.adoc [Eventually]
+// * installing/installing_bare_metal/installing-restricted-networks-bare-metal.adoc
+// * installing/installing_vsphere/installing-restricted-networks-vsphere.adoc [Eventually]
+// * installing/installing_vsphere/installing-vsphere.adoc [Eventually]
+// * installing/installing_ibm_z/installing-ibm-z.adoc [Eventually]
+
+[id="installation-three-node-cluster_{context}"]
+= Running a three-node cluster
+
+You can install and run three-node clusters in {product-title} with no workers. This provides smaller, more resource efficient clusters for cluster administrators and developers to use for deployment, development, and testing.
+
+.Procedure
+
+* Edit the `install-config.yaml` file to set the number of compute replicas, which are also known as worker replicas, to `0`, as shown in the following `compute` stanza:
++
+[source,yaml]
+----
+- compute:
+    name: worker
+    platform: {}
+    replicas: 0
+----
--- a/installing/installing_aws/installing-restricted-networks-aws.adoc
+++ b/installing/installing_aws/installing-restricted-networks-aws.adoc
@@ -76,6 +76,8 @@ include::modules/installation-generate-aws-user-infra-install-config.adoc[levelo
 
 include::modules/installation-configure-proxy.adoc[leveloffset=+2]
 
+//include::modules/installation-three-node-cluster.adoc[leveloffset=+2]
+
 include::modules/installation-user-infra-generate-k8s-manifest-ignition.adoc[leveloffset=+2]
 
 include::modules/installation-extracting-infraid.adoc[leveloffset=+1]
--- a/installing/installing_azure/installing-azure-user-infra.adoc
+++ b/installing/installing_azure/installing-azure-user-infra.adoc
@@ -83,6 +83,7 @@ include::modules/ssh-agent-using.adoc[leveloffset=+1]
 include::modules/installation-user-infra-generate.adoc[leveloffset=+1]
 include::modules/installation-initializing.adoc[leveloffset=+2]
 include::modules/installation-configure-proxy.adoc[leveloffset=+2]
+//include::modules/installation-three-node-cluster.adoc[leveloffset=+2]
 include::modules/installation-user-infra-exporting-common-variables-arm-templates.adoc[leveloffset=+2]
 include::modules/installation-user-infra-generate-k8s-manifest-ignition.adoc[leveloffset=+2]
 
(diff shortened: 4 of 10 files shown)


bmcelvee:
@wking would you mind reviewing this PR please? Here's the aws user-provisioned install preview: https://osdocs-939-3nodebaremetal--ocpdocs.netlify.app/openshift-enterprise/latest/installing/installing_aws/installing-aws-user-infra.html#installation-three-node-cluster_installing-aws-user-infra

 It's my understanding from the comments in https://issues.redhat.com/browse/NE-247 that setting the .compute[].replicas to 0 in your install-config.yaml file is the procedure for installing 3-node clusters, but we have it documented as the general install procedure: 
https://docs.okd.io/latest/installing/installing_aws/installing-aws-user-infra.html#installation-generate-aws-user-infra-install-config_installing-aws-user-infra

I still included it in the new module, but I feel like I'm missing something. Thanks! 

@knobunc, @ironcladlou - tagging you for awareness.

bmcelvee:
@lihongan please review this PR when you get a chance. Also - this is tied to Network Edge so I figured I'd reach out to you for QE review, but if it's not in your area please let me know.

Thanks!

installing/installing_aws/installing-aws-user-infra.adoc
```diff
@@ -59,6 +59,8 @@ include::modules/installation-generate-aws-user-infra-install-config.adoc[levelo
 
 include::modules/installation-configure-proxy.adoc[leveloffset=+2]
 
+include::modules/installation-three-node-cluster.adoc[leveloffset=+2]
+
```
lihongan, line 63: Since the title of the story is `NE-247 Support 3-node bare metal clusters with the routers` (see also: https://issues.redhat.com/browse/NE-247) , and QE just ensured  this can be supported on **Bare Metal** , so I don't think we should add it to other platform like aws, azure etc.

modules/installation-three-node-cluster.adoc
```diff
@@ -0,0 +1,50 @@
+// Module included in the following assemblies:
+//
+// * installing/installing_aws/installing-aws-user-infra.adoc
+// * installing/installing_azure/installing-azure-user-infra.adoc
+// * installing/installing_gcp/installing-gcp-user-infra.adoc
+// * installing/installing_gcp/installing-restricted-networks-gcp.adoc [Eventually?]
+// * installing/installing_bare_metal/installing-bare-metal.adoc
+// * installing/installing_aws/installing-restricted-networks-aws.adoc
+// * installing/installing_bare_metal/installing-restricted-networks-bare-metal.adoc
```
lihongan, line 9: For now we just support it on bare metal.

modules/installation-three-node-cluster.adoc
```diff
@@ -0,0 +1,50 @@
...
+// * installing/installing_vsphere/installing-restricted-networks-vsphere.adoc
+// * installing/installing_vsphere/installing-vsphere.adoc
+// * installing/installing_ibm_z/installing-ibm-z.adoc [Eventually?]
+
+[id="installation-three-node-cluster_{context}"]
+= Running a three-node cluster
+
```
lihongan, line 16: @gpei Could you help review this part?

modules/installation-three-node-cluster.adoc
```diff
@@ -0,0 +1,50 @@
...
+See the link:https://access.redhat.com/support/offerings/techpreview/[Red Hat
+Technology Preview features support scope] for more information.
+endif::[]
+====
+
+////
+.Prerequisites
+
+* You must create at least two compute machines, which are also known as worker machines, during installation.
```
lihongan, line 37: No sure why we need extra compute/worker machines during installation. From my understanding, three-node cluster means that the three nodes/machines have both master and worker role.

modules/installation-three-node-cluster.adoc
```diff
@@ -0,0 +1,50 @@
...
+////
+
+.Procedure
+
+* Edit the `install-config.yaml` file to set the number of compute replicas, which are also known as worker replicas, to `0`, as shown in the following `compute` stanza:
++
+[source,yaml]
+----
+compute:
+  name: worker
```
gpei, line 47: The first line of the compute section should begin with a hyphen "-"

installing/installing_aws/installing-aws-user-infra.adoc
bmcelvee: Sounds good! I'll comment out the includes for everything but bare metal. It will only build in bare metal, but the other includes will remain in place for when they're available.

modules/installation-three-node-cluster.adoc
bmcelvee: Commented out of the others and added [Eventually] to the list

bmcelvee:
@kalexand-rh tagging you for awareness on this one. Thanks!

bmcelvee:
@lihongan and @gpei thank you for the review! I mentioned this in an earlier comment, but I'm still a bit confused on the install-config.yaml file being the same for general install and running a three-node cluster. Do you know what the differentiating factor is there?

Earlier comment: 
"It's my understanding from the comments in https://issues.redhat.com/browse/NE-247 that setting the .compute[].replicas to 0 in your install-config.yaml file is the procedure for installing 3-node clusters, but we have it documented as the general install procedure:
https://docs.okd.io/latest/installing/installing_aws/installing-aws-user-infra.html#installation-generate-aws-user-infra-install-config_installing-aws-user-infra"

@smarterclayton you left feedback on the NE epic, would you mind taking a look too? Thanks!

bmcelvee:
Will also require changes to: 
https://docs.openshift.com/container-platform/4.3/installing/installing_bare_metal/installing-bare-metal.html#creating-machines-bare-metal
and
https://docs.openshift.com/container-platform/4.3/installing/installing_bare_metal/installing-bare-metal.html#installation-user-infra-generate-k8s-manifest-ignition_installing-bare-metal

gpei:
> @lihongan and @gpei thank you for the review! I mentioned this in an earlier comment, but I'm still a bit confused on the install-config.yaml file being the same for general install and running a three-node cluster. Do you know what the differentiating factor is there?

In the UPI install, after we setting .compute[].replicas to 0, the masters would become schedulable which is not expected in our cluster, so we'll also set mastersSchedulable to False in the following step:  https://docs.okd.io/latest/installing/installing_aws/installing-aws-user-infra.html#installation-user-infra-generate-k8s-manifest-ignition_installing-aws-user-infra . And then we'll create the dedicated workers manually.

For the three-node cluster, we'll allow the masters to be schedulable AFAIK. This is something I can think of.

bmcelvee:
Thank you, @gpei! I added a note in the Creating the Kubernetes manifest and Ignition config files section to skip setting the `mastersSchedulable` parameter to `False` if running a three-node cluster.

bmcelvee:
@openshift/team-documentation PTAL? Thanks!

modules/installation-three-node-cluster.adoc
```diff
@@ -0,0 +1,44 @@
...
+// * installing/installing_aws/installing-restricted-networks-aws.adoc [Eventually]
+// * installing/installing_bare_metal/installing-restricted-networks-bare-metal.adoc
+// * installing/installing_vsphere/installing-restricted-networks-vsphere.adoc [Eventually]
+// * installing/installing_vsphere/installing-vsphere.adoc [Eventually]
+// * installing/installing_ibm_z/installing-ibm-z.adoc [Eventually]
+
+[id="installation-three-node-cluster_{context}"]
+= Running a three-node cluster
+
+[IMPORTANT]
```
bergerhoffer, line 17: Are you able to use the `technology-preview.adoc` file here instead of copy/pasting the important note?

modules/installation-three-node-cluster.adoc
```diff
@@ -0,0 +1,44 @@
...
+not recommend to use them for production. These features provide early access to
+upcoming product features, enabling customers to test functionality and provide
+feedback during the development process.
+
+See the link:https://access.redhat.com/support/offerings/techpreview/[Red Hat
+Technology Preview features support scope] for more information.
+endif::[]
+====
+
+You can install and run a three-node clusters in {product-title}. This provides smaller, more resource efficient clusters for cluster administrators and developers to use for deployment, development, and testing.
```
bergerhoffer, line 32: Either "run a three-node cluster" or "run three-node clusters"

bergerhoffer: Also just wondering whether you can give any more info on exactly what three-node clusters consist of. I assume per the procedure below that it's 3 masters and no workers, but might be helpful to point out here.

bergerhoffer (commented):
Just a few comments

PR #20675, opened 2020-03-25
Follow-up Edits to log forwarding per email

mburke5678:
Follow-up to https://github.com/openshift/openshift-docs/pull/20108#discussion_r397512074


mburke5678:
@anpingli PTAL

anpingli (commented):
> If you set disableDefaultForwarding: true you must configure a pipeline and output for all three types of logs, application, infrastructure, and audit. If you do not specify a pipeline and output for a log type, those logs are not stored and will be lost.

I don't see any difference between disableDefaultForwarding: true and disableDefaultForwarding: false.  @jcantrill Is this a bug?

PR #20683, opened 2020-03-25
Cnv rename

kalexand-rh:
These terms still need to be checked/updated:

cnv-bridge
openshift-cnv
cnv-2.3-for-rhel-8-x86_64-rpms
*{VirtProductName}*


kalexand-rh:
@aburdenthehand @ousleyp  @vikram-redhat,  PTAL?

ousleyp (commented):
This LGTM other than the things you called out in the description, which we can update as they become clear (and as part of our pre-release checklist for 2.4). We'll also need to check/update the "CNV Operator Deployment" tab name and other instances where the name appears in the UI, as well as the links to the Red Hat Container Catalog (virtIO drivers, etc.).

I think we can merge this.

 @aburdenthehand @vikram-redhat

kalexand-rh:
Then I'm going to merge. Thanks!

vikram-redhat:
Thanks for this @kalexand-rh !

PR #20685, opened 2020-03-25
Release Notes 4.4 entries

codyhoag:


release_notes/ocp-4-4-release-notes.adoc
```diff
@@ -67,7 +67,8 @@ Red Hat to assist in the deployment process, or create your own. You are also
 free to create the required resources through other methods; the ARM templates
 are just an example.
 
-//See [link] for details.
+See xref:../installing/installing_azure/installing-azure-user-infra.adoc#installing-azure-user-infra[Installing a cluster on Azure using ARM templates]
```
jboxman: Would we want to define _ARM_? I'm thinking the CPU architecture, but that can't be correct in this instance.

release_notes/ocp-4-4-release-notes.adoc
```diff
@@ -106,6 +107,33 @@ the original scheduling decisions are no longer appropriate for certain nodes.
...
+[id="ocp-4-4-logging"]
+=== Logging
+
+[id="ocp-4-4-sending-cluster-logs-using-fluent-syslog-plug-in"]
+==== Sending cluster logs using the Fluentd syslog plug-in (RFC 3164)
+
+Due to changes introduced with the Log Forwarding feature in {product-title}
+4.3, you could no longer use the Fluentd syslog plug-in to forward logs to an
+external syslog server. In {product-title} 4.4, this functionality is restored
+and you can use the syslog plug-in. Note that the procedure for configuring the
```
jboxman: s/Note that the procedure/The procedure/

Changed since when, though? 4.3?

release_notes/ocp-4-4-release-notes.adoc
```diff
@@ -106,6 +107,33 @@ the original scheduling decisions are no longer appropriate for certain nodes.
...
+plug-in has changed. For more information, see
+xref:../logging/config/cluster-logging-external.adoc#cluster-logging-collector-syslog_cluster-logging-external[Sending logs using the Fluentd syslog plug-in (RFC 3164)].
+
+[id="ocp-4-4-networking"]
+=== Networking
+
+[id="ocp-4-4-sctp-on-ocp"]
+==== Stream Control Transmission Protocol (SCTP) on {product-title}
+
+If your cluster is installed on bare metal infrastructure, you can enable SCTP
```
jboxman: Not sure if I should peer review my own text, but maybe this instead:

> SCTP is a reliable message based protocol that runs on top of an IP network.
> If your cluster is installed on bare metal infrastructure, you can enable SCTP on the hosts in the cluster.
> When enabled, you can use SCTP as a protocol with both Pods and Services.

jboxman (changes requested):
Few comments, thanks!

release_notes/ocp-4-4-release-notes.adoc
codyhoag: It was actually defined earlier in that entry; it didn't show up as new in this PR since I only added the new link. It's for "Azure Resource Manager".

release_notes/ocp-4-4-release-notes.adoc
codyhoag: Haha! I'll go with your new text.

jboxman:
Cool, LGTM

PR #20604, opened 2020-03-20
update sriov operator config

zenghui-shi:

PR #20686, opened 2020-03-25
Fix-up commit for SR-IOV Operator config

jboxman:
- ref: https://github.com/openshift/openshift-docs/pull/20604

### DIFF
--- a/modules/nw-sriov-configuring-operator.adoc
+++ b/modules/nw-sriov-configuring-operator.adoc
@@ -1,8 +1,8 @@
 // Module included in the following assemblies:
 //
-// * networking/multiple_networks/configuring-sr-iov.adoc
+// * networking/hardware_networks/configuring-sriov-operator.adoc
 
-[id="configuring-sr-iov-operator_{context}"]
+[id="nw-sriov-configuring-operator_{context}"]
 = Configuring the SR-IOV Network Operator
 
 [IMPORTANT]
@@ -12,46 +12,31 @@ The default configuration is recommended for most use cases.
 Complete the steps to modify the relevant configuration only if the default behavior of the Operator is not compatible with your use case.
 ====
 
-The SR-IOV Network Operator adds the `SriovOperatorConfig.sriovnetwork.openshift.io` Custom Resource Definition (CRD) to {product-title}.
-You can configure the SR-IOV software compoenents by modifying the default 
-SriovOperatorConfig Custom Resource (CR). The Operator supports the following optional behavioral changes:
-
-- Disabling the Network Resources Injector
-- Disabling the Operator admission web controller
-- Configuring a custom node selector for the SR-IOV Network Config daemon. The daemon runs on all the worker nodes by default.
+The SR-IOV Network Operator adds the `SriovOperatorConfig.sriovnetwork.openshift.io` CustomResourceDefinition resource.
+The operator automatically creates a SriovOperatorConfig custom resource (CR) named `default` in the `openshift-sriov-network-operator` namespace.
 
 [NOTE]
 =====
-There is one Custom Resource of SriovOperatorConfig being created when
-the operator is up. Change the operator configuration by modifying this
-pre-created Custom Resource. The `default` Custom Resource is created in
-`openshift-sriov-network-operator` namespace.
+The `default` CR contains the SR-IOV Network Operator configuration for your cluster.
+To change the operator configuration, you must modify this CR.
 =====
 
-`SriovOperatorConfig` provides three options for configuring operator:
-* `enableInjector` allows project administrators to enable or disable Network
-Resources Injector DaemonSet.
-* `enableOperatorWebhook` allows project adminstrators to enable or disable
-Operator Admission Controller Webhook DaemonSet.
-* `configDaemonNodeSelector` allows project adminstrators to schedule SR-IOV
-Network Config Daemon on selected nodes.
+The SriovOperatorConfig CR provides several fields for configuring the operator:
 
+* `enableInjector` allows project administrators to enable or disable the Network Resources Injector DaemonSet.
+* `enableOperatorWebhook` allows project administrators to enable or disable the Operator Admission Controller webook DaemonSet.
+* `configDaemonNodeSelector` allows project administrators to schedule the SR-IOV Network Config Daemon on selected nodes.
 
-[id="disabling-network-resource-injector_{context}"]
-== Disabling Network Resources Injector
+[id="about-network-resource-injector_{context}"]
+== About the Network Resources Injector
 
 The Network Resources Injector is a Kubernetes Dynamic Admission Controller 
-application that:
+application. It provides the following capabilities:
 
-* Mutates resource requests and limits in pod specifiction to add SR-IOV
-resource name according to SR-IOV network attachment definition annotation.
-* Mutates pod specifications with downward API volume to expose pod annotations
-and labels to the running container as files under `/etc/podnetinfo`.
+* Mutation of resource requests and limits in Pod specification to add an SR-IOV resource name according to an SR-IOV network attachment definition annotation.
+* Mutation of Pod specifications with downward API volume to expose pod annotations and labels to the running container as files under the `/etc/podnetinfo` path.
 
-By default the Network Resources Injector will be enabled by SR-IOV operator
-and runs as a Kubernetes DaemonSet on all master nodes. Below is an example
-of Network Resources Injector pods running in a cluster with three master
-nodes:
+By default the Network Resources Injector is enabled by the SR-IOV operator and runs as a DaemonSet on all master nodes. The following is an example of Network Resources Injector Pods running in a cluster with three master nodes:
 
 ----
 $ oc get pods -n openshift-sriov-network-operator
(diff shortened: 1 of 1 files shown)


jboxman:
@zshi-redhat, I'll complete edits here and cherry-pick to 4.4 and 4.3 when done.

zshi-redhat:
@jboxman thanks for the fix!

/lgtm

modules/nw-sriov-configuring-operator.adoc
```diff
@@ -103,45 +95,52 @@ Follow this procedure to disable it.
...
+
+. To update the `default` SriovOperatorConfig CR, enter the following command:
 +
 ----
-$ oc patch sriovoperatorconfig default --patch '[{"op": "replace", "path": "/spec/configDaemonNodeSelector", "value": {<node-label>}}]' --type=json -n openshift-sriov-network-operator
+$ oc patch sriovoperatorconfig default --type=json \
+  -n openshift-sriov-network-operator \
+  --patch '[{
+      "op": "replace",
+      "path": "/spec/configDaemonNodeSelector",
```
jboxman, line 137: @zshi-redhat, I've never seen this kind of `patch` before. What is a _path_? What does `/spec/configDaemonNodeSelector` mean? Is that a file on the user's file system? An xpath?

modules/nw-sriov-configuring-operator.adoc
```diff
@@ -103,45 +95,52 @@ Follow this procedure to disable it.
...
+      "value": {<node-label>}
```
jboxman, line 138: @zshi-redhat, does `value` expand out to

```
"value": {
 node-role.kubernetes.io/worker: "",
  some-other-key/worker: "",
  ...
}
```

Or can there be only 1 key provided? Or is this a different data structure than I think it is?

Thanks.

jboxman:
@zshi-redhat, I've updated the content to sync with our contributor style guide. Let me know if the content is still correct.

Thanks!

modules/nw-sriov-configuring-operator.adoc
```diff
@@ -59,32 +49,34 @@ Follow this procedure to disable it.
...
+  --type=merge -n openshift-sriov-network-operator \
+  --patch '{ "spec": { "enableInjector": false } }'
 ----
 
-. To re-enable Network Resources Injector
+. To reenable the Network Resources Injector, enter the following command:
 +
 ----
-$ oc patch sriovoperatorconfig default --patch '{ "spec": { "enableInjector": true}}' --type=merge -n openshift-sriov-network-operator
+$ oc patch sriovoperatorconfig default \
```
bergerhoffer, line 64: This procedure is confusing to me. At the end of the procedure, is the Network Resources Injector disabled, as the procedure heading implies? Or is the second step actually enabling it after step 1 disabled it? If it's the latter, then I don't think it should be a step in this procedure. It should either be its own, or mentioned after the procedure somehow.

modules/nw-sriov-configuring-operator.adoc
```diff
@@ -59,32 +49,34 @@ Follow this procedure to disable it.
...
 +
 ----
-$ oc patch sriovoperatorconfig default --patch '{ "spec": { "enableInjector": false}}' --type=merge -n openshift-sriov-network-operator
+$ oc patch sriovoperatorconfig default \
+  --type=merge -n openshift-sriov-network-operator \
+  --patch '{ "spec": { "enableInjector": false } }'
 ----
 
-. To re-enable Network Resources Injector
+. To reenable the Network Resources Injector, enter the following command:
```
bergerhoffer, line 61: I think you had it right the first time. The IBM style guide says to use hyphens when two of the same letters are next to each other.

> Generally, hyphenate a prefix only to prevent misreading or ambiguity. Specifically, hyphenate a prefix in the following situations:
> • In some words in which identical letters would be next to each other (for example, anti-intellectual, bell-like, co-opt, de-emphasize)

modules/nw-sriov-configuring-operator.adoc
```diff
@@ -59,32 +49,34 @@ Follow this procedure to disable it.
 * You must have installed the SR-IOV Operator.
 
 .Procedure
-. Update the default SriovOperatorConfig Custom Resource
+
+. To update the `default` SriovOperatorConfig CR, enter the following command:
```
bergerhoffer, line 53: Suggest being clearer on what the user is doing instead of just saying "update". 

"To disable the Network Resources Injector, set the `enableInjector` field to `false`:

modules/nw-sriov-configuring-operator.adoc
```diff
@@ -103,45 +95,52 @@ Follow this procedure to disable it.
...
```
bergerhoffer, line 99: Same comments as above for this procedure.

modules/nw-sriov-configuring-operator.adoc
```diff
@@ -103,45 +95,52 @@ Follow this procedure to disable it.
...
 ----
-$ oc patch sriovoperatorconfig default --patch '{ "spec": { "enableOperatorWebhook": true}}' --type=merge -n openshift-sriov-network-operator
+$ oc patch sriovoperatorconfig default --type=merge \
+  -n openshift-sriov-network-operator \
+  --patch '{ "spec": { "enableOperatorWebhook": true } }'
 ----
 
 [id="configuring-custom-nodeselector_{context}"]
-== Configuring Custom NodeSelector for SRIOV Network Config Daemon
+== Configuring a custom NodeSelector for SR-IOV Network Config daemon
```
bergerhoffer, line 116: Should there be a "the" in front of "SR-IOV Network Config daemon" ?

modules/nw-sriov-configuring-operator.adoc
```diff
@@ -103,45 +95,52 @@ Follow this procedure to disable it.
...
 
-The SR-IOV Network Config Daemon is in charge of discovering and configuring
-SRIOV network devices on the nodes.
+The SR-IOV Network Config daemon discovers and configures the SR-IOV network devices on cluster nodes. By default, it will be deployed to all the `worker` nodes in the cluster. You can use node labels to specify on which nodes the SR-IOV Network Config daemon runs.
```
bergerhoffer, line 118: will be -> is

modules/nw-sriov-configuring-operator.adoc
```diff
@@ -103,45 +95,52 @@ Follow this procedure to disable it.
...
 
-By default, it will be deployed to all the `worker` nodes in the cluster. You
-can follow this procedure to select the nodes, where the SR-IOV Network Config
-Daemon shall be deployed to, with node labels.
+To specify the nodes where the SR-IOV Network Config daemon is deployed, complete the following procedure.
+
+[IMPORTANT]
+=====
+When you update the `configDaemonNodeSelector` parameter, the SR-IOV Network Config daemon is recreated on each selected node.
+While the daemon is recreated, cluster users will be unable to apply any new SR-IOV Network Node Policy or create new SR-IOV pods.
```
bergerhoffer, line 125: will be -> are

modules/nw-sriov-configuring-operator.adoc
```diff
@@ -103,45 +95,52 @@ Follow this procedure to disable it.
...
+=====
 
 .Procedure
-. Update the default SriovOperatorConfig Custom Resource
+
+. To update the `default` SriovOperatorConfig CR, enter the following command:
```
bergerhoffer, line 130: A single step procedure should be an unnumbered bullet.

bergerhoffer (commented):
Just a few comments and suggestions.

jboxman:
@bergerhoffer thanks!

modules/nw-sriov-configuring-operator.adoc
```diff
@@ -1,46 +1,36 @@
...
+// * networking/hardware_networks/configuring-sriov-operator.adoc
 
-[id="configuring-sr-iov-operator_{context}"]
+[id="nw-sriov-configuring-operator_{context}"]
 = Configuring the SR-IOV Network Operator
 
-The SR-IOV Network Operator adds the `SriovOperatorConfig.sriovnetwork.openshift.io` Custom Resource Definition (CRD) to {product-title}.
-You can configure the SR-IOV software compoenents by modifying the default 
-SriovOperatorConfig Custom Resource (CR).
+The SR-IOV Network Operator adds the `SriovOperatorConfig.sriovnetwork.openshift.io` CustomResourceDefinition resource to {product-title}.
```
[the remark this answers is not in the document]
jboxman: @bergerhoffer still trying to finalize a coherent proposal for these. It presents itself as intractable.

So like our other objects, like ReplicaSet and so forth, a custom resource definition is defined by creating a CustomResourceDefinition object. (In various places on the Internet, both seem to be abbreviated as CRD, because no consistent style exists for any of this.)

I might have another proposal attempt for these soon, maybe.

This was an attempt to use `<meta.name> <ResourceName>` as an approach. So the `SriovOperatorConfig.sriovnetwork.openshift.io` CustomResourceDefinition object is how SriovOperatorConfig as a resource comes to be.

_Custom Resource Definition (CRD)_ is an artifact of my not having understood that that construction simply originated out of someone abbreviating CustomResourceDefinition as CRD and my thinking it was A Big Proper Noun Thing that was abbreviated. It's not; Developers just like to abbreviate it as CRD, _because_.

All of this is just misunderstanding, all around, I think. (CR is a little different; Creating a CustomResourceDefinition does indeed provide a new OCP custom resource, which we're abbreviating as CR.)

modules/nw-sriov-configuring-operator.adoc
jboxman: True. I consolidated it into the same procedure.

jboxman:
@zshi-redhat, I re-worked the content:

- [preview](https://sriov-operator-config-fixup--ocpdocs.netlify.app/openshift-enterprise/latest/networking/hardware_networks/configuring-sriov-operator.html)

What do you think?

Thanks!

zshi-redhat:
> @zshi-redhat, I re-worked the content:
> 
>     * [preview](https://sriov-operator-config-fixup--ocpdocs.netlify.app/openshift-enterprise/latest/networking/hardware_networks/configuring-sriov-operator.html)
> 
> 
> What do you think?
> 
> Thanks!

@jboxman thanks for updating the content, it looks good to me.

PR #20691, opened 2020-03-26
Add note regarding network policy behavior

jboxman:
- https://bugzilla.redhat.com/show_bug.cgi?id=1813846

Didn't want to let this get away. Not everyone will upgrade instantaneously when a fix is released, and one isn't available yet in any case.

### DIFF
--- a/modules/nw-networkpolicy-create.adoc
+++ b/modules/nw-networkpolicy-create.adoc
@@ -9,6 +9,12 @@
 To define granular rules describing Ingress network traffic allowed for projects
 in your cluster, you can create NetworkPolicy objects.
 
+[IMPORTANT]
+====
+If you create a network policy that uses a `podSelector`, and later you create a new Pod that will be matched by the `podSelector`, then you must delete and recreate the network policy for it to apply to the new Pod.
+This will be fixed in a future release of {product-title}.
+====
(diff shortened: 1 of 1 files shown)


modules/nw-networkpolicy-create.adoc
```diff
@@ -9,6 +9,12 @@
 To define granular rules describing Ingress network traffic allowed for projects
 in your cluster, you can create NetworkPolicy objects.
 
+[IMPORTANT]
+====
+If you create a network policy that uses a `podSelector`, and later you create a new Pod that will be matched by the `podSelector`, then you must delete and recreate the network policy for it to apply to the new Pod.
```
danwinship: The bug is a lot more specific and obscure than that; it's that if you have a namespace that does not have a blanket "default-deny" policy, and then you create a NetworkPolicy that creates a "default-deny" rule for *some* (but not all) pods in that namespace, and then add additional pods that ought to match that policy, they don't get added.

This bug has always been there but no one ever noticed because... people don't normally create policies like that. In general, if you care about security you're going to create a blanket default-deny policy first, and if you do that then you can't hit this bug.

jboxman:
Closing this issue.

Issue #12824, opened 2018-11-09
[enterprise-3.11] Edit suggested in file install_config/master_node_configuration.adoc


briantward:
### Which section(s) is the issue in?
https://docs.openshift.com/container-platform/3.11/install_config/master_node_configuration.html#master-config-image-policy-config

### What needs fixing?

Image Policy Config can now take this variable: "additionalTrustedCA"
This was required for a bugfix: https://bugzilla.redhat.com/show_bug.cgi?id=1592936
Comment #40, 41, 45

This was needed to do a valid import-image from registry.redhat.io

mburke5678:
@bparees We are going through older issues. 
Does `AdditionalTrustedCA` = `additionalTrustedCA`? 

We have this note in the docs. Does it cover this concern here?

> You can also point the registry import controller to another filesystem path for certificates
> by setting the `AdditionalTrustedCA` parameter in the `Image Policy Configuration` section of the master configuration file.

bparees:
yes.  in the yaml it will be `additionalTrustedCA`

PR #20712, opened 2020-03-26
Fixing ifdef statement in docs

mburke5678:
When researching https://github.com/openshift/openshift-docs/issues/12824 I noticed an incorrect `ifdef`. This PR fixes the error.


mburke5678:
/enterprise-3.11

PR #20718, opened 2020-03-27
Add missing module for SR-IOV

jboxman:


jboxman:
@zshi-redhat, adding missing module for master and 4.4.

PR #20728, opened 2020-03-27
Update Kuryr known limitations

luis5tb:
Some of the limitations has been fixed when using OpenStack 16
version of Octavia. This patch adds that information

### DIFF
--- a/modules/installation-osp-kuryr-known-limitations.adoc
+++ b/modules/installation-osp-kuryr-known-limitations.adoc
@@ -5,20 +5,57 @@
 [id="installation-osp-kuryr-known-limitations_{context}"]
 = Known limitations of installing with Kuryr
 
-There are known limitations when using Kuryr SDN:
+Using {product-title} with Kuryr SDN has several known limitations.
 
-* An Amphora load balancer VM is deployed per OpenShift Service with
-the default Octavia load balancer driver (Amphora driver). If the environment is
-resource constrained, creating a large amount of Services could be a problem.
-* Depending on the Octavia version, UDP listeners are not supported. This means
-that OpenShift UDP Services are not supported.
-* There is a known limitation of Octavia not supporting listeners on different
-protocols, like UDP and TCP, on the same port. Thus, Services exposing the same
-port for different protocols are not supported.
-* Due to the above UDP limitations of Octavia, Kuryr forces Pods to use TCP
-for DNS resolution. This is set with the `use-vc` option in `resolv.conf`. This
-might be a problem for Pods running Go applications compiled with the `CGO_ENABLED`
-flag disabled, as that uses the `go` resolver that only leverages UDP and is not
-considering the `use-vc` option added by Kuryr to the `resolv.conf`. This is a
-problem also for musl-based containers as its resolver does not support the
-`use-vc` option. This includes images built from `alpine`.
+[discrete]
+[id="openstack-resource-limitations_{context}"]
+== {rh-openstack} resource limitations
+
+* An Amphora load balancer VM is deployed per OpenShift Service that uses the
+default Octavia load balancer driver (Amphora driver). Creating too many Services
+can cause you to run out of resources.
+
+[discrete]
+[id="openstack-version-limitations_{context}"]
+== {rh-openstack} version limitations
+
+Using {product-title} with Kuryr SDN has several limitations that depend on the {rh-openstack} version.
+
+* Octavia {rh-openstack} versions before 16 do not support UDP listeners. Therefore,
+OpenShift UDP services are not supported.
+
+* Octavia {rh-openstack} versions before 16 cannot listen to multiple protocols on the
+same port. Services that expose the same port to different protocols, like TCP
+and UDP, are not supported.
+
+[IMPORTANT]
+====
+The OVN Octavia driver does not support listeners that use different protocols on
+any {rh-openstack} version.
+====
+
+[discrete]
+[id="openstack-go-limitations_{context}"]
+== {rh-openstack} environment limitations
+
+There are limitations when using Kuryr SDN that depend on your deployment environment.
+
+Because of Octavia's lack of support for the UDP protocol and multiple listeners, Kuryr forces Pods to use TCP
+for DNS resolution if:
+
+* The {rh-openstack} version is earlier than 16
+* The OVN Octavia driver is used
+
+In Go versions 1.12 and earlier, applications that are compiled with CGO support disabled use UDP only. In this case,
+the native Go resolver does not recognize the `use-vc` option in `resolv.conf`, which controls whether TCP is forced for DNS resolution.
+As a result, UDP is still used for DNS resolution, which fails.
+
+To ensure that TCP forcing is allowed, compile applications either with the environment variable `CGO_ENABLED` set to `1`, i.e. `CGO_ENABLED=1`, or ensure that the variable is absent.
+
+In Go versions 1.13 and later, TCP is used automatically if DNS resolution using UDP fails.
+
+[INFO]
+====
+musl-based containers, including Alpine-based containers, do not support the `use-vc` option.
+====
+
--- a/modules/installation-osp-kuryr-octavia-configuration.adoc
+++ b/modules/installation-osp-kuryr-octavia-configuration.adoc
@@ -116,13 +116,23 @@ backend is ML2/OVS. There is no need for modifications if the backend is
 ML2/OVN.
 ====
 
-. To enforce network policies across Services, like when traffic goes through
-the Octavia load balancer, you must ensure Octavia creates the Amphora VM
-security groups on the user project. To do that, you must add the project ID
+. In {rh-openstack} versions 13 and 15, add the project ID
 to the `octavia.conf` configuration file after you create the project.
+* To enforce
+network policies across Services, like when traffic goes through
+the Octavia load balancer, you must ensure Octavia creates the Amphora VM
+security groups on the user project.
 +
-This ensures that required LoadBalancer security groups belong to that project
+This change ensures that required LoadBalancer security groups belong to that project,
 and that they can be updated to enforce Services isolation.
++
+[NOTE]
+====
+This task is unnecessary in {rh-openstack} version 16 or later.
+
+Octavia implements a new ACL API that restricts access to the Load
+Balancers VIP.
+====
 
 .. Get the project ID
 +


maxwelldb:
@luis5tb Is there a ticket somewhere that we can add to this PR?

maxwelldb:
Context: https://issues.redhat.com/browse/OSASINFRA-1459

modules/installation-osp-kuryr-known-limitations.adoc
```diff
@@ -5,20 +5,26 @@
...
 
-There are known limitations when using Kuryr SDN:
+There are known limitations when using Kuryr SDN, depending on the
+{rh-openstack} version:
 
 * An Amphora load balancer VM is deployed per OpenShift Service with
 the default Octavia load balancer driver (Amphora driver). If the environment is
-resource constrained, creating a large amount of Services could be a problem.
+resource constrained, creating a large amount of Services could be a problem as
+one VM is created for each Service.
```
maxwelldb: Rewrite:

"A VM is created for each Service. In a resource-constrained environment, creating too many Services can cause problems."

If we want to be specific about the problems that can happen, even if it seems obvious, that's a good move, too.

modules/installation-osp-kuryr-known-limitations.adoc
```diff
@@ -5,20 +5,26 @@
...
 * Depending on the Octavia version, UDP listeners are not supported. This means
-that OpenShift UDP Services are not supported.
-* There is a known limitation of Octavia not supporting listeners on different
-protocols, like UDP and TCP, on the same port. Thus, Services exposing the same
-port for different protocols are not supported.
+that OpenShift UDP Services are not supported. UDP listeners are supported from
+{rh-openstack} 16+.
+* Depending on the Octavia version, there is a known limitation of Octavia
```
maxwelldb: Be explicit about the version in this line.

maxwelldb: "Octavia version X cannot listen to multiple protocols, like UDP and TCP, on the same port. Service that expose the same port to different services are not supported. Listeners that use different protocols are supported in {rh-openstack} version 16 and later."

modules/installation-osp-kuryr-known-limitations.adoc
```diff
@@ -5,20 +5,26 @@
...
+not supported. Listeners on different protocols are supported from
+{rh-openstack} 16+.
 * Due to the above UDP limitations of Octavia, Kuryr forces Pods to use TCP
-for DNS resolution. This is set with the `use-vc` option in `resolv.conf`. This
-might be a problem for Pods running Go applications compiled with the `CGO_DEBUG`
-flag disabled, as that uses the `go` resolver that only leverages UDP and is not
-considering the `use-vc` option added by Kuryr to the `resolv.conf`. This is a
-problem also for musl-based containers as its resolver does not support the
-`use-vc` option. This includes images built from `alpine`.
+for DNS resolution if the {rh-openstack} release is not 16 or newer. This is
```
maxwelldb: "for DNS resolution in {rh-openstack} versions 15 or earlier. Set this option by changing the value of `use-vc" in `resolve.conf`. Because they use the `go` resolver, which relies on UDP, Pods that run Go applications that were compiled with the `CGO_DEBUG` flag disabled may experience problems [which problems?] unless this option is set. musl-based containers, including those based on `alpine`, do not support the `use-vc` option."

I might need to take another pass at this, but does it preserve the meaning?

modules/installation-osp-kuryr-octavia-configuration.adoc
```diff
@@ -116,13 +116,20 @@ backend is ML2/OVS. There is no need for modifications if the backend is
 ML2/OVN.
 ====
 
-. To enforce network policies across Services, like when traffic goes through
+. (Optional, not needed for Red Hat OpenStack Platform 16+) To enforce
```
maxwelldb: Would I take this option with RHOSP 16+ for other reasons?

modules/installation-osp-kuryr-octavia-configuration.adoc
```diff
@@ -116,13 +116,20 @@ backend is ML2/OVS. There is no need for modifications if the backend is
...
 the Octavia load balancer, you must ensure Octavia creates the Amphora VM
 security groups on the user project. To do that, you must add the project ID
 to the `octavia.conf` configuration file after you create the project.
 +
 This ensures that required LoadBalancer security groups belong to that project
 and that they can be updated to enforce Services isolation.
++
+[NOTE]
+====
+This is not needed for Red Hat OpenStack Platform 16+ as Octavia implements a
```
maxwelldb: "You do not need to ensure that Octavia creates the security groups in {rh-openstack} version 16 or later."

modules/installation-osp-kuryr-octavia-configuration.adoc
maxwelldb: Can probably delete 'optional' section if the note later on is correct.

maxwelldb (changes requested):

modules/installation-osp-kuryr-known-limitations.adoc
luis5tb: first part of the sentence is repeated in the sentence before that. I'll add the second part

modules/installation-osp-kuryr-known-limitations.adoc
luis5tb: no it does not. I'll adapt a bit

modules/installation-osp-kuryr-octavia-configuration.adoc
luis5tb: ok

modules/installation-osp-kuryr-octavia-configuration.adoc
luis5tb: I think the current text gives more information that the proposed change

maxwelldb: I'd probably just keep my version and link out to the RHOSP release notes if the why is necessary, but I'm okay with this.

modules/installation-osp-kuryr-known-limitations.adoc
```diff
@@ -5,20 +5,27 @@
 [id="installation-osp-kuryr-known-limitations_{context}"]
 = Known limitations of installing with Kuryr
 
-There are known limitations when using Kuryr SDN:
+There are known limitations when using Kuryr SDN, depending on the
+{rh-openstack} version:
 
 * An Amphora load balancer VM is deployed per OpenShift Service with
```
maxwelldb: The line before this list says that limitations depend on the OSP version. Which version does this first item apply to?

modules/installation-osp-kuryr-known-limitations.adoc
```diff
@@ -5,20 +5,27 @@
...
-resource constrained, creating a large amount of Services could be a problem.
-* Depending on the Octavia version, UDP listeners are not supported. This means
-that OpenShift UDP Services are not supported.
-* There is a known limitation of Octavia not supporting listeners on different
-protocols, like UDP and TCP, on the same port. Thus, Services exposing the same
-port for different protocols are not supported.
+the default Octavia load balancer driver (Amphora driver). In a
+resource-constrained environment, creating too many Services can cause
+problems.
+* Octavia {rh-openstack} version 13 does not support UDP listeners. This means
```
maxwelldb: Suggestion:

* Octavia {rh-openstack} version 13 does not support UDP listeners. Therefore, OpenShift UDP services are not supported. UDP listeners are supported in Octavia {rh-openstack} version 16 and later.

modules/installation-osp-kuryr-known-limitations.adoc
```diff
@@ -5,20 +5,27 @@
...
+that OpenShift UDP services are not supported. UDP listeners a re supported
+from Octavia {rh-openstack} version 16 and later.
+* Octavia {rh-openstack} version 13 cannot listen to multiple protocols,
```
maxwelldb: Suggestion (should "services" on line 20 be "protocols?"):

* Octavia {rh-openstack} version 13 cannot listen to multiple protocols on the same port. Services that expose the same port to different protocols, like TCP and UDP, are not supported.

modules/installation-osp-kuryr-known-limitations.adoc
```diff
@@ -5,20 +5,27 @@
...
+like UDP and TCP, on the same port. Service that expose the same port to
+different services are not supported. Listeners that use different protocols
+are supported in {rh-openstack} version 16 and later. Note this only applies
```
maxwelldb: Suggestion:

[NOTE]
Only the Octavia Amphora driver supports different protocols on the same port in {rh-openstack} version 16. The OVN Octavia driver does not support listeners that use different protocols.

modules/installation-osp-kuryr-known-limitations.adoc
```diff
@@ -5,20 +5,27 @@
...
+to Octavia amphora driver. OVN Octavia driver still does not support listeners
+that use different protocols in {rh-openstack} version 16.
 * Due to the above UDP limitations of Octavia, Kuryr forces Pods to use TCP
-for DNS resolution. This is set with the `use-vc` option in `resolv.conf`. This
-might be a problem for Pods running Go applications compiled with the `CGO_DEBUG`
-flag disabled, as that uses the `go` resolver that only leverages UDP and is not
-considering the `use-vc` option added by Kuryr to the `resolv.conf`. This is a
-problem also for musl-based containers as its resolver does not support the
-`use-vc` option. This includes images built from `alpine`.
+for DNS resolution in {rh-openstack} version 15 or earlier, or if OVN Octavia
```
maxwelldb: Would it make sense to rephrase or rearrange this whole section so that it's in terms of <= version 15 vs. version 16? I could give that a shot, if so.

maxwelldb (changes requested):

modules/installation-osp-kuryr-known-limitations.adoc
maxwelldb: Or if it doesn't break down like that, we can rearrange :)

luis5tb: that is what it is explained below

modules/installation-osp-kuryr-known-limitations.adoc
luis5tb: First point is independent of the OSP version. 2nd and 3rd point are dependent on that. And the last one is a consequence of those. Perhaps that could be added in a follow up PR. Feel free to open that and ping me to take a look. What do you think?

eurijon:
This doc patch looks good to me as well

maxwelldb:
Added a PR of a restructuring idea to Luis' branch.

maxwelldb (changes requested):
Check out the PR to kuryr-octavia-configuration. Should be able to merge this soon, regardless, but I think the restructuring gets it there faster.

maxwelldb (approved):

maxwelldb:
Just needs QE and peer review.

eurijon:
lgtm

maxwelldb:
@openshift/team-documentation For peer review. I was somewhat involved with this, so a fresh pair of eyes would be good.

modules/installation-osp-kuryr-known-limitations.adoc
```diff
@@ -5,20 +5,58 @@
...
-protocols, like UDP and TCP, on the same port. Thus, Services exposing the same
-port for different protocols are not supported.
-* Due to the above UDP limitations of Octavia, Kuryr forces Pods to use TCP
-for DNS resolution. This is set with the `use-vc` option in `resolv.conf`. This
-might be a problem for Pods running Go applications compiled with the `CGO_ENABLED`
-flag disabled, as that uses the `go` resolver that only leverages UDP and is not
-considering the `use-vc` option added by Kuryr to the `resolv.conf`. This is a
-problem also for musl-based containers as its resolver does not support the
-`use-vc` option. This includes images built from `alpine`.
+There are a number of known limitations when using Kuryr SDN.
```
kalexand-rh, line 8: > ! there is, there are v
Avoid at the beginning of an independent clause.

modules/installation-osp-kuryr-known-limitations.adoc
```diff
@@ -5,20 +5,58 @@
...
+any {rh-openstack} version.
+====
+
+[discrete]
+[id="openstack-go-limitations_{context}"]
+== {rh-openstack} environment limitations
+
+There are limitations when using Kuryr SDN that depend on your deployment environment.
+
+* Because of Octavia's lack of support for the UDP protocol and multiple listeners, Kuryr forces Pods to use TCP
```
kalexand-rh, line 44: This section isn't rendering correctly. I think one solution would be to remove the asterisk from this line.

modules/installation-osp-kuryr-known-limitations.adoc
```diff
@@ -5,20 +5,58 @@
...
+
+* An Amphora load balancer VM is deployed per OpenShift Service that uses the
+default Octavia load balancer driver (Amphora driver). Creating too many Services
+can cause you to run out of resources.
+
+[discrete]
+[id="openstack-version-limitations_{context}"]
+== {rh-openstack} version limitations
+
+There are limitations when using Kuryr SDN that depend on the {rh-openstack}
```
kalexand-rh, line 22: > ! there is, there are v
Avoid at the beginning of an independent clause.

modules/installation-osp-kuryr-known-limitations.adoc
```diff
@@ -5,20 +5,58 @@
...
+== {rh-openstack} environment limitations
+
+There are limitations when using Kuryr SDN that depend on your deployment environment.
+
+* Because of Octavia's lack of support for the UDP protocol and multiple listeners, Kuryr forces Pods to use TCP
+for DNS resolution if:
+** The {rh-openstack} version is earlier than 16
+** The OVN Octavia driver is used
+
+The `use-vc` option defined in each Pod's `resolv.conf` file controls whether TCP is forced for DNS resolution.
```
kalexand-rh, line 49: This line seems to contradict the previous UL.

modules/installation-osp-kuryr-known-limitations.adoc
```diff
@@ -5,20 +5,58 @@
...
+
+In Go versions 1.12 and earlier, applications that are compiled with CGO support disabled use UDP only. In this case,
+the native Go resolver does not recognize the `use-vc` option in `resolv.conf`. As a result, UDP is still used for DNS resolution.
+
+To ensure that TCP forcing is allowed, compile applications either with the environment variable `CGO_ENABLED` set to `1`, i.e. `CGO_ENABLED=1`, or ensure that the variable is absent.
```
kalexand-rh, line 54: Wait, so you have to specially compile applications so that Kuryr is able to force TCP? What happens if it can't force TCP?

modules/installation-osp-kuryr-known-limitations.adoc
```diff
@@ -5,20 +5,58 @@
...
+
+In Go versions 1.13 and later, TCP is used automatically if UDP fails.
+
+[INFO]
+====
+musl-based containers, including Alpine-based containers, do not support the `use-vc` option.
```
kalexand-rh, line 60: What's the end result here? Do you have to compile using Go 1.13? Why don't you just make using Go 1.13 a requirement?

modules/installation-osp-kuryr-octavia-configuration.adoc
```diff
@@ -116,13 +116,21 @@ backend is ML2/OVS. There is no need for modifications if the backend is
 ML2/OVN.
 ====
 
-. To enforce network policies across Services, like when traffic goes through
+. In {rh-openstack} versions 13 and 15, to enforce
+network policies across Services, like when traffic goes through
 the Octavia load balancer, you must ensure Octavia creates the Amphora VM
-security groups on the user project. To do that, you must add the project ID
+security groups on the user project. Add the project ID
```
kalexand-rh, line 122: It's better if the action goes first in the step.

modules/installation-osp-kuryr-octavia-configuration.adoc
```diff
@@ -116,13 +116,21 @@ backend is ML2/OVS. There is no need for modifications if the backend is
...
 to the `octavia.conf` configuration file after you create the project.
 +
-This ensures that required LoadBalancer security groups belong to that project
+This change ensures that required LoadBalancer security groups belong to that project,
 and that they can be updated to enforce Services isolation.
++
+[NOTE]
+====
+You do not need to perform this task in {rh-openstack} version 16 or later.
```
kalexand-rh, line 130: s/You do not need to/Do not

> ! need to v
Avoid using “need to” to refer to a requirement that a product or procedure imposes. Use “

kalexand-rh (commented):
@maxwelldb, PTAL.

modules/installation-osp-kuryr-known-limitations.adoc
maxwelldb: UDP is still used for DNS resolution, which I'd imagine makes DNS resolution fail. Is that correct, @luis5tb ?

modules/installation-osp-kuryr-known-limitations.adoc
maxwelldb: @luis5tb This is a fair point. Why not just update the prereqs? We can get rid of much/all of this if we do that, right?

maxwelldb:
@luis5tb Could you enable this option in your upstream branch so that I can make a few edits?

https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

modules/installation-osp-kuryr-known-limitations.adoc
luis5tb: yep, the app will try to use udp for dns resolution and that traffic won't reach the dns pods that are behind an octavia loadbalancer (dns svc)

modules/installation-osp-kuryr-known-limitations.adoc
luis5tb: Still, it is a limitation and the use-vc is needed. I would prefer to leave this section as there may be other things (perhaps in other languages) doing similar things, and having the GO example here will help to understand if the problem is due to that

maxwelldb:
@luis5tb Thanks! Any concerns? Otherwise, just squash and we can merge ASAP.

luis5tb:
> @luis5tb Thanks! Any concerns? Otherwise, just squash and we can merge ASAP.

Done!

maxwelldb:
@luis5tb Thanks!

PR #20738, opened 2020-03-27
edits to channel topic

kalexand-rh:
Edits to https://github.com/openshift/openshift-docs/pull/19088.

@jottofar, will you PTAL? These changes are mostly to meet our various style requirements.

### DIFF
--- a/modules/understanding-upgrade-channels.adoc
+++ b/modules/understanding-upgrade-channels.adoc
@@ -7,159 +7,140 @@
 // * updating/updating-disconnected-cluster.adoc
 
 [id="understanding-upgrade-channels_{context}"]
-= {product-title} Upgrade Channels and Releases
+= {product-title} upgrade channels and releases
 
 In {product-title} 4.1, Red Hat introduced the concept of channels for
 recommending the appropriate release versions for cluster upgrade. By controlling
-the pace of upgrades, these upgrade channels allow users to choose an upgrade
+the pace of upgrades, these upgrade channels allow you to choose an upgrade
 strategy. Upgrade channels are tied to a minor version of
 {product-title}. For instance, {product-title} {product-version}
-upgrade channels will never include an upgrade to a 4.5 release. This ensures
-administrators make an explicit decision to upgrade to the next minor version of
-{product-title}. Upgrade channels only control release selection and have
-no impact on the version of the cluster you install; the `openshift-install`
-binary for a given patch level of {product-title} always installs that patch level.
+upgrade channels will never include an upgrade to a 4.5 release. This strategy ensures that
+administrators explicitly decide to upgrade to the next minor version of
+{product-title}. Upgrade channels control only release selection and do not impact the version of the cluster that you install; the `openshift-install`
+binary file for a specific version of {product-title} always installs that version.
 
-{product-title} {product-version} has the following upgrade channels to choose from:
+{product-title} {product-version} offers the following upgrade channels:
 
-* candidate-{product-version}
-* fast-{product-version}
-* stable-{product-version}
+* `candidate-{product-version}`
+* `fast-{product-version}`
+* `stable-{product-version}`
 
 [discrete]
-== candidate-{product-version} Channel
+== candidate-{product-version} channel
 
-The candidate-{product-version} channel will contain candidate builds for a z-stream
+The `candidate-{product-version}` channel contains candidate builds for a z-stream
 ({product-version}.z) release.
-Release candidates contain all the features of the product but they are not supported and
-should only be used to test feature acceptance and assist in qualifying the next version
+Release candidates contain all the features of the product but are not supported. Use release candidate versions to test feature acceptance and assist in qualifying the next version
 of {product-title}.
-A release candidate is any build (e.g. 4.3.0-rc.3, 4.3.0) that is available in the candidate
-channel.
-After a version lands in the candidate channel, it goes through more quality checks and if
-it meets the quality standard it is promoted to fast-{product-version} or stable-{product-version}
-channels.
-So if a given release is available in the candidate-{product-version} channel and also in the fast-{product-version}
-or stable-{product-version} channels, it is a Red Hat supported version.
-Additionally candidate-{product-version} may include dead end releases from which there are no or ever
-be recommended upgrades.
+A release candidate is any build that is available in the candidate channel, including ones that do not contain `-rc` in their names.
+After a version is available in the candidate channel, it goes through more quality checks. If
+it meets the quality standard, it is promoted to the `fast-{product-version}` or `stable-{product-version}` channels.
+Because of this strategy, if a specific release is available in both the `candidate-{product-version}` channel and in the `fast-{product-version}`
+or `stable-{product-version}` channels, it is a Red Hat supported version.
+The `candidate-{product-version}` channel can include release versions from which there are no recommended updates in any channel.
 
-The candidate-{product-version} channel also allows upgrading from a previous minor version of
-{product-title} to {product-version}.
+You can use the `candidate-{product-version}` channel to upgrade from a previous minor version of
+{product-title}.
 
 [NOTE]
 ====
-Release candidates differ from the nightly builds found on try.openshift.com. Nightly
-builds are available for early access to features but updating to or from nightly
-builds is neither recommended nor supported. Nightly builds are not avilable in
+Release candidates differ from the nightly builds found on the https://www.openshift.com/try site. Nightly
+builds are available for early access to features, but updating to or from nightly
+builds is neither recommended nor supported. Nightly builds are not available in
 any upgrade channel.
 ====
 
 [discrete]
-== fast-{product-version} Channel
+== fast-{product-version} channel
 
-The fast-{product-version} channel is updated with new {product-version} patch
-versions as soon as Red Hat declares the given patch as a general availability
-release. As such, these releases are fully supported, are production quality and have
-performed well while available as a release candidate in the candidate-{product-version}
-channel from where they were promoted. Some time after a release appears in
-fast-{product-version} it will also appear in stable-{product-version}. Releases will
-never appear in stable-{product-version} before they appear in fast-{product-version}.
+The `fast-{product-version}` channel is updated with new {product-version}
+versions as soon as Red Hat declares the given version as a general availability
+release. As such, these releases are fully supported, are production quality, and have
+performed well while available as a release candidate in the `candidate-{product-version}`
+channel from where they were promoted. Some time after a release appears in the
+`fast-{product-version}` channel, it is added to the `stable-{product-version}` channel. Releases
+never appear in the `stable-{product-version}` channel before they appear in the `fast-{product-version}` channel.
 
-The fast-{product-version} channel also allows upgrading from a previous minor version of
+You can use the `fast-{product-version}` channel to upgrade from a previous minor version of
 {product-title}.
 
 [discrete]
-== stable-{product-version} Channel
+== stable-{product-version} channel
 
-Like the fast-{product-version} channel, the stable-{product-version} channel will
-only contain releases that have been declared general availability are therefore
-fully supported. However the stable-{product-version} channel will gradually roll out
-releases to customers based on data from our SRE teams, support services, and
-pre-production and production environments that participate in our connected customer
-program, rather than being immediately available as they are in the fast-{product-version}
-channel. For patch and CVE fixes this delay can range from several hours to a day
-and allows an extra period of assessment in how the software performs.
+While the `fast-{product-version}` channel contains releases as soon
+as their errata are published, releases are added to the `stable-{product-version}` channel
+after a delay of several hours to a day. During this delay, data is collected from Red Hat SRE teams, Red Hat support services, and pre-production and production environments that participate in connected customer program about the stability of the release.
 
-The stable-{product-version} channel also allows upgrading from a previous minor version of
+You can use the `stable-{product-version}` channel to upgrade from a previous minor version of
 {product-title}.
 
 [discrete]
-== Upgrade Version Paths
+== Upgrade version paths
 
 {product-title} maintains an upgrade recommendation service that understands the
 version of {product-title} you have installed as well as the path to take within
 the channel you choose to get you to the next release. You can imagine seeing the
-following in the fast-{product-version} channel:
+following in the `fast-{product-version}` channel:
 
 * {product-version}.0
 * {product-version}.1
 * {product-version}.3
 * {product-version}.4
 
-The service only recommends upgrades that have been tested and have no serious issues.
-If you are on {product-version}.1 and {product-title} is allowing you to select
-{product-version}.4, then it is safe for you to go from .{product-version}.1 to .{product-version}.4. Likewise,
-the absence of {product-version}.2 may be due to a CVE that was fixed in {product-version}.3 and Red Hat no
-longer suggests upgrading to a known vulnerable version.
+The service recommends only upgrades that have been tested and have no serious issues.
+If your cluster is on {product-version}.1 and {product-title} suggests
+{product-version}.4, then it is safe for you to update from .{product-version}.1 to .{product-version}.4. Do not rely on consecutive patch numbers. In this example, {product-version}.2 is not, and never was, available in the channel. The update service will not suggest updating to a version of {product-title} that contains known vulnerabilities.
 
 Update stability depends on your channel. The presence of an update recommendation in
-the candidate-{product-version} channel does not imply that the update is supported.
-It means that no serious issues have been found with the update yet, but there may
+the `candidate-{product-version}` channel does not imply that the update is supported.
+It means that no serious issues have been found with the update yet, but there might
 not be significant traffic through the update to suggest stability. The presence of
-an update recommendation in the fast-{product-version} or stable-{product-version}
+an update recommendation in the `fast-{product-version}` or `stable-{product-version}`
 channels is a declaration that the update is fully supported while it is in the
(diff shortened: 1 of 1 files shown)


modules/understanding-upgrade-channels.adoc
```diff
@@ -7,159 +7,145 @@
...
 
-The candidate-{product-version} channel will contain candidate builds for a z-stream
+The `candidate-{product-version}` channel contains candidate builds for a z-stream
 ({product-version}.z) release.
-Release candidates contain all the features of the product but they are not supported and
-should only be used to test feature acceptance and assist in qualifying the next version
+Release candidates contain all the features of the product but are not supported. Use release candidate versions to test feature acceptance and assist in qualifying the next version
 of {product-title}.
-A release candidate is any build (e.g. 4.3.0-rc.3, 4.3.0) that is available in the candidate
+A release candidate is any build, such as `4.3.0-rc.3` that is available in the candidate
```
jottofar: We do want the other build, 4.3.0, mentioned as well to stress that "release candidate" does not only refer to a build  with the '-rc' extension.

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,159 +7,145 @@
...
-channels.
-So if a given release is available in the candidate-{product-version} channel and also in the fast-{product-version}
-or stable-{product-version} channels, it is a Red Hat supported version.
-Additionally candidate-{product-version} may include dead end releases from which there are no or ever
-be recommended upgrades.
-
-The candidate-{product-version} channel also allows upgrading from a previous minor version of
+After a version is available in the candidate channel, it goes through more quality checks, and if
+it meets the quality standard, it is promoted to the `fast-{product-version}` or `stable-{product-version}` channels.
+Because of this testing strategy, if a specific release is available in both the `candidate-{product-version}` channel and in the `fast-{product-version}`
```
jottofar: Since it's more than just testing it would be more accurate to drop "testing -  "Because of this strategy..."

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,159 +7,145 @@
...
-stable-{product-version} channels, and a new version may be introduced that will be the new
+later time, upgrades to that version might be blocked in both the `fast-{product-version}` and
+`stable-{product-version}` channels, and a new version might be introduced that becomes the new
 preferred upgrade target.
 
 Customers can improve this process by configuring pre-production systems on the
-fast-{product-version} channel, production systems on the stable-{product-version} channel,
-and participating in Red Hat’s xref:../support/remote_health_monitoring/about-remote-health-monitoring.adoc[connected customer program]. This program allows Red
+`fast-{product-version}` channel, configuring production systems on the `stable-{product-version}` channel,
+and participating in Red Hat’s connected customer program. This program allows Red
```
jottofar: Why removal of the cross-ref link that further explains this?

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,159 +7,145 @@
...
 
-Customers who have chosen to not be connected to Red Hat and are controlling their
-own {product-title} container image content manually should consult the Red Hat
-errata associated with product releases and note any comments impacting upgrades.
-During upgrade the user interface may caution about switching between these versions
-and it is up to the customer to ensure they have correctly selected the appropriate
-version before bypassing those cautions.
+If you manage the container images for your {product-title} clusters yourself, you must consult the Red Hat errata that is associated with product releases and note any comments that impact upgrades.
+During upgrade, the user interface provides warnings about updating to versions that contain
+vulnerabilities, so you must ensure that you selected an appropriate version before bypassing those warnings.
```
jottofar: "that contain vulnerabilities" is too specific, so ->

"During upgrade, the user interface provides warnings about updating to **certain versions**, so you must ensure that you selected an appropriate version before bypassing those warnings."

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,159 +7,145 @@
...
-supported. A customer can switch from candidate-{product-version} to fast-{product-version}
-if their current release is a general availability release. Customers can always
-switch from fast-{product-version} to stable-{product-version}, although there may
-be a delay of up to a day if their current release was recently promoted to
-fast-{product-stable} while they wait for the release to be promoted to
-stable-{product-version}. If you change to a channel that does not include your
-current release, an alert will fire, no updates will be recommended, and you can
+== Switching between channels
+
+Your cluster is always supported if you change from the `stable-{product-version}` channel to
```
jottofar: "Your cluster is still..." or "Your cluster continues to be..."

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,159 +7,145 @@
...
+
+The `candidate-{product-version}` channel also allows upgrading from a previous minor version of
 {product-title} to {product-version}.
 
 [NOTE]
 ====
-Release candidates differ from the nightly builds found on try.openshift.com. Nightly
-builds are available for early access to features but updating to or from nightly
-builds is neither recommended nor supported. Nightly builds are not avilable in
+Release candidates differ from the nightly builds found on the link:https://www.openshift.com/try site. Nightly
```
jottofar: I think this needs to be fixed - "the link:https://www.openshift.com/try site"

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,159 +7,145 @@
 // * updating/updating-disconnected-cluster.adoc
 
 [id="understanding-upgrade-channels_{context}"]
-= {product-title} Upgrade Channels and Releases
+= Understanding {product-title} upgrade channels
```
wking: I'd asked for dropping "Understanding" [here][1].  Is there a reason this particular header needs to call out "Understanding" but not all the other headers in openshift-docs?

[1]: https://github.com/openshift/openshift-docs/pull/19088#discussion_r367762031

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,159 +7,145 @@
...
-After a version lands in the candidate channel, it goes through more quality checks and if
-it meets the quality standard it is promoted to fast-{product-version} or stable-{product-version}
-channels.
-So if a given release is available in the candidate-{product-version} channel and also in the fast-{product-version}
-or stable-{product-version} channels, it is a Red Hat supported version.
-Additionally candidate-{product-version} may include dead end releases from which there are no or ever
-be recommended upgrades.
-
-The candidate-{product-version} channel also allows upgrading from a previous minor version of
+After a version is available in the candidate channel, it goes through more quality checks, and if
```
wking: This sentence is pretty long (and it was long before).  Maybe:

> ... more quality checks.  If it meets...

?

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,159 +7,145 @@
...
+it meets the quality standard, it is promoted to the `fast-{product-version}` or `stable-{product-version}` channels.
+Because of this testing strategy, if a specific release is available in both the `candidate-{product-version}` channel and in the `fast-{product-version}`
+or `stable-{product-version}` channels, it is a Red Hat supported version.
+The `candidate-{product-version}` channel can include release versions that you can never upgrade from.
```
wking: I think there's a useful distinction between upgrade possibilities, which you _can_ attempt and which might succeed, and recommended updates.  Channels are our recommended updates.  If you actually attempt those updates, you have a good chance of succeeding, but you might fail.  If you attempt an update that we do not recommend, your odds are probably not as good, but there's still a chance that you succeed for at least some version pairs.  How about:

> The `candidate-{product-version}` channel can include release versions from which there are no recommended updates in any channel.

Re: _in any channel", we will eventually end-of-life 4.1.z, after which the last 4.1.z will have no outgoing edges _in stable-4.1_.  But you'll be able to change into stable-4.2 and update out of that 4.1.z tip and on into 4.2 releases.  Candidate channels are special in that we are fine just writing off a broken release and never provide folks with any offramp.

Complicating this a bit further, I have fuzzy recollections that that long-term-support channels will continue to see z-stream releases beyond the point where updates to the next minor version are supported, and folks will be recommended to use some non-cluster-version-operator migration tool?  If that's true, we probably want text talking about it somewhere in this file.

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,159 +7,145 @@
...
+versions as soon as Red Hat declares the given version as a general availability
+release. As such, these releases are fully supported, are production quality, and have
+performed well while available as a release candidate in the `candidate-{product-version}`
+channel from where they were promoted. Some time after a release appears in the
+`fast-{product-version}` channel, it is added to the `stable-{product-version}` channel. Releases
+never appear in the `stable-{product-version}` channel before they appear in the `fast-{product-version}` channel.
 
-The fast-{product-version} channel also allows upgrading from a previous minor version of
-{product-title}.
+By subscribing to the `fast-{product-version}` channel, you have the ability to upgrade from a previous minor version of {product-title}.
```
wking: Can we echo the earlier candidate wording for this?  That has "allows", which seems weaker to me than "have the ability to".  Although thinking about this more, I can't put my finger on why.  In any case, the idea we're trying to express for all of these channels is that they _may_ include both `{product-version}` releases and releases from the previous minor version of `{product-title}`.  We should avoid phrasing that sounds like we are committing to those channels _actually_ including releases from either channel or having any cross-minor update recommendations.  Both of those are possible, but not required.

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,159 +7,145 @@
...
+== stable-{product-version} channel
 
-Like the fast-{product-version} channel, the stable-{product-version} channel will
-only contain releases that have been declared general availability are therefore
-fully supported. However the stable-{product-version} channel will gradually roll out
-releases to customers based on data from our SRE teams, support services, and
+Like the `fast-{product-version}` channel, the `stable-{product-version}` channel
+contains only releases that have general availability and are
+fully supported. While the `fast-{product-version}` channel contains releases as soon
+as they are released, releases are added to the the `stable-{product-version}` channel
```
wking: "contains releases as soon as they are released" is not true, although "as soon as we declare support for them" would be closer.  Or maybe your "released" is short for "claim support for them" or "publish an Errata" or something?  If so, I'd rather spell that out.  Because my initial reaction is that "released" means "pushed to Quay" which happens before a release lands in _any_ channel, let alone the fast channel.

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,159 +7,145 @@
...
+after data from Red Hat SRE teams, Red Hat support services, and
 pre-production and production environments that participate in our connected customer
-program, rather than being immediately available as they are in the fast-{product-version}
-channel. For patch and CVE fixes this delay can range from several hours to a day
+program confirm that they are stable. This delay can range from several hours to a day
```
wking: Stable confirmation does not impact promotion into the stable channel.  Current policy is that if a release goes into fast, it will go into the corresponding stable channel within at least a day or two, even if it suffers from critical bugs.  The impact of serious bugs is that we remove incoming update recommendations.  I think the previous master wording was closer to this behavior.  Can we make smaller pivots to address any style issues with the master language?

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,159 +7,145 @@
...
 * {product-version}.4
 
-The service only recommends upgrades that have been tested and have no serious issues.
-If you are on {product-version}.1 and {product-title} is allowing you to select
-{product-version}.4, then it is safe for you to go from .{product-version}.1 to .{product-version}.4. Likewise,
-the absence of {product-version}.2 may be due to a CVE that was fixed in {product-version}.3 and Red Hat no
-longer suggests upgrading to a known vulnerable version.
+The service recommends only upgrades that have been tested and have no serious issues.
+If your cluster is on {product-version}.1 and {product-title} suggests
+{product-version}.4, then it is safe for you to update from .{product-version}.1 to .{product-version}.4. If {product-version}.2 is no longer available from the update service, it might be due to a vulnerability that was fixed in {product-version}.3. The update service will not suggest updating to a version of {product-title{ that contains known vulnerabilities.
```
wking: "no longer available" implies that the release was in the channel at some point, but we never remove releases from channels, we just block incoming edges.  Sometimes we cut a release and then decide before it ever lands in a channel (even candidate) that it is too buggy to ship.  An example is [4.3.4][1].  We also skip patch numbers if we go more than a week between releases (e.g. because we have no bugs to fix).  An example is [4.1.32][2].  Can we update this wording to be closer to:

> Do not rely on consecutive patch numbers.  For example, {product-version}.2 is not, and never was, available in the example above.

And not get into why we sometimes skip over patches?

[1]: https://github.com/openshift/cincinnati-graph-data/blob/61f8a42332b767613859535bd53c22ba24b6665c/channels/candidate-4.3.yaml#L24
[2]: https://github.com/openshift/cincinnati-graph-data/blob/61f8a42332b767613859535bd53c22ba24b6665c/channels/prerelease-4.1.yaml#L40

modules/understanding-upgrade-channels.adoc
wking: This section is not about vulnerabilities, it's about:

> If your cluster cannot reach it's upstream Cincinnati service (e.g. our Red Hat Cincinnati service), it will not know about any recommended updates, so any update will need the admin to say:
> > It's ok, cluster-version operator, I'm comfortable with this update despite you never having heard of it before.

I think the previous master wording was closer to this behavior. Can we make smaller pivots to address any style issues with the master language?

wking:
If we're looking for common ground to land quickly so we can focus on the trickier bits, the "fast-4.3" -> "the `fast-4.3` channel" changes seem pretty easy to spin out and land without contention.

modules/understanding-upgrade-channels.adoc
kalexand-rh: @wking, I am happy to not get into why we skip patches.

modules/understanding-upgrade-channels.adoc
kalexand-rh: I'm happy to pull "allows" from all of them. @wking, PTAL?

kalexand-rh (commented):
@wking, @jottofar, thanks for the feedback! Will you please take another look?

modules/understanding-upgrade-channels.adoc
wking: Found some internal docs talking about a migration tool like [the 4.3 -> 4.4 tool][1] for things like 4.1 -> 4.3 updates.  But the docs said that we'd support 4.y -> 4.(y+1) updates via the cluster-version operator for the life of 4.y.z, so no need to pull in migrations tooling here; just ignore that last paragraph in my earlier comment.

[1]: https://docs.openshift.com/container-platform/4.2/migration/migrating_3_4/deploying_cam.html

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,159 +7,146 @@
...
-the pace of upgrades, these upgrade channels allow users to choose an upgrade
+the pace of upgrades, these upgrade channels allow you to choose an upgrade
 strategy. Upgrade channels are tied to a minor version of
 {product-title}. For instance, {product-title} {product-version}
-upgrade channels will never include an upgrade to a 4.5 release. This ensures
-administrators make an explicit decision to upgrade to the next minor version of
-{product-title}. Upgrade channels only control release selection and have
-no impact on the version of the cluster you install; the `openshift-install`
-binary for a given patch level of {product-title} always installs that patch level.
+upgrade channels will never include an upgrade to a 4.5 release. This strategy ensures that
```
wking, line 17: Hmm, is master now tracking 4.5, in which case this should be bumped to "a 4.6 release"?

modules/understanding-upgrade-channels.adoc
wking: > If that's what you're trying to emphasize, I think it would be clearer to say...

I like the wording you propose here, but you seem to have marked this thread resolved without updating the pull request to include it?

modules/understanding-upgrade-channels.adoc
```diff
@@ -7,159 +7,146 @@
...
-be recommended upgrades.
+After a version is available in the candidate channel, it goes through more quality checks. If
+it meets the quality standard, it is promoted to the `fast-{product-version}` or `stable-{product-version}` channels.
+Because of this strategy, if a specific release is available in both the `candidate-{product-version}` channel and in the `fast-{product-version}`
+or `stable-{product-version}` channels, it is a Red Hat supported version.
+The `candidate-{product-version}` channel can include release versions from which there are no recommended updates in any channel.
 
-The candidate-{product-version} channel also allows upgrading from a previous minor version of
-{product-title} to {product-version}.
+You can also use the `candidate-{product-version}` channel to upgrade from a previous minor version of
```
wking, line 43: nit: Drop "also"?  You don't include it in the analogous `fast-...` section below.

modules/understanding-upgrade-channels.adoc
wking: > Would something like this be better?

Yeah, that proposal looks good to me.

modules/understanding-upgrade-channels.adoc
wking: > I am happy to not get into why we skip patches.

Right, but I still think we still need to say something like my above proposal to mention that release versions may not be consecutive without getting into why.

modules/understanding-upgrade-channels.adoc
kalexand-rh: No, docs master is still tracking 4.4.

wking (approved):
Looks good to me, thanks :)

PR #20739, opened 2020-03-27
ClipboardJS feature for docs.openshift.com 

lbarbeevargas:
New clipboard js file, added clipboard css, added link to clipboardJS cdn to template.

@vikram-redhat PTAL - I changed the links to _javascripts and _stylesheets back to 4.1 from my local links. I tested using local/relative links for these files, so further testing is needed.

### DIFF
--- a/_javascripts/clipboard.js
+++ b/_javascripts/clipboard.js
@@ -1,0 +1,43 @@
+// This file runs the Clipboard.js functionality
+document.querySelectorAll('div.listingblock').forEach((codeblock, index) => {
+  codeblock.getElementsByTagName('pre')[0].insertAdjacentHTML("beforebegin", "<p class='clipboard-button-container'><span class='clipboard-button fa fa-clipboard'></span></p>");
+  document.getElementsByTagName('pre')[index].setAttribute('id',`clipboard-${index}`);
+});
+
+document.querySelectorAll('span.clipboard-button').forEach((copybutton, index) => {
+  copybutton.setAttribute('data-clipboard-target',`#clipboard-${index}`);
+});
+
+var clipboard = new ClipboardJS('.clipboard-button', {
+    text: function(target) {
+      const targetId = target.getAttribute('data-clipboard-target').substr(1);
+      const clipboardText = document.getElementById(targetId).innerText.replace(/\$\s/g, "");
+
+      if (clipboardText.slice(0, 2) === "# ") {
+        return clipboardText.substr(2);
+      }
+
+      return clipboardText;
+    }
+});
+
+clipboard.on("success", function (e) {
+  const triggerId = e.trigger.getAttribute("data-clipboard-target").substr(1);
+  const triggerNode = document.getElementById(triggerId);
+
+  const selection = window.getSelection();
+  selection.removeAllRanges();
+
+  const range = document.createRange();
+  range.selectNodeContents(triggerNode);
+  selection.addRange(range);
+
+  e.trigger.classList.toggle("fa-clipboard");
+  e.trigger.classList.toggle("fa-check");
+
+  setTimeout(function () {
+    e.clearSelection();
+    e.trigger.classList.toggle("fa-clipboard");
+    e.trigger.classList.toggle("fa-check");
+  }, 2000);
+});
--- a/_stylesheets/docs.css
+++ b/_stylesheets/docs.css
@@ -141,6 +141,38 @@ License: https://creativecommons.org/licenses/by/2.0/
 
 /* End footer edits */
 
+/* ClipboardJS */
+p.clipboard-button-container {
+  position: relative;
+}
+
+span.clipboard-button {
+  position: absolute;
+  right: 0px;
+  bottom: -28px;
+  z-index: 1;
+  display: block;
+  width: 30px;
+  padding: 5px;
+  text-align: center;
+}
+
+span.clipboard-button:hover {
+  cursor: pointer;
+}
+
+.fa-clipboard {
+  color: #404040;
+}
+
+.fa-clipboard:hover {
+  color: #AAAAAA;
+}
+
+.fa-check {
+  color: #73804E;
+}
+/* End ClipboardJS edits */
 
 .fa-inverse:hover {
   color: #ccc;
--- a/_templates/_page_openshift.html.erb
+++ b/_templates/_page_openshift.html.erb
@@ -11,7 +11,8 @@
   <%= (distro_key == "openshift-webscale") ? '<meta name="robots" content="noindex,nofollow">' : '' %>
   <title><%= [topic_title, subgroup_title].compact.join(' - ') %> | <%= group_title %> | <%= distro %> <%= version %></title>
   <link href="https://assets.openshift.net/content/subdomain.css" rel="stylesheet" type="text/css">
-  <link href="https://docs.openshift.com/container-platform/4.1/_stylesheets/docs.css" rel="stylesheet" />
+  <!--<link href="https://docs.openshift.com/container-platform/4.1/_stylesheets/docs.css" rel="stylesheet" />-->
+  <link href="../../_stylesheets/docs.css" rel="stylesheet" />
   <link href="https://docs.openshift.com/container-platform/4.1/_stylesheets/search.css" rel="stylesheet" />
   <link href="https://assets.openshift.net/content/subdomain/touch-icon-precomposed.png" rel="apple-touch-icon-precomposed" type="image/png">
   <link href="https://assets.openshift.net/content/subdomain/favicon32x32.png" rel="shortcut icon" type="text/css">
@@ -150,6 +151,9 @@
   <script src="https://docs.openshift.com/container-platform/4.1/_javascripts/reformat-html.js" type="text/javascript"></script>
   <script src="https://docs.openshift.com/container-platform/4.1/_javascripts/hc-search.js" type="text/javascript"></script>
   <script src="https://docs.openshift.com/container-platform/4.1/_javascripts/page-loader.js" type="text/javascript"></script>
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/clipboard.js/2.0.6/clipboard.min.js" type="text/javascript"></script>
+  <!--<script src="https://docs.openshift.com/container-platform/4.1/_javascripts/clipboard.js" type="text/javascript"></script>-->
+  <script src="../../_javascripts/clipboard.js" type="text/javascript"></script>
 
   <script>
   var dk = '<%= distro_key %>';


vikram-redhat:
Hey @lbarbeevargas this is fantastic and works like a charm. Great work.

I suggest two fixes/additions:
1. If we can move the copy button away from the top of the listing to the right, within the code blocks? This might require some CSS work to make it align. Having that button interrupt the reading is a little distracting, especially on pages where there are a lot of single line code blocks.

2. When copying single line content, with $ shell prompt in the beginning, it copies the $ as well. We might need to change the behaviour to detect the $ and exclude that.

It will be great if you can set up a preview for this as well using the manual preview method (the automatic preview doesn't work on javascript/CSS updates).

lbarbeevargas:
@vikram-redhat Great, I'm glad the functionality is working as expected. I'll start working on the two requests (CSS & $ omission) as well as the manual preview.

lbarbeevargas:
@vikram-redhat 

Updated CSS to move the clipboard button to top right of the code block. Also, If the first line of code starts with either `$` or `#`, these characters are removed and not copied.

vikram-redhat:
@lbarbeevargas works great. Let's demo in the next meeting and get it ready for the 4.4 release.

lbarbeevargas:
> @lbarbeevargas works great. Let's demo in the next meeting and get it ready for the 4.4 release.

Sounds great! I'll have a demo ready for the meeting this coming Monday.

lbarbeevargas:
Preview of this feature is available here:
http://file.bos.redhat.com/lbarbeev/041720/ocp-clipboard-feature/welcome/index.html

The following are direct links to general examples of the feature:
http://file.bos.redhat.com/lbarbeev/041720/ocp-clipboard-feature/cli_reference/openshift_cli/getting-started-cli.html

http://file.bos.redhat.com/lbarbeev/041720/ocp-clipboard-feature/installing/install_config/installing-customizing.html

To build and view this feature, a relative path is set for the clipboard.js and docs.css files. Not all pages will show the clipboard button feature because of this limitation. If you notice that the CSS is not working as expected, the feature will not show up on that page.

_javascripts/clipboard.js
```diff
@@ -0,0 +1,43 @@
...
+document.querySelectorAll('span.clipboard-button').forEach((copybutton, index) => {
+  copybutton.setAttribute('data-clipboard-target',`#clipboard-${index}`);
+});
+
+var clipboard = new ClipboardJS('.clipboard-button', {
+    text: function(target) {
+      const targetId = target.getAttribute('data-clipboard-target').substr(1);
+      const clipboardText = document.getElementById(targetId).innerText;
+
+      if (clipboardText.charAt(0) === '$') {
```
jboxman, line 16: We'll need to loop this to handle the case of multiple commands in a block.

bergerhoffer:
Love love love this. Love the icon and placement. But I really don't think we should be trying to remove the beginning prompt. I think that (right or wrong) we have a decent bit of code blocks that have multiple commands. And the copy is copying everything, including output (if the code block has that, and plenty do), and we're not trying to remove the output.

Also, the way it is coded, it only searches for the starting char of `$` or `#` and removes it. There might possibly be starting code blocks that start with these, but aren't prompts. I can give an example now of this here [1]. This gives an example bash file, starting with `#`. I'm guessing this copy command would remove that.

I think the right answer here is to have this copy functionality copy text exactly as is. And then maybe we investigate another way to show prompts that doesn't include explicitly typing them before the command. I'm sure I've seen sites that have the prompt displayed as not part of the actual text, but beside it, not copyable. But we'd probably need to coordinate something like this so that it can be done on the portal as well.

[1] https://docs.openshift.com/container-platform/4.3/cli_reference/openshift_cli/extending-cli-plugins.html#cli-writing-plugins_cli-extend-plugins

jboxman:
@bergerhoffer, there are a few technical ways to resolve that; The larger issue is the one you raised: What kind of click-to-copy behavior makes the most sense for our users?

Creating a list of what other documentation teams are doing in this area might provide a useful baseline.

lbarbeevargas:
@bergerhoffer  and @jboxman, thank you for your feedback! All great things to consider and research.

Also, @bergerhoffer, re the bash file example, the `#` would be removed from the first line. That's a good use case where the intended functionality would not be appropriate.

vikram-redhat:
Agree with everything @bergerhoffer has said. Some great points. I had the initial suggestion to remove the $/# but I didn't consider the case she has mentioned where the # needs to be part of copy.

With the copy-everything approach, I worry about our liability when the copy command copies everything including the $/# prompts and folks run it verbatim. They will need to know that whenever they see the $/# sign, they will need to paste the text somewhere else first, remove the prompt and then copy the resulting text, OR just select the text like they are used to currently. This is compounded when the copy command copies over outputs and multiple commands as well.

I see two things we may need to do going forward: 
1. Remove the $/# prompts everywhere where they are cosmetic, as part of this exercise. When we need to tell the reader to run those commands as root ($ is normal user, # is root), we say so in text leading up to it.
2. Separate output and multiple commands out, again as part of this PR/exercise. It might get a little ugly to have many in one page, but I think the copy command should be exactly copy-able without any more changes required, otherwise it will lose it usefulness. 

This is specific to docs.openshift.com/docs.okd.io. This will not work on the customer portal and there is no intention to even try and make it work on the portal.

jboxman:
@vikram-redhat, what about the commands with `<token>`, that are not runnable without being modified? All of the commands for the CNO and Multus are of this form.

What about commands that break across lines `\`? Many commands overflow and create a horizontal scroll. Without breaks, it is not possible to usefully use callouts for a command across lines.

It may be instructive to see the complete universe of code blocks in the doc set.

jboxman:
@openshift/team-documentation 

So here are all the code blocks we have that are neither JSON or YAML. There are 2,887 distinct code blocks. We have 1,740 modules. That's 1.66 code blocks per module, for those keeping score at home.

http://file.bos.redhat.com/jboxman/code_blocks/

We have a huge variety of styles.

vikram-redhat:
@jboxman impressive use of the syntax highlighter :)

For the ones with tokens, I think there is enough common understanding that commands with tokens are to be replaced so we are not liable there if folks don't replace them and run them as is.

With the copy functionality, I haven't yet seen one where the copy button copies across the prompt. In fact, I think commands that can be copied don't have the prompts at all, AFAIK.

jboxman:
@vikram-redhat, we can make the copy optional, I believe; There's a role attribute that can be used for code blocks:

```
[source.role,language]
```

And thus we can selectively enable it by selecting on a specific CSS class. I think this may help with much of the challenge here.

jboxman:
And that looks like this:

```
<div class="listingblock somerandomthing">
<div class="title">Network.config.openshift.io custom resource</div>
<div class="content">
<pre class="CodeRay highlight nowrap"><code data-lang="yaml"><span style="color:#606">apiVersion</span>: <span style="background-color:hsla(0,100%,50%,0.05)"><span style="color:#D20">config.openshift.io/v1</span></span>
<span style="color:#606">kind</span>: <span style="background-color:hsla(0,100%,50%,0.05)"><span style="color:#D20">Network</span></span>
<span style="color:#606">metadata</span>:
  <span style="color:#606">name</span>: <span style="background-color:hsla(0,100%,50%,0.05)"><span style="color:#D20">cluster</span></span>
<span style="color:#606">spec</span>:
  <span style="color:#606">externalIP</span>:
    <span style="color:#606">autoAssignCIDRs</span>: <span style="background-color:hsla(0,100%,50%,0.05)"><span style="color:#D20">[] </span></span><i class="conum" data-value="1"></i><b>(1)</b>
    <span style="color:#606">policy</span>: <i class="conum" data-value="2"></i><b>(2)</b>
      <span style="color:#F00;background-color:#FAA">...</span></code></pre>
</div>
</div>
```

vikram-redhat:
@openshift/team-documentation @openshift/team-aligned-docs 

Want to open this up to a wider audience to try and build a consensus/background around this as it will impact all docs. 

Please weigh in (see demo here: https://github.com/openshift/openshift-docs/pull/20739#issuecomment-615452097) and the other discussions around this.

jboxman:
Also of note, the portal also offers a click-to-copy feature for images (usage of `$` is inconsistent):

https://access.redhat.com/containers/?tab=images&get-method=registry-tokens#/registry.access.redhat.com/rhscl/mysql-80-rhel7

lbarbeevargas:
> Also of note, the portal also offers a click-to-copy feature for images (usage of `$` is inconsistent):
> 
> https://access.redhat.com/containers/?tab=images&get-method=registry-tokens#/registry.access.redhat.com/rhscl/mysql-80-rhel7

@jboxman Good example of the click-to-copy feature in the portal. It looks like it's not working though. I'm not able to copy any of the examples and can see an error in the console.

bobfuru:
All good points here. I've changed my mind repeatedly. :)

The copy highlighting that @lbarbeevargas uses is great because it shows what you're copying. But it could be confusing to click copy where there's a $/# and not see it pasted. 

I wonder how GCP is able to block output from being copied from the same code block here: 
https://cloud.google.com/compute/docs/images/create-delete-deprecate-private-images#setting_families (third code block example down)

^^ They also seem to limit docs to one command per code block. And they strip the $ from copy: https://cloud.google.com/compute/docs/disks/optimizing-local-ssd-performance#multiqueue

In the end, for simplification, I am good with the copy command simply copying everything...especially if the highlighting is indicating that.

lbarbeevargas:
Here's a great resource from Google on [Documenting command-line syntax](https://developers.google.com/style/code-syntax) that @vikram-redhat shared with me. 

@bobfuru  RE GCP, it looks like a `<span>` tag is wrapped around content that isn't copied with a "no-select" class (CSS - `user-select: none;`). There may be a way to dynamically add a class like this to each part of the code we don't want copied or at least the instances of `$` and `#`. If not, each block of code would need to be manually modified in the asciidoc for sections that we don't want copied.

lbarbeevargas:
Switched to using a regular expression to remove instances of "$ " and slice(0, 2) to check for "# " at the beginning of the copied text so as not to remove "#!/bin/bash" examples. @bergerhoffer

Let me know if there are any suggestions for using a reg expression to check for multiple root command prompts ("# "). I could not use the same reg expression as with "$ " because "# " is used for comments in some code blocks.

http://file.bos.redhat.com/lbarbeev/042320/ocp-clipboard-feature/cli_reference/openshift_cli/extending-cli-plugins.html#cli-writing-plugins_cli-extend-plugins

http://file.bos.redhat.com/lbarbeev/042320/ocp-clipboard-feature/installing/install_config/installing-customizing.html#building-and-testing-the-kernel-module-container

jboxman:
@lbarbeevargas, I'd only remove `#` if it is not followed by `!`. This is the syntax:

```
#!interpreter [optional-arg]
```

A space is optional, but possible. I think there will almost always be a path, as well; maybe the following, for each line:

```
^#\!\s*?\/
```

bergerhoffer:
Personally, I'd rather see us just copy everything as is, rather than try to parse out command prompts. It might be a different story if we were more consistent in our usage. Like if we only ever put one command in a block at a time, and didn't include output. But I feel that we're putting too much at risk if we don't find every situation that shouldn't be excluded, and confusing for users if the first prompt is removed, but no others (if there are more).

I like the way the button works, and it highlights everything including the prompt, so it's showing you what is all being copied. It would be ideal if we could get to a place where a copy would copy commands without the prompt, but I don't think we can get there without some work on code blocks on our end, which might not be worth the effort.

codyhoag:
Having a copy all button is definitely better than nothing at all. Some of my opinions and experiences (more to explain need than propose a solution :sweat_smile:):

- I had initiated (or revived) this conversation a few months ago based on our installer templates being massive. It was a terrible user experience having readers copy and scroll 100+ lines to copy a template. Although there are better solutions for how to present this in the works, the Copy button was proposed as a temporary solution for that issue.

- I've experienced reviewers request I remove the beginning `$`/`#` from a PR because they couldn't simply copy/paste the text and execute the command. Also, our commands with multiple info markers add line breaks to the commands when copying/pasting. That makes it very inconvenient when commands start getting massive. I would love to see that resolved with the Copy button because the installer docs specifically would be greatly improved if we could allow it. I do see the challenges in other parts of the docs; it'd be nice to find a hack to bypass it.

- My two cents: I do find it strange that we string multiple commands together along with output, etc in a single code block. It seems messy and hard to read. I'd love to see that massaged (or broken out into multiple code blocks), but I also know that's a huge undertaking that probably requires more effort than we can spare. Just an opinion of mine though. :wink:

maxwelldb:
As another person who's recently worked with example- and script-heavy docs, +1 to everything Cody said.

PR #20760, opened 2020-03-30
Updates for BZ1766200 - clarifying TLS version and router cipher suites

lamek:
https://bugzilla.redhat.com/show_bug.cgi?id=1766200

### DIFF
--- a/architecture/networking/routes.adoc
+++ b/architecture/networking/routes.adoc
@@ -151,8 +151,10 @@ ciphers for the connection to be complete:
 See the link:https://wiki.mozilla.org/Security/Server_Side_TLS[Security/Server
 Side TLS] reference guide for more information.
 
-The router defaults to the `intermediate` profile. You can select a different
-profile using the `--ciphers` option when creating a router, or by changing
+By default, the router selects the intermediate profile and sets ciphers based on this profile.
+When a profile is selected, only the ciphers are set. The TLS version is not governed by the profile.
+
+You can select a different profile by using the `--ciphers` option when creating a router, or by changing
 the `ROUTER_CIPHERS` environment variable with the values `modern`,
 `intermediate`, or `old` for an existing router. Alternatively, a set of ":"
 separated ciphers can be provided. The ciphers must be from the set displayed


architecture/networking/routes.adoc
```diff
@@ -151,8 +151,10 @@ ciphers for the connection to be complete:
 See the link:https://wiki.mozilla.org/Security/Server_Side_TLS[Security/Server
 Side TLS] reference guide for more information.
 
-The router defaults to the `intermediate` profile. You can select a different
-profile using the `--ciphers` option when creating a router, or by changing
+By default the router selects the intermediate profile and sets ciphers based on this profile.
```
kalexand-rh:
```suggestion
By default, the router selects the intermediate profile and sets ciphers based on this profile.
```

architecture/networking/routes.adoc
```diff
@@ -151,8 +151,10 @@ ciphers for the connection to be complete:
...
+When a profile is selected, only the ciphers are set. The TLS version is not governed by the profile.
+
+You can select a different profile using the `--ciphers` option when creating a router, or by changing
```
kalexand-rh:
```suggestion
You can select a different profile by using the `--ciphers` option when creating a router or by changing
```

architecture/networking/routes.adoc
```diff
@@ -151,8 +151,10 @@ ciphers for the connection to be complete:
 See the link:https://wiki.mozilla.org/Security/Server_Side_TLS[Security/Server
 Side TLS] reference guide for more information.
 
-The router defaults to the `intermediate` profile. You can select a different
-profile using the `--ciphers` option when creating a router, or by changing
+By default the router selects the intermediate profile and sets ciphers based on this profile.
+When a profile is selected, only the ciphers are set. The TLS version is not governed by the profile.
```
kalexand-rh: Do we provide information about setting the TLS version elsewhere?

kalexand-rh (commented):
Couple of picks and a question, then it LGTM.

architecture/networking/routes.adoc
lamek: The link to the Mozilla page has all the version specific info.

In this context (I believe) you're configuring your router to handle connections from clients supporting a specific version of TLS.

So, for example, if you wanted to only support secure connections from clients using TLS 1.3, you'd set the router to the Modern profile, and anyone using a client implementing an older version of TLS would have their connection request rejected.

lamek: *this link - https://wiki.mozilla.org/Security/Server_Side_TLS

PR #20762, opened 2020-03-30
Add optional annotation step to block default node selector in LSO

### LINKED ISSUE #20181: [enterprise-4.3] Edit suggested in file storage/persistent_storage/persistent-storage-local.adoc

https://docs.openshift.com/container-platform/4.3/storage/persistent_storage/persistent-storage-local.html#local-storage-install_persistent-storage-local

### What needs fixing?

The project creation step should include an annotation which blocks a default node selector, such as with:

```
oc annotate project local-storage openshift.io/node-selector=''
```

Without this annotation pods will inherit the cluster-wide default selector.

Clusters with infrastructure nodes are likely to use a default node selector that includes only worker nodes. This prevents the scheduler from running user apps on infra nodes in compliance with licensing.

However one might want to use the local storage operator to create volumes on infrastructure nodes in support of logging and monitoring components.

This is similar to https://bugzilla.redhat.com/show_bug.cgi?id=1782151.

bobfuru:
Adds an optional step to LSO Install procedure for blocking cluster-wide default node selector, as described in [ISSUE-20181](https://github.com/openshift/openshift-docs/issues/20181). 

@huffmanca or @gnufied PTAL - does this addition look accurate and make sense to you?

@liangxia PTAL

Preview link here: https://issue-20181--ocpdocs.netlify.com/openshift-enterprise/latest/storage/persistent_storage/persistent-storage-local.html#local-storage-install_persistent-storage-local

### DIFF
--- a/modules/persistent-storage-local-install.adoc
+++ b/modules/persistent-storage-local-install.adoc
@@ -19,6 +19,19 @@ The Local Storage Operator is not installed in {product-title} by default. Use t
 $ oc new-project local-storage
 ----
 
+. Optional: Allow local storage creation on infrastructure nodes.
++
+You might want to use the Local Storage Operator to create volumes on infrastructure nodes in support of components such as logging and monitoring.
++
+You must adjust the default node selector so that the Local Storage Operator includes the infrastructure nodes, and not just worker nodes.
++
+To block the Local Storage Operator from inheriting the cluster-wide default selector, enter the following command:
++
+----
+$ oc annotate project local-storage openshift.io/node-selector=''
+----
+
+
 . Install the Local Storage Operator from the web console.
 
 .. Log in to the {product-title} web console.


liangxia:
@qinpingli PTAL

qinpingli:
LGTM

modules/persistent-storage-local-install.adoc
```diff
@@ -19,6 +19,19 @@ The Local Storage Operator is not installed in {product-title} by default. Use t
 $ oc new-project local-storage
 ----
 
+. Optional: Add an annotation to the project that blocks a default node selector.
```
huffmanca: I would make the title of this step a bit more descriptive to the user's end goal. "Default node selector" is an implementation detail, and doesn't answer the "why" behind this step. What about something like:

> Optional: Allow local storage to be created on infrastructure nodes.

modules/persistent-storage-local-install.adoc
```diff
@@ -19,6 +19,19 @@ The Local Storage Operator is not installed in {product-title} by default. Use t
...
++
+You might want to use the Local Storage Operator to create volumes on infrastructure nodes in support of components such as logging and monitoring.
++
+However, clusters with infrastructure nodes are likely to use a default node selector that includes only worker nodes. To comply with licensing, this prevents the scheduler from running user apps on infrastructure nodes.
```
huffmanca: I think this needs one additional sentence at the end to fully indicate that the LSO wouldn't be able to be scheduled on these nodes with the default permissions. Perhaps it's overkill, but maybe something similar to the following added at the end of this paragraph:

> Without adjusting the default node selector, the Local Storage Operator will only be scheduled on worker nodes.

modules/persistent-storage-local-install.adoc
```diff
@@ -19,6 +19,19 @@ The Local Storage Operator is not installed in {product-title} by default. Use t
...
++
+To block Pods from inheriting the cluster-wide default selector, enter the following command:
```
huffmanca: s/Pods/Local Storage Operator/g

huffmanca (commented):
I've left some suggestions (that are more from a writer perspective than a technical one). None of these are required, but have been left for your consideration.

modules/persistent-storage-local-install.adoc
bobfuru: Yes, I agree it's a good idea to be more explicit. I slightly worded your suggested sentence: 
> You must adjust the default node selector so that the Local Storage Operator includes the infrastructure nodes, and not just worker nodes.

bobfuru:
> I've left some suggestions (that are more from a writer perspective than a technical one). None of these are required, but have been left for your consideration.

Thanks @huffmanca , this is great feedback! I've applied your suggestions and made notes based on my slight mods.

bmcelvee:
LGTM

PR #20783, opened 2020-03-31
Added details of SinkBinding for Serverless

abrennan89:
Please do not merge until requested.

### DIFF
--- a/modules/serverless-sinkbinding-kn.adoc
+++ b/modules/serverless-sinkbinding-kn.adoc
@@ -1,0 +1,128 @@
+// Module included in the following assemblies:
+//
+// serverless/knstive_eventing/serverless-sinkbinding.adoc
+
+[id="serverless-sinkbinding-kn_{context}"]
+= Using SinkBinding with the Knative CLI (`kn`)
+
+This guide describes the steps required to create, manage, and delete a SinkBinding instance using `kn` commands.
+
+.Prerequisites
+
+* You have Knative Serving and Eventing installed.
+* You have the `kn` CLI installed.
+
+.Procedure
+
+. To check that SinkBinding is set up correctly, create a Knative event display service, or event sink, that dumps incoming messages to its log:
++
+----
+$ kn service create event-display --image quay.io/openshift-knative/knative-eventing-sources-event-display:v0.13.2
+----
+
+. Create a SinkBinding that directs events to the service:
++
+----
+$ kn source binding create bind-heartbeat --subject Job:batch/v1:app=heartbeat-cron --sink svc:event-display
+----
+
+. Create a CronJob.
+.. Create a file named `heartbeats-cronjob.yaml` and copy the following sample code into it:
++
+----
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: heartbeat-cron
+spec:
+spec:
+  # Run every minute
+  schedule: "* * * * *"
+  jobTemplate:
+    metadata:
+      labels:
+        app: heartbeat-cron
+    spec:
+      template:
+        spec:
+          restartPolicy: Never
+          containers:
+            - name: single-heartbeat
+              image: quay.io/openshift-knative/knative-eventing-sources-heartbeats:v0.13.2
+              args:
+                - --period=1
+              env:
+                - name: ONE_SHOT
+                  value: "true"
+                - name: POD_NAME
+                  valueFrom:
+                    fieldRef:
+                      fieldPath: metadata.name
+                - name: POD_NAMESPACE
+                  valueFrom:
+                    fieldRef:
+                      fieldPath: metadata.namespace
+----
+.. After you have created the `heartbeats-cronjob.yaml` file, apply it by entering:
++
+----
+$ oc apply --filename heartbeats-cronjob.yaml
+----
+
+. Check that the controller is mapped correctly by entering the following command and inspecting the output:
++
+----
+$ kn source binding describe bind-heartbeat
+----
++
+The output should be similar to:
++
+----
+Name:         bind-heartbeat
+Namespace:    demo-2
+Annotations:  sources.knative.dev/creator=minikube-user, sources.knative.dev/lastModifier=minikub ...
+Age:          2m
+Subject:
+  Resource:   job (batch/v1)
+  Selector:
+    app:      heartbeat-cron
+Sink:
+  Name:       event-display
+  Resource:   Service (serving.knative.dev/v1)
+
+Conditions:
+  OK TYPE     AGE REASON
+  ++ Ready     2m
+----
+
+.Verification steps
+
+You can verify that the Kubernetes events were sent to the Knative event sink by looking at the message dumper function logs.
+
+You can view the message dumper function logs by entering:
+----
+$ oc get pods
+$ oc logs $(oc get pod -o name | grep event-display) -c user-container
+----
+
+The logs should contain lines similar to the following:
+----
+☁️  cloudevents.Event
+Validation: valid
+Context Attributes,
+  specversion: 1.0
+  type: dev.knative.eventing.samples.heartbeat
+  source: https://knative.dev/eventing-contrib/cmd/heartbeats/#event-test/mypod
+  id: 2b72d7bf-c38f-4a98-a433-608fbcdd2596
+  time: 2019-10-18T15:23:20.809775386Z
+  contenttype: application/json
+Extensions,
+  beats: true
+  heart: yes
+  the: 42
+Data,
+  {
+    "id": 1,
+    "label": ""
+  }
+----
--- a/modules/serverless-sinkbinding-yaml.adoc
+++ b/modules/serverless-sinkbinding-yaml.adoc
@@ -1,0 +1,164 @@
+// Module included in the following assemblies:
+//
+// serverless/knstive_eventing/serverless-sinkbinding.adoc
+
+[id="serverless-sinkbinding-yaml_{context}"]
+= Using SinkBinding with the YAML method
+
+ This guide describes the steps required to create, manage, and delete a SinkBinding instance using YAML files.
+
+.Prerequisites
+
+* You have Knative Serving and Eventing installed.
+
+.Procedure
+
+. To check that SinkBinding is set up correctly, create a Knative event display service, or event sink, that dumps incoming messages to its log.
+.. Copy the following sample YAML into a file named `service.yaml`:
++
+----
+apiVersion: serving.knative.dev/v1
+kind: Service
+metadata:
+  name: event-display
+spec:
+  template:
+    spec:
+      containers:
+        - image: quay.io/openshift-knative/knative-eventing-sources-event-display:v0.13.2
+----
+.. After you have created the `service.yaml` file, apply it by entering:
++
+----
+$ oc apply --filename service.yaml
+----
+
+. Create a SinkBinding that directs events to the service.
+.. Create a file named `sinkbinding.yaml` and copy the following sample code into it:
++
+----
+apiVersion: sources.knative.dev/v1alpha1
+kind: SinkBinding
+metadata:
+  name: bind-heartbeat
+spec:
+  subject:
+    apiVersion: batch/v1
+    kind: Job <1>
+    selector:
+      matchLabels:
+        app: heartbeat-cron
+
+  sink:
+    ref:
+      apiVersion: serving.knative.dev/v1
+      kind: Service
+      name: event-display
+----
++
+<1> In this example, any Job with the label `app: heartbeat-cron` will be bound to the event sink.
+.. After you have created the `sinkbinding.yaml` file, apply it by entering:
++
+----
+$ oc apply --filename sinkbinding.yaml
+----
+
+. Create a CronJob.
+.. Create a file named `heartbeats-cronjob.yaml` and copy the following sample code into it:
++
+----
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: heartbeat-cron
+spec:
+spec:
+  # Run every minute
+  schedule: "* * * * *"
+  jobTemplate:
+    metadata:
+      labels:
+        app: heartbeat-cron
+    spec:
+      template:
+        spec:
+          restartPolicy: Never
+          containers:
+            - name: single-heartbeat
+              image: quay.io/openshift-knative/knative-eventing-sources-heartbeats:v0.13.2
+              args:
+                - --period=1
+              env:
+                - name: ONE_SHOT
+                  value: "true"
+                - name: POD_NAME
+                  valueFrom:
+                    fieldRef:
+                      fieldPath: metadata.name
+                - name: POD_NAMESPACE
+                  valueFrom:
+                    fieldRef:
+                      fieldPath: metadata.namespace
+----
+.. After you have created the `heartbeats-cronjob.yaml` file, apply it by entering:
++
+----
+$ oc apply --filename heartbeats-cronjob.yaml
+----
+
+. Check that the controller is mapped correctly by entering the following command and inspecting the output:
++
+----
+$ oc get sinkbindings.sources.knative.dev bind-heartbeat -oyaml
+----
++
+The output should be similar to:
++
+----
+spec:
+  sink:
+    ref:
+      apiVersion: serving.knative.dev/v1
+      kind: Service
+      name: event-display
+      namespace: default
+  subject:
+    apiVersion: batch/v1
+    kind: Job
+    namespace: default
+    selector:
+      matchLabels:
+        app: heartbeat-cron
+----
+
+.Verification steps
+
+You can verify that the Kubernetes events were sent to the Knative event sink by looking at the message dumper function logs.
+
+You can view the message dumper function logs by entering:
+----
+$ oc get pods
+$ oc logs $(oc get pod -o name | grep event-display) -c user-container
+----
+
+The logs should contain lines similar to the following:
+----
+☁️  cloudevents.Event
+Validation: valid
+Context Attributes,
+  specversion: 1.0
+  type: dev.knative.eventing.samples.heartbeat
+  source: https://knative.dev/eventing-contrib/cmd/heartbeats/#event-test/mypod
+  id: 2b72d7bf-c38f-4a98-a433-608fbcdd2596
+  time: 2019-10-18T15:23:20.809775386Z
+  contenttype: application/json
+Extensions,
+  beats: true
+  heart: yes
+  the: 42
+Data,
+  {
+    "id": 1,
+    "label": ""
+  }
+----
--- a/serverless/knative_eventing/serverless-sinkbinding.adoc
+++ b/serverless/knative_eventing/serverless-sinkbinding.adoc
@@ -1,0 +1,19 @@
+include::modules/serverless-document-attributes.adoc[]
+[id="serverless-sinkbinding"]
+= Using SinkBinding
+include::modules/common-attributes.adoc[]
+:context: serverless-sinkbinding
+
+toc::[]
+
+SinkBinding is used to connect event producers, or _event sources_, to an event consumer, or _event sink_, for example, a Knative service or application.
+
+[NOTE]
+====
+Both of the following procedures require you to create YAML files.
+
+If you change the names of the YAML files from those used in the examples, you must ensure that you also update the corresponding CLI commands.
+====
+
+include::modules/serverless-sinkbinding-kn.adoc[leveloffset=+1]
+include::modules/serverless-sinkbinding-yaml.adoc[leveloffset=+1]
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -1646,6 +1646,8 @@ Topics:
 ###    File: serverless-kn-source
 ###  - Name: Using ApiServerSource
 ###    File: serverless-apiserversource
+###  - Name: Using SinkBinding
+###    File: serverless-sinkbinding
 ### Triggers
 ###  - Name: Using triggers with Knative Eventing
 ###    File: serverless-kn-trigger


modules/serverless-sinkbinding-kn.adoc
```diff
@@ -0,0 +1,72 @@
...
+----
+$ kn service create event-display --image quay.io/openshift-knative/knative-eventing-sources-event-display
+----
+
+. Create a SinkBinding that will direct events to the service.
++
+You can create the SinkBinding by entering the following command:
++
+----
+$ kn source sinkbinding create testevents --sink broker:default --resource "event:v1" --service-account events-sa
```
rhuss, line 30: The options are not correct (they are different than for the api server source). Actually a SinkBinding example is a bit more involved as you also need the other end for the binding (one is the broker). This can be a `Deployment`, `Job`, `StatefulSet` or anything which carries a Pod template and eventually fire up a Pod.

I would suggest to use a `Deployment` but the problem here, I don't know if we have a sample image out of the box that could be used here. Because the application within the image needs to honor the `K_SINK` environment variable which will get injected by Knative Operator to point to the broker's URL so that the application can send CloudEvents to the broker.

Let me check whether there is an example already available upstream. If not, I'm afraid we would need to create first such an application and make it generally available (like the `knative-eventing-sources-eventing-display` image mentioned above.

modules/serverless-sinkbinding-kn.adoc
```diff
@@ -0,0 +1,72 @@
...
+----
+
+. Create a trigger from the `default` broker, to filter events to the service created in the previous step, by entering the following command:
++
+----
+$ kn trigger create event-display-trigger --broker default --sink svc:event-display \ source=dev.knative.samples/testevents \
```
rhuss, line 36: this should be probably `--filter source=dev.knative.samples/testevents` but the value totally depends on the CloudEvents that the Deployment mentioned above is emitting.

modules/serverless-sinkbinding-yaml.adoc
```diff
@@ -0,0 +1,150 @@
...
++
+----
+$ oc apply --filename sinkbinding.yaml
+----
+
+. Create a CronJob.
++
+You can do this by creating a file named `heartbeats-source.yaml` and copying the following sample code into it:
++
+----
```
rhuss, line 78: Ah, here is the missing piece which is needed for the `kn` command.

@matzew what do you think, wouldn't it be better to use a `Deployment` instead of a `CronJob` as an example here ? Mostly because the `CronJob` is tricky because the sink binding needs to point to the included Job (not the CronJob itself), which makes it kind of hard to understant. That is not the case for a regular Deployment (and tbh I think Deployments are a much more likely use case then a CronJob).

The question is whether we can craft a `Deployment` descriptor with e.g. the heartbeat image.

rhuss (changes requested):
This PR needs some more work. The `kn` examples uses an Trigger / Broker which is not the case in the YAML section. There the sinkbinding is directly connected to the edisplay service (without an intermediate broker).

I'm also not sure that the CronJob sample is easy to understand. I would prefer a Deployment here as the concept of the SinkBinding is better reflected.

Let me write down a script for `kn` based on the CronJob sample and keep the discussion whether to change to a Deployment separately.

modules/serverless-sinkbinding-yaml.adoc
```diff
@@ -0,0 +1,150 @@
...
+    selector:
+      matchLabels:
+        app: heartbeat-cron
+
+  sink:
+    ref:
+      apiVersion: serving.knative.dev/v1
+      kind: Service
+      name: event-display
+  ceOverrides:
```
rhuss, line 61: @matzew Why is this required ?

modules/serverless-sinkbinding-yaml.adoc
```diff
@@ -0,0 +1,150 @@
...
++
+After you have created the `sinkbinding.yaml` file, apply it by entering the following command:
++
+----
+$ oc apply --filename sinkbinding.yaml
+----
+
+. Create a CronJob.
++
+You can do this by creating a file named `heartbeats-source.yaml` and copying the following sample code into it:
```
rhuss, line 76: I would call this `heartbeats-cronjob.yaml` as its technically not a source.

rhuss:
For using the SinkBinding with `kn` that reflect the YAML example, the following steps are needed:

* (Ignore the comment about the broker, we don't need it because we are not using triggers)
* Create the event-display service: `kn service create event-display --image gcr.io/knative-releases/github.com/knative/eventing-contrib/cmd/event_display`
* Create the sink binding: `kn source binding create --subject job:batch/v1:app=heartbeat-cron --sink svc:event-display --ce-override sink=bound` (not sure if the `--ce-override` option is really needed
* Create the CronJob with the YAML provided in the YAML section (including the `oc apply`)
* Verification is the same as for the YAML section

Does this make sense ?

modules/serverless-sinkbinding-yaml.adoc
```diff
@@ -0,0 +1,150 @@
...
+----
+apiVersion: serving.knative.dev/v1
+kind: Service
+metadata:
+  name: event-display
+spec:
+  template:
+    spec:
+      containers:
+        - image: gcr.io/knative-releases/github.com/knative/eventing-contrib/cmd/event_display
```
lberk, line 30: @abrennan89 do we want to change this to a quay.io image?

abrennan89: Oops, thought I did that already... yes, and I have the one from the conversation with you and Matthias the other day :smile: thanks!

modules/serverless-sinkbinding-yaml.adoc
lberk: I don't think it is strictly required.  It allows for outbound filtering -- which I don't think is used in this example, but it does show that you can add the overrides here and that they still exist in the `event-display`'s logs.

modules/serverless-sinkbinding-yaml.adoc
lberk: @rhuss I'm not sure I completely agree with that analysis.  With both a `deployment` as well as the `CronJob` the sinkbinding duck-types the underlying `container` partial schema.  Hence the 'podspecable' name.  I think this example works.  If anything perhaps we just make that more explicit that the sinkbinding will end up targeting the resulting pod?

lberk: Actually, one thing that could further emphasize/clarify this point is printing out the sinkbinding itself, ie:

```
oc get sinkbinds bind-heartbeat -oyaml
...
spec:
  ceOverrides:
    extensions:
      sink: bound
  sink:
    ref:
      apiVersion: serving.knative.dev/v1
      kind: Service
      name: event-display
      namespace: default
  subject:
    apiVersion: batch/v1
    kind: Job
    namespace: default
    selector:
      matchLabels:
        app: heartbeat-cron
```
which shows the sink, as well as the subject the sinkbinding is acting on

modules/serverless-sinkbinding-yaml.adoc
```diff
@@ -0,0 +1,150 @@
...
+$ oc get pods
+$ oc logs $(oc get pod -o name | grep event-display) -c user-container
+----
+
+The logs should contain lines similar to the following:
+----
+☁️  cloudevents.Event
+Validation: valid
+Context Attributes,
+  specversion: 0.3
```
lberk, line 134: if we're using the latest quay images, this is going to be `1.0`

modules/serverless-sinkbinding-yaml.adoc
abrennan89: Removed it from here and the `kn` command, lmk if this changes :)

modules/serverless-sinkbinding-yaml.adoc
abrennan89: Should I remove `ceOverrides` from this too @lberk if it's not being included in the YAML or `kn` commands now?

lberk: Ya if you're not going to use the `ceOverrides` it'd make sense to remove it here too

modules/serverless-sinkbinding-kn.adoc
abrennan89: Removed the trigger part altogether since it's not relevant?

modules/serverless-sinkbinding-yaml.adoc
```diff
@@ -0,0 +1,172 @@
...
+  specversion: 1.0
+  type: dev.knative.eventing.samples.heartbeat
+  source: https://knative.dev/eventing-contrib/cmd/heartbeats/#event-test/mypod
+  id: 2b72d7bf-c38f-4a98-a433-608fbcdd2596
+  time: 2019-10-18T15:23:20.809775386Z
+  contenttype: application/json
+Extensions,
+  beats: true
+  heart: yes
+  sink: bound
```
lberk, line 165: we dropped this override in the event-display yaml so we can drop `sink: bound` here as well :)

modules/serverless-sinkbinding-yaml.adoc
```diff
@@ -0,0 +1,172 @@
...
+----
+apiVersion: serving.knative.dev/v1
+kind: Service
+metadata:
+  name: event-display
+spec:
+  template:
+    spec:
+      containers:
+        - image: quay.io/openshift-knative/knative-eventing-sources-heartbeats:v0.13.2
```
lberk, line 30: I'd use `quay.io/openshift-knative/knative-eventing-sources-event-display:v0.13.2` here instead

dsimansk: A little observation, the image should be the same in `kn` section vs `yaml` one. I'll add a comment to the exact place. 

The actual value from Lukas's comment is probably the best candidate.

modules/serverless-sinkbinding-kn.adoc
```diff
@@ -0,0 +1,137 @@
...
+* You have the `kn` CLI installed.
+
+.Procedure
+
+. To check that SinkBinding is set up correctly, create a Knative event display service, or event sink, that dumps incoming messages to its log.
++
+You can create the service by entering the following command:
++
+----
+$ kn service create event-display --image gcr.io/knative-releases/github.com/knative/eventing-contrib/cmd/event_display
```
dsimansk, line 25: Per my [comment](https://github.com/openshift/openshift-docs/pull/20783/files#r403036649), this `--image` value should be the same as in yaml.

```
--image quay.io/openshift-knative/knative-eventing-sources-event-display:v0.13.2
```

modules/serverless-sinkbinding-kn.adoc
```diff
@@ -0,0 +1,137 @@
...
+----
+$ kn service create event-display --image quay.io/openshift-knative/knative-eventing-sources-event-display:v0.13.2
+----
+
+. Create a SinkBinding that will direct events to the service.
++
+You can create the SinkBinding by entering the following command:
++
+----
+$ kn source binding create --subject job:batch/v1:app=heartbeat-cron --sink svc:event-display
```
rhuss, line 33:
```suggestion
$ kn source binding create bind-heartbeat --subject job:batch/v1:app=heartbeat-cron --sink svc:event-display
```

modules/serverless-sinkbinding-kn.adoc
```diff
@@ -0,0 +1,137 @@
...
+After you have created the `heartbeats-cronjob.yaml` file, apply it by entering the following command:
++
+----
+$ oc apply --filename heartbeats-cronjob.yaml
+----
+
+. Check that the controller is mapped correctly by entering the following command and inspecting the output:
++
+----
+$ oc get sinkbindings.sources.knative.dev bind-heartbeat -oyaml
```
rhuss, line 84: I suggest better to use `kn` here, too:

```
$ kn source binding describe bind-heartbeat

Name:         bind-heartbeat
Namespace:    demo-2
Annotations:  sources.knative.dev/creator=minikube-user, sources.knative.dev/lastModifier=minikub ...
Age:          2m
Subject:
  Resource:   job (batch/v1)
  Selector:
    app:      heartbeat-cron
Sink:
  Name:       event-display
  Resource:   Service (serving.knative.dev/v1)

Conditions:
  OK TYPE     AGE REASON
  ++ Ready     2m
```

modules/serverless-sinkbinding-kn.adoc
rhuss, line 33:
```suggestion
$ kn source binding create bind-heartbeat --subject job:batch/v1:app=heartbeat-cron --sink svc:event-display
```

rhuss (approved):
Looks good to me from a kn pov. Added one correction and a minor comment.

However, I tried to replay the example, but it looks like that the sink binding is not picked properly. Could be also my setup though. Its then more a serving bug, the documentation is correct and is supposed to work.

modules/serverless-sinkbinding-kn.adoc
abrennan89: Awesome, thank you!

modules/serverless-sinkbinding-kn.adoc
```diff
@@ -0,0 +1,138 @@
...
+* You have the `kn` CLI installed.
+
+.Procedure
+
+. To check that SinkBinding is set up correctly, create a Knative event display service, or event sink, that dumps incoming messages to its log.
++
+You can create the service by entering the following command:
++
+----
+$ kn service create event-display --image quay.io/openshift-knative/knative-eventing-sources-event-display:v0.13.2
```
maschmid, line 25: Should use a specific tag :v0.13.2 for the current release

kn service create event-display --image quay.io/openshift-knative/knative-eventing-sources-event-display:v0.13.2

modules/serverless-sinkbinding-kn.adoc
```diff
@@ -0,0 +1,138 @@
...
+----
+
+. Create a SinkBinding that will direct events to the service.
++
+You can create the SinkBinding by entering the following command:
++
+----
+$ kn source binding create bind-heartbeat --subject job:batch/v1:app=heartbeat-cron --sink svc:event-display
```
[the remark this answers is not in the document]
lberk: I'm not sure with `kn` I'd have to defer to @rhuss

modules/serverless-sinkbinding-kn.adoc
```diff
@@ -0,0 +1,138 @@
...
+  specversion: 1.0
+  type: dev.knative.eventing.samples.heartbeat
+  source: https://knative.dev/eventing-contrib/cmd/heartbeats/#event-test/mypod
+  id: 2b72d7bf-c38f-4a98-a433-608fbcdd2596
+  time: 2019-10-18T15:23:20.809775386Z
+  contenttype: application/json
+Extensions,
+  beats: true
+  heart: yes
+  sink: bound
```
maschmid, line 131: I believe the idea in the original demo was that the "sink: bound" would be added by the ceOverrides , see https://knative.dev/docs/eventing/samples/sinkbinding/#create-our-sinkbinding  

(it is not shown if I follow the steps as written here)

so I believe it should be added to the "kn source binding create..." command above   "--ce-override sink=bound" maybe? cc: @rhuss

maschmid: (alternatively, if we don't want to show ceOverride, we can just remove the "sink: bound" here from the logs.

modules/serverless-sinkbinding-kn.adoc
```diff
@@ -0,0 +1,138 @@
...
+  the: 42
+Data,
+  {
+    "id": 2,
```
maschmid, line 135: should be "id": 1 

it can only be 1, as the CronJob repeatedly deploys the heartbeats container with ONE_SHOT=true  , so it will produce a sequence of one-shot heartbeats, so the id will always be just 1

modules/serverless-sinkbinding-yaml.adoc
```diff
@@ -0,0 +1,170 @@
...
+  id: 2b72d7bf-c38f-4a98-a433-608fbcdd2596
+  time: 2019-10-18T15:23:20.809775386Z
+  contenttype: application/json
+Extensions,
+  beats: true
+  heart: yes
+  the: 42
+Data,
+  {
+    "id": 2,
```
maschmid, line 167: should be "id": 1

modules/serverless-sinkbinding-kn.adoc
abrennan89: We actually removed this, must've missed this instance of `sink: bound`, but I'll remove it. Thanks!

modules/serverless-sinkbinding-kn.adoc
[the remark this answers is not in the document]
navidshaikh: It should be `Job`.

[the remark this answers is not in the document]
navidshaikh: Value for `--subject` should be in the form of `Kind:api/version:nameOrSelector`.

maschmid:
QE verified

modules/serverless-sinkbinding-kn.adoc
```diff
@@ -0,0 +1,134 @@
+// Module included in the following assemblies:
+//
+// serverless/knstive_eventing/serverless-sinkbinding.adoc
+
+[id="serverless-sinkbinding-kn_{context}"]
+= Using SinkBinding with the Knative CLI (`kn`)
+
+This guide describes the steps required to create, manage and delete a SinkBinding instance using `kn` commands.
```
mburke5678, line 8: s/manaage/manage,

modules/serverless-sinkbinding-kn.adoc
```diff
@@ -0,0 +1,134 @@
...
+.Prerequisites
+
+* You have Knative Serving and Eventing installed.
+* You have the `kn` CLI installed.
+
+.Procedure
+
+. To check that SinkBinding is set up correctly, create a Knative event display service, or event sink, that dumps incoming messages to its log.
++
+You can create the service by entering the following command:
```
mburke5678, line 19: Do you need these substeps? 
...create a Knative event display service, or event sink, that dumps incoming messages to its log by entering the following command:

modules/serverless-sinkbinding-kn.adoc
```diff
@@ -0,0 +1,134 @@
...
++
+----
+$ kn service create event-display --image quay.io/openshift-knative/knative-eventing-sources-event-display:v0.13.2
+----
+
+. Create a SinkBinding that will direct events to the service.
++
+You can create the SinkBinding by entering the following command:
```
mburke5678, line 27: Same here. You could remove the substep.
_Create a SinkBinding that will direct events to the service:_

modules/serverless-sinkbinding-kn.adoc
```diff
@@ -0,0 +1,134 @@
...
++
+----
+$ kn source binding create bind-heartbeat --subject Job:batch/v1:app=heartbeat-cron --sink svc:event-display
+----
+
+. Create a CronJob.
++
+You can do this by creating a file named `heartbeats-cronjob.yaml` and copying the following sample code into it:
```
mburke5678, line 35: Maybe?
.Create a CronJob
.. Create a file named...
.. Apply the `heartbeats-cronjob.yaml` file:

modules/serverless-sinkbinding-yaml.adoc
```diff
@@ -0,0 +1,170 @@
...
+        - image: quay.io/openshift-knative/knative-eventing-sources-event-display:v0.13.2
+----
++
+After you have created the `service.yaml` file, apply it by entering the following command:
++
+----
+$ oc apply --filename service.yaml
+----
+
+. Create a SinkBinding that will direct events to the service.
```
mburke5678, line 38: s/that will direct/that directs

modules/serverless-sinkbinding-kn.adoc
mburke5678: Must the YAML file be called `heartbeats-cronjob.yaml`?

modules/serverless-sinkbinding-yaml.adoc
```diff
@@ -0,0 +1,170 @@
...
+
+.Prerequisites
+
+* You have Knative Serving and Eventing installed.
+
+.Procedure
+
+. To check that SinkBinding is set up correctly, create a Knative event display service, or event sink, that dumps incoming messages to its log.
++
+Copy the following sample YAML into a file named `service.yaml`:
```
mburke5678, line 18: Must the file be named `service.yaml`?

modules/serverless-sinkbinding-yaml.adoc
```diff
@@ -0,0 +1,170 @@
...
++
+After you have created the `service.yaml` file, apply it by entering the following command:
++
+----
+$ oc apply --filename service.yaml
+----
+
+. Create a SinkBinding that will direct events to the service.
++
+You can do this by creating a file named `sinkbinding.yaml` and copying the following sample code into it:
```
mburke5678, line 40: Must the file be named `sinkbinding.yaml`?

serverless/knative_eventing/serverless-sinkbinding.adoc
```diff
@@ -0,0 +1,12 @@
+include::modules/serverless-document-attributes.adoc[]
+[id="serverless-sinkbinding"]
+= Using SinkBinding
+include::modules/common-attributes.adoc[]
+:context: serverless-sinkbinding
+
+toc::[]
+
+SinkBinding is used to connect event producers, or event sources, to an event consumer, or event sink, for example, a Knative service or application.
```
mburke5678, line 9: Should event sources and event sink by marked-up, as introducing new terms? 
connect event producers, or _event sources_, to an event consumer, or _event sink_,

mburke5678:
@abrennan89 I had some comments on the procedure format, where you use substeps. Is this a style in your docs? If so, ignore. If not, should you combine some of the intro sentences and substeps together? I offered a few examples, but didn't point them all out.

abrennan89:
Thanks so much for the review, @mburke5678 !
I've implemented the suggestions re structure etc, I think this way is much better. LMK if you think it all looks OK now :slightly_smiling_face: 

The only ones I haven't made changes to are the names of the yaml files; I think users can call these whatever they want, but then I'll need to add some notes to say "make sure you use the name of your yaml file in the commands" basically. I'll just get someone from eng to confirm before I do this.

abrennan89:
@mburke5678 I was able to confirm that the user can call those yaml files whatever they want technically, but I've added a note to the assembly: https://github.com/openshift/openshift-docs/pull/20783/files#diff-e1f22bfa9d3a7d1f271cf77b1774d019R11-R16

mburke5678:
@abrennan89 LGTM

PR #20789, opened 2020-03-31
updating config for hive metatstore

lamek:
Updating the config path for Hive metastore.


lamek:
@timflannagan1 PTAL.

bmcelvee:
LGTM!

PR #20796, opened 2020-03-31
Bug 1654320, added more details about openshift_console_auth_ca_file

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1654320

### DIFF
--- a/install/configuring_inventory_file.adoc
+++ b/install/configuring_inventory_file.adoc
@@ -2255,7 +2255,11 @@ following variables can be used to customize the cluster console.
 |Optional base path for the cluster console. If set, it should begin and end with a slash like `/console/`. Defaults to `/` (no base path).
 
 |`openshift_console_auth_ca_file`
-|Optional CA file to use to connect to the OAuth server. Defaults to `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`. Typically this does not need to be changed.
+|Optional CA file to use to connect to the OAuth server. Defaults to
+`/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`. Typically, this value
+does not need to be changed. You must create a ConfigMap that contains the CA
+file and mount it into the `console` Deployment in the `openshift-console`
+namespace at the same location as the location that you specify.
 
 |===
 


ahardin-rh:
@spadgett PTAL. This fix is for 3.11 docs only. Thanks!

install/configuring_inventory_file.adoc
```diff
@@ -2255,7 +2255,10 @@ following variables can be used to customize the cluster console.
 |Optional base path for the cluster console. If set, it should begin and end with a slash like `/console/`. Defaults to `/` (no base path).
 
 |`openshift_console_auth_ca_file`
-|Optional CA file to use to connect to the OAuth server. Defaults to `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`. Typically this does not need to be changed.
+|Optional CA file to use to connect to the OAuth server. Defaults to
+`/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`. Typically this does not
+need to be changed. You must create a secret containing the CA and mount it in
```
spadgett: This could be a config map instead of a secret.

Specifically this would be mounted into the `console` deployment in the `openshift-console` namespace.

ahardin-rh: @spadgett Thanks! Updated. PTAL

install/configuring_inventory_file.adoc
```diff
@@ -2255,7 +2255,11 @@ following variables can be used to customize the cluster console.
...
+the same location as set in the variable, or create a ConfigMap and mount it
+into the console deployment in the `openshift-console` namespace.
```
spadgett: Sorry, I wasn't clear. The only way to mount it in is by adding it to the `console` deployment.

```suggestion
need to be changed. You must create a config map containing the CA file and mount it into
the `console` deployment in the `openshift-console` namespace at the same location as
the variable.
```

ahardin-rh: @spadgett Thank you! Updated.

spadgett:
lgtm

ahardin-rh:
@yapei Can you please provide a QE review?

yapei:
> same location as the variable

I'm a little confused about this sentence, currently we have 

```
# oc get deployment console -n openshift-console -o yaml
...
      containers:
      - command:
        - /opt/bridge/bin/bridge
        - --public-dir=/opt/bridge/static
        - --config=/var/console-config/console-config.yaml
        image: registry-proxy.engineering.redhat.com/rh-osbs/openshift3-ose-console:v3.11
        imagePullPolicy: IfNotPresent
.....
        volumeMounts:
        - mountPath: /var/serving-cert
          name: serving-cert
          readOnly: true
        - mountPath: /var/oauth-config
          name: oauth-config
          readOnly: true
        - mountPath: /var/console-config
          name: console-config
.......
      volumes:
      - name: serving-cert
        secret:
          defaultMode: 288
          secretName: console-serving-cert
      - name: oauth-config
        secret:
          defaultMode: 288
          secretName: console-oauth-config
      - configMap:
          defaultMode: 288
          name: console-config
        name: console-config
```

yapei:
@spadgett  after searching in our cases pool I think we didn't cover this variable in 3.11 cluster console installation, I'm also confused about its usage

spadgett:
Yes, it sets the value of `oauthEndpointCAFile` in the console config map. If you change it, you have to also mount the file into the pod (unfortunately). But I wouldn't expect to need to change this in the normal case.

yapei:
@ahardin-rh  

> same location as the variable 

I think it should be `same location as other variable` ?

ahardin-rh:
@yapei After some discussion, we decided that the current wording is more accurate. I will omit "other". Thank you!

install/configuring_inventory_file.adoc
```diff
@@ -2255,7 +2255,11 @@ following variables can be used to customize the cluster console.
 |Optional base path for the cluster console. If set, it should begin and end with a slash like `/console/`. Defaults to `/` (no base path).
 
 |`openshift_console_auth_ca_file`
-|Optional CA file to use to connect to the OAuth server. Defaults to `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`. Typically this does not need to be changed.
+|Optional CA file to use to connect to the OAuth server. Defaults to
+`/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`. Typically this does not
```
kalexand-rh, line 2259: > Typically, this value

install/configuring_inventory_file.adoc
```diff
@@ -2255,7 +2255,11 @@ following variables can be used to customize the cluster console.
...
+need to be changed. You must create a ConfigMap containing the CA file and mount
```
kalexand-rh, line 2260: > ConfigMap that contains

install/configuring_inventory_file.adoc
```diff
@@ -2255,7 +2255,11 @@ following variables can be used to customize the cluster console.
...
+it into the `console` deployment in the `openshift-console` namespace at the
+same location as the variable.
```
kalexand-rh, line 2262: maybe
> location that you specify.

kalexand-rh (commented):
I have a couple of picks to consider!

install/configuring_inventory_file.adoc
```diff
@@ -2255,7 +2255,11 @@ following variables can be used to customize the cluster console.
 |Optional base path for the cluster console. If set, it should begin and end with a slash like `/console/`. Defaults to `/` (no base path).
 
 |`openshift_console_auth_ca_file`
-|Optional CA file to use to connect to the OAuth server. Defaults to `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`. Typically this does not need to be changed.
+|Optional CA file to use to connect to the OAuth server. Defaults to
+`/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`. Typically this does not
+need to be changed. You must create a ConfigMap containing the CA file and mount
+it into the `console` deployment in the `openshift-console` namespace at the
```
spadgett, line 2261: You might want to use `Deployment` with a capital here since you use `ConfigMap` above

ahardin-rh:
@yapei @spadgett @kalexand-rh Thank you! All comments are now addressed.

PR #20811, opened 2020-04-01
BZ#1726773 Enhance 4.x required ports tables

codyhoag:
https://bugzilla.redhat.com/show_bug.cgi?id=1726773

### DIFF
--- a/modules/installation-network-user-infra.adoc
+++ b/modules/installation-network-user-infra.adoc
@@ -52,8 +52,8 @@ of all other machines in the cluster.
 |Description
 
 |ICMP
-|`0`
-|
+|N/A
+|Network reachability tests
 
 .3+|TCP
 |`9000`-`9999`
--- a/modules/installation-network-user-infra.adoc
+++ b/modules/installation-network-user-infra.adoc
@@ -51,19 +51,23 @@ of all other machines in the cluster.
 |Port
 |Description
 
+|ICMP
+|`0`
+|
+
 .3+|TCP
 |`9000`-`9999`
 |Host level services, including the node exporter on ports `9100`-`9101` and
 the Cluster Version Operator on port `9099`.
 
-|`10249`-`10259`
+|`10250`-`10259`
 |The default ports that Kubernetes reserves
 
 |`10256`
 |openshift-sdn
 
 
-.4+|UDP
+.3+|UDP
 |`4789`
 |VXLAN and GENEVE
 
@@ -73,6 +77,7 @@ the Cluster Version Operator on port `9099`.
 |`9000`-`9999`
 |Host level services, including the node exporter on ports `9100`-`9101`.
 
+|TCP/UDP
 |`30000`-`32767`
 |Kubernetes NodePort
 
--- a/modules/installation-network-user-infra.adoc
+++ b/modules/installation-network-user-infra.adoc
@@ -51,13 +51,7 @@ of all other machines in the cluster.
 |Port
 |Description
 
-.5+|TCP
-|`2379`-`2380`
-|etcd server, peer, and metrics ports
-
-|`6443`
-|Kubernetes API
-
+.3+|TCP
 |`9000`-`9999`
 |Host level services, including the node exporter on ports `9100`-`9101` and
 the Cluster Version Operator on port `9099`.
@@ -81,6 +75,23 @@ the Cluster Version Operator on port `9099`.
 
 |`30000`-`32767`
 |Kubernetes NodePort
+
+|===
+
+.All machines to control plane
+[cols="2a,2a,5a",options="header"]
+|===
+
+|Protocol
+|Port
+|Description
+
+.2+|TCP
+|`2379`-`2380`
+|etcd server, peer, and metrics ports
+
+|`6443`
+|Kubernetes API
 
 |===
 


codyhoag:
Preview of additional table: https://required-ports--ocpdocs.netlify.com/openshift-enterprise/latest/installing/installing_bare_metal/installing-bare-metal-network-customizations.html#installation-network-user-infra_installing-bare-metal-network-customizations

codyhoag:
@sdodson can you confirm the accuracy of the additional table I added for the *All machines to control plane* table? Also, can you verify this applies to OCP 4.1-4.4?

From what I've gathered, this change aligns the docs with how we'd like to present our port requirements to the public for 4.x. I'll wrap up any additional edits for providing this info for other cloud platforms in another PR. Thanks!

sdodson:
/lgtm
Thank you @codyhoag

codyhoag:
@gpei PTAL?

gpei:
According to the Ingress rules set in upi-on-aws doc https://github.com/openshift/installer/blob/5e0bb6753f13da91ffa83eee2e5a99f411b6f5af/upi/aws/cloudformation/03_cluster_security.yaml 
I could see something different in the detailed Protocol&Port rules:

Ingress for masters from the VPC:
- ICMP
- TCP/22
- TCP/6443
- TCP/22623

Ingress for workers from the VPC:
- ICMP
- TCP/22

Ingress for masters from masters:
- TCP/2379-2380 "etcd"


Ingress for masters from masters and workers:
- UDP/4789 "Vxlan packets"
- UDP/6081 "Geneve packets"
- TCP/9000-9999 "Internal cluster communication"
- TCP/10250-10259 "Kubernetes kubelet, scheduler and controller manager"
- TCP/30000-32767 "Kubernetes ingress services"

Ingress for workers from masters and workers:
- UDP/4789 "Vxlan packets"
- UDP/6081 "Geneve packets"
- TCP/9000-9999 "Internal cluster communication"
- TCP/10250-10259 "Kubernetes kubelet, scheduler and controller manager"
- TCP/30000-32767 "Kubernetes ingress services"

sdodson:
@gpei Can you pinpoint what you're saying is different?

- ICMP
- TCP/22
- TCP/10250-10259 vs TCP/10249-10259

I guess we can add ICMP to the all machines to all machines portion.
The TCP/22 is not strictly required of the platform but is if you intend to SSH to hosts, I would expect admins to decide if that's necessary for their use or not and wouldn't document that.
We should fix the last item.

Anything else that I missed?

gpei:
> @gpei Can you pinpoint what you're saying is different?
> 
> * ICMP
> * TCP/22
> * TCP/10250-10259 vs TCP/10249-10259
> 
> I guess we can add ICMP to the all machines to all machines portion.
> The TCP/22 is not strictly required of the platform but is if you intend to SSH to hosts, I would expect admins to decide if that's necessary for their use or not and wouldn't document that.

+1 , TCP/22 should not be a strictly requirement. 

Besides these 3 items, the other differences I noticed:

- TCP/22623 should be opened on masters for MCS.
- TCP/2379-2380 should be control plane to control plane.
- 9000-9999 ports with only TCP configured in the AWS cloudformation template.
- 30000-32767 ports with TCP configured in the AWS cloudformation template.
I'm also not very clear about the port usage in the last two items, just noticed they're configured like that in the template.

sdodson:
> TCP/22623 should be opened on masters for MCS.

That's covered in the load balancer section, while it doesn't explicitly say that the firewall needs to be open I would assume that if the load balancer is used to connect to the control plane hosts on that port it need not be explicitly stated.

>     * 30000-32767 ports with TCP configured in the AWS cloudformation template.
>       I'm also not very clear about the port usage in the last two items, just noticed they're configured like that in the template.

This is the range used by NodePort and should be TCP and UDP on all hosts. @codyhoag lets fix that too.

> 9000-9999 ports with only TCP configured in the AWS cloudformation template.

That's out of sync with IPI terraform code, we'll fix the cloud formations template, docs are correct.

codyhoag:
@sdodson two questions:
1. What about the last difference in https://github.com/openshift/openshift-docs/pull/20811#issuecomment-614663660?
    >TCP/2379-2380 should be control plane to control plane.

    Should I move those ports to a new table for control plane to control plane?

2. What should I include as the description for ICMP (port `0`)?

codyhoag:
@sdodson when you have a moment, can you answer the two questions in my previous comment? Thanks!

sdodson:
@codyhoag 
#1 is correct as it is already, all hosts to control plane, this appears to be potentially a descrepancy between the reference that @gpei used versus what's defined in the IPI installation here https://github.com/openshift/installer/blob/master/data/data/aws/vpc/sg-master.tf#L226-L234

#2 ICMP doesn't have a port so put "N/A" and description of "Network reachability tests" or something to that effect would be fine. "ICMP" is also fine for a description, networking people should know what it means.

codyhoag:
@gpei can you take another look? Thanks!

Preview: https://required-ports--ocpdocs.netlify.app/openshift-enterprise/latest/installing/installing_bare_metal/installing-bare-metal-network-customizations.html#installation-network-user-infra_installing-bare-metal-network-customizations

gpei:
Thanks to the confirmation from @sdodson , lgtm now.

lamek:
LGTM.

codyhoag:
The manual 4.1 PR was sent in #21392

PR #20814, opened 2020-04-01
CNV-4285 not all bond modes supported

aburdenthehand:
Adding note that only bond modes 1,5,6 are supported for CNV 2.3. Also changed the mode used in the example.

enterprise-4.4 only

### DIFF
--- a/modules/cnv-example-bond-nncp.adoc
+++ b/modules/cnv-example-bond-nncp.adoc
@@ -8,6 +8,15 @@
 Create a bond interface on nodes in the cluster by applying a `NodeNetworkConfigurationPolicy` manifest
 to the cluster. 
 
+[NOTE]
+====
+{CNVProductNameStart} only supports the following bond modes:
+
+* mode=1 active-backup +
+* mode=5 balance-tlb +
+* mode=6 balance-alb
+====
+
 The following YAML file is an example of a manifest for a bond interface.
 It includes samples values that you must replace with your own information.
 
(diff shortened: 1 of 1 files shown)


modules/cnv-example-bond-nncp.adoc
```diff
@@ -8,6 +8,14 @@
 Create a bond interface on nodes in the cluster by applying a `NodeNetworkConfigurationPolicy` manifest
 to the cluster. 
 
+[NOTE]
+====
+{CNVProductNameStart} 2.3 only supports the following bond modes: +
```
phoracek, line 13: The documentation is already versioned. Do we need to keep the version here? Seems like something that can get outdated quite fast.

modules/cnv-example-bond-nncp.adoc
```diff
@@ -8,6 +8,14 @@
...
+* mode=1 active-backup +
```
phoracek, line 14: The asterisk is not rendered as a list item in the doc. Maybe we should drop them. I'm not an UX person though.

modules/cnv-example-bond-nncp.adoc
aburdenthehand: It'll be fixed before then though, right? :)

modules/cnv-example-bond-nncp.adoc
aburdenthehand: No, you're right. It was missing a linebreak. I either need glasses or to change my monitor resolution

phoracek:
/lgtm

Thanks!

phoracek (approved):

yossisegev:
LGTM (QE)

PR #20821, opened 2020-04-01
Updated upgrading section to add API group update

abrennan89:


serverless/installing_serverless/upgrading-serverless.adoc
```diff
@@ -8,6 +8,39 @@ toc::[]
 
 If you have previously installed a Technology Preview version of {ServerlessProductName}, follow the instructions in this guide to upgrade to the latest version.
 
+[id="update-api-group_{context}"]
+== Updating the API group
+
+In {ServerlessProductName} 1.4.0 and newer versions, the `KnativeServing` object's API group changed from `serving.knative.dev` to `operator.knative.dev`.
+You must adjust any of your scripts or applications that rely on the old API group to use the new API group.
+
+The {ServerlessOperatorName} creates a `KnativeServing` custom resource (CR) in the new API group for you.
```
mgencur, line 17: I'd say it would be better to mention "custom resource definition" (CRD) instead of CR. 
The Serverless operator creates this new CRD and later the user has to manually create the KnativeServing CR in order to install Knative. The Serverless operator doesn't automatically create KnativeServing CR
Something like this:
"The {ServerlessOperatorName} creates the `knativeserving` custom resource definition (CRD) in the new API group for you."

serverless/installing_serverless/upgrading-serverless.adoc
```diff
@@ -8,6 +8,39 @@ toc::[]
...
+
+After upgrading to {ServerlessProductName} 1.6.0, you must remove the old CRD to fix this issue.
+You can remove the old CRD by entering the following command:
+----
+$ oc delete crd knativeservings.serving.knative.dev
+----
+
+[id="upgrade-sub-channel_{context}"]
+== Upgrading the Subscription Channel
+
```
mgencur, line 43: Is this heading expected to be here? I don't see any removed lines in this PR so this is new and not just moved.

abrennan89: Yeah, it is. It's because we have two sub-sections now since I added **Updating the API group**
It's just to help conform with our modular docs approach and improve readability - make sense?

serverless/installing_serverless/upgrading-serverless.adoc
abrennan89: Ah OK. Odd, the release notes have the same content and were already QE'd. I will update here and in the release notes. Thanks!

mgencur: Yeah. That was probably an oversight.

serverless/installing_serverless/upgrading-serverless.adoc
```diff
@@ -8,6 +8,39 @@ toc::[]
 
 If you have previously installed a Technology Preview version of {ServerlessProductName}, follow the instructions in this guide to upgrade to the latest version.
 
+[id="update-api-group_{context}"]
+== Updating the API group
+
+In {ServerlessProductName} 1.4.0 and newer versions, the `KnativeServing` object's API group changed from `serving.knative.dev` to `operator.knative.dev`.
+You must adjust any of your scripts or applications that rely on the old API group to use the new API group.
+
+The {ServerlessOperatorName} creates a `KnativeServing` custom resource definition (CRD) in the new API group for you.
```
mgencur, line 17: @abrennan89 Sorry for confusion, Ashleigh. But I really meant changing just this single sentence to use CRD. It doesn't make so much sense in the other parts of documentation. We have CRD and CR (both of them), it's just the serverless operator will automatically create only the CRD. And the user has to manually create the CR.

mgencur: The user mostly manipulates with `CR`. The only part where they have to work with CRD is when removing it via `oc delete crd knativeservings.serving.knative.dev`

abrennan89: No worries, sorry it just seems a bit confusing that we go between the two. Is this accurate now?

mgencur:
QE Approved

serverless/installing_serverless/upgrading-serverless.adoc
```diff
@@ -8,6 +8,37 @@ toc::[]
 
 If you have previously installed a Technology Preview version of {ServerlessProductName}, follow the instructions in this guide to upgrade to the latest version.
 
+[id="update-api-group_{context}"]
+== Updating the API group
```
kalexand-rh, line 12: I'd make this section a separate procedure module. It sounds like you need to perform these steps after you upgrade, so I'd move it after that section.

serverless/installing_serverless/upgrading-serverless.adoc
```diff
@@ -8,6 +8,37 @@ toc::[]
...
+Once you have updated references to the new API group, you must remove any older custom resource (CR) versions and use the newly deployed `KnativeServing` CR instead.
+
+To safely do this without downtime, remove the owner reference from the newly deployed `KnativeServing` CR by entering the following command:
+----
+ $ oc edit knativeserving.operator.knative.dev knative-serving -n knative-serving
+----
+
+After the owner reference has been removed, you can safely remove any older CR versions and start using the new one.
+
+[IMPORTANT]
```
kalexand-rh, line 26: I might move this note after the next step.

serverless/installing_serverless/upgrading-serverless.adoc
```diff
@@ -8,6 +8,37 @@ toc::[]
...
+You must adjust any of your scripts or applications that rely on the old API group to use the new API group.
+
+Once you have updated references to the new API group, you must remove any older custom resource (CR) versions and use the newly deployed `KnativeServing` CR instead.
+
+To safely do this without downtime, remove the owner reference from the newly deployed `KnativeServing` CR by entering the following command:
+----
+ $ oc edit knativeserving.operator.knative.dev knative-serving -n knative-serving
+----
+
+After the owner reference has been removed, you can safely remove any older CR versions and start using the new one.
```
kalexand-rh, line 24: Is there any detriment to removing the old CR before you fix all these references? Is there a method for finding which ones to update? (If so, and it's not covered elsewhere, consider adding that here.)

serverless/installing_serverless/upgrading-serverless.adoc
```diff
@@ -8,6 +8,37 @@ toc::[]
...
+
+[IMPORTANT]
+====
+If a previous version of the CR exists, changes to the new CR will be overwritten by the {ServerlessOperatorName}. While the old CR is still active, all changes need to be made to that CR.
+====
+
+This change to the API group causes commands without a fully qualified API group and kind, such as `oc get knativeserving`, to become unreliable and not always work correctly.
```
kalexand-rh, line 31: ~I think this might make more sense after line 14.~ If this is only temporary, I might present it slightly differently. It goes from a "you need to know to not do this" fact to just extra context information that you could present after the step to remove the old CRD.

serverless/installing_serverless/upgrading-serverless.adoc
```diff
@@ -8,6 +8,37 @@ toc::[]
...
+
+After upgrading to {ServerlessProductName} 1.6.0, you must remove the old custom resource definition (CRD) to fix this issue.
```
kalexand-rh, line 33: Wait, so after you remove the old CRD, the not-fully-qualified commands work as expected again?

serverless/installing_serverless/upgrading-serverless.adoc
```diff
@@ -8,6 +8,37 @@ toc::[]
...
+You can remove the old CRD by entering the following command:
+----
+$ oc delete crd knativeservings.serving.knative.dev
+----
+
+[id="upgrade-sub-channel_{context}"]
```
kalexand-rh, line 39: I think this section should be a separate procedure module, too.

kalexand-rh (commented):
I have a couple of questions and suggestions!

serverless/installing_serverless/upgrading-serverless.adoc
abrennan89: cc @markusthoemmes wdyt?

serverless/installing_serverless/upgrading-serverless.adoc
abrennan89: I moved the note as requested, but I'm not sure if there's something else you are asking me to do with this?

serverless/installing_serverless/upgrading-serverless.adoc
markusthoemmes: Yep, cause there is no ambiguity about the name anymore.

serverless/installing_serverless/upgrading-serverless.adoc
markusthoemmes: Yeah if you delete before you remove the owner reference, everything will get deleted because of the ref being there. There's unfortunately no way of querying that really.

You could consider doing: `oc get knativeserving.operator.knative.dev --all-namespaced` to find them all and make sure none have any ownerReferences left.

serverless/installing_serverless/upgrading-serverless.adoc
abrennan89: @kalexand-rh is there an issue here or is it OK to resolve this comment now?

serverless/installing_serverless/upgrading-serverless.adoc
abrennan89: After discussions in Slack, I really don't think there's a need to add this step re `oc get knativeserving.operator.knative.dev --all-namespaced`. It refers to just listing all new CRs, which there should only be one of anyway, so it seems like it will just add confusion for users.

PR #20822, opened 2020-04-01
BZ1679419 change Ansible version from 2.5 to 2.6

jeana-redhat:
https://bugzilla.redhat.com/show_bug.cgi?id=1679419

### DIFF
--- a/install_config/topics/ocp_osp_deploy_instance.adoc
+++ b/install_config/topics/ocp_osp_deploy_instance.adoc
@@ -179,7 +179,7 @@ $ sudo subscription-manager repos \
     --enable="rhel-7-server-rpms" \
     --enable="rhel-7-server-extras-rpms" \
     --enable="rhel-7-server-ose-3.11-rpms" \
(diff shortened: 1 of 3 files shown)


jeana-redhat:
Hi @rlopez133 - 
I'm working on https://bugzilla.redhat.com/show_bug.cgi?id=1679419, which is asking to update the statement "Requires Ansible 2.5" to say "Requires Ansible 2.6" 
That's easy enough, but I noticed several other places on that page reference "rhel-7-server-ansible-2.5-rpms" and am thinking they should also be changed to "rhel-7-server-ansible-2.6-rpms" 
Let me know if that's right please!

bmcelvee:
LGTM!

rlopez133:
@jeana-redhat At the time of the publishing, that was the version we tested on. It would likely still work on Ansible 2.6 so I think that is a pretty safe change to make at this point.

jeana-redhat:
@rlopez133 Thank you!
@bmcelvee Can you do the things I don't have permission for yet? 
- Apply labels
- Merge
- CP to 3.11

PR #20834, opened 2020-04-01
Fix labelSelector syntax for PodAfinity docs - about

StevenBarre:
https://kubernetes.io/docs/concepts/configuration/assign-pod-node/

### DIFF
--- a/modules/nodes-scheduler-pod-affinity-about.adoc
+++ b/modules/nodes-scheduler-pod-affinity-about.adoc
@@ -41,7 +41,7 @@ spec:
   affinity:
     podAffinity: <1>
       requiredDuringSchedulingIgnoredDuringExecution: <2>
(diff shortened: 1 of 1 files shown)


kalexand-rh:
@lyman9966, will you PTAL and confirm which versions this change is valid for?

jboxman:
This has it as an array:

- https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#an-example-of-a-pod-that-uses-pod-affinity

And the API docs state that it is an array:

- https://docs.openshift.com/container-platform/4.5/rest_api/objects/index.html#podaffinity-core-v1

jboxman:
@sbarre-esit, thanks for reporting this. In the latest version of the page, it looks to be an array. If I'm wrong, feel free to re-open. Thanks!

PR #19721, opened 2020-02-13
Remove: apiserver cannot handle logLevel and operatorLogLevel

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1791888

New BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1816115
Need to set a CR(s) to Unmanaged?

Update: Per Tushar Underlying bug might get fixed. Topic might not need to be removed.
https://bugzilla.redhat.com/show_bug.cgi?id=1791888#c6

PR #20856, opened 2020-04-02
Fix audit logs

soltysh:
This came to my attention from https://github.com/openshift/api/issues/614.
There are several issues I've fixed in this PR:
1. you can't configure audit log as of yet
2. the output is not two lines any more but a structured JSON. 

I've also fixed minor issues here and there as I was touching this.

/assign @bergerhoffer


modules/nodes-nodes-audit-log-basic-viewing.adoc
```diff
@@ -32,7 +32,7 @@ ip-10-0-170-165.ec2.internal audit.log
 .. View the {product-title} API server log for a specific master node and timestamp or view all the logs for that master:
 +
 ----
-$ oc adm node-logs <node-ip> <log-name> --path=openshift-apiserver/<log-name>
+$ oc adm node-logs <node-ip> --path=openshift-apiserver/<log-name>
```
bergerhoffer, line 35: @soltysh You remove <log-name> here, but the two example below still provide that argument:
* audit-2019-04-08T13-09-01.227.log
* audit.log

Should those be removed from the examples?

modules/nodes-nodes-audit-log-basic-viewing.adoc
```diff
@@ -69,7 +69,7 @@ ip-10-0-170-165.ec2.internal audit.log
 .. View the Kubernetes API server log for a specific master node and timestamp or view all the logs for that master:
 +
 ----
-$ oc adm node-logs <node-ip> <log-name> --path=kube-apiserver/<log-name>
+$ oc adm node-logs <node-ip> --path=kube-apiserver/<log-name>
```
bergerhoffer, line 72: Same question here about whether `audit-2019-04-09T14-07-27.129.log` and `audit.log` should be removed from the examples below?

modules/nodes-nodes-audit-log-basic.adoc
```diff
@@ -6,27 +6,12 @@
...
-. The request line containing:
-.. A Unique ID allowing to match the response line (see #2)
-.. The source IP of the request
-.. The HTTP method being invoked
-.. The original user invoking the operation
-.. The impersonated user for the operation (`self` meaning himself)
-.. The impersonated group for the operation (`lookup` meaning user's group)
-.. The namespace of the request or <none>
-.. The URI as requested
+The audit policy object structure is defined in the `audit.k8s.io` API group.
```
bergerhoffer, line 10: What does this mean? How/where would a user actually go to find the structure? I don't see this in our API docs at all.

bergerhoffer (commented):
Thanks for the updates @soltysh! Just a few questions if you could help clarify.

vikram-redhat:
> you can't configure audit log as of yet

@soltysh does that mean we had docs on this for the 4.x stream but the feature isn't/wasn't working? If yes, @bergerhoffer we need to go through change control on this.

modules/nodes-nodes-audit-log-basic-viewing.adoc
soltysh: Good catch - fixed.

modules/nodes-nodes-audit-log-basic-viewing.adoc
soltysh: Fixed.

modules/nodes-nodes-audit-log-basic.adoc
soltysh: Yeah, I couldn't find those either, I'm not sure how we generate API docs these days but we're missing this bit.

soltysh: https://github.com/kubernetes/kubernetes/blob/ea3eeb494a528edcc02f8887d3098f33fca547b3/staging/src/k8s.io/apiserver/pkg/apis/audit/v1/types.go#L72-L136 is the structure that describes this. Not sure how we could expose that best.

soltysh:
> @soltysh does that mean we had docs on this for the 4.x stream but the feature isn't/wasn't working? If yes, @bergerhoffer we need to go through change control on this.

Future is working, but its configuration was/is not possible.

soltysh:
I've pushed with updated changes that @bergerhoffer pointed out.

vikram-redhat:
> @vikram-redhat Let me know if you disagree, but I don't think this requires change control (unless we're broadening the scope for what we're doing that for).

@bergerhoffer If I understand this right, we had documented previously that the configuration of audit logs is possible (https://docs.openshift.com/container-platform/4.3/nodes/nodes/nodes-nodes-audit-log.html#nodes-nodes-audit-log-basic-config_nodes-nodes-audit-log), which your new PR (when merged) will effectively say is not. This is a degradation of a previously known documented feature and within change control as defined currently.

mburke5678:
This issue has been through change control and I received approval to remove the topic last week.
I sent an email to aos-devel and openshift-sme on Firday, April 3. 
https://github.com/openshift/openshift-docs/pull/19721

PR #20877, opened 2020-04-03
Add installing on IBM Power, add installing on IBM Power with restricted network, updates to installing on IBM Z

rh-max:


_topic_map.yml
```diff
@@ -199,6 +199,12 @@ Topics:
   Topics:
   - Name: Installing a cluster on IBM Z
     File: installing-ibm-z
+- Name: Installing on IBM Power
```
kalexand-rh, line 202: @vikram-redhat, we're going to need a new book on the portal.

vikram-redhat:
@rh-max is this ready to go? And for what versions would this be for?

modules/installation-requirements-user-infra.adoc
```diff
@@ -140,7 +147,9 @@ On your z/VM instance, set up:
 [discrete]
 === Disk storage for the z/VM guest virtual machines
 
-* FICON attached disk storage (DASDs). These can be z/VM minidisks, fullpack minidisks, or dedicated DASDs. If available, use HyperPAV to ensure optimal performance.
+* FICON attached disk storage (DASDs). These can be z/VM minidisks, fullpack minidisks, or dedicated DASDs.
+ifdef::ibm-z[To reach the minimum required DASD size for RHCOS installations, you need extended address volumes (EAV).]
```
bergerhoffer, line 151: Is this ifdef::ibm-z really necessary? It looks like it's already inside a big ibm-z ifdef on line 121.

modules/installation-requirements-user-infra.adoc
```diff
@@ -173,7 +183,9 @@ On your z/VM instances, set up:
...
```
bergerhoffer, line 187: Same question as above on whether this ifdef is necessary.

modules/installation-requirements-user-infra.adoc
```diff
@@ -158,6 +167,7 @@ On your z/VM instance, set up:
 
 * 3 LPARs with 6 IFLs that support SMT2
 * 1 or 2 OSA or RoCE network adapters, or both
+* Hipersockets, which are attached to a node either directly as a device or by bridging with one zVM vSwitch to be transparent to the z/VM guest. To directly connect Hipersockets to a node you must set up a gateway to the external network via a RHEL 8 guest to bridge to the Hipersockets network. See the "Additional resources" section.
```
bergerhoffer, line 170: Suggest against the last sentence to see the additional resources section.

bergerhoffer: Also the first sentence uses both `zVM` and `z/VM`. Should they both be the same, or are they different?

installing/installing_ibm_power/installing-ibm-power.adoc
```diff
@@ -0,0 +1,91 @@
...
+[NOTE]
+====
+Be sure to also review this site list if you are configuring a proxy.
+====
+
+include::modules/cluster-entitlements.adoc[leveloffset=+1]
+
+include::modules/installation-requirements-user-infra.adoc[leveloffset=+1]
+
+.Additional resources
```
bergerhoffer: I don't think we're supposed to have an additional resources section right in the middle of an assembly like this. But I'm not positive.

installing/installing_ibm_power/installing-ibm-power.adoc
```diff
@@ -0,0 +1,91 @@
...
+include::modules/installation-initializing-manual.adoc[leveloffset=+1]
+
+include::modules/installation-bare-metal-config-yaml.adoc[leveloffset=+2]
+
+include::modules/installation-user-infra-generate-k8s-manifest-ignition.adoc[leveloffset=+1]
+
+[id="creating-machines-bare-metal-power"]
+== Creating {op-system-first} machines
+
+Before you install a cluster on bare metal infrastructure that you provision,
```
bergerhoffer, line 59: Should this say something other than bare metal, or is this correct?

bergerhoffer (commented):
Submitting the feedback so far. Will continue reviewing shortly.

modules/installation-bare-metal-config-yaml.adoc
```diff
@@ -7,9 +7,23 @@
 ifeval::["{context}" == "installing-restricted-networks-bare-metal"]
 :restricted:
 endif::[]
+ifeval::["{context}" == "installing-ibm-z"]
```
bergerhoffer, line 10: Is there a reason that IBM Z and power are using bare metal's config instead of having their own file (like the other cloud providers do)?

modules/installation-bare-metal-config-yaml.adoc
```diff
@@ -103,7 +117,10 @@ load balancers and routers to manage the traffic.
 one IP address pool. If you need to access the services from an external network,
 configure load balancers and routers to manage the traffic.
 <10> You must set the platform to `none`. You cannot provide additional platform
-configuration variables for bare metal infrastructure.
+configuration variables for
+ifndef::ibm-z[bare metal]
```
bergerhoffer, line 121: Does power need to be represented here, or is it okay that it says bare metal?

modules/installation-ibm-z-troubleshooting-and-debugging.adoc
```diff
@@ -0,0 +1,43 @@
+// Module included in the following assemblies:
+//
+// * installing/installing_ibm_z/installing-ibm-z.adoc
+
+[id="installation-ibm-z-troubleshooting-and-debugging_{context}"]
+= Completing installation on user-provisioned infrastructure
```
bergerhoffer, line 6: I'm really confused as to what this module is supposed to be. If you look at the assembly for installing on IBM Z/LinuxOne, it shows the same thing twice:

```
Completing installation on user-provisioned infrastructure
Completing installation on user-provisioned infrastructure
```

But it does look like it's supposed to be troubleshooting/debugging info like the file is named. Is the title just wrong?

modules/installation-ibm-z-troubleshooting-and-debugging.adoc
```diff
@@ -0,0 +1,43 @@
...
+
+You can gather the debugging information that might help you to troubleshoot and
```
bergerhoffer, line 8: the debugging information -> debugging information

modules/installation-ibm-z-troubleshooting-and-debugging.adoc
```diff
@@ -0,0 +1,43 @@
...
+debug certain issues with {product-title} installation on IBM Z.
```
bergerhoffer, line 9: with {product-title} installation -> with an {product-title} installation

modules/installation-ibm-z-troubleshooting-and-debugging.adoc
```diff
@@ -0,0 +1,43 @@
...
+
+.Prerequisites
+
+* The `oc` CLI tool installed.
+
+.Procedure
+
+. Log into the cluster:
```
bergerhoffer, line 17: Log into -> Log in to

modules/installation-ibm-z-troubleshooting-and-debugging.adoc
```diff
@@ -0,0 +1,43 @@
...
+$ toolbox
+----
+
+. Collect the `dbginfo` data:
++
+----
+$ dbginfo.sh
+----
+
+. You get then retrieve the data, for example, using `scp`.
```
bergerhoffer, line 43: This sentence is confusing. Should it be "You can then retrieve..."?

modules/installation-installing-bare-metal.adoc
```diff
@@ -36,12 +36,12 @@ endif::restricted[]
...
 ----
-$ ./openshift-install --dir=<installation_directory> wait-for bootstrap-complete \ <1>
-    --log-level=info <2>
-
-INFO Waiting up to 30m0s for the Kubernetes API at https://api.test.example.com:6443...
-INFO API v1.14.6+c4799753c up
-INFO Waiting up to 30m0s for the bootstrap-complete event...
+$ ./openshift-install --dir=/root/manual_installation/installation_directory wait-for bootstrap-complete \
+    --log-level=info
+INFO Waiting up to 30m0s for the Kubernetes API at https://api.m3559ocp.lnxne.boe:6443...
```
bergerhoffer, line 41: Just wondering if this should stay `api.test.example.com` instead of `api.m3559ocp.lnxne.boe`

modules/installation-installing-bare-metal.adoc
```diff
@@ -36,12 +36,12 @@ endif::restricted[]
 . Monitor the bootstrap process:
 +
 ----
-$ ./openshift-install --dir=<installation_directory> wait-for bootstrap-complete \ <1>
-    --log-level=info <2>
-
-INFO Waiting up to 30m0s for the Kubernetes API at https://api.test.example.com:6443...
-INFO API v1.14.6+c4799753c up
-INFO Waiting up to 30m0s for the bootstrap-complete event...
+$ ./openshift-install --dir=/root/manual_installation/installation_directory wait-for bootstrap-complete \
```
bergerhoffer, line 39: Should this stay `<installation_directory>` instead of `/root/manual_installation/`?

You removed the callouts here but not below the code block.

modules/installation-operators-config.adoc
```diff
@@ -52,5 +57,43 @@ service-catalog-apiserver            4.4.0     True        False         False
 service-catalog-controller-manager   4.4.0     True        False         False      5m25s
 storage                              4.4.0     True        False         False      5m30s
 ----
+endif::ibm-z[]
+ifdef::ibm-z[]
+----
+$ oc get clusteroperators
+NAME                                       VERSION                                   AVAILABLE   PROGRESSING   DEGRADED   SINCE
+authentication                             4.3.0-0.nightly-s390x-2020-03-09-183623   True        False         False      40d
```
bergerhoffer, line 65: Is it correct for the versions here to say 4.3 nightly?

modules/installation-operators-config.adoc
```diff
@@ -52,5 +57,43 @@ service-catalog-apiserver            4.4.0     True        False         False
 service-catalog-controller-manager   4.4.0     True        False         False      5m25s
 storage                              4.4.0     True        False         False      5m30s
 ----
+endif::ibm-z[]
+ifdef::ibm-z[]
+----
+$ oc get clusteroperators
```
bergerhoffer, line 63: Does this command need watch like the other non-ibm z one above (since the command says "Watch the cluster components come online")?

modules/installation-user-infra-machines-pxe.adoc
```diff
@@ -107,3 +146,10 @@ You must create the bootstrap and control plane machines at this time. Because
 some pods are deployed on compute machines by default, also create at least two
 compute machine before you install the cluster.
 ====
+
+ifeval::["{context}" == "installing-ibm-power"]
```
bergerhoffer, line 150: It looks like you need to unset the variables like how they were set. You need to unset both variables for both ifevals

modules/registry-configuring-storage-baremetal.adoc
```diff
@@ -5,18 +5,42 @@
...
 
 * Cluster administrator permissions.
-* A cluster on bare metal.
-* Provision persistent storage
-for your cluster, such as Red Hat OpenShift Container Storage. To deploy a private image registry, your storage must provide
+* A cluster on
+ifndef::ibm-z,ibm-power[bare metal]
+ifdef::ibm-z[IBM Z]
+ifdef::ibm-power[IBM Power]
+.
```
bergerhoffer, line 39: There is a space after each of these in front of the period. Suggest moving the period to after each of the above bare metal/IBM Z/IBM Power

modules/registry-configuring-storage-baremetal.adoc
```diff
@@ -5,18 +5,42 @@
...
+* Provision persistent storage for your cluster, such as
+ifndef::ibm-z[Red Hat OpenShift Container Storage]
+ifdef::ibm-z[NFS]
```
bergerhoffer, line 42: Put the period after each of these two as well. This is the only bullet in this list without a period.

modules/registry-configuring-storage-baremetal.adoc
```diff
@@ -5,18 +5,42 @@
...
+. To deploy a private image registry, your storage must provide
```
bergerhoffer, line 43: This is rendering as a number 1 here, which I don't think was the intention. If you move the period as suggested above, it should get rid of that problem.

bergerhoffer (commented):
A few more bits of feedback. And don't forget to squash.

Also, is this just for 4.4, or does it need cherry picked any farther back?

bergerhoffer:
@rh-max - @vikram-redhat mentioned that this might be for 4.3 (and not 4.4)? If so, then it should be targeted against enterprise-4.3 instead of master.

modules/installation-requirements-user-infra.adoc
rh-max: Good catch, removed the inner conditional.

modules/installation-requirements-user-infra.adoc
rh-max: Same, removed.

modules/installation-requirements-user-infra.adoc
rh-max: Yeah, that's the first time I've referred to the Additional resources from main text. In this particular case, the hint to the reader that there's more info on this is really lacking, IMO. But I've removed it to comply with the guidelines.
Yes, should be z/VM, fixed.

installing/installing_ibm_power/installing-ibm-power.adoc
rh-max: That's the solution Vikram allowed because of the recent directive to not keep any resources inside modules. I don't know where we'd put them otherwise.
But if you mean to put them all at the very end of the assembly, then IMO we might as well get rid of them. They'd be hopelessly disconnected from the place they are relevant to and there would potentially be a big bunch of them together at the end (from the whole assembly content).
So, unless there's a directive to not do this, I'd definitely keep them here.

installing/installing_ibm_power/installing-ibm-power.adoc
rh-max: Should be IBM Power, fixed.

modules/installation-bare-metal-config-yaml.adoc
rh-max: To make use of modularity.
Plus IBM Z was already using the bare metal's config when I started work, so I went with it. Then, when creating the IBM Power one, it was very similar to the IBM Z one, so I did it the same way.
I think the real (small) problem here is the naming of this thrice-used file. Do you agree that I should rename it into `installation-config-yaml.adoc`? To remove the implication that it's supposed to be about bare metal and the other two are living off it.

modules/installation-bare-metal-config-yaml.adoc
rh-max: Good catch, added the IBM Power conditional.

modules/installation-ibm-z-troubleshooting-and-debugging.adoc
rh-max: That's right, I somehow used the wrong title. Fixed to be "Collecting debugging information".

modules/installation-ibm-z-troubleshooting-and-debugging.adoc
rh-max: Fixed.

modules/installation-ibm-z-troubleshooting-and-debugging.adoc
rh-max: Fixed.

modules/installation-ibm-z-troubleshooting-and-debugging.adoc
rh-max: Yep, fixed.

modules/installation-installing-bare-metal.adoc
rh-max: Are we allowed to just change the outputs, trading their "realism" for better presentation? If yes, I'll change this and in a couple of other outputs.

modules/installation-installing-bare-metal.adoc
rh-max: Correct, changed to <installation_directory> and returned the callouts.

modules/installation-operators-config.adoc
rh-max: Yeah, `nightly` isn't acceptable. Do you think changing `4.3.0-0.nightly-s390x-2020-03-09-183623` to `4.3.0-s390x` is reasonable here?

modules/installation-operators-config.adoc
rh-max: Yep, fixed.

modules/installation-user-infra-machines-pxe.adoc
rh-max: That was bad. Fixed.

modules/registry-configuring-storage-baremetal.adoc
rh-max: Done.

modules/registry-configuring-storage-baremetal.adoc
rh-max: Done. It's not actually a bullet, it's the first sentence of a bullet.

modules/registry-configuring-storage-baremetal.adoc
rh-max: Yes, fixed.

rh-max:
> @rh-max - @vikram-redhat mentioned that this might be for 4.3 (and not 4.4)? If so, then it should be targeted against enterprise-4.3 instead of master.

Ah, I forgot to open it against `enterprise-4.3`. Will do. Plus there'll also be a small PR for `enterprise-4.2` that will exclude IBM Power from it (as it's only for 4.3) and will exclude some IBM Z things that only apply to 4.3.

rh-max:
@bergerhoffer And, to your previous question, this goes to `enterprise-4.2` and `enterprise-4.3`.

installing/installing_ibm_power/installing-ibm-power.adoc
bergerhoffer: Fair enough!

modules/installation-bare-metal-config-yaml.adoc
bergerhoffer: Fair enough. The name of the file is what caused me to ask. Though I don't know whether it's the right thing to do, to change the file name, since it is still only for a subset of the install options. Up to you, but I'd say it's not critical.

I would at least add the power assemblies that it's newly included for to the list of assemblies at the top of the module.

modules/installation-installing-bare-metal.adoc
bergerhoffer: Absolutely, I do it all the time. Better to present a generic example instead of the exact machine/cluster info from where we actually ran the tests.

modules/installation-operators-config.adoc
bergerhoffer: I would say that's at least better than nightly, but it would be good to find out what the actual output would be here for a 4.3 release.

modules/installation-bare-metal-config-yaml.adoc
```diff
@@ -103,7 +117,11 @@ load balancers and routers to manage the traffic.
 one IP address pool. If you need to access the services from an external network,
 configure load balancers and routers to manage the traffic.
 <10> You must set the platform to `none`. You cannot provide additional platform
-configuration variables for bare metal infrastructure.
+configuration variables for
+ifndef::ibm-z,ibm-power[bare metal]
+ifdef::ibm-z[IBM Z]
+ifdef::ibm-power[IBM Power]
```
bergerhoffer, line 123: This (and the heading for the module) still say bare metal for IBM Power. Looks like the top of the page only sets the ibm-power attribute for the restricted network version.

bergerhoffer (commented):
Just one more issue

bergerhoffer:
@rh-max Other than my last comment and a few other open conversations that you might need to act on, these changes look pretty good to me. Other than applying these changes directly to enterprise-4.3, and then commits squashed still. 

However, If IBM power doesn't apply for 4.2, then I don't think you should cherry pick this PR back to 4.2 and then do another PR to fixup. You might run a risk of content inappropriate to 4.2 showing up live in the docs. You should just submit a separate PR directly to 4.2 that only includes the relevant parts of this PR for 4.2.

modules/installation-bare-metal-config-yaml.adoc
rh-max: Hmm, you are right. I couldn't come up with a reasonable name which would indicate that it's just for a subset. Leaving it as it is, might address in a future release.

modules/installation-installing-bare-metal.adoc
rh-max: Great, will keep it in mind. Changed to `api.test.example.com` and changed in a few other places.

modules/installation-operators-config.adoc
rh-max: Changed.

modules/installation-bare-metal-config-yaml.adoc
rh-max: Yes, fixed.

rh-max:
@bergerhoffer The squashing and change to enterprise-4.3 - ACK.

Regarding handling the difference between 4.2 and 4.3. What you describe would make future changes that apply to both 4.2 and 4.3+ uncherrypickable from 4.3+ to 4.2 because of the conditionals that deal with both IBM Z and IBM Power. Every such change would need to be done for 4.2 differently, which is a huge and unnecessary overhead.
I don't think that we lose much by putting IBM Power material into enterprise-4.2 but then removing it from _topic_map.yaml (in enterprise-4.2). I don't think that the risk of inappropriate content appearing in 4.2 is that high. I can't see how parts pertaining to IBM Power can appear in the text, since they rely on the correct context set in the assembly file, and the Power assemblies won't be included in _topic_map.yaml. For good measure, even the Power assemblies can be deleted. That way, it will only remain as those conditionals.

bergerhoffer:
@rh-max I see what you're getting at for the conditionals. And I think I'd be fine with that but yeah as long as the power assemblies are removed from the topic map, and yes I think the assemblies should be deleted as well.

One of the points I was trying to make is that if you cherry pick this PR to 4.2, and it takes you a bit to get that follow up PR to remove things from 4.2, power might temporarily show up on 4.2.

What I'd recommend then, is to merge this to 4.3, and then do a manual cherry pick of this to 4.2, removing the power assemblies and power from the topic map (and any other updates needed). Just so there's no chance for incorrect updates to show up on 4.2, however temporarily.

rh-max:
@bergerhoffer Great, agree with every point, going ahead with that!

rh-max:
@bergerhoffer The 4.3 PR is https://github.com/openshift/openshift-docs/pull/21723.

rh-max:
Closing this in favor of https://github.com/openshift/openshift-docs/pull/21723 and https://github.com/openshift/openshift-docs/pull/21727.

PR #20878, opened 2020-04-03
Bug 1816773: add information about the possibility to remove UsingDeprecatedAPIExtensionsV1Beta1 alert

soltysh:
We removed the alert `UsingDeprecatedAPIExtensionsV1Beta1` as part of https://bugzilla.redhat.com/show_bug.cgi?id=1795617 but after the upgrade the alert still persists as reported in https://bugzilla.redhat.com/show_bug.cgi?id=1816773. We figured out that we'll inform users they can safely remove that alert if they wish to.

This is ONLY for 4.3. 
/assign @bergerhoffer

### DIFF
--- a/monitoring/cluster_monitoring/managing-cluster-alerts.adoc
+++ b/monitoring/cluster_monitoring/managing-cluster-alerts.adoc
@@ -13,6 +13,7 @@ include::modules/monitoring-silencing-alerts.adoc[leveloffset=+1]
 include::modules/monitoring-getting-information-about-silences.adoc[leveloffset=+1]
 include::modules/monitoring-editing-silences.adoc[leveloffset=+1]
 include::modules/monitoring-expiring-silences.adoc[leveloffset=+1]
+include::modules/monitoring-known-issues.adoc[leveloffset=+1]
 
 .Next steps
 
--- a/modules/monitoring-known-issues.adoc
+++ b/modules/monitoring-known-issues.adoc
@@ -1,0 +1,36 @@
+// Module included in the following assemblies:
+//
+// * monitoring/cluster_monitoring/managing-cluster-alerts.adoc
+
+[id="known-issues_{context}"]
+= Known Issues
+
+For short period of time `UsingDeprecatedAPIExtensionsV1Beta1` alert existed
+in {product-title} version 4.3. Starting from version 4.3.8 the alert is no
+more required, if you notice it is firing in your cluster and you are on a
+{product-title} version 4.3.8 or newer you can safely remove that `PrometheusRule`.
+
+
+.Procedure
+
+. Verify your cluster version:
++
+----
+$ oc get clusterversion
+----
+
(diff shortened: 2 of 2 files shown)


monitoring/cluster_monitoring/managing-cluster-alerts.adoc
```diff
@@ -13,6 +13,7 @@ include::modules/monitoring-silencing-alerts.adoc[leveloffset=+1]
 include::modules/monitoring-getting-information-about-silences.adoc[leveloffset=+1]
 include::modules/monitoring-editing-silences.adoc[leveloffset=+1]
 include::modules/monitoring-expiring-silences.adoc[leveloffset=+1]
+include::modules/monitoring-known-issues.adoc[leveloffset=+1]
```
bergerhoffer, line 16: @maximsvistunov Do you have any thoughts on how to add this information (about removing this alert that is no longer required) to the alerts section? Should we keep it as a known issues section like this? Or do you think we should be more specific and call it "Removing the UsingDeprecatedAPIExtensionsV1Beta1 alerting rule" or something similar?

bergerhoffer:
@maximsvistunov Would you be able to give us some input on the addition of this information?

rh-max:
@bergerhoffer Sorry for the delay. (It's better to tag @rh-max). I would prefer having a known issues section, which then contains a collection of the specific issues (only one in this case). So something like:

```
= Expiring silences
= Known Issues
== Removing the `UsingDeprecatedAPIExtensionsV1Beta1` alert
introduction
steps
```

Unfortunately, we need to keep the structure flat. Your suggestion:

```
= Expiring silences
= Known Issues
introduction
steps
```

is flat, but I think the combination of the title "Known Issues" and of the content (description of one specific alert) doesn't work, it is a mismatch. So the only viable option, I think, is to abandon any "Known Issues" structure and just have the section that deals specifically with this alert:

```
= Expiring silences
= Removing the `UsingDeprecatedAPIExtensionsV1Beta1` alert
introduction
steps
```

I must say that such additions really clutter the page https://docs.openshift.com/container-platform/4.3/monitoring/cluster_monitoring/managing-cluster-alerts.html and break the flow (notice how consecutive pages under `Monitoring -> Cluster monitoring` suggest to the reader what to do using the "Next steps" section at the end). @sichvoge, do you have a strong opinion on the placement of this text?

bergerhoffer:
Thanks @rh-max (and sorry for the wrong GH tag)! I've overtaken this update and opened a new PR w/ some cleanup and renaming the section away from Known Issues. I can of course update per feedback from @sichvoge, but at least wanted to get these updates ready to help see how it would look. Let me know what you both think.

New PR: https://github.com/openshift/openshift-docs/pull/21247
Preview: http://file.rdu.redhat.com/~ahoffer/2020/pr-20878-update/monitoring/cluster_monitoring/managing-cluster-alerts.html

bergerhoffer:
Closing this PR in favor of #21247.

PR #20880, opened 2020-04-03
CNV - BZ#1818068 removing supplmentary NAD example

aburdenthehand:
BZ#1818068, removing the supplementary example for adding NAD outside of namespace. This is not possible.

cp to enterprise-4.3 and enterprise-4.4

### DIFF
--- a/modules/cnv-creating-bridge-nad-cli.adoc
+++ b/modules/cnv-creating-bridge-nad-cli.adoc
@@ -8,6 +8,11 @@
 As a network administrator, you can configure a NetworkAttachmentDefinition
 of type `cnv-bridge` to provide Layer-2 networking to Pods and virtual machines.
 
+[NOTE]
+====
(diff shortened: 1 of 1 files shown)


modules/cnv-creating-bridge-nad-cli.adoc
```diff
@@ -66,19 +66,6 @@ spec:
 ----
 <1> You must substitute the actual `name` value from the
 NetworkAttachmentDefinition.
-+
```
phoracek, line 69: Can we add a note explaining that the NetworkAttachmentDefinition is available only to VMs within the same Namespace?

phoracek (changes requested):

phoracek (approved):
Thanks!

yossisegev:
LGTM (CNV Network QE)

PR #20885, opened 2020-04-03
BZ1731358 Added valid time units, fixed typo.

jeana-redhat:
Addressing https://bugzilla.redhat.com/show_bug.cgi?id=1731358

### DIFF
--- a/install_config/aggregate_logging.adoc
+++ b/install_config/aggregate_logging.adoc
@@ -493,7 +493,7 @@ When `openshift_logging_es_pvc_prefix` is set, and:
 * `openshift_logging_es_pvc_dynamic`=`false`, the value for `openshift_logging_es_pvc_size` must be set.
 
 |`openshift_logging_es_recover_after_time`
-|The amount of time Elasticsearch will wait before it tries to recover.
+|The amount of time Elasticsearch will wait before it tries to recover. Supported time units are seconds (s) or minutes (m).
 
 |`openshift_logging_es_storage_group`
 |Number of a supplemental group ID for access to Elasticsearch storage volumes.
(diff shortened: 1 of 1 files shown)


jeana-redhat:
@anpingli please check this change for BZ1731358 :)

install_config/aggregate_logging.adoc
```diff
@@ -552,8 +552,8 @@ when `openshift_logging_use_ops` is set to `true`.
 when `openshift_logging_use_ops` is set to `true`.
 
 |`openshift_logging_es_ops_recover_after_time`
```
anpingli, line 554: The variable doesn't work. https://bugzilla.redhat.com/show_bug.cgi?id=1822093

anpingli: @jcantrill  Will we still fix the low bug in 3.11?

jcantrill: I dont see us fixing this unless there is a customer case that blocks this.  "Ops" deployments are uncommon to begin with

jeana-redhat: Missed this discussion when I was looking at the comments on the bugs themselves earlier. Should the doc change be posted as proposed above then?
@vikram-redhat FYI I'm not sure how to proceed here.

vikram-redhat: @jeana-redhat we should remove this variable and its explanation since it doesn't work. Normally, I wouldn't advocate for this and request a separate bug but in this case the change about time units applied to this variable as well.

jeana-redhat:
@anpingli Ready for a re-check :)

anpingli (commented):
LGTM

anpingli (commented):
LGTM

mburke5678:
@jeana-redhat LGTM

PR #20890, opened 2020-04-03
OSDOCS-893: Updating procedure to replace a single unhealthy etcd member

bergerhoffer:
https://issues.redhat.com/browse/OSDOCS-893

Preview: https://osdocs-893-single-member--ocpdocs.netlify.com/openshift-enterprise/latest/backup_and_restore/replacing-unhealthy-etcd-member.html


modules/restore-determine-state-etcd-member.adoc
```diff
@@ -0,0 +1,78 @@
...
++
+----
+$ oc get pods -n openshift-etcd | grep etcd
+etcd-ip-10-0-131-183.ec2.internal                2/3     Error       7          6h9m <1>
+etcd-ip-10-0-164-97.ec2.internal                 3/3     Running     0          6h6m
+etcd-ip-10-0-154-204.ec2.internal                3/3     Running     0          6h6m
+----
+<1> Since this status of this Pod is `Error`, then the *node is running, but etcd is crashlooping*.
+
+// TODO: Are there any other possible situations other than the above 2 that an unhealthy member could be in? If so, which procedure should be the default?
```
bergerhoffer, line 78: @alaypatel07 PTAL at this question

modules/restore-replace-crashlooping-etcd-member.adoc
```diff
@@ -0,0 +1,125 @@
...
+sh-4.2# chroot /host
+----
+
+.. Move the existing etcd Pod file out of the Kubelet manifest directory:
++
+----
+sh-4.4# mv /etc/kubernetes/manifests/etcd-pod.yaml /var/lib/etcd-backup/
+----
++
+// TODO: Confirm sh-4.4# prompt is okay.
```
bergerhoffer, line 47: @alaypatel07 When I tested, the prompt was 4.2 after you oc debug, then changed to 4.4 after you run `chroot /host`. Just wanted to verify with you that that's correct as shown here.

modules/restore-replace-crashlooping-etcd-member.adoc
```diff
@@ -0,0 +1,125 @@
...
+. Force etcd redeployment.
++
+In a terminal that has access to the cluster as a `cluster-admin` user, run the following command:
++
+----
+$ oc patch etcd cluster -p='{"spec": {"forceRedeploymentReason": "single-master-recovery-'"$( date --rfc-3339=ns )"'"}}' --type=merge <1>
+----
+<1> The `forceRedeploymentReason` value must be unique, which is why a timestamp is appended.
++
+// TODO: Add a statement similar to what we used for the other DR procedure: "When the etcd cluster Operator performs a redeployment, the existing nodes are started with new Pods similar to the initial bootstrap scale up."
```
bergerhoffer, line 116: @alaypatel07 The sentence I mention here is what we have after forcing etcd deployment in the new restore procedure here [1]. Can we adjust this sentence to add here? Would this be accurate: 

"When the etcd cluster Operator performs a redeployment, a new member is started with new Pods similar to the initial bootstrap scale up."

or just?

"When the etcd cluster Operator performs a redeployment, a new member is started similar to the initial bootstrap scale up."

[1] https://github.com/openshift/openshift-docs/pull/20053/files#diff-4445e0411093c0b0dc43b767e97b0557R131

modules/restore-replace-stopped-etcd-member.adoc
```diff
@@ -0,0 +1,131 @@
...
+
+... Update the `selfLink` field to use the new machine name from the previous step.
+
+.. Delete the machine of the unhealthy member:
++
+----
+$ oc delete machine ci-ln-mpprd6t-d5d6b-n94jk-master-0 -n openshift-machine-api
+----
++
+// TODO: This command takes a few minutes to come back, just wait it out?
```
bergerhoffer, line 123: @alaypatel07 Deleting the machine takes a while. It's a few minutes before the prompt is returned to you in the terminal. Should we tell users to wait until the prompt is returned before proceeding to the next step? Should we tell them to wait until the machine is actually deleted (which is a few more minutes past the prompt returning) before going to the next step?

backup_and_restore/replacing-unhealthy-etcd-member.adoc
```diff
@@ -0,0 +1,37 @@
+[id="replacing-unhealthy-etcd-member"]
+= Replacing an unhealthy etcd member
+include::modules/common-attributes.adoc[]
+:context: replacing-unhealthy-etcd-member
+
+toc::[]
+
+This document describes the process to replace a single unhealthy etcd member.
+
+This procedure assumes that there is still an etcd quorum in the cluster. The process to replace the unhealthy member depends on whether the etcd member is stopped, or whether it is running but etcd is crashlooping.
```
alaypatel07, line 10: @bergerhoffer I don't understand in what case the member will be stopped, we probably need to reword

```
The process to replace the unhealthy member depends on whether the etcd member is stopped,

File: modules/restore-replace-stopped-etcd-member.adoc [hunk 6]
```diff
@@ -0,0 +1,131 @@
...
+
+.. Create the new machine using the `new-master-machine.yaml` file.
++
+----
+$ oc apply -f new-master-machine.yaml
+----
+
+// TODO: Need to add a final verification step. The process seems to take quite a few minutes.
```
bergerhoffer, line 131: @alaypatel07 Can you recommend what kind of final verification step we could add here?

modules/restore-determine-state-etcd-member.adoc
```diff
@@ -0,0 +1,78 @@
+// Module included in the following assemblies:
+//
+// * backup_and_restore/replacing-unhealthy-etcd-member.adoc
+
+[id="restore-determine-state-etcd-member_{context}"]
+= Determining the state of the unhealthy etcd member
+
+The steps to replace an unhealthy etcd member depend on which of the following states your etcd member is in:
+
+* The node is not running.
```
alaypatel07, line 10: When users run `oc get nodes`, they will see the conditions in `Ready/NotReady` status, when they run `oc get machine`, they will see `Running/Failed` etc amongst other things hence for the docs I suggest:

if the use of the term Node, the corresponding metric is Ready/NotReady
if the use of the term machine, the corresponding metric is Running/Not Running

when users run oc get node, they would see Ready states

backup_and_restore/replacing-unhealthy-etcd-member.adoc
bergerhoffer: @alaypatel07 It was meant to be the first procedure for when the node is stopped. Should it be "whether the node is stopped" instead of "whether the etcd member is stopped"? Or is there something else incorrect about this wording?

modules/restore-determine-state-etcd-member.adoc
alaypatel07: actually there is one:

If the admin knows that the underlying node is NotReady or machine is not Running and the admin expects the same machine/node to be `Ready/Running` soon, they do not need to perform any steps, cluster etcd operator will automatically sync when the unhealthy node/machine goes into expected state

bergerhoffer: Thanks, I can work in a mention of this.

modules/restore-determine-state-etcd-member.adoc
bergerhoffer: Updated terminology across the doc.

backup_and_restore/replacing-unhealthy-etcd-member.adoc
bergerhoffer: Updated per our slack conversation.

modules/restore-determine-state-etcd-member.adoc
bergerhoffer: @alaypatel07 I added a note here on this: https://github.com/openshift/openshift-docs/pull/20890/files#diff-1fca3c59dde22eaa3ca910449ad421d0R17

Can you see if the wording is okay?

bergerhoffer:
@geliu2016 We still have a few items to wrap up (marked as TODO in the source), but can you take a look at running through these procedures? Thanks!

https://osdocs-893-single-member--ocpdocs.netlify.app/openshift-enterprise/latest/backup_and_restore/replacing-unhealthy-etcd-member.html

modules/restore-replace-stopped-etcd-member.adoc
alaypatel07: @bergerhoffer yes, this is expected as the VM is being deleted under the hood. 

> Should we tell users to wait until the prompt is returned before proceeding to the next step?
yes, I think not only we should tell users to wait for the prompt to return, but I know some anxious ones like me will just ctrl+C out of it. So we should ask users to wait until  

`oc get machine ci-ln-mpprd6t-d5d6b-n94jk-master-0 -n openshift-machine-api` returns "No resources fouond"

make be even automate that with:

`$ until $(oc get machine ci-ln-mpprd6t-d5d6b-n94jk-master-0 -n openshift-machine-api | grep "No resources found"); do sleep 1; echo "waiting for the machine to be deleted"; done` 

as the next step after deletion.

modules/restore-replace-crashlooping-etcd-member.adoc
alaypatel07: @bergerhoffer I think the user can stay oblivious to the fact that this happens during scale-up, so I am not able to wrap my head around why we need to mention about the similarity. Maybe something like this:

> "When the etcd cluster Operator performs a redeployment, it will make sure all the master nodes will have a functioning etcd pod"

alaypatel07 (commented):
apologies for the delay, added responses inline

modules/restore-replace-stopped-etcd-member.adoc
alaypatel07: note, it will take some time for the machine to be created. We can add a step to watch on the etcd resource to include the third member with something like:

```
$ until $(oc get etcd  -o=jsonpath='{range .items[0].status.conditions[?(@.type=="EtcdMembersAvailable")]}{.message}{"\n"} | grep <third-master-node-name>) ; do sleep 1; echo "waiting for the third etcd"; done
```

modules/restore-replace-crashlooping-etcd-member.adoc
bergerhoffer: Thanks, added!

modules/restore-replace-stopped-etcd-member.adoc
bergerhoffer: Thanks, I'll add and will test it out tomorrow.

modules/restore-replace-stopped-etcd-member.adoc
bergerhoffer: @alaypatel07 To confirm, this isn't the same node name as it originally was as the unhealthy member, correct? So we should probably add a quick step in there to get the new node name. Thoughts on the easiest way of telling which node is the new one?

modules/restore-replace-stopped-etcd-member.adoc
bergerhoffer: @alaypatel07 I tried this just now with a machine that doesn't exist, and I get `Error from server (NotFound): machines.machine.openshift.io "ci-ln-mpprd6t-d5d6b-n94jk-master-0" not found`

I haven't tested with an actual machine I've deleted, but if the message is the same, this until command doesn't work.

jboxman: I don't know if this will help, but you can use the `--wait` option for `oc`:

```
$ oc delete pod sctpserver --wait
pod "sctpserver" deleted
... prompt waits until success ...
$ 
```

Perhaps this works for the Machine resource as well?

geliu2016:
LGTM, thanks

geliu2016:
LGTM, thx

hexfusion (commented):
This look good but I want to stress the fact that a user that is coming to these docs might be under duress. While verbose 1 liners are nice to review output they are prone to copy-paste errors. I think we should show simple commands to verify success and also the more robust 1 liner for validation if they choose to do so.

Also we should not reference open PRs in user-facing docs I have never seen this before. If something is not the way it should be we show workaround until it is fixed. At that time an update to the docs should also be done.

@bergerhoffer input?

modules/restore-replace-crashlooping-etcd-member.adoc
```diff
@@ -0,0 +1,135 @@
...
++
+In a terminal that has access to the cluster as a `cluster-admin` user, run the following command:
++
+----
+$ oc get etcd  -o=jsonpath='{range .items[0].status.conditions[?(@.type=="EtcdMembersAvailable")]}{.message}{"\n"}'
+ip-10-0-131-183.ec2.internal,ip-10-0-164-97.ec2.internal,ip-10-0-154-204.ec2.internal members are available,  have not started,  are unhealthy,  are unknown
+----
+
+////
+Once https://github.com/openshift/cluster-etcd-operator/pull/307 merges for 4.4, use this output:
```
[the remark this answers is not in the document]
bergerhoffer: @hexfusion This is commented out, it's not visible when you view the doc. Did you think that it was rendering, or are you saying that I shouldn't even have this commented out in my source in the repo?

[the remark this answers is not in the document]
hexfusion: ok cool that was my concern that it was user-facing thanks for the update.

modules/restore-replace-stopped-etcd-member.adoc
```diff
@@ -0,0 +1,147 @@
...
+sh-4.2# etcdctl member remove 6fc1e7c9db35841d
+Member 6fc1e7c9db35841d removed from cluster baa565c8919b060e
+----
+
+.. Verify that the member was removed:
++
+----
+sh-4.2# etcdctl endpoint health --cluster
+https://10.0.154.204:2379 is healthy: successfully committed proposal: took = 16.698331ms
+https://10.0.164.97:2379 is healthy: successfully committed proposal: took = 16.621645ms
```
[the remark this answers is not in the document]
bergerhoffer: Agree, switching this step to use `etcdctl member list -w table`.

modules/restore-replace-stopped-etcd-member.adoc
```diff
@@ -0,0 +1,147 @@
...
++
+----
+$ oc delete machine clustername-8qw5l-master-0 -n openshift-machine-api
+----
+
+.. Wait until the machine is deleted before proceeding to the next step:
++
+----
+$ until $(oc get machine clustername-8qw5l-master-0 -n openshift-machine-api | grep "No resources found"); do sleep 1; echo "Waiting for the machine to be deleted"; done
+----
```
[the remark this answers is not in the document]
bergerhoffer: @hexfusion so this will watch forever, even once the machine has been deleted?

Could we just run that command w/o watching, (`oc get machines -n openshift-machine-api`) and have them proceed with adding the new machine once the old machine is gone from the list?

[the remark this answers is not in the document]
hexfusion: absolutely fine with me, the point is simple commands.

bergerhoffer:
@geliu2016 We made a few command changes to simplify them, and a few other minor updates. Can you please take another look to make sure that the procedures still work as expected?

One thing I would definitely appreciate you taking a look at are my example YAML snippets in step 2 where we're editing the machine's file. I want to make sure that there isn't anything that shouldn't be shown or if any of the unique IDs need to be changed. Thanks!

Preview: https://osdocs-893-single-member--ocpdocs.netlify.app/openshift-enterprise/latest/backup_and_restore/replacing-unhealthy-etcd-member.html

geliu2016:
LGTM, and I have run step2 for many times last week ago, and there is not problem even if there is a little changes compare to it before. thanks

backup_and_restore/replacing-unhealthy-etcd-member.adoc
```diff
@@ -0,0 +1,37 @@
+[id="replacing-unhealthy-etcd-member"]
+= Replacing an unhealthy etcd member
+include::modules/common-attributes.adoc[]
+:context: replacing-unhealthy-etcd-member
+
+toc::[]
+
+This document describes the process to replace a single unhealthy etcd member.
+
+The process to replace the unhealthy member depends on whether the etcd member is unhealthy because the machine is not running or the node is not ready, or whether it is unhealthy because the etcd Pod is crashlooping.
```
bobfuru, line 10: This sentence is long; can you break it up or pare it down? Maybe something like:
```
The process to replace the unhealthy member depends on whether:
* The machine is not running or the node is not ready; or,
* The etcd Pod is crashlooping.
```
Or maybe:
`This process depends on whether the etcd member is unhealthy because the machine is not running or the node is not ready, or because the etcd Pod is crashlooping.`

modules/restore-determine-state-etcd-member.adoc
```diff
@@ -0,0 +1,93 @@
...
++
+----
+$ oc get nodes -l node-role.kubernetes.io/master | grep "NotReady"
+ip-10-0-131-183.ec2.internal   NotReady   master   122m   v1.17.1 <1>
+----
+<1> If the node is listed as `NotReady`, then the *node is not ready*.
+
++
+// TODO: xref
+If the *node is not ready*, then follow the _Replacing an unhealthy etcd member whose machine is not running or whose node is not ready_ procedure.
```
bobfuru, line 64: Same xref comment as above.

modules/restore-determine-state-etcd-member.adoc
```diff
@@ -0,0 +1,93 @@
...
+$ oc get pods -n openshift-etcd | grep etcd
+etcd-ip-10-0-131-183.ec2.internal                2/3     Error       7          6h9m <1>
+etcd-ip-10-0-164-97.ec2.internal                 3/3     Running     0          6h6m
+etcd-ip-10-0-154-204.ec2.internal                3/3     Running     0          6h6m
+----
+<1> Since this status of this Pod is `Error`, then the *etcd Pod is crashlooping*.
+
++
+// TODO: xref
+If the *etcd Pod is crashlooping*, then follow the _Replacing an unhealthy etcd member whose etcd Pod is crashlooping_ procedure.
```
bobfuru, line 93: Same xref comment as above.

modules/restore-identify-unhealthy-etcd-member.adoc
```diff
@@ -0,0 +1,38 @@
+// Module included in the following assemblies:
+//
+// * backup_and_restore/replacing-unhealthy-etcd-member.adoc
+
+[id="restore-identify-unhealthy-etcd-member_{context}"]
+= Identifying an unhealthy etcd member
+
+You can identify whether your cluster has an unhealthy etcd member.
```
bobfuru, line 8: s/whether/if/

modules/restore-replace-crashlooping-etcd-member.adoc
```diff
@@ -0,0 +1,140 @@
...
+.Prerequisites
+
+* You have identified the unhealthy etcd member.
+* You have verified that the etcd Pod is crashlooping.
+* You have access to the cluster as a user with the `cluster-admin` role.
+* You have taken an etcd backup.
++
+[IMPORTANT]
+====
+It is important to take an etcd backup before performing this procedure, so that your cluster can be restored if you encounter any issues.
```
bobfuru, line 19: s/procedure, so that/procedure so that/

modules/restore-replace-crashlooping-etcd-member.adoc
```diff
@@ -0,0 +1,140 @@
...
++
+----
+sh-4.4# mv /var/lib/etcd/ /tmp
+----
++
+You can now exit the node shell.
+
+. Remove the unhealthy member.
+
+.. Choose a Pod _not_ on the affected node.
```
bobfuru, line 57: s/Pod _not_/Pod  that is _not_/

modules/restore-replace-crashlooping-etcd-member.adoc
```diff
@@ -0,0 +1,140 @@
...
+----
+<1> Replace this with the name of the unhealthy node.
+
+.. Change your root directory to the host:
++
+----
+sh-4.2# chroot /host
+----
+
+.. Move the existing etcd Pod file out of the Kubelet manifest directory:
```
bobfuru, line 41: s/Kubelet/kubelet/
I've seen it both uppercase and lowercase in our docs, but k8s docs uses lowercase so that is my suggestion.

bobfuru (commented):
Some minor suggestions and a question about xrefs as anchor links, and then LGTM!

modules/restore-determine-state-etcd-member.adoc
```diff
@@ -0,0 +1,93 @@
...
++
+----
+$ oc get machines -A -ojsonpath='{range .items[*]}{@.status.nodeRef.name}{"\t"}{@.status.providerStatus.instanceState}{"\n"}' | grep -v running
+ip-10-0-131-183.ec2.internal  stopped <1>
+----
+<1> If this output lists the machine as anything other than `running`, then the *machine is not running*.
+
++
+// TODO: xref
+If the *machine is not running*, then follow the _Replacing an unhealthy etcd member whose machine is not running or whose node is not ready_ procedure.
```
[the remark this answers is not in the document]
bergerhoffer: @bobfuru No we can't yet. I just do this so that I can easily find places to add xrefs once we're allowed to again.

modules/restore-replace-crashlooping-etcd-member.adoc
bergerhoffer: You're right, I did just look that up the other day as well and agree that it's lowercase, but I had copied this from another procedure and missed it!

backup_and_restore/replacing-unhealthy-etcd-member.adoc
```diff
@@ -0,0 +1,37 @@
...
+
+// Identifying an unhealthy etcd member
+include::modules/restore-identify-unhealthy-etcd-member.adoc[leveloffset=+1]
+
+// Determining the state of the unhealthy etcd member
+include::modules/restore-determine-state-etcd-member.adoc[leveloffset=+1]
+
+== Replacing the unhealthy etcd member
+
+Depending on the state of your unhealthy etcd member, either follow the steps to xref:../backup_and_restore/replacing-unhealthy-etcd-member.adoc#restore-replace-stopped-etcd-member_replacing-unhealthy-etcd-member[replace an unhealthy etcd member whose machine is not running or whose node is not ready] or to xref:../backup_and_restore/replacing-unhealthy-etcd-member.adoc#restore-replace-crashlooping-etcd-member_replacing-unhealthy-etcd-member[replace an unhealthy etcd member whose etcd Pod is crashlooping].
```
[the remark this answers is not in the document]
bergerhoffer: Thanks yeah I know these weren't ideal sentences because they're a bit complicated. I'll try to break this up a little better.

PR #20900, opened 2020-04-03
The default values for OpenShift metrics(hawkular) on installation should be documented

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1730179


mburke5678:
@juzhao Please take a look

PR #20913, opened 2020-04-04
Clarify phrasing for choosing a region

bparry02:
Clarify phrasing for criteria to pick an AWS region.

`You can install a single cluster in many regions` sounds like creating a stretched cluster across multiple regions, which is not the intent here.

Applies to 4.1, 4.2, and 4.3 docs.

### DIFF
--- a/modules/installation-aws-limits.adoc
+++ b/modules/installation-aws-limits.adoc
@@ -52,10 +52,10 @@ and each NAT gateway requires a separate
 link:https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/elastic-ip-addresses-eip.html[elastic IP].
 Review the
 link:https://aws.amazon.com/about-aws/global-infrastructure/[AWS region map] to
-determine how many availability zones are in each region. You can install a
-single cluster in many regions without increasing your EIP limit, but to take
-advantage of the default high availability, install the cluster in a region with
-at least three availability zones.
+determine how many availability zones are in each region. To take advantage of
+the default high availability, install the cluster in a region with at least
+three availability zones. To install a cluster in a region with more than five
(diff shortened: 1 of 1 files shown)


modules/installation-aws-limits.adoc
```diff
@@ -53,7 +53,7 @@ link:https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/elastic-ip-addresses-ei
 Review the
 link:https://aws.amazon.com/about-aws/global-infrastructure/[AWS region map] to
 determine how many availability zones are in each region. You can install a
-single cluster in many regions without increasing your EIP limit, but to take
+single cluster in most regions without increasing your EIP limit, but to take
```
kalexand-rh, line 56: @bparry02, I'd rather say "many of these" than "most." (I don't track data on what percentage of regions you can install without increasing your EIP limit, so we might be introducing a different inaccuracy with "most.")

bparry02:
@kalexand-rh Fair point. I have force pushed the change you suggested.

I think where I'm struggling with this Description is that it's dancing around the requirement a bit. I wonder if it's better to just rewrite the sentence to something like:

> Review the AWS region map to determine how many availability zones are in each region. To take advantage of the default high availability, install the cluster in a region with at least three availability zones. To install a cluster in a region with more than 5 Availability Zones, an EIP limit increase is required.

kalexand-rh:
@bparry02, that's fair. Maybe:

> Review the AWS region map to determine how many availability zones are in each region. To take advantage of the default high availability, install the cluster in a region with at least three availability zones. To install a cluster in a region with more than five availability zones, you must increase the EIP limit.

bparry02:
@kalexand-rh yes that's better. Updated the branch. Thanks!

PR #20917, opened 2020-04-06
CNV: VMware VM import troubleshooting

apinnick:
Added section for troubleshooting CNV VMware VM import
Known issue for CNV 2.3 only (conditionally tagged): migration progress status not displated

Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1829828

### DIFF
--- a/modules/cnv-troubleshooting-vmware-vm-import.adoc
+++ b/modules/cnv-troubleshooting-vmware-vm-import.adoc
@@ -1,0 +1,41 @@
+// Module included in the following assemblies:
+// * cnv/cnv_virtual_machines/cnv_importing_vms/cnv-importing-vmware-vm.adoc
+[id='troubleshooting-vmware-vm-import_{context}']
+= Troubleshooting a VMware virtual machine import
+
+If an imported virtual machine's status is `Import error: (VMware)`, you can check the Conversion Pod log for errors:
+
+. Obtain the Conversion Pod name:
++
+----
+$ oc get pods -n <project> | grep v2v <1>
+kubevirt-v2v-conversion-f66f7d-zqkz7            1/1     Running     0          4h49m
+----
+<1> Specify the project of your imported virtual machine.
+
+. Obtain the Conversion Pod log:
++
+----
+$ oc logs kubevirt-v2v-conversion-f66f7d-zqkz7 -f -n <project>
+----
+
+== Error messages
+
+* If an imported virtual machine event displays the error message, `Readiness probe failed`, the following error messsage appears in the Conversion Pod log:
++
+----
+INFO - have error: ('virt-v2v error: internal error: invalid argument: libvirt domain ‘v2v_migration_vm_1’ is running or paused. It must be shut down in order to perform virt-v2v conversion',)"
+----
++
+You must ensure that the virtual machine is shut down before importing it.
+
+ifeval::["{CNVVersion}" == "2.3"]
+== Known issues
+
+* Migration progress status is not displayed in the Cluster Application Migration tool web console. link:https://bugzilla.redhat.com/show_bug.cgi?id=1830780[(*1830780*)]
++
+[NOTE]
+====
+This issue does not affect the migration itself.
+====
+endif::[]
--- a/cnv/cnv_virtual_machines/cnv_importing_vms/cnv-importing-vmware-vm.adoc
+++ b/cnv/cnv_virtual_machines/cnv_importing_vms/cnv-importing-vmware-vm.adoc
@@ -4,7 +4,7 @@ include::modules/cnv-document-attributes.adoc[]
 :context: cnv-importing-vmware-vm
 toc::[]
 
-You can import a single VMware virtual machine or template using the virtual machine wizard.
+You can import a single VMware virtual machine or template into your {product-title} cluster.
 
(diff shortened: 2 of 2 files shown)


modules/cnv-troubleshooting-vmware-vm-import.adoc
```diff
@@ -0,0 +1,39 @@
...
+----
+<1> Specify the project of your imported virtual machine.
+
+. Obtain the Conversion Pod log:
++
+----
+$ oc logs kubevirt-v2v-conversion-f66f7d-zqkz7 -f -n <project>
+----
+
+= Error messages
```
bergerhoffer, line 22: There shouldn't be another heading level 1 (=) in the same module. I assume that this should be a level 2, but if this really should be at level 1, then it should be in a separate module file.

modules/cnv-troubleshooting-vmware-vm-import.adoc
```diff
@@ -0,0 +1,39 @@
...
+
+* Virtual machine event error message: `Readiness probe failed`
++
+Log message: `INFO - have error: ('virt-v2v error: internal error: invalid argument: libvirt domain ‘v2v_migration_vm_1’ is running or paused. It must be shut down in order to perform virt-v2v conversion',)"`
```
bergerhoffer, line 26: This should probably go into a code block. When looking at the preview, it cuts off at ‘v2v and I can't scroll to the right at all to see the rest.

modules/cnv-troubleshooting-vmware-vm-import.adoc
```diff
@@ -0,0 +1,39 @@
...
++
+Resolution: Shut down the virtual machine before importing it.
+
+ifeval::[{CNVVersion} == "2.3"]
+= Known issues
```
bergerhoffer, line 31: Same comment here that there shouldn't be another level 1 heading in the same module file. But I'm not sure what the appropriate level heading for this one is.

modules/cnv-troubleshooting-vmware-vm-import.adoc
```diff
@@ -0,0 +1,39 @@
...
+
+* Migration progress status is not displayed in the CAM web console. link:https://bugzilla.redhat.com/show_bug.cgi?id=1830780[(*1830780*)]
```
bergerhoffer, line 33: This is the first I see the CAM tool mentioned in this assembly, is this correct? And if so, suggest spelling out what CAM is.

modules/cnv-troubleshooting-vmware-vm-import.adoc
```diff
@@ -0,0 +1,39 @@
...
+
+= Error messages
+
+* Virtual machine event error message: `Readiness probe failed`
++
+Log message: `INFO - have error: ('virt-v2v error: internal error: invalid argument: libvirt domain ‘v2v_migration_vm_1’ is running or paused. It must be shut down in order to perform virt-v2v conversion',)"`
++
+Resolution: Shut down the virtual machine before importing it.
+
+ifeval::[{CNVVersion} == "2.3"]
```
bergerhoffer, line 30: Should this be showing for 4.4/4.5? Because I can see that the CNVVersion is set to 2.3 in the cnv-document-attributes.adoc file in master, but this section doesn't display in the preview build.

modules/cnv-troubleshooting-vmware-vm-import.adoc
```diff
@@ -0,0 +1,39 @@
+// Module included in the following assemblies:
+// * cnv/cnv_virtual_machines/cnv_importing_vms/cnv-importing-vmware-vm.adoc
+[id='troubleshooting-vmware-vm-import-{context}']
```
bergerhoffer, line 3: it should be an underscore, not a dash before _{context}

bergerhoffer (commented):
Just a few things!

modules/cnv-troubleshooting-vmware-vm-import.adoc
apinnick: Strange. This worked when I created the preview build. But if the topic_map is commented out, I imagine it won't appear at all.

apinnick: > Should this be showing for 4.4/4.5? Because I can see that the CNVVersion is set to 2.3 in the cnv-document-attributes.adoc file in master, but this section doesn't display in the preview build.

That depends on whether CNV 2.3 is going to be visible for 4.4/4.5 or not. As long as CNV version is 2.3, that section should appear.

bergerhoffer: So the known issues section is still not showing for me when I do a build. I checked other places where we do an ifeval, and they all put quotes around the variable.

```
ifeval::["{CNVVersion}" == "2.3"] 
```

I tried this and then this section shows for me.

bergerhoffer:
@apinnick LGTM other than still the thread on the known issues section not showing as is.

bergerhoffer:
The cnv/cnv_virtual_machines/cnv_importing_vms/cnv-importing-vmware-vm.adoc file doesn't exist in 4.2. After discussing with Avital, this doesn't need to go to 4.2 after all.

PR #20934, opened 2020-04-06
Updates to docs for OCS 4.2

Erin-Donnelly:
sections "About dynamic provisioning" and "Access modes"

### DIFF
--- a/modules/dynamic-provisioning-about.adoc
+++ b/modules/dynamic-provisioning-about.adoc
@@ -5,21 +5,27 @@
 [id="about_{context}"]
 = About dynamic provisioning
 
-The StorageClass resource object describes and classifies storage that can 
+The StorageClass resource object describes and classifies storage that can
 be requested, as well as provides a means for passing parameters for
-dynamically provisioned storage on demand. StorageClass objects can also 
-serve as a management mechanism for controlling different levels of 
+dynamically provisioned storage on demand. StorageClass objects can also
+serve as a management mechanism for controlling different levels of
 storage and access to the storage. Cluster Administrators (`cluster-admin`)
- or Storage Administrators (`storage-admin`) define and create the 
-StorageClass objects that users can request without needing any intimate 
+ or Storage Administrators (`storage-admin`) define and create the
+StorageClass objects that users can request without needing any intimate
 knowledge about the underlying storage volume sources.
 
-The {product-title} persistent volume framework enables this functionality 
-and allows administrators to provision a cluster with persistent storage. 
-The framework also gives users a way to request those resources without 
+The {product-title} persistent volume framework enables this functionality
+and allows administrators to provision a cluster with persistent storage.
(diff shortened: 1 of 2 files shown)


Erin-Donnelly:
Note: these changes are for OCP 4.2 docs

modules/dynamic-provisioning-about.adoc
```diff
@@ -5,21 +5,27 @@
...
 Many storage types are available for use as persistent volumes in
 {product-title}. While all of them can be statically provisioned by an
 administrator, some types of storage are created dynamically using the
 built-in provider and plug-in APIs.
+
+When using Red Hat OpenShift Container Storage, the dynamic StorageClasses are
+created when Red Hat OpenShift Container Storage 4.2 is deployed from the
+Operator Hub as described in the
+link:https://access.redhat.com/documentation/en-us/red_hat_openshift_container_storage/4.2/html-single/deploying_openshift_container_storage/index#verify_that_the_storage_classes_are_created_and_listed[Verify that the storage classes are created and listed]
+section.
```
bobfuru, line 31: I don't think we should be adding OCS-specific documentation to this section for OCS as then we would need to include content for all other supported plug-ins. OCS content placement should make sense in the context of our docs, which it doesn't here.

My understanding from Ken is that the OCS content and links in OCP would be maintained on this page: https://docs.openshift.com/container-platform/4.2/storage/persistent_storage/persistent-storage-ocs.html

modules/storage-persistent-storage-pv.adoc
```diff
@@ -141,6 +141,16 @@ the Pods that use these volumes are deleted.
...
+Red Hat OpenShift Container Storage 4.2 provides the following backends for these
+access modes:
+
+* ReadWriteOnce: ceph-rbd (block mode)
+* ReadWriteMany: cephfs (file mode)
+
+See
+link:https://access.redhat.com/documentation/en-us/red_hat_openshift_container_storage/4.2/html-single/managing_openshift_container_storage/index#available-plug-ins_rhocs[Available dynamic provisioning plug-ins] 
+for more information.
+
```
bobfuru, line 153: We already have OCS documented in our "Access modes" table beginning in 4.2 docs: https://docs.openshift.com/container-platform/4.2/storage/understanding-persistent-storage.html#pv-access-modes_understanding-persistent-storage

If helpful, we can add `ceph-rbd (block mode)` in the **RWO** column and `cephfs (file mode)` in the **RWX** column.

It looks like your feature branch (`ocs4.2-updates`) is checked out from OCP `master` so I'm not sure why the table in this file does not already show OCS in the "Volume Plug-in" column. (We added OCS as a supported plug-in starting in OCP 4.2.)

Also, I don't think we should be linking documentation from this section for OCS as then we would potentially need to include links for all other plug-ins. OCS content placement should make sense in the context of our docs, which linking doesn't here. (Similar to my comment above.)

PR #20936, opened 2020-04-06
Update command to run ovs-* in OVS pods

jboxman:
- https://bugzilla.redhat.com/show_bug.cgi?id=1651778

Preview: http://file.bos.redhat.com/jboxman/20200406/bz-1651778/admin_guide/sdn_troubleshooting.html#is-the-open-vswitch-ovs-configured-correctly

### DIFF
--- a/admin_guide/sdn_troubleshooting.adoc
+++ b/admin_guide/sdn_troubleshooting.adoc
@@ -398,21 +398,30 @@ default via 192.168.122.1 dev ens3
 
 You should see the 10.128.x.x lines (assuming you have your pod network set to the default range in your configuration). If you do not, check the {product-title} logs (see the <<reading-the-logs>> section)
 
-=== Is the Open vSwitch configured correctly?
+=== Is the Open vSwitch (OVS) configured correctly?
 
-Check the Open vSwitch bridges on both sides:
+You must run the `ovs-vsctl` and `ovs-ofctl` commands on one of the OVS pods.
+
+To list the OVS pods, enter the following command:
 
 ----
-# ovs-vsctl list-br
+$ oc get pod -n openshift-sdn -l app=ovs
+----
+
+Check the Open vSwitch bridges on both sides.
+Replace `<ovs_pod_name>` with the name of one of the OVS pods.
+
+----
+$ oc exec -n openshift-sdn <ovs_pod_name> -- ovs-vsctl list-br
 br0
 ----
 
-This should be `br0`.
+The previous command should return `br0`.
 
-You can list all of the ports that ovs knows about:
+You can list all of the ports that OVS knows about:
 
 ----
-# ovs-ofctl -O OpenFlow13 dump-ports-desc br0
+$ oc exec -n openshift-sdn <ovs_pod_name> -- ovs-ofctl -O OpenFlow13 dump-ports-desc br0
 OFPST_PORT_DESC reply (OF1.3) (xid=0x2):
  1(vxlan0): addr:9e:f1:7d:4d:19:4f
      config:     0
@@ -439,7 +448,7 @@ be listed as ports.
 Next, list the flows that are configured on that bridge:
 
 ----
-# ovs-ofctl -O OpenFlow13 dump-flows br0
+$ oc exec -n openshift-sdn <ovs_pod_name> --  ovs-ofctl -O OpenFlow13 dump-flows br0
 ----
 
 The results will vary slightly depending on whether you are using the


admin_guide/sdn_troubleshooting.adoc
```diff
@@ -398,7 +398,14 @@ default via 192.168.122.1 dev ens3
...
 
-=== Is the Open vSwitch configured correctly?
+=== Is the Open vSwitch (OVS) configured correctly?
+
+You must install the `openvswitch` package on the host to run the `ovs-vsctl` and `ovs-ofctl` commands.
+To install the `openvswitch` package, enter the following command:
+
+----
+# yum install openvswitch
+----
```
zhaozhanqi, line 408: hmm, for 3.11 maybe host do not have `yum` command.  So I suggest run this ovs command in the container: eg
`oc exec -n openshift-sdn ovs-45dzz -- ovs-vsctl list-br`

jboxman: @zhaozhanqi, how do I find the names of the SDN pods? Thanks!

zhaozhanqi:
```
oc get pod -n openshift-sdn -l app=ovs
NAME        READY     STATUS    RESTARTS   AGE
ovs-45dzz   1/1       Running   0          17h
ovs-8njfh   1/1       Running   0          17h
ovs-zw5tz   1/1       Running   0          17h
oc get pod -n openshift-sdn -l app=sdn
NAME        READY     STATUS    RESTARTS   AGE
sdn-jssrh   1/1       Running   0          17h
sdn-mqbhf   1/1       Running   0          17h
sdn-qzvks   1/1       Running   0          17h
```

jboxman: Thanks!

jboxman:
@zhaozhanqi, I've updated the commands based on your feedback.

admin_guide/sdn_troubleshooting.adoc
```diff
@@ -398,21 +398,30 @@ default via 192.168.122.1 dev ens3
...
-Check the Open vSwitch bridges on both sides:
+You must run the `ovs-vsctl` and `ovs-ofctl` commands on one of the OVS pods.
+
+To list the OVS pods, enter the following command:
+
+----
+$ oc get pod -n openshift-sdn -l app=ovs
+----
+
+Check the Open vSwitch bridges on both sides.
```
kalexand-rh, line 411: Is this going to be obvious to the user?

admin_guide/sdn_troubleshooting.adoc
```diff
@@ -398,21 +398,30 @@ default via 192.168.122.1 dev ens3
...
+Replace `<ovs_pod_name>` with the name of one of the OVS pods.
 
 ----
-# ovs-vsctl list-br
+$ oc exec -n openshift-sdn <ovs_pod_name> -- ovs-vsctl list-br
 br0
 ----
 
-This should be `br0`.
+The previous command should return `br0`.
```
kalexand-rh, line 419: What if it's not? Is this not clear enough from the command output?

kalexand-rh (commented):
The free-form nature of the 3.x docs makes this section scan oddly, but the change itself LGTM.

jboxman:
Hi @juanluisvaladas, I'm updating this section of the 3.11 SDN debugging docs; Do you have any insight into these commands and whether there is any other information that might be helpful when probing if OVS is correctly configured?

Thanks!

admin_guide/sdn_troubleshooting.adoc
juanluisvaladas: That should never happen, it would need some more investigation to determine why. I don't think we can't give a generic answer to that because it wouldn't be a well-known issue.

If there are more switches present something or someone is creating them, and if br0 isn't present there is a problem with the SDN. Either way it would need further investigation.

juanluisvaladas:
Hi Jason,
I don't have any further suggestions. This looks more complete than I would expect, unless we want to document how the flows structure, which I'm not sure if that is the case.

The changes look good but there is one additional change necessary, In the docs I can read:

> You must install the openvswitch package on the host to run the ovs-vsctl and ovs-ofctl commands. To install the openvswitch package, enter the following command:
> 
> # yum install openvswitch
> 

This should be removed. You're doing oc exec, so it's not necessary at all.
Also it's important to say that if you install OVS in the host, most of the times you'll be lucky enough and running the host's ovs-ofctl or ovs-vsctl will work without problems with the vswitchd running in the ovs container, however, sometimes they'll be incompatible.

Finally, I don't know if you're still updating 3.10's docs, but all these commands are valid for 3.10 and 4.x as well. Not quite sure how this fits with our 4.x doc structure, but there shouldn't be any difference between 3.10 or 3.11 in this regard .

juanluisvaladas (changes requested):
The yum install openvswitch should be removed

jboxman:
@juanluisvaladas, oops, didn't update the previews. The latest version omits any mention of installing `openvswitch`.

We've deprecated the documentation for OCP <= `3.10`.

juanluisvaladas (approved):

PR #20938, opened 2020-04-06
GarbageCollector section is not clear

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1704598

### DIFF
--- a/admin_guide/garbage_collection.adoc
+++ b/admin_guide/garbage_collection.adoc
@@ -14,10 +14,10 @@ toc::[]
 
 The {product-title} node performs two types of garbage collection:
 
-* xref:container-garbage-collection[Container garbage collection] Removes
-terminated containers.
+* xref:container-garbage-collection[Container garbage collection]: Removes
+terminated containers. Enabled by default.
 * xref:image-garbage-collection[Image garbage collection]: Removes images not
-referenced by any running pods.
+referenced by any running pods. Not enabled by default.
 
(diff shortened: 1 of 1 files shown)


sunilcio:
LGTM

sunilcio:
@mburke5678 I just checked your question in bz. Garbage collection for containers is enabled by default. For image garbage collection, user need to manually enable it. So we need to modify the statement to reflect that in image garbage collection.

admin_guide/garbage_collection.adoc
```diff
@@ -15,9 +15,9 @@ toc::[]
 The {product-title} node performs two types of garbage collection:
 
 * xref:container-garbage-collection[Container garbage collection] Removes
```
kalexand-rh, line 17: I'd add a : after the xref.

kalexand-rh (commented):
I'd consider moving each of the notes before their respective code blocks. It's not in the section that you changed, but I'm a little confused how we have default parameters for the image garbage collection if it's not enabled by default.

mburke5678:
@kalexand-rh Thanks for the tips. I moved the notes and they do look better. 

@sunilcio I removed the references to the default values for the image garbage collection parameters based on Kathryn's comment above. It seems non-intuitive to say a parameter is not default but has default values. (It does seems that the parameters do have defaults.). Let me know if you think I should put them back.

sunilcio:
I agree, looks better now. LGTM

PR #20968, opened 2020-04-07
CNV2928 - OVN support

aburdenthehand:
Added NEW release note for OVN support in CNV

### DIFF
--- a/cnv/cnv_release_notes/cnv-release-notes.adoc
+++ b/cnv/cnv_release_notes/cnv-release-notes.adoc
@@ -15,7 +15,10 @@ include::modules/technology-preview.adoc[leveloffset=+2]
 
 == New and changed features
 
-// CNV-2928/CNV-4166 OVN-Kubernetes network provider
+[id="cnv-2-3-networking"]
+=== Networking
+
+* You can now use {CNVProductName} with either the xref:../../networking/ovn_kubernetes_network_provider/about-ovn-kubernetes.adoc#learn-about-ovn-kubernetes[OVN-Kubernetes] or the xref:../../networking/openshift_sdn/about-openshift-sdn.adoc#about-openshift-sdn[OpenShiftSDN] network provider.
 
 // CNV-4091 nmstate
 
--- a/modules/cnv-what-you-can-do-with-cnv.adoc
+++ b/modules/cnv-what-you-can-do-with-cnv.adoc
@@ -20,3 +20,6 @@ custom resources to enable virtualization tasks. These tasks include:
 
 An enhanced web console provides a graphical portal to manage these virtualized
 resources alongside the {product-title} cluster containers and infrastructure.
+
+You can use {CNVProductName} with either the xref:../networking/ovn_kubernetes_network_provider/about-ovn-kubernetes.adoc#learn-about-ovn-kubernetes[OVN-Kubernetes] or the xref:../networking/openshift_sdn/about-openshift-sdn.adoc#about-openshift-sdn[OpenShiftSDN] network provider.
+
--- a/cnv/cnv-about-cnv.adoc
+++ b/cnv/cnv-about-cnv.adoc
@@ -10,5 +10,5 @@ include::modules/cnv-what-you-can-do-with-cnv.adoc[leveloffset=+1]
 
 == {CNVProductNameStart} support
 
-:FeatureName: {CNVProductName}
+:FeatureName: {CNVProductNameStart}
 include::modules/technology-preview.adoc[leveloffset=+2]


phoracek (approved):
/lgtm

phoracek (approved):
Very good!

cnv/cnv_release_notes/cnv-release-notes.adoc
```diff
@@ -15,7 +15,12 @@ include::modules/technology-preview.adoc[leveloffset=+2]
 
 == New and changed features
 
-// CNV-2928/CNV-4166 OVN-Kubernetes network provider
+[id="cnv-2-3-networking"]
+=== Networking
+
+* The Open Virtual Networking (OVN) Kubernetes network plug-in is a network virtualization platform developed by the Open vSwitch community.
```
dankenigsberg, line 21: I'm not sure we have to define ovn here. I think that the xref is enough.

But even with this, the PR is acceptable.

dankenigsberg (approved):

duyanyan:
Just a minor question: seems the Open Virtual Networking (OVN) Kubernetes is still a Technology Preview feature, do we need to mention that in the note? else LGTM

dankenigsberg:
I don't think we should mention the support state of other features because it may change soon and we'd become outdated. It is enough to add a link to OVNKubernetes definition.

aburdenthehand:
Thanks for the review folks. As suggested, I've simplified the Release Note.

modules/cnv-what-you-can-do-with-cnv.adoc
```diff
@@ -19,4 +19,7 @@ custom resources to enable virtualization tasks. These tasks include:
 * Live migrating virtual machines between nodes
 
 An enhanced web console provides a graphical portal to manage these virtualized
-resources alongside the {product-title} cluster containers and infrastructure.
\ No newline at end of file
+resources alongside the {product-title} cluster containers and infrastructure.
+
+You can use {CNVProductName} with either the xref:../networking/ovn_kubernetes_network_provider/about-ovn-kubernetes.adoc#learn-about-ovn-kubernetes[OVN-Kubernetes] or the xref:../networking/openshift_sdn/about-openshift-sdn.adoc#about-openshift-sdn[OpenShiftSDN] network provider.
```
vikram-redhat, line 24: @aburdenthehand no xrefs in modules. Can you remove this?

dankenigsberg: Pardon my ignorance, @vikram-redhat, but why aren't xref allowed? What would you suggest CNV to have instead? We can refer to the SDNs by name and describe where more info can be found about them, but nothing beats a clickable URL.

vikram-redhat: Hey @dankenigsberg this is, unfortunately, a long running battle within our (RH) internal tools and development of modular docs for the customer portal. 2 years ago a decision was taken (for technical and conceptual reasons) to exclude xrefs from within modules. 

xrefs within assemblies are ok. I would suggest @aburdenthehand moves this section to the enclosing assembly instead, if possible, otherwise this needs to be removed.

dankenigsberg: @vikram-redhat imho there should be a static analysis tool rejecting `xref` when they are proposed. The decision and its reasoning should be well-documented too. To me, it still sounds arbitrary.

vikram-redhat: @dankenigsberg it is well documented to not use it in our repo: https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#links-hyperlinks-and-cross-references, I remind all docs folks to not use it as many times as I can in our weekly meetings.

It has had lots of discussions going back to two years within the docs folks at Red Hat: https://github.com/redhat-documentation/modular-docs/issues/64 and decisions from the tooling team: https://github.com/redhat-documentation/modular-docs/issues/64#issuecomment-416423780 and you are welcome to wade in. Including the most recent confirmation that it is not to be used: https://github.com/redhat-documentation/modular-docs/issues/113#issuecomment-586044976.

Tooling team for Red Hat portal docs drives the decision behind this and have after extensive discussions decided to overturn this ban for a future version of portal docs.

A build process to reject xrefs in modules and fail it is in the works for OpenShift based docs.

aburdenthehand: Thanks for the catch. Corrected in #22419
No impact to the built doc:
https://cnv-ovn-naughty-xref--ocpdocs.netlify.app/openshift-enterprise/latest/virt/about-virt.html

Issue #20979, opened 2020-04-07
`odo init` in documentation makes it sound like it's a command


cdrage:
### Which section(s) is the issue in?

https://docs.openshift.com/online/starter/cli_reference/openshift_developer_cli/using_odo_in_a_restricted_environment/about-odo-in-a-restricted-environment.html (and others, the entire section)

### What needs fixing?

I was going through the documentation on how to use odo in an offline environment and found multiple references to `odo init` image (in that way).

To me (and maybe to others) it makes it sound as though `odo init` is an actual command. Perhaps we can remove the whole `odo init` image and just say: "odo init image" ? 

ping @boczkowska

cdrage:
![Screenshot from 2020-04-07 14-45-31](https://user-images.githubusercontent.com/6422176/78707967-5f4ad000-78df-11ea-998a-d1064f6bdb47.png)

bergerhoffer:
Closing as PR #20991 addressed this.

PR #20991, opened 2020-04-08
Removed unnecessary backticks.

yhontyk:
Removed backticks around "odo image". It is unnecessary and might confuse users: https://github.com/openshift/openshift-docs/issues/20979

Branches 4.2, 4.3 and 4.4

It doesn't require QE review since it's just a formatting fix.


kalexand-rh:
@boczkowska, our style guide does say to enclose specific names, packages, in backticks. See https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#quick-markup-reference

I don't think this change is necessary.

boczkowska:
@kalexand-rh "init image" is not a specific name in this case I think, just a category of an image. 

I searched how other doc teams use this term, not too many instances, unfortunately. :( RH docs, for example, capitalize it: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/building_running_and_managing_containers/using_red_hat_universal_base_images_standard_minimal_and_runtimes and use backticks only when referring to a specific image ("`ubi8-init` image" but "Init image"). Should I maybe do the same? 

@cdrage would appreciate your input here as both an SME and a user in this case.

cdrage:
To me, I 100% agree on removing the backticks. As a user it makes me think as though the `odo init` command exists, which it does not.

It makes it even more confusing because `odo init` was a *possible* command we were thinking of implementing, but we decided against it.

PR #20996, opened 2020-04-08
Remove dash for CSI features in TP Tracker

bobfuru:
Removing dashes in TP tracker table for CSI snapshots/cloning because, as explained above the table, "features marked as *-* indicate that the feature is removed from the release or deprecated" (which is not the case for CSI snapshots and CSI cloning).

### DIFF
--- a/release_notes/ocp-4-4-release-notes.adoc
+++ b/release_notes/ocp-4-4-release-notes.adoc
@@ -662,13 +662,13 @@ indicate that the feature is removed from the release or deprecated.
 |
 
 |CSI volume snapshots
-|-
-|-
+|
+|
 |TP
 
 |CSI volume cloning
-|-
-|-
+|
+|
 |TP
 
 |====


bobfuru:

bmcelvee:
LGTM!

codyhoag:
Hey @bobfuru; the dashes have been used in several releases for our RNs to indicate that the feature did not exist or was in its very beginning days (and not in TP). I don't think it indicates uncertainty. Blank spaces, for how I've implemented it, are where we're unsure of its status.

codyhoag:
With that said, looking back, there's quite a bit of inconsistency with its usage. So maybe we should reassess how we're using them :thinking:

bobfuru:
> With that said, looking back, there's quite a bit of inconsistency with its usage. So maybe we should reassess how we're using them 🤔

I think I was using them in the same way that you described until I saw the note above the table:
`Features marked as - indicate that the feature is removed from the release or deprecated.`
And because they hadn't been removed or deprecated, I removed the dashes. hmm

codyhoag:
Aha, that's a great point. OK, proceed with the removal :sweat_smile:. Those are applied inconsistently throughout then. I'll go back through today and update the others.

Thanks for spotting that!

PR #20998, opened 2020-04-08
BZ1739420 - Clarify selector-label volume binding

bobfuru:
[BZ1739420](https://bugzilla.redhat.com/show_bug.cgi?id=1739420)
Doc bug asks to include optional steps for how to reserve a PV for specific tasks, as described in [BZ1717746](https://bugzilla.redhat.com/show_bug.cgi?id=1717746).
Applies to `enterprise-3.11` only.

### DIFF
--- a/install_config/persistent_storage/selector_label_binding.adoc
+++ b/install_config/persistent_storage/selector_label_binding.adoc
@@ -21,14 +21,14 @@ storage] by identifiers defined by a cluster administrator.
 [[selector-label-volume-motivation]]
 == Motivation
 In cases of statically provisioned storage, developers seeking persistent
-storage are required to know a handful identifying attributes of a PV in order
+storage are required to know a handful of identifying attributes of a PV in order
 to deploy and bind a PVC. This creates several problematic situations. Regular
 users might have to contact a cluster administrator to either deploy the PVC or
 provide the PV values. PV attributes alone do not convey the intended use of the
 storage volumes, nor do they provide methods by which volumes can be grouped.
 
 Selector and label attributes can be used to abstract away PV details from the
-user while providing cluster administrators a way of identifying volumes by a
+user while providing cluster administrators with a way of identifying volumes by a
 descriptive and customizable tag. Through the selector-label method of binding,
 users are only required to know which labels are defined by the administrator.
 
@@ -53,7 +53,7 @@ xref:./index.adoc#install-config-persistent-storage-index[storage provider]
 [[selector-label-volume-define]]
 === Define the Persistent Volume and Claim
 
-. As the *cluser-admin* user, define the PV. For this example, we will
+. As the *cluster-admin* user, define the PV. For this example, we will
 be using a
 xref:./persistent_storage_glusterfs.adoc#install-config-persistent-storage-persistent-storage-glusterfs[GlusterFS]
 volume.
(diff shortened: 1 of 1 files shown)


install_config/persistent_storage/selector_label_binding.adoc
```diff
@@ -111,6 +111,58 @@ spec:
 labels of targeted PV.
 ====
 
+[[selector-label-volume-bind]]
+=== Optional: Bind the PVC to a specific PV when using a selector
```
bobfuru, line 115: Is this the right placement for this optional step (between Define and Deploy)? Also, not certain that "when using a selector" is necessary here?

install_config/persistent_storage/selector_label_binding.adoc
```diff
@@ -111,6 +111,58 @@ spec:
...
+
+As the *cluster-admin* user, you can bind the PVC to a specific PV only.
+
+To reserve a PV for specific tasks, you have two options: create a specific storage class, or pre-bind the PV to your PVC.
+
+. Request a specific storage class for the PV by specifying the storage class’s name.
++
+The following resource shows the required values that you use to configure a StorageClass. This example uses the AWS ElasticBlockStore (EBS) object definition.
+
```
bobfuru, line 124: The bug reporter requested a storage class spec for binding purposes, so I borrowed from this sample: https://docs.openshift.com/container-platform/4.3/storage/dynamic-provisioning.html#basic-storage-class-definition_dynamic-provisioning
Could use help if there is a better example we could use for this context.

install_config/persistent_storage/selector_label_binding.adoc
```diff
@@ -111,6 +111,58 @@ spec:
...
+metadata:
+  name: gp2
+provisioner: kubernetes.io/aws-ebs
+...
+----
+====
++
+[IMPORTANT]
+====
+If necessary in a multi-tenant environment, use a quota definition to reserve the storage class and PV(s) only to a specific namespace.
```
bobfuru, line 140: Should this be called out like this and will this statement be clear to a cloud admin?

bobfuru:
@jsafrane PTAL - this is based on a bug fix you documented in [BZ1717746](https://bugzilla.redhat.com/show_bug.cgi?id=1717746).

bobfuru:
@liangxia or @qinpingli PTAL

bobfuru:
@jkaurredhat PTAL

qinpingli:
@chao007 Could you help review this patch?

install_config/persistent_storage/selector_label_binding.adoc
```diff
@@ -111,6 +111,60 @@ spec:
...
+The following resource shows the required values that you use to configure a StorageClass. This example uses the AWS ElasticBlockStore (EBS) object definition.
+
++
+.StorageClass definition for EBS
+====
+----
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: gp2
```
jsafrane, line 134: I would perhaps use a specific name, for example "kafka", so it's clear the storage class is reserved to it.

"gp2" is our default storage class that might confuse users.

install_config/persistent_storage/selector_label_binding.adoc
jsafrane: The placement is OK, however, the heading does not match the content. Heading says that it's about label selectors, but the two solutions don't use selector at all.

There are two mixed topics:

1. bind a PVC to a specific PV
  * This can be done either by specifying pvc.spec.volumeName (when knowing the PV name) or using pvc.spec.selector (when knowing PV labels). 
2. reserve a PV to specific PVC (or namespace) 
  * This can be done by using a specific storage class or setting pv.spec.claimRef, as outlined correctly below.

I would not mix those in a single chapter.

bobfuru: Thanks, @jsafrane. I understand more now that you spelled out #1. 

I've added an additional section for using a selector or volumeName, so there are now two distinct chapters. Let me know if this looks good.

install_config/persistent_storage/selector_label_binding.adoc
bobfuru: Changed `gp2` to `kafka`

install_config/persistent_storage/selector_label_binding.adoc
```diff
@@ -111,6 +111,69 @@ spec:
...
+To bind a PVC to a specific PV as a cluster administrator:
+
+* Use `pvc.spec.volumeName` if you know the PV name.
+* Use `pvc.spec.selector` if you know the PV labels.
++
+By specifying a selector, the PVC requires the PV to have specific labels.
+
+
+[[selector-label-volume-reserve]]
+=== Optional: Reserve a PV to a specific PVC or namespace
```
chao007, line 128: Could not reserve a pv to namespace.
It is better to drop namespace

install_config/persistent_storage/selector_label_binding.adoc
```diff
@@ -111,6 +111,69 @@ spec:
...
+...
+----
+====
++
+[IMPORTANT]
+====
+If necessary in a multi-tenant environment, use a quota definition to reserve the storage class and PV(s) only to a specific namespace.
+====
+
+. Pre-bind the PV to your PVC using the PVC namespace and name. A PV defined as such will bind only to the specified PVC and to nothing else, as shown in the following example:
```
bobfuru, line 154: @chao007 - I dropped "namespace" from the subhead as you suggested, but is it okay to still have it "PVC namespace" here or should it also be removed?

jsafrane:
lgtm

bobfuru:
@chao007 Can you please approve from QE side?

bobfuru:

bmcelvee:
LGTM

chao007:
LGTM.

PR #19093, opened 2020-01-16
Release notes minor updates

codyhoag:
@sheriff-rh can you take a quick look at this? Just minor updates I had stashed away yesterday.

PR #21026, opened 2020-04-08
Revert dash additions added in 4.3 RNs

codyhoag:
The dashes were added for several features in the 4.3 TP table for consistency purposes (#19093), to show when a feature was not in TP or GA. After a discussion in #20996, several should be removed since the dash (`-`) is strictly used for deprecated/removed features.


kalexand-rh (commented):
LGTM

PR #21051, opened 2020-04-09
OSDOCS-959 cluster-admin documentation added

sheriff-rh:
https://issues.redhat.com/browse/OSDOCS-959
This has been verified by QA. 

Build preview:
http://file.rdu.redhat.com/~antaylor/osdocs-959/administering_a_cluster/cluster-admin-role.html

### DIFF
--- a/modules/dedicated-cluster-admin-enable.adoc
+++ b/modules/dedicated-cluster-admin-enable.adoc
@@ -1,0 +1,15 @@
+// Module included in the following assemblies:
+//
+// administering_a_cluster/cluster-admin-role.adoc
+
+[id="dedicated-cluster-admin-enable"]
+= Enabling the cluster-admin role for your cluster
+
+The cluster-admin role must be enabled at the cluster level before it can be assigned to a user.
+
+.Prerequisites
+. Open a technical support case with Red Hat to request that `cluster-admin` be enabled for your cluster. 
+
+.Procedure
+. In the OpenShift Cluster Manager, select the cluster you want to assign cluster-admin privileges.
+. Under the *Actions* dropdown menu, select *Allow cluster-admin access*.
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -112,6 +112,8 @@ Distros: openshift-dedicated
 Topics:
 - Name: The dedicated-admin role
   File: dedicated-admin-role
+- Name: The cluster-admin role
+  File: cluster-admin-role
 ---
 Name: Installing
 Dir: installing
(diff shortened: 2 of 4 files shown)


modules/dedicated-cluster-admin-enable.adoc
```diff
@@ -0,0 +1,12 @@
+// Module included in the following assemblies:
+//
+// administering_a_cluster/cluster-admin-role.adoc
+
+[id="dedicated-cluster-admin-enable"]
+= Enabling the cluster-admin role to your cluster
```
lamek, line 6: Should this be "... on your cluster" or "...for your cluster"?

modules/dedicated-cluster-admin-enable.adoc
```diff
@@ -0,0 +1,12 @@
...
+
+The cluster-admin role must be enabled at the cluster level before it can be assigned to a user.
+
+.Procedure
+. In the OpenShift Cluster Manager, select the cluster you want to assign cluster-admin privileges.
+. Under the *Actions* drop down menu, select *Allow cluster-admin access*.
```
lamek, line 12: Nit - we usually use "dropdown" instead of "drop down", but no firm rule one way or the other that I know of.

lamek (commented):
One suggestion and one nit. Otherwise LGTM.

jewzaam:
@rogbas @sheriff-rh there isn't in this explaining how to get cluster-admin enabled.  Customer has to be CCS and has to request via ticket enabling this feature for their subscription.  It isn't on by default.

sheriff-rh:
@patrickstrick can you PTAL when you get a chance? thanks!

modules/dedicated-cluster-admin-enable.adoc
```diff
@@ -0,0 +1,15 @@
...
+//
+// administering_a_cluster/cluster-admin-role.adoc
+
+[id="dedicated-cluster-admin-enable"]
+= Enabling the cluster-admin role for your cluster
+
+The cluster-admin role must be enabled at the cluster level before it can be assigned to a user.
+
+.Prerequisites
+. Submit a ticket to Service Reliability Engineering (SRE) to enable `cluster-admin` for your cluster.
```
patrickstrick, line 11: @sheriff-rh @jewzaam @lamek The customer will not know how to contact SRE, and in fact doesn't have a way to do so directly. I believe this should instead say something like, "Open a technical support case with Red Hat to request that `cluster-admin` be enabled for your cluster."

PR #21076, opened 2020-04-09
OSDOCS-973 migrating legacy Azure DNS zones

kalexand-rh:
See https://issues.redhat.com/browse/OSDOCS-973

### DIFF
--- a/installing/installing_azure/migrating-azure-dns-zones.adoc
+++ b/installing/installing_azure/migrating-azure-dns-zones.adoc
@@ -1,0 +1,16 @@
+[id="migrating-azure-dns-zones"]
+= Migrating legacy Azure DNS zones
+include::modules/common-attributes.adoc[]
+:context: migrating-azure-dns-zones
+
+toc::[]
+
+If you installed an {product-title} version 4.2 or 4.3 cluster on Azure that uses installer-provisioned infrastructure, you must migrate the legacy private DNS zone to the new type of private DNS zone.
+
+For more information about private zones in Azure, see link:https://azure.microsoft.com/en-us/updates/announcing-preview-refresh-for-azure-dns-private-zones-2/[Announcing Preview Refresh for Azure DNS Private Zones].
+
+include::modules/migrating-azure-dns-zones.adoc[leveloffset=+1]
+
+.Next steps
+
+* Update your cluster to {product-title} version {product-version}. See xref:../../updating/updating-cluster-between-minor.adoc#updating-cluster-between-minor[Updating a cluster between minor versions].
--- a/installing/installing_azure/migrating-azure-dns-zones.adoc
+++ b/installing/installing_azure/migrating-azure-dns-zones.adoc
@@ -5,7 +5,7 @@ include::modules/common-attributes.adoc[]
 
 toc::[]
 
-If you installed an {product-title} version 4.2 or 4.3 cluster on Azure that uses installer-provisioned infrastructure, you must migrate the legacy private DNS zone to the new type of private DNS zone.
+If you installed an {product-title} version 4.2 or 4.3 cluster on Azure that uses installer-provisioned infrastructure, your cluster might use a legacy private DNS zone. If it does, you must migrate it to the new type of private DNS zone.
 
 For more information about private zones in Azure, see link:https://azure.microsoft.com/en-us/updates/announcing-preview-refresh-for-azure-dns-private-zones-2/[Announcing Preview Refresh for Azure DNS Private Zones].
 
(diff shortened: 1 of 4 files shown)


kalexand-rh:
@jhixson74, will you PTAL?

modules/migrating-azure-dns-zones.adoc
```diff
@@ -0,0 +1,47 @@
...
++
+----
+$ openshift-install migrate azure privatedns migrate \
+  --old-resource-group=<resource_group> \ <1>
+  --new-resource-group=<new_resource_group> \ <2>
+  --zone=example.com \ <3>
+  --virtual-network=<vnet> \ <4>
+  --virtual-network-resource-group=<resource_group>  \ <5>
+  --link <6>
+----
```
jhixson74, line 39: This is still a moving target. I should have the user doc updated in my next push. For now, the current commands for listing eligible zones and migrating them are "azure-privatedns-eligible" and "azure-privatedns-migrate". I've also combined the --old-resource-group and --new-resource-group options into just --resource-group. Other than that, this looks good.

kalexand-rh: Thanks! I'll make these changes now and check back in after your PR merges. Do you know when that will be? We're trying to get all the 4.4 content merged by EOD Friday for our initial translation drop.

jhixson74: My hope is that is is merged by then as well ;-)

jhixson74: It appears it has been merged. So the official method for doing migrations is [here](https://github.com/openshift/installer/pull/3343/files#diff-1a58973fd4d2a19922fce799fdcea569).

kalexand-rh: @jhixson74, thanks! I've updated the commands to match your PR.

kalexand-rh:
@e-tienne, will you PTAL?

modules/migrating-azure-dns-zones.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+.Procedure
+
+. Determine your legacy zones:
++
+----
+$ ./openshift-install migrate azure-privatedns-eligible
+
+INFO[0000] Credentials loaded from file "<path_to>/ServicePrincipal.json"
+INFO[0002] legacy zone=<legacy_zone> resourceGroup=<resource_group_name>
```
jhixson74, line 24: The [0000] and [0002] are removed now, so only INFO shows.

kalexand-rh: Fixed! Thank you!

installing/installing_azure/migrating-azure-dns-zones.adoc
```diff
@@ -0,0 +1,16 @@
+[id="migrating-azure-dns-zones"]
+= Migrating legacy Azure DNS zones
+include::modules/common-attributes.adoc[]
+:context: migrating-azure-dns-zones
+
+toc::[]
+
+If you installed an {product-title} version 4.2 or 4.3 cluster on Azure that uses installer-provisioned infrastructure, you must migrate the legacy private DNS zone to the new type of private DNS zone.
```
bmcelvee, line 8: This introductory content is repeated in the assembly and module. It shouldn't be an issue, but I wanted to make sure it was intentional.

bmcelvee:
LGTM!

e-tienne:
I would like to test this against an actual cluster and when I run `./openshift-install migrate azure-privatedns-eligible` on a Azure IPI 4.3.14 cluster I could not find any legacy Azure DNS zones. The current doc says `4.2` and `4.3`.  What exact versions should have those zones by default?

e-tienne:
> I would like to test this against an actual cluster and when I run `./openshift-install migrate azure-privatedns-eligible` on a Azure IPI 4.3.14 cluster I could not find any legacy Azure DNS zones. The current doc says `4.2` and `4.3`. What exact versions should have those zones by default?

I was able to create a `privatedns-eligible` zone with `4.2.0` but I think we need to add more context and make it obvious when and how this should be used (including when to use the options to link the zone). I'm not sure of the result I currently achieved by running the migration and customers might feel the same.

I think that the feature should require proper testing before we can adjust the docs.

kalexand-rh:
@e-tienne, after you do proper testing, let me know so that I can adjust the docs.

This change has been moved to 4.5.

e-tienne:
I was able to confirm that the first part of the procedure (1. Determine your legacy zones) worked.

However, I could not finish testing as the ability to create new private DNS zone was removed completely in Azure while I was working on the second part (2. Migrate a legacy zone)

I think the doc looks good, but we could also add some details or example scenarios regarding when to use the following options (as a user might be unsure if needs to use those or not):

--virtual-network=<vnet>
--link

PR #21106, opened 2020-04-13
disableDefaultForwarding=false doesnt work

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1818034

Added new module on disabling Log Fowarding.


mburke5678:
@jcantrill @anpingli PTAL

modules/cluster-logging-log-forwarding-about.adoc
```diff
@@ -86,7 +86,7 @@ spec:
      - secureforward-offcluster
 ----
 <1> The name of the log forwarding CR must be `instance`.
-<2> Parameter to enable log forwarding. Set to `true` to enable log forwarding.
+<2> Parameter to enable log forwarding. Set to `true` to enable log forwarding. Do not set to `false`. Setting this parameter to `false` prevents cluster logging from sending logs to the specified endpoints *and* to default internal {product-title} Elasticsearch instance.
```
jcantrill, line 89: This is incorrect.  It should be something like:

```
<2> Parameter to disable the operator from creating log forwarding config to the default cluster logging managed store. Set to `true` to disable generation of default store logforwarding config. Note: this change assumes administrators will define logforwarding for default storage.
```

modules/cluster-logging-log-forwarding-configure.adoc
```diff
@@ -85,7 +85,7 @@ spec:
 ----
 <1> The name of the log forwarding CR must be `instance`.
 <2> The namespace for the log forwarding CR must be `openshift-logging`.
-<3> Set to `enabled` to enable log forwarding.
+<3> Set to `true` to enable log forwarding.
```
jcantrill, line 88: Same comment as previously

jcantrill (changes requested):

modules/cluster-logging-log-forwarding-about.adoc
mburke5678: @jcantrill Thanks for the clarification. I meant here to use `true` to enable the Log Forwarding feature, to keep the wording simple. I can see where that would be confusing as set the parameter to enable forwarding is wrong, as log forwarding is enabled by default. 
I put wording similar to your further up in the file. Again to keep it simple here is the call-out. 
`<2> Parameter to disable the default log forwarding behavior. Set to `true` to enable log forwarding to the specified outputs.`
The default behavior is explained above.

modules/cluster-logging-log-forwarding-configure.adoc
mburke5678: Changed to:
`<3> Set to `true` to enable log forwarding to the specified outputs. `

At this point, I assume that users are not as concerned with turning off the default behavior, just how to make the log forward feature work.

mburke5678: @jcantrill In the Disabling section, I have: 
`Setting disableDefaultForwarding=false prevents cluster logging from sending logs to the specified endpoints and to default internal OpenShift Container Platform Elasticsearch instance.` 
Based on https://bugzilla.redhat.com/show_bug.cgi?id=1818034, which doesn't sound like will get fixed. If we want this info here, do we want it earlier when I introduce the parameter?

modules/cluster-logging-log-forwarding-about.adoc
```diff
@@ -86,7 +88,7 @@ spec:
      - secureforward-offcluster
 ----
 <1> The name of the log forwarding CR must be `instance`.
-<2> Parameter to enable log forwarding. Set to `true` to enable log forwarding.
+<2> Parameter to disable the default log forwarding behavior. Set to `true` to enable log forwarding to the specified outputs.
```
jcantrill, line 91: This switch does not explicitly enable `enable log forwarding to the specified outputs`; that is what the annotation essentially does.  This switch only *disables* the operator from automatically forwarding logs to its own managed logstore

jcantrill (changes requested):

modules/cluster-logging-log-forwarding-configure.adoc
```diff
@@ -85,7 +85,7 @@ spec:
 ----
 <1> The name of the log forwarding CR must be `instance`.
 <2> The namespace for the log forwarding CR must be `openshift-logging`.
-<3> Set to `enabled` to enable log forwarding.
+<3> Set to `true` to enable log forwarding to the specified outputs.
```
jcantrill, line 88: Looks like this was missed from the previous change `<2> Parameter to disable the default log forwarding behavior.`

jcantrill (changes requested):

modules/cluster-logging-log-forwarding-about.adoc
```diff
@@ -9,17 +9,19 @@ The {product-title} cluster log forwarding feature uses a combination of _output
 
 [NOTE]
 ====
-If you want to use only the default internal {product-title} Elasticsearch instance, do not configure any outputs and pipelines. 
+If you want to use only the default internal {product-title} Elasticsearch logstore, do not configure any outputs and pipelines. 
 ==== 
 
-An output is the destination for log data and a pipeline defines simple routing for one source to one or more outputs. 
+By default, the Cluster Logging Operator creates `logforwarding` configuration file that forwards logs to the default internal Elasticsearch logstore. To use the log forwarding feature, you create a custom `logforwarding` configuration file to send logs to endpoints you specify. In this configuration file, the `disableDefaultForwarding=true` parameter disables the default forwarding behavior. This change assumes you define outputs and pipelines to forward logs for storage.
```
anpingli, line 15: "By default, the Cluster Logging Operator creates `logforwarding` configuration file that forwards logs to the default internal Elasticsearch logstore. ".  That confuse me. 
Could we use 'If elasticsearch are defined in the clusterlogging CRD. The Cluster Logging Operator will enable the fluentd to ship logs the default internal Elasticsearch logstore'

kind: "ClusterLogging"
metadata:
  name: "instance"
  namespace: openshift-logging
spec:
  managementState: "Managed"
  logStore:
    type: "elasticsearch"
    elasticsearch:
      nodeCount: 3
      resources:
        limits:
          memory: 2Gi
        requests:
          cpu: 200m
          memory: 2Gi
      storage:
        size: "20G"
        storageClassName: "gp2"
      redundancyPolicy: "SingleRedundancy"
  visualization:
    type: "kibana"
    kibana:
      replicas: 1
  curation:
    type: "curator"
    curator:
      schedule: "30 3 * * *"
  collection:
    logs:
      type: "fluentd"
      fluentd: {}

anpingli: There is a plugin named forward. How about replace some the word 'forward' with 'sent , ship, copy or transfer'?

anpingli, line 15: Could we remove the disableDefaultForwarding  here?

mburke5678: _By default, the Cluster Logging Operator sends logs to the default internal Elasticsearch logstore, defined in the Cluster Logging Custom Resource. To use the log forwarding feature, you create a custom `logforwarding` configuration file to send logs to endpoints you specify._

mburke5678: Yes. See https://github.com/openshift/openshift-docs/pull/21106#discussion_r413353866

mburke5678:
@anpingli Do these changes meet your expectations?

modules/cluster-logging-log-forwarding-about.adoc
anpingli: I don't like the word '`logforwarding` configuration file '  in "The Cluster Logging Operator creates `logforwarding` configuration file that forwards logs to the default internal Elasticsearch logstore".  as we couldn't find a file named logforwarding.

mburke5678:
@anpingli Do these changes meet your expectations?

anpingli (commented):
LGTM

modules/cluster-logging-log-forwarding-disable.adoc
```diff
@@ -0,0 +1,45 @@
...
+[source,yaml]
+----
+apiVersion: "logging.openshift.io/v1"
+kind: "ClusterLogging"
+metadata:
+  annotations:
+    clusterlogging.openshift.io/logforwardingtechpreview: enabled <1>
+  name: "instance"
+  namespace: "openshift-logging"
+....
```
jboxman, line 36: s/..../.../

logging/config/cluster-logging-log-forwarding.adoc
```diff
@@ -33,7 +33,8 @@ include::modules/cluster-logging-log-forwarding-about.adoc[leveloffset=+1]
 
 include::modules/cluster-logging-log-forwarding-configure.adoc[leveloffset=+1]
 
-== Additional resources
+include::modules/cluster-logging-log-forwarding-disable.adoc[leveloffset=+1]
 
-Alternatively, you can use Fluentd plugins to forward logs. For more information, see xref:../../logging/config/cluster-logging-external.adoc#cluster-logging-external[Sending logs to external devices using Fluentd plugins].
+== Additional resources
```
jboxman, line 38: s/== Additional resources/.Additional References/

logging/config/cluster-logging-log-forwarding.adoc
```diff
@@ -33,7 +33,8 @@ include::modules/cluster-logging-log-forwarding-about.adoc[leveloffset=+1]
...
 
+Alternatively, you can use Fluentd *forward* protocol or syslog to forward logs. For more information, see xref:../../logging/config/cluster-logging-external.adoc#cluster-logging-external[Forward logs to third party systems].
```
jboxman, line 40: possibly;

s/can use/can use the/

modules/cluster-logging-log-forwarding-about.adoc
```diff
@@ -9,17 +9,19 @@ The {product-title} cluster log forwarding feature uses a combination of _output
 
 [NOTE]
 ====
-If you want to use only the default internal {product-title} Elasticsearch instance, do not configure any outputs and pipelines. 
+If you want to use only the default internal {product-title} Elasticsearch logstore, do not configure the Log Forwarding feature. 
 ==== 
 
-An output is the destination for log data and a pipeline defines simple routing for one source to one or more outputs. 
+By default, the Cluster Logging Operator sends logs to the default internal Elasticsearch logstore, as defined in the Cluster Logging Custom Resource. To use the Log Forwarding feature, you create a custom `logforwarding` configuration file to send logs to endpoints you specify.
```
jboxman, line 15: I don't think we've been internally consistent with our usage of custom resource, partially because upstream and developer usage may be inconsistent.

In any case, a custom resource is a resource defined using k8s CustomResourceDefinition object. So that's where _custom resource_, and thus _CR_ comes from. I don't think it should be capitalized. It's doubly confusing because developers will abbreviate resources, such as CustomResourceDefinition or HorizontalPodAutoscaler as CRD or HPA, but they'r really resource names... So it's different than abbreviating, say, RAM, which is really a proper noun, abbreviated.

Sigh.

So;

s/Custom Resource/custom resource/

modules/cluster-logging-log-forwarding-disable.adoc
```diff
@@ -0,0 +1,45 @@
+// Module included in the following assemblies:
+//
+// * logging/cluster-logging-log-forwarding.adoc
+
+[id="cluster-logging-log-forwarding-disable_{context}"]
+= Disabling the Log Forwarding feature
+
+To disable the Log Forwarding feature, remove the `clusterlogging.openshift.io/logforwardingtechpreview:enabled` parameter from the Cluster Logging CR and delete the Log Forwarding CR. The container and node logs will be forwarded to the internal {product-title} Elasticsearch instance.
```
jboxman, line 8: I'd define custom resource (CR) here, and then not define it in the _Procedure_ that follows later.

modules/cluster-logging-log-forwarding-disable.adoc
```diff
@@ -0,0 +1,45 @@
...
+[NOTE]
+====
+You cannot disable Log Forwarding by setting the `disableDefaultForwarding` to `false` in the Log Forwarding Custom Resource. This prevents cluster logging from sending logs to the specified endpoints *and* to default internal {product-title} Elasticsearch instance. 
+====
+
+.Procedure
+
+To disable the Log Forwarding feature:
+
+. Edit the Cluster Logging Custom Resource (CR) in the `openshift-logging` project:
```
jboxman, line 19: This usage of CR can be defined earlier, and not defined here.

modules/cluster-logging-log-forwarding-disable.adoc
```diff
@@ -0,0 +1,45 @@
+// Module included in the following assemblies:
+//
+// * logging/cluster-logging-log-forwarding.adoc
+
+[id="cluster-logging-log-forwarding-disable_{context}"]
+= Disabling the Log Forwarding feature
+
+To disable the Log Forwarding feature, remove the `clusterlogging.openshift.io/logforwardingtechpreview:enabled` parameter from the Cluster Logging CR and delete the Log Forwarding CR. The container and node logs will be forwarded to the internal {product-title} Elasticsearch instance.
+
+[NOTE]
```
jboxman, line 10: Is this _[IMPORTANT]_ instead?

modules/cluster-logging-log-forwarding-configure.adoc
```diff
@@ -83,18 +83,18 @@ spec:
...
-<3> Set to `enabled` to enable log forwarding.
+<1> The name of the Log Forwarding CR must be `instance`.
+<2> The Namespace for the Log Forwarding CR must be `openshift-logging`.
+<3> Set to `true` to disable the default log forwarding behavior of sending all logs to the internal Elasticsearch instance. 
 <4> Add one or more endpoints:
 * Specify the type of output, either `elasticseach` or `forward`.
 * Enter a name for the output.
 * Enter the endpoint, either the server name or FQDN.
-* Optionally, enter the name of the secret required by the endpoint for TLS communication. The secret must exist in the openshift-logging project.
+* Optionally, enter the name of the secret required by the endpoint for TLS communication. The secret must exist in the `openshift-logging` project.
```
jboxman, line 93: s/Optionally,/Optional:/

modules/cluster-logging-log-forwarding-configure.adoc
```diff
@@ -83,18 +83,18 @@ spec:
      outputRefs:
      - secureforward-offcluster
 ----
-<1> The name of the log forwarding CR must be `instance`.
-<2> The namespace for the log forwarding CR must be `openshift-logging`.
-<3> Set to `enabled` to enable log forwarding.
+<1> The name of the Log Forwarding CR must be `instance`.
+<2> The Namespace for the Log Forwarding CR must be `openshift-logging`.
```
jboxman, line 87: This one is fun, because [Namespace](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#namespace-v1-core) is a k8s resource, but it's also a field in every single CR, I think without exception, and references a Namespace object.

Here we mean the `namespace` field, though. So I'd say:

The `namespace` for the Log Forwarding CR must be `openshift-logging`.

jboxman (changes requested):
Made some observations; let me know if anything doesn't make sense.

Thanks!

modules/cluster-logging-log-forwarding-about.adoc
mburke5678: I'm going to leave this one for now, if you don't mind. it would require 44 changes in 28 files (per Atom) to make `Cluster Logging custom resource` consistent. I can change these all in a single sweep in a separate PR.

logging/config/cluster-logging-log-forwarding.adoc
mburke5678: @jboxman 
s/== Additional resources/.Additional **resources**/ ??

PS. I hate this one. The Additional resources should be a heading that appears in the TOC, which allows the resource to apply to the entire assembly. As `.Additional reource`, it's not in the TOC and seems to apply to the topic that happens to be last. 

Sigh.

modules/cluster-logging-log-forwarding-about.adoc
jboxman: @mburke5678, oh, that's interesting; Usually it's [Object] CR. This is _feature name_ CR? Or is there a ClusterLogging object?

mburke5678: @jboxman It's a custom resource that the Cluster Logging Operator uses when creating the various cluster logging components (Elasticearch, Kibana, Fluentd, Curator, etc.). It's `kind: "ClusterLogging"` and named _instance_. 
Probably should refer to it as the _ClusterLogging CR_?

jboxman: @mburke5678 probably for now. I wanted to devise a convention for this stuff, but it's intractable. It would be nice to be able to write about _cluster logging_, and mention the ClusterLogging CR when directly interacting with it. But that approach might not stand up to scrutiny.

Issue #18426, opened 2019-12-03
OCP 4.4 Release Notes Tracker


ahardin-rh:
Please leave comments here for anything that should be highlighted in the 4.4 release notes. Thank you!

Draft Preview: https://docs.openshift.com/container-platform/4.4/release_notes/ocp-4-4-release-notes.html

deads2k:
In 4.3 and before, the kube-apiserver checked tokens before client certificates for authentication.  In 4.4 and after, the kube-apiserver will check client-certificates before tokens.  As an example, this means that if you have a system:admin kubeconfig and run `oc --token=foo get pod` in 4.3 that would authenticate as a the user with token foo, in 4.4 and after it will authenticate as system:admin.  The recommendation for many releases has been to use `--as` in such cases instead of overriding the token when using a client-certificate as well.

soltysh:
`oc build-logs` was deprecated in 3.11 (https://github.com/openshift/origin/pull/20322) and we are now removing it entirely from `oc` (https://github.com/openshift/oc/pull/232). Users should use `oc logs` instead.

stlaz:
As stated in 4.1 release notes, the deprecated `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt` file will be removed in 4.5.

soltysh:
oc's `--config` flag is being deprecated in 4.4, users should start using `--kubeconfig`, instead.
`OS_EDITOR` is also being deprecated in 4.4, users should start using `KUBE_EDITOR` or `EDITOR`, instead.

soltysh:
`oc rsh --timeout` flag is being deprecated in https://github.com/openshift/oc/pull/260, users should use `--request-timeout`, instead.

codyhoag:
The `machineCIDR` network parameter has been deprecated. It's recommended to use `machineNetwork.cidr` instead.

See #19212 for sample entry for release notes.

https://issues.redhat.com/browse/OSDOCS-877

soltysh:
`oc secrets` subcommands (`new`, `new-basicauth`, `new-dockercfg`, `new-sshauth`) were deprecated in 3.9 (https://github.com/openshift/origin/commit/5046e19f71a6480c131fbf6bd19c73b8a9d47018), they are being removed entirely now. Users should use `oc create secret` command, instead.

jboxman:
[...]

PR #21112, opened 2020-04-13
CephFS and Manila provisioners were removed for 4.x

codyhoag:
This was manually cherry-picked to 4.1-4.3 RNs too:
#21109 (4.1)
#21110 (4.2)
#21111 (4.3)

### DIFF
--- a/release_notes/ocp-4-4-release-notes.adoc
+++ b/release_notes/ocp-4-4-release-notes.adoc
@@ -538,10 +538,10 @@ In the table below, features are marked with the following statuses:
 |-
 |TP
 
-|CephFS
-|TP
-|TP
-|
+|CephFS Provisioner
+|-
+|-
+|-
 
 |Podman
 |TP
@@ -559,9 +559,9 @@ In the table below, features are marked with the following statuses:
 |
 
 |Manila Provisioner
-|TP
-|TP
-|
+|-
+|-
+|-
 
 |Cluster Administrator console
 |GA


codyhoag:
This was proposed/confirmed by QE via RN Tracker [comments](https://github.com/openshift/openshift-docs/issues/18426#issuecomment-611329059) and Slack:
>Qin Ping  9:30 PM
Hi Cody, for the CephFS and Manila provisoiner,, I checked the images we released, did not find 4.x images for these two images.
cephfs-provisioner: https://comet.engineering.redhat.com/containers/repositories/5ad908debed8bd441305cce3#tab-configuration (looks we never release the image.)
manila-provisioner: https://comet.engineering.redhat.com/containers/repositories/5ad80b8fbed8bd4413058022#tab-configuration (only images for 3.10 and 3.11)
9:31
I checked the JIRA ticket, it's csi  manila, not the same with manila-provisioner.

>Qin Ping  9:43 PM
Another thing is from the Jan's comment, we can see for we used RHCOS from 4.x, so in-tree ceph can't be used.

>Cody Hoag  11:54 AM
Thanks for investigating! I took a look back at previous release notes, and it appears we started listing CephFS and Manila provisioners as TP starting in 3.10. Sounds like it has never been TP for any 4.x releases. I'll mark them as removed in the TP table for 4.1-4.4.

kalexand-rh (commented):
Part of me wants to question why we're including items in the TP table that have never been TP for this major version, but the change LGTM.

vikram-redhat:
@codyhoag sounds good. Let's change it for the next release cycle.

PR #21137, opened 2020-04-14
BZ1704090 - Add registry encryption variables for configuring S3

bobfuru:
[BZ 1704090](https://bugzilla.redhat.com/show_bug.cgi?id=1704090) - adds the four encryption variables to set initial registry to encryption configuration, as  noted in the BZ.
@wzheng1 PTAL

### DIFF
--- a/install_config/topics/configuring_for_aws_registry-proc.adoc
+++ b/install_config/topics/configuring_for_aws_registry-proc.adoc
@@ -76,16 +76,19 @@ openshift_hosted_registry_storage_s3_bucket=openshift-registry-storage <3>
 openshift_hosted_registry_storage_s3_region=us-east-1 <4>
 openshift_hosted_registry_storage_s3_chunksize=26214400
 openshift_hosted_registry_storage_s3_rootdirectory=/registry
+openshift_hosted_registry_storage_s3_encrypt=false
+openshift_hosted_registry_storage_s3_kmskeyid=aws_kms_key_id <5>
 openshift_hosted_registry_pullthrough=true
 openshift_hosted_registry_acceptschema2=true
 openshift_hosted_registry_enforcequota=true
 openshift_hosted_registry_replicas=3
+
 ----
 <1> The access key for the IAM user. (Not required with IAM Roles in place)
 <2> The secret key for the IAM user. (Not required with IAM Roles in place)
 <3> The S3 storage bucket name.
 <4> The region in which the bucket exists.
-
+<5> The AWS Key Management Service (AWS KMS) key ID of the encryption key used to encrypt data in the cluster.
 
 === Manually configuring {product-title} registry to use S3
 


install_config/topics/configuring_for_aws_registry-proc.adoc
```diff
@@ -80,12 +80,18 @@ openshift_hosted_registry_pullthrough=true
 openshift_hosted_registry_acceptschema2=true
 openshift_hosted_registry_enforcequota=true
 openshift_hosted_registry_replicas=3
+# AWS Registry Encryption Configuration
+openshift_hosted_registry_storage_s3_encrypt
```
dmage, line 84: openshift_hosted_registry_storage_s3_encrypt=false (if we want to have defaults there)

And I'd put it after `openshift_hosted_registry_storage_s3_rootdirectory`, because it's also about s3 storage.

install_config/topics/configuring_for_aws_registry-proc.adoc
```diff
@@ -80,12 +80,18 @@ openshift_hosted_registry_pullthrough=true
...
+openshift_hosted_registry_storage_s3_kmskeyid
+secure: default ('true') <5>
+v4auth: default ('true') <6>
```
dmage, line 87: We don't have `secure` not `v4auth` in Ansible.

install_config/topics/configuring_for_aws_registry-proc.adoc
```diff
@@ -80,12 +80,18 @@ openshift_hosted_registry_pullthrough=true
 openshift_hosted_registry_acceptschema2=true
 openshift_hosted_registry_enforcequota=true
 openshift_hosted_registry_replicas=3
+# AWS Registry Encryption Configuration
+openshift_hosted_registry_storage_s3_encrypt
+openshift_hosted_registry_storage_s3_kmskeyid
```
dmage, line 85: I'm not sure if you can have variables without any value in Ansible inventory files.

@mtnbikenc you made changes in our playbooks recently, is it valid syntax?

install_config/topics/configuring_for_aws_registry-proc.adoc
bobfuru: @dmage I think we do want defaults, thanks. I added `=false` and moved as indicated.

install_config/topics/configuring_for_aws_registry-proc.adoc
bobfuru: Removed `secure` and `v4auth` here.

install_config/topics/configuring_for_aws_registry-proc.adoc
bobfuru: @mtnbikenc @dmage I dug a little and added what I could about this key id (including annotation), but not sure if there is an actual value that should be included here instead?

bobfuru:

install_config/topics/configuring_for_aws_registry-proc.adoc
dmage: :+1:, looks good

wzheng1:
@bobfuru Thanks @dmage 's help reviewing it, LGTM now.

PR #21175, opened 2020-04-15
moving 3.x contrib guide to 3.11 branch

kalexand-rh:
@vikram-redhat, I think we should retain the 3.x contribution guide. We changed enough between the 3.x and 4.x stream that it's useful to retain the old information.

What do you think?

### DIFF
--- a/contributing_to_docs/contributing_user_stories.adoc
+++ b/contributing_to_docs/contributing_user_stories.adoc
@@ -1,0 +1,124 @@
+[[contributing-user-stories]]
+= Contribute user stories to OpenShift documentation
+:icons:
+:toc: macro
+:toc-title:
+:toclevels: 1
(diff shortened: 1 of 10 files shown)


vikram-redhat:
Great point. Yes, makes absolute sense.

contributing_to_docs/contributing_user_stories.adoc
```diff
@@ -0,0 +1,124 @@
...
+
+== What if I don't want to write in modules?
+If you don't want to write the modules yourself but have a content change,
+write a user story, provide details to support the story, and reach out to the
+OpenShift docs team.
+
+== Where do I contribute modularized content?
+If you have new content for OCP 4.1, it must be modularized before it is merged
+to the collection.
+Here is an overview of the active branches:
```
vikram-redhat, line 30: Might need editing to include more recent branches

PR #21170, opened 2020-04-14
BZ1823977 - Better approval command for CSR

vwalek:

PR #21176, opened 2020-04-15
BZ-1823977: Updating command to approval all pending CSRs

bergerhoffer:
Apply change from https://bugzilla.redhat.com/show_bug.cgi?id=1823977 to 4.x. Command accomplishes the same thing, but this one doesn't require the jq package to be installed.


kalexand-rh (commented):
LGTM

bergerhoffer:
@barleyer Can you check this as well? This applies the change from #21170 to 4.x. I tested it on 4.4 and the command works well there as well.

PR #21186, opened 2020-04-15
CNV-3573 clone local volume to different node

aburdenthehand:
New module and assembly to cover moving a VM that uses local storage to a specific node. This process requires the user to label a PV on the destination node and clone the underlying volume of the VM using a DataVolume that references the PV label.

I've included this under 'Virtual Machine Disks' instead of 'Cloning Virtual Machines' because although the process requires cloning, the user ultimately wants to move the virtual machine to a specific node.

### DIFF
--- a/modules/cnv-cloning-local-volume-to-another-node.adoc
+++ b/modules/cnv-cloning-local-volume-to-another-node.adoc
@@ -1,0 +1,130 @@
+// Module included in the following assemblies:
+//
+// * cnv/cnv_virtual_machines/cnv_virtual_disks/cnv-moving-local-vm-disk-to-different-node.adoc
+
+[id="cnv-cloning-local-volume-to-another-node_{context}"]
+= Cloning a local volume to another node
+
+You can move a virtual machine disk so that it runs on a specific node by cloning the underlying PersistentVolumeClaim (PVC).
+
+To ensure the virtual machine disk is cloned to the correct node, you must either create a new PersistentVolume (PV) or identify one on the correct node.
+Apply a unique label to the PV so that it can be referenced by the DataVolume.
+
+[NOTE]
+====
+The destination PV must be the same size or larger than the source PVC.
+If the destination PV is smaller than the source PVC, the cloning operation fails.
+====
+
+.Prerequisites
+
+* The virtual machine must not be running. Power down the virtual machine before cloning the virtual machine disk. 
+
+.Procedure
+
+. Either create a new local PV on the node, or identify a local PV already on the node:
+
+* Create a local PV that includes the `nodeAffinity.nodeSelectorTerms` parameters. The following manifest creates a `10Gi` local PV on `node01`. 
++
+[source,yaml]
+----
+kind: PersistentVolume
+apiVersion: v1
+metadata:
+  name: <destination-pv> <1>
+  annotations:
+spec:
+  accessModes:
+  - ReadWriteOnce
+  capacity:
+    storage: 10Gi <2>
+  local:
+    path: /mnt/local-storage/local/disk1 <3>
+  nodeAffinity:
+    required:
+      nodeSelectorTerms:
+      - matchExpressions:
+        - key: kubernetes.io/hostname
+          operator: In
+          values:
+          - node01 <4>
+  persistentVolumeReclaimPolicy: Delete
+  storageClassName: local
+  volumeMode: Filesystem
+----
+<1> The name of the PV.
+<2> The size of the PV. You must allocate enough space, or the cloning operation fails. The size must be the same as or larger than the source PVC.
+<3> The mount path on the node.
+<4> The name of the node where you want to create the PV.
+
+* Identify a PV that already exists on the target node. You can identify the node where a PV is provisioned by viewing the `nodeAffinity` field in its configuration:
++
+----
+$ oc get pv <destination-pv> -o yaml
+----
++
+The following snippet shows that the PV is on `node01`:
++
+[source,yaml]
+----
+...
+spec:
+  nodeAffinity:
+    required:
+      nodeSelectorTerms:
+      - matchExpressions:
+        - key: kubernetes.io/hostname <1>
+          operator: In
+          values:
+          - node01 <2>
+...
+----
+<1> The `kubernetes.io/hostname` key uses the node host name to select a node.
+<2> The host name of the node. 
+
+. Add a unique label to the PV:
++
+----
+$ oc label pv <destination-pv> node=node01
+----
+
+. Create a DataVolume manifest that references the following:
+
+* The PVC name and namespace of the virtual machine.
+* The label you applied to the PV in the previous step.
+* The size of the destination PV.
++
+[source,yaml]
+----
+apiVersion: cdi.kubevirt.io/v1alpha1
+kind: DataVolume
+metadata:
+  name: <clone-datavolume> <1>
+spec:
+  source:
+    pvc:
+      name: "<source-vm-disk>" <2>
+      namespace: "<source-namespace>" <3>
+  pvc:
+    accessModes:
+      - ReadWriteOnce
+    selector:
+      matchLabels:
+        node: node01 <4>
+    resources:
+      requests:
+        storage: <10Gi> <5>
+----
+<1> The name of the new DataVolume.
+<2> The name of the source PVC. If you do not know the PVC name, you can find it in the virtual machine configuration: `spec.volumes.persistentVolumeClaim.claimName`.
+<3> The namespace where the source PVC exists.
+<4> The label that you applied to the PV in the previous step.
+<5> The size of the destination PV.
+
+. Start the cloning operation by applying the DataVolume manifest to your cluster:
++
+----
+$ oc apply -f <clone-datavolume.yaml>
+----
+
+The DataVolume clones the PVC of the virtual machine into the PV on the specific node.
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -1527,6 +1527,8 @@ Topics:
       File: cnv-uploading-local-disk-images-virtctl
     - Name: Uploading a local disk image to a block storage DataVolume
       File: cnv-uploading-local-disk-images-block
+    - Name: Moving a local virtual machine disk to a different node
+      File: cnv-moving-local-vm-disk-to-different-node
     - Name: Expanding virtual storage by adding blank disk images
       File: cnv-expanding-virtual-storage-with-blank-disk-images
     - Name: Preparing CDI scratch space
--- a/cnv/cnv_virtual_machines/cnv_virtual_disks/cnv-moving-local-vm-disk-to-different-node.adoc
+++ b/cnv/cnv_virtual_machines/cnv_virtual_disks/cnv-moving-local-vm-disk-to-different-node.adoc
@@ -1,0 +1,22 @@
+[id="cnv-moving-local-vm-disk-to-different-node"]
+= Moving a local virtual machine disk to a different node
+include::modules/cnv-document-attributes.adoc[]
+:context: cnv-moving-local-vm-disk-to-different-node
+toc::[]
+
+Virtual machines that use local volume storage can be moved so that they run on a specific node. 
+
+You might want to move the virtual machine to a specific node for the following reasons:
+
+* The current node has limitations to the local storage configuration.
+* The new node is better optimized for the workload of that virtual machine. 
+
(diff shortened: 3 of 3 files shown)


modules/cnv-cloning-local-volume-to-another-node.adoc
```diff
@@ -0,0 +1,121 @@
+// Module included in the following assemblies:
+//
+// * cnv/cnv_virtual_machines/cnv_virtual_disks/cnv-moving-local-vm-disk-to-different-node.adoc
+
+[id="cnv-cloning-local-volume-to-another-node_{context}"]
+= Cloning a local volume to another node
+
+You can move a virtual machine so that it runs on a specific node by cloning the underlying PersistentVolumeClaim (PVC).
```
awels, line 8: You can move a virtual machine __disk__ so that it resides on a specific node by cloning the underlying PersistentVolumeClaim (PVC)

modules/cnv-cloning-local-volume-to-another-node.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+To ensure the virtual machine is cloned to the correct node, either create a PersistentVolume (PV) or identify one already on the node and apply a unique label.
```
awels, line 10: To ensure the virtual machine __disk__ is cloned to the correct node,

modules/cnv-cloning-local-volume-to-another-node.adoc
```diff
@@ -0,0 +1,121 @@
...
+  persistentVolumeReclaimPolicy: Delete
+  storageClassName: local
+  volumeMode: Filesystem
+----
+<1> The name of the PV.
+<2> The size of the PV. You must allocate enough space, or the cloning operation fails. The size must be the same as or larger than the source PVC.
+<3> The mount path on the node.
+<4> The name of the node on which you want the PV to be created.
+
+* You can identify the node on which the PV is provisioned by examining the `nodeAffinity` parameter of the PV configuration:
```
awels, line 55: You can identify the node on which the PV is provisioned by examining the `nodeAffinity` **field** of the PV. In particular look for the kubernetes.io/hostname key and the matching values field to determine the node the PV is on.

awels (commented):
Please note that this procedure only works 'offline' aka the VM must be powered off before transferring the disk

aburdenthehand:
@awels Thanks for the feedback - I've updated as suggested, and it reminded me to add a note about additional user permissions to clone across namespaces (if user is not a cluster-admin)

qwang1:
LGTM, thanks.

modules/cnv-cloning-local-volume-to-another-node.adoc
```diff
@@ -0,0 +1,125 @@
...
+= Cloning a local volume to another node
+
+You can move a virtual machine disk so that it runs on a specific node by cloning the underlying PersistentVolumeClaim (PVC).
+
+To ensure the virtual machine disk is cloned to the correct node, either create a PersistentVolume (PV) or identify one already on the node and apply a unique label.
+Reference this label in the DataVolume so that the underlying PVC of the virtual machine is cloned to the PV on the correct node. 
+
+[NOTE]
+====
+The destination PV must be the same size or larger than the source PVC or else the cloning operation will fail.
```
ousleyp, line 15: s/will fail/fails

Suggest: `The destination PV must be the same size or larger than the source PVC. If the destination PV is smaller than the source PVC, the cloning operation fails.`

modules/cnv-cloning-local-volume-to-another-node.adoc
```diff
@@ -0,0 +1,125 @@
...
+
+.Prerequisites
+
+* The virtual machine must not be running. Power down the virtual machine before cloning the virtual machine disk. 
+
+.Procedure
+
+. Either create a new local PV on the node, or identify a local PV already on the node:
+
+* Create a local PV that includes the `nodeAffinity.nodeSelectorTerms` parameters. The following manifest creates a 10Gb local PV on node01.
```
ousleyp, line 26: Should this be 10Gi (as it's listed in the YAML)?  I would also consider putting the values that you take from the YAML in backticks here.

modules/cnv-cloning-local-volume-to-another-node.adoc
```diff
@@ -0,0 +1,125 @@
...
+          values:
+          - <node01> <4>
+  persistentVolumeReclaimPolicy: Delete
+  storageClassName: local
+  volumeMode: Filesystem
+----
+<1> The name of the PV.
+<2> The size of the PV. You must allocate enough space, or the cloning operation fails. The size must be the same as or larger than the source PVC.
+<3> The mount path on the node.
+<4> The name of the node on which you want the PV to be created.
```
ousleyp, line 57: I think that all of these values (in the YAML) should either be something like `<node-name>` or a literal example name like `node01` without `<>`. Currently there's a mix.

ousleyp: Maybe:`The name of the node where you want to create the PV.`

modules/cnv-cloning-local-volume-to-another-node.adoc
```diff
@@ -0,0 +1,125 @@
...
+
+* You can identify the node on which the PV is provisioned by examining the `nodeAffinity` field of the PV configuration, in particular the `kubernetes.io/hostname` key and matching values:
```
ousleyp, line 59: I think this bullet point should start with something that more clearly differentiates it as a separate option from the previous bullet point, if that makes sense. Like "Identify a PV that already exists on the target node. You can identify the node where a PV is provisioned by viewing the `nodeAffinity` field in its configuration (manifest?):" (or maybe just say "by viewing its configuration")

Then I'd put callouts in the YAML itself to show which specific fields you're referring to (the `kubernetes.io/hostname` and `values:` parts).

modules/cnv-cloning-local-volume-to-another-node.adoc
```diff
@@ -0,0 +1,125 @@
...
+...
+spec:
+  nodeAffinity:
+    required:
+      nodeSelectorTerms:
+      - matchExpressions:
+        - key: kubernetes.io/hostname
+          operator: In
+          values:
+          - node01
```
ousleyp, line 78: add callout

modules/cnv-cloning-local-volume-to-another-node.adoc
```diff
@@ -0,0 +1,125 @@
...
+...
+----
+
+. Once you have created or identified the PV on the correct node, add a unique label to that PV:
```
ousleyp, line 82: Nit: It might be fine to just say "Add a unique label to the PV" here

modules/cnv-cloning-local-volume-to-another-node.adoc
```diff
@@ -0,0 +1,125 @@
...
++
+----
+$ oc label pv <destination-pv> node=node01
+----
+
+. Create a DataVolume manifest that references the PVC of the virtual machine you are cloning to the new node, and the `selector` label that was applied to the PV in the previous step:
```
ousleyp, line 88: I don't think this should say `selector` because you didn't mention `selector` in the previous step. I would also rather this be more minimalistic. 

Suggestion:
> Create a DataVolume manifest that references:
> - The PVC of the virtual machine that you want to clone.
> - The label that you applied to the PV in the previous step.
> For example:

modules/cnv-cloning-local-volume-to-another-node.adoc
```diff
@@ -0,0 +1,125 @@
...
+  pvc:
+    accessModes:
+      - ReadWriteOnce
+    selector:
+      matchLabels:
+        node: node01 <4>
+    resources:
+      requests:
+        storage: <10Gi> <5>
+----
```
ousleyp, line 110: Same comments as the previous YAML

modules/cnv-cloning-local-volume-to-another-node.adoc
```diff
@@ -0,0 +1,125 @@
...
+<1> The name of the new DataVolume.
+<2> The name of the source PVC. If you do not know the PVC name, you can find it in the virtual machine configuration: `spec.volumes.persistentVolumeClaim.claimName`.
+<3> The namespace where the source PVC exists.
+<4> The label applied to the PV in the previous step.
```
ousleyp, line 114: s/applied/that you applied

modules/cnv-cloning-local-volume-to-another-node.adoc
```diff
@@ -0,0 +1,125 @@
...
+<5> The size of the new DataVolume. You must allocate enough space, or the
+cloning operation fails. The size must be the same as or larger than the source PVC.
+
+. Apply the DataVolume manifest to initiate the cloning operation:
```
ousleyp, line 118: Nit: how about `Start the cloning operation by applying the DataVolume manifest to your cluster:`

modules/cnv-cloning-local-volume-to-another-node.adoc
```diff
@@ -0,0 +1,125 @@
...
++
+----
+$ oc apply -f <clone-datavolume.yaml>
+----
+
+The DataVolume clones the PVC of the virtual machine into the new PV on the specific node. After the cloning operation is complete, you can edit the virtual machine configuration so that it uses the new DataVolume, or add the new DataVolume to another virtual machine.
```
ousleyp, line 124: I wonder if you should put this info (starting with "After the cloning operation...") as two bullet points at the end of the assembly instead, so that you can xref to assemblies that describe how to do these actions.

modules/cnv-cloning-local-volume-to-another-node.adoc
```diff
@@ -0,0 +1,125 @@
...
+cloning operation fails. The size must be the same as or larger than the source PVC.
+
+. Apply the DataVolume manifest to initiate the cloning operation:
++
+----
+$ oc apply -f <clone-datavolume.yaml>
+----
+
+The DataVolume clones the PVC of the virtual machine into the new PV on the specific node. After the cloning operation is complete, you can edit the virtual machine configuration so that it uses the new DataVolume, or add the new DataVolume to another virtual machine. 
+
```
ousleyp, line 125: rogue `+`?

modules/cnv-cloning-local-volume-to-another-node.adoc
ousleyp, line 124: re: `the new PV` -- it isn't necessarily new, right?

ousleyp (commented):
This mostly LGTM with some nits and suggestions (mostly related to minimalism).

modules/cnv-cloning-local-volume-to-another-node.adoc
aburdenthehand: Great idea

modules/cnv-cloning-local-volume-to-another-node.adoc
```diff
@@ -0,0 +1,131 @@
+// Module included in the following assemblies:
+//
+// * cnv/cnv_virtual_machines/cnv_virtual_disks/cnv-moving-local-vm-disk-to-different-node.adoc
+
+[id="cnv-cloning-local-volume-to-another-node_{context}"]
+= Cloning a local volume to another node
+
+You can move a virtual machine disk so that it runs on a specific node by cloning the underlying PersistentVolumeClaim (PVC).
+
+To ensure the virtual machine disk is cloned to the correct node you must either create a new PersistentVolume (PV) or identify one on the correct node.
```
ousleyp, line 10: s/node/node,

modules/cnv-cloning-local-volume-to-another-node.adoc
```diff
@@ -0,0 +1,131 @@
...
+Apply a unique label to the PV so that it can be referenced by the DataVolume.
+The DataVolume clones the underlying PVC to the PV on the correct node.
```
ousleyp, line 12: nit: I'm not sure this information needs to be included in the module intro.

modules/cnv-cloning-local-volume-to-another-node.adoc
```diff
@@ -0,0 +1,131 @@
...
+      nodeSelectorTerms:
+      - matchExpressions:
+        - key: kubernetes.io/hostname <1>
+          operator: In
+          values:
+          - node01 <2>
+...
+----
+<1> The `kubernetes.io/hostname` key uses the node hostname to select a node.
+<2> The hostname of the node.
```
ousleyp, line 84: hostname n
Use only as a variable. Otherwise, use “host name.”

ousleyp (commented):
Mostly LGTM! I made note of a couple nits, but otherwise it's good to go from my perspective.

PR #21208, opened 2020-04-15
Fix invalid admonition type in ShiftStack Kuryr limitations

### LINKED ISSUE #21207: Invalid admonition in 4.3 ShiftStack w/Kuryr limitations

https://docs.openshift.com/container-platform/4.3/installing/installing_openstack/installing-openstack-installer-kuryr.html#installation-osp-kuryr-known-limitations_installing-openstack-installer-kuryr

### What needs fixing?

The musl section uses "INFO" as an admonition, which is not a valid type. As a result, the admonition doesn't render properly.

maxwelldb:
Resolves #21207


maxwelldb:
@kalexand-rh Could you +1 this? It fixes a tiny issue that's introduced by #20728.

kalexand-rh (commented):
+1

PR #21210, opened 2020-04-16
Add information on Watchdog

rh-max:


modules/monitoring-alertmanager-default-configuration.adoc
```diff
@@ -28,4 +28,6 @@ receivers:
 
 {product-title} monitoring ships with the Watchdog alert that always triggers by default to ensure the availability of the monitoring infrastructure.
 
-// FIXME perhaps explain what individual entries in the listing mean
+Watchdog is a simple Prometheus alerting rule that always triggers. The Alertmanager continuously sends notifications for Watchdog to the notification provider that supports this functionality. The notification provider can be configured to issue alerts to the administrator if the monitoring stack fails. An example of a third-party notification provider is PagerDuty.
```
simonpasquier, line 31: "is always active" rather than "always triggers"?

"The Alertmanager continuously sends notifications for Watchdog to the notification provider that supports this functionality. The notification provider can be configured to issue alerts to the administrator if the monitoring stack fails." -> "Alertmanager continuously sends notifications for the Watchdog alert to the notification provider which is configured to notify the administrator when it stops receiving the Watchdog alert."

modules/monitoring-alertmanager-default-configuration.adoc
```diff
@@ -28,4 +28,6 @@ receivers:
 
 {product-title} monitoring ships with the Watchdog alert that always triggers by default to ensure the availability of the monitoring infrastructure.
```
simonpasquier, line 29: this sentence seems redundant with the new section below.

rh-max: ACK, I've merged the first sentence and the first sentence of the second paragraph.

modules/monitoring-alertmanager-default-configuration.adoc
rh-max: Thank you, @simonpasquier, good feedback.
* "is always active" might imply that it is enabled, but not firing. How about "fires continuously"?
* Your suggested sentences assume that the provider has been configured to alert the administrator. While it's reasonable, I don't think we can assume that, so let's go with "usually".

Could you take another look at the resultant changes? Thank you.

simonpasquier (commented):
:+1:

rh-max:
@simonpasquier Thank you!
@kalexand-rh Could you PTAL and merge + cherry-pick into `enterprise` `4.1` to `4.4`? Thank you.

kalexand-rh:
IDK what went wrong with the cherry-pick PR for 4.2 - it said that it wasn't merged, but it was: https://github.com/openshift/openshift-docs/commit/e23c9824bac22a18a58337c1a853f084c0a8a2bd

rh-max:
@kalexand-rh Thank you. Yeah, I can see it in `enterprise-4.2`. I've had this happen before, and Vikram saw it before too. It's not a problem.

kalexand-rh:
@rh-max, thanks! I hadn't seen this weirdness before. I'm just glad that the content moved over correctly.

PR #21216, opened 2020-04-16
CNV: Removed ssh keys workaround

apinnick:
CP for OCP 4.3 and 4.4

Removed procedure for editing imported VM yaml.


modules/cnv-importing-vmware-vm.adoc
```diff
@@ -24,6 +24,7 @@ If you try to import a virtual machine whose disk size is larger than the availa
 
 . In the {CNVProductName} web console, click *Workloads* -> *Virtual Machines*.
 . Click *Create Virtual Machine* and select *Import with Wizard*.
+
```
bergerhoffer, line 27: This isn't necessary between steps (when there's no text or code block following the step) but it's not hurting anything,

bergerhoffer (commented):
Just one comment, but nothing to hold up the PR.

PR #21232, opened 2020-04-16
updating 3.11 FIPS note

kalexand-rh:
@xiaojiey, PTAL?


kalexand-rh:

PR #20175, opened 2020-03-02
Sending a copy of the logs to an external log aggregator while keeping the logs local

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1792999

PR #21240, opened 2020-04-16
Removing note on @type:copy added to 4.3 and 4.4 in error

mburke5678:
Added note to logging docs in error. Functionality was removed in 4.3+. Might still exist in 4.1 and 4.2
https://github.com/openshift/openshift-docs/pull/20175

### DIFF
--- a/modules/cluster-logging-collector-fluentd.adoc
+++ b/modules/cluster-logging-collector-fluentd.adoc
@@ -7,11 +7,6 @@
 
 You can use the Fluentd *forward* plug-ins to send a copy of your logs to an external log aggregator, instead of the default Elasticsearch.
 
-[NOTE]
-====
-You can configure {product-title} to send logs to both the internal Elasticsearch instance and an outside log aggregator using the '@type copy' plug-in within the Fluentd `output-application` configuration file. Specific instructions for this configuration are beyond the scope of this documentation.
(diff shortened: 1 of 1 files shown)


mburke5678:
@anpingli PTAL

modules/cluster-logging-collector-fluentd.adoc
```diff
@@ -7,11 +7,6 @@
 
 You can use the Fluentd *forward* plug-ins to send a copy of your logs to an external log aggregator, instead of the default Elasticsearch.
```
anpingli, line 8: You can use the Fluentd *forward* plug-ins to send your logs to both the internal Elasticsearch instance and an outside log aggregator

modules/cluster-logging-collector-fluentd.adoc
```diff
@@ -7,11 +7,6 @@
...
 
-[NOTE]
-====
-You can configure {product-title} to send logs to both the internal Elasticsearch instance and an outside log aggregator using the '@type copy' plug-in within the Fluentd `output-application` configuration file. Specific instructions for this configuration are beyond the scope of this documentation.
```
anpingli, line 12: That is ok to drop these lines. In fact, the 'copy' plugin are still be used in background.

PR #21272, opened 2020-04-17
Remove SCTP from 4.4.0 release notes

jboxman:
@codyhoag, this feature won't land until 4.4.z; Should I comment this out or delete it?

Thanks!

### DIFF
--- a/release_notes/ocp-4-4-release-notes.adoc
+++ b/release_notes/ocp-4-4-release-notes.adoc
@@ -169,6 +169,7 @@ Developer Catalog.
 [id="ocp-4-4-networking"]
 === Networking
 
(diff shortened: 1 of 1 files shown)


codyhoag:
@jboxman it's fine how you have it here (commenting out) for now. I'll figure out the protocol for adding an item to RNs released in a z-stream release. We're a little unclear how to handle this at the moment, so it's safer to just keep it as a comment for now. Thanks!

vikram-redhat:
It is an unusual one in that we have a feature that **we know** will land in a z stream. Commenting it out is fine, although note that a promise to release at a future date is just that, a promise :)

PR #21311, opened 2020-04-20
Add clusterOSImage parameter to install config table

### LINKED ISSUE #21308: Explain install-config OSP clusterOSImage property

* https://docs.openshift.com/container-platform/4.3/installing/installing_openstack/installing-openstack-installer-custom.html#installation-configuration-parameters_installing-openstack-installer-custom

### What needs fixing?

Docs should note OS image override info: https://github.com/openshift/installer/blob/release-4.4/docs/user/openstack/customization.md#image-overrides

maxwelldb:
Resolves #21308

### DIFF
--- a/modules/installation-configuration-parameters.adoc
+++ b/modules/installation-configuration-parameters.adoc
@@ -331,7 +331,7 @@ ifdef::osp[]
 |`platform.openstack.clusterOSImage`
 |The location from which the installer downloads the {op-system} image.
 
-You must to set this parameter to perform an installation in a restricted network.
+You must set this parameter to perform an installation in a restricted network.
 |An HTTP or HTTPS URL, optionally with an SHA-256 checksum.
 
 For example, `\http://mirror.example.com/images/rhcos-43.81.201912131630.0-openstack.x86_64.qcow2.gz?sha256=ffebbd68e8a1f2a245ca19522c16c86f67f9ac8e4e0c1f0a812b068b16f7265d`.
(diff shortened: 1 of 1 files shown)


maxwelldb:
For your review @Fedosin

modules/installation-configuration-parameters.adoc
```diff
@@ -312,6 +326,16 @@ ifdef::osp[]
    }
 }
 ----
+
+|`platform.openstack.clusterOSImage`
+|The location from which the installer downloads the {op-system} image.
+
+You may need to set this parameter to perform an installation in a restricted network.
```
racedo, line 333: It saus "you may need it" but this is required. I'd suggest something like the following:

If not specified, the installer will download the image from a predetermined location described in https://github.com/openshift/installer/blob/master/data/data/rhcos.json 

This is required to perform an installation in a restricted networking environment.

@Fedosin FYI

maxwelldb: Is there a reason to specify where the installer fetches the image from by default? If not, I'd prefer to leave it out--space is at a premium in the param tables. 

Will strike "may."

maxwelldb:
Putting this on ice pending input from CS.

maxwelldb:
FYI @vikram-redhat This is not intended for 4.4 GA.

maxwelldb:
I'm incorporating this change into the in-progress restricted installation docs, e.g. #22478 

Closing.

PR #21317, opened 2020-04-20
Add SCTP feature to release notes

jboxman:
@codyhoag we're back on!


codyhoag:
Great! Much easier this way then adding in z-stream :sweat_smile:

PR #21336, opened 2020-04-21
[WIP] Adding more detail to load balancer setup documentation

etsauer:
The current load balancer documentation is difficult to distinguish between the two required load balancers, and some critical information is easy to miss. This is an attempt to improve that section.

### DIFF
--- a/modules/installation-network-user-infra.adoc
+++ b/modules/installation-network-user-infra.adoc
@@ -114,48 +114,99 @@ for platform containers and provide telemetry data to Red Hat.
 ====
 endif::[]
 
-.Load balancers
+[discrete]
+=== Load balancers
 
-Before you install {product-title}, you must provision two layer-4 load
-balancers. The API requires one load balancer and the default Ingress Controller
-needs the second load balancer to provide ingress to applications.
+Before you install {product-title}, you must provision two load balancers that meet the following requirements:
 
-[cols="2a,5a,2a,2a,2a",options="header"]
+. *API load balancer*: Provides a common endpoint for users, both human and machine, to interact with and configure the platform. Configure the following conditions:
++
+--
+  ** Layer 4 load balancing only. This can be referred to as Raw TCP, SSL Passthrough, or SSL Bridge mode.
+  ** A stateless load balancing algorithm. The options vary based on the load balancer implementation.
+--
++
+[NOTE]
+====
+Session persistence is not required for the API load balancer to function properly.
+====
++
+Configure the following ports on both the front and back of the load balancers:
++
+.API load balancer
+[cols="2,5,^2,^2,2",options="header"]
 |===
 
 |Port
-|Machines
+|Back-end machines (pool members)
 |Internal
 |External
 |Description
 
 |`6443`
 |Bootstrap and control plane. You remove the bootstrap machine from the load
-balancer after the bootstrap machine initializes the cluster control plane.
-|x
-|x
+balancer after the bootstrap machine initializes the cluster control plane. You
+must configure the `/readyz` endpoint for the API server health check probe.
+|X
+|X
 |Kubernetes API server
 
 |`22623`
 |Bootstrap and control plane. You remove the bootstrap machine from the load
 balancer after the bootstrap machine initializes the cluster control plane.
-|x
+|X
 |
 |Machine Config server
 
+|===
++
+[NOTE]
+====
+The load balancer should be configured to take a maximum of 30 seconds from the
+time the API server turns off the `/readyz` endpoint to the removal of the API
+server instance from the pool. Within the time frame after `/readyz` returns an
+error or becomes healthy, the endpoint must have been removed or added. Probing
+every 5 or 10 seconds, with two successful requests to become healthy and three
+to become unhealthy, are well-tested values.
+====
+
+. *Application Ingress load balancer*: Provides an Ingress point for application traffic flowing in from outside the cluster. Configure the following conditions:
++
+--
+  ** Layer 4 load balancing only. This can be referred to as Raw TCP, SSL Passthrough, or SSL Bridge mode.
+  ** A connection-based or session-based persistence is recommended, based on the options available and types of applications that will be hosted on the platform.
+--
++
+Configure the following ports on both the front and back of the load balancers:
++
+.Application Ingress load balancer
+[cols="2,5,^2,^2,2",options="header"]
+|===
+
+|Port
+|Back-end machines (pool members)
+|Internal
+|External
+|Description
+
 |`443`
 |The machines that run the Ingress router pods, compute, or worker, by default.
-|x
-|x
+|X
+|X
 |HTTPS traffic
 
 |`80`
 |The machines that run the Ingress router pods, compute, or worker by default.
-|x
-|x
+|X
+|X
 |HTTP traffic
 
 |===
+
+[TIP]
+====
+If the true IP address of the client can be seen by the load balancer, enabling source IP-based session persistence can improve performance for applications that use end-to-end TLS encryption.
+====
 
 [NOTE]
 ====


vikram-redhat:
@kalexand-rh can you move this forward with @etsauer?

etsauer:
Just as a note, I noticed the CI failed, but it looks to have been due to git clone issues, maybe just needs kicked off again?

etsauer:
@vikram-redhat @kalexand-rh bump.

vikram-redhat:
Thanks. @etsauer. GitHub has been playing up in the last few days. I restarted the build and it works ok now.

@kalexand-rh PTAL

etsauer:
@kalexand-rh bump.

etsauer:
@kalexand-rh @vikram-redhat bump.

etsauer:
@kalexand-rh @vikram-redhat any feedback on this change?

vikram-redhat:
@xltian Hey Xiaoli, can someone in QE help us ack this?

xltian:
@gpei @lihongan  to review from QE

xingxingxia:
Plus me, because the "API Load Balancer" part includes "control plane". The PR looks good from my side. For the wording "Layer 4 load balancing only" and "Round robin load balancing", @gpei and @lihongan , you can confirm these?

modules/installation-network-user-infra.adoc
```diff
@@ -125,6 +129,24 @@ balancer after the bootstrap machine initializes the cluster control plane.
 |
 |Machine Config server
 
+|===
+
+
+2. *Application Ingress Load Balancer*: Provides an ingress point for application traffic flowing in from outside of the cluster.
+  * Layer 4 load balancing only (this is sometimes referred to as Raw TCP, SSL Passthrough, or SSL Bridge mode)
+  * Round robin load balancing
```
lihongan, line 137: `Round robin` is not always true in various cloud platform, for example the default distribution mode for Azure Load Balancer is a five-tuple hash. The tuple is composed of the: 
```
Source IP
Source port
Destination IP
Destination port
Protocol type
```
see also https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-distribution-mode

gpei:
The bootstrap related in `Table 2. API Load Balancer` lgtm.

modules/installation-network-user-infra.adoc
lihongan: Anyway, Round robin is most used load balancing algorithm. And both `round robin` and `hash` can used for Application Ingress LB.

vikram-redhat: @etsauer - if you can address @lihongan comment, we can merge this.

@codyhoag

modules/installation-network-user-infra.adoc
```diff
@@ -96,17 +96,21 @@ network topology requirements.
...
+== Load balancers
 
-Before you install {product-title}, you must provision two layer-4 load
-balancers. The API requires one load balancer and the default Ingress Controller
-needs the second load balancer to provide ingress to applications.
+Before you install {product-title}, you must provision two load balancers. The requirements for these load balancers are as follows:
 
+1. *API Load Balancer*: Provides a common endpoint for users (both human and machine) to interact with and configure the platform.
+  * Layer 4 load balancing only (this is sometimes referred to as Raw TCP, SSL Passthrough, or SSL Bridge mode)
+  * Round robin load balancing
```
xingxingxia, line 105: @sttts , hi Stefan, could you double check above two lines whether looking good or needing update? I see https://github.com/openshift/machine-config-operator/pull/1670/files says "Google cloud load balancer is a L3LB that is special", not "Layer 4" here. Thanks!

sttts: I totally miss what is described https://github.com/openshift/installer/blob/master/docs/dev/kube-apiserver-health-check.md.

sttts: https://www.quora.com/What-is-the-difference-between-layer-3-and-layer-4-load-balancing-Why-is-layer-7-LB-used-inspite-of-its-drawbacks-of-being-a-bottleneck#:~:text=A%20layer%203%20load%2Dbalancer,and%20TCP%20or%20UDP%20ports.&text=Layer%207%20load%20balancer%20operates,actual%20content%20of%20each%20message.

Sounds like L4 is correct. But this is really a detail IMO. @squeed has more background.

modules/installation-network-user-infra.adoc
etsauer: @lihongan @vikram-redhat I'm not clear on what you are asking me to change here. Do you want me to just remove the `round robin` line since it doesn't apply across the board?

vikram-redhat: @lihongan did you need @etsauer to do any update here?

lihongan: Since Round robin is not required here, so it is better to remove this line. Thanks.

kalexand-rh:
@lihongan, @etsauer, I removed the round robin line and added some edits for style. Does everything look right?

lihongan:
LGTM. Thank you @kalexand-rh

sttts:
/hold

https://github.com/openshift/openshift-docs/pull/21336#discussion_r442761292 is still not addressed. The load balancers described in this doc don't use supported configurations. The customer will experience frequent outages from the API, e.g. on cert rotation events, configuration changes or upgrades. Please describe a supported configuration with proper `/readyz` wiring.

etsauer:
@sttts I'm unclear on what the ask is in that comment, considering that the line about Round Robin load balancing was removed. Are you saying we need to add content covering the configuration of apiserver healthchecks via `/readyz`? If so, I'm ok taking a stab at that, but none of that is covered in the current documentation either.

sttts:
> Are you saying we need to add content covering the configuration of apiserver healthchecks via /readyz? 

Yes. Customer clusters without proper `/readyz` use are not supported and to be expected to have outages at random times (during updates, but also independently of updates, e.g. when configs change or cert rotation happens).

codyhoag:
@sttts @etsauer do you think we can add a statement about `/readyz` wiring to the API load balancer port 6443 table section?:
```
You must configure the `/readyz` endpoint for the API server health check probe.
```
Our UPI config sections referencing this info should have the `/readyz` wiring configured in their examples (e.g., [GCP](https://docs.openshift.com/container-platform/4.4/installing/installing_gcp/installing-gcp-user-infra.html#installation-deployment-manager-dns_installing-gcp-user-infra) and [Azure](https://docs.openshift.com/container-platform/4.4/installing/installing_azure/installing-azure-user-infra.html#installation-arm-dns_installing-azure-user-infra)).

sttts:
@codyhoag yes. Something along these lines. We should also define the expected timing. We do 10s intervals, 3 failure/successes for some platforms, 5s for others afaik. 

@tkashem can you comment? Which are the timing we want in the future everywhere?

codyhoag:
After a second look, it appears several cloud provider UPI docs do not instruct to configure `/readyz` health checks. The docs currently follow what our UPI templates contain in the [openshift/installer](https://github.com/openshift/installer/tree/master/upi) repo. According to [BZ#1836017](https://bugzilla.redhat.com/show_bug.cgi?id=1836017), those UPI templates should be updated soon, which will trickle down to the docs. In the meantime, we'll call it out as I suggested above.

We'll wait for any further suggestions from @tkashem before proceeding.

tkashem:
> @codyhoag yes. Something along these lines. We should also define the expected timing. We do 10s intervals, 3 failure/successes for some platforms, 5s for others afaik.
> 
> @tkashem can you comment? Which are the timings we want in the future everywhere?

Hi @codyhoag,  it depends on the platform. In essence, we want the load balancer to take at most `30s` (from the moment the kube-apiserver turns off `/readyz`)  to take the kube-apiserver instance out of the pool. 

AWS has the following configuration:
Path: /readyz
Unhealthy threshold: 2
Interval: `10s`
Timeout: `10s`

If we imagine a sequential process doing the probes, then it will take at most `30s` to take an unhealthy instance out of the pool. (disclaimer: the AWS health check algorithm is distributed and involves consensus)

Some platforms enforce some constraints on the probe configuration as well, for example on Azure TCP probe has the following constraints: source: https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-custom-probe-overview
```
The minimum probe interval is 5 seconds and the minimum number of unhealthy responses is 2. The total duration of all intervals cannot exceed 120 seconds.
```
I think as long as the health check probe respect the `at most 30s` rule, it should be okay. We can use the AWS health check probe as an example.

codyhoag:
Thanks for that info, @tkashem. I added a note in ceea96c803c56c0967b02e1d1bb20cac04e68089 describing the 30 second health check probe expectation. Can you review?

The `/readyz` config should be outlined in the UPI docs that reference this information, so I think this note will flag its importance and then it'll be demonstrated later in the assembly.

codyhoag:
@sttts I think we've addressed everything. Can you give a final verification before we merge?

modules/installation-network-user-infra.adoc
```diff
@@ -114,24 +114,34 @@ for platform containers and provide telemetry data to Red Hat.
...
+|Back-end machines (pool members)
 |Internal
 |External
 |Description
 
 |`6443`
 |Bootstrap and control plane. You remove the bootstrap machine from the load
-balancer after the bootstrap machine initializes the cluster control plane.
+balancer after the bootstrap machine initializes the cluster control plane. You
+must configure the `/readyz` endpoint for the API server health check probe.
```
sttts, line 144: why not be clear about the timings?
```
You must configure the API server `/readyz` endpoint as the health check of the LB: within at most 30s after `/readyz` returns error or becomes healthy, the endpoint must be removed or added. Probing every 5 or 10 seconds, with 2 successful requests to become healthy and 3 to become unhealthy, are well-tested values.
```

codyhoag: @sttts I added more detail for timings in 5ec28c182871d4ecbff344c68daf7d9718867510. Can you verify?

sttts: Commented. 30s is really an upper bound. That's why I wrote "at most".

codyhoag: Since the first sentence in this note states the 30 second max upper bound, I generically refer to the time frame afterwards since it can vary.

codyhoag:
I've addressed all feedback here and pushed the changes. @sttts are we clear to remove the `do-not-merge` labels?

modules/installation-network-user-infra.adoc
```diff
@@ -114,24 +114,34 @@ for platform containers and provide telemetry data to Red Hat.
...
-Before you install {product-title}, you must provision two layer-4 load
-balancers. The API requires one load balancer and the default Ingress Controller
-needs the second load balancer to provide ingress to applications.
-
+[id="installation-network-user-infra-load-balancers_{context}"]
+=== Load balancers
+
+Before you install {product-title}, you must provision two load balancers that meet the following requirements:
+
+. *API Load Balancer*: Provides a common endpoint for users, both human and machine, to interact with and configure the platform. Configure one of the following strategies:
```
etsauer, line 122: @codyhoag the phrase `Configure one of the following strategies` suggests that the below bullets represent two mutually exclusive approaches. They don't. I think that what we are actually trying to state is that

1. In ALL cases, you need a Layer 4 (as opposed to Layer 7) load balancer
2. The proper Load Balancing algorithm to use will vary based on the type of load balancer used, whether it is on prem (e.g. F5, Netscaler) or cloud-based, however, session persistence is not required for the API load balancer to function properly, as there is no session persistence at play. (its important to note this because this may be different for the ingress LB, where it may be serving requests for back end applications that use session persistence).

codyhoag: @etsauer I've adjusted that wording in c0649a6302f3d1294b0dc06b1bb3d04aaa49d3d7. Can you confirm?

etsauer:
@codyhoag adjusted a bit more.

codyhoag:
I've squashed and performed a final wordsmithing on this text. @etsauer @sttts can you confirm the latest edition and that it's accurate for OCP 4.3+?

modules/installation-network-user-infra.adoc
```diff
@@ -143,6 +158,37 @@ balancer after the bootstrap machine initializes the cluster control plane.
 |
 |Machine Config server
 
+|===
++
+[NOTE]
+====
+The load balancer should be configured to take a maximum of 30 seconds from the
```
jboxman, line 165: Is it _should be_ or _must_?

modules/installation-network-user-infra.adoc
```diff
@@ -114,24 +114,39 @@ for platform containers and provide telemetry data to Red Hat.
...
++
+[NOTE]
+====
+Session persistence is not required for the API load balancer to function properly.
+====
++
+Configure the following ports on both the front and back of the load balancers:
++
+.API load balancer
 [cols="2a,5a,2a,2a,2a",options="header"]
```
jboxman, line 137: I don't know if we care, but it's possible to get the `x` column to center the X. Also, the `a` in each `cols` allows for full Asciidoc syntax, but as none is being used, it can probably be omitted from `cols`:

```
[cols="2,5,^2,^2,2",options="header"]
```

Works locally.

jboxman (approved):
@codyhoag few notes, LGTM

modules/installation-network-user-infra.adoc
codyhoag: This is a recommended time frame; so I think ***must*** would be too strong, since users could technically configure it to a different amount.

modules/installation-network-user-infra.adoc
sttts: looks good.

modules/installation-network-user-infra.adoc
sttts: It *must* be configured with a value <=30s. 20s is also fine. 40s is unsupportable.

codyhoag:
This has been confirmed by final SME via email.

PR #21340, opened 2020-04-21
Added details for configuration of HA and config field note

abrennan89:


modules/serving-HA-components.adoc
```diff
@@ -0,0 +1,46 @@
+// Module included in the following assemblies:
+//
+// * serverless/serverless-config-HA.adoc
+
+[id="serving-HA-components_{context}"]
+= Knative Serving HA components
+
+Active/passive high availability (HA) is now available by default for the `autoscaler-hpa`, `controller`, `activator` , `kourier-control`, and `kourier-gateway` components.
```
maschmid: kourier-gateway is actually active/active

modules/serving-HA-components.adoc
```diff
@@ -0,0 +1,46 @@
...
+. In the {product-title} web console, navigate to the *Catalog* → *OperatorHub* → *Installed Operators* page.
+. Select the *{ServerlessOperatorName}*.
+. Click *Knative Serving* in the list of *Provided APIs* for the {ServerlessOperatorName} to go to the *Knative Serving* tab.
++
+image::serverless-installed-operator.png[Installed Operators page]
+. Click on `knative-serving` in the *Knative Serving* tab.
+. Click on `YAML` in the *knative-serving* page.
++
+image::serving-YAML-HA.png[Knative Serving YAML]
+. In the `spec` section of the YAML, you will see the `high-availability` field configured with `2` replicas:
```
mgencur, line 32: Unfortunately. This image has so much text that it is not readable in the preview. Maybe we could cut off the right side (let's say from the middle of the image) and zoom in?

modules/serving-HA-components.adoc
```diff
@@ -0,0 +1,46 @@
+// Module included in the following assemblies:
+//
+// * serverless/serverless-config-HA.adoc
+
+[id="serving-HA-components_{context}"]
+= Knative Serving HA components
+
+Active/passive high availability (HA) is available by default for the `autoscaler-hpa`, `controller`, `activator` , `kourier-control`, and `kourier-gateway` components.
+
+You can disable HA for these components by changing the configuration of the `KnativeServing.spec.highAvailability` field.
```
mgencur, line 10: It would be good to put a heading here. Something like "Disabling High Availability". The whole section below is related to disabling HA but we only get to know that from this sentence on line 10

modules/serving-HA-components.adoc
abrennan89: Just removed Active/passive entirely to simplify.

modules/serving-HA-components.adoc
abrennan89: Tried removing the the outer part and just showing the YAML - we need to make sure the image still makes sense and isn't cutting off in a weird way. Unfortunately it's an issue with most images in OCP docs that the text can look very small.

modules/serving-HA-components.adoc
```diff
@@ -0,0 +1,49 @@
+// Module included in the following assemblies:
+//
+// * serverless/serverless-config-HA.adoc
+
+[id="serving-HA-components_{context}"]
+= Knative Serving high availability components
+
+High availability (HA) is available by default for the `autoscaler-hpa`, `controller`, `activator` , `kourier-control`, and `kourier-gateway` components.
```
abrennan89, line 8: High availability (HA) ->  HA
(already defined earlier in the assembly).
Will update alongside any additional QE or OCP doc review comments.

mgencur:
QE Approved

_topic_map.yml
```diff
@@ -1678,6 +1680,9 @@ Topics:
 ### Brokers
 ###  - Name: Using the broker
 ###    File: serverless-using-brokers
+### HA
+### - Name: Configuring High Availability components
```
bergerhoffer, line 1684: Unless there's something special w/ it in serverless, "high availability" doesn't need to be capitalized. Here and in the heading of the assembly.

modules/serving-HA-components.adoc
```diff
@@ -0,0 +1,49 @@
+// Module included in the following assemblies:
+//
+// * serverless/serverless-config-HA.adoc
+
+[id="serving-HA-components_{context}"]
+= Knative Serving high availability components
```
bergerhoffer, line 6: The headings/TOC seem a little off on this. I'd suggest instead having a concept module and then a procedure module, at the same level. Something like:

```
= About Knative Serving high availability
= Disabling high availability
```

modules/serving-HA-components.adoc
```diff
@@ -0,0 +1,49 @@
...
+Do not modify any YAML contained inside the `config` field. Some of the configuration values in this field are injected by the {ServerlessOperatorName}, and modifying them will cause your deployment to become unsupported.
+====
+
+.Prerequisites
+* An {product-title} account with cluster administrator access.
+* Installed the {ServerlessOperatorName} and Knative Serving.
+
+.Procedure
+
+. In the {product-title} web console, navigate to the *Catalog* → *OperatorHub* → *Installed Operators* page.
```
bergerhoffer, line 26: It might all be the same in the end, but suggest using `->` instead of the character you use for arrows. It's what we say to use in our guidelines, and good to follow just in case there's any difference in appearance.

modules/serving-HA-components.adoc
```diff
@@ -0,0 +1,49 @@
...
+. Select the *{ServerlessOperatorName}*.
+. Click *Knative Serving* in the list of *Provided APIs* for the {ServerlessOperatorName} to go to the *Knative Serving* tab.
++
+image::serverless-installed-operator.png[Installed Operators page]
+. Click on `knative-serving` in the *Knative Serving* tab.
+. Click on `YAML` in the *knative-serving* page.
```
bergerhoffer, line 32: Is this a tab in the GUI? If so, YAML should be in bold, not backticks

modules/serving-HA-components.adoc
```diff
@@ -0,0 +1,49 @@
...
+* Installed the {ServerlessOperatorName} and Knative Serving.
+
+.Procedure
+
+. In the {product-title} web console, navigate to the *Catalog* → *OperatorHub* → *Installed Operators* page.
+. Select the *{ServerlessOperatorName}*.
+. Click *Knative Serving* in the list of *Provided APIs* for the {ServerlessOperatorName} to go to the *Knative Serving* tab.
++
+image::serverless-installed-operator.png[Installed Operators page]
+. Click on `knative-serving` in the *Knative Serving* tab.
```
bergerhoffer, line 31: If it's a GUI item, knative-serving should be in bold, not backticks.

modules/serving-HA-components.adoc
bergerhoffer: I would just say to consider whether these screenshots are truly necessary or not. We don't use screenshots a lot and have plenty of GUI procedures in the docs.

bergerhoffer: Also, "you will see the `high-availability` field" -> "the `high-availability` field is configured "

modules/serving-HA-components.adoc
```diff
@@ -0,0 +1,49 @@
...
+image::serving-YAML-HA.png[Knative Serving YAML]
+. In the `spec` section of the YAML, you will see the `high-availability` field configured with `2` replicas:
++
+[source,yaml]
+----
+  high-availability:
+    replicas: 2
+----
++
+Change the value of `replicas` to `1`:
```
bergerhoffer, line 43: Consider making this it's own step since it's actionable. Or consider condensing the step into something like:

```
. In the `spec` section of the YAML, update the `high-availability.replicas` field to `1`:
+
[source,yaml]
----
  high-availability:
    replicas: 1
----
```

serverless/serverless-config-HA.adoc
```diff
@@ -0,0 +1,11 @@
+include::modules/serverless-document-attributes.adoc[]
```
bergerhoffer, line 1: This is an odd place to have the attribute file included. Though as I'm typing I see why, because you're using one in the title.

bergerhoffer (commented):
Just a few comments!

modules/serving-HA-components.adoc
abrennan89: Didn't even realise this was in the guidelines, that's so much better, thanks!

serverless/serverless-config-HA.adoc
abrennan89: Yeah, I used to have it in the normal place but it was breaking the titles :)

abrennan89:
Thanks, @bergerhoffer ! This should be ready for a merge now :slightly_smiling_face:

bergerhoffer:
Thanks for the updates, LGTM!

PR #21355, opened 2020-04-22
Add info about environment used for scale testing

chaitanyaenr:
This commit:
- Adds details about the environment including cloud platform used
  for running the performance and scale tests as well as info about
  the cluster configuration like instance types, CPU, memory, disk,
  node count e.t.c.
- Adds info on why we went with particular configuration as well as
  pointers to the docs wherever required.


ahardin-rh:
Thanks!

ahardin-rh:
Note: This PR is master only, as other PRs applied the changes to the appropriate branches

PR #21360, opened 2020-04-22
update sriov supported devices

zenghui-shi:


zshi-redhat:
@jboxman an update of supported sriov devices, PTAL

This will be backported to 4.4 and 4.3.

jboxman:
This needs special handling for 4.3.

PR #21388, opened 2020-04-22
Add ExternalIP configuration

jboxman:
- https://issues.redhat.com/browse/OSDOCS-536

This PR adds longstanding absent `ExternalIP` configuration. An update to the ingress configuration documentation that utilizes this functionality will follow under a separate cover.

### DIFF
--- a/modules/nw-externalip-about.adoc
+++ b/modules/nw-externalip-about.adoc
@@ -1,0 +1,179 @@
+// Module included in the following assemblies:
+//
+// * networking/configuring_ingress_cluster_traffic/configuring-externalip.adoc
+
+[id="nw-externalip-about_{context}"]
+= About ExternalIP
+
+For non-cloud environments, {product-title} supports the assignment of external IP addresses to a Service `spec.externalIPs` field through the *ExternalIP* facility.
+This exposes an additional virtual IP address, assigned to the Service, that can be outside the service network defined for the cluster.
+A Service configured with an external IP functions similarly to a Service with `type=NodePort`, allowing you to direct traffic to a local node for load balancing.
+
+You must configure your networking infrastructure to ensure that the external IP address blocks that you define are routed to the cluster.
+
+{product-title} extends the ExternalIP functionality in Kubernetes by adding the following capabilities:
+
+- Restrictions on the use of external IP addresses through a configurable policy
+- Allocation of an external IP address automatically to a Service upon request
+
+By default, only a user with `cluster-admin` privileges can create a Service with `spec.externalIPs[]` set to IP addresses defined within an external IP address block.
+
+[CAUTION]
+====
+Disabled by default, use of ExternalIP functionality can be a security risk, because in-cluster traffic to an external IP address is directed to that Service.
+This could allow cluster users to intercept sensitive traffic destined for external resources.
+====
+
+[IMPORTANT]
+====
+This feature is supported only in non-cloud deployments.
+For cloud deployments, use the load balancer services for automatic deployment of a cloud load balancer to target the endpoints of a service.
+====
+
+You can assign an external IP address in the following ways:
+
+Automatic assignment of an external IP::
+{product-title} automatically assigns an IP address from the `autoAssignCIDRs` CIDR block to the `spec.externalIPs[]` array when you create a Service with `spec.type=LoadBalancer` set.
+In this case, {product-title} implements a non-cloud version of the load balancer Service type and assigns IP addresses to the services.
+Automatic assignment is disabled by default and must be configured by a cluster administrator as described in the following section.
+
+Manual assignment of an external IP::
+{product-title} uses the IP addresses assigned to the `spec.externalIPs[]` array when you create a Service. You cannot specify an IP address that is already in use by another Service.
+
+[id="configuration-externalip_{context}"]
+== Configuration for ExternalIP
+
+Use of an external IP address in {product-title} is governed by the following fields in the `Network.config.openshift.io` CR named `cluster`:
+
+* `spec.externalIP.autoAssignCIDRs` defines an IP address block used by the load balancer when choosing an external IP address for the service. {product-title} supports only a single IP address block for automatic assignment. This can be simpler than having to manage the port space of a limited number of shared IP addresses when manually assigning ExternalIPs to services. If automatic assignment is enabled, a Service with `spec.type=LoadBalancer` is allocated an external IP address.
+* `spec.externalIP.policy` defines the permissible IP address blocks when manually specifying an IP address. {product-title} does not apply policy rules to IP address blocks defined by `spec.externalIP.autoAssignCIDRs`.
+
+If routed correctly, external traffic from the configured external IP address block can reach service endpoints through any TCP or UDP port that the service exposes.
+
+[CAUTION]
+====
+You must ensure that the IP address block you assign terminates at one or more nodes in your cluster.
+====
+
+{product-title} supports both the automatic and manual assignment of IP
+addresses, and each address is guaranteed to be assigned to a maximum of one
+service. This ensures that each service can expose its chosen ports regardless
+of the ports exposed by other services.
+
+[NOTE]
+====
+To use IP address blocks defined by `autoAssignCIDRs` in {product-title}, you must configure the necessary IP address assignment and routing for your host network.
+====
+
+The following YAML describes a Service with an external IP configured:
+
+.Example Service object with `spec.externalIPs[]` set
+[source,yaml]
+----
+apiVersion: v1
+kind: Service
+metadata:
+  name: http-service
+spec:
+  clusterIP: 172.30.163.110
+  externalIPs:
+  - 192.168.132.253
+  externalTrafficPolicy: Cluster
+  ports:
+  - name: highport
+    nodePort: 31903
+    port: 30102
+    protocol: TCP
+    targetPort: 30102
+  selector:
+    app: web
+  sessionAffinity: None
+  type: LoadBalancer
+status:
+  loadBalancer:
+    ingress:
+    - ip: 192.168.132.253
+----
+
+[id="restrictions-on-ip-assignment_{context}"]
+== Restrictions on the assignment of an external IP address
+
+As a cluster administrator, you can specify IP address blocks to allow and to reject.
+
+You configure IP address policy with a `policy` object defined by specifying the `spec.ExternalIP.policy` field.
+The policy object has the following shape:
+
+[source,json]
+----
+{
+  "policy": {
+    "allowedCIDRs": [],
+    "rejectedCIDRs": []
+  }
+}
+----
+
+When configuring policy restrictions, the following rules apply:
+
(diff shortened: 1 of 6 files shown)


jboxman:
Hi @squeed,

This PR is my first pass at adding documentation for ExternalIP in 4.x; Can you provide a sanity check? Initially I used the 3.x documentation as a guide, but the 4.x implementation seems quite different.

Thanks!

cc @danwinship

modules/nw-externalip-about.adoc
```diff
@@ -0,0 +1,184 @@
+// Module included in the following assemblies:
+//
+// * networking/configuring_ingress_cluster_traffic/configuring-externalip.adoc
+
+[id="nw-externalip-about_{context}"]
+= About external IP addresses
+
+You can allow external traffic into the cluster by defining an *external IP* configuration.
```
squeed, line 8: A better wording might be something like:

One way to make Kubernetes Services reachable by external clients is with the use of External IPs. External IPs are an additional virtual IP assigned to a Service that does not need to be inside the cluster's ServiceCIDR. External IPs are, generally, most useful for clusters in a bare-metal environment.

ExternalIPs function similarly to NodePort services, where traffic directed to a local node will be accepted and load-balanced. However, the administrator needs to configure their routing environment to ensure ExternalIPs are routed to the OpenShift cluster.

modules/nw-externalip-about.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+The configuration for external IP address blocks is defined by a Network custom resource (CR) named `cluster`.
```
squeed, line 10: OpenShift builds some additional functionality on top of Kubernetes' ExternalIPs.

* Administrators can restrict the use of ExternalIPs (ExternalIPPolicy)
* Services can request an ExternalIP be automatically allocated to them. ( AutoAssignExternalIPs)

modules/nw-externalip-about.adoc
```diff
@@ -0,0 +1,184 @@
...
+The Network custom resource (CR) is defined by the `Network.config.openshift.io` CustomResourceDefinition.
+
+The object is created during cluster installation by the Cluster Version Operator (CVO).
+
+// If it is a cloud deployment, what happens?
```
squeed, line 15: to answer your question, you can set up external IPs on cloud deployments, but it's not useful.

modules/nw-externalip-about.adoc
```diff
@@ -0,0 +1,184 @@
...
+This feature is only supported in non-cloud deployments.
+For cloud deployments, use the load balancer services for automatic deployment of a cloud load balancer to target the service's endpoints.
+====
+
+Use of external IP in {product-title} is governed by the following parameters:
+
+// What happens to existing services if this is changed to another IP block?
+// What happens if policy rules change an address becomes rejected?
+
+* `spec.externalIP.autoAssignCIDRs` defines an IP address block used by the load balancer when choosing an external IP address for the service. {product-title} supports only a single IP address block for automatic assignment.
```
squeed, line 28: If automatic assignment is enabled, then a Service of type LoadBalancer will automatically have an ExternalIP allocated for it.

This should not be used on cloud deployments.

modules/nw-externalip-about.adoc
squeed: Also, somewhere add:

ExternalIPs can be a security risk, because in-cluster traffic to an ExternalIP are directed to that service. This could allow cluster users to intercept sensitive traffic destined for external resources. Therefore, OpenShift, by default, disables use of ExternalIPs. See below for how to enable limited ExternalIPs.

modules/nw-externalip-about.adoc
squeed: By default, only administrators can create services with an ExternalIP set. A policy must be set to allow non-administrator use of the feature.

modules/nw-externalip-about.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+// Do we offer this in 4.x?
+// Also, these addresses can be used as virtual IPs (VIPs) when configuring xref../admin_guide/high_availability.adoc#configuring-ip-failover[high availability].
+
+{product-title} supports both the automatic and manual assignment of IP
+addresses, and each address is guaranteed to be assigned to a maximum of one
+service. This ensures that each service can expose its chosen ports regardless
+of the ports exposed by other services.
+
+For manually-configured external IPs, potential port clashes are handled on a first-come, first-served basis. If you request a port, it is only available if it has not yet been assigned for that IP address. For example:
```
squeed, line 47: Not sure what you mean by port collisions. We also have to worry about IP collisions if externalips are manually allocated.

squeed:
Added some comments; we don't have great messaging in 3.x about how to expose ExternalIPs.

IPFailover might need some expansion...

jboxman:
@squeed thank you!

modules/nw-externalip-about.adoc
jboxman: @squeed, does an External IP need not (but can) be in the ServiceCIDR, or is there a collision risk if it is? Can IP addresses automatically assigned from the ServiceCIDR co-exist with an External IP range?

Thanks.

zhaozhanqi:
@weliang1  could you help double confirm this? thanks

jboxman:
@zhaozhanqi @weliang1 thanks! I came up with this based on `oc explain network.spec.ExternalIP` and local CRC, so it's a guess. This only covers ExternalIP; I'll update the [ingress docs](https://docs.openshift.com/container-platform/4.5/networking/configuring_ingress_cluster_traffic/configuring-ingress-cluster-traffic-service-external-ip.html) (which seem wrong) that reference an external IP address separately.

modules/nw-externalip-about.adoc
```diff
@@ -0,0 +1,189 @@
...
+[id="automatic-external-ip_assignment_{context}"]
+== Automatic assignment of an external IP
+
+{product-title} automatically assigns an IP address from the `autoAssignCIDRs` CIDR range to the `spec.externalIPs[]` array when you create a Service CR object with `spec.type=LoadBalancer`.
+In this case, {product-title} implements a non-cloud version of the load balancer service type and assigns IP addresses to the services.
+
+[id="manual-external-ip_assignment_{context}"]
+== Manual assignment of an external IP
+
+{product-title} uses the IP addresses assigned to the `spec.externalIPs[]` array when you create a Service CR object with `spec.type=LoadBalancer`.
```
weliang1, line 113: I do not think Manual assignment of an external IP need a Service CR object with `spec.type=LoadBalancer`.

modules/nw-externalip-object.adoc
```diff
@@ -0,0 +1,90 @@
...
+  ...
+  externalIP:
+    autoAssignCIDRs:
+    - 192.168.132.254/29
+----
+
+The following example configures policy rules for allowed and rejected CIDR ranges.
+Any IP address that exists in both the allowed and rejected ranges is rejected.
+
+.Example configuration with `spec.externalIP.autoAssignCIDRs` set
```
weliang1, line 74: Line 57 is for spec.externalIP.autoAssignCIDRs set.
This line should be 'spec.externalIP.policy.allowedCIDRs and spec.externaIP.policy.rejectedCIDRs' set

jboxman: @weliang1, oops, thanks!

jboxman:
Hi @danwinship,

When you've a moment, can you review this draft for ExternalIP?

Thanks.

jboxman:
Hi @squeed,

Can you review the [latest version](http://file.bos.redhat.com/jboxman/OSDOCS-536/networking/configuring_ingress_cluster_traffic/configuring-externalip.html) (preview) of this and let me know what you think?

Thanks!

modules/nw-externalip-about.adoc
```diff
@@ -0,0 +1,178 @@
...
+[IMPORTANT]
+====
+This feature is supported only in non-cloud deployments.
+For cloud deployments, use the load balancer services for automatic deployment of a cloud load balancer to target the endpoints of a service.
+====
+
+You can assign an external IP address in the following ways:
+
+Automatic assignment of an external IP::
+{product-title} automatically assigns an IP address from the `autoAssignCIDRs` CIDR block to the `spec.externalIPs[]` array when you create a Service CR object.
```
squeed, line 36: Three corrections
1 - add a reference below to how administrators can set `autoAssignCIDRs` - those aren't on the Service object
2 - Service isn't a custom resource (CR).
3 - This only happens when you create a service with `spec.type = LoadBalancer`. Otherwise, we won't automatically allocate an IP.

modules/nw-externalip-about.adoc
```diff
@@ -0,0 +1,178 @@
...
+In this case, {product-title} implements a non-cloud version of the load balancer Service type and assigns IP addresses to the services.
+
+Manual assignment of an external IP::
+{product-title} uses the IP addresses assigned to the `spec.externalIPs[]` array when you create a Service CR. You cannot specify an IP address that you have already used in another Service.
+
+[id="configuration-externalip_{context}"]
+== Configuration for ExternalIP
+
+Use of an external IP address in {product-title} is governed by the following fields:
```
squeed, line 45: I would call out that this is in the `config.openshift.io/v1/Network` object.

modules/nw-externalip-about.adoc
```diff
@@ -0,0 +1,178 @@
...
+  sessionAffinity: None
+  type: LoadBalancer
+status:
+  loadBalancer:
+    ingress:
+    - ip: 192.168.132.253
+----
+
+[id="restrictions-on-ip-assignment_{context}"]
+== Restrictions on IP address assignment
```
squeed, line 98: Probably worth noting that this is external IP address assignment, unless it's clear from the context.

modules/nw-externalip-about.adoc
jboxman: @squeed,

1. We have a weird restriction where we can't link internally to other sections, so instead I added:
   > Automatic assignment is disabled by default and must be configured by a cluster administrator as described in the following section.

   Or did you mean something else?

1. Oops, thanks. That's a legacy from when I didn't know anything about k8s extension mechanisms.

1. Oops, for some reason I removed that language; I've added it back.

modules/nw-externalip-about.adoc
jboxman: @squeed, oh, wow, somehow I never mention that important detail. How about

> Use of an external IP address in {product-title} is governed by the following fields in the `Network.config.openshift.io` CR named `cluster`:

modules/nw-externalip-about.adoc
jboxman: Done!

jboxman:
@openshift/team-documentation 

- [latest preview](http://file.bos.redhat.com/jboxman/OSDOCS-536/networking/configuring_ingress_cluster_traffic/configuring-externalip.html)

codyhoag (commented):
This is really well done. Just a couple minor suggestions. Otherwise, LGTM!

jboxman:
@codyhoag, thanks, it's been six months!

PR #21391, opened 2020-04-22
Suggested changes to the log forwarding section

mburke5678:
https://docs.google.com/document/d/1abKabUALXa2O6-cbFu8iGY7uo5vcLR3ago9JyQC7xjE/edit?ts=5e9ee5b7#


mburke5678:
@sichvoge PTAL

modules/cluster-logging-collector-log-forward.adoc
```diff
@@ -0,0 +1,34 @@
+// Module included in the following assemblies:
+//
+// * logging/cluster-logging-external.adoc
+
+[id="cluster-logging-collector-log-forward_{context}"]
+= Forwarding cluster logs using the Log Forwarding API
+
+The cluster logging Log Forwarding API enables administrators to configure custom pipelines to send your container and node logs to specific endpoints within or outside of your cluster. You can send logs by type to the internal {product-title} Elasticsearch instance and to remote destinations not managed by {product-title} cluster logging, such as your existing logging service, an external Elasticsearch cluster, external log aggregation solutions, or a Security Information and Event Management (SIEM) system.
```
sichvoge, line 8: How about changing this to "The Log Forwarding API" only?

modules/cluster-logging-collector-log-forward.adoc
```diff
@@ -0,0 +1,34 @@
+// Module included in the following assemblies:
+//
+// * logging/cluster-logging-external.adoc
+
+[id="cluster-logging-collector-log-forward_{context}"]
+= Forwarding cluster logs using the Log Forwarding API
```
sichvoge, line 6: We should remove "cluster" to be consistent with the other sub section.

sichvoge (changes requested):
Two small changes but otherwise LGTM

mburke5678:
@jcantrill @alanconway PTAL
@sichvoge proposed merging the two log forwarding topics in hte docs into one topic, plus other suggestions in a Google doc. [1] 
Please take a look at the preview [2] and let us know what you think.
[1] https://suggested-changes-to-the-log-forwarding-section--ocpdocs.netlify.app/openshift-enterprise/latest/logging/config/cluster-logging-external.html
[2] https://suggested-changes-to-the-log-forwarding-section--ocpdocs.netlify.app/openshift-enterprise/latest/logging/config/cluster-logging-external.html

mburke5678:
@jcantrill @alanconway PTAL

mburke5678:
Preview: https://suggested-changes-to-the-log-forwarding-section--ocpdocs.netlify.app/openshift-enterprise/latest/logging/config/cluster-logging-external.html

sichvoge:
@mburke5678 based on some recent discussions, we decided to not restrict forwarding using Fluentd's forward protocol to Fluentd only. It seems that there are systems that accept data using that protocol as well and we should handle it like the syslog protocol.

Obviously, it is important to note that like the syslog protocol, we only support the sending side not the receiver.

Therefore, can we make the appropriate changes so it looks similar to syslog and remove any text that restricts only to fluentd. For example: _s/Forwarding logs to another Fluentd instance/Forwarding logs using the out_forward protocol_

alanconway:
> @mburke5678 based on some recent discussions, we decided to not restrict forwarding using Fluentd's forward protocol to Fluentd only. It seems that there are systems that accept data using that protocol as well and we should handle it like the syslog protocol.

Makes sense. The fluent forward protocol is supported by several log stores and log processing tools. 

> Therefore, can we make the appropriate changes so it looks similar to syslog and remove any text that restricts only to fluentd. For example: _s/Forwarding logs to another Fluentd instance/Forwarding logs using the out_forward protocol_

I would suggest "Forwarding logs using the Fluent forward protocol"

Then name of the protocol is "forward", but you need to prefix with "Fluent" as it's a pretty vague name otherwise. The names "in_forward" and "out_forward" refer to the fluentd input/output *plugins* that speak the receiving/sending ends of the "forward" protocol.

mburke5678:
@sichvoge @alanconway PTAL: http://file.rdu.redhat.com/~mburke/Suggested-changes-to-the-log-forwarding-section/logging/config/cluster-logging-external.html#cluster-logging-collector-log-forward_cluster-logging-external

sichvoge:
PM: ack :)

jcantrill (approved):
minor edit that @alanconway introduced https://github.com/openshift/cluster-logging-operator/pull/532 for GA

PR #21395, opened 2020-04-22
Expand explanation of SCTP capabilities

jboxman:
Explain a use case for SCTP.

### DIFF
--- a/modules/nw-sctp-about.adoc
+++ b/modules/nw-sctp-about.adoc
@@ -8,7 +8,16 @@
 As a cluster administrator, you can enable SCTP on the hosts in the cluster.
 On {op-system-first}, the SCTP module is disabled by default.
 
-SCTP is a reliable message based protocol that runs on top of an IP network.
+SCTP is a reliable message based protocol that runs on top of an IP network. The protocol is commonly used in the telecommunications industry. It differs from TCP in several ways, including the following:
+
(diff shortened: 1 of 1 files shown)


vikram-redhat:
@jboxman what is this waiting on?

jboxman:
@vikram-redhat, Monday morning probably; it's a not entirely satisfying attempt at specifying a coherent use case. But it may be off the mark. I haven't finished my next attempt yet. And someone that knows better will then need to call my buff.

It's not urgent for 4.4, it's just a little extra.

I can change the label or push :delete it or whatever until it's complete, if necessary.

vikram-redhat:
@jboxman that's ok if it can come post 4.4. I wanted to close up all 4.4 necessary work for l10n.

jboxman:
Hi @danwinship, I felt the urge to expand upon the description. Is what I wrote technically sound?

Thanks!

danwinship:
The use case for SCTP is "my OpenShift cluster needs to interoperate with my other systems which are already using SCTP". If the customer is not already using SCTP outside of OpenShift, there is no reason for them to use it inside OpenShift. None of the nominal benefits of SCTP over TCP are relevant at this point because SCTP is a dead and abandoned protocol that nobody supports (except in telecom where they're stuck with it for compatibility reasons). (Also, things like multihoming wouldn't work inside an OpenShift cluster anyway.) We should be actively discouraging customers from using SCTP. Every time a RHEL customer enables SCTP, God kills a kitten. Think of the kittens. :joy_cat:

jboxman:
@danwinship, I'll close this PR, then. Thanks!

PR #20769, opened 2020-03-30
RHV-16816 RHV install docs

rolfedh:

PR #21402, opened 2020-04-22
RHV-16816 squashing master 4-13 06:10

rolfedh:


rolfedh:
See https://github.com/openshift/openshift-docs/pull/20769 instead.

PR #21447, opened 2020-04-23
removing product-title when referring to an admin

lamek:


jboxman:
@lamek as expected, many are mine!

jboxman:
@lamek LGTM

jboxman:
Yeah, I was just thinking about that; It's going to need all manual cherry picks; I have my own style fix that will cause the same result that I've been putting off for that reason.

PR #21475, opened 2020-04-24
Added tags example for Azure MachineSet yaml

fmartingr:
I've added the `tags` key to the yaml example for the document "[Creating a MachineSet in Azure](https://docs.openshift.com/container-platform/4.3/machine_management/creating_machinesets/creating-machineset-azure.html#machineset-yaml-azure_creating-machineset-azure)", since it was supported and missing.

### DIFF
--- a/modules/machineset-yaml-azure.adoc
+++ b/modules/machineset-yaml-azure.adoc
@@ -75,8 +75,8 @@ spec:
           sshPrivateKey: ""
           sshPublicKey: ""
           tags:
-            tag1: value1
-            tag2: value2
+            - name: kubernetes.io/cluster/<infrastructureID> <1>
+              value: owned
           subnet: <infrastructureID>-<role>-subnet <1> <2>
           userDataSecret:
             name: <role>-user-data <2>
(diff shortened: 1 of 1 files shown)


vikram-redhat:
@jianlinliu is this ok?

@kalexand-rh

jianlinliu:
@jhou1 @sunzhaohua2 ^^

kalexand-rh:
Adding the example is fine, but please also provide context about how these tags are meant to be used.

jhou1:
@miyadav PTAL

miyadav:
@fmartingr , would suggest to add more info about the field values value1 and value2 , so that it would help users to make use of those values in future and give better idea of its usefulness .
May be similar to what is mentioned in example of [AWS](https://docs.openshift.com/container-platform/4.3/machine_management/creating_machinesets/creating-machineset-aws.html) - 
 tags:
            - name: kubernetes.io/cluster/<infrastructureID> 
              value: owned

fmartingr:
@miyadav I've updated the example to match the one in the AWS procedure, sorry about the delay.

PR #21493, opened 2020-04-24
osdocs-948 updating AWS permissions in installer

kalexand-rh:
https://issues.redhat.com/browse/OSDOCS-948

@jianlinliu, will you PTAL?

### DIFF
--- a/modules/installation-aws-permissions.adoc
+++ b/modules/installation-aws-permissions.adoc
@@ -18,6 +18,7 @@ cluster, the IAM user requires the following permissions:
 * `ec2:AuthorizeSecurityGroupIngress`
 * `ec2:CopyImage`
 * `ec2:CreateNetworkInterface`
+* `ec2:AttachNetworkInterface`
 * `ec2:CreateSecurityGroup`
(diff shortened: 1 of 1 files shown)


modules/installation-aws-permissions.adoc
```diff
@@ -71,7 +71,6 @@ cluster, the IAM user requires the following permissions:
 * `ec2:CreateVpc`
 * `ec2:CreateVpcEndpoint`
 * `ec2:ModifySubnetAttribute`
-* `ec2:ModifyVpcAttribute`
```
jianlinliu, line 74: Why we remove this line here? Per install source code - https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/aws/permissions.go, we still need this permission.

kalexand-rh: Thank you @jianlinliu. I've fixed this error. Will you please take another look?

kalexand-rh:
@jianlinliu, when you have a moment, will you please +1 this change?

jianlinliu:
LGTM.

PR #21508, opened 2020-04-24
Add note about link-local IP addresses

jboxman:
- https://bugzilla.redhat.com/show_bug.cgi?id=1718389

### DIFF
--- a/networking/understanding-networking.adoc
+++ b/networking/understanding-networking.adoc
@@ -12,4 +12,13 @@ IP address means that Pods can be treated like physical hosts or virtual
 machines in terms of port allocation, networking, naming, service discovery,
 load balancing, application configuration, and migration.
 
+[NOTE]
+====
+Some cloud platforms offer metadata APIs that listen on the 169.254.169.254 IP address, a link-local IP address in the IPv4 `169.254.0.0/16` CIDR block.
+
+This CIDR block is not reachable from the pod network. Pods that need access to these IP addresses must be given host network access by setting the `spec.hostNetwork` field in the Pod spec to `true`.
+
+If you allow a Pod host network access, you grant the Pod privileged access to the underlying network infrastructure.
+====
+
 include::modules/nw-ne-openshift-dns.adoc[leveloffset=+1]


networking/understanding-networking.adoc
```diff
@@ -12,4 +12,9 @@ IP address means that Pods can be treated like physical hosts or virtual
 machines in terms of port allocation, networking, naming, service discovery,
 load balancing, application configuration, and migration.
 
+[NOTE]
+====
+Use of link-local IP addresses, which are assigned from the `169.254.0.0/16` CIDR block, are supported in {product-title} only for Pods with the `spec.hostNetwork` parameter set to `true`.
```
bobfuru, line 17: I'm a bit unclear as to the use of "only for Pods" here. I take it to mean that OCP only supports link-local IP addresses when a Pod uses the `spec.hostNetwork` parameter set to `true`. The sentence is also a bit long and hard to follow all the relationships. Suggest breaking into two sentences, maybe something like:

> Link-local IP addresses are assigned from the `169.254.0.0/16` CIDR block. OCP only supports link-local IP addresses that have the `spec.hostNetwork` parameter set to `true`.

^^ Not sure if that's the correct meaning behind this, so finding a way to clarify what the user needs to do with the parameter could be helpful.

bobfuru (commented):
Suggestion to clarify the note.

networking/understanding-networking.adoc
jboxman: @bobfuru, `spec.hostNetwork` is a field for the Pod resource. The CIDR block is unreachable by a pod unless that field is set to `true`. That is what I'm trying, with limited success, to explain.

bobfuru: That's the message I was getting, @jboxman. It could be helpful to state that more clearly up front, perhaps: 

`For the CIDR block to be reachable by a Pod, the `spec.hostNetwork" field must be set to `true`.

And then fill in the rest? Definitely tricky to shoehorn all the details here.

jboxman:
@bobfuru, updated; Maybe this take instead?

bobfuru:
> @bobfuru, updated; Maybe this take instead?

Yes, that's much better!

networking/understanding-networking.adoc
```diff
@@ -12,4 +12,10 @@ IP address means that Pods can be treated like physical hosts or virtual
 machines in terms of port allocation, networking, naming, service discovery,
 load balancing, application configuration, and migration.
 
+[NOTE]
+====
+Link-local IP addresses are assigned from the `169.254.0.0/16` CIDR block.
+For the CIDR block to be reachable by a Pod, the `spec.hostNetwork` field for the Pod CR must be set to `true`.
```
danwinship, line 18: I would say something like:

> Some cloud platforms provide APIs on addresses in the IPv4 link-local CIDR block (`169.254.0.0/16`). This CIDR block is not reachable from the pod network; Pods that need access to these IPs will need to be given host network access by setting the `spec.hostNetwork` field in the Pod spec to `true`.

Especially, it should be clearer that setting `spec.hostNetwork` is granting additional privileges to the pod (the wording I suggested may not be the best way to do that).

Also, "the Pod CR" is wrong. "CR" = "Custom Resource", which applies only types that are defined by Custom Resource Definitions. Built-in types like Pod are just "resources", not "custom resources".

juanluisvaladas:
Technically there is nothing wrong with the statement but I think it would be good to add a clarification of why this is important.

Something like:
Link-local IP addresses are assigned from the `169.254.0.0/16` CIDR block. These addresses are often used to reach the cloud provider's metadata service such as AWS, OpenStack or azure.

Also what Dan said.

networking/understanding-networking.adoc
jboxman: @danwinship, thanks. I've updated the text; How does it look now?

networking/understanding-networking.adoc
```diff
@@ -12,4 +12,13 @@ IP address means that Pods can be treated like physical hosts or virtual
 machines in terms of port allocation, networking, naming, service discovery,
 load balancing, application configuration, and migration.
 
+[NOTE]
+====
+Link-local IP addresses are assigned from the IPv4 `169.254.0.0/16` CIDR block. Some cloud platforms provide APIs reachable from only the link-local CIDR block.
```
juanluisvaladas, line 17: > Some cloud platforms provide APIs reachable from only the link-local CIDR block.

I understand from the word "from" in this statement that it's the cloud provider who limits the pod ability to reach the metadata service and it's not, the SDN limits it. Instead you could say that some cloud providers have APIs that listen on the link-local CIDRblock. That sounds more adequate to me

English isn't my first language, so feel free to disregard this comment if you don't agree with my interpretation.

jboxman: @juanluisvaladas you're absolutely right, and that's why trying to write this has been a challenge, beyond also not being an expert in the subject.

I'll try again. It's definitely important that the limitation is in the SDN.

danwinship, line 17: The first sentence makes it sound like someone is assigning IPs from that block to pods (or nodes), but that's not accurate. OCP itself does not make any use of link-local IPs. And the problem isn't that the cloud only allows access _from_ link-local IPs, it's that the cloud API is provided _on_ a link-local IP, and OCP doesn't allow access _to_ that IP from the pod network.

The top-level story here is "OCP blocks access to cloud metadata APIs, but you can use hostNetwork to get around that". The fact that the IPv4 link-local network block is involved is not really that important beyond the fact that if someone searches the docs for "IPv4 link local" we want them to find this note. Also, probably, if they search for "metadata api" or "169.254.169.254" because those are the other things they might search for when they're trying to figure out why their pod can't access the API it's trying to access.

danwinship:
yup lgtm

ahardin-rh:
LGTM!

jboxman:
@danwinship @juanluisvaladas thanks!

@ahardin-rh thanks!

PR #21510, opened 2020-04-24
Update the metric names for OVN-Kubernetes

jboxman:
These actually include the namespace and subsystem in the wild.


jboxman:
Hi @zhaozhanqi @huiran0826 

Can you confirm that the metrics names are correct? This is how they appear for me on the web console metrics page when using a cluster installed with the _OVN-Kubernetes_ network provider.

Thanks!

huiran0826:
LGTM, add @anuragthehatter , could you help double check? thanks!

anuragthehatter:
That looks right. Thanks @jboxman 

/LGTM

PR #21531, opened 2020-04-25
Remaining bug fix summaries

codyhoag:
@vikram-redhat this adds the remaining bug fix summaries. They are now all added. :tada:


vikram-redhat:
Thanks @codyhoag. I will ping the l10n team to see if they can squeeze these in.

PR #21540, opened 2020-04-26
Removed tag causing verification step to appear twice

apinnick:
Removed tag that caused verification step to appear twice in CAM operator installation on 4.x target cluster
CP for all 4.x versions


bobfuru:
LGTM

PR #21522, opened 2020-04-25
updates to oke

vikram-redhat:
@lbarbeevargas - I apologize. I realize now why there were issues with images. oce_about was created in 3.x docs, which was previously on non-modular docs format. This new page needed to be in the modular docs format in the 4.x release and needed symbolic links to images and modules folders to keep it compatible. 

This PR fixes this.

PR #21544, opened 2020-04-27
link redirects for oke and main page

vikram-redhat:
follow up to: https://github.com/openshift/openshift-docs/pull/21522


vikram-redhat:
No CP required

PR #21554, opened 2020-04-27
Added a note to installation modules.

yhontyk:
Branch 4.4

### DIFF
--- a/cli_reference/openshift_developer_cli/installing-odo.adoc
+++ b/cli_reference/openshift_developer_cli/installing-odo.adoc
@@ -5,13 +5,14 @@ include::modules/common-attributes.adoc[]
 :context: installing-odo
 toc::[]
 
-The following section describes how to install `{odo-title}` on different platforms.
+The following section describes how to install `{odo-title}` on different platforms using the CLI.
 
 [NOTE]
 ====
 Currently, `{odo-title}` does not support installation in a restricted network environment.
 ====
 
+You can also find the URL to the latest binaries from the {product-title} web console by clicking the *?* icon in the upper-right corner and selecting *Command Line Tools*
 
 include::modules/openshift-developer-cli-installing-odo-on-linux.adoc[leveloffset=+1]
 
--- a/modules/helm-installing-helm.adoc
+++ b/modules/helm-installing-helm.adoc
@@ -5,6 +5,10 @@
 [id="installing-helm_{context}"]
 = Installing Helm
 
+The following section describes how to install Helm on different platforms using the CLI.
+
+You can also find the URL to the latest binaries from the {product-title} web console by clicking the *?* icon in the upper-right corner and selecting *Command Line Tools*
+
 .Prerequisites
 * You have installed Go, version 1.13 or higher.
 


cli_reference/openshift_developer_cli/installing-odo.adoc
```diff
@@ -7,12 +7,13 @@ toc::[]
 
 The following section describes how to install `{odo-title}` on different platforms.
 
+Currently, `{odo-title}` does not support installation in a restricted network environment.
+
 [NOTE]
 ====
-Currently, `{odo-title}` does not support installation in a restricted network environment.
+You can always find the URL to the latest binaries in the web console of your cluster. To access them using the web console; click the *?* icon in the upper-right corner of the window, and select *Command Line Tools* in the drop-down menu.
```
Preeticp, line 14: s/always/also

cli_reference/openshift_developer_cli/installing-odo.adoc
```diff
@@ -7,12 +7,13 @@ toc::[]
 
 The following section describes how to install `{odo-title}` on different platforms.
```
Preeticp, line 8: Suggest: The following...different platforms using the CLI.

cli_reference/openshift_developer_cli/installing-odo.adoc
Preeticp: Suggest same change in the Helm module too.

modules/helm-installing-helm.adoc
```diff
@@ -8,6 +8,11 @@
 .Prerequisites
```
Preeticp, line 8: Suggest adding an intro line here similar to that in odo: The following section describes how to install Helm on different platforms using the CLI.

Suggest adding the note before the prerequisite section, immediately after the above intro line.

cli_reference/openshift_developer_cli/installing-odo.adoc
Preeticp: I think the placement is good. The other option is to add it to the end of the section. But I think the current option tells users that this section has CLI instructions, but if he wants to use the console he can access it before reading all of the CLI instructions.

cli_reference/openshift_developer_cli/installing-odo.adoc
```diff
@@ -5,14 +5,15 @@ include::modules/common-attributes.adoc[]
 :context: installing-odo
 toc::[]
 
-The following section describes how to install `{odo-title}` on different platforms.
+The following section describes how to install `{odo-title}` on different platforms using the CLI.
+
+Currently, `{odo-title}` does not support installation in a restricted network environment.
```
bergerhoffer, line 10: It's up to you, but to me, it feels like this warning about not being able to use odo in a restricted env belongs in the NOTE. And then the additional info about finding the binaries in the web console belongs in normal text below the NOTE.

cli_reference/openshift_developer_cli/installing-odo.adoc
```diff
@@ -5,14 +5,15 @@ include::modules/common-attributes.adoc[]
...
 
 [NOTE]
 ====
-Currently, `{odo-title}` does not support installation in a restricted network environment.
+You can also find the URL to the latest binaries in the web console of your cluster. To access them using the web console; click the *?* icon in the upper-right corner of the window, and select *Command Line Tools* in the drop-down menu.
```
bergerhoffer, line 14: * Suggest simplifying to something like "You can also download the latest binaries from the {product-title} web console." Or even combining into "You can also download the latest binaries from the {product-title} web console by clicking the *?* icon in the upper-right corner and select *Command Line Tools*."
* Use a comma instead of a semicolon in the second sentence.

bergerhoffer (commented):
Just two comments

cli_reference/openshift_developer_cli/installing-odo.adoc
boczkowska: You can't really _download_ them this way. Clicking the ? icon in the upper-right corner and selecting Command Line Tools will lead user to the page with titles and URLs of different CLI tools' folders on the file server (odo, helm etc). And there, on the file server, they can download the actual binary file. 

Otherwise, good call, thank you! I've shortened the sentence and included it as a part of introduction in both modules.

bergerhoffer: Fair enough!

boczkowska:
@bergerhoffer I completely forgot about this PR! Could you merge please if everything's okay?

bergerhoffer:
Sure thing @boczkowska, merging! Just FYI this will go into both 4.5 and 4.4 now.

PR #21584, opened 2020-04-27
Fixing version number for new features.

JStickler:
Need to update this to the correct Jaeger version that was released.


JStickler:
@neal-timpe, one tiny update.  The version number here is hardcoded so that we maintain the history for each release, but I forgot to update it because I was thinking it was a variable.
Need to be cherry picked to 4.3 and 4.4.

PR #21591, opened 2020-04-27
Release notes adjustments

codyhoag:
FYI @Preeticp - the other part of our discussions regarding 4.4 RN adjustments.


Preeticp:
@codyhoag Pipeline mentions from the Bug Fixes section need to be removed too. I see them on L 914, 981, 986, and 993.
Also, I wonder if the 2 web console sections should be immediately after each other. Probably something to be considered for later?

PR #21628, opened 2020-04-28
no base64 encoding allowed in additionalTrustedCA

miminar:
Just struggled with this until I foudnd out that the value shall be a pure unencoded PEM certificate.

Just like suggested in [Setting up additional trusted certificate authorities for builds](https://docs.openshift.com/container-platform/4.3/builds/setting-up-trusted-ca.html):

> The ConfigMap must be created in the openshift-config namespace.
>   - domain is the key in the ConfigMap; value is the PEM-encoded certificate.
>     - Each CA must be associated with a domain. The domain format is hostname[..port].

Applies to all 4.* releases.

### DIFF
--- a/modules/images-configuration-file.adoc
+++ b/modules/images-configuration-file.adoc
@@ -60,8 +60,8 @@ administrators will have the appropriate permissions.
 should be trusted during `ImageStream import`, `pod image pull`,
 `openshift-image-registry pullthrough`, and builds. The namespace for this ConfigMap is
 `openshift-config`. The format of the ConfigMap is to use the registry hostname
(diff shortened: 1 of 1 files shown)


kalexand-rh:
@wzheng1, will you PTAL?

vikram-redhat:
@wzheng1 ^ 

@bmcelvee please also follow up.

@xltian ^

wzheng1:
@kalexand-rh @vikram-redhat Sorry for missing this review, ack now : )

kalexand-rh:
Thanks @wzheng1! Merging.

PR #21629, opened 2020-04-28
BZ1828926 - Apply minimalism workshop feedback

bobfuru:
[BZ 1828926](https://bugzilla.redhat.com/show_bug.cgi?id=1828926)
This assembly was previously reviewed and approved by Eng, QE, and Docs. It then went through a review during a recent Minimalism workshop. This PR incorporates the feedback from that workshop, so I'd like another round of Docs review. Hoping that it can be completed in time for 4.4 GA, though not critical.

Preview build link: https://bz1828926--ocpdocs.netlify.app/openshift-enterprise/latest/storage/container_storage_interface/persistent-storage-csi-cloning.html


modules/persistent-storage-csi-cloning-overview.adoc
```diff
@@ -3,19 +3,27 @@
 // * storage/container_storage_interface/persistent-storage-csi-cloning.adoc
 
 [id="persistent-storage-csi-cloning-overview_{context}"]
-= Overview of CSI volume cloning
+= How CSI volume cloning works
 
-A Container Storage Interface (CSI) volume clone represents a duplicate of an existing persistent volume that is its own unique volume on the system at a particular point in time. With cloning, the back-end device creates an exact duplicate of the specified volume, rather than creating a new empty volume. After it has been dynamically provisioned, a volume clone can be consumed as any standard volume would be.
+Cloning creates an exact duplicate of the specified volume on the back-end device, rather than creating a new empty volume. Once dynamically provisioned, you can use a volume clone just as you would use any standard volume.
```
maxwelldb, line 8: Drive-by comment, here, but "once" should only be used to mean "one time" per ISG.

bobfuru: Ah, good catch! Fixed, thanks!

storage/container_storage_interface/persistent-storage-csi-cloning.adoc
```diff
@@ -5,7 +5,7 @@ include::modules/common-attributes.adoc[]
 
 toc::[]
 
-This document describes how to use volume cloning with supported Container Storage Interface (CSI) drivers to protect against data loss in {product-title}. Familiarity with xref:../../storage/understanding-persistent-storage.adoc#persistent-volumes_understanding-persistent-storage[persistent volumes] is suggested.
+Volume cloning duplicates an existing persistent volume to protect against data loss in {product-title}. This feature is only available with supported Container Storage Interface (CSI) drivers. You should be familiar with xref:../../storage/understanding-persistent-storage.adoc#persistent-volumes_understanding-persistent-storage[persistent volumes] before you begin.
```
bergerhoffer, line 8: "before you begin" what? Before you provision a CSI volume clone?

modules/persistent-storage-csi-cloning-overview.adoc
```diff
@@ -3,19 +3,27 @@
 // * storage/container_storage_interface/persistent-storage-csi-cloning.adoc
 
 [id="persistent-storage-csi-cloning-overview_{context}"]
-= Overview of CSI volume cloning
+= How CSI volume cloning works
```
bergerhoffer, line 6: I know that the minimalism feedback hasn't been in favor of Overview of/Understanding in headings, but this heading looks odd to me. Especially because we don't use heading like this elsewhere in the OCP docs (that I'm aware of). I thought as a team we determined that we weren't going to worry about this, and could use "About xyz", etc.

Also, saying how something works isn't exactly interchangeable with an overview/general introduction of something. Just my 2 cents.

modules/persistent-storage-csi-cloning-overview.adoc
```diff
@@ -3,19 +3,27 @@
...
+Cloning creates an exact duplicate of the specified volume on the back-end device, rather than creating a new empty volume. After dynamic provisioning, you can use a volume clone just as you would use any standard volume.
 
-Volume cloning is similar to volume snapshots, although it offers a more efficient method for creating a pre-populated volume copy. For example, a cluster administrator can duplicate a cluster volume by creating another instance of the existing cluster volume.
+No new API objects are required for cloning. The existing `dataSource` field in the PersistentVolumeClaim object is expanded so that it can accept the name of an existing PersistentVolumeClaim in the same namespace.
 
-There are no new API objects required for cloning. The existing `dataSource` field in the PersistentVolumeClaim object is expanded to be able to accept the name of an existing PersistentVolumeClaim in the same namespace.
+== Definition of CSI volume cloning
```
bergerhoffer, line 12: So this definitely seems odd to me. You've described how CSI volume cloning works first, and then give a definition of it here? Seems backwards, but also I'm not a fan of the "Definition of" heading.

modules/persistent-storage-csi-cloning-provisioning.adoc
```diff
@@ -5,22 +5,20 @@
...
 
 .Prerequisites
-* Logged in to a running {product-title} cluster.
-* A PVC created using a CSI driver that supports volume cloning.
-* Storage back end is configured for dynamic provisioning. Cloning support is not available for static provisioners.
-* Source and destination PVCs must exist in the same namespace.
-* Source and destination storage class must be the same.
+* You are logged in to a running {product-title} cluster.
+* Your PVC is created using a CSI driver that supports volume cloning.
+* Your storage back-end is configured for dynamic provisioning. Cloning support is not available for static provisioners.
```
bergerhoffer, line 15: back end is two words when it's a noun.

modules/persistent-storage-csi-cloning-provisioning.adoc
```diff
@@ -43,7 +41,7 @@ spec:
     name: pvc-1
 ----
 +
-<1> Name of the storage class that provisions the storage back end. Default storage class can be used and `storageClassName` can be omitted in the spec.
+<1> Name of the storage class that provisions the storage back-end. The default storage class can be used and `storageClassName` can be omitted in the spec.
```
bergerhoffer, line 44: Same here, back end was correct.

bergerhoffer: Also if you're changing the second sentence to start with "The", you should do the same for the first sentence, "The name of..."

modules/persistent-storage-csi-cloning-provisioning.adoc
```diff
@@ -58,5 +56,31 @@ A new PVC `pvc-1-clone` is created.
 ----
 $ oc get pvc pvc-1-clone
 ----
+
+To use the newly cloned PVC to configure a Pod:
+
+. Create and save a file with the Pod object described by the YAML. For example:
```
bergerhoffer, line 62: Single step procedures should be an unordered bullet.

storage/container_storage_interface/persistent-storage-csi-cloning.adoc
bergerhoffer, line 8: I know the wording was already there, but I'd just be careful about saying that it protects against data loss. It might be correct as is, but I might use something like "to help protect against data loss".

modules/persistent-storage-csi-cloning-provisioning.adoc
```diff
@@ -5,22 +5,20 @@
 [id="persistent-storage-csi-cloning-provisioning_{context}"]
 = Provisioning a CSI volume clone
 
-Provisioning of a CSI volume clone is triggered by the creation of a cloned persistent volume claim (PVC) API object. The cloned PVC is pre-populated with the contents of another PVC. After creation, the cloned volume adheres to the same rules as any other persistent volume.
+When you create a cloned persistent volume claim (PVC) API object, you trigger the provisioning of a CSI volume clone. The clone pre-populates with the contents of another PVC, adhering to the same rules as any other persistent volume.
+
+The one exception of cloned provisioning is that you must add a `dataSource` that references an existing PVC in the same namespace.
```
bergerhoffer, line 10: I had to re-read this line a few times to know what you were talking about that this was an exception to. I'd suggest adding this to the paragraph above, since you're talking about that it's the same rules as any other PV, with this exception.

modules/persistent-storage-csi-cloning-provisioning.adoc
```diff
@@ -58,5 +56,31 @@ A new PVC `pvc-1-clone` is created.
 ----
 $ oc get pvc pvc-1-clone
 ----
+
+To use the newly cloned PVC to configure a Pod:
```
bergerhoffer, line 60: This seems odd to have a second procedure in here like this. It should either be a step in the above procedure (if it's truly necessary to accomplish that task), or it should be a separate procedure.

bergerhoffer (commented):
A few thoughts

modules/persistent-storage-csi-cloning-overview.adoc
bobfuru: Agree! It was suggested that I break this out and use those subheads but your reaction makes sense. I'm going to revert to "Overview" and change the sentence order again slightly.

modules/persistent-storage-csi-cloning-provisioning.adoc
bobfuru: Thanks, I forgot why I had it two different ways. I hope they change their guidance to "backend" (always one word), as that's how I most often see it.

modules/persistent-storage-csi-cloning-provisioning.adoc
bobfuru: Agree. Originally, it was a separate procedure but the feedback was that it made more sense as part of this one. I've changed this so that the intro sentence is now a validation statement for the previous step, and I've made "Create and save..." the next step.

storage/container_storage_interface/persistent-storage-csi-cloning.adoc
bobfuru: Yes. I've changed to say "before you provision a CSI volume clone."

PR #21640, opened 2020-04-28
modules/installation-requirements-user-infra: add missing RHEL major version

vrutkovs:
@mburke5678  PTAL


mburke5678:
@vrutkovs This will show as _Fedora 8_ on OKD. Is that right?

mburke5678:
LGTM

PR #21664, opened 2020-04-29
typo fix

vikram-redhat:
@rolfedh I had to take out the use of the icon. I am not sure if we use that, and it was breaking the portal build.


rolfedh:
Okay. Sorry about that. No problem with removing the icon. I had verified that it didn't cause build failures in the checks.

PR #21688, opened 2020-04-29
Updated code block guidelines

lbarbeevargas:
@openshift/team-documentation PTAL, thank you!

Updated the docs guidelines regarding code blocks after team discussions around the new click-to-copy button and standardizing how we write codes blocks.

### DIFF
--- a/contributing_to_docs/doc_guidelines.adoc
+++ b/contributing_to_docs/doc_guidelines.adoc
@@ -484,6 +484,39 @@ $ lorem.sh
 ----
 ....
 
+* Do not use more than one command per code block. For example, the following must be split up into three separate code blocks.
++
+Run the following commands to create templates you can modify:
++
+----
+$ oc adm create-login-template > login.html
+----
++
+----
+$ oc adm create-provider-selection-template > providers.html
+----
++
+----
+$ oc adm create-error-template > errors.html
+----
++
+
+* Separate a command and its related output into individual code blocks. For example:
++
+Use the `oc new-project` command to create a new project.
++
+----
+$ oc new-project my-project
+----
+
++
+The output verifies that a new project was created.
++
+
+----
+Now using project "my-project" on server "https://openshift.example.com:6443".
+----
+
 * Try to use callouts to provide information on what the output represents when required.
 +
 Use this format when embedding callouts into the code block:


contributing_to_docs/doc_guidelines.adoc
```diff
@@ -484,6 +484,32 @@ $ lorem.sh
...
+* Separate a command and its related output into individual code blocks. For example:
++
+Use the `oc new-project` command to create a new project.
++
+----
+$ oc new-project my-project
+----
+
++
+Output that verifies a new project was created.
```
lbarbeevargas, line 506: I'm rethinking this line. The example could read, "The output verifies that a new project was created." Or something similar.

contributing_to_docs/doc_guidelines.adoc
```diff
@@ -484,6 +484,32 @@ $ lorem.sh
...
 
+* Use only one command per code block. For example, do not include more than one command in the same code block:
++
+Run the following commands to create templates you can modify:
++
+----
+$ oc adm create-login-template > login.html
+$ oc adm create-provider-selection-template > providers.html
+$ oc adm create-error-template > errors.html
+----
```
bobfuru, line 495: This is confusing to me. It is saying to only use one, then it's saying for example, not more than one. Then the example is something we shouldn't be doing. 

I suggest either:
A) Changing the guideline to something like, "Do not use more than one command per code block. For example, the following should be split up into three separate code blocks."
B) Change the example to show a single-command code block, following the guideline as it's laid out.

contributing_to_docs/doc_guidelines.adoc
bobfuru: ^^Yes, this is better.

bobfuru (commented):
Couple of comments before merging.

lbarbeevargas:
@bobfuru PTAL with your peer review feedback applied. Thank you!

bobfuru:
LGTM!

contributing_to_docs/doc_guidelines.adoc
```diff
@@ -484,6 +484,39 @@ $ lorem.sh
 ----
 ....
 
+* Do not use more than one command per code block. For example, the following should be split up into three separate code blocks.
```
jboxman, line 487: But is it _should_ or _must_?

lbarbeevargas: Good point, @jboxman. I can change it to "must" since this is the guideline moving forward. @bobfuru, thoughts?

bobfuru: Good catch; my vote is _must_.

maxwelldb: Seconding "must" as an RFC 2119 booster. ISG agrees, too.

lbarbeevargas:
@jboxman @bobfuru Updated "should" to "must".

bobfuru:
> @jboxman @bobfuru Updated "should" to "must".

Cool. I'll merge but can override if sentiment changes.

vikram-redhat:
git cherrypick enterprise-3.11

vikram-redhat:
@lbarbeevargas how does it work with multi-line commands and the edge case Jason mentioned? I think it would work as expected, but am not sure.

PR #21705, opened 2020-04-29
ES High Availability example in OpenShift documentation is wrong

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1702665

### DIFF
--- a/install_config/aggregate_logging.adoc
+++ b/install_config/aggregate_logging.adoc
@@ -814,13 +814,13 @@ to a value of `1` or higher to create replicas.
 
 Use the following scenarios as a guide for an {product-title} cluster with three Elasticsearch nodes:
 
-* If you can tolerate one Elasticsearch node going down,
-set `openshift_logging_es_number_of_replicas` to `1`. This ensures
-that two nodes have a copy of all of the Elasticsearch data in the cluster.
+* With `openshift_logging_es_number_of_replicas` set to `1`, two nodes have a copy of all of the Elasticsearch data in the cluster.
+This ensures that if a node with Elasticsearch data goes down, another node has a copy of all of the Elasticsearch data in the cluster.  
 
-* If you must tolerate two Elasticsearch nodes going down,
-set `openshift_logging_es_number_of_replicas` to `2`. This ensures that
-every node has a copy of all of the Elasticsearch data in the cluster.
+* With `openshift_logging_es_number_of_replicas` set to `3`, four nodes have a copy of all of the Elasticsearch data in the cluster.
+This ensures that if three nodes with Elasticsearch data go down, one node has a copy of all of the Elasticsearch data in the cluster.  
++
+In this scenario, with multiple Elasticsearch nodes going down, Elasticsearch status would be RED, and new Elasticsearch shards would not be allocated.  However, because of the high availability, you do not lose your Elasticsearch data.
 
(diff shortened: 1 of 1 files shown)


vikram-redhat:
@anpingli PTAL

anpingli:
Why we mix up the quorum and replicas together? 

if replicas=1, the replica shard will be created and locate in the other nodes.that needs at least two nodes in the cluster.
if replicas=3, two replica shards will be created and locate in different nodes, that need at least three nodes in the cluster.

To prevent split-brain, it is better to use an odd number, to provide toleration and auto-recovery ability,  n/2+1 nodes should be alive in the ES cluster.  so when replicas=1, the cluster number should >=3. when replics=2, the cluster node number should >=5.

mburke5678:
@anpingli I believe this section is intended to describe how using multiple replicas can help if some of your nodes go down. 

If you increase the number of replicas, you are less likely to lose data if nodes go down. However, there is a trade-off as more replicas require more resources.

anpingli (commented):
/lgtm

mburke5678:
Preview: http://file.rdu.redhat.com/~mburke/BZ-1702665/install_config/aggregate_logging.html#aggregated-elasticsearch

Aggregating Container Logs -> Elasticsearch -> Configuring Elasticsearch for High Availability

install_config/aggregate_logging.adoc
```diff
@@ -814,13 +814,13 @@ to a value of `1` or higher to create replicas.
 
 Use the following scenarios as a guide for an {product-title} cluster with three Elasticsearch nodes:
 
-* If you can tolerate one Elasticsearch node going down,
-set `openshift_logging_es_number_of_replicas` to `1`. This ensures
-that two nodes have a copy of all of the Elasticsearch data in the cluster.
+* With `openshift_logging_es_number_of_replicas` set to `1`, two nodes have a copy of all of the Elasticsearch data in the cluster.
+This ensures that if a node with the Elasticsearch data goes down, another node has a copy of all of the Elasticsearch data in the cluster.
```
bergerhoffer, line 818: Probably fine, but just wondering if "if a node with the Elasticsearch data" should be "if a node with Elasticsearch data"

install_config/aggregate_logging.adoc
```diff
@@ -814,13 +814,13 @@ to a value of `1` or higher to create replicas.
...
-set `openshift_logging_es_number_of_replicas` to `1`. This ensures
-that two nodes have a copy of all of the Elasticsearch data in the cluster.
+* With `openshift_logging_es_number_of_replicas` set to `1`, two nodes have a copy of all of the Elasticsearch data in the cluster.
+This ensures that if a node with the Elasticsearch data goes down, another node has a copy of all of the Elasticsearch data in the cluster.  
 
-* If you must tolerate two Elasticsearch nodes going down,
-set `openshift_logging_es_number_of_replicas` to `2`. This ensures that
-every node has a copy of all of the Elasticsearch data in the cluster.
+* With `openshift_logging_es_number_of_replicas` set to `3`, four nodes have a copy of all of the Elasticsearch data in the cluster.
+This ensures that if three nodes with the Elasticsearch data go down, one node has a copy of all of the Elasticsearch data in the cluster.
```
bergerhoffer, line 821: Same question on if the "the" is necessary here.

PR #21723, opened 2020-04-30
Installing: add IBM Power, add IBM Power w/ resticted network, update…

rh-max:
… IBM Z

Add version of the installing section for IBM Power

Link changes for RHCOS in case of IBM Power

Add section on restricted network install for IBM Power

Change resource requirements for IBM Power

IBM Z installation: change image names and preferred requirements

Fix duplicate ID

Updates for IBM Z installation section

Fixes in references to the system on which installing

Create a dedicated context and use it for conditionals

Fix conditionals

Conditional fixes

Remove the Telemetry section from the restricted network installation part

Add a missing :context:

Fix context and conditionals

Several updates

Move an "Addtional resources" section to the correct page

Implement feedback from peer review

Substitute server names in examples

Fix conditionals

Change version format in an command output


rh-max:
@bergerhoffer This commit is a squash of the many commits in https://github.com/openshift/openshift-docs/pull/20877 without further changes, apart from conflict resolutions when cherry-picking into `enterprise-4.3`. It's ready for merging.

modules/installation-bare-metal-config-yaml.adoc
```diff
@@ -135,3 +156,15 @@ endif::restricted[]
 ifeval::["{context}" == "installing-restricted-networks-bare-metal"]
 :!restricted:
 endif::[]
+ifdef::openshift-origin[]
```
bergerhoffer, line 159: It looks like this ifdef got added probably during a conflict from 4.4, but it wasn't defined at the top of the file.

rh-max: You are right, fixed.

rh-max:
@bergerhoffer Thank you!

PR #21744, opened 2020-05-01
Odo correction for no additional server in understanding docs

cdrage:
Based on user feedback, updates the documentation to suggest that there
is no need for an **additional** server other than OpenShift / the
cluster you're using. Rather than saying there is *no server required*.

Signed-off-by: Charlie Drage <charlie@charliedrage.com>


cdrage:
ping @boczkowska

PR #21750, opened 2020-05-01
BZ1785123 internal Ingress Controller for AWS restricted networks

kalexand-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1785123

@jianlinliu, will you PTAL?

### DIFF
--- a/modules/installation-generate-aws-user-infra-install-config.adoc
+++ b/modules/installation-generate-aws-user-infra-install-config.adoc
@@ -80,6 +80,7 @@ is required for an installation in a restricted network.
 .. Update the `pullSecret` value to contain the authentication information for
 your registry:
 +
+[source,yaml]
 ----
 pullSecret: '{"auths":{"<local_registry>": {"auth": "<credentials>","email": "you@example.com"}}}'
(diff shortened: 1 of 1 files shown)


kalexand-rh:
@jianlinliu, will you please take another look? I think that the request to clarify how to make the ingress controller private makes sense in this context, even if the private cluster method exists. I think that making the step optional is the right thing to do, since it's not actually mandatory.

modules/installation-generate-aws-user-infra-install-config.adoc
```diff
@@ -109,6 +109,14 @@ imageContentSources:
 ----
 +
 Use the `imageContentSources` section from the output of the command to mirror the repository or the values that you used when you mirrored the content from the media that you brought into your restricted network.
+
+.. Optional: Set the publishing strategy to `Internal`
++
+----
```
adellape, line 115: Would adding `[source,yaml]` to this (and any other relevant code blocks in this module) be appropriate?

adellape (commented):
Just one musing.

PR #21761, opened 2020-05-03
Add Rouge syntax highlighter CSS

jboxman:
Add the necessary CSS classes for syntax highlighting.

- https://spsarolkar.github.io/rouge-theme-preview/

### DIFF
--- a/_stylesheets/docs.css
+++ b/_stylesheets/docs.css
@@ -180,6 +180,225 @@ span.clipboard-button:hover {
 }
 /* End ClipboardJS edits */
 
+/* Rouge syntax highlighting: 'GitHub' style */
+/* From: https://spsarolkar.github.io/rouge-theme-preview/ */
+
+.highlight table td {
+     padding: 5px;
+}
+ .highlight table pre {
+     margin: 0;
+}
+ .highlight .cm {
+     color: #999988;
+     font-style: italic;
+}
+ .highlight .cp {
+     color: #999999;
+     font-weight: bold;
+}
+ .highlight .c1 {
+     color: #999988;
+     font-style: italic;
+}
+ .highlight .cs {
+     color: #999999;
+     font-weight: bold;
+     font-style: italic;
+}
+ .highlight .c, .highlight .cd {
+     color: #999988;
+     font-style: italic;
+}
+ .highlight .err {
+     color: #a61717;
+     background-color: #e3d2d2;
+}
+ .highlight .gd {
+     color: #000000;
+     background-color: #ffdddd;
+}
+ .highlight .ge {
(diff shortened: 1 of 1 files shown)


jboxman:
Preview: http://file.bos.redhat.com/jboxman/rouge-style/networking/using-sctp.html

cc @vikram-redhat

jboxman:
I cheated and dumped the styles in a `<style>` tag.

jboxman:
@vikram-redhat, this has the GH theme ready to go already.

vikram-redhat:
@openshift/team-documentation @openshift/team-aligned-docs FYI that this will be merged this week.

maxwelldb:
Just reiterating because poll comments are easy to lose: many of these styles, including the GH one, don't quite meet the latest WCAG AA contrast requirements. They'd need some modification to be accessible.

jboxman:
@maxwelldb, did we meet that standard before?

maxwelldb:
@jboxman Just doing a few spot checks, yeah, looks like it. I'm actually seeing AAA for terminal and YAML contrast ratios--8.38 and 8.97, respectively.

maxwelldb:
As an aside, WCAG encourages allowing people to select from different themes. :smiling_imp: 

>Note: Some people with cognitive disabilities require color combinations or hues that have low contrast, and therefore we allow and encourage authors to provide mechanisms to adjust the foreground and background colors of the content. Some of the combinations that could be chosen may have contrast levels that will be lower than those found in the Success Criteria. This is not a violation of this Success Criteria provided there is a mechanism that will return to the default values set out in the Success Criteria.

jboxman:
@maxwelldb, it looks like the styles can be adjusted manually with: https://accessible-colors.com

maxwelldb:
Might make sense to implement an accessible theme from a common palette rather than adjusting text color + background color for each item in isolation, but that's definitely a quick and painless resource.

jboxman:
@maxwelldb, all of this is completely outside my wheelhouse. So I wouldn't really know where to begin.

maxwelldb:
@jboxman This is a good review: https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Perceivable/Color_contrast

jboxman:
@maxwelldb, but not color palettes and such. I've never been a unicorn. I can do backend stuff on occasion, but I never took to frontend work. So my ability in this area is limited.

maxwelldb:
@jboxman Could always just steal some high-contrast stylesheets from elsewhere. Ex: https://maxchadwick.xyz/pygments-high-contrast-stylesheets/

PR #21771, opened 2020-05-04
Remove downward api that does not work.

shawn-hurley:
* remove docs that will not work
* needs to be backported to 4.3 and 4.4


kalexand-rh:
@jianzhangbjz, will you please confirm this change?

jianzhangbjz:
@kalexand-rh Looks good to me, thanks! BTW, I didn't see this section in [4.4 release doc](https://docs.openshift.com/container-platform/4.4/operators/olm-restricted-networks.html). Am I missing something?

jianzhangbjz:
Never mind, I find it, it's here: https://docs.openshift.com/container-platform/4.4/operators/operator_sdk/osdk-generating-csvs.html#olm-enabling-operator-for-restricted-network_osdk-generating-csvs

PR #21818, opened 2020-05-04
fix anchors for ingress controller

git001:
The links for Ingress Enhancements have used wrong anchors


codyhoag:
Thanks, @git001. Those edits do make the links more useful.

I've tested the link updates and they function properly.

PR #21823, opened 2020-05-05
Fix extra indentation for 3scale YAML

jboxman:


jboxman:
Hi @neal-timpe,

I adjusted the indentation for this YAML, but it might be deeply indented by intention?

jboxman:
@JStickler, this seems to be deeply indented; do you know if that was the intent?

neal-timpe:
I don't think this module is being used, but you're probably right.

jboxman:
@neal-timpe, I see it here:

```
  - Name: Customizing the installation
    File: customizing-installation-ossm
```

`service_mesh/service_mesh_install/customizing-installation-ossm.adoc`

```
include::modules/ossm-cr-threescale.adoc[leveloffset=+1]
```

neal-timpe:
Ok, thanks. I can't speak to intention, but I don't see why it should be indented that way in any case. @JStickler Does @jboxman 's change look okay to you?

JStickler:
@jboxman and @neal-timpe, I didn't write this so I'll have to take a read through and see what the context is.  Although I suspect that rather than removing the indentations, that what we need to do is add the beginning of the file (spec:istio:) so that it's clear where that chunk should be nested.  

@jboxman I'm curious what you were doing that flagged this but not the other YAML snippets in this assembly?

jboxman:
@JStickler, it came up in a search I was doing for YAML files for reasons I no longer remember; it looked suspect, so I created a PR for it.

JStickler:
@dfennessy, care to weigh in here?

dfennessy:
> @dfennessy, care to weigh in here?

Hi @JStickler 

The changes to the indentation by Jason look good to me.

philipgough (approved):
@dfennessy - Both YAML snippets are valid but I don't see any reason for the extra indentation so better to remove

dfennessy:
> @dfennessy - Both YAML snippets are valid but I don't see any reason for the extra indentation so better to remove

Thanks @PhilipGough

JStickler (approved):
Darren and Phillip are on the 3Scale team, so if they approve, I think we're OK making this change.

PR #21850, opened 2020-05-06
dedicated fixes

vikram-redhat:
@sheriff-rh some fixes for dedicated. Could you please check if these need to be put back to master/4.4/4.5?


sheriff-rh:
Thanks @vikram-redhat  - not sure how that happened. I've manually cherry picked to 4.4+.

PR #20939, opened 2020-04-06
Bug 1793648, added disk recommendations for etcd

ahardin-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1793648

PR #21922, opened 2020-05-07
Follow-up edits to PR#20939

ahardin-rh:
Follow-up edits to https://github.com/openshift/openshift-docs/pull/20939


ahardin-rh:
FYI @dmoessne @hexfusion

PR #21925, opened 2020-05-07
Put tech preview notification below section heading

russellb:
While looking up the documentation on three-node bare metal clusters,
I missed the tech preview note because it is displayed above the
three-node cluster section heading.  I think it would be more clear if
it was displayed within the section.

### DIFF
--- a/installing/installing_bare_metal/installing-bare-metal.adoc
+++ b/installing/installing_bare_metal/installing-bare-metal.adoc
@@ -52,8 +52,11 @@ include::modules/installation-bare-metal-config-yaml.adoc[leveloffset=+2]
 
 include::modules/installation-configure-proxy.adoc[leveloffset=+2]
 
+[id="installing-bare-metal-three-node"]
+== Running a three-node cluster
+
 :FeatureName: Three-node cluster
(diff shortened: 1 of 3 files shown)


vikram-redhat:
@kalexand-rh thoughts?

kalexand-rh:
I see the point, but this proposed change isn't mod docs compliant - you can't include a module in another module. You'd have to make a heading in the assembly and nest both the tech preview module and the feature module under it.

russellb:
> I see the point, but this proposed change isn't mod docs compliant - you can't include a module in another module. You'd have to make a heading in the assembly and nest both the tech preview module and the feature module under it.

I've tried to update this to follow this suggestion.  It's honestly not tested, though.

kalexand-rh:
@russellb, close! Each module's supposed to keep an H1 heading. I added a commit to your branch, and here's a rendered version of it: http://file.rdu.redhat.com/kalexand/062520/3-node-clusters-tech-preview/installing/installing_bare_metal/installing-bare-metal.html#installation-three-node-cluster_installing-bare-metal

russellb:
Cool, that works.  Thanks for the help!

PR #21943, opened 2020-05-08
Bug 1833239: [DOCS] Add warning message about blocking all RH registry when configuring "allowedRegistries"

bysnupy:
* Versions: OCP4.4, and other OCP4.x either

* Descriptions:
  `allowedRegistries` can block Red Hat registries without specific configurations about that. And it's big impact on running OCP4 cluster, so we should warn about that in advance on the docs.

* Reference: [[DOCS] Add warning message about blocking all RH registry when configuring "allowedRegistries"](https://bugzilla.redhat.com/show_bug.cgi?id=1833239)

### DIFF
--- a/modules/images-configuration-parameters.adoc
+++ b/modules/images-configuration-parameters.adoc
@@ -77,6 +77,11 @@ registries are allowed.
 `allowedRegistries`: Whitelisted for image pull and push actions. All other
 registries are blocked.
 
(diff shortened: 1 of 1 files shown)


bysnupy:
Another PR has been resolved this here, so I close this.
https://github.com/openshift/openshift-docs/pull/21983

PR #21973, opened 2020-05-08
Adding link to Support policy for unmanaged Operators

mburke5678:
Until we decide, if we decide, to remove unmanaged configurations, adding a link to the new _Support policy for unmanaged Operators_ module.


mburke5678:
@sichvoge PTAL

sichvoge:
Should that we added to all sections that mention "Unmanaged"?

mburke5678:
@sichvoge We cannot add links to module files, those files added using `include` statements. I added a note to those files, in the prerequisite section where we advise that unmanaged is required. 
I added the link to the assemblies that call the modules.

PR #21978, opened 2020-05-09
availability regions for ARO 4x

lamek:
We're temporarily publishing the available regions for aro 4.x here until we can get the automated tool updated on the MSFT side.

### DIFF
--- a/welcome/index.adoc
+++ b/welcome/index.adoc
@@ -46,6 +46,21 @@ a| * Overview of OpenShift architecture
 * Developing on OpenShift
 * Managing your cluster 
 |===
+
+== Available Regions
+
+{product-title} is currently available in the following regions:
+
+* East US (Public/United States)
+* East US 2 (Public/United States)
+* South Central US (Public/United States)
+* North Europe (Public/Europe)
+* West Europe (Public/Europe)
+* Southeast Asia (Public/Asia Pacific)
+* Australia East (Public/Australia)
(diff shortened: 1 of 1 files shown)


welcome/index.adoc
```diff
@@ -46,6 +46,21 @@ a| * Overview of OpenShift architecture
...
+
+* East US (Public / United States)
+* East US 2 (Public / United States)
+* South Central US (Public / United States)
+* North Europe (Public / Europe)
+* West Europe (Public / Europe)
+* Southeast Asia (Public / Asia Pacific)
+* Australia East (Public / Australia)
+* UK South (Public / United Kingdom)
+* West US 2 (Public / United States)
```
bobfuru, line 62: Not sure if this is copied from elsewhere in docs, but I prefer not to use spaces between forward slashes. There's not specific guidance in ISG, but they don't use them when writing things like `I/O` and `read/write`, so I personally wouldn't use them here. My 2¢.

LGTM otherwise!

bobfuru (commented):
> We're temporarily publishing the available regions for aro 4.x here until we can get the automated tool updated on the MSFT side.

Just double-checking then that this is meant only for `enterprise-4.3`.

welcome/index.adoc
lamek: Got it. I'll update that. And yes, copied right out of an email -- good eyes!

PR #21981, opened 2020-05-11
CNV - change wording and version for VM api ref link

aburdenthehand:
PM and eng recommended we re-phrase the 'definitive' part of the API reference and link to the specific version instead of master.

Also removed the duplicate link.

### DIFF
--- a/cnv/cnv_virtual_machines/cnv-create-vms.adoc
+++ b/cnv/cnv_virtual_machines/cnv-create-vms.adoc
@@ -37,6 +37,12 @@ include::modules/cnv-creating-vm.adoc[leveloffset=+1]
 
 include::modules/cnv-vm-storage-volume-types.adoc[leveloffset=+1]
 
-See the
-https://kubevirt.io/api-reference/master/definitions.html#_v1_virtualmachinespec[kubevirt
-API Reference] for a definitive list of virtual machine settings.
+== Additional resources
+
+The `VirtualMachineSpec` definition in the link:https://kubevirt.io/api-reference/v0.26.5/definitions.html#_v1_virtualmachinespec[KubeVirt v.0.26.5 API Reference] provides broader context for the parameters and hierarchy of the virtual machine specification.
+
+[NOTE]
+====
+The KubeVirt API Reference is the upstream project reference and might contain parameters that are not supported in {CNVProductName}. 
+====
+
--- a/modules/cnv-vm-storage-volume-types.adoc
+++ b/modules/cnv-vm-storage-volume-types.adoc
@@ -5,9 +5,7 @@
 [id="cnv-vm-storage-volume-types_{context}"]
 = Virtual machine storage volume types
 
-Virtual machine storage volume types are listed, as well as domain and volume settings. See the
-https://kubevirt.io/api-reference/master/definitions.html#_v1_virtualmachinespec[kubevirt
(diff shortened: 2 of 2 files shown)


modules/cnv-vm-storage-volume-types.adoc
```diff
@@ -5,9 +5,10 @@
 [id="cnv-vm-storage-volume-types_{context}"]
 = Virtual machine storage volume types
 
-Virtual machine storage volume types are listed, as well as domain and volume settings. See the
-https://kubevirt.io/api-reference/master/definitions.html#_v1_virtualmachinespec[kubevirt
-API Reference] for a definitive list of virtual machine settings.
+Virtual machine storage volume types are listed, as well as domain and volume settings. +
```
dankenigsberg, line 8: would this show only in the storage section, not the virt one?

How about defining a variable for `KUBEVIRT_VERSION_SHIPPED=v0.26.5` (or something like this) so it is easier to remember to update every type we release, and easier to refer in multiple places (now only two)

aburdenthehand: Yeah, this PR is just a quick fix to remove the issue in the current published docs - targetted to enterprise-4.4 only. The link is in the storage types module as part of the 'creating a VM' assembly.

For the long term I like the idea of a kubevirt version variable for the docs. Is that something we can rely on going forward? If we don't mind linking to the kubevirt API reference in future versions - just as long as it's linked to the appropriate version - then that is a fine solution. We could add the variable to our pre-flight checklist to ensure it gets updated.

dankenigsberg: I think that there is a lot of value in explicitly stating the version of KubeVirt used - as long as we keep the language defensive. It has to be clear that not everything that is upstream is supported within OpenShift. I'd prefer if the same language is kept on the virt side, too.

aburdenthehand: @dankenigsberg I've moved it out of the storage section and at end of the 'create VM' assembly. There is now only one reference/link to the API reference, and it includes a note calling out that not everything in the API reference is supported.
Using explicit version here because it's published doc. When this is acked/merged I'll do the same in master with the KUBEVIRT_VERSION variable. 

Screenshot of built doc:
Screenshot from 2020-05-12 12-42-37

cnv/cnv_virtual_machines/cnv-create-vms.adoc
```diff
@@ -37,6 +37,12 @@ include::modules/cnv-creating-vm.adoc[leveloffset=+1]
...
-See the
-https://kubevirt.io/api-reference/master/definitions.html#_v1_virtualmachinespec[kubevirt
-API Reference] for a definitive list of virtual machine settings.
+== Additional resources
+
+The `VirtualMachineSpec` definition in the link:https://kubevirt.io/api-reference/v0.26.5/definitions.html#_v1_virtualmachinespec[KubeVirt v.0.26.5 API Reference] provides broader context for the parameters and hierarchy of the virtual machine specification.
+
+[NOTE]
+====
+The KubeVirt API Reference is the upstream project reference and may contain parameters that are not supported in {CNVProductName}.
```
ousleyp, line 46: s/may/might

ousleyp (commented):
One nit, otherwise LGTM

PR #21987, opened 2020-05-11
LogForwarding endpoint must be an node name or FQDN, not an IP Address

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1831552

### DIFF
--- a/modules/cluster-logging-log-forwarding-about.adoc
+++ b/modules/cluster-logging-log-forwarding-about.adoc
@@ -18,8 +18,13 @@ An _output_ is the destination for log data and a _pipeline_ defines simple rout
 
 An output can be either:
 
-* `elasticsearch` to forward logs to an external Elasticsearch v5.x cluster, specified by server name or FQDN, and/or the internal {product-title} Elasticsearch instance. 
-* `forward` to forward logs to an external log aggregation solution. This option uses the Fluentd *forward* plug-ins.
+* `elasticsearch` to forward logs to an external Elasticsearch v5.x cluster and/or the internal {product-title} Elasticsearch instance. 
+* `forward` to forward logs to an external log aggregation solution. This option uses the Fluentd *forward* plug-ins. 
+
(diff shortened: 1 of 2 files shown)


modules/cluster-logging-log-forwarding-about.adoc
```diff
@@ -13,11 +13,15 @@ If you want to use only the default internal {product-title} Elasticsearch insta
...
 An output can be either:
 
-* `elasticsearch` to forward logs to an external Elasticsearch v5.x cluster, specified by server name or FQDN, and/or the internal {product-title} Elasticsearch instance. 
-* `forward` to forward logs to an external log aggregation solution. This option uses the Fluentd *forward* plug-ins.
+* `elasticsearch` to forward logs to an external Elasticsearch v5.x cluster and/or the internal {product-title} Elasticsearch instance. 
+* `forward` to forward logs to an external log aggregation solution. This option uses the Fluentd *forward* plug-ins. 
+
+[NOTE]
+====
+The endpoint must be a server name or FQDN, not an IP Address.
```
anpingli, line 23: @mburke5678  @jcantrill   "Must be"?  The IP Address works in my testing.

anpingli: According to https://bugzilla.redhat.com/show_bug.cgi?id=1825923#c10.  
It should be 'The endpoint must be a server name or FQDN, not an IP Address **when cluster-wide proxy are enabled**.'

mburke5678: You seem to be right. I was looking at comment 12:
_It's indicating that it should be a server name of FQDN, but I believe that it should be more clear in a Warning indicating that it shouldn't be used an IP Address._

I made the change as you suggested.

mburke5678:
@periklis PTAL

anpingli (commented):
LGTM

PR #19266, opened 2020-01-23
Bug 1794466: modules/installation-mirror-repository: Pivot to by-digest pullspecs

wking:
The `OCP_RELEASE` stuff landed in ced98c67ad (#16678).  Since then, 4.2.14 and later (including some 4.1 releases) have been [suffixing the Quay tags with architecture information][1].  By pivoting to digests pulled from the release errata, we decouple these docs from tag names.  Also pull the repo portion of the pullspec out of the errata, so there's less hard-coding in the docs.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1794466

PR #21993, opened 2020-05-11
Adding documentation for updating a disconnected cluster

LalatenduMohanty:
Signed-off-by: Lalatendu Mohanty <lmohanty@redhat.com>


updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,134 @@
...
-= Updating a disconnected cluster to a minor version
+= Updating a disconnected cluster
 include::modules/common-attributes.adoc[]
 :context: updating-disconnected-cluster
 
 toc::[]
 
-You can update, or upgrade, an {product-title} cluster that you installed in a
-restricted network to a minor version by using the web console.
+You can update, or upgrade a disconnected or fully air gapped {product-title} cluster using `oc` commandline.
```
wking, line 8: nit: "disconnected or fully air gapped" -> "restricted network"?

Also, first references to `oc` in openshift-docs seem to be:

    ... the `oc` command-line interface (CLI).

and similar.  I don't see existing references like:

    ... `oc` commandline.

wking:
Cross-linking the previous (abandoned?) attempts in #16894 and #17896.

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,134 @@
...
 
 .Prerequisites
 
+* Have access to a conatiner registry in the disconnected environment to push and pull images. The container registry should be compatible with Docker registry API v2.
```
wking, line 12: nit: "conatiner" -> "container" and "disconnected" -> "restricted-network".

I dunno if "have `oc` installed" gets listed as a prerequisite or not.  Looks like we do list it explicitly in some places (e.g. [here][1]).

[1]: https://github.com/openshift/openshift-docs/blame/4abd206b42d38e5bd66423d780c4c900b395e8ff/modules/cli-configuring-completion.adoc#L13

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,134 @@
...
 * Have access to the cluster as a user with `admin` privileges.
 See xref:../authentication/using-rbac.adoc[Using RBAC to define and apply permissions].
 * Have a recent xref:../backup_and_restore/backing-up-etcd.adoc#backup-etcd[etcd backup] in case your upgrade fails and you must xref:../backup_and_restore/disaster_recovery/scenario-2-restoring-cluster-state.adoc#dr-restoring-cluster-state[restore your cluster to a previous state].
```
wking, line 15: This is pretty grim, and we don't call it out in our [usual CLI update docs][1].  But if we want to punt and keep this existing wording for now, that's fine with me too.  Also not clear to me why this is over in `updating/` while the existing stuff is under `modules/`, but I'm ok punting on that as well ;).

[1]: https://github.com/openshift/openshift-docs/blame/4abd206b42d38e5bd66423d780c4c900b395e8ff/modules/update-upgrading-cli.adoc

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,134 @@
...
 
-//mirror module
+[IMPORTANT]
+====
+If you are upgrading from earlier release than 4.5.0, you must must use the manual method of creating configmap.
```
wking, line 19: As far as *creating* the ConfigMap goes, the version that matters is `oc`, not the original cluster version.  Not clear to me what the doc policy is about talking about version numbers, but my impression is that these master docs (so tracking 4.6?) just assume that folks will understand `oc` references to be referenced to (the latest?) 4.6.z `oc`.  If that's true, we can drop this callout.

I did raise "what happens if docs assume latest z-stream `oc` and users are running an earlier `oc` from that z stream that lacks $FEATURE" with @smarterclayton at some point, and he said that while that might happen, it hasn't happened often enough to be a problem.  I think that was in the context of openshift/enhancements#188, but it might have been an out of band discussion, because I don't see anything about it in the PR comments.

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,134 @@
...
+You can update, or upgrade a disconnected or fully air gapped {product-title} cluster using `oc` commandline.
 
 .Prerequisites
 
+* Have access to a conatiner registry in the disconnected environment to push and pull images. The container registry should be compatible with Docker registry API v2.
 * Have access to the cluster as a user with `admin` privileges.
 See xref:../authentication/using-rbac.adoc[Using RBAC to define and apply permissions].
 * Have a recent xref:../backup_and_restore/backing-up-etcd.adoc#backup-etcd[etcd backup] in case your upgrade fails and you must xref:../backup_and_restore/disaster_recovery/scenario-2-restoring-cluster-state.adoc#dr-restoring-cluster-state[restore your cluster to a previous state].
 
-//mirror module
```
wking, line 17: I don't think we should inline so much of the discussion in this file.  This placeholder was calling for a new:

```asciidoc
include::modules/update-mirror-restricted-network.adoc[leveloffset=+1]
```

or some such, to isolate the workflow that is unique to restricted-network clusters, while continuing to share the other modules which apply to both restricted- and connected-network clusters.  For an example of how to structure that, see #16894.

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,134 @@
...
 
-include::modules/update-service-overview.adoc[leveloffset=+1]
+This is because `oc` command enhancement is not available in previous releases.
+====
 
-include::modules/understanding-upgrade-channels.adoc[leveloffset=+1]
+[id="update-airgapped-env"]
+== Updating an air gapped {product-title} cluster
+
+An air gapped environment is the one in which your cluster nodes cannot access the internet. For this reason you must mirror the images to a filesystem disconnected from that environment and then bring that host or removable media across that gap. It may also be the case that there are multiple clusters within the air gapped network. In such a case it makes sense to configure a central registry in the air gapped environment from which every cluster can pull required images for upgrade.
```
wking, line 27: > ...In such a case it makes sense to...

This feels overly prescriptive to me.  I doubt many users will think "no thanks, I'll push the new update into each cluster's local registry", but I don't think it really matters to us.  Can we say something like:

> If there are multiple clusters within the restricted network, consider mirroring the release an related images to a single container image registry which all the clusters can access.  This can simplify maintenance by reducing the number of mirror repositories which need to be manually maintained.

or some such to sketch out the benefit.

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,134 @@
...
+
+.Procedure
+
+. Findout if there is link:https://access.redhat.com/solutions/4583231[an update path] exists between your current version and the version you want the cluster to move.
```
wking, line 31: I think Knowledgebase articles are a great stopgap while formal docs are worked out, but I don't like having the formal docs link out to them.  That sort of thing eventually ends up in things like #18944, where we spend a month and many people-hours hashing out how to bring the content into openshift-docs to ease future maintenance.  For the purpose of this PR, can we assume that the user already knows (e.g. from errata) which pullspec they want to mirror and that they don't care about validating the update vs. our upgrade recommendation service (i.e. Cincinnati)?  Because I don't think we have a good official story around figuring out the best next-hop yet (although [this script][1] is in an official-ish repository and automates at least some of what the referenced Knowledgebase article is covering).

[1]: https://github.com/openshift/cincinnati/blob/master/hack/available-updates.sh

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,134 @@
...
+
+. Mirror the version images to the internal container registry.
+
+** Connect the removable media to a system connected to the internet.
+** Mirror the images and configuration manifests to a directory on the removable media.
++
+----
+$ oc adm release mirror "--to-dir=${REMOVABLE_MEDIA_PATH}/mirror" quay.io/openshift-release-dev/ocp-release:4.3.0-x86_64
```
wking, line 39: This overlaps with [our existing mirror module][1].  But because #16894 and #17896 have stalled out on previous attempts at making that existing module generic enough to cover this case, I'm fine if we want to punt on a DRY fix and copy/edit over in a new location to address the update-specific variations.

[1]: https://github.com/openshift/openshift-docs/blame/4abd206b42d38e5bd66423d780c4c900b395e8ff/modules/installation-mirror-repository.adoc#L59

wking: And while I dislike using tags for this (#19266), if we do use them we probably want to use `...:{product-version}.0-x86_64` to get something that at least feels reasonable for whatever version of docs are actually building.  Or maybe we wrap this all up in a shell variable (`${OCP_RELEASE}`?) and avoid having any example tag or digest in the docs?

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,134 @@
...
+----
++
+
+** Take the media to the disconnected environment and upload the images to the local container registry
++
+----
+$ oc image mirror "--from-dir=${REMOVABLE_MEDIA_PATH}/mirror" file://openshift/release:4.3.0* REGISTRY/REPOSITORY
```
wking, line 46: `REGISTRY` and `REPOSITORY` should be actual environment variables and be declared with docs around their semantics (like [this][1], although ideally without the `export`, #22008).

[1]: https://github.com/openshift/openshift-docs/blame/4abd206b42d38e5bd66423d780c4c900b395e8ff/modules/installation-mirror-repository.adoc#L35-L54

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,134 @@
...
+$ oc apply -f mirror/config/signature-sha256-81154f5c03294534.yaml
+----
+. Perform the upgrade
++
+----
+$ oc adm upgrade --allow-explicit-upgrade --to-image REGISTRY/REPOSITORY@sha256:81154f5c03294534e1eaf0319bef7a601134f891689ccede5d705ef659aa8c92
+----
+
+[id="update-disconnected-manual_method_configmap_creation"]
+== Manual method to create the image signature configmap
```
wking, line 70: Can we split the procedure into _mirror_, _push_, and _upgrade_ subsections?  Then this manual workaround would be a subsection of _mirror_, instead of showing up here as a potential gotcha for someone who is reading along and has already walked their media into their restricted datacenter?  Odds are low, but still, I think it's easiest to follow if we lay it out in chronological order.

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,134 @@
...
+The following commands will create and apply the image signature configmap to the cluster you want to update. 
+
+.Procedure
+
+. Find the link:https://access.redhat.com/solutions/4583231[version] you want the cluster to update.
+
+. Add the version to OCP_RELEASE_NUMBER environment variable.
++
+----
+$ export OCP_RELEASE_NUMBER=<4.y.z>
```
wking, line 81: no need for `export`, #22008.

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,134 @@
...
+----
++
+. Get the release image digest from link:https://quay.io/[Quay].
++
+----
+$ export DIGEST="$(oc adm release info quay.io/openshift-release-dev/ocp-release:${OCP_RELEASE_NUMBER}-x86_64 | sed -n 's/Pull From: .*@//p')"
```
wking, line 87: should we paramaterize the arch too, instead of hard-coding `-x86_64`?  I really need to get back around to fixing [this][1], so we can stop talking about tags in these docs...

[1]: https://github.com/openshift/openshift-docs/pull/19266#discussion_r375878575

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,134 @@
...
+----
++
+. Get the image signature from link:https://mirror.openshift.com/pub/openshift-v4/signatures/openshift/release[mirror.openshift.com] website.
```
wking, line 90: We also canonically store signatures [on GCS][1].  Can we mention both locations?

[1]: https://github.com/openshift/cluster-update-keys/blob/cca4ce696383e70ae669e770bd63265a9540b721/stores/store-openshift-official-release

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,134 @@
...
+. Get the digest algorithm.
++
+----
+export DIGEST_ALGO=$(echo $DIGEST | cut -d: -f1)
+----
++
+. Get the digest signature.
++
+----
+export DIGEST_SIGNATURE=$(echo $DIGEST | cut -d: -f2)
```
wking, line 105: If you're going to split out the algo and hash (formally ["the encoded portion"][1]) into variables, can we do that before `curl`ing the signature so we don't have to drop the logic inline?

And no need to shell out to replace `:` with `=`.  If we can assume [a POSIX shell][2], we can:

```console
$ DIGEST_ALGO="${DIGEST%%:*}" # remove first colon and everything after
$ DIGEST_ENCODED="${DIGEST#*:}" # remove everything up to and including the first colon
```

`${DIGEST/:/=}` would be even more compact, but is a Bashism.

[1]: https://github.com/opencontainers/image-spec/blob/v1.0.1/descriptor.md#digests
[2]: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,134 @@
...
+----
++
+. Create the configmap.
++
+----
+$ cat >checksum-${OCP_RELEASE_NUMBER}.yaml <<EOF
```
wking, line 111: Probably better to use:

```console
$ DIGEST=sha256=a3fe9de9b338abc80e2afafdf38dff0c2de3efb61c6896e8e16495f59e717f53
$ DIGEST_ALGO="${DIGEST%%:*}"
$ DIGEST_ENCODED="${DIGEST#*:}"
$ curl -s "https://mirror.openshift.com/pub/openshift-v4/signatures/openshift/release/${DIGEST_ALGO}=${DIGEST_ENCODED}/signature-1" >"${DIGEST_ALGO}-${DIGEST_ENCODED}-1"
$ oc -n openshift-config-managed create configmap -o yaml --dry-run --from-file="${DIGEST_ALGO}-${DIGEST_ENCODED}-1" "release-image-signature-${OCP_RELEASE_NUMBER}"
```

Not sure if there's a convenient way to stuff in the `release.openshift.io/verification-signatures` label though...

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,134 @@
...
++
+. Install the config map in the cluster which needs to be updated.
++
+----
+$ oc apply -f checksum-${OCP_RELEASE_NUMBER}.yaml
+----
++
+. Get the created configmap to verify.
+----
+$ oc get cm release-image-${OCP_RELEASE_NUMBER} -o yaml
```
wking, line 132: Hopefully `oc apply` tells you if you fail.  I don't think we need to suggest folks double-check.

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,134 @@
...
++
+. Apply the the mirrored release image signature ConfigMap to the connected cluster
++
+----
+$ oc apply -f mirror/config/signature-sha256-81154f5c03294534.yaml
+----
+. Perform the upgrade
++
+----
+$ oc adm upgrade --allow-explicit-upgrade --to-image REGISTRY/REPOSITORY@sha256:81154f5c03294534e1eaf0319bef7a601134f891689ccede5d705ef659aa8c92
```
wking, line 66: This should probably use the canonical pullspec (the output from the earlier `Pull From` `grep`).  No need to use the local mirror pullspec, because this will get translated via the mirror's `ImageContentSourcePolicy` entry.

updating/updating-disconnected-cluster.adoc
LalatenduMohanty: I think the generic idea behind modules is that if you expect to reuse some doc in multiple places it should be a module. May be we want to create a module for some of the docs here. Lets move forward and take a look a back after some iteration.

updating/updating-disconnected-cluster.adoc
LalatenduMohanty: I agree that the issue is only with the version of `OC` command line. But I am assuming customers use same version of `oc` and OpenShift version. The reason I mentioned this because we plan to back port this document to 4.4 and 4.3. I am not confident what approach will best serve the backporting.

updating/updating-disconnected-cluster.adoc
LalatenduMohanty: I like the idea. I am going to create at least one module.

updating/updating-disconnected-cluster.adoc
LalatenduMohanty: It is very clear that we have a gap in the user experience workflow around how users can find the right version easily. My plan is to keep the KB article here as it is easier to find this way and then replace it with a official way (with a better user experience). I am planning to create a Jira to track this down.

updating/updating-disconnected-cluster.adoc
LalatenduMohanty: This (URL)[https://storage.googleapis.com/openshift-release/official/signatures/openshift/release] does not seems to work for me.

updating/updating-disconnected-cluster.adoc
smarterclayton: I am not a huge fan of the shell scripting approach, but I understand why people do it.  The goal is NOT to have fully executable scripts in our docs.  The goal is not to need that.  Having to bash is a gap in the product experience - we shouldn't just paper over those.

smarterclayton:
This doesn't include what I expected to see from https://github.com/openshift/openshift-docs/pull/17896 which is the terminology and flow changes (especially the removal of the atrocious registry creation).  Is that a follow up I'm missing?

17896 is what the flow *should* be and captures the flow and terminology that we want to use with customers (I haven't heard a concrete counter).  If that's not included here it needs to be a follow up.

Our current flow is not correct.

updating/updating-disconnected-cluster.adoc
wking: >  Having to bash is a gap in the product experience - we shouldn't just paper over those.

I don't think this is a product gap; we're just using shell as a convenient, familiar placeholder syntax here, and uesrs can plug in their chosen registry and repository without bothering with actual shell variables if they want.  The alternative of using placeholders like `REGISTRY` (instead of `$REGISTRY`) is just defining our own, non-standard placeholder syntax, and I see no benefit to doing that.

updating/updating-disconnected-cluster.adoc
wking: GCS does not provide directory listing.  But you can pull signatures:

```console
$ curl -s https://storage.googleapis.com/openshift-release/official/signatures/openshift/release/sha256=96cb606d04e94de1954a50ddb965eb02b76e182478bef9225df0ad523cfa6b38/signature-1 | gpg --decrypt
{"critical": {"image": {"docker-manifest-digest": "sha256:96cb606d04e94de1954a50ddb965eb02b76e182478bef9225df0ad523cfa6b38"}, "type": "atomic container signature", "identity": {"docker-reference": "quay.io/openshift-release-dev/ocp-release:4.4.0-x86_64"}}}gpg: Signature made Wed 29 Apr 2020 08:45:51 AM PDT using RSA key ID FD431D51
gpg: Good signature from "Red Hat, Inc. (release key 2) <security@redhat.com>"
gpg: WARNING: This key is not certified with a trusted signature!
gpg:          There is no indication that the signature belongs to the owner.
Primary key fingerprint: 567E 347A D004 4ADE 55BA  8A5F 199E 2F91 FD43 1D51
```

wking:
> This doesn't include what I expected to see from #17896...

If that PR gets picked back up and lands, or is carried and lands, I think we'd be happy to rebase on top.  But are there any corners we can cut to get _something_ landed around manual ConfigMap creation without having to resolve all the outstanding issues in one big PR?

smarterclayton:
I was expecting you guys to pick that up and drive it to completion - that was what I thought the last communication on it was.

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,131 @@
...
++
+----
+$ oc adm release mirror "--to-dir=${REMOVABLE_MEDIA_PATH}/mirror" quay.io/openshift-release-dev/ocp-release:{product-version}.0-x86_64
+----
++
+
+** Take the media to the disconnected environment and upload the images to the local container registry
++
+----
+$ oc image mirror "--from-dir=${REMOVABLE_MEDIA_PATH}/mirror" file://openshift/release:{product-version}.0* REGISTRY/REPOSITORY
```
jianlinliu, line 49: 1. This make me have to raise https://bugzilla.redhat.com/show_bug.cgi?id=1806779 again here. Per my testing, `oc image mirror` does not print out `ImageContentSourcePolicy` entry.  Maybe we need to raise the bug priority, so that QE can do adjustment in time for our test cases and automation scripts to align with this new change. 
2. Here we will only support user to upload files to registry (registry-mirror-to-disk-to-registry scenario), but not support mirror image from quay.io to local registry directly  (registry-mirror-to-registry scenario) any more? In present QE's disconnected cluster upgrade testing, we mainly tested registry-mirror-to-registry scenario.

LalatenduMohanty: Looks like https://bugzilla.redhat.com/show_bug.cgi?id=1806779 ON QA now. Can you test it and see if it works for you now?

jianlinliu: I just completed my verification some hours ago. Now looks more clear.

About how to mirror release image to private registry, there are two ways, one is registry-mirror-to-disk-to-registry, another is registry-mirror-to-registry, here we only mentioned one scenario.
And for registry-mirror-to-disk-to-registry scenario, we also need tell user https://github.com/openshift/openshift-docs/pull/17896/files#diff-21ddacc727855ec1dc3074e6e76e430dR56-R57 so that get ImageContentSourcePolicy, and applied it to connected cluster. 

I know this PR focus on upgrade part, but from customer view, this is the 1st time to mention registry-mirror-to-disk-to-registry scenario, install section did not mentioned this yet. Or do we intend to publish this one and https://github.com/openshift/openshift-docs/pull/17896 together?

LalatenduMohanty: @jianlinliu I have mentioned `--from-dir` and `--to-dir` workflow. In the first iteration  I have no intention of mentioning `-from=file:` workflow.  May be in a follow up PR. I have no plans around https://github.com/openshift/openshift-docs/pull/17896.

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,131 @@
...
+. Perform the upgrade
++
+----
+$ oc adm upgrade --allow-explicit-upgrade --to-image REGISTRY/REPOSITORY@sha256:81154f5c03294534e1eaf0319bef7a601134f891689ccede5d705ef659aa8c92
+----
+
+[id="update-restricted network-manual_configmap_creation"]
+== Manual method to create the image signature configmap
+
+The following commands will create and apply the image signature configmap to the cluster you want to update.
```
mhrivnak, line 75: Every customer who needs to do this is eventually going to copy and paste each line into a file and make a script that just does the whole thing. Should we do that once for everyone and just condense this into a script that is well-documented?

mhrivnak (changes requested):

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,131 @@
...
+* Have access to a container registry in the restricted-network environment to push and pull images. The container registry should be compatible with Docker registry API v2.
+* You must have the `oc` command-line interface (CLI) tool installed.
 * Have access to the cluster as a user with `admin` privileges.
 See xref:../authentication/using-rbac.adoc[Using RBAC to define and apply permissions].
 * Have a recent xref:../backup_and_restore/backing-up-etcd.adoc#backup-etcd[etcd backup] in case your upgrade fails and you must xref:../backup_and_restore/disaster_recovery/scenario-2-restoring-cluster-state.adoc#dr-restoring-cluster-state[restore your cluster to a previous state].
 
-//mirror module
+[IMPORTANT]
+====
+If you are upgrading from earlier release than 4.5.0, you must use the manual method of creating configmap.
```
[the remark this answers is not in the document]
LalatenduMohanty: Agree, going to remove this note and will keep it in the section about ConfigMaps.

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,131 @@
...
++
+----
+$ oc image mirror "--from-dir=${REMOVABLE_MEDIA_PATH}/mirror" file://openshift/release:{product-version}.0* REGISTRY/REPOSITORY
+----
++
+. Use `oc` to log in to a given cluster to be upgraded.
++
+[NOTE]
+====
+If you are upgrading from earlier release than 4.5.0, you must must use the manual method of creating configmap.
```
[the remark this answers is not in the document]
LalatenduMohanty: I am removing the top one, so it we will have just one instance of it.

updating/updating-disconnected-cluster.adoc
LalatenduMohanty: Not sure at this point. But in future we can look in to that.

updating/updating-disconnected-cluster.adoc
LalatenduMohanty: I agree with @smarterclayton  about having a bunch of shell scripts is the ideal user experience we want for our customers. I am on board with figuring out how we can get rid of the shell scripts and may be provide a better user experience in future.

updating/updating-disconnected-cluster.adoc
LalatenduMohanty: Thinking it again, I am inclined to use the  ${LOCAL_REGISTRY}/${LOCAL_REPOSITORY} here. Just to be more explicit in the communication. may be we can add note that users can use use quay.io here because of `ImageContentSourcePolicy`

LalatenduMohanty:
> This doesn't include what I expected to see from #17896 which is the terminology and flow changes (especially the removal of the atrocious registry creation). Is that a follow up I'm missing?

There was too much confusion around what docs are update vs what are for installation. So I started with this one. I am planning to pick some of the flow from the existing PR and add it here to keep it consistent. In the same time want to keep it to a size which can be relatively easier to merge and then do follow up PRs.

e-minguez:
Some comments:
* Shouldn't instructions on how to create an `ImageContentSourcePolicy` be added just in case? Or at least, refer to the section on the docs on how to do that.
* Shouldn't the ClusterVersion's .spec.channel be an empty string to tell the CVO to not bother attempting to fetch the graph?
* You can avoid using `--allow-explicit-upgrade` by creating a custom upgrade graph and referring to it in the ClusterVersion's .spec.upstream URL, so you can just `oc adm upgrade --to="4.X.Y"`

LalatenduMohanty:
> * Shouldn't instructions on how to create an `ImageContentSourcePolicy` be added just in case? Or at least, refer to the section on the docs on how to do that.
> 

Will point to the section on the docs on how to do that

> * Shouldn't the ClusterVersion's .spec.channel be an empty string to tell the CVO to not bother attempting to fetch the graph?

Good point. Need to confirm this and include. 

> * You can avoid using `--allow-explicit-upgrade` by creating a custom upgrade graph and referring to it in the ClusterVersion's .spec.upstream URL, so you can just `oc adm upgrade --to="4.X.Y"`

This will be addresses when users can run local Cincinnati instance and they point the CVO to get the graph information. So it will come as part of local Cincinnati docs.

updating/updating-disconnected-cluster.adoc
ncameronbritt: @LalatenduMohanty  Not specifically related to this PR, but seems like maybe we should be building a workflow around for whatever graph viewer we come up with in OCM. I.e. you tell us what version you're on; we recommend some options; you choose a version; and then we help you create whatever needs creating so that you can go off and do the mirroring/upgrade. cc @katherinedube

wking: I would [still][1] rather punt on target discovery, and have this PR scoped to the mechanics of post-discovery application.

[1]: https://github.com/openshift/openshift-docs/pull/21993#discussion_r423287574

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,135 @@
 [id="updating-disconnected-cluster"]
-= Updating a disconnected cluster to a minor version
+= Updating a disconnected cluster
```
wking, line 2: "disconnected" -> "restricted network", here and throughout the content you're touching?  Previous thread [here][1].  And I realize that these `disconnected` references exist in master, but we're uncommenting them here, so I think we should move to the standard wording while we uncomment.

[1]: https://github.com/openshift/openshift-docs/pull/21993#discussion_r423267306

modules/images-configuration-create-config-map.adoc
```diff
@@ -0,0 +1,60 @@
...
+The following commands will create and apply the image signature configmap to the cluster you want to update. 
+
+.Procedure
+
+. Find the link:https://access.redhat.com/solutions/4583231[version] you want the cluster to update.
+
+. Add the version to OCP_RELEASE_NUMBER environment variable.
++
+----
+$ export OCP_RELEASE_NUMBER=<{product-version}.0>
```
wking, line 13: [Still][1] no need to `export` this or other other variables you're using.  Although I see #22008 is still open, so you're still free to punt ;).  Also in this space is @smarterclayton's pushback [here][2] (which I attempted to rebut in my follow-up comment in that thread).

[1]: https://github.com/openshift/openshift-docs/pull/21993#discussion_r423298514
[2]: https://github.com/openshift/openshift-docs/pull/21993#discussion_r424620899

modules/images-configuration-create-config-map.adoc
```diff
@@ -0,0 +1,60 @@
...
+. Get the image signature from link:https://mirror.openshift.com/pub/openshift-v4/signatures/openshift/release[mirror.openshift.com] website.
++
+----
+$ export SIGNATURE_BASE64=$(curl -s "https://mirror.openshift.com/pub/openshift-v4/signatures/openshift/release/$(echo $DIGEST | cut -d: -f1)=$(echo $DIGEST | cut -d: -f2)/signature-1" | base64 -w0 && echo)
+----
++
+. Get the digest algorithm.
++
+----
+export DIGEST_ALGO=$(echo $DIGEST | cut -d: -f1)
```
wking, line 31: I [still][1] think this and `DIGEST_SIGNATURE` should be shifted up above the `curl` so you don't have to inline the `$(...)` command substitution in the `curl` line.

[1]: https://github.com/openshift/openshift-docs/pull/21993#discussion_r423305654

updating/updating-disconnected-cluster.adoc
wking: This thread has been replaced by [this new thread][1] and can be marked resolved.

[1]: https://github.com/openshift/openshift-docs/pull/21993#discussion_r437015337

updating/updating-disconnected-cluster.adoc
LalatenduMohanty: @wking I do not understand that point. You need to be more verbose :)

wking: You need [the label][1] you were setting in the here-doc.  I could not find a way to get `oc create configmap` to set labels.  Maybe I missed it?  Maybe we can add the label in post-processing with `-o json` and `jq`.  Maybe we want the here-doc and manual base64ing after all?

[1]: https://github.com/openshift/oc/blob/8dda2e784003f0bd16142a3c4a63924768bdd891/pkg/helpers/release/configmap.go#L28-L32

LalatenduMohanty: I reverted to here-doc.

modules/images-configuration-create-config-map.adoc
```diff
@@ -0,0 +1,61 @@
+[id="steps-for-manual-configmap-creation"]
+= Steps to create the image signature ConfigMap
+
+The following commands will create and apply the image signature ConfigMap to the cluster you want to update. 
+
+.Procedure
+
+. Find the link:https://access.redhat.com/solutions/4583231[version] you want the cluster to update.
```
jottofar, line 8: This sounds a bit awkward. How about something like: "Use the following solution article to determine a valid upgrade path for your cluster".
We should also mention that the following steps need to be ran for each upgrade target node in the path or explain that the steps are for doing a single upgrade of the path.

modules/images-configuration-create-config-map.adoc
```diff
@@ -0,0 +1,61 @@
...
+$ OCP_RELEASE_NUMBER=<{product-version}.0>
+----
++
+. Get the release image digest from link:https://quay.io/[Quay].
++
+----
+$ DIGEST="$(oc adm release info quay.io/openshift-release-dev/ocp-release:${OCP_RELEASE_NUMBER}-${ARCHITECTURE} | sed -n 's/Pull From: .*@//p')"
+----
++
+. Get the digest algorithm.
```
jottofar, line 22: nit but I think the next three "Get"s should be "Set"

modules/images-configuration-create-config-map.adoc
```diff
@@ -0,0 +1,61 @@
...
+The following commands will create and apply the image signature ConfigMap to the cluster you want to update. 
+
+.Procedure
+
+. Find the link:https://access.redhat.com/solutions/4583231[version] you want the cluster to update.
+
+. Add the version to OCP_RELEASE_NUMBER environment variable.
++
+----
+$ OCP_RELEASE_NUMBER=<{product-version}.0>
```
jottofar, line 13: This is implying that product-version should only be set to major.minor? If so we should explain that and perhaps even give an example. I just see it not being clear to some customers and resulting in support cases.

modules/images-configuration-create-config-map.adoc
```diff
@@ -0,0 +1,61 @@
...
+----
```
jottofar, line 14: Need step to set ARCHITECTURE and a ref back to the solution article, or something up front that keys them into what they are taking away from the solution article, so they know how to set it.

modules/images-configuration-create-config-map.adoc
wking: This will render as 4.5.0 for 4.5 docs, etc., and my impression is that is intended to be enough of an example for users to replace with 4.5.9, or whatever version they actually want to mirror.  Maybe a `For example:` at the end of the above paragraph would make that clearer?

jottofar: NM. It's been a while since I've looked at at our docs and I totally forgot about the macros.

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,145 @@
...
+= Updating a restricted network cluster
 include::modules/common-attributes.adoc[]
-:context: updating-disconnected-cluster
+:context: updating-restricted-network-cluster
 
 toc::[]
 
-You can update, or upgrade, an {product-title} cluster that you installed in a
-restricted network to a minor version by using the web console.
+You can update, or upgrade a restricted network {product-title} cluster using the `oc` command-line interface (CLI).
```
jottofar, line 8: Change to "upgrade" per our recent slack conversation

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,145 @@
...
+
+Before you perform the mirror procedure, you must prepare the host to retrieve content
+and push it to the remote location.
+
+include::modules/cli-installing-cli.adoc[leveloffset=+2]
+
+include::modules/installation-adding-mirror-registry-pull-secret.adoc[leveloffset=+1]
+
+[IMPORTANT]
+===
```
jottofar, line 36: I think this and closing delimiter should be `====`

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,145 @@
...
+Before you update a cluster on infrastructure that you provision in a restricted network, you must mirror the required container images into that environment. You may also use this procedure in unrestricted networks to ensure your clusters only use container images that have satisfied your organizational controls on external content.
+
+.Procedure
+
+. Find out if there is link:https://access.redhat.com/solutions/4583231[an update path] that exists between your current version and the desired version for the cluster to move.
+
+. Set the required environment variables:
++
+----
+$ OCP_RELEASE=<release_version> <1>
```
jottofar, line 54: These are coming out a `i`, `ii`, etc. Believe you need `# <1>`

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,145 @@
...
+$ LOCAL_REGISTRY='<local_registry_host_name>:<local_registry_host_port>' <2>
+$ LOCAL_REPOSITORY='<repository_name>' <3>
+$ PRODUCT_REPO='openshift-release-dev' <4>
+$ LOCAL_SECRET_JSON='<path_to_pull_secret>' <5>
+$ RELEASE_NAME='ocp-release' <6>
+$ ARCHITECTURE=<server_architecture> <7>
+----
+<1> For `<release_version>`, specify the tag that corresponds to the version of {product-title} you want
```
jottofar, line 62: `to which you want to upgrade`

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,145 @@
...
+$ oc adm release mirror "--to-dir=${REMOVABLE_MEDIA_PATH}/mirror" quay.io/openshift-release-dev/${RELEASE_NAME}:${OCP_RELEASE}-${ARCHITECTURE} --dry-run
+----
++
+----
+$ oc adm release mirror "--to-dir=${REMOVABLE_MEDIA_PATH}/mirror" quay.io/openshift-release-dev/${RELEASE_NAME}:${OCP_RELEASE}-${ARCHITECTURE}
+----
++
+[NOTE]
+====
+If the local container registry and the cluster connected to the mirror host, you can directly push the release images to the local registry and apply the ConfigMap to the cluster using following command.
```
jottofar, line 92: and the cluster are connected

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,145 @@
...
++
+----
+$ oc adm release mirror "--to-dir=${REMOVABLE_MEDIA_PATH}/mirror" quay.io/openshift-release-dev/${RELEASE_NAME}:${OCP_RELEASE}-${ARCHITECTURE} --dry-run
+----
++
+----
+$ oc adm release mirror "--to-dir=${REMOVABLE_MEDIA_PATH}/mirror" quay.io/openshift-release-dev/${RELEASE_NAME}:${OCP_RELEASE}-${ARCHITECTURE}
+----
++
+[NOTE]
```
jottofar, line 90: Also need to note need for 4.5.0 oc here as below since it's required for use of `--apply-release-image-signature`

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,145 @@
...
++
+** Take the media to the restricted network environment and upload the images to the local container registry.
++
+----
+$ oc image mirror  -a ${LOCAL_SECRET_JSON} "--from-dir=${REMOVABLE_MEDIA_PATH}/mirror" file://openshift/release:${OCP_RELEASE}* ${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}
+----
+
+=== Create the ConfigMap for image signature verification
+
+The Cluster Version Operator (CVO) in the cluster performs a verification to make sure the signatures of the release images are not tampered. To do that it compares expected VS actual image signatures. For the verification process it needs to know the image signatures. The image signatures are available in link:https://mirror.openshift.com/pub/openshift-v4/signatures/openshift/release[mirror.openshift.com]  and link:https://storage.googleapis.com/openshift-release/official/signatures[Google Cloud Storage (GCS)]. However in a restricted network environment we do not have access to internet, so we we need to create ConfigMap which CVO can use to get the image signatures.
```
jottofar, line 107: to the internet, so we need

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,145 @@
...
+
+[NOTE]
+====
+If you are upgrading from an earlier release than 4.5.0, you must must use the manual method for creating the ConfigMap.
```
jottofar, line 111: Not sure if there is a doc's standard for this but why don't we link to the manual method adoc here and remove the 'Manual method...' section below.

modules/images-configuration-create-config-map.adoc
LalatenduMohanty: Added the architecture and a note, but did not added much info around what they need from the KB article as this is a module and will be called from the update doc which would have the context.  Also the KB article has some information.

updating/updating-disconnected-cluster.adoc
LalatenduMohanty: The reason I kept it separate because most likely less customers need manual method of creating the configmap as we have back ported the feature to 4.4.8. So want to keep the less text for default steps.

updating/updating-disconnected-cluster.adoc
LalatenduMohanty: Added the note before the installing OC CLI module

updating/updating-disconnected-cluster.adoc
LalatenduMohanty: I tried to fix it but still getting issues. I will let doc team fix it :)

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,150 @@
...
+`ocp-release`.
+<7> For `server_architecture`, specify the architecture of the server e.g. x86_64.
+
+. Mirror the version images to the internal container registry.
+
+** Connect the removable media to a system connected to the internet.
+** Mirror the images and configuration manifests to a directory on the removable media.
++
+----
+$ oc adm release mirror "--to-dir=${REMOVABLE_MEDIA_PATH}/mirror" quay.io/openshift-release-dev/${RELEASE_NAME}:${OCP_RELEASE}-${ARCHITECTURE} --dry-run
```
jianlinliu, line 88: Should we replace "quay.io/openshift-release-dev/${RELEASE_NAME}:${OCP_RELEASE}-${ARCHITECTURE}" to "quay.io/${PRODUCT_REPO}/${RELEASE_NAME}:${OCP_RELEASE}-${ARCHITECTURE}" ?

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,150 @@
...
+----
++
+----
+$ oc adm release mirror "--to-dir=${REMOVABLE_MEDIA_PATH}/mirror" quay.io/openshift-release-dev/${RELEASE_NAME}:${OCP_RELEASE}-${ARCHITECTURE}
```
jianlinliu, line 92: Should we replace "quay.io/openshift-release-dev/${RELEASE_NAME}:${OCP_RELEASE}-${ARCHITECTURE}" to "quay.io/${PRODUCT_REPO}/${RELEASE_NAME}:${OCP_RELEASE}-${ARCHITECTURE}"?

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,150 @@
...
+----
++
+[NOTE]
+====
+If the local container registry and the cluster are connected to the mirror host, you can directly push the release images to the local registry and apply the ConfigMap to the   cluster using following command.
+----
+$ oc adm release mirror --from=quay.io/openshift-release-dev/${RELEASE_NAME}:${OCP_RELEASE}-${ARCHITECTURE} \
+  --to=${LOCAL_REGISTRY}/${LOCAL_REPOSITORY} --apply-release-image-signature
+----
```
jianlinliu, line 101: Suggest to add one more sentence:
Once that, skip the following setps for ConfigMap signiature creation, directly perform the upgrade.

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,150 @@
...
+
+include::modules/cli-installing-cli.adoc[leveloffset=+2]
+
+include::modules/installation-adding-mirror-registry-pull-secret.adoc[leveloffset=+1]
+
+[IMPORTANT]
+====
+This step requires that you have write access to a container image registry on the
+mirror registry and adding those credentials to a registry pull secret. This secret
+should NOT be used to install a cluster as it would give machines within the cluster
```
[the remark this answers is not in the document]
LalatenduMohanty: Thats right.

updating/updating-disconnected-cluster.adoc
LalatenduMohanty: I copied these from some other PRs, so trying to keep these consistent. If you do not have strong opinions we should go ahead with it.

updating/updating-disconnected-cluster.adoc
LalatenduMohanty: https://github.com/openshift/openshift-docs/pull/21993/files#r444313036

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,150 @@
...
++
+----
+$ oc adm release mirror "--to-dir=${REMOVABLE_MEDIA_PATH}/mirror" quay.io/openshift-release-dev/${RELEASE_NAME}:${OCP_RELEASE}-${ARCHITECTURE} --dry-run
+----
++
+----
+$ oc adm release mirror "--to-dir=${REMOVABLE_MEDIA_PATH}/mirror" quay.io/openshift-release-dev/${RELEASE_NAME}:${OCP_RELEASE}-${ARCHITECTURE}
+----
++
+[NOTE]
```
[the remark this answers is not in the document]
LalatenduMohanty: Thanks for pointing out. Will fix it.

updating/updating-disconnected-cluster.adoc
LalatenduMohanty: Thanks, added `If you are using `--apply-release-image-signature` you do not need to create the ConfigMap for image signature verification.`

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,150 @@
...
+$ oc image mirror  -a ${LOCAL_SECRET_JSON} "--from-dir=${REMOVABLE_MEDIA_PATH}/mirror" file://openshift/release:${OCP_RELEASE}* ${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}
+----
+
+=== Create the ConfigMap for image signature verification
+
+The Cluster Version Operator (CVO) in the cluster performs a verification to make sure the signatures of the release images are not tampered. To do that it compares expected VS actual image signatures. For the verification process it needs to know the image signatures. The image signatures are available in link:https://mirror.openshift.com/pub/openshift-v4/signatures/openshift/release[mirror.openshift.com]  and link:https://storage.googleapis.com/openshift-release/official/signatures[Google Cloud Storage (GCS)]. However in a restricted network environment we do not have access to the internet, so we need to create ConfigMap which CVO can use to verify the image signatures.
+
+[NOTE]
+====
+If you are upgrading from an earlier release than 4.4.8, you must must use the manual method for creating the ConfigMap.
```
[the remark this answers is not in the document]
LalatenduMohanty: Fixed

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,150 @@
...
+$ oc adm upgrade --allow-explicit-upgrade --to-image ${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}@sha256:81154f5c03294534e1eaf0319bef7a601134f891689ccede5d705ef659aa8c92
+----
+[NOTE]
+====
+You can use ImageContentSourcePolicy if you do want to use canonical registry name instead of LOCAL_REGISTRY.
+====
+
+include::modules/images-configuration-registry-mirror.adoc[leveloffset=+2]
+
+== Manual method to create the image signature configmap
```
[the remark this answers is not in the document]
LalatenduMohanty: Fixed.

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,150 @@
...
+$ oc adm release mirror --from=quay.io/openshift-release-dev/${RELEASE_NAME}:${OCP_RELEASE}-${ARCHITECTURE} \
+  --to=${LOCAL_REGISTRY}/${LOCAL_REPOSITORY} --apply-release-image-signature
+----
+====
++
+** Take the media to the restricted network environment and upload the images to the local container registry.
++
+----
+$ oc image mirror  -a ${LOCAL_SECRET_JSON} "--from-dir=${REMOVABLE_MEDIA_PATH}/mirror" file://openshift/release:${OCP_RELEASE}* ${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}
+----
```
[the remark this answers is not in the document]
LalatenduMohanty: the wild card character `*` will take care of it. These commands are tested by us and it works so kept it that way as the chance of user using same directory for another architecture  is very remote and even if they do it wont cause any issue.

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,150 @@
...
+.Procedure
+
+. Use `oc` command-line interface (CLI) to log in to a given cluster which going to be upgraded.
+
++
+. Apply the the mirrored release image signature ConfigMap to the connected cluster
++
+----
+$ oc apply -f mirror/config/signature-sha256-81154f5c03294534.yaml
+----
```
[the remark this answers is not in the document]
LalatenduMohanty: `images-configuration-create-config-map.adoc` is for a different purpose so it can not be linked from here.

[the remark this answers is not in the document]
LalatenduMohanty: I added `Example:` here.

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,150 @@
...
+
+=== Perform the upgrade
+
+.Procedure
+
+. Perform the upgrade
++
+----
+$ oc adm upgrade --allow-explicit-upgrade --to-image ${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}@sha256:81154f5c03294534e1eaf0319bef7a601134f891689ccede5d705ef659aa8c92
+----
```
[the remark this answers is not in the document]
LalatenduMohanty: I could not find any way to communicate this better. May be we can improve this later (i.e. post the PR merge). The basic documentation is missing from official docs for long. Lets get some decent docs first and then improve on these.

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,150 @@
...
+[NOTE]
+====
+You can use ImageContentSourcePolicy if you do want to use canonical registry name instead of LOCAL_REGISTRY.
```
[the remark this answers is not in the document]
LalatenduMohanty: The idea is to communicate that if you are using `ImageContentSourcePolicy` you can use canonical registry names and I believe it communicates that. let me know if you disagree.

modules/images-configuration-create-config-map.adoc
```diff
@@ -0,0 +1,70 @@
...
++
+----
+$ DIGEST_ENCODED="${DIGEST#*:}"
+----
++
+. Get the image signature from link:https://mirror.openshift.com/pub/openshift-v4/signatures/openshift/release[mirror.openshift.com] website.
++
+
+----
+$ SIGNATURE_BASE64=$(curl -s "https://mirror.openshift.com/pub/openshift-v4/signatures/openshift/release/$(echo $DIGEST | cut -d: -f1)=$(echo $DIGEST | cut -d: -f2)/signature-1" | base64 -w0 && echo)
```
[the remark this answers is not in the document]
LalatenduMohanty: Thanks for the suggestion.

updating/updating-disconnected-cluster.adoc
jianlinliu: I meant should we replace the hardcode - 'openshift-release-dev' with the prior defined PRODUCT_REPO var?

updating/updating-disconnected-cluster.adoc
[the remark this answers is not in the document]
jianlinliu: Maybe my previous comment is not clear enough. When user mirror release image to a different PRODUCT_REPO for this upgrade with the source PRODUCT_REPO for disconnected install, then user have to capture `ImageContentSourcePolicy` from --dry-run output, and create it before upgrade:
https://github.com/openshift/openshift-docs/pull/21993/files#diff-8afabd3ee5993f3a7b9595c12f39a3baR88

For example:
1. mirror release image to 'release-4.3/ocp' in the container registry
2. run a disconnected install using the 'release-4.3/ocp' release image.
3. mirror release image to 'release-4.4/ocp' in the same container registry for disconnected cluster upgrade

In this case, user have to create `ImageContentSourcePolicy` before upgrade.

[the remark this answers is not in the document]
LalatenduMohanty: > In this case, user MUST create `ImageContentSourcePolicy` before upgrade, or else, upgrade would fail.

You mean it will fail if user used quay.io as registry name right?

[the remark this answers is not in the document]
LalatenduMohanty: The idea is to capture the good path in the doc and leave enough pointers to users to address their workflow. Also PRs are welcome if you want to improve the doc post this PR merge.

updating/updating-disconnected-cluster.adoc
LalatenduMohanty: I am not sure. I do not see this changing for OpenShift 4. Also reducing environment variables is helpful as it is reduces the hassle for customers. @wking @jottofar WDYT?

wking: I want this to eventually be "the by-digest pullspec you extracted from a Cincinnati graph, errata text, or whatever".  I am agnostic about what we do in the meantime.

updating/updating-disconnected-cluster.adoc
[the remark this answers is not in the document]
LalatenduMohanty: @wking WDYT?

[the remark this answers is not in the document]
jianlinliu: > You mean it will fail if user used quay.io as registry name right?
As far as I know, this is not related to registry name, but PRODUCT_REPO path.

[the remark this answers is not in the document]
wking: Folks will need an ICSP so their cluster knows to check the mirror(s) to find the quay.io referenced images.  They may already have a sufficient ICSP in place, and we have no generic tooling to help users decide how/when to merge, dedup, and retire ICSPs.  I'm fine calling out tge "if you mirror all releases to the same place, you don't need to update your original ICSP" case.  But I don't think we want to go very far down the "I mirror each release to a different place" rabbit hole.

updating/updating-disconnected-cluster.adoc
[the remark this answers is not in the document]
LalatenduMohanty: Also `$ oc adm release mirror -a /home/lmohanty/.pull-secret --from=quay.io/openshift-release-dev/ocp-release@sha256:0d1ffca302ae55d32574b38438c148d33c2a8a05c8daf97eeb13e9ab948174f7 --to-dir='/home/lmohanty/kubec
onfigs/mirror/' --apply-release-image-signature` output prints the below command as suggestion

updating/updating-disconnected-cluster.adoc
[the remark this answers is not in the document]
LalatenduMohanty: Yeah it is tricky to explain it and at the same time not fall in to the trap of rabbit hole of explain how user needs to adjust the `ImageContentSourcePolicy`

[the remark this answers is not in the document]
LalatenduMohanty: @jianlinliu I have made some changes, also tried the steps at https://gist.github.com/LalatenduMohanty/dcecc4dda52fcce6d17b4a6ef1249043 . Let me know if it looks fine to you.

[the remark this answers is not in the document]
jianlinliu: In QE's testing, we never use OpenShift Registry to host mirror release image. A registry running in a OCP cluster, which maybe also need a upgrade using a mirror release image hosted in itself, sound like not productive enough or not a customer case. 
Beside that, the steps in the gist looks good to me (Actually I did not find many difference with the existing steps in this doc PR).

LalatenduMohanty:
@wking @jottofar @jiajliu @jianlinliu All your comments are resolved. Let me me know if you still have comments for this PR.

LalatenduMohanty:
@kalexand-rh  can you please take a look? we need this docs for 4.5.

kalexand-rh:
@LalatenduMohanty, thanks for letting me know that you'd resolved all the comments! I will push another commit to your branch with some edits as soon as I can get them together.

updating/updating-disconnected-cluster.adoc
```diff
@@ -1,23 +1,157 @@
...
+$ oc adm upgrade --allow-explicit-upgrade --to-image ${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}@sha256:81154f5c03294534e1eaf0319bef7a601134f891689ccede5d705ef659aa8c92
+
+The sha256sum value can be found in the image signature ConfigMap in the previous step.
+----
+[NOTE]
+====
+If you are using ImageContentSourcePolicy for the mirror registry you can use the canonical registry name instead of LOCAL_REGISTRY.
+====
+
+include::modules/images-configuration-registry-mirror.adoc[leveloffset=+2]
```
kalexand-rh, line 153: Is this included module in the right place?

kalexand-rh:
@LalatenduMohanty @wking @jottofar @jiajliu @jianlinliu
Here's a built draft: http://file.rdu.redhat.com/kalexand/063020/OTA-102/updating/updating-restricted-network-cluster.html

Do any of you see any other changes?

modules/installation-mirror-repository.adoc
```diff
@@ -38,9 +38,10 @@ $ export LOCAL_REPOSITORY='<repository_name>' <3>
 $ export PRODUCT_REPO='openshift-release-dev' <4>
 $ export LOCAL_SECRET_JSON='<path_to_pull_secret>' <5>
 $ export RELEASE_NAME="ocp-release" <6>
+$ export ARCHITECTURE=<server_architecture> <7>
```
kalexand-rh, line 41: @jianlinliu, since this variable's split for upgrade, does it also need to be split for install?

jianlinliu: To keep consistent, yeah, it also can be split for install.

updating/updating-restricted-network-cluster.adoc
```diff
@@ -0,0 +1,50 @@
+[id="updating-restricted-network-cluster"]
+= Updating a restricted network cluster
+include::modules/common-attributes.adoc[]
+:context: updating-restricted-network-cluster
+
+toc::[]
+
+You can upgrade a restricted network {product-title} cluster by using the `oc` command-line interface (CLI).
+
+A restricted network environment is the one in which your cluster nodes cannot access the internet. For this reason you must mirror the images to a file system that disconnected from that environment and then bring that host or removable media across that gap.
```
jianlinliu, line 10: "Mirror the image to a file system, then bring that host or removable media across that gap" is only one case, to be more accurate, we also need to take "directly push the image to the local container registry" into consideration (refer to "If the local container registry and the cluster are connected to the mirror host, directly push the release images to the local registry and apply the ConfigMap to the cluster by using following command" section).

jianlinliu:
> Let me me know if you still have comments for this PR. 
Though I still think need more instruction about how to capture ImageContentSourcePolicy and apply it in this disconnected cluster upgrade doc (As a contrast, ImageContentSourcePolicy is mentioned for install), if you think it is tricky to explain it, I can let it go and wait for real customer's feedback as future enhancement.

jianlinliu:
@kalexand-rh BW, shall we link this into 4.5 Release Note? From https://docs.openshift.com/container-platform/4.5/release_notes/ocp-4-5-release-notes.html#ocp-4-5-installation-and-upgrade, I did not see any section to mention this improvement.

kalexand-rh:
> > Let me me know if you still have comments for this PR.
> 
> Though I still think need more instruction about how to capture ImageContentSourcePolicy and apply it in this disconnected cluster upgrade doc (As a contrast, ImageContentSourcePolicy is mentioned for install). If you think it is tricky to explain it, I can let it go and wait for real customer's feedback as future enhancement.

@jianlinliu, I'd rather wait for customer feedback to address that enhancement.

I opened a PR to add this doc to the release notes: https://github.com/openshift/openshift-docs/pull/23404

And I pushed a few last changes. With those caveats, is this PR ready to merge?

LalatenduMohanty:
> Do any of you see any other changes?

@katherinedube Looks [1] looks very good to me. Thanks for fixing the PR. 
[1] http://file.rdu.redhat.com/kalexand/063020/OTA-102/updating/updating-restricted-network-cluster.html

updating/updating-disconnected-cluster.adoc
LalatenduMohanty: As per http://file.rdu.redhat.com/kalexand/063020/OTA-102/updating/updating-restricted-network-cluster.html I think it is in the right place. Also we can fix it later if we get feedback from users.

LalatenduMohanty:
>  @jianlinliu, I'd rather wait for customer feedback to address that enhancement.

+1. I agree

kalexand-rh:
Thanks @LalatenduMohanty. If Jianlin approves the last changes, it should go in tomorrow. Which versions should I add these changes to?

LalatenduMohanty:
> @kalexand-rh @LalatenduMohanty there is a pending RN submission for updating a disconnected cluster: #23258. Let me know if I should delete this PR in favor of the new one. I've been waiting for feedback on it before merging.

Sorry missing out the PR. I have added comments to #23258

modules/installation-mirror-repository.adoc
```diff
@@ -52,14 +53,16 @@ registry, such as `ocp4/openshift4`.
...
 
 . Mirror the repository:
 +
 ----
 $ oc adm -a ${LOCAL_SECRET_JSON} release mirror \
-     --from=quay.io/${PRODUCT_REPO}/${RELEASE_NAME}:${OCP_RELEASE} \
+     --from=quay.io/${PRODUCT_REPO}/${RELEASE_NAME}:${OCP_RELEASE}-${ARCHITECTURE} \
      --to=${LOCAL_REGISTRY}/${LOCAL_REPOSITORY} \
-     --to-release-image=${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}:${OCP_RELEASE}
+     --to-release-image=${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}:${OCP_RELEASE}-${ARCHITECTURE}
```
jianlinliu, line 65: @kalexand-rh The following step 5 missed ${ARCHITECTURE}

kalexand-rh: Fixed. Thanks!

kalexand-rh:
@jianlinliu, I know this change is valid for 4.5. Which earlier versions can I backport this change to?

LalatenduMohanty:
> @jianlinliu, I know this change is valid for 4.5. Which earlier versions can I backport this change to?

We should backport to 4.4. As this feature is available from 4.4.8.

jianlinliu:
LGTM

kalexand-rh:
Per https://bugzilla.redhat.com/show_bug.cgi?id=1828214#c14 and an ongoing discussion on Slack, this change also needs to be ported to 4.3.

kalexand-rh:
@jianlinliu, do you have any objection to applying this change to 4.3? @LalatenduMohanty  says fix is in 4.3.12.

jianlinliu:
This is a big PR. Some pieces of process does not work in 4.3. Such as: oc mirror payload to disk scenario. But 4.3.13 indeed support user manually create signature configmap, then apply it to existing clusters, then trigger disconnected cluster upgrade.

LalatenduMohanty:
@kalexand-rh  For versions in 4.3.z (i.e. after 4.3.12) will not get the the ConfigMap created by the `oc adm release mirror` command. So they can not apply the ConfigMap directly i.e. `oc apply -f <image_signature_file>` [1] post the `oc adm release mirror` command. User need to create the ConfigMap manually using the steps given in the document and then apply it. 

[1] https://docs.openshift.com/container-platform/4.5/updating/updating-restricted-network-cluster.html#update-oc-configmap-signature-verification_updating-restricted-network-cluster

PR #22006, opened 2020-05-11
updating 4.2 AWS permissions

kalexand-rh:
Updating to match https://github.com/openshift/installer/blob/release-4.2/pkg/asset/installconfig/aws/permissions.go

@jianlinliu, will you PTAL?


jianlinliu:
LGTM

PR #22008, opened 2020-05-11
modules/installation-mirror-repository: Drop unnecessary 'export'

wking:
The export landed with the module in ced98c67ad (#16678).  But these are local variables that we expand when calling the mirror command.  The mirror command receives them from its command line arguments; we don't need [`export`][1] to expose them in the mirror command's environment as well.

/assign @kalexand-rh

[1]: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#export

### DIFF
--- a/modules/installation-mirror-repository.adoc
+++ b/modules/installation-mirror-repository.adoc
@@ -32,12 +32,12 @@ to determine the version of {product-title} that you want to install and determi
 . Set the required environment variables:
 +
 ----
-$ export OCP_RELEASE=<release_version> <1>
-$ export LOCAL_REGISTRY='<local_registry_host_name>:<local_registry_host_port>' <2>
-$ export LOCAL_REPOSITORY='<repository_name>' <3>
-$ export PRODUCT_REPO='openshift-release-dev' <4>
-$ export LOCAL_SECRET_JSON='<path_to_pull_secret>' <5>
-$ export RELEASE_NAME="ocp-release" <6>
+$ OCP_RELEASE=<release_version> <1>
+$ LOCAL_REGISTRY='<local_registry_host_name>:<local_registry_host_port>' <2>
+$ LOCAL_REPOSITORY='<repository_name>' <3>
+$ PRODUCT_REPO='openshift-release-dev' <4>
+$ LOCAL_SECRET_JSON='<path_to_pull_secret>' <5>
+$ RELEASE_NAME="ocp-release" <6>
 ----
 <1> For `<release_version>`, specify the tag that corresponds to the version of {product-title} to
 install for your architecture, such as `4.4.0-x86_64`.


kalexand-rh:
@wking, is this true back to 4.2?

@jiajliu, will you PTAL?

jiajliu:
If no need export the variables, then we may update the mirror command with the local variables instead of env variables in "Mirror the repository" step too.
```
oc adm -a ${LOCAL_SECRET_JSON} release mirror \
     --from=quay.io/${PRODUCT_REPO}/${RELEASE_NAME}:${OCP_RELEASE} \
     --to-release-image=${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}:${OCP_RELEASE}
```
hdyt? @wking

wking:
> If no need export the variables, then we may update the mirror command...

The example you have is what [we have now in master][1].  How did you want me to change it?  Here's an example showing local variables expanded by the shell before they are fed to a subcommand, in case that helps clarify:

```console
$ A=whatever
$ python -c 'import os, sys; print(sys.argv, [k for k in os.environ.keys() if "A" in k])' "${A}"
(['-c', 'whatever'], [..., 'PATH', ...])
```

The `${A}` is expanded to `whatever` by the shell, but `A` is not an environment variable in the Python process.

[1]: https://github.com/openshift/openshift-docs/blame/428118501b18a9a81061f4ad3ae254bcbf83a391/modules/installation-mirror-repository.adoc#L59-L62

jiajliu:
I mean, if we want to keep all variables for long-time usage, then better to set env variables permanently. And if we update it to use local variables in one-time shell, then we can just use the command with correct variables. Maybe no need to set local variables before it and then call it.
Now:
```
 OCP_RELEASE=<release_version> 
 LOCAL_REGISTRY='<local_registry_host_name>:<local_registry_host_port>' 
 LOCAL_REPOSITORY='<repository_name>' 
 PRODUCT_REPO='openshift-release-dev' 
 LOCAL_SECRET_JSON='<path_to_pull_secret>' 
 RELEASE_NAME="ocp-release" 

oc adm -a ${LOCAL_SECRET_JSON} release mirror \
     --from=quay.io/${PRODUCT_REPO}/${RELEASE_NAME}:${OCP_RELEASE} \
     --to-release-image=${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}:${OCP_RELEASE}
```
I meant:
```
oc adm -a <path_to_pull_secret> release mirror \
     --from=quay.io/<repository_name>/ocp-release:<release_version> \
     --to=<local_registry_host_name>:<local_registry_host_port>/<repository_name>\
     --to-release-image=<local_registry_host_name>:<local_registry_host_port>/<repository_name>:<release_version> 
```
Anyway, it's not a big deal for this part, i think users can understand either, so it is ok for me to merge the pr without any changes.

kalexand-rh:
@jiajliu, @wking, any objections to applying this change back to 4.2?

wking:
> I mean, if we want to keep all variables for long-time usage, then better to set env variables permanently(not just export in the shell).

Folks _can_ do that.  But I don't think we want to get into that level of detail in the docs.  Folks who are interested in declaring these in their `~/.bashrc` or whatever will already understand that they can add an `export ...` for that sort of thing.

> And if we update it to use local variables in one-time shell, then we can just use the command with correct variables.

Isn't that what I'm doing?  I think declaring them on separate lines like we've been doing gives space for the `<1>` and such that would make a single line that plugged in a bunch of placeholders complicated.  Also, shell variables are already an established syntax for placeholders.  I'd rather use them and not our ad-hoc `<...>` placeholders.  Like:

```console
$ oc adm -a "${LOCAL_SECRET_JSON}" release mirror \
     --from="quay.io/${PRODUCT_REPO}/${RELEASE_NAME}:${OCP_RELEASE}" \
     --to="${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}" \
     --to-release-image="${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}:${OCP_RELEASE}"
```

and then a section that walked through:

* `${LOCAL_SECRET_JSON}` is a...

etc.  But that sort of change seemed like a bigger pivot than I want to argue through, so I'm just focusing on the `export` ;).

> ... any objections to applying this change back to 4.2?

I am agnostic about any backporting.  "No backporting at all", "backporting through all the versions", and anything in between are all fine with me.

PR #22017, opened 2020-05-12
Migration: organizational changes

apinnick:
I renamed the "migration application workloads..." assembly to "understanding CAM" because I thought it was a better description and less repetitious. 

Moved "understanding data copy methods" to "understanding CAM" assembly because it made more sense to put it in assembly dealing with concepts.

Moved CAM deployment before replication repository. It seemed to make more sense that way.

Moved "launching CAM" to CAM deployment.

CP for 4.2, 4.3, 4.4, 4.5


bergerhoffer:
@apinnick what version(s) is this for?

If you're changing file and anchor names for an existing published version, we'll probably need to coordinate with @vikram-redhat to set up redirects.

vikram-redhat:
Hoping this is for a future release so we don't require redirects :) 

But, yes, if this is for a published release, this will definitely require redirects as this is popular content.

apinnick:
> Hoping this is for a future release so we don't require redirects :)
> 
> But, yes, if this is for a published release, this will definitely require redirects as this is popular content.

@vikram-redhat I have added the old and new links to the [Redirects Register](https://docs.google.com/spreadsheets/d/1KH2_p54Fr2KRFTUGzQdpqExox7iL9RM4_yiPNQpArjk/edit#gid=0)

migration/migrating_3_4/understanding-migration-toolkit.adoc
```diff
@@ -1,5 +1,5 @@
-[id="migrating-application-workloads-3-to-4"]
-= Migrating application workloads from {product-title} 3.7 to {product-version}
+[id="understanding-migration-toolkit_{context}"]
+= Understanding the migration toolkit
```
bergerhoffer, line 2: Just an observation that I don't see any other references to this set of tools as the "migration toolkit", so just wondering if we should be generically saying that. Also, this reminds me of the "Red Hat Application Migration Toolkit": https://access.redhat.com/documentation/en-us/red_hat_application_migration_toolkit/

modules/migration-adding-replication-repository-to-cam.adoc
```diff
@@ -3,7 +3,7 @@
 // migration/migrating_3_4/migrating-applications-with-cam.adoc
 // migration/migrating_4_1_4/migrating-applications-with-cam.adoc
 // migration/migrating_4_2_4/migrating-applications-with-cam.adoc
-[id='migration-adding-replication-repository-to-cam_{context}']
+[id='adding-replication-repository-to-cam_{context}']
```
bergerhoffer, line 6: Typically we do like the anchor to match the file name, so I'd recommend leaving this as it was. Same for other instances like this in the PR.

_topic_map.yml
```diff
@@ -1285,12 +1285,12 @@ Topics:
   Dir: migrating_4_2_4
```
bergerhoffer, line 1285: I stumbled across this typo in the above line (it's outside your PR changes so it wouldn't let me comment on line 1284) if you wouldn't mind fixing OpenShift here: `Migrating Openshift Container Platform 4.2 to 4.3`

bergerhoffer (commented):
Sorry for the delay @apinnick! Just a few comments.

bergerhoffer:
@apinnick this PR has some build errors if you can check them out:

```
--------------------------------------

>>> Working on migration book in openshift-enterprise <<<

Transforming the AsciiDoc content to DocBook XML...

Unknown ID or title "migration-understanding-cpma_migrating-3-4", used as an internal cross reference

Unknown ID or title "migration-configuring-aws-s3_migrating-3-4", used as an internal cross reference

Unknown ID or title "migration-configuring-aws-s3_migrating-4_1-4_x", used as an internal cross reference

Unknown ID or title "migration-configuring-aws-s3_migrating-4_2-4_x", used as an internal cross reference

>>> Validation of book migration in openshift-enterprise failed <<<

>>> Finished with migration book in openshift-enterprise <<<

---------------------------------------
```

And then I still just need to know which versions this applies to.

_topic_map.yml
```diff
@@ -1255,12 +1255,12 @@ Topics:
...
+    File: understanding-migration-toolkit
+  - Name: Deploying the Cluster Application Migration tool
+    File: deploying-cam
   - Name: Configuring a replication repository
     File: configuring-replication-repository
-  - Name: Deploying the Cluster Application Migration tool
-    File: deploying_cam
   - Name: Migrating applications with the CAM web console
     File: migrating-applications-with-cam
   - Name: Migrating control plane settings with the Control Plane Migration Assistant
```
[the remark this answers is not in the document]
apinnick: Oops! Thanks for catching that!

migration/migrating_3_4/configuring-replication-repository.adoc
```diff
@@ -10,13 +10,11 @@ You must configure an object storage to use as a replication repository. The Clu
 
 The following storage providers are supported:
 
+* xref:migration-configuring-mcg_{context}[Multi-Cloud Object Gateway (MCG)]
```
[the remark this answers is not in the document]
apinnick: OK, thanks. I wasn't sure about that.

migration/migrating_3_4/migrating-with-cpma.adoc
```diff
@@ -1,12 +1,13 @@
 [id='migrating-with-cpma_{context}']
-= Migrating control plane settings with the Control Plane Migration Assistant (CPMA)
+= Analyzing control plane settings for migration
```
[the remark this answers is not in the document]
apinnick: Correct. CPMA generates a support and can create some of the CR manifest files from 3.x configurations, but it doesn't actually migrate the settings. The actual migration of settings is done manually. Most of CPMA is the generation of a very nice report that shows which configs are supported in 4.x.

migration/migrating_3_4/migrating-with-cpma.adoc
```diff
@@ -1,12 +1,13 @@
...
 include::modules/common-attributes.adoc[]
 :context: migrating-3-4
 :migrating-3-4:
 
+You can analyse your control plane settings with the Control Plane Migration Assistant (CPMA). The CPMA is a CLI-based tool that assists you in migrating the control plane from {product-title} 3.x to {product-title} {product-version}. The CPMA processes the {product-title} 3 configuration files and generates Custom Resource (CR) manifest files, which are consumed by the {product-title} {product-version} Operators.
```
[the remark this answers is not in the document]
apinnick: Oops, me bad. Bad Canadian. Thanks for catching.

migration/migrating_3_4/understanding-migration-toolkit.adoc
apinnick: It's a gradual move toward the rebranding. I have another branch in progress. CAM is going to become the "Migration Toolkit for Containers". RHAMT will be MTA and CNV (OpenShift virtualization)/IMS will be MTV.

In this context, the term 'toolkit' is fairly generic, so it seemed OK to me. Official rebranding date is Aug. 10, not far off.

modules/migration-adding-replication-repository-to-cam.adoc
apinnick: OK. I started removing them because the links had "_migration_migration-3-4", etc., which looked odd to me. I'll change them back.

bergerhoffer:
Hi @apinnick I'll re-review once you let me know what versions this is for please. And it looks like there are conflicts that need to be resolved. Thanks!

apinnick:
> Hi @apinnick I'll re-review once you let me know what versions this is for please. And it looks like there are conflicts that need to be resolved. Thanks!

@bergerhoffer Yes, it had to be rebased. Had a couple other PRs merged today.

I added versions in the PR description. Thanks!

migration/migrating_3_4/understanding-migration-toolkit.adoc
bergerhoffer: Okay cool, just wanted to make sure RHAMT was being considered in this. Glad to hear that there is a plan for it all!

migration/migrating_3_4/configuring-replication-repository.adoc
```diff
@@ -1,22 +1,20 @@
-[id='configuring-replication-repository_{context}']
+[id='configuring-replication-repository_migrating-3-4']
```
bergerhoffer, line 1: I can't speak for sure because I'm not familiar with nesting assemblies, but I think that this _{context} should actually stay here. It's only when you're actually using an xref to reference that it should be resolved out.

migration/migrating_3_4/deploying-cam.adoc
```diff
@@ -1,12 +1,14 @@
-[id='deploying-cam_{context}']
+[id='deploying-cam_migrating-3-4']
```
bergerhoffer, line 1: Same here about keeping these _{context} in the anchor definition. There are a few other places as well.

migration/migrating_3_4/configuring-replication-repository.adoc
bergerhoffer, line 1: I can't speak for sure because I'm not familiar with nesting assemblies, but I think that this _{context} should actually stay here. It's only when you're actually using an xref to reference that it should be resolved out.

bergerhoffer (commented):
Just two more questions on whether the nested assemblies need to keep _{context} in the anchors, or resolve them.

Otherwise then this PR is ready to go. @vikram-redhat let us know when the redirects would be ready and when we should merge this. Thanks!

migration/migrating_3_4/configuring-replication-repository.adoc
apinnick: I see what you're saying. I'll revert the ids.

migration/migrating_3_4/deploying-cam.adoc
```diff
@@ -6,7 +6,9 @@ include::modules/common-attributes.adoc[]
 
 toc::[]
 
-Deploying the Cluster Application Migration (CAM) tool requires installing the Cluster Application Migration Operator on the xref:../../migration/migrating_3_4/deploying_cam.adoc#installing-cam-operator-ocp-3_sourcecluster-3-4[{product-title} 3 source] and xref:../../migration/migrating_3_4/deploying_cam.adoc#installing-cam-operator-ocp-4_targetcluster-3-4[{product-title} {product-version} target] clusters.
+You can deploy the Cluster Application Migration (CAM) tool by installing the Cluster Application Migration Operator on the {product-title} 3 source cluster and {product-version} target cluster.
+
+After installing the Cluster Application Migration Operator, you can launch the CAM web console.
 
 :context: sourcecluster-3-4
```
vikram-redhat, line 13: This PR is testing the limits of our (docs.openshift.com) implementation of modular docs :). I don't know how this PR is gonna go on the customer portal, now, and in the future implementation of FCC.

A few things: 

* Why is the context being switched here? The context for this assembly should stay at migrating-3-4 as defined earlier in this file.

* Assemblies shouldn't have an id with an undefined context (see the first line) https://github.com/openshift/openshift-docs/pull/22017/files#diff-db29aa1eeb8fc2419fcf94ed09c98bd5R1. This assembly should match the file name: https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#assembly-file-metadata. 

* I think you mean to define a new attribute and then unset that attribute, rather than changing the context. You have defined the attribute sourcecluster-3-4 and then unset it immediately after that is used in the assembly, which is ok. But, I also think that instead of creating these new attributes, all you need to do is to leave the context that is already defined (migrating-3-4). The included modules will pick that context up and create the unique ids as required. However, that is not what you really need to do here.. see below.

Overall, as we discussed late last week, this is going to be complex and difficult. My suggestion is:

Create unique assemblies for each migration. So instead of deploy-cam.adoc, create deploy-cam-3-4.adoc, deploy-cam-4-1-4-3.adoc etc, and then include the modules that are common to all these. That will still give you the separate navigation, and you can reuse modules and content.

apinnick: > Why is the context being switched here? The context for this assembly should stay at migrating-3-4 as defined earlier in this file.

The context is being switched because the same module (installing CAM on 4.x cluster) is used for installing on a target or on a source cluster. **However**, there are significant differences between the installations, depending on whether the CAM is installed on the target or source cluster. These differences are tagged, so they require separate contexts. BTW, I removed these links, since I didn't think they were necessary and created additional possible points of failure. :-)

> Assemblies shouldn't have an id with an undefined context 

OK. Will check for undefined context and match assembly to file name.

> Create unique assemblies for each migration. 

Will do. That looks like the safest course of action.

PR #22021, opened 2020-05-12
feat(router): Adds documentation for haproxy.router.openshift.io/rewrite-target

benoitf:
related to https://github.com/openshift/origin/issues/20474

and https://github.com/openshift/enhancements/pull/287

### DIFF
--- a/modules/nw-route-specific-annotations.adoc
+++ b/modules/nw-route-specific-annotations.adoc
@@ -30,6 +30,7 @@ and "-". The default is the hashed internal key name for the route. |
 |`haproxy.router.openshift.io/ip_whitelist` | Sets a whitelist for the route. |
 |`haproxy.router.openshift.io/hsts_header` | Sets a Strict-Transport-Security header for the edge terminated or re-encrypt route. |
 |`haproxy.router.openshift.io/log-send-hostname` | Sets the `hostname` field in the Syslog header. Uses the host name of the system. `log-send-hostname` is enabled by default if any Ingress API logging method, such as sidecar or Syslog facility, is enabled for the router. |
+|`haproxy.router.openshift.io/rewrite-target` | Sets the rewrite path of the request on the backend. |
 |===
 
 [NOTE]
@@ -54,3 +55,38 @@ metadata:
 Setting a server-side timeout value for passthrough routes too low can cause
 WebSocket connections to timeout frequently on that route.
 ====
+
+
+.A route specifying a rewrite target
+[source,yaml]
+----
+apiVersion: v1
+kind: Route
+metadata:
+  annotations:
+    haproxy.router.openshift.io/rewrite-target: / <1>
+...
+----
+<1> Sets `/` as rewrite path of the request on the backend.
+
+Setting the `haproxy.router.openshift.io/rewrite-target` annotation on a route specifies that the Ingress Controller should rewrite paths in HTTP requests using this route before forwarding the requests to the backend application.
+The part of the request path that matches the path specified in `spec.path` is replaced with the rewrite target specified in the annotation.
+
+The following table provides examples of the path rewriting behavior for various combinations of `spec.path`, request path, and rewrite target.
+
+.rewrite-target examples:
+[cols="4*", options="header"]
+|===
+|Route.spec.path|Request path|Rewrite target| Forwarded request path
+|/foo|/foo|/|/
+|/foo|/foo/|/|/
+|/foo|/foo/bar|/|/bar
+|/foo|/foo/bar/|/|/bar/
+|/foo|/foo|/bar|/bar
+|/foo|/foo/|/bar|/bar/
+|/foo|/foo/bar|/baz|/baz/bar
+|/foo|/foo/bar/|/baz|/baz/bar/
+|/foo/|/foo|/|N/A (request path does not match route path)
+|/foo/|/foo/|/|/
+|/foo/|/foo/bar|/|/bar
+|===


modules/nw-route-specific-annotations.adoc
```diff
@@ -53,3 +54,34 @@ metadata:
...
+|===
+|Route.spec.path|Request path|Rewrite target| Output
+|/foo|/foo|/|/
+|/foo|/foo/|/|/
+|/foo|/foo/bar|/|/bar
+|/foo|/foo/bar/|/|/bar/
+|/foo|/foo|/bar|/bar
+|/foo|/foo/|/bar|/bar/
+|/foo|/foo/bar|/baz|/baz/bar
+|/foo|/foo/bar/|/baz|/baz/bar
```
elafontaine, line 82: Are we saying we'll be removing trailing `/` ? If yes, isn't it inconsistent with line 78?

benoitf: I made an error when recopying result, fixed

modules/nw-route-specific-annotations.adoc
```diff
@@ -54,3 +55,33 @@ metadata:
 Setting a server-side timeout value for passthrough routes too low can cause
 WebSocket connections to timeout frequently on that route.
 ====
+
+
+.A route rewrite-target annotation
```
Miciah, line 60: To be more consistent with the previous heading:
```suggestion
.A route specifying a rewrite target
```

modules/nw-route-specific-annotations.adoc
```diff
@@ -54,3 +55,33 @@ metadata:
...
+[source,yaml]
+----
+apiVersion: v1
+kind: Route
+metadata:
+  annotations:
+    haproxy.router.openshift.io/rewrite-target: / <1>
+...
+----
+<1> Sets / as rewrite path of the request on the backend
```
Miciah, line 70:
```suggestion
<1> Sets `/` as rewrite path of the request on the backend.
```

modules/nw-route-specific-annotations.adoc
```diff
@@ -54,3 +55,33 @@ metadata:
...
+
```
Miciah, line 71: A paragraph explaining the annotation and the table below would be helpful.  For example,

> Setting the `haproxy.router.openshift.io/rewrite-target` annotation on a route specifies that the Ingress Controller should rewrite paths in HTTP requests using this route before forwarding the requests to the backend application.  The part of the request path that matches the path specified in `spec.path` is replaced with the rewrite target specified in the annotation.  The following table provides examples of the path rewriting behavior for various combinations of `spec.path`, request path, and rewrite target.

modules/nw-route-specific-annotations.adoc
```diff
@@ -54,3 +55,33 @@ metadata:
...
+|Route.spec.path|Request path|Rewrite target| Output
+|/foo|/foo|/|/
+|/foo|/foo/|/|/
+|/foo|/foo/bar|/|/bar
+|/foo|/foo/bar/|/|/bar/
+|/foo|/foo|/bar|/bar
+|/foo|/foo/|/bar|/bar/
+|/foo|/foo/bar|/baz|/baz/bar
+|/foo|/foo/bar/|/baz|/baz/bar/
+|/foo/|/foo|/|Application is not available (Error 404)
```
Miciah, line 84: In this example, the request does not use the route, right?  But the request could use another route, if there is some route with the same host but with `spec.path` omitted or set to `/`, so the response will not necessarily be a 404.  Better to say something along the lines of, "N/A (request path does not match route path)".

modules/nw-route-specific-annotations.adoc
```diff
@@ -54,3 +55,33 @@ metadata:
...
+  annotations:
+    haproxy.router.openshift.io/rewrite-target: / <1>
+...
+----
+<1> Sets / as rewrite path of the request on the backend
+
+.rewrite-target examples:
+[cols="4*", options="header"]
+|===
+|Route.spec.path|Request path|Rewrite target| Output
```
Miciah, line 75: Instead of "Output", how about "Rewritten path", or "Forwarded request path"?

modules/nw-route-specific-annotations.adoc
benoitf: paragraph added 👍

benoitf:
PR updated with feedback

modules/nw-route-specific-annotations.adoc
```diff
@@ -54,3 +55,38 @@ metadata:
...
+kind: Route
+metadata:
+  annotations:
+    haproxy.router.openshift.io/rewrite-target: / <1>
+...
+----
+<1> Sets `/` as rewrite path of the request on the backend
+
+Setting the `haproxy.router.openshift.io/rewrite-target` annotation on a route specifies that the Ingress Controller should rewrite paths in HTTP requests using this route before forwarding the requests to the backend application.
+The part of the request path that matches the path specified in spec.path is replaced with the rewrite target specified in the annotation.
```
Miciah, line 73: One more tiny suggestion:
```suggestion
The part of the request path that matches the path specified in `spec.path` is replaced with the rewrite target specified in the annotation.
```

benoitf: fixed

Miciah:
Many thanks!
/lgtm

@bmcelvee, the path-rewriting feature is new in 4.6.  I don't believe any other documentation changes are needed for the feature.  Is there anything left for engineering to do to move this PR along?

bmcelvee:
@benoitf thank you so much! The PR looks good to me. I only see one thing (missing period on "Sets `/` as rewrite path of the request on the backend), but I don't mind opening a follow-up PR to fix that. 

@Miciah this is great, thank you! I just need to request QE review before I can merge, so I'll go ahead and do that now.

@quarterpin would you mind reviewing this PR please? Thank you!

benoitf:
@bmcelvee I've added the missing `.`

bmcelvee:
Thank you!

bmcelvee:
@quarterpin would you mind reviewing this PR please? Thank you!

PR #22047, opened 2020-05-12
Standardize navigation capitalization to match OpenShift

JStickler:
Navigation capitalization in topic map switched from title case to sentence case (with product names capitalized) to conform with OpenShift standards.


neal-timpe (approved):
LGTM

PR #21894, opened 2020-05-06
BZ1804489 - Add device path note for local storage

bobfuru:
This is a work in progress attempt to add instructions for configuring an additional storage device to be used by local volumes as described in [BZ1804489](https://bugzilla.redhat.com/show_bug.cgi?id=1804489).

I've made an attempt, as suggested, to adapt the following article for storage rather than network devices: https://access.redhat.com/articles/4731401.

Lacking additional info and technical expertise, I'm not sure if this procedure is even close to meeting the request of providing a working example for RHCOS.

@liangxia or @anpingli Can one of you take a look at this and suggest how the example file and machine-config instructions can be more accurately modified to be used for a storage device?

PR #22075, opened 2020-05-13
BZ1835077 - Fix file system path note

bobfuru:
[BZ1835077](https://bugzilla.redhat.com/show_bug.cgi?id=1835077)
Updates the filesystem annotation and description to match filesystem path, rather than block device path.
This is an update to [PR 21894](https://github.com/openshift/openshift-docs/pull/21894).

Preview link here:
https://bz1835077--ocpdocs.netlify.app/openshift-enterprise/latest/storage/persistent_storage/persistent-storage-local.html#local-volume-cr_persistent-storage-local

### DIFF
--- a/modules/persistent-storage-local-create-cr.adoc
+++ b/modules/persistent-storage-local-create-cr.adoc
@@ -50,7 +50,7 @@ spec:
       volumeMode: Filesystem <3>
       fsType: xfs <4>
       devicePaths: <5>
-        - /path/to/device <6>
+        - /path/to/filesystem <6>
 ----
 <1> The namespace where the Local Storage Operator is installed.
 <2> Optional: A node selector containing a list of nodes where the local storage volumes are attached. This
@@ -62,7 +62,7 @@ local volumes.
 <4> The file system that is created when the local volume is mounted for the
 first time.
 <5> The path containing a list of local storage devices to choose from.
-<6> Replace this value with your actual local disks filepath to the LocalVolume resource, such as `/dev/xvdg`. PVs are created for these local disks when the provisioner is deployed successfully.
+<6> Replace this value with your actual local file system path to the LocalVolume resource, such as `/etc/motd`. PVs are created for these local file systems when the provisioner is deployed successfully.
 +
 .Example: Block
 [source,yaml]


bobfuru:
@qinpingli 
@openshift/team-documentation 
PTAL - thanks!

sheriff-rh (approved):
Ah, good catch. LGTM!

modules/persistent-storage-local-create-cr.adoc
```diff
@@ -62,7 +62,7 @@ local volumes.
 <4> The file system that is created when the local volume is mounted for the
 first time.
 <5> The path containing a list of local storage devices to choose from.
-<6> Replace this value with your actual local disks filepath to the LocalVolume resource, such as `/dev/xvdg`. PVs are created for these local disks when the provisioner is deployed successfully.
+<6> Replace this value with your actual local file system path to the LocalVolume resource, such as `/etc/motd`. PVs are created for these local file systems when the provisioner is deployed successfully.
```
qinpingli, line 65: No, `/path/to/device` is correct.
filesystem means when the pv is exported to a container it's a filesystem.

qinpingli:
We do not need this pr.

modules/persistent-storage-local-create-cr.adoc
bobfuru: > No, `/path/to/device` is correct.
> filesystem means when the pv is exported to a container it's a filesystem.

I didn't change `/path/to/device` in the annotation, just the description to use something else besides `dev/xvdg`, which we use in the block device example.

So if this is still not correct, can you please report why in the new [BZ1835077](https://bugzilla.redhat.com/show_bug.cgi?id=1835077) that was opened by the TAM in response to the original PR? 

He is asking that we update the `Filesystem` example to show a filesystem path, rather than a block device name/path. 

Also, can you suggest another alternative for description <6> that will address the TAM's request? Thanks!

qinpingli: Added comment in bug.

bobfuru:
Waiting for TAM to confirm that corresponding BZ should be closed; if so, this PR should be closed/not merged.

bobfuru:
Closing this PR as the QE review found that this was not a bug: https://bugzilla.redhat.com/show_bug.cgi?id=1835077#c5

PR #22079, opened 2020-05-13
Remove deprecated snapshots files for 4.3+

bobfuru:
Snapshots was deprecated in 4.2 (not to be confused with CSI snapshots, which was introduced in 4.4). This PR removes old files that are no longer needed in 4.3+.


vikram-redhat:
@bobfuru I am assuming this didn't require change control?

bobfuru:
@vikram-redhat Correct. These were orphaned files that I left in case I'd want to repurpose the content when deprecated in-tree snapshotter was replaced with CSI snapshotter. Everything was documented in Storage docs and Release Notes at the time. I just wanted to remove these old files that were not used after 4.2.

PR #22090, opened 2020-05-13
Adding docs on Tracing and Jaeger for Serverless

abrennan89:
WIP

- Included Jaeger modules, updated list of "included in" assemblies
- Created Serverless tracing assembly, Jaeger serverless config module
- Clean up of links related to services and topic map


modules/serverless-jaeger-config.adoc
```diff
@@ -0,0 +1,41 @@
...
+----
+tracing:
+  sample-rate: "0.1"
+  backend: zipkin
+  zipkin-endpoint: http://jaeger-collector.<JAEGER_NAMESPACE>.svc.cluster.local:9411/api/v2/spans
+  debug: "true"
+----
++
+* The backend must be `zipkin`.
+// why? Check Jaeger docs.
```
nak3, line 36: It is a Knative's specific configuration. `zipikin` is has same tracing API and we can use it via Jaeger.

It is strange, but it is correct :smile_cat:

modules/serverless-jaeger-config.adoc
```diff
@@ -0,0 +1,41 @@
...
+kind: Jaeger
+metadata:
+  name: jaeger
+----
+
+// TODO: Add images to show configuring this
+. Enable tracing on for Knative Serving:
++
+----
+$ oc edit ks -n knative-serving knative-serving
```
nak3, line 23: Sorry, I had a typo in original doc... Can you remove second `knative-serving`.   
```suggestion
$ oc edit ks -n knative-serving
```

markusthoemmes (commented):
Thanks for the quick turnaround time!

modules/serverless-jaeger-config.adoc
```diff
@@ -0,0 +1,57 @@
...
+  debug: "true"
+----
++
+.. The backend must be `zipkin`.
+.. The `zipkin-endpoint` must point to your `jaeger-collector` service endpoint.
++
+You can find this endpoint by getting the Jaeger URL:
++
+----
+$ oc get route
```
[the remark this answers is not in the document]
nak3: Yes, it should be a cluster local address service. So, the command should be:

```
$ oc get svc
```

modules/serverless-jaeger-config.adoc
```diff
@@ -0,0 +1,57 @@
...
+----
++
+The service endpoint will be in the format `https://jaeger-<namespace>.apss.<snip>`.
```
nak3, line 54: The `https://jaeger-<namespace>.apss.<snip>` is an URL for the Jaeger's web UI. The endpoint which we want to set to `zipkin-endpoint` is different from the URL for Jaeger's web UI.

modules/serverless-jaeger-config.adoc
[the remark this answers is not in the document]
markusthoemmes: Even that is unnecessary right? The command you show only has a placeholder in the namespace, so the user just needs to throw the namespace they installed the Jaeger resource in in there?

modules/serverless-jaeger-config.adoc
markusthoemmes: This one will need the `oc get route` command I believe? To easily fetch the URL. Example output would be great here.

modules/serverless-jaeger-config.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+. Enable tracing on for Knative Serving, by editing the namespace and adding the tracing YAML:
++
+----
+$ oc edit ks -n knative-serving
+----
++
+.Tracing YAML example
+----
+tracing:
```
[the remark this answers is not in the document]
nak3: Yes, here is the entire context:

```yaml
apiVersion: operator.knative.dev/v1alpha1
kind: KnativeServing
metadata:
  name: knative-serving
  namespace: knative-serving
spec:
  config:

    ... snip ..

    tracing:
      sample-rate: "0.1"
      backend: zipkin
      zipkin-endpoint: http://jaeger-collector.<JAEGER_NAMESPACE>.svc.cluster.local:9411/api/v2/spans
      debug: "true"
```

modules/serverless-jaeger-config.adoc
```diff
@@ -0,0 +1,57 @@
...
+kind: Jaeger
+metadata:
+  name: jaeger
+EOF
+----
+
+. Enable tracing on for Knative Serving, by editing the namespace and adding the tracing YAML:
++
+----
+$ oc edit ks -n knative-serving
```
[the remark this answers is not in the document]
nak3: I tested `oc apply` against `KnativeServing` resource. It seems `apply` for KnativeServing overwrites configs and so it is dangerous.

I applied this:
```
cat <<EOF | oc apply -f -
apiVersion: operator.knative.dev/v1alpha1
kind: KnativeServing
metadata:
  name: knative-serving
  namespace: knative-serving
spec:
  config:
    tracing:
      backend: zipkin
      zipkin-endpoint: http://jaeger-collector.default.svc.cluster.local:9411/api/v2/spans
      debug: "true"
EOF
```

Then, all configurations which are not reconciled, they are gone: https://gist.github.com/nak3/7f7092d9fa42dcbbd14e5bf85e0c0e96

modules/serverless-jaeger-config.adoc
nak3: Umm.. Then, we should write the docs like "To access Jaeger's web UI, run `oc get route` and access  to https://jaeger-<namespace>.apss.yordomain.example.com".

modules/serverless-jaeger-config.adoc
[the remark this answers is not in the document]
markusthoemmes: I _think_ that's expected. You would use apply only if you have the entire state of the KnativeServing local of course (like in GitOps).

All I'm saying is: This is normal operation for any Kubernetes resource. KnativeServing isn't special in any way here imo.

modules/serverless-jaeger-config.adoc
markusthoemmes: Right.

modules/serverless-jaeger-config.adoc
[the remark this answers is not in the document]
abrennan89: Awesome, thanks folks!

modules/serverless-jaeger-config.adoc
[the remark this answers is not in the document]
abrennan89: I made some changes to make this more readable if you want to take another look. I think I just need to know what a user would do to get that endpoint so they can use it in the YAML?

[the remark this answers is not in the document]
markusthoemmes: Point is: The endpoint you get here is not what you use in the YAML, but to open the Jaeger dashboard in your browser! @nak3 could you provide an example output here maybe?

serverless/serverless-tracing.adoc
```diff
@@ -0,0 +1,31 @@
...
+It is used to tie information about different units of work together, to understand a whole chain of events in a distributed transaction.
+The units of work may be executed in different processes or hosts.
+
+Developers can visualize call flows in large architectures with distributed tracing. which is useful for understanding serialization, parallelism, and sources of latency.
+
+For more information about Jaeger, see the documentation on xref:../jaeger/jaeger_arch/rhbjaeger-architecture.adoc#rhbjaeger-architecture[Jaeger architecture].
+
+== Using Jaeger with {ServerlessProductName}
+
+To use Jaeger with {ServerlessProductName}, you will need to install the *Elasticsearch Operator*, and then the *Jaeger Operator*.
```
markusthoemmes, line 22: This is techically not true I believe. Should we link to the jaeger docs here? They state you can use jaeger without elasticsearch but you should install elasticsearch if you want to use it in production.

See https://docs.openshift.com/container-platform/4.4/jaeger/jaeger_install/rhbjaeger-installation.html

modules/serverless-jaeger-config.adoc
```diff
@@ -0,0 +1,65 @@
...
+      backend: zipkin
+      zipkin-endpoint: http://jaeger-collector.<JAEGER_NAMESPACE>.svc.cluster.local:9411/api/v2/spans
+      debug: "true"
+----
++
+.. The backend must be `zipkin`.
+.. The `zipkin-endpoint` must point to your `jaeger-collector` service endpoint from the previous step.
+
+.Verification  steps
+Access the Jaeger web console to see tracing data. You can access the Jaeger web console by using the `jaeger-collector` service endpoint.
```
abrennan89, line 58: There's an issue tracking docs for accessing the Jaeger console https://issues.redhat.com/browse/OSSMDOC-91 - so maybe I can leave this here for now and then switch to including that doc module once it exists? Should this be the same thing @nak3 ?

modules/serverless-jaeger-config.adoc
```diff
@@ -0,0 +1,64 @@
...
+  config:
+    tracing:
+      sample-rate: "0.1"
+      backend: zipkin <1>
+      zipkin-endpoint: http://jaeger-collector.<JAEGER_NAMESPACE>.svc.cluster.local:9411/api/v2/spans <2>
+      debug: "true"
+----
++
+<1> The backend must be `zipkin`.
+<2> The `zipkin-endpoint` must point to your `jaeger-collector` service endpoint. To get this endpoint, substitute the namespace where the Jaeger Operator is installed.
```
markusthoemmes, line 49: This should be the namespace where the **resource** above is applied (the Jaeger resource), not the Jaeger operator.

modules/serverless-jaeger-config.adoc
```diff
@@ -0,0 +1,64 @@
...
++
+.Example Jaeger service endpoint
+----
+http://jaeger-collector.openshift-operators.svc.cluster.local:9411/api/v2/spans
```
markusthoemmes, line 53: We shouldn't use openshift-operators here as that'd hint at this being the location where the operator is installed. We could use `default` as the default namespace (as its literally the default namespace :D).

Maybe we should add the namespace to the Jaeger resource above to make that clear too?

modules/serverless-jaeger-config.adoc
```diff
@@ -0,0 +1,64 @@
...
+
+.Verification  steps
+Access the Jaeger web console to see tracing data. You can access the Jaeger web console by using the `jaeger-collector` service endpoint.
+
+. Get the `jaeger-collector` service endpoint:
++
+----
+$ oc get route
+----
+. Open the endpoint address in your browser. The service endpoint should be in the format `pass:c[https://jaeger-<namespace>.apps.<snip>]`.
```
markusthoemmes, line 64: Here's an example command:

```console
$ oc get route jaeger
NAME     HOST/PORT                         PATH   SERVICES       PORT    TERMINATION   WILDCARD
jaeger   jaeger-default.apps.example.com          jaeger-query   <all>   reencrypt     None
```

Let's change the command to be "oc get route jaeger" or "oc -n default get route jaeger" if the default namespcae is where the Jaeger resource is applied (see comments above).

modules/serverless-jaeger-config.adoc
```diff
@@ -0,0 +1,64 @@
...
++
+.Example Jaeger service endpoint
+----
+http://jaeger-collector.openshift-operators.svc.cluster.local:9411/api/v2/spans
+----
+
+.Verification  steps
+Access the Jaeger web console to see tracing data. You can access the Jaeger web console by using the `jaeger-collector` service endpoint.
+
+. Get the `jaeger-collector` service endpoint:
```
markusthoemmes, line 59:
```suggestion
Access the Jaeger web console to see tracing data. You can access the Jaeger web console by using the `jaeger` route.

. Get the `jaeger` route's hostname
```

modules/serverless-jaeger-config.adoc
```diff
@@ -0,0 +1,64 @@
...
+metadata:
+  name: knative-serving
+  namespace: knative-serving
+spec:
+  config:
+    tracing:
+      sample-rate: "0.1"
+      backend: zipkin <1>
+      zipkin-endpoint: http://jaeger-collector.<JAEGER_NAMESPACE>.svc.cluster.local:9411/api/v2/spans <2>
+      debug: "true"
```
maschmid, line 45: I think we should not set the debug to true here.  (Certainly not without an explanation of what it does?)

abrennan89: @markusthoemmes @nak3 what do you think?

nak3: Yes, agree. We should not set to false.

```suggestion
      debug: "false"
```

(If we add an explanation, we can note as "Enabling debug mode allows all spans to be sent to the server bypassing sampling.", which is just same explanation in [upstream](https://github.com/knative/serving/blob/a0c33cf6d3c8645118ded05b7399d428ec970650/config/core/configmaps/tracing.yaml#L52-L53) )

modules/serverless-jaeger-config.adoc
```diff
@@ -0,0 +1,63 @@
...
+    tracing:
+      sample-rate: "0.1"
+      backend: zipkin <1>
+      zipkin-endpoint: http://jaeger-collector.default.svc.cluster.local:9411/api/v2/spans <2>
+      debug: "false" <3>
+----
++
+<1> The backend must be `zipkin`.
+<2> The `zipkin-endpoint` must point to your `jaeger-collector` service endpoint. To get this endpoint, substitute the namespace where the Jaeger custom resource is applied.
+
```
nak3, line 51: There are no `<3>` for ` debug: "false" <3>` above?

modules/serverless-jaeger-config.adoc
```diff
@@ -0,0 +1,63 @@
...
+----
+apiVersion: operator.knative.dev/v1alpha1
+kind: KnativeServing
+metadata:
+  name: knative-serving
+  namespace: knative-serving
+spec:
+  config:
+    tracing:
+      sample-rate: "0.1"
```
maschmid, line 43: Might as well document the `sample-rate` ,  something like:

Defines sampling probability. With `sample-rate: "0.1"`, approximately 1 in 10 traces will be sampled.

modules/serverless-jaeger-config.adoc
```diff
@@ -0,0 +1,65 @@
...
+      sample-rate: "0.1" <1>
+      backend: zipkin <2>
+      zipkin-endpoint: http://jaeger-collector.default.svc.cluster.local:9411/api/v2/spans <3>
+      debug: "false" <4>
+----
++
+<1> The `sample-rate` defines sampling probability. Using `sample-rate: "0.1"`` means that 1 in 10 traces will be sampled.
+<2> The backend must be `zipkin`.
+<3> The `zipkin-endpoint` must point to your `jaeger-collector` service endpoint. To get this endpoint, substitute the namespace where the Jaeger custom resource is applied.
+<4> Enabling debug mode by setting `debug: "false"` allows all spans to be sent to the server, bypassing sampling.
```
maschmid, line 52: Setting `debug: true` would enable the debug mode.  `debug: "false"` doesn't really do anything (expect pointing to users that they can enable it... )

abrennan89: Oops, good catch, thanks @maschmid :slightly_smiling_face:

maschmid:
QE approved

abrennan89:
**Note to docs folks: please do not merge until requested, this one is on hold.**

modules/serverless-jaeger-config.adoc
```diff
@@ -0,0 +1,65 @@
...
+  config:
+    tracing:
+      sample-rate: "0.1" <1>
+      backend: zipkin <2>
+      zipkin-endpoint: http://jaeger-collector.default.svc.cluster.local:9411/api/v2/spans <3>
+      debug: "false" <4>
+----
++
+<1> The `sample-rate` defines sampling probability. Using `sample-rate: "0.1"`` means that 1 in 10 traces will be sampled.
+<2> The backend must be `zipkin`.
```
ahardin-rh, line 50: `backend` must be set to `zipkin`.

modules/serverless-jaeger-config.adoc
```diff
@@ -0,0 +1,65 @@
...
+<3> The `zipkin-endpoint` must point to your `jaeger-collector` service endpoint. To get this endpoint, substitute the namespace where the Jaeger custom resource is applied.
+<4> Debugging should be set to false. Enabling debug mode by setting `debug: "true"` allows all spans to be sent to the server, bypassing sampling.
```
ahardin-rh, line 52: s/false/`false`

serverless/installing_serverless/upgrading-serverless.adoc
```diff
@@ -18,6 +18,6 @@ Before upgrading to the latest Serverless release, you must remove the community
 
 When upgrading from older versions of {ServerlessProductName} to 1.7.0, support for HTTPS requires a change to the format of routes. Knative services created on {ServerlessProductName} 1.6.0 or older versions are no longer reachable at the old format URLs. You must retrieve the new URL for each service after upgrading {ServerlessProductName}.
 
-For more information on retrieving Knative services URLs, see the documentation on xref:../../serverless/knative_serving/interacting-serverless-apps.adoc#interacting-serverless-apps[Interacting with your serverless application].
+For more information on retrieving Knative services URLs, see the documentation on xref:../../serverless/serving-creating-managing-apps.adoc#serving-creating-managing-apps[Verifying your serverless application deployment].
```
ahardin-rh, line 21: I would drop "see the documentation on"

serverless/serverless-tracing.adoc
```diff
@@ -0,0 +1,21 @@
...
+include::modules/common-attributes.adoc[]
+
+toc::[]
+
+Using Jaeger with {ServerlessProductName} allows you to enable _distributed tracing_ for your serverless applications on {product-title}.
+
+Distributed tracing records the path of a request through the various services that make up an application.
+
+It is used to tie information about different units of work together, to understand a whole chain of events in a distributed transaction.
+The units of work may be executed in different processes or hosts.
```
ahardin-rh, line 14: s/may/might

serverless/serverless-tracing.adoc
```diff
@@ -0,0 +1,21 @@
...
+
+Developers can visualize call flows in large architectures with distributed tracing. which is useful for understanding serialization, parallelism, and sources of latency.
+
+For more information about Jaeger, see the documentation on xref:../jaeger/jaeger_arch/rhbjaeger-architecture.adoc#rhbjaeger-architecture[Jaeger architecture] and xref:../jaeger/jaeger_install/rhbjaeger-installation.adoc#rhbjaeger-installation[Installing Jaeger].
```
ahardin-rh, line 18: s/see the documentation on/see

ahardin-rh:
@abrennan89 Just a few minor comments from me

PR #22128, opened 2020-05-14
[cnv] adding empty toc-title var to cnv attributes

ousleyp:
Adding empty `:toc-title:` variable to the CNV-specific attributes file so that the "Table of Contents" header is omitted from CNV assemblies.

@vikram-redhat should I cherrypick this to previous versions as well?
(edited to add: this file exists as far back as 3.11, but it was named cnv_document_attributes.adoc in 3.11.)


vikram-redhat:
@ousleyp anywhere we have cnv docs.

PR #22158, opened 2020-05-14
Enable schema1 support in configured registries

mtrmac:
Many historically-published images in Quay.io use the schema1 format, and both `oc image mirror` and `podman push` fail when pushing schema1 images to registries that reject schema1 (and even if they didn't fail, the push would modify the manifest digest, which is inconsistent with
our disconnected operation design and the general desire to use manifest digests for image references).

Warning: I have tested that this setting works as expected, but I haven’t tested either of the full processes described by the edited documentation modules.

### DIFF
--- a/modules/installation-creating-mirror-registry.adoc
+++ b/modules/installation-creating-mirror-registry.adoc
@@ -101,6 +101,7 @@ OpenSSL documentation.
      -v /opt/registry/certs:/certs:z \
      -e REGISTRY_HTTP_TLS_CERTIFICATE=/certs/domain.crt \
      -e REGISTRY_HTTP_TLS_KEY=/certs/domain.key \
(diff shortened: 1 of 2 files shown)


kalexand-rh:
@jiajliu, @wking, will you PTAL?

jiajliu:
@jianlinliu will you PTAL?

jianlinliu:
Though I did not hit any problems before this new options is introduced, after added this options, image mirror is running well, should be harmless. I am okay with it.

kalexand-rh:
@jianlinliu, thank you! Do you agree that this change should be valid in version 4.2 and later?

jianlinliu:
Actually this is only a docker registry configuration, nothing with OCP (even I never hit any issue whatever the option is there or not). I am okay with it in any version.

PR #22164, opened 2020-05-15
aro 4 fixes

vikram-redhat:
@lamek For ARO 4. Can you check the exclusions make sense, especially excluding the release notes? If release notes should be included, we will need to exclude a lot of content within it.


vikram-redhat:
Also, not sure if the same commit will work for 4.4/4.5/master which should also get these in.

PR #22174, opened 2020-05-15
Secondary NIC support for OSP IPI - OSDOCS-1033

maxwelldb:
Context: https://issues.redhat.com/browse/OSDOCS-1033


maxwelldb:
At the moment, this PR just includes relevant params. Unsure about additional work required at this point.

Fedosin (approved):

maxwelldb:
@wjiangjay Would you be able to review this PR?

maxwelldb:
QE ack in Jira w/CSS comment to address (c115197)

maxwelldb:
@kalexand-rh Would you be able to look at this given that your tables are magnificent? :1st_place_medal:

modules/installation-configuration-parameters.adoc
```diff
@@ -299,9 +305,40 @@ ifdef::osp[]
...
 |An existing floating IP address to associate with the load balancer API.
 |An IP address, for example `128.0.0.1`.
+|====
+
+.Optional {rh-openstack} parameters
+[%header, cols=".^2,.^3,.^5a"]
+|====
+|Parameter|Description|Values
+
+|`compute.platform.openstack
```
kalexand-rh, line 315: I'm not wild about the extra hard return. It makes it look like it's rendering like the other tables, but it isn't. The only difference I see between your table and mine is that I have a variable in angled brackets.

modules/installation-configuration-parameters.adoc
```diff
@@ -299,9 +305,40 @@ ifdef::osp[]
...
+.additionalNetworkIDs`
+|Additional networks that are associated with compute machines. Allowed address pairs are not created for additional networks.
+|A list of one or more UUIDs as strings, for example `fa806b2f-ac49-4bce-b9db-124bc64209bf`.
```
kalexand-rh, line 318: for example,

kalexand-rh (commented):
One pick, one gripe. I hope you can file a bug for that wrapping problem.

modules/installation-configuration-parameters.adoc
maxwelldb: Agreed. Reverting that.

modules/installation-configuration-parameters.adoc
maxwelldb: Just repeating what I see elsewhere in the module. I'll fix this here and fix it in other places in a new PR.

maxwelldb:
@vikram-redhat Would you be able to take a look at this? Kathryn and I do not understand how we can use identical column options for the "Required parameters" and "Optional {rh-openstack} parameters" tables with different results. :(

modules/installation-configuration-parameters.adoc
maxwelldb: Created #23257

modules/installation-configuration-parameters.adoc
maxwelldb: Created #23257

maxwelldb:
@adellape No rush at all, but Vikram mentioned that you might have an idea about this problem.

This row:
image

This row:
image

And this row: 
image

Look quite different. :(

All have the same col options:
`cols=".^2,.^3,.^5a"`

Their content is similar, too, but text wrapping is only borked in the final example. 

Any thoughts about what's going on?

adellape:
@maxwelldb It looks like since the verbatim marked-up names in the "Parameter" column are being considered single words that can't wrap, that first column becomes as wide at the longest entry, squishing the other columns. In that final example's case, it's `controlPlane.platform.openstack.additionalSecurityGroupIDs`. In your first example, however, there is a `<platform>`, and it looks like for some reason it _does_ wrap when it hits a `<`. When I threw some `<>` around the long ones, it wrapped them and narrowed the columns.

maxwelldb:
@adellape :bowing_man: https://giphy.com/gifs/cat-dead-NjevnbNiUmeLm

PR #22214, opened 2020-05-17
Experimental build of OCP 4.4.3 APIs

jboxman:
This is a remix of the OCP 4 APIs. Still to be completed:

- Better grouping, and group names; This uses the API group as a demarcation point
- Spec splitting; Some of these are 10 levels deep; That's just how the APIs are designed
- Reintroduction of API endpoint URLs
- Inclusion of `format` property


jboxman:
@vikram-redhat, so this addresses the biggest issue: missing keys. It now fully resolves each node. This leads to some untidy output, but it reflects the OCP API as it is designed.

The deepest fields are similar to: `.spec.template.spec.affinity.nodeAffinity.preferredDuringSchedulingIgnoredDuringExecution[].preference.matchExpressions`

I'd wanted to represent the APIs as YAML, with callouts, but some of these are 100 rows. So that idea is infeasible. Tables are the only real option. (And some of the description text is four or five paragraphs!)

Unsatisfying output:

Screen Shot 2020-05-17 at 4 35 40 PM

vikram-redhat:
That is still an improvement over what we have now :) 

Would the API docs be better without a navigation bar to the left?

jboxman:
@vikram-redhat, maybe. Our CSS includes a forced horizontal scroll bar, so it is possible to scroll the table. But it may not be immediately obvious that scrolling is possible.

I've also debated stuffing the type information into the top of the description table or under the property name.

The current plan is to split the tables for each object or array, into a sub-table. The code for this exists, but in my tests a month ago it ran slowly enough, iterating thousands of keys many times, to give me pause. A more robust implementation or better foresight might have avoided this.

It may help with clarity; It mostly will just buy space, as by splitting, the table can reference a much condensed property path:

_.spec.blah.foo.bar properties_
.baz - A property
.alpha.beta - Another property.
ect.

But we'll see, I suppose. It's odd, the amount of code, just to display a couple of tables.

jboxman:
@vikram-redhat, so I can confirm that that approach works, but naturally requires some fixes before a useful build is possible. I need to

- Repair handling of `.`
- Sort tables by depth, alphabetically
- Allow customized paths for display in _resources_ configuration file
- Add automatic splitting by path based on density; if a spec has 10 paths overall, maybe not split; if it has 50, maybe split for every path with more than 10 children or whatever

(A subset of paths, split on `.metadata`)

```
[ '.metadata.annotations',
  '.metadata.clusterName',
  '.metadata.creationTimestamp',
  '.metadata.deletionGracePeriodSeconds',
  '.metadata.deletionTimestamp',
  '.metadata.finalizers',
  '.metadata.generateName',
  '.metadata.generation',
  '.metadata.labels',
  '.metadata.managedFields',
  '.metadata.managedFields[]',
  '.metadata.managedFields[].apiVersion',
  '.metadata.managedFields[].fieldsType',
  '.metadata.managedFields[].fieldsV1',
  '.metadata.managedFields[].manager',
  '.metadata.managedFields[].operation',
  '.metadata.managedFields[].time',
  '.metadata.name',
  '.metadata.namespace',
  '.metadata.ownerReferences',
  '.metadata.ownerReferences[]',
  '.metadata.ownerReferences[].apiVersion',
  '.metadata.ownerReferences[].blockOwnerDeletion',
  '.metadata.ownerReferences[].controller',
  '.metadata.ownerReferences[].kind',
  '.metadata.ownerReferences[].name',
  '.metadata.ownerReferences[].uid',
  '.metadata.resourceVersion',
  '.metadata.selfLink',
  '.metadata.uid' ]
```

jboxman:
@vikram-redhat, so I did that thing; it now correctly splits paths. In truth, the output looks insane. The volume of properties is simply unimaginable.

I'm going to split each API roughly by API group, then by individual API:

```
apis
├── cluster
│   ├── api1.adoc
│   └── api2.adoc
├── ect
│   └── api3.adoc
└── storage
    └── apiN.adoc
```

That won't help with the volume of content on each page, though; One strategy is to suppress expansion for certain properties, and link them to definition pages. There is a lot of repetition around Pod, Metadata, Storage, and a few others.

The OpenAPI spec file has nothing specific to facilitate this; the Prometheus CRD directly embeds other schemas it wants, so OpenAPI `$ref` is useless here. Other OCP CRDs do this as well. I can't even guess an approach at present.

The best way to describe it is thus: In the Go code, a struct is mounted and expressed through a property. In the OpenAPI spec, it manifests as mounted. The information about what that structure is is lost. So is `.metadata` actually `meta.v1.ObjectMeta`? Who knows.

Best bet is probably to suppress these by extracting and relocating them further down on each page, so there's a _related specs_ section with `.meta` and whatever else. Could possibly help. I'll have to experiment further. Linking to an outside page is probably out of the question, because I can't reliably know what struct is mounted at `.metadata`.

Or I can use `$ref` where available (all of k8s, some of OCP) and file bugs on any resource that directly embeds structs, because that's awful and no one should ever do that. And that's probably the best approach.

jboxman:
Closing in favor of https://github.com/openshift/openshift-docs/pull/22732.

PR #22225, opened 2020-05-18
Adjust wording in ingress cert docs

ahardin-rh:


danehans:
@ahardin-rh you may want to backport.

/lgtm

PR #22249, opened 2020-05-19
Documentation mentioning the deprecated /etc/sysconfig/docker

mburke5678:
https://bugzilla.redhat.com/show_bug.cgi?id=1767374

### DIFF
--- a/install_config/registry/extended_registry_configuration.adoc
+++ b/install_config/registry/extended_registry_configuration.adoc
@@ -80,42 +80,53 @@ We recommend against rebooting the entire cluster because that incurs
 unnecessary downtime for pods and does not actually clear the cache.
 
 [[whitelisting-docker-registries]]
-== Whitelisting Docker Registries
+== Configuring an External Registry Search List
 
-You can specify a whitelist of docker registries, allowing you to curate a set
-of images and templates that are available for download by {product-title}
-users. This curated set can be placed in one or more docker registries, and then
-added to the whitelist. When using a whitelist, only the specified registries
-are accessible within {product-title}, and all other registries are denied
-access by default.
+You can use the *_/etc/containers/registries.conf_* file to create a list of 
+Docker registries to search for container images. 
 
-To configure a whitelist:
+The *_/etc/containers/registries.conf_* file is a list of registry servers 
+that {product-title} should search against when a user pulls an image using
+the image short name, such as: `myimage:latest`, You can customize the order 
+of the search, specify secure and insecure registries, and define a 
+blocked registry list. {product-title} does not search or allow pulls 
+from registries on the blocked list.
 
-. Edit the *_/etc/sysconfig/docker_* file to block all registries:
+For example, if a user wants to pull the `myimage:latest` image, {product-title} 
+searches the registries in the order they appear in the list until it finds 
+the `myimage:latest`. 
+
+The registry search list allows you to curate a set of images and templates 
+that are available for download by {product-title} users. You can place these 
+images in one or more Docker registries, add the registry to the list, 
+and pull those images into your cluster.
+
+[NOTE]
+====
+When using the registry search list, {product-title} will not pull images 
+from a registry that is not in the search list. 
+====
+
+To configure a registry search list:
+
+. Edit the *_/etc/containers/registries.conf_* file to add or edit the following
+parameters as needed:
 +
 ----
-BLOCK_REGISTRY='--block-registry=all'
-----
-+
-You may need to uncomment the `*BLOCK_REGISTRY*` line.
-. In the same file, add registries to which you want to allow access:
-+
-----
-ADD_REGISTRY='--add-registry=<registry1> --add-registry=<registry2>'
-----
-+
-.Allowing Access to Registries
-----
-ADD_REGISTRY='--add-registry=registry.redhat.io'
-----
-+
-This example would restrict access to images available on the
-link:https://access.redhat.com/search/#/container-images[Red Hat Customer Portal].
+[registries.search] <1>
+registries = ["reg1.example.com", "reg2.example.com"]
 
-Once the whitelist is configured, if a user tries to pull from a container image registry
-that is not on the whitelist, they will receive an error message stating that
-this registry is not allowed.
+[registries.insecure] <2>
+registries = ["reg3.example.com"] 
 
+[registries.block] <3>
+registries = ['docker.io']
+----
+<1> Specify the secure registries from which users can download images 
+using SSL/TLS.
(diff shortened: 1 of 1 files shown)


install_config/registry/extended_registry_configuration.adoc
```diff
@@ -82,40 +82,40 @@ unnecessary downtime for pods and does not actually clear the cache.
 [[whitelisting-docker-registries]]
 == Whitelisting Docker Registries
 
-You can specify a whitelist of docker registries, allowing you to curate a set
-of images and templates that are available for download by {product-title}
-users. This curated set can be placed in one or more docker registries, and then
-added to the whitelist. When using a whitelist, only the specified registries
-are accessible within {product-title}, and all other registries are denied
-access by default.
+You can specify a whitelist of Docker registries, allowing you to curate a set
```
control-d, line 85: The registries configuration isn't a whitelist.  A client can pull from any registry that uses SSL/TLS (i.e., a "secure" registry) as long as it's not in the `registries.block` list (i.e., "blacklist").  The registries configuration only exists to allow customization of the search, insecure, and block list.

install_config/registry/extended_registry_configuration.adoc
```diff
@@ -82,40 +82,40 @@ unnecessary downtime for pods and does not actually clear the cache.
...
+[registries.insecure] <2>
+registries = ["reg3.example.com"] 
 
-Once the whitelist is configured, if a user tries to pull from a container image registry
-that is not on the whitelist, they will receive an error message stating that
-this registry is not allowed.
+[registries.block] <3>
+registries = ['docker.io']
+----
+<1> Specify the secure registries that users can download images.
```
control-d, line 110: The `registries.search` list is the list of registry servers the client (docker or cri-o in this situation) should search against when only an image's short name is specified.  For example if `registries.search` is `['registry.redhat.io','docker.io','registry.example.com']` and the image `myimage:latest` is specified, then the client will start trying to pull from the search registries in the order they're listed (`registry.redhat.io/myimage:latest`, `docker.io/myimage:latest`...) until the image is finally found.  For search it does not matter if the registry is secure or not.  How the client tries to communicate with an secure/insecure registry is handled by `registry.insecure`.

install_config/registry/extended_registry_configuration.adoc
```diff
@@ -82,40 +82,40 @@ unnecessary downtime for pods and does not actually clear the cache.
...
+<2> Specify the insecure registries that users can download images.
```
control-d, line 111: The `registries.insecure` list tells the client which registries are not secured with SSL/TLS, so that it knows not to try to set up a secure connection with them.  This list is independent of `registries.search`, though if there are insecure registries in the search list they should also be added here.

sunilcio:
@mburke5678 LGTM

install_config/registry/extended_registry_configuration.adoc
```diff
@@ -80,42 +80,58 @@ We recommend against rebooting the entire cluster because that incurs
...
-access by default.
+== Configuring an External Registry Search List
+
+You can use the *_/etc/containers/registries.conf_* file to create a list of 
+Docker registries to search for container images. 
+
+The *_/etc/containers/registries.conf_* file is a list of registry servers 
+that {product-title} should search against when a user pulls an image using
+the image short name. For example, if the registry search list contains 
+`registry.redhat.io`, `docker.io`, and `registry.example.com`, and the
```
jboxman, line 91: It might help to list the list;

- `registry.redhat.io`
- `docker.io`
- `registry.example.com`

> For example, if the previous list is searched for `myimage:lastest`, OCP searches from first to last until it finds a match or no match is found.

Or some such; but maybe that's excessive and it's clear enough as written.

install_config/registry/extended_registry_configuration.adoc
```diff
@@ -80,42 +80,58 @@ We recommend against rebooting the entire cluster because that incurs
...
+`registry.redhat.io`, `docker.io`, and `registry.example.com`, and the 
+image short name is `myimage:latest`, {product-title} searches 
+the registries in that order until it finds the `myimage:latest`. 
+
+The *_/etc/containers/registries.conf_* file allows you to customize the 
+order of the search, specify secure and insecure registries, and define a blocked
+registry list. 
+
+{product-title} can pull images from any secure registry, which is a registry 
+that uses SSL/TLS, or insecure registries, which require standard web protocols
```
jboxman, line 100: I'd be surprised if _standard web protocols_ isn't just HTTP. So I'd say an insecure registry just uses HTTP.

install_config/registry/extended_registry_configuration.adoc
```diff
@@ -80,42 +80,58 @@ We recommend against rebooting the entire cluster because that incurs
...
+{product-title} can pull images from any secure registry, which is a registry 
+that uses SSL/TLS, or insecure registries, which require standard web protocols 
+instead of TLS. {product-title} does not search or allow pulls from 
+registries on the blocked list and returns an error message stating 
+that the registry is not allowed.
+
+The registry search list allows you to curate a set of images and templates 
+that are available for download by {product-title} users. You can place these 
+images in one or more Docker registries, then add the registry 
+to the list, which you can pull into your cluster.
```
jboxman, line 108: What can I pull into the cluster? The images or the registry list? It may be obvious, but I've never worked with registries before, so I'm confused.

install_config/registry/extended_registry_configuration.adoc
```diff
@@ -80,42 +80,58 @@ We recommend against rebooting the entire cluster because that incurs
...
+
+When using the registry search list, {product-title} pulls only
```
jboxman, line 110: I think this is duplicated earlier in this file?

install_config/registry/extended_registry_configuration.adoc
```diff
@@ -80,42 +80,58 @@ We recommend against rebooting the entire cluster because that incurs
...
-this registry is not allowed.
+[registries.insecure] <2>
+registries = ["reg3.example.com"] 
 
+[registries.block] <3>
+registries = ['docker.io']
+----
+<1> Specify the secure registries from which users can download images 
+using SSL/TLS.
+<2> Specify the insecure registries from which users can download images
```
jboxman, line 132: I think this is just _HTTP_.

jboxman: Or it could be _without_ TLS; because actually HTTP is always used; It happens either over TLS/SSL or not. The underlying protocol is always HTTP.

jboxman (changes requested):
@mburke5678, left several comments.

Thanks!

PR #22252, opened 2020-05-19
modules/understanding-upgrade-channels: Recommend clearing channel

wking:
I would prefer to configure this by clearing the `upstream` setting, which seems more intuitive for "there is no upstream" to me.  But sadly, it seems that the CVO has been falling back to a default URI when the ClusterVersion upstream is empty since [way][1] [back][2], and that this behavior is [enshrined in the API][3].  Although the channel docs also [talk about defaults][4], the only channel defaulting in the CVO is when creating a ClusterVersion object after the in-cluster copy was (accidentally?) [deleted][4].  So maybe we could talk folks into adjusting the CVO logic to return `NoUpstream` in the empty-upstream case, but at the moment, clearing the channel is the best approach for "the CVO keeps complaining that it can't hit the upstream, and I want to quiet it down" (rhbz#1827378)[5].

[1]: https://github.com/openshift/cluster-version-operator/blame/2c4931dc283c551938be1a00fee290de0b79d99c/pkg/cvo/availableupdates.go#L27-L31
[2]: https://github.com/openshift/cluster-version-operator/commit/ab4d84a021c38fea5e5973287e1f5580928907c1#diff-78f2af341fa49292dd6930f378018867R24
[3]: https://github.com/openshift/api/blame/0422dc17083e9e8df18d029f3f34322e96e9c326/config/v1/types_cluster_version.go#L56-L57
[4]: https://github.com/openshift/api/blame/0422dc17083e9e8df18d029f3f34322e96e9c326/config/v1/types_cluster_version.go#L62-L63
[5]: https://bugzilla.redhat.com/show_bug.cgi?id=1827378

### DIFF
--- a/modules/understanding-upgrade-channels.adoc
+++ b/modules/understanding-upgrade-channels.adoc
@@ -21,6 +21,8 @@ ifndef::openshift-origin[]
 * `stable-{product-version}`
 * `eus-4.6` (only available when running 4.6)
 
+If you do not want the Cluster Version Operator to fetch available updates from the upgrade recommendation service, you can use the `oc adm upgrade channel` command in the OpenShift CLI to configure an empty channel. This configuration can be helpful if, for example, a cluster has restricted network access and there is no local, reachable upgrade recommendation service.
+
 endif::openshift-origin[]
 ifdef::openshift-origin[]
 {product-title} {product-version} offers the following upgrade channel:
--- a/updating/updating-cluster-between-minor.adoc
+++ b/updating/updating-cluster-between-minor.adoc
@@ -9,7 +9,7 @@ You can update, or upgrade, an {product-title} cluster between minor versions.
 
 [NOTE]
 ====
-Because of the difficulty of changing update channels by using `oc`, use the web console to change the update channel. It is recommended to complete the update process within the web console. You can follow the steps in xref:../updating/updating-cluster-cli.adoc#updating-cluster-cli[Updating a cluster within a minor version by using the CLI] to complete the update after you change to a {product-version} channel.
+Use the web console or `oc adm upgrade channel _<channel>_` to change the update channel. You can follow the steps in xref:../updating/updating-cluster-cli.adoc#updating-cluster-cli[Updating a cluster within a minor version by using the CLI] to complete the update after you change to a {product-version} channel.
 ====
 
 == Prerequisites


modules/understanding-upgrade-channels.adoc
```diff
@@ -25,6 +25,8 @@ binary file for a specific version of {product-title} always installs that versi
 * `fast-{product-version}`
 * `stable-{product-version}`
 
+You can also configure an empty channel to tell the Cluster Version Operator to not fetch available updates from the upgrade recommendation service (for example, because your cluster has restricted network access and there is no local, reachable upgrade recommendation service).
```
LalatenduMohanty, line 28: We should add the command to set empty channel here.

LalatenduMohanty (changes requested):

modules/understanding-upgrade-channels.adoc
wking: I'd rather link to (existing?) docs about managing the channel via `oc` and the web console.  Can we address that in follow-up work?

kalexand-rh: As long as the upgrade docs for the relevant scenarios have the command, I think that it's fine to say that you can do it here without the link or the command itself.

I'm not sure this statement makes sense here, though. You haven't introduced the upgrade service yet. Do we expect users to run this command outside of a restricted use case? If not, I'd put it under the *Restricted network clusters* heading.

If it stays here, I might say something like:
> You can also configure an empty upgrade channel, which prevents the cluster's Cluster Version Operator from fetching available upgrades from the upgrade recommendation service. Configure an empty upgrade channel only if the cluster cannot reach the upgrade recommendation services, such as when you have restricted network access.

If you move it, I'd say something like:
> In a cluster with restricted network access, the Cluster Version Operator cannot reach the upgrade recommendation service to fetch upgrades. You can configure an empty upgrade channel to prevent it from attempting to fetch upgrades.

kalexand-rh (commented):
I have a question and some suggestions.

modules/understanding-upgrade-channels.adoc
wking: Rebased on master and picked up your wording for the first sentence with 635018ce5 -> eb93d29ca.  I kept my parenthetical instead of your "only if the cluster cannot reach" sentence, because I don't have a problem with folks opting out of the recommendation service in connected clusters.  I'd be fine expanding on the pros and cons of self-managing updates vs. using a recommendation service if we want to get into that here, but I'm also fine leaving both options on the table without much discussion (which would be distracting noise for folks who want update recommendations) and letting users sort that out on their own.

kalexand-rh: @wking, the parenthetical sentence is half the reason I rewrote it - here's what our style guide says about them:

> Avoid using parentheses in running text because text in parentheses is decreased in importance.

The sentence is also a bit long as-is. Would this work?

> You can also configure an empty upgrade channel, which prevents the cluster's Cluster Version Operator from fetching available upgrades from the upgrade recommendation service. For example, you might configure an empty upgrade channel because your cluster has restricted network access and cannot reach the upgrade recommendation services.

Is opting out of the recommendation service in a connected cluster something that would give the users a good experience in production? If so, I think the future work might also need to include adding a section that provides that explanation later in the file.

wking: > Would this work...

Done with eb93d29ca -> e8552a573.

> Is opting out of the recommendation service in a connected cluster something that would give the users a good experience in production?

Maybe?  I could see folks disabling this if they had their own tooling that always called `oc adm upgrade --allow-explicit-upgrade --to-image ...` (like we do in CI, where we also have channels cleared since openshift/release#8631).  It's going to have no effect outside of whoever is managing cluster updates, and they're presumably the ones setting the channel property.  I'm fine allowing them to make whatever choice makes sense to them without putting an official thumb on the scale.

kalexand-rh:
@jianlinliu, will you PTAL? I think this change should go back to 4.2. Do you agree?

jianlinliu:
LGTM. @jiajliu Can you help review this from your side?

jiajliu:
There is a note in official doc https://docs.openshift.com/container-platform/4.4/updating/updating-cluster-between-minor.html.
```
Because of the difficulty of changing update channels by using oc, use the web console to change the update channel. It is recommended to complete the update process within the web console. You can follow the steps in Updating a cluster within a minor version by using the CLI to complete the update after you change to a 4.4 channel.
```
But from web console, we don't provide null channel in "Channel" option currently. So my suggestion here is that, we'd better to recommend users to clear channel trough CLI and provide the command to clear channel.

kalexand-rh:
@LalatenduMohanty, @wking, do you have the command to clear the upgrade channel?

LalatenduMohanty:
> @LalatenduMohanty, @wking, do you have the command to clear the upgrade channel?

`$ oc patch clusterversion/version --patch '{"spec":{"channel":""}}' --type=merge` would clear the channel. But I think on a fresh installed cluster the channel will not be set. Users need to clear it only if it is set already.

wking:
> But I think on a fresh installed cluster the channel will not be set.

This is only true in CI since openshift/release#8631.  User installs should have default channels, see openshift/installer#3848.

wking:
Is this blocked on docs with an `oc patch ...` command to clear the channel?  I'll try to get an `oc` wrapper around channel maintenance in place in the meantime...

wking:
I've floated openshift/oc#576 for convenient, `patch`-free channel management.

wking:
/lifecycle frozen

Still trying to get review on openshift/oc#576.  Now that 4.7 is in feature freeze, I'll push again once 4.7 forks off master and 4.8 opens up for new features.

vikram-redhat:
@wking is this still valid?

wking:
> @wking is this still valid?

Yeah, we still want to recommend clearing the channel as the way to get the CVO to stop trying to poll the configured `upstream`.  @jiajliu [sort of blocked][1] this PR because we didn't have a convenient way to clear the channel short of `oc patch ...`.  I have not been successful in motivating folks to review openshift/oc#576, but will do another round of begging once master opens for 4.9.  It's also possible that the console folks have landed something that makes clearing your channel easy; I'll check on that front.  And personally, I think we should be distributing this advice, even if we can't make up our minds about how we want customers to accomplish the clearing.  And I'm fine recommending the `patch` command if that's our current best option, even if it isn't particularly elegant.

[1]: https://github.com/openshift/openshift-docs/pull/22252#issuecomment-650694537

wking:
Rebased onto master with e8552a5732 -> bf742cc4e5.

wking:
I've pushed bf742cc4e -> e494b8661, rebasing onto master and updating to mention `oc adm upgrade channel` for clearing channels (just landed for 4.9 via openshift/oc#576).

wking:
And I've pushed e494b8661 -> a0f42786e to update the:

> Because of the difficulty of changing update channels by using `oc`...

language mentioned [earlier][1], and instead mention `oc adm upgrade channel ...`.

[1]: https://github.com/openshift/openshift-docs/pull/22252#issuecomment-650694537

jiajliu:
LGTM.

jiajliu:
@wking @LalatenduMohanty Since we introduce `oc adm upgrade channel` command in this pr(for official doc), and i notice pr #576 has been merged now. I think it will be available soon in v4.9. Do you think QE should have some test against the new added subcommand? If so, we'd better have a jira us to track it, so that QE can involve it. HDYT?

LalatenduMohanty (approved):
/lgtm

wking:
> If so, we'd better have a jira us to track it, so that QE can involve it. 

[OTA-416][1] is our tracker for this change, if that's useful for QE to hang some test-cases on.

[1]: https://issues.redhat.com/browse/OTA-416

updating/updating-cluster-between-minor.adoc
```diff
@@ -9,7 +9,7 @@ You can update, or upgrade, an {product-title} cluster between minor versions.
 
 [NOTE]
 ====
-Because of the difficulty of changing update channels by using `oc`, use the web console to change the update channel. It is recommended to complete the update process within the web console. You can follow the steps in xref:../updating/updating-cluster-cli.adoc#updating-cluster-cli[Updating a cluster within a minor version by using the CLI] to complete the update after you change to a {product-version} channel.
+Use the web console or `oc adm upgrade channel ...` to change the update channel. You can follow the steps in xref:../updating/updating-cluster-cli.adoc#updating-cluster-cli[Updating a cluster within a minor version by using the CLI] to complete the update after you change to a {product-version} channel.
```
jeana-redhat, line 12:
```suggestion
Use the web console or `oc adm upgrade channel` to change the update channel. You can follow the steps in xref:../updating/updating-cluster-cli.adoc#updating-cluster-cli[Updating a cluster within a minor version by using the CLI] to complete the update after you change to a {product-version} channel.
```
Unless they would literally type `...`. 
If `...` is meant to stand in for a parameter, it could be included like this:
> Use the web console or `oc adm upgrade channel _<parameter_name_or_type>_` to change the update channel.

jeana-redhat (commented):
Just a couple minor suggestions.

modules/understanding-upgrade-channels.adoc
```diff
@@ -21,6 +21,8 @@ ifndef::openshift-origin[]
 * `stable-{product-version}`
 * `eus-4.6` (only available when running 4.6)
 
+You can also configure an empty channel to tell the Cluster Version Operator to not fetch available updates from the upgrade recommendation service (for example, because your cluster has restricted network access and there is no local, reachable upgrade recommendation service).  You can configure an empty channel with the `oc adm upgrade channel` command from the OpenShift CLI.
```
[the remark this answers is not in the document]
wking: I've pushed a0f42786e -> 4085e6dff, rebasing on `main` and moving to your wording.

updating/updating-cluster-between-minor.adoc
wking: moved to `_<channel>_` with 4085e6dff -> 4244510df

wking: And just confirming, the `_` give us italics despite being inside backticks:

image

jeana-redhat: yep, italics and brackets denote user-replaced content :) Thanks!

PR #22286, opened 2020-05-19
CO-936, Azure ownedBy permission no longer required

ahardin-rh:
https://issues.redhat.com/browse/CO-936

### DIFF
--- a/modules/installation-azure-service-principal.adoc
+++ b/modules/installation-azure-service-principal.adoc
@@ -124,8 +124,7 @@ Retrying role assignment creation: 4/36
 output. You need these values during {product-title} installation.
 
 . Grant additional permissions to the service principal. The service principal
-requires the legacy `Azure Active Directory Graph -> Application.ReadWrite.OwnedBy`
-permission and the `User Access Administrator` role for the cluster to assign
+requires the `User Access Administrator` role for the cluster to assign
 credentials for its components.
 .. To assign the `User Access Administrator` role, run the following command:
 +


dgoodwin:
/assign @joelddiaz 

Looks ok to me, is this all we need Joel?

joelddiaz:
> /assign @joelddiaz
> 
> Looks ok to me, is this all we need Joel?

I thought what we were documenting was the option to have the Azure install be done with the cloud credentials in `passthrough` mode. To have `mint` mode, the permissions required have not changed.

I believe the docs should be updated to cover both possibilities.

joelddiaz:
Let me try to summarize where I think these docs need to go.

We still support installing a cluster where the ownedBy permissions is granted to the ServicePrincipal used during installation. This allows the cluster to create distinct users for the various in-cluster components that need cloud credentials (eg image registry, machine api, and cluster ingress).

What we are doing here is documenting an additional option where the cluster will not create unique users for the various in-cluster components that need cloud credentials. Those components will simply share the same credentials that were used during installation.

The difference between these two options is that if during an upgrade, a new or existing component needs additional permissions, the installation performed with the ownedBy permission can create/update a user to grow additional permissions.

With the other set of permissions, if additional permissions are needed that are not already granted to the existing cloud credentials used during installation, the cluster cannot resolve the request and will need someone to put new cloud credentials into the cloud credential secret (kube-system/azure-credentials), or the existing credentials will need to be granted the new permissions.

At present, there is no system in place to communicate the fact that new permissions are needed, and the cluster components that don't have sufficient permissions will simply error when trying to make use of the new API calls/permissions. This is because without the ownedBy permission, the cluster simply assumes that the existing cloud credentials are good enough and will blindly pass them to the components that are requesting cloud credentials.

jeana-redhat:
Hey @joelddiaz - Ashley sent me this link yesterday. I don't see a fix version on the Jira linked, so where does this fit into the release timeline?

joelddiaz:
The changes that allowed for this mode of operation (essentially Passthrough mode on Azure) have been in OpenShift for a while (it was needed for ARO). So current releases all support this.
I think there is some confusion in the title stating that "ownedBy permissions is no longer required". We still require these powerful permissions for OpenShift to run in Mint mode, but there are reduced permissions where Passthrough mode can work, and we've never documented this alternative set of permissions.

ahardin-rh:
Closing this PR because my work here is stale, but this is in Jeana's very capable hands now.

jeana-redhat:
In case anyone lands on this PR, the new PR with this content is over here: https://github.com/openshift/openshift-docs/pull/26303

PR #22348, opened 2020-05-21
Bug 1827550: Add the empty node selector namespace annotation when installing Metering.

timflannagan:
The metering-operator recently had a BZ filed where the `defaultNodeSelector`, if configured, in the cluster-wide Pod Scheduler object would inject that node selector to all of the Pods in the metering stack. In the case where a user configured a MeteringConfig custom resource to place each of the metering-operator's operands on a specific set of node selectors, this could potentially collide with the cluster-wide default set of node selectors.

If we add this empty openshift.io/node-selector namespace annotation, it will override the cluster-wide default node selector configuration.

### DIFF
--- a/modules/metering-install-operator.adoc
+++ b/modules/metering-install-operator.adoc
@@ -8,11 +8,41 @@
 
 .Procedure
 
+
+
+. Create a Namespace for the Metering Operator:
+
+.. Create a Namespace object YAML file (for example, `metering-namespace.yaml`) for the Metering Operator:
++
+[source,yaml]
+----
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-metering <1>
+  annotations:
+    openshift.io/node-selector: ""
+  labels:
+    openshift.io/cluster-monitoring: "true"
+----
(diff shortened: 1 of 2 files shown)


timflannagan1:
I just wanted to get the ball rolling on - happy to explore other alternatives as it's a bit awkward to create this namespace locally, using the CLI, and the rest of the installation through the web console.

modules/metering-install-operator.adoc
```diff
@@ -8,11 +8,41 @@
...
+[source,yaml]
+----
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-metering <1>
+  annotations:
+    openshift.io/node-selector: ""
+  labels:
+    openshift.io/cluster-monitoring: "true"
```
timflannagan1, line 26: We may want to call out the purpose of the `openshift.io/node-selector` annotation and the `openshift.io/cluster-monitoring` label in separate bullet points. Or we could have a single bullet point that says to just mirror everything listed in this namespace manifest like cluster-logging does in their docset.

PR #22395, opened 2020-05-26
Removed commands as per msivak's note of May26.

StephenJamesSmith:
Changes from msivak - removed commands in "Configuring huge pages".


ahardin-rh:
@StephenJamesSmith Looks good! Thanks!

PR #22403, opened 2020-05-26
Update CLO to generate Elasticsearch Index Management Retention Policy

mburke5678:
https://issues.redhat.com/browse/LOG-596
https://issues.redhat.com/browse/OSDOCS-904
Design doc: https://github.com/jcantrill/enhancements/blob/3117b47fd72f12394618366989398ad072ee8f3c/enhancements/cluster-logging/cluster-logging-es-rollover-data-design.md


modules/cluster-logging-elasticsearch-retention.adoc
```diff
@@ -0,0 +1,104 @@
...
+//
+// * logging/cluster-logging-elasticsearch.adoc
+
+[id="cluster-logging-elasticsearch-retention_{context}"]
+= Configuring Elasticsearch retention policy
+
+
+Each Elasticsearch index adds load and overhead to the Elasticsearch cluster. You can use the Elasticsearch rollover index API to automatically rollover indices when they reach a specified age. When an index reaches the designated age, Elasticsearch rolls over the index, deleting the current index and creating a new index. Rolling over and deleting indices optimizes the indices and helps Elasticsearch performance. 
+
+You specify how long Elasticsearch stores indices using a separate _retention policy_ for each of the three log sources, infrastucture logs, application logs, and audit logs, using the Cluster Logging Custom Resource (CR), which writes the policy to the Elasticsearch Custom Resource (CR).
```
mburke5678, line 11: @jcantrill Are we supporting only _maxAge_? And no support for _maxDocs_ and _maxSize?

modules/cluster-logging-elasticsearch-retention.adoc
```diff
@@ -0,0 +1,104 @@
...
+      audit:
+        maxAge: 7d
+...
+----
+<1> Specify the maximum age that is allowed before triggering a rollover for each log source.
++
+The Elasticsearch Operator creates or modifies the `indexmanagement-scripts` configMap in the {PROJECT} namespace with the retention policy.
++
+----
+$ oc get cm
```
mburke5678, line 87: @jcantrill Which project is the configMap in?

modules/cluster-logging-elasticsearch-retention.adoc
```diff
@@ -0,0 +1,104 @@
...
+
+////
+Only maxAge in current implementation:
+You can specify any of the following conditions:
+* `max_age`. The maximum age that is allowed before triggering a rollover for the _hot_ indices, or those ​indices that are actively being updated and queried. Specify a value in days (d), hours (h), minutes (m), or seconds (s). 
+* `max_docs`. The maximum number of documents allowed in an index before triggering a rollover. Documents added since the last refresh are not included in the document count. The document count does not include documents in replica shards.
+* `max_size`. The maximum size the index can be before a rollover is triggered. Specify a value in bytes (b), kilobytes (kb), megabytes (mb), or gigabytes (gb). 
+////
+
+For example, the following retention policy rolls over active indices for the infrastructure logs every three days and the old indices are deleted seven days after rollover. {product-title} checks every 15 minutes to determine if the infrastructure indices need to be rolled over.
```
mburke5678, line 21: @jcantrill Do you typically create a policy for all three log sources?

modules/cluster-logging-elasticsearch-retention.adoc
```diff
@@ -0,0 +1,104 @@
...
+    - name:  node.infra 
+      policyRef: infra-policy
+      aliases:
+      - infra  
+...
+----
+<1> For each log source, the retention policy indicated when to delete and rollover logs for that source. The policy includes: 
+* When to rollover the indices, determine by `maxAge`. This setting comes from the Cluster Logging CR.
+* When to check if rollover should be performed, determined by `pollInterval`. This setting is the default.
+* When the rolled-over indicies will be deleted after the rollover, determined by `minAge`. This setting is the default.
```
mburke5678, line 50: @jcantrill I assume the `pollInterval` and `minAge` parameters are defaulted and cannot be set without taking the EO inti unmanged state?

modules/cluster-logging-elasticsearch-retention.adoc
```diff
@@ -0,0 +1,104 @@
...
+
+You specify how long Elasticsearch stores indices using a separate _retention policy_ for each of the three log sources, infrastucture logs, application logs, and audit logs, using the Cluster Logging Custom Resource (CR) to set the number of days after which the indices should be deleted. The Cluster Logging CR writes the full policy to the Elasticsearch Custom Resource (CR).
+
+////
+Only maxAge in current implementation:
+You can specify any of the following conditions:
+* `max_age`. The maximum age that is allowed before triggering a rollover for the _hot_ indices, or those ​indices that are actively being updated and queried. Specify a value in days (d), hours (h), minutes (m), or seconds (s). 
+* `max_docs`. The maximum number of documents allowed in an index before triggering a rollover. Documents added since the last refresh are not included in the document count. The document count does not include documents in replica shards.
+* `max_size`. The maximum size the index can be before a rollover is triggered. Specify a value in bytes (b), kilobytes (kb), megabytes (mb), or gigabytes (gb). 
+////
```
periklis, line 19: Do we need this block? we support only maxAge.

modules/cluster-logging-elasticsearch-retention.adoc
periklis: The config-map `indexmanagement-scripts` is in `openshift-logging`.

modules/cluster-logging-elasticsearch-retention.adoc
mburke5678: @periklis I commented-this section out. Wasn't sure if it was needed. I can remove.

mburke5678:
@QiaolingTang PTAL

modules/cluster-logging-deploy-cli.adoc
```diff
@@ -308,35 +315,42 @@ spec:
...
         size: 200G
       redundancyPolicy: "SingleRedundancy"
   visualization:
-    type: "kibana"  <6>
+    type: "kibana"  <7>
     kibana:
       replicas: 1
   curation:
-    type: "curator"  <7>
+    type: "curator"  <8>
```
QiaolingTang, line 336: The curator will be dropped, so I think we'd better not add the curator in the clusterlogging instance.

modules/cluster-logging-deploy-cli.adoc
```diff
@@ -308,35 +315,42 @@ spec:
...
     curator:
       schedule: "30 3 * * *"
```
QiaolingTang, line 338: Curator will be dropped in the future, so I think we'd better remove this filed in 4.5 DOC.

QiaolingTang:
Hi @mburke5678 , start from 4.5, we use index management to replace the curation, so I suggest to remove the curation statements in the doc, and add some words to tell the users that the curation will be dropped in the future. Although the curation still works in 4.5, but we recommend to use index management.

modules/cluster-logging-elasticsearch-retention.adoc
```diff
@@ -0,0 +1,102 @@
+// Module included in the following assemblies:
+//
+// * logging/cluster-logging-elasticsearch.adoc
+
+[id="cluster-logging-elasticsearch-retention_{context}"]
+= Configuring Elasticsearch retention policy
+
+
+Each Elasticsearch index adds load and overhead to the Elasticsearch cluster. You can use the Elasticsearch rollover index API to automatically rollover indices when they reach a specified age. When an index reaches the designated age, Elasticsearch rolls over the index, moving the current index and creating a new index. After seven days, the old index is automatically deleted. Rolling over and deleting indices optimizes the indices and helps Elasticsearch performance.
```
QiaolingTang, line 9: Per my understanding, the rollover will happens when the index matches any one of: 
1. the index has been a write_index for more than `maxAge` 
2. the index size > 40gb * number_of_primary_shards
3. the index doc count > 40960 * number_of_primary_shards

Note: the `maxAge` is the phases.hot.actions.rollover.maxAge in the elasticsearch CR

modules/cluster-logging-elasticsearch-retention.adoc
```diff
@@ -0,0 +1,102 @@
...
+//
+// * logging/cluster-logging-elasticsearch.adoc
+
+[id="cluster-logging-elasticsearch-retention_{context}"]
+= Configuring Elasticsearch retention policy
+
+
+Each Elasticsearch index adds load and overhead to the Elasticsearch cluster. You can use the Elasticsearch rollover index API to automatically rollover indices when they reach a specified age. When an index reaches the designated age, Elasticsearch rolls over the index, moving the current index and creating a new index. After seven days, the old index is automatically deleted. Rolling over and deleting indices optimizes the indices and helps Elasticsearch performance. 
+
+You specify how long Elasticsearch stores indices using a separate _retention policy_ for each of the three log sources, infrastucture logs, application logs, and audit logs, using the Cluster Logging Custom Resource (CR) to set the number of days after which the indices should be rolled over. The Cluster Logging CR writes the full policy to the Elasticsearch Custom Resource (CR).
```
QiaolingTang, line 11: Except days, we support weeks(w), hours(h/H), minutes(m) and seconds(s)

modules/cluster-logging-elasticsearch-retention.adoc
```diff
@@ -0,0 +1,102 @@
...
+    policies: <1>
+    - name: infra-policy
+      pollInterval: 15m
+      phases:
+        hot:
+          actions:
+            rollover:
+              maxAge: 3d
+        delete:
+          minAge: 7d
```
QiaolingTang, line 31: If the `minAge` is `7d`, then the `maxAge` should be `8h`, I don't think users could set the `maxAge` by themselves:
```
      - name: app-policy
        phases:
          delete:
            minAge: 7d
          hot:
            actions:
              rollover:
                maxAge: 8h
```

modules/cluster-logging-elasticsearch-retention.adoc
```diff
@@ -0,0 +1,102 @@
...
+    - name:  node.infra 
+      policyRef: infra-policy
+      aliases:
+      - infra  
+...
+----
+<1> For each log source, the retention policy indicates when to delete and rollover logs for that source. The policy includes: 
+* When to rollover the indices, determine by `maxAge`. This setting comes from the Cluster Logging CR.
+* When to check if rollover should be performed, determined by `pollInterval`. This setting is the default and cannot be changed.
+* When the rolled-over indicies will be deleted after the rollover, determined by `minAge`. This setting is the default and cannot be changed.
```
QiaolingTang, line 42: The `minAge` in the elasticsearch CR equals to the `maxAge` in the clusterlogging CR.

modules/cluster-logging-elasticsearch-retention.adoc
```diff
@@ -0,0 +1,102 @@
...
+    retentionPolicy: <1>
+      application:
+        maxAge: 1d
+      infra:
+        maxAge: 7d
+      audit:
+        maxAge: 7d
+...
+----
+<1> Specify the maximum age that is allowed before triggering a rollover for each log source.
```
QiaolingTang, line 74: The `maxAge` is for deleting, not for rollover.

modules/cluster-logging-elasticsearch-retention.adoc
jcantrill: Let me provide a broader review the the docs but my initial response is this is too much detail for users.  User's wont use the rollover api, the operator makes use of it.  User's don't need to understand the elasticsearch CR, the CLO will manage it for them.  The biggest idea to keep in mind is the cluster logging CR is an abstraction over logging that is supposed to make management of the cluster easier.  User's can only specify what happens with it; anything downstream of it is out of their control unless they wish to go unmanaged.

modules/cluster-logging-deploy-cli.adoc
jcantrill: It's probably better to add a note this is deprecated and will be removed in future.  Curator settings are still valid but the expectation is they are only used to remove data from the old index format.  "Curation" going forward is by using the new datamodel and rolling indices

modules/cluster-logging-deploy-cli.adoc
```diff
@@ -308,35 +315,42 @@ spec:
...
+However, an unmanaged deployment does not receive updates until the cluster logging is placed back into a managed state.
 <3> Settings for configuring Elasticsearch. Using the Custom Resource (CR), you can configure shard replication policy and persistent storage.
-<4> Specify the number of Elasticsearch nodes. See the note that follows this list.
-<5> Enter the name of an existing StorageClass for Elasticsearch storage. For best performance, specify a StorageClass that allocates block storage.
-<6> Settings for configuring Kibana. Using the CR, you can scale Kibana for redundancy and configure the CPU and memory for your Kibana nodes. For more information, see *Configuring Kibana*.
-<7> Settings for configuring Curator. Using the CR, you can set the Curator schedule. For more information, see *Configuring Curator*.
-<8> Settings for configuring Fluentd. Using the CR, you can configure Fluentd CPU and memory limits. For more information, see *Configuring Fluentd*.
-endif::[]
+<4> Specify the time that Elasticsearch should retain each log source. Logs older than the `maxAge` are deleted.
```
jcantrill, line 348: The code refers to this as "TimeUnit" which is an integer number and time designation (e.g. w, M, h)  I wonder if we should use similar and specify the actual units allowed for clarity

modules/cluster-logging-elasticsearch-retention.adoc
```diff
@@ -0,0 +1,121 @@
+// Module included in the following assemblies:
+//
+// * logging/cluster-logging-elasticsearch.adoc
+
+[id="cluster-logging-elasticsearch-retention_{context}"]
+= Configuring Elasticsearch retention policy
+
+
+Each Elasticsearch index adds load and overhead to the Elasticsearch cluster.
```
jcantrill, line 9: @sichvoge I have reservations about this entire section.  I think its too much detail and it describes a resource a user can not edit without moving to unmanaged.   This seems like we may want a technical document someplace but do we really want it here?

mburke5678: @jcantrill @sichvoge Is the retention policy only for how long to keep the rolled-over indices before deletion (as the name suggests!) using the cluster logging CR. Tthe user cannot change the rollover schedule without going unmanaged? 
In this paragraph, I wanted to describe the rollover process for folks who might not know the process and show the additions to the Elasticsearch CR made by by Cluster Logging CR.

modules/cluster-logging-elasticsearch-retention.adoc
mburke5678: @jcantrill @sichvoge Would an admin type be interested in when the roll overs are performed and how it is related to the maxAge they set?

mburke5678: Particularly, if the admin looks at the Elasticsearch CR and sees the `spec.indexManagement.policies` section, the admin might wonder what it is (?).
And how the `maxAge` the admin set in the cluster logging CR becomes the `delete.minAge` in the Elasticsearch CR. 
I removed much of the rollover details and added a note: _Modifying the Elasticsearch CR is not supported.  All changes to the retention policies must be made in the cluster logging CR._  WDYT?

modules/cluster-logging-deploy-cli.adoc
mburke5678: @QiaolingTang @jcantrill @sichvoge 
I left the Curator setting in the installing topics, where you create the cluster logging CR with new wording. I removed all other sections on configuring curator.
_Settings for configuring the Curator schedule. Curator is used to remove data that is in the Elasticsearch index format prior to {product-title} 4.5 and will be removed in a later release._

modules/cluster-logging-elasticsearch-retention.adoc
mburke5678: See: https://github.com/openshift/openshift-docs/pull/22403#discussion_r437790865

modules/cluster-logging-elasticsearch-retention.adoc
mburke5678: @QiaolingTang @jcantrill @sichvoge What are the defaults for the `maxAge` or does the user need to set a value for all three log types?

modules/cluster-logging-elasticsearch-retention.adoc
sichvoge: Not sure about this as well. @jcantrill Are you saying that an admin will not be able to configure the retention time?

sichvoge: But essentially agree that everything should be done through the ClusterLogging CR and we would need to basically expose fields we want to support out of the ES/Kibana CR there.

mburke5678: @sichvoge If I understand correctly, the admin can configure the retention time and the rollover schedule is devised from the retention time (?). 
I wonder if we should publish where the rollover schedule comes from.

mburke5678: @sichvoge @jcantrill ^^

modules/cluster-logging-elasticsearch-retention.adoc
mburke5678: @sichvoge @jcantrill @QiaolingTang  ^^

QiaolingTang: @mburke5678, there is several conditions:
 1. when creating the clusterlogging instance, if users don't specify the retentionPolicy field, the CLO/EO(?) will create policies for all the 3 log types: http://pastebin.test.redhat.com/875892
2. if users only set retentionPolicy for one or two log types, then there will only have policy for the specified log type(s), and only the specified log type(s) logs could be received, details are in https://bugzilla.redhat.com/show_bug.cgi?id=1845788#c0 and https://bugzilla.redhat.com/show_bug.cgi?id=1845788#c4

mburke5678: @QiaolingTang How does this look:
https://github.com/openshift/openshift-docs/pull/22403/commits/7fbda8283104fa1f931381d658718609cdde17fd

modules/cluster-logging-elasticsearch-retention.adoc
```diff
@@ -0,0 +1,123 @@
...
+= Configuring log retention time
+
+You can specify how long the default Elasticsearch log store keeps indices 
+using a separate _retention policy_ for each of the three log sources: 
+infrastructure logs, application logs, and audit logs. The retention policy, 
+configured using the `maxAge` parameter in the Cluster Logging Custom Resource 
+(CR), determines when to delete the Elasticsearch indices. 
+
+If you do not create a retention policy, logs are retained for seven days by
+default. If you do not specify a retention policy for all three log sources,
```
jcantrill, line 15: From where are you getting this information; it seems like a bug and I do not believe this to be correct if that's the behavior.  Retention policy is about keeping a certain number of TIMEUNIT of logs ... not about  logs being stored or not;  that is what logforwarding is about

modules/cluster-logging-elasticsearch-retention.adoc
```diff
@@ -0,0 +1,123 @@
...
+...
+----
+<1> Specify the time that Elasticsearch should retain each log source. Enter an 
+integer and a time designation: weeks(w), hours(h/H), minutes(m) and seconds(s). 
+For example, `1d` for one day. Logs older than the `maxAge` are deleted. 
+By default, logs are retained for seven days. 
+
+. You can verify the settings in the Elasticsearch Custom Resource (CR).
++
+For example, the following Elasticsearch CR has a retention policy to roll over
```
jcantrill, line 56: For example, the cluster logging operator will generate the following....

Need to make it clear this is a managed resource

modules/cluster-logging-elasticsearch-retention.adoc
```diff
@@ -0,0 +1,123 @@
...
+----
+<1> Specify the time that Elasticsearch should retain each log source. Enter an 
+integer and a time designation: weeks(w), hours(h/H), minutes(m) and seconds(s). 
+For example, `1d` for one day. Logs older than the `maxAge` are deleted. 
+By default, logs are retained for seven days. 
+
+. You can verify the settings in the Elasticsearch Custom Resource (CR).
++
+For example, the following Elasticsearch CR has a retention policy to roll over 
+active indices for the infrastructure logs every eight hours and the old indices
```
jcantrill, line 57: not sure how exactly to express this but the criteria for rollover is when and index meets one of several conditions:
* age
* size
* number of documents

the index will be rolled if any of those conditions are exceeded.  It could be the index is significantly less then 8 hours but  its size in Gb is met

modules/cluster-logging-elasticsearch-retention.adoc
```diff
@@ -0,0 +1,123 @@
...
+<4> When to check if rollover should be performed. This setting is the default 
+and cannot be changed.
++
+[NOTE]
+====
+Modifying the Elasticsearch CR is not supported. All changes to the retention 
+policies must be made in the cluster logging CR.
+==== 
++
+The Elasticsearch Operator creates or modifies the `indexmanagement-scripts`
```
jcantrill, line 95: @sichvoge do you think we need this much detail? IMO we could remove any mention of these scripts

modules/cluster-logging-updating-elasticsearch.adoc
```diff
@@ -0,0 +1,25 @@
...
+.Prerequisites
+
+* Update cluster logging from 4.5 to 4.6.
+
+* Make sure the cluster logging status is healthy:
++
+** All Pods are `ready`.
+** Elasticsearch cluster is healthy.
+
+* Take a snapshot of your Elasticsearch and Kibana indices. For more information, see link:https://www.elastic.co/guide/en/elasticsearch/reference/7.6/snapshot-restore.html[Snapshot and restore] in the Elasticseach documentation.
```
jcantrill, line 19: In line with my other comment, I don't believe we want to ref the snapshot api at all.  In the past we have simply suggested to make a backup of PVCs and in the context of AWS that is also a "snapshot" but not the same as ES snapshot

jcantrill (changes requested):

modules/cluster-logging-deploy-cli.adoc
```diff
@@ -308,35 +315,42 @@ spec:
...
 <3> Settings for configuring Elasticsearch. Using the Custom Resource (CR), you can configure shard replication policy and persistent storage.
-<4> Specify the number of Elasticsearch nodes. See the note that follows this list.
-<5> Enter the name of an existing StorageClass for Elasticsearch storage. For best performance, specify a StorageClass that allocates block storage.
-<6> Settings for configuring Kibana. Using the CR, you can scale Kibana for redundancy and configure the CPU and memory for your Kibana nodes. For more information, see *Configuring Kibana*.
-<7> Settings for configuring Curator. Using the CR, you can set the Curator schedule. For more information, see *Configuring Curator*.
-<8> Settings for configuring Fluentd. Using the CR, you can configure Fluentd CPU and memory limits. For more information, see *Configuring Fluentd*.
-endif::[]
+<4> Specify the time that Elasticsearch should retain each log source. Enter an integer and a time designation: weeks(w), hours(h/H), minutes(m) and seconds(s). For example, `7d` for seven days. Logs older than the `maxAge` are deleted. You must specify a retention policy or the Elasticsearch indices will not be created.
```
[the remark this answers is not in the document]
mburke5678: Per @QiaolingTang 

> if users only set retentionPolicy for one or two log types, then there will only have policy for the specified log type(s), and only the specified log type(s) logs could be received, details are in https://bugzilla.redhat.com/show_bug.cgi?id=1845788#c0 and https://bugzilla.redhat.com/show_bug.cgi?id=1845788#c4
https://github.com/openshift/openshift-docs/pull/22403#discussion_r441253688

Per @vimalk78 in the referenced BZ:

> Elasticsearch Operator will generate index templates for the mappings defined in the CR. (app only in this case)

modules/cluster-logging-elasticsearch-retention.adoc
mburke5678: Per @QiaolingTang 

> the rollover will happen when the index matches any one of:
>    * the index has been a write_index for more than maxAge
>    * the index size > 40gb * number_of_primary_shards
>    * the index doc count > 40960 * number_of_primary_shards
 

https://github.com/openshift/openshift-docs/pull/22403#discussion_r437078356

modules/cluster-logging-about-crd.adoc
```diff
@@ -25,7 +25,7 @@ spec:
   logStore:
     type: "elasticsearch"
     elasticsearch:
-      nodeCount: 3
+      nodeCount: 2
```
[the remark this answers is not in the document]
mburke5678: This CR is specific to Dedicated. I assume it is correct.

modules/cluster-logging-deploy-cli.adoc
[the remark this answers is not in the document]
QiaolingTang: > ...You must specify a retention policy or the Elasticsearch indices will not be created. I do not believe this to be true and if it is then its likely a bug. Failure to define the policy should result in simply infinite retention

@jcantrill could you please confirm this with @vimalk78 ? It seems you have different opinions.

modules/cluster-logging-elasticsearch-retention.adoc
sichvoge: Are we exposing anything else than the age?

modules/cluster-logging-deploy-cli.adoc
```diff
@@ -308,35 +315,42 @@ spec:
...
+<5> Specify the number of Elasticsearch nodes. See the note that follows this list.
+<6> Enter the name of an existing StorageClass for Elasticsearch storage. For best performance, specify a StorageClass that allocates block storage.
+<7> Settings for configuring Kibana. Using the CR, you can scale Kibana for redundancy and configure the CPU and memory for your Kibana nodes. For more information, see *Configuring Kibana*.
```
QiaolingTang, line 351: @mburke5678 `scale Kibana for redundancy`? Do you want to say `replicas`?

`Kibana nodes` --> `Kibana pods` ?

modules/cluster-logging-elasticsearch-retention.adoc
```diff
@@ -0,0 +1,120 @@
...
+
+* the index is older than than the `rollover.maxAge`.
+* the index size is greater than 40gb*number of primary shards.
+* the index doc count is greater than 40960*number of primary shards.
+
+If you do not create a retention policy for any of the log sources, logs 
+are retained for seven days by default. If you do not specify a retention policy 
+for all three log sources, logs from only the sources with a retention policy 
+are stored. For example, if you do not set a retention policy for audit logs, 
+the audit logs will not be retained and there will be no *audit-* index in
```
QiaolingTang, line 25: @mburke5678 Per my understanding, the audit logs won't be collected by default, even though there has `audit-*` index in the ES.
Besides, I'm not sure if this behavior(If you do not specify a retention policy 
for all three log sources, logs from only the sources with a retention policy 
are stored.) is expected or not, let's wait for Jeff's reply.

modules/cluster-logging-elasticsearch-retention.adoc
QiaolingTang: @sichvoge In the ClusterLogging, only the age and the log type.

@mburke5678 the `maxAge` I said is the value of `maxAge` in the Elasticsearch, not the value in the ClusterLogging.

modules/cluster-logging-elasticsearch-retention.adoc
```diff
@@ -0,0 +1,130 @@
...
+...
+----
+<1> Specify the time that Elasticsearch should retain each log source. Enter an 
+integer and a time designation: weeks(w), hours(h/H), minutes(m) and seconds(s). 
+For example, `1d` for one day. Logs older than the `maxAge` are deleted. 
+By default, logs are retained for seven days. 
+
+. You can verify the settings in the Elasticsearch Custom Resource (CR).
++
+For example. the Cluster Logging Operator updated the following
```
lamek, line 73: Change the period to a comma.

lamek (commented):
Few nits and one suggestion for clarification. Otherwise LGTM.

PR #22407, opened 2020-05-26
Create subscription docs + deadlettersink for Serverless

abrennan89:


modules/serverless-creating-subscriptions.adoc
```diff
@@ -0,0 +1,47 @@
...
+
+* Create a Subscription object that uses the default InMemoryChannel type Channel by entering the following:
++
+[source,yml]
+----
+$ oc  create -f - <<EOF
+apiVersion: messaging.knative.dev/v1
+kind: Subscription
+metadata:
+  name: my-subscription
```
maschmid, line 33: We should consider if we want to have "namespace" here (we do so in previous examples) ... if "namespace" is not specified, it would be created in the "current namespace", as set by "oc namespace"

modules/serverless-creating-subscriptions.adoc
```diff
@@ -0,0 +1,47 @@
...
+spec:
+  channel:
+    apiVersion: messaging.knative.dev/v1
+    kind: InMemoryChannel
+    name: default
```
maschmid, line 38: In previous section, we have created a channel named "example-channel" , so we should probably refer to that one.

modules/serverless-creating-subscriptions.adoc
```diff
@@ -0,0 +1,47 @@
...
+// * If you want to use other Channel types with Subscriptions...
+// TODO: Create docs for these, link Kafka etc?
+
+.Procedure
+
+* Create a Subscription object that uses the default InMemoryChannel type Channel by entering the following:
++
+[source,yml]
+----
+$ oc  create -f - <<EOF
```
maschmid, line 29: In other examples we split this to listing the file contents and creating it from the file... while using this is convenient to *nix users, as they can just copy-paste to terminal, it would probably be annoying to windows users...  (not sure what the general openshift doc guidelines say... )

modules/serverless-creating-subscriptions.adoc
```diff
@@ -0,0 +1,47 @@
...
+kind: Subscription
+metadata:
+  name: my-subscription
+spec:
+  channel:
+    apiVersion: messaging.knative.dev/v1
+    kind: InMemoryChannel
+    name: default
+  subscriber:
+    uri: <SERVICE-URI> <1>
```
maschmid, line 40: I believe the "uri" should only be used if the subscriber isn't something a kubernetes resource by itself, as it is safer to use "ref", eg.

```
  subscriber:
    ref:
      apiVersion: serving.knative.dev/v1
      kind: Service
      name: foobar
```

modules/serverless-creating-subscriptions.adoc
```diff
@@ -0,0 +1,54 @@
...
+// * If you want to use other Channel types with Subscriptions...
+// TODO: Create docs for these, link Kafka etc?
+
+.Procedure
+
+. Create a Subscription object that uses the default InMemoryChannel type channel, by creating a YAML file containing the following:
++
+[source,yml]
+----
+apiVersion: messaging.knative.dev/v1
```
maschmid, line 31: at least with 1.8.0, subscription is still not "v1" yet, should be "v1beta1" for now.

modules/serverless-creating-subscriptions.adoc
```diff
@@ -0,0 +1,54 @@
...
+kind: Subscription
+metadata:
+  name: my-subscription
+  namespace: default
+spec:
+  channel:
+    apiVersion: messaging.knative.dev/v1
```
maschmid, line 38: Channel is also not yet "v1", should be "v1beta1" for now (at least as of 1.8.0)

modules/serverless-creating-subscriptions.adoc
```diff
@@ -0,0 +1,54 @@
...
+    kind: InMemoryChannel
```
maschmid, line 39: While InMemoryChannel would work, we have actually created InMemoryChannel indirectly, by creating "Channel" . The point of the "Channel" CRD is to let other objects be independent on actual channel implementation, so we should probably show this by referencing the Channel with the "Channel" kind, instead of "InMemoryChannel" , for the purpose of this example.

modules/serverless-creating-subscriptions.adoc
```diff
@@ -0,0 +1,54 @@
...
+  subscriber:
+    ref:
+      apiVersion: serving.knative.dev/v1
+      kind: Service
+      name: example-service
+----
+. Apply the YAML file by entering:
++
+----
+$ oc apply <FILENAME>
```
maschmid, line 50: oc apply --filename    (or -f for the short version)

modules/serverless-creating-subscriptions.adoc
```diff
@@ -0,0 +1,54 @@
...
+
+* You must have a current installation of xref:../../serverless/installing_serverless/installing-openshift-serverless.adoc#serverless-install-web-console_installing-openshift-serverless[{ServerlessProductName}], including Knative Serving and Eventing, in your {product-title} cluster. This can be installed by a cluster administrator.
+* If you do not have an existing sink that you wish to use, you can create a Service to use as a sink by following the documentation on xref:../../serverless/serving-creating-managing-apps.adoc#serving-creating-managing-apps[Creating and managing serverless applications].
+* You will need a channel to connect your Subscription to. For more information about creating an InMemoryChannel type channel, see xref:../../serverless/knative_eventing/serverless-channels.adoc#serverless-channels[Using channels with Knative Eventing].
+// * If you want to use other Channel types with Subscriptions...
+// TODO: Create docs for these, link Kafka etc?
+
+.Procedure
+
+. Create a Subscription object that uses the default InMemoryChannel type channel, by creating a YAML file containing the following:
```
maschmid, line 27: "Subscription object using channel" sound weird to me.  As stated above, a "subscription" connects "service" to a "channel".  So maybe something like "Create a Subscription object to connect a Channel to a Knative Service, by creating... "

modules/serverless-creating-subscriptions.adoc
```diff
@@ -0,0 +1,54 @@
...
+  name: my-subscription
+  namespace: default
+spec:
+  channel:
+    apiVersion: messaging.knative.dev/v1beta1
+    kind: Channel
+    name: example-channel
+  subscriber:
+    ref:
+      apiVersion: serving.knative.dev/v1beta1
```
maschmid, line 43: a serving Service should actually be "v1"

modules/serverless-creating-subscriptions.adoc
```diff
@@ -0,0 +1,54 @@
+// Module included in the following assemblies:
+//
+// <List assemblies here, each on a new line>
+// * serverless/knative_eventing/serverless-subscriptions.adoc
+
+[id="serverless-creating-subscriptions_{context}"]
+= Creating a Subscription
+
+You can create a Subscription to connect a Service or other event sink to a channel.
+The following procedure shows how to create a Subscription object that uses the default InMemoryChannel type channel.
```
maschmid, line 10: another use of "subscription object uses a channel", not terribly wrong, but it probably be better with "connecting" or something///  (as subscription isn't really "using" a channel).

Also, in the example we're no longer explicitly mentioning the channel type, so we don't need to mention it here either. We could probably make this whole subscription section a bit more concise if we drop any mention of the InMemoryChannel or channel types altogether, and just keep the reference to serverless-channels.adoc, where they are described.

serverless/knative_eventing/serverless-channels.adoc
```diff
@@ -1,13 +1,13 @@
 include::modules/serverless-document-attributes.adoc[]
 [id="serverless-channels"]
-= Using channels
+= Using channels with Knative Eventing
```
abrennan89, line 3: This will all be overwritten by https://github.com/openshift/openshift-docs/pull/23815 anyway, so either rebase that on top of this or vice versa. This section shouldn't take up much review time since it's changing anyway, IMHO.

modules/serverless-creating-subscriptions.adoc
```diff
@@ -0,0 +1,66 @@
...
+    deadLetterSink:
+      ref:
+        apiVersion: serving.knative.dev/v1
+        kind: Service
+        name: error-handler
+  subscriber: <4>
+    ref:
+      apiVersion: serving.knative.dev/v1
+      kind: Service
+      name: example-service
```
cardil, line 48: "Creating and managing serverless applications" page is showing `hello` ksvc. We also use `event-display` ksvc in some places. I didn't see `example-service` anywhere. Maybe better to use one of already existing examples?

cardil (commented):
Only some questions. If resolved without change then I can give LGTM.

modules/serverless-creating-subscriptions.adoc
```diff
@@ -0,0 +1,66 @@
+// Module included in the following assemblies:
+//
+// <List assemblies here, each on a new line>
+// * serverless/knative_eventing/serverless-subscriptions.adoc
+
+[id="serverless-creating-subscriptions_{context}"]
+= Creating a subscription
+
+You can create a subscription to connect a service or other event sink to a channel.
```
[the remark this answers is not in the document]
abrennan89: We don't really have a dedicated section about it, but we do explain that it's an event consumer: https://docs.openshift.com/container-platform/4.5/serverless/architecture/serverless-event-architecture.html

It would be nice to have a better section on this, but it probably won't be immediately, so probably not worth blocking the PR on.

modules/serverless-creating-subscriptions.adoc
```diff
@@ -0,0 +1,66 @@
...
+      kind: Service
+      name: example-service
+----
++
+<1> Name of the subscription.
+<2> Configuration settings for the channel that the subscription connects to.
++
+[NOTE]
+====
+Currently, {ServerlessProductName} only supports the use of InMemoryChannel type channels as part of the Knative Eventing Technology Preview.
```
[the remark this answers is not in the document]
abrennan89: I plan to remove this when we have docs for Kafka here. For now I'll leave it since it's true of the current version.

alanfx:
QE Approved

modules/serverless-creating-subscriptions.adoc
```diff
@@ -0,0 +1,66 @@
...
+      apiVersion: serving.knative.dev/v1
+      kind: Service
+      name: event-display
+----
++
+<1> Name of the subscription.
+<2> Configuration settings for the channel that the subscription connects to.
++
+[NOTE]
+====
```
codyhoag, line 55: This Note is throwing off the callout numbering. After some research/testing, this is a bit tricky due to the callouts being within a procedure and the note being within the callouts :sweat_smile:. I was able to get the intended outcome by doing this:
```
<last line of yaml code>
----
+
--
<1> text
<2> text
+
[NOTE]
====
text
====
<3> text
<4> text
--
```
There may be a better way, but I couldn't find it.

serverless/knative_eventing/serverless-channels.adoc
codyhoag, line 3: Should this title change also be reflected in the `topic_map.yml`?

codyhoag (commented):
A few suggestions. Everything else looks good!

modules/serverless-creating-subscriptions.adoc
abrennan89: I actually think it's OK to remove this because it's mentioned earlier, my bad. Thanks though!

serverless/knative_eventing/serverless-channels.adoc
abrennan89: Decided to leave as was, looking back idk why I changed this :joy:

PR #22419, opened 2020-05-27
CNV OVN naughty xref 

aburdenthehand:
Moving out a line about supported network providers from the 'what-you-can-do-with-virt' module and into the 'about-virt' assembly due to the xref ban in modules. Included comments in both to inform future writers looking to re-use the `wycdwv` module.

@vikram-redhat


ousleyp:
LGTM!

PR #22420, opened 2020-05-27
CNV Adding KubeVirtVersion variable

aburdenthehand:
PM and eng recommended we re-phrase the 'definitive' part of the API reference and link to the specific version instead of master. This is declared as a variable in the doc-attributes file, as suggested in #21981 

Also removed the duplicate link. Moved the API reference link to the create-vm assembly under the heading 'Additional resources' and added a note to call out that options in upstream are not supported.

The "KubeVirtVersion: will need to be added to the pre-flight checklist to ensure it is updated prior to release

### DIFF
--- a/virt/virtual_machines/virt-create-vms.adoc
+++ b/virt/virtual_machines/virt-create-vms.adoc
@@ -37,6 +37,11 @@ include::modules/virt-creating-vm.adoc[leveloffset=+1]
 
 include::modules/virt-vm-storage-volume-types.adoc[leveloffset=+1]
 
-See the
-https://kubevirt.io/api-reference/master/definitions.html#_v1_virtualmachinespec[kubevirt
-API Reference] for a definitive list of virtual machine settings.
+== Additional resources
+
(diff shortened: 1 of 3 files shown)


dankenigsberg (approved):

virt/virtual_machines/virt-create-vms.adoc
```diff
@@ -37,6 +37,11 @@ include::modules/virt-creating-vm.adoc[leveloffset=+1]
...
-See the
-https://kubevirt.io/api-reference/master/definitions.html#_v1_virtualmachinespec[kubevirt
-API Reference] for a definitive list of virtual machine settings.
+== Additional resources
+
+The `VirtualMachineSpec` definition in the link:https://kubevirt.io/api-reference/{KubeVirtVersion}/definitions.html#_v1_virtualmachinespec[KubeVirt {KubeVirtVersion} API Reference] provides broader context for the parameters and hierarchy of the virtual machine specification.
+
+[NOTE]
+====
+The KubeVirt API Reference is the upstream project reference and may contain parameters that are not supported in {VirtProductName}.
```
ousleyp, line 46: s/may/might

ousleyp (commented):
One nit, otherwise LGTM

PR #22459, opened 2020-05-28
Using ApiServerSource event sources in dev console

abrennan89:


serverless/event_sources/knative-event-sources.adoc
```diff
@@ -0,0 +1,31 @@
...
+PingSource:: Periodically sends ping events with a constant payload. It can be used as a timer.
+
+SinkBinding is also supported, which allows you to connect core Kubernetes resources such as `Deployment`, `Job`, or `StatefulSet` with a sink.
+
+You can create and manage Knative event sources using the Developer perspective, the `kn` CLI, or by applying YAML files.
+
+.Prerequisites
+
+* You must have a current installation of xref:../../serverless/installing_serverless/installing-openshift-serverless.adoc#serverless-install-web-console_installing-openshift-serverless[{ServerlessProductName}], including Knative Serving and Eventing, in your {product-title} cluster. This can be installed by a cluster administrator.
+* You must create the `default` broker in the same namespace that the event source will be installed in. For more information, see xref:../../serverless/knative_eventing/serverless-using-brokers.adoc#serverless-using-brokers[Using brokers with Knative Eventing].
```
matzew, line 20: that's not correct.

The Sources do NOT require a broker to be installed. Any compliant source can sink to:
* Knative Serving Services
* Knative Eventing Channel
* Knative Eventing Broker

More than that, they can emit their events to:
* standard k8s Services
* URI (including off cluster urls)

serverless/event_sources/knative-event-sources.adoc
```diff
@@ -0,0 +1,31 @@
...
+* You must create or update a service account, role and role binding for event sources.
```
matzew, line 21: the service account is needed here, b/c the ApiServerSources is accessing "k8s events".

The ping source, for instance does not need that, to give a counter example

serverless/event_sources/knative-event-sources.adoc
```diff
@@ -0,0 +1,31 @@
...
+* You must create a serverless application, or Knative service, before you can create an event source. Event sources need a service to use as an event _sink_. The sink is the service or application that events are sent to from the event source.
```
matzew, line 22: you can create the source, but it does not fully reconcile.

Once the sink (e.g. Knative service) is there, the pod for the source will be created

modules/odc-creating-apiserversource.adoc
```diff
@@ -0,0 +1,41 @@
...
++
+image::event-sources-odc.png[Add an Event Source in the Developer perspective]
+. Select **Event Sources** -> **ApiServerSource**.
++
+image::apiserversource-odc.png[Create an ApiServerSource]
+. Configure the **ApiServerSource** settings:
++
+image::apiserver-odc-config-1.png[Configuring the ApiServerSource]
++
+.. Enter `sources.knative.dev/v1alpha1` as the **APIVERSION**, and `ApiServerSource` as the **KIND**.
```
invincibleJai, line 20: here `ApiVersion` and `kind` is not of ApiServerSource for any sources this will be handled internally. Here these are for resources needs to be watched it could be more then one pair there is option in UI to add more.

```
  resources:
  - apiVersion: v1
    kind: Event
```

https://github.com/knative/docs/blob/8dcca2764db60006bf256fb98f5ba32783b80ec5/docs/eventing/samples/kafka/channel/020-k8s-events.yaml#L9-L11

even Screenshot may need update.

@matzew can add more input

modules/odc-creating-apiserversource.adoc
```diff
@@ -0,0 +1,41 @@
...
+.Verification steps
+
+. After you have created the ApiServerSource, you will see it connected to the service it is sinked to in the **Topology** view.
++
+image::toplogy-odc-apiserver.png[ApiServerSource Topology view]
+
+// . Create events by launching a pod in the default namespace.
+// TODO: How to do this? What should a user see?
+
+= Deleting the ApiServerSource
```
invincibleJai, line 39: EventSources can be deleted on right click via context menu or on click it'll open side there are option menu with which we can delete.

Screenshot 2020-06-16 at 12 14 27 PM

modules/odc-creating-apiserversource.adoc
```diff
@@ -0,0 +1,41 @@
...
+. Click **Create**.
+
+.Verification steps
+
+. After you have created the ApiServerSource, you will see it connected to the service it is sinked to in the **Topology** view.
++
+image::toplogy-odc-apiserver.png[ApiServerSource Topology view]
+
+// . Create events by launching a pod in the default namespace.
+// TODO: How to do this? What should a user see?
```
invincibleJai, line 37: to verify can follow steps as in here for Event Creation if watching events or can deploy image through `container image` in Add flow from ODC with image `busybox`

```
kubectl run busybox --image=busybox --restart=Never -- ls
kubectl delete pod busybox
``` 
https://github.com/knative/docs/tree/master/docs/eventing/samples/kubernetes-event-source#create-events

and check pods logs in devConsole can see events

Screenshot 2020-06-16 at 12 33 23 PM

modules/odc-creating-apiserversource.adoc
```diff
@@ -0,0 +1,41 @@
...
+.. Select the **Service Account Name** for the service account that you created for event sources.
+.. Select the targeted Knative service from the dropdown menu in **Sink** -> **Knative Service**.
+// .. Add details about the image:
+// General -> Application Name, Name
+// image::apiserver-odc-config-2.png[Configuring the ApiServerSource]
+. Click **Create**.
+
+.Verification steps
+
+. After you have created the ApiServerSource, you will see it connected to the service it is sinked to in the **Topology** view.
```
invincibleJai, line 32: Can even show sidebar on click to show associated deployments and pods for `ApiServerSource` are up and running.

Screenshot 2020-06-16 at 12 39 30 PM

serverless/event_sources/knative-event-sources.adoc
abrennan89: Thanks Matthias - I'm trying to get this structured right across our event source docs, so I appreciate the feedback.

serverless/event_sources/knative-event-sources.adoc
abrennan89: So only the ApiServerSource needs the service a/c + the broker?

serverless/event_sources/knative-event-sources.adoc
abrennan89: It doesn't let me create it at all without a service if you're using dev console? Is that expected behavior?

modules/odc-creating-apiserversource.adoc
abrennan89: @matzew can you clarify please, thanks.

modules/odc-creating-apiserversource.adoc
abrennan89: Thank you!

modules/odc-creating-apiserversource.adoc
matzew: I don't understand what you mean. Clarify what ? 

Let me guess here.  Since the for is already representing the `ApiServerSource`, the fields capture what's supported for its `spec`.  So the two fields on the form (`apiVersion` and `kind`) are to capture the K8s type to be observed, eg:

```
  resources:
  - apiVersion: v1
    kind: Event
```

As in the above example the `v1` versioned `Event` object, representing k8s platform events

matzew: @invincibleJai the fields are not to describe the source, right? Or am I getting that wrong?

It;s for the APi of the API_SERVERS_SOURCE, right ?

serverless/event_sources/knative-event-sources.adoc
matzew: The APIServerSource needs the service account b/c it accesses k8s event, like

```
  resources:
  - apiVersion: v1
    kind: Event
```

see here: https://github.com/matzew/knative-eventing-samples/blob/master/01-source_service/020-k8s-events.yaml#L9-L11

The source does NOT need the service-account, b/c of the broker.  Generally the `APIServrSrc` needs a ServiceAccount, it is completely independent if it does sink to a `Service`, a `Channel` or a `Broker`.
* K

serverless/event_sources/knative-event-sources.adoc
matzew: I only know the yaml, and I can create it there, but the pod does not come up.

I guess the UX prevents this, as this might be undesired experience. @invincibleJai can u confirm?

serverless/event_sources/knative-event-sources.adoc
abrennan89: I'm not saying it needs the service a/c _because_ of the broker, I'm just saying ApiServerSource needs both a service a/c and a broker, right?

modules/odc-creating-apiserversource.adoc
invincibleJai: > @invincibleJai the fields are not to describe the source, right? Or am I getting that wrong?
> 
> It;s for the APi of the API_SERVERS_SOURCE, right ?

yes fields here are not to describe `ApiVersion` and `kind` for apiServerSource but to resources i.e events as described above.

serverless/event_sources/knative-event-sources.adoc
abrennan89: Assuming that this is the way it works/is supposed to work since we haven't heard anything. @invincibleJai let me know if this needs changes.

modules/odc-creating-apiserversource.adoc
abrennan89: On second thoughts, I think this might be sort of out of scope for this PR. This is because verification steps are optional, and adding details about creating an ODC flow using busybox isn't really something we require. Thanks for the info anyway though, we could maybe add this in a future PR when there's more time to doc optional steps.

cardil (approved):
LGTM

modules/serverless-service-ac-event-sources.adoc
```diff
@@ -0,0 +1,62 @@
...
+----
++
+<1> Change this namespace to the namespace that you have selected for installing the event source.
++
+[NOTE]
+====
+If you want to re-use an existing service account with the appropriate permissions, you must modify the `authentication.yaml` for that service account.
+====
+
+. Create the service account, role binding and cluster binding by entering the following command:
```
ahardin-rh, line 58: role binding, and cluster

modules/odc-creating-apiserversource.adoc
```diff
@@ -0,0 +1,39 @@
+// Module included in the following assemblies:
+//
+// * applications/application_life_cycle_management/odc-knative-event-sources.adoc
```
abrennan89, line 3: Actually it's not here, it's in the Serverless section so I need to update this, but does it need to be in this or some other assembly as well for ODC @Preeticp ?

modules/odc-creating-apiserversource.adoc
```diff
@@ -0,0 +1,39 @@
...
+
+[id="odc-creating-apiserversource_{context}"]
+= Creating an ApiServerSource event source using the Developer perspective
+
+.Procedure
+
+. Navigate to the **Add** page.
++
+image::event-sources-odc.png[Add an Event Source in the Developer perspective]
```
Preeticp, line 12: Suggestion: maybe drop this image. It keeps getting updated and becomes a maintenance overhead. For instance in 4.6 they will be adding a Channel card and you will need to update it.

modules/odc-creating-apiserversource.adoc
```diff
@@ -0,0 +1,39 @@
...
+. Select **Event Sources** -> **ApiServerSource**.
```
Preeticp, line 13: s/**Event Sources** /**Event Source**:  In the Add page it is **Event Source** which then ironically takes you to the **Event Sources** page. I will let the Devs know that it should be made consistent.

modules/serverless-service-ac-event-sources.adoc
```diff
@@ -0,0 +1,62 @@
+// Module included in the following assemblies:
+//
+// serverless/knative_eventing/serverless-kn-source.adoc
+
+[id="serverless-service-ac-event-sources_context"]
+= Creating a service account, role and binding for event sources
+
+.Procedure
+
+. Create a service account, role, and role binding for the event source, by creating a file named `authentication.yaml` and copying the following sample code into it:
```
ahardin-rh, line 10: s/event source, by/event source by

ahardin-rh:
@abrennan89 Just some minor comments from me. Looks great!

modules/odc-creating-apiserversource.adoc
```diff
@@ -0,0 +1,39 @@
...
++
+image::apiserversource-odc.png[Create an ApiServerSource]
```
Preeticp, line 15: Probably crop the image to remove the left nav panel?
If you have the screenshot maybe show the other fields you cover in the doc like the Service account name etc. if not this is good enough.
Do we want to ask the ODC devs for input on the commented out sections? If not now, maybe for later?

modules/odc-creating-apiserversource.adoc
```diff
@@ -0,0 +1,39 @@
...
+// General -> Application Name, Name
+// image::apiserver-odc-config-2.png[Configuring the ApiServerSource]
+. Click **Create**.
+
+.Verification steps
+
+. After you have created the ApiServerSource, you will see it connected to the service it is sinked to in the **Topology** view.
++
+image::toplogy-odc-apiserver.png[ApiServerSource Topology view]
+
```
Preeticp, line 31: Do we plan to update the above image as per the dev input, later?

Preeticp:
Hi @abrennan89 the DevConsole sections looks great. I have left some minor suggestions. Thank you for doing this!

modules/odc-creating-apiserversource.adoc
Preeticp: Alternatively, first step could be: Navigate to the **Add** page and select **Event Source**.
and second could be: In the **Event Sources** page, select **ApiServerSource** in the **Type** section.

modules/odc-creating-apiserversource.adoc
abrennan89: I expanded on the comments, but I don't think these fields are required for now, so I wouldn't block the PR to explain them. It wasn't flagged by the eng SMEs who reviewed it or by QE so I'd say it's OK as is for now.
Maybe add a follow up docs issue in the backlog?

abrennan89:
@Preeticp @ahardin-rh if this looks OK to you both now it should be ready to merge :slightly_smiling_face: I've resolved all the comments except for the additional fields that AFAIK aren't required (see comments).

abrennan89:
@Preeticp this is for OCP 4.5 + 4.6, right?

modules/odc-creating-apiserversource.adoc
Preeticp: I agree, no reason to block this PR. Also, I would go by your judgement Ashleigh since you have tested these scenarios. I will discuss this with the Dev and only if required create an issue.

Issue #19847, opened 2020-02-20
OCP 4.5 Release Notes Tracker


soltysh:
Please leave comments here for anything that should be highlighted in the 4.5 release notes. Thank you!

Rendered draft of 4.5 release notes: https://docs.openshift.com/container-platform/4.5/release_notes/ocp-4-5-release-notes.html

soltysh:
https://github.com/openshift/oc/pull/310 drops the following commands and flags:
- `oc policy can-i` command deprecated in 3.9 (https://github.com/openshift/origin/commit/73338c4), users should use `oc auth can-i` instead
- `oc new-app --image` and `oc new-build --image` flags deprecated in 3.2 (https://github.com/openshift/origin/commit/6517d76), users should use `--image-stream` instead
- `oc process -t` flag deprecated in 3.11 (https://github.com/openshift/origin/commit/6a8cc0b), users should use `--template` instead
-  `oc process --output-version` flag deprecated in 3.11 (https://github.com/openshift/origin/commit/6a8cc0b), the flag was already ignored
- `oc set deployment-hook -v` flag deprecated in 3.11 (https://github.com/openshift/origin/commit/2fcf185), users should use `--volumes` instead
- `oc set volumes --list` flag deprecated in 3.3 (https://github.com/openshift/origin/commit/af79d7d), `oc set volumes` w/o flags will list volumes
- `oc status -v` and `oc status --verbose` flags deprecated in 3.11 (https://github.com/openshift/origin/commit/2fcf185), users should use `--suggest` instead

codyhoag:
The deprecated `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt` file will be removed in 4.5. From https://github.com/openshift/openshift-docs/issues/18426#issuecomment-574239835.

soltysh:
https://github.com/openshift/oc/pull/355 changes default behavior of `oc new-app`, starting since 4.5 the default resource produced by `oc new-app` will be Deployment and not DeploymentConfig (for more information about differences see https://docs.openshift.com/container-platform/4.3/applications/deployments/what-deployments-are.html). Users interested in generating DeploymentConfig can pass `--as-deployment-config` flag when invoking `oc new-app`.

jatanmalde:
With OpenShift 4.4, I see some change with installation format, should we include one such paragraph of it in the release docs. I see etcd-member and etcd-metrics running on the bootstrap and not on the masters as per previous installation methods?

[...]

PR #22467, opened 2020-05-28
Updated RNs on Infrastructure Features filtering

ahardin-rh:
This addresses https://github.com/openshift/openshift-docs/issues/19847#issuecomment-634401043

We're not going to call out FIPS MODE and PROXY at this time. That may change in the future. 

@codyhoag FYI


codyhoag (commented):
LGTM!

PR #22477, opened 2020-05-28
Security assembly and mtls

neal-timpe:
https://issues.redhat.com/browse/OSSMDOC-37
https://issues.redhat.com/browse/OSSMDOC-29

Please cherrypick to 4.2-4.5.

### DIFF
--- a/modules/ossm-security-mtls.adoc
+++ b/modules/ossm-security-mtls.adoc
@@ -1,0 +1,63 @@
+// Module included in the following assemblies:
+//
+// * service_mesh/service_mesh_user_guide/ossm-security.adoc
+
+[id="ossm-security-mtls_{context}"]
+= Enabling mutual Transport Layer Security (mTLS)
+
+Mutual Transport Layer Security (mTLS) is a protocol where two parties authenticate each other at the same time. It is the default mode of authentication in some protocols (IKE, SSH) and optional in others (TLS).
+
+MTLS can be used without changes to the application or service code. The TLS is handled entirely by the service mesh infrastructure and between the two sidecar proxies.
+
+By default, {ProductName} is set to permissive mode, where the sidecars in {ProductShortName} accept both plain-text traffic and connections that are encrypted using mTLS. If a service in your mesh is communicating with a service outside the mesh, strict mTLS could break communication between those services. Use permissive mode while you migrate your workloads to {ProductShortName}.
+
+== Enabling strict mTLS across the mesh
+
+If your workloads do not communicate with services outside your mesh and communication will not be interrupted by only accepting encrypted connections, you can enable mTLS across your mesh quickly. Set `spec.istio.global.mtls.enabled` to `true` in your ServiceMeshControlPlane resource. The operator creates the required resources.
+
+[source,yaml]
+----
+apiVersion: maistra.io/v1
+kind: ServiceMeshControlPlane
+spec:
+  istio:
+    global:
+      mtls:
+        enabled: true
+----
+
+[id="ossm-security-mtls-sidecars-incoming-services_{context}"]
+=== Configuring sidecars for incoming connections for specific services
+
+You can also configure mTLS for individual services or namespaces by creating a policy.
+
+[source,yaml]
+----
+apiVersion: "authentication.maistra.io/v1"
+kind: "Policy"
+metadata:
+  name: "default"
+  namespace: <NAMESPACE>
+spec:
+  peers:
+  - mtls: {}
+----
+
+[id="ossm-security-mtls-sidecars-outgoing_{context}"]
+== Configuring sidecars for outgoing connections
+
+Create a destination rule to configure {ProductShortName} to use mTLS when sending requests to other services in the mesh.
+
+[source,yaml]
+----
+apiVersion: "networking.istio.io/v1alpha3"
+kind: "DestinationRule"
+metadata:
+  name: "default"
+  namespace: <CONTROL_PLANE_NAMESPACE>
+spec:
+  host: "*.local"
+  trafficPolicy:
+    tls:
+      mode: ISTIO_MUTUAL
+----
--- a/_topic_map.yml
+++ b/_topic_map.yml
@@ -1445,10 +1445,8 @@ Topics:
     File: ossm-traffic-manage
   - Name: Data visualization and observability
     File: ossm-observability
-#  - Name: Grafana tutorial
-#    File: ossm-tutorial-grafana
-#  - Name: Prometheus tutorial
-#    File: ossm-tutorial-prometheus
+  - Name: Security
+    File: ossm-security
 - Name: Support
   Dir: service_mesh_support
   Topics:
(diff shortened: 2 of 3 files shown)


modules/ossm-security-mtls.adoc
```diff
@@ -0,0 +1,52 @@
...
+By default, {ProductName} is set to permissive mode, where the sidecars in {ProductShortName} accept both plain-text traffic and connections that are encrypted using mTLS. If a service in your mesh is communicating with a service outside the mesh, strict mTLS could break communication between those services. Use permissive mode while you migrate your workloads to {ProductShortName}.
+
+== Enabling strict mTLS across the mesh
+
+If your workloads do not communicate with services outside your mesh and communication will not be interrupted by only accepting encrypted connections, you can enable mTLS across your mesh quickly. Set `.Spec.istio.global.mtls.enabled` to `true` in your ServiceMeshControlPlane resource. The operator create the required resources.
+
+[id="ossm-security-mtls-sidecars-incoming_{context}"]
+== Configuring sidecars for incoming connections
+
+Create a ServiceMeshPolicy to enforce mTLS encryption on incoming connections. It instructs all sidecars to reject plain text connections and accept mTLS connections. The resource must be named `"default"` and it must be created in a project in your control plane.
```
dgn, line 21: - [...] and *only* accept mTls connections.`
- [...] and it must be created in the same project as your ServiceMeshControlPlane resource.

modules/ossm-security-mtls.adoc
```diff
@@ -0,0 +1,52 @@
...
+kind: "DestinationRule"
+metadata:
+  name: "default"
+  namespace: <CONTROL_PLANE_NS>
+spec:
+  host: "*.local"
+  trafficPolicy:
+    tls:
+      mode: ISTIO_MUTUAL
+----
```
rcernich, line 52: Setting mtls=true in the smcp spec has the same effect.  It also configures disables mtls for the k8s api server.

mtls can be configured for individual services/namespaces by using `Policy` instead of `ServiceMeshPolicy`, the latter applying to all namespaces/services in the mesh.

neal-timpe: @rcernich I made some changes here. I added a more explicit yaml example to enabling strict mtls across the mesh. 

And I added another section about using a policy. Did I understand you correctly?

modules/ossm-security-mtls.adoc
```diff
@@ -0,0 +1,80 @@
...
+
+Mutual Transport Layer Security (mTLS) is a protocol where two parties authenticate each other at the same time. It is the default mode of authentication in some protocols (IKE, SSH) and optional in others (TLS).
+
+MTLS can be used without changes to the application or service code. The TLS is handled entirely by the service mesh infrastructure and between the two sidecar proxies.
+
+By default, {ProductName} is set to permissive mode, where the sidecars in {ProductShortName} accept both plain-text traffic and connections that are encrypted using mTLS. If a service in your mesh is communicating with a service outside the mesh, strict mTLS could break communication between those services. Use permissive mode while you migrate your workloads to {ProductShortName}.
+
+== Enabling strict mTLS across the mesh
+
+If your workloads do not communicate with services outside your mesh and communication will not be interrupted by only accepting encrypted connections, you can enable mTLS across your mesh quickly. Set `.Spec.istio.global.mtls.enabled` to `true` in your ServiceMeshControlPlane resource. The operator creates the required resources.
```
JStickler, line 16: `spec.istio.global.mtls.enabled`  lower case 'spec' and I'm not sure we need the period in front of 'spec'?   Maybe confirm with one of the developers what the standard is.

rcernich: Yep, s/b lowercase

modules/ossm-security-mtls.adoc
```diff
@@ -0,0 +1,80 @@
...
+    global:
+      mtls:
+        enabled: true
+----
+
+[id="ossm-security-mtls-sidecars-incoming_{context}"]
+== Configuring sidecars for incoming connections with a policy
+
+Create a ServiceMeshPolicy to enforce mTLS encryption on incoming connections. It instructs all sidecars to reject plain text connections and only accept mTLS connections. The resource must be named `"default"` and it must be created in the same project as your ServiceMeshControlPlane resource. Change `<CONTROL_PLANE_NAMESPACE>` in the following sample to your ServiceMeshControlPlane namespace.
+
```
rcernich, line 33: This resource is created when you set mtls.enabled=true.

rcernich:
Added a few more comments.  Other than those, looks good.

modules/ossm-security-mtls.adoc
rcernich: Yep.  Looks good.

modules/ossm-security-mtls.adoc
```diff
@@ -0,0 +1,63 @@
...
+  istio:
+    global:
+      mtls:
+        enabled: true
+----
+
+[id="ossm-security-mtls-sidecars-incoming-services_{context}"]
+=== Configuring sidecars for incoming connections for specific services
+
+You can also configure mTLS for individual services or namespaces by using a `Policy` instead of `ServicMeshPolicy`.
```
rcernich, line 32: Not sure if you want to leave `ServiceMeshPolicy` here, as it's configured through mtls.enabled=true

neal-timpe: @rcernich Ok, good catch. I removed that and added a line about making sure to change the control plane namespace. However, I have two questions, in Istio, policies are deprecated. Is that still our strategy? And how does this policy apply to a specific service? I don't see a place to add the service name.

Should the <CONTROL_PLANE_NAMESPACE> variable be <Namespace>?

modules/ossm-security-mtls.adoc
```diff
@@ -0,0 +1,63 @@
...
+  istio:
+    global:
+      mtls:
+        enabled: true
+----
+
+[id="ossm-security-mtls-sidecars-incoming-services_{context}"]
+=== Configuring sidecars for incoming connections for specific services
+
+You can also configure mTLS for individual services or namespaces by creating a policy. Change the following example to include your ServiceMeshControlPlane namespace.
```
rcernich, line 32: I think you can remove the last sentence here (same below for DestinationRule), as these can be created in any namespace in the mesh and, if named default, will serve as the default policy for that namespace.  Regarding the control plane namespace, I think it's best to leave it out of the docs, as it should be configured through mtls.enabled in the SMCP to avoid issues with control plane services (e.g. prometheus, jaeger, etc.).

rcernich: lgtm

rcernich (approved):

modules/ossm-security-mtls.adoc
```diff
@@ -0,0 +1,63 @@
...
+----
+
+[id="ossm-security-mtls-sidecars-incoming-services_{context}"]
+=== Configuring sidecars for incoming connections for specific services
+
+You can also configure mTLS for individual services or namespaces by creating a policy.
+
+[source,yaml]
+----
+apiVersion: "authentication.maistra.io/v1"
```
JStickler, line 36: I don't think we need any of the quotes in these example YAML files.  I don't think they break anything, but I don't think they're necessary either.

JStickler (approved):
Just one small comment, other than that LGTM.

neal-timpe:
@yxun I tagged Tiago and Guillerme also. If you'd like to take a look, it's ready for a QE review.

yxun (approved):

PR #22484, opened 2020-05-29
Update TP Table in 4.4 RN to include Pipelines

nkakkar81:
Uncommented the Pipelines entries in the 4.4 TP RN Table


ahardin-rh:
Looks good!

